PR #613, opened 2018-01-22
Typo3 support for ddev, fixes #570

### LINKED ISSUE #570: Support TYPO3 CMS

* Feature Request tied to [v1.2.0](https://github.com/drud/ddev/wiki/roadmap#120-cms-support) of the roadmap.

We've tested and [manually stood up a TYPO3 site](https://github.com/drud/ddev/issues/500) to get a sense of what would be necessary to make it a supported, 1st class citizen alongside WordPress and Drupal 7/8. Once we [refactor how we handle CMS specific logic](https://github.com/drud/ddev/issues/535), we will be using that as a way to officially support TYPO3.

More details and acceptance criteria to follow once we get within 2 sprints of starting this issue.

rfay:
## The Problem/Issue/Bug:

We want to officially detect and support typo3.

* Typo3 is detected
* Import-db and related actions work

Spin up a few typo3 sites, and import the db for them. I would imagine we'll learn something along the way. 

The assumption (enforced by the code here) is that typo3 must already have been through an installation process and a typo3conf/LocalConfiguration.php has been created *by typo3*. On import-db, we add an AdditionalConfiguration.php that overrides the database credentials.

Typo3 creates many more files and has much more outside-the-db configuration than we've experienced before, so I'm not entirely sure what the result of that will be.

Tests still being visualized.

## Related Issue Link(s):

OP #570 - "Official support for typo3"

## Release/Deployment notes:

It looks to me like typo3 is pretty specific about versions. For example. a database from 8.0.0 doesn't seem to work in 8.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,9 @@ ddev is an open source tool that makes it simple to get local PHP development en
 2. **Install ddev:** Options include [macOS homebrew](https://ddev.readthedocs.io/en/latest/#homebrew-macos) (recommended), an [install script](https://ddev.readthedocs.io/en/latest/#installation-script-linux-and-macos), or [a manually download](https://ddev.readthedocs.io/en/latest/#manual-installation-linux-and-macos).
 3. **Choose a CMS Quick Start Guide:** 
   - [WordPress](https://ddev.readthedocs.io/en/latest/users/cli-usage#wordpress-quickstart)
-  - [Drupal 7](https://ddev.readthedocs.io/en/latest/users/cli-usage#drupal-7-quickstart)
+  - [Drupal 6 and 7](https://ddev.readthedocs.io/en/latest/users/cli-usage#drupal-6/7-quickstart)
   - [Drupal 8](https://ddev.readthedocs.io/en/latest/users/cli-usage#drupal-8-quickstart)
+  - [TYPO3](https://ddev.readthedocs.io/en/latest/users/cli-usage#typo3-quickstart)
 
 Having trouble? See our [support options below](#support). Additionally, you may have trouble if [another local development tool is already using port 80 or 443](https://ddev.readthedocs.io/en/latest/#using-ddev-with-other-development-environments).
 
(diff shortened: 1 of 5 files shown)


cweagans:
Would it be possible to add a test for Typo3 in ddevapp_test.go? I think those tests add a lot of value and make sure that we don't break things. If ddev is going to be widely used in the Typo3 community, I'd really like to make sure that we don't accidentally break support.

cweagans:
Other than the testing note, this looks great! I haven't done any manual testing at this point, but from a code standpoint, everything seems pretty solid.

rfay:
Thanks, please do get some time for a manual check. I know that costs learning effort.

cweagans:
Will do, but probably tomorrow.

cweagans:
Testing notes:

* Detection worked and container spin up worked, but hitting the site in a browser, I get this:
```
This version of TYPO3 CMS requires PHP 7.2 or above
```

* Turns out that Typo3 v9.0.0 won't work with our PHP container _at all_. This is a problem that we should probably resolve soonish if Typo support is going to be a big thing for us.
* Blew everything away and downgraded to Typo3 v8.7.9
* Installation and everything worked great. Created a page and a template and dumped the db)
* Blew everything away again (including ~/.ddev/typo3)
* Created a brand new typo3 environment
* Went through the installer again (to regenerate config) and used a different admin password
* `ddev import-db`
* Logged in with admin password for first install
* Verified that the page I created in the first install was all there

I think we should move forward with this, and add support for Typo3 v9.0.0 in a followup.

cweagans:
(sorry about that - misclick)

rfay:
Thanks! - Will do a quick check-in with them about typo3 v9.

rfay:
Asked in typo3 #ddev and they're good with 7.1 for now. They say there's nothing currently that requires 7.2 anyway, and of course we should support it in the future as we will.

cweagans:
> They say there's nothing currently that requires 7.2 anyway

Their download page and installer disagree: http://take.ms/MmCCU  -- Typo3 v9.0.0 bombs out right out of the gate for me with our PHP container. If it's not a big deal to support that version of Typo3 right now, that's fine, but we should at least make a note on the docs page that support is only in place for up to v8.x.x to set expectations appropriately.

PR #618, opened 2018-01-24
Make ddev config work with args when no --apptype provided, fixes #616

### LINKED ISSUE #616: ddev v0.12.0 config command requires --apptype and shouldn't

Somewhere we made it so ddev v0.12.0 `ddev config --docroot=docroot` can't work because the apptype detection doesn't get to the right place at the right time. (It works with --apptype)

```
ddev config --docroot=.
You are reconfiguring the app at /Users/rfay/workspace/randyfay.
apptype must be one of drupal7, drupal8, wordpress, drupal6, typo3, php
```

This breaks ddev-ui, with configures sites expecting autodetect to do the job.

rfay:
## The Problem/Issue/Bug:

In the multi-CMS rework, I seem to have made it so you can't use `ddev config --docroot=.`, for example, without specifying the apptype (you have to do `ddev config --apptype=drupal6 --docroot=.`

This problem breaks ddev-ui (which does not provide --approot)

Check to see if appRoot is empty and if so detect. The detection was already there, but the failure here was higher up.

* Run `ddev config --docroot=.` and see if it works. Try other combinations of things with ddev config. Look at `ddev config -h`.
* Test with ddev-ui. Without this patch, "add from repo" is nonfunctional. With this patch it should work OK.

## Related Issue Link(s):

OP #616

## Release/Deployment notes:

This adds a new flag --projectroot, which is the same as --approot. That should become the new standard. But since people don't use ddev config with flags, we should be ok.

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
@@ -32,7 +33,7 @@ var appType string
 // ConfigCommand represents the `ddev config` command
 var ConfigCommand = &cobra.Command{
 	Use:   "config [provider]",
-	Short: "Create or modify a ddev application config in the current directory",
+	Short: "Create or modify a ddev project configuration in the current directory",
 	Run: func(cmd *cobra.Command, args []string) {
 
 		appRoot, err := os.Getwd()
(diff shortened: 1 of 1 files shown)


rickmanelius:
Not necessary for this PR, but noting that we should probably change --sitename to --projectname as it relates to this meta https://github.com/drud/ddev/issues/526

PR #622, opened 2018-01-25
Support router_http_port and router_https_port in config.yaml, fixes #481

### LINKED ISSUE #481: Allow webserving on an alternate port instead of port 80/443

In #126 we'll solve the problem of explaining gracefully to people that their port 80 is already occupied, but some people may want to leave it that way. Other people don't know how to find out what is taking up the port, etc., and may want to just move ddev's webserving to a different port. We should consider providing configuration for that. 

This type of change may be pretty intrusive, because we'll have to add the port number to every reference of the URL (when it's not 80/443). That's a bit of impact on the code, but it's also less useful for developers, because in general they would not want to see a port in their URL, they want to see URLs that are structurally equivalent to production.

However, we should consider providing it.

In addition, the closed #323 and #376  advocate for reporting to users the actual localhost port binding for the webserver, so that should be done in this issue.

rfay:
## The Problem/Issue/Bug:

OP #481: We need to allow people to choose a different router port

New elements in config.yaml for each site allow the user to override router_http_port and router_https_port

1. Stop the project and delete the .ddev/docker-compose.yml (New respect for an environment variable in the docker-compose.yml is required)
2. Add "router_http_port: 8080" to a site's config.yaml and `ddev start`. The correct URL should show everywhere that URLs are displayed.

## Release/Deployment notes:

- [ ] Depending on how ddev-ui calculates or uses the URL in describe, it may have to add respect for the port number.

### DIFF
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -116,7 +116,8 @@ func StartDdevRouter() error {
 }
 
 // findDdevRouter usees FindContainerByLabels to get our router container and
-// return it
+// return it. This is currently unused but may be useful in the future.
+// nolint: deadcode
 func findDdevRouter() (docker.APIContainers, error) {
 	containerQuery := map[string]string{
 		"com.docker.compose.service": RouterProjectName,
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -59,7 +59,7 @@ func StartDdevRouter() error {
 	if _, err = os.Stat(certDir); os.IsNotExist(err) {
 		err = os.MkdirAll(certDir, 0755)
 		if err != nil {
-			return fmt.Errorf("unable to create directory for ddev router: %v", err)
+			return fmt.Errorf("unable to create directory for ddev certs: %v", err)
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


rfay:
This now has docs and tests, and the formatting problem Rick caught should be fixed.

pkg/ddevapp/router.go
```diff
@@ -87,17 +86,20 @@ func StartDdevRouter() error {
...
-	if err != nil || container.State != "running" {
-		err = CheckRouterPorts()
-		if err != nil {
-			return fmt.Errorf("Unable to listen on required ports, %v,\nTroubleshooting suggestions at https://ddev.readthedocs.io/en/latest/users/troubleshooting/#unable-listen", err)
-		}
+	// Since the ports in use may have changed, stop the router and make sure
+	// we can access all ports.
+	// It might be possible to do this instead by reading from docker all the
+	// existing mapped ports.
+	_, _, err = dockerutil.ComposeCmd([]string{routerComposePath}, "-p", RouterProjectName, "down", "-v")
```
tannerjfco: Do we actually need to stop the router for this? Shouldn't changes to the ports be picked up by running "docker-compose up" again?

rfay: docker-compose up would fail if the ports weren't available right? And show some complaints?

PR #639, opened 2018-02-06
WIP Use Buildkite for macOS and Windows testing

cweagans:
## The Problem/Issue/Bug:

We're switching to Buildkite. This PR will contain the necessary changes to make it happen.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,33 +1,41 @@
 #!/bin/bash
 
-# This script is used to build drud/ddev using surf 
-# (https://github.com/surf-build/surf)
+# This script is used to build drud/ddev using buildkite
 
-# Manufacture a $GOPATH environment that can mount on docker (when surf build)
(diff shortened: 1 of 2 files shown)


cweagans:
Do we need to keep Surf around? I thought the idea was to replace Surf with Buildkite. Are you saying that you want to run them in parallel for a while?

rfay:
There's no intent to keep surf, but imo it never helps to burn your bridges during development. Better to get the new thing in place without destroying the old, then retire the old.

cweagans:
Fair enough. I can get behind that. Probably going to change up the approach here a bit anyhow, so it shouldn't interfere with the surf stuff. I've got some other things to do before I can get back to this though, so it'll be a bit before that happens.

rfay:
I fully support this, but there's no expectation that it will be completed soon, so I'm going to temporarily close it, we'll reopen when it's time to work on it again.

PR #640, opened 2018-02-06
Add TYPO3 to the list of quickstart guides

beeradb:
## The Problem/Issue/Bug:
The wonderful @horncologne noticed we don't have TYPO3 in the list of quickstart guides at the top of this document.

Add TYPO3 to the list!


beeradb:
Looks like my editor nabbed some trailing whitespace in here too. I'm usually for removing trailing whitespace where possible, but I can rebase that out if there are objections.

Issue #637, opened 2018-02-05
Question about documentation configuration 


nearlythere:
Is this by design? Seems like these links are missing, and I don't know if it's intentional or an error with RTD. Looks like an error. 

## What you expected to happen:

- Missing in sidebar: download links to versions and formats. 
- Missing in upper right hand corner:  "Edit on GitHub"

Examples of what I expected to see (screenshotted from other projects)

<img width="257" alt="missing-edit-github" src="https://user-images.githubusercontent.com/330552/35816412-8cdaa1e8-0a92-11e8-8690-6fec3fd4bd66.png">

<img width="317" alt="read-docs-downloads" src="https://user-images.githubusercontent.com/330552/35816304-467c5908-0a92-11e8-94e4-37b8088ba066.png">

## How to reproduce this:

Visit the site https://ddev.readthedocs.io/ 

## Anything else do we need to know:

These links help people know which version of docs they are looking at and how to suggest improvements.

rfay:
Assigned to @tannerjfco for review. Thanks for pointing this out, it seems like the approach in some other projects could help us.

rickmanelius:
Related issue https://github.com/drud/ddev/issues/630. This is yet another issues/snag we've hit upon using readthedocs. It might be worth re-evaluating if build/search issues persist.

tannerjfco:
We've gained the "Edit on Github" link with the fix in #641, but unfortunately it looks as though the workaround we're using to fix search breaks the RTD footer pop-up menu. Given that we only provide one version of our docs and dont have PDF or Epub versions, I don't think the loss of this menu impacts our project all that much. If the situation with search issues changes, or a better workaround surfaces, we may be able to have both working, but I think we've addressed what we can at this time.

Thanks for letting us know about these issues!

nearlythere:
@tannerjfco @rfay Can we reopen this? Or is it simply "won't fix" at this time? (I couldn't find info about the search issue, is that related to the RTD service?)

I should have specified in the original post: We would like to have downloadable PDFs for installation instructions, especially for use in onboarding newcomers at sprints and training. 

If they can be elegantly generated from Docs directly this would ensure the materials are up to date.

rickmanelius:
Linking to our documentation audit #242 because we need to do a more significant update/upgrade. To that end, I would prefer to solve that there at both a content and a structure level. If that works, let's close this sub-issue out, @nearlythere!

nearlythere:
@rickmanelius sounds good to me!

Issue #630, opened 2018-02-02
Wrong link to troubleshooting documentation (port 80 being occupied) 


tmotyl:
## What happened (or feature request):
during `ddev start` command en error occurred with information:

```
Failed to start core-master: Unable to listen on required ports, localhost port 80 is in use,
Troubleshooting suggestions at https://ddev.readthedocs.io/en/latest/users/troubleshooting/#unable-listen 
```
Unfortunatelly the url result in 404

## What you expected to happen:

To see a troubleshooting page based on https://github.com/drud/ddev/blob/master/docs/users/troubleshooting.md

## Version: Please include the output of `ddev version`, `docker version` and the project's .ddev/config.yaml.

ddev version
web   	drud/nginx-php-fpm-local:v0.9.1
db    	drud/mariadb-local:v0.7.1      
dba   	drud/phpmyadmin:v0.2.0         
router	drud/ddev-router:v0.5.0        
commit	v0.12.0                        
domain	ddev.local                     
cli   	v0.12.0  

## Related source links or issues:
https://github.com/drud/ddev/issues/542

rickmanelius:
Hi @tmotyl. I can confirm the issue. Readthedocs is not picking up the page. We will get this updated in the next sprint.

tannerjfco:
This has been fixed with #641, thanks for letting us know!

PR #641, opened 2018-02-07
add troubleshooting to toc, update mkdocs meta

tannerjfco:
## The Problem/Issue/Bug:
#630 - Troubleshooting doc missing from readthedocs site
#637 - Missing "Edit on Github" links on readthedocs pages, missing readthedocs footer

- The troubleshooting docs page was missing from the mkdocs.yml TOC. Adding it restores the missing page and fixes the link we provide in ddev that's currently broken
- Updates the mkdocs.yml metadata to get the "Edit on Github" links to show in readthedocs. From what I can tell, this update makes it a static link back to the ddev github page. The only examples of RTD sites I can find with more specific edit links are using RST. Similarly, any of the markdown driven sites I could find with the edit link were static to their repo homepages as well.

How it doesn't :( :

- readthedocs has a footer in the menu sidebar that pops-up, with links to EPUB, PDF versions of the site, and any available alternate versions of the docs. On our site, this pop-up is empty. This appears to be due to the workaround we implemented to resolve the broken docs search. I checked other sites with the search workaround and similarly found empty footer pop-ups. IMO, search is far more important than this footer menu, especially for our project. I don't believe the the PDF and EPUB options work for our project, and we only provide one version of our docs, so there are no other versions to navigate to. I propose that we accept the lost footer and keep the search.

An alternate version of the docs site is temporarily online to preview this change: http://ddev.readthedocs.io/en/doc-fixes/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,7 @@
 site_name: ddev Documentation
-edit_uri: edit/master/docs/
+repo_url: https://github.com/drud/ddev
+repo_name: GitHub
+edit_uri: blob/master/docs/
 copyright: DRUD Technology, LLC
 theme: readthedocs
 extra_javascript:
@@ -20,5 +22,6 @@ pages:
     - 'Integration with Hosting Providers':
       - 'Pantheon': 'users/providers/pantheon.md'
     - 'Notes for Linux Users': 'users/linux_notes.md'
+    - 'Troubleshooting': 'users/troubleshooting.md'
   - 'Developer Documentation':
     - 'Building, Testing, and Contributing': 'developers/building-contributing.md'


cweagans:
Yeah, please ignore Buildkite for now. It's not reliably passing yet.

rickmanelius:
Noting that this hasn't appeared on the official docs yet. If I made a mistake breaking protocol by merging before @tannerjfco reviewed, I apologize. 
/me slaps his own wrists.

tannerjfco:
The RTD build failed on merge, so the site didn't update. I've kicked the build, but I'm not sure why we didn't get a notification for the failed build. At any rate, the docs site is now updated.

PR #644, opened 2018-02-09
Add intl extension for TYPO3

cweagans:
## The Problem/Issue/Bug:

TYPO3 requires the intl extension

The container already has the intl extension in v0.9.2, so we should use that version.

No user facing changes other than the extension being installed. `php -i | grep intl` should return some things in the container.

No automated test changes.

## Related Issue Link(s):

https://github.com/drud/ddev/issues/634


rickmanelius:
Some additional info. When I login and run `php -i`, I get the following error:

```
root@4622a522f474:/var/www/html# php -i
php: error while loading shared libraries: /usr/lib/x86_64-linux-gnu/libssl.so.1.0.2: invalid ELF header
```

cweagans:
I cannot reproduce this. Can you let me know specifically what you're doing to see that problem?

PR #646, opened 2018-02-12
Remove parsing of invited_by_id because it breaks and we don't use it, fixes #645

### LINKED ISSUE #645: json parsing error running `ddev config pantheon`

* Error while setting up a Drupal 8 Pantheon site using the `ddev config pantheon` method.
* Error Message: "json: cannot unmarshal object into Go struct field Site.invited_by_id of type String". See attachment.
* From the user: "The only not-completely-ordinary thing I did was change local branches after I cloned."

* I was able to confirm that this was not an issue for a D7 repo being configured using ddev v0.12.0. I was able to fully execute `ddev config pantheon` with a working .ddev/config.yml file and spun up a site using `ddev start`.

## How to reproduce this:

* Need to test against a Windows 10 Pro machine.

* @ultimike was/is running DDEV v0.12.0

1-2

rfay:
## The Problem/Issue/Bug:

The invited_by_id field is apparently not always what we expect, and we don't use it. So stop parsing/using it.

This PR brings in v0.1.0 of go-pantheon, which just ignores the invited_by_id, https://github.com/drud/go-pantheon/pull/14

OP: #645 

Basically we only know one person who can reproduce this. He's tested it. It's probably good to go. We don't know what the bad json looks like.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,7 +4,10 @@
 [[projects]]
   branch = "master"
   name = "github.com/Azure/go-ansiterm"
-  packages = [".","winterm"]
+  packages = [
+    ".",
+    "winterm"
+  ]
   revision = "19f72df4d05d31cbe1c56bfc8045c96babff6c7e"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


rickmanelius:
Given this was approved in the upstream repo, I would vote for a +1, but I think @beeradb would be more appropriate. Adding him to the reviewer field.

PR #656, opened 2018-02-18
Make drupal 7 and drupal 8 settings distinct, recover settings[hash_salt], fixes #647

### LINKED ISSUE #647: generated settings incorrect for drupal8: use $settings['hash_salt] not $drupal_hash_salt

Our generated settings.php/settings.local.php for D8 is incorrect; we're settings $drupal_hash_salt, but for d8 it should be $settings['hash_salt'].

rfay:
## The Problem/Issue/Bug:

Drupal8-specific config values got lost in the appmatrix refactor, we need to get them back. The most obvious was the loss of `$settings["hash_salt"]`

* Splits the d7 and d8 config sections. 
* Uses the right one for the detected apptype
* Removes IsDrupal8 from the config struct

I'm not happy with the very, very not-DRY way we're doing Drupal and Backdrop settings creation. Loads of code and it's all repeated. However, it's what we've decided to go with for now, and this just does it again.  Maybe sometime when we're feeling more pure we can fiddle with it and clean it up.

* Do a ddev import-db on a d7 and d8 site, and verify the settings.local.php that gets created.

## Related Issue Link(s):

OP #647

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -50,17 +50,17 @@ var appTypeMatrix map[string]AppTypeFuncs
 func init() {
 	appTypeMatrix = map[string]AppTypeFuncs{
 		"php": {},
+		"drupal6": {
+			createDrupal6SettingsFile, getDrupalUploadDir, getDrupal6Hooks, setDrupalSiteSettingsPaths, isDrupal6App, nil, drupal6ConfigOverrideAction, drupal6PostConfigAction,
+		},
 		"drupal7": {
-			createDrupalSettingsFile, getDrupalUploadDir, getDrupal7Hooks, setDrupalSiteSettingsPaths, isDrupal7App, nil, drupal7ConfigOverrideAction, nil,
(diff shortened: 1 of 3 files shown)


rickmanelius:
Updating reviewers.

@rfay Regarding the need to consolidate code on the database credential file management. I think we're going to hit this this relatively soon with the potential splitting up of TYPO3 v8 and v9  (maybe). Should we file a follow-up issue or is the LOE less than a day and therefore worth tackling now versus later?

rfay:
Any TYPO3 work should be completely unrelated to this ticket, which affects only d7/d8. 

Also, we're just learning here about TYPO3 and haven't even quite stabilized with v8, so it's better to get v8 fully working and then add v9. Especially since those with even modest TYPO3 skill can already do what they need to do with v9.

rickmanelius:
@rfay I agree the scope of this is related to D7/D8 and not TYPO3. I was more focused on the comment that suggested refactoring and consolidating the code for existing CMS support, which would also help if/when additional CMS's are added OR we have additional distinctions between versions.

Let's proceed as per the PR for now, but you bring up an excellent point/concern that warrants a follow-up issue (if it doesn't already exist).

Issue #501, opened 2017-10-09
Proposal: Package ddev in a manner that is turnkey for Drupalcon Code Sprints


rickmanelius:
## What happened (or feature request):

Feature Request

## What you expected to happen:

As per a conversation with @horncologne, those leading code sprints are currently evaluating the possibility of a Docker based stack for sprint camps at Camps/Cons. This is a use case envisioned a while ago and we are highly motivated to make it happen.

## Reference

* Internal collaboration document to bake into a proposal.
https://docs.google.com/document/d/1slnml-Y_XRPEcmmd6XNK21xp01g1NR8AFnSkIy3DZ9A/edit

## Participants

* @rickmanelius
* @rfay 
* @cweagans 
* @cweagans 
* @horncologne

## Acceptance Criteria

* [x] Brainstorm and document ideas on how to address this in a small/medium/large bucketing (@cweagans and @rfay)
* [ ] Split out a new issue to address the most basic (small) use case of providing docker images in the tarball (@rickmanelius)
* [ ] Consolidate efforts across all fronts, including @BrianGilbert and @HornCologne.
* [ ] Generate a proposal that addresses the medium/large LOE versions (@rickmanelius, @cweagans, and @rfay )
* [ ] Circulate to code sprint leads at Drupalcon/Drupal camps to gauge their interest and solicit feedback (@rickmanelius)
* [ ] Convert this from a proposal to a deliverable with specific actions, acceptance criteria, deadlines, etc.

rickmanelius:
Current notes from a meeting w/@rfay and Brian Gilbert
https://docs.google.com/document/d/1OJucfCMiToVf05tZaUzlsM5BPD61A0eZKwcR9RDcxWs/edit#

rickmanelius:
Next actions:

* [ ] Reach out to Brian for confirmation of notes.
* [ ] Break out research acceptance criteria

rickmanelius:
@cweagans Just drawing your attention to this.

rickmanelius:
Trying to consolidate this because this is spanning 5 different individuals on the team that are all going to be a part of this process. @cweagans and @rfay had some great ideation on what an implementation strategy would look like (whether we were actively there or not). We also have @CRReigan looking to help coordinate outreach with the various camp organizers. Additionally, we have our partners at OSP starting to generate some materials that could prove to be of value.

[...]

PR #662, opened 2018-02-19
Add docker images to artifacts for #501

rfay:
## The Problem/Issue/Bug:

For OP #501 one of the things we want is to prevent the initial docker pull that happens initially, so bundle all the containers in the artifacts we provide. 

* Change "osx" to "macos" in artifact names
* Add ddev_docker_images.tar to the artifacts, so all can be downloaded at once.

* Delete your drud containers any way you like
* Use the install script to install ddev
* Download the ddev_docker_images.tar
* Load the images with `docker load -i /path/to/ddev_docker_images.tar`
* Verify with `docker images | grep drud` that the new images were loaded
* When you do your first "ddev start" nothing should have to be pulled.

This didn't fully work out as I had hoped. I had been planning to put the images into the download tarball, but that resulted in a tarball over 1G, which would be an unacceptable download for many people. Instead, the docker images are in a separate tarball.

We should consider whether the script should prompt about downloading it, or ask if it's already available.

## Related Issue Link(s):

OP #501  - turnkey install for sprints.

### DIFF
--- a/.circleci/generate_artifacts.sh
+++ b/.circleci/generate_artifacts.sh
@@ -8,7 +8,11 @@ BASE_DIR=$PWD
 sudo mkdir $ARTIFACTS && sudo chmod 777 $ARTIFACTS
 export VERSION=$(git describe --tags --always --dirty)
 
+# Make sure we have all our docker images, and save them in a tarball
 $BASE_DIR/bin/linux/ddev version | awk '/drud\// {print $2;}' >/tmp/images.txt
+for item in $(cat /tmp/images.txt); do
+  docker pull $item
+done
 docker save -o /tmp/docker_images.tar $(cat /tmp/images.txt)
 
 # Generate and place the extra items like images and autocomplete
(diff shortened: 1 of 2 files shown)


cweagans:
So what's the end result here? All ddev downloads are packaged with docker images included? Or we have new _additional_ packages with the images included?

rfay:
The end result was not what I hoped exactly.  ddev_docker_images.tar gets added to the artifacts, see https://circleci.com/gh/drud/ddev/2355#artifacts/containers/0 - I had intended it put it inside each tarball/zipball, but got cold feet when I found that our downloads were going over 1GB. I think we should provide the ddev_docker_images.tar as a separate file when in a faster environment.

cweagans:
Excellent. I think that's a really solid way to go.

rfay:
Oh wow, good point. I thought those docker images were already well-compressed. Added https://github.com/drud/ddev/pull/662/commits/67f29af1043e3692ce66c125cf987ddeefbecfec to gzip them.

rickmanelius:
Re-ran again after fulling blowing out my docker containers and images. I confirmed that ddev start had every container it needed on `ddev start`. +1

rfay:
My only hesitance about pulling this one is whether we want to do more with the install script in this PR. Should we make it ask "Do you want to download the images?", "Do you already have the images?". Or maybe we should just say at the end "You can get the images at ... and install with ..."

rickmanelius:
Just to document one of the conversations that we had during our sprint planning meeting. We want to make this zip more useful for end-users wanting to download and install the containers right from the get-go rather than waiting for the first `ddev start`. To that end, we would have to update our install script and our homebrew installation methods to include an optional download and import of this zip using the instructions above (e.g. docker load -i /path/to/ddev_docker_images.tar). This would make for a faster first-time experience by shifting the delay at install time versus the first usage.

PR #667, opened 2018-02-20
Updated build-tools to 1.5.5 (golang 1.10)

rfay:
## The Problem/Issue/Bug:

Updating everything to golang 1.10. This shouldn't change anything particularly.

### DIFF
--- a/build-tools/makefile_components/base_build_python-docker.mak
+++ b/build-tools/makefile_components/base_build_python-docker.mak
@@ -11,7 +11,7 @@ SHELL := /bin/bash
 
 PWD := $(shell pwd)
 
-BUILD_IMAGE ?= drud/golang-build-container:v0.5.2
+BUILD_IMAGE ?= drud/golang-build-container:v0.5.4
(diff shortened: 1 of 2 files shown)


build-tools/makefile_components/base_build_python-docker.mak
```diff
@@ -11,7 +11,7 @@ SHELL := /bin/bash
 
 PWD := $(shell pwd)
 
-BUILD_IMAGE ?= drud/golang-build-container:v0.5.2
+BUILD_IMAGE ?= drud/golang-build-container:v0.5.4
```
tannerjfco: shouldn't this be 0.5.5. too?

rfay: Really, it should be removed, as the BUILD_IMAGE is used nowhere. It's actually a result of me not catching it in the build-tools PR. 

I think I should do a PR to just remove this from the python-docker, pretty sure it can do no harm.

PR #669, opened 2018-02-21
Bump version to bring in separate-nginx-per-cms, fixes #648

### LINKED ISSUE #648: nginx config improvement for drupal 8: Don't do the ?q= part

These Drupal 8 redirect module issues point out problems with our nginx config for d8: https://www.drupal.org/project/redirect/issues/2895845 and https://www.drupal.org/project/redirect/issues/2870533

berdir seems to say that the ?q= config just needs to be removed for d8, https://www.drupal.org/project/redirect/issues/2895845#comment-12174240

I propose that we do a separate nginx config for every project type, and just link them in on container start. 

## How to reproduce this:

Using redirect module enabled with 
`route_normalizer_enabled: true`

seems to trigger this, infinite redirects. At least with the test db we have available.

rfay:
## The Problem/Issue/Bug:

https://github.com/drud/docker.nginx-php-fpm-local/pull/51 solves the problem in drud/ddev#648 where D8 doesn't work with redirect module. 

It also aims to fix related problems with TYPO3, and adds a TYPO3-specific config.

This PR just uses the image from that PR. 

It must be updated with an actual release before this can be pulled. 

OP drud/ddev#648
PR we're pulling in from nginx-php-fpm-local: https://github.com/drud/docker.nginx-php-fpm-local/pull/51

* Remove any .ddev in a project that uses redirect module (or TYPO3) and demonstrates the redirect problem.
* ddev config
* ddev start
* Verify that the problem is fixed.
* `ddev ssh` to verify the correct linked nginx config for the apptype in use. This can be tested with other apptypes as well


rfay:
Before this one can be (updated with new image) and pulled we need to make sure it has a full smoke test on drupal6, drupal7, drupal8, backdrop, wordpress, and TYPO3. All of those nginx configs are now separate, and there could be impact.

rfay:
We haven't had any specific feedback about changes needed for wordpress, I guess that's why I didn't add a wordpress version, so it uses the default. It's easy enough to copy the default and call it wordpress, but doesn't accomplish a lot.  I think it's worth looking for a wordpress config though and explicitly adding what they would normally do.

rfay:
I did a basic smoke test on this with drupal6, drupal7, drupal8, wordpress, backdrop, TYPO3. Included accessing important URLs, logging in, managing imagecache-type images. Sad that I found all kinds of problems. Glad that I seem to have solved them.

jdarrohn:
I was spun up a WordPress site successfully and went through some basic WP functions that I feel might be effected by nginx configuration:

- I successfully set permalinks to be the post's name
- I edited a permalink from within the post after saving
- I successfully created a custom post type with a rewrite rule on the slug for the archive page

I did find something interesting that I found a little concerning. I was going through my normal development workflow and hooked up this [WP CLI Login](https://aaemnnost.tv/wp-cli-commands/login/) plugin. It creates a "magic link" that, when clicked, will redirect and log you into the website. For some reason, the links were throwing a 404 when I used them, which I felt might be related to the configuration file. I tested `wp_redirect` in some debug code, and redirects appeared to be working fine.

Just wanted to bring this up because maybe it will uncover something larger. I wish I had more info, but I was unable to figure out exactly would could be causing that issue.

rfay:
I tested with WP 4.7.3 and 4.9.4 and WP-CLI v 1.2 and it worked out of the box with `wp login create rfay`, so going to write @jdarrohn 's problem off to some interaction/problem in the site he was working with, until we hear of further problems.

rfay:
@rickmanelius you've blocked this one as well, so please approve when you're ready, thanks!

PR #671, opened 2018-02-22
Add ddevapp tests for drupal6, backdrop, TYPO3, fixes #617 Pull on 2018-03-6 #TRIVIALREVIEW

### LINKED ISSUE #617: Add full tests for drupal6, backdrop, typo3

ddevapp_test.go has long had full downloads and various tests for a d7, d8, and wordpress. We should expand that to do our new project types, drupal6, backdrop, and typo3. 

I began just a bit of work on this [here](https://github.com/rfay/ddev/blob/20180123_drupal6_typo3_tests/pkg/ddevapp/ddevapp_test.go#L46-L68)

But it seems that we should get the typo3 support in rather than getting stuck on the tests.

rfay:
## The Problem/Issue/Bug:

We didn't add the full suite of tests for d6/backdrop/TYPO3 when implementing those and it's time.

This is not intended to solve the fact that we don't currently use any of the sites that we spin up with ddev, or see if they're working. That problem is deferred to https://github.com/drud/ddev/issues/677

This PR simply fills in test tarballs and related info for other CMSs, and does some minor improvements to the testing process. New tarballs are provided in the new repo set up for their storage.

Look at the Circleci test results and make sure everything is being hit. 

## Related Issue Link(s):

Fixes #617

### DIFF
--- a/cmd/ddev/cmd/logs_test.go
+++ b/cmd/ddev/cmd/logs_test.go
@@ -40,7 +40,7 @@ func TestDevLogs(t *testing.T) {
 		cleanup := v.Chdir()
 
 		confByte := []byte("<?php trigger_error(\"Fatal error\", E_USER_ERROR);")
-		err := ioutil.WriteFile(filepath.Join(v.Dir, v.DocrootBase, "index.php"), confByte, 0644)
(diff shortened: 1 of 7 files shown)


rfay:
I consider this a pretty non-intrusive PR. It adds a few CMSs to our stable. The improved testing is not great, but I think this can be pulled on 2018-03-06

PR #675, opened 2018-02-23
[BUGFIX] fix typo in install instruction output for bash completion

d3pendent:
## The Problem/Issue/Bug:
instruction for installing bash completion shows wrong filename "ddev_bash_completions.sh"
change filename in output to "ddev_bash_completion.sh"


d3pendent:
@rfay done ;)

Issue #545, opened 2017-11-20
Edge case. Running "ddev config" can detect D7 in a D8 repo


rickmanelius:
## What happened (or feature request):

* Download D8
* Run `ddev config`
* Mistakenly answer "core" as "docroot" instead of "./"
* The app type is detected as Drupal7

## What you expected to happen:

No app detection instead of a false positive for Drupal 7.

## Acceptance Criteria

* [ ] Entry in the ddev documentation that specifies apptype is now a suggestion versus a detection.
* [ ] Entry in the ddev documentation that notes that the php apptype is a generic fallback that will be used if a supported CMS is not detected (regardless of this is caused by user error OR it's simply not a folder strucutre containing a supported CMS).

rfay:
This will have changed now to detecting php. However, I don't know any way around it (except more automatic docroot detection). We're not going to detect the right thing in the wrong place.

I'd say we should either repurpose this and plan for more robust docroot detection, or else close this one.

rickmanelius:
Fair enough. I'll consider this done when we note that apptype is now a suggestion and that we have PHP as our generic fallback!

rickmanelius:
Issue acceptence criteria updated and milestone tagged for v1.2.0.

mglaman:
Here's my PHP toolstack identifier library https://github.com/mglaman/toolstack-helper/tree/develop/src/Stacks

One check is to see if it is a built or not Drupal site, from the days of yore with Drush make.

But the real identifier is

```
return (file_exists($dir . '/core/composer.json')) ? self::DRUPAL8 : self::DRUPAL7;
```

rfay:
This issue is outdated and we have much better detection now. However, with a bogus docroot, "php" would be detected, and I think that's correct behavior. 

I'm going to close this in favor of the real better alternative, [almost always detecting and suggesting the right docroot](https://github.com/drud/ddev/issues/663)

rfay:
Reopening so we at least get to see @mglaman's PR for this!

rfay:
Closing again. @mglaman's https://github.com/drud/ddev/pull/678 is going to be great for this, and a very adequate solution.

PR #679, opened 2018-02-26
Harden Drupal version detection #545

mglaman:
## The Problem/Issue/Bug:

Detecting the Drupal version can be weak if wrong docroot chosen.

There must be an `index.php` and contain the word "Drupal", then follow structure patterns.

### DIFF
--- a/pkg/ddevapp/apptypes_test.go
+++ b/pkg/ddevapp/apptypes_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/testcommon"
 	asrt "github.com/stretchr/testify/assert"
+	"io/ioutil"
 )
 
 // TestApptypeDetection does a simple test of various filesystem setups to make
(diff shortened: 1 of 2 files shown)


rickmanelius:
Added reviewers and placed on our project board.

rickmanelius:
I'll wait for @rfay to take a first pass on code/tests before I jump in.

rfay:
* I'm not convinced that this adds significant value over what was there before.
* We've been trying to deliberately treat d6/d7/d8 as separate things and not mix the code, so it's easier to maintain (despite some copying).

So I'm uncomfortable with this one.

mglaman:
@rfay  that's fine, we can close. The main thing is the `index.php` peek to ensure it's Drupal. That might help with other CMS support down the road. "We know that a file within X bytes will say its name"

rfay:
Thanks, sorry. I hate to have people spend time on things and it turn out not used. But it's just a different direction than the code has been going. Thanks so much for your absolutely excellent contributions on so many fronts.

PR #686, opened 2018-02-27
Change "app" and "application" to "project" in user-facing docs/output, fixes #526

### LINKED ISSUE #526: Standardize usage to "project", away from "app" or "site", at least in interacting with user

In ddev we use "app" and "site" quite a lot. We should probably use one or the other, at least in user-facing stuff.  I'm a fan of "site" because I think it's what we're generally working with.  To me "app" usually refers to a smartphone app or an application that runs on the OS, and a "site" usually means HTML/css/js served from a web server.  I know that the usage has started to shift a bit (like some companies with smartphone apps will refer to their "web app" as an alternative). 

Anyway, we should choose the terminology that is most meaningful to our developers and go with it consistently.

Edit: We've agreed (below) to go with "project" when interacting with the user.

## Remaining Tasks

* [ ] Search/Replace and ensure all ddev CLI tests pass.
* [ ] Open up a ticket for DDEV UI

rfay:
## The Problem/Issue/Bug:

OP #526: We're trying to use "project" everywhere when talking about a ddev site/app/whatever, in our docs and whenever ddev interacts with the user. This PR attempts to do that.

Use ddev. Look at the output. Review the docs.

## Related Issue Link(s):

OP #526

### DIFF
--- a/cmd/ddev/cmd/sequelpro.go
+++ b/cmd/ddev/cmd/sequelpro.go
@@ -27,7 +27,7 @@ var SequelproLoc = "/Applications/sequel pro.app"
 var DdevSequelproCmd = &cobra.Command{
 	Use:   "sequelpro",
 	Short: "Easily connect a dev site to sequelpro",
-	Long:  `A helper command for easily using sequelpro (OSX database browser) with a running ddev app.`,
+	Long:  `A helper command for easily using sequelpro (OSX database browser) with a running DDEV-Local project.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) != 0 {
 			output.UserOut.Fatalf("invalid arguments to sequelpro command: %v", args)
--- a/cmd/ddev/cmd/sequelpro.go
+++ b/cmd/ddev/cmd/sequelpro.go
@@ -26,8 +26,8 @@ var SequelproLoc = "/Applications/sequel pro.app"
 // DdevSequelproCmd represents the sequelpro command
 var DdevSequelproCmd = &cobra.Command{
 	Use:   "sequelpro",
-	Short: "Connect a project database to sequelpro",
-	Long:  `A helper command for using sequelpro (macOS database browser) with a running DDEV-Local project.`,
+	Short: "Connect sequelpro to a project database",
+	Long:  `A helper command for using sequelpro (macOS database browser) with a running DDEV-Local project's database'.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) != 0 {
 			output.UserOut.Fatalf("invalid arguments to sequelpro command: %v", args)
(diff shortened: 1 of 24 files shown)


rickmanelius:
Nice. Thanks for moving on this one!

beeradb:
I'd love to know we've thought through the implications for this change in DDEV Live, as any terminology change like this will inevitably effect that as well.

If it has been then 👍. Just chiming in to make sure.

rfay:
Certainly we have not thought through anything about DDEV-Live! But it doesn't make sense for DDEV-Live. For Pantheon, which is the analog of DDEV-Live, we use "site", which is the terminology used by Pantheon.

This transition has been in progress for quite some time, this is just the (hopefully, nearly) last bit.

rfay:
@rickmanelius loads o' changes in 190b9873cb6541618ae4434fd3e053373682b7b3 site->project

rickmanelius:
Loads of fun! Will review later tonight...

rickmanelius:
Hi @rfay. This is very close. One changed didn't make it in. I also found that some of the json responsese need to be vetted.

Version: v0.14.0-11-g7469fc97
* `ddev config` CONFIRMED CHANGES
* `ddev list`
  * This still says "List applications"
* `ddev sequelpro` CONFIRMED CHANGES
* `ddev start` CONFIRMED CHANGES
* `ddev stop` CONFIRMED CHANGES
* `-j`
  * `ddev config --projectname test --projecttype drupal7 --docroot . -j`
  	* This returns "You are reconfiguring the app at". Needs s/app/project/

Other details:
* Note we are going to have to be very clear in our release notes that there are BCs in place.
* There also may be BCs with DDEV-UI with respect to json output, which has not been tested.

rfay:
Thanks for the good eye Rick. It's so hard because we're not trying to change the code usages. "app" hides in there pretty well.

There shouldn't be any backward compatibility issues, as the old commands are still supported (but deprecated). 

I don't *think* the json output should have changed.

cmd/ddev/cmd/sequelpro.go
```diff
@@ -26,8 +26,8 @@ var SequelproLoc = "/Applications/sequel pro.app"
 // DdevSequelproCmd represents the sequelpro command
 var DdevSequelproCmd = &cobra.Command{
 	Use:   "sequelpro",
-	Short: "Easily connect a dev site to sequelpro",
-	Long:  `A helper command for easily using sequelpro (OSX database browser) with a running ddev app.`,
+	Short: "Connect a project database to sequelpro",
```
cweagans: I don't think this should hold up the PR, but I think something to the effect of "Connect to a project database with Sequel Pro" would be more accurate.

rickmanelius:
Yay!!! This was a big, tedious one :)

PR #690, opened 2018-02-28
Bump mariadb tag to bring in static db version

rfay:
## The Problem/Issue/Bug:

In https://github.com/drud/mariadb-local/pull/42 a new container uses static database initialization in order to speed our first-time-project startup. It improves startup times by double.


rfay:
#707 will be used to bring in the new images for ddev v0.15.0.

Issue #663, opened 2018-02-20
"Docroot Location" should pick up normal directory roots


mglaman:
## What happened (or feature request): 
Feature request

## What you expected to happen:
When I configured a composer based Drupal project, I figured it would pick up `web` or `docroot` when "Docroot Location"  was asked.

## How to reproduce this:
Run `ddev config`

## Related source links or issues:

Example of what I did in my tool `platform-docker`

```
                // Platform.sh scaffold docroot.
                if (is_dir($this->cwd . '/' . Platform::DEFAULT_WEB_ROOT)) {
                  Config::set('docroot', Platform::DEFAULT_WEB_ROOT);
                }
                // Typical app location.
                elseif (is_dir($this->cwd . '/web')) {
                    Config::set('docroot', 'web');
                }
                // Acquia.
                elseif (is_dir($this->cwd . '/docroot')) {
                    Config::set('docroot', 'docroot');
                }
                else {
                  $question = new Question("<info>What is the document root for the project?");
                  $answer = $helper->ask($input, $output, $question);
                  Config::set('docroot', $answer);
                }
```

source: https://github.com/mglaman/platform-docker/blob/develop/src/Command/InitCommand.php#L48

## Please use a complexity rating of 1-5 (5 is high) for a feature request. (High complexity implies more PR planning)

1

rfay:
I think this is totally reasonable. I'd think 

* probe those in order of most likely, looking for an index.php in them. 
* Prompt with the discovered docroot as default rather than just going with it.

I actually can't tell you how many times I've had to ctrl-c in the middle of a config operation or use another window to go find what the docroot is.

rickmanelius:
Flagging in the milestone to provide better CMS UX. There is a little bit of a challenge in that this may be CMS specific because some codebases (notably TYPO3) can sometimes have two index.php files and we need to have more than just a file/directory check to accurately detect with a 99+% hit rate.

mglaman:
So, I know this is in Go. But are option selects possible, like in Symfony Console?

" We detected these docroots, please select correct one", because don't some even use `index_dev.php`? So if someone want to point to dev or something crazy.

mglaman:
Okay, giving this a whirl. Looks like all the work needs to be done in `pkg/ddevapp/config.go:358` at `func (app *DdevApp) docrootPrompt() error {`

rickmanelius:
[...]

PR #693, opened 2018-02-28
Create settings file when configuring project, fixes #692

### LINKED ISSUE #692: Create project settings file (settings.local.php) during "ddev config" instead of import-db

When I ran `ddev config` and tried to install Drupal, I had to run `ddev describe` and copy my database information.

I expected a scaffolded CMS settings file which provided my required configuration.

## How to reproduce this:

Run `ddev config`. The `settings.php` is only generated if `ImportDB` is executed.

```
$ ddev version
commit	v0.14.0                        
domain	ddev.local                     
cli   	v0.14.0                        
web   	drud/nginx-php-fpm-local:v1.0.0
db    	drud/mariadb-local:v0.7.1      
dba   	drud/phpmyadmin:v0.2.0         
router	drud/ddev-router:v0.5.0  
```

This was due to debugging #454 and when working on #691.

1

mglaman:
## The Problem/Issue/Bug:

See issue #692

After writing the project config, also write a settings files via `app.CreateSettingsFile()`

## Related Issue Link(s):

#692

## Release/Deployment notes:

* No obvious break changes come to mind.

## Remaining Tasks
* [x] Error Messages: As discussed in [the following review ](https://github.com/drud/ddev/pull/693#pullrequestreview-101537297), there are syntax errors when attempting to output the message beginning with “Could not write settings file” for Drupal 6, 7, 8, Backdrop, and TYPO3.
* [ ]  WordPress Local Settings:
	* [ ] We need to replicate the pattern used for Drupal by defaulting to wp-config.php if it doesn’t exist and falling back to generating wp-config-local.php if #ddev-generated is not found in wp-config.php.
	* [x] We need to fix the 500 error for WP by correcting the require_once() [syntax described here](https://github.com/drud/ddev/pull/693#issuecomment-370833403) .
	* [ ]  We need to append an inclusion of wp-config-local.php within the wp-config.php template [similar to how D8 does it]([https://github.com/drupal/drupal/blob/8.5.x/sites/default/default.settings.php#L785).
* [x] Backdrop: We need to ensure the error message for Backdrop doesn’t state Drupal.
* [ ] TYPO3:
  * [x]  Convert Error to Warning: It's not an error to not have LocalConfiguration.php (but we should util.Warning() it.
  * [ ] If AdditionalConfiguration.php does not exist, create the typo3conf directory if it doesn't exist and add our AdditionalConfiguration.php to it.
  * [ ] If typo3conf/LocalConfiguration.php didn't already exist, util.Warning() them that they have to do an install manually.
* [ ] Documentation: Any end-user documentation necessary to instruct users on this new workflow (@rickmanelius can address).
* [ ] Config => Start: File an issue to address [this use case](https://github.com/drud/ddev/pull/693#issuecomment-370893215) where an end-user may need the settings file generated at `ddev start` (@rickmanelius)

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -165,7 +165,7 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 
 		_, err = app.CreateSettingsFile()
 		if err != nil {
-			util.Failed("Could not write settings file: %w", err)
+			util.Failed("Could not write settings file: %v", err)
 		}
 
 		// If a provider is specified, prompt about whether to do an import after config.
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -163,10 +163,7 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
-		_, err = app.CreateSettingsFile()
-		if err != nil {
-			util.Failed("Could not write settings file: %w", err)
-		}
+		_,_ = app.CreateSettingsFile()
 
 		// If a provider is specified, prompt about whether to do an import after config.
 		switch provider {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -152,7 +152,10 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			} else {
 				util.Success("Using project name '%s' and environment '%s'.", app.Name, pantheonEnvironment)
 			}
-			_ = app.ConfigFileOverrideAction()
+			err = app.ConfigFileOverrideAction()
+			if err != nil {
+				util.Failed("Failed to run ConfigFileOverrideAction: %v", err)
+			}
 
 		}
 		err = app.WriteConfig()
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -163,7 +163,7 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
-		_,_ = app.CreateSettingsFile()
+		_, _ = app.CreateSettingsFile()
 
 		// If a provider is specified, prompt about whether to do an import after config.
 		switch provider {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -152,10 +152,7 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			} else {
 				util.Success("Using project name '%s' and environment '%s'.", app.Name, pantheonEnvironment)
 			}
-			err = app.ConfigFileOverrideAction()
-			if err != nil {
-				util.Failed("Failed to run ConfigFileOverrideAction: %v", err)
-			}
+			_ = app.ConfigFileOverrideAction()
 
 		}
 		err = app.WriteConfig()
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -163,7 +163,10 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
-		_, _ = app.CreateSettingsFile()
+		_, err = app.CreateSettingsFile()
+		if err != nil {
+			util.Failed("Could not write settings file: %w", err)
+		}
 
 		// If a provider is specified, prompt about whether to do an import after config.
 		switch provider {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -163,6 +163,11 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
+		_, err = app.CreateSettingsFile()
+		if err != nil {
+			util.Failed("Could not write settings file: %w", err)
+		}
+
 		// If a provider is specified, prompt about whether to do an import after config.
 		switch provider {
 		case ddevapp.DefaultProviderName:
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -94,9 +94,11 @@ func IsValidAppType(apptype string) bool {
 func (app *DdevApp) CreateSettingsFile() (string, error) {
 	app.SetApptypeSettingsPaths()
 
-	// If neither settings file options are set, then don't continue
+	// If neither settings file options are set, then don't continue. Return
+	// a nil error because this should not halt execution if the apptype
+	// does not have a settings definition.
 	if app.SiteLocalSettingsPath == "" && app.SiteSettingsPath == "" {
-		return "", fmt.Errorf("Neither SiteLocalSettingsPath nor SiteSettingsPath is set")
+		return "", nil
 	}
 
 	// Drupal and WordPress love to change settings files to be unwriteable.
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -2,6 +2,7 @@ package ddevapp
 
 import (
 	"fmt"
+	"github.com/drud/ddev/pkg/util"
 	"os"
 	"path/filepath"
 )
@@ -98,6 +99,7 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 	// a nil error because this should not halt execution if the apptype
 	// does not have a settings definition.
 	if app.SiteLocalSettingsPath == "" && app.SiteSettingsPath == "" {
+		util.Warning("Project type has no settings paths configured, so not creating settings file.")
 		return "", nil
 	}
 
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -125,7 +125,10 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 	// just ignore.
 	if appFuncs, ok := appTypeMatrix[app.GetType()]; ok && appFuncs.settingsCreator != nil {
 		settingsPath, err := appFuncs.settingsCreator(app)
-		return settingsPath, err
+		if err != nil {
+			util.Warning("Unable to create settings file: %v", err)
+		}
+		return settingsPath, nil
 	}
 	return "", nil
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,7 @@ jobs:
       - run:
           command: ./.circleci/generate_artifacts.sh $ARTIFACTS
           name: tar/zip up artifacts and make hashes
+          no_output_timeout: "20m"
 
       - store_artifacts:
           path: /artifacts
@@ -79,6 +80,7 @@ jobs:
       - run:
           command: ./.circleci/generate_artifacts.sh $ARTIFACTS
           name: tar/zip up artifacts and make hashes
+          no_output_timeout: "20m"
 
       - store_artifacts:
           path: /artifacts
@@ -114,6 +116,7 @@ jobs:
       - run:
           command: ./.circleci/generate_artifacts.sh $ARTIFACTS
           name: tar/zip up artifacts and make hashes
+          no_output_timeout: "20m"
 
       - store_artifacts:
           path: /artifacts
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -57,6 +57,8 @@ func TestConfigWithSitenameFlagDetectsDocroot(t *testing.T) {
 	if err != nil {
 		t.Errorf("Could not create %s directory under %s", testDocrootName, tmpdir)
 	}
+	err = os.MkdirAll(filepath.Join(tmpdir, testDocrootName, "sites", "default"), 0755)
+	assert.NoError(err)
 	_, err = os.OpenFile(filepath.Join(tmpdir, testDocrootName, "index.php"), os.O_RDONLY|os.O_CREATE, 0666)
 	assert.NoError(err)
 
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -68,7 +68,7 @@ ini_set('session.cookie_lifetime', 2000000);
 func createBackdropSettingsFile(app *DdevApp) (string, error) {
 	settingsFilePath, err := app.DetermineSettingsPathLocation()
 	if err != nil {
-		return "", fmt.Errorf("Failed to get Backdrop settings file path: %v", err)
+		return "", fmt.Errorf("Failed to get Backdrop settings file path: %v", err.Error())
 	}
 	output.UserOut.Printf("Generating %s file for database connection.", filepath.Base(settingsFilePath))
 
@@ -76,7 +76,7 @@ func createBackdropSettingsFile(app *DdevApp) (string, error) {
 
 	err = writeBackdropSettingsFile(backdropConfig, settingsFilePath)
 	if err != nil {
-		return settingsFilePath, fmt.Errorf("Failed to write Backdrop settings file: %v", err)
+		return settingsFilePath, fmt.Errorf("Failed to write Backdrop settings file: %v", err.Error())
 	}
 
 	return settingsFilePath, nil
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -76,7 +76,7 @@ func createBackdropSettingsFile(app *DdevApp) (string, error) {
 
 	err = writeBackdropSettingsFile(backdropConfig, settingsFilePath)
 	if err != nil {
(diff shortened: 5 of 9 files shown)


mglaman:
This shouldn't run when `php` I suppose. Whoops!

cmd/ddev/cmd/config.go
```diff
@@ -160,6 +160,11 @@ var ConfigCommand = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
+		_, err = app.CreateSettingsFile()
+		if err != nil {
```
mglaman: Hrm. Maybe this isn't the correct spot to run this? Because it errors if the app type doesn't support settings. But we want to check that before running it.

rfay: CreateSettingsFile() should just do nothing if there's no settings support. If that's not what's happening then something is wrong.

mglaman:
Shoot. GitHub UI bad edit.

cmd/ddev/cmd/config.go
```diff
@@ -160,6 +160,8 @@ var ConfigCommand = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
+		_,_ = app.CreateSettingsFile()
```
rfay: We do have to capture and check the err here right? I don't think we need to capture the location of the file though for this context.

cmd/ddev/cmd/config.go
mglaman: Fixed to not throw an error if there is no support (ie: PHP apptype)

mglaman:
Tried to grok surf-windows error. Looks like randoms? Not sure.

rfay:
Yes, that's randoms. Never have figured out why that happens sometimes. 

surf is completely polling, so the only way to rerun (other than me going upstairs and doing it manually) is to push a change. When I have nothing to change, I just `git commit --amend` and change the message by adding a space or something and then `git push --force-with-lease` (only in a fork of course)

rfay:
If it seems right, I think we should fix https://github.com/drud/ddev/issues/704 in this PR as well. Basically CreateSettingsFile() returns an error if ddev can't manage any settings files, and that's wrong. It should be emitting a warning instead.

mglaman:
@rfay so change the fail to a warning in import-db command and call it 💯 ?

rfay:
Yes, just do a util.Warning() with that message and return a nil for the err

rickmanelius:
I would like to review this from an end-user perspective before this gets pulled. I have one more review ahead of this.

mglaman:
I screwed up `TestConfigWithSitenameFlagDetectsDocroot` test.

```
	Error Trace:	config_test.go:72
	Error:		"Creating a new ddev project config in the current directory (/tmp/config-with-sitename395145304)
			Once completed, your configuration will be written to /tmp/config-with-sitename395145304/.ddev/config.yaml
			
			Found a php codebase at /tmp/config-with-sitename395145304
			Using project name 'config-with-sitename' and environment ''.
			time="2018-03-04T23:02:21Z" level=debug msg="Config Directory does not exist, attempting to create." directory=/tmp/config-with-sitename395145304/.ddev
			" does not contain "Found a drupal8 codebase"
```

It tests the case where there is an empty `web` directory but also `docroot/index.php`

rickmanelius:
Hi @mglaman. We're *almost* there. It looks the only remaining issue is what you mentioned [regarding a test in your last comment](https://github.com/drud/ddev/pull/693#issuecomment-370524676). I'll wait until you're done before I review.

mglaman:
@rickmanelius Sorry! That comment was meant for #663 not here.

rickmanelius:
Ok great! I'll get this reviewed next.

rickmanelius:
Initial test with a Fresh Drupal 7.56 Site...

* `ddev config` generates settings.php w/#ddev-generated
* Re-running `ddev config` doesn't generate a settings.local.php, but overwrites the existing settings.php
* Removing #ddev-generated from settings.php and re-running `ddev config` generates a settings.local.php w/#ddev-generated
* Re-running `ddev config` overwrites the existing settings.local.php w/#ddev-generated
* Removing #ddev-generated in settings.local.php and re-running `ddev config` generates an error: "Could not write settings file: &{%!w(string=Failed to get Drupal settings file path: settings files already exist and are being managed by the user)}"

That last error is fine because such a file should not be checked into version control. However, the warning information should probably not contain "{%!w(string=" and it should probably suggest deleting or renaming settings.local.php to proceed.

rickmanelius:
I'll return to this later. Remaining things to test:

* Drupal 6 and 8
* Backdrop (comes with stock settings.php)
* WP (I don't believe they use a settings.local.php approach, and instructing to remove from version control could be problematic).
* TYPO3. Similar to WP. Need to understand what to do when a DB config file is checked in and whether a local.db.php version is acceptable.

mglaman:
IMO local development tools should be opinionated and document them. There's no reason WP or TYPO3 couldn't support local settings and document how they work with DDEV.

See my WP local configs https://github.com/mglaman/platform-docker/tree/develop/resources/stacks/wordpress

rfay:
@rickmanelius there are a pile of things required of a TYPO3 install, you can't actually just use a checkout with a db load. We can work with them to see if there might be a way around, but I'm not optimistic. In short: for TYPO3 an install must have been done at some point. THEN ddev can add the AdditionalSettings. But ddev does not attempt to create the LocalConfiguration.php for this reason. It has to have been created, and contains a pile of info that ddev could never infer.

rickmanelius:
@rfay Let me retest with a fresh checkout of TYPO3.

rfay:
So my comment above https://github.com/drud/ddev/pull/693#issuecomment-370779690 was wrong - the problem is not that this patch introduced an error, it's that it points out a chicken-and-egg problem with TYPO3. TYPO3 has to get installed on local *sometime*.  And with this patch we get an error in the early stages, so it at least appears that we are saying to the user they have no path forward.

It's possible we can just turn the complaint into a warning? And change the message? 

I don't think this patch is wrong, but we need to figure out how it is going to interact with TYPO3.

rfay:
I proposed over in the TYPO3 slack that we should consider just not whining about the missing LocalConfiguration.php.

mglaman:
We should update the summary with tasks. But as far as I can see, we need to:

-  [ ] Create a wp-config-local.php 
-  [ ] Backdrop: We need to replace s/Drupal/Backdrop/.
-  [ ] Grok TYPO3 error.

All of those seem to be missing features or

rickmanelius:
Hi @mglaman. I'll get the issue summary updated, but in terms of what falls in/out of scope, here's where I land. If the intent is to get up and running faster with a pre-baked db file, then I think the wp-config.php would fall into scope because of WP equally could benefit (although I may have found a regression on that one as well). TYPO3 we are in an active conversation in their Slack and we'll grok the importance of this before proceeding.

IMHO, there might be a need to categorize CMS support into two buckets: those that generate the settings during config and those that generate ONLY during an import.

mglaman:
> and those that generate ONLY during an import.

It sounds like TYPO3 would be in this bucket. That's unfortunate :/

rickmanelius:
When the #ddev-generated wp-config.php file is generated beforehand, this stanza results in a 500 error.

```
/**
Sets up WordPress vars and included files.

wp-settings.php is typically included in wp-config.php. This check ensures it is not
included again if this file is written to wp-config-local.php.
*/
if (__FILE__ == "wp-config.php") {
	require_once(ABSPATH . '/wp-settings.php');
}
```

I can correct it by looking at wp-config.php that WP generates during install time.

```
require_once(ABSPATH . '/wp-settings.php');
```

And modifying it as such

```
/**
Sets up WordPress vars and included files.

wp-settings.php is typically included in wp-config.php. This check ensures it is not
included again if this file is written to wp-config-local.php.
*/
if (__FILE__ == "/var/www/html/wp-config.php") {
	require_once(ABSPATH . '/wp-settings.php');
}
```

I know there is a better way to just get the file matching regex above, but I'm just noting this as the blocker. Since this relates to which CMSes can benefit from this approach, it would make sense to correct in this PR and then WordPress is all set. The remaining open loop is TYPO3.

rfay:
For TYPO3 my current thinking is let's just notify that the LocalConfiguration.php has not been set up yet and that an install will have to be done. I haven't tried that, so we'll have to try it, but I suspect it's a valid approach.

rickmanelius:
So after thinking further, I think the proper path forward is the following for TYPO3:

## Mid/Long term approach:

* Optional: TYPO3 adds AdditionalConfiguration.php to .gitignore (not mandatory for this to work).
1. If no AdditionalConfiguration.php exists, add from template.
2. If AdditionalConfiguration.php exists, post a warning and point to documentation on how to modify DB credentials.
3. After `ddev config`, always state DB credentials and note `ddev describe` for more information.
4. If import-db runs, generate LocalConfiguration.php and AdditionalConfiguration.php if it doesn't exist.

## Short term approach:

* Remove the check for LocalConfiguration.php before attempting to write AdditionalConfiguration.php.
* Link to documentation if AdditionalConfiguration.php can't be written.

## Discussion

Here's why this could work:

* AdditionalConfiguration.php is not touched by TYPO3 and *should* not be checked in https://docs.typo3.org/typo3cms/InsideTypo3Reference/CoreArchitecture/Configuration/AdditionalConfiguration/Index.html.
* It always loads after LocalConfiguration.php, so it ensures the DDEV will always pick up the proper credentials.

rickmanelius:
One other use case to not ignore... one of the advantages of "DDEV configuration in code" is that on a multi-team project, one can just hit git clone and go. In moving the DB settings generation to the configuration step, it's very much possible for someone to clone and hit a wall if the settings.php file is stored in code but the settings.local.php is not (and it shouldn't be due to .gitignore). This is mitigated by the fact that someone working this way will have to import a DB, but if we remove the responsibility of generating that DB settings file from import-db, it will bork.

rickmanelius:
@rfay I'm also happy to proceed with that approach. It was still helpful for me to step through this so I understood the entire scope of what would be impacted and how TYPO3 approaches the setup process.

rickmanelius:
One final thought about the other use case I surfaced in https://github.com/drud/ddev/pull/693#issuecomment-370893215. We could probably get around that by running the DB settings generation after `ddev config` and again for `ddev start`. That way, it's there no matter what.

rfay:
I propose this approach, which solves most short and long-term:

1. It's not an error to not have LocalConfiguration.php (but we should util.Warning() it. 
2. If AdditionalConfiguration.php does not exist, create the typo3conf directory if it doesn't exist and add our AdditionalConfiguration.php to it.
3. If typo3conf/LocalConfiguration.php didn't already exist, util.Warning() them that they have to do an install manually.

This seems to work fine on a manual runthrough with TYPO3 8.7.10, and I think it solves all of the short- and long-term issues above. 

@mglaman I know that this isn't what you signed up for, so if you have it where you want it and want me to push a commit doing the TYPO3 stuff, I will.

The only discussion missing now is that we have discussed moving this whole step of this PR to "start" instead of "config". That's a valid idea. I'm willing to either do it or wait and see how we do with config.

mglaman:
I'll review. But I think on start should be a later discussion. You're
beginning to merge into my points that in start we should write the ports
somewhere. Generating a local.php on start if missing is smart. But I think
it's out of scope here.

I'm willing to push this forward. Will read on computer later and grok next
steps.

rickmanelius:
I think we're good to go with respect to updating the acceptance criteria. Working on that now for the remaining tasks!

rickmanelius:
Issue summary updated on this PR to reflect remaining tasks. The last two I've assigned to me to address documentation and filing of a follow-up issue on the ddev config vs ddev start item. @mglaman I know you've been super generous with your time on this one, so let us know if this is too much to put on your plate.

mglaman:
@rickmanelius thanks for the clarification! Scope looks fine to tackle :) The only grey area will be TYPO3. But the tests _should_ mock the environment enough. And if not then I will improve those tests!

mglaman:
>  We need to append an inclusion of wp-config-local.php within the wp-config.php template similar to how D8 does it.

WordPress uses constants. So the logic for loading needs to be reviewed. There's no way to reassign constants. So the local would have to be loaded first and then `if (!defined)`

rickmanelius:
I talked with @mglaman at Midcamp. I think we need to split this out so that we can get the Backdrop/Drupal changes in place while we reach out to some of our WordPress colleagues to address that use case and have @rfay or @cweagans (with support from the TYPO3 community) to finish the work on TYPO3.

rickmanelius:
To clarify this one (my last comment wasn't clear because I was moving fast).

Ultimately, this is becoming a big PR and therefore harder to test. To that end, I would prefer that we get it merged while simplifying the path forward for the open items related to WordPress and TYPO3. To that end, I say we call @mglaman's work done and I'll create a revised punch list with items associated to myself and @rfay to address from here on out.

Off the top of my head, these would be:

* [ ] WordPress: Postpone the mirroring of the version control + local DB settings file for now. Split into a new ticket because this is not a blocker (@rickmanelius)
* [ ] TYPO3: Generate AdditionalConfiguration.php and parent folder (if they don't exist) and warn if  LocalConfiguration.php doesn't exist (@rfay)
* [ ] Docs: Update to specify the new timing of when the DB settings is generated (@rickmanelius)

cmd/ddev/cmd/config.go
```diff
@@ -163,6 +163,11 @@ var ConfigCommand *cobra.Command = &cobra.Command{
 			util.Failed("Could not write ddev config file: %v", err)
 		}
 
+		_, err = app.CreateSettingsFile()
+		if err != nil {
+			util.Failed("Could not write settings file: %w", err)
```
rfay: This should be %v, not %w

pkg/ddevapp/apptypes.go
```diff
@@ -94,9 +94,11 @@ func IsValidAppType(apptype string) bool {
 func (app *DdevApp) CreateSettingsFile() (string, error) {
 	app.SetApptypeSettingsPaths()
 
-	// If neither settings file options are set, then don't continue
+	// If neither settings file options are set, then don't continue. Return
+	// a nil error because this should not halt execution if the apptype
+	// does not have a settings definition.
 	if app.SiteLocalSettingsPath == "" && app.SiteSettingsPath == "" {
```
rfay: I think we probably need a util.Warning() here.

rfay:
I rebased and made the minor changes I suggested, and snuck in a timeout increase for the circleci archive creation; we'll probably have to solve that problem elsewhere, as it's taking far longer than we want it to. After @cweagans has taken a gander and @rickmanelius approves we should be ready to go. @mglaman please mention if you have any reservations before it goes in.

mglaman:
👍 here

rickmanelius:
## Preparation

* `git checkout mglaman/gh-692-always-create-settings-file`
* `make darwin` 
* `ddev version` v0.15.1-15-g922d16ba
* Fresh CMS Including w/default DB settings templates/files
  * Drupal 7.57
  * Drupal 6.38
  * Drupal 8.5.0
  * Backdrop 1.9.2
  * WordPress 4.9.4
  * TYPO3 8.7.10
* Settings command to test `ddev config --projectname settings --docroot . --projecttype INSERT_TYPE`
* Documentation to Review
  * [Getting Started](https://ddev.readthedocs.io/en/latest/users/cli-usage/#getting-started). We should note the generation of these files alongside .ddev/config.yml
    * Example Addition: Also, if a Database configuration file doesn't already exist, it will be generated from a template.
  * [Import Assets](https://ddev.readthedocs.io/en/latest/users/cli-usage/#importing-assets-for-an-existing-project). This will no longer need "Running this command will provide a prompt for you to specify the location of your database import"

rfay:
Followup issues for wordpress, TYPO3, and docs created.

rickmanelius:
I did hit the approve button, but there are two documentation items to hit:

* Getting Started. We should note the generation of these files alongside .ddev/config.yml. Example Addition: Also, if a Database configuration file doesn't already exist, it will be generated from a template.
* Import Assets. This will no longer need "Running this command will provide a prompt for you to specify the location of your database import". This should be deleted.

And then once @cweagans reviews, ready to merge!

rfay:
This can't go in until the item @rickmanelius discovered is resolved,
> Error vs Warning. I realize that we're trying to take over the backup settings.local.php, but I'm not sure that warrants an error message versus a warning with a link to more information. I can see how an error would make sense because a ddev start would then fail, but someone could override our #ddev-generated label in order to lock in some custom changes they wish to keep and that's fully valid.

rickmanelius:
The recommendation that I put re:error vs warning turns out to be an error that we need to correct to ensure we don't exit early. Here are two attachments showing what occurs when using either the flags or interactive prompts.

Fastest way to recreate:

1. Download/unzip https://www.drupal.org/project/drupal/releases?api_version%5B%5D=103
2. Create blank settings.php and settings.local.php in sites/default/
3. Run `ddev config --projectname settings --docroot . --projecttype drupal7`

screen shot 2018-03-16 at 1 25 25 pm
screen shot 2018-03-16 at 1 43 14 pm

rfay:
Added https://github.com/drud/ddev/pull/693/commits/c2a08ad573959c5f4e11355568e1dc4d3e6ed8f3 to warn on the settings-are-managed failure, but not prevent the operation from completing. IMO when this passes tests it can be pulled.

rickmanelius:
Nice. Will re-review a final time before the v0.16.0 release, but very excited to see this land!

PR #695, opened 2018-02-28
Don't remove settings.local.php when removing database, fixes #478 PULL 2018-03-06 #TRIVIALREVIEW

### LINKED ISSUE #478: Failure to remove settings.php on ddev rm --remove-data (Don't remove settings file!)

I did a `ddev rm --remove-data` and got the error message

`Failed to remove gchistory: remove /Users/rfay/workspace/gchistory/docroot/sites/default/settings.php: permission denied
`

The settings.php in this case is in fact ddev-generated, so removing it would be legit in almost any context.

I didn't think it was such a hard task. I didn't expect an error. 

The problem here is just the classic problem that Drupal is *always* trying to set sites/default to unwriteable, even by the owner of the directory. So any user that wants to change sites/default/settings.php has to make the sites/default directory writeable first. We need to do that.

rfay:
## The Problem/Issue/Bug:

OP #478 

We used to try to delete the settings.php or settings.local.php when deleting database (`ddev remove --remove-data`) but this isn't necessary and can just create confusion. 

It also doesn't make sense to delete the settings file just because we're dropping the database.

* Create a Drupal site and use it
* check that Drupal has done its silly thing and set sites/default to 0444.
* `ddev rm --remove-data` and look for no error.

## Related Issue Link(s):

OP #478

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -788,7 +788,6 @@ func (app *DdevApp) DetermineSettingsPathLocation() (string, error) {
 // Down stops the docker containers for the project in current directory.
 func (app *DdevApp) Down(removeData bool) error {
 	app.DockerEnv()
-	settingsFilePath := app.SiteSettingsPath
 
 	// Remove all the containers and volumes for app.
 	err := Cleanup(app)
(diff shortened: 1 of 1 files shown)


rickmanelius:
@rfay Can you tell me the mechanism you're using to create the database file? I've attempted a drush and install.php for D7/D8 and sites/default isn't set to 0444 while settings.php is. Using `ddev rm --remove-data` doesn't remove settings.php. I'm trying to find a way to replicate generating a settings.local.php (I'm assuming Backdrop?) because it's been a while since I've seen that come up in my testing.

rfay:
So @rickmanelius the long way is:

1. Install Drupal 7 (or 8, or 6). It will create a settings.php
2. Import a database
3. Visit the site, log in, I'm not sure when Drupal makes the changes.
4. ls -ld sites/default should show it unwriteable.
5. You should see a sites/default/settings.local.php
6. Try to `ddev rm --remove-data` the site.

You could also just import a db and visit the site and you'll see the above with settings.php created by ddev. To short-circuit (without visiting the site) you could just set sites/default to 0444.

The good news is none of this will happen with this patch :) But to demo the failure you need to do something like this procedure.

rickmanelius:
I was finally able to recreate. It was a slightly different path:

* New D7 codebase with no settings.php.
* ddev config and start.
* Import DB and confirmed settings.php generated by ddev
* Logout, login, and visit an admin page (this seemed to trigger the sites/default perm change)
* `ddev rm --remove-data` throws the error below:

```
Failed to remove drupal-7.56: remove /Users/rmanelius/git/drupal-7.56/sites/default/settings.php: permission denied
```

PR #707, opened 2018-03-02
Allow override of mysql and php configuration, fixes #654

### LINKED ISSUE #654: "Index column size too large" MySQL error when importing TYPO3 DB
When importing db dump of the TYPO3 9 I've got mysql error:

"Index column size too large. The maximum column size is 767 bytes."
An example failing query is:
```
CREATE TABLE sys_file_processedfile (
uid int(11) NOT NULL auto_increment,
tstamp int(11) NOT NULL default '0',
crdate int(11) NOT NULL default '0',
storage int(11) NOT NULL default '0',
original int(11) NOT NULL default '0',
identifier varchar(512) NOT NULL default '',
name tinytext,
configuration text,
configurationsha1 char(40) NOT NULL default '',
originalfilesha1 char(40) NOT NULL default '',
task_type varchar(200) NOT NULL default '',
checksum char(10) NOT NULL default '',
width int(11) default '0',
height int(11) default '0',
PRIMARY KEY (uid),
KEY combined_1 (original,task_type,configurationsha1),
KEY identifier (storage,identifier(199))
);
```
Error:
Index column size too large. The maximum column size is 767 bytes.

The same error was reported by user trying to install fresh TYPO3 v7.6. However he reported that installation of the fresh 8.7 or 9 went through (probably because starting from 8.7 TYPO3 adds default charset when creating tables).

## How to reproduce this:
log in to ddev mysql, and try to run query from above.

rfay:
## The Problem/Issue/Bug:

Users need to be able to override php and mysql configuration in the same way as they can override nginx configuration.

* Update docker-compose.yml to mount .ddev as /ddev_config in  both web and db containers. This allows those containers to easily change their behavior based on the contents of /ddev_config.
* Remove an obsolete section that created an nginx-site.conf for nginx. After https://github.com/drud/ddev/pull/669 went in this is totally obscene :)

The web and db images have been updated in this PR. 

* Delete .ddev and reconfigure
* Add a php.ini and my.cnf to your heart's delight
* ddev start
* Verify that the site admin/reports/status/php shows the changed value
* Verify that the mysql behavior changes. A nice test is to fix the utf8mb4 problem.

Tests are added in both containers for the new behavior. Since the behavior is essentially all in the containers that's probably good enough.

## Related Issue Link(s):

Mariadb container change: https://github.com/drud/mariadb-local/pull/46
Nginx container change: https://github.com/drud/docker.nginx-php-fpm-local/pull/54
OP: #654 - need to override my.cnf

OP: #654 problems with importing mysql databases.


rfay:
Remaining todo:
- [ ] Make the PHP and mysql settings *overrides* (add-ons) instead of full replacement files.
- [ ] Notify on start about the existence of any of the 3 (nginx, mysql, php) overrides that are being loaded. (Problems with these have confused many people, including me)
- [ ] Test D6 extensively, since this removes a conflicting nginx config that was accidentally left in.
- [ ] Update to bring in a real release of the nginx and mysql containers.

rfay:
This issue will be used to bring in the new db and web image tags for v0.15.0.

pkg/ddevapp/config.go
```diff
@@ -309,6 +309,30 @@ func (app *DdevApp) WriteDockerComposeConfig() error {
...
+	if _, err := os.Stat(filepath.Join(ddevDir, "mysql")); err == nil {
+		customConfig = true
+	}
+
+	if _, err := os.Stat(filepath.Join(ddevDir, "php")); err == nil {
+		customConfig = true
+	}
+
+	if customConfig {
+		util.Warning("The current project utilizes custom configuration.")
```
rfay: I'd like this to be much stronger: This project is using custom configuration for PHP (list files). This project is using custom configuration for mariadb (list files)

docs/users/extend/customization-extendibility.md
```diff
@@ -22,7 +22,21 @@ The default web container for ddev uses NGINX as the web server. A default confi
 - **NOTE:** The "root" statement in the server block must be `root $NGINX_DOCROOT;` in order to ensure the path for NGINX to serve the project from is correct.
 - Save your configuration file and run `dev start` to start the project environment. If you encounter issues with your configuration or the project fails to start, use `ddev logs` to inspect the logs for possible NGINX configuration errors (or use `ddev ssh` and inspect /var/log/nginx/error.log.)
 - Any errors in your configuration may cause the web container to fail and try to restart, so if you see that behavior, check your container.
-- Changes to .ddev/nginx-site.conf take place only after you do a `ddev rm` followed by `ddev start`.
+- **IMPORTANT**: Changes to .ddev/nginx-site.conf take place only after you do a `ddev rm` followed by `ddev start`.
```
rfay: `ddev restart` is the ticket here.

docs/users/extend/customization-extendibility.md
```diff
@@ -22,7 +22,21 @@ The default web container for ddev uses NGINX as the web server. A default confi
...
+
+## Providing custom PHP configuration (php.ini)
+
+You can provide additional PHP configuration for a project by creating a directory called `.ddev/php/` and adding any number of php configuration ini files. Normally, you should just override the specific option that you need to override. Note that any file that exists in `.ddev/php/` will be copied into `/etc/php/[version]/(cli|fpm)/conf.d`, so it's possible to replace files that already exist inside the container, so if in doubt, put your custom overrides in a file called `my-php.ini`.
+
+One interesting implication of this behavior is that it's possible to disable extensions by replacing the configuration file that loads them. For instance, if you were to create an empty file at `.ddev/php/20-xdebug.ini`, it would replace the configuration that loads xdebug, which would cause xdebug to not be loaded!
+
+To load the new configuration, just run a `ddev stop` and `ddev start`.
```
rfay: Conflicts with earlier statement to use rm. Of course if stop and start works, then we can use `ddev restart`.

docs/users/extend/customization-extendibility.md
```diff
@@ -22,7 +22,21 @@ The default web container for ddev uses NGINX as the web server. A default confi
...
+
+## Providing custom mysql/MariaDB configuration (my.cnf)
+
+You can provide additional PHP configuration for a project by creating a directory called `.ddev/mysql/` and adding any number of MySQL configuration files. These files will be automatically included when MySQL is started.
+
+To load the new configuration, just run a `ddev stop` and `ddev start`.
```
rfay: `ddev restart` right? As would be expected, `ddev restart` does work, just tested it.

rfay:
In the docs, lets make sure to give a trivial example. For example, the mysql configs have to be .cnf? PHP definitely has to be .ini. And the section header has to be included in the file. So I think we should mention gotchas like that and show a simple example.

rfay:
I *think* we just have the problem with TestDevLogs to resolve. Might be timing or how much of the log we're grabbing.

rfay:
@cweagans:

* Updated docs per comments
* Added more detail to the config override warnings
* Added a couple of tests

Please take a look. We may be closing in, thanks!

pkg/ddevapp/config.go
```diff
@@ -309,6 +309,43 @@ func (app *DdevApp) WriteDockerComposeConfig() error {
...
+	if _, err := os.Stat(phpPath); err == nil {
+		phpFiles, err := fileutil.ListFilesInDir(phpPath)
+		util.CheckErr(err)
+		if len(phpFiles) > 0 {
+			util.Warning("Using custom PHP configuration: %v", phpFiles)
+			customConfig = true
+		}
+	}
+	if customConfig {
+		util.Warning("Custom configuration takes effect after 'ddev restart'.")
```
cweagans: I don't think this is a valid message. These checks happen when the container is started, so it may be confusing to have a message telling you to restart while you're in the process of starting the container.

rfay: No, unless I'm lost, these checks happen when `ddev start` is executed. That may or may not result in a container start. Until [ddev start should use docker-compose up --force-recreate](https://github.com/drud/ddev/issues/716)

Tell me if I'm just lost.

rickmanelius:
I believe this also resolved https://github.com/drud/ddev/issues/660.

PR #708, opened 2018-03-03
Switch restart to 'no' for main containers, fixes #676

### LINKED ISSUE #676: Remove restart configs from ddev's docker-compose actions

For whatever reason we've always had "restart" as the action on failure in our ddev containers. But this actually never works out. There are in fact no cases where a restart fixes anything. 

All our containers should be set to not restart on failure.

rfay:
## The Problem/Issue/Bug:

OP #676: Our web and db containers never do recover when a restart happens, so restarting them just obscures the cause. This PR changes the restart for the "working" containers from "always" to "no".

@mglaman also points out the unexpected side-effect of the current restart: always policy: A machine reboot or docker restart results in stopped projects magically starting unexpectedly. 

This change should result in more predictable ddev container behavior.

Failing containers won't restart any more.

ddev ssh into the web container. `killall supervisord`. This will ruin the container, and it will show up as unhealthy in a bit, but won't restart.

ddev ssh -s db into the db container. `killall mysqld`. This will ruin the container, and it will stop, and it will show up as stopped in `ddev list`

## Related Issue Link(s):

OP #676

### DIFF
--- a/pkg/ddevapp/templates.go
+++ b/pkg/ddevapp/templates.go
@@ -11,7 +11,7 @@ services:
     volumes:
       - "${DDEV_IMPORTDIR}:/db"
       - "${DDEV_DATADIR}:/var/lib/mysql"
-    restart: always
+    restart: "no"
     ports:
       - "3306"
     labels:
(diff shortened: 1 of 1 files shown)


rfay:
I went ahead and added the "no" for the router as well.

mglaman:
That's weird. The router had the setting but didn't seem to run.

PR #718, opened 2018-03-08
Disable xdebug by default for 6x speed improvement fixes #664

### LINKED ISSUE #664: TYPO3 environment extremely slow
Working on TYPO3 in the DDEV environment is extrem slow.  Saving Content needs 1-3 Seconds. Also FE needs lots of time. 

Any ideas what i could do to speed up? 

Working on:
macOS 10.12
Docker Version 17.12.0-ce-mac49 (21995)

bildschirmfoto 2018-02-20 um 09 59 15

rfay:
## The Problem/Issue/Bug:

I think this will fix #664 "TYPO3 environment extremely slow"

From https://github.com/drud/docker.nginx-php-fpm-local/pull/57 description of the container change:

People complain that ddev is slow. It was.

If we turn off xdebug by default it's 6x faster. 

I was always wrong about its impact

d8 web-based install:

macOS 10.13.3 8GB memory MacBook Pro (Retina, 13-inch, Early 2015
Processor: 2.7 GHz Intel Core i5
Docker 17.09.0-ce-mac35 (19611) default 2CPU, 2GB memory

Test: Install d8 8.5 HEAD via gui/web. "Install site"
with xdebug on: 12:34 
with xdebug off (phpdismod xdebug +killall -1 php-fpm): 2:08

6x faster

Delete your .config.yaml (or delete the web image line) and delete your docker-compose.yaml
ddev start.
Try things out.

## Related Issue Link(s):

nginx container PR: https://github.com/drud/docker.nginx-php-fpm-local/pull/57
#664: Discussion of slow TYPO3 environment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ VERSION_VARIABLES = DdevVersion WebImg WebTag DBImg DBTag RouterImage RouterTag 
 # These variables will be used as the default unless overridden by the make
 DdevVersion ?= $(VERSION)
 WebImg ?= drud/nginx-php-fpm-local
-WebTag ?= v1.1.0
+WebTag ?= 20180308_disable_xdebug_by_default
 DBImg ?= drud/mariadb-local
 DBTag ?= v0.8.0
 RouterImage ?= drud/ddev-router
--- a/cmd/ddev/cmd/logs_test.go
+++ b/cmd/ddev/cmd/logs_test.go
@@ -58,7 +58,7 @@ func TestDevLogs(t *testing.T) {
 
 		assert.NoError(err)
 		assert.Contains(string(out), "Server started")
-		assert.Contains(string(out), "PHP message: PHP Stack trace:")
+		assert.Contains(string(out), "PHP Fatal error:")
 
 		cleanup()
 	}
--- a/docs/users/step-debugging.md
+++ b/docs/users/step-debugging.md
@@ -1,12 +1,13 @@
 <h1>Step-debugging with ddev and xdebug</h1>
 
-Every ddev project is automatically configured with xdebug so that popular IDEs can do step-debugging of PHP code.
+Every ddev project is automatically configured with xdebug so that popular IDEs can do step-debugging of PHP code. It is disabled by default for performance reasons, so you'll need to enable it in your config.yaml.
 
 xdebug is a server-side tool: It is installed automatically on the container and you do *not* need to install it on your workstation. All you have to do on your workstation is to add an extra IP address (below) and perhaps add a browser extension or bookmark.
 
 All IDEs basically work the same: They listen on a port and react when they're contacted there. So IDEs other than those listed here should work fine, if listening on port 11011.
 
 **Key facts:**
+* You need to explicitly enable xdebug in your config.yaml as a post-start step.
 * The debug server port on the IDE must be set to port 11011. Although the xdebug default is port 9000, that port often has conflicts for PHP developers, so 11011 is used with ddev.
 * An IP-address *alias* of 172.28.99.99 must be added to your workstation host's loopback address. On macOS this is done with the command `sudo ifconfig lo0 alias 172.28.99.99`. On Ubuntu 16.04 and probably other Linux variants, `sudo ifconfig docker0:0 172.28.99.99 up` **This must currently be done after each reboot.**
 
@@ -15,6 +16,17 @@ For more background on XDebug see [XDebug documentation](https://xdebug.org/docs
 For each IDE the link to their documentation is provided, and the skeleton steps required are listed here.
 
 ## Setup Instructions
+
+### Enable or disable xdebug in your config.yaml
+
+Use a post-start hook to enable or disable xdebug on startup:
+
+```
+hooks:
+    post-start:
+      - exec: enable_xdebug
+```
+
 
 * [PHPStorm](#phpstorm)
 * [NetBeans](#netbeans)
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -20,7 +20,7 @@ var DockerComposeVersionConstraint = ">= 1.10.0-alpha1"
 var WebImg = "drud/nginx-php-fpm-local" // Note that this is overridden by make
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "v1.1.0" // Note that this is overridden by make
+var WebTag = "20180308_disable_xdebug_by_default" // Note that this is overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/mariadb-local" // Note that this is overridden by make


cweagans:
I mentioned this quickly in Slack last night, but aren't we just moving the point at which ddev is slow here? Backend devs in particular will need a debugger, so when they go to turn it on, everything will get slower, and that's a really frustrating experience.

I think that there's a lot of value in having xdebug on by default. It's one less thing for a large swath of developers to have to worry about, and having it working out of the box with zero configuration is a selling point of the product, and I think that there was agreement on that point whenever xdebug was added to the default configuration.

We've lived with the slowness for a while, and while I can recognize that it's not ideal, I'm not so sure just killing xdebug is the answer here. There's something else at play because xdebug is not nearly this slow outside of Docker. I'd like to propose that we work around the filesystem issues first, because that's always been a major point of slowdown, and it's the only thing that I can think of that would exacerbate the xdebug slowness. I think we should look at adopting [bg-sync](https://github.com/cweagans/docker-bg-sync) as a standard part of ddev. We used it at NBC for a while, and it was something like a 30x speedup across the board (this was in the earlier days of Docker for Mac, so the speedup now might not be quite as substantial).

rickmanelius:
I'm wanting to roll back to an earlier version of ddev to see if I can determine when this was introduced because it wasn't this severe before. I'm curious if there is some other factor that caused this and I'd prefer to know that before going into the file system, which I know is where much enthusiasm has gone to die in the past for us and other teams.

rfay:
git bisect should be able to tell us what happened (if something did). It won't help us with the fact that docker has changed over that time. I'll give that a shot this weekend, it's interesting.

revolutionate:
I'm convinced that "xdebug is slow" is not the entire story.

I've been testing today on MacOS, re-installing Drupal 8 via drush (`ddev ssh` then `time drush si`).

To begin with, install was ~50s with xdebug enabled and ~30s without - certainly not the 6x difference. This was an installation from a week or so ago (nginx-php-fpm-local:v0.9.3 I _think_), so I tried trashing my containers and starting again with ddev config.

After this, I'm seeing 3m20s(!) with xdebug, ~35s without.

So obviously I tried changing my webimage back to v0.9.3, but, this doesn't make any difference! Even another dev site I created the same day as the first one is similarly slow with xdebug enabled... this is too weird.

I'll try and maybe `strace` it or something to see if I can see what's going on.

rickmanelius:
Thank you @revolutionate for confirming that I'm not crazy that this was not an issue from the past, but a recent issue. If we can trace this down, we can cut a minor release because we obviously want xdebug available out of the gates, but not with the massive performance penalty that can creep up.

Question: which version of OS and Docker version are you running? I see a few dimensions that can be causing this: Docker version, container images, ddev version. Hoping we don't need to step through all 3, but I'll also try bisecting (which will change the default images as well).

rickmanelius:
Going to move this testing to a Google Sheet to try and track this down.

rickmanelius:
So I'm also able to see a situation where there is no issue with v0.15.0 with a stock D7 site. Sharing my data. Going to see if I can recreate the D8 issue next. 
https://docs.google.com/spreadsheets/d/1QMo6Oq2lBYGex0wF2TvGzCOgcFdEH0lF-7XTPZNFP_g/edit#gid=0

rickmanelius:
I'm starting to catch the inconsistent nature of this bug. With a fresh D7 site, there's not issue all the way through v0.11.0, v0.14.0, and v0.15.0. D6 also is performant on v0.15.0. D8 was slow again (3+ minutes) and then when I disabled xdebug it went down to ~ 35s. When I enabled it again, it went up to ~ 1 minute for 3 drush si's in a row.

I'm thinking this is might be a memory limit issue or something that crosses a major threshold and xdebug is simply the cause of something else.

rickmanelius:
Trying a few other dimensions... including docker memory, php memory, and other potential solutions. It's not consistent and sometimes D8 has trouble on `ddev start` and results in an unhealthy container due to a time out.

rfay:
Thanks for this, both of you! Don't forget that while `time drush si` is ever so much easy to test, it may or may not be representative of php-fpm behavior. So far it seems representative to me, but I don't trust it. 

IMO process should be something like this (interested in your additions):
* rm -r .ddev ~/.ddev/projname
* Get new ddev (perhaps `make clean darwin`)
* ddev config
* ddev start
* `ddev exec time drush si` (Works fine after settings.local.php has been created, or settings.php has db/db/db/db)

rickmanelius:
Alright. I *think* I've been able to consistently get to a successful `drush si`. In Docker for Mac, I updated the memory from 2GB to 3GB and that has made work every time. Going to try the UI installs to confirm.

rickmanelius:
As per @rfay's comment, now I'm going to test through the UI to confirm this is repeatable.

rickmanelius:
With the increase of the memory limit for Docker for Mac from 2 GB to 3 GB, the UI install time was ~ 2:30s before and after Xdebug.

My vote is that this is NOT an xdebug issue, but a resource/Docker level issue. I suggest we focus on, when upgrading and in the release notes, highlighting this as a solution.

rickmanelius:
It is worth noting that on v0.14.0 I had the web install time down to 1:30 w/out xdebug. So there is still a performance delta worth noting even though the average experience (and notably w/xdebug scenario) was better.

revolutionate:
From running `strace` against drush, the difference I'm seeing between fast(ish) and slow is many thousands of calls to `gettimeofday()`.

I wondered whether https://blog.packagecloud.io/eng/2017/03/08/system-calls-are-much-slower-on-ec2/ was something to do with this, but the number of `gettimeofday` calls from the test program in that post never seems to vary, so probably not(?).

My theory is the difference is caused by xdebug ignoring the request versus actively tracing all the calls... once it starts being slow though, I have to restart docker before it's quick again - `ddev remove`/`ddev start` somehow isn't enough!

I'll try bumping up the memory limit now though, and see if I can still reproduce it.

rickmanelius:
@revolutionate Yes, Docker sometimes needs to be reset because it doesn't always release resources. Hopefully, that is something they solve upstream because, short of forcing a docker reset every so often, it does very much affect the end-user experience.

Thanks for tracing that performance issue down... curious to see if the memory limit change addresses this at a symptom level, but the number of calls is obviously something that we hope can be addressed.

revolutionate:
I'd actually given up trying to reproduce it for the moment, but just had this show up in my screen session into the linuxkit VM:

```
[ 2109.731574] clocksource: timekeeping watchdog on CPU0: Marking clocksource 'tsc' as unstable because the skew is too large:
[ 2109.733504] clocksource:                       'hpet' wd_now: eac938e1 wd_last: e9908d0c mask: ffffffff
[ 2109.734845] clocksource:                       'tsc' cs_now: 13612bd25e8d cs_last: 13601178370b mask: ffffffffffffffff
[ 2109.741802] clocksource: Switched to clocksource hpet
```

... and now seems to be borked again.

```
Congratulations, you installed Drupal!                                                                                                                                                                                                    

real	3m25.527s
user	0m44.580s
sys	2m13.390s
```

This is Docker 17.12.0-ce-mac55 on MacOS High Sierra 10.13.3 by the way, with memory set back to 2GB.

Showing 875MB of usable free memory with 5 ddev sites running, but has been eating into swap:
```
linuxkit-025000000001:/var/log# free -m
             total       used       free     shared    buffers     cached
Mem:          1998       1795        203        232        100        571
-/+ buffers/cache:       1123        875
Swap:         1023        777        246
```

revolutionate:
Back to the actual PR for a moment...

Even if it's not a 6x improvement, I do think almost doubling the speed by running with the extension disabled is still an attractive proposition, and something I'll personally be doing until I need to debug anything.

Maybe a command like `ddev debug` could make it a bit friendlier, and even spit out some help about IDE configuration...

rickmanelius:
@revolutionate I concur that we may default that until off because while xdebug is common, it's not used universally and getting the speed gain is compelling. Thanks for all your help tracing this!

revolutionate:
And yes, deliberately switching the clocksource to `hpet` in the linuxkit VM also elicits the same behaviour, so has to be the culprit. (I saw [this thread](https://news.ycombinator.com/item?id=13814771) earlier, but now it makes sense!) When the kernel decides tsc is unstable, it can be fixed only by quitting docker and re-opening (or adjusting the memory, cause that would restart the VM too 😄 )

I suppose ddev could conceivably warn the user that xdebug is going to be slow, and that they should restart docker...
e.g.
`php -i | grep xdebug`
`cat /sys/devices/system/clocksource/clocksource0/current_clocksource | grep hpet` (may use a different fallback clocksource on other platforms, mind)

Still don't know what particularly causes the two timers to drift, but maybe irrelevant to this discussion.

revolutionate:
Tried leaving a script running to poll for the clocksource change and display a notification.

One thing which reliably screws it up here is putting the Macbook to sleep - the clocksource failover happens right away on resume, and xdebug performance is then bad until the docker VM gets restarted.

rfay:
Interestingly enough, sleep is also implicated on the docker-compose hang in 17.12.

rfay:
Just a note that back in ddev v0.6 the database name was "data" and you could install with user=root pass=root (host=db)

rfay:
Please remember in testing that we use opcache, so it's not a fair test if the container is not created from scratch before the test.

rfay:
I haven't been able to get any of the results reported here using Docker 18.03 and a pretty fast 2015 MBP with the default 2GB memory settings and 2 CPU. I haven't tried increasing docker memory, but basically with any version of ddev (including way back to v0.6) I get slow with xdebug (3.5-4 minute drush si) and maybe 6x faster without (0:33-0:41). 

`time drush si` may not be representative of web install times, but it does seem to track the same.

revolutionate:
@rfay did you check what the VM is using as clocksource? Would be interested to hear if this behaves differently on a newer Docker.
`cat /sys/devices/system/clocksource/clocksource0/current_clocksource` from within any container.

Just to reiterate, for me (on 17.12-ce-mac55) while it says `tsc` the xdebug overhead is low (~1min to install D8), if it says `hpet` it's back to 3.5 minutes again, because of _lots_ of `gettimeofday` syscalls.

It's `tsc` here when docker first starts up, but ends up on `hpet` later after the machine has resumed from sleep. (my [script to detect when it goes wrong](https://gist.github.com/revolutionate/7ba1eea3c9e9d0fd4d1be4c696c477c2))

(Maybe something to do with the time synchronization fix added in 17.06 https://github.com/docker/for-mac/issues/17#issuecomment-300734810 or just generally the time drift issue they were "fixing"...)

rfay:
It seems to be hpet always on this machine with Docker 18.03. I tried a clean `ddev rm && ddev start`, got hpet. Did a manual sleep, checked on return and it was hpet.

revolutionate:
>  I tried a clean ddev rm && ddev start, got hpet

Removing containers won't do anything to change it, since that's coming from the underlying shared linux kernel.

Quitting Docker and re-launching is the only way to go back to the preferred `tsc` (if it got blacklisted, like in first part of [this comment](https://github.com/drud/ddev/pull/718#issuecomment-372074750).)

rfay:
Clean start of docker shows tsc as timesource, as you suggested. Docker 18.03.0-ce-rc1-mac54 (23022)

cweagans:
I did some digging into this today, and on MAMP, I was able to see around a 4x slowdown with xdebug enabled. It makes sense that there is some additional decrease on osxfs due to the known performance issues with that component. Given @rickmanelius has given this the thumbs up to move forward, one last thing that I think is important is just a note in the documentation stating that enabling xdebug substantially decreases performance. It's a small thing, but I think it's important to note that behavior + why it happens so that we're setting expectations properly.

rickmanelius:
I'm good with this going in as is. The documentation does note the performance hit. Also, the instructions on adding it back in as a hook. We do have the capability of using php.ini overrides, but the phpenmod is probably more elegant. At the moment. Let's pull and we can tweak based on feedback. Midcamp did highlight for me how exceptionally important and how valued having xdebug support readily available was to those evaluating projects... but the performance hit of 4X-6X on certain operations is just a lot to stomach for some use cases.

Approved

rfay:
This is approved so can come in once the upstream nginx-php-fpm-local release is done. We probably need to see if we want to get anything else into that one.

rickmanelius:
Is there a specific nginx-php-fpm-local PR/issue to track? No rush on my side... just trying to understand the path to completion.

rfay:
Already pulled in https://github.com/drud/docker.nginx-php-fpm-local/pull/57

rfay:
Pulling this even though it has temporary container reference for the web container. That way we can make sure that https://github.com/drud/ddev/pull/745 is properly up to date with this one and it can be reviewed properly.

PR #738, opened 2018-03-15
docker-compose version check failure, fixes #711, fixes #737

### LINKED ISSUE #711: ddev hostname should not check for docker (Windows)

`ddev hostname` is the way we recommend for people to edit their hosts file on windows. However, if you try to run it in a window where docker is not working (with docker toolbox, missing environment variables), you can't run it. And there's no reason for it to check for docker.

### LINKED ISSUE #737: docker-compose version check is incorrect, results in failure on latest docker Edge

Started up Docker for Windows on latest docker-ce edge (18.03)

`
The docker-compose version currently installed does not meet ddev's requirements: 1.20.0-rc2 is less than 1.10.0
`

Obviously the error message is incorrect, and this is a major blocker.

`ddev version` does not report because of the docker-compose problem. This should also be fixed.

```
C:\Users\rfay\workspace\randyfay.com>docker version
Client:
 Version:       18.03.0-ce-rc3
 API version:   1.37
 Go version:    go1.9.4
 Git commit:    e730959
 Built: Thu Mar  8 20:11:53 2018
 OS/Arch:       windows/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.03.0-ce-rc3
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   e730959
  Built:        Thu Mar  8 20:20:22 2018
  OS/Arch:      linux/amd64
  Experimental: true
```

rfay:
## The Problem/Issue/Bug:

#737: Docker-compose latest edge is 1.20.0-rc2. But [semver does not support versions with a pre unless the check has a pre](https://github.com/Masterminds/semver/issues/21). So our docker-compose version check was broken.

* Update the constraint to include a "pre" for both docker and docker-compose.
* Fix the logic about skipping pre-validation on commands like "version" and "hostname" to fix #711

With latest edge docker-compose, try `ddev version`, `ddev hostname` with and without this. 

For brute-force testing, create a script called docker-compose in a location which is *not* normally in your $PATH (like ~/tmp) with this content:
```
#!/bin/bash
echo "0.5 "
```

```
export PATH=~/tmp:$PATH
```
and run `ddev start` and `ddev list`, which should fail with that content. Run `ddev version` and `ddev hostname something.ddev.local 127.0.0.1`, which should succeed with that content. Change the script to use other versions valid and invalid.

I'm not sure how to make a test for this behavior, since it depends on the compiled ddev behavior so I don't have a way to override the version constraints cleanly.

## Related Issue Link(s):

OP #737 (docker-compose breaking everything)
OP #711 (ddev hostname shouldn't check docker)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -13,8 +13,8 @@
 [[projects]]
   name = "github.com/Masterminds/semver"
   packages = ["."]
-  revision = "517734cc7d6470c0d07130e40fd40bdeb9bcd3fd"
-  version = "v1.3.1"
+  revision = "15d8430ab86497c5c0da827b748823945e1cf1e1"
+  version = "v1.4.0"
 
 [[projects]]
   name = "github.com/Masterminds/sprig"
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -29,14 +29,14 @@ var RootCmd = &cobra.Command{
 	Long:  "This Command Line Interface (CLI) gives you the ability to interact with the ddev to create a development environment.",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		ignores := []string{"list", "version", "describe", "config", "hostname"}
-		command := strings.Join(os.Args, " ")
+		command := strings.Join(os.Args[1:], " ")
 
 		output.LogSetUp()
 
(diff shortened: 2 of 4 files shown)


cweagans:
:+1: on the code, but I haven't manually tested it yet.

PR #745, opened 2018-03-19
Always set DDEV_UID/DDEV_GUID to run web container as host uid, fixes #635, fixes #740

### LINKED ISSUE #635: File permissions on linux host (folders not being writable from within docker)
Following documentation, especially http://ddev.readthedocs.io/en/latest/users/cli-usage/#typo3-quickstart 
TYPO3 failed to run because some subfolder of typo3temp was not writable from within docker.
After little googling I found : http://ddev.readthedocs.io/en/latest/users/linux_notes/
Changing permissions to make files and folders world writable in typo3temp fixed the issue, 

At later point I had similar issue - some CMS functions were failing because of the insufficient file permissions of the fileadmin and typo3conf folders.

I would expect that 
a) ideally, ddev would not require any fiddling with file permissions
b) if a) is not possible, then  setting file permissions should be documented in the quickstart guide
c) ddev should detect if folders have wrong permissions and warn the user about it.

### LINKED ISSUE #740: web container startup fails on uid or gid collision (linux)

Our web container (and mariadb container) try to make the uid in the container match the uid on the host. 

However, this leads to unintended consequences if the UID or GID already exists in the container. For example, if somebody runs ddev as root, the DDEV_UID will be 0, and so when we try to usermod, it tries to use uid 0:

```
usermod -u 0 nginx
usermod: UID '0' already exists
```

We'll have to detect this and work with it somehow.

rfay:
## The Problem/Issue/Bug:

We need to run our web and db containers predictably as the same uid and gid as the host user. This also prevents security issues where a container process might create files with permissions we don't want.

#635 (File write permissions fail on linux) is one example of this general problem.
#740 (web container fails on gid collision) is another example

The web and db containers are reworked to be able to run as an unprivileged user. This situation resolves all the problems associated with running the container as root. However, it doesn't work everywhere. Linux containers can have UIDs only up to 60000, and the "uid" equivalent on Windows may be much higher than that. 

As a result, on Windows or whenever the UID is high, the containers are still run as root. This is logic decided in ddev.

* Remove .ddev from the project
* ddev config
* ddev start
* Use the site as normally intended. You can `ddev ssh` to see the uid that the user is running as, `ps -ef` in there to see what user the important processes are running as.
* Try some composer commands in there. 

The same automated testing is in play, as no functionality really changes, except the warning to the user when containers are running as root.

## Related Issue Link(s):

#635 
#740 
web container PR: https://github.com/drud/docker.nginx-php-fpm-local/pull/60
db container PR: https://github.com/drud/mariadb-local/pull/51

## Release/Deployment notes:

- [ ] Make new release and push for docker.nginx-php-fpm-local
- [x] Make new release and push for mariadb-local
- [ ] Update this PR with the new released tags

## Security Implications

In the vast majority of cases on Linux or Mac, the web and db containers will run as an unprivileged user. This solves all kinds of file sharing and access problems.

In another set of cases (Docker for Windows and Docker for Mac) the Docker environment prevents any bad coming from containers even if they are run with root privileges (as seems necessary on Windows).

In the remaining cases (say a user ID > 60000 on linux, or some Windows situation we don't have experience with, not Docker for Windows) the container could be run as root and have privileged access to the filesystem where it's being run. We don't have any knowledge of this situation or how it can occur, but it probably would not be best from a security perspective.


rickmanelius:
@rfay I'm fine with performing a review on this, but I'm curious if you feel we have a path forward or if this is not really something that would be ready in time for this release window.

rickmanelius:
I'm assuming myself and @cweagans would be the best reviewers given @tannerjfco is out. Updating to reflect that, but happy to pivot. I'll wait to chime in until I get the affirmative that tests are passing,e etc.

rfay:
@rickmanelius There's a lot of important work in this that resolves a variety of instabilities, so I'd be in favor of getting it in. I'm optimistic. Of course depends on review. Anyway, either this or master is now a proxy for the release for you.

Issue #743, opened 2018-03-16
Document settings file creation and what it means to users


rfay:
## What happened (or feature request):

Followup to #693 (moving settings files creation to start from import-db). We need to document when settings files are created and what it means to the user (and how they can override, etc. AFAIK this has never been in the docs.

@rickmanelius lists there these additional items:

* Getting Started. We should note the generation of these files alongside .ddev/config.yml. Example Addition: Also, if a Database configuration file doesn't already exist, it will be generated from a template.
* Import Assets. This will no longer need "Running this command will provide a prompt for you to specify the location of your database import". This should be deleted.

rickmanelius:
@rfay I'm happy to take care of this within this PR. I found 3 areas that I think could be affected and briefly describe this here https://github.com/drud/ddev/pull/693#issuecomment-373803401.

rickmanelius:
Blocker for the v0.16.0 because we need to ensure that end-users understand that we've changed this behavior. Should be a simple set of doc changes as well as a highlight in the release notes.

rickmanelius:
Fixed in #752

Issue #704, opened 2018-03-02
import-db fails when unable to manage settings files, should warn instead


rfay:
## What happened (or feature request):

import-db fails when the settings files slots are both occupied. It should warn instead. This failure is actually after the db has successfully been loaded.

`Failed to import database for typo3-kit: failed to write settings file for typo3-kit: Failed to get TYPO3 AdditionalConfiguration.php file path: settings files already exist and are being managed by the user`,

rickmanelius:
Similar to https://github.com/drud/ddev/issues/743, I'm going to push to have this documented before we release [v0.16.0](https://github.com/drud/ddev/milestone/26) because this might result in a few support issues. This should also less of concern (hopefully) now that we are more proactive about taking on the settings file generation as a result of https://github.com/drud/ddev/issues/692.

rfay:
This was *solved* in #692, but I'm fine if you want to keep it open as the docs placeholder.

rickmanelius:
This was addressed in code and the docs were touched up in #752.

Issue #468, opened 2017-09-19
Always create settings.ddev.php, Add conditional include to user's settings.php


rfay:
## What happened (or feature request):

Currently, we generate a settings.php or wp-config.php if one doesn't exist, and we have some [explicit support](https://github.com/drud/ddev/blob/master/pkg/ddevapp/config.go#L492-L508) for settings.local.php. However, we don't handle all cases, and we don't communicate to the user what the situation for their particular site.

I'll speak just in Drupal terms here but we need to document and notify the user what is happening if:
* They have a settings.php that does include settings.local.php
* They have a settings.php that does *not* include settings.local.php
* They have no settings.php file (and we generate it for them)

Also we need to be clear about ownership of the settings.php and settings.local.php; we may want to have a settings.ddev.php as well for ddev-only settings.

And our *generated* settings.php should include settings.local.php. It does not currently.

In general: Try for flexible and clearly communicated management of settings files, even though different sites manage these differently  and different platforms like Pantheon manage them differently.

## Related source links or issues:

[#407 - Provide a warning if settings.local.php not included](https://github.com/drud/ddev/issues/407)

ultimike:
Big +1 to a settings.ddev.php file for all the stuff that DDEV generates and currently puts in settings.local.php. 

I _always_ put my own local environment settings in settings.local.php including settings for:

*  Stage File Proxy  
*  Environment Indicator  
*  All of the stuff from Drupal 8's example.settings.local.php  

When using DDEV, I have to create a settings.local2.php for all of my stuff, otherwise DDEV will overwrite it.

ultimike:
Another other option would be for DDEV to not overwrite the _entire_ settings.local.php file - just it's own section.

ultimike:
After reading #692 I'm still not 100% sure that will solve the issue for everyone. I can envision and edge case where someone re-runs `ddev config` (either on purpose or by accident) and blows away any customizations they have in settings.local.php. 

I'm still of the mind that DDEV should only rewrite its own section of settings.local.php on `ddev config` and not the entire file.

rfay:
@ultimike ddev will never ever touch a settings file except one that it generated and that has its own signature in it. There's a big warning about that in every generated file. 
```
/**
 #ddev-generated: Automatically generated Drupal settings.php file.
 ddev manages this file and may delete or overwrite the file unless this comment is removed.
 */
```

rickmanelius:
[...]

PR #752, opened 2018-03-26
Documentation updates to address #468, #704, #743

rickmanelius:
## The Problem/Issue/Bug:

Features in [Milestone v0.16.0](https://github.com/drud/ddev/milestone/26) required documentation.

The three issues were related as it pertains to the timing and mechanism by which DB settings files were generated. This attempts to address these deficiencies.

## Related Issue Link(s):

* [User documentation/support: settings.php/settings.local.php support](https://github.com/drud/ddev/issues/468)
* [import-db fails when unable to manage settings files, should warn instead](https://github.com/drud/ddev/issues/704)
* [Document settings file creation and what it means to users #743](https://github.com/drud/ddev/issues/743)

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -184,10 +184,6 @@ Your application can be reached at: http://example-backdrop-site.ddev.local
 
 ### Database Imports
 
-**Important:** Before importing any databases for your project, please remove its' wp-config.php if using WordPress - or settings.php file in the case of Drupal 7/8, if present.
-
-_ddev will create a wp_config.php or settings.php file automatically if one does not exist. If you already have one you'll need to set the database credentials (user=db, password=db, host=db, database=db)._
-
 Import a database with just one command; We offer support for several file formats, including: **.sql, sql.gz, tar, tar.gz, and zip**.
 
 Here's an example of a database import using ddev:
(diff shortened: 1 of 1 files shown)


rickmanelius:
Added @rfay for review for the [v0.16.0 milestone release](https://github.com/drud/ddev/issues/746)

docs/users/cli-usage.md
```diff
@@ -221,9 +221,24 @@ Docroot Location: web
...
+How do you know if DDEV manages a database settings file? You will see the following comment. Remove the comment and DDEV will not attempt to overwrite it!
+
+```
+/**
+ #ddev-generated: Automatically generated Drupal settings.php file.
+ ddev manages this file and may delete or overwrite the file unless this comment is removed.
+ */
+```
+
+Now that the configuration files have been created, you can start your project with `ddev start` (still from within the project working directory):
```
rfay: Since we're on the subject and have never mentioned this in docs before, why don't we say that the whole strategy fails if settings.php doesn't have the include of settings.local.php. Same basic idea on Wordpress and Backdrop. Not relevant to TYPO3.

rickmanelius: Good point and this might be even more complicated. First, backdrop already has the local include by default https://github.com/backdrop/backdrop/blob/1.9.x/settings.php. WordPress has some issues to work out in https://github.com/drud/ddev/issues/742. TYPO3 is all good.

Let's scope this to include the CMS specific call outs and perhaps reference issues that exist and that need to get updated/removed when resolved (WordPress).

I'll take another stab at this after planning meeting.

Issue #659, opened 2018-02-19
Add docs HOWTO section for small challenges so people can solve them themselves


rfay:
## What happened (or feature request):

People are asking lots of great questions. Let's add the solutions to a docs HOWTO.

- [ ] How to set an environment variable in web container (we probably can't yet without editing of docker-compose.yml, but we should)
- [x] How to turn off xdebug
- [x] How to install an extra needed package in web container (or other random changes to web container)
- [x] Running composer on start (in container)
- [x] Running composer on start (on host)
- [ ] Using drush inside and outside the container
- [x] Viewing nginx, php-fpm, mariadb logs, ddev logs -f
- [x] Solving "Index column size too large" as in #654  ([answer](https://stackoverflow.com/questions/49785419/ddev-import-db-index-column-size-too-large-the-maximum-column-size-is-767-byte/49785420#49785420)
- [ ] Other examples of PHP config override and mysql override
- [x] Drupal 8 and composer workflow with ddev by @dsquaredb ([composer answers](https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev))
- [x] Composer create-project by @ultimike (see the cross-referenced issue) ([composer answers](https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev))
- [ ] Linux users often miss that they have to add the normal user to the "docker" group to be able to use docker without sudo.
- [x] Using `ddev logs` to debug 500 or related errors.
- [ ] Platform.sh instructions/howto
- [x] Getting to the D7 /install.php page for a fresh website ([answer](https://stackoverflow.com/questions/49696010/i-set-up-a-drupal-project-in-ddev-and-i-get-a-pdoexception-when-hitting-the-proj))
- [x] xdebug troubleshooting ([answer](https://stackoverflow.com/questions/49677199/how-do-i-get-xdebug-step-debugging-working-with-ddev))

Please edit this or comment to add howtos as they come up or are requested.

rickmanelius:
Related to this https://github.com/drud/ddev/issues/654.

rickmanelius:
I'm willing to take a stab at some of the bulleted items.

rfay:
I'm stuck on where this should live. Normally things like this would gestate in blog posts or something, or wikis. But our normal approach is in the docs. I'm convinced we should get this going ASAP though, even if we choose the wrong place to start.

rickmanelius:
[...]

PR #758, opened 2018-03-28
Add Information about env vars into documentation

kaystrobach:
## The Problem/Issue/Bug:

Not easy to find how to change / add env vars manually.

add documentation

read documentation

no

### DIFF
--- a/docs/users/extend/customization-extendibility.md
+++ b/docs/users/extend/customization-extendibility.md
@@ -55,3 +55,43 @@ To load the new configuration, run `ddev restart`.
 
 ## Overriding default container images
 The default container images provided by ddev are defined in the `config.yaml` file in the `.ddev` folder of your project. This means that _defining_ an alternative image for default services is as simple as changing the image definition in `config.yaml`. In practice, however, ddev currently has certain expectations and assumptions for what the web and database containers provide. At this time, it is recommended that the default container projects be referenced or used as a starting point for developing an alternative image. If you encounter difficulties integrating alternative images, please [file an issue and let us know](https://github.com/drud/ddev/issues/new).
+
+## Adding environment variables
+
+The prefered way to set env vars is the following composer package:
+
+* https://github.com/helhum/dotenv-connector
+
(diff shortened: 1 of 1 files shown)


docs/users/extend/customization-extendibility.md
```diff
@@ -55,3 +55,43 @@ To load the new configuration, run `ddev restart`.
 
 ## Overriding default container images
 The default container images provided by ddev are defined in the `config.yaml` file in the `.ddev` folder of your project. This means that _defining_ an alternative image for default services is as simple as changing the image definition in `config.yaml`. In practice, however, ddev currently has certain expectations and assumptions for what the web and database containers provide. At this time, it is recommended that the default container projects be referenced or used as a starting point for developing an alternative image. If you encounter difficulties integrating alternative images, please [file an issue and let us know](https://github.com/drud/ddev/issues/new).
+
+## Adding environment variables
```
rfay: ddev is used for many different CMSs, dotenv-connector is only for TYPO3, right? That implies that we have to find a better place for this. Also, we're hoping for a much more general solution in [Make an easy way to populate environment variables in web container](https://github.com/drud/ddev/issues/757).

rfay:
Oh and *thanks so much* for your contribution and reaching out with another perspective! 

Please do update the title and fill in the PR to make the case for this, even though IMO it should land elsewhere.

docs/users/extend/customization-extendibility.md
kaystrobach: dotenv is for any composer based project - not TYPO3 related

rickmanelius:
Agreed. Let's leave this PR open until we find a proper home in #659

rfay:
We're about to make the main docker-compose.yaml ddev-managed, meaning that users will be making a mistake by editing it. We plan to allow it, but it's not going to be mainline. We also plan to allow explicit setting of environment variables in .ddev/config.yaml, https://github.com/drud/ddev/issues/757 .

With your permission I'd like to close this in favor of those future paths.

rfay:
@kaystrobach I hope you can take a look at https://github.com/drud/ddev/pull/816 and see what you think. Thanks!

PR #759, opened 2018-03-28
Update cli-usage.md

kaystrobach:

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -182,9 +182,14 @@ Successfully started example-backdrop-site
 Your application can be reached at: http://example-backdrop-site.ddev.local
 ```
 
-### Flow Framework Quickstart
+### flow framework and neos cms Quickstart
 
 To get started with flow Framework or neos, clone your project repository and navigate to the project directory.
+
+More information can be found here:
+
+* flow: https://flow.neos.io/ | http://flowframework.readthedocs.io/en/stable/
+* neos: https://www.neos.io/ | http://neos.readthedocs.io/en/stable/
 
 ```
 git clone https://github.com/example-user/example-neos-site
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -42,7 +42,7 @@ Quickstart instructions regarding database imports can be found under [Database 
 
 ### Drupal 6/7 Quickstart
 
-Beginning to use ddev with a Drupal 6 or 7 project is as simple as cloning the project's repository and checking out its directory.
+Beginning to use ddev with a Drupal 6 or 7 project is as simple as cloning **your** project's repository and checking out its directory.
 
 ```
 git clone https://github.com/user/my-drupal-site
@@ -74,7 +74,7 @@ Quickstart instructions for database imports can be found under [Database Import
 
 ### Drupal 8 Quickstart
 
-Get started with Drupal 8 projects on ddev either by cloning a git repository or using a new or existing composer project.
+Get started with Drupal 8 projects on ddev either by cloning **your** git repository or using a new or existing composer project.
 
 **Git Clone Example**
 
@@ -115,7 +115,7 @@ Your project can be reached at: http://my-drupal8-site.ddev.local
 
 ### TYPO3 Quickstart
 
-To get started using ddev with a TYPO3 project, clone the project's repository and checkout its directory.
+To get started using ddev with a TYPO3 project, clone **your** project's repository and checkout its directory.
 
 ```
 git clone https://github.com/example-user/example-typo3-site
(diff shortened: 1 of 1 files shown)


docs/users/cli-usage.md
```diff
@@ -42,7 +42,7 @@ Quickstart instructions regarding database imports can be found under [Database
 
 ### Drupal 6/7 Quickstart
 
-Beginning to use ddev with a Drupal 6 or 7 project is as simple as cloning the project's repository and checking out its directory.
+Beginning to use ddev with a Drupal 6 or 7 project is as simple as cloning **your** project's repository and checking out its directory.
```
rfay: I think the use of "your" instead of "the" is appropriate. I think your *emphasis* with double-asterisks, while explaining the reason for the PR, isn't as useful to the general user. So let's just go with "your" in all these cases instead of `**your**`` - I assume you were thinking there was ambiguity and thought people might have checked out the ddev or TYPO3 repo?

docs/users/cli-usage.md
```diff
@@ -182,6 +182,66 @@ Successfully started example-backdrop-site
 Your application can be reached at: http://example-backdrop-site.ddev.local
 ```
 
+### Flow Framework Quickstart
```
rfay: Could this be generalized for neos rather than just "Flow Framework"? Not everybody knows what neos is (I had to google it), perhaps add a link. It may be

docs/users/cli-usage.md
```diff
@@ -182,6 +182,66 @@ Successfully started example-backdrop-site
...
+Run ddev using and you are good to go:
+
+```
+ddev start
+```
+
+When `ddev start` runs, it outputs status messages to indicate the project environment is starting. When the startup is complete, ddev outputs a message like the one below with a link to access your project in a browser.
+
+```
+Successfully started example-backdrop-site
```
rfay: Make sure to update from "backdrop" to the actual names you use in your example.

docs/users/cli-usage.md
kaystrobach: the problem was that **the** suggested that there is an dedicated ddev repository which shoul be used. **your** indicates clearly, that i have to use my own project repository

docs/users/cli-usage.md
kaystrobach: the uri might be wrong, as far as i understood it is **projectname.ddev.local** -> right? maybe this needs to be more generic

docs/users/cli-usage.md
kaystrobach: ok, will chech that again

rickmanelius:
This is neat! It looks like this is a headless TYPO3 setup with https://flow.neos.io/ as the front-end. To that end, we appreciate the PR as it stands, but it probably needs a new section on how to address a headless setup in general with the information re:TYPO3 configuration within #659. It's neat to see CMSes/frameworks that we haven't yet provided official support making their way to DDEV! :)

For now, I think this will need to wait until we get the FAQ section in #659 set up.

kaystrobach:
Please do not mix flow / neos and TYPO3. 

Flow as framework started in the TYPO3 familiy but got separated later, the leftover of flow is called extbase.

Neos is an CMS **competitor** of TYPO3. 

The nice thing is that both communities work together in some parts and both use the same template engine called fluid.

rfay:
We so much appreciate this contribution! We're not ready to support flow though, but would love to have you post your flow setup solution in https://stackoverflow.com/tags/ddev (tagged with "ddev"), if you would! With your permission I'd like to close this and hope to get full flow support over there in Stack Overflow.

kaystrobach:
feel free to do it ... i will also publish the blog article in the next days, maybe it makes sense to link it then

rickmanelius:
Hi @kaystrobach. Thank you for the additional context surrounding the relationship of flow/neo and TYPO3. I must say that I'm less familiar with the history of these projects so I apologize for misspeaking. To that end, there are two categories of feedback on this PR.

* Using "the project" versus "your project." @nearlythere has been spearheading the initiative to ensure our documentation has a consistent style/tone/voice, so we'll want to address this globally if we decide to go down that route. I also lean towards referring to "the" project because it avoids mental thinking around "well technically this repo refers to my companies project and not my project" while the usage of "the project" is more likely to be accurate in most/all use cases.
* A proper home for this information. We're still making a determination of where information like this should best live. We recently started pushing tutorials and howtos to either blog articles or Stackoverflow (see #659). This also brought up the question regarding what it means when we say we "officially support" something. In this case, neo/flow is not part of our core competencies, so we would rather direct someone to a post on Stackoverflow than be prepared to answer support questions ourselves.

To that end, my vote would be to close this PR and then post this on stackoverflow with the [ddev tag](https://stackoverflow.com/questions/tagged/ddev) and then check the box to "Answer your own question" where you can then post the instructions you provided. We are more than happy to do this if you don't have the time, but we want to give you credit!

rickmanelius:
Randy and I crossed paths. All good :)

rfay:
Closing per above, looking forward to blog. Thanks!

nearlythere:
@kaystrobach Please ping me when you publish, I'd love to share on social media and in a community round-up! Thank you!

Issue #748, opened 2018-03-21
Database corruption on host/docker daemon restart "Failed to start sitename: db service health check timed "


rfay:
## What happened (or feature request):

We sometimes see "Failed to start sitename: db service health check timed " situations. The log often has something like "InnoDB: The log sequence number 1600607 in ibdata file do not match the log sequence number 10696720565 in the ib_logfiles!" and implies a crash.

Perhaps as in https://stackoverflow.com/questions/19378565/log-sequence-number-in-ibdata-files-does-not-match

This may be happening on a host restart; the db container may not have a reasonable shutdown sequence, so we end up with a corrupted db. We *could* delete the logfiles on start (so they'd be created automatically). There are probably other options. 

**Note: This is almost always caused by a corrupted database, most often in a larger database. Since v0.16.0, this is generally only caused by docker being shut down in an ungraceful way. Unfortunately, both Docker for WIndows and Docker for Mac shut down without notifying the container during upgrade, with a manual Docker exit, or at system shutdown.  To fix this, remove the database with `ddev remove --remove-data` and then `ddev start` and re-import the db. It can be avoided by stopping or remove-ing your projects before letting Docker exit.**

rfay:
Studying https://docs.docker.com/compose/compose-file/#stop_grace_period I see a stop_grace_period that we might want to increase. The default stop_signal SIGTERM that is normally sent should work OK with the db container, and mysqld is pid 1, so that's correct.

rfay:
I added     stop_grace_period: 1m30s to the docker-compose.yaml in the db service, but it did not affect the behavior. However, on a stopped/broken container I did a `rm ~/.ddev/<project>/mysql/ib*` and then a start and it did file.

rfay:
Interesting paths also might include [live restore](https://docs.docker.com/config/containers/live-restore)

rfay:
[innodb_fast_shutdown](http://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_fast_shutdown) ?

rfay:
Adding this to .ddev/mysql/recovery.cnf seems to completely resolve this problem:
```
[mysqld]
innodb_force_recovery = 1
```
I think we should consider just adding that to the mariadb config. See [forcing recovery](https://dev.mysql.com/doc/refman/5.6/en/forcing-innodb-recovery.html); This would be questionable on a prod system, but that's not what we have here. The '1" value is the gentlest, and it works for me.

[...]

PR #763, opened 2018-03-29
Fix database corruption problem on ddev rm

rfay:
## The Problem/Issue/Bug:

In the recent Drupal-2-geddon updates I noted that a few of my sites (perhaps with larger dbs) would fail to start after ddev start, ddev import-db, ddev rm, ddev start. After further investigation I found that there was database corruption as described in https://github.com/drud/ddev/issues/748 (which is about the situation of a reboot or docker exiting). 

* We weren't stopping containers before forcing their removal, so db container didn't get a chance to clean up and exit correctly. I'm sure we used to, but in our effort to clean up when docker-compose.yaml is missing, we apparently removed that. 
* Added to docker-compose.yaml an increased exit timeout for the db container (default is 10s, that normally is totally adequate, but it doesn't do any harm to have it higher).

With a larger site (larger db) do `ddev start`, `ddev import-db`, `ddev rm`, `ddev start`. It should work without trouble.

I can write a test for this, but it seems quite heavyweight. Not sure.

## Related Issue Link(s):

Definitely relates to #748, which is about corruption when docker goes away (like at reboot of host).

https://github.com/drud/mariadb-local/pull/52 was a PR aimed at fixing #748 (it failed, but was interesting).


rickmanelius:
@rfay I see the @cweagans approved. I can't manually test until tomorrow or Monday... or I can wait for the end-user testing on v0.17.0 review.

rickmanelius:
I'll probably need to use devel to bloat up the DB to a certain threshold and then I can test.

PR #765, opened 2018-03-29
Docs: Remove special linux filesystem instructions, add docker-ce post-install, fixes #754

### LINKED ISSUE #754: Document that users should not run ddev with sudo (docker post-install)

We have had a fair number of linux users who are coming to docker for the first time, and don't know that they should not be using sudo to run ddev. On linux, you have to add your normal user to the "docker" group to run docker (and thus ddev). People don't know that. We should tell them.

rfay:
## The Problem/Issue/Bug:

Since #745 we Linux users don't have the writability problems they once had with files. 

However, #754 reminds us that Linux users need to do the required docker-ce setup steps.

Out with the old, in with the new. Docs only

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,11 +6,11 @@
 
 ## System Requirements
 
-- [Docker](https://www.docker.com/community-edition) version 17.05 or higher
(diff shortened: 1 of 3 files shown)


rfay:
This should not affect Brian.

PR #787, opened 2018-04-12
Windows installer, fixes #786, fixes #780, fixes #798, fixes #701

### LINKED ISSUE #701: Windows install package or script

It's well past time for us to provide a Windows install package or script, rather than the manual instructions we currently give. This is not too hard to do, especially since installation is so easy.

### LINKED ISSUE #780: Poor handling of "sudo ddev hostname" - panic when sudo fails

Over in https://github.com/drud/quicksprint/issues/41 @jeolpittet reported what appears to be a perfectly obvious problem with `sudo ddev hostname` - We do a util.CheckErr() after it, which causes a panic and stacktrace, but the reality is that it can easily fail for various reasons (inadequate permissions, wrong password, etc). 

https://github.com/drud/ddev/blob/73517397dcc6707411a7c961d9fe8be146731ed2/pkg/ddevapp/ddevapp.go#L936-L937

### LINKED ISSUE #786: Implement windows installer

We need a windows installer. [NSIS](http://nsis.sourceforge.net/Main_Page) is quite popular, and can be executed from Linux or macOS, so can be integrated into our CI system.

### LINKED ISSUE #798: Windows: Admin-priv approach to /etc/hosts (ddev hostname)

It's high time we provide a decent approach for windows users to doing the `ddev hostname`. Currently we mention that they should do it in a different (admin) window, but we need to solve this. It's not *that* hard. We avoided it long ago because we just weren't really ready to do full Windows support, but now we're fully committed and we have lots of Windows users... and we just need to make this easier. Soon.

This can also be solved with [DNS support](https://github.com/drud/ddev/issues/416) if we end up doing that. But it needs to be solved.

rfay:
## The Problem/Issue/Bug:

We've not had a windows installer, we need one.

This uses the excellent NSIS to build an installer (and uninstaller).  The installer is built in the regular Circleci build and is available in the artifacts on the Circleci build. For example, on https://circleci.com/gh/drud/ddev/2785#artifacts/containers/0

* Adds a windows installer (`make windows_install`) which packages ddev and sudo (#786)
* Adds sudo so we no longer make Windows users go manually edit hosts file.  (#798)
* Fixes a bug discovered at Drupalcon where a failed sudo attempt results in a stack trace, (#780)

* Remove any existing ddev instances from your %PATH%
* Install using the downloaded installer (Perhaps https://circleci.com/gh/drud/ddev/2726#artifacts/containers/0 )
* Open a new cmd or git-bash window (to get the new %PATH%
* Test basic usage
* Uninstall and verify that the binary, the directory it was installed in, and the %PATH% component are properly updated.
* Take pleasure in not having to manually add hostnames on Windows.

## Related Issues

#786: Implement Windows Installer
#780: Poor handling of `sudo ddev hostname` - We now warn instead of stack trace.
#798: Admin-priv handling for windows users (add sudo command in our install)

### DIFF
--- a/.circleci/circle_vm_setup.sh
+++ b/.circleci/circle_vm_setup.sh
@@ -6,7 +6,7 @@ set -x
 # Basic tools
 
 sudo apt-get update -qq
-sudo apt-get install -qq mysql-client realpath zip
+sudo apt-get install -qq mysql-client realpath zip nsis
 
 # golang of the version we want
 sudo apt-get remove -qq golang && sudo rm -rf /usr/local/go &&
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,9 @@ jobs:
 
       # Now build using the regular ddev-only technique - this results in a fully clean set of executables.
       - run:
-          command: make -s clean linux darwin windows
+          command: |
+            make -s clean linux darwin windows
+            makensis -DVERSION=$(git describe --tags --always --dirty) winpkg/ddev.nsi
           name: Build the ddev executables
 
       # Run the built-in ddev tests with the executables just built.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,8 @@ jobs:
           command: |
             make -s clean linux darwin windows
             makensis -DVERSION=$(git describe --tags --always --dirty) winpkg/ddev.nsi
+            curl -o /tmp/sudo.zip -O -L https://github.com/mattn/sudo/releases/download/v0.0.1/sudo-x86_64.zip
+            unzip -d bin/windows/windows_amd64 /tmp/sudo.zip
           name: Build the ddev executables
 
       # Run the built-in ddev tests with the executables just built.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,10 +24,7 @@ jobs:
       # Now build using the regular ddev-only technique - this results in a fully clean set of executables.
       - run:
           command: |
-            make -s clean linux darwin windows
-            makensis -DVERSION=$(git describe --tags --always --dirty) winpkg/ddev.nsi
-            curl -o /tmp/sudo.zip -O -L https://github.com/mattn/sudo/releases/download/v0.0.1/sudo-x86_64.zip
-            unzip -d bin/windows/windows_amd64 /tmp/sudo.zip
+            make -s clean linux darwin windows_install
           name: Build the ddev executables
 
       # Run the built-in ddev tests with the executables just built.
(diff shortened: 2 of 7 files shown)


rfay:
I've removed the WIP from this. What I'd like to do is go ahead and get it in, and then have @jdarrohn "theme" it before the release. Functionality first, then the theming. None of that should stop Jed from working on it.

rfay:
One thing I noted testing this the other day is that Windows challenged the install exe, and you had to go around the challenge to get it done. I hadn't seen this before, and the workaround is one click, but it surprised me. We need to document it and work toward solving it, (solving it probably not in this PR)

tannerjfco:
I took this for a spin on my Windows 10 Pro machine. This machine did not have docker or anything ddev-related installed previously. The installer seems to work well, though I did hit a snag or two below.

- Blocked by Windows: On my first attempt running the ddev installer, Windows blocked the application from running. I expect there's some kind of application signing we might need to look into.

winstaller-block


- Question on Installer options - I might not be recalling correctly, but I thought we were going to have the option to not install sudo? Not sure how big of a deal this would be.

winstaller-options

- Question on dependencies - we don't seem to check for them. I was able to install without any sort of Docker installation present. Is this feasible? I think it could be helpful to at least warn users if they don't have docker installed that they will need it. If this is too complex to address currently, perhaps we could add some text to one of the installer windows to warn they need docker installed?

These smaller issues aside, this work is fantastic, and once I was past the initial block of the installer by windows, everything worked fine. I installed, opened command prompt, and had ddev available. I also tested the `ddev hostname` command to test sudo, which worked just as I expected.

rfay:
Thanks @tannerjfco - 
* I didn't see the "Windows protected your PC" until just the other day, not sure if that was just lurking somewhere or Windows changed something or started not liking the install. I doubt that we should deal with it in this round other than by telling people to click "More info" and "Run anyway"
* I'm sad to say my mastery of the language wasn't enough to add sudo as a separate component. It really should be easy, but I just didn't get it. I should take another try.
* Definitely agree that we should look for docker, or at least tell people about it. Regardless, when they run ddev it will tell them about it. Homebrew install only tells them about it.
* It also needs the sudo license.

rfay:
I separated the ddev and sudo components so sudo is now optional; also added sudo license. I'll spend a bit more time on it but it could be infinite... And it's such a step forward from what we've been doing to windows users.

rickmanelius:
* I liked the warnings for Docker and Docker compose. It might be useful to note the need for Docker for Windows with a link.
* I saw the security warning. Once we can get the signed certificate, that will be helpful. The only way around it was for me to download it versus run directly from Microsoft Edge.
* Let's link ddev.com instead of ddev.drud.com.
* I must confess that when I had Docker for Windows and ddev installed, I'm a novice Windows user and feel naked without a normal terminal. I know with 30 minutes of digging I will find my way, but if there are any handy references to accelerate that process, I would appreciate it.

PR #790, opened 2018-04-16
Add nodejs and npm to web container

rfay:
## The Problem/Issue/Bug:

People want node and npm for their web container.

Add it, from https://github.com/drud/docker.nginx-php-fpm-local/pull/68

Use this (or just set your config.yaml to use `webimage: drud/nginx-php-fpm-local: 20180416_add_node`

`ddev ssh` in there and do something useful.


misterboe:
I checked this in two installations - everything works fine

PR #791, opened 2018-04-17
Remove com.ddev.container-type from documentation

tmotyl:
PR https://github.com/drud/ddev/pull/215 has changed the `com.ddev.container-type` label to 
`com.docker.compose.service`.

Documentation  https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/
and solr example needs to be updated.

### DIFF
--- a/docs/users/extend/custom-compose-files.md
+++ b/docs/users/extend/custom-compose-files.md
@@ -24,8 +24,6 @@ When defining additional services for your project, we recommended you follow th
 
   - `com.ddev.app-url: $DDEV_URL`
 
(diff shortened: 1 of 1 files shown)


tannerjfco:
This is actually a label that was deprecated some time ago, and should be removed from the documentation. ddev uses the `com.docker.compose.service` provided by docker-compose to determine the 'container type'.

tmotyl:
updated the issue description and PR to match comment from @tannerjfco

tmotyl:
here you go

PR #802, opened 2018-04-20
Only build xz images on tag build, fixes #801

### LINKED ISSUE #801: Remove xz build from (most?) circleci builds

We added xz compression into the Circleci build for Quicksprint and because it was Brian's favorite. But it extends our build by more than 10 minutes, which is too much.

We can either completely remove it, or we can remove it from all but the tag build.

rfay:
## The Problem/Issue/Bug:

The xz compression of our docker images takes *forever*. It was added for quicksprint, and is only needed on releases. So remove it from normal builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,9 +113,12 @@ jobs:
           command: bin/linux/ddev version
           name: ddev version information
 
+      # We only build the xz version of the docker images on tag build.
       - run:
-          command: ./.circleci/generate_artifacts.sh $ARTIFACTS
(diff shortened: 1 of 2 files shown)


rfay:
Tag build seems to have the right artifacts in https://circleci.com/gh/rfay/ddev/337#artifacts/containers/0 (Had to build on rfay account to avoid pushing all kinds of weird tags)

rfay:
I pulled this without review because it's trivial and it sat there for more than a week without a review.

Issue #800, opened 2018-04-20
Windows: Warning: containers will run as root. 


rfay:
## What happened (or feature request):

Currently this warning shows to Windows users 2x for every command. It's not useful and it's annoying:

`Warning: containers will run as root. This is fine on Docker for Windows or Docker for Mac, but could be a security risk on Linux.`

I think we should probably run windows containers as an arbitrary user (not root), and that would just make this go away when it's not needed.

PR #803, opened 2018-04-20
Run containers as UID 1000 on Windows

rfay:
## The Problem/Issue/Bug:

OP #800 - Awkward warning (repeated lots) about Windows containers running as root.

There's really no reason to run them as root, we just couldn't work with Windows UID/GID. Go with uid 1000, which should work fine at least on Docker for Windows.

Run on windows, verify correct behavior:
* Should be able to create upload files in sites/default/files
* Files created should be owned by the *host* user
* The warning should be gone.


rickmanelius:
I think @cweagans  will be able to test this more quickly than me. I'm still slow on Windows because I don't spend a lot of time using Parallels except when absolutely necessary. Adding him to the reviewer list.

rfay:
This is WIP, not ready for testing.

cweagans:
I don't have a functioning Windows install at the moment, but it's on my list for the weekend. I can do it sooner if necessary though.

rfay:
This is ready for review; I went another way, and added a test for writeability of files directory.

rfay:
Note that the latest push of this has the windows installer, so should be pretty easy to do a runthrough.

PR #804, opened 2018-04-20
Set COLUMNS and LINES in container shell, fixes #799

### LINKED ISSUE #799: ddev ssh does a poor job with long command lines and $TERM in general

When lines get long, `ddev ssh` starts to get really ugly, wrapping things wrong and messy. This is really from `docker exec`. 

[This docker issue](https://github.com/moby/moby/issues/33794) mentions an excellent workaround which we could just implement.

rfay:
## The Problem/Issue/Bug:

OP #799 explains our trouble with $COLUMNS and $LINES inside `ddev ssh`.

Set those values when we use docker-compose.

`ddev ssh` with and without and observe the behavior on a long line.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2,6 +2,7 @@ package ddevapp
 
 import (
 	"fmt"
+	"golang.org/x/crypto/ssh/terminal"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


rfay:
This doesn't seem to work on Windows, at least in the git-bash window or the CMD window.

rfay:
This is still a little tweaky. The LINES and COLUMNS are set on `ddev start`, they don't seem to be reset on `ddev ssh`, which is pretty strange. So bottom line, if the window you're using is either the same (unresaized) or has the same size that you did the start from, things work right. But if you have a few windows of different sizes, doesn't seem to work right.

PR #816, opened 2018-04-30
Document adding extra env variables, fixes #757, also #758 (docs only)

### LINKED ISSUE #757: Make an easy way to populate environment variables in the web container

TYPO3 users often request a way to set environment variables in the web container. 

This is not hard to do with docker-compose.yaml, but we're really on course to take complete control of that file and recreate it on every start.

So we should provide a way to do it in config.yaml.

rfay:
## The Problem/Issue/Bug:

OP #757 points out that especially TYPO3 users need to set environment variables in the container. It's actually quite easy to do (without editing the main docker-compose.yaml). This just tells how to do it.

Thanks @kaystrobach - Hope you can review this.

Adds docs with simple technique.

### DIFF
--- a/docs/users/extend/customization-extendibility.md
+++ b/docs/users/extend/customization-extendibility.md
@@ -13,6 +13,24 @@ A collection of vetted service configurations is available in the [Additional Se
 
 If you need to create a service configuration for your project, see [Defining an additional service with Docker Compose](custom-compose-files.md)
 
+## Providing custom environment variables to a container
+
+Each project can have an unlimited number of .ddev/docker-compose.*.yaml files, so it's easy to maintain custom environment variables in a .ddev/docker-compose.environment.yaml file (the exact name doesn't matter, if it just matches docker-compose*.y*ml).
+
+For example, 
+
+```
+version: '3'
+
+services:
+  web:
+    environment:
+      - TYPO3_CONTEXT=Development
+  db:
+    environment:
+      - SOMETHING=something special
+```
+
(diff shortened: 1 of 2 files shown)


docs/users/extend/customization-extendibility.md
```diff
@@ -13,6 +13,24 @@ A collection of vetted service configurations is available in the [Additional Se
 
 If you need to create a service configuration for your project, see [Defining an additional service with Docker Compose](custom-compose-files.md)
 
+## Providing custom environment variables to a container
+
+Each project can have an unlimited number of .ddev/docker-compose.*.yaml files, so it's easy to maintain custom environment variables in a .ddev/docker-compose.environment.yaml file (the exact name doesn't matter, if it just matches docker-compose*.y*ml).
```
tannerjfco: nit, the last mentioned file name format should be docker-compose.*.y*ml. the first mentioned one also uses .yaml vs .y*ml. i wonder if we should just document one of the two yaml extensions rather than using the *. not sure how much confusion that could be for some people

docs/users/extend/customization-extendibility.md
```diff
@@ -13,6 +13,24 @@ A collection of vetted service configurations is available in the [Additional Se
...
+
+For example,
```
tannerjfco: it might help to provide an example file name here for extra clarity. I think it could also be useful to spell out that this example would add env vars to both the web and db containers.

PR #817, opened 2018-05-01
Revamp docker-compose file ordering and choosing, fixes #815

### LINKED ISSUE #815: docker-compose.override.y*ml is ignored

Due to a flaw in DdevApp.ComposeFiles the docker-compose.yaml is always discarded. Doesn't do much good that way :)

rfay:
## The Problem/Issue/Bug:

There were two problems here:
1. It would pick up docker-compose* (like docker-compose.yaml.bak)
2. The docker-compose.override.yaml was being discarded, so never did any overriding at all.

* Complete logic change; a new slice is built instead of trying to re-use the old one.
* Error checking in case both docker-compose.yml and docker-compose.yaml exist, and in case docker-compose.override.yaml and yml exist. 

* Create docker-compose.*.yaml with files like this example docker-compose.1.yaml:
```
version: '3'
services:
  web:
    environment:
      - TEST1="from docker-compose.1.yaml"
      - TEST="TEST value provided by docker-compose.1.yaml"
```

* Create a docker-compose.override.yaml that overrides TEST
* Start the project, `ddev ssh`. Verify that TEST1 or related exists and has right value. Verify that TEST is the value provided by docker-compose.override.yaml.

Added TestMultipleComposeFiles() to make sure the logic ends up correct.

## Related Issue Link(s):

OP #815 discovered the discarding of docker-compose.override.yaml. 
No issue was created for the importing of non-yaml files

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -518,24 +518,24 @@ func (app *DdevApp) ImportFiles(imPath string, extPath string) error {
 func (app *DdevApp) ComposeFiles() ([]string, error) {
 	files, err := filepath.Glob(filepath.Join(app.AppConfDir(), "docker-compose*.y*l"))
 	if err != nil || len(files) == 0 {
-		return []string{}, fmt.Errorf("Failed to load any docker-compose.*y*l files: %v", err)
+		return []string{}, fmt.Errorf("failed to load any docker-compose.*y*l files: %v", err)
 	}
 
 	mainfiles, err := filepath.Glob(filepath.Join(app.AppConfDir(), "docker-compose.y*l"))
 	// Glob doesn't return many errors, so just CheckErr()
 	util.CheckErr(err)
 	if len(mainfiles) == 0 {
-		return []string{}, fmt.Errorf("Failed to find a docker-compose.yml or docker-compose.yaml")
+		return []string{}, fmt.Errorf("failed to find a docker-compose.yml or docker-compose.yaml")
 
 	}
 	if len(mainfiles) > 1 {
-		return []string{}, fmt.Errorf("There are more than one docker-compose.y*l, unable to continue.")
+		return []string{}, fmt.Errorf("there are more than one docker-compose.y*l, unable to continue")
 	}
 
 	overrides, err := filepath.Glob(filepath.Join(app.AppConfDir(), "docker-compose.override.y*l"))
 	util.CheckErr(err)
 	if len(overrides) > 1 {
-		return []string{}, fmt.Errorf("There are more than one docker-compose.override.y*l, unable to continue.")
+		return []string{}, fmt.Errorf("there are more than one docker-compose.override.y*l, unable to continue")
 	}
 
 	orderedFiles := make([]string, 1)
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -513,28 +513,47 @@ func (app *DdevApp) ImportFiles(imPath string, extPath string) error {
 }
 
 // ComposeFiles returns a list of compose files for a project.
+// It has to put the docker-compose.y*l first
+// It has to put the docker-compose.override.y*l last
 func (app *DdevApp) ComposeFiles() []string {
-	files, err := filepath.Glob(filepath.Join(app.AppConfDir(), "docker-compose*"))
-	if err != nil {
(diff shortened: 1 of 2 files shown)


tannerjfco:
So I took this for a spin, but I wasn't able to reproduce the original issue against master. Here's how I tested:

- ddev config a site
- add a docker-compose.1.yaml as noted in instructions. I used the same contents provided in example.
- add a docker-compose.override.yaml, as that was reported to be discarded. I used similar contents provided for the docker-compose.1.yaml example, changing the text of the vars to indicate what file was setting them
- ddev start
- inspect output of `env` in web container.
- the test vars added matched the values set in docker-compose.override.yaml, which indicates to me that the override file was picked up, and was the last file loaded

I had similar result from testing the binary from this PR, but I'm not sure what we're addressing here since I can't seem to reproduce the original issue.

rfay:
@tannerjfco thanks!

Could you try this with v0.16.0.... Just create a docker-compose.override.yaml with these contents:

```
version: '3'

services:
  web:
    environment:
      - OVERRIDE=Overridden by docker-compose.override.yaml
```

`ddev rm && ddev start`
`ddev ssh` and see if $OVERRIDE is set.

(Note that this will *work* if the docker-compose.override.yaml is changed to docker-compose.tanner.yaml.)

Then do the same with this patch.

rickmanelius:
I must confess that I, like @tannerjfco in https://github.com/drud/ddev/pull/817#issuecomment-385743223, I was unable to recreate the issue with v0.17.0 because I can run `env | grep TEST` and I see the original entry for TEST1 from docker-compose.1.yaml and the overridden value of TEST from docker-compose.overridden.yaml. I see the request to review using v0.16.0, but I'm unclear why to step back a version if this appears to be addressed in v0.17.0?

rickmanelius:
FWIW, I can confirm that v0.16.0 wasn't properly reading docker-compose.override.yaml based on my pre/post test. Still curious what the issue/regression is/was on v0.17.0

rfay:
I apologize for mentioning v0.16.0, I meant v0.17.0. I just repeated the test in https://github.com/drud/ddev/pull/817#issuecomment-385743223 with v0.17.0 and it fails, as I would expect, because I know of no work in this area for many, many months.

rickmanelius:
Retesting this PR https://github.com/drud/ddev/pull/817 at commit 1b25a76f91fe054ac47afbd9fa1665941bf0d943

## Preparation

Attempting to start from ground zero while paying extra special attention to persisting config.yml and docker-compose.yml files.

* Remove all active projects from `ddev list`
* Remove all trace of drupal-7.59 by `rm -rf ~/.ddev/drupal-7.59`
* Remove old drupal-7.59 folder. Unzip two version (drupal-7.59a and drupal-7.59b)

PR at 1b25a76f91fe054ac47afbd9fa1665941bf0d943

* `ddev version` => v0.17.0-15-g1b25a76f
* `ddev config` with stock answers
* Created .ddev/docker-compose.override.yaml with the following information https://github.com/drud/ddev/pull/817#issuecomment-385743223
* Ran `ddev start` and I confirmed that a stock .ddev/docker-compose.yaml was generated.
* `ddev ssh` and then `echo $OVERRIDE` shows "Overridden by docker-compose.override.yaml"
* Ran `ddev rm` and `ddev start` and `ddev ssh` and `echo $OVERRIDE` shows "Overridden by docker-compose.override.yaml"

v0.17.0

* `brew install ddev`
* `ddev version` => v0.17.0
* `ddev config` with stock answers
* Created .ddev/docker-compose.override.yaml with the following information https://github.com/drud/ddev/pull/817#issuecomment-385743223
* `ddev ssh` and then `echo $OVERRIDE` shows up blank

It turns out, I was not properly refreshing my $PATH once I switched from the PR back to brew. Next time, I will simply checkout the tag directly so as to avoid this mishap.

rickmanelius:
Running one last test to confirm Solr works as per https://ddev.readthedocs.io/en/latest/users/extend/additional-services/.

rickmanelius:
So I ran another test with the solr service. I added 3 envirionment variables for web (TEST1, TEST2, TEST2), I then had the first two overridden in Solr and the last one overriden in override. I confirmed that the overrides stacked in the proper order.

+1 from me!

rfay:
I went ahead and pulled as Tanner didn't seem to have concerns, and it has a pretty solid test.

PR #819, opened 2018-05-01
Override docker-compose.yaml on ddev start, fixes #762

### LINKED ISSUE #762: ddev should always recreate docker-compose.yaml

Nearly every release we're having to tell people to delete docker-compose.yaml. We should just take it over and always recreate it. It's possible it should be renamed to something like ddev-docker-compose.yaml. Additional services can add properly named alternates that we won't touch.

rfay:
## The Problem/Issue/Bug:

OP #762: We keep having to tell people to delete docker-compose.yaml on ddev upgrades. We made the decision to take control of docker-compose.yaml

This PR addresses the changes for docker-compose.yaml, another PR will address the changes to config.yaml handling.

* Looks for docker-compose.yaml `#ddev-generated` in docker-compose. If not found, renames the file to docker-compose.yaml.bak and warns the user.
* Creates the docker-compose.yaml on every start
* Overrides can be done in docker-compose.*.yaml

* Run `ddev start` with this version
* Existing docker-compose.yaml should be renamed to .bak, and new one created.
* Run it again, with or without changes. A new one should be created. No warning should be given.

Nothing at this point.

## Related Issue Link(s):

OP #762

### DIFF
--- a/docs/users/troubleshooting.md
+++ b/docs/users/troubleshooting.md
@@ -77,7 +77,7 @@ We've seen cases where this is caused by old databases that are not compatible w
 Note: Your project database will be destroyed by this procedure.
 
 1. `ddev remove --remove-data`
-2. rm -r .ddev (removes the config.yaml and docker-compose.yml, do this only if you haven't modified those)
+2. rm -r .ddev (removes the config.yaml and docker-compose.yml and any custom nginx/php/mariadb config you may have added. It makes sense to make a backup of these before proceeding.)
 3. `ddev start` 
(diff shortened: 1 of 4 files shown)


rfay:
@rickmanelius I think you requested changes because of required docs. Please check the paragraph that I already had slipped in upstream, https://github.com/drud/ddev/blob/master/docs/users/extend/custom-compose-files.md#restrictions-on-the-docker-composeyaml-file and see if that does it for you, and approve if it does.

PR #820, opened 2018-05-01
Stop using container versions in config.yaml, version the APIVersion, fixes #762, fixes #713

### LINKED ISSUE #713: config.yaml includes explicit container versions, maybe it should not

* config.yaml (as created by `ddev config` includes explicit web and nginx image versions.
* If you leave those explicit definitions out, then the versions specified by ddev are used by all commands.
* If you `ddev config` with an existing project, all the container versions are updated (whether or not they were empty before).

There are a number of good things about this approach:

* If people check in or otherwise respect.ddev/config.yaml, they can expect ddev behavior to be unchanged and unaffected by ddev version upgrades.

Bad things about this approach:

* ddev upgrades don't automatically result in the required upgrade for the project. This can occasionally be bad, when ddev expects a new behavior and the older container doesn't provide it, or when the user is expecting to automatically have new ddev features.

Paths forward:

* Leave as is
* Stop unconditionally writing the version to config.yaml (but respect if it's there).

I would prefer to go forward with *not* populating those in the .config.yaml, but providing them as commented-out options. The related problem is that we need to rewrite docker-compose.yml all the time (on every start) to properly respect ddev and container upgrades.

This is closely related to our long-running discussion about who owns what in .ddev.

### LINKED ISSUE #762: ddev should always recreate docker-compose.yaml

rfay:
## The Problem/Issue/Bug:

OP #762: 
* Stop default override of webimage etc. Now, if they are included in config.yaml, they will be respected, but they are not written by `ddev config`
* Version the config.yaml file with the ddev version.
* Pester the user if their config.yaml doesn't have the same version as ddev's version.

* Try a `ddev start` on a currently-setup project. It should complain about the config.yaml version.
* Try adding a specific webimage: yaml element, like `webimage: drud/nginx-php-fpm-local:v1.2.0` and `ddev start`. It should start with that image.
* Try `ddev config` and verify that the new config.yaml has no image statements populated.

The existing tests are probably adequate and were updated for this.

## Related Issue Link(s):

OP #762 
Related #713

### DIFF
--- a/cmd/ddev/cmd/describe_test.go
+++ b/cmd/ddev/cmd/describe_test.go
@@ -77,33 +77,21 @@ func TestDescribe(t *testing.T) {
 		args = []string{"describe", "-j"}
 		out, err = exec.RunCommand(DdevBin, args)
 		assert.NoError(err)
-		logStrings := strings.Split(out, "\n")
+		logItems, err := unmarshallJsonLogs(out)
+		assert.NoError(err)
 
-		assert.True(len(logStrings) >= 2)
+		// The description log should be the last item; there may be a warning
+		// or other info before that.
+		data := logItems[len(logItems)-1]
+		assert.EqualValues(data["level"], "info")
+		raw, ok := data["raw"].(map[string]interface{})
+		assert.True(ok)
+		assert.EqualValues(raw["status"], "running")
+		assert.EqualValues(raw["name"], v.Name)
+		assert.EqualValues(raw["shortroot"].(string), ddevapp.RenderHomeRootedDir(v.Dir))
+		assert.EqualValues(raw["approot"].(string), v.Dir)
 
-		var descWasHit bool
-		// Wander through the json output lines looking for our describe.
-		for _, entry := range logStrings {
-			if entry != "" { // Ignore empty line.
-				// Unmarshall the json results. Normal log entries have 3 fields
-				data := make(log.Fields, 4)
-				err = json.Unmarshal([]byte(entry), &data)
-				assert.NoError(err)
-				if data["level"] != "info" {
-					continue
-				}
-				raw, ok := data["raw"].(map[string]interface{})
-				assert.True(ok)
-				assert.EqualValues(raw["status"], "running")
-				assert.EqualValues(raw["name"], v.Name)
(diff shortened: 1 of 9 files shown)


rfay:
I added a couple of lines of context on the docker images, and improved the documentation in config.yaml a couple of other way. Also caught a bug where the APIVersion was not updated on `ddev config`. 

```
# This config.yaml was created with ddev version v0.17.0-26-g172430da 
# webimage: drud/nginx-php-fpm-local:v1.3.0
# dbimage: drud/mariadb-local:v0.9.0
# dbaimage: drud/phpmyadmin:v0.2.0
# However we do not recommend explicitly wiring these images into the
# config.yaml as they may break future versions of ddev.
# You can update this config.yaml using 'ddev config'.

# Key features of ddev's config.yaml:

#   http://projectname.ddev.local and https://projectname.ddev.local

# type: <projecttype>  # drupal6/7/8, backdrop, typo3, wordpress, php

# docroot: <relative_path> # Relative path to the directory containing index.php.

# php_version: "7.1"  # PHP version to use, "5.6", "7.0", "7.1", "7.2"

# You can explicitly specify the webimage, dbimage, dbaimage lines but this 
# is not recommended, as the images are often closely tied to ddev's' behavior,
# so this can break upgrades.

# webimage: <docker_image>  # nginx/php docker image. 
# dbimage: <docker_image>  # mariadb docker image. 
# dbaimage: <docker_image>

# router_http_port: <port>  # Port to be used for http (defaults to port 80)
# router_https_port: <port> # Port for https (defaults to 443)

# xdebug_enabled: false  # Set to true to enable xdebug and "ddev start" or "ddev restart"

#additional_hostnames:
# - somename
# - someothername
# would provide http and https URLs for "somename.ddev.local"
# and "someothername.ddev.local".

# provider: default # Currently either "default" or "pantheon"
#
# Many ddev commands can be extended to run tasks after the ddev command is
# executed.
# See https://ddev.readthedocs.io/en/latest/users/extending-commands/ for more
# information on the commands that can be extended and the tasks you can define
# for them. Example:
#hooks:
#  post-import-db:
#    - exec: "drush cc all"
```

PR #824, opened 2018-05-04
Support .mysql files for import along with related files, fixes #812, fixes #516

### LINKED ISSUE #516: Consider supporting *.mysql file type for import-db

Rick noted with a ddev demo that an import of a backup-migrate module backup failed.  For example, `ddev import-db --src=APLVDatabaseBackup-2017-10-23T06-30-32.mysql.gz`:

```
Failed to import database for aplv: no .sql files found to import
```
The issue here is that backup-migrate uses the very nonstandard "*.mysql" file type inside the gzip file, and ddev recognizes only the more-likely "*.sql" filetype.

backup-migrate has gzip files like `APLVDatabaseBackup-2017-10-23T06-30-32.mysql.gz` containing a file like `APLVDatabaseBackup-2017-10-23T06-30-32.mysql`. If you gunzip and give the .mysql file to ddev import-db, it imports just fine. 

It shouldn't be hard to expand our vision to looking for *.mysql as equivalent to *.sql

### LINKED ISSUE #812: Add ".mysql" to allowable database extensions for import
Databases that are exported from Drupal using the backup_migrate module are exported with the file extension "mysql.gz." Ignoring the gzip extension, DDev's import-db command accepts only "sql," not "mysql," as an extension, forcing the user to rename the file in order to import it.

Could you please add "mysql" as an acceptable database file-extension for import-db?

rfay:
## The Problem/Issue/Bug:

OP #812 and #516: People who use backup_migrate like to use .mysql files and the related items. This PR just allows us to import those and their variants as if they were .sql files.

Download a mysql file, maybe from a backup_migrate instance. Use ddev  import-db on it, verify. 

This adds all variants to TestDevImportDB()

## Related Issue Link(s):

#812, #516


rickmanelius:
I'm actually having a few issues with this.

* The default export of backup migrate is actually .mysql.gz. I attempt to use gunzip on the command line but I get that the file is "not in gzip format".
* When I tell backup_migrate to export a .mysql file, it does so successfully. However, if I already had a DB before, it doesn't appear to bring in the node table. However, this seems to be an issue with backup migrate, which is odd. I'll need to pull a different DB to test with to verify that the import happens correclty and this oddity is more on an incorrect export from backup migrate.

rfay:
Hmm, the tests do mysql.gz, mysql, mysql.tar.gz, etc. You shouldn't have to unzip. But gzip should let you. 

It may be that backup_migrate does not do a "drop table" before bringing stuff in (you can look inside your export).  I think we should be dropping the db on import:  https://github.com/drud/ddev/issues/810 - very important IMO

PR #827, opened 2018-05-07
Add windows_install to nightly and tag builds, fix nightly build

rfay:
## The Problem/Issue/Bug:

https://github.com/drud/ddev/pull/787 went in last week, adding a windows installer, but the nightly and tag builds didn't include it, so the artifact script failed. Fix that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
 
       - run:
           command: |
-            make -s linux darwin windows
+            make -s linux darwin windows_install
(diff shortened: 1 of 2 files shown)


rfay:
As of 20180507 10:49am the IP routing from Circleci to ftp.drupal.org is broken; it only gets 2 hops and stalls. Best to just let this sit a while.

rfay:
This now passed manual nightly, so I pulled it, no real need for trivial review.

PR #857, opened 2018-05-15
Document AUR repository for installing ddev

cweagans:
## The Problem/Issue/Bug:

There's an AUR repo for ddev, so we should mention it in the docs.

Adds docs.

## Related Issue Link(s):

Fixes https://github.com/drud/ddev/issues/818


rfay:
Could you just add the command people would use to install, for those that are new to AUR?

rfay:
I guess that's a more varied request than I knew, so this is good to pull then.

PR #864, opened 2018-05-17
[regression fix] Make project uniqueness constraint work again, fixes #136

### LINKED ISSUE #136: ddev project name (and hostname) uniqueness should be enforced

I tried a `ddev config` and used an existing 'name'. Then I did a `ddev start`. The new repo/ddev app took over the containers of the previously-existing app without warning.

I would expect that either
1. 'name' is *not* unique, where multiple apps named 'drupal8' could co-exist
or
2. uniqueness of names is enforced
or 
3. there's a "machine name" of an app name that is enforced to be unique

Anyway, I don't expect a `ddev start` in another repo to take over the containers already built for another app.

## How to reproduce it (as minimally and precisely as possible):

1. Config and start a container with the name 'abracadabra`
2. Go to another repo and config and start a container with the name 'abracadabra'

rfay:
## The Problem/Issue/Bug:

#136: We had made running projects have to have unique names, but that got lost somewhere. This restores it.

This does not solve:
* Duplicate project name when one is not running
* One project taking over another project's database inappropriately (databases are uniquely named as well)
* Duplicated hostnames (across projects) breaking ddev-router. (Adding an additional_hostname to one project that is already taken by another)

Take seriously when the error says there's a duplicate site.

Start a site.
In another directory, start a site with the same name. You should get a failure.

* The test for uniqueness was updated to catch this case. 

## Related Issue Link(s):

OP #136

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -102,7 +102,7 @@ func (app *DdevApp) Init(basePath string) error {
 	if err == nil {
 		containerApproot := web.Labels["com.ddev.approot"]
 		if containerApproot != app.AppRoot {
-			return fmt.Errorf("a web container in %s state already exists for %s that was created at %s", web.State, app.Name, containerApproot)
+			return fmt.Errorf("a project (web container) in %s state already exists for %s that was created at %s", web.State, app.Name, containerApproot)
 		}
 		return nil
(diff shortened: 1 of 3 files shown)


rfay:
There was actually already a test for uniqueness, but it was on app.Init() only. I added a stanza to that test in 4907e43 that also checks in GetActiveApp() which was where it was also needed.

Issue #421, opened 2017-07-23
Consider improving test independence by a site naming strategy


rfay:
## What happened (or feature request):

We still use preconfigured container names and site names in our testing, despite our best intentions. This means that only one set of tests can run on one system at a time, which is kind of limiting, especially for our new Windows and Mac tests, which don't have unlimited systems to run on.

IMO we can:

1. Name test sites with a unique appendage (but still cache downloads without the appendage)
2. Fix the router test so that it either doesn't run in some cases or ???

This approach will also prevent accidental reuse of database information for sites. Currently, tests that run on a pre-existing system like our dev laptops or Windows or OSX tests could conceivably (likely) accidentally reuse databases from previous runs.

## Related source links or issues:

rfay:
This one is not hard and is worth doing. It will save us massive testing time because we stumble on it repeatedly, both in development and in test.

rfay:
Not happening at this time.

PR #876, opened 2018-05-23
Add debugging in build.sh for running sites problem

rfay:
## The Problem/Issue/Bug:

Our windows build especially often fails on running tests (pkg tests require no projects running at all, which is a problem in its own right - #421, which is the whole problem with this.)

Example: https://gist.github.com/drud-test-machine-account/943c09556790ade2a7edbfc07bb6434b

Does a ddev list and docker ps to help us understand the failures we see.

Note that this PR affects *only* the macOS and Windows testbots, and nothing else, so it's pretty harmless. Just some debugging info.


andrewfrench:
Looks good to me as far as providing debug context.

PR #877, opened 2018-05-23
Add linux deriving host.docker.internal, fixes #843

### LINKED ISSUE #843: v0.18.0 xdebug instructions don't work on linux due to no "host.docker.internal" hostname

It turns out Xdebug doesn't work on linux with our provided instructions on v0.18.0 because Docker didn't get `host.docker.internal` into the linux version. 

https://github.com/docker/for-linux/issues/264

This isn't too hard to sort out, but it's sad.

rfay:
## The Problem/Issue/Bug:

Although the hostname "host.docker.internal" is defined and supported for Docker for Windows and Docker for Mac, it's not provided in the container on linux docker at this time. That breaks our approach to xdebug in v0.18.0, which worked great on the other situations.

On linux, use `ip address show dev docker0` to get the IP address, then add the address to `extra hosts`.

Note that this is a bit awkward, and a throwaway address is used on non-linux. This is because golang templating by default escapes the required quotes, breaking yaml :(  There are workarounds for it, but it's a bit of work (You have to treat things as HTML attributes instead of as strings in the golang templating.)

xdebug instructions should work for Linux, Mac, and Windows as shown in [our docs](https://ddev.readthedocs.io/en/latest/users/step-debugging/)

I haven't added a test for this.

## Related Issue Link(s):

The problem was introduced in https://github.com/drud/ddev/pull/785

## Release/Deployment notes:

- [ ] Update Stack Overflow article with Linux fix on release.
- [ ] When Docker for Linux is fixed, remove this code.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -396,8 +396,8 @@ func (app *DdevApp) CheckCustomConfig() {
 func (app *DdevApp) RenderComposeYAML() (string, error) {
 	var doc bytes.Buffer
 	var err error
-	var docker0Addr string = "127.0.0.1"
-	var docker0Hostname string = "unneeded"
+	var docker0Addr = "127.0.0.1"
+	var docker0Hostname = "unneeded"
 	templ := template.New("compose template")
 	templ, err = templ.Parse(DDevComposeTemplate)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


rickmanelius:
I'd actually be curious if @andrewfrench could take the first pass of a manual test on this in addition to code/test coverage review. Given that this is a OS specific regression, if he can confirm it's working then that would be sufficient for me to skip this review.

andrewfrench:
Agreed, working towards that right now @rickmanelius

rfay:
It's actually a great project for that: Experience with PHP debugging, getting linux review going, all kinds of good stuff.

PR #890, opened 2018-05-29
Include direct web container access in GetAllURLs(), resolves #796

### LINKED ISSUE #796: "ddev describe" should show direct ephemeral port/url via localhost

`docker ps` shows the localhost port that nginx is listening on, which can often be useful when things go wrong. For example, in the example below, you can easily hit the webserver at http://localhost:32795.

We should show the direct/localhost URL in `ddev describe`.

`76d9bc4f6cb8        drud/nginx-php-fpm-local:v1.2.2           "/start.sh"              26 hours ago        Up 26 hours (healthy)   0.0.0.0:32795->80/tcp, 0.0.0.0:32794->8025/tcp                                             ddev-d8git-web`

andrewfrench:
## The Problem/Issue/Bug:
In order to access a project's web container directly, a user would need to know how to obtain the container's ephemeral port via a command like `docker ps`.

This PR adds code that queries the project's web container for its public ephemeral port and local IP, returning this direct address at the end of the URLs list presented in `ddev describe`. A user can find this address without any Docker knowledge and access this address to interact directly with the web container, bypassing the ddev router.

- `ddev start` a project 
- Execute `ddev describe` within the project
- The last address in the URLs list under Project Information should be in the format `http://<ip>:<port>`
- Visit this direct address and interact with the site, bypassing the ddev router

A basic test (`TestGetAllURLs()`) was added to ensure the direct web container address is returned.

## Related Issue Link(s):
#796

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -950,7 +950,7 @@ func (app *DdevApp) GetAllURLs() []string {
 	} else {
 		for _, p := range webContainer.Ports {
 			if p.PrivatePort == 80 {
-				URLs = append(URLs, fmt.Sprintf("http://%s:%d", "127.0.0.1", p.PublicPort))
+				URLs = append(URLs, fmt.Sprintf("http://%s:%d", p.IP, p.PublicPort))
 				break
 			}
 		}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -946,13 +946,9 @@ func (app *DdevApp) GetAllURLs() []string {
 
 	webContainer, err := app.FindContainerByType("web")
 	if err != nil {
-		util.Error("Unable to find web container for app %s: %s", app.Name, err)
+		util.Error("Unable to find web container for app: %s, err %s", app.Name, err)
 	} else {
 		for _, p := range webContainer.Ports {
-			// TODO: Remove
-			fmt.Printf("%s %s %d %d\n", p.Type, p.IP, p.PublicPort, p.PrivatePort)
-
-			// TODO: Always private port 80?
 			if p.PrivatePort == 80 {
 				URLs = append(URLs, fmt.Sprintf("http://%s:%d", p.IP, p.PublicPort))
 			}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -950,7 +950,7 @@ func (app *DdevApp) GetAllURLs() []string {
 	} else {
 		for _, p := range webContainer.Ports {
 			if p.PrivatePort == 80 {
-				URLs = append(URLs, fmt.Sprintf("http://%s:%d", p.IP, p.PublicPort))
+				URLs = append(URLs, fmt.Sprintf("http://%s:%d", "127.0.0.1", p.PublicPort))
 				break
 			}
 		}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -975,7 +975,7 @@ func (app *DdevApp) HostName() string {
 func (app *DdevApp) AddHostsEntries() error {
 	dockerIP, err := dockerutil.GetDockerIP()
 	if err != nil {
-		return fmt.Errorf("could not get Docker IP: %s", err)
+		return fmt.Errorf("could not get Docker IP: %v", err)
 	}
 
 	hosts, err := goodhosts.NewHosts()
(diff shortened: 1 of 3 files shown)


rfay:
@rickmanelius we can't have https for the direct-to-container URL because HTTPS is terminated on the router and we're avoiding the router.

@andrewfrench I don't think 0.0.0.0 is a correct approach, should be 127.0.0.1.

rfay:
Hmm, I note that 0.0.0.0 is apparently what's in p.IP and that it also works correctly. Fascinating. I've seen 0.0.0.0 used for binding to all interfaces, but never seen it used this way.

rfay:
Fascinating, https://superuser.com/a/949522/75275 - I guess this should be just fine. It needs testing on Windows and Linux though.

Also, despite your super-fast response, it's not correct to set it to 127.0.0.1 in all cases, because on Docker toolbox it may be another address.

andrewfrench:
@rfay: Reverted the IP change back to using `p.IP`, and I'll wait for a go-ahead before pulling.

Re: test -- good point! I don't currently see a test for GetAllURLs, adding one now.

rickmanelius:
@rfay As a tactical point of order, I know that Docker Toolbox is a separate PR that hasn't been accepted yet. Does it looks like there is a relatively clear path forward to making them both work? FWIW, if Docker Toolbox support ultimately breaks certain functionality, that might be part of the consideration on whether or not to officially support it at all.

rfay:
@rickmanelius yes, we already do the correct IP lookup in the linked code, so there's no problem making this work.

rickmanelius:
Ok! I saw some additional commits and this is still in the Needs Review column. Not sure if it's ready, but I may retest tonight.

rfay:
Please rebase this against upstream/master to pick up the buildkite testing and such, thanks! Or if you're around tonight or in the morning, authorize me to do that.

rickmanelius:
Ah yes I forgot about that. I just took it for a quick spin and I noted that the proper IP address came through. However, I've already tested and if it works after the rebaase, then feel free to pull it. Thanks!

pkg/ddevapp/ddevapp.go
```diff
@@ -939,6 +942,26 @@ func (app *DdevApp) GetAllURLs() []string {
 		}
 		URLs = append(URLs, "http://"+name+httpPort, "https://"+name+httpsPort)
 	}
+
+	dockerIP, err := dockerutil.GetDockerIP()
+	if err != nil {
+		util.Error("Unable to get Docker IP: %s", err)
```
rfay: This would be better as a %v, but I'm concerned that we're not doing the right thing by continuing. Probably the signature of this function is wrong, should have an error. But it's probably OK anyway.

pkg/ddevapp/ddevapp.go
```diff
@@ -939,6 +942,26 @@ func (app *DdevApp) GetAllURLs() []string {
 		}
 		URLs = append(URLs, "http://"+name+httpPort, "https://"+name+httpsPort)
 	}
+
+	dockerIP, err := dockerutil.GetDockerIP()
```
rfay: Thanks so much for refactoring this correctly. Takes longer to do and is always worth it in a case like this.

pkg/ddevapp/ddevapp.go
```diff
@@ -949,15 +972,11 @@ func (app *DdevApp) HostName() string {
...
-	dockerHostRawURL := os.Getenv("DOCKER_HOST")
-	if dockerHostRawURL != "" {
-		dockerHostURL, err := url.Parse(dockerHostRawURL)
-		if err != nil {
-			return fmt.Errorf("failed to parse $DOCKER_HOST: %v, err: %v", dockerHostRawURL, err)
-		}
-		dockerIP = dockerHostURL.Hostname()
+	dockerIP, err := dockerutil.GetDockerIP()
+	if err != nil {
+		return fmt.Errorf("could not get Docker IP: %s", err)
```
rfay: Please use %v for err

andrewfrench:
Holy cow, this rebase trainwreck is a good argument against Git after dark.  In an effort to avoid more Git spaghetti, I've recreated this PR at #892, addressing all outstanding comments and adding a basic automated test case for `GetAllURLs()`.

rickmanelius:
Needs retesting after the rebase or are we good to go?

rfay:
Andrew has one tiny thing he wants to do with it, and add some possible test coverage. But it should be OK. If you want to just include the testing in your release test it would probably be good enough.

andrewfrench:
This one's good to go pending ✅ s from Circle and Buildkite.

PR #893, opened 2018-05-30
If a ddev-local specific config file exists, include it.

cweagans:
## The Problem/Issue/Bug:

We need a place where users can reliably put ddev-local related configuration for their applications.

If settings.ddev-local.php, wp-config-ddev-local.php, or LocalConfigurationDdev.php exist (for Backdrop/Drupal, Wordpress, and Typo3 respectively), we'll require it.

Spin up a vanilla site, create one of the files mentioned above and put something simple in it (for instance, just `<?php die('file was included');` should verify that the file is included properly), and then hit any page on the site.

This should only happen if ddev generated the config file.

### DIFF
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -60,6 +60,11 @@ ini_set('session.gc_probability', 1);
 ini_set('session.gc_divisor', 100);
 ini_set('session.gc_maxlifetime', 200000);
 ini_set('session.cookie_lifetime', 2000000);
+
+// Include local settings if it exists.
+if (file_exists(__DIR__ . '/settings.ddev-local.php')) {
+	require_once __DIR__ . '/settings.ddev-local.php';
+}
 `
 
 // createBackdropSettingsFile creates the app's settings.php or equivalent,
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -95,6 +95,11 @@ $settings['file_scan_ignore_directories'] = [
 if (!empty($_SERVER["argv"]) && strpos($_SERVER["argv"][0], "drush") && empty($_ENV['DEPLOY_NAME'])) {
   include __DIR__ . '../../../drush.settings.php';
 }
+
+// Include local settings if it exists.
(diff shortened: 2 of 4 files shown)


rfay:
Our current thinking was to append the include to an existing settings file if it exists. We don't think that approach will probably be required for TYPO3.

cweagans:
Yeah. We talked about this in a zoom this morning. We didn't feel like editing existing config files was appropriate in the short term, so the MVP was to start with a `require_once` in settings file that we control.

rfay:
I don't agree with that at this point; I don't understand how this prioritized work was started without a discussion, even though it was assigned to Andrew.

cweagans:
That was the takeaway from the call with @rickmanelius, myself, @alkymst, and @cyberswat. It was all of about 10 minutes of work. The only thing that was assigned to Andrew was a review.

Please let me know how to proceed. This is an annoyance for a client migration and merging this will resolve it in a way that doesn't negatively impact anyone.

rickmanelius:
Please, let's pause on this. There is a considerable amount of discussions happening in parallel and this is not a critical feature in the current v0.19.0 release and in the roadmap. So I would like to take our foot off the gas on this. I recognize that this has many contexts and I did in fact meet with @cyberswat @alkymst and @cweagans to get ahead of this because I saw the conversation moving rapidly without me or @rfay (primary leads on this) getting to have input before it was to a PR stage.

When I'm at a place where I can take a breath and prepare for the v0.19.0 release, I'll get a more thorough response to this. Thank you for your patience. I see there is a lot of good intent and desire in here... but we're colliding a bit and that can have secondary, negative effects.

rfay:
This effort is very much appreciated, and we should have done just this ages ago. However, as overall settings work is going on, I'd like to close this and then incorporate the ideas in @andrewfrench 's upcoming settings work. I *think* we're going to go ahead and add an include for ddev settings file in unmanaged settings files. Anyway, this was brilliant and appreciated, and the ideas will get in this summer.

dclear:
we're going to the roll the work done here into consideration for this week's sprint (which has some related WP issues)

PR #895, opened 2018-05-30
Remove max_allowed_packet override, fixes #894

### LINKED ISSUE #894: Unable to override mysql max_allowed_packet setting
**Describe the bug**
I am following instructions on https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/ to override mysql settings using a custom `.cnf` file. I am able to override some settings, but I'm unable to override `max_allowed_packet` specifically, even when other settings in the override file are being applied successfully.

**To Reproduce**
1. Create a ddev environment
2. Login to mysql (`ddev ssh` followed by `mysql -u root -proot -h db`)
3. Check settings for `max_allowed_packet` (and other settings to verify they are later overridden -- for example, both `tmp_table_size` and `max_heap_table_size` default to `64M`, those will be overridden as a test).
```
mysql> show global variables like 'max_allowed_packet';
+--------------------+----------+
| Variable_name      | Value    |
+--------------------+----------+
| max_allowed_packet | 16777216 |
+--------------------+----------+
1 row in set (0.01 sec)

mysql> show global variables like 'tmp_table_size';
+----------------+----------+
| Variable_name  | Value    |
+----------------+----------+
| tmp_table_size | 67108864 |
+----------------+----------+
1 row in set (0.00 sec)

mysql> show global variables like 'max_heap_table_size';
+---------------------+----------+
| Variable_name       | Value    |
+---------------------+----------+
| max_heap_table_size | 67108864 |
+---------------------+----------+
1 row in set (0.01 sec)
```
4. Add a custom cnf file within `.ddev/mysql/foo.cnf` with the following contents:
```
[mysqld]
max_allowed_packet=32M
tmp_table_size=128M
max_heap_table_size=128M
```
5. Run `ddev restart`, verify the new cnf file is being loaded with the following output: `Using custom mysql configuration: [foo.cnf]`.
6. Login to mysql again and check settings for the above overridden variables. While some will be overridden successfully, `max_allowed_packet` remains at the default `16M`:
```
mysql> show global variables like 'max_allowed_packet';
+--------------------+----------+
| Variable_name      | Value    |
+--------------------+----------+
| max_allowed_packet | 16777216 |
+--------------------+----------+
1 row in set (0.00 sec)

mysql> show global variables like 'tmp_table_size';
+----------------+-----------+
| Variable_name  | Value     |
+----------------+-----------+
| tmp_table_size | 134217728 |
+----------------+-----------+
1 row in set (0.00 sec)

mysql> show global variables like 'max_heap_table_size';
+---------------------+-----------+
| Variable_name       | Value     |
+---------------------+-----------+
| max_heap_table_size | 134217728 |
+---------------------+-----------+
1 row in set (0.00 sec)
```

**Expected behavior**
`max_allowed_packet` should be overridden with the value from `foo.cnf`.

 - mac OS 10.13.4
 - Docker version 18.03.1-ce
 - ddev version:
```
db    	drud/mariadb-local:v0.9.0
dba   	drud/phpmyadmin:v0.2.0
router	drud/ddev-router:v0.5.1
commit	v0.18.0
domain	ddev.local
cli   	v0.18.0
web   	drud/nginx-php-fpm-local:v1.3.0
```
 - config.yaml contents for the misbehaving project:
```
APIVersion: v0.18.0
name: foo
type: drupal7
docroot: docroot
php_version: "5.6"
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
provider: default


# This config.yaml was created with ddev version v0.18.0
# webimage: drud/nginx-php-fpm-local:v1.3.0
# dbimage: drud/mariadb-local:v0.9.0
# dbaimage: drud/phpmyadmin:v0.2.0
# However we do not recommend explicitly wiring these images into the
# config.yaml as they may break future versions of ddev.
# You can update this config.yaml using 'ddev config'.

# Key features of ddev's config.yaml:

#   http://projectname.ddev.local and https://projectname.ddev.local

# type: <projecttype>  # drupal6/7/8, backdrop, typo3, wordpress, php

# docroot: <relative_path> # Relative path to the directory containing index.php.

# php_version: "7.1"  # PHP version to use, "5.6", "7.0", "7.1", "7.2"

# You can explicitly specify the webimage, dbimage, dbaimage lines but this
# is not recommended, as the images are often closely tied to ddev's' behavior,
# so this can break upgrades.

# webimage: <docker_image>  # nginx/php docker image.
# dbimage: <docker_image>  # mariadb docker image.
# dbaimage: <docker_image>

# router_http_port: <port>  # Port to be used for http (defaults to port 80)
# router_https_port: <port> # Port for https (defaults to 443)

# xdebug_enabled: false  # Set to true to enable xdebug and "ddev start" or "ddev restart"

additional_hostnames:
# - someothername
# would provide http and https URLs for "somename.ddev.local"
# and "someothername.ddev.local".

# provider: default # Currently either "default" or "pantheon"
#
# Many ddev commands can be extended to run tasks after the ddev command is
# executed.
# See https://ddev.readthedocs.io/en/latest/users/extending-commands/ for more
# information on the commands that can be extended and the tasks you can define
# for them. Example:
#hooks:
#  post-import-db:
#    - exec: "drush cc all"

hooks:
  post-start:
    # Enable ping within Solr so that Drupal acknowledges it is up and running.
    - exec: "curl -s http://foo.ddev.local:8983/solr/dev/admin/ping?action=enable -o /dev/null"
```

rfay:
## The Problem/Issue/Bug:

A long-ago feature of the mariadb_local container had an override of max_allowed_packet on the command line when running the container. In the meantime we've allowed custom config, and OP #894 points out that it's impossible to override max_allowed_packet.

* Remove the command line override.
* Update default max-allowed-packet to better 256M value per discussion in https://github.com/drud/ddev/issues/894#issuecomment-393286945

* This is pushed as drud/mariadb-local:20180530_remove_max_allowed_packet
* Use `dbimage: drud/mariadb-local:20180530_remove_max_allowed_packet` in a project's config.yaml to use it. 
* Verify that max_allowed_packet defaults to 256M
* Override the max_allowed_packet as described in #894 and verify that behavior is now correct.

## Related Issue Link(s):

OP #894 

## Release/Deployment notes:

This needs to be added to the version.go before release, so a commit needs to be added updating version.go

This is probably OK to add go v0.19.0, but OTOH nobody has discovered it until Jeff, so there's probably no pressure.


rfay:
This could be a candidate for v0.19.0, a very small change. Or we can wait for next time.

rickmanelius:
I don't see this as urgent. That said, if I can squeeze it in during the release and it's a 👍 it looks benign enough.

rfay:
This has been rebased to be standalone and is referenced in the version.go, so the built executable can be used as is. the latest intermediate container has been pushed.

rfay:
This failed on win10pro because docker hung. I'm fixing the testbot, but think we can pull it.

PR #897, opened 2018-05-30
Bump container versions to planned v0.19.0 for release

rfay:
## The Problem/Issue/Bug:

Our container code is now part of the main code tree, so during development we used temporary pushed container.

Now, though, we're about to do a release with the code in our containers. But we want to be able to test and reference the new containers (all tagged v0.19.0 on dockerhub) at the time of the tag.

To do that we have to temporarily push v0.19.0 containers, so I've just done that with the current code.

Use it, look at the tests.

## Release/Deployment notes:

The 4 containers should be re-pushed at tag time.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -20,25 +20,25 @@ var DockerComposeVersionConstraint = ">= 1.10.0-alpha1"
 var WebImg = "drud/nginx-php-fpm-local"
 
 // WebTag defines the default web image tag for drud dev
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for taking the extra effort for the manual review @andrewfrench !

rfay:
I'm having a bit of trouble with the testbots, but it doesn't seem to have anything to do with this PR. So as soon as both are up again, I'll pull this and we'll rely on the master test.

PR #913, opened 2018-06-07
Fix broken containers resulting from CRs in scripts

rfay:
## The Problem/Issue/Bug:

In https://github.com/drud/ddev/pull/902 I deliberately let slide test failures on Windows of the containers. This is to fix that. Hopefully getting buildkite to check out as-on-server, with linefeeds and no CRs will solve that problem.  

I *think* the largest part of the problem was that Windows agents have to check out at least the container stuff with core.autocrlf=false, so we have shell scripts that don't have CRs in them. This meant a change in the Windows *agent* testbots:

`git-clone-flags="-v --config core.autocrlf=false"`

And that was the most important change.

### DIFF
--- a/.buildkite/test_containers.sh
+++ b/.buildkite/test_containers.sh
@@ -7,6 +7,14 @@ set -o pipefail
 set -o nounset
 set -x
 
+# Make sure we don't have any existing containers on the testbot that might
+# result in this container not being built from scratch.
+VERSION=$(make version | sed 's/^VERSION://')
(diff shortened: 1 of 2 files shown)


rfay:
[ddev-containers-windows](https://buildkite.com/drud/ddev-containers-windows/builds/32) was clean on this one, not sure why buidkite and thus github report it as a failure. Retrying ddev-containers-macos.

PR #914, opened 2018-06-08
Improve reliability with a more unique tmpdir and outdir

rfay:
## The Problem/Issue/Bug:

Working on reliability issues with the container build I noted possible collisions of the directory we're using, so added $SECONDS to the tmp directory name maker. We'd be better to use TMPDIR, but that provides an unusable directory on macOS.

Improve tmp directory name; should not do any harm.

### DIFF
--- a/containers/mariadb-local/test/testserver.sh
+++ b/containers/mariadb-local/test/testserver.sh
@@ -6,7 +6,7 @@ IMAGE="$1"  # Full image name with tag
 MYSQL_VERSION="$2"
 CONTAINER_NAME="testserver"
 HOSTPORT=33000
-MYTMPDIR=~/tmp/testserver-sh_$$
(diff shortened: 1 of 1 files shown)


containers/mariadb-local/test/testserver.sh
```diff
@@ -6,7 +6,7 @@ IMAGE="$1"  # Full image name with tag
 MYSQL_VERSION="$2"
 CONTAINER_NAME="testserver"
 HOSTPORT=33000
-MYTMPDIR=~/tmp/testserver-sh_$$
+MYTMPDIR=~"/tmp/testserver-sh_${SECONDS}_$$"
```
andrewfrench: If the goal is uniqueness, would it be better to use `$RANDOM` over `$SECONDS` here?

PR #922, opened 2018-06-14
Allow a user to issue root commands from ddev, fixes #918

### LINKED ISSUE #918: Allow a user to issue root commands from ddev
We provide users a way to execute arbitrary commands on a container, but we offer no way for a user to execute a command with root privileges.  A user can still do this, but it requires a user to work around `ddev` (instead of with it), requires some specific Docker knowledge, and can't be integrated with `ddev`'s native tooling.

`ddev exec` already handles issuing commands within a container -- we can add an optional flag indicating the command should be executed as root:

`ddev exec -R <command>`

It'd be a good idea to alias `ddev exec -R` to something like `ddev rexec` so a root command can be issued in an event hook from `config.yml`:
```
hooks:
  post-start:
    rexec: <command_1>
    rexec: <command_2>
```

**Additional context**

andrewfrench:
## The Problem/Issue/Bug:
Users are currently unable to issue commands within a container as root from ddev. 

This PR adds the `--root/-R` flag to `ddev exec`, which executes the commands as the root user. A new hook command, `exec-root`, allows a user to define a hook command to be run as root, allowing for automated modifications of the standard container. Documentation on hooks and commands has been updated.

With a running ddev project, execute a command that requires root privileges with `ddev exec`. For example, `ddev exec apt-get update` should fail due to lack of permissions.  Execute the same command with root flags: `ddev exec --root apt-get update`.  The command will complete successfully.

Simple test cases that confirm the `--root` flag provides a root user have been added.

## Related Issue Link(s):
#918

### DIFF
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -11,12 +11,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var root bool
+
 // DdevExecCmd allows users to execute arbitrary bash commands within a container.
 var DdevExecCmd = &cobra.Command{
 	Use:     "exec <command>",
 	Aliases: []string{"."},
 	Short:   "Execute a shell command in the container for a service. Uses the web service by default.",
-	Long:    `Execute a shell command in the container for a service. Uses the web service by default. To run your command in the container for another service, run "ddev exec --service <service> <cmd>"`,
+	Long:    `Execute a shell command in the container for a service. Uses the web service by default. To run your command in the container for another service, run "ddev exec --service <service> <cmd>". To run your command as root, run "ddev exec --root <cmd>".`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) == 0 {
 			err := cmd.Usage()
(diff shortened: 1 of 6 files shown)


rickmanelius:
@andrewfrench Thanks for getting this PR in place! One question before a review by me/@rfay would be any issues related to the docker toolbox test failing.

andrewfrench:
It looks like either Randy's test configuration changes stabilized the Buildkite tests or it's a Monday miracle, but all the tests are giving us legitimate greens. I think this one's good for manual review.

andrewfrench:
That change to the container is an elegant solution that I'd support, but doesn't address the entirety of this PR.  Either way, a new hook command is needed to allow a command to be executed either with sudo or as root.

I don't have a ton of background information on this issue, but it sounded like there was a previous change that removed root/sudo access by default from the `ddev ssh` user &mdash; I want to make sure I'm not undoing intentional changes without understanding the reasoning.

rfay:
Wouldn't the current exec hook just work, with people using sudo in the command?

rfay:
As far as the background of running containers as the ordinary users, there are two major reasons:

1.  host filesystem security on Linux. On Linux, the docker filesystem mount is native, meaning that a root user in the container can do amazingly nasty things on the host filesystem. 
2. It also can (and has) resulted in file ownership that is just inconvenient.  If the webserver creates files with ownership of the www-data user, that comes through to the host with ownership that is really difficult to deal with, especially for more naive users.

andrewfrench:
> Wouldn't the current exec hook just work, with people using sudo in the command?

Whoops, it would. I was imagining the sudo having to somehow come before the exec.

rfay:
We've tentatively decided to go with the approach in https://github.com/drud/ddev/pull/935, so closing this one for now. If we think we should go back this way, we'll reopen.

Issue #924, opened 2018-06-15
Mailhog doesn't work with TYPO3 ddev configuration


spoonerWeb:
**Describe the bug**
TYPO3 in its basic configuration cannot send test mails due to an error in the sendmail command: `/usr/local/bin/mailhog sendmail test@example.org --smtp-addr 127.0.0.1:1025`

**To Reproduce**
1. Setup TYPO3
2. Go to Install Tool (`/typo3/install.php`)
3. Go to module `Environment`
4. Go to `Test Mail Setup`
5. Send a test mail
6. Error `Something went wrong`, in the TYPO3 log `Uncaught TYPO3 Exception: Unsupported sendmail command flags [/usr/local/bin/mailhog sendmail test@example.org --smtp-addr 127.0.0.1:1025]. Must be one of "-bs" or "-t" but can include additional flags.`

**Expected behavior**
It should send an email.

**Screenshots**

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: macOS High Sierra
 - Docker version information (use `docker version`) and insert it here.
```
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true
```
 - ddev version information (use `ddev version`)
```
commit	v0.19.0                         
domain	ddev.local                      
cli   	v0.19.0                         
web   	drud/nginx-php-fpm-local:v0.19.0
db    	drud/mariadb-local:v0.19.0      
dba   	drud/phpmyadmin:v0.19.0         
router	drud/ddev-router:v0.19.0
```
 - config.yaml contents for the misbehaving project
```
APIVersion: v0.19.0
name: typo3-9
type: typo3
docroot: public
php_version: "7.2"
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
provider: default
```
 - Do you have any custom configuration (nginx, php, mysql) in the .ddev folder? If so, have you tried without them?

rickmanelius:
[...]

PR #925, opened 2018-06-15
Make TYPO3 work with mailhog

spoonerWeb:
## The Problem/Issue/Bug:
Wrong sendmail command to send mails in TYPO3 environment

Add the missing parameter

See #924 

There are no tests for this, are they?

## Related Issue Link(s):
#924 

## Release/Deployment notes:
Not afaik

### DIFF
--- a/containers/nginx-php-fpm-local/files/etc/php/5.6/fpm/php.ini
+++ b/containers/nginx-php-fpm-local/files/etc/php/5.6/fpm/php.ini
@@ -81,7 +81,7 @@ pdo_mysql.default_socket=
 SMTP = localhost
 smtp_port = 25
 mail.add_x_header = On
-sendmail_path="/usr/local/bin/mailhog sendmail test@example.org --smtp-addr 127.0.0.1:1025"
+sendmail_path="/usr/local/bin/mailhog sendmail -t test@example.org --smtp-addr 127.0.0.1:1025"
 
 [SQL]
 sql.safe_mode = Off
--- a/containers/nginx-php-fpm-local/files/etc/php/7.0/fpm/php.ini
+++ b/containers/nginx-php-fpm-local/files/etc/php/7.0/fpm/php.ini
@@ -75,7 +75,7 @@ pdo_mysql.default_socket=
 SMTP = localhost
 smtp_port = 25
 mail.add_x_header = On
-sendmail_path="/usr/local/bin/mailhog sendmail test@example.org --smtp-addr 127.0.0.1:1025"
+sendmail_path="/usr/local/bin/mailhog sendmail -t test@example.org --smtp-addr 127.0.0.1:1025"
 
 [SQL]
 sql.safe_mode = Off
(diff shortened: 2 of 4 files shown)


rickmanelius:
If the tests pass, I say we pull this in.

rfay:
This needs to be linked to mailhog documentation that confirms the usage before it can be pulled. The previous usage was correct for years and years, so I assume that something has shifted in mailhog.

rfay:
I don't think we need to get this into v0.20.0, and there hasn't been any followup, but I'm pretty sure the appropriate fix is to set `transport => 'mail' ` in the AdditionalSettings.php. For TYPO3 users the immediate approach is just to set the transport in TYPO3 settings

rfay:
I think the path forward is to open another PR that does:

* transport => 'mail'
* `SYS/displayErrors => 1`
*  `SYS/devIPmask => '*'

psychomieze:
Especially if we want to make this compatible with as many systems (read: Windows) as possible, we should probably switch to the following configuration for TYPO3 (and not touch the php.ini files at all):

```
    'MAIL' => [
        'transport' => 'smtp',
        'transport_smtp_server' => 'localhost:32769',
    ],
```
with the host and port of the mailhog container. This should be the most reliable option.

rfay:
Just to be clear: None of this is related to or even has anything to do with WIndows. This all happens inside the web container, which is Debian Linux. If we were going to use SMTP, the mailhog smtp port is port 1025, and if that seems better to you, we could probably do it. But then any change to the mailhog config would break TYPO3 projects. As it is, the configuration is encapsulated in sendmail_path so it should work for any CMS that is using mail()

rfay:
Very thoughtful response from @jigal in [slack](https://typo3.slack.com/archives/C8TRNQ601/p1530002562000405):

> I did some digging in mail handling in PHP and how we can use this in TYPO3.
TYPO3 uses SwiftMailer to send emails. SwiftMailer has deprecated the use of the PHP `mail()` function as mail transport and can remove support completely at any moment. Therefore we have to switch to either "sendmail" or "SMTP".
According to the PHP documentation `mail()` uses either sendmail (or something compatible) or SMTP (only on Windows and only if `sendmail_path` is empty). This is configured in the php.ini settings. 
> TYPO3 can use the php.ini settings (at least for new installations) to configure the mail settings as most installations have a working `mail()` function. If sendmail_path is empty (and SMTP is empty on Windows) we can only ask the user how installs TYPO3.
So, please don't rely on the 'mail' transport, but make sure the php.ini setting `sendmail_path` can be copied to the TYPO3 mail settings and SwiftMailer can use that command. If someone prefers to use SMTP they can configure this themselves IMO.

rfay:
Thanks so much for bringing this to the fore and taking the initiative to do a PR @spoonerWeb - Per the discussion here, the php.ini technique is not what's needed, so I'm going to close this one out. I think the key info is in https://github.com/drud/ddev/issues/924, and I hope to get that into the next release. A followup PR from you is welcome if you want, probably best setting the transport to SMTP and port 25 in the AdditionalSettings.php

Issue #269, opened 2017-05-24
Create a service configuration for memcache


tannerjfco:
## What happened (or feature request):
Feature Request
In #121 we determined memcache was one of several popular services that can be a requirement for more complex applications. In order for ddev to facilitate local environments for these more complex applications, we should provide an optional service configuration for memcache.

## What you expected to happen:
A docker-compose configuration should be made, named `docker-compose.redis.yml`. This configuration should:

- Determine if an existing memcache docker image can be used, or if we need to develop / maintain our own.
- Define a service named memcache that provides a memcache container.
- Where possible, pre-configure the memcache container to reduce required setup for a user.
- Document any steps that are necessary after adding the service file to .ddev to be able to use memcache with a Drupal site.

The [example solr service in documentation](https://ddev.readthedocs.io/en/latest/users/customization-extendibility/#defining-an-additonal-service-apache-solr-example) is a good reference for how this service file should work.

## Related source links or issues:

rickmanelius:
Closing for now until this becomes an end-user request, at which time we can re-open.

rickmanelius:
Hi @jeffsheltren! This looks great and I agree with the approach of installing the PHP extensions would make sense. Long term, I want to include a better ability for users to add these additional docker-compose.service.yml files in a programmatic way, but I think we could easily start by adding another example in our documentation.

Summary. I think this is actionable with the PHP extensions installed and the above example file added to our docs. @jeffsheltren if you can add a PR, we'll get it reviewed and in. And as someone not very experienced with memcached, if you could give me a few ways to quickly test and assert its functional I would appreciate it.

jeffsheltren:
I submitted #923 with the change to the PHP container. I realized that adding both php-memcache and php-memcached extensions added a lot of complexity to enable one but not the other, so I ended up only installing php-memcached -- since php-memcache appears unmaintained at this point.

For testing memcached itself, I would test to ensure that the PHP container can reach the memcached container on port 11211. If you can connect on that port via `telnet` or `nc`, typing `stats` and seeing stats returned should be enough to verify memcached is up and running.

dclear:
thanks @jeffsheltren !

rfay:
This was auto-closed, but if I'm not mistaken, #923 does not actually fix it right, just enables it to be fixed.

andrewfrench:
Closing, this was resolved in #927. 👍

PR #927, opened 2018-06-17
Add Memcached service configuration file and documentation.

jeffsheltren:
## The Problem/Issue/Bug:
Currently ddev does not provide a Memcached container. Memcached is used as a caching layer by many applications.

Adds an example docker-compose file and documentation to setup a Memcached container within a ddev environment.

Follow instructions in documentation included in PR to create a Memcached container.

This should get picked up automatically by the already-existing servicetest test script.

## Related Issue Link(s):
#269

### DIFF
--- a/pkg/servicetest/servicetest_test.go
+++ b/pkg/servicetest/servicetest_test.go
@@ -7,12 +7,17 @@ import (
 
 	"path/filepath"
 
+	"net"
+
+	"fmt"
+
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/dockerutil"
 	"github.com/drud/ddev/pkg/fileutil"
 	"github.com/drud/ddev/pkg/output"
 	"github.com/drud/ddev/pkg/testcommon"
 	"github.com/drud/ddev/pkg/util"
+	"github.com/fsouza/go-dockerclient"
 	log "github.com/sirupsen/logrus"
 	asrt "github.com/stretchr/testify/assert"
 )
@@ -27,9 +32,26 @@ var (
 			Type:    "drupal8",
 		},
 	}
-	ServiceFiles []string
-	ServiceDir   string
+	ServiceFiles   []string
+	ServiceDir     string
+	serviceConfigs = map[string]serviceTestConfig{
+		"solr": {
+			proto: "http",
+			port:  8983,
+			path:  "/solr/",
+		},
+		"memcached": {
+			proto: "tcp",
+			port:  11211,
+		},
+	}
 )
+
+type serviceTestConfig struct {
+	proto string
+	port  int64
+	path  string
+}
 
 // TestMain runs the tests in servicetest
 func TestMain(m *testing.M) {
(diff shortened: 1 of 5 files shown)


pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -0,0 +1,30 @@
...
+    image: memcached:1.5
+    restart: always
+    ports:
+      - 11211 # memcached is available at this port inside the container
+    labels:
+    # These labels ensure this service is discoverable by ddev
+      com.ddev.site-name: ${DDEV_SITENAME}
+      com.ddev.approot: $DDEV_APPROOT
+      com.ddev.app-url: $DDEV_URL
+    environment:
```
rfay: I'm pretty sure the entire environment stanza is not useful; we don't support access via http on named project via port 11211 because 11211 is a tcp interface, not http. So ddev-router is not involved at all.  (I tested and it works fine without this section)

pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -0,0 +1,30 @@
...
+  memcached: # This is the service name used when running ddev commands accepting the --service flag
+    container_name: ddev-${DDEV_SITENAME}-memcached # This is the name of the container. It is recommended to follow the same name convention used in the main docker-compose.yml file.
+    image: memcached:1.5
+    restart: always
+    ports:
+      - 11211 # memcached is available at this port inside the container
+    labels:
+    # These labels ensure this service is discoverable by ddev
+      com.ddev.site-name: ${DDEV_SITENAME}
+      com.ddev.approot: $DDEV_APPROOT
```
rfay: The "labels" section is also not useful since site-name, app-root, and app-url aren't used here. The commend about "Discoverable by ddev" is incorrect, I think, I know it's from the example.

pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -0,0 +1,30 @@
...
+      com.ddev.app-url: $DDEV_URL
+    environment:
+      - VIRTUAL_HOST=$DDEV_HOSTNAME # This defines the host name the service should be accessible from. This will be sitename.ddev.local
+      - HTTP_EXPOSE=11211
+
+    command: ["-m", "128"]
+
+# This links the memcached service to the web service defined in the main docker-compose.yml, allowing applications running in the web service to access the memcached service at memcached:11211
```
rfay: The 'web' stanza is also not useful. Pretty sure it should be removed.

rfay:
Hoping to get this in before it's too stale, know you're busy. Would love to have a rebase of it and get it going again. I don't think it will take too much to get it in. Let us know if you want us to run with it instead (and you don't mind commits being pushed into your fork).

rfay:
Moving to v1.1.0

jeffsheltren:
Thanks for the feedback @rfay! I believe I addressed everything in the docker-compose file with the commit today. I'd love your help with adding a test for it. I believe we can simply test a TCP connection to port 11211, but I'm not sure how to do so. Feel free to push to my fork if that's easiest.

rfay:
Could you rebase this? Or mind if I rebase and push back into your fork?

pkg/servicetest/servicetest_test.go
```diff
@@ -136,3 +153,46 @@ func TestServices(t *testing.T) {
...
 }
+
+// checkHTTP will attempt to communicate with a service over HTTP
+func checkHTTP(t *testing.T, container docker.APIContainers, targetPort int64, path string) {
+	t.Log("Checking HTTP communication")
+	assert := asrt.New(t)
+
+	for _, port := range container.Ports {
+		if port.PrivatePort == targetPort && port.PublicPort != 0 {
+			address := fmt.Sprintf("http://127.0.0.1:%d%s", port.PublicPort, path)
```
rfay: This is a decent test, but I think it may be leaving out the important fact that we have to be able to access the port from *within* the web container. I imagine this approach is what we were doing before, but we should consider using app.Exec() to take an action within the web container. nc (netcat) is installed in there, it would do fine I think.

pkg/servicetest/servicetest_test.go
```diff
@@ -136,3 +153,46 @@ func TestServices(t *testing.T) {
...
+}
+
+// checkTCP will attempt to communicate with a service over TCP
+func checkTCP(t *testing.T, container docker.APIContainers, targetPort int64, path string) {
+	t.Log("Checking TCP communication")
+	assert := asrt.New(t)
+
+	for _, port := range container.Ports {
+		if port.PrivatePort == targetPort {
+			address := fmt.Sprintf("127.0.0.1:%d%s", port.PublicPort, path)
```
rfay: As with the other one, this tests from the host's perspective, but may not be an adequately valid test because it's not testing from the web container's perspective.

PR #929, opened 2018-06-17
Add Composer Setup Example for TYPO3

michadu:
With a composer example for TYPO3 you can setup a fresh installation.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -117,12 +117,23 @@ Your project can be reached at: http://my-drupal8-site.ddev.local
 
 To get started using ddev with a TYPO3 project, clone the project's repository and checkout its directory.
 
+**Git Clone Example**
+
 ```
 git clone https://github.com/example-user/example-typo3-site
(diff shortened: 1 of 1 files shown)


docs/users/cli-usage.md
```diff
@@ -117,13 +117,24 @@ Your project can be reached at: http://my-drupal8-site.ddev.local
...
 git clone https://github.com/example-user/example-typo3-site
 cd example-typo3-site
 ```
 
 If necessary, run build steps that you may require, like `composer install` in the correct directory.
 
+**Composer Setup Example**
+
+```
+composer create-project typo3/cms-base-distribution my-typo3-site ^9
```
rfay: The [official docs](https://docs.typo3.org/typo3cms/InstallationGuide/QuickInstall/Composer/Index.html) would have us use Stable 8.7, rather than 9. What do you think?

michadu: You are right, we should use ^8 until v9 is released in October.

PR #932, opened 2018-06-18
Update WordPress settings management, fixes #756, fixes #1156

### LINKED ISSUE #756: Wordpress: ddev overwrites user-managed wp-config.php

In a wordpress site, if you `ddev config` it will overwrite an existing user-managed wp-config.php.

Shouldn't do that.

## How to reproduce this:

* Create a wp-config.php any way you want. Make sure it doesn't have #ddev-generated in it.
* `ddev config`. It will be overwritten.

### LINKED ISSUE #1156: Wordpress projects don't need the warning to run wp to change URL
**Describe the bug**

Wordpress projects don't need the warning to run wp to change the URL. Instead, we should be adding 
define('WP_HOME', <site_url>);, and no changes will be required. 

So:
1. Remove that from post-import-db action (or remove the action if that's all that's in there)
2. Add WP_HOME setting to config-wp-ddev.php

alkymst:
## The Problem/Issue/Bug:
#756 
#1156

- Adds checks for isWordpressApp.
- It properly checks for WordPress config files and doesn't overwrite them if they are user managed.
- Instructs the user on how to complete configuration if a user-managed wp-config.php file exists.

**Base WordPress Install**
- From a new WordPress install, without a wp-config.php file, run `ddev config`
- Ensure that ddev has written both a wp-config.php and a wp-config-ddev.php
- Ensure that specific ddev database connection information exists in wp-config-ddev.php
- Ensure that wp-config.php contains other required non-ddev-specific configurations

**A Previously Configured WordPress Install**
- In the same directory, edit wp-config.php to remove the `#ddev-generated` signature
- Edit wp-config-ddev.php in some noticeable way 
- Run `ddev config`
- Ensure that ddev warns the user that a user-managed wp-config.php already exists and must be edited to include the PHP snipped written to the terminal
- Ensure that ddev **does not** indicate that configuration is complete or that the user can run ddev start to start the project
- Ensure that wp-config.php has not been overwritten
- Ensure that wp-config-ddev.php has been overwritten

**WordPress-in-a-directory**
- From a new WordPress-in-a-directory install, run `ddev config`
- Ensure that ddev can detect the correct app type (`wordpress`)
- Ensure that wp-config.php and wp-config-ddev.php have been written to the docroot
- Ensure that wp-config.php includes the appropriate `ABSPATH` value

## Related Issue Link(s):
#756 
#1156 

## Release/Deployment notes:

The DDEV docs have been updated to reflect the changes made here.


rfay:
Hmm, I seem to have skipped by the comment form in "request changes" somehow, but basically the need is to recognize that *we're generating this file* and thus the comments and actions taken need to recognize that. 

I also note this stanza at the end of the file:

```
wp-settings.php is typically included in wp-config.php. This check ensures it is not
included again if this file is written to wp-config-local.php.
*/
if (basename(__FILE__) == "wp-config.php") {
	require_once(ABSPATH . '/wp-settings.php');
}
```

That I don't think makes sense does it?

pkg/ddevapp/wordpress.go
```diff
@@ -184,7 +286,7 @@ func setWordpressSiteSettingsPaths(app *DdevApp) {
 	settingsFileBasePath := filepath.Join(app.AppRoot, app.Docroot)
 	var settingsFilePath, localSettingsFilePath string
 	settingsFilePath = filepath.Join(settingsFileBasePath, "wp-config.php")
-	localSettingsFilePath = filepath.Join(settingsFileBasePath, "wp-config-local.php")
+	localSettingsFilePath = filepath.Join(settingsFileBasePath, "wp-config-ddev.php")
```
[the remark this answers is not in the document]
alkymst: I can rename the var `localSettingsFilePath` to `ddevSettingsFilePath`, that makes sense.

It will not break anything necessarily, because additional/environment configs requires manual work in `wp-config.php` to be included in the first place. Before these changes when DDEV created `wp-config-local.php` it was not be used unless a user added it by hand.

So `wp-config-local.php` is a non-standard WP way to handle multiple environments. If it exist, and we didn't create it. It's being used for a different local setup most likely. The user has to do work to include and additional configs by hand in the `wp-config.php`. That's why I felt for users trying out DDEV with an existing site, this naming convention made more sense. I believe passed on past PR's I have sen from @cweagans I feel other team members felt the same way.

Environment Based  WP Config Examples:
https://abandon.ie/notebook/wordpress-configuration-for-multiple-environments
https://jonsuh.com/blog/configure-wordpress-for-multiple-environments/
https://github.com/studio24/wordpress-multi-env-config
https://github.com/Abban/WordPress-Config-Bootstrap/blob/master/wp-config.php
https://gist.github.com/ahaywood/c11f1d756521c3edee9a822c0b9cccc7

I'm totally open to including both or using `wp-config-local.php`. Ideally we use env vars.

pkg/ddevapp/wordpress.go
```diff
@@ -194,6 +296,10 @@ func isWordpressApp(app *DdevApp) bool {
 	if _, err := os.Stat(filepath.Join(app.AppRoot, app.Docroot, "wp-login.php")); err == nil {
 		return true
 	}
+	// TODO: Add wildcard or ENV var to make more flexible, ie wordpress/
```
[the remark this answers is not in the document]
alkymst: I'll add better notes as to why we are doing the additional check.

We are looking for WP in a sub-directory, the earlier check was looking for it only in the docroot. Ideally it's a config settings we can control. All I'm adding is a check for WordPress in `wp`, but it could be in `wordpress` or whatever they want to call it by defining `WP_SITEURL`.

`define('WP_SITEURL', 'https://www.sitename.com/wp);`

pkg/ddevapp/wordpress.go
```diff
@@ -78,46 +81,80 @@ const (
  ddev manages this file and may delete or overwrite the file unless this comment is removed.
  */
 
+// Include local settings if it exists.
+$docroot = getenv('NGINX_DOCROOT');
+if (file_exists($docroot . '/wp-config-ddev.php')) {
+	require_once $docroot . '/wp-config-ddev.php';
+}
+
 // ** MySQL settings - You can get this info from your web host ** //
```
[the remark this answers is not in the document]
alkymst: We don't want this file pushed to any other environment therefor it only exist when running locally. That is why we are conditionally including it. Most users will not version their `wp-config.php` if they do they are 100% using it for production settings, and conditionally loading environment specific configs. Now if we generate the `wp-config.php` I agree we wouldn't want to include it, however if we generate the `wp-config-ddev.php` we need to tell the user to include it, or include it for them.

pkg/ddevapp/wordpress.go
```diff
@@ -78,46 +81,80 @@ const (
...
+	define( 'DB_PASSWORD', '{{ $config.DatabasePassword }}' );
 
 /** MySQL hostname */
-define('DB_HOST', '{{ $config.DatabaseHost }}');
+if (!defined('DB_HOST'))
+ 	define( 'DB_HOST', '{{ $config.DatabaseHost }}' );
 
 /** Database Charset to use in creating database tables. */
-define('DB_CHARSET', 'utf8mb4');
+define( 'DB_CHARSET', 'utf8mb4' );
```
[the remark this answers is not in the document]
alkymst: I'm not 100% sure, it's something that I have always set with DB creds.

```/**
 * DB settings
 */
define('DB_NAME', env('DB_NAME'));
define('DB_USER', env('DB_USER'));
define('DB_PASSWORD', env('DB_PASSWORD'));
define('DB_HOST', env('DB_HOST') ?: 'localhost');
define('DB_CHARSET', 'utf8mb4');
define('DB_COLLATE', '');
$table_prefix = env('DB_PREFIX') ?: 'wp_';

File: pkg/ddevapp/wordpress.go [hunk 5]
```diff
@@ -78,46 +81,80 @@ const (
...
 
 /** The Database Collate type. Don't change this if in doubt. */
-define('DB_COLLATE', '');
+define( 'DB_COLLATE', '' );
```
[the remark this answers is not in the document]
alkymst: I agree. It's carried over from the default WP config

pkg/ddevapp/wordpress.go
```diff
@@ -78,46 +81,80 @@ const (
...
 
 /**
  * WordPress Database Table prefix.
  */
-$table_prefix  = '{{ $config.TablePrefix }}';
+if(!isset($table_prefix))
```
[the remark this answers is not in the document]
alkymst: Not sure I understand the question 100%.

If `wp-config-ddev.php` is includes it would be coming from there. So it's falling back to `$config.TablePrefix` if not set.

You have to set the DB prefix for WP which by default is `wp_`. However a decent amount of suer have a different prefix, hosts will have unique prefixes etc.

pkg/ddevapp/wordpress.go
```diff
@@ -137,24 +174,89 @@ if (basename(__FILE__) == "wp-config.php") {
...
+	}
+
+	return settingsFilePath, nil
 }
 
 // WriteWordpressConfig dynamically produces valid wp-config.php file by combining a configuration
 // object with a data-driven template.
 func WriteWordpressConfig(wordpressConfig *WordpressConfig, filePath string) error {
-	tmpl, err := template.New("wordpressConfig").Funcs(sprig.TxtFuncMap()).Parse(wordpressTemplate)
+
```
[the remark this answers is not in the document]
alkymst: Well all that really matters to get the site working on DDEV is what's in `worpressTemplateSimple`. The `wordpressTemplate` is more closely mirroring an end to end wp-config. `worpressTemplateSimple` is a way to set what DDEV needs to connect.

What makes this weird is really how WP handles config settings. Being constants makes it really weird to "overwrite" settings. Environment variables being the best option really.

https://github.com/studio24/wordpress-multi-env-config

alkymst:
@rfay I have made some of the requested updates and responded to all of your review questions. Let me know when you can take it for a spin again.

pkg/ddevapp/wordpress.go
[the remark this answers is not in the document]
rfay: Looks like you changed from wordpressTemplateSimple to wordpressTemplateDdev?

We really don't want more than one type of template do we? 

AFAICT there should just be wordpressTemplateDdev, and it should not include the "If" statements, those should just be defines.

rfay:
Moved this to v1.1.0, since it's not going to get short-term attention.

rfay:
I took the liberty of rebasing this.

alkymst:
The ifdef's can't go if we want WordPress to work.

rfay:
@alkymst please expound on that. We *must* have settings.ddev.php *overriding* any existing definitions, not respecting them.

rfay:
I just removed all the ifdefs from the generated wp-config-ddev.php and it *seems* to work ok.

rfay:
My take on this is that we can't do what we do for other CMSs if the (unmanaged) wp-config.php already exists. In that case I suspect we just have to:

1. Create the wp-config-ddev.php
2. Add to their wp-settings.php:
```
 // Automatically generated include for wp-config-settings.php managed by ddev.
if (file_exists(ABSPATH . 'wp-config-ddev.php') && !defined('DB_USER')) {
  require_once(ABSPATH . 'wp-config-ddev.php');
}
```
3. Output instructions explaining that they have to comment out the db config in their wp-config.php (and to be careful not to commit of course. There may be people who don't know that you shouldn't commit that stuff anyway....). 

Note: I do *not* think we should define anything in the wp-settings-ddev.php *except* the db credentials (at this time). In other words, nothing should be there except things that are specific to local development. (No AUTH_KEY, etc.). 

The wp-settings.php that we might generate if *no* wp-settings.php exists though should include whatever is necessary for a working wordpress project, which I assume includes tings like AUTH_KEY and such. So ddev-managed wp-settings.php and ddev-managed wp-settings-ddev.php would *not* be the same.

rfay:
For conversation here are the generated files. 

This doesn't exactly match the proposal in https://github.com/drud/ddev/pull/932#issuecomment-425934263 above.

* We were going to go ahead and edit/add the include if there was an existing wp-config.php (like we do for other CMSs)
* No use of NGINX_DOCROOT is appropriate anywhere, as that isn't even a thing any more. Please use the code suggested in the proposal.
* My understanding of the proposal and the discussion was that we were going to tell them to *comment out* any db configuration in their wp-config.php. (See the proposal above)
* Current output to them is in yellow. Since their site is now not functional, I think maybe util.Error()? 

I haven't looked at any of the golang code, but this gives us a start for conversation.

pkg/ddevapp/apptypes_test.go
```diff
@@ -23,7 +24,7 @@ func TestApptypeDetection(t *testing.T) {
 		"drupal6":   "misc/ahah.js",
 		"drupal7":   "misc/ajax.js",
 		"drupal8":   "core/scripts/drupal.sh",
-		"wordpress": "wp-login.php",
+		"wordpress": "wp-settings.php",
```
rfay: I actually thought that @alkymst had said either of these would be legit.

pkg/ddevapp/ddevapp_test.go
```diff
@@ -538,11 +538,10 @@ func TestDdevImportDB(t *testing.T) {
 				assert.True(settingsHashSalt)
 			case "wordpress":
 				// nolint: vetshadow
-				hasAuthSalt, err := fileutil.FgrepStringInFile(app.SiteSettingsPath, "SECURE_AUTH_SALT")
+				hasAuthSalt, err := fileutil.FgrepStringInFile(app.SiteLocalSettingsPath, "SECURE_AUTH_SALT")
```
rfay: This is odd because it's actually in our generated SiteSettingsPath, not in the SiteLocalSettingsPath (wp-config-ddev.php). Are these two files swapped somehow? I'd expect SiteSettingsPath to be wp-config.php and SiteLocalSettingsPath to be wp-config-ddev.php. BTW, I'm *way* ok with renaming that to SiteDdevSettingsPath.

pkg/ddevapp/settings_test.go
```diff
@@ -41,7 +42,7 @@ func TestWriteSettings(t *testing.T) {
 		"drupal6":   "sites/default/settings.ddev.php",
 		"drupal7":   "sites/default/settings.ddev.php",
 		"drupal8":   "sites/default/settings.ddev.php",
-		"wordpress": "wp-config.php",
+		"wordpress": "wp-config-ddev.php",
```
rfay: Yeah, I think this is where things got backwards. If I'm reading right the settings path should be wp-config.php.

rfay: The reasoning behind the original naming of "local" and such: "settings.local.php" is the traditional name of a site-specific add-on for a drupal settings file. So D8 settings.php has a commented out "load settings.local.php" stanza. When we switched to settings.ddev.php, the name "local" followed, which is now quite confusing because people are referring to ddev as "local" or DDEV-Local, but that use of "local" has nothing to do with what was going on here.

pkg/ddevapp/wordpress.go
```diff
@@ -100,75 +137,73 @@ define('DB_PASSWORD', '{{ $config.DatabasePassword }}');
...
- * WordPress Database Table prefix.
- */
-$table_prefix  = '{{ $config.TablePrefix }}';
-
-/**
- * For developers: WordPress debugging mode.
- */
-define('WP_DEBUG', false);
+/** WP_SITEURL location */
+define('WP_SITEURL', WP_HOME . '/{{ $config.AbsPath  }}');
```
rfay: I think @alkymst said this wasn't necessary, but I don't think it does harm?

pkg/ddevapp/wordpress.go
```diff
@@ -182,37 +217,68 @@ func WriteWordpressConfig(wordpressConfig *WordpressConfig, filePath string) err
...
+	_, err := getRelativeAbsPath(app)
+	if err != nil {
+		return false
 	}
-	return false
+
+	return true
 }
 
 // wordpressPostImportDBAction just emits a warning about updating URLs as is
```
rfay: Let's just remove this since it does nothing. There's no need to keep wordpressPostImportDBAction

pkg/ddevapp/apptypes_test.go
alkymst: @rfay wp-settings is what WP core looks for, so I felt it made sense to align with that.

pkg/ddevapp/wordpress.go
alkymst: They are both necessary if WP is in a sub-directory. It's more failsafe in general too.

pkg/ddevapp/wordpress.go
```diff
@@ -260,3 +326,40 @@ func wordpressImportFilesAction(app *DdevApp, importPath, extPath string) error
 
 	return nil
 }
+
+// getRelativeAbsPath returns the portion of the ABSPATH value that will come after "/" in wp-config.php -
+// this is done by searching (at a max depth of one directory from the docroot) for wp-settings.php, the
+// file we're using as a signal to indicate that this is a WordPress project.
+func getRelativeAbsPath(app *DdevApp) (string, error) {
```
rfay: I think you'd be much happier here with a very simple `matches, err := filepath.Glob("*/<string>")` - I haven't tried that globbing directories, but I'd think it would work.

If this is just about Wordpress, let's name it as being for wordpress. It looks like it's intended to be a general-purpose func by the naming. But I do think you'll like the glob.

rfay:
I'm not sure, but think the OP will need some attention. This PR ended up doing more things than originally envisioned, and I think the testing instructions will require more.

pkg/ddevapp/wordpress.go
andrewfrench: Good point on renaming, I'll make that change.

I'm looking into updating this to use filepath.Glob() now -- it's not _quite_ as simple as that example as that _only_ looks in subdirectories, but I think there's some simplification there.

pkg/ddevapp/wordpress.go
andrewfrench: Sure, removed.

pkg/ddevapp/wordpress.go
rfay: Hope you saw the example on github, which looks exactly one directory down for a file.

andrewfrench:
(I can only respond directly to code comments, so I'm dumping a few responses here)

The manual testing instructions and summary of changes do need some work, I'll update those.

I agree with the note about output color - I can change that to red to indicate the required changes necessary.

I don't believe [this comment](https://github.com/drud/ddev/pull/932#issuecomment-425934263) captures the most up-to-date plan, if only because we've had two substantial discussions in the time since it was posted and the comment hasn't been significantly changed to reflect those. To be clear (and I should have captured this in a comment here immediately after the call), my understanding of the config flow was the following:

- Always create/overwrite wp-config-ddev.php on config
- If no wp-config.php exists, create it and include wp-config-ddev.php
- If wp-config.php exists and is ddev-managed, overwrite on config
- If wp-config.php exists and is not ddev-managed, point the user to wp-config-ddev.php and instruct them to make the required config changes to use the ddev values

rfay:
The first three bullets are the same as I think; only the last do we differ on. I did make significant edits to https://github.com/drud/ddev/pull/932#issuecomment-425934263 as we talked in the clocktower, and I thought I got it. But the key difference in opinion is what the user should be told to do:

My memory (and what I tried to edit at the time) was that 
* we were going to tell the user to comment out any db settings in the main (non-ddev-managed) wp-config.php. And
*  we would add the include of wp-config-ddev.php to the bottom as we've done in other CMSs. If that's not possible (and I do remember something about it not being possible because it had to be before the include of wp-settings.php?), then we need to give them explicit instructions of how to add the include (verbatim).

andrewfrench:
Agreed on all points, we can use this as a starting point for wider support of WordPress variants.

PR #936, opened 2018-06-21
Rename ssh command

jdoubleu:
I'd like to use this PR as a discussion on whether or not to rename the ssh command.

Please also take https://github.com/drud/ddev/issues/484 into account.

## The Problem/Issue/Bug:
My propose is to rename the `ssh` command to `shell`. What the command actually does is to open a shell in a service and attaches it to the current *tty*.
The *ssh* indicates that it runs a ssh server inside some container and connects to it using a local ssh client.

Using `shell` is just cleaner in my opinion.

However I see the sometimes misuse of the term *ssh* and same behavior of a ssh client so I added the `ssh` command as alias to the `shell` command. Also this is required for backwards compatibility.

## Related Issue Link(s):
* Consider reorganizing start/stop/rm commands https://github.com/drud/ddev/issues/484


rfay:
Thanks so much for your effort on this! It's probably better to finish a conversation in the issue before investing the effort in a PR, but 💯 for your initiative! We normally like to think through changes like this for some time, as they impact all users. So I'd like to leave your issue/suggestion open and close this PR for now.  I don't disagree with your thinking, and it's true that `ddev ssh` isn't actually ssh. But we don't need to move quite that fast at this time. Thanks, and again your initiative is much appreciated.

PR #938, opened 2018-06-21
Remove duplicate config path resolving in ddevapp/config.go

jdoubleu:
## The Problem/Issue/Bug:
The path to the config file (`config.yml`) is set multiple times (see https://github.com/drud/ddev/blob/master/pkg/ddevapp/config.go#L61 and https://github.com/drud/ddev/blob/master/pkg/ddevapp/config.go#L64) in config class when creating a new application.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -58,7 +58,6 @@ type Provider interface {
 func NewApp(AppRoot string, provider string) (*DdevApp, error) {
 	// Set defaults.
 	app := &DdevApp{}
-	app.ConfigPath = filepath.Join(AppRoot, ".ddev", "config.yaml")
 
 	app.AppRoot = AppRoot
 	app.ConfigPath = app.GetConfigPath("config.yaml")


rickmanelius:
Hi @jdoubleu! Thanks for submitting this. Question: Does this issue result in something that causes a break for the end-user? Is there a way to show it breaking that can be quickly testable (e.g. First run a ddev config, run ddev start, re-run ddev config and do X, Y, and Z)?

Added @rfay and @andrewfrench to review, but curious how to test to verify pre/post.

jdoubleu:
I don't think it does change anything.
It was just a duplicate variable declaration.

pkg/ddevapp/config.go
```diff
@@ -58,7 +58,6 @@ type Provider interface {
 func NewApp(AppRoot string, provider string) (*DdevApp, error) {
 	// Set defaults.
 	app := &DdevApp{}
-	app.ConfigPath = filepath.Join(AppRoot, ".ddev", "config.yaml")
```
andrewfrench: Good removal here for two reasons: we're redeclaring `app.ConfigPath` immediately below without having used the initial value, and `app.GetConfigPath()` will build the value using the same process.

PR #940, opened 2018-06-25
Drupal8 settings improvements, fixes #504, fixes #680

### LINKED ISSUE #504: Generated settings.php needs 'sync' key in drupal8

Drupal 8 insists on a `sync` key, to a writeable directory, in settings.php. We should add that. 

* For extra credit, add the "Trusted Hosts" directive to avoid that complaint from D8

status_report___rfay-drupal8

### LINKED ISSUE #680: Local settings for Drupal 8 and TYPO3 need to support trusted_host patterns (and with https)

When installing TYPO3 9 for the first time on ddev using https, an warning is thrown during installation wizard.

```
Trusted hosts pattern mismatch
The trusted hosts pattern will be configured to allow all header values. This is because your $SERVER_NAME:[defaultPort] is "typo3-master-intro.ddev.local:443" while your HTTP_HOST:SERVER_PORT is "typo3-master-intro.ddev.local:80". Check the pattern defined in Install Tool -> All configuration -> System -> trustedHostsPattern and adapt it to expected host value(s).
```

image

No warning - as it is when installing using http.

## How to reproduce this:

clone TYPO3 repo, run composer install, ddev config, ddev start, and follow the install wizard

The issue exists in both default image and the new:
drud/nginx-php-fpm-local:20180224_multi_cms_nginx

andrewfrench:
## The Problem/Issue/Bug:
Two minor settings issues that cause errors on a normal Drupal8 install.

This PR adds two lines to the generated Drupal8 settings file: the first defines a trusted hosts pattern, the second defines the config sync directory.  The existence of these settings lines prevent the errors outlined in #504 and #680.

- Install a fresh Drupal8 site
- Navigate to `<site>.ddev.local/admin/reports/status`
- Ensure no errors are presented 

## Related Issue Link(s):
#504 
#680

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -12,6 +12,8 @@ import (
 	"path"
 	"path/filepath"
 	"text/template"
+
+	"github.com/drud/ddev/pkg/fileutil"
 )
 
 // DrupalSettings encapsulates all the configurations for a Drupal site.
@@ -503,8 +505,7 @@ func createDrupal8SyncDir(app *DdevApp) error {
 	drupalConfig := NewDrupalSettings()
 
 	syncDirPath := path.Join(app.GetAppRoot(), app.GetDocroot(), drupalConfig.SyncDir)
-	if _, err := os.Stat(syncDirPath); !os.IsNotExist(err) {
-		// Directory already exists.
+	if fileutil.FileExists(syncDirPath) {
 		return nil
 	}
 
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -509,7 +509,7 @@ func createDrupal8SyncDir(app *DdevApp) error {
 	}
 
 	if err := os.MkdirAll(syncDirPath, 0755); err != nil {
-		return fmt.Errorf("failed to create sync directory: %v", err)
+		return fmt.Errorf("failed to create sync directory (%s): %v", syncDirPath, err)
 	}
 
 	return nil
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -27,6 +27,7 @@ type DrupalSettings struct {
 	DatabasePrefix   string
 	HashSalt         string
 	Signature        string
+	SyncDir          string
 }
 
 // NewDrupalSettings produces a DrupalSettings object with default.
@@ -41,6 +42,7 @@ func NewDrupalSettings() *DrupalSettings {
 		DatabasePrefix:   "",
 		HashSalt:         util.RandString(64),
 		Signature:        DdevFileSignature,
+		SyncDir:          "sites/default/files/sync",
 	}
 }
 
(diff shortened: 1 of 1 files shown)


rfay:
Here's what Drupal does when it's creating a site from scratch (and a settings.php from scratch): 
```
$config_directories['sync'] = 'sites/default/files/config_J-q1PFJ0IVTR2JCk6vnFr69fv5H6r8EsCbxINa6ArAOoaFyFUI--C1m4jxO77DJ6xmOUcN8idA/sync';
```

It also creates that directory and creates a README.txt containing

```
This directory contains configuration to be imported into your Drupal site. To make this configuration active, visit admin/config/development/configuration/sync. For information about deploying configuration between servers, see https://www.drupal.org/documentation/administer/config
```

I'm not sure we need to do the README or do a fancy directory name, but we should create the directory, and based on this I think putting it in sites/default/files would be appropriate. 

But I suspect we should *not* overwrite the `$config_directories['sync']` if it's already set. That would be the case in a user-owned settings.php that included our stuff in a settings.local.php.  So probably 

```
if !empty($config_directories['sync']) {
   set it
}
```

andrewfrench:
That sounds good to me, but it's a different level of effort that won't make this round.  I'll close this and circle back.

andrewfrench:
I've added a few lines that check if the sync directory has been set and create the directory if it doesn't exist. It's been a real long time since I've written any PHP, so let me know if there are any style or function choice issues, but I got it to do what it should.

pkg/ddevapp/drupal.go
```diff
@@ -93,6 +93,18 @@ $settings['file_scan_ignore_directories'] = [
...
 
+// This will ensure the site can only be accessed through the intended host names.
+// Additional host patterns can be added for custom configurations.
+$settings['trusted_host_patterns'] = ['.*'];
+
+// This specifies the default configuration sync directory.
+if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
+  $config_directories[CONFIG_SYNC_DIRECTORY] = 'sites/default/files/sync';
+  if (!file_exists($config_directories[CONFIG_SYNC_DIRECTORY])) {
+    mkdir($config_directories[CONFIG_SYNC_DIRECTORY]);
```
rfay: Oh, I was thinking we should do the mkdir in go, interesting. I wonder where it belongs. in php though it probably needs 'recursive' set in case the dir doesn't exist yet, http://php.net/manual/en/function.mkdir.php

pkg/ddevapp/drupal.go
```diff
@@ -41,6 +42,7 @@ func NewDrupalSettings() *DrupalSettings {
 		DatabasePrefix:   "",
 		HashSalt:         util.RandString(64),
 		Signature:        DdevFileSignature,
+		SyncDir:          "sites/default/files/sync",
```
rfay: This isn't correct, and results in loads of trouble. This should be relative to the docroot.

andrewfrench:
I've updated this to specify the absolute path to the sync dir in the container with a new Docroot field in the Drupal settings struct (similar to how it was done in the Wordpress settings struct). Since we're making the path to the dir on the Go side, I found that we needed different host and container paths (/Users/andrew/... vs /var/www/html/...), so I'm now doing some path construction on the fly in the template and in the function that creates the path.

rfay:
Please rebase this when you're working on it as well, hoping to see those windows container improvements, etc.

andrewfrench:
Rebased, and looks like the dockerforwindows is having a little bit of trouble.  Looks like it failed twice immediately during/after TestGetVersionInfo, running one more time to see if that test is the issue or if it was a coincidence.

rfay:
Fixes look good to me. I tested actually using the sync, and this works fine.

The only thing that was a gotcha for me was that drupal-composer build sets up and creates <webroot>/config/sync, but does it at the end of the settings.php, so I wasn't looking for another instance of it there.

PR #945, opened 2018-06-27
Don't use apk cache to reduce image size

t3easy:
## The Problem/Issue/Bug:
Apk cache remains in image and makes it larger as it must be
Use apk --no-cache parameter

### DIFF
--- a/containers/mariadb-local/Dockerfile.in
+++ b/containers/mariadb-local/Dockerfile.in
@@ -8,7 +8,7 @@ ENV MYSQL_ROOT_PASSWORD root
 RUN mkdir /docker-entrypoint-initdb.data
 
 # Install mariadb and other packages
-RUN apk update && apk add mariadb mariadb-client bash tzdata shadow
+RUN apk add --no-cache mariadb mariadb-client bash tzdata shadow
 # Remove the installed version as we need to set up our own from scratch
 
 RUN rm -rf /var/lib/mysql/* /etc/mysql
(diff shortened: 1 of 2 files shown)


rickmanelius:
Added @rfay to review because I wasn't sure if there was a particular reason for this.

rfay:
This looks great to me, thanks. I pushed this to drud/mariadb-local:20180627_mariadb_no_apk_cache - @t3easy if you could please add a commit setting that container in pkg/version/version.go it would be fantastic. Otherwise ask us to do it and tell us you don't mind us pushing into your fork.

rfay:
I should note that this seems to reduce the image size by less than 1MB, so might not be as effective as you hoped.

andrewfrench:
Just to confirm @rfay's point about the image size, here's a comparison of this version and the v0.20.0 image:
```
CONTAINER ID        IMAGE                         SIZE
340e94a2a9b6        drud/mariadb-local:v0.20.0    151MB (virtual 362MB)
d6918a2aceca        drud/mariadb-local:945-test   151MB (virtual 361MB)
```

t3easy:
It's not dramatically but best practice. 

Regarding
> if you could please add a commit setting that container in pkg/version/version.go it would be fantastic

Just rais `var DBTag = "v0.20.0"` to v0.20.1?

rfay:
`var DBTag = 20180627_mariadb_no_apk_cache` - the tag I pushed. 

We use temporary tags between releases, when we release ddev v1.0.0 (the next one) we'll retag the latest version of the mariadb-local container to v1.0.0 to match the ddev version.

PR #964, opened 2018-07-02
ConfigFileOverrideAction() should only override if no config exists yet, fixes #900

### LINKED ISSUE #900: ddev config on an existing config updates the php version inappropriately
**Describe the bug**

`ddev config` with an existing configuration should respect most settings, certainly php version. 

However, it's updating in some situations, but not all.

Thanks to @jeffsheltren for finding and describing this.

**To Reproduce**

Drupal 7:
* Set the php_version in an existing config.yaml to 5.6
* Run ddev config and click through
* Note that the php_version is now set to 7.1

Drupal 6:
* Set the php_version in an existing config.yaml to 7.0
* Run ddev config and click through
* note the php_version is now set to 5.6

It doesn't seem to happen with Drupal 8. 

**Expected behavior**

Most config.yaml settings should be preserved by ddev config. Certainly php_version

 - Host computer OS and Version: macOS High Sierra
 - Docker version: 18.03
 - ddev version: v0.19.0

rfay:
## The Problem/Issue/Bug:

#900 - ddev config incorrectly updates php version in some situations.

ConfigOverrideAction() really should only override on a new config, and had only been used for D7 and D6. Both places it didn't work correctly.

It now only does its job if the config doesn't already exist.

Described in OP #900 

TestPostConfigAction() tests the expected behavior of ConfigFileOverrideAction().

## Related Issue Link(s):

OP #900

### DIFF
--- a/pkg/ddevapp/apptypes_test.go
+++ b/pkg/ddevapp/apptypes_test.go
@@ -5,9 +5,13 @@ import (
 	"path/filepath"
 	"testing"
 
+	"bufio"
+	"fmt"
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/testcommon"
+	"github.com/drud/ddev/pkg/util"
 	asrt "github.com/stretchr/testify/assert"
+	"strings"
 )
 
 // TestApptypeDetection does a simple test of various filesystem setups to make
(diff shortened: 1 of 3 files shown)


pkg/ddevapp/apptypes_test.go
```diff
@@ -44,3 +48,54 @@ func TestApptypeDetection(t *testing.T) {
...
+		"wordpress": ddevapp.DdevDefaultPHPVersion,
+		"backdrop":  ddevapp.DdevDefaultPHPVersion,
+	}
+
+	for appType, expectedPHPVersion := range appTypes {
+		testDir := testcommon.CreateTmpDir("TestApptype")
+
+		// testcommon.Chdir()() and CleanupDir() checks their own errors (and exit)
+		defer testcommon.CleanupDir(testDir)
+		defer testcommon.Chdir(testDir)()
```
andrewfrench: This may be what you intend, but these deferred calls won't be executed until after the entire test is complete (when `TestPostConfigAction` returns) and not at the end of the block.

rfay: Yes, that pattern is used all over the place, it's just what we did back in the day. The cleanup after test completion is fine, it's all just about cleaning up the various crap directories.

PR #966, opened 2018-07-03
Yet another round of battling building and testing containers on Windows

rfay:
## The Problem/Issue/Bug:

We seem to never be able to build and test our containers on WIndows *and* elsewhere. 

Trying again.

Script and docs (mostly) updates, and agent debugging

Build should run reliably on each testbot, especially windows testbots.

### DIFF
--- a/.buildkite/test_containers.sh
+++ b/.buildkite/test_containers.sh
@@ -7,6 +7,7 @@ set -o pipefail
 set -o nounset
 set -x
 
+echo "--- Cleanup docker"
 echo "Warning: deleting all docker containers and deleting ~/.ddev/Test*"
 if [ "$(docker ps -aq | wc -l)" -gt 0 ] ; then
 	docker rm -f $(docker ps -aq)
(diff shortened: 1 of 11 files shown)


rfay:
Much pain, much learning. I think this is ready for a quick review. Mostly I learned about agent setup, and may still have some work to do on that, maybe an agent qualifier script that makes sure git works right and all the right tools are installed?

.buildkite/test_containers.sh
```diff
@@ -7,6 +7,12 @@ set -o pipefail
 set -o nounset
 set -x
 
+echo "--- Cleanup docker"
+echo "Warning: deleting all docker containers and deleting ~/.ddev/Test*"
```
andrewfrench: I don't think ~/.ddev/Test* gets deleted, if we're still expecting that to happen.

rfay: Great point - It's a copy-from-original-script problem. This script doesn't ever create anything in ~/.ddev, I'll fix that.

rfay:
I'm probably good with this now, my energy may have run its course again :)

PR #970, opened 2018-07-04
Update docs about composer on Windows

rfay:
## The Problem/Issue/Bug:

We advocated using composer on the host in our quickstart, because that's fairly easy to do on macOS and Linux.

However, users on Windows will never be able to succeed with that, and have to be much more sophisticated.

Point them to the Stack Overflow article about ddev with composer.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -84,6 +84,7 @@ cd my-drupal8-site
 ```
 
 **Composer Setup Example**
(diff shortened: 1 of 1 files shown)


docs/users/cli-usage.md
```diff
@@ -127,6 +128,7 @@ cd example-typo3-site
 If necessary, run build steps that you may require, like `composer install` in the correct directory.
 
 **Composer Setup Example**
+*Do not use this technique with Windows, as composer on an NTFS filesystem can result in major problems with NTFS symlinks that can't be read in the Linux web container. Instead, use the [community-provided hints on Stack Overflow](https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev)*
```
andrewfrench: I believe lines 131 and 87 need another newline to put this warning in its own Markdown paragraph.

Issue #971, opened 2018-07-05
php.ini should add switch "-t" for the sendmail-call


neufeind:
With ddev and a typo3-project the default of sending a mail from TYPO3 to MailHog is using the sendmail-functionality of php. TYPO3 uses the famous SwiftMailer-component to send mail.

Trying to send a testmail SwiftMailer then fails with an error because it expects either "-bs" or "-t" in the sendmail-commandline. MailHog doesn't seem to care if -t is provided or not.

So let's imho just add "-t" to the sendmail-call in php.ini-files.

rfay:
This was just fixed in https://github.com/drud/ddev/pull/954

rfay:
And thanks for being so proactive and going out to fix it!

neufeind:
You're welcome. I used the installer-script which gave me the latest release (0.20.0) and not master - so your fix was not included. Also you fixed it different (by not using the sendmail-way but switching to SMTP).

Should we maybe still adjust the sendmail-binary-path in php.ini, as some other environment/component might also be using SwiftMailer some day and failing for the same reason?

rfay:
It's a bug in TYPO3 that it even grabs and uses the php.ini setting for sendmail_path (and sendmail does not exist in the web container). The sendmail_path is *correct* in place there. TYPO3 just uses it incorrectly (if configured for 'sendmail' sending), which is the default. mailhog with the args provided there works perfectly as discussed in those issues, TYPO3 (or swiftmailer?) is just doing some overly aggressive and odd things with that.

neufeind:
Well, it's about Swiftmailer imho. They removed (for security-reasons) support for using the php-mail()-function, with SwiftMailer v6. But they still have a possibility to directly use a sendmail-compatible transport by calling a binary. So what TYPO3 does is going and fetching the configured sendmail-command. And that's where things fail - since SwiftMailer does not just execute what is configured but checks the sendmail-command with a strpos() for "-bs" or "-t".
So if you say it "works as configured" (from the mailhog/php-side) then it might be on SwiftMailer to remove their check? But they might feel they are doing it for a good reason :-) But if you suggest I can reach out to them if they maybe want to fix it on the SwiftMailer-side.

rfay:
For TYPO3 and ddev (and Drupal and other CMSes) I think we have this sorted out just fine. But it is really strange what (swiftmailer) did with that config.

PR #954, opened 2018-06-28
Add configuration for typo3 mail and errors, default to php 7.2, fixes #924

rfay:
## The Problem/Issue/Bug:

* TYPO3 wasn't sending mail to mailhog correctly, per #924 
* TYPO3 config needed to display errors, something we'd always want in a dev environment
* TYPO3 default php version was 7.1, and 7.2 is required for TYPO3 v9 (and best elsewhere really)

## How this PR Solves The Problem:

* Adds config to the AdditionalConfiguration.php
* Sets default php to 7.2

## Manual Testing Instructions:

* Create a TYPO3 site with composer
* `ddev config` and step through
* Check that .ddev/config.yaml has php 7.2
* `ddev start`
* Visit and install the site
* In the ENVIRONMENT section of the backend "Admin Tools", use "Test Mail Setup" to send an email. The email should be delivered to mailhog (See `ddev describe` for accessing mailhog)

## Automated Testing Overview:

No testing was added.

## Related Issue Link(s):

#924 is fixed by this
@spoonerweb's excellent #925 PR provided the start for this, but it went in a different direction in the end. 

It would be lovely to get a review of this from the TYPO3 community.

PR #972, opened 2018-07-05
php.ini should add switch "-t" for the sendmail-call

neufeind:
The famous SwiftMailer-component (used for example by
TYPO3) expects either -bs or -t in the sendmail-call.
Provide -t by default, as it is usually done when
calling the real sendmail-binary.

Resolves: #971

## The Problem/Issue/Bug:
With ddev and a typo3-project the default of sending a mail from TYPO3 to MailHog is using the sendmail-functionality of php. TYPO3 uses the famous SwiftMailer-component to send mail.

Trying to send a testmail SwiftMailer then fails with an error because it expects either "-bs" or "-t" in the sendmail-commandline. MailHog doesn't seem to care if -t is provided or not.

Let's add "-t" to the sendmail-call in php.ini-files.

Installing a TYPO3-project with ddev, using the test-mail-functionality from the install-tool.

Calling the sendmail-binary with or without -t doesn't make a difference for MailHog.

## Related Issue Link(s):
https://github.com/drud/ddev/issues/971

### DIFF
--- a/containers/nginx-php-fpm-local/files/etc/php/5.6/cli/php.ini
+++ b/containers/nginx-php-fpm-local/files/etc/php/5.6/cli/php.ini
@@ -82,7 +82,7 @@ pdo_mysql.default_socket=
 SMTP = localhost
 smtp_port = 25
 mail.add_x_header = On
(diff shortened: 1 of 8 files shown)


rfay:
This was just fixed in #954, so but so appreciate your willingness to go find a solution to it!

PR #973, opened 2018-07-05
Fix case of Sequel Pro.app, fixes #969

### LINKED ISSUE #969: Sequelpro command fails for case sensitive filesystems
**Describe the bug**
I try to call `ddev sequelpro` on my mac. Sequelpro is installed and works fine. When running the command, I get the error _The sequelpro command is not available because sequel pro.app is not detected on your workstation_

I figured out, the ddev is looking for `/Applications/sequel pro.app` folder. But Sequel Pro is installed to `/Applications/Sequel Pro.app` by default. When you got a case-sensitive formatted hard disk, the call will fail because the folder can not be found.

**To Reproduce**
1. Use a Mac with case sensitive file system
2. Install Sequel Pro to the Applications folder
3. Setup a ddev project
4. Run `ddev sequelpro`

**Expected behavior**
Sequel Pro will start and automatically connect to the ddev database.

 - macOS High Sierra 10.13.5 with case-sensitive filesystem
 - Docker version 18.03.1-ce
 - ddev version 0.20.0

rfay:
## The Problem/Issue/Bug:

The app location for looking for Sequel Pro uses wrong capitalization, breaking the command for macOS systems with case-sensitive file systems.

Change the capitalization.

Unfortunately I wasn't able to figure out an easy way to test this, so will have to ask the issue originator to test.

This is too hard and too obscure to do tests.

## Related Issue Link(s):

OP #969

### DIFF
--- a/cmd/ddev/cmd/sequelpro.go
+++ b/cmd/ddev/cmd/sequelpro.go
@@ -21,7 +21,7 @@ import (
 
 // SequelproLoc is where we expect to find the sequel pro.app
 // It's global so it can be mocked in testing.
-var SequelproLoc = "/Applications/sequel pro.app"
+var SequelproLoc = "/Applications/Sequel Pro.app"
 
 // DdevSequelproCmd represents the sequelpro command
 var DdevSequelproCmd = &cobra.Command{


codemonkey1988:
I tested the  command `ddev sequelpro` for version https://circleci.com/gh/drud/ddev/3054#artifacts/containers/0 an it works on my machine with case-sensitive filesystem

jdoubleu:
Please see my notes: https://github.com/drud/ddev/commit/ccedee37bbb2a960b346be5c3a22cabdb17fc5d6#r29619377

rfay:
@jdoubleu you posted over there

> What if, for some reason, the app was installed/saved in lower case?
> Or if it was put in another location (e.g. `/Users/john/Applications`)?

> Due to the fact, that you are using the `open` command (see https://github.com/drud/ddev/blob/ccedee37bbb2a960b346be5c3a22cabdb17fc5d6/cmd/ddev/cmd/sequelpro.go#L88), I'd recommend to use the `-a` option:
```
open -a Sequel\ Pro
```

> This way macOS will deal with case-sensitivity and different locations.

And that does solve the *open* but not the *detection*.  But since we already asked for comments, waited for them, got confirmation, and pulled this... and since it took more than a year to discover and report the bug anyway... I think I'm good with it. But would imagine we'd accept a PR changing to open -a .... But if you have an idea how to solve the *detection* as well as the running that would be best.

PR #975, opened 2018-07-05
Improve mariadb-local test to work on linux without making un-deletable files

rfay:
## The Problem/Issue/Bug:

Nightly build is breaking due to changes in how we test/generate the mysql starter db. See https://circleci.com/gh/drud/ddev/3052?utm_campaign=build-failed&utm_medium=email&utm_source=notification

Try to make the output not violate fs permissions.

Run the nightly build against this PR.

### DIFF
--- a/.buildkite/macos_container.yml
+++ b/.buildkite/macos_container.yml
@@ -4,4 +4,5 @@
       - "dockertype=dockerformac"
     env:
       BUILDKITE_CLEAN_CHECKOUT: true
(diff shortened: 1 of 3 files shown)


rfay:
Looks to me like this will solve the problem @andrewfrench , and it has a little bit to say about macOS testing too, the buildkite builds were using /usr/local...buildkite as $HOME, and that isn't configured for mounting onto docker on default macOS docker-for-windows.

PR #986, opened 2018-07-11
Check testbot sanity before continuing with tests

rfay:
## The Problem/Issue/Bug:

We sometimes have a broken buildkite testbot in any of a number of ways, and never find out about it until it's run forever and ever. This additional script does an initial test to see if the testbot is sane before continuing the test.

### DIFF
--- a/.buildkite/sanetestbot.sh
+++ b/.buildkite/sanetestbot.sh
@@ -24,4 +24,4 @@ if [ $(go env GOOS) = "windows"  -a $(git config core.autocrlf) != "false" ] ; t
  exit 3
 fi
 
-echo "testbot seems to be set up OK"
+echo "=== testbot $HOSTNAME seems to be set up OK ==="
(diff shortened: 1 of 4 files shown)


.buildkite/sanetestbot.sh
```diff
@@ -0,0 +1,27 @@
...
+# Test that docker can allocate 80 and 443, get get busybox
+docker pull busybox:latest >/dev/null
+docker run -t -p 80:80 -p 443:443 -p 1081:1081 -p 1082:1082 --rm busybox:latest ls >/dev/null
+
+# Check that required commands are available.
+for command in mysql git go make; do
+    command -v $command >/dev/null || ( echo "Did not find command installed '$command'" && exit 2 )
+done
+
+if [ $(go env GOOS) = "windows"  -a $(git config core.autocrlf) != "false" ] ; then
```
andrewfrench: This line produces an error for me on OSX:
```
./sanetestbot.sh: line 22: [: too many arguments
```

This goes away when either the command outputs are wrapped in quotes:
```
"$(go env GOOS)"
```

Issue #774, opened 2018-04-05
Make ddev-router more transparent about failures


rfay:
## What happened (or feature request):

In https://github.com/drud/ddev/issues/760 @rickmanelius found that accidentally duplicating a hostname in additional_hostnames (either by duplicating the base name or by any number of other techniques) completely breaks the ddev-router. 

The reason is that the generated nginx config is invalid in this case (duplicate upstreams), so nginx just *fails* and there's no easy way to understand why.

We need to:

**On ddev-router**
* Make the container fail immediately when nginx fails, rather than restarting.
* Make the container fail immediately if the templating fails (the most common situation)
* Make it easy to get details of the failure.
* Suggest a troubleshooting route to the user when this happens.
* Handle the case of duplicate hostnames coming from different projects. Warn the user, or something.

**In ddev**
* Do a simple check to uniquify the array of hostnames used. This is not hard. 
* Perhaps check on start to see if there's a running project with the hostname offered. This is probably harder.

luksak:
@rfay I got this error as well:

```
$ docker logs ddev-router                                                                                       1 ↵  353  14:07:33
forego     | starting nginx.1 on port 5000
forego     | starting dockergen.1 on port 5100
forego     | starting dockergen.1 on port 5300
dockergen.1 | 2018/11/02 13:06:31 Generated '/gen-cert.sh' from 9 containers
dockergen.1 | 2018/11/02 13:06:31 Running 'chmod ugo+x /gen-cert.sh && /gen-cert.sh'
dockergen.1 | 2018/11/02 13:06:31 Generated '/etc/nginx/conf.d/default.conf' from 9 containers
dockergen.1 | 2018/11/02 13:06:31 Watching docker events
dockergen.1 | 2018/11/02 13:06:32 Contents of /gen-cert.sh did not change. Skipping notification 'chmod ugo+x /gen-cert.sh && /gen-cert.sh'
dockergen.1 | 2018/11/02 13:06:32 Error running notify command: sleep 1 && nginx -s reload, exit status 1
dockergen.1 | 2018/11/02 13:06:32 Watching docker events
dockergen.1 | 2018/11/02 13:06:32 Contents of /etc/nginx/conf.d/default.conf did not change. Skipping notification 'sleep 1 && nginx -s reload'
nginx.1    | 127.0.0.1 - - [02/Nov/2018:13:06:36 +0000] "GET /healthcheck HTTP/1.1" 404 169 "-" "curl/7.52.1" "-"
[...]

PR #988, opened 2018-07-12
Allow arbitrary additional_fqdns, fixes #868

### LINKED ISSUE #868: Allow custom hostnames
I wish that additional hostnames didn't get `.ddev.local` affixed to them.

I want to be able to define hostnames like `dev.fillpdf.io` and have them added to `/etc/hosts` as-is. I'm OK with it being a separate config variable than the existing one. 

Manually adding it to my `/etc/hosts` myself...

**Additional context**
I'm starting to move over my lesser-used VMs that were previously on `hashbangcode/vlad` to DDEV. It'd be nice to keep the same domain names I'm habituated to.

rfay:
## The Problem/Issue/Bug:

OP #868: People want to use URLs that do not container ddev.local. 

* Add additional_fqdns[] array to the config.yaml. People can then have as many URLs pointing to their project as they could ever want.
* Add docs explain it, but mentioning why to avoid this feature.

Set up config.yaml with additional_fqdns as in the docs added here. 

Added additional_fqdns testing into TestDdevStartMultipleHostnames()

## Related Issue Link(s):

OP #868 
[Make ddev-router more transparent about failures #774](https://github.com/drud/ddev/issues/774) - It's a known issue that two live sites can't share the same fqdn or additional_hostname, and it causes an annoying and hard-to-understand ddev-router failure to start.

### DIFF
--- a/docs/users/extend/additional-hostnames.md
+++ b/docs/users/extend/additional-hostnames.md
@@ -13,3 +13,18 @@ additional_hostnames:
 ```
 
 This configuration would result in working hostnames of mysite.ddev.local, extraname.ddev.local, fr.mysite.ddev.local, es.mysite.ddev.local, and it.mysite.ddev.local (with full http and https URLs for each).
+
+You can also provide additional_fqdn entries, which don't use the ".ddev.local" top-level domain. 
+
+```
+name: somename
+
+additional_fqdns:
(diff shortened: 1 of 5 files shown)


rfay:
I think this is ready for review.

PR #990, opened 2018-07-13
Improved management of settings files, introduce settings.ddev.php

andrewfrench:
## The Problem/Issue/Bug:
#468, improved management of settings files.

This PR creates settings.ddev.php, placing ddev-specific configurations in this file to preserve user settings.  The settings.php file is only modified to include settings.ddev.php.

[Randy's comment here](https://github.com/drud/ddev/issues/468#issuecomment-395548090) describes the desired behavior and was used as the template for the changes in this PR. 

For Drupal 6/7/8 and Backdrop, begin with the situation in which no settings files exist.
- Begin with no settings.php file and no settings.ddev.php file, and no .gitignore in the directory containing the settings files,
- `ddev config`
- Ensure a settings.php has been created that does nothing but include settings.ddev.php
- Ensure a settings.ddev.php has been created that contains ddev-generated configurations
- Ensure a .gitignore has been created that ignores settings.ddev.php

Now, test the situation in which a settings.php file already exists but does not include settings.ddev.php
- Begin with no settings.ddev.php and a settings.php file that has contents, but does not include settings.ddev.php
- `ddev config`
- Ensure settings.php has been modified to include settings.ddev.php
- Ensure settings.ddev.php has been created
- Ensure a .gitignore has been created that ignores settings.ddev.php

Finally, test the situation in which settings.pho and .gitignore already exist and won't be modified, and a settings.ddev.php will be overwritten.
- Begin with a settings.php that includes settings.ddev.php, an existing settings.ddev.php filled with "junk" data, and an existing user-managed .gitignore in the settings directory
- `ddev config`
- Ensure settings.php has not been modified to include settings.ddev.php
- Ensure settings.ddev.php has been overwritten and contains the settings as generated by ddev
- Ensure .gitginore has not been overwritten

Several tests have been added to ensure the desired creation/modification behavior of settings.php, settings.ddev.php, and .gitignore in settings_test.go.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -5,7 +5,7 @@ Type `ddev` or `ddev -h`in a terminal windows to see the available ddev commands
 
 ## Quickstart Guides
 
-These are quickstart instructions for WordPress, Drupal 7, Drupal 8, TYPO3, and Backdrop.
+These are quickstart instructions for WordPress, Drupal 6, Drupal 7, Drupal 8, TYPO3, and Backdrop.
 
 **Prerequisites:** Before you start, follow the [installation instructions](../index.md#installation). Make sure to [check the system requirements](../index.md#system-requirements), you will need *docker* and *docker-compose* to use ddev.
 
@@ -213,7 +213,7 @@ For in-depth application monitoring, use the command `ddev describe` to see deta
 
 ## Getting Started
 
-Check out the git repository for the project you want to work on. `cd` into the directory and run `ddev config` and follow the prompts.
+Check out the git repository for the project you want to work on. `cd` into the directory, run `ddev config`, and follow the prompts.
 
 ```
 $ cd ~/Projects
@@ -232,22 +232,7 @@ Docroot Location: web
 Found a drupal8 codebase at /Users/username/Projects/drupal8/web
 ```
 
-Configuration files have now been created for your project. (Take a look at the file on the project's .ddev/ddev.yaml file). Additionally, the `ddev config` steps attempts to create a CMS specific database settings with the DDEV specific credentials pre-populated. Here's a Drupal specific example that is mirrored in the other CMSes.
-
-* If settings.php does not exist, create it.
-* If a settings.php file exists that DDEV manages, recreate it.
-* If a settings.php file exists that DDEV does not manage, create settings.local.php.
-* If a settings.local.php file exists that DDEV manages, recreate it.
-* If a settings.local.php file exists that DDEV does not manage, warn the user and proceed.
-
-How do you know if DDEV manages a database settings file? You will see the following comment. Remove the comment and DDEV will not attempt to overwrite it!
-
-```
-/**
- #ddev-generated: Automatically generated Drupal settings.php file.
- ddev manages this file and may delete or overwrite the file unless this comment is removed.
- */
-```
+Configuration files have now been created for your project. Take a look at the project's .ddev/config.yaml file.
 
 Now that the configuration files have been created, you can start your project with `ddev start` (still from within the project working directory):
 
@@ -264,7 +249,29 @@ Your project can be reached at: http://drupal8.ddev.local
 
 And you can now visit your working project. Enjoy!
 
+### Configuration files
 _Please note that if you're providing the settings.php or wp-config.php and ddev is creating the settings.local.php (or wordpress wp-config-local.php), the main settings file must explicitly include the appropriate "settings.local.php" or equivalent._
+
+The `ddev config` command attempts to create a CMS-specific settings file with DDEV credentials pre-populated.
+
+For **Drupal** and **Backdrop**, DDEV settings are written to a DDEV-managed file, settings.ddev.php. The `ddev config` command will ensure that these settings are included in your settings.php through the following steps:
+
+* Write DDEV settings to settings.ddev.php
+* If no settings.php file exists, create one that includes settings.ddev.php
+* If a settings.php file already exists, ensure that it includes settings.ddev.php, modifying settings.php to write the include if necessary
+
+For **TYPO3**, DDEV settings are written to AdditionalConfiguration.php.  If AdditionalConfiguration.php exists and is not managed by DDEV, it will not be modified.
+
+For **Wordpress**, DDEV settings are written to wp-config.php.
+
+How do you know if DDEV manages a settings file? You will see the following comment. Remove the comment and DDEV will not attempt to overwrite it!
+
+```
+/**
+ #ddev-generated: Automatically generated Drupal settings.php file.
+ ddev manages this file and may delete or overwrite the file unless this comment is removed.
+ */
+```
 
 ## Listing project information
 
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -250,7 +250,7 @@ Your project can be reached at: http://drupal8.ddev.local
 And you can now visit your working project. Enjoy!
 
 ### Configuration files
-_Please note that if you're providing the settings.php or wp-config.php and ddev is creating the settings.local.php (or wordpress wp-config-local.php), the main settings file must explicitly include the appropriate "settings.local.php" or equivalent._
+_**Note:** If you're providing the settings.php or wp-config.php and DDEV is creating the settings.ddev.php (or wp-config-local.php, AdditionalConfig.php, or similar), the main settings file must explicitly include the appropriate DDEV-generated settings file._
 
 The `ddev config` command attempts to create a CMS-specific settings file with DDEV credentials pre-populated.
 
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -119,7 +119,7 @@ func createBackdropSettingsFile(app *DdevApp) (string, error) {
 	}
 
 	if err := createGitIgnore(filepath.Dir(app.SiteLocalSettingsPath), settings.SiteSettingsLocal); err != nil {
-		output.UserOut.Warnf("Failed to write .gitignore: %v", err)
+		output.UserOut.Warnf("Failed to write .gitignore in %s: %v", app.SiteLocalSettingsPath, err)
 	}
 
 	return app.SiteLocalSettingsPath, nil
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -143,7 +143,7 @@ func writeBackdropMainSettingsFile(settings *BackdropSettings, filePath string) 
 
 	// Ensure target directory is writable.
 	dir := filepath.Dir(filePath)
-	if err := os.Chmod(dir, 0755); err != nil {
+	if err = os.Chmod(dir, 0755); err != nil {
 		return err
 	}
 
@@ -170,7 +170,7 @@ func writeBackdropDdevSettingsFile(settings *BackdropSettings, filePath string) 
 
 	// Ensure target directory is writable
 	dir := filepath.Dir(filePath)
-	if err := os.Chmod(dir, 0755); err != nil {
+	if err = os.Chmod(dir, 0755); err != nil {
 		return err
 	}
 
@@ -242,7 +242,7 @@ func addIncludeToBackdropSettingsFile(settings *BackdropSettings, siteSettingsPa
 
 	// If settings.php is empty, write the complete settings template
 	if len(currentSiteSettings) == 0 {
-		if err := writeBackdropMainSettingsFile(settings, siteSettingsPath); err != nil {
+		if err = writeBackdropMainSettingsFile(settings, siteSettingsPath); err != nil {
 			return err
 		}
 	}
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -51,8 +51,8 @@ func NewBackdropSettings() *BackdropSettings {
 const backdropMainSettingsTemplate = `<?php
 {{ $config := . }}
 // {{ $config.Signature }}: Automatically generated Backdrop settings file.
-if (file_exists('{{ $config.SiteSettingsLocal }}')) {
-  include '{{ $config.SiteSettingsLocal }}';
+if (file_exists(__DIR__ . '/{{ $config.SiteSettingsLocal }}')) {
+  include __DIR__ . '/{{ $config.SiteSettingsLocal }}';
 }
 `
 
@@ -60,8 +60,8 @@ if (file_exists('{{ $config.SiteSettingsLocal }}')) {
 // settings.php in the event that one exists.
 const backdropSettingsAppendTemplate = `{{ $config := . }}
 // Automatically generated include for settings managed by ddev.
-if (file_exists('{{ $config.SiteSettingsLocal }}')) {
-  include '{{ $config.SiteSettingsLocal }}';
+if (file_exists(__DIR__ . '/{{ $config.SiteSettingsLocal }}')) {
+  include __DIR__ . '/{{ $config.SiteSettingsLocal }}';
 }
 `
 
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -119,7 +119,7 @@ func createBackdropSettingsFile(app *DdevApp) (string, error) {
 	}
 
 	if err := createGitIgnore(filepath.Dir(app.SiteLocalSettingsPath), settings.SiteSettingsLocal); err != nil {
-		output.UserOut.Warnf("Failed to write .gitignore in %s: %v", app.SiteLocalSettingsPath, err)
+		output.UserOut.Warnf("Failed to write .gitignore in %s: %v", filepath.Dir(app.SiteLocalSettingsPath), err)
 	}
 
 	return app.SiteLocalSettingsPath, nil
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -54,7 +54,7 @@ const backdropMainSettingsTemplate = `<?php
 /**
  This was automatically generated to include settings managed by ddev.
  */
-include {{ '$config.SiteSettingsLocal }}';
+include '{{ $config.SiteSettingsLocal }}';
 `
 
 const backdropSettingsAppendTemplate = `{{ $config := . }}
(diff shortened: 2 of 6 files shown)


andrewfrench:
Opening this to get some initial build-and-test coverage and opening the door for sanity-check reviews, but it has at least a little work to be done still before a formal review.

rickmanelius:
This one is a beast of an issue to review, so I'm going to have to block at least an hour to go through the use cases. I'm set for tomorrow morning and, hopefully, I can give you feedback by noon. Thanks for all the work and providing a series of testing instructions, @andrewfrench!

One thing I will note right out of the gates is that there are no documentation changes with this PR. Given the extent of this change, I'm anticipating there will be potential support requests and such.

pkg/ddevapp/drupal.go
```diff
@@ -182,79 +210,153 @@ $databases['default']['default'] = array(
...
+	if fileutil.FileExists(app.SiteLocalSettingsPath) {
+		output.UserOut.Printf("ddev settings file %s already exists", drupalConfig.SiteSettingsLocal)
+	} else {
+		output.UserOut.Printf("No %s exists, creating one", drupalConfig.SiteSettingsLocal)
+		if err := writeDrupal7SettingsFile(drupalConfig, app.SiteLocalSettingsPath); err != nil {
+			return "", fmt.Errorf("failed to write Drupal settings file %s: %v", app.SiteLocalSettingsPath, err)
+		}
+	}
+
+	if err := createGitIgnore(filepath.Dir(app.SiteLocalSettingsPath), drupalConfig.SiteSettingsLocal); err != nil {
```
rfay: Creating a gitignore here seems like a pretty reasonable approach. I'm just slightly torn about intruding in their world, but think it's probably the exact right approach.

pkg/ddevapp/drupal.go
```diff
@@ -182,79 +210,153 @@ $databases['default']['default'] = array(
...
+		output.UserOut.Warnf("Failed to write .gitignore: %v", err)
```
rfay: I don't think this will happen, right? We've already done the permission setting on sites/default before this? Or maybe not, this is at config time and we do that at start time. But anyway, please improve the message to say where the gitignore creation failed (full path of failed creation) (And of course this pattern is repeated elsewhere.)

pkg/ddevapp/drupal.go
```diff
@@ -522,3 +622,55 @@ func createDrupal8SyncDir(app *DdevApp) error {
...
+	}
+
+	b := bytes.NewBuffer([]byte{})
+	if err := tmpl.Execute(b, drupalConfig); err != nil {
+		return err
+	}
+
+	if _, err := file.Write(b.Bytes()); err != nil {
+		return err
+	}
```
rfay: Silly me, I had envisioned just an append to the file using perhaps unix tools, but that certainly wouldn't work. But wouldn't `f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0644)` work?

I do understand that we might be creating as opposed to appending, but seems like a large-hammer approach to appending.

pkg/ddevapp/settings_test.go
```diff
@@ -8,19 +8,30 @@ import (
...
 	"github.com/drud/ddev/pkg/util"
 	"github.com/stretchr/testify/assert"
 )
 
+var appTypeSettingsLocations = map[string][]string{
+	"drupal6":  {"sites/default/settings.php", "sites/default/settings.ddev.php"},
+	"drupal7":  {"sites/default/settings.php", "sites/default/settings.ddev.php"},
+	"drupal8":  {"sites/default/settings.php", "sites/default/settings.ddev.php"},
+	"backdrop": {"settings.php", "settings.ddev.php"},
+}
```
rfay: No need to change this, but it would be a lot more flexible if the members were named as opposed to a positional array. But we can do that when we come back and need to change it. Positional arrays are pretty lousy for future expansion, using numeric indexes that are magical and hard to maintain.

pkg/ddevapp/drupal.go
```diff
@@ -182,79 +210,153 @@ $databases['default']['default'] = array(
...
-	err = writeDrupal6SettingsFile(drupalConfig, settingsFilePath)
-	if err != nil {
-		return settingsFilePath, fmt.Errorf("Failed to write Drupal settings file: %v", err.Error())
+	if err := manageDrupalCommonSettingsFile(app, drupalConfig); err != nil {
+		return "", err
 	}
 
-	return settingsFilePath, nil
+	// Create settings.ddev.php if one doesn't already exist
+	if fileutil.FileExists(app.SiteLocalSettingsPath) {
```
rfay: Eventually we'll hate the repeated nature of these things and the fact that we have to maintain the same thing in multiple places  that we'll consolidate with functions (or somehow).  Eventually.

pkg/ddevapp/drupal.go
```diff
@@ -182,79 +210,153 @@ $databases['default']['default'] = array(
...
+		output.UserOut.Printf("ddev settings file %s already exists", drupalConfig.SiteSettingsLocal)
+	} else {
+		output.UserOut.Printf("No %s exists, creating one", drupalConfig.SiteSettingsLocal)
+		if err := writeDrupal6SettingsFile(drupalConfig, app.SiteLocalSettingsPath); err != nil {
+			return "", fmt.Errorf("failed to write Drupal settings file: %v", err)
```
rfay: This is failing to write the settings.ddev.local, right? Probably the message should be slightly improved; if the err doesn't have the full path info, please add that to the message.

pkg/ddevapp/settings_test.go
```diff
@@ -82,3 +93,174 @@ func TestWriteDrushConfig(t *testing.T) {
 	err = os.RemoveAll(dir)
 	assert.NoError(t, err)
 }
+
+// TestIncludeSettingsDdevInNewSettingsFile verifies that when no settings.php file exists,
+// a settings.php file is created that includes settings.ddev.php.
+func TestIncludeSettingsDdevInNewSettingsFile(t *testing.T) {
+	dir := testcommon.CreateTmpDir("")
+	err := os.MkdirAll(filepath.Join(dir, "sites/default"), 0777)
```
rfay: This doesn't seem to take the docroot into account? This should be docroot + sites/default.

Also, this is drupal only right? Probably should update the test name for that.

pkg/ddevapp/settings_test.go
```diff
@@ -82,3 +93,174 @@ func TestWriteDrushConfig(t *testing.T) {
 	err = os.RemoveAll(dir)
 	assert.NoError(t, err)
 }
+
+// TestIncludeSettingsDdevInNewSettingsFile verifies that when no settings.php file exists,
+// a settings.php file is created that includes settings.ddev.php.
+func TestIncludeSettingsDdevInNewSettingsFile(t *testing.T) {
+	dir := testcommon.CreateTmpDir("")
```
rfay: IMO it's a little better to give a prefix here, like `dir := testcommon.CreateTmpDir("TestIncludeSettingsDdevInNewSettingsFile")`; it doesn't affect functionality or improve the test, but makes debugging (and possibly cleanup) easier and more obvious.

pkg/ddevapp/settings_test.go
```diff
@@ -82,3 +93,174 @@ func TestWriteDrushConfig(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	app, err := NewApp(dir, DefaultProviderName)
+	assert.NoError(t, err)
+
+	for appType, relativeSettingsLocations := range appTypeSettingsLocations {
+		app.Type = appType
+
+		relativeSettingsLocation := relativeSettingsLocations[0]
+		relativeSettingsDdevLocation := relativeSettingsLocations[1]
```
rfay: This is the magic positional params I was talking about earlier, hard to read and understand. Again, no need to change, especially in this very simple situation. I just overall prefer approaches that give a named parameter (map, struct, whatever)

pkg/ddevapp/settings_test.go
```diff
@@ -82,3 +93,174 @@ func TestWriteDrushConfig(t *testing.T) {
...
+		expectedSettingsLocation := filepath.Join(dir, relativeSettingsLocation)
+		expectedSettingsDdevLocation := filepath.Join(dir, relativeSettingsDdevLocation)
+
+		// Ensure that no settings.php exists
+		os.Remove(expectedSettingsLocation)
```
rfay: These os.Remove() are  definitely going to get gometalinter complaints when you get here, err = and assert.NoError(), or _ =?

pkg/ddevapp/settings_test.go
```diff
@@ -82,3 +93,174 @@ func TestWriteDrushConfig(t *testing.T) {
...
+		assert.True(t, newSettingsIncludesSettingsDdev)
+
+		// Ensure that settings.ddev.php exists
+		assert.True(t, fileutil.FileExists(expectedSettingsDdevLocation))
+	}
+}
+
+// TestIncludeSettingsDdevInExistingSettingsFile verifies that when a settings.php file already exists,
+// it is modified to include settings.ddev.php
+func TestIncludeSettingsDdevInExistingSettingsFile(t *testing.T) {
```
rfay: Thanks for thinking through the testing on these various situations!

pkg/ddevapp/utils.go
```diff
@@ -164,3 +168,48 @@ func ddevContainersRunning() (bool, error) {
 	}
 	return false, nil
 }
+
+// getTemplateFuncMap will return a map of useful template functions.
+func getTemplateFuncMap() map[string]interface{} {
```
rfay: /me likes maps for this kind of thing :)

rfay:
Just a question to ask: Should we delete the settings.local.php if it's #ddev-generated ?

rickmanelius:
@andrewfrench Hi! I'm planning on reviewing this afternoon, but I wasn't sure if, based on any feedback from @rfay, you need/want me to hold off. I'm more interested in end-user facing items, so if it's behind the scenes items then I won't consider myself blocked in my review.

andrewfrench:
There are a couple important logical flow things that I see in these comments that I want to address immediately this morning, the rest are behind-the-scenes and would only impact the source itself or the comments in the generated files. As of now, I'd say consider yourself unblocked, but if things go sideways I'll update you within the hour.

pkg/ddevapp/backdrop.go
```diff
@@ -6,43 +6,69 @@ import (
...
 
-// Note that this template will almost always be used for settings.local.php
-// because Backdrop ships with it's own default settings.php.
-const backdropTemplate = `<?php
+// backdropMainSettingsTemplate defines the template that will become settings.php in
+// the event that one does not already exist.
+const backdropMainSettingsTemplate = `<?php
+{{ $config := . }}
+/**
+ This was automatically generated to include settings managed by ddev.
```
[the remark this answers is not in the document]
andrewfrench: This is a point I've been considering over the weekend.  It's certainly true that we're generating some content inside the settings file, but I was hesitant to output the signature as it's used as a signal to determine ownership and management of settings files elsewhere/previously.  I assume we agree that we can't consider settings.php _managed_ by ddev after these changes, but I don't think there'd be harm in including the signature here with the updated logic (unless a user downgrades ddev versions in the future).

pkg/ddevapp/settings_test.go
andrewfrench: Interesting, gometalinter hasn't complained about these.  I've updated to `_ =` anyway.

pkg/ddevapp/drupal.go
andrewfrench: This got a little logically polluted from the an earlier revision of this function that used file.WriteAt().  There were a couple refactor/flow improvement opportunities here that I took advantage of, it should now be clearer.

rickmanelius:
Ok. So I've reviewed the PR summary and I'm ready to begin testings. However, there are still two things that are worth calling out:
* Documentation: This PR doesn't add any (yet) and the OP specifically called out the fact that part of the problem was that we were not communicating this to the end-user strongly enough.
* The OP had a WP related item by @alkymst. I'm not seeing that, but I'm not sure if there is another PR that addresses it. We don't need to solve this here, but we should have a separate issue filed to track it IMHO.

The only bullet blocking an approval will be item 1. The second one is more of a request so we don't lose that work/goal when the PR is merged and the OP is closed.

rfay:
@rickmanelius 
* I do think we should document the new behavior, people need to know what settings.ddev.php is.
* Mostly the output of ddev itself when creating the files is the best documentation though. Release notes need quite a bit of focus on it. 
* the Wordpress fix is still a PR in progress, it may not be high enough priority for @alkymst any more. It's not actually part of this PR IMO.

pkg/ddevapp/backdrop.go
```diff
@@ -62,50 +88,102 @@ ini_set('session.gc_maxlifetime', 200000);
...
-// adding things like database host, name, and password.
-// Returns the full path to the settings file and err.
+// createBackdropSettingsFile manages creation and modification of settings.php and settings.ddev.php.
+// If a settings.php file already exists, it will be modified to ensure that it includes
+// settings.ddev.php, which contains ddev-specific configuration.
 func createBackdropSettingsFile(app *DdevApp) (string, error) {
-	settingsFilePath, err := app.DetermineSettingsPathLocation()
+	settings := NewBackdropSettings()
+
+	if !fileutil.FileExists(app.SiteSettingsPath) {
```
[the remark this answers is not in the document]
andrewfrench: I agree -- this is checking existence of the "main" settings.php file and writing that if one doesn't exist.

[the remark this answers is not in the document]
rfay: I was saying that perhaps it should check #ddev-generated and if it *is* always recreate. That lets us future-proof. Although I certainly understand that if we were making changes we would put them into the settings.ddev.php.

rickmanelius:
* Backdrop
  * Scenario A + B
  	* Basically not possible because.
  	  1. Stock Backdrop always includes settings.php in the repo root.
  	  2. There is no ability to create a separate .gitignore for settings.ddev.php
  * Scenario C
    * Observed the output
      * "Existing settings.php file does not include settings.ddev.php, modifying to include ddev settings"
    * Verified that settings.ddev.php was regenerated
    * Verified that settings.php now links to settings.ddev.php

pkg/ddevapp/backdrop.go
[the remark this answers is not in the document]
andrewfrench: If I understand this correctly, I think this means we'd need two signatures (as @rickmanelius pointed out in his review) because this currently writes the `#ddev-generated` signature to the appended portion of settings.php, so it would nearly always be present in the file.  I'm not against that at all, as it does clearly separate ownership vs. modification.

andrewfrench:
Thanks @rickmanelius! I agree that this should be blocked until docs are in.  It sounds like this will need an update to write our .gitignore if there is an existing, but empty, .gitignore file. I'll update this.

pkg/ddevapp/backdrop.go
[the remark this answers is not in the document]
rfay: Yeah, we should *not* be writing any signature when *appending* to a user settings.php, IMO.

andrewfrench:
@rickmanelius, you'll find updated settings management documentation in [`docs/users/cli-usage.md`](https://github.com/andrewfrench/ddev/blob/2018-07-10_settings-files/docs/users/cli-usage.md).  I've gone through the settings management code and tried to reflect the current state of what each app type will do when writing settings files, as the existing explanation was very Drupal-heavy and didn't include other app type processes.

docs/users/cli-usage.md
```diff
@@ -264,7 +249,29 @@ Your project can be reached at: http://drupal8.ddev.local
...
 
-_Please note that if you're providing the settings.php or wp-config.php and ddev is creating the settings.local.php (or wordpress wp-config-local.php), the main settings file must explicitly include the appropriate "settings.local.php" or equivalent._
+### Configuration files
+_**Note:** If you're providing the settings.php or wp-config.php and DDEV is creating the settings.ddev.php (or wp-config-local.php, AdditionalConfig.php, or similar), the main settings file must explicitly include the appropriate DDEV-generated settings file._
+
+The `ddev config` command attempts to create a CMS-specific settings file with DDEV credentials pre-populated.
+
+For **Drupal** and **Backdrop**, DDEV settings are written to a DDEV-managed file, settings.ddev.php. The `ddev config` command will ensure that these settings are included in your settings.php through the following steps:
+
+* Write DDEV settings to settings.ddev.php
```
rickmanelius: Small nit in that it doesn't just write, but it forces it to factory default each time.

rickmanelius:
## Preparation

* https://github.com/drud/ddev/pull/990
* `git checkout 250e9084`
* `make darwin`
* `ddev version` => v0.20.0-71-g250e9084

## Testing

There are 3 scenarios across 4 CMSs

Scenarios:

A. No settings.php, settings.ddev.php, .gitignore
B. Settings.php with no include or existence of settings.ddev.php and no .gitignore
C. All 3 exist

The 4 CMSs to test are Drupal 6, 7, 8, and Backdrop.

I'll start with Drupal 7 and walk through the scenarios before switching to other CMS flavors.

* Drupal 7
  * Scenario A
    * `ddev config --projectname drupal7 --projecttype drupal7 --docroot .`
    * I observed that the output states:
	  * "No settings.php file exists, creating one"
  	  * "Existing settings.php file includes settings.ddev.php"
  	* I verified the generated files are in place at sites/default/
    * Cleaned up via `rm settings.php settings.ddev.php .gitignore`
  * Scenario B
  	* Created a mock settings with `cp default.settings.php settings.php`
    * Re-ran `ddev config --projectname drupal7 --projecttype drupal7 --docroot .`
    * I observed that the output states:
      * "Existing settings.php file does not include settings.ddev.php, modifying to include ddev settings"
    * I verified the creation of .gitignore and settings.ddev.php
    * I verified that settings.php has a #ddev message in the comment.
  * Scenario C
    * Preparation: Opened and blanked out all 3 files
    * Re-ran `ddev config --projectname drupal7 --projecttype drupal7 --docroot .`
    * I observed that the output states:
      * "Existing settings.php file does not include settings.ddev.php, modifying to include ddev settings"
    * I opened all 3 files and verified:
      * settings.php has the appropriate include
        * NOTE. Since the file was blanked out, the opening <?php wasn't there. Not likely to be an issue.
      * settings.ddev has been refreshed.
      * .gitignore will only populate when when it's blank, which is fine.

Looks like everything behind the scenes was addressed. I also see the improvement of switching from "#ddev-generated" to a more accurate comment of "// Automatically generated include for settings managed by ddev."

Documentation changes look good. One small nit, but nothing I would block the release on.

PR #991, opened 2018-07-14
Rename web and db containers (nginx-php-fpm-local=>ddev-webserver, mariadb-local=>ddev-dbserver)

rfay:
## The Problem/Issue/Bug:

Our naming of the web and db containers has always been a little too much technology-specific (for example, we changed from mysql to mariadb; and we're likely to ship apache2 on the web container soon). 

However, in the upcoming v1.0.0 release, web would actually have a *tag conflict*  on v1.0.0 and some following as well for nginx-php-fpm-local, so it only makes sense to just rename.

Rename nginx-php-fpm-local => ddev-webserver
Rename mariadb-local => ddev-dbserver

Run a project. Verify that it uses the new containers.

## Release/Deployment notes:

I verified that ddev-router will not have any version conflicts, and didn't change the name of phpmyadmin.
I pushed the initial tag for each of these to dockerhub with the new repo name but the exact same tag and content (from master).


rfay:
I'm OK with changing it; it's such a backwater that it seems hardly to matter. I don't really like the use of 'dba' because I always stumble on it thinking it's the 'db'. Got a suggestion for a more obvious name?

andrewfrench:
I was thinking the same, dbaserver could be confusingly similar to dbserver. Is ddev-dbadminserver better? Could be a little long. Ultimately, I agree that it's not a big deal either way.

rfay:
The difference between this and the others is that this isn't even a server, just a php app, and it's nothing but phpmyadmin.  And I'm not completely sure anybody likes it.

rickmanelius:
Until we have the ability for (optional) user statistics, it's unclear if removing a feature like phpmyadmin would be well received. After all, there are several instances where our decision to add a shortcut to sequelpro was very well received.

What would probably be useful (issue needs to be filed) is have the ability to pick from a menu of default additional services, such as phpmyadmin, Solr, Varnish, Memcached, etc. If this approach has merit/buy-in, I'd vote we just leave the container label as "dba" until we do that. Alternatively, "tools" might be reasonable if we want to have this contain (no pun intended) optional dev tools like phpmyadmin and mailhog.

rfay:
That was just random talk about removal, we certainly wouldn't do it here. 

The name of the image (and the code directory) is and always has been (at least for a year or more?) drud/phpmyadmin, Andrew's suggestion was about *changing* it from "phpmyadmin" to something like "dbaserver". 

IMO that wouldn't be helpful at this time, as "dba" is not a very good description, and there's no particular reason to fiddle with it, since we never even look at it. If somebody has a great idea what to change it to, I'd be ok with it, but it doesn't seem very worthy to me. Changing the other two names is quite good and helpful.

rfay:
I'll pull this today if we think the conversation about the name of the phpmyadmin container is adequately resolved. (I think we should leave it as is until we think of something better.)

Issue #677, opened 2018-02-24
Add in-depth testing of sites to ddev tests


rfay:
## What happened (or feature request):

In our tests we have a variety of CMSs that we install and start ddev.... but we don't actually do anything with the sites after starting, just test ddev functions like import-db and the like. For real test coverage we need to verify:

- [ ] Hitting the front page works
- [ ] Logging in works
- [ ] Visiting some important URLs works
- [ ] Imagecache/image styles or the equivalent work properly

We'll need to add to this list as we go.

cweagans:
I've been using https://github.com/headzoo/surf on a personal project with a decent amount of success. Might be useful here.

rickmanelius:
I'm not sure if this was already addressed in #671. Seems useful and will pin to v1.2.0 as a milestone related to CMS support.

rfay:
No, this was a spin-off from #671  - We've never had any nontrivial tests for the test sites we install, and it will really improve our coverage and detection when we add them.

rfay:
These should probably be spun off into separate tickets.

When doing Wordpress, please work closely with @alkymst to do something reasonable there.

rfay:
So the way I'm thinking about this now is pretty easy to implement.

1. After https://github.com/drud/ddev/pull/1151 lands, it adds a URI+Expectation.
2. Add a non-static expectation URL (or array of expectations?) to the TestSite type
3. In TestDdevImportDB, test appropriate expected URLs for each project.

I think that's all we really have to do. Not very ambitious at all.

PR #1009, opened 2018-07-23
Add CMS interaction to tests

andrewfrench:
## The Problem/Issue/Bug:
Our tests previously didn't interact with the CMS applications being deployed to confirm they actually worked.

This PR provides Interactor, an interface that provides methods to test interactivity of an app using the basic points outlined in #677. 

The vast majority of changes here are updates to vendor/ in order to use [Surf](https://github.com/headzoo/surf), a programmatic browser for Go that handles interaction with a web page and is super cool.

Examples of use are found in start_test.go.


andrewfrench:
Opening this PR to get some early opinions and test coverage. Very much a work in progress.

pkg/testinteraction/common.go
```diff
@@ -0,0 +1,74 @@
+package testinteraction
```
rfay: I would have expected to see this in testcommon, but don't have any objection if this is where you want it.

pkg/testinteraction/common.go
```diff
@@ -0,0 +1,74 @@
...
+	"regexp"
+
+	"github.com/drud/ddev/pkg/exec"
+	"github.com/headzoo/surf/browser"
+)
+
+// commonInteraction holds values that all interaction structs will need, allowing for common
+// methods to be defined.
+type commonInteraction struct {
+	adminUsername string
```
rfay: I'm not all the way through this yet, but is it important that it be an adminUsername or just a provided and valid user/pass?

andrewfrench:
This is absolutely intended for use by ordinary tests, it's just packaged independently (which is definitely not required). The idea is that each app type will support the same general steps (configure, install, login, visit), but each will have their own workflow and criteria, so each should implement the specifics on its own.

I'm going to close this for now as full test coverage uncovered some issues that might take a good amount of work to resolve, but I expect this to resurface in one form or another sometime soon.

PR #1014, opened 2018-07-25
Fix ddev exec message offered to wp users on import

rfay:
## The Problem/Issue/Bug:

A user in slack pointed out that our prompted suggestion of wp search-replace results in an error because we quote it.

The prompt says

> Wordpress sites require a search/replace of the database when the URL is changed. You can run "ddev exec 'wp search-replace [http://www.myproductionsite.example] http://wpsite.ddev.local'" to update the URLs across your database. For more information, see http://wp-cli.org/commands/search-replace/

The result if you do that is:

`ddev exec 'Failed to execute command [wp search-replace [http://www.myproductionsite.exampleto update the URLs across your da] http://wpsite.ddev.local]: Failed to run docker-compose [-f /Users/brentr
obbins/wpsite/.ddev/docker-compose.yaml exec -T web wp search-repl
ace [http://www.myproductionsite.example] http://wpsite.ddev.local], err='exit status 126', stdout='OCI runtime exec failed: exec failed: container_linux.go
:348: starting container process caused "exec: \"wp search-replace [http://www.my
productionsite.example] http://wpsite.ddev.local\": stat wp search-replace
[http://www.myproductionsite.example] http://wpsite.ddev.local: no such file or directory": unknown
', stderr=''`

Because ddev exec is trying to run an executable called "wp search-replace ..." instead of just running wp. 

Remove the quotes around the command.

### DIFF
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -189,7 +189,7 @@ func getBackdropUploadDir(app *DdevApp) string {
 func getBackdropHooks() []byte {
 	backdropHooks := `
 #  post-import-db:
-#    - exec: "drush cc all"`
+#    - exec: drush cc all`
 	return []byte(backdropHooks)
 }
 
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -394,7 +394,7 @@ func TestWriteConfig(t *testing.T) {
 
 	out, err = ioutil.ReadFile(filepath.Join(testDir, "config.yaml"))
 	assert.NoError(err)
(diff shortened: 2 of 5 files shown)


andrewfrench:
A project-wide search for `exec: "` shows this incorrect pattern being generated and suggested in many places:
```
grep -R --exclude-dir .gotmp --exclude-dir bin --exclude-dir vendor 'exec: "' .
```
Could these be updated as well?

rfay:
Interestingly enough, the command to exec *works* inside quotes when used as a hook, and in some cases the quotes improve things, I can't give an example though. I think we're better to offer it without quotes where it doesn't need them.

PR #1008, opened 2018-07-23
Install php-redis module in the web container

tmotyl:
## The Problem/Issue/Bug:
This PR adds php-redis module to the web container.

## Manual Testing Instructions:

check in phpInfo if the redis module is installed, or check through php code `extension_loaded('redis');`

## Related Issue Link(s):

Resolves: #267

## Release/Deployment notes:

nothing to be done.

Issue #267, opened 2017-05-24
Create a service configuration for redis


tannerjfco:
## What happened (or feature request):
Feature Request
In #121 we determined redis was one of several popular services that can be a requirement for more complex applications. In order for ddev to facilitate local environments for these more complex applications, we should provide an optional service configuration for redis.

## What you expected to happen:
A docker-compose configuration should be made, named `docker-compose.redis.yml`. This configuration should:

- Determine if an existing redis docker image can be used, or if we need to develop / maintain our own.
- Define a service named redis that provides a redis container.
- Where possible, pre-configure the redis container to reduce required setup for a user.
- Document any steps that are necessary after adding the service file to .ddev to be able to use redis with a Drupal site.

The [example solr service in documentation](https://ddev.readthedocs.io/en/latest/users/customization-extendibility/#defining-an-additonal-service-apache-solr-example) is a good reference for how this service file should work.

## Related source links or issues:

rickmanelius:
Closing for now until this becomes an end-user request, at which time we can re-open.

tmotyl:
I would love to have redis availavle in ddev. This includes
- adding php extension redis to the web container
- providing example redis service configuration in documentation
This is what I'm playing around with right now:
```
# ddev redis recipe file
#

version: '1'

services:
  redis: # This is the service name used when running ddev commands accepting the --service flag
    container_name: ddev-${DDEV_SITENAME}-redis # This is the name of the container. It is recommended to follow the same name convention used in the main docker-compose.yml file.
    image: redis:4
    restart: always
    ports:
      - 6379 # Redis is served from this port inside the container
    labels:
    # These labels ensure this service is discoverable by ddev
      com.ddev.site-name: ${DDEV_SITENAME}
      com.ddev.approot: $DDEV_APPROOT
      com.ddev.app-url: $DDEV_URL
    environment:
      - VIRTUAL_HOST=$DDEV_HOSTNAME # This defines the host name the service should be accessible from. This will be sitename.ddev.local
      - HTTP_EXPOSE=6379 # This defines the port the service should be accessible from at sitename.ddev.local
    volumes: []
# This links the redis service to the web service defined in the main docker-compose.yml, allowing applications running
# in the web service to access the redis service at sitename.ddev.local:6379
  web:
    links:
      - redis:$DDEV_HOSTNAME

```

rickmanelius:
We have a live request. Re-opening...

tmotyl:
[...]

PR #1015, opened 2018-07-26
Revert "Install php-redis module in the web container (#1008), for #267"

rfay:
Although #1008 did in fact pass all tests and was able to build
containers, I'm completely unable to get a successful container build
whether using the code here or just adding an apt-get install -y php-redis

I suspect something may have changed in the upstream PHP repo.

The failure breaks nightly builds and all container builds.

This reverts commit f6784fe02d9ba02d2a96fc0ab96939e2b538a924.


rfay:
This was not the problem, closing.

PR #1016, opened 2018-07-26
Improve buildkite test_containers.sh by not erroring if docker rm/rmi --fail

rfay:
## The Problem/Issue/Bug:

* There seem to be times on the buildkite testbots where docker rm or docker rmi can fail... but we're only doing it in the context of cleaning up anyway, so it shouldn't cause the tests to fail.
* Docker images should be built in testing with --no-cache. Turns out build-tools has a DOCKER_ARGS makefile variable for this exact purpose.

* Ignore failure on docker rm.
* Add --no-cache to container build in circle and buildkite tests.

### DIFF
--- a/.buildkite/test_containers.sh
+++ b/.buildkite/test_containers.sh
@@ -38,7 +38,7 @@ set -x
 for dir in containers/*
     do pushd $dir
     echo "--- Build container $dir"
-    time make container
+    time make container DOCKER_ARGS=--no-cache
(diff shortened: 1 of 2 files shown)


rfay:
The premise here was false, I was not analyzing the behavior of the failures very well. However I think the script improvements do no harm, and certainly don't affect the end-user. 

The big argument is to use BATS to build readable tests on our container tests with parseable test results.

PR #1017, opened 2018-07-26
Clean up hostnames when removing project data, closes #831

### LINKED ISSUE #831: Clean up /etc/hosts on ddev remove (remove unused hostnames)
## Feature request:
Please add some automatic cleanup to remove unused hostnames from `/etc/hosts`.
When you use a lot of projects or test a lot, the `/etc/hosts` gets cluttered with all the hostnames.
It would be great if `ddev remove --remove-data` would also remove the hostname from `/etc/hosts`.
Or maybe you want to introduce another flag that triggers the removal of unused hostnames.

andrewfrench:
## The Problem/Issue/Bug:
#831: The user's hosts file eventually becomes polluted after having many project hostnames added to it. Currently, ddev makes no effort to remove a hostname entry after it has been added.

Add the ability to remove a hostname via the `ddev hostname` command:
```
ddev hostname --remove my.host 127.0.0.1 
```

Remove hostname entries from the user's hosts file on `ddev remove --remove-data`.  Removing a hostname on `ddev remove --remove-data` and `ddev hostname --remove` mimic the output and structure of existing hostname addition code.

User-facing output in `ddev hostname` has been improved.

- Add a hostname via `ddev hostname test.host 1.2.3.4` and ensure the hostname has been added to the hosts file
- Remove the same hostname via `ddev hostname test.host 1.2.3.4 --remove` and ensure the hostname has been removed from the hosts file

Test the above steps with and without `sudo` and the `--json-output` to inspect output and error logging.

- Add a new and unique FQDN or hostname to a ddev project's .ddev/config.yaml
- Execute `ddev start` on the project and ensure the new hostnames have been added to the hosts files
- Execute `ddev remove --remove-data` on the project and ensure the project's hostnames have been removed from the hosts file

Test the above steps with and without `DRUD_NONINTERACTIVE` to see non-interactive instruction logging.

### DIFF
--- a/cmd/ddev/cmd/hostname.go
+++ b/cmd/ddev/cmd/hostname.go
@@ -162,8 +162,8 @@ func removeInactiveHostnames(hosts goodhosts.Hosts) {
 	var detail string
 	rawResult := make(map[string]interface{})
 
-	// Get the list active hosts names to preserve, always including localhost
-	activeHostNames := map[string]bool{"localhost": true}
+	// Get the list active hosts names to preserve
+	activeHostNames := make(map[string]bool)
 	for _, app := range ddevapp.GetApps() {
 		for _, h := range app.GetHostnames() {
 			activeHostNames[h] = true
@@ -196,7 +196,7 @@ func removeInactiveHostnames(hosts goodhosts.Hosts) {
 					continue
 				}
 
-				// Silently ignore those that may not be ddev-managed to not spam the user's terminal
+				// Silently ignore those that may not be ddev-managed
 				if !strings.HasSuffix(h, version.DDevTLD) {
 					continue
 				}
@@ -223,11 +223,6 @@ func removeInactiveHostnames(hosts goodhosts.Hosts) {
 		rawResult["full_error"] = detail
 		output.UserOut.WithField("raw", rawResult).Fatal(detail)
 	}
-
-	detail = "Cleaned up inactive host names"
-	rawResult["error"] = "SUCCESS"
-	rawResult["detail"] = detail
-	output.UserOut.WithField("raw", rawResult).Info(detail)
 
 	return
 }
--- a/cmd/ddev/cmd/hostname.go
+++ b/cmd/ddev/cmd/hostname.go
@@ -25,10 +25,11 @@ var HostNameCmd = &cobra.Command{
 
 		hosts, err := goodhosts.NewHosts()
 		if err != nil {
+			detail := fmt.Sprintf("Could not open hosts file for reading: %v", err)
 			rawResult := make(map[string]interface{})
 			rawResult["error"] = "READERROR"
-			rawResult["full_error"] = fmt.Sprintf("%v", err)
-			output.UserOut.WithField("raw", rawResult).Fatal(fmt.Sprintf("could not open hosts file for read: %v", err))
+			rawResult["full_error"] = detail
+			output.UserOut.WithField("raw", rawResult).Fatal(detail)
 
 			return
 		}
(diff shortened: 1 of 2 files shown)


rfay:
Quick glance looks fantastic. How about a --all which would only remove those that match our default domain (and maybe that aren't currently active? or could only be executed when nothing is running?)

rickmanelius:
Wow... as a side comment, thank you for such thoroughness on the coverage as well as the testing instructions.

andrewfrench:
- Agreed, I'll add this here.
- I disallowed `ddev remove --remove-data --all` because that seemed to be an exceptionally destructive command for users that are doing real work with ddev.  I'm willing to revisit this, but giving a user the ability to blow away everything seemed dangerous.
- I can see that becoming annoying.  I specifically put this under the `--remove-data` option because I'm an avid `ddev rm` user and didn't want to constantly type my password for sudo, either.

rickmanelius:
Thanks for the follow-ups. Everything you said makes sense and I'm +1

rfay:
First thing I wanted to do was use --fire-bazooka (and thanks for adding that). But:

```
$ ddev hostname --remove-inactive
Hostname localhost at 127.0.0.1 is active, preserving
Removed hostname randyfay.ddev.local at 127.0.0.1
Removed hostname d8git.ddev.local at 127.0.0.1
Removed hostname one.ddev.local at 127.0.0.1
Removed hostname two.ddev.local at 127.0.0.1
Removed hostname testoverride.ddev.local at 127.0.0.1
Removed hostname a.b.c.ddev.local at 127.0.0.1
Removed hostname coe.ddev.local at 127.0.0.1
Removed hostname typo3-v9-composer-from-instructions.ddev.local at 127.0.0.1
Removed hostname d8composer.ddev.local at 127.0.0.1
Removed hostname peakhrconsulting.com.ddev.local at 127.0.0.1
Removed hostname rfay-drupal8.ddev.local at 127.0.0.1
Removed hostname ddevt3dd18.ddev.local at 127.0.0.1
Removed hostname typo3-v9-composer.ddev.local at 127.0.0.1
Removed hostname satirtraining.ddev.local at 127.0.0.1
Removed hostname junker999.ddev.local at 127.0.0.1
Removed hostname d7git.ddev.local at 127.0.0.1
Removed hostname wordpress.ddev.local at 127.0.0.1
Removed hostname d8incontainer.ddev.local at 127.0.0.1
Removed hostname gchistory.ddev.local at 127.0.0.1
Removed hostname typo3-v8-composer.ddev.local at 127.0.0.1
Removed hostname junk.ddev.local at 127.0.0.1
Removed hostname junk.d8git.com at 127.0.0.1
Removed hostname mysite.com at 127.0.0.1
Removed hostname somesite.yoursite.com at 127.0.0.1
Removed hostname anothersite.yoursite.com at 127.0.0.1
Removed hostname junk.com at 127.0.0.1
Removed hostname drupal-kickstart.ddev.local at 127.0.0.1
Removed hostname drupal8.ddev.local at 127.0.0.1
Removed hostname backdrop.ddev.local at 127.0.0.1
Removed hostname d6git.ddev.local at 127.0.0.1
Removed hostname d7-kickstart.ddev.local at 127.0.0.1
Removed hostname wordpress4.ddev.local at 127.0.0.1
Removed hostname d8-umami.ddev.local at 127.0.0.1
Removed hostname junktypo3.ddev.local at 127.0.0.1
Removed hostname t3git.ddev.local at 127.0.0.1
Removed hostname testpkgdrupal7.ddev.local at 127.0.0.1
Removed hostname testvalidtestsitewordpress.ddev.local at 127.0.0.1
Could not write hosts file: open /etc/hosts: permission denied
```

Note that it shows many successes, but nothing actually worked (as it tells us at the very end)

I'd hate to have `ddev hostname` have to sudo itself; maybe it can just check for writeablility before taking action (using goodhosts doing a no-op) and if it can't do it, suggest the correct command?

Also, remember that this is a really serious problem for windows users, so we may have to help them along the way on it.

FInally, really important: If we don't have a reason to believe ddev put the hostname in there, we shouldn't remove it (That's why I was suggesting filtering on ddev.local - or the configured TLD.) 

I saw this:
```
Removed hostname something.com at 127.0.0.1
Removed hostname somethingelse.com at 127.0.0.1
```
and realized that ddev could have cleaned up some things it absolutely didn't own.

andrewfrench:
Good points @rfay, this will now only remove `ddev.local` domains (at least until we can get a persistent record of all projects to work from) and attempts a trivial write _first_ to confirm permissions.

cmd/ddev/cmd/hostname.go
```diff
@@ -5,58 +5,235 @@ import (
...
+			rawResult["full_error"] = detail
+			output.UserOut.WithField("raw", rawResult).Fatal(detail)
+
+			return
+		}
+
+		// Attempt to write the hosts file first to catch any permissions issues early
+		if err := hosts.Flush(); err != nil {
+			rawResult := make(map[string]interface{})
+			detail := fmt.Sprintf("Could not write hosts file: %v", err)
```
rfay: Let's just prompt them with what to do about it, 
> Please use "sudo hostname <args they used>" or execute with administrative privileges.

cmd/ddev/cmd/hostname.go
```diff
@@ -5,58 +5,235 @@ import (
...
+			return
 		}
+
+		// By default, add a host name
+		addHostname(hosts, ip, hostname)
 	},
 }
 
+// addHostname encapsulates the logic of adding a hostname to the system's hosts file.
+func addHostname(hosts goodhosts.Hosts, ip, hostname string) {
```
rfay: Seems like a relic of long ago that all this has remained in cmd instead of getting refactored into pkg. OK with me, and this is a very, very isolated command.

pkg/ddevapp/ddevapp.go
```diff
@@ -887,13 +887,18 @@ func (app *DdevApp) Down(removeData bool) error {
 		return fmt.Errorf("failed to remove %s: %v", app.GetName(), err)
 	}
 
-	// Remove data/database if we need to.
+	// Remove data/database/hostname if we need to.
 	if removeData {
```
rfay: I think it's worth just a small conversation on whether we should just do this on regular rm as well. I guess the argument against is that people hit sudo too often.

rfay:
Still looks great to me. Thanks!

PR #1022, opened 2018-07-27
Fix Windows mount of .ddev/mysql configuration (Support for Docker 18.06)

rfay:
## The Problem/Issue/Bug:

With the move to Docker 18.06 *something* had to break, and our container test for ddev-dbserver broke on windows. It turns out that mysqld won't (at least now) read config that is world-writable, as we've run into other places on ~/.my.cnf; I don't know why this would show up with a docker version change.

* Change the *test* to mount ro
* Change so ddev generates a docker-compose.yaml that has this mount set to ro so our users don't discover this. 

* If the tests success on the windows container that's good
* For extra credit try changing config in `.ddev/mysql/<something.cnf>` ON WINDOWS and see that your change takes effect.

### DIFF
--- a/containers/ddev-dbserver/Dockerfile
+++ b/containers/ddev-dbserver/Dockerfile
@@ -19,7 +19,7 @@ RUN chmod ugo+x /healthcheck.sh
 # Security-sensitive changes: Make sure our start script can do what is needed
 # But make sure these are right
 RUN chmod ugo+wx /mnt /var/tmp
-RUN chmod -R ugo+wx /var/log /var/tmp/mysqlbase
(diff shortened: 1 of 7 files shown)


rfay:
This is on the right track, but not done yet. What's going on here is if you manually make the my.cnf read-only (in the container) then it passes tests. 

There are two problems I think:
 * if people using windows provide a writable whatever.cnf override file, the server won't respect it.
* Making the whole mount ro doesn't allow the container to create the mysql directory, which may not exist, and mysqld wants to importdir a whole directory.

rfay:
Hoping this one may have all it needs. Starting nightly on it as well.

rfay:
Yay 8 successes!

PR #1031, opened 2018-08-02
Experimental: Mariadb 10.2

rfay:
## The Problem/Issue/Bug:

* We haven't been able to use debian-based containers for mysql/mariadb in the past on *windows*
* The mariadb library container, which we'd like to use is debian-based
* Since we are trying to get away from bind-mounting in https://github.com/drud/ddev/pull/987 and using a docker volume instead, it just might work now. This is based on #987. 

Experimental only.


rfay:
WIll try this again later, have to get the docker volume PR in first.

PR #1034, opened 2018-08-06
Move static tests to a new Buildkite pipeline 

andrewfrench:
## The Problem/Issue/Bug:
`make -s staticrequired`, which lints the source, is currently run *after* the long suite of functional and unit tests.  In situations where the project can successfully be built but doesn't meet the standards of the linter, it takes more than half an hour to find out about it.

Creates a new Buildkite pipeline to run static tests independently of other tests.

Inspect the results of the new Buildkite test, if those still work, we're good.


rfay:
It used to run before, and that was yet more annoying, because you wouldn't find out about the things that take a long time and are really important (the actual tests). 

I'd say let's run the static check as a separate workflow (or even on buildkite as a separate job maybe).  The problem with Circle is it (seems to) count all workflows equal, so it says things are fixed when one of them starts working. Might be a way around this.

rfay:
Yeah, the problem is that the tests are more important than the static analysis, and require far more processing power.

Issue #941, opened 2018-06-26
Stable port number for accessing mysql and http on localhost


tmotyl:
**Is your feature request related to a problem? Please describe.**
I'm using mysqlworkbench to connect to ddev database using external port (found using ddev describe). The problem is that every time ddev is restarted  the port changes, which means I need to reconfigure mysqlworkbench. This is very annoying.

**Describe the solution you'd like**
The mysql port should not change after restarting ddev.

rfay:
I don't know of a way to accomplish this, since the localhost port binding is managed on ephemeral ports by docker. Of course http-based services (like phpmyadmin) use ddev-router for access, so they stay in a defined http-space.

tmotyl:
can the port binding be generated once on ddev config and then saved in config.yaml?

rfay:
Maybe. Seems like this wouldn't work at all for teams that check in their config.yaml. Every machine would have various kinds of conflicts based on what projects they were running.

tmotyl:
In case of conflict I'm ok if the port is adapted.

rickmanelius:
FWIW, there was a related request by @mglaman but solving this in a different way. See https://github.com/drud/ddev/issues/691. Basically, rather than pinning the port numbers (and this will become problematic if users are working across multiple teams and they start to conflict), surface a way to expose the port number by exporting a file OR exposing it via `ddev describe -j`. Of course this might not help the use case of this issue because I don't think mysqlworkbench could pull this information.

rfay:
in #691 I asked if this current output (published_port) from `ddev describe -j` is adequate to solve this problem:

```
    "raw": {
        "approot": "/Users/rfay/workspace/d7git",
        "dbinfo": {
            "dbname": "db",
            "host": "db",
            "password": "db",
            "port": "3306",
            "published_port": "32848",
            "username": "db"
        },
```

tmotyl:
an alternative solution which would work for mysql workbench, would be to allow ddev to expose ssh serwer, then mysql workbench could connect through ssh tunnel. 
But then again we have an issue with ports.

Or maybe we could have a similar proxy/router we have for http traffic, so stuff like:
`mysql -h myprojectname.ddev.local -p`
would work from the host machine?

rfay:
Yeah, there's no ssh tunneling with `ddev ssh`, because it's actually `docker exec` which isn't actually related to ssh.

There's no way to redirect mysql traffic like we do with ddev-router, because there's no Host header or equivalent. 

[...]

PR #1036, opened 2018-08-07
Docs: Fix example path to drush.settings.php

JeremySkinner:
Fix missing / in example for loading drush.settings.php (`__DIR__` does not contain a trailing slash).


JeremySkinner:
Thanks @rfay, I didn't realise this was auto-generated as part of settings.ddev.php, just assumed it was an incorrect example. I've gone ahead and modified the templates and removed the snippet from the docs page and replaced it with some new wording. 

I haven't changed the use of the `DEPLOY_NAME` as the way of checking we're outside the container...it seems good enough for now, and at least drush.settings.php is then usable. 

Let me know if there's anything else you'd like changed.

rfay:
This probably needs an automated test, but it's nontrivial to check the host-side drush usage on every platform, where we don't currently have drush installed. I *think* there are drush usage tests sprinkled around that would make sure it's working inside the web container.

JeremySkinner:
I'm happy to try and add some tests for drush usage on the host side, although I'm not familiar with go so might take me a bit of time.

rfay:
That seems like a lot to ask. When we test this we'll see if we can do some tests, if you'll allow push by rfay and andrewfrench to your fork (if you haven't already).

JeremySkinner:
Thanks - sounds good. Yes you should have access to push to the branch.

docs/users/developer-tools.md
```diff
@@ -53,7 +53,7 @@ The port referenced is unique per running project, and randomly chosen from avai
 **Note:** The host database port is likely to change any time a project is stopped/removed and then later started again.
 
 ### Using Drush Installation on Host Machine
-If you have Drush installed on your host system, you can use it to interact with a ddev project by defining a `drush.settings.php` file at the docroot of your code base, and referencing it from your `settings.php` file. The `drush.settings.php` file should look similar to below, using the host port information for your project retrieved from `ddev describe`:
+If you have Drush installed on your host system, you can use it to interact with a ddev project by defining a `drush.settings.php` file at the docroot of your code base. This will be automatically referenced from the generated `settings.ddev.php` file which is located under `web/sites/default`. The `drush.settings.php` file should look similar to below, using the host port information for your project retrieved from `ddev describe`:
```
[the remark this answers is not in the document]
JeremySkinner: I've changed to `<docroot>/sites/default`. I tested this with D8 and the include statement is automatically generated in settings.ddev.php

pkg/ddevapp/drupal.go
```diff
@@ -163,7 +163,7 @@ if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
 // This determines whether or not drush should include a custom settings file which allows
 // it to work both within a docker container and natively on the host system.
 if (!empty($_SERVER["argv"]) && strpos($_SERVER["argv"][0], "drush") && empty($_ENV['DEPLOY_NAME'])) {
-  include __DIR__ . '../../../drush.settings.php';
+  include __DIR__ . '/../../../drush.settings.php';
```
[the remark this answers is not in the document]
JeremySkinner: I'd assumed there was a reason why it was referencing the doc root, rather than looking in sites/default, but if there isn't then this seems better as all the config files are then in one place.

JeremySkinner:
Re #941 and transient ports, personally I'd much prefer to have a single db container exposed on a single port which can then be used by all the dev sites. I know that's not the 'docker way' and others probably wouldn't like this, but would certainly fit my workflow better!

At the moment I'm just running `ddev describe` and updating the port in drush.settings.php manually whenever I start the project. Would it be feasible to automate that following a successful `ddev start`?

JeremySkinner:
Thanks @rfay, that looks good and will be much more useful. Looking forward to playing with it. Thanks again for getting this in so quickly 👍👍

rfay:
Artifacts from this PR can be downloaded for testing from https://circleci.com/gh/drud/ddev/3225#artifacts/containers/0 for testing - there are builds for macOS, Linux, and Windows there.

JeremySkinner:
I've been using this today and it works nicely with no issues.

rfay:
@andrewfrench there are probably a couple of things I slipped in here that you'll want to look at again and make sure you're good with them:

* Settings directory gitignore creation happens in app.Init() instead of in post-config, so we always write our own in there. It's the same for every CMS type.
* The settings.php or main settings file is *not* included in the gitignore. So many people actually check in a settings.php (without db credentials), that I thought it shouldn't be in there. Of course... maybe *i* added that and then reconsidered.

pkg/ddevapp/apptypes.go
```diff
@@ -133,6 +133,10 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 		if err != nil {
 			util.Warning("Unable to create settings file: %v", err)
 		}
+		if err := CreateGitIgnore(filepath.Dir(app.SiteSettingsPath), filepath.Base(app.SiteLocalSettingsPath), "ddev_drush_settings.php"); err != nil {
```
andrewfrench: This may be my CMS ignorance showing, but isn't this very drupal-centric? Do TYPO3 and Wordpress projects have a `ddev_drush_settings.php` file?

I suppose there's no harm in ignoring something that may not exist, but I think this block could be flagged for a CMS-specific refactor.

rfay: My assumption is that it's ok to ignore something that doesn't exist to make the code simpler. Drush is for Drupal only, but I guess it shouldn't be offensive to people to find a non-their-CMS file gitignored.

This is actually *undoing* the previous CMS-specific approach, which involved 4x duplicated code.

andrewfrench: Gotcha, I see that now. That's fine with me.

rfay:
This one fails with the same #fail as master on TestDdevRestoreSnapshot, but only there, and only on Windows 10 Home. We'll definitely have to find out what's going on there, hopefully tomorrow, but it's not related to this PR. I'm going to go ahead and pull it.

PR #1041, opened 2018-08-09
Add docker installation and testing instructions, fixes #1002

### LINKED ISSUE #1002: Add extensive Linux Docker setup instructions
Tried ddev for the first time, on a brand new Ubuntu 18.04 install. I've previously only played with Docker on a mac, so this was new ground for me, making me your ideal test subject. My experience was bumpy, I spent an under-caffeinated hour getting everything to work.

Once I installed ddev, I got a message saying that Docker and Docker Composer are not installed.
My expectation was that the ddev install script would do this for me, since I'm on a OS where both are a few commands away. Nevermind, I'll just paste the commands from the docs. However, ddev didn't give me a link to its docs that shows what I need to paste, it sends me to Docker's confusing marketing website, which takes me to a [page](https://store.docker.com/editions/community/docker-ce-server-ubuntu) listing supported Ubuntu versions, none of them 18.04. I get confused and give up. Decide to install Docker myself. Google for "Install docker" and click the first link, which tells me to do an "apt install docker-io". This is the Docker from the official Ubuntu repo, and it's too old for ddev. I learn that after I've installed it. I go back and uninstall docker-io. Then I follow the next link, which tells me to add the external Docker repo and do an apt install.

The tutorial that did the trick was https://www.digitalocean.com/community/tutorials/how-to-install-and-use-docker-on-ubuntu-18-04, though I also initially skipped the "optional" step of configuring Docker for non-root, which required me to sudo all ddev commands. This step was mentioned in the ddev docs, but I was already away from them due to needing to figure out what to do with Docker.

My bumpy experience would have been much less bumpy if there was an official ddev docs page that at least sent me to the digitalocean tutorial, with notes saying "don't use ubuntu's docker" and "now setup docker to run without root".

rfay:
## The Problem/Issue/Bug:

#1002: Add docker installation instructions

Adds the instructions

* Read the [main page](https://github.com/drud/ddev/blob/98d7259394221ac25b314f173173b259d51f6c50/docs/index.md) and the [docker installation page](https://github.com/drud/ddev/blob/98d7259394221ac25b314f173173b259d51f6c50/docs/users/docker_installation.md)
* Do the installations
* Try the tests suggested in the testing section

### DIFF
--- a/docs/users/docker_installation.md
+++ b/docs/users/docker_installation.md
@@ -1,0 +1,62 @@
+<h1>Docker Installation</h1>
+
+## macOS Installation: Docker-ce For Mac
+
+Most MacOS versions and computers will run Docker for Mac. Homebrew users can `brew cask install docker` or you can download from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg). 
+
+## Windows Installation: Docker-ce For Windows
+
+Docker For Windows is the preferred docker environment for Windows 10 Pro and Windows 10 Enterprise. 
+
+[Download Docker-ce for Windows](https://download.docker.com/win/stable/Docker%20for%20Windows%20Installer.exe)
+
+[Chocolatey](https://chocolatey.org/install) users: `choco install docker-for-windows`
+
+__Please note that you *must* share your local drives in the "settings" after installation or ddev will not be able to mount your project.__
+
+
+## Windows Installation: Docker Toolbox
+
+Docker Toolbox is only recommended for systems that absolutely won't run Docker-ce for Windows (Windows 10 Home, etc.)
+
+[Download and install docker toolbox](https://download.docker.com/win/stable/DockerToolbox.exe). 
+
+[Chocolatey](https://chocolatey.org/install) users: `choco install docker-toolbox`
+
+
+## Linux Installation: Docker-ce
+
+__Please don't forget that Linux installation absolutely requires post-install steps (below).__
+
+__docker-compose must be installed separately, as it is not bundled with docker-ce on Linux, see below.__
+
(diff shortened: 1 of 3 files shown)


docs/users/docker_installation.md
```diff
@@ -0,0 +1,62 @@
+<h1>Docker Installation</h1>
+
+## macOS Installation: Docker-ce For Mac
+
+Most MacOS versions and computers will run Docker for Mac. Homebrew users can `brew cask install docker` or you can download from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg).
```
rickmanelius: Given that we tend to bias/prefer Docker for Mac, should we promote that first in the 2nd sentence? Maybe "Docker For Mac is the preferred docker environment for macOS. Alternatively, you can use the Homebrew option by running `brew cask install docker`."

rfay: Well, we *only* support Docker-for-mac, so that's the only product mentioned here. Let me know what you think is ambiguous and I can improve it. (Docker-ce == Docker-for-mac == docker-ce-for-mac).  They still have docker toolbox for Mac but we've never tried it and haven't ahd any reason to. It might work. Maybe a couple of those really, really old macs we ran into at Badcamp. Not sure.

rickmanelius: Sure. I guess what tripped me us I almost always went for the DMG approach, but if brew cask is installing the CE edition than my comments are irrelevant. Sorry for the noise there!

rfay: Thanks, I do wish the command was `brew cask install docker-ce`, not sure why it isn't. If this is OK then please approve and we'll get it in.

rfay:
👍 and thanks for the review (and for pointing out the lack!)

PR #1043, opened 2018-08-10
Change the TYPO3 AdditionalConfiguration.php template to extend the MAIL array - fixes #1042

### LINKED ISSUE #1042: Automatically generated TYPO3 AdditionalConfiguration.php overwrites MAIL configuration
**Describe the bug**
The automatically generated TYPO3 AdditionalConfiguration.php overwrites the complete MAIL configuration so that the $GLOBALS['TYPO3_CONF_VARS']['MAIL'] array only contains those two values making it impossible to set or use any other values.

**To Reproduce**
1. Go to 'Configuration' module in the backend
2. Click on the Global Configuration
3. Scroll down to the MAIL section
4. Only two values are present

**Expected behavior**
I would expect a much larger array with the complete MAIL configuration (including defaultMailFromAddress, etc.)

**Screenshots**
bildschirmfoto 2018-08-10 um 21 22 44

 - macOS Mojave
 - TYPO3 CMS 9.3.3
 - Docker 18.06.0-ce
 - ddev v1.0.0

**Possible solution:**
The part of the file currently looks like this:
```
// This mail configuration sends all emails to mailhog
$GLOBALS['TYPO3_CONF_VARS']['MAIL'] = [
    'transport' => 'smtp',
    'transport_smtp_server' => 'localhost:1025',
];
```
… and it should probably look like this:
```
// This mail configuration sends all emails to mailhog
$GLOBALS['TYPO3_CONF_VARS']['MAIL']['transport'] = 'smtp';
$GLOBALS['TYPO3_CONF_VARS']['MAIL']['transport_smtp_server'] = 'localhost:1025';
```

dmnkhhn:
## The Problem/Issue/Bug:
Problem described in #1042 

The code in the template file is changed so that the array is extended instead of overwritten so that we have access to the full $GLOBALS['TYPO3_CONF_VARS']['MAIL'] array in TYPO3. 

Testing this automatically was not possible as far as I know.

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -16,7 +16,7 @@ const typo3AdditionalConfigTemplate = `<?php
 /** ` + DdevFileSignature + `: Automatically generated TYPO3 AdditionalConfiguration.php file.
  ddev manages this file and may delete or overwrite the file unless this comment is removed.
  */
- 
+
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['trustedHostsPattern'] = '.*';
 
 $GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default'] = array_merge($GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default'], [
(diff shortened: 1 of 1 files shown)


dmnkhhn:
There should be no negative impact as all other smtp settings are empty by default and they are not touched by this change.
There is only one MAIL setting that's not empty which is `transport_sendmail_command` and it is ignored because the transport is already set to smtp.

rfay:
@dmnkhhn please do add the email you used for this commit to github per https://github.com/drud/ddev/pull/1043#issuecomment-412189451 - Thanks!

dmnkhhn:
Done, thanks for the reminder!

PR #1045, opened 2018-08-12
Add staticrequired to pre-push hooks

rfay:
## The Problem/Issue/Bug:

The provided (optional) githooks don't provide `make staticrequired` which is one thing we'd like them to have.

Adds it.

This change affects only developers who link the pre-push hooks, it doesn't affect users, testing, etc.


rfay:
Thanks @andrewfrench  - I actually just moved the `make staticrequired` into the pre-push.allchecks. I guess maybe I'll add a script to turn it on and off still.

PR #1046, opened 2018-08-13
Don't use err when it's nil

rfay:
## The Problem/Issue/Bug:

In a couple of our tests we occasionally get panics because we get a nil for err and then use err.Error(). We shouldn't use a nil err.

* If err == nil, make sure we don't try to use err.Error()

This affects only tests, and only occasionally, so is a code-review.

See the fail/panic in https://buildkite.com/drud/ddev-macos/builds/489#6c3c127b-8d24-40f2-861e-9ea7704c07e3

### DIFF
--- a/pkg/appimport/appimport_test.go
+++ b/pkg/appimport/appimport_test.go
@@ -57,7 +57,9 @@ func TestValidateAsset(t *testing.T) {
 	// db no sql
 	gofilePath := filepath.Join(testdata, "junk.go")
 	_, isArchive, err = appimport.ValidateAsset(gofilePath, "db")
-	assert.Contains(err.Error(), "provided path is not a .sql file or archive")
+	if err != nil {
+		assert.Contains(err.Error(), "provided path is not a .sql file or archive")
+	}
 	assert.Error(err)
 	assert.False(isArchive)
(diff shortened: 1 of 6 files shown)


pkg/appimport/appimport_test.go
```diff
@@ -52,16 +52,22 @@ func TestValidateAsset(t *testing.T) {
 	testArchivePath := filepath.Join(testdata, "somedb.sql.gz")
 	_, err = appimport.ValidateAsset(testArchivePath, "db")
 	assert.Error(err)
-	assert.Equal(err.Error(), "is archive")
+	if err != nil {
+		assert.Equal(err.Error(), "is archive")
+	}
```
andrewfrench: We'll want to wait until the import-files refactor PR is in and rebase this as the function signature for ValidateAsset has changed, it now returns a bool isArchive value and doesn't rely on the error value.

rfay:
Rebased this, think it does what it should.

I did figure out that one test failure - there was a cached download for a tarball that should never have succeeded. Not sure how that happened, but cleaning up ~/.ddev/testcache/<whateveritwas> fixed it I think.

PR #1047, opened 2018-08-13
Extend default wait time for GetLocalHttpResponse to 20 sec

rfay:
## The Problem/Issue/Bug:

On Windows testing, we're seeing timeouts on GetLocalHttpResponse on a newly brought-up site. Perhaps the 10-second default isn't enough.

Increase the default to 20s. This is only used in testing, but we may want to add a timeout parm to EnsureLocalHTTPContent() for this reason.


rfay:
Well, this doesn't pass on dockertoolbox, same problem. It may be worth pulling but it doesn't solve the problem we're seeing.

rfay:
I don't see any reason to pursue this right now since it doesn't solve the problem.

PR #1048, opened 2018-08-14
Provide a warning when a nonexistent docroot is supplied, closes #1035

### LINKED ISSUE #1035: Allow non existing document root in config
When you want to start a new project or start working on an existing one, there might be the case, that the actual document root for the project does not exist because it is generated by composer. Currently you either have to run composer install on your host machine or set the project root directory as document root.
In the first case, you need to have php and composer running on your local machine (and also the correct php version) which seems the opposite goal of ddev. In the second case, you can start ddev, run composer install or create-project, then shut down ddev and reconfigure the docroot.

Both solutions are not very nice. One actual example is running TYPO3 with the `helhum/typo3-secure-web` package. The document root for this project needs not to be stored in version control (normally).

Instead of an error, ddev should throw a warning, that the document root does not exist but continue starting up. So the developer can run composer install inside the docker container.

When a non existing document root breaks nginx startup, ddev might create the docroot folder structure by itself. But this would be a worse solution then the given one above I think.

andrewfrench:
## The Problem/Issue/Bug:
#1035: The current brand-new project installation workflow requires the docroot to exist during initial configuration, which means that one can't install a site from a completely empty directory using composer from inside the web container.

This requires the site to be initially configured on the host, meaning the user has to have php and composer installed on the host system. Not unreasonable, but annoying.

Warn the user when a non-existent docroot is supplied, but allow the configuration process to continue. This allows site installation to be completed within the container using composer, which ends up being a bit slower but potentially more convenient for the user.

- Begin with an empty directory
- `ddev config` and define a new project
- Provide a docroot directory that does not exist, like `web/public`
- Create a `composer.json` file like the one below, which defines a TYPO3 project:
```
{
    "repositories": [
        {
            "type": "composer",
            "url": "https://composer.typo3.org/"
        }
    ],
    "require": {
        "typo3/cms": "^8.7"
    },
    "extra": {
        "typo3/cms": {
            "web-dir": "web/public"
        }
    }
}
```
- Execute `ddev start`
- Confirm the docroot has been created
- Invoke composer from inside the web container to install the CMS: 
```
ddev exec composer install -d ../../
```
- Load the project URL and complete installation

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -172,7 +172,7 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
 	if docrootRelPath != "" {
 		app.Docroot = docrootRelPath
 		if _, err = os.Stat(docrootRelPath); os.IsNotExist(err) {
-			util.Failed("The docroot provided (%v) does not exist", docrootRelPath)
+			output.UserOut.Warnf("Warning: the provided docroot at %s does not currently exist.", docrootRelPath)
 		}
 	} else if !cmd.Flags().Changed("docroot") {
 		app.Docroot = ddevapp.DiscoverDefaultDocroot(app)
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -168,7 +168,7 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
 		app.Name = filepath.Base(pwd)
 	}
 
-	// docrootRelPath must exist
+	// Warn the user if the supplied docrootRelPath doesn't exist yet
 	if docrootRelPath != "" {
 		app.Docroot = docrootRelPath
 		if _, err = os.Stat(docrootRelPath); os.IsNotExist(err) {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -25,6 +25,9 @@ var appType string
 // showConfigLocation if set causes the command to show the config locatio
 var showConfigLocation bool
 
+// createDocroot will allow a nonexistent docroot to be created
+var createDocroot bool
+
 // extraFlagsHandlingFunc does specific handling for additional flags, and is different per provider.
 var extraFlagsHandlingFunc func(cmd *cobra.Command, args []string, app *ddevapp.DdevApp) error
 
@@ -103,6 +106,7 @@ func init() {
 	ConfigCommand.Flags().StringVarP(&appType, "apptype", "", "", apptypeUsage+" This is the same as --projecttype and is included only for backwards compatibility.")
 	ConfigCommand.Flags().BoolVarP(&showConfigLocation, "show-config-location", "", false, "Output the location of the config.yaml file if it exists, or error that it doesn't exist.")
 	ConfigCommand.Flags().StringVarP(&siteName, "sitename", "", "", projectNameUsage+" This is the same as projectname and is included only for backwards compatibility")
+	ConfigCommand.Flags().BoolVar(&createDocroot, "create-docroot", false, "Prompts ddev to create the docroot if it doesn't exist")
 	err := ConfigCommand.Flags().MarkDeprecated("sitename", "The sitename flag is deprecated in favor of --projectname")
 	util.CheckErr(err)
 	err = ConfigCommand.Flags().MarkDeprecated("apptype", "The apptype flag is deprecated in favor of --projecttype")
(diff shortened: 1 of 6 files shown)


cmd/ddev/cmd/config.go
```diff
@@ -166,7 +166,7 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
 	if docrootRelPath != "" {
 		app.Docroot = docrootRelPath
 		if _, err = os.Stat(docrootRelPath); os.IsNotExist(err) {
-			util.Failed("The docroot provided (%v) does not exist", docrootRelPath)
+			output.UserOut.Warnf("Warning: the provided docroot at %s does not currently exist.", docrootRelPath)
```
rfay: util.Warning()? Easier to say.

pkg/ddevapp/config.go
```diff
@@ -518,10 +513,9 @@ func (app *DdevApp) docrootPrompt() error {
 	// Ensure the docroot exists. If it doesn't, prompt the user to verify they entered it correctly.
 	fullPath := filepath.Join(app.AppRoot, app.Docroot)
 	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
-		output.UserOut.Errorf("No directory could be found at %s. Please enter a valid docroot\n", fullPath)
-		app.Docroot = ""
-		return app.docrootPrompt()
+		output.UserOut.Warnf("Warning: the provided docroot at %s does not currently exist.", fullPath)
```
rfay: util.Warning() is easier to say :)

mfrieling:
The same issue also applies for `ddev ssh` (and depending commands like `ddev exec`) which fails with the error message

> Failed to ssh: error restoring AppConfig: no project name given

**Solution:**
Provide a warning similar the one to `ddev start` and when SSHing into the container cd to `/var/www/html` instead of the not yet existing `/var/www/html/[docroot]`.

This is especially required where permissions can make problems if git & composer commands executions are mixed between Windows and the container and e. g. the Drupal 8 project (containing the docroot `web` folder) lives in a different git repo than the DDEV project.

andrewfrench:
+1 to the can of worms: looks like there's definitely some undesirable behavior in certain Drupal states that deserves more investigation. To your first point, I'm seeing some weird results as well, I thought we were good because `ddev` creates the docroot if it doesn't exist, but it looks like the order of operations makes that difficult to rely on and could potentially be more trouble than it's worth to unpack. I'll look into it.

To your second point: maybe I need to try to break it harder, but because `ddev start` creates the docroot path, I was able to `ddev ssh` right into `/var/www/html/99/`.

andrewfrench:
@rfay We're back to requiring a docroot, but now give an opportunity to create it via prompt/flag as discussed. One remaining question that came up during my testing: as we're writing config files on `ddev config`, should we go ahead and create the entire path to the config files regardless of whether or not they exist? Currently, we'll just warn the user that the parent file doesn't exist and move on (for example, when configuring a Drupal project and sites/default doesn't exist yet).

rfay:
I guess let's try creating the path to the settings files. We'd have to change a lot to stop doing that with the files path. And composer is going to need an empty directory anyway, so not creating them is not going to solve that. 

Is that the right answer?

cmd/ddev/cmd/config.go
```diff
@@ -168,11 +172,26 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
...
 
-	// docrootRelPath must exist
+	// Ensure that the docroot exists
 	if docrootRelPath != "" {
 		app.Docroot = docrootRelPath
 		if _, err = os.Stat(docrootRelPath); os.IsNotExist(err) {
-			util.Failed("The docroot provided (%v) does not exist", docrootRelPath)
+			// If the user has indicated that the docroot should be created, create it.
+			if !createDocroot {
+				util.Failed("The provided docroot %s does not exist. Allow ddev to create it with the --create-docroot flag.", docrootRelPath)
```
rfay: Nice prompting/error message here.

pkg/ddevapp/config.go
```diff
@@ -518,10 +507,28 @@ func (app *DdevApp) docrootPrompt() error {
...
 	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
-		output.UserOut.Errorf("No directory could be found at %s. Please enter a valid docroot\n", fullPath)
-		app.Docroot = ""
-		return app.docrootPrompt()
+		util.Warning("Warning: the provided docroot at %s does not currently exist.", fullPath)
+
+		// Ask the user for permission to create the docroot
+		for {
+			fmt.Printf("Create docroot at %s? [Y/n]: ", fullPath)
+			resp := util.GetInput("")
```
rfay: util.Prompt() is just a wrapper on util.GetInput() that does the prompt also.

pkg/ddevapp/wordpress.go
```diff
@@ -164,10 +164,13 @@ func WriteWordpressConfig(wordpressConfig *WordpressConfig, filePath string) err
 		return err
 	}
 
-	// Ensure target directory is writable.
+	// Ensure target directory exists and is writeable
 	dir := filepath.Dir(filePath)
-	err = os.Chmod(dir, 0755)
-	if err != nil {
+	if err = os.Chmod(dir, 0755); os.IsNotExist(err) {
```
rfay: This seems like an improved pattern :)

pkg/ddevapp/config.go
rfay: Note that [Y/n] normally means that 'Y' is the default, but if you hit <enter> here, it just asks you again.

```
$ ddev config
You are reconfiguring the project at /tmp/junk.
Project name (junk):

The docroot is the directory from which your site is served. This is a relative path from your project root (/tmp/junk)
You may leave this value blank if your site files are in the project root
Docroot Location (docroot): junker99
Warning: the provided docroot at /tmp/junk/junker99 does not currently exist.
Create docroot at /tmp/junk/junker99? [Y/n]:
Create docroot at /tmp/junk/junker99? [Y/n]:
Create docroot at /tmp/junk/junker99? [Y/n]: y
Created docroot at /tmp/junk/junker99.
Found a php codebase at /tmp/junk/junker99.
Project Type [drupal6, drupal7, drupal8, wordpress, typo3, backdrop, php] (php):
```

andrewfrench: Good point - I agree that's how it should behave. I'll update this.

Issue #994, opened 2018-07-16
v1.1.0 Release Checklist Due 2018-08-15


dclear:
Remaining actions:

* [x]  Review release (product owner)
* [x] Approve release (product owner and release manager)
* [x] Create binaries (any DRUD maintainer)
* [x] Draft release notes (release manager)
* [x] Draft additional announcements for blog, newsletter, etc (not applicable at the moment)
* [x] Create release (release manager)
* [x] Update Roadmap (TPM/release manager)
* [x] Followup issues spun off if we need them.

For additional background information, please see our [Product Release](https://github.com/drud/community/blob/master/development/product_release.md) instructions here.

rfay:
Artifacts for testing v1.1.0 are at https://circleci.com/gh/rfay/ddev/637#artifacts/containers/0

rickmanelius:
Current diff https://github.com/drud/ddev/compare/v1.0.0...HEAD

rickmanelius:
Based on the release notes and diff, I believe these are the end-user features to focus on.

* [files-import](https://github.com/drud/ddev/pull/1037)
* [drud-s3 pluging](https://github.com/drud/ddev/issues/998)
* [covert to docker volumes, snapshotting](https://github.com/drud/ddev/issues/714)
* [memcached](https://github.com/drud/ddev/pull/927)
* [hostname clean ups](https://github.com/drud/ddev/issues/831)
* [drush on host machine](https://github.com/drud/ddev/issues/454)

Luckily, all but one I've tested before at some level. The drush command is the only one I'll be using for the first time.

rickmanelius:
Here is my working copy so far.

# ddev v1.1.0 Review

## Executive Summary

Based on the following analysis, [ddev v1.1.0](https://github.com/drud/ddev/milestone/32) is not ready for release.

## Milestone Overview (@TODO)

@TODO

## Testing

Performed by Rick Manelius as of 2018-08-14

### Hardware Setup

I’m using a laptop with the following high-level specifications:

* MacBook Pro: Retina 15-inch Mid 2015
* 2.2 Ghz Intel Core i7
* 16 GB 1600 MHz DDR3 RAM
* 250 GB Hard Drive
* macOS Sierra 10.12.4

### Software Setup

* Xcode
* Docker 18.03.1-ce
* Git 2.15.1

### Repositories Used

* 

### Preparation

* TODO

## Results

### To Test

* [files-import](https://github.com/drud/ddev/pull/1037)
[...]

Issue #944, opened 2018-06-27
web container: Warning: World-writable config file '/home/.my.cnf' is ignored


rfay:
**Describe the bug**

/home/my.cnf doesn't work if writeable, and we have ended up with it writeable.

Warning: World-writable config file '/home/.my.cnf' is ignored

**To Reproduce**

```
ddev ssh
mysql
Warning: World-writable config file '/home/.my.cnf' is ignored
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

**Expected behavior**

Connect to  the mysql db named 'db'

Actual behavior: 

```
Warning: World-writable config file '/home/.my.cnf' is ignored
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

The workaround:
Use `mysql -udb -pdb -h db db` 
or
`sudo chmod ugo-w ~/.my.cnf`

PR #1054, opened 2018-08-15
Improve archive extraction path error handling

andrewfrench:
## The Problem/Issue/Bug:
The review in #994 describes some issues with import-files, namely one where a bad extraction path would return a success despite having extracted zero files into the project. 

Additionally, a regression was identified where a user would no longer be prompted for an archive extraction path when using the import-files interactive mode.

If no files match the extraction path in an archive, an error is returned to the user.

The interactive extraction path regression was fixed as well.

Provide an archive in interactive mode and ensure the extraction path prompt is presented.

Provide an invalid extraction dir and ensure an error is generated.

## Related Issue Link(s):
#944

### DIFF
--- a/cmd/ddev/cmd/import-files.go
+++ b/cmd/ddev/cmd/import-files.go
@@ -48,15 +48,13 @@ imported.`,
 
 		importPath, isArchive, err := appimport.ValidateAsset(sourcePath, "files")
 		if err != nil {
-			// Ensure we prompt for extraction path if an archive is provided, while still allowing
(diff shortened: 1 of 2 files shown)


rfay:
Ooops, I get `No upload directory has been specified for the project type drupal6` on a d6 site - that's doing it from just `ddev import-files` with no flags

rfay:
Cancel that. I had wrong ddev. quicksprint installed a new one for me!

PR #1067, opened 2018-08-20
Add some debugging so we know which project fails in TestDevLogs

rfay:
## The Problem/Issue/Bug:

We have regular failures in TestDevLogs, such as https://buildkite.com/drud/ddev-macos/builds/524#11fd5348-c02c-45eb-ab31-9c3444953b0d, but a retry usually sorts it out. We would save ourselves a lot of time by understanding this.

It seems to happen on only one project, so the first step here is to add output identifying the project.

I've manually tested and the procedure in this test works fine on d8 and wordpress. Wordpress is likely the culprit, since I think buildkite is just building one testsite on macos.

### DIFF
--- a/cmd/ddev/cmd/logs_test.go
+++ b/cmd/ddev/cmd/logs_test.go
@@ -54,7 +54,7 @@ func TestDevLogs(t *testing.T) {
 		o.Timeout = 30
 		o.Headers["Host"] = v.Name + "." + version.DDevTLD
 		err = util.EnsureHTTPStatus(o)
-		assert.NoError(err)
+		assert.NoError(err, "expected 200 status not found for project %s: %v", v.Name, err)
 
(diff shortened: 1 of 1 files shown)


rfay:
Some [googling around](https://www.google.com/search?q=php-fpm+%7C+WARNING%3A+%5Bpool+www%5D+child+exited+on+signal+7+(SIGBUS)+after+seconds+from+start&oq=php-fpm+%7C+WARNING%3A+%5Bpool+www%5D+child+exited+on+signal+7+(SIGBUS)+after+seconds+from+start&aqs=chrome..69i57.14418j1j7&sourceid=chrome&ie=UTF-8) is pretty unsatisfying,  The takeaway though is "too much memory use" which seems unlikely, "coding error", kind of odd.

Issue #1072, opened 2018-08-21
Nightly build fails because it tries to pull container it already made


rfay:
**Describe the bug**
We had to do a quick fix to pull the new db image during db migration in https://github.com/drud/ddev/pull/1058

Unfortunately, that code always pulls, it doesn't check to make sure that the image might not already exist. So we have to figure out how to do what docker or docker-composer would do and check for the image locally before pulling it.

The fail looks like this:

`--- FAIL: TestDbMigration (15.75s)
        ddevapp_test.go:1455: TestMain startup: app.Start() failed on site TestPkgDrupal7, err=Failed to migrate db from bind-mounted db: failed to run migrate_file_to_volume.sh, err=failed to pull image drud/ddev-dbserver:v1.1.1-3-g1e554381: API error (404): manifest for drud/ddev-dbserver:v1.1.1-3-g1e554381 not found output=`

There's a second thing: Because TestDbMigration fails, it leaves a bind-mount db around, so that causes TestWriteDrushConfig to fail trying the same thing again:

`--- FAIL: TestWriteDrushConfig (69.44s)
        settings_test.go:106:
                        Error Trace:    settings_test.go:106
                        Error:          Received unexpected error:
                                        Failed to migrate db from bind-mounted db: failed to run migrate_file_to_volume.sh, err=failed to pull image drud/ddev-dbserver:v1.1.1-3-g1e554381: API error (404): manifest for drud/ddev-dbserver:v1.1.1-3-g1e554381 not found output=`

So TestDbMigration needs to unconditionally clean up after itself and get rid of the loaded migration stuff in ~/.ddev.

PR #1075, opened 2018-08-21
Check if image exists locally before forcing a pull

andrewfrench:
## The Problem/Issue/Bug:
See #1072, DDEV tries to pull images that only exist locally.

Check if the image already exists locally before attempting a pull.

Ensure that the nightly tests pass as expected.

## Related Issue Link(s):
#1072

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -406,17 +406,25 @@ func GetDockerIP() (string, error) {
 }
 
 // RunSimpleContainer runs a container (non-daemonized) and captures the stdout.
-// It will block, so not to be run on a container whose entryoint or cmd might hang or run too long.
+// It will block, so not to be run on a container whose entrypoint or cmd might hang or run too long.
 // This should be the equivalent of something like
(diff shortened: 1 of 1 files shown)


andrewfrench:
Yeah, I bet there's a way to do the filtering upstream. Looking into it now.

andrewfrench:
@rfay This one's ready, assuming the nightly gives the thumbs up.

pkg/dockerutil/dockerutils.go
```diff
@@ -487,3 +495,31 @@ func RunSimpleContainer(image string, name string, cmd []string, entrypoint []st
...
+
+	if err != nil {
+		return false, err
+	}
+
+	if len(images) == 0 {
+		return false, nil
+	}
+
+	for _, i := range images {
```
rfay: I don't think we need this stanza below at all? It seems like the filter already got us the exact one and only image we want, at least that's what it looks like to me stepping through. If I'm not mistaken we can "if we found one, true, if we didn't, false"

rfay:
Yes, it does fail to ever find "busybox:latest", so it will always pull a "latest" tag. Not sure we have to solve this. Why don't we just change the requirement on RunSimpleContainer to require a full image:tag (just check to make sure it has a :tag on it) and change the 2 usages in the test from "busybox" to "busybox:latest"?

rfay:
New commit (c2b0dbc  checking for tag) looks good to me, thanks.

PR #1076, opened 2018-08-21
Parallelize Circle build

rfay:
## The Problem/Issue/Bug:

* We're going to need to be able to do more parallel builds for Apache webserver types, and it wouldn't do us any harm to do it for multiple CMSs.
* We need parallel of staticrequired anyway.
* It's been too long coming :)

Reorganize the circleci build

Review the builds (and look at the fancy workflows tab on circleci also, https://circleci.com/gh/drud/workflows/ddev

* This will actually let us lose the buildkite staticrequired build, so will save some resources on our buildkite stuff.

## Related Issue Link(s):

* Apache PR: https://github.com/drud/ddev/pull/1007

## Release/Deployment notes:

* After pulling this, we can disable/remove the buildkite static test.

### DIFF
--- a/.buildkite/static.yml
+++ b/.buildkite/static.yml
@@ -1,7 +1,0 @@
-  - command: ".buildkite/test_static.sh"
-    agents:
-      - "os=macos"
-      - "dockertype=dockerformac"
-    env:
-      BUILDKITE_CLEAN_CHECKOUT: true
-    parallelism: 1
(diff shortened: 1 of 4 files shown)


rfay:
@andrewfrench I think this is ready for review. This is building at https://circleci.com/workflow-run/de08e69e-093c-4e73-9624-7da7b3de3d27 (unfortunately on rfay account instead of drud, I'll try to change that after this build completes)

I'm not going to try to do the nightly build or tag_build yet because they're lightly used at this point and I don't want to break something that might trip us up like that. We'll see regular build stuff right away so we can fix it, but those might trip us in the future. 

If we can get this in I'll try to do nginx and apache versions of the build by keying on the build name and making minor changes to our code to test with different config.

PR #1081, opened 2018-08-24
Relocate the misplaced 'Stopping a project' section and give it text.

rszrama:
## The Problem/Issue/Bug:

There's a misplaced 'Stopping a project' section in the CLI documentation that was never given content.

It moves it down to the section on interacting with a project and gives it text.

Review my notation for the projectname parameter. When I looked at the diff, I realized I used the brackets from the CLI help text whereas the web based documentation for removing a project used less than / greater than signs.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -443,7 +443,7 @@ To manually review additional log files use the `ddev ssh` command and review lo
 
 ## Stopping a project
 
-To stop the development environment for a project run `ddev stop` in its working directory. You can also stop a particular project's environment from any directory by running `ddev stop [projectname]` or stop every running project via `ddev stop --all`.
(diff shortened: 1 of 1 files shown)


rfay:
Oh, please do remember to sign the cla @rszrama https://github.com/drud/ddev/pull/1081#issuecomment-415678595

rszrama:
Reviewed the docs and updated to less / greater-than signs for consistency. Signed the CLA. : )

rfay:
BTW, my own personal preference is to always use `ddev rm`, which removes the containers and reduces docker resources. I don't really even know a downside to it, we've debated moving the names around and getting rid of "stop", because it comes up so fast even after rm.

rszrama:
@rfay Oh! Interesting. Good deal - I just assumed I was supposed to use stop. I only have a handful of sites on this laptop anyways, but it's good to know there's no downside to removing. To date my development environment has been MAMP Pro, where "removing" a host would be completely destructive. : D

PR #1083, opened 2018-08-24
Include empty array in JSON output when no projects are active, fixes #588

### LINKED ISSUE #588: ddev list -j should/could return empty array (and router status) when no sites running

ddev-ui has to resort to backup techniques to handle the case where no sites are running. Instead it would be better if ddev gave it an empty array, perhaps with some context.

More discussion at https://github.com/drud/ddev-ui/pull/62#discussion_r158038523

andrewfrench:
## The Problem/Issue/Bug:
#588

Prints the raw output array even if no projects are active.

Execute `ddev list -j` both with and without active projects.

### DIFF
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -11,7 +11,7 @@ import (
 
 var continuous bool
 
-// DevListCmd represents the list command
+// DdevListCmd represents the list command
 var DdevListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List projects",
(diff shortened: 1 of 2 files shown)


andrewfrench:
Bump it to 2.5s? Sure, I can do that.

rfay:
No, I'd say provide an arg for the period, so it can be adjusted as he sees fit and needs.

rfay:
Probably a separate PR is better I guess. Just seemed like the same territory.

andrewfrench:
Right, agreed. I'll do that in another PR.

PR #1106, opened 2018-09-13
Add locales to web container, fixes #822

### LINKED ISSUE #822: Install system locales in web image
I was using the method `strftime('%B', mktime(0, 0, 0, 1, 1))` in PHP to format a date (here to get the month name), but unfortunately the month was not localized to German but English only, no matter if I changed the default locale in PHP with `setlocale(LC_TIME, "de_DE")` before.

What the method does:
`strftime `— Format a local time/date according to locale settings

The method `strftime` should be able to use the systems locale to format any date correctly.

## How to reproduce this:
Execute this snippet in any DDEV environment:
```php
setlocale(LC_TIME, "de_DE");
$january = strftime('%B', mktime(0, 0, 0, 1, 1));
echo $january; // should print the german localized month name 'Januar' but shows 'January'
```

```
cli     v0.17.0
web     drud/nginx-php-fpm-local:v1.2.2
db      drud/mariadb-local:v0.9.0
dba     drud/phpmyadmin:v0.2.0
router  drud/ddev-router:v0.5.0
commit  v0.17.0
domain  ddev.local
```

The problem is based on the fact that no locales are installed in the web container except:
```
C
C.UTF-8
POSIX
```

My workaround was to use http://php.net/manual/en/class.intldateformatter.php to format the month name which works great without having any locale installed in the web container.

http://php.net/manual/en/function.strftime.php

1

rfay:
## The Problem/Issue/Bug:

in #822 a couple of users have requested locales be installed in the web container. It's easy enough to add that package locales-all

Add locales-all

* Edit .ddev/config.yaml to use `webimage: drud/ddev-webserver:20180913_locales`
* Add the LANG environment variable with a docker-compose.env.yaml as shown below (example, uses de_DE.iso88591)
* ddev start
* `ddev ssh` to visit the web container

```
version: '3.6'

services:
  web:
    environment:
      - LANG=de_DE.iso88591
`


rfay:
I wish it didn't, but this does add 28MB to that container, a pretty big hit for a fairly small gain. I wonder if there are smaller solutions.

drud_ddev-webserver_-_docker_hub

sbaghdadi:
I can confirm that this change works. I tested on mac osx 10.12.04 with ddev 1.1.1 . See Screenshot attached. 

bildschirmfoto 2018-09-17 um 10 00 39


I also want to add, that this not set a correct date in the ddev container. 

bildschirmfoto 2018-09-17 um 10 00 52

If you want to have the correct date and time you also have to add the TZ in environments. Example:

`
version: '3.6'

services:
  web:
    environment:
      - LANG=de_DE.iso88591
      - TZ=Europe/Berlin
`
bildschirmfoto 2018-09-17 um 10 03 21

rfay:
Thanks for the testing! Getting TZ is documented in https://stackoverflow.com/questions/50255466/how-can-i-set-the-timezone-in-my-ddev-containers - I wish I'd come up with an easier way, maybe will revisit. Don't think it's much related to this though.

rfay:
The thing that makes TZ hard is it's not easily accessible on macOS. Explored in https://github.com/drud/ddev/issues/709

PR #1111, opened 2018-09-13
Make apache run properly on Windows, fixes #1109

### LINKED ISSUE #1109: apache-fpm (and probably apache-cgi) webserver_type don't work on windows
**Describe the bug**
Ooops. We never tested apache-fpm on Windows, (and don't have automated tests yet). And we probably weren't planning on automated tests on Windows for apache-fpm :(

Basically, we generate a username in apache's envvars directory, and it comes out to be uid_0, but we don't have uid_0 in /etc/passwd. 

**To Reproduce**

Try running any project with apache-fpm for the webserver type on Windows. It doesn't work.

**Additional context**

This can be fixed by just adding uid_0 and gid_0 to the /etc/passwd and /etc/group file, I think.

rfay:
## The Problem/Issue/Bug:

Apache doesn't work at all on Windows. 

There are two problems. 

First:  https://github.com/drud/ddev/blob/5ed08dc93b285b00fb77ab5ec35fabcaf375ad64/containers/ddev-webserver/files/etc/apache2/envvars#L49-L50 is overwritten with uid_0 (which didn't exist in /etc/passwd)

Second:

Apache just won't run as root. Period. Without recompiling.

Use uid 1000 in windows instead of using uid 0. This seems to work fine. I've checked to make sure that at least with docker-for-windows files created by the webserver come through as the correct user on the host.

Use apache-fpm as webserver_type on Windows.

## Related Issue Link(s):

OP #1109 

## Release/Deployment notes:

This changes the UID used inside the web container to UID 1000 instead of root (uid 0). Although this seems to have no detectable problems on docker for windows in terms of creating files with improper uid there could be side-effects. But it's sure nice to not be running as root inside the container.

This also might causes issues to some Windows user doing an exec hook without sudo.

### DIFF
--- a/containers/ddev-webserver/files/etc/group
+++ b/containers/ddev-webserver/files/etc/group
@@ -1,5 +1,6 @@
 root:x:0:
 nginx:x:0:
+gid_0:x:0:
 gid_1:x:1:
 gid_2:x:2:
 gid_3:x:3:
--- a/containers/ddev-webserver/files/etc/passwd
+++ b/containers/ddev-webserver/files/etc/passwd
@@ -1,5 +1,6 @@
 root:x:0:0:root:/home:/bin/bash
 nginx:x:0:0:root:/home:/bin/bash
+uid_0:x:0:0:root:/home:/bin/bash
 uid_1:x:1:1:gid_1:/home:/bin/bash
 uid_2:x:2:2:gid_2:/home:/bin/bash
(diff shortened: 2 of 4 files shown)


rfay:
I think I'd like to wait to pull this until we have at least basic apache test coverage, which is queued up in https://github.com/drud/ddev/pull/1101

PR #1116, opened 2018-09-18
Docs - change docker and docker-compose req version

lukaszuznanski:
## The Problem/Issue/Bug:
With ddev in version 1.20.0, you need Docker 18.06. With ddev in version 1.20.0, file docker-compose.yaml require version: '3.6', but that property is only available from docker-compose in version 1.20, which supports Docker in version 18.02 or higher. Version 1.22.0 of docker-compose supports Docker 18.06 or higher, which is also required for ddev to run, so I propose to change docs to require Docker in version 18.06 or above, and docker-compose in version 1.22.0 or above.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,8 +6,8 @@
 
 ## System Requirements
 
-- [Docker](https://www.docker.com/community-edition) version 18.03 or higher. Linux users make sure you do the [post-install steps](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user)
-- docker-compose 1.18.0 and higher (bundled with Docker in Docker for Mac and Docker for Windows)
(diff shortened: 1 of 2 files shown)


lukaszuznanski:
@rfay are you sure about this docker compose version? From what I see, version 3.6 of docker-compose 
(currently supported in ddev 1.20) file is introduced in docker compose 1.20 
https://github.com/docker/compose/blob/master/CHANGELOG.md#1200-2018-03-20

rfay:
Thanks for paying close attention! In https://docs.docker.com/compose/compose-file/ they say that 3.6 was introduced in docker *engine* release 18.02.0. 

But... your link wins! Our docker-compose requirement was wrong, and needs to be updated in version.go.  You want to push it or should I?

rfay:
Updated to current upstream master with distasteful web ui merge :) Couldn't push rebase into the fork.

PR #1123, opened 2018-09-22
Upgrade containers (and mariadb to 10.2), fixes #947, fixes #1068

### LINKED ISSUE #947: DB Container - Upgrade MariaDB to 10.2
When using DDEV with recent versions of Neos the mariadb version that comes with ddev is not sufficient since Maria 10.2 is requried.

I suggest to raise the version of the alpine-linux the db-container is based on from 3.7 to 3.8. Alpine 3.8 comes with MariaDB 10.2.15.

Alternatively some documentation on how a default mariadb:10.2 container can be used in ddev could solve the problem aswell.

### LINKED ISSUE #1068: Upgrade web container to Debian 9 (stable, stretch)
**Describe the bug**

Our web container is built on debian jessie, which is obsolete. We'll get better support if we upgrade it. It may be (almost) as easy as changing the base image.

There may be some impact on users who add custom packages in hooks.

rfay:
## The Problem/Issue/Bug:

* ddev-webserver needs to go to current stable debian stretch, fixing #947 
* ddev-dbserver goes to alpine:3.8
* mariadb version goes to 10.2, fixing #1068
* Bump versions and update sources for all containers.

Try it out, it should work.
Preferably use the full circleci build so you get all 3 new images. 
Risks include doing a restore-snapshot (onto mariadb 10.2) on a previous snapshot (Created on mariadb 10.1). So try that.  Also *starting* an existing 10.1 project with this.

## Related Issue Link(s):

#947 was about upgrading the db container to mariadb 10.2
#1068 is about upgrading the web container to debian stable.


rfay:
I manually tested with starting this with an existing 10.1 database and also doing a restore-snapshot of a 10.1 database (after running with 10.2) and didn't find any obvious problems.

PR #1124, opened 2018-09-23
Add rewrites for ssl so apache behaves properly

rfay:
## The Problem/Issue/Bug:

@ondrogo in [Typo3 slack](https://typo3.slack.com/archives/C8TRNQ601/p1536766643000100) pointed out that with apache-fpm or apache-cgi he was unable to log into the backend. The reason ended up being that both Apache and TYPO3 were redirecting to plain https, losing the context in the process.

Apache is *not* very good at being on the receiving end of a reverse proxy. But there are workarounds, as in this PR. You have to convince mod_dir not to redirect to http when operating in https, and you have to make sure that the HTTPS environment variable gets propagated to php.

The base problem is described in http://httpd.apache.org/docs/2.4/mod/core.html#servername: 

>Sometimes, the server runs behind a device that processes SSL, such as a reverse proxy, load balancer or SSL offload appliance. When this is the case, specify the https:// scheme and the port number to which the clients connect in the ServerName directive to make sure that the server generates the correct self-referential URLs.

Unfortunately, we are supporting both http and https at the same time, so can't use the ServerName workaround.

The internet is not all that good at solving this problem for us, but these two obscure links provided hints:
* https://mail-archives.apache.org/mod_mbox/httpd-users/201403.mbox/%3C49404A24C7FAD94BB7B45E86A9305F6214D04652@MSGEXSV21103.ent.wfb.bank.corp%3E
* https://gist.github.com/nurtext/b6ac07ac7d8c372bc8eb

It tries to do those things.

You can either use the full ddev download from Circleci or just use `webimage: drud/ddev-webserver:20180922_apache_https` in your config.yaml. 

* Nontrivial explorations on TYPO3 with apache-cgi and apache-fpm, especially using the backend.
* Nontrivial explorations of Drupal with apache-cgi and apache-fpm
* Make sure to use custom ports on both http and https as another permutation.

We may be able to add cases for this into apache-specific tests, but I'm not sure.

### DIFF
--- a/containers/ddev-webserver/files/etc/apache2/apache-site-default.conf
+++ b/containers/ddev-webserver/files/etc/apache2/apache-site-default.conf
@@ -8,6 +8,16 @@
 	# However, you must set it for any further virtual host explicitly.
 	#ServerName www.example.com
 
+    # Workaround from https://mail-archives.apache.org/mod_mbox/httpd-users/201403.mbox/%3C49404A24C7FAD94BB7B45E86A9305F6214D04652@MSGEXSV21103.ent.wfb.bank.corp%3E
+    # See also https://gist.github.com/nurtext/b6ac07ac7d8c372bc8eb
+
+    RewriteEngine On
+    RewriteCond %{HTTP:X-Forwarded-Proto} =https
+    RewriteCond    %{DOCUMENT_ROOT}%{REQUEST_FILENAME} -d
+    RewriteRule    ^(.+[^/])$           https://%{HTTP_HOST}$1/ [redirect,last]
+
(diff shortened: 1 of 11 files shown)


peterbenke:
Adding `webimage: drud/ddev-webserver:20180922_apache_https` to my config.yaml solved the problem for me. Thank you!

ogrosko:
Testing it with `webimage: drud/ddev-webserver:20180922_apache_https` and clean Typo3 `v9.4` all works fine ;)

rfay:
It's worth considering adding a test for this one.

PR #1139, opened 2018-09-28
Validate config before writing .yaml or starting project, resolves #1095

### LINKED ISSUE #1095: Validate ddev config flag input values
**The Issue**
In #1092, support for defining config values can through flags passed into `ddev config` was added, but these values are currently naively written to `config.yaml` without any validation.  There are several instances where providing wrong values might lead to cryptic error messages at the container or application level, but we can do basic validation on the values before accepting and writing them.

**The Solution**
Validate values passed into `ddev config` flags before writing `config.yaml`. In some cases, like PHP versions, one source of truth should be defined to indicate which values are supported and valid config values. In other cases, like FQDNs and hostnames, basic string validation can be used.

andrewfrench:
## The Problem/Issue/Bug:
Only certain config values were being validated before attempting to start a project, which could lead to cryptic errors. When supplying config value flags to `ddev config`, no validation was done before writing the config file. 

These changes do more to ensure that configuration values are valid before either 1) writing a config file, or 2) attempting to start a project. Additionally, many string values in the source code were constantized and a few helper functions were added as a requirement for valid value lookup.

Similarly, instead of inspecting error values, a small set of error types has been added allowing for error type comparisons, decoupling the error from its text value.

Using the CLI's config flags, ensure that invalid choices lead to validation errors:

- `--projectname invalid!name`
- `--php-version 1.9`
- `--projecttype typo4`
- `--webserver-type nginx-cgi`
- `--additional-hostnames valid,in_valid`
- `--additional-fqdns valid.com,in_valid.com`

Provide combinations of these invalid values alongside valid values to ensure ddev rejects invalid values and, when appropriate, outputs the valid values for that flag.

Using config.yaml, ensure that invalid values lead to validation errors after saving the file and when attempting to `ddev start` the project:

- `name: "invalid name"`
- `type: drupal10`
- `php-version: 9.9`
- `webserver-type: nginx-cgi`
- `additional-hostnames: [valid, in_valid]`
- `additional-fqdns: [valid.com, in_valid.com]`
- `provider: unknown`

Constantization updates will largely rely on the build process completing successfully, ensuring all new 
symbols are resolved correctly.

Test cases have been added to ensure invalid config values generated appropriate errors.

## Related Issue Link(s):
#1095

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -263,10 +263,6 @@ func (app *DdevApp) PromptForConfig() error {
 
 // ValidateConfig ensures the configuration meets ddev's requirements.
 func (app *DdevApp) ValidateConfig() error {
-	if _, err := os.Stat(app.ConfigPath); os.IsNotExist(err) {
-		return fmt.Errorf("no valid project config.yaml was found at %s", app.ConfigPath).(invalidConfigFile)
-	}
-
 	// validate hostnames
 	for _, hn := range app.GetHostnames() {
 		if !hostRegex.MatchString(hn) {
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -23,12 +23,6 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-// DdevDefaultRouterHTTPPort is the starting router port, 80
-const DdevDefaultRouterHTTPPort = "80"
-
-// DdevDefaultRouterHTTPSPort is the starting https router port, 443
-const DdevDefaultRouterHTTPSPort = "443"
-
 // Regexp pattern to determine if a hostname is valid per RFC 1123.
 var hostRegex = regexp.MustCompile(`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`)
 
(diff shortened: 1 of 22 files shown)


pkg/ddevapp/config.go
```diff
@@ -273,20 +257,36 @@ func (app *DdevApp) PromptForConfig() error {
 
 // ValidateConfig ensures the configuration meets ddev's requirements.
 func (app *DdevApp) ValidateConfig() error {
-	if _, err := os.Stat(app.ConfigPath); os.IsNotExist(err) {
-		return fmt.Errorf("no valid project config.yaml was found at %s", app.ConfigPath)
+	provider, err := app.GetProvider()
+	if err != nil {
+		return err.(invalidProvider)
```
rfay: I love typing errors. That's going to be a big step forward for us. Thanks!

pkg/ddevapp/ddevapp.go
```diff
@@ -1205,9 +1205,23 @@ func GetActiveApp(siteName string) (*DdevApp, error) {
 	// Mostly ignore app.Init() error, since app.Init() fails if no directory found. Some errors should be handled though.
 	// We already were successful with *finding* the app, and if we get an
 	// incomplete one we have to add to it.
-	err = app.Init(activeAppRoot)
-	if err != nil && (strings.Contains(err.Error(), "is not a valid hostname") || strings.Contains(err.Error(), "is not a valid apptype") || strings.Contains(err.Error(), "config.yaml exists but cannot be read.") || strings.Contains(err.Error(), "a project (web container) in ")) {
-		return app, err
+	if err = app.Init(activeAppRoot); err != nil {
```
rfay: This is a massive improvement, great with the error typing. 

It will read a lot better with `case webContainerExists, invalidHostname, invalidAppType, ... : return app, err`

This is absolutely suspect code though, quite fragile, and as you'd expect it has a lot of history. I'd *rather* we had the types of error we're going to *ignore* here instead of the types we're going to return. That would be less fragile. We could do that either this time or next time.

pkg/ddevapp/values.go
```diff
@@ -0,0 +1,158 @@
...
+	ProviderDrudS3   = "drud-s3"
+	ProviderPantheon = "pantheon"
+
+	// ProviderDefault contains the name of the default provider which will be used if one is not otherwise specified.
+	ProviderDefault = "default"
+)
+
+// ValidProviders should be updated whenever provider plugins are added or removed, and should
+// be used to ensure user-supplied values are valid.
+var ValidProviders = map[string]bool{
```
rfay: You coming over to the dark side of maps?

PR #1146, opened 2018-09-29
Do not try to start project for a snapshot, fixes #1129

### LINKED ISSUE #1129: Confusing/improper behavior on "ddev rm" when db container is broken
**Describe the bug**

If the db container is broken (won't come up, usually due to corrupted database) and you get "db service stopped" on `ddev list` then:

`ddev rm --R` may fail with confusing errors where it immediately *starts* the project (to make a snapshot, but it hasn't said that yet). During the start, it fails, and never gets to the snapshot. 

**To Reproduce**

1. Make a snapshot of the good db
1. Break a project's db container. `ddev ssh -s db` and `rm /var/lib/mysql/mysql/*.MYD` and `ddev restart`. `ddev list` should show "db container stopped".  Pretty nasty thing to do to a database.
2. `ddev rm -R`. 

Result is 
```
$ ddev rm -R
Starting ddev-t3git-db ... done
ddev-t3git-dba is up-to-date
ddev-t3git-web is up-to-date
Stopping ddev-router ... done
Removing ddev-router ... done
Network ddev_default is external, skipping
Creating ddev-router ... done
```
and `ddev list` shows just "db container stopped" again. No hint is given of what the problem or procedure is.

`ddev rm -R --omit-snapshot` solves the problem just fine.

**Expected behavior**

I expect `ddev rm -R` to always succeeed. It should probably just warn if a snapshot can't be created and continue to removal. It's also questionable whether we should always be creating the snapshot. Maybe we should ask. 

ddev v1.2.0

**Additional context**

rfay:
## The Problem/Issue/Bug:

#1129 describes the terrible cyclic behavior people have when trying to destroy a project but it's already hurt (like the db is broken, so the db container is down). Essentially you do a `ddev rm -R` and it tries to *start* a broken project (in order to snapshot it), which can't possibly work, therefore people end up broken all the way. (`ddev rm -RO` works in this case, but people shouldn't have to know that). 

Don't try to start project to do a snapshot. Require the project to be running in the first place.

See the break-db instructions in #1129. Break it. Do a `ddev rm --remove-data` to see the correct behavior.

### DIFF
--- a/cmd/ddev/cmd/remove.go
+++ b/cmd/ddev/cmd/remove.go
@@ -58,7 +58,7 @@ To snapshot the database on remove, use "ddev remove --snapshot"; A snapshot is 
 			// We do the snapshot if either --snapshot or --remove-data UNLESS omit-snapshot is set
 			doSnapshot := ((createSnapshot || removeData) && !omitSnapshot)
 			if err := app.Down(removeData, doSnapshot); err != nil {
-				util.Failed("Failed to remove ddev project %s: %v", app.GetName(), err)
+				util.Failed("Failed to remove project %s: \n%v", app.GetName(), err)
 			}
 
 			util.Success("Project %s has been removed.", app.GetName())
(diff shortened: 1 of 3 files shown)


rfay:
Thanks for the good suggestions.

1__default__bash_
1__default__bash_

PR #1148, opened 2018-10-01
 Add test to make sure we get the webserver we configure, fixes #1094

### LINKED ISSUE #1094: Create tests for experimental Apache support
**The Issue**
In #1007, experimental support for the `apache-cgi` and `apache-fpm` server types was added. The current test infrastructure, when simple setting the server type, would lead to consistent flakiness in the Apache tests, while the Apache server types functioned well in local testing.

**The Solution**
Revisit the server type testing strategy and ensure Apache tests are solid.

rfay:
## The Problem/Issue/Bug:

#1094 promises tests for Apache webserver_type. We already run all tests against each webserver_type but this adds an explicit test to bring up under each webserver_type and make sure the results are what we expect.

If the code is OK and the test passes, we're good enough, no manual testing is probably required.

### DIFF
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -1562,6 +1562,7 @@ func TestWebserverType(t *testing.T) {
 			err = app.Start()
 			assert.NoError(err)
 
+			// nolint: vetshadow
 			out, _, err := testcommon.GetLocalHTTPResponse(t, app.GetHTTPURL()+"/phpinfo.php")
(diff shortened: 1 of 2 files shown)


rfay:
I meant to swing back to this one to resolve a question I had regarding the commented-out check on the actual header. It deserves a little more attention. The test here is OK. But I'm confused why https://github.com/drud/ddev/blob/66a6eaea16b74035b7ed30112d27816891258ca1/pkg/ddevapp/ddevapp_test.go#L1573 didn't work.

PR #1149, opened 2018-10-02
Add debugging for intermittent failures on TestDdevRestoreSnapshot

rfay:
## The Problem/Issue/Bug:

We see intermittent failures on TestDdevRestoreSnapshot that we've never been able to chase. They happen on Docker Toolbox and I think on macOS as well. This extra debugging info should help us get more information about what is happening.


rfay:
Since the problem may possibly have to do with db problems, I think it's important to report the db logs as well, so this needs one more commit.

PR #1160, opened 2018-10-06
Explicitly support mariadb upgrade from 10.1 to 10.2

rfay:
## The Problem/Issue/Bug:

We incidentally upgraded mariadb from 10.1 to 10.2 when upgrading the container in https://github.com/drud/ddev/pull/1123

However, you need to run `mysql_upgrade` when you do that. So we need to run mysql_upgrade on existing databases when they come up.

* Test to find out the expected mariadb version of the existing database. A new file is added to /var/lib/mysql to suggest this. If not found, it's assumed the version is/was 10.1
* If we have a different mariadb version, run `mysql_upgrade`

* Create a database/project using ddev v1.2.0 (or with `drud/ddev-dbserver:v1.2.0` in config.yaml; it seems to be compatible.
* `ddev rm`
* Bring up the db with this version of ddev or with `dbimage: drud/ddev-dbserver:20181006_mariadb_upgrade` in config.yaml
* `ddev logs` should show no errors, and you should see in there "Running mysql_upgrade" and a successful run.
* Snapshot your new db.
* `ddev rm`
* `ddev start`
* Inspect the db logs with `ddev logs -s db`. You should not see the mysql_upgrade, and should not see errors.
* Restore your snapshot and then inspect with `ddev logs -s db`. You should not see the "Running mysql_upgrade"

**Test bind-migration workflow:**
* Use a project with a bind-mounted (old) db or just mkdir ~/.ddev/<project>/mysql
* `ddev start`
* You should get an error explaining what to do.

**Consider testing restore-snapshot with old snapshot**

* Create a snapshot with v1.2
* Try to restore it with v1.3
* You should get a comprehensible error and a link to docs to workaround.

* I added a test in containers/ddev-dbserver that loads a 10.1 database, runs our container with it, checks for the upgrade happening, then rm's the container and restarts it and checks that the upgrade doesn't happen (since it's already 10.2).

## Related Issue Link(s):

The mariadb upgrade happened in #1123

## Release/Deployment notes:

**SUPER IMPORTANT**: Unfortunately, 
* Since old snapshots are not compatible with this version, we can't migrate bind-mounted databases any more. The code detects this and tells people what to do.
* mariadb/mariabackup 10.1 snapshots are not compatible with mariabackup 10.2. This means that people who have pre-ddev-1.3 snapshots will not be able to use them easily in ddev v1.3+ - but there is a workaround:

1. Set `webimage: drud/ddev-dbserver:v1.2.0` in .ddev/config.yaml
2. `ddev start`
3. `ddev restore-snapshot <name>`
4. `ddev rm`
5. Remove or comment out the webimage line
6. `ddev start`

This PR links people to what to do.


rfay:
Thanks for the review @andrewfrench . I'm going to do one more little bit with this, detect when we can't restore a snapshot and give them a link to the workaround.

PR #1161, opened 2018-10-08
Adds " around the version parameter to avoid trouble with zsh, update to TYPO3 v9 (docs only)

tomasnorre:
## The Problem/Issue/Bug:
For me on macOS it doesn't work without "" around the version string.. 
Further more I would suggest to update the documentation to TYPO3 9LTS.


rfay:
Definitely this should be updated to LTS, you can do that i this PR if you want. 

But this *does* work fine on macOS, in fact this suggested approach is directly from https://composer.typo3.org/ - Are you using a shell other than sh or bash that makes special use of the ^ ?

tomasnorre:
The 9LTS is already in my PR. 
I'm using ZSH as shell, and that's exactly the problem. 

I've tested my fix on Bash, sh and ZSH and it's working in all of them with the "^9"

Issue #1170, opened 2018-10-11
v1.3.0 regression: Restoring snapshots from previous versions


rfay:
**Describe the bug**

* At one point it was easily possible to restore snapshots by just changing the dbcontainer in config.yaml per https://ddev.readthedocs.io/en/latest/users/troubleshooting/#old-snapshot

However, I seem to have increased the error-checking in the snapshot code in ddev itself and it now prevents that. So you still get `$ ddev restore-snapshot umami_maria_10_1
Failed to restore snapshot umami_maria_10_1 for project d8git: snapshot db_snapshots/umami_maria_10_1 is not compatible with this version of ddev and mariadb. Please use the instructions at https://ddev.readthedocs.io/en/latest/users/troubleshooting/#old-snapshot for a workaround to restore it` but unfortunately that workaround doesn't work.

* An additional step was required anyway. You had to `ddev rm -RO` before starting up, because the v1.2.0 container was unable to function with the v1.3.0 database, so came up broken.

rfay:
I think this can be fixed by changing https://github.com/drud/ddev/blob/c9e3cb59ca1d56f12b5c91eac464f2bcc0bdd413/pkg/ddevapp/ddevapp.go#L909-L911
app.Exec("db", "bash", "-c", "mysql --database=mysql -e 'DROP DATABASE IF EXISTS db; CREATE DATABASE db;' && cat /db/*.*sql | mysql db")
to add an `app.Exec('db', 'bash', '-c', 'mariabackup --version | grep 10.2')` or something like that.

PR #1176, opened 2018-10-11
Inspect mysql version before failing when restoring a snapshot, #1170

andrewfrench:
## The Problem/Issue/Bug:
#1170

This allows the snapshot error help text method of recovery to work by inspecting the mysql version currently being used and only failing if the current database version requires the `db_mariadb_version.txt` file to exist.

- Configure and start a project using ddev v1.2.0
- Populate the database
- `ddev snapshot`
- `ddev rm`
- Switch to ddev v1.3.0
- `ddev config`
- `ddev start`
- `ddev restore-snapshot <snapshot>`; this will fail - below are the steps recommended in the help link
- `ddev rm`
- Edit the database image in .ddev/config.yaml to be v1.2.0
- `ddev restore-snapshot <snapshot>`; this should succeed
- `ddev rm`
- Edit the database image in .ddev/config.yaml to be v1.3.0, or comment out the specific image to use the default
- `ddev start`
- Project should start successfully
- `ddev logs -s db` will show "Running mysql_upgrade" 

## Related Issue Link(s):
#1170

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -907,20 +907,21 @@ func (app *DdevApp) RestoreSnapshot(snapshotName string) error {
 		return fmt.Errorf("Failed to find a snapshot in %s", hostSnapshotDir)
 	}
 
-	// This command returning an error indicates grep has failed to find the value
-	_, _, err := app.Exec("db", "sh", "-c", "mysql -V | grep 10.2")
-	if !fileutil.FileExists(filepath.Join(hostSnapshotDir, "db_mariadb_version.txt")) && err == nil {
-		return fmt.Errorf("snapshot %s is not compatible with this version of ddev and mariadb. Please use the instructions at %s for a workaround to restore it", snapshotDir, "https://ddev.readthedocs.io/en/latest/users/troubleshooting/#old-snapshot")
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -906,18 +906,21 @@ func (app *DdevApp) RestoreSnapshot(snapshotName string) error {
...
+
+	// This command returning an error indicates grep has failed to find the value
+	_, _, err := app.Exec("db", "sh", "-c", "mysql -V | grep 10.2")
+	if !fileutil.FileExists(filepath.Join(hostSnapshotDir, "db_mariadb_version.txt")) && err == nil {
 		return fmt.Errorf("snapshot %s is not compatible with this version of ddev and mariadb. Please use the instructions at %s for a workaround to restore it", snapshotDir, "https://ddev.readthedocs.io/en/latest/users/troubleshooting/#old-snapshot")
 	}
 
 	if app.SiteStatus() == SiteRunning || app.SiteStatus() == SiteStopped {
-		err := app.Down(false, false)
+		err = app.Down(false, false)
```
rfay: I don't think we want this, people don't need an extra snapshot when they're trying to restore one?

PR #1179, opened 2018-10-12
Add timeouts to queries on dynamic-generated URLS (tests only), fixes #1172

### LINKED ISSUE #1172: Add timeout in content tests in TestDdevRestoreSnapshot
**Describe the bug**

We have intermittent test timeouts in TestDdevRestoreSnapshot. I thought I had put increased timeouts in there, but apparently not everywhere we need them. For example:

https://github.com/drud/ddev/blob/3456f8c541f2c33de1f095e7aa96365389efaac9/pkg/ddevapp/ddevapp_test.go#L711

No reason not to put a 45 second timeout in here.

rfay:
## The Problem/Issue/Bug:

We have some occasional timeouts in TestDdevRestoreSnapshot when a snapshot restores dynamic content and then we test the content. 

This just increases the timeout on those; the default is 20s, which you'd think would be enough, but sometimes a dynamic project takes a really long time to come up.

### DIFF
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -627,7 +627,6 @@ func TestDdevFullSiteSetup(t *testing.T) {
 		// Test static content.
 		testcommon.EnsureLocalHTTPContent(t, app.GetHTTPURL()+site.Safe200URIWithExpectation.URI, site.Safe200URIWithExpectation.Expect)
 		// Test dynamic php + database content.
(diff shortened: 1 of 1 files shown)


rfay:
Hmm, buildkite/ddev-macos *still* failed on this in https://buildkite.com/drud/ddev-macos/builds/832#f0a6d5d0-1764-449e-8c5e-f43c41be4f73, so maybe there's something more to this.

PR #1193, opened 2018-10-13
Add command to output final docker-compose config, closes #1192

### LINKED ISSUE #1192: Add a command to output final docker-compose config
Debugging docker-compose configurations can be difficult when multiple docker-compose.*.yaml files exist, and many values are generated and inserted from environment variables. 

Add a command to allow for easy debugging of the values being used to create and configure service containers. The command should print the end-state docker-compose configuration after all override files and environment variable substitution.

andrewfrench:
## The Problem/Issue/Bug:
#1192: Debugging docker-compose configurations can be difficult when multiple docker-compose.*.yaml files exist, and many values are generated and inserted from environment variables. 

This PR adds a hidden debugging command, `ddev debug compose-config`, that will go through the same steps to create a full docker-compose configuration as when starting a project. After combining any docker-compose.*.yaml files and inserting environment variables, the command will output the final configuration state. This is especially helpful for determining the exact values that are passed to a container when debugging override files or trying to figure out exactly what values are available to the web/db/dba container.

When in a project's approot, run `ddev debug compose-config` and ensure that a correct docker-compose configuration is printed.

- Update values like the project name, server type, etc. and ensure that these updates are propagated to the docker-compose config output
- Create a docker-compose.override.yaml and add additional values, override existing values, etc. and ensure that these updates are propagated to the docker-compose config output

Run `ddev debug compose-config <projectname>`, where <projectname> is the name of a running project, and ensure that a correct docker-compose configuration for the provided project has been printed

A functional test has been added that the function outputs expected values and that values from a docker-compose.override.yaml are included in the result.

## Related Issue Link(s):
#1192


rfay:
This is a great step. I kind of wonder if we should hide all debugging things under a single debugging command though?

andrewfrench:
The motivation behind this was tinkering with override .yamls and there not being a straightforward way to have ddev emit the configuration it was building. Uses I can imagine:

- Removes a layer of opacity in the configuration process as merging multiple .yaml files and inserting environment variables is a multi-step process that can't be trivially replicated through the terminal 
- Provides a very simple way to sanity check docker-compose configurations
- Allows a user to experiment with changes to their docker-compose configuration without having to start a project and inspect values from within the container
- Allows us to ask for the output of `ddev compose-config` if we're debugging a problem that may have to do with composer configuration

I'm very open to restructuring this, but I'll ask for a example of what you're imagining as far as debugging commands under a top-level command. Which other commands live in that group (`ddev log`?)? What is that top-level command called? Is that command hidden?

PR #1222, opened 2018-10-22
ddev-dbserver based on mariadb 10.1 and 10.2 images, fixes #1221

### LINKED ISSUE #1221: ddev should support  both mariaDB 10.1 and 10.2

TYPO3 v8.7 unfortunately is incompatible with MariaDB 10.2, which is used in ddev v1.3.0. https://github.com/davdenic/typo3-ddev-project-setup explains the issues. 

We also have regular client requirements for other versions of MariaDB and Mysql. We probably have to figure out how to support them either as we do for nginx/apache, or as separate docker images.

It won't surprise me if this forces us to move to debian (or library/mariadb) as a base image, which should work now that we mount the db on docker volumes.

**Additional context**

rfay:
## The Problem/Issue/Bug:

#1221: We probably need to support more versions of mysql/mariadb. This is relatively hard (and unfamiliar) on Alpine. 

We had the db container running on Alpine because it was the only distro that would work with a bind-mount of the database onto Windows. But we no longer use bind mounts. So maybe we can use library/mariadb (here) or directly derive from bitnami/minideb (like we do for ddev-webserver)


rfay:
Unfortunately, this was just a bare beginning; it's obvious this approach can work (I think), but it's not still the right way to solve all the needs in OP #1221. But it's a thought and a starting place. Don't want to give it too much attention until the ssh one matures.

WebsiteDeveloper:
@rfay any news on this?

rfay:
It's prioritized for this release @WebsiteDeveloper, not completely sure we'll get it. But WAY glad to know that you're interested. Are you currently using `dbimage: drud/ddev-dbserver:v1.2.0` successfully?

I guess one of the things we could do is to only support mariadb 10.1 and 10.2 and that would simplify a bit.

rfay:
It's possible that the `ERROR: for db  Cannot create container for service db: Duplicate mount point: /var/lib/mysql` is a problem that has been fixed in Docker/Docker-compose:
* https://github.com/docker/compose/issues/4751
* https://github.com/docker/compose/issues/5885

rfay:
Closing this in favor of #1318, a less-ambitious approach that sticks with the same Alpine base.

PR #1223, opened 2018-10-23
Add ddev-ssh-auth container to provide ssh-auth, fixes #414, fixes #1152

### LINKED ISSUE #414: As a ddev user, I want to be able to use ssh/rsync from within the web container

* Feature request

In a conversation with @schnitzel, he mentioned a workflow within his organization where they leverage an SSH agent docker container to make it easier for team members to communicate with an API endpoint without having to re-authenticate by typing in their SSH passphrases.

Given that ddev already allows the ability to extend/add other services as described here (https://ddev.readthedocs.io/en/latest/users/extend/additional-services/), this feature could be provided by essentially creating a docker compose file into the ddev library. Example, we did the same with Apache Solr here https://ddev.readthedocs.io/en/latest/users/extend/additional-services/#apache-solr.

Ideally, we can get access to the docker compose file used by @schnitzel along with any other business logic needed to connect in their workflow.

* The ability to add custom containers #121

### LINKED ISSUE #1152: Allow disabling containers (phpmyadmin, etc)

People regularly ask for the ability to disable phpmyadmin container. On occasion they ask to disable the ddev-dbserver. 

This would be a very simple add to the config.yaml. 

```
omit_containers: 
    - phpmyadmin
    - ddev-dbserver
```

rfay:
## The Problem/Issue/Bug:

People want to be able to use their ssh keys inside the web container, mostly for composer actions against private repositories, but also for things like drush access and direct ssh access.

* Adds a ddev-ssh-auth container that serves all other ddev projects with ssh-auth running
* Incidentally solves #1152 (request to allow turning off phpMyAdmin container) with an omit-containers array in config.yaml, which was added in order to allow turning off the ddev-ssh-agent container (but might as well turn off dba=phpMyAdmin as well)

This will probably need a special ddev command for step 2, but for now, do this:

1. `ddev start` a project
2. Assuming your password-protected ssh key is in ~/.ssh/, run `ddev auth ssh` and give it your key password. 
3. In a project, `ddev ssh`. You should be able to `ssh <username>@somessh-protected-host` successfully. If you don't have something useful to test with, give me your pubkey and I'll set you up.
4. Put a password-protected ssh key in another directory and use `ddev auth ssh -d /somewhere/else` to add it (and test that you can now access what you ought to be able to access)
5. Try using `ddev rm --stop-ssh-agent` to stop the ssh agent container; when you do that, auth from within the web container should be lost; to re-establish it a `ddev start` should get it going again, and ddev auth ssh is required again.
6. Try omit_containers: ["ddev-ssh-agent", "dba"] with a `ddev start` to verify that the dba (phpMyAdmin) and ddev-ssh-agent containers don't start.

This doesn't yet have tests and it will need some.

## Related Issue Link(s):

OP #414 
Random issue about omitting containers #1152 

## TODO: 

- [x] Needs tests
- [x] The shared volume may need to be deleted at first run of the ddev-ssh-agent (right before running it?)
- [x] ddev probably needs a special command to add the ssh key. `ddev auth ssh` ?
- [x] Figure out the correct way for users to kill off the ddev-ssh-agent if they don't want it
- [x] Consider configuration for disabling the ddev-ssh-agent
- [x] Add ddev-ssh-agent status to ddev describe (maybe ddev list?)
- [x] Add ddev-ssh-agent version to `ddev version`
- [x] Consider whether we need a command-line switch to `ddev config` for omit_containers.


rfay:
@andrewfrench I'd love to have you give this a go today. The code is pretty much in place; it still needs a test for the new omit_containers piece, and depending on your view it may need a config flag for that. 

The thing that I'm least happy with is using the docker command directly in the `ddev auth ssh` implementation. Second least happy is using docker directly in the pkg test. 

If you would like a different command setup than `ddev auth ssh` I'm open to suggestions. Open to suggestions on anything, but I think this is now accomplishing what we'd hoped for in the basics.

rfay:
This now has config flags, config value validation, docs, and tests. I think I'm pretty well done with it until review.  I did give up on testing it in docker-toolbox environment. Hopefully it will pass at least *most* of the tests with this last push.

rfay:
> Running ddev rm --stop-ssh-agent appears to be equivalent to running ddev rm -a --stop-ssh-agent

Just checked, they are not equivalent.
```
cd d8git
ddev start
cd d7git
ddev start
ddev rm --stop-ssh-agent
```
Only d7git is stopped, not all projects (as with ddev rm -a), it is not equivalent to ddev rm -a --stop-ssh-agent

We could, of course, create a whole separate command for destroying the ssh-agent, but I think this should be a very unusual operation anyway. It's really just for people who just discovered it was running and don't understand why.

andrewfrench:
You're right about that - I only had one project running and apparently forgot that a `ddev rm` without args will stop the current project. In that case, one can't currently stop the ddev-ssh-agent unless you're also stopping a project with it (`ddev rm --stop-ssh-agent` fails in non-project-dir `~`)

rfay:
Right, assuming it's an unusual event, I tacked it onto rm rather than adding a new command. Although `docker rm -f ddev-ssh-agent` is the obvious technique. I really think this is only useful to people who are seeing it for the very first time and don't want or understand it.

rfay:
Thanks for the prods, there were a lot of rough spots, like `ddev auth ssh` outside of a project directory, no keys in the ~/.ssh, directory doesn't exist, all kinds of rough spots. I think they're better now.

You'll want to `docker pull drud/ddev-ssh-agent:20181022_add_ssh_auth` before retesting.

It does now start the agent on `ddev auth ssh` if it's not already running.

PR #1228, opened 2018-10-24
Add open CORS to ddev-router nginx config.

Mogtofu33:
## The Problem/Issue/Bug:

Setting up a ddev stack for Drupal 8 with [ContentaCMS](https://github.com/contentacms) (a headless Drupal distribution) and a separated frontend as a new service (with a different hostname as _additional_hostnames_), the ddev-router container do not allow access from the frontend because of a different _Origin_ (403).
In the same time using CORS, ddev-router do not provide or relay any _Access-Control-Allow-Origin_ information.

Here is an installation instruction used to setup ddev with Contenta:
* https://github.com/Mogtofu33/contenta-ddev

I assume the same kind of problem can occur with ddev on any type of stack when dealing with CORS on a project.

Clean the header _Origin_ for all request to the router and add an open _Access-Control-Allow-Origin_.

This configuration is full open and not safe for production, so perhaps could be a better approach to allow _proxy.conf_ override on ddev directly.

## Related Issue Link(s):

* https://github.com/Mogtofu33/contenta-ddev/issues/1

### DIFF
--- a/containers/ddev-router/nginx.tmpl
+++ b/containers/ddev-router/nginx.tmpl
@@ -78,6 +78,10 @@ proxy_set_header X-Forwarded-Port $proxy_x_forwarded_port;
 
 # Mitigate httpoxy attack (see README for details)
 proxy_set_header Proxy "";
+
+# Open CORS
+proxy_set_header Origin "";
+add_header "Access-Control-Allow-Origin" "*" always;
 {{ end }}
 
 {{ $enable_ipv6 := eq (or ($.Env.ENABLE_IPV6) "") "true" }}
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -45,7 +45,7 @@ var DBATag = "v1.3.0" // Note that this can be overridden by make
 var RouterImage = "drud/ddev-router"
 
 // RouterTag defines the tag used for the router.
-var RouterTag = "v1.3.0" // Note that this can be overridden by make
+var RouterTag = "Mogtofu33-master" // Note that this can be overridden by make
 
 // COMMIT is the actual committish, supplied by make
 var COMMIT = "COMMIT should be overridden"


rfay:
@Mogtofu33 the artifacts from this PR are at https://circleci.com/gh/drud/ddev/4874#artifacts/containers/0

If you or preferably someone else using contenta could manually test and review this, I'll be happy to approve it. 

Also please add explicit manual testing instructions. Thanks!

rfay:
Ping @Mogtofu33 https://github.com/drud/ddev/pull/1228#issuecomment-433236490

mglaman:
I'm confused. I've used a JSON API Drupal project with CORS configured and had no problems. Why does ddev-router need specific configuration?

rfay:
@mglaman did yours use the same hostname or a different hostname (this one uses a different hostname).

mglaman:
It was different. API hosted via `*.ddev.local` and I was accessing it over `localhost:8000`, which `localhost` still respects CORS (because I had to configure it.)

Or am I missing that `additional_hostnames` was populated, and somehow those don't have CORS being passed along the headers?

Because this has nothing to do with the router, but the script creating the response and what headers it should pass on.

rfay:
Yes, here they're accessing via FQDN rather than via an exposed port on localhost. I don't actually see why the router can't add the header; but I'm no expert on this.

Mogtofu33:
@mglaman you are right, CORS was working fine with  Drupal+jsonapi and with ContentaCMS (which is Drupal+jsonapi) and ContentaJS (nodejs server that proxies to Contenta CMS).
The problem was when adding a frontend to this in ddev stack (angular, react, vue...).

Now it seems _access-control-allow-origin_ propagate correctly through the router. I am a bit confused about why and what happened... But looks like the problem was in the front cors management.

rfay:
Thanks @Mogtofu33  - Your opinion is that this should be closed then?

Mogtofu33:
Yes, thanks @rfay, sorry about the time consuming on this, this high mix of different technologies confused me.

rfay:
Thanks for the thoughtful PR, we hope for more in the future! Thanks for your excellent thoughts and experience @mglaman

Moongazer:
Was this PR implemented into DDEV? I couldn't find any CORS related topics in the documentation and having the same issue:

- `apiproject.ddev.site` runs a TYPO3 API where CORS headers are set in TYPO3 config, so far OK.
- Vue frontend is developed outside of DDEV running as `localhost:8080`, normal API access works fine.
- But some Node components making requests for files containing a `Origin` header, e.g. `Origin: http://localhost:8080`, which can not be avoided.

The problem:
1) With this header set, the browser expects a `Access-Control-Allow-Origin` set in the response of that file, but
2) this header is not for direct file access (because the header is set in TYPO3, it will only send if PHP is executed).

In this case it would be great if we have the config option in DDEV to tell the webserver which header it should respond.

rfay:
@Moongazer this was closed by the author, as being unnecessary. 

I could be wrong, but it sounds like your problem is whatever you're doing with the vue header. You're running it outside of ddev? It could be run inside ddev, and you could put it on https://<yourproject.site.ddev>:8000 or whatever, and then you would actually be using ddev-router. 

Maybe you can say more about your setup, but it would be better to open a dedicated issue or use one of the other support channels, https://ddev.readthedocs.io/en/stable/#support-and-user-contributed-documentation

Moongazer:
> @Moongazer this was closed by the author, as being unnecessary.

Well, I'm still not sure if this is the case, as I run into CORS issues multiple times with different setups and try-outs. E.g. see this comment: "About CORS, you should set correct headers in webserver configuration" (https://github.com/TYPO3-Initiatives/headless/pull/157#issuecomment-683849901)... He's right, in this case I'd also like to set the header in the webserver config^^

Will it be a huge or complicated change in DDEV to give developers the freedom to adjust the nginx and/or router config to add such headers like they want?
I mean, it's a development system which should be as flexible as possible to imitate the production system in my opinion (And for the production system usually I'm able to adjust the webserver config to reflect the needed environment).

> It could be run inside ddev...

Might be true, but the frontend dev-stack can be already quite complex with all the frameworks, languages, compilers, tools, etc. so I'd prefer not to mix DDEV as another layer of complexity into it. DDEV = backend, NodeJS = frontend ... separated and simple  ;-)

rfay:
@Moongazer  If you can be very specific about the problem you're seeing, provide a demonstration case, and provide the specific suggestion, I'm happy to see this go forward. It seems like people find that it's in the headers and working OK at this point? But if you can provide the actual recreation, we can either figure out a workaround for you and others running into it or provide some kind of way to do custom config. It's definitely not easy right now, see the template https://github.com/drud/ddev/blob/master/containers/ddev-router/nginx.tmpl

Moongazer:
OK I see. Next time I stumble upon the issue I will extract/reduce it into a more simple setup for reproduction which I can describe here.

rfay:
If you think your addition should be added to the default config, we're sure open to that, just send a PR or describe the change. Also, if you think the ddev-router (which is the nginx-based reverse proxy that provides the *.ddev.site URLs) should have it, we're open to that as well.

Moongazer:
I think its not needed to be included, as the settings depending on the project IMO, e.g. I had to add `Access-Control-Allow-Origin http://localhost:8080` header because that is where I develop the frond-end. Others might use `localhost:3000` etc...

rfay:
Thanks. You might consider developing the front-end and hosting it in ddev, and then I don't think you'd see this problem anyway. It's easy to do it as a second nginx stanza in the .ddev/nginx_full, or as a second container/service.

Moongazer:
> You might consider developing the front-end and hosting it in ddev, and then I don't think you'd see this problem anyway. It's easy to do it as a second nginx stanza in the .ddev/nginx_full, or as a second container/service.

That's kind-of right for smaller projects, but because of separation, lose coupling and different dev teams (backend, frondend) I find it more "clean" and better to manage to have it in two projects/repositories (DDEV with API services, and the JS frontend completly independend of it. I made a prototype in Vue, but now outsource it to someone else, and I will not give all the backend project to them). But thanks for thinking along :)

PR #1230, opened 2018-10-25
Automatically create release on tag creation, closes #1166

### LINKED ISSUE #1166: Automatic push of release artifacts to update homebrew-ddev and homebrew-ddev-edge
A release build should update homebrew-ddev if possible. 

* It's critical that only a completely accepted build gets pushed to homebrew. It may be that it should be a manual action/push.
* Note that our homebrew approach (and its docs) are pretty rudimentary. We don't have a source build, and our actual build claims to be a source build. So homebrew-ddev may need some love there. And our understanding of homebrew. @cweagans is a great resource on this.
* It has been proposed to try to get ddev into the master homebrew catalog instead of having a tap.

andrewfrench:
## The Problem/Issue/Bug:
Moving artifacts from Circle to Github is a manual process.

After release artifacts are created, a Github release for the tag will be automatically created and the artifacts will be uploaded to the release.

A new file, `.github/RELEASE_NOTES_TEMPLATE.md`, has been added to serve as an autogenerated template for release notes that should be filled in by the release manager in the Github UI before publishing the release.

The release process documentation located at `docs/developers/release-checklist.md` has been updated to reflect changes in the release process related to this PR. The most significant change is that we should no longer publish an empty release to initiate the tag build, but we should instead create and push a tag independently of a release, then let CircleCI create and populate the release when the artifacts are built. This process is documented in the release checklist.

Proper configuration for releases can be applied to the CircleCI settings for any fork of drud/ddev, so it may be easiest to walk through the configuration and release process in real-time on a fork of this repository to confirm behavior.

## Related Issue Link(s):
#1166 

## Release/Deployment notes:
Existing environment variables should be removed from the CircleCI environment variables management system and moved to `./.circleci/.publicenv`. This file is `source`-ed at the beginning of each job to ensure the expected environment variables existing during a typical public build-and-test workflow.

A secret environment variable, **GITHUB_TOKEN**, is required by `ghr` to complete the upload process as a result of these changes. This should be managed using CircleCI and environment variables managed by the CI system should be changed to be private (that is, not shared with build from remote forks). Environment variables specific to the release process, but not strictly private, are also expected to be provided by CircleCI: **GITHUB_USERNAME** and **GITHUB_REPOSITORY**. On my fork, GITHUB_USERNAME was set to `andrewfrench` to correlate my account with the release. We can continue this with the Drud repository or create a new user (like `drud-release-bot`) to be responsible for releases.


.circleci/circle_vm_setup.sh
```diff
@@ -34,3 +33,6 @@ sudo add-apt-repository \
 sudo apt-get update -qq
 sudo apt-get install -qq docker-ce
 
+# Install ghr
+go get -u -v github.com/tcnksm/ghr
```
rfay: I think the specify-a-release-and-download-it approach would be more robust.

.circleci/config.yml
```diff
@@ -204,6 +204,20 @@ jobs:
...
           name: Artifact storage
 
+      - run:
+          command: |
+            /home/circleci/go/bin/ghr \
+            -draft \
+            -prerelease \
+            -r $GITHUB_REPOSITORY \
+            -u $GITHUB_USERNAME \
+            -t $GITHUB_TOKEN \
```
rfay: I see these variables are not set in CircleCI on the Drud/ddev config. How can they be set in Circleci without exposing them to forked builds?

release_notes.md
```diff
@@ -0,0 +1,28 @@
+# Installation/Upgrade
```
rfay: Is inclusion of this file a mistake? It has specifics about ddev v1.3.0 that I wouldn't expect to see in this PR.

andrewfrench: No, not a mistake. The thought process is in the PR description, but this gives us an opportunity to either build our release notes as we make changes and submit it automatically, or create a basic release notes template that we can then edit after the release is created . It looks like there are a couple outstanding references to "this upgrade" (referring to v1.3) that will definitely get removed, but the upgrade caveats are still relevant assuming we stick with MariaDB v10.2 in v1.4+.

rfay: OK, we'll just have to make it more vanilla, as it currently contains things that would not be good for the basics of release notes.

.circleci/config.yml
andrewfrench: The proposal in this PR is to:

1) Check public environment variables into Git and load them at the beginning of a job
2) Update Circle to make environment variables private and prevent them from being shared with forked builds

andrewfrench:
This PR forced me learn the difference between "draft" and "pre-release" in the context of Github releases. We'll want to make sure that the automated releases are marked "draft", as this means they won't be returned in the list of releases returned to an API query like...

`curl https://api.github.com/repos/drud/ddev/releases`

...until we finalize the release notes and artifacts and click "publish release". The automation in this PR creates a draft release, so we shouldn't have to worry about that. Similarly, a draft won't show up to a public (non-admin) viewer, so we can safely let the draft release wait for as long as we'd like without worrying about what a user can see or has access to download. "Pre-release" doesn't prevent a release from showing up in the releases list and indicates a non-production ready release, so that should stay unchecked unless we have a good reason to indicate that.

rfay:
I do appreciate you learning that. So now I don't have to just worry about it instead of learning it! Thanks.

rfay:
I see several references to [private environment variables](https://circleci.com/docs/2.0/oss/#private-environment-variables) [and here](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project) in the Circleci docs, but  no actual definition. It seems to me that "Private environment variable" == "Environment variable set via Circleci web UI with "Pass secrets to builds from forked pull requests" turned off. Is that also your understanding? If so, it will mean that we won't be able to test any forked PRs that deal with Pantheon or drud-s3 (or other future external interaction things). 

I do note something I never saw before - that you can inject environment variables via the CircleCI api. That approach would let us kick off a tag build with injected secret env vars.

andrewfrench:
We share the same understanding of what a private environment variable means here, but I disagree with the conclusion. 

Working off the premise that an environment variable we're willing to share with a forked build is entirely and safely public, let's just check those values into this repository and load them before or during a test from a file. I've created `.circleci/.publicenv` for this purpose.

rfay:
Well, I'm willing to share our limited-privilege Pantheon and drud-s3 credentials with the occasional crafty hacker that knows how to harvest from a forked build. Unfortunately, I'm absolutely not willing to check those into a public git repo, where they can be harvested by every github-scanner.

rfay:
An alternate variation would be to build the artifacts on tag, but to push them on a separate workflow. I am not entirely sure how a separate workflow would find them to push though.

andrewfrench:
Closing this for now, I'll circle back later.

dclear:
re-opening for current sprint

rfay:
Rebased to current master.

PR #1240, opened 2018-11-05
Improve healthchecks, especially ddev-router, fixes #774, fixes #1237

### LINKED ISSUE #774: Make ddev-router more transparent about failures

### LINKED ISSUE #1237: Failed to start ddev router
**Bugdescription**
I tried to start ddev with "ddev start" and everything starts but the ddev-router. 
It says 
> Failed to start demo: ddev-router failed to become ready: health check timed out: labels map[com.docker.compose.service:ddev-router] timed out without becoming healthy, status=unhealthy

**To Reproduce**
For me it's reproducable by just starting ddev.

**Expected behavior**
I expected it to start normally without an error.

 - Host computer OS and Version: Windows 10
 - Docker version information:
> Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Tue Aug 21 17:21:34 2018
 OS/Arch:           windows/amd64
 Experimental:      false

> Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Tue Aug 21 17:29:02 2018
  OS/Arch:          linux/amd64
  Experimental:     false

 - ddev version information:
> dba     drud/phpmyadmin:v1.3.0
router  drud/ddev-router:v1.3.0
commit  v1.3.0
domain  ddev.local
cli     v1.3.0
web     drud/ddev-webserver:v1.3.0
db      drud/ddev-dbserver:v1.3.0

 - config.yaml contents:
> APIVersion: v1.3.0
name: demo
type: typo3
docroot: public
php_version: "7.2"
webserver_type: nginx-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
additional_fqdns: []
provider: default

**Additional context**
We are using a company proxy but I dont know if it effects the problem, because it's defined in docker and everything else like pulling containers works fine.

rfay:
## The Problem/Issue/Bug:

* Our healthchecks were relying on "localhost", which requires name resolution, and we don't actually need name resolution like that for things to work.
* The router healthcheck (#774) was always questionable about giving adequate info about failures, especially missing nginx config failures.

* Use 127.0.0.1 in healthcheck
* Improve ddev-router healthcheck to check nginx config so one can use `docker inspect ddev-router` to see what went wrong.
* Add log information from container checks to give context to failures.

* Break ddev-router by configuring the same additional_hostnames[] on two projects (or `docker exec -it ddev-router bash` and edit the /etc/nginx/conf.d/default.conf to remove the healthcheck). Examine the result. 
* Break other containers with `ddev exec -it <service>` and breaking the healthcheck one way or another.

No changes were made.

## Related Issue Link(s):

OP #774: Make ddev-router more transparent about failures.

### DIFF
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -143,7 +143,7 @@ func RenderRouterStatus() string {
 	case "exited":
 		fallthrough
 	default:
-		renderedStatus = color.RedString(status) + badRouter + ":" + logOutput
+		renderedStatus = color.RedString(status) + badRouter + ":\n" + logOutput
 	}
 	return fmt.Sprintf("\nDDEV ROUTER STATUS: %v", renderedStatus)
 }
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -133,7 +133,7 @@ func findDdevRouter() (*docker.APIContainers, error) {
 func RenderRouterStatus() string {
 	status, logOutput := GetRouterStatus()
 	var renderedStatus string
-	badRouter := "\nThe router is not currently healthy. Your projects may not be inaccessible.\nTry running 'ddev start' on a site to recreate the router."
+	badRouter := "\nThe router is not currently healthy. Your projects may not be accessible.\nTry running 'ddev start' on a site to recreate the router."
 
 	switch status {
 	case SiteNotFound:
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -107,9 +107,9 @@ func StartDdevRouter() error {
 
 	// ensure we have a happy router
 	label := map[string]string{"com.docker.compose.service": "ddev-router"}
-	err = dockerutil.ContainerWait(containerWaitTimeout, label)
+	log, err := dockerutil.ContainerWait(containerWaitTimeout, label)
 	if err != nil {
-		return fmt.Errorf("ddev-router failed to become ready: %v", err)
+		return fmt.Errorf("ddev-router failed to become ready: log=%s, err=%v", log, err)
 	}
 
 	return nil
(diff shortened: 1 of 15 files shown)


pkg/ddevapp/router.go
```diff
@@ -131,9 +131,9 @@ func findDdevRouter() (*docker.APIContainers, error) {
 
 // RenderRouterStatus returns a user-friendly string showing router-status
 func RenderRouterStatus() string {
-	status := GetRouterStatus()
+	status, logOutput := GetRouterStatus()
 	var renderedStatus string
-	badRouter := "\nThe router is not currently running. Your sites are likely inaccessible at this time.\nTry running 'ddev start' on a site to recreate the router."
+	badRouter := "\nThe router is not currently healthy. Your projects may not be inaccessible.\nTry running 'ddev start' on a site to recreate the router."
```
andrewfrench: Quick typo fix here: `inaccessible` => `accessible`

pkg/ddevapp/router.go
```diff
@@ -143,30 +143,27 @@ func RenderRouterStatus() string {
 	case "exited":
 		fallthrough
 	default:
-		renderedStatus = color.RedString(status) + badRouter
+		renderedStatus = color.RedString(status) + badRouter + ":" + logOutput
```
andrewfrench: The output from `ddev describe` looks like it needs a newline here after ':'

Issue #1027, opened 2018-08-01
Consider composer cache volume


rfay:
We have lots of people using composer, both on the host and in the container. 

However, 
* If they run composer on the host, the cache is not in the container.
* If they run composer in the container, it's only cached there, and only for the lifetime of the container. Other projects get no benefit of the download.

The Stack Overflow solution [How can I use the composer cache from my host machine in DDEV containers?](https://stackoverflow.com/questions/51626715/how-can-i-use-the-composer-cache-from-my-host-machine-in-ddev-containers#51626716) provides a one-off way to do this, but it would be wonderful if we could figure out a generic way to provide this capability for people. 

There are issues with it, since we may not know where the composer cache is located on the host, although the command `composer config cache-dir` should give it to us.

rfay:
Edited the title: I think this is probably better as a docker volume than as a bind-mounted cache, especially for performance.

PR #1241, opened 2018-11-06
Add 'ddev composer' command, closes #1131 

### LINKED ISSUE #1131: Add explicit ddev composer (in container) support "ddev composer" command

macOS and Linux users often use composer on the host, and it *can* work out OK, although there are problems with matching php versions and available packages.

However, on Windows a "composer install" can result in generated code that is not even reasonable correct for the deployment environment (ddev web container or Linux server). For example, if you use `composer require drush/drush` on windows, you get completely different content for vendor/bin/drush than you would get with the same command in the container. 

As a result, it's my opinion that we should be pushing all users toward using composer in the container, using a `ddev composer` command that would support most composer actions. 

It should be able to do at least `ddev composer install`, `ddev composer upgrade` and `ddev composer create-project`. The first two are trivial (already). create-project is harder because it has to be done in a dummy directory and then the contents copied in.

I don't actually see how there's a lot of work to be done to accomplish any of these other than create-project. The others should just be pass-through commands working just like `ddev exec composer ...`

It's actually possible that we don't need anything more than a `ddev composer-create-project`, and that all the other composer commands are so easily handled with exec.

https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev shows the current state of the art; The last two links are user-provided techniques that work, but are awkward. 

**Additional context**

Related problems: 
* [As a ddev user, I want to be able to use ssh/rsync from within the web container #414](https://github.com/drud/ddev/issues/414) would allow those with private repositories to easily access them, and there are a number of people with private composer repos.
* The solution here should probably provide for a shared (mounted from host) composer cache for each php version.

andrewfrench:
## The Problem/Issue/Bug:
#1131 

Adds a `ddev composer` command, allowing for seamless use of composer within the context of the container. Specific support has been added for `ddev composer create-project`, installing the project into `/tmp/composer` and moving the resulting project into the app root when complete.

- Execute composer commands with `ddev composer`
- Create projects using `ddev composer create-project`
- Test `create-project` args and flags: one argument should be interpreted as the package, two should be interpreted as the package and the installation directory, and three should be interpreted as the package, installation directory, and the version
- Test something like `ddev composer create-project --no-dev -v typo3/cms-base-distribution myproj ^8` and ensure it installs successfully to `${APP_ROOT}/myproj`
- Test other composer commands with and without flags, like: `ddev composer require --quiet twitter/bootstrap` and confirm that commands behave as if arguments are passed directly to composer

**Note:** `ddev` will make no effort to parse any flags provided to it, they will instead be passed to composer.

composer_test.go has been added to test `ddev composer` and `ddev composer create-project` commands.

## Related Issue Link(s):
#1131


rfay:
Taking this for a quick spin, reminded me of https://github.com/drud/ddev/issues/1027 - Shared composer cache. We may want to add that to this release.

rfay:
I took this for a quick spin and things seemed to work out just fine! Haven't looked at the code at all but was just eager to take it for a spin. 

```
mkdir junk
cd junk
ddev config --projecttype=php
ddev start
# This composer create-project is better example, builds in project dir, and v9
ddev composer create-project --no-dev -v typo3/cms-base-distribution . ^9

# composer require fails
ddev composer require typo3-ter/static-info-tables
Using version ^6.7 for typo3-ter/static-info-tables
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 74 installs, 0 updates, 0 removals


  [RuntimeException]
  /vendor does not exist and could not be created.


require [--dev] [--prefer-source] [--prefer-dist] [--no-progress] [--no-suggest] [--no-update] [--no-scripts] [--update-no-dev] [--update-with-dependencies] [--update-with-all-dependencies] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [--sort-packages] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--] [<packages>]...
```

The create-project went great, but the `composer require` apparently didn't execute in the expected directory (/var/www/html in this case)

andrewfrench:
Right, the example given was more to showcase version selection and that this now supported non-cwd installations. I'm having trouble recreating that issue you ran into given the commands you ran:

```
mkdir junk
cd junk
ddev start 
ddev composer create-project --no-dev -v typo3/cms-base-distribution . ^9
ddev composer require typo3-ter/static-info-tables
```

Each of the above completed successfully. Can you confirm that after a `ddev rm -a`?

 #1027 will bring a ton of efficiency and greatly speed things up, but I think that should be its own PR.

rfay:
I can't recreate the `ddev composer require` failure with current, yay.

However, `ddev composer update --dev` fails. as does `ddev composer update --no-dev` and `ddev composer require --no-dev monolog/monolog`

I imagine that's because you haven't figured out how to get cobra to be gentle on flags it doesn't understand.

pkg/ddevapp/config_test.go
```diff
@@ -214,6 +214,9 @@ func TestConfigCommandInteractiveCreateDocrootDenied(t *testing.T) {
 	// Set up tests and give ourselves a working directory.
 	assert := asrt.New(t)
 
+	err := os.Unsetenv("DRUD_NONINTERACTIVE")
```
rfay: AH! This is the cause of our trouble with root-owned .ddev/Test*. Need to not do this!

andrewfrench: Good call - note that a util.Confirm() was added (to reduce code duplication and standardize interactions) that respects the DRUD_NONINTERACTIVE env var:

https://github.com/drud/ddev/pull/1241/files#diff-427f16bfc90f69b5bd53189bb655137bR41

So for now, I've updated this to unset and reset the variable after this test to allow the interaction output to be captured.

andrewfrench:
I believe this one's ready for review. 
Spinoff issues that I'd like to add to improve this in the next release:

1) Improving the create-project flow to be less cumbersome and (maybe?) even handle some config interactivity if no config currently exists, 
2) Sharing the composer cache between projects

docs/users/cli-usage.md
```diff
@@ -130,11 +130,11 @@ If necessary, run build steps that you may require, like `composer install` in t
...
 **Composer Setup Example**
 
-*Do not use this technique with Windows, as composer on an NTFS filesystem can result in major problems with NTFS symlinks that can't be read in the Linux web container. Instead, use the [community-provided hints on Stack Overflow](https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev)*
-
 ```
-composer create-project typo3/cms-base-distribution my-typo3-site "^9"
+mkdir my-typo3-site
 cd my-typo3-site
+ddev config --project-type php
+ddev composer create-project typo3/cms-base-distribution . ^9
```
rfay: 👍 I only wonder if the "." will confuse people, but not sure of a better way. Could use /var/www/html there. I guess let's just maybe add an explanation below it? Maybe 
```
# This does "composer create-project" in the current directory (".") (inside the container)
```

rfay: Doesn't the Drupal 8 composer example need this also?

pkg/ddevapp/config_test.go
```diff
@@ -214,6 +214,12 @@ func TestConfigCommandInteractiveCreateDocrootDenied(t *testing.T) {
 	// Set up tests and give ourselves a working directory.
 	assert := asrt.New(t)
 
+	noninteractive := "DRUD_NONINTERACTIVE"
```
rfay: I was confused by this env name variable, assuming it was the value, not the name. Silly nit, but maybe name it noninteractiveEnv?

pkg/ddevapp/config_test.go
```diff
@@ -214,6 +214,12 @@ func TestConfigCommandInteractiveCreateDocrootDenied(t *testing.T) {
...
+	// nolint: errcheck
+	defer os.Setenv(noninteractive, os.Getenv(noninteractive))
```
rfay: Nicely done.

docs/users/cli-usage.md
rfay: Probably need `ddev config --project-type typo3 --php-version 7.2` after that stuff.

cmd/ddev/cmd/composer-createproject.go
```diff
@@ -0,0 +1,129 @@
...
+				util.Failed("Failed to start app %s to run create-project: %v", app.Name, err)
+			}
+		}
+
+		// Make the user confirm that existing contents will be deleted
+		// If the user hasn't passed any positional arguments, we're not actually executing an
+		// install and nothing will be deleted -- for example, 'ddev composer create-project -h'
+		doInstall := false
+		if len(positionals) > 0 {
+			util.Warning("Warning: Any existing contents of the project root will be overwritten")
```
rfay: I noticed this warning; it's not necessarily true is it? Depending on whether they used the all-powerful "." as the destination (which is probably what they want). Maybe it needs to get the path of what it's going to blow away and provide that in the warning?

cmd/ddev/cmd/composer-createproject.go
```diff
@@ -0,0 +1,129 @@
...
+			util.Failed("Failed to execute create-project command")
+		}
+
+		if len(stdout) > 0 {
+			fmt.Println(strings.TrimSpace(stdout))
+		}
+
+		// Move project to app root if an install is underway
+		if doInstall {
+			output.UserOut.Printf("Moving installation to project root")
```
rfay: This one too might not be exactly correct. Not sure how we should work this in the very odd case where they don't do the project root. I'm kind of wondering if we should take away the option of *where* they can do the composer create-project, because if they don't use "." they're shooting themselves in the foot, and they're likely to use some example that messes them up. 

It's possible that we should adjust to the more normal examples, like `composer create-project typo3/cms-base-distribution CmsBaseDistribution ^9` and do all the things for them after that, like going into the dir, config. But that's too complex and too much change for where we are now. But people will definitely use examples like that one and get unexpected results that we'll have to support.

docs/users/cli-usage.md
rfay: And it probably needs an extra `ddev start` after the config, so it gets the right nginx config.

cmd/ddev/cmd/composer-createproject.go
andrewfrench: Well, it's true as far as whatever exists in `/var/www/html` will be explicitly `rm`ed before continuing with the installation. The catch, as I think you're pointing out here, is if a user provided an absolute path to the installation directory, then there's really no reason to delete the project root. At the very least this needs the path of what's being deleted.

rfay:
There's a problem with our Drupal example.

`ddev composer create-project drupal-composer/drupal-project:8.x-dev . --stability dev --no-interaction` somehow gets the args moved around to `Executing composer command: composer create-project --stability --no-interaction drupal-composer/drupal-project:8.x-dev /tmp/composer dev` (see dev at end).

The result is
```
  [Symfony\Component\Console\Exception\RuntimeException]
  The "--stability" option requires a value.
```

andrewfrench:
Yikes, that's some complicated parsing, but I'll admit what's being done to intercept args and put them where they need to be is perhaps overly naive and trying to do too much. I see two ways of addressing this:

- Lock down argument positions further, forcing flags to come before package/version and not allowing arbitrary directories,
- Recreating Composer's flag parsing set to ensure all possible flags get put where they need to be

The second option would be preferable for a 1:1 recreation of the composer create-project user experience, but that's a lot to ask. The first option can be communicated through usage text and effective documentation, but will likely confuse users and prevent them from running commands how they're used to.

rfay:
I would be in favor of using our own create, maybe renaming it to make it so it's clearly *not* transparent pass-through? And providing our own args/flags that work with it.  That way we're in control. And people who need to do something else can do it the old fashioned way.

rfay:
Another anomaly that needs to be discussed and/or documented: "ddev composer xxx" seems to only work in the default working directory, whereas `composer install` and friends normally work in the current directory.

andrewfrench:
This has been rebased, doc updates made, and a quick fix to `ddev composer` to ensure we're working in the app root.

rfay:
@cweagans we should have asked for your comment already on the minimal list of flags we're supporting for the `ddev composer create` command. Currently 
```
Flags:
      --dev                Pass the --dev flag to composer create-project
  -h, --help               help for create
      --no-dev             Pass the --no-dev flag to composer create-projet
      --stability string   Pass the --stability <arg> option to composer create-project
```
but we'll add `--no-interaction`

Thoughts? We want to limit the flags we pass through to `composer create-project` for maintainability.

cweagans:
I would recommend allowing users to specify extra flags to pass (`--extra-composer-args=""` or something). People use Composer in all kinds of funky ways, and someone is inevitably going to want something that isn't provided there. Composer could also add a flag someday and we shouldn't make people wait to be able to use it IMO.

docs/users/cli-usage.md
```diff
@@ -130,11 +130,11 @@ If necessary, run build steps that you may require, like `composer install` in t
...
 
-*Do not use this technique with Windows, as composer on an NTFS filesystem can result in major problems with NTFS symlinks that can't be read in the Linux web container. Instead, use the [community-provided hints on Stack Overflow](https://stackoverflow.com/questions/49660082/how-can-i-run-composer-with-ddev)*
-
 ```
-composer create-project typo3/cms-base-distribution my-typo3-site "^9"
+mkdir my-typo3-site
 cd my-typo3-site
+ddev config --project-type php
+ddev composer create typo3/cms-base-distribution ^9
 ```
```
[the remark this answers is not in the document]
andrewfrench: The diff doesn't show this, but those commands are mentioned just underneath these lines. The code snippets are broken up with some text between them:

>_From here we can start setting up ddev..._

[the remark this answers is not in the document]
rfay: Sorry, I was pretty dense. You already tried to explain that to me. I see it there in context.

rfay:
It looks like something is not quite copacetic with args. I guess the version spec must be optional?
```
$ ddev composer create drupal-composer/drupal-project:8.x-dev --stability dev --no-interaction
Error: accepts 2 arg(s), received 1
Usage:
  ddev composer create [flags] <package> <version>
```

Yup, `ddev composer create drupal-composer/drupal-project:8.x-dev --stability dev --no-interaction ^8` works. So version spec must be optional.

andrewfrench:
The version argument is now optional, so usage will be:
`ddev composer create [flags] <package> [<version>]`

Tested with TYPO3 and D8 with and without version specifications, `--no-interaction`, and other flags.

PR #1243, opened 2018-11-07
Stop mounting from global ddev directory, fixes #883

### LINKED ISSUE #883: Stop bind-mounting from global ddev directory (~/.ddev)
ddev's use of ~/.ddev for temporary storage of SSL certs and for db imports is problematic because docker can't mount network drives into containers.

We have already moved the database out of there, and that was the big thing.

Now this problem can be solved with:
1. Use a docker volume for ddev-router's SSL cert cache
2. Use project's .ddev for the temporary mount used by the db container for `ddev import-db`

This is a pretty simple approach that will solve things for our users who have network-mounted home directories. They still have to have their code/project on a regular drive, but that's easy enough to do.

----- Original issue below ----

Currently we use ~/.ddev for storage of many things. However, there are people who have that on a network drive, and it can have catastrophic performance implications.

Allow overriding the global ddev dir somehow.  It can't be in the project .ddev. 

A workaround is to set %HOMEPATH% and/or %HOMEDRIVE% on Windows, or $HOME on *nix/macOS. One could make a script that brought up a ddev-specific terminal for this purpose; it wouldn't be a very good thing to change it globally.

**Additional context**

rfay:
## The Problem/Issue/Bug:

OP #883: People who have network-mounted home directories have loads of trouble using ddev, and we don't really have any reason to be bind-mounting from there any more.

* Stop using bind-mounted router cache and use docker volume instead
* Use import-db directory in project .ddev instead of global ddev dir

Start a project
Use `docker inspect ddev-router` to check mounts
Use `docker inspect ddev-<project>-db` to check mounts

Nothing has been added. 

## Related Issue Link(s):

OP #883 

## Release/Deployment notes:

This will probably deserve a minor mention in release notes.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -221,8 +221,7 @@ func (app *DdevApp) ReadConfig() error {
 		app.DBAImage = version.GetDBAImage()
 	}
 
-	dirPath := filepath.Join(util.GetGlobalDdevDir(), app.Name)
-	app.ImportDir = filepath.Join(dirPath, "import-db")
+	app.ImportDir = app.GetConfigPath("import-db")
 
 	app.SetApptypeSettingsPaths()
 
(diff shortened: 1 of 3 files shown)


rfay:
Thanks for the good comment on ~/.ddev. IMO there's no reason to put project directories in there any more at all. I think this might be the last gasp of project directories, hoping!

Regardless of that, it's a mess.

PR #1246, opened 2018-11-08
Minor cleanup of test and buildkite setup (tests only)

rfay:
## The Problem/Issue/Bug:

We had some minor issues with tests after #1243 (stop mounting homedir) went in. There are a couple of things showing up:

* There are unnecessary attempts to clean up ~/.ddev/Test* when it shouldn't have been used.
* Only the test for upgrade of bind-mounted databases still should be looking in there. 
* Those directories are showing up as owned by root in our macOS testbots, which really shouldn't be able to happen.  (This ends up being a [PR removal of DRUD_NONINTERACTIVE](https://github.com/drud/ddev/pull/1241/files#r231994483), so will be fixed in that PR.
* There's currently no error checking on the rm -rf ~/.ddev/Test*, so the buildkite-agent startup is not detecting the root ownership problem.

Note that [TestConfigOverrideDetection](https://github.com/drud/ddev/blob/e87a4586fba2bc9d5832e5e8bddc0213b5a78daf/pkg/ddevapp/config_test.go#L543) is trying to use sudo on a build for the composer PR, see [test](https://buildkite.com/drud/ddev-macos/builds/984#5141d296-cffb-4140-be93-d7c6f8e5be40). 

* Make sure that non-removable ~/.ddev/Test* directories cause a build start failure on buildkite.
* Explicitly set DRUD_NONINTERACTIVE in buildkite setup

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -13,6 +13,7 @@ if [ ! -z "$BUILDKITE_JOB_ID" ]; then
 fi
 
 export GOTEST_SHORT=1
+export DRUD_NONINTERACTIVE=true
 
 echo "--- cleaning up docker and Test directories"
 echo "Warning: deleting all docker containers and deleting ~/.ddev/Test*"
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -20,16 +20,16 @@ if [ "$(docker ps -aq | wc -l)" -gt 0 ] ; then
 	docker rm -f $(docker ps -aq)
 fi
 docker system prune --volumes --force
-rm -rf ~/.ddev/TestPkg*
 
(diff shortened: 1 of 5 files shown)


rfay:
The DRUD_NONINTERACTIVE has always been set (by https://github.com/drud/ddev/blob/master/pkg/ddevapp/ddevapp_test.go#L132 more than anywhere else). I just freaked out and added this when the bug in that one PR caused it to be unset. Essentially, it always *must* be set for any ddev commands that do a start (or really anything that might invoke `ddev hostname`, and thus sudo). So it should be totally safe here. However... this *was* reactionary, and we *did* find the cause. So if you'd prefer I'll pull them back out.

PR #1253, opened 2018-11-13
add _apt user back in to /etc/passwd so apt operations do not exit wi…

nedsbeds:
Hi,
The /etc/passwd files in the ddev web image overwrites the _apt user (and various others)
https://github.com/drud/ddev/blob/master/containers/ddev-webserver/files/etc/passwd

This appears to be causing various odd issues such as when trying to add additional packages to the container once booted since there is no sandbox _apt user and the OS cannot create any more users

**To Reproduce**
`ddev start`
`ddev . sudo apt-get update`

response:
`W: No sandbox user '_apt' on the system, can not drop privileges`

```
# ddev version
dba   	drud/phpmyadmin:v1.3.0
router	drud/ddev-router:v1.3.0
commit	v1.3.0
domain	ddev.local
cli   	v1.3.0
web   	drud/ddev-webserver:v1.3.0
db    	drud/ddev-dbserver:v1.3.0
```

This PR adds back the _apt user

Thanks

* `ddev ssh`
* `sudo apt-get update && sudo apt-get install htop`
* You should see htop get installed with no error and no complaint about _apt being missing.

### DIFF
--- a/containers/ddev-webserver/files/etc/passwd
+++ b/containers/ddev-webserver/files/etc/passwd
@@ -1,5 +1,4 @@
 root:x:0:0:root:/home:/bin/bash
-_apt:x:100:65534::/nonexistent:/bin/false
 nginx:x:0:0:root:/home:/bin/bash
 uid_0:x:0:0:root:/home:/bin/bash
 uid_1:x:1:1:gid_1:/home:/bin/bash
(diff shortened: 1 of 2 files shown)


containers/ddev-webserver/files/etc/passwd
```diff
@@ -60001,4 +60002,3 @@ uid_59997:x:59997:59997:gid_59997:/home:/bin/bash
 uid_59998:x:59998:59998:gid_59998:/home:/bin/bash
 uid_59999:x:59999:59999:gid_59999:/home:/bin/bash
 uid_60000:x:60000:60000:gid_60000:/home:/bin/bash
-nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
```
rfay: It's much more important to keep "nobody" than to have "apt"

containers/ddev-webserver/files/etc/passwd
```diff
@@ -1,4 +1,5 @@
 root:x:0:0:root:/home:/bin/bash
+_apt:x:100:65534::/nonexistent:/bin/false
```
rfay: You can have two usernames with the same uid here. SO let's put _apt with uid_100. Not sure why uid 100 had group 101.

PR #1262, opened 2018-11-14
Create a new randomized installation directory for 'ddev composer create'

andrewfrench:
## The Problem/Issue/Bug:
Re-using and clearing out a common tmp install directory caused issues on Windows.

Creates a new directory for each installation attempt.
This also adds support for the `--prefer-dist` flag.


andrewfrench:
This is looking promising on my Windows 10 Home setup with Docker for Windows 18.06.1-ce-win73.

cmd/ddev/cmd/composer-create.go
```diff
@@ -119,22 +117,26 @@ project root will be deleted when creating a project.`,
 			composerCmd = append(composerCmd, "--no-interaction")
 		}
 
+		if preferDistArg {
+			composerCmd = append(composerCmd, "--prefer-dist")
+		}
+
 		composerCmdString := strings.TrimSpace(strings.Join(composerCmd, " "))
 		output.UserOut.Printf("Executing composer command: %s\n", composerCmdString)
 		stdout, _, err := app.Exec(&ddevapp.ExecOpts{
```
rfay: This really should grab the stderr and report it if there is any. Alternately, you can explicitly set ExecOpts.NoCapture and let it just go out to them. If you think Exec() isn't handling stderr correctly, that's another thing and we can open an issue for it.

cmd/ddev/cmd/composer-create.go
```diff
@@ -119,22 +117,26 @@ project root will be deleted when creating a project.`,
...
 			Service: "web",
 			Cmd:     composerCmd,
 		})
 		if err != nil {
-			util.Failed("Failed to execute create-project command")
+			util.Failed("Failed to create project")
```
rfay: Let's at least add the error code. If the actual output is being reported elsewhere then that's ok.

cmd/ddev/cmd/composer-create.go
```diff
@@ -119,22 +117,26 @@ project root will be deleted when creating a project.`,
...
 		}
 
 		if len(stdout) > 0 {
 			fmt.Println(strings.TrimSpace(stdout))
 		}
 
 		output.UserOut.Printf("Moving installation to project root")
-		bashCmdString := fmt.Sprintf("if [ -d %s ]; then mv %s /var/www/html/; fi", installDir, path.Join(installDir, "*"))
+		bashCmdString := fmt.Sprintf("if [ -d %s ]; then mv %s /var/www/html/; fi", containerInstallPath, path.Join(containerInstallPath, "*"))
```
rfay: Now it makes me think... Why don't we just do *this* on the host as well? Probably faster and more reliable.

cmd/ddev/cmd/composer-create.go
```diff
@@ -143,13 +145,16 @@ project root will be deleted when creating a project.`,
 			util.Failed("Failed to create project: %v", err)
```
rfay: Isn't this message incorrect? Also, if it's just a fail to mv, probably not fatal any more?

cmd/ddev/cmd/composer-create.go
```diff
@@ -170,4 +175,5 @@ func init() {
 	ComposerCreateCmd.Flags().BoolVar(&noDevArg, "no-dev", false, "Pass the --no-dev flag to composer create-project")
 	ComposerCreateCmd.Flags().StringVar(&stabilityArg, "stability", "", "Pass the --stability <arg> option to composer create-project")
 	ComposerCreateCmd.Flags().BoolVar(&noInteractionArg, "no-interaction", false, "Pass the --no-interaction flag to composer create-project")
+	ComposerCreateCmd.Flags().BoolVar(&preferDistArg, "prefer-dist", false, "Pass the --prefer-dist flag to composer create-project")
```
rfay: awesome, thanks.

cmd/ddev/cmd/composer-create.go
rfay: util.Failed though will mean that the existing directories don't get cleaned up. So we should probably defer the directory cleanup instead of doing it inline?

cmd/ddev/cmd/composer-create.go
andrewfrench: Definitely agreed, I'll add this

cmd/ddev/cmd/composer-create.go
```diff
@@ -69,18 +75,6 @@ project root will be deleted when creating a project.`,
...
-		_, _, err = app.Exec(&ddevapp.ExecOpts{
-			Service: "web",
-			Cmd:     []string{"sh", "-c", fmt.Sprintf("rm -rf %s", installDir)},
-		})
-		if err != nil {
-			util.Failed("Failed to create project: %v", err)
-		}
-
 		// Remove any contents of project root
 		util.Warning("Removing any existing files in project root")
```
rfay: This isn't actually true is it? We *overwrite* but don't currently remove any existing files otherwise? Maybe the warning should be "Existing files may be overwritten" or perhaps we should actually delete everything but .ddev.

cmd/ddev/cmd/composer-create.go
andrewfrench: My thinking was that, at this point, the project hasn't been installed to the expected directory and likely isn't easily startable by a typical user, since the source code and docroot will be hidden inside a .tmp... dir. That would mean that the top-level intention of the command has failed, with the specifics detailed in the error text.

rfay: Well, the mv will now be handled on the host, right, per other discussion. So we can give more details in the error reporting of that code.

cmd/ddev/cmd/composer-create.go
andrewfrench: We are explicity removing everything except .ddev in /var/www/html here, though.

rfay: Ah, not we're definitely not removing everything. It does not remove dot-files or dot-directories. But I see it does everything else. (I did lose track of context, but also this is not removing like the old failed composer installs, which isn't a big deal). But it will be much improved by doing this on the host.

andrewfrench:
Alright, lots of testing over here on the subject of capturing stderr. It turns out that Composer's "streaming" create-project update text, as well as any error text, is coming through on stderr. Capturing that and printing it to the terminal after the command has completed results in double-printing the progress information and double-printing any error that was encountered, definitely undesirable behavior. The bright side is that this confirms that stderr isn't being suppressed, and we shouldn't have to explicitly print err.Error() if the Composer command fails, because the error information should already be the last thing printed.

Stdout is not, by default, printed to the terminal. If ExecOpts.NoCapture is set to true, stdout and stderr are both printed to the terminal in real-time, but Exec() will now no longer return an error if the underlying command fails. I don't think any of this is necessarily how things should work, but the necessary changes will have to be made in Exec() and its dependencies.

cmd/ddev/cmd/composer-create.go
```diff
@@ -119,38 +127,33 @@ project root will be deleted when creating a project.`,
...
+			util.Failed("Failed to create project")
 		}
 
 		if len(stdout) > 0 {
 			fmt.Println(strings.TrimSpace(stdout))
 		}
 
 		output.UserOut.Printf("Moving installation to project root")
-		bashCmdString := fmt.Sprintf("if [ -d %s ]; then mv %s /var/www/html/; fi", installDir, path.Join(installDir, "*"))
+		bashCmdString := fmt.Sprintf("if [ -d %s ]; then mv %s /var/www/html/; fi", containerInstallPath, path.Join(containerInstallPath, "*"))
```
rfay: Since this mv was one of the first and most important failures we saw, could we do it in golang on the host as well?

rfay: I was thinking it could be done the way you did the initial directory cleanup: Walk the members of the subdirectory, and os.Rename() each.

andrewfrench: This has been updated to walk the tmp dir and rename its files, good suggestion!

rfay: Thanks, I think we'll all like that better in the end.  We'll want to move this to fileutil a bit later, but we don't have to do it for this release.

You can't rename a dir to move it huh? Yuckie Shucks.

PR #1269, opened 2018-11-16
Make TestComposerCmd run faster, fixes #1268

### LINKED ISSUE #1268: Use a simpler and faster composer build in TestComposerCmd()
**Describe the bug**

We do a full TYPO3 composer build in TestComposerCmd, which takes a really long time. We can demonstrate the build much faster with a smaller composer build.

rfay:
## The Problem/Issue/Bug:

The full TYPO3 composer build in first pass of TestComposerCmd took more time than warranted.

Use a simpler build


rfay:
This should be fine, and we've figured out what was going on, so that's a win. Ready for review.

PR #1272, opened 2018-11-16
Add support for pre-gzipped CSS and JS files

schliesser:
## The Problem/Issue/Bug:

Like in the TYPO3 .htaccess file which supports gzip compression this should also be supported by the nginx configuration.
Solution found in Slack posted by @supergrav https://typo3.slack.com/archives/C8TRNQ601/p1528290423000106

Adds headers for .js.gzip and .css.gzip files

Run a TYPO3 project in ddev with nginx-fpm
Include some gzipped files
```
page.includeJSFooter.jsFile = EXT:example/Resources/Public/JavaScript/file.js.gzip
page.includeCSS.cssFile = EXT:example/Resources/Public/Styles/file.css.gzip
```
Network analyze in any browser if the files get deflated correctly

## Related Issue Link(s):
none

### DIFF
--- a/containers/ddev-webserver/files/etc/nginx/nginx-site-typo3.conf
+++ b/containers/ddev-webserver/files/etc/nginx/nginx-site-typo3.conf
@@ -48,6 +48,16 @@ server {
         fastcgi_param SERVER_NAME $host;
         fastcgi_param HTTPS $fcgi_https;
     }
+    
+    # Compression: Support gzipped js/css files
+    location ~ \.js\.gzip {
+        add_header Content-Encoding gzip;
+        add_header Content-Type application/javascript;
+    }
+    location ~ \.css\.gzip {
+        add_header Content-Encoding gzip;
+        add_header Content-Type text/css;
+    }
 
     # Expire rules for static content
     # Feed


containers/ddev-webserver/files/etc/nginx/nginx-site-typo3.conf
```diff
@@ -48,6 +48,16 @@ server {
         fastcgi_param SERVER_NAME $host;
         fastcgi_param HTTPS $fcgi_https;
     }
+    
+    # Compression: Support gzipped js/css files
```
rfay: Shouldn't this be 3 rules? One that does gzip (and adds Content-Encoding) and one each for js and css?

schliesser:
We have several projects with precompiled css & js files. They get minified and gzipped during deployment process. Allmost all projects are running on apache servers where we use the default TYPO3 .htaccess file compression settings (see: https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_8-7/_.htaccess#L36 ).
My goal was to achieve the same with the nginx config in ddev.

But maybe you are right and it's time to think the other way round and let the web server do the compression. I'll discuss this also with my team.

schliesser:
We did some testing and discussed this in our team. Thx for your advise, that nginx/apache will do the compression automatically. So this PR can be closed without merging.

rfay:
Thanks so much for the PR in the first place, and if you decide this is a common TYPO3 approach, we're happy to support it.

PR #1306, opened 2018-11-26
Use gotestsum for test output on circleci

rfay:
## The Problem/Issue/Bug:

gotestsum looks like a promising formatter for test output. 
https://circleci.com/blog/level-up-go-test-with-gotestsum/
https://github.com/gotestyourself/gotestsum

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
     - run:
         command: make -s test
         name: ddev tests
-        no_output_timeout: "40m"
+        no_output_timeout: "120m"
     - store_test_results:
         path: /tmp/test-results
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -42,6 +44,8 @@ jobs:
         command: make -s test
         name: ddev tests
         no_output_timeout: "40m"
+    - store_test_results:
```
jrthrawny: This needs to point to the full path of where the gotestsum command placing the XML. Or we can add a step after make test to mv/cp the file over

rfay:
Well, this is pretty. (https://circleci.com/gh/drud/ddev/6440). But I'm not sure what it *did* with the uploaded test results, or why it matters.

rfay:
OK, let's try this out. @andrewfrench seemed to like the format. We'll just pull it out if we don't like it.

One thing I did here that might result in issues is just increasing the test timeout quite a lot, and I notice we actually do have a hanging test (TestCaptureLogs) on Apache. So that one will pain us when it uses up a whole 2 hours. Could be a mistake.

PR #1307, opened 2018-11-28
Use sentry to log usage events, fixes #307

### LINKED ISSUE #307: Gather (Optional/Opt-in) Usage Statistics to Inform Roadmap Decisions

* Feature Request

Given that we do not put up any barrier to downloading/using ddev, it would be valuable to have some mechanism to (optionally and with the ability to opt-in/out) have a sense of the total number of active ddev users. Additionally, we could go a step further and gather statistics on the frequency of commands used. Obviously, we will want to sanitize the specific parameters specified so that there are no information disclosures. Still, it would be valuable to know what percentage of end-users use `ddev exec` and how frequently they use it.

More work and discussion would need to happen before this becomes actionable.

Any step in this direction would require us to have an explicit policy around data collection, retention.

rfay:
## The Problem/Issue/Bug:

* We want opt-in stats and crash reports for ddev OP #307 

* Add global configuration (~/.ddev/global_config.yaml)
* Add `ddev config global` as command for people to change global config
* Add opt-in prompt
* Integrate with Sentry for most commands (reports the command and environment) and also for panics and the like. 
* Add docker and docker-compose version to `ddev version`
* `ddev version` now presents sorted, stable list of values.

* This must be built either by CircleCI or by make with `export SentryDSN=<dsn_value>` set
* Get SentryDSN at https://sentry.io/settings/drud/ddev/keys/
* Do some things, create some errors
* Visit the ddev dashboard on Sentry (you can log in with github credentials)
* Validate what you find there.

Misc:
* Remove ~/.ddev/global_config.yaml and verify expected prompting behavior'
* Change last_run_version there to verify expected prompting behavior
* Set omit_containers to omit dba and ddev-ssh-agent in the global config, verify behavior.
* Look at `ddev version` and verify content
* Experiment with `ddev config global`, introducing errors, etc.

Still needs tests

## Related Issue Link(s):

OP #307

ddev_rm


andrewfrench:
Initial tests with this shows ddev prompting as expected when:
- ~/.ddev/global_config.yaml is removed
- ~/.ddev/global_config.yaml is updated with a stale last_used_version

Errors seem to be making it to Sentry reliably and the values are accurate, looks super cool!

rfay:
This is ready for review, both from the user-interaction perspective and code and results. 

* Here are [artifacts for testing](https://circleci.com/gh/drud/ddev/6718#artifacts/containers/0)
* Please pay careful attention to the opt-in experience
* You should see your events and errors show up in https://sentry.io/drud/ddev/
* Other testing instructions in the OP

Of course if users want it a different way, we can change it in a following release. We don't want to annoy anybody, but we can also respond  to concerns within a month or so. But we *do* want to see initial value (and no negative user experience) immediately.

unn:
I have the same concerns from earlier that we're misusing Sentry for statistics gathering. 

@rfay can you point me to documentation with regard to how to test using the artifacts from CircleCI?

rfay:
Totally reasonable concerns. The goal is to just get started here at this point.

To test using the artifacts posted, download ddev and put it in your path. I assume you're on macOS? If so, `brew unlink ddev` and put the downloaded untarred ddev from the mac tarball into your path, or use the [script](https://ddev.readthedocs.io/en/stable/#manual-installation-or-upgrade-linux-and-macos) to install. Ping me if you need more guidance. Thanks for testing!

unn:
👍 with the getting started goal. 

Thanks, that's easy. I didn't know if I needed to grab the container images or something else.

andrewfrench:
An interesting thing I'm consistently noticing when reporting errors &mdash; and I'm curious if you're seeing the same thing on your machine, @rfay &mdash; is that the reporting process seems to drastically inflate the execution time of a simple ddev command.

Some examples, with no extant projects (all commands fail):
### Wifi on, reporting opt-in
`ddev rm aaa` => 10.725s
`ddev start bbb` => 11.243s
`ddev config --php-version abcd` => 10.859s

### Wifi off, reporting opt-in
`ddev rm aaa` => 0.453s
`ddev start bbb` => 0.467s
`ddev config --php-version abcd` => 0.443s

### Wifi on, reporting opt-out
`ddev rm aaa` => 0.429s
`ddev start bbb` => 0.425s
`ddev config --php-version abcd` => 0.030s

### Wifi off, reporitng opt-out
`ddev rm aaa` => 0.432s
`ddev start bbb` => 0.437s
`ddev config --php-version abcd` => 0.030s

Is it possible this is the result of us using a Sentry trial with reduced performance, or could this be overhead introduced by Raven? My connection while testing is fairly strong, pinging sentry.io at about 110ms round-trip and 20+ Mbps.

Also, when testing the wifi off/reporting opt-in combination, the following error was printed to the terminal only with the `ddev config --php-version abcd` command, not the `ddev start` or `ddev rm` commands:
```
Post https://sentry.io/api/160826/store/: dial tcp: lookup sentry.io on [::1]:53: read udp [::1]:54219->[::1]:53: read: connection refused
```

rfay:
@andrewfrench I think this does what you rightly suggested now. Thanks.

PR #1315, opened 2018-11-30
Document how to add a beanstalk queue, fixes #1313

### LINKED ISSUE #1313: Add Beanstalk Queue
We're developing an application that generates tasks to a queue. There are multiple workers that resolve one task at a time.

To ensure the tasks are handled in a First-In-First-Out order - and to avoid handing the same task to two parallel running workers, we've settled for [Beanstalk](https://beanstalkd.github.io/) as the queue of choice.

I think it would be cool to have a simple possibility to include a beanstalk-service to ddev, similar to the additional services like Solr and Memcached as described on https://ddev.readthedocs.io/en/stable/users/extend/additional-services/

For now, a possible workaround would be to run the queue worker via CLI to test the execution of the single tasks. This is no problem via `ddev ssh` and issuing the command to the application. But writing new tasks to or fetching the next task from the queue would probably need changes to our application (to circumvent the situation that there just is no Beanstalk-service available within ddev).

mrimann:
Resolves #1313


rfay:
Thanks for this! It looks like it was pretty straightforward.

PR #1318, opened 2018-12-03
Provide mariadb 10.1 and 10.2, fixes #1221

### LINKED ISSUE #1221: ddev should support  both mariaDB 10.1 and 10.2

rfay:
## The Problem/Issue/Bug:

It turns out TYPO3 v8 and earlier require Mariadb 10.1 and can't function with 10.2. This PR provides explicit 10.1 support.

The first attempt at this was #1222, which tried it using library/mariadb, but that made images that were twice as big and swich to debian base image. Too much change for too little value. This one maintains two almost identical images.

* 10.1 and 10.2 are now available with setting `mariadb_version: 10.1` etc in config.yaml
* You can also `ddev config --mariadb-version=10.1` and `ddev start`.
* You can also restore 10.1 snapshots (pre v1.3) with it running in 10.1 mode.
* An attempt to start a 10.2 database with 10.1 mariaDB will fail and the logs will tell why (and it fails before the db can be damaged).
* An attempt to restore a 10.1 snapshot with 10.2 is warned.
* An attempt to restore a 10.2 snapshot with 10.1 is warned.

To test out the mariadb 10.1 db, please:
• dump your db to make a backup: `ddev export-db >dump.sql.gz`
• `ddev remove --remove-data` which will create an automatic snapshot also (or `ddev remove --remove-data --omit-snapshot` if you don't want it.
• `ddev config --mariadb-version=10.1`
• `ddev start`

* There's a new test just for 10.1 that goes over the basics of database usage.

## Related Issue Link(s):

OP #1221


rfay:
I didn't mean to put so much effort into this, but am pretty happy with the end result. It even has error checking on wrong-db-version and wrong-snapshot type. Even though the audience is probably small, they'll appreciate it. And we have enough test coverage for it I think. And if we're forced to add another type, the model is at least in place.

PR #1322, opened 2018-12-07
Fix reported panic in GetContainerHealth()

rfay:
## The Problem/Issue/Bug:

@mafriend reported a panic when testing ddev-ui for release. Full information wasn't available, but this was the gist of it:

screen_shot_2018-12-07_at_1 01 23_pm

Check the error before using the results.


rfay:
I brought this in because I'm using `ddev log --continuous` in testing other PR and I'd like to make sure this panic goes away.

PR #1323, opened 2018-12-08
Add utilities and tests for CIFS simulated symlinks, add to ddev composer, fixes #1283 

### LINKED ISSUE #1283: ddev composer still doesn't quite do symlinks that result in proper behavior
**Describe the bug**

In v1.4.0 we introduced `ddev composer`, whose intent is to provide in-container predictable behavior, especially on Windows.

Inside the linux containers, creating a symlink appears to work (ln -s), and most linux commands respect it.

Sadly, php does not respect the "XSym" file that is created. So if you try to read a symlink with PHP, you'll get the contents of the XSym file, not the contents of the target file.

**To Reproduce**

* On Windows, in the container, on the mounted NTFS volume, create a file named README.md with the contents `This is just a simple readme for testing purposes.`
* create a symlink to the README.md. For example, `ln -s README.md mysymlink.md`
* Read the file with cat. It will work fine. `cat mysymlink.md` will result in the proper output (the content of README.md).
* Read the file with PHP instead, using this script:
```
<?php

if ($fh = fopen('mysymlink.md', 'r')) {
    while (!feof($fh)) {
        $line = fgets($fh);
        echo $line;
    }
    fclose($fh);
}
```

**Expected behavior**

php should read the *target* file instead. Which is what it does if you create a proper symlink on the Windows host:

* Developer mode must be turned on on the host (See https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/)
* In git-bash `export MSYS=winsymlinks:nativestrict`
* ln -s README.md mysymlink.md
* Read the file with the same php script inside the container. It gets the correct content.

**Additional context**

* @ohader notes that this project creates unusable symlinks even when run inside the container: https://github.com/HofUniversityIndieOne2018/red-cross-project
* @ohader has also written a script (to be run on the host, after composer install, https://gist.github.com/ohader/f4b792e0bb99e79efe75605008ce4a0d#file-use-native-symlinks-sh, to relink XSym as "junction symlinks"
* This [old forum post](https://forums.docker.com/t/symlinks-on-shared-volumes-not-supported/9288/45) gives some context about "junction links"
* This docker-for-win issue seems to deal with the same problems: https://github.com/docker/for-win/issues/137

rfay:
## The Problem/Issue/Bug:

OP #1283: On Win10 with docker-for-windows, the CIFS filesystem inside the container uses *simulated* XSym symlinks, which mostly work inside the container, are completely incomprehensible on the host, and don't *completely* work, especially on nontrivial TYPO3 projects (see OP for complete details).

Mostly this doesn't affect most Drupal projects, and doesn't affect plain-vanilla TYPO3 projects.

However, Windows 10 on the *host* has the capability of creating real, valid symlinks that are perfectly fine inside the container. But only if it's set to "developer" mode.

* Add simulated XSym symlink detection and replacement tools (and tests)
* Add to the "composer" commands a run through of replacing, if on docker-for-windows.

* With a Windows 10 Pro system with developer mode, use `ddev composer install` or `ddev composer create` to create something that has symlinks. It should notify you of the creation of the symlinks.
* Without developer mode, try the same thing; it should warn about the issue.

* The utilities added have their own tests.
* The composer tests already do what's needed

Probably we should add a symlink to the repos used in the composer tests.

## Related Issue Link(s):

OP #1283


docs/users/developer-tools.md
```diff
@@ -28,6 +28,16 @@ To execute a fully-featured `composer create-project` command, you can execute t
 
 `ddev exec composer create-project ...`
 
+#### Windows OS and `ddev composer`
```
andrewfrench: Came for instructions on how to enable developer mode, but there aren't any!

rfay: Search for "developer" in settings. Thanks for pointing this out :) But if you want to continue it's not hard.

andrewfrench: I did eventually find it, but wasn't sure if it was a Docker setting or a Windows setting. Just referring to this:

`please see the docs to enable developer mode: https://ddev.readthedocs.io/en/latest/users/developer-tools/#windows-os-and-ddev-composer`

rfay: See if this works for you, https://github.com/drud/ddev/blob/ab4965c15cdc5b893e2224deef4ebe136dcef729/docs/users/developer-tools.md#windows-os-and-ddev-composer

andrewfrench: Woah, screenshots and everything! Awesome.

PR #1324, opened 2018-12-09
Check to see if two paths are the same in app.Init, fixes #1168

### LINKED ISSUE #1168: Accessing project via non-canonical path makes ddev fail
**Describe the bug**

On macOS with case-insensitive filesystem, 
* create a project named "junk" in ~/Sites
* `ddev start`
* Now cd to ~/sites/junk
* `ddev rm` - Will fail `Unable to get project(s): a project (web container) in running state already exists for junk that was created at /Users/rfay/Sites/junk`

You can get the exact same thing to happen with a symlink. 
* ln -s ~/Sites/junk /tmp/junk
* cd /tmp/junk
* `ddev rm`

The reason we have this is that we have a test to make sure people don't run two projects of the same name (at least at the same time). 

But it doesn't check for an actual "official" location (following a symlink or ???). If we could add just a bit of resiliency to that check by following symlink or perhaps getting golang to give us the official full path of the file we'd be out of hell here.

rfay:
## The Problem/Issue/Bug:

There have been lots of confused people over time with symlinks and with directories in case-insensitive filesystems.

* `ddev start` in ~/Sites/junk when the project is running in ~/sites/junk (even though on case-insensitive filesystem they're the same)
* `ddev start` in ~/somelinkwhen project is running in  ~/Sites/other but ~/somelink is s symlink to ~/Sites/other

People are easily trained out of it, but it's been a support issue more than once.

Check to see if the two named paths are actually the same before throwing an error.

Try the examples listed in the summary above.

* Adds test for IsSameFile(), the new utility.
* Adds testing in TestDdevStart for this case.

## Related Issue Link(s):

OP #1168


rfay:
To pass on Windows this will need creating the symlink for the test in the test, rather than trying to use symlinks in git. I made the mistake of setting `core.symlinks=true` for git on the testbots and it blew things sky-high for all windows tests. Not sure why.

PR #1327, opened 2018-12-12
Fix problem where ddev-ssh-agent volume causes error, fixes #1312

### LINKED ISSUE #1312: If ddev-ssh-agent is not running, the volume won't exist, and we still depend on it.
**Describe the bug**

If you start clean after a docker factory-reset AND have disabled the ddev-ssh-agent either globally or in the first project, the ddev-ssh-agent_socket_dir  volume won't have been created and you get this error:

`
Volume ddev-ssh-agent_socket_dir declared as external, but could not be found. Please create the volume manually using `docker volume create --name=ddev-ssh-agent_socket_dir` and try again.
Failed to start d8git: Failed to run docker-compose [-f C:\Users\rfay\workspace\d8git\.ddev\docker-compose.yaml up -d], err='exit status 1', stdout='', stderr='Volume ddev-ssh-agent_socket_dir declared as external, but could not be found. Please create the volume manually using `docker volume create --name=ddev-ssh-agent_socket_dir` and try again.'
`

**To Reproduce**

* docker volume rm ddev-ssh-agent_socket_dir 
* omit_containers["ddev-ssh-agent"] in global config or in the only project to run.
* ddev start

**Expected behavior**

Normal behavior, but without the ssh-agent

rfay:
## The Problem/Issue/Bug:

OP #1312 there was an edge case that resulted in a complaint about ssh-agent volume not existing.

If omit_containers["ddev-ssh-agent"] then don't refer the the volumes

* Make sure normal use of ssh authing works correctly. `ddev auth ssh` and access a host requiring ssh auth from within the web container.
* Start with omit_containers: ["ddev-ssh-agent"]  in the ~/.ddev/global_config.yaml; the ssh agent should not start, and you should get no errors.
* Start with omit_containers: ["ddev-ssh-agent"] at the project level. You should get no errors.

Nothing was added for this one.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -469,8 +469,8 @@ type composeYAMLVars struct {
 	ComposeVersion  string
 	MountType       string
 	WebMount        string
-	IncludeDBA      bool
-	IncludeSSHAgent bool
+	OmitDBA         bool
+	OmitSSHAgent    bool
 	WebcacheEnabled bool
 	IsWindowsFS     bool
 }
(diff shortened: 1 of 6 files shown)


pkg/ddevapp/config.go
```diff
@@ -463,6 +463,10 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	if util.ArrayContainsString(app.OmitContainers, "dba") {
 		includeDBA = ""
 	}
+	var includeSSHAgent = "includeItByDefault"
```
rfay: My bet is you can school me in a cleaner way to do this in the template @andrewfrench . For some reason I just hate golang templating and haven't achieved much competency.

rfay:
Thanks so much for the excellent cleanup on this. That stuff preceded even me, and I just worked with it over time. This is much better.

Rebase, though was pretty intense because of the important stuff coming in from bgsync. Could you please take quick look through the code @andrewfrench ?

Thanks!

pkg/ddevapp/templates.go
```diff
@@ -155,11 +165,13 @@ networks:
 volumes:
   mariadb-database:
     name: "${DDEV_SITENAME}-mariadb"
+  {{ if .OmitSSHAgent }}
```
andrewfrench: Needs a `not` here

rfay: Wow, and here I thought I tested that successfully. Thanks.

PR #1332, opened 2018-12-14
Regression: docker and docker-compose versions empty in `ddev version`

rfay:
## The Problem/Issue/Bug:

I noted that `ddev version` had empty lines for docker and docker-compose versions (recently added).

Use the getter for them instead of using the variable directly.

Sad dance to resolve import cycles. Our packages are kind of out of control.

Use `ddev version`

Added trivial stuff in the related test functions.


pkg/version/version.go
```diff
@@ -95,8 +99,12 @@ func GetVersionInfo() map[string]string {
 	versionInfo["commit"] = COMMIT
 	versionInfo["domain"] = DDevTLD
 	versionInfo["build info"] = BUILDINFO
-	versionInfo["docker"] = DockerVersion
-	versionInfo["docker-compose"] = DockerComposeVersion
+	if versionInfo["docker"], err = GetDockerVersion(); err != nil {
+		_ = fmt.Errorf("failed to GetDockerVersion(): %v", err)
```
andrewfrench: These error values are formatted, but never returned.

PR #1345, opened 2018-12-19
Allow a user to remove a project by name after deleting project directory, closes #1338

### LINKED ISSUE #1338: Unable to remove a project by name when the project directory no longer exists
**Describe the bug**
If the project directory is deleted while the app is running, you'll need to `ddev rm -a` to remove the project.

**To Reproduce**
For a running project in `~/Sites/abc`, 
- `rm -rf ~/Sites/abc`
- `ddev rm abc`

```
Unable to get project(s): failed to get abc:  is not a valid project name. Please
enter a project name in your configuration that will allow for a valid hostname.
See https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames for
valid hostname requirements
```

**Expected behavior**
`ddev rm <project>` should remove the named project because we can still query project containers by their tags.

andrewfrench:
## The Problem/Issue/Bug:
#1338 describes how a project can't be deleted by name (`ddev remove <name>`) after the project's directory has been removed. 

This updates the `remove` command to no longer use a convenience method that returns full-fledged apps because that method tried to read the config to fully initialize the app. This isn't necessary when removing an app, so now we're just pulling existing containers by names found in the container tags.

`mkdir ~/project`
`cd ~/project`
`ddev config --project-type php`
`ddev start`
`cd ~`
`rm -r ~/project`
`ddev remove project`

A test case has been added to ensure a project can be removed by name after deleting the project directory. Other test cases have been added to ensure that `start`, `stop`, and `describe` return consistent help text when the project directory is missing.

## Related Issue Link(s):
#1338

### DIFF
--- a/cmd/ddev/cmd/remove.go
+++ b/cmd/ddev/cmd/remove.go
@@ -39,36 +39,32 @@ To snapshot the database on remove, use "ddev remove --snapshot"; A snapshot is 
 "ddev remove --remove-data" unless you use "ddev remove --remove-data --omit-snapshot".
 `,
 	Run: func(cmd *cobra.Command, args []string) {
-		// Prevent users from destroying everything
-		if removeAll && removeData {
-			util.Failed("Illegal option combination: --all and --remove-data")
-		}
 		if createSnapshot && omitSnapshot {
 			util.Failed("Illegal option combination: --snapshot and --omit-snapshot:")
 		}
 
-		apps, err := getRequestedApps(args, removeAll)
+		projects, err := getRequestedProjects(args, removeAll)
 		if err != nil {
-			util.Failed("Unable to get project(s): %v", err)
+			util.Failed("Failed to get project(s): %v", err)
 		}
 
-		// Iterate through the list of apps built above, removing each one.
-		for _, app := range apps {
-			if app.SiteStatus() == ddevapp.SiteNotFound {
-				util.Warning("Project %s is not currently running. Try 'ddev start'.", app.GetName())
+		// Iterate through the list of projects built above, removing each one.
+		for _, project := range projects {
+			if project.SiteStatus() == ddevapp.SiteNotFound {
+				util.Warning("Project %s is not currently running. Try 'ddev start'.", project.GetName())
 			}
 
 			// We do the snapshot if either --snapshot or --remove-data UNLESS omit-snapshot is set
-			doSnapshot := ((createSnapshot || removeData) && !omitSnapshot)
-			if err := app.Down(removeData, doSnapshot); err != nil {
-				util.Failed("Failed to remove project %s: \n%v", app.GetName(), err)
+			doSnapshot := (createSnapshot || removeData) && !omitSnapshot
+			if err := project.Down(removeData, doSnapshot); err != nil {
+				util.Failed("Failed to remove project %s: \n%v", project.GetName(), err)
 			}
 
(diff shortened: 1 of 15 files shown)


cmd/ddev/cmd/remove.go
```diff
@@ -39,17 +39,38 @@ To snapshot the database on remove, use "ddev remove --snapshot"; A snapshot is
 "ddev remove --remove-data" unless you use "ddev remove --remove-data --omit-snapshot".
 `,
 	Run: func(cmd *cobra.Command, args []string) {
-		// Prevent users from destroying everything
-		if removeAll && removeData {
```
andrewfrench: This is no longer an illegal combination thanks to the database snapshot feature that's a default on `--remove-data`.

rfay: So it's fine to do `ddev rm -a --remove-data` ? Also fine to do `ddev rm -a --remove-data --omit-snapshot` That actually sounds pretty useful.

andrewfrench: Right. Before snapshotting was the default, `ddev rm -aR` was a great way to lose a bunch of data accidentally, but snapshots should prevent that. `ddev rm -aRO` will blow everything away, but it requires actively asking for snapshots to be skipped.

cmd/ddev/cmd/remove_test.go
```diff
@@ -54,3 +46,23 @@ func TestDevRemove(t *testing.T) {
 	err = addSites()
 	require.NoError(t, err)
 }
+
+// TestDdevRemoveDeletedDir tests that a project can be `ddev remove`d when its project directory has been removed
+func TestDdevRemoveDeletedDir(t *testing.T) {
```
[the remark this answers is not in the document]
andrewfrench: This PR (at the moment) only deals with collecting projects passed in as args, so it's kind of arg processing -- maybe one step removed. The test you're referring to is testing `app.Stop()`, which is important because it's wrapping a `docker-compose` command that needs project configuration files to get started, so it will return an error if the project directory is missing. The TestDdevStopMissingDirectory test is passing because the error is getting returned as expected. The `ddev remove` command will _attempt_ an `app.Stop()`, but when it returns an error because the project directory is missing, will emit a warning and continue to remove the containers without using `docker-compose`. 

That's not to say they shouldn't behave consistently, just explaining why it's like that now. Some commands operate on containers directly, some rely on `docker-compose`, I'll see if it needs to be that way.

andrewfrench:
This ended up becoming a bit larger than I'd hoped, but not as huge a change set as it could have been. I expect some tests to fail, but I believe this is okay for another spin.

rfay:
I'd been delaying looking at this because I thought it had legit test failures, but looks like it's pretty clean now with just windows nastiness?

It does look like there's an attempt to remove files before the test has let go of them, which you can't do on Windows, https://buildkite.com/drud/ddev-windows-dockerforwindows/builds/1674#a2dfff69-1e1d-46cd-9359-9780d2a0108f ?

andrewfrench:
Yep, just one last test update to appease Windows. I'll ping you when that's pushed and tested, working on that now.

rfay:
I haven't finished manual testing on this but expect to do so today and get it in.

rfay:
Going with it! Let's retest it carefully in release test.

PR #1347, opened 2018-12-20
Simplify nightly build, build PRs in nightly build, fixes #1165

### LINKED ISSUE #1165: Each open PR should have a nightly build in Circleci
Per our Q4 OKR list, we want to have a nightly build for each open PR. 

So for each open PR:
* Circleci should run a nightly_build (which can be reorganized if we want)
* buildkite should run a build for each test

If this is too aggressive we can not do it all. I had originally just planned the first bullet.

There are frictional issues with builds that we should be thinking about as well. For example, CircleCI does not do a container build except in the nightly build. Much of our CircleCI approach hasn't been rethought for a really long time. 

(This OP can be edited as we discuss this one.)

rfay:
## The Problem/Issue/Bug:

The nightly build is broken, I think mostly because of the version-replacement activities in the extensive nightly build.

I don't think we need that stuff any more with current strategy.

* Remove the special build for nightly and just use an ordinary build.
* Build all PRs nightly, fixes #1165


rfay:
This did *not* cause the nightly to run on this PR last night. But it might not have. We probably have to pull it. Will probably need a followup to make it possible to run it using api. Requesting @andrewfrench review.

PR #1349, opened 2018-12-21
Use pv to show progress bar during import-db

### LINKED ISSUE #1348: Add progress bar during import-db
Importing multi gigabyte-sized databases can take several minutes or more and there is no feedback given until it is complete.

Add a progress bar to the `import-db` command so the user has an idea of how long it will take.

longwave:
## The Problem/Issue/Bug:
There is no feedback during long running database imports.

This PR uses `pv` instead of `cat` to read the database dump which provides a progress bar on stderr.

- Import a (preferably large) database dump file with `ddev import-db` and check that a progress bar is displayed.
- Import a database dump with `ddev import-db -p=false` and check the progress bar is hidden.

Changes are visual only, and require a tty, so automated testing is quite tricky.

## Related Issue Link(s):
Fixes #1348

## Release/Deployment notes:
Automated processes that use `ddev import-db` should not be affected as we detect presence of a tty and disable the progress bar in these environments. There is also the `-p=false` switch to disable the progress bar if needed.

### DIFF
--- a/cmd/ddev/cmd/import-db.go
+++ b/cmd/ddev/cmd/import-db.go
@@ -11,6 +11,7 @@ import (
 
 var dbSource string
 var dbExtPath string
+var progressOption bool
 
 // ImportDBCmd represents the `ddev import-db` command.
 var ImportDBCmd = &cobra.Command{
@@ -41,7 +42,7 @@ can be provided if it is not located at the top level of the archive.`,
 			}
 		}
 
-		err = app.ImportDB(dbSource, dbExtPath)
+		err = app.ImportDB(dbSource, dbExtPath, progressOption)
 		if err != nil {
 			util.Failed("Failed to import database for %s: %v", app.GetName(), err)
 		}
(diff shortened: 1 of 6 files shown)


longwave:
The tests fail because they are not connected to a tty, so it looks like I need to use github.com/mattn/go-isatty to detect whether stdout is a tty or not, and call `Exec` or `ExecWithTty` as appropriate. This might also be a good point to add a switch to disable progress even if the tty is present.

rfay:
You're awesome, thanks for the PR! Ambitious!

longwave:
A followup PR could refactor away `ExecWithTty` as `Exec` now has a tty option, and only the ssh command uses `ExecWithTty`.

rfay:
Assuming this is just a test issue, happy to help when and if you're ready for help.

longwave:
Any review comments or help with test fixes would be appreciated. I don't have time to look at this again until next year now.

longwave:
@rfay The lint and compile steps seem to be working but it looks like the CircleCI tests don't rebuild the container images, so `pv` is not available during the tests?

rfay:
What we have to do is push a new version of the container, and update in pkg/version/version.go with the new version. I just pushed drud/ddev-dbserver:20190103_import-db-progress-10.1 and drud/ddev-dbserver:20190103_import-db-progress-10.2, please add a commit to pkg/version/version.go that updates the version there (temporarily, until next release), and then you'll get the right container.

longwave:
@rfay I can't see the buildkite results so don't know why the Windows builds have failed. The dockerforwindows one is especially suspicious at only 2 seconds runtime.

rfay:
I've got those tests restarted for you, looks like you're mighty close!

Thanks for the awesome work on this. I'm looking forward to taking a look at it.

rfay:
You're awesome, thanks so much for this excellent contribution @longwave - I like it!

PR #1351, opened 2018-12-21
Fix apache-fpm behavior with php7.3, fixes #1346

rfay:
## The Problem/Issue/Bug:

OP #1346: The ddev-webserver didn't come up at all when configured with apache-fpm and php7.3. This was due to missing apache config for php7.3

## How this PR Solves The Problem:

* Add in the missing files
* Add a full complement of nginx/apache/php tests into the container test.
* Bump the web container version number
* Remove symlinks that were checked in for the apache section of the web container. These didn't get brought into the docker build right (they came in a symlinks to paths with "\" in them). Previously we hadn't really been testing the Apache configurations, and this PR adds tests for every permutation of apache and php, so it uncovered the Windows build problem.   (This is what all the removals in containers/ddev-webserver/files/etc/apache2/conf-enabled are about, they're replaced with `a2enmod` command in start.sh.)
* Minor change to sanetestbot.sh to resolve a problem where it couldn't mount a directory which was a symlink.
* Added .gitattributes to explicitly control line endings in this repo; they never need to be anything but \n. The creeping-in of wrong linefeeds was caught in another PR. (I tried to solve this in a separate PR but ended up having to transplant it here to get a clean build)
* Remove "indexes" from apache config, since it should *not* be indexing directories with no index.php/index.html. This broke one of the tests.

## Manual Testing Instructions:

* Set your config.yaml for php7.3 and apache-fpm
* Configure your .ddev/config.yaml with `webimage: drud/ddev-webserver:20181221_php73_apache`
* `ddev start` should bring it up successfully.

## Automated Testing Overview:

This adds container-level tests for the full matrix of php vs webserver versions.

## Related Issue Link(s):

OP #1346

PR #1354, opened 2018-12-22
Ensure that our containers only have files with LFs in them, not CRLF

rfay:
## The Problem/Issue/Bug:

We've always had the problem that if Windows CRLFs get into scripts on the containers then those scripts are broken. 

Add a .gitattributes that ensures that checkouts of all text files in the containers directory will be treated as Linux-style files. 

Now a user can `git config --global core.autocrlf true` and things will work out OK. This also means we should be able to remove the requirement for special configuration of the testbots.

* Check out the repo on Windows. You should see only LFs inside text files in containers/ (And there's a commit there with files to look at). You can use `od -xc <file>` to see the actual linefeeds. Also vim will typically show the file as "dos" if it has CRLF. I'm sure there are other ways.
* Check out the repo on mac or Linux. You should see only LFs 
* Adding files with CRLF (on windows) in directories other than containers should be handled with autocrlf true. Adding files in the containers area should end up with commits that convert them to LFs.

No testing change, but this *should* mean that we can remove the requirement for special git setup on windows testbots.

## Release/Deployment notes:

Before pull, get rid of the test commit.

### DIFF
--- a/.buildkite/test_containers.sh
+++ b/.buildkite/test_containers.sh
@@ -39,11 +39,13 @@ cleanup
 
 set -x
 
-for dir in containers/*
-    do pushd $dir
-    echo "--- Build container $dir"
-    time make container DOCKER_ARGS=--no-cache
-    echo "--- Test container $dir"
-    time make test
-    popd
+for dir in containers/*; do
+    if [ -d "$dir" ] ; then
+        pushd $dir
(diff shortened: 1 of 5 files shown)


rfay:
I ended up having to introduce this and work through it in #1351, so closing here.

Issue #1380, opened 2019-01-15
[BUG] AdditionalConfiguration.php array_merge() throws warning due to not existing array


tomasnorre:
**Describe the bug**

When making a fresh installation with TYPO3 7.6, and there is no LocalConfiguration.php present or the configuration `$GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default']` is not set. 

Then the warning is shown: 
`Warning: array_merge(): Argument #1 is not an array in /var/www/html/public/typo3conf/AdditionalConfiguration.php on line 9`

**To Reproduce**
1. Copy the content from https://gist.github.com/tomasnorre/79f69249fb8a47d2795a342ed72f1fe1 to composer.json in a empty local directory
2. run `composer update --no-dev`
3. `ddev config`
4. `ddev config --mariadb-version=10.1` <-- Due to TYPO3 problem with mariadb 10.2+
5. `ddev start`
6. Go throug the TYPO3 First Install setup
7. Go to the Install Tool, and you will see the warning in the top, 

I've seen it on more places, but here is shown every time. 

**Expected behavior**
No warning should be shown

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: macOS Mojave 10.14.2
 - Docker version: 
```
Client: Docker Engine - Community
 Version:           18.09.1
 API version:       1.39
 Go version:        go1.10.6
 Git commit:        4c52b90
 Built:             Wed Jan  9 19:33:12 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.1
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       4c52b90
  Built:            Wed Jan  9 19:41:49 2019
  OS/Arch:          linux/amd64
  Experimental:     true
```
 - ddev version: 
```
bgsync        	drud/ddev-bgsync:v1.5.0
cli           	v1.5.0
commit        	v1.5.0
db            	drud/ddev-dbserver:v1.5.0-10.2
dba           	drud/phpmyadmin:v1.5.0
ddev-ssh-agent	drud/ddev-ssh-agent:v1.5.0
docker        	18.09.1
docker-compose	1.23.2
domain        	ddev.local
router        	drud/ddev-router:v1.5.0
web           	drud/ddev-webserver:v1.5.0
```

 - config.yaml contents for the misbehaving project
```
APIVersion: v1.5.0
[...]

PR #1381, opened 2019-01-15
[BUGFIX] Removes warning from empty DB Connections array (#1380)

tomasnorre:
## The Problem/Issue/Bug: 
Removes the warning when the DB Connection array is empty

Checking is the array is set if not set, an empty array is assigned to allow the `merge_array()`

Run through the Steps to reproduce, and one shouldn't see the warning anymore. 

## Related Issue Link(s):
#1380

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -19,8 +19,9 @@ const typo3AdditionalConfigTemplate = `<?php
  */
 
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['trustedHostsPattern'] = '.*';
+$ORIGINAL_CONFIG = isset($GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default']) ? $GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default'] : [];
 
-$GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default'] = array_merge($GLOBALS['TYPO3_CONF_VARS']['DB']['Connections']['Default'], [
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for this. We had never actually intended to support anything before TYPO3 v8, but if this does no harm and gets things OK for v7, I see no reason not to.  Are there other things you have to do (besides mariadb 10.1) to have v7 work ok?

tomasnorre:
I wasn't aware of the "not supporting" TYPO3 v7.. No I didn't do anything else besides the MariaDB 10.1 version. 

I'm going to set up a TYPO3 v8 and v9 too, so I can test this before we merge anything, to make sure it doesn't break anything else.. But it shouldn't harm other versions (famous last word) ;) 

I'll get back to you when  I have this tested, thanks for fast reply.

tomasnorre:
This is working on TYPO3 v8 and v9 as well, but gives some issue on Settings up TYPO3 that shouldn't happen. 

This problem only appear on TYPO3 v7, as the database setup changed in TYPO3 v8, so I'll close this PR and have it patched locally only where needed.

rfay:
I'm definitely open to supporting TYPO3 v7 and very much appreciate your contribution here. We probably should have made separate support for v8 and v9 so they could have different code, and v7 would have been easier to do there. 

Anyway, if you think this is worth fighting and getting to work, I'm sure there's a way to do it.

You should know that the more common approach to various not-fully-supported CMSs or platforms is just to use project type "php", which tells ddev not to try to fiddle with the settings.

tomasnorre:
"Where there is will there is a way", so it's just a matter if we want to follow up on this. 

TYPO3 v7 is now in ELTS which means, that most people will move away from it over time. 
https://typo3.org/cms/roadmap/

So for now I think it's ok. If someone else reaches out to your/drud regarding this, feel free to reach out to me. I'll be happy to help supporting this.

PR #1392, opened 2019-01-20
Don't ask about sentry if DDEV_NO_SENTRY

rfay:
## The Problem/Issue/Bug:

It's just annoying to be asked about sentry when you're developing and you do a new build. No reason to ask if DDEV_NO_SENTRY.

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -151,7 +151,7 @@ func sentryNotSetupWarning() {
 // from the last saved version. If it is, prompt to request anon ddev usage stats
 // and update the info.
 func checkDdevVersionAndOptInSentry() error {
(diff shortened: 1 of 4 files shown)


cmd/ddev/cmd/root.go
```diff
@@ -151,7 +151,7 @@ func sentryNotSetupWarning() {
 // from the last saved version. If it is, prompt to request anon ddev usage stats
 // and update the info.
 func checkDdevVersionAndOptInSentry() error {
-	if !output.JSONOutput && version.COMMIT != globalconfig.DdevGlobalConfig.LastUsedVersion && globalconfig.DdevGlobalConfig.InstrumentationOptIn == false {
+	if !output.JSONOutput && version.COMMIT != globalconfig.DdevGlobalConfig.LastUsedVersion && globalconfig.DdevGlobalConfig.InstrumentationOptIn == false && os.Getenv("DDEV_NO_SENTRY") == "" {
```
andrewfrench: I think it'd be a good idea to define the environment variable name as a constant in values.go because it's hard-coded in several places, but that's a small point.

rfay:
Good idea @andrewfrench - please see how this looks.

PR #1396, opened 2019-01-23
Implement nfs mounted project directory, fixes #976

### LINKED ISSUE #976: Use NFS mounts instead of osxfs
As you may know or experience by yourself volumes mounted by Docker for Mac's _osxfs_ provide slow I/O performance. There is a lengthy topic about this at the [Docker Forum](https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076). 

Even if osxfs seems to do well reading files using the _cached_ configuration it still is slow at writes compared to NFS. Also - as I understand it - _cached_ may have it's drawbacks like delayed synchronisation of host changes to containers.

Are there any plans to use NFS mounts instead of osxfs which seems to be possible since Docker for Mac (around) 18.04? Or is it just not worth it because _cached_ already works well enough?

rfay:
## The Problem/Issue/Bug:

As requested in OP #976, an NFS mount of the project directory gives an enormous speed boost on both MacOS and Windows (it's probably unnecessary on Linux, but there's no reason not to support it). 

Add app features and  config to use NFS. 
Inspired by https://medium.com/@sean.handley/how-to-set-up-docker-for-mac-with-native-nfs-145151458adc, but works on Windows too, should work everywhere.

* Configure NFS server with proper share
* Use `nfs_mount_enabled: true` in config.yaml
* `ddev start`
* Try to break things.

## Related Issue Link(s):

* OP #976 
* I expect that the experimental webcache_enabled feature will be obsoleted by this feature. They are incompatible with each other (cannot be selected at the same time)

## Release/Deployment notes:

* Testing requires NFS server running on testbots (so that's more docs)

## TODO:

* Docs on testbot setup
* User-oriented docs, probably including the mac setup and a windows setup.
* Scripts for users to use to set up NFS


rfay:
Latest build artifacts: https://circleci.com/gh/drud/ddev/8864#artifacts/containers/0

rfay:
@andrewfrench I'd love it if you could take this for a spin. See [basic docs](https://github.com/drud/ddev/blob/eeb9732351f98a4e77b341d041b7b23fc5787c76/docs/users/performance.md).  The manual testing instructions are mostly good enough. There's a script for each OS to set up NFS.

rfay:
[Latest build artifacts](https://circleci.com/gh/drud/ddev/8990#artifacts/containers/0); If WIndows, use the installer, but it's not signed, so you have to say "more info" when it gets blocked, and then "run anyway". 

See [docs](https://github.com/drud/ddev/blob/eeb9732351f98a4e77b341d041b7b23fc5787c76/docs/users/performance.md) for guidance.

wizonesolutions:
I followed the instructions here and enabled NFS. I'm able to `ddev ssh` and get dropped in the project directory.

Summarizing caveats I ran into:

`/etc/exports` shouldn't have anything below `/Users` in it. It might if e.g. Vagrant has modified the file previously. I had to empty it _before_ running the setup script. `nfsd` was silently failing to start otherwise.

macOS: The terminal environment the script gets run in will probably need Full-Disk Access in the Security > Accessibility > Full Disk Access settings.

This feature will probably conflict with Vagrant's NFS feature, since it exports paths more granularly. This will probably present issues for people transitioning from Vagrant to DDEV. Perhaps DDEV could try to detect these conflicts or embed the setup script into a command and prompt the user to try running it if the web container fails to start and NFS is configured, and then do a more robust check.


I also tried restoring a snapshot. It seemed to work. It errored in connecting to the DB, but I think that was a timing issue, as I was able to connect to it from the web container using `drush sqlc` (and from PhpStorm).

So the basic use case seems OK.

jfastnacht:
Have used the artifacts from here: https://circleci.com/gh/drud/ddev/9034#artifacts/containers/0

OS: Windows 8.1

I've set `COMPOSE_HTTP_TIMEOUT` to 180 after it told me to increase this value on the first try.

```
λ ddev start
Starting phpbenchmark...
Removing ddev-phpbenchmark-web
ddev-phpbenchmark-db is up-to-date
ddev-phpbenchmark-dba is up-to-date
Recreating 19bac77686f6_ddev-phpbenchmark-web ... error

ERROR: for 19bac77686f6_ddev-phpbenchmark-web  Cannot start service web: error while mounting volume '/mnt/sda1/var/lib/docker/volumes/ddev-phpbenchmark_nfsmount/_data': error while mounting volume with options: type='nfs' device=':/C/Users/***/phpbenchmark' o='addr=192.168.99.1,hard,nolock,rw': connection timed out

ERROR: for web  Cannot start service web: error while mounting volume '/mnt/sda1/var/lib/docker/volumes/ddev-phpbenchmark_nfsmount/_data': error while mounting volume with options: type='nfs' device=':/C/Users/***/phpbenchmark' o='addr=192.168.99.1,hard,nolock,rw': connection timed out
Encountered errors while bringing up the project.
Failed to start phpbenchmark: Failed to run docker-compose [-f C:\Users\***\phpbenchmark\.ddev\docker-compose.yaml up -d], err='exit status 1', stdout='', stderr='Removing ddev-phpbenchmark-web
ddev-phpbenchmark-db is up-to-date
ddev-phpbenchmark-dba is up-to-date
Recreating 19bac77686f6_ddev-phpbenchmark-web ... error

ERROR: for 19bac77686f6_ddev-phpbenchmark-web  Cannot start service web: error while mounting volume '/mnt/sda1/var/lib/docker/volumes/ddev-phpbenchmark_nfsmount/_data': error while mounting volume with options: type='nfs' device=':/C/Users/***/phpbenchmark' o='addr=192.168.99.1,hard,nolock,rw': connection timed out

ERROR: for web  Cannot start service web: error while mounting volume '/mnt/sda1/var/lib/docker/volumes/ddev-phpbenchmark_nfsmount/_data': error while mounting volume with options: type='nfs' device=':/C/Users/***/phpbenchmark' o='addr=192.168.99.1,hard,nolock,rw': connection timed out
Encountered errors while bringing up the project.'
```

Not sure how to check what the actual problem is.

rfay:
@jfastnacht It almost certainly means that 1) winnfsd is not running or 2) Your firewall is blocking it. Could you please use the services app to check the status of nfsd and then turn off your firewall temporarily and restart nfsd service?

If you didn't already, please work through the [debugging](https://github.com/drud/ddev/blob/eeb9732351f98a4e77b341d041b7b23fc5787c76/docs/users/performance.md#debugging-ddev-start-failures-with-nfs_mount_enabled-true) section in the docs.

jfastnacht:
Classical case of RTFM, sorry. It's working as intended with Docker Toolbox on Windows 8.1. Decreased the loading time of an old Symfony 3 project from ~1600ms to less than 100ms. That's pretty neat, I guess. :)

rfay:
@jfastnacht Yay, great to hear it, and I love it too, especially on Windows, where I had little hope of it working with bgsync/webcache. 

Could you say a little about how the docs could be improved so you wouldn't have hit this?

I do plan to add a debug command that will check whether nfs is functioning adequately.

jfastnacht:
> Could you say a little about how the docs could be improved so you wouldn't have hit this?

The docs are fine, it was just me not reading them in the first place. I have skipped that bullet point in the Manual Testing Instructions.

asacolips:
Chiming in that performance is great in my testing! Drupal install time for a standard profile dropped from 2m28s to 1m5s, and I imagine the gap will only get larger as I test more complicated setups. Seems to be comparable to my performance with Vagrant.

One caveat that I ran into is that the entry for /etc/exports that allows NFS to work doesn't play well with Vagrant setups that are also using NFS on MacOS, since Vagrant also needs access to the /Users directory and that entry sort of hijacks it. The caveat/workaround that I used was to move my ddev testing project into a different directory, like `/DdevSites/some-project` rather than `/Users/myuser/projects/some-project` and then set up the nfs share for `/DdevSites`.

rfay:
Thanks so much @asacolips - We will probably just have to document that people can manage NFS as they choose, since there's not a cross-platform way to manage exports. 

Could you edit your review to say what OS you tested on? Thanks!

asacolips:
Updated. After additional testing, the main issue is that the NFS exports can't overlap. So in my case, I could set up a structure like the following and then do an NFS export for `/Users/myuser/projects/docker`.

* /Users
    * myuser
        * projects
            * docker
                * ddev-project
                * another-ddev-project
            * vagrant
                * vagrant-project
                * another-vagrant-project

My exports entry then looks like this:

```
# Docker
/Users/myuser/projects/docker -alldirs -mapall=502:20 localhost
# Vagrant
/Users/myuser/projects/vagrant/vagrant-project 192.168.33.01 --alldirs -mapall=502:20
/Users/myuser/projects/vagrant/another-vagrant-project 192.168.33.01 --alldirs -mapall=502:20
```

rfay:
Thanks @asacolips - @wizonesolutions reported the same thing. I think we're just going to have to say "You are in charge of setting up NFS as you'd like to set it up, the script is just for those who are only using ddev with NFS". I don't think there's a way we could change exports on the fly (per project) with Windows, and don't think that would necessarily solve the overlap problem anyway.

Bottom line is that for your use, you can use any export strategy you like, as long as the ddev project needed is exported somehow.

rfay:
Outstanding, thanks for the update!

PR #1413, opened 2019-01-30
Specify how project names work with Pantheon

jodyHamilton:
I had a hard time entering a project name as I did not understand that it is not the human readable name of my project in the Pantheon dashboard but rather the machine name used in the Pantheon URL. This is probably an issue for sites with spaces in human name, not sure about capitalization. Let's better explain what is meant by 'name' here.


rfay:
We'll pull it as soon as you sign the CLA, thanks @jodyHamilton

PR #1428, opened 2019-02-05
Minor: Better volume cleanup before running tests and full remove

rfay:
## The Problem/Issue/Bug:

I noticed a comment in the TestMain for ddevapp about deleting more volumes when webcache PR landed. This just addresses that.

Oops, also in the process discovered a bug where we don't properly remove the unisoncatalogvol in ddev rm -RO


rfay:
That was a pretty minor fixup, no need to waste extra time on review. I took one more look before pulling.

PR #1438, opened 2019-02-08
Push al(most) all artifacts to where we want them to go

rfay:
## The Problem/Issue/Bug:

Releases are getting more complicated, and we need to make them documented in code and automated as much as possible.

* Does the windows signing of the ddev_windows_installer
* Creates a new release and pushes artifacts to github.
* Tries to push to Chocolatey (but per below there may be a blocking issue)

TODO:
- [ ] Push artifacts to bintray (or else figure out how homebrew can use github release artifacts)
- [ ] See if there's any solution to the chocolatey problem.

## Release/Deployment notes:

Automation of the chocolatey push may be blocked by this issue with mono: https://github.com/mono/mono/issues/10192,


rfay:
@andrewfrench I think I'm out of gas on this one, so if you have time for a little look it would be great. I was hoping to get the bintray Homebrew stuff in, but I guess that will have to be manual this time.

Issue #1410, opened 2019-01-29
Allow overriding config.yaml so that .ddev config can be meaningfully committed to Git


wizonesolutions:
**Is your feature request related to a problem? Please describe.**
If I commit the DDEV directory to Git with, say, `webcache_enabled: true`, then my colleague on Linux will have to turn it off (there is no point in using it on Linux). Same with the upcoming NFS feature.

Ideally, I could specify:

````yaml
additional_config:
  - ./local.config.yaml
````

and ignore the local config in Git. Then in my local config I could turn on webcache or NFS, or set up additional domains that not everyone needs, etc.

**Describe the solution you'd like**
````yaml
additional_config:
  - ./local.config.yaml
````

Additional config is read after `.ddev/config.yaml`.

**Describe alternatives you've considered**
A private Git branch?
Just letting the config file be dirty on my local?
Not committing `.ddev`?

**Additional context**
N/A

rfay:
Just FYI, in the next release webcache_enabled is ignored on Linux and Windows. The issue was https://github.com/drud/ddev/issues/1395, the PR was https://github.com/drud/ddev/pull/1390

Does that make this a dup of #1395?

wizonesolutions:
Not quite. That's only one example. Another colleague wanted to change
ports on his system, a need I don't have on mine.

rfay:
Somebody suggested a config.local.yaml, it's a good idea.

wizonesolutions:
A hard-coded filename would work too, and that might be an easier starting
point. Then everyone knows what it's called.

wizonesolutions:
Another thought here is per-platform config.

e.g. webcache on macOS, NFS on Windows, neither on Linux. I know some settings will soon be ignored based on platform, but that would still then use native file syncing on Windows if webcache was configured.

`config.{linux,windows,mac}.yaml` would help people managing dev environments for other developers (for minimum friction) to provide more ready-to-use configuration, so that `config.local.yaml` would truly be for edge cases where the developers knew what they were doing, like using a different hostname or ports or whatever. I do admit that this is primarily file syncing-motivated, though, as developers could probably just agree on ports and hostnames and then not have to override them. But file syncing is less of an edge case in a multi-platform environment.

kaystrobach:
go with a convention please, no additional config

rfay:
@kaystrobach Could you propose a convention?  That would be awesome.

kaystrobach:
I’d vote for config.local.yaml


[...]

PR #1446, opened 2019-02-12
Add webcache_excludes option to config.yaml

unherz:
## The Problem/Issue/Bug:
As mentioned in issue #1410 there is no feature to to override the config.yaml webcache_enabled property.

It adds the opportunity to exclude specific operation systems from webcache enabled even if the option is enabled in general

Add webcache_excludes options as mentioned in the updated docs/users/performance.md 

## Related Issue Link(s):
[#1410 ](https://github.com/drud/ddev/issues/1410)


unherz:
@
Can sb tell me why the following build checks are failing? (Im not allowed to lookup the details)

- `buildkite/ddev-windows-apache-fpm`
- `buildkite/ddev-windows-dockerforwindows`
- `buildkite/ddev-windows-dockerforwindows-nfs`

rfay:
I'd be happy to re-run the tests, but this PR is obsolete in v1.6.0, which just ignores webcache_enabled on Windows and Linux.  This issue was resolved in https://github.com/drud/ddev/pull/1390

I'd love to see you work on a general solution to local overriding of config (https://github.com/drud/ddev/issues/1410) but please work to propose an approach there before writing a PR, since we want it to work for everybody (and not make the config more complex).

PR #1453, opened 2019-02-13
Fix Windows NFS script when username has spaces, fixes #1452

### LINKED ISSUE #1452: windows_ddev_nfs_setup.sh issue with spaces in path name
**Describe the bug**
If the Windows username has a space in it, the nfsd service will not be able to start due to the `pathFile` argument not being escaped properly.

**To Reproduce**
1. Run `windows_ddev_nfs_setup.sh`
2. You get the error "nfsd: Unexpected status SERVICE_PAUSED in response to START control".

**Expected behavior**
The nfsd service should start without issues.

**Screenshots**

 - Host computer OS and Version: Windows 10
 - Docker version 2.0.2
 - ddev version 1.6.0

rfay:
## The Problem/Issue/Bug:

OP #1452 reports that the Windows NFS installation script fails when there are spaces in the user's names.

Use $HOME instead of ~, and put path in quotes.

### DIFF
--- a/scripts/windows_ddev_nfs_setup.sh
+++ b/scripts/windows_ddev_nfs_setup.sh
@@ -30,7 +30,7 @@ if ! command -v winnfsd.exe >/dev/null; then
     echo "winnfsd.exe does not seem to be installed or is not in the PATH"
     exit 101
 fi
-winnfsd=$(command -v winnfsd.exe)
+winnfsd=$(where winnfsd.exe)
 
 if [ -f "$HOME/.ddev/nfs_exports.txt" ]; then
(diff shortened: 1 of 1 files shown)


rfay:
Per #1452 this needs to be tested on git-bash; look at the results there as well.

rfay:
Please test https://github.com/drud/ddev/blob/e0fdc174813b2c5e0ef5aa8267ef6a9a022ce547/scripts/windows_ddev_nfs_setup.sh [raw](https://github.com/drud/ddev/raw/e0fdc174813b2c5e0ef5aa8267ef6a9a022ce547/scripts/windows_ddev_nfs_setup.sh)

PR #1475, opened 2019-02-26
Bump mariadb to 10.1.38-r1 and 10.2.22-r0

rfay:
## The Problem/Issue/Bug:

I don't know why we can't use plain pinned mariadb versions in alpine/apk, but until we learn how, we have to bump when the upstream changes.

Bump mariadb to 10.1.38-r1 and 10.2.22-r0


rfay:
As far as I can tell, neither of these actually works with mariabackup, and I know they made a change to mariabackup in these versions. https://jira.mariadb.org/browse/MDEV-18129 - I tried the --backup-encrypted=false, with no success.

rfay:
Giving up on Alpine/apk for now, so closing in favor of #1476

PR #1484, opened 2019-03-05
Make very long server names possible

janhelke:
## The Problem/Issue/Bug:

Having long server names causes the ddev-router to cease working.
Error message during start:
```nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64```

This patch increases server_names_hash_bucket_size to the maximum possible size (128).

Create a server name longer than 32 characters and you should see, that none of your projects will start again.

### DIFF
--- a/containers/ddev-router/nginx.tmpl
+++ b/containers/ddev-router/nginx.tmpl
@@ -38,6 +38,9 @@ map $http_upgrade $proxy_connection {
   '' close;
 }
 
+# Apply fix for very long server names
+server_names_hash_bucket_size 128;
+
 # Default dhparam
 # ssl_dhparam /etc/nginx/dhparam/dhparam.pem;
 
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -67,7 +67,7 @@ var BgsyncTag = "v1.6.0" // Note that this can be overridden by make
 var RouterImage = "drud/ddev-router"
 
 // RouterTag defines the tag used for the router.
-var RouterTag = "v1.6.0" // Note that this can be overridden by make
+var RouterTag = "20190309_longer_server_names" // Note that this can be overridden by make
 
 var SSHAuthImage = "drud/ddev-ssh-agent"
 
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -43,7 +43,7 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "20190305_longer_nginx_server_names" // Note that this can be overridden by make
+var WebTag = "v1.6.0" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"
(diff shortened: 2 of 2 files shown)


rfay:
Thanks for this! How long are your server names? :)

janhelke:
Currently? 45 chars. But as you might know that will grow :-)

janhelke:
Anything I can do about the failing tests?

kaystrobach:
Jan i thnik you need to sign the sla ... then it will be fine.

andrewfrench:
@janhelke I'm looking into the test failures, they don't seem to be related to your changes.

rfay:
Appveyor isn't yet supported in master (and may not be) so it's no-worries. 

I pushed a web container version with this in it: drud/ddev-webserver:20190305_longer_nginx_server_names

It can now be tested with that by altering the config.yaml.

We need an edit in pkg/version/version.go pointing to that tag though please (20190305_longer_nginx_server_names) .

rfay:
I took the liberty of pushing in that one commit to push the new docker container version tag, hope that's OK.

rfay:
Artifacts for manual testing are at https://circleci.com/gh/drud/ddev/10522#artifacts/containers/0

janhelke:
I can confirm that after inserting the modified webimage in the affected config.yaml, shutting down all containers and restart, everything works fine.

rfay:
Hmm, trying to test using `name: a1234567890b1234567890c1234567890d1234567890e1234567890f1234567890` in .ddev/config.yaml, and I note that while curl, ping, and telnet can handle the resultant hostname, both Chrome and Safari (on macOS) refuse to resolve it. So that could be a problem.

rfay:
Note that testing this PR requires a `ddev rm -a` before the `ddev start` with the new version; otherwise the new router isn't generated.

janhelke:
Sorry. My bad. I tried the webcontainer change and was under the impression, that this would help. But   it doesn't. So from my point of view no change on the webcontainer is needed, only the one on the router.

rfay:
It was my bad. I have now pushed the router, and will post the artifacts for ddev shortly (you have to use the ddev version from artifacts to test). But again, I find the results in https://github.com/drud/ddev/pull/1484#issuecomment-470284175 - If you want to use Chrome or Safari on macOS to use huge hostnames it doesn't seem to work (curl works fine).  [This article](https://stackoverflow.com/questions/15090220/maximum-length-for-url-in-chrome-browser) has links to all the necessary info on URL lengths, and implies that there should be no problem, but I sure have a problem.

rfay:
Artifacts for testing are at https://circleci.com/gh/drud/ddev/10617#artifacts/containers/0 - I would imagine that just configuring a simple project with a long hostname is good enough. So download the ddev you need, `ddev rm -a` and `ddev start` your long-name project.

rfay:
I tested this with `name: a1234567890b1234567890c1234567890d1234567890` (resulting in a 44+5+6=56 character URL) and it seemed to work everywhere. 

However, this will need to be tested and approved by the OP to get in, we need to make sure it does what you need, otherwise it will be closed. (That means you @janhelke !)

* Configure a long, problematic name
* `ddev rm -a`
* `ddev start`
* Confirm adequate behavior for your use case.

janhelke:
I can confirm, that using the provided artifacts our projects (even the one with the very long names) work like a charm.

PR #1485, opened 2019-03-05
Package php5.6 and php7.0 so we can still provide them after sury stops providing, fixes #1448

### LINKED ISSUE #1448: deb.sury.org (was) ending support/distribution of PHP 5.6/7.0 in April 2019
**Describe the bug**

Please see https://www.patreon.com/posts/ubuntu-14-04-php-22093917 - 

Sury is ending 5.6 and 7.0, so we either have to drop them or find a way to provide them without him (like downloading and stashing the .deb's?) 

Of course they have no support from anybody already as both are EOL.

This means that to support those (which we should) we need to figure out a way to stash the .deb's or learn how to build, or something.

rfay:
## The Problem/Issue/Bug:

OP #1448: deb.sury.org will no longer be providing php5.6 and php7.0, but our users will still want those. 

* Put the php5.6 and php7.0 .deb packages on bintray as tarballs
* Download them during the docker build and install as files.

All this really requires is checking to make sure that php5.6 and php7.0 can still be used.

It would be good to check on xdebug on all versions

Existing tests should catch regressions.

## Related Issue Link(s):

OP #1448

## Release/Deployment notes:

This requires having the old package tarballs in a safe location so we can repeat this build in the future. I put it at bintray (https://dl.bintray.com/drud/oldphptarballs) but I'm not totally sold that that's the best place. Suggestions are welcome.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -83,7 +83,7 @@ RUN apt-get -qq update && \
         php-imagick \
         php-uploadprogress && \
     for v in $PHP_VERSIONS; do apt-get -qq install --no-install-recommends --no-install-suggests -y $v-apcu $v-bcmath $v-bz2 $v-curl $v-cgi $v-cli $v-common $v-fpm $v-gd $v-intl $v-json $v-mysql $v-mbstring $v-memcached $v-opcache $v-redis $v-soap $v-sqlite3 $v-readline $v-xdebug $v-xml $v-xmlrpc $v-zip libapache2-mod-$v ; done && \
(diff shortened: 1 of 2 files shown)


unn:
Maybe this is a documentation task, but it would be nice if we could inform the user that these versions of PHP are at end of life and are likely not compatible with many hosting providers.

```
# The php_versions below contain several options for your development requirements, however, PHP 5.6 and 7.0 are at end of life (eol) and are deprecated. You should not begin a new project or attempt to maintain an old project on these versions. You will not be able to host your project on these versions of PHP.

php_version: "5.6-eol"  # PHP version to use, "5.6-eol", "7.0-eol", "7.1", "7.2", "7.3"
```

rfay:
I've tested this on all PHP versions and it seems to work fine. The one question that I have is whether we should be installing those unversioned php items (xdebug, apcu) at all, or if we're installing them, should they be installed before the regular PHP installs. Either way, we may get a situation in the future where php-xdebug doesn't work with old PHP.

rfay:
I tested this without the unversioned packages and it seems to work fine. Pushed that in.

Now we'll close this PR until further need because as noted in https://github.com/drud/ddev/issues/1448#issuecomment-470922422, Sury has relented and is going to keep 5.6 and 7.0 in the repo for now.

Issue #642, opened 2018-02-07
Manage ddev project list independently of docker container labels (yaml)


rfay:
## What happened (or feature request):

Over in https://github.com/drud/ddev/issues/484#issuecomment-363656446 the discussion is underway about ddev start/stop/rm, and one of the key reasons we've kept containers around (the "stopped" state) is just because ddev doesn't know what projects it has unless docker tells it. 

In reality we need to track projects separately from docker, because it's impossible for docker containers to represent the actual state of what a user may have set up. 

We've talked at times about using a ddev-managed yaml file in ~/.ddev to track project state; It would probably be more appropriate to use a real database like sqlite instead. Basically every start would put the key information into the db, and `ddev list` would report the info from that database (whether sqlite or yaml or whatever). Additional docker container information would be used to supplement the information about project state.

This allows us to:

* Fully catalog the projects that a user has created (and show their state)
* Fully prevent name collisions on projects
* Fully prevent databases accidentally being shared by projects.

Essentially, way back in the day (drud local days) docker containers were the easiest way to get info about projects, and we used it. But we've outgrown that, and it was inadequate always. It's time to leave it behind. 

I'd envision the following behavior:

* `ddev start` updates the database 
* `ddev stop` (our current "remove") also updates the db
* `ddev list` uses the database and supplements with live docker info to present its list. But it would know of all projects. It could *also* know the status of db (size, and whether it exists or has been removed, which can be pretty important and useful)
* Probably we'd need a new and completely separate `ddev rmdb` or something to trash the database, as database existence isn't really about start/stop/rm
* There would probably need to be a new command to remove projects from the list that the user didn't want to show any more, and probably `ddev list` would need some new filter options, for example to not show projects that have no database, or to not show projects that are inactive.

Destruction of the list db would be nonfatal (just deletion of a file) and the database could be recreated naturally just by starting projects.

This is most likely a prerequisite to #484 (reorganize stop/rm commands)

## Please use a complexity rating of 1-5 (5 is high) for a feature request. (High complexity implies more PR planning)

3 - This appears relatively easy to implement, but will likely expose weaknesses in our code that will have to be dealt with.

rfay:
Note that #524 was the previous place this was being discussed, and there was an associated [google doc](https://docs.google.com/document/d/1Xy0ag25AA9NqmH6MhLIph5abcp9i_YD5mG7F5RS14jQ/edit#heading=h.jdj6auaan21x) with the various issues

[...]

PR #1502, opened 2019-03-20
Static configuration of webserver and dbserver localhost ports, bind only to localhost, fixes #1491, fixes #941, #642 too

### LINKED ISSUE #941: Stable port number for accessing mysql and http on localhost

### LINKED ISSUE #1491: Host binding of database should be to 127.0.0.1 only, not to 0.0.0.0
**Describe the bug**
Don't map db container to 0.0.0.0

**To Reproduce**
1. Configure ddev and start the project
2. run `docker ps`
3. See that the db container maps it's port to 0.0.0.0

**Expected behavior**
Map to just 127.0.0.1 like:

```
  db:
    ports:
    - 127.0.0.1::3306
```

rfay:
## The Problem/Issue/Bug:

fixes #1491, fixes #941

OP #1491: For security reasons, host binding of the dbserver should only be localhost (127.0.0.1) not 0.0.0.0 (the default, all interfaces), because this allows other people on the network to possibly probe for the database connection, depending on firewall configuration.

OP #941: Keep a stable port number for the localhost ephemeral ports so that people can use standard techniques to access them.

#642: We've long talked about keeping a catalog of metadata even for projects that aren't running. This PR introduces that.

* add `host_db_port` and `host_webserver_port`to config.yaml, use those values in the docker-compose.yaml.
* When a new host_db_port is written, add it to ~/.ddev/global_config.yaml so it's blacklisted from being used again for another project.
* Change the generated docker-compose.yaml to use explicit binding to 127.0.0.1 instead of the implicit 0.0.0.0

This is actually an amazing step forward for us. For the first time it provides a global database of projects, as requested in #642 . It's now trivial to add additional project information in the global_config.yaml.

* Experiment with host_db_port and host_webserver port, both generated and explicitly configured.
* Verify that one project can't poach another project's ports.
* Verify that used ports show up in ~/.ddev/global_config.yaml
* Verify that db and web ports can only be accessed on the localhost interface, not via other interfaces on the machine. (You can use `ifconfig` on macOS and Linux to view other interfaces). You can test with `telnet localhost 33434` for port 33434, `telnet 172.28.99.33 33434` for a denied access to the same port on the 172.28.99.33 IP address (you must use the one that pertains to your system). Alternately, check from another machine on your network, but this requires the firewall to be turned off for correctness.
* Verify maintenance of ~/.ddev/global_config.yaml when projects are removed and rm --omit-data (which should completely remove the host)

- [x] Test for util.GetFreePort()
- [x] Test for access to used ports. (The existing router tests have some code that should help). 
- [x] Test to make sure that one project's ports can be reused deliberately or accidentally by another one.
- [x] Explicitly test to make sure localhost mysql and http port can be connected to correctly. This was only caught by accident on TestWebserverType (Docker Toolbox)

## Release/Deployment notes:

We'll probably want to tell people to do a `ddev config` on existing projects, which will add the new items in config.yaml.


rfay:
Artifacts for this round are at https://circleci.com/gh/drud/ddev/10976#artifacts/containers/0 and testing would be fantastic. @tmotyl and @t3easy

rfay:
@andrewfrench this one requires a significant effort both with the code review and manual testing and your review of the basic approach. I think it's good, but sure look forward to your opinion. 

I guess the biggest basic design question is whether the host database should be kept in global_config.yaml (as currently) or split into a different file. We'll see what you think.

andrewfrench:
Hey @rfay, I'm working through review of this now and it adds a huge amount of both immediate utility and future potential for expanding on the local database of projects idea. 

Managing port claims in global_config.ddev works for me, but I was able to spot some issues regarding the way project ports are claimed in global_config.ddev: after a project is configured and its ports are claimed under the project's name, the normal methods of editing a project cause failed starts, like editing the "name" field in .ddev/config.yaml or `ddev config`. A user can bypass this by editing global_config.yaml, but we don't have any messaging to indicate this, only that the port is claimed by the project's previous name.

I can imagine a few solutions to this:
- Creating a primary key that's not editable by the user like a UUID, for example
- Handling the project name rewrite inside `ddev config [--project-name xyz]`
- More verbose error messaging when a project fails to start

rfay:
Wow @andrewfrench I think you just saved a lot of people a lot of pain. Thanks for checking that carefully.

I think the fundamental error I've introduced is reserving the port for every single project, when only a few people have actually requested it. I'll experiment with only reserving globally when explicitly specified in project config.yaml. It was just wrong to have a second global reservation at this point (besides project name)

My initial response to the general issue is that we've always used project name as the unique systemwide identifier, but we haven't ever enforced it. As a result of this undocumented assumption, we have occasional failures like

* Stolen databases (database is keyed by name) so if you start a new project with the same name it adopts (and may damage) another project's db.
* Conflicting containers: Most of us have seen this occasionally, and we even have tests for it. And it's annoying. 

I suspect we can solve both of those pretty easily as this functionality gets fleshed out. In the next sprint we can add the project directory and it will solve an enormous amount of this kind of trouble, and also allow `ddev start` of a rm'd project.

rfay:
@andrewfrench I changed the logic so that the port is dynamic *unless* specified in the config.yaml, and not reserved in the global_config.yaml unless it's set as static in config.yaml. I think this is far more in keeping with the request, and should solve the problem you described.

pkg/globalconfig/global_config.go
```diff
@@ -155,3 +166,74 @@ func GetValidOmitContainers() []string {
...
+		return port, nil
+	}
+	return "-1", fmt.Errorf("GetFreePort() failed to find a free port")
+
+}
+
+// ReservePorts() adds the ProjectInfo if necessary and assigns the reserved ports
+func ReservePorts(projectName string, ports []string) error {
+	// If the project doesn't exist, add it.
+	_, ok := DdevGlobalConfig.ProjectList[projectName]
```
andrewfrench: After playing around with this post-update, I think there should be a check either here or perhaps in the caller that returns early if `len(ports) == 0`. As written, an empty `ProjectInfo` struct and `usedhostports` list are written to global_config.yaml, even if no ports are reserved -- not the end of the world, but adds some unneeded clutter to that config file.

andrewfrench:
I noticed that `host_webserver_port` and `host_db_port` aren't configurable through command-line flags, would you be open to adding those options (`ddev config --host-db-port 1515 --host-webserver-port 5151`)?

pkg/globalconfig/global_config.go
rfay: This sounded like a great idea, and I started working on it... but since I was a good boy and used a struct instead of a map, I can't do it without removing the entire ProjectInfo, which would mess up future uses of the list.

rfay:
I added the config flags and some test coverage. Couldn't get rid of the empty used_host_ports in global_config unless you have a suggestion about how to do that @andrewfrench . Thanks for the careful review. Saved an enormous amount of future pain.

pkg/globalconfig/global_config.go
```diff
@@ -29,7 +29,7 @@ func init() {
 }
 
 type ProjectInfo struct {
-	UsedHostPorts []string `yaml:"used_host_ports,flow"`
+	UsedHostPorts []string `yaml:"used_host_ports,omitempty,flow"`
```
andrewfrench: I confirmed that this will hide the `used_host_ports` list when none were requested, but also played around with the logic in CheckAndReserveHostPorts (ddevapp/config.go) that prevents the project info map record from being generated when no relevant port information was provided, I'll comment there as well.

pkg/ddevapp/config.go
```diff
@@ -191,6 +199,27 @@ func (app *DdevApp) WriteConfig() error {
...
+	if app.HostWebserverPort != "" {
+		portsToReserve = append(portsToReserve, app.HostWebserverPort)
+	}
+	if len(portsToReserve) > 0 {
+		err := globalconfig.CheckHostPortsAvailable(app.Name, portsToReserve)
+		if err != nil {
+			return err
+		}
+	}
+	err := globalconfig.ReservePorts(app.Name, portsToReserve)
```
andrewfrench: If this call to globalconfig.ReservePorts is moved up inside the if block right before it (`if len(portsToReserve) > 0 {`), the project info map won't be written to the global config file in the case where no ports are requested. There may be some unintended consequences there that I'm not considering, but it does seem a touch cleaner in the general case from what I can tell.

rfay: Yeah, but I want the project info map always, for future use. I'm expecting it to be there for every project by the next sprint, containing a minimum of the approot. Given that I want it in the projectlist, are you good with it?

andrewfrench: Yep, in that case it's totally fine after your most recent commit.

PR #1504, opened 2019-03-21
Allow config.*.y*ml files to override config.yaml, fixes #1410

### LINKED ISSUE #1410: Allow overriding config.yaml so that .ddev config can be meaningfully committed to Git

rfay:
## The Problem/Issue/Bug:

OP #1410 requests config.yaml overrides. There are lots of different opinions in the thread, so a very simple approach is used here.

* .ddev/config.yaml is loaded first
* After that, override files in .ddev/config.*.y*ml are loaded, in alphabetical/glob order. 

Users can choose the workflow they want for these and can choose how to manage them in git.

The simplest possible technique could be to use a `config.local.yaml` for things that should not be shared with the team.

## TODO:

- [x] Add `ddev debug configyaml` to show the resultant config.yaml totally loaded, and also make sure it shows what files are being loaded.
- [x] Make sure that `ddev config` and anything else that writes config.yaml *only* writes stuff that was in the config.yaml, not in the add-ons. Probably need a test for this.
- [x] Add docs on how this works

* Add a `.ddev/config.local.yaml` with custom configuration. `ddev start` and inspect the results.
* Try this with multiple files.
* Test multiple variants of `ddev config` (which writes out config.yaml) to make sure that config.*.yaml values cannot sneak into the written config.yaml.
* Verify the correctness of the output of `ddev debug configyaml`
* Verify the correctness of new docs

- [x] Test this basic behavior
- [x] Test that config.*.yaml are properly ordered by the glob on all test environments.

## Related Issue Link(s):

OP #1410


rfay:
Artifacts for this version for testing are at https://circleci.com/gh/drud/ddev/11014#artifacts/containers/0

rfay:
As I think about this there's probably an untested flaw; `ddev config` will probably write out the fully-concatenated version of the config...

Lefaux:
I strongly recommend to not implement this change as is.
glob() orders vary between OSs which can lead to unexpected results which are hard, if not impossible to debug.

Manual sorting could help here, so an alphabetical sorting in go could solve that problem.

pkg/ddevapp/config.go
```diff
@@ -195,25 +195,27 @@ func (app *DdevApp) WriteConfig() error {
...
-		return fmt.Errorf("invalid configuration in %s: %v", app.ConfigPath, err)
+		return fmt.Errorf("unable to load config file %s: %v", app.ConfigPath, err)
 	}
-
 	app.DBImage = "" // DBImage will be set below
 
-	// ReadConfig config values from file.
-	err = yaml.Unmarshal(source, app)
+	// Load config.*.y*ml after in glob order
+	glob := filepath.Join(filepath.Dir(app.ConfigPath), "config.*.y*ml")
```
unn: A `config.*.y?ml` makes more sense to me.

rfay:
Thanks @Lefaux - This isn't ready for commit by any means, and I certainly agree that the results of the glob have to be consistent, and we need to test for this. I do note that glob is not an operating system function, but (normally) a shell function, and that here we're using golang's globbing, not the shell's. Sorting is easy if it's not irrelevant, but it seems to me that [the golang glob code](https://golang.org/src/path/filepath/match.go?s=5600:5655#L305) is committed to lexicographical order.

pkg/ddevapp/config.go
rfay: Thanks for taking a look @unn - but y?ml doesn't match config.*.yml, which is the intent here. It seems [glob has no concept of 0 or 1 characters](https://en.wikipedia.org/wiki/Glob_(programming)) like some versions of regex do. Of course we could use regex here.

unn: I tried! On the upside we can say we support `yaaaaaaaml` files.

rfay: And config.yWHATDOYOUMEANml!

rfay: Oh, I guess we do config.what.in.the.hell.is.in.this.yANIml!

rfay:
Latest artifacts for testing are at https://circleci.com/gh/drud/ddev/11174#artifacts/containers/0 - Thanks to all who are willing to take it for a spin!

alesrebec:
Just for reference... because we used this structure `config.*.yaml` we can experiment more with file naming.
Just tried using ddev v.1.7.1 with config override named `config.os.linux.local.yaml` and it works as it should.
I only included those two lines and they were both used as overrides from original config.yaml.

```
APIVersion: v1.7.1
xdebug_enabled: true
```

rfay:
Thanks for the feedback! Glad it's going well.

PR #1505, opened 2019-03-22
Explain how to connect Solr with Drupal Search API

amitaibu:
It wasn't clear to me which values I should enter there. I thought in the beginning since it's the web container that I should use ``http://solr`. Also took me some time to figure I should add the core name in the solr path.

Hopefully, this will make it easier for others :)

### DIFF
--- a/docs/users/extend/additional-services.md
+++ b/docs/users/extend/additional-services.md
@@ -18,6 +18,14 @@ This recipe adds an Apache Solr 5.4 container to a project. It will setup a solr
 - To access the Solr container from the web container use `http://solr:8983/solr/`
 - The Solr core will be "dev"
 
+### Drupal's Search API configuration
+
+In order to connect Search API with Solr:
+
+1. Add a new Solr server via `admin/config/search/search_api/add_server`, or edit an existing one.
+1. Under "Solr host" write `<projectname>.ddev.local`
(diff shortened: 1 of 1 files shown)


docs/users/extend/additional-services.md
```diff
@@ -18,6 +18,14 @@ This recipe adds an Apache Solr 5.4 container to a project. It will setup a solr
 - To access the Solr container from the web container use `http://solr:8983/solr/`
 - The Solr core will be "dev"
 
+### Drupal's Search API configuration
```
rfay: This seems to add very Drupal-centric stuff to the docs, which aren't a Drupal tutorial at all. Would this be better as a full Drupal SOLR tutorial on Stack Overflow or a blog? I note also @mglaman 's tutorial on multiple cores for Drupal, https://glamanate.com/blog/using-multiple-solr-cores-ddev?utm_source=Drupal%20Planet&utm_medium=feed&utm_campaign=drupal

amitaibu: Sure, that can be changed and moved to SO. I can create that.

amitaibu:
Closing in favor of SO Q & A - https://stackoverflow.com/questions/55322080/how-to-configure-search-api-with-ddev/55322081#55322081

PR #1506, opened 2019-03-22
The old bind-mounted db code is now fully obsolete, remove it

rfay:
## The Problem/Issue/Bug:

We used to bind-mount the database to the host. It was a long time ago. It's time to remove that code and the related tests and scripts.

Remove the code.


rfay:
@andrewfrench this is pretty easy, I just wanted to get your concurrence on it. We haven't even supported bind-mount since you got here have we?

Issue #1374, opened 2019-01-11
ddev no longer respects dbimage: <imagespec> in config.yaml


rfay:
**Describe the bug**

If you put an alternate dbimage: in config.yaml, it is no longer respected; the version wired into ddev is used instead.

**To Reproduce**

Put 
```
dbimage: drud/ddev-dbserver:v1.4.0-10.2
```
into config.yaml. It will be ignored when you `ddev start`

**Expected behavior**

It should kind of be respected.

**Version and configuration information (please complete the following information):**

```
bgsync        	drud/ddev-bgsync:v1.5.0
cli           	v1.5.1-19-g44175b7f
commit        	v1.5.1-19-g44175b7f
db            	drud/ddev-dbserver:20190103_import-db-progress-10.2
dba           	drud/phpmyadmin:v1.5.0
ddev-ssh-agent	drud/ddev-ssh-agent:v1.5.0
docker        	18.09.0
docker-compose	1.23.2
domain        	ddev.local
router        	drud/ddev-router:v1.5.0
web           	drud/ddev-webserver:20190107_web_container_maint
```
**Additional context**
Add any other context about the problem here. Thanks!

longwave:
Hah, I thought this was just me doing something wrong when trying to locally develop #1349, never realised it was an actual real bug!

PR #1520, opened 2019-04-02
Fix regression with mariadb-version arg, fixes #1517

### LINKED ISSUE #1517: ddev config --mariadb-version=10.1 injects incorrect dbimage into config.yaml
**Describe the bug**

In candidate v1.7.0, `ddev config --mariadb-version=10.1` will create a config.yaml with 

`dbimage: drud/ddev-dbserver:v1.7.0-10.2`

Which is wrong in so many ways. It shouldn't be emitted in the first place.

This doesn't happen if you just don't use the flag; for example, if they .ddev/config.yaml already has `mariadb_version: 10.1` then a `ddev config` won't break anything.

rfay:
## The Problem/Issue/Bug:

#1517: Sadly, `ddev config --mariadb-version 10.1` creates a config.yaml with the `dbimage...10.2`... and it shouldn't even have a dbimage at all. 

In the midst of config spaghetti, this makes sure the default dbimage in config handling gets set right.

Oh, this has to be tested against every regression in its sad history.

But 
* `ddev config --mariadb-version 10.1` should work with the .ddev/config.yaml showing the right stuff (and not having a dbimage spec)
* `ddev config --mariadb-version 10.2` should do the same
* Manually changing to 10.1 or 10.2 in the config.yaml should work
* Manually setting the dbimage for 10.1 or 10.2 should work

## Release/Deployment notes:

I'm not sure if it's worth getting this into v1.7.0 or just telling people not to use `ddev config --mariadb-version` for this release. I'm not entirely sure that anybody uses it.

It's clear that interacting specifications give us enormous pain and we'll have to figure out a coherent approach to them.

### DIFF
--- a/cmd/ddev/cmd/config-global.go
+++ b/cmd/ddev/cmd/config-global.go
@@ -20,8 +20,7 @@ var configGlobalCommand *cobra.Command = &cobra.Command{
 	Use:     "global [flags]",
 	Short:   "Change global configuration",
 	Example: "ddev config global --instrumentation-opt-in=false\nddev config global --omit-containers=dba,ddev-ssh-agent",
-	//Args:    cobra.NoArgs,
-	Run: handleGlobalConfig,
+	Run:     handleGlobalConfig,
 }
 
 // handleGlobalConfig handles all the flag processing for global config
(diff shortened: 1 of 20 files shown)


rfay:
Wow, so sad. I'm pretty sure that this PR addresses #1517 correctly, but in appropriately testing it you've discovered yet another round of #1374, which is truly annoying. It's obvious that we need better tests and some work on design.

I think what I'll do is add more testing to this one to address the specific problems with the `ddev config` action and confusion. Then I'll probably reopen #1374 or open another issue for it to chase the NewApp() problems.

rfay:
Thanks for preventing a descent into chaos @andrewfrench , would love to have you look again. So many permutations. The only good thing is that correctness of `ddev config` behavior can  be determined just by looking at the resulting config.yaml, doesn't require a start. Correctness of the NewApp() behavior can be determined on start and depends only on the config.yaml.

We need to avoid interdependent config at all costs. I wonder how we're going to do that. I wonder if we ever needed mariadb_version.

PR #1526, opened 2019-04-09
Reorganize stop and remove commands (use stop), fixes #484

### LINKED ISSUE #484: Reorganize start/stop/rm commands

Our start/stop/remove commands have kind of morphed and they may be confusing to the end user.

The way things are set up at this point (with the db mounted), there's never a reason to stop and keep the docker containers around, since new ones are identical. And the different modes of "remove" could be confusing.

I propose:

* Remove the current "stop" command implementation
* Change the basic "remove" command to "stop"
* Remove the "restart" command, as "start" does all the work it would do.
* Change the "remove --remove-data" command to "remove" and add a confirmation step.

I think these semantics would be more likely for users to relate to clearly, and it might give the GUI a bit more of a handhold as  well.

rfay:
## The Problem/Issue/Bug:

OP #484 : Make the "stop" command actually do a "remove". The original "stop" was analogous to docker's stop and caused no end of confusion. People will understand "start" and "stop" better

* Rename "stop" to "stop-containers"
* Rename "remove" to "stop"
* Alias "remove" and "rm" to "stop"  so `ddev rm` will still work the same.
* Rename app.Down() to app.Stop() (affects only internal usage)
* Change behavior of "restart" to "stop and start" (as it was... but stop is a different thing now)

* Use `ddev stop` and view behavior.
* Experiment with `ddev stop-containers`
* Experiment with `ddev restart`

* The existing stop, stop-containers, and remove tests remain the same, just renamed.

## Related Issue Link(s):

OP #484 

## Release/Deployment notes:

We'll want to be specific about explaining this in release notes.


rfay:
@andrewfrench Open to any thoughts about this. It was surprisingly easy. Want to make it easy for people to understand and use... and transition to.

andrewfrench:
I've tested this manually, behaves as I'd expect and no comments regarding the changes to the source, :+1: from me.  The only potential problem I can see with these changes are the naming of the commands, which make a lot of sense to us as developers familiar with Docker and the container lifecycle, but might be more confusing to a Docker-novice ddev user:

- `ddev start` is obvious, start the project (and its containers)
- `ddev remove` is `start`'s opposite, but will _not_ remove project data (but _will_ remove containers)
- `ddev stop` was previously stopping the containers, not performing the 'opposite of `start`', but not `ddev stop-containers` is more explicit about its effect

I guess my point is if deciding on naming the command `ddev remove` over `ddev stop` will prove to be more or less confusing to a new ddev user.

rfay:
@andrewfrench I reorganized this so that "start" and "stop" are the lead commands. The new "stop" does what "remove" used to do. "rm" and "remove" are aliases of stop, do what they always did.  Thanks for your thoughtful approach to this, and please schedule time for another look at it.

Issue #1495, opened 2019-03-13
Deprecate experimental webcache feature


rfay:
**Describe the bug**

Webcache (bgcache) has so very many problems in a cross-platform or a stressing environment; we'll likely never be able to get predictable behavior out of it on Windows. 

We should consider deprecating this experimental feature, perhaps removing it in v1.8.0 or somewhere there.

rfay:
Responses on slack have been consistent that NFS is better and we should let webcache go. I think we should announce immediately:

* No further support for it
* Removal of the feature June 1, 2019 or so.

In addition, we should remove the docs immediately.

pcambra:
I want to point out that *based on my benchmarks*, in my experience, running tests from within the container using webcache is ~10x faster than NFS.

rfay:
@pcambra we're leaving it in there for now, you were the only one who spoke up in its favor. We're just not running tests against it any more, and docs indicate deprecated.

PR #1528, opened 2019-04-10
[TASK] add hint of deprecated webcache

kaystrobach:

### DIFF
--- a/docs/users/performance.md
+++ b/docs/users/performance.md
@@ -66,6 +66,8 @@ Tools to debug and solve permission problems:
 
 ## Using webcache_enabled to Cache the Project Directory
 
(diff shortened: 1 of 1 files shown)


rfay:
Thanks, #1495 is on my list for this round, I think there's more to do than just this though. Docs go lower,  I'm going to close this as a sort-of-duplicate but thanks very much.  Interestingly enough, there was one real, real fan of webcache that objected to the deprecation plans. He uses it for Drupal unit tests, and of course it's blazingly fast and consistency doesn't matter. It's a valid use case. May add a note that it's mostly appropriate for that case.

rfay:
@kaystrobach would appreciate your review of https://github.com/drud/ddev/pull/1530

PR #1534, opened 2019-04-12
ddev-webserver Nginx - do not intercept 400 errors

AronNovak:
## The Problem/Issue/Bug:

The issue is that currently Nginx swallows all the upstream error messages. This is particularly painful for instance for webservices where the error message would be crucial to get in addition to the status code.

It changes the default configuration of Nginx to pass all the upstream errors.

Put a tiny PHP script inside the container:
```php
header($_SERVER['SERVER_PROTOCOL'] . ' 500 Internal Server Error', true, 500);
echo "moo";
```

`moo` should be visible when accessing the script.
Right now the PR does not contain tests, if it's approved that it's a wanted change, it should have a test that contains a dummy script that provides error and it tries to access that URL via `wget`˛and see if the body of the request contains the output from PHP.

## Release/Deployment notes:
DDEV users need to be warned that the default behavior changes, it alters the behaviors of the applications in case of HTTP error responses.


rfay:
This seems reasonable to me.

rfay:
Should we also be removing these? 

https://github.com/drud/ddev/blob/4f6b8ff01c3096be3148a16d8aa80468da1fc51e/containers/ddev-webserver/files/etc/nginx/nginx-site-drupal7.conf#L119-L122

AronNovak:
@rfay Thanks, then I check now where and how to add the test.

rfay:
I took the liberty of pushing an updated ddev-webserver and the version to point to it.

AronNovak:
> Should we also be removing these?
> 
> [ddev/containers/ddev-webserver/files/etc/nginx/nginx-site-drupal7.conf](https://github.com/drud/ddev/blob/4f6b8ff01c3096be3148a16d8aa80468da1fc51e/containers/ddev-webserver/files/etc/nginx/nginx-site-drupal7.conf#L119-L122)
> 
> Lines 119 to 122 in [4f6b8ff](/drud/ddev/commit/4f6b8ff01c3096be3148a16d8aa80468da1fc51e)
>  error_page 400 401 /40x.html; 
>  location = /40x.html { 
>          root   /usr/share/nginx/html; 
>  }

Good question. I'd say we might get 4XX code from PHP-FPM, then it could make sense to specify these.

rfay:
The actual error pages are also there. 

Unfortunately, I know that those were added to make things easier for some people (quite a long time ago) and can't remember the exact reason.

AronNovak:
@rfay Is there an easy way how I could test the test I added locally, without waiting for the CI? Thanks!

rfay:
Is it container test or golang test? If it's a container test (in cmd/ddev-webserver` just run `make test` in the ddev-webserver directory.

AronNovak:
@rfay container test, thanks, I will do so.

AronNovak:
@rfay Ready for review, the test passed.

rfay:
Thanks so much! Probably should remove containers/ddev-webserver/files/usr/share/nginx/html/40x.html also, right?

AronNovak:
@rfay Just pushed the removal of that file, for sure.

AronNovak:
@rfay Done, but CI fails, it seems I am not authorized to access the build details. Do we have an error message of the build failure?

rfay:
The failure was just an internet outage on one of the testbots, retry started. Thanks so much!

rfay:
https://github.com/drud/ddev/pull/1540 does a bunch of reorganization of the nginx files, so this will have to go in after that. Shouldn't be hard to do.

AronNovak:
@rfay Okay, sure, I will revisit this after https://github.com/drud/ddev/pull/1540 is in.

PR #1538, opened 2019-04-16
Use docker volume in sample apachesolr config

wizonesolutions:
## The Problem/Issue/Bug:
The suggested DDEV Solr configuration at https://ddev.readthedocs.io/en/stable/users/extend/additional-services/#apache-solr does not include a persistent Docker volume.

It adds one.

1. Copy the Compose file, then run `ddev start`. DDEV will tell you that it created a volume.
1. Run `docker volume ls` and ensure that the volume is there. It will be named `ddev-PROJECTNAME-data`.

This is something the user would add. It won't break DDEV even if it's broken.

## Related Issue Link(s):
- https://github.com/drud/ddev/pull/1526

### DIFF
--- a/pkg/servicetest/testdata/services/docker-compose.solr.yaml
+++ b/pkg/servicetest/testdata/services/docker-compose.solr.yaml
@@ -25,7 +25,7 @@ services:
       - HTTP_EXPOSE=8983 # This defines the port the service should be accessible from at sitename.ddev.local
     volumes:
       - "./solr:/solr-conf" # This exposes a mount to the host system `.ddev/solr-conf` directory.
-      - data:/opt/solr/server/solr/mycores # This uses the 'data' volume defined below
+      - solrdata:/opt/solr/server/solr/mycores # This uses the 'data' volume defined below
     entrypoint:
       - docker-entrypoint.sh
       - solr-precreate
@@ -36,4 +36,4 @@ services:
     links:
       - solr:$DDEV_HOSTNAME
 volumes:
-  data: # This creates a Docker volume that sticks around even if you remove or rebuild the container
+  solrdata: # This creates a Docker volume that sticks around even if you remove or rebuild the container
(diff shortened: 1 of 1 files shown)


rfay:
How about a more descriptive name than "data"? THANKS!

wizonesolutions:
Ironically, that's what the image documentation itself actually uses. I guess it is confusing in the context of DDEV, though. I'll change it to `solrdata`.

rfay:
@wizonesolutions could you please provide more extensive instructions on how to test this with a real Drupal project, make sure that it works properly?

rfay:
Now actually looking at it... it does use a bind-mount already. I'd think a docker volume would be better, but the bind mount does provide persistence right?

mglaman:
Testing this.

Basically, the problem is this:

```
ddev start
ddev exec ../vendor/bin/drush sapi-i # ... indexing
ddev exec ../vendor/bin/drush sapi-s # ... verifying indexed
# ... open Solr admin via `docker ps` and find port, see documents
ddev rm
ddev start
# ... open Solr admin via `docker ps` and find port, see no documents
ddev exec ../vendor/bin/drush sapi-s # ... should show nothing indexed
```

mglaman:
Results!

I did a `ddev rm` and then updated my Solr definition. Then:

```
ddev start
ddev exec ../bin/drush sapi-c store_locator # clear values
ddev exec ../bin/drush sapi-i store_locator                                                                                                                        mglaman@Matts-iMac
 [ok] Successfully indexed 50 items.
 [ok] Successfully indexed 100 items.
 [ok] Successfully indexed 105 items.
 [notice] Message: Successfully indexed 105 items.

105	0
```

Checked the Solr index (11 stores get skipped, so index has 94.)

```
{
  "response":{"numFound":94,"start":0,"docs":[
      {
        "id":"oth2o0-store_locator-entity:commerce_store/100:en",
        "index_id":"store_locator",
```

I then did `ddev rm`

I then checked Solr. Data persisted 🙌.

Issue #1235, opened 2018-10-28
Access to webserver by name within container doesn't work if additional_hostnames invoked


rfay:
**Describe the bug**

In https://github.com/drud/ddev/pull/1151 we made it so you could access the webserver by full URL from within the web container. For example, `curl https://d8git.ddev.local`. The way it's set up, d8git.ddev.local resolves to the IP address of the *ddev-router*. 

However, once you add anything to additional_hostnames, all the IPs resolve to 127.0.0.1. As a result, http to the hostname *may* work if there's no alternate port, but https or most everything is broken.

**To Reproduce**

* ddev ssh
* ping main hostname of project. It should be the same as the ddev-router address (maybe 172.18.0.5 or something) and *not* 127.0.0.1
* exit
* Add to additional_hostnames in .ddev/config.yaml
* ddev start
* ddev ssh
* ping main hostname or any other additional_hostnames; they will resolve to 127.0.0.1, not going through the router. You can demonstrate the problem by hitting any of the supported URLs via https, for example for the main project being d8git, hit https://d8git.ddev.local and it will fail with connection refused (because it's hitting port 443 on the web container instead of going to the router)

**Expected behavior**

**Screenshots**

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: [e.g. Windows 10, macOS High Sierra]
 - Docker version information (use `docker version`) and insert it here.
 - ddev version information (use `ddev version`)
 - config.yaml contents for the misbehaving project
 - Do you have any custom configuration (nginx, php, mysql) in the .ddev folder? If so, have you tried without them?

**Additional context**
Add any other context about the problem here. Thanks!

rfay:
I think this workaround (though fragile) might work until we can sort this out:

This depends on the router normally getting the IP address 172.18.0.6, if your router has a different address, change it. You can find out its address with `ping ddev-router` inside the web container.

* Add a .ddev/docker-compose.hosts.yaml (particular name doesn't matter)

```
version: '3.6'
services:
  web:
    extra_hosts:
      - extraname1.ddev.local:172.18.0.6
      - extraname2.ddev.local:172.18.0.6
    environment:
      - VIRTUAL_HOST=originalname.ddev.local,extraname1.ddev.local,extraname2.ddev.local
```

d3pendent:
Is there a variable that contains the IP addess of the docker0 network gateway?
Since developers could have different system setups, it's not perfect to commit this file into any projects repository.

[...]

PR #1540, opened 2019-04-17
Trust SSL/TLS Certificates, fixes #1247, fixes #1501, fixes #849

### LINKED ISSUE #849: Reorganize nginx configs to share most elements
_From @rfay on March 22, 2018 21:2_

Currently the entire nginx config is repeated for every project type, which is unnecessary and confusing and hard to maintain. They should be reorganized so that only the key things that are *different* are in the separate configs.

_Copied from original issue: drud/docker.nginx-php-fpm-local#63_

### LINKED ISSUE #1247: Add self-signed SSL certs to CA chain in web container
In some circumstances TYPO3 9.5 uses curl to fetch own content, f.e. for rendering error pages.
Trying to fetch these pages via https results in an error due to invalid SSL certificates, because they are self-signed but not known by the CA inside the web container.
As a workaround we added a volume pointing to ~/.ddev/certs and added a post-start hook to add all certs found in this directory to the CA:

.ddev/config.yaml

```
hooks:
    post-start:
      # add self-signed certs to ca to be used by curl, wget etc.
      - exec: bash -c "sudo /usr/local/bin/add-certs-to-ca"
```

.ddev/docker-compose.override.yaml

```version: '3.6'

# use this file to make the hostnames available for dns in web container
# only works with http because https is terminated by ddev-router (nginx proxy)
# and self-signed ssl certificate is invalid for curl
services:
  web:
    external_links:
      - ddev-router:$DDEV_HOSTNAME
    volumes:
      # mount certs into web container to add them to CA for using curl with https
      # without complaining about insecure self-signed certificates
      - ~/.ddev/certs:/usr/share/ca-certificates/ddev
      - ../Build/Ddev/add-certs-to-ca.sh:/usr/local/bin/add-certs-to-ca
```
Build/Ddev/add-certs-to-ca.sh

```
#!/bin/bash

search_dir=/usr/share/ca-certificates/ddev
ca_certs_file=/etc/ca-certificates.conf

# remove already existing certs of ddev from certs file
sed -i '/^ddev\//d' "$ca_certs_file"

# add all ddev certs
for entry in "$search_dir"/*.crt
do
  b=ddev/$(basename "$entry")
  echo "$b" >> "$ca_certs_file"
done

update-ca-certificates --fresh
```

### LINKED ISSUE #1450: Accept https callbacks with valid (not self-signed) certificate
It's not unusual for a project to interact with other systems, and good practice requires data to be sent over https. Ddev uses its own certificate in when responding to https callbacks from external systems, and this certificate is often not acceptable to services sending data, which makes it difficult to use ddev to develop when external systems are involved.

_Notes_

- ddev uses the same certificate for all sites. In fact the router software jwilder/nginx-proxy has the ability to use the correct certificate. I think the nginx template file would need alteration to more closely match the jwilder version. https://github.com/jwilder/nginx-proxy/blob/master/README.md#ssl-support
-  jwilder/nginx-proxy has an associated container which works with LetsEncrypt. This requires port 80, so it won't work with projects that use other ports. https://github.com/jwilder/nginx-proxy/blob/master/README.md#ssl-support-using-letsencrypt 

_Workarounds_
- As a hacked workaround, you can set up a non-ddev Nginx reverse proxy outside of ddev with the correct certificate. This terminates the https request, and passes the request into ddev. It's clutzy as a solution because the whole project is no longer in ddev. If you want to use LetsEncrypt, you'll need to set this up on port 80. You will need to switch your ddev containers off port 80. This workaround requires port 80 to be reachable from the internet, and of course having a domain correctly set up.
- An alternative workaround is to tunnel from a computer on the internet, through to your dev machine. See rfay's article here - https://randyfay.com/content/tunneling-http-debug-or-develop-external-webservice-call . This will not work if the external service does not accept the ddev self-signed certificate.

### LINKED ISSUE #1501: Allow extending nginx config
Would be cool if nginx conf could be extended instead of just replacing it. The possible solution could be the same as in lagoon platform -> https://github.com/amazeeio/lagoon/blob/master/images/nginx-drupal/drupal.conf

The key is to add the similar line in the server directive:
`include /etc/nginx/conf.d/ddev/server_prepend*.conf;`

So a developer would just create a partial configs which would be copied during docker build.

Use case. For example we have drupal private folder inside the public folder under `sites/default/files/private`. We obviously want to protect the private files. Ideally we would just create an nginx conf file like server_prepend_protect_private_files.conf with the content:
```
    location /sites/default/files/private/ {
        internal;
    }
```

And that's it. The default nginx config is extend with this part and our private files are protected. With this users can have much more control over their nginx configs instead of fully replacing them.

rfay:
## The Problem/Issue/Bug:

https is now standard everywhere, but ddev has never had a way to present a trusted certificate.
mkcert makes it possible.

* Use mkcert to create a trusted CA on the host
* Copy the trusted CA files into the router and the webserver container
* Generate certs for each name needed (or wildcard)
* Provide a cert for 127.0.0.1 for web container
* Make sure web container trusts the certs as well
* Reorganize nginx config so that there's not so much repetition
* Add simple include nginx config add-ons per #1501 - simple item while doing the nginx config reorg

* Install mkcert (`brew install mkcert`, `choco install -y mkcert`)
* `mkcert -install`
* Use browser and curl against https URLs on host
* Test complex URLs introduced with additional_hostnames and additional_fqdns. 
* Try just regular old https
* Try curl within the web container against localhost and against the FQDN with https
* Try curl on the host against localhost-https port as shown in `ddev describe`
* Test behavior without mkcert installed. It should "work", but not have trusted browser responses.
* Use nginx snippet overrides in `.ddev/nginx/<something>.conf` and verify that they get detected.
* Review updated docs for mkcert and nginx config and other (docs/index.md, docs/users/extend/customization-extendibility.md
* Test `ddev config --host-https-port`

- [x] We may have places where we're currently ignoring https failure and we can respect it now!
- [x]  Start a project with no router running
- [x]  Start up with another project running, both should be valid
- [x]  Container tests for ddev-router, cert should work inside (`curl https://d8composer.ddev.local --resolve "d8composer.ddev.local:443:127.0.0.1"`)
- [x]  Container tests for ddev-webserver, cert should work inside against 127.0.0.1 address. #1235 will prevent access to router via hostname

## Related Issue Link(s):

#1247 

## Release/Deployment notes:

* Add mkcert dependency to homebrew
* Explain in release how nginx snippets work
* Explain mkcert in release
* Explain `ddev config --host-https-port`
* Explain that https is now the first option in lists of URLs and why


rfay:
Thanks so much @andrewfrench - I'm in the process of adding nss.

Question though on your install of nss... This is what I see on `brew install mkcert` on macOS, I don't get a prompt about nss:

```
$ brew uninstall --ignore-dependencies mkcert nss
Uninstalling /usr/local/Cellar/mkcert/1.3.0... (6 files, 4.8MB)
Uninstalling /usr/local/Cellar/nss/3.43... (222 files, 27.5MB)
rfay-macos-mbp2013:d8git rfay$ brew install mkcert
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 3 taps (rfay/ddev, homebrew/core and homebrew/cask).
==> Updated Formulae
asciidoctor         gitlab-runner       newsboat            ruby-build
erlang              glooctl             purescript

==> Downloading https://homebrew.bintray.com/bottles/mkcert-1.3.0.mojave.bottle.
Already downloaded: /Users/rfay/Library/Caches/Homebrew/downloads/8811bf4b440be280e5f5685d7f54a55e4620638b3676953ffc30b12f455ead7a--mkcert-1.3.0.mojave.bottle.tar.gz
==> Pouring mkcert-1.3.0.mojave.bottle.tar.gz
🍺  /usr/local/Cellar/mkcert/1.3.0: 6 files, 4.8MB
```

rfay:
Forget it @andrewfrench , it's very clear on the mkcert project page,

> brew install mkcert
> brew install nss # if you use Firefox

I'll update docs and such for that.

andrewfrench:
Just ran through again and noticed that the message actually shows up in the `mkcert -install` step, not the `brew` step:

```
Using the local CA at "/Users/andrew/Library/Application Support/mkcert" ✨
Warning: "certutil" is not available, so the CA can't be automatically installed in Firefox! ⚠️
Install "certutil" with "brew install nss" and re-run "mkcert -install" 👈
```

But I see that both mkcert and nss have been added as dependencies, thanks for that!

PR #1549, opened 2019-04-27
Improved Solr install docs, update to 6.6, fixes #1547

### LINKED ISSUE #1547: Update Solr config to v6 or newer
**The problem**
The available [configuration file for Solr](https://github.com/drud/ddev/blob/master/pkg/servicetest/testdata/services/docker-compose.solr.yaml) mentions 5.4. This is currently no longer supported, the [Search API Solr](https://www.drupal.org/project/search_api_solr) module for Drupal 8 now only supports 6.4 or newer.

**Requested solution**
Update the Solr configuration so that it defaults to 6.4 or newer.

damienmckenna:
Updated the Apache Solr installation instructions to default to v6.6.
It also makes it more clear what the expected directory structure
should be.

## The Problem/Issue/Bug:
Per issue #1547, the included config file for Solr installed v5.4, which is no longer supported.

This PR updates it the Solr install to v6.6 and improves the documentation.

Use the config file to install Solr. Expected results: it installs Apache Solr 6.6 instead of the previous 5.4.

None.

## Related Issue Link(s):
https://github.com/drud/ddev/issues/1547

## Release/Deployment notes:
Don't know (-:

### DIFF
--- a/pkg/servicetest/testdata/services/docker-compose.solr.yaml
+++ b/pkg/servicetest/testdata/services/docker-compose.solr.yaml
@@ -1,39 +1,58 @@
-# ddev apache solr recipe file
+# DDev Apache Solr recipe file.
 #
-# To use this in your own project: Copy this file to your project's .ddev folder,
-# and create the folder path .ddev/solr/conf. Then, copy the solr configuration
-# files for your project to .ddev/solr/conf. E.g., using Drupal Search API Solr, 
-# you would copy the solr-conf/5.x/ contents into .ddev/solr/conf. The configuration
-# files must be present before running `ddev start`.
+# To use this in your own project:
+# 1. Copy this file to your project's ".ddev" directory.
+# 2. Create the folder path ".ddev/solr/conf".
+# 3. Copy the Solr configuration files for the appropriate plugin/module to
+#    ".ddev/solr/conf". For example, using Drupal 7's Search API Solr module,
+#    copy the files from its "solr-conf/6.x" directory into ".ddev/solr/conf"
+#    so that a file exists with the path ".ddev/solr/conf/solrconfig.xml".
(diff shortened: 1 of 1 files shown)


rfay:
Thanks so much for this! Do you think you could consolidate adding a volume as in https://github.com/drud/ddev/pull/1538 as well?  It does seem like we shouldn't throw it away every restart.

rfay:
@mglaman could you review this one also, looks all good to me. Would love any suggestions you might have.

pkg/servicetest/testdata/services/docker-compose.solr.yaml
```diff
@@ -1,36 +1,53 @@
...
-    image: solr:5.4
+  # This is the service name used when running ddev commands accepting the
+  # --service flag.
+  solr:
+    # This is the name of the container. It is recommended to follow the same
+    # name convention used in the main docker-compose.yml file.
+    container_name: ddev-${DDEV_SITENAME}-solr
+    # Controls the version of Solr which is installed.
+    image: solr:6.6
     restart: always
```
mglaman: If we're triaging, let's change this to `no`.

```
    restart: "no"
```

Defaulting to `restart: always` was old DDEV

damienmckenna: `Defaulting to restart: always was old DDEV`
Sounds like there needs to be a new issue for just that change, so make it on all of the example config files.

rfay:
@damienmckenna I took the liberty of rebasing this after the pull of #1538 

https://github.com/drud/ddev/pull/1559 was opened to change to `restart: "no"`

rfay:
Thanks so much for this excellent contribution.

PR #1555, opened 2019-04-30
ddev-webserver Nginx - do not intercept 400 errors

AronNovak:
supersedes #1534 

## The Problem/Issue/Bug:

The issue is that currently Nginx swallows all the upstream error messages. This is particularly painful for instance for webservices where the error message would be crucial to get in addition to the status code.

It changes the default configuration of Nginx to pass all the upstream errors.

Put a tiny PHP script inside the container:
```php
header($_SERVER['SERVER_PROTOCOL'] . ' 500 Internal Server Error', true, 500);
echo "moo";
```

`moo` should be visible when accessing the script.
Right now the PR does not contain tests, if it's approved that it's a wanted change, it should have a test that contains a dummy script that provides error and it tries to access that URL via `wget`˛and see if the body of the request contains the output from PHP.

## Release/Deployment notes:
DDEV users need to be warned that the default behavior changes, it alters the behaviors of the applications in case of HTTP error responses.

Related: https://github.com/drud/ddev/issues/1367#issuecomment-485874564 neos needed this change.


AronNovak:
@rfay Here's the reworked PR, ready for a last review.

rfay:
Wow thanks. I didn't expect you to do that. Thanks. BTW, you could have just rebased the other one, would have saved some effort. But thanks!  I pushed the updated ddev-router.

AronNovak:
@rfay Yep, rebasing works, but I do not prefer force pushes and with so many conflicts, it's the same order of magnitude of work in my opinion.

AronNovak:
Hm, it seems it needs a little work due to the CI failure.

rfay:
You had used the same tag, so that image wouldn't have passed tests, etc. It's pretty confused now. I tried to restart all of them and it's *really* confused.

rfay:
Looks like you used `intercept_errors` instead of `fastcgi_intercept_errors` :)

rfay:
I took the liberty of fixing the problems and rebasing this, hope you'll take a look and make sure you're OK with it @AronNovak

AronNovak:
@rfay Thanks, the change looks good of course, now I try to understand the CI failure.

rfay:
I think the TestConfigOverrideDetection failure was a (now) obsolete nginx-site.conf in that test. Pushed a commit to update it.

rfay:
Rebased again; it seems the pushed ddev-webserver image didn't match the tests that were being run against it (php overrides).

PR #1568, opened 2019-05-06
Further improved Solr install docs.

damienmckenna:
## The Problem/Issue/Bug:
The post-setup instructions for Solr are not completely clear.

Some improvements to the Additional Services page, and the Solr config yml.

rtfm (-:

### DIFF
--- a/pkg/servicetest/testdata/services/docker-compose.solr.yaml
+++ b/pkg/servicetest/testdata/services/docker-compose.solr.yaml
@@ -9,6 +9,17 @@
 #    so that a file exists with the path ".ddev/solr/conf/solrconfig.xml".
 #
 # Note: The configuration files must be present before running `ddev start`.
+#
+# To access Solr after it is installed:
+# - The Solr admin interface will be accessible at:
+#   http://<projectname>.ddev.local:8983/solr/
+#   For example, if the project is named "myproject" the hostname will be:
+#   http://myproject.ddev.local:8983/solr/
+# - To access the Solr container from the web container use:
+#   http://solr:8983/solr/
+# - A Solr core is automatically created with the name "dev", i.e. it can be
+#   accessed at the URL: http://solr:8983/solr/dev
+
 
 version: '3.6'
 
@@ -47,12 +58,13 @@ services:
       - solr-precreate
       - dev
       - /solr-conf
-# This links the Solr service to the web service defined in the main
-# docker-compose.yml, allowing applications running in the web service to
-# access the Solr service at sitename.ddev.local:8983.
+  # This links the Solr service to the web service defined in the main
+  # docker-compose.yml, allowing applications running in the web service to
+  # access the Solr service at sitename.ddev.local:8983.
   web:
     links:
       - solr:$DDEV_HOSTNAME
 volumes:
-  # This creates a Docker volume that sticks around even if you remove or rebuild the container
+  # This creates a Docker volume that sticks around even if you remove or
+  # rebuild the container
   solrdata: 
--- a/pkg/servicetest/testdata/services/docker-compose.solr.yaml
+++ b/pkg/servicetest/testdata/services/docker-compose.solr.yaml
@@ -8,7 +8,9 @@
 #    copy the files from its "solr-conf/6.x" directory into ".ddev/solr/conf"
 #    so that a file exists with the path ".ddev/solr/conf/solrconfig.xml".
 #
-# Note: The configuration files must be present before running `ddev start`.
+# Notes:
+# - The configuration files must be present before running "ddev start".
+# - Changes to the configuration files will not be automatically applied.
 #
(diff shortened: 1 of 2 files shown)


rfay:
Yay, thanks!

pkg/servicetest/testdata/services/docker-compose.solr.yaml
```diff
@@ -9,6 +9,17 @@
 #    so that a file exists with the path ".ddev/solr/conf/solrconfig.xml".
 #
 # Note: The configuration files must be present before running `ddev start`.
```
heddn: We should add a comment in here that the config is read once for each core defined and if there's a typo or change needed, the volume needs to be deleted i.e. `docker volume rm ddev-myproject_solrdata`, where myproject is the name of site.

Another interesting note for Drupal, the default solr config in search_api_solr is setup for the tarball location of things. Change the defaults to get the core to spin up correctly. `solr.install.dir=/opt/solr` is the correct entry in `.ddev/solr/conf/solrcore.properties`

mglaman: Well, the volume doesn't need to be deleted - the core just needs to be recreated. And, actually, the config just needs to be copied over. I don't see any scripts for that here https://github.com/docker-solr/docker-solr/tree/6f7a6e812247d00f9f3c293993e26d6d041c119e/6.6/scripts

heddn: @mglaman What are you suggesting as the mitigation since the copy doesn't occur?

mglaman: From the image's docs: https://github.com/docker-solr/docker-solr-examples/tree/master/custom-configset#note

> One note of caution: we've copied the configset from a specific version of Solr. It's possible that a future version will have changes to those config files; it will be up to you to re-generate your config.

🤔

mglaman: I was hoping there was a script in the image we could re-use, I don't have time to dig in at the moment

heddn: There doesn't seem to be. I did find https://github.com/docker-solr/docker-solr-examples/tree/master/schema-api-startup, which makes me think we should be using schema api. 

For now I wonder if we should just copy/paste that folks who need to make changes to their solr config files need to modify the files inside the docker instance. They won't be overwritten after initial creation.

mglaman: Works for me. If someone is doing this, they're probably an advanced enough user.

> Note: changes to your configuration will not automatically be applied to your Solr core

damienmckenna: I added an extra note similar to mglaman's wording above.

damienmckenna: The question is, how can config changes be loaded? Would it be worth noting the steps here?

mglaman: It becomes a question of: is it something to document in DDEV or general official Solr Docker image documentation. I think a fair warning is "good enough"

pkg/servicetest/testdata/services/docker-compose.solr.yaml
```diff
@@ -8,7 +8,9 @@
 #    copy the files from its "solr-conf/6.x" directory into ".ddev/solr/conf"
 #    so that a file exists with the path ".ddev/solr/conf/solrconfig.xml".
 #
-# Note: The configuration files must be present before running `ddev start`.
+# Notes:
+# - The configuration files must be present before running "ddev start".
+# - Changes to the configuration files will not be automatically applied.
```
heddn: `...even with a ddev rm --data. They must be manually updated inside the docker instance after the initial core creation.`

pkg/servicetest/testdata/services/docker-compose.solr.yaml
wizonesolutions: Ultimately, this is an issue with the Solr Docker image, no? It's not DDEV-specific other than that the recommended DDEV config creates a volume. However, not creating a volume means potentially huge reindexing for developers.

Maybe we just need separate examples for various common scenarios so that people have a high chance of getting it right the first time.

wizonesolutions: oh, looks like I am just echoing @mglaman. My review was very slow :)

jeffsheltren:
👍 from me

rfay:
@damienmckenna it seems everybody is OK with this, I'll pull it when you're ready.

rfay:
Thanks so much to everybody for your collaboration on this!

PR #1585, opened 2019-05-11
Improvements to the memcache compose file

damienmckenna:
## The Problem/Issue/Bug:
The docker-compose.memcache.yaml file's comments needs a little tidying.

This tidies up the file to be easier to follow.

## Related Issue Link(s):
#1549

## Release/Deployment notes:
n/a

### DIFF
--- a/pkg/servicetest/testdata/services/docker-compose.memcached.yaml
+++ b/pkg/servicetest/testdata/services/docker-compose.memcached.yaml
@@ -1,20 +1,30 @@
-# ddev memcached recipe file
+# DDev memcached recipe file.
 #
-# To use this in your own project: Copy this file to your project's .ddev folder.
-# Edit the 'command' settings to change CLI flags sent to memcached.
-# Defaults to '-m 128' to allocate 128 MB of memcached storage.
+# To use this in your own project:
+# 1. Copy this file to your project's ".ddev" directory.
+# 2. Launch "ddev start".
+# 3. Configure the project to look for memcached at the hostname "memcache" and
+#    port 11211.
+# 4. Optional: adjust the 'command' line below to change CLI flags sent to
+#    memcached.
 
 version: '3.6'
 
 services:
-  memcached: # This is the service name used when running ddev commands accepting the --service flag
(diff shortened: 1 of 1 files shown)


pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -1,20 +1,30 @@
-# ddev memcached recipe file
+# DDev memcached recipe file.
```
rfay: I guess if we're going to be official it should be `DDEV-Local`

pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -1,20 +1,30 @@
...
 #
-# To use this in your own project: Copy this file to your project's .ddev folder.
-# Edit the 'command' settings to change CLI flags sent to memcached.
-# Defaults to '-m 128' to allocate 128 MB of memcached storage.
+# To use this in your own project:
+# 1. Copy this file to your project's ".ddev" directory.
+# 2. Launch "ddev start".
+# 3. Configure the project to look for memcached at the hostname "memcache" and
```
rfay: I think the hostname is "memcached" right? At least we should make it that way.

pkg/servicetest/testdata/services/docker-compose.memcached.yaml
```diff
@@ -1,20 +1,30 @@
-# ddev memcached recipe file
+# DDev memcached recipe file.
 #
-# To use this in your own project: Copy this file to your project's .ddev folder.
-# Edit the 'command' settings to change CLI flags sent to memcached.
-# Defaults to '-m 128' to allocate 128 MB of memcached storage.
```
jeffsheltren: I would like to keep this comment somewhere so it's obvious what the default memory allocation is. Could be down below with the actual arguments if that makes more sense.

rfay:
There are a couple of minor open comments on this; I'll be happy to pull it when you say it's ready @damienmckenna ; if you want me to address those I can also do it.

rfay:
I took the liberty of pushing the minor review comment changes in. Thanks so much for this!

PR #1586, opened 2019-05-11
Remove drush --yes behavior when in ddev exec behavior, fixes #1584

### LINKED ISSUE #1584: "ddev exec drush" hardcodes certain responses?
**Describe the bug**
This may be fixed in 1.8.0, I have not tested the latest codebase yet..

A common practice I have is to use "drush upc -n" to work out what updates are available, before deciding which updates to download. With ddev this is not possible as it seems it detects when drush is being executed and passes the "-y" argument to skip all prompts.

**To Reproduce**
* Run `ddev exec drush upc -n`
* Expected results: a list of available updates is presented, but none are applied.
* Actual results: all available updates are applied.

**Expected behavior**
No updates are applied when the "-n" argument is passed to `ddev exec drush upc`.

```ddev version
bgsync        	drud/ddev-bgsync:v1.7.0       
cli           	v1.7.1                        
commit        	v1.7.1                        
db            	drud/ddev-dbserver:v1.7.0-10.2
dba           	drud/phpmyadmin:v1.7.0        
ddev-ssh-agent	drud/ddev-ssh-agent:v1.7.0    
docker        	18.09.2                       
docker-compose	1.23.2                        
domain        	ddev.local                    
router        	drud/ddev-router:v1.7.0       
web           	drud/ddev-webserver:v1.7.0
```

rfay:
## The Problem/Issue/Bug:

From long ago, long, long ago, `drush` defaulted to --yes when in `ddev exec` context. It wasn't probably ever reasonable behavior, but it was there to avoid the interactive prompt mostly in exec hooks.

Remove that special behavior.

`ddev exec upc --security-only` on a Drupal site that needs update, it should stop to confirm.
`ddev exec drush sql-drop` - it should stop to confirm.

## Related Issue Link(s):

OP #1584, thanks @damienmckenna 

## Release/Deployment notes:

There may be people that have counted on this in exec hooks. So release notes have to tell them to add -y to their exec hooks.

### DIFF
--- a/containers/ddev-webserver/files/etc/drush/drushrc.php
+++ b/containers/ddev-webserver/files/etc/drush/drushrc.php
@@ -2,7 +2,3 @@
 if (!empty($_ENV['DDEV_URL'])) {
     $options['uri'] = $_ENV['DDEV_URL'];
 }
-# Skip confirmations since `ddev exec` cannot support interactive prompts
-if (!empty($_ENV['DDEV_EXEC'])) {
-    $options['yes'] = 1;
-}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -812,7 +812,7 @@ func (app *DdevApp) Exec(opts *ExecOpts) (string, string, error) {
 		return "", "", fmt.Errorf("no service provided")
 	}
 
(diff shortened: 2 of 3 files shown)


rfay:
It's possible this could be a candidate for v1.8.0, but since the issue has been there for a couple of years, it seems not that urgent.

PR #1598, opened 2019-05-16
Create settings.ddev.php on ddev start, fixes #1423

### LINKED ISSUE #1423: ddev only creates settings.ddev.php on `ddev config`
**Describe the bug**
ddev only creates `settings.ddev.php` when `ddev config` is run (https://github.com/drud/ddev/issues/692) and adds it to `.gitignore`, but when multiple developers are collaborating on a repository, the settings file is not present as expected.

**To Reproduce**
1. Clone repo.
1. Run `ddev config`
1. Commit ddev configuration to repository.
1. Push repo.
1. Have a collaborator clone the repository.
1. Collaborator runs `ddev start`.
1. Collaborator attempts to run `drush site-install`, but since `settings.ddev.php` does not exist, this fails.

**Expected behavior**
Since `settings.ddev.php` is ignored by git by default, I would expect it to be regenerated on `ddev start` if it does not exist.

An alternative could be to remove `settings.ddev.php` from `.gitignore` and commit it to the repository. Is that what is recommended in this scenario?

 - macOS Mojave 10.14.2
 - Docker version 18.09.1
 - ddev v1.5.2

rfay:
## The Problem/Issue/Bug:

OP #1423: Teams often share a project and don't run `ddev config` individually. So the settings.ddev.php or equivalent needs to be created on `ddev start`

Add settings creation into the post-start action.

* Remove settings.ddev.php
* ddev start
* Verify that it exists again

Unfortunately this needs to be tested in every supported CMS; TYPO3 and Wordpress and Backdrop are the most likely places for unexpected consequences.

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -93,7 +93,7 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 	// If neither settings file options are set, then don't continue. Return
 	// a nil error because this should not halt execution if the apptype
 	// does not have a settings definition.
-	if app.SiteLocalSettingsPath == "" && app.SiteSettingsPath == "" {
+	if app.SiteDdevSettingsFile == "" && app.SiteSettingsPath == "" {
 		util.Warning("Project type has no settings paths configured, so not creating settings file.")
 		return "", nil
 	}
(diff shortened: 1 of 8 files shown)


rfay:
Thanks for taking a look at this @andrewfrench . It's a simple need and a pretty simple solution, but could use extra eyes. I did change the word "Local" when referring to settings.ddev.php to "ddev". It was a carryover from when we used to use settings.local.php.

PR #1600, opened 2019-05-16
Track and list stopped/removed projects, fixes #642

### LINKED ISSUE #642: Manage ddev project list independently of docker container labels (yaml)

rfay:
## The Problem/Issue/Bug:

Not all projects have been visible to `ddev list` and related commands. This is a longstanding issue and affects ddev-ui, but is more relevant now that `ddev stop` actually removes containers, effectively removing ddev's knowledge of the existence of the project.

* Add stopped projects (which have no containers) to global config, and query from there.

## TODO:

- [x] Sort projects by active and then by name
- [x] TestMain must not remove all projects, it needs to just remove projects like "Test*"
- [x] `ddev list` should probably show only active and paused projects. Add `ddev list --all` for the rest.
- [x] Need a new `ddev stop` flag that just stops and removes from global config, doesn't do everything else (nondestructive). Use that in cmd's root_test.go setup so it's not so dangerous.
- [x] Docs: 
    - Projects don't get added until they're started once
    - How to delete a project from global config

* Test functionality with ddev-ui if it's still working
* `ddev stop -a` and `ddev list` and `ddev list --all` (which will show stopped/removed projects)
* Start a few projects and `ddev list`

There are so *very* many things that can be broken with this (starting/stopping/pausing by name, whether running or not, conflicting stuff in docker from what's in global config). Please spend some serious time poking at it, and trying to make sure you can recreate bugs.

- [x] Test for consistency between docker containers and global config project list
- [x] Test for behavior when global projectlist is not fully populated (as will be the case for lots of people)
- [ ] Test for situation where there are broken or empty projects in global config

## Related Issue Link(s):

OP #642 
This should also make ddev-ui work consistently assuming it gets revived.

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -14,7 +14,7 @@ set -o pipefail
 set -o nounset
 set -x
 
-rm -rf ~/.ddev/Test*
+rm -rf ~/.ddev/Test* ~/.ddev/global_config.yaml
 
 # There are discrepancies in golang hash checking in 1.11+, so kill off modcache to solve.
 # See https://github.com/golang/go/issues/27925
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,6 @@ testpkg: setup
 	DDEV_NO_SENTRY=true CGO_ENABLED=0 go test $(USEMODVENDOR) -p 1 -timeout $(TEST_TIMEOUT) -v -installsuffix static -ldflags '$(LDFLAGS)' ./pkg/... $(TESTARGS)
 
 setup:
-	@(mv -f ~/.ddev/global_config.yaml ~/.ddev/global_config.yaml.bak 2>/dev/null && echo "Warning: Removed your global ddev config file") || true
 	@mkdir -p $(GOTMP)/{src,pkg/mod/cache,.cache}
 	@mkdir -p $(TESTTMP)
 
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -169,8 +169,8 @@ func TestConfigSetValues(t *testing.T) {
 		"--mailhog-port", mailhogPort,
 	}
 
-	_, err = exec.RunCommand(DdevBin, args)
-	assert.NoError(err)
+	out, err := exec.RunCommand(DdevBin, args)
+	assert.NoError(err, "error running ddev %v: %v, output=%s", args, err, out)
 
 	configFile := filepath.Join(tmpdir, ".ddev", "config.yaml")
 	configContents, err := ioutil.ReadFile(configFile)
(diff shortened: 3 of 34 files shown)


rfay:
@andrewfrench this is a biggie, requesting some serious look at this from both code and manual testing perspective. Since it changes the whole model from docker-label-query to global-yaml-projectlist, there's just so very much that can go wrong. Thanks for scheduling this in.

ultimike:
I've been looking forward to this for awhile - I'm sure I've deleted a project's directory or two and forgetting to `ddev remove --remove-data` along the way.

I've been playing with the [alpha 1 version](https://twitter.com/randyfay/status/1133769670014586880) and so far, so good. I haven't run into any issues.

My only comment/request would be that the `ddev list --all` result be the default. With most folks using a standard `ddev start`/`ddev stop` workflow, using `ddev list` doesn't show anything different. I think it would be super-helpful to show stopped projects by default. Maybe include a `ddev list --no-stopped` switch or something like that...

-mike

rfay:
Thanks @ultimike. I think that's a valid request, I thought it would be too big a list for people who work on a lot of projects. Maybe it should go the other way with a --active-only.

rfay:
https://github.com/drud/ddev/pull/1636 will change the behavior to `ddev list` shows all projects, `ddev list --active-only` shows just active. Thanks!

PR #1605, opened 2019-05-17
Fix d6/d7 quick start install step

powpow12:
Users are now instructed to either run the install via '/install.php' or 'druhs si'.

## The Problem/Issue/Bug:
The quick start docs were missing the d6/d7 step of running the install script. 

Updated the doc to include those changes

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -106,7 +106,7 @@ Successfully started my-drupal7-site
 Your project can be reached at: http://my-drupal7-site.ddev.local
 ```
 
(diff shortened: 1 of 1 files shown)


rfay:
Thanks so much for this! The test fails are just a result of recent change in testbots, and will fix easily with a rebase (but really this can be [skip ci]) 

Please do sign the CLA.

powpow12:
@rfay, signed the CLA and thanks for the updates.

rfay:
Thanks so much!

PR #1611, opened 2019-05-20
Use DNS for name resolution in most cases, fixes #416, fixes #834

### LINKED ISSUE #416: Subdomain, Local DNS support: Don't require hosts file manipulation
Summarizing the acceptance criteria in this thread:

* [ ] Replace naming convention to *.ddev.site.
* [x] *.ddev.site records pointing to 127.0.0.1.
* [ ] Test and fallback to /etc/host when unable to lookup public DNS records.

### LINKED ISSUE #834: Configure the TLD, change from ddev.local to something else
## Feature request:
Would it be possible to completely remove the .dev.local and replace it with another domain name? For example, dev.example.edu.

We have SSO that only works with one domain naming convention.

Alter the YAML file to specific a completely different hostname.

rfay:
## The Problem/Issue/Bug:

It's icky messing with the hosts file and requires admin privileges.

Windows has a limit of 9 aliases on a line in the hosts file. 

OP #416 requested non-hosts-file name resolution
OP #834 requested the ability to change the TLD in use, as is provided here.

When the hostname is already resolvable, do not add it to the hosts file.

Note that this will require a transition to using "ddev.site" instead of "ddev.local"

## TODO:
- [x] Deal with the problem of a possible additional_fqdn that resolves already, provide a way to work around that.
- [x] Provide a way to completely turn off this behavior in config.yaml
- [x] This probably can't work with Docker Toolbox as currently envisioned (because the IP address is different)
- [x] Provide configuration that allows an orderly transition to using "ddev.site" instead of "ddev.local"
- [x] Change docs to use "ddev.site"
- [x] Docs must explain how this works
- [x] docs needs to say how to use ddev offline (may need `ddev start` to update, consider tld and useDNSwhen possible

* Remove all ddev hostnames from /etc/hosts
* `ddev start` should not result in hostnames being added, and project should work normally.
* Try additional_hostnames and additional_fqdns. additional_fqdns that resolve in DNS should work just fine anyway. (the resolved IP is not correct, so we add to /etc/hosts)
* Test offline. It won't work until a restart, the restart should add to /etc/hosts
* Try changing the project_tld; everything should then use /etc/hosts, but it should work fine.

- [ ] Test behavior with DNS resolution turned off (I don't think we can do this successfully because the tests don't actually modify /etc/hosts or use sudo)
- [ ] Test behavior with TLD changed in config

### DIFF
--- a/cmd/ddev/cmd/cmd_version_test.go
+++ b/cmd/ddev/cmd/cmd_version_test.go
@@ -40,7 +40,6 @@ func TestCmdVersion(t *testing.T) {
 	assert.Contains(versionData["msg"], version.GetDBImage())
 	assert.Contains(versionData["msg"], version.DBAImg)
 	assert.Contains(versionData["msg"], version.DBATag)
-	assert.Contains(versionData["msg"], version.DDevTLD)
 	assert.NotEmpty(version.DockerVersion)
 	assert.NotEmpty(version.DockerComposeVersion)
 	assert.Contains(versionData["msg"], version.DockerVersion)
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -127,6 +127,12 @@ var (
 	// lists of Debian packages to be added to related containers on build
 	webimageExtraPackages string
 	dbimageExtraPackages  string
+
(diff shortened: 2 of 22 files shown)


rfay:
@andrewfrench Another big one for your plate, but not as big. This one is easier to navigate and the results should be clear enough. Thanks!

PR #1625, opened 2019-05-24
ddev and drush should use https URL when possible inside and outside container, fixes #1624, fixes #1612

### LINKED ISSUE #1612: Disable http links after successful startup (and in ddev describe)
With the release 1.8.0 it's quite easy to use SSL certificates in DDEV. So I would like to have the option the disable the http hostnames after successful DDEV startup.
This might be no bigger problem when using one or two additional hostnames. But in a multidomain installation (i.e. TYPO3) this can get very confusing. When working with for example 10 hostnames the list will grow to 20 URL's.

Implement an option in the config.yaml like:
- list_https_only: true
- list_http_hostnames: false

**Additional context**
Bildschirmfoto 2019-05-21 um 09 59 30

### LINKED ISSUE #1624: Drush commands should use HTTPS hostname by default
Now that ddev supports pain-free HTTPS automagically in 1.8.0, Drush commands should reference the HTTPS hostname instead of the HTTP one.

All Drush commands would use the HTTPS hostname instead of the HTTP one.

damienmckenna:
## The Problem/Issue/Bug:
Drush commands should use the HTTPS hostname when possible, instead of HTTP.

* Stop using the $DDEV_URL environment variable
* Remove the /etc/drush/drushrc.php file from the container
* Use sites/default/drushrc.php (drush8) and sites/all/drush/drush.yml (drush9) to determine the uri for drush
* Abandon use of drush_ddev_settings.php and instead make the decisions it used to make inside settings.ddev.php
* Update tests
* Track the (believed) setup of mkcert during startup
* Use http or https URL based on mkcert status
* `ddev list`, `ddev start`, and `ddev describe` use the https URL when possible and show the http URL when mkcert is not usable.

* Start ddev and `drush uli` both inside and outside container. Should give correct https URL
* `rm -rf $(mkcert -CAROOT)` to mess up mkcert. Start ddev. `drush uli` should work both inside and outside container, and give http url.
* Inspect `ddev start`, `ddev describe`, and `ddev list` for usage of appropriate URLs

## Related Issue Link(s):
#1624 OP
#1612: Request to only show appropriate http/https links in ddev start/ddev describe


rfay:
As far as I can tell, DDEV_URL is targeted only at the /etc/drush/drushrc.php in the container: https://github.com/drud/ddev/blob/master/containers/ddev-webserver/files/etc/drush/drushrc.php

And it would make more sense to have that code in the project anyway (then `drush uli` would give the correct results both on the host and in the container)

And ddev knows at start time if we have pushed the CA, so could put the right thing in there. 

So I think we should go for a slightly more ambitious approach: 
* Remove DDEV_URL completely
* Add the correct URL config in ddev_drush_settings.php
* Use ddev_drush_settings.php inside the container as well as outside.

rfay:
I took the liberty of pushing in a one-line change to fix a test assumption that resulted in a panic. 

I know that my more ambitious goals above might not fit with your ambitions for this, I can follow it up if you'd like. 

I'm not sure how drush on host can know that it's operating in ddev environment, will have to think about that.

rfay:
Well, that was a lot more ambitious than you had in mind, but it cleans up a lot of things. The drushrc in  the container was unnecessary and probably misplaced, we got rid of one environment variable, got rid of ddev_drush_settings.php. And... actually accomplished the objective of using https URL if they have mkcert working.

damienmckenna:
The improvements are great, and I bow to your greater knowledge of the system, thanks for continuing the work!

rfay:
Thanks @damienmckenna . I'll do another prerelease with this in it later this week, hope you'll be able to take it for a spin.

PR #1629, opened 2019-05-27
Router should not blow away existing certificates on startup

rfay:
## The Problem/Issue/Bug:

If the ddev-router is paused and then started, it loses any existing certificates in the router's nginx config. It shouldn't do that; it works OK in many cases because of default "*.ddev.local" added there, but doesn't work if people have subdomains like "xx.yy.ddev.local", for example. 

Don't recreate certs in master.crt unless the file does not exist.

* Create a project with name "x.y.ddev.local"
* Verify that https works with it.
* `ddev pause`
* `ddev start`
* It should still work, even though you see the router being restarted


jonaseberle:
I tested this. It fixed "missing certificates when rebooting with containers running" for me (KDE Neon, Ubuntu 18.04). Thank you.

PR #1636, opened 2019-06-03
Change ddev list to default to showing all

rfay:
## The Problem/Issue/Bug:

@ultimike and @rickmanelius both recommended that the new global site listing default to `ddev list --all`, as they both got caught by missing actual global projects. 

`ddev list` shows all.
`ddev list --active-only` shows only sites active in docker.

### DIFF
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -12,8 +12,8 @@ import (
 // continuous, if set, makes list continuously output
 var continuous bool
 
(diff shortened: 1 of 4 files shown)


rickmanelius:
Awesome! Thank you for the quick turn on this :)

rfay:
And... if you have the energy to try another round or just to experience this change @rickmanelius it's in the prerelease: https://github.com/drud/ddev/releases/tag/v1.9.0-alpha2, thanks!

PR #1661, opened 2019-06-24
Provide docker pull input on ddev start, fixes #1656

### LINKED ISSUE #1656: Provide more feedback when docker-compose is pulling an image

Especially on slow connections, and when a new image is being pulled, the pull of the ddev-webserver image takes forever, and people think it has hung. 

* Get docker-compose to provide verbose information/feedback like `docker pull` does
* Warn the user if the image doesn't already exist on the machine so they know it may take some time.

**Additional context**

This is particularly bad on some Windows consoles, as you can't even hit return and get a response.

rfay:
## The Problem/Issue/Bug:

Our webimage is quite large, and downloading a new one takes some time, especially on a slower internet connection.

#1656 points out that we could at least provide some feedback while that's happening. `docker-composer up` is not that good at doing this; `docker-compose pull` can be forced to. But rather than that, 

Implement dockerutil.Pull() to pull the image with feedback, and do that on Start (if the container doesn't exist) 

* Stop all projects, `ddev rm -a`
* Find out the current web image using `ddev version` and delete it, for example `docker rmi drud/ddev-webserver:20190622_timezone`
* ddev start

You should see a wonderful amount of feedback about the download

Nothing added

## Related Issue Link(s):

OP #1656

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -747,6 +747,14 @@ func (app *DdevApp) Start() error {
 	_ = dockerutil.RemoveVolume(app.GetWebcacheVolName())
 	_ = dockerutil.RemoveVolume(app.GetNFSMountVolName())
 
+	// Pull the main images with full output, since docker-compose up won't
+	// show enough output.
+	for _, imageName := range []string{app.WebImage, app.DBImage} {
+		err = dockerutil.Pull(imageName)
+		if err != nil {
+			return err
+		}
+	}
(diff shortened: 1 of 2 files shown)


rfay:
@andrewfrench I guess just shelling out to the docker pull is another option that might be prettier, with the terminal in place so all the terminal stuff happens.

rfay:
I changed to the direct exec of `docker pull` for much prettier results, as you pointed out @andrewfrench , thanks

PR #1666, opened 2019-06-25
Special-case validation of timezone on windows, since golang windows can't do it

rfay:
## The Problem/Issue/Bug:

Testing for v1.9.0 release [here](https://github.com/drud/ddev/issues/1595#issuecomment-505458860) found that timezone validation on Windows was failing to validate correct timezones. It turns out that golang on Windows mostly can't do a timezone.LoadLocation() unless golang is installed on the machine and $GOROOT points to it. Neither of which are likely with our users.

If we get the error of type syscall.errno on Windows, assume it's a result of this and don't validate, but emit a message instead.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/drud/ddev/pkg/globalconfig"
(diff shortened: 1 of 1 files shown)


reloxx13:
So what is the solution to this? to install golang on the windows host?
(fresh win11, wsl2, latest ddev, "failed to start project: unknown timezone europe/berlin")

tyler36:
Hi @reloxx13 

The [fix](https://github.com/ddev/ddev/pull/1666/commits/79e0470f096a9a6e739a630e6986c9c583d0c73a) was merged and should be present in DDEV `1.9.0` and above. 
You can check your version with: `ddev -v`

If you are experencing an issue, please make an issue and attach the debug log. This really helps us to find a solution quicker.

stasadev:
@reloxx13, this PR targets traditional Windows, not WSL2, and the current error message in DDEV v1.25.1 doesn’t match what you provided:

> failed to start project: unknown timezone europe/berlin

https://github.com/ddev/ddev/blob/fae2ec79dd469b24f593fb40f7eab4c1a3479081/pkg/ddevapp/config.go#L583

Please open a new issue and include all the required information.

PR #1674, opened 2019-06-26
Quiet down healthchecks to save battery and CPU usage, fixes #1663

### LINKED ISSUE #1663: `ddev-ssh-agent` CPU consumption
**Describe the bug**
For a container that does not do a lot of complicated actions the CPU consumption seems a bit high: On my system (i7-7700HQ) it jumps to about 20% every couple of seconds.

To be fair, this is not a 'hard bug' - but the categorization as 'bug' seemed most fitting.

**To Reproduce**
1. Start a ddev project to ensure that the ssh agent is up
2. Observe CPU usage using `docker stats`

**Expected behavior**
I would expect the `ssh-agent` in the container to have a similar resource consumption as on my host system (as in, close to none)

**Screenshots**
screenrecord-2019-06-24_19 18 40

```bash
$ uname -a 
Linux jo-tp 5.1.8-arch1-1-ARCH #1 SMP PREEMPT Sun Jun 9 20:28:28 UTC 2019 x86_64 GNU/Linux

$ docker version
Client:
 Version:           18.09.6-ce
 API version:       1.39
 Go version:        go1.12.4
 Git commit:        481bc77156
 Built:             Sat May 11 06:11:03 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.09.6-ce
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.12.4
  Git commit:       481bc77156
  Built:            Sat May 11 06:10:35 2019
  OS/Arch:          linux/amd64
  Experimental:     false

$ docker-compose version 
docker-compose version 1.24.0, build unknown
docker-py version: 4.0.2
CPython version: 3.7.3
OpenSSL version: OpenSSL 1.1.1c  28 May 2019

$ ddev version 
bgsync        	drud/ddev-bgsync:v1.5.2       
cli           	v1.5.2                        
commit        	v1.5.2                        
db            	drud/ddev-dbserver:v1.5.2-10.2
dba           	drud/phpmyadmin:v1.5.2        
ddev-ssh-agent	drud/ddev-ssh-agent:v1.5.2    
docker        	18.09.6-ce                    
docker-compose	1.24.0                        
domain        	ddev.local                    
router        	drud/ddev-router:v1.5.2       
web           	drud/ddev-webserver:v1.5.2 
```

**Additional context**
Having had a quick look at the Dockerfile and looking at the timing of the spikes I would suspect the healthcheck script.

Also, while composing this report I noticed that I am not using the current version of ddev on this machine - however I am fairly certain that I have seen similar behaviour on my work PC. Will add a quick update when I have confirmed this behaviour on the current version.

rfay:
## The Problem/Issue/Bug:

#1663 complains about the constant running of the ddev-ssh-agent healthcheck, which uses CPU and battery. But really most of the containers have this problem, and with the other ones it multiplies.

Docker provides us only the "interval" configuration for healthcheck, which is *both* the time to wait before the first healthcheck *and* how often to run it after that. 

In ddev, we mostly care about when it *first* becomes ready and usable, and it doesn't often need to be checked after that, but we were always stymied by Docker's design mistake here.

This is admittedly a hacky approach, but

* Use a 1 second interval. Docker will wait only 1 second before trying it the first time, and will try thereafter every 1 second.
* Use a long 120s timeout, meaning docker will wait that long for a response
* Track the health status out-of-band, by creating /tmp/healthy when we find it's healthy.
* If the healthcheck finds on startup that it most recently was healthy, sleep for a defined time (configured here to a minute or so) before checking. That slows down the whole healthcheck routine.
* The start script deletes /tmp/healthy. This solves the case where a container is docker-stopped (paused in ddev parlance) and then started. The start script on resuming deletes /tmp/healthy so we get a fresh start checking healthy.

`ddev stop -a --stop-ssh-agent` and `ddev start`
Look at your computer's CPU usage.
Look at `docker events`, you should see far less than previously.
An interesting tool is `docker inspect ddev-router | jq ".[].State.Health"`, which shows the last few healthchecks and their result (change container name as necessary). This is particularly interesting early in the life of a container.

None provided. 

## Related Issue Link(s):

OP #1663

### DIFF
--- a/containers/ddev-dbserver/10.1/Dockerfile
+++ b/containers/ddev-dbserver/10.1/Dockerfile
@@ -1,4 +1,4 @@
-FROM mariadb:10.1.38-bionic
+FROM mariadb:10.1.40-bionic
 
 ENV MYSQL_DATABASE db
 ENV MYSQL_USER db
@@ -26,4 +26,4 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 EXPOSE 3306
 # The following line overrides any cmd entry
 CMD []
-HEALTHCHECK --interval=2s --retries=30 CMD ["/healthcheck.sh"]
+HEALTHCHECK --interval=1s --retries=30 --timeout=120s CMD ["/healthcheck.sh"]
--- a/containers/ddev-dbserver/10.1/files/docker-entrypoint.sh
+++ b/containers/ddev-dbserver/10.1/files/docker-entrypoint.sh
@@ -4,6 +4,7 @@ set -eu
 set -o pipefail
 
 SOCKET=/var/tmp/mysql.sock
+rm -f /tmp/healthy
 
 # Wait for mysql server to be ready.
 function serverwait {
--- a/containers/ddev-dbserver/10.1/files/healthcheck.sh
+++ b/containers/ddev-dbserver/10.1/files/healthcheck.sh
@@ -3,6 +3,25 @@
 ## mysql health check for docker. original source: https://github.com/docker-library/healthcheck/blob/master/mysql/docker-healthcheck
 
 set -eo pipefail
+sleeptime=59
 
-mysql --host=127.0.0.1 -udb -pdb --database=db -e "SHOW DATABASES LIKE 'db';" >/dev/null
+# Since docker doesn't provide a lazy period for startup,
+# we track health. If the last check showed healthy
+# as determined by existence of /tmp/healthy, then
+# sleep at startup. This requires the timeout to be set
+# higher than the sleeptime used here.
+if [ -f /tmp/healthy ]; then
+    printf "container was previously healthy, so sleeping ${sleeptime} seconds before continuing healthcheck...  "
+    sleep ${sleeptime}
+fi
+
+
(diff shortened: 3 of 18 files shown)


rfay:
This is close, but `ddev pause` (docker stop) leaves that /tmp/healthy in there, so on start, it doesn't update soon enough. Need to figure out a way to use the actual date on /tmp/healthy. Or kill it off on startup, or something.

rfay:
Artifacts for this round are at https://circleci.com/gh/drud/ddev/15097#artifacts/containers/0 for anybody who would like to take a look.

rfay:
@j6s reviewed and ccommented in https://github.com/drud/ddev/issues/1663#issuecomment-507045785

j6s:
One minor question (not sure how this project handles configuration): Would it maybe make sense to pass the sleep time down via environment variable? That way we could have something like

```
ENV HEALTHCHECK_SLEEP=59
HEALTHCHECK...
```

in the dockerfile

rfay:
I think it's absolutely worth making that configurable, thanks. I thought about it as the work went on, but just wanted to prove the concept at the time, and never got back. I'll try to add a global config for that setup if this flies with everybody.

Issue #926, opened 2018-06-17
Allow user additions to bashrc in web container


klonos:
**Is your feature request related to a problem? Please describe.**
Coming from https://github.com/beetboxvm/beetbox I have gotten used to have the `ll` command (short for `ls -la`) available OOTB when I ssh into the VM. With ddev, I have to type the whole command, which is an annoyance.

**Describe the solution you'd like**
I would like it if `ll` worked out of the box when I ssh to the VM.

**Describe alternatives you've considered**
I can set up the shortcut inside the VM, but having to do this every time I destroy/rebuild a VM is really tedious.

rfay:
I think we should probably have a directory that auto-mounts if it exists, and have the container's .bashrc include scripts in there. That way every request like this could be resolved by the user.

rfay:
I changed the title to a more generic solution. I propose that we add .ddev/bash_profile and .ddev/bashrc handling to the container's .bash_profile and .bashrc, so those get included. That way every project can have custom bash setup as needed.

johndwells:
+1 for this please.

rfay:
From https://github.com/drud/ddev/issues/1459 - host gitconfig would be completely reasonable here also.

rfay:
@pcambra requested ~/.ssh/config configuration...

rfay:
Another related request: Keep bash history around.

rfay:
Related: Request for composer auth file in #1641

PR #1688, opened 2019-07-03
Reorganize web and db containers to have real user and hostname, persistent bash history, for #926

rfay:
## The Problem/Issue/Bug:

This reorganizes the web and db containers to have a real named user which gets added at docker build time. It is intended to be the same username and id as on the host, and gets added dynamically.

It also sets the hostname inside the container to a useful value, like `<projectname>-web`

This sets the stage for several related pieces, including #926 (user .bashrc additions)

## TODO
- [x] tests
- [x] Make sure that oddball uids have a fallback
- [x] Docs? In some ways this is just a better experience and doesn't need more docs.
- [x] Make bgsync container work with the same strategy.

* `ddev start`
* `ddev ssh` and `ddev ssh -s db` - you should see a real user created with your username, and the hostname should be a nice hostname. There should be actual files in ~/, and the homedir is no longer /home.
* `ddev restart` and `ddev ssh` and you should be able to up/down through bash history from previous session. `cat $HISTFILE` should show you all that.

## Related Issue Link(s):

#926 is for lots of related things, includes the .bash_history, but this doesn't solve all yet, it's a prereq.


rfay:
@unn @andrewfrench I'd love it if you'd just take this for a spin and provide any feedback. Mostly the changes are pretty hard to parse and scattered all over creation, but I'm most worried about improving (while not breaking) the user experience in the container.

rfay:
Most recent artifacts for download are at https://circleci.com/gh/drud/ddev/15554#artifacts/containers/0, thanks for taking a spin.

rfay:
I think this is good to go. It's a lot of churn, but makes the container environment a lot easier to live in. Hoping for feedback, but will go with it if don't get any.

Issue #1706, opened 2019-07-16
Improve the suggestion in case of DDEV version mismatch


AronNovak:
The issue:
```
Your /foobar/drupal/.ddev/config.yaml version is v1.9.1, but ddev is version v1.9.0. 
Please run 'ddev config' to update your config.yaml.
```

**Describe the solution you'd like**:
Actually the message should encourage a local DDEV version update, `ddev config` is not really meaningful in this case. So we could compare version strings and have two cases in `pkg/ddevapp/ddevapp.go` , one for lower local version, one for greater local version.

**Describe alternatives you've considered**
Alternatively we should change the text to cover both cases, less logic, a bit less user friendly of course.

rfay:
This was fixed in #1707, thanks!

PR #1707, opened 2019-07-16
Better version suggestion

AronNovak:
## The Problem/Issue/Bug:
#1706 

Provides two ways to encourage the user to upgrade DDEV (config).

Downgrade local DDEV, execute on a latest configfile and check the warning message.

Tests would be nice to add later.

## Related Issue Link(s):
#1706 

## Release/Deployment notes:
No changes are needed on the user side.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -689,9 +689,13 @@ func (app *DdevApp) Start() error {
 	if err != nil {
 		return err
 	}
-	if APIVersion.LessThan(DdevVersion) {
+
+	// It returns -1, 0, or 1 if the version smaller, equal, or larger than the other version.
+	CompareVersions := APIVersion.Compare(DdevVersion)
+
+	if CompareVersions == -1 {
 		util.Warning("Your %s version is %s, but ddev is version %s. \nPlease run 'ddev config' to update your config.yaml. \nddev may not operate correctly until you do.", app.ConfigPath, app.APIVersion, version.DdevVersion)
-	} else if DdevVersion.LessThan(APIVersion) {
+	} else if CompareVersions == 1 {
 		util.Warning("Your %s version is %s, but ddev is version %s. \nPlease refer to https://ddev.readthedocs.io/en/stable/ to update your local installation.\nddev may not operate correctly until you do.", app.ConfigPath, app.APIVersion, version.DdevVersion)
 	}
 
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -682,7 +682,13 @@ func (app *DdevApp) Start() error {
 	app.DockerEnv()
 
 	APIVersion, err := semver.NewVersion(app.APIVersion)
+	if err != nil {
+		return err
+	}
 	DdevVersion, err := semver.NewVersion(version.DdevVersion)
+	if err != nil {
+		return err
+	}
 	if APIVersion.LessThan(DdevVersion) {
 		util.Warning("Your %s version is %s, but ddev is version %s. \nPlease run 'ddev config' to update your config.yaml. \nddev may not operate correctly until you do.", app.ConfigPath, app.APIVersion, version.DdevVersion)
 	} else if DdevVersion.LessThan(APIVersion) {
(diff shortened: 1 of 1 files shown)


AronNovak:
The PR still needs some love. Better code styling / formatting. Possibly some tests.

rfay:
Thanks!

AronNovak:
Tested in a separate Go file for the sake of simplicity:
```go
package main

import "fmt"
import "strings"
import "strconv"

func compareVer(a, b string) (ret int) {
  as := strings.Split(a, ".")
  bs := strings.Split(b, ".")
  loopMax := len(bs)
  if len(as) > len(bs) {
    loopMax = len(as)
  }
  for i := 0; i < loopMax; i++ {
    var x, y string
    if len(as) > i {
      x = as[i]
    }
    if len(bs) > i {
      y = bs[i]
    }
    xi,_ := strconv.Atoi(x)
    yi,_ := strconv.Atoi(y)
    if xi > yi {
      ret = -1
    } else if xi < yi {
      ret = 1
    }
    if ret != 0 {
      break
    }
  }
  return
}


func main() {
  APIVersion := "v1.9.0"
  DdevVersion := "v1.9.1."

  if compareVer(APIVersion, DdevVersion) == 1 {
          fmt.Print("ddev config")
  } else if compareVer(APIVersion, DdevVersion) == -1 {
          fmt.Print("ddev upgrade")
  }
}
```

image

rfay:
Looks like you've about got it! All go code needs to be gofmt'd, so much better than the php way, no arguments ever about code formatting. `make staticrequired` will show you any problems in advanced. `gofmt -w -s <filename.go>` will fix up the file.

AronNovak:
@rfay Ready for review. Let me know if you prefer to have a basic test for it.

AronNovak:
Test the logic in an isolated file:
image

AronNovak:
@rfay So please review the current status and let me know if it still needs a bit of work. Thanks!

rfay:
Please rebase this against current master, or ask me to do it, thanks.

AronNovak:
@rfay Sure, it's rebased now.

rfay:
And a minor comment on what we should do with that error message, just thinking out loud.

It has been more than a year since the APIVersion in the config.yaml actually meant anything or broke anything.

OTOH, we really should be using it for some things. In v1.10.0, older pinned webimage and dbimage versions were all broken, and we could have caught that by checking the APIversion (or at least looking at the webimage/dbimage versions. Also, we should could force a stop/start==restart when we see an obvious ddev upgrade happening. Those aren't in scope for this PR, I'm just thinking about what we're doing and why.

rfay:
I tested this locally, and I think that you did mean to have different error messages and don't at this point.

I took the liberty of rebasing against current master and pushing it into your branch, so you'll want to reset to the rebased fetched remote.

AronNovak:
Thanks for the test, I continue according to your advice above!

rfay:
Thanks for working on this. I'm sorry for the CircleCI build failure, has nothing to do with your code. Silly chocolatey has weird version requirements and we've entered into the twilight zone with them again. But the buildkite tests should give a good indication.

AronNovak:
Thanks. It took me a little time to make sure I can compile my own DDEV, as `make` behaved weirdly, it seemingly did nothing, then I realized that first it fetches a large Docker image before doing anything. Anyways, now I have the ability to confirm locally that with a real DDEV configfile, it gives the appropriate, different messages in both scenarios.

AronNovak:
The original scenario works well:
image

The scenario where DDEV needs an upgrade:
image

AronNovak:
@rfay So it seems the goal is reached, could you review it now, with one comparison only?

rfay:
@AronNovak should we change the wording and say something like "version mismatch" or something?

rfay:
I made minor verbiage changes, thanks so much for all your work on this. Let's get it in :)

rfay:
Thanks so much for the hard work on this @AronNovak !

AronNovak:
@rfay Thanks! I was thinking on how to defend the fact to print out different messages in different scenarios. I think it's mostly relevant for teamwork, and only for such edge cases when a functionality is used what was introduced by a later DDEV version - that's of course not typical.

Thanks for not abandoning this PR!

Issue #1710, opened 2019-07-18
Configure ImageMagick for TYPO3 inside generated AdditionalConfiguration.php


Emad-Bitar:
**Is your feature request related to a problem? Please describe.**
When Project comes configured for a production system with GraphicsMagick or another non compatible GFX configuration, image processing will not work.

**Describe the solution you'd like**
Add the following Configuration to auto generated `AdditionalConfiguration.php`:
```
$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor'] = 'im6';
$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path'] = '/usr/bin/';
$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path_lzw'] = '/usr/bin/';
```

PR #1712, opened 2019-07-18
Add TYPO3 GFX configuration for ImageMagick6

DanielSiepmann:
In order to configure system to process images. Otherwise possible
existing project configuration within LocalConfiguration.php might
prevent processing, due to configured GraphicsMagick or different paths.

Resolves: #1710 

I've didn't find instructions how to execute existing tests, therefore none were executed or adjusted or added.

## Related Issue Link(s):
#1710

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -34,6 +34,11 @@ $GLOBALS['TYPO3_CONF_VARS']['MAIL']['transport_smtp_server'] = 'localhost:1025';
 
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['devIPmask'] = '*';
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['displayErrors'] = 1;
+
+// This GFX configuration allows processing by installed ImageMagick 6
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor'] = 'ImageMagick';
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path'] = '/usr/bin/';
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path_lzw'] = '/usr/bin/';
 `
 
 // createTypo3SettingsFile creates the app's LocalConfiguration.php and


pkg/ddevapp/typo3.go
```diff
@@ -34,6 +34,11 @@ $GLOBALS['TYPO3_CONF_VARS']['MAIL']['transport_smtp_server'] = 'localhost:1025';
 
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['devIPmask'] = '*';
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['displayErrors'] = 1;
+
+// This GFX configuration allows processing by installed ImageMagick 6
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor'] = 'im6';
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path'] = '/usr/bin/';
+$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor_path_lzw'] = '/usr/bin/';
```
rfay: I see this already in the standard LocalConfiguration.php
``` 
   'GFX' => [
        'processor' => 'ImageMagick',
        'processor_allowTemporaryMasksAsPng' => false,
        'processor_colorspace' => 'sRGB',
        'processor_effects' => true,
        'processor_enabled' => true,
        'processor_path' => '/usr/bin/',
        'processor_path_lzw' => '/usr/bin/',
    ],
```

so the only change required is `$GLOBALS['TYPO3_CONF_VARS']['GFX']['processor'] = 'im6';`  right?

Is this change required for all TYPO3 projects? Or should the change be made in LocalConfiguration.php for a particular project?

DanielSiepmann: This is environment specific. As DDEV delivers ImageMagick6 at this location, this should be set.

Also some agencies might have an existing `LocalConfiguration.php` with setup for production which might differ, I would still set all three options, as they are environment specific, and therefore ddev specific.

rfay: Sorry, I'm not understanding fully. The current configuration works to generate images, we had an issue about it many months ago. The difference is using im6 right?

Please note that if you're talking about something environment-specific, you can delete the `#ddev-generated` line from the AdditionalConfiguration and take ownership of the file and ddev won't touch it any more.

DanielSiepmann: The "environment specific" is ddev specific, as this file get's generated by ddev and therefore should configure the ddev environment.

The path and version are environment (ddev) specific.

If someone has configured `/usr/local/bin` within `LocalConfiguration.php` this will not work in current setup with ddev, as ddev has installed image magick 6 at `/usr/bin/`.

In my opinion ddev should configure TYPO3 through `AdditionalConfiguration.php` to work with ddev environment.

rfay: OK, I'm good with overriding those values in AdditionalConfiguration.php, makes sense. 

I don't understand the switch from ImageMagick to im6 though. Is that something that all TYPO3 users would want?

DanielSiepmann: Yes.

DanielSiepmann: You could start a vote on Twitter or at typo3.slack.com :)

rfay: I'd just ask in TYPO3 slack...

DanielSiepmann:
Yes, the use case we encounter are agencies migrating existing project to ddev. Therefore an `LocalConfiguration.php` already exists on production system, or within VCS already. Whether it's versioned depends on the agencies and people involved. We, as most of our customers, always track this file.

rfay:
So although we'd expect LocalConfiguration.php to have "local configuration", it actually is general config, and stuff like credentials is always put into AdditionalConfiguration? I assume they're not checking credentials into LocalConfiguration? 

Anyway, thanks for helping me understand.

DanielSiepmann:
That depends on the users ;) Sometimes the AdditionalConfiguration is used to make use of a different loader, e.g. env or yaml loader.

But in context of ddev, this is the place where ddev can configure TYPO3 to work. We could also add something to the docs instead.

rfay:
I wonder if there are other things that need to be overwritten by the AdditionalConfiguration for clients in this situation.

Anyway, I'm fine with this, I just have to manually test it.

PR #1741, opened 2019-07-29
Get rid of TYPO3 trustedHostsPattern warning, fixes #1739

### LINKED ISSUE #1739: Get rid of TYPO3 trustedHostsPattern warning
TYPO3 will respect the trustedHostsPattern when this is merged: https://review.typo3.org/c/Packages/TYPO3.CMS/+/61374

Anyways, the setting trustedHostsPattern: .* will trigger a warning because the setting is potentially harmful if the server can also be reached directly (without proxy). That is OK but the warning might be disturbing.

If we add all configured hosts to trustedHostsPattern, this will not trigger a warning any more.

**Additional context**

jonaseberle:
This sets trustedHostsPattern to all configured hosts + "localhost" and "127.0.0.1".

Please excuse my zero Go skills and correct me where needed.

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -23,8 +23,8 @@ func typo3AdditionalConfigTemplate(app *DdevApp) string {
  */
 
 $GLOBALS['TYPO3_CONF_VARS']['SYS']['trustedHostsPattern'] = '` +
-		strings.Join(hostNames, "|") +
-		`';
+		strings.Join(hostNames, "(:\\d+)?|") +
+		`(:\\d+)?';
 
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/typo3.go
```diff
@@ -211,3 +219,4 @@ func typo3PostStartAction(app *DdevApp) error {
 	}
 	return nil
 }
+fixes #1739 Get rid of TYPO3 trustedHostsPattern warning
```
rfay: This line looks mighty suspicious.

jonaseberle: lol, yes

jonaseberle:
Please standby. I discovered a problem with 127.0.0.1:..

jonaseberle:
Yes, sure. That works, too :)

PR #1745, opened 2019-07-30
Move Segment items into Context, better instrumentation_user, fixes #1726

### LINKED ISSUE #1726: Update segment to send "common fields"
**Describe the bug**
Our custom `OS`, `ddev_version`, and `instrumentation_user` should be updated to leverage common fields, see links below.

**Expected behavior**
Our custom implementation of `OS` should be replaced by the common field os.
Our custom implementation of `ddev_version` should be replaced by the common field app version.
Our custom implementation of `instrumentation_user` should be replaced with a common trait like name or a custom trait. 

**Additional context**
https://segment.com/docs/spec/common/#context
https://github.com/segmentio/analytics-go/blob/v3.0/context.go#L95 
https://github.com/segmentio/analytics-go/blob/v3.0/context.go#L39

rfay:
## The Problem/Issue/Bug:

Per #1726 Segment needs to place a few things into Context, and handle instrumentation_user in its own first-class field.

This PR also adds comments with more info to the global_config.yaml

### DIFF
--- a/pkg/ddevapp/instrumentation.go
+++ b/pkg/ddevapp/instrumentation.go
@@ -9,8 +9,10 @@ import (
 	"github.com/drud/ddev/pkg/version"
 	"github.com/getsentry/raven-go"
 	"gopkg.in/segmentio/analytics-go.v3"
+	"os"
 	"runtime"
 	"strconv"
+	"time"
 )
 
 var hashedHostID string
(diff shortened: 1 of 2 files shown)


rfay:
I think this is going the way you wanted it @unn - Would appreciate a look around.  Note that the "Context" stuff is only available in the "raw" window, no idea why.

instrumentation_user is now a *trait*, and you'll see some rfay. You can configure your own in ~/.ddev/global_config.yaml with `instrumentation_user: unn`

Sources_-_ddev-local-prod_-_Segment


Sources_-_ddev-local-prod_-_Segment

PR #1760, opened 2019-08-03
[TASK] Add ngrok.io to trustedHostsPattern TYPO3 configuration

briezler:
with the implementation of the "ddev share" option accessing the page via *.ngrok.io
throws an exception and the trustedHostsPattern must be altered manually.


briezler:
Close as escaping is wrong. New PR follows

rfay:
Just FYI, you can always push new commits to an existing PR. It's much easier.

PR #1766, opened 2019-08-06
Use --renew-anon-volumes with docker-compose up to improve test reliability

rfay:
## The Problem/Issue/Bug:

In the long-running issue about `mkdir /c; file exists` there was a comment that --renew-anonymous-volumes fixed this for them. https://github.com/docker/for-win/issues/1560#issuecomment-516773475


rfay:
I don't think I'm going to pursue this one as the `mkdir /c` problem happens lots with docker, not just docker-compose.

PR #1793, opened 2019-08-14
Fix example name and Github links for all examples

arafalov:
## The Problem/Issue/Bug:
A directory must have been changed at some point.
Also, file links in Github seems to have 'blob' instead of 'tree' in the path.

Fixes the name and URLs to make them work.

Do the links work?

None. Only change is to the documentation.

## Release/Deployment notes:
Maybe the example name is used somewhere else as well? I did not check.


rfay:
Thanks for this. These examples are actually populated in .ddev, so maybe we don't even need to link to them? But showing them online is a better way to explain how they work.

arafalov:
I don't have a strong opinion. But I think that page is linked to from the release notes, so a cross-link to source does help.

rfay:
Thanks so much for catching and fixing this.

PR #1799, opened 2019-08-16
Make shebang lines more portable in host example.

wizonesolutions:
It is typically recommended to use `#!/usr/bin/env bash` in shebang lines so that the correct version of Bash gets used. I'm updating the host example here. I've left the container examples alone because we have more knowledge of the container.

I think clarifying that these scripts ought to work with any programming language that can be executed from the command line (e.g. PHP) would also be good, but I'm not sure if I'm correct on that or if it's a useful addition.


rfay:
My own experience is that using the env shebang technique is a recipe for disaster, so I always avoid it. I don't dispute that it's often recommended.

rfay:
You're correct that any reachable scripting language should work in shebang.

PR #1804, opened 2019-08-19
Update mkcert for catalina compatibility

rfay:
## The Problem/Issue/Bug:

mkcert v1.3.0 is incompatible with up-and-coming macOS Catalina

This is probably blocked by https://github.com/FiloSottile/mkcert/issues/192 - linuxbrew mkcert doesn't work right?

Upgrade to mkcert v1.4.0

On Catalina, 
* `ddev poweroff`
* `ddev start`
* Make sure the https link created is valid in Chrome


rfay:
I've tested this now in several different environments including macOS Catalina beta 6. I think it's good to go!

PR #1817, opened 2019-08-29
Stop building image tarballs by default

rfay:
## The Problem/Issue/Bug:

* The Circleci build has been creating image tarballs for every single build. We only need them on release builds.

Change the default behavior

* Make sure that tarballs aren't built in normal circle build
* Do a dummy release to make sure that they *are* built on a release build (if specified by the script)


rfay:
I tested this with 
* Regular build (no tarballs, yay!)
* Release build with tarballs (made tarballs)
* Release build without tarballs (did not make tarballs)

Issue #1490, opened 2019-03-12
Allow omitting DB container


maddy2101:
**Is your feature request related to a problem? Please describe.**
Not really a problem, just ecological and performance related concerns. I'd like not to boot up any database, if my project doesn't call for one.

**Describe the solution you'd like**
A config option, that allows to omit the configuration and creation of a DB container.

rfay:
Could you please say more about the environment in which you don't need a database please?

t3easy:
TYPO3 with SQLite Database will not need a mariadb container.

jfastnacht:
@maddy2101 Not a config option, but I asked how to do that on Stack Overflow before: https://stackoverflow.com/questions/54252928/running-ddev-without-database

Just in case you need a solution until the issue is solved.

rfay:
Just a note about sqlite3: My own experience trying to use it in a multiuser environment with Drupal was dismal, not recommended. Hoping for better with TYPO3!

lolli42:
+1 to this: I'm using ddev to run TYPO3 and symfony projects. Both run fine with sqlite, and some symfony projects don't even need a db at all. In those cases the db container is unused complexity. It would be great if it could be turned off by allowing 'db' as value for 'omit_containers' config option.

mglaman:
Chiming in: https://contribkanban.com/ is D8 and run on SQLite for ~1 year without problems.

rfay:
It's not too terribly hard to omit the db container. The thing that gets more difficult is re-doing the automatic settings.ddev.php generation and such. And adding new config that tells it that stuff, etc.

Lefaux:
I wonder if the scope is important.
I see several usecases described here.

a) The original request which is "My App simply does not need a database".
b) The requests in the sense of "well, I can run project types other than 'PHP' like Drupal or TYPO3 with SQLite.

Case A seems to be the more prominent one to me.
I currently run 10+ ddev projects based on symfony and none of them needs a DB container.

Case B seems a bit like getting out of ddev's original scope - making things **easy** for developers.
I guess we can all agree that the "sane" default for TYPO3 is a MariaDB (or MySQL) DB backend.
I could imagine the same being true for Wordpress and Drupal.

So while it's nice that you can run your CMS/App with SQlite or Postgres or Orcale or MSSQL in my opinion this gets out of hand. This is where I'd personally draw the line.
ddev does **not** offer a SQLite container, nor a Postgres or Oracle or whatever container.
[...]

PR #1819, opened 2019-08-29
Allow omit_containers["db"] for projects

rfay:
## The Problem/Issue/Bug:

#1490 requested the ability to not create the db (mariadb) container in cases where it's not used.

People want this when they're using sqlite, or postgres, or another custom database setup.

* Allow omit_containers["db"] at the project level (not global)
* If omit_containers["db"] the dba container is also omitted because it's useless
* Remove most commands that access the db in the case it's omitted (like export_db)
* Update ddev describe to leave stuff out.

* Use omit_containers["db"] and see how it goes
* Make sure things work without it

No changes at this point


rfay:
Artifacts for this are at https://circleci.com/gh/drud/ddev/17589#artifacts/containers/0

PR #1827, opened 2019-09-11
Rework pantheon integration to use terminus inside web container, fixes #1731, fixes #1232

### LINKED ISSUE #1232: Allow ddev config pantheon to specify an explicit site hash to override what the API may find (or fail to find)

I'm an admin for UC Berkeley -- we have something like 1500 campus sites in our Pantheon org.  When I run `ddev config pantheon` sometimes it can't find the site (I get a "could not find a pantheon site named sitename" message.). Passing a `projectname` doesn't appear to help find the site either.

I tried running `ddev config` instead.  Then I changed the following in config.yaml: `provider: pantheon`. Then I copied an import.yaml file from a different site that Pantheon could find, and changed `site` to the correct Pantheon machine name.

Unfortunately when I try running `ddev pull` (or `ddev start`) it still doesn't find the site -- it appears to still be searching Pantheon as opposed to using my local yaml changes.

It appears that ddev's `projectname` is searching for and trying to match the sitename string, not the hash.

It would be helpful if we could specify the site via `ddev config pantheon --projectname string` or something like `ddev config pantheon --projectID hashnum` since we can find the Pantheon site ID via `terminus site:info sitename`.  I'm hoping that would avoid having ddev try to look up the site by name each time and failing if it exceeds the Pantheon API limit.

We're able to use `ddev config` and then manually import db/files, but we're looking to use a solution that provides better Pantheon integration.

### LINKED ISSUE #1731: Pantheon/Terminus upgrade required soon (Sept 2019?)
**Describe the bug**

@cyberswat received this from Pantheon; my assumption is that this means we need at least a modest rework of the [go-pantheon](https://github.com/drud/go-pantheon) library @beeradb built: 

> On August 26, 2019, Pantheon will no longer support Terminus versions older than 1.9. We've noticed that in the last month you've used one of these versions, so we wanted to reach out to ask you to update your scripts. Please check all cron jobs or other scripted processes which rely on Terminus to make sure they are on 1.9 or greater.

> We've released Terminus 2.0, hoping to solve many of the concerns of why users haven't upgraded. If your team has questions, concerns, or needs help updating your scripts to the latest Terminus version, please reply to this ticket—we are here to help.

Strategy:
* Try just updating the api version and see if it works
* Postman api study time
* Try to fix the many-sites issue in https://github.com/drud/ddev/issues/1232 while we're there.

rfay:
## The Problem/Issue/Bug:

#1731: The Pantheon API is changing (and remains undocumented), so we'll use terminus instead.
#1232: Ddev config didn't seem to work in particularly large environments. Test to see if this solves it since we're using terminus directly.

* Include terminus inside the web container
* Since terminus isn't supported on Windows, that solves the windows problem
* `ddev pull` can use terminus with an exec to accomplish what's needed

- [x] `ddev auth-pantheon` should become `ddev auth pantheon`
- [x] Update docs for all usages, but change from auth-pantheon to `auth pantheon`

* Read the [(new) docs](https://github.com/drud/ddev/blob/ba93db069c7d17f509ee3eb53160c6ecd6c4dbcd/docs/users/providers/pantheon.md), as `ddev auth-pantheon` becomes `ddev auth pantheon`. 
As in the standard instructions, you need a pantheon site to work with, and get a token there.
* `ddev auth pantheon <token>`
* `ddev config pantheon`
* `ddev pull`

* Return to #1232 and see of one of those folks can test to see if that problem is resolved.

The original tests are still there.
One previously commented-out test has been added back in.

## Related Issue Link(s):

OP #1731 

## Release/Deployment notes:

* ddev auth-pantheon becomes ddev auth pantheon


rfay:
Current artifacts for testing are at https://circleci.com/gh/drud/ddev/20987#artifacts/containers/0

gchaix:
Testing on Ubuntu 19.10
```
gchaix@sonic:~/repos/ddev$ ./.gotmp/bin/ddev version 
DDEV-Local version      v1.12.0-21-gd5387a7c
commit                  v1.12.0-21-gd5387a7c
db                      drud/ddev-dbserver-mariadb-10.2:v1.12.0
dba                     drud/phpmyadmin:v1.12.0
ddev-ssh-agent          drud/ddev-ssh-agent:v1.12.0
docker                  19.03.3
docker-compose          1.25.0-rc4
os                      linux
router                  drud/ddev-router:v1.12.0
web                     drud/ddev-webserver:20190911_terminus_pantheon
```
Successfully tested using a Pantheon machine token with access to 20 separate sites.

`ddev auth pantheon <token>` 👍 
`ddev config pantheon` 👍 
`ddev pull` 👍 at least as far as pulling and loading the database - the site I'm testing has a *huge* files directory that is still downloading. :-)

Issue #1794, opened 2019-08-14
Configure ddev-router to listen on 0.0.0.0


froemken:
Hello DDev Team,

I have my own test remote server where I can play around with DDev and other services.
I'm nearly sure that it was possible in earlier DDev versions to allow ddev-router to listen on 0.0.0.0, so that I can open my ddev-Projects with project.my-domain.de

It seems, that you have switched everything to ngrok. You have remove ddev-router configuration from docker-compose.yaml so that I can override it anymore. I have configured additional hosts and port to web-service, but ddev-router still listen on default ports.

Any idea how to listen ddev-router on 0.0.0.0 again?

Stefan

rfay:
Please try `ddev share` for this, or use the technique in https://stackoverflow.com/questions/57480512/expose-ddev-web-port-to-network

rfay:
You're definitely right that ddev-router is no longer listening on 0.0.0.0, that was a pretty big security hole to have it exposed to the entire local network. Will be interested to hear if `ddev share` does the job for you. 

This is also a duplicate of #1792

froemken:
Hi @rfay 

I have created a ngrok account, but I can't access my TYPO3 System. Error 500. In logs I found that trustedHostPattern does not match. Ok. I have started ddev share to get ngrok URI. Copy URI -> CTRL+C -> edit AdditionalConfiguration.php -> update trustedHostPattern -> save -> start ddev share again -> WTF -> The URI has changed. Nice. CTRL+C -> ddev share -> Again, a new URI.

OK, I have tried ddev share --subdomain myproject
Ding! You're using the free account. Narf! Yes, I'm using the free account, because it's my personal private server and there is no boss/company who will pay me my private test/playing/whatever server.

Any ideas?

Stefan

rfay:
ddev start updates the trustedHostPattern in AdditionalConfiguration.php on every start unless you remove the #ddev-generated at the top of the file.

rfay:
Maybe you can say a little more about why you liked your entire network to be able to reach your ddev project, and how you set that up. You had edited /etc/hosts on every machine that you wanted to use to access it? Was it for testing different browsers? Or what was the purpose?

froemken:
Hi @rfay 

[...]

Issue #1792, opened 2019-08-14
Web environment not accessable through network since 1.10.0


FamousWolf:
**Describe the bug**
Since version 1.10.0 the web environment is only accessible from the host, not from any other computer on the network (see #1662). I understand this is done because on public networks it is a security problem if the environment is publicly accessible. However we use ddev on a secure company network. We use an internal DNS and `additional_fqdns` in .ddev/config.yaml to be able to access each developer's environment using project.developername.domain. Using `ddev share` or manually setting the port each time a ddev environment is started are both not ideal solutions. It would be great if this change can be made optional with a setting in .ddev/config.yaml.

**To Reproduce**
1. Start a ddev environment
2. Try to access it from a different computer on the same network

**Expected behavior**
The environment should be accessible from a different computer on the same network

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: Ubuntu 18.04
 - Docker version: 19.03.1
 - ddev version: 1.10.2

WebsiteDeveloper:
@FamousWolf couldn't you just also use [docker-compose.letthemin.yaml](https://stackoverflow.com/a/57481412/215713) as suggesten in the 1.10 release notes

FamousWolf:
Unless I misunderstood that will make it available through the random port which changes every time ddev (re)starts. I'm not that familiar with docker, so chances are I did misunderstand, but I did try it and it didn't work as desired.

rfay:
@FamousWolf if the stack overflow technique didn't work... Please say what you wanted differently. Also, isn't `ddev share` a better solution?

What did you used to do? Because managing hostnames on each client on your network is difficult.  And of course it's a massive security flaw to *assume* that people want their ddev project exposed to everyone on their network. That's why it was tightened up.

FamousWolf:
Thanks. #1794 Looks like exactly what I need. I'll try it later today.

FamousWolf:
The solution in #1794 does work. It would be nice if there was just an option you could set in config.yaml, but this will do for now.

`ddev share` is fine to quickly share your environment with someone once, but unless you have a paid account the changing subdomains will become a pain quickly. It also feels dirty (for lack of a better description) to route a local development environment through an external service online just so someone who's on the same private network as me can access it.

We have an internal DNS that handles *.developername.internaldomain. We then use the additional_fqdns option to add the domain for the project (projectname.developername.internaldomain) to ddev.

[...]

PR #1834, opened 2019-09-16
Allow users to bind router on all network interfaces, fixes #1798

### LINKED ISSUE #1798: Option to expose web environment to network
Since version 1.10.0 the web environment is only accessible from localhost. Before it was accessible from other devices on the same network. We use ddev for multiple projects. Often we need to try/view something on a colleague's environment. So we have an internal DNS to point *.developername.internaldomain to the development environments. We then use `additional_fqdns` in config.yaml to make it available through projectname.developername.internaldomain.

I would like a configuration option `exposeToNetwork` (or something similar) in config.yaml to expose the web environment to the network. Default should be false.

We've tried `ddev share`, but without a paid account the subdomain keeps changing. We're currently using the solution in #1794, which works, but a cleaner solution would be great.

rfay:
## The Problem/Issue/Bug:

OP #1798: In ddev v1.10, the feature/bug where ddev bound to all network interfaces was removed. Previously everyone on a local network could access ddev sites (and PHPMyAdmin). 

It was a surprise to find out that there were people relying on this feature/bug to share projects with others or to use ddev for CI/CD/Staging applications.

Allow a new global configuration option  in ~/.ddev/global_config.yaml: `router_bind_all_interfaces: true` that uses the original behavior. 

```
ddev poweroff
ddev config global --router-bind-all-interfaces=true
ddev start
```
* Inspect ~/.ddev/router_compose.yaml; you should see the port bindings without IP addresses specified.
* Access project from an IP address on the machine that is *not* localhost

* Adds testing only for the ddev config global option; it seems too hard to know what ip addresses to use in so many testing environments.

## Related Issue Link(s):

#1794 requested this configuration
#1792 also
https://github.com/drud/ddev-contrib/pull/9 provided an interesting workaround that used the ddev-router in an unusual way.

## Release/Deployment notes:

It's important for people to understand the security implications. 
At least using `omit_containers: [dba]` in the ~/.ddev/global_config.yaml is recommended, as that prevents the PHPMyAdmin interface from being used across the local network.

### DIFF
--- a/cmd/ddev/cmd/config-global.go
+++ b/cmd/ddev/cmd/config-global.go
@@ -43,6 +43,9 @@ func handleGlobalConfig(cmd *cobra.Command, args []string) {
 			globalconfig.DdevGlobalConfig.OmitContainers = strings.Split(omitContainers, ",")
 		}
 	}
+	if cmd.Flag("router-bind-all-interfaces").Changed {
+		globalconfig.DdevGlobalConfig.RouterBindAllInterfaces, _ = cmd.Flags().GetBool("router-bind-all-interfaces")
+	}
 	err = globalconfig.ValidateGlobalConfig()
 	if err != nil {
 		util.Failed("Invalid configuration in %s: %v", globalconfig.GetGlobalConfigPath(), err)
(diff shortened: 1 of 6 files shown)


andrewbelcher:
Huge +1 from me. As you said:

> It was a surprise to find out that there were people relying on this feature/bug to share projects with others or to use ddev for CI/CD/Staging applications.

We have been looking at DDev as a possible way to build review apps as part of automated CI. Our production system doesn't use docker, but we have DDev configured to mirror it's setup for easy local development. Spinning up instances in DDev hopefully requires only setting a custom project name and additional FQDN, rather than having to deal with configuring multiple vhosts in apache or nginx.

I'm a bit backed up at the moment, but when I get a chance, I will give setting this up a go. If it's still awaiting review I'll give feedback on my experience!

rfay:
I expect this to make a prerelease this week, so should be easy to test there. If anybody wants to test now, the artifacts from this PR are at https://circleci.com/gh/drud/ddev/17811#artifacts/containers/0

PR #1835, opened 2019-09-17
Use much simpler TYPO3 trustedHostsPattern

rfay:
## The Problem/Issue/Bug:

Several recent PRs have tried to catch up with TYPO3 v1.10's rejection of ".*" as trustedHostsPattern, and it's gotten super complex.

Since TYPO3 only checks and complains about the literal `.*` we might as well fool it by using `.*.*`. It's for development after all!

Install TYPO3 v10 using composer, then do an install
* Test with various hostnames/fqdns
* Test with ngrok/ddev share

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -3,29 +3,23 @@ package ddevapp
 import (
 	"fmt"
 	"io/ioutil"
-	"strings"
 
 	"os"
 	"path/filepath"
 
 	"github.com/drud/ddev/pkg/archive"
-	"github.com/drud/ddev/pkg/dockerutil"
 	"github.com/drud/ddev/pkg/fileutil"
 	"github.com/drud/ddev/pkg/output"
 	"github.com/drud/ddev/pkg/util"
 )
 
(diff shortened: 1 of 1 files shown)


rfay:
Rebased this and tested with a TYPO3 v10 install and then access, and with `ddev share` and it seems to work fine. When it passes tests I think it's good to pull.

gilbertsoft:
@rfay This does not work for TYPO3 8.7.29, getting a warning in the install tool. But as this version is EOL in 5 months I don't think we have to change this again. And I'm not sure if there was a running version for 8.7 at all. The user could simply click ignore and the installation continues without any other issues.

rfay:
The problem on TYPO3 v8 is that it doesn't actually include AdditionalSettings.php during install. As a result we recommend installing using http in the quickstart.  @jonaseberle made a patch to fix this in v9, https://review.typo3.org/c/Packages/TYPO3.CMS/+/61374 

So the answer in v8 is "use http"

gilbertsoft:
Yeah, I remember now. It's only a warning on the first install, afterwards everything works as expected. No need to not to use TLS for this one time step imho. But thanks anyway!

PR #1836, opened 2019-09-17
Make settings.php manipulation optional

cweagans:
## The Problem/Issue/Bug:

Our settings.php already has configuration for ddev and we'd prefer to own the responsibility of managing the file on our own.

Allows users to opt out of settings file management by doing `ddev config --omit-settings-php`. Subsequent `ddev config`s will remember this and continue to not manipulate settings.php.

Probably still needs docs and tests, but I'm opening this PR early to see if this is a desired feature.

Run `ddev config --omit-settings-php` and verify the following:

1. `settings.php` isn't modified
2. `settings.ddev.php` isn't generated
3. `sites/default/.gitignore` isn't generated
4. `omit_settings_php: true` is included in `.ddev/config.yaml`
5. Another `ddev config` does all of those same things ^^

No tests currently. Happy to add them if needed. Same with docs.

## Related Issue Link(s):

n/a

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -2,11 +2,12 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"strings"
+
 	"github.com/drud/ddev/pkg/globalconfig"
 	"github.com/drud/ddev/pkg/version"
 	"github.com/mitchellh/go-homedir"
-	"os"
-	"strings"
 
 	"path/filepath"
 
@@ -136,6 +137,9 @@ var (
 
 	// ngrokArgs provides additional args to the ngrok command in `ddev share`
 	ngrokArgs string
+
+	// omitSettingsPhp skips generating a database settings file.
+	omitSettingsPhp bool
 )
 
 var providerName = ddevapp.ProviderDefault
@@ -293,6 +297,8 @@ func init() {
 	ConfigCommand.Flags().StringVarP(&ngrokArgs, "ngrok-args", "", "", "Provide extra args to ngrok in ddev share")
 
 	ConfigCommand.Flags().String("timezone", "", "Specify timezone for containers and php, like Europe/London or America/Denver or GMT or UTC")
+
+	ConfigCommand.Flags().BoolVar(&omitSettingsPhp, "omit-settings-php", false, "Do not write a database settings file and do not modify the existing settings file")
 
 	RootCmd.AddCommand(ConfigCommand)
 }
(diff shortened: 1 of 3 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -32,7 +33,7 @@ import (
 	"github.com/drud/ddev/pkg/output"
 	"github.com/drud/ddev/pkg/util"
 	"github.com/drud/ddev/pkg/version"
-	"github.com/fsouza/go-dockerclient"
+	docker "github.com/fsouza/go-dockerclient"
```
cweagans: Seems to still compile, but one of the tools vscode uses under the hood made this change. Seems to still compile, so maybe this is a desirable change?

pkg/ddevapp/apptypes.go
```diff
@@ -88,6 +88,13 @@ func init() {
 // CreateSettingsFile creates the settings file (like settings.php) for the
 // provided app is the apptype has a settingsCreator function.
 func (app *DdevApp) CreateSettingsFile() (string, error) {
+	// If the user has asked us to skip settings file manipulation, then just bail
+	// out early.
+	if app.OmitSettingsPhp {
+		util.Warning("Skipping creation of settings file.")
```
cweagans: The styling/wording here probably needs some thought. It's more of a placeholder.

cmd/ddev/cmd/config.go
```diff
@@ -294,6 +298,8 @@ func init() {
 
 	ConfigCommand.Flags().String("timezone", "", "Specify timezone for containers and php, like Europe/London or America/Denver or GMT or UTC")
 
+	ConfigCommand.Flags().BoolVar(&omitSettingsPhp, "omit-settings-php", false, "Do not write a database settings file and do not modify the existing settings file")
```
cweagans: The actual flag probably needs a better name too. Open to suggestions.

rfay:
Why not just use project type 'php' and avoid this altogether?

cweagans:
I forgot that was a thing :sweat_smile:

Are there any other differences between the `drupal8` project type and the `php` project type?

rfay:
I think that drupal8 and php nginx configs are the same, but other than that there isn't any difference that I remember. Experienced people like you are probably better with php project type, just handle all settings yourself. Even if there is something you want to do in nginx, do it in custom nginx config, which you may want anyway.

There is a slight difference:

```
$ diff nginx_default.conf nginx_drupal8.conf
1d0
<
14c13
<         try_files $uri $uri/ /index.php?q=$uri&$args;
---
>         try_files $uri /index.php?$query_string; # For Drupal >= 7
29,30c28
<     location ~ \.php$ {
<         try_files $uri =404;
---
>     location ~ '\.php$|^/update.php' {
52,58d49
<     # Media: images, icons, video, audio, HTC
<     location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
<         expires 1M;
<         access_log off;
<         add_header Cache-Control "public";
<     }
<
81a73,79
>     }
>
>     # Media: images, icons, video, audio, HTC
>     location ~* \.(js|css|png|jpg|jpeg|gif|ico|svg)$ {
>         try_files $uri @rewrite;
>         expires max;
>         log_not_found off;
```

cweagans:
Meh, close enough. Thanks for the pointer!

rfay:
Always nice to see you around here!

PR #1842, opened 2019-09-19
Updated custom command docs with a little more detail

chellman:
## The Problem/Issue/Bug:

When creating my first custom command, there were a few subtleties that I didn't follow. 

I moved the bit about scripts needing to be executable to the top, along with a couple other things I discovered that should be helpful.


rfay:
Thanks for this! In the next version there's a bunch of maturing that needs to be done for custom commands. Do you think there would be anything wrong with ddev making those executable just as part of the process?

chellman:
> Thanks for this! In the next version there's a bunch of maturing that needs to be done for custom commands. Do you think there would be anything wrong with ddev making those executable just as part of the process?

I don't see any reason why not - automating any simple steps like that seems like the right call.  If I need the file not to be executable, I can just remove the flag myself.

rfay:
Thanks for this!

PR #1854, opened 2019-09-28
WordPress: Use the secure URL for `WP_HOME`

gormus:
## The Problem/Issue/Bug:
Every time I run `ddev config`, DDEV overrides the `wp-config-ddev.php` for the `WP_HOME` constant, and uses HTTP protocol for that URL.
However DDEV already defaults to secure URLs, this setting should match that too. 

Changes `DeployURL` to use a secure URL, so that `WP_HOME` can inherit that too.

### DIFF
--- a/pkg/ddevapp/wordpress.go
+++ b/pkg/ddevapp/wordpress.go
@@ -46,7 +46,7 @@ func NewWordpressConfig(app *DdevApp, absPath string) *WordpressConfig {
 		DatabaseUsername: "db",
 		DatabasePassword: "db",
 		DatabaseHost:     "db",
-		DeployURL:        app.GetHTTPURL(),
+		DeployURL:        app.GetHTTPSURL(),
 		Docroot:          "/var/www/html/docroot",
 		TablePrefix:      "wp_",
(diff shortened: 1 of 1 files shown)


rfay:
Seems reasonable to me.

I see this actually ends up getting used as $WP_HOME at  https://github.com/drud/ddev/blob/2948e7e5d250abfc85dbed76e5af6804e15a912c/pkg/ddevapp/wordpress.go#L143

rfay:
In https://github.com/drud/ddev/pull/1862 DDEV_PRIMARY_URL and app.GetPrimaryURL() are introduced. I think that will be better to use than just using the secure URL by default.

rfay:
#1862 has landed, so this can now use the primary URL, will take a look a bit later. In the meantime, rebased this to current upstream/master.

rfay:
I took the liberty of rebasing this and using the app.GetPrimaryURL(). However, it doesn't actually work yet with WP because of the basic problem described in https://github.com/drud/ddev/issues/1736 - we need to get that one fixed before this has any value.

PR #1856, opened 2019-10-01
Add 'IS_DDEV_PROJECT' to environment, require it for drush-on-host

andy-blum:
## The Problem/Issue/Bug:
We occasionally need to use the same settings/configuration files between local development and dev/test/prod environments, but want to set different settings for those environments.

* This creates a "IS_DDEV_PROJECT" variable on $_ENV with a set value of "true" to ease code readability (e.g. `if ($_ENV['IS_DDEV_PROJECT']) { ... }`)
* ddev-related settings in settings files are only applied if IS_DDEV_PROJECT==true
* Running drush on the host now will only work if the host-side environment variable IS_DDEV_PROJECT==true


rfay:
Thanks! In what context do you want to use $IS_DDEV_PROJECT? Only in web container? On host? 

Mostly we've used any one of the other DDEV environment variables for this purpose (check existence).  

Note that you can also accomplish this for your own team or projects with [custom environment variables](https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-environment-variables-to-a-container) if you're interested in setting them inside the various containers.

andy-blum:
The biggest way I envision using this is in Drupal 8's settings files when doing the initial project configuration. We had an issue the other day with ddev's drushrc overriding something internal in Pantheon hosting leaving the Site URI as 'https://site-name.ddev.site' while on pantheon servers. Additionally, we had the settings.ddev.php in the repository and it's included if the file is present, but we really only want the file included if running on a ddev environment.

rfay:
BTW, the reason all the tests fail is that it doesn't compile.
```pkg/testcommon/testcommon.go:218:20: syntax error: unexpected newline, expecting comma or }
```

Please compile and test, it doesn't even actually require golang installed, although that probably comes out cleaner; actual building can be done with just `make` and it will build the OS you're working on. 

Minor build/test docs at https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md
 





pkg/testcommon/testcommon.go:218:20: syntax error: unexpected newline, expecting comma or }
 




pkg/testcommon/testcommon.go:218:20: syntax error: unexpected newline, expecting comma or }

andy-blum:
Interesting, I'm also not sure where it came from then, as I'm running v1.11.1

The issue I experienced came after merging a couple branches back together, so I'll have to investigate that.

rfay:
Thanks. Avoiding leakage of ddev-specific things is a high priority, so please let us know if you can figure out how the leakage happened. We're always working on better ways to do the drush integration too.

rfay:
One thing I'd like to make sure happens perhaps in this PR is to make absolutely sure that settings.ddev.php doesn't get included unless ddev is the local dev.

rfay:
Rebased this to current upstream/master

rfay:
Suggestion from @warped in slack: We should have a way to determine if a project is *started* as well as whether it's a project.

rfay:
This is still valid and alive. I just want to evaluate it in the context of improving drush support.

rfay:
Rebased this, we're going to get it in!

rfay:
I'm still saving this for when I work on drush multisite support again, which I think will be soon. Maybe I'll get it in there before then. It's a great idea and going to be great benefit, sorry it's going so slowly.

rfay:
@andy-blum I finally got back to this, after some drush study and settings study and such. I hope you'll take a look and see if you like the results.

andy-blum:
I don't really understand how ddev's source code works, but from what I can gather this looks good.

Is there a way to download this and test it (preferably without breaking my Homebrew install)?

rfay:
Great question, yes artifacts for this one are at https://circleci.com/gh/drud/ddev/21963#artifacts/containers/0

You have to open the hierarchy to get to the downloads:

Continuous_Integration_and_Deployment

PR #1874, opened 2019-10-09
Allow mariadb_version 5.5, 10.0/1/2/3/4 and mysql 5.5/6/7 and 8.0, fixes #1870, fixes #1627

### LINKED ISSUE #1627: Consider supporting additional mysql variants (mysql 5.7, 8, mariadb 10.3+) and Postgres?

People generally like to have the dbserver be the same software they're using on the server. Since mariadb isn't really a drop-in replacement for anything any more, the differentiation continues. 

Unfortunately, we hitched our wagon to mariadb some time ago, when it was in fact pretty much a drop-in replacement and could meet people's needs. So we use the terminology "mariadb_version" in config instead of something more general. So this would be something that would have to change.

Maybe Postgres too?

I'm pretty sure we'd need a separate container for each, but can do a far better job building from the same base than we're currently doing, which is all copy/paste and not well maintainable. 

We'd have to change the configuration mechanism and naming of the image.

### LINKED ISSUE #1870: MariaDB v10.3
Is it possible to provide MariaDB 10.3 in the existing DB container?

rfay:
## The Problem/Issue/Bug:

#1870 (along with #1627) requested MariaDB 10.3 and mysql versions. People have requested it for compatibility with various CMSs like Shopware.

* Allow MariaDB 5.5 through 10.4
* Add Oracle MySQL 5.5, 5.6, 5.7, 8.0
* Dramatic simplification of the dbserver image

Try using `mariadb_version: 5.5` etc.
Try using `mysql_version: 8.0` etc
Try using both together (it should complain)
Try snapshots of some. And restores.

* The container tests use all these versions
* The ddevapp tests also use all these versions.

## Related Issue Link(s):

OP #1870 (mariadb 10.3)
OP #1627 (mysql variants)


rfay:
@jeffsheltren (and anybody else!) thanks for reviewing the my.cnf.

The goal are (if possible) is to 
* use a single my.cnf for all supported versions of MariaDB and MySQL. It's a bit hard, mostly because of MySQL 8. 
* Have a reasonable config for the vast majority of people
* Not break people upgrading to the next version of ddev.

So I'm open to both suggestions about how we can do it better and also to concerns about the changes here.

Here's the [new my.cnf](https://github.com/drud/ddev/blob/2f2349dfa23a5e27bfbbf67793bc41fab44b1243/containers/ddev-dbserver/files/etc/my.cnf)
Here's the [old my.cnf](https://github.com/drud/ddev/blob/master/containers/ddev-dbserver/10.2/files/etc/my.cnf)

Here's the diff:
```
diff old.my.cnf new.my.cnf
29c29
< # log-bin=on fixes a bug with Virtualbox (used by docker toolbox)
---
> # log-bin fixes a bug with Virtualbox (used by docker toolbox)
60,61c60,62
< query-cache-type               = 0
< query-cache-size               = 0
---
> ; these two are incompatible with mysql 8.0
> ;query-cache-type               = 0
> ;query-cache-size               = 0
72c73,75
< innodb-log-file-size           = 64M
---
> ; 48M is chosen here only to work around xtrabackup bug
> ; for mariadb 5.5, https://bugs.launchpad.net/percona-xtrabackup/+bug/1527081
> innodb-log-file-size           = 48M
76,77c79,81
< innodb_large_prefix=true
< innodb_file_format=barracuda
---
> ; these two are incompatible with mysql 8.0
> ;innodb_large_prefix=true
> ;innodb_file_format=barracuda
```

jeffsheltren:
Thanks for the ping @rfay 

I don't see an issue with disabling the query cache. That may cause performance issues on some sites, but it's generally better to fix poorly performing queries than to rely on query cache. This page sums it up pretty well: https://mysqlserverteam.com/mysql-8-0-retiring-support-for-the-query-cache/

I do think there will be an issue with removing the Barracuda config on older MySQL versions because that's required for utf8mb4. I haven't seen docs about why that config option (innodb_file_format) went away in MySQL 8, but my guess is they've just dropped support for the older file format, so there's only one possible format any longer. 5.7 defaulted to Barracuda, but previous versions defaulted to Antelope. Maybe that's fine, as users can override in their own cnf file, but you'd want to document it at the very least.

rfay:
Thanks so much for your review of my.cnf in https://github.com/drud/ddev/pull/1874#issuecomment-542385827 @jeffsheltren  -

I haven't had any trouble with any versions of mysql and the omitted Barracuda type... and we have automated tests for all. Does that mean we're off the hook? Or would it show up in a more interesting way?

rfay:
Yeah, @andrewfrench I think just having mutually incompatible mysql_version and mariadb_version is pretty questionable :)

PR #1884, opened 2019-10-14
Use nsis from linuxbrew (3.04) instead of from apt-get (ancient)

rfay:
## The Problem/Issue/Bug:

In https://github.com/drud/ddev/pull/1880 @gilbertsoft noted that the Circleci/Linux makensis we're using is really old. It seems we were getting it from apt-get (and those testbots are Ubuntu 16.04, a few years old). 

* Use homebrew to build makensis
* Unzip the standard 3.04 nsis tarball into /usr/local/share/nsis so that we get Stubs


gilbertsoft:
Nice :)

PR #1891, opened 2019-10-21
Added links to ddev-contrib for some extra services and old PHP

chellman:
## The Problem/Issue/Bug:

As discussed on Slack, people don't always know to look at ddev-contrib for support of relatively common, but not built-in, services.

This PR updates three pages to add references to ddev-contrib recipes for things like Blackfire, MongoDB, and older versions of PHP.  It also links to ddev-contrib from the "defining additional services" page so people can use them as examples.

### DIFF
--- a/docs/users/extend/additional-services.md
+++ b/docs/users/extend/additional-services.md
@@ -70,21 +70,12 @@ This recipe adds a [Beanstalk](https://beanstalkd.github.io/) container to a pro
 - The Beanstalk instance will listen on TCP port 11300 (the beanstalkd default).
 - Configure your application to access Beanstalk on the host:port `beanstalk:11300`.
 
-## MongoDB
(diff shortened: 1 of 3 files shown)


rfay:
Thanks!

rfay:
@chellman I rebased and took the liberty of making the main stanza a little more terse. If this change is OK with you, this is ready to pull. Let me know. Diff in https://github.com/drud/ddev/pull/1891/commits/94717793679e2452db58ebee3032ebd6323cb58f

chellman:
I think that's just fine! I was a little verbose because ~~I can't help myself~~ I thought it might be useful for search purposes to have things a little more spelled out, but I do think it reads better this way. Great!

rfay:
Thanks so much for taking the initiative on this.

PR #1894, opened 2019-10-23
Update php nginx config to be more compatible with modern PHP applications

cweagans:
## The Problem/Issue/Bug:

Running Drupal 8 + the redirect module with the php config results in an endless redirect. More generally, other modern PHP applications don't expect `?q=` to designate what the path is.

Just re-use the D8 rewrite for the php app type.

### DIFF
--- a/containers/ddev-webserver/files/etc/nginx/nginx_default.conf
+++ b/containers/ddev-webserver/files/etc/nginx/nginx_default.conf
@@ -11,7 +11,7 @@
 
     location / {
         absolute_redirect off;
-        try_files $uri $uri/ /index.php?q=$uri&$args;
+        try_files $uri /index.php?$query_string;
     }
 
     location @rewrite {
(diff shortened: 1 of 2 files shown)


rfay:
@cweagans this seems to break the https redirection fix added in https://github.com/drud/ddev/pull/1124, seems to break it on every test platform.

rfay:
I'm looking forward to a more generic default nginx config, but it will need more than this, and has to work for most CMSs and platforms. The current one does work for lots of things, but this one doesn't. So I'm going to close this out. It's a valid objective, but not met with this one.

cweagans:
FWIW, I don't think one config _can_ work correctly for both legacy applications and modern applications. If that's what's required of the `php` app type, then I'd like to revisit https://github.com/drud/ddev/pull/1836 if you don't mind. The `php` app type simply doesn't work for Drupal 8 as-is, but having ddev mess with our settings.php even though we've already taken ddev environments into account is pretty undesirable.

rfay:
I'm all for another try at #1836, and happy to reopen it if you want to push stuff into it. But it needs to be tested with at least some basic projects. 

FYI on your situation, of course, you can check in an nginx config in .ddev/ to do whatever your project requires, and you're in complete control that way.

PR #1906, opened 2019-10-28
Improved Node install docs

boulaffasae:
## The Problem/Issue/Bug:

- Setting up a Local dev environment with Browsersync (a Node.js application that watches your files for changes then refreshes your browser window or injects new CSS into the page.), require exposing additional port (e.g.: 3000). `web` service wont let us alter `HTTP_EXPOSE` to expose additionals ports to serve for container.

- This recipe adds a Node container to a project. It will set run an arbitrary command specified in the package’s (custom Drupal theme) "start" property of its "scripts" object.

- Create a custom Drupal theme, add a `package.json` to the theme with a "start" property in its "scripts" object.

- Test require creating a custom Drupal theme

## Release/Deployment notes:

- No, this is only an improved installation docs.

### DIFF
--- a/docs/users/extend/additional-services.md
+++ b/docs/users/extend/additional-services.md
@@ -70,6 +70,14 @@ This recipe adds a [Beanstalk](https://beanstalkd.github.io/) container to a pro
 - The Beanstalk instance will listen on TCP port 11300 (the beanstalkd default).
 - Configure your application to access Beanstalk on the host:port `beanstalk:11300`.
 
+## Node
+This recipe adds a [Node](https://github.com/nodejs/docker-node) container to a project.
+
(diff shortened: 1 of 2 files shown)


rfay:
You're awesome! Thanks so much for pursuing this and sharing it.

Could you please do a PR in https://github.com/drud/ddev-contrib instead? We try not to add every possible recipe to the ddev main docs because they move slower and are harder to change. Thanks!

Also, please sign the CLA, thanks.

rfay:
Thanks @boulaffasae sorry to ask for extra work, but would really appreciate you adding this with a full tutorial to https://github.com/drud/ddev-contrib

Thanks! Looking forward to it.

Issue #1925, opened 2019-11-07
Bundle ngrok with the Windows installer


gilbertsoft:
The ngrok.exe can be easy integrated to the Windows installer too and would make life of devs a little bit easier.

I already asked for the permission to do so, if we don't get an (positive) answer we can simply integrate it as a automatic download similar to Docker Desktop.

gilbertsoft:
I have not received answer from ngrok. How will we proceed? I suggest to inculde the binary to the installer as own component.

rfay:
Let's keep waiting a bit. I pinged them also just now on their "contact us" form on their web page. 

It's quite possible that the ngrok in Chocolatey (2.0) is unmaintained, as their current is 2.3.35, so that would argue even more for bundling it. 

Interestingly... https://chocolatey.org/packages/ngrok : It shows in the license link a link to ngrok v1's Apache license, clearly inappropriate.  Shows software source as https://chocolatey.org/packages/ngrok, not true. 

This is kind of too bad.

rfay:
Let's go ahead and bundle. And probably remove from the chocolatey recipe, since that one isn't kept up-to-date. If they ever get back to us and object, we'll happily remove it. They must be living life large on a beach or something.

rfay:
Finally got a response!


> From: Alan Shreve <support@ngrok.com> 
> hi randy -
> sorry about the delayed response. yep, you bet, please bundle away
> - alan
> Founder, ngrok.com

Link: https://mail.google.com/mail/u/2/#inbox/FMfcgxwGCQbMZZnvcqrBpXfzgswVxgjz

rfay:
ngrok approved us bundling; And we include ngrok as a dependency in chocolatey.

rfay:
I don't think we need to do this since people generally install with chocolatey.

PR #1935, opened 2019-11-11
Ngrok, nssm, mkcert should be a dependency of ddev, why not?

rfay:
## The Problem/Issue/Bug:

We've been telling people complicated things to install ngrok. We *bundle* nssm and mkcert in the installer, but they should probably also be dependencies.  

Note that #1925 also proposes to add ngrok to the installer, which is appropriate.

Make those 3 things dependencies for ngrok.

This will have to be tested with a real pre-release in choco.

## Related Issue Link(s):

#1925 proposes to bundle ngrok with the installer, which I still think is fine. 

## Release/Deployment notes:

I'm concerned about the lack of version requirements. My assumption is that chocolatey upgrades to current when you upgrade ddev.

### DIFF
--- a/winpkg/chocolatey/ddev.nuspec
+++ b/winpkg/chocolatey/ddev.nuspec
@@ -20,9 +20,15 @@
     <summary>ddev is a local web development tool optimized for PHP projects and CMSs.</summary>
     <description>ddev allows developers to use both nginx and apache, and many versions of PHP, with no host configurations. It works for most web apps, but has a bias toward TYPO3 and Drupal and Wordpress.</description>
     <releaseNotes>See https://github.com/drud/ddev/releases/tag/vREPLACE_DDEV_VERSION</releaseNotes>
+  <dependencies>
+      <dependency id="ngrok" />
+      <dependency id="mkcert" />
+      <dependency id="nssm" />
+  </dependencies>
+
   </metadata>
   <files>
     <file src="tools/**" target="tools" />
-
   </files>
+
 </package>


winpkg/chocolatey/ddev.nuspec
```diff
@@ -20,9 +20,15 @@
     <summary>ddev is a local web development tool optimized for PHP projects and CMSs.</summary>
     <description>ddev allows developers to use both nginx and apache, and many versions of PHP, with no host configurations. It works for most web apps, but has a bias toward TYPO3 and Drupal and Wordpress.</description>
     <releaseNotes>See https://github.com/drud/ddev/releases/tag/vREPLACE_DDEV_VERSION</releaseNotes>
+  <dependencies>
```
gilbertsoft: If you install this 3 tools by choco dependencies and later two of them are installed by the installer too we could finally end in a installation of different versions of this tools. If doing so we should prevent the installation of them in the silent mode which perhaps is not a bad idea at all. I will come up with an PR for the Windows installer...

rfay: Complicated! But it does seem that if they're using choco, then the choco dependency should be preferred. Of course Chocolatey discourages bundling things as opposed to making them dependencies, but I don't think we have any choice since we certainly don't expect that everybody uses choco.

rfay:
> You should wait with the merge until the Windows installer is updated and merged to avoid conflicts

I'm fine with that... I assume you mean you want to make a PR that will detect whether these already exist? There's no innate merge conflict here, as this is chocolatey-land. And of course you know that your main PR is merged and released.

winpkg/chocolatey/ddev.nuspec
gilbertsoft: What about adding sudo too here?

rfay: I looked at it and am open to it. But it looks to me like Chocolatey's sudo is unrelated to the one we've always used.  https://chocolatey.org/packages/Sudo - I wasn't quite able to track its parentage.

gilbertsoft: I say go without it. We always can add it later on...

gilbertsoft:
> And of course you know that your main PR is merged and released.
Yes, used it before to upgrade again after the downgrade

PR #1805, opened 2019-08-20
Temporarily use a custom-built mkcert that works on Ubuntu 16.04 for CircleCI

rfay:
## The Problem/Issue/Bug:

mkcert v1.4.0, recently released, is broken on Ubuntu 16.04, see PR https://github.com/FiloSottile/mkcert/pull/193 and issue https://github.com/FiloSottile/mkcert/pull/192

We use Ubuntu 16.04 for our CircleCI linux builds. 

## How this PR Solves The Problem:

The PR https://github.com/FiloSottile/mkcert/pull/193 solves this. Until it or something related comes through, we'll have to dodge with a private build (from https://github.com/rfay/mkcert/releases/tag/v1.4.1-alpha1)

## Manual Testing Instructions:

See if CircleCI builds

PR #1936, opened 2019-11-11
Revert "Temporarily use a custom-built mkcert that works on Ubuntu 16.04"

rfay:
## The Problem/Issue/Bug:

In #1805 we had to go to a custom-made mkcert because v1.4 did not work right on Ubuntu 16.04 (which is what runs our CircleCI tests)

This reverts commit 95ab5532f1d35246a6d1433cfb7bf46135012a65.

Note that mkcert v1.4.1 has been released, which works OK with Ubuntu 16.04

This should be OK if CircleCI tests pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,12 +11,12 @@ jobs:
     - run: sudo mkdir /home/linuxbrew && sudo chown $(id -u) /home/linuxbrew
     - restore_cache:
         keys:
-        - homebrew-linux-v7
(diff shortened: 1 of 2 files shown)


rfay:
This is probably going to pass testing for the wrong reasons. linuxbrew's mkcert v1.4.1 is not yet available, for unknown reasons.  

But to do a valid test on this we'll have to bump the cache so it doesn't use it.

PR #1937, opened 2019-11-11
Skip tools installation in chocolatey mode (Windows installer)

gilbertsoft:
## The Problem/Issue/Bug:
Installing DDEV with Chocolaty should prefer the tools installed as dependencies instead of the ones bundled with the DDEV Windows installer
Skips mkcert and nssm from installation in the new choco mode /S /C

Run the installer with options /S /C and check the tools are not installed. If they are already present they will be kept and prefered by DDEV
## Related Issue Link(s):
#1935

### DIFF
--- a/winpkg/chocolatey/tools/chocolateyinstall.ps1
+++ b/winpkg/chocolatey/tools/chocolateyinstall.ps1
@@ -15,7 +15,7 @@ $packageArgs = @{
   checksumType64= 'sha256'
 
   validExitCodes= @(0, 3010, 1641)
-  silentArgs   = '/S'
+  silentArgs   = '/S /C'
 }
 
 Install-ChocolateyPackage @packageArgs
--- a/winpkg/ddev.nsi
+++ b/winpkg/ddev.nsi
@@ -454,29 +454,32 @@ SectionGroup /e "mkcert"
    * mkcert application install
    */
   Section "mkcert" SecMkcert
-    SectionIn 1 2
-    SetOutPath "$INSTDIR"
-    SetOverwrite try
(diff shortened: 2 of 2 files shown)


rfay:
It seems that this makes the assumption "silent mode" == "choco install". That might not be correct.

I hate to add too much complexity. Wonder if there's a way to detect choco install.

It's always possible that we should have a separate chocolatey installer :(

gilbertsoft:
> 
> 
> It seems that this makes the assumption "silent mode" == "choco install". That might not be correct.
> 
> I hate to add too much complexity. Wonder if there's a way to detect choco install.
> 
> It's always possible that we should have a separate chocolatey installer :(

You're absolutely right but I don't think one will install DDEV in silent mode. So we can assume silent mode = choco install

gilbertsoft:
> As far as I know we can not check for an additonal cmd line parameter but will have a look. But I don't think we have take care

There is a way to get parameters and options from the command line. So I will add a choco parameter to the next commit...

gilbertsoft:
We have now a special Choco mode (see commit) enabled with the parameters /S /C

rfay:
Let's add this to the chocolatey build, please add to silentArgs, see https://github.com/drud/ddev/blob/master/winpkg/chocolatey/tools/chocolateyinstall.ps1

gilbertsoft:
> 
> 
> Let's add this to the chocolatey build, please add to silentArgs, see https://github.com/drud/ddev/blob/master/winpkg/chocolatey/tools/chocolateyinstall.ps1

That's already done with the related commit.

Currently I don't know why the test fails. Did a restart two times...

PR #1944, opened 2019-11-13
Change silent uninstaller to not showing messages

gilbertsoft:
## The Problem/Issue/Bug:
Message boxes are shown during the uninstall.
This patch changes the behavior of the uninstaller to not to show message boxes during the process. The only exception is the message if the install directory could not be removed.
Run the uninstaller with the parameter /S or with chocolatey uninstall


gilbertsoft:
I tried it here on my other dev system and it worked fine too!

Perhaps you have an issue with cmder. Try to launch it by the Windows run dialog (press Win+R or right click to Windows/Run)

PR #1948, opened 2019-11-14
[FEATURE] Allow an optional path for the launch command

gilbertsoft:
## The Problem/Issue/Bug:
Sometime the user likes to open a specific path of the website instead of the entry point.

This patch allows the user to provide an additional path for the launch command to directly head to the defined path of the website instead of the root entry point.

1. Run `ddev launch mypath` to see that the browser opens the correct URL. Replace `mypath` by the path you like to open.
2. Run `ddev launch` to see that the browser opens the root entry point of the website as before.


cmd/ddev/cmd/dotddev_assets/commands/host/launch
```diff
@@ -1,17 +1,17 @@
 #!/bin/bash
 
 ## Description: Launch a browser with the current site
-## Usage: launch
-## Example: "ddev launch"
+## Usage: launch [path]
+## Example: "ddev launch" or "ddev launch mypath"
```
rfay: Wouldn't this be better with `ddev launch /mypath` ?

cmd/ddev/cmd/dotddev_assets/commands/host/launch
```diff
@@ -1,17 +1,17 @@
...
 
 case $OSTYPE in
   linux-gnu)
-    xdg-open ${DDEV_PRIMARY_URL}
+    xdg-open ${DDEV_PRIMARY_URL}/$1
```
rfay: I think each of these would be better and simpler if we construct a single env var (FULLURL?) that includes the $1. Something like this at the top:
```
FULLURL=${DDEV_PRIMARY_URL}
if [ ! -z ${1:-} ] ; then FULLURL="${DDEV_PRIMARY_URL}/${1}"; fi
```
Then use FULLURL in these places. This avoids extra "/" in url in some cases as well

cmd/ddev/cmd/dotddev_assets/commands/host/launch
gilbertsoft: I think we should allow both variants to avoid errors

cmd/ddev/cmd/dotddev_assets/commands/host/launch
gilbertsoft: Change is made, but the variant with the leading slash does not run here. The URL `https://typo3-8-mysql-8-test.ddev.site/C:/Program` is opened. I guess a kind of excaping issue but I don't know how to solve and if it's more a problem of the call or in the script

gilbertsoft: Using two slahes works fine in git-bash on Windows

PR #1975, opened 2019-11-25
Tests only: Get mysql client working right in High Sierra

rfay:
## The Problem/Issue/Bug:

#1973 rolled the CircleCI macOS install back to High Sierra. 

Although that one passed the macOS test, when merged there were still problems with the mysql command on some at least one build. 

This attempts to use the mariadb package instead of mysql-client. We'll see how it goes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,14 +73,14 @@ jobs:
         at: ~/
     - restore_cache:
         keys:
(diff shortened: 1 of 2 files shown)


rfay:
CircleCI seems to have solved the underlying problem, making this unnecessary, thankfully.

PR #1976, opened 2019-11-25
Remove file_scan_ignore_directories

webflo:
## The Problem/Issue/Bug:

The setting for file_scan_ignore_directories is Drupal default settings.php. There is no need to override it in ddev.

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -149,11 +149,6 @@ ini_set('session.cookie_lifetime', 2000000);
 
 $settings['hash_salt'] = '{{ $config.HashSalt }}';
 
-$settings['file_scan_ignore_directories'] = [
-  'node_modules',
-  'bower_components',
-];
-
 // This will prevent Drupal from setting read-only permissions on sites/default.
 $settings['skip_permissions_hardening'] = TRUE;
 


webflo:
I configured an existing project with ```ddev config```. The ```file_scan_ignore_directories``` was configured in ```settings.php``` with an additional folder. It was overwritten by ```settings.ddev.php```.

rfay:
I guess that stanza is directly from default.setting.php. It could certainly be left out here, but as noted above, many projects don't have a settings.php, so ddev generates one for them. Using this same template. 

I'm sure you're already aware that there are a couple of easy ways for you to deal with this for your project:
* Take over the settings.ddev.php by just removing the #ddev-generated line and making it your own.
* Make the change to this section *after* the include of settings.ddev.php

If it's reasonable to have a default file_scan_ignore_directories, then we should probably leave it in there, because it serves many people as it is.

webflo:
> Take over the settings.ddev.php by just removing the #ddev-generated line and making it your own.

Thanks for the hint. I didn't know that.

rfay:
I think this is the correct thing to do. The stuff shown as defaults is just an example config, and we have no business adding it to settings.ddev.php

PR #1980, opened 2019-11-26
[DOCS] Streamline markdown

gilbertsoft:
Following rules were applied:
* write **DDEV** uppercase and bold
* max. line length of 80 chars
* use `*` for lists
* use `**` for bold text
* use `*` for italic text
* use language tags for code highlighting in code blocks

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -12,10 +12,11 @@ A clear and concise description of what the bug is.
 
 **To Reproduce**
 Steps to reproduce the behavior:
+
 1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
-4. See error
+1. Click on '....'
+1. Scroll down to '....'
+1. See error
 
 **Expected behavior**
 A clear and concise description of what you expected to happen.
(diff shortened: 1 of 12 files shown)


gilbertsoft:
@rfay first PR which tries to streamline the docs, have a look at the changes list above. The content has not changed see https://github.com/drud/ddev/pull/1980/files?short_path=1a523bd#diff-1a523bd9fa0dbf998008b37579210e12

rfay:
Thanks for this.

One thing that would be lovely to add is to make sure there's a newline before bulleted lists. readthedocs is very picky about that and makes a mess if there isn't one.

Would you mind fixing here the formatting problem in https://ddev.readthedocs.io/en/latest/users/cli-usage/#wordpress-quickstart (list of env variables is immediately inline followed by "for more details" on same line. It looks OK on https://github.com/drud/ddev/blob/master/docs/users/cli-usage.md

gilbertsoft:
> 
> 
> Thanks for this.
> 
> One thing that would be lovely to add is to make sure there's a newline before bulleted lists. readthedocs is very picky about that and makes a mess if there isn't one.
> 
> Would you mind fixing here the formatting problem in https://ddev.readthedocs.io/en/latest/users/cli-usage/#wordpress-quickstart (list of env variables is immediately inline followed by "for more details" on same line. It looks OK on https://github.com/drud/ddev/blob/master/docs/users/cli-usage.md

Yeah, will do. Just started with the first file so far to get your input...

rfay:
Could we close this one or else update it based on the new .markdownlintrc. I'd be happy to redo also, and the new .markdownlintrc makes it easy to do. Can also add a pre-push hook for it.

rfay:
Please take a look at #2013 and see if it does everything you wanted here. It uses the .markdownlintrc we introduced, but it might not have caught everything you wanted in this PR.

rfay:
Thanks to you @gilbertsoft we've made significant strides in checking markdown now, with a .markdownlintrc. The rules we ended up with aren't exactly the same as applied in this PR, but the replacement PR, https://github.com/drud/ddev/pull/2013, does comply with the agreed rules, and I'm hoping you can review that. #2013 also starts enforcing markdown formatting in the staticrequired rule and  circleci test.

I'm hoping there's nothing here that you don't find dealt with in #2013, closing for now in favor of #2013

PR #1981, opened 2019-11-26
Fix enable_xdebug / disable_xdebug not fully working for apache-cgi

carstendietrich:
## The Problem/Issue/Bug:
Due to the nature of apache-cgi we need an config reload to make php config changes work. Therefore running `ddev exec enable_xdebug` with `apache-cgi` as webserver turns on Xdebug only for the CLI and not for normal webserver requests.

This adds an apache2 reload to the enable_xdebug and disable_xdebug bash scripts in case the current webserver is apache-cgi.

Use the containertest.sh which now checks if both the CLI and the webserver have Xdebug enabled/disabled after using the commands.

### DIFF
--- a/containers/ddev-webserver/files/usr/local/bin/disable_xdebug
+++ b/containers/ddev-webserver/files/usr/local/bin/disable_xdebug
@@ -2,4 +2,7 @@
 export PATH=$PATH:/usr/sbin:/sbin
 phpdismod xdebug
 killall -HUP php-fpm 2>/dev/null || true
+if [[ $DDEV_WEBSERVER_TYPE == "apache-cgi" ]]; then
+  /etc/init.d/apache2 reload >/dev/null || true
+fi
 echo "Disabled xdebug"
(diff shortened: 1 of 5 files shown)


rfay:
Wow, so impressive, you even went after the tests!  For this to work though we'll need to push a separate image. 

Could you say a little about why you use apache-cgi? I didn't actually think anybody at all used that.

rfay:
I pushed an image for you. Please push another commit that thanges the WebTag to `20191126_carstendietrich_enable_xdebug` in pkg/version/version.go at https://github.com/drud/ddev/blob/f99c8db7f7b484739714e3ccfa892ae6246fce8c/pkg/version/version.go#L52

carstendietrich:
Hey @rfay,

thanks for the super fast response :) I updated the WebTag accordingly.

Why we use apache-cgi is a good question. We switched to DDEV for our Magento 2 project. For a simple kickstart we chose Apache, so we could rely on the existing .htaccess files instead of adding a custom nginx configuration. With apache-fpm we encountered several problems and therefore tried apache-cgi, which worked for us immediately.

rfay:
I'm surprised. I'd encourage you to just try apache-fpm quickly one day (and with NFS if you're on macOS or Windows) and think you'll find a major speed boost. BTW, explicit magento support is on the way...

PR #1986, opened 2019-11-28
Better example on drupal7/8 config

AronNovak:
## The Problem/Issue/Bug:

The issue is that the documentation suggests that it's the best practice to reinstall the site at every `ddev start`, but it's not so efficient, basically it loses all the benefits of database persistence. After stopping a container, using the example, you cannot continue using your existing site.

It makes the installation step conditional.

Apply the config, start it, stop it, start it, let's make sure that the 2nd start is much faster, as no installation is involved, but the site is still usable.

Not applicable, only documentation.

## Release/Deployment notes:
Not applicable, only documentation.

### DIFF
--- a/docs/users/extending-commands.md
+++ b/docs/users/extending-commands.md
@@ -123,8 +123,8 @@ hooks:
 ```
 hooks:
   post-start:
-    # Install Drupal after start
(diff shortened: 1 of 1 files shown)


AronNovak:
image
And the outcome of my local test

PR #1993, opened 2019-12-04
Use drupal_recommended template now that Drupal 8.8 is here (docs only), fixes #1920

### LINKED ISSUE #1920: [docs] demonstrate drupal/recommended-project for Drupal install instead of drupal-project
**Describe the bug**

Per https://www.drupal.org/docs/develop/using-composer/starting-a-site-using-drupal-composer-project-templates from Drupal 8.8.x forward, there is a composer template in Drupal core, and using drupal-project is no longer the recommended technique

So we'll want to change the D8/D9 quickstart docs to use that instead of drupal-project.

Currently we have
`ddev composer create drupal-composer/drupal-project:8.x-dev --stability dev --no-interaction --prefer-dist`

And we'll want to change to something like:

`ddev composer create drupal/recommended-project:^8.8--stability dev --prefer-dist`

And probably also add D9 version...

rfay:
## The Problem/Issue/Bug:

* Drupal 8.8 now has its own composer template, drupal/recommended-project, so we no longer need to use the traditional template.
* I updated the bedrock/WordPress and TYPO3 quickstarts to not require switching the project type.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -17,7 +17,7 @@ Each of these commands has full help. For example, `ddev start -h` or `ddev help
 * `ddev launch` or `ddev launch some/uri` will launch a browser with the current project's URL (or a full URL to `/some/uri`)
 * `ddev delete` is the same as `ddev stop --remove-data` and will delete a project's database and ddev's record of the project's existence. It doesn't touch your project or code. `ddev delete -O` will omit the snapshot creation step that would otherwise take place, and `ddev delete images` gets rid of spare Docker images you may have on your machine.
 
-## Bundled Tools List
+## Partial Bundled Tools List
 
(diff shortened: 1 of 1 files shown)


rfay:
This needs more review:
* The recommended template is pretty poor on Windows, at least with my one try. I don't know why it would be so bad.
* Drush isn't included in this template. I have no idea why they'd leave it out.

rfay:
I find a number of things about recommended-template that aren't working as well as the old way, so think we should hold off on this:
* On Windows, unzipping fails and the whole thing often fails. This can be mitigated by using NFS.
* Actual behavior (especially on Windows) seems broken
* drush is not included by default, although we can require it in the recipe.

The recipe here also doesn't match the current docs. Now they're saying just `composer create-project drupal/recommended-template`

Closing this until recommended template is better.

PR #1994, opened 2019-12-06
New location to define sync directory

shaal:
## The Problem/Issue/Bug:

Drupal 8.8 changed where the sync directory is defined
(https://www.drupal.org/node/3018145)

Changing the code to use $settings rather than $config_directories.

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -161,6 +161,12 @@ $settings['trusted_host_patterns'] = ['.*'];
 $settings['class_loader_auto_detect'] = FALSE;
 
 // This specifies the default configuration sync directory.
+// Both $config_directories (pre-Drupal 8.8) and
+// $settings['config_sync_directory'] are provided
+// so it should work on any Drupal 8 or 9 version.
+if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
+  $config_directories[CONFIG_SYNC_DIRECTORY] = '{{ joinPath $config.SitePath $config.SyncDir }}';
+}
 if (empty($settings['config_sync_directory'])) {
   $settings['config_sync_directory'] = '{{ joinPath $config.SitePath $config.SyncDir }}';
 }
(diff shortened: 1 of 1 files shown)


rfay:
Thanks - but how do we handle multiple versions of Drupal 8? Is there a reasonable way to detect 8.8?

shaal:
@rfay we can get the Drupal core version from `core/lib/Drupal.php`:
https://github.com/drupal/drupal/blob/8.8.x/core/lib/Drupal.php#L85

```
 /**
   * The current system version.
   */
  const VERSION = '8.8.1-dev';
```

rfay:
If... you parse PHP... And if nobody changes version with a composer upgrade...

shaal:
The way I understand it, `composer upgrade` will keep the correct drupal
core version when it gets upgraded.

rfay:
But composer wouldn't know to upgrade settings.ddev.php. Seems like a hard problem.

OH... Why don't we put BOTH versions in there? It would do no harm right?

pkg/ddevapp/drupal.go
```diff
@@ -166,8 +166,8 @@ $settings['trusted_host_patterns'] = ['.*'];
 $settings['class_loader_auto_detect'] = FALSE;
 
 // This specifies the default configuration sync directory.
-if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
-  $config_directories[CONFIG_SYNC_DIRECTORY] = '{{ joinPath $config.SitePath $config.SyncDir }}';
+if (empty($settings['config_sync_directory'])) {
+  $settings['config_sync_directory'] = '{{ joinPath $config.SitePath $config.SyncDir }}';
```
rfay: Let's keep both of these here. It should work on most any version of Drupal 8 then true?

Azovsky: Right, this should working.

rfay:
@shaal I pushed a commit that sets *both* variables (and rebased). If you're OK with this (and everybody else is) let's get it in. Thanks for this!

rfay:
The way I did this wasn't quite right, as php complains about undefined CONFIG_SYNC_DIRECTORY in D9, etc. I hope all interested can take a look at https://github.com/drud/ddev/pull/2039 - I think it will be OK.

PR #1998, opened 2019-12-10
Create .ddev/.gitignore on ddev start *and* config (for teams), fixes #1997

### LINKED ISSUE #1997: `ddev start` has side-effects
**Describe the bug**
since .ddev/.gitignore contains itself now and gets created at `ddev config` but not on `ddev start`, the created files after a `ddev start` remain unignored.

**To Reproduce**
Check out any repository with a `.ddev` configuration that was created with v1.12.
Run `ddev start`.

**Expected behavior**
`git status` should not show any changed files.

ddev v1.12.0

jonaseberle:
fixes  #1997

## The Problem/Issue/Bug:
`ddev start` produces files that make a working copy unclean.

It unignores .ddev/.gitignore - thus it encourages users to check this file in. 

This re-enables the old behaviour before v1.12.

* `git clone` a repository with .ddev configuration that was created by `ddev config` (this patch changes the behaviour of ddev config)
* `ddev start` 
* `git status`

## Related Issue Link(s):
https://github.com/drud/ddev/issues/1997


rfay:
Thanks for this. I'm going to study the context and I imagine based on your intense interest it will get in :) I liked having it totally the responsibility of ddev.

What did you think about the option of creating it on `ddev start` as opposed just on config? That's what currently happens to settings files like AdditionalSettings.php.

jonaseberledmind:
> What did you think about the option of creating it on `ddev start` as opposed just on config? That's what currently happens to settings files like AdditionalSettings.php.

That would solve the problem, too. Thank you for looking into it.

rfay:
I do note the inclusion of self-ignore for the gitignore being added in 70a8b0d1e5bc382307fa6552c0e4187fbf83cc48 (v1.12) apparently as an afterthought so you were right, as you knew all along :)

gilbertsoft:
Imho because the .gitignore is managed by DDEV it's fine to just ignore it. If someone wants to add it to the VCS he can still do it. But on the other hand with the current solution I don't have to commit changes in this file.

Recreating it on `ddev start` or `ddev config` is imho the right solution.

gilbertsoft:
Imho because the .gitignore is managed by DDEV it's fine to just ignore it. If someone wants to add it to the VCS he can still do it. But on the other hand with the current solution I don't have to commit changes in this file.

Recreating it on `ddev start` or `ddev config` is imho the right solution.

rfay:
@jonaseberle (and all) I took the liberty of changing the approach to creating the gitignore on `ddev start` as well as `ddev config`, would appreciate your review and testing. I'll post the link to artifacts when the build completes.

rfay:
Artifacts are at https://circleci.com/gh/drud/ddev/21163#artifacts/containers/0 - thanks for testing and reviewing @jonaseberle

rfay:
@jonaseberle @jonaseberledmind  @gilbertsoft hoping you can review this and try out the artifacts. If we're going to get it the way you like it... we'll have to make sure we get it the way you like it :)

jonaseberle:
Hi @rfay,

I tested with 
v1.12.0-3-g7fd9f02f

`.gitignore` is created as expected on `ddev start`. A cloned git repository is thus clean after starting ddev. 

Thank you, I am very happy with that.

rfay:
Thanks so much. I also created https://github.com/drud/ddev/issues/2024 with the idea of adding an option for ddev not to touch settings files and the like. I guess that wouldn't affect the .ddev directory which is at issue here. But it's certainly a related topic.

PR #2003, opened 2019-12-16
Add .markdownlintrc defining markdown, especially for readthedocs

rfay:
## The Problem/Issue/Bug:

@gilbertsoft pointed out github.com/DavidAnson/markdownlint and https://github.com/igorshubovych/markdownlint-cli - these can be used to specify consistent markdown in our docs. Since readthedocs/mkdocs is particularly susceptible to markdown style, this will be great for fixing up our markdown in the docs.

Markdown will either be updated later bit by bit or with a mass update like https://github.com/drud/ddev/pull/1980, or perhaps that one will be updated to meet the requirements of this markdownlintrc

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,0 +1,28 @@
+{
+
+  // These rules mostly come from mkdocs rules
+  // https://github.com/mkdocs/mkdocs/blob/master/.markdownlintrc
+
+  // Enable all markdownlint rules
+  "default": true,
+
+  // Fine with long lines, editor can wrap
+  "line-length": false,
+
+  "MD004": { "style": "consistent" },
+
+  // Set Ordered list item prefix to "ordered" (use 1. 2. 3. not 1. 1. 1.)
+  "MD029": { "style": "ordered" },
+
+  // Set list indent level to 4 which Python-Markdown requires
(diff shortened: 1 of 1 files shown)


.markdownlintrc
```diff
@@ -0,0 +1,28 @@
...
+  "MD004": { "style": "consistent" },
+
+  // Set Ordered list item prefix to "ordered" (use 1. 2. 3. not 1. 1. 1.)
+  "MD029": { "style": "ordered" },
+
+  // Set list indent level to 4 which Python-Markdown requires
+  "MD007": { "indent": 4 },
+
+  // Embedded html allowed; often used for named anchors with mkdocs/readthedocs
+  "no-inline-html": false,
```
gilbertsoft: Not sure about this, normally you would prefer headings.

rfay: The reason for this is essentially to create permalinks in readthedocs that aren't dependent on the actual verbiage in a title. For example, "unable-listen" in https://raw.githubusercontent.com/drud/ddev/master/docs/users/troubleshooting.md

If you know a better way, I'd sure prefer one.

gilbertsoft: I guess there isn't an other way. But this was more a question, not a blocker btw

PR #2006, opened 2019-12-17
Add utf8mb4 client collation to my.cnf in containers

rfay:
## The Problem/Issue/Bug:

mysql import-db and mysql are affected by client config which lacks utf8mb4 (defaults to latin1). 

Add client config to set connection to utf8mb4

Example is coming from @RobertLang, who ran into trouble with imported dbs because of this.

We need to add a test to the mysql container to make sure that everything works properly, should be easy enough.

### DIFF
--- a/containers/ddev-dbserver/files/etc/my.cnf
+++ b/containers/ddev-dbserver/files/etc/my.cnf
@@ -2,6 +2,7 @@
 # CLIENT #
 port                           = 3306
 socket                         = /var/tmp/mysql.sock
+default-character-set = utf8mb4
 
 [mysqld]
 
--- a/containers/ddev-webserver/files/home/.my.cnf
+++ b/containers/ddev-webserver/files/home/.my.cnf
@@ -2,6 +2,7 @@
 user=db
 password=db
 host=db
+default-character-set = utf8mb4
 
 [mysql]
 database=db
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -49,13 +49,13 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "20190911_terminus_pantheon" // Note that this can be overridden by make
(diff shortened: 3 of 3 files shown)


RobertLang:
Please check this very basic project:
[dbtest.zip](https://github.com/drud/ddev/files/3985006/dbtest.zip)


Steps to reproduce the problem with MySQL 5.7 on macOS Catalina 10.15.1:

`ddev start`
`ddev import-db --src=correct_import.sql `
=> check phpMyAdmin 
image 
(result looks good)

`ddev import-db --src=error_import.sql `
=> check phpMyAdmin 
image
(result is scrambled)

When the sql file doesn't contain the line `/*!40101 SET NAMES utf8mb4 */;` the import is done using the default connection encoding which is latin1 for the ddev database container.

When checking the db variables with `ddev mysql` you see that the default character set for client and connection is latin1.
```
mysql> show variables like "%character%";
+--------------------------+----------------------------+
| Variable_name            | Value                      |
+--------------------------+----------------------------+
| character_set_client     | latin1                     |
| character_set_connection | latin1                     |
| character_set_database   | utf8mb4                    |
| character_set_filesystem | binary                     |
| character_set_results    | latin1                     |
| character_set_server     | utf8mb4                    |
| character_set_system     | utf8                       |
| character_sets_dir       | /usr/share/mysql/charsets/ |
+--------------------------+----------------------------+
```

I would suggest to make the default connection encoding fit with the default database encoding (which is utf8mb4) to fix this issue.

rfay:
Thanks for the excellent test. It broke for me with both imports, but with `dbimage: drud/ddev-dbserver-mysql-5.7:20191217_utf8mb4_connection` (from this PR) in .ddev/config.yaml, it worked fine for me. I'd love it if you'd confirm that behavior, but I think this does it.

I was just using `ddev mysql` and 

```
mysql> select * from test_table;
+----+---------------------------------------+
| id | testfield                             |
+----+---------------------------------------+
|  1 | Some german umlauts: äöü ß ÄÖÜ        |
|  2 | Some czech text:  Přímé doručení      |
+----+---------------------------------------+
2 rows in set (0.00 sec)
```

PR #2007, opened 2019-12-19
[TASK] Remove empty trailing lines from PHP files and templates

gilbertsoft:
This patch removes empty trailing lines from PHP files and configuration
templates to comply with PSR-12 which defines the following on file endings:

All PHP files MUST end with a non-blank line, terminated with a single LF.

See https://www.php-fig.org/psr/psr-12/#22-files


gilbertsoft:
Seems I messed up my master. But no issue for this PR. Will clean it afterwards...

gilbertsoft:
It's rebased now

pkg/ddevapp/typo3.go
```diff
@@ -29,28 +32,27 @@ $GLOBALS['TYPO3_CONF_VARS'] = array_replace_recursive(
...
-            'transport_smtp_server' => 'localhost:1025'
+            'transport_smtp_server' => 'localhost:1025',
         ],
         'SYS' => [
             'trustedHostsPattern' => '.*.*',
             'devIPmask' => '*',
-            'displayErrors' => 1
-        ]
+            'displayErrors' => 1,
+        ],
```
gilbertsoft: btw this trailing comma is recommended because it's less error prone if you add something new afterwards

rfay: Always. But of course they ruin js/json :)

PR #2010, opened 2019-12-22
Use recent phpmyadmin version (4.9)

thomaskieslich:
## The Problem/Issue/Bug:
the phpmyadmin Version two Years old

raise phpmyadmin Version


rfay:
I took the liberty of adding a couple of commits with a slightly different approach: Just use the upstream phpmyadmin, 4.9.  I think we built the drud-specific image a few years ago when there was a problem with the upstream version. 

This change means no healthcheck on phpmyadmin, but that's not that important I don't think.

PR #2011, opened 2019-12-23
Edit the default php version.

Ahmed-Ayman:
now whenever I start a new ddev project php is set by default to 7.2 not 7.1

### DIFF
--- a/docs/users/extend/customization-extendibility.md
+++ b/docs/users/extend/customization-extendibility.md
@@ -3,7 +3,7 @@ ddev provides several ways in which the environment for a project using ddev can
 
 ## Changing PHP version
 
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for catching this and doing a PR too! Much appreciated. I must have also caught it when working on #1982 (which changes to php 7.3 as default). See https://github.com/rfay/ddev/blob/d3c18845201dc04b7d8f71b49d591f5192619da0/docs/users/extend/customization-extendibility.md#L6 and see if it works for you. 

Again, thanks so much for flagging this, 7.2 has been the default for a really long time. Closing in favor of #1982

Ahmed-Ayman:
alright, thank you rfay :)

PR #2023, opened 2020-01-07
auth pantheon missing the dash

kreynen:
I was running through our setup documentation before on boarding a new dev. I was still running v1.10.2 and thought this may have changed in the current version, but after updating to v1.12.1 I was still running into an issue trying to configuration this using `ddev auth pantheon`. I'm not sure if this changed at some point and needs to be broken out by version, but `ddev auth-pantheon <YOUR TOKEN>` is required now.

### DIFF
--- a/docs/users/providers/pantheon.md
+++ b/docs/users/providers/pantheon.md
@@ -12,7 +12,7 @@ If you have ddev installed, and have an active Pantheon account with an active W
 
     a. Login to your Pantheon Dashboard, and [Generate a Machine Token](https://pantheon.io/docs/machine-tokens/) for ddev to use.
 
-    b. Run `ddev auth pantheon <YOUR TOKEN>` (This is a one-time operation, and configures ddev to work with all the sites on your account.)
+    b. Run `ddev auth-pantheon <YOUR TOKEN>` (This is a one-time operation, and configures ddev to work with all the sites on your account.)
 
 2. Choose a Pantheon site and environment you want to use with ddev.
 


rfay:
Sorry, you were looking at the ["latest" docs](https://ddev.readthedocs.io/en/latest/) instead of the ["stable" docs](https://ddev.readthedocs.io/en/stable/). In the next release (per "latest" docs), it will in fact be `ddev auth pantheon`. I'd love to have you give it a spin if you're interested. https://github.com/drud/ddev/pull/1827

Sorry to waste your time on docs fix, but your effort and careful attention is appreciated.

I don't actually know how to prevent this type of confusion. Google indexes both of course. If you just go to ddev.readthedocs.io it gets the stable. Open to suggestions.

kreynen:
Sorry. The PR isn't valid, but there is still a problem... or room for improvement.

I was just being lazy and the latest page of documentation is what https://www.google.com/search?q=ddev+pantheon returned. Didn't notice that I was on latest vs. stable... but I'm guessing I'm not the only lazy dev who is going to run into this. It makes sense to use auth for multiple hosts vs. host specific commands, but you might want to note that this changed in the latest docs and/or do more in the UX (beyond the URL and floating version switch) to let people know where they've landed on the latest version. 

Please don't hate me for pointing this out here vs. opening another issue, but when a user lands on the latest, the floating version switch doesn't work to switch to stable.  Results in an unthemed 404. 

Screen Shot 2020-01-07 at 12 54 13 PM

rfay:
I love to hate readthedocs. After 3 years working with it I'm hoping to come to terms with it. 

And it sure won't make me hate *you* for pointing out issues :)

I don't actually know how to solve this though.

PR #2025, opened 2020-01-08
Move drush.yml into /drush from sites/all/drush

rfay:
## The Problem/Issue/Bug:

Drupal 8 drush drush.yml has always been placed in <docroot>/sites/all/drush. However, community opinion seems to be that putting drush.yml in <projectroot>/drush. 

Both of these 
* Can break multisite in the same way by overriding the default uri
* Create a new drush.yml file that is annoying to people and they may have to take over. 

But this is a slightly better place, outside Drupal's docroot, etc.


heddn:
I think the drush folder in the root makes more sense. That seems to be a more standard approach.

PR #2040, opened 2020-01-21
Prevent inappropriate rewrite of global_config.yaml, fixes #1882, fixes #2014

### LINKED ISSUE #1882: instrumentation_opt_in not respected (mistaken overwrite of global_config.yaml)
**Describe the bug**

If you have a new version of ddev and you ddev start you get this prompt every time you start. Even if you say "n" every time. 

```
It looks like you have a new ddev release.
May we send anonymous ddev usage statistics and errors?
To know what we will see please take a look at
https://ddev.readthedocs.io/en/latest/users/cli-usage/#opt-in-usage-information
Permission to beam up? [Y/n] (yes):
```

**To Reproduce**

* Any environment
* Have more than one project going and listed in ~/.ddev/global_config.yaml
* Create a project of type "wordpress" (doesn't need any code, just type of wordpress)
* `ddev start`
* ~/.ddev/global_config.yaml will have been rewritten with default information.

**Expected behavior**

**Screenshots**

This is recreatable on Windows and macOS

**Additional context**

### LINKED ISSUE #2014: ddev launch: DDEV_PRIMARY_URL uses first hostname from alphabetically sorted list of hostnames
**Describe the bug**
The `ddev launch` command uses `$DDEV_PRIMARY_URL` to set the hostname to open in a browser. It seems that the list of hostnames is sorted alphabetically.

**Expected behavior**
I'd expect ddev to use the project name as DDEV_PRIMARY_URL.

**To Reproduce**
Create a project with this configuration:
```
APIVersion: v1.12.0
name: ekw
type: typo3
docroot: public
php_version: "7.3"
webserver_type: apache-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames:
- ckm
additional_fqdns: []
mariadb_version: "10.2"
nfs_mount_enabled: false
provider: default
use_dns_when_possible: true
timezone: Europe/Berlin
```

When you start ddev, you can already see the alphabetically sorted list of hostnames:
```
Successfully started ekw
Project can be reached at https://ckm.ddev.site https://ekw.ddev.site https://127.0.0.1:32772
```

If you hit `ddev launch`, https://ckm.ddev is opened instead of the expected https://ekw.ddev.site

**Additional context**
I've already discussed this issue with Randy on Slack:
https://typo3.slack.com/archives/C8TRNQ601/p1577731610020600

Seems that the hostnames are sorted here:
https://github.com/drud/ddev/blob/4d7cf50357e76654e37b6bd01549480047cd84e0/pkg/ddevapp/ddevapp.go#L1511
https://github.com/drud/ddev/blob/06754053ebecabe991fb25019eeb1a6fd8d85b81/pkg/ddevapp/config.go#L495

rfay:
## The Problem/Issue/Bug:

#1882 reports that the instrumentation opt-in gets overwritten somehow so people get prompted again and again. Or else it's the last-started-ddev-version that gets overwritten. But probably it's a result of ddev somehow writing the global config without having read it.

#2014 is randomly fixed here because I was in the neighborhood. It complains that the primary URL reported on ddev start is not the primary URL. Now it is.

* Look for places where we might somehow write without reading first. Fix them.
* Look for general logic flaws

* The easy one: `ddev start` with multiple hostnames, including as least one alphabetically lower than the primary. The primary should be listed first.
* The hard one: Figure out a way to recreate the problem in #1882 and then demonstrate that it's fixed here. 

Nothing new.


rfay:
It turns out that our use of init() to populate commands is a disaster, with undefined results if the init() does anything that affects global state. This pattern has been in there from the beginning and of course we resist changing things like this. Anyway, the bottom line is that globalconfig should never be written without being read first (or if doesn't exist, populate with reasonable defaults).

This basic pattern and its results could affect so many things so very badly. And the best answer is to put the most important stuff in a.go? Yuck :)

PR #2041, opened 2020-01-21
v1.12.2: branch 20200106 notarize v1.12.1 (dummy for testing only, do not pull)

rfay:
## The Problem/Issue/Bug:

This is a dummy PR generated from the branch that will be used for v1.12.2

It is created only to make sure we've done a full set of tests, since working on that branch is not a very clean thing to do.

Next time: Create a PR against the branch!


rfay:
Tests all passed, release done. This is golden.

PR #2073, opened 2020-02-06
Expose DDEV version in containers

gilbertsoft:
## The Problem/Issue/Bug:
The currently running version of DDEV is unknown in the containers but this is a useful and important information if you relay on a specific version.

This PR exposes the DDEV version as environment variable.

- ddev ssh
- export

Currently no test is added.


rfay:
I'd be interested to know what caused you to want the ddev version inside the container.

FYI there's an interesting tidbit in a file in the root of the container that explains what the *container* was built from, which is different from what you're seeking here. 

The file drud_ddev-webserver_VERSION_INFO.txt has the nominal tag and the commit the container was built from
```
rfay@d7git-web:/$ cat drud_ddev-webserver_VERSION_INFO.txt
drud/ddev-webserver:v1.13.0 commit=v1.13.0-rc2-1-g63f667ea
```

pkg/ddevapp/ddevapp.go
```diff
@@ -1120,6 +1120,7 @@ func (app *DdevApp) DockerEnv() {
 		"DDEV_ROUTER_HTTPS_PORT":        app.RouterHTTPSPort,
 		"DDEV_XDEBUG_ENABLED":           strconv.FormatBool(app.XdebugEnabled),
 		"DDEV_PRIMARY_URL":              app.GetPrimaryURL(),
+		"DDEV_VERSION":                  version.GetVersionInfo(),
```
rfay: GetVersionInfo() returns a map of lots of stuff, it's not what you want. You just want version.Version.

You can compile things and test them by just using `make` and then running the artifact in .gotmp/bin/windows_amd64/ddev.exe

I *think* that this also has to be added in template.go to the docker-compose template, or the env var won't make it inside the container.

gilbertsoft:
> 
> 
> I'd be interested to know what caused you to want the ddev version inside the container.
> 

I like to check the DDEV version by composer or to be more precise checking if a DDEV feature is available or not. There are so many users who don't regulary update their DDEV and this seems to me a possible solution to do this. To get this also running inside the container with `ddev composer` I need a version hint.
If you have a better idea I'm happy to know. And yes, there is a hint on `ddev start` if the APIVersion does not match but that does not block the execution. An other way could be adding a min version to the config.yaml which also would solve some issues.
But I really like to have a solution for composer and I think this will be a nice feature. I'm also investigating how to implement this exactly to composer. My idea is finally to have a requirement for DDEV like this:

```
    "require": {
        "ddev": "^1.12.0"
    }
```

rfay:
Thanks, super interesting. But then that composer recipe can't function outside ddev right? 

There are so many things I don't know about composer!

gilbertsoft:
Outside the container I'll check the version by calling `ddev version`, that's the idea. But personally I normally never run composer outside the container if possible. It's seems also I need a composer-plugin to get rid of this. So finally one has to require this plugin too to be able to fetch the DDEV version inside or outside the container. That's what I'm currently looking for... I'm not sure if I really get this running but so far it looks good. I already did this PR the get my changes built and tested and start a discussion on this with you ;)

pkg/ddevapp/ddevapp.go
gilbertsoft: Thanks for the hint, fixed now

rfay:
You'll want `ddev --version` rather than `ddev version` outside.

gilbertsoft:
Will try to use drud_ddev-webserver_VERSION_INFO.txt for the detection, so I close this PR for the moment.

PR #2077, opened 2020-02-09
Fix Magento 2 quickstart default url

rafal-kos:
## The Problem/Issue/Bug:
Install clean 1.13 ddev. 
Install Magento 2 using the quickstart tutorial.
Accessing https://magento2.ddev.site/ gives an error page with message "There is no back-end webserver at the url you specified".
Use correct url in quickstart

## Release/Deployment notes:
No

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -272,12 +272,12 @@ Note that Magento 1 is a huge codebase and using `nfs_mount_enabled: true` is re
 Normal details of a composer build for Magento 2 are on [Magento 2 site](https://devdocs.magento.com/guides/v2.3/install-gde/composer.html) You must have a public and private key to install from Magento's repository; when prompted for "username" and "password" in the composer create it's asking for your public and private keys.
 
 ```
-ddev config --project-type=magento2 --docroot=pub --create-docroot=true
+ddev config --project-type=magento2 --docroot=pub --create-docroot=true --project-name=magento2
 ddev start
 ddev composer create --repository=https://repo.magento.com/ magento/project-community-edition
 ddev ssh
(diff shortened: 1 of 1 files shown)


docs/users/cli-usage.md
```diff
@@ -277,7 +277,7 @@ ddev start
 ddev composer create --repository=https://repo.magento.com/ magento/project-community-edition
 ddev ssh
 bin/magento setup:install  --db-host=db --db-name=db --db-user=db --db-password=db  --admin-firstname=Magento --admin-lastname=User --admin-email=user@example.com  --admin-user=admin --admin-password=admin123 --language=en_US
-bin/magento setup:store-config:set --base-url="https://magento2.ddev.site/"
+bin/magento setup:store-config:set --base-url="https://magento2-ddev.ddev.site/"
```
rfay: If you had created a directory named "magento2" instead of one named "magento2-ddev", it would have been right, I think. 

Probably the best fix here is to mention that. Add "mkdir magento2" to the front end, probably use `ddev config --project-type=magento2 --docroot=pub --create-docroot=true --project-name=magento2` to be even more explicit.

rafal-kos: @rfay you are right. I missed that default directory name was different on my side. I've added --project-name to be more explicit as you said. It will be easier for new users to test ddev without going deeply into documentation

rfay:
Thanks so much for this, @rafal-kos . I ended up changing the approach a bit and adding more text to try to make clear where the project name came from and such. For most people it's best for the project name to match the directory name, so I wasn't comfortable with just forcing the project name here. 

Thank you very much for catching this problem and then coming up with a solution!

PR #2079, opened 2020-02-10
Update pantheon.md

katielgc:
Hi there, I just wanted to make an edit from auth pantheon to auth-pantheon as the command is incorrect in the doc. 

Thanks


rfay:
@katielgc thanks so much for taking the time to provide this.

However, in the latest version of ddev, v1.13.0, auth-pantheon changed to `ddev auth pantheon`, please see https://github.com/drud/ddev/releases/tag/v1.13.0

Happy to continue the conversation here. If you're using an earlier version of ddev, then auth-pantheon is in fact correct.

PR #2086, opened 2020-02-13
Clarify TYPO3 "Trusted host patterns mismatch"

jonaseberledmind:
## The Problem/Issue/Bug:
Installing via https:// does not give problems any more since TYPO3 v9

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -208,7 +208,7 @@ ddev composer create "typo3/cms-base-distribution:^9" --prefer-dist
 ddev start
 ```
 
-**A Some versions of TYPO3 an install may fail if you use the https URL to install ("Trusted hosts pattern mismatch"). Please use "http" instead of "https" for the URL while doing the install.**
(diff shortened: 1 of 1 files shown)


docs/users/cli-usage.md
```diff
@@ -208,7 +208,7 @@ ddev composer create "typo3/cms-base-distribution:^9" --prefer-dist
 ddev start
 ```
 
-**A Some versions of TYPO3 an install may fail if you use the https URL to install ("Trusted hosts pattern mismatch"). Please use "http" instead of "https" for the URL while doing the install.**
+**On TYPO3 versions < 9 an install may fail if you use the https URL to install because the allowed proxy is not configured yet ("Trusted hosts pattern mismatch"). Please use "http" instead of "https" for the URL while doing the install.**
```
rfay: I could have sworn this also happened on v9 - or did that patch make it into TYPO3 v9 as well?

jonaseberle: Hi Randy, yes, it is in 9.5, too: https://review.typo3.org/c/Packages/TYPO3.CMS/+/61363 (9.5)
https://review.typo3.org/c/Packages/TYPO3.CMS/+/61374 (master)

rfay: Thanks!

PR #2093, opened 2020-02-28
NFS Installer on Windows breaks when spaces present

dasginganinja:
## The Problem/Issue/Bug:
The NFS Installer on Windows errored out.

The $HOME variable isn't quoted and when a space is present in the profile it gets expanded into another parameter to the docker command.
This PR quotes the value to prevent unintended consequences.


rfay:
Thanks for this! I could have sworn that class of failures was already fixed in there :)

PR #2096, opened 2020-02-29
Prevent .gitignore from ignoring itself

paulvandermeijs:
## The Problem/Issue/Bug:

.gitignore ignores itself

Removes .gitignore from .gitignore

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -914,7 +914,7 @@ func PrepDdevDirectory(dir string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rfay:
This wasn't an error, but an attempt to get all of this working for teams. From email to you:

> * The inclusion of .ddev/.gitignore (https://github.com/drud/ddev/pull/2096) ignoring itself is not an error, it was done with community feedback.  Here's the PR: https://github.com/drud/ddev/pull/1998 - basically `ddev start` creates that .gitignore for teams. I do agree that it's a bit awkward. 

Closing, but happy to continue the conversation!

PR #2102, opened 2020-03-02
Provide launch capability for PHPMyAdmin and MailHog interfaces

wolffc:
this allows the ddev launch command to launch PHPMyAdmin and and Mailhog

## The Problem/Issue/Bug:
the developer has to type urls for php my admin / mailhog and needs to rember the port numbers of the tools 

with `ddev launch` there is already a script in place to launch the webbrowser
this patch introduces two "keywords" `pma` and `mailhog` which might be used instead of the path

PHPMyAdmin: ` ddev launch pma`
Mailhog: `ddev launch mailhog`
regular website: `ddev launch`
regular website with path: `ddev launch some/url`

## Related Issue Link(s):
-

### DIFF
--- a/cmd/ddev/cmd/dotddev_assets/commands/host/launch
+++ b/cmd/ddev/cmd/dotddev_assets/commands/host/launch
@@ -2,18 +2,28 @@
 
 ## #ddev-generated
 ## Description: Launch a browser with the current site
-## Usage: launch [path]
-## Example: "ddev launch" or "ddev launch /admin/reports/status/php" or "ddev launch phpinfo.php"
+## Usage: launch [path|pma|mailhog]
+## Example: "ddev launch" or "ddev launch /admin/reports/status/php" or "ddev launch phpinfo.php" or "ddev launch pma" or "ddev launch mailhog"
 
-FULLURL=${DDEV_PRIMARY_URL}
-
-if [ -n "${1:-}" ] ; then 
-  if [[ ${1::1} != "/" ]] ; then 
-    FULLURL="${FULLURL}/"; 
-  fi
-
-  FULLURL="${FULLURL}${1}"; 
-fi
+case "${1}" in
+  mailhog)
+    # build the MailHog URL
+    FULLURL="http://${DDEV_HOSTNAME}:${DDEV_MAILHOG_PORT}"
+    ;;
+  pma)
+    # build the PhpMyAdmin URL
+    FULLURL="http://${DDEV_HOSTNAME}:${DDEV_PHPMYADMIN_PORT}"
+    ;;
+  *)
+    FULLURL=${DDEV_PRIMARY_URL}
+    if [ -n "${1:-}" ]; then
+      if [[ ${1::1} != "/" ]]; then
+        FULLURL="${FULLURL}/"
+      fi
+    fi
+    FULLURL="${FULLURL}${1}"
+    ;;
+esac
 
 case $OSTYPE in
   linux-gnu)


cmd/ddev/cmd/dotddev_assets/commands/host/launch
```diff
@@ -2,18 +2,28 @@
...
-
-if [ -n "${1:-}" ] ; then 
-  if [[ ${1::1} != "/" ]] ; then 
-    FULLURL="${FULLURL}/"; 
-  fi
-
-  FULLURL="${FULLURL}${1}"; 
-fi
+case "${1}" in
+  mailhog)
```
gilbertsoft: Perhaps we should use some kind of escaping character here to avoid conflicts with normal URLs

cmd/ddev/cmd/dotddev_assets/commands/host/launch
```diff
@@ -2,18 +2,28 @@
...
+    # build the MailHog URL
+    FULLURL="http://${DDEV_HOSTNAME}:${DDEV_MAILHOG_PORT}"
+    ;;
+  pma)
```
gilbertsoft: Here too ofc

cmd/ddev/cmd/dotddev_assets/commands/host/launch
wolffc: if your webserver has a mailhog domain you could call it like `ddev launch /mailhog` so i think escape char is not really nessasary. but im open to it (if you find a good escape char)

here are the options i have considred:

a) `--` like `ddev launch --mailhog` this feels CLI like, but `-` is still valid in urls
b) add it as a Second parameter `ddev launch - mailhog` this would cleanly remove the conflict 
c) open for other ideas?

cmd/ddev/cmd/dotddev_assets/commands/host/launch
wolffc: see discussion here: https://github.com/drud/ddev/pull/2102#discussion_r386459331

rfay:
I think I'd prefer a new command, since "launch" already takes an arg which is the URI.

leymannx:
`ddev open` maybe?

gilbertsoft:
I vote for `ddev mailhog` and `ddev pma`

rfay:
I worry about adding too many default commands, afraid that people will be overwhelmed by the number of options. I agree that two new custom commands would be a very easy option.

wolffc:
I am with Randy on adding not to many default commands,
but fear its confusing for a user to understand/rember the dfference between `ddev launch` and `ddev open` (as both commands) will open the webbrowser. (not s

maybe the way is to integrate both into one command with "clean" syntax. (not overloading parameters) 

for example `ddev open mailhog|pma|site| [path]`

so launch command could be removed in future versions of ddev.  `ddev open site home` would open `https://project.ddev.site/home` 
this way it would be possible to bundle all calls that open a webbrowser.

what do you think about this suggestion

leymannx:
I find `ddev open` much more intuitive than `ddev launch`. Either way wouldn't it still be better to just have one command? `ddev open` opens project base URL. `ddev open /some-dir` (note the starting slash) opens project base URL/some-dir. `ddev open pma` opens project PHPMyAdmin. `ddev open mailhog` opens project Mailhog.  Or `ddev launch` of course.

rfay:
I think we got lost on this one, but I'm open to various solutions, including opening the web interface of 3rd-party projects like solr.  I don't have any objection to "open" and "launch" being aliases, that's not hard to do. But closing this for now, I certainly have no objection, but we didn't come up with an actionable path.

rfay:
How about doing it with `ddev launch --<service>`, for example, `ddev launch --phpmyadmin` or `ddev launch --service=phpmyadmin` with single-letter aliases like `ddev launch -p` ?

rfay:
I'm going to reopen and retitle this so it doesn't get dropped. I think we can do something that will make us all happy.

rfay:
Closing this in favor of #2137, which adds -p and -m args to launch. Would appreciate your review over there @wolffc @leymannx @gilbertsoft , and thanks for kicking off this conversation.

Issue #2076, opened 2020-02-07
Configuration router_bind_all_interfaces is not working


benjifisher:
**Describe the bug**

I set `router_bind_all_interfaces: true` in `~/.ddev/global_config.yaml` but my site is still available only from localhost.

**To Reproduce**

1. Either edit `~/.ddev/global_config.yaml` and set `router_bind_all_interfaces: true` or use the command  `ddev config global --router-bind-all-interfaces=true`. (I tried both.)
2. Check the assigned ports with `ddev describe`.
3. Check the active ports with `sudo netstat -lntp`.
4. Visit the site from another computer on the same network.

**Expected behavior**

I expect the ports I find in Step 2 to be open to all interfaces in Step3.

I expect to see my site in a browser in Step 4.

**Screenshots**

How about screen dumps instead of shots?

Step 2:

```
$ ddev describe
NAME        TYPE     LOCATION            URL                           STATUS        
demo-site2  drupal8  ~/sites/demo/site2  https://demo-site2.ddev.site  running

Project Information
-------------------
PHP version:   	7.3
MariaDB version	   
MySQL version  	   

URLs
----
http://demo-site2.ddev.site 
http://127.0.0.1:32792      
https://demo-site2.ddev.site
https://127.0.0.1:32791     

MySQL/MariaDB Credentials
-------------------------
Username: "db", Password: "db", Default database: "db"

or use root credentials when needed: Username: "root", Password: "root"

Database hostname and port INSIDE container: db:3306
To connect to db server inside container or in project settings files: 
mysql --host=db --user=db --password=db --database=db
Database hostname and port from HOST: 127.0.0.1:32790
To connect to mysql from your host machine, 
mysql --host=127.0.0.1 --port=32790 --user=db --password=db --database=db

Other Services
--------------
MailHog:   	http://demo-site2.ddev.site:8025
phpMyAdmin:	http://demo-site2.ddev.site:8036

DDEV ROUTER STATUS: healthy	
ssh-auth status: healthy 
```

Step 3:

```
$ sudo netstat -lntp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      910/systemd-resolve 
tcp        0      0 127.0.0.1:32790         0.0.0.0:*               LISTEN      3339/docker-proxy   
[...]

PR #2103, opened 2020-03-03
Fix some typos

benjifisher:
## The Problem/Issue/Bug:

There are some typos in the documentation.

Er, fix the typos. ;)

- Review the updated files in the GitHub UI.
- Generate the docs using `mkdocs`. This may be overkill.

N/A

## Related Issue Link(s):

N/A

## Release/Deployment notes:

Rebuild the docs as usual.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,7 +15,7 @@
 
 ### Using ddev alongside other development environments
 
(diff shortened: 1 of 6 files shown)


benjifisher:
You may not consider all of these to be typos. For example, my spell checker and I agree that "problem solving" is two words. If you disagree, then you can remove the space added in this PR.

@rfay, This PR is my way of saying "thank you" for your help on #2076.

rfay:
Thanks for this!

rfay:
Awesome, thanks! I think we should probably add spellcheck into  the pre-push hooks to catch these things. I pushed one more commit with minor rewordings. It's ready to go. Thanks!

PR #2112, opened 2020-03-13
Add timeout to IsInternetActive(), fixes #1969

### LINKED ISSUE #1969: Add timeout to IsInternetActive()
**Describe the bug**

IsInternetActive() does a DNS lookup to determine whether we have internet. In terrible internet situations, that can be really slow and annoying.

https://github.com/drud/ddev/blob/ae1940866813afa426245ac32d37e0280be79f1a/pkg/nodeps/utils.go#L39-L42

Add a timeout to that usage. Maybe a warning?

**To Reproduce**

**Expected behavior**

**Screenshots**

 - Host computer OS and Version: [e.g. Windows 10, macOS High Sierra]
 - Docker version information (use `docker version`) and insert it here.

**Additional context**

atomicptr:
## The Problem/Issue/Bug:

#1969 

Stops the DNS lookup if it takes longer then 500ms (the value might need adjustments).

### DIFF
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -2,6 +2,8 @@ package nodeps
 
 import (
 	"context"
+	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"time"
@@ -37,12 +39,36 @@ func IsDockerToolbox() bool {
 
 //IsInternetActive() checks to see if we have a viable
 // internet connection. It just tries a quick DNS query.
-// This requires that the named record be query-able.
+// This requires that the named record be query-able
 func IsInternetActive() bool {
 	const timeout = 500 * time.Millisecond
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	_, err := net.DefaultResolver.LookupHost(ctx, "i-exist.ddev.site")
-	return err == nil && ctx.Err() == nil
+	randomURL := RandomString(10) + ".ddev.site"
+	addrs, err := net.DefaultResolver.LookupHost(ctx, randomURL)
+
+	// Internet is active (active == true) if both err and ctx.Err() were nil
+	active := err == nil && ctx.Err() == nil
+	if os.Getenv("DDEV_DEBUG") != "" {
+		fmt.Printf("IsInternetActive DEBUG: err=%v ctx.Err()=%v addrs=%v IsInternetactive==%v, randomURL=%v\n", err, ctx.Err(), addrs, active, randomURL)
+	}
+	if active == false {
+		fmt.Println("Internet connection not detected")
+	}
+	return active
+}
+
+// From https://www.calhoun.io/creating-random-strings-in-go/
+var seededRand *rand.Rand = rand.New(
+	rand.NewSource(time.Now().UnixNano()))
+
+func RandomString(length int) string {
+	const charset = "abcdefghijklmnopqrstuvwxyz"
+
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
 }
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -1,8 +1,10 @@
 package nodeps
 
 import (
+	"context"
 	"net"
 	"os"
+	"time"
 )
 
 // ArrayContainsString returns true if slice contains element
@@ -37,6 +39,10 @@ func IsDockerToolbox() bool {
 // internet connection. It just tries a quick DNS query.
 // This requires that the named record be query-able.
 func IsInternetActive() bool {
+	const timeout = 500 * time.Millisecond
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
 	_, err := net.LookupHost("i-exist.ddev.site")
-	return err == nil
+	return err == nil && ctx.Err() == nil
 }
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -39,6 +39,9 @@ func IsDockerToolbox() bool {
 
 var isInternetActiveAlreadyChecked = false
 var isInternetActiveResult = false
+var isInternetActiveNetResolver interface {
+	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+} = net.DefaultResolver
 
 //IsInternetActive() checks to see if we have a viable
 // internet connection. It just tries a quick DNS query.
@@ -55,7 +58,7 @@ func IsInternetActive() bool {
 	defer cancel()
 
 	randomURL := RandomString(10) + ".ddev.site"
-	addrs, err := net.DefaultResolver.LookupHost(ctx, randomURL)
+	addrs, err := isInternetActiveNetResolver.LookupHost(ctx, randomURL)
 
 	// Internet is active (active == true) if both err and ctx.Err() were nil
 	active := err == nil && ctx.Err() == nil
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -39,6 +39,9 @@ func IsDockerToolbox() bool {
 
 var isInternetActiveAlreadyChecked = false
 var isInternetActiveResult = false
+
+// In order to override net.DefaultResolver with a stub, we have to define an
+// interface on our own since there is none from the standard library.
 var isInternetActiveNetResolver interface {
 	LookupHost(ctx context.Context, host string) (addrs []string, err error)
 } = net.DefaultResolver
@@ -80,6 +83,7 @@ func IsInternetActive() bool {
 var seededRand *rand.Rand = rand.New(
 	rand.NewSource(time.Now().UnixNano()))
 
+// RandomString creates a random string with a set length
 func RandomString(length int) string {
 	const charset = "abcdefghijklmnopqrstuvwxyz"
 
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -37,10 +37,19 @@ func IsDockerToolbox() bool {
 	return false
 }
 
+var isInternetActiveAlreadyChecked = false
+var isInternetActiveResult = false
+
 //IsInternetActive() checks to see if we have a viable
 // internet connection. It just tries a quick DNS query.
-// This requires that the named record be query-able
+// This requires that the named record be query-able.
+// This check will only be made once per command run.
 func IsInternetActive() bool {
+	// if this was already checked, return the result
+	if isInternetActiveAlreadyChecked {
+		return isInternetActiveResult
+	}
+
 	const timeout = 500 * time.Millisecond
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
@@ -56,6 +65,11 @@ func IsInternetActive() bool {
 	if active == false {
 		fmt.Println("Internet connection not detected")
 	}
+
+	// remember the result to not call this twice
+	isInternetActiveAlreadyChecked = true
+	isInternetActiveResult = active
+
 	return active
 }
 
--- a/pkg/nodeps/utils.go
+++ b/pkg/nodeps/utils.go
@@ -43,6 +43,6 @@ func IsInternetActive() bool {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	_, err := net.LookupHost("i-exist.ddev.site")
+	_, err := net.DefaultResolver.LookupHost(ctx, "i-exist.ddev.site")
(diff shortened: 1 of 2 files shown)


atomicptr:
I was browsing the open issues and found this, I haven't really encountered this, tho there are two places where this function is called / are affected https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/root.go#L77 and https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/root.go#L127 ).

As described in the related issue, when your connection is dodgy the DNS lookup might take a while. You can try to prolong this by adding a ``time.Sleep(5 * time.Second)`` inside the function to reproduce this behaviour, which will be cancelled after 500ms (and the function will return false).

rfay:
Thanks! Glad you poked this along. I think it might be possible to recreate the scenario by using a bogus DNS server, something like that.

atomicptr:
Is there still something I have to do? I sadly can't see why the buildkite checks failed.

rfay:
Thanks so much @atomicptr - Don't worry about the buildkite tests, I'll look at them and get things fixed up. Enough tests succeeded that I can see that nothing is broken as a result of this PR.  Some of the testbots got sick (not coronavirus) while I was on vacation and I haven't really got them all going again yet because I've been working on things other than code in the last week or so.

What's really needed here is manual testing instructions. In a perfect world we'd have an automated test for this, but I haven't come up with an idea for one. But it would be lovely if you could pioneer a manual testing approach for this. I think most likely a virtual machine approach with either faked-out DNS or a slowed network are the options I can think of. 

On macOS the lovely Parallels package has  a "network conditioner" that might get us somewhere for manual testing.

_Ubuntu_18_04__Configuration_and_Control_Center_and_Add_timeout_to_IsInternetActive____fixes__1969_by_atomicptr_·_Pull_Request__2112_·_drud_ddev

atomicptr:
After setting up a small VM (using multipass) with ddev (and also a binary from this PR), I used [netem](https://wiki.linuxfoundation.org/networking/netem) to slow down my network connection:

(This will also slow down SSH...)

```bash
$ sudo tc qdisc add dev [network-interface] root netem delay 1s
```

Example pinging the google DNS:

```bash
$ ping 8.8.8.8
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=56 time=1019 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=56 time=1019 ms
64 bytes from 8.8.8.8: icmp_seq=3 ttl=56 time=1019 ms
64 bytes from 8.8.8.8: icmp_seq=4 ttl=56 time=1019 ms
64 bytes from 8.8.8.8: icmp_seq=5 ttl=56 time=1019 ms
...
```

At first I tried to show the difference by checking the execution times, sadly the results are generally all over the place to show anything meaningful here.

Another (not so great) way to test this is (with the slowed down connection) adding some prints:

```go
func IsInternetActive() bool {
	fmt.Println("@@@ IsInternetActive Called!")
	const timeout = 500 * time.Millisecond
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()

	_, err := net.DefaultResolver.LookupHost(ctx, "i-exist.ddev.site")
	if ctx.Err() != nil {
		fmt.Println("@@@", ctx.Err())
	}
	return err == nil && ctx.Err() == nil
}
```

Which now prints out as expected:

```bash
$ ddev poweroff
$ rm ~/.ddev/.update
$ ./ddev-with-errprint-f27de22 start
@@@ IsInternetActive Called!
@@@ context deadline exceeded
Starting test-project... 
...
```

Without the slowdown:

```bash
./ddev-with-errprint-f27de22 start
@@@ IsInternetActive Called!
Starting test-project... 
```

If nothing else tho, this made me realize i forgot to pass down the context to LookupHost :).

rfay:
Will look forward to another commit to fix that, thanks!

atomicptr:
I already added the commit https://github.com/drud/ddev/pull/2112/commits/b20e6ac2b55367ac1d898e4c643ea34be2c77332 :)

rfay:
Testing this, I note that IsInternetActive() unfortunately doesn't return valid responses in the case of bad internet. Once the (static) queried hostname is resolved finally, IsInternetActive() will return true, because it can suddenly be resolved.

atomicptr:
It might probably make more sense to ping something instead of trying to resolve a hostname then 🤔 . I'll change this. Some preference on what to ping tho? *.ddev.site is a loopback to localhost after all wouldn't make much sense to ping that.

rfay:
Right now I'm working on it. The solution I'm currently working on is to use a randomstring.ddev.site dns lookup.

ping isn't a very good approach since it's often blocked in networks.

Please don't make any changes while I'm working on this, thanks.

atomicptr:
Ok ;)

rfay:
Do you want to or have time to work on this? If so, go to town. I'll push a commit into my own fork with the work I've done so far, and make some suggestions about what could happen next...

rfay:
OK, assuming and hoping for a "yes", I pushed one commit, which is my WIP here. Happy to have it improved.

Things that I'd still like to see:
* Only execute the both of the function once per ddev execution (stash result globally and short-circuit if it's already been set? )
* Careful manual testing of both the good internet and bad internet cases
* Evaluation of the 500ms number you started with.

`export DDEV_DEBUG=true` will get you the debug printout of what's going on.

atomicptr:
Yeah I'd love to :)

atomicptr:
> * Only execute the both of the function once per ddev execution (stash result globally and short-circuit if it's already been set? )

Done with: https://github.com/drud/ddev/pull/2112/commits/7936ea44efcd4bbbc4ace8740131559fa73372cc

> * Careful manual testing of both the good internet and bad internet cases

Slow internet:

```
IsInternetActive DEBUG: err=lookup gmnjirnahi.ddev.site on 127.0.0.53:53: dial udp 127.0.0.53:53: i/o timeout ctx.Err()=context deadline exceeded addrs=[] IsInternetactive==false, randomURL=gmnjirnahi.ddev.site
Internet connection not detected
Starting test-project... 
```

Normal internet:

```
IsInternetActive DEBUG: err=<nil> ctx.Err()=<nil> addrs=[127.0.0.1] IsInternetactive==true, randomURL=jeinzoiolc.ddev.site
Starting test-project... 
```

> * Evaluation of the 500ms number you started with.

I do think it's a bit on the upper end on what I'd like to use, but we can be fairly sure that the connection is terrible when a DNS lookup takes half a second.

rfay:
In testing with the 500ms timeout, I didn't find it annoying on slow internet.

atomicptr:
> In testing with the 500ms timeout, I didn't find it annoying on slow internet.

What I meant is that I think this is a good value between "don't be annoying" and "a DNS lookup shouldn't take this long -> internet is probably slow"

rfay:
Yeah, I think it's probably OK.  Thanks for the swift effort on this! I look forward to trying it out.

atomicptr:
Adding a test for this is probably a good idea, I'll have to make the net.DefaultResolver exchangeable tho (to replace it with a fake one).

rfay:
Weird - I see that TestIsInternetActiveTimeout failed on *nfs* on macOS ONLY.  Probably worth paying attention to, looks like it might be a hard one.  I sent you an invite to add you to buildkite team so you can see buildkite builds, sent it to your github email. 

```
=== RUN   TestIsInternetActiveTimeout
--
  | TestIsInternetActiveTimeout: utils_test.go:48:
  | Error Trace:	utils_test.go:48
  | Error:      	Should be false
  | Test:       	TestIsInternetActiveTimeout
  |  

```

atomicptr:
That is an odd one, I removed the sleep and replaced it with something that works better with the context, hoping that this will fix this :thinking:.

rfay:
Totally weird, same exact failure! It must be real. I guess you work on Linux and don't have access to macOS? I'll try to test this tomorrow on macOS with NFS going. *Very* unexpected.

rfay:
I haven't been able to get a test failure locally just running any or all of the nodeps tests in Goland, but I'll try running the whole thing with `export DDEV_TEST_USE_NFSMOUNT=true && make test`

I was wondering if it was a specific test runner, but it's happened on two different macOS machines.

Of course, the nodeps tests don't have any dependency on any projects running, and they certainly don't care about nfs. So...

atomicptr:
I have an old Mac Mini somewhere for testing stuff like this, I'll see if I can reproduce the issue there.

rfay:
Repeatable on my own laptop (also running Catalina). I let the test run all night. Same exact results. Baffled. I guess the next step is to try running a subset of the tests. Something apparently causes interaction between tests or the environment *only* on NFS????

Here's the result, https://gist.github.com/rfay/481f372e740197cbf46792d903b32cf4#file-gistfile1-txt-L4319

https://gist.github.com/rfay/481f372e740197cbf46792d903b32cf4

pkg/nodeps/utils.go
```diff
@@ -33,10 +37,55 @@ func IsDockerToolbox() bool {
 	return false
 }
 
+var isInternetActiveAlreadyChecked = false
+var isInternetActiveResult = false
+var isInternetActiveNetResolver interface {
+	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+} = net.DefaultResolver
```
rfay: I guess I'm not terribly happy that the desire to add a test imposes mocking complexity into the tested code here.

rfay:
Looking at utils_test.go, it seems super impressive to me, a heroic effort.  I'm pretty sure we just have a test interaction that we don't understand. But I sure don't understand it.

pkg/nodeps/utils.go
atomicptr: Hm, do you have a better idea? Sadly there is no standard interface for the resolver (afaik) and we have to do something to replace the default one.

atomicptr:
I wonder if we can reproduce this behavior in isolation, like taking the code / tests without anything else around it. If it works in isolation, there is probably something wrong somewhere.

rfay:
I was unable to recreate it with `export DDEV_TEST_USE_NFSMOUNT=true; make test TESTARGS="-run TestIsInternet"`

pkg/nodeps/utils_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	assert.False(t, IsInternetActive())
+}
+
+func TestIsInternetActiveTimeout(t *testing.T) {
+	resetVariables()
+
+	isInternetActiveNetResolver = netResolverStub{
+		sleepTime: 501 * time.Millisecond,
```
rfay:
```suggestion
		sleepTime: 1000 * time.Millisecond,
```

Let's set sleepTime in TestIsInternatActiveTimeout dramatically higher, like 1000 * time.Millisecond. There's no reason to have 501 when 500 is the set difference. The possibility of a 1ms deficiency in resolution is high. I'd say that's probably the first thing to do, unless I'm misunderstanding the code.

rfay:
(I went ahead and committed that code suggestion right from the github interface. Pretty slick.)

pkg/nodeps/utils_test.go
atomicptr: Yeah, thats a good idea

atomicptr:
> (I went ahead and committed that code suggestion right from the github interface. Pretty slick.)

I tried to do the same thing but you were a bit faster haha :)

atomicptr:
Looks like the macOS NFS one passed, if this was the solution I'm kinda lost on why this only failed on this specific setup 🤔.

rfay:
Yup, passed, yay! I have no idea why that one setup would be different. OTOH 1ms is a pretty small window. 

I think the only thing we need to do now is add 
* Add headers to the functions
* Add comments to explain the unusual  isInternetActiveNetResolver

Awesome work, thanks. I'd like to have a zoom with you soon if you're open to that to meet in person.

atomicptr:
Hopefully the tests stay green :).

> I'd like to have a zoom with you soon if you're open to that to meet in person.

Sounds good to me.

rfay:
I think this is ready to go, just needs a final round of manual testing. I'll take a look in the next little bit. 

Do you have any time for a zoom in the next hour @atomicptr, or between 4-5PM CET? If not, please suggest a time Monday afternoon your time and we'll do it!  Thanks!

rfay:
Manual testing still shows it working fine. The half-second delay when in bad internet situation is hardly noticeable, but I manually increased it in the code to 3s and also increased the horribleness of the custom network conditioner in Parallels to 4s, and it was easier to see and understand the delay.

IMO it's doing great and ready to pull when you say it's ready.

atomicptr:
> Do you have any time for a zoom in the next hour @atomicptr, or between 4-5PM CET?

I guess I already missed the "in the next hour" mark, rn (4-5PM CET) would be fine though.

> IMO it's doing great and ready to pull when you say it's ready.

Ready when you are! :)

PR #2150, opened 2020-04-06
Create DDEV-Live.md

ekl1773:
Stub for adding info and links for DDEV-Live

## The Problem/Issue/Bug:
DDEV-Live is missing from the hosting providers in the Local docs.

Adds a simple starter page that links to the DDEV-Live docs site.


rfay:
I edited this a bit for clarity @ekl1773 @unn. If you think it's good we'll get it in.

unn:
Looks good to me.

PR #2153, opened 2020-04-07
Windows: mkcert: add to firefox: add mkcert CA

josefglatz:


josefglatz:
I had to use `rootCA.pem` instead of `rootCA-key.pem` to add the CA to my firefox trusted CA

rfay:
I'm pretty sure you're right.

PR #2175, opened 2020-04-15
No need for double "ddev start" in D8/9 Composer Setup Examples

alesrebec:
Tested with and without "ddev start" after "ddev config ..." command when testing ddev 1.14-rc1 and it worked the same in both cases.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -137,7 +137,6 @@ Get started with Drupal 8 projects on ddev either using a new or existing compos
 mkdir my-drupal8-site
 cd my-drupal8-site
 ddev config --project-type=drupal8 --docroot=web --create-docroot
-ddev start
 ddev composer create drupal/recommended-project:^8
 ddev composer remove drupal/core-project-message
 ddev composer require drush/drush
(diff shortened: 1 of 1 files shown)


alesrebec:
@rfay is the extra "ddev start" there for apache and/or windows?

rfay:
It's actually important. `The ddev composer create` completely destroys everything in the directory. So there's no settings.php or settings.ddev.php or anything. Those are created both on `ddev start` and on `ddev config`. But since the settings get destroyed by the composer build... the extra `ddev start` is there to get them created again.

rfay:
And THANK YOU FOR TESTING!

rfay:
It's certainly possible that the `ddev composer create` could execute a start after it was done. that would clean this up.

alesrebec:
I was referring to an extra `ddev start` before `ddev composer create` if you check the commit.
The second one is fine but the first one wasn't necessary when I successfully spinned up the project without it.

rfay:
Ah, I misunderstood. Yes, you're right that because `ddev composer` automatically does the "start", that first start is unnecessary. We could get rid of the other one, as I said, but having `ddev composer create` do a "start" at the end. I think it would be an improvement. Mind if I add that to this PR?

alesrebec:
Sure, I don't mind ;)

rfay:
Here's what I'm thinking right now. 
* I'll make the second start unnecessary by adding a start to the end of `ddev composer create`.
* Remove the second start from the docs.
* I'll leave the first start (the one you removed) in the docs just for instructional purposes, even though `ddev composer` does the start automatically. 

Could you live with that?

alesrebec:
Hehe, yea, I'm totally fine with it. ;)

Just wondered intially why two ddev starts and when I tried with just one, it worked.

rfay:
Closing in favor of #2184 - thanks so much for pointing this out.

Issue #2196, opened 2020-04-17
Try to put ddev drush.yml in more than one place in filesystem


jcnventura:
**Is your feature request related to a problem? Please describe.**
Since #2025 was merged , ddev tries to create the drush/drush.yml file instead of web/sites/all/drush.yml.

Because of this change, projects that already had a drush/drush.yml file no longer get a ddev-generated file with the correct site URL, and, as a result, `drush uli` returns `http://default` as ddev's base path.

**Describe the solution you'd like**
According to the [example.drush.yml](https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml) (line 17), drush will try to locate the drush.yml file in 
> 2. Drupal /drush and sites/all/drush folders, or the /drush folder in the directory above the Drupal root.

Of the three locations, described above, ddev used to place the file in the second option, and now places it in the third location.

My preferred solution is that ddev first try to create the file in drush/drush.yml (current behaviour), and if that fails, attempt to create it in web/sites/all/drush.yml (previous behaviour), and only then give up with the "Failed to WriteDrushYML" error.

**Describe alternatives you've considered**
Another option would be for ddev to modify the existing drush.yml, but since that is a committed file in our case, it would mark it as modified, introducing the risk that somebody could commit it.

rfay:
You might be a candidate for managing your own settings files. There are many ways to do that, summary in https://ddev.com/ddev-local/controlling-cms-settings-files-in-ddev-local/

jcnventura:
We really, really would like to have ddev still create the drush.yml automatically.

jcnventura:
Note that the previous behaviour worked fine for us with v1.12. It was only now that we started using v1.13 that my colleagues complained about the missing correct URL configuration.

One option would be for us to go back to 1.12, but then we would start to be pestered by #2116.

I believe the solution in the PR improves the `disable_settings_management: false` option, keeps the location added in #2025 as the preferred location, by taking advantage of drush's multiple lookup paths for this file.

rfay:
Closing this in favor of #2216

rfay:
@jcnventura and @truls1502 I hope you can test the prerelease [v1.15-alpha5](https://github.com/drud/ddev/releases/tag/v1.15-alpha5) for the DRUSH_OPTIONS_URI support while it's still in prerelease, thanks!

PR #2197, opened 2020-04-17
#2196: Attempt twice to create the drush.yml file

jcnventura:
## The Problem/Issue/Bug:
Since #2025 was merged , ddev tries to create the drush/drush.yml file instead of web/sites/all/drush.yml.

Because of this change, projects that already had a drush/drush.yml file no longer get a ddev-generated file with the correct site URL, and, as a result, drush uli returns http://default as ddev's base path.

This PR restores the previous behaviour of ddev <1.13, as a a fallback in case the new behaviour fails.

Run ddev start on a project with an existing drush/drush.yml. Verify that it creates the web/sites/all/drush/drush.yml file.

Haven't created any automated test.

## Related Issue Link(s):
https://github.com/drud/ddev/issues/2196

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -695,7 +695,10 @@ func drupal8PostStartAction(app *DdevApp) error {
 		// drush9 wants drush.yml
 		err := WriteDrushYML(app, filepath.Join(app.AppRoot, "drush", "drush.yml"))
 		if err != nil {
-			util.Warning("Failed to WriteDrushYML: %v", err)
+			err := WriteDrushYML(app, filepath.Join(filepath.Dir(app.SiteSettingsPath), "..", "all", "drush", "drush.yml"))
+			if err != nil {
+				util.Warning("Failed to WriteDrushYML: %v", err)
+			}
 		}
 
 		err = WriteDrushrc(app, filepath.Join(filepath.Dir(app.SiteSettingsPath), "drushrc.php"))


rfay:
CircleCI changed something about their build image, so you'll need to rebase against upstream master to make this pass the CircleCI tests. Thanks!

jcnventura:
Sorry, only saw this now. Rebased now.

jcnventura:
Still broken. Not sure if it is possible to rerun the failed tests, which seem to be CI problems.

rfay:
I restarted the affected tests. One of them was a network error. The CircleCI ones looked like they might be real fails, not sure.

What I'd really like to do is set up a conversation with you and @truls1502 to brainstorm about your team's actual situation and various ways to solve it.  Every time we consider making a change to fix one team's workflow we end up breaking it for other teams. This one was mostly done for people who do multisite, as I remember, because they didn't want a sites/all in the first place, and ddev had to create one.

Would there be a time some afternoon your time that we could chat this out? Maybe bring in some Drush people or talk in the #drush channel to get some help too.

jcnventura:
Indeed, I wonder if there's an environmnent variable to set the URL.. That way, ddev could just make sure to pass that variable, and skip the creation of this file altogether..

jcnventura:
I just have done the research before answering the above.

There is a DRUSH_OPTIONS_URI variable that drush can use, and could probably be used by ddev instead of creating these drush files:
* 9.x: https://docs.drush.org/en/9.x/using-drush-configuration/
* 10.x: https://docs.drush.org/en/master/using-drush-configuration/

Unfortunately, it's only for Drush 9 and above, This means that the code to write drushrc.php will remain as is, but ddev should be able to no longer write the drush.yml file.

I'll keep this PR around, as maybe this could be something going into 1.13.3, to partly undo the BC-break for those (like us) relying on <1.13 behaviour that ddev would create a drush/sites/all/drush.yml file.

jcnventura:
All tests passed.

rfay:
Thanks so much for this, but as we discussed the environment variable technique is better, so going to close this in favor of https://github.com/drud/ddev/issues/2216, which I definitely hope to get into v1.15

rfay:
@jcnventura and @truls1502 I hope you can test the prerelease [v1.15-alpha5](https://github.com/drud/ddev/releases/tag/v1.15-alpha5) for the DRUSH_OPTIONS_URI support while it's still in prerelease, thanks!

PR #2201, opened 2020-04-19
Add creation of .ddev/.gitignore to post-start actions on php project, fixes #2200

### LINKED ISSUE #2200: .ddev/.gitignore not updated/maintained on start with projects of type 'php'
I had a little issue today with the DDEV RC:

On ddev start the new xdebug command was created as expected. But because you removed the api version check there was no warning and in the first place the .gitignore was not updated with the new command. So my first tought was you forgot to add the new command to the .gitignore generator but ofc the mistake was from my site not running ddev config.

My suggestion is now to also recreate the .gitignore at the startup to avoid this kind of error and this will make the devs life easier....

rfay:
# The Problem/Issue/Bug:

#2200 points out that with a  project of type 'php', the .ddev/.gitignore does not get created/managed on `ddev start`. It ought to update.

Add post-start hook for 'php' project type to update.

1. Create a project of type 'php'
2. Edit the .ddev/.gitignore
3. `ddev start`

Your edits should be gone.

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -65,7 +65,7 @@ var appTypeMatrix map[string]AppTypeFuncs
 
 func init() {
 	appTypeMatrix = map[string]AppTypeFuncs{
-		nodeps.AppTypePHP: {},
+		nodeps.AppTypePHP: {postStartAction: phpPostStartAction},
 		nodeps.AppTypeDrupal6: {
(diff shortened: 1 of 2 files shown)


rfay:
@gilbertsoft the artifacts for this are at https://app.circleci.com/pipelines/github/drud/ddev/2007/workflows/6dadd9fd-1b77-4b23-a5d2-347b6405adf9/jobs/24107/artifacts and would love to see if it works for you.

gilbertsoft:
Your changes worked fine here, probem solved!

Issue #2203, opened 2020-04-20
[DOCS] Strange syntax highlighting in yaml source blocks


gilbertsoft:
For some cases readthedocs creates a strange escaping in code blocks:

![grafik](https://user-images.githubusercontent.com/25326036/79755573-a4c4bf80-8319-11ea-8f1f-f401cce021ad.png)

This is perhaps because no type is explicit declared. So we should give _```bash_ a try.

This example is from https://ddev.readthedocs.io/en/stable/users/extending-commands/#exec-execute-a-shell-command-in-a-container-defaults-to-web-container and https://github.com/drud/ddev/blob/48cf885f01b67add90a40048a9127f246c286608/docs/users/extending-commands.md#exec-execute-a-shell-command-in-a-container-defaults-to-web-container

PR #2208, opened 2020-04-21
[DOCS] Update TYPO3 quick start to new LTS

gilbertsoft:
This patch also tries to solve issue #2203 for the bash commands in this area.

## The Problem/Issue/Bug:
Today we released a new LTS version of the TYPO3 CMS.

The version constraint is updated to the new LTS.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -211,18 +211,17 @@ Quickstart instructions for database imports can be found under [Database Import
 
 #### TYPO3 Composer Setup Example
 
-```
+```bash
 mkdir my-typo3-site
 cd my-typo3-site
 ddev config --project-type=typo3 --docroot=public --create-docroot=true
 ddev start
-ddev composer create "typo3/cms-base-distribution:^9" --prefer-dist
+ddev composer create "typo3/cms-base-distribution:^10" --prefer-dist
+ddev exec touch public/FIRST_INSTALL
 ddev launch
 ```
 
 **On TYPO3 versions < 9 an install may fail if you use the https URL to install because the allowed proxy is not configured yet ("Trusted hosts pattern mismatch"). Please use "http" instead of "https" for the URL while doing the install.**
-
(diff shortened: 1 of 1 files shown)


rfay:
Yay for LTS!

docs/users/cli-usage.md
```diff
@@ -211,26 +211,25 @@ Quickstart instructions for database imports can be found under [Database Import
...
 
-```
+```bash
 mkdir my-typo3-site
 cd my-typo3-site
 ddev config --project-type=typo3 --docroot=public --create-docroot=true
 ddev start
-ddev composer create "typo3/cms-base-distribution:^9" --prefer-dist
+ddev composer create "typo3/cms-base-distribution:^10" --prefer-dist
+ddev exec touch public/FIRST_INSTALL
```
rfay, line 220: Great addition. I wish we could count on Windows users having "touch", and this could just be `touch public/FIRST_INSTALL`

docs/users/cli-usage.md
```diff
@@ -211,26 +211,25 @@ Quickstart instructions for database imports can be found under [Database Import
...
-If doing a basic TYPO3 install, you can then `touch public/FIRST_INSTALL` and hit the http URL top begin an installation.
-
 To connect to the database within the database container directly, please see the [developer tools page](developer-tools.md#using-development-tools-on-the-host-machine).
 
 To get started using ddev with a TYPO3 project, clone the project's repository and checkout its directory.
 
 #### TYPO3 Git Clone Example
 
-```
+```bash
```
rfay, line 232: I used ` ```yaml ` for the other ones I updated. Seemed to work OK.

gilbertsoft: Yes, saw that. For your cases this is fine, but here we talk about bash.

docs/users/cli-usage.md
gilbertsoft: Other option would be `echo "">public/FIRST_INSTALL` which also should work on every platform. But for me the touch approach looks more common.

Issue #2132, opened 2020-03-28
Host ssh config not copied to docker, ssh keys won't connect


damienmckenna:
**Is your feature request related to a problem? Please describe.**
It is commonplace to have multiple ssh keys on a computer - different servers have different key requirements, if one is compromised it doesn't mean all accounts on all servers are compromised, etc.

The "auth ssh" command in ddev-local makes the ssh keys available to connect to other servers, but it does not copy the ssh config file so there is no connection between those ssh keys and the individual servers being connected to. As a result, some commands will fail because they do not have the appropriate ssh key assigned for the given ssh commands.

**Describe the solution you'd like**
I'd like for there to be a scenario that results in host computer's ssh config file's contents getting copied to the docker instances. I noted that the web instance has a config file already, hence suggesting the contents be copied instead of just copying the file wholesale.

**Describe alternatives you've considered**
None right now.

**Additional context**
This problem arose when using the `ddev blt` command, which passes all arguments through to the Acquia `blt` shell command.

rfay:
You can do this by adding your .ssh/config into .ddev/homeadditions/.ssh, and it will be copied into the container on start

```
mkdir <projectroot>/.ddev/homeadditions/.ssh
cp ~/.ssh/config <projectroot>/.ddev/homeadditions/.ssh
ddev restart
```

I experimented with this by creating a .ddev/homeadditions/.ssh/config that looked like this:
```
Host <somehost>.thefays.us
 LogLevel debug
```

and after `ddev restart` a connection to the named host was done with LogLevel debug.

If you want to get fancier, you could add a pre-start hook that copied the config from ~/.ssh into .ddev/homeadditions/.ssh

damienmckenna:
Testing through it some more, just having the config present doesn't work as the config points to specific key files that aren't available...

So what does `ddev auth ssh` actually do? I'll have to dig into the code a little.

rfay:
`ddev auth ssh` runs an ssh-agent, it does not bring the keys into the container or copy them.

You *can* copy the keys into the container with the same strategy, just copy your whole .ssh directory into .ddev/homeadditions in a pre-start exec-host command.

damienmckenna:
Thanks for your suggestions. I was able to solve the problem by copying the `~/.ssh` directory to `.ddev/homeadditions/.ssh` and then it worked.

rfay:
Glad it's working!

tobiasbaehr:
I stumbled into that trap too.

[...]

PR #2224, opened 2020-04-27
Improve ssh-key usage doc

tobiasbaehr:
## The Problem/Issue/Bug:

Without that ssh-keys are physically inside the web-container, the ssh-agent can not found the keys.

## Related Issue Link(s):

#2132

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -619,7 +619,7 @@ The `ddev ssh` command will open an interactive bash or sh shell session to the 
 If you want to use your personal ssh keys within the web container, that's possible. 
 
 * copy the required keys to .ddev/homeadditions/.ssh, and a prepared SSH-Config (.ddev/homeadditions/.ssh/config), see [In-Container home directory configuration](users/extend/in-container-configuration.md)
-* restart via `ddev restart` so that ddev can copy them into the ssh-agent service
+* restart via `ddev restart` so that ddev can copy them into the web service
 * finaly use `ddev auth ssh` to add the keys from your ~/.ssh directory and provide a passphrase
 
 And then those keys will be usable from within the web container. You generally only have to `ddev auth ssh` one time per computer reboot.
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -616,7 +616,13 @@ Commands can also be executed using the shorter `ddev . <cmd>` alias.
 
 The `ddev ssh` command will open an interactive bash or sh shell session to the container for a ddev service. The web service is connected to by default. The session can be ended by typing `exit`. To connect to another service, use the `--service` flag to specify the service you want to connect to. For example, to connect to the database container, you would run `ddev ssh --service db`. To specify the destination directory, use the `--dir` flag. For example, to connect to the database container and be placed into the `/home` directory, you would run `ddev ssh --service db --dir /home`.
 
-If you want to use your personal ssh keys within the web container, that's possible. Use `ddev auth ssh` to add the keys from your ~/.ssh directory and provide a passphrase, and then those keys will be usable from within the web container. You generally only have to `ddev auth ssh` one time per computer reboot.
+If you want to use your personal ssh keys within the web container, that's possible. 
+
+* copy the required keys to .ddev/homeadditions/.ssh, and a prepared SSH-Config (.ddev/homeadditions/.ssh/config), see [In-Container home directory configuration](users/extend/in-container-configuration.md)
+* restart via `ddev restart` so that ddev can copy them into the ssh-agent service
+* finaly use `ddev auth ssh` to add the keys from your ~/.ssh directory and provide a passphrase
+
+And then those keys will be usable from within the web container. You generally only have to `ddev auth ssh` one time per computer reboot.
 
 ### Log Access
 


rfay:
Sorry, keys should generally *not* be physically inside the web container. That's a security issue, that's why the ddev-ssh-agent exists.  Your config files need to be adjusted to refer to the identity, not to the key itself.  As I don't know your exact use-case I can't say exactly what to do, but am happy to help with it.

It's appropriate to copy the config files in. 

`ddev auth ssh` has to be done only once (until `ddev poweroff`, it uses the keys on the host to provide identity management inside the container.

tobiasbaehr:
I have a command which wants to fetch a branch inside the web service to import the database for this task is a ssh-key required. With `ddev auth ssh` you only add them to the agent. (ssh-add -L) But when you to try something which requires a key. Then the agent tries all keys but do not find the key in your home directory.

rfay:
It's working for lots and lots of people accessing things there. 

What do you see when you do `ddev auth ssh` ? Happy to help with this in any of the more interactive support channels as well, https://ddev.readthedocs.io/en/stable/#support-and-user-contributed-documentation

tobiasbaehr:
```
╰─ ddev auth ssh
Identity added: /home/tba/.ssh/key1(/home/tba/.ssh/key1)
Identity added: /home/tba/.ssh/key2 (/home/tba/.ssh/key2)
Identity added: /home/tba/.ssh/key3 (Foobar)
...
```

Inside web service via `ssh-add -l`

```
tba@applications:/var/www/html/web$ ssh-add -l
2048 SHA256:ADFEFDVDVopsvzrKRZ/126197615NovzWdiM /home/tba/.ssh/key (RSA)
...
```

My ssh config is same + 

```
Host *
  UserKnownHostsFile=/home/.ssh-agent/known_hosts
  StrictHostKeyChecking=no
```

rfay:
You don't do an `ssh-add` within the webservice. It's all set up to use the agent, which you've already primed with `ddev auth ssh`.

I may be wrong, but I'm pretty sure your `ssh-add` replaces the keys that have been set up by `ddev auth ssh`

tobiasbaehr:
I know. With `ssh-add -l` I only list the fingerprints to show you, that the keys are added.
But when I just try ssh example.com. SSH tries all keys and ends with Too many authentication failures. Perhaps I have too much keys and the server says, ok, you tried 5 times to connect, now I quit. :D

rfay:
How many keys do you have? I assumed you had three. You know that you can use a single key in most cases right? I use one for all servers and most other situations.

Does this article help you? https://www.tecmint.com/fix-ssh-too-many-authentication-failures-error/

tobiasbaehr:
6. With IdentitiesOnly ssh skips the agent.But I found a host where it was possible.

So the solution is: do not add all keys when you more then 3? keys.

rfay:
I think the issue is how many identities the *server* is willing to let the client try. 

I just stopped after one google on "ssh Too many authentication failures", and there were lots of other good resources there.

rfay:
This is an interesting problem, I'll try experimenting with it after a bit. In the future though, if you have the option, try using a single key rather than generating multiple private keys.

rfay:
I used this test situation. 

1. I created a directory `~/tmp/sshdir` with 10 keys in it: `mkdir ~/tmp/sshdir && cd ~/tmp/sshdir && for item in 1 2 3 4 5 6 7 8 9 10; do ssh-keygen -f $PWD/$item; done`
2. I added those keys with `ddev auth ssh`: `ddev auth ssh -d ~/tmp/sshdir`
3. I set up a server's /etc/ssh/sshd_config with default config `MaxAuthTries 6`

Now I can recreate your situation; only 6 of my 10 keys get tried, `ssh -v <myhostname.mydomain`

```yaml
debug1: Authentications that can continue: publickey
debug1: Next authentication method: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:LrokWMbl1bD0vV0z7Qpn4HLd168NYSIAbqsek6aXIaE agent
debug1: Authentications that can continue: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:ecpRhfcaRWS8EfmYyLuJ81ayhyPWAZd9MG3mKOUKMqA agent
debug1: Authentications that can continue: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:07LrVlDSWu4r+4Eb6WP8FpWYYcREw7IcGm4rtp5v+Ws agent
debug1: Authentications that can continue: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:6L9cIsLlu858CPgb5zZ3v3+5p808uNencyAxJ0S9wOM agent
debug1: Authentications that can continue: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:HwksLkZqEXAK6Zo21+y/C508Mjx2I7EvUQWFScKHsAQ agent
debug1: Authentications that can continue: publickey
debug1: Offering public key: rfay@rfay-mbp-2017.local RSA SHA256:UWDk6gbg2juqMykK0bvJfcSmel4EY/ngNKbX0OygPEM agent
Received disconnect from 174.127.116.75 port 22:2: Too many authentication failures
Disconnected from 174.127.116.75 port 22
```

rfay:
I agree that as far as I can tell, there's no way to specify a key that's provided from an agent, as opposed to the physical key. I tried a number of options for IdentityFile, which seems to be the only selector, and which seems to need to be an actual file, not  a hash or anything else.

Mostly, you'll note that the default server configuration `MaxAuthTries 6` implies that the designers of sshd would be fairly surprised to see you having more private keys than that. 

There's an easy workaround for this that may be useful to you.

Create a directory, maybe ~/ddev-ssh-keys or something. In that directory, either copy or symlink the 6 keys you use most often (assuming 6 is the limit you're running up against). So `cd ~/ddev-ssh-keys && for item in goodkey1 goodkey2 ... googdkey6; do ln -s ~/.ssh/$item; done` (or any way you want to accomplish it. 

Now `ddev auth ssh -d ~/ddev-ssh-keys` and the ddev-ssh-agent will only have those 6 keys. If they're the right keys to solve most of your problems, you may be OK.

Another simple approach if you really want the keys in the container (as opposed to using the agent), then `mkdir -p .ddev/homeadditions/.ssh && cp ~/.ssh/* .ddev/homeadditions/.ssh && chmod 700 .ddev/homeadditions/.ssh && chmod 600 .ddev/homeadditions/.ssh/*`. You can then use the .ddev/homeadditions/.ssh/config file any way you want, including specifying keys.

tobiasbaehr:
That would be my workaround to create a directory only with my personal key. So for me its done. I know now what was my problem. ;-)

rfay:
I added this as a stack overflow question/answer, https://stackoverflow.com/questions/61485061/when-i-try-to-use-ssh-in-ddev-web-container-after-ddev-auth-ssh-the-ssh-keys/61485062#61485062

Issue #2232, opened 2020-05-01
Bump mariadb default install to 10.3.x


damienmckenna:
Drupal 9 requires mariadb 10.3.7, the default 10.2.31 doesn't work.

Given there is likely not much harm in doing so, I think it would be useful to bump the default mariadb to 10.3.

rfay:
We'd need to know what CMSs this might affect before doing it. 10.2 has been great for everything, until Drupal 9 decided at the last minute that 10.3 was a requirement.

Would 10.3 break any versions of TYPO3? WordPress? Magento? Magento2? Drupal 6 or 7? 

If we don't know, I don't really see the reason to move on and change the default.

TYPO3 v7 and below have long had trouble with MariaDB 10.2, so those folks always have trouble and have to use 10.1. And, of course, the results are very obscure and only a person in the know would ever be able to "read" them.

rfay:
For the reasons above (https://github.com/drud/ddev/issues/2232#issuecomment-622553095) this isn't going to happen at this point.

mvance:
I just finished installing [ddev on a Chromebook](https://ddev.com/ddev-local/ddev-docker-chromebook/); thanks @rfay!

I ended up here because I ran the `ddev config --project-type=drupal8 --docroot=web --create-docroot` command but surprisingly the site ended up being Drupal 9.0.2 instead of D8 and it won't let me complete the install process due to the Mariadb 10.3.7 requirement.

I thought I'd mention it here because the out-of-the-box experience for someone new to ddev is a bit confusing due to this issue.

rfay:
Thanks for pointing out the problem with the article, @mvance. I'm confused why you're commenting here I guess, but I've just updated the ddev-on-chrome article to use the new (post-Drupal9-release) recipe. 

When Drupal 9 was released, it started being the default target of `composer create-project drupal/recommended-project`, so this is just a matter of time moving along in the tech world.

Since articles like this can always get out of date, it's important to look at the docs, especially the quickstarts, for the current best practice. See the [drupal 8 quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-8-quickstart) and [drupal 9 quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-9-quickstart)

mvance:
Wow, thanks for the quick reply. I was just coming back to comment that using the following allowed me to install D8:

    ddev composer create drupal/recommended-project:8.9.2

[...]

PR #2233, opened 2020-05-01
Bump default mariadb to 10.3

damienmckenna:
This solves issue #2232 by bumping the default mariadb to 10.3.


rfay:
Sorry, I don't think I'm willing to do this at this point, since it could affect so many people in different situations.

rfay:
But I do appreciate the contribution!

PR #2234, opened 2020-05-04
Support zsh, fish, and Powershell autocompletion, fixes #327

### LINKED ISSUE #327: ddev should support zsh autocomplete

* Feature request.

* We support bash autocomplete https://github.com/drud/ddev/issues/125.
* We're waiting on Cobra to support zsh https://github.com/spf13/cobra/issues/107 before we proceed on this.

* https://github.com/drud/ddev/issues/125.

rfay:
## The Problem/Issue/Bug:

#327 requests zsh autocomplete.

- [x] Generate the autocomplete scripts (bash, fish, zsh, powershell)
- [x] Include them in artifacts generation (CircleCI)
- [x] In install script and homebrew and AUR and Windows installer put them where they belong
- [ ] Docs for installation of each autocomplete (including manual bash autocomplete)
- [x] homebrew: Install _ddev in brew's `/usr/local/share/zsh/site-functions`

- [ ] With zsh, follow the instructions in the docs to install ddev completions.
- [ ] With zsh, run the install script (install_ddev.sh) and verify that the zsh instructions there work.
- [ ] Test out autocompletion with lots of ddev commands and flags. (They may not all work.)

## Release/Deployment notes:

- [x] homebrew install output needs to be updated to explain how to install _ddev zsh autocomplete
- [x] Check AUR path (I think it just uses install_ddev.sh) and see if output is adequate.


rfay:
This zipfile contains _ddev, the zsh autocomplete file. To test, please unzip it into your local $fpath, then `rm -f ~/.zcompdump && compinit`. You should then have zsh autocompletion working.

[ddev_zsh_autocompletes.zip](https://github.com/drud/ddev/files/4575934/ddev_zsh_autocompletes.zip)

rfay:
Autocompletions for fish, zsh, bash, and PowerShell are in this zipfile:

[completions.zip](https://github.com/drud/ddev/files/4576027/completions.zip)

NBZ4live:
Maybe it's the best to do it like in Kubernetes?
https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion

In short:

Instead of generating the files in CI and copy them around and dealing with differences in OS and Linux Distro placing K8S provides a `kubectl completion` command which accepts the shell type as a parameter.

This way for bash you do:
`echo 'source <(kubectl completion bash)' >>~/.bashrc`

For Zsh:
`echo 'source <(kubectl completion zsh)' >>~/.zshrc`

rfay:
That's what we actually do for generation; we have a separate executable (in this PR) that generates the completions. It actually uses the same generating code that kubectl does. The problem here is that you can't just add completions to bashrc or .zshrc, that's not where they belong. They're huge, and it would make an enormous mess of .bashrc or .zshrc.  And I don't even understand why they suggest changing anything in .bashrc, since it should already be set up on macOS (and usually on Linux) to source the correct directory. 

Also, if you read closely in the kubectl instructions, you'll see that they require that you install bash 4, which of course changes the behavior of many things on macOS and would be unacceptable. 

Anyway, it's a morass :) But we'll get through it.

PR #2236, opened 2020-05-04
Encourage people to use Linefeeds in commands, fixes #2126

### LINKED ISSUE #2126: Add .gitattributes to .ddev/commands to keep CRLF out of commands

Windows users who create custom commands often run into a failure like this when they run the command:

```
bash: /mnt/ddev_config/commands/web/artisan: /bin/bash^M: bad interpreter: No such file or directory
Failed to run artisan : exit status 126
```

This is caused by CRLF in the custom command script, and bash never ever will accept that. 

Mostly people need to use LF only in their editor, and that's an easy way to avoid this. But we can add a .gitattributes to .ddev/commands, something like this:

```
#ddev-generated
# Make sure that all command scripts don't get CRLF in them
* -text eol=lf
```

rfay:
## The Problem/Issue/Bug:

Windows users are perpetually discovering the hard way that ddev custom commands have to have linefeeds, not CRLF, because bash can't handle CRLF.

1. Warn people and skip over commands that have CRLF
2. Add .ddev/commands/.gitattributes to coax git into fixing them maybe.

* Create a command with CRLF in it.
* Use `ddev` any way you want. It will nag you to death.


rfay:
Tested on Windows with a CRLF file and seems to work as expected.

PR #2242, opened 2020-05-07
fix(nginx): Remove duplicated image location

klausi:
The `location ~* \.(js|css|png|jpg|jpeg|gif|ico|svg)$ {` entry is duplicated in the drupal7 nginx config.


rfay:
Thanks! This may sit for a while, as the nginx configs in general need to be redone, likely in this release cycle.

PR #2246, opened 2020-05-11
Only use //c on Docker Toolbox, not Docker Desktop for Windows

rfay:
## The Problem/Issue/Bug:

For Docker Toolbox we had a path massaging function that changed C:\Users\rfay\workspace to //c/Users/rfay/workspace. This // really helped toolbox work.

But now in Docker 2.3.0.1 (edge) and 2.3.0.2 (released today as stable) it doesn't work any more on Docker Desktop for Windows.

Only add the extra slash on the front when running Docker Toolbox.

`ddev auth ssh` with Docker Desktop for Windows 2.3.0.2


gnuget:
Thanks a lot!

rfay:
Thank you for the report. Hope to have a point release out by tomorrow.

rfay:
Actually, I'm out tomorrow and it needs to have docker upgraded all the test runners for a final go, so probably day after tomorrow.

Issue #2084, opened 2020-02-12
'php' project type uses nginx config derived from Drupal nginx config, should be more generic - Laravel Conflict


jdoubleu:
**Describe the bug**
Nginx inside the webserver is [adding the requested relative URL as `q` query parameter](https://github.com/drud/ddev/blob/master/containers/ddev-webserver/files/etc/nginx/nginx_default.conf#L14) to the request which is then passed to PHP (fcgi).
This parameter appears in the global `$_GET` PHP array.

Since I am using Laravel with DDEV this pollutes Laravel's signature verification. The signature verification functions [collects all query params from the current request](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Routing/UrlGenerator.php#L383) and verifies it against the previously calculated signature. However this current request includes the `q` query param, while the precalculated signature doesn't include it.

**To Reproduce**
1. Create an empy ddev project with the PHP configration.
2. Create a `index.php` in the `public` folder and add a debug breakpoint.
3. Open the project in your browser and append an arbitraty path (e.g. `myproject.ddev.site/some/path/`).
4. See that the global `$_GET` array contains `q=some/path/`.

**Expected behavior**
There are no additional query params in the `$_GET` or `$_REQUEST` array other than the user-provided ones.

**Version and configuration information (please complete the following information):**
 - ddev version information: v1.12.1

**Screenshot**
![Screenshot 2020-02-12 at 16 22 37](https://user-images.githubusercontent.com/11274444/74349068-fe59bc80-4db3-11ea-8848-d63915d38868.png)

**Additional context**
As mentioned before, this confuses Laravel's signature validation middleware which is used for email verification: https://github.com/laravel/framework/issues/30594#issuecomment-585235650.

rfay:
This is likely a result of the default nginx configuration being derived from a default Drupal configuration.

Probably:
* You should use a custom nginx configuration (https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-nginx-configuration) - you may want to replace the whole thing, referred to as "legacy"
* ddev should probably use a more generic configuration for "php" projects. If you get yours working, we'll consider doing that.

jdoubleu:
Unfortunately it is not possible to override the `location` directive to remove the additional query parameter.

I'd have to replace the whole site configuration which is deprecated, as you said.

I'm trying to create a generic nginx configuration for PHP projects in general and move the current customizations to the Drupal configs?

[...]

Issue #898, opened 2018-05-31
Add support for Laravel (First step: Try it out as PHP)


cweagans:
**Is your feature request related to a problem? Please describe.**

Laravel doesn't have explicit support in ddev. We should consider adding it.

**Describe the solution you'd like**
We can template out the config/database.php file with the appropriate credentials added, but this would be a fairly undesirable solution in the laravel community. If you look at [that file](https://github.com/laravel/laravel/blob/master/config/database.php), it's reading credentials from the environment variables `DB_HOST`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD`, and `DB_DATABASE`. If we set those environment variables, Laravel should "just work".

We could also write out a .env file ([example](https://github.com/laravel/laravel/blob/master/.env.example)) with our creds prepopulated.

**Describe alternatives you've considered**

**Additional context**

dapseen:
+1

beeradb:
Any thoughts about what would make this more attractive than https://laravel.com/docs/5.6/homestead ?

I'd taken a look at adding this a year or so ago but ultimately became bearish because we didn't have contacts in the laravel community and homestead exists and looks to be officially supported.

cweagans:
Homestead is a gigantic Vagrant box. Personally, I'm pretty happy with the idea of never installing Virtualbox again :) There's also Laravel Valet, which is another officially supported solution for macos, but I had lots of problems with it.

alkymst:
I think the argument is that agencies will do some WP, some Drupal, and a few Laravel sites. So one platform to manage it all.

alkymst:
I'm first just going to take it for a test ride and document getting it spun up with generic PHP.

rfay:
Moved to v1.1.0 release since this isn't going to get attention on this round.

rickmanelius:
Need @alkymst time on this, but I know he's tied up at the moment. Same is true for GravCMS, which we've had at least 3 or 4 requests or inquiries about.

bessone:
Some update on "official" Laravel support or is the idea dead?

rfay:
"Official" support just means customizing settings files and sometimes nginx/apache config. So the reality is it's mighty easy to work with any PHP framework just using the 'php' project type. A few people have done that, but we haven't had a lot of activity here as you can see, so it hasn't made the priority list.

unherz:
@rfay Can we start to work on it and open a PR or is this task already assigned and planned for a future release?

[...]

PR #2249, opened 2020-05-13
Add Laravel support

NBZ4live:
## The Problem/Issue/Bug:
There is currently no official Laravel support in form of a project-type.
https://github.com/drud/ddev-contrib/tree/master/recipes/laravel suggests to use the generic `php` project type, but it's Nginx config is more Drupal inspired than generic (for example the `try_files $uri $uri/ /index.php?q=$uri&$args;`)

This PR does add the `laravel` project type and official Laravel config for Nginx.
Additionally, in the `postStartAction`, it does check if the the ddev db container is configured in the .env and prints some helpful instructions how to do it with a one-liner.

How I tested:

1. Create a new Laravel app same as I described in the quickstart docs:
```
mkdir my-laravel-app
cd my-laravel-app
ddev config --project-type=laravel --docroot=public --create-docroot
ddev start
ddev composer create --prefer-dist laravel/laravel
ddev exec "cat .env.example | sed  -E 's/DB_(HOST|DATABASE|USERNAME|PASSWORD)=(.*)/DB_\1=db/g' > .env"
ddev exec "php artisan key:generate"
ddev launch
```

2. Clone any working Laravel demo app from GitHub. (I used the Backpack demo):
```
git clone git@github.com:Laravel-Backpack/demo.git laravel-demo
cd laravel-demo
ddev config --project-type=laravel
ddev composer install
ddev exec "cat .env.example | sed  -E 's/DB_(HOST|DATABASE|USERNAME|PASSWORD)=(.*)/DB_\1=db/g' > .env"
ddev exec "php artisan key:generate"
ddev exec "php artisan migrate"
ddev exec "php artisan db:seed"
ddev launch
```

Will add some demo app to ddevapp_test.go later. Will probably need to write one, because using some random demo app from GitHub for this purpose does feel wrong.

## Related Issue Link(s):
#898, #2084

## Release/Deployment notes:
No. This should not affect any other project types

PS: I did add a wrapper for artisan (Laravel's cli), but decided to do this in a separate branch, because it's first time I do something in Go and I need to study the codebase a bit more to be able to make a test for it. (And at the moment I have no idea how to test the wrapper for a command which does exist only inside a Laravel project codebase)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ ddev is an open source tool that makes it simple to get local PHP development en
     * [TYPO3 Quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage#typo3-quickstart)
     * [Magento 1 Quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage#magento-1-quickstart)
     * [Magento 2 Quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage#magento-2-quickstart)
+    * [Laravel Quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage#laravel-quickstart)
 
 Having trouble? See our [support options below](#support). You might have trouble if [another local development tool is already using port 80 or 443](https://ddev.readthedocs.io/en/stable/users/troubleshooting/#unable-listen).
 
--- a/containers/ddev-webserver/files/etc/nginx/nginx-site-laravel.conf
+++ b/containers/ddev-webserver/files/etc/nginx/nginx-site-laravel.conf
@@ -1,0 +1,42 @@
+# ddev laravel config
+
+# You can override ddev's configuration by placing an edited copy
+# of this config (or one of the other ones) in .ddev/nginx-site.conf
+# See https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-nginx-configuration
+
+# Set https to 'on' if x-forwarded-proto is https
+map $http_x_forwarded_proto $fcgi_https {
+    default off;
+    https on;
+}
+
+server {
+    listen 80; ## listen for ipv4; this line is default and implied
+
+
+    # The WEBSERVER_DOCROOT variable is substituted with
+    # its value when the container is started.
+    root $WEBSERVER_DOCROOT;
+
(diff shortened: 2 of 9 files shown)


rfay:
Yay!

rfay:
@NBZ4live I took the liberty of rebasing this, so you'll need to update your local branch against the github fork.  Hope that's OK. But I know you were concerned about the broken Windows container test.

I also pushed a branch-specific version of ddev-webserver and added that into version.go, so you should be able to use your built ddev as is.

rfay:
Looks like you're right on track here. 

The key thing that needs to be added is a set of data to test against, as in https://github.com/drud/ddev/blob/f6e43e080196e5ade4f1e8f878d3bf3645878457/pkg/ddevapp/ddevapp_test.go#L41-L166

NBZ4live:
Yes, I know. As I wrote in the PR, I will need to writ an app because I think taking something random from GitHub doesn't feel to be right.

But I will be able to do this only on the weekend.

rfay:
Ah, thanks. I didn't understand the context of your comment about "writing an app", had thought you were going to do more go test writing than was necessary. However, just simple codebase built with `ddev composer create laravel/laravel` probably wouldn't do any harm.  But that wouldn't give much opportunity for actual test URLs, so I think your plan is better. Thanks!

NBZ4live:
This would give test urls, but would not give a proper dbimport test

rfay:
Yes, agreed. It's best to have real URLs that depend on a matching database.

rfay:
I'll be back to review this hopefully in the next week. It will also have to get moved probably to a new repo (drud/ddev-images), but I can take care of that.

rfay:
To be more specific, here's the pages admin at http://laravel-backpack-demo.ddev.site/admin/page:

Pages____Backpack_Admin_Panel

If I click where the arrow is, it takes me to http://laravel-backpack-demo.ddev.site/mollitia, which is a 404. 

If I go to News->Articles, I'm able to preview the items there, but don't actually know how to see if they're really published or if the published URL works.

rfay:
@NBZ4live confirmed in slack that the behavior here is correct.

PR #2258, opened 2020-05-15
Add composer bin directory to PATH in ddev-webserver

atomicptr:
## The Problem/Issue/Bug:

Most ddev projects probably use composer and many will have tools installed with composer. Sadly right now in order to execute them you have to do something like this:

```bash
$ ddev exec ./vendor/bin/tool --parameter
```

By adding the vendor/bin directory to the container PATH you can access the script directly

```bash
$ ddev exec tool --parameter
```

Steps to test if this works (assuming you're using a built of ddev/containers from my PR)

```bash
$ mkdir test-this-change
$ cd test-this-change
$ composer req typo3/minimal helhum/typo3-console
# ... wait a moment
$ ddev config
# all the defaults
# now you'd need to add the modified container to .ddev/config.yml
# webimage: local-build-of-ddev-webserver
$ ddev start
# ... wait a moment
$ ddev exec typo3cms cache:flush
Flushed all file caches.
```

I couldn't think of a better example :).

## Release/Deployment notes:

This won't affect non composer projects at all, because directories that don't exist don't cause any issues when inside the PATH variable. If for some reason the composer bin directory is somewhere else you can apply that to the project by adding an environment variable COMPOSER_BIN_DIR ( which is also an option in composer https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin- ) to the ddev configuration.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -25,6 +25,9 @@ ENV COMPOSER_ALLOW_SUPERUSER 1
 ENV COMPOSER_CACHE_DIR /mnt/ddev-global-cache/composer
 # Windows, especially Win10 Home/Docker toolbox, can take forever on composer build.
 ENV COMPOSER_PROCESS_TIMEOUT 2000
+# add composer bin dir to PATH
+ENV COMPOSER_BIN_DIR $WEBSERVER_DOCROOT/vendor/bin
+ENV PATH "$COMPOSER_BIN_DIR:$PATH"
 
 ENV TERMINUS_CACHE_DIR=/mnt/ddev-global-cache/terminus/cache
 
@@ -54,7 +57,7 @@ RUN set -o errexit && apt-get -qq update && \
         gnupg \
         jq \
         locales-all \
-		lsb-release && \
+        lsb-release && \
     wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg && \
     echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/php.list && \
     wget -q -O - https://packages.blackfire.io/gpg.key | apt-key add - && \
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -39,6 +39,11 @@ ENV DDEV_LIVE_DOWNLOAD_URL https://downloads.ddev.com/ddev-live-cli/latest/linux
 ENV DDEV_LIVE_CONFIG_FILE_PATH /mnt/ddev-global-cache/ddev-live/cli-config.json
 ENV DDEV_LIVE_NO_VERSION_PROMPT true
 
+# From "man bash":
+#    When bash is started non-interactively, to run a shell script, for example,
+#    it looks for the variable BASH_ENV in the environment, expands its value
+#    if it appears there, and uses the expanded value as the name of a file to read and execute.
+ENV BASH_ENV /etc/bash.nointeractive.bashrc
 
 RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime && dpkg-reconfigure --frontend noninteractive tzdata
 
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -25,9 +25,6 @@ ENV COMPOSER_ALLOW_SUPERUSER 1
 ENV COMPOSER_CACHE_DIR /mnt/ddev-global-cache/composer
 # Windows, especially Win10 Home/Docker toolbox, can take forever on composer build.
 ENV COMPOSER_PROCESS_TIMEOUT 2000
-# add composer bin dir to PATH
-ENV COMPOSER_BIN_DIR $WEBSERVER_DOCROOT/vendor/bin
-ENV PATH "$COMPOSER_BIN_DIR:$PATH"
 
 ENV TERMINUS_CACHE_DIR=/mnt/ddev-global-cache/terminus/cache
 
@@ -41,6 +38,7 @@ ENV CAROOT /mnt/ddev-global-cache/mkcert
 ENV DDEV_LIVE_DOWNLOAD_URL https://downloads.ddev.com/ddev-live-cli/latest/linux/ddev-live.zip
 ENV DDEV_LIVE_CONFIG_FILE_PATH /mnt/ddev-global-cache/ddev-live/cli-config.json
 ENV DDEV_LIVE_NO_VERSION_PROMPT true
+
 
 RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime && dpkg-reconfigure --frontend noninteractive tzdata
 
@@ -57,7 +55,7 @@ RUN set -o errexit && apt-get -qq update && \
         gnupg \
         jq \
         locales-all \
-        lsb-release && \
+		lsb-release && \
     wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg && \
     echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/php.list && \
     wget -q -O - https://packages.blackfire.io/gpg.key | apt-key add - && \
--- a/containers/ddev-webserver/files/etc/skel/.bashrc
+++ b/containers/ddev-webserver/files/etc/skel/.bashrc
@@ -111,3 +111,11 @@ if ! shopt -oq posix; then
     . /etc/bash_completion
   fi
 fi
+
+# set COMPOSER_BIN_DIR to $WEBSERVER_DOCROOT/vendor/bin if unset
+if [[ ! -v $COMPOSER_BIN_DIR ]]; then
+    export COMPOSER_BIN_DIR="$WEBSERVER_DOCROOT/vendor/bin"
+fi
+
+# add composer bin dir to PATH
+export PATH="$COMPOSER_BIN_DIR:$PATH"
--- a/containers/ddev-webserver/files/etc/skel/.bashrc
+++ b/containers/ddev-webserver/files/etc/skel/.bashrc
@@ -112,10 +112,4 @@ if ! shopt -oq posix; then
   fi
 fi
 
-# set COMPOSER_BIN_DIR to $WEBSERVER_DOCROOT/vendor/bin if unset
-if [[ ! -v $COMPOSER_BIN_DIR ]]; then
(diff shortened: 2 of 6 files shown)


rfay:
vendor/bin is commonly linked with the Drupal recipes, but not for everybody else. 

I'm sure you know you can add to the PATH by adding .ddev/homeadditions/.profile with what you want, which would solve your short-term problem. 

I think this would be most easily accomplished by just adding vendor/bin to the PATH in https://github.com/drud/ddev/blob/a0e2d30d42e4b1febb7c962c56b86d0a3f31bb3d/containers/ddev-webserver/files/etc/skel/.profile

atomicptr:
> I'm sure you know you can add to the PATH by adding .ddev/homeadditions/.profile with what you want, which would solve your short-term problem.

I wouldn't go so far to call it a problem more an inconvenience, in my opinion having the composer bin dir in the PATH is a sensible default.

> I think this would be most easily accomplished by just adding vendor/bin to the PATH in https://github.com/drud/ddev/blob/a0e2d30d42e4b1febb7c962c56b86d0a3f31bb3d/containers/ddev-webserver/files/etc/skel/.profile

I picked the Dockerfile for registering COMPOSER_BIN_DIR mostly to make it changeable, tho I agree that the /etc/skel/.profile file is probably better for adding it to the PATH.

I'll change this in a moment...

containers/ddev-webserver/Dockerfile
```diff
@@ -25,6 +25,7 @@ ENV COMPOSER_ALLOW_SUPERUSER 1
 ENV COMPOSER_CACHE_DIR /mnt/ddev-global-cache/composer
 # Windows, especially Win10 Home/Docker toolbox, can take forever on composer build.
 ENV COMPOSER_PROCESS_TIMEOUT 2000
+ENV COMPOSER_BIN_DIR $WEBSERVER_DOCROOT/vendor/bin
```
rfay, line 28: I'd rather see all the changes just be in the skel .profile. I don't think there's any reason not to.

atomicptr:
Something I didn't realize earlier, the /etc/skel/.profile is only evaluated when you use "ddev ssh"

For instance:

```bash
$ ddev exec env |grep PATH            
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
$ ddev ssh
(inside the container) $ env |grep PATH
PATH=/var/www/html/vendor/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

rfay:
Shoot, you're right. Please experiment with adding the path in the .bashrc, but I'm not sure that will work either, because `ddev exec` is just an invocation of `bash ...` 

It also wouldn't work with the few people who switch to zsh.

If .bashrc doesn't work, we'll have to switch to your original approach if we don't come up with another idea.

atomicptr:
Adding this to the bashrc has as far as I can see the exact same effect, works as soon as you're in the container but not from the outside with exec

atomicptr:
I rebased onto master and changed it back to the prior Dockerfile solution

rfay:
I studied this a little bit today, and it looks like if we wanted to use the "normal" techniques (profile, bashrc, /etc/profile, /etc/environment, /etc/bash.bashrc, then ddevapp.Exec would have to use `bash -ic` instead of just `bash -c`, which doesn't read any config files. But I wouldn't count on that not breaking a thing or two, but it's tempting. https://github.com/drud/ddev/blob/bb80820bfa3e2d65e6ed1790ec82b1b942e05c24/pkg/ddevapp/ddevapp.go#L975

If you'd like to actually test this, you'll need to push an image that matches, or at least test one locally. 

I also pushed a PR with more information about building images, https://github.com/drud/ddev/pull/2264 - would love to have you take a look and see what more you need to know.

atomicptr:
This seems to work:

```bash
$ ./local-ddev-build exec env |grep PATH
PATH=/var/www/html/vendor/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
$ ddev exec env |grep PATH
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

I'll push the changes in a bit

atomicptr:
The tests are quite red now :D.

rfay:
`make push` and `make container` work in the containers/ddev-webserver directory, not at the root. I just added more on that and on automated tests to the dev docs PR. 

Wow, your failure is one I've never seen before, in TestCmdCustomCommands  : "bash: cannot set terminal process group (-1): Inappropriate ioctl for device\nbash: no job control in this shell\n99\n99\n"
 
Ah, the problem is you used incorrect syntax. With bash '-c' takes an argument, so the command should be '-ic', not '-ci'. The way you have it, there is no argument to the 'c'. It could also be "-i -c"

atomicptr:
Ok, weird that this #WorksOnMyMachine haha. I'll change that in a sec.

containers/ddev-webserver/files/etc/skel/.bashrc
```diff
@@ -111,3 +111,11 @@ if ! shopt -oq posix; then
...
   fi
 fi
+
+# set COMPOSER_BIN_DIR to $WEBSERVER_DOCROOT/vendor/bin if unset
+if [[ ! -v $COMPOSER_BIN_DIR ]]; then
+    export COMPOSER_BIN_DIR="$WEBSERVER_DOCROOT/vendor/bin"
+fi
+
+# add composer bin dir to PATH
+export PATH="$COMPOSER_BIN_DIR:$PATH"
```
rfay, line 121: This should not be at the beginning of the $PATH, it should be at the end IMO.

atomicptr: IMHO it should be at the start, because if there is an executable in both directories with the same name you explicitly installed it to be used instead of whatever is in the rest of the path.

rfay: The problem is that you're changing behavior for people that way. For example, on a Drupal 8 site you'll get vendor/bin/drush instead of /usr/local/bin/drush. I'm more or less OK with this PR if we don't create any surprises for people, but not OK with it if it changes behavior.

containers/ddev-webserver/files/etc/skel/.bashrc
```diff
@@ -111,3 +111,11 @@ if ! shopt -oq posix; then
     . /etc/bash_completion
   fi
 fi
+
+# set COMPOSER_BIN_DIR to $WEBSERVER_DOCROOT/vendor/bin if unset
+if [[ ! -v $COMPOSER_BIN_DIR ]]; then
```
rfay, line 116: I haven't used `-v` before. I would have expected `if [ -d ${COMPOSER_BIN_DIR} ]` here?

atomicptr: The intention was: If the variable is unset (so that users could still use a different directory for the composer bin dir) set it to a sensible default (/var/www/html/vendor/bin).

rfay: I think `-z` is the normal usage?

atomicptr: There are quite a few ways to do this tbh, -v Is just what I usually do as it just checks if a variable is set.

```
-v varname

    True if the shell variable varname is set (has been assigned a value).
```

( From https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html )

atomicptr:
I found an IMHO better solution by using BASH_ENV for the non interactive shells and adding a seperate file which will be loaded by it.

rfay:
I'll be back to review this hopefully in the next week. It will also have to get moved probably to a new repo (drud/ddev-images), but I can take care of that.

containers/ddev-webserver/Dockerfile
```diff
@@ -35,6 +35,11 @@ ENV APACHE_SITE_VARS '$WEBSERVER_DOCROOT'
 
 ENV CAROOT /mnt/ddev-global-cache/mkcert
 
+# From "man bash":
+#    When bash is started non-interactively, to run a shell script, for example,
+#    it looks for the variable BASH_ENV in the environment, expands its value
+#    if it appears there, and uses the expanded value as the name of a file to read and execute.
+ENV BASH_ENV /etc/bash.nointeractive.bashrc
```
rfay, line 42: One thing we could do that would make a lot of sense is to set BASH_ENV to /etc/profile or /etc/bash.bashrc.  Then questions like [this one](https://stackoverflow.com/questions/58415512/how-can-i-add-and-use-nvm-in-a-ddev-web-container/61934500?noredirect=1#comment109655087_61934500) wouldn't have to happen.

atomicptr: Dunno, /etc/profile will not be loaded as exec isn't a login shell and /etc/bash.bashrc is as stated in the file 

```
# System-wide .bashrc file for interactive bash(1) shells.
```

for **interactive bash shells** which would in my opinion make this a lot more confusing. (It also probably won't be loaded either at that point).

> Then questions like this one wouldn't have to happen.

Putting something else to the path (or generally bashrc modifications) for both interactive and non-interactive shells can be done by putting something into the /etc/bashrc (maybe a better name is necessary?) directory. Which in my opinion is a fairly clean solution to this issue, the only thing that we absolutely need is specifying the BASH_ENV somewhere and the Dockerfile is so far the best place for this (IMHO).

rfay: What I'm saying is use /etc/profile or /etc/bash.bashrc as the BASH_ENV (and keep them simple). So BASH_ENV=/etc/bash.bashrc would or BASH_ENV=/etc/profile would be the most likely to give people the path and other variables that they expect. It's the same as the solution you have here, but doesn't require an extra file to maintain, and changes that get put into one of those files (we'd have to choose which) would end up in `ddev exec` environment.

atomicptr: If you prefer it that way we can obviously do that, but I'm not sure if that doesn't cause confusion as the /etc/bash.bashrc is explicitly intended to be used for interactive shells as is /etc/profile for login shells. I think a new file is a lot clearer in the intent of what we're doing, though thats just my opinion.

rfay: It's a reasonable opinion. I just thought about it when answering that Stack Overflow and since we'd been in the territory lately. His question is super similar to what you've been exploring here.

atomicptr: Yes, it's very similar. Technically his issue could be solved by putting a file into /etc/bashrc (with the build of this PR) which adds whatever he needs to the PATH.

PR #2267, opened 2020-05-19
fix(nginx): Fix rewrite location for Drupal 7 to make subfolder setups work

klausi:
The `@rewrite` location does not forward requests correctly to Drupal if it is configured to run in a subfolder. We should just use the same rewrite line as we do in the `/` location, then everything works in both normal and subfolder setups.

### DIFF
--- a/containers/ddev-webserver/files/etc/nginx/nginx_drupal7.conf
+++ b/containers/ddev-webserver/files/etc/nginx/nginx_drupal7.conf
@@ -18,8 +18,8 @@
 
     location @rewrite {
         # For D7 and above:
-        # Clean URLs are handled in drupal_environment_initialize().
-        rewrite ^ /index.php;
+        # Directly rewrite to $_GET['q'] so that subfolder setup rewrites work as well.
+        rewrite ^ /index.php?q=$uri&$args;
     }
 
     # Handle image styles for Drupal 7+
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -45,7 +45,7 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "20200518_add_ddev_live" // Note that this can be overridden by make
+var WebTag = "20200527_klausi_drupal7_subfolder" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"


rfay:
Please add manual testing instructions, thanks. 

I assume you've already tested this with .ddev/nginx-site.conf, per https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-nginx-configuration ?

klausi:
Yes, we tested this with the guide in #2268 .

Manual testing: thinking about some rough steps:

### Test without subfolder (normal setup as DDEV works)
* Checkout drupal 7
* ddev config
* ddev start
* ddev exec drush si -y
* Got to front page in browser, make sure Drupal loads normally
* Log in as admin and go to content admin view, make sure Drupal loads normally

### Test with subfolder
* Checkout drupal 7
* ddev config
* Follow instructions from #2268 (just the settings file, we don't need the full nginx override after this change hopefully)
* create `.ddev/nginx/subfolder.conf`:
```
# Rewrite the /jobs subfolder.
# Special rule for the front page: we need to pass "/" explicitly here to
# $_GET['q'] and PHP. This is not done by Nginx automatically and then
# Drupal falls back to the full request URI which would include the
# subfolder /jobs.
rewrite ^/jobs/?$    /index.php?q=/&$args   last;
rewrite ^/jobs/(.*)$   /$1   last;
```
* ddev start
* ddev exec drush si -y
* Go to page `/` in browser, make sure error page from settings.php appears
* Go to page `/jobs` in browser, make sure Drupal front page loads normally
* Log in as admin and go to content admin view, make sure Drupal loads normally

Will do this testing myself to verify.

klausi:
Confirming the manual testing, everything working in the normal and in the subfolder scenario.

rfay:
I wasn't able to get this to work at all after spending a couple of hours on it, and don't think it makes sense to change the core D7 nginx config for a special case. 

I recommend that you just include the full nginx-site.conf (including the redirects) in your ddev-contrib recipe and I think that will do the job. 

I also hope that we'll soon have a way to easily override the nginx configuration without this much effort, as outlined in https://github.com/drud/ddev/issues/2263

If you'd like to experiment with the artifacts from this build to find out whether it did what you wanted, they're at https://app.circleci.com/pipelines/github/drud/ddev/2210/workflows/c762cd81-c11a-4593-ad20-3d0d8526d0dc/jobs/25119/artifacts
So reluctantly I'm going to close this, but you have an easy and already tested way forward.

klausi:
Oh, sorry about that. Did you get any specific error message?

It is strange because the official Nginx Drupal config also uses a similar rewrite rule 
```
    location @rewrite {
        rewrite ^/(.*)$ /index.php?q=$1;
    }
```
https://www.nginx.com/resources/wiki/start/topics/recipes/drupal/

rfay:
I got 404's no matter what I did. Anyway, this is a special case and best handled with a replacement nginx config. And I expect that custom nginx config will be much, much easier to do in an upcoming ddev version.

PR #2277, opened 2020-05-24
Improve examples and help in several commands

rfay:
## The Problem/Issue/Bug:

Some older commands had no examples at all. Some help text was a bit out of date.

Add examples to most commands.


rfay:
Thanks for taking a look @AlexCreamer ! The "short" text is shown when you just type "ddev", the long text when you type `ddev help <command>` or `ddev command -h`

PR #2282, opened 2020-05-29
Adds ddev-live as a default custom command

unn:
## The Problem/Issue/Bug:
Now that the ddev-live tool is in the web container we should ship a default custom command for `live`

* introduces `live` command resulting in commands `ddev live`


rfay:
One request @unn - Please don't push directly into the drud/ddev repository. It makes a mess. Use your unn/ddev fork and use the forked-pr workflow, thanks.

PR #2295, opened 2020-06-04
Make sure that Exec() runs commands with bash error checking, especially pipefail

rfay:
## The Problem/Issue/Bug:

@edysmp pointed out in https://github.com/drud/ddev/issues/2278#issuecomment-636215134 that `ddev import-db` was failing in the tested Mutagen edge version of Docker, but reporting success. 

The reason this was happening was that in app.Exec bash was being invoked without the `pipefail` option, so a failure in the pipe wasn't caught, and the ddev import-db uses a pipe with pv. 

Make sure that `set -o pipefail` is set

## Release/Deployment notes:

This change needs to be mentioned in release notes, as some people might be counting on failures not being reported.


rfay:
Possible failure on apache fpm windows, https://buildkite.com/drud/ddev-windows-apache-fpm/builds/2630#79b04fb2-0ca4-4cb1-844c-04adbf7bc2fe

PR #2297, opened 2020-06-05
Apache2 config: Enforce file types matching at end of filename only

gilbertsoft:
see https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/Security/GuidelinesAdministrators/Index.html#file-extension-handling
see https://httpd.apache.org/docs/2.4/mod/mod_mime.html#multipleext

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/apache2/apache2.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/apache2/apache2.conf
@@ -185,24 +185,6 @@ AccessFileName .htaccess
 </FilesMatch>
 
 
-# Security: Enforce file types matching at end of filename only
-# see https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/Security/GuidelinesAdministrators/Index.html#file-extension-handling
-# see https://httpd.apache.org/docs/2.4/mod/mod_mime.html#multipleext
-<IfModule mod_mime.c>
-	RemoveType .html .htm
-	<FilesMatch ".+\.html?$">
-		AddType text/html .html
-		AddType text/html .htm
-	</FilesMatch>
-
-	RemoveType .svg .svgz
-	<FilesMatch ".+\.svgz?$">
-		AddType image/svg+xml .svg
-		AddType image/svg+xml .svgz
-	</FilesMatch>
-</IfModule>
-
-
 #
 # The following directives define some format nicknames for use with
 # a CustomLog directive.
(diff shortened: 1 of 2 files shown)


gilbertsoft:
I've tested this change with the .ddev/apache/apache-site.conf but I think it's better to place it in the global configuration.

containers/ddev-webserver/ddev-webserver-base-files/etc/apache2/apache2.conf
```diff
@@ -185,6 +185,24 @@ AccessFileName .htaccess
 </FilesMatch>
 
 
+# Security: Enforce file types matching at end of filename only
```
rfay, line 188: Is it OK that this affects all project types?

gilbertsoft: I'm not 100% sure but this should normally work with all project types. If a project relays on this unsecure default configuration it should be solved there directly. This change makes sure the proper file extension is detected means the last .*

rfay:
Updated artifacts are at https://app.circleci.com/pipelines/github/drud/ddev/2259/workflows/274e99ef-24e2-463c-96e3-34a16a635547/jobs/25356/artifacts

I recommend doing a `docker pull drud/ddev-webserver:20200606_gilbertsoft_harden_apache` to make sure you've got the updated image.

gilbertsoft:
Still the same error
grafik

Will have to dig deeper....

rfay:
I really, really think I'll get ddev changed to *mount* the default nginx config into the container in the next release, rather than building it into the container. That would make working on changes like this enormously easier. If you want to wait on that change it's OK with me.

gilbertsoft:
I'm fine with both, it's not a very important change but more a nice to have. Also next TYPO3 releases should have this included in the .htaccess. This change is more important for production environments and for DDEV only to have a as much as possibe environment similar to the production one.

rfay:
I recommend picking up work on this after https://github.com/drud/ddev/pull/2305 goes in.

rfay:
This is now much easier to do and test, but this PR itself is no longer relevant. 

I'd recommend completely pushing a whole new approach.

Just adjust the .ddev/apache/apache-site.conf to meet the needs here. You'll want to remove the "#ddev-generated" while working with it. Then get it working and add the #ddev-generated back in, the only file that has to be updated is easy to find.

rfay:
I'm going to close this for now, as it needs a do-over with how configs are done now. Happy to re-open this one if you want to work with it, and of course it's fine to do a new one.  It's vastly easier to do a PR like this now because it's easy to test it completely without building containers or referring to them.

PR #2315, opened 2020-06-16
Add support for XDEBUG_HOST_IP to override all other ways of figuring it out

rfay:
## The Problem/Issue/Bug:

There are times (like running PHPStorm on the Windows host but using ddev inside WSL2 environment) when it's useful to override host.docker.internal, the *normal* way you figure out where the xdebug-listening IDE may be.

This adds support for an environment variable that ddev will respect, XDEBUG_HOST_IP.

It should be set with `export XDEBUG_HOST_IP=<ip addr>`. On WSL2 with PHPStorm on the host, `export XDEBUG_HOST_IP=$(awk '/^nameserver/ {print $2; exit;}' </etc/resolv.conf)` will normally do the job

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -705,6 +705,11 @@ func CreateVolume(volumeName string, driver string, driverOpts map[string]string
 func GetHostDockerInternalIP() (string, error) {
 	hostDockerInternal := ""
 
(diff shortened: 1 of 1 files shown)


rfay:
As far as I can tell this isn't actually useful, so closing for now. I wrote up the WSL2 workflow with PHPStorm using ddev v1.14.2, https://ddev.com/ddev-local/ddev-local-and-phpstorm-debugging-with-wsl2/

Issue #1648, opened 2019-06-11
As a developer, I want to run Drush commands as easily with ddev as without (with no quoting issues)


damienmckenna:
**Is your feature request related to a problem? Please describe.**
Currently in order to run Drush commands you have to pass them through the exec system. This runs into problems in 1.8.0 due to the bash execution system as quotation marks and backslashes are mangled as part of the execution. This results in needing to do a really cumbersome combination of quotation marks and backslashes in order to run commands.

With a native system you can run Drush commands like this:
`drush ev "cache_clear_all('something:*', 'cache', TRUE);"`

The current DDEV-Local equivalent would be this:
`ddev . drush ev "cache_clear_all('something:*', 'cache', TRUE);"`

Unfortunately that will result in errors from the underlying system that handles the exec functionality because of unescaped characters.

There should be a way of running Drush commands with ddev that doesn't require either huge mounts of character escaping or ssh'ing to the web instance.

**Describe the solution you'd like**
I'd like to be able to run Drush commands as simply as I can with a local system so that I can do something like this:
`ddev drush ev "cache_clear_all('something:*', 'cache', TRUE);"`

**Describe alternatives you've considered**
I've attempted to work out the appropriate combination of backslashes and quotation marks to use, it turns into a PITA.

DDEV-Local is supposed to be aimed at making life easier, having to run `ddev ssh` doesn't match my concept of "easy" as it moves the user into the virtual machine, removing all of their local aliases, commands, etc.

**Additional context**
n/a

rthideaway:
I know exactly what you're talking about. Eval commands with `drush ev` are difficult to execute. Here is an example of a command we execute on CircleCI which just sets default mail system for Drupal 7 app:
```
ddev exec drush ev 'variable_set\(\"mail_system\",\[\"default-system\"\=\>\"DefaultMailSystem\"\]\)\;'
```

Doesn't look really nice. And it took some time to realize where is the problem and write it in a ways
 it's executable.

rfay:
You both know that drush runs on the host too right? Most commands work fine. For example:
```
rfay@rfay-mbp-2017:~/workspace/d7git$ drush ev 'variable_set("junker99","someval");'
somevalrfay@rfay-mbp-2017:~/workspace/d7git$ drush ev 'print_r(variable_get("junker99")."\n\n");'
someval
```

rthideaway:
Ah okay, it really works thanks to `ddev_drush_settings.php`, very true. One just needs to be in docroot.
[...]

PR #2318, opened 2020-06-18
Updated docs for CLI debugging with drush

chellman:
## The Problem/Issue/Bug:

I had some trouble successfully connecting drush, xdebug, and PhpStorm.  The PHP_IDE_CONFIG setting wasn't exactly what I needed.

The docs are updated to note that PHP_IDE_CONFIG should be exported as `export PHP_IDE_CONFIG=serverName=<phpstorm_project_name>`.  I added this in a couple other places where it seemed to make sense.  I also updated the drush.example command with a comment noting that you can execute with the full path to drush instead of using the launcher to potentially make this connection more reliable.

n/a

Just docs and a comment in a command, no tests needed.

## Related Issue Link(s):

None explicitly, but maybe #1648 is a little bit relevant.

## Release/Deployment notes:

n/a


rfay:
Thanks!

docs/users/extend/custom-compose-files.md
```diff
@@ -26,6 +26,7 @@ version: '3.6'
 services:
   web:
     environment:
+      - PHP_IDE_CONFIG=serverName=mysite.ddev.site
```
heddn, line 29: Instead of having to add it custom every time, couldn't we add it to the list of default env variables generated by ddev?

heddn: BTW, I already add this customization to all my ddev containers, so +1 on its addition. The only question is about implementation.

chellman: I like the idea, but I don’t think this can be automated because the value depends on what you‘ve set up for the server name in PhpStorm. It’s not the host name that ddev generates (though of course they could be the same).

heddn: I guess what I'm suggesting is why not default it to the ddev generated name? Its what I've been using. Naming things is hard so I just use the name provided so I don't have to think of a new one.

rfay: Note that PHP_IDE_CONFIG can easily be set in the bash session inside the container. That's my own preference. I just set it when I need it.  And yes $SITENAME (I think that's what's needed) would be better here than mysite.ddev.site.

rfay: maybe PHP_IDE_CONFIG should be added as an env in the code, rather than having to do any of this.

chellman: FWIW, I'm happy wherever it goes as long as it's documented.  I'd love it if this were deployed now since that's how it is right now, and then move the config to a better place, if needed, separately.

rfay:
I went ahead added PHP_IDE_CONFIG always (with the default serverName). 

I also rebased this against current, so if you have any more commits, make sure to reset against the fork before continuing.

I hope that both of you can take a look at this. If you want, you can take a look at it in upcoming alpha5, but you can also download the artifacts from this PR once the build is done. If you want to do that I'll get the link.

cmd/ddev/cmd/dotddev_assets/commands/web/drush.example
```diff
@@ -6,3 +6,8 @@
 ## Example: "ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version"
 
 drush $@
+
+# To use the site-installed drush directly (rather than running /usr/local/bin/drush),
+# you might want to execute it directly:
+#
+# /var/www/html/vendor/bin/drush $@
```
heddn, line 13: I tend to use the relative path i.e. `vendor/bin/drush`. Not sure if that is just me or if other do that too.

pkg/ddevapp/templates.go
```diff
@@ -111,6 +111,7 @@ services:
       - COLUMNS=$COLUMNS
       - LINES=$LINES
       - TZ={{ .Timezone }}
+      - PHP_IDE_CONFIG=serverName=${DDEV_SITENAME}.${DDEV_TLD}
```
heddn, line 114: :+1:

pkg/servicetest/testdata/services/docker-compose.solr.yaml
```diff
@@ -71,7 +71,7 @@ services:
     entrypoint: [ "sh", "-c", "docker-entrypoint.sh solr-precreate dev /solr-conf" ]
 
     external_links:
-      - "ddev-router:${DDEV_SITENAME}.ddev.site"
+      - "ddev-router:${DDEV_SITENAME}.${DDEV_TLD}"
```
heddn, line 74: Can I assume this is the only other place this surfaces?

cmd/ddev/cmd/dotddev_assets/commands/web/drush.example
rfay: Just bare vendor/bin/rush doesn't work normally with drupal8+ project type because the working directory is /var/www/html/web, so it needs to be ../vendor/bin/drush in that case.

pkg/servicetest/testdata/services/docker-compose.solr.yaml
rfay: I tried to look carefully. That's just an example file, but I was surprised to see the "ddev.site" wired in there. I'll bet there are several more in ddev-contrib.

cmd/ddev/cmd/dotddev_assets/commands/web/drush.example
heddn: if the site is using any fairly modern version of drush, then it should work. Its worked from above the web docroot for a couple years now. I know it didn't used to, but that hasn't been true for a long time now.

rfay: Here's the problem:
```
$ ddev exec pwd
/var/www/html/web
```
When using `ddev exec` with a Drupal project, the working directory inside the container is the docroot (explicitly to make drush work on older versions of Drupal, so you're right there, it could be changed for drupal8+, but not in this PR). But `ddev exec`  on drupal executes in the docroot, meaning that the relative path to site-local drush is not vendor/bin/drush, it would have to be ../vendor/bin/drush, and that's pretty ambiguous. 

But I do agree that we could change the working directory in drupal8+

rfay:
If you are OK with it @chellman we'll pull it into alpha5.

chellman:
@rfay Yes, I think this looks good.  I kind of want that "e.g." at the end of the line in `drush.example` so it's totally clear that the path might be different, but that's my only thought.  Otherwise, go go go!

PR #2366, opened 2020-07-08
Improve Shell Completion documentation

peterkraume:


ggothe:
nice, was able to install it with this hints :+1:

PR #2370, opened 2020-07-08
Added commands for Sequel Ace, the successor to Sequel Pro, fixes #2369

### LINKED ISSUE #2369: Support sequelace in addition to sequelpro

[sequelace](https://github.com/Sequel-Ace/Sequel-Ace) seems to be a drop-in replacement for sequelpro, which is poorly maintained these days (and the released version doesn't work with mysql 8)

- [ ]  Support SequelAce as if it were sequelpro
- [ ] Add "sequelace" as an alias for "sequelpro"
- [ ] Docs update

**Additional context**

chellman:
(Dippy Dog returns with a PR on the correct repo. 🤦‍♂️)

## The Problem/Issue/Bug:

Sequel Pro still works, but it's crashy and hasn't been updated in a very long time.  There's a successor to the app called Sequel Ace that's based on the same code, and can connect using the same plist spec.

https://github.com/Sequel-Ace/Sequel-Ace
https://sequel-ace.com/

This adds a `sequelace` command, with an accompanying sequelace_test.

I basically copy/pasted the existing commands and modified things where appropriate, so I can't say exactly how this would work, except that after the next compilation of the codebase, you'll hopefully have `sequelace` available.

sequelace_test.go is available, and hopefully does what it should.

## Release/Deployment notes:

If this is approved, I'll make sure to update any other docs and files that reference the generated .spf file.

### DIFF
--- a/cmd/ddev/cmd/sequelace.go
+++ b/cmd/ddev/cmd/sequelace.go
@@ -1,0 +1,126 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/drud/ddev/pkg/nodeps"
+
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+
+	"runtime"
+
+	"github.com/drud/ddev/pkg/ddevapp"
+	"github.com/drud/ddev/pkg/dockerutil"
+	"github.com/drud/ddev/pkg/output"
+	"github.com/drud/ddev/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// SequelAceLoc is where we expect to find the Sequel Ace.app
+// It's global so it can be mocked in testing.
+var SequelAceLoc = "/Applications/Sequel Ace.app"
+
+// DdevSequelAceCmd represents the sequelpro command
+var DdevSequelAceCmd = &cobra.Command{
+	Use:     "sequelace",
+	Short:   "Connect sequelace to a project database",
+	Long:    `A helper command for using sequelace (macOS database browser) with a running DDEV-Local project's database'.`,
+	Example: `ddev sequelace`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 0 {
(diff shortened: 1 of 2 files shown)


rfay:
I assume this means you're having a good experience with Sequel Ace?

chellman:
Yes, so far so good. It's available through the App Store, which means sandboxing, which means socket files don't work OOB, but I can close tabs without the app crashing, and connecting to IP addresses and anything I actually care about seems to work well.

I like the idea of a custom shell command.  I'm trying to find how I could connect to Sequel Pro or Ace on the cli without a plist file, and haven't found it yet.  If that exists, that would certainly be much easier and better.

NBZ4live:
Hi. I switched to Ace, too. And like it as well. But imho it's better to make the `SequelproLoc` configurable rather than copy and pasting the entire command. This will keep it as Sequel Pro command, and will allow to use Sequel Pro forks.

chellman:
I've found out that it's possible to construct a connection to Sequel Ace, or Sequel Pro I assume, from the command line, e.g.

`open "mysql://root:root@127.0.0.1:3306/db" -a "Sequel Ace"`

If someone else doesn't get to it first, I'll make a custom command for this. I expect a separate branch for it is probably the right call, but if it should go on this one, let me know.

chellman:
I'm not going to say this should definitely be closed, but fwiw, the custom command is available at #2376.

rfay:
I thoroughly appreciate this but am reluctant to go with this one and add a whole new copy/paste command, as I think you are as well. I think https://github.com/drud/ddev/pull/2376 is a better approach (and can actually replace `ddev sequelpro` perhaps). Thanks for both of them. Closing this for now, happy to continue the conversation. The biggest win would be if we could figure out a way to launch Sequel Pro with args like we can do with Sequel Ace. Unfortunately, it's not easy to get Sequel Ace.app associated with .spf files, and Sequel Pro.app doesn't have the URL launch approach.

PR #2372, opened 2020-07-09
Use MariaDB 10.3 with Drupal 8, fixes #2369

### LINKED ISSUE #2369: Support sequelace in addition to sequelpro

damienmckenna:
This is to aid with future compatibility for Drupal 9.

## The Problem/Issue/Bug:

Drupal 8 sites that upgrade to Drupal 9 will be stuck with databases that won't work anymore, because ddev uses mariadb 10.2 by default whereas D9 requires 10.3.

This changes the configuration to use mariadb 10.3 when building an instance for Drupal 8.

Build an instance for D8.

TBD.

## Related Issue Link(s):

#2232

## Release/Deployment notes:

TBD.

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -583,6 +583,13 @@ func drupal7ConfigOverrideAction(app *DdevApp) error {
 	return nil
 }
 
+// drupal8ConfigOverrideAction overrides mariadb_version for Druapl 8 for future
+// compatibility with Drupal 9, since it requires at least 10.3.
+func drupal8ConfigOverrideAction(app *DdevApp) error {
+	app.MariaDBVersion = nodeps.MariaDB103
+	return nil
+}
+
 // drupal9ConfigOverrideAction overrides mariadb_version for D9,
 // since it requires at least 10.3
 func drupal9ConfigOverrideAction(app *DdevApp) error {
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -73,7 +73,7 @@ func init() {
 			settingsCreator: createDrupal7SettingsFile, uploadDir: getDrupalUploadDir, hookDefaultComments: getDrupal7Hooks, apptypeSettingsPaths: setDrupalSiteSettingsPaths, appTypeDetect: isDrupal7App, postImportDBAction: nil, configOverrideAction: drupal7ConfigOverrideAction, postConfigAction: nil, postStartAction: drupal7PostStartAction, importFilesAction: drupalImportFilesAction, defaultWorkingDirMap: docrootWorkingDir,
 		},
 		nodeps.AppTypeDrupal8: {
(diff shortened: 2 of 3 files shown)


rfay:
Does D8 officially support MariaDB 10.3? Note that an autoupgrade from 10.2 to 10.3 probably works, but I'll have to give it a try.

rfay:
FYI, the worry about upgradeability doesn't seem to be valid. 10.2 upgrades automatically to 10.3 if changed:

With a d8 site running 10.2:
```
ddev stop
ddev config --mariadb-version=10.3
ddev start
```
You'll see if come up successfully. `ddev logs -s db`:

```
Starting with db server version=10.3 but database was created with '10.2'.
Attempting upgrade, but it may not work, you may need to export your database, 'ddev stop -RO', start, and reimport.
+ '[' 10.3 '!=' 10.2 ']'
+ echo 'Starting with db server version=10.3 but database was created with '\''10.2'\''.'
+ echo 'Attempting upgrade, but it may not work, you may need to export your database, '\''ddev stop -RO'\'', start, and reimport.'
+ pid=19
+ set +x
+ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/sbin:/usr/local/bin:/usr/local/mysql/bin
+ mysqld --skip-networking --skip-grant-tables --socket=/var/tmp/mysql.sock
+ echo 'Attempting mysql_upgrade because db server version 10.3 is not the same as database db version 10.2'
Attempting mysql_upgrade because db server version 10.3 is not the same as database db version 10.2
+ mysql_upgrade --socket=/var/tmp/mysql.sock
Phase 1/7: Checking and upgrading mysql database
```

rfay:
I see that the [Drupal database requirements](https://www.drupal.org/docs/system-requirements/database-server) say that any MariaDB 5.5+ will work, so there shouldn't be anything wrong with 10.3. However, I distrust that page because it doesn't bother to mention that 10.3 is required for Drupal 9.

rfay:
One shortsighted thing I did a couple of months ago was to hide the mariadb_version setting in config.yaml if it was the default 10.2. This makes it not obvious to people when they upgrade what to do, or what the current is. 

To solve the testing problems here I think it's probably important to remove some of the code that does that and let the 10.2 default shine through so that people can just see it and change it. I think it might be worth doing that when fixing the tests for this PR.

I took a quick attempt at fixing that late in the v1.15 release cycle but decided it was too dangerous to do late in the cycle, but can be done with this PR or another one early in the 1.16 cycle.

rfay:
Judging from the responses to https://twitter.com/randyfay/status/1285240561071349760 I think we'll be able to go ahead and do as you wanted to do earlier, and just move the default for everybody to 10.3

The one problem that will have to be addressed is as above, https://github.com/drud/ddev/pull/2372#issuecomment-656453481

pkg/ddevapp/drupal.go
```diff
@@ -626,6 +626,13 @@ func drupal7ConfigOverrideAction(app *DdevApp) error {
 	return nil
 }
 
+// drupal8ConfigOverrideAction overrides mariadb_version for Druapl 8 for future
```
nlisgo, line 629: typo: `Drupal 8`

rfay:
Rebased this to current. I don't see why we shouldn't go forward with it, pending manual testing.

However, for the longer-term objective of making 10.3 the default for everything, https://github.com/drud/ddev/issues/2555 is a prerequisite.

PR #2376, opened 2020-07-10
Add example custom command for Sequel Ace or Sequel Pro (macOS)

chellman:
## The Problem/Issue/Bug:

We'd like to support either Sequel Ace or Sequel Pro for Mac users to connect directly to the db container visually.  This example custom command does that, assuming the apps are installed in the default location (`/Applications/`)

This might be a better solution than #2370.

Adds a custom command example file `sequel.example` to `.ddev/commands/host`

- Rename `sequel.example` to `sequel` and run `ddev sequel`.

If Sequel Ace is installed, it is preferred.  If neither is installed, the command should just print an error message.

n/a

## Related Issue Link(s):

#2369 

## Release/Deployment notes:

Might possibly want to update the docs to note this.

### DIFF
--- a/cmd/ddev/cmd/dotddev_assets/commands/host/mysqlworkbench.example
+++ b/cmd/ddev/cmd/dotddev_assets/commands/host/mysqlworkbench.example
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 ## #ddev-generated
-## Description: Run MySQLWorkbench against current db
+## Description: Run MySQL Workbench against current db
 ## Usage: mysqlworkbench
 ## Example: "ddev mysqlworkbench"
 
-# Note that this examle uses $DDEV_HOST_DB_PORT to get the port for the connection
-# Mysql Workbench can be obtained from https://dev.mysql.com/downloads/workbench/
+# Note that this example uses $DDEV_HOST_DB_PORT to get the port for the connection
+# MySQL Workbench can be obtained from https://dev.mysql.com/downloads/workbench/
 
 query="root:root@127.0.0.1:${DDEV_HOST_DB_PORT}"
 
--- a/cmd/ddev/cmd/dotddev_assets/commands/host/sequel.example
+++ b/cmd/ddev/cmd/dotddev_assets/commands/host/sequel.example
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+## #ddev-generated
+## Description: Run Sequel Ace or SequelPro against current db
+## Usage: sequel
+## Example: "ddev sequel"
+
+# This command uses $DDEV_HOST_DB_PORT to get the port for the connection
+#
+# Both apps are free, but available only for macOS-only 
+# Find Sequel Ace in the Mac App Store or at https://sequel-ace.com/
+# Find Sequel Pro at https://www.sequelpro.com/
+
+query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/db"
+
+app=""
+if [ -f "/Applications/Sequel Ace.app/Contents/MacOS/Sequel Ace" ]; then
+    app="Sequel Ace"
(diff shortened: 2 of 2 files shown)


cmd/ddev/cmd/dotddev_assets/commands/host/mysqlworkbench.example
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 ## #ddev-generated
-## Description: Run MySQLWorkbench against current db
+## Description: Run MySQL Workbench against current db
```
rfay, line 4: I imagine your editor did this to "help:" you, but in general, try to keep PRs confined to the scope you want them confined to. If there are two unrelated issues, do two PRs.

cmd/ddev/cmd/dotddev_assets/commands/host/sequel.example
```diff
@@ -0,0 +1,28 @@
...
+## Usage: sequel
+## Example: "ddev sequel"
+
+# This command uses $DDEV_HOST_DB_PORT to get the port for the connection
+#
+# Both apps are free, but available only for macOS-only 
+# Find Sequel Ace in the Mac App Store or at https://sequel-ace.com/
+# Find Sequel Pro at https://www.sequelpro.com/
+
+query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/db"
```
rfay, line 14: This query strategy unfortunately doesn't work with Sequel Pro, only with Sequel Ace. 
Sequel_Pro

cmd/ddev/cmd/dotddev_assets/commands/host/mysqlworkbench.example
chellman: Roger. I wasn't sure if separate PRs for minor typos was a good thing or not, but I can certainly see the logic.

cmd/ddev/cmd/dotddev_assets/commands/host/sequel.example
chellman: Not sure if you've seen this, but if Sequel Pro ever gets up off the mat (and [maybe it will](https://github.com/sequelpro/sequelpro/issues/3705#issuecomment-653959299)), it will support these schemes. The [nightly builds](https://sequelpro.com/test-builds) already do (of course the most recent of these is over a year old, whatever that says).

That doesn't solve anything for people using the current release on the sequelpro website, of course.

If I ruled the world, I might be inclined to deprecate support for Sequel Pro with the hope that its actual released version becomes less crashy.

I could also see if it's possible to create that spf file without writing it to disk and pipe it into an `open` command.  Seems like a bit of a hassle, but ddev has support sequelpro for long enough (I'm guessing) that removing it would be uncharitable.

cmd/ddev/cmd/dotddev_assets/commands/host/mysqlworkbench.example
rfay: It's no biggie. Just saying. And it's OK here.

cmd/ddev/cmd/dotddev_assets/commands/host/sequel.example
rfay: I do hate that these similar tools have completely different launch options. 

I think maybe we should add sequel-ace as an official shell command (like launch), based on this one, and also provide an *example* for sequel-pro.  And remove the existing golang one. 

Then our problem is how to hide it if Sequel-Ace isn't installed or if we're not on macOS. That requires a bit of extra effort. I guess I'll work on https://github.com/drud/ddev/issues/2209 and try to make it so individual commands can determine whether they show up?

rfay:
Closing this in favor of #2452, which brings this in, and only shows the commands where the app exists. It supports both Sequel Ace and Sequel Pro, depending on which may be installed. Thanks for all the good work here.

PR #2393, opened 2020-07-18
Allow wildcard additional_hostnames, fixes #2391

### LINKED ISSUE #2391: Wildcard subdomains / hostnames
I have recently found myself working on a project that does sub-domain routing based on generated data
(e.g. `{UUID}.demo.myapp.com`).

While developing I found it a bit cumbersome to add new hostnames to the configuration file and restart ddev
everytime I added a new demo page. That in turn made me wish for wildcard subdomains / hostnames

Ideally, it would be possible to use wildcards in `additional_hostnames` like this:
```yaml
additional_hostnames:
    - '*.demo.myapp'
```

**Additional context**
While trying to find out whether or not this already existed I stumbled over the implementation detail, that
the `ddev.site` domain is actually registered and just points to `127.0.0.1` to make setup easier. If a user is
offline however, entries in `/etc/hosts` are being generated.

With the online DNS setup wildcards should be fairly easy to support since you would just need to tell
ddev-router (which runs haproxy I think?) about the wildcard.
I suspect that the offline `/etc/hosts` setup is why wildcard hostnames could be an issue since there
are no wildcards in `/etc/hosts`.

rfay:
## The Problem/Issue/Bug:

#2391 requests wildcards for additional_hostnames. I experimented with it and it was pretty easy. Here's a PR. It will need to be looked at much more closely later in the release cycle.

* Add finesse to validation to remove the "*." on front  of additional_hostnames.

- [x] Consider whether this is "OK" to go in, since it doesn't work  unless using DNS.
- [x] Needs test for wildcard.
- [x] Needs docs
- [x] Prevent simple "*", as that should be invalid. Validators should make sure empty string is not accepted.

Allow *. in validation - everything else seems to work at first blush.

- [x] Needs test for wildcard


j6s:
I checked out and built your branch locally to test it.

1. Create an empty project (in my case I had an `var_dump($_SERVER);` in there to check the vars passed to PHP)
2. Add a wildcard hostname (in my case `*.ddev-test`)
3. Open in browser using different hostnames

Wildcard hostnames now seem to work when online and (as expected) not work when offline. I cannot determine what effects this change has on the rest of ddev but the wildcard part seems to work.

j6s:
Just out of curiosity I checked the following 2 things as well:

* Wildcard in the middle of a hostname (e.g. `foo.*.ddev-test`)
    * Does not work, throws up the `invalid hostname` message. I wouldn't know what usecase would need this over a leading wildcard however
* Only wildcard as `additional_hostnames` entry
    * This, surprisingly worked and now routes all ddev domains to my project. It may be a good idea to not allow just `*` as a hostname to prevent accidentally shadowing other projects to easily

Example wildcard only config:

```yaml
additional_hostnames:
  - '*'
```

PR #2394, opened 2020-07-18
New warning about Catalina; related cleanup

damienmckenna:
## The Problem/Issue/Bug:

The notes about the new /sbin/nsfd requirement for Catalina was not included in the Catalina section of the document.

Some rewording of the performance file to draw attention to the Catalina requirements.


rfay:
Thanks so much. I added one more [commit](https://github.com/drud/ddev/pull/2394/commits/f27064f01f125561a3e63f234a72e4b8675e9622) that adds screenshots to deal with what I perceived to be the issues we ran into.

Please take a look at https://github.com/drud/ddev/blob/f27064f01f125561a3e63f234a72e4b8675e9622/docs/users/performance.md and see if you like it.

I also removed the apparently-unnecessary /System/Volumes/Data from the setup script. 

Would appreciate your review @damienmckenna

damienmckenna:
Thank looks good, thank you.

PR #2395, opened 2020-07-18
Don't use pipefail with non-bash exec/custom commands, fixes drud/ddev-contrib#80

rfay:
##  The Problem/Issue/Bug:

drud/ddev-contrib#80 points out that custom commands (and all execs) on non-ddev-provided containers will fail because the "sh" shell is used by default, and it doesn't provide the "pipefail" error check.

Only use -o pipefail if bash is the shell.

* Consider dropping pipefail out completely to simplify the code.

* Test drud/ddev-contrib#80
* Test other ordinary `ddev exec` with ddev and non-ddev containers.

- [ ] Needs a test with 3rd party service to make sure this is maintained.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1066,9 +1066,6 @@ func (app *DdevApp) Exec(opts *ExecOpts) (string, string, error) {
 		shell = "sh"
 	}
 	errcheck := "set -eu"
-	if shell == "bash" {
-		errcheck = errcheck + " -o pipefail "
-	}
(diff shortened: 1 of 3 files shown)


icanhazstring:
Is it possible to run this "from source" to check if the fix works?

rfay:
The easiest way is to download the artifacts from the CircleCI tests, https://app.circleci.com/pipelines/github/drud/ddev/2437/workflows/d8ad956f-dca0-4639-9b62-2a533416eb5c/jobs/26223/artifacts

There are also some basic docs on how to build at https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md

PR #2402, opened 2020-07-20
Reset global mkcert var if  has changed, fixes #2373

### LINKED ISSUE #2373: global_config.yaml mkcert_caroot never gets updated
**Describe the bug**

The value global_config.mkcert_caroot never gets updated once it is set.

This means that if somebody (especially installing WSL2) installs with $CAROOT set wrong, it can only be fixed by removing that line. I imagine it should be updated on every `ddev start` ?

**To Reproduce**

In WSL2, do a `mkcert -install` and then `ddev start`  a project (with $CAROOT unset). This will set mkcert_caroot in ~/.ddev/global_config.yaml

Now actually use setx on the Windows site to set $CAROOT correctly. `ddev poweroff && ddev start`. The value in global_config will not be updated.

**Expected behavior**

**Screenshots**

 - Host computer OS and Version: [e.g. Windows 10, macOS High Sierra]
 - Docker version information (use `docker version`) and insert it here.

**Additional context**

rfay:
## The Problem/Issue/Bug:

#2373 points out that `DdevGlobalConfig.MkcertCARoot` never gets updated of $CAROOT changes, which means that especially in WSL2, if any mistake is made in $CAROOT, it never gets corrected.

If $CAROOT isn't the same as `DdevGlobalConfig.MkcertCARoot` then re-read the correct value.

Test in WSL2. 

* Mess up the setx command so that WSLENV contains a semicolon instead of colon.
* You should now have a broken CAROOT on WSL side
* `ddev start` and see that the bad CAROOT gets into ~/.ddev/global_config.yaml
* Now fix CAROOT via a proper setx. 
* `ddev start` and see that it get fixed in global_config.
* Verify that chrome trusts a site.

### DIFF
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -100,7 +100,9 @@ func ReadGlobalConfig() error {
 	if DdevGlobalConfig.ProjectList == nil {
 		DdevGlobalConfig.ProjectList = map[string]*ProjectInfo{}
 	}
-	if DdevGlobalConfig.MkcertCARoot == "" {
+	// Set/read the CAROOT if it's unset or different from $CAROOT (perhaps $CAROOT changed)
+	caRootEnv := os.Getenv("CAROOT")
+	if DdevGlobalConfig.MkcertCARoot == "" || (caRootEnv != "" && caRootEnv != DdevGlobalConfig.MkcertCARoot) {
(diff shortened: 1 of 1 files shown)


rfay:
Tested this on WSL2 as suggested, it solves the problem. Discovered yet another minor problem with the doc.

PR #2408, opened 2020-07-21
Docs: More failsafe on WSLENV having colon [skip ci][ci skip]

rfay:
## The Problem/Issue/Bug:

People have had trouble with ";" in the WSLENV environment variable. The result is not being able to get a trusted connection in chrome.

Make sure that the WSLENV variable gets set to a usable value, even if it was previously set wrong.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -63,7 +63,7 @@ We'll walk through these in more detail. You may prefer other techniques of inst
 2. **Chocolatey:** We recommend using Chocolatey for [Chocolatey](https://chocolatey.org/install) installing required Windows apps like mkcert. In an administrative PowerShell, `Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))`
 3. In an administrative PowerShell: `choco install -y mkcert`
 4. In administrative PowerShell, run `mkcert -install` and answer the prompt allowing the installation of the Certificate Authority.
-5. In administrative PowerShell, run the command `setx CAROOT "$(mkcert -CAROOT)"; If ($Env:WSLENV -notlike "*CAROOT*") { setx WSLENV "CAROOT/up:$Env:WSLENV" }`. This will set WSL2 to use the Certificate Authority installed on the Windows side.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -63,7 +63,7 @@ We'll walk through these in more detail. You may prefer other techniques of inst
 2. **Chocolatey:** We recommend using Chocolatey for [Chocolatey](https://chocolatey.org/install) installing required Windows apps like mkcert. In an administrative PowerShell, `Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))`
 3. In an administrative PowerShell: `choco install -y mkcert`
 4. In administrative PowerShell, run `mkcert -install` and answer the prompt allowing the installation of the Certificate Authority.
-5. In administrative PowerShell, run the command `setx CAROOT "$(mkcert -CAROOT)"; If ($Env:WSLENV -notlike "*CAROOT*") { setx WSLENV "CAROOT/up:$Env:WSLENV" }`. This will set WSL2 to use the Certificate Authority installed on the Windows side.
+5. In administrative PowerShell, run the command `setx CAROOT "$(mkcert -CAROOT)"; If ($Env:WSLENV -notlike "*CAROOT/up:*") { setx WSLENV "CAROOT/up:$Env:WSLENV" }`. This will set WSL2 to use the Certificate Authority installed on the Windows side.
```
gilbertsoft, line 66:
```suggestion
5. In administrative PowerShell, run the command `setx CAROOT "$(mkcert -CAROOT)"; If ($Env:WSLENV -notlike "*CAROOT/up:*") { setx WSLENV ("CAROOT/up:" + ($Env:WSLENV -replace '(.*)CAROOT[^:;]*[:;](.*)', '$1$2').Trim(':')) }`. This will set WSL2 to use the Certificate Authority installed on the Windows side.
```

gilbertsoft: Here we go, this new PS script should now also fix the errors from the first version with using `;` instead of `:` as separator. It also fixes wrong options if a user has set it before.

rfay: I don't object to the trim, but WSENV works fine without it.  Also doubled `CAROOT/up:CAROOT/up:` works fine. 

I really appreciate you adding this, but I think it's simpler as is. I think well-tested version of this will be OK in your script, but I'm going to pass on it here.

gilbertsoft: The trim is to ensure we don't have a separator at the begin and end which could lead then to two separators, it's not really important but produces a cleaner variable value.

Issue #2410, opened 2020-07-22
Unable to use gulp-livereload on macOS


graham73may:
**Describe the bug**
I've just moved from Windows to macOS, the only thing I'm having trouble with is getting [gulp-livereload](https://www.npmjs.com/package/gulp-livereload) to apply the updated styles. 

Does anyone have any experience setting up a Gulp watch task in a ddev site on macOS?
(For reference, it worked fine on Windows, usually these problems are the opposite way round :joy: )

**To Reproduce**
Start a WordPress project with the default ddev configs for a WP site. 

I've taken this from a WordPress project, structure is:

```
/wp-content/themes/{THEME_NAME}/assets/src/package.json
/wp-content/themes/{THEME_NAME}/assets/src/gulpfile.js
```
[gulp.zip](https://github.com/drud/ddev/files/4961057/gulp.zip)

From the host Run `npm install` inside the `src` folder to install all of our node_modules...

Once that's ready, run the gulp watch task on the host, when finished livereload will be ready to use!
Personally I'm running this all through PhpStorm -> Right click gulpfile.js -> Show Tasks -> Click the Watch task. 

This is the recommended Chrome extension to go along with livereload: 
https://chrome.google.com/webstore/detail/livereload/jnihajbhpnppcggbcgedagnkighmdlei

Watched files:
```
/wp-content/themes/{THEME_NAME}/assets/scss/style.scss
/wp-content/themes/{THEME_NAME}/assets/scss/{loads of other .scss files that are included in to style.scss}
```

When a change is detected are compiling to:
```
/wp-content/themes/{THEME_NAME}/assets/css/style.css
/wp-content/themes/{THEME_NAME}/assets/css/style.min.css
```

**Expected behavior**

- [x] You make a change in an .scss file
- [x] This triggers the watch task, triggering the relevant gulp task in this case `build-css`
- [x] .css and .min.css are generated on the host successfully
- [ ] The browser automatically applies the new styles without requiring a full refresh

**Version and configuration information (please complete the following information):**

- Host computer OS and Version:  macOS Catalina
- Docker Desktop version `2.3.0.3`
- ddev version `v1.15.0`
- config.yaml contents for the misbehaving project `defaults for a WP project`
- Do you have any custom configuration (nginx, php, mysql) in the .ddev folder? `none`

**config.yaml**

```yaml
APIVersion: v1.12.0
name: site.co.uk
type: wordpress
docroot: ""
php_version: "7.4"
webserver_type: apache-fpm
[...]

PR #2413, opened 2020-07-22
nginx shouldn't have long expires on css/js, fixes #2404

### LINKED ISSUE #2404: Expires header "max" on JS and CSS files in Nginx configuration; resulting in unhappy front-end developers
We recently converted all our projects from our old VM-based development solution to ddev and we are very happy with the improved developer experience. However, our front-end developers do have an issue with the JS and CSS files getting an expires header that results in the browsers caching JS and CSS files and not seeing changes when JS or CSS is changed.

The expires header is set in the Nginx configuration:

    # Media: images, icons, video, audio, HTC
    location ~* \.(js|css|png|jpg|jpeg|gif|ico|svg)$ {
        try_files $uri @rewrite;
        expires max;
        log_not_found off;
    }

Is there a reason to set this header to max for JS and CSS files? I cannot believe other developers don't have this issue so maybe we are overlooking something obvious...

I would like to have a separate location directive for just JS and CSS files that sets expires -1;

    # JS and CSS
    location ~* \.(js|css)$ {
        try_files $uri @rewrite;
        expires -1;
        log_not_found off;
    }

rfay:
## The Problem/Issue/Bug:

Front-end developers really want changes to css and js to come through right away, and ddev is a developers' tool. 

#2404 explains how the Expires setting in nginx makes it so devs can't see their CSS changes.

I ran into this exact problem exploring #2410 , so figure it will probably affect others. 

Turn off the expires header for css and js.

Must test for every changed CMS project type.
* Start the project
* Visit pages
* Use Chrome debugger or whatever to see the expires situation for css

### DIFF
--- a/pkg/ddevapp/webserver_config_packr_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_packr_assets/nginx-site-backdrop.conf
@@ -92,9 +92,16 @@ server {
     }
 
     # Media: images, icons, video, audio, HTC
-    location ~* \.(js|css|png|jpg|jpeg|gif|ico|svg)$ {
+    location ~* \.(png|jpg|jpeg|gif|ico|svg)$ {
         try_files $uri @rewrite;
         expires max;
+        log_not_found off;
+    }
+
+    # js and css always loaded
+    location ~* \.(js|css)$ {
+        try_files $uri @rewrite;
+        expires -1;
         log_not_found off;
     }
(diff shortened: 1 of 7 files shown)


shaal:
Sounds like a good plan!
Will it be a configuration that can be set to on/off ?

rfay:
@shaal all nginx config can be changed, just a quick edit in .ddev/nginx_full/nginx-site.conf, and remove the #ddev-generated there.

PR #2423, opened 2020-07-27
[docs] Improve wording of "In-container home directory configuration"

ahillio:
In response to https://stackoverflow.com/questions/63084369/in-ddev-how-to-automatically-use-custom-shell-config-inside-docker-container I tried to make the shell configuration more explicit... particularly including the exact words "shell config" and "dotfiles" so the readthedocs search can find them.

In regards to "If you want to use a shell other than bash, add it as a package to [customize the Docker image](https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/).".... I presume that's correct for adding a shell like zsh but I'm not sure..?

### DIFF
--- a/docs/users/extend/in-container-configuration.md
+++ b/docs/users/extend/in-container-configuration.md
@@ -1,10 +1,10 @@
 ## In-Container home directory configuration
 
-Custom shell config (bash or your preferred shell), your usual git configuration, a composer auth.json and more can be achieved within your containers.
+Custom shell config (bash or your preferred shell), your usual git configuration, a composer auth.json and more can be achieved within your containers.  Place all your dotfiles in your `~/.ddev/homeadditions` directory and DDev will use these in all your projects' Docker containers.
 
 On `ddev start`, ddev attempts to create a user inside the web and db containers with the same name and use id as the one you have on the host.
 
-If you have `homeadditions` directory either in `~/.ddev/homeadditions` (the global .ddev directory) or the `.ddev/homeadditions` directory of a particular project, their contents will be copied recursively into the in-container home directory during `ddev start`. (Note that project homeadditions contents override the global homeadditions.)
+DDev looks for `homeadditions` directory either in `~/.ddev/homeadditions` (the global .ddev directory) or the `.ddev/homeadditions` directory of a particular project, and will copy their contents recursively into the in-container home directory during `ddev start`. (Note that project homeadditions contents override the global homeadditions.)
 
 If you want to use a shell other than bash, add it as a package to [customize the Docker image](https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/).
 
--- a/docs/users/extend/in-container-configuration.md
+++ b/docs/users/extend/in-container-configuration.md
@@ -1,16 +1,14 @@
-## In-Container home directory configuration
+## In-Container home directory and shell configuration
 
-Custom shell config (bash or your preferred shell), your usual git configuration, a composer auth.json and more can be achieved within your containers.  Place all your dotfiles in your `~/.ddev/homeadditions` directory and DDev will use these in all your projects' Docker containers.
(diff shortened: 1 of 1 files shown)


docs/users/extend/in-container-configuration.md
```diff
@@ -1,11 +1,12 @@
 ## In-Container home directory configuration
 
-There are a number of reasons that your web container home directory may need custom configuration. You may want your normal git configuration in there, or a composer auth.json, etc.
+Custom shell config (bash or your preferred shell), your usual git configuration, a composer auth.json and more can be achieved within your containers.  Place all your dotfiles in your `~/.ddev/homeadditions` directory and DDev will use these in all your projects' Docker containers.
```
rfay, line 3: The only thing that's wrong with this is that you're just thinking about shell configuration (.profile, .zshrc, stuff like that). But there are many other things that go into a home directory. Of course, an auth.json has nothing to do with shell configuration, so that's a great example.

docs/users/extend/in-container-configuration.md
```diff
@@ -1,11 +1,12 @@
...
 
 On `ddev start`, ddev attempts to create a user inside the web and db containers with the same name and use id as the one you have on the host.
 
-If you have a ~/.ddev/homeadditions
-If you have a ~/.ddev/homeadditions directory (in the global .ddev directory) or your project has a .ddev/homeadditions directory, their contents will be copied recursively into the in-container home directory during `ddev start`. (Note that project homeadditions contents override the global homeadditions.)
+DDev looks for `homeadditions` directory either in `~/.ddev/homeadditions` (the global .ddev directory) or the `.ddev/homeadditions` directory of a particular project, and will copy their contents recursively into the in-container home directory during `ddev start`. (Note that project homeadditions contents override the global homeadditions.)
+
+If you want to use a shell other than bash, add it as a package to [customize the Docker image](https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/).
```
rfay, line 9: I don't see a reason for this line to be here, as it's on a completely different topic. But `webimage_extra_packages` would be the way to do it, that's covered over there.

docs/users/extend/in-container-configuration.md
```diff
@@ -1,11 +1,12 @@
 ## In-Container home directory configuration
```
rfay, line 1: Aren't you going to want to improve this title?

docs/users/extend/in-container-configuration.md
ahillio: Would it make sense to break out the "In-container Configuration" into two/multiple pages then?  1) Shell configuration, and 2) Other stuff like auth.json.

Or perhaps to just expand more on what that other stuff is in that section?  I'm not at all aware of what else could be done.

docs/users/extend/in-container-configuration.md
ahillio: The benefit I see in that line is if someone's thinking "I want to customize my shell" and they go into that page but that page would then be insufficient if their desired shell customizations require a different shell, so to save them from having to then read through the menu/index and try to find the right page (which might be easy enough for some people) linking to that other page directly could make it easier for others.

docs/users/extend/in-container-configuration.md
ahillio: yeah I considered that but wasn't sure.   I guess it'd depend on whether you think it'd be good to break it out into smaller more specific pages like I mentioned in the other comment.  What do you think?

docs/users/extend/in-container-configuration.md
rfay: It's always a tradeoff - do you try to teach people how to use the shell and every single way you can use it? It's a hopeless task I think.

This feature is actually for advanced users, so I'm not very enthusiastic about telling people every option they could ever have.

PR #2424, opened 2020-07-27
PHP_IDE_CONFIG must not be set for php-fpm

rfay:
## The Problem/Issue/Bug:

This fixes a regression in v1.15.0 - 

DDEV-Local v1.15 adds automatic settings of PHP_IDE_CONFIG to help people who are doing command-line debugging and want to use xdebug. Otherwise they'd have to set it themselves. 

However, setting this throughout the web container (in php-fpm's context) results in PHPStorm not popping up the normal server/mapping creation dialog box. It's possible to work around it, but we've always been proud of how ddev makes PHPStorm so easy... and suddenly it's not.

Only set PHP_IDE_CONFIG in the `ddev ssh` or `ddev exec` context

Set up PHPStorm
Remove all servers
`ddev xdebug on`
Set a breakpoint in index.php
Visit the site

It should pop up a mapping dialog.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -86,8 +86,8 @@ ADD ddev-webserver-base-files /
 RUN phpdismod xdebug
 RUN curl -sSL "https://github.com/mailhog/MailHog/releases/download/v${MAILHOG_VERSION}/MailHog_linux_amd64" -o /usr/local/bin/mailhog
 
-RUN curl -ssL -O https://raw.githubusercontent.com/pantheon-systems/terminus-installer/master/builds/installer.phar && php installer.phar install
-RUN curl -ssL -O $DDEV_LIVE_DOWNLOAD_URL && unzip ddev-live.zip && mv ddev-live /usr/local/bin && chmod +x /usr/local/bin/ddev-live && rm ddev-live.zip
+RUN curl -sSL -O https://raw.githubusercontent.com/pantheon-systems/terminus-installer/master/builds/installer.phar && php installer.phar install
+RUN curl -sSL -O $DDEV_LIVE_DOWNLOAD_URL && unzip ddev-live.zip && mv ddev-live /usr/local/bin && chmod +x /usr/local/bin/ddev-live && rm ddev-live.zip
 
 # magerun and magerun2 for magento
 RUN curl -sSL https://files.magerun.net/n98-magerun-latest.phar -o /usr/local/bin/magerun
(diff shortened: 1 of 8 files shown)


brace110:
`To fix it set server name by environment variable PHP_IDE_CONFIG and restart debug session.`

Updating to 1.15.2 broke my command line debugging, how can I get this back?

rfay:
Wow, thanks for the quick report. I didn't expect anybody to have updated yet. 

Please help me recreate your error. PHP_IDE_CONFIG is set in the web container, right?

Do you have some homeadditions added, overwriting the .profile or something?

Please `ddev ssh` and show me the output of `echo $PHP_IDE_CONFIG`
and `ddev exec 'echo $PHP_IDE_CONFIG'`

rfay:
I was able to recreate this problem and will have a new release shortly. 

Simple answer for the short term, 
```
export PHP_IDE_CONFIG=serverName=${DDEV_SITENAME}.${DDEV_TLD}
```

The easiest thing to do would be to (in the web container) `sudo vi /etc/bashrc/php_ide.bashrc` and add the "export" there.

Ironically, the behavior you got on the command line (not finding the server/mapping) is exactly the behavior that everybody else was having with v1.15.1 when using non-command-line :)

rfay:
And unfortunately, the fix for the command-line debugging breaks web debugging, this may take me a bit more time. I thought this was a clean workaround to prevent that.

If PHP_IDE_CONFIG is *set* with web debugging then the Server auto-config doesn't work.
If PHP_IDE_CONFIG is *unset* with command-line debugging, then the server auto-config doesn't work, but no config works.

Your best immediate solution, should work everywhere I think, is to use the v1.14.2 solution as described here:

https://ddev.readthedocs.io/en/v1.14.2/users/step-debugging/#phpstorm-and-command-line-debugging

rfay:
Except that those docs were always wrong! Should be `export PHP_IDE_CONFIG=serverName=${DDEV_SITENAME}.${DDEV_TLD}` of course.

rfay:
https://github.com/drud/ddev/pull/2427 should fix it, after tests succeed with that (and another manual test run) it will become v1.15.3

rfay:
@brace110 I'd love it if you could confirm whether v1.15.3 solves your problem, https://github.com/drud/ddev/releases/tag/v1.15.3

brace110:
@rfay Yes, confirmed, 1.15.3 solved my issue. I had a temporary fix where I included a 'docker-compose.env.yaml' with this content:

```
version: '3.6'

services:
  web:
    environment:
      - DRUSH_ALLOW_XDEBUG=1
      - PHP_IDE_CONFIG=serverName=${DDEV_SITENAME}.${DDEV_TLD}
```

But this file is no longer necessary, thanks for the quick response.

PR #2427, opened 2020-07-29
Make PHP_IDE_CONFIG work for both command-line and web debugging

rfay:
## The Problem/Issue/Bug:

Unfortunately version v1.15.2 made Xdebug with PHPStorm work fine for people using it for web debugging, but broke it for command-line debugging.  See https://github.com/drud/ddev/pull/2424

This revision works with both command-line and web debugging. 

Make sure that PHP_IDE_CONFIG does *not* get loaded by php-fpm (which breaks web debugging) but *does* get loaded by `ddev ssh` (using /etc/bash.bashrc)

Try debugging with PHPStorm both with a web setup and a command-line setup.


brace110:
Thank you for fixing this promptly.

rfay:
Thanks! Note that you should use `ddev ssh` for command-line debugging, not `ddev exec`

PR #2431, opened 2020-07-30
Allow processing of PDF files with ImageMagick

gilbertsoft:
## The Problem/Issue/Bug:
TYPO3 is able to convert PDF files to images for a preview with the help of ImageMagick / GraphicsMagick but ImageMagick blocks PDF files by default on Debian.

The policy.xml file distributed by the APT package on Debian is added to the web server container image and the related line is commented out to have PDF conversion enabled by default in DDEV.

Go to the TYPO3 Backend and under environment in the admin / install tool run the image test. There should not be an error and the PDF should be shown as image in the related test.

grafik

**The important test:**

grafik

**In case something is broken it looks like this:**

grafik

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/ImageMagick-6/policy.xml
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/ImageMagick-6/policy.xml
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policymap [
+  <!ELEMENT policymap (policy)+>
+  <!ATTLIST policymap xmlns CDATA #FIXED ''>
+  <!ELEMENT policy EMPTY>
+  <!ATTLIST policy xmlns CDATA #FIXED '' domain NMTOKEN #REQUIRED
+    name NMTOKEN #IMPLIED pattern CDATA #IMPLIED rights NMTOKEN #IMPLIED
+    stealth NMTOKEN #IMPLIED value CDATA #IMPLIED>
+]>
+<!--
+  Configure ImageMagick policies.
+
+  Domains include system, delegate, coder, filter, path, or resource.
+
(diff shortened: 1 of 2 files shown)


gilbertsoft:
As you can see from the Slack thread the issue is solved by this change.

rfay:
I rebased this, pushed an image for it, gave it a specific image.

containers/ddev-webserver/ddev-webserver-base-files/etc/ImageMagick-6/policy.xml
```diff
@@ -0,0 +1,96 @@
...
+  <policy domain="delegate" rights="none" pattern="HTTPS" />
+  <policy domain="delegate" rights="none" pattern="HTTP" />
+  <!-- in order to avoid to get image with password text -->
+  <policy domain="path" rights="none" pattern="@*"/>
+  <!-- disable ghostscript format types -->
+  <policy domain="coder" rights="none" pattern="PS" />
+  <policy domain="coder" rights="none" pattern="PS2" />
+  <policy domain="coder" rights="none" pattern="PS3" />
+  <policy domain="coder" rights="none" pattern="EPS" />
+  <!-- <policy domain="coder" rights="none" pattern="PDF" /> -->
```
gilbertsoft, line 94: @rfay this is the only difference, I commented out that line

rfay:
```suggestion
 <policy domain="coder" rights="none" pattern="PDF" />
```

rfay: Is that what you're saying?

gilbertsoft: Original:  <policy domain="coder" rights="none" pattern="PDF" />
My change:

Issue #1881, opened 2019-10-10
macOS Catalina: ddev command execution seems sluggish


rfay:
**Describe the bug**

It seems to me that all ddev commands on Catalina are really slow to execute.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: [e.g. Windows 10, macOS High Sierra]
 - Docker version information (use `docker version`) and insert it here.
 - ddev version information (use `ddev version`)
 - config.yaml contents for the misbehaving project
 - Do you have any custom configuration (nginx, php, mysql) in the .ddev folder? If so, have you tried without them?

**Additional context**
Add any other context about the problem here. Thanks!

rfay:
A quick check says this is instrumentation (segment or sentry?) causing the problem. I set `instrumentation_opt_in: false` and it didn't seem to be so slow.

rfay:
I also see that `docker-compose --version` is quite slow on at least Catalina. We shouldn't really have to do that every single time.

rfay:
This is mostly due to docker-compose problems as described in docker/compose#6956 (comment) and potentially fixed by docker/compose#7010

#1905 (fixed by #1916) was also a problem here, not just for Catalina, but for everybody that had a Pantheon project active.

rfay:
I think this is mostly solved by https://github.com/docker/compose/pull/7010, which made it into docker-compose 1.25.1-rc1 which made it into Docker Desktop 2.1.7.0. `docker-compose --version` now takes .5s, much better. The problem is loads less annoying. I still thing ddev does too much rooting around and doing things over again, but the Catalina problem is solved with the latest docker desktop edge, which is also the release candidate for stable.

rfay:
I think it's worth opening this up again. There's a lot of profiling that can be done and significant speedup in ddev command execution. #1962 made some good progress, but I'd like to see it better. Every time I use Linux or WSL2 I see ddev behaving much faster.

PR #2441, opened 2020-08-05
Speed up ddev list for #1881

rfay:
## The Problem/Issue/Bug:

#1881 : ddev commands like `list` are quite slow, and slower the more projects you have.


rfay:
#2445 is better.

PR #2454, opened 2020-08-15
DDEV ARM64 Docker images

dennisameling:
Related to https://github.com/drud/ddev/issues/2338 and https://github.com/drud/ddev/issues/2322

This is a draft PR to track work on DDEV ARM64 Docker images and to share experiences.

- [x] Build `ddev-dbserver` for ARM64
- [x] Build `ddev-router` for ARM64 https://hub.docker.com/repository/docker/dennisameling/ddev-router
- [x] Build `ddev-ssh-agent` for ARM64 https://hub.docker.com/repository/docker/dennisameling/ddev-ssh-agent
- [x] Build `ddev-webserver` for ARM64 https://hub.docker.com/repository/docker/dennisameling/ddev-webserver
  - [x] Needs `drud/ddev-php-base` for ARM64 (https://github.com/drud/ddev-images/pull/16)
- [x] ~Build `test-ssh-server` for ARM64~

**ddev-dbserver**
- While MariaDB has support for ARM64 in their Docker images [since 10.0, based on Ubuntu Xenial](https://hub.docker.com/_/mariadb), percona-xtrabackup-24 for ARM64 only works on Ubuntu Bionic (18.04) and higher: https://launchpad.net/ubuntu/bionic/arm64/percona-xtrabackup. Therefore MariaDB 10.1 is the minimum version I could build with in the current state
- MySQL only has official ARM64 support since version 8.0: https://www.mysql.com/support/supportedplatforms/database.html - the official Docker MySQL image still doesn't have ARM64 support (work has been happening recently though): https://github.com/docker-library/mysql/issues/318

**ddev-webserver**
- I can't test this properly yet since `drud/ddev-php-base` [is only available for `amd64`](https://hub.docker.com/r/drud/ddev-php-base/tags). @rfay I can't find the source code for these images, is it closed source?

**How to build**
- You'll need `docker buildx` [as described here](https://www.docker.com/blog/multi-arch-images/). Installation instructions can be found here: https://github.com/docker/buildx/#building. DDEV's CI probably needs to be updated for this.
- Run `make container` as usual. I've only tested `ddev-webserver` and `ddev-dbserver` so far.

**Test images**
- For the images that I was able to build, I'm uploading test images for my Surface Pro X (Windows ARM64 with WSL2) to https://hub.docker.com/u/dennisameling.


containers/ddev-dbserver/Makefile
```diff
@@ -46,7 +46,7 @@ build: container
 
 # The format here is majorversion:pinnedimageversion
 # This allows to pin the version (as currently required with 8.0.19
-MARIADB_VERSIONS=5.5:5.5 10.0:10.0 10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
+MARIADB_VERSIONS=10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
```
dennisameling, line 49: This is just a quick fix to get the images to build. If 5.5 and 10.0 support are required moving forward, we can consider adding some conditionals to the build script.

dennisameling:
Just tried to update CircleCI to use [ubuntu-1604:202007-01](https://circleci.com/docs/2.0/configuration-reference/#available-machine-images), which bumps the Docker version from `v18.09.3` to `v19.03.12` (we need this for `buildx`). Used the instructions at https://github.com/docker/buildx/#building. Let's see if it works 👀

rfay:
Yay, nice work! Sorry a PR will also be required for [drud/ddev-images](https://github.com/drud/ddev-images), not sure if that was true when we last talked about this.

dennisameling:
Ah, that was exactly the one I was looking for! Thanks, will check it 🚀

dennisameling:
So, `buildx` is now working in CircleCI for arm64 🎉 we'll just need to fix the MariaDB/MySQL-specific issues that pop up. Will continue work at a later stage due to time constraints - to anyone who reads this, feel free to try/test/make changes.

containers/ddev-dbserver/Makefile
rfay: We do need 5.5 and 10.0 I think. So conditionals will be the ticket.

dennisameling:
@rfay I just made the following changes:
- Changed webserver image as requested in https://github.com/drud/ddev-images/pull/16: `FROM dennisameling/ddev-php-base:20200817 as ddev-webserver-base` - now the build process indeed pulls the `arm64` image so that's a great start 🚀 
- Got `Unable to locate package blackfire-php` on `arm64` so I excluded that for now.
- Made a custom build for `mkcert` arm64 as it's not available yet, created a PR in their repo: https://github.com/FiloSottile/mkcert/pull/284
- Same story for `MailHog` (just couldn't provide a PR there as they seem to release manually): https://github.com/mailhog/MailHog/issues/307

I was able to build an `arm64` image locally, also pushed it to Docker Hub: https://hub.docker.com/repository/docker/dennisameling/ddev-webserver/tags?page=1 🚀 . The image now boots on my `arm64` device, only MailHog is crashing on boot. Think I'll need to create another build for that as it was quite complex to get their repo set up quickly and correctly 😉 

I don't have any more time to work on this until the weekend. Just [added a separate step](https://github.com/drud/ddev/pull/2454/commits/31199bf05c794279aca3ed35fec9cc47e3913e52) to CircleCI to build only `arm64` separately. That way the images will build + test on `amd64` and if all that passes, it'll try to create a build for `arm64`. Unfortunately we can't run tests on `arm64` yet as there's no `arm64` hosts available in CircleCI.

rfay:
Fantastic, thanks! Great progress. I'll be on vacation from Friday for a week, probably much of the time without internet, so if I'm not reacting that's why. I hope that doesn't get in your way. Thank you so much for taking the lead on this!

dennisameling:
@rfay Enjoy your vacation! I should be able to continue with the other images without any feedback needed. Have a large backlog of other work, so not sure when I'll be able to spend some more time on this, but will report back in this PR. Only hope that we can get https://github.com/drud/ddev/pull/2462 merged before your vacation (hoping all tests will pass 🤞), as that'll make things a bit easier for me 😊

rfay:
Yes, I'll merge #2462 as soon as tests pass this morning.

dennisameling:
@rfay update: got things working on my `arm64` device now (webserver, router and ssh-agent) 🚀 

If you want to test you'll need to have the following:
- In `pkg/version.go`:
```Go
// RouterImage defines the image used for the router.
var RouterImage = "dennisameling/ddev-router"

// RouterTag defines the tag used for the router.
var RouterTag = "20200829-test" // Note that this can be overridden by make

var SSHAuthImage = "dennisameling/ddev-ssh-agent"

var SSHAuthTag = "20200823-test"
```
- In `.ddev/config.yaml`:
```yaml
omit_containers: [db, dba]
webimage: dennisameling/ddev-webserver:20200829-test2
```

Also added a separate step to the CircleCI config to just build the `arm64` images without testing them (maybe you can hook up your Raspberry Pi to CircleCI to run tests?). At least from Azure Pipelines I know they have an ARM64 agent which can be installed on e.g. Raspberry Pi https://github.com/Microsoft/azure-pipelines-agent/releases.

The `arm64` images [only take 7 mins to build](https://app.circleci.com/pipelines/github/drud/ddev/2677/workflows/36a73f22-8dea-4a8c-881b-28ad1f1b2acd/jobs/27243) (webserver, router, ssh-agent) so that's reasonable I think.

rfay:
I temporarily updated https://github.com/drud/ddev/pull/2474#issuecomment-683816031 to use your docker images, see https://github.com/drud/ddev/pull/2474#issuecomment-683816031 - artifacts are at https://app.circleci.com/pipelines/github/drud/ddev/2683/workflows/ce711e0e-e277-41d0-ac43-ba94bc27216e/jobs/27268/artifacts

rfay:
https://twitter.com/randyfay/status/1300570109166010368 - It's WORKING!

build-tools/makefile_components/base_container_arm64.mak
```diff
@@ -0,0 +1,28 @@
+# Container section of standard makefile (arm64)
```
rfay, line 1: The stuff in build-tools is not to be edited here, it's essentially an include. 

That said, DDEV-Local is now the only relevant consumer of build-tools, so it's my opinion that we should move the piece-parts  that we need out of build-tools/makefile_components and into the Makefile.

containers/ddev-router/Dockerfile
```diff
@@ -6,22 +6,50 @@ ENV DEBIAN_FRONTEND noninteractive
...
     apt-get -qq install --no-install-recommends --no-install-suggests -y \
         ca-certificates certbot curl less python-certbot-nginx procps telnet vim wget && \
     apt-get autoremove -y && \
     apt-get clean -y && \
 	rm -rf /var/lib/apt/lists/*
 
-ADD https://github.com/jwilder/forego/releases/download/v0.16.1/forego /usr/local/bin/forego
+RUN if [[ $TARGETPLATFORM == "linux/arm64" ]]; then \
+        # jwilder/forego doesn't have an arm64 download so we resort to the original forego repo
+        wget -q https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-arm64.tgz && \
```
rfay, line 25: As I studied forego for the cert/LetsEncrypt PR I started to understand that it is just another way to approach the problems we solve with supervisord in ddev-webserver. IMO we should switch to using supervisord. I'm just mentioning this in passing because I note that you hit this problem. No action is required.

containers/ddev-router/Dockerfile
```diff
@@ -6,22 +6,50 @@ ENV DEBIAN_FRONTEND noninteractive
...
+        curl -sSL https://github.com/jwilder/forego/releases/download/v0.16.1/forego -o /usr/local/bin/forego && \
+        chmod ugo+x /usr/local/bin/forego; \
+    fi
 
-RUN wget -q https://github.com/jwilder/docker-gen/releases/download/$DOCKER_GEN_VERSION/docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz && \
-    tar -C /usr/local/bin -xvzf docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz && \
-    rm /docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz && \
-    chmod ugo+x /usr/local/bin/forego && \
-    mkdir -p /etc/nginx/certs /mnt/ddev-global-cache/mkcert
+RUN if [[ $TARGETPLATFORM == "linux/arm64" ]]; then \
```
rfay, line 34: I think it would probably be best to group all the arm64  in a single if/else instead of doing it multiple times.

rfay:
Next time you work on this, could you please rebase the codebase and push images? 

I have invited you to the ddevlocal team on hub.docker.com, so once you accept the invite, you'll also be able to push to the ddev-local images, so we won't have to go through the dance of putting them in your user space.

dennisameling:
Thanks! I was away for the weekend, but planning to continue work on this in the coming weekend :)

rfay:
I'll try to make https://github.com/drud/ddev/issues/2507 - reorganizing the build - a priority in the coming week. The intent there isn't to actually do the build but to reorganize Makefile and deployment so they're aware of multiple-architecture issues. If I can get that sorted out, it will make both this and #2474 less hacky in those areas.

dennisameling:
Just pushed to the `drud/ddev-*` repos on Docker Hub:
- drud/ddev-webserver:20200913_test_version
- drud/ddev-router:20200913_test_version
- drud/ddev-ssh-agent:20200913_test_version

Didn't get to work on the other things unfortunately, have blocked some time in the coming days. Focus will mainly be on grouping the if/else logic and the `dbserver` Docker image. Keep you posted

dennisameling:
@rfay think I found a nice approach for the CI scenario and creating multi-arch builds. Just tested with `ddev-webserver`: https://github.com/drud/ddev/pull/2454/commits/943a25cb7a5c86364ef8bd4006a02debab02c80d. 

The approach:
- First build an `amd64`-only image with `make container`. The `amd64` build is cached for later stages, so we don't have to build twice 🚀 Notice `-o type=docker` here: it makes the image available to the local Docker instance, e.g. when running tests.
- Run `make test`. This will run tests against the `amd64` image.
- Run `make multi-arch`. This will create a build for all architectures that are set in the `BUILD_ARCHS` variable. Since `linux/amd64` is cached, we'll save time here. Benefit of this step instead of going to `make push` directly, is that builds will be created for all architectures in PRs as well (`make push` only runs when releasing a new version).
- Run `make push`. This will use the cached builds and pushes them to Docker Hub. If no cached builds are available for whatever reason, they will be created on the spot, then pushed right after.

The code can be optimized here and there, but in general I think this is a good approach moving forward (also if additional architectures need to be supported in the future). If this sounds OK to you, I'll go ahead and implement this for the other images as well :)

rfay:
Sounds reasonable. Please look also at the alternate approach in https://github.com/drud/ddev-images/pull/18/files - It does a push if logged in. 

That's actually my own workflow most of the time.

OTOH, an explicit push (as I *removed* in https://github.com/drud/ddev-images/pull/18/files) is nice... because it's explicit.

rfay:
I keep being tempted to add in the arm(32) build, but I know it will just make things slower. 

But there's no way to run the builds one-at-a-time for non-amd64 right?

rfay:
You can ignore the ddev-containers-windows build for now. I'll probably just remove that from the builds anyway.  I didn't realize that `docker buildx` was only in edge versions of docker desktop right now.

dennisameling:
I was able to create an arm64 Docker image for MariaDB 😍 🎉 

Docker Hub: https://hub.docker.com/repository/docker/dennisameling/ddev-dbserver-mariadb-10.2/tags?page=1
In `config.yaml`, e.g. use `dbimage: dennisameling/ddev-dbserver-mariadb-10.2:v1.16.0-alpha2-33-g16c506d8-dirty`

This is the result: 

```
denni@DESKTOP-8HTP3NV:~/repos/ddev-test$ ddev mysql
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 9
Server version: 10.2.33-MariaDB-1:10.2.33+maria~bionic-log mariadb.org binary distribution

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [db]>
```

One thing I'm not sure about is how to deal with the fact that only MariaDB supports arm64 currently, MySQL only builds for amd64 currently (https://hub.docker.com/_/mysql?tab=tags). GitHub issue for that: https://github.com/docker-library/mysql/issues/318. I now fixed it quick & dirty in https://github.com/drud/ddev/pull/2454/commits/73cf730d43dc45606ff237c58a83f924636bbb28 by creating a `MARIADB_VERSIONS_arm64` variable, but there must be a better way to do this. @rfay any suggestions?

Regarding your question for building arm32/building one-at-a-time: this is possible with `docker buildx build --platform=PLATFORM_NAME`. Only when pushing the images, e.g. with `docker buildx build --push platform=PLATFORM_NAMES_SEPARATED_BY_COMMA`, do you need to build them all at once. But it's no problem if they were built separately before, it will just use the build cache then. Is that what you mean?

rfay:
Yay for mariadb 10.2! Congrats!

It does look like there's a mysql 8.0 (only) at https://hub.docker.com/r/mysql/mysql-server (as you also noted)

On the one hand it's ugly to have to work around so many things. On the other hand, this is all experimental and will serve a limited audience at first, so it's OK if some features are pared back. Let's just do our best to make sure it's clean enough that we can clean it up and add the features (like old mysql or mariadb) back in without making a mess if they become available.

.circleci/linux_docker_buildx_setup.sh
```diff
@@ -0,0 +1,18 @@
...
+mkdir -p ~/.docker/cli-plugins
+mv docker-buildx ~/.docker/cli-plugins
+chmod a+x ~/.docker/cli-plugins/docker-buildx
+
+# We need this to get arm64 qemu to work https://github.com/docker/buildx/issues/138#issuecomment-569240559
+docker run --rm --privileged docker/binfmt:66f9012c56a8316f9244ffd7622d7c21c1f6f28d
+
+docker buildx create --name ddev-builder-multi
+docker buildx use ddev-builder-multi
+docker buildx inspect --bootstrap
```
rfay, line 18: These are needed for a local push as well, so IMO should be in the Makefile. Here's what I did in the ddev-images Makefile, I think it's going to work out OK. https://github.com/drud/ddev-images/blob/5b04b136cf05e61aae55219d3322687164d72eee/Makefile#L54-L58

The other Linux setup does definitely belong here.

build-tools/makefile_components/base_container_arm64.mak
```diff
@@ -0,0 +1,28 @@
...
+SANITIZED_DOCKER_REPO = $(subst /,_,$(DOCKER_REPO))
+
+DOTFILE_IMAGE = $(subst /,_,$(IMAGE))-$(VERSION)
+
+container-arm64: .container-arm64-$(DOTFILE_IMAGE) container-arm64-name
+
+# In CI environments, use the plain Docker build progress to not overload the CI logs
+PROGRESS := $(if $(CI),plain,auto)
+
+.container-arm64-$(DOTFILE_IMAGE): $(wildcard Dockerfile Dockerfile.in) container-arm64-name
```
rfay, line 16: The Dockerfile.in stuff is from ancient time before multi-stage builds were available, so completely irrelevant and obsolete. I understand where it's coming from, but it's probably easier to just have `Dockerfile` here instead of the wildcard statement.

build-tools/makefile_components/base_container_arm64.mak
```diff
@@ -0,0 +1,28 @@
...
+    # UPSTREAM_REPO in the Dockerfile.in will be changed to the value from Makefile; this is deprecated.
+    # There's no reason not to just use Dockerfile now.
+	@if [ -f Dockerfile.in ]; then sed -e 's|UPSTREAM_REPO|$(UPSTREAM_REPO)|g' Dockerfile.in > .dockerfile; else cp Dockerfile .dockerfile; fi
```
rfay, line 19: This is the old stuff pre-multistage-build, so can be dropped out (or not if you don't want to)

build-tools/makefile_components/base_container_arm64.mak
```diff
@@ -0,0 +1,28 @@
+# Container section of standard makefile (arm64)
+##### PLEASE DO NOT CHANGE THIS FILE #####
```
rfay, line 2: I'm removing the build-tools include from the main ddev Makefile, and consolidating needed parts of build-tools/makefile_components into that Makefile. You're welcome to do that here if it makes things easier, but at some point we have to have this logic in the proper Makefile, and of course, it's not legit to edit build_tools/makefile_components as long as we're using build-tools, since it's an upstream repo. Anyway, you can either continue this direction and I'll do a PR to remove build-tools or if it's easier you can consolidate yourself.

containers/ddev-dbserver/Makefile
```diff
@@ -46,29 +46,44 @@ build: container
 
 # The format here is majorversion:pinnedimageversion
 # This allows to pin the version (as currently required with 8.0.19
-MARIADB_VERSIONS=5.5:5.5 10.0:10.0 10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
-MYSQL_VERSIONS=5.5:5.5 5.6:5.6 5.7:5.7 8.0:8.0.19
+MARIADB_VERSIONS_amd64=5.5:5.5 10.0:10.0 10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
+MYSQL_VERSIONS_amd64=5.5:5.5 5.6:5.6 5.7:5.7 8.0:8.0.19
+MARIADB_VERSIONS_arm64=10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
```
rfay, line 51: This strategy seems OK to me for now.

rfay:
I used your technique for push in https://github.com/drud/ddev-images/pull/18, thanks for that. It's better.

dennisameling:
I just did a refactoring of the build process (https://github.com/drud/ddev/pull/2454/commits/417d58d87b73be398417d32461b64a6e3f0ee1d7), similar to what was done in `ddev-images`. 

Some remarks:
- Added extra step `make multi-arch` to CircleCI config after `make test`
- Added `prep` and `multi-arch` to all separate Makefiles. I realize this is a lot of duplicate code, but I couldn't quickly come up with a better solution for now. At least it moves the new bits out of `build-tools` for now.
- **Haven't worked on `make push` yet, only focusing on getting CI to pass now**
- Will work on "single if/else for arm64" in the Dockerfiles in the coming days

Very curious to see if CircleCI will pass now 🤞

rfay:
Wow, nice work. 

It's possible we'll want to add back in some equivalent to build-tools, but for now duplication is where we're at I guess. build-tools just got obsolete over time, and was mostly only used for ddev, so updating meant a lot of steps.

dennisameling:
@rfay CircleCI is failing: https://app.circleci.com/pipelines/github/drud/ddev/2841/workflows/24ccda4a-662d-4d20-bb6b-2b74c3e49abf/jobs/27901

Would it help to rerun the process? I don't have rights for that. It's failing at the MySQL 5.6 test which I didn't touch AFAIK:

```
# testserver
# dbserver_test-5450_9167
bats tests failed for mysql 5.6
Makefile:104: recipe for target 'test' failed
make: *** [test] Error 5

Exited with code exit status 2
CircleCI received exit code 2
```

rfay:
No, I don't think it will help to rerun, because the bats tests are failing. You should be able to use `make test` to debug locally,  and see the same results. 

But it looks like things just aren't working right in the build to begin with

```
# mysqlcheck: Got error: 1045: Access denied for user 'root'@'localhost' (using password: NO) when trying to connect
# FATAL ERROR: Error during call to mysql_check for fixing the db/tables names on all db(s) except mysql
```

That is happening because it's trying to do a mysql_upgrade when it probably shouldn't:

```
# + echo 'Attempting mysql_upgrade because db server version 5.6 is not the same as database db version unknown'
```

But this would all imply that we're working on versions of mariadb that I think we can't work on right? Didn't we say we couldn't do pre-10.1 versions of mariadb?

dennisameling:
Okay will test locally then. DB images have almost finished building after 30+ mins 😂 The tests will only run on the `amd64` image since that's the host OS. Cross-compilation to `arm64` only happens after the tests have finished. I'll check tomorrow since it's getting late in NL 🌙

rfay:
I also added you to github for this project, so that should mean you'll be able to restart tests in CircleCI. You have to accept the invite though (in your email)

rfay:
It's easy to demonstrate the failure locally, but the good news is it's only mysql 5.6 that's failing.

You can temporarily comment out 5.6 from test.sh, and we can solve this later. It seems that the image build is not building a valid database for a starter database, and it's too late for me to understand why. 

Anyway, you have a path forward with it, and THANKS!

dennisameling:
`lx_container_test` has passed 😍  https://app.circleci.com/pipelines/github/drud/ddev/2844/workflows/6e7f6831-9ea4-4b8e-acac-6504c69d4fdf/jobs/27915

Will work on the remaining bits ASAP, hopefully tonight already. Let's get this pushed over the finish line 🚀

rfay:
@dennisameling I was able to track down what was going on with 5.6, but don't understand how it could happen. The most likely cause I can understand is buildkit doing something wrong with caching. Anyway, the image's /var/lib/mysql/mysql directory was getting created even though mysql hadn't been initialized... and *only* for 5.6. So probably a caching problem... but it happened also with --no-cache.

Anyway, you should be able to undo the exclusion of 5.6 and use the docker-entrypoint.sh from #2523 (which may already be in upstream/master by the time you get this; I'll pull it when it passes tests)

Please go ahead and update pkg/version/version.go to the versions being created/used in this PR, so the ddev created here can work with the versions created here. that's a good debugging technique. This does still require manual pushing, so `cd containers and for item in *; do pushd $item; make push VERSION=<this_version>  ; popd; done`

Also, I recognize the error of my way in removing build-tools without an alternative. The right thing to do is to include what's needed probably from something in a subdirectory of the containers directory. I was so eager to get rid of build-tools that I wasn't thinking about the fact that we need a replacement for it, just for ddev.

rfay:
I pulled #2523 so it should be a pretty easy rebase against upstream/master now.

dennisameling:
@rfay update (https://github.com/drud/ddev/pull/2454/commits/9286fe226518ce22c3b83e9b89f1ce1269e4ba4f):
- Updated the `make push` commands so it's now possible to run `for item in *; do pushd $item; make push VERSION=<this_version> ; popd; done` which will push either `amd64` (MySQL + MariaDB 5.5 & 10.0) or `amd64 + arm64` (all other images) images
- Pushed all images to Docker Hub. Only MariaDB 10.5 didn't work for some reason due to `failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed`. Guess that's okay for now as 10.2 is the base version
- Updated `version.go` with the updated images
- Created a `ddev` executable for linux_arm64 with the updated versions in `version.go`: [ddev.zip](https://github.com/drud/ddev/files/5272018/ddev.zip)
- I can now run everything "as normal" on my Surface Pro X, including PHPMyAdmin and MailHog. No manual changes in `config.yaml` needed anymore 😍 
- CircleCI is failing on building MariaDB 10.1, while this works locally (`make multi-arch` in `containers/ddev-dbserver`), even with `--no-cache` https://app.circleci.com/pipelines/github/drud/ddev/2847/workflows/853d2b1e-a452-445a-9118-98cd5266c07b/jobs/27930, could you please restart the job? For me the "rerun" options are greyed out:
image

Now working on some optimizations in the logic/code. Quite confident that we can get close to finishing this PR tomorrow or on Friday 🚀

rfay:
You're awesome!

I've seen the `failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed` locally as well, it seems related to some `docker buildx build` push trouble on large pushes. A lot of moaning out there about it. I had to retry a couple of times. 

I can't restart a specific job until the full workflow has failed (unless I kill off the workflow first, but I'd rather see everything else succeeed). But I'll restart the db job when lx_nginx_fpm_teat finishes (I'm betting successfully!)

I hope we don't see too many of these weird buildx messages. I certainly saw more than I'd wish last night and today.

rfay:
I see you're still working away! I'm pushing mariadb 10.5, so hopefully everything will finish green or close this time!

Yay, [mariadb 10.5](https://hub.docker.com/layers/drud/ddev-dbserver-mariadb-10.5/20200929_arm64_testing/images/sha256-d02a05560655aa485bb6bee7182f16e13b0caac53cdae7a0dce4a0bdf9a97b83?context=repo) is up there

rfay:
We're doing pretty well, but the MariaDB 10.1 failure in the container build on Linux is problematic. I'll start it again, but we've already seen it a few times.

```
#12 [ 8/13] RUN /create_base_db.sh;
#12 1.223 Initializing mysql
#12 1.337 /usr/sbin/mysqld  Ver 10.1.46-MariaDB-1~bionic for debian-linux-gnu on aarch64 (mariadb.org binary distribution)
#12 1.512 version=10.1.46-MariaDB-1~bionic
#12 1.956 Installing MariaDB/MySQL system tables in '/var/lib/mysql' ...
#12 2.099 2020-09-23 23:08:51 274921006176 [Note] /usr/sbin/mysqld (mysqld 10.1.46-MariaDB-1~bionic) starting as process 113 ...
#12 7.866 Segmentation fault (core dumped)
```

It kind of looks like we have an invalid or wrong-arch mysqld, and when create_base_db.sh runs it, it croaks. It's definiltely the aarch64 version that is failing. 

I'm OK with dropping 10.1 out of the supported mariadb versions if that's what we need to do. It's old anyway. People do use it to solve problems with older versions of TYPO3 though.

However, I don't think that's what we need to do. We're able to build and push it, at least with a few tries, and it passes tests when built in my environment. So if we can't get by it we can leave it out of the tests.

Anyway, restarting it now.

dennisameling:
I just double-checked docker buildx requirements and found the following:

> For QEMU binaries registered with binfmt_misc on the host OS to work transparently inside containers they must be registed with the fix_binary flag. This requires a kernel >= 4.8 and binfmt-support >= 2.1.7

Howerver, Ubuntu 16.04 (CircleCI's latest image) mentions 2.1.6-1 as the latest version (see https://app.circleci.com/pipelines/github/drud/ddev/2848/workflows/17dca9a3-161c-44af-aa3d-b1d13dbd251a/jobs/27943 under "buildx setup")

> binfmt-support is already the newest version (2.1.6-1).

That might be our problem here. I'm now trying to update buildx and the `binfmt` Docker image in CI, let's see if that helps in any way. Otherwise I'm happy with dropping MariaDB 10.1 if that's indeed the only culprit for now.

.circleci/config.yml
```diff
@@ -272,9 +275,12 @@ jobs:
               time make container DOCKER_ARGS=--no-cache EXTRA_PATH=/home/linuxbrew/.linuxbrew/bin
               echo "--- Test container $dir"
               time make test EXTRA_PATH=/home/linuxbrew/.linuxbrew/bin
+              echo "--- Cross-compile container $dir"
+              time make multi-arch EXTRA_PATH=/home/linuxbrew/.linuxbrew/bin
```
dennisameling, line 279: Do we want this here? Currently, the script loops through all containers one by one and then 1) makes the container, 2) runs tests and 3) cross-compiles.

I think it makes more sense to have a separate CircleCI step "Linux container multi-arch" - if one or more tests fail, CircleCI will fail early, prior to cross-compiling for other architectures (now it's only arm64, but might be more in the future). Curious to hear what you think about this.

containers/ddev-dbserver/Dockerfile
```diff
@@ -11,6 +11,11 @@ ENV MYSQL_USER db
 ENV MYSQL_PASSWORD db
 ENV MYSQL_ROOT_PASSWORD root
 
+# TARGETPLATFORM is Docker buildx's target platform (e.g. linux/arm64), while 
+# BUILDPLATFORM is the platform of the build host (e.g. linux/amd64)
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
```
dennisameling, line 17: This is currently not used, will remove it

containers/ddev-dbserver/Makefile
```diff
@@ -46,43 +48,62 @@ build: container
...
+		export archs=$$(if [[ "$(MARIADB_VERSIONS_arm64)" == *"$$item"* ]]; then echo "linux/amd64,linux/arm64"; else echo "linux/amd64"; fi);\
+		docker buildx build --push --platform $${archs} $(DOCKER_ARGS) --progress=$(PROGRESS) --build-arg="DBVERSION=$${baseversion}" --build-arg="PARENTIMAGEVERSION=$$fullversion" --build-arg="DBTYPE=mariadb" -t "drud/ddev-dbserver-mariadb-$${baseversion}:$(VERSION)" .; \
 	done
-	for item in $(MYSQL_VERSIONS) ; do \
+	for item in $(MYSQL_VERSIONS_amd64) ; do \
 		set -euo pipefail ;\
 		export baseversion=$$(echo $$item | awk -F ':' '{print $$1}'); \
-		docker push "drud/ddev-dbserver-mysql-$${baseversion%%%.*}:$(VERSION)"; \
+		export fullversion=$$(echo $$item | awk -F ':' '{print $$2}'); \
+		docker buildx build --push --platform linux/amd64 $(DOCKER_ARGS) --progress=$(PROGRESS) --build-arg="DBVERSION=$$baseversion" --build-arg="PARENTIMAGEVERSION=$$fullversion" --build-arg="DBTYPE=mysql" -t "drud/ddev-dbserver-mysql-$${baseversion%%%.*}:$(VERSION)" .; \
```
dennisameling, line 103: This file has become quite a mess, but it works 😅 let me know if you have any feedback on this approach and if you'd do some things differently

containers/ddev-router/Dockerfile
```diff
@@ -6,22 +6,60 @@ ENV DEBIAN_FRONTEND noninteractive
...
+# PR for official arm64 support pending: https://github.com/FiloSottile/mkcert/pull/284 \n\
+MKCERT_URL="https://github.com/dennisameling/mkcert/releases/download/${MKCERT_VERSION}-ddev/mkcert-${MKCERT_VERSION}-linux-arm64"' > /tmp/ddev/vars; \
+
+    else \
+        printf 'FORGO_IS_TGZ="false" \n\
+FORGO_URL="https://github.com/jwilder/forego/releases/download/v0.16.1/forego" \n\
+DOCKER_GEN_URL="https://github.com/jwilder/docker-gen/releases/download/${DOCKER_GEN_VERSION}/docker-gen-linux-amd64-${DOCKER_GEN_VERSION}.tar.gz" \n\
+DOCKER_GEN_FILE="docker-gen-linux-amd64-${DOCKER_GEN_VERSION}.tar.gz" \n\
+MKCERT_URL="https://github.com/FiloSottile/mkcert/releases/download/$MKCERT_VERSION/mkcert-${MKCERT_VERSION}-linux-amd64"' > /tmp/ddev/vars; \
+    fi
```
dennisameling, line 34: Hope the approach above is OK for multi-arch support! Was the only proper way I could find to share variables cross-build steps

containers/ddev-router/Dockerfile
```diff
@@ -6,22 +6,60 @@ ENV DEBIAN_FRONTEND noninteractive
...
+    if [[ $TARGETPLATFORM == "linux/arm64" ]]; then \
+        printf '# jwilder/forego doesnt have an arm64 download so we resort to the original forego repo \n\
+FORGO_IS_TGZ="true" \n\
+FORGO_FILENAME="forego-stable-linux-arm64.tgz" \n\
+FORGO_URL="https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-arm64.tgz" \n\
+# jwilder/docker-gen doesnt have an arm64 download so we created a custom build \n\
+DOCKER_GEN_URL="https://github.com/dennisameling/docker-gen/releases/download/0.7.5/docker-gen-linux-arm64-0.7.5.tar.gz" \n\
+DOCKER_GEN_FILE="docker-gen-linux-arm64-0.7.5.tar.gz" \n\
+# PR for official arm64 support pending: https://github.com/FiloSottile/mkcert/pull/284 \n\
+MKCERT_URL="https://github.com/dennisameling/mkcert/releases/download/${MKCERT_VERSION}-ddev/mkcert-${MKCERT_VERSION}-linux-arm64"' > /tmp/ddev/vars; \
```
dennisameling, line 26: Shall we fork mkcert or do we wait for my PR in their repo to be merged?

containers/ddev-router/Dockerfile
```diff
@@ -6,22 +6,60 @@ ENV DEBIAN_FRONTEND noninteractive
...
+SHELL ["/bin/bash", "-c"]
+
+RUN mkdir /tmp/ddev && \
+    if [[ $TARGETPLATFORM == "linux/arm64" ]]; then \
+        printf '# jwilder/forego doesnt have an arm64 download so we resort to the original forego repo \n\
+FORGO_IS_TGZ="true" \n\
+FORGO_FILENAME="forego-stable-linux-arm64.tgz" \n\
+FORGO_URL="https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-arm64.tgz" \n\
+# jwilder/docker-gen doesnt have an arm64 download so we created a custom build \n\
+DOCKER_GEN_URL="https://github.com/dennisameling/docker-gen/releases/download/0.7.5/docker-gen-linux-arm64-0.7.5.tar.gz" \n\
```
dennisameling, line 23: Shall we fork docker-gen for the arm64 build? What do you think?

dennisameling:
Excluding MariaDB from 10.1 indeed did the trick :) left some comments above, curious to hear your feedback

.circleci/config.yml
rfay: I'm fine with separate CircleCI step "Linux container multi-arch".

containers/ddev-router/Dockerfile
rfay: I PM'd FiloSottile on Twitter the other day about this. Got no response. I was asking him if he'd be open to adding a maintainer for mkcert. I think he's quite a bigwig on golang at google, probably way too busy. But no response at all.

I think we have to fork. I'll fork in drud's namespace and give you privileges. 

The other option is just to maintain the builds we need. Maybe that's more sustainable? I'm not sure exactly where to put them. A fork is such a convenient place to put them.

rfay: I forked and made you maintainer. https://github.com/drud/mkcert

containers/ddev-router/Dockerfile
rfay: I forked and made you maintainer.

rfay: https://github.com/drud/forego

rfay:
mariadb 10.1 still failing with the same SEGV. 

1. We need to change the Makefile so a single version can be a target for both build and test. Obviously this lack has made it really difficult to build and debug specific things.
2. With that, we could build just the buildable versions in the circleCI test if we wanted to.

But still... I'd like to update CircleCI release build to *push* and if we can't get a successful build on 10.1 it's hard to do that.

dennisameling:
Wait... Did you check the most recent run? That one has passed: https://github.com/drud/ddev/pull/2454/checks?check_run_id=1159689700

rfay:
Sorry, fantastic. Github has been having some real problems with updating status on a page reload. Sometimes you have to reload a few times before it shows current status. And I apparently didn't.

containers/ddev-router/Dockerfile
dennisameling: BOOM and we have a release: https://github.com/drud/mkcert/releases/tag/v1.4.4

Instructions for creating a new release are here: https://github.com/drud/mkcert#creating-a-new-release

containers/ddev-webserver/Dockerfile
```diff
@@ -78,7 +88,12 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -o Dpkg::Options::="--for
     zip
 
 ADD ddev-webserver-base-files /
-RUN curl -sSL "https://github.com/mailhog/MailHog/releases/download/v${MAILHOG_VERSION}/MailHog_linux_amd64" -o /usr/local/bin/mailhog
+RUN if [[ $TARGETPLATFORM == "linux/arm64" ]]; then \
+        # Issue for official arm64 support: https://github.com/mailhog/MailHog/issues/307
+        curl -sSL "https://github.com/dennisameling/MailHog/releases/download/v${MAILHOG_VERSION}-ddev/MailHog_linux_arm64" -o /usr/local/bin/mailhog; \
```
dennisameling, line 93: Same for MailHog, can we also fork that? Forgot about this one, sorry

containers/ddev-router/Dockerfile
dennisameling: @rfay I actually need docker-gen instead of Forgo 😅

rfay: Sorry, got docker-gen forked for you.

containers/ddev-webserver/Dockerfile
rfay: mailhog also forked for you, thanks!

containers/ddev-router/Dockerfile
dennisameling: Done: https://github.com/drud/docker-gen/releases/tag/0.7.5

containers/ddev-webserver/Dockerfile
dennisameling: Done: https://github.com/drud/MailHog/releases/tag/v1.0.1-ddev

rfay: You are absolutely astonishing.

rfay:
The test failure in container tests is ddev-router not having docker-gen installed, and it's easy to demonstrate locally with `cd containers/ddev-router && make test`. The only thing is that because it fails to run and it's invoked with `docker run --rm` you can't see the logs unless you remove the `--rm`

But this is what's going on:
```
$ docker run  --name ddev-router-test -p 8080:80 -p 8443:443 --mount type=bind,src=/var/run/docker.sock,target=/tmp/docker.sock -v ddev-global-cache:/mnt/ddev-global-cache --name ddev-router-test -d drud/ddev-router:v1.16.0-alpha4-41-g8c2239b8
657fb1f316596f9fb6f96819d3bd0c558225c6b759c5f39bed4e45d767ac2c68

$ docker logs -f 657fb1f316596f9fb6f96819d3bd0c558225c6b759c5f39bed4e45d767ac2c68
Using the local CA at "/root/.local/share/mkcert" ✨
The local CA is now installed in the system trust store! ⚡️

Using the local CA at "/root/.local/share/mkcert" ✨

Created a new certificate valid for the following names 📜
 - "*.ddev.local"
 - "*.ddev.site"
 - "127.0.0.1"
 - "localhost"
 - "ddev-router"
 - "ddev-router.ddev_default"

Reminder: X.509 wildcards only go one level deep, so this won't match a.b.ddev.local ℹ️

The certificate is at "/etc/nginx/certs/.crt" and the key at "/etc/nginx/certs/.key" ✅

Using the local CA at "/root/.local/share/mkcert" ✨

Created a new certificate valid for the following names 📜
 - "*.ddev.local"
 - "*.ddev.site"
 - "ddev-router"
 - "ddev-router.ddev_default"
 - "127.0.0.1"
 - "localhost"

Reminder: X.509 wildcards only go one level deep, so this won't match a.b.ddev.local ℹ️

The certificate is at "/etc/nginx/certs/master.crt" and the key at "/etc/nginx/certs/master.key" ✅

forego     | starting nginx.1 on port 5000
forego     | starting dockergen.1 on port 5100
dockergen.1 | /bin/bash: /usr/local/bin/docker-gen: No such file or directory
forego     | starting dockergen.1 on port 5200
forego     | sending SIGTERM to dockergen.1
forego     | sending SIGTERM to nginx.1
```

I'm pretty sure the problem here is that the binary is dynamically linked, and it needs to be statically linked so it has no external dependencies. You'll note that it has dependences on /home/linuxbrew/.linuxbrew/lib/ld.so, which doesn't exist in this context of course.

```
# file docker-gen
docker-gen: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /home/linuxbrew/.linuxbrew/lib/ld.so, for GNU/Linux 3.2.0, Go BuildID=kXFM6l2if4EYnxMeB8Sb/KrWHn0D5kFElWWv6E30d/S826RzMkSjyBTv4IV5bJ/XOehIkiZbIE0T-lyKgsD, not stripped
```

One thing to add to the lovely github action that builds docker-gen would be executing it once, although it may have the ld.so in that context, not sure.

dennisameling:
Wow, thanks for finding this out! Learning a lot about Go haha 👀 to make it statically linked I had to add `CGO_ENABLED=0` to the build command, looks like that did the trick 🚀 

GitHub actions now works for `docker-gen` and automatically builds a release version (with `CGO_ENABLED=0`) when a new tag is created: https://github.com/drud/docker-gen/compare/0.7.5...drud:0.7.7#diff-b67911656ef5d18c4ae36cb6741b7965R20

rfay:
I think it's awesome that you went with github actions right away, I've been looking forward to moving some things to that and this opportunity is perfect. In a future PR it will make easy sense to move the db build and test into github actions.

rfay:
Yippee!

dennisameling:
I think we're good for a (final?) review now. Created a new file `containers_shared.mak` and moved away from build-tools. If you want, I can do one more push now and then update `version.go` with those versions.

I removed the `prep` stage from the Makefiles as I suspected it to cause some errors here and there, and `buildx` should work in Docker for Mac stable versions 2.1.0.1+ https://docs.docker.com/docker-for-mac/release-notes/#docker-desktop-community-2101 and by setting `experimental` to enabled in the `config.json` https://docs.docker.com/buildx/working-with-buildx/. Let me know if you want me to add it back, could add it to the `containers_shared.mak` then.

rfay:
As far as changing versions in versions.go, my usual workflow (which can certainly be improved) is to keep pushing to dockerhub the same tag (usually matching the branch name) all the way through the life of a PR, and not changing version.go.  When a version is out in the wild, I do not typically push an update to dockerhub.

Thanks for the info about pushing from macOS with experimental. Seems OK to me, I'll explore it. I can always push from Linux as well. I certainly found that I needed those pushing from edge. 

Feel free to git-push as needed, I don't think at this final phase anything will change the review. I imagine we'll find some nits anyway.

dennisameling:
⚠️ **ALL CHECKS HAVE PASSED** ⚠️ This is exciting!! 🚀 ✅ 

(btw I won't push new images as per your comment, they should be OK given the tests have passed)

build-tools/makefile_components/base_container.mak
```diff
@@ -8,6 +8,9 @@ SANITIZED_DOCKER_REPO = $(subst /,_,$(DOCKER_REPO))
 
 DOTFILE_IMAGE = $(subst /,_,$(IMAGE))-$(VERSION)
 
+# In CI environments, use the plain Docker build progress to not overload the CI logs
```
rfay, line 11: These changes to base_container.mak are irrelevant I think. No need to push a change though, because as long as we're not using this, it will come out in another PR.

build-tools/makefile_components/base_push.mak
```diff
@@ -6,7 +6,7 @@
 
 push: .push-$(DOTFILE_IMAGE) push-name
 .push-$(DOTFILE_IMAGE): .container-$(DOTFILE_IMAGE)
-	docker push $(DOCKER_REPO):$(VERSION)
```
rfay, line 9: Same, no need to change, but I'm assuming this code doesn't get included any more.

containers/containers_shared.mak
```diff
@@ -0,0 +1,25 @@
+# In CI environments, use the plain Docker build progress to not overload the CI logs
```
rfay, line 1: Well done.

containers/ddev-dbserver/Makefile
```diff
@@ -9,6 +9,8 @@
 DOCKER_REPO ?= drud/ddev-dbserver
 SHELL=/bin/bash
 
+.PHONY: prep
```
rfay, line 12: I think you removed the prep target.

containers/ddev-router/Makefile
```diff
@@ -8,6 +8,8 @@
 # Docker repo for a push
 DOCKER_REPO ?= drud/ddev-router
 
+.PHONY: prep
```
rfay, line 11: I think you removed the prep target.

containers/ddev-ssh-agent/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM bitnami/minideb:buster
+FROM debian:buster-slim
```
rfay, line 1: FYI I did a simple test of difference of size between using a base image of debian:buster-slim and debian:buster the other day when debugging one of the build problems we had (wondering if slim was a problem). It turned out that even though we add enormously to the image, there's still a 50MB or so benefit (as I remember) of using buster-slim.

containers/ddev-ssh-agent/Makefile
```diff
@@ -5,6 +5,8 @@
 # Docker repo for a push
 DOCKER_REPO ?= drud/ddev-ssh-agent
 
+.PHONY: prep
```
rfay, line 8: Was removed.

rfay (approved):
Looks great to me. The highest priority is not breaking anything that currently works mainline. I think the tests are helping us with that. I made minor comments. The .PHONY entries can be removed (if I understand right) and if you want to.

Let me know when you're ready to have me merge!

dennisameling:
Just applied your suggestions, thanks for those. I also removed the separate `multi-arch` step in CircleCI and moved it back into `linux container test`. It was definitely not using all cached builds from the previous steps for some reason, increasing the CI time significantly. Should be faster now.

If CI passes this should be ready to merge 🚀

dennisameling:
From 76m [down to 53m](https://app.circleci.com/pipelines/github/drud/ddev/2867/workflows/15831c43-b11e-41cf-8f03-85c338157784/jobs/28023) by moving `multi-arch` back to `linux container test` instead of in a separate step 🚀 that's quite OK I guess for build, test & multi-arch! In the future it should be possible to add a native `arm64` builder (like the Raspberry Pi 4) to the CI pipeline for better performance. For now, at least we have cross-compilation set up at least 😄

rfay:
Outstanding. If you get a chance could you do a PR to do the same thing on ddev-images?

dennisameling:
So, specifically, that'd mean adding `multi-arch` into `ddev-images` so that it not only runs on `push`, but also for PRs etc., right? Together with optimizing the Makefile.

rfay:
I'm not sure it has to run on PRs, that's so painful. Take a look at what's there and see what you think is appropriate to change. It's just that we've been learning a lot and ddev-images has been kind of frozen in place.

rfay:
Truly heroic @dennisameling ! Such an awesome effort. Thanks!

Issue #2403, opened 2020-07-20
RSS feeds are broken when using TYPO3 and Nginx


stucki:
**Describe the bug**
When using TYPO3 and Nginx, any RSS feeds that should be served by TYPO3 will return a HTTP 404 error.

**To Reproduce**

1. Open any (non-existing) RSS URL on the ddev webserver.
2. Alternatively, set up a news feed including RSS feed in TYPO3

**Expected behavior**
It should return the output as configured by TYPO3. Instead, what I get is a HTTP 404 error by Nginx.

**Additional context**
The issue seems to be caused by the `location ~* \.(?:rss|atom)$ {` block in `nginx_typo3.conf`. Defining a new `location` means that the `location / {` block won't have effect here:

https://github.com/drud/ddev/blob/3976b9431b9b87101ba2416268f6f3cbfe317ffd/pkg/ddevapp/webserver_config_packr_assets/nginx-site-typo3.conf#L52-L56

The fix is to duplicate the contents from this block and define it again. (Alternatively, the code block which causes the issue could be removed, but then the `expires` setting is gone...)

rfay:
Thanks so much for the report!

I assume this isn't new; I don't know of any changes that would have affected this.

You didn't fill out the required information about ddev version, etc. I assume this is ddev v1.15.0?

If you know the fix, you can easily demonstrate it by changing the .ddev/nginx_full/nginx-site.conf (don't forget to remove the #ddev-generated line at the top). Please do that and then submit a PR. 

But please provide full manual testing instructions.

Please give full recreation instructions

gilbertsoft:
I see no reason for this definition, if you setup a rss feed with TYPO3 the expiration is defined anyway by TYPO3 and we will have the next issue... Also the next lines concerning the caching of images should be removed.
Imho we safely can remove the lines https://github.com/drud/ddev/blob/3976b9431b9b87101ba2416268f6f3cbfe317ffd/pkg/ddevapp/webserver_config_packr_assets/nginx-site-typo3.conf#L52-L63, do you agree @stucki ?

stucki:
@gilbertsoft I agree that the feed rule can be removed. The other part about media rules won't hurt here, but it's also not needed on a local system. Anyway, it is not related to this problem.

@rfay Yes it's not new at all. I could reproduce this with ddev 1.15.3, but the configuration exists at least since 2018: https://github.com/drud/ddev/commit/e36b1f6eba25df2e761271d368282333db946306

[...]

PR #2461, opened 2020-08-19
Forward requests for *.rss + *.atom to PHP (#2403)

stucki:
## The Problem/Issue/Bug:
Defining expire headers for *.rss and *.atom in the Nginx configuration will skip execution of any other section in the Nginx configuration file. This means that RSS feeds processed by a PHP application won't be forwarded to PHP right now.

Remove expire header which is not needed here. It may be relevant on production servers to optimize performance. But then again, PHP should send the expire headers instead. Sites that contains static rss files will still work, but the expire header is just missing, which should be ok.

The PR removes the configuration not only for TYPO3 but also for any other configuration where the rule is currently set. Not sure if this is ok, but I think it makes sense.

- Create a TYPO3 website
- Request a random URL that ends with `.rss`, e.g. https://typo3.ddev.site/test.rss
- Without this fix, the request is answered by Nginx (it will return a 404 error)
- With this fix, the request should print a TYPO3 error about the site not being configured.

n/A

## Related Issue Link(s):
#2403 

## Release/Deployment notes:
n/A


rfay:
This looks totally reasonable to me. I pushed an additional commit to actually add those files to the binary (result of `make packr2`) so you'll need to `git pull` to pick that up.

rfay:
Interestingly enough, Drupal puts its rss feed at /rss.xml, so is unaffected by this.

stucki:
@rfay Thanks for the feedback! Looks good.

> Interestingly enough, Drupal puts its rss feed at /rss.xml, so is unaffected by this.

Well it's a different suffix, so that's no surprise. As long as there is no other location rule which matches `.xml` then the problem won't happen...

stucki:
@rfay Hmm could it be that TYPO3 is configured and sends its a 404 header, too? Please try dropping the database, so that thw whole site fails. You should get a TYPO3 error if the request was processed by PHP, and a simple Nginx error if not.

rfay:
The 404 I'm getting is in fact from TYPO3, 
Page_Not_Found

stucki:
@rfay Ok so that means that the fix works as expected. Without the fix, you wouldn't see a TYPO3 page here.

PR #2462, opened 2020-08-19
Update CircleCI version to ubuntu-1604:202007-01

dennisameling:
Needed for https://github.com/drud/ddev/pull/2454 - creating a separate PR to first upgrade the image and see if everything still works as it should in the current state.


rfay:
Tests have all passed on CircleCI, although it's not reported here for some reason.

PR #2466, opened 2020-08-20
Fix various errors in typo3 and typo3cms global commands, add wp command, fixes #2467

### LINKED ISSUE #2467: A global wp command (wp-cli)

```
#!/bin/bash

## Description: Run WordPress CLI inside the web container
## Usage: wp [flags] [args]
## Example: "ddev wp core version" or "ddev wp plugin install user-switching --activate"
## ProjectTypes: wordpress

wp "$@"
```

gilbertsoft:


rfay:
Pushed a commit to pack it into the binary.

Glad you checked those out! 

The alpha1 release was a mistake. I must have pushed the wrong button :(

rfay:
Since I was fiddling with packr2 and #2467 came in I hope you don't mind my hijacking this to add that in here. It's worth doing the packr2 as little as possible.

gilbertsoft:
> The alpha1 release was a mistake. I must have pushed the wrong button :(

I already thought so because it's not marked as pre release :(

rfay:
I did notice and check the prerelease button.

gilbertsoft:
> 
> 
> Since I was fiddling with packr2 and #2467 came in I hope you don't mind my hijacking this to add that in here. It's worth doing the packr2 as little as possible.

Yeah this one came up today in Slack and I advised the user to open an issue for this as it looks WP CLI should be included also. No problem with your hijacking ;)

Issue #2338, opened 2020-06-25
Support darwin/arm64 on Apple Silicon


rfay:
The announcement was made this week that macOS and mac hardware are going to switch to ARM chips.

This has major impact on DDEV-Local. We'll want to be prepared for the new hardware before it is released in Fall, 2020

- [x] Obtain ARM macOS hardware for testing with macOS 11 Big Sur beta
- [x] Prepare build techniques for Docker images (#2454)
- [x] Improve build techniques to build ddev fat binary with ARM (#2474)
- [x] Inform ddev-live cli build process of required changes
- [x] Make sure that all related resources are updated for ARM
    - [x] Docker Desktop
    - [x] mkcert (https://github.com/FiloSottile/mkcert/pull/284) - we may have to fork, as mkcert maintainer hasn't been responsive lately.

In order to have DDEV-Local ready for before new hardware comes out, we should
- [x] Subscribe to the [Universal App Quick Start Program](https://developer.apple.com/programs/universal/), which costs $500 and includes a loaner ARM machine which is to be returned after the program is complete.

Related: #2322 is about building binary and docker images, and it may come sooner.

unn:
FYI @cyberswat

rfay:
@cyberswat did submit the application for the program, thanks!

cyberswat:
Still have not received a response on this.

rfay:
The application was accepted, but when we got the email, it had like a 24-hour deadline and it was missed among all the other email. 

@cyberswat has resubmitted the application, hopefully we'll do better this time.

* I purchased a Raspberry Pi 4 (and installed 64-bit Raspberry Pi OS) for testing and was able to run ddev on it using https://github.com/drud/ddev/pull/2474 and https://github.com/drud/ddev/pull/2454 - we're making real progress there.
* I spun up an AWS EC2 Ubuntu machine and was able to run things fine (with no db container, those aren't done yet)

Quite a significant amount of build-reorganization will need to be done, as the assumption of amd64 is wired into all the builds.

rfay:
gotip (upcoming v1.16) now has darwin/arm64:
```
rfay@rfay-mbp-2017:~/sdk/gotip/bin$ go tool dist list | grep darwin
darwin/amd64
darwin/arm64
```

rfay:
We will purchase Apple Silicon as soon as it's available; gotip already builds arm64 binary, but of course we have nowhere to run it.

rfay:
[...]

PR #2474, opened 2020-08-22
Reorganize Makefile to build ARM binaries for ddev and stop using build-tools, fixes #2507, for #2338, fixes #2322

### LINKED ISSUE #2322: Build multi-arch (supporting ARM64) versions of ddev binary and docker images
DDEV currently doesn't work on Windows (WSL2) or Linux on ARM devices. We'd need Docker images for ARM64 in order to run DDEV on these platforms.

Docker images for ARM64 can potentially be built by using the `buildx` command https://www.docker.com/blog/multi-arch-images/, but this would need to be researched. Then, automated tests should run on an ARM64 host (either Linux or Windows WSL2) so that we can be sure that DDEV runs (and keeps running) properly on ARM64.

N/A

**Additional context**
N/A

### LINKED ISSUE #2507: Reorganize Makefile/build to support multiple architectures, remove use of build-tools (main binary build)
We need to support arm64 architecture, and perhaps arm32. This requires new naming conventions and build capabilities. Some of these have been temp-hacked in #2454  and #2474, but this needs to be done carefully thinking about the overall architecture of the build (and the CircleCI artifacts creation).

- [x] Use a real target (.gotmp/bin/windows_arm/ddev) instead of the fake target currently created with `touch`
- [x] DDEV_BINARY_FULLPATH needs to be correct for each architecture
- [x] Remove include of build-tools
- [x] Names of all artifacts should reflect the architecture. 
- [x] Verify that `make test` and `make staticrequired` and all other targets work (golangci-lint, darwin_signed, darwin_notarized, etc)
- [x] Can we successfully notarize darwin_arm64 ddev (assuming we can build it)? (Can't build it at this time, waiting for golang)
- [x] Can we make an installer for windows_arm machines? (No, can't build makensis at this point)
- [x] .circleci config should handle all relevant artifacts
- [ ] This also affects Quicksprint, which consumes the binaries, and affects the homebrew pushes and (perhaps?) Chocolatey pushes. 
- [x] I don't believe that homebrew is willing to accept arm bottles, so maybe it just won't happen.
- [x] Chocolatey doesn't seem to support ARM, see https://github.com/chocolatey/choco/issues/1803

rfay:
## The Problem/Issue/Bug:

This builds linux_amd64, linux_arm, windows_arm. 

We can't build darwin_arm64 yet, and windows_arm64 is not on the tool list. (`go tool dist list`)

See checklist in #2507


rfay:
I updated this to use @dennisameling 's images in version.go, so it will start to work (without db container at this point). 

Artifacts are at https://app.circleci.com/pipelines/github/drud/ddev/2683/workflows/ce711e0e-e277-41d0-ac43-ba94bc27216e/jobs/27268/artifacts

rfay:
https://twitter.com/randyfay/status/1300570109166010368 - It's WORKING!

rfay:
@dennisameling one thing I note that I hope won't discourage you. windows/arm64 is not listed as a target in `go tool dist list`, although windows/arm is there. Compiling for windows/arm isn't hard, but results in 32-bit AFAICT, see https://medium.com/@tux0010/compile-golang-for-arm64-on-an-arm64-machine-7a86c133430e

Also, https://build.golang.org/ does *not* show windows/arm as a currently built version. https://github.com/golang/go/issues/26148#issuecomment-494021129 questions whether it has been abandoned, with no response.

Important resources for building:

- [gist with variety of build options](https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63)
- `go tool dist list`
- [golang Port to windows/arm64](https://github.com/golang/go/issues/36439) issue
- [Use of GOARM to target various ARM families](https://github.com/golang/go/wiki/GoArm) (more complexity!) (It looks like GOARM is only necessary where GOARCH=arm (32-bit) so if we stick with arm64, it's not relevant. windows/arm is explicitly not listed as supported here)

dennisameling:
@rfay I don't think a DDEV executable for Windows arm64 should be a priority right now, as I expect most users to be working in WSL2 which uses arm64 Linux executables anyway. Windows arm64 can also run 32-bit applications, so if we could somehow get a x86 or ARM32 executable, that'd suffice for now :)

rfay:
Ah, glad to hear it! Thanks @dennisameling

rfay:
@dennisameling I think all the boxes are checked now. In some ways it's better than it's ever been, but in other ways it has so far to go.

.circleci/generate_artifacts.sh
```diff
@@ -52,33 +42,53 @@ $BUILTPATH/ddev_gen_autocomplete
...
 if [ -d chocolatey ]; then
-    tar -czf $ARTIFACTS/ddev_chocolatey.$VERSION.tar.gz chocolatey
+    tar -czf $ARTIFACTS/ddev_chocolatey_amd64-.$VERSION.tar.gz chocolatey
 fi
+popd
 
-cp ddev_windows_installer*.exe $ARTIFACTS
+# generate windows-arm tarball/zipball
+pushd $BASE_DIR/.gotmp/bin/windows_arm
+curl -sSL -o mkcert.exe https://github.com/drud/mkcert/releases/download/${MKCERT_VERSION}/mkcert-${MKCERT_VERSION}-windows-arm.exe
```
dennisameling, line 88: `windows-arm` is not part of the 1.4.4 release: https://github.com/drud/mkcert/releases. I tried with `windows-arm` but it doesn't seem to work on arm64. i386 (32-bit) does work however (32-bit emulation on arm64). So please:

- Update mkcert version to v1.4.6 (just released)
- use `-windows-386.exe` on windows-arm :)

dennisameling (commented):
I have the feeling that `windows-arm` won't work (at least it didn't work with mkcert, application crashed). Could we do `windows-386` (32-bit) for the time being until go supports arm64? 32-bit works through emulation on Windows arm64 :)

.circleci/generate_artifacts.sh
rfay: So you want windows/386 (See `go tool dist list`) ? I wonder if that will work without building 32-bit versions of the images? It may not be worth fighting this, of course. I was hoping that the windows-arm would work on your machine. 

I agree about pull windows/arm out of there. But do you think it's useful to have windows/i386?

dennisameling: That would be windows/386 indeed. Thing is, Windows arm64 doesn't even have the ability to run Docker Desktop, so it's actually a waste of effort for now. Users of Windows arm64 will need to be doing everything from within WSL2, so they'll be using DDEV Linux arm64 binaries in that case. Let's just remove windows_arm and re-consider in a few months 👍

rfay: Oh, I guess I didn't understand how you're running docker on your surface device. You must be running Linux docker inside WSL2?

I've removed windows/arm

dennisameling: Yes, correct :) no other way to run Docker currently on arm64 Windows devices 😊

rfay: I sure hear another blog post coming! :)

dennisameling: Haha, so much to blog about after this whole experience! Def will write something soon 🚀

rfay: We may want to collaborate on a single blog that explains windows/arm64, linux/arm64, raspberry pi/arm64, raspberry pi/arm. /cc @ekl1773

dennisameling: Yes sounds good!

rfay:
Doggone it @dennisameling I just now discovered that the library/mariadb upstream images have no arm/v7 builds.

dennisameling:
That's a shame regarding MariaDB :( hope they'll add such image soon!

PR #2489, opened 2020-09-03
Fix faq entry about required tools like PHP

AMartinNo1:
## The Problem/Issue/Bug:

> **Do I need to install PHP or Composer or Nginx or DDEV on my computer?**
> Absolutely *not*. All of these tools live inside DDEV's docker containers, so you need only Docker and DDEV. This is especially handy for Windows users where there's a bit more friction installing those tools.

The question lists among other DDEV. Yet the answer says in the first sentence "no, not necessary" but in the second it is listed in the two required tools (Docker and DDEV).

It removes the tool DDEV from the question which makes the answer valid.

No testing needed as there are no code changes involed.


AMartinNo1:
Happy to help 😃

PR #2493, opened 2020-09-05
Add the option to add flags to custom commands, fixes #2491

### LINKED ISSUE #2491: Add the option to add flags to custom commands
More and more bundled custom commands are providing flags but there is not option to describe this flags in the custom command.

There should be an option to add provided flags to the Cobra defintion (commandToAdd.Flags().BoolVarP() in commands.go should be sufficient).

Currently the flags can be shown in the example section.

**Additional context**

gilbertsoft:
## The Problem/Issue/Bug:
More and more bundled custom commands are providing flags but there is not option to describe this flags in the custom command using the cobra definition which prints the flags for a command.

A new option _Flags_ is recognized now to describe the flags with a JSON comment in the custom command.

Example:

`## Flags: [{"Name":"test","Shorthand":"t","Usage":"Usage of test"}]`

The _Shorthand_ is optional and can be omitted.

Create a new custom command and add a flags definition to the comment header see above or the detailed explanation in the docs.

The provided tests fully cover all cases including errors and panics.

## Related Issue Link(s):
Closes #2491 

## Release/Deployment notes:
Nothing special is needed.


rfay:
The failure in CircleCI is in https://app.circleci.com/pipelines/github/drud/ddev/2744/workflows/36fa626d-ad99-4c90-a739-670728b3ab85/jobs/27496 - 

You can run `make staticrequired` to verify gofmt issues, and I try to link up the pre-push hook, see https://github.com/drud/ddev/blob/master/.githooks/README.md - to avoid this happening. It's a little annoying waiting for the push, but OTOH better than finding out later that the build failed. I do the `ln -s ../../.githooks/pre-push.allchecks pre-push`

cmd/ddev/cmd/commands.go
```diff
@@ -3,6 +3,13 @@ package cmd
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+
```
gilbertsoft, line 12: This change was not really intended but made by the linter of VS Code. And looking to other packages this looks like a best practise to place interal packages on the top and gh imports afterwards.

gilbertsoft:
> 
> 
> The failure in CircleCI is in https://app.circleci.com/pipelines/github/drud/ddev/2744/workflows/36fa626d-ad99-4c90-a739-670728b3ab85/jobs/27496 -
> 
> You can run `make staticrequired` to verify gofmt issues, and I try to link up the pre-push hook, see https://github.com/drud/ddev/blob/master/.githooks/README.md - to avoid this happening. It's a little annoying waiting for the push, but OTOH better than finding out later that the build failed. I do the `ln -s ../../.githooks/pre-push.allchecks pre-push`

Yeah thanks! I've added more go extensions to VS Code which should also solve this issue :)

gilbertsoft:
@rfay Change is ready to review now. Will add the docs once #2494 and #2495 are merged....

cmd/ddev/cmd/flags.go
```diff
@@ -0,0 +1,357 @@
...
+	return nil
+}
+
+// AssignToCommand iterates the flags and assigns it to the provided command.
+func (f *Flags) AssignToCommand(command *cobra.Command) error {
+	var errors string
+
+	for _, flag := range f.Definition {
+		// Create the flag at the command
+		switch flag.Type {
```
gilbertsoft, line 319: As you can see from the list above there are many more types which could be implemented in the future if needed. For now, I've focused on the most important.

cmd/ddev/cmd/commands_test.go
```diff
@@ -2,6 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	osexec "os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
```
gilbertsoft, line 10: Same as above

cmd/ddev/cmd/flags_test.go
```diff
@@ -0,0 +1,193 @@
...
+	assert.EqualValues("1", f.DefValue)
+	assert.EqualValues("2", f.NoOptDefVal)
+	assert.Empty(f.Annotations)
+
+	// No fully implemented type
+	subject = getSubject()
+	c = getCommand()
+	assert.NoError(subject.LoadFromJSON(`[{"Name":"test","Usage":"Usage of test","Type":"_test3_"}]`))
+	assert.PanicsWithValue("Error implementation missing for type '_test3_'", func() {
+		assert.NoError(subject.AssignToCommand(&c))
```
gilbertsoft, line 184: As I'm checking here for a panic this check makes imho no sense but `make staticrequired` complains if this is missing. Same above btw.

docs/users/extend/custom-commands.md
```diff
@@ -8,8 +8,6 @@ There are example commands provided in `ddev/commands/*/*.example` that can just
 
 * The command filename is not what determines the name of the command.  That comes from the Usage doc line (`## Usage: commandname`).
```
gilbertsoft, line 9: @rfay  Just saw this line and from reading the code it's not checked the user provides this value and we also do not provide a default value here. So if this sentence is correct a command without usage should / must fail.

gilbertsoft: I had a short look at the cobra command code and this works like written here. So I suggest we use the file name as default value. What do you think? Other solution would be we introduce a validation like done for the flags which would be a good long term solution but is a little bit more work ofc. If you don't mind I will prepare a solution here and clean up the whole part in commands.go.

gilbertsoft:
@rfay examples to test the flags can be found in the tests as starting point

docs/users/extend/custom-commands.md
rfay: The filename is already the default. If you drop a random shell script into .ddev/commands/web it will show up with that name.

cmd/ddev/cmd/flags_test.go
rfay: Wow, I've never seen an assert for a panic! 

You can use `//nolint: <linter>` to specify the lint-checking behavior you want, for example `// nolint: errcheck`. 

It's not clear to me what makes no sense (the func()?) but anyway, you can turn off linting for a specific linter.

docs/users/extend/custom-commands.md
gilbertsoft: Ah okey, so the docs are not fully correct here. But I think we can leave it untouched for the moment.

cmd/ddev/cmd/flags_test.go
gilbertsoft: It makes no sense to check the return value of the method because it does not return anything but should panic instead.

Yeah, checking for a panic is ofc an edge case. But here I like to make sure the method panics to be sure dev errors are discovered during the CI :)

But no need to change the linter behavior here but nice to know there are ways to fix such cases for the future. Thanks for that hint!

rfay: Ah, if it's errcheck that's the issue, just use `_ = whateverfunction();` and all the linters will be quiet. Easier than `//nolint errcheck`. "Throw away the return value. Yes I know I'm doing it."

cmd/ddev/cmd/flags.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// LoadFromJSON imports the defs provided by the custom command as json into
+// the flags structure.
+func (f *Flags) LoadFromJSON(data string) error {
+	if data == "" {
+		return nil
+	}
+
+	var err error
```
rfay, line 294: Explicitly defining err isn't necessary or particularly useful. Just use err := ... on the first error. OTOH, it doesn't actually do any harm.

rfay (changes requested):
Minor change requests. 

Have you looked at the performance impact at all? ddev is already doing way too much with these commands every time ddev is executed, which is already worrisome.

I do see a persistent failure, didn't debug it, it's the fail that's breaking the tests:

```
Error 'invalid character '[' looking for beginning of object key string', command 'test' contains an invalid flags definition '{[{"Name":"test","Shorthand":"t","Usage":"Usage of test"}]', skipping add flags of /Users/rfay/workspace/d9composer/.ddev/.global_commands/web/test
```

gilbertsoft:
> Have you looked at the performance impact at all? ddev is already doing way too much with these commands every time ddev is executed, which is already worrisome.

No, I did not. But saw this behavior during the development which is really not so nice. I have an idea how we can improve this situation but I first have to check if this works at all. Should be done in a next or prior change imho.

> 
> I do see a persistent failure, didn't debug it, it's the fail that's breaking the tests:
> 
> ```
> Error 'invalid character '[' looking for beginning of object key string', command 'test' contains an invalid flags definition '{[{"Name":"test","Shorthand":"t","Usage":"Usage of test"}]', skipping add flags of /Users/rfay/workspace/d9composer/.ddev/.global_commands/web/test
> ```

This is intended but ofc the test should not fail. But I was not aware about a failing test, will have a deeper look.

docs/users/extend/custom-commands.md
```diff
@@ -104,13 +99,92 @@ Useful variables for container scripts are:
...
+
+Usage: `## Usage: <command-usage>`
+Example: `## Usage: commandname [flags] [args]`
+
+#### Example
+
+`Example` is used for the help message to provide some usage examples to the user. Use `\n` to force a line break.
+
+Usage: `## Example: <command-example>`
+Example: `## Example: commandname\commandname -h`
```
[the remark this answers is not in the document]
gilbertsoft: Nope, this was a typo of mine

cmd/ddev/cmd/flags.go
gilbertsoft: Yes, I'm aware of this. I did it like this because the var is used multiple times and for me it's better readable like this for this case.

cmd/ddev/cmd/flags.go
```diff
@@ -0,0 +1,357 @@
...
+type Flags struct {
+	CommandName string
+	Script      string
+	Definition  FlagsDefinition
+}
+
+func formatErrorItem(level int, prefix string, format string, a ...interface{}) error {
+	return fmt.Errorf("%s %s %s", strings.Repeat(" ", (level-1)*2), prefix, fmt.Sprintf(format, a...))
+}
+
```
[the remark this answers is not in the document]
gilbertsoft: Yes, will do...

rfay:
I created this global web command named "test"
```
#!/bin/bash

## Description: test
## Usage: test
## Flags: {[{"Name":"test","Shorthand":"t","Usage":"Usage of test"}]
## Example: "ddev test"

echo "test was executed with args=$@ on host $(hostname)"
```

But `ddev help test` doesn't tell me anything about flags:
```
$ ddev test --help
test (global shell web container command)

Usage:
  ddev test [flags]

Examples:
  "ddev test"

Flags:
  -h, --help   help for test

Global Flags:
```

gilbertsoft:
I did the same like you and get this expectet output:

```ps
PS E:\drud\ddev\.gotmp\bin\windows_amd64> .\ddev test -h
Error 'invalid character '[' looking for beginning of object key string', command 'test' contains an invalid flags definition '{[{"Name":"test","Shorthand":"t","Usage":"Usage of test"}]', skipping add flags of E:\drud\ddev\.gotmp\bin\windows_amd64\.ddev\commands\web\test
test (shell web container command)

Usage:
  ddev test [flags]

Examples:
  "ddev test"

Flags:
  -h, --help   help for test

Global Flags:
```

There is a error in the JSON, the starting `{` should be removed. Then I get the following:
```ps
PS E:\drud\ddev\.gotmp\bin\windows_amd64> .\ddev test -h
test (shell web container command)

Usage:
  ddev test [flags]

Examples:
  "ddev test"

Flags:
  -h, --help   help for test
  -t, --test   Usage of test

Global Flags:
```

Did you really not get an error message like me? That would mean yousomehow use the wrong binary...

rfay:
OK. I took that from https://github.com/gilbertsoft/ddev/blob/21082e1b30b802728df129530884dd182b16f600/cmd/ddev/cmd/testdata/TestCustomCommands/project_commands/host/testhostcmdflags#L5, and I must have messed it up somehow. 

I'm not entirely sure that people will be comfortable with json here but I'm OK to go with it.

gilbertsoft:
> OK. I took that from https://github.com/gilbertsoft/ddev/blob/21082e1b30b802728df129530884dd182b16f600/cmd/ddev/cmd/testdata/TestCustomCommands/project_commands/host/testhostcmdflags#L5, and I must have messed it up somehow.

Yeah, this one was buggy :(

> I'm not entirely sure that people will be comfortable with json here but I'm OK to go with it.

I'm open for other ideas here but it seems to be the easiest. And nowadays people should be able to write such simple JSON. To support less experienced people I added the whole validation thing which outputs as clear as possible error messages for user misstakes...

gilbertsoft:
Two good and working examples can be found here:
- https://github.com/drud/ddev/pull/2493/files#diff-8d9431d2887832d2de5cfd81bbfe6495R48
- https://github.com/drud/ddev/pull/2493/files#diff-8d9431d2887832d2de5cfd81bbfe6495R58

rfay:
Those links aren't helpful, sorry. Must be a copy/paste problem?  Please link to code in your fork rather than to a files diff in PR. 

Here's what https://github.com/drud/ddev/pull/2493/files#diff-8d9431d2887832d2de5cfd81bbfe6495R48 results in: 

Add_the_option_to_add_flags_to_custom_commands__fixes__2491_by_gilbertsoft_·_Pull_Request__2493_·_drud_ddev

gilbertsoft:
> Those links aren't helpful, sorry. Must be a copy/paste problem? Please link to code in your fork rather than to a files diff in PR.

Links are fine but the browser jumps here to the wrong line. Here the two examples I'd like to show you:

- `[{"Name":"test","Usage":"Usage of test"}]`
- `[{"Name":"test-1","Shorthand":"t","Usage":"Usage of test 1","Type":"string","DefValue":"true","NoOptDefVal":"true","Annotations":{"test-1":["test-1-1","test-1-2"]}},{"Name":"test-2","Usage":"Usage of test 2","Type":"bool","DefValue":"true","NoOptDefVal":"true","Annotations":{"test-2":["test-2-1","test-2-2"]}}]`

The annotation in the second one is not useful, it was meant for testing only. But the enhanced docs have a link to spf13 where this is explained if you like to test it. It's for autocompletion only so not so imortant in the first place but I thought if somebody likes to use it e.g. to define possible file extensions for a file name paramter...

docs/users/extend/custom-commands.md
```diff
@@ -104,13 +99,92 @@ Useful variables for container scripts are:
...
+#### Example
+
+`Example` is used for the help message to provide some usage examples to the user. Use `\n` to force a line break.
+
+Usage: `## Example: <command-example>`
+Example: `## Example: commandname\ncommandname -h`
+
+#### Flags
+
+`Flags` is used for the help message. All defined flags here are listed with their shorthand if available. It has to be encoded according the following definition:
```
rfay, line 127: This should be a bit stronger. I know you're planning another PR or two, so let's not delay this one, but when you next work on commands, let's make this say something like "THIS DOES NOT DEFINE THE FLAG HANDLING, ONLY WHAT COMES OUT IN THE HELP"

rfay (approved):
Looking good! Thanks for all the great work on this.

docs/users/extend/custom-commands.md
gilbertsoft: Yes, agree...

PR #2494, opened 2020-09-06
Fix missing global prefix of command suffix

gilbertsoft:
## The Problem/Issue/Bug:
The global prefix in the command suffix of global commands is missing.

This patch fixes the condition and adds tests to avoid regressions in this place.

run `ddev` and check if the global prefix is added to global commands suffixes e.g.

`Run TYPO3 CLI (typo3) command inside the web container (global shell web container command)`

The new tests are checking the correct behavior of global host and web commands. Also a new test is provided to ensure global host commands gets properly overwritten by the project ones.

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -135,7 +135,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 				}
 
 				descSuffix := " (shell " + service + " container command)"
-				if serviceDirOnHost[0:1] == "." {
+				if commandSet == targetGlobalCommandPath {
 					descSuffix = " (global shell " + service + " container command)"
 				}
 				commandToAdd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/ddev/cmd/commands.go
```diff
@@ -135,7 +135,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 				}
 
 				descSuffix := " (shell " + service + " container command)"
-				if serviceDirOnHost[0:1] == "." {
+				if commandSet == targetGlobalCommandPath {
```
rfay, line 138: The serviceDirOnHost[0:1] was completely wrong. It wasn't just Windows, glad you got this fixed!

rfay (approved):
Nice job with both the fix and the test update. Thanks!

cmd/ddev/cmd/commands.go
rfay: Also, the way you did it is far more readable, although this whole function is pretty obtuse.

PR #2495, opened 2020-09-06
Allow multiline examples for custom commands

gilbertsoft:
## The Problem/Issue/Bug:
Currently the examples section of a custom command does only support a one line description which ends in confusing examples. Examples in multiple line is highly desirable.

This patch introduces the possibility of using `\n` in the examples to force a line break.

Change the examples comment and use `\n` to force a new line and run `ddev [command] -h` to see the new output.

The line breaks and fromatting are tested in a new test to avoid regressions.

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -108,7 +108,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 					usage = val
 				}
 				if val, ok := directives["Example"]; ok {
-					example = val
+					example = "  " + strings.ReplaceAll(val, `\n`, "\n  ")
 				}
 				if val, ok := directives["ProjectTypes"]; ok {
 					projectTypes = val
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -2,6 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	osexec "os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/exec"
 	"github.com/drud/ddev/pkg/fileutil"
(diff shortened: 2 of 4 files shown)


docs/users/extend/custom-commands.md
```diff
@@ -102,6 +102,16 @@ Useful variables for container scripts are:
...
+### Annotations supported
+
+The custom commands support various annotations in the header which are used to provide additional information about the command to the user:
+
+* Description
+* Usage
+* Example (use `\n` to force a line break)
+* ProjectTypes
+* OSTypes
+
```
gilbertsoft, line 114: This whole section will be changed again with #2493 . There I will provide an extended description of all annotations.

cmd/ddev/cmd/commands_test.go
```diff
@@ -2,6 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	osexec "os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
```
gilbertsoft, line 11: This change was not really intended but made by the linter of VS Code. And looking at other packages this looks like a best practise to place interal packages on the top and gh imports afterwards.

rfay: I always let goland put them where they belong. That's probably what you're seeing. 

I doubt you should be using any linter beyond gofmt, which I think also organizes them.

cmd/ddev/cmd/commands.go
```diff
@@ -108,7 +108,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 					usage = val
 				}
 				if val, ok := directives["Example"]; ok {
-					example = val
+					example = "  " + strings.ReplaceAll(val, `\n`, "\n  ")
```
rfay, line 111: I wish there was an easier way to do this (for the command writer) but I'm fine with this. It's far easier to parse things as lines than otherwise.

cmd/ddev/cmd/commands_test.go
gilbertsoft: No, gofmt does not organize them as I've seen with my tests. I'm wondering about the difference we have here between your editor and VS Code

rfay (approved):
Fantastic improvement, thanks! I manually tested and it did great, and the code looks great.

cmd/ddev/cmd/commands.go
gilbertsoft: Yeah same thought. There are ways but I guess this is the easiest solution also for the users.

PR #2498, opened 2020-09-09
Update Atom XDebug Instructions

WidgetsBurritos:
## The Problem/Issue/Bug:

The `php-debug` package requires `atom-ide-ui`, but `atom-ide-ui` has it’s own debugger system which isn’t compatible with `php-debug`. So you essentially have to disable the debugger that comes with `atom-ide-ui` to use the `php-debug` debugger. After doing that everything works.

It updates the XDebug + Atom documentation to provide instructions on doing this

Step through setup instructions on Atom and ensure Xdebug works

Documentation change only, so no tests.

## Related Issue Link(s):
https://github.com/gwomacks/php-debug/issues/310#issuecomment-419751273

## Release/Deployment notes:
The `atom-ide-ui` dependency adds a few features to Atom, so it might slightly change the user experience, but overall it should be pretty seamless.

### DIFF
--- a/docs/users/snippets/atom_config_cson_snippet.txt
+++ b/docs/users/snippets/atom_config_cson_snippet.txt
@@ -1,3 +1,7 @@
+  "atom-ide-ui":
+    "atom-ide-debugger": {}
+    use:
+      "atom-ide-debugger": "never"
   "php-debug":
     PathMaps: [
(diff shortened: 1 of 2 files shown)


rfay:
Thanks for this @WidgetsBurritos !

PR #2504, opened 2020-09-10
ddev-router should not accept hostnames it's not configured for, fixes #2345

### LINKED ISSUE #2345: PHPMyAdmin page remains active (with contents of another project), after ddev stop issued
**Describe the bug**
When i say `ddev stop` to stop a project i can still go to the PHPMyAdmin Page.

**Screenshots**
If i type `ddev describe` it looks like this: 

Bildschirmfoto 2020-06-29 um 20 06 57

**Versions**
DDEV Version: v1.15-rc1 
Operating System: MacOS Catalina
Docker Version: 19.03.8
Docker-Compose: Version: 1.25.5

But it happens not on every project.

And it seems, when i type `ddev start` on `project1`, it also starts `project2`.

rfay:
## The Problem/Issue/Bug:

#2345 points out that if you run the PHPMyAdmin interface or mailhog UI, but then stop the project that it's referring to, it still shows there. 

It turns out that *any* server on *any* port besides 80 and 443 demonstrates this behavior, it's not just mailhog and phpmyadmin. If you use `router_http_port: "8080"` in a project, then stop that project, anything else listening on 8080 will show up.

Rework the catch-all (not matched) server stanza in ddev-router to list *all* available ports, so that it can properly redirect any of those if they're invalid. 

- [x] Start two projects, preferably two projects with dissimilar databases so you can see the difference easily
- [x] Open the PHPMyadmin interface for one of them with `ddev launch -p`
- [x] Stop that project.
- [x] Referesh the PHPMyAdmin page. You should get "no backend" error message.


rfay:
The artifacts for this PR are at https://app.circleci.com/pipelines/github/drud/ddev/2760/workflows/252d6e80-289d-464f-a17c-b51da905f989/jobs/27564/artifacts

PR #2541, opened 2020-10-07
Images of omitted containers should not be pulled, fixes #2262

### LINKED ISSUE #2262: phpMyAdmin image gets pulled even if the container is globally omitted
**Description**
According to [documentation](https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/#global_configyaml-options) it's possible to omit containers on a global basis. This makes sense for e.g. the `dba` container. Adding `dba` to the `omit_containers` in `$HOME/.ddev/global_config.yaml` however doesn't prevent ddev from pulling the image (~ 500MB for me) and storing it on the hard drive. As the image will never be used, because the container will never be build, it's unecessary to pull it. 

**To Reproduce**
- Run: `docker images | grep "phpmyadmin"`
- If found, copy the `IMAGE ID`
- Run: `docker image rm <IMAGE ID>`
- Add `omit_containers: [dba]` to `$HOME/.ddev/global_config.yaml `
- Start any project with `ddev start`
- ddev still pulls the image

**Expected behavior**
I would epexct that ddev doesn't pull the image if the container is globally omitted.

NBZ4live:
## The Problem/Issue/Bug:
Images for omitted containers are pulled from registry

Prevents pulling of images for omitted containers

## Related Issue Link(s):
#2262

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gobuffalo/packr/v2"
 	"github.com/lextoumbourou/goodhosts"
 	"github.com/mattn/go-isatty"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh/terminal"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
Awesome!

Just two things:
1. db container can be omitted (in project) so it should be handled differently
2. Probably some output (if globalconfig.DdevVerbose util.Success()?) on pull that will make it easier to test. It's hard to test as is because nothing happens

Since it's getting more complex, you may want to make a static map[imagename]imagespec that will allow you to put it in one loop and test, for example images := map[string]string{"web":app.WebImage, "db": app.DBImage, etc}

rfay (approved):
This is exactly what I was hoping for, thanks!

I made minor stylistic changes, but this is good to go. Thanks!

rfay:
Thanks for this!

PR #2548, opened 2020-10-09
Discourage users more strongly from modifying DDEV's settings.ddev.php

chellman:
## The Problem/Issue/Bug:

Maybe this was just me, but I thought it was fine to modify the settings.ddev.php file in a Drupal project after the initial config.  I know it does technically work, but I've found out it's not recommended.  I think the comment template, and docs, should be updated to make this a smidge clearer.

Added a line to the docs, and modified the comment templates for Drupal, Typo3, and WordPress settings files (the only ones I saw that seemed applicable).

Only comments changed, so as long as the build succeeds, all should be well.

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -385,7 +385,7 @@ And you can now visit your working project by URL or just give the command `ddev
 
 ### Configuration files
 
-_**Note:** If you're providing the settings.php or wp-config.php and DDEV is creating the settings.ddev.php (or wp-config-local.php, AdditionalConfig.php, or similar), the main settings file must explicitly include the appropriate DDEV-generated settings file._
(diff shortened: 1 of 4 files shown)


rfay:
Excellent. 

I'm thinking maybe we should add something that says "Make your changes to the settings.php after the include of settings.ddev.php?)

How about also improving https://ddev.readthedocs.io/en/stable/users/cli-usage/#configuration-files ?

chellman:
I added a new section called "Adding Configuration" under that Configuration files section, with some (I think correct) practices for adding config for Drupal, Backdrop, and WordPress.  I don't use Typo3 or Magento, so I'm not sure what should go there.

rfay (approved):
Thanks for this! I rebased and made minor edits. Good to go.

PR #2558, opened 2020-10-15
Fix config sync directory for Drupal 8 websites that are using a custom sync directory

shaal:
## The Problem/Issue/Bug:
Drupal 8.9.6 website, that is using a custom deprecated (but still supported) setup for configuration sync directory, gets overridden by ddev default setup.
```
CONFIG_SYNC_DIRECTORY = 'sync'
$config_directories[CONFIG_SYNC_DIRECTORY] = `/var/www/html/config/sync`
```

Rearranging the conditions that check if a configuration sync directory already exist.

I am thinking testing for this, would require 6 scenarios, installing Drupal Umami + Drush 10, and then run `drush cex -y`.

(no custom config directory)
Drupal 8.7 Umami
Drupal 8.8 Umami
Drupal 9.0 Umami

(with custom config directory)
Drupal 8.7 Umami
Drupal 8.8 Umami
Drupal 9.0 Umami

Some of these tests should fail with the conditions set in ddev `master`, but pass with the new set of conditions (in this patch)

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -118,14 +118,21 @@ $settings['trusted_host_patterns'] = ['.*'];
 $settings['class_loader_auto_detect'] = FALSE;
 
 // This specifies the default configuration sync directory.
-// $config_directories (pre-Drupal 8.8) and
-// $settings['config_sync_directory'] are supported
-// so it should work on any Drupal 8 or 9 version.
-if (defined('CONFIG_SYNC_DIRECTORY') && empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
-  $config_directories[CONFIG_SYNC_DIRECTORY] = '{{ joinPath $config.SitePath $config.SyncDir }}';
+// If a configuration sync directory wasn't set yet, ddev is going to set a default one.
+// Ddev checks first which setup (Drupal 8 or Drupal 9) is being used.
+
+// Pre-Drupal 8.8 configuration
+if (defined('CONFIG_SYNC_DIRECTORY')) {
+  if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
+    // Set the configuration sync directory if it is not set for the project.
+    $config_directories[CONFIG_SYNC_DIRECTORY] = 'sites/default/files/sync';
+  }
 }
+
+// Post-Drupal 8.8 configuration
 elseif (empty($settings['config_sync_directory'])) {
-  $settings['config_sync_directory'] = '{{ joinPath $config.SitePath $config.SyncDir }}';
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/drupal.go
```diff
@@ -118,14 +118,24 @@ $settings['trusted_host_patterns'] = ['.*'];
...
+    if (empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
+      // Set a default configuration sync directory.
+      $config_directories[CONFIG_SYNC_DIRECTORY] = 'sites/default/files/sync';
+    }
+  }
+  // Post-Drupal 8.8 configuration is needed.
+  else {
+    // Set a default configuration sync directory.
+    $settings['config_sync_directory'] = 'sites/default/files/sync';
+  }
```
rfay, line 138: This ends up using CONFIG_SYNC_DIRECTORY on D8.8/8.9, which doesn't seem like best practice. 

I suggest this, which is very clear and chooses based on version. 

```suggestion
// For D8 before 8.8.0, we set $config_directories[CONFIG_SYNC_DIRECTORY] if not set
if (version_compare(Drupal::VERSION, "8.8.0", '<') &&
  empty($config_directories[CONFIG_SYNC_DIRECTORY])) {
  $config_directories[CONFIG_SYNC_DIRECTORY] = 'sites/default/files/sync';
}
// For D8.8/D8.9, set $settings['config_sync_directory'] if neither 
// $config_directories nor $settings['config_sync_directory is set
if (version_compare(DRUPAL::VERSION, "8.8.0", '>=') &&
  version_compare(DRUPAL::VERSION, "9.0.0", '<') &&
  empty($config_directories[CONFIG_SYNC_DIRECTORY]) &&
  empty($settings['config_sync_directory'])) {
  $settings['config_sync_directory'] = 'sites/default/files/sync';
}
// For Drupal9, it's always $settings['config_sync_directory']
if (version_compare(DRUPAL::VERSION, "9.0.0", '>=') &&
  empty($settings['config_sync_directory'])) {
  $settings['config_sync_directory'] = 'sites/default/files/sync';
}
```

I tested this on D8.7, 8.9, and 9.0 and it seemed to behave correctly.

rfay (changes requested):
Please take a look at the suggestion. If it seems correct, you can just click the button to commit it.

pkg/ddevapp/drupal.go
shaal: 👍 
I really like it! and it's clear to read and understand.

rfay:
Tested again on all 3 cases, with and without modifications, seems OK.

rfay (approved):

PR #2568, opened 2020-10-20
Add Travis config for arm64 tests, fixes #2543

### LINKED ISSUE #2543: arm64 cleanup tasks: Exclude mysql, old mariadb versions, run tests
**Describe the bug**

DDEV-Local now supports linux/arm64 and the binaries are provided, but:

- [x] Only mariadb 10.1+ is supported because there are no packages/upstream docker images for mariadb <10.1 or mysql (any version). But ddev doesn't know this.
- [x] Tests don't know not to run with those values (Fixed in #2568)
- [x] Tests are not yet run on arm64 (fixed in #2568)

/cc @dennisameling

dennisameling:
As discussed in https://github.com/drud/ddev/issues/2543#issuecomment-706209596, this PR tries to add a TravisCI config for running tests on arm64 Linux. It's using the brand-new [arm64-graviton2 processor](https://docs.travis-ci.com/user/multi-cpu-architectures/#multi-cpu-availaibility) which should result in good performance 🚀 

First test run: https://travis-ci.com/github/dennisameling/ddev/builds/191266714

Small note: Linuxbrew is not supported on arm64 ([no binary packages](https://docs.brew.sh/Homebrew-on-Linux#arm)), and mkcert is not available for arm64 as well. I therefore used our custom mkcert version: https://github.com/drud/mkcert/releases/download/v1.4.6/mkcert-v1.4.6-linux-arm64. 

The config likely needs some tweaks so that it will support more tests. I'll try to take inspiration from https://github.com/drud/ddev/blob/master/.circleci/linux_circle_vm_setup.sh where applicable.

- [x] Please add Travis to the `drud/ddev` repo so that it can run here as well 😊

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,9 @@ jobs:
      virt: lxd
      group: edge
 
+before_install:
+  # Install mkcert
+  - curl -sSL https://github.com/drud/mkcert/releases/download/v1.4.6/mkcert-v1.4.6-linux-arm64 -o /usr/local/bin/mkcert && chmod +x /usr/local/bin/mkcert
+  - mkcert -install
+
 script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,3 +58,4 @@ jobs:
           popd
         done
         popd
+        set +eu
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
   global:
     - BUILDKIT_PROGRESS=plain
     - GOTEST_SHORT=true
+    - DOCKER_CLI_EXPERIMENTAL=enabled
 
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,11 @@ language: go
 go:
 - 1.x
 
-jobs:
-  include:
-   - os: linux
-     arch: arm64-graviton2
-     virt: lxd
-     group: edge
+arch: arm64-graviton2
+virt: lxd
+os: linux
+dist: focal
+group: edge
 
 before_install:
   # Install mkcert
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,17 +16,6 @@ env:
 
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
-  # Configure environment so changes are picked up when the Docker daemon is restarted after upgrading
-  - echo '{"experimental":true}' | sudo tee /etc/docker/daemon.json
-  - export DOCKER_CLI_EXPERIMENTAL=enabled
-  # Upgrade to Docker CE 19.03 for BuildKit support
-  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-  - sudo add-apt-repository "deb [arch=arm64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-  - sudo apt-get update
-  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
-  # Show info to simplify debugging and create a builder
-  - docker info
-  - docker buildx create --name ddev-builder-multi --use  
 
 # Group tests to break up travis-CI build so it doesn't hit time limit.
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ jobs:
     - name: container tests
       script: |
         set -eu -o pipefail
-        pushd containers
         for dir in containers/*/
           do pushd $dir
           echo "--- Build container $dir"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ os: linux
 dist: focal
 group: edge
 
+env:
+  global:
+    - BUILDKIT_PROGRESS=plain
+
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
   # Configure environment so changes are picked up when the Docker daemon is restarted after upgrading
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,5 +38,4 @@ jobs:
           time make test
           popd
         done
-        popd
         set +eu
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
       script: make test
     - name: Linux container tests
       script: |
+        set -eu -o pipefail 
         for dir in containers/*/
           do pushd $dir
           echo "--- Build container $dir"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ jobs:
 
 before_install:
   # Install mkcert
-  - curl -sSL https://github.com/drud/mkcert/releases/download/v1.4.6/mkcert-v1.4.6-linux-arm64 -o /usr/local/bin/mkcert && chmod +x /usr/local/bin/mkcert
+  - sudo curl -sSL https://github.com/drud/mkcert/releases/download/v1.4.6/mkcert-v1.4.6-linux-arm64 -o /usr/local/bin/mkcert && sudo chmod +x /usr/local/bin/mkcert
   - mkcert -install
 
 script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,8 +60,9 @@ jobs:
       script: make testcmd
     - name: Linux container tests
       script: |
-        set -eu -o pipefail 
-        for dir in containers/*/
+        set -eu -o pipefail
+        pushd containers
+        for dir in ddev-dbserver ddev-router ddev-ssh-agent ddev-webserver
           do pushd $dir
           echo "--- Build container $dir"
           time make container DOCKER_ARGS=--no-cache
@@ -69,5 +70,6 @@ jobs:
           time make test
           popd
         done
+        popd
 
 script: make testpkg
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,15 +37,13 @@ before_install:
 #
 jobs:
   include:
-    - stage: test
-      name: test
-    - name: make testcmd
+    - name: testcmd tests
       script: make testcmd
-    - name: make testnotddevapp
+    - name: testpkg tests except ddevapp
       script: make testnotddevapp
-    - name: make testddevapp
+    - name: ddevapp tests
       script: make testddevapp
-    - name: Container tests
+    - name: container tests
       script: |
         set -eu -o pipefail
         pushd containers
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,11 @@ before_install:
   - echo '{"experimental":"enabled"}' | sudo tee /etc/docker/daemon.json
   - mkdir -p $HOME/.docker
   - echo '{"experimental":"enabled"}' | sudo tee $HOME/.docker/config.json
-  - sudo service docker start
+  - export DOCKER_CLI_EXPERIMENTAL=enabled
+  - sudo service docker restart
 
 install:
+  - docker info
   - docker buildx create --name ddev-builder-multi --use
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ dist: focal
 group: edge
 
 before_install:
-  # Install mkcert
-  - sudo curl -sSL https://github.com/drud/mkcert/releases/download/v1.4.6/mkcert-v1.4.6-linux-arm64 -o /usr/local/bin/mkcert && sudo chmod +x /usr/local/bin/mkcert
-  - mkcert -install
+  - ./.travis/linux_travis_arm64_setup.sh
 
 script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: go
+
+go:
+- 1.x
+
+arch:
+- arm64-graviton2
+
+script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,4 +70,4 @@ jobs:
           popd
         done
 
-script: make test
+script: make testpkg
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ group: edge
 env:
   global:
     - BUILDKIT_PROGRESS=plain
+    - GOTEST_SHORT=true
 
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
@@ -27,11 +28,36 @@ before_install:
   - docker info
   - docker buildx create --name ddev-builder-multi --use  
 
+# Group tests by directory to logically break up travis-CI build.
+#
+# Here's the bash if you need to update this. Be sure to maintain the 
+# lines that are currently commented out (these cannot run in Travis)
+#   cd pkg
+#   find . -maxdepth 1 -type d | sed 's|./|    - env: TESTDIR=|' | sed 's/$//g'
+#
 jobs:
   include:
     - stage: test
-      name: Run make test
-      script: make test
+      name: testpkg
+      env: TESTDIR=nodeps
+    - env: TESTDIR=appports
+    - env: TESTDIR=fileutil
+    - env: TESTDIR=version
+    - env: TESTDIR=ddevhosts
+    - env: TESTDIR=exec
+    - env: TESTDIR=output
+    - env: TESTDIR=globalconfig
+    - env: TESTDIR=util
+    - env: TESTDIR=testcommon
+    - env: TESTDIR=appimport
+    - env: TESTDIR=ddevapp
+    - env: TESTDIR=netutil
+    - env: TESTDIR=servicetest
+    - env: TESTDIR=updatecheck
+    - env: TESTDIR=archive
+    - env: TESTDIR=dockerutil
+    - name: make testcmd
+      script: make testcmd
     - name: Linux container tests
       script: |
         set -eu -o pipefail 
@@ -43,3 +69,5 @@ jobs:
           time make test
           popd
         done
+
+script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jobs:
       script: |
         set -eu -o pipefail
         pushd containers
-        for dir in ddev-dbserver ddev-router ddev-ssh-agent ddev-webserver
+        for dir in containers/*/
           do pushd $dir
           echo "--- Build container $dir"
           time make container DOCKER_ARGS=--no-cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ go:
 - 1.x
 
 arch: arm64-graviton2
-virt: lxd
+virt: vm
 os: linux
 dist: focal
 group: edge
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: go
 go:
 - 1.x
 
-services:
-  - docker
-
 arch: arm64-graviton2
 virt: vm
 os: linux
@@ -14,6 +11,13 @@ group: edge
 
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
+  - echo '{"experimental":"enabled"}' | sudo tee /etc/docker/daemon.json
+  - mkdir -p $HOME/.docker
+  - echo '{"experimental":"enabled"}' | sudo tee $HOME/.docker/config.json
+  - sudo service docker start
+
+install:
+  - docker buildx create --name ddev-builder-multi --use
 
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,18 @@ group: edge
 before_install:
   - ./.travis/linux_travis_arm64_setup.sh
 
-script: make test
+jobs:
+  include:
+    - stage: test
+      name: Run make test
+      script: make test
+    - name: Linux container tests
+      script: |
+        for dir in containers/*/
+          do pushd $dir
+          echo "--- Build container $dir"
+          time make container DOCKER_ARGS=--no-cache
+          echo "--- Test container $dir"
+          time make test
+          popd
+        done
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: go
 go:
 - 1.x
 
+services:
+  - docker
+
 arch: arm64-graviton2
 virt: vm
 os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,8 @@ before_install:
   - mkdir -p $HOME/.docker
   - echo '{"experimental":"enabled"}' | sudo tee $HOME/.docker/config.json
   - export DOCKER_CLI_EXPERIMENTAL=enabled
-  - sudo service docker restart
-
-install:
   - docker info
-  - docker buildx create --name ddev-builder-multi --use
+  - docker buildx create --name ddev-builder-multi --use  
 
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,6 @@ go:
 arch:
 - arm64-graviton2
 
+virt: lxd
+
 script: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,11 @@ language: go
 go:
 - 1.x
 
-arch:
-- arm64-graviton2
-
-virt: lxd
+jobs:
+  include:
+   - os: linux
+     arch: arm64-graviton2
+     virt: lxd
+     group: edge
 
 script: make test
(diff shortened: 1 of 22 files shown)


dennisameling:
These tests failed (https://travis-ci.com/github/dennisameling/ddev/builds/191266714):

--- FAIL: TestDdevStartMultipleHostnames (95.94s)
--- FAIL: TestDdevXdebugEnabled (5.83s)
--- FAIL: TestGetApps (110.74s)
--- FAIL: TestDdevAllDatabases (235.62s)
--- FAIL: TestDdevFullSiteSetup (184.42s)
--- FAIL: TestDdevExec (69.63s)
--- FAIL: TestHttpsRedirection (20.10s)
--- FAIL: TestMultipleComposeFiles (0.61s)
--- FAIL: TestWebserverType (20.48s)
--- FAIL: TestNFSMount (3.69s)
--- FAIL: TestHostDBPort (39.77s)
--- FAIL: TestPortSpecifications (20.48s)
--- FAIL: TestEnvironmentVariables (0.00s)
--- FAIL: TestRouterConfigOverride (14.85s)
--- FAIL: TestWriteDrushConfig (35.91s)
--- FAIL: TestSSHAuth (0.01s)
--- FAIL: TestSshAuthConfigOverride (14.91s)

Will take a closer look during the weekend. If anyone wants to work on this before that, please feel free to do so 😊

rfay:
I'm fine with not using linuxbrew in the setup script.

rfay:
I added drud/ddev to travis-ci

dennisameling:
So the good news is that the failing NFS test is solved now with some tweaks ✅ 

I see some errors related to MariaDB 5.5 and PHP 5.6:

MariaDB 5.5 (error makes sense because there's no `arm64` image, it'll need to be excluded from tests on arm64):
```
standard_init_linux.go:211: exec user process caused "exec format error"
```

PHP 5.6 (I thought 5.6 is available for arm64 now, right?):
```
update-alternatives: error: alternative /usr/bin/php5.6 for php not registered; not setting
```

Any thoughts on this or required actions to move forward?

--- FAIL: TestDdevStartMultipleHostnames (68.27s)
--- FAIL: TestDdevXdebugEnabled (4.89s)
--- FAIL: TestGetApps (110.13s)
--- FAIL: TestDdevAllDatabases (200.40s)
--- FAIL: TestDdevFullSiteSetup (119.65s)
--- FAIL: TestDdevExec (73.91s)
--- FAIL: TestHttpsRedirection (13.28s)
--- FAIL: TestMultipleComposeFiles (0.40s)
--- FAIL: TestWebserverType (19.47s)
--- FAIL: TestHostDBPort (37.15s)
--- FAIL: TestPortSpecifications (20.99s)
--- FAIL: TestEnvironmentVariables (0.00s)
--- FAIL: TestRouterConfigOverride (12.49s)
--- FAIL: TestWriteDrushConfig (33.39s)
--- FAIL: TestSSHAuth (0.00s)
--- FAIL: TestSshAuthConfigOverride (13.29s)

rfay:
Thanks for the incredible work on this. The problem with php5.6 will be solved starting in https://github.com/drud/ddev-images/pull/31 - Your Travis tests caught a genuine bug, so already paid their way. We also would have caught the bug if we ran arm64 container tests, we'll have to think about that as well.

One option is to have me work on those tests for a while, since they've been my friends and enemies for so long. 

Let's disable the tests that still fail after fixing php5.6 using t.Skip() if the arch is arm64, then open a new issue for the remaining thests. That way we'll be able to get basic testing going, but still have a good issue to solve those with.

But first, let's get the php5.6 stuff in there so we can see how much that fixes.

dennisameling:
Just added arm64 container tests to Travis, let's see how that goes: https://travis-ci.com/github/drud/ddev/builds/192948148

As soon as https://github.com/drud/ddev/pull/2584 is merged, we can rebase this PR and see which tests are still failing 👍

rfay:
Hopefully soon! I was blindsided by them going with 2.0 as the default composer version with the release of 2.0. That's going to break a lot of people and a lot of tests. Luckily it broke our tests.

rfay:
Just tested with #2584 on my Raspberry Pi and php 5.6 seems to be working fine now with it.

dennisameling:
The arm64 container job failed (on PHP 5.6 as far as I can see, which would make sense), but it's not passing the exit/error code to Travis, so Travis thinks the job succeeded: https://travis-ci.com/github/drud/ddev/jobs/408005818

Any tips? Could probably do `time make test || travis_terminate 1;`, as described here:
https://medium.com/@manjula.cse/how-to-stop-the-execution-of-travis-pipeline-if-script-exits-with-an-error-f0e5a43206bf - would that help you think? You have more experience with Bash/shell than me 😄 if you also don't know, I'll go ahead and tests some things

.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+  - docker buildx create --name ddev-builder-multi --use  
+
+jobs:
+  include:
+    - stage: test
+      name: Run make test
+      script: make test
+    - name: Linux container tests
+      script: |
+        for dir in containers/*/
```
rfay, line 37:
```suggestion
        set -eu -o pipefail 
        for dir in containers/*/
```

rfay:
(suggestion above)

I see

```
#7 0.071 standard_init_linux.go:219: exec user process caused: exec format error
#7 ERROR: executor failed running [/bin/sh -c chmod -R go-rwx /root/.ssh]: buildkit-runc did not terminate successfully
```

which would normally be the result of using the wrong kind of container?

Then of course 

```
make: *** [../containers_shared.mak:9: container] Error 1
```

so make did the right thing.  It's just that bash was happily willing to continue on because we didn't have a `set -eu` on the front.

dennisameling:
> which would normally be the result of using the wrong kind of container?

Looks like that's because the image that's used in `test-ssh-server`, `gotechnies/alpine-ssh`, is only available for `amd64`: https://hub.docker.com/r/gotechnies/alpine-ssh/tags

I don't know what the purpose of the `test-ssh-server` image is, at least it doesn't block the DDEV workloads on my arm64 device. Do we need this image to be built + tested on `arm64`?

rfay:
test-ssh-server is used in TestCmdAuthSSH  to test `ddev ssh auth`, and I think it would be fine to t.Skip() TestCmdAuthSSH on arm64 for now. 

https://github.com/drud/ddev/blob/8de02b79dfef70f15ed8149fc392d07eb10dca4f/cmd/ddev/cmd/auth-ssh_test.go#L19-L39

rfay:
OK, #2584 is in.

dennisameling:
Nice! Just rebased, let's see how far we can get now 🚀 Will check the results in the morning.

dennisameling:
**make test**
The Travis build timed out after 50 minutes. [This is a hard limit](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts) in Travis. A build doesn't have a timeout limit, but a single job can only run for 50 minutes.

Was considering to split `make test` into two separate jobs for `make testpkg` and `make testcmd`, but looking at the CI log, it seems that `make testpkg` alone isn't even able to finish within 50 mins.

Do you have any suggestions how we can split these tests into separate jobs?

**Linux container tests**
This one fails on MariaDB 5.5, which makes sense as we don't build an image for it. What about we move the `MARIADB_VERSIONS_arm64` etc. variables to a separate file, so that it can be shared between the Makefile for containers and the test file `test_dbserver.sh`?

rfay:
I'd say make separate jobs for testcmd and also for each of the items in pkg/* that has tests. We may still hit the limit with pkg/ddevapp, but it's worth a try. 

TestDdevXdebugEnabled probably doesn't want a t.Skip, but rather "if 5.6 and arm64 t.Log("skipping 5.6 on arm64 because not supported") && continue

> What about we move the MARIADB_VERSIONS_arm64 etc. variables to a separate file, so that it can be shared between the Makefile for containers and the test file test_dbserver.sh?

I'm OK with that.

rfay:
OH another important thing: Set the environment variable GOTEST_SHORT=true. This limits testing on a number of things that are very costly, especially TestDdevFullSiteSetup, which takes a really long time if you test every single CMS. We have only one test on CircleCI (and none in buildkite) that doesn't have this set. 

(But good news: TestDdevFullSiteSetup passed in this run before it timed out, but it took 7 of our 50 minutes to do it)

dennisameling:
Created separate jobs for the different folders now (5 simultaneous jobs), but it's far from efficient: https://travis-ci.com/github/drud/ddev/builds/193511452

Travis runs the init steps on **all** jobs, which adds unnecessary overhead IMO. I'm considering to add an env variable or two to the Makefile, potentially something like TESTPKG_EXCLUDE and TESTPKG_INCLUDE, so that just we need two jobs:
1. "run all tests except ddevapp" and 
2. "run only ddevapp tests". 

First, let's give Travis some time and see if it finishes the `ddevapp` tests within 50 mins.

rfay:
I don't think we need env vars, but rather new (additional) targets in the Makefile. One target for each item in pkg. So the targets can be named for each, and then `testpkg` can depend on all of them and do nothing. And you can have another Make target called `make testddevapp` just tests ddevapp and another called `make otherpkg` which depends on all the other non-ddevapp targets (and then of course `make testcmd`

dennisameling:
Some good news: the only failing tests currently are:

--- FAIL: TestDdevXdebugEnabled (132.89s)
--- FAIL: TestDdevAllDatabases (57.41s)
--- FAIL: TestSSHAuth (60.27s)
--- FAIL: TestCmdAuthSSH (17.96s)

Just pushed https://github.com/drud/ddev/pull/2568/commits/5439059a8cd27bbde68081770ff2d95d0b2422a7, hopefully container tests will pass now 🤞 

Would you mind looking into those failing tests to see how we can fix/skip those on arm64? And the Makefile for the pkg tests with the approach you just mentioned? Let me know where I can be of further help 👍

rfay:
Pretty amazingly awesome!

rfay:
I'll try to look at those this afternoon. May I push into your fork, or would you rather do it another way?

dennisameling:
Yes please go ahead by pushing into the fork! 👍

rfay:
I got GoLand running on my Raspberry Pi, seems to be doing OK. step-debugging, everything. Used this guide, should be the same for your system. https://blog.jetbrains.com/go/2020/02/18/running-goland-on-a-raspberry-pi-4/

dennisameling:
All container tests have passed (🎉) except for `test-ssh-server`: https://travis-ci.com/github/drud/ddev/jobs/411198889 (I excluded XDebug on PHP 5.6 in https://github.com/drud/ddev/pull/2568/commits/09bdce176590ff724a769c1d91a557aa082f55c3).

We'll just need to prevent the `test-ssh-server` image from building on arm64/Travis and then should be good to go with container tests 🚀 my brain is too tired to come up with a solution for that now, so if you don't mind looking into that one as well, that'd be fantastic. I'll try to be around here again before Friday, otherwise during the weekend.

We're almost there! 😄

rfay:
I pushed a rebase, so please be careful when you come back to this. Thanks for the amazing work!

rfay:
We have green Travis! And it got done before any of the others!

dennisameling:
Awesome!! So I guess this PR is ready for review. Let me know if I need to add any optimizations.

We could remove the `multi-arch` step from CircleCI now that the arm64 image builds and tests run natively on Travis. Multi-arch builds are then only relevant when pushing the images to Docker Hub. What do you think?

https://github.com/drud/ddev/blob/8de02b79dfef70f15ed8149fc392d07eb10dca4f/.circleci/config.yml#L279-L280

Oh Yeah

rfay:
I agree that the multi-arch target can come out; I always use `push` anyway. Let's do that in another PR unless we have more work on this one. I'll review more carefully today.

The Makefile thing that confuses me the most is that `make push VERSION=xxx` doesn't leave a tagged amd64 image locally.  I can't see why that is, and perhaps there's a simple workaround. But the dev-side results are bad. Push an image (with tag, for a PR that depends on it), then `ddev start` and it gets the old version of the tagged image. You have to `docker pull <imagespec>` to get the new one.

I imagine this is an artifact (bug?) of `docker buildx --push`, but I guess we could work around it by having a 2-step push that did build and then push. 

https://github.com/drud/ddev/blob/5d7c9b83b479cddb8fd83948e2fecbe9b357879e/containers/ddev-webserver/Makefile#L22-L27

.travis/linux_travis_arm64_setup.sh
```diff
@@ -0,0 +1,41 @@
...
+  set +x
+  echo "${DOCKERHUB_PULL_PASSWORD}" | docker login --username "${DOCKERHUB_PULL_USERNAME}" --password-stdin
+  set -x
+fi
+
+sudo apt-get update -qq
+sudo apt-get install -qq mysql-client zip jq expect nfs-kernel-server build-essential curl git libnss3-tools libcurl4-gnutls-dev docker-compose
+
+# Copy docker-compose to /usr/local/bin because Travis' pre-installed version leads to exec format error
+sudo cp /usr/bin/docker-compose /usr/local/bin/docker-compose
```
rfay, line 19: Is this Ubuntu 20.04? I found on Raspberry Pi (Debian 10) that the only way to get a working docker-compose was to install with `pip3 install docker-compose`; the bundled one has problems, and there is no other distribution path. But I guess if this is Ubuntu 20.04... the apt version works there.

.travis/linux_travis_arm64_setup.sh
```diff
@@ -0,0 +1,41 @@
...
+
+curl -sSL --fail -o /tmp/ngrok.zip https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-arm64.zip && sudo unzip -o -d /usr/local/bin /tmp/ngrok.zip
+
+# Without this .curlrc CircleCI linux image doesn't respect mkcert certs
+echo "capath=/etc/ssl/certs/" >>~/.curlrc
+
+. ~/.bashrc
+
+npm install -g bats
```
rfay, line 28: I didn't know you could do it this way. Easier than most of the ways I've done it. Oh, but I wonder if this gets bats-core/bats-core or the old unmaintained one. (Edit: I checked; it gets bats 1.2.1, which is from bats-core)

containers/containers_shared.mak
```diff
@@ -6,7 +6,7 @@ SANITIZED_DOCKER_REPO = $(subst /,_,$(DOCKER_REPO))
 DOTFILE_IMAGE = $(subst /,_,$(IMAGE))-$(VERSION)
 
 container: container-name
-	docker buildx build --platform linux/amd64 -o type=docker -t $(DOCKER_REPO):$(VERSION) $(DOCKER_ARGS) --label "build-info=$(DOCKER_REPO):$(VERSION) commit=$(shell git describe --tags --always)" .
+	docker buildx build -o type=docker -t $(DOCKER_REPO):$(VERSION) $(DOCKER_ARGS) --label "build-info=$(DOCKER_REPO):$(VERSION) commit=$(shell git describe --tags --always)" .
```
rfay, line 9: This doesn't work on arm64, gets a syntax error. 
```
$ make container
Makefile:21: warning: overriding recipe for target 'push'
../containers_shared.mak:15: warning: ignoring old recipe for target 'push'
container: drud/ddev-webserver:v1.16.0-alpha8-35-g192ca359
docker buildx build -o type=docker --label com.ddev.buildhost=pi1.thefays.us --target=ddev-webserver  -t drud/ddev-webserver:v1.16.0-alpha8-35-g192ca359  .
unknown shorthand flag: 'o' in -o
See 'docker --help'.
```

I also see
```
Makefile:21: warning: overriding commands for target `push'
../containers_shared.mak:15: warning: ignoring old commands for target `push'
```
Which probably says to me that we're not using "push" from containers_shared.mak and it should be removed?

containers/ddev-dbserver/Makefile
```diff
@@ -4,13 +4,9 @@ VERSION := $(shell git describe --tags --always --dirty)
...
-# The format here is majorversion:pinnedimageversion
-# This allows to pin the version (as currently required with 8.0.19
-MARIADB_VERSIONS_amd64=5.5:5.5 10.0:10.0 10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
-MYSQL_VERSIONS_amd64=5.5:5.5 5.6:5.6 5.7:5.7 8.0:8.0.21
-MARIADB_VERSIONS_arm64=10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5
-# MySQL images on Docker Hub currently are amd64 only
-MYSQL_VERSIONS_arm64=
+include database-versions
+
+CURRENT_ARCH= $(shell ../get_arch.sh)
```
rfay, line 9: I'm sure it doesn't matter but I'd be happier with 
```suggestion
CURRENT_ARCH=$(shell ../get_arch.sh)
```

containers/ddev-dbserver/test/test_dbserver.sh
```diff
@@ -13,24 +16,36 @@ function cleanup {
...
+    printf "Test successful for $DB_TYPE $baseversion\n\n"
 done
 
 export DB_TYPE=mysql
-for v in 5.5 5.6 5.7 8.0; do
-    export IMAGE="drud/ddev-dbserver-$DB_TYPE-$v:$tag"
-    export DB_VERSION=$v
+for v in ${!MYSQL_VERSIONS}; do
+    export baseversion=$(echo $v | awk -F ':' '{print $1}')
+	export fullversion=$(echo $v | awk -F ':' '{print $2}')
```
rfay, line 42: Looks like simple indent problem here.

pkg/ddevapp/ddevapp_test.go
```diff
@@ -669,6 +669,12 @@ func TestDdevXdebugEnabled(t *testing.T) {
 	runTime := util.TimeTrack(time.Now(), fmt.Sprintf("%s %s", site.Name, t.Name()))
 
 	phpVersions := nodeps.ValidPHPVersions
+
+	// arm64 builds from deb.sury.org do not have xdebug in php5.6
+	if runtime.GOARCH == "arm64" {
+		t.Log("Skipping php5.6 test on arm64 because deb.sury.org packages do not include it currently")
+		delete(phpVersions, "5.6")
+	}
```
rfay, line 677: Silly me. This shouldn't be done in the tests, but instead in the actual definition of ValidPHPVersions on arm64. Same with Mysql and MariaDB.

pkg/ddevapp/ssh_auth_test.go
```diff
@@ -22,6 +23,9 @@ import (
 
 // TestSSHAuth tests basic ssh authentication
 func TestSSHAuth(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip(("Skipping TestSSHAuth on arm64 because the test-ssh-server is not available for arm64"))
+	}
```
rfay, line 28: I'm sure we can build a test-ssh-server that would work, but I don't think it's worth bothering at this point. However, manual testing shows that `ddev auth ssh` and related capabilities work fine on arm64.

rfay (changes requested):
Mostly nits, but there are a few things here that we need to get done. Thanks for the amazing work on this!

rfay:
test-ssh-server in https://github.com/drud/ddev/pull/2587

rfay:
Assuming you're done for the day @dennisameling , I'll fiddle with a bit of this.

.travis/linux_travis_arm64_setup.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# Basic tools
+
+set -x
+
+if [ ! -z "${DOCKERHUB_PULL_USERNAME:-}" ]; then
```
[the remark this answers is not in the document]
rfay: I added these into the travis config.

.travis/linux_travis_arm64_setup.sh
rfay: Ah, I do see it's focal, thanks.

containers/ddev-dbserver/Makefile
rfay: Fixed this.

containers/ddev-dbserver/test/test_dbserver.sh
rfay: Fixed this.

pkg/ddevapp/ssh_auth_test.go
rfay: Fixed this, added new test-ssh-server and removed this code.

pkg/ddevapp/ddevapp_test.go
rfay: Fixed this with conditional compilation. And it even fixes the ddev binary's available choices of php, mysql, and mariadb versions.

containers/containers_shared.mak
rfay: Moved this to https://github.com/drud/ddev/issues/2588, so marking this resolved here.

.travis.yml
```diff
@@ -0,0 +1,59 @@
...
+  - echo '{"experimental":true}' | sudo tee /etc/docker/daemon.json
+  - export DOCKER_CLI_EXPERIMENTAL=enabled
+  # Upgrade to Docker CE 19.03 for BuildKit support
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=arm64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+  # Show info to simplify debugging and create a builder
+  - docker info
+  - docker buildx create --name ddev-builder-multi --use
```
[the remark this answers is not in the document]
rfay: > Why aren't these bundled into the linux_travis_arm64_setup.sh?

This is probably the last question to be answered, and thankfully it's a small one.

rfay (approved):

.travis.yml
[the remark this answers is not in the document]
dennisameling: Fixed in https://github.com/drud/ddev/pull/2568/commits/ca995d51142c02e0df423a91d58e8dbc78306856 😄

dennisameling:
If you could restart de failed Buildkite jobs, then CI should be green and this PR ready to merge 🚀

rfay:
I probably shouldn't have completely excluded php5.6 instead of just skipping it in the xdebug test, but oh well, can be changed later. It's not like there will be a lot of people screaming for php5.6 on arm64 in the near future. Also opened https://github.com/oerdnj/deb.sury.org/issues/1475 about that.

rfay:
OK CONGRATULATIONS @dennisameling ! 🚀 🚀 🚀 !

rfay:
Ooops, travis tests aren't running on master build, perhaps because I turned off "build pushed branches". I turned that back on just now.

drud_ddev_-_Travis_CI

rfay:
Unfortunately, not running on my PR either, https://github.com/drud/ddev/pull/2589

dennisameling:
I'm on mobile now so can't check, but does it work if you create a brand-new PR now?

rfay:
CircleCI has quite exotic treatments of forked PRs, so I imagine it's an issue for travis as well. Also, it looks like the docker login isn't working (probably also because forked PR), so we'll have to work on that.

rfay:
Oh, I see travis running on #2589 now. So I guess the forked PR problem is at least partially solved.

dennisameling:
From the Travis docs:
> Travis only runs builds on the commits you push after you’ve added a .travis.yml file.

If I understand this correctly,  it means that on the next commit to master,  it should start running there as well. 

https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci-using-github

rfay:
BTW, the travis tests are completing far faster than anything else we have. I think I'll split up the CircleCI tests the way we did, and split the full-size TestFullSiteSetup into its own `make testfullsitesetup`. Sadly, there's nothing I can think of to do about the 2+ hour Windows runs. And I haven't even started WSL2 testing, although that's mostly covered by the Linux testing.

dennisameling:
So what about we disable the `buildkite/ddev-windows-dockerforwindows-nfs` job? Assuming that NFS on Windows isn't explicitly supported anymore now that WSL2 is here.

Regarding Buildkite, would you be up for moving those tests to GitHub Actions? That'd allow me to troubleshoot things much more easily as I don't have access to Buildkite.

rfay:
NFS is still widely used, and regular Windows still widely used. 

I absolutely would like to move to github actions; but might have to be self-hosted based on our experience as we go. When I've tried macOS and Windows on services like Appveyor and CircleCI there have been lots of gotchas with trying to test ddev (because it itself needs Docker, and it's hard to manage Docker Desktop inside a test environment).

I'm sorry you haven't had access to buildkite @dennisameling - if you'll PM me your email address I'll send you the invite!

PR #2582, opened 2020-10-25
[wip] Manage WSL2 hosts on Windows side using ddev.exe, fixes #2533

### LINKED ISSUE #2533: Non-ddev.site hostnames can't work on WSL2, and neither can ddev.site if DNS rebinding disallowed
**Describe the bug**

When the hostname required for a project is not *.ddev.site, or when DNS rebinding is not allowed, WSL2 browser stuff can't work.

So:
1. If DNS rebinding is disallowed, ddev will add the hostname to /etc/hosts. But unfortunately it's the /etc/hosts in the WSL2 system, which is *not* the hosts file that the browser will use (which is C:\Windows\System32\drivers\etc\hosts). So the browser can't recognize it.
2. If project_tld is set to something other than ddev.site, same problem
3. If additional_fqdns is set, that has the same problem.

I'm not sure there's a good solution to this, because WSL2 code can't change things on the Windows side...

Windows users can work around this issue by manually adding the needed hostnames to C:\Windows\System32\drivers\etc\hosts or by using the *windows* ddev to issue the command `ddev hostname <neededfqdn> 127.0.0.1`

rfay:
## The Problem/Issue/Bug:

On Windows WSL2, /etc/hosts manipulation needs to be done on the Windows WSL2 side, or the browser won't know about it. 

Run `ddev.exe hostname` (if it exists) to update the *Windows*-side hosts file.

### DIFF
--- a/cmd/ddev/cmd/hostname.go
+++ b/cmd/ddev/cmd/hostname.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/drud/ddev/pkg/exec"
 	"github.com/drud/ddev/pkg/nodeps"
 	"github.com/drud/ddev/pkg/util"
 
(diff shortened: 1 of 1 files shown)


rfay:
Closed in favor of #2655

PR #2626, opened 2020-11-09
Pin mariadb versions to previous release because of regression that affects php7.2

rfay:
## The Problem/Issue/Bug:

The recently released versions of MariaDB have a regression that affects all PHP versions < 7.3. https://jira.mariadb.org/browse/MDEV-24121

Pin MariaDB versions to the previous minor release. 

Try installing TYPO3 v10 using php7.2. I think it will succeed, whereas it would have failed with php7.2 and ddev v1.16.0-rc2

Some kind of manual test that's more authoritative would be better

I'd love to have a test that clearly demonstrates this, so we could know when to unpin these releases.

### DIFF
--- a/containers/ddev-dbserver/database-versions
+++ b/containers/ddev-dbserver/database-versions
@@ -1,7 +1,7 @@
 # The format here is majorversion:pinnedimageversion
 # This allows to pin the version (as currently required with 8.0.21)
-MARIADB_VERSIONS_amd64="5.5:5.5 10.0:10.0 10.1:10.1 10.2:10.2 10.3:10.3 10.4:10.4 10.5:10.5"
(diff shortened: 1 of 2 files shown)


rfay:
I tested this with TYPO3 v10 install; it worked fine.

I then reverted to ddev v1.16.0-rc2 and the TYPO3 install hangs at step 4 again.

rfay:
It looks like https://jira.mariadb.org/browse/MDEV-24121?focusedCommentId=171135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-171135 is a repro case for the regression.

rfay:
They've just announced new versions of MariaDB, https://mariadb.org/mariadb-10-5-8-10-4-17-10-3-27-and-10-2-36-now-available/

PR #2635, opened 2020-11-12
Add default usage of custom commands again, fixes #2633

### LINKED ISSUE #2633: [Minor] Ignore or complain about custom commands which don't have required annotations
**Describe the bug**

Apparently in v1.16+, custom/script commands that don't have a "Usage" annotation don't get properly shown in the list of commands (they're without a command name). 

**To Reproduce**

Use this command as .ddev/comamnds/web/badcmd:
```
#!/bin/bash
## Description: This command won't show up properly

tr -cd '[:alnum:]' < /dev/urandom | fold -w32 | head -n1
```

If you type `ddev` you'll see it doesn't get labelled correctly:

```
Create and maintain a local web development environment.
Docs: https://ddev.readthedocs.io
Support: https://ddev.readthedocs.io/en/stable/#support

Usage:
  ddev [command]

Available Commands:
                   This command won't show up properly (shell web container command)
  auth             A collection of authentication commands
```

**Expected behavior**

In v1.15.3 a command like this would work, with the name of the script as the name.

IMO it's fine for commands to require the Usage annotation, but if they don't have it ddev should complain I think.

Note that this command works fine in .ddev/commands/web/ttpwd:

```
#!/bin/bash
## Description: Generate 32 character alphanumeric password
## Usage: ttpwd

tr -cd '[:alnum:]' < /dev/urandom | fold -w32 | head -n1
```

gilbertsoft:
## The Problem/Issue/Bug:
In case of missing usage annotation no command name is shown anymore. This was accidentally removed with https://github.com/drud/ddev/commit/c34907d7d6cd836e0571ad0b0effcaa1d48a7b83.

This patch reintroduces the default value again.

Create a custom command without the `usage` annotation and check the command name is shown properly on the help screen.

## Related Issue Link(s):
#2633
#2452

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -106,6 +106,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 					description = val
 				}
 
+				usage = commandName + " [flags] [args]"
 				if val, ok := directives["Usage"]; ok {
 					usage = val
 				}


rfay:
Rebased again. Hope to review and get these in this week.

rfay (approved):
Looks perfect to me, thanks. Tested with v1.16.0, doesn't show it. Tested with this, shows it and does the right thing.

PR #2646, opened 2020-11-18
Fix description of HTTPS_EXPOSE

liayn:
## The Problem/Issue/Bug:
Wrong description in the docs

Corrects the naming of the port setup and how the exposed port is used

### DIFF
--- a/docs/users/extend/custom-compose-files.md
+++ b/docs/users/extend/custom-compose-files.md
@@ -53,7 +53,7 @@ When defining additional services for your project, we recommended you follow th
 
         * `VIRTUAL_HOST=$DDEV_HOSTNAME`
         * `HTTP_EXPOSE=portNum` The `hostPort:containerPort` convention may be used here to expose a container's port to a different external port. To expose multiple ports for a single container, define the ports as comma-separated values.
-        * `HTTPS_EXPOSE=<exposedPortNumber>:portNum` This will expose an https interface on `<exposedPortNumber>` to the host (and to the web container) as `https://<project>.ddev.site:exposedPortNumber`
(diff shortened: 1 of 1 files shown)


docs/users/extend/custom-compose-files.md
```diff
@@ -53,7 +53,7 @@ When defining additional services for your project, we recommended you follow th
 
         * `VIRTUAL_HOST=$DDEV_HOSTNAME`
         * `HTTP_EXPOSE=portNum` The `hostPort:containerPort` convention may be used here to expose a container's port to a different external port. To expose multiple ports for a single container, define the ports as comma-separated values.
-        * `HTTPS_EXPOSE=portNum:<exposedPortNumber>` This will expose an http interface on `portNum` to the host (and to the web container) as `https://<project>.ddev.site:exposedPortNumber`
+        * `HTTPS_EXPOSE=<exposedPortNumber>:portNum` This will expose an https interface on `<exposedPortNumber>` to the host (and to the web container) as `https://<project>.ddev.site:exposedPortNumber`
```
rfay, line 56: Let's make it the same terminology as HTTP_EXPOSE; it can also be improved with an example. 
```suggestion
        * `HTTPS_EXPOSE=<exposedPortNumber>:portNum` This will expose an https interface on `<exposedPortNumber>` to the host (and to the web container) as `https://<project>.ddev.site:exposedPortNumber`. To expose multiple ports for a single container, use comma-separated definitions, as in `HTTPS_EXPOSE=9998:80,9999:81`, which would expose http port 80 from the container as `https://<project>.ddev.site:9998` and http port 81 from the container as `https://<project>.ddev.site:9999`.
```

rfay (changes requested):
Thanks for this, looks good. So easy to get confused. I added a suggestion that adds an example, which I think is important, consier clicking the button to commit that.

liayn:
I think I do not need a contributor agreement for this low profile PR, right?

rfay:
It would be nice if you'd sign the CLA. However, I won't insist.

liayn:
Sry, no time to read that atm. Fixing ddev setup stuff which got broken with v1.16. this is just a drive-by change.
No problem if you take the owner/authorship so you get the legal part working.

PR #2652, opened 2020-11-20
Allow configuration to fail when a hook execution fails, fixes #2536

### LINKED ISSUE #2536: Add config.yaml option to fail an action when hooks fail
image
```yml
hooks:
  post-start:
    - exec: exit 1
    # Private files directory.
    - exec: mkdir /var/www/private

    # Install profile
    - exec: drush site-install server -y --existing-config

    # Compile theme
    - exec-host: ddev robo theme:compile

    # Generate one-time login link
    - exec: drush uli
```

Let's say we have a failing post-start command in our custom configuration.
Right now there's no trivial way to know if all the commands succeeded or not, as `ddev start` exits with code 0 even if something in the hooks failed.

We could stop on the first failure for `post-start` and propagate the received exit code for the caller of `ddev start`.

As it might break existing configs, it could be a config option, like `stop_on_hook_failure: true` , with a default value false, it would ensure the backwards compatibility.

majamee:
## The Problem/Issue/Bug:
https://github.com/drud/ddev/issues/2536

Introduction of a new option (global & local): fail-on-hook-fail

Easiest way to test is probably by using:
`ddev config global --fail-on-hook-fail=true`
After that, just include a pre-start hook with an exit code (!=0) for instance use the following script saved as test.sh in the config.yaml of your project like this for example:
```
  pre-start:
  - exec-host: bash ./test.sh
```
The contents of the example `test.sh` can be found here (place in root folder of your project)
```
#!/bin/bash
exit 1
```
After you have done all that make sure, you are using the self-build version of `ddev` either from my branch or from the PR, run `ddev poweroff` to ensure, that there are no side-effects from other versions and start your project as usual by calling `ddev start`

`make staticrequired` finished without errors and without warnings.
Manual testing was successful.
Tested:
- inclusion of new option for all projects via `ddev config global --fail-on-hook-fail=true`
- inclusion of new option for specific project via adding `fail_on_hook_fail: true` to projects `.ddev/config.yaml`

## Release/Deployment notes:
Nothing else should be affected, even though this was my first project using "Go" 🙂


rfay:
Look at you go!

rfay:
Please improve this PR description with full manual testing instructions, including example hooks, and please confirm that you've tested it manually.

majamee:
> Please improve this PR description with full manual testing instructions, including example hooks, and please confirm that you've tested it manually.
> 
> I'm sure this will need a test as well. We can cross that bridge a little farther down the road.

Done 👍

rfay (approved):
Nice work on this! 

I did these things you may want to review:

* Added test coverage for this case
* Prohibit exec and composer tasks in pre-start. This was a fail I discovered in testing this and had nothing to do with the PR.
* Fixed a panic where non-string values were used as the value of an exec, like `exec: false`. This again was a panic I discovered in testing and had nothing to do with this PR.

Would love it if you could give a look and a test and make sure you're good with it.

Thanks for the nice work on this!

majamee:
@rfay I tested on https://github.com/drud/ddev/commit/6c4c1246ca3550353676697464f22a57fe01b8da and everything is looking fine plus working as expected. Thank you :)

PR #2655, opened 2020-11-21
Manage WSL2 hostnames on Windows side where necessary

gilbertsoft:


rfay:
Rebased and fixed a couple of linting errors.

rfay:
This doesn't currently build for windows at all, 
```
$ make windows_amd64
make: Circular .gotmp/bin/windows_amd64/ddev.exe <- windows_amd64 dependency dropped.
building .gotmp/bin/windows_amd64/ddev.exe from ./cmd/... ./pkg/...
# github.com/drud/ddev/pkg/exec
pkg/exec/exec_windows.go:33:28: not enough arguments in call to "golang.org/x/sys/windows".CreateFileMapping
	have ()
	want ("golang.org/x/sys/windows".Handle, *"golang.org/x/sys/windows".SecurityAttributes, uint32, uint32, uint32, *uint16)
pkg/exec/exec_windows.go:34:24: not enough arguments in call to "golang.org/x/sys/windows".MapViewOfFile
	have ()
	want ("golang.org/x/sys/windows".Handle, uint32, uint32, uint32, uintptr)
pkg/exec/exec_windows.go:41:31: undefined: event
make: *** [.gotmp/bin/windows_amd64/ddev.exe] Error 2
```

rfay:
But I guess we only care about how it builds for linux? It does build on linux, haven't tried the functionality yet.

rfay:
Manually testing this on WSL2, it seems not to change behavior at all; we still have the exact same problem. I haven't stepped through the code yet.

rfay:
Rebased.

rfay:
Given that MS is supposed to provide native X11 support "real soon now", it's going to be simpler for WSL2 users to use the browser inside wsl2. 

IMO we should keep things fairly simple instead of trying to solve this problem with everything we have. I'm afraid that the launching into Windows would add complexity that we'd never unravel.

So if it's OK, I'd like to abandon this approach right now.  Closing for now, but we can reopen if we see fit.

PR #2659, opened 2020-11-24
Bump NSIS to 3.06.1 for Linux builds

gilbertsoft:
This patch bumps NSIS to the last stable release 3.06.1 with Linux builds. This ensures we get the same result on MacOS and Linux but does not influence the production release cycle.

### DIFF
--- a/.circleci/linux_circle_vm_setup.sh
+++ b/.circleci/linux_circle_vm_setup.sh
@@ -45,7 +45,7 @@ pyenv global 3.8.3 # added to make CircleCi give us pip3
 pip3 install -q yq mkdocs==0.17.5
 
 # Get the Stubs and Plugins for makensis; the linux makensis build doesn't do this.
-wget https://sourceforge.net/projects/nsis/files/NSIS%203/3.04/nsis-3.04.zip/download && sudo unzip -o -d /usr/local/share download && sudo mv /usr/local/share/nsis-3.04 /usr/local/share/nsis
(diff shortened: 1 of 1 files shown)


rfay:
I'll certainly manually test this before pulling, but are there any specific testing techniques?

gilbertsoft:
We only have to do a quick test if the path manipulation still works correctly, I see the following cases:

- fresh install, DDEV install folder is not set in $PATH
- update, DDEV install folder extists once in $PATH
- update with multiple DDEV install folders in $PATH (fix for the wrong installer which added the path again on updates)
- removal of the folder on uninstall

gilbertsoft:
I've checked the steps above and it works fine like expected. Releasing now a new EnVar plugin...

gilbertsoft:
Release is done and will be used for all new installer builds here. As said this patch will only update the assets on Linux to latest NSIS.

rfay:
Thanks so much. Is there any urgency to this? Does it need to go into v1.16 point relesaes? (I expect at least one more point release)

gilbertsoft:
No, not urgent. This will only ensure the Windows installer built on Linux uses the same assets like the one on MacOS does. But anyway it wont hurt anything as this part is not used for the production builds. It's just a cosmetic change.

Because we always use the latest stable release of the EnVar plugin the new release will of course be included with the next release.

rfay (changes requested):
It turns out that makensis is also installed (with 3.06.1) via homebrew. We should either remove that or use this one, I'm not sure which is appropriate. I see that line 48 puts it in /usr/local/share, but the homebrew install puts it in ~linuxbrew.

PR #2661, opened 2020-11-24
Docs: Add a little extra info on snapshot deletion [skip ci][ci skip]

chellman:
## The Problem/Issue/Bug:

I wasn't 100% clear on whether I could/should delete snapshots on my own, or use a ddev command.

Updates the docs to explain that yes, you can just delete them however you want.


rfay (approved):
Thanks for this! I made a minor edit and it's good to go.

PR #2663, opened 2020-11-25
Solve problem with unreachable sites due to ddev-router config fail, fixes #2648

### LINKED ISSUE #2648: Can't access more than 2 ddev sites at the same time (after updating to v1.16.0, MacOS)
**Describe the bug**
After updating to the latest version 1.16.0 on macOS I find that I can't browse to any site for any ddev project I start after the first started project.

**To Reproduce**

ddev start site1
open browser with this site: https://site1.ddev.site.  (all good)

then start new a project:
ddev start site2

browse to site2: https://site2.ddev.site. 
I get this error:

503: No ddev back-end site available.
This is the ddev-router container: There is no back-end webserver at the URL you specified. You may want to use "ddev start" to start the site.

**Expected behavior**
I expect the new site to open in the browser

**Screenshots**
image

- Host computer OS and Version: macOS Catalina 10.15.7
- Docker Desktop version: 2.5.0.1 (49550)

DDEV-Local version	v1.16.0
commit            	COMMIT should be overridden
db                	drud/ddev-dbserver-mariadb-10.2:v1.16.0
ddev-ssh-agent    	drud/ddev-ssh-agent:v1.16.0
docker            	19.03.13
docker-compose    	1.27.4
os                	darwin
router            	drud/ddev-router:v1.16.0
web               	drud/ddev-webserver:v1.16.0

rfay:
## The Problem/Issue/Bug:

#2648 : In ddev v1.16.0 and v1.16.1, a `ddev start` might result in a site that was unreachable; the newly started site would get "503: No ddev back-end site available."

503__No_ddev_back-end_site_available

It turns out that this was timing related, but more importantly, 

* the listener (`-watch`) docker-gen process creates a script which is then run to create the nginx configuration
* However, the script is not recreated if docker-gen does not see that it changed
* However, the generated script had hostnames in it (container names) but it did *not* include the ports from HTTP_EXPOSE or HTTPS_EXPOSE. So if those came through in separate events, docker-gen would read the later events but decide that because the generated script did not change, it did not need to be recreated (and the script didn't get run... so the nginx configuration was not updated)

This also
* Improves the healthcheck to detect whether the /etc/nginx/conf.d/default.conf has been properly generated
* Cleans up the healthcheck in general
* Provides a default.conf that is more appropriate than the one put there by the nginx package (and includes healthcheck)

- [x] If you do *not* have the ddev binary build with this PR, then add this file as ~/.ddev/router-compose.nginx.yaml (router-compose.nginx.yaml in your global .ddev directory, the one in your home directory), then `ddev poweroff` and do the normal testing that was earlier able to create the problem: 

```yaml
version: '3.6'
services:
  ddev-router:
    image: drud/ddev-router:20201124_router_fails_to_expose
```
- [x] Otherwise, test with the ddev binary from this PR
- [x] Make sure to test with the varnish recipe from ddev-contrib and other 3rd party services. 
- [x] The script linked in #2648 may induce the failure described there, so try it.

This seems to be timing related and dependent on docker's state somehow and difficult to test.

## Related Issue Link(s):

#2648

### DIFF
--- a/containers/ddev-dbserver/README.md
+++ b/containers/ddev-dbserver/README.md
@@ -8,4 +8,3 @@ make push
 make push VERSION=someversion
 make clean
 ```
-
--- a/containers/ddev-router/etc/nginx/conf.d/default.conf
+++ b/containers/ddev-router/etc/nginx/conf.d/default.conf
@@ -1,0 +1,86 @@
+# DEFAULT default.conf SOMETHING IS WRONG
+# This is the "default"" default.conf and if you see this it means
+# that a proper default.conf has not been generated and something
+# is wrong
+
+# If we receive X-Forwarded-Proto, pass it through; otherwise, pass along the
+# scheme used to connect to this server
+map $http_x_forwarded_proto $proxy_x_forwarded_proto {
+  default $http_x_forwarded_proto;
+  ''      $scheme;
+}
+# If we receive X-Forwarded-Port, pass it through; otherwise, pass along the
+# server port the client connected to
+map $http_x_forwarded_port $proxy_x_forwarded_port {
+  default $http_x_forwarded_port;
+  ''      $server_port;
+}
+# If we receive Upgrade, set Connection to "upgrade"; otherwise, delete any
+# Connection header that may have been passed to this server
+map $http_upgrade $proxy_connection {
+  default upgrade;
+  '' close;
+}
+# Default dhparam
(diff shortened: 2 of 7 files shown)


rfay:
I note that when I use the varnish 3rd party service with this PR the healthcheck stanza does not get generated in the ddev-router nginx config (default.conf).

rfay:
I tried the script from #2648 and ran it repeatedly with this ddev version and was no longer able to demonstrate the problem.

PR #2667, opened 2020-11-30
Use root user for PHPMyAdmin

rfay:
## The Problem/Issue/Bug:

It makes sense for PHPMyAdmin (dba) to use the root user instead of the 'db' user.


gilbertsoft:
We just need to change this two lines to achieve this https://github.com/drud/ddev/blob/a7172b79e7e4eb5a990142321053cd28ddb080a2/pkg/ddevapp/templates.go#L174-L175

gilbertsoft (approved):

PR #2669, opened 2020-11-30
Using fully-qualified name `brew install drud/ddev/ddev`

shaal:
When running the existing command `brew tap drud/ddev && brew install ddev`, I got this error:

```
Error: Formulae found in multiple taps:
       * drud/ddev/ddev
       * drud/ddev-edge/ddev

Please use the fully-qualified name (e.g. drud/ddev/ddev) to refer to the formula.
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -33,10 +33,10 @@ Docker and docker-compose are required before anything will work with DDEV. This
 For macOS and Linux users, we recommend installing and upgrading via [Homebrew](https://brew.sh/) (macOS) or [Homebrew on Linux](https://docs.brew.sh/Homebrew-on-Linux) (Linux):
 
 ```
-brew tap drud/ddev && brew install ddev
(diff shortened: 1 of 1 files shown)


rfay:
This is a result of you having both edge and stable taps installed. I'm not sure we need to confuse people by changing it, but maybe...

shaal:
oh, that could be the case, I thought something changed in Brew, that now requires using FQN.
I guess for new users this will not be a problem.

rfay:
It still might not be a bad idea.

rfay:
I appreciate this PR, but have never quite figured out how to do this "just right" for people. For you, I recommend that you `brew untap drud/ddev` and only use the edge versions. It's hard to keep this so that people can use the familiar usage (most people never use edge, and the vast majority of people will never have two taps). 

Way opt to completely updated approach to this, but for now I think we might be adding complexity instead of solving a common problem. Closing for now, converstaion open.

PR #2674, opened 2020-12-02
Update database_types.md for mysql 5.7

programmieraffe:
Just a propopal to give DDEV users a clear hint for mysql 5.7 support (which still runs on many  webspace hosting providers, therefore having the same version on DDEV is very helpful). 

Best regards.& thanks for providing DDEV!

### DIFF
--- a/docs/users/extend/database_types.md
+++ b/docs/users/extend/database_types.md
@@ -1,20 +1,26 @@
 ## Database Server Types
 
-ddev supports most versions of MariaDB and MySQL database servers. You can configure the type that you want using the `.ddev/config.yaml` file, for example:
+DDEV-Local supports most versions of MariaDB and MySQL database servers, but of course the two types are mutually exclusive.
 
-`mariadb_version: 10.4` or `mysql_version: 8.0`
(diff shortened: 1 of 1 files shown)


rfay (commented):
Thanks for taking the time to update the docs! 

I don't understand why this sentence is needed when at the beginning of the section it says 

"ddev supports most versions of MariaDB and MySQL database servers. You can configure the type that you want using the .ddev/config.yaml file, for example:

mariadb_version: 10.4 or mysql_version: 8.0
"

ddev also supports mysql 5.5 and 5.6 and 8.0. Why do you single out 5.7 here?

programmieraffe:
Imho a specific example is very helpful for new users who might overlook sentences like "we support all version".

I just picked the latest 5.x version which is used by webspace hosters in germany to give a specific example on how to set it. Examples for 5.5,5.6 could be added as well :-)

Best regards

rfay:
Thanks @programmieraffe - I rewrote most of this page and added a new commit (and rebased). Could you see if the rewrite addresses your concerns?   (Updated [database_types.md](https://github.com/drud/ddev/blob/05e23bdd363f5a15728f588226765119c3a38893/docs/users/extend/database_types.md))

programmieraffe:
Yes, very nice! 👍 👍

PR #2675, opened 2020-12-02
Docs: Escape asterisks in glob patterns

mootari:
## The Problem/Issue/Bug:
The documentation uses glob patterns with asterisks in several places. In Markdown, asterisks have to be escaped if they need to be rendered verbatim.

Escapes asterisks in glob patterns in ~~the document "Extending and Customizing Environments"~~ all Markdown documents below `/docs`.

Note: There are likely more occurences in other files.


rfay (approved):
Perfect, thanks so much!

mootari:
Some asterisks already rendered properly without being escaped (e.g. in a headline), likely because they had no terminating asterisk in the same block. I'd recommend to still be consistent and escape all occurrences.

rfay:
Could you please push another commit to fix all of them? Thanks!

mootari:
> Could you please push another commit to fix all of them?

Done! I only checked for `.*` and `*.` (separately), but imo this should cover all instances.

At some point someone will have to do another pass over the docs to wrap plain text file names/patterns and variables as inline code (after which the escaping won't be necessary any longer).

rfay:
And of course I'd love it if you'd sign up for that :)

PR #2684, opened 2020-12-09
Ignore drushrc.php writing when settings disabled

joelpittet:
## The Problem/Issue/Bug:
`public/sites/default/drushrc.php` being written when settings disabled.

Skips writing the file when settings are disabled.

On a D7 site
```
 ddev config \
  --docroot=public \
  --disable-settings-management
```

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -456,6 +456,11 @@ func writeDrupal6DdevSettingsFile(settings *DrupalSettings, filePath string) err
 // WriteDrushrc writes out drushrc.php based on passed-in values.
 // This works on Drupal 6 and Drupal 7 or with drush8 and older
 func WriteDrushrc(app *DdevApp, filePath string) error {
+	// Ignore because this is a settings file.
+	if app.DisableSettingsManagement {
+		return nil
+	}
+
 	if fileutil.FileExists(filePath) {
 		// Check if the file is managed by ddev.
 		signatureFound, err := fileutil.FgrepStringInFile(filePath, DdevFileSignature)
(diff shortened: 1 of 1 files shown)


rfay:
@joelpittet I remember you mentioning this in Slack. Question though - Was it a Drupal 7 site? Because this isn't to be written at all in Drupal 8+ I don't think.

pkg/ddevapp/drupal.go
```diff
@@ -456,6 +456,11 @@ func writeDrupal6DdevSettingsFile(settings *DrupalSettings, filePath string) err
 // WriteDrushrc writes out drushrc.php based on passed-in values.
 // This works on Drupal 6 and Drupal 7 or with drush8 and older
 func WriteDrushrc(app *DdevApp, filePath string) error {
+	// Ignore because this is a settings file.
+	if app.DisableSettingsManagement {
+		return nil
+	}
```
rfay, line 462: This isn't incorrect IMO, but I think it would be better to fix this in drupal7PostStartAction and also in drupal6PostStartAction. It seems to me like those were both somehow neglected on this. Use drupal8PostStartAction as example.

Congrats on your first golang PR! Please make sure to test it manually.

rfay (changes requested):

joelpittet:
Yes this is for Drupal 7, sorry this got lost in my emails

pkg/ddevapp/drupal.go
joelpittet: Thanks I'll take a look at that now!

joelpittet:
@rfay I posted a new patch, I simplified the conditional to one level so the nesting is easier to read and just exit early, I hope you don't mind, I can revert this change if you think it's overstepping scope?

rfay:
You didn't post a new PR right? Just added a commit to this one? Thank you ! I'll be catching up on all of these before too long.

rfay (approved):
This looks perfect to me, thanks! I reviewed the code and manually tested d6/7/8/9 and then rebased to upstream/master

joelpittet:
Yeah tried not to create queue noise and just reused this fork/PR hope that is a good approach?

Thanks for merging!

PR #2700, opened 2020-12-17
Make command work with TablePlus installed different places

Graloth:
## The Problem/Issue/Bug:
Some have TablePlus installed differently, in our case some used Setapp (https://setapp.com/), which installs the app in a different location.
When researching a more generic solution, all my testing showed that simply specifying the app name "TablePlus" should be enough, and it should launch regardless of install location.

Additionally, we have also included additional information in the query, so that TablePlus uses the project name and gets a grey color (maybe it should be a different color to differentiate it as a local db).

- Path to application removed in favor of App name.
- The OS knows where the app is located, no need to keep track of it ourselves.

## Release/Deployment notes:

----

I would very much like if other people on MacOS helps test that this doesn't break the command for them, on all the tests I've done with my colleagues it still works.


rfay (changes requested):
* Please add extra locations, delimited by commas to HostBinaryExists
* Please run `make packr2` after your changes and check in the results.

Thanks!

Graloth:
> * Please add extra locations, delimited by commas to HostBinaryExists
> * Please run `make packr2` after your changes and check in the results.
> 
> Thanks!

Thanks :) Just did both, the `make packr2` didn't seem to return any output other than the docker run command it uses, but it did make some changes to go.mod (included in the pull request update).

I hope this makes the checks succeed :)

cmd/ddev/cmd/global_dotddev_assets/commands/host/tableplus
```diff
@@ -7,10 +7,9 @@
 ## Usage: tableplus
 ## Example: "ddev tableplus"
 ## OSTypes: darwin
-## HostBinaryExists: /Applications/TablePlus.app
+## HostBinaryExists: /Applications/TablePlus.app,/Applications/Setapp/TablePlus.app
```
rfay, line 10: I don't understand the usage of `/Applications/Setapp/TablePlus.app` - Is /Applications/Setapp a directory that is commonly used for installs? I've never seen it. 

If this is just something for your personal use, then just remove the #ddev-generated line at the beginning of the script and manage it yourself.

rfay (changes requested):

cmd/ddev/cmd/global_dotddev_assets/commands/host/tableplus
rfay: Oh, also, I think you must have done the `make packr2` on Windows? I haven't figured out how or why, but it does fail to do that correctly on Windows. We just need to make sure that it gets done on macos or linux before merge.

Graloth: If you have the Setapp installed, it installs applications to it's own directory instead of the default applications directory, as it's a subscription service where you get a bunch of paid software rather cheaply it has become popular with mac users in our office.

I've made all pull requests and modifications on my mac, as I'm with family and don't have access to my windows desktop,  should it have returned any specific output? I've just run it again and the output is still just the normal docker run command output.

rfay: Your Windows desktop won't be able to demonstrate anything anyway, unfortunately, only the mac can do it. Windows TablePlus doesn't have the ability to be launched with a dburl, https://github.com/TablePlus/TablePlus/issues/2224

To test this, rebase it and run `make packr2` and commit the results if anything changes, then `make` and run the resultant binary. 

I'm reluctant to make a change though that seems very special to your environment, when you can just edit the file in your own ~/.ddev/commands/host and be done with it.

Graloth: I think that you can just close this commit, I agree that since it can be done locally and is a more niche environment, it shouldn't have to be supported out of the box, and we already run with the customized version just fine.

rfay:
Thanks so much for thoughtfully providing this in the first place. Maybe someday we can figure out how to find the app on macOS, and that would help things. I'm not even sure that things install into /Applications on non-English macOS.

Graloth:
The `open` command should accept just the application name instead of the full path if the application is properly installed, I know it works personally for a normally installed TablePlus, but there might be other ways to install it so it doesn't get properly registered with the OS and thus not able to be started without the full path.

rfay:
I guess you're right. The thing then is how to *detect* whether it exists, to decide whether or not to show the command.

rfay:
I guess we should have had a command like `dbbrowser` with subcommands for the various things. Then they could show under that. I just don't like to pollute the command namespace with every little thing people could use.

Graloth:
I completely agree, it would also mean you could end up with supporting 3rd party apps forever, leaving less and less time for work on what is actually important for ddev. I think it is fine to support the default and most common install location, but you shouldn't need to be responsible for every little need there might be, especially since the system you have in place already allow overwriting and modifying existing commands to fit ones needs.

I am sadly not aware of any decent way to find the path to an application, all I know use the same registry as the `open` command, so would be subject to the same issues.

rfay:
TablePlus is quite a nice cross-platform tool too. I was really sad there was no way to launch it properly on Windows. Thought I was going to be able to just add some install locations and have it work on Windows, but no dice.

rfay:
But the heidisql PR just went in, and it's nice for both Windows and WSL2.

peterkraume:
I just stumbled over this pull request and I'm not sure why it was closed. On my MacBook Pro, TablePlus is also installed in `/Applications/Setapp/TablePlus.app`. So ddev currently can't detect that the app is present. I would love to have this great automatic TablePlus support for Mac and SetApp.

rfay:
You know you can just change the shell command and remove the #ddev-generated from the top of it...

peterkraume:
Thx a lot for the hit. I just modified `~/.ddev/commands/host/tableplus` and now it works like a charm!

Schrank:
> Thanks so much for thoughtfully providing this in the first place. Maybe someday we can figure out how to find the app on macOS, and that would help things. I'm not even sure that things install into /Applications on non-English macOS.

German mac here. It does. Mac OS does weird things by translating directory names in Finder 🙄  but thankfully not in the terminal.

---

And in case someone needs the content of ``, here it is, works for me:
```
#!/bin/bash

# Support for TablePlus, https://tableplus.com/
# This command is available if macOS and TablePlus is installed in the normal place
## Description: Run tableplus with current project database
## Usage: tableplus
## Example: "ddev tableplus"
## OSTypes: darwin
## HostBinaryExists: /Applications/TablePlus.app,/Applications/Setapp/TablePlus.app

set -x
query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/db"

open "$query" -a "/Applications/Setapp/TablePlus.app/Contents/MacOS/TablePlus"
```

Thanks @rfay for the support! <3

rfay:
Ah, I missed the context that this PR didn't actually go in. Sorry for the confusion.

PR #2704, opened 2020-12-19
Add TYPO3 11 Backend URL rewriting support and php7.4 default for TYPO3

gilbertsoft:
## The Problem/Issue/Bug:

With the [patch 67083](https://review.typo3.org/c/Packages/TYPO3.CMS/+/67083) TYPO3 introduces URL rewriting for the Backend and therefor the Nginx configuration needs to be adapted like explained for Apache see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-93048-BackendURLRewrites.html#migration. Currently after the login to the backend this error occurs:
`The page did not exist or was inaccessible. Reason: The requested page does not exist`

Also minimal support for PHP was raised to 7.4 which must also be addressed.

This change adds an additional configuration for Nginx and the default PHP version for the project type typo3 was changed to 7.4.

Without this change the backend will fail after login with this error: `The page did not exist or was inaccessible. Reason: The requested page does not exist`

Testing instructions:

```
ddev config --project-type=typo3 --docroot=public --create-docroot
ddev composer create "typo3/cms-base-distribution:dev-master"
ddev exec touch public/FIRST_INSTALL
ddev launch
```

- Finish the Install Wizard
- Login to the Backend

If no error is shown the test is successful.

This change was tested with TYPO3 11 (master), 10.4 and 9.5 and works fine so far.

## Related Issue Link(s):

- https://github.com/TYPO3/TYPO3.CMS/commit/07bfbc849140d6164e9a146b70ab3643d7557ef8
- https://github.com/TYPO3/TYPO3.CMS.BaseDistribution/pull/30


maddy2101:
did you leave out the apache config change on purpose?

gilbertsoft:
> 
> 
> did you leave out the apache config change on purpose?

Yes, Apache is handled with the default .htaccess delivered by the install tool see https://review.typo3.org/c/Packages/TYPO3.CMS/+/67083/22/typo3/sysext/install/Resources/Private/FolderStructureTemplateFiles/root-htaccess. We do not have a special configuration in DDEV her. But thanks for the hint!

maddy2101:
all right, better safe than sorry :) Thank you for the effort and all the work. I have no idea what I should do these days if there were no ddev.

pkg/ddevapp/webserver_config_packr_assets/nginx-site-typo3.conf
```diff
@@ -31,6 +31,12 @@ server {
         try_files $uri $uri/ /index.php$is_args$args;
     }
 
+    # TYPO3 11 Backend URL rewriting support
+    location /typo3/ {
+        absolute_redirect off;
+        try_files $uri $uri/ /typo3/index.php$is_args$args;
```
bnf, line 37: I would recommend to use:
```
try_files $uri /typo3/index.php$is_args$args;
```
The TYPO3 backend does not nested index.php files to be evaluated and directory indexes aren't desired either. Avoid `$uri/` saves a roundtrip for "good" routes.


Also an additional explicit route for `/typo3` would be nice:
```
location = /typo3 {
        return 302 /typo3/;
}
```

gilbertsoft: Thanks @bnf !

rfay (changes requested):
Please do a 'make packr2: and commit the results.

gilbertsoft:
> 
> 
>     * Please do a 'make packr2: and commit the results.
> 
>     * Please explain more clearly how to test and verify that this does what's needed ior doesn't). Does it ail on the backend without this patch?

Changed the PR description above. Will check again once my PR for the base-dist was merged...

gilbertsoft:
Tested again now with the latest (green) build and everything works fine and as expected.

Updated also the testing instructions.

rfay (approved):
I tested this without the patch and got the failure logging into the backend.

I tested with the patch and it logs in fine.

I tested a `ddev config` from scratch, and when typo3 project type is detected or specified, it uses php7.4 by default. 

Seems good to go. 

I note in passing that I'm unable to log into the backend using http, and no message is given, it just keeps coming back to the login page.

liayn:
For the record: the referenced documentation has been moved to: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.0/Breaking-93048-BackendURLRewrites.html#migration

PR #2708, opened 2020-12-21
Use Composer 2 for project type typo3

gilbertsoft:


rfay (changes requested):
It looks like this has the already-committed stanza for TYPO3 in it, but otherwise should be OK. Time to rework it. Shouldn't need the packr2 stuff any more right?

gilbertsoft:
Let's close this for the moment until all issues with Composer 2 are solved especially this one https://github.com/composer/installers/issues/470

PR #2711, opened 2020-12-21
Update auth ssh doc to correct --ssh-key-path switch name

chellman:
## The Problem/Issue/Bug:

The docs for `ddev auth ssh` were still referencing `--keydir` as the long name of the `-d` switch.

Corrected the long description to display `--ssh-key-path` instead.

`ddev auth ssh --help` and see the glory of the correct switch. 😄


rfay (approved):
Looks great, thanks. Could you please review the docs and see if anything should be updated there with this PR?

chellman:
I see no other ssh auth-related funny business in the docs. Should be good to go.

PR #2716, opened 2020-12-29
Bump PHP default to php7.4, fixes #2657

### LINKED ISSUE #2657: Bump PHP default to 7.4
**Describe the bug**

7.4 will be the most appropriate version of PHP for most sites by the time v1.17 comes out. Change the default to 7.4

rfay:
## The Problem/Issue/Bug:

Time marches on and now PHP does too. PHP7.4 is now the stable version, so we should use it by default. 

Change the default in both web container and ddev code to 7.4

This also removes the typo3ConfigOverrideAction that was added for TYPO3 v11.

- [x]  `ddev config` a project without specifying PHP version.
- [x]  It should have PHP7.4. `ddev exec php --version`
- [x] `docker run -it --rm drud/ddev-webserver:20201229_php7.4 php --version` should also give 7.4


rfay:
This is failing on magento 2. And I think TestDdevFullSiteSetup is using an older version of Magento 2, probably 2.2 or something. So the test files need to be rejiggered. And it looks a little complicated. 

https://github.com/drud/ddev/blob/5d17f0e5dccc321104489bc1ca9586186b6de552/pkg/ddevapp/ddevapp_test.go#L137-L150

When fixing this, name those files better so we know what magento2 version is inside.

PR #2717, opened 2020-12-29
Bump default mariadb version to 10.3, fixes #2232

### LINKED ISSUE #2232: Bump mariadb default install to 10.3.x

rfay:
## The Problem/Issue/Bug:

MariaDB 10.3 is required by several CMSs now, we're leaving 10.2 behind. #2232 

Change the default to 10.3. 10.2 is still fully supported, and any project that has it configured will still use it.

- [ ] `ddev config` a project from scratch. The .ddev/config.yaml should get mariadb 10.3


rfay:
Finally for @damienmckenna ! Yay!

Issue #2547, opened 2020-10-09
'ddev composer' does not support quotes, making aliases impossible


marcvangend:
**Describe the bug**

The way to require a package with an alias, as the [Drupal 9 upgrade documentation](https://www.drupal.org/docs/upgrading-drupal/upgrading-from-drupal-8-to-drupal-9-or-higher) tells you to do, is `composer require "drupal/core:9.0.0 as 8.9.0"`.

However with `ddev composer ...` this results in an error:

```bash
$ ddev composer require "drupal/core:9.0.0 as 8.9.0" --no-update

                              
  [InvalidArgumentException]  
  Could not find package as.  
                              
  Did you mean one of these?  
      ext-hash                
      ramsey/uuid             
      drupal/zen              
      drupal/cog              
      drupal/bfd              
                              

require [--dev] [--prefer-source] [--prefer-dist] [--fixed] [--no-progress] [--no-suggest] [--no-update] [--no-scripts] [--update-no-dev] [--update-with-dependencies] [--update-with-all-dependencies] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [--sort-packages] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--] [<packages>]...

composer [require drupal/core:9.0.0 as 8.9.0 --no-update] failed, composer command failed: exit status 1. stderr= 
```

It seems that ddev composer is stripping out the quotes, which causes "as" to be interpreted as a package name.

**To Reproduce**

1. Have a standard composer-based Drupal 8 ddev installation.
2. Run `ddev composer require "drupal/core:9.0.0 as 8.9.0" --no-update`

**Expected behavior**
I expect composer to add "drupal/core:9.0.0 as 8.9.0" as a requirement in composer.json.

**Version and configuration information (please complete the following information):**

- Host computer OS and Version: macOS Catalina
- Docker Desktop version: 2.4.0.0
- ddev version information:
  ```
  DDEV-Local version      v1.14.2                                
  commit                  v1.14.2                                
  db                      drud/ddev-dbserver-mariadb-10.2:v1.14.1
  dba                     phpmyadmin/phpmyadmin:5                
  ddev-ssh-agent          drud/ddev-ssh-agent:v1.14.0            
  docker                  19.03.13                               
  docker-compose          1.27.4                                 
  os                      darwin                                 
  router                  drud/ddev-router:v1.14.0               
  web                     drud/ddev-webserver:v1.14.2 
[...]

PR #2718, opened 2020-12-30
[docs only] Add "Limitations using ddev composer"

rowild:
`ddev composer` does not work in every scenario (e.g. `ddev composer --version`). This PR adds examples and workarounds to the docu based on #2547 and Slack exchange from 2020-21-30.


rfay:
Thanks!

rfay (approved):
Thanks for this!

PR #2726, opened 2021-01-04
Remove duplicate menu entry for shell autocompletion in docs

rowild:
The menu entries "Shell completion" and "Bach and Zsh Completion" did the same thing. Removed the latter.

WARNING:
My local copy behaves differently than the live page: it shows only the first h3 submenu level instead of all h2 and h3. Not sure why...


rfay (approved):
Perfect, thanks!

PR #2742, opened 2021-01-11
Allow setting web environment in config.yaml, fixes #871

### LINKED ISSUE #871: Support setting environment variables directly from config.yaml

TYPO3 uses environment variables (in php) for various uses that are important to development, especially TYPO3_CONTEXT=Development. 

We have various ways to set these but they're pretty twiddly. See https://stackoverflow.com/questions/50109020/how-can-i-set-environment-variables-inside-ddevs-containers and https://stackoverflow.com/questions/49997842/how-to-set-typo3-context-in-ddev

Add environment variable spec to config.yaml:
```
environment:
  web:
    - WEBVAR1=value1
    - WEBVAR2=value2:
  db:
    - DBVAR1=value1
```

Related PR that does this in a more complicated way: https://github.com/drud/ddev/pull/870

Note: After this is addressed the related Stack Overflow articles should be addressed.

rfay:
## The Problem/Issue/Bug:

#871: Allow setting environment variables in global and local configuration config.yaml

Add `web_environment` key

`ddev config --web-environment="SOMEENV=somevalue" && ddev start && ddev exec 'echo $SOMEENV'`
`ddev config global --web-environment="SOMEGLOBALENV=someglobalvalue" && ddev start && ddev exec 'echo $SOMEGLOBALENV'`

## Release/Deployment notes:

* TYPO3 folks often use environment variable via docker-compose.env.yaml. Teach how to not need that.
* There are Stack Overflow articles that explain this as well. Update them.


rfay:
Still needs 
* Replacement of example in docs (because it's unneeded)
* test coverage

PR #2745, opened 2021-01-14
Improve DDEV configuration validation errors (php_version, mysql_version, etc.)

markhalliwell:
## The Problem/Issue/Bug:

Container config validation errors can be a little confusing as, while a previous version may have (at one point) been "valid", the current version of ddev may no longer actually support it. Note: this came up when a project which worked on previous Macs (using MariaDB 10.0) would show this error because of the new Apple Silicon arm64 architecture; of which MariaDB does not include packages for this version.

Changes the verbiage of "invalid" to "unsupported" and as specifies that it is ddev (and current architecture) that does not support the requested version.

Specify a version that isn't supported and check that the message matches the PR.

Not sure if this is tested anywhere?

## Release/Deployment notes:
Doesn't affect anything, just a better error message.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -439,28 +439,31 @@ func (app *DdevApp) ValidateConfig() error {
 
 	// validate PHP version
 	if !nodeps.IsValidPHPVersion(app.PHPVersion) {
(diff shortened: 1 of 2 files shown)


rfay:
Rebased.

rfay (approved):
This is a great improvement. Thanks so much for taking the time to contribute it. Feel free to take a look at my [commit](https://github.com/drud/ddev/pull/2745/commits/a354f498c0f7cecaedfe15b78432078ddf05fe19) and see if you have any comments about it.

markhalliwell:
LGTM

PR #2747, opened 2021-01-16
Large snapshot restore shouldn't return until it's done

rfay:
## The Problem/Issue/Bug:

The db container shows healthy during a snapshot restore, meaning the project can come up and look healthy, but in reality it's still restoring.

* Make sure that db container shows unhealthy while restoring.

## TODO

- [x] Remove the verbiage that says the log may still be going. This is both in snapshot restore and also in the docs.
- [x] Tell people they may have to wait. Maybe guess based on size of snapshot?
- [x] Change the timeout/numtries on snapshot restore so it waits a long time
- [x] See if there's any way to give feedback about how long it's going to be.

### DIFF
--- a/containers/ddev-dbserver/files/docker-entrypoint.sh
+++ b/containers/ddev-dbserver/files/docker-entrypoint.sh
@@ -55,7 +55,11 @@ if command -v xtrabackup; then BACKUPTOOL="xtrabackup"; fi
 # If mariadb has not been initialized, copy in the base image from either the default starter image (/mysqlbase)
 # or from a provided $snapshot_dir.
 if [ ! -f "/var/lib/mysql/db_mariadb_version.txt" ]; then
-    touch /tmp/initializing
+    # If snapshot_dir is not set, this is a normal startup, so
+    # tell healthcheck to wait by touching /tmp/initializing
+    if [ -z "${snapshot_dir}" ] ; then
+      touch /tmp/initializing
+    fi
     target=${snapshot_dir:-/mysqlbase/}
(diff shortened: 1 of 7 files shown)


croane:
I am using the latest alpha version of ddev. The snapshot stuff was working great, but then all of a sudden the snapshots wouldn't restore and the db was empty. After looking at the log, it looks like my mac somehow created .DB_Store files in the snapshot directories, which broke the restoration. I was able to manually fix by running: find . -name '.DS_Store' -type f -delete ... in the snapshot directory.

Would it be possible to delete these .DS_Store files before restoring, to prevent this issue?

Also, another idea that might help with space issues. I work on a massive site where the DB snapshot is >15GB. Could you compress the snapshot directory and decompress it when a restore is needed? That would save a ton of space with a bunch of snapshots. Just a thought. :)

rfay:
Well, you're the second person to report this problem with .DS_Store, so I guess we have to do something about it. So you must have been using the finder to look at the .ddev/db_snapshots directory? 

I doubt that this is a new thing at all, maybe the two of you that experienced this are just using the Finder the way you are.

Could you please open a separate issue with the exact experience you had, and perhaps a way to recreate it?

croane:
I think I might have just done some random browsing in finder or something. I'm not doing anything specific with the snapshot files. After I get through some tasks, I'll go ahead and gather some more details and create an issue. Thanks!

rfay:
For anybody who sees this, the followup issue for the problem with .DS_Store files is https://github.com/drud/ddev/issues/2765

PR #2768, opened 2021-02-01
Remove magerun from web container as people use composer-installed version

rfay:
## The Problem/Issue/Bug:

Since the beginning of magento/magento2 support we've installed magerun and magerun2 in /usr/local/bin, but all instructions and expectations say to use bin/magento, the composer-installed binary.

Remove magerun, magerun2.


cmuench:
@rfay n98-magerun is an addition to the bin/magento. It's not the same tool.
It's like wp-cli or drush.

cmuench:
I added a wiki page to describe how the tool can be installed in a ddev environment.

See: https://github.com/netz98/n98-magerun2/wiki/ddev

rfay:
OK, i never heard of anybody using them, and thought they were the same! If you think they should go back in, we can put them back in. 

As far as your wiki page, I'd suggest that it would be nicer to do it in a .ddev/web-build/Dockerfile than a post-start hook though, so it doesn't have to be done on every ddev start.

PR #2783, opened 2021-02-05
Adds mime type for compressed javascript and css with a .gzip extension to fix TYPO3 compression problem

unn:
## The Problem/Issue/Bug:
When compression is configured with either `$GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel']`  or `$GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel']` the nginx default mime type for .gzip sends octet-stream resulting in a broken site.

Adds mime type for compressed javascript and css with a .gzip extension per https://forge.typo3.org/issues/54316


rfay (changes requested):
This requires a `make packr2` to take effect.

tbal:
I would suggest taking the part from https://docs.typo3.org/m/typo3/guide-installation/master/en-us/In-depth/SystemRequirements/Index.html#nginx with added `add_header Content-Encoding gzip` and `gzip off`. That's how we use it in production too for several instances.

gilbertsoft:
Will check it tomorrow...

rfay:
Thanks so much for this contribution @unn 

I tried to demonstrate the problem using TYPO3 v11 and wasn't able to make it break by adding `$GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel'] = 9;` to the bottom of AdditionalConfiguration.php (and killing off the #ddev-generated line) and then visiting the backend.

Could you please give more details on how to recreate the error so I can test the fix?

unn:
I was testing with a typo3 site, so maybe there's a difference between typo3 10 and 11 when it comes to the way frontend/backend compression work.

It's pretty easy to reproduce in 10, unless there is something special about my setup, which could be the case, though I doubt it.

mabolek:
As far as I know, the compression works the same in TYPO3 v10 and v11, but I haven't checked the latest dev version. Please note that, v11 won't have a long-term support release (i.e. recommended for production) until October. A lot can happen with v11 until then. For most people, v10 will probably the default version used until closer to the v11 release date.

rfay:
I tried the same approach as in https://github.com/drud/ddev/pull/2783#issuecomment-777942813 with TYPO3 10.4.12, on the back-end. I couldn't recreate the problem. I know there's something important I'm missing.

unn:
For clarity's sake I can reproduce this 100% of the time: typo3 10.4.12, fresh install using this AdditionalConfiguration.php

```
<?php
if (getenv('IS_DDEV_PROJECT') == 'true') {
    $GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel'] = 9;
    $GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel'] = 9;
    $GLOBALS['TYPO3_CONF_VARS'] = array_replace_recursive(
        $GLOBALS['TYPO3_CONF_VARS'],
        [
            'DB' => [
                'Connections' => [
                    'Default' => [
                        'dbname' => 'db',
                        'host' => 'db',
                        'password' => 'db',
                        'port' => '3306',
                        'user' => 'db',
                    ],
                ],
            ],
            // This GFX configuration allows processing by installed ImageMagick 6
            'GFX' => [
                'processor' => 'ImageMagick',
                'processor_path' => '/usr/bin/',
                'processor_path_lzw' => '/usr/bin/',
            ],
            // This mail configuration sends all emails to mailhog
            'MAIL' => [
                'transport' => 'smtp',
                'transport_smtp_server' => 'localhost:1025',
            ],
            'SYS' => [
                'trustedHostsPattern' => '.*.*',
                'devIPmask' => '*',
                'displayErrors' => 1,
            ],
        ]
    );
}
```

rfay:
I'm completely unable to recreate the behavior described in the OP using ddev v1.16.6 or v1.17.0-alpha4. I used your exact configuration from https://github.com/drud/ddev/pull/2783#issuecomment-778261595 @unn 

Is there some configuration that you have set in TYPO3 that I need to do?

Please give complete recreation instructions, including what URL to hit. Please give your recreation step-by step.

gilbertsoft (changes requested):
I'm unsure about this change. The default behavior of the DDEV Nginx config is to compress these files out of the box. The settings     `$GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel']` and `$GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel']` are used to enabled the compression of PHP which is in every case bad in performance compared to the webserver compression. These settings should only be changed if you do not have the possibility to enable compression at the webserver level.
So I would like not to include this change to the default configuration of DDEV because it's superfluous at all.

unn:
I don't agree with that. If the argument is I shouldn't be using it, I don't have a rebuttal to that. It's in use and as such I would like a working site.

The issue ultimately is https://forge.typo3.org/issues/75043 in OPs point 2 and the fix is https://forge.typo3.org/issues/93182

In either case until this is fixed or editable in typo3s upstream, ddev should make efforts to support this minor configuration in nginx.

gilbertsoft:
TBH if you have to use this configuration you should probably change the hosting provider. The best option is using the webserver's capabillities here and not rely on PHP features which are a workaround only for these bad wrong configured servers.

> In either case until this is fixed or editable in typo3s upstream, ddev should make efforts to support this minor configuration in nginx.

Agree, but if you like to add support for these projects then a good workaround would be overwriting the two settings in the DDEV's default AdditionalConfiguration.php with 0 and not the other way around. If somebody likes to use the bad way with PHP's compression he can change it by his own.

DDEV is using here the best possible configuration afaik, so why making things worser? Not DDEV has here to change something in the first place but the providers using a bad webserver configuration. Your change disables this good configuration and introduces a bad one. Everybody who does not use these two configuration options in TYPO3 afterwards will end with no compression at all.

rfay:
@gilbertsoft I note that @unn is not suggesting to change the generated AdditionalConfiguration.php at all, only to handle the .gzip file created when somebody *does* add `$GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel']` manually or whatever.

gilbertsoft:
> 
> 
> @gilbertsoft I note that @unn is not suggesting to change the generated AdditionalConfiguration.php at all, only to handle the .gzip file created when somebody _does_ add `$GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel']` manually or whatever.

Yes, that's clear. But I suggest to add the two lines from the comment above with 0 instead of changing the Nginx configuration. This will make sure DDEV still uses the best possible configuration regarding compression but also support projects which have these two (or one of them) settings in the LocalConfiguration.php. That's exactly the point where the issue is generated though.

rfay:
So you suggest overriding anything the user would have done and turn off compression. That makes some sense, unless they're just trying to match server config.

gilbertsoft:
> 
> 
> So you suggest overriding anything the user would have done and turn off compression. That makes some sense, unless they're just trying to match server config.

Exactly. For the case where the users likes to match exactly the server's configuration the user should change the Nginx configuration like suggested in this PR by his own.

We have here two different setups, one good like done by DDEV using the webserver's compression and one bad which is using PHP's compression. We only can support one of them as long as we do not introduce an option in the config.yaml to change this behavior.

I already had this issue one or two times in the DDEV channel and I always suggested the users to disable PHP's compression and nobody every compained about it. That's why I think we do not need an other option to change this behavior here in the first place. A hint in the docs or an example how to achieve this in ddev-contrib would be easier imho.

gilbertsoft:
Just tested the following and works just fine. This could get the proposition for this issue. I've simply added the file `.ddev/nginx/compression.conf` with the following content and activate the PHP's compression with the two TYPO3 settings:

```
    # Compressing resource files will save bandwidth and so improve loading speed especially for users
    # with slower internet connections. TYPO3 can compress the .js and .css files for you.
    # *) Set $GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel'] = 9 for the Backend
    # *) Set $GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel'] = 9 together with the TypoScript properties
    #    config.compressJs and config.compressCss for GZIP compression of Frontend JS and CSS files.
    location ~ \.js\.gzip$ {
        add_header Content-Encoding gzip;
        gzip off;
        types { text/javascript gzip; }
    }
    location ~ \.css\.gzip$ {
        add_header Content-Encoding gzip;
        gzip off;
        types { text/css gzip; }
    }
```

rfay:
@unn I do note that there was strong pushback in #typo3 about ever using the "compressionlevel" feature, so there seems to be consensus that it's not best practice. @gilbertsoft 's solution does seem elegant - comes at it from the other direction.

gilbertsoft:
Oh, just changed my mind about this change. It does only disable the webserver's compression in case the file is already compressed by PHP. So this change does not hurt at al, sorry I didn't get that before.

But the other part of course is still valid, one should never use the PHP's compression if it could be done by the webserver.

gilbertsoft (approved):
Finally I'm fine with this. Thanks!

rfay:
Yay that we have community approval, even with community disapproval of the php-level compression option.

All I need now is a way to manually recreate this. Perhaps a screencast would help.

rfay:
I'd love to have this go in @unn but I haven't figured out a way to demonstrate the failure in a manual test. So if you can give me a step-by-step technique to do that, I'll demonstrate the bug and the fix. Otherwise, I'll close this in a week or so.

gilbertsoft:
To reproduce do the following steps:

- download or clone my TYPO Dirstribution demo see https://github.com/GsTYPO3/typo3-introduction#quick-start
- run `ddev start` -> the backend opens, we'll need it later to clear the caches
- run `ddev launch` -> the frontend opens and should show the following screen

grafik

- edit the public/typo3conf/AdditionalConfiguration.php and add the following two lines and remove the ddev-generated comment

```
    $GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel'] = 9;
    $GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel'] = 9;
```

Now go to the previously opened backend (or run `ddev launch typo3`), login with the choosen credentials during the installation and flush all caches:

grafik

Head to the frontend again and it's now broken:

grafik

Now use the new ddev binary or add the fix to `.ddev/nginx/compression.conf`:

```
    # Compressing resource files will save bandwidth and so improve loading speed especially for users
    # with slower internet connections. TYPO3 can compress the .js and .css files for you.
    # *) Set $GLOBALS['TYPO3_CONF_VARS']['BE']['compressionLevel'] = 9 for the Backend
    # *) Set $GLOBALS['TYPO3_CONF_VARS']['FE']['compressionLevel'] = 9 together with the TypoScript properties
    #    config.compressJs and config.compressCss for GZIP compression of Frontend JS and CSS files.
    location ~ \.js\.gzip$ {
        add_header Content-Encoding gzip;
        gzip off;
        types { text/javascript gzip; }
    }
    location ~ \.css\.gzip$ {
        add_header Content-Encoding gzip;
        gzip off;
        types { text/css gzip; }
    }
```

Restart the project by `ddev restart` and head again to the frontend (`ddev launch`). The frontend should be working again now.

rfay:
Rebased.

rfay (approved):
This is working for me now, thanks for the repro coaching @gilbertsoft . 

The latest problem I had was that the *browser* was caching the bad admin page even though I had the inspector open. Surprised me.

rfay:
Thanks for all the work on this! Sorry it was so hard to get manually tested. Lots of things can go wrong.

PR #2785, opened 2021-02-05
Add support for persistent data on old Solr.

damontgomery:
This provides support for persistent data on older versions of Solr. Tested with 5.

It seems like 5 stores data in `/opt/solr/server/solr/mycores/dev/data` while 8 stores data in `/var/solr/data/dev/data`.

Solr_Admin_and_Solr_Admin

This breaks the `The persistent volume should have the same name as the service so it can be deleted when the project is deleted.` part.

I'm not sure if `5` uses `/var/solr` at all, although there was stuff in it. Maybe an alternative is to add a comment to pick the relevant directory based on versions?

## The Problem/Issue/Bug:

Solr 5 does not persist data.

Add an additional volume that is used by Solr 5 and maybe other older versions.

Follow the standard Solr setup direction, but set `image: solr:5`.

Without this code:
- Spin up the server
- View the core at `:8983/solr/#/dev`
- Add a document from the UI `:8983/solr/#/dev/documents`
- View the core and see the document count
- Restart DDEV `ddev restart`
- View the core, it will be `0` again

No tests were added. For automated tests, we'd have to include Solr 5 configuration files in addition to 8.

### DIFF
--- a/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml
+++ b/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml
@@ -51,10 +51,12 @@ services:
       # to the host port 8984 vid ddev-router reverse proxy.
       - HTTPS_EXPOSE=8984:8983
     volumes:
-      # solr core *data* is stored on the 'solr' docker volume
+      # solr core *data* is stored on the 'solr_var' or `solr_opt` docker
+      # volumes depending on the version of solr.
       # This mount is optional; without it your search index disappears
       # each time the ddev project is stopped and started.
-      - solr:/var/solr
+      - solr_var:/var/solr
+      - solr_opt:/opt/solr
 
       # This mounts the conf in .ddev/solr into the container where
       # the solr-precreate command in the entrypoint uses it as a one-time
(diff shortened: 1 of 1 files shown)


rfay:
If I understand you right, you're saying that the current recipe doesn't work with an ancient version of solr, true?

Please open a PR to explain this in https://github.com/drud/ddev-contrib for non-current versions of solr, just like there's one there for TYPO3. We can't support every permutation of every product in every situation or everybody will just be confused out of their wits.

damontgomery:
I think that makes sense. I see the file in https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/typo3-solr.

The current documentation suggests we should be able to use older versions and I think it would be helpful to mention this to people there: https://ddev.readthedocs.io/en/stable/users/extend/additional-services/

`The recommended Solr version is: image: solr:8, from hub.docker.com.`

I agree that supporting long outdated versions of Solr sounds pointless, but until recently, this was a requirement when working with Acquia. More of an Acquia problem then a DDEV problem, but given the overlap between DDEV / Drupal / Acquia, I don't think it's pointless to call out that issue.

I can put together an example / notes probably next week. Please feel free to close this if that makes sense.

rfay:
Thanks, and again congratulations for tracking this down. Solr is one of the most convoluted things out there. To solve problems you have to understand solr itself, perhaps many versions, the dockerhub solr implementation, the client software (like drupal's solr module), and then docker on top of all of it. It's craziness. 

Closing this one and looking forward to your ddev-contrib contribution.

PR #2793, opened 2021-02-09
Update OSX install step order for clarity.

adamzimmermann:
## The Problem/Issue/Bug:

The setup.sh script previously seemed like a step under the `macOS Catalina (and above) warning:` heading. Once I looked at the markdown I realized that wasn't a heading, but instead just bold text. However it caused me a lot of confusion since I wasn't running the `setup.sh` script and everything was failing.

I re-ordered the text so it is clear that the general section is not a sub-section to the Catalina specific heading. Even if my change isn't what is desired, some structure change to indicate the correct associations would be great!

N/A

N/A

## Related Issue Link(s):
N/A

### DIFF
--- a/docs/users/performance.md
+++ b/docs/users/performance.md
@@ -26,14 +26,14 @@ Note that you can use the NFS setup described for each operating system below (a
 
 ### macOS NFS Setup
 
-__macOS Mojave (and above) warning:__ You'll need to temporarily give your terminal "Full disk access" before you (or the script provided) can edit /etc/exports. If you're using iterm2, here are [full instructions for iterm2](https://gitlab.com/gnachman/iterm2/wikis/fulldiskaccess). The basic idea is that in the System Preferences -> Security and Privacy -> Privacy you need to give "Full Disk Access" permissions to your terminal app. Note that the "Full Disk Access" privilege is only needed when the /etc/exports file is being edited by you, usually a one-time event.
-
-__macOS Catalina (and above) warning:__ If the projects are in a subdirectory of the ~/Documents directory or on an external drive, it is necessary to grant the "Full Disk Access" permission to the `/sbin/nfsd` binary. Full details are [below](#upgrading-catalina).
-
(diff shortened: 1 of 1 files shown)


adamzimmermann:
Screen Shot 2021-02-09 at 2 30 58 PM

rfay:
Thanks for this! I think the language can be toned down a little too, since macOS has matured a bit in how those full disk access privs are granted.

docs/users/performance.md
```diff
@@ -26,15 +26,15 @@ Note that you can use the NFS setup described for each operating system below (a
...
 
 If your DDEV-Local projects are set up outside your home directory, you'll need to edit /etc/exports to add a line for that share as well.
 `sudo vi /etc/exports` and copy the line the script has just created (`/System/Volumes/Data/Users/username -alldirs -mapall=<your_user_id>:20 localhost`), editing it with the additional path, e.g: `/Volumes/SomeExternalDrive -alldirs -mapall=<your_uid>:20 localhost`.
 
+__macOS Mojave (and above) warning:__ You'll need to temporarily give your terminal "Full disk access" before you (or the script provided) can edit /etc/exports. If you're using iterm2, here are [full instructions for iterm2](https://gitlab.com/gnachman/iterm2/wikis/fulldiskaccess). The basic idea is that in the System Preferences -> Security and Privacy -> Privacy you need to give "Full Disk Access" permissions to your terminal app. Note that the "Full Disk Access" privilege is only needed when the /etc/exports file is being edited by you, usually a one-time event.
```
rfay, line 34:
```suggestion
__Warning:__ You may need to temporarily give your terminal "Full disk access" before you (or the script provided) can edit /etc/exports. If you're using iterm2, here are [full instructions for iterm2](https://gitlab.com/gnachman/iterm2/wikis/fulldiskaccess). The basic idea is that in the System Preferences -> Security and Privacy -> Privacy you need to give "Full Disk Access" permissions to your terminal app. Note that the "Full Disk Access" privilege is only needed when the /etc/exports file is being edited by you, usually a one-time event.
```

docs/users/performance.md
```diff
@@ -26,15 +26,15 @@ Note that you can use the NFS setup described for each operating system below (a
...
+
+__macOS Catalina (and above) warning:__ If the projects are in a subdirectory of the ~/Documents directory or on an external drive, it is necessary to grant the "Full Disk Access" permission to the `/sbin/nfsd` binary. Full details are [below](#upgrading-catalina).
```
rfay, line 36:
```suggestion
__Warning:__ If the projects are in a subdirectory of the ~/Documents directory or on an external drive, it is necessary to grant the "Full Disk Access" permission to the `/sbin/nfsd` binary. Full details are [below](#upgrading-catalina).
```

rfay (approved):
Thanks, minor rewording suggestions as this text has gotten old and stale.

PR #2806, opened 2021-02-14
Remove sudo from ddev-dbserver and ddev-ssh-agent, limit use in ddev-webserver, fixes #2607

### LINKED ISSUE #2607: Remove sudo from ddev-dbserver or at least from a hardened version
**Describe the bug**

ddev-dbserver has sudo and uses it. A fair bit. And it has a writeable .ddev mount in /mnt/ddev_config, so a break-in might be able to use sudo to make a setuid file on the host. 

It's unlikely that shell-access to the dbserver would be obtained, but it would sure be better if we could just get sudo off of here. 

A related technique (leaving sudo here) would be to share only the .ddev/db_snapshots directory as writeable, and share the .ddev directory as read-only. Might be easier and better.

**To Reproduce**

**Expected behavior**

**Screenshots**

- Host computer OS and Version: [e.g. Windows 10, macOS Catalina]
- Docker Desktop version if on macOS or Windows (from "About Docker Desktop")

**Additional context**

rfay:
## The Problem/Issue/Bug:

#2607: Having sudo in images at least on a server is a security risk.

This also (if it works) could alleviate the Gitpod setuid problem in https://github.com/gitpod-io/gitpod/issues/3174

* remove sudo and its usages from both ddev-dbserver and ddev-ssh-agent


rfay:
Consequences of this:
* Fails if you add config in .ddev/mysql, but it may well be better to just include /mnt/ddev_config/mysql instead of copying in: https://github.com/drud/ddev/pull/2806/checks?check_run_id=1895958454
  ```
  # + '[' -d /mnt/ddev_config/mysql -a /mnt/ddev_config/mysql/utf.cnf '!=' '/mnt/ddev_config/mysql/*.cnf' ']'
  # + cp /mnt/ddev_config/mysql/utf.cnf /etc/mysql/conf.d
  # + chmod -R ugo-w /etc/mysql/conf.d
  ```
* chmod (and certainly chown) don't work where used: `chmod -R ugo+w /var/lib/mysql` https://github.com/drud/ddev/pull/2806/checks?check_run_id=1895958441
* TestServices seems to fail on Linux and macOS
* TestGetLocalHTTPResponse fails on macOS
* TestComposer and TestComposerVersion fail on WIndows https://buildkite.com/drud/ddev-macos-use-nfsmount/builds/2834#c940fcce-b90e-4c93-b936-b39dca4cbdc8
* SO many things fail on Windows. Probably because of difference in filesystem mount behavior and default perms
*

PR #2807, opened 2021-02-14
Remove `ddev start` before `ddev composer` or `ddev launch`

jonaseberle:
`ddev launch` or `ddev composer` will start the project if not up yet. So if the project was not started before, it can be safely omitted.

I think the behaviour that `ddev composer` and `ddev launch` bring up the project was added recently and I love it! 

## The Problem/Issue/Bug:
The "Getting started with ddev" could be even shorter.

It remove a superfluous command from the documentation.

Test the setup of different project types as given in the documentation. 

## Release/Deployment notes:
none, just a documentation change

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -64,7 +64,6 @@ DDEV works happily with most any PHP or static HTML project, although it has spe
 mkdir my-wp-bedrock-site
 cd my-wp-bedrock-site
 ddev config --project-type=wordpress --docroot=web --create-docroot
-ddev start
 ddev composer create roots/bedrock
 ```
 
@@ -83,7 +82,7 @@ Now, since [Bedrock](https://roots.io/bedrock/) uses a configuration technique w
     WP_ENV=development
 ```
 
-You can then `ddev start` and `ddev launch`.
+You can then `ddev start` (or directly `ddev launch`).
 
 For more details see [Bedrock installation](https://roots.io/bedrock/docs/installing-bedrock/)
 
@@ -137,7 +136,6 @@ Get started with Drupal 8 projects on ddev either using a new or existing compos
 mkdir my-drupal8-site
 cd my-drupal8-site
 ddev config --project-type=drupal8 --docroot=web --create-docroot
-ddev start
 ddev composer create "drupal/recommended-project:^8"
 ddev composer require drush/drush
 ddev drush site:install -y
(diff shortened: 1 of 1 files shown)


rfay:
I guess the question is not just about making it as short as possible (a good goal) but also instructing people in basic commands. It's true that shell-based commands and `ddev composer` start ddev if it hasn't been started (and this has always been this way), but `ddev launch` is definitely not a replacement for `ddev start`, the behavior does what people need in this place, but they need to know about `ddev start`.

So I'm open to this, but interested in what you think about the educational vs the quick-starting goals.

jonaseberle:
I would say a Quickstart should really try to showcase how easy it is to use the subject. Even from an educational perspective I would "market" `ddev launch` more, because it makes working with ddev one little step easier. 

I have copy-pasted the project URL from `ddev describe` for some time until I discovered `ddev launch`... I think it sneaked into some tutorials early on (for example https://docs.typo3.org/m/typo3/guide-contributionworkflow/master/en-us/Appendix/SettingUpTypo3Ddev.html) and it stuck a bit. 

I think that page ([Using the CLI](https://ddev.readthedocs.io/en/stable/users/cli-usage/)) is a bit too convoluted at the moment. Maybe it could be split into "Getting started", "Commands", "Start a ddev project", "Import volatile user data/assets" or similar?

- so I think this patch would not really help much.

*About "learning ddev":*

PS: For my peers I set up `ddev` so that starting a project will trigger `composer install` and compiling assets etc. That way I usually introduce them to `ddev launch` and tell them "If after a pull something is missing, run `ddev (re)start`". 

PPS: I recently advocated against more ddev-docs inside TYPO3 docs because it will never be as complete and up-to-date as the official docs. It occured that the official (fantastic!) troubleshooting docs were not known – we shall really point people more towards the official documentation and contribute. In my company, a self-written "ddev on Windows"-guide is circulating but I have not been successful yet in putting an end to it... Your documentation is so well-thought and tackles all the (cross-platform) questions but somehow users do want to re-invent it.

rfay:
Thanks for the thoughtful response. My problem is that `ddev launch` only does a start (if the project is stopped) as a side-effect. Mostly it's *intended* as a way to launch a browser, not a way to start the project. In fact it does *not* do a start if the project is already started, which can lead to confusion. 

I'd love to see a docs reorg. Unfortunately, when something grows organically like this, it can get to be too much. But issues/PRs with specific proposals are completely welcome.

jonaseberle:
That's ok. If I come up with an idea for the docs, I'll let you know. Thank you!

PR #2829, opened 2021-02-23
Reintroduce magerun and magerun2

rfay:
## The Problem/Issue/Bug:

magerun and magerun2 were mistakenly removed from the web image in #2768 

Bring it back in. Turns out it wasn't exactly completely removed, and this should be a better installation anyway

`ddev exec magerun2`


rfay:
@cmuench seemed to me like these should go back in. Would appreciate your review.  Sorry I didn't understand.

PR #2831, opened 2021-02-24
Be way less aggressive with removing CREATE DATABASE or USE statements in ddev import-db, fixes #2787

### LINKED ISSUE #2787: `ddev import-db` Removal of DDL statements in SQL dump can corrupt data if it contains DDL in the data
**Describe the bug**

I have SQL exported from my prod server (MySQL 5.7) using `drush sql:dump` into a gzip SQL file.

I import it into my multisite setup like this:

`ddev import-db --target-db=SITE --src=/Users/nthompson/Downloads/SITE.sql.gz`

However... The `node__body` table is empty. There are no errors shown during import. The SQL file definitely has INSERT statements for this table.

If I to to PHPMyAdmin and import the same SQL file, the table then has data in it.

Equally, if I import it via SequelPro (invoked using `ddev sequelpro`, the same gzip file produces a DB with data in `node__body`.

I cannot see any "verbose" or "debug" flags on `import-db` to see what commands it runs.

Other tables do have data in them, although I've not done extensive inspection to see if its complete or there are is other missing data. The missing node body is enough of a bug for me at the moment 😉

**To Reproduce**

Command above: Simply import the DB.

Host: Big Sur 11.1
Docker Desktop:
image

```
DDEV-Local version	v1.16.2
db                	drud/ddev-dbserver-mariadb-10.2:v1.16.0
ddev-ssh-agent    	drud/ddev-ssh-agent:v1.16.0
docker            	20.10.2
docker-compose    	1.27.4
os                	darwin
router            	drud/ddev-router:v1.16.2
web               	drud/ddev-webserver:v1.16.2
```

Config:
```
name: tmj-d8
type: drupal8
docroot: web
php_version: "7.4"
webserver_type: nginx-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
additional_fqdns: []
mariadb_version: ""
mysql_version: "5.7"
provider: default
use_dns_when_possible: true
composer_version: ""
```

## .ddev/config.multisite.yaml
```
additional_hostnames:
  - tmj
  - ktw
hooks:
  post-start:
  - exec: mysql -uroot -proot -e "CREATE DATABASE IF NOT EXISTS tmj; GRANT ALL ON tmj.* to 'db'@'%';"
    service: db
  - exec: mysql -uroot -proot -e "CREATE DATABASE IF NOT EXISTS ktw; GRANT ALL ON ktw.* to 'db'@'%';"
    service: db
```

Let me know if there is anything you'd like me to test..

rfay:
## The Problem/Issue/Bug:

#2787 and https://ddevhq.atlassian.net/browse/DEV-502 - `ddev import-db` can corrupt database that contains mysql DDL inside the content of the rows, but it's only supposed to affect actual DDL "CREATE DATABASE" and "USE"

Tighten up the regex significantly.

- [x] Import the db from https://ddevhq.atlassian.net/browse/DEV-502 and verify that the entire site is functional and that the URLs in that issue are working.
- [x] Import the [test db dump](https://github.com/drud/ddev/blob/327cd23dffd070ca9994ff354c48e671fc8d74c8/pkg/ddevapp/testdata/TestDdevImportDB/posts_with_ddl_content.sql) for the tests in this issue and verify that you get lots of rows in the table. Previously the table would be empty.
- [x] Import the test db dump from stdin, should work.
- [x] Go back to v1.17.0-alpha6 and do the import, the test db dump will be empty.
- [x] Manually test dump provided in https://github.com/drud/ddev/issues/2787#issuecomment-778675152 - should work with new, fail with old ddev (158 rows in node__body with this PR, 0 rows in node__body with ddev v1.17.0-alpha6)

TestDdevImportDB has been updated to test this case both with stdin and with a file.


pkg/ddevapp/ddevapp.go
```diff
@@ -507,12 +507,16 @@ func (app *DdevApp) ImportDB(imPath string, extPath string, progress bool, noDro
...
+	// as has actually happened with https://www.ddev.com/ddev-local/ddev-local-database-management/
+	// and in https://github.com/drud/ddev/issues/2787
+	// The backtick after USE is inserted via fmt.Sprintf argument because it seems there's
+	// no way to escape a backtick in a string literal.
+	inContainerCommand := fmt.Sprintf(`mysql -uroot -proot -e "%s" && pv %s/*.*sql | perl -p -e 's/^(CREATE DATABASE \/\*|USE %s)[^;]*;//' | mysql %s`, preImportSQL, insideContainerImportPath, "`", targetDB)
 
 	// Handle the case where we are reading from stdin
 	if imPath == "" && extPath == "" {
-		inContainerCommand = fmt.Sprintf(`mysql -uroot -proot -e "%s" && perl -p -e 's/^(CREATE DATABASE|USE|.*DROP DATABASE).*$//' | mysql %s`, preImportSQL, targetDB)
+		inContainerCommand = fmt.Sprintf(`mysql -uroot -proot -e "%s" && perl -p -e 's/^(CREATE DATABASE \/\*|USE %s)[^;]*;//' | mysql %s`, preImportSQL, "`", targetDB)
```
wozniakjan, line 519: do we still desire to remove `DROP DATABASE`? I'm wondering if those can be dangerous too. I can imagine a scenario when a user has a number of sites, intends to import to a site `A` but the SQL dump contains `DROP DATABASE b;` and what appeared to be an unrelated site for the `ImportDB` suddenly stops working.

wozniakjan (approved):
Thanks for bringing this issue to my attention, I think I got the gist of it. If some user table contain `CREATE DATABASE`, `DROP DATABASE` or `USE`, we can strip the SQL dump from important lines that we should leave there, correct?

Wondering if by using any form of regexp, we just delay the issue and maybe we need something a bit more elaborate to be correct always:
1) SQL => AST parser
2) remove the undesired statements
3) AST => SQL

But in general, this is likely a good improvement, I will keep this in mind for ddev-live.

pkg/ddevapp/ddevapp.go
rfay: I looked at several dumps and did not find any use of `DROP DATABASE` in a standard dump. However... in a case like this it would be DROP DATABASE IF EXISTS... and the database would not exist. 

But your point about breaking a *different* site is interesting. 

Still, one would expect that if DROP DATABASE *were* in there, it would refer to the current database, so the current site, so would not do harm?

wozniakjan: I acknowledge it's a hypothetical maybe type of scenario that we probably don't need to worry about here. Still wondering if for ddev-live we may need to parse AST, modify there and then generate back the SQL dump to eliminate even any hypothetical scenarios.

rfay: IMO this is all bad. We shouldn't be doing sed/perl/awk against a db dump. I imagine that we could figure out a way to do this only to the header section of the dump.

I hope you noticed some of the limiting factors now added to this version - using [^;]* instead of .* for the removal, and making sure it was anchored.

PR #2840, opened 2021-02-26
Updates local docs theme to match live

unn:
A preview of what this will look like read the docs is available via https://unnddev.readthedocs.io/en/latest/


rfay:
@unn thanks for this work! 

You'll note that the tests in the build fail, it's tested like this: https://github.com/drud/ddev/blob/dc26ba17b1c6a934ded420cae94bebf641e82033/.github/workflows/docscheck.yml#L30-L34

Please update what you think should be updated to make that work correctly.

unn:
Thanks @rfay! On my list.

rfay (changes requested):
The requirements.txt and related things seem like a great step forward. Could we put them in a .mkdocs directory or something so they have context? Maybe put everything mkdocs related into .mkdocs?

requirements.txt
```diff
@@ -0,0 +1,29 @@
+Babel==2.8.0
```
rfay, line 1:
```suggestion
# This requirements.txt file is exclusively for mkdocs docfile creation, and doesn't have anything
# specifically to do with ddev.
Babel==2.8.0
```

rfay (approved):
Looking good, I'll poke around your forked docs and this should go in. Thanks so much!

rfay:
I *love* the table of contents on the right side. Since it's context-sensitive it was a little hard to figure out what it was contents *for* at first, but it's a great step forward. 
Using_the_CLI_-_DDEV-Local_Documentation

rfay:
Yay, thanks and congratulations @unn !

rfay:
Makes me want to do a prerelease just to make this show up.

PR #2841, opened 2021-02-26
move Gitpod's docker-up to a "secondary" terminal

shaal:
## The Problem/Issue/Bug:
Gitpod requires `sudo docker-up` to run docker commands.
That process run indefinitely.
The other terminal that run `make` command and can show progress and becomes available for additional commands of the user is "hidden" as a second terminal.

Switch between the terminal order, so `sudo docker-up` is on secondary terminal, and `make` commands runs in main terminal.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,8 +3,6 @@ image:
   file: .gitpod.Dockerfile
 
 tasks:
-  - name: docker_up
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks good to me, thanks!

PR #2844, opened 2021-03-01
[Gitpod integration] Prepare make (with docker-up) during prebuild

shaal:
## The Problem/Issue/Bug:
Users have to wait for make to run when starting a workspace.

Make (with docker-up) can run during prebuild.
A user can start workspace after make already finished running, and ready to work with `ddev`.

Run Gitpod in this PR, and confirm you can run `ddev` in the terminal.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -10,9 +10,16 @@ tasks:
         sleep 1
       done
       .githooks/linkallchecks.sh
-      make
   - name: docker_up
     command: sudo docker-up      
+  - init: |
+      # Wait for docker to come up before doing make (make requires docker)
+      while ! docker ps; do
+        sleep 1
+      done
+      .githooks/linkallchecks.sh
+      make
+  - init: sudo docker-up &
 
 vscode:
   extensions:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,7 +3,7 @@ image:
   file: .gitpod.Dockerfile
 
 tasks:
-  - name: make
+  - name: ready
     command: |
       # Wait for docker to come up before doing make (make requires docker)
       while ! docker ps; do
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -3,16 +3,23 @@ image:
...
-      make
   - name: docker_up
     command: sudo docker-up      
+  - prebuild: |
+      # Wait for docker to come up before doing make (make requires docker)
+      while ! docker ps; do
+        sleep 1
+      done
+      .githooks/linkallchecks.sh
+      make
```
rfay, line 21: This is awesome. Gets the golang-build-container pulled in advance! Hopefully there aren't permissions problems in there.

rfay (approved):
Love it!

rfay:
I see this in the terminal once it comes up:

```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
Linked pre-push.allchecks as pre-push git hook
```

Looks like you needed to redirect your `docker ps` - `docker ps 2>/dev/null`

rfay (changes requested):
Unfortunately, I don't think this can work until the gitpod permissions bug is fixed.

```
gitpod /workspace/ddev $ make 
building .gotmp/bin/linux_amd64/ddev from ./cmd/... ./pkg/...
go: creating work dir: mkdir /tmp/go-build1509598969: permission denied
make: *** [Makefile:115: .gotmp/bin/linux_amd64/ddev] Error 1```

shaal:
@rfay Thanks for looking into it.
We'll have to wait until the Gitpod bug is fixed - https://github.com/gitpod-io/gitpod/issues/3174

rfay:
It also means (I think) that we'll need the `docker rmi` or docker pull to even make restored workspaces work.

shaal:
True! We forgot to test restarting the workspace

rfay:
I'd like to close this for now and reopen it when it becomes actionable. So many things are not practical until they fix that bug as well as https://github.com/gitpod-io/gitpod/issues/2459 (so we can build docker images).

PR #2847, opened 2021-03-03
Use the correct variable name for Pantheon token

shaal:
## The Problem/Issue/Bug:
Documentation is different than the instructions in `.ddev/providers/pantheon.yaml.example`
I confirmed this syntax is working.


rfay (approved):
Thanks for testing, and thanks for the PR. I note that this is correct in pantheon.yaml.example

rfay:
@shaal if you could fix the markdownlint errors it would be appreciated, https://github.com/drud/ddev/pull/2847/checks?check_run_id=2024738642

PR #2858, opened 2021-03-08
Implement -d argument to ddev composer to specify relative dir, fixes #1699

### LINKED ISSUE #1699: ddev could have a config.yaml item to determine its expected composer.json location
I do not have the composer.json on the root level. I cannot use `ddev composer` then.

For convenience, I already have set the `config.yaml` setting `working_dir=` to the main web directory. I would expect that to influence composer's working directory as well.

```
$ ddev composer install
```
runs 
```
Failed to run docker-compose [-f /home/jonas/www/ratbacher.de/.ddev/docker-compose.yaml -f /home/jonas/www/ratbacher.de/.ddev/docker-compose.env.yaml -f /home/jonas/www/ratbacher.de/.ddev/docker-compose.redis.yaml exec -w /var/www/html -T web bash -c composer install], err='exit status 1', stdout='', stderr='Composer could not find a composer.json file in /var/www/html
```
which ignores the `working_dir=` - setting but instead uses explicitly `exec -w /var/www/html`

`ddev composer` should either have the possibility to use a `composer_root_dir=` setting or just use the already existing `working_dir=`. Both would make sense to me. The 1st implementation would not be breaking, though.

`ddev composer --working-dir=/var/www/html/webroot ...` works, but is tedious.

`ddev ssh`, then `composer` works also, but defeats the purpose of a `ddev composer` command and the according *-composer hooks.

Thank you for considering!

rfay:
## The Problem/Issue/Bug:

Despite the convention of having composer.json in the root of the project, some projects and some people have it elsewhere.

Add `ddev composer -d`  to identify the target directory

```
mkdir junk
ddev composer -d junk composer init
ddev composer -d junk require drush/drush
```
Verify the composer.json in the junk directory has reasonable content with drush.

Added to TestComposerCmd and TestComposer to cover this case

## Related Issue Link(s):

Fixes #1699

### DIFF
--- a/cmd/ddev/cmd/composer.go
+++ b/cmd/ddev/cmd/composer.go
@@ -8,6 +8,8 @@ import (
 	"os"
 )
 
+var execDir = ""
+
 var ComposerCmd = &cobra.Command{
 	Use:   "composer [command]",
 	Short: "Executes a composer command within the web container",
@@ -17,7 +19,8 @@ the command with 'ddev'.`,
 	Example: `ddev composer install
 ddev composer require <package>
 ddev composer outdated --minor-only
-ddev composer create drupal/recommended-project`,
+ddev composer create drupal/recommended-project
+ddev composer -d docroot install`,
 	Run: func(cmd *cobra.Command, args []string) {
 		app, err := ddevapp.GetActiveApp("")
 		if err != nil {
@@ -30,7 +33,7 @@ ddev composer create drupal/recommended-project`,
 			}
 		}
 
-		stdout, stderr, err := app.Composer(args)
+		stdout, stderr, err := app.Composer(execDir, args)
 		if err != nil {
 			util.Failed("composer %v failed, %v. stderr=%v", args, err, stderr)
 		}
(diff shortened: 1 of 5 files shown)


jonaseberle:
I am not sure I understand the testing instructions. We were using `-d` before already.

I would expect a testing regime like:
`mkdir junk`
`ddev config --working-dir=./junk`
`ddev composer init`
`[ -f junk/composer.json ] && echo OK`

rfay:
Hi @jonaseberle - thanks for testing! I guess you're right. This adds -d as a flag to *ddev composer*, you were adding -d as a flag to composer. Maybe this isn't useful.

Unfortunately I don't think it would be reasonable to change `ddev composer` to use the configured working_dir, because that already has meant from the beginning "the directory that ddev ssh and ddev exec land in". So it would have other unintended consequences both for you and others. 

AFAICT #1699 intended what's done here, but maybe I'm confused. With this, you can use `ddev composer -d <dir> require ...` etc, not just install.

Keep talking!

rfay:
Maybe this is a bad idea, and just using composer's -d flag is better.

jonaseberle:
It's not bad, but it doesn't give anything new.

We have the same behaviour already with `ddev composer -- --working-dir=./junk init` or `ddev composer -- -d ./junk init`..

Sorry that my intentions weren't clear.
I was after a `config.yaml` setting `composer_root_dir` which would be used by `ddev composer` by default. Actually just something like a `cd ...` before executing the composer command.

PR #2866, opened 2021-03-10
Update docker installation instructions.

oknate:
## The Problem/Issue/Bug:

See

https://stackoverflow.com/questions/30413621/homebrew-cask-option-not-recognized

as of 2021, command has changed.

It updates the command and it worked for me.

### DIFF
--- a/docs/users/docker_installation.md
+++ b/docs/users/docker_installation.md
@@ -1,6 +1,6 @@
 ## macOS Installation: Docker Desktop for Mac
 
-Docker Desktop for Mac can be installed via Homebrew (`brew cask install docker`) or can be downloaded from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg).
+Docker Desktop for Mac can be installed via Homebrew (`brew install --cask docker`) or can be downloaded from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg).
 
 ## Windows Installation: Docker Desktop for Windows
 


rfay (approved):
Thanks for this! They just changed this in homebrew 3.0.0, which came out a few weeks ago.

docs/users/docker_installation.md
```diff
@@ -1,6 +1,6 @@
 ## macOS Installation: Docker Desktop for Mac
 
-Docker Desktop for Mac can be installed via Homebrew (`brew cask install docker`) or can be downloaded from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg).
+Docker Desktop for Mac can be installed via Homebrew (`brew install --cask docker`) or can be downloaded from [download.docker.com](https://download.docker.com/mac/stable/Docker.dmg).
```
rfay, line 3: Actually, I don't think this is correct. The new usage to get docker desktop is `brew install homebrew/cask/docker` - Doing what you have here just gets the docker client.

rfay (changes requested):

docs/users/docker_installation.md
rfay, line 3:
```suggestion
Docker Desktop for Mac can be installed via Homebrew (`brew install  homebrew/cask/docker`) or can be downloaded from [desktop.docker.com](https://desktop.docker.com/mac/stable/Docker.dmg).
```

PR #2867, opened 2021-03-10
Touch up Windows WSL2 hosts file support and docs, rename DRUD_INTERACTIVE, fixes #2533

### LINKED ISSUE #2533: Non-ddev.site hostnames can't work on WSL2, and neither can ddev.site if DNS rebinding disallowed

rfay:
## The Problem/Issue/Bug:

#2533: In the current state of WSL2, there's quite a lot of confusion in the computer about how to look up hostnames. However, ddev already has warnings to users, and it rarely matters to users who use *.ddev.site.

* Improved docs, minor addition to code that warns WSL2 users
* Rename DRUD_INTERACTIVE  environment variable to DDEV_INTERACTIVE

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1976,10 +1976,10 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 // But if it's not, try to add one.
 func addHostEntry(name string, ip string) error {
 	_, err := osexec.LookPath("sudo")
-	if (os.Getenv("DRUD_NONINTERACTIVE") != "") || err != nil {
+	if (os.Getenv("DDEV_NONINTERACTIVE") != "") || err != nil {
 		util.Warning("You must manually add the following entry to your hosts file:\n%s %s\nOr with root/administrative privileges execute 'ddev hostname %s %s'", ip, name, name, ip)
-		if os.Getenv("WSL_DISTRO") != "" {
-			util.Warning("For WSL2, execute 'sudo ddev hostname %s %s' on Windows", name, ip)
+		if nodeps.IsWSL2() {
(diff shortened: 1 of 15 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -1988,8 +1988,8 @@ func addHostEntry(name string, ip string) error {
 	util.CheckErr(err)
 
 	output.UserOut.Printf("ddev needs to add an entry to your hostfile.\nIt will require administrative privileges via the sudo command, so you may be required\nto enter your password for sudo. ddev is about to issue the command:")
-	if os.Getenv("WSL_DISTRO") != "" {
-		output.UserOut.Printf("You are on WSL2, so should manually execute 'sudo ddev hostname %s %s' on Windows", name, ip)
+	if nodeps.IsWSL2() {
+		util.Warning("You are on WSL2, so should also manually execute 'sudo ddev hostname %s %s' on Windows if you use a Windows browser.", name, ip)
```
pointmetothedocs, line 1992: how does one execute sudo on windows?

PR #2869, opened 2021-03-13
Update acquia.md

KarimBoudjema:


rfay (approved):
Thanks! So it survived here huh? Good job. 

Thanks so much for testing!

PR #2879, opened 2021-03-17
Clarify caveat about snapshot compatibility

mootari:
## The Problem/Issue/Bug:
Snapshot docs mention the database type, but the example describes different database versions.

Adds mention of version constraints, per discussion in https://gitter.im/drud/ddev?at=60510519b5131f4f281d10ac


rfay (approved):
Thanks!

PR #2881, opened 2021-03-17
Update acquia.yaml.example

KarimBoudjema:
Replace 'eeamoreno.dev' with the project_id variable. So user only has to make changes in one place.


rfay (commented):
This is correct. I stumbled on it today. However, if you don't mind I'll make this change in the work I'm currently doing.

KarimBoudjema:
No problem @rfay

rfay:
I checked carefully just now that I have both of your changes. 

Beware that I'm changing the acquia and pantheon recipes a bit as I'm working on "push".

Thanks so much for the testing and for this contribution!

PR #2888, opened 2021-03-24
Update brew upgrade command

moritzkrauss:
## The Problem/Issue/Bug:
Tried to update ddev with `ddev poweroff && brew upgrade ddev`to latest version and got the following error:

``` 
Formulae found in multiple taps: 
 * drud/ddev/ddev
 * drud/ddev-edge/ddev
```

Updated the existing upgrade command with `ddev poweroff && brew upgrade drud/ddev/ddev` to match correct package/tap.


rfay (approved):
Thanks for this. I think the fully-qualified package name is the least confusing in the long run. People who know brew know what to do, but lots of people just use it casually and occasionally. 

I searched through for other usage changes and added a commit or two.

PR #2892, opened 2021-03-24
[docs only] Use fully-qualified name for service when configuring in client, fixes #2744 [skip ci][ci skip]

### LINKED ISSUE #2744: Services are accessible by other projects
**Describe the bug**
If one adds a Service to ddev like for example solr like described here: https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/typo3-solr, the service is also available in other ddev projects and probably misused by accident.

**To Reproduce**
- Setup solr service like described in the link above. 
- create or switch to an other (simular) project and use that solr core
- even if the second project has its own service (setup the same way) its hard to guess which service container is taken.

**Expected behavior**
A service container should by default somehow only be accessible by the same container stack.

**Hint**
Its of cause possible to use different ports for each solr service in each project, but thats not very practical because:
- one has to know each already used ports
- its easy to accidently take a port already used from something else on the host

- Host computer OS and Version: Linux
- ddev version information v1.16.5

rfay:
## The Problem/Issue/Bug:

#2744: It's possible to use ambiguous service names in client configuration. This changes the docs to suggest using names like `ddev-<projectname>-solr` instead of the default (and sometimes ambiguous) `solr`

@tgaertner I'd definitely appreciate your review of this, and would appreciate if you could re-test. I think this will fix the problem you described so clearly.

### DIFF
--- a/docs/users/extend/additional-services.md
+++ b/docs/users/extend/additional-services.md
@@ -23,7 +23,7 @@ This recipe adds an Apache Solr container to a project. It will set up a solr co
 * Add a solr server at `https://<projectname>>.ddev.site/en/admin/config/search/search-api/add-server`.
     * Use the "standard" Solr connector
     * Use the "http" protocol
-    * The "solr host" should be "ddev-<projectname>-solr" **NOT the default "localhost"**, because it does not run in the same container as the webserver.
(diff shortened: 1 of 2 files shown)


tgaertner:
I would like to see even a hint or warning why its not save to use "solr" as solr host.
"solr" works, but bears the risk of cross connecting ddev web containers to solr containers of other projects if running multiple projects at the same time.

rfay:
Added the suggestion, thanks!

PR #2897, opened 2021-03-26
Latest docker (version 20)

shaal:
## The Problem/Issue/Bug:
By running this command - we get docker version 20 installed in Gitpod.
Does it help resolve - https://github.com/gitpod-io/gitpod/issues/3051#issuecomment-789246763?

Update gitpod to use docker version 20.


rfay (approved):
Well that's mighty exciting! It works perfectly, can build images now. Thanks for catching this. It's the normal way to upgrade of course, but I thought they had to be in control of it. 

I guess we should do this for ddev-gitpod as well.

shaal:
Gitpod still has that issue open [[docker] Make Docker 20.10.2 work #3051](https://github.com/gitpod-io/gitpod/issues/3051)
Is it possible that other parts are not working?

rfay:
Sure worked perfectly for me. It may be their docker vs our docker.

rfay:
Unfortunately... I hadn't tested a simple `make`:
```
gitpod /workspace/ddev $ make
building .gotmp/bin/linux_amd64/ddev from ./cmd/... ./pkg/...
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: write sysctl key net.ipv4.ip_unprivileged_port_start: open /proc/sys/net/ipv4/ip_unprivileged_port_start: read-only file system: unknown.
make: *** [Makefile:115: .gotmp/bin/linux_amd64/ddev] Error 125
```

It's great that with 20.10 you can build images... but you can't build ddev. Catch 22. I should have tested it.

PR #2921, opened 2021-04-02
Update mkcert-in-Firefox docs;  Firefox tab naming changed

ijf8090:
Firefox tab names and selections have changed slightly


rfay (approved):
I took a quick look at Firefox on both Windows and Mac, and this update seems correct to me.

rfay:
Thanks so much for catching this!

PR #2933, opened 2021-04-07
gitpod: Move make into init in .gitpod.yml

rfay:
## The Problem/Issue/Bug:

The gitpod check is likely not running because there is no init.

With the latest gitpod as of yesterday, there's no need to delay the make and ddev start

Add an init that does basic make and setup.

See also #2935


rfay:
This definitely isn't perfect yet (the add-on junk project doesn't work right in gitpod), but I think it's worth pulling because it gets the gitpod check working.

PR #2935, opened 2021-04-07
proposed Gitpod pre-build config

JanKoehnlein:
I played around a bit with your Gitpod setup. This is the result
- pre-install `markdownlint` in the dockerfile as it's required by the build
- run `docker-up` as `before` task as it's needed in all scenarios. Save its PID in a tempfile
- run the `make build` as an `init` task
- add a `prebuild` task that waits for `make` to finish and tears down docker using the saved PID, otherwise pre-builds won't terminate


rfay (commented):
Thanks so much for this. I've been experimenting with several of your ideas in https://github.com/drud/ddev/pull/2933, which had been opened before you did this one. I really appreciate your time and tutoring!

rfay:
I think the key ideas got into #2933, thanks! Looking forward to more collaboration.

PR #2946, opened 2021-04-12
Use terminus to invoke drush to pull DB from Pantheon

amitaibu:
## The Problem/Issue/Bug:

With the existing comman to pull db from Pantheon, I got an error:

`[error]  Could not locate a site your user may access identified by @foo.bar `

Trying to move to use `terminus drush` instead

### DIFF
--- a/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
+++ b/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
@@ -51,7 +51,7 @@ db_pull_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
-    terminus remote:drush -- ${project} sql-dump --gzip >/var/www/html/.ddev/.downloads/db.sql.gz
+    terminus drush ${project} -- sql-dump --gzip >/var/www/html/.ddev/.downloads/db.sql.gz
 
 files_pull_command:
   command: |
--- a/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
+++ b/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
@@ -51,7 +51,7 @@ db_pull_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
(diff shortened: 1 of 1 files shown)


amitaibu:
Actually, I was too quick on the submit. It recognizes the site, but I got a different error

```
 [notice] Command: foo.bar -- drush sql-dump [Exit: 0]
Database downloaded to: /home/amitaibu/Sites/foo/drupal/.ddev/.downloads/db.sql.gz
Importing database...
Pull failed: failed to extract provided archive: gzip: invalid header
```

amitaibu:
Ok, some understanding I have about current implementation:

Right now in `master` upon authentication it does `terminus aliases`. However when using `drush @${project}` it would expect `pantheon.foo.bar` and not `foo.bar` as the current example suggests. So I do think that removing the `terminus aliases` and switching to `terminus drush` might be a better option.

I'm not clear about the `gzip: invalid header` error though.

rfay:
Usually gzip invalid header means that the data isn't gzipped. And you're doing a sql-dump without --gzip.

amitaibu:
btw, another advantage of using `terminus drush` is that it recognizes sites that I belong to through "Organization", but not a member of specifically.

> And you're doing a sql-dump without --gzip.

It's actually there, but for some reason it's not working. `-- sql-dump --gzip >...`

rfay:
You can do the gzipping on the local side then...

amitaibu:
Haven't circled back to it yet. I'll close it for now, until I find a working solution.

rfay:
Thanks, sure appreciate you working on it. I'm not having any trouble with the existing one, so if I can help with the problem you're having there, let me know. I'm definitely not a terminus or pantheon expert.

rfay:
There was a version of drush that didn't properly set the site alias location, so you might want to add `--alias-path=~/.drush` to the drush command...

rfay:
Would love to have you try and review the new strategy (backup-based) in https://github.com/drud/ddev/pull/2957 which is far, far better for a large site.  Thanks @amitaibu .

amitaibu:
Sure, I'll check it (Probably in a couple of days).
Using `terminus` looks like a great approach. Should I install ddev with that new branch? Is there a README on how to do that (I currently use brew for installation)

rfay:
All that changes here is the example file, so you can just take the "pull" lines from the pantheon.yaml.example and put them in your pantheon.yaml. You don't have to change ddev at all unless you don't have v1.17+, which I'm sure you do.

PR #2949, opened 2021-04-15
Add ~/bin and /var/www/html/bin to PATH in web container

rfay:
## The Problem/Issue/Bug:

A number of people have their composer bin directory in /bin instead of /vendor/bin - this is fairly nonstandard, but there's no reason not to add that to the default path. 

Add /var/www/html/bin to path

Also adds ~/bin to front of $PATH, allows people to solve problems like this by just putting whatever they need in global homeadditions.

`ddev exec 'echo $PATH'`

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
@@ -1,1 +1,1 @@
-export PATH="$PATH:/var/www/html/vendor/bin"
+export PATH="$PATH:/var/www/html/vendor/bin:/var/www/html/bin"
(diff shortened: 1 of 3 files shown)


mootari:
The addition of `~/bin` causes problems with `find -execdir` due to the relative path:
> find: The relative path ‘~/bin’ is included in the PATH environment variable, which is insecure in combination with the -execdir action of find. Please remove that entry from $PATH

Is it possible to use `$HOME` instead?
~~~
export PATH="$HOME/bin:$PATH:/var/www/html/vendor/bin:/var/www/html/bin"
~~~

rfay:
I had never heard of `-execdir` before, and I would think it would be pretty bad anyway. Why were you using it? What are you trying to accomplish?

rfay:
I'm sure you know that you can set the PATH yourself before doing a find (and -execdir is a pretty strange thing to use, people often use -exec, but not so much -execdir). `PATH=/some/path1:/some/path2 find /someotherpath -name=something -exec-dir ...`

PR #2957, opened 2021-04-21
Change pantheon pull to use terminus backup:get

rfay:
## The Problem/Issue/Bug:

The previous (drush rsync) technique for pulling a backup from Pantheon was just way too tedious, and could even fail due to connection timeouts

Use `terminus backup:get`. command and require that a backup already exists.

Use the "pull" sections of the new strategy with your site and see if it works.


amitaibu:
@rfay Worked great! (I only wish there was a progress report in terminus for the files get -which in my case took quite a long time)

Two suggestions:

1. Since `db_push_command` and `files_push_command` are dangerous, how about commenting them out? Let's force user for one more step, before they accidentally destroy their env.
2. Since `files_push_command` is now the only command using drush directly, I think we can remove the `terminus aliases 2>/dev/null` from the `auth_command` and move it to be under `files_push_command`. Like that we'd save a bit of time on each authentication.

rfay:
I agree with you on both.

It was actually DDEV (the company) that kept pushing for "push" for integration with DDEV-Live, and I probably wouldn't have done it anyway. 

The time for `terminus aliases` is irrelevant compared to the time for pull though.

PR #2963, opened 2021-04-24
Comment out dangerous commands, and move terminus alias

amitaibu:
## The Problem/Issue/Bug:

Comment out dangerous commands in Pantheon integration.

## Related Issue Link(s):

https://github.com/drud/ddev/pull/2957#issuecomment-826106788

### DIFF
--- a/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
+++ b/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
@@ -32,7 +32,7 @@
 #
 # 10. Run `ddev pull pantheon`. The ddev environment  download the Pantheon database and files, and import the database and files into the ddev environment. You should now be able to access the project locally.
 #
-# 11. Optionally use `ddev push pantheon` to push local files and database to DDEV-Live. Note that `ddev push` is a command that can potentially damage your production site, so this is not recommended.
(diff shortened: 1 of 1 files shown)


cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example
```diff
@@ -64,18 +63,19 @@ files_pull_command:
...
+#    pushd /var/www/html/.ddev/.downloads >/dev/null;
+#    terminus remote:drush ${project} -- sql-drop -y
+#    gzip -dc db.sql.gz | terminus remote:drush ${project} -- sql-cli
+#
+## push is a dangerous command. If not absolutely needed it's better to delete these lines.
+#files_push_command:
+#  command: |
+#    # set -x   # You can enable bash debugging output by uncommenting
+#    ls ${DDEV_FILES_DIR} >/dev/null # This just refreshes stale NFS if possible
+#    terminus aliases 2>/dev/null
```
amitaibu, line 80: `terminus aliases` moved here

rfay (changes requested):
* This approach will break the push tests, so we probably need to do something like adding an environment variable like "DDEV_ALLOW_PUSH=true" instead of the commenting-out technique you've done.
* We need to do it with all the example providers, not just this one.

rfay:
This one is superceded by #2966 right @amitaibu ? Did you get everything you wanted in there and we should close this one?

amitaibu:
Yes, this one can be closed. To be honest, I never got back to https://github.com/drud/ddev/pull/2966 , sorry :smiley_cat:

rfay:
#2966 has excellent potential. One of us will get back to it :)

Issue #2195, opened 2020-04-17
Update & Redesign ddev describe output (and ddev list and ddev version)


andy-blum:
The `ddev describe` output is a bit cluttered and has some pieces missing now. Setting this issue up as meta/discussion to encourage community input.

Below is the current output for a drupal8 project 'example.com'

```
NAME         TYPE     LOCATION             URL                            STATUS
example.com  drupal8  ~/Sites/example.com  https://example.com.ddev.site  running

Project Information
-------------------
PHP version:      	7.3
NFS mount enabled:	true
Database type:    	mariadb
MariaDB version:  	10.2

URLs
----
https://example.com.ddev.site
https://127.0.0.1:32791
http://example.com.ddev.site
http://127.0.0.1:32792

MySQL/MariaDB Credentials
-------------------------
Username: "db", Password: "db", Default database: "db"

or use root credentials when needed: Username: "root", Password: "root"

Database hostname and port INSIDE container: db:3306
To connect to db server inside container or in project settings files:
mysql --host=db --user=db --password=db --database=db
Database hostname and port from HOST: 127.0.0.1:32790
To connect to mysql from your host machine,
mysql --host=127.0.0.1 --port=32790 --user=db --password=db --database=db

Other Services
--------------
MailHog (https):	https://example.com.ddev.site:8026
MailHog:        	http://example.com.ddev.site:8025

DDEV ROUTER STATUS: healthy
ssh-auth status: healthy
```

andy-blum:
[My initial complaint](https://github.com/drud/ddev/issues/2174#issuecomment-615333930) is regarding the output of the database credentials. I suggest:

```
MySQL/MariaDB Credentials
-------------------------

Username (root):  root
Password (root):  root
Username:         db
Password:         db

host:port (internal):  db:3306
host:port (external): 127.0.0.1:XXXXX
```

andy-blum:
I think another immediate improvement would be to transfer the first two lines into the same table format

```
example.com
-------------------------------------------------
Type:               drupal8
Location:           ~/Sites/example.com
URL:                https://example.com.ddev.site
Status:             Running
-------------------------------------------------
```

andy-blum:
I think it might also be worth splitting the container/server info to sit along side other relevant information, giving us 5 main sections:

- Basic Info
- Webserver Info
- Database Info
- Services/Settings/Tools/Misc
- Global DDEV status info (I don't ever use this info and wouldn't be sad to see it gone)

```
example.com
[...]

PR #2965, opened 2021-04-26
Update & redesign ddev describe output #2195

andrewbelcher:
## The Problem/Issue/Bug:
Redesigns `ddev describe` output

## Related Issue Link(s):
https://github.com/drud/ddev/issues/2195


andrewbelcher:
First bash at using go, so I would not be surprised if I've done things in improper ways - please do point things out!

So things I _know_ are not even attempted yet:

- [ ] Updating any relevant tests
- [ ] Need to update gosuri/uitable for fix to column width with colors - this was fixed in https://github.com/gosuri/uitable/issues/14 is in v0.0.4, but I couldn't figure out the combo of `go get`/`go mod`/`make` that it would like for updating!
- [ ] Need statuses for web/db, maybe MailHog?
- [ ] Probably a better way to handle versions rather than using image name
- [ ] Label/Host differentiation for extra services?
- [ ] Skipping extra hosts that have no additional info
- [ ] Ability to add customised additional info for extra services?

Probably a host of other things to...

rfay:
Bravo!  

Some things that might be fun to experiment with:
* [gchalk](https://github.com/jwalton/gchalk/blob/master/README.md) apparently is a go module for easily adding colored/styles cli output.
* Anybody can test this PR as you work on it at https://gitpod.io/#https://github.com/drud/ddev/pull/2965 by just typing `ddev describe` .

cmd/ddev/cmd/describe.go
```diff
@@ -133,4 +258,23 @@ func renderAppDescribe(desc map[string]interface{}) (string, error) {
 
 func init() {
 	RootCmd.AddCommand(DescribeCommand)
+	DescribeCommand.Flags().StringVarP(&service, "service", "s", "", "The service for additional information")
+	DescribeCommand.Flags().BoolVarP(&verbose, "verbose", "v", false, "Show extended output for all services")
+}
+
+func formatStatus(status string) string {
```
rfay, line 265: I think gchalk would be easier than this and far more powerful.

andrewbelcher: Can you give me a quick summary of how to manage dependencies for DDev? I tried updating `gosuri/uitable` using `go get -d github.com/gosuri/uitable@v0.0.4`, which appeared to work, but them `make` started failing because of explicit dependencies in `vendor/modules.txt`. I found references to that with regard to `go mod vendor`, but running that failed due to `embed: package embed is not in GOROOT (/usr/local/go/src/embed)` and I didn't want to go further down the rabbit hole!

rfay: What I do is the `go get` and then `go mod vendor && go mod tidy`

andrewbelcher: Updating go seemed to fix my issue! So got uitable updated now.

What's the reason for the gchalk switch? Whilst it looks more powerful, nothing I've done so far (or am expecting to do) adds to what was already being done using `fatih/color`, so swapping seems a bit of tangent?

pkg/ddevapp/ddevapp.go
```diff
@@ -317,9 +317,12 @@ func (app *DdevApp) Describe(short bool) (map[string]interface{}, error) {
 				if svc, ok = v.(map[interface{}]interface{}); !ok {
 					continue
 				}
-
 				extraServices[serviceName] = map[string]string{}
 
+				extraServices[serviceName]["version"] = svc["image"].(string)
+
+				extraServices[serviceName]["status"], _ = dockerutil.GetContainerStateByName(svc["container_name"].(string))
+
```
andrewbelcher, line 325: @rfay My new table also has the status for non-extra services. How would you like me to handle adding that information? Perhaps we should have a separate key for container statuses and set it before skipping the default containers?

rfay: That sounds reasonable.

cmd/ddev/cmd/describe.go
rfay: I was just thinking you'd have a lot of fun with it :) Might give more flexibility than just colors. It's been on my list to try.

Yes, go 1.16+ is required.

rfay:
Rebased and fixed gofmt problem.

rfay:
Just a note that this is very much appreciated and I'll be getting to it!

rfay:
Rebased

rfay:
See https://github.com/drud/ddev/issues/3054 - `webserver_type` needs to be added to `ddev describe`

tyler36:
I really like the services table; it's so clean and consice.

IMHO, I would prefer the current project overview
```
NAME        TYPE  LOCATION               URL                                STATUS 
simpleproj  php   /workspace/simpleproj  https://simpleproj.ddev.site:8443  running        
````
... instead of the expanded one from this PR
```
Project Information
-------------------
Name      simpleproj                       
Type      php                              
Location  /workspace/simpleproj            
URL       https://simpleproj.ddev.site:8443
Status    running         
```

- I work in VSCode and the current project overview with the PR services tables would allow for the most information in the small terminal without scrolling up.
- I also like the idea that it matches closely to the `ddev list` output. `list` shows the overview, `describe` shows the same overview but with additional services.

My personal vote would be for something like:
```
gitpod /workspace/simpleproj $ ddev list
NAME        TYPE  LOCATION               URL                                STATUS 
simpleproj  php   /workspace/simpleproj  https://simpleproj.ddev.site:8443  running


gitpod /workspace/simpleproj $ ddev describe
NAME        TYPE  LOCATION               URL                                STATUS 
simpleproj  php   /workspace/simpleproj  https://simpleproj.ddev.site:8443  running                

Services
-------------------
SERVICE    | HOST | STATE | URL/PORT                          | INFO              
Web        | web  | @todo | https://simpleproj.ddev.site:8443 | PHP 7.4, nginx-fpm
Database   | db   | @todo | 127.0.0.1:32768                   | MariaDB 10.3      
MailHog    |      | @todo | https://simpleproj.ddev.site:8026 |                   
phpMyAdmin |      | @todo | https://simpleproj.ddev.site:8037 |        

````

I think this best of both worlds keeps the details without the noise; add some color on the status and links/paths and would be perfect.

**Comment by rfay** [comment]
Here's what it's looking like right now:

Cursor_and_c8819936_pr3-hrsitesp-lapp114__ngs_app_hrsitesp_and_rfay_d9-web___var_www_html

Unfortunately, that requires a window 104 characters wide (because of mutagen status)

Suggestions welcome.

You can experiment with this yourself at https://gitpod.io/#https://github.com/drud/ddev/pull/2965

**Comment by tyler36** [comment]
Just had a look at your latest Gitpod

- 'mutegen' is spelt lower case but your previous screen has it upper.
- 'Mutegen enabled (Watching for changes)' ... I think reducing this to just  'Mutegen enabled (watching)' still conveys the meaning but without the extra characters
 - 'Host' & 'Inside' ... I'm not sure why but 'inside' feels off. Originally, I was think 'SSH' instead since this is th prefered way of getting inside the container (even though currently it's not a *true* ssh session). This might help tie the comment / command connection. Not sure how strongly I feel about it though.
- Final though, would 'disabled/enabled' states be better communicated through a character / icon / emoji? 
```
NFS X
Mutagen O

X NFS
O Mutagen

❌ NFS
✔️ Mutagen

✖NFS
✅ Mutagen
```

DDEV-describe

**Comment by rfay** [comment]
Thanks so much for the quick review @tyler36 !

* I don't agree that Mutagen must be lower case...

Mutagen___Cloud-based_development_using_your_local_tools

* The "Watching for changes" is something that comes out of Mutagen, and it will be one of the many statuses of Mutagen. It wouldn't be reasonable to try to change it, although I'd love a suggestion of how to present it in a more compact way.
* "Host" and "Inside" - I struggle with explaining this also, but I find that new folks are often confused by this difference, so want to make sure to present this info somehow.
* I like the emoji idea, might be more compact. Pretty sure I don't know how to do it in a terminal environment in a cross-platform situation. It needs to work right on Windows, Linux, WSL2, macOS, and in a number of different terminal clients.

Absolutely looking for ways to communicate more concisely.

**Comment by tyler36** [comment]
>I don't agree that Mutagen must be lower case...

My bad ... I was pointing out the inconsistency. Yes, capitalized is prefered

>  "Watching for changes"

Might be best to stick to "official" messages from Mutagen, would make debugging easier for new comers.

> "Host" and "Inside" - I struggle with explaining this also, but I find that new folks are often confused by this difference, so want to make sure to present this info somehow.

Yeah, that why I though "SSH" might be good alternative since one would `ddev ssh` then run commands with appropriate SSH IP.  Even more so if DDEV ever intergrates DDEV-contrib SSH or #3097 . This would help drive home a SSH environment concept.

>  I like the emoji idea

I like the emojis too and there add a hint of color. I've been using [Starship](https://starship.rs/guide) which is a cross-platform shell prompt that uses emojis. If you went down that road, 
- using `enabled / disabled` is redundant
- You could use a `🐳` for `inside`, maybe a `🌐` for host/web services?

```
Service    | Hostname            | Status  | URL/Port                          | Info             
Web        | ddev-simpleproj-web | running | 🌐: https://simpleproj.ddev.site:8443 | PHP 7.4 nginx-fpm
           |                     |         | 🐳: http://localhost              | ✔️ NFS 
           |                     |         |                                    | ❌ Mutagen
Database   | ddev-simpleproj-db  | running | 🌐: 127.0.0.1:32768              | MariaDB 10.3     
           |                     |         | 🐳: ddev-simpleproj-db:3306      |                  
MailHog    |                     |         | https://simpleproj.ddev.site:8026 |                  
           |                     |         | SMTP: web:1025                    |                  
phpMyAdmin | dba                 | running | https://simpleproj.ddev.site:8037 |       

I tried updated the `describe.go` file in gitpod but didn't know how to get it show.

rfay:
Thanks for the pointer to starship. I'd say that since it requires a custom font to be installed it's a non-starter though. 

Mutagen vs mutagen is like DDEV vs ddev. It's hard, and there are lots of inconsistencies. My goal is to use "Mutagen" when talking about the product, and "mutagen" when talking about the command. Same with "DDEV" and "ddev". But I certainly am not completely consistent.

"ssh" really isn't the right description of the ports and such inside the container. You normally don't need to know about those when doing `ddev ssh`, you need them when configuring things like `docker-compose.*.yaml` or just understanding the whole topology. Lots of people don't even understand that there's another whole set of computers running inside that docker/ddev thingie.

My biggest problem with what's there right now is I wish it weren't taking two lines in those rows to get all the info across.

And it could use a lot more color.

tyler36:
I might be wrong here but I though Starship used a mixture of nerd fonts and emojis. So if you don't have nerd fonts installed, you don't get the `node`  () or `git branch` ( ) icons but the `php` (🐘) and `time` (⏳ clock) emoji show. 

Personally, except for the brands, I use emojis where possible because they are more colorful. See example below without / with nerd fonts

Win10 Windows Terminal WSL2: Cascadia Code
cascadiacode

Win10 Windows Terminal WSL2: Cascadia Code with Nerd Font
cascadiacode-nf

Since emoji are now [unicode blocks](https://en.wikipedia.org/wiki/Emoji#Unicode_blocks), they are supported since Win8, Mac OS, Ubuntu 18.04 and Fedora 28. 

Ideally, you'd have config settings with *sensible* defaults:
```
{
 termDisabled:  'X';
 termEnabled = 'O';
 termHost = 'Host';
 termInside = 'Inside';
}
```
then people could customize/theme it as a DDEV-contrib style project.

```
{
 termDisabled:  '❌';
 termEnabled = '✔️';
 termHost = '🌐';
 termInside = '🐳';
}
```

rfay:
I sure like the idea of emojis. 

I tried pasting "❌🌐🐳✔️" into lots of places. 

* Good on macOS iterm2 and terminal
* Good on Ubuntu 20.04
* Black-and white on Windows git-bash window (and interpreted funny)
* Just little boxes on WSL2 Ubuntu 20.04
* Black-and-white on Windows cmder running git-bash
* Little boxes on Windows PowerShell

I sure love the idea. Not sure the world is ready for it.

I have always loved what `mkcert -install` does with emojis, but didn't know the limitations, and never noticed that it fails all these places too.

Here's what it looks like most places:

Ubuntu_20_04_ARM64

But it's little ugly boxes or grayscale emojis in the same places as above.

tyler36:
That's strange ... I wonder why it's working in my Powershell & WSL2 then? (obviously some dark magic spun in the middle of the nite)

Thank you for the follow-through ... it seems the world isn't ready for an emoji-fied DDEV just yet. :(

rfay:
@andrewbelcher I'd love it if you'd weigh in on this. Is this going to fly? Is it a step forward?

rfay:
I apologize for having to rebase this into a single commit in order to solve massive vendor directory conflicts with latest upstream. I think it's still the same and behaves the same though. Seems to be OK.

andrewbelcher:
image

I think overall this is a good step forward in improving this! Thanks for your work on it @rfay and your feedback @tyler36! I'd personally be happy for it to go in pretty much as is and future iterations can improve further...

There's some issue with the alignment of the table headers. This appears to be to do with colourisation, which I thought we'd solved... I think this should be sorted before it goes in.

I personally prefer a vertical table to an inline one for the summary. I find it much easier to read, though I definitely appreciate the concern about it taking up more space!

Do we need both Hostname (which appears to be mix of container name & internal hostname) and "Inside" URL/Port? Perhaps we could replace Hostname with the internal addresses (which is what I think I had in mind initially). Then maybe we could get down to one line per service? Something like:

```
Service    | Status  | Host address                   | Internal address | Info
Web        | running | https://example.ddev.site      | web              | PHP 7.4 nginx-fpm
Database   | running | 127.0.0.1:32771                | db:3306          | MariaDB 10.3
MailHog    | running | https://example.ddev.site:8026 | web:1025 (SMTP)  |
phpMyAdmin | running | https://example.ddev.site:8037 |                  |
solr       | running | https://example.ddev.site:8983 | solr:8983        | solr:7.5
```

I think if the services are going to end up with multiple lines on anything other than Info, there needs to be some kind of separator between the services. Unfortunately that adds quite a bit of height, but I think it might be necessary for readability? 

Do _NFS disabled_ / _mutagen disabled_ need to go in web info? Are they not 'global' (although primarily web)? If they are helpful enough to go in the non verbose version, perhaps they could go outside of the table?

Definitely a potential for follow up would be how to declare the information for additional services, e.g. SOLR, node for frontend stuff etc that people might add.

I also agree, colour & emoji could be helpful, though I would probably put them again in the category of suitable to be follow ups.

tyler36:
I agree with @andrewbelcher in that this is a great improvement over the current offering and should be merge sooner rather than later. This will probably cause further feedback from other which might bring to light some issues or use cases we haven't thought about.

rfay:
Thanks. I expect to spend a little more time with it and get it into the next alpha.

rfay:
OK, spent a lot of time on this and hope you like it. You can try it out at https://gitpod.io/#https://github.com/drud/ddev/issues/2195 or download the binaries and run it in your own context. (I'll add a link to the artifacts in a minute)

rfay_covid19-web___var_www_html

tyler36:
Some quick feedback based on #issuecomment-902296653

- When I opened your [gitpod link](https://gitpod.io/#https://github.com/drud/ddev/pull/2965), my colors were off compared to your screenshot. The dark grey background is hard to read.
ddev-gitpod-details

 
  - Colors seemed fine in my terminal though on native Win10 (not WSL). I was about to try WSL but you made some changes
ddev-win10-terminal


- Personally, I find the default view too long. I generally keep my terminal window small (VScode default-ish) and `ddev status` & `ddev describe` overflow and I'm left with a links only view
ddev-gitpod-overflow
 
Perhaps something like
- `ddev status` : table view only
- `ddev describe` / `ddev status --info` : table with more info

- The URL on the `InDocker: ddev-...`  lines are not recognized as a link in gitpod or my terminal (ie. I can't click through). I guess this is because they are missing `http` because line length. Should they be made clickable? 


UPDATE: Just checked your [last update](https://github.com/drud/ddev/pull/2965#commits-pushed-c6b3a9e). Colors still appear off in Chrome 92, & Firefox 91.0.1

rfay:
Thanks for taking a look @tyler36 - It's. good to have different views on different systems. 

It's super easy to adjust the colors, this is done with https://github.com/jedib0t/go-pretty and we can find a custom style that we like. As you can see there I ended up using `table.StyleColoredBright`, and we could easily do something else.

tyler36:
Yeah, the current colors on Win10 Chrome 92 & Win10 Firefox 91 on gitpod website are *not* usuable. It looks fine in my terminal though (I really like it), so not sure if Gitpod is doing something funky.

rfay:
I'm worried about usability problems as well (especially colorblindness) so added an environment variable DDEV_USE_SIMPLE_FORMATTING=true to go to plain text formatting.

Definitely want to avoid unusable combinations. 

If you want, it's easy to fiddle with the styles in gitpod and change the color combinations.

rfay:
It's horrible on gitpod, I agree.

This is what it looks like on untouched Ubuntu:

Ubuntu_20_04_ARM64

andrewbelcher:
Love it as an approach, but agree the colours need some tweaking. I think the unformatted version is a good idea too. How does all the colourisation affect piping, e.g. Into grep?

rfay:
Colorization makes a mess for piping, but that can be fixed and should be (detect is stdout is a. terminal, not just stdin)

For future reference though, `ddev describe -j` with jq is a lot better than grepping plain describe output.

rfay:
OK, it's over the top and not what I intended to do, but I got worried about what would happen in various environments, so we have complete style selection for the list and describe tables. I hope you'll try them out, including on gitpod, https://gitpod.io/#https://github.com/drud/ddev/pull/2965

`ddev config global --table-style=something` will show you the options. 

To try it out on your own computer, I'll update this in a minute with the artifacts from this build.

rfay:
Final TODO:
- [ ] Choose a default theme
- [ ] Make a separate followon table in describe to show the various details. It can just be a new table following the other one.
- [ ] In `ddev list` make the router status a separate followon container. Make sure it looks good when router is stopped.

rfay:
Here's the current very textual look with the default table theme and many services:

rfay_ddev-ssh-agent____and_root_d9-web___var_www_html

tyler36:
## Windows Terminal on Win10

- Like the simple text-based table
- `ddev describe` is broken when in a project that is stopped:
    - `1.17.7`: it displays name, type, ... status etc.
    - Now, it shows the "More info:" details.
- TIL `ddev describe -j`
- You went all out on theming there! Cool selection & variety
- I think "ROUTER" looks out of place on the colored themes.
  - Maybe make it the same color style as the header?
  - Or, just make it a part of the table. Currently, "ROUTER" with a "HEALTHY" type doesn't make sense.
ddev-router

- Will the website have screenshots for all available themes?
- Do you anticipate / how will you handle requests for user contributed themes?

rfay:
TODO:
- [ ] Clean up "servicename" and "verbose" flags, at least for now, cmd/describe.go
- [ ] Clean up usability questions in cli_usage.md, is that even needed any more? Remove the $DDEV_USE_SIMPLE_FORMATTING?, see also pkg/ddevapp/config.go, nodeps.UseSimpleFormatting
- [ ] Remove table separators if output not a tty
- [ ] Make sure `ddev list` output is OK when router has problems - fit it into correct cell
- [ ] Use util.ColorizeText("color") instead of direct use of text color
- [ ] Mutagen status should show in `ddev list` and `ddev describe`
- [ ] global_config.go instructions about table_style are hanging.
- [ ] RequireSimpleFormatting() use?
- [ ] Make sure json output still works.
- [ ] Review version output. It can probably go ahead and show expected mutagen version.
- [ ] Describe should handle the new elements added
- [ ] Describe should use IsRouterDisabled() to determine whether to output the main stuff.

rfay:
@tyler36 thanks for the detailed review!

I'm kind of thinking the huge number of themes is just too much and most could be commented out. Choice is a sickness of the mind. Just allow 3 maybe?

tyler36:
@rfay I'd comment them out. No point destroying work you've done, it might help out others who want to theme and it's a nice "easter egg" for people that poke through code.

rfay:
OK, for those who can still scroll in this one! Hopefully we'll be done now.

Cursor_and_rfay_d9-web___var_www_html

rfay_d9-web___var_www_html

rfay_d9-web___var_www_html_and_rfay_d9-web___usr_local_bin_and_rfay_d9-web___var_www_html

rfay_d9-web___var_www_html_and_rfay_d9-web___usr_local_bin_and_rfay_d9-web___var_www_html

rfay:
Artifacts for local testing will shortly be at https://github.com/drud/ddev/actions/runs/1184327314, to test with gitpod, https://gitpod.io/#https://github.com/drud/ddev/pull/2965

tyler36:
There is an issue with the default table setting, at least in the [gipod ](https://gitpod.io/#https://github.com/drud/ddev/pull/2965).

In the screen shot below, you can see
- `ddev list` shows an unstyled stream of consciousness
- setting the table to `bright`, shows the table (but with the ugly gitpod colors 😢 , waiting for artifacts for local testing)
- setting the table to `default` results in a proper table display
ddev-gitpod-list

- Tables (default & bright) show nicely when project is stopped

Still not a fan of "HEALTHY" & "STOPPED" for "ROUTER" listed under "LOCATION" but 🤷

rfay:
Thanks - looks like an empty (initial value) of table-style doesn't equal "default", which I thought it would. Best to avoid this for users, thanks for pointing it out. 

Gitpod has terrible colors, if you want to explore colors that might work for gitpod, I'm happy to enable another style. But that's the reason that the uncolored version is now the default.

I tried a number of other things for the router status (and moved the failure status below the whole thing), but never came up with anything I liked better. Suggestions still welcome. I wasn't able to concatenate two tables and have them the same width, which would have been the preferable solution.

rfay:
New artifacts at https://github.com/drud/ddev/actions/runs/1184441145 gitpod still https://gitpod.io/#https://github.com/drud/ddev/pull/2965

Thanks for catching the empty global config situation.

tyler36:
Hmmm .... table don't seem to work on Win10.

- I downloaded the [artifact](https://github.com/drud/ddev/actions/runs/1184441145)
- removed old version via `choco uninstall ddev`
- installed artifact
```bash
$ ddev --version
ddev version v1.18.0-alpha4-85-g9012eb52
```
- moved into a project with a DDEV config and ran `ddev list`  table 'old' style was display
- I tried setting it via CLI
   - ` > $null` is the Powershell eq of `>/dev/null`
   - I confirmed that the setting was being changed in my `~/.ddev/global_config.yaml`
DDEV-win10

- I tried manually editting the my `~/.ddev/global_config.yaml` with `default`, `bold` and `bright` without change
- I tried setting `table_style: invalid`, to see if would throw an error but it didn't. Should there be a warning for invalid style?
- Confirmed in 
   - native `Windows PowerShell 5.1.19041.1151 
   - `Powershell Core 7.1.4` via Windows Terminal

I'm new to WSL so I'm not sure how to install the file there.

rfay:
Nice work @tyler36 AGAIN.

There was some terminal detection code that didn't work exactly the same on Windows as it does everywhere else, and so your check and comment actually fixed a couple of other related bugs as well. The result was that it was not detecting either PS or git-bash as a terminal, and not getting the size properly. And it has a switch where if it detects not a terminal (like a pipe) it turns off styling. You can also turn it off deliberately in global config.

This one is at https://github.com/drud/ddev/actions/runs/1184800446 and https://gitpod.io/#https://github.com/drud/ddev/pull/2965

You're a delight to work with!

tyler36:
OK ... checked in Powershell & Powershell Core

- There appears to be a `=` added at the end of `default` table
ddev-win10-default

- There appears to be a `=` added at the end of `bold` table
ddev-win10-bold

- `bright` style looks good
ddev-win10-bright


- `ddev describe` always appears as below regardless of style setting (default, bright, bold). I would expect it to appear styled, or at the lest similar to `default` in appearance (IE. lines instead of dash)
ddev-win10-describe


- just to be "delightful" (😄)  I threw some Japanese (kanji) at it to see if it would freak out.
  - it didn't like the kanji as a hostname (didn't expect it too)
  - table displayed the folder name correctly which is great!
ddev-japanese-hostname

tyler36:
Tested on WSL:
- `default`, `bold`, `bright` all look good
- `ddev describe` displays with the "dash" style instead new themed styles (not implemented yet?)
- same great results with Japanese as above 👍

rfay:
The "=" is a truncation marker, it's telling you that the window isn't quite wide enough for what's to be displayed. Judging from your screenshot it sure looks wide enough, but the behavior is the same elsewhere, and when I widen my window I get rid of the truncation marker.

`ddev describe` seems to revert to the "dash" setup only when project is stopped. Will have to take a look at that.

tyler36:
Congrats on the merge!

Regarding the `=`, on ddev version v1.18.0-rc1, it seems to happen between 1447px & 1456px (snap-points) ... not sure why the table thinks it truncated though 🐛 .

- 1447px
win10-1447px

- 1456px
win10-1456px

rfay:
What terminal is this, PowerShell? I don't have any trouble with PowerShell, but I'm not a big user.

tyler36:
The above were in Powershell Core, but that was running through Windows Terminal which, in hindsite, might be complicating the matter.

I switch to a native Powershell 5.1.19041.1151 prompt.

- 1124px
ddev-powershell-1124

- 1125px
ddev-powershell-1125

The scrollbar is showing which may be affecting it.

rfay:
@tyler36 due to a mistake in the code you don't even have to turn debug on to find out what the perceived terminal width is. Use `ddev describe`:

Cursor_and_rfay-win10pro

rfay:
Are you using utf-8? Plain vanilla PS seems to work perfectly for me.

Cursor_and_rfay-win10pro

tyler36:
Oh, that seems easier then taking screenshot and using those sizes.

- `ddev describe` never shows the marks, so the issue seems to be with `ddev list`

I played around with the window

- `w158`:  everything bigger than this appears OK
ddev-w158

- `w157`: everything smaller than this shows the marks
ddev-w157

ddev-properties

rfay:
Sure wonder what the difference is.

Cursor_and_rfay-win10pro

rfay:
Native windows can sure be hard. Let's do follow this up, and nail it. Let's do followups in 
* https://github.com/drud/ddev/issues/3207

PR #2966, opened 2021-04-26
Check for environment variable DDEV_ALLOW_PUSH before allowing provider push

amitaibu:
## The Problem/Issue/Bug:

Allow the dangerous push of DB or files only after an env is set

## Related Issue Link(s):

https://github.com/drud/ddev/pull/2963#pullrequestreview-644054303

I've never written in Go, so this is just a naive (and so far untested) implementation.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/2966"></a>

### DIFF
--- a/pkg/ddevapp/provider.go
+++ b/pkg/ddevapp/provider.go
@@ -226,7 +226,7 @@ func (p *Provider) UploadDB() error {
 	}
 
 	if os.Getenv("DDEV_ALLOW_PUSH") != "true" {
-		util.Warning("You must set the enviorement DDEV_ALLOW_PUSH=true to allow DB push")
+		util.Warning("You must set the environment variable DDEV_ALLOW_PUSH=true to allow DB push")
 		return nil
 	}
 
--- a/pkg/ddevapp/provider.go
+++ b/pkg/ddevapp/provider.go
@@ -2,6 +2,7 @@ package ddevapp
 
 import (
 	"github.com/drud/ddev/pkg/output"
+	"github.com/pkg/errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


pkg/ddevapp/provider.go
```diff
@@ -225,6 +225,11 @@ func (p *Provider) UploadDB() error {
 		return nil
 	}
 
+	if os.Getenv("DDEV_ALLOW_PUSH") != "true" {
+		util.Warning("You must set the enviorement DDEV_ALLOW_PUSH=true to allow DB push")
```
rfay, line 229:
```suggestion
		util.Warning("You must set the environment variable DDEV_ALLOW_PUSH=true to allow DB push")
```

rfay (changes requested):
I added DDEV_ALLOW_PUSH=true to the CircleCI tests so push will still be executed there, so the push tests will run OK.

Then we just need text in the example files and the docs explaining how to enable push. 

But This should be on the right path.

rfay:
This PR can be tested easily using gitpod, https://gitpod.io/#https://github.com/drud/ddev/pull/2966

All the necessary build tools are added there, and you can run `make` with no trouble, and the built ddev is automatically in the PATH

rfay (changes requested):
Rebased this to current, but realized it's not really right. DDEV_ALLOW_PUSH has to be specified in the *host* environment for this to work, but the docs I added expect it to be in the web container.

rfay:
I guess I don't think that people are going to do this accidentally, but I'm fine with finding a path to prevent it in the long term. For now, though, neither you nor I seem to be getting to it, so I'm going to close this. happy to reopen it, and happy for the contribution!

PR #2982, opened 2021-05-01
Gitignore .ddev/commands/web/live

jonaseberle:
This adds the autogenerated ddev command "live" to the autogenerated .ddev/.gitignore

## The Problem/Issue/Bug:
After updating to v1.17.2, ddev-using repositories would show `.ddev/commands/web/live` as newly added file.

Add file to .gitignore

1. `ddev start` in a clean git repository
2. Check if `git status` reports `.ddev/commands/web/live` as new file

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/2982"></a>

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1005,7 +1005,7 @@ func PrepDdevDirectory(dir string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for this. I guess it was shortsighted to remove that gitignore. Of course... the file can just be deleted, and it's best to remove it (and have one less command listed). But that doesn't mean we can't gitignore it.

jonaseberle:
I actually misinterpreted that: I really though `live` was autogenerated. But actually it is not anymore but was already there from an earlier version.

I don't know how to proceed.

rfay:
It was previously autogenerated, so any instances of it that are left are from previous versions. 

I think your PR is probably the best path. 

This is probably the first time we've removed a shell command in an update; there's no logic to remove a previously-generated command, and it would probably be complex and risky anyway.

rfay:
I added a caveat to the release notes about this, https://github.com/drud/ddev/releases/tag/v1.17.2

rfay (approved):
Thanks for this.

PR #2983, opened 2021-05-01
Add xhprof to ddev, fixes #2974

### LINKED ISSUE #2974: Add a command to enable xhprof for ddev (like xdebug)
I am always frustrated when I want to profile a web page and last I followed Daniels post here:

(https://daniel.town/profiling-php-in-ddev-using-tideways-locally/)

@rfay now showed me that it's as simple as doing:

```
webimage_extra_packages: [php7.4-xhprof]
```

but I was not able to find that solution anywhere.

I would love that to be as simple as with xdebug to just do:

```
ddev xhprof on
```

and if I don't need it anymore:

```
ddev xhprof off
```

It can be done manual, but a command sounds nicer. OSS should not only support proprietary (even if free) solutions like blackfire, but always also provide a free alternative with equal ease of use. (My old Debian / GNU background shines through here :D)

**Additional context**

The command will be great!

penyaskito:
## The Problem/Issue/Bug:

#2974 Add xhprof 

* Adds the xhprof php extension
* Adds command to enable
* Adds the XHProf UI (php) to allow exploring results

Added a bash script as xdebug.

* Use the [ddev download from this build](https://github.com/drud/ddev/actions/runs/958476279); you can `brew unlink ddev` and just put the downloaded ddev in your PATH. Alternately, use gitpod by visiting [link](https://gitpod.io/#https://github.com/drud/ddev/pull/2983) and you'll have the built item all set up. (You have to be logged into github to see the artifacts)
* If your project has a custom .ddev/apache/apache-site.conf or a custom .ddev/nginx_full/nginx-site.conf, please remove it temporarily (the URL for the xhprof app is added in the default configs)
* `ddev xhprof on`
* Visit a page; Follow (and test) instructions in https://github.com/drud/ddev/blob/8074d8e00a15051385149b873f5b6b0f66b999e7/docs/users/xhprof-profiling.md

* TestDdevXhprofEnabled has pretty much full coverage.

## Related Issue Link(s):

#2974 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/2983"></a>


penyaskito:
I looked at how xdebug works, and it depends on the container /usr/local/bin/xdebug_enable  and /usr/local/bin/xdebug_disable scripts. Should we do the same here? If yes, what's the preferred way of putting those in the container?

rfay:
Looks like you've got it just fine here. 

The only reason the xdebug script uses enable_xdebug and disable_xdebug is that they were there before it was.

penyaskito:
@rfay I think this is ready then. I just added the PHP version to php(dis|en)mod as I was getting two warnings for each of the versions that were inactive, e.g. using 7.4:

```
WARNING: Module xhprof ini file doesn't exist under /etc/php/8.0/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/8.0/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.3/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.3/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.2/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.2/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.1/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.1/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.0/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/7.0/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/5.6/mods-available
WARNING: Module xhprof ini file doesn't exist under /etc/php/5.6/mods-available
```

rfay:
Sounds like the right thing to do. deb.sury.org keeps changing how that kind of things works, sounds like what you did should be fine.

rfay (changes requested):
I think we should just add php*-xhprof to the standard extensions. 

Unfortunately, this is still done upstream in drud/ddev-images, 

So I'll do a PR there and we can bring it in here.

rfay:
I took the liberty of pushing upstream installation of xhprof in there, so make sure you pull before working further.

This will still need 
* docs
* A golang test, probably based on the xdebug tests
* Consideration of whether xhprof should be enabled by default
* Consideration of how the gui part should work in general.

penyaskito:
Added docs to cli-usage. Added DDEV_XHPROF_ENABLED but didn't document it. For the record, DDEV_XDEBUG_ENABLED is not documented neither.

Did my best to add a golang test at least for the enabling/disabling.

Consideration of whether xhprof should be enabled by default: we agree it shouldn't. Disabled that, which forced to refactor the xhprof and add scripts for enable_xhprof and disable_xhprof that can be run manually inside the container. That mimics what xdebug does already, so much better.  

Consideration of how the gui part should work in general: Not sure what you mean here. Do you want to have a gui in ddev by default, as in xhgui?

rfay:
upstream/master got a change to ddev-webserver, so
* I rebased
* I pushed an updated version of drud/ddev-webserver:20210502_penyaskito_xhprof

If you work on it, make sure you pull the docker image and reset your branch to the current upstream.

thomasdiluccio:
There may be conflicts between Blackfire and Xhprof. Activating one should deactivate the other.

thomasdiluccio:
@penyaskito You may have already did. But if that's not the case, you may want to have a look at the Blackfire commands
[https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/global_dotddev_assets/commands/web/blackfire](https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/global_dotddev_assets/commands/web/blackfire)

You should be able to tweak of copy of it for your use case.

And, we may have to update both commands to make sure Blackfire, Xhprof, and Xdebug won't conflict one another.

rfay:
It's true that the blackfire command would have been a better starter than the xdebug one. And yes, I think xdebug and blackfire should probably be disabled when starting.

penyaskito:
@thomasdiluccio would be enough to call "blackfire disable" in "xhprof_enable"? If I'm right, blackfire spawns its own process, while xhprof doesn't.

thomasdiluccio:
> @thomasdiluccio would be enough to call "blackfire disable" in "xhprof_enable"? If I'm right, blackfire spawns its own process, while xhprof doesn't.

It should be. It's quite unlucky that someone may want to switch between profiling and monitoring solutions.

rfay (approved):
Rebased and pushed the ddev-webserver image. I think this is ready to go. I added minor docs, but now there's a place for them and we can add more details as we figure out how you actually use xhprof.

rfay:
Updated OP with review instructions

rfay:
@penyaskito @LionsAd Would appreciate if you could review this. Manual testing is the most important, but code and approach review are also appreciated.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/enable_xhprof
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+export PATH=$PATH:/usr/sbin:/sbin
+phpdismod blackfire
+mkdir -p ${XHPROF_OUTPUT_DIR}
+phpenmod xhprof
+killall -HUP php-fpm 2>/dev/null || true
```
andypost, line 6: I think it should use `USR2` signal instead of hup

containers/ddev-webserver/Dockerfile
```diff
@@ -125,14 +127,14 @@ RUN chmod -R ugo+w /usr/sbin /usr/bin /etc/nginx /var/cache/nginx /var/lib/nginx
 
 RUN source /tmp/ddev/vars && curl -sSL https://github.com/drud/mkcert/releases/download/${MKCERT_VERSION}/mkcert-${MKCERT_VERSION}-${MKCERT_ARCH} -o /usr/local/bin/mkcert && chmod 777 /usr/local/bin/mkcert
 
+RUN git clone https://github.com/longxinH/xhprof.git /var/www/xhprof
```
andypost, line 130: Why you're using git? Maybe better to use pecl source, because no reason to fetch all history

shaal (approved):
I tested it using the Gitpod link 😁 

I installed Umami demo on Drupal 9.2
I ran `ddev xhprof`
After opening the homepage, I opened `/xhprof/` link, and I was able to see all the data from xhprof.

image

This was so easy to setup, great work!

LionsAd (approved):
RTBM - Looks like a great start to me. We can always improve the prepend and append more later.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/enable_xhprof
rfay: It looks like perhaps php-fpm could have ended up with SIGHUP in addition, https://bugs.php.net/bug.php?id=67553, but I don't see traction on that.

andypost: Thanks for links, this bug is old and as PHP is shipping init.* scripts nobody cares about to add HUP support. Maybe filing PR could help as PHP 8.1 going to feature freeze

rfay:
Congratulations and thanks for making this happen @penyaskito ! Thanks to Tag1 for financial support!

penyaskito:
yay! Thanks for your support @rfay, and thanks Tag1 for supporting maintainers!

emb03:
Yay! This should be the docs!
🥇

PR #3008, opened 2021-05-18
Adding PHIVE to the PHP image

### LINKED ISSUE #2990: Add Phive to the PHP image?
To install development aids like PHPUnit or similar tools, we use [Phive](https://github.com/phar-io/phive). It would be great to include their Phar in the ddev PHP images, such that it can be used in the same way as composer

NicoHaase:
This PR adds the PHIVE binary to the PHP image used by ddev

Fixes #2990 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3008"></a>

### DIFF
--- a/.spellcheckwordlist.txt
+++ b/.spellcheckwordlist.txt
@@ -73,6 +73,8 @@ Netbeans
 Nginx
 OSs
 OpenMage
+PHAR
+PHIVE
 PHP
 PID
 PWD
--- a/cmd/ddev/cmd/dotddev_assets/commands/web/phive
+++ b/cmd/ddev/cmd/dotddev_assets/commands/web/phive
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run PHIVE inside the web container
+## Usage: phive [flags] [args]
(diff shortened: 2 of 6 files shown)


rfay (changes requested):
Thanks! 

I think given this is getting installed, it should then be used for the various phar installs that get done in the container build?

For testing, this will require an update of pkg/version/version.go and a push of the requisite version.

NicoHaase:
> I think given this is getting installed, it should then be used for the various phar installs that get done in the container build?

You mean that installing the tools for the images should also take place using Phive? That sounds plausible, but I'd have to check how to do that...

> For testing, this will require an update of pkg/version/version.go and a push of the requisite version.

Yeah, I'm already reading the tutorial about how to test this using Gitpod - that's pretty slow, so I might check this on my own machine

rfay:
Yes, I just mean that the phive could go in first, then the parts of the build that download phars could use it.

Building and testing is fine on your own machine, but gitpod is usually lots faster. 

Thanks!

But then you deleted it!

rfay:
Shucks, branch naming is not that important.

PR #3011, opened 2021-05-19
Allow disabling http/2 support in ddev-router

rfay:
## The Problem/Issue/Bug:

There are situations where http/2 is not wanted; for example the Safari browser seems not to get along well with large numbers of cookies and http/2

*. Add `ddev global config --disable-http2` (and disable_http in global_config.yaml)
* Add support in ddev-router to respect that config

* `ddev poweroff`, `ddev config global --disable-http2`, `ddev start`, `curl -I <site>`. The first line should show HTTP/1.1
* `ddev poweroff`, `ddev config global --disable-http2=false`, `ddev start`, `curl -I <site>`. The first line should show HTTP/2

Added to the test of global config
Added TestDisableHTTP2

## Related Issue Link(s):

https://github.com/drud/ddev/pull/2698 (which increased http2 header size) and https://github.com/drud/ddev/discussions/2697

### DIFF
--- a/cmd/ddev/cmd/config-global.go
+++ b/cmd/ddev/cmd/config-global.go
@@ -74,6 +74,12 @@ func handleGlobalConfig(cmd *cobra.Command, args []string) {
 		dirty = true
 	}
 
+	if cmd.Flag("disable-http2").Changed {
+		val, _ := cmd.Flags().GetBool("disable-http2")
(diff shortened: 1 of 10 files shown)


LionsAd (approved):
RTBM - if tests pass — looks great to me

PR #3017, opened 2021-05-20
Check ENV variable first before doing a FS stat in settings.ddev.php

markhalliwell:
Very minor nit, but no need to check whether the file is readable if it's not a DDEV environment.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3017"></a>

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -69,7 +69,7 @@ func NewDrupalSettings(app *DdevApp) *DrupalSettings {
 const settingsIncludeStanza = `
 // Automatically generated include for settings managed by ddev.
 $ddev_settings = dirname(__FILE__) . '/settings.ddev.php';
-if (is_readable($ddev_settings) && getenv('IS_DDEV_PROJECT') == 'true') {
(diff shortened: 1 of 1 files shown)


plach79 (approved):
Makes sense to me

rfay:
Could you please say why this matters?

markhalliwell:
Logically, it just doesn't make a lot of sense to access the physical system when it can first check the environment variable which is already loaded in memory.

From a performance standpoint, I get that this isn't going to matter much. However, this code is injected into `settings.php` which is something that is still executed in a production environment. While one tiny `is_readable` call isn't going to do much, it also doesn't hurt avoiding a completely unnecessary `stat()` on each request when it doesn't have to.

rfay:
I guess this is OK. One note though: settings.ddev.php is never intended to exist in a normal production environment. It's gitignored and shouldn't show up there at all.

markhalliwell:
Thanks for the merge 😄 I am aware that settings.ddev.php shouldn't/doesn't get committed. But it would still check for one, even on prod because this was autogenerated into `settings.php` which is committed.

PR #3034, opened 2021-06-03
Increase ddev-webserver apache/nginx limit request field size for large cookie payloads

rfay:
## The Problem/Issue/Bug:

Some users have run into trouble with huge cookie payloads, both in nginx and apache. This doubles the relevant [nginx default value](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers) is "large_client_header_buffers 4 8k;". This raises the size of the nginx buffers to "large_client_header_buffers 4 16k;"

Apache LimitRequestFieldSize  is doubled to 16380.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3034"></a>


rfay:
@TravisWhitehead would love it if you could review this, thanks.

TravisWhitehead (approved):
This LGTM. :+1:

PR #3039, opened 2021-06-05
Download WinNFSd and NSSM on demand fixes #2973, ddev_windows_installer detected as virus

### LINKED ISSUE #2973: Sophos considers ddev_windows_installer v1.17.1 a threat or virus in Windows.
In windows 10, Sophos will delete ddev as soon as it's downloaded. "choco install -y ddev" will be stopped too. 

"Generic ML PUA detected at \Downloads\ddev_windows_installer.v1.17.1.exe"

gilbertsoft:
## The Problem/Issue/Bug:
Virus tools more and more are recognizing WinNFSd and NSSM as malicious software and so these get not longer bundled to the installer but downloaded on demand during the installation to avoid issues.

WinNFSd and NSSM are now downloaded on demand, so only local virus scanners can complain.

Remove the previous DDEV installation or in minimum remove the nssm.exe and WinNFSd.exe so see if it's correctly downloaded and installed with the new installer.

## Related Issue Link(s):
#2973 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3039"></a>


.gitattributes
```diff
@@ -19,3 +19,7 @@
 *.tar binary
 *gz binary
 *MY? binary
+
+# Force line endings to CRLF for NSIS files
+*.nsi text eol=crlf
+*.nsh text eol=crlf
```
rfay, line 25: This would be better in the winpkg directory wouldn't it?

gilbertsoft:
Finally I've decided to stay on CRLF for the Windows related files as this is the default file ending for Windows. Therefor the .gitattributes changes are now done in winpkg like suggested. I've also saved all NSIS files with the new .editorconfig, I guess the whitespace changes are related to the change to UTF8 because line endings did not change, the initial development of these files was done with the NSIS editor which did not support UTF8 at this time two years ago I guess. So after merging this PR that mess should be finally solved.

rfay:
Awesome, thanks. I'll try to test this tomorrow. 

I imagine we can remove the related downloading of winnfsd and nssm from the Makefile too, right?

winpkg/ddev.nsi
```diff
@@ -1,1045 +1,1083 @@
...
+   */
+  Section "NSSM" SecNSSM
+    ; Install in non choco mode only
+    ${IfNot} ${Chocolatey}
+      SectionIn 1
+      SetOutPath "$INSTDIR"
+      SetOverwrite try
+
+      ; Copy files
+      ;File "..\.gotmp\bin\windows_amd64\nssm.exe"
```
rfay, line 564:
```suggestion
```

winpkg/ddev.nsi
```diff
@@ -1,1045 +1,1083 @@
...
+  /**
+   * WinNFSd application install
+   */
+  Section "WinNFSd" SecWinNFSd
+    SectionIn 1
+    SetOutPath "$INSTDIR"
+    SetOverwrite try
+
+    ; Copy files
+    ;File "..\.gotmp\bin\windows_amd64\winnfsd.exe"
```
rfay, line 529:
```suggestion
```

winpkg/ddev.nsi
```diff
@@ -1,1045 +1,1083 @@
...
+    File "..\.gotmp\bin\windows_amd64\winnfsd_license.txt"
```
rfay, line 530: This file is currently downloaded by the Makefile, it's OK, but a little odd since we get the binary while running the thing. Probably slightly better to remove from the Makefile and download with the binary.

rfay (changes requested):
I manually tested this from the github actions build artifacts, https://github.com/drud/ddev/actions/runs/911370731 and it worked fine. Minor suggestions, and we should remove the binary downloads from the Makefile.

It might be cleaner to download the license files with the binaries as opposed to downloading them in the Makefile, but that doesn't seem like a big deal.

gilbertsoft:
> 
> 
> Awesome, thanks. I'll try to test this tomorrow.
> 
> I imagine we can remove the related downloading of winnfsd and nssm from the Makefile too, right?

Yes. I can include this into my PR and remove the unneeded parts you mentioned above

gilbertsoft:
Final cleanup is done

rfay:
Thanks so much!

rfay (approved):
The changes look fine, the intent looks fine. I tested manually by removing ddev and uninstalling the nfsd service and ~/.ddev/nfs_exports.txt, then installing using the artifacts from https://github.com/drud/ddev/actions/runs/915604283 and running the nfsd installation script, and everything seemed to work fine.

I'm not sure the way the licenses are presented is altogether clear, but I can definitely live with that. 

Thanks so much for understanding what to do here and then doing it!

I think this will solve the problem for most people, and we may have a few problems with the few people who use Sophos until they (Sophos) get this straightened out.

gilbertsoft:
> I'm not sure the way the licenses are presented is altogether clear, but I can definitely live with that.
> 
In the registry at `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\DDEV` there are various `NSIS:*LicenseAccepted` values, if you delete those the licenses will be shown during the installation again.

PR #3042, opened 2021-06-07
Don't attempt to sign windows binaries we don't have

rfay:
## The Problem/Issue/Bug:

In #3039 we stopped downloading winnfsd.exe and nssm.exe because they caused false virus checker warnings.

But we can't sign them then. Stop signing them.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3042"></a>


gilbertsoft:
Ops, thought I've catched them all but now I remeber I did only search for the URLs

PR #3050, opened 2021-06-14
Improve composer_version docs

jameswilson:
## The Problem/Issue/Bug:

[Documentation](https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/) states that specifying "2" for `composer_version` in .ddev/config.yaml would install the latest version of composer, but currently "2.1.1" is installed using latest current ddev version 1.17.5 (at a time when "2.1.3" is the latest available version of composer). Furthermore, when using an older version of ddev 1.17.2, the latest version installed is "2.0.13".

Also, the docs are unclear as it states that v1 is the default, then later states that composer 2 is default.

This is just a rewording of the docs to be more explicit:

* Remove "v1" being mentioned as default.
* Reword language around "latest v1"  and "latest v2" so it is clear that this isn't the actual latest version released, but the latest version available at the time the ddev-webserver image was bundled for a ddev release.

N/a

No tests needed. Docs update only.

## Related Issue Link(s):
https://drupal.slack.com/archives/C5TQRQZRR/p1623701639291500

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3050"></a>


docs/users/extend/config_yaml.md
```diff
@@ -23,7 +23,7 @@ the .ddev/config.yaml is the primary configuration for the project.
...
 | webserver_type | nginx-fpm or apache-fpm | The default is nginx-fpm, and it works best for many projects.|
 | timezone | timezone to use in container and in PHP configuration | It can be set to any valid timezone, see [timezone list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). For example "Europe/Dublin" or "MST7MDT". The default is UTC. |
-| composer_version | version of composer to use in web container and `ddev composer` | It can be left blank to get the default composer v1, or set to "1", for latest composer v1, "2" for latest composer v2, or to an explicit composer version. (Composer v2 is currently the default for ddev.) |
+| composer_version | version of composer to use in web container and `ddev composer` | Set to "2" (default) for composer v2 or "1" for composer v1 to use the latest major.minor.patch versions available at the time your currently installed ddev version was bundled and released. Note that the bundled default version might be behind the latest available composer release. Alternatively, an explicit composer version may be specified. Leaving the value blank assumes the default. |
```
rfay, line 26:
```suggestion
| composer_version | version of composer to use in web container and `ddev composer` | It defaults to composer v2; you can set it to "" or "2" (default) for composer v2 or "1" for composer v1 to use the latest major.minor.patch versions available at the time your currently installed ddev version was bundled and released. Note that the bundled default version might be behind the latest available composer release. Alternatively, an explicit composer version may be specified, for example `composer_version: 1.0.22`. |
```

rfay (approved):
Thanks for taking the time to improve this!  I added a minor wording update.

PR #3053, opened 2021-06-15
Use large_client_header_buffers instead of http2_max_header_size for nginx config

rfay:
## The Problem/Issue/Bug:

I noticed this error in the ddev-webserver log:

> the "http2_max_header_size" directive is obsolete, use the "large_client_header_buffers" directive instead in /etc/nginx/nginx.conf:133

So somewhere in our doubling of the http2_max_header_size journey we did/did not switch to the new format. 

This switches to the new format in ddev-webserver and ddev-router.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3053"></a>

### DIFF
--- a/containers/ddev-router/etc/nginx/nginx.conf
+++ b/containers/ddev-router/etc/nginx/nginx.conf
@@ -31,10 +31,6 @@ http {
 
   #gzip  on;
 
-  # Double default value to accommodate larger requests (often due to big cookies)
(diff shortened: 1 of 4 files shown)


rfay:
@TravisWhitehead would appreciate your review of this. I guess nginx got ahead of us during the various times we were working on this.

TravisWhitehead:
@rfay Looks good to me. :+1:

Issue #3051, opened 2021-06-15
findDirectivesInScriptCommand trims quotes from custom command files, breaks help text of provided commands


diegoe:
**Describe the bug**
In `findDirectivesInScriptCommand`, spaces, quotes, and double quotes, are trimmed from strings in the command description:

```go
    for scanner.Scan() {
        line := scanner.Text()
        if strings.HasPrefix(line, "## ") && strings.Contains(line, ":") {
            line = strings.Replace(line, "## ", "", 1)
            parts := strings.SplitN(line, ":", 2)
            parts[1] = strings.Trim(parts[1], " \"'")
            directives[parts[0]] = parts[1]
        }
    }
```

This breaks the formatting of help lines in the default provided commands (example, `cmd/ddev/cmd/global_dotddev_assets/commands/web/drush`)

```shell
$ ddev drush --help
(...)
Examples:
  ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version
(...)
```

Versus the source:
```shell
#ddev-generated
## Description: Run drush CLI inside the web container
## Usage: drush [flags] [args]
## Example: "ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version"
## ProjectTypes: drupal7,drupal8,drupal9,backdrop
(...)
```

**To Reproduce**
Run commands like `ddev drush --help`

**Expected behavior**
String is passed as is, or at least the provided commands don't fall into this bug.

**Version and configuration information (please complete the following information):**

- macOS Big Sur 10.4
- Docker Desktop version 3.4.0
- ddev version information (use `ddev version`)
```
DDEV-Local version	v1.17.5
architecture      	arm64
db                	drud/ddev-dbserver-mariadb-10.3:v1.17.3
dba               	phpmyadmin:5
ddev-ssh-agent    	drud/ddev-ssh-agent:v1.17.0
docker            	20.10.7
docker-compose    	1.29.2
os                	darwin
router            	drud/ddev-router:v1.17.3
web               	drud/ddev-webserver:v1.17.4
```

**Additional context**

rfay:
Good eye! 

Want to do a PR? You don't even need a development environment to work on it  and test with gitpod. https://gitpod.io/#https://github.com/drud/ddev

More hints in https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md

rfay:
This was fixed in https://github.com/drud/ddev/pull/3057

PR #3057, opened 2021-06-17
Do not trim quotes in findDirectivesInScriptCommand (#3051)

diegoe:
## The Problem/Issue/Bug:
https://github.com/drud/ddev/issues/3051

In 2ff4f0b17, triming was added for readability of file paths but the code
does not actually depend on quotes around paths. It only cares for
commas to differentiate multiple paths in a given directive.

This partially reverts:
```
    commit 2ff4f0b178734f602a4f26013fb48fb528ac0f30
    Author: Andreas Hager <3351175+andreashager@users.noreply.github.com>
    Date:   Tue Dec 29 05:47:40 2020 +0100

        Create new HeidiSQL command for Windows and WSL2 (#2679)
```

Call `ddev drush --help` or any command help in `cmd/ddev/cmd/global_dotddev_assets/commands/web/`

NA

## Related Issue Link(s):
https://github.com/drud/ddev/issues/3051

## Release/Deployment notes:
I can't test on Windows, which was the focus of the original commit

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3057"></a>

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -296,7 +296,7 @@ func findDirectivesInScriptCommand(script string) map[string]string {
 		if strings.HasPrefix(line, "## ") && strings.Contains(line, ":") {
 			line = strings.Replace(line, "## ", "", 1)
 			parts := strings.SplitN(line, ":", 2)
-			parts[1] = strings.Trim(parts[1], " \"'")
+			parts[1] = strings.Trim(parts[1], " ")
 			directives[parts[0]] = parts[1]
 		}
 	}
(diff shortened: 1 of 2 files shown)


rfay:
Thanks for this. I wonder if we should be trimming in most directives, but not in examples?

diegoe:
> Thanks for this. I wonder if we should be trimming in most directives, but not in examples?

FWIW the quotes here only help with readability, since you are still just passing the value along. I thought it was simpler to pass directives as-is, just to avoid surprises like the original bug.

Intuitively, I would interpret the example commands/directives as a simplistic "explode at marker" format. I usually don't expect much handling or formatting of the value. But that's just me 😆 

(Re pushed because I made a typo)

rfay:
Rebased and updated it so it doesn't trim quotes on Examples, but leaves alone everything else. I didn't want unexpected results, and this resolves the concern in #3051 

Thanks so much for this!

PR #3060, opened 2021-06-19
[docs] Minor improvements to solr recipe [skip ci][ci skip]

rfay:
## The Problem/Issue/Bug:

Minor docs improvements from a session walking through solr installation.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3060"></a>

### DIFF
--- a/docs/users/extend/additional-services.md
+++ b/docs/users/extend/additional-services.md
@@ -10,7 +10,7 @@ This recipe adds an Apache Solr container to a project. It will set up a solr co
 
 #### Installation
 
-1. Copy [docker-compose.solr.yaml](https://github.com/drud/ddev/tree/master/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml) to the .ddev folder for your project.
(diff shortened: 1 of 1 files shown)


rfay:
@hotwebmatter I'd love it if you could check through this set of changes, (full doc at https://github.com/drud/ddev/blob/9bd89ca723de7e05a6c9d505722ed0ffd8802922/docs/users/extend/additional-services.md) Working through it step-by-step on a WSL2 dummy instance would be marvellous!

I know you'll be doing more later.

Thanks!

rfay:
I think this just needs to be rethought. The whole weird add-a-script-at-the-right-time is just wrong.

hotwebmatter:
Hmm, I'm surprised you closed this PR, rather than doing the rethinking here.

It may be that the directory _can_ be mounted with the script inside it, in which case both need to exist before starting the Docker container. Maybe we can automate `mkdir -p ./.ddev/solr/docker-entrypoint-initdb.d` (this is idempotent and will cause no ill effects if the directory exists) and even put the config update script in that directory as a pre-start hook, if indeed it is possible to do such things before starting the containers.

PR #3071, opened 2021-06-25
Add docker-php-ext-* helper tools for odd php extensions, fixes #2987 [skip ci][ci skip]

### LINKED ISSUE #2987: Build PHP extensions from source with docker-php-ext-*
The official [PHP Docker Images](https://hub.docker.com/_/php/#how-to-install-more-php-extensions) provides convenient `docker-php-ext-configure`, `docker-php-ext-enable`, `docker-php-ext-install` commands for building PHP extensions from source. This is often useful when you need to build an extension from source on the GitHub, when the branch is still under development and is not in any repositories.

An example: Build [Phalcon Framework v5.0.x](https://github.com/phalcon/cphalcon/tree/5.0.x), which is still in alpha for further testing and adaptation with the codebase of our project.

Reimplement and provide these helper scripts in DDEV images.

https://github.com/docker-library/php/tree/master/7.4/buster/fpm

rfay:
## The Problem/Issue/Bug:

#2987 proposes adding docker-php-ext-* scripts to the web container to make it easy to add non-mainstream build-from-source php extensions.

Adds the scripts... but there's more to be done to make them work. And I imagine they're not useful without build-essential, which I probably wouldn't add because of size.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3071"></a>

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -43,6 +43,10 @@ RUN curl -sSL https://raw.githubusercontent.com/netz98/n98-magerun/develop/res/a
 RUN curl -sSL https://files.magerun.net/n98-magerun2-latest.phar -o /usr/local/bin/magerun2 && chmod 777 /usr/local/bin/magerun2
 RUN curl -sSL https://raw.githubusercontent.com/netz98/n98-magerun2/develop/res/autocompletion/bash/n98-magerun2.phar.bash -o /etc/bash_completion.d/n98-magerun2.phar && chmod +x /usr/local/bin/magerun
 
+RUN set -eu -o pipefail && for item in docker-php-ext-configure docker-php-ext-enable docker-php-ext-install docker-php-source; do \
+    curl -sSL -o /usr/local/bin/${item} https://raw.githubusercontent.com/docker-library/php/master/${item} && chmod ugo+rwx /usr/local/bin/${item} ; \
+done
+
 # /usr/local/bin may need to be updated by start.sh, etc
 RUN chmod -f ugo+rwx /usr/local/bin /usr/local/bin/composer
 
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -40,7 +40,7 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "v1.17.6" // Note that this can be overridden by make
+var WebTag = "20210625_docker-php-ext" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"


rfay:
I ran into some trouble with this because there are obviously missing steps before you can use these tools. Anyway, this build has the scripts in it. 

You can try it out with `docker run -it --rm drud/ddev-webserver:20210625_docker-php-ext bash` and do whatever you'd like to do. Help me understand what's missing. Show me a set of steps that will successfully use this for the currently-configured PHP version.

Thanks for your help @StudioMaX

StudioMaX:
That would be great.
As far as I understand, now the scripts are just downloaded from the original repository at build-time. In my opinion, it would be more correct to make copies of these scripts in your repository with the same names, adapt them, and when we run each script, add some pre-initialization script to download the PHP sources with the required version (depending on the DDEV config), and install all the necessary dependencies. 

Unfortunately I am not familiar with the internals of DDEV build scripts, but as I understood most of the steps can be taken from the [original Dockerfile](https://github.com/docker-library/php/blob/master/7.4/buster/fpm/Dockerfile). 

> And I imagine they're not useful without build-essential, which I probably wouldn't add because of size.

Can it just be installed the first time when someone execute docker-php-ext-* scripts?

rfay:
You don't need to know anything at all about ddev and its build scripts. As I said above, you just need to run `docker run -it --rm drud/ddev-webserver:20210625_docker-php-ext bash` and figure out how to use the scripts. If you figure that out, I'll add the necessary prereqs to the Dockerfile. But I think it will be way too heavyweight :(

StudioMaX:
Sure, I tried using your test image by running `docker run -it --rm drud/ddev-webserver:20210625_docker-php-ext bash` and ran into the same error as yours.
```bash
# docker-php-ext-install iconv
tar (child): /usr/src/php.tar.xz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
```

First, I downloaded archive with sources itself by manually executing the same commands from the original dockerfile:
https://github.com/docker-library/php/blob/541ab0e70f98cc3efce08ead9548285c984d6d33/7.4/buster/fpm/Dockerfile#L66-L98

After that, a new error occurred:
```bash
# docker-php-source extract
tar (child): xz: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
```
I realized that by default `tar` does not support unpacking `xz` archives at all, so I installed the `xz-utils` package (it was also mentioned in the original Dockerfile). After that, the archive was successfully unpacked using the `sudo docker-php-source extract` command. Then it became difficult to understand, since the scripts just don't find the dependencies they need. For example,
```bash
# docker-php-ext-configure gd
/usr/local/bin/docker-php-ext-configure: 68: /usr/local/bin/docker-php-ext-configure: phpize: not found
root@7b548e4b48c4:/usr/src# phpize
bash: phpize: command not found
```
`phpize` was not found, since in the original Docker PHP images PHP is built from source, but in our case it is just installed from the official repositories without dev-packages.

Therefore, I assumed that we would need some additional script (if we won't require all this stuff with the standard DDEV-PHP Docker images) that, depending on the DDEV config, will install the required php-dev package (for example `php7.4-dev` which contains `phpize` and header files) and all needed dependencies (`xz-utils`, etc).

If you're wondering how these scripts work together in a final Dockerfile, here's an example (Some dev packages of the libraries are unnecessary as this is only part of the Dockerfile):
```bash
RUN apt-get update && apt-get install -y \

        libfreetype6-dev \
        libjpeg62-turbo-dev \
        libmcrypt-dev \
        libpng-dev \
        zlib1g-dev \
        libbz2-dev \
        libxslt-dev \
        libicu-dev \
        libcurl3-dev \
        libc-client-dev \
        libkrb5-dev \
        libedit-dev \
        libxslt1-dev \
        libreadline6-dev \
        libtidy-dev \
        libxml++2.6-dev \
        g++ \
        unixodbc-dev \
        libxml2-dev \
        libaio-dev \
        freetds-dev \
        libssl-dev \
        openssl \
        curl \
        bzip2 \
        libfontconfig1 \
        libfontconfig1-dev \
        libzip-dev \
    && docker-php-ext-install -j$(nproc) iconv \
    && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
    && docker-php-ext-install -j$(nproc) gd \
    && pecl install redis-3.1.1 \
    && pecl install xdebug-2.7.2 \
    && docker-php-ext-enable redis xdebug \
    && docker-php-ext-configure imap --with-imap --with-imap-ssl --with-kerberos \
    && docker-php-ext-install bcmath bz2 curl dom fileinfo gettext imap intl json \
        mbstring mysqli pcntl pdo pdo_mysql phar posix readline session \
        shmop simplexml soap sockets tidy tokenizer xml zip \
    && pecl install apcu \
    && pecl install apcu_bc-1.0.4 \
    && docker-php-ext-enable apcu --ini-name 10-docker-php-ext-apcu.ini \
    && docker-php-ext-enable apc --ini-name 20-docker-php-ext-apc.ini
```

rfay:
It actually sounds like the original approaches as suggested in https://github.com/drud/ddev/issues/2987#issuecomment-831855689 are more appropriate, true?

rfay:
This will be a great project for [ddev-contrib](https://github.com/drud/ddev-contrib), but it's not appropriate for the huge image size increase that would result from adding it. I didn't realize how much beyond the scripts was required.

PR #3078, opened 2021-06-30
Use relative paths for dockerfile and context, for docker/compose-cli#1854

rfay:
## The Problem/Issue/Bug:

See docker/compose-cli#1854 - It seems that in the `build:` stanza of a docker-compose file, a Windows path (or at least an absolute Windows path) is not accepted for `context:` or for `dockerfile`

The various examples in the [compose spec](https://github.com/compose-spec/compose-spec/blob/b7fb5b65fff19651ecdb2d4d3232b01fb6866f70/build.md) all show relative paths. So I'm just changing this to relative paths in those places. It seemed to work in manual testing.

`ddev start` on Windows with compose v2 should work.

Will have to figure out how to get a full set of tests with docker-compose v2.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3078"></a>


rfay:
I tested this on windows and it seems to work OK!

PR #3086, opened 2021-07-03
Don't allow docker-compose v2 to be used

rfay:
## The Problem/Issue/Bug:

Currently there are so many bugs on docker-compose v2 beta that it's unusable for most ddev purposes.

This PR checks for compose v2 and refuses to use it, and gives instruction. 

For a current list of incompatibilities and problems, see https://github.com/drud/ddev/pull/3083#issuecomment-873060929

Improve the version constraint to disallow v2
Add an extra error message explaining about v2

With compose v2 enabled, `ddev start`. It should fail, explaining why.
Then `docker-compose disable-v2` and `ddev start`. It should work. 

## Release/Deployment notes:

Hopefully we can revert this when compose v2 is fully functional and compatible.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3086"></a>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,7 +6,7 @@
 
 * [Docker](https://www.docker.com/products/docker-desktop) version 18.06 or higher. Linux users make sure you upgrade docker-compose and do the [post-install steps](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user)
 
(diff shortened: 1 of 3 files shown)


ddltest:
Tested this on macOS, Windows 11, WSL2, and Ubuntu 20.04 and it seems to work correctly.

rfay:
Tested this on macOS, Windows 11, WSL2, and Ubuntu 20.04 and it seems to work correctly.

PR #3094, opened 2021-07-08
Update xhprof-profiling.md

iammati:
Dot after "PHP7" removed since "and PHP8" follows up

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3094"></a>

### DIFF
--- a/docs/users/xhprof-profiling.md
+++ b/docs/users/xhprof-profiling.md
@@ -1,6 +1,6 @@
 ## Profiling with xhprof
 
-DDEV-Local has built-in support for [xhprof](https://www.php.net/manual/en/book.xhprof.php). The official PECL xhprof extension does not support PHP5.6, but only PHP7 *and PHP8.*.
+DDEV-Local has built-in support for [xhprof](https://www.php.net/manual/en/book.xhprof.php). The official PECL xhprof extension does not support PHP5.6, but only PHP 7.\* and PHP 8.\*.
 
 ### Basic xhprof Usage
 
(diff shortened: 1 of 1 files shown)


docs/users/xhprof-profiling.md
```diff
@@ -1,6 +1,6 @@
 ## Profiling with xhprof
 
-DDEV-Local has built-in support for [xhprof](https://www.php.net/manual/en/book.xhprof.php). The official PECL xhprof extension does not support PHP5.6, but only PHP7.*and PHP8.*.
+DDEV-Local has built-in support for [xhprof](https://www.php.net/manual/en/book.xhprof.php). The official PECL xhprof extension does not support PHP5.6, but only PHP7 *and PHP8.*.
```
rfay, line 3: Actually it was supposed to be like this but was a markdown error. Thanks for catching it. Does this work for you?
```suggestion
DDEV-Local has built-in support for [xhprof](https://www.php.net/manual/en/book.xhprof.php). The official PECL xhprof extension does not support PHP5.6, but only PHP 7.\* and PHP 8.\*.
```

rfay: If so you can just commit the suggestion.

rfay (changes requested):
See if the new markdown works for you and commit it if you like it. Thanks for catching this.

rfay:
Looks OK to me now, thanks for catching that. 
ddev_xhprof-profiling_md_at_1ca83e23511e42aade4b411be0922131ebc338fd_·_drud_ddev

PR #3095, opened 2021-07-08
Remove version: key from docker-compose files, fixes #3067

### LINKED ISSUE #3067: Remove `version:` from docker-compose files
`version:` is no longer a "thing" in docker-compose, so our `version: "3.6"` just gets in the way.

* Remove from all ddev
* Remove from ddev-contrib
* Do this on a major version upgrade so it doesn't mess up too many projects.

rfay:
## The Problem/Issue/Bug:

#3067: docker-compose no longer requires the `version:` key in compose config files, and it has always been a bit of a mess.

Remove these things

If it passes testing it should be OK.

No changes

## Release/Deployment notes:

This has dramatic changes:
* Everything on ddev-contrib will have to be updated
* The minimum docker-compose version (of the tool) will have to be changed
* Docs will need updating (I think; this may already do enough)

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3095"></a>

### DIFF
--- a/cmd/ddev/cmd/debug-compose-config_test.go
+++ b/cmd/ddev/cmd/debug-compose-config_test.go
@@ -11,7 +11,6 @@ import (
 )
 
 var override = `
-version: '3.6'
 services:
   web:
     labels:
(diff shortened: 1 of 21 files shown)


rfay:
I don't think we'll be able to do this.

Per https://docs.docker.com/compose/compose-file/  
> The latest Compose file format is defined by the Compose Specification and is implemented by Docker Compose 1.27.0+

Unfortunately Ubuntu 20.04 has 1.25 by default, Debian 10 will be lower than that I think.

The motivation here was to lose the version field and let it go, but there are too many old docker-compose things out there. 

It would be safe, though, to move to docker-compose config version 3.7 instead of 3.6. It would just break a lot of custom configs.

rfay:
I won't try this for at least a year as it would require so many users, especially Linux users, to update docker-compose

PR #3107, opened 2021-07-20
Add suggestion how to convert the distro in step 12

velletti:
needed ,if it was created earlier as WLS 1 distro.

just added : 

If not , and your linux distro was created earlier with version 1, you can convert it in Powershell: `wsl --set-version <distro> 2`. Replace `<distro>` with the shown distro

## The Problem/Issue/Bug:
Running in Error in step 12 as my distro was set up more thatn 12 month ago without wsl --set-default-version 2 before

just tried it once and helped me

as it is an optional step, only needed when Step 12 does not show the  wanted : all distros have version 2 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3107"></a>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -74,7 +74,8 @@ We'll walk through these in more detail. You may prefer other techniques of inst
 9. Install a WSL2 distro from the Microsoft Store. We recommend [Ubuntu](https://www.microsoft.com/store/productId/9NBLGGH4MSV6). (You do not have to log in to the Microsoft Store, just "X" out the dialog box requesting that.)
 10. **Docker Desktop for Windows:** If you already have the latest Docker Desktop, configure it in the General Settings to use the WSL2-based engine. Otherwise install the latest Docker Desktop for Windows and select the WSL2-based engine (not legacy Hyper-V) when installing. Install via Chocolatey with `choco install docker-desktop` or it can be downloaded from [download.docker.com](https://download.docker.com/win/stable/Docker%20for%20Windows%20Installer.exe).  Start Docker. It may prompt you to log out and log in again, or reboot.
 11. Go to Docker Desktop settings > Resources > WSL integration > enable integration for your distro (now `docker` commands will be available from within your WSL2 distro).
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -74,7 +74,8 @@ We'll walk through these in more detail. You may prefer other techniques of inst
...
 11. Go to Docker Desktop settings > Resources > WSL integration > enable integration for your distro (now `docker` commands will be available from within your WSL2 distro).
-12. Double-check in PowerShell: `wsl -l -v` should show three distros, and your Ubuntu should be the default. All three should be WSL version 2.
+12. Double-check in PowerShell: `wsl -l -v` should show three distros, and your Ubuntu should be the default. All three should be WSL version 2. 
+If not , and your linux distro was created earlier with version 1, you can convert it in Powershell: `wsl --set-version <distro> 2`. Replace `<distro>` with the shown distro name, that still uses version 1.
```
rfay, line 78: I'm not entirely sure this is our job to coach this here, but I'll think about it. Unfortunately adding solutions for every single possible situation can really clutter up docs.

rfay: Really appreciate your contribution and hope you'll fix other things you find in the docs!

tyler36:
Step 9 references distro with [Ubuntu] recommended. Step 12 references Ubuntu but also  ....
``wsl --set-version <distro> 2`. Replace `<distro>``

For consistency,  I suggest the following changes 
```diff
- 9. Install a WSL2 distro from the Microsoft Store. We recommend [Ubuntu]
+ 9. Install a WSL2 `<distro>` from the Microsoft Store. We recommend [Ubuntu]
...
- 12. Double-check in PowerShell: `wsl -l -v` should show three distros, and your Ubuntu should be the default. All three should be WSL version 2. 
+ 12. Double-check in PowerShell: `wsl -l -v` should show three distros, and your `<distro>` should be the default. All three should be WSL version 2.

rfay:
I very much appreciate you taking the time to add this, but after consideration I don't think it belongs in ddev docs. 

Also, I recommend to people to create a new distro rather than converting one. It takes absolutely forever to convert one, and who knows what might go wrong.

PR #3113, opened 2021-07-22
Make mysql client history survive a ddev restart

cspitzlay:
## The Problem/Issue/Bug:
It would be nice if the SQL statements written to the client history file were still there after a `ddev restart`.
Currently they are not.

It sets an environment variable so the mysql client will write
its history to a location on a persistent volume.
That location is shared between the web and the db container.

- Remove the existing history files with `ddev start && ddev exec rm -rf /mnt/ddev-global-cache/mysqlhistory`
- `ddev restart` and verify that the directories now exist, `ddev exec ls -lR /mnt/ddev-global-cache/mysqlhistory`
- `ddev drush sqlc`
- Issue an SQL statement
- Quit
- `ddev restart`
- `ddev drush sqlc`
- Hit arrow up -> Statement should still be there, in v1.17.7 it is not.

## Notes:
1. There is no variable to limit the size of the file like there is for bash. From https://planet.mysql.com/entry/?id=5991652 : 
```
Despite the similarities to Bash, there is no MYSQL_HISTSIZE environment variable
to be like Bash's HISTSIZE. With some unreliable fiddling, 
one can try to tell the underlying library what the maximum history size should be
[...]
Usually this particular trick is best avoided because of its side effects.
```
2. The directory is not created automatically, if it doesn't already exist the history file is not written.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3113"></a>

### DIFF
--- a/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
+++ b/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
@@ -6,8 +6,6 @@
 ## Example: "ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version"
 ## ProjectTypes: drupal7,drupal8,drupal9,backdrop
 
-source ~/.bashrc
-
 if ! command -v drush >/dev/null; then
   echo "drush is not available. You may need to 'ddev composer require drush/drush'"
   exit 1
--- a/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
+++ b/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
@@ -6,6 +6,8 @@
 ## Example: "ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version"
 ## ProjectTypes: drupal7,drupal8,drupal9,backdrop
 
+source ~/.bashrc
+
 if ! command -v drush >/dev/null; then
   echo "drush is not available. You may need to 'ddev composer require drush/drush'"
   exit 1
--- a/containers/ddev-dbserver/files/docker-entrypoint.sh
+++ b/containers/ddev-dbserver/files/docker-entrypoint.sh
@@ -89,7 +89,7 @@ fi
 echo $server_db_version >/var/lib/mysql/db_mariadb_version.txt
 
 cp -r /home/{.my.cnf,.bashrc} ~/
-mkdir -p /mnt/ddev-global-cache/bashhistory/${HOSTNAME} || true
+mkdir -p /mnt/ddev-global-cache/{bashhistory,mysqlhistory}/${HOSTNAME} || true
 
 echo
 echo 'MySQL init process done. Ready for start up.'
--- a/containers/ddev-dbserver/files/home/.bashrc
+++ b/containers/ddev-dbserver/files/home/.bashrc
@@ -2,7 +2,3 @@
 export HISTFILE=/mnt/ddev-global-cache/bashhistory/${HOSTNAME}/bash_history
 export HISTSIZE=10000
 export HISTFILESIZE=100000
-
-# mysql history to be shared between web and db
-mkdir -p /mnt/ddev-global-cache/mysqlhistory/${DDEV_PROJECT}
-export MYSQL_HISTFILE=/mnt/ddev-global-cache/mysqlhistory/${DDEV_PROJECT}/mysql_history
--- a/containers/ddev-dbserver/files/home/.bashrc
+++ b/containers/ddev-dbserver/files/home/.bashrc
@@ -2,3 +2,7 @@
 export HISTFILE=/mnt/ddev-global-cache/bashhistory/${HOSTNAME}/bash_history
 export HISTSIZE=10000
 export HISTFILESIZE=100000
+
+# mysql history to be shared between web and db
+mkdir -p /mnt/ddev-global-cache/mysqlhistory/${DDEV_PROJECT}
+export MYSQL_HISTFILE=/mnt/ddev-global-cache/mysqlhistory/${DDEV_PROJECT}/mysql_history
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.bashrc
@@ -56,4 +56,8 @@ fi
 
 export HISTFILE=/mnt/ddev-global-cache/bashhistory/${HOSTNAME}/bash_history
 
+# mysql history to be shared between web and db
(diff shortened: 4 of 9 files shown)


cspitzlay:
I've finally gotten around to testing it. It works as expected from inside the containers (my-built-ddev ssh then drush sqlc). It survives a my-built-ddev restart and sharing the file with the command line mysql client in the DB container works, too). But it does not work when I use my-built-ddev drush sqlc or my-built-ddev exec drush sqlc from the host machine. I guess in that case the bashrc file with the environment variable is not read.

These are the changes I've made locally to the pkg/version/version.go file:

```
diff --git a/pkg/version/version.go b/pkg/version/version.go
index c669e743..a4b83cbc 100644
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -40,13 +40,13 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "v1.17.7" // Note that this can be overridden by make
+var WebTag = "20210722_cspitzlay_mysql_history" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"
 
 // BaseDBTag is the main tag, DBTag is constructed from it
-var BaseDBTag = "20210709_mariadb_versions"
+var BaseDBTag = "20210722_cspitzlay_mysql_history"
```

rfay:
Thanks, great progress! 

If you want to make it work with `ddev drush sql-cli` the answer would be in fixing https://github.com/drud/ddev/issues/2998, and I'd be OK with sneaking that in here.

I'll try to push your images later today.

rfay (approved):
This worked well, but still has the problems you've pointed out. Doesn't include working with `ddev mysql` or `ddev drush` (or `ddev exec anything`) just because we followed the bash history way of doing it, which was totally reasonable. But with bash history... we're talking about a bash session, and here we're not. 

Here (https://github.com/drud/ddev/pull/3113/commits/b5782ccb7e0ca36a9c1ee17d9ec73b801f1ee2b0) I experimented with just adding it in the template file that generates the .ddev-docker-compose-base.yaml, and I think it works out OK, if you're good with it. Simple and effective.

I wasn't able to use a single shared history, as the two seemed to write on top of each other. I'm fine with more experimentation on that though. I think you can see how easy it is to experiment.

 So if you test this and approve of it we can go with it. Thanks for pioneering all the other ideas. Sorry I didn't think of this approach in the first place.

cspitzlay:
I think the "writing on top of each other" is to be expected if you mean that "the last one wins". There is no live-sync. The same happens if you open two mysql command line clients on the same system, and I guess the same thing will happen with the bash history file if you have two shells in parallel.

The only situation where we make things worse it that it now extends to clients on web _and_ db. But I'd consider the feature that the user is able to reuse a statement on db that he used on web earlier outweighs this issue with possibly conflicting writes.

rfay:
What I found was that using mysql in the web container corrupted the file so it couldn't be used in the db container if the file was shared between them. I have no idea why, unless it keeps the file open and has an index in it or something. 

Anyway, it's easy to try out, just edit so they have the same file and build the binary and try it out using clients in both the web container and the db container. Maybe your experience will be different from what I thought happened. 

I actually examined the file and was horrified by what I saw :)

cspitzlay:
I'm not sure if we're both seeing the same kind of corruption but it's certainly different from what I've seen in my earlier tests. I hadn't looked at the actual contents of the file but given that the feature worked in all three use cases I'd say there hadn't been any corruption at the time.  But maybe I just didn't test thoroughly enough.

What I am seeing now is that the web container encodes spaces differently than the db container and the db container cannot read the format while the web container still can. The file written from web also has a '_HiStOrY_V2_' header that the db version does not.

The clients are using different libraries:
On web:
```
$ mysql --version
mysql  Ver 15.1 Distrib 10.5.10-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper
```
On db:
```
$ mysql --version
mysql  Ver 15.1 Distrib 10.5.11-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
```
I think this difference in encoding is related to this mysql issue: https://bugs.mysql.com/bug.php?id=72108
One comment even says "AFAIR this is a readline vs. libedit problem ...?"


**I suggest we go with separate files then to stay clear of these version / config / library issues.**
Which is what your latest version does.

There is one TODO left: Creation of the parent directory. If it does not already exist no history file is created. Do you know a good place to do that?
`mkdir -p /mnt/ddev-global-cache/mysqlhistory/${DDEV_PROJECT}`

rfay:
Excellent. I think separate files won't inconvenience many people. Glad you were able to sort out what's going on with the conflict. I have no idea offhand why the versions would be different, except that in the db container we install everything with server of course, and on the web we just install the client.

Are you sure we have to pre-create the directory? I never did that, but did wonder how it worked. If we had to do it it would be in the start script of the container I think.

cspitzlay:
Yes, I tested it a couple of minutes ago:

- The feature worked with your latest version.
- Then I removed the code. No history was stored any more.
- Then I manually created the directory again. History was stored again.

Could it be that you tested my version at least once before you started working on a better place to set the var?
In that case the dir had been created and since the whole point is that it is located on the persistent volume you already had it when you ran your version.

rfay:
Yes, I definitely used your version. So I guess we have to add the `mkdir -p` to the startup script in both the web and db containers.

cspitzlay:
Do you mean the `docker-entrypoint.sh`? That's where the bash history dir is created.
Is the variable `$DDEV_PROJECT` known in that script?

rfay:
It's start.sh in the webserver, and you'll see that DDEV_PROJECT is used there.
DDEV_PROJECT and DDEV_SITENAME are also available in the db container, where the start script is docker-entrypoint.sh

cspitzlay:
How about making the structure like the one for the bash history? Since the file is no longer shared there is no reason to use a different naming.

rfay:
Totally agreed. In fact, you can just add to the mkdir -p command used for bash history.

rfay (approved):
I rebased this and pushed, so `git fetch && git reset --hard origin/master`

I tested this after rebasing and it did fine.

I added instructions that include removing the /mnt/ddev-global-cache/mysqlhistory before restarting.

Thanks for this! When you think it's OK I'll pull it.

rfay:
Thanks for the great work on this. I've always secretly wished for it. And I'm very impressed with your persistence and care in manual testing. It would be great if everybody was that careful.

cspitzlay:
Thanks for your support in this. I couldn't have done it alone.

PR #3140, opened 2021-08-03
Add create config for D9 Git section

tekNorah:
## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3140"></a>


tekNorah:
This command will not work if you clone Drupal to the root directory (not using a separate docroot)

PR #3143, opened 2021-08-05
Make PHP_IDE_CONFIG available in 'ddev drush'.

cspitzlay:
## The Problem/Issue/Bug:

The PHP_IDE_CONFIG variable is not set when drush is started from the host machine via `ddev drush`.
As a result breakpoints in phpstorm do not work for code started that way.

It moves the line that sets the variable into a separate file that is included both in interactive shells, e.g. when running `ddev ssh` and from the drush wrapper script.

Just sourcing ~/.bashrc instead would not have worked because the line that sets the variable is after a check for interactive mode.  From /etc/bash.bashrc:
```
# If not running interactively, don't do anything
[ -z "$PS1" ] && return
```
In the drush script that variable is not set so the rest of the script ist not executed.

- In ~/.ddev/commands/web/drush, remove the #ddev-generated
- Put `export | grep PHP_IDE_CONFIG` before and after the line that sources the new file
- You will see the line with the variable only once.

## Related Issue Link(s):
See also https://github.com/drud/ddev/issues/2998

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3143"></a>

### DIFF
--- a/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
+++ b/cmd/ddev/cmd/global_dotddev_assets/commands/web/drush
@@ -6,6 +6,9 @@
 ## Example: "ddev drush uli" or "ddev drush sql-cli" or "ddev drush --version"
 ## ProjectTypes: drupal7,drupal8,drupal9,backdrop
 
+# Set up variable needed for debugging.
+source /etc/bash.php-ide-config
+
 if ! command -v drush >/dev/null; then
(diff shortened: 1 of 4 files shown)


rfay:
Please try #3149 - you might like it! Thanks for prodding me to try it.

cspitzlay:
I'm afraid I won't be around for the best part of August. Will try it when I'm back and it is still useful.

rfay:
I think you're going to like #3149 better, so closing this for now.

PR #3149, opened 2021-08-08
Remove PHP_IDE_CONFIG in php-fpm instead of trying to add it elsewhere, replaces #3143, fixes #2998, fixes #3106

### LINKED ISSUE #2998: Allow overriding PHP_IDE_CONFIG
I would like to change the value of `PHP_IDE_CONFIG` environment variable in the web container. The reason is that I'd like to unify the server name for XDebug requests coming from HTTP and the CLI. 

My projects typically have a name like `acme-corp-website` and I use `acme-corp.localhost` as local domain name. So when I start a Xdebug session from the browser PHPStorm recognizes a connection from the «server» `acme-corp.localhost`. Now that DDEV sets the value of this environment variable to `serverName=acme-corp.ddev.site` any Xdebug connection stared by CLI would result in PHPStorm assuming a different server. So I have to maintain two servers in PHPStorm for the same project.

The obvious way to solve this is something like this:

```yaml
# .ddev/docker-compose.override.yaml

version: '3.6'

services:
    web:
        environment: 
            PHP_IDE_CONFIG: serverName=acme-corp.localhost
```

I tired this with version 1.17.2 of DDEV but it seems that this very variable is somehow locked for overriding:
```
$ env | grep PHP_IDE
PHP_IDE_CONFIG=serverName=acme-corp.ddev.site
```

I could try to set  this variable via a custom Dockerfile or via a post-start hook in `.ddev/config.yaml` or even inside `.ddev/homeadditions/.bashrc` but all of these are not really convenient.

### LINKED ISSUE #3106: PHP_IDE_CONFIG - dont set via .bashrc
I got confused when I ran `ddev exec drush` and PHPStorm was telling me that PHP_IDE_CONFIG needs to be set in the container. I would do a `ddev ssh` and then see it there.  

Set PHP_IDE_CONFIG as  a regular env variable thats set even for non-login shells like DDEV_TLD, DDEV_PHP_VERSION, etc.

I prefer running commands via ddev exec versus ssh into the container because they stay in my bash history versus disappearing whenever the container is rebuilt. Plus my zsh is confugured on my host but I cant be bothered to do that in container.

rfay:
## The Problem/Issue/Bug:

* For PhpStorm zero-configuration debugging to work, PHP_IDE_CONFIG must not exist (for php-fpm)
* For command-line debugging, PHP_IDE_CONFIG must be set

After discovering this problem, the attempt was to *add* PHP_IDE_CONFIG only for login sessions, but people didn't like that for a variety of reasons (doesn't work then for `ddev exec` or `ddev drush`)

Instead of trying to ADD PHP_IDE_CONFIG for the command-line, instead add it everywhere (via docker-compose.yaml) and *hide* it from php-fpm by unsetting it before php-fpm or even supervisord is started

Thanks to @cspitzlay for pushing me to try another time!

This testing is critical:

* `docker pull  drud/ddev-webserver:20210808_PHP_IDE_CONFIG` to make sure you have the latest
* Start a project with this ddev (you can use gitpod per the links here)
* Remove any existing "Server" from the "PHP" section in PhpStorm
* Use the regular instructions for [zero-configuration xdebug](https://ddev.readthedocs.io/en/stable/users/step-debugging/#phpstorm) for PhpStorm.  You should get a popup that lets you choose a mapping. Verify that this works. (This also creates a new server/mapping). Verify that debugging a web app works now.
* Now go to the server/mapping in "PHP" and make sure the root of the project is mapped to /var/www/html
* Set a breakpoint and turn on xdebug and use `ddev drush` or `ddev exec someotherphp` to verify that debugging also works on the command line.

This is so PhpStorm fiddly that I didn't try adding anything.

## Related Issue Link(s):

* #3143
* #2998 
* #3106
* #2427
* #2424

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3149"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.bashrc
@@ -55,5 +55,3 @@ if [ -x /usr/lib/command-not-found -o -x /usr/share/command-not-found/command-no
 fi
 
 export HISTFILE=/mnt/ddev-global-cache/bashhistory/${HOSTNAME}/bash_history
-
-export PHP_IDE_CONFIG=serverName=${DDEV_SITENAME}.${DDEV_TLD}
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/php-fpm.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/php-fpm.conf
@@ -1,5 +1,5 @@
 [program:php-fpm]
-command = bash -c "unset PHP_IDE_CONFIG && exec /usr/sbin/php-fpm --nodaemonize --force-stderr"
+command = /usr/sbin/php-fpm --nodaemonize --force-stderr --allow-to-run-as-root
 priority=5
 stdout_logfile=/proc/self/fd/2
 stdout_logfile_maxbytes=0
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/php-fpm.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/php-fpm.conf
@@ -1,5 +1,5 @@
 [program:php-fpm]
-command = /usr/sbin/php-fpm --nodaemonize --force-stderr --allow-to-run-as-root
+command = bash -c "unset PHP_IDE_CONFIG && exec /usr/sbin/php-fpm --nodaemonize --force-stderr"
 priority=5
 stdout_logfile=/proc/self/fd/2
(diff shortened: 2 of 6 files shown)


rfay:
It looks like every runner fails on TestContainerWait, and when you look there, you see that php-fpm has died. I don't know why this would be, but it's one of the potential risks of putting an extra process between supervisord and php-fpm.

rfay:
I ran this through the test routines mentioned and it all worked perfectly. If it passes tests after rebasing, it's good to pull.

weitzman:
Thanks for improving this.

PR #3150, opened 2021-08-09
Use posix-raw as default symlink mode in mutagen.yml

rfay:
The only mutagen symlink mode that looks like it will work in general is the posix-raw, where mutagen "just syncs" the symlink. Because we want things to end up on the host if created in container, and vice versa.


rfay:
This doesn't work at all on Windows, so we'll have to figure something out.

* Don't use posix-raw on Windows
* Find out what the error is that's happening, because we haven't caught it here.

rfay:
Got [payment required](https://buildkite.com/drud/ddev-macos-m1/builds/783#4fd713e0-c68a-4fe3-bf49-0e4d2082425f) on blackfire!

PR #3154, opened 2021-08-10
Include memory consumption by default in xhprof output.

mrbaileys:
## The Problem/Issue/Bug:

When enabling xhprof, the output in the UI does not contain information about memory consumption.

Enable profiling of memory consumption alongside CPU profiling by adding the XHPROF_FLAGS_MEMORY when xhprof_enable is called.

This can be tested using gitpod: https://gitpod.io/#https://github.com/drud/ddev/pull/3154

Enable xhprof (ddev xhprof on).
Visit at least one page on your site.
Visit /xhprof and view the profiling output, it should contain information on memory consumption.

Not sure if (or how) this can be included in automated testing. Happy to add testing to this PR if someones points me in the right direction.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3154"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/xhprof_prepend.php
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/xhprof_prepend.php
@@ -7,7 +7,7 @@ if (!empty($_SERVER) && array_key_exists('REQUEST_URI', $_SERVER)) {
 
 // Enable xhprof profiling if we're not on an xhprof page
 if (extension_loaded('xhprof') && strpos($uri, '/xhprof') === false) {
-    xhprof_enable();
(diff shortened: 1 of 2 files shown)


rfay:
Thanks for this!

rfay (approved):
I tested this on gitpod and it works exactly as proposed.  I could wish for a larger window to view it in, but think this is what people will want. I'll plan to do a followup that allows people to edit their own prepend.

LionsAd:
@rfay Yes, let’s please prioritize overriding of the xhprof_prepend file via global ddev config.

rfay:
@LionsAd thanks for the ping,
* https://github.com/drud/ddev/issues/3168

PR #3160, opened 2021-08-12
Use working_dir in generated compose instead of always overriding, fixes #3158

### LINKED ISSUE #3158: Use `working_dir` in generated docker-compose instead of dynamically setting working_dir on launch
I've been following along with https://github.com/drud/ddev/issues/3130 and https://drupalize.me/blog/debug-any-drupals-phpunit-tests-phpstorm-ddev-local-environment but ran in to an issue when running [Drupal Test Traits](https://gitlab.com/weitzman/drupal-test-traits).

When phpstorm runs the test via `docker exec` it runs from `/` instead of `/var/www/html` and therefore DrupalFinder can't find Drupal. This is because [the working directory is dynamically added](https://github.com/drud/ddev/blob/07954280aaf2adc98613e6984f1aeeaf568d8ab1/pkg/ddevapp/ddevapp.go#L1217-L1219) to the ddev exec commands. Would it be possible to add the working_dir to the docker-compose file instead or as well?

To get around the issue for now I'm adding a custom docker compose file to set this:

.ddev/docker-compose.env.yml:

```
version: '3.6'
services:
  web:
    # This makes sure the working dir is set for running tests via phpstorm
    working_dir: /var/www/html
```

rfay:
## The Problem/Issue/Bug:

* #3158 

#3158 points out that we could more efficiently do working_dir with the docker-compose yaml instead of only passing it on the docker-compose command line. 

Put working_dir in the generated `.ddev/.ddev-docker-compose-*.yaml`

- [x] Use `ddev exec` and `ddev ssh` both with and without the `-d` flag, make sure it works all those ways.
- [x] Change working_dir in the .ddev/config.yaml and make sure that it becomes the default

I'm hoping the existing tests are adequate.

The behavior shouldn't change.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3160"></a>

### DIFF
--- a/pkg/ddevapp/templates.go
+++ b/pkg/ddevapp/templates.go
@@ -18,7 +18,7 @@ services:
         gid: '{{ .GID }}'
     image: ${DDEV_DBIMAGE}-${DDEV_SITENAME}-built
     stop_grace_period: 60s
-    working_dir: "{{ .WebWorkingDir }}"
+    working_dir: "{{ .DBWorkingDir }}"
     volumes:
       - type: "volume"
         source: mariadb-database
@@ -167,6 +167,7 @@ services:
   dba:
     container_name: ddev-${DDEV_SITENAME}-dba
     image: $DDEV_DBAIMAGE
+    working_dir: "{{ .DBAWorkingDir }}"
     restart: "{{ if .AutoRestartContainers }}always{{ else }}no{{ end }}"
     labels:
       com.ddev.site-name: ${DDEV_SITENAME}
--- a/pkg/ddevapp/templates.go
+++ b/pkg/ddevapp/templates.go
@@ -18,6 +18,7 @@ services:
         gid: '{{ .GID }}'
     image: ${DDEV_DBIMAGE}-${DDEV_SITENAME}-built
     stop_grace_period: 60s
+    working_dir: "{{ .WebWorkingDir }}"
     volumes:
       - type: "volume"
         source: mariadb-database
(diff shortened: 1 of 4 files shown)


rfay:
You can use gitpod to review this, https://gitpod.io/#https://github.com/drud/ddev/pull/3160

Or download and install in your path the artifacts, from https://github.com/drud/ddev/actions/runs/1126173412

rfay:
I manually tested all the permutations, and even though this was not planned work, it seems to do fine. When it passes tests it can be pulled.

pkg/ddevapp/templates.go
```diff
@@ -18,6 +18,7 @@ services:
         gid: '{{ .GID }}'
     image: ${DDEV_DBIMAGE}-${DDEV_SITENAME}-built
     stop_grace_period: 60s
+    working_dir: "{{ .WebWorkingDir }}"
```
mstrelan, line 21: Should this be DBWorkingDir?

mstrelan:
Will take a look next week, thanks for this!

pkg/ddevapp/templates.go
rfay: Great catch, thanks. I note that there aren't tests for this, and also found that .DbaWorkingDir needed to be supported.

rfay:
This is not passing [TestCmdWorkingDir](https://buildkite.com/drud/ddev-macos-use-nfsmount/builds/3625#a553f7fc-8b5b-4aca-993b-3efd1e99c733) at least not now. I don't think it has to do with this PR, others are seeing it, but I don't think it can go in until we understand why that test fails.

PR #3169, opened 2021-08-14
Blackfire packages are currently broken, embargo TestExtraPackages, force image removal

rfay:
## The Problem/Issue/Bug:

* Blackfire.io's Debian repositories are currently refusing `apt-get update`, see https://twitter.com/randyfay/status/1426535913094467584
* Image removal is not currently forced. I think it will do no harm to force removal

* Embargo the TestExtraPackages test until 20 AUG 21 10:00 MDT, meaning we won't test it until then, but tests will resume then.
* Change dockerutil.RemoveImage() to force removal. This fixes a random test failure that was identified at the same time as the Blackfire problem.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3169"></a>

### DIFF
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -794,6 +794,11 @@ func TestPHPOverrides(t *testing.T) {
 // TestExtraPackages tests to make sure that *extra_packages config.yaml directives
 // work (and are overridden by *-build/Dockerfile).
 func TestExtraPackages(t *testing.T) {
+	embargoTime := "20 AUG 21 10:00 MDT"
+	if util.IsBeforeCutoffTime(embargoTime) {
+		t.Skipf("Skipping %s until %s", t.Name(), embargoTime)
+	}
+
 	assert := asrt.New(t)
 	app := &DdevApp{}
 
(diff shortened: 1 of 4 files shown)


croane:
I setup deev on a new project this morning and I kept on getting the "payment required" error from balckfire. It is driving me nuts. If you find a temporary workaround, please let me know. Thanks!

rfay:
@croane there are two workarounds:

 1. Temporarily remove your custom .ddev/web-build/Dockerfile or `webimage_extra_packages` in the .ddev/config.yaml
 2. Delete the blackfire config in a custom .ddev/web-build/Dockerfile (or adapt your existing Dockerfile) with something like this:

```dockerfile
# work around blackfire.io problems 
ARG BASE_IMAGE
FROM $BASE_IMAGE
# Remove blackfire apt config while it's broken.
RUN rm /etc/apt/sources.list.d/blackfire.list
RUN apt-get update -qq

# Now add any apt-get installs that you might need
# see https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/#adding-extra-dockerfiles-for-webimage-and-dbimage.
# The example here is just php7.4-tidy
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests php7.4-tidy
```

PR #3209, opened 2021-09-01
Update acquia.yaml.example

greggmarshall:
We noticed the ddev pull acquia was significantly slower than a direct ddev blt drupal:sync, but drupal:sync started failing.  Two suggested optimizations:
1.  In the db pull, we switched from acli remote:drush to the local drush (using it for the files pull), which allows using the --gzip option and downloads directly into db.sql.gz.  On our relatively large DB (about 300 mb) this change reduced the time from about 23 minutes to 15 minutes (using Linux time command)
2.  In the files_pull, we added an --exclude-paths='styles:css:js' to not bother with what are usually Drupal's aggregation temporaries.  We actually added a project specific directory with PDFs to reduce transfer size.  This change reduce the time from about 66 minutes to 43 minutes (there are A LOT of files in the project).  A better optimization would be to not copy the files into the .ddev/.downloads directory then have ddev copy it into the Drupal site directory, but instead do a direct rsync to the Drupal site directory which would avoid a lot of files being copied after the first time but that would require changes to the Golang.....  When we tested that option outside of the pull the rsync time went down to 15 minutes.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3209"></a>

### DIFF
--- a/pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example
+++ b/pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example
@@ -21,7 +21,7 @@
 # 9. Use `ddev pull acquia` to pull the project database and files.
 # 10. Optionally use `ddev push acquia` to push local files and database to Aquia. Note that `ddev push` is a command that can potentially damage your production site, so this is not recommended.
 
-# Debugging: Use `ddev exec acli command` and `ddev exec acli auth:info`
+# Debugging: Use `ddev exec acli command` and `ddev exec acli auth:login`
 # Make sure you remembered to `ddev auth ssh`
 
 environment_variables:
@@ -35,6 +35,8 @@ auth_command:
     ssh-add -l >/dev/null || ( echo "Please 'ddev auth ssh' before running this command." && exit 1 )
     acli auth:login -n --key="${ACQUIA_API_KEY}" --secret="${ACQUIA_API_SECRET}"
     acli remote:aliases:download -n >/dev/null
+    drush core:init -y   # make aliases available to Drush
+    source ~/.bashrc
 
 db_pull_command:
   command: |
@@ -42,8 +44,7 @@ db_pull_command:
     set -eu -o pipefail
     ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
     pushd /var/www/html/.ddev/.downloads >/dev/null
-    acli remote:drush -n ${project_id} -- sql-dump --extra-dump=--no-tablespaces >db.sql
-    gzip db.sql
+    drush @${project_id} sql-dump -n --extra-dump=--no-tablespaces --gzip >db.sql.gz
 
 files_pull_command:
   command: |
@@ -51,7 +52,7 @@ files_pull_command:
     set -eu -o pipefail
     ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
     pushd /var/www/html/.ddev/.downloads >/dev/null;
-    drush -r docroot rsync --alias-path=~/.drush -q -y @${project_id}:%files ./files
+    drush -r docroot rsync --exclude-paths='styles:css:js' --alias-path=~/.drush -q -y @${project_id}:%files ./files
 
 # push is a dangerous command. If not absolutely needed it's better to delete these lines.
 db_push_command:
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example
```diff
@@ -42,15 +42,14 @@ db_pull_command:
     # set -x   # You can enable bash debugging output by uncommenting
     ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
     pushd /var/www/html/.ddev/.downloads >/dev/null
-    acli remote:drush -n ${project_id} -- sql-dump --extra-dump=--no-tablespaces >db.sql
-    gzip db.sql
+    drush @${project_id} sql-dump -n --extra-dump=--no-tablespaces --gzip >db.sql.gz
```
rfay, line 45: Doesnt' this assume you have custom aliases set up for the remote?

rfay (commented):
Thanks for this. We'll see if it can pass the tests. I'm worried that you have custom aliases set up.

pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example
greggmarshall: I suppose it does, but doesn't the acli remote:aliases:download in the auth_command download them?

rfay: If it's just relying on the ones that acli downloads, it will work fine. If you project has custom aliases, it won't work when they're not there.

greggmarshall: Actually it must work since the files_pull_command below, line 52/53, is (and was) using the same alias to get the files.

rfay (changes requested):
I experimented with this for quite a while and was not able to get it to work with anything I tried. Always get

```
        	            	drush @${project_id} sql-dump -n --extra-dump=--no-tablespaces --gzip >db.sql.gz
        	            	)], err='exit status 1', stdout='', stderr=' [preflight] The alias @eeamoreno.dev could not be found.'
```

That was after I had changed it to look in the alias path with `    drush @${project_id} sql-dump --alias-path=~/.drush -n --extra-dump=--no-tablespaces --gzip >db.sql.gz`

Maybe there's something I was doing wrong, but it doesn't pass tests and I'm unable to get it to work fiddling with it. Seems the drush alias isn't found. Perhaps your Acquia environment has aliases that another one doesn't?

greggmarshall:
That's very possible.  I'm pretty sure we had the aliases in the repository.  But i am still confused why the drush command in line 51 works then, its using the same alias from what I can see...

`drush -r docroot rsync --alias-path=~/.drush -q -y @${project_id}:%files ./files`

rfay:
maybe rsync works with it and sql-dump doesn't?

greggmarshall:
I have no idea.  Shall we just cancel this then?

rfay:
If you would be willing to take another shot at it and use the test Acquia credentials I could give them to you privately. You'd want to make `ddev pull acquia` work manually and also make TestAcquiaPull work with those credentials. 

I'm quite sure you know vastly more about drush aliases and remotes and Acquia than I do, so it's a great honor to have you working on this, and if you could get this solved it would be great. So let me know if you'd like the credentials. I'll be happy to give them to you privately in slack or whatever.

rfay:
I may be wrong but... https://github.com/drud/ddev/issues/3237

I may have had mutagen enabled globally when testing this. I'll recheck when that goes in or before.

greggmarshall:
Ok, if not, send me the credentials via DM on Slack and I'll try to figure out why it works in one case but not another.  I might need help understanding how to run the tests.

pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example
greggmarshall: I can see one possible issue.   After doing a `ddev pull acquia`, I tried running `ddev drush sa` and got  `[success] No site aliases found.`

Then I tried running `ddev exec acli remote:aliases:download -n` to see the output
```
Drush changed how aliases are defined in Drush 9. Drush 8 aliases are PHP-based, while Drush 9+ aliases are YAML-based.
Cloud Platform Drush aliases installed into /home/gregg/.drush/sites
Drush 9+ does not automatically read aliases from this directory. Run drush core:init to ensure these aliases are discovered.
For more details, see https://github.com/drush-ops/drush/blob/master/examples/example.site.yml
```

So I'm guessing I'll need to add `drush core:init -y` and probably `source ~/.bashrc `so the aliases downloaded show up

rfay (approved):
This looks like a significant improvement to me, thanks so much!

PR #3211, opened 2021-09-02
Provide TYPO3 and Webserver specific help

gilbertsoft:
## The Problem/Issue/Bug:

Some questions show up regulary so these should be covered by the docs.

### Provide TYPO3 Base Variant example

This patch adds an example on how to add a Base Variant to TYPO3 for the local ldevelopment with DDEV which is often asked for.

### Provide Apache TLS redirect example

Because of the proxy setup used with DDEV normal redirect examples from plain HTTP to TLS won't work so and extended example is provided.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3211"></a>

### DIFF
--- a/docs/users/topics/cms_specific_help.md
+++ b/docs/users/topics/cms_specific_help.md
@@ -8,6 +8,28 @@
 
 * **Settings Files**: On `ddev start`, DDEV creates a `public/typo3conf/AdditionalConfiguration.php` with database configuration in it.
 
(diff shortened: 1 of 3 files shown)


rfay:
Thanks!

rfay (approved):
Thanks so much for this. Looks great to me!

PR #3213, opened 2021-09-03
Document how to restore DDEV management to a file

mlncn:
## The Problem/Issue/Bug:

DDEV warns you when a file is not managed by it anymore but does not indicate how to restore management, with the exception of a closed issue https://github.com/drud/ddev/issues/468 indicating how to take a file *out* of DDEV management.

Adds documentation (so searching for the message will bring you to a documentation page) but does *not*, as yet, add a link in the `ddev start` messages.

N/A documentation only.

N/A documentation only.

## Related Issue Link(s):

https://github.com/drud/ddev/issues/468

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3213"></a>

### DIFF
--- a/docs/users/topics/cms_specific_help.md
+++ b/docs/users/topics/cms_specific_help.md
@@ -11,3 +11,20 @@
 ### Running any PHP App with DDEV
 
 Nearly any PHP app will run fine with DDEV, and lots of others. If your project type is not one of the explicitly supported project types, that's fine. Just set the project type to 'php' and go about setting up settings files or .env as you normally would.
+
+### DDEV-managed Files
+
+These files created by DDEV are also updated and managed by DDEV by default.  If this is changed, you will receive the message "settings.ddev.php already exists and is managed by the user."
+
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
Thanks so much for this!

Actually, the best way to work around this is just to delete the file in question and do a `ddev start` and ddev will recreate it. That will make this section much simpler. Feel free to push another commit with that suggestion. However, this isn't CMS-specific. It probably belongs in another file in the same directory, call it hints_and_tips.md and add it to mkdocs.yml.

rfay:
On second thought, can we just add this to https://ddev.readthedocs.io/en/latest/users/topics/settings_files/ ?

mlncn:
Sure!  I don't think i found that page :sob:, but as long as the warning messages are added there people will find it!

rfay:
I added this in https://github.com/drud/ddev/commit/9cc8fe4c5f2a540061d117901013ceade2af3d87 - please see if you like it and it solves the problem you were raising. Thanks @mlncn !

Issue #3216, opened 2021-09-03
`ddev pull` changes public/fileadmin to a directory (project type: typo3) 


jonaseberle:
**Run a Diagnostic and Paste Link Here**
https://gist.github.com/jonaseberle/01aa87572b248d093109180e997d04b1

**Describe the bug**

In my case, `public/fileadmin` and `public/uploads` are symlinks.
After running `ddev pull`, the symlink `public/fileadmin` is replaced by an empty directory.

**To Reproduce**

```sh
ddev config --docroot=public --create-docroot --project-type=typo3

mkdir -p data/files/fileadmin data/files/uploads
ln --symbolic --relative data/files/* public/

ddev start

cat > .ddev/providers/test.yaml <<EOF
files_pull_command:
  command:  
EOF

ddev pull test --skip-db -y

# the former symlink `fileadmin` is now a folder
ls -als public/
```

**Expected behavior**

The symlink stays in place

**Version and configuration information (please complete the following information):**

- Host computer OS and Version: Linux (Ubuntu)
- ddev version information: v1.18.0-rc1

rfay:
Thanks for the report. This is an unusual situation, right? What should the correct behavior be, just follow the symlink (if it's relative and also inside the filespace?) Pretty sure this behavior is a result of the lower-level utilities we use to copy directories, and I think they ignore symlinks.

Would it be possible to share your pull yaml?

jonaseberle:
Hi Randy, the content of the pull yaml:

The used commands are those from ddev-contrib.
These commands don't do anything with `public/fileadmin`

```
environment_variables:
  target: live

auth_command:
  command: |
    ssh-add -l >/dev/null || ( echo "Please 'ddev auth ssh' before running this command." && exit 1 )

db_pull_command:
  command: |
    # Dumps and downloads from server
    /var/www/html/.ddev/commands/web/dump-db "${target}" -y
    # ddev only supports single-file-gzipped here currently so we gzip the downloaded SQL for ddev to consume it...
    cat /var/www/html/data/sql/struct/* /var/www/html/data/sql/records/* | gzip > /var/www/html/.ddev/.downloads/db.sql.gz

files_pull_command:
  command: |
    # This rsyncs to ./data/files/fileadmin and ./data/files/uploads because the default ddev location
    # is cleared before pulling
    # in our case, ./data/files/* is symlinked to appropriate locations in a post-start hook
    /var/www/html/.ddev/commands/web/dump-files "${target}" -y
```

[...]

PR #3223, opened 2021-09-05
Make more robust and flexible options for how importing works in `ddev pull`

jonaseberle:
## The Problem/Issue/Bug:
Previously it was not possible to disable automatic imports by configuration.

This adds a <provider>.yaml option `skip_import` (bool).
When `false` (default), "ddev pull" automatically imports downloaded artefacts
from `.ddev/.downloads/`.
When `true`, "ddev pull" will not act upon artefacts. In that case it is the
command's responsibility to take all necessary steps to make them available
in the ddev environment.

Setting this for both `files_pull_command` and `db_pull_command` is functionally
equivalent to using "ddev pull ... --skip-import".

This also changes when the message "... downloaded to: ..." is shown:
If either the command line argument "--skip-import" has been given or any of the
newly introduced <provider>.yaml options are set, the message is not shown.

```
cat > .ddev/providers/test.yaml <<EOF
db_pull_command:
  command: |
    echo | gzip > /var/www/html/.ddev/.downloads/db.sql.gz
    echo "[db with default] After this message is shown, ddev will output 'Database downloaded to: ...' and 'Importing database...'"

files_pull_command:
  skip_import: true
  command: |
    echo "[files with skip_import:true] After this message is shown, ddev will output 'Skipping default file locations import.'"EOF
EOF

ddev pull test -y
```

Sorry, none so far. 

## Related Issue Link(s):
#3216

## Release/Deployment notes:

none

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3223"></a>

### DIFF
--- a/pkg/ddevapp/provider.go
+++ b/pkg/ddevapp/provider.go
@@ -15,11 +15,13 @@ import (
 
 // ProviderCommand defines the shell command to be run for one of the commands (db pull, etc.)
 type ProviderCommand struct {
-	Command string `yaml:"command"`
-	Service string `yaml:"service,omitempty"`
+	Command    string `yaml:"command"`
+	Service    string `yaml:"service,omitempty"`
+	SkipImport bool   `yaml:"skip_import"`
 }
 
 // ProviderInfo defines the provider
+// @todo: CodePullCommand unused
 type ProviderInfo struct {
 	EnvironmentVariables map[string]string `yaml:"environment_variables"`
 	AuthCommand          ProviderCommand   `yaml:"auth_command"`
@@ -91,11 +93,10 @@ func (app *DdevApp) Pull(provider *Provider, skipDbArg bool, skipFilesArg bool, 
 			return err
 		}
 
-		output.UserOut.Printf("Database downloaded to: %s", fileLocation)
-
-		if skipImportArg {
-			output.UserOut.Println("Skipping database import.")
+		if skipImportArg || provider.DBPullCommand.SkipImport {
+			output.UserOut.Println("Skipping default database import.")
 		} else {
+			output.UserOut.Printf("Database downloaded to: %s", fileLocation)
 			output.UserOut.Println("Importing database...")
 			err = app.ImportDB(fileLocation, importPath, true, false, "db")
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/provider.go
```diff
@@ -15,11 +15,13 @@ import (
 
 // ProviderCommand defines the shell command to be run for one of the commands (db pull, etc.)
 type ProviderCommand struct {
-	Command string `yaml:"command"`
-	Service string `yaml:"service,omitempty"`
+	Command    string `yaml:"command"`
+	Service    string `yaml:"service,omitempty"`
+	SkipImport bool   `yaml:"skip_import"`
```
rfay, line 20: Wouldn't it make sense to add a files_import_command and a db_import_command, which could be empty?

I guess that leads to potential backward-compatibility problems, but it's worth thinking about.

Perhaps add an import-api version to solve that?

Overall, I think the actual db import logic and files import logic should be moved into the yaml file. What you you think of that?

rfay (commented):
Thanks for the good thinking about this. Haven't read everything yet, but here's a thought.

pkg/ddevapp/provider.go
jonaseberle: I understand you mean the actual last "import into ddev" step, do you?

I wonder if it's worth it to provide an API catering for all scenarios like multiple databases/file locations and multiple artifact formats.

btw: I am not sure I defined that correctly here. What I wanted to achieve: `SkipImport` shall be an optional boolean, defaulting to `false`. I just found out through testing that it worked like that.

rfay: Right, I'm saying actually define the import-into-ddev (db and files) steps as separate items. 

Skipping the import isn't exactly what we're trying to accomplish, right? We're trying to *delegate* the import. So I'm saying why not just always delegate it to a script?

rfay: I like where you're going... and the "leave out for default" solves so many problems.

I agree that pull implies everything... but `pull` has a long history in ddev and it wouldn't probably be worth changing it.

jonaseberle:
I can't promise anything in the next 2 weeks and am not fluent in Go either. I am happy to test everything you throw at me ;)

rfay:
Totally understood, but your thinking here is much appreciated.

rfay:
Closing in favor of #3373 - hope you'll be able to take a look at that @jonaseberle.

rfay:
This can be tested in gitpod using https://gitpod.io/#https://github.com/drud/ddev/pull/3223

Artifacts can be downloaded for local experimentation at https://github.com/drud/ddev/actions/runs/1469618009

PR #3243, opened 2021-09-13
Improve table formatting and presentation, fixes #3242

### LINKED ISSUE #3242: Suggestions for the `ddev describe` and `ddev list` output
I've taken a look at the improved and updated output of `ddev describe` and `ddev list` in DDEV 1.18.0-rc2. I've noticed one or two details which might be discussed or perhaps improved further. 

**_ddev described_**

ddev_describe

-  would it make sense to increase the width of the `SERVICE` column in the `describe.go` file in line 80 from `WidthMax: 8` to `WidthMax: 10` so there is no wrapping for the listed default containers provided by DDEV? 

-  would it make sense to switch the position of the headline row (SERVICE, STAT, URL/PORT, INFO) and the one currently underneath (Project, name,path)? Basically the Project row provides the context for the whole `describe` output underneath. It might be a better fit to place it as the title of the output on top of the table like: 

Project: danse ( ~/Sites/danse )

**_ddev list_**
ddev_list

-  you have the first row with NAME, TYPE, LOCATION, URL and STATUS. The title is written in capital letters. If you take now a look at the last row you also have capital letters. that gives the initial visual queue  that the last row would consist of  titles as well. often longer tables are bordered/cased by title rows but in this case ROUTER is an actual container like the others listed above. so why not change ROUTER to router and also move HEALTHY to the STATUS column and rename it to `running (health)`. Semantically it would be also a better fit there. Healthy is more of a status than a location.

rfay:
## The Problem/Issue/Bug:

* #3242

Fixes the items listed there.

Try it out on gitpod or locally.

Screen Shot 2021-09-13 at 8 06 25 AM

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3243"></a>

### DIFF
--- a/cmd/ddev/cmd/describe.go
+++ b/cmd/ddev/cmd/describe.go
@@ -77,7 +77,7 @@ func renderAppDescribe(app *ddevapp.DdevApp, desc map[string]interface{}) (strin
 		t.SetColumnConfigs([]table.ColumnConfig{
 			{
 				Name:     "Service",
-				WidthMax: 8,
+				WidthMax: 10,
 			},
 			{
 				Name:     "URL/Port",
@@ -89,8 +89,8 @@ func renderAppDescribe(app *ddevapp.DdevApp, desc map[string]interface{}) (strin
 			},
 		})
 	}
+	t.SetTitle(fmt.Sprintf("Project: %s %s %s", app.Name, desc["shortroot"].(string), app.GetPrimaryURL()))
 	t.AppendHeader(table.Row{"Service", "Stat", "URL/Port", "Info"})
-	t.AppendRow(table.Row{"Project", "", app.Name, desc["shortroot"].(string)})
 
 	// Only show extended status for running sites.
 	if status == ddevapp.SiteRunning {
@@ -164,7 +164,7 @@ func renderAppDescribe(app *ddevapp.DdevApp, desc map[string]interface{}) (strin
 				k = "PHPMyAdmin"
 				urlPortParts = append(urlPortParts, "`ddev launch -p`")
 			}
-			t.AppendRow(table.Row{k, v["status"], strings.Join(urlPortParts, "\n"), strings.Join(extraInfo, "\n")})
+			t.AppendRow(table.Row{k, ddevapp.FormatSiteStatus(v["status"]), strings.Join(urlPortParts, "\n"), strings.Join(extraInfo, "\n")})
 		}
 
 		if !ddevapp.IsRouterDisabled(app) {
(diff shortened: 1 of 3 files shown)


rpkoller:
I've taken a look at your pull request on Gitpod. :) 

- extending the width of the service column was the right step imho 
- moving the project row to the top is also an improvement. the only nitpick i might have is currently the row looks like that: `Project: simpleproj /workspace/simpleproj https://simpleproj.ddev.site:8443          ` but i think there is not really a need for the link `https://simpleproj.ddev.site:8443`  in the top row. the link is displayed often enough in the table cells underneath but the filesystem path it is not so `Project: simpleproj /workspace/simpleproj` might be enough? 
- and about the mixed case table headers. i think for the table headers it is totally fine that they remain in capital letters. in capital letters they are easier to distinguish from regular table data. with the case i referred mainly to ROUTER and HEALTHY implying those would be titles as well.  
 
p.s. one detail which is probably gitlab related but somehow the decender of the used typeface gets cut? is that a known thing? 
Bildschirmfoto 2021-09-13 um 16 31 20

rpkoller:
ahhhh and `ddev list` got also updated. 
Bildschirmfoto 2021-09-13 um 16 41 16

same like for `ddev describe` i think the table header would be ok or even easier to distinguish in capital letters. 
and it looks good that the table footer (router) is displayed in lower case. the only nitpick i would have. would it be possible that `healthy` is displayed in the `status` column instead of the `location` one?

rfay:
Thanks, I've updated with some of your requests.

The router status is in the "Location" column because sometimes it spits out a huge amount of stuff and that column adapts better. 

I don't understand about the descender being cut off, I imagine that has to do with the environment you're running with. Nothing about this sets the font.

rpkoller:
cool thanks!

ahhh and i see. if there is much output for the router status then it makes sense to use the cell with more available width. thought the output would be just plain `healthy` then it would been a better fit semantically in the status column. but if there could be more output then its the right choice you took

hmmm and about the environment. i've just clicked the gitpod icon in the pull request in firefox. that is the output out of the box no changes nothing. maybe i should post that screenshot  in the drupalpod channel?

rfay:
I don't see the problem with descenders on my mac. You can download and experiment on your machine at https://github.com/drud/ddev/actions/runs/1231131984

rpkoller:
on my mac i havent had the descender issue so far. will check out the build on my machine if a high sierra one is available. but it seems to me that it is more of a gitpod exclusive issue here imho. will check it out.

rpkoller:
tested with the build from the pull request. the display is correct no issues with the descender: 
Bildschirmfoto 2021-09-13 um 23 59 24
as expected that seems an gitpod issue or better an issue with the gitpod settings?

PR #3247, opened 2021-09-16
Ensure monitoring is accessible also with basic authentication

gilbertsoft:
## The Problem/Issue/Bug:
With Nginx if basic authentication is enabled the monitoring endpoints are not longer accessible and this cannot be changed in a custom configuration because the locations are already defined.

For Apache this is not an issue because an alias is used for `/phpstatus`.

In addition, this need uncovered problems in how apache is run via supervisord, and those had to be fixed. This was originally https://github.com/drud/ddev/pull/3249 but for practicality had to be rolled in here. 

* To avoid this problem this patch sets `auth_basic` always to `off` for phpstatus
* Fix a bug from #3249 where `child_exit_monitor entered FATAL state, too many start retries too quickly` showed up in the logs at startup in v1.18 (using nonexistent python (python2) in /usr/local/bin/kill_supervisor.py)
* Fix a bug where reloads/restarts of apache were not successful when done by supervisord

Enable auth_basic with a custom config `.ddev/nginx/auth.conf`:

```
auth_basic		"auth test DDEV";
auth_basic_user_file	/mnt/ddev_config/nginx/htpasswd;
```

Create a `.ddev/nginx/htpasswd` file using apache2-utils see https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/.

Access the webroot with `ddev launch` and an authentication dialog should be shown, abort the authentication. Now check the two monitoring end points to be accessible without authentication. ATTENTION: the browser remembers a basic auth so once authenticated you are not able to log off without closing the browser.

Also check the logs for errors with `ddev logs`.

Added a full extra stanza to container testing for this situation, which covers both apache and nginx.

## Related Issue Link(s):

Some of what gets fixed here was first attempted in #3249, but unfortunately they were too closely related to keep them separate.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3247"></a>

### DIFF
--- a/containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats
+++ b/containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats
@@ -127,3 +127,28 @@
   run docker exec -t $CONTAINER_NAME disable_xdebug
   run docker exec -t $CONTAINER_NAME disable_xhprof
 }
+
+@test "verify htaccess doesn't break ${WEBSERVER_TYPE} php${PHP_VERSION}" {
+  docker cp tests/ddev-webserver/testdata/nginx/auth.conf ${CONTAINER_NAME}:/etc/nginx/common.d
+  docker cp tests/ddev-webserver/testdata/nginx/junkpass ${CONTAINER_NAME}:/tmp
+  docker cp tests/ddev-webserver/testdata/apache/auth.conf ${CONTAINER_NAME}:/etc/apache2/conf-enabled
+  # Reload everything
+  docker exec ${CONTAINER_NAME} kill -HUP 1
+  sleep 4
+  # Make sure we can hit /phpstatus without auth
+  run curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:$HOST_HTTP_PORT/phpstatus
+  echo "# phpstatus status=$output"
+  [ "$status" = 0 ]
+  [ "$output" = "200" ]
+  curl --fail -s http://127.0.0.1:$HOST_HTTP_PORT/phpstatus | egrep "idle processes|php is working"
+  # Make sure the auth requirement is actually working
+  curlstmt="curl --fail -s -o /dev/null -w "%{http_code}" http://127.0.0.1:$HOST_HTTP_PORT/test/phptest.php"
+  run ${curlstmt}
+  [ "$output" = "401" ]
+
+  # Make sure it works with auth when hitting phptest.php
+  AUTH=$(echo -ne "junk:junk" | base64)
+  curl --fail --header "Authorization: Basic $AUTH" 127.0.0.1:$HOST_HTTP_PORT/test/phptest.php
+  docker exec ${CONTAINER_NAME} rm /etc/nginx/common.d/auth.conf /etc/apache2/conf-enabled/auth.conf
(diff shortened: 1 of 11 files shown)


rfay:
Artifacts to test this will be available at https://github.com/drud/ddev/actions/runs/1242263546 in a couple of minutes.

This probably needs a couple of container-level tests, or maybe just the one will do it.

gilbertsoft:
> 
> This probably needs a couple of container-level tests, or maybe just the one will do it. Perhaps something added to https://github.com/drud/ddev/blob/master/containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats, which would get all permutations. It's not easy to add a mount for one test and not the others though. But putting it there would get all the permutations. I think maybe just writing the config and reloading internally would be fine, then make sure to remove the config. (I can do it)

I have currently no idea about the container tests, if you already know what's to do feel free to do so. I would first have to invest some time to it. But if you do and once it's finished I like to learn from your changes, so please ping me at Slack...

rfay:
@gilbertsoft would appreciate if you could review this one as it is now, thanks.

rfay:
Artifacts for testing are at https://github.com/drud/ddev/actions/runs/1246233873

Or you can use gitpod, https://gitpod.io/#https://github.com/drud/ddev/pull/3247

containers/ddev-webserver/tests/ddev-webserver/testdata/nginx/auth.conf
```diff
@@ -0,0 +1,2 @@
+auth_basic   "auth test DDEV";
```
gilbertsoft, line 1: Lets use the same value here as for apache
```suggestion
auth_basic   "Restricted Files";
```

containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats
```diff
@@ -127,3 +127,32 @@
...
+  # Reload webserver
+  if [ "${WEBSERVER_TYPE}" = "apache-fpm" ]; then
+    docker exec ${CONTAINER_NAME} apache2ctl -k graceful
+  else
+    docker exec ${CONTAINER_NAME} nginx -s reload
+  fi
+  sleep 2
+  # Make sure we can hit /phpstatus without auth
+  run curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:$HOST_HTTP_PORT/phpstatus
+  echo "# phpstatus status=$output"
```
gilbertsoft, line 144: Where does this $output come from? Same for $status below.

gilbertsoft (commented):
Looks good! One little suggestion and two questions...

containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats
rfay: The bats `run` command creates $STATUS (the return code of the command) and $OUTPUT (The actual output). So in this case we're making curl output the http status code and gathering it in $OUTPUT.

gilbertsoft: Okay, was my assumption. Some hidden magic :)

PR #3249, opened 2021-09-16
Fix Debian 11 bullseye problems with supervisord

rfay:
## The Problem/Issue/Bug:

There were a couple of problems with the startup of the web container with Debian 11 Bullseye. They weren't really obvious if you weren't watching the logs.

1. `child_exit_monitor entered FATAL state, too many start retries too quickly`
2. A reload of supervisord reported a failure (but error 0) restarting apache

1. The problem with `child_exit_monitor` was that the script used for this (/usr/local/bin/kill_supervisord.py) was set to use `#!/usr/bin/env python` and that referenced python2, which is gone from Debian Bullseye. This changes that to `#!/usr/bin/env python3`
2. Apache restart (when doing `kill -1 1` to get supervisord to restart did not work because it returns right away, so this changes to `startsecs=0` to allow `apachectl` to return right away. This was almost certainly this way since the beginning of apache support in ddev.

- [ ] `ddev start` and watch `ddev logs`. There shouldn't be failures of child_exit_monitor
- [ ] After starting with apache-fpm, `ddev exec kill -1 1` to reload supervisord, you should see it reload with no errors

I'm not sure this needs any particular automated testing

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3249"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/conf.d/supervisor.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/conf.d/supervisor.conf
@@ -1,7 +1,7 @@
 [supervisord]
 logfile=/var/log/supervisord.log ; (main log file;default $CWD/supervisord.log)
 loglevel=info                ; (log level;default info; others: debug,warn,trace)
-pidfile=//var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
+pidfile=/var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
 nodaemon=false               ; (start in foreground if true;default false)
 
 [eventlistener:child_exit_monitor]
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/apache.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/apache.conf
@@ -8,3 +8,4 @@ stdout_logfile=/proc/self/fd/2
 stdout_logfile_maxbytes=0
 redirect_stderr=true
 exitcodes=0,1
+startsecs=0 # See https://github.com/Supervisor/supervisor/issues/212#issuecomment-47933372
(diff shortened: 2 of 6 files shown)


containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/conf.d/supervisor.conf
```diff
@@ -1,7 +1,7 @@
 [supervisord]
 logfile=/var/log/supervisord.log ; (main log file;default $CWD/supervisord.log)
 loglevel=info                ; (log level;default info; others: debug,warn,trace)
-pidfile=//var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
+pidfile=/var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
```
gilbertsoft, line 4: Same also in the prod folder

gilbertsoft (changes requested):

containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/conf.d/supervisor.conf
rfay: Of course `//var` or `//anything` or `/anything//something` is all valid in Linux and most linux-like systems.

rfay: But it also needed the apache fix in prod, so thanks!

gilbertsoft: Yeah I know, it's just superfluous like written in the Slack thread.

gilbertsoft (approved):

rfay:
Closing this in favor of #3247 - I wasn't able to complete that one without incorporating the changes from this PR.

PR #3261, opened 2021-09-24
Clarify usage of the "solr path" setting in implementation details

damienmckenna:
## The Problem/Issue/Bug:
Different systems use Solr API v1 and others use v2.

Noted that the "Solr path" setting must be set to "/solr" when using Solr API v1 and just "/" when using Solr API v2.

## Related Issue Link(s):
https://drupal.slack.com/archives/C5TQRQZRR/p1632433128297900

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3261"></a>


rfay (approved):
Thanks!

PR #3262, opened 2021-09-24
Add topic on PhpStorm integration, fixes #3130 [skip ci]

### LINKED ISSUE #3130: How to use PhpStorm on Windows WSL2 with full integration including tests
I try to run ddev as good as possible integrate in PHPStorm on Windows using WSL and Docker Desktop. Sadly PHPStorm can't use the generated docker-compose file for integration of a remote interpreter on Windows (WSL) because the absolute paths for the volumes can not be mapped corrrectly.
See this issue for more information:
https://youtrack.jetbrains.com/issue/WI-61699

A solution for this issue would be to use relative paths in the generated docker-compose file as shown by Eugene Morozov in the video you find attached to the ticket.

An alternative / workaround might be a post-x script that is manipulating the docker-compose.yaml after its generation.
Also there is a workaround Alexandru found, that you can mount your WSL as a Network Drive in Windows and reference the docker-compose.yml by this path.

If there is no need for the absolute path I would really appreciate marking them relative. Of course I am also interested in any other solution that does not require much efford.

rfay:
## The Problem/Issue/Bug:

This adds a topic to the docs about PhpStorm integration. 

Would very much appreciate feedback from wonderful contributors @eojthebrave @susannemoog @nico-loeber @deviantintegral @cmuench @SuddenlyRust @mstrelan @tyler36 

Of course we can improve this as time goes on, counting on all of you.


cmuench:
@rfay Thanks for providing the docs. It looks very complete to me. We are already testing it. I will try to collect the feedback of my colleagues, too.
BTW... have an idea. Maybe it's possible to generate the PhpStorm/Idea config by a new ddev command. It's only an idea but in the past I created already a command as composer plugin to generate PhpStorm settings for PSR/4 mappings by writing XML config files in the .idea directory.
In our case the config is stored in the `.idea/php.xml` file.

rfay:
@cmuench that's great thinking about generating the config, and it would work well with a custom command that could be community-maintained.

The biggest problem I see on a first look is that the "Server" is actually in global .idea, and it has the all-important path mappings in it.  Interesting that it's called `DOCKER_ACCOUNT_NAME`

```xml
        <remote_data INTERPRETER_PATH="php" HELPERS_PATH="/opt/.phpstorm_helpers" INITIALIZED="false" VALID="true" RUN_AS_ROOT_VIA_SUDO="false" DOCKER_ACCOUNT_NAME="d9 DDEV Integration" DOCKER_COMPOSE_SERVICE_NAME="web" DOCKER_REMOTE_PROJECT_PATH="/opt/project">
```

That Server file mapping doesn't seem to be in the php.xml, not sure where it's hidden. 

Anyway, if you want to take a crack at a custom command it has potential. I imagine they'd break it periodically. The current process is awfully involved for such a simple thing.

mstrelan:
Is the server only in the global config if you have the "Shared" option checked?

rfay:
@mstrelan good question... but I don't know where the "shared" checkbox is. Can you give a screenshot or something?

mstrelan:
On my phone at the moment, from memory when you add the server in phpstorm
is in the top right of the dialog box.

mstrelan:
See right at the bottom of https://www.jetbrains.com/help/phpstorm/servers.html#configurationDetails

rfay:
Thanks, that's the debug server, and I've never had it set to shared, as the default is not.  The "server" I mentioned in https://github.com/drud/ddev/pull/3262#issuecomment-927118701 is the Docker server. There's a lot of ambiguity in their teminology!

deviantintegral:
> Maybe it's possible to generate the PhpStorm/Idea config by a new ddev command.

I'm doing this on a project now, but we're creating configuration in the UI and then adding it to the repo. We ended up moving away from directly committing files in `.idea`, to copying them from a template folder in a composer post-install hook. The main issues we ran into was that it was possible for `composer` operations to change include paths, which are stored along side the PHP interpreter info.

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+2. Set up your project to do normal Xdebug, as described in the [Step Debugging section](../step_debugging.md). This will result in a PhpStorm "Server" with the proper name, normally the same as the FQDN of the project. In this example, "d9.ddev.site". (All you have to do here is click the little telephone to "Start listening for PHP Debug Connections", then `ddev xdebug on`, then visit a web page and choose the correct mapping from host to server. )
+3. Under File→Settings→PHP (Windows) or Preferences→PHP (macOS), click the "..." to the right of "CLI Interpreter"
+    1. Use the "+" to select "From Docker, Vagrant, VM..."
+    2. Choose "Docker Compose"
+    3. Create a server; the default name is "docker", but since the "server" for each project will be different, name it for the project, for example "DDEV d9". Choose "Docker for Windows" or "Docker for Mac"
+    4. In the "Path mappings" of the "Server" map the local paths (which on WSL2 means /home/...) to the in-container paths. So "Virtual Machine Path" would be "/var/www/html" and "Local path" would be something like /Users/rfay/workspace/d9 (on macOS) or \\wsl$\Ubuntu\home\rfay\workspace\d9 on Windows using WSL2
```
deviantintegral, line 22: When you set the compose file in step 5, aren't the path mappings set up automatically? So far they have been for us on macOS - we won't be trying things out on the Windows side for a while, if ever.

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+    5. Now back in the "Configure Remote PHP Interpreter" for "Configuration files" use .ddev/.ddev-docker-compose-full.yaml
+    6. Service: web
+    7. Add an environment variable `COMPOSE_PROJECT_NAME=ddev-<projectname>`. In my case, it's `ddev-d9`
```
deviantintegral, line 25:
```suggestion
    7. Add an environment variable `COMPOSE_PROJECT_NAME=ddev-<projectname>`. In this case, it's `ddev-d9`
```

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+    8. In the CLI interpreter "Lifecycle" select "Connect to existing container"
+    9. In the PHP Interpreter path, you can just put "php" if you're using the default PHP version (currently 7.4). Due to a [PhpStorm bug](https://youtrack.jetbrains.com/issue/WI-62463) you'll want to put the full name of the binary, like "php8.0" if you're not using the default version.
```
deviantintegral, line 27:
```suggestion
    9. In the PHP Interpreter path, you can just put `php` if you're using the default PHP version (currently 7.4). Due to a [PhpStorm bug](https://youtrack.jetbrains.com/issue/WI-62463) you'll want to put the full name of the binary, like `php8.0` if you're not using the default version.
```

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+3. Under File→Settings→PHP (Windows) or Preferences→PHP (macOS), click the "..." to the right of "CLI Interpreter"
+    1. Use the "+" to select "From Docker, Vagrant, VM..."
+    2. Choose "Docker Compose"
+    3. Create a server; the default name is "docker", but since the "server" for each project will be different, name it for the project, for example "DDEV d9". Choose "Docker for Windows" or "Docker for Mac"
+    4. In the "Path mappings" of the "Server" map the local paths (which on WSL2 means /home/...) to the in-container paths. So "Virtual Machine Path" would be "/var/www/html" and "Local path" would be something like /Users/rfay/workspace/d9 (on macOS) or \\wsl$\Ubuntu\home\rfay\workspace\d9 on Windows using WSL2
+    5. Now back in the "Configure Remote PHP Interpreter" for "Configuration files" use .ddev/.ddev-docker-compose-full.yaml
```
deviantintegral, line 23:
```suggestion
    5. Now back in the "Configure Remote PHP Interpreter" for "Configuration files" use `.ddev/.ddev-docker-compose-full.yaml`
```

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+    6. Service: web
+    7. Add an environment variable `COMPOSE_PROJECT_NAME=ddev-<projectname>`. In my case, it's `ddev-d9`
+    8. In the CLI interpreter "Lifecycle" select "Connect to existing container"
+    9. In the PHP Interpreter path, you can just put "php" if you're using the default PHP version (currently 7.4). Due to a [PhpStorm bug](https://youtrack.jetbrains.com/issue/WI-62463) you'll want to put the full name of the binary, like "php8.0" if you're not using the default version.
+    10. Here's an example filled out ![example configuration](images/cli_interpreter.png)
+4. In the main PHP setup dialog, add an entry to the path mappings, as it doesn't correctly derive the full path mapping. Add an entry that maps your project location to /var/www/html. So in this example, the Local Path is /Users/rfay/workspace/d9 and the Remote Path is /var/www/html. ![example mapping](images/mapping.png)
```
deviantintegral, line 29: I think all paths should be codefenced, but I'll stop calling them out now.

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+5. Configure composer under PHP→Composer.
+    - Use "remote interpreter"
+    - CLI Interpreter will be "web"
+6. Under "Test Frameworks" click the "+" to add phpunit
+    - PHPUnit by remote interpreter
+    - Interpreter "web"
+    - Choose "Path to phpunit.phar" and use /var/www/html/vendor/bin/phpunit (or wherever your phpunit is inside the container) (You need phpunit properly composer-installed for your CMS. For example, for Drupal 9, `ddev composer require --dev --with-all-dependencies drupal/core-dev:^9`  and `ddev composer require --dev phpspec/prophecy-phpunit:^2`
```
deviantintegral, line 36: Is this better than using "Use composer autoloader" and putting in vendor/autoload.php?

docs/users/topics/phpstorm.md
```diff
@@ -0,0 +1,50 @@
...
+    - Default configuration file: /var/www/html/web/core/phpunit.xml or wherever yours is inside the container.
```
deviantintegral, line 37: Related: https://youtrack.jetbrains.com/issue/WI-39137

Would it be better to default to `phpunit.xml.dist`, given that's likely to exist?

deviantintegral (commented):
This is great! It very nearly matches what I've been finding over the past two weeks.

docs/users/topics/phpstorm.md
rfay: Oh, I was probably testing with mutagen enabled, which would prevent PhpStorm from being able to derive the main mount.

rfay:
```suggestion
    4. In the "Path mappings" of the "Server" you may have to map the local paths (which on WSL2 means /home/...) to the in-container paths, especially if you have mutagen enabled. So "Virtual Machine Path" would be "/var/www/html" and "Local path" would be something like /Users/rfay/workspace/d9 (on macOS) or \\wsl$\Ubuntu\home\rfay\workspace\d9 on Windows using WSL2.
```

docs/users/topics/phpstorm.md
rfay: I appreciated the ones you did :)

docs/users/topics/phpstorm.md
rfay: When I tested this time, I couldn't get "Use composer autoloader" to work, but I have seen it work before.

rfay: But there's nothing wrong with the other way, of course... if it works.

docs/users/topics/phpstorm.md
rfay: I was just assuming it couldn't be used because of the missing settings in it? I'm no expert on this stuff.  I imagine that anybody who's doing this already better understand what they want.

rfay:
@deviantintegral it would be absolutely fantastic if you'd be willing to share what your team is doing to create the config. Maybe in ddev-contrib or on Stack Overflow. Or a blog post. Or whatever.  I think it would be a real step forward, and could be incorporated here eventually.

cmuench:
Maybe we should shift the discussion here to a new Github issue.

rfay:
Yeah, let's put discussion of generated config in a new issue, thanks!

docs/users/topics/phpstorm.md
nico-loeber: Using the autoloader did work well for me. 
Maybe adding some sort of FAQ section instead with "PHPStorm can not auto detect my phpunit version" and as fix these steps?

rfay: Suggestions welcome. (Use the `Insert a suggestion` on comments.). Not sure we have to solve every problem or provide every option.

docs/users/topics/phpstorm.md
nico-loeber: This step is optional I think. If I skip this step PHPStorm seems to autoamtically check for a phpunit.xml or if not present for a phpunit.xml.dist.

docs/users/topics/phpstorm.md
deviantintegral: That would make sense! Also, I think path mappings will be preserved if you switch from a normal mount to using mutagen on an existing project.

docs/users/topics/phpstorm.md
deviantintegral: I think the current steps are fine.

rfay: I think it's obvious this will evolve as time goes on, it doesn't have to be perfect right now, and people can contribute improvements as we get experience. And as PhpStorm fixes things up.

docs/users/topics/phpstorm.md
deviantintegral: When I remove it I get:

> Error: Default Configuration file is not specified in PHPUnit configuration. Press 'Fix' to edit PHPUnit configuration.

Screen Shot 2021-09-27 at 2 47 48 PM

deviantintegral:
> it would be absolutely fantastic if you'd be willing to share what your team is doing to create the config. Maybe in ddev-contrib or on Stack Overflow. Or a blog post. Or whatever. I think it would be a real step forward, and could be incorporated here eventually.

I will! It's still early, and I haven't written anything because I'm not sure that even having templates will be viable. But so far, the high level is:

1. Configure PHPStorm.
2. Copy `.idea` to `templates/idea/...`
3. Pare down the files as needed, only committing those without local paths, or editing the XML etc.
4. The following in a composer install script:

```
<?php

declare(strict_types=1);

namespace Organization\ProjectNameScripts;

use Composer\Script\Event;
use Composer\Util\Filesystem;

/**
 * Installs default PHPStorm configuration from the templates directory.
 */
class PhpStormTemplate {

  /**
   * Install PHPStorm's configuration if .idea doesn't already exist.
   */
  public static function installConfiguration(Event $event): void {
    if ($event->isDevMode() && !file_exists('./.idea')) {
      $event->getIO()->write('Creating default PHPStorm configuration.');
      $fs = new Filesystem();
      if (!$fs->copy('templates/idea', '.idea')) {
        $event->getIO()->write('<warning>Unable to create the .idea directory. Check directory permissions?');
      }

      $fs->rename('.idea/.gitignore.template', '.idea/.gitignore');
    }
  }

}
```

Note this also requires configuring `mutagen.yml` if using it, since `ddev composer install` will create the idea files that won't get sync'ed back out.

PR #3274, opened 2021-09-29
Support both amd64 and  arm64 image tarballs for quicksprint usage

rfay:
## The Problem/Issue/Bug:

Quicksprint needs to support mac M1, see https://github.com/drud/quicksprint/pull/203

* Separately package up amd64 and arm64 images

Note that using the script here I manually created the ddev_docker_images*.tar.xz that are needed and added them to the v1.18.0 release. That allows Quicksprint to move forward

- [ ] This will have to be tested using a release creation on the fork (rfay).

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3274"></a>

### DIFF
--- a/.ci-scripts/generate_artifacts.sh
+++ b/.ci-scripts/generate_artifacts.sh
@@ -7,7 +7,7 @@ set -o pipefail
 set -o nounset
 
 MKCERT_VERSION=v1.4.6
-BUILD_IMAGE_TARBALLS=false
+BUILD_IMAGE_TARBALLS=${BUILD_IMAGE_TARBALLS:-false}
 
 ARTIFACTS=${1:-/artifacts}
 BASE_DIR=$PWD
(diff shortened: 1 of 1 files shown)


rfay:
Experimental release on rfay/ddev: https://github.com/rfay/ddev/releases/tag/v1.18.1-alpha1

rfay:
Unfortunately the Windows self-hosted used for signing the windows binaries is set up just for drud/ddev, so couldn't get the build fully completed in https://github.com/rfay/ddev/actions/runs/1305850683, which is annoying. 

I did add SegmentKey secret to rfay/ddev, which was another issue. 

But the flow will either have to be changed or something to make this testable. Or just test it live on poor drud/ddev.

rfay:
It was too much brain damage to do an actual dummy release on rfay org, but I did manually test the script with `BUILD_IMAGE_TARBALLS=true ./.ci-scripts/generate_artifacts.sh /tmp/artifacts` and it built the correct set of tarballs.

PR #3275, opened 2021-09-30
Add example for arguments forwarding / passing in ddev custom commands

JPustkuchen:
## The Problem/Issue/Bug:
Passing arguments into custom commands is quite typical, but there's no documentation about that yet.
See https://github.com/drud/ddev/issues/3272

Adds documentation and an example

## Related Issue Link(s):
https://github.com/drud/ddev/issues/3272

### DIFF
--- a/docs/users/extend/custom-commands.md
+++ b/docs/users/extend/custom-commands.md
@@ -57,7 +57,21 @@ For example, to add a "solrtail" command that runs in a solr service, add `.ddev
 tail -f /opt/solr/server/logs/solr.log
 
 ```
+#### Arguments forwarding
+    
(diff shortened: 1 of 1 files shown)


rfay:
I'm a little reluctant because this has nothing to do with ddev, it's all about basic bash usage. 

Note also that there are many examples both in the global and per-project commands dirs.

rfay:
This tech stuff is so hard to sort out. There are so many layers to everything! And it's not clear at first what belongs to what layer. But then if a tool like ddev documents all the syntax of bash .. and docker... etc. then you can't figure out anything either because it gets so bloated.

rfay:
Please take a look at this @JPustkuchen and see if it addresses what you were after, https://github.com/drud/ddev/pull/3275/files - I think your concern was totally valid, but want to give more general instruction.

PR #3284, opened 2021-10-02
LDAP missing TLS_CACERT path from libldap-common

joelpittet:
## The Problem/Issue/Bug:
LDAP+SSL aka `ldaps://` doesn't work without the OS CA cert being set. In version 1.17.0, libldap-common was a dependency of libldap-2.4-2, and in 1.18.0 it moved it to "recommends" and doesn't install it.

```
DDEV 1.17 
Package: libldap-2.4-2
Version: 2.4.47+dfsg-3+deb10u6
Depends: libc6 (>= 2.28), libgnutls30 (>= 3.6.6), libsasl2-2 (>= 2.1.27+dfsg), libldap-common

DDEV 1.18
Package: libldap-2.4-2
Version: 2.4.57+dfsg-3
Depends: libc6 (>= 2.28), libgnutls30 (>= 3.7.0), libsasl2-2 (>= 2.1.27+dfsg)
Recommends: libldap-common
```

The file needed is:
`/etc/ldap/ldap.conf` which has `TLS_CACERT /etc/ssl/certs/ca-certificates.crt` pointing to the OS CA's path which is needed for client SSL connections.

Explicitly adds the package back.

`apt install ldap-utils` and use ldapsearch with ldaps:// like `ldapsearch -x -v -H ldaps://ldap.example.org`
before it will say `ldap_sasl_bind(SIMPLE): Can't contact LDAP server (-1)`
and after 
```
filter: (objectclass=*)
requesting: All userApplication attributes
# extended LDIF
#
# LDAPv3
# base <> (default) with scope subtree
# filter: (objectclass=*)
# requesting: ALL
#

# search result
search: 2
result: 32 No such object
```

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3284"></a>

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -83,6 +83,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -o Dpkg::Options::="--for
     iproute2 \
     iputils-ping \
     jq \
+    libldap-common \
     libpcre3 \
     locales-all \
     nano \
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -41,7 +41,7 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag for drud dev
-var WebTag = "v1.18.0" // Note that this can be overridden by make
+var WebTag = "20211004_joelpittet_ldap" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"


rfay:
Notes from slack conversation before it goes away:

> And there is notes that say we should add the TLS_CACERT manually ourselves: https://www.openldap.org/faq/data/cache/185.html

> So one recommends and the other depends
DDEV 1.17 
Package: libldap-2.4-2
Version: 2.4.47+dfsg-3+deb10u6
Depends: libc6 (>= 2.28), libgnutls30 (>= 3.6.6), libsasl2-2 (>= 2.1.27+dfsg), libldap-common

DDEV 1.18
Package: libldap-2.4-2
Version: 2.4.57+dfsg-3
Depends: libc6 (>= 2.28), libgnutls30 (>= 3.7.0), libsasl2-2 (>= 2.1.27+dfsg)
Recommends: libldap-common

rfay:
I'm fine with this PR, but I think this should probably be opened with the upstream, https://github.com/oerdnj/deb.sury.org/issues

rfay:
This won't be testable until I push an image for it and update the webimage version tag.

joelpittet:
@rfay I could see why they don't want to require this package because it's adding config file that may be user maintained. In our case it's not, and maybe any people don't tweak this but `/etc/ldap/ldap.conf` is really not required.

rfay:
Artifacts are at https://github.com/drud/ddev/actions/runs/1305287555 - If you could test the actual artifact @joelpittet - I don't think any of the rest of us have an https ldap server to hit, unless you can suggest one. You can just download the artifact for your OS and put it in your PATH before the regular one, or temporarily replace the regular one.

I note that if I run `docker run -it --rm drud/ddev-webserver:20211004_joelpittet_ldap bash`

I see `TLS_CACERT      /etc/ssl/certs/ca-certificates.crt` in `/etc/ldap/ldap.conf`.

Don't worry about the container build failure, it's unrelated. 

Thanks so much for all your work sorting this out!

joelpittet:
I might not be super quick on the draw, that was cool, I've never used a build artifact like that before. Worked like a charm v1.18.0-5-gb9c5b826 was the `ddev --version` that ran, and the ldap conf was there and the connection test I did with PHP worked great @rfay

PR #3298, opened 2021-10-10
Avoid redeclaration of constant `ABSPATH`

### LINKED ISSUE #3297: Autogenerated code / templates on ddev start
Hello, 
I d like to ask if there is a way to custom auto generated code / template after ddev start?

I know I can change it manually each time but imo its inefficient.

on example:
ddev generates this ``wp-config.php`` file on each wp project while ``ddev start``

and ABSPATH with combination of WP CLI commands generates this annoying notice in output in terminal:
"Notice: ABSPATH is already defined please bla bla bla"

To fix that, the template of ``wp-config.php`` needes to be changed for sth like that:

```
if( defined('ABSPATH')){
/** Absolute path to the WordPress directory. */
define('ABSPATH', dirname(__FILE__) . '/');
}
```

I would like to be able to change autogenerated template for ``wp-config.php`` file globally. 
Or have ability to define global hook which will be lunching command for each new created project. 
That command would be overriding this newly auto generated file.

That would solve the problem.

Awesome project!

gilbertsoft:
## The Problem/Issue/Bug:

If the constant `ABSPATH` is already declared there will be an PHP notice on `ddev start`.

This patch checks if the constant is already declared and only defines it if it's not already declared before.

## Related Issue Link(s):

Fixes #3297

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3298"></a>

### DIFF
--- a/pkg/ddevapp/wordpress.go
+++ b/pkg/ddevapp/wordpress.go
@@ -103,7 +103,9 @@ define('LOGGED_IN_SALT',   '{{ $config.LoggedInSalt }}');
 define('NONCE_SALT',       '{{ $config.NonceSalt }}');
 
 /** Absolute path to the WordPress directory. */
-define('ABSPATH', dirname(__FILE__) . '/{{ $config.AbsPath }}');
+if (!defined('ABSPATH')) {
+  define('ABSPATH', dirname(__FILE__) . '/{{ $config.AbsPath }}');
+}
 
 // Include for settings managed by ddev.
 $ddev_settings = dirname(__FILE__) . '/wp-config-ddev.php';


gilbertsoft:
> 
> 
> This is great, thanks. Should we be doing the same thing with everything else there?
> 
> https://github.com/drud/ddev/blob/0eab46ae39a303c6e998e12fd78e72ba15d589cb/pkg/ddevapp/wordpress.go#L129-L147

At the moment I would not touch these constants because it's used in the ddev webcontainer only. Everything outside this if block can be discussed ofc.

PR #3302, opened 2021-10-12
Add debug router-nginx-config command

cmuench:
## The Problem/Issue/Bug:
Sometimes are issues in the router nginx config.

Prints the generated Nginx config of the currently running router.

Start a project and run `ddev debug router-nginx-config`.

Test `TestDebugRouterNginxConfigCmd` is included in the PR.

## Related Issue Link(s):
#648

## Release/Deployment notes:
No backwards incompatible changes.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3302"></a>

### DIFF
--- a/cmd/ddev/cmd/debug-router-nginx-config_test.go
+++ b/cmd/ddev/cmd/debug-router-nginx-config_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/exec"
 	asrt "github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


cmuench:
> This is just doing docker-gen and looking for error right? It could do an `nginx -t` to see if there's a problem with the generated code as well, and `docker logs ddev-router` (although that's a little harder to comprehend.
> 
> I'd be interested in knowing the specific problem that made you work this hard :) And also whether `ddev list` showed the reason for the problem.
> 
> Thanks!

At netz98 we have ported all our projects to ddev. Several developers are modifying/extending the settings from time to time.
Issues like duplicate upstreams are very often. So I thought it's a good idea to make the router a bit more transparent. I know that we can also provide this information by using the docker cli.

I was also thinking about a `ddev doctor` command to automate the troubleshoot stuff.

PR #3305, opened 2021-10-14
[docs] explain xdebug profiling with ddev

AronNovak:
## The Problem/Issue/Bug:
If someone prefers to use XDebug for profiling, it should be documented how to do it.

Extends the documentation

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3305"></a>


rfay:
Updated this to get it into the mkdocs.yml, and also made it automatically put the files in .ddev/xdebug so you don't have to work so hard to get them.

Thanks so much for this contribution!

PR #3308, opened 2021-10-15
Add `--file=` option to `ddev import-db` (same as `ddev import-db --src=`)

rowild:
## The Problem/Issue/Bug:
While `export-db` uses `file` as the parameter to provide a name for the output file, `import-db` uses `src`, even though it is also a parameter that eventually references a file. 

For the sake of clarity and consistency a `file` flag is added to `import-db`, so users can simply think of a `file`, whether they use `import-db` or `export-db`.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3308"></a>

### DIFF
--- a/cmd/ddev/cmd/import-db.go
+++ b/cmd/ddev/cmd/import-db.go
@@ -25,9 +25,9 @@ can be provided if it is not located at the top level of the archive. An optiona
 can also be provided; the default is the default database named "db".
 Also note the related "ddev mysql" command`,
 	Example: `ddev import-db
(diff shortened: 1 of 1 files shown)


rfay:
I spent some more time on this
* Hid --src
* Checked tests

But then I notice that `ddev import-files` also has `--src`, and there it's probably because import-file can take a tarball or a directory.

So having some cold feet about this. You're getting import-db to match export-db, but then import-db and import-files don't match, and if I do the same thing to import-files it's not probably as good as --src

PR #3317, opened 2021-10-17
Support mysql on arm64 (mac M1) machines, fixes #3002

### LINKED ISSUE #3002: Support MySQL on arm64 (Apple Silicon M1)
Unfortunately, Oracle does not provide Debian packages for MySQL (any version). 

As noted in the [docs](https://ddev.readthedocs.io/en/stable/users/special-environments/#mac-m1-apple-silicon-and-docker-desktop-technical-preview) and because there are no options for mysql packages or images, DDEV on the Mac M1 (and any other arm64) does not currently have the capability to use MySQL as a db server.

There are a few ways we may be able to solve this:

1. Of course Oracle could support Debian arm64 and provide packages for it. Then there would be a docker image, or we could build one.  The request for this is https://github.com/docker-library/mysql/issues/318
2. We could build MySQL for arm64 ourselves and incorporate it into an image. This is likely to work but means yet another thing to maintain, and it's not easy. And of course there is risk, because we're not Oracle and don't know the ins and outs of what might need to be enabled in the build. This would be a whole effort to automate the build of MySQL and then build images for it. 
3. We could allow the use of the amd64 MySQL images with ddev on arm64. This can work on the Apple M1 (doesn't help at all on other arm64 architectures). However, because it's emulated, it's both risky (lots of crashes reported) and has performance problems.

The most likely option at this point is number 2, but it will likely require somebody to sponsor the effort or take it on themselves.

rfay:
## The Problem/Issue/Bug:

* #3002 

It's long been a problem that mac M1 users can't use mysql

Provide at least 5.7 and 8.0 for mac M1 users

1. Configure mysql for each of the supported versions and take it for a spin.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3317"></a>

### DIFF
--- a/docs/users/extend/config_yaml.md
+++ b/docs/users/extend/config_yaml.md
@@ -16,7 +16,7 @@ the .ddev/config.yaml is the primary configuration for the project.
 | webimage | docker image to use for webserver | It is unusual to change the default and is not recommended, but the webimage can be overridden with a correctly crafted image, probably derived from drud/ddev-webserver |
 | dbimage | docker image to use for db server | It is unusual to change the default and is not recommended, but the dbimage can be overridden with a correctly crafted image, probably derived from drud/ddev-dbserver |
 | dbaimage | docker image to use for dba server (phpMyAdmin server) | It is unusual to change the default and is not recommended, but the dbimage can be overridden with a correctly crafted image, probably derived from drud/phpmyadmin |
-| mariadb_version | Version of MariaDB to be used |  Defaults to 10.3, but 5.5, 10.0, 10.1, 10.2, 10.3, 10.4, 10.5 are available. Cannot be used with mysql_version. See [Database Server Types](database_types.md) for details and caveats. |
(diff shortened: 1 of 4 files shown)


rfay:
Apple arm64/M1 artifacts are at https://github.com/drud/ddev/suites/4111351442/artifacts/105008784
All artifacts are at https://github.com/drud/ddev/actions/runs/1364614285

To test, 
* download and untar the artifacts and put in your $PATH. You may want to `brew unlink ddev` to get the homebrew version out of the way.
* Export your existing database (because you have to switch database types), `ddev export-db --file=/path/to/db.sql.gz`
* `ddev stop`
* `ddev delete -Oy` your project, or of course you can `docker volume rm <project>-mariadb` instead
* Change the .ddev/config.yaml to use `mysql_version: 5.7` or `mysql_version: 8.0` depending on your upstream database type. `mariadb_version: ""`
* `ddev start` and `ddev import-db --src=/path/to/db.sql.gz`

PR #3324, opened 2021-10-20
Fix pantheon.yaml.example auth section so it checks for successful login

tsidel:
## The Problem/Issue/Bug:
Adding debugging/verbose code that Randy walked me through when experiencing trying to do a ddev pull pantheon -y with a bad machine token. It wasn't clear that the token was bad, and it was returning a "Pull failed: exit status 1 " status that wasn't at all verbose/clear what was going on. The issue was a bad character in the machine token.

The PR displays the error message of where the pull is failing if it's caused by a bad machine token

Insert a bad machine token into the global_config.yaml file, and you'll see the error messages. (Although, it's not exiting the pull as Randy had expected)

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3324"></a>


rfay (approved):
I reviewed the similar recipes and it seems like they're already OK. Thanks for taking the time to work through this and for the PR.

PR #3346, opened 2021-10-28
Use docker interface to determine host.docker.internal, fixes #2904

### LINKED ISSUE #2904: Linux: Get host.docker.internal with "host-gateway" with docker v20.10+

An oddity of docker has always been that the key "host.docker.internal" name has not been available on Linux, but has been in Docker Desktop. 

With Docker 20.10 they made it more available on linux, but didn't put it there by default for some reason.

https://github.com/docker/for-linux/issues/264#issuecomment-784985736 explains how you do it now. This should be a far more reliable way to get host.docker.internal, as using the docker0 interface as we currently do does not work everywhere; for example it doesn't work on gitpod.io (docker in docker) and most likely won't work in rootless docker. 

1. Detect docker 20.10+ before creation of the .ddev/.ddev-docker-compose-base.yaml
2. If 20.10+, use the newer technique to define host.docker.internal in the template there.

Related issues:

Rootless docker: https://github.com/drud/ddev/issues/2899
Gitpod.io and ddev, https://github.com/shaal/ddev-gitpod

rfay:
## The Problem/Issue/Bug:

* Our technique for determining host.docker.internal didn't work on Gitpod, and may have occasionally failed elsewhere
* There is a "correct" way to get it on linux (`docker network inspect bridge`), see https://docs.docker.com/engine/tutorials/networkingcontainers/
* We can use the docker api to get it as well, as done here.
* Thanks to @shaal for the inspiration for the better detection technique!
* This would have fixed #2904 - On docker 20.10+ we use host-gateway. But it doesn't work out, because we need host.docker.internal for NFS usage as well, and we don't get that then. So giving up on #2904

- [x] Test xdebug usage on normal Linux. (You can also `nc -l 0.0.0.0 9000` on the host and inside container `telnet host.docker.internal 9000` and type something - it should appear in the host window.)
- [x] Test on WSL2
- [x] Test on GitPod (without a docker-compose.host-docker-internal.yaml)

The existing xdebug tests should exercise this just fine.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3346"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: drud/ddev-gitpod-base:20211028_fix_gitpod_setup
+image: drud/ddev-gitpod-base:20211028_new_host_docker_internal_detection
 
 tasks:
   - name: ddev
--- a/.gitpod/images/gitpod-setup-ddev.sh
+++ b/.gitpod/images/gitpod-setup-ddev.sh
@@ -28,16 +28,3 @@ host_mailhog_port: "8025"
 # Assign phpMyAdmin port
 host_phpmyadmin_port: 8036
 CONFIGEND
-
-# We need host.docker.internal inside the container,
-# So add it via docker-compose.host-docker-internal.yaml
-hostip=$(awk "\$2 == \"$HOSTNAME\" { print \$1; }" /etc/hosts)
-
-cat <<COMPOSEEND >"${PROJDIR}"/docker-compose.host-docker-internal.yaml
-#ddev-gitpod-generated
-version: "3.6"
-services:
-  web:
-    extra_hosts:
-    - "host.docker.internal:${hostip}"
-COMPOSEEND
--- a/.gitpod/images/push.sh
+++ b/.gitpod/images/push.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eu -o pipefail
 
-DOCKER_REPO=drud/ddev-gitpod-base:20211028_fix_gitpod_setup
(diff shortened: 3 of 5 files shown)


shaal (approved):
It works great!
image

What if people use self-hosted Gitpod, would `GITPOD_WORKSPACE_ID` be available?

Would it make sense to first check if it's Linux and if `bridge` network interface is available, but if not - check if the Gitpod setting would work?

rfay:
Thanks for reviewing!

> What if people use self-hosted Gitpod, would GITPOD_WORKSPACE_ID be available?
I don't know, happy to look at other detection techniques, probably down the line. I thought of looking to see if `gp` was available.

> Would it make sense to first check if it's Linux and if bridge network interface is available, but if not - check if the Gitpod setting would work?

On both gitpod and WSL2 the bridge network interface shows as available... but doesn't exist. But I like your idea. Maybe checking for docker0 is still a good idea. 

Interestingly I finally know why TestDdevXdebugEnabled() doesn't work on WSL2

https://github.com/drud/ddev/blob/791cf945b9669d882729afcf1f03ea5a8bca737d/pkg/ddevapp/ddevapp_test.go#L656-L665

It's because WSL2 (with Docker Desktop) exposes host.docker.internal as the network interface on the WINDOWS side, but docker0 doesn't exist on the WSL2 side, so actually PhpStorm running inside WSL2 (as opposed to on Windows) can't actually do xdebug. 

The problem on both WSL2 and Gitpod is that without the docker0 endpoint, PhpStorm listening on all interfaces doesn't work... because there's no docker0 interface to listen on!

PR #3358, opened 2021-11-08
Adds further mutagen hints/refs to the caveats

garvinhicking:
## The Problem/Issue/Bug:

People may stop regarding mutagen as a possibility, if they read "doubles project size" with UGC or other data in mind. There however is a way to exclude specific things.

The docs now point to the section of the docs to address caveat of not needing to sync large UGC directories.

(none)

(none)

## Related Issue Link(s):

https://github.com/drud/ddev/issues/3357

(none)

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3358"></a>


docs/users/performance.md
```diff
@@ -148,7 +148,7 @@ You can run mutagen on all your projects, there's no limit. To configure it glob
...
 * **Best on macOS**: This is mostly for macOS users. WSL2 is already the preferred environment for Windows users, but if you're still using traditional Windows this makes a huge difference. Turning on mutagen doesn't make sense on Linux or WSL2.
-* **Increased disk usage**: Mutagen integration ends up at least doubling the size of your project code disk usage, because the code exists both on your computer and also inside a docker volume. So take care that you have enough overall disk space, and also (on macOS) that you have enough file space set up in Docker Desktop.
+* **Increased disk usage**: Mutagen integration ends up at least doubling the size of your project code disk usage, because the code exists both on your computer and also inside a docker volume. So take care that you have enough overall disk space, and also (on macOS) that you have enough file space set up in Docker Desktop. If you have a large amount of data like user-generated content that does not need syncing (i.e. "fileadmin" for TYPO3 or "upload_dir" for Drupal), you can exclude specific directories from getting duplicated. See [below for Advanced Mutagen configuration options](#mutagen-config).
```
rfay, line 151:
```suggestion
* **Increased disk usage**: Mutagen integration ends up at least doubling the size of your project code disk usage, because the code exists both on your computer and also inside a docker volume. So take care that you have enough overall disk space, and also (on macOS) that you have enough file space set up in Docker Desktop. If you have a large amount of data like user-generated content that does not need syncing (i.e. "fileadmin" for TYPO3 or "sites/default/files" for Drupal), you can exclude specific directories from getting synced and use regular docker mount for them instead. See [below for Advanced Mutagen configuration options](#mutagen-config).
```

rfay:
Thanks!

PR #3374, opened 2021-11-17
Allow project type PHP to import-files if upload_dir is set, fixes #2880

### LINKED ISSUE #2880: I want to use `ddev import-files` even though I use project type "php"
When I use the project type **php** I'm not able to use the command `ddev import-files` to copy files to a specific folder location.
This is annoying because `ddev pull <provider>` relies on the import-files command to copy pulled files to it's destination folder.

I would expect the `ddev import-files` command to work if the config option **upload_dir** is set in `.ddev/config.yaml`.

I've considered to ignore the error message when pulling files but there is no real reason why import-files needs more information than it's destination folder.

rfay:
## The Problem/Issue/Bug:

* #2880 

There's no reason a project type of type `php` shouldn't be able to `ddev import-files` if `upload_dir` is set.

Allows PHP projects with `upload_dir` set to import files

- [x] Set up php project
- [x] Set `upload_dir` and `ddev import-files` - should succeed and put imported files into the `upload_dir`. `upload_dir` should be relative to docroot
- [x] Unset `upload_dir` - `ddev import-files` should fail

- [x] Added test coverage for 'php' type, both with and without upload_dir

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3374"></a>


rfay:
This can be tested on gitpod with https://gitpod.io/#https://github.com/drud/ddev/pull/3374
Or download the build artifacts for your OS/Architecture from https://github.com/drud/ddev/actions/runs/1483723247 and put the binary in your PATH.

PR #3382, opened 2021-11-18
[docs] Use dynamic local home path (~) for homebrew install

codebymikey:
## The Problem/Issue/Bug:
The `/home/linuxbrew` path is hardcoded.

Replace it with an `~` so that it resolves to the appropriate home directory.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3382"></a>


rfay (changes requested):
Unfortunately, the way homebrew does the install, there is no user for homebrew, just a directory (/home/linuxbrew). The preferred, standard technique is to put it in /home/linuxbrew, not in ~/.linuxbrew. I think you can convince it to go into ~/.linuxbrew. But your suggestion doesn't work for people who follow the instructions on https://brew.sh and just follow the prompts.

codebymikey:
OK thanks for the clarification, I just [checked the code](https://github.com/Homebrew/install/blob/13b2d49/install.sh#L67-L68) and noticed where I went wrong with the install.

I typically use a passwordless sudo on my WSL2 box, so it was a surprise when I was prompted for my password.
And since I had accidentally put in [the wrong password](https://github.com/Homebrew/install/blob/13b2d49/install.sh#L487-L492), it opted to install it in `~/.linuxbrew` instead despite having sudo access.

I had passwordless sudo enabled using:
```
# file: /etc/sudoers.d/codebymikey
codebymikey ALL=(ALL) NOPASSWD: ALL
```

But `/usr/bin/sudo -n -v` doesn't seem to take that into account.

I've now reinstalled homebrew with the correct sudo password.

rfay:
Thanks! Just a note - sudo is required on Linux/WSL2 here because homebrew has to create /home/linuxbrew... after that there's no need for sudo.

PR #3389, opened 2021-11-21
Set correct permission for bash completion in AUR

stasadev:
It works fine with 644, no executable permissions required.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3389"></a>

### DIFF
--- a/.ci-scripts/bump_aur.sh
+++ b/.ci-scripts/bump_aur.sh
@@ -59,7 +59,7 @@ sha256sums=("${LINUX_HASH}")
 
 package() {
 	install -D -m 0755 ddev "\$pkgdir/usr/bin/ddev"
-	install -D -m 0755 ddev_bash_completion.sh "\$pkgdir/usr/share/bash-completion/completions/ddev"
+	install -D -m 0644 ddev_bash_completion.sh "\$pkgdir/usr/share/bash-completion/completions/ddev"
 }
 END
 


rfay:
Well, you won't be able to run the script without `bash <script>` if you remove the executable bit. 755 makes more sense in my opinion, it's the normal setting for any script. Why would you want to change it?

stasadev:
I asked to change the permissions for `ddev_bash_completion.sh` because I see everything here with 644:
```
ls -l /usr/share/bash-completion/completions
```

And here is my quick search for PKGBUILDs with bash completion (644 everywhere):

* [docker](https://github.com/archlinux/svntogit-community/blob/540286c35242a60c8653a0f661bbbafbbbccff90/trunk/PKGBUILD#L142-L143)
* [git](https://github.com/archlinux/svntogit-packages/blob/ed7474a15418cae054765600fe5f1e769b2d6aa6/trunk/PKGBUILD#L98-L100)
* [code](https://github.com/archlinux/svntogit-community/blob/e1f5cc22b4b6aa9eea3dd3fd6e8bf437d58f724e/trunk/PKGBUILD#L131-L133)
* [visual-studio-code-bin](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=visual-studio-code-bin#n61)

rfay (approved):
Wow do I appreciate your contributions and care looking at this stuff.

I assume this means that AUR (with which you obviously have far more experience than I do) executes the script with an explicit "bash". 

There's nothing wrong with 755 from a security perspective or anything else, but I'm fine with your suggestion to change it. 

But it's on you to check bash completions on AUR on the next release :)

PR #3391, opened 2021-11-22
Set writable perms on /var/lib/php in ddev-webserver start.sh

rfay:
## The Problem/Issue/Bug:

Problems cleaning up php gc() were reported in https://gitter.im/drud/ddev?at=619b5d2fcd3f06175dcbff24

`Uncaught PHP Exception ErrorException: "Notice: SessionHandler::gc(): ps_files_cleanup_dir: opendir(/var/lib/php/sessions) failed: Permission denied (13)` when PHP tries to clear / delete old sessions

## TODO

- [x] This doesn't solve the problem for hardened images (casual webhosting), because there we don't have sudo inside the container. But also, a 777 on /var/lib/php/sessions would be inappropriate and a security problem. So probably we have to run an elevated container during app.Start() to set this properly, as in https://github.com/drud/ddev/blob/d80a4d5c7f74d3aa43fbab366bcce7ff6b9d71aa/pkg/ddevapp/ddevapp.go#L851
- [x] This fix will allow rogue PHP code to explore php sessions, which doesn't matter much in a development environment, but definitely does in real webhosting.

Make sure that /var/lib/php is owned by the php-fpm user.

Not sure how to recreate this.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3391"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
@@ -79,7 +79,7 @@ ls /var/www/html >/dev/null || (echo "/var/www/html does not seem to be healthy/
 sudo mkdir -p ${TERMINUS_CACHE_DIR}
 
 sudo mkdir -p /mnt/ddev-global-cache/{bashhistory,mysqlhistory}/${HOSTNAME}
-sudo chown -R "$(id -u):$(id -g)" /mnt/ddev-global-cache/
+sudo chown -R "$(id -u):$(id -g)" /mnt/ddev-global-cache/ /var/lib/php
 
 # /mnt/ddev_config/.homeadditions may be either
 # a bind-mount, or a volume mount, but we don't care,
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -41,7 +41,7 @@ var DockerComposeFileFormatVersion = "3.6"
 var WebImg = "drud/ddev-webserver"
 
 // WebTag defines the default web image tag
-var WebTag = "20220124_postgres" // Note that this can be overridden by make
+var WebTag = "20211122_php_sessions" // Note that this can be overridden by make
 
 // DBImg defines the default db image used for applications.
 var DBImg = "drud/ddev-dbserver"


blankse:
I have the error very often. Is it possible to fix it in the next version or is there a workaround (move path for session via php.ini?)

rfay:
@blankse I would love to get it in, but haven't found anybody that could offer a recreation scenario. Can you please explain how to recreate this issue?

blankse:
@rfay I have this error in my Pimcore X projects with PHP 8. I analyzed it and should be reproducible with these steps:

# Create a Pimcore X demo instance
```
git clone https://github.com/pimcore/demo.git pimcore-demo
cd pimcore-demo
ddev config --project-type=php --docroot=public --webserver-type=apache-fpm --create-docroot --php-version=8.0
ddev start
ddev composer install
ddev exec php vendor/pimcore/pimcore/bin/pimcore-install --ignore-existing-config --admin-username admin --admin-password demo --mysql-host-socket db --mysql-username db --mysql-password db --mysql-database db --no-interaction
```

# Login in Backend

1. Open Backend https://pimcore-demo.ddev.site/admin/
2. Login with username: admin and password: demo
3. Stay at this window

# Call the Symfony Console
```
ddev exec php bin/console
```

# Watch the session files
```
ddev exec sudo ls -la /var/lib/php/sessions
```

- Should return two sess_* file (one from the backend and one from the console)
- The session file from the backend should be updated every 30sec (/admin/notification/find-last-unread ajax call)

# Wait for error

1. Wait > 24min (default session.gc_maxlifetime)
2. Afterwards max ~100 calls (default session.gc_divisor) from /admin/notification/find-last-unread ajax call or own calls with interacting with the backend should appear the error.

The backend complete freeze. No ajax call will be completed. A reload doesn't help to fix it. I must restart ddev "ddev restart" to clear all sessions. Than I can login to the backend again.

blankse:
@rfay You can set the session.gc_* configs in the php.ini to get the error faster.
.ddev/php/my-php.ini
``` 
[Session]
session.gc_probability = 1
session.gc_divisor = 100
session.gc_maxlifetime = 120
```

rfay:
I'd love it if you could test this @blankse - Artifacts for testing are at https://github.com/drud/ddev/actions/runs/1797622114

blankse:
@rfay I'm glad I could help you here :)
I will test it. 

I notice that the /var/lib/php/modules folder is set to 777. If you now assign /var/lib/php to the ddev user, that's obsolete, isn't it?
Here:
https://github.com/drud/ddev/blob/22803bc0519073e1493c777e6b6d661622aea55e/containers/ddev-webserver/Dockerfile#L127
And here:
https://github.com/drud/ddev/blob/22803bc0519073e1493c777e6b6d661622aea55e/containers/ddev-webserver/Dockerfile#L242

blankse:
@rfay I test it and it works 🎉 

The garbage collection deletes now the obsolete session files without error :)

rfay:
The issue is /var/lib/php/sessions, which gets created later (not during the docker image build I don't think). So that's why it has to be chowned on startup.

Thanks so much for your work on this!

PR #3392, opened 2021-11-22
AUR: Add zsh and fish completions

stasadev:
## The Problem/Issue/Bug:
I did not notice at first, but after reading the docs, I discovered that there are completions for all shells.
Tested in `zsh` and `fish`, everything works as expected.
It will be useful to have this functionality for all users of AUR out of the box.

Adds completions to standard `zsh` and `fish` folders.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3392"></a>


rfay (approved):
This is relatively hard to test in advance of a release but it seems pretty safe. Thanks!

I'm never quite sure who's currently using AUR/Arch and who has bounced off of it, so very much appreciate your maintenance here!

PR #3398, opened 2021-11-25
.git needs to be synced for Composer self.version

deviantintegral:
## The Problem/Issue/Bug:

I ran into issues trying to run `ddev composer require drush/drush` on a Drupal 9 checkout. Syncing the `.git` directory fixed it, allowing composer to properly resolve version contraints.

Updates the docs to call this situation out.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3398"></a>


rfay:
Thanks, this is a good addition. See also https://github.com/drud/ddev/issues/3384

I excluded .git by default in mutagen.yml because the maintainer of mutagen was concerned that the low-level (and so-many-files) filesystem usage of .git could lead to trouble.

rfay:
It's interesting, the two places where this has been a problem, git most likely won't be used, and certainly won't be used in the container, but it breaks composer. Is this a composer flaw in the end I wonder? I wonder how we can avoid people having to discover this over and over. (Docs are good, but mostly people don't read them.)

deviantintegral:
> I excluded .git by default in mutagen.yml because the maintainer of mutagen was concerned that the low-level (and so-many-files) filesystem usage of .git could lead to trouble.

That's really interesting. Given default Drupal 9 has ~25K files in in `core` and `vendor`, it doesn't seem like there would be an order of magnitude difference compared to the operations Composer or npm makes. Actually, npm as well would usually dwarf a .git directory.

> I wonder how we can avoid people having to discover this over and over. (Docs are good, but mostly people don't read them.)

We could test for `self.version` in `composer.json`?

rfay:
I have an idea: We should just bind-mount .git, it would solve so many things. Use the technique in https://ddev.readthedocs.io/en/latest/users/performance/#advanced-mutagen-configuration-options - but do it automatically.

rfay:
Opened 
* https://github.com/drud/ddev/issues/3400

rfay (approved):
I hope to add .git and upload_dir direct mounting in a release soon, but will go ahead and do this now and it can be edited with that PR.

rfay:
Thanks so much for this!

Issue #3369, opened 2021-11-15
Connecting to wrong database: docker-compose v2 is not respecting `links` any more  - ambiguous "db" hostname


rfay:
If multiple DDEV projects are running at the same time, they may try to connect to each other's databases.

docker-compose v2.1.1+ is not respecting the `links` statement any more, so if you use a servicename as a hostname... and more than one project is running, the hostname will be ambiguous, resulting in roundn-robin behavior.

1. It's recommended to all users to use `ddev-<project>-<servicename>` for configurations. So `ddev-<project>-db` for example. In that case you won't have this problem.
2. docker-compose v1 does not have this problem, so uncheck docker-compose v2 in Docker Desktop "Experimental settings" or `docker-compose disable-v2` (these work on macOS/Windows/WSL2 only, not Linux)
3. docker-compose v2 before 2.1.1 does not have this problem - see https://github.com/docker/compose/issues/8935

ddev-webserver ~/.my.cnf uses the ambiguous hostname "db" instead of `ddev-<project>-db` - that's fixed in https://github.com/drud/ddev/pull/3376

So:
- [x] Make sure ~/.my.cnf on ddev-webserver has fullname of container
- [x] This is another docker-compose v2 bug - file it
- [x] Review all ddev-contrib to make sure everything is using the unambiguous naming.

wtfred:
When having multiple project start at the same time, i randomly had errors like this on page refresh :
`TYPO3.CMS.Frontend.Controller.TypoScriptFrontendController": The page is not configured! [type=0][]`

- with docker-compose 2.1.1, when i hit F5, i have errors like 3 times every 5 refresh
- when downgraded to 2.0.1, i refresh like 10-15 times and never got the errors
- fixing my `.env` with `ddev-<projectname>-db` instead of just `db` and upgraded to 2.1.1 : refresh 10-15 times, no errors

(I don't use ddev default `AdditionalConfiguration.php`, i use `disable_settings_management: true` on all my project and have my own `AdditionalConfiguration.php` where i read db values from ENV vars)

jonaseberle:
I am using docker-compose v1 in production without any `links` with identical `docker-compose.yaml` in multiple directories (aka different projects). 

I have a pretty similar setup to `ddev` with a frontend proxy and an additional "external" network where the frontend proxy and some containers are in. (I am mentioning that because that might be were project-local name resolution gets now overridden in v2?)

[...]

PR #3403, opened 2021-11-28
Use simple container name as hostname inside docker network (use "db" instead of "ddev-<project>-db")

jonaseberle:
## The Problem/Issue/Bug:

* In recent versions of docker-compose there have been problems with the `links` stanza, where it does not properly bind "db" for example, to the project's "db" service. These have been solved in the docker-compose that is currently bundled by ddev.
* But for years (since the beginning of ddev) containers included without a "links" (like solr) have resulted in ambiguity and unpredictable behavior between projects. One ddev project might accidentally connect to another project's solr service, for example.

Unambiguity in name resolution was recently only possible in `ddev`when using `ddev-<projectname>-<service>` which requires a level of environment-awareness in apps that should usually not be necessary in a container environment.

This PR sets up networking where the service name of a project's container is perfectly adequate to address it. For example, `ping db` in the web container will always get the related project's db container. `ping solr` will always get the related solr container, not one from another service.

This makes ddev use  standard docker-compose networking as described in https://docs.docker.com/compose/networking/.

The "ddev" network is now only used to communicate across projects and with the proxy.

Contrib-service-recipes that only need per-project networking (via its service name or `ddev-<projectname>-<service>:<port>`) will not need any changes.
Those that use HTTP_EXPOSE and HTTPS_EXPOSE to be reachable externally via `http[s]://<project>:<port>` (via ddev-router) or cross-project need to add the "ddev" network:
```
networks: [default, ddev]
```
This can be given as general advise, too because it does not hurt.

All ddev-contrib service examples should be updated.

To test this: 
* Use [gitpod](https://gitpod.io/#https://github.com/drud/ddev/pull/3403)
* Download the artifacts in the latest [build](https://github.com/drud/ddev/actions/workflows/pr-build.yml?query=branch%3A20211128__docker-compose-networking++) and [put ddev in your path](https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md#testing-a-pr)

Create multiple ddev projects (`ddev config` with all defaults).
Run `ddev exec ping db` from them.

You should see answers only from the project-local "db" container.
With docker-compose v2.1.1 you would see answers from other "db" containers.

In order to test cross-project connectivity, run 
* `ddev exec ping ddev-<other-projectname>-db`.

To test this properly, use docker-compose v2.1.1, which had the bug about not respecting `links`. `ddev config global --docker-compose-version=v2.1.1`

- [x] Added TestNetworkAmbiguity

## Related Issue Link(s):

* #3369
* docker/compose#8935

## Remarks/tbd:
* This change tries to minimize code changes. For improved self-documentation the network should rather be called "ddev_proxy" or "ddev_frontend".
- [x] Removed ddev-ssh-agent networks completely
- [x] Usage of docker-compose service "links" should not be necessary any more.
- [ ] ddev-contrib services (especially containing HTTP_EXPOSE) needs to be updated with the new networking on release
- [ ] Release notes need to explain the change and what it means to 3rd party projects

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3403"></a>

### DIFF
--- a/cmd/ddev/cmd/auth-ssh_test.go
+++ b/cmd/ddev/cmd/auth-ssh_test.go
@@ -35,7 +35,7 @@ func TestCmdAuthSSH(t *testing.T) {
 	assert.NoError(err)
 
 	// Run a simple ssh server to act on and get its internal IP address
-	_, err = exec.RunCommand("docker", []string{"run", "-d", "--name=test-cmd-ssh-server", "--network=ddev_default", "drud/test-ssh-server:v1.16.0"})
+	_, err = exec.RunCommand("docker", []string{"run", "-d", "--name=test-cmd-ssh-server", "--network=ddev", "drud/test-ssh-server:v1.16.0"})
 	assert.NoError(err)
 	//nolint: errcheck
 	defer dockerutil.RemoveContainer("test-cmd-ssh-server", 10)
--- a/cmd/ddev/cmd/poweroff.go
+++ b/cmd/ddev/cmd/poweroff.go
@@ -5,6 +5,7 @@ import (
 	"github.com/drud/ddev/pkg/dockerutil"
 	"github.com/drud/ddev/pkg/util"
 	"github.com/drud/ddev/pkg/version"
+	docker "github.com/fsouza/go-dockerclient"
 	"github.com/spf13/cobra"
 )
 
@@ -47,5 +48,15 @@ func powerOff() {
 
 	if err := ddevapp.RemoveSSHAgentContainer(); err != nil {
 		util.Error("Failed to remove ddev-ssh-agent: %v", err)
+	}
+	// Remove current global network ("ddev") plus the former "ddev_default"
+	removals := []string{"ddev", "ddev_default"}
+	for _, networkName := range removals {
+		err = dockerutil.RemoveNetwork(networkName)
+		_, isNoSuchNetwork := err.(*docker.NoSuchNetwork)
+		// If it's a "no such network" there's no reason to report error
+		if err != nil && !isNoSuchNetwork {
+			util.Warning("Unable to remove network %s: %v", "ddev", err)
+		}
 	}
 }
--- a/containers/ddev-router/docker-entrypoint.sh
+++ b/containers/ddev-router/docker-entrypoint.sh
@@ -40,9 +40,9 @@ mkcert -install
 
 # It's unknown what docker event causes an attempt to use these certs/.crt files, but they might as well exist
 # to prevent it.
-mkcert -cert-file /etc/nginx/certs/.crt -key-file /etc/nginx/certs/.key "*.ddev.local" "*.ddev.site" 127.0.0.1 localhost ddev-router ddev-router.ddev_default
+mkcert -cert-file /etc/nginx/certs/.crt -key-file /etc/nginx/certs/.key "*.ddev.local" "*.ddev.site" 127.0.0.1 localhost ddev-router ddev-router.ddev
 
 if [ ! -f /etc/nginx/certs/master.crt ]; then
-  mkcert -cert-file /etc/nginx/certs/master.crt -key-file /etc/nginx/certs/master.key "*.ddev.local" "*.ddev.site" "ddev-router" "ddev-router.ddev_default" 127.0.0.1 localhost
+  mkcert -cert-file /etc/nginx/certs/master.crt -key-file /etc/nginx/certs/master.key "*.ddev.local" "*.ddev.site" "ddev-router" "ddev-router.ddev" 127.0.0.1 localhost
 fi
 exec "$@"
--- a/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
@@ -94,10 +94,6 @@ sudo mkdir -p ${CAROOT} && sudo chmod -R ugo+rw ${CAROOT}
 # This will install the certs from $CAROOT (/mnt/ddev-global-cache/mkcert)
 mkcert -install
 
-if [ -f ~/.my.cnf ]; then
-  perl -pi -e "s/host=db/host=ddev-${DDEV_PROJECT}-db/" ~/.my.cnf
-fi
-
 # VIRTUAL_HOST is a comma-delimited set of fqdns, convert it to space-separated and mkcert
 CAROOT=$CAROOT mkcert -cert-file /etc/ssl/certs/master.crt -key-file /etc/ssl/certs/master.key ${VIRTUAL_HOST//,/ } localhost 127.0.0.1 ${DOCKER_IP} web ddev-${DDEV_PROJECT:-}-web ddev-${DDEV_PROJECT:-}-web.ddev_default
 echo 'Server started'
--- a/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
@@ -95,7 +95,7 @@ sudo mkdir -p ${CAROOT} && sudo chmod -R ugo+rw ${CAROOT}
 mkcert -install
 
 # VIRTUAL_HOST is a comma-delimited set of fqdns, convert it to space-separated and mkcert
-CAROOT=$CAROOT mkcert -cert-file /etc/ssl/certs/master.crt -key-file /etc/ssl/certs/master.key ${VIRTUAL_HOST//,/ } localhost 127.0.0.1 ${DOCKER_IP} web ddev-${DDEV_PROJECT:-}-web ddev-${DDEV_PROJECT:-}-web.ddev_default
+CAROOT=$CAROOT mkcert -cert-file /etc/ssl/certs/master.crt -key-file /etc/ssl/certs/master.key ${VIRTUAL_HOST//,/ } localhost 127.0.0.1 ${DOCKER_IP} web ddev-${DDEV_PROJECT:-}-web ddev-${DDEV_PROJECT:-}-web.ddev
 echo 'Server started'
 
 # We don't want the various daemons to know about PHP_IDE_CONFIG
--- a/containers/ddev-webserver/ddev-webserver-prod-scripts/start.sh
+++ b/containers/ddev-webserver/ddev-webserver-prod-scripts/start.sh
@@ -89,10 +89,6 @@ mkdir -p ${CAROOT}
 # This will install the certs from $CAROOT (/mnt/ddev-global-cache/mkcert)
 mkcert -install
 
-if [ -f ~/.my.cnf ]; then
-  perl -pi -e "s/host=db/host=ddev-${DDEV_PROJECT}-db/" ~/.my.cnf
-fi
-
 # VIRTUAL_HOST is a comma-delimited set of fqdns, convert it to space-separated and mkcert
 CAROOT=$CAROOT mkcert -cert-file /etc/ssl/certs/master.crt -key-file /etc/ssl/certs/master.key ${VIRTUAL_HOST//,/ } localhost 127.0.0.1 ${DOCKER_IP} web ddev-${DDEV_PROJECT:-}-web ddev-${DDEV_PROJECT:-}-web.ddev_default
 echo 'Server started'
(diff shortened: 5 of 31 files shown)


rfay:
Thanks for pursuing this! I know it has great potential (haven't tried it out yet)

All services that use HTTP_EXPOSE (most of them that have a web interface, including solr, etc) need to communicate with the ddev-router. So all of those would have the traditional problem of ambiguity of service name? 

An FYI - https://github.com/docker/compose/issues/8935  claimed to be fixed the other day.

jonaseberle:
> All services that use HTTP_EXPOSE (most of them that have a web interface, including solr, etc) need to communicate with the ddev-router. So all of those would have the traditional problem of ambiguity of service name?

They would have to add
```
networks:
 - default
 - ddev_default
```
`solr` can then be used unambiguously from the same project. From another container it is still accessible as `ddev-<projectname>-solr`. 

I added that use case to the manual testing instruction.

jonaseberle:
> An FYI - [docker/compose#8935](https://github.com/docker/compose/issues/8935) claimed to be fixed the other day.

That might mean we do not have to make any changes :)
I think "links" is rather a workaround, though. "links" needs all local services added to any container to work totally hassle-free and as expected for users adding services. The "default" network solves that better in my opinion.

rfay:
Why would we not provide the two networks on every service?

jonaseberle:
I'd say because it is not needed.

Pro "ddev_default" on all services:
* Easy to communicate ("Breaking change: You need to add `networks: [default, ddev_default]` to all custom services")
* Even the default `db` container would still be reachable cross-project (for whoever uses that now).

Pro "ddev_default" only on services that need it:
* Isolation: Only open connectivity for services that really need them
* The change is breaking anyways, so we could use it to educate about network segmentation. I think it could be explained well enough with two sentences. "If your service needs to be reached from your browser or from other projects, ... . By default all services of a project can communicate with each other".

So yes, when thinking about the ecosystem it might be easier to just use `[default, ddev_default]` throughout. The latter might be only interesting to network nerds and security fanatics.

rfay:
It's not uncommon to use `db` service between projects - most commonly for migrations; the new (migrating) project accesses the database of the old project.

jonaseberle:
I updated the PR description to reflect that.

jonaseberle:
Tests are still erroring at `TestGetLocalHTTPResponse`. Do you know how I could test that locally? Which `docker-compose` version is used in the tests?

rfay:
Details about running tests are in https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md - you'll want `export GOTEST_SHORT=true` and `make testpkg TESTARGS="-run TestGetLocalHTTPResponse` 

My preference is to run in GoLand or vscode. It's easy to do this in Gitpod's vscode as well.

jonaseberle:
I am trying to reproduce this run https://github.com/drud/ddev/runs/4380370899?check_suite_focus=true

This is ok for me locally and in Gitpod:
```
make testpkg TESTARGS="-run TestCmdStopMissingProjectDirectory"
```

I am puzzled by the `mutagen: false` errors (If at all I would expect problems with mutagen, not without).

Also the failures on `arm` puzzle me https://github.com/drud/ddev/pull/3403/checks?check_run_id=4380370760

If you don't have an idea I'll set up an Ubuntu 20.04 VM and do exactly what' done here https://github.com/drud/ddev/blob/master/.github/workflows/tests.yml.

rfay:
Wow, I've never seen anything like that before. 

Fails exactly the same way in [CircleCI](https://app.circleci.com/pipelines/github/drud/ddev/5530/workflows/71aa03a4-c211-4ce0-8314-6d82d83cc32e/jobs/35652) on arm64 :(

The key thing about those TestCmd tests is they don't necessarily get your ddev right if you're running them one at a time, so you'll want to make sure that the ddev in your PATH is the one you built. I often put a link to it in ~/bin . You can also use `export DDEV_BINARY_FULLPATH=/path/to/built/ddev` - if using the Makefile (`make testcmd`) it does this for you. But the TestCmd tests actually run the ddev binary.

Interesting factoid is that it worked on WSL2, which uses docker-compose v2.1.1 (WSL2 is the only place on test system that is using docker-compose v2 right now)
Also worked on ddev-windows-mutagen, which uses docker-compose v1

jonaseberle:
Thank you for these hints.

The tests were flaky for me so I prepared a Ubuntu 20.4 VM and always ran them from a prepared snapshot ([Gist](https://gist.github.com/jonaseberle/619d3d3c8c8cd3a1f9f49668c5c4f610)). Not sure if I went a bit over the top there.

I've worked on removing auto-created project networks today. 
In the end it was easiest to use `docker-compose down` rather than trying to produce the correct network name (ddev project "a.b.c.DE" => network "ddev-abcde_default"...).

PS:
Both GoLand and VSCode (with the Go extension) are annoyingly removing and rewriting Go imports all the time. Do you know how I could stop them from doing that?

rfay:
Thanks for all your work on this @jonaseberle - 

On goimports, I have everything turned on on Goland and it seems to do what I want. I wonder if maybe you and I just have different versions of goimports. 
```
$ ls -l $(which goimports)
-rwxr-xr-x  1 rfay  admin  5057410 Jan 28  2021 /usr/local/bin/goimports
```

Here's my setting for GoLand
Cursor_and_Preferences_and_ddev_–_utils_go

I imagine you can just turn that stuff off if it's not useful.

jonaseberle:
I had these boxes unticked but it would still try to remove on save. Thank you for `goimports`!
I did `go install golang.org/x/tools/cmd/goimports@latest` and now GoLand is happy ;)

rfay:
Mostly if you can get general manual usage working, then get the github tests working, you can ignore any other false failures on the buildkite builds. Those are more fragile because... "Docker Desktop".

jonaseberle:
I am using the patched version for day-to-day use. 
The currently failing tests are reproducible between test runs but I haven't looked into it yet. Many are about problems with writing config. I'll check it latest on the weekend.

rfay:
This will need to be rebased to pass tests since I updated the macOS/Windows test runners to Docker Desktop 4.3.0

jonaseberle:
When running the failing `TestSshAuthConfigOverride` alone, it passes. When running the whole suite, it fails.

I am using it in day-to-day work (with the required addition to additional services) and have not met any problems.

I am going to rebase now but I am not sure if you think it is worth to push it further.

rfay:
Usually if a test fails only when the tests around it are failing, it means there are side-effects from another test. It's annoying chasing those, but it can usually be narrowed down by running the tests that are just before it, etc. 

I'm all in favor of this, so there's no need to give up on it. It's just not my focus right now. I also haven't looked closely, but my hope is that this is a better solution than `links`.

rfay:
Rebased.   I apologize but the main yaml files got moved into their own separate places, so there was no way to successfully rebase this to current. As a result, I manually picked apart the changes to apply them for testing.
    
Your original commit tree, still available, was 8e3b43a84dd84420294c57ba2565f7e044ed4461

I've been studying your approach and the networking page and very much appreciate your work on this. I think the true test will be to remove the `links` statements and see if everything still works.

rfay:
Looks like we've about got it. Now for a good automated test, manual testing, and reverting all the references to `ddev-><project>-db` !

rfay:
Created https://github.com/drud/ddev/pull/3479 as a baseline, it should be adequate to test this.

jonaseberle:
Thank you for picking that up. That indeed looks good. Are you planning to merge the new test into this one?

rfay:
I was going to pull the new test after it passes, then rebase. Same thing I guess.

rfay:
I've rebased this and think it's ready. Please take a look and see what you think @jonaseberle - It's an awesome step forward for ddev.

jonaseberle:
I found 2 problems. 

A minor one: Running `ddev stop`/`ddev delete -Oy` on an already stopped project yields a warning:
```
$ ddev stop
Project www.ratbacher.de is already stopped. 
ddev-wwwratbacherde  Warning: No resource found to remove 
Project www.ratbacher.de has been stopped. 
```

A major one: The project network does not get removed for some projects. I think that is a problem because that can leave unnecessarily many virtual network interfaces. Those networks even persist across docker restarts. At least `ddev poweroff` should leave no project networks behind IMHO.
```
$ ddev stop
Container ddev-www.schempp-hirth.com-web  Stopping 
... # no errors or warnings, but also nothing about the network removal
    # that seems to imply that the guard clause fileutil.FileExists(app.DockerComposeFullRenderedYAMLPath()) yields "false"
    # - but I am not sure
Project www.schempp-hirth.com has been stopped. 

$ docker network ls
NETWORK ID     NAME                               DRIVER    SCOPE
c4767a2d6b82   bridge                             bridge    local
5e25314ecc00   ddev-wwwschempp-hirthcom_default   bridge    local
eea130851847   ddev_default                       bridge    local
bd0294da3e6a   host                               host      local
246a612e7b0c   none                               null      local

$ docker network rm ddev-wwwschempp-hirthcom_default 
Error response from daemon: error while removing network: network ddev-wwwschempp-hirthcom_default id db8cff7f4aa5854e7dd6cf603734c42a29269fe6c4973e55d96152c40c14efe2 has active endpoints

$ docker network inspect ddev-wwwschempp-hirthcom_default 
  ... # the ddev agent is somehow in that project network:
        "Containers": {
            "a4889472d1f6728d4e3200cdcd7c8c84d27efe48c08cb96d38f3f2285dfc3214": {
                "Name": "ddev-ssh-agent",
                "EndpointID": "baa3a01000c1e07596185a8230429a2fc7f1b8acf7bc8bd1a69f91eb7aadae4e",
                "MacAddress": "02:42:ac:18:00:02",
                "IPv4Address": "172.24.0.2/16",
                "IPv6Address": ""
            }
        },
  ...

$ docker stop ddev-ssh-agent && docker network rm ddev-wwwschempp-hirthcom_default 
# network gets removed

```

With the previous https://github.com/drud/ddev/commit/8e3b43a84dd84420294c57ba2565f7e044ed4461 that is not reproducible:
```
$ ddev stop
Container ddev-www.schempp-hirth.com-web  Stopping 
...
Network ddev-wwwschempp-hirthcom_default  Removing  # < here
Network ddev-wwwschempp-hirthcom_default  Removed   # < docker-compose down removes the project networks
Project www.schempp-hirth.com has been stopped. 
```

It does not happen with all of my projects. 
Here's a reproducible example:
```
$ mkdir a.b.c && cd a.b.c
$ ddev config --project-type php && ddev start && ddev stop
$ docker network ls
# project network was not removed 
```
Could we have tests for unusual project names? Multiple `.` and `-`seem to be problematic. Uppercase is also valid as a project name currently.

rfay:
Thanks for your careful eye!

1. I took the opportunity to limit the chattiness (or at least the amount we output) of docker-compose. I never have thought we needed 3 separate messages about every container creation. That gets rid of the "no resource" when stopping and nothing exists.
2. The failure of network to delete was very odd. It turns out that when I removed all network declaration from the ddev-ssh-agent compose file, it then somehow attached itself to the network of the ddev project that was starting it. As a result of that, with `ddev ssh-agent` still running, the network wasn't removed on `docker-compose down` because it was still in use. I fixed this just by adding the network statements back into the ssh-auth-compose file. No idea why it did that. I also added a little test coverage to make sure the network is gone.

Again, thanks so much for the brilliance of this approach and for the careful testing and good eye.

rfay:
And then I went and...
* changed the name of the global network to "ddev" instead of "ddev_default". Never understood why it was called "ddev_default".
* Make `ddev poweroff` also delete the "ddev" network.

jonaseberle:
I tried to break it really hard. This works beautifully now. Thank you so much!

I especially played around with shared and project-local services again - and am really pleased how stable things are.
* Local services: Usual project-local services do not need any changes. Their local (or full) name **always** resolves from the same project and **never** resolves from other projects.
* Shared services: In order to make a container available to other projects (or to the router), one has to add at least `networks: [ddev]`. This resolves the short service name as well as the full container name (`ddev-...`) from everywhere. It is advisable to always add `[default, ddev]`, though in order for the next point to work as expected.
* Unambiguous: A local service of the same name always takes precedence in resolving. Short service name and full container name can be used.

The messages on `ddev start`/`stop`/`poweroff` are so noticeably cleaner now, too! This alone makes my heart happy.
The network name `ddev` is also very sensible I think. (I am adapting the first post)
`ddev poweroff` does not leave any trace behind.

Get well into the new year, Randy.

jonaseberle:
Sorry, I really didn't want to confuse you - or did you pick the wrong emoji?

rfay:
I'm confused. What emoji? :)

jonaseberle:
I meant that one under my last post:
image
:)

rfay:
Oh, my eyes just aren't what they ought to be. Must not have had my glasses on. Was trying for a smiley face. Sorry!

PR #3407, opened 2021-11-30
Truncate existing file when `ddev export-db --file=`, fixes #3402

### LINKED ISSUE #3402: export-db sometimes writes files with trailing garbage
**Describe the bug**

Sometimes, `ddev export-db` appears to write trailing garbage to gzipped files.

**To Reproduce**

This is very inconsistent, I can't reproduce on demand and when it does happen then immediately rewriting the file succeeds:

```
$ ddev export-db -f db.sql.gz
Wrote database dump from site database 'db' to file db.sql.gz in gzip format.
$ gunzip db.sql.gz

gzip: db.sql.gz: decompression OK, trailing garbage ignored

$ rm db.sql 
$ ddev export-db -f db.sql.gz
Wrote database dump from site database 'db' to file db.sql.gz in gzip format.
$ gunzip db.sql.gz
$
```

**Expected behavior**

The file should always gunzip cleanly with no warnings.

Ubuntu 20.04
```
DDEV-Local version	v1.17.7                                
architecture      	amd64                                  
db                	drud/ddev-dbserver-mariadb-10.3:v1.17.7
ddev-ssh-agent    	drud/ddev-ssh-agent:v1.17.0            
docker            	20.10.10                               
docker-compose    	1.23.2                                 
os                	linux                                  
router            	drud/ddev-router:v1.17.6               
web               	drud/ddev-webserver:v1.17.7         
```

longwave:
## The Problem/Issue/Bug:

When using `ddev export-db` if the output file already exists and is larger than the new output file, the file will be the same size as before. This causes `ddev import-db` to fail again, `gunzip` to consider that the file has garbage at the end, etc.

Closes #3402

The file is truncated at opening.

Copy a very large file to db.sql.gz.
Run `ddev export-db -f db.sql.gz` on a site with a small database.
The file should be smaller than before, and `gunzip` should not complain about it.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3407"></a>

### DIFF
--- a/cmd/ddev/cmd/export-db_test.go
+++ b/cmd/ddev/cmd/export-db_test.go
@@ -60,6 +60,8 @@ func TestCmdExportDB(t *testing.T) {
 	err = os.Chdir(tmpDir)
 	assert.NoError(err)
 
+	err = app.MutagenSyncFlush()
+	assert.NoError(err)
 	err = os.RemoveAll(outputFileName)
 	assert.NoError(err)
 	command = exec.Command(DdevBin, "export-db", site.Name, "-f="+outputFileName, "--gzip=false")
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -576,7 +576,7 @@ func (app *DdevApp) ExportDB(outFile string, gzip bool, targetDB string) error {
 		opts.Cmd = fmt.Sprintf("mysqldump %s | gzip", targetDB)
 	}
 	if outFile != "" {
-		f, err := os.OpenFile(outFile, os.O_RDWR|os.O_CREATE, 0644)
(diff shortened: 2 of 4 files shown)


rfay (approved):
I manually tested this on macOS and linux (gitpod) and can recreate the problem with released ddev and the problem is gone with this version. Nice work!

It may need a test. If so I can add something to the test that will verify. Probably can just create a big file before the existing tests.

rfay:
I added to TestDdevExportDB a section that first exports the uncompressed, then renames that to the compressed (it being larger). Then exports the gzipped. Verifies success and size.

Also added a tiny mutagen-race fix (Windows only) in TestCmdExportDB that I happened to see in the tests of this one. It's been around a while but is transient.

PR #3409, opened 2021-12-01
Download and use a private docker-compose, fixes #3405

### LINKED ISSUE #3405: Bundle docker-compose with ddev so we don't have unexpected breakage
docker-compose v2 has had a lot of growing pains. Nearly every release has a new problem that breaks ddev. 

We should bundle docker-compose with ddev and download it on demand just like we do with mutagen. That way a known version would always be in use. 

* Download explicit version named, just as we do with mutagen
* Add a config.yaml option to use docker-compose in PATH instead of downloaded one (for testing, overriding, etc.) - This also allows docker-compose v1.

rfay:
## The Problem/Issue/Bug:

docker-compose versions just break too often

Use a specified, known good, private docker-compose version

## TODO
- [x] Add global configuration to override the docker-compose version
- [x] Add global configuration to allow using docker-compose from system $PATH
- [x] Consider adding support for docker-compose v1 (different filenames/download pattern, and no arm64 support)

* Download the specified docker-compose version when needed during `ddev start`
* Use that version in all docker-compose commands

- [x] Normal casual testing
- [x] `ddev version` should show the specified version, regardless of system version
- [x] Run it against the current broken Docker Desktop 4.3.0 with docker-compose 2.2.1 in it
- [x]  Remove the system's docker-compose and verify it still works
- [x] Remove the ~/.ddev/bin/docker-compose and view the single download of docker-compose on ddev start
- [x] Verify that docker-compose versions 1.x can be downloaded
- [x] Verify that global config using external (PATH) docker-compose can work
- [x] Test basic usage on
  - [x] macOS amd64
  - [x]  macos arm64
  - [x] Windows
  - [x] Linux amd64
  - [x] WSL2
  - [x] Gitpod

* Adds TestDockerComposeDownload to test the various types of usage based on global config

## Related Issue Link(s):

One hundred docker-compose regressions

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3409"></a>


.buildkite/test.sh
```diff
@@ -60,7 +60,7 @@ docker run --rm alpine sh -c '
   done
 '
 
-rm -rf ~/.ddev/Test* ~/.ddev/global_config.yaml ~/.ddev/homeadditions ~/.ddev/commands
+rm -rf ~/.ddev/Test* ~/.ddev/global_config.yaml ~/.ddev/homeadditions ~/.ddev/commands ~/.ddev/bin/docker-comnpose*
```
jonaseberle, line 63: small typo here *comnpose*

rfay: I guess that's not going to work out! Thanks for paying attention! :)

rfay:
```suggestion
rm -rf ~/.ddev/Test* ~/.ddev/global_config.yaml ~/.ddev/homeadditions ~/.ddev/commands ~/.ddev/bin/docker-compose*
```

stasadev: This typo somehow made its way to master branch :)

Issue #1678, opened 2019-06-29
Package ddev using native linux packages using openbuildservice or goreleaser


rfay:
**Is your feature request related to a problem? Please describe.**

Linux people are used to native package managers like apt and yum. 
linuxbrew is good, but it's not their normal situation.

However it's way beyond our capacity to maintain multiple packages

**Describe the solution you'd like**

Consider using OpenSuse's https://openbuildservice.org/ (See also https://en.wikipedia.org/wiki/Open_Build_Service) , which builds all variants as needed.

**Describe alternatives you've considered**

**Additional context**

rfay:
I spent quite a few hours exploring [open build service](https://openbuildservice.org), and am afraid it probably isn't worth the effort or maintenance risk involved. It seems that although a fantastic project, it's quite unevenly maintained and documented. 

* I was able to build RPMs and debs using the technique described in the off-site tutorial [Build packages with OpenBuild service](https://www.sidorenko.io/post/2016/05/build-packages-with-openbuild-service/). His approach was really very much what we'd want to do - build package from release artifacts rather than the source. It would be a shame to build from source when we already have a consistent method of doing that. However, once I built the debian repo, it took ages to figure out how to use it, and then when I got that right, the `apt-get install` hit something on debian's hosting that [redirected to an http url and broke everything](https://gist.github.com/rfay/ba875e761ae41d86c433fbe4f4259d1c). They seem to have only [IRC support](https://openbuildservice.org/support/) and that tends to be pretty spotty. I didn't get a response. Note that the correct sources.list line was `deb https://download.opensuse.org/repositories/home:/rfay/Debian_9.0/ ./`
* The same author's [Short way to a deb package](https://www.sidorenko.io/post/2016/05/a-short-way-to-a-deb-package/) was a good reintroduction to debian packages, and showed the technique of using binaries instead of source.
* I failed several times trying to make the [OBS Beginner's Guide](https://openbuildservice.org/help/manuals/obs-beginners-guide/) official documentation work. 

* [open-build-service github project](https://github.com/openSUSE/open-build-service)
* [Debian builds](https://en.opensuse.org/openSUSE:Build_Service_Debian_builds) official docs

rfay:
Opened https://github.com/openSUSE/open-build-service/issues/7830 to chase this

rfay:
The problem with redirecting to https seems to have been solved on download.opensuse.org, as I can successfully install my test project today on debian9. Maybe this is worth some more effort after all.

[...]

PR #3414, opened 2021-12-02
Compile binary with goreleaser

bserem:
## The Problem/Issue/Bug:
#1678 Consider packaging native linux packages.

- [x] Proof of concept that binaries can be achieved with goreleaser
- [x] Create linux (and windows/mac) binaries
- [ ] Add binaries into github release with `goreleaser release`
- [ ] Create linux packages (deb etc)

- [Install goreleaser](https://goreleaser.com/install/) and go 1.16+
- Run `goreleaser release --snapshot --rm-dist`
- Check `dist` folder.

**See it in action: https://asciinema.org/a/1ys721qaRcSc3HhJbo474xt4a**
[asciicast](https://asciinema.org/a/1ys721qaRcSc3HhJbo474xt4a)

From here on, DEB files can *probably* be created, but that is not well documented (for me) in goreleaser docs.
It is also possible to automatically create binaries and add then in the release page of github, this is handled by goreleaser, but for the moment I was only looking for a proof of concept.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3414"></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# Compiled binaries by goreleaser
+/dist
 # Temporary build-tools artifacts to be ignored
 /.gotmp/
 /VERSION.txt
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,35 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
+    - go mod tidy
+    # you may remove this if you don't need go generate
+    - go generate ./...
+builds:
+  - env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
+      - darwin
+    main: ./cmd/ddev
+    goarch:
+      - amd64
+archives:
+  - replacements:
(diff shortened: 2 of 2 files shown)


rfay:
Thanks for the contribution and good thinking.

I've been looking at goreleaser for building and releasing for some time, but haven't been break loose from the safety net of building in a container, so that no matter where you're building, you get the same go version and you're always building on linux. 

I know goreleaser is super popular. Probably not willing to change the whole build and deployment scheme at this time. Also, this PR doesn't change the build... it just provides a goreleaser config.

bserem:
My goal was just a POC that it can work and maybe get a .deb file out of it. Not much, I know.
I can't do much about it running in a container, it is beyond my understanding. Maybe someone can continue from this POC and do better than me :)

rfay:
One of the great potentials of goreleaser is building packages. 
* https://github.com/drud/ddev/issues/1678

Would absolutely love to have you work on that, but doing it means big changes to the Makefile, and to deployment, etc.

rfay:
Thanks so much for taking the time to do this contribution. I'm closing this for now, but happy to go down this path if you think you can go farther with it, especially if you can build and deploy packages with the free version of goreleaser.

PR #3424, opened 2021-12-06
[Just experimenting] Remove db volume

weitzman:
Just exploring if our custom Db image works right if DDEV skips its usual volume. No reviews needed.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3424"></a>

### DIFF
--- a/pkg/ddevapp/templates.go
+++ b/pkg/ddevapp/templates.go
@@ -20,11 +20,6 @@ services:
     stop_grace_period: 60s
     working_dir: "{{ .DBWorkingDir }}"
     volumes:
-      - type: "volume"
-        source: mariadb-database
-        target: "/var/lib/mysql"
-        volume:
-          nocopy: true
       - type: "bind"
         source: "."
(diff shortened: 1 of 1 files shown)


rfay:
You're not interested in `omit_containers: [db]` just whether it can work if the db is not persistent/mounted on volume right?

weitzman:
Yes, not looking for omit-comtainers ... I am supplying own db image whose data is baked into the image (by design). DDEV's volume is removing all the tables that are in the image, and replacing with emptiness. I am currently using an additional service but that makes my db not work with snapshot, `ddev mysql`, PhpMyAdmin, etc. Maybe there is some way to coax these tools to use my `dbmass` service instead of `db`?

Anyway, this experiment didnt work. I still gets startup errors when my image is used as `db` service.

rfay:
This doesn't need to be PR for now, so closing again. Can just do builds locally or whatever. Local and gitpod build info in https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md

PR #3426, opened 2021-12-07
Use "ddev delete --omit-snapshot" instead of "ddev stop --remove-data"

douggreen:
## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3426"></a>

### DIFF
--- a/docs/users/extend/database_types.md
+++ b/docs/users/extend/database_types.md
@@ -22,5 +22,5 @@ mysql_version: 8.0
 
 * If you change the database type or version in an existing project, the existing database may not be compatible with your change, so you'll want to use `ddev export-db` to save a dump first.
 * When you change database type, it's best to destroy the existing database using `ddev delete --omit-snapshot` before changing, then use `ddev import-db` to import the db you already exported.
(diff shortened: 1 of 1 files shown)


docs/users/extend/database_types.md
```diff
@@ -21,6 +21,6 @@ mysql_version: 8.0
 ### Caveats
 
 * If you change the database type or version in an existing project, the existing database may not be compatible with your change, so you'll want to use `ddev export-db` to save a dump first.
-* When you change database type, it's best to destroy the existing database using `ddev stop --remove-data` before changing, then use `ddev import-db` to import the db you already exported.
+* When you change database type, it's best to destroy the existing database using `ddev delete --omit-snapshot` before changing, then use `ddev import-db` to import the db you already exported.
 * Despite those warnings, the db container will do its best to upgrade if you're upgrading from compatible versions. For example, a change from MariaDB 10.1 to 10.4 is likely to work and upgrade correctly, but your mileage may vary. If you change from MySQL 8.0 to MariaDB 5.5, all hell will break loose and you'll have to `ddev stop --remove-data --omit-snapshot` to get back where you were.
```
rfay, line 25:
```suggestion
* Despite those warnings, the db container will do its best to upgrade if you're upgrading from compatible versions. For example, a change from MariaDB 10.1 to 10.4 is likely to work and upgrade correctly, but your mileage may vary. If you change from MySQL 8.0 to MariaDB 5.5, all hell will break loose and you'll have to `ddev delete --omit-snapshot` to get back where you were.
```

rfay (approved):
Thanks!

PR #3427, opened 2021-12-07
Clean up ports in `.gitpod.yml`

shaal:
## The Problem/Issue/Bug:

Many ports belong to previous setup that is no longer needed in current ddev workspace in Gitpod.
Each port that is defined in `.gitpod.yml`, is added to the list of ports, and can be confusing for users -
image

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3427"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -64,34 +64,12 @@ ports:
   # Ignore host https port
   - port: 8443
     onOpen: ignore
-  - port: 3306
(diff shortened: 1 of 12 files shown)


shaal:
I opened this by mistake, here's the correct PR - https://github.com/rfay/ddev/pull/13

PR #3445, opened 2021-12-11
Use volumes and copies instead of bind-mounts with db container

rfay:
## The Problem/Issue/Bug:

To make a number of things work in new environments where bind-mounts don't exist, we have to stop using bind-mounts. 

This affects possible future usage in colima, lima, podman, and remote docker usage.

* Copy the /mnt/ddev_config into containers instead of bind-mount
* Move snapshot usage into docker volumes (which may speed up the whole process)

## Changes:
* Minimum docker version
* Warn instead of fail if wrong docker version
* Ability to use remote docker
* Ability to use work where read-write mounts are not available

## TODO
- [x] Place snapshots into their own volume
- [x] Compare performance of bind-mount approach and this one on database commands (import/export/snapshot/restore)
- [x] Consider using the non-compose Exec implementation in this code instead of app.Exec()
- [x] Rethink the push-config-into-volume approach being used right now. It's ugly and fragile. But could also just push into container. But surely something is better. Could start.sh pull something in that's waiting for it?
- [x] Document change to projects like solr that mount .ddev (mount the volume instead of bind-mount)
- [x] Use bind-mount in templates.go when a global command requires it - global or project no-bind-mounts.
  - [x] Rethink how apache/nginx configs get loaded on start
  - [x] Rethink how homeadditions gets loaded. (Is there a need for two execs for homeadditions?)
  - [ ] 
- [x] File a bug with Docker Desktop for Windows about the problem with untarring incorrectly. Use their own API to demonstrate it.
- [x] Use a script built into web container to do the make-executable work for Windows.
- [x] Experiment with pushing stdin into a container instead of using their API
- [x] Add some output showing when volumes are being pushed
- [x] Random use of GetContainerID() and GetContainerNameI() should be replaced by something simpler. GetContainer("db")
- [ ] Consider making `ddev start` always be `ddev restart`
- [x] Add github tests both with and without bind-mount
- [x] Compare normal and hardened start.sh in ddev-webserver, consider consolidating them
- [x] ddev_config volume is being created even when we're bind-mounting
- [x] `docker context` lookup should be secondary to $DOCKER_HOST
- [x] if `docker context` isn't there, it's not actually an error. Maybe we can move back to previous allowed versions? But since we're now being more generous with versions, maybe it doesn't matter.

- [ ] Manual test with
  - [ ] Colima
  - [x] Podman
  - [ ] Remote docker 
  - [x] Rancher Desktop

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3445"></a>


rfay:
Shocking that no-bind-mount beats the bind-mount behavior:



  | traditional/docker | nobindmount
-- | -- | --
import-db hugedb | 6m48s | 5m34s
snapshot hugedb | 4m2s | 3m0s
restore snapshot hugedb | 4m42s | 3m38s
import-db d9-umami | 4.3s | 4.6s
snapshot d9-umami | 15s | 8.9s
restore snapshot d9-umami | 26s | 23s

PR #3454, opened 2021-12-16
Remove unnecessary rules for static content

gilbertsoft:
## The Problem/Issue/Bug:

The configuration for php project type should be kept as simple as
possible to avoid conflicts with the various needs of the various
applications. It's easy possible to enhance the Nginx configuration
with a custom include in .ddev/nginx afterwards.

The static file configuration is removed.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3454"></a>

### DIFF
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-php.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-php.conf
@@ -52,15 +52,6 @@ server {
         fastcgi_param HTTPS $fcgi_https;
     }
 
-    # Expire rules for static content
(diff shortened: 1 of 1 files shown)


rfay (approved):
Seems to me these are probably a decendant of some rules intended for normal hosting.

gilbertsoft:
> Seems to me these are probably a descendant of some rules intended for normal hosting.

Exactly. For normal hosting this makes absolutely sense but for developement it can cause more troubles than advantages.

PR #3469, opened 2021-12-24
Don't build buildkite/circleci tests on docs pushes

rfay:
## The Problem/Issue/Bug:

We spend a lot of testing time on docs builds, that don't need to be tested.

Check to see if code has been changed in buildkite test.sh, if it has not, then just exit early
Same thing in CircleCI

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3469"></a>


rfay:
I'm going to go ahead and pull this, and we can watch the results carefully.

PR #3471, opened 2021-12-25
Updating the checkout directory from ~/tmp to /tmp

shaal:
## The Problem/Issue/Bug:
I am familiar with `/tmp`, but not with `~/tmp`

Update the checkout directory to `/tmp`

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3471"></a>


rfay (changes requested):
This doesn't work on Windows, where /tmp may not exist or be accessible. 

And please remember that these provider examples are just examples. Anybody can use anything tmp directory they want.

shaal:
Thanks!

PR #3482, opened 2021-12-28
MariaDB 10.1 doesn't work on arm64/m1 mac

funkvn:
MariaDB 10.1 is now out of support, so DDEV runs only with 10.2 or higher.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3482"></a>

### DIFF
--- a/docs/users/special-environments.md
+++ b/docs/users/special-environments.md
@@ -9,6 +9,6 @@ Docker on the Mac M1 requires the Apple Silicon version of Docker Desktop, which
 There are a few limitations of DDEV-Local on the Mac M1.
 
 * Only mysql 5.7 and 8.0 are available for arm64/Mac M1 users.
-* Because MariaDB does not publish arm64 packages or images for versions that are out of support, only MariaDB versions 10.1 and higher are supported by DDEV.
+* Because MariaDB does not publish arm64 packages or images for versions that are out of support, only MariaDB versions 10.2 and higher are supported by DDEV.
 
 Both [Mutagen](performance.md#using-mutagen) and [NFS mounting](performance.md#macos-nfs-setup) are supported and work great on the Mac M1.


rfay (changes requested):
The images are still there, see https://hub.docker.com/_/mariadb?tab=tags&page=1&name=10.1 - I doubt they'll remove them. 

It would take more than just the doc to remove the 10.1 option for arm64.

Are you having trouble with arm64 MariaDB 10.1?

funkvn:
> It would take more than just the doc to remove the 10.1 option for arm64.

well, but...

> Are you having trouble with arm64 MariaDB 10.1?

yes. It does not work with 10.1 and 10.2 works fine.
If there is no 10.1 anymore on https://mariadb.org/ (header) or https://mariadb.org/download/ – don't know about the details of removing 10.1 from the official support.

I got this error while `ddev start`:
```
#12 0.187 standard_init_linux.go:228: exec user process caused: exec format error
#12 ERROR: executor failed running [/bin/bash -c (groupadd --gid $gid "$username" || groupadd "$username" || true) && (useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' --uid $uid "$username" || useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' "$username" || useradd  -l -m -s "/bin/bash" --gid "$gid" --comment '' "$username")]: exit code: 1
```

A `docker run` failed as well:
```
$ docker run ab420cdc23d0                         
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
standard_init_linux.go:228: exec user process caused: exec format error
```

Switching to 10.2 (or 10.3/later) fixes that problem. 

Greetings
  Valentin

rfay:
Thanks for pointing out this problem. It was actually an error in the build process, fixed in 

* #3483 

I pushed the missing arm64 image for mariadb 10.1

Please `ddev poweroff` and try again with mariadb 10.1

rfay:
Looking forward to hearing from you if that sorted it out, but closing this for now. Feel free to comment here and let me know.

funkvn:
Thanks Randy! Now it works with MariaDB 10.1 fine.

I misunderstood the documentation in that way, that only supported versions of MariaDB are available, because only for those the images were created (by the vendor). Blame it on the non-native speaker. Sorry for the trouble.

rfay:
No, you didn't really misunderstand. Mostly the docs weren't completely correct.  Actually, it looks like we could now do 10.0 as well, but there hasn't really been any call for that. 

Again, thanks for catching this. Old versions of TYPO3 actually need 10.1, but that's the only old-db use I know of. 

Was there a particular reason you needed 10.1 @funkvn ?

PR #3495, opened 2022-01-04
add `mkdocs` VScode setup  [skip ci]

tyler36:
## The Problem/Issue/Bug:

DDEV uses `mkdocs` to generate the website. It uses additional plugins and a theme which are hard-coded in the config.

Remembering the setup is hard.

This PR provides a container for use with VSCode & with ["Remote: Containers"](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extenstion.

VSCode will then be able to open the repo in the customized Python continer which includes `mkdocs` with the required extensions.

- Install VSCode
- Open folder
- Install VSCode "Remote: containers" (it will be suggested automatically)
- VSCode should detect container config files and offer to re-open in container
- Alternatively, open the command pallete and select "Remote Containers: Open folder in container"
- In VScode terminal, type: `mkdocs serve`

## Versions

This works with VSCode desktop(local) versions. The remote containers extension is not supported in web versions of VSCode (Gitpod, Github vscode)

- VSCode: 1.63
- Remote - Continers: 0.209.6

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3495"></a>

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,21 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.209.6/containers/python-3/.devcontainer/base.Dockerfile
+
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+ARG VARIANT="3.10-bullseye"
+FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+
+# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
+ARG NODE_VERSION="none"
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
+
+# [Optional] If your pip requirements rarely change, uncomment this section to add them to the image.
+# COPY requirements.txt /tmp/pip-tmp/
+# RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt \
+#    && rm -rf /tmp/pip-tmp
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment this line to install global node packages.
+# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -19,3 +19,6 @@ RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/
 
 # [Optional] Uncomment this line to install global node packages.
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
+
+# The following code sets up DDEV's "mkdoc" environment
+RUN pip install mkdocs mkdocs-material mdx_truly_sane_lists mkdocs-git-revision-date-localized-plugin
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,48 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.209.6/containers/python-3
+{
+	"name": "Python 3",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"context": "..",
+		"args": { 
+			// Update 'VARIANT' to pick a Python version: 3, 3.10, 3.9, 3.8, 3.7, 3.6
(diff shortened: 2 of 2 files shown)


rfay:
Thanks! But you're going to make me install and use vscode when all I currently have to do is `mkdocs serve` ? :)

tyler36:
If you want to test it, yes. :)

I use it so I don't have to remember stuff as I've been working on the bookmarklet and other doc updates. If/when VScode online versions support the remote container extension it might make things easier.

Probably not worthy of inclusion, I know, but even as a closed PR, it will be documentated.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment this line to install global node packages.
+# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
+
+# The following code sets up DDEV's "mkdoc" environment
+RUN pip install mkdocs mkdocs-material mdx_truly_sane_lists mkdocs-git-revision-date-localized-plugin
```
rfay, line 24: I don't understand why we're installing these few things instead of installing everything in requirements.txt?

Also don't understand using pip instead of pip3?

rfay (changes requested):
Thanks for this contribution, and for scratching your itch! It would be worth getting it in there just for your contributions, but...

* Why not just use https://hub.docker.com/r/squidfunk/mkdocs-material/ or https://hub.docker.com/r/polinux/mkdocs instead of doing all this? And the build is really slow. I gave up on it and came here to make this comment.
* Didn't understand using pip instead of pip3, not using the requirements.txt
* I think this ends you up with your terminal inside the container, right? Confusing people because they expect it to be ddev since they're working on ddev?
* This requires you to install python inside the container, you'd need to add that to the Dockerfile, `sudo apt-get install python3 python3-venv python3-pip`
*

rfay:
I opened up vscode for another reason today and this clogged everything up. Could we please use a simple command-line solution to run mkdocs in a container? Shouldn't be hard. Happy to push it into project namespace on dockerhub.

.devcontainer/Dockerfile
tyler36: I'm still learning docker & how to best intergrate it into my workflow.

Instead of installing the requirements for mkdocs in WSL, I thought I'd try the VSCode remote container to install them.
As I hit the errors listed below, I updated the `RUN pip` line to resolve them:

```
ERROR    -  Config value: 'theme'. Error: Unrecognised theme name: 'material'. The available installed themes are: mkdocs, readthedocs
ERROR    -  Config value: 'markdown_extensions'. Error: Failed loading extension "mdx_truly_sane_lists".
ERROR    -  Config value: 'plugins'. Error: The "git-revision-date-localized" plugin is not installed
```

For my setup, this PR spun up a working mkdocs server. After reading through your comments though, I see my process has several flaws.

tyler36:
The following command runs a dockerized `mkdocs` with live-reload on the DDEV docs 

```
docker run \
    -ti \
    -p 9000:9000 \
    -v "${PWD}:/docs" \
    -e "ADD_MODULES=mkdocs-material mdx_truly_sane_lists mkdocs-git-revision-date-localized-plugin" \
    -e "LIVE_RELOAD_SUPPORT=true" \
    -e "FAST_MODE=true" \
    -e "DOCS_DIRECTORY=./docs" \
    -e "DEV_ADDR=0.0.0.0:9000" \
    polinux/mkdocs
```

rfay:
I like it. We could actually just put that into the Makefile.

rfay:
How do you prefer to go forward with this? Want to add a `mkdocs-serve` command to the Makefile that just does this docker command? 

```Makefile
mkdocs-serve:
    docker run \
    -ti \
    -p 9000:9000 \
    -v "${PWD}:/docs" \
    -e "ADD_MODULES=mkdocs-material mdx_truly_sane_lists mkdocs-git-revision-date-localized-plugin" \
    -e "LIVE_RELOAD_SUPPORT=true" \
    -e "FAST_MODE=true" \
    -e "DOCS_DIRECTORY=./docs" \
    -e "DEV_ADDR=0.0.0.0:9000" \
    polinux/mkdocs
```

(sorry about the spaces for indentation and Makefile requires tab.

rfay:
Or maybe `if command -v mkdocs; then mkdocs serve; else <docker>; fi`

tyler36:
Not sure I understand the pros/cons of each.

rfay:
So `make mkdocs-serve` would require no system setup. 

It would just be a command like

```
mkdocs-serve: 
    if command -v mkdocs >/dev/null ; then mkdocs serve; else docker run \
    -ti \
    -p 9000:9000 \
    -v "${PWD}:/docs" \
    -e "ADD_MODULES=mkdocs-material mdx_truly_sane_lists mkdocs-git-revision-date-localized-plugin" \
    -e "LIVE_RELOAD_SUPPORT=true" \
    -e "FAST_MODE=true" \
    -e "DOCS_DIRECTORY=./docs" \
    -e "DEV_ADDR=0.0.0.0:9000" \
    polinux/mkdocs
fi
``` 

Requires no setup or installation. Execute with `make mkdocs-serve`

rfay:
Please review https://github.com/drud/ddev/pull/3541, thanks!

tyler36:
I checked out your PR.

> ... if mkdocs isn't installed, `make mkdocs-serve`

```shell
make mkdocs-serve

Command 'make' not found, but can be installed with:

sudo apt install make        # version 4.2.1-1.2, or
sudo apt install make-guile  # version 4.2.1-1.2
```

rfay:
Thanks so much for pioneering all this! Since it's now in the docs along with the raw docker command to use (not just make) I'd rather have this contribution somewhere else. Maybe Stack Overflow? I'm so happy that you have gone through the effort to make mkdocs work, and the excellent contributions you've been able to do as a result.

PR #3503, opened 2022-01-08
Added missing sudo to wsl2 docker install instructions

chellman:
There appears to be a missing `sudo` in the install commands for docker in WSL2 — I got "Could not open lock file /var/lib/dpkg/lock-frontend" before adding the `sudo` shown here.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3503"></a>


rfay (approved):
Yup, thanks!

PR #3510, opened 2022-01-11
v1.19.x vs v1.18.x compatibility with 3rd party services, especially solr

rfay:
## The Problem/Issue/Bug:

* People on versions below v1.19 are referred to the HEAD version of docker-compose.solr.yaml, which does not work until v1.19.x
* People upgrading to v1.19 with solr and other 3rd party services have to *add* the `networks:` line to make their project work.
* The failure mode if people are missing the networks line is awful - ddev-router has a fatal that's hard to understand and says nothing about solr.

- [x] Add comment in `docker-compose.solr.yaml` explaining to remove the networks line below v1.19
- [x] Figure out how to reduce the problems with people hitting the ddev-router issue with solr

- [x] Read docs
- [x] Try adding a service without the required `networks` line, `ddev start` and get the error
- [x] Try adding a service *with* the required `networks` line and `ddev start` and see no error

* Nothing added at this point. It's possible we should add a test for the network-checking behavior

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3510"></a>

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -8,10 +8,10 @@ on:
   pull_request:
     paths:
       - "go.*"
-      - "pkg"
-      - "cmd"
+      - "pkg/**"
+      - "cmd/**"
       - "Makefile"
-      - "vendor"
+      - "vendor/**"
 
 env:
   BUILDKIT_PROGRESS: plain
--- a/cmd/ddev/cmd/auth-ssh_test.go
+++ b/cmd/ddev/cmd/auth-ssh_test.go
@@ -22,30 +22,32 @@ func TestCmdAuthSSH(t *testing.T) {
 		t.Skip("Skipping TestCmdAuthSSH because expect scripting tool is not available")
 	}
 
-	testDir, _ := os.Getwd()
-	defer cmd.TestSites[0].Chdir()()
-
-	_, err := exec.RunCommand(cmd.DdevBin, []string{"start"})
+	origDir, _ := os.Getwd()
+	err := os.Chdir(cmd.TestSites[0].Dir)
 	require.NoError(t, err)
-	// nolint: errcheck
-	defer exec.RunCommand(cmd.DdevBin, []string{"stop", "--remove-data", "--omit-snapshot"})
+
+	app, err := ddevapp.NewApp("", false)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		err = app.Stop(true, false)
+		assert.NoError(err)
+		err = os.Chdir(origDir)
+		assert.NoError(err)
+		err = dockerutil.RemoveContainer("test-cmd-ssh-server", 10)
+		assert.NoError(err)
+	})
 
 	// Delete any existing identities from ddev-ssh-agent
 	_, err = exec.RunCommand("docker", []string{"exec", "ddev-ssh-agent", "ssh-add", "-D"})
 	assert.NoError(err)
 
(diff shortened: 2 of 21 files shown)


rfay:
Here's what I can think of so far @jonaseberle - thanks so much! Will look at your proposals for release notes now though.

jonaseberle (approved):

heddn:
The notes are specific to solr. But what about redis, memcache, or any of a variety of other external services?

jonaseberle:
This is actually just version-specific test data.
But the `solr` recipe is linked here https://ddev.readthedocs.io/en/stable/users/extend/additional-services/#apache-solr. It already mentions that is going to be replaced by https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/solr
or am I missing something?

rfay:
@jonaseberle @heddn I requested review again because I did add a check for the networks element. Would appreciate if you could review and do the manual test. I'll put a link to the artifacts here as soon as it's built.

jonaseberle:
Sorry, I first misunderstood the direction this PR took ;)

It works fine and also works find if I happen to have additional custom networks.

But I seem to always get cached results. Here's my test setup:
`docker-compose.redis.yaml`:
```yaml
services:
  redis:
    container_name: ddev-${DDEV_SITENAME}-redis
    # networks: [default, ddev]
    image: "redis"
    labels:
      com.ddev.approot: $DDEV_APPROOT
```

* `ddev start` - see the error message
* uncomment the `networks:` line
* `ddev start` - still see the error message
* again `ddev start` - now it is ok

Also the other way around (removing the `networks:` line by e.g. checking out another branch of a project) yields the result from the previous `ddev start`.

I currently don't understand why, but `app.ComposeYaml` seems to contain cached data.

rfay:
Thanks, I'll see if I can figure out why that happens. I know that the .ddev/.ddev-docker-compose-full.yaml doesn't get generated until `ddev start` but don't know how caching will happen otherwise, but I'll chase it.

Another thing I noticed in seeing test results is that the called code does a `util.Failed()`, terminating ddev, and it should instead bubble up the error to a more appropriate point.

jonaseberle:
You fixed the caching problem :)


A simple
```
services:
  redis:
    container_name: ddev-${DDEV_SITENAME}-redis
    networks: [default, ddev_default]
    image: "redis"
    labels:
      com.ddev.approot: $DDEV_APPROOT
```
yields `service "redis" refers to undefined network ddev_default: invalid compose project` in ddev v1.1*8*.

But adding it as a global network fixes that
```
# ddev v1.18 compatibility
networks:
  ddev_default:
    external: true
```

So yes, this saves v1.18 users from having to create a docker network themselves! Splendid, thank you! I am really happy.

I also tested with a Selenium setup (with `VIRTUAL_HOST` and `HTTP_EXPOSE`).

jonaseberle (approved):

rfay:
Ah, the problem is that the v1.18 network is defined like this:

```yaml
networks:
  default:
    name: ddev_default
    external: true
```

So I didn't quite do it right. But not sure how to do it right, because of the namespace conflict over "default"

jonaseberle:
Yeah, I also tried if I could get rid of the external network definition but in the end couldn't find a way.
I am really happy that you integrated docker-compose by the way - that makes the testing matrix a lot smaller :)

Anyways, this is easy enough IMHO. We can give a good recommendation so people can prepare services beforehand and your error message smoothly guides casual users.

PR #3519, opened 2022-01-13
Use regular colima release instead of HEAD, change composer create-project approach

rfay:
## The Problem/Issue/Bug:

Previously colima installation with `--HEAD` was required, but the needed things have made it into regular release.

- [x] Change docs and tests to use regular colima brew install
- [x] Use mutagen for tests since colima's sshfs is quite problemmatic.
- [x] Docs should recommend/require mutagen

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3519"></a>


rfay:
TestComposer doesn't seem to work in latest colima/lima. Real oddities. It's actually `ddev composer create` that doesn't work, and `composer create-project drupal/recommended-project /tmp/somepath` seems to create the project in a subdirectory of /tmp/somepath instead of in /tmp/somepath itself.

PR #3523, opened 2022-01-15
Create upload dir when creating settings, fixes #3521

### LINKED ISSUE #3521: ddev should create upload_dir if it doesn't exist (before generating docker-compose config)
On mutagen, with latest changes, the upload_dir is bind-mounted into the container (instead of copying it all using mutagen). But if it doesn't exist, then the mount isn't created.

ddev should create the directory if upload_dir is specified, just like it creates settings.php. Should do no harm

rfay:
## The Problem/Issue/Bug:

* Since with mutagen syncing we now bind-mount the upload_dir, but since it doesn't get mounted unless it exists, we need to make sure it exists if it's defined.

* Create the upload_dir with the settings files

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3523"></a>


rfay:
TestMutagenSimple is [failing all the time on wsl2](https://buildkite.com/drud/ddev-wsl2/builds/1645#7b653725-29c8-47b9-ab61-2094e606d316) in the same place. So something interesting is going on here.

PR #3541, opened 2022-01-22
Add make mkdocs-serve

rfay:
## The Problem/Issue/Bug:

@tyler36 pioneered all of this while trying to test docs as they would really look in https://github.com/drud/ddev/pull/3495 - so this is all credited to him. 

`make mkdocs-serve` will run either local mkdocs or run a docker command to serve it.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3541"></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,17 @@ mkdocs:
 		echo "Not running mkdocs because it's not installed"; \
 	fi
 
+# To see what the docs build will be, you can use `make mkdocs-serve`
+# It works best with mkdocs installed, `pip3 install mkdocs`,
+# see https://www.mkdocs.org/user-guide/installation/
+# But it will also work using docker.
+mkdocs-serve:
+	if command -v mkdocs >/dev/null ; then \
+  		mkdocs serve; \
+	else \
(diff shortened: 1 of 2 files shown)


tyler36:
Sorry ... originally posted in wrong issue.

```shell
make mkdocs-serve

Command 'make' not found, but can be installed with:

sudo apt install make        # version 4.2.1-1.2, or
sudo apt install make-guile  # version 4.2.1-1.2
```

rfay:
Interesting, you're on linux without `make` there. I would have thought it would be everywhere. But yes, `make` is required, but I would expect people ambitious enough to run mkdocs to have make at least. Probably the most common way to install it on Linux would be `sudo apt-get update && sudo apt-get install build-essential`, which installs all the normal things a dev needs on Linux.

tyler36:
Tested from WSL

rfay:
Yup, Linux.

rfay:
I went ahead and added the raw docker command into the docs as well, so now people without make can use this.

PR #3557, opened 2022-01-28
Add full support for Postgresql, fixes #2187

### LINKED ISSUE #2187: Support Postgresql in ddev
When DDEV is building a Drupal site (Drupal 7 in my case), it's using a settings template described here:

https://github.com/drud/ddev/blob/8be3d67394f157bbe3554bca0e8e0be9f62aaf7d/pkg/ddevapp/drupal.go#L53

The database driver is hard coded, which requires me to override the database connection on my local settings, when I use the PostgreSQL contrib from here: https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/postgres

This is the generated code in my settings.ddev.php file:

`$databases['default']['default'] = array(`
`  'database' => "db",`
`  'username' => "db",`
`  'password' => "db",`
`  'host' => $host,`
`  'driver' => "mysql",`
`  'port' => $port,`
`  'prefix' => "",`
`);`

Since I only need one database server and I'm using the postgresql from above, I disabled the db + dba  containers and added postgresql-client to my config.yaml:

`webimage_extra_packages: [postgresql-client]`
`omit_containers: ["db","dba"]`

However, I have no way to configure the database host, port nor driver. Would be great to have them as an option. Because right now I have to add an ugly extra setting file which obfuscates my project repository and settings.php:

`// #ddev-custom: load postgres db connection (not supported by DDEV config yet)`
`if (getenv('IS_DDEV_PROJECT') == 'true') {`
`  include dirname(__FILE__) . '/settings.ddev.postgres.php'`
`}`

Which contains:
`<?php`
`# DDEV POSTGRES SETTINGS`
`$databases['default']['default'] = array(`
`  'database' => "db",`
`  'username' => "db",`
`  'password' => "db",`
`  'host' => 'postgres',`
`  'port' => '5432',`
`  'driver' => 'pgsql',`
`  'prefix' => '',`
`);`

This solution works, but isn't beautiful - especially since Drupal supports mysql + pgsql by default, so why doesn't DDEV?  Adding full postgresql support for ddev config is a lot to ask, so I rather ask for the little config change (support of config variables for `host`,`port` and `driver`). This would be much appreciated.

rfay:
## The Problem/Issue/Bug:

* #2187

Full support for Postgresql v10-14, with snapshots and all

Run it through its paces

All key tests updated to test Postgres

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3557"></a>


tyler36:
Looking forward to trying this out in an beta.

PR #3573, opened 2022-02-04
Change keys of ddev version json object for more consistency, fixes #3572

### LINKED ISSUE #3572: Optimize JSON Output of ddev version Command
I currently try to parse the output of ddev version -j in an java application. Sadly the keys of the output are quite inconsistent as you can see:
```json
{
        "DDEV version": "v1.18.2",
        "architecture": "amd64",
        "build info": "BUILDINFO should have new info",
        "db": "drud/ddev-dbserver-mariadb-10.3:v1.18.2",
        "dba": "phpmyadmin:5",
        "ddev-ssh-agent": "drud/ddev-ssh-agent:v1.18.0",
        "docker": "20.10.12",
        "docker-compose": "v2.2.2",
        "mutagen": "0.12.0",
        "os": "linux",
        "router": "drud/ddev-router:v1.18.2",
        "web": "drud/ddev-webserver:v1.18.2"
}
``` 

The output of ddev status -j for example is consistently using snake_case. Could you please adopt this here?

```json
{
        "ddev_version": "v1.18.2",
        "architecture": "amd64",
        "build_info": "BUILDINFO should have new info",
        "db": "drud/ddev-dbserver-mariadb-10.3:v1.18.2",
        "dba": "phpmyadmin:5",
        "ddev_ssh_agent": "drud/ddev-ssh-agent:v1.18.0",
        "docker": "20.10.12",
        "docker_compose": "v2.2.2",
        "mutagen": "0.12.0",
        "os": "linux",
        "router": "drud/ddev-router:v1.18.2",
        "web": "drud/ddev-webserver:v1.18.2"
}
``` 

This would add to the overall consistency and save me some hassle in this specific use case.

nico-loeber:
The Problem/Issue/Bug:
Inconsistent keys with spaces and alternating capitalisation
make the output hard to parse.

How this PR Solves The Problem:
To be consistent with the output of ddev status -j the keys get changed to snake_case.

Manual Testing Instructions:
N/A

Automated Testing Overview:
The test case has been adapted accordingly.

Related Issue Link(s):
* https://github.com/drud/ddev/issues/3572

Release/Deployment notes:
<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3573"></a>

### DIFF
--- a/cmd/ddev/cmd/cmd_version_test.go
+++ b/cmd/ddev/cmd/cmd_version_test.go
@@ -25,14 +25,14 @@ func TestCmdVersion(t *testing.T) {
 	raw, ok := versionData["raw"].(map[string]interface{})
 	require.True(t, ok, "raw section wasn't found in versioninfo %v", out)
 
-	assert.Equal(version.DdevVersion, raw["DDEV version"])
+	assert.Equal(version.DdevVersion, raw["ddev_version"])
 	assert.Equal(version.WebImg+":"+version.WebTag, raw["web"])
 	assert.Equal(version.GetDBImage(nodeps.MariaDB), raw["db"])
 	assert.Equal(version.GetDBAImage(), raw["dba"])
 	dockerVersion, _ := version.GetDockerVersion()
 	assert.Equal(dockerVersion, raw["docker"])
 	composeVersion, _ := version.GetDockerComposeVersion()
-	assert.Equal(composeVersion, raw["docker-compose"])
+	assert.Equal(composeVersion, raw["docker_compose"])
 
 	assert.Contains(versionData["msg"], version.DdevVersion)
 	assert.Contains(versionData["msg"], version.WebImg)
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -93,28 +93,28 @@ func GetVersionInfo() map[string]string {
 	var err error
 	versionInfo := make(map[string]string)
 
-	versionInfo["DDEV version"] = DdevVersion
+	versionInfo["ddev_version"] = DdevVersion
 	versionInfo["web"] = GetWebImage()
 	versionInfo["db"] = GetDBImage(nodeps.MariaDB)
 	versionInfo["dba"] = GetDBAImage()
 	versionInfo["router"] = RouterImage + ":" + RouterTag
-	versionInfo["ddev-ssh-agent"] = SSHAuthImage + ":" + SSHAuthTag
-	versionInfo["build info"] = BUILDINFO
+	versionInfo["ddev_ssh_agent"] = SSHAuthImage + ":" + SSHAuthTag
+	versionInfo["build_info"] = BUILDINFO
 	versionInfo["os"] = runtime.GOOS
 	versionInfo["architecture"] = runtime.GOARCH
 	if versionInfo["docker"], err = GetDockerVersion(); err != nil {
 		versionInfo["docker"] = fmt.Sprintf("failed to GetDockerVersion(): %v", err)
 	}
-	if versionInfo["docker-platform"], err = GetDockerPlatform(); err != nil {
-		versionInfo["docker-platform"] = fmt.Sprintf("failed to GetDockerPlatform(): %v", err)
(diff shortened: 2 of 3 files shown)


rfay:
Looks like TestCmdVersion needs some more work. You can test this a number of ways, including running in gitpod per the link in the OP.  Also see https://github.com/drud/ddev/blob/master/docs/developers/building-contributing.md

nico-loeber:
Oh yes, I see there are references in lines 28 and 35. Sorry is my 2nd time working on a golang project and I thought all test cases would in the same directory as the file itself. But it looks like this is just the case for unit tests.

As this test seems to use the actual binary and is testing its output I need to to install the artifacts locally to get green tests?
This is the common workflow? To be honest I am a bit overwhelmed installing the right artifacts on my system...
I think I will try to fix it "blind" and try it out in gitpod after pushing if this is okay.

nico-loeber:
I have noticed that chaning the keys of the raw array does also effect the human readable output. In case we would like to keep a pretty formatted label here we could add something like a snakeCaseToSentenceCase method call in cmd_version.go:60.

What do you think?

rfay:
The thing that bothers me the most is the change from `-` where it's relevant. The name of the `ddev-ssh-agent` is actually `ddev-ssh-agent`. So changing to underscores is odd there.

nico-loeber:
> The thing that bothers me the most is the change from - where it's relevant. The name of the ddev-ssh-agent is actually ddev-ssh-agent. So changing to underscores is odd there.

Yes I know what you mean. Stuff like docker-compose are fixed expressions.
I think we just encountered a concept weakness. There is no distinction between the output made for humans or for the communication between software. I absolutly agree, that "DDEV version" is better to read than ddev_version and that docker-compose is a common name.
But thinking as a programmer you can't name your variables like that as docker-compose would cause substracting compose from docker in most languages. Thats why variables are mostly dromedaCase or snake_case written.
Thinking of JSON as an API it makes absolutely sense to use a format that is as compatible as possible.

Sadly the current implementation of --json-output switches just the logger to JSON Output. This might be useful if you have monitoring software to collect the messages but it is not well suited as an API as the purpose of log messages is the the presentation to humans adn the data bag is made for easy filtering and giving additional context reading those log messages in some tool.

For perfect API usage --json-output would output only the json representation of an (go lang) object, probably the current raw data, and would have no wrapper by the logger, so it can be reconstructed to an qual object in the consuming software.
Warnings and Errors would be written to stderr output channel as only stdout is consumed for parsing.
The human Readable output can be created from the same object but probably with some manual optimization for data presentation as it is already done, table stlying, manually setting labels and so on.

So there are some options now:
1. Rejecting this pull request and doing nothing (The lazy solution)
2. Adding a mapping for labels in human readable output like ['ddev_version' => 'DDEV version'] to this pull request. This would ensure the same human readable output like we had before (pragmatic solution)
3. Rejecting this pull request and adding an optimized API for the describe and version commands according to my suggestions above, keeping --json-output as it is for Monitoring tools (Ideal solution for API integration)

Solution 3 would also address https://github.com/drud/ddev/pull/3574 as the warning would no longer be part of the stdout.
For the implementation of version 3 I would need some help but it looks possible without causing any bigger BC breaks.

rfay:
* Warnings and errors should absolutely go to stderr, but that may be easier said than done. 
* Your suggestion #3 would be super interesting and not harmful IMO, but might be more than you were hoping to undertake.

I'm quite sure that we can do *something* that is good for your situation. The reality is I don't know of many consumers of the json format. [ddev-ui](https://github.com/drud/ddev-ui) did/does, but I don't know of anyone using it. There are probably several people with one-off solutions that use jq with `ddev describe -j` and `ddev list -j`.

rfay:
Overall, I'm OK with changing the "DDEV version" to something more appropriate, but would be a little reluctant to change some of the others. "DDEV version" seems like a poor choice. I think a long time ago it was confusing because there were so *many* versions in there that it wanted to be differentiated.

rfay:
We both decided this had too many tradeoffs at this point, and @nico-loeber gave permission to close this for now.

PR #3574, opened 2022-02-05
Remove warning about missing SegmentKey in json mode

nico-loeber:
## The Problem/Issue/Bug:
The warning about the missing SegmentKey generates a second obejct on output which causes it to be invalid for parsers.

In json output mode the warning is no longer triggert.

Execute ddev describe -j on a self build ddev binary. The following warning must not appear:
```json
{
  "level": "warning",
  "msg": "Instrumentation is opted in, but SegmentKey is not available.",
  "time": "2022-02-05T16:52:26+01:00"
}
```

N/A

## Related Issue Link(s):
N/A

## Release/Deployment notes:
N/A

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3574"></a>

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -167,7 +167,7 @@ func init() {
 }
 
 func instrumentationNotSetUpWarning() {
-	if version.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
+	if !output.JSONOutput && version.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
 		output.UserOut.Warning("Instrumentation is opted in, but SegmentKey is not available.")
 	}
 }
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -168,7 +168,7 @@ func init() {
 
 func instrumentationNotSetUpWarning() {
 	if !output.JSONOutput && version.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
-		output.UserOut.Warning("Instrumentation is opted in, but SegmentKey is not available.")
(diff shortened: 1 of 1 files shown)


rfay (approved):
This looks perfect. Please just manually check to see that the warning shows in non-json mode still. Thanks!

nico-loeber:
The warning is still shown in the human readable output of the commands:

image

Can be tested in gitpod

- run ddev poweroff
- edit .ddev/config.yml
- set `instrumentation_opt_in: true`
- run ddev start
- run ddev describe
- run dddev describe --json-output

rfay:
Awesome. Let me know when you're happy with it and I'll pull (after tests run)

nico-loeber:
> Awesome. Let me know when you're happy with it and I'll pull (after tests run)

Ready when you are

rfay:
The weird thing here is that output.UserOut.Warning() should already have been outputting json and not breaking the json output. Maybe it's not properly initialized at this point.

cmd/ddev/cmd/root.go
```diff
@@ -167,7 +167,7 @@ func init() {
 }
 
 func instrumentationNotSetUpWarning() {
-	if version.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
+	if !output.JSONOutput && version.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
 		output.UserOut.Warning("Instrumentation is opted in, but SegmentKey is not available.")
```
rfay, line 171:
```suggestion
		output.UserOut.Warning("Instrumentation is opted in, but SegmentKey is not available. This usually means you have a locally-built ddev binary or one from a PR build. It's not an error. Please report it if you're using an official release build.")
```

rfay (changes requested):
Hmm, I'd love to figure out how to solve the json format problem, but I guess we'll be OK with the current approach.

The one thing we can surely do though is to make this message less concerning to people who are testing PR builds or local builds. What do you think about the suggested terminology here?

rfay:
I'm sure you already know that your code can set DDEV_NO_INSTRUMENTATION=true before interacting with ddev and also solve all this.

PR #3581, opened 2022-02-08
Remove duplicate networks key in docs

webflo:
## The Problem/Issue/Bug:

Small typo in docs. The network key is already defined a few lines above.


jonaseberle (approved):

rfay:
Thanks. BTW, I'm going to see if I can find a way to make this unnecessary by having ddev add this info.

PR #3589, opened 2022-02-09
Command `docker context show` not available on all docker client versions

gambry:
## The Problem/Issue/Bug:
According to [official Docker CLI documentation](https://docs.docker.com/engine/reference/commandline/context/), `docker context show` command does not exist. 

Replacing it with `docker context ls`.

To replicate run `docker context show` on latest docker CLI. 
Note: apparently the command works on "some" docker CLI installation. Still the command is not documented in the official docs.

## Release/Deployment notes:
This doesn't affect anything. It's just a documentation update, and not related to any setting changes but just how to check the docker active context.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3589"></a>


rfay:
Thanks!

rfay (approved):
Thanks!

PR #3597, opened 2022-02-12
Remove special handling for gitpod getting host.docker.internal

rfay:
## The Problem/Issue/Bug:

Xdebug wasn't working any more on gitpod because they changed the networking situation. Now the networking is normal Linux, so we don't need the previous technique (which now doesn't work).

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3597"></a>

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -65,7 +65,7 @@
             "hostname": "0.0.0.0",
             "port": 9003,
             "pathMappings": {
-                "/var/www/html": "${workspaceRoot}/d9simple"
+                "/var/www/html": "${workspaceRoot}/../d9simple"
             }
         }
     ]
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -65,7 +65,7 @@
             "hostname": "0.0.0.0",
             "port": 9003,
             "pathMappings": {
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -65,7 +65,7 @@
             "hostname": "0.0.0.0",
             "port": 9003,
             "pathMappings": {
-                "/var/www/html": "${workspaceRoot}"
+                "/var/www/html": "${workspaceRoot}/d9simple"
```
shaal, line 68:
```suggestion
                "/var/www/html": "${workspaceRoot}/../d9simple"
```
`workspaceRoot` = `/workspace/ddev`
But the location of d9simple directory is `/workspace/d9simple`

shaal (changes requested):
1 small change in `.vscode/launch.json`

rfay:
I too found it awkward to test xdebug, but my technique was to `ddev stop --unlist` and `mv /workspace/d9 /workspace/ddev` and then `ddev start` again. That way I could use vscode. 

It seems like the d9 used to be in subdirectory of ddev by default, maybe it should go back there, just for times like this. Most things don't need debugging of PHP but this one sure did.

PR #3604, opened 2022-02-14
install_ddev.sh script should handle `HEAD` option for latest

shaal:
## The Problem/Issue/Bug:
The `install_ddev.ssh` support only released versions, but doesn't support getting the latest unreleased version - `HEAD` 

Adding `HEAD` as a version option, ie. `install_ddev.sh HEAD`

Run `scripts/install_ddev.sh HEAD`

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3604"></a>

### DIFF
--- a/scripts/install_ddev.sh
+++ b/scripts/install_ddev.sh
@@ -153,7 +153,9 @@ fi
 curl -fsSL "https://raw.githubusercontent.com/${GITHUB_USERNAME}/ddev/master/scripts/macos_ddev_nfs_setup.sh" -o "${TMPDIR}/macos_ddev_nfs_setup.sh" || (printf "${RED}Failed downloading "https://raw.githubusercontent.com/${GITHUB_USERNAME}/ddev/master/scripts/macos_ddev_nfs_setup.sh"${RESET}\n" && exit 110)
 
 cd $TMPDIR
-$SHACMD -c "$SHAFILE"
+if [ ${VERSION} != "HEAD" ]; then
+  $SHACMD -c "$SHAFILE"
+fi
 tar -xzf $TARBALL
 
 printf "${GREEN}Download verified. Ready to place ddev and mkcert in your /usr/local/bin.${RESET}\n"
--- a/scripts/install_ddev.sh
+++ b/scripts/install_ddev.sh
@@ -156,7 +156,11 @@ cd $TMPDIR
 if [ ${VERSION} != "HEAD" ]; then
   $SHACMD -c "$SHAFILE"
 fi
-tar -xzf $TARBALL
+if [ ${VERSION} != "HEAD" ]; then
+  tar -xzf ${TARBALL}
+else 
+  unzip ${TARBALL}
+ fi
 
 printf "${GREEN}Download verified. Ready to place ddev and mkcert in your /usr/local/bin.${RESET}\n"
 
--- a/scripts/install_ddev.sh
+++ b/scripts/install_ddev.sh
@@ -139,7 +139,7 @@ if ! docker --version >/dev/null 2>&1; then
     printf "${YELLOW}Docker is required for ddev. Please see https://ddev.readthedocs.io/en/stable/#docker-installation.${RESET}\n"
 fi
 
-if [ "$VERSION" == 'HEAD' ]; then 
+if [ "$VERSION" = 'HEAD' ]; then 
   TARBALL="$FILEBASE.zip"
 else
   TARBALL="$FILEBASE.$VERSION.tar.gz"
--- a/scripts/install_ddev.sh
+++ b/scripts/install_ddev.sh
@@ -99,7 +99,7 @@ if [ $# -ge 1 ]; then
   VERSION=$1
 fi
 
-if [ "$VERSION" == 'HEAD' ]; then
+if [ "$VERSION" = 'HEAD' ]; then
   RELEASE_BASE_URL="https://nightly.link/drud/ddev/workflows/master-build/master"
 else
   RELEASE_BASE_URL="https://github.com/${GITHUB_USERNAME}/ddev/releases/download/$VERSION"
--- a/scripts/install_ddev.sh
+++ b/scripts/install_ddev.sh
@@ -125,7 +125,7 @@ else
 fi
 
 if [ "$VERSION" == 'HEAD' ]; then
-  FILEBASE="${FILEBASE//-/_}"
+  FILEBASE="${FILEBASE//_/-}"
   USE_ARCH=1
 else
   USE_ARCH=$(semver_compare "${VERSION}" "v1.16.0-alpha4")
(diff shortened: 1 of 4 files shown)


rfay:
Awesome, thanks!

rfay:
Minor complaints from docs checker.

rfay (changes requested):
Unfortunately, I get 

```
$ bash ./install_ddev.sh HEAD
curl: (22) The requested URL returned error: 404
Failed downloading https://nightly.link/drud/ddev/workflows/master-build/master/ddev_macos-arm64.zip
```

The correct link would have been:
https://nightly.link/drud/ddev/workflows/master-build/master/ddev-macos-arm64.zip

Looks like you got an underscore where a hyphen belonged?

scripts/install_ddev.sh
```diff
@@ -98,12 +98,19 @@ VERSION=$LATEST_VERSION
 if [ $# -ge 1 ]; then
   VERSION=$1
 fi
-RELEASE_BASE_URL="https://github.com/${GITHUB_USERNAME}/ddev/releases/download/$VERSION"
 
-rv=$(semver_compare "${VERSION}" "v1.10.0")
-if [[ ${rv} -lt 0 ]]; then
-  printf "${RED}Sorry, this installer does not support specifying versions of ddev prior to v1.10.0${RESET}\n"
-  exit 1
+if [ "$VERSION" == 'HEAD' ]; then
```
rfay, line 102:
```suggestion
if [ "$VERSION" = 'HEAD' ]; then
```

scripts/install_ddev.sh
```diff
@@ -117,29 +124,38 @@ else
...
 fi
 
-
 if ! docker --version >/dev/null 2>&1; then
     printf "${YELLOW}Docker is required for ddev. Please see https://ddev.readthedocs.io/en/stable/#docker-installation.${RESET}\n"
 fi
 
-TARBALL="$FILEBASE.$VERSION.tar.gz"
-SHAFILE="$TARBALL.sha256.txt"
+if [ "$VERSION" == 'HEAD' ]; then
```
rfay, line 142:
```suggestion
if [ "$VERSION" = 'HEAD' ]; then 
```

scripts/install_ddev.sh
```diff
@@ -117,29 +124,38 @@ else
...
 
 curl -fsSL "$RELEASE_BASE_URL/$TARBALL" -o "${TMPDIR}/${TARBALL}" || (printf "${RED}Failed downloading $RELEASE_BASE_URL/$TARBALL${RESET}\n" && exit 108)
-curl -fsSL "$RELEASE_BASE_URL/$SHAFILE" -o "${TMPDIR}/${SHAFILE}" || (printf "${RED}Failed downloading $RELEASE_BASE_URL/$SHAFILE${RESET}\n" && exit 109)
+if [ "$VERSION" != 'HEAD' ]; then
+  curl -fsSL "$RELEASE_BASE_URL/$SHAFILE" -o "${TMPDIR}/${SHAFILE}" || (printf "${RED}Failed downloading $RELEASE_BASE_URL/$SHAFILE${RESET}\n" && exit 109)
+fi
 curl -fsSL "https://raw.githubusercontent.com/${GITHUB_USERNAME}/ddev/master/scripts/macos_ddev_nfs_setup.sh" -o "${TMPDIR}/macos_ddev_nfs_setup.sh" || (printf "${RED}Failed downloading "https://raw.githubusercontent.com/${GITHUB_USERNAME}/ddev/master/scripts/macos_ddev_nfs_setup.sh"${RESET}\n" && exit 110)
 
 cd $TMPDIR
 $SHACMD -c "$SHAFILE"
```
rfay, line 156:
```suggestion
if [ ${VERSION} != "HEAD" ]; then
  $SHACMD -c "$SHAFILE"
fi
```

scripts/install_ddev.sh
```diff
@@ -117,29 +124,38 @@ else
...
 tar -xzf $TARBALL
```
rfay, line 157:
```suggestion
if [ ${VERSION} != "HEAD" ]; then
  tar -xzf ${TARBALL}
else 
  unzip ${TARBALL}
 fi
```

rfay (changes requested):
I think the suggestions here will get you started, but it still has a ways to go.

scripts/install_ddev.sh
uberhacker: HEAD should also work with the `.tar.gz` extension if downloaded that way.

scripts/install_ddev.sh
```diff
@@ -117,29 +124,38 @@ else
...
 
-
 if ! docker --version >/dev/null 2>&1; then
     printf "${YELLOW}Docker is required for ddev. Please see https://ddev.readthedocs.io/en/stable/#docker-installation.${RESET}\n"
 fi
 
-TARBALL="$FILEBASE.$VERSION.tar.gz"
-SHAFILE="$TARBALL.sha256.txt"
+if [ "$VERSION" == 'HEAD' ]; then 
+  TARBALL="$FILEBASE.zip"
```
uberhacker, line 143: This could also be: `TARBALL="$FILEBASE.tar.gz"`

scripts/install_ddev.sh
```diff
@@ -98,12 +98,19 @@ VERSION=$LATEST_VERSION
...
+  RELEASE_BASE_URL="https://nightly.link/drud/ddev/workflows/master-build/master"
+else
+  RELEASE_BASE_URL="https://github.com/${GITHUB_USERNAME}/ddev/releases/download/$VERSION"
+fi
+
+if [ "$VERSION" != 'HEAD' ]; then
+  rv=$(semver_compare "${VERSION}" "v1.10.0")
+  if [[ ${rv} -lt 0 ]]; then
+    printf "${RED}Sorry, this installer does not support specifying versions of ddev prior to v1.10.0${RESET}\n"
+    exit 1
```
uberhacker, line 112: Why limit?

shaal: @rfay?

rfay: I can't remember, but I think the artifacts were in a different naming format before v1.10.

scripts/install_ddev.sh
rfay: If you look at the artifacts for what this is trying to get, you'll see only zip artifacts, see https://nightly.link/drud/ddev/workflows/master-build/master

Why that is I don't know/remember. It would be fine to improve the build process, but that may be out of scope?

rfay (changes requested):
This is making progress. It still fails installing the completions, so it should skip that I guess.

```
cp ddev_bash_completion.sh /opt/homebrew/etc/bash_completion.d/ddev
cp: ddev_bash_completion.sh: No such file or directory
```

rfay:
I'd like to go ahead and start over on this. The artifacts for the HEAD are quite different from the release artifacts, so it's very complicated to add support in this script. I'd like to start over with install_ddev_head.sh if that's OK.

PR #3607, opened 2022-02-17
Look for tableplus in setapp directory as well, fixes #3606

### LINKED ISSUE #3606: Recognise TablePlus from SetApp Subscriptions
I'm using TablePlus via a setapp.com Subscription. The TablePlus binary is located here /Applications/Setapp/TablePlus.app instead.

That causes the `ddev tableplus` command to be not available on my machine.

Perhaps it would be nice to add an additional check for this subdirectory.

Thanks
Rico

rfay:
## The Problem/Issue/Bug:

Some people have tableplus installed in /Applications/Setapp

* #3606

Look for it there as well

Use Tableplus on a machine that has it intalled in /Applications/Setapp

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3607"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
@@ -7,7 +7,7 @@
 ## Usage: tableplus
 ## Example: "ddev tableplus"
 ## OSTypes: darwin
-## HostBinaryExists: /Applications/TablePlus.app
+## HostBinaryExists: /Applications/TablePlus.app,/Applications/Setapp/TablePlus.app
 
 set -x
 query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/db"


rfay:
I don't think this can work because of the awkward way that Tableplus provides to give the query. 

```
query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/db"
open "$query" -a "/Applications/TablePlus.app/Contents/MacOS/TablePlus"
```

We have to reach completely inside the app.

luke3butler:
For anyone looking for a solution, this should solve the problem. 
It also doesn't have any of the side-effects that come with copy/pasting the entire application.

```sh
mkdir -p "/Applications/TablePlus.app/Contents/MacOS" && ln "/Applications/Setapp/TablePlus.app/Contents/MacOS/TablePlus" "/Applications/TablePlus.app/Contents/MacOS/TablePlus"
```

PR #3612, opened 2022-02-19
Allow overriding system nodejs version and add nvm support, fixes #2651

### LINKED ISSUE #2651: Select node/npm version in config.yaml
Quite often the scripts that we need to compile need an specific version of node, not being compatible with modern versions of node.js In order to continue compiling these assets inside the web container, it would be ideal to have the possibility to choose which version of node.js to have installed in the container, in a simple way as we can do to choose PHP or composer version.

To have a new option in config.yaml where to specify the major version of node installed in the container.

I am currently using the n command as an alternative, as it allows to install a different version of node. However a more global solution would be more elegant.

**Additional context**
It's common to have npm dependencies in modern projects, and some of them are not compatible with newer versions of node.js In order to being able of maintain old sites using modern versions of ddev, this feature is really needed.

rfay:
## The Problem/Issue/Bug:

* #2651 

Allow setting node version with `nodejs_version` in config.yaml or `ddev config --nodejs-version=14`

Supported versions are 12, 14, 16, 17. Default is current LTS, "16".

- [x] `ddev config --nodejs-version=<someversion>` and `ddev start`. 
- [x] `ddev start` && `ddev exec node --version`
- [x] Do anything else you should do with a node version

* Added node startup test to make sure we can do each supported version
* Added to configuration tests

## Related Issue Link(s):

#2651

## Release/Deployment notes:

- [ ] Explain this feature

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3612"></a>


rfay:
The reality is we should just make a separate script for installing HEAD. We're just making a mess of install_ddev.sh here.

pkg/ddevapp/config.go
```diff
@@ -416,6 +417,10 @@ func (app *DdevApp) ValidateConfig() error {
 		return fmt.Errorf("unsupported webserver type: %s, ddev (%s) only supports the following webserver types: %s", app.WebserverType, runtime.GOARCH, nodeps.GetValidWebserverTypes()).(invalidWebserverType)
 	}
 
+	if !nodeps.IsValidNodeVersion(app.NodeJSVersion) {
+		return fmt.Errorf("unsupported NodeJS version: '%s', ddev only supports the following node versions: %s", app.NodeJSVersion, nodeps.GetValidNodeVersions())
```
jonaseberle, line 421:
```suggestion
		return fmt.Errorf("unsupported NodeJS version: '%s', ddev only supports the following node versions: %s. You can use 'ddev nvm' at runtime to use more.", app.NodeJSVersion, nodeps.GetValidNodeVersions())
```

PR #3626, opened 2022-02-24
[docs] Explain relationship between composer project and docroot

iamsophiesk:
## The Problem/Issue/Bug:

I tried to spin up a brand new Drupal site using the instructions at https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-9-quickstart 

Years of using Acquia meant I set the DDEV config to use `docroot`: 

```
ddev config --project-type=drupal9 --docroot=docroot --create-docroot
```

Because the Composer create command automatically sets the Composer `[web-root]` to `web`, this caused conflicts in my install. DDEV looked for `docroot` while Drupal was in `web`. **This is expected behaviour** but it wasn't immediately apparent from the docs that this would be an issue.

Just adds a little bit of context to the docs page, telling people not to be dumb like I was :) 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3626"></a>

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -172,6 +172,8 @@ Quickstart instructions regarding database imports can be found under [Database 
 
 #### Drupal 9 Composer Setup Example
 
+NB: `ddev composer create "drupal/recommended-project"` will, by default, set the webroot to `web`. If you set a different value in the `ddev config` command, Drush won't be able to find the database, and the install commands will fail.
+
 ```bash
 mkdir my-drupal9-site
 cd my-drupal9-site


rfay (changes requested):
Thanks so much for taking the time to make this contribution, and your fixes for docs and other things are always welcome!

Since the problem you had here has nothing to do with Drupal and everything to do with using composer and configuring webservers, I don't think it belongs in this exact place. Maybe in belongs in the discussion of [ddev and composer](https://ddev.readthedocs.io/en/latest/users/developer-tools/#ddev-and-composer)? Anyway, please think back about your experience and consider where the docs would most likely help the former you :)

iamsophiesk:
> please think back about your experience and consider where the docs would most likely help the former you :) 

I needed to make a new Drupal 9 project and wanted to use DDEV to spin it up quickly, so I went to the bit of the docs that described how to set up a new D9 project with DDEV/Composer - hence changing the value here :)

I guess I would use the `ddev config...` commands in the following situations (bearing in mind I am a Drupal dev and haven't used DDEV for anything other than Drupal):
* Creating a brand new project with Composer: wouldn't be able to set `--docroot` to something other than the one that Drupal provides. 
* Creating a brand new project with Git: there is no `--docroot` flag set in those commands, it uses `--auto` instead.
* Adding DDEV to an existing Composer project: won't run the Composer create command, the `web-root` should already be set in `composer.json`, and `--docroot` will set the config correctly.
* Adding DDEV to an existing non-Composer project: you have bigger problems :D 

I'm not sure whether the statement is applicable more globally because I don't use DDEV for anything other than Drupal - if you think it should go somewhere else then I can move it, but if it's an issue that came up in this situation, then it should be mentioned here as well, IMO... either as a note like this PR, or as a link off to elsewhere ("go see the Composer caveats").

rfay:
These are all issues that you would have in any environment, and don't actually have to do with ddev. In every hosting environment there's a "docroot", and if you use composer. it usually has its own idea about the docroot. 

The reason that `--docroot` is required with the composer examples is that we're starting with a clean slate, so ddev can't auto-discover the docroot (as it would with a checked-out project). There's nothing there until the composer create-project runs. 

I think I understand why you were put off by this. I'll continue pondering it. But I don't know how to solve this right now in ddev docs, since the problem isn't a ddev issue. But I'll keep thinking about it. I'm going to close this for now, happy to continue the conversation here or elsewhere.

PR #3627, opened 2022-02-24
Suppress feedback from command failure, fixes #3518

### LINKED ISSUE #3518: Add option to not display result of `ddev exec`
I am working on adding [support for DDEV on GrumPHP (https://github.com/phpro/grumphp/pull/972/files)
At the moment `ddev exec` outputs the command that failed to execute, either in full or in json with `ddev exec -j`.

**Full Output of a failing commit**
image
**JSON Output of a failing commit**
image

I'd like  a `ddev exec` option to not output anything, GrumPHP will be red or green anyway and that's all the verbosity that is required.
Something like `ddev exec --no-output COMMAND`

Run `git` commands from inside the container, but it is not really an option because one would have to copy over ssh keys, git config (easy) and gpg keys (I don't know how to do this yet).

**Additional context**
Grum seems to be widely adopted. If we manage to get https://github.com/phpro/grumphp/pull/972 in a good state it would mean out-of-the-box support for DDEV. PR over there is working at the moment, though I'd 'love to improve it to run both inside and outside of ddev containers.

bserem:
## The Problem/Issue/Bug:
* #3518 

It does not print a message when a command fails, thus only display the original command response and not the one from ddev/docker

| Before | After |
| --- | --- |
| image | image |

Do `ddev exec "ls whatever"` with and without the changes

## Related Issue Link(s):
#3518 

## Release/Deployment notes:
Unaware if it affects things elsewhere.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3627"></a>

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -146,6 +146,7 @@ func Execute() {
 
 func init() {
 	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "json-output", "j", false, "If true, user-oriented output will be in JSON format.")
+	RootCmd.PersistentFlags().BoolVarP(&output.NoOutput, "no-output", "n", false, "If true, error messages output will be suppressed.")
 
 	output.LogSetUp()
 
--- a/pkg/output/output_setup.go
+++ b/pkg/output/output_setup.go
@@ -3,6 +3,7 @@ package output
 import (
 	log "github.com/sirupsen/logrus"
 	"os"
+	"io"
 )
 
 var (
@@ -14,6 +15,8 @@ var (
 	UserOutFormatter = new(TextFormatter)
 	// JSONOutput is a bool telling whether we're outputting in json. Set by command-line args.
 	JSONOutput = false
+	// NoOutput is a bool telling whether we want to supress any output. Set by command-line args.
+	NoOutput = false
 )
 
 // LogSetUp sets up UserOut and log loggers as needed by ddev
@@ -41,6 +44,10 @@ func LogSetUp() {
 		logLevel = log.DebugLevel
 	}
 	log.SetLevel(logLevel)
+
+	if NoOutput {
+		UserErr.Out = io.Discard
+	}
 }
 
 // ErrorWriter allows writing stderr


rfay (changes requested):
I don't think we ever discussed just *not* outputting the failure. We talked about a config element that would allow it to be suppressed. This will work for you but nobody else. I imagine you're planning to work toward that config element.

bserem:
I can give it a try when time allows it, but Go is not my cup of tea, so it will take some time

bserem:
Let's see if I got that right:

1. Normal command (unaffected by PR)
2. JSON Output (unaffected by PR)
3. **New** global command flag, `--no-output`, supressing output to StdErr
4. **Updated** Help message, with the new flag

image

rfay (commented):
If no-output is set... this will affect everything, right? I think your goal was just to affect `ddev exec`, right? This will affect hooks, output from loads of other things...

bserem:
Right... Well then, I honestly don't know if I can do it. I can move the flag under `exec`, but then it does not output an error, so no `exit` code is sent to bash, so a sequence of commands triggered by exec will run to the end (example: a failing `pre-commit` will continute to `commit-msg`).

rfay:
Are you committed to doing this with `ddev exec`? There are so many ways to do these things. You can `ddev ssh` as you already know. You can `ddev exec` a bash script that manages the output. There are so many different things you can do. 

In this case, I expected that you would change `ddev exec` to redirect the stderr output, or something like that. 

But still, it seems to me like you're killing the messenger instead of finding a way to manage the problem.

gilbertsoft:
I'm not sure if this a good option. To surpress warnings to upstream simply do something like this:

```
ddev exec 'ls randomdir || true'
```

Why should we surpress a real failure? Does not make much sense to me.

bserem:
@gilbertsoft I was not aware of `|| true`, and it seems none of us was (see #3518). I'll give it a go soon(ish).

The idea was not to supress the failure, only to hide the docker/ddev message, since it is a duplicate. I have added a lot of screenshots in the issue to make this as clear as possible.

This task apparently was more complex than I initially anticipated and like both you and Randy said I am very far from an ideal solution.
I am taking notes on all the suggestions and try them when I can, but I am starting to believe it doesn't make much sense. Maybe I should pause on this after all.

Thanks

gilbertsoft:
The ` || true` is a common bash workaround to surpress errors. Wondering nobody had suggested it before. I'm often using it for composer scripts and also in bash scripts where an error may happen but should not block the further execution.

I also had a look at your issue yesterday and for me it looks like you should fix the input which seems to be wrong and that's why you get a warning.

rfay:
This has been idle for 3 weeks, so I'm going to close it for now, happy to continue the conversation here or elsewhere, and we can reopen if you want to push more into it.

PR #3630, opened 2022-02-26
Don't allow snapshots for postgres:9, fixes #3583

### LINKED ISSUE #3583: `ddev snapshot restore` doesn't work with postgres:9
Although `postgres:9` is supported and seems to work, `ddev snapshot restore` fails. If anybody has experience with `pg_basebackup` and can help study this it would be appreciated.

Options include:
* Fixing this
* Removing snapshot support for postgres:9
* Removing all support for postgres:9

rfay:
## The Problem/Issue/Bug:

Snapshots don't work with postgres:9. Nobody has really stepped up to study it, and postgres:9 is probably a very limited case anyway, so just don't let people use snapshots there.

- [x] `ddev snapshot` and `ddev snapshot restore` on a project with database=`postgres:9`

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3630"></a>

### DIFF
--- a/cmd/ddev/cmd/snapshot.go
+++ b/cmd/ddev/cmd/snapshot.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"github.com/drud/ddev/pkg/ddevapp"
+	"github.com/drud/ddev/pkg/nodeps"
 	"github.com/drud/ddev/pkg/util"
 	"github.com/spf13/cobra"
 	"strings"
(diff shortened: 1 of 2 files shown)


rfay:
No, your problem would be that you're using `omit_containers: [db]`. Perhaps you're using the old, old way of using postgres, where it's not configured via DDEV but via a separate add-on service? You probably have a `.ddev/docker-compose.postgres.yaml` and probably have `omit_containers: [db]`.

rfay:
Yes, but your snapshot has no data unless you exported from the old system and imported into the new, you'll have to migrate your data from the old db to the new one.

rfay:
Oh, and until you remove the `docker-compose.postgres.yaml` and change your db to postgres:11, you won't be using the ddev-built-in facility. I'm sure you already know that.

alesrebec:
The snapshot was made on the old system and it appears it has some data in but I also have dump done with `drush sql:dump` to play it safe.

Yes, I'm aware that all the old references of postgres need to be removed to use the built-in one. Thanks.

PR #3631, opened 2022-02-26
Update the docs on the use of DDEV with Compose V2 support in PHPStorm

nico-loeber:
## The Problem/Issue/Bug:
The documentation currently contains information about a bug in PHPStorm that does not allow the use of Docker Compose V2 and provides a workaround by downgrading DDEV to use V1. This is no longer necessary as PHPStorm can use Compose V2 if it is available on the system, alternatively the bundled Docker Compose binary can be used.

The documentation section has been updated to recommend using your system's Docker Compose binary in V2 mode or using the bundled binary.

None

None

## Related Issue Link(s):
https://youtrack.jetbrains.com/issue/WI-61205

## Release/Deployment notes:
None

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3631"></a>

### DIFF
--- a/docs/users/topics/phpstorm.md
+++ b/docs/users/topics/phpstorm.md
@@ -10,7 +10,7 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 - PhpStorm 2021.2 or higher
 - You may need to add the `PHP Docker` plugin.
-- docker-compose v1. Due to a [bug or two in PhpStorm](https://youtrack.jetbrains.com/issue/WI-61205), it can't properly interpret docker-compose v2 output as of Sept 2021.  If this is still a problem, you can `ddev config global --required-docker-compose-version=v1.29.2` to globally require docker-compose v1.
+- As of version v1.18.2, DDEV is bundled with its own docker-compose binary, using Compose V2 by default. Enable _Use Compose V2_ in PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools_, in case you have an up to date docker-compose binary installed (Check with `docker-compose version` > 2.0). Otherwise, you can also set the bundled docker-compose binary as _Docker Compose executable_ which you can be found in `~/.ddev/bin`. In this case, you do not need to select the _Use Compose V2_ Checkbox.
 - Any OS platform, but if you're using Windows PhpStorm with WSL2 the path mappings are slightly more complex. WSL2 instructions are provided where necessary.
 
 #### Setup Technique
--- a/docs/users/topics/phpstorm.md
+++ b/docs/users/topics/phpstorm.md
@@ -10,7 +10,9 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 - PhpStorm 2021.3 or higher
 - You may need to add the `PHP Docker` plugin.
-DDEV uses docker-compose v2 by default, so enable the **Use Compose V2** checkbox  in `PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools` and set the "Docker Compose executable" to ddev's private docker-compose binary at `~/.ddev/bin/docker-compose` (Home directory, `.ddev/bin/docker-compose.). However, if using PhpStorm on Windows side with WSL2 you'll need to use a docker-compose v2 binary as provided by Docker Desktop, normally `C:\Program Files\Docker\Docker\resources\bin\docker-compose`.
+- You need to enable Compose V2 Support in PHPStorm. This can be done in `Settings/Preferences | Build, Execution, Deployment | Docker | Tools` by either
+   1. set the bundled docker-compose binary as **Docker Compose executable** which can be found in `~/.ddev/bin`.
+   2. (Suggested for WSL) enabling **Use Compose V2** in PHPStorm and your Docker installation.
 - Any OS platform, but if you're using Windows PhpStorm with WSL2 the path mappings are slightly more complex. WSL2 instructions are provided where necessary.
 
 #### Setup Technique
--- a/docs/users/topics/phpstorm.md
+++ b/docs/users/topics/phpstorm.md
@@ -8,7 +8,7 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 #### Requirements
 
-- PhpStorm 2021.2 or higher
+- PhpStorm 2021.3 or higher
 - You may need to add the `PHP Docker` plugin.
 - As of version v1.18.2, DDEV is bundled with its own docker-compose binary, using Compose V2 by default. Enable _Use Compose V2_ in PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools_, in case you have an up to date docker-compose binary installed (Check with `docker-compose version` > 2.0). Otherwise, you can also set the bundled docker-compose binary as _Docker Compose executable_ which you can be found in `~/.ddev/bin`. In this case, you do not need to select the _Use Compose V2_ Checkbox.
 - Any OS platform, but if you're using Windows PhpStorm with WSL2 the path mappings are slightly more complex. WSL2 instructions are provided where necessary.
--- a/docs/users/topics/phpstorm.md
+++ b/docs/users/topics/phpstorm.md
@@ -10,7 +10,7 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 - PhpStorm 2021.3 or higher
 - You may need to add the `PHP Docker` plugin.
-- As of version v1.18.2, DDEV is bundled with its own docker-compose binary, using Compose V2 by default. Enable _Use Compose V2_ in PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools_, in case you have an up to date docker-compose binary installed (Check with `docker-compose version` > 2.0). Otherwise, you can also set the bundled docker-compose binary as _Docker Compose executable_ which you can be found in `~/.ddev/bin`. In this case, you do not need to select the _Use Compose V2_ Checkbox.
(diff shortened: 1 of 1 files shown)


docs/users/topics/phpstorm.md
```diff
@@ -10,7 +10,7 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 - PhpStorm 2021.2 or higher
 - You may need to add the `PHP Docker` plugin.
-- docker-compose v1. Due to a [bug or two in PhpStorm](https://youtrack.jetbrains.com/issue/WI-61205), it can't properly interpret docker-compose v2 output as of Sept 2021.  If this is still a problem, you can `ddev config global --required-docker-compose-version=v1.29.2` to globally require docker-compose v1.
+- As of version v1.18.2, DDEV is bundled with its own docker-compose binary, using Compose V2 by default. Enable _Use Compose V2_ in PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools_, in case you have an up to date docker-compose binary installed (Check with `docker-compose version` > 2.0). Otherwise, you can also set the bundled docker-compose binary as _Docker Compose executable_ which you can be found in `~/.ddev/bin`. In this case, you do not need to select the _Use Compose V2_ Checkbox.
```
rfay, line 13: Thanks so much for chasing this and this contribution!

I would say we don't need to make it this complex. That's a lot of information.

```suggestion
- DDEV uses docker-compose v2 by default, so enable the **Use Compose V2** checkbox  in `PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools`.
```

docs/users/topics/phpstorm.md
```diff
@@ -10,7 +10,7 @@ This is based on the wonderful [article](https://susi.dev/fully-integrate-ddev-a
 
 - PhpStorm 2021.2 or higher
```
rfay, line 11:
```suggestion
- PhpStorm 2021.3 or higher
```

rfay (changes requested):
Awesome, thanks. Please check my suggestions, commit if you like them.

docs/users/topics/phpstorm.md
nico-loeber: I am not sure if we should ommit all of the information. You are right, that it might be to detailed. But Compose V2 is currently only avaiable by default on Systems with Docker Desktop (Windows, Mac). I think Linux Users might not have Compose V2 support of their system binary, so we should give information on how to use it anyway, or at least how to find out if they have.
What do you think?

rfay: No, all ddev users on all platforms have docker-compose v2 unless they override globally.

nico-loeber: Eanbling the Use Compose V2 Checkbox will let PHPStorm use the docker-compose executable of your system to parse the DDEV generated configuration. Selecting the Checkbox will not guarantee a V2 capable binary will be used. As far as I understood it will just change its internal handling of the binary and validation.
If you like I can try to confirm this on a Debian distribution.

rfay: I recommend telling people to set the docker-compose binary to ~/.ddev/bin/docker-compose I guess. That's a consistent solution. People who need to do something more exotic can sort it out, but since PhpStorm is now working with docker-compose v2, they shouldn't need anything more exotic.

nico-loeber: This is a too restricted view in my opinion. I work on many projects every day and only for a few I use DDEV. Setting the docker-compose binary is a global setting affecting all Projects. This might result in unexpected behavior for any non ddev project. It is a valid solution for people using docker-compose only in DDEV context, that's why I listed it as an option.
I would recomment using the system docker-compose binary, as it is in sync with your installed docker version (As V2 is a docker plugin) and it is the same in both your OS and WSL (Thanks to the WSL2 Backend of Docker Desktop). Also it is automatically updated by Docker Desktop or the package manager of your choice. 
Another less idealistic reason for my recommendation is, that I work mostly in WSL. It is currently not possible to select a Linux executable (non .exe) in PHPStorm as docker-compose executable. So you would need to install DDEV in WSL, as this is currently the recommended way, and in windows to select its binary here. 
This makes in total 3 docker-compose binaries you have which might all have different versions. This will cause trouble for sure.

rfay: Sorry, these are the ddev instructions, so let's use the ddev docker-compose. People like you will know what to do, and people who have need for another docker-compose will know what to do. 

It's always complicated to understand which PhpStorm instructions are global for all projects and which are per-project, so that may be confusing, but again, we're doing the ddev instructions here.

nico-loeber: Okay, so let's summarize:
- The original version works in all environments but might be too complicated, I admit that.
- The version that is currently suggested by you won't work for Linux users in most cases (No V2 binary)
- Your last suggestion was using the bundled version, this will not work for Windows with WSL (Can't select Linux binaries)

We could close this pull request, as downgrading DDEV should work on all environments. You should have permission to push to this pull request, so you can choose whatever version you like most.

rfay: Thanks so much for your work on this and for pioneering all this and testing it. 

If you're using PhpStorm on the Windows side (and Docker Desktop) with WSL2, yes, you can't point it directly at ddev's binary.  However, the bundled Docker Desktop version will work, and they all default to v2. Thanks for pointing that out. WSL2 always brings out the complexity. Or at least WSL2 with Docker Desktop. 

How about this?

```suggestion
DDEV uses docker-compose v2 by default, so enable the **Use Compose V2** checkbox  in `PHPStorm _Settings/Preferences | Build, Execution, Deployment | Docker | Tools` and set the "Docker Compose executable" to ddev's private docker-compose binary at `~/.ddev/bin/docker-compose` (Home directory, `.ddev/bin/docker-compose.). However, if using PhpStorm on Windows side with WSL2 you'll need to use a docker-compose v2 binary as provided by Docker Desktop, normally `C:\Program Files\Docker\Docker\resources\bin\docker-compose`.
```

nico-loeber: This one seems to cover all use cases that come to my mind. I accepted your suggestion.

rfay:
No rush... just let me know when you're happy with it.

blankse:
@nico-loeber I test it with docker composer V2. It still shows the wrong PHP version (I'm on version 8.0):
image

And the PHP Stan Inspector doesn't work correct anymore:
image
With the old doks the PHPStan inspector show the errors on code lines.

Before I checked the docker compose V2 checkbox:
image

After:
image

Does the PHPStan Inspector work for you with docker compose V2?

I'm working on windows 11 with PhpStorm/ddev on WSL2. I tested it with newest PhpStorm stable and EAP version.

rfay:
@blankse thanks for chiming in. The bug where PhpStorm uses the wrong method to poll the php version is still not fixed, see https://youtrack.jetbrains.com/issue/WI-62463

This is #9 in the setup doc, https://ddev.readthedocs.io/en/latest/users/topics/phpstorm/#setup-technique
> 9. In the PHP Interpreter path, you can just put php if you're using the default PHP version (currently 7.4). Due to a [PhpStorm bug](https://youtrack.jetbrains.com/issue/WI-62463) you'll want to put the full name of the binary, like php8.0 if you're not using the default version.

They're making progress, at least they're supporting docker-compose v2 now.

blankse:
@rfay Yes I know I can put "php8.0" in the input field to force the right version. But the PHP Stan interpretor doesn't work with docker compose V2. There seems to be another bug. So I think we should hold the docu with docker compose V1 for this usecase.

nico-loeber:
Hi @blankse thanks for testing!
I prefer Psalm and didn't notice any error yet. I will take a look at it. Till then I have marked this pull request as a draft.
I wonder how using Compose v2 could be related to this phpstan issue. What are the steps to reproduce?

@rfay as we discussed yesterday I just pushed a new variant describing both possible ways of integration, as both ways should work for you. Using the internal binary is stated first as you prefer this solution. I tried to be brief.

rfay:
@blankse again thanks so much for reviewing and contributing on this.

Could you say exactly how the PHP Stan interpreter doesn't work with docker compose V2? What are the issues? I wouldn't think that PHP Stan would know anything about docker-compose. 

The class of bugs associated with PhpStorm may not go away, it's pretty complex. You may want to write up a separate approach to PHP Stan and we can link to it?

blankse:
@nico-loeber Try steps to reproduce:

1. Create an empty project
2. `ddev config` with default values
3. `ddev composer require phpstan/phpstan`
4. Create CLI interpreter
5. Close and reopen Settings (Interpreter seems not available instantly)
6. Setup PHPStan "PHP - Quality Tools - PHPStan"
Select created ddev web CLI interpreter
Select phpstan path: "vendor/bin/phpstan"
7. Activate PhpStan inspection "Editor - Inspections - PHP - Quality tools - PHPStan validation"
8. Create a PHP file with errors, for example index.php
```
<?php

if (false) {
    Calulator::test();
}
```

nico-loeber:
@blankse, could you please also provide the output of `docker-compose -v` and wich OS you are using, in case of Windows, do you use WSL? Thanks!
Please note that it might take some days, till I find time to trace the root of this issue.

blankse:
@nico-loeber 

> I'm working on windows 11 with PhpStorm/ddev on WSL2. I tested it with newest PhpStorm stable and EAP version.

I added the ddev buildin docker-compose to PhpStorm, so it should these versions:
V1
```
ddev config global --required-docker-compose-version=v1.29.2
ddev restart
~/.ddev/bin/docker-compose -v
docker-compose version 1.29.2, build 5becea4c
```
V2
```
ddev config global --required-docker-compose-version=""
ddev restart
~/.ddev/bin/docker-compose -v
Docker Compose version v2.2.2
```
If you want to know the normal docker-compose:
V1 (deactivate V2 Checkbox in Docker Desktop)
```
docker-compose -v
docker-compose version 1.29.2, build 5becea4c
```
V2 (activate V2 Checkbox in Docker Desktop)
```
docker-compose -v
Docker Compose version v2.2.3
```

rfay (approved):
I think this is as good as we're going to get, thanks for much for the effort on this. I'm sure we'll have to continue working on this as time goes on, and maybe you'll get it automated for us. 

@blankse we're going to go with this for now, but there will be more edits in the future, and your comments are very welcome. Thanks for testing and continuing to test. This has been pretty hard with PhpStorm.  And if you want to put up a gist or something specific to phpstan, just to a docs PR to link to it in this section.

PR #3646, opened 2022-03-01
The absolute available space value was wrong for checking docker disk space

rfay:
## The Problem/Issue/Bug:

I noticed in a casual test on Colima that the value provided for absolute disk space was calculated wrong.

Try to get it right

Try testing with various values and make sure it works

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3646"></a>


rfay:
I tested this by fiddling with the minimum space required. I hope 5GB is the right value there. It will work for all but the biggest import-db actions.

PR #3654, opened 2022-03-03
Make composer root directory configurable, fixes #1699

### LINKED ISSUE #1699: ddev could have a config.yaml item to determine its expected composer.json location

gilbertsoft:
## The Problem/Issue/Bug:

* #1699 

Sometimes the `composer.json` is not located in the project root and for each composer command the `working-dir` option must be provided.

This patch introduces a new config option to change the composer root to a sub folder. Composer can now be used without the `working-dir` option.

### config options

- Configure the composer root directory using the new option
- Reset the configuration to the default using the new option

### composer command

Run composer commands and check if the command is executed in the correct directory.

### composer-create command

Run composer-create command and check if the project is created the correct directory.

- the existing composer tests are extended with the new configuration
- the config tests are extended with the new options

## Related Issue Link(s):

Resolves #1699

## Release/Deployment notes:

Nothing special is needed.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3654"></a>


pkg/ddevapp/templates.go
```diff
@@ -80,6 +80,11 @@ const ConfigInstructions = `
 # These values specify the destination directory for ddev ssh and the
 # directory in which commands passed into ddev exec are run.
 
+# composer_root_dir: app
+# would set the composer root directory to sub directory "app". All composer
+# related commands will then be executed in this directory instead of project
+# root directory. The default is "/var/www/html".
```
gilbertsoft, line 86: default like this is the wrong term here, will change...

jonaseberle:
Just a note: It configures a bit differently from the `working_dir` (https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/#ddevconfigyaml-options)

This
```
composer_root_dir: /var/www/html/src
# for comparison:
working_dir:
  web: /var/www/html/src
```
would not work (tested with `ddev composer show`). It currently needs to be relative like `composer_root_dir: src` or `./src` or `./src/`. I would be nice to allow for relative and absolute paths.

The `ddev config composer-root-dir ...` and `ddev config composer-root-dir-default` work fine for me.

Following [this](https://ddev.readthedocs.io/en/stable/users/cli-usage/#typo3-quickstart) with `ddev config --project-type=typo3 --docroot=some/public --create-docroot --composer-root-dir=./some/` leads to a problem, though:

The `some/public` directory existed but has been removed in this command:
```
$ ddev composer create "typo3/cms-base-distribution:^11"

Warning: MOST EXISTING CONTENT in the project root (/home/jonas/tmp/my-typo3-site) will be deleted by the composer create-project operation. .git and .ddev will be preserved.
Would you like to continue? [Y/n] (yes): yes
Removing any existing files in project root
Executing composer command: [composer create-project typo3/cms-base-distribution:^11 /tmp/wSURnj]

Creating a "typo3/cms-base-distribution:^11" project at "/tmp/wSURnj"
Installing typo3/cms-base-distribution (v11.5.1)
  - Downloading typo3/cms-base-distribution (v11.5.1)
  - Installing typo3/cms-base-distribution (v11.5.1): Extracting archive
Created project in /tmp/wSURnj
Loading composer repositories with package information
Updating dependencies
... # Downloading, extracting and installing of packages
Generating autoload files
> typo3cms install:fixfolderstructure
...
Moving installation to composer root
Using '/var/www/html/some' as composer root directory
Using '/var/www/html/some' as composer root directory
Failed to create project: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-my-typo3-site docker-compose -f /home/jonas/tmp/my-typo3-site/.ddev/.ddev-docker-compose-full.yaml exec -w /var/www/html/some -T web bash -c set -eu && ( rsync -a "/tmp/wSURnj/" "/var/www/html/some/")', action='[exec -w /var/www/html/some -T web bash -c set -eu && ( rsync -a "/tmp/wSURnj/" "/var/www/html/some/")]', err='exit status 126', stdout='OCI runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/var/www/html/some") set in config.json failed: no such file or directory: unknown
', stderr=''
```
Is the generating in `/tmp`, then moving it somewhere else really still needed?

gilbertsoft:
Thanks for your feedback and testing!

> Just a note: It configures a bit differently from the `working_dir` (https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/#ddevconfigyaml-options)

Yes, it's currently relative to `/var/www/html`, the default for `working_dir`. I'm absolutely open to change this.

Also thought about to make it always relative to `working_dir`, would also be a good option. I did not because this would be breaking.

> Following [this](https://ddev.readthedocs.io/en/stable/users/cli-usage/#typo3-quickstart) with `ddev config --project-type=typo3 --docroot=some/public --create-docroot --composer-root-dir=./some/` leads to a problem, though:
> 
> The `some/public` directory existed but has been removed in this command:

Well, that's intended and a warning is shown. This behavior has not changed by the patch.


> ```
> $ ddev composer create "typo3/cms-base-distribution:^11"
> 
> Warning: MOST EXISTING CONTENT in the project root (/home/jonas/tmp/my-typo3-site) will be deleted by the composer create-project operation. .git and .ddev will be preserved.
> Would you like to continue? [Y/n] (yes): yes
> Removing any existing files in project root

Looks these messages should be optimized....

> Is the generating in `/tmp`, then moving it somewhere else really still needed?

 Yes, that's because git clone does not work in a non empty directory and I do not like a different behavior, if `composer_root_dir` is set.


@rfay what would you suggest, making this path also absolute? I'm fine with it.

rfay:
`composer create-project` absolutely will not work in a dir that has contents, so has to be done somewhere else; current code rsyncs it into place then. That's what the creation in /tmp and then the move is about. The reason for using /tmp is it's fast there no matter what the setup, docker mount, NFS, or mutagen. The copy into place is sometimes slow though. All this makes me hate `composer create-project` and wonder why it's even useful at all.

I'd be interested in changing all the quickstarts to `ddev composer create-project --no-install` followed by `ddev composer install` which loses most of that nonsense.

rfay:
IMO composer_root_dir should be relative to /var/www/html - I think that's what you're asking.

I don't recommend using working_dir for something like this as it may mean other things to people. It's mostly the default dir for `ddev exec`, `ddev ssh`. I don't think it's widely used.

gilbertsoft:
> I'd be interested in changing all the quickstarts to `ddev composer create-project --no-install` followed by `ddev composer install` which loses most of that nonsense.

This will not solve the cloning issue. I don't see a solution if the composer root is equal to the project root. But with this change I could make it simpler by directly using the destination. As this is more an edge case, this would still not solve the problem for the most cases though...

rfay:
BTW I probably won't review this until I get v1.19.0 out. Thanks for the good work here!

gilbertsoft:
> BTW I probably won't review this until I get v1.19.0 out. Thanks for the good work here!

Yeah, was expected, was a bit late to the party :)


@rfay can we agree using an absolute path here?

docs/users/extend/config_yaml.md
```diff
@@ -26,10 +26,11 @@ the .ddev/config.yaml is the primary configuration for the project.
...
 | working_dir                      | explicitly specify the working directory used by `ddev exec` and `ddev ssh`                | `working_dir: { web:  "/var/www", db: "/etc" }` would set the working directories for the web and db containers.                                                                                                                                                                                                                                                                                                                               |
+| composer_root_dir                | explicitly specify the composer root directory used by `ddev composer`                     | `composer_root_dir: custom-dir` would set the composer root directory to the `custom-dir` sub folder. All composer related commands will run in this directory.                                                                                                                                                                                                                                                                                |
```
jonaseberle, line 29:
```suggestion
| composer_root_dir                | explicitly specify the composer root directory used by `ddev composer`                     | `composer_root_dir: /var/www/html/custom-dir` would set the composer root directory to the `custom-dir` sub folder. All composer related commands will run in this directory.                                                                                                                                                                                                                                                                                |
```

rfay: This is OK (mentioning /var/www/html) but so many people don't understand the difference between in-container and on-host paths...

gilbertsoft: That's currently as the path is always calculated relative to `/var/www/html`. But as said I think we should streamline this configuration and make it absolute too.

gilbertsoft:
> Currently absolute has been shown in the docs for `working_dir` so I agree with using this here, too. My point was just about a bit of commonality between the two options ;)

Absolutely agree. And this will also make it more flexible. Going to change it...

gilbertsoft:
Changed now, please check the description here for the new behavior https://github.com/drud/ddev/pull/3654/files#diff-5d46f5073025478314cefbf1d169ed7014159e6f7c20235703b85bb50343c672R83-R90

jonaseberle (approved):
Thanks Simon! It works fine for me.
Even `ddev composer create ...` is fixed now.

The test failure in `buildkite/ddev-macos-use-nfsmount` seems unrelated:
```
error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.41/containers/429b59675e6b12e8725433a91763e053e8749948dfab926e7d8f454c4437f17a/start": EOF'
```

gilbertsoft:
> The test failure in `buildkite/ddev-macos-use-nfsmount` seems unrelated:
> 
> ```
> error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.41/containers/429b59675e6b12e8725433a91763e053e8749948dfab926e7d8f454c4437f17a/start": EOF'
> ```

Thanks for testing! Yes, already talked to Randy, this are common hiccups which happens regularly. Not related to my changes.

josefglatz:
Thanks for adding this feature so fast! I've tested it successfully. Based on the manual testing instructions @gilbertsoft

rfay (approved):
Rebased and added docs, thanks for this excellent contribution. I'm tempted to go ahead with it for v1.19.0.

rfay:
Minor rebase was required after https://github.com/drud/ddev/pull/3677, so letting it go through another test run.

PR #3663, opened 2022-03-05
Update typo "Pull path" vs "Full path"

scotthooker:
Pull path is misleading. I think it should be full path?

## The Problem/Issue/Bug:

Typo

Fixes typo

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3663"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -442,7 +442,7 @@ func (app *DdevApp) ImportDB(imPath string, extPath string, progress bool, noDro
 			extPathPrompt = true
 		}
 		output.UserOut.Println("Provide the path to the database you want to import.")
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for taking the time to make this contribution! There are actually 2 instances of this, also one for import-files. I guess nobody uses the prompted usage, everybody uses the flags. It would be good to make it clearer in both cases. I don't think it was a typo, as it went in four years ago and has been that way ever since :)

Did you actually use the prompted usage? How could it be better in general? We can change more than one letter if needed :)

scotthooker:
Oh god I didn't mean to open a PR 🤦 this isn't a contribution... It just confused me for a while as to where it was "pulling from".

rfay:
Closing this now per your request, but let's keep thinking about it.

PR #3671, opened 2022-03-06
Make tests working from foreign repositories

gilbertsoft:
## The Problem/Issue/Bug:

Access to secrets is denied by default for PRs from foreign repositories and all tests using these will fail therefor.

The solution is to use the `pull_request_target` event instead of the simple `pull_request` event. This changes the behavior of the workflows for security reasons, the workflow is not loaded from the PR but always the base of it. This makes sure nobody can get access to the token and abuse it.

Merge this PR and create a new PR from a foreign repository to verifiy if the tests are working as expected now.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3671"></a>

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -1,6 +1,6 @@
 name: Colima tests
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - "go.*"
       - "pkg/**"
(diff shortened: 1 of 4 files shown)


rfay:
Thanks!

PR #3683, opened 2022-03-10
 Add `ddev clean` command, fixes #2332

### LINKED ISSUE #2332: Add a new `ddev clean` command to remove unused or replaceable items
While on a quest to clear up some hard drive space on my local machine, I discovered that the `~/.ddev/` directory was hiding a large amount of database backups. After confirming with @rfay on Slack, I set about to delete them - when I was through I reclaimed over 20GB of hard drive space (granted, I've been using DDEV for a few years, and have a significant number of projects).  

I think it would be a good idea to add a new command to automate the process. I suggested `ddev clean` above, but I'm not tied to that command name. Maybe `ddev cache-clear` or `ddev reclaim` or `ddev gimme-my-hd-space-back`.

Ideally, when run, the command should remove the following:

- `~/.ddev/<project-name>/mysql.saved/` - possibly from when the DB was moved from bind-mounted to a Docker volume.  
- `~/.ddev/<project-name>/mysql_migrated.back/` - possibly from when the DB was moved from bind-mounted to a Docker volume(?)    
- `~/.ddev/pantheon/<project-name>/` - database backups from each and every `ddev pull`...
- Any other location where .ddev is squirreling away DB backups.  

The command should also report how much disk space was reclaimed.

I would also suggest a couple of options on the command:

- `--dry-run` - does everything except actually deleting any files.  
- `--keep-recent` - keeps any files less than 1 year (or some reasonable time frame) old.

mikebarkas:
## The Problem/Issue/Bug:
Provide a new clean command

This creates a new command

Run `ddev clean projectname` and verify the snapshots have been removed

No tests have been created. I would like to create the tests.

## Related Issue Link(s):
#2332 

## Release/Deployment notes:

No

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3683"></a>

### DIFF
--- a/cmd/ddev/cmd/clean.go
+++ b/cmd/ddev/cmd/clean.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/globalconfig"
@@ -32,7 +33,7 @@ Additional commands that can help clean up resources:
 		// Make sure the user provides a project or flag
 		cleanAll, _ := cmd.Flags().GetBool("all")
 		if len(args) == 0 && !cleanAll {
-			util.Failed("No project provided. See ddev clean --help for options")
+			util.Failed("No project provided. See ddev clean --help for usage")
 		}
 
 		util.Success("Powering off ddev to avoid conflicts")
@@ -64,8 +65,8 @@ Additional commands that can help clean up resources:
 			os.Exit(1)
 		}
 
-		confirm := util.Prompt("Are you sure you want to continue? Y/N", "N")
-		if confirm == "Y" {
+		confirm := util.Prompt("Are you sure you want to continue? y/n", "n")
+		if strings.ToLower(confirm) == "y" {
 			globalDdevDir := globalconfig.GetGlobalDdevDir()
 			_ = os.RemoveAll(filepath.Join(globalDdevDir, "testcache"))
 			_ = os.RemoveAll(filepath.Join(globalDdevDir, "bin"))
@@ -73,12 +74,18 @@ Additional commands that can help clean up resources:
 			output.UserOut.Print("Deleting snapshots and downloads for selected projects...")
 			for _, project := range projects {
 				// Delete snapshots and downloads for each project
-				_ = os.RemoveAll(project.GetConfigPath(".downloads"))
-				_ = os.RemoveAll(project.GetConfigPath("db_snapshots"))
+				err = os.RemoveAll(project.GetConfigPath(".downloads"))
+				if err != nil {
+					util.Warning("There was an error removing .downloads for project %v", project)
+				}
+				err = os.RemoveAll(project.GetConfigPath("db_snapshots"))
+				if err != nil {
+					util.Warning("There was an error removing db_snapshots for project %v", project)
+				}
 			}
 
 			output.UserOut.Print("Deleting Docker images that ddev created...")
-			err = deleteDdevImages(true)
+			err = deleteDdevImages(cleanAll)
 			if err != nil {
 				util.Failed("Failed to delete image tag", err)
 			}
--- a/cmd/ddev/cmd/clean.go
+++ b/cmd/ddev/cmd/clean.go
@@ -1,13 +1,12 @@
 package cmd
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/drud/ddev/pkg/dockerutil"
 	"github.com/drud/ddev/pkg/globalconfig"
+	"github.com/drud/ddev/pkg/output"
 	"github.com/drud/ddev/pkg/util"
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/spf13/cobra"
@@ -21,36 +20,27 @@ var CleanCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 
 		util.Success("Powering off ddev to avoid conflicts")
-		powerOff()
+		// ddevapp.PowerOff()
 
-		util.Success("Gathering ddev items to be cleaned")
-		// Display a list of ddev projects that will be removed
-		projectList := globalconfig.GetGlobalProjectList()
-		fmt.Println("Snapshots and downloads from the following project will be removed")
-		for appRoot := range projectList {
-			fmt.Printf("\t Project %s will be cleaned \n", appRoot)
+		util.Warning("Warning - Snapshots for the following project(s) will be permanently deleted")
+
+		cleanAll, _ := cmd.Flags().GetBool("all")
+		projects, err := getRequestedProjects(args, cleanAll)
+		if err != nil {
+			util.Failed("Failed to get project(s): %v", err)
 		}
 
-		// Create a list of ddev images
-		client := dockerutil.GetDockerClient()
-		imagesToRemove, _ := client.ListImages(docker.ListImagesOptions{
-			All: true,
-		})
-
-		// Provide space between sections for readability
-		fmt.Println(" ")
-		fmt.Println("The following ddev images will be removed")
-		for _, image := range imagesToRemove {
-			for _, tag := range image.RepoTags {
-				if strings.HasPrefix(tag, "drud/ddev-") {
-					fmt.Printf("\t %s \n", tag)
-				}
+		// Show the user how many snapshots per project that will be deleted
+		for _, project := range projects {
+			snapshots, err := project.ListSnapshots()
+			if err != nil {
+				util.Failed("Failed to get project %s snapshots: %v", project.Name, err)
 			}
-		}
-		// Show the user a list of everything and ask to confirm
-		if !util.Confirm("Confirm removal of the listed items?") {
-			util.Warning("Program terminated")
-			os.Exit(1)
+			if len(snapshots) > 0 {
+				output.UserOut.Printf("%v Snapshots for project %s will be deleted", len(snapshots), project.Name)
+			} else {
+				output.UserOut.Printf("No snapshots found for project %s", project.Name)
+			}
 		}
 
 		dryRun, _ := cmd.Flags().GetBool("dry-run")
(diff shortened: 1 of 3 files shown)


cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+var CleanCmd = &cobra.Command{
+	Use:   "clean [projectname ...]",
+	Short: "Removes items ddev has created",
+	Long: `Stops all running projects and then removes downloads and snapshots
+for the selected projects. Then clean will remove ddev images.
+
+Warning - This command will permanently delete your snapshots and you will lose
+your database data in these snapshots.
```
rfay, line 23:
```suggestion
Warning - This command will permanently delete your snapshots for the named project[s].
```

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+
+		dryRun, _ := cmd.Flags().GetBool("dry-run")
+		if dryRun {
+			util.Warning("Dry run terminated without removing items")
+			os.Exit(1)
+		}
+
+		if util.Confirm("Confirm removal of the listed items?") {
```
rfay, line 65: How about using util.Prompt() here instead, with default 'N' ?

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+			ddevDir := globalconfig.GetGlobalDdevDir()
```
rfay, line 66: Let's use a variable that explains what it is here, maybe globalDdevDir. I always hate when I confuse the two dirs.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+			_ = os.RemoveAll(filepath.Join(ddevDir, "testcache"))
+			_ = os.RemoveAll(filepath.Join(ddevDir, "bin"))
+			_ = os.RemoveAll(filepath.Join(ddevDir, "downloads"))
```
rfay, line 69: I don't think the global ddev dir has ever had a downloads folder.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+
+			output.UserOut.Print("Deleting snapshots and downloads for selected projects...")
+			for _, project := range projects {
+				// Delete snapshots and downloads for each project
+				_ = os.RemoveAll(filepath.Join(project.AppRoot, ".ddev", "downloads"))
```
rfay, line 74: Easier usage for these is app.GetConfigPath("downloads")

But I think the thing we want to delete is ".downloads", with the dot on the front.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+				_ = os.RemoveAll(filepath.Join(project.AppRoot, ".ddev", "db_snapshots"))
+			}
+
+			output.UserOut.Print("Deleting images that ddev created...")
```
rfay, line 78:
```suggestion
			output.UserOut.Print("Deleting docker images that ddev created...")
```

People are so confused about images/containers/docker that we can help them a little here.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+			client := dockerutil.GetDockerClient()
+			images, err := client.ListImages(docker.ListImagesOptions{
```
rfay, line 80: Let's promote the code you wrote for deleting all images to a function and not duplicate the code.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+				All: true,
+			})
+			if err != nil {
+				util.Failed("Failed to list images: %v", err)
+			}
+			err = deleteDdevImages(images)
```
rfay, line 86: Oh, I see you promoted, but didn't promote all of it. Shouldn't you be able to promote the whole thing? With a deleteAll arg?

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,99 @@
...
+				util.Failed("Failed to delete image tag", err)
+			}
+			util.Success("Finished cleaning ddev projects")
+		}
+	},
+}
+
+func init() {
+	CleanCmd.Flags().BoolP("all", "a", false, "Clean all ddev projects")
+	CleanCmd.Flags().Bool("dry-run", false, "Run the clean command without deleting")
```
rfay, line 97: Thanks for trying out .Bool() - I just hate all those single-purpose variables everywhere.

cmd/ddev/cmd/delete-images.go
```diff
@@ -130,3 +122,21 @@ func init() {
 	DeleteImagesCmd.Flags().BoolVarP(&deleteAllImages, "all", "a", false, "If set, deletes all Docker images created by ddev.")
 	DeleteCmd.AddCommand(DeleteImagesCmd)
 }
+
+// deleteDdevImages removes Docker images prefixed with ddev-
+func deleteDdevImages(images []docker.APIImages) error {
```
rfay, line 127: As above, I think this can be done more completely by making this function find the images itself. Then the arg would be `deleteAll bool`. Without deleteAll it would just do old images.

rfay (changes requested):
Making progress, thanks so much for this!

cmd/ddev/cmd/clean.go
mikebarkas: Yeah, I confused this with the .downloads at the project level.
I will fix.

cmd/ddev/cmd/clean.go
mikebarkas: Like this?
```
    confirm := util.Prompt("Confirm removal of the listed items? Y/N", "N")
    if confirm == "Y"  {
```

cmd/ddev/cmd/clean.go
mikebarkas: Good idea, thanks. I changed it.

cmd/ddev/cmd/clean.go
mikebarkas: Ah.. that method is better. I renamed to ".downloads" too. Thanks.

cmd/ddev/cmd/delete-images.go
mikebarkas: How do we define "old" images?

cmd/ddev/cmd/clean.go
mikebarkas: I will move this to the function.

cmd/ddev/cmd/clean.go
mikebarkas: I will move this to the function.

mikebarkas: I think it makes sense to have `deleteDdevImages()` get its images and have a deleteAll argument.

I can then move the delete all logic to be first in the command.

cmd/ddev/cmd/clean.go
mikebarkas: I added docker

cmd/ddev/cmd/clean.go
mikebarkas: Updated, thanks.

cmd/ddev/cmd/clean.go
mikebarkas: I implemented this.

cmd/ddev/cmd/clean.go
```diff
@@ -62,28 +64,21 @@ Additional commands that can help clean up resources:
...
+		confirm := util.Prompt("Are you sure you want to continue? Y/N", "N")
+		if confirm == "Y" {
+			globalDdevDir := globalconfig.GetGlobalDdevDir()
+			_ = os.RemoveAll(filepath.Join(globalDdevDir, "testcache"))
+			_ = os.RemoveAll(filepath.Join(globalDdevDir, "bin"))
 
 			output.UserOut.Print("Deleting snapshots and downloads for selected projects...")
 			for _, project := range projects {
 				// Delete snapshots and downloads for each project
-				_ = os.RemoveAll(filepath.Join(project.AppRoot, ".ddev", "downloads"))
```
rfay: Let's check the error and `util.Warn()` about it, but not fail. If we found it and can't delete it, there's something going on.

cmd/ddev/cmd/delete-images.go
rfay: The "old" images are the ones that don't go with the current version. That's what you did in the delete images work - looking to see what matched current and what didn't. And the --all was all of them regardless of match.

cmd/ddev/cmd/clean.go
rfay: Lower-case n will work right IIRC, lower-case `y` won't here. But don't ask me... try it out! Step through it, improve it :)

rfay:
You'll want to rebase this when you get back to it, thanks.

cmd/ddev/cmd/clean.go
mikebarkas: I will add the error check.

mikebarkas:
@rfay 

- I refactored the delete images function to be reusable by other commands.
- I removed the flag variable placeholders for delete images and used .BoolP
- Created a vscode debug launcher for delete images command
- Compiled and tested locally deleting images with and without --all
- Rebased local branch and pushed

This is ready for review when the builds and tests finish.

rfay:
It seems like the CircleCI failure is a problem with Pantheon, so can be ignored. It's failing there on other tests as well.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,101 @@
...
+				output.UserOut.Printf("%v Snapshots for project %s will be deleted", len(snapshots), project.Name)
+			} else {
+				output.UserOut.Printf("No snapshots found for project %s", project.Name)
+			}
+		}
+
+		dryRun, _ := cmd.Flags().GetBool("dry-run")
+		if dryRun {
+			util.Warning("Dry run terminated without removing items")
+			os.Exit(1)
```
rfay, line 65: No biggie, but I think this should just be a `return`, which will exit success rather than error.

If it weren't to be a success (which seems ok with dry run) then I would have terminated it with utiil.Failed() which does an exit.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		confirm := util.Prompt("Are you sure you want to continue? y/n", "n")
+		if strings.ToLower(confirm) == "y" {
```
rfay, line 69: I think it would be more natural to check for *not* "y" and return if not. Then the rest could continue without being in an "if" stanza.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,101 @@
...
+			globalDdevDir := globalconfig.GetGlobalDdevDir()
+			_ = os.RemoveAll(filepath.Join(globalDdevDir, "testcache"))
+			_ = os.RemoveAll(filepath.Join(globalDdevDir, "bin"))
+
+			output.UserOut.Print("Deleting snapshots and downloads for selected projects...")
+			for _, project := range projects {
+				// Delete snapshots and downloads for each project
+				err = os.RemoveAll(project.GetConfigPath(".downloads"))
```
rfay, line 77: I didn't know that RemoveAll() returns nil if the path doesn't exist, so this works out fine.

rfay: This would be cleaner with a for loop of a slice of strings, since the action is the same everywhere. No change required.

cmd/ddev/cmd/clean.go
```diff
@@ -0,0 +1,101 @@
...
+					util.Warning("There was an error removing .downloads for project %v", project)
+				}
+				err = os.RemoveAll(project.GetConfigPath("db_snapshots"))
+				if err != nil {
+					util.Warning("There was an error removing db_snapshots for project %v", project)
+				}
+			}
+
+			output.UserOut.Print("Deleting Docker images that ddev created...")
+			err = deleteDdevImages(cleanAll)
```
rfay, line 88: cleanAll was set based on "all projects" here, but your function is deciding between "all DDEV images" and "current DDEV images", which is a different decision.

IMO you can just use `true` here.

rfay (changes requested):
Looks like almost done, thanks for all the good work. I think there are a couple of comments here that are actionable.

cmd/ddev/cmd/clean.go
mikebarkas: I had true there and then changed it. haha.. I will change it back, I agree.

cmd/ddev/cmd/clean.go
mikebarkas: ok, I agree.

cmd/ddev/cmd/clean.go
mikebarkas: Happy path.. is better.

rfay (approved):
Looks good! I tested on macOS and Linux, blew away everything a couple of times. 

Please take another read-through looking especially for mistakes that might cause a deletion-of-wrong-thing disaster, and let me know when you're ready for me to pull it.

Congrats!

mikebarkas:
Looks good to me, I think it can be merged. Thanks for your feedback and guidance, it is appreciated.

PR #3689, opened 2022-03-11
Bump docker-compose to 2.3.3 to solve problems with ddev exec stdout capture, fixes #3698

### LINKED ISSUE #3698: 1.19.0: Assignment of container stdout to a variable doesn't work (stdout/stderr issues)

_No response_

### Current Behavior

Hi,

I am using some custom commands and I assign the result of such a command in a variable (in some other bash script or even in other ddev command) : 

`MY_VARIABLE=$(ddev my-custom-command)`

The current result is an empty variable.

### Expected Behavior

Before 1.19.0, the variable was well assigned.

### Steps To Reproduce

As I guess my real custom commands won't be useful, I tested with the following:

- Create a custom command `.ddev/commands/web/web-command-test`

```
#!/usr/bin/env bash

## Description: TEST
## Usage: web-command-test value
## Example: "ddev web-command-test hello"

echo "$@"
exit 0
```

This command just returns a value that you pass as a parameter:

- Launch `ddev start`

- Then, in your terminal, run : 

```
export MY_VALUE=$(ddev web-command-test hello)

```
With 1.18.2 : nothing is displayed on the terminal; 
With 1.19.0: "hello" is displayed

- Then: 

```
echo ${MY_VALUE}

```
With 1.18.2 : "hello" is displayed on the terminal (this is the expected behavior)
With 1.19.0: a blank line is displayed

### Anything else?

_No response_

rfay:
## The Problem/Issue/Bug:

In v1.19.0 it is impossible to capture stdout from `ddev exec`. For example `ddev exec ls -lR >/dev/null` will display the output. This seems to be a regression in docker-compose v2.2.3, which was used in ddev v1.19.0

Pointed out in:

* https://github.com/drud/ddev-contrib/issues/206

Bump to docker-compose v2.3.3, seems to solve it.

`ddev exec ls -l >/tmp/out.txt` should show nothing. And you should find the output in /tmp/out.txt

* I added to TestCmdExec a section that deliberately redirects output to a file. However, I wasn't able to recreate the actual behavior that we all saw even when I ran the test against v1.19.0 (and docker-compose v2.2.3). The test succeeded. I assume this must have something to do with pty detection by docker-compose.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3689"></a>

### DIFF
--- a/docs/users/troubleshooting.md
+++ b/docs/users/troubleshooting.md
@@ -202,10 +202,10 @@ For example, if your Dockerfile were
 ```dockerfile
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
-RUN npm install --global gulp-cli
+RUN npm install --global forever
 ```
 
-You could test it with `ddev ssh`, `sudo -s` and then `npm install --global gulp-cli`
+You could test it with `ddev ssh`, `sudo -s` and then `npm install --global forever`
 
 The error messages you get will be more informative than messages that come when the Dockerfile is processed.
 
--- a/cmd/ddev/cmd/exec_test.go
+++ b/cmd/ddev/cmd/exec_test.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"fmt"
 	"github.com/drud/ddev/pkg/ddevapp"
+	"github.com/drud/ddev/pkg/fileutil"
+	"github.com/drud/ddev/pkg/util"
 	"github.com/stretchr/testify/require"
 	"os"
 	"path/filepath"
@@ -112,4 +114,17 @@ func TestCmdExec(t *testing.T) {
 	content, err := os.ReadFile(filepath.Join(site.Dir, filename))
 	assert.NoError(err)
 	assert.Equal("This file was piped into ddev exec", string(content))
+
+	// Make sure that redirection of output from ddev exec works
+	bashPath := util.FindBashPath()
+	f, err := os.CreateTemp("", "")
+	err = f.Close()
+	assert.NoError(err)
+	defer os.Remove(f.Name()) // nolint: errcheck
+
+	_, err = exec.RunHostCommand(bashPath, "-c", fmt.Sprintf("%s exec ls -l /usr/local/bin/composer >%s", DdevBin, f.Name()))
+
+	out, err = fileutil.ReadFileIntoString(f.Name())
+	assert.NoError(err)
+	assert.Contains(out, "/usr/local/bin/composer")
 }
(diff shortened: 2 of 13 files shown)


rfay:
* https://github.com/docker/compose/issues/9277
makes this not work at all on Linux. A partial workaround (not sure if other bugs may be encountered) will be to consolidate the context directory by copying anything in .ddev/web-build into .ddev/.webimageBuild, so the Dockerfile is in the same directory as context directory.

bserem:
I tested variable assignment in gitpod and it works fine.

**This is 1.19.0.** You can the variable assignment does not work:
image

**This is with the fix from this PR. It works as expected:**
image

docs/users/troubleshooting.md
```diff
@@ -202,10 +202,10 @@ For example, if your Dockerfile were
 ```dockerfile
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
-RUN npm install --global gulp-cli
+RUN npm install --global forever
```
jonaseberle, line 205: The second commit seems unrelated to the PR.
Careful when merging.

jonaseberle (approved):
That was weird with docker-compose v2.2.3 ;) I couldn't even figure out to which output stream the output went.

This is now back to sane again.

Nota bene: This is a different output stream issue but would be nice if we could get it under control once:
```
# this should not output anything
ddev exec 'sh -c "echo goes to stderr 1>&2"' 2> /dev/null
# compare with
webContainer="$(ddev describe -j | jq -r '.raw.services.web.full_name')"
docker exec "$webContainer" sh -c "echo goes to stderr 1>&2" 2> /dev/null
```

rfay:
Unfortunately there's a new set of problems with docker-compose 2.3+, with `docker-compose build`. Not sure it will be possible to solve, but I'll study it and open an issue. I'm thinking we'll have to go back to 2.2.2 and just clean up the leftover build containers that were the reason for moving to 2.2.3.

rfay:
@jonaseberle I believe there's significant progress on a number of the stdout vs stderr issues in the docker-compose queue. Not sure it will ever stabilize enough for us to use them :(

rfay:
@jonaseberle it's also possible to go a completely different way with `ddev exec` and either use docker or docker api instead of using docker-compose, and I think that might work out. I already did some of this a few months ago, but didn't want to disrupt things as they were... but the basics are there.

docs/users/troubleshooting.md
rfay: Thanks so much @jonaseberle - I happened to be testing a trivial Dockerfile to figure out what was going on with https://github.com/docker/compose/issues/9277 (and https://github.com/docker/compose/issues/9276) and so was actually looking at the example and docs. It turned out that the Dockerfile.example and docs were outdated because gulp-cli has long been installed by default. So I just fixed the example and docs.

rfay:
stdout/stderr issues with exec are longstanding and closely tied to docker-compose issues, 
* https://github.com/drud/ddev/issues/2839

PR #3690, opened 2022-03-11
More accurately reflect necessary setup for D9.

chrisfromredfin:
## The Problem/Issue/Bug:
When the user clones a lean repository but has not yet run the
`composer install` command, then using `ddev config --auto` will not
work because the files that the detection needs are not yet present.

Updates the documentation that users were following and not having luck with.

Clone a lean repository for a Drupal 9 site (say, from a Pantheon site using Integrated Composer). Run `ddev config --auto` before doing a `composer install`. Note that the generated config.yaml will not be Drupal 9.

Documentation update only.

## Related Issue Link(s):
Issue not filed. Discussed in Drupal Slack: https://drupal.slack.com/archives/C5TQRQZRR/p1646971651968869

## Release/Deployment notes:
n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3690"></a>

### DIFF
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -192,7 +192,7 @@ Note that the git URL shown below is an example only, you'll need to use your ow
 ```bash
 git clone https://github.com/example/example-site
 cd example-site
-ddev config --project-type=drupal9
+ddev config --project-type=drupal9 --docroot=web --create-docroot
 ddev composer install
 ddev launch
 ```
--- a/docs/users/cli-usage.md
+++ b/docs/users/cli-usage.md
@@ -192,7 +192,6 @@ Note that the git URL shown below is an example only, you'll need to use your ow
 ```bash
 git clone https://github.com/example/example-site
 cd example-site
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
I think mostly we just need to specify project type, would "web" not be in a normal checkout?

docs/users/cli-usage.md
```diff
@@ -192,7 +192,7 @@ Note that the git URL shown below is an example only, you'll need to use your ow
 ```bash
 cd example-site
-ddev config --auto
+ddev config --project-type=drupal9
```
rfay, line 195:
```suggestion
ddev config --project-type=drupal9 --docroot=web --create-docroot
```

This resolves the problem that the comment was talking about. I never actually knew that the docroot might not exist in a Drupal9 project.

rfay:
Thanks so much for taking the time to improve this! I learn something every day.

rfay (approved):

PR #3695, opened 2022-03-13
Automatically run unknown commands in the web container

gilbertsoft:
## The Problem/Issue/Bug:

Currently to run a command in the web container a proper custom command has to be defined. It would be nice if unknown commands are just forwarded to the web container without the need of defining a custom command. If the command is not found it still can fail later.

Unknown commands are simply forwarded to the web container and if found the exit code is properly returned.

If the command is not found an error is shown like it's done before.

- run `ddev ls`, the content of `var/www/html` should be listed
- run `ddev invalid`, an error should be shown

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3695"></a>

### DIFF
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -185,6 +185,11 @@ func TestCustomCommands(t *testing.T) {
 	app.Type = nodeps.AppTypePHP
 	err = app.WriteConfig()
 	assert.NoError(err)
+	// Drush may still be installed in /usr/local/bin, get rid of it
+	_, _, err = app.Exec(&ddevapp.ExecOpts{
+		Cmd: "rm -f /usr/local/bin/drush",
+	})
+	assert.NoError(err)
 
 	// Make sure that all the official ddev-provided custom commands are usable by just checking help
 	for _, c := range []string{"launch", "mysql", "php", "xdebug"} {
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -2,6 +2,10 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/dockerutil"
 	"github.com/drud/ddev/pkg/globalconfig"
@@ -13,8 +17,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"gopkg.in/segmentio/analytics-go.v3"
-	"os"
-	"path/filepath"
 
 	"time"
 )
@@ -139,7 +141,33 @@ func Execute() {
 	// bind flags to viper config values...allows override by flag
 	viper.AutomaticEnv() // read in environment variables that match
 
-	if err := RootCmd.Execute(); err != nil {
+	RootCmd.SilenceErrors = true
+	err := (error)(nil)
+
+	if err = RootCmd.Execute(); err == nil {
+		return
+	}
+
+	if !strings.Contains(err.Error(), "unknown command") {
+		os.Exit(-1)
+	}
+
+	// Try to run the command in the web container directly
+	args := os.Args[1:]
+	RootCmd.SetArgs(append([]string{"exec"}, args...))
+
+	if err = RootCmd.Execute(); err == nil {
+		return
+	}
+
+	if strings.Contains(err.Error(), "command not found") {
+		// Run the inital command again without surpressing errors
+		RootCmd.SilenceErrors = false
+		RootCmd.SetArgs(args)
+		if err = RootCmd.Execute(); err == nil {
+			// Should never happen
+			os.Exit(-2)
+		}
 		os.Exit(-1)
 	}
 }


rfay:
You'll want to rebase when you get back to this, thanks.

rfay:
Looking forward to seeing this see the light of day. The only issue is whether it can be predictable for people. How will help work? How will people know what anything does?

rfay:
Do you think this has a future @gilbertsoft ? It's nice, but the ambiguity of what to expect is a little troublesome. Easy enough to fix the test failure.

gilbertsoft:
Lets finish this one, will rebase later...

gilbertsoft:
Tested now and works like expected. I'm currently restarting the failing tests which don't have real issues...

rfay:
It turns out that buildkite tests are run with Drupal6 by default, meaning that TestCustomCommands starts out with Drupal6. Then when the test changes the config type to php, it doesn't restart the container, meaning drush is still in there, so the test detects drush in that circumstance. At least I think this was the issue. I pushed https://github.com/drud/ddev/pull/3695/commits/b5b4f6349a9a0d8b652296290ad3710b6f6b2a14 to try to fix this.

rfay:
@cmuench @jonaseberle Would love it if you could review this and try it out. It's a slick enhancement, my only concern is that it could cause confusion because of the undefined nature of commands in the future.

jonaseberle (approved):
That works very well. Thank you @gilbertsoft!

I've tested with 
`ddev ssh` (DDEV builtin)
`ddev mysql` (DDEV global command)
`ddev dump-db live` (DDEV custom command)
`ddev vendor/bin/phpunit`
`ddev ls` (command from $PATH)
`ddev jq --help` (check if parameter is appended to ddev or to the sub command)
`ddev 'jq --help'`
`ddev 'source .env; echo $TYPO3_CONTEXT'` (bash builtin)
`ddev 'ls -als | grep public' && echo success || echo failed` (pipes and exit codes)
`ddev 'ls -als | grep nonexisting' && echo success || echo failed`

jonaseberle:
I think it is conceptually in a tough spot: 
For working inside a container we have `ddev ssh`, for executing commands there is `ddev exec`.

This feature works beautifully, but on the other hand it only saves the user from using `ddev exec`.

My opinion: Having 2 ways for the same thing should generally be avoided. It makes it more difficult to show a coherent usage in docs/blog posts/support channels. Some will love it because they need to type 5 characters less, others will not use it in examples because they don't like ambiguity. I am quite sure a Twitter post about it would get generally good responses, too.
My question is if the advantages outweigh.

gilbertsoft:
Please see https://discord.com/channels/664580571770388500/950743021790576700 where this idea was born. I personally don't have a final opinion, it was just an idea so simlify things so I prepared this POC.

rfay:
There are `ddev exec`, `ddev .`, and loads of shell-command wrappers. I think the confusion among all those things is in fact the real problem. 

One dramatic option would be to put this on major-release track (v1.20) and at the same time remove the wrappers, like `ddev php` and `ddev typo3`, etc. I guess the issue with some of those is that the command is not necessarily in the $PATH.

gilbertsoft:
Yeah, removing existing custom commands sounds like a good idea. But this also removes them from the command list which was always a nice feature, but well, personally could live with that. We could introduce here sth like done for pull and push for the less experienced users to get a hint on project specific commands maybe...

The $PATH should be set correctly as long as the user doesn't change the install structure. And if he does he also should take care of a proper $PATH override which I'm also doing for some special cases:

```
web_environment:
- PATH=/var/www/html/public/bin:$PATH
```

gilbertsoft:
We also can enhance the `exec` command description and say explicit this is the default command and `exec` can be omitted.

rfay:
> We also can enhance the exec command description and say explicit this is the default command and exec can be omitted.

Interesting thinking @gilbertsoft - However, exec is still used for other services (db, add-on, etc). But still worthy thinking.

gilbertsoft:
> > We also can enhance the exec command description and say explicit this is the default command and exec can be omitted.
> 
> Interesting thinking @gilbertsoft - However, exec is still used for other services (db, add-on, etc). But still worthy thinking.

I guess sth like `ddev -s db ls` should also work but not tested so far.

cmuench:
I'm still a little unsure about this change. There are pros and cons.
IMHO one of the advantages of ddev is that it provides such a simple way to add custom commands for developers which can be shared in a team. There is no other tool with that simple approach. We should do anything to keep custom commands.
Custom commands whould be listed if we see the command list.

If I type only "ddev", then we should see the command list.
On the other hand this change provides a way to deduce a lot of this one line proxy shell scripts.

Then we have `ddev .` which really less people know.

rfay:
Rebased.

shaal:
My 2 cents -
* I like the separation between `ddev <custom-ddev-command>` and `ddev . <any-command>`.
* I think this feature could be helpful - when you run `ddev unknown_command` and it's not found in custom commands, `ddev` could prompt - "Did you want to run `unknown_command` inside the container?", hitting `y` would run that command in the container for you.
* I was sure I opened an issue about this before, but I guess I didn't 😅, in DrupalPod I added generic custom command `exec_d` that check your current directory, and runs the command in the same relative directory in the container
https://github.com/shaal/DrupalPod/blob/main/.ddev/commands/host/exec_dir

gilbertsoft:
@rfay well, my PoC here showed it would work. But like others I'm also not sure if we should merge it at this point. So I suggest to close this PR until a real issue pops up and get back on it if really needed....

rfay:
I do think it is brilliant, but the fact that we lose control of the commands, and don't have a way to provide help, and it's pretty disruptive. But I think there might be a way to provide a small bit of help for the most important commands, not sure. But probably you're right. It certainly can't go in until I start on non-minor-release candidate work.

rfay:
I think we have a general consensus that we lose too much control of the user workflow with this approach. Closing.

PR #3700, opened 2022-03-14
Use bin/magento in magento command, fixes #3699

### LINKED ISSUE #3699: Magento command does not work if working_dir is changed

https://gist.github.com/cmuench/a939548463eedc656c358a0f6f5c7f62

### Current Behavior

If a working_dir is set to a sub-directory and the shop exits there the `ddev magento` command does not work.

The command prints the error message: "magento is not available in your in-container $PATH"

The ddev command is looking for the command "magento" in the *$PATH* which does not exist, because it's shipped by Magento/Adobe Commerce and available after a successful `composer install`.

### Expected Behavior

The `bin/magento` CLI tool should be called. I expect the output of the tool.

### Steps To Reproduce

1. Open an old project with a Magento installed in a sub-directory.
2. Verify that the working_dir is set to this subdirectory.
3. Start the project
4. Run `ddev magento`

### Anything else?

_No response_

cmuench:
## The Problem/Issue/Bug:

* #3699

Currently the command looks for a script/binary "magento"
in the $PATH which is not available in a general Magento installation.
Change the script to use the shipped bin/magento.
If Magento is not installed via "composer install", the "bin/magento"
script is not available. So the check if ok with the right path.

Fix the path to magento script.

Start a Magento project with a Magento installation in a sub-directory and set working_dir.

Automatic test requires a Magento installation or a dummy script. Let's see if we can add that in the future.

## Release/Deployment notes:

Issue was introduced in ddev v1.19.0.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3700"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/magento
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/magento
@@ -7,10 +7,10 @@
 ## ProjectTypes: magento2
 ## ExecRaw: true
 
-if ! command -v magento >/dev/null; then
-  echo 'magento is not available in your in-container $PATH'
+if ! command -v bin/magento >/dev/null; then
+  echo 'bin/magento is not available in your installation.'
+  echo 'Please verify that you installed the shop in your working dir.'
   exit 1
 fi
 
-magento "$@"
-
+bin/magento "$@"


rfay:
@cmuench I'm surprised this was introduced in ddev v1.19. Not sure how that could be.

cmuench:
@rfay I am not 100% sure if it was introduced in v1.19.
We have only two projects where the Magento root is a sub directory of the repo root.
In this projects we did not use the working_dir option.
Maybe the issue is older.

rfay (approved):
I'm good with this if it solves your problem and doesn't break regular default working_dir usage.

PR #3717, opened 2022-03-18
Document how to add env-file support

mrhorse:
## The Problem/Issue/Bug:
Please see https://github.com/drud/ddev-contrib/pull/210#issuecomment-1071916362

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3717"></a>


rfay (approved):
Awesome. This should be default for ddev, just have to check and see if the whole thing fails if .env doesn't exist.

PR #3721, opened 2022-03-19
Add xz and bzip2 support for import-db, export-db, import-files, fixes #3609

### LINKED ISSUE #3609: DDEV commands "import-db" and "export-db" should support bz2 and xz compressed dumps
We usually transfer our SQL files in "bzip2" or even "xz" these days, however importing them into DDEV Local only supports plain SQL and "gzip".

It would be really cool, if DDEV Local should also detect "bz2" and "xz" file extensions.

I could do some bash piping, but that's just ugly :)

**Additional context**
I think the place for import-db is around here (but I'm no go expert) https://github.com/drud/ddev/blob/master/pkg/ddevapp/ddevapp.go#L466-L470

I'm willing to sponsor this feature (depends on the amount of work to be done of course) - just contact me with a quote, if you think this feature request is a good idea (I'm 100% OK if you think it's not needed).

rfay:
## The Problem/Issue/Bug:

* #3609 

Add bzip/xz formats for import and export

## TODO: 
- [x] Add test coverage for compression formats on export-db and import-files
- [x] Make sure the TestCmd commands have proper coverage for the new types

- [x] `ddev import-db --src=` with a *.sql.bz2 and *.sql.xz 
- [x] `ddev export-db --bzip2 --file=something.bz2`
- [x] `ddev export-db --xz --files=something.sql.xz`
- [x] Verify (re-import) the output files

Each of the tests that deal with import-db, export-db, and export-files has been updated to try all 3 compression techniques.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3721"></a>


rfay:
@bmack (and anybody else) I'd love if you could give this a quick test. You can just use gitpod, https://gitpod.io/#https://github.com/drud/ddev/pull/3721 or you can download artifacts and put them in path, https://github.com/drud/ddev/actions/runs/2012196073

PR #3739, opened 2022-03-30
[Tests only] TestComposerVersion fixup for composer 2.3.x

rfay:
In Composer 2.3.x they changed the output of composer --version
and it broke TestComposerVersion

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3739"></a>


rfay:
Github never was actually testing this code, but rather other code. Apparently due to an outage. Replaced with #3742

PR #3744, opened 2022-04-01
Try docker-compose 2.4.0

rfay:
## The Problem/Issue/Bug:

The previous release, 2.3.4, was unusable, so we'll run a set of tests against this one.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3744"></a>


rfay:
docker-compose 2.4.0 is nonfunctional - mounts parent directory instead of what it should mount. 

* https://github.com/docker/compose/issues/9344

PR #3749, opened 2022-04-02
ddev snapshot -a should snapshot all apps and not fail just because one isn't running, fixes #3707

### LINKED ISSUE #3707: ddev snapshot --all fails if any projects are stopped

https://gist.github.com/FLeudts/51c1c7ab6d0051c1939d1305b2cd3507 

### Current Behavior

If I run the ddev snapshot --help command it tells me that ddev snapshot --all should only snapshot running projects. It still tries to snapshot a stopped project and then fails because the db container is stopped. 

### Expected Behavior

ddev snapshot --all should ignore stopped projects or the description should be edited. 

### Steps To Reproduce

1. Have two or more projects
2. Stop the first project in ddev list and start the others
3. Execute ddev snapshot --all
4. No snapshots will be created

### Anything else?

_No response_

mikebarkas:
## The Problem/Issue/Bug:
`ddev snapshot --help` command says ddev snapshot --all should only snapshot running projects.
It still tries to snapshot a stopped project and then fails.

`ddev snapshot --all` will now start any stopped or paused apps and create a snapshot and then place the app back to its original state of stopped or paused.

- Have three or more projects
- Stop one project and pause another and start the others
- Execute ddev snapshot --all
- Will create a snapshot of all projects
- Project stopped or paused will be set back to stopped or paused after snapshots

Run existing tests

## Related Issue Link(s):
#3707

## Release/Deployment notes:

No deployment changes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3749"></a>

### DIFF
--- a/cmd/ddev/cmd/snapshot.go
+++ b/cmd/ddev/cmd/snapshot.go
@@ -68,6 +68,13 @@ func listAppSnapshots(app *ddevapp.DdevApp) {
 
 func createAppSnapshot(app *ddevapp.DdevApp) {
 
+	// If the database is omitted, do not snapshot
+	omittedContainers := app.GetOmittedContainers()
+	if nodeps.ArrayContainsString(omittedContainers, "db") {
+		util.Warning("Database is omitted for project %s, skipping snapshot", app.GetName())
+		return
+	}
+
 	appStatus := app.SiteStatus()
 	// If the app is not running, then start it to create a snapshot.
 	if appStatus != ddevapp.SiteRunning {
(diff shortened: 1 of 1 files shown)


rfay:
How about starting apps one at a time that aren't running, snapshot them and then stop them? Snapshot auto-starts, right?

What your strategy here is is good, but most people who want a snapshot-all really want all to be done.

FLeudts:
One issue that could then exist is that it is possible to have projects without a DB. If they are started and the snapshot is tried it would fail again? In my usecase I had projects that weren't started because of this issue.

rfay:
@FLeudts is right, if `omit_containers[db]` then there's nothing to be done, and the app.Snapshot() fails with
```
Failed to snapshot j: unable to snapshot database,
your db container in project j is not running.
Please start the project if you want to snapshot it.
If deleting project, you can delete without a snapshot using
'ddev delete --remove-data --yes',
which will destroy your database
```

This isn't a bug with this PR, but it wouldn't do any harm to have app.Snapshot() output a message saying "not snapshotting because no database" and return nil.

rfay (changes requested):
This works well and looks fine, thanks. Just the comments in https://github.com/drud/ddev/pull/3749#issuecomment-1089852818

mikebarkas:
I will make the requested changes. Thanks for your feedback.

mikebarkas:
Update

- I checked for omittedContainers before app.Snapshot. I don't think we should start an app and then try app.Snapshot if omit_container[db]
- I changed util.Failed to util.Warning so the snapshots will continue even if one fails. I colorized the warning message to red

mikebarkas:
Restarted `buildkite/ddev-macos-m1`

rfay:
Rebased. Should get it in shortly, thanks!

PR #3752, opened 2022-04-08
More precise output for xhprof enable

AronNovak:
## The Problem/Issue/Bug:
`ddev xhprof on` states that it generates a profile for web requests.

It makes a precision that it works for CLI processes too.

`ddev xhprof on` should have a more precise output now.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3752"></a>


rfay (approved):
Thanks!

PR #3763, opened 2022-04-14
Set permissions for GitHub actions

nathannaveen:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3763"></a>

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -146,9 +146,6 @@ jobs:
           key: ${{ github.sha }}-${{ github.ref }}-notarize-macos
 
   artifacts:
-    permissions:
-      actions: write  # for skx/github-action-publish-binaries to attach binaries to release artifacts
-      contents: read  # for actions/checkout to fetch code
     name: Upload artifacts
     runs-on: ubuntu-20.04
     needs: [build-most, sign-windows, notarize-macos]
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -14,6 +14,9 @@ env:
   DOCKER_CLI_EXPERIMENTAL: enabled
   DDEV_DEBUG: true
 
+permissions:
+  contents: read
+
 jobs:
   build-most:
     name: Build DDEV executables except Windows
@@ -143,6 +146,9 @@ jobs:
           key: ${{ github.sha }}-${{ github.ref }}-notarize-macos
 
   artifacts:
+    permissions:
+      actions: write  # for skx/github-action-publish-binaries to attach binaries to release artifacts
+      contents: read  # for actions/checkout to fetch code
     name: Upload artifacts
     runs-on: ubuntu-20.04
     needs: [build-most, sign-windows, notarize-macos]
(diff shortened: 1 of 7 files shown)


rfay:
Thanks for this @nathannaveen - 

I'm still no expert at github workflows. But I do note from [update doc](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#:~:text=GitHub%20Actions%20now%20lets%20you,token%20when%20a%20job%20completes) that

> Pull requests from public forks are still considered a special case and will receive a read token regardless of these settings

So since all work on ddev is done with public forks.. does that make this PR irrelevant?

Thanks so much for your careful review of this. I've spent some time reviewing the difference between `pull_request` and `pull_request_target` this morning and am still a bit hazy, but it doesn't seem like we're necessarily doing the right thing in each case.

nathannaveen:
> Thanks for this @nathannaveen -
> 
> I'm still no expert at github workflows. But I do note from [update doc](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#:~:text=GitHub%20Actions%20now%20lets%20you,token%20when%20a%20job%20completes) that
> 
> > Pull requests from public forks are still considered a special case and will receive a read token regardless of these settings
> 
> So since all work on ddev is done with public forks.. does that make this PR irrelevant?
> 
> Thanks so much for your careful review of this. I've spent some time reviewing the difference between `pull_request` and `pull_request_target` this morning and am still a bit hazy, but it doesn't seem like we're necessarily doing the right thing in each case.

I will let you decide, thanks!

rfay:
Still interested in our opinion, and why this piqued your interest. I assume you're a ddev user and were studying the workflow...

.github/workflows/master-build.yml
```diff
@@ -143,6 +146,9 @@ jobs:
           key: ${{ github.sha }}-${{ github.ref }}-notarize-macos
 
   artifacts:
+    permissions:
+      actions: write  # for skx/github-action-publish-binaries to attach binaries to release artifacts
+      contents: read  # for actions/checkout to fetch code
```
rfay, line 151:
```suggestion
```

This section is superfluous because you set it globally above, right?

rfay (changes requested):
This all looks great, and thanks for taking the initiative to do it. I've spent more time with GITHUB_TOKEN and permissions now thanks to you. I do think the one item is superfluous, so please take a look.

nathannaveen:
> This all looks great, and thanks for taking the initiative to do it. I've spent more time with GITHUB_TOKEN and permissions now thanks to you. I do think the one item is superfluous, so please take a look.

Thanks! I accepted the request!

.github/workflows/master-build.yml
rfay: All you had to do was click "Commit suggestion" on this :)

rfay (approved):
Thanks!

PR #3768, opened 2022-04-21
Added DB_CHARSET and DB_COLLATE constants to the wp-config.php

### LINKED ISSUE #3767: WordPress -- created wp-config-ddev.php is missing some settings

_No response_

### Current Behavior

When non-latin texts (in my case Japanese characters) are saved into database, it is corrupted/garbbled.

I found that this is due to `wp-config-ddev.php` created during `ddev config --project-type=wordpress` is missing following two `define()`

```
/** Database charset to use in creating database tables. */
define( 'DB_CHARSET', 'utf8' );

/** The database collate type. Don't change this if in doubt. */
define( 'DB_COLLATE', '' );
```

These two, particulary the `DB_CHARSET`, when missing, WordPress will save into DB in latin and not utf8. This is not a custom extra setting needs to be added for unusual situation as such. It is coded in `wp-config-sample.php` of the WordPress Core
https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php

### Expected Behavior

`wp-config-ddev.php` created during `ddev config --project-type=wordpress` should include `DB_CHARSET` and `DB_COLLATE` thus non-latin characters should be saved without corrupting.

Currently, to avoid this, after `ddev config --project-type=wordpress` I have to add below lines to autocreated `wp-config.php`, after `ddev wp core install` step, and I also have to drop all tables and restart the install. I expect this fix extra step should not be needed.

```
/** Database charset to use in creating database tables. */
define( 'DB_CHARSET', 'utf8' );

/** The database collate type. Don't change this if in doubt. */
define( 'DB_COLLATE', '' );
```

### Steps To Reproduce

1. launch new WordPress site as per https://ddev.readthedocs.io/en/stable/users/cli-usage/#wordpress-quickstart
2. login into admin, create a new post. Copy and paste below Japanese text.
3. Publish the post.
4. Check the text on the editor and the frontend. Texts are corrupted to something like `~~~~~~~~` (or something similar, sorry, forgot to capture but you will know)

Please try with these Japanese text

For title: 

```
私の個人主義
```

For content: 

```
あなたは先刻まあどんな下宿学として気のためをうろついですます。

すでに生涯に参考心はすこぶるこうした満足ならないかもを申し上げていないをも相当潰さたないて、更にには与えでしございませない。

主義にもっならのは無論今をしかるにたでた。もっと槙さんで矛盾人そうお話をできるあっ人その一つこれか腐敗がというご評だろただうて、その今もどこか泰平空の買うば、大森君の事を手の私の同時にご病気と見えるば私個人にお意味が流行るようにどうか実楽に云っんないので、あたかもさぞ安心の曲げなりてならだろ事をなっですた。すなわちあるいは大素因を云わ事はいろいろ不愉快と弱らでしと、その去就のは充たすですてについてらであるてみありた。こういう上西洋のためこの生徒もあなた上をするたかと大森君が引張りだろでしょ、かたがたの今日ですにおいてご講演ですたましながら、金力のうちに通りに今でもの根本の時間思うているて、それほどの今へしのでこういう以上へついに抱いなくありと喜ぶたものでと、好いましたがこう小他人作ったのでんた。
```

### Anything else?

_No response_

rawdreeg:
## The Problem/Issue/Bug:

closes #3767.

When non-latin texts (in my case Japanese characters) are saved into the database, it is corrupted/garbled.

This is due to wp-config-ddev.php created during ddev config --project-type=wordpress is missing the following two constants.
```

/** Database charset to use in creating database tables. */
define( 'DB_CHARSET', 'utf8' );

/** The database collate type. Don't change this if in doubt. */
define( 'DB_COLLATE', '' );
```

These two, particularly the DB_CHARSET, when missing, WordPress will save into DB in Latin and not utf8.

Added the constant to the wp-config-ddev.php template file

Run `ddev config --project-type=wordpress` with this branch build and confirm the wp-config-ddev.php file contained definition constants for DB_CHARSET and DB_COLLATE

## Related Issue Link(s):
https://github.com/drud/ddev/issues/3767

## Release/Deployment notes:
Not that I know of.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3768"></a>

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Homebrew cache/restore
         uses: actions/cache@v3
         env:
-          cache-name: cache-homebrew-cache-1
+          cache-name: cache-homebrew-cache-2
         with:
           path: ~/Library/Caches/Homebrew
           key: ${{ runner.os }}-build-${{ env.cache-name }}
@@ -68,7 +68,7 @@ jobs:
       - name: Lima cache/restore
         uses: actions/cache@v3
         env:
-          cache-name: cache-lima-1
+          cache-name: cache-lima-2
         with:
           # npm cache files are stored in `~/.npm` on Linux/macOS
           path: ~/.lima
@@ -118,7 +118,7 @@ jobs:
           ddev poweroff
           colima stop
       - name: Store test results
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-test-results-${{ matrix.webserver }}
           path: /tmp/testresults/
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -39,7 +39,7 @@ jobs:
         if: matrix.os == 'ubuntu-20.04'
         run: ./.github/workflows/linux-setup.sh
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v3
         with:
           go-version: 1.*
           
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: rojopolis/spellcheck-github-actions@0.22.1
+      - uses: rojopolis/spellcheck-github-actions@0.23.0
         name: Spellcheck
       - name: Run markdownlint on docs
         uses: docker://avtodev/markdown-lint:v1
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -21,7 +21,7 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/setup-go@v2
+    - uses: actions/setup-go@v3
       with:
         go-version: '1.17'
     - uses: actions/checkout@v3
--- a/.github/workflows/macos-colima-setup.sh
+++ b/.github/workflows/macos-colima-setup.sh
@@ -5,9 +5,11 @@ set -eu -o pipefail
 # colima has golang as dependency, so is going to install go anyway.
 # So we have to get rid of it somehow.
 brew uninstall go@1.15 || true
+brew unlink go || true
+brew uninstall go@1.17 || true
 brew uninstall postgresql || true
-brew update && brew install colima docker docker-compose golang libpq mkcert mysql-client
-brew link --force libpq mysql-client
+brew update && brew install colima docker docker-compose go libpq mkcert mysql-client
+brew link --force go libpq mysql-client
 
 # This command allows adding CA (in mkcert, etc) without the popup trust prompt
 # Mentioned in https://github.com/actions/virtual-environments/issues/4519#issuecomment-970202641
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install build tools
         run: ./.github/workflows/linux-build-setup.sh
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v3
         with:
           go-version: 1.*
       
@@ -108,7 +108,7 @@ jobs:
         with:
           # We need to get all branches and tags for git describe to work properly
           fetch-depth: 0
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v3
         with:
           go-version: 1.*
 
@@ -200,32 +200,32 @@ jobs:
         run: ./.ci-scripts/generate_artifacts.sh ${{ github.workspace }}/artifacts
 
       - name: Upload all artifacts
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: all-ddev-executables
           path: ${{ github.workspace }}/artifacts/*
       - name: Upload macos-amd64 binary
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-macos-amd64
           path: .gotmp/bin/darwin_amd64/ddev
       - name: Upload macos-arm64 binary
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-macos-arm64
           path: .gotmp/bin/darwin_arm64/ddev
       - name: Upload linux-arm64 binary
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-linux-arm64
           path: .gotmp/bin/linux_arm64/ddev
       - name: Upload inux_amd644 binary
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-linux-amd64
           path: .gotmp/bin/linux_amd64/ddev
       - name: Upload windows_amd64 installer
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: ddev-windows-amd64-installer
           path: .gotmp/bin/windows_amd64/ddev_windows_installer*.exe
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install build tools
         run: ./.github/workflows/linux-build-setup.sh
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v3
         with:
           go-version: 1.*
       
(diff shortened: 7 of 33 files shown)


rfay:
@mandrasch could you please review and manually test this change?

mandrasch:
> @mandrasch could you please review and manually test this change?

Sure, I'll check and test this regarding the WordPress quickstart! 👍 

First thing I noticed right now is that we probably should discuss if we should use `defined() || define()` pattern as well for these values. 

btw: The official documentation for these config values can be found here:
https://wordpress.org/support/article/editing-wp-config-php/#database-character-set

I'll get back to both of you when I have tested this. Thanks @rawdreeg for submitting this! 👍

rawdreeg:
@mandrasch Thanks for looking at this. 

> First thing I noticed right now is that we probably should discuss if we should use defined() || define() pattern as well for these values.

Yes. But this is only an issue when DDEV generates wp-config.php. which is not based on[ wp-config-sample.php ](https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php).

Alternatively, I could define the additional property in WordpressConfig

https://github.com/drud/ddev/blob/0aa4e1d27cd77d73c06b026a0ff0629dac61603d/pkg/ddevapp/wordpress.go#L41-L64

then modify the wp-config template - adding the DB_CHARSET and DB_COLLATE constants as default.:

https://github.com/drud/ddev/blob/985d02017b96ccb80b9b974f9108c2e21069aa3c/pkg/ddevapp/wordpress/wp-config.php#L11-L18

I initially chose the path with the least amount of change but let me know how you want to move forward with this :-)

mandrasch:
Hey, I just had some time to think briefly about this.

First of all: I finally figured out why this bug is introduced, thanks for submitting this! Good catch! DDEV creates the empty 'db' with collation `latin1_swedish_ci`. WordPress reads this collation when DB_CHARSET/DB_COLLATE is not set and uses it (instead of utf8mb4). Thought all the time that this might be a bug within WordPress because 'utf8' should be used as default... took me some time ;-) 

image

So we should definitely fix this via initial config to avoid this:

A) The default and official [`wp-config-sample.php`](https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php) from WordPress has the following content (as you mentioned in your issue):

```
/** Database charset to use in creating database tables. */
define( 'DB_CHARSET', 'utf8' );

/** The database collate type. Don't change this if in doubt. */
define( 'DB_COLLATE', '' );
```

So my first vote would be to add DB_CHARSET and DB_COLLATE to DDEVs initial config when a new project is configured with `ddev config project-type=wordpress`. Then we would be also compliant to the current official WordPress suggestion.

```php
/* wp-config.php created by DDEV */
define( 'DB_CHARSET', 'utf8' );
define( 'DB_COLLATE', '' );
```

_Bonus info: WordPress will automatically upgrade to `utf8mb4` if possible, there is a ticket (https://core.trac.wordpress.org/ticket/48285) to use this as new default in wp-config.php. Its handled via [determine_charset()](https://github.com/WordPress/WordPress/blob/c6dbcf85766063b20aeb23ca207cc476818cdda2/wp-includes/wp-db.php#L832). More technical background infos: https://www.usablewp.com/character-set-and-character-encoding/_

<hr>

B) Regarding the additional DDEV specific overrides in `wp-config-ddev.php`: 

I can't think really think of a use case where this is needed? 

One general use case is that users import an existing wordpress project and these constants should be already defined in their wp-config.php and shouldn't be overridden by DDEV. DDEV in general doesn't really care / has a specific requirement like it is the case for DB_HOST, DB_PASSWORD as far as I understand.  

So my open question would be if maybe a fallback like the one for table prefix is useful?

```
/**
	 * Set WordPress Database Table prefix if not already set.
	 *
	 * @global string $table_prefix
	 */
	if ( ! isset( $table_prefix ) || empty( $table_prefix ) ) {
		// phpcs:disable WordPress.WP.GlobalVariablesOverride.Prohibited
		$table_prefix = 'wp_';
		// phpcs:enable
	}
```

$table_prefix is not a constant, for DB_CHARSET we would need to use the pattern mentioned above:

```
// Fallback to default values
defined( 'DB_CHARSET' ) || define( 'DB_CHARSET'', 'utf8' );
defined( 'DB_COLLATE' ) || define( 'DB_COLLATE', '' );
...
```

BUT: The big question would be if we break existing projects which accidently run by now in latin instead of utf8? WordPress docs state: 

> If DB_COLLATE and DB_CHARSET do not exist in your wp-config.php file, DO NOT add either definition to your wp-config.php file unless you read and understand [Converting Database Character Sets](https://codex.wordpress.org/Converting_Database_Character_Sets). And you may be in need of a WordPress upgrade.
https://wordpress.org/support/article/editing-wp-config-php/#database-collation

Long story short: Option A) should be definitely implemented in my opinion, about B) I'm not really sure. 

@rfay Is there a special reason why latin is used as default?

image

rfay:
@mandrasch the problem with db collation of latin1 was a bug and has been resolved (will be in v1.19.2 before long). Iit affected only mariadb 10.3

mandrasch:
Thanks for clarification, Randy! (And sorry for not researching in the open issues before asking )

I would propose to implement variant A) as describe above as fix for the initial problem. @rawdreeg Would you please adjust this PR accordingly (if you agree)?

I would also propose *not* to implement variant B since users should handle this themselves if they need other config values for their use cases.

rawdreeg:
Thanks @mandrasch, @rfay for your feedback and insights on this. I will update the PR with your recommendations. Thanks

rawdreeg:
@mandrasch I have updated this PR with the proposed changes. Thanks, R

rfay:
To be clear @rawdreeg - If you use mariadb 10.4, do you even have the problem this is trying to solve? Or is it a result of the bug where mariadb has latin1 charset instead of utf8mb4?

rfay:
But regardless, wordpress.go needs to be reformatted, `gofmt -w -s wordpress.go`

rawdreeg:
> To be clear @rawdreeg - If you use mariadb 10.4, do you even have the problem this is trying to solve? Or is it a result of the bug where mariadb has latin1 charset instead of utf8mb4? You can test by just upgrading to mariadb 10.4 (assuming you were using 10.3), ddev config --database=mariadb:10.4

Thanks, @rfay. I have tested this with MariaDB 10.6 and the issue is resolved on newer versions of MariaDB which means updating MariaDB should be enough in solving the problem. Otherwise, I will leave the PR open in case we'd want to keep wp-config.php in line with [wp-config-sample.php](https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php)

> But regardless, wordpress.go needs to be reformatted, gofmt -w -s wordpress.go
Thanks, I fixed this.

rfay:
It's not "newer" versions I don't think, just the one problem with a failed mariadb 10.3 build in ddev v1.19.1. I imagine that mariadb 10.2 would work fine as well.

I'm just saying... I'm sorry that bug wasted your time to investigate this. Do we need this PR if your problem was just a result of that bug? (I'm still open to it just fine).

rawdreeg:
> It's not "newer" versions I don't think, just the one problem with a failed mariadb 10.3 build in ddev v1.19.1. I imagine that mariadb 10.2 would work fine as well.
>
> I'm just saying... I'm sorry that bug wasted your time to investigate this. Do we need this PR if your problem was just a result of that bug? (I'm still open to it just fine).

Ah. I understand -- I have used DDEV for so long, that I am happy to help wherever I can. 

I'd say to align the ddev generated wp-config.php with [wp-config-sample.php](https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php) this PR might be useful still.

rfay:
@mandrasch if you approve this I think it can go in fine. Thanks!

rfay:
Oh, sorry, please rebase this as well to pick up some fixes that will help the tests succeed.

mandrasch:
I generally approve this, thanks @rawdreeg  👍 

@rfay: I only had 10 minutes today and couldn't figure out how to test this via Gitpod (https://gitpod.io/#https://github.com/drud/ddev/pull/3768) to confirm https://ddev.readthedocs.io/en/stable/users/cli-usage/#wordpress-quickstart is still working. Guess I'll need to read https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr in detail. 
Let me know if manual testing is required for approval, happy to check and verify it in the next days.

rfay:
I would love your manual approval, even though I think this PR grew out of the mariadb 10.3 bug in v1.19.1, which will be fixed in v1.19.2. 

If you'd like to download the artifacts from this PR, they're at https://github.com/drud/ddev/actions/runs/2258351791 - you can just get the ddev you want, put it in your $PATH, and test away with a Wordpress project. THANKS!

mandrasch:
> I would love your manual approval, even though I think this PR grew out of the mariadb 10.3 bug in v1.19.1, which will be fixed in v1.19.2.
> 
> If you'd like to download the artifacts from this PR, they're at https://github.com/drud/ddev/actions/runs/2258351791 - you can just get the ddev you want, put it in your $PATH, and test away with a Wordpress project. THANKS!

Oh wow, on mac it was way easier than I thought! Nice! Thanks for guidance!
(For anyone reading this, here are the steps: https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr)

mandrasch (approved):
Works great, thanks @rawdreeg! 

Tested the https://ddev.readthedocs.io/en/stable/users/cli-usage/#wordpress-quickstart with it.

Tables were created with 'utf8mb4_unicode_ci' which is correct (Official info here, WordPress tries to upgrade tables since v4.2 https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/)

Thanks very much for contributing! This should also prevent (edge case) errors when DDEV created WP sites are exported (as tarball) to other hosts which may have latin or other collations set as default.  Great that we are now compliant to the official wp-config-sample.php :+1:

image

rfay:
Thanks for reviewing! Of course this would have been fixed by the upstream fix with MariaDB 10.3 if you were using latest.

mandrasch:
> Thanks for reviewing! Of course this would have been fixed by the upstream fix with MariaDB 10.3 if you were using latest.

You're welcome, I learned a lot about PR, Gitpod and DDEV! 👍 

Just as side note: I think it generally adds to the robustness of DDEV. Because when a user starts a fresh WordPress project and later exports the site to a webspace host which has another collation default, this would have caused problems importing the WordPress created by DDEV. So it's good that we stick to the official https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php.

I even thought today that it is maybe a good thing if we copy all the contents and comments of https://github.com/WordPress/WordPress/blob/master/wp-config-sample.php as well, therefore users will know that it is just standard WordPress (with some lines added by DDEV). They would maybe feel more "at home".

rfay:
I absolutely agree that we should use the full config sample. We already do that for Drupal.

Issue #3741, opened 2022-03-31
ddev yarn --cwd XYZ unexpected EOF while looking for matching


HeineDeelstra:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Run a Diagnostic and Paste Link Here

https://gist.github.com/HeineDeelstra/a5e647aa35b2f5a56e2a8e21a8f8ae0a

### Current Behavior

ddev yarn with a cwd given via --cwd results in an error message "unexpected EOF while looking for matching `" instead of executing the yarn command.

For example:

$ ddev yarn --cwd frontend build
frontend: unexpected EOF while looking for matching `''
frontend: syntax error: unexpected end of file
Failed to execute command bash -ic yarn '--cwd frontend build': exit status 2

### Expected Behavior

I would expect the yarn command to run in the given cwd.

### Steps To Reproduce

In a drupal9 or php project with a dir projectroot/subdir run on the host:
ddev yarn --cwd subdir [command]

### Anything else?

I normally use /opt/homebrew/bin/bash on the host (GNU bash, version 5.1.16(1)-release (aarch64-apple-darwin21.1.0). This bash is also in PATH.

rfay:
Hmm, shouldn't do that. 

Note that `ddev exec yarn --cwd web install` works fine.

rfay:
I think `ddev yarn` should be moved from a host command to web command to make it behave better.

m4olivei:
I ran into this today as well. Working on a Drupal 10 core patch, wanting to compile post css for core, trying:

```
$ ddev yarn --cwd core install
core: unexpected EOF while looking for matching `''
core: syntax error: unexpected end of file
Failed to execute command bash -ic yarn '--cwd core install': exit status 2
```

As noted, the following work around works for me:

```
$ ddev exec yarn --cwd core install
```

rfay:
* The `yarn` shell command should really be a web command instead of host command, will make things simpler.
* I think we should add an instruction to the custom commands to say "do a mutagen sync after", and that would be a better workaround to the current situation.

rfay:
This was fixed by 
* https://github.com/drud/ddev/pull/3836
You can test it with `brew install --HEAD ddev`

PR #3769, opened 2022-04-21
Allow further sub commands for yarn

liayn:
The yarn host command now accepts further arguments for scripts,
e.g. ddev yarn gulp <task>

## The Problem/Issue/Bug:

#3741 

By changing the quotation of the `$@`

* Create a package.json with gulp and add a task.
* Try to run this task

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3769"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/yarn
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/yarn
@@ -4,5 +4,5 @@
 ## Usage: yarn [flags] [args]
 ## Example: "ddev yarn install" or "ddev yarn add learna" or "ddev yarn --cwd web/core add learna"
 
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
I think I would just rather move yarn from host command to web command, which would solve all of this I think. It was a mistake to make it a host command. 

There would be two steps to do that:
1. Add a stanza to app.FixObsolete() which will remove the old one
2. Add the new one, it's pretty easy to just base on `ddev php` or whatever

Are you interested in taking that on? You can just push new commits here if you'd like, or do a new PR (since it's a different thing).

liayn:
Fine with me. Will see what I can do - with a new PR then.

liayn:
Follow up in #3836

PR #3771, opened 2022-04-23
Clarify that only images from drud/ddev are deleted

weitzman:
See companion PR at #3770 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3771"></a>


rfay:
If it's not working it's a bug not docs fix. @mikebarkas tried to make this all work in https://github.com/drud/ddev/pull/3683 etc

rfay:
Looking at the code, it definitely only finds drud/ddev images. I guess there must have been a conversation I don't remember where we deferrred doing this.

rfay (approved):
Thanks!

PR #3779, opened 2022-04-27
Fix TYPO3 Git clone example

jonaseberle:
## The Problem/Issue/Bug:

The example missed the `ddev config` step.

This removes a sentence that was not really necessary. It add `ddev config` and finally a `ddev launch` to be in line with the other examples.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3779"></a>


rfay (approved):
Thanks!

PR #3785, opened 2022-04-29
Try out docker-compose 2.5.0

rfay:
## The Problem/Issue/Bug:

docker-compose 2.5.0 has been released, and it may solve some of the problems that have plagued previous versions. 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3785"></a>

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -81,7 +81,7 @@ var DockerComposeVersion = ""
 
 // This is var instead of const so it can be changed in test, but should not otherwise be touched.
 // Otherwise we can't test if the version on the machine is equal to version required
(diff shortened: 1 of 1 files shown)


rfay:
This time the problem is that relative paths (or at least short relative paths) are no longer supported for bind mounts. So `.:/mnt/ddev_config` and `../:/var/www/html` seem to fail.

The `../:/var/www/html` is probably essential for PhpStorm with WSL2.

rfay:
Opened
* https://github.com/docker/compose/issues/9428
for this issue. Will try again when it's fixed and released.

PR #3795, opened 2022-05-03
Composer should fail if patches fail

AronNovak:
## The Problem/Issue/Bug:

It's a great thing to get information on failing patches immediately.

Uses https://github.com/cweagans/composer-patches/blob/1.x/src/Patches.php#L285

Apply a broken patch on the codebase and see the exit code of `ddev composer install`

-

## Release/Deployment notes:

As it might be a breaking change for some, we might need to wire it into the config and make it adjustable.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3795"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1789,6 +1789,7 @@ func (app *DdevApp) DockerEnv() {
 		// Without COMPOSE_DOCKER_CLI_BUILD=0, docker-compose makes all kinds of mess
 		// of output. BUILDKIT_PROGRESS doesn't help either.
 		"COMPOSE_DOCKER_CLI_BUILD": "0",
+		"COMPOSER_EXIT_ON_PATCH_FAILURE":"1",
 		// The compose project name can no longer contain dots
 		// https://github.com/compose-spec/compose-go/pull/197
 		"COMPOSE_PROJECT_NAME":          "ddev-" + strings.Replace(app.Name, `.`, "", -1),
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1788,8 +1788,8 @@ func (app *DdevApp) DockerEnv() {
 	envVars := map[string]string{
 		// Without COMPOSE_DOCKER_CLI_BUILD=0, docker-compose makes all kinds of mess
 		// of output. BUILDKIT_PROGRESS doesn't help either.
-		"COMPOSE_DOCKER_CLI_BUILD": "0",
-		"COMPOSER_EXIT_ON_PATCH_FAILURE":"1",
+		"COMPOSE_DOCKER_CLI_BUILD":       "0",
+		"COMPOSER_EXIT_ON_PATCH_FAILURE": "1",
 		// The compose project name can no longer contain dots
 		// https://github.com/compose-spec/compose-go/pull/197
 		"COMPOSE_PROJECT_NAME":          "ddev-" + strings.Replace(app.Name, `.`, "", -1),


rfay:
I'll be interested to see what the community says about this. Seems to me like it should be the default!

longwave:
:+1: - @cweagans implied a long time ago that this should become the default in v2.0 of the composer-patches plugin, it's only disabled by default for backward compatibility with v1.0: https://github.com/cweagans/composer-patches/issues/40#issuecomment-209465274

ashabed:
I would also prefer the default to be for composer to fail when patches fail.

cweagans:
This is what 2.0 of the patches plug-in will do (someday when I get around to finishing it). +1 on making this the default.

gilbertsoft:
This would mean we silently change the default behavior which could be very confusing for some. Instead you can easily add this environment variable to your config.yaml which will only influence you and not every DDEV instance. Regarding the package this would also mean it's a breaking change, we should not introduce breaking changes in minor or dot releases at all.

cweagans:
I'd agree, but the behavior doesn't actually work correctly right now. If you try to run a composer command that installs patches and the patches don't actually apply, it'll do the wrong thing silently.

If you want a more constrained version of this, you could set this behavior as the default for composer commands run before a ddev project is up and running.

gilbertsoft:
> I'd agree, but the behavior doesn't actually work correctly right now. If you try to run a composer command that installs patches and the patches don't actually apply, it'll do the wrong thing silently.
> 
> If you want a more constrained version of this, you could set this behavior as the default for composer commands run before a ddev project is up and running.

I guess you didn't change it for v1 for the same reason I came up here with. Hopefully you find the time soon for the v2 release so this issue is solved at the upstream properly.

As said, everyone who like a clean behavior already now can add the the two lines to the config.yaml and the variable is set and no existing project breaks :)

```
web_environment:
- COMPOSER_EXIT_ON_PATCH_FAILURE=1
```

And I also fully agree this would be the better default behavior but for sure not with the risk of breaking existing projects.

rfay:
I'm not entirely sure that the existing behavior isn't what's broken, and it would be OK to fix it. I can't imagine anybody would ever want a silent patch application failure.

gilbertsoft:
> I'm not entirely sure that the existing behavior isn't what's broken, and it would be OK to fix it. I can't imagine anybody would ever want a silent patch application failure.

Yeah, it's broken. But chaning the behavior is a breaking change which could finally break existing setups. And it's not on DDEV to fix it but the upstream!

jonaseberle:
See also the second option to achieve that: [`extra.composer-exit-on-patch-failure: 1` in composer.json ](https://github.com/cweagans/composer-patches#error-handling.)

rfay:
As @gilbertsoft pointed out, anybody really concerned about this can `ddev config global --web-environment-add= COMPOSER_EXIT_ON_PATCH_FAILURE=1` and fix it for all their projects. (`--web-environment-add` is coming in v1.19.2)

So it seems reasonable to just save this for when we're moving from v1.19 patch releases to v1.20 work.

josefglatz:
Should fail per default! 🤗

rfay (approved):
Everybody seems fine with this, but maybe it can wait for merge until we stop doing minor releases and move onto new features for v1.20.

jonaseberle:
Ok, to make my point more clear: Basically I think that is "workaround creep". Also it changes default behaviour and everybody that cares should have added it to their `composer.json` anyways.
It's not a biggy but not "Everybody seems fine with this" ;)

AronNovak:
@jonaseberle 
Well, according to our experience, when I consider the day-to-day-work, what it really means: many developers are not aware that patches fail silently, so sometimes regressions happen because of that, and it takes much more time/money to address the problem if that's discovered at a later point of the development flow. It's a breaking change what's easy to undo, so if someone would like to keep buggy patches in `composer.json`, after scanning the release notes, it will be trivial to do that.

jonaseberle:
@AronNovak If you rely on that behaviour (I do), then why don't you [set it in composer.json](https://github.com/cweagans/composer-patches#error-handling)? That way it is well defined in all environments.

AronNovak:
@jonaseberle As we try to use DDEV to enforce best practices as much as possible, not at another level. If we get a foreign project, we can quickly put it in DDEV with our usual config and we are more or less comfortable to start the work. Your reasoning totally makes sense as well, it really depends on what do we expect from this tool.

jonaseberle:
Has somebody explored what happens if we set `COMPOSER_PATCHES_EXIT_ON_PATCH_FAILURE=1` and there is `extra.composer-exit-on-patch-failure: false` in `composer.json`?

AronNovak:
@jonaseberle 
https://github.com/cweagans/composer-patches/blob/1.x/src/Patches.php#L285
`getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure']);`
That answers this question, it seems.

jonaseberle:
Thanks. So in order to reset it I could use `COMPOSER_EXIT_ON_PATCH_FAILURE=0` (the value of `extra.composer-exit-on-patch-failure` is not evaluated).

I still think we should not change the behaviour implicitly in this certain environment. It is not helping really, just obfuscating a potential problem that you might encounter in your deployment systems or CI or other non-DDEV places.
When v2 of composer-patches is released this is obsolete anyways and then it is just dead legacy code.

AronNovak:
> When v2 of composer-patches is released this is obsolete anyways

Do you have a reference for this? In case, that's indeed a deal-breaker for this PR.

jonaseberle:
That's how I understood it from this comment by @longwave:
> +1 - @cweagans implied a long time ago that this should become the default in v2.0 of the composer-patches plugin, it's only disabled by default for backward compatibility with v1.0: [cweagans/composer-patches#40 (comment)](https://github.com/cweagans/composer-patches/issues/40#issuecomment-209465274)

cweagans:
Can confirm, this will be the default in composer patches 2.0.0.

It may be a _while_ before that comes out though.

rfay:
Rebased.

h1nds1ght:
My opinion is that people who are not aware that a patch has failed (composer shows a red error message that it failed) are also not aware that composer process explicitly fails and does not go on (and shows an additional more prominent red error message). Composer applies patches at the very end of composer install, actually a project still might work if composer exits on first failed patch vs. all patches were applied except of failed ones.

I would vote for "whatever is the default, let it be the default".

rfay:
Rebased

rfay:
Does everybody like this in the end? If so, please rebase and let's get it into v1.20.0

AronNovak:
Just rebased then.

PR #3807, opened 2022-05-05
[Docs] Improve docs for PR testing, add step for checking

mandrasch:
## The Problem/Issue/Bug:

When I tried to use ddev artifacts from a PR first, I wasn't sure which DDEV version was used and if the replacement worked on my laptop.

Also the link step should be at the end of all steps (in my humble opionion).

Add info about `ddev -v` which will give you an exact info whether it worked (or not ;-))

No, just docs

No, just docs

## Release/Deployment notes:
No, just docs

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3807"></a>


rfay (approved):
Thanks!

PR #3812, opened 2022-05-06
Add support for all `composer self-update` version formats

gilbertsoft:
## The Problem/Issue/Bug:

The preinstalled Composer in the prebuilt base image is very fast out of date and therefor the latest features of the self-update command are not available.

The PR https://github.com/drud/ddev/pull/3808 was the first step to solve long standing issues regarding the installed Composer version. But as we can see it goes not far enough e.g. Composer supports now also to select minor versions which currently fails because not the latest version is installed in the container. This can be tested my adding the `composer_version: "--2.2"` to the `config.yaml` which currently fails also on master.

Before updating to the specified version by the user the latest stable release is always installed. This makes sure latest features can be properly used like the newly introduced minor version which is available in the latest Composer releases on various branches.

In the same time also the predefined key words like `stable`, `preview` and `snapshot` are now supported so the user has now the same options like with manually run `composer self-update`. This makes the Composer version selection in DDEV more streamlined to Composer itself and also others tools like https://github.com/shivammathur/setup-php which also fully supports the various versions formats supported by Composer. So DDEV not longer uses an own strategy here but is more compatible with other tools which should it make easier for the users in the end.

The versioning scheme used before is still fully supported, so no breaking change for existing project but new options where needed.

- Add an invalid version to the `config.yaml` e.g. `invalid` and check if latest stable version is installed afterwards
- Add a minor version e.g. `2.2` and check if the latest version of this branch is installed afterwards
- Add a version key word like `snapshot` and check if the current master is properly installed

The existing test is enhanced to also test major and minor versions support. The version key words are also used in the iteration but currently I see no easy way to also properly test if the correct version was installed. Added a TODO therefor to enhance these tests too at some time.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3812"></a>


rfay:
> Composer supports now also to select minor versions which currently fails because not the latest version is installed in the container. This can be tested my adding the composer_version: "--2.2" to the config.yaml which currently fails also on master

`composer_version: "--2.2"` was never a supported option. You can use "2" or an explicit patch version.

gilbertsoft:
> > Composer supports now also to select minor versions which currently fails because not the latest version is installed in the container. This can be tested my adding the composer_version: "--2.2" to the config.yaml which currently fails also on master
> 
> `composer_version: "--2.2"` was never a supported option. You can use "2" or an explicit patch version.

True but that's a new feature of Composer since the last releases and does not work because in the container an old version is installed. If you use a custom Docker file like my example later it works.

```
ARG BASE_IMAGE
FROM $BASE_IMAGE

RUN export XDEBUG_MODE=off && composer self-update --stable
```

Afterwards you are able to set the version like explained!

pkg/ddevapp/config.go
```diff
@@ -874,26 +872,30 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg:
...
+		}
+
 		// Try composer self-update twice because of troubles with composer downloads
 		// breaking testing.
+		// First of all Composer is updated to latest stable release to ensure
+		// new options of the self-update command can be used properly e.g.
+		// selecting a branch instead of a major version only.
 		contents = contents + fmt.Sprintf(`
-RUN export XDEBUG_MODE=off && ( composer self-update %s || composer self-update %s || true )
+RUN export XDEBUG_MODE=off && ( composer self-update --stable || composer self-update --stable || true ) && ( composer self-update %s || composer self-update %s || true )
```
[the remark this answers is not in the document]
gilbertsoft: Yeah, absolutely!

pkg/ddevapp/config.go
```diff
@@ -874,26 +872,30 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg:
...
+		}
+
 		// Try composer self-update twice because of troubles with composer downloads
 		// breaking testing.
+		// First of all Composer is updated to latest stable release to ensure
+		// new options of the self-update command can be used properly e.g.
+		// selecting a branch instead of a major version only.
 		contents = contents + fmt.Sprintf(`
-RUN export XDEBUG_MODE=off && ( composer self-update %s || composer self-update %s || true )
+RUN export XDEBUG_MODE=off; composer self-update --stable || composer self-update --stable; composer self-update %s || composer self-update %s || true
```
rfay, line 898:
```suggestion
RUN export XDEBUG_MODE=off; composer self-update --stable || composer self-update --stable || true; composer self-update %s || composer self-update %s || true
```

We need the `|| true` here on the first update because errors cause failure, and this can fail when no internet or when getcomposer is misbehaving, etc.

rfay: Do we even need the initial `self-update --stable` here, which has some additional cost? I know that it would have been an issue before v1.19.2, but since v1.19.2 now ships with a composer version that has all these features... do we need to do this stuff?

docs/users/extend/config_yaml.md
```diff
@@ -20,9 +20,9 @@ the .ddev/config.yaml is the primary configuration for the project.
...
+| composer_version                 | version of Composer to use in web container and `ddev composer`                            | It defaults to Composer v2; you can set it to "" or "2" (default) for Composer v2 or "1" for Composer v1 to use the latest major version available at the time your container is built. It is also possible to select a minor version for example `2.2` which will install the latest release of that branch. Alternatively, an explicit Composer version may be specified, for example `1.0.22`. Finally, it is also possible to use one of the key words `stable`, `preview` or `snapshot` see Composer documentation. |
```
rfay, line 24: This stuff should also be added to template.go, where the comments for config.yaml are added.

rfay: It also needs to be added (at least a little) to the `--composer-version` help.

rfay (changes requested):
Thanks for this contribution, I know it will be helpful for power users. I had a few documentation comments (needs to go in template.go and in command help) and asked whether the initial self-update is really necessary now that ddev v1.19.2 has current composer. I know that they might add features in the future and that this could be a thing, but not sure we really have to do that to defend against something that is now in the past.

pkg/ddevapp/config.go
gilbertsoft: The idea behind this is to avoid similar issues in the future when Composer introduces new version features again. Having the latest stable version installed first makes sure we can use everything Composer supports at this time. And it's a one time thing on the first image creation. As we saw Docker build cache is very effective here and caches also between various images so I really don't think this has a big impact but helps to avoid issues.

docs/users/extend/config_yaml.md
gilbertsoft: Yeah, didn't think about these two, will add it...

pkg/ddevapp/config.go
gilbertsoft: Maybe we should do some tests again with the Docker build cache and the new `debug refresh` command to see how the behavior is at the moment. Needs ofc a new Composer release...

rfay:
Rebased. Your editor had rearranged a bunch of imports, which made rebasing messy - if you could turn off that feature it would be great, thanks!

rfay:
Now is the get this rebased and reviewed and in, if everybody likes it. Please rebase - you'll have to put the docs in a new place. If you can get your editor not to move imports around and not to change things in the docs that you're not fiddling with it will be great. Thanks!

docs/users/extend/config_yaml.md
gilbertsoft: Templates is changed now too.

rfay (approved):
Thanks for this. I manually tested, rebased, and added a minor commit to make the `ddev config --composer-version` help reflect the new options. 

I wonder if they're now going to have a 2.3 channel, now that 2.4 is about to be released. It doesn't show at this time when running 2.4.0-RC1.

eliashaeussler:
I think it's not specifically an issue with 2.3, but rather a general potentially wrong assumption that I can specify each minor version of Composer and will get the latest patch release. My above mentioned scenario neither works for 2.3, nor does it for 2.1, 2.0 and so on.

I'll create an issue to

- either add support for specifying Composer minor versions (which would be against @gilbertsoft's strategy of streamlining `composer_version` with officially available Composer channels)
- or correct the information provided in `config.yaml` about possible config values for the `composer_version` config, supporting only the officially supported Composer channels (as per https://github.com/composer/composer/blob/2.4.2/src/Composer/SelfUpdate/Versions.php#L31)

PR #3820, opened 2022-05-09
Add `--wide` flag to `ddev list`

jerodev:
## The Problem/Issue/Bug:
In our team, we use the command bellow in several hooks to check if the ddev container is running. 

    ddev list -A | grep '<projectname>' | grep 'running' --count

However, because the terminal is not wide enough, the `running` part is omitted from the first grep command and second grep always fails.

Add a `--wide` flag to the `ddev list` command to make sure the output is not truncated.

```bash
$ ddev list -A
┌──────────┬─────────┬─────────────────────┬─────────────────────────────────────────────────────────┬────────┐
│ NAME     │ TYPE    │ LOCATION            │ URL                                                     │ STATUS │
├──────────┼─────────┼─────────────────────┼─────────────────────────────────────────────────────────┼────────┤
│ d9simple │ drupal9 │ /workspace/d9simple │ https://8080-jerodev-ddev-843s3hl1910.ws-eu44.gitpod.io │ OK     │
├──────────┼─────────┼─────────────────────┼─────────────────────────────────────────────────────────┼────────┤
│ Router   │         │ stopped             │                                                         │        │
└──────────┴─────────┴─────────────────────┴─────────────────────────────────────────────────────────┴────────┘
 
$ ddev list -A | grep d9simple
│ d9simple │ drupal9 │ /workspace/d9simple │ https://8080-jerodev-ddev-843s3hl ≈

$ ddev list -Aw | grep d9simple
│ d9simple │ drupal9 │ /workspace/d9simple │ https://8080-jerodev-ddev-843s3hl1910.ws-eu44.gitpod.io │ OK     │
```

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3820"></a>

### DIFF
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -14,6 +14,9 @@ var activeOnly bool
 // continuousSleepTime is time to sleep between reads with --continuous
 var continuousSleepTime = 1
 
+// Allow the output to be wider than the terminal width
+var unlimitedOutputWidth bool
+
 // ListCmd represents the list command
 var ListCmd = &cobra.Command{
 	Use:   "list",
@@ -23,7 +26,7 @@ var ListCmd = &cobra.Command{
 ddev list --active-only
 ddev list -A`,
 	Run: func(cmd *cobra.Command, args []string) {
-		ddevapp.List(activeOnly, continuous, continuousSleepTime)
+		ddevapp.List(activeOnly, continuous, continuousSleepTime, !unlimitedOutputWidth)
 	},
 }
 
@@ -31,6 +34,7 @@ func init() {
 	ListCmd.Flags().BoolVarP(&activeOnly, "active-only", "A", false, "If set, only currently active projects will be displayed.")
 	ListCmd.Flags().BoolVarP(&continuous, "continuous", "", false, "If set, project information will be emitted until the command is stopped.")
 	ListCmd.Flags().IntVarP(&continuousSleepTime, "continuous-sleep-interval", "I", 1, "Time in seconds between ddev list --continuous output lists.")
+	ListCmd.Flags().BoolVarP(&unlimitedOutputWidth, "wide", "w", false, "Unlimited output width")
 
 	RootCmd.AddCommand(ListCmd)
 }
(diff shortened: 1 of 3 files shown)


jonaseberle:
Maybe you can elaborate a bit more on the use case? Do you need to check for some other command if a certain DDEV project is running? And then start DDEV? I don't really get it yet and we always like to hear use cases.

Output can change (the text is *currently* `OK` btw for running sites).
Gladly we have `ddev list --json-output` for usage in scripts so you could just do:

```sh
#! /bin/sh
project="example.com"
if ddev list --json-output --active-only \
    | jq --exit-status ".raw[] | select(.name == \"$project\")" &> /dev/null; then
  # $project is active
  ...
fi
```
`jq` is available on relevant platforms nowadays.

Here is another (less expressive) example:
```sh
#! /bin/sh
if ddev exec true
  # is active
  ...
fi
```


Related: The [config option `table-style`](https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/#global_configyaml-options).

rfay:
I'm really frustrated by truncation as well. When I hit it (especially with long project names) I `ddev config global --simple-formatting` - that might solve your problem as well @jerodev , but I don't object to this option and will take a look. Also keep wanting to return to the basic table setup, the underlying library has improved a little in the meantime.

jerodev:
@jonaseberle I attempted to use `jq`, but it seems neither Ubuntu nor Centos has this by default.
Using `ddev exec true` is possible, but then the working directory has to be in the project folder.

If not for this specific use case, I think this flag can also be useful for other commands.

rfay (changes requested):
This is great, and will help me quite a lot. I hope there are better options in the future.

But could you please make it not wrap the "location" (filepath) when --wide is on? 

Would dearly love it if you became an expert on https://github.com/jedib0t/go-pretty and could make these things better yet! (how about describe having --wide also?)

rfay:
Another thing to consider is whether truncation should just be turned off. I'm not sure it's ever had a benefit. It sure is ugly when it wraps and you're not expecting it to... but completely useless when you need the info.

rfay:
Please take a look at #3865 - I'd like to replace this one with that one if you're good with it.

jerodev:
> Please take a look at #3865 - I'd like to replace this one with that one if you're good with it.

Perfect, I'll take a look at the new PR later and see if I can help. :)

rfay:
I'm going to close this for now, but it's worth considering coming back to the flag approach used here, it would probably make me happy either way. Truncation never ever helps me, and I often have long project names and URLs that I want to copy (in tests).

PR #3836, opened 2022-05-12
Move yarn command from host to web (#3741)

liayn:
## The Problem/Issue/Bug:

https://github.com/drud/ddev/issues/3741

Executes the command right inside the web container so no further CLI parameter tweaking is necessary.

`yarn gulp <someTaskName>`

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3836"></a>

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -579,6 +579,17 @@ func (app *DdevApp) FixObsolete() {
 			}
 		}
 	}
+
+	// Remove old global commands
+	for _, command := range []string{"host/yarn"} {
+		cmdPath := filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/", command)
+		if _, err := os.Stat(cmdPath); err == nil {
+			err1 := os.Remove(cmdPath)
+			if err1 != nil {
+				util.Warning("attempted to remove %s but failed, you may want to remove it manually: %v", cmdPath, err)
+			}
+		}
+	}
 }
 
 type composeYAMLVars struct {
--- a/docs/users/performance.md
+++ b/docs/users/performance.md
@@ -65,9 +65,11 @@ In general, it's best practice on most projects to do significant git operations
 ddev mutagen sync || true
 ```
 
-### Syncing after yarn actions
+### Syncing after yarn, npm, pnpm actions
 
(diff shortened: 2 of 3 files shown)


pkg/ddevapp/config.go
```diff
@@ -569,7 +569,7 @@ func (app *DdevApp) CheckDeprecations() {
 func (app *DdevApp) FixObsolete() {
 
 	// Remove old in-project commands (which have been moved to global)
-	for _, command := range []string{"db/mysql", "host/launch", "web/xdebug"} {
+	for _, command := range []string{"db/mysql", "host/launch", "web/xdebug", "host/yarn"} {
```
rfay, line 572: This isn't exactly correct - this is removing from the project's .ddev, and we need to remove from global. So we need another stanza like this for global.

rfay (changes requested):
Thanks so much for taking this up! Has a little farther to go.

pkg/ddevapp/config.go
liayn: Isn't the global commands stuff rewritten each time anyways?

rfay: Yes, but no removal happens.

liayn: I guess you need to help me out then, because I obviously misunderstand what you mean.
My experience is that the whole `.global_commands` folder is eliminated every time I `ddev start`
(at least the handles are gone if I have wsl open in one of those subfolders in the meantime)

rfay: Yeah, .global_commands is not where global commands live, they're just copied in there on start/etc. They live in ~/.ddev/commands

I'm happy to add the stanza though. Have to get the global config dir and do the work there.

liayn: Oh, I didn't know that. Thanks for explaining. I can add the go stuff then.

rfay:
Note that the DdevFullSiteSetup tests can't succeed right now due to 
* #3837

gilbertsoft:
Thought we need this workaround because of mutagen users?

rfay (approved):
Manually tested and it does exactly what's needed, as long as `ddev yarn` is now working as you want it to from the original issue. Thanks!

The one thing I'd appreciate is if you'd add a minor edit to https://github.com/drud/ddev/blob/master/docs/users/performance.md#syncing-after-yarn-actions - Obviously the command no longer mitigates. So change it to "Syncing after yarn, npm, pnpm actions" and just give the good advice to `ddev mutagen sync` after those, or wait for them.

Issue #3839, opened 2022-05-16
Colima/Lima 9p mounting support


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Colima has released support for Lima with 9p mounting instead of sshfs. This should be a win for reliability of in-container mounts. However, 9p doesn't support symlinks on the host... and it certainly doesn't work yet with ddev.

* Postgres doesn't work:
```
postgres: could not access the server configuration file "/etc/postgresql/postgresql.conf": No such file or directory
```
* MariaDB import-db doesn't work:
```
$ ddev import-db --src=.tarballs/db.sql.gz
pv: /mnt/ddev_config/.importdb3908252741/*.*sql: No such file or directory
Failed to import database db for d7: failed to import database: exit status 10
stdout:
stderr:
```

### Expected Behavior

Would like it to work

rfay:
I find that colima 9p is working OK with colima v0.4.1 and limactl 0.10.0, not sure what was happening earlier.

Unfortunately 9p mounting is a lot slower than the sshfs default mounting. But it seems to work fine with mutagen.

PR #3849, opened 2022-05-18
Run a set of tests with colima 9p mounting, for #3839

rfay:
## The Problem/Issue/Bug:

* #3839 

9p is working OK in manual testing, but this will try to run a full set of tests

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3849"></a>


rfay:
This was successful, but slower than sshfs, so not going to start using it for test right now.

PR #3858, opened 2022-05-20
Web container depends on db container, fixes #3702

### LINKED ISSUE #3702: Web container should depend on db container readiness unless omit_containers[db]

I have a supervisor configuration that executes a Symfony message consumer when starting the project. The boot process tends to fail in case the `db` container is not ready yet.

The message related to this issue found in the logs is as follows:
```
An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refused
```

The following configuration stored in `.ddev/docker-compose.overwrite.yaml` workarounds the issue:
```
services:
  web:
    depends_on:
      - "db"
```

This could be done automatically by ddev, unless `omit_containers[db]` is set.

_No response_

### Additional context

_No response_

mikebarkas:
## The Problem/Issue/Bug:
* #3702 

Web container should depend on db container readiness unless omit_containers

This adds a `depends_on` for the web container if no `omit db`

Start a ddev in a new project, look in the `.ddev-docker-compose-full.yaml `file

No tests added

## Related Issue Link(s):
#3702 

## Release/Deployment notes:

N/A

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3858"></a>

### DIFF
--- a/cmd/ddev/cmd/composer_test.go
+++ b/cmd/ddev/cmd/composer_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"github.com/drud/ddev/pkg/nodeps"
 	"github.com/stretchr/testify/require"
 	"os"
 	"path/filepath"
@@ -17,6 +18,13 @@ import (
 func TestComposerCmd(t *testing.T) {
 	if dockerutil.IsColima() {
 		t.Skip("Skipping test on Colima because of odd unable to delete vendor error")
+	}
+	// 2022-05-24: I've spent lots of time debugging intermittent `composer create` failures when NFS
+	// is enabled, both on macOS and Windows. As far as I can tell, it only happens in this test, I've
+	// never recreated manually. I do see https://github.com/composer/composer/issues/9627 which seemed
+	// to deal with similar issues in vagrant context, and has a hack now embedded into composer.
+	if nodeps.NFSMountEnabledDefault {
+		t.Skip("Composer has strange behavior in NFS context, so skipping")
 	}
 	assert := asrt.New(t)
 
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -144,7 +144,8 @@ services:
 
     {{ if not .OmitDB }}
     depends_on:
-      - db
+      - db:
+          condition: service_healthy
     {{ end }}
 
     ports:
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -144,8 +144,8 @@ services:
 
     {{ if not .OmitDB }}
     depends_on:
-      - db:
-          condition: service_healthy
+      db:
+        condition: service_healthy
     {{ end }}
 
     ports:
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -142,6 +142,11 @@ services:
     user: '$DDEV_UID:$DDEV_GID'
     hostname: {{ .Name }}-web
 
+    {{ if not .OmitDB }}
+    depends_on:
+      - db
+    {{ end }}
+
     ports:
       - "{{ .DockerIP }}:$DDEV_HOST_WEBSERVER_PORT:80"
       - "{{ .DockerIP }}:$DDEV_HOST_HTTPS_PORT:443"


rfay:
I'm trying to figure out a manual test that will show this well. A huge DB might do it, may be able to see how long it takes for db container to get ready in that case, and keep a curl going in the db container.

rfay:
I think this is causing macos-nfs to fail (more),  https://buildkite.com/drud/ddev-macos-use-nfsmount/builds/5108#bbcfd70f-75f9-474e-9604-dc494f7b0946 

I've seen this plenty of times with NFS, never understood it. Essentially, the composer.json doesn't get there in time to be used, which is probably an NFS failure. But we're seeing it here on 3 different test runners. 

Line 90 is this:
```go
		args = []string{"composer", "create", "--prefer-dist", "--no-dev", "--no-install", "psr/log:1.1.0"}
		out, err = exec.RunHostCommand(DdevBin, args...)
		assert.NoError(err, "failed to run %v: err=%v, output=\n=====\n%s\n=====\n", args, err, out)
```

It's got --no-install even, so it's not trying to do a lot. 

It will be interesting to see if it's recreatable in local with NFS. And of course this may just be random behavior, because it's been seen before. It's possible that slightly slower startup could affect this, or that the exec that's done doesn't respect the startup, etc.

mikebarkas:
Perhaps the complexity this is introducing isn't worth the change?
Maybe a better approach is, recommend someone to add an override compose file as needed for their project?

rfay:
I don't consider this very complex. Just want to understand what might be happening. Would be lovely to understand why the composer.json doesn't exist when it needs to, since that's been an intermittent problem for years, and especially (only?) with NFS.  Since this one seem to fail every single time with macOS/NFS it may be a lovely situation to learn from. I'll spend a little time with it, and of course you're welcome to also. And I'm still in search of manual testing that will show the scope of this change and the impact on real users.

rfay:
To test the behavior, I created an alternate healthcheck.sh for the database container (mysql/mariadb) that waits 15s before declaring the db container healthy.  I put that in `.ddev/db-build`
Then I added a `.ddev/db-build/Dockerfile` with
```dockerfile
ARG BASE_IMAGE
FROM $BASE_IMAGE
COPY healthcheck.sh /
RUN chmod +x /healthcheck.sh
```

Then... `export PROJ=d9simple; while true; do date && echo "===== Web container" && docker inspect --format "{{json .State.Health.Log }}" ddev-$PROJ-web |jq -r last && echo "==== DB container" && docker inspect --format "{{json .State.Health.Log }}" ddev-$PROJ-db |jq -r last && sleep 1; done`

Do the `ddev start` in another windows.

That shows the health status of both containers, but you'll see that the web container has in fact been started and becomes healthy before the db container does. However, if you wait until `ddev start` completes to do things... it is behaving correctly waiting for both containers to be up. But I'm pretty sure that's true without this change.

rfay (changes requested):
I don't think this accomplishes what we want yet, see my testing example. Looks like in https://github.com/compose-spec/compose-spec/blob/master/spec.md#depends_on we need to use the long form, so 
```yaml
    depends_on:
      db:
        condition: service_healthy
```

rfay:
@mikebarkas don't worry about the NFS test failure right now. It's lovely that it's predictable, I can reproduce it, have no idea why it would be impacted by the depends_on here. 

I've been experimenting with the various things in 
* https://github.com/composer/composer/issues/9627
where composer added an environment variable `COMPOSER_RUNTIME_ENV=virtualbox` to deal with things just like this, doing a sleep to let the filesystem catch up. But I haven't been able to solve the problem using either that or the suggested dummy unzip command in that issue.

rfay:
Always try a `ddev start && ddev launch`  :)

The syntax is yaml object syntax, not array syntax, 

```
    depends_on:
      db:
        condition: service_healthy
```

rfay:
We've all done it! :)

rfay:
It's wonderful how consistent that with-NFS failure is, but I'm completely baffled how it relates to this change, and although I've chased a lot of angles with it here I haven't found anything that affects it. And since I can't recreate it with manual testing (which works fine) I'm a bit at a loss.

rfay:
And now, unlike last time, we have the test failure on Windows NFS as well :( How in the world could this be affected by this PR... ??? :)

rfay (approved):
This seems to work perfectly now, tested on Windows WSL2 with Ubuntu 22.04. Using the (edited recipe above - forgot the healthcheck link before) I see this:

```json
Mon May 23 12:20:54 PM MDT 2022
===== Web container
Template parsing error: template: :1:13: executing "" at <.State.Health.Log>: map has no entry for key "Health"
==== DB container
null
Mon May 23 12:20:56 PM MDT 2022
===== Web container
Template parsing error: template: :1:13: executing "" at <.State.Health.Log>: map has no entry for key "Health"
==== DB container
null
Mon May 23 12:20:57 PM MDT 2022
===== Web container
null
==== DB container
{
  "Start": "2022-05-23T12:20:42.002494-06:00",
  "End": "2022-05-23T12:20:57.1979635-06:00",
  "ExitCode": 0,
  "Output": "waiting a little bit before declaring healthyhealthy"
}
Mon May 23 12:20:59 PM MDT 2022
===== Web container
{
  "Start": "2022-05-23T12:20:59.218305-06:00",
  "End": "2022-05-23T12:20:59.3281947-06:00",
  "ExitCode": 1,
  "Output": "phpstatus: FAILED /var/www/html: OK mailhog: FAILED "
}
==== DB container
{
  "Start": "2022-05-23T12:20:42.002494-06:00",
  "End": "2022-05-23T12:20:57.1979635-06:00",
  "ExitCode": 0,
  "Output": "waiting a little bit before declaring healthyhealthy"
}
Mon May 23 12:21:01 PM MDT 2022
===== Web container
{
  "Start": "2022-05-23T12:21:01.524198-06:00",
  "End": "2022-05-23T12:21:01.6891125-06:00",
  "ExitCode": 0,
  "Output": "phpstatus: OK /var/www/html: OK mailhog: OK "
}
==== DB container
{
  "Start": "2022-05-23T12:20:42.002494-06:00",
  "End": "2022-05-23T12:20:57.1979635-06:00",
  "ExitCode": 0,
  "Output": "waiting a little bit before declaring healthyhealthy"
}
Mon May 23 12:21:03 PM MDT 2022
===== Web container
{
  "Start": "2022-05-23T12:21:01.524198-06:00",
  "End": "2022-05-23T12:21:01.6891125-06:00",
  "ExitCode": 0,
  "Output": "phpstatus: OK /var/www/html: OK mailhog: OK "
}
==== DB container
{
  "Start": "2022-05-23T12:20:42.002494-06:00",
  "End": "2022-05-23T12:20:57.1979635-06:00",
  "ExitCode": 0,
  "Output": "waiting a little bit before declaring healthyhealthy"
}
```

The web container doesn't start to come up until the db container is fully ready.

`ddev start` completes when both are ready.

Will continue trying to figure out the NFS situation. Thanks for this!

rfay:
Thanks, I'm not sure it's actually changing "performance" as opposed to giving a realistic view of readiness, but I'm also interested in what you're seeing. Your result would seem to indicate that the db container is taking a really long time to come up, or that the go code that waits for it is doing something wrong. 

Could you check the results of this with a plain vanilla project (`ddev debug test` for example) and see what you see. 

I'm actually not used to `ddev start` taking anywhere near as long as you're seeing. You're on Linux right? I would expect 10-30s.

Oh, you're doing 10 restarts, so that's reasonable. I thought you were doing one restart.

jonaseberle:
`restart` overemphasizes the effect by also showing docker-compose safely tearing down the project without breaking the constraints. We lose some parallel execution that docker-compose can take advantage of if no constraints are there.

```
ddev poweroff
time for i in $(seq 5); do ddev debug test; done
```
I see no real difference:
main: 1m57s
this PR: 2m2s

@rfay are you sure we are testing the right thing here? Looking into the `ddev debug test` container does not show the `depends_on`. My test project shows it.

```
# with this PR:
ddev debug test
debugContainer=$(docker ps -q --filter=name='ddev-tryddevproject-.*-db')
docker inspect $debugContainer \
  | jq -r '.[].Config.Labels'

mkdir /tmp/a && cd /tmp/a
ddev config --auto
ddev start
# outputs:   "com.docker.compose.depends_on": "db:service_healthy",
docker inspect ddev-a-web \
  | jq -r '.[].Config.Labels'
ddev delete -Oy
```

jonaseberle:
I spotted my error ;) `--filter=name='ddev-tryddevproject-.*-db'` -> I checked the `db` container ...

Sorry for the fuss.

jonaseberle (approved):

rfay:
Rebased and excluded TestCmdComposer when running NFS-based tests. I've always thought this had to be a composer/filesystem bug, and after following the linked issues I think so more... but I can't sort it out, and I can't get composer to output enough verbosity to help me.

rfay:
Thanks for this!

rfay:
Unfortunately, this leads to some unfortunate behavior from docker-compose, see 
* https://github.com/docker/compose/issues/9732

The result is that docker-compose can just hang if the db container exits early. 

I think I'll have to revert at least part of this PR.

PR #3859, opened 2022-05-20
Dont' use compose project name with dots, add dot test

rfay:
## The Problem/Issue/Bug:

With docker-compose 2.5.1...

* https://github.com/drud/ddev/pull/3851#issuecomment-1132745998 (discussion)
* https://github.com/docker/compose/issues/8881
* https://github.com/compose-spec/compose-go/pull/197

Bottom line is that `ddev ssh` and `ddev exec` (possibly other things) don't work if COMPOSE_PROJECT_NAME has a dot in it. You get:

```
$ ddev ssh
service "web" is not running container #1
Failed to ddev ssh web: exit status 1
```

I imagine since ddev was using COMPOSE_PROJECT_NAME initially with `docker-compose config`, which inserts `name` into the .ddev-docker-compose-full.yaml (without dots) and then uses that .ddev-docker-compose-full again with COMPOSE_PROJECT_NAME set (and with a dot in it), that it triggers an obscure edge case.  It's easier to remove the dot here than to chase it in compose.

* Remove the dot
* Add a test that uses project names with dots in them.

* https://github.com/drud/ddev/pull/3851#issuecomment-1132745998

* Added TestCmdSSH

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3859"></a>

### DIFF
--- a/cmd/ddev/cmd/ssh_test.go
+++ b/cmd/ddev/cmd/ssh_test.go
@@ -1,0 +1,66 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/drud/ddev/pkg/ddevapp"
+	"github.com/drud/ddev/pkg/exec"
+	"github.com/drud/ddev/pkg/fileutil"
+	"github.com/drud/ddev/pkg/testcommon"
+	"github.com/drud/ddev/pkg/util"
+	asrt "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"os"
+	"testing"
+)
+
+// TestCmdSSH runs `ddev ssh` on basic apps, including with a dot and a dash in them
+func TestCmdSSH(t *testing.T) {
+	assert := asrt.New(t)
+
+	// Create a temporary directory and change to it for the duration of this test.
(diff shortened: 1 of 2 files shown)


jonaseberle:
I tested and this works on docker. Are the colima failures real?
```
$ ddev version
 ITEM             VALUE                                   
 DDEV version     v1.19.2-25-g6bee2020                    
 architecture     amd64                                   
 db               drud/ddev-dbserver-mariadb-10.3:v1.19.2 
 dba              phpmyadmin:5                            
 ddev-ssh-agent   drud/ddev-ssh-agent:v1.19.0             
 docker           20.10.15                                
 docker-compose   v2.5.1                                  
 docker-platform  e480                                    
 mutagen          0.14.0                                  
 os               linux                                   
 router           drud/ddev-router:v1.19.0                
 web              drud/ddev-webserver:v1.19.2  
```

Thank you so much for the added tests!

rfay:
Thanks so much as always.

Colima has a perpetual problem with port bindings. Doesn't affect ordinary users very much, but breaks the tests at least half of the time and I have to restart. 
* https://github.com/abiosoft/colima/issues/189

PR #3861, opened 2022-05-22
ddev-router should only output one upstream per container port

rfay:
## The Problem/Issue/Bug:

@deviantintegral pointed out in Discord that ddev-router is creating two upstream hosts for every port, resulting in inappropriate round-robin behavior of the reverse-proxy nginx inside ddev-router. https://discord.com/channels/664580571770388500/976181828341866526

More discussion in the PR that probably introduced this behavior inadvertently:

* https://github.com/drud/ddev/pull/3403#issuecomment-1132892462

Upstreams in ddev router (shown by `ddev debug router-nginx-config`) show both of the IP addresses (for the two networks that the container belongs to), resulting in inappropriate round-robin behavior, as if there were actually two different containers out there:

```
        upstream d7.ddev.site-80 {
        keepalive 100;
            # Container=ddev-d7-web 172.21.0.5:80
            server ddev-d7-web:80;
            # Container=ddev-d7-web 172.19.0.8:80
            server ddev-d7-web:80;
        }
```

Instead, with this PR, the upstream definitions show only the ddev_default network IP address:
```
        upstream d7.ddev.site-80 {
        keepalive 100;
            # Container=ddev-d7-web 172.19.0.8:80
            server ddev-d7-web:80;
        }
```

Only evaluate the network named ddev_default

`ddev poweroff && ddev start`
Review `ddev debug router-nginx-config`

I think this is OK as long as it passes existing tests. 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3861"></a>

### DIFF
--- a/containers/ddev-router/nginx.tmpl
+++ b/containers/ddev-router/nginx.tmpl
@@ -124,7 +124,7 @@ server {
 
         {{/* jonaseberle/github-action-setup-ddev version */}}
         {{ range $containerNetwork := $container.Networks }}
-            {{ if (and (ne $containerNetwork.Name "ingress") (or (eq $containerNetwork.Name $containerNetwork.Name) (eq $containerNetwork.Name "host"))) }}
+            {{ if eq $containerNetwork.Name "ddev_default" }}
                 {{/* If only 1 port exposed, use that */}}
                 {{ if eq $addrLen 1 }}
                     {{ $address := index $container.Addresses 0 }}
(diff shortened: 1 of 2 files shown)


containers/ddev-router/nginx.tmpl
```diff
@@ -124,7 +124,7 @@ server {
 
         {{/* jonaseberle/github-action-setup-ddev version */}}
         {{ range $containerNetwork := $container.Networks }}
-            {{ if (and (ne $containerNetwork.Name "ingress") (or (eq $containerNetwork.Name $containerNetwork.Name) (eq $containerNetwork.Name "host"))) }}
+            {{ if eq $containerNetwork.Name "ddev_default" }}
```
deviantintegral, line 127: I like this much simpler conditional 👏

deviantintegral (approved):
I've confirmed this fixes the issue I had with noVNC, and that only a single upstream server is emitted in the nginx config. Nice, and thank you!

jonaseberle (approved):
Nice catch!

Issue #2839, opened 2021-02-26
app.Exec() doesn't properly split stdout and stderr, can result in data corruption (ddev drush sql-dump)


rfay:
**Describe the bug**

[app.Exec()](https://github.com/drud/ddev/blob/8cf313e106558ac1034ce37a8d9124eb5576dfc2/pkg/ddevapp/ddevapp.go#L1058) doesn't properly return separate stdout and stderr streams, but instead bundles them together. 

```
$ ddev exec junk 2>/dev/null
bash: junk: command not found
```

This seems to be a result of `docker exec` doing the same thing:

```
$ docker exec -it ddev-d9-web bash -c junk 2>/dev/null
bash: junk: command not found
```

The result is that a command that happens to return valid stdout but also some stderr will return them all together in the stdout. 

I noticed this today using `drush sql-dump` with mysql 5.7.33.  mysql 5.7.31+ has a new security feature that breaks drush sql-dump, see https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-31.html#mysqld-5-7-31-security and https://github.com/drush-ops/drush/issues/4489 ([drupal slack discussion regarding impact on Acquia hosting](https://drupal.slack.com/archives/C2HRNCHV4/p16142849300138000

Anyway, the result with our using of app.Exec() is that `ddev drush sql-dump >/tmp/db.sql` using current mysql 5.7 (and 8.0) will result in a mysql error being embedded in the dump, which isn't correct. 

Since this seems to have origins in docker-world, there may not be a good fix.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version and configuration information (please complete the following information):**

- Host computer OS and Version: [e.g. Windows 10, macOS Catalina]
- Docker Desktop version if on macOS or Windows (from "About Docker Desktop")
- ddev version information (use `ddev version`)
- config.yaml contents for the misbehaving project
- Do you have any custom configuration (nginx, php, mysql) in the .ddev folder? If so, have you tried without them?

**Additional context**
Add any other context about the problem here. Thanks!

rfay:
Dependent on either
* https://github.com/docker/compose/issues/8098
or
* Just using docker exec (via api) instead of using `docker-compose exec`. This may be a better choice.

Will wait to address it until that one comes through.

rfay:
* #3869 
deals with this problem for the Acquia integration.

GuySartorelli:
[...]

PR #3863, opened 2022-05-23
Make sure that errors and warnings are written to stderr

nico-loeber:
## The Problem/Issue/Bug:

Warnings and errors are not written to stderr consistently.

Apart from a general unpredictability, warnings can disturb automatically evaluated output.

To make the output more predictable all warnings and errors, I was able to find, are now written to stderr.

Executing any ddev commands on the command line should still show exactly the same result.

To test if warnings are written to stderr you can run the command outputting the error / warning to show stdout only by adding `2>/dev/null`. 
For example `ddev describe 2>/dev/null` should now show a clean output. 

N/A

## Related Issue Link(s):
This will help to improve stability when using the [DDEV Integration plugin](https://github.com/php-perfect/ddev-intellij-plugin).

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3863"></a>

### DIFF
--- a/cmd/ddev/cmd/hostname.go
+++ b/cmd/ddev/cmd/hostname.go
@@ -52,7 +52,7 @@ to allow ddev to modify your hosts file. If you are connected to the internet an
 		// If requested, remove all inactive host names and exit
 		if removeInactive {
 			if len(args) > 0 {
-				output.UserOut.Fatal("Invalid arguments supplied. 'ddev hostname --remove-all' accepts no arguments.")
+				util.Failed("Invalid arguments supplied. 'ddev hostname --remove-all' accepts no arguments.")
 			}
 
 			util.Warning("Attempting to remove inactive hostnames which use TLD %s", nodeps.DdevDefaultTLD)
@@ -63,7 +63,7 @@ to allow ddev to modify your hosts file. If you are connected to the internet an
 
 		// If operating on one host name, two arguments are required
 		if len(args) != 2 {
-			output.UserOut.Fatal("Invalid arguments supplied. Please use 'ddev hostname [hostname] [ip]'")
+			util.Failed("Invalid arguments supplied. Please use 'ddev hostname [hostname] [ip]'")
 		}
 
 		hostname, ip := args[0], args[1]
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -161,7 +161,7 @@ func init() {
 	// Populate custom/script commands so they're visible
 	// We really don't want ~/.ddev or .ddev/homeadditions or .ddev/.globalcommands to have root ownership, breaks things.
 	if os.Geteuid() == 0 {
-		output.UserOut.Warning("Not populating custom commands or hostadditions because running with root privileges")
+		util.Warning("Not populating custom commands or hostadditions because running with root privileges")
 	} else {
 		err := ddevapp.PopulateExamplesCommandsHomeadditions("")
 		if err != nil {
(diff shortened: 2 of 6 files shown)


rfay:
Wouldn't it be better to fix output.UserOut.Fatal() and output.UserOut.Error() and warning?

Also, note 
* https://github.com/drud/ddev/issues/2839

nico-loeber:
> Wouldn't it be better to fix output.UserOut.Fatal() and output.UserOut.Error() and warning?

You mean using `output.UserErr.Fatal()` instead? 
I think there should be a consistent way for writing output.
I thought there is a reason why someone created the util.* wrapper functions, they also seem to do some coloring, so I prefered that one.

> Also, note
> 
> * [app.Exec() doesn't properly split stdout and stderr, can result in data corruption (ddev drush sql-dump) #2839](https://github.com/drud/ddev/issues/2839)

That doesn't sound good... I had no issues when manually testing with the "Instrumentation is opted in" warning for example.

rfay:
Hmm... util.Warning and friends are just wrappers on the other... https://github.com/drud/ddev/blob/0aa4e1d27cd77d73c06b026a0ff0629dac61603d/pkg/util/utils.go#L43-L50

So not sure what this would accomplish.

And the reason that we're using the output wrapper is explicitly so that json consumers can be notified... using json.

nico-loeber:
We can use whatever function call you like, but there should be some rule like:

> Whenever there is a warning use `util.Warning()` to output it.

I would advise against using `output.UserErr.Warn` without a wrapper directly in your buisiness logic, so you segerate your code from logrus logger, as it is a 3rd party dependency.
Using a wrapper makes it easier to react on changes of the dependency or add functionality like verbosity or a quite mode later on.

rfay:
The util.Warning stuff predated the logrus adoption (which was done to support json output). So that's why you'll see a mix. Still interested in why this PR satisfies the needs of the plugin... I don't see a problem in it, but not sure I see the benefit either. The one big problem is using util.Failed() when it shouldn't be used, because it immediately terminates the app, which shouldn't usually be done during internal logic (like in test infrastructure).

nico-loeber:
> Still interested in why this PR satisfies the needs of the plugin...

It makes it easyer to distinguish between relevant content and warnings / errors / infos that should be reported to the user executing it.
While stdout is expected to contain the acutal result of the command, warnings are expected to be in stderr.
Currently DDEV does write warnings sometimes to stderr, and sometimes to stdout which makes error handling hard to handle, without redirecting stderr to stdout which you obviously don't want to do.

This does not solve an actual problem, it is more an improvement to be more compliant to the unix output standards and will be neccessary for another pull request I am planning.

> The one big problem is using util.Failed() when it shouldn't be used, because it immediately terminates the app, which shouldn't usually be done during internal logic (like in test infrastructure).

For context: util.Failed is a wrapper for logrus fatal

```golang
func (logger *Logger) Fatal(args ...interface{}) {
	logger.Log(FatalLevel, args...)
	logger.Exit(1)
}
```

Yes, I also noticed this exit and dind't like it much, because it is a bit unexpected and can make things hard to test.
I didn't change aynthing regarding this behavior as this pull request just aims to tidy up the error streams.

nico-loeber:
Another comment on the wrapper topic:
I am not saying that we should use `util.` I am just suggesting any wrapper. Something like `logger.warning()` would allow you to have one single point to change the logging behavior later on. For example if you like to swtich the warnings back to stdout (please don't do ;) ) you only have to edit one line in `logger.warning()` instead of editing all places where a warning is caused.

Just tell me which wrapper to use or if I should create a new one. I will adjust my pull request accordingly.

cmd/ddev/cmd/hostname.go
```diff
@@ -52,7 +52,7 @@ to allow ddev to modify your hosts file. If you are connected to the internet an
 		// If requested, remove all inactive host names and exit
 		if removeInactive {
 			if len(args) > 0 {
-				output.UserOut.Fatal("Invalid arguments supplied. 'ddev hostname --remove-all' accepts no arguments.")
+				util.Failed("Invalid arguments supplied. 'ddev hostname --remove-all' accepts no arguments.")
```
rfay, line 55: When I first looked at this I thought you were changing output.UserErr.Fatal() to util.Failed() and they looked the same :) That's why I was confused in the comments. But you're properly moving from output.UserOut.Fatal to (essentially) output.UserErr.Fatal. Which makes plenty of sense.

rfay:
Sometimes you just have to keep the conversation going long enough until the other person understands! 😄

nico-loeber:
Haha, yes, if you have any further concerns, I'm happy to clarify misunderstandings! ;)

rfay:
Thanks, and thanks for the patience!

PR #3864, opened 2022-05-24
Use logrus json formatter to ensure valid json output per line

nico-loeber:
## The Problem/Issue/Bug:

As discussed with @rfay, there is a problem parsing the `--json-output` of commands, as the JSON is not valid per line or as a whole if there is more than one log output at a time.

This appears to be caused by a custom implementation of the json_formatter class, as it attempts to pretty print json when a terminal is detected. The pretty printed json contains line breaks that break its validity on a per-line basis.

The json_formatter implementation of logrus ensures that every line is valid at any time.

As json is a data exchange format meant for machines and not humans there is no need to pretty print this output. In case pretty printed output is needed for debugging purposes one can still use `ddev describe -j | jq`.

I could not find any other deviating behavior of the json_formatter compared to the logrus implementation.
As the custom implementation is no longer used I removed it.

Run any ddev command with `-j` flag to enable json output.
To get more than one line of output  `export DDEV_DEBUG=true` can help.
Each outputed log line `{"msg": "", "time": "" ...}` should be printed stand alone in one line.
Each line should pass a json validity check.
Use `jq` to pretty print the json output and to check it's contents. 
No contents should be changed by this pull request.
 
N/A

## Related Issue Link(s):
- https://github.com/php-perfect/ddev-intellij-plugin/issues/39
- https://github.com/php-perfect/ddev-intellij-plugin/issues/56

## Release/Deployment notes:
N/A

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3864"></a>


rfay:
Did you compare the original linked json_formatter to the forked one currently in use to understand the difference?

rfay:
I'd much rather use current logrus... but I wouldn't think I would have gone to so much effort to fork it if that were the only thing. I'll take a look.

nico-loeber:
> This also makes it behave like I implied it behaved before also right, with one line per object?

Thats right. :)

PR #3872, opened 2022-05-28
Change app.SiteStatus() to return a predictable and parseable status

nico-loeber:
## The Problem/Issue/Bug:

Status may contain values other than the expected Site* status constants. This is difficult to parse. 

app.SiteStatus() will now returns one ofthe Site* status constants along with a descriptive status.

ddev list will still show detailed status per container, if containers have deviating statuses.
ddev describe -j should contain a status field containing one of the Site* constan values and status_desc with a descriptive status used for ddev list output.
No other behavioral changes are intended.

N/A

## Related Issue Link(s):
N/A

## Release/Deployment notes:
N/A

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3872"></a>


cmd/ddev/cmd/pause_test.go
```diff
@@ -51,7 +52,8 @@ func TestCmdPauseContainers(t *testing.T) {
 	// Confirm all sites are stopped.
 	apps = ddevapp.GetActiveProjects()
 	for _, app := range apps {
-		assert.True(app.SiteStatus() == ddevapp.SitePaused, "All sites should be stopped, but %s status: %s", app.GetName(), app.SiteStatus())
+		status, _ := app.SiteStatus()
+		assert.True(status == ddevapp.SitePaused, "All sites should be stopped, but %s status: %s", app.GetName(), status)
```
rfay, line 56: This assertion comment would be better to give the statusDesc, a good place to use it. Maybe this never happens, don't know.

cmd/ddev/cmd/ssh.go
```diff
@@ -30,14 +28,11 @@ ddev ssh -d /var/www/html`,
 		app := projects[0]
 		instrumentationApp = app
 
-		if strings.Contains(app.SiteStatus(), ddevapp.SiteStopped) {
+		status, _ := app.SiteStatus()
+		if status != ddevapp.SiteRunning {
```
rfay, line 32: Massively better, again.

cmd/ddev/cmd/start_test.go
```diff
@@ -34,7 +34,8 @@ func TestCmdStart(t *testing.T) {
 	// Confirm all sites are running.
 	apps := ddevapp.GetActiveProjects()
 	for _, app := range apps {
-		assert.True(app.SiteStatus() == ddevapp.SiteRunning, "All sites should be running, but %s status: %s", app.GetName(), app.SiteStatus())
+		status, _ := app.SiteStatus()
+		assert.True(status == ddevapp.SiteRunning, "All sites should be running, but %s status: %s", app.GetName(), status)
```
rfay, line 38: This would also be better with showing the statusDesc.

cmd/ddev/cmd/start_test.go
```diff
@@ -53,7 +54,8 @@ func TestCmdStart(t *testing.T) {
 
 	// Confirm all sites are running
 	for _, app := range apps {
-		assert.True(app.SiteStatus() == ddevapp.SiteRunning, "All sites should be running, but %s status: %s", app.GetName(), app.SiteStatus())
+		status, _ := app.SiteStatus()
+		assert.True(status == ddevapp.SiteRunning, "All sites should be running, but %s status: %s", app.GetName(), status)
```
rfay, line 58: Show statusDesc in the assertion as well

rfay (changes requested):
This is looking great. Minor comments and suggestions. I'd like to do some manual testing where we break the web container and db container and verify the behavior of `ddev list` and `ddev describe`. It's easy to break the db container by just trying to start it  with a lower db version. And testing with an add-on service that fails would be great also.

cmd/ddev/cmd/start_test.go
nico-loeber: I've added statusDesc to the message output but I think it might be a bit confusing, that the value shown in the message differs from the one that is asserted on.
I've changed assert.true to assert.equal as this should provide a better native message on fail as it will show the expected status and the given one instead of just saying "expected true, got false".
I also added an assertion on the statusDesc because... why not?

cmd/ddev/cmd/pause_test.go
nico-loeber: See comment below.

cmd/ddev/cmd/start_test.go
rfay: Sorry, I was saying to show BOTH 'status=%v' and statusDesc=%v, not to show statusDesc as status.

cmd/ddev/cmd/start_test.go
```diff
@@ -53,7 +55,9 @@ func TestCmdStart(t *testing.T) {
 
 	// Confirm all sites are running
 	for _, app := range apps {
-		assert.True(app.SiteStatus() == ddevapp.SiteRunning, "All sites should be running, but %s status: %s", app.GetName(), app.SiteStatus())
+		status, statusDesc := app.SiteStatus()
+		assert.Equal(ddevapp.SiteRunning, status, "All sites should be running, but %s status description is: %s", app.GetName(), statusDesc)
```
rfay, line 59:
```suggestion
		assert.Equal(ddevapp.SiteRunning, status, "All sites should be running, but project=%s status=%s statusDesc=%s", app.GetName(), status, statusDesc)
```

cmd/ddev/cmd/pause_test.go
```diff
@@ -51,7 +52,9 @@ func TestCmdPauseContainers(t *testing.T) {
 	// Confirm all sites are stopped.
 	apps = ddevapp.GetActiveProjects()
 	for _, app := range apps {
-		assert.True(app.SiteStatus() == ddevapp.SitePaused, "All sites should be stopped, but %s status: %s", app.GetName(), app.SiteStatus())
+		status, statusDesc := app.SiteStatus()
+		assert.Equal(ddevapp.SitePaused, status, "All sites should be stopped, but %s status description is: %s", app.GetName(), statusDesc)
```
rfay, line 56:
```suggestion
		assert.Equal(ddevapp.SitePaused, status, "All sites should be stopped, but project=%s status=%s, statusDesc=%s", app.GetName(), status, statusDesc)
```

rfay (commented):
Added suggestions. Thanks for your patience. I know how hard it is to work with an opinionated maintainer. Have done it before.

rfay (approved):
Thanks for all the work on this! I went ahead and pushed the buttons on my tiny suggestions and rebased. This can go in when tests are done.

PR #3873, opened 2022-05-29
On traditional Windows, don't use pv on mounted file (ddev import-db)

rfay:
## The Problem/Issue/Bug:

There is an obscure bug apparently in pv (used in `ddev import-db` inside the db container) where on Windows 11 it has a buffer overflow ([link](https://gist.github.com/rfay/18ed619edc366a73b289bd23bfb70bbc)). 

On windows, don't use the mounted database export, but instead push it into the container in the way we do with `--no-project-mount`

- [ ] Do a `ddev import-db` using mariadb:10.3 on a known fussy machine like tb-win11-5

This has not shown up in automated testing except on the new tb-win11-5. I was able to recreate it on another Windows 11 machine by doing a full uninstall/reinstall of Docker Desktop 4.8.2 

Hopefully we can get this tested on tb-win11-5 as well as other machines that weren't misbehaving.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3873"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -547,7 +547,10 @@ func (app *DdevApp) ImportDB(imPath string, extPath string, progress bool, noDro
 	// default insideContainerImportPath is the one mounted from .ddev directory
 	insideContainerImportPath := path.Join("/mnt/ddev_config/", filepath.Base(dbPath))
 	// But if we don't have bind mounts, we have to copy dump into the container
-	if globalconfig.DdevGlobalConfig.NoBindMounts {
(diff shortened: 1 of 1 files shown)


rfay:
Thankfully, this code does not need to be changed. The problem was an out-of-date WSL2 kernel, which was only a problem in Windows 11, which is why it didn't affect the older Windows test runners. The pv maintainer pointed out the [Dirty Pipe Vulnerability](https://dirtypipe.cm4all.com/), It did turn out that the two Windows 11 machines that demonstrated it did not have updates that would update the kernel to 5.10.102.

It turns out that the Windows WSL2 kernel is now automatically updated by Windows Update...  But only if you click the Windows Update "Advanced Options" -> "Receive updates for other Microsoft Products", which is not enabled by default.

PR #3882, opened 2022-06-03
Downgrade drush8 to 8.4.8, refresh to get latest platform cli tool

### LINKED ISSUE #3867: Unable to import postgres database pulled from platform.sh

https://gist.github.com/penance316/f84d993406836bc5e9cf368e64b287ae

### Current Behavior

When running the `ddev pull` command for platform.sh integration the automatic import of my postgres database is failing and reporting an error like 
```
ERROR:  relation "public.users" does not exist
  68KiB 0:00:00 [1.65MiB/s] [>                                                                                                                         ]  0%
Failed to import database db for MyDBName: failed to import database: exit status 3
stdout:
stderr:
```

Running `ddev export-db` on my local DB then `ddev import-db` works correctly as expected.

### Expected Behavior

The database should be pulled from platform.sh provider and imported into my local environment.

### Steps To Reproduce

1. Setup platform.sh integration using `providers/platform.yaml`
2. run pull DB command `ddev pull platform --skip-files`
3. See database backup is pulled down successfully and stored in `.ddev/.downloads/db.sql.gz` as defined by platform.yaml.example file.
4. The command will now move onto importing the database, this is when the failure occurs due to a relation not existing.

### Anything else?

To get around this issue i have been using a custom `db_import_command` hook and a custom postgres script like this
```
\# First drop all tables.
psql -U db -c 'DROP SCHEMA public CASCADE; CREATE SCHEMA public; GRANT ALL ON SCHEMA public TO db; GRANT ALL ON SCHEMA public TO public;'

\# Import via user postgres to avoid credentials prompt
psql -U db db < /mnt/ddev_config/.downloads/db.sql
```

It seems to be an issue with how the platform.sh command takes the backup as it appears to not be compatible with the ddev built in import-db command. 
But it also seems that running this custom import to drop all SCHEMA etc before importing the DB works fine. So I am not sure from which end this issue should be tackled.

Thanks

rfay:
## The Problem/Issue/Bug:

* https://github.com/drush-ops/drush/issues/4979 - The latest drush8 versions have odd output and some problems. Previous 8.4.8 seemed to work right for a long time. This is typically only used with Drupal 7 and lower PHP versions. It should be fine. 
* https://github.com/drud/ddev/issues/3867 - After previous image build platformsh-cli should fix #3867 

Downgrade drush8, rebuild

- [x] `ddev exec 'cd /tmp && /usr/local/bin/drush8 --version` should get 8.4.8
- [x] `ddev exec platform self:update` should show nothing new

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3882"></a>

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -42,7 +42,7 @@ ENV DDEV_PHP_VERSION=$PHP_DEFAULT_VERSION
 ENV PHP_VERSIONS="php5.6 php7.0 php7.1 php7.2 php7.3 php7.4 php8.0 php8.1"
 ENV PHP_INI=/etc/php/$PHP_DEFAULT_VERSION/fpm/php.ini
 ENV YQ_VERSION=v4.7.1
-ENV DRUSH_VERSION=8.4.10
+ENV DRUSH_VERSION=8.4.8
 ENV NODE_LTS=16
 # composer normally screams about running as root, we don't need that.
 ENV COMPOSER_ALLOW_SUPERUSER 1
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -3,7 +3,7 @@
 ### Build ddev-php-base from ddev-webserver-base
 ### ddev-php-base is the basic of ddev-php-prod
 ### and ddev-webserver-* (For DDEV-Local)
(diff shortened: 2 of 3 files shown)


rfay:
Tested and works fine with correct versions on ChromeOS Flex on a 2013 Macbook Air

rfay:
This also works for `ddev pull platform` with Postgres database now, as expected this fixes 
* #3867

PR #3887, opened 2022-06-05
Try docker-compose 2.6.0 in a test run

rfay:
## The Problem/Issue/Bug:

docker-compose 2.6.0 was released

Try it out in test run

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3887"></a>


rfay:
It's good that this passes. I'm not sure we need to update at this point.

PR #3909, opened 2022-06-12
[WIP] Implement goreleaser for build and release creation

rfay:
## The Problem/Issue/Bug:

This was an attempt to implement goreleaser for build and release.

I think most of it will have to be taken on on the build, because the build we have is already a pretty mature process. I tried to recreate it and found trouble at nearly every turn, even though goreleaser is lovely and well-supported. It just doesn't seem to have what it takes to replace what we already have.

What I think I'll do is try to leave the build part as is, and try to use goreleaser for the release parts, using the already built (and signed and notarized) artifacts. That's less disruptive, and probably using goreleaser more for the things it's best at.

So most of this existing work will need to be backed out. I'm putting it here so I can see what may be useful to keep.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3909"></a>


rfay:
I think this can be successful. It's just that
- [ ] Build and release phases have to be completely separate. (This means using the pay-version of goreleaser)
- [ ] Window signing, installer creation, and installer signing have to be broken out into their own scripts
- [ ] macOS signing has to be broken out

Issue #3920, opened 2022-06-17
Allow PHPStorm Code Coverage integration by providing `/opt/phpstorm-coverage` in web container


nico-loeber:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

PHPStorm provides you with the ability to run your PHP unit tests with coverage report directly from the GUI. After execution, you can see in your IDE which lines are covered and which are not.

Sadly this does fail currently in DDEV with the error message 
```
Generating code coverage report in Clover XML format ... failed [00:00.345]
Directory "/opt/phpstorm-coverage" could not be created
```

The path is auto generated by PHPStorm, for example `--coverage-clover /opt/phpstorm-coverage/acol@UdpConnectionTest_testConnect.xml` and I didn't find an option to change the prefix of the generated path.

### Describe the solution you'd like

You can manually create this directory and give it write permission for your user so PHPStorm can create the coverage reports in there but it would be great if this directory would already exist with appropriate permissions, so you can use this PHPStorm feature with zero configuration.

In this issue most of the people add this directory as a docker volume.
https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006971460-Phpstorm-code-coverage-with-docker-container-
I am not aware of the advantges / disadvantages of mounting a docker volume instead of just providing the directory in the container for the coverage reports. I guess it would make the coverage reports persistent when restarting ddev? Your feedback is welcome.

So the solution I would like would be adding a docker volume mount or a directory at /opt/phpstorm-coverage with write permissions for the ddev user to the web container.

### Describe alternatives you've considered

The alternative is adding a DDEV post-start hook executing:

```bash
mkdir -p "/opt/phpstorm-coverage" && chmod a+rw "/opt/phpstorm-coverage"
```

### Additional context

In combination with https://github.com/drud/ddev/pull/3754 and the [DDEV Integration Plugin](https://github.com/php-perfect/ddev-intellij-plugin) solving this issue will allow the use of all test run features (Run, Debug, Run with Coverage) of PHPStorm with zero configuration.

rfay:
You have 3 choices:
1. Bind-mount to something like .ddev/code-coverage or something (persists)
2. Docker volume (not available on the host side but persists)
3. Just create the directory in the image and it doesn't persist. (`ddev restart` throws it all away)

If it persistence doesn't matter the third option works. 

You could probably put it in the ddev-global-cache (indexed by project name)

[...]

PR #3754, opened 2022-04-09
Configure Xdebug debugging sessions to be started by a trigger by default, fixes #3593, fixes #3731

nico-loeber:
## The Problem/Issue/Bug:

* #3593 
* #3731 

As an experienced PHP developer 
- I would like to be able to use all of Xdebug's features with zero configuration.
- I would like to use the development features without triggering a debugging session.
- I do not expect any errors or warnings when executing PHP scripts on the command line.
- I would like to be able to generate line and path coverage reports of my code using phpunit and xdebug.
- I expect to be able to start debugging session via the XDebug browser extension or from my IDE.

As a non experienced PHP developer 
- I would like to start a xdebug debugging session with minimal effort.
- I would like to use the improved error traces provided by xdebug.

## How this PR solves the problem:

This PR is based on the results of the issue #3731.

Xdebug is enabled by default with all of its features. A debugging session is no longer started automatically. 
This allows the usage of the Xdebug development features during development of your application (and not debugging it) as well as triggering alternative features like profiling or tracing.
In this mode you can enable debugging sessions by cookies (browser extension or bookmarks) or by environment variables on the cli (set by your IDE).
To enable the automatic start of a debugging session you can run `ddev xdebug autostart` or by setting the value `xdebug_autostart: true` in the .ddev/config.yml. 
This will start a debugging session on each request and on every started script on the command line. This is the current default behavior, when enabling xdebug.
The Xdebug extension can still be disabled by running `ddev xdebug off` or setting `xdebug: false` in the configuration.

#### Preferring "autostart" over "trigger" as a new mode
As discussed in #3731 the initial idea was to introduce a new mode "trigger" to provide new behavior to stay compatible to the current default behavior. 
I did some research and asked developers what they would expect would happen when they execute "ddev xdebug trigger", in most cases they could not imagine the concrete behavior or describe it precise. 
At the same time "ddev xdebug autostart" was easy to understand and describe for everyone.
The current default behavior of xdebug in ddev was also not expected by the asked developers (no ddev users).

#### Enabling Xdebug by default
Enabling Xdebug is a reasonable default as it provides useful development features like better var_dump function and traces for errors. 
Using Xdebug in a development environment is a common default. To test your application in a more production like environment you can still disable the extension.
With disabled autostart of debugging sessions there are no drawbacks in running Xdebug during development. 

#### Disabling autostart mode by default
Automatically starting debugging sessions is not always intended. 
There are serveral common situations in which you do not like to start a debugging session when Xdebug is active:
[...]

PR #3924, opened 2022-06-18
Add `/opt/phpstorm-coverage` directory to web-container

nico-loeber:
## The Problem/Issue/Bug:

https://github.com/drud/ddev/issues/3920

This PR adds the PHPStorm required directory `/opt/phpstorm-coverage`  to the web-container with read/write premissions for the cli user.

As discussed in #3920, a Docker volume would allow persistence of the coverage reports, but I don't think this is necessary. Therefore, adding the directory was the easiest solution.
PHPStorm does allow loading coverage data from previous runs, but it does not load it automatically at startup. This is a pretty hidden feature and I don't think anyone would use it.

The directory `/opt/phpstorm-coverage` should exist right after `ddev start`. 
Your default user (not root) should be able to read / write files in this directory.

I don't think tests for this are needed.

## Related Issue Link(s):

* https://github.com/drud/ddev/issues/3920

## Release/Deployment notes:

Please don't forget to remove the reference to my tagged container in versionconstants.go.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3924"></a>

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -101,6 +101,9 @@ RUN phpdismod xdebug && curl -sSL --fail --output /usr/local/bin/phive "https://
 RUN set -o pipefail && curl -sSL https://github.com/pantheon-systems/terminus/releases/download/$(curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/terminus.phar --output /usr/local/bin/terminus && chmod 777 /usr/local/bin/terminus
 RUN set -o pipefail && curl -sSL https://github.com/platformsh/platformsh-cli/releases/download/$(curl --silent "https://api.github.com/repos/platformsh/platformsh-cli/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/platform.phar --output /usr/local/bin/platform && chmod 777 /usr/local/bin/platform
 
+RUN mkdir -p "/opt/phpstorm-coverage" && \
+    chmod a+rw "/opt/phpstorm-coverage"
+
 # TODO: v1.18.0 had numerous problems, see https://github.com/acquia/cli/issues/508#issuecomment-967185296
 # and https://github.com/acquia/cli/issues/721 and https://github.com/acquia/cli/issues/722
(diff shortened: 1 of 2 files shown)


containers/ddev-webserver/Dockerfile
```diff
@@ -101,6 +101,9 @@ RUN phpdismod xdebug && curl -sSL --fail --output /usr/local/bin/phive "https://
 RUN set -o pipefail && curl -sSL https://github.com/pantheon-systems/terminus/releases/download/$(curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/terminus.phar --output /usr/local/bin/terminus && chmod 777 /usr/local/bin/terminus
 RUN set -o pipefail && curl -sSL https://github.com/platformsh/platformsh-cli/releases/download/$(curl --silent "https://api.github.com/repos/platformsh/platformsh-cli/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/platform.phar --output /usr/local/bin/platform && chmod 777 /usr/local/bin/platform
 
+RUN mkdir -p "/opt/phpstorm-coverage" && \
+    chmod a+rw "/opt/phpstorm-coverage"
```
rfay, line 105: So you decided to let the coverage stuff be transient, not keep it in a volume?

nico-loeber: Yes, a volume would probably be the 100% solution, but as nobody else asked for the coverage feature till now, I would like to keep it simple. 
PHPStorm doesn't use older coverage reports by default anyway (No coverage indicatos available till you re run the test or you manually load reports).
So let's keep it simple. :)

rfay: OK, thanks. I assume this PR is working for you with PhpStorm + coverage, so I'll pull it when you're ready.

rfay (approved):

nico-loeber:
> OK, thanks. I assume this PR is working for you with PhpStorm + coverage, so I'll pull it when you're ready.

Yes it's tested with PHPStorm of course. You can merge whenever you like.

Please note that there is a dependency to #3754 as xdebug has to be enabled to generate a coverage report at all.

rfay:
Rebased and pushed to drud/ddev-dbserver.

PR #3929, opened 2022-06-20
restore-snapshot → snapshot restore

brandonkelly:
## The Problem/Issue/Bug:

The “Managing Databases” page in the docs was referencing a `ddev restore-snapshot` command that doesn’t exist anymore.

Replaces the reference with `ddev snapshot restore`.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3929"></a>


rfay (approved):
Thanks!

PR #3945, opened 2022-06-28
Bind ddev-router to IPv6 addresses

gilbertsoft:
## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3945"></a>


gilbertsoft:
Will stop here because WSL2 does not support IPv6 without a special kernel :(

PR #3946, opened 2022-06-28
Only copy custom commands files on start and restart

rfay:
## The Problem/Issue/Bug:

A major thread in discord, https://discord.com/channels/664580571770388500/986929382822150154 exposed a problem that appeared when using the PhpStorm extension.

If you tried to use custom commands while it was running, you'd get odd things like

```
OCI runtime exec failed: exec failed: unable to start container process: exec: "/mnt/ddev_config/.global_commands/web/magento": stat /mnt/ddev_config/.global_commands/web/magento: no such file or directory: unknown
Failed to run magento s:up: exit status 126
```

and 
```
ddev nvm use 18
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "/mnt/ddev_config/.global_commands/web/nvm": stat /mnt/ddev_config/.global_commands/web/nvm: no such file or directory: unknown
Failed to run nvm use 18: exit status 126
```

and 
```
ddev nvm use 18
Adding custom/shell commands failed: Failed to chmod file /Users/rfay/workspace/d9/.ddev/.global_commands/web/platform to mode -rwxr-xr-x, err=chmod /Users/rfay/workspace/d9/.ddev/.global_commands/web/platform: no such file or directory
```

@nico-loeber proposed that this might be just because the PhpStorm plugin does a `ddev describe -j` every 10 seconds.

I ran `while true; do ddev describe -j | jq -r .raw.approot ; sleep 10; done` in a window and used custom commands like `ddev nvm` in the other, and immediately ran into this problem.

Only copy/recreate the commands directories on `ddev start` and `ddev restart`.

Use custom/shell commands while the PhpStorm extension is running or while `while true; do ddev describe -j | jq -r .raw.approot ; sleep 10; done` is running.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3946"></a>

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -13,40 +13,10 @@ import (
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/exec"
 	"github.com/drud/ddev/pkg/fileutil"
-	"github.com/drud/ddev/pkg/globalconfig"
 	"github.com/drud/ddev/pkg/util"
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 )
-
-// populateCustomCommandFiles sets up the needed directories and files
-func populateCustomCommandFiles(app *ddevapp.DdevApp) error {
-
-	sourceGlobalCommandPath := filepath.Join(globalconfig.GetGlobalDdevDir(), "commands")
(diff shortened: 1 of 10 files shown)


rfay (commented):
I tested this extensively with custom commands in both project and global web directories, with the 10-second describe going, and it seemed to me like everything was working fine.

PR #3949, opened 2022-07-03
[Docs] Laravel Quickstart: Add command to automatically replace APP_URL with $DDEV_PRIMARY_URL

mandrasch:
## The Problem/Issue/Bug:

`APP_URL` should be set correctly in Laravels `.env`, this was missing in the [quickstart](https://ddev.readthedocs.io/en/stable/users/cli-usage/#laravel-quickstart).

I added and tested the following on OSX 11.6.5 and ddev version v1.19.1.

```bash
ddev exec 'sed -i "/APP_URL=/c APP_URL=$DDEV_PRIMARY_URL" .env'
```

A more readable version (as described [here](https://programarivm.medium.com/how-to-set-a-web-env-variable-from-the-command-line-on-linux-unix-a06016bb9b89)) would be the following. 

```bash
sed -i "s/APP_URL=.*/APP_URL=${DDEV_PRIMARY_URL}/g" .env
```

But I couldn't get it to work with quote escaping. 🤔

No tests needed, just docs improvement

## Related Issue Link(s):

- https://github.com/drud/ddev/issues/3636

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3949"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -284,6 +284,7 @@ Here's a quickstart instructions for a number of different environments:
     ddev start
     ddev composer create --prefer-dist laravel/laravel
     ddev exec "cat .env.example | sed  -E 's/DB_(HOST|DATABASE|USERNAME|PASSWORD)=(.*)/DB_\1=db/g' > .env"
+    ddev exec 'sed -i "s#APP_URL=.*#APP_URL=${DDEV_PRIMARY_URL}#g" .env'
     ddev exec "php artisan key:generate"
     ddev launch
     ```


stasadev:
> But I couldn't get it to work with quote escaping.

It has nothing to do with quotes, just use a different delimiter instead of `/`, e.g. `#`.
It is because  `${DDEV_PRIMARY_URL}` has a `/` inside.

```bash
ddev exec 'sed -i "s#APP_URL=.*#APP_URL=${DDEV_PRIMARY_URL}#g" .env'
```

mandrasch:
Ah, thanks very much @stasadev for the enlightenment! :-o :-o 

(For interested readers: https://stackoverflow.com/questions/16790793/how-to-insert-strings-containing-slashes-with-sed)
 
Not quite sure though if a custom delimiter would be confusing for new users who never worked with `sed` before? 🤔

```
ddev exec 'sed -i "s#APP_URL=.*#APP_URL=${DDEV_PRIMARY_URL}#g" .env'
ddev exec 'sed -i "s@APP_URL=.*@APP_URL=${DDEV_PRIMARY_URL}@g" .env'
ddev exec 'sed -i "s:APP_URL=.*:APP_URL=${DDEV_PRIMARY_URL}:g" .env'
ddev exec 'sed -i "s~APP_URL=.*~APP_URL=${DDEV_PRIMARY_URL}~g" .env'
```

(btw: The cleanest solution would be a CLI tool for dotenv, but did not found a big and major one 🤔)

tyler36 (approved):
Works as intended.

rfay (approved):
Thanks. As @tyler36 is working on, we should figure out how to do this with higher quality in the actual ddev code, but this is a great start.

Issue #3960, opened 2022-07-06
Proxy in custom Dockerfile in Version v.1.19.3


sbiniok:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

The update from v1.19.2 to v1.19.3 introduced a change in the way the web-build/Dockerfile is integrated into the .webimageBuild/Dockerfile. Since custom Dockerfiles are now appended below the auto-generated apt-get command for extra packages, our previously utilized way of setting the proxy for apt (detailed [here](https://github.com/drud/ddev-contrib/blob/master/recipes/proxy/Dockerfile)) is no longer valid. 

### Describe the solution you'd like

May we suggest, either giving developers a way of forcing a custom Dockerfile to be the first in .webimageBuild/Dockerfile (via naming convention for instance), or maybe introducing a new config.yaml parameter (e.g. web_container_proxy) that, if present, would add the following to the auto-generated code:

```
ENV HTTP_PROXY [value from web_container_proxy]
ENV HTTPS_PROXY [value from web_container_proxy]

RUN if [ ! -z "${HTTP_PROXY}" ]; then printf "Acquire {\nHTTP::proxy \"$HTTP_PROXY\";\nHTTPS::proxy \"$HTTPS_PROXY\";\n}\n"  > /etc/apt/apt.conf.d/proxy.conf ; fi
```

rfay:
Thanks @sbiniok - Sorry to break that workflow. I'm going to see if there's a different solution path, and if not we'll have to add a way to introduce those ENV earlier. I think that those could be introduced all the time if they exist, possibly. 

BTW, You can currently do a `docker-compose.*.yaml` that would override the build...

rfay:
It looks like docker-compose v2 had a fix for this, https://github.com/docker/compose-cli/pull/2053 - so there may not be a need for this approach any more. I'll take a look. but bottom line we're hoping for is that the docker cli's proxy settings will automatically be used in build.

hanoii:
While I was working on #3870 I kind of thought something like this was bound to happen, not specifically this, but someone wanting to add something before what you are injecting.

At that point I couldn't think of anything, and there should definitely be a way for ddev to ship their own Dockerfile on top of any project specific ones. The naming convention mentioned here was also my first thought, something like `Dockerfile.pre` but we are adding probably too opinionated ways of doing this. Will think a bit more, and maybe this is a non issue all along. 

[...]

PR #3976, opened 2022-07-08
Use HTTP_PROXY and friends to provide direct proxy support

rfay:
## The Problem/Issue/Bug:

* https://github.com/drud/ddev/issues/3960

Since Dockerfile.* was introduced in DDEV v1.19.3, you can't manually introduce proxy behavior in the build

If $HTTP_PROXY, $http_proxy, or $NO_PROXY is set in the environment (it normally is if configured on the system) then introduce those automatically into the build. 

For example, although these are normally set by the system, we'll set manually:
```bash
export NO_PROXY=127.0.0.1
export HTTP_PROXY=http://proxy:8888
```

After a `ddev start` we see in `.ddev/.webimageBuild/Dockerfile`:

```Dockerfile
### DDEV-injected base Dockerfile contents
ARG BASE_IMAGE
FROM $BASE_IMAGE

ENV http_proxy ""
ENV HTTP_PROXY "http://proxy:8888"
ENV NO_PROXY "127.0.0.1"

ARG username
ARG uid
ARG gid
RUN (groupadd --gid $gid "$username" || groupadd "$username" || true) && (useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' --uid $uid "$username" || useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' "$username" || useradd  -l -m -s "/bin/bash" --gid "$gid" --comment '' "$username")

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests autojump

RUN export XDEBUG_MODE=off && ( composer self-update --2 || composer self-update --2 || true )

RUN chmod 600 ~rfay/.pgpass ~rfay/.my.cnf
ENV NVM_DIR=/home/rfay/.nvm
```

- [ ] Try this out in a real proxy environment
- [ ] Try it out in a non-proxy environment

I don't expect to add proxy testing

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3976"></a>

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -872,11 +872,17 @@ func WriteBuildDockerfile(fullpath string, userDockerfilePath string, extraPacka
 	}
 
 	// Normal starting content is just the arg and base image
+	// The proxy stuff may be removable when https://github.com/docker/compose/issues/9034 is resolved
 	contents := `
 ### DDEV-injected base Dockerfile contents
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
-`
+` + fmt.Sprintf(`
+ENV http_proxy "%s"
+ENV HTTP_PROXY "%s"
+ENV NO_PROXY "%s"
+`, os.Getenv("http_proxy"), os.Getenv("HTTP_PROXY"), os.Getenv("NO_PROXY"))
+
 	contents = contents + `
 ARG username
 ARG uid


rfay:
Considering that 
* https://github.com/drud/ddev/issues/3988
is affected by the same basic thing, I'm going to give up on this PR and do what was originally requested in 
* #3960 
and let people prepend things in the Dockerfile stacking order.

Issue #3977, opened 2022-07-08
docs: 404 not found is broken (again)


tyler36:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Run a Diagnostic and Paste Link Here

n/a

### Current Behavior

I wanted to try the new symlink feature (#3904), so I went to the "latest" docs page and searched for "symlinks".
This should me a link to the "[ddev version history](https://ddev.readthedocs.io/en/latest/users/topics/version-history/)" page but when I clicked on it, I got the screen shot below.

![ddev readthedocs io_en_latest_users_topics_version-history_](https://user-images.githubusercontent.com/7234392/177914909-d01d0842-43bb-4370-b19a-032e56748869.png)

### Expected Behavior

At first, I though the version history page was bad, but after scrolling down, I realised it was a 404 page.

- [ ] fix page display on 404
- [ ] add version history page to new docs (or correct search result?)

### Steps To Reproduce

1. Visit https://ddev.readthedocs.io/en/latest/users/basics/developer-tools/
2. Search for "symlink",  
3. Click the second result, "[DDEV version history](https://ddev.readthedocs.io/en/latest/users/topics/version-history/)". It appear as the second result for me (you need to scroll down due to #3967)
4. Confirm link/page is broken

### Anything else?

Test on:

- Chrome 103
- Firefox 102

rfay:
I removed the version history from the docs because it so often resulted in irrelevant search results. I think it might be a great clue about https://github.com/drud/ddev/issues/3967 that it even showed up in the search results, because if you look in the docs directory you won't find version_history in there any more.

rfay:
BTW, 
* I haven't been adequately maintaining version-history.md since you fixed it up, so it wouldn't have the symlink thing in it.
* The doc was updated to include mention of that in the homeadditions page
* That feature hasn't been released yet so also wouldn't be there even if I'd been adequately maintaining version history.
* The page you want to look at is the homeadditions page, https://ddev.readthedocs.io/en/latest/users/extend/in-container-configuration/?h=homeaddition - I did find it searching for homeaddditions.

And... the 404 problem, like https://ddev.readthedocs.io/en/latest/users/extend/junk - is in fact a problem, and is another thing that I don't think was a problem running locally.

rfay:
This was previously sorted out in 
* #2939


I think those strategies will make it go.

Issue #3967, opened 2022-07-07
docs: search broken, snippets huge, don't use mkdocs-material-insiders


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Run a Diagnostic and Paste Link Here

The new docs has search that is not useful, has to be fixed. Go to https://ddev.readthedocs.io/en/latest/ and search and you get a wall of text like this:

<img width="1066" alt="Get_Started_with_DDEV_-_DDEV-Local_Documentation" src="https://user-images.githubusercontent.com/112444/177801390-7e579648-c6f7-49bb-818c-9edeaba86010.png">
 

### Current Behavior

Wall of text

### Expected Behavior

Useful results

rfay:
Chased this in https://github.com/readthedocs/readthedocs.org/issues/9419 but it definitely doesn't appear to be readthedocs.

tyler36:
Confirmed. Encounted this this morning when I search for "addon".

rfay:
The bottom-line problem here is that I had started to use features from [mkdocs-material-insiders](https://squidfunk.github.io/mkdocs-material/insiders/) because it seemed like a good idea. But then I realized it wasn't a good idea, because it messes up our entire open-source workflow. 

#3980 restores search functionality, but does it by adding mkdocs-material-insiders back in as a stop-gap. Next stop here will be to figure out how to completely back out mkdocs-material-insiders.

PR #3980, opened 2022-07-08
Readthedocs features update and debug current problems for #3967

rfay:
## The Problem/Issue/Bug:

* #3977 
* #3967 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3980"></a>

### DIFF
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -42,6 +42,10 @@ jobs:
 #          file-path: 'README.md', 'CONTRIBUTING.md'
           config-file: 'markdown-link-check.json'
       - name: Test mkdocs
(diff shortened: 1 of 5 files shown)


rfay:
The problem in #3967 is that I had started to use features from [mkdocs-material-insiders](https://squidfunk.github.io/mkdocs-material/insiders/) because it seemed like a good idea. But then I realized it wasn't a good idea, because it messes up our entire open-source workflow. 

This PR restores search functionality, but does it by adding mkdocs-material-insiders back in as a stop-gap. In #3967 we'll figure out how to back out insiders completely, but it's important to have the search working in the meantime.

PR #3983, opened 2022-07-10
Merge the contents of config.yaml and `config.*.yaml`, fixes #3598

### LINKED ISSUE #3598: Merge `config.*.yaml` and `config.yaml environment variables and hooks
[Sorry for the vague title, I struggled to think of what to call it]

I am new to DDEV and am currently in the process of setting up some config files to share with my team.

We use TYPO3 set up with `.env` files but, as it is just the development environment (and for quick setup), I have committed the following to my `config.yaml`

```
web_environment:
- TYPO3_CONTEXT=Development/Local
- TYPO3_DB_HOST=db
- TYPO3_DB_USER=db
- TYPO3_DB_PASSWORD=db
- TYPO3_DB_NAME=db
```

The issue arose when I wanted to change the `TYPO3_CONTEXT` for my local install, without changing the committed file (for risk of accidental committing). Suggest [on Discord](https://discord.com/channels/664580571770388500/941661807855157258) was the creation of a `config.env.yaml` to override settings.

This worked, however I had to re-declare **all** of my variables.

It would be great if you could just declare the variable you wish to overwrite

E.g.

```
web_environment:
- TYPO3_CONTEXT=Production
```

Rather than needing to reset the `DB` credentials again.

torenware:
## The Problem/Issue/Bug:

* #3598

Before this fix, adding a web_environment key to an override config file (`config.SOMENAME.yaml`) would completely replace the content of that string list of environment variables. A fix needs to allow some reasonable notion of a merge, preserving old keys if not restated, updating them if changed.

This PR modifies ddevapp.ReadConfig() to special case the web_environment item in the config data structure. In the WebEnvironment case, for `config.yaml` and `config.override.yaml `(any overriding name),

* The original keys are added to a list
* The new keys are compared with an regexp and will match the key portion of `key=value`. If the key matches for an existing item, the new key=value will replace it.
* Any item in the new version of web_environment that did not match any in the old version will be put at the end of the list.

The PR also adds a bit of testdata and a unit test that confirms
* Old items that are not matched are preserved.
* New items with the same key will replace the old.
* Other new items are indeed added.

1. Note the contents of the web_environment section of your config.yaml file.
2. Add these test keys to that section, say, 
```yaml
web_environment:
- TESTKEY1=oldvalue
- TESTKEY2=notchanged
```
3. Create a config.some-name.yaml file with at least this structure.
```yaml
web_environment:
- TESTKEY1=newvalue
- TESTKEY3=notinold
```
4. Run `ddev restart`, and do `ddev exec printenv | grep TESTKEY` to see the keys on the web container.
5. You should see TESTKEY1 with its new value, TESTKEY2 preserved and with its original value, and TESTKEY3 as well.

The PR  adds a bit of testdata and a unit test that confirm the fix:
* Test data is in pkg/ddevapp/testdata/TestConfigMergeItems/.ddev and includes a simple config.yaml and config.override.yml file, both implementing the web_environment key.
* Unit test `TestConfigMergeItems()` that has asserts for the following:
  * Old items that are not matched are preserved.
  * New items with the same key will replace the old.
  * Other new items are indeed added.

## Release/Deployment notes:

In all cases other than web_environment, the current clobber behavior is maintained. Anyone with a config file that duplicates the keys in config.yaml should see no change in behavior, since the new code will guarantee that all the keys in the override file will still be present and as they are in the override file.

I don't see any need anything different done in deployment unless you want to override the web_environment.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3983"></a>


rfay (changes requested):
Awesome start, thanks!

I like the fact that you're trying to make it possible to *turn off* a previously-set env var.

I was hoping for a full merge of the read apps, because that solves both environment and hooks. Hooks are more complex, because they are two-levels deep, and of course future config might include other multi-level elements. 

There are a number of struct-merge hits on google, https://www.google.com/search?q=golang+merge+structs&oq=golang+merge+stru&aqs=chrome.0.0i512j69i57j0i512l2j0i22i30l6.3528j0j7&sourceid=chrome&ie=UTF-8, I haven't tried them out.  https://github.com/imdario/mergo looks like it has real potential, and looks to be maintained. 

(There are also some yaml-merge hits, but I don't think the yaml is necessarily the thing to merge since we have everything already successfully into structs, but doing it at the yaml level may be appropriate as well.)

Note that other problematic sections that have to be merged will include
* AdditionalHostnames and Additional FQDNs (could be done same as environment I guess)
* Database (but I suspect the overwrite model is OK for it)
* WorkingDir
* OmitContainers
* WebImageExtraPackages
* DbImageExtraPackages

I so much appreciate you getting your fingers into this!

torenware:
> 
> I was hoping for a full merge of the read apps, because that solves both environment and hooks. Hooks are more complex, because they are two-levels deep, and of course future config might include other multi-level elements.
>

So, with a couple exceptions, items typed as `[]string`.

I looked at `type DdevApp struct` with the following questions in mind:

* How many items are likely to be each item. If the number is small, then using a clobber strategy is less work for a developer.
* Why would a developer want to "edit" their config.yaml settings for that particular item.  Is there a use case we can discern.
* Are there use cases for removal of items? I'd say yes for additional_hostnames, additional_fqdns, omit_containers, webimage_extra_packages, dbimage_extra_packages (all simple []string), and hooks (a map).
* Is there a reasonable way to indicate removal?
  + There was for web_environment: `KEYNAME=` is how you'd do it. I'm not sure if this really works with ddev's current code, since I expect the items are parsed and passed to os,Setenv. Not sure if it will be happy with an empty string, so a test is probably needed.
  + Not obvious how to handle any of the other []string variables. We *could* prefix the string with a `-` (dash) or `!`, but we'd need to document this pretty well. Might want to use a regexp to prevent trouble with people doing things we don't expect.
  + Items with map structures probably can do removal; for hooks, for example a removed set of hooks would be something like `post-start: []`. Removing a key from the []string list of a key like `post-start` has the same dev experience issues as the other []string, but likely could be solved the same way.
* Can this be split into another PR later on, with us just solving the items we know developers are asking for. Sounds like web_environment and hooks are the top ask.

Generally speaking, adding items is easy, removing items will take a bit more work.

> There are a number of struct-merge hits on google, https://www.google.com/search?q=golang+merge+structs&oq=golang+merge+stru&aqs=chrome.0.0i512j69i57j0i512l2j0i22i30l6.3528j0j7&sourceid=chrome&ie=UTF-8, I haven't tried them out. https://github.com/imdario/mergo looks like it has real potential, and looks to be maintained.
> 
> (There are also some yaml-merge hits, but I don't think the yaml is necessarily the thing to merge since we have everything already successfully into structs, but doing it at the yaml level may be appropriate as well.)
> 
A library would be helpful, since manipulating data structures once they have been marshalled typically requires the Reflection API. _Be afraid. Be very, *very* afraid_.  Going around go's type checking can be bad juju.

torenware:
Additional question: do you have any idea why TestDdevFullSiteSetup might be failing for Mac/Mutgen? I'm not clear how this would relate to changing the behavior of ReadConfig().

rfay:
Yes, web_environment and hooks are the key asks, the only ones I've ever heard. But a generic approach would cover all (but would not allow removal)

The only failure I see right now (just restarted it) is https://buildkite.com/drud/ddev-macos-mutagen/builds/1924#0181e880-0907-4f9c-b330-b24a655f2a10, and it was a network or upstream issue. But the mac and Windows tests often fail because Docker Desktop is not that reliable, and also because the tests are mostly end-to-end so are impacted by upstream and network issues.

1, `type DdevApp struct`: It can be whatever it is. Some people have lots of environment variables or lots of hooks.
2. "Why would a developer want to "edit" their config.yaml settings for that particular item. Is there a use case we can discern" - Not sure I understand the question. People edit and override all the time, using `ddev config` or editing config.yaml or editing `config.*.yaml`. 
3. Removal: I appreciate you looking carefully at removal; docker-compose has never succeeded at that after all these years. However, it's mostly an edge case asl well. I'm not sure it's fundamental to this PR. 
4. Followup-PR later: If env and hooks are well-handled and we like the result and the resultant code, I would be OK with it, because those are the asks. However, I do think an actual merge might be easier.

torenware:
> Yes, web_environment and hooks are the key asks, the only ones I've ever heard. But a generic approach would cover all (but would not allow removal)
> 

Of course, since the current strategy is "replace", users already have removal.  We may not want to change that unless there's a perceived user need.  Since there are reasons why the current behavior might be near optimal anyway.

> The only failure I see right now (just restarted it) is https://buildkite.com/drud/ddev-macos-mutagen/builds/1924#0181e880-0907-4f9c-b330-b24a655f2a10, and it was a network or upstream issue. But the mac and Windows tests often fail because Docker Desktop is not that reliable, and also because the tests are mostly end-to-end so are impacted by upstream and network issues.

I figured that might be it. I tried to run all the pkg/ddev/ddevapp tests, and got just those kind of failures. Gave up :-)

 
> 2. "Why would a developer want to "edit" their config.yaml settings for that particular item. Is there a use case we can discern" 
> - Not sure I understand the question. People edit and override all the time, using `ddev config` or editing config.yaml or editing `config.*.yaml`.
 
If we know what they're trying to accomplish, then we know whether people mostly want to add items to a list, and which people have instead wanted to remove them.  This should affect our implementation, making either adding or removing as easy as possible.

 > 4.  5. Removal: I appreciate you looking carefully at removal; docker-compose has never succeeded at that after all these years. However, it's mostly an edge case asl well. I'm not sure it's fundamental to this PR. 4. Followup-PR later: If env and hooks are well-handled and we like the result and the resultant code, I would be OK with it, because those are the asks. However, I do think an actual merge might be easier.

Yep. Removal is hard.  Personally, I'd treat handling of docker-compose.yaml type files to another issue.  There are yaml equivalents of the `jq` utility that give you a command-based way of editing that kind of file, and we're better off IMNSHO using one of those systaxes (or better: go libraries if they exist) to handle this.

torenware:
Another question: what is the best way to report config errors back to the user?  If we allow for negation (I'm leaning towards using '!' as a prefix, since YAML syntax doesn't use it), then we need to check that a string doesn't look "screwed up".  And given that for the hooks issue we have the types:

```golang
// for hooks:
type Hooks map[string][]YAMLTask
// YAMLTask defines tasks like Exec to be run in hooks
type YAMLTask map[string]interface{}
```
the best way to parse hooks is likely a set of type switches.  I don't think a merge library will give you the needed tools to handle a merge intelligently, give how the structure is typed.  But since our object is not highly nested, type switches will work well and be easy to understand for anybody playing in that code.

But allowing users to extend in a config.OVERRDE.yaml file will require us to handle bad input; I'll need to give feedback when that happens.

rfay:
> Another question: what is the best way to report config errors back to the user? If we allow for negation (I'm leaning towards using '!' as a prefix, since YAML syntax doesn't use it), then we need to check that a string doesn't look "screwed up". And given that for the hooks issue we have the types:

The current behavior is quite bad IIRC. There are two cases:
1. yaml is invalid: I think the message implies that config.yaml is bad and it's confusing
2. DDEV won't allow whatever's in the yaml (maybe bad mariadb version or whatever). I think this also is a confusing result

We could easily validate each yaml file we load with the regular app validation. The only problem with that is they may get a validation error that isn't relevant because it's not something that actually gets used. (This would be the case with using mariadb:5.2 in the config.yaml and having mariadb:10.4 in the config.something.yaml)

torenware:
Been a while since I've had to do commit squashing, and I'm clearly out of practice.

I may need @rfay to fix my damn commits once the tests start passing. Sorry about that, but I figure that's not a bad trade for the code if it actually works :-)

rfay:
You don't have to do any squashing, we do squash-on-pull. The one awkward thing for this PR is since it's your first, Github won't run tests until I approve them. After the first, they run automatically.

davereid-pfg:
I would definitely like to also have the `hooks` information merged in addition to the web_environment listing, looking forward to updates on this!

rfay:
@davereid-pfg artifacts are updated in https://github.com/drud/ddev/pull/3983#issuecomment-1179739882 and your review would be welcome. (I'll look today or tomorrow @torenware, thanks so much for this great initiative!)

torenware:
Does the asset list need updating? It looks like the version from two days back might be different than last night's build.

rfay:
@torenware could you please update the OP to update to the current approach? Thanks!

rfay (changes requested):
This isn't quite working yet; hooks aren't being merged, although environment variables are. 

I used the example config files in https://gist.github.com/rfay/91fcaefa0c085c696f8f12713f265d95

And the result of `ddev debug configyaml` is as shown in https://gist.github.com/rfay/91fcaefa0c085c696f8f12713f265d95#file-ddev-debug-configyaml-txt - you'll see that the hooks weren't merged, the ones in config.1.yaml overwrote. You can also see in `ddev start` that only the config.1.yaml hooks are getting run, https://gist.github.com/rfay/91fcaefa0c085c696f8f12713f265d95#file-ddev-start-txt

Please add test coverage for the merging of hooks, and if it's not too hard, turn off the feature of your IDE that is moving imports around. The world won't end if you can't figure that out.

I haven't studied your actual code yet. Thanks so much for the great effort on this!

pkg/ddevapp/config.go
```diff
@@ -3,19 +3,20 @@ package ddevapp
...
 	"text/template"
 	"time"
 
+	"github.com/Masterminds/sprig/v3"
+	"github.com/drud/ddev/pkg/dockerutil"
+	"github.com/drud/ddev/pkg/globalconfig"
+	"github.com/drud/ddev/pkg/nodeps"
+	"github.com/drud/ddev/pkg/versionconstants"
+	copy2 "github.com/otiai10/copy"
+
```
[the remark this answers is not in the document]
torenware: Changing vcode's handling go code is *hard*.  It's using official golang language tools under the hood, which is not readily configurable. If the imports in config.go are getting reorganized, you're basically fighting city hall. 

I can do manual merges with git to prevent this, if you're really committed to the current ordering.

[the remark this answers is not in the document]
rfay: You don't have to change it but no, gofmt doesn't automatically do this.

torenware:
> This isn't quite working yet; hooks aren't being merged, although environment variables are.
> 
> I used the example config files in https://gist.github.com/rfay/91fcaefa0c085c696f8f12713f265d95
>

I'll see what I get with those files.  You expect the items to be added if they do not match, correct?
 
> Please add test coverage for the merging of hooks, and if it's not too hard, turn off the feature of your IDE that is moving imports around. The world won't end if you can't figure that out.
> 

I actually *have* test cases for this, which using my test data in `pkg/ddevapp/testdata/TestConfigMergeItems`, and `pkg/ddevapp/config_test.go::TestConfigHooksMerge`.  Not sure why your test data is handled differently, but I'll take a look at it.

Current test files are:

config.yaml:

```yaml
webimage: config.yaml
web_environment:
  - LARRY=curly
  - MOE=standard
  - CURLEY=bald
additional_hostnames:
 - somename
 - someothername
hooks:
  post-start:
    - exec: "simple web command"
    - exec-host: "simple host command"
  post-import-db:
    - exec: "drush uli"

```

and 

config.override.yaml

```yaml
web_environment:
  - SHEMP=fake
  - LARRY=balding
  - MOE=hair-piece
additional_hostnames:
 # add a hostname
 - somename-new
 # remove a hostname
 - "!someothername"
hooks:
  post-start:
    - exec: "simple random expression"

```

The asserts tested are:

```golang
	assertTask(true, "post-start", "exec", "simple random expression")
	assertTask(true, "post-start", "exec-host", "simple host command")
	assertTask(true, "post-start", "exec", "simple web command")
	assertTask(true, "post-import-db", "exec", "drush uli")
```

i.e., all of the listed tasks were present, from both tasks.

I'll try your cases as well, and see why they are not handled as expected.

> I haven't studied your actual code yet. Thanks so much for the great effort on this!

torenware:
OK, looks like you're using stale binaries.  The assets you sent Dave too are the original versions; there's no hook processing code. You may want to generate the current assets and make those available.

I substituted my original test data for your config files; these asserts all pass, indicating locally that the hook code was invoked and that both the config.yaml and config.1.yaml versions got merged as we'd want them to be.

```golang
	// assertTask(true, "post-start", "exec", "simple random expression")
	// assertTask(true, "post-start", "exec-host", "simple host command")
	// assertTask(true, "post-start", "exec", "simple web command")
	// assertTask(true, "post-import-db", "exec", "drush uli")

	assertTask(true, "pre-start", "exec-host", "echo pre-start-orig")
	assertTask(true, "post-start", "exec", "echo post-start-orig")
	assertTask(true, "pre-start", "exec-host", "echo pre-start-1")
	assertTask(true, "post-start", "exec", "echo post-start-1")

```

rfay:
I apologize.

You're right. I used gh pr checkout, which I haven't used until lately, and it did me wrong and gave an old version. I have only recently been trying it out, and I guess it won't remain in my toolkit. My normal technique is to add a remote, then check out the branch from the remote. And when I do that I see the hooks working fine in ddev debug configyaml

pkg/ddevapp/config_merge.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/imdario/mergo"
+	"gopkg.in/yaml.v2"
+)
+
+// mergeConfigToApp does an unmarshall with merging
+func (app *DdevApp) mergeConfigToApp(source []byte) error {
+	// save away state before merges.
+	unmergedApp := &DdevApp{}
+	err := mergo.Merge(unmergedApp, app)
```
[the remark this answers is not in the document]
rfay: If I'm not mistaken, all the logic you have here can be replaced with this pseudocode:

```
// start with app as the loaded official starting ddevapp
for _, i := range extraFiles {
        unmergedNewConfig := &DdevApp{}
	err = yaml.Unmarshal(i, unmergedNewConfig)
	if err != nil {
		return err
	}
        err = unmergedNewConfig.Validate()
         if err != nil {
		return err
	}
        err = mergo.Merge(app, unmergedNewConfig)
         if err != nil {
		return err
	}
}

What do you think about trying that out? It may require a little more work, but seems like mergo is doing the work it's intended to do here.

File: pkg/ddevapp/config_merge.go [hunk 3]
```diff
@@ -0,0 +1,206 @@
...
+	// anything else from the upstream config for any
+	// key except for web_environment.
+	err = yaml.Unmarshal(source, app)
+	if err != nil {
+		return err
+	}
+
+	// add merges here.  Default strategy is to clobber old
+	// keys.
+	mergeableItems := map[string]mergeData{
```
rfay, line 36: As above, I think mergo should do *all* of this for you, without any need for any of this?

pkg/ddevapp/config_merge.go
```diff
@@ -0,0 +1,206 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// merge an arbitrary string list.
+func (app *DdevApp) mergeStringList(ptr interface{}, oldList []string) error {
```
rfay, line 90: Don't think this is needed?

pkg/ddevapp/config_merge.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	// save back the merged results
+	*newList = results
+	return nil
+}
+
+// merge the web_environment string list.
+func (app *DdevApp) mergeWebEnvironment(ptr interface{}, oldEnv []string) error {
```
rfay, line 120: Don't think this is needed?

pkg/ddevapp/config_merge.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		if !found {
+			result = append(result, newItem)
+		}
+	}
+	app.WebEnvironment = result
+	return nil
+}
+
+func (app *DdevApp) mergeHooks(ptr interface{}, oldHooks map[string][]YAMLTask) error {
```
rfay, line 172: Or this?

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
 		assert.Empty(app.MariaDBVersion)
 	}
 }
+
+// TestConfigMergeStringList verifies that string merges update w/o clobber
```
rfay, line 1323: Thanks so much for getting tests together for this!

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
...
+		err = app.Stop(true, false)
+		assert.NoError(err)
+		err = os.RemoveAll(projDir)
+		assert.NoError(err)
+	})
+	err = os.Chdir(app.AppRoot)
+	assert.NoError(err)
+
+	// test matches allowing for delete syntax (a prefixed !)
+	assertSimpleMatch := func(expected bool, match string, setting []string) {
```
rfay, line 1349: Let's not test for features we're not going to introduce yet.

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
...
+
+// TestConfigMergeStringList verifies that string merges update w/o clobber
+func TestConfigMergeStringList(t *testing.T) {
+	assert := asrt.New(t)
+	pwd, _ := os.Getwd()
+
+	projDir, err := filepath.Abs(testcommon.CreateTmpDir(t.Name()))
+	require.NoError(t, err)
+
+	err = fileutil.CopyDir("./testdata/TestConfigMergeItems/.ddev", filepath.Join(projDir, ".ddev"))
```
rfay, line 1331: Please use separate testdata for separate tests, even if they're identical. That way fiddling with the testdata for one test won't break another one. And use t.Name() here instead of the literal `TestConfig...`, so if you happen to change the test name it can follow you.

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
...
+	})
+	err = os.Chdir(app.AppRoot)
+	assert.NoError(err)
+
+	// test matches allowing for delete syntax (a prefixed !)
+	assertSimpleMatch := func(expected bool, match string, setting []string) {
+		test := assert.True
+		if !expected {
+			test = assert.False
+		}
```
rfay, line 1353: Please just use assert.True() in these cases instead of using the `test` alias. The `test` alias here just obscures what's going on.

pkg/ddevapp/config_test.go
rfay: Am I mistaken that this function really only checks for deletions, and thus should have a name indicating that? (Of course, I've already said I'd rather not include tests for features we're not going to include yet.)

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
...
+// TestConfigMergeEnvItems verifies that config overrides for web_environment
+// override and do not clobber settings from config.yaml.
+func TestConfigMergeEnvItems(t *testing.T) {
+	assert := asrt.New(t)
+	pwd, _ := os.Getwd()
+
+	projDir, err := filepath.Abs(testcommon.CreateTmpDir(t.Name()))
+	require.NoError(t, err)
+
+	err = fileutil.CopyDir("./testdata/TestConfigMergeItems/.ddev", filepath.Join(projDir, ".ddev"))
```
rfay, line 1381: Again, use t.Name() for your testdata access, thanks.

rfay: But I imagine you don't have to do the copying at all, since NewApp() takes a project directory, so you can just NewApp(path/to/testdata). That would not be true if the test were messing with the app, but it's just reading it here.

pkg/ddevapp/config_test.go
```diff
@@ -1317,3 +1319,220 @@ func TestDatabaseConfigUpgrade(t *testing.T) {
...
+		assert.NoError(err)
+		err = os.RemoveAll(projDir)
+		assert.NoError(err)
+	})
+	err = os.Chdir(app.AppRoot)
+	assert.NoError(err)
+
+	// make test results a bit clearer with our
+	// own matcher.
+	assertHasKey := func(expected bool, match string) {
```
rfay, line 1400: I think it might be more correct to just provide a testdata config.yaml file with everything included in it and compare it to the merged result. Wouldn't that be more explicit that just listing keys?

rfay (changes requested):
I did a lot better this time with the correct code instead of the old code gh gave me. Ah, new tools :( 

First, thanks so much for taking this on, and for providing tests. That's fantastic, and it's going to work, and it's going to be a great contribution.

* I'm pretty sure the logic in the code can be massively simplified using mergo, and explained that in near-code.
* The tests can be more comprehensive and much easier if you just provide finished-config yaml files then marshall your built-with-overrides config and compare the resulting files.

Other misc comments, but this is going to be great.

rfay:
Sadly, I've discovered something that can never be merged, must be replaced. If you have this in config.yaml:
```yaml
database:
  type: mariadb
  version: "10.3"
```

Then add this in `config.x.yaml`:
```yaml
database:
  type: postgres
  version: "14"
```

You'll end up with:
```yaml
database:
  type: mariadb
  type: postgres
  version: "10.3"
  version: "14"
```

Amazingly though, the last thing configured wins, so in this case that stanza actually works out and gets postgres:14

And also amazingly, this also works:
```yaml
database:
  type: postgres
  type: mariadb
  version: "14"
  version: "10.3"
```

So maybe there's no problem here, but wow.

rfay:
No, that was me experimenting with different options. I just manually created the merged option and tried it out in ddev.

My proposal of a simple full deep merge would result in the examples I show.

rfay:
Hoping we can move forward on this (without putting pressure on you!). Please let me know if you have questions about the suggestions in the review, especially https://github.com/drud/ddev/pull/3983#discussion_r919501242

pkg/ddevapp/config.go
```diff
@@ -334,6 +335,23 @@ func (app *DdevApp) LoadConfigYamlFile(filePath string) error {
...
 
+// LoadConfigYamlFile loads one config.yaml into app, overriding what might be there.
+func (app *DdevApp) LoadConfigAndMerge(filePath string) error {
+	source, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("could not find an active ddev configuration at %s have you run 'ddev config'? %v", app.ConfigPath, err)
+	}
+
+	// validate extend command keys
+	err = validateHookYAML(source)
```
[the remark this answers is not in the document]
torenware: Might be a good idea. Main issue is that `LoadConfigYamlFile()` is used elsewhere in an independent test, and I did not want to change its behavior.  Probably a good idea to add a call to `validateHookYAML()` post merge though; I'll try that.

pkg/ddevapp/config_merge.go
torenware: Needed.  It allows me to use the same code for multiple fields.  Unlike PHP, go structs don't have a syntax to update a field of struct like you do in most languages, so I need to carry around a pointer to the member of the struct in the instance I need to update.

pkg/ddevapp/config_test.go
torenware: What I'd really like to do is refactor the set up here so that I'm not just copying the first half of the test.  Hell: I copied similar code from another test :-) But using t.Name() will resolve the issue to both our satisfaction, I think.  I'll do that.

pkg/ddevapp/config_test.go
torenware: Typically test data will have a few items in them; the assert wrapper is also munging the data structure, since there's no assert in your library (AFAICT) that does a "string in slice" test.  Keeps the test DRY.

pkg/ddevapp/config_merge.go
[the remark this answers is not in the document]
torenware: Problem is that the app object wraps reference types like maps and slices.  Doing a copy as you suggest screwed up the reference types, with changes in app bleeding into unmergedApp. Hence the deep copy library.

Was a slightly puzzling issue that took a spell to debug.

pkg/ddevapp/config_merge.go
torenware: Open to that. Not hard, but as we've seen, it's not straightforward to merge structs, which don't act like most languages. So each struct needs to be handled a la carte.  

That list tells you what I'm processing currently.  I think the one thing I skipped is the database struct, which I can add to this PR if you like.

pkg/ddevapp/config_merge.go
torenware: Since this one is called only once, I could do this w/o redirection.  It's most valuable for the string lists; I'd need separate but very similar routines for each and every string list w/o it.  The redirection makes the processing logic a lot more compact.

Also, if we ever do tackle deletion: all that code would need to be touched. I'd as soon do it once, rather than modify very similar code 5 or 6 times.

pkg/ddevapp/config_test.go
torenware: Done.

rfay:
This should be ready to pull @torenware - Could you please rebase and add a commit that adds "config-star-yaml-merging" to `ddev debug capabilities` in https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/debug-capabilities.go - thanks!

torenware:
I believe this is properly merged including the debug-capabilities command -- which appears to have been among the 17 commits that are now merged. Kick off the tests at your leisure.

rfay:
Unfortunately, you didn't do it right, not sure if it will be a problem. You'll want to learn how to use `git fetch upstream && git rebase upstream/master` and then `git push --force` to put your commits on top of the existing upstream/master. Do you want to do it again? When things are working right, you wont' see yourself pushing upstream commits in the PR, just the ones that belong to this PR.  

This is all wrong now. (You can tell by just looking at what's changed, and you see all kinds of things you didn't intend to do).

The previous working commit was at https://github.com/drud/ddev/pull/3983/commits/7fe85b286afd0fda19f69fea691cdbd036de66dc, which is still there and can be recovered and worked from.

rfay:
You may also have forgotten that you pulled my PR in using the Github Web UI, so you needed to update your remote before doing anything else, `git fetch <forkremote>` (my forked remote is always "rfay", I recommend that you use "torenware". `git remote -v` will show you your remotes. 

So you would have needed to:

1. `git fetch --all` to get everything updated in your remotes
2. `git reset --hard torenware/issue-3558-merge-config` to get your local branch equal to the one on github. 
3. `git rebase upstream/master` to rebase against current code. 
4. Add your new commit. 

These are worthy things to master.  And there is more than one way to do it.

rfay:
The docs fix is in. I can rebase this or you can, as you like.

rfay (changes requested):
There are complexities here that still have to be dealt with. 

if `config.1.yaml` does not set the `type`, it gets a default `type`, which is `php`, which then overrides the actual project type coming from `config.yaml`. 

So there are some defaults that must not be set. We already prevent overwriting name and approot, but there are probably more. It's the defaults that are the issue.

Test coverage needs to be updated to cover these cases as well. 

(Discovered with manual testing)

rfay:
Unfortunately, the idea of creating a NewApp() results in this bad behavior for things like webserver_type, etc. Any default-set thing in the config.*.yaml will overwrite the config with the wrong data. I think the problems are all about scalar settings.

rfay:
I hope you'll take a look at this @torenware - It seems to me that we probably need some more permutations in TestConfigMerge.

torenware:
HUZZAH!!

rfay:
Congratulations, thanks so much for your leadership and exploration on this.

davereid-pfg:
Could I get some clarification on the expectation that hooks in two different config files get merged together or do they override each other? I'm using the current release and finding that adding an additional post-db-import hook in a config.local.yaml overrides the existing post-import-db hook I have set in my config.base.yaml file.

davereid-pfg:
Ah! This hasn't been actually released yet!

rfay:
@davereid-pfg you can now use this with `brew install --HEAD --fetch-head` and would love your reports. It will be in v1.20.0, hopefully in a couple of weeks or so.

rfay:
If you want to know if a version of ddev supports it, `ddev debug capabilities` will contain `config-star-yaml-merging`

PR #3990, opened 2022-07-12
Insert user Dockerfiles earlier, fixes #3988, fixes #3960

### LINKED ISSUE #3960: Proxy in custom Dockerfile in Version v.1.19.3

### LINKED ISSUE #3988: Need to be able to prepend Dockerfile commands before nodejs is installed due to corporate network SSL termination proxy

Due to our corporate network using a SSL termination proxy for all HTTPS traffic, we need to install some manual CA certificates before our environments attempt to "talk" to anything using cURL or apt.

I would like to be able to *prepend* Dockerfile commands before the commands are generated by DDEV to install NodeJS via cURL and apt. The only way to modify the generated Dockerfile is by appending things after the generated contents.

We currently have to turn off the `nodejs_version` setting in our config.yaml file, and manually add NodeJS back in our Dockerfile appending file:

```Dockerfile
ARG BASE_IMAGE
FROM $BASE_IMAGE

COPY *.crt /usr/local/share/ca-certificates/
RUN update-ca-certificates --fresh

# This line is basically just copied from the .ddev/.webimageBuild/Dockerfile that was auto-generated by DDEV but now it runs after our certificates
RUN curl -sSL --fail https://deb.nodesource.com/setup_14.x | bash - && apt-get install nodejs && npm config set unsafe-perm true && npm install --global gulp-cli yarn
```

### Additional context

This is related to https://github.com/drud/ddev/issues/3960 but is a more generic ask to solve our problem since we not actually using an actual proxy.

rfay:
## The Problem/Issue/Bug:

* #3988 
* #3960 

In DDEV v1.19.3, multiple user Dockerfiles were supported, and they were all added *after* the initial portions of generated Dockerfile content. This meant that things like HTTP_PROXY/HTTPS_PROXY could not be added before the `apt-get` things or related things. And it also meant that things like custom-cert installation couldn't be done before the other things happened. 

- [x] Move user-provided Dockerfile content to near the beginning of the generated .ddev/webimageBuild/Dockerfile, so user-generated Dockerfile content can take precedence.
- [x] Adds explicit support for proxy, using the HTTP_PROXY and HTTPS_PROXY and NO_PROXY environment variables. docker-compose should be doing this, and claims to, but it doesn't actually work out.

- [x] Create a .ddev/web-build/Dockerfile and/or `Dockerfile.*`
- [x] `ddev start`
- [x] Inspect the .ddev/.webimageBuild/Dockerfile to review the actual order of inclusions
- [x] Verify that the actual build is done right by `ddev ssh` and inspect the build, or `docker image inspect` or review in Docker Desktop.
- [x] Verify that setting HTTP_PROXY in the traditional way works

## Release considerations

- [ ] Since ddev now directly supports proxies (via HTTP_PROXY and friends) the ddev-contrib recipe should be updated to point to the ddev docs.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3990"></a>

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -74,6 +74,10 @@ ENV COMPOSER_HOME=""
 
 **Remember that the Dockerfile is building a docker image that will be used later with ddev.** At the time the Dockerfile is executing, your code is not mounted and the container is not running, it's just being built. So for example, an `npm install` in /var/www/html will not do anything useful because the code is not there at image building time.
 
+## HTTP proxy support inside containers and during build
+
+DDEV will automatically recognize systems that have the environment variables HTTP_PROXY, HTTPS_PROXY, and NO_PROXY set to configure proxy behavior. It will then configure generated images to include those values and set them up to be able to use the proxy during build time and at run time.
+
 ### Debugging the Dockerfile build
 
 It can be complicated to figure out what's going on when building a Dockerfile, and even more complicated when you're seeing it go by as part of `ddev start`.
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -36,7 +36,7 @@ For more complex requirements, you can add:
 * `.ddev/db-build/Dockerfile`
 * `.ddev/db-build/Dockerfile.*`
 
-These files' content will be appended to ddev's own Dockerfile for each image.
+These files' content will be inserted into the constructed Dockerfile for each image. They are inserted *before* most of the rest of the things that are done to build the image, and are done in alpha order, so `Dockerfile` is inserted first, followed by `Dockerfile.*` in alpha order. You can examine the resultant Dockerfile (which should not be changed as it is generated) at `.ddev/.webimageBuild/Dockerfile` and you can force a rebuild with `ddev debug refresh`.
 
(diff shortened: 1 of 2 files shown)


rfay:
I've tested this, so going to pull it and it will now be testable with `brew unlink ddev && brew install --HEAD --fetch-head drud/ddev/ddev`

PR #3999, opened 2022-07-15
Support for pre.Dockerfile.*

hanoii:
Partially reverts 
* #3990 
with a `pre.Dockerfile.*` approach and keeping regular `Dockerfile*` at the end of the insertions.

From Discord: https://discord.com/channels/664580571770388500/997163304822771842

# Release Consequences

- [ ] Update https://github.com/drud/ddev-contrib/tree/master/recipes/proxy to show the new technique

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3999"></a>

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -5,7 +5,7 @@ defaults:
     shell: bash
 on:
   push:
-    branches: [ master, main, 20220714_publish_binaries_on_master_workflow ]
+    branches: [ master, main ]
   release:
     types: [ created ]
 
--- a/cmd/ddev/cmd/debug-capabilities.go
+++ b/cmd/ddev/cmd/debug-capabilities.go
@@ -11,7 +11,7 @@ var DebugCapabilitiesCmd = &cobra.Command{
 	Use:   "capabilities",
 	Short: "Show capabilities of this version of ddev",
 	Run: func(cmd *cobra.Command, args []string) {
-		capabilities := []string{"multiple-dockerfiles", "interactive-project-selection", "ddev-get-yaml-interpolation"}
+		capabilities := []string{"multiple-dockerfiles", "interactive-project-selection", "ddev-get-yaml-interpolation", "pre-dockerfile-insertion"}
 		output.UserOut.WithField("raw", capabilities).Print(strings.Join(capabilities, "\n"))
 	},
 }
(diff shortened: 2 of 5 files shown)


rfay (commented):
Just took a quick look before bed, looks like it's probably fine. Will look more closely tomorrow. The docs problem is just a silly markdown rule.  Thanks!

rfay:
Oh, there is a very specific reason for the explicit proxy support - docker-compose doesn't actually put the HTTP_PROXY into the build stage at this point, meaning that you can't build images or add to images that require proxying without detecting the HTTP_PROXY as was done in that code. It's unfortunate but true.  We have to figure out how to get HTTP_PROXY into the build stage.

hanoii:
> Oh, there is a very specific reason for the explicit proxy support - docker-compose doesn't actually put the HTTP_PROXY into the build stage at this point, meaning that you can't build images or add to images that require proxying without detecting the HTTP_PROXY as was done in that code. It's unfortunate but true. We have to figure out how to get HTTP_PROXY into the build stage.

I don't think this is correct. First, the only thing your code seems to be doing is checking for the presence of the env var and if so, making it available on the Dockerfile through `ENV ` and then doing the APT part. The checking is not necessary and you can just add it to your Dockerfile without that. To second this we have the actual use case of #3960, the fact that we was able to make it work on 1.19.2 means that you don't really need the code you added to make it work. Am I missing something here?

hanoii:
Fixed the md error.

hanoii:
Hmm, it still didn't pass the docs check but I see no error in the job, and I can't seem to be able to trigger a re-run.

rfay:
Don't worry about that docs check. Sometimes random sites start blocking various kinds of access, we'll sort it out. Obviously not the fault of this PR.

rfay:
Works with https://github.com/rfay/ddev-proxy-support - would love anybody else to try that out. You need the artifacts from https://github.com/drud/ddev/pull/3999#issuecomment-1185132955 (or build this PR) and then `ddev get rfay/ddev-proxy-support` - You'll need at least HTTP_PROXY set in your environment.

rfay:
Rebased and added capabilities marker in `ddev debug capabilities`

rfay (approved):
Looks great to me, tests great. Thanks! And thanks for the collaboration. I am so very happy not to be putting explicit proxy support in here.

Issue #3854, opened 2022-05-20
Add PATH_INFO CGI variable to nginx configuration


longwave:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using SimpleSAMLphp inside a ddev project with nginx, it crashes because it cannot find the `$_SERVER['PATH_INFO']` variable, which is expected to be set. This is documented at https://simplesamlphp.org/docs/1.17/simplesamlphp-install#configuring-nginx

### Expected Behavior

The PATH_INFO variable should be passed from nginx to PHP.

This should just be a case of adding the following line to the `.php` location block:
```
fastcgi_param PATH_INFO $fastcgi_path_info;
```

I don't think adding this will harm any other projects using DDEV.

rfay:
@longwave please respond, and PR is welcome.

rfay:
I'm happy to continue thinking about this with a PR or more discussion, but closing for now, as no response from OP. Happy to continue the conversation here or elsewhere.

davereid:
We just ran into this as well on a project.

rfay:
Still happy to return to @longwave 's PR, 
* https://github.com/ddev/ddev/pull/4001

I'll reopen or you can just redo.

mark-felix:
We need path info support also. Why this is not enabled by default?

rfay:
@mark-felix please read the (long) discussion in the associated PR:
* https://github.com/ddev/ddev/pull/5454

PATH_INFO is actually something from the Apache/CGI world I guess, and as far as we were able to figure out it was a flaw in Simplesaml that it was trying to use it. 

If you can figure out how to make it work, we can return to it.

rfay:
I put Claude to work on it and seem to have a solution. 
* https://github.com/ddev/ddev/pull/8844

mark-felix:
Hello, @rfay !
Thanks a lot for the information and for the fix! I am working on a Moodle site, and it still needs PATH_INFO for proper functioning. Thanks a lot for the fast response and for taking care of it! Also thanks for all who contributed before!

PR #4001, opened 2022-07-15
Add PATH_INFO CGI variable to nginx config.

longwave:
## The Problem/Issue/Bug:

When using SimpleSAMLphp inside a ddev project with nginx, it crashes because it cannot find the `$_SERVER['PATH_INFO']` variable, which is expected to be set. This is documented at https://simplesamlphp.org/docs/1.17/simplesamlphp-install#configuring-nginx

The `PATH_INFO` variable will be passed from nginx to PHP.

This should have no side effects on other platforms as they don't use the `PATH_INFO` variable

Check `phpinfo();` after applying the fix to see that `$_SERVER['PATH_INFO']` is present and correct in the PHP Variables section

(or set up SimpleSAMLphp inside a ddev project, but that is non trivial)

## Related Issue Link(s):

#3854

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4001"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/nginx/sites-enabled/nginx-site-default.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/nginx/sites-enabled/nginx-site-default.conf
@@ -39,6 +39,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;


rfay (changes requested):
This looks innocuous, but the tests don't reflect what it did at all, because this is the default nginx in the upstream container. The place it should really be changed is in https://github.com/drud/ddev/blob/master/pkg/ddevapp/webserver_config_assets/nginx-site-drupal9.conf and all the siblings in that directory.

Note that you can change this in any project in the .ddev/nginx-full directory with the instructions given there, I'm sure you know that already. But the place for this change is in the actual configs that gets used, not in the container. (But it's fine to change it here as well.) But what we want is all the things in https://github.com/drud/ddev/blob/master/pkg/ddevapp/webserver_config_assets

longwave:
Ah OK, will push some fixes for that soon. I work around this at present by changing `.ddev/nginx-full` but it would be helpful if I didn't have to ship an nginx config at all, as we use SimpleSAMLphp on more than one project.

rfay:
Closing this for now, happy to return to it and reopen when you get some time to put into it. Should be easy.

rfay:
If you're still interested in this after all these years @longwave 

* https://github.com/ddev/ddev/pull/8844

Issue #3456, opened 2021-12-17
Add daemon support and http/s proxying for Node JS


andy-blum:
**Is your feature request related to a problem? Please describe.**
As a front-end dev that's used DDEV for Drupal development, DDEV is _fantastic_ for the standard, monolithic PHP CMS, but takes some work to get things really well customized and set up for any node projects. This work might be "simple" in terms of Docker config or server setup, but as someone that doesn't have much experience in the devops/sysadmin world, this is frequently unfamiliar and sometimes feels unattainable without a step-by-step recipe.

With the rise of decoupled/headless CMSes, it would be nice to have better support of node surfaced to the ddev config file so average users can setup and maintain this type of project with a simple local development environment. In my mind, the primary things I would need:

1. Be able to pick a node version per #2651.
2. Be able to specify a npm/yarn command (and the relevant location for that command). Ideally this would be able to run as a background process, returning control of the CLI to me. I've tried running a node server via a hook before and since the process doesn't complete until I kill the server, the ddev start command never finishes
3. Be able to proxy a ddev domain or subdomain to the port that the command starts.

**Describe the solution you'd like**
I'd like to be able to run `ddev start` and have it do the following:
- Use the correct version of node
- Spin up my Drupal/Wordpress/TYPO3/Magento/PHP backend
- Spin up my gatsby/next.js/other node app frontend
- Complete the ddev start command, and let me continue to use the same terminal window
- Serve both the drupal backend and my JS frontend at `.ddev.site` urls

**Additional context**
If ddev starts to venture into supporting decoupled JS frontends, I wonder if there's interest/value in supporting JS backends like Strapi or Keystone CMSes

rfay:
nodejs has to run in a particular directory, true? How would that be specified and what's a good example?

rfay:
Would you be willing to provide an example project (just a dummy simple project) that could be used for testing this? You can specify manual steps required.

andy-blum:
Typically yes. My personal site is running a [Drupal 9 backend ](https://github.com/andy-blum/portfolio-drupal) and a [Gatsby frontend](https://github.com/andy-blum/portfolio-gatsby).

On my local machine, these two respositories are stored together in `~/Sites/andy-blum.com/`, so I would imagine that would be the new home for the `.ddev` directory

```
Sites/
|-- andy-blum.com/
|   |-- .ddev/
|   |-- drupal/
|   |-- gatsby/
|-- example.com/
```

[...]

PR #4006, opened 2022-07-16
#3456 - add project type "node" and configure nginx as reverse proxy for it

hchonov:
This PR aims at providing a solution for running Node.js apps behind a proxy with DDEV. 

See https://github.com/drud/ddev/issues/3456

## The Problem/Issue/Bug:
Currently it is not possible to use ddev to start directly a project unless one configures a custom nginx config with reverse proxy settings.

1. A new project type "node" is being added that can be specified in the .ddev/config.yml.
2. The new project type comes without any configuration, but a dedicated nginx config preconfigured as reverse proxy.
3. The reverse proxy will initially forward to https://localhost:3000 so you would have to start your application on that port. This can be changed later, but should be sufficient for an initial implementation.

Further with the current changes it is no longer required to set `disable_settings_management: true` in the config.yml of the project.
The following config.yml configuration should be enough to run a NextJS app using express:

```
name: my-node-app
type: node
webserver_type: nginx-fpm
nodejs_version: "16"
omit_containers: ['db', 'dba']
hooks:
  post-start:
    - exec: NODE_ENV=development PORT=3000 node server.js
```

If you are using express-ipfilter you might need something like this to allow the request initiator's address after a check that it is running on a dev environment:
```
    if (dev) {
      return IpFilter(allowedIpList, {
        mode: 'allow',
        // Allow the ddev private addresses.
        trustProxy: 'uniquelocal',
      })(req, res, _next);
    }
```
 
No test coverage yet.

## Related Issue Link(s):
https://github.com/drud/ddev/issues/3456

## Release/Deployment notes:
No.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4006"></a>


rfay (changes requested):
This is interesting thinking. I hadn't thought of node being a project type, which may be a reasonable thing. But... aren't there going to be many different types of project that would use node?

Also, nothing you're doing here launches nodejs or tells it where it should be running. The nginx config just assumes it can pass things off to node.

We'd want to start nodejs in supervisord I'd imagine, so add another conf file to the supervisor config?

hchonov:
> This is interesting thinking. I hadn't thought of node being a project type, which may be a reasonable thing. But... aren't there going to be many different types of project that would use node?

With my PR I am covering the use case of a NextJS decoupled app that lives in its own repository and retrieves the data from the backend via APIs. I think that we can start with one use case and go further adding support for other use cases if you agree. If needed we can introduce the type nextjs instead of node in order to make it more clear that we currently only support nextjs apps, but since probably the way of starting applications might be pretty similar I am not sure if it worth it to make the separation?

> Also, nothing you're doing here launches nodejs or tells it where it should be running. The nginx config just assumes it can pass things off to node.

If you look at the config.yml the first implementation was intending that the developers take care of starting the node app via a post-start hook. My thought here was that we do not have a project that has a standard entry point like index.php, but everyone wants to start their app in a different way. 

> We'd want to start nodejs in supervisord I'd imagine, so add another conf file to the supervisor config?

I agree and I've pushed a new commit, and now instead of using a post-start hook we can define the command that the supervisor should use.

Now it is possible to start the node app with ddev using the following `config.yml`:
```
name: my-node-app
type: node
webserver_type: nginx-fpm
omit_containers: ['db', 'dba']
nodejs_version: "16"
web_environment:
  - NODE_ENV=development
  - PORT=3000
node_cmd: node server.js --prefix /var/www/html
```

where the environment variables that I am setting ared used by my `server.js`.

rfay:
Why shouldn't we just add nginx proxying of port 3000 to all projects? Super simple thing to do. Why add a new project type for that?

What I'd prefer to do here is:
1. A PR to proxy port 3000 using both HTTP_EXPOSE and HTTPS_EXPOSE, and solve https://github.com/drud/ddev/issues/4032 by adding NODE_EXTRA_CA_CERTS. 
2. A ddev add-on that explores this territory as well as we can, to try to find out what the "general" solution for nodejs is. Already https://github.com/drud/ddev-browsersync and https://github.com/torenware/ddev-viteserve are exploring this territory, but there's more to be done.

rfay:
I so appreciate your inspiration thinking about this. I think we'll do as I suggested - 
* https://github.com/drud/ddev/pull/4057

Could you please review that? Please try it out and say what you think is missing. I really want to be able to provide first-class nodejs support for all the many ways it's used.

Closing this in favor of #4057 - which you absolutely inspired.

PR #4010, opened 2022-07-17
Bump docker images for v1.19.4

rfay:
## The Problem/Issue/Bug:

Time for new release. Bump to new images.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4010"></a>


rfay:
Looks like the behavior here is the same as in 
* https://github.com/drud/ddev/issues/3748#issuecomment-1186381830

PR #4016, opened 2022-07-18
Support for HostWorkingDir commands directive (`ddev npm` and `ddev yarn` use relative dir)

hanoii:
Documentation change: https://github.com/hanoii/ddev/blob/commands-relative/docs/content/users/extend/custom-commands.md#hostworkingdir-annotation-container-commands-only

Discord https://discord.com/channels/664580571770388500/998670519383638087

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4016"></a>

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -287,7 +287,7 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 		if relative {
 			wd, err := app.GetRelativeWorkingDirectory()
 			if err == nil {
-				opts.Dir = opts.Dir + wd
+				opts.Dir = path.Join(opts.Dir, wd)
 			}
 		}
 
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -285,7 +285,7 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			NoCapture: true,
 		}
 		if relative {
-			wd, err := app.GetRelativeWorkingDirectory()
+			wd := app.GetRelativeWorkingDirectory()
 			if err == nil {
 				opts.Dir = path.Join(opts.Dir, wd)
 			}
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -285,10 +285,7 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			NoCapture: true,
 		}
 		if relative {
-			wd := app.GetRelativeWorkingDirectory()
-			if err == nil {
-				opts.Dir = path.Join(opts.Dir, wd)
-			}
+			opts.Dir = path.Join(opts.Dir, app.GetRelativeWorkingDirectory())
 		}
 
 		if execRaw {
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -193,9 +193,10 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 					continue
 				}
 
-				// Default is to exec with bash interpretation (not raw)
+				// Execute the command matching the host working directory relative
+				// to the app root.
 				relative := false
-				if val, ok := directives["WorkingDirRelative"]; ok {
+				if val, ok := directives["HostWorkingDir"]; ok {
 					if val == "true" {
 						relative = true
 					}
@@ -284,8 +285,8 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			NoCapture: true,
 		}
 		if relative {
-			wd := app.GetRelativeWorkingDirectory()
-			if wd != `` {
+			wd, err := app.GetRelativeWorkingDirectory()
+			if err == nil {
 				opts.Dir = opts.Dir + wd
 			}
 		}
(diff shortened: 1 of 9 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -388,6 +390,22 @@ func (app DdevApp) GetDocroot() string {
...
+	re, err := regexp.Compile(app.AppRoot + `/`)
+
+	if err != nil {
+		return ``
+	}
+
+	pwd, err := os.Getwd()
+	if err != nil {
+		return ``
+	}
```
hanoii, line 404: @rfay is this how you would handle errors in such a standard function?

rfay: No, in almost every case I would return `(string, error)` and return the error. 

There are a few exceptions where it doesn't seem like the error is important (then emit complaint) and at the top-level, sometimes a util.Fail(), which stops with error.

hanoii: I was going to try that initially, will change it.

hanoii:
@rfay I see you flag it as draft again, what do you think is missing here to consider it? tests?

rfay:
I apologize, I haven't yet looked at it, but you said it didn't have a test yet, and so I'm trying to prioritize the many PRs that are in play. Yours is super interesting but not critical for v1.20, and I'm trying to get those critical ones, some of which have been on hold for a long time, reviewed and in.

hanoii:
That's good, thanks! If you have a vague idea on how such a test could be conducted/worked on, I'll give it a try. It doesn't look like there are tests for commands annotations, but it was a conclusion after a very quick look.

Is there any functional tests that tests ddev specific commands?

rfay:
Thanks - there are several tests in https://github.com/drud/ddev/blob/e0536988ce32f21afecafc24255d9d1875c1ceda/cmd/ddev/cmd/commands_test.go (like the launch and mysql tests) that would probably do.


Those are *command* tests, meaning they use the actual ddev binary, so you'll want to make sure it's built and use DDEV_BINARY_FULLPATH in your environment to make sure it's the binary that gets used.

pkg/ddevapp/global_dotddev_assets/commands/web/nvm
```diff
@@ -5,5 +5,6 @@
 ## Usage: nvm [flags] [args]
 ## Example: "ddev nvm install 6"
 ## ExecRaw: true
+## HostWorkingDir: true
```
rfay, line 8: nvm shouldn't have this, nvm sets up a global nvm. 
```suggestion
```

cmd/ddev/cmd/commands.go
```diff
@@ -275,6 +284,13 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			Tty:       isatty.IsTerminal(os.Stdin.Fd()),
 			NoCapture: true,
 		}
+		if relative {
+			wd, err := app.GetRelativeWorkingDirectory()
+			if err == nil {
+				opts.Dir = opts.Dir + wd
```
rfay, line 290:
```suggestion
				opts.Dir = path.Join(opts.Dir, wd)
```

pkg/ddevapp/ddevapp.go
```diff
@@ -388,6 +390,22 @@ func (app DdevApp) GetDocroot() string {
 	return app.Docroot
 }
 
+// GetAppRoot return the full path from root to the app directory
```
rfay, line 393:
```suggestion
// GetRelativeWorkingDirectory returns the relative working directory relative to project root
```

pkg/ddevapp/ddevapp.go
```diff
@@ -388,6 +390,22 @@ func (app DdevApp) GetDocroot() string {
...
+	if err != nil {
+		return ``, err
+	}
+
+	pwd, err := os.Getwd()
+	if err != nil {
+		return ``, err
+	}
+
+	return re.ReplaceAllString(pwd, ""), nil
```
rfay, line 406: You don't want more than one replacement do you? A simple string replacement would work wouldn't it? strings.TrimPrefix()?

pkg/ddevapp/ddevapp.go
```diff
@@ -388,6 +390,22 @@ func (app DdevApp) GetDocroot() string {
 	return app.Docroot
 }
 
+// GetAppRoot return the full path from root to the app directory
+func (app *DdevApp) GetRelativeWorkingDirectory() (string, error) {
+	re, err := regexp.Compile(app.AppRoot + `/`)
```
rfay, line 395: I don't think regex is the right thing to do here. The AppRoot may have regex characters in it, like a "." for example, which could match anything.

How about strings.TrimPrefix() - seems easier.

rfay (changes requested):
Thanks so much for this, looks like it's going to be great. 

I haven't manually tested yet, but it sounds like you had good results. 

It does need a test, but you know how to do that per other comment. 

Thanks!

rfay:
I took this for a spin and it sure seems to be working great!

rfay:
I think this is a lovely initiative and it would be great to get it in for v1.20.0, thanks!

hanoii:
@rfay I was away for a few days and will be again next week. I am not sure I'll be able to move forward a lot on this this week. Not sure what's the timeline for v1.20.0 is. I will still keep it in the radar and see how it goes.

rfay:
If I get to it before you do is that OK?

hanoii:
Sure, you don't really need asking :)

rfay:
@hanoii did the requested rework tasks and added tests. Could you take a look and manually test please?

rfay:
Pretty sure that when the package.json isn't there both npm and yarn fail, and it shouldn't be in any directory but the one we're looking at. Important thinking and worth checking though.

hanoii:
The test Is creating the package on the root folder. So when you run the test in the current version it will work because the command will always run on root. It won't fail

hanoii:
I missed that. Sorry

PR #4026, opened 2022-07-20
Bump mysql to 8.0.29

rfay:
## The Problem/Issue/Bug:

We're finally able to upgrade mysql:8.0 to 8.0.29.

This is complex for arm64 because Oracle doesn't publish arm64 packages, and percona doesn't provide arm64 versions of xtrabackup.

As a result, we have to wait for arm64 Ubuntu:20.04 to have the right version of mysql 8.0, then wait for percona to get a matching xtrabackup version.

Today the stars aligned and we have both, so I built the xtrabackup, then the arm64 images, and now here is the actual PR.

- [ ] Configure a project for mysql:8.0. 
- [ ] `ddev exec -s db mysql --version` should show 8.0.29 on both amd64 and arm64 
- [ ] Use the project and see how it works.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4026"></a>

### DIFF
--- a/containers/ddev-dbserver/Makefile
+++ b/containers/ddev-dbserver/Makefile
@@ -10,7 +10,7 @@ CURRENT_ARCH=$(shell ../get_arch.sh)
 # So has to explicitly declare anything it might need from there (like SHELL)
 SHELL = /bin/bash
 
-BUILD_TARGETS=mariadb_5.5_amd64 mariadb_10.0_amd64 mariadb_10.1_both mariadb_10.2_both mariadb_10.3_both mariadb_10.4_both mariadb_10.5_both mariadb_10.6_both mariadb_10.7_both mariadb_10.8_both mysql_5.5_amd64 mysql_5.6_amd64 mysql_5.7_both mysql_8.0_both_8.0.26
+BUILD_TARGETS=mariadb_5.5_amd64 mariadb_10.0_amd64 mariadb_10.1_both mariadb_10.2_both mariadb_10.3_both mariadb_10.4_both mariadb_10.5_both mariadb_10.6_both mariadb_10.7_both mariadb_10.8_both mysql_5.5_amd64 mysql_5.6_amd64 mysql_5.7_both mysql_8.0_both_8.0.29
 TEST_TARGETS=$(shell if [ "$(CURRENT_ARCH)" = "amd64" ] ; then \
 	echo "mariadb_5.5_test mariadb_10.0_test mariadb_10.1_test mariadb_10.2_test mariadb_10.3_test mariadb_10.4_test mariadb_10.5_test mariadb_10.6_test mariadb_10.7_test mariadb_10.8_test mysql_5.5_test mysql_5.6_test mysql_5.7_test mysql_8.0_test"; \
 	else \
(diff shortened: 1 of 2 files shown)


rfay:
snapshots won't work yet due to https://docs.percona.com/percona-xtrabackup/8.0/em/instant.html

The default db creation needs to have `ALGORITHM=INSTANT` TURNED OFF.

rfay:
Judging from a reply on https://www.percona.com/blog/percona-xtrabackup-8-0-29-and-instant-add-drop-columns/#comment-10973674 mysql 8.0.29 with xtrabackup 8.0.29 will not be usable with ddev, as it requires application-level changes to do "the right thing" with table creation.

> Hey Randy,
There is no system option to alter the default algorithm of “ALTER TABLE … ADD|DROP COLUMN” in either Oracle MySQL Community or Percona Server for MySQL. You must alter your SQL SYNTAX to explicitly use the “ALGORITHM=INPLACE” modifier. In the upcoming Percona Server 8.0.29 release (coming soon) we have temporarily changed the default algorithm for these “ALTER TABLE” operations to use the “INPLACE” algorithm instead of the unstable “INSTANT” algorithm. We will likely revert this change once the “INPLACE” feature has stabilized, hopefully as early as 8.0.30.

jkabat:
@rfay v8.0.30 is available, can you please create a new image?

jkabat:
By the way, what is the difference between the tags [drud/ddev-dbserver-mysql-8.0:20220719_mysql_8_0_29] and [drud/ddev-dbserver-mysql-8.0:v1.21.3] ?

tyler36:
@jkabat could you link to the release notes with the required update?

jkabat:
@tyler36 done.

tyler36:
Thanks!

rfay:
It does look like there's an xtrabackup for 8.0.30, and it does look like they believe it will work, https://www.percona.com/blog/percona-xtrabackup-8-0-29-and-instant-add-drop-columns/#comment-10973674

Unfortunately, the Ubuntu version has moved forward to 8.0.31, and we use Ubuntu as the base image, since Oracle never packages arm64 packages.  And there is not xtrabackup fro 8.0.31 yet. You can find out how difficult this problem is by reading https://ddev.readthedocs.io/en/latest/developers/release-management/#maintaining-ddev-dbserver-mysql-57-80-arm64-images

Anyway, please open an issue requesting 8.0.30. I'll see if there's a way to get Ubuntu to install 8.0.30 instead of the default current 8.0.31.

rfay:
Seems like there's no way to get the Ubuntu 8.0.30, perhaps they passed it by. 

Our best bet will be when xtrabackup finally comes out with an 8.0.31 version. It does look like that may be close, https://github.com/percona/percona-xtrabackup/pull/1396

rfay:
> By the way, what is the difference between the tags [drud/ddev-dbserver-mysql-8.0:20220719_mysql_8_0_29] and [drud/ddev-dbserver-mysql-8.0:v1.21.3] ?

The 20220719_mysql_8_0_29 tag was the attempt (from this PR) to get 8.0.29 to work. But then the Percona folks said "Oh, but 8.0.29 doesn't actually work with xtrabackup". 

* Oracle doesn't package arm64 packages for anything, so we use Ubuntu, which has curated mysql-server-8.0 package
* Percona doesn't package arm64 packages, but we build xtrabackup from source
* xtrabackup-8.0 must be used with the matching mysql-server minor version. 
* xtrabackup-8.0 always lags behind mysql-8.0 releases (which are every 6 months)

Happy to have help out of this hell :) Maybe we'll get it with coming Xtrabackup version.

jkabat:
Thanks. Not even 8.0.30 has been officially released in docker hub: https://github.com/docker-library/official-images/pull/13573
Lets see how long it takes to get 8.0.31.

rfay:
I keep a close watch on xtrabackup, they're moving the PR along for 8.0.31. Hopefully we'll get those two in sync.

rfay:
@jkabat finally succeeded in the upgrade to 8.0.31, https://github.com/drud/ddev/pull/4644

PR #4033, opened 2022-07-20
Skip option validation for composer create-project

gilbertsoft:
## The Problem/Issue/Bug:

If `ddev composer create-project` is called with an option the message about the usage of `ddev composer create` instead is not shown but an error about an unknown option.

The issue poped up in this DDEV Discord thread https://discord.com/channels/664580571770388500/999382823033118801.

This PR enforces to skip the option validation and shows the correct message in any case.

Run e.g. `ddev composer create-project -s centarro/commerce-kickstart-project kickstart` which will show the correct message about the correct usage with `ddev composer create` now.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4033"></a>

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -154,6 +154,9 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 // when they try ddev composer create-project
 var ComposerCreateProjectCmd = &cobra.Command{
 	Use: "create-project",
+	FParseErrWhitelist: cobra.FParseErrWhitelist{
+		UnknownFlags: true,
+	},
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


cmd/ddev/cmd/composer-create.go
```diff
@@ -154,6 +154,9 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 // when they try ddev composer create-project
 var ComposerCreateProjectCmd = &cobra.Command{
 	Use: "create-project",
+	FParseErrWhitelist: cobra.FParseErrWhitelist{
```
rfay, line 157: I just want to know how you figured out how to do this!

rfay (approved):

cmd/ddev/cmd/composer-create.go
gilbertsoft: Well, it's used in other commands too and I've had also a look at the cobra library to verify it.

PR #4040, opened 2022-07-24
Custom command for Craft CMS "craft" utility

bencroker:
## The Problem/Issue/Bug:

Running Craft CMS commands requires one of the following syntaxes, which is fine but somewhat verbose.

```shell
ddev exec php craft

ddev exec ./craft
```

This PR adds a new `craft` convenience command so that running the command is less verbose.

```shell
ddev craft
```

Running `ddev craft` should show the help text for `craft`.

## Release/Deployment notes:

This is a small step in getting better support for Craft CMS in DDEV. I have added it for the `php` project type, which may be too broad. It is my hope that @pixelandtonic team will consider submitting (or perhaps are already working on) a complete Craft CMS project type for DDEV in due course.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4040"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/craft
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/craft
@@ -4,7 +4,7 @@
 ## Description: Run Craft CMS command inside the web container
 ## Usage: craft [flags] [args]
 ## Example: "ddev craft db/backup" or "craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
-## ProjectTypes: craftcms
+## ProjectTypes: php
 ## ExecRaw: true
 
 php ./craft "$@"
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/craft
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/craft
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run Craft CMS commands inside the web container
+## Usage: craft [flags] [args]
+## Example: "ddev craft db/backup" or "craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
+## ProjectTypes: craftcms
+## ExecRaw: true
+
+php ./craft "$@"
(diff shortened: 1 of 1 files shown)


bencroker:
I was just reminded of https://twitter.com/angrybrad/status/1541844776521170944, so perhaps better to wait for the "official" PR.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run Craft CMS command inside the web container
+## Usage: craft [flags] [args]
+## Example: "ddev craft db/backup" or "craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
+## ProjectTypes: php
```
rfay, line 7: This is the one thing I thought of after we talked, there's not yet a craft project type. But using it with project type php is a pretty decent solution probably. May annoy some people using the "php" project type on other CMSs that "craft" shows up as a command.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
```diff
@@ -0,0 +1,10 @@
...
+## ExecRaw: true
+
+php ./craft "$@"
```
rfay, line 10: Probably don't need the `php` here but it does no harm.

rfay (commented):
This looks fine to me, but I haven't manually tested yet. Hope to do that. And maybe we should go for the craft project type.

Thanks!

pkg/ddevapp/global_dotddev_assets/commands/web/craft
bencroker: Yes, agreed that it's not ideal. I think it might be better to wait for @angrybrad to submit his PR, which I assume will include this command anyway.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
bencroker: Right, I just followed the convention set by the other commands, for example https://github.com/drud/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/artisan

pkg/ddevapp/global_dotddev_assets/commands/web/craft
rfay: I think it's OK though.

rfay (approved):
I spent some time looking at craft and manually tested this and think it's OK. Let's go with it.

PR #4049, opened 2022-07-28
Add NODE_EXTRA_CA_CERTS to web container, fixes #4032

### LINKED ISSUE #4032: Set up node.js to trust mkcert CA

We're trying to run a node.js script in the web container that sends requests to Drupal at the ddev project URL over HTTPS. Node.js has made the decision to maintain their own list of trusted certs for reasons that are unclear to me, so the CAs that are configured to be trusted at the OS-level don't help.

By default, let's set `NODE_EXTRA_CA_CERTS=/mnt/ddev-global-cache/mkcert/rootCA.pem` in the web container. This will tell Node to trust the mkcert CA.

_No response_

### Additional context

_No response_

rfay:
## The Problem/Issue/Bug:

* #4032

Add the suggested env var inside container

- [x] `ddev exec 'echo $NODE_EXTRA_CA_CERTS'`
- [x] Try this script inside container, call it try.js; change the target URL. From https://stackoverflow.com/a/48729712/215713. it should work with `ddev exec node try.js`, but if you `ddev ssh` and `unset NODE_EXTRA_CA_CERTS` it should fail when running it there with `node try.js`
```js
const getScript = (url) => {
    return new Promise((resolve, reject) => {
        const http      = require('http'),
              https     = require('https');

        let client = http;

        if (url.toString().indexOf("https") === 0) {
            client = https;
        }

        client.get(url, (resp) => {
            let data = '';

            // A chunk of data has been recieved.
            resp.on('data', (chunk) => {
                data += chunk;
            });

            // The whole response has been received. Print out the result.
            resp.on('end', () => {
                resolve(data);
            });

        }).on("error", (err) => {
            reject(err);
        });
    });
};

(async (url) => {
    console.log(await getScript(url));
})('https://d9.ddev.site/');
```

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4049"></a>

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -193,6 +193,7 @@ services:
     - LINES
     - MYSQL_HISTFILE=/mnt/ddev-global-cache/mysqlhistory/${DDEV_SITENAME}-web/mysql_history
     - MYSQL_PWD=db
+    - NODE_EXTRA_CA_CERTS=/mnt/ddev-global-cache/mkcert/rootCA.pem
     - PGDATABASE=db
     - PGHOST=db
     - PGPASSWORD=db


cweagans:
Apologies for the slow response. This works great, thank you!

rfay:
Thanks!

PR #4061, opened 2022-08-01
[docs] Mention use of upload_dir for TYPO3 projects with nonstandard fileadmin

Kayu84:
## The Problem/Issue/Bug:
Using mutagen with TYPO3 and secured-web results in an issue where composer cannot create a symlink for fileadmin under public/

Setting the ddev "upload_dir" to `../private/fileadmin` prevent the creation of a docker volume (type: bind), see: .ddev-docker-compose-full.yaml.
The right directory here is the one under "private/"

## Related Issue Link(s):
https://github.com/mutagen-io/mutagen/issues/359

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4061"></a>


rfay:
Thanks! This at least gets us a good way to provide it to people!

rfay:
I did some minor rewrites @Kayu84 would appreciate it if you'd take a look and see if it's OK.

PR #4078, opened 2022-08-04
Add PHPUnit command with path remapping

### LINKED ISSUE #4077: Add PHPUnit command with path remapping
I would like a command on the host that will run PHPUnit in the DDEV web container.

It should
- [ ] display output with color codes
- [ ] remap ddev container paths to project paths

tyler36:
Fixes #4077

## The Problem/Issue/Bug:
I run different projects with different PHP requirements.
I generally use an alias to switch my host PHP version but sometimes forget when I switch projects. This causes my tests to fail.

I been using the command on ddev-contrib but it has a problem:
- file paths are displayed relative to the container and not the host

This means I can not click on the filename in VScode to open them

This PR adds a command to run PHPUnit tests inside the container.
It attempts to rewrite the path which allows me to click through.

It does have some issues though:
- it only works on linux
- it does not display out put in color

I am unsure how to correct this issues so this PR is "draft". 
I feel like there might be a "better" solution but I do not know this area well enough.
Perhaps the community can help.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4078"></a>

### DIFF
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -190,7 +190,7 @@ func TestCustomCommands(t *testing.T) {
 	assert.NoError(err)
 
 	// Make sure that all the official ddev-provided custom commands are usable by just checking help
-	for _, c := range []string{"launch", "mysql", "npm", "php", "xdebug", "yarn"} {
+	for _, c := range []string{"launch", "mysql", "npm", "php", "phpunit", "xdebug", "yarn"} {
 		_, err = exec.RunHostCommand(DdevBin, c, "-h")
 		assert.NoError(err, "Failed to run ddev %s -h", c)
 	}
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/phpunit
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/phpunit
@@ -1,0 +1,9 @@
+#!/bin/bash
+#ddev-generated
+
+## Description: Run Symphony phpunit tests
+## Usage: phpunit <test>
+## Example: "ddev phpunit" or "ddev phpunit tests/Util"
+CWD=$(pwd)
+
+ddev . ./vendor/bin/phpunit "$@" | sed "2,\$s|/var/www/html|$CWD|"


rfay (commented):
I think what you actually want is the new 

`# HostWorkingDir: true`

just introduced for `ddev npm` and `ddev yarn`? And don't you want a web command?

https://github.com/drud/ddev/blob/f4e9b2af5e3fbc3d19e00ec2c7cca334bd683d36/pkg/ddevapp/global_dotddev_assets/commands/web/yarn#L8

I also think you'd be better to just rely on the $PATH instead of giving an explicit path to vendor/bin. vendor/bin is already in the default $PATH.  I have a note to also make $PATH work when the composer_dir is elsewhere (and thus the vendor is elsewhere), https://github.com/drud/ddev/issues/4060

So I'd think you'd want to just make a phpunit command that is a web command, just like `yarn`, and that doesn't require a particular explicit path.  But I may not understand what you're up to.

This also may be a good candidate for a `ddev-get` add-on if it's a specialized use of phpunit, not sure it is.

tyler36:
Goal:
If I run phpunit on my WSL host with a broken test:

```shell
$ phpunit
PHPUnit 9.5.21 #StandWithUkraine

F.                                                                  2 / 2 (100%)

Time: 00:00.368, Memory: 22.00 MB

There was 1 failure:

1) Tests\Unit\ExampleTest::test_that_true_is_true
Failed asserting that false is true.

/home/user13/code/lara-vite/tests/Unit/ExampleTest.php:16
```

Running the command via DDEV `ddev . ./vendor/bin/phpunit` gives me

```shell
...
1) Tests\Unit\ExampleTest::test_that_true_is_true
Failed asserting that false is true.

/var/www/html/tests/Unit/ExampleTest.php:16
```

When my project is open in VSCode this path is obviously incorrecty.  And because the path is incorrect, I can't click-to-open.

rfay:
I'm all in favor of this, but it's best to work it outside the ddev repo; when it matures it can become an add-on, and of course if we can figure out how to do it for all CMSs we could include it in DDEV. Thanks for all your contributions! Closing this for now.

PR #4080, opened 2022-08-05
Provide DDEV_VERSION env variable in container

cmuench:
## The Problem/Issue/Bug:

As a developer I want to execute scripts in the container. Sometimes it's good to know which ddev version is currently used to have a different logic or to warn that some features require an ddev update.

Register a new DDEV_VERSION variable in the container.

Start a new project and run `echo $DDEV_VERSION`. It should print the currently used ddev version as string.
For stable versions it's a tag like `v1.20.0`.

Test was added to `pkg/ddevapp/ddevapp_test.go`.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4080"></a>

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -178,6 +178,7 @@ services:
     - DDEV_FILES_DIR
     - DDEV_WEBSERVER_TYPE
     - DDEV_XDEBUG_ENABLED
+    - DDEV_VERSION
     - DEPLOY_NAME=local
     {{ if not .DisableSettingsManagement }}
     - DRUSH_OPTIONS_URI=$DDEV_PRIMARY_URL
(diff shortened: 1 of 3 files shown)


rfay:
Seems reasonable. I've been using `ddev debug capabilities` for this sort of thing though, because it doesn't always require a semver library to use :)

rfay:
Actually, ddev could have a semver check in `ddev debug`. Then we wouldn't have to do that in bash or php or whatever.

cmuench:
@rfay Yes, but this runs then on the host, right?
I want to use it in script inside the container. The scripts should be universal and contain conditions for different ddev versions.

rfay:
Agreed, I was just thinking out loud that maybe we'd want to figure out how to provide that capability elsewhere. I prefer semver checks, but generally find them really annoying. Bash doesn't have good tools. PHP and golang require a bunch of extra scaffolding to check (but do a good job when you do it). Etc.

cmuench:
No worry. In the meantime I understand your hints. :-)
BTW I missed the capability feature. There are still hidden features in ddev.

rfay (approved):
Seems. fine to me. Worked with manual testing. Thanks!

PR #4095, opened 2022-08-08
Don't use docker-compose to wait for db container to become healthy

rfay:
## The Problem/Issue/Bug:

* https://github.com/docker/compose/issues/9732

If the db container exits, for example because the db type has been changed and the existing database is incompatible with the new one, the `docker-compose up` hangs forever without any information. Unfortunately we can't let docker-compose do the work for us as we had anticipated in #3858.

* Remove the docker-compose depends_on
* Add inline code to check db container

Note that DDEV v1.20 web_extra_daemons already waits for these things (now), so the original problem described in #3702 should be resolved *if* it uses web_extra_daemons, which would be great.

- [x] Make db container fail by changing its type. `ddev restart`. You should get clear feedback, not a hang.

## Related Issue Link(s):

* https://github.com/drud/ddev/issues/3702
* https://github.com/drud/ddev/pull/3858

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4095"></a>

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -149,12 +149,6 @@ services:
     user: '$DDEV_UID:$DDEV_GID'
     hostname: {{ .Name }}-web
 
-    {{ if not .OmitDB }}
-    depends_on:
-      db:
-        condition: service_healthy
-    {{ end }}
-
     ports:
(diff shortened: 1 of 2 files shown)


rfay:
@mikebarkas and @andreasfernandez would love it if you could review this. 

@andreasfernandez your best strategy for your extra daemon is to try out web_extra_daemons.

PR #4102, opened 2022-08-09
[docs] fix colima installation step 2

mikesnoeren:
## The Problem/Issue/Bug:
The 2nd step assumed you had DDEV installed, which isn't necessarily true as setting up Docker is usually the first step.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4102"></a>


rfay (approved):
Looks good to me, thanks!

PR #4105, opened 2022-08-10
Check database type, allow migration, fixes #4089, fixes #3923

### LINKED ISSUE #3923: `ddev start` could check to see if database type matches existing database type

When you change the database type in config.yaml, you may or may not be able to start the project. It tries to upgrade if possible, but `mariadb:10.3` to `mysql:8.0` or `postgres:14` will never ever work.

The previous version is actually stored in /var/lib/mysql/db_mariadb_version.txt, but it's hard to get at that (requires running a container with the volume mounted), so would slow down start. 

But the results when somebody does a change are pretty ugly, and they have to do a `ddev logs -s db` to figure out what happened.

But ddev could store away the previous version, for example in ~/.ddev/global_config.yaml like it does with `used_host_ports`:
```
  d9mysql:
    approot: /Users/rfay/workspace/d9mysql
    used_host_ports: ["9999"]
```

Above

Mounting the db volume and looking at the file.

### Additional context

_No response_

### LINKED ISSUE #4089: Provide easy technique to change database type/version

Currently if you want to make a significant change to database type/version, it requires an awkward dance. For example, if you want to change from MariaDB 10.4 to MySQL 8.0, you have to 

1. Export your database in a text format (`ddev export-db`)
2. `ddev delete` to destroy the old database
3. `ddev config --database=mysql:8.0` to change database type
4. `ddev start`
5. `ddev import-db` the exported sql.gz file. 

Worse than this is the fact that if you change the database type naively - if you just change it, you get a broken db container and a cryptic error message saying to look at the db logs. 

It would be great if we could:

1. Detect a database change without crashing like this
2. Provide a tool that does the above steps (`ddev debug change-database-type` ?)

The current frictional path doesn't harm very many people, but it harms them when they get blind-sided by it.

### Additional context

This problem is exactly the same as everywhere else in the database world. You can't just change from mariadb to mysql on the server without knowing what you're doing.

rfay:
## The Problem/Issue/Bug:

* #4089 
* #3923 

It's not a common occurrence, but people are really frustrated when they configure a database and then can't start it. 

* Check on start to see if the existing database matches the configuration
* Check on DB config change to see if the existing database matches the configuration
* `ddev debug check-db-match` checks to see if they match
* `ddev debug get-volume-db-version` gets the existing in-container in-volume value
* `ddev debug migrate-database` will attempt a database migration

- [x] Test changing config in .ddev/config.yaml and then `ddev start`. It should fail and complain with a good message
- [x] Try changing with `ddev config --database-version` and it should complain with a good message.
- [x] Check performance. I think some of these commands may be feeling too sluggish. Check through.
- [x] Test `ddev debug check-db-match`
- [x] Test `ddev debug get-volume-db-version`

* TestDdevAllDatabases has additional coverage.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4105"></a>

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -709,6 +709,13 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
 		return fmt.Errorf("failed to validate config: %v", err)
 	}
 
+	// If the database already exists in volume and is not of this type, then throw an error
+	if !nodeps.ArrayContainsString(app.GetOmittedContainers(), "db") {
+		if dbType, err := app.GetExistingDBType(); err != nil || (dbType != "" && dbType != app.Database.Type+":"+app.Database.Version) {
+			return fmt.Errorf("Unable to configure project %s with database type %s because that database type does not match the current actual database. Please change your database type back to %s and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=%s', and you can try a migration with 'ddev debug migrate-database %s' see docs at %s", app.Name, dbType, dbType, dbType, app.Database.Type+":"+app.Database.Version, "https://ddev.readthedocs.io/en/latest/users/extend/database_types/")
+		}
+	}
+
 	if err := app.WriteConfig(); err != nil {
 		return fmt.Errorf("could not write ddev config file %s: %v", app.ConfigPath, err)
 	}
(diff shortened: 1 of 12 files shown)


linawolf:
Since I updated ddev to v1.21.1 none of my projects starts. I get the error

``` Failed to start example: Unable to start project example because the configured database type does not match the current actual database. Please change your database type back to  and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=' and then you might want to try 'ddev debug migrate-database mariadb:10.2', see docs at https://ddev.readthedocs.io/en/latest/users/extend/database_types/```

I tried the new commands from this PR but they always gave me the error

```Failed to check existing DB type: could not split version string 10.2```

rfay:
Hi @linawolf - This problem and its solution are described both in the pinned issue at the top of the issues and in the release notes.

* https://github.com/drud/ddev/issues/4129
* https://github.com/drud/ddev/releases/tag/v1.21.1

PR #4106, opened 2022-08-11
Update platform integration documentation

bserem:
- Split one time config (cli_token) from per project config
- Improve formating (using double spaces) so that both mkdocs and github print new bullets on a new line

## The Problem/Issue/Bug:
https://ddev.readthedocs.io/en/latest/users/providers/platform/ is a bit hard to follow because of poorly rendered markdown list and the fact that global configuration is mixed with per-project config

image

This MR separates one time config and improves lists

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4106"></a>

### DIFF
--- a/docs/content/users/providers/platform.md
+++ b/docs/content/users/providers/platform.md
@@ -21,21 +21,25 @@ web_environment:
 ## Platform.sh per Project Configuration
 
 1. Check out the site from platform.sh and then configure it with `ddev config`. You'll want to use `ddev start` and make sure the basic functionality is working.
-2. Add PLATFORM_PROJECT and PLATFORM_ENVIRONMENT variables to your project `.ddev/config.yaml` or a `.ddev/config.*.yaml`:
+2. Add PLATFORM_PROJECT and PLATFORM_ENVIRONMENT variables to your project.
 
-```yaml
-web_environment:
-  - PLATFORM_PROJECT=nf4amudfn23biyourproject
-  - PLATFORM_ENVIRONMENT=main
-```
-or with
-```bash
-ddev config --web-environment-add="PLATFORM_PROJECT=nf4amudfn23bi,PLATFORM_ENVIRONMENT=main"
-```
+    * Either in `.ddev/config.yaml` or a `.ddev/config.*.yaml` file:
 
-4. `ddev restart`
-5. Run `ddev pull platform`. After you agree to the prompt, the current upstream database and files will be downloaded.
-6. Optionally use `ddev push platform` to push local files and database to platform.sh. Note that `ddev push` is a command that can potentially damage your production site, so this is not recommended.
+        ```yaml
+        web_environment:
+          - PLATFORM_PROJECT=nf4amudfn23biyourproject
+          - PLATFORM_ENVIRONMENT=main
+        ```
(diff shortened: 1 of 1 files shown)


bserem:
Screenshot from github with new layout:

image

rfay:
Wow thanks. You don't really have to go to that extent to do a docs PR :) You can just explain why and do it. But it's appreciated.

I'd love it if you'd try out and follow along with https://github.com/platformsh/ddev-platformsh as well. It's not mature yet, but if you're using platform I'd love to have feedback about it.

Also note that the tests automatically give a view URL. The one here is https://ddev--4106.org.readthedocs.build/en/4106/users/providers/platform/

You'll note that the numbering still doesn't seem to be quite right.

bserem:
Thanks for the tips!

Numbered-lists are messy in markdown, it's true. Github gets it right, mkdocs gets it wrong.
I'll have to take this locally to be able to preview mkdocs output right away and hope we get lucky.
We could use bullets though.

_As for https://github.com/platformsh/ddev-platformsh, only had a small glimpse a couple of days ago, not enough experience by my side to share info, but I am aware of its existence and I plan to provide feedback when I can. Time for ddev is limited sadly, I'm split between ddev and pre-existing tools in my team._

bserem:
Tried this locally, I can't get mkdocs to order the numbered-list properly while using code block(s) inside the list at the same time.

I see two options from here:
- use bullets
- leave a is, hoping mkdocs fixes this someday

@rfay it's your call I guess :)

rfay:
I think you just have to indent the code blocks. Thanks for chasing this!

bserem:
Yeah, indented code blocks work, but not if you use two of them together. Took me a minute to suspect it.
I split things, if anything I'd say it looks even better now:

image

docs/content/users/providers/platform.md
```diff
@@ -6,30 +6,40 @@ ddev provides integration with the [Platform.sh Website Management Platform](htt
...
-   - PLATFORM_ENVIRONMENT=main
- ```
-
-You can also do this with `ddev config --web-environment-add="PLATFORM_PROJECT=nf4amudfn23bi,PLATFORM_ENVIRONMENT=main"`
-4. `ddev restart`
-5. Run `ddev pull platform`. After you agree to the prompt, the current upstream database and files will be downloaded.
-6. Optionally use `ddev push platform` to push local files and database to platform.sh. Note that `ddev push` is a command that can potentially damage your production site, so this is not recommended.
+    * Or with a command from your terminal:
+
+        ```bash
```
rfay, line 36: This isn't working out quite right, it shows as `bash ddev config` - above I see you did a newline below the ```yaml, maybe that will solve it. Ir just use ``` and not ````bash

Thanks!
Platform_sh_Integration_-_DDEV_Docs

rfay (changes requested):
Almost there, thanks!

docs/content/users/providers/platform.md
bserem: There was one extra whitespace. I removed it

Thanks

rfay (approved):
Looks great, thank you so much.

PR #4118, opened 2022-08-13
Allow overriding config in `config.*.yaml` using `override_config: true`, fixes #4100, fixes #4079, fixes #4099

### LINKED ISSUE #4079: Override in config.*.yaml of false/null/empty values does not work anymore in v1.20.0

https://gist.github.com/cmuench/6c633b58d33ec9fcf802b6bc2dda551e

### Current Behavior

I have the NFS mount enabled in the projects `.ddev/config.yaml` file.
In a `.ddev/config.local.yaml` the setting is set to `false`.

If I start the project then the NFS mount is used.

A `ddev debug configyaml | grep nfs` prints `nfs_mount_enabled: true`.

I also tested with other values likes "timezone" or "composer_version". That values are overridden as expected.
My suggestion is that the NFS mount config has a different handling since v1.20.0

### Expected Behavior

If value in config.local.yaml file should override.
NFS mount should *not* be used in the project.

### Steps To Reproduce

1. Create project
2. Set "nfs_mount_enabled" to true in config.yaml
3. Create a config.local.yaml and set "nfs_mount_enabled" to false
4. Start the project
5. NFS mount is used

### Anything else?

It's currently only nfs_mount_enabled affected.

### LINKED ISSUE #4099: additional_hostnames could not be overwritten in config.local.yaml anymore

_No response_

### Current Behavior

Since the upgrade to ddev 1.20 additional_hostnames could not be overwritten in config.local.yaml anymore.
The router just respects  additional_hostnames in config.yaml

### Expected Behavior

if present in config.local.yaml , additional_hostnames have to be taken from that file

### Steps To Reproduce

_No response_

### Anything else?

_No response_

### LINKED ISSUE #4100: Provide feature flag for overriding config.*.yaml instead of merging

In our setup, we're overriding hooks with a limited set when we're running certain operations - e.g. we're not installing drupal. Since #3983 changes behaviour, we're no longer able to do it.

I would like to set up a (project) flag in `config.*.yaml` file that would indicate if I want to merge multiple config files or if those should be overriden.

One alternative would be to use `yq` or similar scripting solution, requiring additional tooling on the server and user machines. 

### Additional context

_No response_

rfay:
## The Problem/Issue/Bug:

* #4079 
* #4100 
* #4099 

DDEV v1.20.0 introduced the *merging* approach to `config.*.yaml` that most people seemed to expect over the years, but it turned out there were many people who had used the old (overwrite everything) behavior to good effect and needed the old behavior.

Add `override_config: true` option that can be added to any `config.*.yaml` file to signal that its contents should override instead of merging. 

Add `override_config: true` to any `config.*.yaml` and the behavior for *that file only* is changed to override instead of merge. 

Try out the things people complained about:
- [ ] `nfs_mount_enabled: false`
- [ ] `additional_hostnames: []`

* TestConfigMerge gets a whole new section that tests this.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4118"></a>


nhp:
Tested with `nfs_mounts_enabled` and works as expected.

rfay:
A related advance in docker-compose, https://github.com/compose-spec/compose-spec/pull/340

PR #4124, opened 2022-08-16
Remove support for docker-compose v1, Include 'name' in compose-yaml so PhpStorm doesn't have to specify it

rfay:
## The Problem/Issue/Bug:

DDEV has always specified the project name using COMPOSE_PROJECT_NAME, probably for historical reasons. But that means that other users of the `.ddev/.ddev-docker-compose-full.yaml` have to specify the COMPOSE_PROJECT_NAME, and it doesn't make sense to have to do this. The biggest issue is with PhpStorm, which tries to use the file on its own.

* Remove support for docker-compose v1
* add the top-level `name` specification to docker-compose files so it can be used without that environment variable.

- [ ] After this goes into a release, we can update the PhpStorm docs to say you don't have to do that.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4124"></a>


rfay:
It looks like we have to completely deprecate docker-compose v1 to do this. I think that's probably OK.

PR #4125, opened 2022-08-17
Fix bash completion installation docs

bserem:
## The Problem/Issue/Bug:
Bash without homebrew completion docs assumed you were in the running the commands from `/usr/bin` which is never the case.

Update the command so it can be run from anywhere and still install the proper file from everywhere.

Note: these docs might be good to be revisited now that there is an apt repo

### DIFF
--- a/docs/content/users/install/shell-completion.md
+++ b/docs/content/users/install/shell-completion.md
@@ -21,7 +21,7 @@ Shells like bash and zsh need help to do this though, they have to know what the
     If you're installing DDEV via homebrew, each new release will automatically get a refreshed completions script.
 
 === "Bash without Homebrew"
-    The completion script is exactly the same, it's just that you have to install it yourself. Each system may have a slightly different technique, and you'll need to figure it out. On Debian/Ubuntu, you would use [these instructions](https://www.cyberciti.biz/faq/add-bash-auto-completion-in-ubuntu-linux/) to enable bash-completion, and then `sudo mkdir -p /etc/bash_completion.d && sudo cp ddev_bash_completion.sh /etc/bash_completion.d`. This deploys the ddev_bash_completion.sh script where it needs to be. Again, every Linux distro has a different technique, and you may have to figure yours out.
+    The completion script is exactly the same, it's just that you have to install it yourself. Each system may have a slightly different technique, and you'll need to figure it out. On Debian/Ubuntu, you would use [these instructions](https://www.cyberciti.biz/faq/add-bash-auto-completion-in-ubuntu-linux/) to enable bash-completion, and then `sudo mkdir -p /etc/bash_completion.d && sudo cp /usr/bin/ddev_bash_completion.sh /etc/bash_completion.d`. This deploys the ddev_bash_completion.sh script where it needs to be. Again, every Linux distro has a different technique, and you may have to figure yours out.
 
 === "Zsh Completion with Homebrew"
 


rfay (changes requested):
Thanks for using, reading carefully, and providing a PR. 

I think the original docs are better... You're assuming that ddev_bash_completion.sh will be in /usr/bin, which it wouldn't be in any way I know of. The idea here is that you have downloaded https://github.com/drud/ddev/releases/download/v1.20.0/ddev_shell_completion_scripts.v1.20.0.tar.gz and untarred it, and have the ddev_bash_completion.sh in your current directory. Is the downloading of the tarball the missing piece?

bserem:
I've installed ddev using `apt install ddev` from the new debian repos. The `*_completion.sh` files are installed in `/usr/bin`, not by me, but by apt.

DPKG verifies that these files are provided by ddev package:
```
$ dpkg -S /usr/bin/ddev_bash_completion.sh 
ddev: /usr/bin/ddev_bash_completion.sh
```

We might need a section for apt installations as it seems

rfay:
But... if you install using apt, the completions are already installed ! I guess that should also be mentioned. You shouldn't have to be doing anything in that case.

bserem:
Apparently they were not. I'll try to see if I can do a re-install without messing up my system.
I'll come back to you tomorrow about this.
Thanks

rfay:
You can just `apt remove ddev` and perhaps `dpkg -P ddev`, then install it again. It should already have the completion script in the right place.

But the autocompletion absolutely should be installed. You do have to `apt install bash-completion` for it to work (debian/ubuntu).

Thanks for working on this!

bserem:
Gave this a try, completion is not installed automatically.
Removing with `apt` or `dpkg -P` made no different (I tried both, in different passes, only one is displayed below).

_output below truncated for readability_

**Checking bash-completion is installed:**
```
$ apt install bash-completion
bash-completion is already the newest version (1:2.11-6).
```

**Remove ddev:**
```
$ apt remove ddev
Removing ddev (1.20.0) ...
```

**Checking completion files after removal:**
```
$ ls /etc/bash_completion.d/
ddev_bash_completion.sh  git-prompt
```

**Remove them manually since they exist, to check if they are automatically installed:**
```
$ rm /etc/bash_completion.d/ddev_bash_completion.sh
```

**Re-install DDEV:**
```
$ apt install ddev
The following NEW packages will be installed:
  ddev
0 upgraded, 1 newly installed, 0 to remove and 18 not upgraded.

Selecting previously unselected package ddev.
(Reading database ... 334527 files and directories currently installed.)
Preparing to unpack .../archives/ddev_1.20.0_amd64.deb ...
Unpacking ddev (1.20.0) ...
Setting up ddev (1.20.0) ...
```

**Check completion files, they are not installed:**
```
$ ls /etc/bash_completion.d/
git-prompt
```

---

**Querying the file with dpkg also doesn't yield results:**
```
dpkg -S /etc/bash_completion.d/ddev_bash_completion.sh 
dpkg-query: no path found matching pattern /etc/bash_completion.d/ddev_bash_completion.sh
```

Are we sure it gets installed automatically? 
I am running debian `testing`, but this shouldn't be affecting this at all.

Some extra info:
```
apt version: 2.5.2 (amd64)
dpkg version: 1.21.9 (amd64)
```

rfay:
We may have a bug in the package builder (nfpm) so will have to chase that. I'll take a look. And no, I wouldn't think debian testing would be related.

rfay:
Asked how to do this right in https://github.com/goreleaser/goreleaser/discussions/3327

It definitely isn't currently being done, see https://github.com/drud/ddev/blob/badc85136b3cf586ebe2a59a9527cad776184f0d/.goreleaser.yml#L237-L253

rfay:
* https://github.com/drud/ddev/pull/4133
should solve the problem, so we'll close this in favor of #4133 if it does, but would love your review there.

PR #4128, opened 2022-08-18
Fix PhpStorm integration instructions, fixes #4083

### LINKED ISSUE #4083: [docs] Rewrite PhpStorm integration section

The longstanding PhpStorm issue is now fixed, https://youtrack.jetbrains.com/issue/WI-62463, and also the ddev-intellij plugin is available, https://github.com/php-perfect/ddev-intellij-plugin

That makes https://ddev.readthedocs.io/en/latest/users/topics/phpstorm/ much easier.

1. Recommend using the plugin, as it fixes so many things right off the bat.
2. Remove all the weird workarounds for specifying the remote php version.

Rewrite. Anybody can take this on. 

_No response_

### Additional context

_No response_

rfay:
## The Problem/Issue/Bug:

* #4083 

Rewrite, please review.

Walk through the instructions.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4128"></a>


docs/content/users/topics/phpstorm.md
```diff
@@ -2,51 +2,47 @@
...
+    4. In the "Path mappings" of the "Server" you may have to map the local paths (which on WSL2 means /home/...) to the in-container paths, especially if you have mutagen enabled. So "Virtual Machine Path" would be "/var/www/html" and "Local path" would be something like `/Users/rfay/workspace/d9` (on macOS) or `\\wsl$\Ubuntu\home\rfay\workspace\d9` on Windows using WSL2. (The plugin will already have done this for you.)
+    5. Now back in the "Configure Remote PHP Interpreter" for "Configuration files" use `.ddev/.ddev-docker-compose-full.yaml`. On macOS, you may need to use `<cmd><shift>.`, (Command+Shift+Dot) to show hidden dotfiles. (The plugin will have done this for you.)
+    6. Service: web (Plugin will have done this.)
+    7. Add an environment variable `COMPOSE_PROJECT_NAME=ddev-<projectname>`. In this case, it's `ddev-d9`. (Note that DDEV project names that contain dots do not currently work due to a [PhpStorm bug](https://youtrack.jetbrains.com/issue/WI-63293). You'll need to rename your project to get these instructions to work.) (Plugin will have done this.)
```
nico-loeber, line 27: Is setting the environment variable still required? I thought this would also be related to phpstorm using docker run instead of exec which should be fixed now.

docs/content/users/topics/phpstorm.md
```diff
@@ -2,51 +2,47 @@
...
-   1. set the bundled docker-compose binary as **Docker Compose executable** which can be found in `~/.ddev/bin`.
-   2. (Suggested for WSL) enabling **Use Compose V2** in PHPStorm and your Docker installation.
-- Any OS platform, but if you're using Windows PhpStorm with WSL2 the path mappings are slightly more complex. WSL2 instructions are provided where necessary.
+- PhpStorm 2022.2 or higher.
+- DDEV v1.20.1 or higher
 
 ### Setup Technique
 
-1. Open a DDEV project. In this example, the project name is "d9" and the site is "d9.ddev.site".
+1. Preferably, install the [DDEV Integration Plugin](https://plugins.jetbrains.com/plugin/18813-ddev-integration) so you don't have to do much of this.
```
nico-loeber, line 14: If you use the plugin you shouldn't need to do any of the following steps. So suggesting the plugin shoudln't be part of the list but an alternative to it.
This would also make the "(plugin will have done this)" redundant.

nico-loeber (commented):
The revision of the documentation looks really good to me. Great work!
As noted above I suggest to make more clear that the DDEV Integration Plugin is an alternative to the manual steps listed.
Please let me know if there is any step on this list that is not covered by the plugin yet.

rfay:
The thing that the plugin does not do is configuring phpunit. Not sure if it can do that.

docs/content/users/topics/phpstorm.md
rfay: The environment variable is not required in v1.21.1, so I guess with an update on the version at the top, it will be OK.

docs/content/users/topics/phpstorm.md
rfay: Well, except the phpunit.

I do think this can be much better with a separate section for phpunit.

rfay:
Thanks @nico-loeber - incorporated your comments. The one thing I've decided to add is using Jetbrains Gateway on WSL2. Tricky because of having to enable ssh.

Issue #2110, opened 2020-03-11
Explore GUI options for DDEV-Local


rfay:
Some people would definitely rather have a GUI to manipulate projects. A GUI can roll tasks into a more familiar and easier-to-remember set of tasks.

Many Go GUI options are listed in [awesome-go](https://awesome-go.com/#gui), these are among them. @andrewfrench had experimented with https://github.com/marcusolsson/tui-go, but it's mostly abandonware now.

A GUI presentation of DDEV-Local has long been a goal, but to date it hasn 't been fully achieved. [ddev-ui](https://github.com/drud/ddev-ui) was built as an Electron app which used `ddev list -j` to get date. Although it hasn't been maintained for a year and a half, it still seems to mostly work with current ddev v1.13.1. But it has no tests and has nobody to take care of it. It's Electron and mostly React.

I've recently experimented with a couple of GUI options simply to explore maintainable GUI options.

## Fyne cross-platform GUI

I experimented with Fyne and ddev in [branch](https://github.com/rfay/ddev/tree/20200309_fyne) and was able to get a table of projects that looks like this:

<img width="968" alt="DDEV-Local_and_rfay_rfay-drupal8-web___var_www_html_and_rfay_monitor___" src="https://user-images.githubusercontent.com/112444/76467283-e2dfd280-63ae-11ea-9fee-d22434aafa35.png">

The key code to get that going is in https://github.com/rfay/ddev/blob/eb69a7e07d9e925d2d73797fabba2a4f69ae277b/cmd/ddev/cmd/gui.go#L19-L99 and it's not tremendously difficult, although it requires jumping through some hoops. 

### Building with Fyne included

Building on macOS isn't very easy.

1. Obtain "develop" version of "fyne" tool as described in https://github.com/fyne-io/fyne/issues/157#issuecomment-597362440
2. `fyne vendor`
3. Buld with `go install ./pkg/... ./cmd/... ` instead of make; the current build-tools won't build this successfully

Note that CGO_ENABLED=1 is required

### Running the experimental branch

I just kept building and running with `go install ./pkg/... ./cmd/...  && ~/go/bin/ddev gui`

### Issues with Fyne

* Outputs complaints at startup on macOS
    When you start a Fyne app there are major deprecation warnings that are output to the screen. Issues suggest that this should be fixed in the next release of Fyne. OTOH the warnings are scary because they imply that Fyne might not work at all in a future version of macOS.

* Cross-compiling is work
[...]

PR #4137, opened 2022-08-21
ddev list: Truncate app list table text to avoid wrapping

cmuench:
## The Problem/Issue/Bug:

If I run `ddev list` in most of the cases the text is wrapped.
That's because the text of the path and the URL in the table is to long for my terminal window.
In most of the cases a user want to get an overview about the existing projects and the current status of
them.

Example screenshot:

image

The size of the *type* column was increased to prevent a wrap of longer project types than "drupal7". The project types "wordpress" and "magento2" are not wrapped anymore with this PR.

I moved the *status* column from the end of the table right after the name.
If the terminal width is not enough the text will now be truncated.

Screenshot:

image

For users who want to have the previous behavior the flag `-W` was added to force the old text wrapping.

Run `ddev list` and change terminal window size.

## Related Issue Link(s):

It's a temporary solution until we use one of the bigger libraries listed in #2110.

## Release/Deployment notes:

If someone parses the raw output of the `ddev list` command then maybe scripts must be changed due to the changed order of the columns.
The order of JSON rendered table is also changed. This should not provide any issues. The key is still the same.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4137"></a>

### DIFF
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -14,6 +14,9 @@ var activeOnly bool
 // continuousSleepTime is time to sleep between reads with --continuous
 var continuousSleepTime = 1
 
+// wrapListTable allow that the text in the table of ddev list wraps instead of cutting it to fit the terminal width
+var wrapListTableText bool
+
 // ListCmd represents the list command
 var ListCmd = &cobra.Command{
 	Use:   "list",
@@ -23,13 +26,14 @@ var ListCmd = &cobra.Command{
 ddev list --active-only
 ddev list -A`,
 	Run: func(cmd *cobra.Command, args []string) {
-		ddevapp.List(activeOnly, continuous, continuousSleepTime)
+		ddevapp.List(activeOnly, continuous, wrapListTableText, continuousSleepTime)
 	},
 }
 
 func init() {
 	ListCmd.Flags().BoolVarP(&activeOnly, "active-only", "A", false, "If set, only currently active projects will be displayed.")
 	ListCmd.Flags().BoolVarP(&continuous, "continuous", "", false, "If set, project information will be emitted until the command is stopped.")
+	ListCmd.Flags().BoolVarP(&wrapListTableText, "wrap-table", "W", false, "Display table with wrapped text if required.")
 	ListCmd.Flags().IntVarP(&continuousSleepTime, "continuous-sleep-interval", "I", 1, "Time in seconds between ddev list --continuous output lists.")
 
 	RootCmd.AddCommand(ListCmd)
--- a/cmd/ddev/cmd/list_test.go
+++ b/cmd/ddev/cmd/list_test.go
@@ -41,7 +41,7 @@ func TestCmdList(t *testing.T) {
 	assert.NoError(err)
 
 	// Execute "ddev list" and harvest plain text output.
-	out, err := exec.RunHostCommand(DdevBin, "list")
+	out, err := exec.RunHostCommand(DdevBin, "list", "-W")
 	assert.NoError(err, "error running ddev list: %v output=%s", out)
 
 	// Execute "ddev list -j" and harvest the json output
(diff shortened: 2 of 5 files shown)


rfay:
I've struggled with this a lot, turned off the truncation because I often need to copy the results (and nobody complained until now). I think the path is the thing I most often need, maybe that's true for you too. It's pretty brilliant that you experimented with changing the order of the columns.

Of course, don't forget that there are different styles in `ddev config global --table-styles` and also that `ddev config global --simple-formatting` turns off all the lines.

rfay:
What would you think about leaving the TYPE column out completely? It's usually location, URL, and status that I'm most interested in, perhaps name.

rfay:
Looks like TestCmdList is failing... but perhaps not showing the item at all?

tyler36:
`Status` column could be remove and replaced with a prefix before name. 
- green `O` for OK
- red `.` / `_`  / `-` / `x` for stopped.

This could save a lot of space.

rfay:
I like it @tyler36 - and the `ddev list --json` could still have all the same stuff in the same place.

cmuench:
@rfay @tyler36 All this options are valid but I think there are more and better options available.
As written it's only a temporary solution until we decide to use maybe a more complex CLI UI library.
IMHO the status column is important but does not cause the layout issue. The maximum length of the status column is defined. The issue are the columns where we do not know how long they can be (url and path).

At netz98 we wrote another CLI application in golang with a much more complex table. A colleague of me used https://github.com/rivo/tview as library to display the table. This package allows to scroll inside the table.
The package also allows interaction.

As you know I am the maintainer of n98-magerun which is also a CLI tool. There are also a lot of list commands in this tool. A solution there was to list only a small set of columns and allow to add additional data columns by an option.

Possible examples:

`ddev list -c path,url` or `ddev list -c path -c url` (I think the first example fits better to ddev)

This would provide then a way to list other values.

We can extend the command later-on to provide handy lists with other values.

`ddev list -c php_version,composer_version`

cmuench:
@rfay Test is fixed but the buildkite builds seem to be broken. Cannot see any dependency to this commits here.

rfay:
Looks like you've got it. Colima tests more than they succeed, re-ran them. The buildkite tests use Docker Desktop and it often fails.

cmuench:
> Looks like you've got it. Colima tests more than they succeed, re-ran them. The buildkite tests use Docker Desktop and it often fails.

Good to know. :-)

rfay (commented):
Thanks for this. It's a problem that needs some attention, and annoys me regularly, which is why I removed the truncation.

I do think this can be massively improved by using a simpler status indicator on the project name line, losing a whole column, as @tyler36 suggested, and how about moving TYPE to the  last column? That way we'd be getting rid of two columns essentially, and most of us would be able to see the name, location, and URL, at least where we don't have things too complicated. But those are the things I'm always copying.

cmuench:
@rfay What's up with the idea that we remove the type column and provide the option to add additional columns by a option?
My opinion is that we need truncation or a better wrapping algorithm.

rfay:
It's not hard to wrap, but it's nearly impossible to copy wrapped stuff. The library used to do this will in fact wrap, but you can't copy it.

Just trying to make this work more places for more people. And especially for you and me :)

rfay:
See https://github.com/drud/ddev/blob/e66448d80ed18b657d651851ece8b6ade11d9f69/pkg/globalconfig/styles.go, and the docs. I'm sure we can wrap, and I'm sure we can do this better, and so appreciate your exploration!

But just getting rid of a couple columns (or rather collapsing one and putting the other at the end where it may or may not be truncated) is a pretty easy path.

cmuench:
> See https://github.com/drud/ddev/blob/e66448d80ed18b657d651851ece8b6ade11d9f69/pkg/globalconfig/styles.go, and the docs. I'm sure we can wrap, and I'm sure we can do this better, and so appreciate your exploration!
> 
> But just getting rid of a couple columns (or rather collapsing one and putting the other at the end where it may or may not be truncated) is a pretty easy path.

Ok. I will then move the type column to the end. Let's see how that looks like.

cmuench:
Bigger terminal. 



Smaller terminal window with and without wrap option.

rfay (approved):
Thanks, this is ok with me, although I like all would like something even more effective in the long run. 
rfay_ddevprod____workspace_stories_grandcountyhistory_org


rfay_ddevprod____workspace_stories_grandcountyhistory_org_and_rfay_d9-web___var_www_html_and_ddev_list__Truncate_app_list_table_text_to_avoid_wrapping_by_cmuench_·_Pull_Request__4137_·_drud_ddev

github.com/jedib0t/go-pretty continues to mature, oh well :)

cmuench:
@rfay yes, this should only be a quick win. I will have a look what's be possible with the go-pretty library.

rfay:
Thanks so much!

PR #4157, opened 2022-08-31
[docs] Mention need to configure docker-compose for PhpStorm

rfay:
For full PhpStorm integration, people need a recent docker-compose v2, preferably the private one provided by DDEV.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4157"></a>


julienloizelet:
Hi,

I just saw your updated doc and I would like to mention that setting `~/.ddev/bin/docker-compose` did not work for me : I had to avoid the `~/` and set the "real path". In my case : `/home/julien/.ddev/bin/docker-compose`

rfay:
Yes, it doesn't accept the shell '`' notation, I'll update it to be clearer.

PR #4166, opened 2022-09-02
[docs] Add note about checking unused space when importing DBs [skip ci]

tyler36:
Saw the conversation on Discord and though it would be wise to drop a quick note in the docs.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4166"></a>


rfay (approved):
Thanks!

rfay:
BTW, you don't need `[skip ci]` on a docs PR any more because the test scripts check to see if it's only docs to be checked. But if you do want to skip all tests, it's `[skip ci]`, not `[skip tests]`.

Thanks!

PR #4176, opened 2022-09-04
Adds Craft CMS project type, docs, and tests closes #3594

### LINKED ISSUE #3594: Explicit Craft CMS support
I manage a lot of Craft CMS (and Bedrock/WP) sites. It would be nice if DDEV had a Craft recipe/support built in. 
(craftcms.com)

Built-in detection of and support for Craft sites. It's basically just a composer-managed LAMP app but it would be nice if existing .env files could be parsed (and new environment variables set or .env generated) for setup.

Building it myself. -- not super familiar with the process here or with golang, however. (Or with setting environment variables in the container.)

Craft also has their own local, containerized runtime called Nitro. But I'd prefer to standardize around a single local dev setup.

**Additional context**
Installation instructions: https://craftcms.com/docs/3.x/installation.html
Configuration: https://craftcms.com/docs/3.x/config/
Craft has a full-featured CLI: https://craftcms.com/docs/3.x/console-commands.html

It occurs to me that handling .env files and/or setting env variables in the container might be useful for Bedrock (WP) support as well – no need to modify wp-config in that case.

khalwat:
## The Problem/Issue/Bug:

There is no built-in Craft CMS project type, and no quickstart documentation for Craft CMS

This PR adds a Craft CMS project type, tests, and quickstart documentation for people working with Craft CMS

#### Composer project

Use this to create a new Craft CMS project from the official [Craft starter project](https://github.com/craftcms/craft) or a third-party starter project using Composer.

```bash
mkdir my-craft-project
cd my-craft-project
ddev config --project-type=craftcms
ddev composer create -y --no-scripts --no-install craftcms/craft
ddev start
ddev composer install
ddev craft setup/welcome
ddev launch
```

#### Manual download

Use this to create a new Craft CMS project using a zipped archive.

```bash
mkdir my-craft-project
cd my-craft-project
curl -LO https://craftcms.com/latest-v4.zip
unzip latest-v4.zip && rm latest-v4.zip
ddev config --project-type=craftcms
ddev start
ddev composer install
ddev craft setup
ddev launch
```

#### Craft git clone

Use this to migrate an existing Craft CMS project from a git repository and import a database dump.

```bash
git clone https://github.com/example/example-site my-craft-project
cd my-craft-project
ddev config --project-type=craftcms
ddev start
ddev composer install
ddev import-db
ddev launch
```

* Craft CMS webserver test
* Craft CMS commands test

## Related Issue Link(s):

https://github.com/drud/ddev/issues/3594

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4176"></a>


rfay:
I recommend using `ddev composer` for everything. Purists would *never* have composer or php on their local machine, so all the quickstarts use `ddev composer create` or `ddev composer install`. There are loads of problems with using local php/composer, including version and extension mismatches.

bencroker:
Yes, you're of course right @rfay, I've updated the quickstart docs to read as follows:

```sh
mkdir path/to/my-project
cd path/to/my-project
ddev config --project-type=craftcms
ddev start
ddev launch
```

rfay:
Sorry the full tests weren't being run. Forgot you were first-time contrib so they have to be authorized just during this PR. If you want to do the docs part first, we can get that one through faster and then (I think) it won't balk about running tests on this one.

khalwat:
No worries @rfay -- the docs are all set!

Let us know what we need to do from here.

rfay:
I was just saying you'll probably have a lot of tests work to do still, and you may need me to ping each push you do. Also remember to run the failing tests manually while debugging them so you don't have to wait for all the test runners.

rfay:
Oh, if you can't see the buildkite failures PM me your email and I'll add you to buildkite.

khalwat:
So near as I can tell, it's failing in `TestApptypeDetection` which I _believe_ is failing because we need to provide you with some kind of a test app?

If you could help us out with where to go from here, that'd be helpful. I have looked at:

https://ddev.readthedocs.io/en/stable/developers/project-types/

& 

https://github.com/drud/ddev/blob/master/docs/content/developers/building-contributing.md

But I'm a little unclear on what we do from here in terms of getting a test project built, delivered, etc.

bencroker:
I've started a PR to add a test site at https://github.com/khalwat/ddev/pull/11, but I think we need direction from @rfay on next steps.

rfay:
I don't think TestAppTypeDetection can work until you have a craftcms stanza in  https://github.com/khalwat/ddev/blob/777386662756585a8d2519e8e3a34a48d481d8db/pkg/ddevapp/ddevapp_test.go#L40

Because IsCraftCMS() can't succeed in https://github.com/khalwat/ddev/blob/1f62a43dd46ae3007b7eafc182046c51a3e17754/pkg/ddevapp/craftcms.go#L13-L15 without the craft command being there.

bencroker:
@rfay we have our test site at https://github.com/khalwat/ddev/blob/ed7cb298f279bf636c49ad12a0ab92810335c185/pkg/ddevapp/ddevapp_test.go#L238-L251 but it looks like `IsCraftCMS()` is still failing. Do you have any pointers for us?

Screenshot 2022-09-05 at 10 39 23

rfay:
Looks like you have
```
			SourceURL:                     "https://github.com/khalwat/ddev/files/9488101/craftcms-4.2.3.tar.gz",
			ArchiveInternalExtractionPath: "craftcms-4.2.3",
```

but there is no subdir in that tarball, no ArchiveInternalExtractionPath. the tarball starts at the root. 

Congrats, though, it certainly works with `ddev config` on the extracted archive! 

I do think the archive extraction should have failed on you because that subdir isn't in there.

This is a good time to get Goland going though (or vscode, but it's more work) so you can easily run a test and step through it.

bencroker:
I updated the source, ensuring that `craftcms-4.2.3` is a subdirectory in the tarball, but tests are still failing. Any other ideas?

khalwat:
We both have it up and running in Goland... it's a little opaque to me what's going on tho.

It _seems_ like what is happening is somehow, the `craftcms` project isn't in the right place, so it fails the check for project type `craftcms` when it looks for the `craft` command.

However, I've compared what Ben did with the archive files to some of the other project types that _do_ seem to pass, and I'm missing what's wrong.

khalwat:
Looks like it was just missing an empty `craft` file in `pkg/ddevapp/testdata/TestApptypeDetection/craftcms/`

Tests are running, so there may still yet be more errors, but at least that one is down...

bencroker:
Are you $#%@ kidding me?! Well, tests seem to be passing, so great catch!!

rfay:
You're awesome, will try to look tomorrow.

khalwat:
So @bencroker I did some sleuthing... you switched over to use `.zip` in the test app:

```go
		// 13: craftcms
		{
			Name:                          "TestPkgCraftCms",
			SourceURL:                     "https://github.com/khalwat/ddev/files/9497111/cms-4.2.3.zip",
			ArchiveInternalExtractionPath: "cms-4.2.3/",
			FilesZipballURL:               "https://github.com/khalwat/ddev/files/9497148/craftcms-4.2.3-files.zip",
			DBZipURL:                      "https://github.com/khalwat/ddev/files/9496399/craftcms-4.2.3-db.sql.zip",
			FullSiteTarballURL:            "",
			Type:                          nodeps.AppTypeCraftCms,
			Docroot:                       "web",
			Safe200URIWithExpectation:     testcommon.URIWithExpect{URI: "/test.html", Expect: "Thanks for testing Craft CMS"},
			DynamicURI:                    testcommon.URIWithExpect{URI: "/", Expect: "Thanks for installing Craft CMS"},
			FilesImageURI:                 "/happy-brad.jpg",
		},
```

But alas, it may allow for `.zip` for the `SourceURL` but it does not for other parts of the tests:

Files:

```go
		// Get files before start, as syncing can start immediately.
		if site.FilesTarballURL != "" {
			_, tarballPath, err := testcommon.GetCachedArchive(site.Name, "local-tarballs-files", "", site.FilesTarballURL)
			require.NoError(t, err)
			err = app.ImportFiles(tarballPath, "")
			assert.NoError(err)
		}
```

Database:

```go
		if site.DBTarURL != "" {
			_, cachedArchive, err := testcommon.GetCachedArchive(site.Name, site.Name+"_siteTarArchive", "", site.DBTarURL)
			if err != nil {
				assert.NoError(err)
				continue
			}
			err = app.ImportDB(cachedArchive, "", false, false, "db")
			if err != nil {
				assert.NoError(err, "failed to import-db with dbtarball %s, app.Type=%s, database=%s", site.DBTarURL, app.Type, app.Database.Type+":"+app.Database.Version)
				continue
			}
		}
```

...so the files were never being imported, and also the database was never being imported, causing the tests to fail.

So I think you'll need to recreate the `.tar.gz` archives for the files and database, and use the `FilesTarballURL` and `DBTarURL` respectively.

khalwat:
New archive files in `tar.gz` format:

[craftcms-4.2.3-files.tar.gz](https://github.com/drud/ddev/files/9501127/craftcms-4.2.3-files.tar.gz)

bencroker:
> But alas, it may allow for .zip for the SourceURL but it does not for other parts of the tests:

So you're saying that our testing is uncovering quirks with ddev's testing framework, rather than with our code ;)

Anyway, great to see you get to the bottom of this and get tests finally green!!

bencroker:
Ironically, Craft just got an update that requires us to change the `.env.example` to `.env.example.dev` 😆

khalwat:
We will have to check for both. Fortunately it is an easy change

cmd/ddev/cmd/commands_test.go
```diff
@@ -254,6 +254,17 @@ func TestCustomCommands(t *testing.T) {
 		assert.NoError(err, "expected to find command %s for app.Type=%s", c, app.Type)
 	}
 
+	// Craft CMS commands should only be available for type craftcms
+	app.Type = nodeps.AppTypeCraftCms
+	_ = app.WriteConfig()
+	_, _ = exec.RunHostCommand(DdevBin)
+	err = app.MutagenSyncFlush()
+	assert.NoError(err)
+	for _, c := range []string{"craft"} {
```
rfay, line 263: Interesting strategy! Not entirely sure this is the place to solve this, but doesn't seem to violate any particular strategies.

docs/content/users/quickstart.md
```diff
@@ -342,6 +342,76 @@ Here's a quickstart instructions for a number of different environments:
...
+        ### Composer build
+        
+        Use this to create a new Craft CMS project from the official [Craft starter project](https://github.com/craftcms/craft) or a third-party starter project using Composer.
+        
+        ```bash
+        mkdir path/to/my-project
+        cd path/to/my-project
+        composer create craftcms/craft . --no-install --no-scripts
+        ddev config --project-type=craftcms
+        ddev start
```
rfay, line 362:
```suggestion
        ddev config --project-type=craftcms
        ddev composer create craftcms/craft  --no-install --no-scripts
        ddev composer install
```

docs/content/users/quickstart.md
```diff
@@ -342,6 +342,76 @@ Here's a quickstart instructions for a number of different environments:
...
+        ```
+        
+        ### Git clone
+        
+        Use this to create a new Craft CMS project from the official [Craft starter project](https://github.com/craftcms/craft) or a third-party starter project using Git.
+        
+        ```bash
+        mkdir path/to/my-project
+        cd path/to/my-project
+        git clone https://github.com/craftcms/craft . && rm -rf .git*
```
rfay, line 375: Why remove .git?

docs/content/users/quickstart.md
```diff
@@ -342,6 +342,76 @@ Here's a quickstart instructions for a number of different environments:
...
+        
+        Use this to migrate an existing Craft CMS project from a git repository and import a database dump.
+        
+        ```bash
+        git clone https://github.com/example/example-site path/to/my-project
+        cd path/to/my-project
+        ddev config --project-type=craftcms
+        ddev start
+        ddev composer install
+        ddev import-db
```
rfay, line 411: Although this works without flags, people seem to be confused by the behavior and an explicit example is probably better.
```suggestion
        ddev import-db --src=/path/to/db.sql.gz
```

pkg/ddevapp/.env
```diff
@@ -0,0 +1,23 @@
+# The application ID used to to uniquely store session and cache data, mutex locks, and more
+CRAFT_APP_ID=CraftCMS--dfd9f530-e7cb-46c3-ab7f-cea4c8d09e3a
```
rfay, line 2: Is this random data or does it have some meaning? If it has meaning, should have a link to explain it?

pkg/ddevapp/.env
```diff
@@ -0,0 +1,23 @@
+# The application ID used to to uniquely store session and cache data, mutex locks, and more
```
rfay, line 1: I'm quite sure this file is in the wrong place, please remove it or put it where it belongs.

pkg/ddevapp/apptypes.go
```diff
@@ -81,23 +85,27 @@ func init() {
...
+		},
+		nodeps.AppTypeShopware6: {
+			appTypeDetect: isShopware6App, apptypeSettingsPaths: setShopware6SiteSettingsPaths, uploadDir: getShopwareUploadDir, configOverrideAction: nil, postStartAction: shopware6PostStartAction, importFilesAction: shopware6ImportFilesAction,
+		},
+		nodeps.AppTypeTYPO3: {
+			settingsCreator: createTypo3SettingsFile, uploadDir: getTypo3UploadDir, hookDefaultComments: getTypo3Hooks, apptypeSettingsPaths: setTypo3SiteSettingsPaths, appTypeDetect: isTypo3App, postImportDBAction: nil, configOverrideAction: nil, postConfigAction: nil, postStartAction: nil, importFilesAction: typo3ImportFilesAction,
+		},
+		nodeps.AppTypeWordPress: {
+			settingsCreator: createWordpressSettingsFile, uploadDir: getWordpressUploadDir, hookDefaultComments: getWordpressHooks, apptypeSettingsPaths: setWordpressSiteSettingsPaths, appTypeDetect: isWordpressApp, postImportDBAction: nil, configOverrideAction: nil, postConfigAction: nil, postStartAction: nil, importFilesAction: wordpressImportFilesAction,
+		},
```
rfay, line 108: Just assure me that nothing was changed when alphabetizing :)

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/drud/ddev/pkg/util"
+	"github.com/pkg/errors"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+// isCraftCmsApp returns true if the app is of type craftcms
+func isCraftCmsApp(app *DdevApp) bool {
+	return fileutil.FileExists(filepath.Join(app.AppRoot, "craft"))
```
rfay, line 16: This seems slightly suspicious as a fundamental detector, but you're the Craft expert(s). Completely valid? Any idea why the craft script is placed in the app root instead of some place like bin or vendor/bin ?

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+	if app.Docroot == "" {
+		app.Docroot = "web"
+	}
+
+	return nil
+}
+
+// Returns the upload directory for importing files, if not already set
+func getCraftCmsUploadDir(app *DdevApp) string {
+	app.UploadDir = "files"
```
rfay, line 30: I assume this needs some comments explaining Craft behavior here because this really isn't the default is it? I thought there was no default?

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+// craftCmsImportFilesAction defines the workflow for importing project files.
+func craftCmsImportFilesAction(app *DdevApp, importPath, extPath string) error {
+	if app.UploadDir == "" {
+		return errors.Errorf("No upload_dir is set for this (craftcms) project")
+	}
+	destPath := app.GetHostUploadDirFullPath()
+
+	// parent of destination dir should exist
+	if !fileutil.FileExists(filepath.Dir(destPath)) {
+		return fmt.Errorf("unable to import to %s: parent directory does not exist", destPath)
```
rfay, line 44: Do you think it's too dangerous to just create the directory? DDEV creates some settings directories if they don't exist, and it hasn't been a problem.

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	// parent of destination dir should be writable.
+	if err := os.Chmod(filepath.Dir(destPath), 0755); err != nil {
```
rfay, line 48: IMO we should set the 0755 here, or try. This is a common problem, and it's worth helping people out by working around it.

rfay: Oh, and it won't work on Windows, so needs some extra care. Take a look at Drupal's handling of settings.php and sites/default

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+
+	// If the destination path exists, remove it as was warned
+	if fileutil.FileExists(destPath) {
+		if err := os.RemoveAll(destPath); err != nil {
```
rfay, line 54: This, of course, is a possible disaster. Have we done adequate vetting of destPath here?

pkg/ddevapp/craftcms.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return nil
+}
+
+// Set up the .env file for ddev
+func craftCmsPostConfigAction(app *DdevApp) error {
+	var err error
+	var envFilePath string
+	envFilePath = filepath.Join(app.AppRoot, ".env")
```
rfay, line 87: Shouldn't this be done in the settingsCreator, not PostConfigAction?

pkg/ddevapp/ddevapp_test.go
```diff
@@ -235,6 +235,20 @@ var (
 			DynamicURI:                    testcommon.URIWithExpect{URI: "/", Expect: "This is a test page"},
 			FilesImageURI:                 "/sites/default/files/2022-07/Logo.png",
 		},
+		// 13: craftcms
+		{
+			Name:                          "TestPkgCraftCms",
+			SourceURL:                     "https://github.com/khalwat/ddev/files/9497111/cms-4.2.3.zip",
```
rfay, line 241: We have to remember to move these over to the official junk testing release. Can do now.

pkg/ddevapp/webserver_config_assets/nginx-site-craftcms.conf
```diff
@@ -0,0 +1,81 @@
+# ddev craftcms config
```
rfay, line 1: Was this copied from the PHP type? Do we even need to do it? (php type is used by default if this doesn't exist)

rfay (changes requested):
It's looking great. I haven't tried it out with manual testing yet, but wanted to give the first-runthrough responses to the code so we could discuss.

One thing I'm concerned about and hope to do better here: the .env file management in Laravel is a mess, and ddev doesn't properly manage the file (or try to manage it) and that causes confusion like https://github.com/drud/ddev/issues/3636 - I hope we're doing better here, and would like to do better in general.

Congrats on getting all the tests green! I'll move the test tarballs over to the ddev test repo now.

rfay:
The test tarballs are now at https://github.com/drud/ddev_test_tarballs/releases/tag/v1.1 but I changed the cms-4.2.3.zip to craft-cms-4.2.3.zip

pkg/ddevapp/ddevapp_test.go
rfay, line 241: Please add a comment about how to recreate this zipball for a new version of craft.

docs/content/users/quickstart.md
khalwat: So we looked at doing it this way, but decided against it. Because if we do it this way, the `.env` and `.env.example` and `.env.example.dev` will not be present, so `ddev config --project-type=craftcms` will do nothing in terms of being able to automatically update the `.env` file for you.

Recommend not changing this.

docs/content/users/quickstart.md
khalwat: Because this is starter template, you don't want the git repo info hanging around. This is essentially what `composer create-project` does: it clones the repo down, and then removes the `.git` dir

docs/content/users/quickstart.md
khalwat: Seems reasonable

pkg/ddevapp/ddevapp_test.go
khalwat: Go ahead and update the URLs to where you have moved them to if you like.

pkg/ddevapp/webserver_config_assets/nginx-site-craftcms.conf
khalwat: Yes, it was copied, and then some additional changes were added (relatively minor, but nice to have). Also this way we have a default file that if we do need additional customizations, it's already there.

pkg/ddevapp/ddevapp_test.go
khalwat: @bencroker you mind taking care of that?

pkg/ddevapp/.env
khalwat: It's just random data, people who use Craft will know what it means. It's not a secret, it just needs to be the same across environments. It's used to hash some data, for example.

pkg/ddevapp/craftcms.go
khalwat: This is a copy & paste from the current generic PHP setup so I'm assuming it's okay? If not, the generic PHP project, and others that have copy/pasted that code block should probably be changed too

pkg/ddevapp/craftcms.go
khalwat: This is a copy & paste from the current generic PHP setup so I'm assuming it's okay? If not, the generic PHP project, and others that have copy/pasted that code block should probably be changed too

pkg/ddevapp/craftcms.go
khalwat: Probably not, because I think `settingsCreator` wants a file returned to it? In our case, it's the `.env` file, and not a settings file we want DDEV to do anything with or manage in any way

pkg/ddevapp/.env
khalwat: Ah yes, that got checked in somehow. Removed.

pkg/ddevapp/apptypes.go
khalwat: Nothing was changed, just alphabetized it, and made the coding style consistent.

pkg/ddevapp/craftcms.go
khalwat: That's where it goes, it is the `cli` for Craft. It's at the root, I presume, to make it convenient to run from the project root

pkg/ddevapp/craftcms.go
khalwat: There is no default, but if we don't provide this, tests do not work for `import-files` -- so it seemed as good as anything. I understand you have a potential FR in the works that may improve on DDEV's assumption of a singleton upload directory, which would help us too

pkg/ddevapp/craftcms.go
khalwat: This is a copy & paste from the current generic PHP setup so I'm assuming it's okay? If not, the generic PHP project, and others that have copy/pasted that code block should probably be changed too

pkg/ddevapp/ddevapp_test.go
bencroker: Will do.

bencroker: This is complete.

bencroker: This is complete.

docs/content/users/quickstart.md
rfay: This will have to be sorted out. Composer and PHP are not required to use ddev.

rfay (changes requested):
I seem to have had some unsent comments... here they come.

khalwat:
This PR is waiting on the addition of a project type action that executes after `ddev composer create`, so that we can rewrite the `.env` or `env.example` / `env.example.dev` then, rather than inside of the `configOverrideAction: craftCmsConfigOverrideAction` as we do now.

@rfay has said he's going to look into adding such a project type action.

This will allow us to greatly simplify the quickstart documentation, and will also allow for the resolution of:

https://github.com/drud/ddev/issues/3636 & https://github.com/drud/ddev/issues/3590

pkg/ddevapp/ddevapp_test.go
```diff
@@ -235,6 +235,20 @@ var (
...
+			ArchiveInternalExtractionPath: "cms-4.2.3/",
+			FilesTarballURL:               "https://github.com/drud/ddev/files/9501156/craftcms-4.2.3-files.tar.gz",
```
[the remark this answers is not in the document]
khalwat: @bencroker I think did this already? If not, can you, Ben?

pkg/ddevapp/ddevapp_test.go
```diff
@@ -235,6 +235,20 @@ var (
...
+			DBTarURL:                      "https://github.com/drud/ddev/files/9501131/craftcms-4.2.3-db.sql.tar.gz",
```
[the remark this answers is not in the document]
khalwat: @bencroker I think did this already? If not, can you, Ben?

pkg/ddevapp/ddevapp_test.go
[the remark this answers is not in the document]
bencroker: Yes, I did.
https://github.com/drud/ddev/blob/91732b471750ac20ba4505874cc530e5fdca380f/pkg/ddevapp/ddevapp_test.go#L238-L247

pkg/ddevapp/ddevapp_test.go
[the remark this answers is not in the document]
bencroker: Yes, I did.
https://github.com/drud/ddev/blob/91732b471750ac20ba4505874cc530e5fdca380f/pkg/ddevapp/ddevapp_test.go#L238-L247

khalwat:
Strike that comment above, after a discussion with @rfay I was able to refactor the code so the setup flow is much cleaner. We may not need the additional project action at all.

rfay (approved):
I manually tested the composer and download options from the docs and it worked flawlessly. 

The .env management is doing fine, and the regex approach should be OK. 

Thanks for a great contribution, and let me know when you're ready to have it pulled.

khalwat:
Thx for all your help @rfay -- I just push some very minor docs changes to clean up the formatting.

But we are _ready_. Let's do this!

rfay:
Congrats, great work!

PR #4179, opened 2022-09-05
[Homebrew] Fix zsh completion file name

voronkovich:
## The Problem/Issue/Bug:

ZSH completion doesn't work if DDEV was installed by Homebrew. It happens because a completion file has a wrong name: `ddev` instead of `_ddev`.

The file was renamed to `_ddev`

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4179"></a>


rfay (approved):
Awesome, thanks.

rfay:
Should work with next release. `brew install --HEAD ddev` doesn't use this goreleaser code.

PR #4184, opened 2022-09-06
fix: no need for SSH when using terminus token

shaal:
## The Problem/Issue/Bug:

Terminus token allow users to connect to Pantheon without SSH.

This PR remove the check for SSH, and update documentation to reflect that change.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4184"></a>

### DIFF
--- a/docs/content/users/basics/developer-tools.md
+++ b/docs/content/users/basics/developer-tools.md
@@ -116,4 +116,4 @@ To use terminus, you'll first need to:
 2. Use `ddev ssh` to tunnel into your container
 3. Issue a command using the keyword `terminus`. For help using terminus try `terminus list` to get a list of possible commands.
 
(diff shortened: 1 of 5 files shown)


rfay (changes requested):
Sadly, this fails TestPantheonPush, can you take a look at that? https://app.circleci.com/pipelines/github/drud/ddev/7483/workflows/cd995e87-a7fa-4c5a-ab61-0678f92b17cf/jobs/39568

shaal:
Is seems that for `ddev pull` there's no need for SSH, but for `remote:drush` SSH is required.

rfay:
Would be cool if there's a better way!

rfay:
Closing, but sure open to a solution that works to accomplish this need.

One thing that you could do to pursue this is to put the ssh check into the push section instead of the auth section. Then you'd be happy. 

Why don't you give it a try?

Also, a technique that doesn't require drush rsync would be excellent.

Issue #4181, opened 2022-09-05
[output] Make the output on the first start more streamlined in the root.go 


rpkoller:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

At the moment the output after the installation of a new version of DDEV is the following `/cmd/ddev/cmd/root.go`
```
Congratulations, you seem to have a new DDEV version.
Please note that the global ~/.mutagen folder is no longer used by DDEV so you can delete it if it's no longer used by other applications.
It looks like you have a new DDEV version. During an upgrade it's important to `ddev poweroff`. May I do `ddev poweroff` before continuing? This does no harm and loses no data.
```
Problem is for one that you have "congratulations you have a new version of DDEV" in the first sentence and in the third sentence you have more or less the same again. the second sentence could be removed imho. 
the second sentence could be broken up into two. the first part should be a plain explanatory statement that `~/.mutagen` isn't used by DDEV any longer. the second part is changed into a clear actionable item in case the folder isn't used by any other application. 

### Describe the solution you'd like

I would suggest the following draft: 
```
Congratulations, you seem to have a new DDEV version.
Please note that the global ~/.mutagen folder is no longer used by DDEV. Delete the folder if it is no longer used by any other application.
During an upgrade DDEV has to be shut down. May I do `ddev poweroff` to be on the safe side before continuing? This causes no harm or data loss.
```
I'll try to open up a pull request later today to test the fork/branch process via the terminal.

rfay:
If you're still interested in this, a new PR is welcome. Otherwise, let's close it about 1 August 2023.

rfay:
Closing, almost a year late :)

PR #4195, opened 2022-09-08
[output] Improve the first start output in root.go

rpkoller:
Updates to the strings for output.UserOut.Print and util.Confirm

## The Problem/Issue/Bug:
PR for #4181 
## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4195"></a>

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -212,9 +212,9 @@ func checkDdevVersionAndOptInInstrumentation(skipConfirmation bool) error {
 			output.UserOut.Print("Congratulations, you seem to have a new DDEV version.")
 			home, _ := homedir.Dir()
 			if globalconfig.DdevGlobalConfig.MutagenEnabledGlobal && fileutil.IsDirectory(filepath.Join(home, ".mutagen")) {
-				output.UserOut.Print("Please note that the global ~/.mutagen folder is no longer used by DDEV so you can delete it if it's no longer used by other applications")
+				output.UserOut.Print("Please note that the global ~/.mutagen folder is no longer used by DDEV. Delete the folder if it is no longer used by any other application.")
 			}
-			okPoweroff := util.Confirm("It looks like you have a new DDEV version. During an upgrade it's important to `ddev poweroff`. May I do `ddev poweroff` before continuing? This does no harm and loses no data.")
+			okPoweroff := util.Confirm("During an upgrade DDEV has to be shut down. May I do a `ddev poweroff` to be on the safe side before continuing? This causes no harm or data loss.")
 			if okPoweroff {
 				ddevapp.PowerOff()
 			}


rfay:
Let me know when you've manually tested and you're happy with it, thanks.

rpkoller:
i forgot to test when i've created the change and the PR directly in the github ui. apologies. but one question. i'Ve downloded the arm64 artifact from 5 days ago in this issue. did a 'brew unlink ddev' then deleted the ddev file in '/opt/homebrew/bin' and then copied the downloaded ddev file into the bin folder. but if i do a `$> ddev --version` i get a `zsh: permission denied: ddev` . i thought i dont need to set the permissions for the file. on my old mac when i manually copied the new releases i never had to change the permissions? thats the only detail what confuses me

rpkoller:
hm manually changed the permissions to '755' with chmod. but when i try now 'ddev --version' i get that? 
Screen Shot 2022-09-14 at 22 44 36

rfay:
Right-click and open and give it permission.

rpkoller:
Screen Shot 2022-09-14 at 23 06 21
got it working now. thanks. and the output looks ok. also ok for a native speaker or would you rephrase anything?

cmd/ddev/cmd/root.go
```diff
@@ -212,9 +212,9 @@ func checkDdevVersionAndOptInInstrumentation(skipConfirmation bool) error {
 			output.UserOut.Print("Congratulations, you seem to have a new DDEV version.")
 			home, _ := homedir.Dir()
 			if globalconfig.DdevGlobalConfig.MutagenEnabledGlobal && fileutil.IsDirectory(filepath.Join(home, ".mutagen")) {
-				output.UserOut.Print("Please note that the global ~/.mutagen folder is no longer used by DDEV so you can delete it if it's no longer used by other applications.")
+				output.UserOut.Print("Please note that the global ~/.mutagen folder is no longer used by DDEV. Delete the folder if it is no longer used by any other application.")
```
rfay, line 215: I apologize, but in use I've found this to be really annoying, and think we should remove it.

cmd/ddev/cmd/root.go
```diff
@@ -212,9 +212,9 @@ func checkDdevVersionAndOptInInstrumentation(skipConfirmation bool) error {
...
 			}
-			okPoweroff := util.Confirm("It looks like you have a new DDEV version. During an upgrade it's important to `ddev poweroff`. May I do `ddev poweroff` before continuing? This does no harm and loses no data.")
+			okPoweroff := util.Confirm("During an upgrade DDEV has to be shut down. May I do a `ddev poweroff` to be on the safe side before continuing? This causes no harm or data loss.")
```
rfay, line 217: I'm afraid I don't think this is useful after all. Thanks for your thoughtfulness in adding it, but IMO the original message is better. We want people to do the poweroff because otherwise they may get an inappropriate ddev-router.

rfay (changes requested):

rfay:
I do apologize. Had some work to do on the conflicts rebase and didn't like it in the end. Thanks so much for the thoughtfulness of taking this on and working through it. Sorry it doesn't seem quite right.

PR #4199, opened 2022-09-09
Add `craft` command transition

bencroker:
## The Problem/Issue/Bug:

Once https://github.com/drud/ddev/pull/4176 is merged, existing Craft projects that have the project type set to `php` will no longer have the `craft` command available to them and will output an error.

Screenshot 2022-09-09 at 19 36 06 2

This PR maintains the `craft` command in the `php` project type, but prompts users to switch to the `craftcms` project type in order to use it, easing the transition for existing users who update DDEV without reading the release notes.

This is primarily to aid in transitioning and the `craft` command could be dropped for`php` project types in the next minor version, if desirable.

Screenshot 2022-09-09 at 19 36 06

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4199"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/craft
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/craft
@@ -10,5 +10,8 @@
 if grep -q "type: php" ./.ddev/config.yaml; then
     echo "The craft command is only available in the craftcms project type. You can update this in your project's config file, followed by restarting the DDEV project."
 else
+    CRAFT_CMD_ROOT=${CRAFT_CMD_ROOT:="./"}
+    
+    cd "${CRAFT_CMD_ROOT}"
     php ./craft "$@"
 fi
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/craft
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/craft
@@ -3,8 +3,12 @@
 #ddev-generated
 ## Description: Run Craft CMS command inside the web container
 ## Usage: craft [flags] [args]
-## Example: "ddev craft db/backup" or "craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
-## ProjectTypes: php
+## Example: "ddev craft db/backup" or "ddev craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
+## ProjectTypes: craftcms,php
 ## ExecRaw: true
 
-php ./craft "$@"
+if grep -q "type: php" ./.ddev/config.yaml; then
+    echo "The craft command is only available in the craftcms project type. You can update this in your project's config file, followed by restarting the DDEV project."
+else
+    php ./craft "$@"
+fi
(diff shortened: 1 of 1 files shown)


brandonkelly:
Would be great to have this, at least temporarily!

khalwat:
While I realize this would still be polluting the `php` command namespace, we're doing that already... and this would ease the transition.

We could set a trigger to remove it in a year or something, once the transition should be mostly complete.

I will also add something to the docs to tell people how to manually change the project type.

rfay:
Another option is just to let it work on php for a release or two, perhaps with a pester when you run it?

bencroker:
> Another option is just to let it work on php for a release or two, perhaps with a pester when you run it?

My preference would be to force people to switch project types. Pestering often goes ignored and once we finally do remove it, there may be confusion for those that simply ignored the message.

rfay:
Plesae rebase too, thanks.

bencroker:
> Plesae rebase too, thanks.

Rebased and ready to merge.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
```diff
@@ -3,11 +3,15 @@
...
-## ProjectTypes: craftcms
+## Example: "ddev craft db/backup" or "ddev craft db/backup ./my-backups" (see https://craftcms.com/docs/4.x/console-commands.html)
+## ProjectTypes: craftcms,php
 ## ExecRaw: true
 
-CRAFT_CMD_ROOT=${CRAFT_CMD_ROOT:="./"}
-
-cd "${CRAFT_CMD_ROOT}"
-php ./craft "$@"
+if grep -q "type: php" ./.ddev/config.yaml; then
```
rfay, line 10:
```suggestion
if [ "${DDEV_PROJECT_TYPE}" != "craftcms" ]; then
```

rfay: I didn't manually test this, so please give it a quick manual review, but think it's better.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
```diff
@@ -3,11 +3,15 @@
...
+    echo "The craft command is only available in the craftcms project type. You can update this in your project's config file, followed by restarting the DDEV project."
+else
+    CRAFT_CMD_ROOT=${CRAFT_CMD_ROOT:="./"}
+    
+    cd "${CRAFT_CMD_ROOT}"
+    php ./craft "$@"
```
rfay, line 16: Just a note that `php ./craft` is pretty much redundant and `php craft` is the same. But that's not important. I do keep trying to understand if craft may be somewhere else, in which case we could consider using `## HostWorkingDir: true` to run it in the same relative directory they're in. I don't think there's any reason to fiddle with this right now, jut talking out loud.

rfay (changes requested):
Thanks for this, and thanks for figuring out how to make things better for the Craft community.

pkg/ddevapp/global_dotddev_assets/commands/web/craft
khalwat: Yeah the check for the environment variable is likely better, I agree. I just didn't know it existed.

bencroker: Updated, thanks for the review!

rfay (approved):
Thanks for this!

rfay:
Manually tested this and it seems to be fine. Thanks!

PR #4218, opened 2022-09-21
Implementation of global project-tld config option

bbrala:
## The Problem/Issue/Bug:
I want to set a global project tld by default to support a custom TLD for out company.

Adds a config global config option for project-tld that overwrites the default for the project and makes it globally settable.

Build, use --project-tld globally and see it work. You can also set it in the global config yaml.

Added the option in the tests, but can't run tests locally, so thats a little bit hard.

I did test this manually by setting the global config in yml and observing our actual project tld be used! Hopefully i can use gitpod to run the tests.

## Related Issue Link(s):
https://github.com/drud/ddev/issues/2746


bbrala:
Hmm, the tests were green, then pushed the whitespace fix, seems a unrelated issue?

bbrala:
See: https://github.com/drud/ddev/runs/8480130526

rfay:
The github tests don't even run because you're first-time contributor, so ping me when I need to run them again. Just started them. Congrats on the PR! After you get this PR in then you don't have to have tests run manually any more.

bbrala:
Yeah seems i was mistaken, there was actually a declaration that was removed for some reason. Guessing my infamiliarity with go.

rfay:
Started tests.

cmd/ddev/cmd/config-global.go
```diff
@@ -159,6 +159,11 @@ func handleGlobalConfig(cmd *cobra.Command, args []string) {
 		globalconfig.DdevGlobalConfig.RequiredDockerComposeVersion = val
 		dirty = true
 	}
+	if cmd.Flag("project-tdl").Changed {
```
JaZo, line 162: There's a typo here: `project-tdl` > `project-tld`

cmd/ddev/cmd/config-global.go
```diff
@@ -202,6 +207,7 @@ func handleGlobalConfig(cmd *cobra.Command, args []string) {
 	output.UserOut.Printf("required-docker-compose-version=%v", globalconfig.DdevGlobalConfig.RequiredDockerComposeVersion)
 	output.UserOut.Printf("use-docker-compose-from-path=%v", globalconfig.DdevGlobalConfig.UseDockerComposeFromPath)
 	output.UserOut.Printf("no-bind-mounts=%v", globalconfig.DdevGlobalConfig.NoBindMounts)
+	output.UserOut.Printf("project-tdl=%v", globalconfig.DdevGlobalConfig.ProjectTldGlobal)
```
JaZo, line 210: Same typo.

JaZo (changes requested):

cmd/ddev/cmd/config-global.go
bbrala: Lovely. Thanks for checking this out.

bbrala: Fixed typo

cmd/ddev/cmd/config-global.go
bbrala: Fixed typo

bbrala:
the fail seems unrelated to the code change:

```
get_test.go:47: 
        	Error Trace:	get_test.go:47
        	            	exit status 1
        	Test:       	TestCmdGet
        	Messages:   	failed ddev get --list: exit status 1 (Failed to list available add-ons: Unable to get list of available services: GET https://api.github.com/search/repositories?q=topic:ddev-get+fork:true+org:drud: 403 You have exceeded a secondary rate limit. Please wait a few minutes before you try again. [] rateinfo=github.Rate{Limit:30, Remaining:28, Reset:github.Timestamp{2022-09-22 13:56:34 +0000 UTC}}
        	            	)
    get_test.go:48: 
        	Error Trace:	get_test.go:48
        	Error:      	"\x1b[31mFailed to list available add-ons: Unable to get list of available services: GET https://api.github.com/search/repositories?q=topic:ddev-get+fork:true+org:drud: 403 You have exceeded a secondary rate limit. Please wait a few minutes before you try again. [] rateinfo=github.Rate{Limit:30, Remaining:28, Reset:github.Timestamp{2022-09-22 13:56:34 +0000 UTC}}\x1b[0m\n" does not contain "drud/ddev-memcached"
        	Test:       	TestCmdGet
```

rfay (changes requested):
Perfect, tested manually and it did great. 

It still needs docs updated. I see references in https://ddev.readthedocs.io/en/latest/users/details/offline-usage/ and https://ddev.readthedocs.io/en/latest/users/configuration/config_yaml/ - but. it needs to be added to the global config on that page too. 

Also needs to be added to global_config.yaml stuff in https://github.com/drud/ddev/blob/1cb188308d85f359285c917b0d24256cac55abbc/pkg/globalconfig/global_config.go#L210-L295

Awesome, thanks!

rfay:
I'd like this to make it into the next point release (within the next week). Hope you can address the minor changes required.

bbrala:
Hmm, I was also thinking though. I think right now global options take precedent. But perhaps this should not be the case. You might want a global option, but if you set something on the project level you probably want that value.

For this particular option it might now matter much. But especially if we look forwards, what about the php version. The global should perhaps just function as a configurable default value.

rfay:
IMO global config should be a default, but there are cases where it isn't implemented that way, like mutagen. But you're certainly right that that's the obvious - global config should be default, and project config should be able to override it.

bbrala:
I've fixed all the requested changed afaik :) Ready for review again.

rfay:
I rebased using the github UI, so you'll need to update your branch if further work is required. Github broke everybody by removing brew from the path lately, and your code didn't have that fix.

I see the docs and link checks failing though, you'll want to look at those.

rfay:
I'll see if I can fix this by rolling back to https://github.com/drud/ddev/pull/4218/commits/a14223d72b3f3674ac64cde9711caf116a49033c and doing a rebase. Seems lots of things were lost.

rfay:
I haven't sorted out what happened, but we'll get it.

rfay:
Seems github was just slow to pick up my push, now it looks like all is OK. Except docs change landed that has conflicts. I'll fix that.

bbrala:
Thank you. :)

docs/content/users/details/offline-usage.md
```diff
@@ -13,7 +13,7 @@ You have general options as well:
 
 In `.ddev/config.yaml`, `use_dns_when_possible: false` will make DDEV never try to use DNS for resolution, instead adding hostnames to `/etc/hosts`. You can also use `ddev config --use-dns-when-possible=false` to set this configuration option.
 
-In `.ddev/config.yaml`, `project_tld: example.com` (or any other domain) can set DDEV to use a project that could never be looked up in DNS. You can also use `ddev config --project-tld=example.com`.
+In `.ddev/config.yaml` `project_tld: example.com` (or any other domain) can set DDEV to use a project TLD that could never be looked up in DNS. You can also use `ddev config --project-tld=example.com`. This also works as a global option in `~/.ddev/global_config.yaml` or use `ddev config global --project-tld=example.com`.
```
mattstein, line 16: I don’t think you have to explain that `example.com` is a placeholder, and it might be good to give a _little_ context to the difference between `.ddev/config.yaml` and `project_tld: example.com` instead of just a space:

```suggestion
In `.ddev/config.yaml`, you can use `project_tld: example.com` to have DDEV use a project TLD that won’t be looked up via DNS. You can do the equivalent with `ddev config --project-tld=example.com`. This also works as a global option in `~/.ddev/global_config.yaml` or running `ddev config global --project-tld=example.com`.
```

mattstein:
Added two minor suggestions, otherwise looks great to me. Thanks @bbrala!

rfay:
Rebased again, another merge conflict. Applied @mattstein 's changes. I'll re-test tomorrow and we can get it in. Would appreciate if you could do manual test and review again @bbrala, thanks.

bbrala:
Tested it locally built from this branch.

* ddev config drupal core.
* ddev config global --project-tld lol.com
* checked if it starts on right domain
* ddev config --project-tld pap.com
* checked if it starts on pap.com

It works :)

rfay:
Thanks so much @bbrala ! Happy to have you do any of those other needed globals if you get inspired. Won't require so much rebasing next time, that was just about the timing of the big docs PR.

rpkoller:
wow nice addition! thanks a lot! i've already tested a few minutes ago by updating to the latest version in head. everything is working as expected. :)  i had only one suggestion i brought up on discord and quickly chatted with @rfay about. currently the setting in the global config  is `project_tld: ""` while in the config for project files you have`# project_tld: ddev.site`

i wonder if it would make sense to change `project_tld: ""` in `~/.ddev/global_config.yaml` to `project_tld: ddev.site` that way it would be self explanatory which TLD ddev is using out of the box. the formatting would be also consistent with the `project_tld` used in project config files where no quotation marks are used (that inconsistency, for me as a none developer, was something that confused me. i didn't know that it doesn't matter if you use quotation marks as long as you dont have any spaces ). and by providing `project_tld: ddev.site` in the global config file it would also prevent the potential uncertainty in case i wanted to change the TLD if i need to prepend a period or not. that way everything would be self explanatory. 

the only caveat with my suggestion @rfay pointed out on discord is that you have to be careful in case you are changing the global project_tld` . ddev sets the default (`project_tld`: `DdevDefaultTLD                  = "ddev.site"`) in https://github.com/drud/ddev/blob/af1918778c56f789568951437f0a7b6ed4b15752/pkg/nodeps/values.go#L109 and then compares the constant `DdevDefaultTLD` with `ProjectTLD` in https://github.com/drud/ddev/blob/5049bdfb0a2dadf1185c63972d2e238cd2120969/pkg/ddevapp/config.go#L103-L106

PR #4228, opened 2022-09-27
Align `composer_version` docs with Composer version channels, fixes #4227, for #3812

### LINKED ISSUE #4227: Description about possible composer version config is insufficient

_No response_

### Expected Behavior

The `config.yaml` contains the following description for the `composer_version` config:

```
# composer_version: "2"
# You can set it to "" or "2" (default) for Composer v2 or "1" for Composer v1
# to use the latest major version available at the time your container is built.
# It is also possible to select a minor version for example "2.2" which will
# install the latest release of that branch. Alternatively, an explicit Composer
# version may be specified, for example "1.0.22". Finally, it is also possible
# to use one of the key words "stable", "preview" or "snapshot" see Composer
# documentation.
# To reinstall Composer after the image was built, run "ddev debug refresh".
```

This assumes that one can configure any Composer minor version, such as

```yaml
composer_version: "2.3"
# or
composer_version: "2.2"
# or
composer_version: "2.1"
# or
composer_version: "2.0"
```

and the latest patch release within the specified version range will be used.

### Actual Behavior

However, the only working minor version is `2.2`. This is because the 2.2 branch is added as additional version channel in Composer itself, see https://github.com/composer/composer/blob/2.4.2/src/Composer/SelfUpdate/Versions.php#L31. Specifying another minor version will result in DDEV installing the latest available Composer version, currently 2.4.2.

My suggestion is to

- either add support for specifying Composer minor versions
- or correct the information provided in `config.yaml` about possible config values for the `composer_version` config, supporting only the officially supported Composer channels (as per https://github.com/composer/composer/blob/2.4.2/src/Composer/SelfUpdate/Versions.php#L31)

Side note: This behavior was originally introduced with #3812.

### Steps To Reproduce

1. Configure a DDEV project with `composer_version: "2.3"`
2. Run `ddev debug refresh`
3. Run `ddev composer -- --version`
4. Compare the printed version with the configured version, the minor version branches should not match.

### Anything else?

_No response_

eliashaeussler:
## The Problem/Issue/Bug:

The `composer_version` config can only handle available Composer version channels as well as explicit Composer versions. However, the current documentation section did not explicitly reflect this behavior.

The appropriate documentation chapter is now adapted to list available Composer version channels.

## Related Issue Link(s):

Resolves: #4227
Related: #3812

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4228"></a>


pkg/ddevapp/templates.go
```diff
@@ -54,11 +54,12 @@ const ConfigInstructions = `
...
-# install the latest release of that branch. Alternatively, an explicit Composer
-# version may be specified, for example "1.0.22". Finally, it is also possible
-# to use one of the key words "stable", "preview" or "snapshot" see Composer
-# documentation.
+# It is also possible to use each other Composer version channel. This includes:
+#   - 2.2 (latest Composer LTS version)
+#   - stable
+#   - preview
+#   - snapshot
+# Alternatively, an explicit Composer version may be specified, for example "1.0.22".
```
rfay, line 62:
```suggestion
# Alternatively, an explicit Composer version may be specified, for example "2.2.18".
```

rfay (approved):
This is great, thanks. But it needs to be done in the docs too right? Maybe you're doing that in another PR. I just suggested updating the explicit composer example to something newer. You can just click the commit button to accept it.

eliashaeussler:
You're right, it should be added to the docs as well. I'd push it in this PR as well, if that's okay for you?

rfay:
Yes, docs in this PR is the best way, thanks!

rfay (approved):
Fantastic, thanks so much!

PR #4248, opened 2022-10-02
Implement traefik as ddev-router, fixes #3458, fixes #2653

### LINKED ISSUE #2653: Stop mounting docker.sock into ddev-router (security)
**Describe the bug**

ddev-router uses listening to docker.sock (via https://github.com/jwilder/docker-gen) to find out what containers have come up, etc.

However,

1. This is a security issue. Mounting docker.sock into ddev-router means that any attacker who was able to get access inside ddev-router could do basically anything to the host computer, because they can use the docker API on docker.sock, and the docker daemon runs as root.
2. Listening to find out what containers have exposed ports (which is what it does) has a flaw in that there might be non-ddev containers running with exposed ports, and ddev-router may then try to create a configuration for them that cannot work.

The second item might be the case if `docker logs ddev-router` shows something like "Processing certs and nginx for hostnames: .app.boilerplate.docker content-block-registration-api.ddev.site pma.boilerplate.docker" - where the hostnames aren't normal ddev hostnames (one here is, the others aren't)

I don't think there's any reason ddev can't deduce what containers need to be routed and to provide this information to ddev-router.

### LINKED ISSUE #3458: Replace ddev-router with Traefik Proxy
As ddev users we love the simplicity which is provided by the tool to handle topics like generation SSL certificates, route a URL to the right project. It's currently based on NGINX which is fine.
With the increasing amount of people who are using ddev and the different kind and mount of projects we should think about a refined router solution with more features like providing middlware logic or a dashboard for a quick overview and a better troubleshooting.
Troubleshooting is currently hard, because a developer needs to investigate the whole router config overall. 
One solution could be a traefik server. Traefik is a proxy solution which is dedicated deisgned to work with containers.
It comes with a extensible middleware logic.The proxy can handle different kind of protocols.
Letsencrypt is one of the supported SSL certificate soltions which can 100% covered by Traefik.
It's easier to troubleshoot issues in the router configuration. The build-in dashboard shows warnings for wrong configured servers without breaking the whole system.

There was already the idea to use Traefik in #956

Traefik can be configured with static config files (toml, yaml, json).
It can also work in a dynamic mode by using labels of running docker containers by using the mounted docker socket. That's the same way as in currenty ddev. Static and dynamic configuration can work together.
One adavantage of Traefik is that we do not need to restart the container if the dynamic config is changed.
Traefik can work in docker-compose environments.
So it should be possible to use the labels to generate Traefik compatible rules.

We have to investigate what's the easiest way to migrate to Traefik without bigger or maybe without any breaks in current project configurations.

There is already the project Warden (a similar project to ddev) which uses Traefik as router . It's popular in the Magento world.
Maybe we can learn from this project.

Resources:

- Traefik Middlewares: https://doc.traefik.io/traefik/middlewares/overview/
- Traefik Dashboard: https://doc.traefik.io/traefik/operations/dashboard/
- Warden: https://github.com/davidalger/warden/
- Current ddev router logic: https://github.com/drud/ddev/tree/master/containers/ddev-router

rfay:
## The Problem/Issue/Bug:

* #3458 
* #2653 

Lots of work
No longer binding docker.sock

* `ddev poweroff && ddev config global --use-traefik`


cmuench:
@rfay wow. You already did the implementation. I will try that out ASAP

cmuench:
@rfay Just started with some tests. First look is awesome. For the test I selected a project which communicates to another ddev project. Seems to work.
One thing I found is that already removed projects are still in the Traefik Dashboard.

cmuench:
@rfay Just started a Magento 2 project with Kibana and Elasticsearch. The Kibana is then listed in the Traefik Dashboard and can be accessed. The Elasticsearch Server is not listed.
`ddev status` shows the port of both services.

rfay:
> Skipping bad HTTP_EXPOSE port pair spec 9200 for service elasticsearch

Ah, the code assumes that the port pair will be a port pair, but a single port has always been acceptable (though rarely used). Thanks for pointing this out. 

`ddev stop` should probably remove config from the router, I'll check that. I think it removes it from the project files (.ddev/traefik) but don't think it actually makes that change in the docker volume that traefik is looking at.

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
 * Rancher Desktop does not provide bind mounts, so use `ddev config global --no-bind-mounts` which also turns on Mutagen.
 * Use a non-`ddev.site` name, `ddev config --additional-fqdns=rancher` for example, because the resolution of `*.ddev.site` seems to make it not work.
 * Rancher Desktop does not seem to currently work with `mkcert` and `https`, so turn those off with `mkcert -uninstall && rm -r "$(mkcert -CAROOT)"`. This does no harm and can be undone with just `mkcert -install` again.
+
+## Traefik Router
+
+The job of the `ddev-router` is to receive most HTTP or HTTPS traffic, like requests to `*.ddev.site`, and deliver it to the correct project's web container, see [Container Architecture](../basics/architecture.md#container-architecture). The router as of DDEV v1.21.3 was always a forked, poorly documented nginx reverse proxy container.
```
mattstein, line 32: Edit for brevity:

```suggestion
DDEV’s router plays an important role in its [container architecture](../basics/architecture.md#container-architecture), receiving most HTTP and HTTPS traffic for requests like `*.ddev.site` and delivering it to the relevant project’s web container.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+The very popular [Traefik Proxy](https://traefik.io/traefik/) will replace the current `ddev-router` in the future, but is available as an experimental configuration today. Feedback is welcome. To enable it:
```
mattstein, line 34: Combine explanation of history and availability of replacement, a bit more concisely:

```suggestion
Since v1.21.3, `ddev-router` has been based on a forked, poorly-documented nginx reverse proxy. A future release will replace this with the popular [Traefik Proxy](https://traefik.io/traefik/), available now as an experimental feature. Run the following to enable it:
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+```
+ddev poweroff && ddev config global --use-traefik
+```
+
+Most DDEV projects will work fine out of the box, but there is a vast array of possible configuration changes and options, and new ways to view what is going on with the router. You don't need to know any of these to use the Traefik router, but they will offer new options in the future.
```
mattstein, line 40: Edit for brevity, move “feedback welcome” sentiment alongside future feature note:

```suggestion
Most DDEV projects will work fine out of the box, with the benefit of vastly more configuration options and ways to work with the router. (This will likely lead to more features in the future, and we’d love your feedback if you’re trying this out now!)
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+### Traefik Configuration
+
+All configuration for the new router is intended to be customizable where needed. As with other files throughout the DDEV ecosystem, if the file has `#ddev-generated` in it, it can and will be overwritten by DDEV. If you want to "take over" the configuration, you remove the `#ddev-generated` and become responsible for the file's content.
```
mattstein, line 44: Shorten and strengthen start, link immediately to configuration docs:

```suggestion
You can fully customize the router’s [Traefik configuration](https://doc.traefik.io/traefik/).
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+All Traefik configuration is described at [docs.traefik.io](https://doc.traefik.io/traefik/).
```
mattstein, line 46: Mention the next-most-important and Traefik-specific provider type next:

```suggestion
All Traefik configuration uses the *file* provider, not the *docker* provider. Even though the Traefik daemon itself is running inside the `ddev-router` container, it uses mounted files for configuration, rather than listening to the Docker socket.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+All Traefik configuration uses the *file* provider, not the *docker* provider. Even though the Traefik daemon itself is running inside the `ddev-router` container, it uses mounted files for configuration, rather than listening to the Docker socket.
```
mattstein, line 48: End on config takeover as an aside, since that’s _not_ Traefik-specific and it may already be a familiar pattern:

```suggestion
!!!tip
    Like other DDEV configuration, any file with `#ddev-generated` will be overwritten unless you choose to “take over” it yourself. You can do this by removing the `#ddev-generated` line. DDEV will stop making changes to that file and you’ll be responsible for updating it.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+#### Global Traefik Configuration
+
+Global configuration for Traefik is automatically generated in ~/.ddev/traefik.
```
mattstein, line 52: Use backticks and “directory”:

```suggestion
Global configuration is automatically generated in the `~/.ddev/traefik` directory:
```

(Also removes redundant “Traefik” here since it’s in the heading and the directory name and hardly ambiguous at this point.)

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+* `static_config.yaml` which is the base configuration.
```
mattstein, line 54: Keep format consistent:

```suggestion
* `static_config.yaml` is the base configuration.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+* `certs/default_cert.*` is the default DDEV-generated certificates.
```
mattstein, line 55: Fix singular+plural conundrum (_is the (...) certificates_):

```suggestion
* `certs/default_cert.*` files are the default DDEV-generated certificates.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+* `config/default_config.yaml` contains global dynamic configuration, including pointers to the default certificates.
+
+#### Project Traefik Configuration
+
+Project configuration is automatically generated in the project's .ddev/traefik directory.
```
mattstein, line 60: Backticks and curly apostrophe:

```suggestion
Project configuration is automatically generated in the project’s `.ddev/traefik` directory.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+
+* The `certs` directory contains the `<projectname>.crt` and `<projectname>.key` certificate generated for this project.
```
mattstein, line 62: Swap “this” for “the” to be consistent. And because we’re talking about a per-project file and not a specific project:

```suggestion
* The `certs` directory contains the `<projectname>.crt` and `<projectname>.key` certificate generated for the project.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+* The `config/<projectname>.yaml` file contains the configuration for the project, including information about routers, services, and certificates.
+
+### Debugging Traefik Routing
+
+* Traefik provides a dynamic description of its configuration that you can visit at `http://localhost:9999`.
```
mattstein, line 67: Un-bullet since there’s no introduction or strong reason for this to be a list item:

```suggestion
Traefik provides a dynamic description of its configuration you can visit at `http://localhost:9999`.
```

docs/content/users/configuration/experimental.md
```diff
@@ -26,3 +26,43 @@ Rancher Desktop integration currently has no automated testing for DDEV integrat
...
+* When things seem to be going wrong, do a `ddev poweroff` and then start your project and see what the Traefik daemon is doing or failing at with `docker logs ddev-router` or `docker logs -f ddev-router`.
```
mattstein, line 68: Un-bullet for the same reason, “run” command rather than “do”, link to command reference, and attempt to subtly clarify the relationship between the container logs and daemon we’re looking to investigate:

```suggestion
When things seem to be going wrong, run [`ddev poweroff`](../basics/commands.md#poweroff) and then start your project again by running [`ddev start`](../basics/commands.md#start). Examine the router’s logs to see what the Traefik daemon is doing (or failing at) by running `docker logs ddev-router` or `docker logs -f ddev-router`.
```

docs/content/users/extend/custom-tls-certificates.md
```diff
@@ -3,5 +3,9 @@
 It’s possible to use “real” TLS certificates issued by a CA rather than the local-development-oriented `mkcert` command.
 
 1. Obtain a certificate and key from Let’s Encrypt or another source.
-2. Install the certificate and key in your project’s `.ddev/custom_certs` directory. Each certificate must be named with the pattern `fqdn.crt` and `fqdn.key`. If you’re working with a project `example.ddev.site`, you would have `example.ddev.site.crt` and `example.ddev.site.key` in `.ddev/custom_certs`. There must be one cert-set for each FQDN handled by the project.
+2. Install the certificate and key in your project’s `.ddev/custom_certs` directory.
+   * If using traditional router (current default), each certificate must be named with the pattern `fqdn.crt` and `fqdn.key`. If you’re working with a project `example.ddev.site`, you would have `example.ddev.site.crt` and `example.ddev.site.key` in `.ddev/custom_certs`. There must be one cert-set for each FQDN handled by the project.
```
mattstein, line 7: Reword slightly to warm up language, clarify example, and more concisely reference default:

```suggestion
   * If you’re using the default router, each certificate must be named with the pattern `fqdn.crt` and `fqdn.key`. A project named `example.ddev.site`, for example, would need `example.ddev.site.crt` and `example.ddev.site.key` in `.ddev/custom_certs`. There must be one cert-set for each FQDN handled by the project.
```

docs/content/users/extend/custom-tls-certificates.md
```diff
@@ -3,5 +3,9 @@
...
+   * If using Traefik (`ddev config global --use-traefik`) then the files should be named `<projectname>.crt` and `<projectname>.key`, for example `exampleproj.crt` and `exampleproj.key`.
```
mattstein, line 8: Replace redundant instruction with link:

```suggestion
   * If you’re [using Traefik](../configuration/experimental.md#traefik-router), the files should be named `<projectname>.crt` and `<projectname>.key`, for example `exampleproj.crt` and `exampleproj.key`.
```

Because there’s more to know about Traefik in general, and in this context the focus is on custom certificates.

docs/content/users/extend/custom-tls-certificates.md
```diff
@@ -3,5 +3,9 @@
...
 3. Run [`ddev start`](../basics/commands.md#start) and verify using a browser that you’re using the right certificate.
+
+When using Traefik you have other options, and can completely override configuration and certificates using the `.ddev/traefik` directory.
```
mattstein, line 11: This is covered on the page we’re now linking to, and it’s odd to _sort of_ explain it here.

```suggestion
```

mattstein (changes requested):
Left a dazzling pile of suggestions I doubt will surprise you! ✨😄

docs/content/users/configuration/experimental.md
rfay, line 34:
```suggestion
Until DDEV v1.21.3, `ddev-router` was been based on a forked, poorly-documented nginx reverse proxy. A future release will replace this with the popular [Traefik Proxy](https://traefik.io/traefik/), available now as an experimental feature. Run the following to enable it:
```

docs/content/users/configuration/experimental.md
rfay: All of which I fundamentally object to, as you know :)  We'll have to find some truce on that. Probably figuring out a way to do *presentation* with curlies without affecting the *text*.

docs/content/users/configuration/experimental.md
mattstein: Whoops! I misread that as `ddev-router` starting with its current forked nginx proxy in v1.21.3.

Your edit fixes my mistake, but also seems to suggest that in DDEV v1.21.3 `ddev-router` is no longer based on that nginx fork—but it still is by default if I’m understanding correctly. I edited my suggestion above to clarify that 1.21.3 adds an opt-in Traefik option that will become the default in some future release:

> The default `ddev-router` is based on a forked, poorly-documented nginx reverse proxy. A future release will replace this with the popular [Traefik Proxy](https://traefik.io/traefik/), available in DDEV v1.21.3 as an experimental feature. Run the following to enable it:

docs/content/users/configuration/experimental.md
mattstein: Wait—I know you don’t like curly quotes for text, but you don’t object to the _backticks_ do you?

docs/content/users/configuration/experimental.md
mattstein: ...and one more revision to try and clarify that 1.21.3 adds a new option for the first time. It’s weird to start the paragraph calling it a “default” when it’s been the _only_ option before this new introduction:

> `ddev-router` has been based on a forked, poorly-documented nginx reverse proxy. DDEV v1.21.3 adds a new router based on the popular [Traefik Proxy](https://traefik.io/traefik/), available as an experimental feature until it becomes the default in a future release. Run the following to enable it:

Updated my suggestion above and I promise I’ll leave it alone now.

docs/content/users/configuration/experimental.md
rfay: No, of course can't ever object to backticks. Skimming thought it had to do with curly quotes.

rfay (commented):
Thanks so much!

PR #4264, opened 2022-10-05
[docs] Add spell check instructions

mattstein:
## The Problem/Issue/Bug:

The enthusiastic writer may not be aware of the spell checker, which would be ideal for the [Working on the Docs](https://ddev.readthedocs.io/en/latest/developers/testing-docs/) page.

This adds a “Check for Spelling Errors” section to that page:

Screen Shot 2022-10-05 at 03 48 08 PM@2x

Terminal commands are grabbed directly [from the Makefile](https://github.com/drud/ddev/blob/master/Makefile), and the `→  ` is used to indicate user CLI input since that style is used elsewhere in the docs.

https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-your-changes

Only Markdown; no new tests needed.

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4264"></a>

### DIFF
--- a/docs/content/developers/testing-docs.md
+++ b/docs/content/developers/testing-docs.md
@@ -1,8 +1,9 @@
 # Working on the Docs
 
-This section is designed for people who want to contribute to the DDEV documentation.
+This section is for people who want to contribute to the DDEV documentation.
 
-On any page in the docs
+!!!tip "No need to work locally!"
+    The documentation is built and checked automatically with various [GitHub Actions workflows](https://github.com/drud/ddev/actions). It may help to check your work locally, particularly for more involved PRs, but you can just as well make suggestions using GitHub in a browser.
 
 ## Fork / Clone the DDEV Repository
 
--- a/docs/content/developers/testing-docs.md
+++ b/docs/content/developers/testing-docs.md
@@ -3,7 +3,16 @@
 This section is for people who want to contribute to the DDEV documentation.
 
 !!!tip "No need to work locally!"
-    The documentation is built and checked automatically with various [GitHub Actions workflows](https://github.com/drud/ddev/actions). It may help to check your work locally, particularly for more involved PRs, but you can just as well make suggestions using GitHub in a browser.
+    The documentation is built and checked automatically with various [GitHub Actions workflows](https://github.com/drud/ddev/actions). It may help to check your work locally, particularly for more involved PRs, but you can just as well make suggestions using [GitHub in a browser](#fix-docs-using-web-browser).
+
+## Fix Docs Using Web Browser
+
+On any page in the docs:
+
+1. Click the pencil in the upper right. That will take you to the the right page on GitHub.
+2. Click the pencil button on GitHub and follow the instructions to create your change.
+3. Save your changes and follow the prompts to create a PR.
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks good to me. I don't think the installation works for all environments, but it's a good start.

Want to get them to install markdownlint too, for the same reason?

mattstein:
> Looks good to me. I don't think the installation works for all environments, but it's a good start.

I forget what I did to get it working on my Mac, but I figured it out once I found that Makefile comment. (Trying on another machine now.) Some clues are better than nothing!

> Want to get them to install markdownlint too, for the same reason?

Aren’t we already doing that in the section above? If you mean something else, I’m not following.

rfay:
You're right, sorry. Markdownlint instructions are already there.

rfay:
Is it worth saying that it's OK to just toss it to the build check if they don't want to mess with it?

mattstein:
I think so! Worth pointing out it’ll be done CI so nobody’s discouraged from submitting a PR. I shall whip up a blurb—one moment.

mattstein:
Screen Shot 2022-10-05 at 04 05 52 PM@2x

rfay (changes requested):
Oh dear. We seem to have lost the whole section I did recently showing this. https://github.com/drud/ddev/pull/4237 - please put that back in, edit as you see fit.

mattstein:
My fault, sorry! Just re-added.

docs/content/developers/testing-docs.md
```diff
@@ -3,7 +3,16 @@
 This section is for people who want to contribute to the DDEV documentation.
 
 !!!tip "No need to work locally!"
-    The documentation is built and checked automatically with various [GitHub Actions workflows](https://github.com/drud/ddev/actions). It may help to check your work locally, particularly for more involved PRs, but you can just as well make suggestions using GitHub in a browser.
+    The documentation is built and checked automatically with various [GitHub Actions workflows](https://github.com/drud/ddev/actions). It may help to check your work locally, particularly for more involved PRs, but you can just as well make suggestions using [GitHub in a browser](#fix-docs-using-web-browser).
```
rfay, line 6: Do you want this tip to be in the section below about Fork/Clone?

rfay:
I'm not sure that was your fault that got removed, must have been a merge things, but things like that are always scary. That was definitely something that went in while you were working on this.

docs/content/developers/testing-docs.md
mattstein: Good point about the flow. I think it’d be better to remove the tip treatment since it’s less urgent, and use that blurb as context. Commit coming in a moment.

mattstein:
> I'm not sure that was your fault that got removed, must have been a merge things, but things like that are always scary. That was definitely something that went in while you were working on this.

I just assume I botched the merge commit even though I recall _seeing_ that new section.

docs/content/developers/testing-docs.md
mattstein: How’s [this](https://github.com/drud/ddev/pull/4264/commits/3312c5fe8f1716088cc76fc2c8726a01a4ebb960) feel instead?

rfay (approved):
Thanks!

PR #4265, opened 2022-10-06
Use proper post-start hook label everywhere

hanoii:
Post-start hook is wrongly stated on two places, fixed that.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4265"></a>


rfay (approved):
Thanks!

PR #4268, opened 2022-10-07
Add support for TYPO3 v12

gilbertsoft:
## The Problem/Issue/Bug:

The new TYPO3 v12 major version has some breaking changes regarding the settings files.

This patch introduces an automatic detection of the TYPO3 version 12 or higher and properly adapts the setting file paths and names.

To avoid the creation of the .gitignore in the wrong location, it is not written anymore before TYPO3 is installed or the version could be properly detected.

Follow the quick start guide to test this change. To test it for a previous version just add a version to the composer create command e.g.:

```
ddev composer create --no-install "typo3/cms-base-distribution:^11"
```

## Related Issue Link(s):

https://github.com/TYPO3/get.typo3.org/issues/386

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4268"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -226,7 +226,7 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ddev start
         ddev composer create "typo3/cms-base-distribution" --no-install
         ddev composer install
-        ddev config --auto
+        ddev restart
         ddev exec touch public/FIRST_INSTALL
         ddev launch
         ```
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -240,7 +240,7 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         cd example-site
         ddev config --project-type=typo3 --docroot=public --create-docroot --php-version 8.1
         ddev composer install
-        ddev config --auto
+        ddev restart
         ddev exec touch public/FIRST_INSTALL
         ddev launch
         ```
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -222,10 +222,11 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ```bash
         mkdir my-typo3-site
         cd my-typo3-site
-        ddev config --project-type=typo3 --docroot=public --create-docroot
+        ddev config --project-type=typo3 --docroot=public --create-docroot --php-version 8.1
         ddev start
         ddev composer create "typo3/cms-base-distribution" --no-install
         ddev composer install
+        ddev config --auto
         ddev exec touch public/FIRST_INSTALL
         ddev launch
         ```
@@ -237,8 +238,10 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ```bash
         git clone https://github.com/example/example-site
         cd example-site
-        ddev config
+        ddev config --project-type=typo3 --docroot=public --create-docroot --php-version 8.1
         ddev composer install
+        ddev config --auto
+        ddev exec touch public/FIRST_INSTALL
         ddev launch
         ```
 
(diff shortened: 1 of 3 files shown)


docs/content/users/quickstart.md
```diff
@@ -222,10 +222,11 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ```bash
         mkdir my-typo3-site
         cd my-typo3-site
-        ddev config --project-type=typo3 --docroot=public --create-docroot
+        ddev config --project-type=typo3 --docroot=public --create-docroot --php-version 8.1
         ddev start
         ddev composer create "typo3/cms-base-distribution" --no-install
         ddev composer install
+        ddev config --auto
```
gilbertsoft, line 229: Instead of `ddev config --auto` we also could recommend `ddev restart` to create the settings file. AFAIK this is a long standing issue anyway since the introduction of the `--no-install` option.

pkg/ddevapp/typo3.go
```diff
@@ -62,7 +69,7 @@ func writeTypo3SettingsFile(app *DdevApp) error {
 		}
 
 		// The directory doesn't exist, create it with the appropriate permissions.
-		if err := os.Mkdir(dir, perms); err != nil {
+		if err := os.MkdirAll(dir, perms); err != nil {
```
gilbertsoft, line 72: This also fixes a long standing issue with older versions too.

docs/content/users/quickstart.md
rfay: Why would we do a `ddev config --auto` when we just did the explicit ddev config above? It doesn't seem like this belongs at all.

gilbertsoft: Other solution would be to create the files even with a simple `ddev composer` command.

gilbertsoft: I've checked the sources again, at the moment the settings files are create with start, config and importdb. Because I now need TYPO3 properly to be installed for the detection of the version, the behavior has changed a bit and doesn't currently work without an additional line her.

rfay: Settings file is created on `ddev start` if the project type is set.

rfay:
This one I'd dearly love to get into v1.21.2, which I hope to cut very soon, would have considered it today if this didn't come up. Thanks for the great work on this!

docs/content/users/quickstart.md
gilbertsoft: Hope this should be clear now after your discussion yesterday and finally it's solved with https://github.com/drud/ddev/pull/4273.

gilbertsoft:
Fixed the failing test today and all green again :)

rfay:
Congrats on tests, and thanks!

docs/content/users/quickstart.md
rfay: No, I never did understand, but glad about #4273.

rfay: Oh, I see now. A `ddev restart` after the `ddev composer install` would have worked and is probably better, especially wrt mutagen. But we can go with this as it is now.

rfay, line 229:
```suggestion
        ddev restart
```

When you commit these, just add [skip ci] in the commit message. You can commit the two together as a batch.

docs/content/users/quickstart.md
```diff
@@ -237,8 +238,10 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ```bash
         cd example-site
-        ddev config
+        ddev config --project-type=typo3 --docroot=public --create-docroot --php-version 8.1
         ddev composer install
+        ddev config --auto
```
rfay, line 243:
```suggestion
        ddev restart
```

rfay (approved):
Thanks so much! Manually tested on v12 and v11, worked fine. 

The two minor docs changes can be committed as a batch with `[skip ci]` in the commit title and it won't run a test.

PR #4273, opened 2022-10-07
Optimize composer create to also work without --no-install

gilbertsoft:
## The Problem/Issue/Bug:

`composer create-project` executes a `git clone` which is possible only into empty directories. Therefor the `ddev composer create` command was introduced to automatically bypass this problem and make the initialization easier for the user. This is achieved by running the `composer create-project` command automatically providing the directory argument and rsync the the files and folders from this directory afterwards to the project root directory. This can be very slow depending on the number of files and folders which must be synchronized and therefor using the `--no-install` option is suggested in many places to avoid issues and make the process faster but requires the user to run `ddev composer install` afterwards, once the `ddev composer create` command has finished. This workflow can be optimized again by handling the `--no-install` option automatically.

The `composer create` command now checks if the `--no-install` option is provided by the user. If not, it's now added by default and once the rsync is completed, `composer install` is called afterwards. This makes the project initialization easier.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4273"></a>

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -98,6 +98,14 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 		tmpDir := util.RandString(6)
 		containerInstallPath := path.Join("/tmp", tmpDir)
 
+		// Remember if --no-install was provided by the user
+		noInstallPresent := nodeps.ArrayContainsString(osargs, "--no-install")
+		if !noInstallPresent {
+			// Add the --no-install option by default to avoid issues with
+			// rsyncing many files afterwards to the project root.
+			osargs = append(osargs, "--no-install")
+		}
+
 		// Build container composer command
 		composerCmd := []string{
 			"composer",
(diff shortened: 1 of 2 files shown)


rfay (commented):
I like it!

gilbertsoft:
> I like it!

Have to handle the options properly, some must be used for the install too.

gilbertsoft:
Done. Hope, all tests pass properly :)

rfay:
This is going to be great. It's too late in the v1.21.2 cycle though, so it can get in early after release and people will be using it then before v1.21.3 or v1.22

gilbertsoft:
There are potentially more cases in the quick start to change. So far I've only changed what works for sure. Other cases must be tested properly....

rfay (changes requested):
This is awesome, a huge step forward. 

* Would it be possible for us to now use the command `ddev composer create-project` since it will actually work pretty well?
* I think we should fix up the "on the way" statements so it outputs "Doing composer create-project --no-install <args>" and then "Moving composer.json to project root" and then "Running `composer install <args>`". Probably already does that last one well enough.

Wonderful and awesome! Best thing that's happened to this since the beginning. Thanks!

rfay:
Rebased and fixed minor conflict.

rfay:
Rebased again.

rfay (approved):
Brilliant approach, well done. Big benefit. Thanks!

rfay:
This causes the laravel quickstart to fail, at least colima/macOS/mutagen.  It seems the create does more work than we usually expect before everything is copied over.

```
Executing composer command: [composer create-project --prefer-dist laravel/laravel --no-install /tmp/LPEgOF]

Creating a "laravel/laravel" project at "/tmp/LPEgOF"
Installing laravel/laravel (v9.3.12)
  - Installing laravel/laravel (v9.3.12): Extracting archive
Created project in /tmp/LPEgOF
> @php -r "file_exists('.env') || copy('.env.example', '.env');"
> @php artisan key:generate --ansi
PHP Warning:  require(/tmp/LPEgOF/vendor/autoload.php): Failed to open stream: No such file or directory in /tmp/LPEgOF/artisan on line 18

Warning: require(/tmp/LPEgOF/vendor/autoload.php): Failed to open stream: No such file or directory in /tmp/LPEgOF/artisan on line 18
PHP Fatal error:  Uncaught Error: Failed opening required '/tmp/LPEgOF/vendor/autoload.php' (include_path='.:/usr/share/php') in /tmp/LPEgOF/artisan:18
Stack trace:
#0 {main}
  thrown in /tmp/LPEgOF/artisan on line 18

Fatal error: Uncaught Error: Failed opening required '/tmp/LPEgOF/vendor/autoload.php' (include_path='.:/usr/share/php') in /tmp/LPEgOF/artisan:18
Stack trace:
#0 {main}
  thrown in /tmp/LPEgOF/artisan on line 18
Script @php artisan key:generate --ansi handling the post-create-project-cmd event returned with error code 255
Failed to create project:exit status 255, stderr=
```

tyler36:
Not sure waht exactly you want to me check:

## Test

```shell
$ ddev --verion
ddev version v1.21.2-4-ga5403f9f

$ ddev config --project-type=laravel --docroot=public --create-docroot
...

$ ddev start
Successfully started ga5403f9f 

$ ddev composer create-project --prefer-dist --no-install --no-scripts laravel/laravel
'ddev composer create-project' is unsupported. Please use 'ddev composer create'
for basic project creation or 'ddev ssh' into the web container and execute
'composer create-project' directly.

$ ddev ssh
$ composer create-project --prefer-dist --no-install --no-scripts laravel/laravel
Created project in /var/www/html/laravel

# Note: Laravel is installed in a subdirectory instead of current directory.

$ ls -a
.  ..  .ddev  laravel  public

```

PR #4150, opened 2022-08-28
Improve WSL2 Installation experience with a PowerShell script [skip ci]

rfay:
## The Problem/Issue/Bug:

The existing installation instructions include a long, long list of activities, most of which can be scripted.

## How this PR Solves The Problem:

Try to script them. `wsl --install` and Docker Desktop installation are prerequisites.

## TODO
- [x] Rebase with current
- [x] Current docs suggest docker-ce as default, it really shouldn't be until it matures, at least with consistent xdebug
- [ ] Add a docker-ce script as well

## Manual Testing Instructions:

Try running it. It shouldn't harm existing installations.

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4150"></a>

PR #4274, opened 2022-10-08
[docs] fix command for installing Ubuntu 20.04 on WSL2

shaal:
## The Problem/Issue/Bug:
The current command listed in ddev docs results in error.

Tested on Windows 11 Pro - this PR uses the new command

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4274"></a>

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -64,7 +64,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     * If you already have Docker Desktop installed, make sure to disable its integration with your WSL2 distro. In *Resources* → *WSL Integration*, disable integration with the default distro and with your particular distro.
     * If you don’t already have WSL2, install it with `wsl --install`. This will likely require a reboot.
     * Run `wsl --set-default-version 2`.
-    * Install a distro. We recommend Ubuntu 20.04: `wsl --install Ubuntu-20.04`.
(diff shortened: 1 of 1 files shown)


docs/content/users/install/docker-installation.md
```diff
@@ -64,7 +64,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     * If you already have Docker Desktop installed, make sure to disable its integration with your WSL2 distro. In *Resources* → *WSL Integration*, disable integration with the default distro and with your particular distro.
     * If you don’t already have WSL2, install it with `wsl --install`. This will likely require a reboot.
     * Run `wsl --set-default-version 2`.
-    * Install a distro. We recommend Ubuntu 20.04: `wsl -s Ubuntu-20.04`.
+    * Install a distro. We recommend Ubuntu 20.04: `wsl --install Ubuntu-20.04`.
```
rfay, line 67: Actually, if you'd done `wsl --install` earlier, this wouldn't have been necessary at all. It installs Ubuntu (which is currently 20.04). So this line should be omitted. However, the MS flow is kind of unpredictable. If you have already installed WSL2 then `wsl --install` doesn't work this way, and you have to `wsl --install Ubuntu`

rfay:
```suggestion
    * If the `Ubuntu` distro is not already installed and the default distro (see `wsl -l -v`) then install it with `wsl --install Ubuntu`.
```

rfay (changes requested):
Thanks so much for exploring this territory (and testing #4150). 

Commit this suggestion if you think it works.

PR #4275, opened 2022-10-08
[docs] auto-install recommended packages

shaal:
## The Problem/Issue/Bug:
When trying to copy all the commands and paste them in terminal of Ubuntu so they all run at once, the `-y` is missing.

Adding `-y` to auto-confirm installing Ubuntu packages.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4275"></a>


rfay (approved):
This is fine. 

Since you're an expert novice now, would you please try this? It would help so much
* https://github.com/drud/ddev/pull/4150 

You can `wsl --unregister Ubuntu` assuming your distro is named Ubuntu.

Issue #4188, opened 2022-09-07
Audit documentation for consistency, accuracy, and flow


mattstein:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Despite being a helpful and detailed resource, the DDEV documentation could be more internally consistent, up to date, and at some times concise. Stronger documentation could reduce some support burden, help people get working or un-stuck faster, and make DDEV more of a pleasure to work with than it already is.

### Describe the solution you'd like

I’d like to suggest the following items and will tackle them myself as long as each seems reasonable:

1. In the spirit of the proposed [Code of Conduct](https://github.com/drud/ddev/pull/4186), draft a writing style guide that broadly encourages consistency of tone and convention.

    a. Describe an ideal tone for writing that’s ideal for DDEV and a natural fit for the code of conduct. (Welcoming, focused, warm, precise.)

    b. Describe the intended audience for the documentation that can be used to evaluate the appropriateness and accessibility of any writing.

    c. DDEV should appear in all caps in sentence form as long as it doesn’t reference the `ddev` command; retire “DDEV-Local” and “ddev”.

    d. Should avoid using “just” and “it’s easy” that subtly talk down to the reader and suggest anyone struggling is wrong, and instead use words like “straightforward” and offer reassurance by demonstration or illustration whenever possible.

2. Audit all documentation and implement suggestions from item 1 above.

3. Fix pedantic things across the board: use curly quotes, fix proper nouns (“Docker” not ”docker”, etc.), consistently capitalize headings, edit for brevity and clarity. Cross-link more liberally so readers can more easily land in the right place. Update `.spellcheckwordlist.txt` accordingly.

4. Rename the existing “Styleguide” page to “Brand Guide” and season accordingly.

5. Update installation instructions with a heading specifically for upgrading. (Easier to scan and link to, since upgrading is likely _more_ common than first-time installation.)

6. Split commands into organized subheadings for more efficient browsing. Emphasize—on that page or elsewhere—day-to-day usage and the most popular or expected commands.

7. Reduce redundant information that increases the risk of becoming stale, with a greater emphasis on links to canonical content.

8. Request related admin-level changes to things like the repository description. (Even disabling the unused [Packages section](https://share.cleanshot.com/3oyLup) and [Projects](https://github.com/drud/ddev/projects) feature—since every bit of cruft removed puts more emphasis on meaningful bits.)

9. Fix any 404s and bad links.

[...]

PR #4293, opened 2022-10-13
[docs] Improve FAQ

mattstein:
## The Problem/Issue/Bug:

The [current FAQ](https://ddev.readthedocs.io/en/stable/users/basics/faq/) is a bit sprawling, and it contains important information that can’t be linked to directly.

This shapes up the FAQ to be more readable and linkable in a few ways:

- Re-titles excessively-explanatory “Frequently-Asked Questions (FAQ)” simply to “FAQ”, while repurposing the phrase “frequently-asked questions” in introductory framing for the one person who searches that term or has never seen an FAQ.
- Gives a proper heading to each Q&A item, further organized categorically for easier browsing and more orderly maintenance.
- Pares down headings wherever possible for brevity and readability, with sense breaks added to text.
- Swaps in the word “workstation” in a few places, since we’re trying to use that more consistently.
- Keeps all existing Q&A items, paring down some in favor of linking (which reinforces documentation structure), and adding some explanatory or illustrative content where more context can be useful. (Most obviously in _How can I connect to my database?_, which I assume is a constant point of friction for any container-based development environment.)
- Rewords the question about contributing financially so it’s less redundant with the preceding question. (This content should get a landing page with more detail on ddev.com!)

This comes with the downside of chaotic sidebar navigation, which I’ve addressed separately in #4294.

https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes

n/a

## Related Issue Link(s):

Part of this overall effort:

- #4188

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4293"></a>

### DIFF
--- a/docs/content/users/basics/faq.md
+++ b/docs/content/users/basics/faq.md
@@ -205,12 +205,12 @@ See the [support options](../support.md), including [Discord](https://discord.gg
 
 We love and welcome contributions of knowledge, support, docs, and code:
 
-- Submit an issue or pull request to the [main repo](https://github.com/drud/ddev).
-- Add your external resource to [awesome-ddev](https://github.com/drud/awesome-ddev).
-- Add your recipe or HOWTO to [ddev-contrib](https://github.com/drud/ddev-contrib).
-- Help others in [Discord](https://discord.gg/kDvSFBSZfs) and on [Stack Overflow](https://stackoverflow.com/tags/ddev).
-- Contribute financially via [GitHub Sponsors](https://github.com/sponsors/rfay).
-- Get involved with DDEV governance and the [Advisory Group](https://github.com/drud/ddev/discussions/categories/ddev-advisory-group).
+* Submit an issue or pull request to the [main repo](https://github.com/drud/ddev).
+* Add your external resource to [awesome-ddev](https://github.com/drud/awesome-ddev).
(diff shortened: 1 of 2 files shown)


rfay:
Thanks, probably won't be able to review until next week, but much appreciated!

docs/content/users/basics/faq.md
```diff
@@ -1,144 +1,217 @@
-# Frequently-Asked Questions (FAQ)
+# FAQ
```
rfay, line 1: I suspect the old way catches more SEO and provides more than one way to find this in search.

rfay (approved):
This is a great improvement. Thanks!

docs/content/users/basics/faq.md
mattstein: I mentioned search in the first bullet point of the PR description: I kept the phrase “frequently-asked questions” on the page with that in mind. Giving each question a proper heading is *probably* better for overall SEO surfacing those questions—and I’m guessing `ddev faq` would be a more common search than `ddev frequently asked questions`, but that could be verified with research.

The real win would be providing [structured data to further expose Q&A items](https://developers.google.com/search/docs/appearance/structured-data/faqpage), but that’d require a far deeper journey into the theme.

rfay:
Will pull when you're ready.

mattstein:
Pull away!

PR #4303, opened 2022-10-17
Craft CMS: Return existing upload directory, if already set, fixes #4302

### LINKED ISSUE #4302: Project type `craftcms` doesn't listen to `upload_dir` config setting.

_No response_

### Expected Behavior

Project type `craftcms` should listen to `upload_dir` config setting.

### Actual Behavior

Project type `craftcms` defaults to `files` as it is hard coded in the project type..
`files` shouldn't be seen as a default because it is opinionated.

### Steps To Reproduce

1. Start a new project `ddev config --project-type=craftcms && ddev start`.
2. Check the default value `ddev exec 'echo $DDEV_FILES_DIR' `  (returns `files`).
3. Change `upload_dir` in `.ddev/config.yaml` to anything.
4. Check for the custom value `ddev exec 'echo $DDEV_FILES_DIR' `  (still returns `files`).

### Anything else?

_No response_

bencroker:
Fixes 
* https://github.com/drud/ddev/issues/4302 
by returning the existing upload directory, if already set.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4303"></a>

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -27,7 +27,9 @@ func craftCmsConfigOverrideAction(app *DdevApp) error {
 
 // Returns the upload directory for importing files, if not already set
 func getCraftCmsUploadDir(app *DdevApp) string {
-	app.UploadDir = "files"
+	if app.UploadDir == "" {
+		return "files"
+	}
 
 	return app.UploadDir
 }


bencroker:
@rfay are tests failing due to my changes or does something need to be enabled for this PR?

rfay:
Mostly not you, but TestDdevFullSiteSetup is you:
```
    ddevapp_test.go:1949: 
        	Error Trace:	ddevapp_test.go:1949
        	            	http status code was 404, not 200
        	Test:       	TestDdevFullSiteSetup
        	Messages:   	failed ImageURI response on project TestPkgCraftCms
    ddevapp_test.go:1951: 
        	Error Trace:	ddevapp_test.go:1951
        	Error:      	Not equal: 
        	            	expected: "image/jpeg"
        	            	actual  : "text/html; charset=UTF-8"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-image/jpeg
        	            	+text/html; charset=UTF-8
        	Test:       	TestDdevFullSiteSetup
```

rfay:
You can run that manually pretty easily to study it. If necessary, we can skip the jpeg check. Can't remember the discussion but it seems at once point we were trying to set upload_dir for this.

bencroker:
Any idea why this would work with ARM but not AMD? 

Screenshot 2022-10-17 at 17 16 12

rfay:
For resource reasons, the CircleCI amd64 test is the only one (of all of them) that runs all of the options in TestDdevFullSiteSetup, which take forever, see https://github.com/drud/ddev/blob/572ec67714b44571bc8dec32ec91c49ca55c6b16/.circleci/config.yml#L38-L44 where it's the only one that doesn't have GOTEST_SHORT set to something.

Unfortunately... that's mostly the only test you care about. As a reminder, you can run just your test using the technique in the last line of https://github.com/drud/ddev/blob/a94f9d7aad2f3ec8012fd3f6f118dac71033557c/docs/content/developers/project-types.md

```
GOTEST_SHORT=13 make testpkg TEST_ARGS="-run TestDdevFullSiteSetup"
```

Or easier, just set it up in Goland or vscode and run it there.

Thank you so much!

bencroker:
> Is there actually a default or even likely files/upload_dir?

There's no default and not really any one commonly used directory name.

> Is it likely to be '/files' ?

No. We chose this simply because we had to choose something to make the `FilesImageURI` test pass.

khalwat:
@bencroker I believe I tried what you have in your PR, and it didn't work (failed the tests), so I assumed `upload_dir` was set to some other default other than being empty but... unclear what that could be.

pkg/ddevapp/craftcms.go
```diff
@@ -27,7 +27,9 @@ func craftCmsConfigOverrideAction(app *DdevApp) error {
 
 // Returns the upload directory for importing files, if not already set
 func getCraftCmsUploadDir(app *DdevApp) string {
-	app.UploadDir = "files"
+	if app.UploadDir == "" {
+		return "files"
```
rfay, line 31: I don't think defaulting to "files" is a reasonable thing to do, given our conversation. It shouldn't have a default if there's no such expectation in Craft.

rfay (changes requested):
I'll take a look and see if I can figure something out that works for the community.

pkg/ddevapp/craftcms.go
bencroker: Great, how do we get that to work with the testing framework?

khalwat: I think there should be a default. The issue this PR is intending to fix is to allow the `upload_dir` to be overridden via config, which is a perfectly reasonable thing to want.

Providing a default in this context is fine; we just need to ensure that there is a way the user can _override_ that default.

The issue I had was the `if app.UploadDir == "" {` always failed, which means it is set to _something_ by default (or maybe it's null?). If we can figure out what that default value is, we can either:

* Properly check for it, and not set a default ourselves
* Use that default value in our tests

The latter of the two requires redoing the test project setup, so I'd have a preference for the first

ref: https://github.com/drud/ddev/issues/4302

rfay: Trying to look at it this morning. We may want to add the upload_dir to the testing framework.

rfay: @khalwat I've asked about this a number of times and the answer was always "there is no default files dir in Craft". So there are two things at play right? We obviously have to make upload_dir work right and override any project-type-provided upload_dir. And we also have to do some reasonable testing.

khalwat: Correct, there is no default for Craft, because people can organize their project any way they want.

Also what you might typically think of as an "upload_dir" might be somewhat of a dated or at least very CMS-specific concept, but generally relates to where user uploaded files (images, etc.) will go, correct?

In Craft, the default project has zero Asset Volumes set up. You set them up yourself, and you choose where they point to, whether local directories, or S3 buckets, or other cloud storage.

So we are providing a default of `files/` for two reasons:

1) We need _something_ for the test project to pass the "uploading file" test, which still seems useful to me
2) It's a sane default that at least some % of people using Craft will use.

Thus I think it's fine to provide said default _providing_ that we can properly allow them to override the default via the `.yaml` config.

The primary issue I was having, and @bencroker has run into, is that we can't do this test:

```go
if app.UploadDir == ""
```

...because it always fails. A default DDEV project has `app.UploadDir` set to _something_ (or maybe null?) which causes that check to fail, which then causes us to not be able to default it to `files/` which then causes the test project to fail, because it's expecting a file to be in `upload_dir` (which in the project is `files/`)

The only other option I can think of is some way to default _just the test project_ to use an `upload_dir` of `files/` but I'm not sure how to do that, and I'm sure involves some redoing of the test project?

khalwat: All that said, if you want to instead change the testing framework, that's fine too. I'm just thinking that could require a good bit more work, in that we'd have to ensure that it doesn't break any of the other test projects, or refactor the other test projects as well.

rfay:
Please take a look at #4317, thanks! We'll see how it does on everything on the CircleCI tests. thanks so much for everybody's work and attention to this. Closing this one.

PR #4307, opened 2022-10-18
Bugfix TYPO3 version detection to fully support semver

gilbertsoft:
## The Problem/Issue/Bug:

The version detection for dev version fails currently, only major.minor.patch version are supported. Installing a dev version of TYPO3 leads to a panic.

The version is now parsed like done in the TYPO3 release app Darth see https://github.com/TYPO3/darth/blob/f807221d2f4619e464813bc6159a603ecce805c5/conf/release.yaml#L15. Afterwards the extracted version string is parsed using the semver package to avoid errors.

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -5,9 +5,9 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"text/template"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/drud/ddev/pkg/archive"
 	"github.com/drud/ddev/pkg/fileutil"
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks so much for diagnosing and fixing.

I was able to recreate the panic and confirm that it doesn't panic with this code.

I added one commit to add extra info to the debug and warning statements, because I had a few T3 projects and couldn't sort out which ones the output was coming from. See if you think they're OK.

I'll plan to do a point release with this soon.

gilbertsoft:
> I added one commit to add extra info to the debug and warning statements, because I had a few T3 projects and couldn't sort out which ones the output was coming from. See if you think they're OK.

Your changes are fine.

Issue #4190, opened 2022-09-07
Single upload_dir is too limiting, doesn't work for lots of CMSs, make it upload_dirs


ursbraem:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It is recommended to exclude the assets directory from mutagen and replace it with a docker bind-mount via `upload_dir`. Alas, this allows only one directory, but quite a few sites have multiple asset directories / file storages. 

Also, the current upload_dir must be in a subdirectory of the docroot; it should not be constrained like that. For example, Drupal's private files should be outside the docroot.

### Describe your solution

Add a new parameter 

```
# upload_dirs list with paths relative to project root
upload_dirs:
- this
- that
- those
```

and deprecate `upload_dir`

### Describe alternatives

I've tried the manual method, but fear it's not cross-platform (I had to hardwire paths from my system for it to work, and I don't know what happens if you're on linux and don't use mutagen). 

Also, it's a quite complex and ddev can't manage `mutagen.yml` anymore by itself.

### Additional context

https://discord.com/channels/664580571770388500/1014794399009476658

rfay:
This is a good idea, and would solve the private_files problem for drupal, and the multiple upload_dir problem of craft cms as well, and I think TYPO3 for some setups. 

It would require adding a flag and logic to `ddev import-files` to specify where you're importing to, and also the logic to completely change a config value.

mikesnoeren:
Are there any plans to add this feature in the future? Would be really convenient for a lot of folks.

johndwells:
I've stumbled upon this ticket while researching switching from Mac Docker For Desktop to Colima, which recommends Mutagen support. I work with Craft CMS, and @rfay as you rightly said, Craft can be configured with multiple local user-generated directories. So it would indeed be great to be able to configure multiple upload directories to run Mutagen & Colima optimally.

riconeitzel:
I noticed that it's only possible to create this inside docroot. For many applications it makes sense to have that outside of docroot. Should I create a separate ticket for it?

rfay:
@riconeitzel Your comment here is adequate, I'll add to the OP as well.

gilbertsoft:
> I noticed that it's only possible to create this inside docroot. For many applications it makes sense to have that outside of docroot. Should I create a separate ticket for it?

@riconeitzel you should already be able to configure a upload dir outside the docroot like this `ddev config --upload-dir=../myfiles`

[...]

PR #4317, opened 2022-10-20
upload_dir support in Craft CMS, fixes #4302

### LINKED ISSUE #4302: Project type `craftcms` doesn't listen to `upload_dir` config setting.

rfay:
## The Problem/Issue/Bug:

* #4302 
* Related attempt and conversation in #4303

* Remove the wired upload_dir
* Remove the wired docroot (seems inappropriate? DDEV will autodetect, and people should be able to override?)
* Support specifying upload_dir in tests

## TODO:
- [x] The Craft CMS test in TestDdevFullSiteSetup is leaving cruft in the project directory (.env file) - need to fix that.

- [ ] `ddev config --upload-dir=files` on a craft project and see if it works with mutagen and `ddev start` - it should not warn. 

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4317"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -452,7 +452,7 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
         ```bash
         mkdir my-craft-project
         cd my-craft-project
-        ddev config --project-type=craftcms
+        ddev config --project-type=craftcms --docroot=web --create-docroot
         ddev composer create -y --no-scripts --no-install craftcms/craft
         ddev start
         ddev composer update
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -68,7 +68,7 @@ func init() {
 			settingsCreator: createBackdropSettingsFile, uploadDir: getBackdropUploadDir, hookDefaultComments: getBackdropHooks, apptypeSettingsPaths: setBackdropSiteSettingsPaths, appTypeDetect: isBackdropApp, postImportDBAction: backdropPostImportDBAction, configOverrideAction: nil, postConfigAction: nil, postStartAction: backdropPostStartAction, importFilesAction: backdropImportFilesAction, defaultWorkingDirMap: docrootWorkingDir,
 		},
 		nodeps.AppTypeCraftCms: {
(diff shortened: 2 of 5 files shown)


khalwat (changes requested):
So it looks good... the wired `web/` is set that way because that's what the default Craft CMS project sets is set to:

https://github.com/craftcms/craft

So I think in this case, it's a sane default, and eliminates having to specify the `--docroot=web` for every project. This is a case where Craft does have a baked in default.

If DDEV will auto-detect `web/` then sure, it can be removed.

That's assuming the override via `.yaml` works as expected.

The rest of it looks good, though!

rfay:
There is no need to set `--docroot=web` unless the code does not yet exist. That's the case in the composer-build quickstart. It will be autodetected otherwise. 

But people certainly may put the project in a subdir or have a different docroot, and the existing code would have blown away their configuration I think, right?

khalwat:
@rfay okay I didn't realize that DDEV would auto-detect `web/` if so then I agree, it's redundant/pointless.

As for the subdir, I'm assuming the check that was in there `if app.Docroot == ""` works, because people are using it (see recent docs PR).

But anyway yeah moot point if DDEV auto-detects `web/` regardless! Looks good!

rfay:
tested with `ddev import-files` with upload_dir set; it imported properly. mutagen/mutagen.yml was properly edited.

riconeitzel:
I'm researching a upload_dir set behaviour and stumbled upon this issue. was part of the issue that upload_dir will always be a sub directory of the docroot (which should not be) or was that behaviour introduced here? From reading the code it doesn't look like.

bencroker:
@riconeitzel The issue here is that Craft can be configured with multiple local user-generated directories, so there’s little point in hard-coding any single `upload_dir`. Read more about the general issue at https://github.com/ddev/ddev/issues/4190.

bencroker:
@riconeitzel The issue here is that Craft can be configured with multiple local user-generated directories, so there’s little point in hard-coding any single `upload_dir`. Read more about the general issue at https://github.com/ddev/ddev/issues/4190.

PR #4330, opened 2022-10-25
Fix TYPO3 project detection to work with non existent symlinks, fixes #4326

### LINKED ISSUE #4326: Missing AdditionalConfiguration.php with TYPO3 v8 non-composer mode

_No response_

### Expected Behavior

`AdditionalConfiguration.php` is generated.

### Actual Behavior

The file is missing, hence the site does not work.

### Steps To Reproduce

- classic installation of TYPO3 v8 (https://github.com/TYPO3/typo3/blob/TYPO3_8-7/INSTALL.md#installation)

- `typo3_src` folder being mounted into the container like below

docker-compose.local.yaml:
```
      - type: bind
        source: '/mnt/c/TYPO3/CMS/TYPO3_8-7/'
        target: '/var/www/html/typo3_src'
```

### Anything else?

ddev 1.21.3

ddev config:

```
name: redacted
type: typo3
docroot: html
php_version: "7.3"
webserver_type: apache-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
additional_fqdns: []
database:
  type: mariadb
  version: "10.3"
nfs_mount_enabled: false
mutagen_enabled: false
use_dns_when_possible: true
timezone: Europe/Vienna
composer_version: "2"
web_environment:
  - TYPO3_CONTEXT=Development/ddev
nodejs_version: "16"
```

gilbertsoft:
## The Problem/Issue/Bug:

If the `typo3` folder is a symlink to a non existing target the detection fails currently.

If the first check fails, an additional check for an existing symlink is introduced.

- Set up a TYPO3 according the [legacy installation](https://docs.typo3.org/m/typo3/tutorial-getting-started/main/en-us/Installation/LegacyInstallation.html)
- Rename the source folder extracted from the package
- Remove `AddintionalConfiguration.php`

The `AddintionalConfiguration.php` is not longer created during `ddev start` or `ddev restart`.

## Related Issue Link(s):

#4326 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4330"></a>

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -152,9 +152,18 @@ func setTypo3SiteSettingsPaths(app *DdevApp) {
 
 // isTypoApp returns true if the app is of type typo3
 func isTypo3App(app *DdevApp) bool {
-	if _, err := os.Stat(filepath.Join(app.AppRoot, app.Docroot, "typo3")); err == nil {
+	typo3Folder := filepath.Join(app.AppRoot, app.Docroot, "typo3")
+
+	// Check if the folder exists, fails if a symlink target does not exist.
+	if _, err := os.Stat(typo3Folder); !os.IsNotExist(err) {
 		return true
 	}
+
+	// Check if a symlink exists, succeeds even if the target does not exist.
+	if _, err := os.Lstat(typo3Folder); !os.IsNotExist(err) {
+		return true
+	}
+
 	return false
 }
 


liayn:
Works perfectly!!

rfay (approved):
I tested this manually and it's good and correct.

I'm not entirely sure that we should be writing code for broken symlinks. 

Thanks so much!

liayn:
> I'm not entirely sure that we should be writing code for broken symlinks.

As I talked to @gilbertsoft in the TYPO3 Slack: This is actually only necessary, as we currently use the `isTYPO3App` function (aka "guess if it's a TYPO3", for generating the configuration.
It would be better in this case to trust what has been provided by the user in ddev config.yaml.
But I don't know the details in the ddev well enough to decide, whether this is doable at this point.

PR #4336, opened 2022-10-25
[docs] Image cleanup

mattstein:
## The Problem/Issue/Bug:

Images could generally use some tidying.

## How this PR Solves The Problem:

- Uses high-DPI-friendly SVG logo in readme. 
- Consolidates screenshots and images into `content/images/` directory for easier maintenance, consistently using lowercase filenames with hypens (`-`) rather than underscores (`_`) as word separators. (Markdown references updated.)
- Deletes a few unused images.
- Adds alt text in a few cases where it was missing or not meaningful.
- Losslessly optimizes all PNG files for ~50% savings in file weight.
- Crops a number of screenshots to exclude useless cruft.
- Adds step callouts to `/sbin/nfsd` selection screenshot.
- Updates container diagram to remove redundant notes, correct casing, prioritize important pieces, and drastically reduce file weight while staying high-DPI-friendly.
- Fixes squished favicon and adds a multi-resolution ICO variant.

## Manual Testing Instructions:

https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes

## Automated Testing Overview:

n/a

## Related Issue Link(s):

More #4188.

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4336"></a>

PR #4337, opened 2022-10-25
Gitignore `.DS_Store` and similar files

mattstein:
## The Problem/Issue/Bug:

One might easily and accidentally commit those annoying macOS `.DS_Store` files to their commit, adding cruft to the project.

This uses the `.gitignore` file to prevent those from being noticed or committed by Git.

🤔 Stage this change, then purposefully try to commit a file named `.DS_Store`. It’ll be trickier, at least.

n/a

## Related Issue Link(s):

Faintly related to #4188; came up during #4336.

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4337"></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@
 /.ddev
 /artifacts
 *.DS_Store
+nbproject/
+.idea
+*.orig
+*.rej
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /dist
 /.ddev
 /artifacts
+*.DS_Store


rfay (approved):
Nothing wrong with that, but I gitignore stuff like that in global gitignore, I recommend to you as well. 

My global .gitignore:
```
.project
.buildpath
.settings
nbproject/
.idea
.DS_Store
db/db_*
docroot/core/phpunit.xml
docroot/drush
docroot/sites/simpletest
docroot/modules/contrib/devel
sites/default/files
*.orig
*.rej
.tarballs
```

mattstein:
Oh, that makes more sense! Thank you.

tyler36:
Yes, a global `.gitignore` is definately great.

However, I think for a project like DDEV that recieves lots of contributions, it's helpful to have a project level one too.

rfay:
May want to add a couple more of those if we're doing this one.

mattstein:
I mean now that we’ve trained me to use a global file, I won’t be noticing whatever those other things are! 😅 🚀

rfay:
Another pro tip: Always use `git add -u` when choosing changes for a commit. Then look and see what might be left over with `git diff`. The leftover stuff is normally cruft, unless you explicitly added a file, in which case `git add <file>`.  If you use `git add -u` you'll almost never check in something you shouldn't have.

.gitignore
```diff
@@ -10,3 +10,4 @@
 /dist
 /.ddev
 /artifacts
+*.DS_Store
```
rfay, line 13:
```suggestion
*.DS_Store
nbproject/
.idea
*.orig
*.rej
```

rfay (changes requested):

PR #4339, opened 2022-10-27
Setting the MySQL transaction isolation level (Fixes #4281)

### LINKED ISSUE #4281: Drupal 9/10 & Mysql: "MySQL transaction isolation level"

While testing the lastest head version (HEAD-c035fa2) with the latest Drupal version (10.1.0-dev), I noticed a new warning:

> Database Isolation Level
> REPEATABLE-READ
> For the best performance and to minimize locking issues, the READ-COMMITTED transaction isolation level is recommended. See the setting [MySQL transaction isolation level](https://www.drupal.org/docs/system-requirements/setting-the-mysql-transaction-isolation-level) page for more information.

Link: https://www.drupal.org/docs/system-requirements/setting-the-mysql-transaction-isolation-level

I made the following changes to `web/sitges/default/settings.ddev.php`  
  - remove `#ddev-generated`
  - add the following to database array

```php
$databases['default']['default'] = [
  ...
  'init_commands' => [
    'isolation_level' => 'SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED',
  ],
];
```

This change removed the warning.

Drupal 10 is due out in 2 months, on Dec 2022.

DDEV: `HEAD-c035fa2`
Drupal: `10.1.0-dev`

### Describe your solution

The config should be added to when
- database type = MySQL or MariaDB
- project type = Drupal 10 

### Describe alternatives

_No response_

### Additional context

_No response_

tyler36:
## The Problem/Issue/Bug:

This resolves the issue of #4281 which requires MySQL tranaction isolation level setting.

Add the required line to 'mysql' driver configurations. 

- Note: I added it on a Drupal 10 Postgres system too which appeared to load correctly, but am unaware of any side affects under "normal" usage.

Comment out line in a Drupal 10 install and a warning appears on `/admin/reports/status`

## Related Issue Link(s):
Fixes #4281

[Setting the MySQL transaction isolation level](https://www.drupal.org/docs/system-requirements/setting-the-mysql-transaction-isolation-level)
[Suggestion for: (3285800) Setting the MySQL transaction isolation level](https://www.drupal.org/project/documentation/issues/3308909)

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4339"></a>

### DIFF
--- a/pkg/ddevapp/drupal/drupal10/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal10/settings.ddev.php
@@ -28,6 +28,13 @@ $databases['default']['default'] = array(
   'prefix' => "{{ $config.DatabasePrefix }}",
 );
 
+
+// Setting the MySQL transaction isolation level.
+// @see https://www.drupal.org/docs/system-requirements/setting-the-mysql-transaction-isolation-level
+if ($driver === "mysql") {
+  $databases['default']['default']['init_commands']['isolation_level'] = 'SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED';
+}
+
 $settings['hash_salt'] = '{{ $config.HashSalt }}';
 
 // This will prevent Drupal from setting read-only permissions on sites/default.
--- a/pkg/ddevapp/drupal/drupal9/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal9/settings.ddev.php
@@ -28,6 +28,13 @@ $databases['default']['default'] = array(
   'prefix' => "{{ $config.DatabasePrefix }}",
 );
 
+
+// Setting the MySQL transaction isolation level.
+// @see https://www.drupal.org/docs/system-requirements/setting-the-mysql-transaction-isolation-level
+if ($driver === "mysql") {
+  $databases['default']['default']['init_commands']['isolation_level'] = 'SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED';
+}
+
 $settings['hash_salt'] = '{{ $config.HashSalt }}';
 
 // This will prevent Drupal from setting read-only permissions on sites/default.


rfay (changes requested):
Thanks for this. I think we should do it for Drupal 9 as well. In fact, @rpkoller, is this also absolutely required by drupal 9.5.x?

tyler36:
Add to Drupal 9 and rebased.

rfay:
I'm going to take a look too and see if there's any way we can do this with extra config, that would represent what would be done in the real world. I imagine what we have here is what will work out easiest and best though.

rpkoller:
here is the change record about the mysql isolation level and the warning if it isn't using READ COMMITTED:  https://www.drupal.org/node/3269885 . according to the CR will be introduced in 9.5.0

rfay (changes requested):
When making a PR, try to change only the things that are relevant to the topic of the PR. If you want to make other changes (if "TRUE" should be "true" due to change in style preferences?) then make those changes in another PR that is style-related.

I'm pretty sure the changes to the go templating here introducing the spaces between members of the go template is unnatural, not sure if it's a syntax error. But that again is something that shouldn't be introduced if not related to the goal of the PR. 

Overall, though, since this approach is currently recommended in the Drupal Change Notice, I think this approach will be fine.

tyler36:
Sorry, Vscode tried to format on save.

I thought the following setting was suppose to prevent that but it doesn't appear to be working.
```json
"editor.formatOnSaveMode": "modifications",
```

rfay:
Those EVIL editors! Sometimes they can help so much... and other times not so much.

rfay (approved):
I tested this on Drupal 9.4 and Drupal 10 and it behaved as expected. Thanks!

I'm a bit reluctant, as discussed before, I don't think the Drupal recommendations are correct, that the config change should be in the mysql config, not done like this. But I'm good with this because it's the Drupal-recommended way and it's also the easiest for us to do. 

When I look at the postgres database requirement I'll be thinking about this.

rfay:
When @rpkoller approves we're good to go.

rpkoller:
hm i am not really qualified for reviewing that. ;) but i've manually tested and added the snippet to one of my existing sites (removed the cnf file beforehand and restarted). adding the according lines set the transaction isolation level and commenting it out set the level back to REPEATABLE-READ. 

the only detail i wonder is, does it have certain performance implications to set the level on the drupal settings level - that way isn't it set on every request? with the cnf solution it would be set once on startup on the database level instead? 

but as @rfay said it is still the current Drupal-recommended best practice therefore a plus one from me. but talking of still, have to see  who best to ping and ask on the drupal end about the underlaying issue. cuz since yesterday when i've learned searching for the change record the issue isnt just limited to the docs page but also the change record itself. there the recommendation is used as well. maybe the best place might be the d10readiness meetings?

rpkoller (approved):
hm i am not really qualified for reviewing that. ;) but i've manually tested and added the snippet to one of my existing sites (removed the cnf file beforehand and restarted). adding the according lines set the transaction isolation level and commenting it out set the level back to REPEATABLE-READ.

the only detail i wonder is, does it have certain performance implications to set the level on the drupal settings level - that way isn't it set on every request? with the cnf solution it would be set once on startup on the database level instead?

but as @rfay said it is still the current Drupal-recommended best practice therefore a plus one from me. but talking of still, have to see who best to ping and ask on the drupal end about the underlaying issue. cuz since yesterday when i've learned searching for the change record the issue isnt just limited to the docs page but also the change record itself. there the recommendation is used as well. maybe the best place might be the d10readiness meetings?

p.s. first added the comment to the general conversation until i found where to approve changes. haven't expected that a comment is required there as well. therefore deleted the comment in the general conversation and readded it here.

rfay:
In the future we have two options:
1. Add a post-start action like is done for pg_trgm, https://github.com/drud/ddev/pull/4347 (which would implement the *other* drupal-recommended approach)
2. Add per-project-type database configuration (my.cnf) which hasn't been necessary up to now, but maybe it's coming.

rfay:
Thanks for making this happen @tyler36 and thanks for all the informed direction on it @rpkoller .

PR #4348, opened 2022-10-29
Add postgres:15 support

rfay:
## The Problem/Issue/Bug:

Postgresql 15 was [released on Oct 13, 2022](https://www.postgresql.org/about/news/postgresql-15-released-2526/#:~:text=October%2013%2C%202022%20%2D%20The%20PostgreSQL,most%20advanced%20open%20source%20database.)

So this adds support

Try out `ddev config --database=postgres:15`

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4348"></a>

### DIFF
--- a/docs/content/users/configuration/config_yaml.md
+++ b/docs/content/users/configuration/config_yaml.md
@@ -81,7 +81,7 @@ The type and version of the database engine the project should use.
 
 | Type | Default | Usage
 | -- | -- | --
-| :octicons-file-directory-16: project | MariaDB 10.4 | Can be MariaDB 5.5–10.7, MySQL 5.5–8.0, or PostgreSQL 9–14.<br>See [Database Server Types](../extend/database_types.md) for examples and caveats.
+| :octicons-file-directory-16: project | MariaDB 10.4 | Can be MariaDB 5.5–10.7, MySQL 5.5–8.0, or PostgreSQL 9–15.<br>See [Database Server Types](../extend/database_types.md) for examples and caveats.
 
 !!!note
     DDEV v1.18 and earlier used `mariadb_version` and `mysql_version`, which are now automatically converted to the `database` format.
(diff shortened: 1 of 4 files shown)


rpkoller (approved):
i've only manually tested. one detail i'Ve noticed. when starting a project with postgres15 for the first time i get the following: 

``` 
15: Pulling from library/postgres
dd6189d6fc13: Already exists 
e83a243abe4a: Already exists 
e1e5d0e9701b: Already exists 
b8172b349685: Already exists 
83dad09f3014: Already exists 
c4849d0ca437: Already exists 
7642d443be37: Already exists 
a40f15729374: Already exists 
d965a7635506: Downloading [==========================>                        ]  46.69MB/88.34MB
aa94e54c6bb6: Download complete 
6d955f7fa1ab: Download complete 
3cd6711cf4eb: Download complete 
2899081d3dc1: Download complete 
```
the "already exists" parts look odd. havent downloaded postgres 15 before? only postgres 9,10 and 14. and when i try to install drupal 10.1.x-dev with postgres 15 the `pg_trgm` extension is missing and i am unable to proceed. 

Screen Shot 2022-10-29 at 13 03 39

added the extension manually with `ddev psql -c 'CREATE EXTENSION pg_trgm;'`and then was able to proceed. status report had no complaints at the end: 

Screen Shot 2022-10-29 at 13 08 03

rfay:
Thanks @rpkoller - "already exists" on a docker layer means that it has already been downloaded as part of a different docker image. So postgres:14 and postgres:15 images share some layers (the OS they're built on).

rpkoller:
oh neat. i've always thought docker would download complete image for every major release. but that way it saves bandwidth. interesting! thanks for the explanation.

rfay:
For ddev-webserver the image is consolidated into one big image, which reduces the total size, but OTOH it doesn't allow this kind of shared layers, and doesn't allow parallel download/extraction of layers.

PR #4360, opened 2022-11-02
[docs] Suggest using gitpod in CMS quickstarts

ibrah3m:
## The Problem/Issue/Bug:
testing area for devlopers
## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4360"></a>


docs/content/users/quickstart.md
```diff
@@ -16,6 +16,11 @@ Once DDEV is installed, you can quickly spin up new projects:
 
 DDEV comes ready to work with any PHP project, and has deeper support for several common PHP platforms and content management systems.
 
+!!!tip
+    If you're looking for place to test all these examples and you don't have time to install docker or create new workspace then hit the following button.
+
+    [![New DdevSpace](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/ibrah3m/DdevSpace.git)
+
```
tyler36, line 23: While I can see value in providing a generic playground area, I think it should be within a `[drud/ddev](https://github.com/drud/ddev)` namespace. This would allow people DDEV-authorized users to manage as appropriate.

ibrah3m: how to make this happen?

tyler36: @rfay will probably decide if its needed or not, especially if we used the offical domain.

Personally, I think it would be helpful if we had a repo for each supported framework.
That would allow anyone to instantly spin up a Drupal, Laravel, CraftCMS, etc. which would also make it easy to test issues.

rfay (changes requested):
Sorry, I don't think we should introduce gitpod here. If people aren't already using it, they have to do a little setup. I think this will be a distraction.

ibrah3m:
> Sorry, I don't think we should introduce gitpod here. If people aren't already using it, they have to do a little setup. I think this will be a distraction.

well gotcha, actually I had an Idea to fork the ddev-launch repo in order to PR a new feature that can specify the project type through one of the environment variables passed through the URL to match the idea that @tyler36  mentioned: "I think it would be helpful if we had a repo for each supported framework." .
anyway just let me know if you decide to use gitpod in the cms quickstart.

rfay:
It's just that people are already wedded to what they're using. Gitpod is a fantastic tool, but I think people are better to use the tool they're familiar with here. Thanks for the good thinking!

PR #4364, opened 2022-11-03
[docs] apt-key is deprecated

mhuber84:
## The Problem/Issue/Bug:
apt-key is deprecated

Update documentation with suggestion from https://itsfoss.com/apt-key-deprecated/

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -53,8 +53,8 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     DDEV’s Debian and RPM packages work with `apt` and `yum` repositories and most variants that use them, including Windows WSL2:
 
     ```bash
-    curl -fsSL https://apt.fury.io/drud/gpg.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/ddev.gpg > /dev/null
-    echo "deb [signed-by=/etc/apt/trusted.gpg.d/ddev.gpg] https://apt.fury.io/drud/ * *" | sudo tee /etc/apt/sources.list.d/ddev.list
+    curl -sS https://apt.fury.io/drud/gpg.key | gpg --dearmor | sudo tee /usr/share/keyrings/ddev.gpg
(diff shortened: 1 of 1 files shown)


rfay:
Hmm, I thought this was already fixed everywhere with 
* https://github.com/drud/ddev/pull/4324

mhuber84:
Sorry, my fault. I haven't noticed that the text here in the repository already doesn't use apt-key anymore. On https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/ are the old instructions. So my pull request is not needed anymore because of the apt-key deprecation.
But https://itsfoss.com/apt-key-deprecated/ says, that there are two ways to import the GPG keys and this way is the proper way. But I don't know if this is really important. In my tests both ways work.

rfay:
Hmm, https://itsfoss.com/apt-key-deprecated/ puts does what the DDEV docs already do, so I'm confused. 

Anyway, we'll close this and please do report if you have any issues with the current recommendation.

Thanks for taking the time to help sort it out! Your docs improvements and other contributions are always welcome.

mhuber84:
> The article you quote says the keys belong in /etc/apt/trusted.gpg.d but for some reason you're changing that.

"Solution 2: Adding the GPG keys to the system to the proper way" says /usr/share/keyrings/. But it doesn't explain why and I'm not that experienced in the linux directory structures.

rfay:
Yet everywhere in the article they use /etc/apt/trusted.gpg.d :( I think it's just a confusing article. 

A quick look around the internet shows it being done into both of these directories. I guess we'll live and learn.

FoxtrotCZ:
Just wanna say, Docker uses etc/apt/keyrings
https://docs.docker.com/engine/install/ubuntu/
Could be good idea to use the same location.

rfay:
If you have something more authoritative than just what docker does, I'm sure open to it. 

The error message says 
> Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).

It looks to me like Ubuntu (and probably Debian) say 

> Note: Instead of using this command a keyring should be placed directly in the
           [/etc/apt/trusted.gpg.d/](file:///etc/apt/trusted.gpg.d/) directory with a descriptive name and either "gpg" or "asc" as
           file extension.

https://manpages.ubuntu.com/manpages/jammy/man8/apt-key.8.html

PR #4366, opened 2022-11-03
[docs] Add docker-inside script for wsl2 [skip ci]

rfay:
## The Problem/Issue/Bug:

The instructions have been way too hard for wsl2 with docker inside

Wrap it up into a script

## TODO

- [x] If Win11, boot docker on startup via /etc/wsl.conf
- [x] If WIn10, tell them what to do to start docker-ce `sudo service docker start`

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4366"></a>


rfay:
@mattstein I should have asked for your review on the previous PR, so you'll likely want to look at the entire WSL2 section. I guess due to the dynamics it will take a PR to fix things you don't like in the WSL2-with-docker-desktop section.

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     * Installing Chocolatey package manager (optional).
     * One time initialization of mkcert.
     * Installing WSL2 and installing a distro like Ubuntu.
-    * Installing or upgrading to the latest Docker Desktop for Windows with WSL2 enabled.
+    * (Optionally) Installing Docker Desktop for Windows and enabling WSL2 integration with the distro.
```
mattstein, line 117: Move “optional” to the end in the style of the first bullet:

```suggestion
    * Installing Docker Desktop for Windows and enabling WSL2 integration with the distro (optional).
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
     * Installing DDEV inside your distro.
 
+    ### WSL2 with Docker-ce Inside Scripted Installation
```
mattstein, line 120: Refer to [Docker CE](https://docker-docs.netlify.app/install/), shorten heading, and use the phrase “install script” to be more consistent with [the rest of this page](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#system-requirements):

```suggestion
    ### WSL2 + Docker CE Inside Install Script
```

On the fence about the word “inside” here, but I assume it’s critical to point out that Docker CE will live inside the WSL2 distro.

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+
+    This scripted installation prepares your default WSL2 Ubuntu distro for use and has no dependency on Docker Desktop.
```
mattstein, line 122: Omit “for use” because it’s safe to assume we’re preparing something in order to take further steps with it.

```suggestion
    This scripted installation prepares your default WSL2 Ubuntu distro and has no dependency on Docker Desktop.
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+
+    You can do these things manually, or you can do most of it with the provided PowerShell script.
```
mattstein, line 124: Start with the preferred/easiest suggestion:

```suggestion
    The provided PowerShell script can do most of the work for you, or you can handle these things manually. 
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
     * Installing DDEV inside your distro.
 
+    ### WSL2 with Docker-ce Inside Scripted Installation
+
+    This scripted installation prepares your default WSL2 Ubuntu distro for use and has no dependency on Docker Desktop.
+
+    You can do these things manually, or you can do most of it with the provided PowerShell script. 
+    In all cases:
+    
+    1. Install WSL2 with an Ubuntu distro. On a system without WSL2, just run:
```
mattstein, line 127: [Omit “just”](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/#avoid-just-and-easy):

```suggestion
    1. Install WSL2 with an Ubuntu distro. On a system without WSL2, run:
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+        ```powershell
+        wsl --install
+        ```
+
+        Verify that you have an Ubuntu distro set to default with `wsl -l -v`.
```
mattstein, line 132: Explicitly clarify that the reader’s supposed to run a command:

```suggestion
        Verify that you have an Ubuntu distro set to default by running `wsl -l -v`.
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+
+        If you already have WSL2 but don't have an Ubuntu distro, install one with `wsl --install Ubuntu`.
```
mattstein, line 134: Explicitly clarify that the reader’s supposed to run a command:

```suggestion
        If you already have WSL2 but don't have an Ubuntu distro, install one by running `wsl --install Ubuntu`. 
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+        wsl --install
+        ```
+
+        Verify that you have an Ubuntu distro set to default with `wsl -l -v`.
+
+        If you already have WSL2 but don't have an Ubuntu distro, install one with `wsl --install Ubuntu`. 
+
+        If that doesn't work for you, see the [manual installation](https://docs.microsoft.com/en-us/windows/wsl/install-manual) and linked [troubleshooting](https://docs.microsoft.com/en-us/windows/wsl/troubleshooting#installation-issues).
+        
+        If you prefer to use another Ubuntu distro, just install it and set it as default. For example, `wsl --set-default Ubuntu-22.04`.
```
mattstein, line 138: [Omit “just”](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/#avoid-just-and-easy):

```suggestion
        If you prefer to use another Ubuntu distro, install it and set it as default. For example, `wsl --set-default Ubuntu-22.04`.
```

docs/content/users/install/ddev-installation.md
```diff
@@ -114,9 +114,38 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
...
+    2. Run [this PowerShell script](https://raw.githubusercontent.com/drud/ddev/master/scripts/install_ddev_wsl2_docker_inside.ps1) in an administrative PowerShell Window to complete DDEV WSL2 with Docker Desktop installation:
+
+        ```powershell
+        Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; 
+        iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/drud/ddev/master/scripts/install_ddev_wsl2_docker_inside.ps1'))
+        ```
+
+    Now you can use the "Ubuntu" terminal app or Windows Terminal to access your Ubuntu distro, which has DDEV and Docker working in it.
+
     ### WSL2 with Docker Desktop Scripted Installation
```
mattstein, line 149: Match heading style and “install script” with the previous one:

```suggestion
    ### WSL2 + Docker Desktop Install Script
```

docs/content/users/install/ddev-installation.md
```diff
@@ -193,7 +222,7 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
 
     ## Traditional Windows
 
-    DDEV does work fine on the Windows side, although it’s quite a bit slower than WSL2 by default. Good results have been reported by users who enabled Mutagen.
+    DDEV does work fine on the Windows side, although it’s quite a bit slower than WSL2 by default. You'll want to enable Mutagen or NFS for tolerable performance. Most users find that Mutagen is the most satisfactory technique. See [Performance](performance.md) for more information, but `ddev config global --mutagen-enabled` is the command to enable mutagen.
```
mattstein, line 225: Get to the point with better links:

```suggestion
    DDEV works fine on the Windows side, but it’s slower than WSL2 by default. Enable either [Mutagen](https://ddev.readthedocs.io/en/stable/users/install/performance/#system-requirements) or [NFS](https://ddev.readthedocs.io/en/stable/users/install/performance/#nfs) for the best performance.
```

In this case, the explanation gets in the way of the instruction:

1. The performance page (and lots of other blurbs) heavily emphasize Mutagen over NFS, so there’s no need to offer commentary here on what most users find best. (We’re leading with that recommendation anyway.)
2. The Mutagen link gets right to the point with that command to enable, so there’s no need to be redundant with it here. (It’s also strangely prescriptive to recommend Mutagen or NFS but only offer the command for Mutagen—might as well link thoughtfully and encourage the reader to either pick the quick + obvious answer or read further and make the choice they’re happiest with. No need to broadly link to the performance page and short-circuit their trip in one breath while distracting from further Windows install notes.)

mattstein (changes requested):
I’ve [been steering clear](https://github.com/drud/ddev/pull/4200#issuecomment-1242996437) of this section until now, so I only left feedback on the changes here. Can swoop through again in a separate PR as you suggested.

docs/content/users/install/ddev-installation.md
rfay:
```suggestion
    * Installing Docker Desktop for Windows and enabling WSL2 integration with the distro (optional, do this if you're using the Docker Desktop approach).
```

docs/content/users/install/ddev-installation.md
rfay: Links should always be relative in the docs, not link out to readthedocs.

```suggestion
    DDEV works fine on the Windows side, but it’s slower than WSL2 by default. Enable either [Mutagen](performance/#system-requirements) or [NFS](performance/#nfs) for the best performance.```

mattstein: Yikes, I don’t know what came over me. Thank you for fixing!

rfay:
Thanks @mattstein. Just the one thing to make sure about, use relative links in docs to docs, not readthedocs links.

I'm a little sad about not giving to-the-point statements about simple things like mutagen. Especially with mutagen, people assume they have to do something significant or install something or understand something significant, which they don't. So that's the reasoning for showing a simple command when it's all they have to know.

mattstein:
> Thanks @mattstein. Just the one thing to make sure about, use relative links in docs to docs, not readthedocs links.
 
Of course! I know better, I just had a momentary lapse of good sense for some reason.

> I'm a little sad about not giving to-the-point statements about simple things like mutagen. Especially with mutagen, people assume they have to do something significant or install something or understand something significant, which they don't. So that's the reasoning for showing a simple command when it's all they have to know.

Those comments are yours to leave if you’d rather, I just think they get in the way of the subject at hand. A thoughtful link helps maintain the focus of the current page, puts emphasis on making a clear case in the right place, and subtly helps someone learn how to find their way around the documentation in general. The more effort there is in thoughtful structuring, the more that pays off.

Clear structure and purpose is teaching someone to fish. Saying the same thing in every relevant place they might end up is throwing them a fish (and creating some ... _fish debt_?).

rfay:
Totally understood, thanks.

rfay:
And I understand the reason for just linking much of the time as well.

mattstein:
> A thoughtful link

I will point out that I’m the same person who un-thoughtfully used a non-relative link. As you’ve pointed out, life is hard.

rfay:
I think this is ready to go in. Currently pantheon docs are not responding so docs check fails.

I have tested both scripts on both clean WIndows 11 and Windows 10 machines, and they seem to be working fine.

PR #4380, opened 2022-11-14
Warn about usage of Docker Desktop for Linux

rfay:
## The Problem/Issue/Bug:

* People on Linux keep finding trouble with Docker Desktop for Linux. 
* This is mentioned in the docs

Warn them so they maybe they won't use it.

Try to `ddev start` with Docker Desktop for Linux

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -626,6 +626,18 @@ func CheckDockerVersion(versionConstraint string) error {
 		return err
 	}
 
+	// See if they're using broken docker desktop on linux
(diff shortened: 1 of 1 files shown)


rfay:
This doesn't prevent `ddev start` currently; what it does is emit "The docker version currently installed does not seem to meet ddev's requirements" with an explanation.

rfay:
This is adequate:

image

rfay:
I was able to get docker desktop for linux to work by using mutagen, `ddev config --mutagen-enabled`, then had to change some permissions, for example on the upload_dir.

PR #4412, opened 2022-11-29
[docs] Fixing indentation in multiple exposed ports example YAML.

quicksketch:
## The Problem/Issue/Bug:

The code sample at https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/#docker-composeyaml-examples has an indentation problem in the `labels` section.

image

Reformats the whole section to use 2-spaces like the other examples on this page.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4412"></a>

### DIFF
--- a/docs/content/users/extend/custom-compose-files.md
+++ b/docs/content/users/extend/custom-compose-files.md
@@ -31,17 +31,17 @@ That approach usually isn’t sustainable because two projects might want to use
 
 ```yaml
 services:
-    someservice:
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks @quicksketch !

rfay:
Interestingly enough, that can be an array as well, in which case it could either have or not have indentation. But we've always used the object syntax.

quicksketch:
Thanks @rfay!

PR #4418, opened 2022-11-30
Add support for .ddev/.env file, fixes #4309

rfay:
## The Problem/Issue/Bug:

* #4309 

## How this PR Solves The Problem:

Import environment variables from project .ddev/.env file if it exists. 

## Manual Testing Instructions:

Create a .ddev/.env file, `ddev restart`, verify that your env vars appear in the web container.

## Automated Testing Overview:

Added TestEnvFile

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4418"></a>

Issue #4309, opened 2022-10-18
Environment variables in .env file are not supported without extra configuration


flynnt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

Based on the instructions in the [CMS Quickstarts](https://ddev.readthedocs.io/en/stable/users/quickstart/) section of the docs for CraftCMS I expected to be able to set the `CRAFT_CMD_ROOT` variable in my `.env` file and have the `ddev craft` command use the value set there. I am running my Craft application in a subdirectory, `./app`, and having set that as the `CRAFT_CMD_ROOT`, and after restarting DDEV, I expected `ddev craft` to work.

### Actual Behavior

Running `ddev craft` after following the steps in the CMS Quickstart docs to change the `CRAFT_CMD_ROOT` setting yields:

`Could not open input file: craft`.

### Steps To Reproduce

1. Have an existing Craft project running in a subdirectory, e.g. `my-project/app`.
2. `ddev config --docroot=app/web --project-type=craftcms --composer-root=app`
3. `ddev start`
4. Import a DB
5. Update the Craft `.env` with `CRAFT_CMD_ROOT=./app`.
6. `ddev restart`
7. `ddev craft`

### Anything else?

For what it's worth, the main issue seems to be that the container has no awareness of that environment variable when it is set via the `.env` file. The `craft` command does work if you set the variable via `ddev config  --web-environment="CRAFT_CMD_ROOT=./app"`. This _seems_ to be a better way to do it in general, since there's a better separation of concerns between DDEV and Craft in that case. This also seems to be the way that some of the existing global web commands handle environment variables. For example, the `blackfire` command recommends setting any environment variables needed by the script via the DDEV global or project-level config.

rfay:
Thanks for raising this issue. `.env` files are not explicitly supported by DDEV, but they probably will be in the future. See https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/#providing-custom-environment-variables-to-a-container for how to support .env files.

flynnt:
Oh nice. It will be pretty cool once `.env` files are supported. In the meantime, maybe the resolution for this specific issue as it relates to the `craftcms` project type would be to update the CMS Quickstart documentation to not reference the `.env` file as a way to modify the `CRAFT_CMD_ROOT` setting. Specifically this:

[...]

PR #4425, opened 2022-12-02
[docs] Improvements on addition of .ddev/.env file (for #4418)

mattstein:
## The Problem/Issue/Bug:

The docs added in #4418 could use some pedantic nitpicking.

This jumps right in with pedantic nitpicking:

- Fixes a backtick formatting issue.
- Introduces environment-variable-setting methods so it seems less like an unfinished sentence.
- Reorders methods from what we know/expect to what’s new/different.
- Rewords slightly and works in “setting” and “command” phrases to gently reiterate how DDEV works.
- Clarifies that we’re using the `ddev config` command in the example, and links to its reference to again subtly reiterate how DDEV works.
- Reformats `ddev config` command for (hopefully) easier readability.
- Moves sensitive value warning into warning callout and further suggests common `.env.example` convention while we’re covering basic secret hygiene—and gives the callout a heading so those familiar with this can scan and skip.
- Uses `MY_` prefix for user-defined things. (Trying to do that generally to make it obvious what’s DDEV’s vs. the user’s.)

[Preview changes locally](https://ddev.readthedocs.io/en/latest/developers/testing-docs/#preview-changes) or take a look at the [automatic build](https://ddev--4425.org.readthedocs.build/en/4425/users/extend/customization-extendibility/#providing-custom-environment-variables-to-a-container).

n/a

## Related Issue Link(s):

- #4418
- #4188

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4425"></a>

### DIFF
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -113,7 +113,7 @@ web_extra_exposed_ports:
 
 You can set custom environment variables in several places:
 
-* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml or `.ddev/config.*.yaml`:
+* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml` or `.ddev/config.*.yaml`:
 
     ```yaml
     web_environment:
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -113,7 +113,7 @@ web_extra_exposed_ports:
 
 You can set custom environment variables in several places:
 
-* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml`:
+* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml or `.ddev/config.*.yaml`:
 
     ```yaml
     web_environment:
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -113,7 +113,7 @@ web_extra_exposed_ports:
 
 You can set custom environment variables in several places:
 
-* The project’s [`web_environment`](../configuration/config.md#web_environment) config setting in `.ddev/config.yaml`:
+* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml`:
 
     ```yaml
     web_environment:
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -139,6 +139,7 @@ ddev config --web-environment-add="MY_ENV_VAR=someval"
 # Set MY_ENV_VAR globally
 ddev config global --web-environment-add="MY_ENV_VAR=someval
 ```
+
 You can use the `--web-environment` flag to overwrite existing values rather than adding them.
 
 !!!warning "Don’t check in sensitive values!"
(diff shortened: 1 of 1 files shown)


rfay:
You seem to have reverted the essential changes in the PR. 

I guess you say above that you intended to reverse the ordering of options, but let's talk about that. It's my expectation that the .env technique will obsolete the traditional technique of putting variables in config.yaml etc.

mattstein:
> It's my expectation that the .env technique will obsolete the traditional technique of putting variables in config.yaml etc.

I didn’t realize this, though I still think the reorder makes sense since it potentially ends on the strongest suggestion without leaving any out.

My original thought was that, in simplistic terms, I can generally think of DDEV as YAML configuration and commands. So if I need to set an environment variable for a container, or all containers, YAML would be the first thing I’ve trained myself to look for. A `.env` file is something I’m used to seeing with my own app, but not something I’d naturally expect from DDEV—and keeping multiple `.env` files for a project seems relatively uncomfortable.

The first two bullets also build on each other, where the third is the outlier. In sheer terms of presentation, it felt right to have _expected DDEV thing_, _variation on expected DDEV thing_, _other (and newer) thing_.

I would personally stick to one of the first two approaches rather than maintain multiple `.env` files, which may needlessly bias my approach here. If I was sharing my work with a team, for example, and we had to discuss where a Craft-specific setting was defined, I’d rather be able to say “DDEV’s project YAML” or “the dot-env file” without having to further qualify where `.env` values come from. Seems like a minor thing, probably, but one that’s easy to avoid sticking with DDEV-like conventions as I see them.

My understanding with [the original issue](#4309) is that the medium (project YAML vs. `.env`) is irrelevant, but that some environment variables must be explicitly set with DDEV when they’re not picked up from the app’s own `.env` file. This new feature supports a `.ddev/.env` file—as an alternative to using `.ddev/config.yaml`—as a sort of half-measure; you’ll still need to define things on DDEV’s side, but you can copy them straight out of your app’s `.env`. Just saying that out loud in case I’ve got that wrong.

rfay:
OK, I buy it, thanks!

docs/content/users/extend/customization-extendibility.md
```diff
@@ -111,28 +111,39 @@ web_extra_exposed_ports:
 
 ## Providing Custom Environment Variables to a Container
 
-Custom environment variables may be set in
+You can set custom environment variables in several places:
 
-* The project `.ddev/.env file`, which could look something like this:
+* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml`:
```
rfay, line 116:
```suggestion
* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml or `.ddev/config.*.yaml`:
```

mattstein: Ha, I almost got rid of that YAML reference to let the config page explain and make this easier to read, but kept it because the location seemed urgently important in this case. Might as well call this out for technical accuracy!

mattstein:
Pull whenever you’d like!

docs/content/users/extend/customization-extendibility.md
```diff
@@ -111,28 +111,39 @@ web_extra_exposed_ports:
 
 ## Providing Custom Environment Variables to a Container
 
-Custom environment variables may be set in
+You can set custom environment variables in several places:
 
-* The project `.ddev/.env file`, which could look something like this:
+* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml or `.ddev/config.*.yaml`:
```
rfay, line 116:
```suggestion
* The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml` or `.ddev/config.*.yaml`:
```

Guess I messed it up. Linting is valuable!

mattstein: I completely missed that, but ol’ linter didn’t! 🎉

PR #4426, opened 2022-12-02
Platform integration: Download all databases, Allow a primary relationship for the main 'db', fixes #4415, fixes platformsh/ddev-platformsh#59

### LINKED ISSUE #4415: Unable to pull database from platform.sh if more than one exists

It passes, but contains several project/client names so I haven't uploaded it. I can do so if absolutely required.

### Expected Behavior

`ddev pull platform --skip-files` should be able to fetch and import a database dump.

### Actual Behavior

When more than one DB is configured, `platform` expects a parameter for which DB to work with.

```
Authenticating...
Obtaining database...
More than one relationship found.
Use the --relationship (-r) option to specify a relationship. Options:
    legacy
    database
Pull failed: Failed to exec #set -x   # You can enable bash debugging output by uncommenting
set -eu -o pipefail
ls /var/www/html/.ddev >/dev/null # This just refreshes stale NFS if possible
platform db:dump --yes --gzip --file=/var/www/html/.ddev/.downloads/db.sql.gz --project="${PLATFORM_PROJECT}" --environment="${PLATFORM_ENVIRONMENT}"
 on web: exit status 1
```

### Steps To Reproduce

1. Create a project with multiple mysql relationships in platform.sh
2. `.platform.app.yaml`
```
relationships:
  database: 'mysqldb:mysql'
  legacy: 'legacydb:mysql'
  redis: 'rediscache:redis'
```
3. Run `ddev pull platform --skip-files`
4. See error (under **Actual Behavior** above)

### Anything else?

This fails (from inside DDEV):
```
platform db:dump --yes --gzip --file=/var/www/html/.ddev/.downloads/db.sql.gz --project="${PLATFORM_PROJECT}" --environment="${PLATFORM_ENVIRONMENT}" -r 0
More than one relationship found.
Use the --relationship (-r) option to specify a relationship. Options:
    legacy
    database
```

This works (from inside DDEV):
```
platform db:dump --yes --gzip --file=/var/www/html/.ddev/.downloads/db.sql.gz --project="${PLATFORM_PROJECT}" --environment="${PLATFORM_ENVIRONMENT}" -r database
Creating gzipped SQL dump file: /var/www/html/.ddev/.downloads/db.sql.gz
```

See the `-r database` at the end (stands for `--relationship`).

_I understand this is an edge case, but it happened to me so it made sense to file it._

rfay:
## The Problem/Issue/Bug:

* #4415 
* https://github.com/platformsh/ddev-platformsh/issues/59
* https://github.com/drud/ddev/pull/4417
* DDEV was focused on downloading a single database

* Use relationships to download all the upstream databases
* If one should be primary, set it with PLATFORM_PRIMARY_RELATIONSHIP and it will go into the database named 'db'

This also fixes a bug in handling blank `--environment` in `ddev pull`

Try it out on a multisite or whatever, `ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=maindb` or `ddev pull platform --environment=PLATFORM_PRIMARY_RELATIONSHIP=maindb` 

Multiple databases are not yet covered in TestPlatformPull. 

## Release/Deployment notes:

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4426"></a>

### DIFF
--- a/docs/content/users/providers/platform.md
+++ b/docs/content/users/providers/platform.md
@@ -42,6 +42,14 @@ web_environment:
 4. Run `ddev pull platform`. After you agree to the prompt, the current upstream database and files will be downloaded.
 5. Optionally use `ddev push platform` to push local files and database to Platform.sh. The [`ddev push`](../basics/commands.md#push) command can potentially damage your production site, so we don’t recommend using it.
 
+If you have more than one database on your Platform.sh project, you'll need to choose which one you want to use
+as the 'db' primary database on DDEV, and that will be the one pulled or pushed.
+Do this by setting PLATFORM_PRIMARY_RELATIONSHIP, for example,
+
+```
+ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=maindb`
+```
+
 ## Usage
 
 * `ddev pull platform` will connect to Platform.sh to download database and files. To skip downloading and importing either file or database assets, use the `--skip-files` and `--skip-db` flags.
--- a/docs/content/users/providers/platform.md
+++ b/docs/content/users/providers/platform.md
@@ -47,7 +47,13 @@ as the 'db' primary database on DDEV, and that will be the one pulled or pushed.
 Do this by setting PLATFORM_PRIMARY_RELATIONSHIP, for example,
 
 ```
-ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=maindb`
+ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=main`
+```
+
+or run `ddev pull platform` with the `--environment` flag, for example,
+
+```
+ddev pull platform --environment="PLATFORM_PRIMARY_RELATIONSHIP=main"
 ```
 
 ## Usage
--- a/cmd/ddev/cmd/pull.go
+++ b/cmd/ddev/cmd/pull.go
@@ -67,7 +67,7 @@ func appPull(providerType string, app *ddevapp.DdevApp, skipConfirmation bool, s
 		for _, v := range envVars {
 			split := strings.Split(v, "=")
 			if len(split) != 2 {
-				util.Failed("unable to parse commanad-line environment variable setting: '%v'", v)
+				util.Failed("unable to parse command-line environment variable setting: '%v'", v)
 			}
 			provider.EnvironmentVariables[split[0]] = split[1]
 		}
--- a/cmd/ddev/cmd/pull.go
+++ b/cmd/ddev/cmd/pull.go
@@ -62,13 +62,15 @@ func appPull(providerType string, app *ddevapp.DdevApp, skipConfirmation bool, s
 	}
 
 	// Add or override the command-line provided environment variables
-	envVars := strings.Split(env, ",")
-	for _, v := range envVars {
-		split := strings.Split(v, "=")
-		if len(split) != 2 {
-			util.Failed("unable to parse environment variable setting: %v", v)
+	if env != "" {
+		envVars := strings.Split(env, ",")
+		for _, v := range envVars {
+			split := strings.Split(v, "=")
+			if len(split) != 2 {
+				util.Failed("unable to parse commanad-line environment variable setting: '%v'", v)
+			}
+			provider.EnvironmentVariables[split[0]] = split[1]
 		}
-		provider.EnvironmentVariables[split[0]] = split[1]
 	}
 
 	if err := app.Pull(provider, skipDbArg, skipFilesArg, skipImportArg); err != nil {
--- a/cmd/ddev/cmd/push.go
+++ b/cmd/ddev/cmd/push.go
@@ -60,14 +60,16 @@ func apppush(providerType string, app *ddevapp.DdevApp, skipConfirmation bool, s
 		util.Failed("Failed to get provider: %v", err)
 	}
 
-	// Add or override the command-line provided environment variables
-	envVars := strings.Split(env, ",")
-	for _, v := range envVars {
-		split := strings.Split(v, "=")
-		if len(split) != 2 {
-			util.Failed("unable to parse environment variable setting: %v", v)
+	if env != "" {
+		// Add or override the command-line provided environment variables
+		envVars := strings.Split(env, ",")
+		for _, v := range envVars {
+			split := strings.Split(v, "=")
+			if len(split) != 2 {
+				util.Failed("unable to parse environment variable setting: %v", v)
+			}
+			provider.EnvironmentVariables[split[0]] = split[1]
 		}
-		provider.EnvironmentVariables[split[0]] = split[1]
 	}
 
 	if err := app.Push(provider, skipDbArg, skipFilesArg); err != nil {
--- a/pkg/ddevapp/dotddev_assets/providers/platform.yaml
+++ b/pkg/ddevapp/dotddev_assets/providers/platform.yaml
@@ -23,10 +23,12 @@
 # 5. Run `ddev pull platform`. After you agree to the prompt, the current upstream database and files will be downloaded.
 # 6. Optionally use `ddev push platform` to push local files and database to platform.sh. Note that `ddev push` is a command that can potentially damage your production site, so this is not recommended.
 
-# If you have more than one database on platform, you'll currently need to choose which one you want to use
-# as the primary database, and that will be the one pulled or pushed.
-# Do this by setting PLATFORM_PRIMARY_RELATIONSHIP, for example, `ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=maindb`
-
+# If you have more than one database on your Platform.sh proect,
+# you will likely to choose which one you want to use
+# as the primary database ('db').
+# Do this by setting PLATFORM_PRIMARY_RELATIONSHIP, for example, `ddev config --web-environment-add=PLATFORM_PRIMARY_RELATIONSHIP=main`
+# or run `ddev pull platform` with the environment variable, for example
+# `ddev pull platform -y --environment=PLATFORM_PRIMARY_RELATIONSHIP=main`
 # If you need to change this `platform.yaml` recipe, you can change it to suit your needs, but remember to remove the "ddev-generated" line from the top.
 
 # Debugging: Use `ddev exec platform` to see what platform.sh knows about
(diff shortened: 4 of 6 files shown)


rfay:
Thanks @bserem - I think I may try to just get it to pull all databases.

rfay:
@bserem @gilzow @lolautruche this should be ready for your review. Much more mature regarding databases and `ddev pull platform` now. If possible, please try it out sooner rather than later so it can make it into v1.21.4, which is in the next couple of days.

gilzow:
I don't know if these are "issues" or not and whether they are ddev issues or add-on issues, but here's what I observed using this version of ddev when adding in the platformsh-ddev addon
* After answering the question regarding my platform project id, ddev "hung" for more than 90 seconds, but less than 120. I was just about to Ctrl+C the process when it finally continued.
* After the above, I was warned about reconfiguring an existing project and that i could now run `ddev start` BUT ddev wasn't actually finished and then prompted me for an environment name
* I was prompted to enter my environment name. after doing so, I was again given the same warning about reconfiguring an existing application and then told I could do a restart but the same as previous, ddev wasn't actually finished.
* DDEV finished the add-on setup and I ran `ddev start`. During start, about half of my composer dependencies failed to install. Example: 
```
  - Installing wpackagist-plugin/guest-author (2.2): Extracting archive
    Install of wpackagist-plugin/guest-author failed
```
After `ddev start` completed, I ssh'ed into the web app container and ran `composer install` manually. Everything installed correctly. 
* Exited ssh, and then ran `ddev pull platform`. I only have a single database. DB pull worked just fine, but DDEV errored out during the files pull portion with 
```
  [ProcessFailedException]
  The command failed with the exit code: 255
  ```

**Comment by rfay** [comment]
Thanks for testing @gilzow !

* A new ddev version has to grab new images, and unless you have DDEV_DEBUG=true that might not be visible to you.
* DDEV doesn't know about environment names, so can't prompt for them, so that is likely something from the add-on I would imagine. But when you give it an environment name, it does a `ddev config` to update it, so that's why you saw the info about updating the config a second time. That will be smoothed out in the future. 
* About `Install of wpackagist-plugin/guest-author failed` - You're using macOS right? But are you using Docker Desktop/Colima/etc? If colima, you really have to have mutagen enabled for `ddev composer` to work right. In that case I recommend using it globally with `ddev config global --mutagen-enabled`
* `but DDEV errored out during the files pull portion with`... The add-on is not compatible with multiple relationships at this point. I should have mentioned that. 


Could you please test without the add-on, since the add-on isn't compatible with multiple db relationships? You can just clone your project into a different directory, and use `ddev pull platform` on it.

**Comment by bserem** [comment]
Thanks @rfay! Here's my test-run and some thoughts.

## Platform.sh Config
For reference, this how DBs are under PSH. This is used for both tests below.
```
relationships:
  database: 'mysqldb:mysql'
  redis: 'rediscache:redis'
  legacy: 'legacydb:mysql'
```

## Test run 1
1. Downloaded the artifact from the comment above. It was different from the one I got 5 days ago, so I suppose it's a new one with all the latest changes.
2. Fired up a project with multiple relationships
3. The project already has `PLATFORM_PRIMARY_RELATIONSHIP=database` into `config.yaml`
4. `ddev pull platform --skip-files` pulls and imports 2 databases, named `db` and `main`
- - `main` doesn't really mean anything to me at this point, so it is hard to understand what's going on while the script runs.
- - It would be great to allow configuration to download+import only one of the databases.
- - Post DB import hooks run twice, but since these are drush commands in my case they run against the same DB twice.

Output:
```
Obtaining databases...
Creating gzipped SQL dump file: /var/www/html/.ddev/.downloads/main.sql.gz
PLATFORM_PRIMARY_RELATIONSHIP=database so using it as database 'db' instead of the upstream 'main'
Creating gzipped SQL dump file: /var/www/html/.ddev/.downloads/db.sql.gz
Downloaded db dumps for databases 'main main'
Importing databases...
```

## Test run 2
5. Try again on a project where `PLATFORM_PRIMARY_RELATIONSHIP` is **not** configured.
6. `ddev pull platform --skip-files` pulls and imports only one database, called `main`
- - `psh:main` gets imported into `ddev:main`
- - `ddev:db` database remains empty
7. Drupal project can't start without `db` database.

Output
```
Obtaining databases...
Creating gzipped SQL dump file: /var/www/html/.ddev/.downloads/main.sql.gz
File exists: /var/www/html/.ddev/.downloads/main.sql.gz. Overwrite? [Y/n] y
Creating gzipped SQL dump file: /var/www/html/.ddev/.downloads/main.sql.gz
Downloaded db dumps for databases 'main main'
Importing databases...
```

## Thoughts

1. It would be nice to display the names as in `.platform.app.yaml`, so developers know what's going when the script runs
2. 1. If no primary_relationship is configured things won't work properly. It would make sense to get the relationship called `database` for the primary, and only that one. _Currently getting only the other one._
3. An option to only download and import one database would make sense, since both of them might not be required for a handful of scenarios (frontenders don't need migration data for example etc)
4. Overall I feel it's getting there, but it's a bit rough on the edges. Verbose output of the script makes you wonder what's going on and post import hooks run twice.
5. Definitely on a good path here! It would make things easier for multi-db projects. Thanks once again!

I hope I expressed my thoughts clearly :)

**Comment by rfay** [comment]
@bserem there isn't any way that DDEV will be able to support multiple databases of different types. 

I used this template, https://github.com/platformsh-templates/drupal9-multisite

* https://github.com/platformsh-templates/drupal9-multisite/blob/7b96ab12b80cb788ea6399f130e65b22f60b31ce/.platform.app.yaml#L29-L33
* https://github.com/platformsh-templates/drupal9-multisite/blob/7b96ab12b80cb788ea6399f130e65b22f60b31ce/.platform/services.yaml#L6-L26

**Comment by rfay** [comment]
I'd be interested in votes on whether this should go in as is. I'm pretty sure the only risk is to platform.sh users doing multiple databases, which they can't do right now, and they or we can improve the platform.yaml as we go. However, it could wait and not go into this release, although there's a bugfix in there that does need to go in.

**Comment by rfay** [comment]
Some responses to your questions @bserem 

1. "main doesn't really mean anything to me at this point. Where does the word main come from? It is hard to understand what's going on while the script runs." - `main` is the name of one of your relationships; you can `platform relationships` to see the relationships.
2. "It would be great to allow configuration to download+import only one of the databases" - I don't disagree, but since this is already an edge case for most people, I think it can wait and see if people need that.
3. "Post DB import hooks run twice, but since these are drush commands in my case they run against the same DB twice" - post-import-db hooks run after a db import, and in your case there are two DB imports. You could improve your post-import-db hook to check which database it's importing and skip the drush install for one that doesn't matter though.
4. Test Run 2: Without PLATFORM_PRIMARY_RELATIONSHIP, you would need to use $PLATFORM_RELATIONSHIPS to determine the proper settings. But your settings.ddev.php won't know what those are. They can be configured a number of ways. PLATFORM_PRIMARY_RELATIONSHIP short-circuits that and puts the primary db into the 'db' database. But in some cases you'd want the databases to be named the same in DDEV as they are on Platform.sh. That's what you were actually doing in this test run.

Also note that platformsh/ddev-platformsh cannot currently work against multiple-database projects at all, unrelated to this PR. 

Thanks so much for the great and detailed feedback.

**Comment by rfay** [comment]
@gilzow I just finally "processed" your statement "Exited ssh, and then ran ddev pull platform. I only have a single database. DB pull worked just fine, but DDEV errored out during the files pull portion with `The command failed with the exit code: 255`"

Would you be able to share that project with me?

**Comment by bserem** [comment]
Good news on this becoming part of DDEV. It'll help a lot :)

Now, regarding `main`, I tested what you mention above and I can't find a `main` relationship:
image

My relationships are:
- `legacydb`
- `database`
Same as I provided in https://github.com/drud/ddev/pull/4426#issuecomment-1343965761

That's why `main` sounds strange, it doesn't seen to come from my config.

**Comment by rfay** [comment]
`main` is the name of the database/schema, given in `path`. Sorry I misled you.

File: docs/content/users/providers/platform.md [hunk 1]
```diff
@@ -42,6 +42,20 @@ web_environment:
 4. Run `ddev pull platform`. After you agree to the prompt, the current upstream database and files will be downloaded.
 5. Optionally use `ddev push platform` to push local files and database to Platform.sh. The [`ddev push`](../basics/commands.md#push) command can potentially damage your production site, so we don’t recommend using it.
 
+If you have more than one database on your Platform.sh project, you'll need to choose which one you want to use
```
mattstein, line 45: @rfay I understand only one database will be handled by the integration, but what would you recommend if I had additional databases I also wanted to pull into DDEV?

Create my own modified version of `platform.yaml` and modify [`db_pull_command`](https://github.com/drud/ddev/blob/master/pkg/ddevapp/dotddev_assets/providers/platform.yaml#L44)?

rfay: In this version, *all* databases will be handled by the `ddev pull platform`, so line 45 is just incorrect. Thanks for catching that.

mattstein: Am I understanding correctly that all databases will be pulled, but that I still need to designate a `PLATFORM_PRIMARY_RELATIONSHIP` so DDEV knows what to use for its (default) `'db'` database?

mattstein: Ultimately asking while I attempt to improve the wording [over here](https://github.com/mattstein/ddev/commit/317f0304cea9c1a4b93ea60e2e7a598611ee22ad).

bserem: @mattstein this is correct.

- All databases will be downloaded
- The remote database that is flagged as primary will be imported into the `db` database of DDEV (the default one)
- All other remote databases will be imported into ddev databases with the same as in the remote

I believe I've never used the word database that much in such a short message

mattstein: @bserem Thanks for clarifying the database behavior for this database-related documentation adjustment regarding databases. 😄

rfay: Not all projects will need for a "db" database to be populated. For example, as the ddev-platformsh add-on matures, it should be able to handle databases of any number that have the same names they do on Platform.sh, and the "db" database not be used at all.

mattstein: So this is more of an *if* you have multiple databases *and* want to designate a primary to use with `'db'`, yes? If so I’ll update https://github.com/drud/ddev/pull/4469 accordingly.

rfay: That is correct. Most people who have ever used DDEV will want that. There may be contexts in the future where that's not needed.  The same may be true for Acquia or Pantheon or other hosts, where the settings are wired to particular database names and people are *not* using DDEV's automatically generated settings, like `settings.ddev.php` or Craft CMS's .env configuration.

mattstein: Revised [in this commit](https://github.com/drud/ddev/pull/4469/commits/6a311d52db800af254c2c9f975edf7e30e362c7f), which is hopefully more accurate! 😅

PR #4432, opened 2022-12-06
Drop the manual download instructions for Craft CMS

brandonkelly:
## The Problem/Issue/Bug:

The CMS quickstart guide for Craft CMS includes instructions for downloading Craft from a zip file. The zip installation is not going to be supported for much longer though, so these need to be removed.

Manual download instructions have been removed.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4432"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -463,8 +463,6 @@ DDEV comes ready to work with any PHP project, and has deeper support for severa
     
     === "New projects"
     
-        ### Composer project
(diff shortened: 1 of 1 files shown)


rfay:
Why this and also
* https://github.com/drud/ddev/pull/4428 

???

brandonkelly:
Doh, I missed that August submitted a PR for this as well 🤦🏻  Will close…

PR #4197, opened 2022-09-09
Add .env file parsing, fix laravel to use it, fix Shopware 6 overwriting, fixes #3636, fixes #3590

rfay:
## The Problem/Issue/Bug:

We need to read, write, and support .env files better for various project types.

* #3636
* #3590

## How this PR Solves The Problem:

* Add reading and writing of .env
* For Laravel, directly edit the .env file
* For Shopware 6, mostly stop editing it except MAILER_URL

## Manual Testing Instructions:

- [x] Do manual installation of Laravel and then use it
- [x] Do manual installation of Shopware 6 and then use it
- [x] Verify that only appropriate changes are made to the .env

## Automated Testing Overview:

No changes, existing should catch it.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4197"></a>

Issue #3636, opened 2022-02-28
Improve Laravel's explicit integration by actually updating the .env file


tyler36:
**Is your feature request related to a problem? Please describe.**

DDEV does a good job of setting up itself and pre-configuring a Drupal project.
It would be nice if we could automate other frameworks too. 

**Describe the solution you'd like**
After running `ddev config` in a new Laravel project, there are 2 additional steps required.
- Updating PHP version. Addressed in #3588 
- Updaitng Laravel `.env` file

```diff
# These are the current defaults
APP_URL=http://localhost
DB_CONNECTION=mysql
DB_HOST=127.0.0.1
DB_PORT=3306
DB_DATABASE=laravel
DB_USERNAME=root
DB_PASSWORD=
```

Eg. Assuming I create  a new project call phpcs-demo 

## Test

- Start a new `Laravel` project

```shell
laravel new phpcs-demo
cd phpcs-demo
```

- Add `DDEV` to project, accepting all default

```shell
$ ddev config
...
Configuration complete. You may now run 'ddev start'.
```

- Bump DDEV `PHP` version in `./.ddev/config.yaml`

```yaml
php_version: "8.0"
```

- Config Laravel `.env`

```env
APP_URL=https://phpcs-demo.ddev.site
DB_CONNECTION=mysql
DB_HOST=ddev-phpcs-demo-db
DB_DATABASE=db
DB_USERNAME=db
DB_PASSWORD=db
```

This assumes a `mysql` database, but I _think_ all of these might be ddev variables already?
It would get great if a laravel-project type ddev config could automatically update the `.env` file too.

**Additional context**
Ideally, we would want to update other supported frameworks (WP, Magento) but I have no idea how involved those are. A Laravel feels like an easy win to start though.

rfay:
Thanks for this!
I imagine you're queueing up a PR for this, but reading this issue it's not clear to me what you want to do.

tyler36:
No PR on the way, more of a `TODO` placeholder

After running `ddev config` , if  the project type = 'laravel', ddev should update the `.env` file with its setting.

```diff
- APP_URL=http://localhost
+ APP_URL=<DDEV_PRIMARY_URL>
- DB_CONNECTION=mysql
+ DB_CONNECTION=<DDEV_DB_TYPE>
- DB_HOST=127.0.0.1
+ DB_HOST=<DDEV_DB_HOST>
- DB_PORT=3306
+ DB_PORT=<DDEV_DB_PORT>
[...]

Issue #3590, opened 2022-02-09
Project type shopware6 keeps overwriting APP_URL


susannemoog:
**Run a Diagnostic and Paste Link Here**
https://gist.github.com/susannemoog/2b36c2feb9746392606c6eedf4353d0c

**Describe the bug**
With a project of type shopware6 manual changes to the APP_URL are not possible, as every restart adds an additional line with the original ddev created value.

**To Reproduce**
```
name: shopware
type: shopware6
docroot: public
php_version: "7.4"
webserver_type: nginx-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
additional_fqdns: []
mariadb_version: "10.3"
mysql_version: ""
nfs_mount_enabled: false
mutagen_enabled: false
use_dns_when_possible: true
composer_version: "2"
web_environment: []
```

- Add APP_URL manually to .env
- (Re)Start ddev 
- Open .env
- Adjust APP_URL ddev created to http
- Restart ddev

Results in something like this in the .env:
```
APP_ENV="dev"
APP_URL="http://shopware.ddev.site"
...
# =================
# Configuration added by ddev
APP_URL="http://shopware.ddev.site"
# =================
# Configuration added by ddev
APP_URL="http://shopware.ddev.site"
# =================
# Configuration added by ddev
APP_URL="http://shopware.ddev.site"
# =================
# Configuration added by ddev
APP_URL="https://shopware.ddev.site"
```

**Expected behavior**

- If APP_URL is already set, do not add another one

**Version and configuration information (please complete the following information):**

config.yaml - see above
version info - see gist above
Shopware installation according to https://susi.dev/ddev-shopware-6/

rfay:
Thanks, will try to get it fixed up.

PR #4437, opened 2022-12-07
Fix over-eager project env replacement, for #4197

rfay:
## The Problem/Issue/Bug:

See bug reported in 
* https://github.com/drud/ddev/pull/4197#issuecomment-1340843620 Thanks @stasadev

The application .env file replacement (for laravel, shopware, etc.) was matching partials on the key, so 

> For example, I have DYNAMODB_CONNECTION in my .env (from [this](https://github.com/baopham/laravel-dynamodb/blob/963811691313124e7562a02aefdcc31fdbd75e78/config/dynamodb.php#L14) package) and ddev thinks it is the same as DB_CONNECTION, and replaced my config with DYNAMODB_CONNECTION="mysql".

* Use a more sophisticated pattern for replacement
* Add TestWriteProjectEnvFile that covers (hopefully) all the things

Try the env file suggested by @stasadev 

Added significant test in TestWriteProjectEnvFile

## Related Issue Link(s):

* https://github.com/drud/ddev/pull/4197#issuecomment-1340843620 
* #3636 
* #3590

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -98,7 +98,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		return nil
 	}
 	// Read in the .env file
-	envMap, envText, err := ReadEnvFile(app)
+	envMap, envText, err := ReadProjectEnvFile(app)
 	if err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("Unable to read .env file: %v", err)
 	}
@@ -138,7 +138,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		}
 	}
 
-	err = WriteEnvFile(app, envMap, envText)
+	err = WriteProjectEnvFile(app, envMap, envText)
 	if err != nil {
 		return err
 	}
--- a/pkg/ddevapp/envfile.go
+++ b/pkg/ddevapp/envfile.go
@@ -9,9 +9,9 @@ import (
 	"strings"
 )
 
-// ReadEnvFile reads the .env in the project root into a envText and envMap
+// ReadProjectEnvFile reads the .env in the project root into a envText and envMap
 // The map has the envFile content, but without comments
-func ReadEnvFile(app *DdevApp) (envMap map[string]string, envText string, err error) {
+func ReadProjectEnvFile(app *DdevApp) (envMap map[string]string, envText string, err error) {
 	envFilePath := filepath.Join(app.AppRoot, ".env")
 	envText, _ = fileutil.ReadFileIntoString(envFilePath)
 	envMap, err = godotenv.Read(envFilePath)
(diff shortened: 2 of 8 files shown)


stasadev (approved):
It works, thank you!

Also, I noticed that:

1. It writes the db config to `.env` even if you have `omit_containers: [db]`
2. I use URL from `additional_hostnames` for `APP_URL` in Laravel, but it is always overridden with the primary URL (maybe you can use curl to check if `APP_URL` is already reachable, and do not override it).

rfay:
Thanks so much @stasadev - 

1. Do you think it matters that it writes DB stuff if omit_containers[db]?
2. I recommend using `ddev config --disable-settings-management` in your situation. Is that adequate for you?

stasadev:
1. Just noticed it, does not matter to me.
2. Yes, that's exactly what I want 👍

rahaddon:
@rfay Hey, every time I perform a `ddev restart` on a laravel project which doesn't have standard ENV vars like DB_USER they are always being either automatically added or overwritten which results in having to undo the changes, is this the desired behaviour? Thanks

rfay:
That is the expected behavior, but you can just turn it off with `disable_settings_management: true`, see https://ddev.readthedocs.io/en/latest/users/usage/cms-settings/

But I'd love to have you explain in more detail why your laravel project doesn't use those so I could understand better.

rahaddon:
Thank you for getting back to me, does `disable_settings_management: true`  disable anything else or it is just related to auto config overwrites in the .env file based on the project type?

> But I'd love to have you explain in more detail why your laravel project doesn't use those so I could understand better.

We are using a remote staging database environment which is shared between development team members so the credentials are different to those defaults which are auto-populated by DDEV. Each team member has their own unique user credentials to access the database.

rfay:
I think `disable_settings_management: true` will do what you want. 

I do recommend using `ddev pull` to get the database rather than this, as it will prevent unexpected results of using a shared db. Or just sharing a db dump. But I know you've already thought this through.

faytecCD:
@rfay 

Hi, The following configuration does not work for me. DDEV still rewrites `DB_*` values in `.env` as well as other keys it thinks it should control (e.g., `APP_URL`):

(Sorry, after reconfirming, the correct name of the new configuration option should be disable_settings_management.)

```diff
type: laravel
docroot: public
php_version: "8.4"
webserver_type: nginx-fpm
xdebug_enabled: false
omit_containers: [db]
+ settings_management_disabled: true
```

The only way to stop DDEV from modifying my `.env` is to switch the project type to PHP:

```diff
- type: laravel
+ type: php
docroot: public
php_version: "8.4"
webserver_type: nginx-fpm
xdebug_enabled: false
omit_containers: [db]
- settings_management_disabled: true
```

---

Why I’m doing this:

~1. I am indeed running a Laravel project that originally used Laravel Sail, so it already has MySQL, Redis, etc. For several reasons, I’m not migrating that data. I simply reuse the existing services (accessible via `host.docker.internal` without any issues).~

(Sorry, I just rechecked and found that when `settings_management_disabled: true` and `omit_containers: [db]` are set, ddev no longer modifies the `DB_*` values in my .env, but it still changes my `APP_URL` and `MAIL_*`.)

2. DDEV makes some unexpected changes.

   For example, DDEV always generates a default domain like `<project>.ddev.site`, which appears impossible to disable. I already have fixed domains and added them via `additional_fqdns`. But every time I run `ddev start`, DDEV rewrites my `APP_URL` to `<project>.ddev.site`, which is disruptive.

3. Unprompted `.env` modifications are destructive.

   In daily development, `.env` is not committed to git. DDEV silently overwriting it on start means changes can go unnoticed until a failure occurs. Without git history, old values may be unrecoverable (thankfully PhpStorm keeps local history).

---

Suggestions:

* Provide a preview before modifying `.env`, and automatically back up the file.
* Provide an option to completely prevent `.env` modification (currently only possible by switching to `type: php`).
* First setup, offer configuration options similar to Sail, allowing developers to enable/disable specific service containers.
* If modifying `.env` is necessary, support prepend/append/modify/never behaviors.
  For Laravel, later `.env` entries override earlier ones, so prepending would avoid breaking existing values. Only when needed should DDEV overwrite or remove existing entries, and such changes should be explicit and manageable.

Thank you.

tyler36:
@faytecCD

The setting was renamed since the 2023 comment you referenced.

 Use the following command to apply the setting

```shell
ddev config --disable-settings-management
```

This adds the following setting to `.ddev/config.yaml`

```conf
```

see [Controlling or Removing CMS Settings](https://docs.ddev.com/en/stable/users/usage/cms-settings/#controlling-or-removing-cms-settings)

faytecCD:
@tyler36 

Thank you for your response. I just discovered the same issue, and after switching to `disable_settings_management`, it looks good. It no longer breaks my `.env`.

rfay:
Sorry I always do have to look and see what it is. 

Config setting: [disable_settings_management](https://docs.ddev.com/en/stable/users/configuration/config/#disable_settings_management)

[ddev config --disable-settings-management](https://docs.ddev.com/en/stable/users/usage/commands/#config)

I'll edit the mistake above.

More detail: [Managing CMS Settings](https://docs.ddev.com/en/stable/users/usage/cms-settings/)

PR #4438, opened 2022-12-07
Fix help message when deleting a project

juampynr:
It references an old option to skip creating a snapshot.

From `ddev delete --remove-data --yes`

To `ddev delete --omit-snapshot --yes`

## The Problem/Issue/Bug:

When you try deleting a project with `ddev delete`, you get the following output:

```
...
your db container in project drupalcms is not running. 
Please start the project if you want to snapshot it. 
If deleting project, you can delete without a snapshot using 
'ddev delete --remove-data --yes', 
which will destroy your database
```

The above is incorrect because `--remove-data` does not exist.

It replaces `--remove-data` by `--omit-snapshot`.

This is a documentation change so there is no need to test anything manually.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4438"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2092,7 +2092,7 @@ func (app *DdevApp) Snapshot(snapshotName string) (string, error) {
 	// Ensure that db container is up.
 	err = app.Wait([]string{"db"})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rfay (approved):
Impressive, thanks for catching that!

PR #4445, opened 2022-12-11
Update php 8.2 extensions (xdebug, memcached, redis now there)

rfay:
## The Problem/Issue/Bug:

deb.sury.org fixed a number of missing php 8.2 extensions, so they should all be here now.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4445"></a>


rfay:
xhprof has a bug with PHP8.1/8.2, see 
* https://github.com/longxinH/xhprof/issues/76

PR #4469, opened 2022-12-19
[docs] Minor tidying for `ddev pull platform` PRs #4441 and #4426

mattstein:
## The Problem/Issue/Bug:

Documentation guy is too slow right now and trying to catch up.

This makes some minor edits for consistency and hopefully clearer phrasing about how databases are handled in the Platform.sh integration.

I brought this up to date with the `master` branch and had to fix a conflict in `docs/content/users/providers/platform.md`, which I think looks good and doesn’t omit any improvements. But you may want to double-check.

[Preview changes locally](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) or inspect the automatic build:

- [Commands → `get`](https://ddev--4469.org.readthedocs.build/en/4469/users/basics/commands/#get)
- [Platforms → Platform.sh → Designating a Primary Database](https://ddev--4469.org.readthedocs.build/en/4469/users/providers/platform/#designating-a-primary-database)

n/a

## Related Issue Link(s):

- #4441 
- #4426 
- #4188

## Release/Deployment notes:

n/a

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4469"></a>

### DIFF
--- a/docs/content/users/basics/commands.md
+++ b/docs/content/users/basics/commands.md
@@ -550,7 +550,7 @@ Flags:
 
 * `--all`: List unofficial *and* official add-ons. (default `true`)
 * `--list`: List official add-ons. (default `true`)
-* `--verbose` or `-v`: Output verbose error information with bash `set -x` (default `false`)
+* `--verbose`, `-v`: Output verbose error information with Bash `set -x` (default `false`)
 
 Example:
 
@@ -566,7 +566,6 @@ ddev get drud/ddev-redis
 
 # Get debug info about `ddev get` failure
 ddev get drud/ddev-redis --verbose
-
 
 # Download the Drupal 9 Solr add-on from its v0.0.5 release tarball
 ddev get https://github.com/drud/ddev-drupal9-solr/archive/refs/tags/v0.0.5.tar.gz
--- a/docs/content/users/providers/platform.md
+++ b/docs/content/users/providers/platform.md
@@ -42,11 +42,13 @@ web_environment:
 4. Run `ddev pull platform`. After you agree to the prompt, the current upstream databases and files will be downloaded.
 5. Optionally use `ddev push platform` to push local files and database to Platform.sh. The [`ddev push`](../basics/commands.md#push) command can potentially damage your production site, so we don’t recommend using it.
 
-### Designating a Primary Database
+### Managing Multiple Databases
 
-All databases for your Platform.sh project will be pulled locally, with each name matching its remote. When you have more than one, you can optionally tell DDEV which to use as the primary and load into the default `'db'` database.
+If your project has only one database, it will automatically be pulled into and pushed from DDEV’s `'db'` database.
 
-You can do this by setting `PLATFORM_PRIMARY_RELATIONSHIP`:
+If your project has multiple databases, they’ll all be pulled into DDEV with their respective remote names. You can optionally designate a *primary* to use DDEV’s default `'db'` database, which may be useful in some cases—particularly if you’ve been using the default solo-database behavior and happened to add another one to your project.
+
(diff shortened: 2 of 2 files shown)


bserem (approved):

docs/content/users/basics/commands.md
```diff
@@ -550,7 +550,7 @@ Flags:
 
 * `--all`: List unofficial *and* official add-ons. (default `true`)
 * `--list`: List official add-ons. (default `true`)
-* `--verbose` or `-v`: Output verbose error information with bash `set -x` (default `false`)
+* `--verbose`, `-v`: Output verbose error information with Bash `set -x` (default `false`)
```
rfay, line 553: I have a hard time capitalizing 'Bash' but can live with it. 

The reality is I have a hard time with 'DDEV' :)

docs/content/users/providers/platform.md
```diff
@@ -42,15 +42,17 @@ web_environment:
 4. Run `ddev pull platform`. After you agree to the prompt, the current upstream databases and files will be downloaded.
 5. Optionally use `ddev push platform` to push local files and database to Platform.sh. The [`ddev push`](../basics/commands.md#push) command can potentially damage your production site, so we don’t recommend using it.
 
-If you have more than one database on your Platform.sh project, you'll need to choose which one you want to use
-as the 'db' primary database on DDEV, and that one will be pulled to the database named `db`.
-Do this by setting PLATFORM_PRIMARY_RELATIONSHIP, for example,
+### Designating a Primary Database
+
+All databases for your Platform.sh project will be pulled locally, with each name matching its remote. When you have more than one, you can optionally tell DDEV which to use as the primary and load into the default `'db'` database.
```
rfay, line 47: Since this won't matter to the vast majority of people, we may want to calm it instead of making it a major header. That's why it was calmed into bullets before; it doesn't matter most of the time, most people have one database. 

```suggestion
If you have just one database on Platform.sh, it will be pulled and loaded into the local default database named 'db'. 

However, if you have more than one database on platform.sh, 
all databases will be pulled locally, with each name matching its remote. If you have more than one, you can optionally tell DDEV which to use as the primary one and load into the default `'db'` database.
```

rfay (changes requested):
Thanks for the multiple efforts on this. It's hard (and was hard to implement :) ). 

But since it only matters to people with more than one db, it probably should be skippable by ordinary folks.

I should have spent more time explaining the 3 cases:

1. One database upstream: Gets downloaded into 'db', nothing required of the user, they don't have to even think. This is the most common case.
2. Multiple databases upstream, PLATFORM_PRIMARY_RELATIONSHIP set: PLATFORM_PRIMARY_RELATIONSHIP's db goes into 'db'. All other databases are pulled and they are named as they were upstream.
3. Multiple databases upstream, PLATFORM_PRIMARY_RELATIONSHIP not set: All databases are downloaded and loaded into databases with the same names they have upstream.

docs/content/users/providers/platform.md
rfay: I'm fine with however you want to do it, knowing what you know. Thanks! Sorry this one is so painful.

mattstein: Not at all, it’s an interesting feature to describe simply! I’ll counter-suggest something again shortly.

mattstein: Alright, another stab: https://github.com/drud/ddev/pull/4469/commits/1c0d63ed0d96446ff4cba4525338a2bb939cb15a

I used my theoretical understanding to try and clarify why you _may_ want to designate a primary database, since it’s optional.

rfay (approved):
Thanks!

Issue #4519, opened 2023-01-04
[Craft CMS] `.env` file is not written to correct location when app is in a sub directory


flynnt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

When a project type of Craft CMS is configured and the Craft application files live in a subdirectory - e.g., `/app` - the project initialization script should write the `.env` file to the location in which the app is installed. In my case, this is also usually the Composer root.

### Actual Behavior

DDEV attempts to write out the `.env` file to the project root (that is, `app.AppRoot`) instead of the subdirectory in which Craft is installed.

### Steps To Reproduce

**Configure the project:**
`ddev config --project-type=craftcms --docroot=app/web --create-docroot --composer-root=app --web-environment-add=CRAFT_CMD_ROOT=./app`

**Install Craft**
`ddev composer create -y --no-scripts --no-install craftcms/craft`
`ddev composer install`

Note that the base `.env.example.dev` file is correctly copied to `.env` within `/app`, but when `WriteProjectEnvFile` is called, it does not actually write to the correct file, so you end up with one `.env` file in the `/app` and one in `app.AppRoot`.

### Anything else?

This looks to be the result of the abstraction related to writing `.env` files that happened in #4197. I hesitate to call it a regression or even a bug, really, but the `.env` file operations previously written in `craftcms.go` were working prior to this. It may just be a matter of parameterizing the path, or adding something like:

`envFilePath := filepath.Join(app.AppRoot, app.ComposerRoot, ".env")`

rfay:
Want to try a PR to fix it? Seems like you've explored it well.. And thanks!

flynnt:
Yep. I'll give it a shot. And thank _you_! :)

rfay:
Some guidance in https://ddev.readthedocs.io/en/latest/developers/building-contributing/ and the rest of that section, thanks for giving it a go!

rfay:
This was fixed in #4526, thanks!

PR #4526, opened 2023-01-05
Updates write path for `.env` files to respect app-specific locations.

flynnt:
## The Issue
There were some abstractions in #4197 that led to the addition of two methods for reading and writing out `.env` files for project types where settings management is enabled: `ReadProjectEnvFile` and `WriteProjectEnvFile` respectively. These methods each presumed that the `.env` file was located at `app.AppRoot`. However, there are cases where the `.env` file might need to live somewhere else, for example, if something like Craft CMS is installed in a subdirectory. In a case like, that, the post start action would keep attempting to write out an `.env` file in the incorrect location.

This PR makes the `envFilePath` a parameter in both `ReadProjectEnvFile` and `WriteProjectEnvFile`, which allows a given application type like Craft CMS or Laravel to be in charge of determining where it thinks its `.env` file should be. This seems like a better approach, since now the reading and writing methods don't have to have an opinion about where the `.env` file lives.

I've been testing this with Craft CMS. The general commands are:

**Configure the project:**
```
ddev config --project-type=craftcms --docroot=app/web --create-docroot --composer-root=app --web-environment-add=CRAFT_CMD_ROOT=./app
```

**Install Craft**
```
ddev composer create -y --no-scripts --no-install craftcms/craft
ddev composer install
```

With these changes, you should see an `.env` file that reads from `app/.env` and writes/appends the appropriate DDEV values to that file.

## Related Issue Link(s)
* #4519
* #4197

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4526"></a>

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -98,7 +98,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		return nil
 	}
 	// Read in the .env file
-	envMap, envText, err := ReadProjectEnvFile(app)
+	envMap, envText, err := ReadProjectEnvFile(envFilePath)
 	if err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("Unable to read .env file: %v", err)
 	}
@@ -138,7 +138,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		}
 	}
 
-	err = WriteProjectEnvFile(app, envMap, envText)
+	err = WriteProjectEnvFile(envFilePath, envMap, envText)
 	if err != nil {
 		return err
 	}
--- a/pkg/ddevapp/envfile.go
+++ b/pkg/ddevapp/envfile.go
@@ -4,15 +4,14 @@ import (
 	"fmt"
 	"github.com/drud/ddev/pkg/fileutil"
 	"github.com/joho/godotenv"
-	"path/filepath"
 	"regexp"
 	"strings"
 )
 
 // ReadProjectEnvFile reads the .env in the project root into a envText and envMap
 // The map has the envFile content, but without comments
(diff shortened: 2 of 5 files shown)


rfay:
Thanks and congratulations, I'll review this in the next few days.

rfay (approved):
Reading this carefully it seems to make no actual changes to projects besides Craft CMS. I kind of wonder if the .env file might be expected in other random places in other CMSs. Will ask @tyler36 to take a look from a Laravel perspective.

flynnt:
I think you're correct. My rationale was just to let other application types figure out on their own how/where they want to discover their individual `.env` files. It may be the case with some of them that they will never have their .env files in any other location than `app.AppRoot`, or it may be the case that individual app types need to be super flexible about where their settings live. My changes just make the read/write methods agnostic to the whims of a given app type. Thanks for reviewing!

rfay:
Manually tested with Craft CMS per the testing instructions and it seemed fine to me.

rfay:
Asked @mattsches to review this from his Laravel perspective in [discord](https://discord.com/channels/664580571770388500/1061945409011986442/1062505856257773724).

tyler36:
Laravel uses the project root for `.env` files so under "typical" installations this isn't required. 

However, I really like the idea of DDEV deferring to the frameworks preferred path for `.env`'s. I think it's brilliant addition!

Should we add a small comment in the docs for how to use it? Perhaps under CraftCMS quickstart and/or Environment Variables Provided?

tyler36 (approved):

rfay:
@tyler36 I think there's not really anything to know here with Craft CMS, and actual code would have to be added to do it elsewhere. With Craft, it's `CRAFT_CMD_ROOT` that determines what's to happen, and DDEV is just following along with this PR.

PR #4532, opened 2023-01-08
[docs] Remove notes about composer limitations

tbal:
Since [v1.21.4](https://github.com/drud/ddev/releases/tag/v1.21.4) command flags for wrapper commands work as expected, `ddev composer --help` works out of the box.

## The Issue

The [docs](https://ddev.readthedocs.io/en/latest/users/usage/developer-tools/#limitations-of-ddev-composer) mention limitions when using composer through ddev. This is not true since latest release v1.21.4.

It removes the composer limitations part in the docs.

## Release/Deployment Notes

Maybe it is too early to remove this notes for users with older ddev version stumpling upon this limitation. I don't know how you handle this..

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4532"></a>


rfay (approved):
Thanks for catching that!

PR #4549, opened 2023-01-12
Provide container start hooks and ability to RUN dockerfile with code mounted, fixes #4233

### LINKED ISSUE #4233: Allow hooks or other interventions earlier in the start and build process
After working on 
* #4215 something else came to mind, which it's probably very custom but I wonder what your thoughts are.

`start.sh` does a lot of cool things, cool things that I may want to change or do different, or do things in between.

On a project I have `pm2` that serves a node app. I added `pm2` as an extra daemon. And `pm2` is being installed globally from a `Dockerfile`.

The first time the project is spun, it fails because there's no dependencies installed. I added `yarn install` to `post-start` hook but that happens after `start.sh` which starts `supervisord` which starts `pm2`. While that's fine, I now learned I can add my own start.sh that `yarn install` before the supervisor is run.

I am ok doing this but I wonder if we can add flexibility here. It may not be worth it, but taking over start.sh will force me to have it updated on newer versions if more things were added. I could probably do some `sed` that might be more update aware.

One thing I thought of is having something like a `/start.d/` directory where you can add things like `90-supervisor.sh`, `50-node-cache.sh` and something like this where the current bits of what start.sh does can go into sub files and `start.sh` executes those in order. In this scenario I might decide to take over a single snippet of `start.sh` or add something I want to do somewhere else by adding a `51-user.sh` to `/start.d`.

This could even be something an addon can add to it. 

Just an idea.

rfay:
## The Issue

* #4233

* Need the ability to execute actions before supervisord starts up (container start time)
* Need the ability to execute scripts in the context of the project code if necessary (build time)

Does both of those. 

* scripts in .ddev/web-entrypoint.d will be executed on start right before supervisord is run
* You can do build actions in the context of the project with a .ddev/web-build/Dockerfile.* like
```
RUN --mount=type=bind,source=.,target=/var/www/html ls /var/www/html >/var/tmp/htmlcontents
```

## TODO

- [x] Needs tests for each type of additional thing
- [x] Needs docs on how to do those
- [x] Consider adding the start hook to db container.

The current directory is already the context, but that RUN would also mount the current directory as /var/www/html to be like container start time.

* Try out adding scripts to `.ddev/web-entrypoint.d`
* Try out adding a Dockerfile that uses mounted code to do something interesting.
* Experiment with complex platform.sh setups as described in https://github.com/drud/ddev-platformsh/issues/89

## Automated Testing Overview

- [x] Needs tests for both of these

## Release/Deployment Notes

- [ ] Dockerfiles that include outside files may have to have the path changed

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4549"></a>


rfay:
Wow, I think this can finally be tested against something more real than what I've been doing.

rfay:
Manually tested with the ddevdotcom WP repository, the one that sent us down this path. In .ddev/web-entrypoint.d I put two scripts:

a-envvars.sh (the same stuff in the config.platformsh.yaml)
```bash
PLATFORM_RELATIONSHIPS=ewoiZGF0YWJhc2UiOiBbCiAgewogICAgInVzZXJuYW1lIjogImRiIiwKICAgICJzY2hlbWUiOiAibXlzcWwiLAogICAgInNlcnZpY2UiOiAiZGF0YWJhc2UiLAogICAgImZyYWdtZW50IjogbnVsbCwKICAgICJpcCI6ICIyNTUuMjU1LjI1NS4yNTUiLAogICAgImhvc3RuYW1lIjogImRiIiwKICAgICJwdWJsaWMiOiBmYWxzZSwKICAgICJjbHVzdGVyIjogImRkZXYtZHVtbXktY2x1c3RlciIsCiAgICAiaG9zdCI6ICJkYiIsCiAgICAicmVsIjogIm15c3FsIiwKICAgICJxdWVyeSI6IHsKICAgICAgImlzX21hc3RlciI6IHRydWUKICAgIH0sCiAgICAicGF0aCI6ICJkYiIsCiAgICAicGFzc3dvcmQiOiAiZGIiLAogICAgInR5cGUiOiAibWFyaWFkYjoxMC40IiwKICAgICJwb3J0IjogMzMwNiwKICAgICJob3N0X21hcHBlZCI6IGZhbHNlCiAgfQpdIH0K
PLATFORM_MOUNTS="web/app/uploads web/app/wp-content/cache"
PLATFORM_APP_DIR=/var/www/html
PLATFORM_PROJECT_ENTROPY=f502ed008b747df6bd0a5a6f21e9afaca21be807cf1f1c68970b7ffbf6752fb9
# Consider commit hash for PLATFORM_TREE_ID
PLATFORM_TREE_ID=2dc356f2fea13ef683f9adc5fc5bd28e05ad992a
PLATFORM_CACHE_DIR=/mnt/ddev-global-cache/ddev-platformsh/ddevdotcom
PLATFORM_DIR=/var/www/html
PLATFORM_ROUTES=eyAiaHR0cHM6Ly9kZGV2ZG90Y29tLmRkZXYuc2l0ZS8iOiB7ICJwcmltYXJ5IjogdHJ1ZSwgImlkIjogbnVsbCwgInByb2R1Y3Rpb25fdXJsIjogImh0dHBzOi8vZGRldmRvdGNvbS5kZGV2LnNpdGUvIiwgImF0dHJpYnV0ZXMiOiB7fSwgInVwc3RyZWFtIjogImRydXBhbCIsICJ0eXBlIjogInVwc3RyZWFtIiwgIm9yaWdpbmFsX3VybCI6ICJodHRwczovL3tkZWZhdWx0fS8iIH0gfQ==
PLATFORM_VARIABLES=e30=
```

And `b-environment.sh`, a copy of the platform .environment:

```
export DB_NAME=$(echo $PLATFORM_RELATIONSHIPS | base64 --decode | jq -r ".database[0].path")
export DB_HOST=$(echo $PLATFORM_RELATIONSHIPS | base64 --decode | jq -r ".database[0].host")
export DB_PORT=$(echo $PLATFORM_RELATIONSHIPS | base64 --decode | jq -r ".database[0].port")
export DB_USER=$(echo $PLATFORM_RELATIONSHIPS | base64 --decode | jq -r ".database[0].username")
export DB_PASSWORD=$(echo $PLATFORM_RELATIONSHIPS | base64 --decode | jq -r ".database[0].password")

export WP_HOME=$(echo $PLATFORM_ROUTES | base64 --decode | jq -r 'to_entries[] | select(.value.primary == true) | .key')
export WP_SITEURL="${WP_HOME}wp"
export WP_DEBUG_LOG=/var/log/app.log
export WP_ENV="${PLATFORM_ENVIRONMENT_TYPE}"
export AUTH_KEY=$PLATFORM_PROJECT_ENTROPY
export SECURE_AUTH_KEY=$PLATFORM_PROJECT_ENTROPY
export LOGGED_IN_KEY=$PLATFORM_PROJECT_ENTROPY
export NONCE_KEY=$PLATFORM_PROJECT_ENTROPY
export AUTH_SALT=$PLATFORM_PROJECT_ENTROPY
export SECURE_AUTH_SALT=$PLATFORM_PROJECT_ENTROPY
export LOGGED_IN_SALT=$PLATFORM_PROJECT_ENTROPY
export NONCE_SALT=$PLATFORM_PROJECT_ENTROPY
```

And it worked, just like that!

rfay:
@mattstein so sorry I didn't ask for your review before now. But it passes tests after 47 commits!

docs/content/users/extend/custom-commands.md
```diff
@@ -73,6 +73,7 @@ A number of environment variables are provided to these command scripts. These a
 * `DDEV_HOST_HTTPS_PORT`: localhost port for HTTPS on web server
 * `DDEV_HOST_MAILHOG_PORT`: localhost port for MailHog
 * `DDEV_HOST_WEBSERVER_PORT`: localhost port of the web server
+* `DDEV_MUTAGEN_ENABLED`: "true" if mutagen is enabled
```
mattstein, line 76:
```suggestion
* `DDEV_MUTAGEN_ENABLED`: `true` if Mutagen is enabled
```

docs/content/users/extend/custom-commands.md
```diff
@@ -90,6 +91,7 @@ Useful variables for container scripts are:
 * `DDEV_DOCROOT`: relative path from approot to docroot
 * `DDEV_FILES_DIR`: directory of user-uploaded files
 * `DDEV_HOSTNAME`: comma-separated list of FQDN hostnames
+* `DDEV_MUTAGEN_ENABLED`: "true" if mutagen is enabled
```
mattstein, line 94:
```suggestion
* `DDEV_MUTAGEN_ENABLED`: `true` if Mutagen is enabled
```

docs/content/users/extend/customizing-images.md
```diff
@@ -78,7 +78,7 @@ RUN chmod -R ugo+rw $COMPOSER_HOME
 ENV COMPOSER_HOME=""
 ```
 
-**Remember that the Dockerfile is building a Docker image that will be used later with DDEV.** At the time the Dockerfile is executing, your code is not mounted and the container is not running, it’s just being built. So for example, an `npm install` in `/var/www/html` will not do anything useful because the code is not there at image building time.
+**Remember that the Dockerfile is normally building a Docker image that will be used later with DDEV.** At the time the Dockerfile is executing, your code by default is not mounted and the container is not running, it’s just being built. So for example, an `npm install` in `/var/www/html` will not do anything useful because the code is not there at image building time. However, you use `RUN` in the context of your codebase using something like `RUN --mount=type=bind,source=.,target=/var/www/html cp /var/www/html/index.php /var/tmp/` would mount your code and make it available at the normal `/var/www/html` mount. However, you can't make changes to the mounted code, so things like an early `npm install` wouldn't work, although `npm install -g` would work.
```
mattstein, line 81: I’m having trouble following this.

It starts with one of those “remember” phrases that’s assuming I already know what you’re allegedly reminding me. If I’m being thick and the order of operations were already described somewhere, it’d be ideal to link to it. If this should have been obvious to the Docker-competent, fine.

Then a couple sentences that made sense are followed by two new “However,” clauses—and I’m not sure if the second however is countering the first or thickening the plot.

I ultimately don’t know what this is saying so I’m not sure what to recommend, other than writing it to explain like I’m five.

docs/content/users/usage/architecture.md
```diff
@@ -67,6 +67,9 @@ A project’s `.ddev` directory can be intimidating at first, so let’s take a
 `web-build` directory
 : Can be used to provide a [custom Dockerfile](../extend/customizing-images.md) for the web container.
 
+`web-entrypoint.d` directory
+: Custom scripts (named `*.sh`) in this directory will be run during web container startup, before the php-fpm server or other daemons are run. This is an advanced topic, so don't put anything here unless you know why you're doing it. It's useful for introducing environment variables into the context of the nginx and php-fpm servers, for example.
```
mattstein, line 71: Restructure to lead with the example, then explain _why_ caution is warranted. “This is an advanced topic” subtly sounds like a status judgment (“this is only the gifted students!”) rather than plainly explaining why someone should be careful.

My reason could be wrong, so that could need to be revised for accuracy. 😅 

```suggestion
: Custom scripts (named `*.sh`) in this directory will be run during web container startup, before the php-fpm server or other daemons are run. This can be useful, for example, for introducing environment variables into the context of the nginx and php-fpm servers. Use this carefully, because custom entrypoints can very easily break things.
```

mattstein (changes requested):
A mere 47 commits! 💪 🎉 👏 

I have left my customary nitpicks.

docs/content/users/extend/customizing-images.md
mattstein: And apologies in advance to any five-year-old Docker developers that would have read this just fine.

docs/content/users/extend/custom-commands.md
rfay: The oddity about this one, and the reason I had left it with quotes, is the difference between the text string "true" and the binary `true`. However, there isn't a difference in the context of an environment variable, so going with your suggestion. It makes enormous difference in yaml.

mattstein: In that case it could just as well be `"true"` to express that I think. I had looked at `IS_DDEV_PROJECT` on this page and assumed this should follow suit.

docs/content/users/extend/customizing-images.md
rfay: It's all that problem of how much people know or don't know, and how little of that you can fix in this context.

The "normal" situation is that in a Dockerfile you're building an image, like building an image of a machine. This may happen months or years before the code of the project is even imagined. In the case of DDEV, most of each image is built during DDEV development. And then when DDEV is used by developers, their code is mounted into that container that was built so long ago.

In the case of additional .ddev/web-build/Dockerfile.*, the time may be closer to their code's lifecycle, but not necessarily. For example, they may be a `ddev start` for the first time that builds their extra stuff on May 1 and that same Docker layer may be in use unchanged when they do a `ddev start` on July 1. Because the Dockerfile content hasn't changed, no new build happens. And the entire thing is not done in the context of code... unless you do the one little thing suggested here, which may offer the possibility of contextual read-only code.

Hope that helps, and if it does, can help communicate this stuff more clearly.

rfay:
Thanks @mattstein!

docs/content/users/extend/customizing-images.md
mattstein: Thanks for that! I think I’m following and that you’re reiterating the varying degrees of difference between when images are being built and their relationship to code that runs—sort of a cautionary warning for Docker adventurers since there are potential pitfalls.

The immediate problems, I think are the “remember” and the dueling “howevers” that are just hard to parse.

More broadly, it’d be ideal to diagram or otherwise illustrate that lifecycle separately for context. Not as a matter of “fixing” peoples’ knowledge (which suggests it’s broken or insufficient) but educating them and helping them understand something you already do. _Then_ the “remember” can be justified, once that thing is clearly explained and something you can link to. This is probably a bigger and longer-term thing to improve, and I’d be happy to help with that if/when there’s a good chance.

rfay: Yeah, would love to find ways to communicate this. Maybe add to https://ddev.readthedocs.io/en/latest/users/usage/architecture/ and explain difference between image, container, and mounted code?

This class of problem is a little like helping people with getting homebrew right when they don't know how $PATH works in bash or cmd. You have no basis for the discussion, so all are flailing in the wilderness.

mattstein: > Maybe add to https://ddev.readthedocs.io/en/latest/users/usage/architecture/ and explain difference between image, container, and mounted code?

I think that‘s a step in the right direction, maybe isolated in a context/page that’s less introductory and more of a deep dive. Specifically not overwhelming to someone just getting curious about how things work, with freedom to explain and illustrate without pressure to keep it light.

While it’s kind of off-topic, I’d still like to depict Linux vs. macOS vs. Windows+WSL2 on that architecture page, since the difference (especially with WSL2) can be conceptually challenging. That’s for anyone, including beginners. The Docker composition is probably for more of a novice—or at least a more daring—DDEV user.

> so all are flailing in the wilderness.

😆 All the more reason to be a vigilant forest ranger! These big gaps are common, and trying to bridge them is hard work, but I think the result is something more welcoming and empowering for everybody trying to understand and use the thing.

rfay: Mostly linux/macOS/Windows are almost the same, with a slight tech difference for Linux. But WSL2 is baffling to people for multiple reasons. 3 levels of computer!

Issue #1858, opened 2019-10-02
app.Exec() doesn't check for running site/container


rfay:
**Describe the bug**

If the code does an app.Exec() and the project (or container) is not running, you can get an obscure result like:

`No container found for web_1
Failed to run docker-compose [-f /Users/rfay/workspace/junk/.ddev/docker-compose.yaml exec -w /var/www/html/ -T web bash -c terminus auth:login --machine-token=1BoWJ7Nhj_MCz7BXdnlqWk42PEcIegoOARIClgzoNMfzN], err='exit status 1', stdout='', stderr='No container found for web_1'`

We've seen this from time to time, mostly with exited containers. And `ddev exec` itself actually does check, and says `Project is not currently running. Try 'ddev start'`, the problem is when app.Exec is used naively.

Issue #3499, opened 2022-01-05
variables in provider environment_variables don't get properly evaluated if the value has spaces in it


schliesser:
**Describe the bug**

String variables with space are not correctly transferred from yaml to bash.

**To Reproduce**

* Create a custom provider configuration
* Add a variable which contains a space in the value
* Use the variable in a command

```yaml
environment_variables:
  test: "variable with spaces"

auth_command:
  command: |
    set -eu -o pipefail

db_pull_command:
  command: |
    # set -x # You can enable bash debugging output by uncommenting
    set -eu -o pipefail
    echo ${test}

files_pull_command:
  command: |
    # set -x   # You can enable bash debugging output by uncommenting
    set -eu -o pipefail
    echo ${test}
  service: web
```

Run with `ddev pull test -y --skip-import`

Result:

```bash
Authenticating... 
Obtaining database... 
variable
Skipping database import. 
Obtaining files... 
variable
Skipping files import. 
Pull succeeded. 
```

**Expected behavior**

The complete string should be available in the bash commands

```bash
Authenticating... 
Obtaining database... 
variable with spaces
Skipping database import. 
Obtaining files... 
variable with spaces
Skipping files import. 
Pull succeeded. 
```

**Version and configuration information (please complete the following information):**

- Host macOS Monterey 12.1 Apple M1
- Docker version 20.10.11, build dea9396
- ddev version v1.18.2

**Additional context**
https://discord.com/channels/664580571770388500/912420298488447026/928257058099302421

rfay:
Thanks!

PR #4569, opened 2023-01-20
Make `ddev ssh` exit more predictable, not outputting useless info, fixes #3738

### LINKED ISSUE #3738: `ddev ssh` should return the exit code that is returned from docker-compose and not add extra output

_No response_

### Current Behavior

When I exit from ddev ssh after having written a command in the container I get an error: "Failed to ddev ssh web: exit status 1" or "Failed to ddev ssh web: exit status 127".

This does not occur if I have not typed a command other than "exit" (i.e. entered and exited again immediately)

### Expected Behavior

no error

### Steps To Reproduce

see above

### Anything else?

_No response_

DigitalFrontiersMedia:
Attempt to check if values were piped in before deciding to return additional error output from ssh.go.  Issue 

* #3738.

## The Issue
`ddev ssh` returns output when not needed.  It should only do so when a command was piped in.

Lots of confusion on this... 
*  #1697 was in 2019... but that code isn't there any more. 

The change reversion happened in 
* https://github.com/drud/ddev/pull/2830

Which was fixing 
* #1858

It doesn't.  It's a direction that needs work.  Reading the stdin/piped-in values removes them from evaluation later down the chain.

Should NOT produce additional output and should exit with code 0:
```
ddev ssh
ls /non-existent
exit
echo $?
```

SHOULD produce error output from the piped in command:
`echo "ls /non-existent" | ddev ssh`

## Automated Testing Overview
No automated tests at this time.

## Related Issue Link(s)
#3738 

## Release/Deployment Notes
In addition to not solving the problem, `ddev ssh` "hangs" because it appears to be waiting for stdin input read.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4569"></a>

### DIFF
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -43,12 +43,6 @@ ddev ssh -d /var/www/html`,
 
 		app.DockerEnv()
 
-		// Determine if values were piped into the command.
-		pipe, err := os.Stdin.Stat()
-		if err != nil {
-			panic(err)
-		}
-		isPiped := pipe.Size() > 0
 
 		// Use bash for our containers, sh for 3rd-party containers
 		// that may not have bash.
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -55,7 +55,7 @@ ddev ssh -d /var/www/html`,
 			Cmd:     shell + " -l",
 			Dir:     sshDirArg,
 		})
-		if err != nil && isPiped {
+		if err != nil  {
 			if exiterr, ok := err.(*exec.ExitError); ok {
 				os.Exit(exiterr.ExitCode())
 			} else {
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"bufio"
+
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/nodeps"
 	"github.com/drud/ddev/pkg/util"
@@ -35,6 +37,11 @@ ddev ssh -d /var/www/html`,
 
 		app.DockerEnv()
 
+		// Determine if values were piped into the command.
+		var reader = bufio.NewReader(cmd.InOrStdin())
+		pipedValues, _ := reader.ReadString('\n')
+		isPiped := len(pipedValues) > 0
+
 		// Use bash for our containers, sh for 3rd-party containers
 		// that may not have bash.
 		shell := "bash"
@@ -46,7 +53,7 @@ ddev ssh -d /var/www/html`,
 			Cmd:     shell + " -l",
 			Dir:     sshDirArg,
 		})
-		if err != nil {
+		if err != nil && isPiped {
 			util.Failed("Failed to ddev ssh %s: %v", serviceType, err)
 		}
 	},
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -43,7 +43,6 @@ ddev ssh -d /var/www/html`,
 
 		app.DockerEnv()
 
-
 		// Use bash for our containers, sh for 3rd-party containers
 		// that may not have bash.
 		shell := "bash"
@@ -55,7 +54,7 @@ ddev ssh -d /var/www/html`,
 			Cmd:     shell + " -l",
 			Dir:     sshDirArg,
 		})
-		if err != nil  {
+		if err != nil {
 			if exiterr, ok := err.(*exec.ExitError); ok {
 				os.Exit(exiterr.ExitCode())
 			} else {
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -1,7 +1,8 @@
 package cmd
 
 import (
-	"bufio"
+	"os"
+	"os/exec"
 
 	"github.com/drud/ddev/pkg/ddevapp"
 	"github.com/drud/ddev/pkg/nodeps"
@@ -35,12 +36,19 @@ ddev ssh -d /var/www/html`,
 			util.Failed("Project is not currently running. Try 'ddev start'.")
 		}
 
+		err = app.Wait([]string{serviceType})
+		if err != nil {
+			util.Failed("Service %s doesn't seem to be running: %v", serviceType, err)
+		}
+
 		app.DockerEnv()
 
 		// Determine if values were piped into the command.
-		var reader = bufio.NewReader(cmd.InOrStdin())
-		pipedValues, _ := reader.ReadString('\n')
-		isPiped := len(pipedValues) > 0
+		pipe, err := os.Stdin.Stat()
+		if err != nil {
+			panic(err)
+		}
+		isPiped := pipe.Size() > 0
 
 		// Use bash for our containers, sh for 3rd-party containers
 		// that may not have bash.
@@ -54,7 +62,11 @@ ddev ssh -d /var/www/html`,
 			Dir:     sshDirArg,
 		})
 		if err != nil && isPiped {
-			util.Failed("Failed to ddev ssh %s: %v", serviceType, err)
+			if exiterr, ok := err.(*exec.ExitError); ok {
+				os.Exit(exiterr.ExitCode())
+			} else {
+				os.Exit(1)
+			}
 		}
 	},
 }
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -31,11 +31,6 @@ ddev ssh -d /var/www/html`,
 		app := projects[0]
 		instrumentationApp = app
 
-		status, _ := app.SiteStatus()
-		if status != ddevapp.SiteRunning {
-			util.Failed("Project is not currently running. Try 'ddev start'.")
-		}
-
 		err = app.Wait([]string{serviceType})
 		if err != nil {
 			util.Failed("Service %s doesn't seem to be running: %v", serviceType, err)


rfay:
You can push new commits into this if you think you see a path forward.

rfay:
I took the liberty of rebasing this since the updated docker-compose will be in there now. I don't think it changes anything, but you'll want to update your git checkout.

cmd/ddev/cmd/ssh.go
```diff
@@ -46,7 +53,7 @@ ddev ssh -d /var/www/html`,
 			Cmd:     shell + " -l",
 			Dir:     sshDirArg,
 		})
-		if err != nil {
+		if err != nil && isPiped {
 			util.Failed("Failed to ddev ssh %s: %v", serviceType, err)
 		}
```
rfay, line 58: The easy part of this PR is to remove the output on failure as it's not useful, and in the case of `ddev ssh` was already output if needed, and just exit with the exit code, which can be done something like this:

```suggestion
		if err != nil { // Didn't experiment with your isPiped here...
			if exiterr, ok := err.(*exec.ExitError); ok {
				os.Exit(exiterr.ExitCode())
			} else {
				os.Exit(1)
			}
		}
```

rfay:
Oh, if you don't have GoLand (or vscode) set up to let you debug through things, make sure to make that a priority. I love Goland, I think you will too. Lots of people use vscode successfully, but it's more tweaky, and I am not a master of the tweaks. But make sure you can do step-debugging. Of course, the pipe thing you're trying to do here is one of the hardest things to do with step debugging.

cmd/ddev/cmd/ssh.go
DigitalFrontiersMedia: Interesting.  So many moving parts on top of just getting my head around Go, so every little suggestion is appreciated.  Will check this out!

DigitalFrontiersMedia:
I'm using VSCode.  I'm cash strapped at the moment and I don't see paying for another tool anytime soon.  And yeah, I was wondering how you do step debugging with Go.  I tried a cursory search but didn't really understand the suggested results.  If you have specific suggestions for VSCode, I'm all ears.  Will keep looking at these issues as I learn though.  Thank you for the opportunity and help!

rfay:
Yeah, IMO the first thing to do with any development project in any language is to get step-debugging going, so spend some time with it. There are docs around, and the Gitpod integration is all set up with it, so you could debug with this PR at https://gitpod.io/#https://github.com/drud/ddev/pull/4569, but since you're an established vscode user I'd try to get it set up locally as well.

rfay:
So see
* https://github.com/docker/compose/issues/10189

It turns out that the phony exit codes are a feature of bash. So all we really need to do here is
* Stop outputting any verbiage
* Return the exit code as in my suggestion above
* Perhaps verify that the target container is running before trying to use it, perhaps something like this:
```
		err = app.Wait([]string{serviceType})
		if err != nil {
			util.Failed("Service %s doesn't seem to be running: %v", serviceType, err)
		}
```

DigitalFrontiersMedia:
I'm not sure, but I think this is now doing what you expect?

```
(base) stephen@MacBook-Pro:dfm-bdw [16:10:22] (master): ddev ssh                 
stephen@dfm-bdw-web:/var/www/html$ ls /xxx
ls: cannot access '/xxx': No such file or directory
stephen@dfm-bdw-web:/var/www/html$ exit
logout
(base) stephen@MacBook-Pro:dfm-bdw [16:15:07] (master): echo $? 
0
```

```
(base) stephen@MacBook-Pro:dfm-bdw [16:15:10] (master): echo "ls /xxx" | ddev ssh
ls: cannot access '/xxx': No such file or directory
(base) stephen@MacBook-Pro:dfm-bdw [16:15:15] (master): echo $?                  
2
```

cmd/ddev/cmd/ssh.go
```diff
@@ -33,8 +36,20 @@ ddev ssh -d /var/www/html`,
...
+		}
+
 		app.DockerEnv()
 
+		// Determine if values were piped into the command.
+		pipe, err := os.Stdin.Stat()
+		if err != nil {
+			panic(err)
+		}
+		isPiped := pipe.Size() > 0
```
rfay, line 51: I might be wrong, but I don't think we need this at all any more. 

```suggestion
```

cmd/ddev/cmd/ssh.go
```diff
@@ -46,8 +61,12 @@ ddev ssh -d /var/www/html`,
 			Cmd:     shell + " -l",
 			Dir:     sshDirArg,
 		})
-		if err != nil {
-			util.Failed("Failed to ddev ssh %s: %v", serviceType, err)
+		if err != nil && isPiped {
```
rfay, line 64:
```suggestion
		if err != nil  {
```

rfay (changes requested):
Making progress! I think we can get rid of more now :)

I'm going to take the liberty of rebasing this now, because you now have lots of bits and pieces you don't want, so when you start working on it, make sure your git remote is updated.

rfay (changes requested):
IMO this is perfect now. Behavior is as expected. The one thing I'd like to ask is to remove 
https://github.com/drud/ddev/blob/995cc6beffe375a5ada8ecdecf17ac707b608ad0/cmd/ddev/cmd/ssh.go#L34-L37

The reason is... we don't have to check siteStatus if the container is running. I think this might be better in the case where one container failed.

DigitalFrontiersMedia:
Yes, I've been meaning all day to say that I think you had it and to thank you for your patience with my tunnel vision as I now totally see what you were getting at the entire time and how what was in the pipe was irrelevant. But I wanted to test it before reporting back on those changes but I seemed to have done something unexplainable that prevents me from running the compiled project.  So I haven't been able to test it.

I am using an M1 (arm64), and for some reason (probably from a previous bad install), things were compiling amd64.  Oddly, now that I think it's "fixed" I get `zsh: killed     ddev` whenever I try to run the compiled project.  But all seems correct:

```
(base) stephen@MacBook-Pro:~ [14:36:43]: go env GOHOSTOS
darwin
(base) stephen@MacBook-Pro:~ [15:00:04]: go env GOHOSTARCH
arm64
(base) stephen@MacBook-Pro:~ [15:00:08]: which ddev
/usr/local/bin/ddev
(base) stephen@MacBook-Pro:~ [15:00:21]: ls -alh /usr/local/bin/ddev
-rwxr-xr-x  1 stephen  wheel    17M Jan 26 14:38 /usr/local/bin/ddev
```
That's definitely the executable I just compiled with "make" but still I get `zsh: killed     ddev`.

I've uninstalled and re-installed Go and tools and blew away my local of ddev and re-cloned.
```
(base) stephen@MacBook-Pro:~ [15:00:34]: go version
go version go1.19.5 darwin/arm64
```

Don't see how it has anything to do with ddev, so trying to figure out what's wrong and will circle back.

Your signature on this quickfix should be "If things get any worse, I'm going to have to ask you to stop helping."  :-D

DigitalFrontiersMedia:
So weird.  Compiling for amd64 works:
```
(base) stephen@MacBook-Pro:dfm-ddev [15:10:16] (master): make darwin_amd64 && cp .gotmp/bin/darwin_amd64/ddev /usr/local/bin/ddev
building .gotmp/bin/darwin_amd64/ddev from ./cmd/... ./pkg/...
(base) stephen@MacBook-Pro:dfm-ddev [15:10:24] (master): ddev version
 ITEM             VALUE                                   
 DDEV version     v1.21.4-57-g1239754d                    
 architecture     amd64                                   
 db               drud/ddev-dbserver-mariadb-10.4:v1.21.4 
 dba              phpmyadmin:5                            
 ddev-ssh-agent   drud/ddev-ssh-agent:v1.21.4             
 docker           20.10.17                                
 docker-compose   v2.15.1                                 
 docker-platform  docker-desktop                          
 mutagen          0.16.0                                  
 os               darwin                                  
 router           drud/ddev-router:v1.21.4                
 web              drud/ddev-webserver:v1.21.4             
```

DigitalFrontiersMedia:
Running amd64, I get the following with the above code, which I believe isn't what you want due to the non-zero exit code on exit:
```
(base) stephen@MacBook-Pro:dfm-bdw [15:14:25] (master): ddev ssh
stephen@dfm-bdw-web:/var/www/html$ ls /xxx
ls: cannot access '/xxx': No such file or directory
stephen@dfm-bdw-web:/var/www/html$ exit
logout
(base) stephen@MacBook-Pro:dfm-bdw [15:14:39] (master): echo $?
2
(base) stephen@MacBook-Pro:dfm-bdw [15:14:46] (master): echo "ls /xxx" | ddev ssh
ls: cannot access '/xxx': No such file or directory
(base) stephen@MacBook-Pro:dfm-bdw [15:15:03] (master): echo $?                  
2
```

but I don't really trust my environment right now, maybe it's a Rosetta thing?

DigitalFrontiersMedia:
"The reason is... we don't have to check siteStatus if the container is running. I think this might be better in the case where one container failed."

Yes, I didn't think that was necessary for same reason but you had suggested it earlier and I just thought I was missing something.  Will address that after I figure out my local issues.

rfay:
You're doing great, congratulations!

The problem with your build is you have some bad-architecture `ddev` in your $PATH. Delete them all :) If you do a `file $(which ddev)` you'll see the architecture of the bad one.

I don't recommend running amd64 *anything* on arm64. Just confuses everything.

If you just run `make` it will build ddev into project .gotmp/bin/darwin_arm64/ddev - you can explicitly path that if you want, but I recommend using another technique.  I always symlink my /usr/local/bin to that file.

DigitalFrontiersMedia:
> The problem with your build is you have some bad-architecture ddev in your $PATH. Delete them all :) If you do a file $(which ddev) you'll see the architecture of the bad one.

When I run `make && cp .gotmp/bin/darwin_arm64/ddev /usr/local/bin/ddev` and do as you suggest, I get the following:
```
(base) stephen@MacBook-Pro:~ [15:39:31]: file $(which ddev)
/usr/local/bin/ddev: Mach-O 64-bit executable arm64
(base) stephen@MacBook-Pro:~ [15:41:37]: ddev version
zsh: killed     ddev version
```

> If you just run make it will build ddev into project .gotmp/bin/darwin_arm64/ddev - you can explicitly path that if you want, but I recommend using another technique. I always symlink my /usr/local/bin/ddev to that file.

Yes, I've just been copying that (as noted above).  But if I do `make darwin_amd64 && cp .gotmp/bin/darwin_amd64/ddev /usr/local/bin/ddev` instead, everything works...but I don't trust it and it makes no sense:

```
(base) stephen@MacBook-Pro:dfm-ddev [15:41:32] (issue3738): make darwin_amd64 && cp .gotmp/bin/darwin_amd64/ddev /usr/local/bin/ddev
building .gotmp/bin/darwin_amd64/ddev from ./cmd/... ./pkg/...
(base) stephen@MacBook-Pro:dfm-ddev [15:47:54] (issue3738): ddev version
 ITEM             VALUE                                   
 DDEV version     v1.21.4-58-g726740e4                    
 architecture     amd64                                   
 db               drud/ddev-dbserver-mariadb-10.4:v1.21.4 
 dba              phpmyadmin:5                            
 ddev-ssh-agent   drud/ddev-ssh-agent:v1.21.4             
 docker           20.10.17                                
 docker-compose   v2.15.1                                 
 docker-platform  docker-desktop                          
 mutagen          0.16.0                                  
 os               darwin                                  
 router           drud/ddev-router:v1.21.4                
 web              drud/ddev-webserver:v1.21.4             
```

DigitalFrontiersMedia:
Twilight Zone level stuff.  Restarted the whole computer.  And now seems to `make` running arm64 versions again (nice loss of a few hours there lol).

That fixed, I don't think that code is working the way you want it to, because if an error is returned from the previous command per os, it spits that out (code 2) and you want 0 on exit:
```
(base) stephen@MacBook-Pro:dfm-bdw [15:56:21] (master): ddev ssh
stephen@dfm-bdw-web:/var/www/html$ ls /xxx
ls: cannot access '/xxx': No such file or directory
stephen@dfm-bdw-web:/var/www/html$ exit
logout
(base) stephen@MacBook-Pro:dfm-bdw [15:58:34] (master): echo $?
2
```

This actually worked correctly with the pipe check code in there because if you're not piping anything, it returns nothing:
and I just made the change and recompiled to show the output and getting `zsh: killed  ddev ssh` again. <headdesk>

rfay:
The case you show is bash's fault, and it's OK. As explored in 
* https://github.com/docker/compose/issues/10189

I don't think there's a "fix" that makes any sense if that's the behavior of bash. But now we don't whine about it with extra output.

rfay:
Use `hash -r` when you change that affects your PATH, and I recommend a symlink,
`ln -sf <project>/.gotmp/bin/darwin_arm64/ddev /usr/local/bin/ddev`

DigitalFrontiersMedia:
> I don't think there's a "fix" that makes any sense if that's the behavior of bash. But now we don't whine about it with extra output.

Oh. Okay.  Was going off of something you said last week in https://github.com/drud/ddev/issues/3738#issuecomment-1397623130:

> Not only is the reporting about "failed to ddev ssh web" incorrect, but the return value should actually be 0.

Because if you really want it 0, I was getting that by bypassing that whole return block if we weren't expecting a return value for a piped in command.

rfay:
That was before I studied the bash behavior.

DigitalFrontiersMedia:
Ah!  Understood.  And I also misread the block you wanted removed.  Thought you were asking to remove the service check (not the SiteRunning check).

I'm 2 steps behind but catching up!

DigitalFrontiersMedia:
Give it a look now.

(symlinking was the answer, btw, copying it = bad)

rfay:
I hope you'll be ready for another challenge! Maybe
* https://github.com/drud/ddev/issues/3499 ?

rfay (approved):
Looking great to me! I'll manually test one more time before pull, but congratulations!

DigitalFrontiersMedia:
> I hope you'll be ready for another challenge! Maybe
> 
> * [variables in provider environment_variables don't get properly evaluated if the value has spaces in it #3499](https://github.com/drud/ddev/issues/3499) ?

Absolutely!  Thank you again for the opportunity and the guidance, even that out of scope of the issue.  And I'll even name the branch correctly as requested for the PR!  :-)

Thank you for the `hash -r` suggestion, too.  I had no idea such a thing existed but fits perfectly with my spidey sense that something akin to a caching problem seemed to be occurring.  Feel free to point out other things I'm missing as I go along, such as at this point if I should click "Comment" or "Close with comment" when we've reached a resolution?  Also, what are expectations for when in a thread the "Update branch" button should be used?

rfay:
I always try to keep PRs rebased to current upstream/master. This can be done 
* locally with `git fetch upstream && git rebase upstream/master && git push --force-with-lease`
* Locally by merging upstream/master in. `git fetch upstream && git merge upstream/master` - I don't ever do this because merge commits hide lots of stuff and I don't like that hidden stuff. But the reality is it works because we always squash PRs here.
* On Github web UI with "update branch", again, I always prefer and recommend rebase.

rfay:
I think you're all done here, but general workflow is:
1. Sometimes start with PR in "draft" mode. Doesn't matter too much.
2. When it's maturing, change to "ready for review"
3. When you want review, request review.

When it gets merged, the PR will be closed. If it says "fixes ..." like the title here does, it will auto-close the related issue, #3738 in this case.

rfay (approved):
Looks perfect; I did another manual test and things look just right. 

It has a great side-effect that I've always wanted. Now when you have a failure to start due to some random container failing, you can still `ddev ssh` into the other containers that were successfully started.

rfay:
Congratulations!

PR #4576, opened 2023-01-23
[tests only] Fix broken docker context list problem, for docker/for-win/#13180

rfay:
## The Issue

* https://github.com/docker/for-win/issues/13180 breaks tests on Windows and wsl2/Docker Desktop

Delete the ~/.docker/contexts on starting tests if it's broken

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4576"></a>

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -5,13 +5,6 @@ export PATH=$PATH:/home/linuxbrew/.linuxbrew/bin
 
 echo "buildkite building ${BUILDKITE_JOB_ID:-} at $(date) on $(hostname) as USER=${USER} for OS=${OSTYPE} in ${PWD} with golang=$(go version | awk '{print $3}') docker-desktop=$(scripts/docker-desktop-version.sh) docker=$(docker --version | awk '{print $3}') ddev version=$(ddev --version | awk '{print $3}'))"
 
-# Broken docker context list from https://github.com/docker/for-win/issues/13180
-# When this is solved this can be removed.
-# The only place we care about non-default context is macOS Colima
-if ! docker context list >/dev/null; then
-  rm -rf ~/.docker/contexts && docker context list >/dev/null
-fi
-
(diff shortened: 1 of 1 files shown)


yilmazdurmaz:
hi @rfay,
I thought you implemented some file testing and removal of only broken contexts, but this `Delete the ~/.docker/contexts` is totally not acceptable and does more harm than it solves.
what if I made multiple custom contexts and only 1 of them was broken. you can confirm this by creating a few more contexts and removing content from one of their `meta.json` file.

````
docker context create test1      # create a new context
docker context create test2      # create another context
docker context ls                # runs fine and will show at least 3 contexts
docker context inspect test1     # get the path from `Storage` property
````
for the test1 context, edit `meta.json` and just remove content and save it.
```
docker context ls                # now you will get only 1 entry because you have deleted all contexts.
```

rfay:
@yilmazdurmaz This is for tests only, has nothing to do with the operation of DDEV, it's a workaround to keep the Docker Desktop for Windows test runners operating until https://github.com/docker/for-win/issues/13180 gets fixed. 

https://github.com/drud/ddev/blob/1f93c39a41e94f2c3ef7df4a06e0c2a97c4fde70/.buildkite/test.sh#L18-L23

In the case of traditional Windows and WSL2, DDEV doesn't use contexts for anything, and this workaround to keep the tests running seems to be adequate. 

Again, has nothing to do with DDEV's operation, and DDEV would never do anything like this. But I'm glad you're paying attention!

yilmazdurmaz:
> workaround to keep the tests running

ok then :) test runners can be remote and do not need many contexts. just make sure this does not slip out to unsuspecting people having many contexts ;)

I actually wrote another bug report to the docker repository about the `\x00` error (12561), and this "JSON" error is a sibling to it. contributors tend to link mine to them and close others in relation to mine (your has it too). I was browsing these issues lately and warn people before deleting things and found a link to this commit, so I felt the urge here too.

PR #4578, opened 2023-01-24
Disable xdebug when running composer

### LINKED ISSUE #4219: Executing `ddev composer` should disable XDebug

When executing `ddev composer ...` while XDebug is enabled, XDebug will debug the composer PHP script. This will lead to the script pausing.

### Describe your solution

Composer should be executed with XDebug disabled. This can be achieved by disabling XDebug. One way is to set the environment variable `XDEBUG_MODE=off`.

### Describe alternatives

Disabling XDebug (effectively running `ddev xdebug off`) before executing the composer command. Then re-enabling XDebug (effectively running `ddev xdebug on`).

A further extension of the feature could be that XDebug is only disabled for the composer script itself, but not for PHP scripts executed by composer, e.g. still having XDebug enabled when executing the unittests script `ddev composer unittests`.

### Additional context

_No response_

brianjhanson:
## The Issue
When running `ddev composer` with xdebug enabled it will start debugging.

Adds `XDEBUG_MODE=off` before the `composer` command. 

## Automated Testing Overview
I'm sure some tests would be good, but I'm not very familiar with Go so I'd need some guidance on where I should put them.

## Related Issue Link(s)
Resolves #4219 

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4578"></a>

### DIFF
--- a/pkg/ddevapp/composer.go
+++ b/pkg/ddevapp/composer.go
@@ -20,7 +20,7 @@ func (app *DdevApp) Composer(args []string) (string, string, error) {
 	stdout, stderr, err := app.Exec(&ExecOpts{
 		Service: "web",
 		Dir:     app.GetComposerRoot(true, true),
-		RawCmd:  append([]string{"composer"}, args...),
+		RawCmd:  append([]string{"XDEBUG_MODE=off composer"}, args...),
 		Tty:     isatty.IsTerminal(os.Stdin.Fd()),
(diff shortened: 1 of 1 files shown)


brianjhanson:
Looks like this breaks all kinds of things. Going to close this until I have more time to dig in.

PR #4590, opened 2023-01-26
Update php-fpm.conf

skleiber:
adapt include path to 8.2 (was 8.0)

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4590"></a>


rfay:
Thanks for paying attention and catching this! But it would sure be less messy if you did a single PR. 

Of course this doesn't mean anything either without pushing the resulting image.

PR #4597, opened 2023-01-27
Document disabling xdebug in RUN commands

deviantintegral:
## The Issue

`ddev start` was hanging for members on our team unexpectedly. The only way I figured this out was when I tried running the container on a completely different computer:

```
docker run --rm -it drud/ddev-webserver:v1.21.4 terminus
```

That showed xdebug failing to connect, and then I could look at `.ddev/.webimageBuild/Dockerfile` to see why xdebug was triggering at all.

I assume there's good reasons why we keep xdebug on. If not, turning it off during project dockerfile builds could be a better approach.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4597"></a>

### DIFF
--- a/docs/content/users/usage/troubleshooting.md
+++ b/docs/content/users/usage/troubleshooting.md
@@ -241,6 +241,14 @@ The error messages you get will be more informative than messages that come when
 
 You can also see the full Docker build using `~/.ddev/bin/docker-compose -f .ddev/.ddev-docker-compose-full.yaml build --no-cache --progress=plain`.
 
+### Disable Xdebug When Running PHP Applications like Composer or Terminus
+
+During the `docker build` process Xdebug is enabled and configured. If you have an IDE open and listening for connections, `ddev start` can hang. Always disable Xdebug in each `RUN` command:
+
+```dockerfile
+RUN export XDEBUG_MODE=off; terminus self:plugin:install terminus-build-tools-plugin
+```
+
 ## DDEV Starts but Browser Can’t Access URL
 
 You may see one of two messages in your browser:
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/troubleshooting.md
```diff
@@ -241,6 +241,23 @@ The error messages you get will be more informative than messages that come when
...
 
+### Disable Xdebug When Running PHP Applications like Composer or Terminus
+
+During the `docker build` process Xdebug is enabled and configured. If you have an IDE open and listening for connections, `ddev start` can hang. Always disable Xdebug in each `RUN` command:
+
+For commands that need to run as root:
+
+```dockerfile
+# Example only - ddev already runs self-update for you.
+RUN export XDEBUG_MODE=off; composer self-update
```
rfay, line 252: I think this would be cleaner...

```suggestion
RUN XDEBUG_MODE=off composer self-update
```

Or how about `ENV XDEBUG_MODE=off` - and why isn't that already in DDEV? Why would we have the extension even enabled at build time?

rfay (changes requested):
Seems to me like the problem is that xdebug shouldn't even be enabled as an extension at build time, or that at least DDEV could `ENV XDEBUG_ENABLED=off` at the minimum.

rfay:
The extension is in fact disabled at build time... https://github.com/drud/ddev/blob/876ae3aaf47f80933898a0b672234915876faa01/containers/ddev-webserver/Dockerfile#L97

rfay:
Xdebug being enabled by default in the ddev-webserver image itself happened because people begged for it in 
* https://github.com/drud/ddev/issues/2245
* https://github.com/drud/ddev/pull/2460

But that was back when PhpStorm integration was far, far different. I think we should consider not doing that any more. 

I think it doesn't belong that way any more.

rfay:
I think 
* https://github.com/drud/ddev/pull/4599

Is probably the answer, sure appreciate you bringing this up.

rfay:
Closing in favor of
* #https://github.com/drud/ddev/pull/4599

Hope you can review that, maybe even try it out.

PR #4599, opened 2023-01-28
Don't enable xdebug by default in ddev-webserver, fixes #4597

rfay:
## The Issue

It doesn't make sense for Xdebug to be enabled by default in ddev-webserver any more:
* #4597 

It was enabled for very different workflows back in the day, see 
* https://github.com/drud/ddev/issues/2245
* https://github.com/drud/ddev/pull/2460

I don't think those workflows are valid any more, but will ping those folks.

Turn off xdebug after it's installed.

As @deviantintegral did in #4597, `docker run -it --rm drud/ddev-webserver:20230128_xdebug_build_time
php --version` and see if xdebug is shown there.

## TODO

It may be worth trying to fix `ddev composer`,  so they doesn't try to use xdebug as well.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4599"></a>


deviantintegral:
This is great. I much prefer this to what was in place before. I probably won't get a chance to test this out until next week. Thank you!

rfay:
Artifacts for testing are at https://nightly.link/drud/ddev/workflows/master-build/master, thanks!

PR #4600, opened 2023-01-29
[docs] Add Drupal Multisite recommendations

laetus007:
Changes are from Discord DDEV multisite [forum question posed by KristofferRom](https://discord.com/channels/664580571770388500/1037263797674594334).  File examples are from Jon McLaughlin and tested by myself  with an existing codebase deployed on Acquia Cloud Site Factory.

## The Issue
**Original Question**
Anyone here with experience in converting separate sites to one multisite using ddev as local environment? I'm aware of https://github.com/drud/ddev-contrib/tree/master/recipes/drupal8-multisite but I was thinking to use an existing site as the code base so installing with --disable-settings-management isn't an option. Is there a workaround or should I create a new installation for the code base?

**Additional Requirements**
We run a Drupal Multisite setup where the default DB manages the bulk of configuration for all our sites.  We have three specific MVP's where we will run the config updates locally before committing to our hosting environment but never export config from them.  

This PR adds additional steps in documentation to assist others that might have a similar setup. 

No tests as this for documentation purposes only.

## Automated Testing Overview
No tests as this for documentation purposes only.

## Related Issue Link(s)
None
## Release/Deployment Notes

N/A

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4600"></a>

### DIFF
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -78,7 +78,7 @@ require $app_root . '/sites/default/settings.ddev.php';
 $databases['default']['default']['database'] = 'site_name';
 ```
 3. if you are using site aliases add the following to your `config.yaml`
-```php   
+```yaml   
 web_environment:
   # Make ddev drush shell PIDs last for entire life of container. So that `ddev drush site:set @alias` will persist
   # for all drush connections.
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -45,6 +45,47 @@ This environment variable is set `true` by default in DDEV’s environment, and 
 ```php
 $settings['container_yamls'][] = DRUPAL_ROOT . '/sites/development.services.yml';
 ```
+* **Multisite**:
+    * See [DDEV-Local Drupal 8 Multisite Recipe] (https://github.com/drud/ddev-contrib/tree/master/recipes/drupal8-multisite) to start
+    * update the following files if you are using an existing site as the code base where installing with `--disable-settings-management` is not an option. 
+1. each `site/{site_name}/settings.php` 
+ ```php
+ /**
+ * ddev local-dev environments will have $databases (and other settings)
+ * set by an auto-generated file. Make alterations here for this site
+ * in a multisite environment.
+ */
+elseif (getenv('IS_DDEV_PROJECT') == 'true') {
+
+  /**
+   * Alter database settings and credentials for ddev-local environment.
+   * This includes loading the ddev-generated default/settings.ddev.php.
+   */
+  include $app_root . '/' . $site_path . '/settings.databases.ddev.inc';
+}
+ ```
+2.  add a `settings.databases.ddev.inc` in each `site/{site_name}/`   
+ ```php
+ /**
+ * Fetch ddev generated database credentials and other settings.
+ */
+require $app_root . '/sites/default/settings.ddev.php';
+
+/*
+ * Alter default database for this site. settings.ddev.php will have "reset"
+ * this to 'db'
+ */
+$databases['default']['default']['database'] = 'site_name';
+```
+3. if you are using site aliases add the following to your `config.yaml`
+```php   
+web_environment:
+  # Make ddev drush shell PIDs last for entire life of container. So that `ddev drush site:set @alias` will persist
+  # for all drush connections.
+  # https://chrisfromredfin.dev/posts/drush-use-ddev/
+  - DRUSH_SHELL_PID=PERMANENT
+```
+
 
 ### TYPO3 Specifics
 
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -47,7 +47,7 @@ $settings['container_yamls'][] = DRUPAL_ROOT . '/sites/development.services.yml'
 ```
 * **Multisite**:
     * See [DDEV-Local Drupal 8 Multisite Recipe] (https://github.com/drud/ddev-contrib/tree/master/recipes/drupal8-multisite) to start
-    * update the following files if you are using an existing site as the code base where installing with `--disable-settings-management` is not an option. 
+    * update the following files
 1. each `site/{site_name}/settings.php` 
  ```php
  /**
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -36,56 +36,69 @@ This environment variable is set `true` by default in DDEV’s environment, and 
 
 ### Drupal Specifics
 
-* **Settings Files**: By default, DDEV will create settings files for your project that make it “just work” out of the box. It creates a `sites/default/settings.ddev.php` and adds an include in `sites/default/settings.php` to bring that in. There are guards to prevent the `settings.ddev.php` from being active when the project is not running under DDEV, but it still should not be checked in and is gitignored.
-* **Database requirements for Drupal 9.5+ +**:
-    * Using MySQL or MariaDB, Drupal requires `SET GLOBAL TRANSACTION ISOLATION LEVEL READ COMMITTED` and DDEV does this for you on [`ddev start`](../usage/commands.md#start).
-    * Using PostgreSQL, Drupal requires the`pg_trm` extension. DDEV creates this extension automatically for you on `ddev start`.
-* **Twig Debugging**: With the default Drupal configuration, it’s very difficult to debug Twig templates; you need to use `development.services.yml` instead of `services.yml`. Add this line in your `settings.php` or `settings.local.php`. See discussion at [drupal.org](https://www.drupal.org/forum/support/module-development-and-code-questions/2019-09-02/ddev-twig-debugging) and the Drupal documentation.
+#### Settings Files
+
+By default, DDEV will create settings files for your project that make it “just work” out of the box. It creates a `sites/default/settings.ddev.php` and adds an include in `sites/default/settings.php` to bring that in. There are guards to prevent the `settings.ddev.php` from being active when the project is not running under DDEV, but it still should not be checked in and is gitignored.
+
+#### Database requirements for Drupal 9.5+
+
+* Using MySQL or MariaDB, Drupal requires `SET GLOBAL TRANSACTION ISOLATION LEVEL READ COMMITTED` and DDEV does this for you on [`ddev start`](../usage/commands.md#start).
+* Using PostgreSQL, Drupal requires the`pg_trm` extension. DDEV creates this extension automatically for you on `ddev start`.
+
+#### Twig Debugging
+
+With the default Drupal configuration, it’s very difficult to debug Twig templates; you need to use `development.services.yml` instead of `services.yml`. Add this line in your `settings.php` or `settings.local.php`. See discussion at [drupal.org](https://www.drupal.org/forum/support/module-development-and-code-questions/2019-09-02/ddev-twig-debugging) and the Drupal documentation.
 
 ```php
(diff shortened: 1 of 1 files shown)


JonMcL (approved):
Looks good to me. Thank you for making these changes.

rfay:
You say
>  I was thinking to use an existing site as the code base so installing with --disable-settings-management isn't an option

I don't really understand your comment here... One can always switch from using settings management to not using it... It has nothing to do with installation... Do we have something to talk about here?

docs/content/users/usage/cms-settings.md
```diff
@@ -45,6 +45,47 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
+require $app_root . '/sites/default/settings.ddev.php';
+
+/*
+ * Alter default database for this site. settings.ddev.php will have "reset"
+ * this to 'db'
+ */
+$databases['default']['default']['database'] = 'site_name';
+```
+3. if you are using site aliases add the following to your `config.yaml`
+```php
```
rfay, line 81:
```suggestion
```yaml   
```

rfay (changes requested):
You'll see the spellcheck is failing in https://github.com/drud/ddev/actions/runs/4037036738/jobs/6959971014 meaning that there's something funky about the ``` fences. I wasn't able to see it just scanning through. But the spellchecker doesn't look inside those.

laetus007:
> You say
> 
> > I was thinking to use an existing site as the code base so installing with --disable-settings-management isn't an option
> 
> I don't really understand your comment here... One can always switch from using settings management to not using it... It has nothing to do with installation... Do we have something to talk about here?

This was added based on the forum question and, in reflection, not needed as I didn't do this when I spun up my instance.

JonMcL:
> > You say
> > > I was thinking to use an existing site as the code base so installing with --disable-settings-management isn't an option
> > 
> > 
> > I don't really understand your comment here... One can always switch from using settings management to not using it... It has nothing to do with installation... Do we have something to talk about here?
> 
> This was added based on the forum question and, in reflection, not needed as I didn't do this when I spun up my instance.

I didn't notice that --disable-settings-management is true on that suggested configuration. If settings management is disabled, then there will be no sites/default/settings.ddev.php file for the multisites to include. I don't think we would want the new instructions to go with the original suggested configuration.

rfay:
I don't really recommend ever using a modified `settings.ddev.php` because in a normal situation it will be overwritten by ddev on start. Best to use a different name?

laetus007:
I updated the code to remove this line - interesting the check failed again - I'll take a look

JonMcL:
> I don't really recommend ever using a modified `settings.ddev.php` because in a normal situation it will be overwritten by ddev on start. Best to use a different name?

The guidance from the readme is to use the ddev-generated settings.ddev.php, not to modify it. Each site has a settings.databases.ddev.inc file that has an include to pull in the ddev-generated settings.ddev.php. It could maybe be simplified by moving the contents of setttings.databases.ddev.inc to the sites/{site}/settings.php file, but by keeping them separate it gives the developer the option of leaving settings.databases.ddev.inc out of the repo.

rfay:
> The guidance from the readme

Maybe you can say which readme?

laetus007:
> > > You say
> > > > I was thinking to use an existing site as the code base so installing with --disable-settings-management isn't an option
> > > 
> > > 
> > > I don't really understand your comment here... One can always switch from using settings management to not using it... It has nothing to do with installation... Do we have something to talk about here?
> > 
> > 
> > This was added based on the forum question and, in reflection, not needed as I didn't do this when I spun up my instance.
> 
> I didn't notice that --disable-settings-management is true on that suggested configuration. If settings management is disabled, then there will be no sites/default/settings.ddev.php file for the multisites to include. I don't think we would want the new instructions to go with the original suggested configuration.

I've removed this line from the documentation.  It clearly was not needed and from the conversation here, clearly it's needed otherwise the user is set doing all settings per the .ddev/config.yml which has the following - total newb issue on my end . . . coming over from lando and now only a week into development with ddev.  Huge kudos to everyone - its sooooooo much faster and out of the box easier to setup than I found going from docksal to lando.

`# disable_settings_management: false
# If true, ddev will not create CMS-specific settings files like
# Drupal's settings.php/settings.ddev.php or TYPO3's AdditionalConfiguration.php
# In this case the user must provide all such settings.`

docs/content/users/usage/cms-settings.md
```diff
@@ -36,16 +36,70 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
+
+1. each `site/{site_name}/settings.php`
+
+     ```php
+     /**
+     * ddev local-dev environments will have $databases (and other settings)
+     * set by an auto-generated file. Make alterations here for this site
+     * in a multisite environment.
+     */
+    elseif (getenv('IS_DDEV_PROJECT') == 'true') {
```
rfay, line 69: This starts with elseif, I'm sure that means something is wrong here.

I also took the liberty of rebasing and fiddling with the flow here, and think I've fixed the spellcheck problem; looks to me like a bug in the spellchecker.

rfay (changes requested):

rfay:
Rebased, make sure you pull before working on it.

tyler36:
If we are recommending the separate databases appraoch, perhaps there should be a note about how `ddev snapshot` only targets the `db` database.

rfay:
Agreed; doesn't everybody do Drupal multisite with separate databases? Or do they sometimes use table prefixes?

rfay:
But actually, do test... I think you'll find that snapshots get the entire database, not just the one db. export-db is per-db, but I think you'll find snapshot gets everything.

tyler36:
@rfay Your right. Just checked it now.

Snapshot does get the both dbs.  So maybe add a note as you said [above](https://github.com/drud/ddev/pull/4600#issuecomment-1418324425).

## Test

1. Setup multisite 
2. Create "base" snapshot
3. Add unique user to site 1.
4. Add unique user to site 2.
5. Create "user" snapshot
6. Restore "base" snapshot; both databases are missing the unique users.
7. Restore "user" snapshot; both databases contain the unique users.

laetus007:
> Agreed; doesn't everybody do Drupal multisite with separate databases? Or do they sometimes use table prefixes?

Separate dbs in our case!

Ok - should I update the patch?

rfay:
I think it's fine; people can figure out their own backup routines.

rfay:
Rebased; sorry I lost track of where we were on this. It's definitely a useful addition.

docs/content/users/usage/cms-settings.md
```diff
@@ -36,16 +36,70 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
+3. if you are using site aliases add the following to your `config.yaml`
+
+    ```yaml
+    web_environment:
+      # Make ddev drush shell PIDs last for entire life of container. So that 'ddev drush site:set @alias' will persist
+      # for all drush connections.
+      # https://chrisfromredfin.dev/posts/drush-use-ddev/
+      - DRUSH_SHELL_PID=PERMANENT
+    ```
+
```
mattstein, line 102: Continue numbered sequence, clarify what step two encapsulates, edit down and re-wrap code comments so they’re readable in browser, capitalize Drush and DDEV proper nouns, and link to `web_environment` reference:

```suggestion
2. Update configuration files.
    1. Update each `site/{site_name}/settings.php`:

        ```php
        /**
         * DDEV environments will have $databases (and other settings) set
         * by an auto-generated file. Make alterations here for this site
         * in a multisite environment.
         */
        elseif (getenv('IS_DDEV_PROJECT') == 'true') {
          /**
           * Alter database settings and credentials for DDEV environment.
           * Includes loading the DDEV-generated `default/settings.ddev.php`.
           */
          include $app_root . '/' . $site_path . '/settings.databases.ddev.inc';
        }
        ```

    2. Add a `settings.databases.ddev.inc` in each `site/{site_name}/`:

        ```php
        /**
         * Fetch DDEV-generated database credentials and other settings.
         */
        require $app_root . '/sites/default/settings.ddev.php';
        
        /*
         * Alter default database for this site. `settings.ddev.php` will have 
         * “reset” this to 'db'.
         */
        $databases['default']['default']['database'] = 'site_name';
        ```

    3. Update your [`web_environment`](../configuration/config.md#web_environment) config option if you’re using site aliases:

        ```yaml
        web_environment:
          # Make DDEV Drush shell PIDs last for entire life of the container
          # so `ddev drush site:set @alias` persists for all Drush connections.
          # https://chrisfromredfin.dev/posts/drush-use-ddev/
          - DRUSH_SHELL_PID=PERMANENT
        ```
```

mattstein (changes requested):
Thanks for this, @laetus007! Made a few suggestions to improve formatting, which will probably make more sense if you look at the [preview build output](https://ddev--4600.org.readthedocs.build/en/4600/users/usage/cms-settings/#cms-specific-help-and-techniques).

docs/content/users/usage/cms-settings.md
```diff
@@ -36,16 +36,70 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
+With the default Drupal configuration, it’s very difficult to debug Twig templates; you need to use `development.services.yml` instead of `services.yml`. Add this line in your `settings.php` or `settings.local.php`. See discussion at [drupal.org](https://www.drupal.org/forum/support/module-development-and-code-questions/2019-09-02/ddev-twig-debugging) and the Drupal documentation.
 
 ```php
 $settings['container_yamls'][] = DRUPAL_ROOT . '/sites/development.services.yml';
 ```
 
+#### Multisite
+
+* See [DDEV-Local Drupal 8 Multisite Recipe] (<https://github.com/drud/ddev-contrib/tree/master/recipes/drupal8-multisite>) to start
```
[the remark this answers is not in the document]
laetus007: Thank you greatly for this feedback.  Documentation is one of our team's weak points so I greatly appreciate the learning opportunity.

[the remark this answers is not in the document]
mattstein: My pleasure! I’ve worked with multi-site Drupal as many times as I’ve landed on the moon, so I’m glad we could improve the docs together. 🎉

rfay:
Minor remaining trouble with markdown, 

> docs/content/users/usage/cms-settings.md:100 MD031/blanks-around-fences Fenced code blocks should be surrounded by blank lines [Context: "```"]
docs/content/users/usage/cms-settings.md:100 MD032/blanks-around-lists Lists should be surrounded by blank lines [Context: "```"]
> docs/content/users/usage/cms-settings.md:[10](https://github.com/drud/ddev/actions/runs/4288791438/jobs/7471186607#step:8:11)1 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above] [Context: "### TYPO3 Specifics"]

docs/content/users/usage/cms-settings.md
```diff
@@ -36,16 +36,68 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
+
+    3. Update your [`web_environment`](../configuration/config.md#web_environment) config option if you’re using site aliases:
+
+        ```yaml
+        web_environment:
+          # Make DDEV Drush shell PIDs last for entire life of the container
+          # so `ddev drush site:set @alias` persists for all Drush connections.
+          # https://chrisfromredfin.dev/posts/drush-use-ddev/
+          - DRUSH_SHELL_PID=PERMANENT
+        ```
```
mattstein, line 100: Adding a line break here should fix the linting error. 🤞 

```suggestion
        ```

```

rfay:
Thanks for all the effort getting this in @laetus007 !

PR #4610, opened 2023-02-04
Make nodejs installation work, changed npm --unsafe-perm behavior

rfay:
## The Issue

npm behavior with regard to `--unsafe-perm` seems to have changed, and now the various versions don't `npm install` in the same way; some need --unsafe-perm, and others need `npm config unsafe-perm`

Try both ways.

`ddev config --nodejs-version= <all>` and `ddev start` and `ddev exec gulp`

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4610"></a>


AronNovak:
I can confirm that this affected one of our client projects and in the alpha release, it does work correctly.

PR #4612, opened 2023-02-04
Switch devcontainer/github codespaces to use apt install

rfay:
## The Issue

When v1.21.5 comes out, codespaces/devcontainer can use regular apt install

Prepare to just use apt install

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4612"></a>


rfay (commented):
This can be pulled now because apt already gets v1.21.5-alpha1

PR #4613, opened 2023-02-05
project type filter for `ddev list` command (`ddev list --type=`)

cmuench:
## The Issue
As a user of many different local installed ddev projects I want to filter the `ddev list` output. Currently we have the `only-active` flag available.
Additionally we could filter by project type.

The PR restructures the parameters of the list command into a struct. This prepares the command to get additional filters in the future like tags.

Run the `ddev list` command with new filter option `-t` or `--type` and set a project type

Example: `ddev list -t php` or `ddev list -t magento2`.

The filter can be combined with active only filter. 
`ddev list -A -t php`

## Automated Testing Overview
I added a simple test which filters by a not existing type. Currently I am looking for a good way to test also the project type filter by an existing type.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4613"></a>

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4613"></a>

### DIFF
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -15,7 +15,9 @@ var ListCmd = &cobra.Command{
 	Aliases: []string{"l", "ls"},
 	Example: `ddev list
 ddev list --active-only
-ddev list -A`,
+ddev list -A
+ddev list --type=drupal8
+ddev list -t drupal8`,
 	Run: func(cmd *cobra.Command, args []string) {
 		ddevapp.List(listCommandSettings)
 	},
--- a/cmd/ddev/cmd/list.go
+++ b/cmd/ddev/cmd/list.go
@@ -5,17 +5,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// continuous, if set, makes list continuously output
-var continuous bool
-
-// activeOnly, if set, shows only running projects
-var activeOnly bool
-
-// continuousSleepTime is time to sleep between reads with --continuous
-var continuousSleepTime = 1
-
-// wrapListTable allow that the text in the table of ddev list wraps instead of cutting it to fit the terminal width
-var wrapListTableText bool
+var ListCommandSettings = ddevapp.ListCommandSettings{}
 
 // ListCmd represents the list command
 var ListCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


rfay:
What do you think @cmuench , does this have a future?

cmuench:
@rfay The filter is already working. I had a bigger idea but we can start with this approach. That's why it's currently marked as draft.

cmd/ddev/cmd/list.go
```diff
@@ -5,17 +5,7 @@ import (
...
-
-// activeOnly, if set, shows only running projects
-var activeOnly bool
-
-// continuousSleepTime is time to sleep between reads with --continuous
-var continuousSleepTime = 1
-
-// wrapListTable allow that the text in the table of ddev list wraps instead of cutting it to fit the terminal width
-var wrapListTableText bool
+var ListCommandSettings = ddevapp.ListCommandSettings{}
```
rfay, line 8: I don't think ListCommandSetings needs to be global does it?

cmd/ddev/cmd/list_test.go
```diff
@@ -103,6 +103,12 @@ func TestCmdList(t *testing.T) {
 	siteList = getTestingSitesFromList(t, jsonOut)
 	assert.Equal(len(TestSites), len(siteList))
 
+	// Now filter the list by a not existing type
```
rfay, line 106: Can we add one that uses existing type? I think there's always a wordpress type in TestCmd.

pkg/ddevapp/list.go
```diff
@@ -14,29 +14,53 @@ import (
 	"time"
 )
 
+// ListCommandSettings conains all filters and settings of the `ddev list` command
+type ListCommandSettings struct {
```
rfay, line 18: This is a significant code improvement, thanks!

rfay (changes requested):
This is looking quite good and is a major code improvement, thanks!

It's missing 
* the proper updates to docs to mention the new feature, https://ddev.readthedocs.io/en/latest/users/usage/commands/#list
* It needs to have examples add in the command so `ddev help list` will show this option (and perhaps others).

cmd/ddev/cmd/list.go
cmuench: Done

cmd/ddev/cmd/list_test.go
cmuench: I added a test but in general I have issues to run them on my local machine.

pkg/ddevapp/list.go
cmuench: I thought it's a way to get rid of all these variables for options.

cmuench: If that works then we can improve also the other commands.

rfay: I don't like to use those statics/globals at all. Mostly newer commands use the if changed stuff.

cmuench:
> This is looking quite good and is a major code improvement, thanks!
> 
> It's missing
> 
> * the proper updates to docs to mention the new feature, https://ddev.readthedocs.io/en/latest/users/usage/commands/#list
> * It needs to have examples add in the command so `ddev help list` will show this option (and perhaps others).

Done

docs/content/users/usage/commands.md
```diff
@@ -716,6 +717,9 @@ ddev list
 
 # List all running projects
 ddev list --active-only
+
+# List all wordpress
```
mattstein, line 721: Minor nitpick to clarify the product name (proper noun) and add “projects” for consistency with the comment above:

```suggestion
# List all WordPress projects
```

mattstein (approved):

docs/content/users/usage/commands.md
cmuench: Absolutly. Good finding.

cmuench:
> This is looking quite good and is a major code improvement, thanks!
> 
> It's missing
> 
> * the proper updates to docs to mention the new feature, https://ddev.readthedocs.io/en/latest/users/usage/commands/#list
> * It needs to have examples add in the command so `ddev help list` will show this option (and perhaps others).

it's done

rfay:
Yay, thanks. Back now and will catch up on PRs eventually :)

cmuench:
> Yay, thanks. Back now and will catch up on PRs eventually :)

No pressure. I was not sure if I had to inform you.

PR #4620, opened 2023-02-07
Make ddev nvm work right again, fixes #4499

### LINKED ISSUE #4499: `ddev exec` doesn't pick up new nvm-defined node version

_No response_

### Expected Behavior

Through a post-start hook I want to call yarn start, but this fails because it does not meet the required node version defined in package.json of a project. Switching the web container to use the required node version works, but yarn start does not seem pickup this and instead continues to use the `nodejs_version` set in `config.yaml`.

### Actual Behavior

I am trying to use a different version of node in the web container and setup a post-start hook for `yarn start`. The version provisioned is `14` however a project needs a lower version "10.16". I setup a post-start hook command like so:

```yaml
hooks:
  post-start:
    - exec-host: ddev nvm install 10.16.0 && ddev nvm use
    - exec: yarn start
```

exec-host is needed for the nvm commands as otherwise it fails.

This outputs:

```
v10.16.0 is already installed.
Now using node v10.16.0 (npm v6.9.0)
Found '/var/www/html/.nvmrc' with version <10.16.0>
Now using node v10.16.0 (npm v6.9.0)
yarn run v1.22.19
error example-project@1.0.0: The engine "node" is incompatible with this module. Expected version "^10.16.0". Got "14.21.1"
error Commands cannot run with an incompatible environment.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

The `node --version` command returns the original node version provisioned in config.yaml `14.21.1` when run via a hook, however if I `ddev ssh` and then run `node --version` it is set to `v10.16.0` and yarn runs fine.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #4499 

NVM regression in DDEV v1.21.4 was caused mostly by dropping the NVM_DIR environment variable in the container.

* Restore the NVM_DIR globally
* Improve the .bashrc (from /etc/skel) by using `source` instead of `.`
* Add better nvm coverage to TestNodeJSVersions

- [x] From #4499:

```
ddev nvm install 10.16.0 && ddev nvm use 10.16.0 && ddev nvm alias default 10.16.0
ddev exec node --version
```
should show 10.16.0

Manually test the things #4346 was aimed at:

#4346 was trying to fix the issues in 
 - [x] https://github.com/drud/ddev/issues/4313
- [ ] And https://github.com/drud/ddev/issues/4243
- [x] And it was trying to fix WSL2 when the uid was 999

## Automated Testing Overview

Improved TestNodeJSVersions to get coverage for this case.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4620"></a>

### DIFF
--- a/cmd/ddev/cmd/root_test.go
+++ b/cmd/ddev/cmd/root_test.go
@@ -165,8 +165,9 @@ func TestGetActiveAppRoot(t *testing.T) {
 
 	// And we should be able to stop it and find it as well
 	app, err := ddevapp.GetActiveApp("")
+	require.NoError(t, err)
 	err = app.Stop(false, true)
-	assert.NoError(err)
+	require.NoError(t, err)
 
 	t.Cleanup(func() {
 		err = os.Chdir(origDir)
--- a/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
@@ -113,8 +113,8 @@ if ! shopt -oq posix; then
 fi
 
 export NVM_DIR="$HOME/.nvm"
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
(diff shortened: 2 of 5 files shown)


rfay:
I was able to test everything but the Ubuntu Live. I have no idea how to get enough ram/disk to make that work. Always ran out of "disk space".

PR #4623, opened 2023-02-09
Mark ddev stop as deprecated command.

miromichalicka:
## The Issue

`ddev stop` command is deprecated, but you're not notified about it.

User gets a message with replacement command.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4623"></a>


rfay:
Why do you think `ddev stop` is deprecated? It's a standard part of DDEV's workflow...

rfay:
`ddev rm` is an alias for `ddev stop`; *That* is deprecated. But it's just an alias.

PR #4635, opened 2023-02-10
Provide more useful message if removing of volume fails.

miromichalicka:
## The Issue

When volume removal fails, it might be tricky for less experienced users to find the reason. This PR adds information which containers are still using the volume.

Create test project:
```
⇒ mkdir -p test-4635 && cd test-4635
⇒ ddev config --docroot=.  --project-type=php --project-name=test
⇒ ddev start
```

Confirm that only this project is using test-mariadb volume
```
⇒ docker ps --filter "volume=test-mariadb" -a
CONTAINER ID   IMAGE                                                COMMAND                  CREATED              STATUS                        PORTS                       NAMES
d586fab3af05   drud/ddev-dbserver-mariadb-10.4:v1.21.4-test-built   "/docker-entrypoint.…"   About a minute ago   Up About a minute (healthy)   127.0.0.1:32768->3306/tcp   ddev-test-db
```

Start two additional containers using that volume:
```
⇒ docker run -d --name test1 -v test-mariadb:/tmp hello-world:latest
⇒ docker run -d --name test2 -v test-mariadb:/tmp hello-world:latest
```

Confirm that all three are using test-mariadb volume now
```
⇒  docker ps --filter "volume=test-mariadb" -a                       
CONTAINER ID   IMAGE                                                COMMAND                  CREATED         STATUS                     PORTS                       NAMES
b807570dc649   hello-world:latest                                   "/hello"                 3 minutes ago   Exited (0) 3 minutes ago                               test2
391002c97206   hello-world:latest                                   "/hello"                 3 minutes ago   Exited (0) 3 minutes ago                               test1
d586fab3af05   drud/ddev-dbserver-mariadb-10.4:v1.21.4-test-built   "/docker-entrypoint.…"   5 minutes ago   Up 5 minutes (healthy)     127.0.0.1:32768->3306/tcp   ddev-test-db
```

Run `ddev delete -Oy` and get message 
```
could not remove volume test-mariadb: Docker volume 'test-mariadb' is in use by 2 containers and cannot be removed. Use 'docker rm -f test2 test1' to remove them 
```

Delete test2 `docker rm -f test2`

Run `ddev delete -Oy` and get message 
```
could not remove volume test-mariadb: Docker volume 'test-mariadb' is in use by a container 'test1' and cannot be removed. Use 'docker rm -f test1' to remove it
```

After deleting test1, you will be able to delete project without any problem

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4635"></a>


rfay (commented):
This looks good to me, but I'd love to know more about how you got this situation. Are you able to reproduce the problem? Are there other solutions we should be recommending? How about suggesting `docker rm -f $(docker ps -aq)` here?

rfay:
golangci-lint complaints:
```
Warning: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
  Warning: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
  Warning: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
```

miromichalicka:
To be honest, I have no idea how my colleagues ended up with this error. I was able to get it only when I knowingly mounted a volume to multiple containers.
The other recommended solution is `docker rm -f $(docker ps -aq)`. However, I found this approach too destructive as we can provide more targeted approach by providing specific containers to delete.

rfay:
I agree, thanks for the info.

rfay (changes requested):
Please re-test manually and provide manual testing instructions, thanks!

miromichalicka:
Steps for manual test added.

rfay (changes requested):
I appreciate the careful effort you went to in order to get great error messages, but this code adds complexity that probably doesn't have to be as detailed. Instead of separate stanzas, could we have a single check that gets the names of the containers and offers the `docker rm -f <container1> <container2>...` recipe to delete them?

miromichalicka:
Thanks for the review. I unified message for one and multiple containers. I would suggest to keep the generic message in case of error. I think I just hit a case when the volume isn't mounted to any other container but it's not possible to delete it. I'll open a followup issue and PR when that's confirmed.

rfay:
See https://discord.com/channels/664580571770388500/1068592266752573440/1078647584970772600 for demonstration that this is a valid effort!

I would love to know how we get into this situation; if you or your coworker hit it again it would be great to study it and find out the source of the problem.

It also would be valid to just recommend `docker rm -f $(docker ps -aq)` which is completely harmless as long as people are only running DDEV containers.

It does seem to me like `ddev poweroff` must be missing some non-running containers somehow.

rfay:
From the discord thread:

> It would be good to know in your case why in the world a spare container was holding the volume. If you get the chance to study it, use `docker ps -a --filter volume=<projectname>-mariadb` to find out what docker container is using the volume.

rfay:
Note that `docker rm -f $(docker ps -a --filter volume=<projectname>-mariadb)` would solve the mysql/mariadb volume issue, would be easy to get that in there.

miromichalicka:
- cool, happy that someone was able to replicate that. I'll definitely ask my colleagues to watch for the error
- `which is completely harmless as long as people are only running DDEV containers` - we're not doing that, so I would be careful to jump to conclusions here
- `docker rm -f $(docker ps -a --filter volume=<projectname>-mariadb)` - yes, in theory this should work, but we're basically doing that in the code. I like it better in the code since you can actually review containers before deleting them and keep those you knowingly started

To conclude, I think we should go with this PR as better instructions gives us some time to search for a bug in `ddev poweroff`

rfay:
I'm not saying we shouldn't go with this PR, just saying that we can fine-tune the last-ditch recommendations. 

Also, now that we have a known situation, we can work in another PR to try to make this less likely to happen.

rfay:
Here's how we get into this trouble in the first place:

* https://github.com/drud/ddev/issues/4676

rfay (approved):
Thanks so much for simplifying this. It's hugely useful, and as the discord thread points out, other people are having this problem. 

Would you be interested in taking on the source of the problem, described in https://github.com/drud/ddev/issues/4676 (as a separate PR)? I suspect that a silent Pull() is happening to people early in `ddev start` and without appropriate output, so people do a <ctrl-c>. I'd be happy to do a call with you to discuss next steps if you're interested in taking it on. You're mighty ambitious, and mighty capable, and very much appreciated.

rfay:
I got a little time and am working on #4676 myself.

PR #4636, opened 2023-02-10
Fix line break in `ddev exec` help, fixes #4634

### LINKED ISSUE #4634: Improve ddev-exec help/examples output

_No response_

### Expected Behavior

Either it should be a real newline/linebreak; or maybe the latter -s example isn't needed as all, as it's the same as --service db; and the -s short-version is already documented.

### Actual Behavior

Current code: https://github.com/drud/ddev/blob/a9b53ba02d066e9879f396e57c15d28857976cfc/cmd/ddev/cmd/exec.go#L23 
It prints \n as the literal characters rather than a line-break between the 2 command examples.

### Steps To Reproduce

Run ddev exec --help

### Anything else?

image

miromichalicka:
## The Issue

https://github.com/drud/ddev/issues/4634

Add hard line break

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4636"></a>


rfay (approved):
Thanks!

PR #4640, opened 2023-02-10
Add global router_http(s)_port config, fixes #3240

### LINKED ISSUE #3240: Add global router_http(s)_port setting (can be overridden by project)
Imho it absolutely would make sense to change the router ports on the machine instead of the project. It looks like this is missing currently see https://ddev.readthedocs.io/en/stable/users/extend/config_yaml/#global_configyaml-options, maybe just the docs leaking here.

Add the options router_http(s)_port to the global configuration and apply it to every project which is started.

Using the project's configuration which already works.

**Additional context**

miromichalicka:
## The Issue

* #3240

- [x] Set router_https_port globally
- [x] Override it with project
- [x] Make sure things work with it not set in either place
- [x] Make sure upgrade from v1.21.6 to HEAD works OK

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4640"></a>

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4640"></a>


miromichalicka:
Next steps:
- add/fix tests
- add validation of ports
- update docs

rfay:
What do you think @miromichalicka , does this have a future?

rfay (changes requested):
Thanks for this! It looks like it's on the right track.

* The config value in global should be same as in project, `router_http_port` and `router_https_port`
* It doesn't have support for the global command yet, `ddev config global --router-http-port=8080 --router-https-port=8443`
* I like that the project config overrides the global; that's the way it should be.
* The comments in global_config.yaml should explain the usage as they do for all the other things.
* Right now this isn't functional (please manually test) because router_http_port and router_https_port are always populated in the project .ddev/config.yaml, so those defaults always override. You'll want those to be empty I think by default.
* But it doesn't work correctly even if I remove the project-level configs of router_http_port, etc.

Please make sure to manually test, thanks!

pkg/globalconfig/global_config.go
```diff
@@ -63,6 +63,8 @@ type GlobalConfig struct {
 	UseTraefik                   bool                    `yaml:"use_traefik"`
 	WSL2NoWindowsHostsMgt        bool                    `yaml:"wsl2_no_windows_hosts_mgt"`
 	ProjectList                  map[string]*ProjectInfo `yaml:"project_info"`
+	RouterHTTPPort               string                  `yaml:"router_http_port"`
```
rfay, line 66: This should go *above* the project_info member, otherwise they're hidden below that.

rfay:
Rebased. Looking forward to this.

miromichalicka:
I did some manual testing and now I'm able to set up global ports with `ddev config global` command. After manually removing values from project config, I have project running on global values. I'll probably continue tomorrow with more testing. Some things ae not clear to me:

> The config value in global should be same as in project, router_http_port and router_https_port
- do you mean this https://github.com/ddev/ddev/pull/4640/files#diff-42da3ac5f97f8c2d00f51c6afe831dadaf9245cc93f258cee0e6f1a6ac2e7a06R65? Because it seems the same to me as https://github.com/ddev/ddev/pull/4640/files#diff-de9be5de8b21f6465a7e0df5c9c2f44638c0a67659ccf812cd043bfee94aeeb1R85

> It doesn't have support for the global command yet, ddev config global --router-http-port=8080 --router-https-port=8443
- I used the same flags as in the project, so the correct command should be `ddev config global --http-port=8080 --https-port=8443`. Should I change that?

> router_http_port and router_https_port are always populated in the project .ddev/config.yaml, so those defaults always override. You'll want those to be empty I think by default.
- I'm not sure I want to touch existing configs. I'll check how to modify this behaviour for new projects. Maybe that's already done with recent changes.

miromichalicka:
> I think it should be the same, router-http-port and router-https-port

 `router_http_port` and `router_https_port` are used in YAML. In CLI it seems to be http-port and https-port - https://github.com/ddev/ddev/blob/master/cmd/ddev/cmd/config.go#L244-L245, or do you have something else in mind?

> I think if we change the defaults for the project-level values to be ,omitempty in

I already have that, I'll test a new project creation tomorrow.

rfay:
Wow, let's make them the same as the yaml. Maybe keep aliases in the project config so as to not break existing?

rfay:
Yay! I'm out for a couple of weeks but look forward to getting this in.

mattstein (approved):

rfay (approved):
I think this is good now, thanks for pioneering it @miromichalicka ! Please take a look at my recent commits and the overall state and see if you like it.

rfay:
I tested all the things. 

1. If people have existing projects, the defaults are there in the .ddev/config.yaml, and will override the global values. That will likely result in some confusion.
2. `ddev config --router-http-port=""` doesn't overwrite :(

hanoii:
Awesome!!

PR #4643, opened 2023-02-12
Add more detail for Backdrop Quickstart

laryn:
I was searching for the Backdrop Quickstart and it took me a while to find it buried inside the Drupal Quickstart. Then once I found it, it didn't have the information I was looking for (how to enable `bee`). Luckily I found that information elsewhere! 

Here's a PR to break Backdrop Quickstart out, and include the information about how to enable `bee`.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4643"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -2,6 +2,21 @@
 
 While the generic `php` project type is [ready to go](./project.md) with any CMS or framework, DDEV offers project types for more easily working with popular platforms and content management systems:
 
+=== "Backdrop"
+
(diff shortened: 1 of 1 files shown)


rfay:
Yay, a Backdrop fan! 

It used to have its own tab, but there haven't been many backdrop fans around lately and so it got consolidated just recently.

laryn:
Three cheers for Backdrop!

rfay (changes requested):
We're not going to move Backdrop out of the Drupal section at this time, as we made that choice recently. I'm not sure that adding information about Bee here is something that belongs in the DDEV docs?

PR #4646, opened 2023-02-13
Add .ddev/.downloads to mutagen exclusions, fixes #4645

### LINKED ISSUE #4645: `.ddev/mutagen/mutagen.yml` need to exclude `.ddev/.downloads`

_No response_

### Expected Behavior

mutagen syncing should happen pretty fast

### Actual Behavior

Lots of dots

### Steps To Reproduce

Put a bunch of files in .ddev/.downloads, like from a failed pull

### Anything else?

Need to exclude that directory in mutagen.yml

rfay:
## The Issue

Big tarballs, or unmanaged files in .ddev/.downloads can cause extraordinary starup pain with mutagen enabled. 

* Add .ddev/.downloads to mutagen exclusion
* Add docs about figuring what's wrong in this situation (infinite startup time when mutagen enabled)

### DIFF
--- a/docs/content/users/install/performance.md
+++ b/docs/content/users/install/performance.md
@@ -35,7 +35,7 @@ Mutagen can offer a big performance boost on macOS and Windows. It’s fast and 
     ### Enabling and Disabling Mutagen
 
     !!!warning "Don’t Install Mutagen"
-        Don’t install the `mutagen` binary separately. If it’s not available, DDEV will install and upgrade it for you.
+        You do not need to install anything to use mutagen. DDEV installs and maintains its own `mutagen` binary.
 
     We recommend enabling Mutagen globally with `ddev config global --mutagen-enabled`. You can disable it again with `ddev mutagen reset && ddev config global --mutagen-enabled=false`.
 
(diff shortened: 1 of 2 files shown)


docs/content/users/install/performance.md
```diff
@@ -137,6 +137,7 @@ Mutagen can offer a big performance boost on macOS and Windows. It’s fast and
 
     ### Troubleshooting Mutagen Sync Issues
 
+    * Does `ddev start` take forever, with perhaps many minutes of "dots"? Take care that you don't have big binaries that mutagen is trying to sync, which will cause endless "dots" as your project is starting. There is already an exclusion place for database tarballs and the like, the `.tarballs` directory; if you put dumps there they will be ignored by mutagen. To find out what mutagen is trying to sync, use `ddev mutagen status -l` in another window.
```
mattstein, line 140: Should this be first? It seems like the existing items are probably more urgent to convey sooner than this.

Rewrite to start with a recommendation instead of a question, and try to be more brief:

```suggestion
    * Avoid having Mutagen sync large binaries, which can cause `ddev` start to take a long time. The `.tarballs` directory is automatically excluded, so Mutagen will ignore anything you move there. To see what it’s trying to sync, run `ddev mutagen status -l` in another window.
```

mattstein (changes requested):

PR #4647, opened 2023-02-13
Add file as alias for "src" flag (fixes #4593)

### LINKED ISSUE #4593: Add "file" as alias for import-db flag "src"

Export-db uses `--file` to specify the file being exported. Import-db uses `--src` to identify the _file_ being imported. (In fact, the shortcut for `--src` is `-f`!) It would be nice to not have to remember which flag goes with which operation, and it would be especially nice not to have the command simply fail when you accidentally use the wrong one.

The safest way to do this is probably to add `--file` as an alias, rather than renaming it.

### Describe your solution

Add `--file` as an alias for `--src` for the import-db command.

### Describe alternatives

You could probably rename `--src` to `--file` for consistency, but that would probably make a lot of people sad who are used to the way things are now. I think we can avoid that with an alias.

### Additional context

_No response_

miromichalicka:
## The Issue
https://github.com/drud/ddev/issues/4593

It's using https://github.com/spf13/pflag#mutating-or-normalizing-flag-names approach to set alias for existing flag.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4647"></a>

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4647"></a>

### DIFF
--- a/cmd/ddev/cmd/export-db.go
+++ b/cmd/ddev/cmd/export-db.go
@@ -5,6 +5,7 @@ import (
 	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 var outFileName string
@@ -64,10 +65,18 @@ ddev export-db someproject --gzip=false --file=/tmp/someproject.sql `,
 }
 
 func init() {
-	ExportDBCmd.Flags().StringVarP(&outFileName, "file", "f", "", "Provide the path to output the dump")
+	ExportDBCmd.Flags().StringVarP(&outFileName, "src", "f", "", "Provide the path to output the dump")
 	ExportDBCmd.Flags().BoolVarP(&doGzip, "gzip", "z", true, "Use gzip compression")
 	ExportDBCmd.Flags().BoolVarP(&doXz, "xz", "", false, "Use xz compression")
 	ExportDBCmd.Flags().BoolVarP(&doBzip2, "bzip2", "", false, "Use bzip2 compression")
 	ExportDBCmd.Flags().StringVarP(&exportTargetDB, "target-db", "d", "db", "If provided, target-db is alternate database to export")
+	ImportFileCmd.Flags().SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
+		switch name {
+		case "src":
+			name = "file"
+			break
+		}
+		return pflag.NormalizedName(name)
+	})
 	RootCmd.AddCommand(ExportDBCmd)
 }
--- a/cmd/ddev/cmd/import-db.go
+++ b/cmd/ddev/cmd/import-db.go
@@ -68,15 +68,15 @@ gzip -dc db.sql.gz | ddev import-db`,
 }
 
 func init() {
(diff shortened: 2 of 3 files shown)


rfay:
On holiday/vacation but will review as soon as I have the spare brainpower :)

rfay (changes requested):
I'd say as it stands it's not good enough, because you didn't add an alias, you replaced "src", with "file"; we don't want to *change* things, just add an alias right? 

```
rfay@rfay-tag1-m1:~/workspace/d9$ ddev import-db -h
Import a sql file into the project.
The database dump file can be provided as a SQL dump in a .sql, .sql.gz, sql.bz2, sql.xz, .mysql, .mysql.gz, .zip, .tgz, or .tar.gz
format. For the zip and tar formats, the path to a .sql file within the archive
can be provided if it is not located at the top level of the archive. An optional target database
can also be provided; the default is the default database named "db".
Also note the related "ddev mysql" command

Usage:
  ddev import-db [project] [flags]

Examples:
ddev import-db
ddev import-db --src=.tarballs/junk.sql
ddev import-db --src=.tarballs/junk.sql.gz
ddev import-db --target-db=newdb --src=.tarballs/db.sql.gz
ddev import-db --src=.tarballs/db.sql.bz2
ddev import-db --src=.tarballs/db.sql.xz
ddev import-db <db.sql
ddev import-db someproject <db.sql
gzip -dc db.sql.gz | ddev import-db

Flags:
      --extract-path string   If provided asset is an archive, provide the path to extract within the archive.
  -f, --file string           Provide the path to a sql dump in .sql or tar/tar.gz/tgz/zip format
  -h, --help                  help for import-db
      --no-drop               Set if you do NOT want to drop the db before importing
  -p, --progress              Display a progress bar during import (default true)
  -d, --target-db string      If provided, target-db is alternate database to import into (default "db")

Global Flags:
```

miromichalicka:
I pushed a change which will keep the original flags there. However, I wasn't able to figure out so far, how to print aliases in the help message.
The other solution would be to add also alias as a standalone flag. That way it would get printed out. However, I'm not sure how it would behave if you fill both flags and they try to access the same variable.
If you have any other suggestions, I'm happy to check them.

rfay:
I don't mind trying a duplicate flag; the help can just say "alias for --src" and what it does? Thanks for your work on this!

rfay (changes requested):
This is looking fine to me; manually tested, it's great. But I think there should be a hint in a couple of places about the `--file` alias. 
1. Add an example that has it in each of them
2. Add to the `--src` description "--file can be used an alias of --src" ?

rfay:
BTW, if you'll PM me your email address I'll add you to the buildkite team so you can restart buildkite tests, etc.

miromichalicka:
Planning to have a look during this weekend, been sick for the past two weeks.

rfay:
When you're well, please update the requested item and hopefully I'll have some time for regular maintenance now. Thanks!

rfay:
What do you think about the simpler approach of just adding a new flag that does the exact same thing? Its description could be "same as --src" for example?

So you'd just add 

```
	ImportDBCmd.Flags().StringVarP(&dbSource, "file", "", "", "Provide the path to a sql dump in .sql or tar/tar.gz/tgz/zip format (same as --src)")
```

rfay:
Closing in favor of 
* #5013
We'll try to get you commit credit on that one.

PR #4656, opened 2023-02-15
Fix `ddev import-db` and `ddev import-files` queries for path to artifacts

gilbertsoft:
## The Issue

In the two queries for a full path, "pull" is used instead of "full".

"pull" is replaced by "full".

Run the two commands `import-db` and `import-files` to see if the query for the path is fine.

### DIFF
--- a/cmd/ddev/cmd/import-files.go
+++ b/cmd/ddev/cmd/import-files.go
@@ -86,7 +86,7 @@ func promptForFileSource(val *string) {
 	// An empty string isn't acceptable here, keep
 	// prompting until something is entered
 	for {
-		fmt.Print("Pull path: ")
+		fmt.Print("Full path: ")
 		*val = util.GetInput("")
 		if len(*val) > 0 {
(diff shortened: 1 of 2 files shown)


gilbertsoft:
Thanks to @sbusemann for this finding!

rfay (changes requested):
I don't believe that this is correct. I think it was intended to be "Pull path", but that was poor. It does *not* need to be "Full path", as a relative path is fine. Instead, we should change the prompt to "Path to import" or something like that.

rfay:
I'm not even convinced that this fmt.Print() is useful. The path has already been requested....

rfay:
@gilbertsoft please take a look at https://github.com/drud/ddev/pull/4656/commits/0871166f126669afc30c204697ad6c4a2af37a95 and see if it works for you.

rfay (approved):
I think this is right now, hope for @gilbertsoft approval.

PR #4657, opened 2023-02-15
[docs] Fix copy pasta on webserver_type description

chellman:
The webserver_type description here:

https://ddev.readthedocs.io/en/stable/users/configuration/config/#webserver_type

Was the same as xdebug_enabled here:

https://ddev.readthedocs.io/en/stable/users/configuration/config/#xdebug_enabled

This PR fixes that.

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Which available [web server type](../extend/customization-extendibility.md/#changing-web-server-type) should be used.
+Which available [web server type](../../extend/customization-extendibility.md/#changing-web-server-type) should be used.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Which available web server should be used.
+Which available [web server type](../extend/customization-extendibility.md/#changing-web-server-type) should be used.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Whether Xdebug should be enabled for [step debugging](../debugging-profiling/step-debugging.md) or [profiling](../debugging-profiling/xdebug-profiling.md).
+Which available web server should be used.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Which available [web server type](../../extend/customization-extendibility.md/#changing-web-server-type) should be used.
+Which available [web server type](../extend/customization-extendibility.md#changing-web-server-type) should be used.
 
 | Type | Default | Usage
 | -- | -- | --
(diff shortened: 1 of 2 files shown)


docs/content/users/usage/developer-tools.md
```diff
@@ -113,7 +113,7 @@ On Drupal 8+, if you want to use `drush uli` on the host (or other Drush command
 
 ## DDEV and Terminus
 
-[Terminus](https://pantheon.io/docs/guides/terminus) is a command line tool providing advanced interaction with [Pantheon](https://pantheon.io/) services. `terminus` is available inside the project’s container, allowing users to get information from, or issue commands to their Pantheon-hosted sites. This is an especially helpful feature for Windows users since Terminus is only officially supported on Unix-based systems.
+[Terminus](https://pantheon.io/docs/terminus) is a command line tool providing advanced interaction with [Pantheon](https://pantheon.io/) services. `terminus` is available inside the project’s container, allowing users to get information from, or issue commands to their Pantheon-hosted sites. This is an especially helpful feature for Windows users since Terminus is only officially supported on Unix-based systems.
```
rfay, line 116: Interesting, this was recently changed from the docs/terminus URL when they started failing on it...

docs/content/users/configuration/config.md
```diff
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Whether Xdebug should be enabled for [step debugging](../debugging-profiling/step-debugging.md) or [profiling](../debugging-profiling/xdebug-profiling.md).
+Which available web server should be used.
```
rfay, line 565:
```suggestion
Which available [web server type](../extend/customization-extendibility.md/#changing-web-server-type
) should be used (`nginx-fpm` (default) or `apache-fpm`).
```

rfay (changes requested):
Thanks for catching and fixing. Just extended the text a bit. You can accept/commit if you want.

docs/content/users/configuration/config.md
mattstein: @rfay The link seems useful, but the parenthetical is redundant with the details that immediately follow:

Rendered output highlighting redundant bits

rfay: That's the problem with tiny changes, can't see the context!

docs/content/users/usage/developer-tools.md
chellman: Wait a minute, I think it just changed again.  Now it's redirecting to https://docs.pantheon.io/terminus.  Updated.

docs/content/users/configuration/config.md
chellman: Thanks to both of you.  I committed with the link, but without the repetition.

docs/content/users/usage/developer-tools.md
mattstein: What a squirrelly link!

rfay: I do think we should be able to convince the link checker to follow redirects.

mattstein: Direct links are best for the reader and redirects risk going stale or turning into long chains. I think flagging redirects and keeping links current is best, even with the inconvenience of outages and weird cases like this.

rfay (changes requested):
The link I suggested wasn't right (404). If you could figure out the correct link it would be appreciated. See https://ddev--4657.org.readthedocs.build/en/4657/users/configuration/config/#webserver_type

The link should actually end being https://ddev--4657.org.readthedocs.build/en/latest/users/extend/customization-extendibility/#changing-web-server-type

but it actually goes to https://ddev--4657.org.readthedocs.build/en/4657/users/configuration/extend/customization-extendibility.md/#changing-web-server-type

right now.

Looks like it just has extra "customization" component that doesn't belong.

docs/content/users/configuration/config.md
```diff
@@ -562,7 +562,7 @@ Example: `webimage_extra_packages: [php-yaml, php-bcmath]` will add the `php-yam
 
 ## `webserver_type`
 
-Whether Xdebug should be enabled for [step debugging](../debugging-profiling/step-debugging.md) or [profiling](../debugging-profiling/xdebug-profiling.md).
+Which available [web server type](../extend/customization-extendibility.md/#changing-web-server-type) should be used.
```
rfay, line 565: Maybe this is it this time :)

```suggestion
Which available [web server type](../../extend/customization-extendibility.md/#changing-web-server-type) should be used.
```

PR #4661, opened 2023-02-17
[docs] pantheon.yaml.example comments: changed sequence of items #6 and #7

ops-andy:
## The Issue
At least in my experience you can't run `ddev composer require drush/drush` before having a project configured. You need to run `ddev config` first, then run `ddev...` commands.


rfay (approved):
Makes sense, thanks!

PR #4675, opened 2023-02-24
[docs] Re-enable copy and edit icons

rfay:
## The Issue

Upgrading to mkdocs-material 9 lost edit and copy icons that are so useful.

* https://github.com/squidfunk/mkdocs-material/issues/5092#issuecomment-1443927417

Enable them explicitly

## Reviewing

See https://ddev--4675.org.readthedocs.build/en/4675/ to review this.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -26,6 +26,8 @@ theme:
     primary: indigo
     accent: indigo
   features:
+  - content.action.edit
+  - content.code.copy
 #  - content.tabs.link
   - navigation.expand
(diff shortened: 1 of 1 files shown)


rfay:
This has a problem... the edit icon is empty, I guess we need to add one. 

image

rfay:
I guess that's the new icon. Not very apparent to me.

rfay:
I'm going to go ahead and pull this because people using "latest" docs need these features, but hopefully we can get a more recognizable icon.

Issue #1919, opened 2019-11-05
'ddev config' should default to existing config values unless overridden with flags


wizonesolutions:
**Describe the bug**
Some complex projects, like multisite Drupal projects, use `type: php` in order to manage settings themselves. However, ddev detects them at `drupal7` or `drupal8` every time it is run. This includes when it is run non-interactively! Thus, developers can get blindsided and have ddev files or file content added to their hand-crafted settings files if they aren't very careful.

**To Reproduce**
1. Run `ddev config` with a Drupal 8 project. Select `php` as the project type.
2. Run `ddev config` again. Notice how it defaults to 'drupal8'.
3. Run `ddev config --use-dns-when-possible=false`. Notice how it **automatically generates Drupal 8 settings and changes the project type without asking**.

**Expected behavior**
The only time project type should be auto-detected is when `.ddev/config.yaml` does not exist at all, and perhaps even when no `.ddev/config.*.ya?ml` files exist.

**Version and configuration information (please complete the following information):**
 - Host computer OS and Version: Manjaro Linux
 - Docker version information (use `docker version`) and insert it here.
```
Client:
 Version:           19.03.4-ce
 API version:       1.40
 Go version:        go1.13.1
 Git commit:        9013bf583a
 Built:             Sat Oct 19 04:40:07 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.4-ce
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.1
  Git commit:       9013bf583a
  Built:            Sat Oct 19 04:39:38 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.3.0.m
  GitCommit:        d50db0a42053864a270f648048f9a8b4f24eced3.m
 runc:
  Version:          1.0.0-rc9
  GitCommit:        d736ef14f0288d6993a1845745d6756cfc9ddd5a
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

 - ddev version information (use `ddev version`)
```
DDEV-Local version	v1.11.2                        
commit            	v1.11.2                        
db                	drud/ddev-dbserver:v1.11.0-10.2
dba               	drud/phpmyadmin:v1.11.0        
ddev-ssh-agent    	drud/ddev-ssh-agent:v1.11.0    
docker            	19.03.4-ce                     
docker-compose    	1.24.1                         
os                	linux                          
[...]

Issue #443, opened 2017-08-24
Mention homebrew as an installation option in readme


cyberswat:
## What happened (or feature request):
I'm walking through the Readme and noticed that the installation instructions do not mention Homebrew.

## What you expected to happen:
The readme should be updated to include mention of homebrew as an installation option.

cyberswat:
Just noting that there doesn't seem to be a mention of homebrew in the documentation site either.

rfay:
Duplicate of #434, right?

rickmanelius:
Yup, duplicate.

PR #4677, opened 2023-02-25
Don't override project type php by detected app, fixes #4403

### LINKED ISSUE #4403: Any use of `ddev config` automatically changes a project type from 'php' to the discovered codebase type

(Skipped the diagnostics for now because this looks like intended but inconvenient behavior than a bug, but I'll add it if you think it is required, also I asked some colleagues if they can reproduce so I assume this is unlikely an issue related to my system) 

### Description

I am working mostly with WordPress project but in contrast to the suggested installation these applications are entirely build with Composer, for example by using a plugin like [WP Starter](https://github.com/wecodemore/wpstarter). So I do want to have the project type set to `php` because the `wp-config.php` is customized.

Now whenever I run a `ddev config` command, for example to change the PHP version, DDEV automatically sets the project type to `wordpress` unless I explicitly add the command parameter.

### Expected Behavior

I would expect `ddev config` to change the project type only if I explicitly add it as parameter.

### Actual Behavior

`ddev config` changes the project type to `wordpress` even if `--project-type` is not mentioned.

### Steps To Reproduce

Run the following commands in an empty (temporary) directory:

Create a new, empty DDEV project:

    ddev config --project-name project-type-issue --project-type php --php-version 8.0 --docroot public --create-docroot

Download WordPress core to the docroot:

    ddev exec wp core download --path=/var/www/html/public

Run DDEV config to update PHP version:

    ddev config --php-version 8.1

Now project-type has changed to `wordpress` in `.ddev/config.yaml`

Cleanup everything:

    ddev delete -O

### Anything else?

_No response_

dnaber-de:
## The Issue

* #4403 
* DDEV's config command will override project type "php" if any app like WordPress or Drupal is detected.

As suggested in #4403 it applies the detected app type only if there is no type specified yet.

–

## Automated Testing Overview

`TestConfigMaintainProjectTypePhp` creates a common DDEV project with type "php" and creates a `{DOC_ROOT}/wp-settings.php` file to mimic a WordPress installation. Then it runs a `ddev config` command to change an arbitrary value (the PHP version in this case) and finally verifies that the project type is still set to "php".

## Related Issue Link(s)

* #1919
* #4403

## Release/Deployment Notes

–

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -443,7 +443,7 @@ func handleMainConfigArgs(cmd *cobra.Command, args []string, app *ddevapp.DdevAp
 	}
 
 	switch {
-	case (app.Type == "" || app.Type == nodeps.AppTypePHP) && (projectTypeArg == "" || projectTypeArg == detectedApptype): // Found an app, matches passed-in or no apptype passed
+	case (app.Type == "") && (projectTypeArg == "" || projectTypeArg == detectedApptype): // Found an app, matches passed-in or no apptype passed
 		projectTypeArg = detectedApptype
 		app.Type = projectTypeArg
 		if app.Type == nodeps.AppTypePHP {
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -615,3 +615,51 @@ func TestConfigGitignore(t *testing.T) {
 	_, err = exec.RunHostCommand("bash", "-c", "git status | grep 'Untracked files'")
 	assert.Error(err, "Untracked files were found where we didn't expect them: %s", statusOut)
 }
+
+func TestConfigMaintainProjectTypePhp(t *testing.T) {
+	assert := asrt.New(t)
+
+	origDir, _ := os.Getwd()
+	tmpDir := testcommon.CreateTmpDir(t.Name())
+	err := os.Chdir(tmpDir)
+	assert.NoError(err)
+
+	t.Cleanup(func() {
+		err = os.Chdir(origDir)
+		assert.NoError(err)
+		out, err := exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
+		assert.NoError(err, "output=%s", out)
+		err = os.RemoveAll(tmpDir)
+		assert.NoError(err)
+	})
+
+	docRoot := "public"
+	args := []string{
+		"config",
+		"--project-name", t.Name(),
+		"--docroot", docRoot,
+		"--create-docroot",
+		"--project-type", "php",
+	}
+
+	// set up a DDEV project
+	out, err := exec.RunHostCommand(DdevBin, args...)
+	assert.NoError(err, "error running ddev %v: %v, output=%s", args, err, out)
+
+	// mimic a WordPress app by creating wp-settings.php file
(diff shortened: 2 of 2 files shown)


dnaber-de:
I'm sorry I was a bit too optimistic here. Since `app.Type` is set to "php" by default, we cannot distinct between "php" as the default value or the value that was read from config.yaml in `handleMainConfigArgs()`. By just removing the `app.Type == "php"` condition from this statement we will circumvent the app detection entirely.

```golang
case (app.Type == "" || app.Type == nodeps.AppTypePHP) && (projectTypeArg == "" || projectTypeArg == detectedApptype):
```

I think what still should solve the mentioned issues and pass the existing tests would be the distinction between "php" being the default value or the one from the config.yaml. This could either be implemented in `handleMainConfigArgs()` or by introducing another parameter `app.IsDefaultType` which would be set in `DdevApp.readConfig()` for example.

rfay:
Thanks for taking the time to explore this and create a PR!

Yes, there have been a variety of problems with this ambiguity over the years. The project type may be default (nothing detected) or it may be from the config.yaml, or it may be from the command line. And all of them have to work... It's complex territory.

Can you speak to the problem you specifically are trying to solve? Why does this give you trouble?

(Most people use `disable_settings_management: true` in the case where they don't want DDEV managing settings files)

dnaber-de:
The issue I have is that I do want to use the project type "php" when working with a customized WordPress installation but every time I run `ddev config ...` to change a value that is not "project-type" it switches the project type to "wordpress" (described more detailed in [#443](https://github.com/drud/ddev/issues/4403) and also for Drupal in [#1919](https://github.com/drud/ddev/issues/1919).

I do want to use `ddev config` over editing config.yaml manually to keep this file up to date (inline documentation, new config keys and so on). I'm not sure if `disable_settings_management: true` would circumvent this issue. But in general I'd like to have DDEV managing configuration files and only use the command line API.

Maybe another config parameter `disable_project_type_detection` would be the most straightforward solution to maintain backward compatibility without increasing the complexity even more?

rfay:
Thanks, @dnaber-de - But why do you care about whether it's project type 'wordpress' or 'php' if settings management is disabled? What is the difference then? (Docs: https://ddev.readthedocs.io/en/latest/users/usage/cms-settings/)

dnaber-de:
I don't care much about the value of the project-type beside that it might be a bit confusing when a different value does not imply any different behavior. The configuration parameter `disable_settings_management` is new to me and if it only affects the CMS settings file management I'm fine with it. However, this PR originated from our short discussion in #4403 where you suggested to try the change this PR provides.

I'm fairly new to Go but if you want to move that forward (in whatever direction) I'd be happy to help. Otherwise just close the PR. I'll add a hint to the tickets how to avoid it using `disable_settings_management`.

rfay:
I didn't understand the very minor reference to #4403, thanks for bringing that up. I'd love to see a solution if you can figure it out. It looked to me like this came out of nowhere, and you were just trying to solve your own problem, but it sounds like that's easier.

dnaber-de:
Sure, I'll give it another try.

rfay:
Hoping you can make it work the way we all want it to! Otherwise let me know and we'll close this one for now.

rfay:
I really appreciate you stepping up with this contribution, but no response for a month and a half, so closing for now. Happy to continue the conversation here or elsewhere, and happy to reopen this when you're ready.

PR #4680, opened 2023-02-26
Make sure all containers get cleaned up on stop and poweroff, fixes #4676

### LINKED ISSUE #4676: Chowning volume contents can result in dead containers that don't get removed

_No response_

### Expected Behavior

* Dead containers should not be left around
* `ddev stop` should remove any related dead containers
* `ddev poweroff` should remove any related dead containers

### Actual Behavior

@yoannisj discovered trouble trying to `ddev delete` where they couldn't delete the database due to other containers holding the docker volume.

See https://discord.com/channels/664580571770388500/1068592266752573440/1078647584970772600

`docker ps -a --filter volume=<projectname>-mariadb` showed:

```
CONTAINER ID   IMAGE                         COMMAND                  CREATED        STATUS                    PORTS     NAMES
b15e2d864da4   drud/ddev-webserver:v1.21.4   "sh -c 'chown -R 501…"   5 hours ago    Exited (0) 5 hours ago              cool_liskov
4e07e105b64a   drud/ddev-webserver:v1.20.0   "sh -c 'chown -R 501…"   5 months ago   Exited (0) 5 months ago             stupefied_bassi
```

So it's obvious that there were two separate instances of containers that were run and exited and weren't cleaned up, and thus made the mariadb volume non-deletable. 

The behavior that created the problem dead containers is at 

https://github.com/drud/ddev/blob/d753cc54922508474e87bc168d3902ef8a3348c4/pkg/ddevapp/ddevapp.go#L1095-L1098

But the question would be what situation caused RunSimpleContainer to *not* successfully clean up.

### Steps To Reproduce

It may be possible to <ctrl-C> during start and get this situation.

### Anything else?

The best solution is where RunSimpleContainer always successfully kills the container when it's done. Maybe a defer() with some checks to make sure it happens can accomplish this.

But while we're at it, we can give a label and a good descriptive name to the container we create; if the label is standard label be should be able to make sure it's killed on stop (and thus on poweroff).

Related:
* https://github.com/drud/ddev/pull/4635 - which tries to give a better message to solve the problem when it happens.

rfay:
## The Issue

* #4676
* https://github.com/drud/ddev/pull/4635
* GetExistingDBType() used heavyweight image to do its work, use busybox instead; this should also make disruption in `ddev config` on first pull be nicer.
* app.PullContainerImages() moved to earlier in Start() so that it behaves more predictably.

- [x] Use debugger to make sure #4676 doesn't happen any more
- [x] Try `ddev config` with images not downloaded, verify that download time is slight.
- [x] Create containers like  a failure would; `docker run --label "com.ddev.site-name:xxx" busybox true` and then `ddev stop` (where name matches) or `ddev poweroff` (where name doesn't match project)
- [x] Verify that unlabeled containers do *not* get deleted by these things.

## Automated Testing Overview

No new tests are added

### DIFF
--- a/cmd/ddev/cmd/debug-dockercheck.go
+++ b/cmd/ddev/cmd/debug-dockercheck.go
@@ -72,14 +72,14 @@ var DebugDockercheckCmd = &cobra.Command{
 		}
 
 		uid, _, _ := util.GetContainerUIDGid()
-		_, _, err = dockerutil.RunSimpleContainer(versionconstants.GetWebImage(), "", []string{"ls", "/mnt/ddev-global-cache"}, []string{}, []string{}, []string{"ddev-global-cache" + ":/mnt/ddev-globa-cache"}, uid, true, false, nil)
+		_, _, err = dockerutil.RunSimpleContainer(versionconstants.GetWebImage(), "dockercheck-runcontainer--"+util.RandString(6), []string{"ls", "/mnt/ddev-global-cache"}, []string{}, []string{}, []string{"ddev-global-cache" + ":/mnt/ddev-globa-cache"}, uid, true, false, map[string]string{"com.ddev.site-name": ""})
 		if err != nil {
 			util.Warning("Unable to run simple container: %v", err)
 		} else {
 			util.Success("Able to run simple container that mounts a volume.")
 		}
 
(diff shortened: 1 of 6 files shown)


rfay:
@miromichalicka I'd love to have you review and test this. 

Assets for testing are in https://github.com/drud/ddev/pull/4680#issuecomment-1445393401

miromichalicka:
Cool, I'll try to test it tonight.

rfay:
Manual testing:

* `ddev config` with brand-new docker works much better, just pulls busybox.
*

PR #4691, opened 2023-03-02
Add dev libraries / function test requirements

cosmicdreams:
## The Issue
Quickstart needs to be updated if we want users to know how to run Functional / FunctionalJavascript tests.  We could do it this way or we could add this knowledge to the phpunit portion of the phpstorm documentation.  

Kind of feels like there should be a phpunit section instead of a phpunit in phpstorm section.

Adds documentation that I needed in order to understand how to debug functional tests in Drupal 10


rfay:
A quickstart is a place for super-start get-going. Not for how to run functional tests. But I know you're getting some improvements into the Drupal Specifics section, which is where this info goes.

Thanks so much for exploring all this territory and contributing!

PR #4692, opened 2023-03-02
improve tableplus connection view

riconeitzel:
with latest tableplus version a connection label and color will be shown prominently in the header bar. By default it is gray and shows "PRODUCTION" for these ddev connections.

This change labels them as "DDEV" and sets the color to ddev's key color so you don't get confused whether it's a PROD connection or a DDEV connection.

See example: ddev tableplus

Reference: https://twitter.com/TablePlus/status/1631300835530665985?s=20

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4692"></a>

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
@@ -14,7 +14,7 @@ driver=mysql
 if [[ $dbtype == "postgres" ]]; then
     driver=$dbtype
 fi
-query="${driver}://db:db@127.0.0.1:${DDEV_HOST_DB_PORT}/db"
+query="${driver}://db:db@127.0.0.1:${DDEV_HOST_DB_PORT}/db?statusColor=00325a&enviroment=DDEV"
 
 set -x
 open "$query" -a "/Applications/TablePlus.app/Contents/MacOS/TablePlus"


rfay:
Thanks!

rfay (changes requested):
I'm glad this works for you, but maybe it doesn't work in the trial version?

image

It think it may be worth considering removal of the `set -x` at this point too? But it's pretty useful right now for studying this.

image

rfay:
I hope you'll be able to respond and check this out and we can get this in.

riconeitzel:
hey @rfay … can you quickly address how I can assist? Removal of set -x seems fine if we don't want the output. I don't care much so … fine to remove. However I cannot test the trial version, but I'm happy to shoot out a message to their dev team. They seemed to be quite fast responding to these issues. I'll keep you posted.

riconeitzel:
https://twitter.com/tableplus/status/1636890112410947589?s=46&t=IEaWKOHqOHo--eDWO8NnJg

So you’re using the current stable, I’m using the beta. We can keep this ticket open, till the next version is released, then it works for everyone.

rico

rfay:
Thanks for checking. Just say the word and we'll reopen this when its' released. You can just comment here. Thanks for your work on this!

We will want to make sure that this doesn't break people who have older versions.

huyphams:
Hi guys, I just wanna confirm that it will work for all. The reasons it does not work for the current version are:
1. There is a typo in the current version, we use enviroment instead of environment. We fixed it in the next update, also alias it to tag/env (you can use any of these params). So please use the correct params: env or tag 

2. The user who created this PR us the hotpatch, not the beta. Which fixed the issue.

huyphams:
You can download the hot patch release here: https://www.dropbox.com/s/3rbnoo74s03ekam/TablePlus.dmg?dl=0

riconeitzel:
I'm just amazed that @huyphams from tableplus follows the tweets to the pull request and takes the time to help! 🚀 😍

PR #4694, opened 2023-03-02
[docs] Update guidance about Colima/Docker Desktop requirements and Mutagen

rfay:
## The Issue

* One person reported confusion about the wording of Docker Desktop vs Colima, thinking that Colima was only for older versions of macOS
* Many people don't understand that Mutagen is the normal standard on macOS (and will be default on v1.22)
* VirtioFS is still problematic on filesystem permissions on Docker Desktop.

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -20,7 +20,7 @@ These environments can be extended, version controlled, and shared, so you can t
     * RAM: 8GB
     * Storage: 256GB
     * [Colima](https://github.com/abiosoft/colima) or [Docker Desktop](https://www.docker.com/products/docker-desktop/)
-    * Docker Desktop requires macOS Catalina (10.15) or higher; Colima runs on older systems
+    * Colima runs on many versions of macOS up through the latest. Docker Desktop requires macOS Big Sur (11) or higher.
 
 === "Windows WSL2"
     ### Windows WSL2
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -20,7 +20,8 @@ These environments can be extended, version controlled, and shared, so you can t
     * RAM: 8GB
     * Storage: 256GB
     * [Colima](https://github.com/abiosoft/colima) or [Docker Desktop](https://www.docker.com/products/docker-desktop/)
-    * Colima runs on many versions of macOS up through the latest. Docker Desktop requires macOS Big Sur (11) or higher.
+    * Colima (preferred by many) runs on many versions of macOS up through the latest. Docker Desktop (fully supported) requires macOS Big Sur (11) or higher.
+    * On macOS with either Colima or Docker Desktop most users will want to enable Mutagen for superb performance, `ddev config global --mutagen-enabled`
 
 === "Windows WSL2"
     ### Windows WSL2
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -20,10 +20,11 @@ These environments can be extended, version controlled, and shared, so you can t
     * RAM: 8GB
     * Storage: 256GB
     * [Colima](https://github.com/abiosoft/colima) or [Docker Desktop](https://www.docker.com/products/docker-desktop/)
-    * Colima (preferred by many) runs on many versions of macOS up through the latest. Docker Desktop (fully supported) requires macOS Big Sur (11) or higher.
-    * On macOS with either Colima or Docker Desktop most users will want to enable Mutagen for superb performance, `ddev config global --mutagen-enabled`
+    * Colima and Docker Desktop both require macOS Big Sur (11) or higher.
+    * On macOS with either Colima or Docker Desktop most users will want to enable Mutagen for best performance, `ddev config global --mutagen-enabled`
 
 === "Windows WSL2"
+
     ### Windows WSL2
 
     * RAM: 8GB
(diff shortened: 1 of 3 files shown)


docs/content/index.md
```diff
@@ -20,7 +20,8 @@ These environments can be extended, version controlled, and shared, so you can t
     * RAM: 8GB
     * Storage: 256GB
     * [Colima](https://github.com/abiosoft/colima) or [Docker Desktop](https://www.docker.com/products/docker-desktop/)
```
mattstein, line 22: Is there a definitive minimum version for Colima? It’d be nice to be able to say “Colima (macOS X–latest) or Docker Desktop (macOS 11 Big Sur or higher)”. This would clarify the confusion about requirements, which is exactly what this section focuses on.

The remaining bullets aren’t especially relevant here:

- What people prefer should probably be a component of whatever we officially recommend, which is still not a requirement.
- Colima and Docker Desktop are both fully supported as far as I know, so the “(fully supported)” parenthetical seems redundant and to apply that perhaps Colima is _not_ fully supported somehow.
- The Mutagen comment is out of place unless it has a direct impact on requirements, and people should be encouraged to install it not in the requirements but on the installation page.

docs/content/users/install/ddev-installation.md
```diff
@@ -41,9 +41,11 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     You can include a `-s <version>` argument to install a specific release or a prerelease version:
 
     ```
-    curl -fsSL https://raw.githubusercontent.com/drud/ddev/master/scripts/install_ddev.sh | bash -s v1.19.5
+    curl -fsSL https://raw.githubusercontent.com/drud/ddev/master/scripts/install_ddev.sh | bash -s v1.21.5
     ```
 
+    Most macOS users will want to enable Mutagen for superb performance; no installation is required, just `ddev config global --mutagen-enabled`.
```
mattstein, line 47: I’d... 

- Drop the redundant “macOS” since we’re in a macOS-specific section. 
- Make a strong recommendation so it’s not about popularity and we’re not leaving someone to wonder what macOS users _wouldn’t_ want to enable Mutagen. 
- Drop the explanation of steps that _don’t_ need to be taken. (Installation not required.)
- Replace “superb” with “the best” because it’s more about the best chance of great performance without being any kind of qualitative guarantee.
- Link to the broader explanation of the feature for context.
- Link the command to start reiterating how commands (and the documentation structure) work
- Remove that “just” and “run” the command [for consistency](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/).

```suggestion
    We recommend [enabling Mutagen](performance.md#mutagen) for the best performance, by running [`ddev config global --mutagen-enabled`](../usage/commands.md#config).
```

mattstein (changes requested):

docs/content/users/install/docker-installation.md
```diff
@@ -49,6 +49,8 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     Docker Desktop for Mac can be installed via Homebrew (`brew install homebrew/cask/docker`) or can be downloaded from [docker.com](https://www.docker.com/products/docker-desktop). It has long been supported by DDEV and has extensive automated testing.
 
+    If you use Docker Desktop for Mac, the `VirtioFS` option, while performant, is *not* recommended at this time and can cause many mysterious problems. Also,, when Mutagen is enabled, the performance implications of `VirtioFS` are not significant.
```
mattstein, line 52: Ah, I missed this while I was reviewing the other bits—sorry! 

I’d activate the language here for warmth and [tonal consistency](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/#use-active-third-person), add a link, and fix the double comma. 

I think it whittles down pretty well:

```
We do not recommend the “VirtioFS” option with Docker Desktop for Mac. While it’s performant, it can cause mysterious problems that are not present with [Mutagen](performance.md#mutagen)—which offers comparable performance when enabled.
```

docs/content/index.md
rfay: Actually, there isn't a minimum requirement for Colima, but there *is* for homebrew, and it would be nearly impossible to install without homebrew, so that's a good point. And... it turns out that Big Sur is also the minimum requirement for homebrew. 

Mutagen is actually a requirement with colima; if you use sshfs with colima then things don't work without mutagen.

mattstein: Ooof. Sounds like there could be a page just for navigating conditional requirements. Maybe it’d be best to list Big Sur as the requirement and add an FAQ like “Can DDEV run on an older Mac” that steps through this?

Or if it turns out nobody’s running older macOS versions and it’s over-serving an edge case, call out Big Sur and make sure the Colima instructions mention Mutagen and move on.

rfay: I just put Big Sur for both in the end.

DDEV works in a lot lot lot of different environments. It would be easier by far to limit them. But I do love that it works so many different places.

Lando actually requires (and installs during installation process, whether you already have docker or not) its own supported version of Docker Desktop.. and if you go outside that you're on your own.

rfay:
Hmm, you only made one actual suggestion, and I added that. Sorry to miss other?

Anyway, we've probably beaten it adequately.

mattstein:
First macOS mention of Mutagen reintroduces a redundant “macOS” and is missing a period.

The Windows Mutagen blurb adds the phrasing and linking issues I called out for macOS.

Further down, VirtioFS is in backticks when it should be quoted as an option or UI reference and not literal code. But I’ll take a hint and give up on this.

rfay:
I used VirtioFS in backticks because it wouldn't be in spellcheck, and I'm lazy.

PR #4695, opened 2023-03-02
Indicate that Colima is preferred.

cosmicdreams:


rfay:
Let's go with 
* https://github.com/drud/ddev/pull/4694

instead. Thanks for pointing this out!

PR #4700, opened 2023-03-03
Bump all images for v1.21.5

gilbertsoft:
## The Problem/Issue/Bug:

Preparing DDEV v1.21.5


rfay (changes requested):
ddev-dbserver needs to be updated here also of course.

rfay (approved):
Thanks!

PR #4705, opened 2023-03-04
[tests only] Use major versions for actions on some tests

gilbertsoft:
## The Issue

To avoid hassle with minor upgrades of Github Actions it is better to not to look a specific version and use a canonical version instead.

This patch replaces the current versions with a canonical one like suggested by the authors of the two actions. See https://github.com/marketplace/actions/markdown-link-check and https://github.com/marketplace/actions/github-spellcheck-action for more information.


rfay (approved):
Thanks!

PR #4712, opened 2023-03-05
Improve .editorconfig [skip ci]

gilbertsoft:
## The Issue

Some essential settings are missing and the editor defaults are used instead which could lead to differences with the various editors.

This patch introduces important default settings to avoid differences with the various editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,8 +5,16 @@ root = true
 
 # Unix-style newlines with a newline ending every file
 [*]
+charset = utf-8
 end_of_line = lf
+indent_style = space
+indent_size = 4
 insert_final_newline = true
+trim_trailing_whitespace = true
+
+# GO files
+[*.go]
+indent_style = tab
 
 # Makefile
 [Makefile]


.editorconfig
```diff
@@ -5,8 +5,16 @@ root = true
...
+charset = utf-8
 end_of_line = lf
+indent_style = space
+indent_size = 4
 insert_final_newline = true
+trim_trailing_whitespace = true
+
+# GO files
+[*.go]
+indent_style = tab
```
rfay, line 17: Nice that you can do this specifically by language. 

Of course... php is yet another world, and we have a few PHP files; not sure we have to deal with it though?

rfay (approved):
Seems reasonable to me.

rfay:
I guess the one question would be if this conflicts with any existing files, so every time sombody touches them with vscode they'll get changed for no reaason.

.editorconfig
gilbertsoft: That's now handled properly with the default config above. By default spaces are used and exceptions are provided later for `*.go` and `Makefile`. Maybe we have to define other exception I don't see at the moment.

gilbertsoft:
> I guess the one question would be if this conflicts with any existing files, so every time sombody touches them with vscode they'll get changed for no reaason.

I've quickly checked the existing files and so far I can see this should work and not make any conflicts.

rfay:
Let me know when you want me to pull it.

gilbertsoft:
> Let me know when you want me to pull it.

Go ahead at any time.

rfay:
I let the tests run on purpose for this one, just to see if we're getting any better commit statuses.

rfay:
I think we probably want an indent of 2 spaces by default and 4 for markdown? /cc @mattstein ?

PR #4722, opened 2023-03-07
[docs] Clarify how you upgrade DDEV

bencroker:
## The Issue

It is not immediately apparent that the installation page in the docs contains upgrading instructions. 

This PR makes it clear as crystal by adding a heading, which can additionally be linked to at https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#upgrading

Screenshot 2023-03-06 at 18 55 09

## Testing

Affects only one page of the docs, no testing required.

## Release/Deployment Notes

> Added an “Upgrading” heading to the installation page for clarity.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4722"></a>

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -4,7 +4,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
 ## Upgrading
 
-Upgrading DDEV is basically the same process as installing it, because you're upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
+Upgrading DDEV is basically the same process as installing it, because you’re upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
 
 === "macOS"
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -2,7 +2,9 @@
 
 Once you’ve [installed a Docker provider](docker-installation.md), you’re ready to install DDEV!
 
-Installing and upgrading DDEV are nearly the same thing, because you're upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
+## Upgrading
+
+Upgrading DDEV is basically the same process as installing it, because you're upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
 
 === "macOS"
 


mattstein:
Thanks @bencroker!

The callout and ability to link to it are obviously an improvement, and I’m wondering if it’s still odd to visit an install page and immediately get a note about upgrading before the actual install instructions. (Your PR doesn’t create this issue, it just further calls attention to it.)

Two counter proposals that might help this:

## Option A

Use a callout, for visual emphasis but also a sort of aside:

Screen Shot 2023-03-06 at 05 43 03 PM@2x

Feels more like the side note it is, but loses the ability to link directly. (Even though we could add a hidden anchor for that.)

## Option B

Keep this exactly how you have it, but move the heading + text _below_ the tabbed fields:

Screen Shot 2023-03-06 at 05 45 45 PM@2x

Most install instructions mention upgrading anyway, and this would allow the reader to get to them sooner while reiterating (with a clear heading) that installing and upgrading are almost the same thing.

bencroker:
Thanks for the considered response, @mattstein.

Option A is visually great, but would require an anchor link to really be effective.

Option B I feel is lost all the way down there, meaning you’d have to scroll back up to find the code samples.

mattstein:
Welp, I can’t get an invisible anchor to work with a callout locally—so ignore me and thanks for the PR! Works for me.

docs/content/users/install/ddev-installation.md
```diff
@@ -2,7 +2,9 @@
 
 Once you’ve [installed a Docker provider](docker-installation.md), you’re ready to install DDEV!
 
-Installing and upgrading DDEV are nearly the same thing, because you're upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
+## Upgrading
+
+Upgrading DDEV is basically the same process as installing it, because you’re upgrading the `ddev` binary that talks with Docker. You can update this file like other software on your system, whether it’s with a package manager or traditional installer.
```
rfay, line 7: The problem is... this isn't actually true here. With brew it would be `brew upgrade`. With apt it would be `apt upgrade` as opposed to `apt install`.  

I do think this needs attention, but maybe it just needs more attention in the separate sections?

rfay (changes requested):

docs/content/users/install/ddev-installation.md
mattstein: What part of that overview isn’t true? The idea, I think, is that you’re updating an executable in the same way you first installed it.

Someone coming back to update needs a clear place to land, which is why Ben’s heading makes sense and my suggestion (B) of moving it further down the page defeats the purpose. I think shifting update notes exclusively into each tab would be a similar problem. Unless each tab had clear, separate _Installing_ and _Upgrading_ subheadings.

Or upgrading could get its own page?

rfay: The idea is correct, but I think it implies the the process is always the same. So it's true that you're mostly just changing out a binary, but the execution, which is what most people care about, is slightly different, based on the flavor of the package manager, except in the case of the script. 

If you look at this in context, it also brings up upgrading before installation. Somehow we have to explain both. 

I was thinking maybe this could change to an overview of upgrading and installing without much change. This paragraph could explain both by getting across the idea that we're just swapping a binary, but the difference may be different with different package managers.

rfay: To be clear, this is an important initiative, we just have to find the way to express it that gets people to the right place.

mattstein:
I didn’t forget about this, I just need time to think about it and I’ll probably end up trying a bigger PR.

bencroker:
Sorry!

mattstein:
Ha, no worries @bencroker! I appreciate the nudge because it’s definitely worth improving.

mattstein:
Finally attempted to address this more thoroughly in #4789.

bencroker:
Perfect, thanks and well done, @mattstein!

PR #4723, opened 2023-03-07
Make sure the .ddev/.downloads directory is available, even on mutagen, fixes #4721

### LINKED ISSUE #4721: v1.21.5: `ddev pull platform` doesn't work with mutagen enabled due to .ddev/.downloads not being synced

_No response_

### Expected Behavior

`ddev pull platform` should "just work"

### Actual Behavior

In v1.21.5, the mutagen.yml was changed to exclude .ddev/.downloads, assuming it didn't need to be used. So now with mutagen enabled, at least `ddev pull platform` doesn't work: 

```
$ ddev pull platform
You're about to delete the current database and files and replace with the results of a fresh pull.
Would you like to continue? [Y/n] (yes):
Authenticating...
Obtaining databases...


  [InvalidArgumentException]
  Directory not found: /var/www/html/.ddev/.downloads
```

This change happened in

* https://github.com/ddev/ddev/pull/4646

### Steps To Reproduce

_No response_

### Anything else?

Why didn't the TestPlatformPull catch this problem?

rfay:
## The Issue

* #4721 

`ddev pull platform` and probably other things are broken with mutagen enabled on DDEV v1.21.5.

Bind-mount the .downloads directory.

Note that this isn't working quite right on linux.

## TODO

- [x] Verify whether the extra ownership and permissions setting code is even relevant.

- [x] `ddev pull platform` on macOS, without mutagen, Colima
- [x] `ddev pull platform` macOS, with mutagen, Colima
- [x] `ddev pull platform` macOS, Docker Desktop, without
- [x] Docker Desktop with mutagen
- [x] Windows Docker Desktop mutagen
- [x] Windows Docker Desktop no mutagen
- [x] See if it can work on Linux

## Automated Testing Overview

I haven't sorted out how to do this better. It would be better if we added a mutagen-based TestPullPlatform, etc.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4723"></a>

### DIFF
--- a/docs/content/users/install/performance.md
+++ b/docs/content/users/install/performance.md
@@ -15,13 +15,13 @@ Mutagen can offer a big performance boost on macOS and Windows. It’s fast and 
     ## Mutagen
 
     If you’re on macOS or Windows, you can skip everything else here and enable Mutagen:
-
+    
     ```
     ddev config global --mutagen-enabled
     ```
 
     ### What Mutagen Does
-
+    
     The [Mutagen](https://mutagen.io) asynchronous caching feature is the best way to improve DDEV’s web-serving performance on macOS and Windows, and we recommend it for most projects. It can be significantly faster than NFS, massively faster than plain Docker or Colima, and it makes filesystem watchers (`fsnotify`/`inotify`) work correctly.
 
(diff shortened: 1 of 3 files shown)


rfay:
* `ddev pull platform` doesn't work on traditional Windows (either git-bash or cmd). AFAICT it never has. I checked v1.21.4. 
```
Downloaded db dumps for databases 'db'
Importing databases [C:\Users\randy\workspace\platform-d9mariadb\.ddev\.downloads\db.sql.gz]

ERROR at line 1: Unknown command '\U'.
Pull failed: failed to import database: exit status 1
stdout:
stderr:
```

* This works fine on Linux with or without mutagen now.
* People who `ddev pull` enormous files or databases when using mutagen will still see some interesting things go on, but files just left laying around are deleted on start.

joelpittet:
Anybody stumbling here the `ddev mutagen reset` got it working again followed by `ddev restart`. Thanks for the fix in v1.21.6

PR #4724, opened 2023-03-07
Append DDEV_EXECUTABLE to the Host commands environment

### LINKED ISSUE #4717: Allow add-ons to declare other add-ons as dependencies
I created a few addons (and sometimes I used others) and I thought it might be good to have a way to bundle add-ons so one can do ddev get xxx/add-ons and get a bunch at once.

One potentially nice way of doing it is with the ability of an addon to declare dependencies, so one would do the usual ddev get and if more dependencies are included, ddev could run ddev get on each. All GH, https and local add-ons can be supported, namely the same as running ddev get locally? I just thought that it can now maybe be done using post actions, but this is declaratively better.

Just an idea.

hanoii:
Tangentially resolves #4717.

As discussed on https://discord.com/channels/664580571770388500/1082435430479306803/1082435430479306803 instead of having an add on define dependencies, the add-on can already install those from post-install actions.

This PR only adds an environment variable so that the running ddev binary can be used in any ddev host Command  (also pre/post install actions) through an env var.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4724"></a>

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -10,10 +10,19 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+func HostCommand(name string, args ...string) *exec.Cmd {
+	c := exec.Command(name, args...)
+	ddevExecutable, _ := os.Executable()
+	c.Env = append(os.Environ(),
+		"DDEV_BINARY_FULLPATH="+ddevExecutable,
+	)
+	return c
+}
+
 // RunCommand runs a command on the host system.
 // returns the stdout of the command and an err
 func RunCommand(command string, args []string) (string, error) {
-	out, err := exec.Command(
+	out, err := HostCommand(
 		command, args...,
 	).CombinedOutput()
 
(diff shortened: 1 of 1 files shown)


hanoii:
I _think_ tests failed for some other reason, can they be re-run?

rfay:
I think TestCmdGetComplex failed everywhere. You should be able to demonstrate it locally.

I'll add you on github and buildkite though so you can re-run yourself.

hanoii:
@rfay so only one test failed and I am not sure how to rerun it. buildkit looks fine and here I coulldn't find anywhere to re reun it. On this failed test I see no error.

rfay:
Congrats. It was a random docker fail. I restarted it.

rfay:
Will take a look when you think this is ready for review.

rfay:
I'll change it to "ready for review" then. You probably didn't notice that it was in "draft" mode.

rfay (approved):
OK, this looks ok and I tested with a pre-start exec-host hook. I only tested on macOS, hope it does the right thing on Windows. I hope it solves a real problem. Not completely excited about it :) But thanks for taking the time to scratch your itch!

PR #4728, opened 2023-03-08
Revert full-project Dockerfile context, fixes #4727

### LINKED ISSUE #4727: v1.21.5: Large projects with many mounted files take way too long to start

_No response_

### Expected Behavior

`ddev start` shouldn't be slower than it was in v1.21.4

### Actual Behavior

With huge projects, particularly with projects that have mounted components as @mkalkbrenner and @cspitzlay have, the `ddev start` time can degenerate to an hour and a half startup time. 

We studied this and found that the `~/.ddev/bin/docker-compose -f .ddev/.ddev-docker-compose-full.yaml up --build -d` step hung for the 90 minutes. Without the `--build -d` it did not. 

Our analysis said that what's happening is that the docker context is being built there for the build. And in v1.21.5, the docker build context includes the entire project. Which in their case includes more thousands of files than even an ordinary modern PHP project. And many thousands of them are mounted on remote filesystems. The result is that the docker build stage is going through the entire mess and copying them into the docker context.

This change was made in 
* https://github.com/ddev/ddev/pull/4549

but that PR did two different changes, so the custom Dockerfile thing will have to be redone to go back to the previous behavior.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #4727

Revert the feature that allowed using entire project for context

Use `ADD <somefile> /var/tmp` in a custom .ddev/web-build/Dockerfile. It should get added.
Use `ADD .ddev/web-build/<somefile /var/tmp` - it should fail.

## Automated Testing Overview

The customizations to tests that were added to that feature were removed.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4728"></a>

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -78,7 +78,7 @@ RUN chmod -R ugo+rw $COMPOSER_HOME
 ENV COMPOSER_HOME=""
 ```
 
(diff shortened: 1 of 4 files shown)


mkalkbrenner:
I tested the uploaded artifacts here. Unfortunately, the problem still exists with v1.21.5-4-g5c9c1828

PR #4735, opened 2023-03-10
[docs] Add ExecRaw annotation documentation.

ultimike:
The title says it all....

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4735"></a>

### DIFF
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -218,6 +218,14 @@ If your container command should run from the directory you are running the comm
 
 Example: `## HostWorkingDir: true`
 
+### “ExecRaw” Annotation (Container Commands Only)
+
+Use `ExecRaw: true` to pass arguments for the command you want to run directly to the command in the container as-is. 
+
+For example, when ExecRaw is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
+
+Example: `## ExecRaw: true`
+
 ## Known Windows Issues
 
 ### Line Endings
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -224,6 +224,8 @@ Use `ExecRaw: true` to pass arguments for the command you want to run directly t
 
 For example, when ExecRaw is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/extend/custom-commands.md
```diff
@@ -218,6 +218,14 @@ If your container command should run from the directory you are running the comm
 
 Example: `## HostWorkingDir: true`
 
+### “ExecRaw” Annotation (Container Commands Only)
+
+Use `ExecRaw: true` to pass arguments for the command you want to run directly to the command in the container as-is. 
+
+For example, when ExecRaw is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
```
rfay, line 225:
```suggestion
For example, when ExecRaw is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.

Most commands will work better with this annotation. This is recommended for almost all container commands.
```

rfay (approved):
Just added one suggestion you can commit if you like it.

docs/content/users/extend/custom-commands.md
```diff
@@ -218,6 +218,14 @@ If your container command should run from the directory you are running the comm
 
 Example: `## HostWorkingDir: true`
 
+### “ExecRaw” Annotation (Container Commands Only)
+
+Use `ExecRaw: true` to pass arguments for the command you want to run directly to the command in the container as-is.
```
mattstein, line 223: This sentence seems a little tricky to parse. Maybe this whittles it down without losing meaning?

```suggestion
Use `ExecRaw: true` to pass command arguments directly to the container as-is. 
```

docs/content/users/extend/custom-commands.md
mattstein: > Most commands will work better with this annotation. This is recommended for almost all container commands.

Two things:

1. This reads like “you should mostly use this” without qualifying why. Is it possible to give an example of when this is and is not ideal to help someone navigate their own decision?
2. Ideally recommendations can be [active third person](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/#use-active-third-person); “this is recommended” → “we recommend”.

mattstein (changes requested):
Added my nitpickery!

docs/content/users/extend/custom-commands.md
rfay: Thanks @mattstein - I updated. But shouldn't that be called "active first person"?

PR #4736, opened 2023-03-10
[docs] Document exposing additional non-HTTP ports on web

deviantintegral:
## The Issue

It wasn't clear to me that a services file could _only_ modify the web service, without declaring an additional container to run. When I first started trying to figure this out, I looked on this page so it's where I've added the new docs.

A new section on exposing non-HTTP ports.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4736"></a>

### DIFF
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -111,7 +111,7 @@ web_extra_exposed_ports:
 
 ## Exposing Extra Non-HTTP Ports
 
-While the `web_extra_exposed_ports` gracefully handles running multiple ddev projects at the same time, it can't forward  ports for non-HTTP TCP or UDP daemons. Instead, ports can be added in a `docker-compose.*.yaml` file. This file does not need to specify an additional services. For example, this configuration exposes port 5900 for a VNC server.
+While the `web_extra_exposed_ports` gracefully handles running multiple DDEV projects at the same time, it can't forward ports for non-HTTP TCP or UDP daemons. Instead, ports can be added in a `docker-compose.*.yaml` file. This file does not need to specify an additional services. For example, this configuration exposes port 5900 for a VNC server.
 
 In `.ddev/docker-compose.vnc.yaml`:
 
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -118,9 +118,9 @@ In `.ddev/docker-compose.vnc.yaml`:
 ```yaml
 version: '3.6'
 services:
-    web:
-        ports:
-            - "5900:5900"
+  web:
+    ports:
+      - "5900:5900"
 ```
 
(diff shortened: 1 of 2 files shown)


rfay (approved):
Looks good to me, thanks!

docs/content/users/extend/customization-extendibility.md
```diff
@@ -109,6 +109,22 @@ web_extra_exposed_ports:
 !!!warning "Fill in all three fields even if you don’t intend to use the `https_port`!"
     If you don’t add `https_port`, then it default to `0` and `ddev-router` will fail to start.
 
+## Exposing Extra Non-HTTP Ports
+
+While the `web_extra_exposed_ports` gracefully handles running multiple ddev projects at the same time, it can't forward  ports for non-HTTP TCP or UDP daemons. Instead, ports can be added in a `docker-compose.*.yaml` file. This file does not need to specify an additional services. For example, this configuration exposes port 5900 for a VNC server.
```
mattstein, line 114: I’d capitalize DDEV in its product form and remove that bonus space:

```suggestion
While the `web_extra_exposed_ports` gracefully handles running multiple DDEV projects at the same time, it can't forward ports for non-HTTP TCP or UDP daemons. Instead, ports can be added in a `docker-compose.*.yaml` file. This file does not need to specify an additional services. For example, this configuration exposes port 5900 for a VNC server.
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -109,6 +109,22 @@ web_extra_exposed_ports:
...
+
+In `.ddev/docker-compose.vnc.yaml`:
+
+```yaml
+version: '3.6'
+services:
+    web:
+        ports:
+            - "5900:5900"
```
mattstein, line 123: Most examples in the docs use two spaces for indentation, so it’s probably best to keep it consistent:

```suggestion
version: '3.6'
services:
  web:
    ports:
      - "5900:5900"
```

mattstein (changes requested):
Added some minor stylistic nitpicks, otherwise looks good to me!

docs/content/users/extend/customization-extendibility.md
deviantintegral: Good catch, I always use 2 spaces too, so I'm not sure what happened!

mattstein: Just realized this could be a side effect from [this PR](https://github.com/ddev/ddev/pull/4712), if it you were editing locally and it formatted on save. I did some battle with it yesterday.

deviantintegral: Yep, I have the editorconfig plugin installed in goland. Thanks for pointing out why that happened!

rfay: Proposal in 
* https://github.com/ddev/ddev/pull/4745

Not sure how we can say indent-4 always in markdown, when it affects fenced codeblocks with yaml.

PR #4739, opened 2023-03-10
[docs] Fix formatting and phrasing

mattstein:
## The Issue

This is a late followup review of #4723, to “run” commands and make sure the summarize+explain list formatting is consistent.

It also adds `<br>` elements, since the deliberate two-space line breaks (`  `) are stripped out after #4712—which is still an improvement since so many docs PRs get snagged on whitespace issues. I’m positive I have more deliberate line breaks scattered around that will get removed, but nobody will be injured.

[Preview changes locally](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) or inspect the [automatic build](https://ddev--4739.org.readthedocs.build/en/4739/users/install/performance/).

## Automated Testing Overview

n/a

## Related Issue Link(s)

- #4723
- #4712

## Release/Deployment Notes

n/a

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4739"></a>

### DIFF
--- a/docs/content/users/install/performance.md
+++ b/docs/content/users/install/performance.md
@@ -15,13 +15,13 @@ Mutagen can offer a big performance boost on macOS and Windows. It’s fast and 
     ## Mutagen
 
     If you’re on macOS or Windows, you can skip everything else here and enable Mutagen:
-    
+
     ```
     ddev config global --mutagen-enabled
     ```
 
     ### What Mutagen Does
-    
+
     The [Mutagen](https://mutagen.io) asynchronous caching feature is the best way to improve DDEV’s web-serving performance on macOS and Windows, and we recommend it for most projects. It can be significantly faster than NFS, massively faster than plain Docker or Colima, and it makes filesystem watchers (`fsnotify`/`inotify`) work correctly.
 
     Mutagen decouples in-container reads and writes from reads and writes on the host machine, so each can enjoy near-native speed. A change on the host gets changed “pretty soon” in the container, and a change in the container gets updated “pretty soon” on the host; neither filesystem is stuck waiting on the other one. This “pretty soon” means, however, that there’s a brief window where files on the host may not exactly match the files inside the container—so files that manage to change in both places can lead to conflicts.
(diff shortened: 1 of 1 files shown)


rfay (commented):
I'm getting new complaints from markdownlint about those trailing spaces, and goland is trying to clean them up for me. I think there's a bunch more, not sure.

But I don't really like the trailing `<br>`, which seems complex and adds mixed markdown/html where we're rather avoid it. Is it a good thing?

mattstein:
> But I don't really like the trailing `<br>`, which seems complex and adds mixed markdown/html where we're rather avoid it. Is it a good thing?

It’s not my preference either, but the usual double-space convention (`  `) gets stripped out. Another convention is ending the line with a slash (`\`) instead of two spaces, but I tried that and it doesn’t parse into a line break.

The breaks themselves are as intentional as the summaries, since these are some really verbose bullet points. The bold summaries and breaks make them much more readable:

Screen Shot 2023-03-10 at 03 45 44 PM@2x

rfay:
> the usual double-space convention (  ) gets stripped out

stripped out by whom?

mattstein:
This old chap: https://github.com/ddev/ddev/pull/4712/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR13

(Editor config.)

rfay:
Ah it's .editorconfig that has convinced my Goland to be fiddling with markdown. I thought it was my markdownlint. 

Should we just remove that?

rfay:
@gilbertsoft just added that the other day.

mattstein:
> Should we just remove that?

Not sure, I think we’re playing whack-a-mole.

Like I wrote in the description, automatically stripping spaces from Markdown seems _overall_ like a good thing since whitespace issues are frequently obstacles to PRs.

It’ll mean some of my precious intentional breaks go away, but that seems less important.

🤷‍♂️

rfay (approved):

mattstein:
Pull away whenever you’d like!

PR #4740, opened 2023-03-11
[docs] Requirements: reduce tab width, eliminate redundancy

mattstein:
## The Issue

A few minor things about this page amount to cruft:

1. The “Gitpod and GitHub Codespaces” tab requires horizontal paging when it doesn’t need to.
2. The macOS requirements instruct the reader to enable Mutagen, which is neither a requirement nor anything they have to install. (I brought that up [here](https://github.com/ddev/ddev/pull/4694#discussion_r1125190704) but I think that got lost.)
3. The macOS version requirement applies to both Colima and Docker Desktop, so calling them both out is redundant.

Screen Shot 2023-03-10 at 04 34 42 PM@2x

1. Whittles the tab name down to “Gitpod & Codespaces”, which fits great and is more consistent with the [DDEV Installation](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/) page.
2. Removes the Mutagen comment/instruction from requirements. It’s an explicit step in [Colima’s Docker install sequence](https://ddev.readthedocs.io/en/latest/users/install/docker-installation/#colima), a recommendation in [Docker Desktop’s](https://ddev.readthedocs.io/en/latest/users/install/docker-installation/#docker-desktop-for-mac), and mentioned again in the [macOS DDEV Installation](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/) instructions.
3. Pares down the OS version requirement line and moves it to the top since any other details are probably a non-starter an old OS. (Edge cases could be covered in the FAQ.)

Screen Shot 2023-03-10 at 04 35 30 PM@2x

[Preview changes locally](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) or inspect the [automatic build](https://ddev--4740.org.readthedocs.build/en/4740/).

## Automated Testing Overview

n/a

## Related Issue Link(s)

Discussed partially in #4694.

## Release/Deployment Notes

n/a

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4740"></a>

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -14,14 +14,13 @@ These environments can be extended, version controlled, and shared, so you can t
 === "macOS"
 
     ### macOS
-    
+
     Runs natively on ARM64 (Apple Silicon) and AMD64 machines.
 
+    * macOS Big Sur (11) or higher
     * RAM: 8GB
     * Storage: 256GB
     * [Colima](https://github.com/abiosoft/colima) or [Docker Desktop](https://www.docker.com/products/docker-desktop/)
-    * Colima and Docker Desktop both require macOS Big Sur (11) or higher.
-    * On macOS with either Colima or Docker Desktop most users will want to enable Mutagen for best performance, `ddev config global --mutagen-enabled`
 
 === "Windows WSL2"
 
@@ -50,7 +49,7 @@ These environments can be extended, version controlled, and shared, so you can t
     * RAM: 8GB
     * Storage: 256GB
 
-=== "Gitpod and GitHub Codespaces"
+=== "Gitpod & Codespaces"
 
     ### Gitpod and GitHub Codespaces
 
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -17,7 +17,7 @@ These environments can be extended, version controlled, and shared, so you can t
 
     Runs natively on ARM64 (Apple Silicon) and AMD64 machines.
 
-    * macOS Big Sur (11) or higher
+    * macOS Big Sur (11) or higher ([mostly](./users/usage/faq.md#can-i-run-ddev-on-an-older-mac))
     * RAM: 8GB
     * Storage: 256GB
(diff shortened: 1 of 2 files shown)


rfay (approved):
This is fine, thanks. 

I feel guilty saying that DDEV requires Big Sur when it's really not true. 

Should we link to @rpkoller 's SO post about older systems? https://stackoverflow.com/a/69964995/215713

mattstein:
> I feel guilty saying that DDEV requires Big Sur when it's really not true.

That’s what I meant by “edge cases could be covered in the FAQ.” If it doesn’t affect most people, but having a nuanced answer is important to catch edge cases, we might as well add a FAQ item either stealing or complementing that answer for visibility.

I can tack that onto this PR if you agree!

mattstein:
Could even say it “mostly” requires Big Sur and link “mostly” to the FAQ item if that’s not too cheeky.

rfay:
You're right, adding to FAQ is a great way to do it. Then link the mostly here to the FAQ?  Do you want to add that here?

Thanks for your patience with this.  DDEV because of golang runs so well so many places, and then time keeps moving on and invalidating everything.

I do think that brew and Docker Desktop will remain in lockstep with their requirements, because they seem to have the same strategy. 

OTOH, colima is planning to vendor Lima (but perhaps not other components) which might make it end up installable on older versions because not dependent on brew. But we can let that come when it comes.

mattstein:
👍 Taking this into draft mode and adding an FAQ item about support for older macOS.

rfay:
It doesn't have to have everything @rpkoller wrote, maybe just say you can use old Docker Desktop and link to his post?

mattstein:
> It doesn't have to have everything @rpkoller wrote, maybe just say you can use old Docker Desktop and link to his post?

Right, I don’t intend to cram too much into the FAQ. If I can summarize and link I will, but I’m going to need a minute to catch up and squeeze some words out.

mattstein:
Ready again! Minimal distraction from requirements, straightforward FAQ answer with a link to the longer one. Feels pretty good.

rfay (approved):
Fantastic, thanks so much.

PR #4744, opened 2023-03-11
[docs] Fix up ExecRaw spellcheck trouble

rfay:
## The Issue

ExecRaw started causing spellcheck trouble. I probably pulled something when tests hadn't finished.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4744"></a>

### DIFF
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -218,11 +218,11 @@ If your container command should run from the directory you are running the comm
 
 Example: `## HostWorkingDir: true`
 
-### “ExecRaw” Annotation (Container Commands Only)
+### `ExecRaw` Annotation (Container Commands Only)
 
 Use `ExecRaw: true` to pass command arguments directly to the container as-is.
 
-For example, when ExecRaw is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
+For example, when `ExecRaw` is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
 
 We recommend  using this annotation for all container commands. The default behavior is retained to avoid breaking existing commands.
 


docs/content/users/extend/custom-commands.md
```diff
@@ -218,11 +218,11 @@ If your container command should run from the directory you are running the comm
 
 Example: `## HostWorkingDir: true`
 
-### “ExecRaw” Annotation (Container Commands Only)
+### `ExecRaw` Annotation (Container Commands Only)
```
mattstein, line 221: @rfay This fixes the spell check issue to the detriment of the reader by not following the format of the page. I’d either wrap all the heading annotations in backticks (`“HostWorkingDir” Annotation` → <code>`HostWorkingDir` Annotation</code>, or add “ExecRaw” to the exclusion list.

rfay: I don't think `ExecRaw` should be in the spellcheck list because it's not a word.

mattstein: These aren’t words either:

- `ProjectTypes`
- `OSTypes`
- `HostBinaryExists`
- `DBTypes`
- `HostWorkingDir`

But they’re quoted in headings on this page, just as `ExecRaw` was. I’m saying that if you’re changing the heading format to appease the spell checker, you might as well update the format for the rest of them so you’re not signaling to the reader that `ExecRaw` is somehow different. (Unless this is intentional and meaningful break in format and I’m missing the point.)

rfay: Hmm, so I probably added all those to spellcheck huh? In that case I'll just do that.

mattstein: Sure, or use backticks for the rest of the headings.

> I’d either wrap all the heading annotations in backticks (`“HostWorkingDir” Annotation` → <code>`HostWorkingDir` Annotation</code>, etc.), or add “ExecRaw” to the exclusion list.

I’m not sure how else to say this; whatever the answer, I think the reader that should matter more than the spell checker. The consistency of the page is for the reader.

rfay: Fixed in https://github.com/ddev/ddev/pull/4746, you don't need to say it any more :)

PR #4745, opened 2023-03-11
Revert indentation specifier from editorconfig settings

rfay:
## The Issue

In 
* #4712 

@gilbertsoft made some very reasonable improvements to the .editorconfig, but they've been causing some trouble in a number of places.

* Embedded yaml in codeblocks in yaml gets 4-space (and ugly) indentation
* Lots of failures are getting shown with existing stuff that gets edited and the editor magically removes trailing spaces.

* Don't force the 4-space indentation (remove indent_size)
* Clean up trailing spaces in our Markdown files so the current restriction about trailing spaces doesn't trigger on Markdown at least. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4745"></a>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,9 +8,13 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 4
+# Forcing indent to 4 seems to break lots of things, like
+# embedded yaml in markdown.
+# indent_size = 4
 insert_final_newline = true
-trim_trailing_whitespace = true
+# Lots of editors and practices leave trailing whitespace
+# that does no harm.
+#trim_trailing_whitespace = true
 
 # GO files
 [*.go]
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,8 +12,6 @@ indent_style = space
 # embedded yaml in markdown.
 # indent_size = 4
 insert_final_newline = true
-# Lots of editors and practices leave trailing whitespace
-# that does no harm.
 trim_trailing_whitespace = true
 
 # GO files
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ indent_style = space
 insert_final_newline = true
 # Lots of editors and practices leave trailing whitespace
 # that does no harm.
-#trim_trailing_whitespace = true
+trim_trailing_whitespace = true
 
 # GO files
 [*.go]
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -29,6 +29,10 @@ jobs:
       - uses: rojopolis/spellcheck-github-actions@v0
         name: Spellcheck
       - run: echo "/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin" >> $GITHUB_PATH
+      - uses: snow-actions/eclint@v1.0.1
+        name: eclint
+        with:
+          args: 'check'
       - name: Run markdownlint on docs
         uses: docker://avtodev/markdown-lint:v1
         with:
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -29,10 +29,6 @@ jobs:
       - uses: rojopolis/spellcheck-github-actions@v0
         name: Spellcheck
       - run: echo "/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin" >> $GITHUB_PATH
-      - uses: snow-actions/eclint@v1.0.1
-        name: eclint
-        with:
-          args: 'check'
       - name: Run markdownlint on docs
         uses: docker://avtodev/markdown-lint:v1
         with:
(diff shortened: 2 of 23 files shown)


mattstein:
I haven’t had responded to [this question](https://github.com/ddev/ddev/pull/4712#issuecomment-1464944807) yet because I think the best answer is a diligent one—not adopting or reverting the change but testing it carefully:

1. Taking inventory of the file types we have and how they’re formatted, making rules that are as specific as possible. Probably specifying four spaces for Markdown and PHP, and two spaces for YAML and JavaScript and JSON—but the answer should include everything we know about.
2. Applying the rule set in a commit, instead of seeing what happens as people submit PRs over a long period of time.

Reverting this change gets rid of the good it does skipping whitespace issues in PRs. Applying it creates new issues. I haven’t had time to take a careful run at it yet.

rfay:
We can keep the `trim_trailing_whitespace` if we do a PR to fix all existing markdown files. Or maybe all files.

mattstein:
I swapped `<br>` elements for intentional line breaks and attempted to be more specific with file types [over here](https://github.com/ddev/ddev/compare/master...mattstein:ddev:docs/editorconfig-refinement), but the settings seem to be only spontaneously applied in VS Code.

I uninstalled Prettier to rule out interference, and while trailing spaces and empty lines are automatically cleaned up on save I still have a few problems:

1. Four-space YAML indentation is not fixed (to two spaces) in fenced code blocks.
2. Shell scripts (ending in `.sh`) can have a mixture of tabs, spaces, and 1-, 2-, 3-, or 4-space indentations—sometimes in [the same file](https://github.com/ddev/ddev/blob/master/.buildkite/sanetestbot.sh)—without being fixed on save.
3. Nothing changes when I save [`docs/content/assets/extra.css`](https://github.com/ddev/ddev/blob/master/docs/content/assets/extra.css).
4. Saving `.go` files automatically adds line breaks between some lists if imports, and I’m not sure if that’s a thing we want.

In same cases above, I can use <kbd>SHIFT</kbd> + <kbd>⌘</kbd> + <kbd>P</kbd> and choose “Reformat lines” to get the desired effect. Not always, though.

I installed and disabled the EditorConfig VS Code extension (even though editorconfig.org says no plugin is necessary), and tried setting it as my default formatter instead of none. No real change.

I’d like to apply the rules to all the files in the project to quantify the impact it has, but I’m struggling to apply the rules to _individual_ files as a precursor.

Maybe my setup’s just a dumpster fire. Anybody else get consistent results with [an updated config](https://github.com/ddev/ddev/compare/master...mattstein:ddev:docs/editorconfig-refinement#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5b)?

rfay:
I suspect that we can apply settings to all *markdown* files with 

`markdownlint --fix <file> -c $ProjectFileDir$/.markdownlint.jsonc`


```
find docs *.md -name "*.md" -exec markdownlint --fix {} \;
```

I've been intending to try it and add results to this PR.

rfay:
That exact recipe doesn't work.

rfay:
Ah, [eclint](https://www.npmjs.com/package/eclint)

`eclint fix`

rfay:
There are [several editorconfig actions](https://github.com/marketplace?type=actions&query=editorconfig+), all seem to be very poorly maintained. The one I tried doesn't have any option to limit the files or file types that it acts on.

rfay:
Here are the 12,000 lines of errors from `eclint check` with our current config:

https://gist.github.com/rfay/dfc23c73b4c95c35f3eb662f5f80c57c

The problem with this is that editors that use it, when the file is visited, change more than the planned change, which makes PRs ugly and hard to comprehend.

rfay:
There are so many, many unintended consequences of .editorconfig.

For example, this one:

> cmd/ddev/cmd/commands_test.go
 >    436:02 ❌ invalid indent style: space, expected: tab
 >                                                  (EditorConfig indent_style https://goo.gl/8Qkrbr)


The code it's complaining about is json embedded in golang:

```
	packageJSON := `
	{
	  "name": "junk",
	  "version": "1.0.0",
	  "description": "",
	  "main": "index.js",
	  "devDependencies": {},
	  "scripts": {
		"test": "echo \"Error: no test specified\" && exit 1"
	  },
	  "author": "",
	  "license": "ISC"
	}
```

It would be incorrect if indented with tabs.

rfay:
This now only removes the indentation restriction, because mixed code doesn't work well with it (like yaml in markdown). 

All of the markdown files have had all of their trailing spaces removed in this PR as well.

I'm not sure I like editorconfig anywhere (except for new code) because of the unintended edits some IDEs do as a result.

mattstein:
I don’t know of a great way to solve code consistency without sharing organization-wide IDE settings, repository configs, and linting tools that go all in on encouraging or enforcing a specific style.

Please give me a poke if you merge this so I can fix any intentional line breaks that get stripped out.

PR #4747, opened 2023-03-12
[docs] Clarification for some MacOS users using install_ddev.sh and missing /usr/local/bin

iparr:
## The Issue

Surprising script failure on a fresh MacOS install.

Provides guidance on this issue.

n/a

## Automated Testing Overview

Documentation change only.

## Related Issue Link(s)

n/a

## Release/Deployment Notes

n/a

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4747"></a>

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -30,7 +30,9 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
 
     !!!tip
         The install script works on macOS, Linux, and Windows WSL2.
-
+    
+    !!!tip
+        You may find that on fresh macOS installs, the `/usr/local/bin` folder does not exist and the script will fail. Create it with `sudo mkdir /usr/local/bin` first and then run the script again.
 
     Run the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) to install or update DDEV. It downloads, verifies, and sets up the `ddev` binary:
 


rfay (changes requested):
You were using the install script and it failed because /usr/local/bin didn't exist? It might be better to update the script in that case? 

Note that if homebrew has been installed (on AMD64) then /usr/local/bin already exists and is also writeable. But on many other environments it might not exist. 

I'd be interested to know why you were choosing to use the install script...

iparr:
I read this page like 3 times. I thought that it was "Step 1: Install using Homebrew" then "Step 2: Setup using the install script." But it's not is it? It's one or t'other. Pretty embarrassing. I had no such problems when I installed on Linux where I'm prepared for multiple steps with everything. I wonder if anyone else has made the same mistake? I may have had a wine but I'm not sure.

rfay:
Well, everybody has a different experience with docs!  Thanks so much for taking the time to share your experience and try to fix it for others.

It does look to me like the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) will fail if /usr/local/bin doesn't exist, so that's what happened to you? It would be gentler the script did this:

```bash
if [ ! -d /usr/local/bin ]; then sudo mkdir -p /usr/local/bin; fi
```

and I don't know a reason not to. 

I guess the first thing I do with a mac is always set up homebrew, so I didn't even know that /usr/local/bin wasn't there by default. And of course on an M1 mac, /usr/local/bin wouldn't be created by homebrew because it's in /opt/homebrew there. 

I do think that the doc in https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#macos could be improved by saying "An alternate way to install is the install script..." Want to try that instead? Interested in giving a go at a PR on the install script?

rfay:
Thanks so much for helping us understand @iparr , please see if 
* https://github.com/ddev/ddev/pull/4752

would have solved it for you.

PR #4752, opened 2023-03-15
Improve install_ddev.sh by making it create /usr/local/bin, replaces #4747 [skip ci]

rfay:
## The Issue

* #4747

1. If /usr/local/bin doesn't exist, the install_ddev.sh fails. In M1 macs, that directory doesn't necessarily exist, although it has always existed almost everywhere else.
2. The docs may not make clear that the script installation is an *alternate* 

* Add creation of /usr/local/bin to the script
* Minor tweak to install doc

1. Temporarily remove /usr/local/bin, `sudo mv /usr/local/bin /usr/local/bin.bak`
2. Run the updated script from this PR, 
`curl -fsSL https://raw.githubusercontent.com/rfay/ddev/90f9d72e02591df1fd1fcc75fa5401bad6aa4113/scripts/install_ddev.sh
 | bash `
3. Put your system back together, `sudo mv /usr/local/bin /usr/local/bin.test && sudo mv /usr/local/bin.bak /usr/local/bin`

## Automated Testing Overview

No testing is done on this.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4752"></a>

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -31,7 +31,9 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     !!!tip
         The install script works on macOS, Linux, and Windows WSL2.
 
-    Run the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) to install or update DDEV. It downloads, verifies, and sets up the `ddev` binary:
+    The [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) is an alternate way to install or upgrade DDEV. It downloads, verifies, and sets up the `ddev` binary.
+
+    To install or update DDEV:
 
     ```
     curl -fsSL https://ddev.com/install.sh | bash
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -10,7 +10,7 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
 
     ### Homebrew
 
-    [Homebrew](https://brew.sh/) is the easiest way to install and upgrade DDEV:
+    We recommend [Homebrew](https://brew.sh/) because it’s the easiest and most reliable way to install and upgrade DDEV:
 
     ```bash
     brew install ddev/ddev/ddev
@@ -30,8 +30,6 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
 
     !!!tip
         The install script works on macOS, Linux, and Windows WSL2.
-
-    We recommend Homebrew because it’s the easiest and most reliable way to install and upgrade DDEV.
 
     Run the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) to install or update DDEV. It downloads, verifies, and sets up the `ddev` binary:
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -31,6 +31,7 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     !!!tip
         The install script works on macOS, Linux, and Windows WSL2.
 
+    We recommend Homebrew for installation on macOS, but the install script is an alternate technique.
 
     Run the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) to install or update DDEV. It downloads, verifies, and sets up the `ddev` binary:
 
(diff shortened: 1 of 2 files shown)


docs/content/users/install/ddev-installation.md
```diff
@@ -31,6 +31,7 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     !!!tip
         The install script works on macOS, Linux, and Windows WSL2.
 
+    We recommend Homebrew for installation on macOS, but the install script is an alternate technique.
```
mattstein, line 34: Does this add anything that couldn’t be solved by strengthening the Homebrew intro?

This could explain _why_ we recommend it, in a more appropriate place, without yanking someone around if they passed on Homebrew and were more interested in the script installer.

> We recommend Homebrew because it’s the easiest and most reliable way to install and upgrade DDEV:

If the script installer comes with caveats, it’d be worth calling them out here so someone can make an informed decision. If the script installer is more risky than Homebrew or it’s not a good way to go, maybe it shouldn’t be an option?

mattstein: As it stands, this reads to me like “You’ve found the install script we pointed out, but we recommend Homebrew so proceed at your own peril!” (Why bother with this “alternate technique”?)

rfay: There are people who don't like homebrew. The script installer doesn't so much have caveats, but it's nonstandard in every environment.  Most professional developers will want to use the "standard" package manager in their environment instead of some random script. 

Supporting less permutations is fine, but it's hard to accomplish I guess. On Linux we support apt, homebrew, and install_ddev.sh :( All work fine. But at least we don't mention so many of them any more.

mattstein: I understand not liking or preferring Homebrew, this sentence just doesn’t seem to leave the reader more informed. “We recommend Homebrew and this is another technique” is kind of redundant, hence my suggestion of either informing someone of the caveats or strengthening the Homebrew recommendation in that section.

rfay: Peril! As always, suggestion is welcome now that we've explored the territory. DDEV's installation is so simple that everything works. The only peril is when people can't remember how they installed, etc. `ddev self-upgrade` tries to address this now, but it's pretty obscure.

```
$ ddev self-upgrade
/opt/homebrew/bin/brew
DDEV appears to have been installed with homebrew, upgrade with 'brew update && brew upgrade ddev'
```

mattstein: I started [with a specific suggestion](https://github.com/ddev/ddev/pull/4752#discussion_r1137524521) to omit this sentence and strengthen the Homebrew recommendation, with a smidge of justification, in the section where it’s relevant.

The `self-upgrade` warning is lovely, even if it’s obscure.

rfay: Sorry, your specific suggestions are usually code suggestions. So I guess my eyes saw it as discussion.

rfay:
```suggestion
    We recommend Homebrew because it’s the easiest and most reliable way to install and upgrade DDEV.
```

mattstein: You were right! I wasn’t sure if I was missing the point you were trying to make with that sentence, either. I should’ve started with a code suggestion though, sorry.

mattstein: @rfay What you applied wasn’t my suggestion though. I’d get _rid of_ this sentence you added, and use the line I suggested to [start the Homebrew section](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#homebrew)—not the Install Script section.

mattstein: I can’t figure out how to make this a code suggestion, but this is what it’d end up being:

```markdown
    ## macOS

    ### Homebrew

    We recommend [Homebrew](https://brew.sh/) because it’s the easiest and most reliable way to install and upgrade DDEV:

    ```bash
    brew install ddev/ddev/ddev
    ```

    ```bash
    brew upgrade ddev
    ```

    As a one-time initialization, run

    ```bash
    mkcert -install
    ```

    ### Install Script

    !!!tip
        The install script works on macOS, Linux, and Windows WSL2.

    Run the [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) to install or update DDEV. It downloads, verifies, and sets up the `ddev` binary:
```

mattstein: Done! I don’t know yet how to PR a PR, so I pushed a commit to your fork.

rfay: Thank you for doing that, I would have gotten to it. But in the future, it's super easy. You just fork the rfay/ddev repo (in this case) (or just add an extra git remote to it) and then make a PR based against rfay/20230315_install_script

A bit of a mess, but super easy to do.

rfay:
Tested this manually by `mv /usr/local/bin /usr/local/bin.bak` and then running it; it behaved as expected. Also worked in the traditional situation.

PR #4763, opened 2023-03-17
[docs] Rescue intentional line breaks

mattstein:
## The Issue

`.editorconfig` improvements in #4712 strip trailing spaces, which in Markdown removes two-space line endings (`  `) than translate into `<br>` elements.

This removes the problematic `trim_trailing_whitespace = true` setting, re-adds two-space line endings that have since been stripped out of Markdown files, and fixes some other mangled spacing I noticed along the way.

The result allows Markdown to continue operating like someone would expect, without needing to resort to using `<br>` elements all over the place or unwittingly having their `  ` breaks stripped on save.

> **Note** 
> After further discussion, this PR changed from originally inserting `<br>` elements to reverting `trim_trailing_whitespace = true`.

[Preview changes locally](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) or inspect [the automatic build](https://ddev--4763.org.readthedocs.build/en/4763/).

## Automated Testing Overview

n/a

## Related Issue Link(s)

- #4745 
- #4712

## Release/Deployment Notes

n/a

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4763"></a>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,7 +12,6 @@ indent_style = space
 # embedded yaml in markdown.
 # indent_size = 4
 insert_final_newline = true
-trim_trailing_whitespace = true
 
 # GO files
 [*.go]
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,9 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 4
+# Forcing indent to 4 seems to break lots of things, like
+# embedded yaml in markdown.
+# indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,19 +26,3 @@ indent_style = tab
 [*.{nsi,nsh}]
 end_of_line = crlf
 indent_size = 2
-
-# YAML files
-[*.{yaml,yml}]
-indent_size = 2
-
-# JavaScript, TypeScript, and JSON
-[*.{json,jsonc,js,ts}]
-indent_size = 2
-
-# styles
-[*.{css,sass,less,pcss,scss}]
-indent_size = 2
-
-# shell scripts
-[*.sh]
-indent_size = 2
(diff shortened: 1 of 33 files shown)


rfay:
Sorry if I'm being dense, but if trim_trailing_whitespace makes us dance around with `<br>` everywhere, and you have to know magic knowledge to accomplish that, then let's remove the trim_trailing_whitespace instead. 

Normal markdown treats double linefeeds with respect. We'd rather have "normal" markdown. 

Am I lost still?

mattstein:
I brought this up [here](https://github.com/ddev/ddev/pull/4739#issuecomment-1464662209), reiterated again why `trim_trailing_whitespace` is useful [here](https://github.com/ddev/ddev/pull/4745#issuecomment-1465268499), and you suggested a PR to fix the Markdown and keep `trim_trailing_whitespace` [here](https://github.com/ddev/ddev/pull/4745#issuecomment-1466339250).

This is that PR.

The `trim_trailing_whitespace` setting [is still enabled](https://github.com/ddev/ddev/blob/master/.editorconfig#L15) and those line breaks are stripped out. I don’t care what happens with the setting, I’m just trying to save the line breaks I added on purpose and I don’t want to keep circling the issue of what we do with that setting.

If you want me to drop `trim_trailing_whitespace` and turn those `<br>`’s back into double spaces again I can do that instead.

rfay:
I'm the one that thinks trailing whitespaces do no harm. I apologize about this confusion, but I never thought this was a problem that needed solving. I guess we all think we're not being listened to. Do you think the trailing spaces is even a problem that needs solving? Does anybody but @gilbertsoft think it's a problem that needs solving?

mattstein:
> Do you think the trailing spaces is even a problem that needs solving?

Not in Markdown, no. But I can’t speak for the whole codebase or @gilbertsoft’s intentions—just defending my precious line breaks.

mattstein:
@rfay Updated my approach here along with the description. Ready to review again.

rfay (approved):
Can we feel the pain receding?

PR #4765, opened 2023-03-18
Improve tableplus cmd

riconeitzel:
@rfay if that works, we might merge?

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4765"></a>


rfay:
Please, let's just wait for reopening and pulling
* https://github.com/ddev/ddev/pull/4692

It shouldn't be long to wait for stable release, right?

PR #4769, opened 2023-03-20
Shopware nginx config: Allow Shopware Installer to run

shyim:
## The Issue

Nginx by default cannot route `foo.phar.php/foo` to the actual file `foo.phar.php`. It thinks foo.phar.php is a folder. 
This is not required to be applied on Apache2 or Caddy as it just works 

Use location to forward `shopware-installer.phar.php` always to that actual file.

See also https://github.com/hetzneronline/community-content/pull/569

- Download the Installer [here](https://github.com/shopware/web-installer/releases/)
- Drop it in a document root
- Open it and see that next pages does not load or CSS/JS is missing

## Automated Testing Overview

You could request the CSS/JS inside

## Related Issue Link(s)

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4769"></a>

### DIFF
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-shopware6.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-shopware6.conf
@@ -50,7 +50,7 @@ server {
     }
 
     # pass the PHP scripts to FastCGI server listening on socket
-    location ~ \.php$ {
+    location ~ ^/(index|shopware-installer\.phar)\.php(/|$) {
         try_files $uri =404;
         fastcgi_split_path_info ^(.+\.php)(/.+)$;
         fastcgi_pass unix:/run/php-fpm.sock;


rfay (commented):
Thanks! Can you please update the OP to explain the practical result of this, and why you kindly took the time to make the change?

shyim:
Hey,

What do you mean with OP?
Yes many hosting providers are also using Apache (I personally prefer Caddy).  But with Apache is no change needed while running phar files in the Web.

rfay:
Sorry, "OP" is your "original post" above. 

So you don't have any problem with shopware installer in ddev with the default apache config that DDEV provides if you set `webserver_type: apache-fpm` ?

Please also add to the original post manual testing instructions; you removed that part of the template.

Please make sure that you have manually tested this and make sure it does what you want.  You can download artifacts with your version at https://github.com/ddev/ddev/pull/4769#issuecomment-1476382243 - there are more instructions at https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr

shyim:
Sorry I updated the PR template :)

I personally don't use ddev, just looked how I can fix nginx with the new shopware-installer. I took the approach how the contao-manager is handled in Nginx. 

The user didn't had any problem with using `apache-fpm` and it worked out of the box. I asked him to test the build binary (I don't want to install Docker Desktop. Just here to adjust the nginx config of Shopware :D )

rfay:
Thanks so much for taking care of things, and for making the PR. 

* You don't need Docker Desktop to use DDEV. I don't know what environment you're on, but it's not required in any environment. 
* Unless somebody can demonstrate the problem and demonstrate that this fixes it, it won't be accepted. 

Thanks!

rfay:
Happy to return to this, but if nobody is actually knowledgable about why it's being done or testing the results, we'll wait until that person comes along. Happy to continue the conversation here or elsewhere. Just say if it should be reopened.

PR #4775, opened 2023-03-21
Overwrite copies of embedded files if they're empty on target filesystem

rfay:
## The Issue

Users have occasionally reported having empty shell commands like `launch` on their system (~/.ddev/commands/host/launch). It's unknown how this happens, perhaps no space available.

https://stackoverflow.com/questions/75801233/problems-with-ddev-after-upgrade-to-1-21-6-launch-file-empty

On empty file, overwrite it. This doesn't solve the original problem, but will fix on `ddev restart` if there's available disk space

`rm ~/.ddev/commands/host/launch && touch ~/.ddev/commands/host/launch && ddev start` - `ddev launch` should work again.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4775"></a>

### DIFF
--- a/pkg/fileutil/embed.go
+++ b/pkg/fileutil/embed.go
@@ -24,8 +24,11 @@ func CopyEmbedAssets(fsys embed.FS, sourceDir string, targetDir string) error {
 		} else {
 			localPath := filepath.Join(targetDir, d.Name())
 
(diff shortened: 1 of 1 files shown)


rfay (commented):
Manually tested again and it seems fine.

PR #4781, opened 2023-03-22
Don't override database prefix in settings.ddev.php

rfay:
## The Issue

In Drupal settings.ddev.php we set the database prefix to ''. This overrides what might be set in the settings.php, and there's no need for it.

Stop setting the database prefix at all.

Set database prefix in settings.php before the include of settings.ddev.php. See if things work with a project that has a prefix in the db.

## Automated Testing Overview

No changes.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4781"></a>

### DIFF
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -23,7 +23,6 @@ type BackdropSettings struct {
 	DatabaseHost     string
 	DatabaseDriver   string
 	DatabasePort     string
-	DatabasePrefix   string
 	HashSalt         string
 	Signature        string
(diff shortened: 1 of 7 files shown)


rfay:
A workaround for people who might run into this... 

If you browse your database and see that all the tables have a prefix (like "xxx_" or "drup_" then add this to the bottom of your settings.php (below the settings.ddev.php import) or to a settings.local.php (also below settings.ddev.php):

```
$databases["default"]["default"]["prefix"] = "xxx_";
```

PR #4788, opened 2023-03-29
Bump go-dockerclient for security issue, replaces #4787

rfay:
## The Issue

#4787 points out a security problem with runc, which we don't use, but go-dockerclient pulls in. Update go-dockerclient. vendor only.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4788"></a>


mattstein (approved):

rfay:
I guess it pinged you because of markdown files in vendor @mattstein ?

mattstein:
@rfay Yep! Double-checked to confirm that was the case and that’s why I approved it without reading the contents since those aren’t our words.

rfay:
I insist that you read every single change in this PR before approving it :)

mattstein:
Oh. Uh ... I did? I did. Yeah I was just joking, I totally read all the Markdown and it was exquisite.

PR #4789, opened 2023-03-29
[docs] Refactor install and upgrade instructions

mattstein:
## The Issue

#4722 pointed out how easy it is to miss the upgrading instructions on the [DDEV Installation](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/) page. We discussed ways of calling attention to a note, which doesn’t address underlying problems:

1. There are lots of ways to install DDEV, which is good for the world and challenging for instructions.
2. Each OS is split into different approaches that have vastly different amounts of explanation.
3. Different install paths reference one another and are not consistent in what they help you accomplish. (Choosing a specific version, for example, or clarifying how to upgrade.)
4. Some details can get lost if you expect to choose a path and not read more broadly. (If I install on macOS with [Homebrew](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#homebrew), for example, it can still be easy to miss the Mutagen recommendation because it looks like it’s part of the install script instructions.)

This PR combines restructuring and edits to solve those problems:

1. **Separates installation and upgrading into separate pages**, allowing each to be more focused:  
    Screenshot of documentation with emphasis on the sidebar, where “DDEV Installation” has a new “Upgrading” page underneath it
2. **Prunes wording and removes comments** that are a barrier to the installation task.  
Example: the reader doesn’t benefit from knowing that the install script or Homebrew work on different OS arrangements. They should see clear options and instructions on the tab they’ve chosen without having to read through everything.
3. **Edits each set of instructions (OS + approach) to be self-contained**.  
This introduces some duplicate information for the writer, but it makes it possible to strip out unnecessary explanation and focus the experience for the reader.
4. **Introduces mkdocs-material’s [Collapsible Blocks](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#collapsible-blocks)** to reduce the visual impact of less-common admonitions without removing them:
    <video src="https://user-images.githubusercontent.com/2488775/228668951-b9122b48-ded7-4cac-9ca4-6dc2ddb9571b.mp4" data-canonical-src="https://user-images.githubusercontent.com/2488775/228668951-b9122b48-ded7-4cac-9ca4-6dc2ddb9571b.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px" width="600"></video>

Here’s a before and after of the macOS install instructions, which are easier to follow when they’re separated from the upgrading instructions:

install-before-after

Here’s a comparison of what the reader _has to pay attention to_ in order to update on macOS with Homebrew, before and after the PR:

upgrade-before-after

[Preview changes locally](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) or inspect the [automatic build](https://ddev--4789.org.readthedocs.build/en/4789/users/install/ddev-installation/).

I recommend comparing the proposed [installation page](https://ddev--4789.org.readthedocs.build/en/4789/users/install/ddev-installation/) to the [current one](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/). Just like the newly-added [upgrading](https://ddev.readthedocs.io/en/latest/users/install/ddev-upgrade/) page, it should feel more focused and easier to follow.

I tested the updated Gitpod install instructions along with the new _updating_ instructions for Gitpod and Codespaces, and everything worked as written.

## Automated Testing Overview

n/a

## Related Issue Link(s)

#4722

## Release/Deployment Notes

~~If it’s possible to redirect URL fragments, I’ll need to do that here.~~ Doesn’t look like it’ll affect any current link fragments, so there shouldn’t be a release impact.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4789"></a>


mattstein:
@rfay While this is a draft, I still have questions for you:

1. How _should_ people upgrade DDEV in a GitHub codespace?
2. Is [this Fedora signed repository note](https://github.com/ddev/ddev/blob/master/docs/content/users/install/ddev-installation.md?plain=1#L91) still relevant?
3. Should we clarify how to install a specific DDEV version everywhere (Homebrew, Gitpod, etc.) like we happen to with the script installer?

mattstein:
@bencroker, if you have any thoughts on this approach I’d love to hear them! (Also anyone else following along, of course.)

rfay:
Thanks @mattstein !

1. Upgrading ddev in codespaces should be the same as linux anywhere, `sudo apt update && sudo apt upgrade` or `sudo apt upgrade ddev`
2. Unfortunately the Fedora signed repository note is still valid. They haven't extended the support to rpm repos yet.
3. I didn't understand this "Should we clarify how to install a specific DDEV version everywhere (Homebrew, Gitpod, etc.) like we happen to with the script installer" - maybe you're asking if we should mention the script everywhere?

rfay:
> But we demonstrate using [Homebrew](https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#gitpod) in Gitpod so that’d be brew upgrade ddev, right?

I don't think we should be using homebrew in the gitpod instructions any more. Should all be apt. (Not that homebrew doesn't work anywhere on amd64 linux, but it's not the "native" technique, and we support the native technique, apt)

> And it’s safe to upgrade with apt for codespaces because the ghcr.io/ddev/ddev/install-ddev:latest [YAML addition](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#github-codespaces) ultimately uses apt under the hood or it’ll work fine?

Agreed. I don't have a pile of experience with Codespaces, but am pretty sure this is great.

> We only point that out for the script installer. But what if I use Homebrew or apt or yum and want to install a specific DDEV version?

Our apt repo will allow people to specify another version, but I haven't experimented with it.  Example resource, https://itsfoss.com/apt-install-specific-version/, which says `apt list --all-versions package_name` so `apt list --all-versions ddev`. I'm horrified to see that there are piles of experimental versions there. Then `sudo apt install ddev=1.21.3` works fine. 

dnf instructions, https://linuxopsys.com/topics/install-specific-version-of-package-using-dnf - `dnf --showduplicates list ddev` does show all versions since we started publishing.  Then `sudo dnf install ddev-1.21.2-1` for example. Seems to all work

It is not currently possible to install a specific version using homebrew; they took that capability away a year or so ago. It would be possible for us to use a new recipe and allow a separate channel.

In general, I prefer for people who want to do exotic things with their exotic systems know how to do it themselves, and giving them poor summaries can do a disservice. OTOH, of course, they *don't* actually know how to use their own OS. 

I guess: Do you think we can communicate this without making a mess and making things worse?

mattstein:
> I guess: Do you think we can communicate this without making a mess and making things worse?

No I don’t. Sounds like it’s better to keep things as they are instead of add complexity that doesn’t benefit most people.

I’ll update the Gitpod + codespaces instructions in this PR and make sure they still work. Thanks for all your feedback on this!

rfay:
It would be OK to say "You can also install a specific version of DDEV using the standard apt techniques on Debian/Ubuntu" and give a link. Same with dnf.

mattstein:
Ready for review! I don’t love that the `apt` install instructions get longer and touch more stuff on the system, and also take longer to upgrade, but I tested and it works as expected with Gitpod and Codespaces.

docs/content/users/install/ddev-installation.md
```diff
@@ -2,50 +2,44 @@
...
+    ```bash
+    # Download and run the install script
     curl -fsSL https://ddev.com/install.sh | bash
-    ```
-
-    You can include a `-s <version>` argument to install a specific release or a prerelease version:
 
-    ```
-    curl -fsSL https://ddev.com/install.sh | bash -s v1.21.4
+    # Optionally enable Mutagen for the best performance (recommended!)
```
rfay, line 32: It looks like this mutagen statement got migrated to just the install script section?

docs/content/users/install/ddev-upgrade.md
```diff
@@ -0,0 +1,117 @@
...
+    ```bash
+    # Upgrade the DDEV package
+    yay -Syu ddev-bin
+    ```
+
+=== "Windows"
+
+    ## Windows
+
+    ### WSL2 + Docker Install Script
```
rfay, line 60: This section is not correct. For all WSL2, the upgrade technique is same as Debian/Ubuntu. Just open Ubuntu and `sudo apt update && sudo apt upgrade ddev` - no mucking about with PowerShell or anything.

rfay (changes requested):
It's brilliant, and the new Collapsible Blocks approach is fantastic. 

There are a couple of things to go, it's easy to get lost in all these changes. But the upgrade of ddev for all WSL2 Ubuntu installation methods is `sudo apt update && sudo apt upgrade`

docs/content/users/install/ddev-installation.md
```diff
@@ -250,8 +270,6 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
     * We recommend using [Chocolatey](https://chocolatey.org/). Once installed, you can run `choco install ddev docker-desktop git` from an administrative shell. You can upgrade by running `ddev poweroff && choco upgrade ddev`.
     * Each [DDEV release](https://github.com/ddev/ddev/releases) includes a Windows installer (`ddev_windows_installer.<version>.exe`). After running that, you can open a new Git Bash, PowerShell, or cmd.exe window and start using DDEV.
 
-    Most traditional Windows users will want to enable Mutagen for superb performance; no installation is required; run `ddev config global --mutagen-enabled`. It still won't be as fast as one of the WSL2 options.
```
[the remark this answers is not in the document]
mattstein: I removed it because it’s redundant; there’s already a note [leading these instructions](https://github.com/ddev/ddev/blob/7e5643e81646becf20bbf3434bf848d89f224898/docs/content/users/install/ddev-installation.md?plain=1#L268).

docs/content/users/install/ddev-installation.md
mattstein: 🤔 I’m not following. It [currently](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#install-script) looks like it might only apply to the macOS install script section, when it applies equally to both methods.

I attempted to clarify that by more concisely recommending it in both cases: [here](https://github.com/ddev/ddev/blob/7e5643e81646becf20bbf3434bf848d89f224898/docs/content/users/install/ddev-installation.md?plain=1#L20-L21) and [here](https://github.com/ddev/ddev/blob/7e5643e81646becf20bbf3434bf848d89f224898/docs/content/users/install/ddev-installation.md?plain=1#L32-L33).

I also deliberately left it out of the Linux install script instructions [here](https://github.com/ddev/ddev/blob/7e5643e81646becf20bbf3434bf848d89f224898/docs/content/users/install/ddev-installation.md?plain=1#L116-L117), under the assumption that Mutagen [doesn’t offer a significant enough benefit on Linux](https://ddev.readthedocs.io/en/stable/users/install/performance/#what-mutagen-does).

Am I missing your point?

docs/content/users/install/ddev-upgrade.md
mattstein: I’m not sure how I managed to get that so wrong!

> But the upgrade of ddev for all WSL2 Ubuntu installation methods is `sudo apt update && sudo apt upgrade`

Your comment above is slightly different. Which should it be?

1. `sudo apt update && sudo apt upgrade`
2. `sudo apt update && sudo apt upgrade ddev`

docs/content/users/install/ddev-installation.md
rfay: OK, if it's being recommended elsewhere I'm good, just couldn't understand in the diff. 

And yes, it is not normally used/useful in Linux or WSL2, although does no harm there.

mattstein: With this PR we now speak directly and specifically to the Linux user instead of punting to the macOS section, so it felt right to remove that part and align with the Linux notes on the Mutagen page.

Sounds like we’re all good on this front!

docs/content/users/install/ddev-upgrade.md
mattstein: So you’re recommending `sudo apt upgrade -y`—are you deliberately leaving out `sudo apt update &&`?

Or is the full and “final” answer `sudo apt update && sudo apt upgrade -y`?

rfay: `sudo apt update && sudo apt upgrade -y` is the full and final answer, thanks.

mattstein: Fixed in d2a35731c2e639edf603b454425c00653a932524, which is refreshingly more concise!

mattstein:
Actually seems fine, so I re-ran the action.

mattstein:
@rfay Buildkite is apparently not a fan of the PR, but if you are please merge whenever you’d like!

rfay:
Thanks. I'll try to take a careful read-through of the rendered docs first. Much appreciated!

rfay (approved):
Read through it all. Thanks for the herculean effort both in thinking/designing and implementing.

PR #4790, opened 2023-03-29
[docs] Update credsStore to Colima to avoid DDEV start errors

chri5tia:
## The Issue
When switching from Docker Desktop and installing colima and docker via homebrew, following this documentation here, I ran into the following issue when running `ddev start` on my project.

```
', stderr='failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: rpc error: code = Unknown desc = error getting credentials - err: docker-credential-desktop resolves to executable in current directory (./docker-credential-desktop), out: ``' 
```

DDev was still looking for a docker desktop file, despite it being fully uninstalled.

This PR updates documentation to provide a configuration tip when migrating from Docker Desktop to Colima. 
The `.docker/config.json` should to be updated to:

```
{
        "auths": {},
        "credsStore": "colima",
        "currentContext": "colima"
}
```

PR does not change code; testing should follow the steps added to documentation for installation.

## Related Issue Link(s)
https://stackoverflow.com/questions/67642620/docker-credential-desktop-not-installed-or-not-available-in-path#comment132716686_68202428

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4790"></a>

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -39,11 +39,8 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     #### Migrating Projects from Docker Desktop to Colima
 
-    !!!tip "Docker Desktop may have left a bad config.json"
-        Remove the `credsStore` line in `~/.docker/config.json` if you have trouble.
-
-    1. Open Docker's `config.json` file (`~/.docker/config.json`)
-    2. Remove the `credsStore` line.
+    !!!tip "Docker Desktop may have left a bad `config.json`"
+        Remove the `credsStore` line in `~/.docker/config.json` if you have trouble running `ddev start` with a project you’ve migrated.
 
     Move your project databases from Docker Desktop to Colima:
 
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -39,6 +39,12 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     #### Migrating Projects from Docker Desktop to Colima
 
+    !!!tip "Change credential storage from Dev Desktop to Colima"
+        Update credential storage to Colima to avoid ddev start errors
+
+    1. Open to edit Docker's `config.json` file (ex: `~/.docker/config.json`)
+    2. Change the `credsStore` value to `colima`
+
     Move your project databases from Docker Desktop to Colima:
 
     1. Make sure all your projects are listed in [`ddev list`](../usage/commands.md#list).
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -39,11 +39,11 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     #### Migrating Projects from Docker Desktop to Colima
 
-    !!!tip "Change credential storage from Dev Desktop to Colima"
-        Update credential storage to Colima to avoid ddev start errors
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
It's easier to just delete ~/.docker. What do you think about that?

rfay:
I'm glad you brought this up!

chri5tia:
> It's easier to just delete ~/.docker. What do you think about that?

I had completely uninstalled and removed docker desktop. It's possible that this directory was still on the system. Since I ran into this after installing docker with brew as part of setting it up with Colima it would be too late at that point to delete the the whole directory, wouldn't it? I assume it was a default value which wasn't set during the Colima set up process. Not sure where to go from here.

rfay:
Yeah, Docker Desktop doesn't remove lots of things, and ~/.docker is one of them. 

I have never heard of anybody needing the contents of ~/.docker when switching away from Docker Desktop. 

Note that it's not necessary to uninstall Docker Desktop to use colima, and the docker client provided with Docker Desktop works fine with colima.

So I think it would be useful to suggest removing ~/.docker/config.json if uninstalling Docker Desktop.

rfay:
The suggestion in Stack Overflow isn't even correct; you can't just change to ` "credsStore": "colima",`, there isn't a credStore AFAIK, and the name would be the name of a profile. 

I would recommend 1) Removing docker.json in case of trouble or 2) Remove the credsStore line.  Why don't we just update this PR to say "Remove the credsStore line".

mattstein:
> I would recommend 1) Removing docker.json in case of trouble or 2) Remove the credsStore line. Why don't we just update this PR to say "Remove the credsStore line".

Sounds more specific and accurate, so great AFAIK!

rfay:
Made the suggested changes.

docs/content/users/install/docker-installation.md
```diff
@@ -39,6 +39,12 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     #### Migrating Projects from Docker Desktop to Colima
 
+    !!!tip "Docker Desktop may have left a bad config.json"
```
mattstein, line 42: Wrap filename in backticks:

```suggestion
    !!!tip "Docker Desktop may have left a bad `config.json`"
```

docs/content/users/install/docker-installation.md
```diff
@@ -39,6 +39,12 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+        Remove the `credsStore` line in `~/.docker/config.json` if you have trouble.
```
mattstein, line 43: Qualify what kind of trouble makes this relevant so it’s less broad and open-ended:

```suggestion
        Remove the `credsStore` line in `~/.docker/config.json` if you have trouble running `ddev start` with a project you’ve migrated.
```

docs/content/users/install/docker-installation.md
```diff
@@ -39,6 +39,12 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+
+    1. Open Docker's `config.json` file (`~/.docker/config.json`)
+    2. Remove the `credsStore` line.
```
mattstein, line 46: I’m not sure why these steps repeat the exact same instruction in a numbered list format outside the admonition, so it seems best to remove them:

```suggestion
```

mattstein (changes requested):

PR #4798, opened 2023-04-01
[docs] Improve colima disk allocation methods to show new increase capability

rfay:
## The Issue

Colima 0.5.4 has the ability to increase disk size.

Tell people about it.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4798"></a>

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -28,7 +28,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     When your computer restarts, you’ll need to `colima start` again. This will eventually be automated in later versions of Colima.
 
     !!!tip "Colima disk allocation"
-        We recommend allocating lots of storage for Colima because there’s no easy way to increase the size later (see [GitHub comment](https://github.com/abiosoft/colima/issues/398#issuecomment-1449801811) for a step-by-step guide to resize disk images with `qemu`). You can reduce usage by running `ddev clean`, and killing off disk images by running `docker rm -f $(docker ps -aq) && docker rmi -f $(docker images -q)`. If you have to rebuild your Colima instance, use the technique described below for migrating from Docker Desktop.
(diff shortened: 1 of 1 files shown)


docs/content/users/install/docker-installation.md
```diff
@@ -28,7 +28,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     When your computer restarts, you’ll need to `colima start` again. This will eventually be automated in later versions of Colima.
 
     !!!tip "Colima disk allocation"
-        We recommend allocating lots of storage for Colima because there’s no easy way to increase the size later (see [GitHub comment](https://github.com/abiosoft/colima/issues/398#issuecomment-1449801811) for a step-by-step guide to resize disk images with `qemu`). You can reduce usage by running `ddev clean`, and killing off disk images by running `docker rm -f $(docker ps -aq) && docker rmi -f $(docker images -q)`. If you have to rebuild your Colima instance, use the technique described below for migrating from Docker Desktop.
+        In Colima versions starting with 0.5.4 you can increase the disk allocation (but not decrease it) by editing the `~/.colima/default/colima.yaml` to change the `disk` setting to a higher value. For example, `disk: 200` will increase allocation to 200 gigabytes. Then `colima restart` will result in the new disk allocation.
```
mattstein, line 31: I’d save a word and keep the increasing/decreasing ideas closer to one another to maybe improve the flow. And get rid of “the,” since it should either be “editing the X file” or “editing X”: 

```suggestion
        In Colima versions starting with 0.5.4 you can increase—but not decrease—the disk allocation by editing `~/.colima/default/colima.yaml` to change the `disk` setting to a higher value. For example, `disk: 200` will increase allocation to 200 gigabytes. Then `colima restart` will result in the new disk allocation.
```

mattstein (approved):
Looks good! Left a soft suggestion, but not a big deal.

PR #4810, opened 2023-04-13
[tests only] Fix broken TestDdevExec on CircleCI arm64

rfay:
## The Issue

CircleCI arm64 test has been failing on TestDdevExec

See if Ubuntu 22.04 fixes it.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4810"></a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   build:
     machine:
-      image: ubuntu-2004:current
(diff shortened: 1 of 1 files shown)


rfay:
Later runs on upstream/master seemed to succeed, so this must have been a CircleCI problem? Anyway, if re-run of the test here works, it's worth upgrading to Ubuntu 22.04.

PR #4811, opened 2023-04-14
Use env.default instead of env.example

ptmkenny:
The docs should specify `env.default` for the value-less environmental variables file as this is a more standard name and `env.example` is gitignored by ddev by default, and it doesn't make sense that the user should have to override the gitignore to follow the instructions in the docs.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4811"></a>

### DIFF
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -159,7 +159,7 @@ ddev config global --web-environment-add="MY_ENV_VAR=someval
 You can use the `--web-environment` flag to overwrite existing values rather than adding them.
 
 !!!warning "Don’t check in sensitive values!"
-    Sensitive variables like API keys should not be checked in with your project. Typically you might use an `.env` file and _not_ check that in, but offer `.env.example` with expected keys that don’t have values. Some use global configuration for sensitive values, as that’s not normally checked in either.
+    Sensitive variables like API keys should not be checked in with your project. Typically you might use an `.env` file and _not_ check that in, but offer `.env.default` with expected keys that don’t have values. Some use global configuration for sensitive values, as that’s not normally checked in either.
 
 ### Altering the In-Container `$PATH`
 


tyler36:
Not sure I agree with changing it to `.env.default`.

A quick google search for

- ".env.example" shows 229,000 results
- "env.default" shows 29,100 results (90% lower)

Laravel uses `.env.example` in their repos.

ptmkenny:
@tyler36 Hmm, ok.  But the problem with `.env.example` is that it is ignored by ddev.

Example of a real problem/reason for this PR: I deleted the `.ddev` directory and then checked it out again from git, and then rebuilt my site.  The rebuild failed because the environmental variables were missing. It took me some time to identify the issue because when I grepped my codebase for the variables, none of the env files turned up because I had originally named my file `.env.example` as stated in the docs and then it never got committed, so it was removed when I deleted `.ddev`.

tyler36:
@ptmkenny 

That sounds terrible. I did a similar thing last week so I know the pain.

I think `.env.example` _should_ be in the repo because it is a guide to what environmental variable are required to get the project functioning. But I don't think it should be initialized by DDEV; it should serve as a guide for required values only.

If there's env stuff that you absolutely require and don't mind exposing, I think the `.ddev/config.yaml` file might be the better way to "hard code" envs into your project. Eg. 

```yaml
web_environment:
 - VITE_SERVER_PORT=3000
```

ptmkenny:
@tyler36 Thanks.  In my case, they are encryption keys, so I don't want them in the repo.

The issue is this: the docs currently recommend using `.ddev/.env.example` for the reference file. Because this file is merely a guide to the environmental variables that need to be set, but doesn't actually set them, that implies it should be committed to the repo.  But the ddev `.gitignore` includes `/**/*.example`, so the file can't be committed without `-f`.

So I proposed renaming the file to `.env.default` so that it could be committed without `-f`.  An alternative would be to ad a note that says something like "be sure to commit this file if you create it because ddev will ignore it by default."  Thoughts?

rfay:
The normal case is `.env` of course, which normally wouldn't be checked in, right? 

I think it would probably be best to open an issue to discuss the problem and the solution to it, rather than just come up with one solution.

Maybe you can explain the problem in an issue and then offer a variety of solutions to it, and some conversation will get somewhere?

Closing this for now, as it needs more discussion. There's nothing wrong with `git add -f` either... And everybody should know how to use it.

PR #4832, opened 2023-04-18
Changed network.go to support the DDEV_GITHUB_TOKEN in DownloadFile

TsPG128:
## The Issue

The get command is actually already supporting the DDEV_GITHUB_TOKEN (see https://github.com/ddev/ddev/blob/master/cmd/ddev/cmd/get.go#L371 ) though for purposes of the github download limit. 

But while the get function passes through the initial github requests where the releases of the repo are detected, it breaks when trying to download the packaged release:

```
$ DDEV_GITHUB_TOKEN=github-token ddev get name/of-addon

Unable to download name/of-addon: Unable to download https://api.github.com/repos/name/of-addon/tarball/1.0.0: download link https://api.github.com/repos/name/of-addon/tarball/1.0.0 returned wrong status code: got 404 want 200

```

I changed the http request method to support the bearer authentication for downloading the tarball

Using the DDEV get command on a private github repository will crash regardless of the set DDEV_GITHUB_TOKEN variable.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4832"></a>

### DIFF
--- a/pkg/util/network.go
+++ b/pkg/util/network.go
@@ -27,8 +27,17 @@ func DownloadFile(destPath string, url string, progressBar bool) (err error) {
 	}
 	defer CheckClose(out)
 
+	request, err := http.NewRequest(http.MethodGet, url, nil)
+
+	// check for github token - used for private addon repositories
+	githubToken := os.Getenv("DDEV_GITHUB_TOKEN")
+	if githubToken != "" {
+		request.Header.Add("Authorization", "Bearer "+githubToken)
+	}
+
 	// Get the data
-	resp, err := http.Get(url)
+	resp, err := http.DefaultClient.Do(request)
+
 	if err != nil {
 		return err
 	}


rfay (changes requested):
My first reaction is that you're making a general change for a specific problem; changing network.go's basic operation with a hack to solve a problem you're having with private add-ons seems a long ways to go.

Do you see any other ways to handle this? Have you considered other ways to solve your private repo problem?

Note that `ddev get` is happy to operate on a local repo, like `ddev get ../some-private-project`, and it can also download nearly any tarball, doesn't have to be a release tarball.

TsPG128:
Hi @rfay ,

Thank you for the review :) 

I had same thoughts first about changing network.go's basic operation.

Regardless, I decided to go along with it because:
- the added snippet is very much low impact, or to say no impact on current functionality
- any other way i can currently see would need more structural changes than this, probably adding a new package and reusing parts of `pkg/archive/archive.go` and `pkg/util/network.go` and changing the way the get command works (i could be wrong about that)
- the current change has potential to be used on other parts that use the network.go functionality 

Thank you for pointing out, I wasn't aware of the local **repo** addon fetching, i did know about the local archive fetching (the fetching of archives didn't suit me as i am planning to version my addons to update in multiple projects). Though private repo functionality included in DDEV sounds even better to me.

P.S. small disclaimer - I'm very new to GO, so please bear with me regarding best practices :see_no_evil: anyways, i am very open to your suggestions and will happily do changes as you see fit best!

rfay:
Yeah, this isn't about golang, it's about adding headers to a general-purpose library function that may or may not be accessing github. It's just not the right way to solve the problem.

Does your add-on need to be private?

This won't be accepted as it is, but happy to discuss with you solutions to your problem.

rfay (changes requested):
I think the right way to do this is to change the signature of DownloadAndExtractTarball to include additional headers (add array of headers param) and change util.DownloadFile() the same.  Then the get logic can pass the header to DownloadAndExtractTarball() if DDEV_GITHUB_TOKEN is set.  Want to try that?

https://github.com/ddev/ddev/blob/bbb2caca961784aa68d0511848823464a009f14b/pkg/archive/archive.go#L474

rfay:
No response after a week, so I'll close this for now. Happy to reopen it and happy to work with you on a solution that's appropriate, https://github.com/ddev/ddev/pull/4832#pullrequestreview-1396020862

I really appreciate you stepping up to try this out, and think you can end up succeeding with it. Happy to continue the conversation here or elsewhere. Closing for now.

rfay:
No response after a week, so I'll close this for now. Happy to reopen it and happy to work with you on a solution that's appropriate, https://github.com/ddev/ddev/pull/4832#pullrequestreview-1396020862

I really appreciate you stepping up to try this out, and think you can end up succeeding with it. Happy to continue the conversation here or elsewhere. Closing for now.

PR #4833, opened 2023-04-18
Switch to NodeJS LTS (18) by default and support NodeJS 20, fixes #4674

### LINKED ISSUE #4674: Add nodejs 20; Make nodejs 18 default in the ddev-webserver image

I know I can change the nodejs version in the configuration but I also use the same image in gitlab CI and is more complicated to change the nodejs version there. I need the fetch API to be available in nodejs

### Describe your solution

Update to v18 of nodejs as default. Nodejs 18 was released on 2022-04-19 and it is LTS. Even the variable in the ddev-php-base dockerfile is called NODE_LTS

### Describe alternatives

_No response_

### Additional context

Is there a policy for ddev as to when to change the default nodejs version?

davereid:
## The Issue

* #4674 

NodeJS 18 is now officially LTS with the NodeJS 20 release today:
https://github.com/nodejs/Release
https://nodejs.org/en/download

## Blockers

The 20 version is not available yet from nodesource distributions, but I have submitted a PR for this at https://github.com/nodesource/distributions/pull/1544

* Updates the Docker image to use `NODE_LTS` of `18`
* ~~Adds support for the `nodejs_version` configuration to be set to `lts` (and is the default). Potentially it seems like we could maybe consider removing this configuration in favor of `ddev nvm`~~

* Restart your DDEV site with this change and running `ddev node --version` should display a version like 18.N.N
* Does this succeed?
  ```
  ddev config --nodejs-version 20
  ```
* Does this fail?
  ```
  ddev config --nodejs-version 19
  ```

## Automated Testing Overview

Should not need any changes to automated tests.

## Related Issue Link(s)

Fixes https://github.com/ddev/ddev/issues/4674

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4833"></a>


pkg/nodeps/values.go
```diff
@@ -77,7 +77,7 @@ var ValidWebserverTypes = map[string]bool{
 	WebserverNginxGunicorn: true,
 }
 
-var ValidNodeJSVersions = []string{"14", "16", "18"}
+var ValidNodeJSVersions = []string{"14", "16", "18", "19", "20"}
```
davereid, line 80: Note, 14 is still valid until 2023-04-30 as per https://github.com/nodejs/Release

pkg/ddevapp/templates.go
```diff
@@ -62,8 +62,8 @@ const ConfigInstructions = `
 # Alternatively, an explicit Composer version may be specified, for example "2.2.18".
 # To reinstall Composer after the image was built, run "ddev debug refresh".
 
-# nodejs_version: "16"
-# change from the default system Node.js version to another supported version, like 12, 14, 17, 18.
+# nodejs_version: "18"
+# change from the default system Node.js version to another supported version, like 12, 14, 17, 18, 19, 20.
```
rfay, line 66: We'll want to get rid of 12 and 17, right? 

I'm also not sure why we would do 19, since you can easily do it with nvm. If it's "supported" I'm OK with it.

rfay:
```suggestion
# change from the default system Node.js version to another supported version, like 16, 18, 19, 20.
```

pkg/nodeps/values.go
rfay: Should we drop 14 then, as this will be in DDEV v1.22, after 2023-04-30? I think probably so?

rfay:
```suggestion
var ValidNodeJSVersions = []string{ "16", "18", "19", "20"}
```

rfay (changes requested):
Thanks for this contribution!

rfay:
Please add the manual testing instructions.

Seems like this isn't really going to be valid until they have the nodesource packages?

davereid:
Yes, I noted in the original PR description that this is blocked by https://github.com/nodesource/distributions/pull/1544 currently.

davereid:
Once the upstream nodesource distribution PR is merged, I'll move this ready to review again.

pkg/nodeps/values.go
davereid: I feel mixed about it, is the policy to always keep this up to date with supported NodeJS versions? Our own internal project didn't switch from 14 to 16 until Sep 29, 2022, which wasn't that long ago. 14 is still valid to download, but just isn't getting any security updates. I would lean towards leaving it as a possible option.

davereid: Should we leave out the odd number (non-LTS) releases like `19`?

rfay: If the idea is "supported", then 19 is supported for now, true?

I'm also mixed, but I'm mixed about this whole thing because we have `ddev nvm` and don't need to even have the package-installed technique :( So maybe we should remove this whole technique and just `ddev nvm install 18` by default.

Also, nodesource packages fail to download have the time, leading to false fails on tests.

rfay:
```
while ! curl -fIL https://deb.nodesource.com/setup_20.x; do
  sleep 3600
done
runtests
```

pkg/nodeps/values.go
davereid: 19 is only supported through the end of May, that's it.

davereid: I would be okay with only LTS versions being listed here and removing 19. We enjoy using this configuration to enforce a standard version of nodejs across 10+ sites with a shared configuration file.

rfay: I dont' think that will meet people's expectations, since people are always using ancient nodejs for ages. They don't know what's lts and what's not...

davereid: If people are already using ancient nodejs, that's their problem and they should be using `ddev nvm`? I think there's an interest here in just always defaulting to the NodeJS LTS.

rfay:
In process of pushing images. `ddev-php- base` takes forever to build.

Please look at https://github.com/ddev/ddev/pull/4833/commits/2b071e0d25961ee89d8d0822b8709c63f0b9c61c - I backed off of your lts approach, sorry. It's admirable, but we don't want people's config for a project changing underneath them.

mattstein (approved):

davereid:
Can we leave in support for 'lts' and 'current' as valid nodejs_version configuration values still though? I'd really like to use that.

rfay:
I don't think it's likely to work as you expect for current code, because of the mismatch between the docker image build getting "lts" and the golang code knowing what "lts" means.

davereid:
I'm not wanting the docker to use lts, I just want it supported in the nodejs_version configuration in DDEV when it goes and fetches a new version that the config has opted into. It should be supported. Based on my reading of https://github.com/ddev/ddev/blob/f26575e808c130c14a823b4c8bf51f85c2b7a5b9/pkg/ddevapp/config.go#L891 it should be supported since nodesource distributions support it.

```go
	if app.NodeJSVersion != nodeps.NodeJSDefault {
		extraWebContent = extraWebContent + "\nRUN (apt-get remove -y nodejs || true) && (apt purge nodejs || true)"
		// Download of setup_*.sh seems to fail a LOT, probably a problem on their end. So try it twice
		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN curl -sSL --fail -o /tmp/setup_node.sh https://deb.nodesource.com/setup_%s.x  ||  curl -sSL --fail -o /tmp/setup_node.sh https://deb.nodesource.com/setup_%s.sh >/tmp/setup_node.sh", app.NodeJSVersion, app.NodeJSVersion)
		extraWebContent = extraWebContent + "\nRUN bash /tmp/setup_node.sh >/dev/null && apt-get install -y nodejs >/dev/null\n" +
			"RUN npm install --unsafe-perm=true --global gulp-cli yarn || ( npm config set unsafe-perm true && npm install --global gulp-cli yarn )"
	}
```

davereid:
Ughhhh, that's a fair point. I'd maybe want to add more logic in there to avoid that, so I could move that to a separate PR.

rfay:
I don't understand why you just did https://github.com/ddev/ddev/pull/4833/commits/d08418f003bbea5211626dbd51952a7675216721

That messes up more than an hour of building that image, but it appears that NODE_LTS isn't used anywhere anyway. All I see is NODE_VERSION,

https://github.com/ddev/ddev/blob/2b071e0d25961ee89d8d0822b8709c63f0b9c61c/containers/ddev-php-base/Dockerfile#L64

Looks like $NODE_LTS was in use previously, https://github.com/search?q=repo%3Addev%2Fddev+NODE_LTS&type=code

But your commit changes nothing that I can see.

rfay:
Now with the images built we can actually see if it works, and get manual testing done, etc. Sorry to revert your reversion.

davereid:
I had change that variable when I updated it to 'lts' but it seemed like an unnecessary change to make with your revertions. Sorry if that caused any hardship, I was just trying to undo more changes that didn't feel appropriate.

rfay:
This will need more from me, not sure why the hardened version didn't get there. Please let me run with it for a while.

rfay:
Looks like it's pretty close. Please manually test and see if it's what you were after @davereid 

We can consider your lts approach in another PR if you think it's useful. I think you'd be disappointed by different results for different team members.

rfay (approved):
I think this is set now, thanks!

rfay:
I wish the ddev-webserver image weren't 10MB bigger with this. I wonder why it is; is it because nodejs 18 is 10MB bigger than node 16? https://hub.docker.com/r/ddev/ddev-webserver/tags

stasadev:
> I wonder why it is; is it because nodejs 18 is 10MB bigger than node 16?

It comes from OpenSSL 3:

- https://github.com/nodejs/node/issues/42081

PR #4653, opened 2023-02-15
Set laravel PHP default to 8.1, fixes #4651

rfay:
## The Issue

Laravel 10 requires PHP 8.1:
* #4651 

## How This PR Solves The Issue

* Change default version in code
* For now, change docs to show explicit --php-version

## Manual Testing Instructions

* Test with laravel
* Test quickstart

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4653"></a>

PR #4834, opened 2023-04-19
Use PHP 8.1 by default, fixes #4652

### LINKED ISSUE #4652: Change PHP Default version to 8.1 in v1.22.0

PHP 8.1 should probably be the default PHP version when next major version (v1.22) comes out.

davereid:
## The Issue

* #4652 

Changes defaults to use PHP 8.1 for new sites or sites that do not have configuration set for `php_version`.

## Related Issue Link(s)

Fixes https://github.com/ddev/ddev/issues/4652

* https://github.com/ddev/ddev/pull/4653

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4834"></a>

### DIFF
--- a/.github/workflows/push-tagged-image.yml
+++ b/.github/workflows/push-tagged-image.yml
@@ -55,7 +55,7 @@ jobs:
       run: |
         cd "containers/${{ github.event.inputs.image }}"
         make push VERSION="${{ github.event.inputs.tag }}"
-    - name: Push ${{  github.event.inputs.image }} image
+    - name: Push all images (not in use)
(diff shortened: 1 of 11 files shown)


rfay (changes requested):
Thanks for this!

This does a lot of it, but you'll find more places. 

* The container tests, for example, don't know about the golang defaults.
* versionconstants.go has to be changed to point to a pushed image. I can do that after we push the image.

rfay (approved):
Thanks, made some changes to docs, pushed images, etc.

PR #4835, opened 2023-04-19
Fix wrong TYPO3 settings path if composer_root is used, fixes #4545

### LINKED ISSUE #4545: TYPO3-project in subfolder leads to wrong configuration path

Diagnostic shows no errors.

### Expected Behavior

When setting up a TYPO3-project in a subfolder `/src` everything should run normal.

### Actual Behavior

When setting up a project in a subfolder `/src` and defining `docroot: src/public` and `composer_root: src` in `config.yaml` there is an additional `/src/src/public/typo3conf/AdditionalConfiguration.php` generated. 

### Steps To Reproduce

Setting up a fresh&clean project with
```
ddev config --project-type=typo3 --docroot=src/public --create-docroot --composer-root=src
ddev composer create --no-install "typo3/cms-base-distribution:^11.5"
ddev composer install
```
Until now, everything is fine. But when doing a
```
ddev start
```
the additional `/src/src/public/typo3conf/AdditionalConfiguration.php` is generated. 

### Anything else?

Only TYPO3 v9-11 are affected. With v12 it seems to behave well.

gilbertsoft:
## The Issue

If composer_root is set for TYPO3 v11 or lower the settings path is wrong calculated. TYPO3 v12 or higher is not affected.

The correct path calculated from the docroot is now used and composer_root is ignored for TYPO3 v11 or lower.

Setup a TYPO3 v11 or lower and set the composer_root. See linked issue for more information.

## Related Issue Link(s)

Fixes 
* #4545

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4835"></a>

### DIFF
--- a/pkg/ddevapp/typo3.go
+++ b/pkg/ddevapp/typo3.go
@@ -129,20 +129,21 @@ func getTypo3Hooks() []byte {
 
 // setTypo3SiteSettingsPaths sets the paths to settings files for templating
 func setTypo3SiteSettingsPaths(app *DdevApp) {
-	settingsFileBasePath := filepath.Join(app.AppRoot, app.ComposerRoot)
 	var settingsFilePath, localSettingsFilePath string
 
 	if isTypo3v12OrHigher(app) {
+		settingsFileBasePath := filepath.Join(app.AppRoot, app.ComposerRoot)
 		settingsFilePath = filepath.Join(settingsFileBasePath, "config", "system", "settings.php")
 		localSettingsFilePath = filepath.Join(settingsFileBasePath, "config", "system", "additional.php")
 	} else if isTypo3App(app) {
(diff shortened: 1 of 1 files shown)


gilbertsoft:
Just found a case where the TYPO3 version isn't recognized at all with TYPO3 v12. On it....

gilbertsoft:
> Just found a case where the TYPO3 version isn't recognized at all with TYPO3 v12. On it....

The case is already mentioned at https://github.com/ddev/ddev/pull/4835/files#diff-5bffff10d8480cab533891faa6c9636a599411f6a2dc1b3dea03882d5886efb1R217-R219 and not easy to solve for the moment. A possible solution would be parsing the composer.json and use the defined vendor path instead.

rfay (approved):
Read through and manually tested with TYPO v11 site, both in normal layout and with src/public and src/vendor and composer_root: src.

PR #4836, opened 2023-04-19
[docs] Add Axiom Repobeats contributor overview

mattstein:
This adds a [Repobeats](https://repobeats.axiom.co) overview to DDEV’s readme to transparently demonstrate the project activity that I think is one of its strengths. It keeps itself up to date and acknowledges contributors.

I deliberately placed it in the _Contributing_ section because it’s relevant there, and I don’t think it should live at the top and distract from a description of what DDEV is that’s for everyone.

Screen Shot 2023-04-19 at 10 31 04 AM@2x


tyler36:
That looks really cool @mattstein 

We should probably reference or include that in any marketing material where appropriate. 
It helps shows active development, engagement with the community and fast response times.

PR #4837, opened 2023-04-19
Recreate mutagen sync session when mutagen.yml changes, fixes #4753

### LINKED ISSUE #4753: Mutagen sync session needs to be recreated when mutagen.yml changes

_No response_

### Expected Behavior

Changing something like upload_dir or project type would "just work"

### Actual Behavior

In some cases where the mutagen.yml is changed, either deliberately by the user or when something like `upload_dir` or project type changes, the mutagen sync session happily continues along without knowing about the new configuration, even after a `ddev restart` or even `ddev poweroff`. 

From @xenoscopic in slack:

> You're correct - the session would need to be recreated in order for changes to mutagen.yml to be picked up; they're current locked-in (mostly so that global configuration changes don't affect sessions unexpectedly).
That is most likely the cause of what you're seeing.  Would it be possible to do something like compute the SHA-1 of the mutagen.yml file that you're using and attach it to sessions as you create them and then do a terminate/create instead of a pause/resume if the associated label changes?

The solution would be to add a label to the mutagen sync session with the hash or checksum of the mutagen.yml file, and then check that on `ddev start` and make sure they match, otherwise recreate the mutagen sync session.

This was experienced by @DuvellAtreus in [Discord](https://discord.com/channels/664580571770388500/1085553833704431636/1085553833704431636)

The easy workaround is a manual `ddev mutagen reset`

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #4753 

A sync session can outlive the mutagen.yml file, as it's not checked again after creation

* Add a label to the mutagen sync session with the SHA1 hash of the existing mutagen.yml
* On start, check to see if the hash remains the same
* If not, recreate the sync session (and volume)

- [x] `ddev start` and review the mutagen sync identifier (`ddev mutagen st -l | grep Identifier`)
- [x] Change mutagen.yml (make sure to remove #ddev-generated)
- [x] `ddev restart` and verify that you have a new mutagen sync session (`ddev mutagen sync st -l | grep Identifier`)
- [x] `ddev restart` without changes and verify you have the same session identifier now

## Automated Testing Overview

Added TestMutagenConfigChange and TestFileHash

## Release/Deployment Notes

* It may be wise for people to `ddev mutagen reset` on the new release. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4837"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -976,7 +976,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 
 	if app.IsMutagenEnabled() {
 		if ok, volumeExists, info := CheckMutagenVolumeSyncCompatibility(app); !ok {
-			util.Debug("mutagen sync session and docker volume are in incompatible status: '%s', Removing mutagen sync session '%s' and docker volume %s", info, MutagenSyncName(app.Name), GetMutagenVolumeName(app))
+			util.Debug("mutagen sync session, configuration, and docker volume are in incompatible status: '%s', Removing mutagen sync session '%s' and docker volume %s", info, MutagenSyncName(app.Name), GetMutagenVolumeName(app))
 			terminateErr := TerminateMutagenSync(app)
 			if terminateErr != nil {
 				util.Warning("Unable to terminate mutagen sync %s: %v", MutagenSyncName(app.Name), err)
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -982,6 +982,14 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 				util.Warning("Unable to terminate mutagen sync %s: %v", MutagenSyncName(app.Name), err)
 			}
 			if volumeExists {
+				// Remove mounting container if necessary.
+				container, err := dockerutil.FindContainerByName("ddev-" + app.Name + "-web")
+				if err == nil && container != nil {
+					err = dockerutil.RemoveContainer(container.ID)
+					if err != nil {
+						return fmt.Errorf(`Unable to remove web container, please 'ddev restart': %v`, err)
+					}
+				}
 				removeVolumeErr := dockerutil.RemoveVolume(GetMutagenVolumeName(app))
 				if removeVolumeErr != nil {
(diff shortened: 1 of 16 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -976,12 +976,20 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 
 	if app.IsMutagenEnabled() {
 		if ok, volumeExists, info := CheckMutagenVolumeSyncCompatibility(app); !ok {
-			util.Debug("mutagen sync session and docker volume are in incompatible status: '%s', Removing mutagen sync session '%s' and docker volume %s", info, MutagenSyncName(app.Name), GetMutagenVolumeName(app))
+			util.Debug("mutagen sync session, configuration, and docker volume are in incompatible status: '%s', Removing mutagen sync session '%s' and docker volume %s", info, MutagenSyncName(app.Name), GetMutagenVolumeName(app))
 			terminateErr := TerminateMutagenSync(app)
```
xenoscopic, line 980: I think my only comment would be that it might be worth flushing the stale session before terminating it, that way both endpoints are theoretically the same when the new session is created.  Doing that will help ensure that the first reconciliation pass of the new synchronization session will construct a virtual ancestor that's (ideally) equal to both endpoints (and that way no pending unsynced changes are lost and converted to conflicts when the new empty ancestor is created in the new session).

However, my guess is that a flush may be done elsewhere, just before `CheckMutagenVolumeSyncCompatibility` (and potentially `TerminateMutagenSync`) is invoked, so if that's the case then just ignore this comment.

Mutagen should probably eventually add a mechanism to transfer the virtual ancestor from one session to a new one (maybe a `mutagen sync update` command, or something).  There are some complexities with that, but it should be possible someday.

xenoscopic (commented):
Hey @rfay, I had a look over this and I think the strategy is good; similar to what Mutagen Compose does (Mutagen Compose actually does an in-memory configuration comparison, but the result is the same).  I only had one comment and it may be something that is already addressed elsewhere in the code.

pkg/ddevapp/ddevapp.go
rfay: Thanks so much, will check!

PR #4853, opened 2023-04-26
[docs] Provide instructions for 3rd party services to trust mkcert

rfay:
## The Issue

It's obscure, but some 3rd party services need to trust the mkcert CA/Certificate

Add docs on how to do it.

Read it and weep.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4853"></a>

### DIFF
--- a/docs/content/users/extend/custom-compose-files.md
+++ b/docs/content/users/extend/custom-compose-files.md
@@ -79,23 +79,23 @@ By default, these commands interact with the `web` container for a project. All 
 
 ## Third Party Services May Need To Trust `ddev-webserver`
 
-In some cases, a third-party service (`docker-compose.*.yaml`) may actually consume content from the `ddev-webserver` container. For example, a PDF generator like `gotenberg` might need to consume webserver content to create a PDF, or a testing service may need to do the same.
+Sometimes a third-party service (`docker-compose.*.yaml`) may need to consume content from the `ddev-webserver` container. A PDF generator like [Gotenberg](https://github.com/gotenberg/gotenberg), for example, might need to read in-container images or text in order to create a PDF. Or a testing service may need to read data in order to support tests.
 
-In this case, the third-party service is by default not configured to trust the certificate authority used by DDEV (via `mkcert`) so you either have to use just HTTP between the two containers, or you have to make the third-party service trust the `mkcert` certificate authority.
+A third-party service is not configured to trust DDEV’s `mkcert` certificate authority by default, so in cases like this you have to either use HTTP between the two containers, or make the third-party service ignore or trust the certificate authority.
 
-The simplest technique is just to use HTTP between the two containers. For example. the [`ddev-selenium-standalone-chrome`](https://github.com/ddev/ddev-selenium-standalone-chrome) service must consume content, and it just conducts interactions with the `ddev-webserver` by accessing `http://web`, see [setup](https://github.com/ddev/ddev-selenium-standalone-chrome/blob/main/config.selenium-standalone-chrome.yaml#L17). In this case, the `selenium-chrome` container just accesses the `web` container via HTTP.
+Using plain HTTP between the containers is the simplest technique. For example, the [`ddev-selenium-standalone-chrome`](https://github.com/ddev/ddev-selenium-standalone-chrome) service must consume content, so it conducts interactions with the `ddev-webserver` [by accessing `http://web`](https://github.com/ddev/ddev-selenium-standalone-chrome/blob/main/config.selenium-standalone-chrome.yaml#L17). In this case, the `selenium-chrome` container accesses the `web` container via HTTP instead of HTTPS.
 
-A second technique is to tell the third-party service to ignore HTTPS/TLS errors. For example, if the third-party service uses `curl`, it could use `curl --insecure https://web` or `curl --insecure https://<project>.ddev.site`.
+A second technique is to tell the third-party service to ignore HTTPS/TLS errors. For example, if the third-party service uses cURL, it could use `curl --insecure https://web` or `curl --insecure https://<project>.ddev.site`.
 
-A third and more complex approach is to actually make the third-party container actually trust the self-signed certificate that the `ddev-webserver` is using. This can be done in many cases using a custom Dockerfile and some extra configuration in the `ddev-config.*.yaml`. An example would be:
+A third and more complex approach is to make the third-party container actually trust the self-signed certificate that the `ddev-webserver` container is using. This can be done in many cases using a custom Dockerfile and some extra configuration in the `ddev-config.*.yaml`. An example would be:
 
 ```yaml
 services:
   example:
     container_name: ddev-${DDEV_SITENAME}-example
     command: "bash -c 'mkcert -install && original-start-command-from-image'"
-    # Add a build stage so we can add mkcert, etc.
-    # The Dockerfile for the build stage goes in the .ddev/example directory here
+    # Add a build stage so we can add `mkcert`, etc.
+    # The Dockerfile for the build stage goes in the `.ddev/example directory` here
     build:
       context: example
     environment:
@@ -103,7 +103,7 @@ services:
       - HTTPS_EXPOSE=3000:3000
       - VIRTUAL_HOST=$DDEV_HOSTNAME
     # Adding external_links allows connections to `https://example.ddev.site`,
-    # which then can go through ddev-router
+    # which then can go through `ddev-router`
     external_links:
       - ddev-router:${DDEV_SITENAME}.${DDEV_TLD}
     labels:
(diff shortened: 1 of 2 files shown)


docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
 [`ddev exec`](../usage/commands.md#exec), [`ddev ssh`](../usage/commands.md#ssh), and [`ddev logs`](../usage/commands.md#logs) interact with containers on an individual basis.
 
 By default, these commands interact with the `web` container for a project. All of these commands, however, provide a `--service` or `-s` flag allowing you to specify the service name of the container to interact with. For example, if you added a service to provide Apache Solr, and the service was named `solr`, you would be able to run `ddev logs --service solr` to retrieve the Solr container’s logs.
+
+## Third Party Services May Need To Trust `ddev-webserver`
+
+In some cases, a third-party service (`docker-compose.*.yaml`) may actually consume content from the `ddev-webserver` container. For example, a PDF generator like `gotenberg` might need to consume webserver content to create a PDF, or a testing service may need to do the same.
```
mattstein, line 82: Revise for flow, use Gotenberg’s proper noun name and link to the project so people can see that glorious image, try to be slightly more specific about what the examples need to read and why:

```suggestion
Sometimes a third-party service (`docker-compose.*.yaml`) may need to consume content from the `ddev-webserver` container. A PDF generator like [Gotenberg](https://github.com/gotenberg/gotenberg), for example, might need to read in-container images or text in order to create a PDF. Or a testing service may need to read data in order to support tests.
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+
+In this case, the third-party service is by default not configured to trust the certificate authority used by DDEV (via `mkcert`) so you either have to use just HTTP between the two containers, or you have to make the third-party service trust the `mkcert` certificate authority.
```
mattstein, line 84: Whittle down slightly, avoid the parenthetical to improve flow, and describe _all_ the options that are about to be explained—which includes both trusting and _ignoring_ SSL issues:

```suggestion
A third-party service is not configured to trust DDEV’s `mkcert` certificate authority by default, so in cases like this you have to either use HTTP between the two containers, or make the third-party service ignore or trust the certificate authority.
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+
+The simplest technique is just to use HTTP between the two containers. For example. the [`ddev-selenium-standalone-chrome`](https://github.com/ddev/ddev-selenium-standalone-chrome) service must consume content, and it just conducts interactions with the `ddev-webserver` by accessing `http://web`, see [setup](https://github.com/ddev/ddev-selenium-standalone-chrome/blob/main/config.selenium-standalone-chrome.yaml#L17). In this case, the `selenium-chrome` container just accesses the `web` container via HTTP.
```
mattstein, line 86: Avoid “just”s, fix punctuation, use links in context for economy and better flow, and contrast with HTTPS since so far it’s only implied:

```suggestion
Using plain HTTP between the containers is the simplest technique. For example, the [`ddev-selenium-standalone-chrome`](https://github.com/ddev/ddev-selenium-standalone-chrome) service must consume content, so it conducts interactions with the `ddev-webserver` [by accessing `http://web`](https://github.com/ddev/ddev-selenium-standalone-chrome/blob/main/config.selenium-standalone-chrome.yaml#L17). In this case, the `selenium-chrome` container accesses the `web` container via HTTP instead of HTTPS.
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+
+A second technique is to tell the third-party service to ignore HTTPS/TLS errors. For example, if the third-party service uses `curl`, it could use `curl --insecure https://web` or `curl --insecure https://<project>.ddev.site`.
```
mattstein, line 88: Use the proper noun of [cURL](https://en.wikipedia.org/wiki/CURL) before the command because words are fun!

```suggestion
A second technique is to tell the third-party service to ignore HTTPS/TLS errors. For example, if the third-party service uses cURL, it could use `curl --insecure https://web` or `curl --insecure https://<project>.ddev.site`.
```

This would be a nice place to demonstrate using cURL via CLI, PHP with `curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false)`, etc. with tabs for each major language DDEV supports.

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+
+A third and more complex approach is to actually make the third-party container actually trust the self-signed certificate that the `ddev-webserver` is using. This can be done in many cases using a custom Dockerfile and some extra configuration in the `ddev-config.*.yaml`. An example would be:
```
mattstein, line 90: Drop an “actually”, describe what’s in backticks.

```suggestion
A third and more complex approach is to make the third-party container actually trust the self-signed certificate that the `ddev-webserver` container is using. This can be done in many cases using a custom Dockerfile and some extra configuration in the `ddev-config.*.yaml`. An example would be:
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+
+```yaml
+services:
+  example:
+    container_name: ddev-${DDEV_SITENAME}-example
+    command: "bash -c 'mkcert -install && original-start-command-from-image'"
+    # Add a build stage so we can add mkcert, etc.
```
mattstein, line 97: Use backticks consistently in these comments:

```suggestion
    # Add a build stage so we can add `mkcert`, etc.
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+    # The Dockerfile for the build stage goes in the .ddev/example directory here
```
mattstein, line 98: Same here:

```suggestion
    # The Dockerfile for the build stage goes in the `.ddev/example directory` here
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+    build:
+      context: example
+    environment:
+      - HTTP_EXPOSE=3001:3000
+      - HTTPS_EXPOSE=3000:3000
+      - VIRTUAL_HOST=$DDEV_HOSTNAME
+    # Adding external_links allows connections to `https://example.ddev.site`,
+    # which then can go through ddev-router
```
mattstein, line 106:
```suggestion
    # which then can go through `ddev-router`
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+    external_links:
+      - ddev-router:${DDEV_SITENAME}.${DDEV_TLD}
+    labels:
+      com.ddev.approot: $DDEV_APPROOT
+      com.ddev.site-name: ${DDEV_SITENAME}
+    restart: 'no'
+    volumes:
+      - .:/mnt/ddev_config
+      # ddev-global-cache gets mounted so we have the CAROOT
```
mattstein, line 115: Pretty sure this should get backticked too:

```suggestion
      # `ddev-global-cache` gets mounted so we have the CAROOT
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+      # This is required so that the CA is available for mkcert to install
```
mattstein, line 116:
```suggestion
      # This is required so that the CA is available for `mkcert` to install
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+      - ddev-global-cache:/mnt/ddev-global-cache
+```
+
+```Dockerfile
+FROM example/example
+
+# CAROOT for mkcert to use, has the CA config
```
mattstein, line 123:
```suggestion
# CAROOT for `mkcert` to use, has the CA config
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+ENV CAROOT=/mnt/ddev-global-cache/mkcert
+
+# If the image build does not run as USER root,
+# temporarily change to root. If the image runs as USER root (default) then
```
mattstein, line 127: I’m not confident that I’m reading this as you intend, but it feels like something should need backticks.

> If the image build does not run as USER root

The uppercase “user” here is throwing me off, but I wonder if this should read “Temporarily change to the `root` user in case the image build is run by another.”

Or even simply “Make sure the image build is running as the `root` user.”

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+# there is no need to change user here.
+USER root
+# Give the "example" user full sudo privileges
```
mattstein, line 130:
```suggestion
# Give the `example` user full `sudo` privileges
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -76,3 +76,61 @@ When defining additional services for your project, we recommended following the
...
+RUN echo "example ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers.d/example && chmod 0440 /etc/sudoers.d/example
+# Install the correct architecture binary of mkcert
```
mattstein, line 132:
```suggestion
# Install the correct architecture binary of `mkcert`
```

mattstein (changes requested):

docs/content/users/extend/custom-compose-files.md
rfay: Agreed, but we're already SO FAR IN THE WEEDS :)

docs/content/users/extend/custom-compose-files.md
rfay: Didn't know we needed backticks inside code demonstration, but works for me!

mattstein: That’s my habit generally, but here I was following your lead slightly later in these examples.

docs/content/users/extend/custom-compose-files.md
rfay: The upper-case USER is referring to the default 

```Dockerfile
USER root
```

So I don't think the usage here is as good as it could be. 

```suggestion
# If the image build does not run as the default `root` user,
# temporarily change to root. If the image already has the default setup
# where it builds as `root`, then
```

PR #4861, opened 2023-05-03
[docs] Try to fix readthedocs builds

rfay:
## The Issue

Readthedocs builds have been broken, see https://readthedocs.org/projects/ddev/builds/20365504/

None of it makes sense to me, but I don't understand why we'd need "future" https://pypi.org/project/future/ which is a compatibility layer for python2/3

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4861"></a>

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -15,5 +15,9 @@ build:
 
 python:
   install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
     - requirements: docs/mkdocs-pip-requirements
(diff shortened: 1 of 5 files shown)


rfay:
The key to this was 
* https://github.com/readthedocs/readthedocs.org/issues/10286#issuecomment-1533206594

which seemed to have nothing to do with anything, but rolling back setuptools using the technique suggested there solved the problem.

mattstein:
What a journey!

PR #4866, opened 2023-05-03
Deliver instrumentation data directly to amplitude, fixes #4335

### LINKED ISSUE #4335: Send telemetry directly to Amplitude instead of via Segment.com

Segment is quite expensive, probably not reasonable to pay them more than $1000/year. We use it to gather data and then it is sent along to amplitude.

We have been on segment's free plan, which used to just cut us off when we got partway through the month and used up our pushes. Now... they have cut us off completely.

### Describe your solution

Apparently we can send data directly to Amplitude, which is currently free.

### Describe alternatives

* Pay segment.
* Drop telemetry

### Additional context

_No response_

gilbertsoft:
## The Issue

* #4335 

Statistics are now directly sent to Amplitude. All events are now properly logged again and to not to affect the user the data is locally cached and sent once the configurable limits are reached. The existing implementation using segment.io is still in place and removed in a second step when the new implementation is tested in the field and confirmed to work fine and like expected.

To build the binary including the Amplitude API key use the following command:

`AmplitudeAPIKey=xxx && make build` where `xxx` has to be replaced by a valid key which can be found at https://data.amplitude.com/ddev/DDEV/sources/name/ddev

How to debug the data is explained at https://www.docs.developers.amplitude.com/data/debugger/.

It's also possible to create new charts with the data to debug it.

Settings `DDEV_DEBUG=true` and `DDEV_VERBOSE=true` will generate some additional output too which helps in testing.

There are also two new command `ddev debug instrumentation flush` to send all cached data to Amplitude and `ddev debug instrumentation clean` which will remove cached data. Helps also while testing.

**Manual mass testing can be done with the following line:**

`export DDEV_VERBOSE=true; for ((i=1;i<=50;i++)); do ddev describe >/dev/null; done 2>&1 | tee ~/tmp/describe.out`

This should do multiple submissions depending on the number of projects on the system.

Submitted events are visible at <https://analytics.amplitude.com/ddev/activity> afterwards.

This patch uses our [forked repo](https://github.com/ddev/analytics-go/tree/fix-failed-transmissions) of https://github.com/amplitude/analytics-go until the issue gets fixed in the upstream.

## Automated Testing Overview

Most parts are covered by unit tests. Currently no real E2E tests are implemented and does look superfluous because no problems occured so far with the delivery and some parts are already tested by the upstream.

* [x] some tests are currently really minimal e.g. for the Tracker functions. Check if better testing is possible.

## Related Issue Link(s)

* https://github.com/amplitude/analytics-go/pull/56

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4866"></a>


rfay:
Please make sure to explain what environment variables have to be set to make this work. I assume there's a secret that has to be compiled in or something. But we'll want to manually test here of course.

rfay:
The problem is that `make test` fails. `pkg/amplitude` has no go files, but pkg/amplitude/storages has one. I think it's just that the test process adds each directory in pkg to the list of items to test.

docs/content/developers/building-contributing.md
```diff
@@ -135,6 +135,38 @@ To manually push using GitHub Actions,
...
+
+Make changes to the event definition using the GUI at <https://data.amplitude.com/ddev/DDEV>:
+
+* create a new branch
+* create or change events and properties
+* save changes to the new branch
+* update the implementation with `ampli checkout <branch name>`
+* make changes to the code
+
+Make sure new properties are not breaking, add it as optional properties though.
```
mattstein, line 157: I’m only following along theoretically, but I’m confused by this sentence. (The plural “new properties” followed by singular “it” could probably be corrected regardless.) Are you saying to make sure that each new property doesn’t break anything, but to make it optional for safety?

docs/content/developers/building-contributing.md
```diff
@@ -135,6 +135,38 @@ To manually push using GitHub Actions,
...
+Once finished, save the changes to publish a new version of the definitions.
+
+Afterwards the changes can be imported running the following command in the
+project root:
+
+```bash
+ampli pull
+```
+
+Make sure the API keys are not included to the sources, they are linked during
```
mattstein, line 167: Semicolon:

```suggestion
Make sure the API keys are not included to the sources; they are linked during
```

docs/content/developers/building-contributing.md
```diff
@@ -135,6 +135,38 @@ To manually push using GitHub Actions,
...
+the compilation using a GitHub secret.
```
mattstein, line 168: Drop “the”:

```suggestion
compilation using a GitHub secret.
```

mattstein (changes requested):

docs/content/developers/building-contributing.md
gilbertsoft: It looks like this can be removed. The API does not check the properties to be there, it's for the code generation only which I did not know at the point where I wrote it.

rfay:
Could you please analyze https://community.amplitude.com/announcements-56/updates-to-our-free-starter-plan-1819 and verify that it's not likely to affect us @gilbertsoft ? I think if *users* are defined in a way that works we should be OK?

gilbertsoft:
> Could you please analyze https://community.amplitude.com/announcements-56/updates-to-our-free-starter-plan-1819 and verify that it's not likely to affect us @gilbertsoft ? I think if _users_ are defined in a way that works we should be OK?

Regarding https://amplitude.com/pricing it's up to 100k users a month. At the moment we should not be affected though, bit more than 10k users a month.

cmd/ddev/cmd/root.go
```diff
@@ -177,8 +182,8 @@ func init() {
 }
 
 func instrumentationNotSetUpWarning() {
-	if !output.JSONOutput && versionconstants.SegmentKey == "" && globalconfig.DdevGlobalConfig.InstrumentationOptIn {
-		util.Warning("Instrumentation is opted in, but SegmentKey is not available. This usually means you have a locally-built ddev binary or one from a PR build. It's not an error. Please report it if you're using an official release build.")
+	if !output.JSONOutput && globalconfig.DdevGlobalConfig.InstrumentationOptIn && (versionconstants.SegmentKey == "" || versionconstants.AmplitudeAPIKey == "") {
```
rfay, line 185: This seems like it will cause confusion, as either of the keys being missing gives the same warning.  Should we remove the checks for SegmentKey in these situations?

docs/content/developers/building-contributing.md
```diff
@@ -135,6 +135,37 @@ To manually push using GitHub Actions,
...
+
+Afterwards the changes can be imported running the following command in the
+project root:
+
+```bash
+ampli pull
+```
+
+Make sure the API keys are not included to the sources; they are linked during
+compilation using a GitHub secret.
```
rfay, line 167: What is the current status of the PR you made upstream? I don't see a mention of that here.

docs/content/developers/release-management.md
```diff
@@ -22,13 +22,14 @@
...
 
+* `AMPLITUDE_API_KEY`: Key that enables Amplitude reporting.
 * `AUR_SSH_PRIVATE_KEY`: Private ssh key for the `ddev-releaser` user. This must be processed into a single line, for example, `perl -p -e 's/\n/<SPLIT>/' ~/.ssh/id_rsa_ddev_releaser| pbcopy`.
 * `CHOCOLATEY_API_KEY`: API key for Chocolatey.
 * `DDEV_GITHUB_TOKEN`: GitHub token that gives access to create releases and push to the Homebrew repositories.
 * `DDEV_MACOS_APP_PASSWORD`: Password used for notarization, see [signing_tools](https://github.com/ddev/signing_tools).
 * `DDEV_MACOS_SIGNING_PASSWORD`: Password for the macOS signing key, see [signing_tools](https://github.com/ddev/signing_tools).
 * `DDEV_WINDOWS_SIGNING_PASSWORD`: Windows signing password.
-* `SegmentKey`: Key that enables Segment reporting.
+* `SEGMENTKEY`: Key that enables Segment reporting.
```
rfay, line 32: The name of this environment variable is SegmentKey still. But I'm not convinced we shouldn't be removing this line completely.

pkg/ddevapp/amplitude.go
```diff
@@ -0,0 +1,53 @@
+package ddevapp
```
rfay, line 1: I usually find it annoying to have two go files with the same name, just because it's easy to get the wrong one when looking for it in the IDE.

pkg/globalconfig/global_config.go
```diff
@@ -36,33 +37,35 @@ type ProjectInfo struct {
...
-	RequiredDockerComposeVersion string                  `yaml:"required_docker_compose_version,omitempty"`
-	UseDockerComposeFromPath     bool                    `yaml:"use_docker_compose_from_path,omitempty"`
-	MkcertCARoot                 string                  `yaml:"mkcert_caroot"`
-	ProjectTldGlobal             string                  `yaml:"project_tld"`
-	XdebugIDELocation            string                  `yaml:"xdebug_ide_location"`
-	NoBindMounts                 bool                    `yaml:"no_bind_mounts"`
-	UseTraefik                   bool                    `yaml:"use_traefik"`
-	WSL2NoWindowsHostsMgt        bool                    `yaml:"wsl2_no_windows_hosts_mgt"`
-	ProjectList                  map[string]*ProjectInfo `yaml:"project_info"`
+	OmitContainersGlobal             []string                `yaml:"omit_containers,flow"`
```
rfay, line 40: In general, please avoid letting your IDE make significant changes that are not related to the PR.

third_party/ampli/ampli.go
```diff
@@ -0,0 +1,527 @@
+// ampli.go
+//
+// Ampli - A strong typed wrapper for your Analytics
+//
+// This file is generated by Amplitude.
+// To update run 'ampli pull ddev'
```
rfay, line 6: Link to docs would be great here.

rfay (changes requested):
Here's my first casual look-through, but I need more help to be able to manually test this. Please add manual testing steps, explaining how to build, and especially how to test, and show where to get the secret key. (I think you told me about the secret key somewhere in Discord, but it needs to be here.)

In Segment there was a nice debugging page where you could see events coming in. Is there something like that for Amplitude?

third_party/ampli/ampli.go
gilbertsoft: This file is generated, we should not manually add something here. But I've added some hint to the docs already.

pkg/globalconfig/global_config.go
gilbertsoft: I've added 2 new configs which caused the new indention, related to this PR though.

docs/content/developers/building-contributing.md
gilbertsoft: I did not submit it because I first wanted to have it tested a litle bit here. So far my changes work, no more issues.

cmd/ddev/cmd/root.go
gilbertsoft: Well, I did not split it because it will anyway get removed soon.

gilbertsoft:
> Here's my first casual look-through, but I need more help to be able to manually test this. Please add manual testing steps, explaining how to build, and especially how to test, and show where to get the secret key. (I think you told me about the secret key somewhere in Discord, but it needs to be here.)
> 
> In Segment there was a nice debugging page where you could see events coming in. Is there something like that for Amplitude?

Done

docs/content/developers/release-management.md
gilbertsoft: Oh that's the name of the env var? Thought this is the secret used at GH... I would not remove it at the moment. Removal is planed with a second PR to remove the old implementation once we are happy with the new data which we know after the app was running in the field for some time.

gilbertsoft: Please check the introduction before the list, it's about the secrets and not the env vars.

pkg/ddevapp/amplitude_project.go
```diff
@@ -0,0 +1,134 @@
...
+			serviceName := strings.TrimPrefix(k.Names[0], "/")
+			shortName := strings.Replace(serviceName, fmt.Sprintf("ddev-%s-", app.Name), "", 1)
+			services = append(services, shortName)
+		}
+	}
+
+	builder := ampli.Project.Builder().
+		Containers(services).
+		ContainersOmitted(containersOmitted).
+		DocumentRoot(app.GetDocroot()).
```
gilbertsoft, line 53: Not sure if this has any value for instrumentation @rfay

pkg/ddevapp/amplitude_project.go
```diff
@@ -0,0 +1,134 @@
...
+	if !nodeps.ArrayContainsString(containersOmitted, "db") {
+		builder.
+			DatabaseType(app.Database.Type).
+			DatabaseVersion(app.Database.Version)
+	}
+
+	ampli.Instance.Project(amplitude.GetUserID(), builder.Build())
+
+	/*
+		appDesc["fail_on_hook_fail"] = app.FailOnHookFail || app.FailOnHookFailGlobal
```
gilbertsoft, line 73: Here is some information which currently is not sent to Amplitude. Maybe you see something with value for instrumentation which should get added @rfay

rfay:
I didn't get any extra output from `ddev restart` when I set `DDEV_DEBUG=true` or `DDEV_VERBOSE=true`

pkg/ddevapp/amplitude_project.go
rfay: I agree. I've never looked at it. I guess the only reason it could exist is if we discover people using new and exotic docroot values. If we were debugging with this we might want it, but we really don't use it.

pkg/ddevapp/amplitude_project.go
rfay: This might be useful, to know how often people set it to fail on hook fail. It used to do that by default. I never did get any pushback when I changed it.

pkg/globalconfig/global_config.go
rfay: Oh, I see, the new configs were longer, and the IDE decided to adjust everything as a result.

docs/content/developers/building-contributing.md
gilbertsoft: PR is submitted in the meantime https://github.com/amplitude/analytics-go/pull/56. Just wanted to make sure it works like expected before sumbitting it.

pkg/ddevapp/amplitude_project.go
gilbertsoft: I don't think we need values here for debugging purposes! Will remove it...

pkg/globalconfig/global_config.go
gilbertsoft: Exactly :)

gilbertsoft:
> I didn't get any extra output from `ddev restart` when I set `DDEV_DEBUG=true` or `DDEV_VERBOSE=true`

Works here

docs/content/developers/release-management.md
gilbertsoft: Documented now, we have to stick to the name used in the source.

pkg/ddevapp/amplitude_project.go
gilbertsoft: Comment is removed now. If we miss sth we can add it at anytime later....

docs/content/developers/building-contributing.md
gilbertsoft: Hint about the fork ist added in various places.

gilbertsoft:
Last thing to do is to check the tests if sth can be optimized like mentioned in the description above.

rfay (changes requested):
This is looking pretty great! I didn't have any trouble and didn't see any slowdowns, and even the flush was fast.

* Please add the global config items to the docs and to https://github.com/ddev/ddev/blob/a27ef6d5e8e56d4ae96b76053f6e9567c05b4420/pkg/globalconfig/global_config.go#L326-L334
* If it's not hard, please support instrumentation_user as segment did. It's just a way to filter the events.

gilbertsoft:
> This is looking pretty great! I didn't have any trouble and didn't see any slowdowns, and even the flush was fast.
> 
>     * Please add the global config items to the docs and to https://github.com/ddev/ddev/blob/a27ef6d5e8e56d4ae96b76053f6e9567c05b4420/pkg/globalconfig/global_config.go#L326-L334
> 
>     * If it's not hard, please support instrumentation_user as segment did. It's just a way to filter the events.

Done

rfay:
Sorry, didn't get to it today, hopefully tomorrow.

mattstein (approved):

rfay:
I'm not sure this is working at all. I spent some time manually testing on gitpod (build with the api key). Get no warnings. Did things like `ddev restart && for i in {1..100}; do ddev list; done` and I did get new entries for "version" and "OS" to show up... but both are "None"

image

I imagine I'm doing something stupid, but may need some help to do a valid test. I do see amplitude.Flush() happening with DDEV_VERBOSE.

rfay:
I see it says "Usage statistics transmitted." on `ddev debug instrumentation flush` and perhaps other times when DDEV_VERBOSE. Could it give a little more info? Like how many or something like that? It doesn't leave me feeling confident. It could also say whether it was delivering to dev or prod.

gilbertsoft:
> I see it says "Usage statistics transmitted." on `ddev debug instrumentation flush` and perhaps other times when DDEV_VERBOSE. Could it give a little more info? Like how many or something like that? It doesn't leave me feeling confident. It could also say whether it was delivering to dev or prod.

The destination environment depends on the injected API key, not possible to know which one is currently used because we can not add the API keys to the source ofc.

Debugging messages are enhanced with some more information now which finally should make debugging and testing easier.

rfay:
At this point, I seem to only get things like this when doing `for ((i=1;i<=200;i++)); do ddev describe >/dev/null; done 2>&1 | tee ~/tmp/describe.out`; I do not see data being sent after grepping out all the PERF: entries

```
time="2023-06-15T12:33:38-06:00" level=error msg="\x1b[31mError 'EOF', while reading event cache.\x1b[0m"
time="2023-06-15T12:31:49-06:00" level=error msg="\x1b[31mError 'unexpected EOF', while reading event cache.\x1b[0m"
```

[describe.txt](https://github.com/ddev/ddev/files/11761605/describe.txt)

rfay:
With `export DDEV_VERBOSE=true; for ((i=1;i<=200;i++)); do ddev describe >/dev/null; done 2>&1 | tee ~/tmp/describe.out` on e2556b775b292eab23e199ccc716da189067c554 (v1.22.0-alpha2-133-ge2556b775) I see the data coming into Amplitude, but the verbose output showing (anything) seems to be lost - no reports about the payload, or anything else. Now all I have is PERF: and such. 

[describe.txt](https://github.com/ddev/ddev/files/11781832/describe.txt)

rfay:
I think it's probably important to update the manual testing instructions with a more significant set of techniques. Please add quick and easy ways to test manually that definitely show the results. For example, some replacement for my for loop, how to look at amplitude, how to capture the output, etc.

gilbertsoft:
> With `export DDEV_VERBOSE=true; for ((i=1;i<=200;i++)); do ddev describe >/dev/null; done 2>&1 | tee ~/tmp/describe.out` on [e2556b7](https://github.com/ddev/ddev/commit/e2556b775b292eab23e199ccc716da189067c554) (v1.22.0-alpha2-133-ge2556b775) I see the data coming into Amplitude, but the verbose output showing (anything) seems to be gone - no reports about the payload, or anything else. Now all I have is PERF: and such.
> 
> Edit: I think maybe you moved some of this to DDEV_DEBUG? I see it all when I do `ddev debug instrumentation flush` with DDEV_DEBUG=true; still don't see it when running the for loop.
> 

It was always outputet with `DDEV_DEBUG=true` and should still be the case, I did not change anything here.

> Edit: Oh, you turned off reporting about many things? https://github.com/ddev/ddev/pull/4866/files#diff-251397554946dd8f37a081ff0d8638b12916434618051440c546fa08c20cf1bcR91
> 
> [describe.txt](https://github.com/ddev/ddev/files/11781832/describe.txt) [flush.txt](https://github.com/ddev/ddev/files/11781892/flush.txt)

This is related to the old Segment instrumentation.

gilbertsoft:
> I think it's probably important to update the manual testing instructions with a more significant set of techniques. Please add quick and easy ways to test manually that definitely show the results. For example, some replacement for my for loop, how to look at amplitude, how to capture the output, etc.
> 
> A script that one could run that would manually verify would be great.

Added an example of mass testing and where to see the incoming data on Amplitude.

Also changed the output behavior in the meantime, output releated to debugging should be printed to stderr now.

rfay (approved):
OK, are you exhausted yet? I am! 

Go ahead and pull this when you're ready.

ampli.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
rfay: Yeah, I just try to keep random files out of the root. Certainly not successful all the time.

rfay:
Yippee! Congratulations!

PR #4873, opened 2023-05-04
Fix log location (supervisord) in several places

rfay:
## The Issue

* In docs and in the build for web_extra_daemons, the logs weren't being sent to the right place
* In the hardened (prod) docker image, the newer gunicorn supervisord config wasn't even being included

* Fix up those things
* Remove the prod files as there was no reason for them to be duplicated

In general, this doesn't affect much; the logs issues are minor, and the prod image is rarely used.

- [ ] Run the prod image (better: Deploy it with ddev) and make sure /etc/supervisor has the gunicorn files.
- [ ] Add a web_extra_daemons entry and verify that logs from the daemon are shown in `ddev logs`

## Automated Testing Overview

No changes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4873"></a>


rfay:
Looks like the hardened/prod healthcheck is broken.

mattstein (approved):

PR #4885, opened 2023-05-08
Feature showing messages to the users, fixes #3799

### LINKED ISSUE #3799: Create custom message/poll on ddev execution  - similar to the new version warning - for use on user feedback

We currently have very limited means to get feedback from the user base. We want to have a way to engage a lot more people, specifically the end-users, and not only the team leaders.

Similar to how we are warning users that there's a new version of ddev available, we should be able to have a mechanism to inject a custom message announcing something like:

"Please help improve DDEV, click on the following link to provide simple feedback on feature X: http://example.com/quiz"

We can continue with Discord and Twitter, but that only reaches those that actually care enough about DDEV to sign in to Discord or to follow DDEV in Twitter.

### Additional context

_No response_

gilbertsoft:
## The Issue

* #3799 

We like to show messages to the user e.g. a hint to support DDEV with sponsoring the project.

This patch introduces an easy way to provide infos, warnings and rotating ticker messages in the remote-config repository which will be regulary downloaded and shown to the user.

Open tasks:
* [x] define a good first message about sponsoring DDEV
* [x] define some first tips

To test this feature run a ddev command e.g. `ddev describe`, messages should be shown. Manipulate or delete the state file `~/.ddev/.state.yaml` to get faster results. It's also possible to change the interval of the ticker in the `~/.ddev/global_config.yaml` or switch to the testing branch of the remote-config, description to do so can be found in the config file.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4885"></a>


docs/content/developers/messages.md
```diff
@@ -0,0 +1,35 @@
+# Messages
+
+## Usage
+
+DDEV supports to show messages to users on every call. The messages are
+regularly downloaded from the Github repository from the `manifest.json`
```
gilbertsoft, line 6: Thinking about using a dedicated file `messages.json` instead.

rfay:
It would be cool to be able to show a variety of messages, so people don't get bored. Instructional, etc.

gilbertsoft:
> It would be cool to be able to show a variety of messages, so people don't get bored. Instructional, etc.

Nice idea. Like Docker Desktop does it on startup. Will think about...

gilbertsoft:
> It would be cool to be able to show a variety of messages, so people don't get bored. Instructional, etc.

Done

docs/content/developers/manifest.md
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+```
+
+## Messages
+
+### Infos and Warnings
+
+The defined messages are shown to the user on every usage of DDEV. Supported
+messages types are `infos` and `warnings`. Every message can optionally include
```
rfay, line 39:
```suggestion
message types are `infos`, `tips`, and `warnings`. Every message can optionally include
```

pkg/github/github.go
```diff
@@ -0,0 +1,26 @@
+package github
+
+import (
+	"context"
+	"os"
+
+	"github.com/google/go-github/v52/github"
```
rfay, line 7: Hmm, we're already using `github.com/google/go-github/github` in get.go. Can these usages be consolidated? What's up with using `v52` here?

pkg/github/github.go
```diff
@@ -0,0 +1,26 @@
...
+	"golang.org/x/oauth2"
+)
+
+type RepositoryContentGetOptions = github.RepositoryContentGetOptions
+
+// GetGithubClient creates the required github client
+func GetGithubClient(ctx context.Context) github.Client {
```
rfay, line 14: What's the difference between this and getGithubClient()? https://github.com/ddev/ddev/blob/18237c7dc49fa67b2be755964e19c5e9323e6f43/cmd/ddev/cmd/get.go#L367

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/ddev/ddev/pkg/manifest/storages"
+	"github.com/ddev/ddev/pkg/manifest/types"
+	"github.com/ddev/ddev/pkg/util"
+)
+
+func NewManifest(updateInterval time.Duration) *Manifest {
+	manifest := &Manifest{
+		fileStorage: storages.NewFileStorage(getLocalFileName()),
+		// TODO change to ddev repo before merge
+		githubStorage:  storages.NewGithubStorage(`gilbertsoft`, `ddev`, `manifest.json`, storages.Options{Ref: "task/upstream-infos"}),
```
rfay, line 18: I think we should put the manifest.json somewhere else so that it's not tied to ddev/ddev? That would make it easier to update, and resolve the problem we're having here with having to use a temporary location. It would also make this testable on PRs, etc.

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (m *Manifest) loadFromLocalStorage() {
+	runTime := util.TimeTrack(time.Now(), "loadFromLocalStorage()")
+	defer runTime()
+
+	m.mu.Lock()
+	defer func() {
+		m.mu.Unlock()
+		/*go*/ m.updateFromGithub()
```
rfay, line 60: What does the /*go*/ notation imply?

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,108 @@
...
+			m.mu.Unlock()
+			m.Write()
+		}()
+
+		// Download the manifest.
+		var err error
+		m.Manifest, err = m.githubStorage.Pull()
+
+		if err != nil {
+			util.Error("Error while downloading manifest: %s", err)
```
rfay, line 100: I'm not sure we should be notifying the user about problems like this.

pkg/manifest/messages.go
```diff
@@ -0,0 +1,75 @@
+package manifest
```
rfay, line 1: As with the name of the file itself, the package name here is too generic and doesn't say anything about the usage. If this is really focused just on tickler messages only, maybe the package name should be ticklermessages or just `ticklers` ? If the package has grander ambitions, then we should name it based on that.  (vscode should be able to rename this without trouble)

pkg/manifest/storages/file_storage.go
```diff
@@ -0,0 +1,94 @@
+package storages
```
rfay, line 1: `storages` is not a normal English usage. Maybe `storage`?

pkg/manifest/storages/file_storage.go
```diff
@@ -0,0 +1,94 @@
...
+func (s *fileStorage) LastUpdate() time.Time {
+	err := s.loadData()
+	if err != nil {
+		return time.Time{}
+	}
+
+	return s.data.LastUpdate
+}
+
+func (s *fileStorage) Pull() (messages types.Manifest, err error) {
```
rfay, line 39: Since Pull() and Push() have such established usages in DDEV it may be better to use a synonym here for what you're trying to accomplish. I can't make a suggestion yet because I'm not sure what Pull and Push mean yet.

rfay (changes requested):
Taking a first casual look through. I don't understand it all yet, but made some suggestions about naming, etc. It's a lot of code for a small feature of course. 

I see lots and lots of unexpected stuff being loaded into vendor, like github.com/ProtonMail (go-crypto I guess, but pulls in lots more, why that and also x/crypto?)? github.com/cloudflare (circl I guess, but what for?)? and of course the duplicated go-github/v52

Currently, every `ddev start` repeats the same message, "Ensure maintenance and further development of DDEV see..."  We don't want to repeat the same message all the time, and don't want to show a message on every command. Maybe once a day?

It does seem to rotate on some of the other messages, but I don't think we want to show any message on most commands. Just daily or whatever?

But it's currently just on start, is that true?

gilbertsoft:
> Taking a first casual look through. I don't understand it all yet, but made some suggestions about naming, etc. It's a lot of code for a small feature of course.
> 
> I see lots and lots of unexpected stuff being loaded into vendor, like github.com/ProtonMail (go-crypto I guess, but pulls in lots more, why that and also x/crypto?)? github.com/cloudflare (circl I guess, but what for?)? and of course the duplicated go-github/v52

I've installed the latest version of go-github (v52) which is breaking and so far did not update the existing usages e.g. for `ddev get` or the update check. Imho we should do this later in a dedicated patch. So long, two version live together which does not hurt. The new version is also the reason for the new dependencies.

> 
> Currently, every `ddev start` repeats the same message, "Ensure maintenance and further development of DDEV see..." We don't want to repeat the same message all the time, and don't want to show a message on every command. Maybe once a day?

There are two features like mentioned in the docs: infos and warnings, which will show on every start, and the tips, which will rotate.

> 
> It does seem to rotate on some of the other messages, but I don't think we want to show any message on most commands. Just daily or whatever?

At the moment the messages are shown on `ddev start`. We should discuss how we like to have it.

The idea of infos and warnings is to have the posibility to show important messages to the user. This could make sense to be shown on most commands and on every call. E.g. if an update of an old version is important because of a security risk.

The tips is someting we also could implement to be shown once a day only, sounds good to me.

An ofc the messages should be printed in a nicer way, just did a simple output so far!

manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+    "messages": {
+        "infos": [
+            {
+                "message": "Ensure maintenance and further development of DDEV see https://github.com/sponsors/rfay."
```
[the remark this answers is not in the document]
gilbertsoft: I have an open task to define the messages see above. This one is meant for testing only at the moment. But will commit your suggestion later....

pkg/github/github.go
gilbertsoft: Like mentioned in the general feed back, I've used the latest version for the new implementation.

pkg/github/github.go
gilbertsoft: It's mostly the same. I've prepared a global factory function which should be used later everywhere. See also general feedback.

pkg/manifest/manifest.go
gilbertsoft: Sure, we also can place it in a dedicated repo. The messages feature is the first one in the manifest and like written in the docs this could also be used for other stuff in the future. That's why I have chosen a generic name. But if we use a dedicated repo, we could change it to messages.json or sth similar.

At the moment the owner/repo/branch is hard coded but would make sense to make it configurable also for testing, true.

pkg/manifest/manifest.go
gilbertsoft: Yes, is currently mainly for testing purposes, Debug would be fine here too!

pkg/manifest/messages.go
gilbertsoft: See answer above.

pkg/manifest/storages/file_storage.go
gilbertsoft: Interesting, didn't think about, just used the terms used by the Amplitide package. Will change....

pkg/manifest/storages/file_storage.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+type fileStorage struct {
+	fileName string
+	loaded   bool
+	data     fileStorageData
+}
+
+// fileStorageData is the structure used for the file.
```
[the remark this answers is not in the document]
gilbertsoft: A lot of comments will follow.... Also I like to have a description in each package what it is meant for.

pkg/manifest/storages/file_storage.go
gilbertsoft: Pull and Push pulls and pushes messages from/to the storage. Terms are imho fine and widely used.

pkg/manifest/manifest.go
gilbertsoft: go starts the method asynchronously, non blocking though. This has the effect the update will be done in the background withouth having the user to wait for. Commented here for testing purposes, is now active again.

rfay:
Requests from our meeting today:

- [ ]  Improve the upstream amplitude-go PR with an example of what could go wrong and more detail. We want them to understand it and pull it. 
- [x] In the fork, https://github.com/ddev/analytics-go#readme, replace the README.md with an explanation of the reason for the fork, the link to the PR, etc. https://github.com/ddev/ddev/pull/4866
- [ ]  Add the PR ink to the OP of this PR

gilbertsoft:
> Requests from our meeting today:
> 
>     * [ ]   Improve the upstream amplitude-go PR with an example of what could go wrong and more detail. We want them to understand it and pull it.
> 
>     * [x]  In the fork, https://github.com/ddev/analytics-go#readme, replace the README.md with an explanation of the reason for the fork, the link to the PR, etc. [Deliver instrumentation data directly to amplitude, fixes #4335 #4866](https://github.com/ddev/ddev/pull/4866)
> 
>     * [x]   Add the PR ink to the OP of this PR
> 
>     * [ ]  (And please rebase, again :) )

This belongs to #4866

gilbertsoft:
@rfay what about `remote config` instead of `manifest`?

rfay:
"remoteconfig" sounds OK to me. 

I just realized that the time interval between showing messages should/could be specified in the manifest.json/remoteconfig.json (preferably remoteconfig.jsonc if we have a way to successfully read jsonc). Then we'd be able to introduce more important warnings, etc.

pkg/github/github.go
gilbertsoft: Solved by https://github.com/ddev/ddev/pull/4915

pkg/github/github.go
gilbertsoft: Solved by https://github.com/ddev/ddev/pull/4915

pkg/manifest/manifest.go
gilbertsoft: Dedicated repo https://github.com/ddev/remote-config

rfay:
Rebased. Is this ready for review?

rfay:
I see this on `ddev start` with latest; neither of these messages are in remote-config.

image

gilbertsoft:
> I see this on `ddev start` with latest; neither of these messages are in remote-config.
> image

These messages are currently hard coded for testing purposes. Will remove 'em again with the next push and add some test messages to the remote-config repo.

rfay:
We have an interesting fragile test in https://github.com/ddev/ddev/actions/runs/5257438889/jobs/9500222563?pr=4885

TestHostDBPort - 

```
2023-06-13T16:31:58.481 Ensuring write permissions for TestPkgDrupal9
    ddevapp_test.go:3776: 
        	Error Trace:	/home/runner/work/ddev/ddev/pkg/ddevapp/ddevapp_test.go:3776
        	Error:      	Not equal: 
        	            	expected: "DDEV_HOST_DB_PORT=9998"
        	            	actual  : "2023-06-13T16:31:58.751 Downloading remote config."
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-DDEV_HOST_DB_PORT=9998
        	            	+2023-06-13T16:31:58.751 Downloading remote config.
        	Test:       	TestHostDBPort
        	Messages:   	dbType=postgres, out=2023-06-13T16:31:58.751 Downloading remote config.
        	            	DDEV_HOST_DB_PORT=9998
        	            	, err=<nil>
```

I guess it happens here because that test has to change the global_config.yaml? Or might it disrupt tests in other places?

rfay:
I just noticed ~/.ddev/state.yaml - 

I think this should probably be ~/.ddev/.state.yaml - it's not anything that anybody should ever look at ...

mattstein:
Is there a way to disable this for users that don’t want a chattier terminal experience?

docs/content/developers/remote-config.md
```diff
@@ -0,0 +1,97 @@
+# Remote config
```
mattstein, line 1: Capitalize heading:

```suggestion
# Remote Config
```

docs/content/developers/remote-config.md
```diff
@@ -0,0 +1,97 @@
...
+
+DDEV supports to download a remote config from the `ddev/remote-config` Github
+repository. At the moment the remote config contains messages which will be
+shown to the user see next chapter but can also be easily enhanced in the
+future with more information and configuration.
```
mattstein, line 6: Shorten and link to repository:

```suggestion
DDEV supports downloading a remote config from the [`ddev/remote-config`](https://github.com/ddev/remote-config) GitHub repository with messages that will be shown to the user. This feature could be enhanced later with more information and configuration.
```

docs/content/developers/remote-config.md
```diff
@@ -0,0 +1,97 @@
...
+## Messages
+
+### Notifications
+
+The defined messages are shown to the user every `interval` as long as not
+`disabled`. Supported message types are `infos` and `warnings` where `infos`
+are printed in a yellow box and `warnings` in a red box.
+
+The user can not influence notifications, they will be shown like defined in
+the `remote-config` repository.
```
mattstein, line 72:
```suggestion
Messages will be shown as configured in the `remote-config` repository and the user cannot influence them.
```

docs/content/developers/remote-config.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Ticker
+
+One of the defined messages is shown to the user every `interval` as long as
+not `disabled`, the messages are rotating.
```
mattstein, line 77: Shorten, improve flow:

```suggestion
Messages rotate, with one shown to the user every `interval` as long as it’s not `disabled`.
```

mattstein (changes requested):

docs/content/users/configuration/config.md
```diff
@@ -237,6 +237,22 @@ Port for project’s MailHog HTTP URL.
...
+| :octicons-globe-16: global | `false` | Can be `true` or `false`; only `true` has any effect.
+
+## `messages:ticker_interval`
+
+How often ticker messages should be shown.
+
+| Type | Default | Usage
+| -- | -- | --
+| :octicons-globe-16: global | `20` | Interval in hours.
+
```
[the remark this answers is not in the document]
gilbertsoft: Agree to combine the two, that's sound good.

The user shouldn't change this setting in the first place and that's why I don't have provided a config command flag.

And this change introduces sub structures in the yaml. So far we only had one level with all options. To have cleaner yaml files I've introduced this sub structures like we already know e.g. from docker-compose. This is what the ":" should outline.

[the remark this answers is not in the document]
mattstein: My understanding is that this YAML config is for any user, and DDEV doesn’t require any familiarity with Docker and Docker Compose for basic setup. So while I understand the `:` should be a familiar thing from Docker Compose, it’s still a new convention in this format. (Commands like `service` have sub-commands, so there’s sort of a precedent for that already.) That’s why I dropped `messages:` from my suggestion altogether, even though I like the idea of using that convention if it’s done more broadly. A one-off change seems more confusing.

If the user shouldn’t change this setting, wouldn’t we want to tell them that directly? Or _not_ document the setting?

I may be part of a rare and needlessly grumpy contingent, but I would immediately disable any messaging that adds noise to the experience without any clear value to me. (Which would probably be concise new feature announcements, or specific suggestions based on how I use DDEV.) This might be one of the first things I figure out how how to disable as a new user. Or it might motivate me to contribute to the project, with the perk of quieting CTAs exactly like this.) So if I’m to change this setting, all the more helpful that I could change it like most other values: as a command flag or YAML setting.

Do whatever you will (or not!) with this feedback, I’m just very sensitive about trying not to take the user’s attention for granted and this feature appears to introduce marketing messages into the daily DDEV experience.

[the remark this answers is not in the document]
gilbertsoft: Btw most messages should be about the usage of DDEV, no marketing at all. Only one message to make people more sensitive to sponsor DDEV is planed so far.

pkg/globalconfig/global_config.go
```diff
@@ -341,6 +364,19 @@ func WriteGlobalConfig(config GlobalConfig) error {
...
+# Controls the display of the ticker messages.
+
+# remote_config:
+#   update_interval: 10 // Interval in hours to download the remote config
+#   remote:
+#     owner: ddev
+#     repo: remote-config
+#     ref: main
+#     filepath: remote-config.jsonc
+# Controls the download of the remote config. Please do not change.
```
gilbertsoft, line 378: @mattstein can you please also have a look at these descriptions, thanks!

mattstein: Looks good to me!

rfay:
Please add manual testing instructions to this, thanks!

rfay:
I added a bunch of messages to https://github.com/ddev/remote-config/commit/b9ba83bd53b621d69c0fa46872fb6f41c143715b

pkg/config/remoteconfig/messages.go
```diff
@@ -0,0 +1,283 @@
...
+	conditionFunc func() bool
+}
+
+var conditionDefinitions = map[string]conditionDefinition{}
+
+func init() {
+	AddCondition("Disabled", "Permanently disables the message", func() bool { return false })
+	AddCondition("Colima", "Running on Colima", dockerutil.IsColima)
+	AddCondition("DockerDesktop", "Running on Docker Desktop", dockerutil.IsDockerDesktop)
+	AddCondition("WSL2", "Running on WSL2", dockerutil.IsWSL2)
```
gilbertsoft, line 35: @rfay any other condition we should add?

rfay: I wish we could consider Colima version and Docker Desktop version, but we don't have easy access to them in golang code at this point.

How about Gitpod and Codespaces while we're here, and OSType?

rfay:
I haven't yet been able to make this work or get any kind of notification. We should do a screenshare to look at it tomorrow.

However, when it does work, it should absolutely not be creating any output except on `ddev start` IMO.

rfay:
Yay, it worked, but we'll probably have to figure out something about wrapping. 

image

rfay:
Do you mind if I push this into the PR? 

```
diff --git a/pkg/config/remoteconfig/messages.go b/pkg/config/remoteconfig/messages.go
index 605aaa54a..c3e649ba9 100644
--- a/pkg/config/remoteconfig/messages.go
+++ b/pkg/config/remoteconfig/messages.go
@@ -265,19 +265,27 @@ func applyTableStyle(preset preset, writer table.Writer) {
 		},
 	})
 
+	style := writer.Style()
+
+	style.Options.SeparateRows = false
+	style.Options.SeparateFooter = false
+	style.Options.SeparateColumns = false
+	style.Options.SeparateHeader = false
+	style.Options.DrawBorder = false
+
 	switch preset {
 	case information:
-		writer.Style().Color = table.ColorOptions{
+		style.Color = table.ColorOptions{
 			Header: text.Colors{text.BgHiYellow, text.FgBlack},
 			Row:    text.Colors{text.BgHiYellow, text.FgBlack},
 		}
 	case warning:
-		writer.Style().Color = table.ColorOptions{
+		style.Color = table.ColorOptions{
 			Header: text.Colors{text.BgHiRed, text.FgBlack},
 			Row:    text.Colors{text.BgHiRed, text.FgBlack},
 		}
 	case ticker:
-		writer.Style().Color = table.ColorOptions{
+		style.Color = table.ColorOptions{
 			Header: text.Colors{text.BgHiWhite, text.FgBlack},
 			Row:    text.Colors{text.BgHiWhite, text.FgBlack},
 		}
```

rfay:
Per discussion in remote-images, let's go ahead and detect terminal width and wrap. I actually think the table code has this built in, and it will work fine for single-column stuff like this.

rfay:
(Working on wrapping right now)

rfay:
I think the remote config isn't created if it doesn't exist and the .state.yaml says that it might have been updated. I doubt this much matters. But it breaks testing.

rfay:
I'm pretty sure these notes would be more effective at the *beginning* of `ddev start` instead of at the end.

rfay:
I took the liberty to push a wrapping change in in https://github.com/ddev/ddev/pull/4885/commits/e6bb8bf5f4b23679486f2aed88738b1541dd3afe, we can revert if you don't like it. But here's what it looks like:

image

image

gilbertsoft:
> I took the liberty to push a wrapping change in in [e6bb8bf](https://github.com/ddev/ddev/commit/e6bb8bf5f4b23679486f2aed88738b1541dd3afe), we can revert if you don't like it. But here's what it looks like:
> image image

Looks good, thanks!

cmd/ddev/cmd/debug-message-conditions.go
```diff
@@ -0,0 +1,38 @@
...
+			t.AppendRow(table.Row{condition, description})
+		}
+
+		t.SortBy([]table.SortBy{
+			{Name: "Condition", Mode: table.Asc},
+		})
+
+		output.UserOut.WithField("raw", conditions).Print("Supported conditions for messages are:\n\n", t.Render())
+	},
+	Hidden: true,
```
rfay, line 33: I like that you hid this. The only thing is how will we ever find it!

docs/content/developers/remote-config.md
```diff
@@ -0,0 +1,95 @@
+# Remote Config
+
+DDEV supports downloading a remote config from the [`ddev/remote-config`](https://github.com/ddev/remote-config)
+GitHub repository with messages that will be shown to the user. This feature
+could be enhanced later with more information and configuration.
+
+Here is an example of a `remote-config.jsonc` file:
```
rfay, line 7: This might be better just as a link to the remote-config repo.

internal/build/remote_config.go
```diff
@@ -0,0 +1,9 @@
+package build
```
rfay, line 1: Thanks for moving to `pkg`. I think `build` isn't a good enough name here. Maybe `remoteconfig` ?

pkg/config/remoteconfig/config.go
```diff
@@ -0,0 +1,91 @@
+package remoteconfig
```
rfay, line 1: Thanks for moving

rfay (changes requested):
Minor change requests, this seems to be working great. It's a really heavy load for what we expected, but I know you have great hopes for some of the things you added. Maybe we'll be able to use them where they can be used.

I tested with various terminal window sizes, and tested with internet disabled, and all seems to go well.

internal/build/remote_config.go
gilbertsoft: build is fine because it's about vars to be replaced during build/linking.

docs/content/developers/remote-config.md
gilbertsoft: Agree, on place less to maintain.

cmd/ddev/cmd/debug-message-conditions.go
gilbertsoft: Therefor it's mentioned in the dev docs see https://github.com/ddev/ddev/pull/4885/files#diff-e7209a6b4e94d84c947308700da7f20a6190c6b8598adb0a0e119e58d9224b90R84

rfay (approved):
Thanks for all the hard work on this, let's get it in. Please pull it when you're ready.  Don't worry about the wsl failure, I'm sure it would clear, not related to this anyway.

gilbertsoft:
> Thanks for all the hard work on this, let's get it in. Please pull it when you're ready. Don't worry about the wsl failure, I'm sure it would clear, not related to this anyway.

Yes, had to restart the WSL test system, was completely broken.

PR #4886, opened 2023-05-08
Remove the vendor folder

gilbertsoft:
## The Issue

The vendor folder was added in the early days because of issues with the package management. In the meantime there are no known issues anymore and the vendor folder can be removed again.

This patch removes the vendor folder and adds it to the .gitignore file.


rfay:
I love the fact that when I try to update a package I can see the exact results in git.

PR #4903, opened 2023-05-11
Add “just simply” link to the writing style guide

mattstein:
This adds an excellent link to the “avoid just and it’s easy” section of the writing style guide that further describes the problem and potential solutions with that sort of language.

Here’s a direct link to it in the preview build: https://ddev--4903.org.readthedocs.build/en/4903/developers/writing-style-guide/#avoid-just-and-easy

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4903"></a>

PR #4905, opened 2023-05-12
[docs] Add textlint

mattstein:
## The Issue

Nobody wants to read a [writing style guide](https://ddev.readthedocs.io/en/latest/developers/writing-style-guide/).

This introduces the [same textlint rule set](https://github.com/ddev/ddev.com-front-end/blob/main/.textlintrc) from the ddev.com project, including a GitHub Actions check, and fixes most of the errors.

This project doesn’t use a `package.json`  file so it’s not a quick `npm install`, but you can install textlint and its rulesets globally like it suggests in the Makefile:

```
npm install -g textlint textlint-filter-rule-comments textlint-rule-no-todo textlint-rule-stop-words textlint-rule-terminology
```

Then run `textlint {README.md,version-history.md,docs/**}` or `make textlint` and have a wondrous time tidying up. (The [VS Code extension](https://marketplace.visualstudio.com/items?itemName=taichi.vscode-textlint) works well!)

## Automated Testing Overview

It tests public-facing plain text in `README.md`, `version-history.md`, and everything in the `docs/` directory.

## Related Issue Link(s)

- https://github.com/ddev/ddev/pull/4903
- https://github.com/ddev/ddev.com-front-end/pull/51
- https://github.com/ddev/ddev.com-front-end/issues/31#issuecomment-1409617516

## Release/Deployment Notes

Won’t mangle DDEV builds or deployment, but I have no idea if the GitHub Action works yet so we’ll be finding out together.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4905"></a>


rfay (commented):
This is just amazing, should save you just hours of reviewing time. Thanks!

I note that some of the rules are inflexible and can force unnecessary changes, but we should just accept that.

mattstein:
@rfay I found a way to _not_ take forever on this! No change to the spell checker, only a sparing introduction of textlint.

Trying to get a simpler GitHub Action working.

I have a rule set that rejects unfinished todo items and the only thing failing here is [this line from the version history](https://github.com/ddev/ddev/blob/master/version-history.md?plain=1#L112). (I also don’t know why that file uses checkboxes and bullets.) How can I fix this?

.github/workflows/docscheck.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: rojopolis/spellcheck-github-actions@v0
         name: Spellcheck
+      - uses: rdohms/textlint-action@latest
+        with:
+          rulesets: 'textlint-filter-rule-comments textlint-rule-no-todo textlint-rule-stop-words textlint-rule-terminology'
+          path: '{README.md,version-history.md,docs/**}'
```
rfay, line 34: IMO version-history.md is not something that needs to be checked at all. Just meant extra work for you in this. It's *not* included in docs because it corrupts the search so much.

mattstein: I don’t mind since it’s both public-facing and important, and I figure it should be held to the same standard as writing in the documentation. (Same with the developer docs this also fixes—though the spell checker would go wild there if we stopped ignoring that section.)

Are you saying you’d like me to remove that from the check? The fixing-up is already done so after this PR it’s only a matter of maintenance/conformance.

rfay:
> I have a rule set that rejects unfinished todo items and the only thing failing here is [this line from the version history](https://github.com/ddev/ddev/blob/master/version-history.md?plain=1#L112). (I also don’t know why that file uses checkboxes and bullets.) How can I fix this?

Easiest way is to exclude version-history.md. 

Second easiest for unfinished checkboxes is to check the box, since it eventually got done. (Not sure why that was listed there because it was not accomplished at that time. (Powers that be in drud wanted it to require a NDA of participants, so I balked and put it off until drud went away.)

Third easiest is to allow unchecked checkboxes.

.github/workflows/docscheck.yml
rfay: Thanks for checking the box. The other option would have been to remove the line, which would have been completely and historically correct.

mattstein: Whoops! I was going for correctness—removed the line instead. The fact that we’re discussing style and accuracy is a reason to include this file, IMO!

mattstein:
@rfay I can’t tell if I managed to break actual important tests or if they’re just having a bad day, but the textlint GitHub Action works now and this is ready for review.

rfay (approved):
Awesome, thanks so much! That was a *lot* less than a week!

WSL2 Docker Desktop is mostly broken since latest release. I'll look for other things.

mattstein:
If you’re bored, try setting up Vale and throw its `alex` package at the docs. It’d be a noble thing to upgrade the spell checker, but it’s a wildly different amount of work. 😅

mattstein:
As usual, merge if/when you’re comfortable with this.

rfay:
Thanks! I'll probably let it finish most tests since it touches a variety of things in case anything has gone wrong.

PR #4908, opened 2023-05-12
Remove unused package spf13/viper and dependencies from vendor, etc.

gilbertsoft:
## The Issue

The usage of package `github.com/spf13/viper` was removed with https://github.com/ddev/ddev/commit/f24e8e5c7abf1a0f7c8a04d8fbcc915a753c0bec 6 years ago but one line was forgotten which had no effect anymore.

This PR removes that last line and the package with its dependencies.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4908"></a>


rfay (commented):
Sure looks good to me! Amazing.

gilbertsoft:
> Sure looks good to me! Amazing.

Reduces the binary size for almost 2MB :)

rfay:
That's weird, the binary shouldn't have been affected, but I guess because of that one single use!

gilbertsoft:
> That's weird, the binary shouldn't have been affected, but I guess because of that one single use!

Yes it does, that's a known problem of Viper. In our case not that big because we already import some dependencies which Viper would do. See also https://github.com/knadh/koanf#alternative-to-viper and https://github.com/knadh/koanf/wiki/Comparison-with-spf13-viper

rfay:
Thanks for this! Is it going to mess up your vendor dir for the amplitude patch? Almost pulled, then thought about that. Would you rather do it after that goes in?

gilbertsoft:
> Thanks for this! Is it going to mess up your vendor dir for the amplitude patch? Almost pulled, then thought about that. Would you rather do it after that goes in?

No, should do any harm. You can merge at any time.

PR #4913, opened 2023-05-16
[docs] fix text to match example [skip ci]

tyler36:
## The Issue
The "desired" version does not match the example given.

This PR updates the desired version to match the example.
It also changes the second example to match the same version. This helps to hightlight the code differences.


mattstein (approved):
🤦‍♂️ Thank you!

rfay (approved):
Thanks, good eye!

Issue #4912, opened 2023-05-15
Possibly add mkcert step for Debian/Ubuntu installation, and Firefox tip about Master password?


gitressa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

Follow the [Debian/Ubuntu installation steps](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#debianubuntu), and get no certificate warnings in Firefox and Chromium.

### Actual Behavior

Got certificate warnings in both Firefox and Chromium.

### Steps To Reproduce

Great project, thanks @rfay! I especially like how there's [built in Solr Cloud integration for Search API Solr](https://git.drupalcode.org/project/search_api_solr#setting-up-solr-cloud-the-modern-way).

- Ubuntu 20.04
- Docker version 23.0.6, build ef23cbc

I have previously had DDEV installed, which may have caused this. But I thought I'd share it anyway, since others may run into this, so they will find this issue and a solution.

If you follow the [Debian/Ubuntu installation steps](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#debianubuntu), and open your Drupal 10 site at https://drupal10.ddev.site/ you may get this warning in Firefox and Chromium:

## Firefox

> ### Warning: Potential Security Risk Ahead
> 
> Firefox detected a potential security threat and did not continue to drupal10.ddev.site. If you visit this site, attackers could try to steal information like your passwords, emails, or credit card details.
> 
> What can you do about it?
> 
> The issue is most likely with the website, and there is nothing you can do to resolve it.
> 
> If you are on a corporate network or using antivirus software, you can reach out to the support teams for assistance. You can also notify the website’s administrator about the problem.
> 
> Someone could be trying to impersonate the site and you should not continue.
>  
> Websites prove their identity via certificates. Firefox does not trust drupal10.ddev.site because its certificate issuer is unknown, the certificate is self-signed, or the server is not sending the correct intermediate certificates.
>  
> Error code: `SEC_ERROR_UNKNOWN_ISSUER`

## Chromium

> ### Your connection is not private
> 
> Attackers might be trying to steal your information from drupal10.ddev.site (for example, passwords, messages, or credit cards). Learn more
> `NET::ERR_CERT_AUTHORITY_INVALID`
> To get Chrome’s highest level of security, turn on enhanced protection
[...]

PR #4914, opened 2023-05-16
[docs] Move mkcert to the top for Linux [skip ci]

gitressa:
Solves #4912.

## The Issue

The mkcert command is placed at the bottom, under the Arch Linux section, but it is a requirement for all Linux installations. 

It moves the mkcert info to the beginning of the Linux section, since it is a requirement for all Linux installations. Repeating it under each Linux distro would break with the DRY principle.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4914"></a>

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -45,6 +45,15 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ## Linux
 
+    ### Locally-trusted certificate with mkcert
+
+    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB` which is the Firefox Master pin. Install it, and then run this:
+
+    ```bash
+    # Initialize mkcert
+    mkcert -install
+    ```
+
     ### Debian/Ubuntu
 
     DDEV’s Debian and RPM packages work with `apt` and `yum` repositories and most variants that use them, including Windows WSL2:
@@ -92,9 +101,6 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     ```bash
     # Install DDEV
     yay -S ddev-bin
-
-    # Initialize mkcert
-    mkcert -install
     ```
 
     ### Homebrew (AMD64 only)
@@ -102,9 +108,6 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     ```bash
     # Install DDEV using Homebrew
     brew install ddev/ddev/ddev
-
-    # Initialize mkcert
-    mkcert -install
     ```
 
     <!-- we’re using HTML here to customize the #install-script-linux anchor -->
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -47,7 +47,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ### Locally-trusted certificate with mkcert
 
-    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB` which is the Firefox Master pin. Install mkcert, and then run this:
+    Modern browsers require valid certificates, which mkcert can create. [Install mkcert](https://github.com/FiloSottile/mkcert#installation), and then run this:
 
     ```bash
     # Initialize mkcert
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -47,7 +47,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ### Locally-trusted certificate with mkcert
 
-    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB` which is the Firefox Master pin. Install it, and then run this:
+    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB` which is the Firefox Master pin. Install mkcert, and then run this:
 
     ```bash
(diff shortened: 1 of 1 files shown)


docs/content/users/install/ddev-installation.md
```diff
@@ -45,6 +45,15 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ## Linux
 
+    ### Locally-trusted certificate with mkcert
+
+    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB` which is the Firefox Master pin. Install mkcert, and then run this:
```
mattstein, line 50: Reading the issue you linked to, looks like this is missing a quote, “master” is not a proper noun that needs to be capitalized, and it would be most helpful to refer to it as a “password” instead of re-using the word “pin” from the description. Also worth adding a link for “install mkcert”:

```suggestion
    Modern browsers require valid certificates, which mkcert can create. Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB"` which is the Firefox master password. [Install mkcert](https://github.com/FiloSottile/mkcert#installation), and then run this:
```

mattstein (commented):
Thanks for this, @gitressa!

I’m usually a fierce advocate of DRY in principle, and these instructions are distinctly not following that which is my intent after a [recent-ish PR](https://github.com/ddev/ddev/pull/4789).

Regardless of whether your new section stays, which seems reasonable because of Firefox, my preference would be to allow each distro’s instructions to be self-contained—in other words move `mkcert -install` above DDEV’s install line if that’s what’s needed. Possibly include mkcert installation for each distro—`sudo apt-get -y install mkcert` for Debian/Ubuntu, for example, if that’s not overly specific.

My thinking is that most people on this page are looking to install DDEV on their distro, not understand how DDEV can be installed on every Linux distro, and therefore instructions for each one should be brief and self-contained. If someone needs to read every part of the page and adventure to other docs in order to install DDEV, I imagine the docs could be making their life easier.

That said, the fact that you’re offering this PR may mean I was wrong.

I’ll leave the decision to @rfay—thanks again for improving these instructions!

docs/content/users/install/ddev-installation.md
rfay, line 50: I have never seen a request for a master PIN or anything of the type on Firefox and that issue (https://github.com/FiloSottile/mkcert/issues/50) is very old. Please remove this from the PR and open an issue with your specific problem.  The mkcert docs problem is one thing, and it definitely should be more prominent, but the issue you're having with Firefox is completely unknown to me, it has never come up in all these years.

rfay (changes requested):

docs/content/users/install/ddev-installation.md
gitressa: Firefox sentence is gone.

gitressa:
Thanks @mattstein and @rfay! I updated the merge request.

I agree that it would be nice to include install commands for each Linux distro. On the other hand, some might want to install with another method. I for example chose to use the pre-built binaries method, for some reason.

docs/content/users/install/ddev-installation.md
gitressa: PS. It just occurred to me that maybe you don't use a Master password in Firefox and this could be the reason?

rfay: I had never heard of "Master Passwords", or heard of anybody using them. But I guess [they're now called "Primary Password](https://support.mozilla.org/en-US/kb/use-primary-password-protect-stored-logins). But this Firefox feature shouldn't have anything to do with DDEV, unless you allow FF to remember your passwords for local sites. Were you saying that it had to do with allowed CAs? (mkcert installs a CA).

gitressa: I think the easiest method is if you try to enable Primary Password in Firefox, and remove and recreate certificates. Then you should see the cryptic `Enter Password or Pin for "NSS Certificate DB":` message for yourself. It doesn't have anything to do with DDEV:

> ... the prompt comes from NSS's `certutil` not `mkcert`. I'm not sure if that prompt can be changed.

https://github.com/FiloSottile/mkcert/issues/50#issuecomment-452013317

I only added the tip here because I spent some time figuring out what it meant. But if you think it's a bad idea, we shouldn't add it.

rfay: Would you consider a PR to certutil? (But I'm not sure at a quick glance where it comes from). I'm glad you sorted it out. It might be best as a Stack Overflow question/answer? You explain the problem as a question, then click the checkbox to answer your own question. Then you can maintain it as it changes, etc. 

It certainly is an obscure prompt.

gitressa: Heh, there's actually already a PR in progress, so I think we're good here.
https://github.com/FiloSottile/mkcert/pull/100

rfay: Unfortunately, as you see, he doesn't stop by the mkcert project very often. Glad you added your "me too" there.

gitressa:
> @rfay: Unfortunately, as you see, he doesn't stop by the mkcert project very often. Glad you added your "me too" there.

Yes ... let's hope that the Firefox tip issue in `mkcert` gets resolved soon. Otherwise we can revisit this issue and consider adding the removed tip.

> Firefox users [may get prompted](https://github.com/FiloSottile/mkcert/issues/50) for `Password or Pin for "NSS Certificate DB"` which is the Firefox master password

Are the doc files getting generated on a cyclical basis (like a house keeping task) every so often?

gitressa:
I got served an old version from February 2023, but now I see the updated version. Weird.

rfay:
There are 2 versions of docs. "stable", which is latest release (which would be v1.21.6) and "latest", which builds from the most recent commits on master.

rfay:
We've struggled with this for some time. The problem is that most people use the stable version of DDEV. So they would find features that don't exist in the "latest" version. OTOH kind people like you come along and fix docs that affect the current stable, so it would be best that the stable docs be updated in that case :)

rfay:
I didn't think you were harsh. It's a problem we've thought about a long time but haven't figured out a path for.

rfay:
Of course, yeah I see everything that happens on that repo.. but not sure how to make progress. He should really add a co-maintainer, but haven't seen that happening either.

PR #4915, opened 2023-05-16
Update github.com/google/go-github to v52

gilbertsoft:
## The Issue

The used package github.com/google/go-github is very outdated and must be updated.

This PR updates the package and makes minor changes in the code because of some breaking changes.


rfay (changes requested):
My intent was that you just upgrade to current go-github. Please don't specify 52 in there. Maybe there's something I don't understand? Just using go-github should get you the right version, and it will be locked until we upgrade it again.

`go get -u github.com/google/go-github` - no changes to imports should be required.

gilbertsoft:
> My intent was that you just upgrade to current go-github. Please don't specify 52 in there. Maybe there's something I don't understand? Just using go-github should get you the right version, and it will be locked until we upgrade it again.
> 
> `go get -u github.com/google/go-github` - no changes to imports should be required.

That does not work with the new modules feature of go see https://github.com/google/go-github#installation

rfay:
This looks fine to me then and can be merged after `ddev get` and the version/update checking has been manually tested. Just let me know.

mattstein (approved):

rfay:
I manually tested `ddev get --list` and update check in root.go, and both seemed right.

gilbertsoft:
> OK, and I see the v52 is what they suggest.
> 
> Don't know what they could be saying about modules being enabled; we never used it any time in its history without modules enabled.

As far as I understand modules is a feature since 1.13 and enabled by default. But not all packages support it by default that's why some packages are requested without the versin suffix.

PR #4917, opened 2023-05-16
Bump docker-compose to v2.18.1

rfay:
## The Issue

Time to catch up to current docker-compose version

Bump to composer 2.18.1

See also
* https://github.com/ddev/ddev/issues/4727

Use it. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4917"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1899,7 +1899,7 @@ func (app *DdevApp) DockerEnv() {
 
 	envVars := map[string]string{
 		// The compose project name can no longer contain dots; must be lower-case
-		"COMPOSE_PROJECT_NAME":          strings.ToLower("ddev-" + strings.Replace(app.Name, `.`, "", -1)),
+		"COMPOSE_PROJECT_NAME":           strings.ToLower("ddev-" + strings.Replace(app.Name, `.`, "", -1)),
 		"COMPOSE_CONVERT_WINDOWS_PATHS":  "true",
 		"COMPOSER_EXIT_ON_PATCH_FAILURE": "1",
 		"DDEV_SITENAME":                  app.Name,
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1898,9 +1898,8 @@ func (app *DdevApp) DockerEnv() {
 	}
 
 	envVars := map[string]string{
-		// The compose project name can no longer contain dots
-		// https://github.com/compose-spec/compose-go/pull/197
-		"COMPOSE_PROJECT_NAME":           "ddev-" + strings.Replace(app.Name, `.`, "", -1),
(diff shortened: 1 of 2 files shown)


gilbertsoft (changes requested):
Sounded promising but makes it even worse here. Also with a small project.

rfay:
@gilbertsoft *what* does it make worse, and in what way?

gilbertsoft:
> @gilbertsoft _what_ does it make worse, and in what way?

`docker-compose build` takes much longer now. It was already bad before since some versions, not sure when it changed to bad but some time ago `ddev start` was much faster than it is at the moment.

rfay:
So you're saying that you tested this PR and were able to detect the difference in timing on build? Did you use `ddev debug refresh` to test it? Can you give an example of what your experience was?

There haven't been reports of build being the problem here; dockerr-compose was frozen at 2.15.1 because of #4727

gilbertsoft:
> So you're saying that you tested this PR and were able to detect the difference in timing on build? Did you use `ddev debug refresh` to test it? Can you give an example of what your experience was?
> 
> There haven't been reports of build being the problem here; dockerr-compose was frozen at 2.15.1 because of #4727

For the test I did:

```
ddev start
ddev stop
ddev start
```

The first start asked for a poweroff which I did and that's why I did a stop and start again afterwards to have clean results only for the start with router and ssh agent already started and the docker-compose binary already downloaded and preset in the desired version.

To compare the time I've taken the last time track start before build and the next perf output after build:

v1.21.6 (16:30:54.899870360 -> 16:31:13.136306597 ~ 8.3 seconds) 
v1.22.0-alpha2-60-gfa7a95f0 (16:23:16.347786863 -> 16:23:35.050146960 ~ 18,7 seconds)

I didn't do a debug refresh for the moment.

rfay:
BTW, this can actually be tested easily without actually building; you can use `ddev config global --required-docker-compose-version=v1.18.0` and that should be adequate to switch back and forth. 

I do see in https://docs.docker.com/compose/environment-variables/envvars/ that we're using some unsupported environment variables still in https://github.com/ddev/ddev/blob/c6518c33c93f6510d22cc43f297798308305a69c/pkg/ddevapp/ddevapp.go#L1903-L1910 - but it doesn't seem like it would affect this.

rfay:
I'm not seeing a difference in `ddev start` times on mac/colima, will try other environments:

image

rfay:
@gilbertsoft I'm unable to recreate the problems you saw. I've done a full comparison at https://docs.google.com/spreadsheets/d/1SXlXanCoVw0SS_-viLueJ7DdJdjQUjrBT6819OW8pV4/edit?usp=sharing

The good news here is the numbers are *all* fine, and the macOS Docker Desktop numbers, which were horrible and probably the reason I was worrying about recent regressions, were fixed by a factory reset. 

image

PR #4924, opened 2023-05-17
Simplify and optimize util.TimeTrack

gilbertsoft:
## The Issue

The util.TimeTrack() function has currently a parameter start which was always set to time.Now() and can therefor set directly in the function instead which will reduce the impact a bit and simplify the usage. Also in many cases the name parameter was manually set to the calling function name which is also error prone.

With this patch, the TimeTrack functionallity was moved to a dedicated file in the same package and two functions are now available, one without parameters where the calling function name is retrieved with the help of the runtime package and a second with a name parameter to individually set a name e.g. used for tests. Source documentation was extended to also show a simpler one line usage which is apropriate in most cases. Also some minor optimization are done like avoiding the expensive format functions e.g. Printf. Finally the outputed messages are streamlined to increase visibility.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4924"></a>

### DIFF
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -8,7 +8,6 @@ import (
 	"runtime"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/exec"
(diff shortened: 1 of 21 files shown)


rfay (approved):
Looks great, thanks for improving this. And I forget this feature even exists, it will be useful for studying my worries about mutagen performance.

rfay:
Also thanks for separating this out into its own PR. It's always great for PRs to have their own specific purpose.

Issue #4650, opened 2023-02-14
Don't manage Craft 3 .env, Fix `*_SITE_URL`


bossanova808:
The recently added settings management for Craft projects broke my ddev / **Craft 4** project - when I came back to it after some time away - and in a way I couldn't immediately understand from this error:

```
Failed to start craft-dev: template: templateStringToFile:/home/jeremy/ddev-projects/c4-dev/.env:7: function "siteUrl" not defined
```

Prior to this the project was working fine for a few months, it was only on coming back to it that the new settings management kicked in, with a ddev upgrade.  

Thus, I think `disable_settings_management ` should definitely default to `false` for _existing_ projects. 

As per the other issue discussion, I think `PRIMARY_SITE_URL` should be added only if the existing & typical `CRAFT_SITE_URL` is not already present - otherwise it's just doubling up right?

The config setting is the default way to define this in Craft:
https://craftcms.com/knowledge-base/how-craft-determines-the-site-url
https://craftcms.com/docs/2.x/config-settings.html#siteurl

And thus, with Craft 4's new approach to .env variables, `CRAFT_SITE_URL` is the de facto default way of actually implementing this.

Personally I am not super fussed about Mailhog, other than to note that Craft has a config setting for a similar thing already:
https://docs.craftcms.com/v3/config/config-settings.html#testtoemailaddress

```
        'testToEmailAddress' => (\craft\helpers\App::env('SEND_ALL_EMAILS_TO')) ?? null,
```

So in Craft's case, Mailhog is likely to be irrelevant to most and I would expect that to be set as a matter of course in `.env` files for dev environments.

_Originally posted by @bossanova808 in https://github.com/drud/ddev/issues/4552#issuecomment-1429132626_
            
(Adding as a new issue per request)

rfay:
I can certainly understand why you would want `disable_settings_management` to default given your experience, but thousands of people have been using it with automatic settings management for years, and there are advantages to it as well, people like it quite a lot that they can spin up a project with no knowledge of settings. The full docs on settings are at https://ddev.readthedocs.io/en/latest/users/usage/cms-settings/, explaining all of the possibilities.

As far as the current usage (aimed at Craft 4), here is the current logic and the environment variables that are changed if settings management is disabled. If you can be specific about what you think is inappropriate, it can be fixed; a PR would be welcome, but please be specific about what you think is not correct and please link to the authority for your opinion.

[...]

PR #4925, opened 2023-05-18
[docs] Mention that `craftcms` is best with Craft CMS 4+

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/4650

docs: Folks on Craft CMS 3 haven't really benefited from the explicit project type `craftcms`. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4925"></a>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -4,12 +4,12 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
 
 === "Craft CMS"
 
-    ## Craft CMS (4)
+    ## Craft CMS
 
     Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing one.
 
     !!!tip "Compatibility with Craft CMS 3"
-        The `craftcms` project type is best with Craft CMS 4+. If you are using Craft CMS 3 you may prefer to use the `php` project type. Craft CMS 4+ (and thus DDEV's support) are more opinionated about some settings, and you may want to just manage your own settings. More information is available in [issue](https://github.com/ddev/ddev/issues/4650).
+        The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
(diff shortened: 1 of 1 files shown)


docs/content/users/quickstart.md
```diff
@@ -4,12 +4,12 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
 
-    ## Craft CMS
+    ## Craft CMS (4)
 
     Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing one.
 
-    !!!tip "Compatibility"
-        The `craft` project type was added to DDEV in version [1.21.2](https://github.com/ddev/ddev/releases/tag/v1.21.2). Check your current version with the `ddev version` command, and [upgrade](../users/usage/faq.md#how-can-i-updateupgrade-ddev) if necessary!
+    !!!tip "Compatibility with Craft CMS 3"
+        The `craftcms` project type is best with Craft CMS 4+. If you are using Craft CMS 3 you may prefer to use the `php` project type. Craft CMS 4+ (and thus DDEV's support) are more opinionated about some settings, and you may want to just manage your own settings. More information is available in [issue](https://github.com/ddev/ddev/issues/4650).
```
mattstein, line 12: Can whittle this down for efficiency, grammar, and link context. If I’m following correctly, it’d also be good to mention Craft 4, then Craft 3 and the recommended action. As written the tip mentions Craft 3, then the text mentions Craft 4, Craft 3, and Craft 4 again and it’s a bit of a jumble. I hope this is more straightforward and still what you intended.

```suggestion
        The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
```

docs/content/users/quickstart.md
```diff
@@ -4,12 +4,12 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
 
 === "Craft CMS"
 
-    ## Craft CMS
+    ## Craft CMS (4)
```
mattstein, line 7: I wouldn’t limit this to v4 here because you’re addressing v3 in this new note.

```suggestion
    ## Craft CMS
```

mattstein (changes requested):

PR #4929, opened 2023-05-18
Delete FUNDING.yml

gilbertsoft:
Use FUNDING.yml from the default community health file instead (https://github.com/ddev/.github/blob/main/FUNDING.yml)

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4929"></a>

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,0 @@
-github: [ddev]
-custom: ["https://www.paypal.com/donate/?hosted_button_id=MCNCSZHC7LHSQ", "Localdev Foundation via PayPal"]


rfay (changes requested):
Last I looked, the org one wasn't working. For example, see https://github.com/ddev/ddev-platformsh or https://github.com/ddev/ddev-addon-template - they don't have the funding info. So I don't think we can remove this.

I guess we should ask their support about it.

gilbertsoft:
> Last I looked, the org one wasn't working. For example, see https://github.com/ddev/ddev-platformsh or https://github.com/ddev/ddev-addon-template - they don't have the funding info. So I don't think we can remove this.
> 
> I guess we should ask their support about it.

Doesn't work there because sponsoring isn't enabled in these repos. Here it works fine https://github.com/ddev/localdev.foundation

gilbertsoft:
Activated for https://github.com/ddev/ddev-addon-template now and works too

gilbertsoft:
Btw fixed the second link before which wasn't a link and has broken custom links.

gilbertsoft:
Sposoring should now be active in all repos.

rfay (approved):
Thanks for sorting this out!

PR #4930, opened 2023-05-18
Only run wsl2-docker-desktop on master

rfay:
## The Issue

WSL2 Docker Desktop crashes so often that it doesn't make sense to run it for every PR/commit. Let's just run it against master. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4930"></a>


gilbertsoft (approved):
Looks good! Sad we have to disable it.

PR #4931, opened 2023-05-19
[docs] Give tip about ddev --version in ddev version section

rfay:
## The Issue

`ddev version` normally gives more output than people want. Give tip about `ddev --version`

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4931"></a>


mattstein (approved):
I’m inclined to think this is a case for flipping the functionality, where `ddev

PR #4932, opened 2023-05-19
Expand files before processing so we can check #ddev-generated, fixes #4253

### LINKED ISSUE #4253: ddev-get: When installing directory, only check #ddev-generated in files that exist in the add-on
The best way to explain this is this commit/project:

https://github.com/hanoii/ddev-z/commit/4dba742c56e00ced4629fdd0d083a6723d575828

I have this add-on where I specified `homeaddtions/` on `project_files`. It worked just fine until I had a project where I had other unrelated stuff in `homeaddtions/` and `ddev get hanoii/ddev-z` was **NOT** installing the file because there was another file on the directory:

```
NOT overwriting file/directory /Users/ariel/Sites/..../.ddev/homeadditions. 
The #ddev-generated signature was not found  in the file, so it will not be overwritten. 
You can just remove the file and use ddev get again if you 
want it to be replaced: signature was not found in file 
/Users/ariel/Sites/..../.ddev/homeadditions/.bashrc.d/path.sh 
```

On the above, `path.sh` is something specific to the ddev project and not present in the above. 

### Expected Behavior

I would expect that when one specifies a directory on `project_files`, only the files present on the whole tree are checked for the `#ddev-generated` signature and not any other potential files in there.

rfay:
## The Issue

* #4253 

* Expand directories found in both ProjectFiles and GlobalFiles before using them, so we process only files
* Improve testing to cover this

Use a sample add-on that has a directory in it; make sure #ddev-generated is respected in files in that directory.

## Automated Testing Overview

TestCmdGetComplex is updated to handle this situation.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4932"></a>

### DIFF
--- a/cmd/ddev/cmd/get.go
+++ b/cmd/ddev/cmd/get.go
@@ -287,7 +287,12 @@ ddev get --remove ddev-someaddonname
 		if len(s.ProjectFiles) > 0 {
 			util.Success("\nInstalling project-level components:")
 		}
-		for _, file := range s.ProjectFiles {
+
+		projectFiles, err := fileutil.ExpandFilesAndDirectories(extractedDir, s.ProjectFiles)
+		if err != nil {
+			util.Failed("Unable to expand files and directories: %v", err)
+		}
+		for _, file := range projectFiles {
 			file := os.ExpandEnv(file)
 			src := filepath.Join(extractedDir, file)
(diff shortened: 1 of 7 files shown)


mattstein (changes requested):
Looks good! Found one character to nitpick.

Issue #4922, opened 2023-05-17
Make documentation version more explicit on ddev.readthedocs.io


gitressa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

The DDEV documentation on https://ddev.readthedocs.io/ is great and very detailed. But for someone new to DDEV and readthedocs.io concept of "stable" and "latest", it is not clear which version you are looking at.

I quite like how the [Drush documentation](https://www.drush.org) is set up, since it's very obvious what you are getting, which is the current official release (Drush 11) prominently displayed next to the logo.

It is also very clear in the URL: https://www.drush.org/11.x/ and older or newer versions are available from the drop-down.

Perhaps something like this is possible, to make it clearer for the user which version they are looking at?

## Add version number in URL's

Show the version number, allowing comparing with current official release in the top right hand side.
- Now: https://ddev.readthedocs.io/en/stable/
- Possible? https://ddev.readthedocs.io/1.21.6

## Add version drop-down

Show version number next to logo, in the URL and right hand side.
![1-version-1 21 6-everywhere](https://github.com/ddev/ddev/assets/3491208/59c7e293-7d45-4903-8b66-4b2c8260cc5f)

Previous (1.19) and dev-version (1.22) is available from drop-down
![2-older-newer-versions-available](https://github.com/ddev/ddev/assets/3491208/020b8bfc-8cea-42df-826d-a72f36098b0c)

1.22 selected. It's clearer that this is the dev-version, since in the right hand side it says 1.21.6
![3-1 22](https://github.com/ddev/ddev/assets/3491208/0f600616-05d7-4dc4-8c13-631bc0d762ea)

### Describe your solution

Make it clearer for the user if the documentation they are looking at match their version of DDEV.

rfay:
It's possible to have multiple versions in readthedocs, but they get things all confused because google latches onto the wrong one and then gets people into the wrong place. 

More thoughts on this @mattstein ?

mattstein:
This is a great suggestion that’d be a lot more useful than that puny selector hiding in the bottom right page corner. It looks like it could be [a matter of mkdocs-material configuration](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/) but I’d need to explore further.

The [version warning](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#version-warning) doesn’t solve the SEO indexing problem, but it seems like a nice way to help.

[...]

PR #4936, opened 2023-05-20
[docs] Improve version selection UX

mattstein:
## The Issue

https://github.com/ddev/ddev/issues/4922

This attempts to enable mike [as described in the Material for MkDocs documentation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#versioning). It’d be amazing if this so easily and immediately improved things, but I have no idea how to test it locally or with a build preview since each is strictly limited to the branch version without a selector. But pushing straight into production seems cavalier.

If I knew what they were, this wouldn’t be a draft PR!

## Automated Testing Overview

n/a

## Related Issue Link(s)

https://github.com/ddev/ddev/issues/4922

## Release/Deployment Notes

This could ruin everything or make it better. Which is tricky.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4936"></a>

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -74,6 +74,8 @@ extra:
   generator: false
   analytics:
     provider: plausible
+  version:
+    provider: mike
 
 # Extensions
 markdown_extensions:


mattstein:
Nope, not that simple.

readthedocs.org does its own thing with versioning and injects that tiny selector into the bottom right corner of published docs, and to take advantage of this we’d need to move from readthedocs.org, take a bit more control over build+structure, and deploy with GitHub pages or something. That’s [sort of on the radar](https://github.com/ddev/ddev.com-front-end/issues/49) already, but something to evaluate carefully so we don’t screw up peoples’ experience with the documentation while trying to improve it.

PR #4945, opened 2023-05-22
[docs] Improve `ddev pull` flags description with `--environment`

rfay:
## The Issue

When `ddev pull --environment=` was added, it didn't get into docs

Add to docs

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4945"></a>

### DIFF
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -1023,7 +1023,7 @@ ddev pull platform --skip-files -y
 # Pull the localfile integration’s files *only* without confirming
 ddev pull localfile --skip-db -y
 
-# Pull from platform specifying PLATFORM_ENVIRONMENT and PLATFORM_CLI_TOKEN on command line
+# Pull from Platform.sh specifying the environment variables PLATFORM_ENVIRONMENT and PLATFORM_CLI_TOKEN on the command line
 ddev pull platform --environment=PLATFORM_ENVIRONMENT=main,PLATFORMSH_CLI_TOKEN=abcdef
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/commands.md
```diff
@@ -1021,6 +1022,9 @@ ddev pull platform --skip-files -y
 
 # Pull the localfile integration’s files *only* without confirming
 ddev pull localfile --skip-db -y
+
+# Pull from platform specifying PLATFORM_ENVIRONMENT and PLATFORM_CLI_TOKEN on command line
```
mattstein, line 1026: This seems like a critical time to clarify it’s pulling from Platform.sh, to align with previous examples and not confuse the otherwise more general term “platform.” Also take a stab at describing what we’re doing instead of listing environment variables verbatim. Wanted to say “inline” instead of “on the command line” but not convinced that’s a term anybody uses. (Attempting to distinguish that those environment variables are being defined with the pull command and not an a `.env` file or some other location.)

```suggestion
# Pull from Platform.sh specifying environment and CLI token variables on the command line
```

mattstein (changes requested):

docs/content/users/usage/commands.md
mattstein: That’s what I meant with my “inline” comment above; the nature of setting those in that call vs. in an environment variable or someplace else. I don’t know that “inline” conveys that, and “on the command line” may do a better job or still be unclear. I was attempting to describe the purpose of those environment variables instead of repeating them, but I agree it could be better to repeat for clarity.

That period you slipped in breaks form with the comments that precede it though!

mattstein (approved):

PR #4951, opened 2023-05-23
[tests only] Use macOS 13 for colima, update colima

rfay:
## The Issue

* Colima should be running on macOS 13
* Colima needs upgrade to 0.5.5 and cache clean

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4951"></a>


rfay:
This was only about Colima testing, so is ready to go.

Issue #4963, opened 2023-06-05
Promote ddev config --auto, which simplifies all the way


gitressa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

I really like how minimal the DDEV start project command `ddev config` is. When I started using DDEV I found the [excellent documentation](https://ddev.readthedocs.io/en/latest/users/usage/commands/#config), where the example almost cannot be any simpler:

```
ddev config --docroot=web --project-type=drupal8
```

### Describe your solution

The final improvements would be:

- Allow omitting `docroot`: Check if a folder named  _web_ is present and if yes, assume `--docroot=web`
- Allow omitting `project-type`: Scan _composer.json_ file if present, and if found with Drupal content assume `--project-type=drupal8`

I was extremely impressed, when I realized that this is already possible with the `--auto` parameter:

```
$ ddev config --auto
Creating a new ddev project config in the current directory (/home/user/test) 
Once completed, your configuration will be written to /home/user/test/.ddev/config.yaml
 
Configuring a drupal10 codebase with docroot 'web' at /home/user/test/web 
Configuration complete. You may now run 'ddev start'. 
```

### Remaining tasks

Update the documentation.

rfay:
Thanks, but I think people just get too confused when things aren't already in place. If you're configuring a project where the docroot doesn't yet exist, everything goes wrong. Closing for now, appreciate the suggestion, happy to continue the conversation here or elsewhere.

PR #4964, opened 2023-06-05
Add --auto under ddev config examples

gitressa:
Solves #4963.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4964"></a>

### DIFF
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -158,7 +158,7 @@ ddev config
 # Configure a Drupal 8 project with a `web` document root
 ddev config --docroot=web --project-type=drupal8
 
-# Automatically detect docroot and project-type without prompting
+# Automatically detect `docroot` and `project-type` without prompting
 ddev config --auto
 
 # Switch the project’s default `nginx-fpm` to `apache-fpm`
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -158,6 +158,9 @@ ddev config
 # Configure a Drupal 8 project with a `web` document root
 ddev config --docroot=web --project-type=drupal8
 
+# Automatically detect docroot and project-type without prompting
+ddev config --auto
+
 # Switch the project’s default `nginx-fpm` to `apache-fpm`
 ddev config --webserver-type=apache-fpm
 ```


mattstein (changes requested):

docs/content/users/usage/commands.md
```diff
@@ -158,6 +158,9 @@ ddev config
 # Configure a Drupal 8 project with a `web` document root
 ddev config --docroot=web --project-type=drupal8
 
+# Automatically detect docroot and project-type without prompting
```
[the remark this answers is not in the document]
gitressa: Good suggestion, I committed your suggestion. (pretty cool feature, just a click to commit)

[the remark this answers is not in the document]
mattstein: I like that feature too! Learned about that from @rfay. 🙌

docs/content/users/usage/commands.md
```diff
@@ -158,6 +158,9 @@ ddev config
 # Configure a Drupal 8 project with a `web` document root
 ddev config --docroot=web --project-type=drupal8
 
+# Automatically detect `docroot` and `project-type` without prompting
```
rfay, line 161:
```suggestion
# Automatically detect `docroot` and `project-type` without prompting
# This assumes that the code is already in place in the directory and auto-detection is possible
```

rfay (changes requested):

docs/content/users/usage/commands.md
mattstein: @rfay I think it might make more sense to point this out [where the documentation explains what the flag does](https://github.com/ddev/ddev/blob/c51a90500868f487dc21b61726599646d729d368/docs/content/users/usage/commands.md?plain=1#L172), instead of using comments on the example. If it’s important to say this here, maybe tacking on `(if possible)` would suffice?

mattstein: Or “Attempt to automatically detect (...)”

rfay: That would be OK. I'm reluctant with --auto because if there's no code people get bogus settings that then haunt them. 

Little-known fact: `--auto` doesn't do anything. It just prevents the interactive questions about docroot and type..  So `ddev config --php-version=8.2` is also auto, but it sort-of does something.

mattstein: Sounds like `--sort-of-auto` defaults to `true`. :)

Tricky when the arguments kind of imply 🤫 and DDEV probably should say “hey your docroot setting may be weird because you’ve given me nothing to work with.” I’d personally prefer a warning or refusal on DDEV’s part but I can see how that’d be an unwanted conflict for somebody else. (Or maybe even DDEV internally in some case?)

rfay: `--auto` was only added after years of recommending some random arg like `--php-version`. People REALLY WANTED IT :) But the thing is those new users can get messed up.

rfay:
It's not that `--auto` isn't adequately robust, it's that people often use `ddev config` on fresh empty repos, as in many of the examples in https://ddev.readthedocs.io/en/latest/users/quickstart/

Back when everybody started with a git checkout, `ddev config --auto` would almost always work fine. 

The thing is once the .ddev/config.yaml is created, ddev respects it. So *then* `ddev config --auto` doesn't do what you might want. So if people `ddev config --auto` and empty directory and get a project type of `php` then it will stay PHP until they override that.

Maybe we need a `ddev config --auto --fire-bazooka` that is willing to override existing .ddev/config.yaml.

rfay:
I use `ddev config --auto` quite a lot with an empty directory :)

gitressa:
Sure, but that's because you know what result will be. You created the behavior after all :)

If you hadn't explained that `--auto` really is `--php-version=8.2`, I wouldn't have guessed it, and expected nothing to happen, running it in an empty folder.

rfay:
It's really not `--php-version=8.2`, it's really "Any argument will do". So any flag gets you the automated behavior. A better example might be something that you set to the default, maybe `ddev config --no-project-mount=false` or `ddev config --omit-containers ""`

tyler36:
> This is safer, and ensures only DDEV's PHP version is in play:

Yes. As my project list grew, and my support PHP versions increased, it is worth getting into the habit of using:

```
ddev config --docroot web --project-name my-site --project-type drupal10 --create-docroot
````

rfay:
@gitressa if you use php or composer on your host machine it will be nothing but pain, lots of problems to sort out. And it's not just PHP version conflicts, but extension differences, etc.

gitressa:
Yes @rfay, that's also what I tried to convey. Sorry if I wasn't clear enough in my formulation.

PR #4971, opened 2023-06-12
[docs] Show details about GitHub PAT for package management

rfay:
## The Issue

I couldn't figure out where the (expired) package management PAT was used. But it was already documented.

Give more context about it.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4971"></a>


mattstein:
Is [this](https://ddev--4971.org.readthedocs.build/en/4971/developers/release-management/#creating-a-release) the formatting you intended, @rfay?

Screen Shot 2023-06-13 at 04 02 43 PM@2x

I’m usually not a fan of “see X” style links that could use more contextual anchor text, but escaping the link itself seems downright reader-hostile since they’ll have to copy and paste it.

rfay:
I think it actually is the right formatting. Hard to work with though. The URL to the PAT can't pass docs check because it's not a public URL, that's why it's back-quoted. The name of the token is "Package management token see ... url"

So very awkward, but probably what I was trying to accomplish.

mattstein (approved):

PR #4983, opened 2023-06-14
Set traefik as default router, fixes #4679

### LINKED ISSUE #4679: Traefik router should become default router in v1.22
### Expected Behavior

In DDEV v1.22, traefik should become the default router.

- [ ] Make it the default
- [ ] Move it out of the experimental section of ddev docs
- [ ] Explain how to use the traditional router
- [ ] Perhaps explain about casual hosting not supported directly by traefik. It's already explained, but will likely have to be reworded.
- [ ] We probably have to change (and migrate) the configuration of `use_traefik: true` and instead use something like `router_type: traefik` or `router_type: traditional`

rfay:
## The Issue

* #4679 

* Traefik is default
* New global config `router: traefik` or `router: nginx-proxy`
* New `ddev config global --router=traefik`

## TODO

- [x] Re-enable and merge TestCmdConfigGlobal - it's pending changes in https://github.com/ddev/ddev/pull/4640
- [x] Add test coverage for `router` into TestCmdConfigGlobal
- [x] `ddev describe` currently shows something other than "traditional" or whatever. Use the correct value.
- [x] Implement @mattstein 's new diagram
- [x] Implement @gilbertsoft suggestion of typed strings.
- [x] Add IsTraefik() instead of awkward `if globalconfig.DdevGlobalConfig.Router == nodeps.TraefikRouter && status == SiteRunning`
- [x] Use constants in initializations in globalconfig.New()

- [x] Try switching between traefik and traditional, verify behavior in each.
- [x] Make sure that existing `use-traefix` configuration in global_config.yaml doesn't break anything.

## Automated Testing Overview

* Updated tests

## Release/Deployment Notes

* Info about this change goes into release notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4983"></a>


pkg/globalconfig/global_config.go
```diff
@@ -60,13 +60,20 @@ type GlobalConfig struct {
 	ProjectTldGlobal             string                  `yaml:"project_tld"`
 	XdebugIDELocation            string                  `yaml:"xdebug_ide_location"`
 	NoBindMounts                 bool                    `yaml:"no_bind_mounts"`
-	UseTraefik                   bool                    `yaml:"use_traefik"`
+	Router                       string                  `yaml:"router"`
```
gilbertsoft, line 63: For a config expecting a group of predefined strings I'd use a typed string like I'm doing it in my change see https://github.com/ddev/ddev/pull/4904/files#diff-42da3ac5f97f8c2d00f51c6afe831dadaf9245cc93f258cee0e6f1a6ac2e7a06R38-R46

pkg/globalconfig/global_config.go
```diff
@@ -60,13 +60,20 @@ type GlobalConfig struct {
...
 	WSL2NoWindowsHostsMgt        bool                    `yaml:"wsl2_no_windows_hosts_mgt"`
 	RouterHTTPPort               string                  `yaml:"router_http_port"`
 	RouterHTTPSPort              string                  `yaml:"router_https_port"`
 	ProjectList                  map[string]*ProjectInfo `yaml:"project_info"`
 }
 
+// NewGlobalConfig() returns a minimally initialized GlobalConfig
+func NewGlobalConfig() GlobalConfig {
```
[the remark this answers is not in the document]
rfay: I saw it more than one way. Plain New() doesn't have a good initialization technique AFAIK. 

In the past, no initialization has been required for this struct.

pkg/globalconfig/global_config.go
rfay: Awesome! Never even heard of typed string before.  I see https://golangbyexample.com/string-constant-golang/#Typed_string_constant and other references; not good google hits though.

rfay: AFAICT there's not a concept of "typed string", but rather a typed constant. But using it does not seem to constrain their use, it just gives the constant a type. I'm not at all sure that your example is constraining the uses. I do agree. that we could use a constant named Router, which is done for the *values* that can be assigned here, using https://github.com/ddev/ddev/pull/4983/files#diff-0caf521c3856fcd77e5a3e51d2c3cf9a925478fa8d1beb69aa654a8cff7fce7eR45-R48 and then the ValidRouterTypes function.

pkg/globalconfig/global_config.go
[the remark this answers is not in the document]
gilbertsoft: This was a misunderstanding, I'm talking about our function name to be kept as short as possible. We already talked about in PMs though.

pkg/globalconfig/global_config.go
gilbertsoft: I guess the compiler currently does not warn if a native string is used instead of the new type. This ofc would be great, well to have a warning in minimum. Also not sure about IDE support if they only show constants with the correct type and may also differ in various IDEs. So currently this is more like a cosmetic thing but it also helps to find usages of the new type. Maybe there will be more benefit in the future. Anyway it's a good practise.

pkg/nodeps/values.go
```diff
@@ -41,6 +41,12 @@ var ValidOmitContainers = map[string]bool{
 	DBAContainer:          true,
 }
 
+// Router types
+const (
+	TraditionalRouter = "traditional"
```
gilbertsoft, line 46: Can we name it by name? `nginx` or `nginx-proxy`. Traditional assumes some knowledge which we both have but not anyone else.

pkg/nodeps/values.go
```diff
@@ -59,8 +65,8 @@ const NodeJSDefault = "18"
 // NoBindMountsDefault is default value for globalconfig.DDEVGlobalConfig.NoBindMounts
 var NoBindMountsDefault = false
 
-// UseTraefikDefault is the default value for globalconfig.DDEVGlobalConfig.UseTraefik
-var UseTraefikDefault = false
+// UseTraditionalRouter is the default value for globalconfig.DDEVGlobalConfig.UseTraditionalRouter
+var UseTraditionalRouter = false
```
gilbertsoft, line 69: Should not be needed anymore, isn't it?

pkg/nodeps/values.go
```diff
@@ -41,6 +41,12 @@ var ValidOmitContainers = map[string]bool{
 	DBAContainer:          true,
 }
 
+// Router types
```
gilbertsoft, line 44: imho belongs to a dedicated types package

pkg/globalconfig/values.go
```diff
@@ -44,3 +50,9 @@ func IsValidXdebugIDELocation(loc string) bool {
 	}
 	return false
 }
+
+// IsValidRouterType limits the choices for XdebugIDELocation
```
gilbertsoft, line 54: imho belongs to a dedicated types package

pkg/globalconfig/values.go
```diff
@@ -23,6 +23,12 @@ var ValidOmitContainers = map[string]bool{
 	DBAContainer:          true,
 }
 
+// ValidRouterTypes is the list of valid router types
```
gilbertsoft, line 26: imho belongs to a dedicated types package

If you use a map for this (which will make you troubles if you try to create a list of valid types btw) you should use `map[string]struct{}` which has the lowest impact on Golang.

Please also have a look at my Mutagen patch how I have solved it there.

pkg/globalconfig/global_config.go
```diff
@@ -208,9 +224,9 @@ func ReadGlobalConfig() error {
 	if nodeps.NoBindMountsDefault == true {
 		DdevGlobalConfig.NoBindMounts = true
 	}
-	// For testing only, override UseTraefikDefault no matter what it's set to
-	if nodeps.UseTraefikDefault == true {
-		DdevGlobalConfig.UseTraefik = true
+	// For testing only, override UseTraditionalRouter no matter what it's set to
+	if nodeps.UseTraditionalRouter == true {
```
gilbertsoft, line 228:
```suggestion
	if nodeps.UseTraditionalRouter {
```

Never compare a bool value. Linter should anyway comlain here or we miss a config.

pkg/globalconfig/global_config.go
```diff
@@ -60,13 +60,20 @@ type GlobalConfig struct {
...
+	Router                       string                  `yaml:"router"`
 	WSL2NoWindowsHostsMgt        bool                    `yaml:"wsl2_no_windows_hosts_mgt"`
 	RouterHTTPPort               string                  `yaml:"router_http_port"`
 	RouterHTTPSPort              string                  `yaml:"router_https_port"`
 	ProjectList                  map[string]*ProjectInfo `yaml:"project_info"`
 }
 
+// NewGlobalConfig() returns a minimally initialized GlobalConfig
+func NewGlobalConfig() GlobalConfig {
+	return GlobalConfig{
```
gilbertsoft, line 72: Doesn't respect init() func. imho should be better done in a dedicated patch, does not belong to this patch.

pkg/ddevapp/ddevapp.go
```diff
@@ -2336,7 +2336,7 @@ func (app *DdevApp) Stop(removeData bool, createSnapshot bool) error {
 		util.Warning("Unable to SyncAndterminateMutagenSession: %v", err)
 	}
 
-	if globalconfig.DdevGlobalConfig.UseTraefik && status == SiteRunning {
+	if globalconfig.DdevGlobalConfig.Router == nodeps.TraefikRouter && status == SiteRunning {
```
gilbertsoft, line 2339: I'd introduce a method IsTrafik() or similar to the global config. Same for the proect config. Makes it much easier than using this comparision again and again.

pkg/ddevapp/config.go
```diff
@@ -48,8 +48,8 @@ func init() {
 	if os.Getenv("DDEV_TEST_NO_BIND_MOUNTS") == "true" {
 		nodeps.NoBindMountsDefault = true
 	}
-	if os.Getenv("DDEV_TEST_USE_TRAEFIK") == "true" {
-		nodeps.UseTraefikDefault = true
+	if os.Getenv("DDEV_TEST_USE_TRADITIONAL_ROUTER") == "true" {
```
gilbertsoft, line 51: Why not simply using DDEV_TEST_ROUTER and provide the correct value like done internally anyway? This would be more flexible to also be able to override with the Traefik router type for tests if needed at some point.

gilbertsoft (changes requested):
Had a frist look and found some minor conceptual issues which we should agree how to handle. Maybe also something to discuss on Tuesday.

mattstein (changes requested):

mattstein:
Here’s an updated file for `/images/container-diagram.png`:

container-diagram

docs/content/users/configuration/config.md
```diff
@@ -119,7 +119,7 @@ Whether to disable HTTP/2 listen in `ddev-router`.
 | -- | -- | --
 | :octicons-globe-16: global | `false` | Can be `true` or `false`.
 
-When `true`, nginx will not listen for HTTP/2, but use HTTP/1.1 SSL. (Some browsers don’t work well with HTTP/2.)
+Only available with `router: traditional`, not available with `router: traefik`, which is the default. When `true`, nginx will not listen for HTTP/2, but use HTTP/1.1 SSL. (Some browsers don’t work well with HTTP/2.)
```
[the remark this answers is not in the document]
rfay: Adapted and added in another commit.

docs/content/users/configuration/config.md
```diff
@@ -505,15 +505,15 @@ Whether to enable Let’s Encrypt integration. (Works in conjunction with [`lets
 
 May also be set via `ddev config global --use-letsencrypt` or `ddev config global --use-letsencrypt=false`. When `true`, `letsencrypt_email` must also be set and the system must be available on the internet. Used with the [casual hosting](../topics/hosting.md) feature.
 
-## `use_traefik`
+## `router`
 
-Whether to enable the [traefik router](experimental.md#traefik-router).
+Whether to enable the (default) [traefik router](traefik-router.md#traefik-router) or the legacy "traditional" router.
```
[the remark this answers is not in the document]
rfay: Changed to `nginx-proxy` for the name of the "traditional" router.

docs/content/users/configuration/traefik-router.md
```diff
@@ -0,0 +1,37 @@
+# Traefik Router
```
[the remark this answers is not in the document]
rfay: Moved to "extend"

pkg/nodeps/values.go
rfay: Changed to nginx-proxy

pkg/ddevapp/ddevapp.go
rfay: Done

pkg/ddevapp/config.go
rfay: I decided not to change this one, as the test being triggered here is a one-off. All other tests will use the default traefik.

docs/content/users/configuration/config.md
```diff
@@ -400,6 +400,16 @@ If set to `v2.8.0`, for example, it will download and use that version instead o
 !!!warning "Troubleshooting Only!"
     This should only be used in specific cases like troubleshooting. Best avoided otherwise.
 
+## `router`
+
+Whether to enable the default [traefik router](../extend/traefik-router.md#traefik-router) or the legacy "nginx-proxy" router.
```
mattstein, line 405: If it’s not a reference to a literal value, Traefik is a proper noun:

```suggestion
Whether to enable the default [Traefik router](../extend/traefik-router.md#traefik-router) or the legacy "nginx-proxy" router.
```

docs/content/users/configuration/traefik-router.md
```diff
@@ -0,0 +1,37 @@
...
+
+!!!note
+Traefik is the default router in DDEV v1.22+.
+
+DDEV’s router plays an important role in its [container architecture](../usage/architecture.md#container-architecture), receiving most HTTP and HTTPS traffic for requests like `*.ddev.site` and delivering them to the relevant project’s web container.
+
+As of DDEV v1.22.0+ the [Traefik router](https://traefik.io/traefik/) is the default router. You will be using it by default unless you configure the traditional router with `ddev poweroff && ddev config global --router=traditional`.
```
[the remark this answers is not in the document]
rfay: Fixed this in other commits.

docs/content/users/configuration/traefik-router.md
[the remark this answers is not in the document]
rfay: Hope you like it.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
```diff
@@ -0,0 +1,31 @@
+package globalconfigtypes
+
+type RouterType = string
```
gilbertsoft, line 3:
```suggestion
type RouterType string
```

We like to have a new type so equal symbol is not needed. With the equal symbol no new type is created.

cmd/ddev/cmd/debug-router-nginx-config.go
```diff
@@ -47,7 +48,7 @@ var DebugRouterNginxConfigCmd = &cobra.Command{
 }
 
 func init() {
-	if !globalconfig.DdevGlobalConfig.UseTraefik {
+	if !globalconfig.IsTraefikRouter() {
```
gilbertsoft, line 51: That's dangerous, I'd create a second func IsNginxRouter instead.

pkg/ddevapp/amplitude_project.go
```diff
@@ -56,14 +56,14 @@ func (app *DdevApp) TrackProject() {
...
 		RouterDisabled(IsRouterDisabled(app)).
-		TraefikEnabled(globalconfig.DdevGlobalConfig.UseTraefik).
 		WebserverType(app.GetWebserverType())
 
 	if !nodeps.ArrayContainsString(containersOmitted, "db") {
 		builder.
 			DatabaseType(app.Database.Type).
 			DatabaseVersion(app.Database.Version)
 	}
+	builder.Router(globalconfig.DdevGlobalConfig.Router)
```
gilbertsoft, line 66:
```suggestion

	if !IsRouterDisabled(app) {
		builder.Router(globalconfig.DdevGlobalConfig.Router)
	}
```

For optional properties we should use a condition or make them required.

pkg/globalconfig/global_config.go
```diff
@@ -627,6 +637,11 @@ func IsInternetActive() bool {
 	return active
 }
 
+// IsTraefikRouter returns true if the router is traefik
+func IsTraefikRouter() bool {
+	return DdevGlobalConfig.Router == globalconfigtypes.RouterTypeTraefik
+}
```
gilbertsoft, line 643:
```suggestion
func (c *GlobalConfig) IsTraefikRouter() bool {
	return c.Router == globalconfigtypes.RouterTypeTraefik
}
```

Because you access config properties this should be a method instead.

gilbertsoft (changes requested):
Had a quick look a made some more suggestions.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// GetValidRouterTypes returns a list of valid router types
+func GetValidRouterTypes() []string {
+	s := make([]string, 0, len(ValidRouterTypes))
+	for p := range ValidRouterTypes {
+		s = append(s, p)
+	}
+	return s
+}
```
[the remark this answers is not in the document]
rfay: I don't care about the order. I just care what's valid. And it's only used in an error message.  Why should we care about the ordering here?

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
```diff
@@ -0,0 +1,31 @@
+package globalconfigtypes
```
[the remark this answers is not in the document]
rfay: Hmm, started this as `types` and had errors.

cmd/ddev/cmd/config-global_test.go
```diff
@@ -54,39 +51,71 @@ func TestCmdGlobalConfig(t *testing.T) {
...
-	require.Contains(t, string(out), "wsl2-no-windows-hosts-mgt=false\nrouter-http-port=80\nrouter-https-port=443")
+	assert.NoError(err, "error running ddev config global; output=%s", out)
+	assert.Contains(out, "instrumentation-opt-in=false\nomit-containers=[]")
+	assert.Contains(out, `web-environment=[]`)
+	assert.Contains(out, "mutagen-enabled=false\nnfs-mount-enabled=false")
+	assert.Contains(out, "router-bind-all-interfaces=false")
+	assert.Contains(out, "internet-detection-timeout-ms=3000")
+	assert.Contains(out, "disable-http2=false")
+	assert.Contains(out, "use-letsencrypt=false\nletsencrypt-email=\n")
+	assert.Contains(out, "table-style=default\nsimple-formatting=false")
```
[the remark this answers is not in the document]
rfay: I was on the path, will get there. This does make it much easier to update this test though. We'll have to return to this on all the pending ones we have, like mutagen etc.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
rfay: We decided not to do this.

cmd/ddev/cmd/debug-router-nginx-config.go
rfay: Changed to direct comparison; this is a very obscure location though, in an obsolete usage, not worth too much worry.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Router Types
+const (
+	RouterTypeTraefik    RouterType = "traefik"
+	RouterTypeNginxProxy RouterType = "nginx-proxy"
+	RouterTypeDefault    RouterType = RouterTypeTraefik
+)
+
+// ValidRouterTypes is the list of valid router types
+var ValidRouterTypes = map[RouterType]bool{
```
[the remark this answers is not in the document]
rfay: Changed in pushed code.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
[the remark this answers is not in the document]
rfay: Changed in pushed code.

cmd/ddev/cmd/config-global.go
```diff
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/ddev/ddev/pkg/globalconfig"
+	"github.com/ddev/ddev/pkg/globalconfig/globalconfigtypes"
```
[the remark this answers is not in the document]
rfay: Changed in pushed

cmd/ddev/cmd/config-global.go
```diff
@@ -262,7 +263,7 @@ func init() {
 	configGlobalCommand.Flags().Bool("use-docker-compose-from-path", true, "If true, use docker-compose from path instead of private ~/.ddev/bin/docker-compose")
 	configGlobalCommand.Flags().Bool("no-bind-mounts", true, "If true, don't use bind-mounts - useful for environments like remote docker where bind-mounts are impossible")
 	configGlobalCommand.Flags().String("xdebug-ide-location", "", "For less usual IDE locations specify where the IDE is running for Xdebug to reach it")
-	configGlobalCommand.Flags().Bool("use-traefik", true, "If true, use traefik for ddev-router")
+	configGlobalCommand.Flags().String("router", globalconfigtypes.RouterTypeTraefik, "Use traefik router (default) or `nginx-proxy` legacy router")
```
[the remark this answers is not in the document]
rfay: Used the existing method to get the list. Good point.

pkg/ddevapp/amplitude_project.go
rfay: Changed in pushed code, thanks.

pkg/globalconfig/globalconfigtypes/globalconfig_types.go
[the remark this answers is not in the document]
gilbertsoft: I have to care because I use it for the command to show the user the valid options.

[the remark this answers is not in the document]
rfay: The question is why you care about the order of the listed options. And what is the correct order? Probably not alphabetic.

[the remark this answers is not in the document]
gilbertsoft: In this case here I would prefer to have Traefik always as first option because this is what people should use in the first place.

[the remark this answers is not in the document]
rfay: OK, changed it to slice in https://github.com/ddev/ddev/pull/4983/commits/63dcec63f9561b4ec186857c350676bdc243cb4b

pkg/globalconfig/global_config.go
```diff
@@ -664,7 +679,7 @@ func GetRequiredDockerComposeVersion() string {
 func GetRouterURL() string {
 	routerURL := ""
 	// Until we figure out how to configure this, use static value
-	if DdevGlobalConfig.UseTraefik {
+	if DdevGlobalConfig.Router == types.RouterTypeTraefik {
```
gilbertsoft:
```suggestion
	if DdevGlobalConfig.IsTraefikRouter() {
```

pkg/globalconfig/global_config.go
```diff
@@ -156,23 +157,28 @@ func ValidateGlobalConfig() error {
...
 	if !IsValidTableStyle(DdevGlobalConfig.TableStyle) {
 		DdevGlobalConfig.TableStyle = "default"
 	}
 
 	if !IsValidXdebugIDELocation(DdevGlobalConfig.XdebugIDELocation) {
 		return fmt.Errorf(`xdebug_ide_location must be IP address or one of %v`, ValidXdebugIDELocations)
 	}
-	if DdevGlobalConfig.DisableHTTP2 && DdevGlobalConfig.UseTraefik {
-		return fmt.Errorf("disable_http2 and use_traefik are mutually incompatible")
+	if DdevGlobalConfig.DisableHTTP2 && DdevGlobalConfig.Router == types.RouterTypeTraefik {
```
gilbertsoft:
```suggestion
	if DdevGlobalConfig.DisableHTTP2 && DdevGlobalConfig.IsTraefikRouter() {
```

gilbertsoft (approved):
Looks good!

rfay:
Thanks, I'll wait until upstream/master tests look mostly done.

PR #4987, opened 2023-06-15
[tests only] Make TestHostDBPort more stable

gilbertsoft:
There maybe additional debug output which currently will break the test. This change makes the test more stable by only verifying the expected string is printed somewhere in the output and does not longer expect to be found at a predefined position.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4987"></a>


rfay:
Thanks, would have been OK to fix it in your PR. I often fix bad tests when they show up due to new input.

rfay (approved):
Looks fine to me, thanks!

PR #4988, opened 2023-06-15
[docs] Fix link to TYPO3 docs

gilbertsoft:


docs/content/users/usage/cms-settings.md
```diff
@@ -125,4 +125,4 @@ baseVariants:
     condition: 'applicationContext == "Development/DDEV"'
 ```
 
-See also [TYPO3 Documentation](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SiteHandling/BaseVariants.html).
+See also [TYPO3 Documentation](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/SiteHandling/BaseVariants.html).
```
rfay, line 128: Is there a better way to make this permanent so it will always track properly? Seems like 12.4 will be very fragile very fast.

rfay (approved):
OK with me then. Please do check with team about whether they'll manage a main branch etc.

rfay:
I saw it was breaking existing tests (why you did it I assume) so pulled it.

docs/content/users/usage/cms-settings.md
gilbertsoft: Rendering of main branch didn't work but up again. We can change again to main.

PR #4994, opened 2023-06-16
Update error message in ddevapp.go

rpkoller:
A small wording change to the "ddev ssh failed" error message. I only wonder in regards of styling if "styling" should be upper case. Usually it is the rule to use upper case after periods and colons?

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4994"></a>


pkg/ddevapp/ddevapp.go
```diff
@@ -1639,7 +1639,7 @@ func (app *DdevApp) ExecWithTty(opts *ExecOpts) error {
 
 	state, err := dockerutil.GetContainerStateByName(fmt.Sprintf("ddev-%s-%s", app.Name, opts.Service))
 	if err != nil || state != "running" {
-		return fmt.Errorf("service %s is not current running in project %s (state=%s)", opts.Service, app.Name, state)
+		return fmt.Errorf("service %s is currently not running in project %s (state=%s)", opts.Service, app.Name, state)
```
rfay, line 1642: I think this might be a little simpler.

```suggestion
		return fmt.Errorf("service %s is not running in project %s (state=%s)", opts.Service, app.Name, state)
```

rfay (changes requested):
Thanks! Minor suggestion.

pkg/ddevapp/ddevapp.go
rpkoller: Ahhhh I thought currently should kept in. But removing it is short and even more clear. Good point I'll change that. But one question i've added in the initial comment. Should "service" be upper case since it is after a colon? Stylistically it is recommended but i suppose out of the scope of this issue. Cuz i think it would require to update a few more strings. Maybe something for a followup?

rfay: I think it's probably fine like this. Thanks!

rfay (approved):
Thanks!

PR #4995, opened 2023-06-17
Fix GetRouterHTTP*Port() to use defaults if empty

rfay:
## The Issue

We had [consistent failures](https://buildkite.com/ddev/ddev-windows-mutagen/builds/4064#0188c267-bc9c-4cbb-87a0-a0016166daa7) of TestPretestAndEnv on Windows/Mutagen (traditional Windows)

It turns out that this situation is where the global config has empty values for router_http_port. I'm not sure why it gets that on Windows, but a default should be set anyway.

This was introduced in 
* https://github.com/ddev/ddev/pull/4640

Set the default in GetRouterHTTPPort() etc.

## TODO:

The actual default global setting should be being set in constructor/New() when we're building the DdevGlobalConfig. That is being worked on for other reasons in https://github.com/ddev/ddev/pull/4983 - so we should set those there and then we can remove the default-setting behavior here.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -452,6 +452,11 @@ func (app *DdevApp) GetWebserverType() string {
 // GetRouterHTTPPort returns app's router http port
 func (app *DdevApp) GetRouterHTTPPort() string {
 	port := app.RouterHTTPPortGlobal
+	// TODO: This default setting should be done in creation of the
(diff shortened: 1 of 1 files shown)


gilbertsoft (approved):
Looks fine for the moment. The only reason I've found for having an empty global config is when the global_config.yaml can not be properly parsed or is not loaded at all. Root cause may be the loading of the global config via init() func which we should get rid of a.s.a.p.

PR #4998, opened 2023-06-17
[docs] Some consistency changes to env variables in custom-commands.md

rpkoller:
While adding DDEV_UID and DDEV_GID in a PR i've noticed that the env variable descriptions are sort of inconsistent. For one sometimes the first word after the colon is in lower case and sometimes in upper case. I thought after colon it should always be upper case? the only exception i've made for the description of DDEV_PROJECT_TYPE that just lists the cms versions which are names - same for DDEV_WEBSERVER_TYPE. And the other detail i wonder but wasn't sure in the end. would it make sense to end each description with a period cuz some already are? cuz the descriptions aren't always full sentences?

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4998"></a>


gilbertsoft (approved):
Thanks! Please rebase your branch, looks like you did not use the latest master.

rfay (approved):
Thanks!

rpkoller:
resolved the conflict in this PR as well. the only question that is still open if there should be periods at the end of each description or if it should be left as is and just the change to upper case after periods should be done?

gilbertsoft:
> resolved the conflict in this PR as well. the only question that is still open if there should be periods at the end of each description or if it should be left as is and just the change to upper case after periods should be done?

Thanks! We are able to merge now, conflicts resolved :)

Imho no periods needed here but maybe a native English has another opinion.... @mattstein ?

rpkoller:
oh I've just noticed, on my initial PR on some descriptions there were periods at the end and on some there were not. after the rebase there aren't any periods anymore only for the two descriptions with etc at the end, but there the period is for the abbreviation etc. and since as already mentioned in the issue summary the descriptions aren't always full sentences i would strike my suggestion about the periods.

mattstein (approved):

mattstein:
> Imho no periods needed here but maybe a native English has another opinion.... @mattstein ?

I agree. These aren’t entire sentences so ending punctuation isn’t necessary.

PR #5006, opened 2023-06-21
build: bump docker-compose to 2.20.0

rfay:
## The Issue

Docker-compose 2.20.0 is out

## How This PR Solves The Issue

Run tests to try it.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5006"></a>

PR #5007, opened 2023-06-21
Implement DdevGlobalConfig initialization in one place

rfay:
## The Issue

@gilbertsoft pointed out elsewhere that globalconfig.DdevGlobalConfig initialization should be encapsulated in a New() function. 

Do it and work it all out. 

This one ends up being a pre-requisite for some other PRs, especially 
* https://github.com/ddev/ddev/pull/4983
* https://github.com/ddev/ddev/pull/5006

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5007"></a>

### DIFF
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -26,10 +26,6 @@ var (
 	DdevGlobalConfig GlobalConfig
 )
 
-func init() {
-	DdevGlobalConfig.ProjectList = make(map[string]*ProjectInfo)
-}
-
 type ProjectInfo struct {
 	AppRoot       string   `yaml:"approot"`
 	UsedHostPorts []string `yaml:"used_host_ports,omitempty,flow"`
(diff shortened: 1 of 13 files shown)


pkg/globalconfig/global_config.go
```diff
@@ -229,7 +221,13 @@ func WriteGlobalConfig(config GlobalConfig) error {
 	if err != nil {
 		return err
 	}
-	cfgbytes, err := yaml.Marshal(config)
+
+	cfgCopy := config
+	// Remove some items that are defaults
+	if cfgCopy.RequiredDockerComposeVersion == RequiredDockerComposeVersionDefault {
+		cfgCopy.RequiredDockerComposeVersion = ""
+	}
```
gilbertsoft, line 229: I'd avoid this completely. Instead use a getter function which returns a sane default if the config value is empty. Similar you did with the http ports above.

gilbertsoft: Oh I see the function bellow which is used in the code. There you also could check for a value in the global config or use the default if empty.

gilbertsoft (changes requested):
Looks very good! Some minor comments.

pkg/globalconfig/global_config.go
gilbertsoft: Instead of this code, better fix it on loading.

pkg/globalconfig/global_config.go
```diff
@@ -70,6 +66,34 @@ type GlobalConfig struct {
...
+
+	cfg := GlobalConfig{
+		RequiredDockerComposeVersion: RequiredDockerComposeVersionDefault,
+		InternetDetectionTimeout:     nodeps.InternetDetectionTimeoutDefault,
+		TableStyle:                   "default",
+		RouterHTTPPort:               "80",
+		RouterHTTPSPort:              "443",
+		LastStartedVersion:           "v0.0",
+		NoBindMounts:                 nodeps.NoBindMountsDefault,
+		UseTraefik:                   nodeps.UseTraefikDefault,
```
rfay, line 80: I see that these should be constants, I don't want to do it right now but will try to fix that up in the traefik PR.

pkg/globalconfig/global_config.go
gilbertsoft: Yes, that's clear, the default value should never be written to the file. But still I think validation and removal should better be done after reading and a getter function should take care to return the default value. Anyway, that's not a stopper, go ahead.

PR #5013, opened 2023-06-23
Use `--file` for import-db and rework import-db and export-db, fixes #4593

### LINKED ISSUE #4593: Add "file" as alias for import-db flag "src"

gilbertsoft:
## The Issue

- #4593

This PR streamlines the flags and descriptions of the import-db and export-db commands. It introduces a new package which makes it possible to write cleaner command definitions regarding indentation. Commands are created with a factory function.

Old flags are still here and marked as deprecated to avoid BC issues.

Thanks to @miromichalicka for the initial work done in #4647.

The impact is minimal and everything should be covered by automated testing.

- import-db command works including deprecated flags
- export-db command works including deprecated flags

## Related Issue Link(s)

Closes:
* #4593

Replaces:
* #4647

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5013"></a>

### DIFF
--- a/cmd/ddev/cmd/import-db.go
+++ b/cmd/ddev/cmd/import-db.go
@@ -1,76 +1,155 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/dockerutil"
+	"github.com/ddev/ddev/pkg/heredoc"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
 )
 
-var dbSource string
-var dbExtPath string
-var targetDB string
-var noDrop bool
-var progressOption bool
+// NewImportDBCmd initializes and returns the `ddev import-db` command.
+func NewImportDBCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import-db [project]",
+		Args:  cobra.RangeArgs(0, 1),
+		Short: "Import a SQL dump file into the project",
+		Long: heredoc.Doc(`
+			Import a SQL dump file into the project.
+			
+			The database dump file can be provided as a SQL dump in a .sql, .sql.gz,
+			sql.bz2, sql.xz, .mysql, .mysql.gz, .zip, .tgz, or .tar.gz format.
+			
+			For the zip and tar formats, the path to a .sql file within the archive
+			can be provided if it is not located at the top level of the archive.
+			
+			An optional target database can also be provided; the default is the
+			default database named "db".
 
-// ImportDBCmd represents the `ddev import-db` command.
-var ImportDBCmd = &cobra.Command{
-	Use:   "import-db [project]",
-	Args:  cobra.RangeArgs(0, 1),
-	Short: "Import a sql file into the project.",
-	Long: `Import a sql file into the project.
-The database dump file can be provided as a SQL dump in a .sql, .sql.gz, sql.bz2, sql.xz, .mysql, .mysql.gz, .zip, .tgz, or .tar.gz
-format. For the zip and tar formats, the path to a .sql file within the archive
-can be provided if it is not located at the top level of the archive. An optional target database
-can also be provided; the default is the default database named "db".
-Also note the related "ddev mysql" command`,
-	Example: `ddev import-db
-ddev import-db --src=.tarballs/junk.sql
-ddev import-db --src=.tarballs/junk.sql.gz
-ddev import-db --target-db=newdb --src=.tarballs/db.sql.gz
-ddev import-db --src=.tarballs/db.sql.bz2
-ddev import-db --src=.tarballs/db.sql.xz
-ddev import-db <db.sql
-ddev import-db someproject <db.sql
-gzip -dc db.sql.gz | ddev import-db`,
-
-	PreRun: func(cmd *cobra.Command, args []string) {
-		dockerutil.EnsureDdevNetwork()
-	},
-	Run: func(cmd *cobra.Command, args []string) {
-		projects, err := getRequestedProjects(args, false)
-		if err != nil {
-			util.Failed("Unable to get project(s): %v", err)
-		}
-
-		app := projects[0]
-		status, _ := app.SiteStatus()
-
-		if status != ddevapp.SiteRunning {
-			err = app.Start()
+			Also note the related "ddev mysql" command.
+		`),
+		Example: heredoc.DocI2S(`
+			$ ddev import-db
+			$ ddev import-db --file=.tarballs/junk.sql
+			$ ddev import-db --file=.tarballs/junk.sql.gz
+			$ ddev import-db --database=other_db --file=.tarballs/db.sql.gz
+			$ ddev import-db --file=.tarballs/db.sql.bz2
+			$ ddev import-db --file=.tarballs/db.sql.xz
+			$ ddev import-db < db.sql
+			$ ddev import-db some-project < db.sql
+			$ gzip -dc db.sql.gz | ddev import-db
+		`),
+		PreRun: func(cmd *cobra.Command, args []string) {
+			dockerutil.EnsureDdevNetwork()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			projects, err := getRequestedProjects(args, false)
 			if err != nil {
-				util.Failed("Failed to start app %s to import-db: %v", app.Name, err)
+				return fmt.Errorf("unable to get project: %v", err)
 			}
-		}
 
-		err = app.ImportDB(dbSource, dbExtPath, progressOption, noDrop, targetDB)
-		if err != nil {
-			util.Failed("Failed to import database %s for %s: %v", targetDB, app.GetName(), err)
-		}
-		util.Success("Successfully imported database '%s' for %v", targetDB, app.GetName())
-		if noDrop {
-			util.Success("Existing database '%s' was NOT dropped before importing", targetDB)
-		} else {
-			util.Success("Existing database '%s' was dropped before importing", targetDB)
-		}
-	},
+			app := projects[0]
+
+			dumpFile, err := cmd.Flags().GetString("file")
+			if err != nil {
+				return err
+			}
+
+			if !cmd.Flags().Lookup("file").Changed && cmd.Flags().Lookup("src").Changed {
+				dumpFile, err = cmd.Flags().GetString("src")
+				if err != nil {
+					return err
+				}
+			}
+
+			extractPath, err := cmd.Flags().GetString("extract-path")
+			if err != nil {
+				return err
+			}
+
+			database, err := cmd.Flags().GetString("database")
+			if err != nil {
+				return err
+			}
+
+			if !cmd.Flags().Lookup("database").Changed && cmd.Flags().Lookup("target-db").Changed {
+				database, err = cmd.Flags().GetString("target-db")
+				if err != nil {
+					return err
+				}
+			}
+
+			noDrop, err := cmd.Flags().GetBool("no-drop")
+			if err != nil {
+				return err
+			}
+
+			noProgress, err := cmd.Flags().GetBool("no-progress")
+			if err != nil {
+				return err
+			}
+
+			if !cmd.Flags().Lookup("no-progress").Changed && cmd.Flags().Lookup("progress").Changed {
+				progress, err := cmd.Flags().GetBool("progress")
+				if err != nil {
+					return err
+				}
+
+				noProgress = !progress
+			}
+
+			return importDBRun(app, dumpFile, extractPath, database, noDrop, noProgress)
+		},
+	}
+
+	cmd.Flags().StringP("file", "f", "", "Path to a SQL dump file, plain text or compressed (tar/tar.gz/tgz/zip)")
+	cmd.Flags().String("extract-path", "", "Path to extract within the archive")
+	cmd.Flags().StringP("database", "d", "db", "Target database to import into")
+	cmd.Flags().Bool("no-drop", false, "Do not drop the database before importing")
+	cmd.Flags().Bool("no-progress", false, "Do not output progress")
+
+	// Backward compatibility
+	cmd.Flags().String("src", "", "Path to a SQL dump file, plain text or compressed (tar/tar.gz/tgz/zip)")
+	_ = cmd.Flags().MarkDeprecated("src", "please use --file instead")
+
+	cmd.Flags().String("target-db", "db", "Target database to import into")
+	_ = cmd.Flags().MarkDeprecated("target-db", "please use --database instead")
+
+	cmd.Flags().BoolP("progress", "p", true, "Display a progress bar during import")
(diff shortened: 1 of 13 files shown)


rfay (changes requested):
I don't completely grok this yet, but please update the description here to 
- [ ] explain the new internal/text package (and why you call it "text", which is not very descriptive)
- [ ] explain what the heredoc package is and why it's used


Although this is apparently simple, it is definitely not easy to review.

Since this will apparently be a pattern for future updates, could you please remove all the `var` usages with pointers, and instead use flags, as is done in newer commands? start.go uses this technique, for example https://github.com/ddev/ddev/blob/bbb2caca961784aa68d0511848823464a009f14b/cmd/ddev/cmd/start.go#L36-L47

gilbertsoft:
> I don't completely grok this yet, but please update the description here to
> 
>     * [ ]  explain the new internal/text package (and why you call it "text", which is not very descriptive)

Highly inspired by GH cli. It's about text helpper functions so the name looks fine to me.

>     * [ ]  explain what the heredoc package is and why it's used

It's explained in the PR, it's about the heredoc syntax which allows to indent texts so we do not longer have texts starting on the first col but can be indented properly like the code.

> Although this is apparently simple, it is definitely not easy to review.

Yeah, the main thing which has changed is the indentation but the diff view is not very helpful here. The biggest change is the access to the flags like requested.

> Since this will apparently be a pattern for future updates, could you please remove all the `var` usages with pointers, and instead use flags, as is done in newer commands? start.go uses this technique, for example

Done.

docs/content/users/usage/commands.md
```diff
@@ -539,9 +539,9 @@ Dump a database to a file or to stdout.
 Flags:
 
 * `--bzip2`: Use bzip2 compression.
-* `--file`, `-f`: Provide the path to output the dump.
-* `--gzip`, `-z`: Use gzip compression. (default `true`)
-* `--target-db`, `-d`: If provided, target-db is alternate database to export. (default `"db"`)
+* `--database`, `-d`: Target database to export from (default "db")
+* `--file`, `-f`: Path to a SQL dump file to export to
+* `--gzip`: Use gzip compression (default true)
```
mattstein, line 544: I’d keep the backticks to represent literal values, unless there’s a good reason to drop them:

```suggestion
* `--database`, `-d`: Target database to export from (default `"db"`)
* `--file`, `-f`: Path to a SQL dump file to export to
* `--gzip`: Use gzip compression (default `true`)
```

Did you remove the `-z` alias on purpose?

docs/content/users/usage/commands.md
```diff
@@ -667,11 +667,11 @@ ddev hostname somesite.ddev.local 127.0.0.1
...
-* `--extract-path`: If provided asset is an archive, provide the path to extract within the archive.
-* `--no-drop`: Set if you do NOT want to drop the db before importing.
-* `--progress`, `-p`: Display a progress bar during import. (default `true`)
-* `--src`, `-f`: Provide the path to a SQL dump in `.sql`, `.tar`, `.tar.gz`, `.tgz`, `.bz2`, `.xx`, or `.zip` format.
-* `--target-db`, `-d`: If provided, target-db is alternate database to import into. (default `"db"`)
+* `--database`, `-d`: Target database to import into (default "db")
+* `--extract-path`: Path to extract within the archive
+* `--file`, `-f`: Path to a SQL dump file, plain text or compressed (tar/tar.gz/tgz/zip)
+* `--no-drop`: Do not drop the database before importing
+* `--no-progress`: Do not output progress
```
mattstein, line 674: For stylistic consistency, literal values and file extensions should be wrapped in backticks—so I’d keep those. The new “plain text or compressed (...)” also seems less specific than what it replaced.

```suggestion
* `--database`, `-d`: Target database to import into (default `"db"`)
* `--extract-path`: Path to extract within the archive
* `--file`, `-f`: Provide the path to a SQL dump in `.sql`, `.tar`, `.tar.gz`, `.tgz`, `.bz2`, `.xx`, or `.zip` format.
* `--no-drop`: Do not drop the database before importing
* `--no-progress`: Do not output progress
```

docs/content/users/usage/commands.md
```diff
@@ -680,19 +680,19 @@ Example:
...
+ddev import-db --file=.tarballs/db.sql
 
 # Import the compressed `.tarballs/db.sql.gz` dump to the project database
-ddev import-db --src=.tarballs/db.sql.gz
+ddev import-db --file=.tarballs/db.sql.gz
 
-# Import the compressed `.tarballs/db.sql.gz` dump to a `newdb` database
-ddev import-db --target-db=newdb --src=.tarballs/db.sql.gz
+# Import the compressed `.tarballs/db.sql.gz` dump to a `other_db` database
+ddev import-db --database=other_db --file=.tarballs/db.sql.gz
```
mattstein, line 689: Maybe `additional_db` if you didn’t feel like `newdb` was working? Seems like a good idea to try and reiterate that it’s referring to the non-primary project database.

```suggestion
# Import the compressed `.tarballs/db.sql.gz` dump to a `additional_db` database
ddev import-db --database=additional_db --file=.tarballs/db.sql.gz
```

docs/content/users/usage/commands.md
```diff
@@ -680,19 +680,19 @@ Example:
...
 
 # Import the `db.sql` dump to the project database
-ddev import-db <db.sql
+ddev import-db < db.sql
 
-# Import the `db.sql` dump to a `newdb` database
-ddev import-db newdb <db.sql
+# Import the `db.sql` dump to a `some-project` project
+ddev import-db some-project < db.sql
```
mattstein, line 695: If I’m following correctly, this is importing to the `db` database in a custom project. So I’d prefix with `my-`, a convention we’re generally using to indicate user-generated handles + details, and adjust the description to clarify:

```suggestion
# Import the `db.sql` dump to the `my-project` default database
ddev import-db my-project < db.sql
```

mattstein (changes requested):

rfay:
Just read the heredoc readme, I like that feature! Not sure why they named it "heredoc", which means something completely different in bash.

gilbertsoft:
> Just read the heredoc readme, I like that feature! Not sure why they named it "heredoc", which means something completely different in bash.

heredoc is a name in the PHP world, not sure if somewhere else too. See https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc

docs/content/users/usage/commands.md
gilbertsoft: Yes, the -z flag is deprecated now. Backticks were removed by accident, I just copied the new texts from the command line output and there of course are none. And thanks for your review!

cmd/ddev/cmd/import-db.go
```diff
@@ -1,76 +1,155 @@
...
+			$ ddev import-db --file=.tarballs/db.sql.bz2
+			$ ddev import-db --file=.tarballs/db.sql.xz
+			$ ddev import-db < db.sql
+			$ ddev import-db some-project < db.sql
+			$ gzip -dc db.sql.gz | ddev import-db
+		`),
+		PreRun: func(cmd *cobra.Command, args []string) {
+			dockerutil.EnsureDdevNetwork()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
```
rfay, line 47: Love the RunE!!!

rfay (changes requested):
This seems like a great improvement and good example of command setup, thanks!

* Please don't change import-db's --target-db. There was no request for this, and there's no reason to break people's existing scripts and such. 
* I can cope with the way you've implemented `--file` but the idea was to support it as an alternate rather than as a replacement for `--src`. It's OK though.
* Please credit @miromichalicka in the commit when this goes in.
* Would you mind changing the text package to something like "indent" instead? "text" is already the name of dozens of packages. 
* Interested why the package doesn't just go into pkg; that would be the natural place for it. Unless you have a strong argument I'd prefer for it to be there.

docs/content/users/usage/commands.md
rfay: How did `-z` get deprecated? :)

cmd/ddev/cmd/import-db.go
gilbertsoft: Me too ;)

docs/content/users/usage/commands.md
gilbertsoft: Well, I deprecate it with this PR see https://github.com/ddev/ddev/pull/5013/files#diff-a08f845a12c4812bdf8f9649f2e8548658439231e8ca9d39038ea03be3f5c70fR97. Imho -z is a very bad name the flag does not need a shorthand in the first place, the other compression flags don't have shorthands either. But if you don't agree I can let it stay like it is.

gilbertsoft:
> This seems like a great improvement and good example of command setup, thanks! I manually tested and everything seems to work fine.
> 
>     * Please don't change import-db's --target-db. There was no request for this, and there's no reason to break people's existing scripts and such.

Nothing will break, --target-db still exists. And I really like to have this streamlined with other tools e.g. mysqldump, that's why I've changed it in the first place. We should always try to use standards from other tools so it easier for people to use.

> 
>     * I can cope with the way you've implemented `--file` but the idea was to support it as an alternate rather than as a replacement for `--src`. It's OK though.

Same as above, src still exists. But we can not use source for import because there it's the target or destination but file we can use in all places.

> 
>     * Please credit @miromichalicka in the commit when this goes in.
> 
>     * Would you mind changing the text package to something like "indent" instead? "text" is already the name of dozens of packages.
> 
>     * Interested why the package doesn't just go into pkg; that would be the natural place for it. Unless you have a strong argument I'd prefer for it to be there.

Changed it to heredoc, best name I've found and also moved to pkg now.

rfay (changes requested):
No, `--target-db` does not work here, which is why the tests are failing.  Looks like it's just a bug that makes it not work though. Please manually test.

```
$ ddev import-db --src=.tarballs/db.sql.gz --target-db=junk
Flag --src has been deprecated, please use --file instead
Flag --target-db has been deprecated, please use --database instead
Error: failed to import database 'db' for d9: Unable to validate import asset junk: invalid asset: file not found
Usage:
  ddev import-db [project] [flags]

Examples:
  $ ddev import-db
  $ ddev import-db --file=.tarballs/junk.sql
  $ ddev import-db --file=.tarballs/junk.sql.gz
  $ ddev import-db --database=other_db --file=.tarballs/db.sql.gz
  $ ddev import-db --file=.tarballs/db.sql.bz2
  $ ddev import-db --file=.tarballs/db.sql.xz
  $ ddev import-db < db.sql
  $ ddev import-db some-project < db.sql
  $ gzip -dc db.sql.gz | ddev import-db


Flags:
  -d, --database string       Target database to import into (default "db")
      --extract-path string   Path to extract within the archive
  -f, --file string           Path to a SQL dump file, plain text or compressed (tar/tar.gz/tgz/zip)
  -h, --help                  help for import-db
      --no-drop               Do not drop the database before importing
      --no-progress           Do not output progress

Global Flags:
```

docs/content/users/usage/commands.md
rfay: Please discuss deprecations rather than just doing them. If there were a discussion in a tracked location, that would be something, but just deprecating because you think it should deprecate is different. 

Are you still supporting `-z` here? It's the default anyway, so maybe people don't use it. But if people have scripts that use it we'll break them, and since it's unlikely to be used anyway... why bother deprecating?

gilbertsoft:
> No, `--target-db` does not work here, which is why the tests are failing. Looks like it's just a bug that makes it not work though. Please manually test.
> 
> ```
> $ ddev import-db --src=.tarballs/db.sql.gz --target-db=junk
> Flag --src has been deprecated, please use --file instead
> Flag --target-db has been deprecated, please use --database instead
> Error: failed to import database 'db' for d9: Unable to validate import asset junk: invalid asset: file not found
> Usage:
>   ddev import-db [project] [flags]
> 
> Examples:
>   $ ddev import-db
>   $ ddev import-db --file=.tarballs/junk.sql
>   $ ddev import-db --file=.tarballs/junk.sql.gz
>   $ ddev import-db --database=other_db --file=.tarballs/db.sql.gz
>   $ ddev import-db --file=.tarballs/db.sql.bz2
>   $ ddev import-db --file=.tarballs/db.sql.xz
>   $ ddev import-db < db.sql
>   $ ddev import-db some-project < db.sql
>   $ gzip -dc db.sql.gz | ddev import-db
> 
> 
> Flags:
>   -d, --database string       Target database to import into (default "db")
>       --extract-path string   Path to extract within the archive
>   -f, --file string           Path to a SQL dump file, plain text or compressed (tar/tar.gz/tgz/zip)
>   -h, --help                  help for import-db
>       --no-drop               Do not drop the database before importing
>       --no-progress           Do not output progress
> 
> Global Flags:
>   -j, --json-output   If true, user-oriented output will be in JSON format.
> ```

Looks like a bug while reading the flags. Will have a look....

gilbertsoft:
Found the problem, copy&paste issue :(

docs/content/users/usage/commands.md
gilbertsoft: Sure, it still works. I did not remove but deprecate it because there is no value and it's unclear in meaning as explained above. If you don't agree, we can remove the deprecation as well.

I did not touch any test which makes sure nothing is breaking here.

rfay: It's so useless anyway since it's the default, so no need to change anything at this point. Let's just have agreement about deprecations before doing them. (I'm not sure there's a test that would use -z)

gilbertsoft: There's no risk -z not working anymore, just marked as deprecated and if --gzip works -z works too.

docs/content/users/usage/commands.md
gilbertsoft: Done

docs/content/users/usage/commands.md
gilbertsoft: Done

docs/content/users/usage/commands.md
gilbertsoft: Done, also fixed in the command

docs/content/users/usage/commands.md
gilbertsoft: Done

rfay (changes requested):
~~As far as I can tell, --target-db still doesn't work:~~


Edit: I messed up the command, it's working

rfay (approved):
Looks like it's ready to go, thanks! Manually tested as well.

PR #5016, opened 2023-06-24
docs: auto create commands documentation

gilbertsoft:
## The Issue

Commands must be maintained in two places, the source and afterwards the docs and docs therefor are not always up to date.

There is the cobra/docs package to auto generate the docs from the command definitions.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5016"></a>


gilbertsoft:
@mattstein a first draft of auto generating the docs for the commands. I did not do much so far because I like to discuss this topic with you. At the moment one file per command is created in contrast to one doc for all commands at the moment.

Please have a look and give me some feedback so I know how to continue on this PR...

Thanks!

rfay:
@mattstein is going to be very happy with you.

mattstein:
I’m a fan of this @gilbertsoft, nice work!

If the docs can be generated from code comments that seems like a win for everybody. In my experience, the efficiency comes with the tradeoff of making it more challenging to customize for the reader. But I’ve also never spent any real time with Cobra (or Golang 😅).

Questions:

1. Is there an option to keep everything to one page? I don’t much care what the URL structure looks like, but a benefit of the one long page right now is discovery; as a new user I can scroll through all the commands and see how they’re used. With a command per page, I worry that the experience could be more cumbersome and fragmented for someone browsing. (Barring any UI work to address that.)
2. What does support look like for documentation-specific thinks like anchors and `::: tip` formatting? The [`ddev blackfire` command](https://ddev--5016.org.readthedocs.build/en/5016/users/usage/commands/#blackfire) uses both, for example:  
Screen Shot 2023-06-26 at 12 05 29 PM@2x

gilbertsoft:
> I’m a fan of this @gilbertsoft, nice work!
> 
> If the docs can be generated from code comments that seems like a win for everybody. In my experience, the efficiency comes with the tradeoff of making it more challenging to customize for the reader. But I’ve also never spent any real time with Cobra (or Golang 😅).
> 
> Questions:
> 
>     1. Is there an option to keep everything to one page? I don’t much care what the URL structure looks like, but a benefit of the one long page right now is discovery; as a new user I can scroll through all the commands and see how they’re used. With a command per page, I worry that the experience could be more cumbersome and fragmented for someone browsing. (Barring any UI work to address that.)

We should be able to change it to one page, yes. At the moment I've only copied the implementation from the GH cli to have a first draft to discuss. 15 minutes of work so far.

> 
>     2. What does support look like for documentation-specific conventions like anchors and `::: tip` formatting? The [`ddev blackfire` command](https://ddev--5016.org.readthedocs.build/en/5016/users/usage/commands/#blackfire) uses both, for example:
> 
> 
> Screen Shot 2023-06-26 at 12 05 29 PM@2x

Guess this should also be possible but have to check the details how we can implement it.

>     3. Would you take a similar approach to automating the [Config Options](https://ddev--5016.org.readthedocs.build/en/5016/users/configuration/config/) reference? One of the few benefits of doing this all by hand right now is that a new feature tends to make it into both the command and config reference when that’s relevant.

Yes, I've already discussed this some weeks ago with Randy. But this needs a bit more preparation and we first have to cleanup and optimize the config stuff before. It's on the list though.

Thanks for your input. Will go ahead and change it to a one page as next step.

rfay:
This is a phenomenal initiative, but of course it's gone way stale and has inadequate information about what it does and where it's going. I sure hope we can pick it up again! And of course we always appreciate your massive contributions to DDEV. Closing for now.

PR #5019, opened 2023-06-25
[tests] Optimize caching of downloaded assets

gilbertsoft:
## The Issue

Download of assets for tests fails regularly and tests have to be restarted.

This patch avoids the cleanup of assets between the test runs and introduces a unique hash for every assets which are now properly shared between the test runs and should minize failures.


rfay (commented):
We don't have *that* much trouble with the caching of CMS tarballs, but improving it is fine. Only one test does the full set of these (Github Pull/TestDdevFullSiteSetup). 

The most common failure is the node test, which absolutely needs improvement and caching. 

The real approach to caching the CMS tarballs though is to cache them in the github test.

gilbertsoft:
> We don't have _that_ much trouble with the caching of CMS tarballs, but improving it is fine. Only one test does the full set of these (Github Pull/TestDdevFullSiteSetup).
> 
> The most common failure is the node test, which absolutely needs improvement and caching.
> 
> The real approach to caching the CMS tarballs though is to cache them in the github test.

About 90% of the failures I have is because of failed downloads!

gilbertsoft:
Happened again in this patch
grafik

rfay:
I see more of the nodejs stuff, but every improvement is welcome.

However, improving the caching in the go code won't help much here unless we actually cache the testcache in the github tests.

gilbertsoft:
> I see more of the nodejs stuff, but every improvement is welcome.
> 
> However, improving the caching in the go code won't help much here unless we actually cache the testcache in the github tests.

Will check for more improvments e.g. GH once the tests are green again...

.github/workflows/tests.yml
```diff
@@ -75,6 +75,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          key: ddev-test-cache
+          path: ~/.ddev/testcache
```
rfay, line 81: This looks great.

cmd/ddev/cmd/clean.go
```diff
@@ -72,7 +72,6 @@ Additional commands that can help clean up resources:
 		ddevapp.PowerOff()
 
 		globalDdevDir := globalconfig.GetGlobalDdevDir()
-		_ = os.RemoveAll(filepath.Join(globalDdevDir, "testcache"))
```
gilbertsoft, line 75: If the testcache needs a cleanup this should be done in the tests instead.

pkg/testcommon/testcommon.go
```diff
@@ -297,8 +296,13 @@ func ContainerCheck(checkName string, checkState string) (bool, error) {
...
 // Returns the extracted path, the tarball path (both possibly cached), and an error value.
-func GetCachedArchive(siteName string, prefixString string, internalExtractionPath string, sourceURL string) (string, string, error) {
-	uniqueName := prefixString + "_" + path.Base(sourceURL)
+func GetCachedArchive(siteName, prefixString, internalExtractionPath, sourceURL string) (string, string, error) {
+	var uniqueName string
+	if strings.Contains(sourceURL, ".tar.") {
+		uniqueName = fmt.Sprintf("%x.tar%s", sha256.Sum256([]byte(sourceURL)), path.Ext(sourceURL))
+	} else {
+		uniqueName = fmt.Sprintf("%x%s", sha256.Sum256([]byte(sourceURL)), path.Ext(sourceURL))
+	}
```
gilbertsoft, line 305: Create a unique name based on the source URL only so we have each asset only once in the cache.

.github/workflows/tests.yml
gilbertsoft: Almost. This does not work correctly since the cache never gets updated with this key. It's important to include something like a look file here to get it working.

.github/workflows/cancel-previous-workflow.yml
```diff
@@ -3,11 +3,15 @@ on:
...
   cancel:
+    name: Cancel Previous Runs
+    permissions:
+      actions: write
     runs-on: ubuntu-latest
     steps:
-    - uses: styfle/cancel-workflow-action@0.11.0
-      with:
-        workflow_id: all
+      - uses: styfle/cancel-workflow-action@0.11.0
```
rfay, line 15: Different yaml editors seem to disagree about the placement of array items, some put at top level, some indented. In general, it's better not to change things that don't relate to the PR being worked on. I'm OK with this, but if we're going to insist on indentation, we should agree on a standard.

.github/workflows/colima-tests.yml
```diff
@@ -34,16 +34,12 @@ env:
...
-
 permissions:
   contents: read
 
 jobs:
   tests:
     name: colima-${{ matrix.tests }}-no-bind-mounts=${{ matrix.no-bind-mounts }}
-    defaults:
-      run:
-        shell: bash
```
rfay, line 46: I think this exists because the default shell for macOS these days is zsh. Why remove it?

.github/workflows/tests.yml
```diff
@@ -77,11 +73,61 @@ jobs:
...
+          key: ddev-test-cache-${{ steps.get-date.outputs.date }}
+          restore-keys: |
+            ddev-test-cache-
+
+      - name: DDEV test cache/restore
+        uses: actions/cache/restore@v3
+        if: ${{ github.ref != 'refs/heads/master' }}
+        with:
+          path: ~/.ddev/testcache
```
rfay, line 96: Since this restores the whole testcache, including the extracted stuff which is duplicate, is there a way we can make it less redundant? We could either restore the extracted or the tarballs. I think just the tarballs would be better, but doing both seems like a lot.

pkg/testcommon/testcommon.go
```diff
@@ -320,17 +327,22 @@ func GetCachedArchive(siteName string, prefixString string, internalExtractionPa
...
 	}
+
 	if err != nil {
 		_ = fileutil.PurgeDirectory(extractPath)
 		_ = os.RemoveAll(extractPath)
 		_ = os.RemoveAll(archiveFullPath)
 		return extractPath, archiveFullPath, fmt.Errorf("archive extraction of %s failed err=%v", archiveFullPath, err)
 	}
+
+	output.UserOut.Printf("Extracted %s into %s", archiveFullPath, extractPath)
```
rfay, line 344: Judging from my manual test this seems not to have the right info.

rfay (changes requested):
This seems to dump the tarball into two places (root of testcache and also inside tarballs directory):

```
$ ls -l . tarballs/
.:
total 0
drwxr-xr-x  21 rfay  staff  672 Jul  3 10:04 82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz
drwxr-xr-x   3 rfay  staff   96 Jul  3 10:04 tarballs

tarballs/:
total 43056
-rw-r--r--  1 rfay  staff  21171263 Jul  3 10:04 82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz
```

The one in the root is the extraction of the one in the tarballs directory I guess. It's pretty confusing. 

However, the output shows otherwise, I guess it shows the old destination?
```
Downloading https://wordpress.org/wordpress-6.0.1.tar.gz
Downloaded https://wordpress.org/wordpress-6.0.1.tar.gz into /Users/rfay/.ddev/testcache/tarballs/82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz
Extracted /Users/rfay/.ddev/testcache/tarballs/82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz into /Users/rfay/.ddev/testcache/82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz
Copying directory /Users/rfay/.ddev/testcache/82f32c881c851f923c3442367116fccc8617c7bcc4efa35ea2e649363a3d1b27.tar.gz to /Users/rfay/tmp/ddevtest/TestPkgWordpress517092105
```

Please make the tarball have a name that can be studied and understood. For example, this one is for wordpress. Can we keep a name and date on it? I'm ok with it having a hash in the name of course. 

My results of the testcache directory after running TestDdevStart with DDEV_DEBUG=true: https://gist.github.com/rfay/585cccdcb0a8814c371ea2655e862df2

.github/workflows/colima-tests.yml
```diff
@@ -69,29 +65,74 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-lima-cache-
+
+      - name: Lima cache/restore
+        uses: actions/cache/restore@v3
+        if: ${{ github.ref != 'refs/heads/master' }}
         with:
-          path: ~/.lima
-          key: ${{ runner.os }}-build-${{ env.cache-name }}
+          path: ~/Library/Caches/lima
```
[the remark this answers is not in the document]
gilbertsoft: According the docs it is https://github.com/lima-vm/lima/blob/master/docs/internal.md#lima-cache-directory-librarycacheslima and your folder is the lima home https://github.com/lima-vm/lima/blob/master/docs/internal.md#lima-home-directory-lima_home

.github/workflows/tests.yml
```diff
@@ -77,11 +73,61 @@ jobs:
...
+          key: ddev-test-cache-${{ steps.get-date.outputs.date }}
+          restore-keys: |
+            ddev-test-cache-
```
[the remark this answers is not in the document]
gilbertsoft: No need here because ddev, ddev-test or ddev-test-cache anyway never exist.

.github/workflows/colima-tests.yml
gilbertsoft: Because it's already defined above.

.github/workflows/colima-tests.yml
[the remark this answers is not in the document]
rfay: I guess there are two things here. One is the Lima *cache* and the other is the Lima install. It's the Lima install where things are happening. 

```
$ du -sh ~/.lima ~/Library/Caches/lima
129G	/Users/rfay/.lima
2.5G	/Users/rfay/Library/Caches/lima
```

.github/workflows/colima-tests.yml
```diff
@@ -69,29 +65,74 @@ jobs:
...
+
+      - name: DDEV test cache
         uses: actions/cache@v3
-        env:
-          cache-name: cache-homebrew-cache-14
+        if: ${{ github.ref == 'refs/heads/master' }}
         with:
-          path: ~/Library/Caches/Homebrew
-          key: ${{ runner.os }}-build-${{ env.cache-name }}
+          path: ~/.ddev/testcache
```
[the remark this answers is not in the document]
gilbertsoft: Only caching ~/.ddev/testcache/tarballs should be fine.

.github/workflows/cancel-previous-workflow.yml
gilbertsoft: I had other changes in this file before but this change should also be reverted, true.

.github/workflows/tests.yml
gilbertsoft: Yeah, tarballs should be fine.

pkg/testcommon/testcommon.go
gilbertsoft: No, this is fine but will change anyway again because of the requested file name change.

.github/workflows/colima-tests.yml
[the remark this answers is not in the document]
gilbertsoft: Yes, exactly. ~/lima should be the place where volumes etc. are located, it's explained in my link before. Storing volumes does not look what we want, we should always have a clean new setup without any data from the last run imho.

cmd/ddev/cmd/clean.go
gilbertsoft: Searched in all tests, did not found anything. So safe to keep it.

.github/workflows/colima-tests.yml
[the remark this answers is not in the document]
rfay: My experience is that ~/.lima is what's really important, but it's been a while. I think including *additionally* ~/Library/Caches/lima has potential value, not sure. The bottom line is that we don't want Lima to take a thousand years to build. We don't care too much how up-to-date it is.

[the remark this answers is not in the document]
gilbertsoft: Lets keep it as is at the moment and only take care of testcache which was the initial intention of this patch anyway.

rfay (approved):
I sure like it better now! I'm very willing to give this a try. Should be a win for us. 

New testcache:

```
testcache
├── 00fbfea3_drupal-9.4.2.tar.gz
│   ├── INSTALL.txt
│   ├── LICENSE.txt
│   ├── README.md
│   ├── autoload.php
│   ├── composer.json
│   ├── composer.lock
│   ├── core
│   ├── example.gitignore
│   ├── index.php
│   ├── modules
│   ├── profiles
│   ├── robots.txt
│   ├── sites
│   ├── themes
│   ├── update.php
│   ├── vendor
│   └── web.config
├── 82f32c88_wordpress-6.0.1.tar.gz
│   ├── index.php
│   ├── license.txt
│   ├── readme.html
│   ├── wp-activate.php
│   ├── wp-admin
│   ├── wp-blog-header.php
│   ├── wp-comments-post.php
│   ├── wp-config-sample.php
│   ├── wp-content
│   ├── wp-cron.php
│   ├── wp-includes
│   ├── wp-links-opml.php
│   ├── wp-load.php
│   ├── wp-login.php
│   ├── wp-mail.php
│   ├── wp-settings.php
│   ├── wp-signup.php
│   ├── wp-trackback.php
│   └── xmlrpc.php
└── tarballs
    ├── 00fbfea3_drupal-9.4.2.tar.gz
    └── 82f32c88_wordpress-6.0.1.tar.gz

13 directories, 29 files
```

PR #5020, opened 2023-06-25
[tests only] Improve TestRouterPortsCheck random failure, limit use of TestNodeJSVersions

rfay:
## The Issue

TestRouterPortsCheck fails often, especially on colima slow Github actions test
TestNodeJSVersions often fails due to upstream problems or network problems. Don't test it in most places.

- Make sure that the port is actually occupied before continuing. 
- Introduce DDEV_SKIP_NODEJS_TEST, when "true", skip. It's set to true now for most test runners

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5020"></a>


rfay:
TestProjectPortsOverride seems to have the exact same problem (Colima) likely for exact same reason.

pkg/ddevapp/ddevapp_test.go
```diff
@@ -2911,21 +2911,30 @@ func TestRouterPortsCheck(t *testing.T) {
...
+	// The ddev-webserver healthcheck should make sure that we have a legitimate occupation
+	// of the port by the time it comes up.
+	portBinding := map[docker.Port][]docker.PortBinding{
+		"80/tcp": {
+			{HostPort: "80"},
+			{HostPort: "443"},
+		},
+	}
+
+	containerID, out, err := dockerutil.RunSimpleContainer(versionconstants.GetWebImage(), t.Name()+"occupyport", nil, []string{}, []string{}, []string{"testnfsmount" + ":/nfsmount"}, "", false, true, nil, portBinding)
```
rfay, line 2925: This must have a dockerutil.WaitContainer() or equivalent

gilbertsoft (approved):
Looks good!

PR #5021, opened 2023-06-25
Try using wsl2 as an OS type for amplitude

rfay:
## The Issue

We'd rather see WSL2 as a separate OS type, instead of seeing it as just Linux. 

Try changing to "wsl2" if it is

Use this on WSL2 and see if it shows up in Amplitude.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5021"></a>

### DIFF
--- a/pkg/amplitude/amplitude.go
+++ b/pkg/amplitude/amplitude.go
@@ -44,11 +44,15 @@ func GetDeviceID() string {
 
 // GetEventOptions returns default options to be used when tracking an event.
 func GetEventOptions() (options ampli.EventOptions) {
+	osType := runtime.GOOS
+	if dockerutil.IsWSL2() {
+		osType = "wsl2"
+	}
 	options = ampli.EventOptions{
 		DeviceID:   GetDeviceID(),
 		AppVersion: versionconstants.DdevVersion,
 		Platform:   runtime.GOARCH,
-		OSName:     runtime.GOOS,
+		OSName:     osType,
 		Language:   os.Getenv("LANG"),
 		ProductID:  "ddev cli",
 	}
--- a/pkg/amplitude/amplitude.go
+++ b/pkg/amplitude/amplitude.go
@@ -48,6 +48,7 @@ func GetEventOptions() (options ampli.EventOptions) {
 	if dockerutil.IsWSL2() {
 		osType = "wsl2"
 	}
+
 	options = ampli.EventOptions{
 		DeviceID:   GetDeviceID(),
 		AppVersion: versionconstants.DdevVersion,
--- a/pkg/amplitude/amplitude_test.go
+++ b/pkg/amplitude/amplitude_test.go
@@ -1,6 +1,7 @@
 package amplitude_test
 
 import (
+	"github.com/ddev/ddev/pkg/dockerutil"
 	"os"
 	"runtime"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/amplitude/amplitude.go
```diff
@@ -44,11 +44,15 @@ func GetDeviceID() string {
 
 // GetEventOptions returns default options to be used when tracking an event.
 func GetEventOptions() (options ampli.EventOptions) {
+	osType := runtime.GOOS
+	if dockerutil.IsWSL2() {
+		osType = "wsl2"
+	}
```
gilbertsoft, line 50:
```suggestion
	}

```

gilbertsoft: I love a bit space after block statements :)

gilbertsoft (approved):

rfay:
The event seems to be right, 
```
AMPLITUDE: Track event:
        {EventType:Project EventOptions:{UserID: DeviceID:70ebc340d86052616a7995b308a4e87c2185e1e9e16c4351963b8661b744117e Time:1687740032448 InsertID: Library: LocationLat:0 LocationLng:0 AppVersion:v1.22.0-alpha3-5-g1f14dd87d VersionName: Platform:amd64 OSName:wsl2 OSVersion: DeviceBrand: DeviceManufacturer: DeviceModel: Carrier: Country: Region: City: DMA: IDFA: IDFV: ADID: AndroidID: Language:C.UTF-8 IP: Price:0 Quantity:0 Revenue:0 ProductID:ddev cli RevenueType: EventID:0 SessionID:0 PartnerID: Plan:0xc00030e980 IngestionMetadata:0xc0003108c0} EventProperties:map[Containers:[db dba web] Containers Omitted:[] Database Type:mariadb Database Version:10.4 Fail On Hook Fail:false ID:db0b110533cf8e1f429afb469bd35f70f684883b2e5381b889bdaab350143754 Mutagen Enabled:false NFS Mount Enabled:false Nodejs Version:18 PHP Version:8.1 Project Type:drupal9 Router:traefik Router Disabled:false Webserver Type:nginx-fpm] UserProperties:map[] Groups:map[] GroupProperties:map[] UserID: DeviceID:}
```

But I don't think we're capturing Amplitude events any more now; this is on WSL2:

```
AMPLITUDE: Error 'EOF', while reading event cache.
```

When it fails like that it absolutely shouldn't be saying "Usage statistics submitted". It should at least say that something went wrong.

rfay:
Seems to work:

image

rfay:
Now we need to decide whether this makes sense. It truly is a separate environment, but "environment" would be a better description than "OS". "Environment', though, is way over-used.

pkg/amplitude/amplitude.go
```diff
@@ -44,11 +44,16 @@ func GetDeviceID() string {
 
 // GetEventOptions returns default options to be used when tracking an event.
 func GetEventOptions() (options ampli.EventOptions) {
+	osType := runtime.GOOS
+	if dockerutil.IsWSL2() {
+		osType = "wsl2"
```
gilbertsoft, line 49:
```suggestion
		osType += " (wsl2)"
```

What about just adding a suffix?

gilbertsoft:
> Removing the .amplitude.cache made it submit again. AFAIK this system (WSL2) had not previously been used for amplitude.

Works also without removing the file. The file was empty before and that's the reason for this output. Optimized now with #5025 .

PR #5028, opened 2023-06-27
refactor: add new Amplitude Property DDEV-Environment

rfay:
## The Issue

In Amplitude, we want to be able to track the environment type we're working in, especially things like WSL2, gitpod, codespaces.

Add it.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5028"></a>


pkg/nodeps/ddev_environment.go
```diff
@@ -0,0 +1,28 @@
...
+	switch {
+	case IsCodespaces():
+		e = DDEVEnvironmentCodespaces
+	case IsGitpod():
+		e = DDEVEnvironmentGitpod
+	case IsWSL2():
+		e = DDEVEnvironmentWSL2
+	}
+	return e
+}
```
gilbertsoft, line 28:
```suggestion
func GetDDEVEnvironment() string {
	switch {
	case IsCodespaces():
		return DDEVEnvironmentCodespaces
	case IsGitpod():
		return DDEVEnvironmentGitpod
	case IsWSL2():
		return DDEVEnvironmentWSL2
	}

	return runtime.GOOS
}
```

pkg/amplitude/amplitude.go
```diff
@@ -233,8 +233,10 @@ func identify() {
 	dockerVersion, _ := dockerutil.GetDockerVersion()
 	dockerPlaform, _ := version.GetDockerPlatform()
 	timezone, _ := time.Now().In(time.Local).Zone()
+	ddevEnv := nodeps.GetDDEVEnvironment()
 
 	builder := ampli.Identify.Builder().
+		DdevEnvironment(ddevEnv).
```
gilbertsoft, line 239:
```suggestion

	builder := ampli.Identify.Builder().
		DdevEnvironment(nodeps.GetDDEVEnvironment()).
```

pkg/nodeps/ddev_environment.go
```diff
@@ -0,0 +1,28 @@
+package nodeps
```
gilbertsoft, line 1:
```suggestion
package environment
```
What about a new package here?

pkg/nodeps/wsl.go
```diff
@@ -0,0 +1,25 @@
+package nodeps
```
gilbertsoft, line 1:
```suggestion
package environment
```

Other environment functions should be moved too if you decide to go with a new package.

gilbertsoft (changes requested):
Changes on Amplitude are fine btw. Last step before we merge this one is to merge your new branch on Amplitude and checkout main here.

gilbertsoft:
> I was thinking if we used a new environment we could get rid of IsGitpod() several other things, and instead something like "GetDdevEnvironment() == Environment.Gitpod"

I would do it the other way around and always use helper methods/funcs for this. In the end leads to less code changes if sth in the config changes. And calling a method/func is easier than looking for the correct constant to use.

pkg/nodeps/ddev_environment.go
rfay: I like the original better, following the old rule of trying to have one return from a function. This could also be done with a `default: `e = runtime.GOOS`

pkg/amplitude/amplitude.go
rfay: Done

pkg/nodeps/wsl.go
rfay: Done.

pkg/nodeps/ddev_environment.go
rfay: Done

pkg/nodeps/ddev_environment.go
gilbertsoft: Today early returns is the thing. But anyway does not matter much in this case. The impact here are the superfluous memory allocation you do with the return value, 3 allocations for the same value where one would be fine. This could be really bad for such a helper function which could be called multiple times during runtime.

pkg/nodeps/wsl.go
gilbertsoft: But it's still in nodeps, what have you done? I would move that file to the package. So in the end you should have 3 additional files in the environment package: wsl.go, gitpod.go and codespaces.go. Then we can call  e.g. `environment.IsWSL2()` which really looks nicer than `nodeps.IsWSL2()`. But I have to say this refactoring is also out of scope, so maybe we better do this cleanup in a dedicated patch anyway.

pkg/environment/ddev_environment.go
gilbertsoft, line 1: I'd name this file just `environment` without the `ddev_` prefix.

rfay: Done

rfay:
This should be good to go.

pkg/amplitude/amplitude.go
```diff
@@ -235,6 +236,7 @@ func identify() {
 	timezone, _ := time.Now().In(time.Local).Zone()
 
 	builder := ampli.Identify.Builder().
+		DdevEnvironment(environment.GetDDEVEnvironment()).
```
gilbertsoft, line 239: Can't we just use Environment here? The Ddev prefix looks superfluous to me.

gilbertsoft (changes requested):

pkg/amplitude/amplitude.go
gilbertsoft: Talking about the Amplitude property, I don't care much about the func name here.

rfay:
This was intended to just be a small fix to allow us to study the various environments. Would you mind if we stop haggling over the little things here?

gilbertsoft:
> This was intended to just be a small fix to allow us to study the various environments. Would you mind if we stop haggling over the little things here?

Problem is once it's merge to main in Amplitude we can not change it anymore. We should take really care about clean data in Amplitude, we already had a mess which makes it hard to set up correct charts.

rfay:
I guess I think "DdevEnvironment" is what we're talking about here, and "environment" is inadequate to describe it. What we're trying to report here is "The environment that DDEV is running in", not more general things that the way-overused term "environment" might mean. 

There may be other ways to describe this that don't even use the word "environment".

PR #5034, opened 2023-06-27
Add healthcheck to router so we wait for it to be started

rfay:
## The Issue

We saw some test failures, TestDdevFullSiteSetup is failing on both [linux](https://github.com/ddev/ddev/actions/runs/5384364660/jobs/9772197856?pr=5028) and [colima](https://github.com/ddev/ddev/actions/runs/5392269854/jobs/9790378473?pr=4904). seems to fail on checking port 8025. Fails on WordPress in each case.  But it may be because that’s the first test? (Traefik has no healthcheck)

## TODO:

- [x] For ddev-webserver and I think dbserver we have a special healthcheck that doesn't keep active all the time, and this is important for battery and CPU usage.

Add healthcheck to ddev-router when Traefik

Ordinary testing. See how our tests go.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5034"></a>

### DIFF
--- a/cmd/ddev/cmd/describe_test.go
+++ b/cmd/ddev/cmd/describe_test.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/stretchr/testify/require"
 	"runtime"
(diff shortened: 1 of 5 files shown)


rfay:
Not waiting for the healthcheck on Traefik could also be the cause of the intermittent TestCmdSSH failures, see https://buildkite.com/ddev/ddev-macos-amd64-mutagen/builds/4541#0188fe35-33af-45e1-a348-1d8a4a05f62e

gilbertsoft (approved):

gilbertsoft:
Saw this failure today which should be already fixed with #4975 . Looks like it's not included, branch here is not up to date with master.
grafik

rfay:
The panic in `ddev delete images` is because in the buildkite test runners the latest stable is used, and the test.sh attempts to `ddev delete images`. I think after we have alpha4 out I'll put it on there. But yesterday I deleted all images on the buildkite runners.

PR #5038, opened 2023-06-28
ci: enforce commit message convention, fixes #5037 [skip ci]

### LINKED ISSUE #5037: Use standardized PR titles, resulting in standardized master-branch commit messages
DDEV should also use some standards in PR titles and resulting merged commit messages. The commit message is generated from the PR title so the PR title has to follow the convention, we do not care about commit messages in a PR itself as it gets squash merged anyway. This has various advantages, e.g. we can find bugfixes to be back ported to a previous minor release. There is the widely used standard <https://www.conventionalcommits.org> which we can relay on.

This opens also new abillities e.g. automatically adding labels to PRs which later can be used to group the change log in the release by type.

## Definition of the commit message

`<type>[optional !]: <description>[, fixes #<issue>]`

This corresponds to the *Conventional Commits* and in addition the following rules must appy:

### Type
Must be one of the following:

* **build**: Changes that affect the build system or external dependencies
* **ci**: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
* **docs**: Documentation only changes
* **feat**: A new feature
* **fix**: A bug fix
* ***perf**: A code change that improves performance* (not sure if this has a value and is needed for DDEV at all)
* **refactor**: A code change that neither fixes a bug nor adds a feature
* **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
* **test**: Adding missing tests or correcting existing tests

Regarding SemVer, all types above except `feat` increase the patch version, `feat` increases the minor version.

### Scope
No scope must be used.

### Breaking Changes
Breaking changes must have a `!` appended after type/scope.

Regarding SemVer, breaking changes increase the major version.

### Subject / Description
The subject contains a succinct description of the change:

* use the imperative, present tense: "change" not "changed" nor "changes"
* don't capitalize the first letter
* no dot (.) at the end

If an issue exists for the change, `, fixes #<issue number>` must be appended to the subject.

### Revert
If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.

## Enforce the convention

The final definition will be enforced using https://github.com/marketplace/actions/gs-commit-message-checker, here is a prepared RegEx https://regex101.com/r/Dq4iGI/1

gilbertsoft:
* #5037 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5038"></a>

### DIFF
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,0 +1,61 @@
+version: v1
+
+labels:
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^build(|!): .+'
+
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^ci(|!): .+'
+
+  - label: 'documentation'
+    sync: true
+    matcher:
+      title: '^docs(|!): .+'
+
+  - label: 'enhancement'
+    sync: true
+    matcher:
+      title: '^feat(|!): .+'
+
+  - label: 'bugfix'
+    sync: true
+    matcher:
+      title: '^fix(|!): .+'
+
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^refactor(|!): .+'
+
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^test(|!): .+'
+
+  - label: 'breaking'
+    sync: true
+    matcher:
+      title: '^[a-z]+!: .+'
+
+  - label: 'dependencies'
+    sync: true
+    matcher:
+      files:
+        any: ["go.*", "vendor/**"]
+
+checks:
+  - context: "Semantic Pull Request"
+    url: "https://ddev.readthedocs.io/en/latest/developers/building-contributing/#pull-request-title-guidelines"
+    description:
+      success: Ready for review.
+      failure: Missing semantic label.
+    labels:
+      any:
+        - bugfix
+        - documentation
+        - enhancement
+        - maintenance
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -1,0 +1,34 @@
+name: PR Check
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize, ready_for_review, review_requested]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+  statuses: write
+  checks: write
+      
+jobs:
+  check_pr_title:
+    name: Check PR Title
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Type
+        uses: gsactions/commit-message-checker@v2
+        with:
+          pattern: '^(build|ci|docs|feat|fix|refactor|test)(|!): [a-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)'
+          flags: ''
+          error: 'Title does not follow our conventions see https://ddev.readthedocs.io/en/latest/developers/building-contributing/#pull-request-title-guidelines.'
+
+  assign_labels:
+    name: Assign Labels
+    needs: check_pr_title
+    runs-on: ubuntu-latest
+    steps:
+      - name: Assign and check labels
+        uses: fuxingloh/multi-labeler@v2
+        with:
+          config-path: .github/pr-labeler.yml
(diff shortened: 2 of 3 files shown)


rfay:
Please make sure to lead the docs (and error messages from github actions) with several examples of valid titles. Most of us can't read through the requirements, we just want to do it right. Examples will help a lot.

.github/pr-labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+labels:
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^build(|!): .+'
+
+  - label: 'maintenance'
+    sync: true
+    matcher:
+      title: '^ci(|!): .+'
```
rfay, line 12: You'll have to explain why there are two of these.

.github/workflows/pr-check.yml
```diff
@@ -0,0 +1,34 @@
...
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+  statuses: write
+  checks: write
+      
+jobs:
+  check_commit_message:
+    name: Check Commit Message
```
rfay, line 16: Looks like this is checking on ordinary commits? Is that true? We really only want it to check on master, as discussed.

docs/content/developers/building-contributing.md
```diff
@@ -266,6 +266,64 @@ Once you’ve opened a pull request, a discussion will start around your propose
 
 If your pull request is merged, great! If not, no sweat; it may not be what the project maintainer had in mind, or they were already working on it. This happens, so our recommendation is to take any feedback you’ve received and go forth and pull request again. Or create your own open source project.
 
+### Pull Request Title Guidelines
+
+We have very precise rules over how our Git commit messages can be formatted. This leads to **more readable messages** that are easy to follow when looking through the **project history**. But also, we use the Git commit messages to **generate the changelog** for the releases.
```
rfay, line 271:
```suggestion
We have very precise rules over how our PR titles (and thus master-branch commits) are to be formatted. This leads to **more readable messages** that are easy to follow when looking through the **project history**. But also, we use the master-branch Git commit messages to **generate the changelog** for the releases.
```

docs/content/developers/building-contributing.md
```diff
@@ -266,6 +266,64 @@ Once you’ve opened a pull request, a discussion will start around your propose
...
+* don't capitalize the first letter
+* no dot (.) at the end
+
+If an issue exists for the change, `, fixes #<issue number>` must be appended to the subject.
+
+#### Revert
+
+If the commit reverts a previous commit, it should begin with `revert:`, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
+#### Examples
```
rfay, line 317: I think move this to the top, where people will see it right away, right before "Type".

docs/content/developers/building-contributing.md
```diff
@@ -266,6 +266,64 @@ Once you’ve opened a pull request, a discussion will start around your propose
...
+`<type>[optional !]: <description>[, fixes #<issue>]`
+
+#### Type
+
+Must be one of the following:
+
+* **build**: Changes that affect the build or external dependencies
+* **ci**: Changes to our CI configuration files and scripts
+* **docs**: Documentation only changes
+* **feat**: A new feature
```
rfay, line 284: I see "enhancement" several places, is "feat" same as that?

rfay (changes requested):
Looking good. I'm concerned that we haven't run this on master yet, have you run these on master on your fork? If not, please experiment a bit on your fork with these on master.

.github/workflows/pr-check.yml
gilbertsoft: That's a question of the configuration of the commit-message-checker bellow. Here it is configured to look at PR titles and body only. And as you can see above the event is pull_request_target (btw we need target here for the labeler or no labels could be added)

docs/content/developers/building-contributing.md
gilbertsoft: done

docs/content/developers/building-contributing.md
gilbertsoft: done

rfay (approved):
Sounds good, thanks!

Please make sure to test to your satisfaction on your fork before pulling.

gilbertsoft:
Ready to be merged once @mattstein approved. Tested with Randy on my cloned repo and works like expected. Next step will be prepare release drafts based on the new labels.

PR #5041, opened 2023-06-28
test: stop testing TestDdevFullSiteSetup on Windows/Colima [skip ci]

rfay:
## The Issue

Traditional Windows and Colima testing are very slow and TestDdevFullSiteSetup is particularly problematic with both.

We don't need TestDdevFullSiteSetup on Windows/Colima. Stop testing it

Future:

For ddevapp package, we could easily change the TestMain to skip certain tests automatically based on some general criteria instead of doing one-off changes like this.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5041"></a>


rfay:
Interestingly enough, https://buildkite.com/ddev/ddev-windows-mutagen/builds/4090#01890347-bfd6-4bba-808d-fc93bf430a6c also failed in TestDdevFullSiteSetup on a different test runner.

PR #5005, opened 2023-06-21
feat: allow multiple upload dirs, fixes #4190, fixes #4796, fixes #5047

gilbertsoft:
## The Issue

- #4190
- #4796
- #5047 

## How This PR Solves The Issue

A new config property `upload_dirs` is introduced to hold more than one upload dir. Existing config `upload_dir` is migrated to the new config and removed from the projects afterwards. For project types which do not predefine a default upload dir like `php` the warning about the missing upload dir can be disabled by setting `upload_dirs: false` which is also mentioned in the warning message.

## Manual Testing Instructions

- [x] `upload_dir` is migrated to `upload_dirs`
- [x] multiple dirs can be set
- [x] `import-files` works correctly when importing in a upload dir using the `--target` flag
- [x] configuration with `ddev config` works properly, using `--upload-dirs` and `--uplod-dir` flags

## Related Issue Link(s)

- #5047

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5005"></a>

Issue #4796, opened 2023-04-01
`upload_dir` should be able to be set to empty to prevent bind-mounting


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

If you set `upload_dir: ""` you'd expect it to then ignore the idea of upload_dir, but right now "empty" is considered "no upload_dir configured"

### Actual Behavior

`upload_dir: ""` will result in warnings about upload_dir under mutagen, and no bind-mount set up. 

### Related 

* https://github.com/ddev/ddev/issues/4190

rfay:
No @gilbertsoft I would expect
```
upload_dirs: []
```

But I do agree that there's no way to distinguish between that and not specifying the item in the first place.

Since upload_dir is currently relative to docroot, your suggestion would mean "upload_dir" is "docroot" I think.

gilbertsoft:
fyi @rfay , solved by letting users set the `upload_dirs` to `false`.

```
You have mutagen enabled and your 'php' project type doesn't have an upload_dirs set. 
For faster startup and less disk usage, set upload_dirs to where your user-generated files are stored. 
If this is intended you can disable this warning by running `ddev config --upload-dir=false`. 
```

PR #5042, opened 2023-06-28
fix: avoid removal of bind mounts in `ddev composer create`, fixes #5031

### LINKED ISSUE #5031: Error when using upload-dirs=cpresources flag with 'ddev composer create' (with mutagen enabled)

```
ddev debug test
Running bash [-c /var/folders/n8/0fzdvc910h3_z38zq4gnp7480000gn/T/test_ddev.sh]
======= Existing project config =========
These config files were loaded for project <project_name>: [/Users/<user>/Sites/<project_name>/www/.ddev/config.yaml /Users/<user>/Sites/<project_name>/www/.ddev/config.arm64.yaml]
name: <project_name>
type: craftcms
docroot: web
php_version: 8.1
webserver_type: apache-fpm
webimage: drud/ddev-webserver:v1.21.5
router_http_port: 80
router_https_port: 443
database: {mysql 8.0}
upload_dir: public
mailhog_port: 8025
mailhog_https_port: 8026
phpmyadmin_port: 8036
phpmyadmin_https_port: 8037
webimage_extra_packages: [chromium]
project_tld: ddev.site
use_dns_when_possible: true
timezone: Europe/Brussels
composer_version: 2
nodejs_version: 16
default_container_timeout: 120
======= Creating dummy project named  tryddevproject-1154 in ../tryddevproject-1154 =========
OS Information: Darwin MBP-van-<user>.telenet.be 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:20 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T6000 arm64
ProductName:		macOS
ProductVersion:		13.4.1
BuildVersion:		22F82
User information: uid=501(<user>) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),98(_lpadmin),701(com.apple.sharepoint.group.1),703(com.apple.sharepoint.group.3),702(com.apple.sharepoint.group.2),33(_appstore),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),398(com.apple.access_screensharing),399(com.apple.access_ssh),400(com.apple.access_remote_ae)
DDEV version:  ITEM             VALUE
 DDEV version     v1.21.6
 architecture     arm64
 db               drud/ddev-dbserver-mariadb-10.4:v1.21.5
 dba              phpmyadmin:5
 ddev-ssh-agent   drud/ddev-ssh-agent:v1.21.5
 docker           24.0.2
 docker-compose   v2.15.1
 docker-platform  docker
 mutagen          0.16.0
 os               darwin
 router           drud/ddev-router:v1.21.5
 web              drud/ddev-webserver:v1.21.5
PROXY settings: HTTP_PROXY='' HTTPS_PROXY='' http_proxy='' NO_PROXY=''
======= DDEV global info =========
Global configuration:
instrumentation-opt-in=true
omit-containers=[]
mutagen-enabled=true
nfs-mount-enabled=false
router-bind-all-interfaces=false
internet-detection-timeout-ms=3000
disable-http2=false
use-letsencrypt=false
letsencrypt-email=
table-style=default
simple-formatting=false
auto-restart-containers=false
use-hardened-images=false
fail-on-hook-fail=false
required-docker-compose-version=
use-docker-compose-from-path=false
project-tld=
xdebug-ide-location=
no-bind-mounts=false
use-traefik=false
wsl2-no-windows-hosts-mgt=false

======= DOCKER info =========
docker location: lrwxr-xr-x@ 1 <user>  staff  53 14 jun 07:18 /Users/<user>/.orbstack/bin/docker -> /Applications/OrbStack.app/Contents/MacOS/xbin/docker
Docker Desktop Version: Print: Entry, ":CFBundleShortVersionString", Does Not Exist
Print: Entry, ":CFBundleVersion", Does Not Exist
Docker Desktop for Mac File build Doesn'tDocker Desktop for Mac Exist, build WillDocker Desktop for Mac Create: build /Applications/Docker.app/Contents/Info.plistDocker Desktop for Mac File build Doesn'tDocker Desktop for Mac Exist, build WillDocker Desktop for Mac Create: build /Applications/Docker.app/Contents/Info.plist
docker version:
Client:
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:51:16 2023
 OS/Arch:           darwin/arm64
 Context:           default

Server:
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f9ee60f147020bdd444b26e4b5c636dc28
  Built:            Sat May 27 17:14:34 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          v1.7.2
  GitCommit:        0cae528dd6cb557f7201036e9f43420650207b58
 runc:
  Version:          1.1.7
  GitCommit:        860f061b76bb4fc671f0f9e900f7d80ff93d4eb7
 docker-init:
  Version:          0.19.0
  GitCommit:
DOCKER_DEFAULT_PLATFORM=notset
======= Mutagen Info =========
Mutagen is installed in ddev, version=0.16.0
--------------------------------------------------------------------------------
Name: <project_name>
Identifier: sync_cxjbKBouzb7HtUSd4APFy3ti5E7v1PLC2GZOhTvAQag
Labels:
	com.ddev.volume-signature: Users-<user>-orbstack-run-1687877890
Configuration:
	Synchronization mode: Two Way Resolved
	Maximum allowed entry count: Default (2⁶⁴−1)
	Maximum staging file size: Default (18 EB)
	Symbolic link mode: POSIX Raw
	Ignore VCS mode: Default (Propagate)
	Ignores:
		/.git
		/.tarballs
		/.ddev/db_snapshots
		/.ddev/.importdb*
		.DS_Store
		.idea
		/web/public
	Permissions mode: Default (Portable)
Alpha:
	URL: /Users/<user>/Sites/<project_name>/www
	Configuration:
		Watch mode: Default (Portable)
		Watch polling interval: Default (10 seconds)
		Probe mode: Default (Probe)
		Scan mode: Default (Accelerated)
		Stage mode: Neighboring
		File mode: Default (0600)
		Directory mode: Default (0700)
		Default file/directory owner: Default
		Default file/directory group: Default
	Connected: No
Beta:
	URL: docker://ddev-<project_name>-web/var/www/html
		DOCKER_HOST=unix:///Users/<user>/.orbstack/run/docker.sock
	Configuration:
		Watch mode: Default (Portable)
		Watch polling interval: Default (10 seconds)
		Probe mode: Default (Probe)
		Scan mode: Default (Accelerated)
		Stage mode: Neighboring
		File mode: Default (0600)
		Directory mode: Default (0700)
		Default file/directory owner: Default
		Default file/directory group: Default
	Connected: No
Status: [Paused]
--------------------------------------------------------------------------------
======= Docker Info =========
Docker platform: docker
Using docker context: default (unix:///Users/<user>/.orbstack/run/docker.sock)
docker-compose: v2.15.1
Using DOCKER_HOST=unix:///Users/<user>/.orbstack/run/docker.sock
Docker version: 24.0.2
Able to run simple container that mounts a volume.
Able to use internet inside container.
Docker disk space:
Filesystem                Size      Used Available Use% Mounted on
overlay                 118.5G     26.4G     92.2G  22% /

Stopped mutagen daemon
The ddev-ssh-agent container has been removed. When you start it again you will have to use 'ddev auth ssh' to provide key authentication again.
Existing docker containers:
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
Creating a new ddev project config in the current directory (/Users/<user>/Sites/<project_name>/tryddevproject-1154)
Once completed, your configuration will be written to /Users/<user>/Sites/<project_name>/tryddevproject-1154/.ddev/config.yaml

Configuring unrecognized codebase as project type 'php' at /Users/<user>/Sites/<project_name>/tryddevproject-1154/web
Network ddev_default created
Starting tryddevproject-1154...
ssh-agent container is running: If you want to add authentication to the ssh-agent container, run 'ddev auth ssh' to enable your keys.
Network ddev-tryddevproject-1154_default  Created
Container ddev-tryddevproject-1154-web  Started
Container ddev-tryddevproject-1154-dba  Started
Container ddev-tryddevproject-1154-db  Started
You have mutagen enabled and your 'php' project type doesn't have an upload_dir set.
For faster startup and less disk usage,
set upload_dir to where your user-generated files are stored.
Starting mutagen sync process... This can take some time.
Mutagen sync flush completed in 1s.
For details on sync status 'ddev mutagen st tryddevproject-1154 -l'
Successfully started tryddevproject-1154
Project can be reached at https://tryddevproject-1154.ddev.site https://127.0.0.1:32935
======== Curl of site from inside container:
HTTP/1.1 200 OK
Server: nginx
Date: Tue, 27 Jun 2023 15:38:01 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Vary: Accept-Encoding

======== curl -I of http://tryddevproject-1154.ddev.site from outside:
HTTP/1.1 200 OK
Server: nginx/1.20.1
Date: Tue, 27 Jun 2023 15:38:01 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Vary: Accept-Encoding

======== full curl of http://tryddevproject-1154.ddev.site from outside:
Success accessing database... db via TCP/IP
ddev is working. You will want to delete this project with 'ddev delete -Oy tryddevproject-1154'
======== Project ownership on host:
drwxr-xr-x  4 <user>  staff  128 27 jun 17:37 ../tryddevproject-1154
======== Project ownership in container:
drwxr-xr-x 1 <user> dialout 16 Jun 27 15:37 /var/www/html
======== In-container filesystem:
Filesystem     Type  1K-blocks     Used Available Use% Mounted on
/dev/vdb1      btrfs 124284420 27851868  96452440  23% /var/www
======== curl again of tryddevproject-1154 from host:
Success accessing database... db via TCP/IP
ddev is working. You will want to delete this project with 'ddev delete -Oy tryddevproject-1154'
Thanks for running the diagnostic. It was successful.
Please provide the output of this script in a new gist at gist.github.com
Running ddev launch in 5 seconds
If you're brave and you have jq you can delete all tryddevproject instances with this one-liner:
    ddev delete -Oy $(ddev list -j |jq -r .raw[].name | grep tryddevproject)
In the future ddev debug test will also provide this option.

Please delete this project after debugging with 'ddev delete -Oy tryddevproject-1154'
```
</details>         

### Expected Behavior

Successful creation of the project with the specified upload directory.

### Actual Behavior

<details><summary>Error occurs during the project creation process when the upload-dir flag is set to 'public'. The following error message is displayed</summary>

```
rsync: [receiver] mkstemp "/var/www/html/web/public/.gitignore.IYVNFK" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1333) [sender=3.2.3]
Failed to create project: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-<project_name> docker-compose -f /<generic_user>/Sites/<project_name>/www/.ddev/.ddev-docker-compose-full.yaml exec -w /var/www/html -T web bash -c set -eu && ( rsync -rltgopD "/tmp/TQYtMb/" "/var/www/html/")', action='[exec -w /var/www/html -T web bash -c set -eu && ( rsync -rltgopD "/tmp/TQYtMb/" "/var/www/html/")]', err='exit status 23', stdout='', stderr='rsync: [receiver] mkstemp "/var/www/html/web/public/.gitignore.IYVNFK" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1333) [sender=3.2.3]'
```
</details> 

### Steps To Reproduce

1. Execute the following command to configure DDEV: 
`ddev config --project-name=<project_name> --project-type=craftcms --docroot=web --create-docroot --upload-dir=public`
3. Run the `ddev composer create -y --no-install --no-scripts <package>` command.
4. Observe the error message mentioned above during the project creation process.

### Anything else?

It is important to note that the issue described above only occurs when using the upload-dir flag. Without the upload-dir flag, the project creation process works correctly.

rfay:
## The Issue

* #5031

Seems only to be a problem with mutagen enabled, when the upload_dir is bind-mounted separately

Create the upload_dir if needed during `ddev composer create`

```
ddev config --project-type=craftcms --docroot=web --create-docroot --upload-dir=cpresources
ddev start
ddev composer create -y --no-scripts craftcms/craft
```
## Automated Testing Overview

Nothing new added

## Related Issue Link(s)

Important:

* https://github.com/ddev/ddev/pull/5005
* https://github.com/ddev/ddev/issues/4190
* https://github.com/ddev/ddev/issues/4796

Probably not relevant:
* https://github.com/ddev/ddev/issues/4650

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5042"></a>

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -79,17 +78,15 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 		// Remove most contents of composer root
 		util.Warning("Removing any existing files in composer root")
 
-		skip := []string{}
-		if app.IsMutagenEnabled() {
-			for _, uploadDir := range app.GetHostUploadDirsFullPath() {
-				uploadDir, found := strings.CutPrefix(uploadDir, composerRoot+"/")
-				if found && uploadDir != "" {
-					skip = append(skip, uploadDir)
-				}
+		exceptions := []string{".ddev", ".git", ".tarballs"}
+		for _, uploadDir := range app.GetHostUploadDirsFullPath() {
+			uploadDir, found := strings.CutPrefix(uploadDir, composerRoot+"/")
+			if found && uploadDir != "" {
+				exceptions = append(exceptions, uploadDir)
 			}
 		}
 
-		err = removeAll(composerRoot, []string{".ddev", ".git", ".tarballs"}, skip)
+		err = fileutil.RemoveAllExcept(composerRoot, exceptions)
 		if err != nil {
 			util.Failed("Failed to create project: %v", err)
 		}
@@ -239,62 +236,4 @@ func init() {
 	ComposerCreateCmd.Flags().BoolVarP(&composerCreateYesFlag, "yes", "y", false, "Yes - skip confirmation prompt")
 	ComposerCmd.AddCommand(ComposerCreateProjectCmd)
 	ComposerCmd.AddCommand(ComposerCreateCmd)
-}
-
-// removeAll removes all files and folders in path except the ones listed in
-// skipRecursive and skip where skip only preserves the folder itself but still
-// removes containing files and folders.
-func removeAll(aPath string, skipRecursive, skip []string) error {
-	fsObjects, err := fileutil.ListFilesInDir(aPath)
-	if err != nil {
-		return err
-	}
-
-ObjectsLoop:
-	for _, fsObject := range fsObjects {
-		// Just skip objects listed in skipRecursive.
-		for _, skipObject := range skipRecursive {
-			if fsObject == skipObject {
-				continue ObjectsLoop
-			}
-		}
-
-		for _, skipObject := range skip {
-			if strings.HasPrefix(skipObject, fsObject) {
-				// Build new skipRecursive removing current object prefix.
-				newSkipRecursive := []string{}
-				for _, skipObject := range skipRecursive {
-					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
-						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
-						newSkipRecursive = append(newSkipRecursive, skipObject)
-					}
-				}
-
-				// Build new skip removing current object prefix.
-				newSkip := []string{}
-				for _, skipObject := range skip {
-					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
-						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
-						newSkip = append(newSkip, skipObject)
-					}
-				}
-
-				// Recursive call of removeAll to remove content of object.
-				err = removeAll(path.Join(aPath, fsObject), newSkipRecursive, newSkip)
-				if err != nil {
-					return err
-				}
-
-				continue ObjectsLoop
-			}
-		}
-
-		// Nothing to skip, remove it all.
-		err = os.RemoveAll(filepath.Join(aPath, fsObject))
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
 }
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -95,6 +95,14 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 			}
 		}
 
+		// The upload_dir needs to exist for composer create-project to work
+		if app.GetHostUploadDirFullPath() != "" {
+			err = os.MkdirAll(app.GetHostUploadDirFullPath(), 0755)
+			if err != nil {
+				util.Failed("failed to create upload dir: %v", err)
+			}
+		}
+
 		err = app.MutagenSyncFlush()
 		if err != nil {
 			util.Failed("Failed to sync mutagen contents: %v", err)
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -80,8 +81,8 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 
 		exceptions := []string{".ddev", ".git", ".tarballs"}
 		for _, uploadDir := range app.GetHostUploadDirsFullPath() {
-			uploadDir, found := strings.CutPrefix(uploadDir, composerRoot+"/")
-			if found && uploadDir != "" {
+			uploadDir, err := filepath.Rel(composerRoot, uploadDir)
+			if err == nil && uploadDir != "" {
 				exceptions = append(exceptions, uploadDir)
 			}
 		}
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -81,9 +81,11 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 
 		skip := []string{}
 		if app.IsMutagenEnabled() {
-			uploadDir, found := strings.CutPrefix(app.GetHostUploadDirFullPath(), composerRoot+"/")
-			if found && uploadDir != "" {
-				skip = append(skip, uploadDir)
+			for _, uploadDir := range app.GetHostUploadDirsFullPath() {
+				uploadDir, found := strings.CutPrefix(uploadDir, composerRoot+"/")
+				if found && uploadDir != "" {
+					skip = append(skip, uploadDir)
+				}
 			}
 		}
 
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -8,13 +8,12 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/fileutil"
 	"github.com/ddev/ddev/pkg/nodeps"
-	"github.com/mattn/go-isatty"
-
-	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/output"
 	"github.com/ddev/ddev/pkg/util"
+	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 )
 
@@ -79,25 +78,18 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 
 		// Remove most contents of composer root
 		util.Warning("Removing any existing files in composer root")
-		objs, err := fileutil.ListFilesInDir(composerRoot)
+
+		skip := []string{}
+		if app.IsMutagenEnabled() {
+			uploadDir, found := strings.CutPrefix(app.GetHostUploadDirFullPath(), composerRoot+"/")
+			if found && uploadDir != "" {
+				skip = append(skip, uploadDir)
+			}
+		}
+
+		err = removeAll(composerRoot, []string{".ddev", ".git", ".tarballs"}, skip)
 		if err != nil {
 			util.Failed("Failed to create project: %v", err)
-		}
-
-		for _, o := range objs {
-			// Preserve .ddev, .git, .tarballs
-			if o == ".ddev" || o == ".git" || o == ".tarballs" {
-				continue
-			}
-
-			if err = os.RemoveAll(filepath.Join(composerRoot, o)); err != nil {
-				util.Failed("Failed to create project: %v", err)
-			}
-		}
-
-		// The upload_dir needs to exist for composer create-project to work
-		if app.GetHostUploadDirFullPath() != "" {
-			app.CreateUploadDirIfNecessary()
 		}
 
 		err = app.MutagenSyncFlush()
@@ -245,4 +237,62 @@ func init() {
 	ComposerCreateCmd.Flags().BoolVarP(&composerCreateYesFlag, "yes", "y", false, "Yes - skip confirmation prompt")
 	ComposerCmd.AddCommand(ComposerCreateProjectCmd)
 	ComposerCmd.AddCommand(ComposerCreateCmd)
+}
+
+// removeAll removes all files and folders in path except the ones listed in
+// skipRecursive and skip where skip only preserves the folder itself but still
+// removes containing files and folders.
+func removeAll(aPath string, skipRecursive, skip []string) error {
+	fsObjects, err := fileutil.ListFilesInDir(aPath)
+	if err != nil {
+		return err
+	}
+
+ObjectsLoop:
+	for _, fsObject := range fsObjects {
+		// Just skip objects listed in skipRecursive.
+		for _, skipObject := range skipRecursive {
+			if fsObject == skipObject {
+				continue ObjectsLoop
+			}
+		}
+
+		for _, skipObject := range skip {
+			if strings.HasPrefix(skipObject, fsObject) {
+				// Build new skipRecursive removing current object prefix.
+				newSkipRecursive := []string{}
+				for _, skipObject := range skipRecursive {
+					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
+						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
+						newSkipRecursive = append(newSkipRecursive, skipObject)
+					}
+				}
+
+				// Build new skip removing current object prefix.
+				newSkip := []string{}
+				for _, skipObject := range skip {
+					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
+						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
+						newSkip = append(newSkip, skipObject)
+					}
+				}
+
+				// Recursive call of removeAll to remove content of object.
+				err = removeAll(path.Join(aPath, fsObject), newSkipRecursive, newSkip)
+				if err != nil {
+					return err
+				}
+
+				continue ObjectsLoop
+			}
+		}
+
+		// Nothing to skip, remove it all.
+		err = os.RemoveAll(filepath.Join(aPath, fsObject))
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -97,10 +97,7 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 
 		// The upload_dir needs to exist for composer create-project to work
 		if app.GetHostUploadDirFullPath() != "" {
-			err = os.MkdirAll(app.GetHostUploadDirFullPath(), 0755)
-			if err != nil {
-				util.Failed("failed to create upload dir: %v", err)
-			}
+			app.CreateUploadDirIfNecessary()
 		}
 
 		err = app.MutagenSyncFlush()
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -79,7 +79,7 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 		// Remove most contents of composer root
 		util.Warning("Removing any existing files in composer root")
 
-		exceptions := []string{".ddev", ".git", ".tarballs"}
+		exceptions := []string{".ddev/*", ".git/*", ".tarballs/*"}
 		for _, uploadDir := range app.GetHostUploadDirsFullPath() {
 			uploadDir, err := filepath.Rel(composerRoot, uploadDir)
 			if err == nil && uploadDir != "" {
--- a/pkg/fileutil/path_test.go
+++ b/pkg/fileutil/path_test.go
@@ -1,0 +1,52 @@
+package fileutil_test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ddev/ddev/pkg/fileutil"
+	"github.com/ddev/ddev/pkg/testcommon"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRemoveAllExcept tests copying a directory.
+func TestRemoveAllExcept(t *testing.T) {
+	testDir, _ := os.Getwd()
+	assert := assert.New(t)
+	require := require.New(t)
+
+	sourceDir := filepath.Join(testDir, "testdata", "remove_all_except")
+	targetBaseDir := testcommon.CreateTmpDir("TestRemoveAllExcept")
+	targetDir := filepath.Join(targetBaseDir, "testdata")
+	defer os.RemoveAll(targetBaseDir)
+
+	err := fileutil.CopyDir(sourceDir, targetDir)
+	assert.NoError(err)
+
+	err = fileutil.RemoveAllExcept(targetDir, []string{"keep/*", "keep_partial", "sub/keep/*"})
+	assert.NoError(err)
+
+	require.DirExists(targetDir)
+
+	assert.DirExists(filepath.Join(targetDir, "keep"))
+	assert.DirExists(filepath.Join(targetDir, "keep", "keep_sub"))
+	assert.FileExists(filepath.Join(targetDir, "keep", "keep.txt"))
+	assert.FileExists(filepath.Join(targetDir, "keep", "keep_sub", "keep.txt"))
+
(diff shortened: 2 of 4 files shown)


gilbertsoft (changes requested):
Does not look this solves the problem. And I wonder if this is the correct place to create the directory, maybe `start` would be better.

rfay:
@gilbertsoft the upload_dir has always been created early in app.Start(), see https://github.com/ddev/ddev/blob/2d481a419f1745b6807da9eaf3781df53e75a4e1/pkg/ddevapp/ddevapp.go#L1155

The way I understood the problem, the upload_dir was being removed by the cleanup before composer actions and that's why it failed and needed to be recrated.

gilbertsoft:
> @gilbertsoft the upload_dir has always been created early in app.Start(), see
> 
> https://github.com/ddev/ddev/blob/2d481a419f1745b6807da9eaf3781df53e75a4e1/pkg/ddevapp/ddevapp.go#L1155
> 
> The way I understood the problem, the upload_dir was being removed by the cleanup before composer actions and that's why it failed and needed to be recrated.

I understand the problem in a different way, the folder is there but rsync is not able to sync the files. My guess is it's about the mounting of the upload_dir which does not allow rsync to access to the folder properly.

gilbertsoft (approved):

cmd/ddev/cmd/composer-create.go
```diff
@@ -79,20 +78,20 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
...
-		if err != nil {
-			util.Failed("Failed to create project: %v", err)
-		}
 
-		for _, o := range objs {
-			// Preserve .ddev, .git, .tarballs
-			if o == ".ddev" || o == ".git" || o == ".tarballs" {
-				continue
+		skip := []string{}
+		if app.IsMutagenEnabled() {
```
rfay, line 83: I think we should do this whether or not mutagen is enabled? No reason to remove those if they're upload_dirs. I know the issue is about when mutagen is enabled, but I don't think we should be removing the upload_dirs normally either.

cmd/ddev/cmd/composer-create.go
```diff
@@ -241,3 +240,61 @@ func init() {
 	ComposerCmd.AddCommand(ComposerCreateProjectCmd)
 	ComposerCmd.AddCommand(ComposerCreateCmd)
 }
+
+// removeAll removes all files and folders in path except the ones listed in
+// skipRecursive and skip where skip only preserves the folder itself but still
+// removes containing files and folders.
+func removeAll(aPath string, skipRecursive, skip []string) error {
```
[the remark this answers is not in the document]
rfay: RemoveAll also needs a test. Could be a very dangerous function.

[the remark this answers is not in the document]
rfay: Naming of args can be improved here. Maybe `skip`  can become `skipExplicit` ?

cmd/ddev/cmd/composer-create.go
```diff
@@ -241,3 +240,61 @@ func init() {
...
+	for _, fsObject := range fsObjects {
+		// Just skip objects listed in skipRecursive.
+		for _, skipObject := range skipRecursive {
+			if fsObject == skipObject {
+				continue ObjectsLoop
+			}
+		}
+
+		for _, skipObject := range skip {
+			if strings.HasPrefix(skipObject, fsObject) {
```
rfay, line 263: These are filepaths, we should be using filepath to do this, not strings.HasPrefix(). Otherwise we can make mistakes with directories that have a partial path (prefix)

cmd/ddev/cmd/composer-create.go
```diff
@@ -241,3 +240,61 @@ func init() {
...
+func removeAll(aPath string, skipRecursive, skip []string) error {
+	fsObjects, err := fileutil.ListFilesInDir(aPath)
+	if err != nil {
+		return err
+	}
+
+ObjectsLoop:
+	for _, fsObject := range fsObjects {
+		// Just skip objects listed in skipRecursive.
+		for _, skipObject := range skipRecursive {
```
rfay, line 256: Maybe I'm mistaken, but can't this look just be done outside the fsObjects loop? Doesn't need to be nested for-loops?

cmd/ddev/cmd/composer-create.go
```diff
@@ -241,3 +240,61 @@ func init() {
...
+			}
+		}
+
+		for _, skipObject := range skip {
+			if strings.HasPrefix(skipObject, fsObject) {
+				// Build new skipRecursive removing current object prefix.
+				newSkipRecursive := []string{}
+				for _, skipObject := range skipRecursive {
+					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
+						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
```
rfay, line 268: Again here we shouldn't be doing string manipulation on filepaths.

cmd/ddev/cmd/composer-create.go
```diff
@@ -241,3 +240,61 @@ func init() {
...
+				newSkip := []string{}
+				for _, skipObject := range skip {
+					if strings.HasPrefix(skipObject, fsObject) && skipObject != fsObject {
+						skipObject, _ = strings.CutPrefix(skipObject, fsObject+"/")
+						newSkip = append(newSkip, skipObject)
+					}
+				}
+
+				// Recursive call of removeAll to remove content of object.
+				err = removeAll(path.Join(aPath, fsObject), newSkipRecursive, newSkip)
```
rfay, line 283: I'm not convinced that recursion is called for here. Recursion should be used only when it's absolutely appropriate.

rfay (commented):
I confirm that the current code solves the immediate problem, but I'd like to see 
* a simpler function for RemoveAll(), without recursion, and perhaps more efficient. 
* Needs a test
* Don't use string manipulation and comparison for filepath tests
* Move to fileutils if it seems like it would be useful there. Perhaps this is a very edge case thing that doesn't need to move. 

Overall, it's too hard to validate the behavior of the new removeAll. I stepped through it and studied it, and it took a lot of work to fully understand. And it did a LOT of loops through there!

cmd/ddev/cmd/composer-create.go
gilbertsoft: No, because on the next line fsObject is used in the condition.

cmd/ddev/cmd/composer-create.go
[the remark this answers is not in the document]
gilbertsoft: Yeah, global function with tests sounds good!

rfay:
Rebased this as it needed to use the new mutagen version so the buildkite test runners can settle down.

rfay:
Needs work for TestComposerCmd; manually run it to demonstrate the problem locally, should be easy to fix.

pkg/fileutil/path_test.go
```diff
@@ -0,0 +1,52 @@
...
+// TestRemoveAllExcept tests copying a directory.
+func TestRemoveAllExcept(t *testing.T) {
+	testDir, _ := os.Getwd()
+	assert := assert.New(t)
+	require := require.New(t)
+
+	sourceDir := filepath.Join(testDir, "testdata", "remove_all_except")
+	targetBaseDir := testcommon.CreateTmpDir("TestRemoveAllExcept")
+	targetDir := filepath.Join(targetBaseDir, "testdata")
+	defer os.RemoveAll(targetBaseDir)
```
rfay, line 23: Much preferred here is t.Cleanup(), and catch and assert on the result of RemoveAll, so we know if there's been a problem.

rfay (commented):
I haven't manually tested this yet, or stepped through it, but some requests below. It's massively appropriate and should solve the problem.

I'm super worried about it deleting something that it shouldn't. What if people put `../../` as upload_dir, and that's outside the project? Let's make sure to test that case.

rfay:
We should consider whether we're doing the right thing at all. Do we need to do something this intrusive and possibly risky? Are there other ways to solve this problem?

rfay:
How about this: Let's just remove as we did previously. We warn them. Then we can recreate the listed upload_dirs. Then on mutagen, we can `app.Start()`?

gilbertsoft:
> How about this: Let's just remove as we did previously. We warn them. Then we can recreate the listed upload_dirs. Then on mutagen, we can `app.Start()` or `app.Restart()` ?

Two restarts in this command? Okay, normally you don't do that often but two restarts take a lot of time.

pkg/fileutil/path_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/ddev/ddev/pkg/testcommon"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRemoveAllExcept tests copying a directory.
+func TestRemoveAllExcept(t *testing.T) {
+	testDir, _ := os.Getwd()
+	assert := assert.New(t)
+	require := require.New(t)
```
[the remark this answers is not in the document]
gilbertsoft: Mine was happy, removal of the alias was intended. But reintroduced now.

pkg/fileutil/path_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	sourceDir := filepath.Join(testDir, "testdata", "remove_all_except")
+	targetBaseDir := testcommon.CreateTmpDir("TestRemoveAllExcept")
+	targetDir := filepath.Join(targetBaseDir, "testdata")
+	defer os.RemoveAll(targetBaseDir)
+
+	err := fileutil.CopyDir(sourceDir, targetDir)
+	assert.NoError(err)
+
+	err = fileutil.RemoveAllExcept(targetDir, []string{"keep/*", "keep_partial", "sub/keep/*"})
```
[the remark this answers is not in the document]
gilbertsoft: Added the two missing cases.

gilbertsoft:
> I haven't manually tested this yet, or stepped through it, but some requests below. It's massively appropriate and should solve the problem.
> 
> I'm super worried about it deleting something that it shouldn't. What if people put `../../` as upload_dir, and that's outside the project? Let's make sure to test that case.

We always delete from the base path and not from the exceptions, not possible something gets removed outside the path. The only risk is we do not delete all we should which shouldn't hurt but would change behavior.

rfay:
I think the current approach is just too risky. Since they can use free-form upload_dirs, we could end up destroying something BIG unintentionally. 

There was a bug early in DDEV's history where for one user we deleted their home directory :(

gilbertsoft:
> I think the current approach is just too risky. Since they can use free-form upload_dirs, we could end up destroying something BIG unintentionally.
> 
> There was a bug early in DDEV's history where for one user we deleted their home directory :(

There is no such risk in this code. Before we just removed all, now we exclude some folders from the removal. The risk to remove something unitentional is exactly the same like before.

rfay:
Congrats on 
> Failed to get project(s): /Users/rfay/workspace/c/.ddev/config.yaml exists but cannot be read. It may be invalid due to a syntax error: unable to load config file /Users/rfay/workspace/c/.ddev/config.yaml: invalid upload dir `../../y` outside of project root `/Users/rfay/workspace/c` found

rfay (commented):
In my test, walking through this, the first file that it wanted to delete was /Users/rfay/workspace/c/.ddev/.ddev-docker-compose-base.yaml, which should never have been deleted because .ddev is excluded. 

The second was /Users/rfay/workspace/c/.ddev/.ddev-docker-compose-full.yaml, which also should not have been deleted.

Then /Users/rfay/workspace/c/.ddev/.gitignore ...

and /Users/rfay/workspace/c/.ddev/.global_commands

/Users/rfay/workspace/c/.ddev/.homeadditions

/Users/rfay/workspace/c/.ddev/.webimageBuild

/Users/rfay/workspace/c/.ddev/apache

upload_dirs were cpresources, ../x, and one/two/three/four

It deleted the entire contents of the .ddev directory. 

What this demonstrates is just that the job is too hard, and we need to figure out an easier and less risky way to solve this problem.

rfay:
```
mkdir c && cd c
ddev config --project-type=craftcms --docroot=web --create-docroot --upload-dir=cpresources,../x,one/two/three/four
ddev start
ddev restart # to fix mutagen problem from https://github.com/ddev/ddev/pull/5094

$ ddev composer create -y --no-scripts craftcms/craft
Warning: MOST EXISTING CONTENT in the composer root (/Users/rfay/workspace/c) will be deleted by the composer create-project operation. Only .ddev, .git and .tarballs will be preserved.
Removing any existing files in composer root
Executing composer command: [composer create-project --no-scripts craftcms/craft --no-install /tmp/XDTzhh]

stat /Users/rfay/workspace/c/.ddev/.ddev-docker-compose-full.yaml: no such file or directory
Failed to create project:exit status 14, stderr=
```

gilbertsoft:
I've tested again according your description above and to verify which folders are deleted I've added additionals output of the walk and folders which are finally deleted. Everything works fine and like expected!

grafik
grafik

```
ddev composer create -y --no-scripts craftcms/craft
Warning: MOST EXISTING CONTENT in the composer root (/home/simon/c) will be deleted by the composer create-project operation. Only .ddev, .git and .tarballs will be preserved.
Removing any existing files in composer root
Walking `/home/simon/c`
Walking `/home/simon/c/.ddev`
Walking `/home/simon/c/.ddev/.dbimageBuild`
Walking `/home/simon/c/.ddev/.ddev-docker-compose-base.yaml`
Walking `/home/simon/c/.env`
Remove `/home/simon/c/.env`
Walking `/home/simon/c/.env.example.dev`
Remove `/home/simon/c/.env.example.dev`
Walking `/home/simon/c/.env.example.production`
Remove `/home/simon/c/.env.example.production`
Walking `/home/simon/c/.env.example.staging`
Remove `/home/simon/c/.env.example.staging`
Walking `/home/simon/c/.gitignore`
Remove `/home/simon/c/.gitignore`
Walking `/home/simon/c/bootstrap.php`
Remove `/home/simon/c/bootstrap.php`
Walking `/home/simon/c/composer.json`
Remove `/home/simon/c/composer.json`
Walking `/home/simon/c/composer.lock`
Remove `/home/simon/c/composer.lock`
Walking `/home/simon/c/config`
Remove `/home/simon/c/config`
Walking `/home/simon/c/config`
Remove `/home/simon/c/config`
Walking `/home/simon/c/craft`
Remove `/home/simon/c/craft`
Walking `/home/simon/c/storage`
Remove `/home/simon/c/storage`
Walking `/home/simon/c/storage`
Remove `/home/simon/c/storage`
Walking `/home/simon/c/templates`
Remove `/home/simon/c/templates`
Walking `/home/simon/c/templates`
Remove `/home/simon/c/templates`
Walking `/home/simon/c/vendor`
Remove `/home/simon/c/vendor`
Walking `/home/simon/c/vendor`
Remove `/home/simon/c/vendor`
Walking `/home/simon/c/web`
Walking `/home/simon/c/web/.htaccess`
Remove `/home/simon/c/web/.htaccess`
Walking `/home/simon/c/web/cpresources`
Walking `/home/simon/c/web/cpresources/.gitignore`
Remove `/home/simon/c/web/cpresources/.gitignore`
Walking `/home/simon/c/web/index.php`
Remove `/home/simon/c/web/index.php`
Walking `/home/simon/c/web/one`
Walking `/home/simon/c/web/one/two`
Walking `/home/simon/c/web/one/two/three`
Walking `/home/simon/c/web/one/two/three/four`
Walking `/home/simon/c/web/web.config`
Remove `/home/simon/c/web/web.config`
Walking `/home/simon/c/x`
Executing composer command: [composer create-project --no-scripts craftcms/craft --no-install /tmp/ngFvbz]
```

The only thing I see here is this

```
Walking `/home/simon/c/.ddev/.dbimageBuild`
Walking `/home/simon/c/.ddev/.ddev-docker-compose-base.yaml`
```

It does not hurt but there is no reason to walk these two folders/files at all. Also related to your following comment and could be optimized.

> This approach actually ends up walking down into excepted directories, which is inefficient but may not do any harm, but it does seem incorrect. For example, .ddev is excepted... but it keeps walking down into /Users/rfay/workspace/c/.ddev/.dbimageBuild

My additional output above proves that my code is almost perfect! Agree on the first condition in the exception loop which could be optimized and maybe afterwards there is no longer the need for normalization of the paths.

Did you use the latest sources? If you're `.ddev` gets removed you're not on the latest, this was fixed here https://github.com/ddev/ddev/commit/80432edc9264396da0bc6ee94cd34046396ae977#diff-b94120791d6f38e370c5ea4d8713d839c4037db6c7858b528c4a294b8a0163b3R82 and as you can see, the reason was not the new func but wrong exceptions were used!

rfay:
We need to find a simpler technique, I'm sure we can. Yes, I tested the rebase of what you had yesterday.

rfay:
@gilbertsoft if you end up thinking that https://github.com/ddev/ddev/pull/5058 is the best solution to this, we can consider going that direction.

rfay:
Closing in favor of 
* https://github.com/ddev/ddev/pull/5132

PR #5058, opened 2023-07-01
refactor: make ddev composer create less error-prone

gilbertsoft:


docs/content/users/quickstart.md
```diff
@@ -250,8 +249,7 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
             mkdir my-statamic-app
             cd my-statamic-app
             ddev config --project-type=laravel --docroot=public --create-docroot
-            ddev composer create --prefer-dist --no-install --no-scripts statamic/statamic
-            ddev composer install
+            ddev composer create --prefer-dist statamic/statamic
```
rpkoller, line 252: The only nitpick i would have is to change that line to `ddev composer create --prefer-dist statamic/statamic -y` that the composer create process could run through without any confirmation by the user.

rpkoller, line 252:
```suggestion
            ddev composer create --prefer-dist statamic/statamic -y
```

docs/content/users/quickstart.md
```diff
@@ -250,8 +249,7 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
             ddev exec "php artisan key:generate"
             ddev launch
```
rpkoller, line 254:
```suggestion
            ddev exec "php artisan key:generate"
            ddev exec "php please make:user" # Add a super user
            ddev launch
```

rpkoller: I was just uncertain if it would be the better choice to add the comment `# Add a super user` to a dedicated line before the please make:user command or if it is the better choice to make the comment in line like i did in the suggestion

docs/content/users/quickstart.md
```diff
@@ -250,8 +249,7 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
             mkdir my-statamic-app
             cd my-statamic-app
             ddev config --project-type=laravel --docroot=public --create-docroot
-            ddev composer create --prefer-dist --no-install --no-scripts statamic/statamic
-            ddev composer install
+            ddev composer create --prefer-dist statamic/statamic -y
             ddev exec "php artisan key:generate"
```
gilbertsoft, line 253:
```suggestion
            ddev composer create --no-install --prefer-dist statamic/statamic -y
            ddev composer run-script post-root-package-install
            ddev composer install
            ddev composer run-script post-create-project-cmd
            ddev php please make:user
```

gilbertsoft, line 253:
```suggestion
            ddev composer create --no-install --prefer-dist statamic/statamic -y
            ddev composer run-script post-root-package-install
            ddev composer install
            ddev composer run-script post-create-project-cmd
            ddev php please make:user
```

rpkoller:
```suggestion
            ddev composer create --prefer-dist statamic/statamic -y
            ddev composer run-script post-create-project-cmd
            ddev php please make:user --super
```

rpkoller: I would leave out the --no-install flag so that the user doesn't have to run the `composer install` since the pull request introduces convenience of not needing the --no-install flag anymore. so one command less to enter.  and the other two i've picked from your previous suggestion. in regards of the make:user `--super` i haven't found that something like that existed. that also saves my previous question in regards of the comment. that isn't necessary anymore that way.

gilbertsoft: Will come up with a better solution, was only playing around here.

rfay:
I appreciate the creativity here and look forward to looking at this more, but you should know that `ddev composer create` has a long history of incremental changes, and there are lots of ways it can go bad in various contexts (including NFS). So I really wouldn't be willing to bring this in before the upcoming release.

gilbertsoft:
> I appreciate the creativity here and look forward to looking at this more, but you should know that `ddev composer create` has a long history of incremental changes, and there are lots of ways it can go bad in various contexts (including NFS). So I really wouldn't be willing to bring this in before the upcoming release.

No hurry to get it in. Has to be tested for all project types. Thanks to @rpkoller who makes a great job in testing :)

But this is a big step for `ddev composer create` and will save us time in implementing project type specific workarounds!

rfay:
Manually tested craftcms without the --no-scripts and it worked fine.

mattstein (approved):

rfay:
Rebased.

rfay:
Please make sure when working on this that `ddev composer create` errors out if no args are provided.

rfay:
I moved this to v1.22.5, assuming we can do it with no disruption to people.

rfay:
Rebased, fixing conflicts. Hopefully correctly.

rfay:
The test failure here in TestConfigSetValues is due to `ddev config --docroot=web` being actually overwritten by the discovered docroot, for example `public` in https://github.com/ddev/ddev/actions/runs/6668917097/job/18125558518?pr=5058

So the test has actually caught a real problem with the code (or the rebase).

hanoii:
after working on #5499 @rfay asked me to review this.. I took a more thorough look and I now understand most of it. It obviously needs to be rebased and probably adapted a bit if my PR gets in... 

Thoughts that I have for now:

> There may be rare cases where plugins and scripts should be run, therefor the new flag --preserve-flags

I wonder what those rare cases are, do we know any?
Also, instead of adding such a flag I wonder if something more like --with-scripts and --with-plugins is more obvious.

Also, while I could test and understand the need for `--no-scripts` like on craftcms, why are we also adding --no-plugins. Is there an example of this.

I would also like to ask, and I just saw #4273, the slowness that trigger that improvements, was it because of the shared folder or something else? Would that still be an issue with mutagen. Isn't running composer install in the root directory after rsyncing what `composer create` creates a similar impact of performance of what an rsync do afterwards? Not trying to undo the whole work, but i like the idea of ddev doing as little as possible so to keep compatibility with the composer create command as much as I can so I am trying to understand that need.

Finally, a crazy idea, not for this PR but just sharing it: I was looking a bit at the code on composer and I saw this: https://github.com/composer/composer/blob/da83d29d8ac634783bc70b0f3b7847bae3e081d9/src/Composer/Command/CreateProjectCommand.php#L232

I am wondering if we can have a global composer plugin of our own (ddev owned) that could hook into this and do the initial rsync and then continue whatever composer create needs to do on a new dir. Obviously this would fail if `--no-plugins` is actually needed for any reason or bypass by the user, so we should have a fallback behavior as well.

rfay:
> and I just saw https://github.com/ddev/ddev/pull/4273, the slowness that trigger that improvements, was it because of the shared folder or something else?

Both Colima (with sshfs) and NFS often fail to sync properly with massive file changes, and of course the changes are really slow with NFS. Mutagen solves this problem, but using a separate directory that's not a mount and then copying has worked far better.

rfay:
Rebased, hope I didn't make any mistakes. 

This now is based on 
* #5499 from @hanoii 

So it will need testing and careful review. I hope we can recover it, as it's important work.

rpkoller:
is it already testable again at this point or is some more work needed? cuz i have to install a few different composer based projects tonight to document how the handle a certain functionality. and it would be a good way testing this patch as well?

rfay:
Please try simple testing, artifacts in https://github.com/ddev/ddev/pull/5058#issuecomment-1616077076 seem to have been updated properly. We'll see if it passes all automated tests also.

rfay:
- [ ] TestConfigSetValues is failing
- [ ] @rpkoller points out that `ddev composer create` with no argument should be disallowed.

rfay:
TestConfigSetValues fails because of the addition of typo3ConfigOverrideAction:

https://github.com/gilbertsoft/ddev/blob/fbd8b3e7045a13c6743ee0344a1f52da93b945d5/pkg/ddevapp/typo3.go#L168-L182

rfay:
* Fixed so you have to have exactly one arg to `ddev composer create`
* Removed typo3ConfigOverrideAction. Maybe we'll find out why he did that now. Seems rather major that it wired the docroot.

GuySartorelli:
> There may be rare cases where plugins and scripts should be run, therefor the new flag --preserve-flags is introduced to fallback to the old behavior and not add these two options. But normally it should be fine to running scripts and plugins later with composer install.

It seems like you could add `--no-install` as part of the `composer create-project` call, and then run `composer install` afterwards. That should ensure that the end result is the same as running `composer create-project` with scripts allowed in most cases, without requiring users to run install separately.

This Pr is meant to resolve this problem:
> Users do not know about the special behavior of `ddev composer create` and the differences to `composer create-project`

Users wouldn't normally need to run `composer install` _after_ running `composer create-project`, so requiring that now will just change how that lack of knowledge trips people up rather than directly alleviating the problem.

rfay:
I think @ultimike has generally proposed the `create --no-install` followed by `install`, it's simpler, easy to understand, and easy to debug.

cmd/ddev/cmd/composer-create.go
```diff
@@ -8,21 +8,20 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
-var composerCreateYesFlag bool
-
 // ComposerCreateCmd handles ddev composer create
 var ComposerCreateCmd = &cobra.Command{
-	Use: "create [args] [flags]",
+	Use:  "create [args] [flags]",
+	Args: cobra.MinimumNArgs(1),
```
stasadev, line 24: `cobra.MinimumNArgs(1)` works for the following input

```
ddev composer create laravel/laravel --prefer-dist
```

But it does not work for
```
ddev composer create --prefer-dist laravel/laravel
```

Because cobra reads it as:
```
ddev composer create --prefer-dist=laravel/laravel
```

Perhaps something like the code below can check `osargs` for the real arg:

```go
hasAtLeastOneArg := false
for _, element := range osargs {
	if !strings.HasPrefix(element, "--") && !strings.HasPrefix(element, "-") {
		hasAtLeastOneArg = true
		break
	}
}

if !hasAtLeastOneArg {
	util.Failed("failed to create project: provide a package name to install")
}
```

rfay: I saw the same problem with exactArgs(1). Thought this would resolve it. Really odd that the flags have to come after the args.

stasadev: Yes, passing flags and arguments to another command complicates things. I looked inside cobra and didn't find anything suitable. That's because when you pass unknown flags, cobra doesn't know if that flag takes any value and treats the next item in the cmd as the value for that flag.

rfay:
@GuySartorelli Would you be willing to spend a little time with composer create here? It fails TestComposerCreateAutocomplete due to interplay between the intention here and your intentions. It would be much appreciated! If you see an obvious solution just push it into a branch and I'll cherry-pick it.

GuySartorelli:
> @GuySartorelli Would you be willing to spend a little time with composer create here? It [fails TestComposerCreateAutocomplete](https://github.com/ddev/ddev/actions/runs/7847667020/job/21417102186?pr=5058) due to interplay between the intention here and your intentions. It would be much appreciated! If you see an obvious solution just push it into a branch and I'll cherry-pick it.

This PR is reverting a change that I noted as necessary in my PR. See https://github.com/ddev/ddev/pull/5756#discussion_r1473974649

I haven't looked too closely at this PR or the discussion here, but a quick text search for the word "flag" doesn't come up with any discussion that indicates why this was being reverted. If reverting it isn't necessary, I'd recommend _not_ reverting that change.

rfay:
Thanks @GuySartorelli - I set it back to the way you had it, passes the test locally now. Will hope to get your full review on this once it's passing tests.

Note that this predated your work by 8 months, just got stalled although it got rave reviews. Hard to get something as stale as this in, but we're trying.

rfay:
Yay, it finally passed tests and perhaps is resurrected. If all of you could review now it would be so fantastic.

rpkoller (approved):
Mandatory changes for the quickstart.md were already done in the previous round of review. I don't see anything else necessary at this point. And I've manually retested the current state of the PR creating new sites with the instructions from the quickstart.md.  Created drupal, craft, grav, laravel, moodle, shopware, silverstripe, statamic, typo3 and kirby sites.Sll that have a composer create quickstart guide. Everything ran through without a problem.

GuySartorelli:
## Flag name
`preserve-flags` is a bit of a confusing name for the new flag, which (if I'm not mistaken) is really just a way to say "Don't add `--no-scripts` and `--no-plugins`". It seems like maybe this should be replaced with two scripts: `--with-scripts` and `--with-plugins`, so it's really clear what the flag does.
`preserve-flags` suggests to me "DDEV will ignore some of your flags, so you need to use `--preserve-flags` to have it respect them" which is obviously not the case.

## Behavioural change
I still think don't really understand what this PR is solving - but based on some of the comments in here I still stand by my comments in https://github.com/ddev/ddev/pull/5058#issuecomment-1928594749 that it might make sense to have the `composer create-project` use `--no-install`, and then run a `composer install` in the project dir (with scripts and plugins allowed) afterward.

Utimately, the behaviour I'd expect of `ddev composer create` would be that it does exactly the same thing as if I ran `composer create-project` in an empty directory. If that wasn't true before, and this PR makes that true, then yay! If this PR moves in the opposite direction, that would seem concerning to me.

rfay:
> I still think don't really understand what this PR is solving - but based on some of the comments in here I still stand by my comments in https://github.com/ddev/ddev/pull/5058#issuecomment-1928594749 that it might make sense to have the composer create-project use --no-install, and then run a composer install in the project dir (with scripts and plugins allowed) afterward.

The `composer create-project` here does in fact use --no-install, then do a second pass with --install. Please give it a try and see what it does. 

> Utimately, the behaviour I'd expect of ddev composer create would be that it does exactly the same thing as if I ran composer create-project in an empty directory. If that wasn't true before, and this PR makes that true, then yay! If this PR moves in the opposite direction, that would seem concerning to me.

That's what @gilbertsoft was trying to accomplish here. (If it were in fact possible to run composer create-project in an empty directory, which it's not)


This is in fact a difficult journey, because it was @gilbertsoft 's initiative, and I was a little reluctant about it. However, @rpkoller has been very supportive, and the idea seems good. Maybe we'll be successful, maybe not.

stasadev (approved):
This looks good, and I would postpone this PR, because of the changes in the docs.
People who are looking at the latest docs and using the stable DDEV won't be able to use the quickstart.

rfay:
Good point.  The flip side is that this will be out of sync with the quickstarts before it gets pulled, but we can survive that.

I sure hope there's nothing bad about this. 

But we also need experience with this in the meantime, so it's a catch-22

rpkoller:
> People who are looking at the latest docs and using the stable DDEV won't be able to use the quickstart.
Out of curiosity, cuz i haven't considered that point as a potential problem yet, are there many people using the current stable version of DDEV with the latest version of the docs on ddev.readthedocs.io?  I wouldn't have expected that this is the case cuz, at least me myself tend to use the docs that come along or are on par with the version i currently use?

rfay:
@rpkoller so often people use the `latest` version and get confused. I think we should figure out a better way to do this, and maybe have another branch for breaking-docs-changes.

Often we *want* to update the docs for current users. Sometimes (like here) we don't want to until they have it in their hands.

Most people don't know there are two versions, or know what the two versions mean. 

Would love to have you think about this (in a new issue if we don't have one). We've talked about it many times.

GuySartorelli:
> I think we should figure out a better way to do this, and maybe have another branch for breaking-docs-changes.

Probably worth opening a discussion thread about this.

rfay:
@GuySartorelli 
* https://github.com/orgs/ddev/discussions/5832

rfay:
Rebased

PR #5072, opened 2023-07-04
refactor: detect more dependency changes

gilbertsoft:
## The Issue

We like to detect dependency changes in the CI to properly label PRs changing dependencies. The required Docker Compose version is currently located in the global config and is not taken into account

This PR moves the Docker Compose version to the version constants. To avoid import cycles the functions from version constants are moved to a new package docker/images.

Changes in the version constants are now properly label with `dependencies`.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5072"></a>

### DIFF
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -45,7 +45,10 @@ labels:
     sync: true
     matcher:
       files:
-        any: ["go.*", "vendor/**"]
+        any:
+          - "pkg/versionconstants/**"
+          - "go.*"
+          - "vendor/**"
 
 checks:
   - context: "Semantic Pull Request"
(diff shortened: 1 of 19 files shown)


rfay (approved):
This refactor works for me; looks like it's trying two different things though, changing the labeler and doing the refactor of Get*Image(). Would probably be better to do those separately, but seems ok.

gilbertsoft:
> This refactor works for me; looks like it's trying two different things though, changing the labeler and doing the refactor of Get*Image(). Would probably be better to do those separately, but seems ok.

Yes, that's why I have renamed it finally to refactor. But topic is anyway the same.

PR #5078, opened 2023-07-05
ci: fix PR build for forks [skip ci]

gilbertsoft:
PR builds should use the AMPLITUDE_API_KEY_DEV secret to build the
executables which currently does not work. To be able to use secrets in
PRs which are created from fork repos pull_request_target must be used
instead of pull_request.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5078"></a>

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,11 +1,7 @@
 name: PR Build
 
-defaults:
-  run:
-    shell: bash
-
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - ".github/workflows/**"
       - "cmd/**"
@@ -15,12 +11,15 @@ on:
       - "go.*"
       - "Makefile"
 
+defaults:
+  run:
+    shell: bash
+  
 env:
   BUILDKIT_PROGRESS: plain
   DOCKER_CLI_EXPERIMENTAL: enabled
   DDEV_DEBUG: true
   HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
 
 permissions:
   contents: read
@@ -37,6 +36,7 @@ jobs:
           # We need to get all branches and tags for git describe to work properly
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
+
       - name: Set up Homebrew
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
IIRC, using pull_request_target means that *head* code will be used and we'll build HEAD instead of the PR.

gilbertsoft:
> IIRC, using pull_request_target means that _head_ code will be used and we'll build HEAD instead of the PR.

Please have a look at the checkout action, the PR's sha is used there.

rfay (commented):
It looks like this PR got some refactoring from another PR mixed into it (versionconstants.GetWebImage(), etc.)

rfay (changes requested):
I'm fine to go with this if you've tested it on your fork (and a fork-of-your-fork) in advance. 

We need to think about security as well with pull_request_target

Doesn't doing this mean that any evil PR can expose the api key? They could just embed it in plain text into the build, and then they own it (and so does everybody else)? They could get access to any other environment variables as well, right?

I think we should talk about that first.

I'm actually fine with PR builds not having amplitude support, if that's the only motivation here.

rfay:
From https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

> TL;DR: Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

rfay:
I don't think there's a secure way to build untrusted code and then provide a binary. It's just too easy to steal all the secrets.

gilbertsoft:
> I don't think there's a secure way to build untrusted code and then provide a binary. It's just too easy to steal all the secrets.

There are currently two possible attack vectors in this workflow which must be addressed first:
- ./.github/workflows/linux-build-setup.sh
- make

Both should be checked out from the default branch. So we need two checkouts currently which will make it safe!

gilbertsoft:
The build worked so far https://github.com/gilbertsoft/ddev/actions/runs/5467078860?pr=33

rfay:
All code which is built using the upstream workflow can capture any environment variable it's built with.

rfay:
One thing we can do here is to not automatically build forked PRs. You and I could start using the main repo for branches, or we could have all forked PRs require approval for this test. I don't think that's hard to do.

gilbertsoft:
> One thing we can do here is to not automatically build forked PRs. You and I could start using the main repo for branches, or we could have all forked PRs require approval for this test. I don't think that's hard to do.

Yes, also thought about this and already had a look at our repo settings therefor. At the moment we have to run jobs for new contributors.

gilbertsoft:
Had a look at the various options we have, GH does not have the possibility to restrict a workflow run like discussed. But we can easily achieve it by a label. So we create a label `build` e.g. and once we manually assign the label to a PR the PR Build workflow will run.

rfay:
On just the topic of instrumentation and PR builds, I would rather just let them not have instrumentation than fiddle with our build process.

PR #5079, opened 2023-07-05
ci: decrease cache update interval to have more up to date caches [skip ci]

gilbertsoft:


rfay (commented):
This will then update the cache every time it runs (if more than a minute)? Is that what you want? I thought you were trying to prevent that.

gilbertsoft:
> This will then update the cache every time it runs (if more than a minute)? Is that what you want? I thought you were trying to prevent that.

This means every run on master will update the cache which is fine. I initially tried to avoid too many updates but maximum time for the cache transfer for the biggest cache is less than 10 seconds, so does not hurt to update the caches on every run.

gilbertsoft:
Changed again after our discussion about the usage of the assets. Now pull-push-test-platforms is prefered to store the cache because all assets are downloaded in theses tests.

rfay (approved):
I'm certainly willing to try it!

PR #5080, opened 2023-07-05
ci: implement clean up and update of Homebrew cache [skip ci]

gilbertsoft:


.github/workflows/colima-tests.yml
```diff
@@ -102,14 +111,17 @@ jobs:
...
           key: ${{ runner.os }}-homebrew-cache-${{ steps.get-date.outputs.date }}
           restore-keys: |
             ${{ runner.os }}-homebrew-cache-
 
-      - name: Set up Homebrew
-        id: set-up-homebrew
-        uses: Homebrew/actions/setup-homebrew@master
+      - name: Clean up Homebrew
+        run: |
+          brew update
```
rfay, line 121: `brew update` is done automatically by `brew upgrade` these days, but this does no harm.

gilbertsoft: Did not find any hint on that in the docs. Only cleanup is mentioned. https://docs.brew.sh/Manpage#upgrade-options-outdated_formulaoutdated_cask-

rfay (approved):
Still fine with doing this. It is also very low risk, affects only colima, etc.

PR #5084, opened 2023-07-05
fix: quote host parameter for HeidiSQL and prevent argument double quoting

andreashager:
## The Issue

When launching HeidiSQL via `ddev heidisql` there is an additional double quote at the end of each database connection name. If you remove the double quotes from the arguments the host parameter must be written in double quotes as the documentation suggests it (Docu: https://www.heidisql.com/help.php#commandline).

This PR removes the double quotes at the end of the string and adds them to the host parameter.

Run`ddev heidisql` to start HeidiSQL from WSL2 with correct name.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
@@ -7,15 +7,15 @@
 ## OSTypes: windows,wsl2
 ## HostBinaryExists: /mnt/c/Program Files/HeidiSQL/heidisql.exe,C:\Program Files\HeidiSQL\Heidisql.exe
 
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for the contribution!

We would have reopened this for you, you didn't have to delete it, just needed the title adjusted :) 

From the other PR:
> Last I tested this it worked OK. What is the symptom that you encountered?

andreashager:
OK didn't know that 🙂

The command itself worked fine but in the connection manager of HeidiSQL shows an additional double quote at the end (have a look at the screenshot). And this happens because of the quotation of $arguments than HeidiSQL adds the last double quote to the database description name. 

Connection-Manager

Additionally I found at that the host parameter need to be embedded in double quotes according to documentation of HeidiSQL.

rfay (approved):
Seems fine to me, thanks!

gilbertsoft (approved):
Thanks! Works here on traditional Windows and WSL 2

PR #5088, opened 2023-07-05
ci: reduce Homebrew cleanup

gilbertsoft:
brew upgrade makes too much noise and takes too long, skip for now.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5088"></a>


gilbertsoft:
Don't skip ci for this one, we like to have a full run to see if Colima works now.

rfay (changes requested):
It's a mistake that we're not running tests on this. We must have something configured wrong.

rfay:
Oh, it's because the skip ci on the commit itself.

rfay:
Let's remove the skip ci (do another commit) and at least see the tests start to work.

PR #5091, opened 2023-07-06
ci: increase cache update interval to one week

gilbertsoft:
The previous tests confirmed the problem with the missing assets in the cache was related to tests not downloading all assets. The ddev-test-cache is only stored on pull-push tests where all assets are downloaded, other tests only read this cache. Therefor the cache storing interval is increased again to one week.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5091"></a>

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -62,26 +62,15 @@ jobs:
       - name: Get Date
         id: get-date
         run: |
-          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
         shell: bash
-
-      - name: DDEV test cache
-        uses: actions/cache@v3
-        if: ${{ github.ref == 'refs/heads/master' }}
-        with:
-          path: ~/.ddev/testcache/tarballs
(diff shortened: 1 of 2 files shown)


.github/workflows/colima-tests.yml
```diff
@@ -108,7 +97,7 @@ jobs:
 
       - name: Homebrew cache/restore
         uses: actions/cache/restore@v3
-        if: ${{ github.ref != 'refs/heads/master' }}
+        if: github.ref != 'refs/heads/master'
```
rfay, line 100: I'm always baffled by the mixed use of golang templates and direct comparisons in Github actions.

rfay (approved):
I haven't tested or looked at the results but I'm fine with this if you are.

.github/workflows/colima-tests.yml
gilbertsoft: Don't get it, these are JS conditions. `${{ }}` is the so called expression syntax which will be replaced by the runner before running the workflow. But as `if` is always assumed to be a condition the expression syntax is superfluous but works for compatibility.

rfay: It's actually a `$` with golang template.

rfay: At least that's what I think!

PR #5094, opened 2023-07-06
fix: create upload dirs before generate Mutagen config, fixes #5093

### LINKED ISSUE #5093: v1.22.0-beta2: Mutagen not configured correctly for upload_dirs

```
```
</details>         

### Expected Behavior

Mutagen should ignore upload_dirs and not try to sync them.

### Actual Behavior

```
 Network ddev-d9_default  Created
 Container ddev-d9-db  Created
 Container ddev-d9-web  Created
 Container ddev-d9-db  Started
 Container ddev-d9-web  Started
Starting mutagen sync process... This can take some time.
Mutagen sync completed with problems in 3s.
For details on sync status 'ddev mutagen st d9 -l'
```

In this case (Drupal9, no customized upload_dirs) the .ddev/mutagen/mutagen.yml has *not* been updated with the proper exclusions for upload_dirs

### Steps To Reproduce

Just `ddev config` a typical Drupal 9 site and `ddev start`, should get this result

Make sure to watch for "Mutagen sync completed with problems in 3s" and check `ddev mutagen st -l`

### Anything else?

_No response_

gilbertsoft:

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -989,6 +989,8 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		}
 	}
 
+	app.createUploadDirsIfNecessary()
+
 	if app.IsMutagenEnabled() {
 		err = app.GenerateMutagenYml()
 		if err != nil {
@@ -1156,8 +1158,6 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 
 	// Fix any obsolete things like old shell commands, etc.
 	app.FixObsolete()
-
-	app.createUploadDirsIfNecessary()
 
 	// WriteConfig .ddev-docker-compose-*.yaml
 	err = app.WriteDockerComposeYAML()


pkg/ddevapp/mutagen_test.go
```diff
@@ -63,6 +63,10 @@ func TestMutagenSimple(t *testing.T) {
 
 	assert.True(dockerutil.VolumeExists(ddevapp.GetMutagenVolumeName(app)))
 
+	mutagenConfig, err := os.ReadFile(ddevapp.GetMutagenConfigFilePath(app))
+	require.NoError(t, err)
+	require.Contains(t, mutagenConfig, app.GetUploadDir())
```
rfay, line 68: Shouldn't it check for each item in app.GetUploadDirs()?

rfay (changes requested):
On a quick look it looks right, I'll manually test. As in comment, I think we should check for all upload_dirs in the test. 

It seems TestMutagenSimple usually tests on drupal9, which has upload dir for sure, so that should be OK.

pkg/ddevapp/mutagen_test.go
gilbertsoft: Also thought about that but as there is only one item anyway I decided to go this simpler path.

rfay:
It will save time and testing resources to run the test manually before pushing it up, thanks.

rfay:
I fixed the test and made it check all upload_dirs, even though there's only ever one right now. Slightly less efficient the way I did it (if more than one dir) but it's in a test.... oh well.

rfay (approved):
I manually tested this to verify that it behaves. I used `ddev delete -Oy && rm -rf web vendor composer.lock && ddev start` to demonstrate on upstream/master; it demonstrated the problem on start with a mutagen warning. With this patch it works and the proper value is in the mutagen.yml. Thanks for the quick fix!

rfay:
I think the failure of the no-bind-mounts test is telling us that there shouldn't be any exclusions on no-bind-mounts, because we don't have any bind mounts!

gilbertsoft:
And I've pushed a small fix for the failures we had here during the clean up see https://github.com/ddev/ddev/pull/5097

PR #5096, opened 2023-07-06
fix: terminate mutagen sessions on DDEV upgrade, fixes #5040

### LINKED ISSUE #5040: DDEV upgrade cleanup: flush and remove existing mutagen sessions

1. Existing mutagen sessions with no labels have to be removed, because we need sessions with labels. So the upgrade should probably just clean up all existing mutagen sessions that have no label. Or maybe all sessions in general?

### Describe your solution

Above

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

* #5040

When DDEV upgrade is detected, after poweroff, destroy mutagen sessions. These should already have been flushed when they were stopped.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5096"></a>

### DIFF
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -787,9 +787,11 @@ func GetMutagenConfigFileHashLabel(app *DdevApp) (string, error) {
 
 // TerminateAllMutagenSync terminates all sync sessions
 func TerminateAllMutagenSync() {
-	out, err := exec.RunHostCommand(globalconfig.GetMutagenPath(), "sync", "terminate", "-a")
-	if err != nil {
-		util.Warning("could not terminate all mutagen sessions (mutagen sync terminate -a), output=%s, err=%v", out, err)
+	if fileutil.FileExists(globalconfig.GetMutagenPath()) {
+		out, err := exec.RunHostCommand(globalconfig.GetMutagenPath(), "sync", "terminate", "-a")
+		if err != nil {
+			util.Warning("could not terminate all mutagen sessions (mutagen sync terminate -a), output=%s, err=%v", out, err)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/mutagen.go
```diff
@@ -729,11 +729,11 @@ func CheckMutagenVolumeSyncCompatibility(app *DdevApp) (ok bool, volumeExists bo
 	if mutagenSyncExists {
 		mutagenLabel, mutagenSyncLabelErr = GetMutagenSyncLabel(app)
 		if mutagenSyncLabelErr != nil {
-			util.Warning("mutagen sync %s exists but unable to get label: %v", app.Name, mutagenSyncLabelErr)
+			util.Warning("mutagen sync session '%s' exists but unable to get sync label '%s': '%v' This is normal on upgrade from v1.21.6; error=%v", app.Name, mutagenSignatureLabelName, mutagenLabel, mutagenSyncLabelErr)
```
gilbertsoft, line 732: Why do we need `mutagenSignatureLabelName` here? It's anyway a constant.

gilbertsoft (approved):
Looks good, just one question.

pkg/ddevapp/mutagen.go
rfay: The only reason is to make debugging it easier. I'm concerned about the various mutagen failures we're hearing about.

gilbertsoft: But does this constant change from time to time? It's a constant, so normally not interesting seeing it in a debug output. But anyway not a blocker, just don't get it why this could be useful.

rfay: It doesn't change, but you'd have to look at the code to figure out what was missing to use the proper mutagen command to see the labels. It's just about making it easy to debug, should we need to do so.

PR #5097, opened 2023-07-07
ci: prevent job from failing if cleanup fails

gilbertsoft:
## The Issue

Cleanup may fail on upstream problems see https://github.com/ddev/ddev/actions/runs/5482145525/jobs/9987195573?pr=5094

Make clean up optional and don't fail anymore.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5097"></a>

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -188,6 +188,7 @@ jobs:
           colima stop
 
       - name: Clean up Homebrew
+        continue-on-error: true
         run: |
           brew update
           brew autoremove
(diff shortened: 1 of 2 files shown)


rfay (approved):
This seems fine. Might be clearer and better with `brew autoremove || true` etc. Then we're relying on explicit "don't worry about it" commands rather than relying on Github's directives. Both ways should be fine.

gilbertsoft:
> This seems fine. Might be clearer and better with `brew autoremove || true` etc. Then we're relying on explicit "don't worry about it" commands rather than relying on Github's directives. Both ways should be fine.

`|| true` is a bad idea because it hides the failure but we like to continue on failures but still see if failures occur which also can mean we have to optimize the workflow again.

PR #4553, opened 2023-01-13
WSL2: Improve hosts management on Windows side, fixes #4482

rfay:
## The Issue

* `ddev hostname -r` and thus `ddev delete` were always escalating to hostname removal on Windows side whether or not the hostname was there (so doing it on `*.ddev.site` for example)
* `ddev hostname` was not allowed to run if the `docker` client wasn't installed on the Windows side.
* #4482 

## How This PR Solves The Issue

* Complete rework of hostname management, with special care for WSL
* Make the removal happen only after a check to see if it exists.
* Allow ddev hostname to run with no docker client
* Improve docs.
* Adjust scripts and docs as in #4482 

## Manual Testing Instructions

- [x] `ddev delete -Oy` a project in WSL2, verify it does right thing with hosts.
- [x] Make sure `ddev.exe hostname -r junk.ddev.site` doesn't escalate to remove
- [x] Make sure `ddev start` with project_tld set to something odd works predictably
- [x] Make sure that `ddev delete` in that case works predictably
- [x] Test `ddev hostname --remove-inactive` on all platforms
- [x] Review docs changes at https://ddev--4553.org.readthedocs.build/en/4553/, especially see [docs files diffs](https://github.com/drud/ddev/pull/4553/files?file-filters%5B%5D=.md&show-deleted-files=true&show-viewed-files=true). The reorganized and consolidated Windows install docs are at https://ddev--4553.org.readthedocs.build/en/4553/users/install/ddev-installation/#windows
- [x] Install with docker-ce script on clean machine with clean wsl2
- [x] Install with Docker Desktop script on clean wsl2

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/4553"></a>

PR #5098, opened 2023-07-07
build: clean up Windows installer, fixes #1945

### LINKED ISSUE #1945: Windows Installer/Uninstaller should stop running processes (winnfsd service) before trying to manipulate nfsd
Some processes may be running on starting the uninstaller as nssm or nfsd. The common problem is the service named "nfsd" which is running winnfsd.exe, and is installed by a script provided with ddev.

The uninstaller should stop these processes/services before starting the removal of this files.

gilbertsoft:
## The Issue

The language strings were accidentially removed with https://github.com/ddev/ddev/pull/4553 and some other parts were not properly removed.

Also fixes 
* #1945 

The missing language strings are added again and left overs are removed properly.

Check build if there are no warnings. When running the installer, on the component page the descriptions should be shown properly for all components now.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5098"></a>

### DIFF
--- a/winpkg/ddev.nsi
+++ b/winpkg/ddev.nsi
@@ -507,14 +507,14 @@ SectionGroupEnd
 /**
  * WinNFSd group
  */
-SectionGroup /e "WinNFSd"
+SectionGroup /e "WinNFSd (deprecated)"
   /**
    * WinNFSd application install
    */
-  Section "${WINNFSD_NAME}" SecWinNFSd
+  Section /o "${WINNFSD_NAME}" SecWinNFSd
     SectionIn 1
     SetOutPath "$INSTDIR"
-    SetOverwrite try
+    SetOverwrite off
 
     ; Copy files
     File "licenses\winnfsd_license.txt"
@@ -542,7 +542,7 @@ SectionGroup /e "WinNFSd"
   /**
    * NSSM application install
    */
-  Section "${NSSM_NAME}" SecNSSM
+  Section /o "${NSSM_NAME}" SecNSSM
     ; Install in non choco mode only
     ${IfNot} ${Chocolatey}
       SectionIn 1
--- a/winpkg/ddev.nsi
+++ b/winpkg/ddev.nsi
@@ -226,10 +226,6 @@ Caption "${PRODUCT_NAME_FULL} ${PRODUCT_VERSION} $InstallerModeCaption"
 !insertmacro MUI_PAGE_LICENSE "..\.gotmp\bin\windows_amd64\sudo_license.txt"
 
 ; Components page
-!ifdef DOCKER_NSH
-  Var DockerVisible
-  Var DockerSelected
-!endif
 Var MkcertSetup
 !define MUI_PAGE_CUSTOMFUNCTION_PRE ComponentsPre
 !insertmacro MUI_PAGE_COMPONENTS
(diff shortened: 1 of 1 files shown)


gilbertsoft:
@rfay just saw the strings are missing because of your removal in https://github.com/ddev/ddev/pull/4553. As far as I can see you started to remove the installation of Docker with the Windows Installer but there are still plenty of things to remove. I guess the removal of the Docker install was intented though?

rfay:
Yes, removal of Docker Desktop install was intentional... most people are using docker-ce now, and it just confuses everything if they install DD.

gilbertsoft:
> Yes, removal of Docker Desktop install was intentional... most people are using docker-ce now, and it just confuses everything if they install DD.

With my initial implementation you can simply provide DOCKER_EXCLUDE to Nsis and Docker won't be included. This would make it possible to provide two installers, one with Docker included, the other without.

Now we have to decide which way we'll go....

rfay:
Mostly we don't want people using traditional Windows. Mostly we don't really want people using WSL with Docker Desktop. Mostly, the only reason to install ddev.exe is so it can update unusual entries to the hosts file on Windows. So we never want Docker Desktop installed unless people explicitly want it.

winpkg/ddev.nsi
```diff
@@ -659,7 +659,11 @@ SectionEnd
  */
 LangString DESC_SecDDEV ${LANG_ENGLISH} "Install ${PRODUCT_NAME_FULL} (required)"
 LangString DESC_SecAddToPath ${LANG_ENGLISH} "Add the ${PRODUCT_NAME} (and sudo) directory to the global PATH"
-LangString DESC_SecMkcert ${LANG_ENGLISH} "mkcert (github.com/ FiloSottile/mkcert) is a simple tool for making locally-trusted development certificates. It requires no configuration"
+!ifdef DOCKER_NSH
+LangString DESC_SecDocker ${LANG_ENGLISH} "Download and install ${DOCKER_DESKTOP_NAME} (www.docker.com) which does not seem to be installed, but is required for $(^Name) to function"
+!endif ; DOCKER_NSH
```
rfay, line 664:
```suggestion
```

rfay (changes requested):
Thanks, just don't want Docker Desktop in there.

winpkg/ddev.nsi
```diff
@@ -659,7 +616,8 @@ SectionEnd
  */
 LangString DESC_SecDDEV ${LANG_ENGLISH} "Install ${PRODUCT_NAME_FULL} (required)"
 LangString DESC_SecAddToPath ${LANG_ENGLISH} "Add the ${PRODUCT_NAME} (and sudo) directory to the global PATH"
-LangString DESC_SecMkcert ${LANG_ENGLISH} "mkcert (github.com/ FiloSottile/mkcert) is a simple tool for making locally-trusted development certificates. It requires no configuration"
+LangString DESC_SecSudo ${LANG_ENGLISH} "Sudo for Windows (github.com/mattn/sudo) allows for elevated privileges which are used to add hostnames to the Windows hosts file (required)"
```
rfay, line 619: We haven't been using mattn/sudo for ages. We use gsudo.

rfay (changes requested):
Looks like it's coming along just fine, just needs manual testing and correct sudo pointer.

winpkg/ddev.nsi
gilbertsoft: Done

winpkg/ddev.nsi
```diff
@@ -616,10 +616,10 @@ SectionEnd
  */
 LangString DESC_SecDDEV ${LANG_ENGLISH} "Install ${PRODUCT_NAME_FULL} (required)"
 LangString DESC_SecAddToPath ${LANG_ENGLISH} "Add the ${PRODUCT_NAME} (and sudo) directory to the global PATH"
-LangString DESC_SecSudo ${LANG_ENGLISH} "Sudo for Windows (github.com/mattn/sudo) allows for elevated privileges which are used to add hostnames to the Windows hosts file (required)"
+LangString DESC_SecSudo ${LANG_ENGLISH} "Sudo for Windows (github.com/ddev/gsudo) allows for elevated privileges which are used to add hostnames to the Windows hosts file (required)"
```
rfay, line 619: Sorry, I think we're always using https://github.com/gerardog/gsudo - I think that's the right reference here.

But per https://github.com/ddev/gsudo/releases/tag/v0.7.3, the reason we're bundling an old one here is because he doesn't publish a raw binary. So we'd have to unbundle it. I guess.  I requested this in https://github.com/gerardog/gsudo/issues/88, he wasn't willing.

rfay: Is it obvious to you how we could download and unbundle the binary during the build process? We could certainly do it in the Makefile.

rfay: Is it obvious to you how we could download and unbundle the binary during the build process? We could certainly do it in the Makefile.

rfay: Is it obvious to you how we could download and unbundle the binary during the build process? We could certainly do it in the Makefile.

winpkg/ddev.nsi
```diff
@@ -616,10 +616,10 @@ SectionEnd
...
 LangString DESC_SecMkcert ${LANG_ENGLISH} "mkcert (github.com/FiloSottile/mkcert) is a simple tool for making locally-trusted development certificates. It requires no configuration"
 LangString DESC_SecMkcertSetup ${LANG_ENGLISH} "Run `mkcert -install` to setup a local CA"
-LangString DESC_SecWinNFSd ${LANG_ENGLISH} "WinNFSd (github.com/ winnfsd/winnfsd) is an optional NFS server that can be used with ${PRODUCT_NAME_FULL}"
+LangString DESC_SecWinNFSd ${LANG_ENGLISH} "WinNFSd (github.com/winnfsd/winnfsd) is an optional NFS server that can be used with ${PRODUCT_NAME_FULL}"
```
rfay, line 622: Obviously we should deprecate this soon. It will be so good when windows users don't need to set up winnfsd at all.

rfay (approved):
This is fine, really, and we didn't intend to work on this. It would be nice to have if:
* Use upstream gsudo (unbundle it)
* Disable winnfsd by default in the dialog. This would also take care of https://github.com/ddev/ddev/issues/1945 in a super-simple way.

We can go forward with this as you like, either as is or with a bit more, your call. 

I manually tested on traditional Windows and it worked great... and frankly ddev worked great. It's quite amazing how well it works these days.

rfay (approved):
I took the liberty of making WinNFSD optional (with the help of ChatGPT4) and changed the named source of gsudo. This is good, I'd love to extend it someday, use it for installing WSL2, etc. but this is actually a step forward, and eliminates the failure to update WinNFSD which has happened to every windows user on every install forever.

PR #5108, opened 2023-07-08
fix: detect Craft CMS version and warn user on type mismatch and introduce new env vars, fixes #4650

### LINKED ISSUE #4650: Don't manage Craft 3 .env, Fix `*_SITE_URL`

gilbertsoft:
## The Issue

- #4650

`CRAFT_WEB_URL` and `CRAFT_WEB_ROOT` are added to the `.env` file. `PRIMARY_SITE_URL` is still left untouched to not break existing installations at the moment. Additionally a version check is implemented and warns the user if using project type `craftcms` for version before 4.

Follow the quick start to setup a new Craft CMS site which should not work out-of-the-box without any additional configuration.

Setup a Craft CMS v3 or earlier with project type `craftcms` and check if a warning is properly shown.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5108"></a>

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/ddev/ddev/pkg/archive"
 	"github.com/ddev/ddev/pkg/fileutil"
@@ -68,6 +69,14 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		return nil
 	}
 
+	// Check version is v4 or higher or warn user about app type mismatch.
+	if !isCraftCms4orHigher(app) {
+		util.Warning("It looks like the installed Craft CMS is lower than version 4 where it's recommended to use project type `php` or disable settings management with `ddev config --disable-settings-management`")
+		if !util.Confirm("Would you like to continue anyway with the automatic configuration?") {
+			return nil
+		}
+	}
+
 	// If the .env file doesn't exist, try to create it by copying .env.example to .env
 	envFilePath := filepath.Join(app.AppRoot, app.ComposerRoot, ".env")
 	if !fileutil.FileExists(envFilePath) {
@@ -104,6 +113,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 
 	// If they have older version of .env with DB_DRIVER, DB_SERVER etc, use those
 	if _, ok := envMap["DB_SERVER"]; ok {
+		// TODO remove, was never an official standard of Craft CMS.
 		envMap = map[string]string{
 			"DB_DRIVER":             driver,
 			"DB_SERVER":             "db",
@@ -124,9 +134,11 @@ func craftCmsPostStartAction(app *DdevApp) error {
 			"CRAFT_DB_DATABASE":     "db",
 			"CRAFT_DB_USER":         "db",
 			"CRAFT_DB_PASSWORD":     "db",
+			"CRAFT_WEB_URL":         app.GetPrimaryURL(),
+			"CRAFT_WEB_ROOT":        app.GetAbsDocroot(true),
 			"MAILHOG_SMTP_HOSTNAME": "127.0.0.1",
 			"MAILHOG_SMTP_PORT":     "1025",
-			"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
+			"PRIMARY_SITE_URL":      app.GetPrimaryURL(), // for backward compatibility only
 		}
 	}
 
@@ -155,4 +167,27 @@ func craftCmsConfigOverrideAction(app *DdevApp) error {
 	app.PHPVersion = nodeps.PHP81
 	app.Database = DatabaseDesc{nodeps.MySQL, nodeps.MySQL80}
 	return nil
+}
+
+// isCraftCms4orHigher returns true if the Craft CMS version is 4 or higher. The
+// proper detection will fail if the vendor folder location is changed in the
+// composer.json.
+// The detection is based on a change starting with 4.0.0-RC1 where deprecated
+// constants were removed in src/Craft.php see
+// https://github.com/craftcms/cms/commit/1660ff90a3a69cec425271d47ade66523a4bd44e#diff-21e22a30e7c48265a4dcedc1b1c8b9372eca5d3fdeff6d72c7d9c6b671365c56
+func isCraftCms4orHigher(app *DdevApp) bool {
+	craftFilePath := filepath.Join(app.GetComposerRoot(false, false), "vendor", "craftcms", "cms", "src", "Craft.php")
+	if !fileutil.FileExists(craftFilePath) {
+		// Sources are not installed, assuming v4 or higher.
+		return true
+	}
+
+	craftFileContent, err := fileutil.ReadFileIntoString(craftFilePath)
+	if err != nil {
+		util.Warning("unable to read file `%s` in project `%s`: %v", craftFilePath, app.Name, err)
+
+		return true
+	}
+
+	return !regexp.MustCompile(`const\s+Personal\s*=\s*0`).MatchString(craftFileContent)
 }
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -72,7 +72,7 @@ func craftCmsPostStartAction(app *DdevApp) error {
 	// Check version is v4 or higher or warn user about app type mismatch.
 	if !isCraftCms4orHigher(app) {
 		util.Warning("It looks like the installed Craft CMS is lower than version 4 where it's recommended to use project type `php` or disable settings management with `ddev config --disable-settings-management`")
-		if !util.Confirm("Would you like to continue anyway with the automatic configuration?") {
+		if !util.Confirm("Would you like to stop here, not do the automatic configuration and change project type?") {
 			return nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -389,22 +389,28 @@ func (app DdevApp) GetDocroot() string {
...
+}
+
+// GetComposerRoot will determine the absolute composer root directory where
+// all Composer related commands will be executed.
+// If inContainer set to true, the absolute path in the container will be
+// returned, else the absolute path on the host.
+// If showWarning set to true, a warning containing the composer root will be
+// shown to the user to avoid confusion.
+func (app *DdevApp) GetComposerRoot(inContainer, showWarning bool) string {
+	absComposerRoot := path.Join(app.GetAbsDocroot(inContainer), app.ComposerRoot)
```
rfay, line 413: This logic isn't correct; The composer_root is not relative to the docroot. So this ends up with the wrong things in a basic craft3 download, where the composer.json is in the project root (where it should be)

rfay (changes requested):
This doesn't work yet; the logic for composer root isn't correct; You can test by downloading a Craft 3 tarball (vendor and composer.json in root, docroot in web) from https://craftcms.com/docs/3.x/installation.html#downloading-an-archive-file-manually

pkg/ddevapp/ddevapp.go
gilbertsoft: Oh it is correct, it has nothing changed here, but the name of the new function is wrong, should be the absolute project root instead.

gilbertsoft: True, messed up sth with the refactoring.

pkg/ddevapp/ddevapp.go
```diff
@@ -389,22 +389,28 @@ func (app DdevApp) GetDocroot() string {
...
+	var basePath string
 
 	if inContainer {
 		basePath = "/var/www/html"
 	} else {
 		basePath = app.AppRoot
 	}
 
-	absComposerRoot := path.Join(basePath, app.ComposerRoot)
+	return path.Join(basePath, app.GetDocroot())
```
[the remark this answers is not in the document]
gilbertsoft: Yes, would be probably better. But so far it seems to work also with path on host. Will change...

rfay (changes requested):
This can't work until 

* https://github.com/ddev/ddev/pull/5128

```
$ ddev start
You seem to have a new DDEV version.
During an upgrade it's important to `ddev poweroff`.
May I do `ddev poweroff` before continuing?
This does no harm and loses no data. [Y/n] (yes):
panic: interface conversion: interface {} is bool, not ddevapp.UploadDirs

goroutine 1 [running]:
github.com/ddev/ddev/pkg/ddevapp.(*DdevApp).validateUploadDirs(0xc000177500)
        /home/rfay/workspace/ddev/pkg/ddevapp/upload_dirs.go:230 +0x485
github.com/ddev/ddev/pkg/ddevapp.(*DdevApp).LoadConfigYamlFile(0xc000177500, {0xc00002a840?, 0xc00056ef70?})
        /home/rfay/workspace/ddev/pkg/ddevapp/config.go:356 +0x1b2
github.com/ddev/ddev/pkg/ddevapp.(*DdevApp).ReadConfig(0xc000177500, 0x1)
        /home/rfay/workspace/ddev/pkg/ddevapp/config.go:310 +0x4b
github.com/ddev/ddev/pkg/ddevapp.NewApp({0xc0000292f0, 0x18}, 0x97?)
        /home/rfay/workspace/ddev/pkg/ddevapp/config.go:112 +0x7fc
github.com/ddev/ddev/pkg/ddevapp.GetProjects(0x1)
        /home/rfay/workspace/ddev/pkg/ddevapp/utils.go:343 +0x1e7
github.com/ddev/ddev/pkg/ddevapp.PowerOff()
        /home/rfay/workspace/ddev/pkg/ddevapp/poweroff.go:11 +0x33
github.com/ddev/ddev/cmd/ddev/cmd.checkDdevVersionAndOptInInstrumentation(0x0)
        /home/rfay/workspace/ddev/cmd/ddev/cmd/root.go:225 +0x2ce
github.com/ddev/ddev/cmd/ddev/cmd.glob..func57(0x10efb20?, {0x1130d80, 0x0, 0x0})
        /home/rfay/workspace/ddev/cmd/ddev/cmd/start.go:73 +0x335
github.com/spf13/cobra.(*Command).execute(0x10efb20, {0x1130d80, 0x0, 0x0})
        /home/rfay/workspace/ddev/vendor/github.com/spf13/cobra/command.go:860 +0x663
github.com/spf13/cobra.(*Command).ExecuteC(0x10ee4a0)
        /home/rfay/workspace/ddev/vendor/github.com/spf13/cobra/command.go:974 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        /home/rfay/workspace/ddev/vendor/github.com/spf13/cobra/command.go:902
github.com/ddev/ddev/cmd/ddev/cmd.Execute()
        /home/rfay/workspace/ddev/cmd/ddev/cmd/root.go:155 +0x25
main.main()
        /home/rfay/workspace/ddev/cmd/ddev/main.go:28 +0xc5
```

pkg/ddevapp/craftcms.go
```diff
@@ -68,6 +69,14 @@ func craftCmsPostStartAction(app *DdevApp) error {
 		return nil
 	}
 
+	// Check version is v4 or higher or warn user about app type mismatch.
+	if !isCraftCms4orHigher(app) {
+		util.Warning("It looks like the installed Craft CMS is lower than version 4 where it's recommended to use project type `php` or disable settings management with `ddev config --disable-settings-management`")
+		if !util.Confirm("Would you like to continue anyway with the automatic configuration?") {
```
rfay, line 75: I sure wish we had an option to offer "no" as the default. Should we do that?

rfay (commented):
I did experiment with squashing #5128 onto this and it seemed to do what we want with Craft 3.

pkg/ddevapp/craftcms.go
```diff
@@ -124,9 +134,11 @@ func craftCmsPostStartAction(app *DdevApp) error {
 			"CRAFT_DB_DATABASE":     "db",
 			"CRAFT_DB_USER":         "db",
 			"CRAFT_DB_PASSWORD":     "db",
+			"CRAFT_WEB_URL":         app.GetPrimaryURL(),
+			"CRAFT_WEB_ROOT":        app.GetAbsDocroot(true),
 			"MAILHOG_SMTP_HOSTNAME": "127.0.0.1",
 			"MAILHOG_SMTP_PORT":     "1025",
-			"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
+			"PRIMARY_SITE_URL":      app.GetPrimaryURL(), // for backward compatibility only
```
rfay, line 141: I thought the decision in https://github.com/ddev/ddev/issues/4650 was not to include this.

pkg/ddevapp/craftcms.go
gilbertsoft: Or rephrase the question to invert it.

pkg/ddevapp/craftcms.go
gilbertsoft: Removal is breaking and imho we should give the users some time to use the new config.

rfay: I think this would only affect new projects.

pkg/ddevapp/craftcms.go
rfay: That's not a bad idea.

pkg/ddevapp/ddevapp.go
```diff
@@ -389,23 +389,41 @@ func (app DdevApp) GetDocroot() string {
...
+	if inContainer {
+		return path.Join(app.GetAbsAppRoot(true), app.GetDocroot())
+	}
+
+	return filepath.Join(app.GetAbsAppRoot(false), app.GetDocroot())
+}
+
+// GetAbsAppRoot returns the absolute path to the project root on the host or if
+// inContainer is set to true in the container.
+func (app DdevApp) GetAbsAppRoot(inContainer bool) string {
```
rfay, line 404: I am very happy to have this in there :)

rfay (approved):
Let's do it!

PR #5110, opened 2023-07-09
refactor: anonymize user-defined custom commands

gilbertsoft:
## The Issue

User-defined custom commands are exposed to instrumentation which is not what we want regarding data protection.

User-defined custom commands are now anonymized with the name `custom-command`. So we still are able to record statistics about the usage of custom commands but have not longer critical data stored in Amplitude.

Run user-defined custom commands and check on Amplitude if they are properly anonymized. This is best done with the user lookup debug technique where all commands are visible.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5110"></a>

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"bufio"
 	"fmt"
-	"github.com/ddev/ddev/pkg/nodeps"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


rfay (commented):
Thanks for remembering this. This is going to be a good approach.

rfay (approved):
Looks perfect and quite elegant to me, thanks. Stepped through and verified behavior.

PR #5117, opened 2023-07-10
fix: empty global table style should revert to "default"

rfay:
## The Issue

Many v1.21.6 projects had `table_style: ""`

v1.22.0 requires `table_style: default`

Disallow "" as value, then the global config reader substitutes "default"

Set `table_style: ""` in global_config.yaml, then `ddev start`

You should now see `table_style: default` in global_config.yaml

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5117"></a>

### DIFF
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -3,7 +3,6 @@ package globalconfig
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


gilbertsoft (approved):
Looks good! Empty value is set to default while validating the global config.

PR #5119, opened 2023-07-11
feat: add Silverstripe project type

Firesphere:
## The Issue

Silverstripe has some requirements that a default `ddev config` does not ask. Instead of being forced to use `ddev config --php-version=8.1 --webserver-type=apache-fpm --project-type=php`, using an easier, shorter command is useful

It adds Silverstripe as a project type, making it possible to start a project with only the project type.

It further pre-creates a standard Silverstripe `.env` file for the database settings and environment type.

See the added quickstart.md commands

## Automated Testing Overview

~I'm not sure if tests are needed, I didn't see anything for Laravel, which I used as an example.~

~On top of that, t~ This is my first Go I've ever done, so I'm not too familiar with all the features and options yet.

Got tests going :)

## Release/Deployment Notes

The PHP type project still works as usual for Silverstripe

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5119"></a>


rfay:
Ambitious! This would have been better discussed in an issue first, as we haven't had many folks asking around for it, and it adds complexity to the code. The biggest thing missing from the PR right now is tests, see top of ddevapp_test.go for the existing tests.

Firesphere:
> Ambitious! This would have been better discussed in an issue first, as we haven't had many folks asking around for it, and it adds complexity to the code.

I realise an issue would have been better, in hindsight. I enjoyed trying to get it to work a bit too much.

Silverstripe is a framework that's largely used in New Zealand (it being NZ based and all), but also Poland, UK, Netherlands, Germany and Canada are known to use it.

> The biggest thing missing from the PR right now is tests, see top of ddevapp_test.go for the existing tests.

Yep, learning GoTesting now!

docs/content/users/quickstart.md
```diff
@@ -375,6 +375,32 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
+
+        Use a new or existing Composer project, or clone a Git repository.
+
+        === "Composer"
+            ```bash
+            mkdir my-silverstripe-app
+            cd my-silverstripe-app
+            ddev config --project-type=silverstripe --docroot=public --create-docroot
+            ddev composer create --prefer-dist --no-install --no-scripts silverstripe/installer -y
+            ddev composer install
```
gilbertsoft, line 392:
```suggestion
            ddev composer create --prefer-dist --no-scripts silverstripe/installer -y
```
This workaround is not needed anymore

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ddev/ddev/pkg/fileutil"
+	"github.com/ddev/ddev/pkg/nodeps"
+	"github.com/ddev/ddev/pkg/util"
+	"os"
+	"path/filepath"
+)
+
+// isSilverstripeApp returns true if the app is of type Silverstripe
+func isSilverstripeApp(app *DdevApp) bool {
+	return fileutil.FileExists(filepath.Join(app.AppRoot, "vendor/bin/sake"))
```
gilbertsoft, line 14:
```suggestion
	return fileutil.FileExists(filepath.Join(app.AppRoot, app.ComposerRoot, "vendor/bin/sake"))
```
Problem with this detection is that if somebody changes the vendor or the bin dir in the `composer.json` this will not work. Please add a TODO comment so we are able to fix it with #5058 where a composer.json parser is introduced to fix issues like this.

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func silverstripePostStartAction(app *DdevApp) error {
+	// We won't touch env if disable_settings_management: true
+	if app.DisableSettingsManagement {
+		return nil
+	}
+	envFilePath := filepath.Join(app.AppRoot, ".env")
```
gilbertsoft, line 22:
```suggestion
	envFilePath := filepath.Join(app.AppRoot, app.ComposerRoot, ".env")
```

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,67 @@
...
+	_, envText, err := ReadProjectEnvFile(envFilePath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("Unable to read .env file: %v", err)
+	}
+	if os.IsNotExist(err) {
+		err = fileutil.CopyFile(filepath.Join(app.AppRoot, ".env.example"), filepath.Join(app.AppRoot, ".env"))
```
gilbertsoft, line 28:
```suggestion
		err = fileutil.CopyFile(filepath.Join(app.AppRoot, app.ComposerRoot, ".env.example"), filepath.Join(app.AppRoot, app.ComposerRoot, ".env"))
```

gilbertsoft (changes requested):
Thanks! Added some minor change requests.

docs/content/users/quickstart.md
Firesphere: Oh? Nice!
It always kinda bugged me that it was this two-step method. I took it pretty much one-on-one from the Laravel docs, so maybe all the quickstarts need to be looked at and simplified?

pkg/ddevapp/silverstripe.go
Firesphere: I'm wondering how realistic is it that someone would do such a thing?

Not that it shouldn't be included, just curious, I never heard anyone actually do that ;)

Firesphere:
@gilbertsoft I've applied all your suggestions.

I'm going through and getting tests to go/work now. So far, so good, I... think?

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,67 @@
...
+			return err
+		}
+	}
+	port := "3306"
+	dbConnection := "MySQLDatabase"
+	if app.Database.Type == nodeps.Postgres {
+		dbConnection = "PostgreSQLDatabase"
+		port = "5432"
+	}
+	envMap := map[string]string{
```
rpkoller, line 44: I wonder would it make sense to add values for`SS_DEFAULT_ADMIN_USERNAME` and `SS_DEFAULT_ADMIN_PASSWORD` as well to the to be created `.env` file? I am in no way an Silverstripe expert but it looked the one or two times I've installed it recently that initially there is no user created by Silverstripe out of the box and the quickstart guide also  recommends setting the credentials for the default user: https://docs.silverstripe.org/en/5/getting_started/

rpkoller (commented):
Added one comment in regards of the `.env` file in silverstripe.go

rfay:
Please try to run the key tests locally before pushing. For example, you'll if you're adding silverstripe as number 17 in https://github.com/ddev/ddev/blob/729471081a245d83fb7e8c48d61ee424e4a51219/pkg/ddevapp/ddevapp_test.go#L41, you'll do 

GOTEST_SHORT=17 make testpkg TESTARGS="-run TestDdevFullSiteSetup"

Or, of course, run it that way in your IDE. 

I recommend that you temporarily create your own test-silverstripe repo in your own account like we do with https://github.com/ddev/test-laravel, and make a release that contains the necessary artifacts. Then we can promote that repo to ddev later.

Firesphere:
> GOTEST_SHORT=17 make testpkg TESTARGS="-run TestDdevFullSiteSetup"
Or, of course, run it that way in your IDE.
I recommend that you temporarily create your own test-silverstripe repo in your own account like we do with https://github.com/ddev/test-laravel, and make a release that contains the necessary artifacts. Then we can promote that repo to ddev later.


This has been what I've been doing locally, before pushing

I've put my artifacts at https://github.com/Firesphere/asset-store
I can somewhat easily make a full repo out of it, but I'm not sure if that's needed?

mattstein (approved):

pkg/ddevapp/silverstripe.go
Firesphere: I did think about that too. Given it's .env, and for local development, it's probably good to do so. 👍

pkg/ddevapp/ddevapp_test.go
```diff
@@ -222,6 +222,19 @@ var (
 			Safe200URIWithExpectation:     testcommon.URIWithExpect{URI: "/README.txt", Expect: "This is a simple readme."},
 			DynamicURI:                    testcommon.URIWithExpect{URI: "/index.php", Expect: "This program makes use of the Zend Scripting Language Engine"},
 		},
+		// 12: Silverstripe
```
rfay, line 225: Please don't inject this here and break the knowledge of all things that come after it, please put it as a new entry at the end of the list, thanks. This will be 17.

Firesphere: :+1:
I went alphabetically :)

Firesphere: And I see I even failed at basic alphabet! 😆

pkg/ddevapp/silverstripe.go
rpkoller: cool! if there would be another way or a username and pw preset i haven't found that would also totally fine. but when i've skimmed through the docs to setup silverstripe i couldnt find anything, the getting started and it's instructions for the .env file were the only guidance. 
and from a new users perspective for someone who is unfamiliar with a CMS such a quick start is super convenient to get around potential stepping stones. that is something i really appreciate from any CMS documentation.

p.s. and thanks for working on that PR! was evaluating and comparing many CMSes recently, silverstripe was one of them, and this project type would have saved me some head scratches :D

Firesphere:
It seems like the tests are breaking due to the tar.gz not being an actual tar.gz.
I've tested this multiple times, re-uploaded it, and it is correct to me.

Can anyone tell me what's wrong with my "standard first set-up" database data?

By the way, whether this will be accepted or not, I am definitely enjoying getting my hands dirty on Go! I can see why people like it!

pkg/ddevapp/silverstripe.go
Firesphere: Thanks :)

If you have any questions, feel free to flick me an email via my address here on GitHub :) I do believe I know a thing or two about Silverstripe... Since... I kinda worked with it for 10 years and have worked at the actual company for 4 years too, in which I rewrote the whole Security implementation, among things :)

rfay: I'm sure impressed! I'm sure you'll be successful.

rfay:
Maybe the tarball has a problem, or maybe there was a download problem, https://github.com/ddev/ddev/actions/runs/5537560997/jobs/10106519571?pr=5119

Firesphere:
> Maybe the tarball has a problem, or maybe there was a download problem, https://github.com/ddev/ddev/actions/runs/5537560997/jobs/10106519571?pr=5119


I've run it locally in a few different ways, on Linux (KDE Neon 22.04 and Kubuntu 22.04) and on my Macbook (MacOS Ventura), and all went through.

Redid the same thing after re-uploading a new file, but I keep getting an error from Silverstripe. Which, by the way, is an improvement, as it's surely not an actual error in the test anymore, but something in the set-up causing the db not being instantiated properly(?)

rfay:
In the meantime I re-ran the test.

Firesphere:
Okay, this was a very dumb mistake on my side.

The problem was my own global `.ddev/global_config.yaml` has a standard DB setting, hence, I forgot to include the configuration required in the `.env` in my tar.gz...

I've fixed that up, removed the static file test (Silverstripe doesn't do static files that way), and changed the homepage check to match on the Generator name instead of a generic bit of the content.

It is now locally passing on Kubuntu 22.04, KDE Neon DEV and MacOS Ventura.

rfay:
global_config.yaml doesn't have any database settings...

Firesphere:
Not usually, no, but because of the amount of Silverstripe based projects, I have this in my own global config:
```
web_environment: 
  - SS_DATABASE_NAME=db
  - SS_DATABASE_SERVER=db
  - SS_DATABASE_USERNAME=db
  - SS_DATABASE_PASSWORD=db
  - SS_ENVIRONMENT_TYPE=dev
```

Which removes my need for adding an initial `.env` file

Firesphere:
I may be looking wrong, but I don't see any failure on the buildkite build, to be errorring due to any changes I made here.

On another note, is it normal that my "standard" ddev installation from apt, re-downloads `docker-compose` every time after I've run tests locally?

I assume it's expected, and the ddev build downloads a newer version than the apt installed version, causing a mismatch and redownload?

rfay:
I restarted it, the WSL builder crashed.

rfay:
> On another note, is it normal that my "standard" ddev installation from apt, re-downloads docker-compose every time after I've run tests locally?

The latest DDEV has a newer docker-compose. If you change your local to use the latest beta that won't happen to you any more. Unfortunately, we don't have a good way to provide pre-releases via apt.

Anyway, every time you switch between this (HEAD, essentially) and your local, it says "Oh, I need that other version". It's annoying. It will be over soon when we get the release out. Almost there.

pkg/ddevapp/silverstripe.go
Firesphere: I've added the todo, by the way :)

I'm interested to see how it's going to be resolved

Firesphere:
A question around the PR process:
- Do you want me to squash the commits in to a single commit?
- Should I rebase, even if there is no merge/rebase conflict?
- Is `[skip ci]` in the commit message not honoured? (it seems not, to me?)

Just making sure I'm doing stuff right :)

Cheers.

rfay:
> Do you want me to squash the commits in to a single commit?
 
No, don't squash. When we pull, we do it with a squash. In the meantime it's best to have all the commits. 

> Should I rebase, even if there is no merge/rebase conflict?

With what you're doing the code is all separate from what's changing underneath it, so doesn't matter much. But it doesn't hurt to rebase when you're pushing new stuff anyway.

> Is [skip ci] in the commit message not honoured? (it seems not, to me?)

[skip ci] on the *last* commit your pushing is honored. But we don't run the tests for each commit, just for each push.

Firesphere:
By the way, given how much I personally work with the Silverstripe framework, and am absolutely loving DDEV, I'm happy to keep the Silverstripe part of DDEV under my wing, so if there are any problems with this specific integration, it would be on me to solve it primarily.

rfay:
Yes, your maintenance of DDEV wrt Silverstripe will be very much appreciated. Making an initial contribution is one thing, but keeping it up over the years is quite another thing. Thanks!

pkg/ddevapp/silverstripe.go
Firesphere: Setting aside potential typos, @rpkoller , I think I've added it properly now. :)

docs/content/users/quickstart.md
```diff
@@ -375,6 +375,31 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
+            ddev composer create --prefer-dist --no-scripts silverstripe/installer -y
+            ddev start
+            ```
+        === "Git Clone"
+            ```bash
+            git clone <your-silverstripe-repo>
+            cd <your-silverstripe-project>
+            ddev config --project-type=silverstripe --docroot=public --create-docroot
+            ddev start
+            ddev composer install
```
rpkoller, line 400: I would suggest to add a `ddev launch /admin` at the end of the `composer` and `git clone` snippets which would be in line with the other quickstarter guides.

docs/content/users/quickstart.md
```diff
@@ -375,6 +375,31 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
+            ```
+
```
rpkoller, line 402: Since the changes that are applied in the .env file are not user facing and could go unnoticed i would suggest to explicitly name the username and password you are able to log into silverstripe on the login page that was launched with `ddev launch /admin`. that way it would be clear to the user how to proceed. with the env changes a default user is in place now but still it would be not directly clear to the user what the necessary credentials are and that those are already added to the .env file. i would suggest something like similar to moodle: 

```
Log into the Silverstripe site you just installed using `admin` and `password`. Visit [Learn Silverstripe](https://www.silverstripe.org/learn/) for more information.

File: pkg/ddevapp/silverstripe.go (hunk 5 above)
Comment (rpkoller) [reply pkg/ddevapp/silverstripe.go:46]: oh awesome! i've left two follow-up comments in the context of the quickstarter.md now that the default username and password are set. one is about informing the user what the set username and password actually (which is inline with the example used for Moodle). plus i've suggested to add `ddev launch /admin`at the end. so that the person installing the CMS is directly forwarded to the url where it is possible to log into the admin interface since all other recipes use that.

File: docs/content/users/quickstart.md [hunk 9]
```diff
@@ -375,6 +375,31 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
...
+        Use a new or existing Composer project, or clone a Git repository.
+
+        === "Composer"
+            ```bash
+            mkdir my-silverstripe-app
+            cd my-silverstripe-app
+            ddev config --project-type=silverstripe --docroot=public --create-docroot
+            ddev composer create --prefer-dist --no-scripts silverstripe/installer -y
+            ddev start
```
rpkoller, line 392: I've downloaded the current artifact of the PR and manually tested the step. Beofre the `ddev launch /admin` another step is missing and should be added to the quickstarted guides:
`ddev exec vendor/bin/sake dev/build`

without the cmd the user runs into an error when trying to launch the site because the database isn't populated yet.

rpkoller:
@rfay not directly related to this issue, but is it correct that [apptypes_test.go](https://github.com/ddev/ddev/pull/5119#diff-002123950b86806ba84f0fd117ae68de752570017706d387fd3f205e2a2a646d) contains only a subset of project types? somehow django4, shoptware6, typo3  are missing there? and in the context of this issue would it make sense to switch the position of line 75 and 76, that way all listed apptypes would be sorted alphabetically? but i suppose currently it does no harm.

docs/content/users/quickstart.md
rfay: Just a note that `vendor/bin` is in the web container $PATH already, so `ddev exec sake` should work.

rfay:
@rpkoller it would be great to add the missing types. That's a pretty low-priority test compared to TestDdevFullSiteSetup.

docs/content/users/quickstart.md
rpkoller: I've quickly tested and it works to just use `ddev exec sake dev/build`. it is tempting to shorten the command. but on second thought i think it might be a good idea to keep the relational path. that way it is easier reproducible for someone to understand where the sake command resides within the project directory inside the container. for me `ddev exec vendor/bin/sake dev/build` is already cryptic for example but having `ddev exec sake dev/build` instead feels even more like voodoo imho.

rpkoller:
> @rpkoller it would be great to add the missing types. That's a pretty low-priority test compared to TestDdevFullSiteSetup.

i could give it a try after the quickstarter draft is finished or could be also an task for one of the contribution training group?

docs/content/users/quickstart.md
rfay: I like people to know that $PATH works and that vendor/bin is in the path. That's how drush works, for example. You certainly don't have to `ddev exec vendor/bin/drush` - true of many of the other quickstarts here.

rfay: And $PATH isn't voodoo :)

rpkoller: but the difference with drush is that is something that you install in the quickstarter while sake ships with silverstripe. but basically you are right. maybe the better approach might be using the short form and explain in the liner notes for the quickstart guide where for example is located within the project. 

and with voodoo i just meant when you read the instructions and see something unusual like sake or /vendor/bin/sake the latter makes it a bit more graspable. ;)

Firesphere:
> @rfay not directly related to this issue, but is it correct that [apptypes_test.go](https://github.com/ddev/ddev/pull/5119#diff-002123950b86806ba84f0fd117ae68de752570017706d387fd3f205e2a2a646d) contains only a subset of project types? somehow `django4`, `shopware6`, `typo3` are missing there? and in the context of this issue would it make sense to switch the position of line 75 and 76, that way all listed apptypes would be sorted alphabetically? but i suppose currently it does no harm.

Are you sure they're missing? django, shopware6 and typo3 are definitely in there, just not alphabetical. After `php`, the list is not alphabetical anymore...

docs/content/users/quickstart.md
Firesphere: Note that a `sake` shorthand is included in this PR.

I've added some of your comments around it to the quickstart.

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"github.com/ddev/ddev/pkg/fileutil"
+	"github.com/ddev/ddev/pkg/nodeps"
+	"github.com/ddev/ddev/pkg/util"
+	"os"
+	"path/filepath"
+)
+
+// TODO Check if location is still correct when composer bin folder is changed.
```
rfay, line 12: Has this TODO been taken care of?

rfay (changes requested):
This is looking great, thanks for the excellent effort. 

I've given you access to https://github.com/ddev/test-silverstripe

Please 

1. Create a repo there (probably based on silverstripe upstream) 
2. Add a README or something out of the way of the actual silverstripe code that explains how to create the proper artifacts for use in the tests here. 
3. Create a release with the proper artifacts
4. Add a commit to use the new artifacts

As time goes on and artifacts get outdated it can be really hard to figure out how to update them, so that's why we want to do this.  

I think we can probably get this into v1.22.0.

pkg/ddevapp/silverstripe.go
Firesphere: No, it is added as per @gilbertsoft his comment higher up, where he asked to add this, so it would be taken care of when the composer bin folder is taken from the composer.json file instead of hardcoded `vendor/bin/sake`.

Firesphere: I believe that with the suggested changes of including the composer root in the `filepath.join()` calls, it is resolved, but I'd like confirmation of that from @gilbertsoft , before I remove the todo.

rfay: You can test manually too... And GetComposerRoot() should help. Note that you will use `filepath.Join()` when operating on the host, and `path.Join()` when doing things inside the container.

rfay: > so it would be taken care of when the composer bin folder is taken from the composer.json file instead of hardcoded vendor/bin/sake

There is no plan for the composer.json to be parsed or used for the composer bin directory; the way things currently are (`composer_root` determines where we look for vendor, and vendor/bin is automatically added to the PATH) is what I'd expect. No changes in how this is being handled are contemplated.

rfay: Read the comment, see what you're saying. I'm not as much a fan as Simon about adding TODOs, would rather fix things when the future hoped-for thing goes in.

Firesphere: @rfay Just to be clear... when you say "Simon"... do you mean me or @gilbertsoft ... because... my name is Simon.... and from a quick web search... it seems I'm not the only one with that name!

I'm not a big fan of leaving in TODOs. I'm going to remove it, as with the suggested changes from @gilbertsoft , it seems to work okay. If anything breaks, I will revisit when it breaks, instead of now. I'm somewhat assuming I'll notice a pre-release will let me know something is wrong.

rfay: Sorry! You're Firesphere and @gilbertsoft is Simon. I didn't want to ping him :)

Firesphere: All good... a bit of confusion on my side. Simon is too common a name 😆 

I'm happy to stick to being called "Firesphere" 👍 

Just getting the artifact repo up and ready. I'd like to run some tests before I do the final push and make it all "okay". Just to make sure the correct artifacts are being pulled in and all.

I'll update again when I get it all done. Probably tomorrow morning NZ time.

Firesphere:
@rfay I've pushed to the repo, added instructions on how to create the artifacts, created a release, and added the artifacts, in the repo you gave me access to.

I've ran tests locally on all the environments I've mentioned above. I _think_ that is all, at the moment?

docs/content/users/quickstart.md
Firesphere: added a note about it.

I'm not a fan of including it inside the command list, as it's not relevant to the actual set-up itself.

rfay:
Here's your latest commit, https://github.com/ddev/ddev/pull/5119/commits/2ffa9ff224d16e6f80b415786e2407c9fea97007

image


You'll see that all the important things were skipped, and if you look closely you'll see that the buildkite builds ran but skipped themselves. 

The best thing to do is to push several commits so that a single push goes up for a single set of tests. Where we're at right now is we haven't seen tests run against this PR lately, so let's go ahead and do that. `git commit -m "dummy commit to run tests" --allow-empty` maybe. Or if you have more work to do, do it and push without [skip ci]. [skip ci] is for when things haven't changed, they've already been tested, etc.

rfay:
Rebased and pushed a dummy commit to run tests

rfay:
db.tar.gz isn't a tarball:

```
$ curl -LO https://github.com/ddev/test-silverstripe/releases/download/1.0.0/db.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  5241  100  5241    0     0   4616      0  0:00:01  0:00:01 --:--:--  4616
rfay@rfay-xps13:~/tmp$ tar -tzf db.tar.gz
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
```

Firesphere:
Re-uploaded. I didn't get around to it yesterday, sorry about that.

rfay:
Thanks, restarted the test.

rfay:
Succeeded, yay!

Firesphere:
I tried to run the `TestCustomCommands` test, but it seems to hang on pulling the FS layers for Docker...

Is this a known issue I overlooked?

docs/content/users/usage/commands.md
rpkoller, line 1: I've noticed one other minor detail. since a global command for sake is being added with this PR `pkg/ddevapp/global_dotddev_assets/commands/web/sake` wouldn't it make sense to add the command to the commands.md as well?

rfay: Good point, yes `sake` should be added to the commands.md

rfay (changes requested):
I tried the quickstart but when I got to the key part it failed miserably:

```
rfay@rfay-xps13:~/workspace/silverstripe$ ddev sake dev/build flush=all
ERROR [UNKNOWN TYPE, ERRNO 2002]: Uncaught mysqli_sql_exception: No such file or directory
IN GET dev/build
Line 120 in /var/www/html/vendor/silverstripe/framework/src/ORM/Connect/MySQLiConnector.php

```

It appears that the .env file isn't created correctly, it's pointing at localhost. Here's what was in my .env at this point:
```
# DB credentials
SS_DATABASE_CLASS="MySQLDatabase"
SS_DATABASE_SERVER="localhost"
SS_DATABASE_USERNAME="db"
SS_DATABASE_PASSWORD="db"
SS_DATABASE_NAME="db"

# WARNING: in a live environment, change this to "live" instead of dev
SS_ENVIRONMENT_TYPE="dev"
SS_DATABASE_TYPE=MySQLDatabase
MAILER_DSN=smtp://localhost:1025
SS_DEFAULT_ADMIN_USERNAME=admin
SS_DEFAULT_ADMIN_PASSWORD=password
SS_DATABASE_HOST=db
SS_DATABASE_PORT=3306
```


Also, the tabs in the docs aren't behaving correctly, you'll probably have to run the local `make mkdocs_serve` to fiddle with it and get it correct.  "Composer" and "Git install" are supposed to be subtabs, not titles.

https://ddev--5119.org.readthedocs.build/en/5119/users/quickstart/#silverstripe

image

Firesphere:
I think that error is simply because I'm not a smart person, and had named the environment variables wrong.

E.g.
`SS_DATABASE_SERVER="localhost"`
versus
`SS_DATABASE_HOST=db`

Same root cause as for the tests... I have my defaults in the global config set... 🤦 

I fixed up your errors, and the docs, which was an indentation error

rfay:
You pushed your last commit with [skip ci], so maybe you didn't get any build done at all, but the doc certainly isn't right yet, https://ddev--5119.org.readthedocs.build/en/5119/users/quickstart/#silverstripe

For now, I'd say don't do [skip ci], as it seems to just keep getting in your way.

Please make sure to manually test at every step along the way. We don't want people to try things out the first time and get big errors. 

Can you take a look and try to figure out why the automated tests didn't catch the failure to create a valid .env? We must not be doing something right.

I recommend that you use a "clean" global_config.yaml for a while... you've tripped on that a few times.

Firesphere:
Yeps, I'm looking at that.

I've cleared my global ddev config. The reason it didn't trip the tests, is because the .env in the repository is _correct_.

Thank you for your patience!

rfay:
> The reason it didn't trip the tests, is because the .env in the repository is correct.

IMO the test-silverstripe repo should not include a .env...

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	envMap := map[string]string{
+		"SS_DATABASE_SERVER":        "db",
+		"SS_DATABASE_PORT":          port,
+		"SS_DATABASE_NAME":          "db",
+		"SS_DATABASE_USERNAME":      "db",
+		"SS_DATABASE_PASSWORD":      "db",
+		"SS_ENVIRONMENT_TYPE":       "dev",
+		"SS_DATABASE_CLASS":         dbConnection,
+		"MAILER_DSN":                "smtp://localhost:1025",
```
rpkoller, line 55: where does mails_dsn come from? noticed it one or two times, same as ss_database_host before, but never questioned it. but i've went to the docs page listing the available environmental variables https://docs.silverstripe.org/en/4/getting_started/environment_management/#core-environment-variables and mails_dsn isnt listed there? would it make sense to remove it as well?

pkg/ddevapp/silverstripe.go
```diff
@@ -0,0 +1,74 @@
...
+		"SS_DEFAULT_ADMIN_USERNAME": "admin",
+		"SS_DEFAULT_ADMIN_PASSWORD": "password",
+	}
```
rpkoller, line 58: and would it make sense to add ss_base_url and set it to the DDEV_PROJECT and DDEV_TLD? so that you get `SS_BASE_URL="http://silverstripe.ddev.site"`?

pkg/ddevapp/silverstripe.go
Firesphere: It's for Silverstripe 5:
https://docs.silverstripe.org/en/5/developer_guides/email

pkg/ddevapp/silverstripe.go
Firesphere: That's only for if it can't detect it. Adding it has caused a myriad of problems with HTTPS settings not being applied properly, and if multiple domains are used (Silverstripe Subsites), it creates a lot of "random" issues with the domains added/used.

It's suggested for using CLI commands, but even then, it's not necessary most of the time.

docs/content/users/quickstart.md
```diff
@@ -375,6 +375,44 @@ While the generic `php` project type is [ready to go](./project.md) with any CMS
 
     For more advanced tasks like adding elasticsearch, building and watching storefront and administration, see [susi.dev](https://susi.dev/ddev-shopware-6).
 
+=== "Silverstripe"
+
+    ## Silverstripe
+
+    === "Silverstripe"
```
rpkoller, line 382: I would strike that single `silverstripe` tab. that way you have silverstripe three times in a row. first the active primary tab, then in the heading and then the label of the secondary tab. and then you have a tertiary tab row with composer and git clone. compared to the other quickstarters i would remove the single silverstripe tab for consistency reasons.

pkg/ddevapp/silverstripe.go
rpkoller: ahhhh! thanks for the clarification. then it totally makes sense. 

p.s. github definitely needs a lightbulb emoji!

pkg/ddevapp/silverstripe.go
rpkoller: good to know. i've only tested it in a single site setup. i was unaware that it might cause some troubles in more complicated setups. then a good choice to leave it out. thank you!

docs/content/users/quickstart.md
Firesphere: Sorry, what do you mean?

rpkoller: I've posted the comment before your latest commit that removes line 382 (`=== "Silverstripe"`) from the quickstarter.md around other things. looks good now, thanks for the fix!

Firesphere:
- I've run all tests locally with a clean slate.
- I've updated all docs to be in line with the others.

Firesphere:
How did `DdevXHProfEnabled` fail on Buildkite?

Firesphere:
> @Firesphere if you're having trouble pulling a docker image it's your docker provider (or your network). Restart your docker provider.

It seems this is a common problem for me... Docker pulls regularly time out for no specific reason



> > The reason it didn't trip the tests, is because the .env in the repository is correct.
> 
> IMO the test-silverstripe repo should not include a .env...

It didn't, it included an example environment file that set things correctly. I've updated to have a different .env.example and no actual .env at all, which all pass locally.

I am however, unsure how it still worked despite the faulty example/actual .env file... I would've really expected an error there for the entire set-up.

> I tried the quickstart but when I got to the key part it failed miserably:

Miserably is probably an overstatement? 😆 

As far as I can see, this is again related to my MacOS set-up doin funky stuff, including files I explicitly want to exclude, and creating a tar.gz that is not valid...
I have a gut feeling, something funny is going on with my personal set-up...

I've tried all tests on Linux Kubuntu and KDE Neon without problems, MacOS throws errors that I can't reproduce once I reset the upload to the release again...
I'm 99% sure this is a combination of (my?) MacOS and caching, but I haven't figured out the root cause yet.

rfay:
@Firesphere you're doing a great job on this and it's going to get in. It's a big effort to take this on and you're getting it done with flying colors. I don't think it makes sense to try for v1.22.0, but it will for sure make it into point release v1.22.1, and everybody in the Silverstripe community will be SO HAPPY.

Firesphere:
> @Firesphere you're doing a great job on this and it's going to get in. It's a big effort to take this on and you're getting it done with flying colors. I don't think it makes sense to try for v1.22.0, but it will for sure make it into point release v1.22.1, and everybody in the Silverstripe community will be SO HAPPY. 

Its the whole reason I started this pr... People in the silverstripe community asking if anyone had used ddev, and if it could be easier 😊

I'm learning go as I go here, so even if it makes a point release, I am happy and the community will appreciate my effort to 😊

I must repeat though, your valid feedback, and patience, have been awesome

rfay:
The docs failure will be resolved by https://github.com/ddev/ddev/pull/5166

Firesphere:
> if you're having trouble pulling a docker image it's your docker provider (or your network). Restart your docker provider.

I _think_ something is muted in the output... because after about 5 minutes or so, all docker pulls suddenly go "yep done", and it continues...

docs/content/users/usage/commands.md
Firesphere: Done :)

docs/content/users/quickstart.md
Firesphere: Yeah, I did "double-push", I left the first push because I was about to go to sleep, and intended to wrap up later the next day.

And then saw your comment, which confused me a bit, but makes sense you responded to my initial "fix" where I only fixed the indentation! Thanks for the clarification :)

Firesphere:
For completeness:
- Tests are running okay, on multiple platforms. Everything is okay.
- I removed the test `.env` that came with my installation from the test-silverstripe repository's artifacts. It might have affected tests, but the current artifact/asset works perfect for all environments (Although, I can not manually test on Windows)
- The docs are okay, but some url's are giving issues, which are resolved in a different PR, as per @rfay his comment. A rebase will solve this when the time is right.
- All comments/reviews have been addressed, in one way or another (either solving, or explaining why not)

And, as a repeat:
- I hereby put myself forward as the maintainer of the Silverstripe integration, to ensure DDEV as a project won't be held back by any oddities/issues related to Silverstripe, and any changes will be implemented as soon as possible*.

*: Given I'm currently in a medical mess, "as soon as possible" might be a week or two. It's not life threatening, but it is holding me back in my day-to-day life in a significant way.

Firesphere:
I just rebased and force-pushed. Awaiting tests both local and workflows

Firesphere:
Local workflows pass on MacOS with Colima.
I have to check Linux based systems tomorrow.

rfay:
> Given I'm currently in a medical mess, "as soon as possible" might be a week or two. It's not life threatening, but it is holding me back in my day-to-day life in a significant way.

Your health is far more important than anything else we're talking about here. WAY more important. Wishing you health and take care of yourself.

> Tests are running okay, on multiple platforms. Everything is okay.

You don't have to run all the tests locally of course, that's what CI is for. I normally only run affected tests locally, and don't run on multiple platforms unless CI is pointing out a problem.

After you get this first PR in, Github won't insist on me manually allowing your tests to run, they'll run automatically on push.

> I don't know why buildkite is breaking on xhprof, it is completely unrelated.

TestDdevXhprofEnabled just has *lots* of random failures. It's on my list to spend some time with.

> I hereby put myself forward as the maintainer of the Silverstripe integration, to ensure DDEV as a project won't be held back by any oddities/issues related to Silverstripe, and any changes will be implemented as soon as possible*.

So appreciated!

Firesphere:
After a rebase on the latest master branch, I still see an error on XHProf, which doesn't make sense, and an "agent lost", for build 5517, which I assume means buildkite crashed again?

rfay:
Yes "agent lost" means Docker Desktop crashed. 

No, TestDdevXhprofEnabled failures can never have anything to do with your code.

rfay:
I'll return to this when v1.22.0 issues have calmed down, no need to worry. No need to rebase either, it doesn't affect (and is not affected by) anything else.

rfay (approved):
I think this should be OK to go into DDEV v1.22.1, congrats!

rfay:
Rebased.

rfay:
Congrats on an amazing effort. Will be in DDEV v1.22.1 shortly.

PR #5122, opened 2023-07-11
refactor: update settings.php for the Drupal 10 project type, fixes #5121

### LINKED ISSUE #5121: Update the settings.php in pkg/ddevapp/drupal/drupal10 to adjust to the changes in drupal 10.1

In todays Lean coffee table call we've discussed a feature that got introduced in Drupal 10.1 and is set in the settings.php only. I was a bit confused when I was unable to find it in my fresh Drupal 10.1 install until i realized i've used `ddev config` to set up the project. When i compared the settings.php with the default.settings.php i actually found the changes introduced by that feature (https://www.drupal.org/project/drupal/issues/3027639). on a diff between the settings.php and the default.settings.php there are a few more changes 

on top `// @codingStandardsIgnoreFile` changed to `// phpcs:ignoreFile` , transaction support section got removed, anything in relation to databases got changed significantly, a fallback to http for update manager got added and so on. there are quite a few changes that got in 10.1, since 10.0 was about deprecation and updating versions mainly. 

### Describe your solution

adjust the current settings.php for the drupal 10 project type to the changes introduced with 10.1.x

### Describe alternatives

_No response_

### Additional context

_No response_

rpkoller:
* #5121

Copied the content of the default.settings.php from a Drupal 10.1.0 install and updated the settings.php for the Drupal 10 project type. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5122"></a>

### DIFF
--- a/pkg/ddevapp/drupal/drupal10/settings.php
+++ b/pkg/ddevapp/drupal/drupal10/settings.php
@@ -1,6 +1,6 @@
 <?php
 
-// @codingStandardsIgnoreFile
+// phpcs:ignoreFile
 
 /**
  * @file
@@ -115,14 +115,6 @@ $databases = [];
  * namespace. This is optional for projects managed with Composer if the
  * driver's namespace is in Composer's autoloader.
  *
- * Transaction support is enabled by default for all drivers that support it,
- * including MySQL. To explicitly disable it, set the 'transactions' key to
- * FALSE.
- * Note that some configurations of MySQL, such as the MyISAM engine, don't
- * support it and will proceed silently even if enabled. If you experience
- * transaction related crashes with such configuration, set the 'transactions'
- * key to FALSE.
- *
  * For each database, you may optionally specify multiple "target" databases.
  * A target database allows Drupal to try to send certain queries to a
  * different database if it can but fall back to the default connection if not.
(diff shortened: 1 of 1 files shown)


rpkoller:
Corresponding issue is: https://github.com/ddev/ddev/issues/5121

rfay (approved):
Surprising amount of change for a point release! Looks good to me, just the one question: Is there any reason this would break anybody on 10.0.*? Looks to me like it's all comments anyway?

rpkoller:
I've compared the current settings.php generated by ddev with the settings.php shipping with 10.1.1. . the active not commented out lines are identical, the differing parts are all commented out. as i said the reason for the surprising amount change is that 10.0 was reserved for deprecations and updating dependencies and 10.1 is for the new features.
and i've did the following. ran `ddev config`and then composer created a recommended-project in 10.0.10. i've then updated the settings.php and then installed the site. it behaves correctly as well (wouldnt expected otherwise based on the fact that the active lines are identical). from my perspective it looks okay. but i think it would be good to ask a drupal developer more into the weeds if there might some potential problems down the road? 

and i wonder also about another detail. i went back until drupal 8.0 and compared the active lines in the settings.php and it looks like the default active lines are identical since drupal 8.0. the differing parts come in the commented out sections only. would it make sense to add a comment telling the user on which version of drupal the settings.php generated with `ddev config` is based on?

rfay:
People who care will create their own settings.php, and that's what should be done in a real site prep. DDEV just provides this as a convenience. This is fine as-is, thanks for paying attention and noticing the change.

PR #5130, opened 2023-07-12
docs: use import-db --file instead of import-db --src

rfay:
## The Issue

When we moved from `ddev import-db --src` to `ddev import-db --files`  we didn't update a number of references to the old usage.


rfay:
There are still `ddev import-files --src=` out there; those should be changed to `--source`

PR #5139, opened 2023-07-14
feat: add new disable-upload-dirs-warning configuration

rfay:
## The Issue

Earlier we had the option of `upload_dirs: false` to turn off the pester about `upload_dirs` not being set and mutagen being enabled. This ended up being a very complex bit of code and was rolled back.

Now we need another way to do it.

Add `disable_upload_dirs_warning`

On a `php` project without `upload_dirs`, `ddev start` should show the warning.
After `ddev config --disable-upload-dirs-warning` it should not show the warning.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5139"></a>

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -139,7 +139,7 @@ Whether to disable the standard warning issued when a project is using `performa
 | -- | -- | --
 | :octicons-file-directory-16: project | `false` | Can be `true` or `false`.
 
-When `true`, DDEV will not issue the normal warning on `ddev start`: "You have Mutagen enabled and your 'php' project type doesn't have `upload_dirs` set".
+When `true`, DDEV will not issue the normal warning on `ddev start`: "You have Mutagen enabled and your 'php' project type doesn't have `upload_dirs` set". See [Mutagen and User-Generated Uploads](../install/performance.md#mutagen-and-user-generated-uploads) for context on why DDEV avoids doing the Mutagen sync on `upload_dirs`.
 
 ## `docroot`
 
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -139,7 +139,7 @@ Whether to disable the standard warning issued when a project is using `performa
 | -- | -- | --
 | :octicons-file-directory-16: project | `false` | Can be `true` or `false`.
 
-When `true`, DDEV will not issue the normal warning on `ddev start`, "You have Mutagen enabled and your 'php' project type doesn't have `upload_dirs` set".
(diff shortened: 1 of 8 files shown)


docs/content/users/configuration/config.md
```diff
@@ -131,6 +131,16 @@ Whether to disable CMS-specific settings file management.
...
 
+## `disable_upload_dirs_warning`
+
+Whether to disable the standard warning issued when a project is using `performance_mode: mutagen` but `upload_dirs` is not configured.
+
+| Type | Default | Usage
+| -- | -- | --
+| :octicons-file-directory-16: project | `false` | Can be `true` or `false`.
+
+When `true`, DDEV will not issue the normal warning on `ddev start`, "You have Mutagen enabled and your 'php' project type doesn't have `upload_dirs` set".
```
mattstein, line 142: Swap comma for colon:

```suggestion
When `true`, DDEV will not issue the normal warning on `ddev start`: "You have Mutagen enabled and your 'php' project type doesn't have `upload_dirs` set".
```

mattstein: It might be useful to clarify why DDEV bothers you with this warning in the first place and when it makes sense to disable it so someone can have more confidence about silencing the warning vs. taking some other action.

The message itself says “if this is intended,” so it’d be a good place to offer an example of what intentional choices look like for someone that’s trying to understand.

docs/content/users/usage/commands.md
```diff
@@ -180,6 +180,7 @@ Flags:
 * `--dbimage-extra-packages`: A comma-delimited list of Debian packages that should be added to db container when the project is started.
 * `--default-container-timeout`: Default time in seconds that DDEV waits for all containers to become ready on start. (default `120`)
 * `--disable-settings-management`: Prevent DDEV from creating or updating CMS settings files.
+* `--disable-upload-dirs-warning`: Don't issue a warning when a project does not have `upload_dirs` set but is using `performance_mode: mutagen`.
```
mattstein, line 183: Use a more active verb, and flip the negative and affirmative so it’s easier to follow:

```suggestion
* `--disable-upload-dirs-warning`: Suppresses warning when a project is using `performance_mode: mutagen` but does not have `upload_dirs` set.
```

mattstein (changes requested):

docs/content/users/configuration/config.md
rfay: Thanks, attempted to improve that in https://github.com/ddev/ddev/pull/5139/commits/4987984231df8ef6853111ff1ab0123cded95866

mattstein: Perfect! 👏

PR #5148, opened 2023-07-15
fix: ddev config should not stumble on invalid app type, fixes #5146

### LINKED ISSUE #5146: Improve the behavior of `ddev config` on an existing project with an invalid project type

When switching back to the HEAD version from the silverstripe PR version i completely forgot for one project that it was created in silverstripe. I ran `ddev config`for that project and the output was the following: 

```
$> ddev config
You are reconfiguring the project at /Users/rkoller/Sites/craft3.
Project name (craft3): 

The docroot is the directory from which your site is served.
This is a relative path from your project root at /Users/rkoller/Sites/craft3
You may leave this value blank if your site files are in the project root
Docroot Location (public): 
Found a php codebase at /Users/rkoller/Sites/craft3/public.
Project Type [backdrop, craftcms, django4, drupal10, drupal6, drupal7, drupal8, drupal9, laravel, magento, magento2, php, python, shopware6, typo3, wordpress] (silverstripe):
```
i just pressed return accepting silverstripe. the output returned was: 

```
'silverstripe' is not a valid project type. Allowed project types are: backdrop, craftcms, django4, drupal10, drupal6, drupal7, drupal8, drupal9, laravel, magento, magento2, php, python, shopware6, typo3, wordpress

Project Type [%s] (%s): backdrop, craftcms, django4, drupal10, drupal6, drupal7, drupal8, drupal9, laravel, magento, magento2, php, python, shopware6, typo3, wordpresssilverstripe
```
three problems first the output is twice. the first time correclt stating silverstripe is not a valid project type. the second time using `Project Type [%s] (%s)`. the first time the last output is wordpress without silverstripe the second time the last output is wordpresssilverstripe and the cursor for text input is right after the `e`of silverstripe.
Screen Shot 2023-07-15 at 22 44 24

### Describe your solution

After a none existing project type is submitted i would return: 

```
[project type] is not a valid project type. 
Choose one of the following project types: [backdrop, craftcms, django4, drupal10, drupal6, drupal7, drupal8, drupal9, laravel, magento, magento2, php, python, shopware6, typo3, wordpress]: [cursor]
```

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

* #5146

* It kept repeating on invalid type
* The second through nth times it used `print` instead of the correct `printf`, so showed an invalid message

* Replace print with printf
* If invalid project type, use detected project type

Set an invalid type in .ddev/config.yaml and then try to do interactive config.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5148"></a>

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1259,7 +1259,7 @@ func (app *DdevApp) AppTypePrompt() error {
 	typePrompt := "Project Type [%s] (%s): "
 
 	defaultAppType := app.Type
-	if app.Type == nodeps.AppTypeNone {
+	if app.Type == nodeps.AppTypeNone || !IsValidAppType(app.Type) {
 		defaultAppType = detectedAppType
 	}
 
@@ -1269,7 +1269,7 @@ func (app *DdevApp) AppTypePrompt() error {
 	for !IsValidAppType(appType) {
 		output.UserOut.Errorf("'%s' is not a valid project type. Allowed project types are: %s\n", appType, validAppTypes)
 
-		fmt.Print(typePrompt, validAppTypes, appType)
+		fmt.Printf(typePrompt, validAppTypes, appType)
 		appType = strings.ToLower(util.GetInput(appType))
 	}
 


rpkoller:
I've tried two scenarios: 

1) the silverstripe install with the project type drupal11
=> the suggested default on step three is php which is correct.  

2) a fresh backdrop install with `ddev config --auto` setting the project type to backdrop. i've then manually changed it to drupal11
=> the suggested default on step three is backdrop which is correct

i really like the approach and i can confirm that it is working correctly in my manual testing. thank you!

PR #5149, opened 2023-07-16
ci: minor improvement to TestDdevNoProjectMount [skip ci]

rfay:
## The Issue

When running locally TestDdevNoProjectMount would fail just because it was set up wrong.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5149"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -820,9 +820,6 @@ func TestDdevStartUnmanagedSettings(t *testing.T) {
 
 // TestDdevNoProjectMount tests running without the app file mount.
 func TestDdevNoProjectMount(t *testing.T) {
-	if nodeps.PerformanceModeDefault == types.PerformanceModeMutagen || nodeps.NoBindMountsDefault {
-		t.Skip("Skipping because this doesn't make sense with mutagen or NoBindMounts")
-	}
 	assert := asrt.New(t)
 	app := &ddevapp.DdevApp{}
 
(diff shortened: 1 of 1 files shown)


Firesphere:
I'm not sure if this is related or not, but on my Linux based machine, I had to remove `~/.ddev/testcache` before running a test, because otherwise, the test would be run against an "old" download.

This specifically happened when I had spotty/bad internet connection though. So I'm not sure how much my experience applies.

rfay:
You probably posted that comment in the wrong place? This PR only slightly changes the behavior of TestDdevNoProjectMount mostly when running it locally.

The stuff stored in ~/.ddev/testcache really mostly affects TestDdevFullSiteSetup. But recent changes were made to explicitly give each version of the tarball a versioned name, so you should see different version names for different tarballs in ~/.ddev/testcache/tarballs:

```
$ ls -l
total 162624
-rw-r--r--  1 rfay  staff  21479296 Jul  4 10:58 00fbfea3_drupal-9.4.2.tar.gz
-rw-r--r--  1 rfay  staff    289978 Jul 15 19:03 181a3a21_wordpress5.8.2_db.sql.tar.gz
-rw-r--r--  1 rfay  staff  22595018 Jul 15 18:09 2e7d88c0_drupal-9.5.10.tar.gz
-rw-r--r--  1 rfay  staff  21171263 Jul  4 10:55 82f32c88_wordpress-6.0.1.tar.gz
-rw-r--r--  1 rfay  staff  15085301 Jul  5 11:45 cbceb39f_wordpress-5.8.2.tar.gz
-rw-r--r--  1 rfay  staff   2634224 Jul 15 19:10 cff473ce_wordpress5.8.2_files.tar.gz
```

It's too bad you killed them off because there might have been a bug to discover there. But all you'd have to do is publish a new release and see what happens when you run it locally, make sure that you get a new fresh tarball. These tarballs are extracted into `~/.ddev/testcache/<name>`

Thanks for paying attention!

Firesphere:
> You probably posted that comment in the wrong place? This PR only slightly changes the behavior of TestDdevNoProjectMount mostly when running it locally.

D'oh! Probably from clicking around in all the issues.
And indeed, I'll have to test a bit more, could be any reason as to why it happened.

Issue #4858, opened 2023-05-01
Add docs and automated testing for Orbstack


rfay:
Casual users have reported good results with Orbstack as an alternative to Docker Desktop or Colima on macOS, https://orbstack.dev/

See https://discord.com/channels/664580571770388500/1090234521091309648/1090234521091309648

We should probably try it out, maybe add automated testing, add it as an option for a docker provider.

highermath:
Is this something that we need to do? With both Colima and Orbstack, we expect them to do what  Docker does, and presumably they test for compatibility. Don't misunderstand, I think that it is awesome that we give this level of support, as long as we can afford the resources to do it.

I've only had one compatibility issue using Colima, and that was on a non-ddev project. It was reported and resolved.

Also, we should not lose site of the fact that Orbstack is a bit of a pig in a poke. They are specific when they say that the beta is free. We don't know what they plan once they get out of beta, but it is clear that they plan to monetize it. Maybe they'll have a free version for devs, and maybe not.

rfay:
Docker Desktop is also not free and hasn't been for some time. Monetizing projects is fine. The goal with DDEV is to make things as useful as possible, and this may be useful.

highermath:
FWIW, they are proposing $8/mo. for devs. I made a counterproposal of $10/yr. We'll see what happens.

rfay:
I just tried out Orbstack on mac Apple Silicon and it passed my trivial tests with DDEV with flying colors, and was really, really fast. I had mutagen turned on... We should support it. 

* #5113

cmuench:
We test it on several Mac's (Intel and M1/M2) for weeks now. It's super stable and we had zero crashes. Docker for Mac is much more unstable.

thamas:
Someone who knows these environments using/switching to OrbStack could be trivial. But for us others it would be nice to have a _step-by-step guide_ in DDEV's documentation about what should be done if we want to try/use it (with existing projects). Thanks!

cmuench:
@thamas 
In my case I did snapshots of the DB in my project.
Then I installed OrbStack (Docker for Mac can still be installed but should be stopped).
The I did a `ddev snapshot restore` in my project which "imports" the DB again.
As I know the OrbStack team works on a way to import existing docker volumes.

You are right. It's trivial but we should document it.

rfay:
[...]

PR #5158, opened 2023-07-18
docs: add FAQ explaining how to migrate between Docker providers

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/4858#issuecomment-1640158592

Migrating between Docker providers is covered in https://ddev--5156.org.readthedocs.build/en/5156/users/install/docker-installation/#migrating-projects-from-docker-desktop-to-colima but the general case is not covered.

Add to FAQ


rfay:
Closing and going to use this as an example in contributor training today.

PR #5162, opened 2023-07-18
feat: create global pip & pip3 cmds and update commands.md accordingly,  fixes #5161

### LINKED ISSUE #5161: Add global commands for pip and pip3

DDEV 1.22 adds projects types for django4 and python as well as ships with pip and pip3 installed. On the other hand DDEV already provides for package managers in php and javascript global commands. it would be a reasonable reasonable to add global commands for the python based package manager as well.  

### Describe your solution
see https://github.com/ddev/ddev/pull/5162

### Describe alternatives

_No response_

### Additional context

_No response_

rpkoller:
## The Issue
https://github.com/ddev/ddev/issues/5161

It creates to global commands (pip & pip3) and updates the commands.md

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5162"></a>

### DIFF
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -929,6 +929,7 @@ Example:
 # Output the web container’s PHP version
 ddev php --version
 ```
+
 ## `pip3`
 
 Run [`pip3`](https://pip.pypa.io/en/stable/) (for Python 3) inside the web container (global shell web container command).
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -944,7 +944,7 @@ ddev pip3 --version
 ddev pip3 install
 
 # Update Python packages using `pip`
-pip3 pip3 install --upgrade
+ddev pip3 install --upgrade
 ```
 
 ## `poweroff`
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -929,24 +929,6 @@ Example:
 # Output the web container’s PHP version
 ddev php --version
 ```
-
-## `pip`
-
-Run [`pip`](https://pip.pypa.io/en/stable/) (for Python 2) inside the web container (global shell web container command).
-
-Example:
-
-```shell
-# Output the web container’s pip version
-ddev pip --version
-
-# Install Python packages using `pip`
-ddev pip install
-
-# Update Python packages using `pip`
-pip3 pip install --upgrade
-```
-
 ## `pip3`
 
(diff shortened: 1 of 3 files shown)


rfay (changes requested):
I think this should probably only be available for python and django4 project types? I hate to see so many thousands of commands cluttering up `ddev -h`. 

I don't think we need both `pip` and `pip3` since `pip` in this container is `pip3` (only python3 is installed in the container by default, but python-is-python3 is installed. I can be talked out of this stance, but regardless I'd rather have just one file and a symlink? Could that be done?

rpkoller:
> I think this should probably only be available for python and django4 project types? I hate to see so many thousands of commands cluttering up ddev -h.

I also considered adding `## ProjectTypes: django4,python` to the two global commands when i created the PR. but couldnt it be the case that someone wants to use a python package for whatever reason in a php based project like drupal? making pip exclusive to django and python would make it opinionated? and the other package managers aren't project type specific either? 

> I don't think we need both pip and pip3 since pip in this container is pip3 (only python3 is installed in the container by default, but python-is-python3 is installed. I can be talked out of this stance, but regardless I'd rather have just one file and a symlink? Could that be done?

oh i never version checked python and python3 with `python --version`. i am not sure then. i would rather tend to remove the global command for `pip` and just keep `pip3` since there is only `python3` available in the web container. keeping `pip` would be wrong and would create false expectations. would it even be possible to remove the `python` command in the web container a well, since there is no python2 available? and the python global command could be renamed to python3 ? 
and in addition also communicate in the docs that python 3 is installed and python2 not available? but by only providing the commands `python3` and `pip3` it should be probably clear enough for regular python users but still having it noted in the docs as well would make sense from my point of view?

mattstein (approved):
Looks good to me! We use the phrase “(global shell web container command)” all over this page without clarifying what that means, but that’s not a problem here since this PR is following suit.

rfay:
I'm sure this will land, but after v1.22.0. Thanks for the initiative!

docs/content/users/usage/commands.md
```diff
@@ -930,6 +930,23 @@ Example:
...
+
+```shell
+# Output the web container’s pip version
+ddev pip3 --version
+
+# Install Python packages using `pip`
+ddev pip3 install
+
+# Update Python packages using `pip`
+pip3 pip3 install --upgrade
```
tbal, line 947: this should probably be more like, right?
```suggestion
ddev pip3 install --upgrade
```

rfay:
Thanks so much for this contribution, but I don't think we need these wrappers at this time. I think people are better to be using `ddev exec` at this point.

PR #5169, opened 2023-07-19
fix: don't report json-formatted commands to Amplitude

rfay:
## The Issue

People who use the PhpStorm extension for DDEV have `ddev describe -j` happening every minute or so in the background. This causes loads of useless traffic.

Don't report commands when using `--json-ouput`

`for ((i=0; i<100; i++)); do ddev describe -j; done >/tmp/out 2>&1 && ddev debug instrumentation flush` should *not* result in new data in amplitude

`for ((i=0; i<100; i++)); do ddev describe; done >/tmp/out 2>&1 && ddev debug instrumentation flush` SHOULD result in new data in amplitude

https://app.amplitude.com/analytics/ddev/chart/vqe8wcsj can be used to study this; the filtering will have to be adjusted for your use.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5169"></a>

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -49,7 +49,12 @@ Support: https://ddev.readthedocs.io/en/stable/users/support`,
 			argsCopy = []string{}
 		}
 
-		amplitude.TrackCommand(&cmdCopy, argsCopy)
+		// We don't want to send to amplitude if using --json-output
+		// That captures an enormous number of PhpStorm running the
+		// ddev describe -j over and over again.
+		if !output.JSONOutput {
+			amplitude.TrackCommand(&cmdCopy, argsCopy)
+		}
 
 		// Skip docker and other validation for most commands
 		if command != "start" && command != "restart" {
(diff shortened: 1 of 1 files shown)


rfay:
I note that PhpStorm plugin uses `DDEV_NONINTERACTIVE=true ddev describe -j` so we could also target DDEV_NONINTERACTIVE.

Firesphere:
I'm not sure if the problem is in DDEV, and if DDEV should fix the "issue". Why not have the PHPStorm add-on fixed?

I noticed while using PHPStorm, that it way too often lets me know that "ddev is unreachable". Even when I simply do a `ddev restart`.
I feel the problem is in the add-on, not in DDEV itself.

I do think though, that maybe DDEV could mitigate the load, possibly via a way of caching the response for X minutes?

I'm not sure if it's possible, but if the JSON is called, the response could be loaded from an existing previous response for 10 minutes or so, reducing the load that way?

rfay:
> I'm not sure if the problem is in DDEV, and if DDEV should fix the "issue". Why not have the PHPStorm add-on fixed?

You'll see a variety of discussions about this topic over in https://github.com/php-perfect/ddev-intellij-plugin

It queries DDEV state using `ddev describe -j`. If it did it every 5 minutes instead of whatever it's doing, it would still completely distort the statistics on Amplitude.  People leave a project running and leave PhpStorm sitting there and ...

But if you have a better idea how to do it, an issue with a suggestion might go ok over there.  

> I'm not sure if it's possible, but if the JSON is called, the response could be loaded from an existing previous response for 10 minutes or so, reducing the load that way?

Why even bother polling DDEV for results if they are cached?

But please, followups over there where they'll do some good.

gilbertsoft (approved):
Good catch!

PR #4961, opened 2023-06-02
20230502 hakre secure apt gpg

hakre:
## How This PR Solves The Issue

By the changes in the last commit.

## Manual Testing Instructions

Boot a new Debian or Ubuntu box and install ddev according to the changed instructions.

## Automated Testing Overview

Coverage should not have changed.

Testing is building the documentation.

On Ubuntu:

~~~
$ ./configure
...
$ make -O -j ci
...
~~~

## Related Issue Link(s)

#4960

## Release/Deployment Notes

There is a new `./configure` script installing the Python / NodeJS dependencies in the Makefiles' CI_GOALS. These are accessible via `make ci` and can run in parallel. 

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/4961"></a>

Issue #4960, opened 2023-06-02
Debian/Ubuntu keyring location should be /etc/apt/keyrings, not /etc/apt/trusted.gpg.d


hakre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

n/a

### Expected Behavior

n/a

### Actual Behavior

n/a

### Steps To Reproduce

n/a

### Anything else?

This has been touched in earlier discussion in #3997 about the correct location, the functional fix to de-armor the gpg key was done in #4324 which missed the location fix.
configuration change request for the installation documentation:

* `/etc/apt/trusted.gpg.d/ddev.gpg` -> `/usr/share/keyrings/ddev.gpg`

cf. https://unix.stackexchange.com/a/582853/9340 and [`sources.list(5)`](https://manpages.debian.org/bullseye/apt/sources.list.5.en.html#THE_DEB_AND_DEB-SRC_TYPES:_OPTIONS)

rfay:
I'm not sure there's an absolutely correct place to put the key. If you can give a more authoritative source than Stack Overflow article, it will be considered.  I closed your PR because it made many, many irrelevant changes.

hakre:
Understood, but unfortunately, you closed the pr which now prevents me from changing the history. I'm sorry if you had no other way to deal with it, that was not my intention (cf. "How This PR Solves The Issue" section). And this is the reason why the _relevant_ changes are in the last commit. Use version control to your benefit, perhaps even despite using GitHub. Please reopen to keep the conversation going.

rfay:
Please continue the conversation here to provide documentation about why you think your solution is better. 

All you have to do to reopen a PR is to take your branch and update it (and rebase it please, and don't change anything but what you want to change) and create a PR.

But before you do that.. please provide authoritative documentation about why you think the key should be elsewhere.

hakre:
> please provide authoritative documentation about why you think the key should be elsewhere.

So in the hope you consider it authoritative, let me start by suggesting you the [Debian Wiki page](https://wiki.debian.org/DebianRepository/UseThirdParty) _Instructions to connect to a third-party repository_ prominently mentioned in that _Unix & Linux Stackexchange_ article:

> You can find official information from Debian here: https://wiki.debian.org/DebianRepository/UseThirdParty

And then ask you to explicitly take notice of the following sentence in the first _OpenPGP certificate distribution_ section:

[...]

PR #5172, opened 2023-07-20
Store ddev.gpg in keyrings instead of trusted.gpg.d

ted933:
## The Issue
On Ubuntu systems keys stored in /etc/apt/trusted.gpg.d validate packages for the whole linux installation. Custom sources should not use this directory for storing their keys as this creates a vulnerability.

See https://www.digitalocean.com/community/tutorials/how-to-handle-apt-key-and-add-apt-repository-deprecation-using-gpg-to-add-external-repositories-on-ubuntu-22-04 for detailed explanation.

Storing the ddev.gpg key in /etc/apt/keyrings/ mitigates this vulnerability.

Follow the installation instructions.

## Release/Deployment Notes
Apt key location changed for Debian/Ubuntu systems.


rfay:
Please take a look at 
* https://github.com/ddev/ddev/issues/4960

and 
* https://github.com/ddev/ddev/pull/4961

There's more to be done than just this little bit. But those will give you more context. Definitely want to do it, would like to do more than you have here.

ted933:
Setting access rights corresponding to https://docs.docker.com/engine/install/ubuntu/

rfay (changes requested):
I think we'll want to make the changes more than just in the docs. If you want to take that on it's welcome. See the previous PR, for example, and the discussion in the previous issue.

ted933:
> I think we'll want to make the changes more than just in the docs.

Please be more precise. What else do you expect?

rfay:
The things that have to be done are listed in the initial comment, https://github.com/ddev/ddev/pull/5172#issuecomment-1643896860 - Just search for the many places that have to be changed as was done in https://github.com/ddev/ddev/pull/4961 and do them. That PR had a number of problems, but the idea was right.

rfay:
I'm working on adding a commit to do the things.

rfay:
@ted933 @hakre please review and see what you think, thanks!

gilbertsoft (changes requested):
This PR introduces two different locations for the keys, one in `/usr/share/keyrings` and a second in `/etc/apt/keyrings`, I think we should use one only!

rfay:
@gilbertsoft The PR uses /usr/share/keyrings only for deb.sury.org packages, and it's because that's what they currently recommend for their case. https://packages.sury.org/php/README.txt

gilbertsoft (approved):
Thanks for the additional information. I had a look at the various links and it looks fine to me.

PR #5175, opened 2023-07-20
fix: don't run MutagenSyncFlush on non-running project during config, fixes #5168

### LINKED ISSUE #5168: `ddev config` when project is stopped gets "Could not write settings file: failed to get web container, err='unable to find container ddev-d10-web: <nil>'"
macOS OrbStack, d10 project stopped, mutagen enabled

### Expected Behavior

`ddev config` should work whether project is running or not

### Actual Behavior

```
ddev poweroff 
ddev config --project-type=drupal10
You are reconfiguring the project at /Users/rfay/workspace/d10.
You have specified a project type of 'php' but a project of type 'drupal10' was discovered in /Users/rfay/workspace/d10/web
Could not write settings file: failed to get web container, err='unable to find container ddev-d10-web: <nil>'
```

This is a flaw in MutagenSyncFlush(). It can be fixed by not continuing if the web container is not running, but it will have effects broader than that.

rfay:
## The Issue

* #5168 

This seems like the lightest touch possible here, just check to see if project is running before trying to sync it. The only reason for the sync is to get the new settings into the container right away, and since it's just one file normally, mutagen has probably already synced it without even a flush. However, if the project is stopped, it will be synced on start.

I first experimented with changing MutagenSyncFlush itself, and it can be done there, but it would have effects on all callers, and I'm risk-averse at this point in the release cycle.

Use #5168 instructions

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5175"></a>

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -281,9 +281,13 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 		}
 		return settingsPath, nil
 	}
-	err = app.MutagenSyncFlush()
-	if err != nil {
-		return "", err
+
+	// If the project is not running, it makes no sense to sync it
+	if s, _ := app.SiteStatus(); s == SiteRunning {
+		err = app.MutagenSyncFlush()
+		if err != nil {
+			return "", err
+		}
 	}
 
 	return "", nil


gilbertsoft (approved):
Looks good

PR #5185, opened 2023-07-22
fix: traefik dashboard should not bind all interfaces by default

stasadev:
## The Issue

There is no difference in `router_bind_all_interfaces` for Traefik:

```
ddev config global --router-bind-all-interfaces=true
ddev start
cat ~/.ddev/.router-compose.yaml | grep 9999
    - 9999:9999

ddev config global --router-bind-all-interfaces=false
ddev start
cat ~/.ddev/.router-compose.yaml | grep 9999
    - 9999:9999
```

Traefik must consider the `router_bind_all_interfaces` config.

```
ddev config global --router-bind-all-interfaces=false
ddev start
cat ~/.ddev/.router-compose.yaml | grep 9999
    - "127.0.0.1:9999:9999"

ddev config global --router-bind-all-interfaces=true
ddev start
cat ~/.ddev/.router-compose.yaml | grep 9999
    - "9999:9999"
```

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5185"></a>

### DIFF
--- a/pkg/ddevapp/router_compose_template.yaml
+++ b/pkg/ddevapp/router_compose_template.yaml
@@ -24,7 +24,7 @@ services:
     {{ if eq .Router "traefik" }}
     # Traefik router; configured in static config as entrypoint
     # TODO: Make this configurable? Put it somewhere else?
-    - 9999:9999
+    - "{{ if not .router_bind_all_interfaces }}{{ $dockerIP }}:{{ end }}9999:9999"
     {{ end }}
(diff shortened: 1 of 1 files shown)


rfay (approved):
I think you're right! Thanks for chasing this down. After all those months it's been in the wild :)

rfay:
Manually tested, works exactly as it should, thanks!

PR #5186, opened 2023-07-22
fix: change the term Docker to upper-case and adjust the term ddev to the DDEV word mark

rpkoller:
## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5186"></a>


rpkoller:
I will include the changes from this PR to one for all go files.

PR #5187, opened 2023-07-22
ci: update .textlintrc to fix capitalization and typo

rpkoller:
- Correct potential typo for Bash
- Add Colima, GitHub and Traefik to the list of terms

Those are the changes i've already suggested in https://discord.com/channels/664580571770388500/1132360586588585995/1132360586588585995 

But i think the list of terms might be even extended by `MariaDB`, `Postgres` or `PostrgreSQL` (it isn't addressed consistently - at least when i've went through the go files) and maybe also `Zsh`. 

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5187"></a>


rfay (approved):
Thanks!

mattstein (approved):
Looks great to me! My vote would be for “PostgreSQL” if you add it. I know Postgres is common and not _incorrect_, but “PostgreSQL” [seems like](https://www.postgresql.org) the canonically-accurate term.

rpkoller:
> Looks great to me! My vote would be for “PostgreSQL” if you add it. I know Postgres is common and not _incorrect_, but “PostgreSQL” [seems like](https://www.postgresql.org) the canonically-accurate term.

I went ahead and posted an initial draft as an PR yesterday night: https://github.com/ddev/ddev/pull/5188 . In the context of Postgres i went with PostgreSQL for comments and user facing output text per the recommendation of @rfay in one conversation on discord (i can't find it right, but it was the most reasonable approach to that imho which would be also in line with the preference you've voiced) 




 (around the other changes) per the suggestion of @rfay

PR #5188, opened 2023-07-24
feat: adjust the ddev golang files to the textlint rules

rpkoller:
## The Issue
i've tried to come up with an initial draft adjusting the comments and text outputs of the DDEV Golang files to the rules found in the .textlintrc file. it is far from perfect and an initial draft. i've searched for the different terms in VSCode (searched solely for go files excluding those located in /vendor) and then tried to adjust them accordingly.
One definite problem is the style requirement for golang files to have the first word for `fmt.Errorf` error strings being lower case. that collides if a sentence starts for example with DDEV or Colima. In those cases i've made the term lower case but the other option might be to rephrase them. feedback welcome. thanks 

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5188"></a>


pkg/ddevapp/config.go
```diff
@@ -930,14 +930,14 @@ redirect_stderr=true
...
 	extraDBContent := ""
 	if app.Database.Type == nodeps.Postgres {
-		// Postgres 9/10/11 upstream images are stretch-based, out of support from Debian.
-		// Postgres 9/10 are out of support by Postgres and no new images being pushed, see
+		// PostgreSQL 9/10/11 upstream images are stretch-based, out of support from Debian.
+		// PostgreSQL 9/10 are out of support by PostgreSQL and no new images being pushed, see
 		// https://github.com/docker-library/postgres/issues/1012
-		// However, they do have a postgres:11-bullseye, but we won't start using it yet
-		// because of awkward changes to $DBIMAGE. Postgres 11 will be EOL Nov 2023
+		// However, they do have a Postgres:11-bullseye, but we won't start using it yet
```
stasadev, line 939:
```suggestion
		// However, they do have a postgres:11-bullseye, but we won't start using it yet
```

pkg/ddevapp/db.go
```diff
@@ -9,13 +9,13 @@ import (
 	"strings"
 )
 
-// GetExistingDBType returns type/version like mariadb:10.4 or postgres:13 or "" if no existing volume
-// This has to make a docker container run so is fairly costly.
+// GetExistingDBType returns type/version like mariadb:10.4 or Postgres:13 or "" if no existing volume
```
stasadev, line 12:
```suggestion
// GetExistingDBType returns type/version like mariadb:10.4 or postgres:13 or "" if no existing volume
```

pkg/ddevapp/config.go
rpkoller: uhhhhh what an oversight on my end, that slipped through with the back and forth. thank you for noticing both cases!

rfay (changes requested):
There are one or more tests that are broken by this, like TestCopyDir, please fix. 

In general, massive changes like this aren't all that welcome, because they're difficult to review and can have unanticipated results. We'll get through this one, but maybe in the future you could choose a smaller scope for your PR and do a few PRs that are easier to review.

rfay:
Next one to solve is TestDebugDownloadImages

rfay (approved):
This is an amazing effort, and thanks for undertaking it! 

I've rebased, resolved many conflicts, and added a couple of the tiniest changes.

Before you undertake a huge thing like this in the future, please
* Talk about it in an issue first
* Do smaller chunks, not the whole thing at once. That way it's likely to get in before it gets stale.
* Focus on customer-facing verbiage. There's no need to fix comments or things in tests
* Always keep in mind that massive changes like this require review and introduce risk. It's always a trade-off.

Please note that I won't be enforcing rules like this in the code on non-user-facing verbiage, so they will sneak in again.

I read all the way through this. I hope I did it at the right time and didn't make any mistakes in the rebasing.

PR #5195, opened 2023-07-25
Update shell-completion.md

apotek:
Remind users they need to reload their shell for the new completions to work. I added the reminder only to the "macOS Bash + Homebrew" and "Bash/Zsh/Fish on Linux" sections as those are the two I could verify needed a shell reload. I have not tested the others. I would assume they all would need to be reloaded so, if someone wants me to, I can add a "reload your shell" reminder to each section.

### DIFF
--- a/docs/content/users/install/shell-completion.md
+++ b/docs/content/users/install/shell-completion.md
@@ -16,7 +16,7 @@ Shells like bash and zsh need help to do this though, they have to know what the
     ```
 
     !!!note "Bash profile"
-        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
+        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately. You will also need to reload your shell either by reopening your terminal window or executing `$ exec zsh`
 
     * Link completions with `brew completions link`.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/install/shell-completion.md
```diff
@@ -16,7 +16,7 @@ Shells like bash and zsh need help to do this though, they have to know what the
     ```
 
     !!!note "Bash profile"
-        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
+        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately. You will also need to reload your shell either by reopening your terminal window or executing `$ exec zsh`
```
rfay, line 19: I don't believe that's true at all. It's certainly not true with bash. All you have to do is source the profile.

docs/content/users/install/shell-completion.md
```diff
@@ -37,6 +37,7 @@ Shells like bash and zsh need help to do this though, they have to know what the
         tar -zxf ddev_shell_completion_scripts.${VERSION}.tar.gz
         ```
     2. Run `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`. This deploys the `ddev_bash_completion.sh` script where it needs to be. Again, every Linux distro has a different technique, and you may have to figure yours out.
+    3. Remember to reload your shell with `$ exec zsh` or by closing and reopening your terminal.
```
rfay, line 40: Yeah, just sourcing should do the job.

rfay (changes requested):
I don't believe this is correct at all.

docs/content/users/install/shell-completion.md
apotek: It didn't work for me until I closed the terminal (in one case), and in the other case, until I ran `exec zsh`.

rfay: You may be missing the required autoload in your zsh setup. However, I use bash and am certainly no expert on zsh, especially Oh-My-Zsh. Zsh can be set up so many different ways it can fry the brain. I think one of those ways bit you here.

apotek: > Yeah, just sourcing should do the job.

I had to run:

```
autoload -Uz compinstall && compinstall
```
And then reload my shell before completions worked. 

Mac OS Ventura 13.5 running zsh. Installed ddev via homebrew.

```
zsh --version
zsh 5.9 (x86_64-apple-darwin22.0)
```

rfay:
I appreciate you raising this, but since this has been the same for years without trouble, and because zsh is so troublesome with its many permutations,  I think we should leave this alone. Happy to return to it as you learn more and experiment. Closing for now, happy to continue the conversation.

Issue #3679, opened 2022-03-08
Improve `ddev debug test`


rpkoller:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Multiple requests:

Add these things to test_ddev.sh:

- [ ]  Show whether mutagen.yml has been changed (Is that possible?) (Can we show some info about the existing project?)
- [x]  Should show what dir it’s executing and then trying to create in.
- [x]  Add cleanup option
- [x]  Less output showing mutagen - maybe just the one project we’re interested in?
- [x]  Add a little wait before outside curl to let traefik catch up. Maybe a while loop?
- [x] Give an automatic way to post to a gist or capture the output
- [x] Add `colima status` and `colima --version` if they're using colima
- [x] Doesn't work when global  `router_http_port` is set to nonstandard (tries port 80)
- [x] Output the filename both to the script (top) and at bottom, and in browser.
- [x] Check to make sure name resolution of ddev.site works
- [x] Check https/cert
- [ ] See if we can limit exposure of project names or other security-sensitive items

When `ddev debug test` is run (tested with 1.9.0-rc1) you currently get at the end of the test the following recommendation in the browser: 
```
ddev is working. You will want to delete this project with "ddev delete -Oy tryddevproject-13959"
```
by applying that line the test project is removed from ddev and its docker assets are removed again. but noticed today that the test creates some artifacts in the user folder. a tmp folder is created (~/tmp) and for each test run a new directory with a index.php is created. it looks like that at the moment
![Bildschirmfoto 2022-03-08 um 01 38 26](https://user-images.githubusercontent.com/1665422/157142515-d94929d6-c3ac-4941-916a-463969473c91.png)
the clutter might increase over time if the tmp folder is unnoticed by the user. 

### Describe the solution you'd like

the easiest fix would be a note along: 

```
ddev is working. You will want to delete this project with "ddev delete -Oy tryddevproject-13959"
```
that there is a temp folder in the user folder being created containing test assets. just inform the user about the folder.  the cleaner and more convenient variant might be in case of those test files on `ddev delete` also to remove them?

rfay:
It would be nice to have an auto-cleanup like `ddev debug test cleanup`

PR #5212, opened 2023-07-28
feat: add `ddev debug testcleanup`, fixes #5210

### LINKED ISSUE #5210: Running the provided clean-up line from `ddev debug test` deletes all existing project volumes

_No response_

### Expected Behavior

Keep my projects volumes especially the db volumes 🤯

### Actual Behavior

`ddev delete -Oy $(ddev list -j |jq -r .raw[].name | grep tryddevproject)` deletes everything I have created but keeps the tryddev stuff running and as volumes/containers.

### Steps To Reproduce

1. Try to file a bug
2. Read the info provided
3. Run `ddev debug test` to copy some stuff for you
4. See the command `ddev delete -Oy $(ddev list -j |jq -r .raw[].name | grep tryddevproject)` that says it would remove the testing stuff
5. see all of your project containers disapear without any notice or backups
6. freak out

### Anything else?

_No response_

stasadev:
## The Issue

The output for `ddev debug test` contains:

```
If you're brave and you have jq you can delete all tryddevproject instances with this one-liner:
    ddev delete -Oy $(ddev list -j |jq -r .raw[].name | grep tryddevproject)
In the future ddev debug test will also provide this option.
```

It is dangerous to run this one-liner twice in a row, because the second run will be the same as `ddev delete -Oy` without any arguments, and it will delete the DDEV project with its volumes.

Add `ddev debug testcleanup` command that will clean up all projects with `tryddevproject` prefix. No `jq` required.

Run `ddev debug test` several times, then `ddev debug testcleanup`

## Related Issue Link(s)

- https://github.com/ddev/ddev/issues/5210

## Alternative:

Use `xargs`:

```
ddev list -j | jq -r '.raw[].name' | grep tryddevproject | xargs -L 1 -r ddev delete -Oy
```

* `-L 1` - process one line at a time
* `-r` - do not run `xargs` if the result of previous command is empty

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5212"></a>

### DIFF
--- a/cmd/ddev/cmd/debug-testcleanup.go
+++ b/cmd/ddev/cmd/debug-testcleanup.go
@@ -1,0 +1,51 @@
+package cmd
+
+import (
+	"strings"
+
+	"github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/output"
+	"github.com/ddev/ddev/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+// DebugTestCleanupCmd implements the ddev debug testcleanup command
+var DebugTestCleanupCmd = &cobra.Command{
+	Use:     "testcleanup",
+	Short:   "Cleanup diagnostic projects named with the tryddevproject prefix",
+	Example: "ddev debug testcleanup",
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 0 {
+			util.Failed("This command takes no additional arguments")
+		}
+		allProjects, err := ddevapp.GetProjects(false)
+		if err != nil {
+			util.Failed("failed getting GetProjects: %v", err)
+		}
+		if len(allProjects) < 1 {
+			output.UserOut.Println("No ddev projects were found.")
+			return
+		}
+		for _, project := range allProjects {
+			name := project.GetName()
+			if !strings.HasPrefix(project.GetName(), "tryddevproject") {
+				continue
+			}
+			output.UserOut.Printf("Running ddev delete -Oy %v", name)
+			err = DeleteCmd.Flags().Set("omit-snapshot", "true")
+			if err != nil {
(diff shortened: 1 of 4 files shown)


rfay:
I'm quite sure this is the right quick fix, but I'd sure like to just add a new command that was actually smart enough to do this. Just adding a `ddev debug testcleanup` would be a great thing and super easy.  Part of

* https://github.com/ddev/ddev/issues/3679

But probably the best part. It could be made quite safe too.

stasadev:
I added a new command `ddev debug testcleanup`.

rfay (commented):
This looks like exactly what I hoped for, thanks! It needs an entry in the docs command section as well.

rfay:
Stretch goal, but now that we have all this wonderful stuff, a test is possible in cmd/ddev/cmd!

rfay:
None of them are optimal! Thanks!

cmd/ddev/cmd/debug-testcleanup.go
```diff
@@ -0,0 +1,51 @@
...
+			output.UserOut.Println("No ddev projects were found.")
+			return
+		}
+		for _, project := range allProjects {
+			name := project.GetName()
+			if !strings.HasPrefix(name, "tryddevproject-") {
+				continue
+			}
+			output.UserOut.Printf("Running ddev delete -Oy %v", name)
+			err = DeleteCmd.Flags().Set("omit-snapshot", "true")
```
rfay, line 35: This is much easier and clearer with `project.Stop(true, true)`  (`ddev delete` was once `ddev stop --remove-data`)

rfay (changes requested):
Thanks so much for this! It's easier to delete as suggested in comment.

cmd/ddev/cmd/debug-testcleanup.go
stasadev: Perfect! One note: I used `project.Stop(true, false)`

PR #5151, opened 2023-07-17
refactor: add -t flag to mailhog call for symfony/mailer native transport compat, fixes #4363

rfay:
## The Issue

This PR replaces 
* #5138 by @rodrigoaguilera

Will make sure you get commit credit for it. I couldn't push the image spec into your PR.

* https://github.com/ddev/ddev/issues/4363

## How This PR Solves The Issue

Add a harmless -t to the mailhog call

## Automated Testing Overview

Mailhog integration is not tested AFAIK and soon will be replaced by mailpit

PR #5219, opened 2023-08-01
fix: remove unneeded php config from ddev-webserver, fixes #5092, replaces #5151, fixes #4363

### LINKED ISSUE #4363: Add compatibility between symfony mailer and php sendmail_path configuration for mailhog

When I configure symfony mailer and set its configuration to "native", trying to send a test email results in an exception from `symfony/mailer`
```
Unsupported sendmail command flags "/usr/local/bin/mailhog sendmail test@example.org --smtp-addr 127.0.0.1:1025"; must be one of "-bs" or "-t" but can include additional flags. 
```
Coming from https://github.com/symfony/mailer/blob/6.1/Transport/SendmailTransport.php#L58

I know there is overrides I can do in settings.php so symfony mailer uses the smtp transport on port 1025 but I would like for this to work out of the box without config overrides.

### Describe your solution

Add a `-t` non-functional flag to the command in the php.ini configurations so the exception is not raised.

### Describe alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #5092: Set correct serialize_precision for most PHP versions

I recently encountered an issue with an ini setting that wasn't set to the expected default for that PHP version. After digging a bit more and talking to @rfay on Discord, the conclusion was that the default config(s) probably haven't been updated for a long time.

Specifically it was the ini "serialize_precision" on PHP 8.0.24, that should default to "-1" from 7.1.0 onwards. DDEV sets that to 17 by default. https://www.php.net/manual/en/ini.core.php

A lot of subtle unexpected behavior can occur, if your production environment uses the current defaults while your local dev with DDEV has outdated settings.

### Describe your solution

The solution would be to go through all the configs and updated them to the latest defaults.

Maybe there is a way to automatically pull the defaults from the PHP repo? For example parsing https://github.com/php/php-src/blob/PHP-8.0.24/php.ini-production to generate the config file for 8.0.24 in DDEV.

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

* #5092
* Replaces #5151 
* #4363 

The configuration of various PHP versions and extensions was scattered all over the place and sometimes duplicated

* Move all PHP config into ddev-php-base image
* Use the default configuration for each PHP version by default, making changes only where they need to be made

- [x] Compare old configuration and new
- [ ] Test xdebug and xhprof
- [ ] Look for other risky areas
- [ ] Verify that mailhog works correctly on several PHP versions, follow #5151 testing to make sure things work there
- [x] Make sure php -r "var_dump(0.6);" works without weird floating point stuff
- [ ] `php -i` should show same config between v1.22.0 and this PR
- [ ] phpinfo(); should show same config between v1.22.0 and this PR

## Automated Testing Overview

- [ ] New TestPHPConfig should test at least for the floating point config issue

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5219"></a>


rfay:
I'm afraid this PR is too risky. Changing 1600 files means that something can absolutely go wrong. I'm very sad to let this go after a couple of days of solid work on it, and figuring out ways to mitigate the risk, but I think I shouldn't be doing a change like this. Future PHP versions will be done using default config by default, starting with PHP 8.3.

rfay:
Techniques that were used in this PR:

* Remove all PHP config from ddev-webserver so it's all coming from ddev-php-base
* Remove all config in ddev-php-base so it uses just what's installed by apt install
* Compare the config and edit all to do the few things that are really important:
  * Use the correct value for serialize_precision
  * Use the new value for sendmail_path (using -t)
  * Enable opcache and fix its settings
* Compare the resulting config by using `docker cp` from new and previous docker images. Looking at this with GoLand was pretty efffective (compare directories)

rfay:
Sadly closing in favor of
* https://github.com/ddev/ddev/pull/5231

PR #5221, opened 2023-08-02
build: update mkdocs version for readthedocs

rfay:
## The Issue

Received an email from readthedocs saying they're going to use newer versions of mkdocs and they want us to make sure we're using up-to-date as they're just going to now.

Update the requirement. It seems like we should already have been OK.

Email:

```
Hello,
You are receving this email because your Read the Docs project is impacted by an upcoming deprecation.

Historically, Read the Docs has installed MkDocs 0.17.3 by default if the project didn't define a specific version of it. Now, this version is a pretty old and we recommend our users to use the latest version of MkDocs: 1.5.1.

To avoid failing builds we kept the old default version of MkDocs. We are changing that behavior to always install the latest available version and we wanted to let you know about that so you can prepare your project.

Starting on October 3rd Read the Docs will install the latest available version of MkDocs. If you want to use a different version, make sure you are pinning your MkDocs dependency.

Here you have a pretty simple example of the section required on the .readthedocs.yaml configuration file:

python:
  install:
    - requirements: docs/requirements.txt
The content of docs/requirements.txt would be similar to:

mkdocs==1.5.1
We are sending you this email because you are a maintainer of the following projects that are affected by this removal:

[ddev](https://readthedocs.org/projects/ddev/)
Read more about this in our [Reproducible builds](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html) guide for more details.

Keep documenting,
Read the Docs
```

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5221"></a>


mattstein (approved):
I’m glad _you_ knew what to do with this, @rfay! Read the email and was trying to find a public announcement to link to or what needed changing.

It looks like it builds fine, so 🎉 !

rfay:
I think they weren't very careful with their email because we already had a newer version specified.

PR #5223, opened 2023-08-02
fix: correct xdebug port detection for colima on linux, fixes #5222

### LINKED ISSUE #5222: Add Colima support for Amd64 Linux

```
Failed to start test-site: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-test-site docker-compose -f /home/username/Sites/test-site/.ddev/.ddev-docker-compose-base.yaml config', action='[config]', err='exit status 15', stdout='', stderr='parsing /home/username/Sites/test-site/.ddev/.ddev-docker-compose-base.yaml: yaml: unmarshal errors:
  line 189: mapping key "extra_hosts" already defined at line 186' 
  ```
coming from https://github.com/ddev/ddev/blob/ef71d2537ec5fc8e6d9aa387a0474aa04cc9765c/pkg/ddevapp/app_compose_template.yaml#L228-L234

### Describe your solution

For one, change to `elseif` in https://github.com/ddev/ddev/blob/ef71d2537ec5fc8e6d9aa387a0474aa04cc9765c/pkg/ddevapp/app_compose_template.yaml#L228-L234 to fix the newly breaking issue.

Next, it would be great to officially support Colima on Linux.

### Describe alternatives

Using docker-ce

### Additional context

_No response_

rfay:
## The Issue

* #5222

There are two code problems;
* In the app_compose_template.yaml there are two "if" statements but it needs an else if instead to avoid duplicating in case of odd situations like this
* In config.go, UseHostDockerInternalExtraHosts should not be set if we're using Colima, when we use a static host.docker.internal.

Fix the two problems above.

- [x] Test it on Linux amd64 colima.
- [x] Test on macOS colima
- [x] Make sure Xdebug works both places

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5223"></a>

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -227,8 +227,7 @@ services:
 
       {{ if .HostDockerInternalIP }}
     extra_hosts: [ "host.docker.internal:{{ .HostDockerInternalIP }}" ]
-      {{ end }}
-      {{ if .UseHostDockerInternalExtraHosts }}
+      {{ else if .UseHostDockerInternalExtraHosts }}
     extra_hosts:
       - "host.docker.internal:host-gateway"
       {{ end }}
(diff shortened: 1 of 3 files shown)


bdanin (approved):
This works for me, thank you.

rfay:
I wasn't able to test this on Linux amd64 colima, but @bdanin did and said that it worked and xdebug worked. 

I tested on macOS arm64 colima and it seemed fine, and it passed tests.

PR #5226, opened 2023-08-03
Add WSL2 Debian install script

reloxx13:
## The Issue
no   
no   
./install_ddev_wsl2_docker_inside_debian.ps1 in a admin ps   
## Automated Testing Overview
no   

## Related Issue Link(s)
no   
## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5226"></a>


reloxx13:
> Could you please say why this is necessary? It seems to me like it would just add confusion.
> 
> And when you create a PR, please explain these things and fill in the requested information.

Hey,
because the default script only supports ubuntu wsl and the alternative manual installation is long.

rfay:
Do you think it would be reasonable to do a script for every Linux distro then? 

How about looking at the existing script and seeing if it can be made more generous about installing to the default distro in some cases?

reloxx13:
Not worth to discuss this further.

PR #5227, opened 2023-08-03
feat: bashrc to support very early bashrc additions that can work for edge cases like sshd

hanoii:
## The Issue

Having worked on https://github.com/hanoii/ddev-sshd I was looking for a way to force the landing directory of ssh to be `/var/www/html`. The logical place to do this is on `.bash_profile` or `.bashrc`. I [did it](https://github.com/hanoii/ddev-sshd/blob/94cb0c50ae2c5f5599c529b0185d25b9143bebc6/homeadditions/.bashrc.d/sshd) by adding a file to `.ddev/homeadditions/.bashrc.d` that does exactly that.

This works when I ssh to the host, but not if I run it with a command:

`ssh ddev-pr-5227.ddev.site ls` list the home directory.
`ddev exec ls` lists the `/var/www/html` directory.

I wanted both to behave the same.

I did some research into the different bash execution scripts and the reason for this happening is:

https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bash-Startup-Files

> **Invoked by remote shell daemon**
>
> Bash attempts to determine when it is being run with its standard input connected to a network connection, as when executed by the historical remote shell daemon, usually rshd, or the secure shell daemon sshd. If Bash determines it is being run non-interactively in this fashion, it reads and executes commands from ~/.bashrc, if that file exists and is readable. It will not do this if invoked as sh. The --norc option may be used to inhibit this behavior, and the --rcfile option may be used to force another file to be read, but neither rshd nor sshd generally invoke the shell with those options or allow them to be specified.

The default skeleton file (on ddev which is very close to debian's one) [has a very early check](https://github.com/ddev/ddev/blob/ce8f23b3a83b627a08bb30938c1cbfd70470366b/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc#L5-L9) that looks to see if the shell is interactive, and if so, continue or halt bashrc executing.

I created a test sandbox repo for this at https://github.com/hanoii/ddev-pr-5227. Besides the add on, it [adds](https://github.com/hanoii/ddev-pr-5227/blob/main/.ddev/web-build/Dockerfile) some `echo`s  to the bash start up file:

- `ddev ssh`
```
!! /etc/bash.nointeractive.bashrc
!! /etc/bash.bashrc - all
!! /etc/bash.bashrc - interactive
!! bash_profile
```

- `ddev exec ls`
```
!! /etc/bash.nointeractive.bashrc
```

- `ssh ddev-pr-5227.ddev.site`
```
!! /etc/bash.bashrc - all
!! /etc/bash.bashrc - interactive
!! bash_profile
```

- `ssh ddev-pr-5227.ddev.site ls`
```
!! /etc/bash.bashrc - all
!! home .bashrc-all
```

Where:
- `bash.bashrc - all`: is in the beginning of the bashrc 
- `bash.bashrc - interactive` is after the interactive check.

It allows addons to add bashrc additions that are both run in interactive and remote shells.

## Automated Testing Overview

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5227"></a>

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.nointeractive.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/bash.nointeractive.bashrc
@@ -3,6 +3,10 @@ for f in /etc/bashrc/*.bashrc; do
   source $f;
 done
 
+for i in $(\ls $HOME/.bashrc.pre.d/* 2>/dev/null); do
+    source $i;
+done
+
 for i in $(\ls $HOME/.bashrc.d/* 2>/dev/null); do
     source $i;
 done
--- a/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
@@ -2,6 +2,8 @@
 # see /usr/share/doc/bash/examples/startup-files (in the package bash-doc)
 # for examples
 
+for i in $(\ls $HOME/.bashrc.pre.d/* 2>/dev/null); do source $i; done
+
 # If not running interactively, don't do anything
 case $- in
     *i*) ;;


hanoii:
@rfay I try adding as much information as possible as well as a test sandbox repo. Hope this helps. I can't use `.bash_profile` because of what I explained and showcased in the description

rfay:
> I can't use .bash_profile because of what I explained and showcased in the description

Sorry, not sure why you can't just replace the .bash_profile in Dockerfile or post-start hook or web_entrypoint.d. Can you explain that?

hanoii:
@rfay `bash_profile` si not being sourced when running a command through ssh

`ssh host command`

hanoii:
I can obviously modify the skeleton file on a Dockerfile, I just felt that doing that was a little bit too intrusive for an add-on

rfay:
I didn't remember that this was about an add-on, but still, not sure why you can't just change the .bash_profile itself in an add-on by adding a .ddev/web-build/Dockerfile.ddev-sshd or whatever.

hanoii:
I can do a lot of different things, I don't have a problem making this work, I maybe overthought this and wanted a way for ddev to let me do it with the minimum changes necessary. Same as the pre dockerfile and other hooks, just another point of extension, like a drupal hook.

hanoii:
> I didn't remember that this was about an add-on, but still, not sure why you can't just change the .bash_profile itself in an add-on by adding a .ddev/web-build/Dockerfile.ddev-sshd or whatever.

As per the description:
- ` ssh ddev-pr-5227.ddev.site ls`
```
!! /etc/bash.bashrc - all
!! home .bashrc-all
```

The above are `echo`s that I added to the sandbox repo, there's on in .bash_profile too. .bash_profile is not being run.

The only thing run is /etc/bash.bashrc and the ~/.bashrc, but both files have a check for interactive shells, which then returns because running a remote command through ssh this is not interactive.

rfay:
Sorry, guess we're lost in a loop. My question is why you don't change the .bash_profile to do what you've proposed here.

hanoii:
Agree on the loop, I can showcase it to you if you think it's useful. `.bash_profile` is never run in this particular situation. I thought the PR was very well documented now, including quoting bash manual page where this is all explained. Maybe I am missing something.

We can otherwise close this PR and I'll find a way around it.

hanoii:
@rfay to try to close the loop, I did [fix my add-on](https://github.com/hanoii/ddev-sshd/blob/85cedf35445f4d78d39a71d10f47b4374cb1d9bd/web-build/Dockerfile.sshd#L3)  by modifying `~/.bashrc` and prepending first of all one `.bashrc.sshd` that the add-on also provides, thus, doing what I wanted which is that `ssh HOST and ssh HOST COMMAND` behaves similar to `ddev ssh` and `ddev exec COMMAND`. as both land on `/var/www/html`.

My intention here was only to provide a mean to do this without having to do any modification on the image main files and giving ddev an extra point of extension that my add-on (and potentially others) could use. If you see my code changes they are very simple and harmless.

I understand though that adding every single feature that everybody wants is not the way to go, and I do have a reasonable filter into what I suggest. I could have done this on the add-on but went this more time consuming path only because i thought it was a good addition to ddev, not for fixing my specific problem, but I know you know this.

You can close this PR.

rfay:
Thanks for working this through. Yeah, we already have so many obscure features that people don't actually use, and experts like you already have many tools to solve things like this. I'll go ahead and close, but maybe we'll be reopening one day.

PR #5231, opened 2023-08-03
fix: php serialize_precision should be -1 in php >= 7.1, fixes #5092

### LINKED ISSUE #5092: Set correct serialize_precision for most PHP versions

rfay:
## The Issue

* #5092 

Change the affected `serialize_precision` values

* `ddev ssh` and `php -r "var_dump(0.6);"` - should show float(0.6)

## Automated Testing Overview

Added TestPHPConfig

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5231"></a>


hanoii:
Looks good to me, I downloaded the binary from this PR and tested against https://github.com/hanoii/ddev-pr-5227

With v1.22 I get:

```
ddev exec 'php -r "var_dump((float) 0.6);"'
float(0.59999999999999998)
```

With the binary from this PR I get:

```
ddev exec 'php -r "var_dump((float) 0.6);"'
float(0.6)
```

This on PHP 8.1 - I haven't tested any other PHP version

hanoii:
I went ahead and tried 7.4 and 7.0, output was fine on both

```
ddev exec 'php -r "var_dump((float) 0.6);"'
float(0.6)
```

pkg/ddevapp/config_test.go
```diff
@@ -869,6 +869,59 @@ func TestPHPOverrides(t *testing.T) {
...
+
+		out, _, err := app.Exec(&ExecOpts{
+			Cmd: "php --version",
+		})
+		require.NoError(t, err)
+		t.Logf("============= PHP version=%s ================", out)
+
+		// Look for problems with serialize_precision,https://github.com/ddev/ddev/issues/5092
+		out, _, err = app.Exec(&ExecOpts{
+			Cmd: `php -r "var_dump(0.6);"`,
```
hanoii, line 914: You added my reproduction step on the test, nice :)

hanoii (approved):
All good codewise too!

PR #5234, opened 2023-08-04
fix: import-files should work with any project type

stasadev:
## The Issue

`ddev import-files` should work for all project types, including `silverstripe`.

`importFilesAction` can be easily forgotten for newly created project types.

Let's use `genericImportFilesAction` if no specific `importFilesAction` is defined.

```
mkdir silverstripe && cd silverstripe && mkdir source && touch source/file
ddev config --project-type=silverstripe
```

Before:

```
ddev import-files --source=source --target=target 
Error: failed to import files for silverstripe: this project type (silverstripe) does not support import-files
```

After:

```
ddev import-files --source=source --target=target 
Successfully imported files for silverstripe
```

## Related Issue Link(s)

- https://github.com/ddev/ddev/pull/5119

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5234"></a>

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -98,7 +98,6 @@ func init() {
 			configOverrideAction: django4ConfigOverrideAction,
 			postConfigAction:     django4PostConfigAction,
 			postStartAction:      django4PostStartAction,
-			importFilesAction:    genericImportFilesAction,
 		},
 
 		nodeps.AppTypeDrupal6: {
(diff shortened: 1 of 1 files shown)


Firesphere:
Interesting I hadn't encountered it while adding Silverstripe as a project, nor it being picked up by the tests.

It's a feature that's not often used I have come to believe, by the looks of it.

Nice catch there. I'll have a look if I can combine this with `sspak` which is Silverstripe's "import export" mechanism, to make it fully work.

rfay:
Rebased.

rfay (approved):
This is great, thanks! I'll rebase and then pull after tests are good.

PR #5237, opened 2023-08-04
fix: drupal10.1 now complains about trusted_hosts_patterns even with general regex

rfay:
## The Issue

Drupal 10.1 is whining about the trusted_host_patterns even though we have always had one.

Add more to the patterns.

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5237"></a>

### DIFF
--- a/pkg/ddevapp/drupal/drupal10/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal10/settings.ddev.php
@@ -32,7 +32,11 @@ $settings['skip_permissions_hardening'] = TRUE;
 
 // This will ensure the site can only be accessed through the intended host
 // names. Additional host patterns can be added for custom configurations.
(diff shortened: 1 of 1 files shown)


tyler36:
I'm not seeing a warning on Status page with Drupal `10.1.0` & DDEV `1.22.1`.
It seems happy with the settings:
drupal-trusted-hosts

If I use [4db441a](https://github.com/ddev/ddev/pull/5237/commits/4db441a78d80ffb5697acb15cc7608a5c58fe42b), the site has more "trusted hosts" but the same number of warnings & errors.
drupal-trusted-hosts-2

rfay:
Thanks for checking this so carefully @tyler36 - I'm sure I had the warnings in 10.1.1. But I definitely don't have them in 10.1.2. I'll close this and move on for now.

PR #5242, opened 2023-08-06
fix: additional MySQL config on Windows, fixes #3801

### LINKED ISSUE #3801: Mysql overrides don't work on traditional Windows

_No response_

### Current Behavior

On Windows, as reported by Jon McLaughlin in https://discord.com/channels/664580571770388500/961636249985237103 

* Mysql overrides in `.ddev/mysql/*.cnf` don't work.

```
Ensuring write permissions for servis2
mysql: [Warning] World-writable config file '/mnt/ddev_config/mysql/group_concat_max_length.cnf' is ignored.
mysql: [Warning] World-writable config file '/mnt/ddev_config/mysql/local_infile.cnf' is ignored.
```

This is a result of trying to use the mounted Windows directory directly (/mnt/ddev-config), for which permissions are not in our control.

DDEV 1.17.0-alpha6 seems to have introduced this; prior to that, the config files were copied into /etc/mysql in the container to solve exactly this problem.  The problem originates with the removal of this line, https://github.com/drud/ddev/pull/2806/files#diff-db674dfd9698544c77815b1df958a12fb4d5b4a2a1ec2b3f167cf30c44be901dL50

(and of course, directly including /mnt/ddev_config/mysql)

### Expected Behavior

You should be able to use mysql overrides on Windows. 

### Steps To Reproduce

On windows, use any mysql override.

### Anything else?

Proposed workaround from the Discord channel is to install the needed overrides using a custom .ddev/db-build/Dockerfile:

```
ARG BASE_IMAGE
FROM $BASE_IMAGE

# MySQL customizations.
RUN echo -e "[mysqld]\n group_concat_max_len = 4096" > /etc/mysql/conf.d/group_concat_max_length.cnf
RUN echo -e "[mysqld]\n local_infile = true" > /etc/mysql/conf.d/local_infile.cnf
```

rokclimb15:
## The Issue

On Windows, the documented steps for additional MySQL configuration do not work. See the filed issue for details.

Reverts to an earlier version of this code which copies the config files and adjusts permissions instead of including them directly from the host mount.

Test deployment of MySQL override on supported operating systems

## Automated testing

- [x] Add TestMysqlConfigOverride to verify config override in general
- [ ] Manually run TestMysqlConfigOverride on traditional Windows, since it doesn't get exercised by PR tests. 

## Related Issue Link(s)

* https://github.com/ddev/ddev/issues/3801

## Release/Deployment Notes

MySQL config overrides now work as expected on Windows

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5242"></a>


rfay:
Great work. I think it's important to add test coverage for this so we don't slip back here again.  Looking around, the closest we currently come is https://github.com/ddev/ddev/blob/7aa50e47e108d09adf8b4fcaf87aa71974558199/pkg/ddevapp/config_test.go#L790, which isn't close enough. So I think a TestMysqlConfigOverride is probably in order here. 

Oh, I guess we have it in the container test, https://github.com/ddev/ddev/blob/6302f01c85c4e1c5bce9372353e4aa114eef81b8/containers/ddev-dbserver/test/custom_config.bats - but we don't run that on Windows over the last few years or it would have been caught. 

Thanks for taking this on!

rokclimb15:
@rfay what do you think is the best way forward here? It sounds like running the existing container tests on Windows might prevent this from regressing? How can I contribute to that work?

rfay:
I'd rather have a Go test that tests overrides; Windows is just too slow to run a lot of things on. The go tests are now run only on master-branch commits for that reason, because they take more than 5 hours to run. But at least we run them.  

TestConfigOverrideDetection isn't a bad model, it just needs to be focused on the actual override happening. 

On Tuesday testing will be our topic if you can join us, https://ddev.com/blog/contributor-training/

rfay:
However, it doesn't look like this is working at this time, it's failing the old container test, see https://github.com/ddev/ddev/actions/runs/5778803537/job/15660423482?pr=5242

Did you manually test?

rokclimb15:
@rfay the https://github.com/ddev/ddev/blob/master/docs/content/developers/building-contributing.md#testing-a-pr section says the PR build creates artifacts you can download for testing. That was my plan. Where can I find those?

rfay:
Because you only changed one of the image builds (and no go files) it didn't build any artifacts for you for you, and it wouldn't have helped if it did. 

For you to test what you've done you'll have to 

* Build an image. let's call it 20230806_mysql_config_windows. `cd containers/ddev-dbserver && make VERSION=20230806_mysql_config_windows`
* In versionconstants.go, change the db image base to that.
* In project root, `make`

You probably won't want to build all the ddev-dbserver because it takes forever, take a look at the Makefile for how to build just the one you're most interested in.

If you get stuck building the image I'll be happy to push an image build up for you.

rokclimb15:
@rfay I was able to get a build environment set up under Linux with my branch, build the image and apply a mysql config override without issue. I'm puzzled as to why the tests are failing. Can you help me understand a bit more about the failure?

I was not able to cross-compile for Windows due to a missing `EnVar.dll`.

```
bmorton@LAPTOP-78EIRP0U:~/test$ ../ddev-bin version
 ITEM             VALUE
 DDEV version     54eb124c-dirty
 architecture     amd64
 db               ddev/ddev-dbserver-mariadb-10.4:20230806_mysql_config_windows
 ddev-ssh-agent   ddev/ddev-ssh-agent:v1.22.0
 docker           24.0.2
 docker-compose   v2.20.0
 docker-platform  wsl2-docker-ce
 mutagen          0.17.2
 os               linux
 router           traefik:v2.10
 web              ddev/ddev-webserver:20230803_php_serialize_precision
```

```
bmorton@LAPTOP-78EIRP0U:~/test$ ../ddev-bin mysql

mysql> select @@sql_mode;
+--------------+
| @@sql_mode   |
+--------------+
| IGNORE_SPACE |
+--------------+
1 row in set (0.00 sec)
```

rfay:
Since this is your first PR I have to approve test runs, so ping me when you do a push.

rokclimb15:
@rfay container tests pass locally on WSL2. I copied my cross-OS compile for Windows and it doesn't fail, but doesn't seem to pick up the override changes either. Very strange.

rfay:
You'll want to inspect your image and see if it's actually doing what you want. Your little script can have `set -x` added to it, and you can `docker run -it --entrypoint=bash <imagespec> ` to go inside the image and see what's there, make sure you're running what you want.

rokclimb15:
@rfay thanks for the tips! I did that, and I see that the `ddev_config` dir seems to not be mounted to the host.

```
root@8eacf6f283c8:/# ls -l /mnt/ddev_config/mysql
total 0

root@8eacf6f283c8:/# mount | grep mnt
root@8eacf6f283c8:/#
```

rfay:
When you do `docker run` without explicitly mounting, you won't have /mnt/ddev_config. If you want to experiment with that, `docker run -it --entrypoint=bash -v <path-to-.ddev>:/mnt/ddev_config <imagespec>`

rokclimb15:
@rfay I'm assuming the `docker_entrypoint.sh` script is intended to run as root, correct? When I run it via bash in the container the last few lines are:

```
2023-08-09T01:40:52.016072Z 0 [ERROR] Fatal error: Please read "Security" section of the manual to find out how to run mysqld as root!

2023-08-09T01:40:52.016102Z 0 [ERROR] Aborting
```

This happens in the 1.22 release container as well, so I think I'm doing something wrong. The file copy and `chmod` works fine, everything appears to be in order with the change so far.

rfay:
No, everything in the key DDEV containers runs as the same UID as the user. You can `ddev ssh -s db` to see what goes on in there.

So if you want to do more than just examine the contents of the image, run it with

```
docker run -it --entrypoint=bash  -u www-data:www-data -v <path-to-.ddev>:/mnt/ddev_config <imagespec>
```

`www-data` here is just an available user that can be used. DDEV creates a user that matches your host-side user and adds a layer to do that.

rfay:
There are a lot of moving pieces here. I'm happy to do more screenshare conversations with you.

rokclimb15:
@rfay I think I've sorted out the problem. Would you mind initiating another test run while I work on a Go-based test?

rfay:
Yay, test is running, but the image will need to be pushed. Do you want to push it to your account and change versionconstants.go to reflect the new (temporary) image location, or would you like me to push to dockerhub and add a commit for that?

rokclimb15:
@rfay would you mind doing that?

rfay:
Rebased; images pushed; versionconstants.go updated.

containers/ddev-dbserver/files/docker-entrypoint.sh
```diff
@@ -74,7 +74,9 @@ rm -f /tmp/raw_mysql_version.txt
 
 # If we have extra mariadb cnf files, copy them to where they go.
 if [ -d /mnt/ddev_config/mysql ] && [ "$(echo /mnt/ddev_config/mysql/*.cnf)" != "/mnt/ddev_config/mysql/*.cnf" ] ; then
-  echo "!includedir /mnt/ddev_config/mysql" >/etc/mysql/conf.d/ddev.cnf
+  cp /mnt/ddev_config/mysql/*.cnf /etc/mysql/conf.d
+  # Ignore errors on files such as .gitmanaged
+  chmod -f -R ugo-w /etc/mysql/conf.d/*
```
rfay, line 79: I wonder about the problems you had earlier. .gitmanaged wouldn't have been found by this chmod...

rfay:
Next steps are:
* manual testing on traditional Windows
* Mysql override test, perhaps based on https://github.com/ddev/ddev/blob/7aa50e47e108d09adf8b4fcaf87aa71974558199/pkg/ddevapp/config_test.go#L740 but actually making sure the content works out right.

rokclimb15:
@rfay I think the additional assertion might belong in the current test since it's checking to ensure that overrides work. The additional assertion might need to connect to the DB container via mysql shell and run some SQL to validate an override before shutting down the app. The current test just checks error log output to ensure the files were copied.

I can come up with the setting and the SQL to run, but do you know how to run a mysql shell from this Go-based test runner? That's the bit I'm missing currently. TIA!

rfay:
Yes, that style of app.Exec in the db container will work fine.

rokclimb15:
@rfay I think this one is ready for review.

rfay (changes requested):
The test looks pretty good, thanks! I wonder if we should try it on more than the default database type (currently mariadb 10.4). Some of the other tests (like TestDdevAllDatabases) do a whole selection of them with GOTEST_SHORT is not set, or a tiny selection when it is set. Here we could just do mariadb 10.4 and mysql 8.0 when GOTEST_SHORT is set, but perhaps (almost?) all when it is set

What do you think?

rfay:
I just realized/remembered that we only run the traditional Windows tests on pull (commits on master). So we'll have to make sure that this passes the test locally on Windows. I'll add that to the OP.

rfay:
I guess an option would have been to add this logic to the TestDdevAllDatabases and then it would have been just a couple of lines added. I should have thought of that earlier. This is going to be great though, and it's very specific.

rfay:
Are you able to run individual tests locally OK? https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing has info, and we recently did a contributor training about running tests locally, recording on https://ddev.com/blog/contributor-training/

rokclimb15:
@rfay yes I am. It looks like the Colima tests are failing for an unrelated reason.

rfay:
Yes, the Github hosted runners are not all the same and colima isn't properly set up on some. Don't worry about that. 

However, if you PM me your email I'll add you to the buildkite setup so you can restart tests

rfay:
Looking great! I'll try it out today.

rfay (approved):
Wow thanks! This looks good to me. I tested manually on traditional Windows using the example in the [docs](https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/#custom-mysqlmariadb-configuration-mycnf). I also manually ran the TestDdevAllDatabases test (with GOTEST_SHORT=true). And of course the automated tests have succeeded, we can hope they'll succeed on pull. Congratulations, and thanks for all the hard work!

PR #5244, opened 2023-08-07
feat: add zstd support for import-files, import-db, export-db

kringkaste:
## The Issue

Missing support for Zstandard (zstd) compression for `ddev import-db`, `ddev export-db`, `ddev import-files`

Add support for Zstandard compression.

e.g. `ddev import-db --file=archive.sql.zst`

## Automated Testing Overview

Added the new compression format to the existing tests of various other compression formats.

## Release/Deployment Notes

None.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5244"></a>


pkg/ddevapp/ddevapp_test.go
```diff
@@ -2411,7 +2415,7 @@ func TestDdevImportFiles(t *testing.T) {
 
 		if app.GetUploadDir() != "" {
 			// Import trivial archive with various types of archive/compression and verify that files arrive
-			extensions := []string{"tgz", "zip", "tar.xz", "tar.bz2", "tar.gz"}
+			extensions := []string{"tgz", "zip", "tar.xz", "tar.bz2", "tar.zst", "tar.gz"}
```
rfay, line 2418: `tar. zst` isn't going to work is it?

rfay (commented):
I guess I thought the file extension was `ztsd` not `zts` ?

rfay:
https://en.wikipedia.org/wiki/Zstd says it's .zts :)

rfay:
It looks like you didn't do `go mod tidy`, please follow up with that.

kringkaste:
> https://en.wikipedia.org/wiki/Zstd says it's .zts :)

You mean `.zst` not `.zts` right?

rfay:
Sorry, its always been backwards in my fingers. I wonder how many times I've messed that up in the past.

mattstein (approved):

rfay:
Since this is your first PR I have to approve test runs, so ping me when you push new stuff, thanks!

rfay (changes requested):
This is looking great, and a great contribution, thanks.

I find that `ddev export-db --zst` doesn't output a zstd file though, it outputs a gzipped file. Please check that (and the other uses) and make sure they're all working. A little test coverage may be useful to make sure we don't have regressions in this area.

```
$ ddev export-db --zstd --file=db.sql.zst
Wrote database dump from project 'd10' database 'db' to file db.sql.zst in gzip format.
rfay@rfay-tag1-m1:~/workspace/d10/.tarballs$ file *.zst
db.sql.zst: gzip compressed data, from Unix, original size modulo 2^32 12219647
```

kringkaste:
Sorry @rfay but I'm getting stalled here. To make zstd compression available for export-db, the zstd package have be added to the db-server images. But the base image for mariadb 5.5 is so old, the package is not available in the used distribution.  All other db images can be build.

Do we find a solution for that? If so, I would add test coverage.

rfay:
It's OK to exclude it for mariadb 5.5. We have to do that for postgres:9, it can't do snapshots. You might have to make an "if can't then error" 

Thanks for the great work!

rfay:
It's also possible to do the compression work on the host/golang side. But no need to worry about that. 

We'll need new images if you're changing those though, you can push your own to your own account temporarily or I can push some up when you're ready. Don't forget about postgres.

rfay:
I also wonder if anybody EVER uses the old mysql and mariadb versions. They're there to help revive old websites, but wow, they are really getting old now. They weren't so old 7 years ago.

rfay:
Hoping you'll be back to this before it gets so stale it's unrecoverable @kringkaste . You've done great work on it.

rfay:
Happy to help get this to completion, but if it's not active before about Sept 15, I'll close it. That doesn't mean we can't reopen it, etc, but it needs to not be cluttering the queue if it's not making progress.

rfay:
Thanks for all the work on this. If you get back to it I'm happy to help and we can re-open it. Closing for now.

PR #5259, opened 2023-08-14
ci: colima no longer needs to be installed on macOS runners on github actions

rfay:
## The Issue

Colima is now pre-installed on macOS GitHub runners, so we don't need to install it.

Stop adding it.

Thanks to @joffrey-bion in 

* https://github.com/actions/runner/issues/1456#issuecomment-1676495453
* https://github.com/actions/runner/issues/1456#issuecomment-1676497504

See also
* https://github.com/actions/runner-images/issues/6216
* https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5259"></a>


dluc:
is it installed?

```
jobs:
  build-and-push-image:
    runs-on: macos-latest
    steps:

      - name: Setup docker (missing on MacOS)
        if: runner.os == 'macos'
        run: |
          brew install docker
          colima start
```

> 2024-09-30T20:28:58.0726300Z /Users/runner/work/_temp/df344a65-b5d4-4784-98d8-532409faa44c.sh: line 2: colima: command not found

rfay:
We haven't been testing Colima on GitHub actions for a long time. It was so underpowered and so inconsistent that we moved to running with real arm64 computers locally using buildkite. 

I'm confused about what you're doing then... Are you trying to use the colima.yml in some other context?

dluc:
The end goal was building a multi-arch docker image with GitHub actions. Eventually I found that ARM support on GitHub actions is incomplete, without going with self-hosted runners. 
I've since split the image in 2 labels (ugly), one for adm64 and one for arm64. GitHub actions builds and published the amd one, and for now I build the arm64 manually on a mac, because the github actions fails in many different ways (ssh not starting for example)

Some posts I was following, trying to make it all work:
https://github.com/actions/runner/issues/1456
https://github.com/abiosoft/colima/issues/684

joffrey-bion:
@dluc you won't have it on `macos-latest` because these images don't support nested virtualization AFAIK. You'll need `macos-13` or maybe even `macos-12` images if you want to use docker at all in any form in GitHub Actions mac runners.

rfay:
If your goal is building multi-arch images, you can do it very well with the technique we use, please take a look at https://github.com/ddev/ddev/blob/master/.github/workflows/push-tagged-image.yml

It's not hard. 

Of course you need the support we use in the Makefile as well, and you need `docker buildx create` to have the buildx builder you need.

rfay:
The basics of multi-arch image building are in this blog and the newer docs: https://www.docker.com/blog/how-to-rapidly-build-multi-architecture-images-with-buildx/

You don't need a mac runner at all, it all works great on Linux using the regular ubuntu-latest runners

dluc:
I've tried multiple ways so far, initially using just ubuntu-latest. The app I'm building is based on .NET, not sure if that's part of the problem: the final result was always a multi-arch image where the ARM64 layer contained AMD64 files, leading to segmentation faults. Thanks for the resources, I'll check them out once I get more time to work on this (for now I'm going with two separate builds).

PR #5262, opened 2023-08-15
fix: provide consistent hash per drupal/backdrop project name, fixes #5255

### LINKED ISSUE #5255: `ddev start` always creates a new hash_salt in settings.ddev.php (Drupal); should be static.
### Expected Behavior

Consistent hash_salt or none at all, not sure.

### Actual Behavior

Whenever ddev builds the containers, they create a new hash_salt every time. 

This creates [issues with the token values on Drupal contextual links](https://www.drupal.org/project/drupal/issues/3010562), as the hash_salt value is [checked here](https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/contextual/src/ContextualController.php#L77). which leads to this issue.

You can reproduce this by opening the `settings.ddev.php` file, doing `ddev poweroff`, and then `ddev start` to view it change in real-time.

So a temporary fix is to place your hash_salt value AFTER the $ddev_settings section that gets added by DDEV in settings.php. This will allow that value to persist despite the new hash salt value that's created everytime DDEV starts up.

```
// Automatically generated include for settings managed by ddev.
$ddev_settings = dirname(__FILE__) . '/settings.ddev.php';
if (getenv('IS_DDEV_PROJECT') == 'true' && is_readable($ddev_settings)) {
  require $ddev_settings;
}

$settings['hash_salt'] = file_get_contents('/home/example/salt.txt');
```

You could also just move your local settings below the ddev settings, and put the hash_salt value in settings.local.php instead.

rfay:
## The Issue

* #5255

Creating a new hash on restart can cause a number of Drupal/Backdrop functions to misbehave

The hash is now calculated based on the project name. This is inappropriate for production, but should work great for developer situations. It also means that the hash will be the same for all developers on a team.

`ddev start` multiple times and review the `settings.ddev.php` for the project. It should not change, but should be different for different projects.

## Automated Testing Overview

Added TestDrupalBackdropConsistentHash

### DIFF
--- a/pkg/ddevapp/backdrop.go
+++ b/pkg/ddevapp/backdrop.go
@@ -43,7 +43,7 @@ func NewBackdropSettings(app *DdevApp) *BackdropSettings {
 		DatabaseHost:     "ddev-" + app.Name + "-db",
 		DatabaseDriver:   "mysql",
 		DatabasePort:     GetExposedPort(app, "db"),
-		HashSalt:         util.RandString(64),
+		HashSalt:         util.HashSalt(app.Name),
 		Signature:        nodeps.DdevFileSignature,
 		SiteSettings:     "settings.php",
 		SiteSettingsDdev: "settings.ddev.php",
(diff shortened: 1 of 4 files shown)


TonyTheFerg:
I applied the fix, and it seemed to work well in my environment. Thanks @rfay!

PR #5272, opened 2023-08-17
ci: colima should use macos-13

rfay:
## The Issue

colima tests have been using macos-latest, which is still macos 12. But github has macos-13 (beta) out there.

Use the newer one.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5272"></a>


rfay:
I guess they haven't added colima to macos-13 yet.

PR #5278, opened 2023-08-19
build: experimental Debian 12 Bookworm for ddev-webserver

rfay:
## The Issue

Debian 12 Bookworm is out and we'd like to use it when it's mature and all packages are available

Use bookworm as base

## TODO and problems

- [ ] php*-xhprof and php*-xmlrpc packages are missing on all php versions on arm64 - https://github.com/oerdnj/deb.sury.org/issues/1977
- [x] php*-ssh2 is missing for most PHP versions in deb.sury.org repos, at least on arm64. It's there for php8.2, maybe that only.
- [x] We're using /etc/apt/trusted.gpg.d for nginx and blackfire keys, and that's not where they belong any more. (fixed elsewhere)
- [x] This script, located at https://deb.nodesource.com/setup_X, used to install Node.js is deprecated now and will eventually be made inactive. Go to https://github.com/nodesource/distributions (fixed elsewhere)
- [x] Work with upstream to get php*-xhprof and php*-xmlrpc and php*-ssh2
- [x] There may be other extensions that people are using that are MIA

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5278"></a>

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -87,21 +87,21 @@ RUN set -eu -o pipefail && LATEST=$(curl -L --fail --silent "https://api.github.
 
 # The number of permutations of php packages available on each architecture because
 # too much to handle, so has been codified here instead of in obscure logic
-ENV php56_amd64="apcu bcmath bz2 curl cli common fpm gd imagick intl json ldap mbstring mcrypt memcached mysql opcache pgsql readline redis soap sqlite3 uploadprogress xdebug xhprof xml xmlrpc zip"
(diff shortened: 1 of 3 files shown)


rfay:
* https://github.com/oerdnj/deb.sury.org/issues/1977#issuecomment-1685131922

rfay:
I don't expect this to be resolvable for some time. deb.sury.org doesn't seem to be getting the new stuff right at all yet. I'm going to close for now, can reopen later. I would like this to be in v1.23.0, we'll see what happens.

PR #5281, opened 2023-08-20
feat: initial JSON schemas for global config and project/app config, fixes #4511

### LINKED ISSUE #4511: Providing DDEV configuration yaml schema dynamically for PHPStorm (ddev-intellij-plugin)

I recently noticed that the configuration yaml schema within the PHPStorm plugin was identifying certain valid configuration values as not recognised or invalid, in my case the project value of `craftcms` https://github.com/drud/ddev/issues/4507 https://github.com/php-perfect/ddev-intellij-plugin/issues/141

It has been determined that the PHPStorm plugin currently has a [static JSON schema definition based on DDEV version 1.19](https://github.com/php-perfect/ddev-intellij-plugin/blob/main/src/main/resources/schema/ddev-config-1.19.schema.json) which predates the addition of this value. Because the schema definition is out of date, this would be why it triggers the PHPStorm IDE to flag values, because it doesn't know they exist. The solution is to update the schema, but it probably goes a bit further on making this dynamic and less fragile going forward.

Speaking with @nico-loeber, it's become apparent that static JSON schemas are problematic for several reasons:

1. They'll likely be out of date often if remaining static in the current implementation.
2. It would require publishing plugin updates for each schema version change.
3. Sometimes, configuration option changes have occurred in incremental releases e.g. the `craftcms` project value was added in 1.21.2, rather than a major release, so if the schema remained static, you'd also have to create schema definitions for incremental releases too, which adds more schema files to be maintained.

Because of the highlighted areas, it would seem more appropriate if the schema lived in the main DDEV repository which the PHPStorm plugin would then pick up on each release.

@nico-loeber has already outlined a potential plan around this: https://github.com/php-perfect/ddev-intellij-plugin/issues/141#issuecomment-1368792258. However the main requirement for this to work would be for a schema definition to exist in the main DDEV repository and be maintained on any changes.

Therefore, I wanted to open up the discussion again on agreeing a potential implementation that works between the two projects while being dynamic.

### Describe your solution

A configuration schema definition to be added and maintained within the main DDEV project which the PHPStorm plugin can be configured to parse the correct release package from GitHub to match against the currently installed DDEV version to pull in the schema to load in the PHPStorm IDE in order provide the correct configuration options, without static schema references or out of date information.

How this is done, I think is important to discuss. Moving the schema into the main project, is part of the solution, but how the schema file is maintained or generated is another. Perhaps this can be generated based off the configuration values within the main DDEV code somehow with an internal command like `ddev schema` to then output the JSON definition format (just as an example).

### Describe alternatives

_No response_

### Additional context

_No response_

jamesmacwhite:
## The Issue

Having IDEs automatically validate the global_config.yaml or config.yaml is useful to help guide valid changes within a specification. This also allows DDEV plugins to leverage a schema file that is maintained in the project, allowing for updates across versions.

This scenario was born from the fact the JetBrains Intellij plugin for DDEV written by @nico-loeber was implementing it's own JSON Schema but it was based on 1.19 and had become outdated. By doing this, we allow other plugins to use a single source of truth to prevent issues with out date options or missing options e.g. project types. Any changes required can be directly submitted to the ddev repository, maintaining a single source of truth,

This PR provides two JSON schema files which have been validated against test files for the global_config.yml and config.yaml respectively. This is part one of having the schemas included in the main DDEV repository, to then get the schema linked in the schemastore once we have a URL reference.

While the schemas pass validation from the test cases, this is the initial version and there may be other improvements possible, I welcome anyone else with experience with JSON Schemas, given this is my first adventure into them!

Due to DDEV configuration options/features changing, hosting this within DDEV itself is likely the best solution to allow for rapid updates and changes when needed.

## Automated Testing Overview

The two schemas have been tested against a sample global_config.yaml and config.yaml and validates to the strict type validation.

## Related Issue Link(s)

#4511

## Release/Deployment Notes

This would not have any impact of the DDEV software itself, other than shipping two JSON schema files for use by other software.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5281"></a>

### DIFF
--- a/pkg/ddevapp/schema.json
+++ b/pkg/ddevapp/schema.json
@@ -1,0 +1,401 @@
+{
+  "$id": "https://json.schemastore.org/base.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "DDEV project config",
+  "description": "Schema for DDEV project config for version 1.22",
+  "type": "object",
+  "additionalProperties": false,
+  "definitions": {
+    "DdevTask": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "exec": {
+            "type": "string"
+          },
+          "exec-host": {
+            "type": "string"
+          },
+          "composer": {
+            "type": "string"
+          }
+        }
+      }
+    }
+  },
+  "properties": {
+    "additional_fqdns": {
+      "description": "A comma-delimited list of FQDNs for the project",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "uniqueItems": true
+    },
+    "additional_hostnames": {
+      "description": "A comma-delimited list of hostnames for the project",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "uniqueItems": true
+    },
+    "bind_all_interfaces": {
+      "description": "Bind host ports on all interfaces, not just on localhost network interface",
+      "type": "boolean"
+    },
+    "composer_root": {
+      "description": "The relative path, from the project root, to the directory containing composer.json. (This is where all Composer-related commands are executed.)",
+      "type": "string"
+    },
+    "composer_version": {
+      "description": "Composer version for the web container and the ddev composer command.",
+      "type": "string"
+    },
+    "database": {
+      "description": "Specify the database type and version to use. Defaults to mariadb:10.3",
+      "type": "object",
+      "properties": {
+        "type": {
+          "description": "Specify the database type to use",
+          "type": "string"
+        },
+        "version": {
+          "description": "Specify the database version to use",
+          "type": "string"
+        }
+      }
+    },
+    "dbimage_extra_packages": {
+      "description": "A list of Debian packages that should be added to db container when the project is started",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "uniqueItems": true
+    },
+    "default_container_timeout": {
+      "description": "Seconds DDEV will wait for all containers to become ready.",
+      "type": "integer"
+    },
+    "disable_settings_management": {
+      "description": "Prevent ddev from creating or updating CMS settings files",
+      "type": "boolean"
+    },
+    "disable_upload_dirs_warning": {
+      "description": "Whether to disable the standard warning issued when a project is using performance_mode: mutagen but upload_dirs is not configured.",
+      "type": "boolean"
+    },
+    "docroot": {
+      "description": "Provide the relative docroot of the project, like 'docroot' or 'htdocs' or 'web', defaults to empty, the current directory",
+      "type": "string"
+    },
+    "fail_on_hook_fail": {
+      "description": "Decide whether 'ddev start' should be interrupted by a failing hook",
+      "type": "boolean"
+    },
+    "hooks": {
+      "description": "Run tasks before or after main DDEV commands are executed",
+      "type": "object",
+      "properties": {
+        "pre-start": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-start": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-import-db": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-import-db": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-import-files": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-import-files": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-composer": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-composer": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-stop": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-config": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-config": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-exec": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-exec": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-pause": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-pause": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-pull": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-pull": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-push": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-push": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-snapshot": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-snapshot": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "pre-restore-snapshot": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-restore-snapshot": {
+          "$ref": "#/definitions/DdevTask"
+        },
+        "post-stop": {
+          "$ref": "#/definitions/DdevTask"
+        }
+      }
+    },
+    "host_db_port": {
+      "description": "The db container's localhost-bound port",
+      "type": "string"
+    },
+    "host_dba_port": {
+      "description": "The dba (PHPMyAdmin) container's localhost-bound port, if exposed via bind-all-interfaces",
+      "type": "string"
+    },
+    "host_https_port": {
+      "description": "The web container's localhost-bound https port",
+      "type": "string"
+    },
+    "host_webserver_port": {
+      "description": "The web container's localhost-bound port",
+      "type": "string"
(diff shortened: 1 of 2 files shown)


jamesmacwhite:
Sorry for the force pushes, the regex rule for PR title doesn't allow the first word to be uppercase, which happened to be "JSON".

rfay:
Your fork is your own personal playground. You can force-push there all day.

rfay:
But.. to change the name of the PR you just edit it in the GitHub web UI, has nothing to do with pushing code.

pkg/ddevapp/schema.json
```diff
@@ -0,0 +1,401 @@
+{
+  "$id": "https://json.schemastore.org/base.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "DDEV 1.22 project config",
```
rfay, line 4: Wouldn't this be better as just "DDEV project config" ? Easier to maintain... 
```suggestion
  "title": "DDEV project config",
```

pkg/globalconfig/schema.json
```diff
@@ -0,0 +1,250 @@
+{
+  "$id": "https://json.schemastore.org/base.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "DDEV 1.22 global config",
```
rfay, line 4:
```suggestion
  "title": "DDEV global config",
```

rfay (commented):
Such great, meticulous work, thanks so much!  This is going to be so valuable. 

Is there a way to test it currently in an IDE like GoLand or PHPStorm or vscode?

jamesmacwhite:
Thanks, I believe this is one way of testing it: https://scottaddie.com/2016/08/02/community-driven-json-schemas-in-visual-studio-2015/

You should be able to manually map one as well, until we can get it published in the JSON Schema library: https://www.jetbrains.com/help/phpstorm/json.html#ws_json_using_schemas

rfay:
Do we have to add a hint into the .ddev/config.yaml etc to tell IDEs where to look for this stuff?

Can we add validation at `ddev start` time to warn people about invalid schemas? Mostly the yaml will fail, but, for example, they might be using obsolete or deprecated or unsupported values etc.

jamesmacwhite:
> Do we have to add a hint into the .ddev/config.yaml etc to tell IDEs where to look for this stuff?
> 
> Can we add validation at `ddev start` time to warn people about invalid schemas? Mostly the yaml will fail, but, for example, they might be using obsolete or deprecated or unsupported values etc.

From what I can tell, the JSON schema being automatically mapped from the JSON Schema Store when published, uses a fileMatch rule, which would target the .ddev/config.yaml path, so it should be done automatically.

https://github.com/jamesmacwhite/schemastore/commit/1a2202e71b0bfa39fbba6e54908c1ca3851b81f2#diff-1b8f038f5afb1158263a1fc83b9c0ca5a7438cceb90adad99d87168f70edc815R1310-R1315

I'm not sure how the global_config.yaml would work, given the path is a user home directory which could be anything and variable, I don't think fileMatch would support $HOME as a variable.

rfay:
Thanks! I was able to do this with

image

It looks like it's pretty angry at me right now.

jamesmacwhite:
I'll locally test as well. First observation is that the schema is under `http://json-schema.org/draft-07/schema#` which might not match the schema version on that screenshot, but given the unit tests passed, it's interesting if it errors.

rfay:
Please re-add the "Manual testing instructions" and explain how to test this against existing config.yaml, global_config.yaml. I assume there's some kind of validator CLI that can be used? Otherwise, explaining how it's done with IDE will be great. 

Also, it's complaining about all the comments in the config.yaml. Is there a way we can stop that? Should we be using a jsonc or somethjing?

jamesmacwhite:
Have you set JSON Schema Version 7? This is what the spec was created as, when doing that on my test, other than a single line comment that causes an "Unpaired symbol" error, because of the two literal quotes. It seems OK.

> # is not recommended, as the images are often closely tied to ddev's' behavior,

Hovering over YAML key names, is pulling the description data, so it's linked with it.

rfay:
This is so great! 

Changing to "JSON Schema Version 7" did not make it ignore comments. 

I just realized we should be validating `config*.yaml` - to bring in config added by add-ons and such.

rfay:
I note that GoLand is providing just a little help for things like php_version:

image



```json
    "php_version": {
      "description": "The PHP version the project should use.",
      "type": "string",
      "enum": [
        "5.6",
        "7.0",
        "7.1",
        "7.2",
        "7.3",
        "7.4",
        "8.0",
        "8.1",
        "8.2",
        "8.3"
      ]
    },
```

It makes me wonder if there's a way to get it to give a more intelligent response, with things like:
* Possible values
* Link to docs

I'm not trying to expand scope on you, just loving what you're doing and think it's going to help people enormously. 

I wonder if we can just lose all the comments at the bottom of the file!

jamesmacwhite:
> I note that GoLand is providing just a little help for things like php_version:
> 
> image
> ```json
>     "php_version": {
>       "description": "The PHP version the project should use.",
>       "type": "string",
>       "enum": [
>         "5.6",
>         "7.0",
>         "7.1",
>         "7.2",
>         "7.3",
>         "7.4",
>         "8.0",
>         "8.1",
>         "8.2",
>         "8.3"
>       ]
>     },
> ```
> 
> It makes me wonder if there's a way to get it to give a more intelligent response, with things like:
> 
> * Possible values
> * Link to docs
> 
> I'm not trying to expand scope on you, just loving what you're doing and think it's going to help people enormously.
> 
> I wonder if we can just lose all the comments at the bottom of the file!

Comments wise, I'm not sure what we can do, from testing locally the YAML comments for me aren't causing issue beside the line where there is two literal quotes which has been interpreted as an "Unpaired symbol", but this is just the spell check system, otherwise I'm not having issues with the comments themselves breaking the schema itself or anything.

From looking at the JSON spec there is description property, which can be used on every property and in the case of objects and arrays any sub keys. There is a comment property, but it's referenced as not being for communicating schema information, so not usable for the purpose: http://json-schema.org/understanding-json-schema/reference/generic.html#id3.

I was hoping that for properties like project type, the schema also uses enums which aligns to how Goland is doing it, but it doesn't seem to flag invalid values. I'll have to have a look at that. I do remember that for @nico-loeber PHPStorm DDEV plugin, this did flag values that weren't part of enums, but not sure how that was done!

rfay:
It's doing great flagging invalid PHP versions:

image

rfay:
webserver_type validation also fantastic.

image

jamesmacwhite:
Ah that's good then, for some reason PHPStorm isn't doing that for me. Not sure why, but messing aorund with local schema definitions shouldn't be required if it gets added into the JSON schema store for automatic matching.

rfay:
I moved this to draft status so it doesn't have to be reviewed yet. I'm sure excited to see it land!

tyler36:
This is great!

## Manually VScode test for project-level `config.yam`

1. Save [local schema](https://github.com/ddev/ddev/pull/5281/files#diff-0a05e6727507923569b02c902b25b193daa3e15d132558f18e2dc04fe43bae17) as `./.ddev/schema.json` in an existing project.
1. Install [YAML extention ](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)
1. Update VScode settings to map `./.ddev/schema.json` to `./.ddev/config.yaml`
    ```json
    "yaml.schemas": {
        ".ddev/schema.json": ".ddev/config.yaml"
    }
    ```
1. Hover over keys on `.ddev/config.yaml` to see description
1. Change value to something invalid to test validation: eg `php_version: "5.0"`

Note: The YAML extension is my preferred extension for YAML files (has 13 million +installs).

jamesmacwhite:
Good to know! I think we can be more expansive on the labelling and description of properties in the config,yaml for better documentation and glad enums work to make sure only valid options for values like project type or php version work. The initial work has been to create a schema that validates against a spec, again acknowledging that a lot of this was done by @nico-loeber originally, but just needed some amends to validate against the spec as well as syncing to DDEV 1.22 properties.

The main intention once merged will be to get it listed in the JSON schema which as far as I understand many IDEs like VSCode, PHPStorm etc all can use JSON Schema store by default to automatically pull schemas automatically, so getting it merged into https://github.com/schemastore/schemastore should allow for auto matching and avoid manual configuration of that part.

I don't know if a schema for the global config is actually something we could add into schema store, because the path is under a HOME dir path which I don't think the fileMatch logic will handle, the config.yaml should be fine with a rule like `**/.ddev/config.yaml` given it always exist under a project folder. The global_config.yaml doesn't.

rfay:
I think this is great, so can merge it when you say, and then we can point to it from the upstream location... and then this can be improved as time goes on.

It would be great if we could have it not flag comments, but I think that's a problem upstream and we may not be able to solve it.

When you're ready for it to be merged, please move it to "Ready for Review" and say we should pull it, then let us know about the upstream PR.

rfay:
OK, thanks! I tried to figure out how to easily exclude it from triggering all the other tests but didn't figure it out right now. We can use `[skip ci]` for minor updates to it. 

Please follow up with the upstream PR, and thanks so much for making this happen. Also, the https://github.com/php-perfect/ddev-intellij-plugin PR or issue.

PR #5295, opened 2023-08-24
docs: add Netbeans IDE Debugger configuration, instructions, and screenshot.

larsdesigns:
## The Issue

Netbeans is a great open source IDE choice for working on Drupal and Symfony projects. It works well with XDEBUG using the Netbeans Debugger but there is not mention of configuration for it in the documentation. Only proprietary IDEs such as PHPStorm and Microsoft's Visual Studio Code documentation is available for DDEV. 

Provides instructions and configuration to connect DDEV's XDEBUG to the Netbeans Debugger for step through debugging.

Give it a try and/or proof read the instructions.

## Automated Testing Overview

This is a documentation update. Automated code testing is no required.

## Related Issue Link(s)

None

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5295"></a>


rfay:
Thanks, we had netbeans docs in there for several years... but nobody ever used it. Maybe it belongs somewhere besides the docs?

larsdesigns:
Oh, I was not aware of the history. I was just thinking it would be helpful or at least helpful for me :-) I am open to suggestions.

rfay:
Have you been using Netbeans through the ages? Or newly discovered?

larsdesigns:
I have using it for a while since it became an Apache project. It has come a long way in the last few years and it feels like there has been a growing user base.

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -87,6 +88,25 @@ However, if you’ve not yet used PhpStorm with Xdebug for a regular web request
...
+Complete the following DDEV configuration steps to connect the Netbeans debugger to XDEBUG:
+
+1. Create a project in Netbeans and set the project URL to use DDEV's project URL. For example use `https://drupal.ddev.site` and specify the index file as index.php. ![debugger-settings-screenshot](../../images/netbeans18-project-configuration.png)
+2. Create the xdebug.ini file in `.ddev/php/` directory. The full path will be: `.ddev/php/xdebug.ini`
+3. Add the PHP configuration to the xdebug.ini file:
+```ini
+[PHP]
+xdebug.mode=debug
+xdebug.idekey="netbeans-xdebug"
+```
```
rfay, line 105: This step should not be necessary at all I wouldn't think. I would expect it to be required only for command-line debugging. Are you sure it's necessary?

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -87,6 +88,25 @@ However, if you’ve not yet used PhpStorm with Xdebug for a regular web request
...
+4. Restart ddev: `ddev restart`
+5. Verify that Netbeans is configured as shown in the screenshot ![project-properties-screenshot](../../images/netbeans18-configuration.png)
+6. For debugging Use the command `telnet localhost 9003` to verify that the xdebug is available. To exit telnet press `control+]`and then enter the `quit` command.
```
rfay, line 108: It's not clear why this is necessary. And `telnet` isn't on everybody's host system.

rfay (changes requested):
I'm downloading netbeans to check... on slow internet. But could you please compare this to the previous version, https://github.com/ddev/ddev/blob/ed03c1bf592aefbf54607dbf5cb98e6999eb8cb1/docs/users/step-debugging.md#netbeans-debugging-setup - I think this is too complicated.

rfay:
Would you mind making this a Stack Overflow question-and-answer instead of adding to the docs here? You just create a question like "How can I use NetBeans and Xdebug with DDEV", and write the question and then before submitting click the "Answer the question yourself". And of course you tag it as "ddev". 

This will give you the ability to easily maintain/edit the answer, and directly support people who ask questions.

I did get it working, and did seem to need the INI file (unnecessary with anything else). Performance of Netbeans with its java setup was pretty sad (macbook air M1) compared to PhpStorm or vscode. You'd probably like vscode if you don't want to pay for PhpStorm, but you like what you have and that's fine!

rfay:
No response yet, but happy to continue the conversation here or elsewhere. This is a valuable contribution for a very small audience, so let's put it where it will get great SEO and be easily findable with "ddev netbeans" on Stack Overflow. Please let me know if you need any help, and I'm happy to continue the conversation.

docs/content/users/debugging-profiling/step-debugging.md
larsdesigns: Yes, we I think we can remove the debugging step.

larsdesigns:
The last I checked, Netbeans has a user base of over 1 million but I am not sure how that compares to the others. Using JDK 11 or higher will give a lot better performance. I am not sure which version of JDK you tested with. Also, for the ARM processors, I recommend compiling from source using Ant to produce native binaries. Things tend to slow down a lot when using Rosetta to run x86 apps.

Anyway, I can understand the direction you are taking. I will post the instructions elsewhere as you suggested. Thank you so much for reviewing the issue.

rfay:
Thanks so much for this contribution. Yes, I was testing on arm64, and the fact that they don't publish arm64 is a sign of things... 

When you post, please add instructions about installing java, which caught me by surprise. 

I used 

```
$ java --version
java 20.0.2 2023-07-18
Java(TM) SE Runtime Environment (build 20.0.2+9-78)
Java HotSpot(TM) 64-Bit Server VM (build 20.0.2+9-78, mixed mode, sharing)
``` 

But IMO people shouldn't have to know that kind of thing.

PR #5307, opened 2023-08-30
docs: mention xdebug_ide_location setting in xdebug troubleshooting section

ultimike:
Xdebug troubleshooting doc update for xdebug_ide_location.

-mike

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5307"></a>

### DIFF
--- a/docs/content/users/debugging-profiling/step-debugging.md
+++ b/docs/content/users/debugging-profiling/step-debugging.md
@@ -118,6 +118,7 @@ Here are basic steps to take to sort out any difficulty:
 * Reboot your computer.
 * If you're running WSL2 and have PhpStorm running inside WSL2 (the Linux version of PhpStorm) then `ddev config global --xdebug-ide-location=wsl2`.
 * Temporarily disable any *firewall* or *VPN* if you’re having trouble. Xdebug is a network protocol, and the PHP process inside the web container must be able to establish a TCP connection to the listening IDE (PhpStorm, for example).
+* Check to confirm that the DDEV configuration parameter `xdebug-ide-location` is set properly. For most situations, this should be set to an empty string. Be sure to check both your project's .ddev/config.yaml and ~/.ddev/global_config.yaml. More information about this parameter can be found in ~/.ddev/global_config.yaml. 
 * Use `ddev xdebug on` to enable Xdebug when you want it, and `ddev xdebug off` when you’re done with it.
 * Set a breakpoint at the first executable line of your `index.php`.
 * Tell your IDE to start listening. (PhpStorm: click the telephone button, VS Code: run the debugger.)
--- a/docs/content/users/debugging-profiling/step-debugging.md
+++ b/docs/content/users/debugging-profiling/step-debugging.md
@@ -131,6 +131,7 @@ Here are basic steps to take to sort out any difficulty:
 * Check to make sure that Xdebug is enabled. You can use `php -i | grep -i xdebug` inside the container, or use any other technique you want that gives the output of `phpinfo()`, including Drupal’s `admin/reports/status/php`. You should see `with Xdebug v3` and `php -i | grep xdebug.mode` should give you `xdebug.mode => debug,develop => debug,develop"`.
 * Set a breakpoint in the first relevant line of your `index.php` and then visit the site in a browser. It should stop at that first line.
 * If you're using a flavor of IDE that connects directly into the web container like VS Code Language Server, you may want to use the [global `xdebug_ide_location` setting](../configuration/config.md#xdebugidelocation) to explain to DDEV the situation. For example, `ddev config global --xdebug-ide-location=container`, which tells the PHP/Xdebug to connect directly to the listener inside the container.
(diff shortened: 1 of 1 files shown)


docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -118,6 +118,7 @@ Here are basic steps to take to sort out any difficulty:
 * Reboot your computer.
 * If you're running WSL2 and have PhpStorm running inside WSL2 (the Linux version of PhpStorm) then `ddev config global --xdebug-ide-location=wsl2`.
 * Temporarily disable any *firewall* or *VPN* if you’re having trouble. Xdebug is a network protocol, and the PHP process inside the web container must be able to establish a TCP connection to the listening IDE (PhpStorm, for example).
+* Check to confirm that the DDEV configuration parameter `xdebug-ide-location` is set properly. For most situations, this should be set to an empty string. Be sure to check both your project's .ddev/config.yaml and ~/.ddev/global_config.yaml. More information about this parameter can be found in ~/.ddev/global_config.yaml.
```
mattstein, line 121: It looks like the config setting should use underscores and that there isn’t an equivalent [config command](https://ddev.readthedocs.io/en/stable/users/usage/commands/#config).

I’d edit this to use underscores, link to the relevant documentation instead of mentioning the comments in the global config file, wrap filenames in backticks for consistency, gently reiterate project YAML vs. global YAML, and whittle things down for brevity:

```suggestion
* Confirm that DDEV’s [`xdebug_ide_location`](../../users/configuration/config/#xdebug_ide_location) config setting is set properly, which in most cases should be set to an empty string. Check both your project's `.ddev/config.yaml` and DDEV’s global `~/.ddev/global_config.yaml`.
```

mattstein (changes requested):
Thank for this! Left some minor suggestions.

docs/content/users/debugging-profiling/step-debugging.md
ultimike: Yep, works for me.

mattstein: You can click **Commit suggestion** if you’re cool with it.

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -118,6 +118,7 @@ Here are basic steps to take to sort out any difficulty:
 * Reboot your computer.
 * If you're running WSL2 and have PhpStorm running inside WSL2 (the Linux version of PhpStorm) then `ddev config global --xdebug-ide-location=wsl2`.
 * Temporarily disable any *firewall* or *VPN* if you’re having trouble. Xdebug is a network protocol, and the PHP process inside the web container must be able to establish a TCP connection to the listening IDE (PhpStorm, for example).
+* Confirm that DDEV’s [`xdebug_ide_location`](../../users/configuration/config/#xdebug_ide_location) config setting is set properly, which in most cases should be set to an empty string. Check both your project's `.ddev/config.yaml` and DDEV’s global `~/.ddev/global_config.yaml`.
```
rfay, line 121: Pretty sure this should be: (anyway, failed link check)

```suggestion
* Confirm that DDEV’s [`xdebug_ide_location`](../../configuration/config/#xdebug_ide_location) config setting is set properly, which in most cases should be set to an empty string. Check both your project's `.ddev/config.yaml` and DDEV’s global `~/.ddev/global_config.yaml`.
```

mattstein: Wow, I checked it locally so it’s kind of amazing I still managed to get it wrong. Thanks for fixing!

rfay:
I fixed the link and added another stanza about `DDEV_DEBUG=true ddev start` to find out the value of `host.docker.internal`

docs/content/users/debugging-profiling/step-debugging.md
rfay: It took me three tries. Then go back to the IDE locally, where it helped me get the right syntax. It would help if the markdown syntax were the same as http syntax.

ultimike:
Quick followup on this - I just confirmed that when using PhpStorm with xdebug set to be run via "Test frameworks" inside the DDEV web container (using the "DDEV" CLI Interpreter automatically provided by [the](https://plugins.jetbrains.com/plugin/18813-ddev-integration), that the `xdebug-ide-container` must be set to an empty string (as described in the doc change for this issue).

The problem I faced which triggered this whole discussion is that I alternate between PhpStorm and VSCode (mainly while teaching) and I had set `xdebug-ide-container=container` in my ~/.ddev/global_config.yaml - which caused xdebugging of PHPUnit tests in DDEV to fail. 

One thing I did notice that is important for xdebugging PHPUnit tests using my setup in PhpStorm is that the "Host" and "Name" in the "PHP | Servers" configuration are the same (and of the form name.ddev.site). This is not the way that the DDEV PhpStorm plugin configures the "server" out-of-the-box. I have documented this in the [DDEV plugin issue queue](https://github.com/php-perfect/ddev-intellij-plugin/issues/227).

I believe that my global setting of`xdebug-ide-container=container` was a remnant of previous experiments of trying to get xdebug working from the VSCode UI via the PHPUnit Test Explorer extension without using the "Remote development" extension.

Issue #4701, opened 2023-03-04
Clean up forked repositories/releases


rfay:
There are several open source repos that have been forked or otherwise copied for release or development reasons. We should mostly get back to the original upstream.

- [x] github.com/ddev/MailHog https://github.com/ddev/ddev/pull/5313
- [x] github.com/ddev/nssm
- [x] github.com/ddev/gsudo https://github.com/ddev/ddev/pull/5314
- [x] github.com/ddev/docker-gen
- [x] github.com/ddev/forego
- [ ] github.com/ddev/wordpress (is it necessary? Can it have a better name)

Edit: Not doing nssm because it hasn't changed and downloading from the fork is absolutely better than nssm.cc

rfay:
* https://github.com/ddev/ddev/pull/5489 moves shopware away from the junk downloads and into its own composer-based repository.

rfay:
We'll just make progress on this when we work on tests. Each test tarball for tests should be moved to its own maintained github repo.

rfay:
* https://github.com/ddev/ddev/pull/5804 moves magento2 stuff into the maintainable https://github.com/ddev/test-magento2 repo

PR #5314, opened 2023-08-31
build: use gsudo.exe directly from gerardog/gsudo, for #4701

rfay:
## The Issue

* #4701

We want to use gsudo.exe directly from the upstream gerardog/gsudo repository, instead of using our forked one with an implanted old release

Use the upstream zipball and extract it.

* Remove any sudo.exe from the installer directory C:\Program Files\DDEV
Download the installer from this PR (https://github.com/ddev/ddev/pull/5314#issuecomment-1701925943) and run the installer. 
* Verify that sudo.exe has been installed and can be used

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5314"></a>


rfay:
Discussion with @gerardog regarding this in:
* https://github.com/gerardog/gsudo/pull/299
* https://github.com/gerardog/gsudo/issues/296

PR #5315, opened 2023-09-01
docs: note that the CLI flag `--project-name` is what sets `name` in config.yaml

mlncn:
## The Issue

name in the config yaml needs to be project-name when used with `ddev config`

Explains that.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5315"></a>

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -291,9 +291,7 @@ The URL-friendly name DDEV should use to reference the project.
 
 | Type | Default | Usage
 | -- | -- | --
-| :octicons-file-directory-16: project | enclosing directory name | Must be unique; no two projects can have the same name. It’s best if this matches the directory name. If this option is omitted, the project will take the name of the enclosing directory.
-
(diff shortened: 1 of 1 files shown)


docs/content/users/configuration/config.md
```diff
@@ -293,6 +293,8 @@ The URL-friendly name DDEV should use to reference the project.
 | -- | -- | --
 | :octicons-file-directory-16: project | enclosing directory name | Must be unique; no two projects can have the same name. It’s best if this matches the directory name. If this option is omitted, the project will take the name of the enclosing directory.
 
+!!!note "May also be set via `ddev config --project-name=example-project` (the flag is `project-name` not 'name')."
```
mattstein, line 296: Minor punctuation fix-up:

```suggestion
!!!note "May also be set via `ddev config --project-name=example-project`. (The flag is `project-name`, not `name`.)"
```

mattstein (changes requested):
Thanks for this! Left a minor suggestion.

rfay (approved):
Thanks so much for this PR. I took the liberty of putting the text into the paragraph instead of in a separate "note", and I added a link to the commands.md doc.

PR #5340, opened 2023-09-16
feat: output that the container images are being built, fixes #5339

### LINKED ISSUE #5339: Output something when an image is being rebuilt

Every now and then I get curious into ddev getting quite a bit to start, prompt me to troubleshoot and then I remember it's probably building the image again (maybe something changed, or the cache bust, or I upgraded docker/ddev).

### Describe your solution

I know you can get a lot more output with debug, but I wonder if a simple line saying this is happening could be useful.

Something like "your xxxx container is being built, it can take a few minutes" or something of the like.

### Describe alternatives

_No response_

### Additional context

_No response_

hanoii:
Fixes 
* #5339.

To test simply restart any project with it.

Screenshot 2023-09-16 at 16 26 49

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5340"></a>

### DIFF
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -42,7 +42,7 @@ var DebugRefreshCmd = &cobra.Command{
 			util.Failed("Failed to get compose-config: %v", err)
 		}
 
-		output.UserOut.Printf("Rebuilding project images... This can take some time.")
+		output.UserOut.Printf("Rebuilding project images...")
 		buildDurationStart := util.ElapsedDuration(time.Now())
 		util.Debug("Executing docker-compose -f %s build --no-cache", app.DockerComposeFullRenderedYAMLPath())
 		_, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -41,7 +41,10 @@ var DebugRefreshCmd = &cobra.Command{
 		}
 
 		util.Debug("Executing docker-compose -f %s build --no-cache", app.DockerComposeFullRenderedYAMLPath())
-		out, stderr, err := dockerutil.ComposeCmd([]string{app.DockerComposeFullRenderedYAMLPath()}, "build", "--no-cache")
+		out, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+			ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+			Action:       []string{"build", "--no-cache"},
+		})
 		output.UserOut.Printf("docker-compose build output:\n%s\n\n", out)
 		if err != nil {
 			util.Failed("Failed to execute docker-compose -f %s build --no-cache: %v; stderr=\n%s\n\n", err, stderr)
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -45,12 +45,11 @@ var DebugRefreshCmd = &cobra.Command{
 		output.UserOut.Printf("Rebuilding project images... This can take some time.")
 		buildDurationStart := util.ElapsedDuration(time.Now())
 		util.Debug("Executing docker-compose -f %s build --no-cache", app.DockerComposeFullRenderedYAMLPath())
-		out, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+		_, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
 			ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
 			Action:       []string{"build", "--no-cache"},
-			Progress:     true,
+			RealTime:     true,
 		})
-		output.UserOut.Printf("docker-compose build output:\n%s\n\n", out)
 		if err != nil {
 			util.Failed("Failed to execute docker-compose -f %s build --no-cache: %v; stderr=\n%s\n\n", err, stderr)
 		}
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"time"
+
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/output"
@@ -40,21 +42,25 @@ var DebugRefreshCmd = &cobra.Command{
 			util.Failed("Failed to get compose-config: %v", err)
 		}
 
+		output.UserOut.Printf("Rebuilding project images... This can take some time.")
+		buildDurationStart := util.ElapsedDuration(time.Now())
 		util.Debug("Executing docker-compose -f %s build --no-cache", app.DockerComposeFullRenderedYAMLPath())
 		out, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
 			ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
 			Action:       []string{"build", "--no-cache"},
+			Progress:     true,
 		})
 		output.UserOut.Printf("docker-compose build output:\n%s\n\n", out)
 		if err != nil {
 			util.Failed("Failed to execute docker-compose -f %s build --no-cache: %v; stderr=\n%s\n\n", err, stderr)
 		}
+		buildDuration := util.FormatDuration(buildDurationStart())
+		util.Success("Refreshed Docker cache for project %s in %s", app.Name, buildDuration)
+
 		err = app.Restart()
 		if err != nil {
 			util.Failed("Failed to restart project: %v", err)
 		}
-
-		util.Success("Refreshed Docker cache for project %s", app.Name)
 	},
 }
 
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1261,7 +1261,11 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	buildDurationStart := util.ElapsedDuration(time.Now())
 	progress := "plain"
 	util.Debug("Executing docker-compose -f %s build --progress=%s", app.DockerComposeFullRenderedYAMLPath(), progress)
-	out, stderr, err := dockerutil.ComposeCmd([]string{app.DockerComposeFullRenderedYAMLPath()}, "--progress="+progress, "build")
+	out, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+		ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+		Action:       []string{"--progress=" + progress, "build"},
+		Progress:     true,
+	})
 	if err != nil {
 		return fmt.Errorf("docker-compose build failed: %v, output='%s', stderr='%s'", err, out, stderr)
 	}
@@ -1272,7 +1276,10 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	util.Success("Project-specific images built in %s.", buildDuration)
 
 	util.Debug("Executing docker-compose -f %s up -d", app.DockerComposeFullRenderedYAMLPath())
-	_, _, err = dockerutil.ComposeCmd([]string{app.DockerComposeFullRenderedYAMLPath()}, "up", "-d")
+	_, _, err = dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+		ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+		Action:       []string{"up", "-d"},
+	})
 	if err != nil {
 		return err
 	}
@@ -1783,7 +1790,10 @@ func (app *DdevApp) Exec(opts *ExecOpts) (string, string, error) {
 	if opts.NoCapture || opts.Tty {
 		err = dockerutil.ComposeWithStreams(files, os.Stdin, stdout, stderr, r...)
 	} else {
-		outRes, errRes, err = dockerutil.ComposeCmd([]string{app.DockerComposeFullRenderedYAMLPath()}, r...)
+		outRes, errRes, err = dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+			ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+			Action:       r,
+		})
 		stdoutResult = outRes
 		stderrResult = errRes
 	}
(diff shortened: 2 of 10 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -1257,6 +1257,8 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	}
 
 	// Build extra layers on web and db images if necessary
+	output.UserOut.Printf("Building project containers... This can take some time.")
```
rfay, line 1260:
```suggestion
	output.UserOut.Printf("Building project-specific images... This can take some time.")
```

pkg/ddevapp/ddevapp.go
```diff
@@ -1269,6 +1271,8 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	if globalconfig.DdevVerbose {
 		util.Debug("docker-compose build output:\n%s\n\n", out)
 	}
+	buildDuration := util.FormatDuration(buildDurationStart())
+	util.Success("Containers built in %s.", buildDuration)
```
rfay, line 1275:
```suggestion
	util.Success("Project-specific images built in %s.", buildDuration)
```

pkg/ddevapp/ddevapp.go
```diff
@@ -1257,6 +1257,8 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	}
 
 	// Build extra layers on web and db images if necessary
+	output.UserOut.Printf("Building project containers... This can take some time.")
+	buildDurationStart := util.ElapsedDuration(time.Now())
 	progress := "quiet"
 	if globalconfig.DdevVerbose {
```
rfay, line 1263: The next thing that would be hugely useful here is a good way to do the verbose build output.

rfay (changes requested):
Thanks for doing this. I haven't tried it yet, but am sure it's great. Minor wording suggestions. But the other big missing thing is how to get the better output from `docker-compose build`

pkg/ddevapp/ddevapp.go
hanoii: Not sure I understand this part, and would this be something you wanna do as part of this PR?

hanoii:
Apply your suggestions and followed up on one of the comments.

>  But the other big missing thing is how to get the better output from docker-compose build

Is this the same thing? Not quite sure what you mean.

rfay:
Not the same thing. Same basic problem area. People don't want to just see the time it's taking, but also what it's doing.

I was just commenting, not asking for change.

Of course this would be a little nicer with dots like the mutagen sync, but not asking for that either.

hanoii:
I'll investigate a bit further

hanoii:
@rfay so I spent quite a bit today on this :)... had to grasp some more go concepts. goroutines are pretty nice! I'd love your review and any insight into the approach. New to channels et all.

Obviously this adds more than, as it's showing dots not just for the compose build for every compose commands. This can be parametrized for sure but I figured maybe it's not that bad.

I also consider using some kind of ascii spinner, kind of how the gatsby cli works, and a quick search gave me some go pkgs, but maybe too much:

gatsby

rfay:
Looking forward to trying it out. Also looking forward to trying all the stuff from the charmbracelet folks, https://github.com/charmbracelet/bubbletea - they have done lovely things.

Thanks, wasn't trying to push you farther just was thinking out loud.

hanoii:
Most tests are failing but unsure if it is because of this, will review though tomorrow.

rfay:
Looks like real test failures, just based on some strict/fragile output checks

rfay:
Rebased.

rfay (changes requested):
I'm liking this, and think it has great value. 

The problem right now is that there are many calls to ComposeCmd() and most of them are short, like exec. So they make dots all over the place without explaining what they're for.

I like the dots for `stop`. But it needs to say why they're being output. 

So I think we want to only show them for specific actions. 

I think this could be used to wrap any function perhaps, and maybe replace the mutagen dots. 

I did try fooling with a commit that tried to add the action being undertaken to the ComposerCmd() function, but wasn't all that successful because it's not clear where the key action (up, stop, build) is in the args array.

For example:
```
$ ddev delete -Oy
.......... Container ddev-d10-web  Stopped
 Container ddev-d10-web  Removed
 Container ddev-d10-adminer  Stopped
 Container ddev-d10-adminer  Removed
. Container ddev-d10-db  Stopped
 Container ddev-d10-db  Removed
 Network ddev-d10_default  Removed
 ```
 
 There is a set of unexplained dots to begin, and one spare dot that is who knows what.

But nice work, and so ambitious! If we can make it so we only get dots for things we want, and perhaps if we can refactor so this is more generally usable, this will go. 

The options are to 
1. Go back to your original
2. Just get dots when we want them (probably build, stop, up)
3. Actually refactor to have reusable dots in a nice package. 

We can do whichever of those you want.

**Comment by hanoii** [comment]
I'd aim for 2. I think 3 is too much for this. I already had an idea to limit this and is not by parsing the argument and instead add an option for the compose function and let the caller control when to add dots. Should be easy enough. I'll look into this

**Comment by hanoii** [comment]
Would you consider changing`.......... Container ddev-d10-web  Stopped` for `Stopping container ddev-d10-web ....`

Logging before doing the action and adding the dots later might be more descriptive. This will likely make a lot of tests fail but they can be adapted

**Comment by rfay** [comment]
> Would you consider changing.......... Container ddev-d10-web  Stopped for Stopping container ddev-d10-web ....

Sure, works for me.

**Comment by hanoii** [comment]
@rfay `Container ddev-d10-web` et all comes right out of docker-compose, so adding dots after means parsing the output and I consider that too much right now.

I decided to modify the function signature to accept a struct rather than a variadic argument, allowing for the Progres flag and eventually more control over the call of that function.

I hope you like it.

Adding dots to any other action will not fit in nicely in between the output, so for now just adding it to the build action as it's mostly where it makes the most sense, and now it kind of behaves and looks the same as mutagen.

Hope tests will pass 🤞

**Comment by hanoii** [comment]
Also I just changed `project-specific` for `project`. I didn't like project-specific that much, but it was your suggestion so if you rather have that I'll bring it in. I left images instead of containers.

**Comment by hanoii** [comment]
@rfay tweaked debug refresh to include this.

I have a question.. I tried to add a real time output of stdout instead of the dots. I couldn't make it work with `output.UserOut.Print()`, as no matter how I write to that it always prints a new line. And because I am treating the stdout as bytes, each byte/character is added to a new line. I couldn't find a way around it. I saw that is coming from logrus and tried to see if there was to useit somehow but couldn't.

Just doing: 

output.UserOut.Print("a")
output.UserOut.Print("b")

Prints it on a line, I would expect it to print it side by side like

fmt.Print("a")
fmt.Print("b")

https://go.dev/play/p/JQh9UEavQcL

I wasn't sure if using fmt.print as I realize logrus was there for a reason.

So for now just keeping it as before only adding dots and a better output. Just curious on this.

**Comment by hanoii** [comment]
I never remember or know if I can request some tests to restart, I think you mentioned I had a way to do it.

**Comment by rfay** [comment]
> I have a question.. I tried to add a real time output of stdout instead of the dots. I couldn't make it work with output.UserOut.Print(), as no matter how I write to that it always prints a new line. And because I am treating the stdout as bytes, each byte/character is added to a new line. I couldn't find a way around it. I saw that is coming from logrus and tried to see if there was to useit somehow but couldn't.

If you use stdout, it will break the json output of `ddev start -j`

> I never remember or know if I can request some tests to restart, I think you mentioned I had a way to do it.

I thought we got you set up on buildkite, which would give you that priv for the buildkite tests. Let me know what else you might need. (You do have to be logged into buildkite)

**Comment by rfay** [comment]
I restarted the failed github tests, but it seems TestDdevXdebugEnabled failed on both of them, with a conflict on port 9003. I haven't seen this problem before, but have no idea how it could be related to this PR.

**Comment by hanoii** [comment]
> If you use stdout, it will break the json output of `ddev start -j`

This was meant to be used with `ddev debug refresh`, but ok, no need right now. The question was if you knew a way to user output.UserOut.Something() that can allow me to print characters on next to the others without a newline at the end.

**Comment by rfay** [comment]
> The question was if you knew a way to user output.UserOut.Something() that can allow me to print characters on next to the others without a newline at the end.

output.UserOut.Printf()

**Comment by hanoii** [comment]
> output.UserOut.Printf()

Nope:

```
	output.UserOut.Printf("a")
	output.UserOut.Printf("b")
	output.UserOut.Printf("Building project images... This can take some time.")
```

```
 Network ddev-wild_default  Removed 
a 
b 
Building project images... This can take some time. 
Project images built in 0s. 
```

**Comment by rfay** [comment]
Here's what we're doing in the other place, which I thought you were imitating:
https://github.com/ddev/ddev/blob/16870030d464e7bb64a569685b65871ce5669469/pkg/ddevapp/mutagen.go#L284-L288

It puts it to stderr to avoid the json problem.

**Comment by hanoii** [comment]
I wasn't talking about the dots.. never mind.. I'll explain tomorrow :)

**Comment by hanoii** [comment]
@rfay I was going to leave it at that but I couldn't help myself. I improved the ComposeCmd function to also accept a RealTime as an option in which case stdout is print as is (to `output.UserOut`). I think this is nicer/useful only for `ddev debug refresh`

See https://github.com/ddev/ddev/pull/5340/commits/252b4b0b59604c8cfbf489620415d8e5ad8f843b

**Comment by hanoii** [comment]
@rfay again some tests are failing with this port error, how odd.

I logged in to buildkite and not too familiar with it. I see three things running but no failure. Do I have to wait until everything finishes before I can restart the test that failed?

**Comment by rfay** [comment]
Again TestDdevXdebugEnabled, again `        	            	listen tcp :9003: bind: address already in use`

This is either somehow impacted by this PR or it has snuck into master, but I don't think I've seen it on master. 

Here it happened on the flagship, most reliable test setup though, nginx on linux in github.

**Comment by rfay** [comment]
The other failure was TestDdevXhprofEnabled (on macOS). I restarted both. But TestDdevXhprofEnabled is a normal flaky test.

**Comment by hanoii** [comment]
@rfay it's an odd test failure for sure, I can't think why it can be related to this fail, and specially as it's random and then works.

**Comment by hanoii** [comment]
Hmmm, I tried running the test locally but complained on being M1

**Comment by rfay** [comment]
I don't know what "Complained about being M1" would mean. If you're using Colima, you will see this:

https://github.com/ddev/ddev/blob/e95a49632139a080e1d0aa07070e5e30767e9e7c/pkg/ddevapp/ddevapp_test.go#L877-L882

So you can set `DDEV_TEST_COLIMA_ANYWAY` if you want to. 

But running it locally isn't likely to catch it, unless perhaps it's a concurrency problem and this is leaving the port occupied. Actually that might be the one thing related to your PR and this test: They both use goroutines extensively. I wonder if something could have gone wrong.

**Comment by hanoii** [comment]
@rfay you linked to TestDdevXdebugEnabled and that one ended up passing, the complain on m1 I mentioned was for TestDdevXhprofEnabled that's now constantly failing:

https://github.com/ddev/ddev/blob/e95a49632139a080e1d0aa07070e5e30767e9e7c/pkg/ddevapp/ddevapp_test.go#L1067

**Comment by hanoii** [comment]
Sorry, accidentally closed it and reopened.

I also ran TestDdevXdebugEnabled locally (I am using orbstack) and it passed.

I restarted it one more time 🤷

**Comment by rfay** [comment]
TestDdevXhprofEnabled has been flaky since its inception. I intend to go work on it like I did a while back on TestDdevXdebugEnabled. That one was in fact goroutine reaping difficulties. So I'm not that worried that TestDdevXhprofEnabled would fail. But TestDdevXdebugEnabled hasn't had these problems with port left connected (ever?) so either something has crept in from master or something about the goroutines here is messing things up.

**Comment by rfay** [comment]
One other interesting factoid: go 1.21 is in use now, and we have `Found in cache @ /opt/hostedtoolcache/go/1.21.1/x64`

**Comment by rfay** [comment]
There are minor goroutine changes in go 1.21, nothing rings a bell, https://tip.golang.org/doc/go1.21

**Comment by rfay** [comment]
One other possibility: Something wrong with the github runner images. They change them all the time. It's even possible they run something that occupies port 9003.

**Comment by hanoii** [comment]
But in that case all of the xedebug tests should fail. Here looks just one and random

**Comment by rfay** [comment]
Inspired by your work:
* https://github.com/ddev/ddev/pull/5370

**Comment by rfay** [comment]
All tests passed this round. Did you restart any?

**Comment by hanoii** [comment]
> Not sure what the go factoid might have to do, but you're the expert. I did push a commit with a shot in the dark to see if it makes any difference to the xdebug test.
> 
> 
> 
> [42bae26](https://github.com/ddev/ddev/pull/5340/commits/42bae26a99d3d63d481a6659516b8fb3ba600ff4)

No but as I mentioned here my last commit added a tiny fix to the xdebug test. Maybe it helped?

**Comment by rfay** [comment]
I looked at that carefully; I'm not sure it's in the right place, but it's a right thing to do.

**Comment by hanoii** [comment]
I think it's right there right after select. At least at matches the net.Listen inside the for

**Comment by hanoii** [comment]
@rfay you left me thinking about where to put it. I looked at the code again and performed a few tests.

I put it now in both branches of the select. I think that's best, my reasoning:

- Just having it below the select work for successful tests, but on the fail, it wasn't really closed.
- `defer` would also not have worked as this is an exclusive resource that should be closed on the loop.

**Comment by rfay** [comment]
Tests looking good. I had to restart tb-wsl12, but that's not unusual, https://buildkite.com/ddev/wsl2-docker-inside/builds/2423#018abed9-4a69-45b9-b2bb-64d119b88a18

**Comment by rfay** [comment]
Could you please pull the changes to TestDdevXdebugEnabled out of this PR and put them in another? We haven't seen those failures here or elsewhere and now we don't know why.

**Comment by hanoii** [comment]
Rebased without the commits of the test fix as per your request, they are now on #5372

**Comment by rfay** [comment]
Thanks so much!

**Comment by rfay** [comment]
This is ready for review again, true?

**Comment by hanoii** [comment]
It is, yes!

File: cmd/ddev/cmd/debug-refresh.go [hunk 4]
```diff
@@ -40,18 +42,24 @@ var DebugRefreshCmd = &cobra.Command{
 			util.Failed("Failed to get compose-config: %v", err)
 		}
 
+		output.UserOut.Printf("Rebuilding project images... This can take some time.")
```
rfay, line 45:
```suggestion
		output.UserOut.Printf("Rebuilding project images...")
```

pkg/ddevapp/ddevapp.go
```diff
@@ -1257,21 +1257,29 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	}
 
 	// Build extra layers on web and db images if necessary
-	progress := "quiet"
-	if globalconfig.DdevVerbose {
-		progress = "auto"
-	}
+	output.UserOut.Printf("Building project images... This can take some time.")
```
rfay, line 1260:
```suggestion
	output.UserOut.Printf("Building project images...")
```

pkg/dockerutil/dockerutils.go
```diff
@@ -33,6 +33,13 @@ import (
 // NetName provides the default network name for ddev.
 const NetName = "ddev_default"
 
+type ComposeCmdOpts struct {
+	ComposeFiles []string
+	Action       []string
+	Progress     bool // Add dots every second while the compose command is running
+	RealTime     bool // Print stdout as it happens
```
rfay, line 40: Wow, do I love this! It makes `ddev debug refresh` so very much better.

rfay (approved):
This is really quite excellent, thanks so much for hanging with it! Great new feature, love the result in `ddev debug refresh`. 

Minor suggestions. I don't think we really have to say "this can take some time" when we're giving them dots, which always tells people the scoop.

hanoii:
@rfay Thanks for nice words! Sure, committed your minor suggestion!

hanoii:
And I am glad you are happy with the results, so am I!

hanoii:
Bear in mind that the same language about that it may take some time is in the mutagen line. I just attempted to leave it consistent. Just saying in case you want to remove it from there as well

rfay:
> Bear in mind that the same language about that it may take some time is in the mutagen line. I just attempted to leave it consistent. Just saying in case you want to remove it from there as well

See 
* https://github.com/ddev/ddev/pull/5370

:)

rfay:
Congrats and thanks!

PR #5354, opened 2023-09-19
fix: add PHP 8.3 to list of versions in config template

peterkraume:
## The Issue

PHP 8.3 is missing in the list of available PHP versions in the template for the configuration file.

This PR adds PHP 8.3 to the list.

After `ddev config` you should be able to see PHP 8.3 at the end of the list of available PHP versions in `config.yaml`.


rfay (approved):
Thanks! Here I thought I caught all the places :)

PR #5366, opened 2023-09-22
Update import-db parameter --target-db to --database

gitressa:
## The Issue

The `import-db` parameter `--target-db` is used, but should be `--database` under [Database Management > Extra Databases](https://ddev.readthedocs.io/en/latest/users/usage/database-management/#extra-databases).

Update parameter `--target-db` to `--database`.


rfay (approved):
This is correct. I liked --target-db better though!

PR #5372, opened 2023-09-23
test: close xdebug listening port after opening to avoid "listen tcp :9003: bind: address already in use" errors

hanoii:
In working on #5340 `TestDdevXdebugEnabled` was failing were not passing with random (but frequent) failure of `TestDdevXdebugEnabled` with

```
    ddevapp_test.go:989: 
        	Error Trace:	/home/runner/work/ddev/ddev/pkg/ddevapp/ddevapp_test.go:989
        	            	listen tcp :9003: bind: address already in use
        	Test:       	TestDdevXdebugEnabled
```

i.e. https://github.com/ddev/ddev/actions/runs/6273418171/job/17036856628#step:17:3502

These commits were previously on that issue and since then this test hasn't failed again.

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5372"></a>

### DIFF
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -1053,11 +1053,12 @@ func TestDdevXdebugEnabled(t *testing.T) {
 
 		select {
 		case <-acceptListenDone:
+			listener.Close()
 			fmt.Printf("Read from acceptListenDone at %v\n", time.Now())
 		case <-time.After(time.Second * 11):
+			listener.Close()
 			t.Fatalf("Timed out waiting for accept/listen at %v, PHP version %v\n", time.Now(), v)
 		}
-		listener.Close()
 	}
 	runTime()
 }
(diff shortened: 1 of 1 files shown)


rfay:
Thanks so much. I wanted to make sure we captured your good idea but didn't get confused by too many moving parts. I kind of wonder if it was something like github runners, since we saw a spurt of those but don't seem to have seen more. We'll see what happens in the original PR now.

hanoii:
Yeah. I still think we should have this. It smells like the potential issue is there.

rfay (approved):
I'm ok with this as-is, but since `select()` is blocking, why not do the listener close after the select?

hanoii:
@rfay select is blocking, but t.FatalF prevents what comes after select from running (I did check this). I am not 100% familiar with the whole safe threading of testing so I am unsure what happens with the listener and if it is indeed close by the goroutine, but is it's not even deferred, I thought being on the safe side.

hanoii:
That was actually what was on my first commit https://github.com/ddev/ddev/pull/5372/commits/6acc3f89c603fd44c974ad04d9104c4ae7669f7b

rfay:
Agreed. Normally this would be cleaned also on t.Cleanup() (like t.Fatalf()) but in this case we don't have the listener in scope. I guess we could have a t.Cleanup() for each PHP version as well. But fine to go with it as is. thanks for all the care on this. Not sure what was actually going on.

PR #5379, opened 2023-09-26
feat: add ddev_version_constraint to config.yaml, fixes #4630

### LINKED ISSUE #4630: Allow setting a minimum ddev version in config.yaml

Sometimes ddev adds a feature or fixes a bug that is critical for a project team. It would be great if a minimum (or semver style?) version could be set in a project configuration.

### Describe your solution

When ddev is run, and it's older than the version in config.yaml, ddev should show an error and prompt the user to upgrade.

### Describe alternatives

_No response_

### Additional context

_No response_

hanoii:
Attempts to enforce a specific ddev version using semver constraints

Fixes 
* #4630

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -107,7 +107,7 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 
 You can configure a [DDEV version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running DDEV executable and prevent any command from running if it doesn't match.
 
-Note: this will only work with ddev versions above v1.22.4, older ddev versions will ignore this setting.
+This will only work with DDEV versions above v1.22.4; older versions will ignore this setting.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -105,13 +105,11 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 
 ## `ddev_version_constraint`
 
-You can configure a ddev version constraint that will be validated against the running ddev and prevent any command from running if it doesn't match.
+You can configure a [ddev version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running ddev and prevent any command from running if it doesn't match.
 
 | Type | Default | Usage
 | -- | -- | --
 | :octicons-file-directory-16: project | | `>= 1.22.4`
-
-You can see the different constraint formats at https://github.com/Masterminds/semver#checking-version-constraints.
 
 ## `default_container_timeout`
 
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -107,6 +107,8 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 
 You can configure a [ddev version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running ddev and prevent any command from running if it doesn't match.
 
+Note: this will only work with ddev versions above v1.22.4, older ddev versions will ignore this setting.
+
 | Type | Default | Usage
 | -- | -- | --
 | :octicons-file-directory-16: project | | `>= 1.22.4`
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -105,7 +105,7 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 
 ## `ddev_version_constraint`
 
-You can configure a [ddev version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running ddev and prevent any command from running if it doesn't match.
+You can configure a [DDEV version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running DDEV executable and prevent any command from running if it doesn't match.
 
 Note: this will only work with ddev versions above v1.22.4, older ddev versions will ignore this setting.
 
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -103,6 +103,16 @@ Extra Debian packages for the project’s database container.
 
 Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the database container is built.
 
+## `ddev_version_constraint`
+
+You can configure a ddev version constraint that will be validated against the running ddev and prevent any command from running if it doesn't match.
+
+| Type | Default | Usage
+| -- | -- | --
+| :octicons-file-directory-16: project | | `>= 1.22.4`
+
+You can see the different constraint formats at https://github.com/Masterminds/semver#checking-version-constraints.
+
 ## `default_container_timeout`
 
 Seconds DDEV will wait for all containers to become ready.
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -643,6 +643,39 @@ func TestConfigValidate(t *testing.T) {
 	app.DdevVersionConstraint = ""
 	versionconstants.DdevVersion = ddevVersion
 
+	// prereleases with a constraint that does not include them
+	versionconstants.DdevVersion = "v1.22.3-11-g8baef014e"
+	app.DdevVersionConstraint = ">= 1.23"
+	err = app.ValidateConfig()
+	assert.Error(err)
+	assert.Contains(err.Error(), "ddev version v1.22.3-11-g8baef014e is not supported by this project")
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
+	versionconstants.DdevVersion = "v1.22.3-11-g8baef014e"
+	app.DdevVersionConstraint = ">= 1.22.0"
+	err = app.ValidateConfig()
+	assert.Error(err)
+	assert.Contains(err.Error(), "ddev version v1.22.3-11-g8baef014e is not supported by this project")
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
+	// prereleases with a constraint that does include them
+	versionconstants.DdevVersion = "v1.22.3-11-g8baef014e"
+	app.DdevVersionConstraint = ">= 1.23.0-0"
+	err = app.ValidateConfig()
+	assert.Error(err)
+	assert.Contains(err.Error(), "ddev version v1.22.3-11-g8baef014e is not supported by this project")
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
+	versionconstants.DdevVersion = "v1.22.3-11-g8baef014e"
+	app.DdevVersionConstraint = ">= 1.22.0-0"
+	err = app.ValidateConfig()
+	assert.NoError(err)
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
 	app.Name = "Invalid!"
 	err = app.ValidateConfig()
 	assert.Error(err)
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -636,6 +636,13 @@ func TestConfigValidate(t *testing.T) {
 	app.DdevVersionConstraint = ""
 	versionconstants.DdevVersion = ddevVersion
 
+	versionconstants.DdevVersion = "2134asdf-dirty"
+	app.DdevVersionConstraint = ">= 1.23"
+	err = app.ValidateConfig()
+	assert.NoError(err)
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
 	app.Name = "Invalid!"
 	err = app.ValidateConfig()
 	assert.Error(err)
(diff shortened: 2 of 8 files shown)


hanoii:
@rfay would you require tests for this feature?

rfay:
Yeah, I'm sure it needs a test, thanks. It can probably be a unit test, or something more extensive, but doesn't need a `ddev start`

hanoii:
@rfay any pointer into unit tests.. is there a way to test against an older version of ddev?

rfay:
I'll look for an example.

hanoii:
I wonder if this really needs a test? It's pretty basic and I am relying on semver which I'd assume it has tests on its own, and it's something that only kicks in if the config key is present..  Not sure how useful a test would be around this particular bit.

jonaseberle:
Thank you!
Tests can show the values we expect, also as a reference later. 
It might also be an incentive the functionality into an own function (or maybe even a class?) ;)

Mind that we should support dev versions like `v1.22.3-11-g8baef014e`, too which as I understand are not totally SemVer (works with your patch now, but let's make sure it does not break).

pkg/ddevapp/ddevapp.go
```diff
@@ -1021,6 +1023,20 @@ func (app *DdevApp) GetDBImage() string {
 func (app *DdevApp) Start() error {
 	var err error
 
+	if app.DdevVersionConstraint != "" {
+		v, err := semver.NewVersion(versionconstants.DdevVersion)
+		if err != nil {
+			util.Warning("%v %v, ddev version constraint cannot be verified", err, versionconstants.DdevVersion)
```
jonaseberle, line 1029:
```suggestion
			util.Warning("%v %v, ddev version could not be parsed", err, versionconstants.DdevVersion)
```

rfay:
For the test, you don't need an old DDEV version, you need a future constraint. So you would try creating app (not starting it) with a number of constraints.

rfay:
TestConfigLoadingOrder is one of many tests that might work as a model, https://github.com/ddev/ddev/blob/c37da314af81d48db54fa774a975f1b68acc4409/pkg/ddevapp/config_test.go#L1283

It just does a NewApp() with different configs. In your case it would be the NewApp() that would reject or accept the constraint I think. (I haven't looked at the PR yet)

Unfortunately, features without tests fail silently during maintenance :(

pkg/ddevapp/ddevapp.go
hanoii: there's an error message around that coming from the server variable, I just wanted to emphasize that the intended goal was not going to work.

hanoii:
The code is in App start, I think I do have to start it up, but future constraints will just fail, so that's not a bad thing.. I'll give it a try and look at those tests. Thanks

rfay:
Code should probably move to the app verification (config verification) then to make things much easier.

hanoii:
Tests added, code refactored and moved! PR is ready for review!

pkg/ddevapp/config_test.go
```diff
@@ -607,12 +608,41 @@ func TestConfigValidate(t *testing.T) {
...
 	appName := app.Name
 	appType := app.Type
+	ddevVersion := versionconstants.DdevVersion
 
 	err = app.ValidateConfig()
 	if err != nil {
 		t.Fatalf("Failed to app.ValidateConfig(), err=%v", err)
 	}
 
+	app.DdevVersionConstraint = ">= 1.twentythree"
```
rfay, line 618: Could you please explore making these constraints more flexible? Otherwise we'll deal with needless breakage over the years going forward.

rfay (changes requested):
I haven't carefully tested this yet, but it looks like it's on the right track, thanks!

It's missing docs of course, so add to config.md. And normally every config item has a `ddev config` flag as well. 

I guess this is mostly for the future anyway, right? A current ddev running against a config.yaml with a constraint would ignore (or remove) it.

Also, the new constraint has to go into pkg/ddevapp/schema.json

pkg/ddevapp/config_test.go
hanoii: I didn't understand this. What do you mean? The constrains format comes straight from semver.

hanoii: On this test I made it so that I fails.

rfay: Sorry, I was referring to the whole of the section. Please make it so that this test doesn't have to be edited when v1.23.0 comes out, for example.

hanoii: It doesn't. I am also setting a specific ddev version so that tit will always fail.

hanoii: The first one will fail because it's just invalid

jonaseberle: I'd like to see `v1.22.3-11-g8baef014e` being tested since that's what I have in `DDEV_VERSION` when I compile from source.
For example `v1.22.3-11-g8baef014e >= 1.22`, `v1.22.3-11-g8baef014e < 1.23`

hanoii: @jonaseberle ok I added more tests. How do you get that version? I am getting `22be588e1-dirty` as version on my compiled `ddev`.

A warning here, `v1.22.3-11-g8baef014e` is considered a pre-release, so `v1.22.3-11-g8baef014e >= 1.22` will actually fail as the constraint, as it's defined by the semver module, excludes pre-releases by default. It will fail with any constraint that doesn't include pre-releases. `v1.22.3-11-g8baef014e >= 1.0` will also fail.

If you want to include pre-releases, you have to use a constraint that includes them:

`v1.22.3-11-g8baef014e >= 1.22.0-0` will pass.

Both things are tested.

https://github.com/Masterminds/semver#working-with-prerelease-versions

I will try documenting this.

hanoii:
> I guess this is mostly for the future anyway, right? A current ddev running against a config.yaml with a constraint would ignore (or remove) it.

Yes

hanoii:
>  And normally every config item has a ddev config flag as well.
Done

> Also, the new constraint has to go into pkg/ddevapp/schema.json
Done

> It's missing docs of course

- [ ] Still needs doing

pkg/ddevapp/config_test.go
jonaseberle: That's the VERSION we get in Makefile by `git describe --tags --always --dirty` when there are tags somewhere up in your commit history (and in my case nothing uncommitted, so not `-dirty`). IMHO we should make that format supported somehow.

rfay: `dirty` means that something isn't committed (`git status` shows uncommitted files)

rfay: `v1.22.3-11-g8baef014e` and `v1.22.3-11-g8baef014e-dirty` should already work with the semver comparisons if they're working right and being used right.

hanoii: As I am in my fork I don't have the tags so I don't get that. Ok, they work as long as you have a constrain that includes them. I am not if there's an actionable thing to do here, and this is very specific for dev verisions.

`v1.22.3-11-g8baef014e` with a constraint of `>= 1.22` will not allow ddev to run
`v1.22.3-11-g8baef014e` with a constraint of `>= 1.22.0-0` will.

If you would like `v1.22.3-11-g8baef014e` with a constraint of `>= 1.22` to be "allowed" I am not sure what to do there if we are going to use this version, which I think it's the version we should use.

If there's something specific for me to do around this I am happy to do it, if there's not and it's OK as it is, let's resolve it.

hanoii: I don't consider this a bad thing but I see how this can confuse users.

One alternative would be to strip pre-releases out of the version

hanoii: that would make any `v1.23.0-whatever` be `>=1.23`

hanoii: If this is desirable I can work on it

hanoii: I think this is the way to go actually

hanoii: the only problem there is that 

`v1.23.0-alpha1` with constrain of `>= v1.23.0-alpha2` will still work if I remove`-alpha1` from the version

🤔

hanoii: If you wish to play with it: https://go.dev/play/p/9-OII74ev21

hanoii: I think this is ok and can be documented, if a user is using an alpha/dev version he's advanced enough, but otherwise I'll wait for guidance.

hanoii: I am not quite sure if that means leaving the way it is or not :)

hanoii:
Added docs, this is ready for review again!

pkg/ddevapp/config_test.go
jonaseberle: >  v1.22.3-11-g8baef014e with a constraint of >= 1.22.0-0 will
I think that's the way to go then. Looks good from my side feature-wise.

hanoii:
I mentioned in the docs that this work on version v1.22.4 and above, if this is not included in that release, it can be edited

hanoii:
@rfay how do you restart a test like `Tests / tests (nginx-proxy, nginx-proxy) (pull_request) `?

buildkite as well? I restarted both ddev-macos but they haven't restarted that one.

rfay:
@hanoii I would have thought you'd have adequate privs. For a single failure:
image


For multiple failure:
image

hanoii:
@rfay nope:

Screenshot 2023-09-27 at 17 47 51

rfay:
Poked around and it looks like you need write access to the repo to run tests, which is hash. Ping me any time when you need a restart. I usually notice anyway.

rfay:
Sorry I'm slow to review. In Barcelona now, may get to it in the next few weeks, but also may not.

pkg/ddevapp/templates.go
```diff
@@ -93,6 +93,11 @@ const ConfigInstructions = `
 # If true, turns off the normal warning that says
 # "You have Mutagen enabled and your 'php' project type doesn't have upload_dirs set"
 
+# ddev_version_constraint: >= 1.22.4
```
rfay, line 96: We normally want to show the default first, then examples.

```suggestion
# ddev_version_constraint: ""
# Example:
# ddev_version_constraint: ">= 1.22.4"
```

rfay (changes requested):
I'm not very happy with the inability to test with valid DDEV versions, I know you struggled with this, and I've struggled with it in the past. 

> failed to validate config: ddev version v1.22.3-36-g3a4d62477 is not supported by this project (>=v1.21.6), use a supported version or update your project config.yaml's `ddev_version_constraint`

I'm still working on manual review.

pkg/ddevapp/config_test.go
```diff
@@ -607,12 +608,74 @@ func TestConfigValidate(t *testing.T) {
...
+	assert.Contains(err.Error(), "ddev version v1.22.3-11-g8baef014e is not supported by this project")
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
+
+	versionconstants.DdevVersion = "v1.22.3-11-g8baef014e"
+	app.DdevVersionConstraint = ">= 1.22.0-0"
+	err = app.ValidateConfig()
+	assert.NoError(err)
+	app.DdevVersionConstraint = ""
+	versionconstants.DdevVersion = ddevVersion
```
rfay, line 677: This would have been better as its own test, but will do for now.

rfay:
Current artifacts at at https://github.com/ddev/ddev/actions/runs/6327702142 for testing.

hanoii:
> failed to validate config: ddev version v1.22.3-36-g3a4d62477 is not supported by this project (>=v1.21.6), use a supported version or update your project config.yaml's ddev_version_constraint

I am sure you know or maybe gathered from the previous comments, but if you use the constraint of `>=1.21.6-0` your version will pass.

rfay:
That seems like trouble I've had with the docker-compose constraint in the past.

docs/content/users/configuration/config.md
```diff
@@ -103,6 +103,16 @@ Extra Debian packages for the project’s database container.
 
 Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the database container is built.
 
+## `ddev_version_constraint`
+
+You can configure a [ddev version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running ddev and prevent any command from running if it doesn't match.
```
mattstein, line 108: Always capitalize DDEV when it’s used as a product name, vs. `ddev` (in backticks) referencing the command. Also added a word attempting to distinguish the two “DDEV” instances in this sentence—which ideally would be pared down if possible.

```suggestion
You can configure a [DDEV version constraint](https://github.com/Masterminds/semver#checking-version-constraints) that will be validated against the running DDEV executable and prevent any command from running if it doesn't match.
```

docs/content/users/configuration/config.md
```diff
@@ -103,6 +103,16 @@ Extra Debian packages for the project’s database container.
...
+
+Note: this will only work with ddev versions above v1.22.4, older ddev versions will ignore this setting.
```
mattstein, line 110: Drop “Note:” which isn’t adding anything—or use an admonition if this needs urgent visual emphasis. Also capitalize DDEV again, fix punctuation, and remove redundant “DDEV”:

```suggestion
This will only work with DDEV versions above v1.22.4; older versions will ignore this setting.
```

mattstein (changes requested):

hanoii:
@rfay 
>That seems like trouble I've had with the docker-compose constraint in the past.

Yes, it's coming from how https://github.com/Masterminds/semver#checking-version-constraints does the validation.

PR #5380, opened 2023-09-26
feat: support Env writing without overwriting existing values

n8-dev:
## The Issue
(Surprised this hasn't been brought up before. 🤔 )

Currently in the `pkg/ddevapp/silverstripe.go` file it has logic to ensure you have some basic required `.env` details present.

The implementation actually overwrites some of these details locally.

Arguably `.env` values are specific and shouldn't necessarily be overwritten.

There should be a way to **create if it does not exist**, I couldn't see this in the `envfile.go` file.

### Current work around
Adding details you wish to persist to `.ddev/config.yaml` `web_environment: []`
This isn't a satisfactory solution as you shouldn't have to maintain two "env" files.

This PR solves this by adding a new function that allows writing to `.env` only when the value **does not** already exist.
It also includes a valid usage of it for Silverstripe sites.

Summary of Changes:
1. Added the function `WriteProjectEnvFileNoOverwrite` to `pkg/ddevapp/envfile.go`
2. Implementation in `silverstripe.go`

### Step 1
Get Silverstripe running via composer
```bash
mkdir my-silverstripe-app
cd my-silverstripe-app
ddev config --project-type=silverstripe --docroot=public --create-docroot
ddev composer create --prefer-dist --no-scripts silverstripe/installer -y
ddev start
ddev sake dev/build flush=all
```

This ensures a silverstripe site is created and running.

### Step 2
Edit the .env file eg
```env
SS_DATABASE_SERVER="foo"
SS_DATABASE_USERNAME="bar"
SS_DATABASE_PASSWORD="fizz"
SS_DATABASE_NAME="bang"
```

```bash
ddev start
```

Your `.env` values above should stay how it was above ⬆️ ✅ 

Without this PR they'd be changing to. i.e. ⬇️ ❌
```env
SS_DATABASE_SERVER="db"
SS_DATABASE_USERNAME="db"
SS_DATABASE_PASSWORD="db"
SS_DATABASE_NAME="db"
```

## Automated Testing Overview
I've never written Go before
Any help on whats needed here would be appreciated.

I would hazard that a new function means new tests

## Release/Deployment Notes
New function might be useful for other project types, worth mentioning in documentation?

### DIFF
--- a/pkg/ddevapp/envfile.go
+++ b/pkg/ddevapp/envfile.go
@@ -42,7 +42,7 @@ func WriteProjectEnvFile(envFilePath string, envMap map[string]string, envText s
 
 // WriteProjectEnvFileNoOverwrite writes the passed envText into the project-root .env file
 // with all items in envMap added that do not already exist
-func WriteProjectEnvFileNoOverwrite(envFilePath string, envMap map[string]string, envText string, overwrite) error {
+func WriteProjectEnvFileNoOverwrite(envFilePath string, envMap map[string]string, envText string) error {
 	// envFilePath := filepath.Join(app.AppRoot, ".env")
 	for k, v := range envMap {
 		// If the item is already in envText (checking via regex)
--- a/pkg/ddevapp/envfile.go
+++ b/pkg/ddevapp/envfile.go
@@ -39,3 +39,23 @@ func WriteProjectEnvFile(envFilePath string, envMap map[string]string, envText s
 	}
 	return nil
 }
+
+// WriteProjectEnvFileNoOverwrite writes the passed envText into the project-root .env file
+// with all items in envMap added that do not already exist
+func WriteProjectEnvFileNoOverwrite(envFilePath string, envMap map[string]string, envText string, overwrite) error {
+	// envFilePath := filepath.Join(app.AppRoot, ".env")
+	for k, v := range envMap {
+		// If the item is already in envText (checking via regex)
+		// do nothing, otherwise append it to the envText.
+		exp := regexp.MustCompile(fmt.Sprintf(`((^|[\r\n]+)%s)=(.*)`, k))
+		if !exp.MatchString(envText) {
+			envText = strings.TrimSuffix(envText, "\n")
+			envText = fmt.Sprintf("%s\n%s=%s\n", envText, k, v)
+		}
+	}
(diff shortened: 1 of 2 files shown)


n8-dev:
FYI @Firesphere 

Seen your other work on Silverstripe in here, many thanks 💙

Firesphere:
I don't believe this kind of exception for just Silverstripe should be implemented, it should be either a blanket for all, or nothing.

I believe he reason for overwriting is to ensure the default for any set-up "just works" out of the box.
If the `.env` is left alone, anyone with a currently-working but different set-up (own docker, other "snowflakes", etc.), would suddenly have a misconfigured/unclear configuration, and having to do the reverse of what you're proposing, but manually. A case I would expect to see more, than the other way around.

I am unsure of your reasoning, as I can't think of a common situation where this might happen.
Unless you explicitly want or need to connect to a completely different database continuously, I would think using the config.yml makes more sense.
Further, the need to connect to a database different from the dev environment, is quite uncommon.

If anything, I think that a feature like this would be better off as a config-like flag instead of a blanket "don't change anything", something among the lines of `ddev config --no-env` or something?

n8-dev:
Just quickly:
> I don't believe this kind of exception for just Silverstripe should be implemented, it should be either a blanket for all, or nothing.

I agree, but also don't work with any of the other supported frameworks and was erring on the caution here
- Proposing new functions require thought and care
- Silverstripe section I know about
- Other people will be more qualified to make the call on other frameworks
- Preventing this PR from being too big

I'll reply to everything else when I have more time.

rfay:
This does seem to break the silverstripe tests, https://github.com/ddev/ddev/actions/runs/6307596274/job/17159826556?pr=5380

rfay:
I haven't studied this carefully yet, but I'm pretty sure with .env people won't know what's wrong if we don't overwrite database variables, etc. .env files should be unique to an environment usually. 

I changed it to "draft" until the tests work.

n8-dev:
I haven't discarded this yet, its still percolating in my head.

In short, I think you're right Firesphere.

A config choice to decide is a better approach, I'll edit this to go that way.

rfay:
It sounds like this PR doesn't have a future right now, is that correct? I'm closing for now, but happy to reopen, and happy to continue the discussion here or elsewhere.

PR #5396, opened 2023-10-02
ci: change technique for PR artifact comment, fixes #5386

### LINKED ISSUE #5386: CI: Fix pr-artifacts-comment with link to artifacts

_No response_

### Expected Behavior

We should have a valid artifacts link in a PR

### Actual Behavior

The artifacts link either isn't created (like https://github.com/ddev/ddev/pull/5383) or doesn't get reliably updated, see https://github.com/ddev/ddev/issues/5193#issuecomment-1736862545) 

I guess we should remove or replace

https://github.com/ddev/ddev/blob/master/.github/workflows/pr-artifacts-comment.yml

That action is based on nightly.link, and has been problematic before. The original issue is
* https://github.com/oprypin/nightly.link/issues/37

But the suggestion there is "don't use what we have".

I think it's just going to have to go away, but it's so very useful when it's useful.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/5386

From what I can see, there are no problems with `nightly.link` itself, the failures for PR artifact comments are from GitHub Actions.

The current implementation of upstream [pr-comment.yml](https://github.com/oprypin/nightly.link/blob/5a34469cc7939d3c6a79ad5badbc62971145fa57/.github/workflows/pr-comment.yml) uses different technique to insert/update a bot comment (it looks for `<!-- bot: nightly-link -->` in the comments), but it doesn't work with forked PRs.

Also, DDEV is currently using the deprecated action `actions/github-script@v3`.

I combined two different workflows (see the changes) and used `actions/github-script@v6`.

I tested the new technique on this repo https://github.com/stasadev/ddev-pr-artifacts-comment

The change affects the current open PRs, a second bot comment like this https://github.com/stasadev/ddev-pr-artifacts-comment/pull/12 will be added (because it looks for `<!-- bot: nightly-link -->`  in the comments).

### DIFF
--- a/.github/workflows/pr-artifacts-comment.yml
+++ b/.github/workflows/pr-artifacts-comment.yml
@@ -11,49 +11,68 @@ jobs:
     if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/github-script@v3
-      with:
-        # This snippet is public-domain, taken from
-        # https://github.com/oprypin/nightly.link/blob/master/.github/workflows/pr-comment.yml
-        script: |
-          const {owner, repo} = context.repo;
-          const run_id = ${{github.event.workflow_run.id}};
-          const pull_head_sha = '${{github.event.workflow_run.head_sha}}';
-          const pull_user_id = ${{github.event.sender.id}};
-
-          const issue_number = await (async () => {
-            const pulls = await github.pulls.list({owner, repo});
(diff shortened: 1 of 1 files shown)


rfay:
Thanks! Look forward to looking a little closer. Sounds like you got to the root of it.

rfay (approved):
Thanks! AFAICT this is going to only affect the artifacts comment on PRs, which is (half the time at least) broken already. So if you're ready you can pull this one. Might as well pull it with `[skip ci]` in the title so as to not waste the full test run.

Issue #5397, opened 2023-10-03
error in docs for DDEV_HOSTNAME


joachim-n:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
```
</details>         

### Expected Behavior

On https://ddev.readthedocs.io/en/latest/users/extend/custom-commands/ it says:

> DDEV_HOSTNAME: Comma-separated list of FQDN hostnames

But when I did printenv in a command script, I see this:

```
DDEV_HOSTNAME=accounts-8.ddev.site
```

### Actual Behavior

There's only one domain.

rfay:
Does your project have more than one hostname? For example, are you using `additional_hostnames` or `additional_fqdns`? If not, you won't expect to see more than one.

rfay:
Closing as I think the behavior is as advertised, but happy to continue the conversation.

PR #5398, opened 2023-10-03
fix: pass domain/project name to `ddev sequelace`, fixes #5394

### LINKED ISSUE #5394: Impossible to determine which project a SequelAce window belongs to

```
```
</details>         

### Expected Behavior

The SequelAce window should give me some indication of which project it is connected to.

### Actual Behavior

I've used `ddev sequelace` in several projects.
I now can't tell which SequelAce window is for which project, because the window title is the same for all of them.

### Steps To Reproduce

1. Do `ddev sequelace` in two projects
2. Observe window titles are identical

### Anything else?

_No response_

joachim-n:
## The Issue

Fixes #5394. The domain name is sent to SequelAce instead of 127.0.0.1, which means that it shows in the SequelAce window title.

See #5397 however -- this was the only env var I could see which was suitable, but the docs for it are suspicious. An array here would clearly be wrong.

Do `ddev sequelace`. The new SequelAce window should show the domain, allowing the project to be identified.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
@@ -10,7 +10,7 @@
 
 DATABASE="${1:-db}"
 
-query="mysql://root:root@${DDEV_HOSTNAME}:${DDEV_HOST_DB_PORT}/${DATABASE}"
+query="mysql://root:root@$${DDEV_PROJECT}.${DDEV_TLD}:${DDEV_HOST_DB_PORT}/${DATABASE}"
 
 set -x
 open "$query" -a "/Applications/Sequel Ace.app/Contents/MacOS/Sequel Ace"
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
@@ -10,7 +10,7 @@
 
 DATABASE="${1:-db}"
 
-query="mysql://root:root@$${DDEV_PROJECT}.${DDEV_TLD}:${DDEV_HOST_DB_PORT}/${DATABASE}"
+query="mysql://root:root@${DDEV_PROJECT}.${DDEV_TLD}:${DDEV_HOST_DB_PORT}/${DATABASE}"
 
 set -x
 open "$query" -a "/Applications/Sequel Ace.app/Contents/MacOS/Sequel Ace"
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
@@ -10,7 +10,7 @@
 
 DATABASE="${1:-db}"
 
-query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/${DATABASE}"
+query="mysql://root:root@${DDEV_HOSTNAME}:${DDEV_HOST_DB_PORT}/${DATABASE}"
 
 set -x
 open "$query" -a "/Applications/Sequel Ace.app/Contents/MacOS/Sequel Ace"


pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
```diff
@@ -10,7 +10,7 @@
 
 DATABASE="${1:-db}"
 
-query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/${DATABASE}"
+query="mysql://root:root@${DDEV_HOSTNAME}:${DDEV_HOST_DB_PORT}/${DATABASE}"
```
stasadev, line 13: I would like to suggest `${DDEV_PROJECT}`. Unique and simple because it is the name of your project.

```suggestion
query="mysql://root:root@${DDEV_PROJECT}:${DDEV_HOST_DB_PORT}/${DATABASE}"
```

joachim-n: DDEV_PROJECT won't work AFAICT because it's just the project name. This needs to be the domain.

rfay: But the bottom line is you want something descriptive that is guaranteed to resolve to 127.0.0.1. It would be better if SequelAce had a command-line arg. Are you sure it doesn't? They're very responsive...

joachim-n: It sounds like it doesn't: https://github.com/Sequel-Ace/Sequel-Ace/issues/189
And anyway, this isn't using SequelAce directly, but using the `open` command, passing SequelAce to its -a flag. It doesn't look like we can pass more to the app: https://ss64.com/osx/open.html

rfay: Probably `${DDEV_PROJECT}.${DDEV_TLD}` will always work for you.

joachim-n: DDEV_PRIMARY_URL won't work because it includes the http://

joachim-n: DDEV_PROJECT -- can we guarantee that there is always a domain of the form  ${DDEV_PROJECT}.${DDEV_TLD} for a project? Could the user have configured it differently?

stasadev: I believe it is a reliable way.

Or use `${DDEV_SITENAME}.${DDEV_TLD}`, the same but perhaps more descriptive.

Also, you can check the values of these variables yourself with a custom command:

- https://github.com/ddev/ddev/issues/3636#issuecomment-1055509594

joachim-n: The docs for DDEV_SITENAME are identical to those for DDEV_PROJECT:

> Project name, like d8composer

rfay: I think DDEV_SITENAME and DDEV_PROJECT will always be the same. 

> can we guarantee that there is always a domain of the form ${DDEV_PROJECT}.${DDEV_TLD} for a project? 

DDEV insists on either having it in hosts or via DNS.

You could also easily use `${DDEV_PRIMARY_URL#*//}` which would be fine. 

The only problem is the situation where there's no internet and somebody hasn't already been forced to `ddev start` so the hostname hasn't been added to /etc/hosts. But in that case their browser doesn't work either.

joachim-n: Fixed.

pkg/ddevapp/global_dotddev_assets/commands/host/sequelace
```diff
@@ -10,7 +10,7 @@
 
 DATABASE="${1:-db}"
 
-query="mysql://root:root@127.0.0.1:${DDEV_HOST_DB_PORT}/${DATABASE}"
+query="mysql://root:root@$${DDEV_PROJECT}.${DDEV_TLD}:${DDEV_HOST_DB_PORT}/${DATABASE}"
```
stasadev, line 13: There is an extra `$`.

P.S. I've noticed that in some IDEs you select a Bash variable, it doesn't select the `$`, and when you paste something it doesn't remove the second `$`.

rfay (approved):
This is an interesting approach and it works. I guess we should go with it!

rfay:
Thanks for all the good work on this!

PR #5405, opened 2023-10-08
refactor: remove support for nodejs version 14

amayer5125:
Node 14 went end of life in April 2023. New versions of the Node installer do not support version 14.

## The Issue

NodeJS version 14 is [end of life](https://endoflife.date/nodejs).

This pull request removes support for version 14.

## Release/Deployment Notes

Users using Node 14 will need to upgrade.

### DIFF
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -194,7 +194,7 @@ func TestConfigSetValues(t *testing.T) {
 	useDNSWhenPossible := false
 	timezone := "America/Chicago"
 	webEnv := "SOMEENV=some+val"
-	nodejsVersion := "14"
+	nodejsVersion := "16"
 	defaultContainerTimeout := 300
 
 	args := []string{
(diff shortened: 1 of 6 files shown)


rfay:
Thanks, don't forget 
* https://ddev.readthedocs.io/en/latest/users/configuration/config/#nodejs_version
* https://ddev.readthedocs.io/en/latest/users/usage/commands/#config (which is already incorrect, mentioning 16. We should be able to do better words there. Just change to "if you don't want to use the default version"

rfay (approved):
This looks perfect to me, thanks. After it goes in you can rebase and finish the other one.

stasadev (changes requested):
Found one more place with `14`:

https://github.com/ddev/ddev/blob/d3e6e5b54296fa3ec421da9bd9424db71e01aae0/pkg/ddevapp/templates.go#L66

Please update and it will be ready to go.

rfay:
You can push that one more commit with `[skip ci]` in the title - it doesn't need another test run.

amayer5125:
@rfay change pushed.

stasadev (approved):

PR #5400, opened 2023-10-05
fix: update node install method, fixes #5363

amayer5125:
## The Issue

* https://github.com/ddev/ddev/issues/5363

The old method (using setup_X script) is deprecated.

## How This PR Solves The Issue
Use the new method outlined in [https://github.com/nodesource/distributions](https://github.com/nodesource/distributions)

## Related Issue Link(s)

* https://github.com/ddev/ddev/issues/5363

PR #5412, opened 2023-10-10
fix: warn about obsolete nodejs_version instead of failing

rfay:
## The Issue

In
* https://github.com/ddev/ddev/pull/5405 and 
* https://github.com/ddev/ddev/pull/5400

we invalidated `nodejs_version: 14` because the upstream package manager does not support it any more.

However, we don't want to break people who happen to have that in their config. 

* Allow incorrect config but warn about it
* Empty `nodejs_version` if it's the default, to try to prevent this in the future. This means people will only configure it if they want something other than DDEV's default. This is slightly intrusive, but not too much.

Here's what happens. It's not as pretty as it should be but it gets the job done:

image

Set `nodejs_version: 14` and `ddev start`. Use `ddev config --auto` to fix.

## Automated Testing Overview

No new tests - the other 2 PRs handled everything.

## Release/Deployment Notes

* Release requires announcement of the v14 removal

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -350,6 +350,8 @@ Node.js version for the web container’s “system” version.
 | -- | -- | --
 | :octicons-file-directory-16: project | current LTS version | Can be `16`, `18`, or `20`.
 
+There is no need to configure the `nodejs_version` unless you want to use a major version that is not the default.
+
 `nvm` is also available inside the container and via [`ddev nvm`](../usage/commands.md#nvm), and can be set to any valid version including much older ones.
 
 ## `omit_containers`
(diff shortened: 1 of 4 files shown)


stasadev (approved):
Tested different commands, `ddev config` removes deprecated version as expected.

And other commands are chatty enough to get the user to take action.

```
# and deprecated line nodejs_version: "14" is removed
$ ddev config --nodejs-version 14
Node.js version '14' is not currently supported by DDEV, ignoring and using default version '18'.
Use `ddev config --auto` to fix, or use `ddev nvm` to support an arbitrary Node.js version.

# and deprecated line nodejs_version: "14" is removed
$ ddev config --omit-containers db
WARN[0000] Node.js version '14' is not currently supported by DDEV, ignoring and using default version '18'.
...

$ ddev start
WARN[0000] Node.js version '14' is not currently supported by DDEV, ignoring and using default version '18'.
...

$ ddev ssh
WARN[0000] Node.js version '14' is not currently supported by DDEV, ignoring and using default version '18'.
...

$ ddev exec node -v
WARN[0000] Node.js version '14' is not currently supported by DDEV, ignoring and using default version '18'.
...
v18.18.0
```

Issue #5365, opened 2023-09-22
Remove support for Segment.io instrumentation


rfay:
Amplitude seems to be working fine. We can pull the plug on Segment.io instrumentation.

Issue #5254, opened 2023-08-11
"Unable to gather manifests" when installing add-on with dependencies from local directory


JanoPL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
.ddev/addon-metadata: no such file or directory
```
</details>         

### Expected Behavior

to install addon from directory with command ```ddev get <local dir>```

when the addon install.yaml has configured option dependencies

whitout dependecies key in yaml file is working fine.

example for adminer addon install.yaml
```yaml 
name: adminer

# list of files and directories listed that are copied into project .ddev directory
project_files:
- docker-compose.adminer.yaml
- adminer

dependencies:
  - db

global_files:
- commands/host/launcha
```

![image](https://github.com/ddev/ddev/assets/37768741/353a24e6-898c-41c0-87fe-19c41e572245)

### Actual Behavior

it can't create addon-metadata directory when addon is installed locally for example: ddev get /home/user/ddev-adminer/
Unable to gather manifests: open /home/user/test-app/.ddev/addon-metadata: no such file or directory

![image](https://github.com/ddev/ddev/assets/37768741/ab37914e-1365-4890-bddb-dc4bb778e7e4)

### Steps To Reproduce

1. generate new config or using existing for ddev
2. check if addon-metada not exist in .ddev directory
3. clone addon repo to local directory
4. execute ddev get addon

commands: 
1. ```ddev config```
2. ```[ ! -d "/.ddev/addon-metadata/" ] && echo "Directory /.ddev/addon-metadata/ DOES NOT exists."```
3. ```git clone git@github.com:ddev/ddev-adminer.git```
4. ```ddev get <local path>/ddev-adminer```

rfay:
I do this all the time in testing. Looks like you're using an absolute URL when you should be using a relative URL?

rfay:
```
$ ddev get /Users/rfay/workspace/ddev-elasticsearch/

Installing project-level components:
👍 docker-compose.elasticsearch.yaml
Overwriting existing manifest file /Users/rfay/workspace/d10/.ddev/addon-metadata/ddev-elasticsearch/manifest.yaml

Installed DDEV add-on /Users/rfay/workspace/ddev-elasticsearch/, use `ddev restart` to enable.
```

Are you perhaps giving it an invalid path?

rfay:
```
[...]

PR #5420, opened 2023-10-11
fix: resolve inappropriate creation of .ddev/addon-metadata, fixes #5419

### LINKED ISSUE #5419: Inappropriate creation of .ddev/addon-metadata in current directory

```
```
</details>         

### Expected Behavior

Spun off from
* #5342 

Spurious .ddev/addon-metadata shouldn't be created in the current directory

### Actual Behavior

Examples:

* Create `.ddev/commands/web/echo-something`

```
#! /usr/bin/env bash

VITE_PROJECT_DIR=${VITE_PROJECT_DIR}

echo $VITE_PROJECT_DIR
```

```
$ rm -rf .ddev && sleep 1 && pwd && ddev echo-something && sleep 1 && ls -ld .ddev
total 0
drwxr-xr-x  2 rfay  staff  64 Oct 11 17:23 addon-metadata
```

* `ddev get --remove` not in main project directory:

```
rfay@rfay-tag1-m1:~/workspace/d10/web$ pwd && rm -rf .ddev && ddev get --remove solr && ls -ld .ddev
/Users/rfay/workspace/d10/web
Removed add-on solr
drwxr-xr-x  3 rfay  staff  96 Oct 11 17:25 .ddev
```

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #5419 

The PersistentPostRun function for the root command handles instrumentation. It was using NewApp(), which will create a new ddevapp right there in the current dir. I'm pretty sure it should be using GetActiveApp() and that will solve this.

The actual problem can also be solved by turning off instrumentation, because that's where the app is being created.

Use the techniques from
* #5419 

Also try it with and without instrumentation.

## Automated Testing Overview

It may deserve a test. I'm not sure a test is easy.

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -101,7 +101,6 @@ Support: https://ddev.readthedocs.io/en/stable/users/support`,
 		}
 	},
 	PersistentPostRun: func(cmd *cobra.Command, args []string) {
-		// TODO: Remove once it's activated directly in ddevapp
 		if instrumentationApp == nil {
 			app, err := ddevapp.GetActiveApp("")
 			if err == nil {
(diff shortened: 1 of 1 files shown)


rfay:
There are some issues here:

* It may be that there are situations where there is no app to be found. So that situation probably needs a workaround, a dummy app. 
* Why does just creating the app result in the addon-metadata being written?

stasadev:
>Why does just creating the app result in the addon-metadata being written?

- https://github.com/ddev/ddev/issues/5254

rfay:
Thanks for that. I see
* https://github.com/ddev/ddev/pull/5271

https://github.com/ddev/ddev/pull/5271/files#diff-df6266d1f671740d9fa1aab2239f8031332b53df789f714809e3aea2753cd294R50-R52

```
func GetInstalledAddons(app *DdevApp) []AddonManifest {
	metadataDir := app.GetConfigPath(AddonMetadataDir)
	err := os.MkdirAll(metadataDir, 0755)
```

So we must be calling that. But now that we understand better, it should be something the debugger can show us.

stasadev:
Yes, that's it. I see the place now.

https://github.com/ddev/ddev/blob/7be069cdf26303acced3e193804f7ec8e0428cc6/pkg/ddevapp/amplitude_project.go#L49

rfay:
Note that all this code is apparently due to be removed shortly, and is only involved with the segment way of sending instrumentation. 

* https://github.com/ddev/ddev/issues/5365

PR #5305, opened 2023-08-29
fix: make project network external, fixes #5193

stasadev:
## The Issue

* #5193 

Sometimes when you start a project, multiple project networks are created (e.g. `ddev-npg_default`) which prevents the project from starting.

```
Error response from daemon: network ddev-npg_default is ambiguous (16 matches found on name)
```

It doesn't happen all the time, but it's really annoying.

## How This PR Solves The Issue

The idea is that if the project's docker network is created before doing the usual stuff, it can prevent the problem from occurring.

Let's make the project's docker network external (just like the `ddev_default` global network).

## Manual Testing Instructions

This PR must *not* break existing features to be considered something valuable to merge.

There is no reliable way to test this, because the issue is related to `docker-compose` (not DDEV itself), and it occurs randomly.

## Related Issue Link(s)

- https://github.com/ddev/ddev/issues/5193

## Release/Deployment Notes

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5305"></a>

PR #5422, opened 2023-10-12
feat: assign label to external docker network, for #5305

stasadev:
## The Issue

- #5305

The change proposed in #5305 has the side effect: when a user downgrades DDEV without `ddev poweroff` (in the new version with #5305), the project cannot be restarted because the external project network cannot be removed automatically https://github.com/ddev/ddev/pull/5305#pullrequestreview-1668407006:

`time="2023-10-10T19:01:03+02:00" level=warning msg="a network with name ddev-junk_default exists but was not created by compose.\nSet external: trueto use an existing network" network ddev-junk_default was found but has incorrect label com.docker.compose.network set to "" Failed to restart junk: composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-junk docker-compose -f /Users/rfay/workspace/junk/.ddev/.ddev-docker-compose-full.yaml up -d', action='[]', err='exit status 1', stdout='', stderr='time="2023-10-10T19:01:03+02:00" level=warning msg="a network with name ddev-junk_default exists but was not created by compose.\nSetexternal: true to use an existing network"
`

This PR introduces a new mechanism for deletion of external networks by adding label `com.ddev.platform` and then filtering them by this label.

I decided not to introduce a new label, because `com.ddev.platform` is already used for containers, and can be reused for other Docker resources:

https://github.com/ddev/ddev/blob/0ceead660b99ef1b7a39d0ea7b9a23cf726c5ef7/pkg/ddevapp/utils.go#L163-L179

1. `ddev start`
2. `docker inspect ddev_default` - to see the label `com.ddev.platform`
3. `ddev poweroff`
4. `docker network ls` - should show no DDEV networks

## Release/Deployment Notes

We can add this change in the next release v1.22.4 and #5305 can be added right after that.
In the future, if users experience issues with #5305, they can safely revert to v1.22.4.

### DIFF
--- a/pkg/ddevapp/poweroff.go
+++ b/pkg/ddevapp/poweroff.go
@@ -50,14 +50,17 @@ func PowerOff() {
 	if err := RemoveSSHAgentContainer(); err != nil {
 		util.Error("Failed to remove ddev-ssh-agent: %v", err)
 	}
-	// Remove current global network ("ddev") plus the former "ddev_default"
-	removals := []string{"ddev_default"}
-	for _, networkName := range removals {
-		err = dockerutil.RemoveNetwork(networkName)
-		_, isNoSuchNetwork := err.(*docker.NoSuchNetwork)
-		// If it's a "no such network" there's no reason to report error
-		if err != nil && !isNoSuchNetwork {
-			util.Warning("Unable to remove network %s: %v", "ddev", err)
+
+	// Remove all global networks created with DDEV
+	removals, err := dockerutil.FindNetworksWithLabel("com.ddev.platform")
+	if err == nil {
+		for _, network := range removals {
+			dockerutil.RemoveNetworkWithWarningOnError(network.Name)
 		}
+	} else {
+		util.Warning("Unable to run dockerutil.FindNetworksWithLabel(): %v", err)
 	}
+
+	// Remove old network from DDEV before v1.22.4
+	dockerutil.RemoveNetworkWithWarningOnError(dockerutil.NetName)
 }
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/poweroff.go
```diff
@@ -50,14 +50,17 @@ func PowerOff() {
 	if err := RemoveSSHAgentContainer(); err != nil {
 		util.Error("Failed to remove ddev-ssh-agent: %v", err)
 	}
-	// Remove current global network ("ddev") plus the former "ddev_default"
-	removals := []string{"ddev_default"}
```
rfay, line 54: I think we should probably still remove ddev_default by name here, in case it was created without label? Open to discussion.

rfay (changes requested):
Minor comments for discussion, but it appears that `ddev_default` does not get deleted even in the happy path. I did a `ddev start`, it wanted a `poweroff`, did it, everything came up fine, but then after `ddev stop`, and router stopped, the `ddev_default` network was still there. And it did in fact have the label on it:

```
        "Labels": {
            "com.ddev.platform": "ddev"
        }
```

The project-related container was in fact deleted.

pkg/ddevapp/poweroff.go
stasadev: This instruction to remove `ddev_default`  is below `dockerutil.RemoveNetworkWithWarningOnError(dockerutil.NetName)`

stasadev:
>after ddev stop, and router stopped, the ddev_default network was still there.

But `ddev_default` should be removed only after `ddev poweroff`.

This is how it worked before.

rfay:
Sorry, I thought `ddev_default` was removed when the `ddev-router` stopped, I'll give it another try!

I thought you preferred not to work on the weekend :)

stasadev:
I'm doing other things on my laptop right now, so I had time to answer here :)

rfay (approved):
OK, it looks good, it tests well. Thanks! And this will be in v1.22.4, then we can move on to the one you've spent so much work on, which of course will require a slightly update. Thanks for your patience with this.

PR #5424, opened 2023-10-13
fix: always include pre-releases on version constraint checking, fixes feedback on #4630

hanoii:
## The Issue

With the new version constraint #5379, it was reported on both the PR as well as on https://github.com/ddev/ddev/issues/4630#issuecomment-1761998687 that it's not clear or is confusing why a built version like `v1.22.3-24-g3d90d5eeb` is not accepted with a constraint of `>= v1.22`. This is because of https://github.com/Masterminds/semver#checking-version-constraints.

This assumes that pre-releases are always included. While this invalidates any means of "excluding" pre-releases with the constraint, for general usage this will lead to less support issues and matches other less strict constraints comparisons.

`ddev config --ddev-version-constraint=">= 1.22"` should work.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -434,6 +434,8 @@ func (app *DdevApp) ValidateConfig() error {
 	if app.DdevVersionConstraint != "" {
 		constraint := app.DdevVersionConstraint
 		if !strings.Contains(constraint, "-") {
+			// Allow preleases to be included in the constraint validation
+			// @see https://github.com/Masterminds/semver#working-with-prerelease-versions
 			constraint += "-0"
 		}
 		c, err := semver.NewConstraint(constraint)
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -447,7 +447,8 @@ func (app *DdevApp) ValidateConfig() error {
 		v, err := semver.NewVersion(versionconstants.DdevVersion)
 		if err == nil {
 			if !c.Check(v) {
-				return fmt.Errorf("ddev version %v is not supported by this project (%v), use a supported version or update your project config.yaml's `ddev_version_constraint`", versionconstants.DdevVersion, app.DdevVersionConstraint)
+				return fmt.Errorf("This project has a DDEV version constraint of '%s' and the version of DDEV you are using ('%s') does not meet the constraint. Please update the `ddev_version_constraint` in your .ddev/config.yaml or use a version of DDEV that meets the constraint.", app.DdevVersionConstraint, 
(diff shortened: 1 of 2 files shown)


rfay:
Should we have just used comparison instead of constraint, for more predictable behavior?

pkg/ddevapp/config.go
```diff
@@ -442,7 +447,7 @@ func (app *DdevApp) ValidateConfig() error {
 		v, err := semver.NewVersion(versionconstants.DdevVersion)
 		if err == nil {
 			if !c.Check(v) {
-				return fmt.Errorf("ddev version %v is not supported by this project (%v), use a supported version or update your project config.yaml's `ddev_version_constraint`", versionconstants.DdevVersion, c)
+				return fmt.Errorf("ddev version %v is not supported by this project (%v), use a supported version or update your project config.yaml's `ddev_version_constraint`", versionconstants.DdevVersion, app.DdevVersionConstraint)
```
rfay, line 450: We need to improve this. I was pretty confused by

```
$ ddev config --ddev-version-constraint=">=v1.22.4"
You are reconfiguring the project at /Users/rfay/workspace/d10.
Configuring a 'drupal10' codebase with docroot 'web' at /Users/rfay/workspace/d10/web
failed to validate config: ddev version v1.22.3-35-g4e684a83a is not supported by this project (>=v1.22.4), use a supported version or update your project config.yaml's `ddev_version_constraint`
```

```suggestion
				return fmt.Errorf("This project has a DDEV version constraint of '%s' and the version of DDEV you are using ('%s') does not meet the constraint. Please update the `ddev_version_constraint` in your .ddev/config.yaml or use a version of DDEV that meets the constraint.", app.DdevVersionConstraint, 
 versionconstants.DdevVersion)
```

rfay (changes requested):
This seems to be on the right track. We need to give improved guidance to the user. Thanks for all the work on this!

hanoii:
@rfay 
> We need to give improved guidance to the user.

Is this your text change suggestion? That's committed, not sure if you meant something else.

rfay:
> Is this your text change suggestion? That's committed, not sure if you meant something else.

Yes, thanks, I made the comment and then realized I could just suggest.

Sorry I capitalized wrong!

hanoii:
tests are failing on the last commit but it was really a comment change, and looking at them looks unrelated

rfay:
It was just the Colima tests, which are always fragile (because of the Github-hosted runner). I restarted it. You should have been able to restart also, right?

deviantintegral:
I just tested the build in this issue and this feature works nicely. Thank you!

PR #5426, opened 2023-10-14
fix: ddev describe shouldn't show router when disabled, fixes #5291

### LINKED ISSUE #5291: `ddev describe` shows `router: traefik` when no router.
If no router (omit_containers: ddev-router, which would be better as `router: none`) `ddev describe` shows `router: traefik`

There's no router. For example, on codespaces.

rfay:
## The Issue

* #5291 

`ddev describe` shows the router as traefik even if `omit_containers[ddev-router]`

Change it

Test with router disabled


rfay (commented):
I manually tested and looked at this again and it seems fine.

PR #5432, opened 2023-10-14
docs: make sure sudo is done early in linux installation, fixes #5327

### LINKED ISSUE #5327: docs: Debian/Ubuntu DDEV setup can fail because sudo eats the later output

_No response_

### Expected Behavior

People should be able to paste and use the Debian/Ubuntu setup in https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#debianubuntu

### Actual Behavior

If they haven't already issued a sudo (or if the sudo timeout is disabled) then the first line's sudo (`sudo install -m 0755 -d /etc/apt/keyrings`) eats part of the second line (`curl -fsSL https://pkg.ddev.com/apt/gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/ddev.gpg > /dev/null`) hoping for a password there. 

The result is that the key doesn't get installed, so people see this:

```
W: GPG error: https://pkg.ddev.com/apt * InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY C16711FC1179C7CA
E: The repository 'https://pkg.ddev.com/apt * InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

Probably the easiest way to fix this is with a simple script and have them sudo the script.

### Steps To Reproduce

With sudo not ready yet (hasn't already asked for a password), execute that whole stanza. 

### Anything else?

Thanks to @jonpugh for working this through.

rfay:
## The Issue

* #5327  

The docs suggestion does `sudo` in a place where it can eat important info. Do it early and rely on the cached status.

## Manual Testing

- [ ] Go through full Debian installation using copy/paste from this

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -54,14 +54,17 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ```bash
     # Add DDEV’s GPG key to your keyring
+    sudo sh -c 'echo ""'
     sudo install -m 0755 -d /etc/apt/keyrings
     curl -fsSL https://pkg.ddev.com/apt/gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/ddev.gpg > /dev/null
     sudo chmod a+r /etc/apt/keyrings/ddev.gpg
 
     # Add DDEV releases to your package repository
+    sudo sh -c 'echo ""'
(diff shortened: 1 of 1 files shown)


rfay:
I manually tested this on Fedora and Ubuntu 22.04, and it seems to be fine.

rfay:
I manually tested this on Fedora and Ubuntu 22.04, and it seems to be fine.

PR #5436, opened 2023-10-16
feat: add output for removed network, for #5305

stasadev:
## The Issue

When you start a project, you always see `Network ddev_default created`.

But when you run `ddev poweroff`, there is no indication at all that `ddev_default` is removed.

With docker-compose you can see the output for the project's network when it was removed, but when #5305 comes in, there's no output for that sort of thing at all.

Outputs the message `Network ddev_default removed` (and for any other removed external DDEV created network).

1. `ddev start`
2. `ddev poweroff` - see `Network ddev_default removed`

## Related Issue Link(s)

- https://github.com/ddev/ddev/pull/5305


rfay (approved):
Seems harmless to me, thanks!

PR #5438, opened 2023-10-17
feat: use the stable release version as part of the HEAD version on homebrew, fixes #5433

### LINKED ISSUE #5433: `brew install --HEAD` should provide a canonical DDEV/git version, not `vHEAD-<hash>`
In working with 
* #4630, 
* #5379 and 
* #5424

I installed HEAD with `brew unlink ddev && brew install ddev/ddev/ddev --HEAD --fetch-HEAD`. 

I get:
```
# ddev --version
ddev version vHEAD-0ceead6
```

This obviously will bypass any constraint as it's not a valid semver. This is not an important issue I think, just something I saw by chance, however, I wonder how hard would it be for the HEAD version to be something more similar to `v1.22.3-hash` or whatever pre-release scheme you can think best for HEAD. This would at least allow it to work with `ddev_version_constraint`.

hanoii:
I thought about this and I think this will work. I am unsure how to test it though.

I did edit the homebrew formulae locally and and tried adding the version and it works ok.

And out of curiosity, what does COMMIT= do as part of the make? 

Fixes #5433

### DIFF
--- a/.ci-scripts/bump_homebrew.sh
+++ b/.ci-scripts/bump_homebrew.sh
@@ -52,7 +52,7 @@ class Ddev < Formula
     sha256 cellar: :any_skip_relocation, arm64_big_sur: "${MACOS_ARM64_BOTTLE_SHA}"
   end
   def install
-    system "make", "VERSION=v#{version}", "COMMIT=v#{version}"
+    system "make"
     system "mkdir", "-p", "#{bin}"
     if OS.mac?
         if Hardware::CPU.arm?
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -188,7 +188,7 @@ brews:
         os = OS.mac? ? "darwin" : "linux"
         arch = Hardware::CPU.arm? ? "arm64" : "amd64"
         system "mkdir", "-p", "#{bin}"
-        system "make", "VERSION=v{{.Version}}-#{version}", "COMMIT=v#{version}"
+        system "make"
         system "cp", ".gotmp/bin/" + os + "_" + arch + "/ddev", "#{bin}/ddev"
     else
         bin.install "ddev"
@@ -223,7 +223,7 @@ brews:
         os = OS.mac? ? "darwin" : "linux"
         arch = Hardware::CPU.arm? ? "arm64" : "amd64"
         system "mkdir", "-p", "#{bin}"
-        system "make", "VERSION=v{{.Version}}-#{version}", "COMMIT=v#{version}"
+        system "make"
         system "cp", ".gotmp/bin/" + os + "_" + arch + "/ddev", "#{bin}/ddev"
     else
         bin.install "ddev"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -188,7 +188,7 @@ brews:
         os = OS.mac? ? "darwin" : "linux"
         arch = Hardware::CPU.arm? ? "arm64" : "amd64"
         system "mkdir", "-p", "#{bin}"
-        system "make", "VERSION=v#{version}", "COMMIT=v#{version}"
+        system "make", "VERSION=v{{.Version}}-#{version}", "COMMIT=v#{version}"
         system "cp", ".gotmp/bin/" + os + "_" + arch + "/ddev", "#{bin}/ddev"
     else
         bin.install "ddev"
@@ -223,7 +223,7 @@ brews:
         os = OS.mac? ? "darwin" : "linux"
         arch = Hardware::CPU.arm? ? "arm64" : "amd64"
         system "mkdir", "-p", "#{bin}"
-        system "make", "VERSION=v#{version}", "COMMIT=v#{version}"
+        system "make", "VERSION=v{{.Version}}-#{version}", "COMMIT=v#{version}"
         system "cp", ".gotmp/bin/" + os + "_" + arch + "/ddev", "#{bin}/ddev"
     else
         bin.install "ddev"


rfay:
To test, fork ddev/homebrew-ddev into hanoii/homebrew-ddev and then `brew install --HEAD --fetch-head hanoii/ddev/ddev`

hanoii:
@rfay 

> To test, fork ddev/homebrew-ddev into hanoii/homebrew-ddev and then `brew install --HEAD --fetch-head hanoii/ddev/ddev`

but this would allow me to test editing the formula, not the releasing with gorelease, wouldn't it? I've already tested this by doing `brew edit ddev` and it works ok.

rfay:
I don't actually understand how this can work, can you explain it?

hanoii:
> I don't actually understand how this can work, can you explain it?

Well, my understanding was (and I just checked on https://github.com/ddev/homebrew-ddev/commits/master) is that a new Formulae is pushed to that repo each time a new ddev is released, so I am adding the latest released version to the `VERSION=` used in the make when head is built. It's a go template so `"VERSION=v{{.Version}}-#{version}" would become `"VERSION=v1.23.4-#{version}"` on your next release.

hanoii:
On another note, the wsl2 buildkite is failing since yesterday, it couldn't because of this PR, could it? I restarted it many times, and just did it again.

rfay:
> the wsl2 buildkite is failing since I pushed this, it couldn't because of this PR, could it?

No, and you can just ignore it anyway, since it doesn't test this. (None of the buildkite formulas test it, they just try to shortcut it but sometimes something goes wrong.)

hanoii:
@rfay just double checking you saw https://github.com/ddev/ddev/pull/5438#issuecomment-1768310500

hanoii:
Hi @rfay , just double checking you saw https://github.com/ddev/ddev/pull/5438#issuecomment-1768310500 as well as the comment you replied.

rfay:
I've been traveling the last 3 weeks, this week at Drupalcon Lille, so will eventually catch up. But it would be lovely if you tried it with `brew install --HEAD --fetch-head hanoii/ddev/ddev`

hanoii:
@rfay but trying will not do anything for this PR, as this needs goreleaser to do its thing.. not sure what do you need me to test

rfay:
goreleaser will do exactly what you would do with this recipe in your homebrew fork. If it works there, it's proven to work here. But we can't test goreleaser without a release.

rfay:
Thanks so much for chasing this over multiple PRs and getting it sorted out. 

So the current version would be something like `v1.23-4-HEAD-hash1234`, which is valid, that's good.

The sad thing is that a real version would be `v1.22.3-44-g907fea6e3` (today's HEAD), which would then be comparable to other things properly.

rfay (changes requested):
I did what you suggested here, using rfay/ddev/ddev, and the result is 
```
$ ddev --version
ddev version v{{.Version}}-HEAD-b24f6e5
```

Not at all what we want. 

You can compare with https://github.com/rfay/homebrew-ddev/blob/master/Formula/ddev.rb

and experiment with `brew install --HEAD --fetch-head rfay/ddev/ddev`

rfay (approved):
Please take a look and see if my commit fixes it for you. I think we just needed to use `make` as-is, without any extras. Not even sure where that may have come from. 

You can test with `brew uninstall -f ddev && brew install --HEAD --fetch-head rfay/ddev/ddev && ddev --version`

hanoii:
> I did what you suggested here, using rfay/ddev/ddev, and the result is

```
$ ddev --version
ddev version v{{.Version}}-HEAD-b24f6e5
Not at all what we want.
```

> You can compare with https://github.com/rfay/homebrew-ddev/blob/master/Formula/ddev.rb

This is what I tried to explain before a few times.. `{{.Version}}` is a go template variable. It's not going to be replaced if you commit it as is to the homebrew Formalue repo. It's **goreleaser** that would replace that string for the release version of the moment. Thus, it would become v1.22.4 the next release and so on. So `{{.Version}}` would never get to the homebrew repo.

> Please take a look and see if my commit fixes it for you. I think we just needed to use make as-is, without any extras. Not even sure where that may have come from.

This may be better, I was worried that homebrew clones the repo singlebranch or something without all of the tags so make would not give the version you expect, but I will try it.

hanoii:
Same way that https://github.com/ddev/ddev/blob/94a62984637f04a43fbe7ed551926db7d821b087/.goreleaser.yml#L147C49-L147C60 works.

rfay:
Ah, so goreleaser is willing to replace it.

But regardless, we just need to go with `make` and it solves everything.

hanoii:
> You can test with brew uninstall -f ddev && brew install --HEAD --fetch-head rfay/ddev/ddev && ddev --version

Yes, this works!

rfay:
So @hanoii you're good with this as it is here? If so we'll get it into v1.22.4-alpha1 release this week and we can more easily test everything about it. Actually, as soon as it's pulled it will be easy to test. So I'll pull when you approve if you do.

hanoii:
@rfay I am good, if there are issues we can revise the previous approach.

rfay:
Thanks for all the work pursuing this! We have a couple of things affecting homebrew in upcoming release, both this and the new autocomplete setup, so we'll try to make sure they both get tested in the alpha. However, I think this one might be testable already.

hanoii:
@rfay FYI, I tested `brew install --HEAD ddev/ddev-edge/ddev` now that alpha was released and it worked as expected!

rfay:
Thanks for testing! I note that the HEAD install doesn't do the autocompletes, but it could, see https://github.com/rfay/homebrew-ddev-edge/commit/7f4933fac2d94eec697a8cc71751abd6d47a642d (which is incomplete, only does bash). `brew install --HEAD` takes just a little longer with that, but might be worth it. Interested in making that work?

hanoii:
it's just copying those lines on the if clause?

rfay:
Yeah, I think that would be it. (Except the other 2-3 completions need to be added)

PR #5440, opened 2023-10-20
fix: replace outdated nfs_mount_enabled comments with use_dns_when_possible

BramDriesen:
## The Issue
During a BOF you mentioned this was a bit outdated, so I updated it with what you said it needed to be 😄 

* Updated the docs!
* Updated other usages in tests which shouldn't have been relying on that. 

I don't think we have to do anything.

## Automated Testing Overview

I think the minor modifications to tests are OK.

### DIFF
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -281,7 +281,7 @@ Team members may choose to use `config.local.yaml` for local non-committed confi
 3. The list of environment variables in [`web_environment`](../configuration/config.md#web_environment) are “smart merged”: if you add the same environment variable with a different value, the value in the override file will replace the value from `config.yaml`.
 4. Hook specifications in the [`hooks`](../configuration/config.md#hooks) variable are merged.
 
-If you need to _override_ existing values, set [`override_config: true`](../configuration/config.md#override_config) in the `config.*.yaml` where the override behavior should take place. Since `config.*.yaml` files are normally _merged_ into the configuration, some things can’t be overridden normally. For example, if you have [`performance_mode: none`](../configuration/config.md#performance_mode) you can’t override it with a merge and you can’t erase existing hooks or all environment variables. However, with `override_config: true` in a particular `config.*.yaml` file,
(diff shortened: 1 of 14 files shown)


docs/content/users/extend/customization-extendibility.md
```diff
@@ -281,11 +281,11 @@ Team members may choose to use `config.local.yaml` for local non-committed confi
...
+If you need to _override_ existing values, set [`override_config: true`](../configuration/config.md#override_config) in the `config.*.yaml` where the override behavior should take place. Since `config.*.yaml` files are normally _merged_ into the configuration, some things can’t be overridden normally. For example, if you have [`performance_mode: none`](../configuration/config.md#performance_mode) you can’t override it with a merge and you can’t erase existing hooks or all environment variables. However, with `override_config: true` in a particular `config.*.yaml` file,
 
 ```yaml
 override_config: true
-nfs_mount_enabled: false
+performance_mode: none
```
rfay, line 288: The one problem with this example is that it isn't actually valid now that `performance_mode` is not a bool. It would be better to choose a bool for this exact example (and above). `use_dns_when_possible: false`

rfay (changes requested):
Thank you SO MUCH for capturing this and fixing it! I was going to try to remember it and YOU DID! Thanks.

rfay:
It turns out `nfs_mount_enabled` is also used inappropriately in a number of tests. If you want to explore that and add it here it will be welcome, otherwise we'll get it added on in the process.

docs/content/users/extend/customization-extendibility.md
BramDriesen: I have fixed those in the docs now

rfay:
Thanks so much for this. I added a commit that removes most of the other uses of nfs_mount_enabled in the code.

rfay (approved):
I tried to fix up the various test usages so they weren't doing an invalid thing, and I think it should be ok now. Thanks so much for remembering and fixing this.

stasadev (approved):

PR #5441, opened 2023-10-20
fix: don't pull omitted images, fixes #5292

### LINKED ISSUE #5292: `ddev start` downloads traefik image when router is excluded
On `ddev start` with `omit_containers: [ddev-router]` we shouldn't be downloading the traefik router.

On Codespaces, for example, `ddev start` will download the unused router image.

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/5292

Do not pull omitted images on `ddev start`.

1. `ddev poweroff`
2. `docker rmi $(docker images ddev/ddev-traefik-router -a -q)`
3. `ddev config global --omit-containers ddev-router`
4. `ddev start && ddev stop` should not pull `ddev-router`
5. `docker images ddev/ddev-traefik-router -a` should return an empty result

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -370,7 +370,7 @@ Containers that should not be loaded automatically for one or more projects.
 
 | Type | Default | Usage
 | -- | -- | --
-| :octicons-file-directory-16: project<br>:octicons-globe-16: global | `[]` | **For projects**, can include `db`, and `ddev-ssh-agent`.<br>**Globally**, can include `ddev-ssh-agent` only.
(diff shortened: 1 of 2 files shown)


stasadev:
The error in [Static Validation](https://github.com/ddev/ddev/actions/runs/6591099198/job/17909033329?pr=5441) comes from new rules in `golangci-lint` v1.55.0.

@rfay These errors are coming from other files (I didn't touch them) and I don't yet know how to resolve `dot-imports: should not use dot imports (revive)`.

rfay (approved):
I think this looks fine and seems to solve the problem. I tested it with just `ddev config global --omit-containers=ddev-router,ddev-ssh-agent` and it behaved as expected. Thanks for catching the doc problem.

PR #5442, opened 2023-10-20
ci: apply new rules from golangci-lint v1.55.0

stasadev:
## The Issue

```
$ make golangci-lint
golangci-lint: 
pkg/ddevapp/providerLocalfile_test.go:14:2: dot-imports: should not use dot imports (revive)
        . "github.com/ddev/ddev/pkg/ddevapp"
        ^
pkg/ddevapp/providerGit_test.go:12:2: dot-imports: should not use dot imports (revive)
        . "github.com/ddev/ddev/pkg/ddevapp"
        ^
pkg/ddevapp/providerPlatform_test.go:15:2: dot-imports: should not use dot imports (revive)
        . "github.com/ddev/ddev/pkg/ddevapp"
        ^
pkg/util/utils.go:132:9: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
        } else {
                output.UserOut.Println("Please type yes or no and then press enter:")
                return AskForConfirmation()
        }
make: *** [Makefile:265: golangci-lint] Error 1
```

Removes dot-imports and fixes indent-error-flow.

## Automated Testing Overview

Tests should pass as before.

### DIFF
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -12,7 +12,7 @@ import (
 	"testing"
 
 	"github.com/Masterminds/semver/v3"
-	. "github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/docker"
 	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/exec"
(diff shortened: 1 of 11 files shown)


rfay:
Thanks for chasing this down. I guess the change must be in revive. We'll get this in soon, I do want to understand what it's complaining about a little better so I learn something about golang.

rfay:
This has a decent explanation, https://www.maragu.dk/blog/using-dot-imports-in-views-in-go#:~:text=This%20practice%20is%20generally%20frowned,lead%20to%20a%20name%20conflict.

rfay:
Thanks for the fix!

Issue #5133, opened 2023-07-13
default_container_timeout not working


introfini:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

I have a rather large database that I snapshot, but when I try to restore it I get the following error:

```
Failed waiting for web/db containers to become ready: db container failed: log=, err=health check timed out after 2m0s: labels map[com.ddev.site-name:cufpt com.docker.compose.service:db] timed out without becoming healthy, status=
Container ddev-router  Running
Failed to restore snapshot cufpt_20230713113115 for project cufpt: failed to start project for RestoreSnapshot: container(s) failed to become healthy before their configured timeout or in 120 seconds. This may be just a problem with the healthcheck and not a functional problem. (health check timed out: labels map[com.ddev.site-name:cufpt] timed out without becoming healthy, status=, detail= ddev-cufpt-db:starting - more info with `docker inspect --format "{{json .State.Health }}" ddev-cufpt-db` )
```

I've override the default-container-timeout  variable as instructed:
`ddev config --default-container-timeout 250`

### Actual Behavior

I still get the error with the same 120 seconds:

`Failed to restore snapshot cufpt_20230713113115 for project cufpt: failed to start project for RestoreSnapshot: container(s) failed to become healthy before their configured timeout or in 120 seconds. This may be just a problem with the healthcheck and not a functional problem. (health check timed out: labels map[com.ddev.site-name:cufpt] timed out without becoming healthy, status=, detail= ddev-cufpt-db:starting - more info with `docker inspect --format "{{json .State.Health }}" ddev-cufpt-db` )`

rfay:
I think if you just wait for it it will work anyway, even though it says that it timed out.

rfay:
I'm not saying that the default-container-timeout function might not be broken, just that it likely won't affect your work.

introfini:
@rfay the ddev list reports "db: unhealthy". Perhaps the snapshot is simply too large; it is only 20GB in size. ;-)

Thanks!

rfay:
It will be unhealthy until it completes; does the log still show things happening? WHen you do the snapshot restore it shows you how to watch it.

Lots of people have done huge snapshots like this, but you need enough disk allocated to docker to do it. You didn't provide the requested `ddev debug test` so I don't know what environment you're working in...

introfini:
❯ docker inspect --format "{{json .State.Health }}" ddev-cufpt-db
[...]

PR #5444, opened 2023-10-21
fix: the web container startup time should be more generous, fixes #5343

### LINKED ISSUE #5343: ddev-webserver healthcheck timeout should begin only after mutagen sync is complete

```
```
</details>         

### Expected Behavior

The default healthcheck timeout for containers is 120s, but with ddev-webserver since v1.22.0, the actual startup of the php-fpm and nginx processes is only *after* the mutagen sync is complete. So the 120s should start only *after* start.sh begins.

### Actual Behavior

I think the mutagen sync is inside that 120s, so for large projects (or traditional Windows) this can run over the 120s.

I think this may be why we have such a hard time getting a successful run on traditional Windows (although we have trouble with WSL-docker-desktop as well, so it's also due to Docker Desktop for Windows being a bad thing.)

### Steps To Reproduce

Watch the tests.

An example of many, many failures is https://buildkite.com/ddev/ddev-windows-mutagen/builds/4230

### Anything else?

I think this can be fixed by changing the healthcheck values (perhaps startup?); and of course we have control of the healthcheck.sh. 

A very poor but very cheap approach would be to just increase the timeout on traditional Windows. 

Note that this may be affecting projects on macOS as well, where there are super-large codebases, or where large files are accidentally being synced.

rfay:
## The Issue

* #5343 
* I have been confused over and over again by the docker healthcheck configurations, and they were incorrect still. 
* #5133 
* #5154 (where this wasn't done right still)

* The `start_period` is the actual time that we want to wait for things to more-or-less be ready
* The `timeout` is actually how long we'll put up with the healthcheck taking (which is normally just over a minute, based on the web container healthcheck.sh)
* On Windows we'd just rather bump this up anyway, because the mutagen sync can take some time
* When doing a snapshot restore of a large snapshot, also bump it up quite a bit

- [x] Test on Windows with a largish project, maybe even a magento2 project
- [x] Test `ddev snapshot restore` with hugedb
- [x] Use `docker inspect` to see the resulting behavior

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -79,8 +79,9 @@ services:
       {{ end }}
       interval: 1s
       retries: 120
-      start_period: 120s
-      timeout: "{{ .DefaultContainerTimeout }}s"
+      start_period: "{{ .DefaultContainerTimeout }}s"
+      start_interval: 1s
+      timeout: 70s
   {{ end }} {{/* end if not .OmitDB */}}
 
   web:
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -240,8 +240,9 @@ services:
     healthcheck:
       interval: 1s
       retries: 120
-      start_period: 120s
-      timeout: "{{ .DefaultContainerTimeout }}s"
+      start_period: "{{ .DefaultContainerTimeout }}s"
+      start_interval: 1s
+      timeout: 70s
 
 networks:
   ddev_default:
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -130,6 +130,10 @@ func NewApp(appRoot string, includeOverrides bool) (*DdevApp, error) {
 
 	if app.DefaultContainerTimeout == "" {
 		app.DefaultContainerTimeout = nodeps.DefaultDefaultContainerTimeout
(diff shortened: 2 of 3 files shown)


rfay:
I found a lovely way to test this, no debugger needed. 

I added .ddev/web-entrypoint.d/wait.sh with contents:

```
#!/bin/bash
set +x
wait=20
echo "waiting $wait seconds at $(date)"
sleep $wait
echo "done waiting $wait seconds at $(date)"
```

This simulates a super slow mutagen or whatever. 

Then I can 
* `ddev restart` in one windows, 
* `ddev logs -f` in another (to see the startup process including the output of waiit.sh) and 
* `while true; do date && docker inspect --format "{{json .State.Health.Status }}" ddev-$PROJ-web | jq; sleep 1; done` to see the health status of the container as it comes up.
* Use `docker inspect <container>` to verify that the healthcheck stuff is set correctly, as well as examining .ddev/.ddev-docker-compose-base.yaml

So far it looks like everything works correctly.

rfay:
I'm pretty happy with this. I tested on macOS and traditional Windows with relatively large codebases (Magento2 on Windows, first time I've done that). It worked fine, the timeouts worked, and changing the `default_container_timeout` did the right thing.

On macOS I have a very large db snapshot, (2GB compressed) that is usually a fair demonstration. I also did one half that size successfully.

stasadev (approved):
I'm just going to approve it.

Testing it in a VM was a bad idea 😅.

I remember being able to test things like this on remote machines

PR #5448, opened 2023-10-23
feat: add `pre-command` and `post-command` hooks, fixes #5557 

### LINKED ISSUE #5557: Feature: A hook that runs every time a ddev command is executed
I would like a hook that would allow me to automate things like:
- `ddev restart` after a git pull if a certain criteria is met
- `notify something to the team members` if a certain criteria is met
- etc.

Although the issue should have come first, there's #5448 that adds an idea for this.

More context on https://discord.com/channels/664580571770388500/1144344589549973636/1166023601540051004

hanoii:
I am looking for ways to do more automations within ddev rather than on the host, and I wanted a hook that I can hook that it's run on every ddev command. I thought of different ways and came up with  pre-cmd and post-cmd command.

This PR adds the following hooks:
- `pre-command`
- `post-command`

That would run for any `ddev COMMAND`.

i.e.
```yaml
hooks:
  pre-command:
    - exec: echo pre-command exec
    - exec-host: echo pre-command exec-host
  post-command:
    - exec: echo post-command exec
    - exec-host: echo post-command exec-host
```

Any hook is supported, *however*, those that cannot be run when the site is not `started` will show a warning and the hook will be ignored (i.e. running an `exec` hook on `pre-command` when running `ddev start`).

More context on https://discord.com/channels/664580571770388500/1144344589549973636/1166023601540051004

Fixes #5557

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -51,7 +51,7 @@ Support: https://ddev.readthedocs.io/en/stable/users/support`,
 
 		app, err := ddevapp.GetActiveApp("")
 		if err == nil {
-			err = app.ProcessHooks("pre-command")
+			err = app.ProcessHooks("pre-command", command)
 			if err != nil {
 				util.Failed("failed to process pre-command hooks: %v", err)
 			}
@@ -109,9 +109,10 @@ Support: https://ddev.readthedocs.io/en/stable/users/support`,
 		}
 	},
 	PersistentPostRun: func(cmd *cobra.Command, args []string) {
+		command := os.Args[1]
 		app, err := ddevapp.GetActiveApp("")
 		if err == nil {
-			err = app.ProcessHooks("post-command")
+			err = app.ProcessHooks("post-command", command)
 			if err != nil {
 				util.Failed("failed to process post-command hooks: %v", err)
 			}
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -51,9 +51,9 @@ Support: https://ddev.readthedocs.io/en/stable/users/support`,
 
 		app, err := ddevapp.GetActiveApp("")
 		if err == nil {
-			err = app.ProcessHooks("pre-cmd")
+			err = app.ProcessHooks("pre-command")
 			if err != nil {
-				util.Failed("failed to process pre-cmd hooks: %v", err)
+				util.Failed("failed to process pre-command hooks: %v", err)
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


rfay:
Looks like you have some work to do yet. I guess my comment just from the title is I'd be more in favor of `pre-command` and `post-command`

hanoii:
Renamed and push a fix for the tests, let's see 🤞

hanoii:
Once I start working on some ddev feat is hard to stop :D

I continued working on this, it's looking working nicely I think, I've updated the PR description with more details.

I want to run an idea by you... I started adding hooks arguments, for now they do nothing (at first I needed it for some validation I was doing on `ProcessHooks()` but in the end I didn't need them. But I think this is also a nice addition. Basically I want to make available the COMMAND being run to the hook so that it can be used if needed. I thought of some non-invasive way of doing and I thought of adding something like:

`DDEV_HOOKS_PRE_COMMAND_ARG0`   environment variable to the os.Exec(). I've looked and I think this can be used for any hook (exec, exec-host, etc..). I would do in such a way that any arguments can be used in any hook. Another usage would be the db being imported on pre-import-db and post-import-db if you have multiple databases.

Would you want me to explore this here, or remove it and work on it on a separate PR.

rfay:
Rebased, I will get to it.

rfay:
Rebased.

rfay (changes requested):
You snuck args to tasks in here without mentioning it in the OP or updating the docs. It really should have been in a separate PR.

However, please update the PR OP with a description of this feature and how it's intended to work. 

And please add to the docs.

My simple manual testing of the base feature (pre-command and post-command hooks) worked fine. 

Do you have test coverage for the args feature?

stasadev (commented):
If these hooks produce output and you use the `--json-output` flag, it may break the PhpStorm plugin for DDEV:

```
de.php_perfect.intellij.ddev.cmd.CommandFailedException: Failed to parse output of command 'ddev describe --json-output': pre-command exec
pre-command exec-host
{"level":"info","msg":"..."}
post-command exec
post-command exec-host
```

rfay:
> If these hooks produce output and you use the --json-output flag, it may break the PhpStorm plugin for DDEV

That would be really, really annoying to anything that does a `ddev exec` or `ddev describe -j` running those hooks. Of course... if you install the hooks you should deal with it. But I agree that this could create a number of new complaints from people expecially using the PhpStorm extension. Unintended consequences.

rfay:
Could you please make a clearer statement about exactly why you want this? It would have been better to start with an issue for discussion. Please be clear about how this will benefit you and also how it can help the general DDEV community.

hanoii:
@rfay added #5557

rfay:
All your many contributions are so appreciated, but I think this one is just too tweaky, and nobody but you will know why or how it exists, so I'm not going to go with it. Thanks for all the effort and skill applied to it!

PR #5449, opened 2023-10-23
ci: improve TestExtraPackages to not use deb.sury.org

rfay:
## The Issue

TestExtraPackages fails intermittently, apparently due to problems with the deb.sury.org CDN.

Don't use packages (php packages) from deb.sury.org, use mainstream packages instead.

## Detail

There are many examples of this. For example, https://github.com/ddev/ddev/actions/runs/6619373200/job/17979758687?pr=5420, which has 
> [web 3/5] RUN apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests php8.1-tidy:
> 4.040 W: GPG error: https://packages.sury.org/php bullseye InRelease: Splitting up /var/lib/apt/lists/packages.sury.org_php_dists_bullseye_InRelease into data and signature failed
> 4.040 E: The repository 'https://packages.sury.org/php bullseye InRelease' is not signed.

There are also many complaints about this in the deb.sury.org queue, but none of them generally acknowledged.

### DIFF
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -1085,48 +1085,43 @@ func TestExtraPackages(t *testing.T) {
 	err = app.Start()
 	require.NoError(t, err)
 
-	// Test db container to make sure no ncdu in there at beginning
+	addedDBPackage := "sudo"
+	addedWebPackage := "tmux"
+
+	// Test db container to make sure no sudo in there at beginning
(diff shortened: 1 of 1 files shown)


rfay:
Sadly enough, this doesn't solve the problem, although it's a reasonable improvement. The problem happens at the `apt update` phase, not at `apt install`.

PR #5451, opened 2023-10-24
docs: add instructions for testing a PR by installing a downloaded binary in the $PATH

gitressa:
## The Issue

It is documented how to [test a PR in macOS with Homebrew](https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr), but not by installing a downloaded binary in the $PATH.

Documents how to test a PR by installing a downloaded binary in the $PATH, to be used for example in Linux and MacOS.

### DIFF
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -19,6 +19,8 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
 
 Download and unzip the appropriate binary and place it in your `$PATH`.
 
+### Homebrew with macOS
+
 If you’re using Homebrew, start by unlinking your current binary:
 
 ```
@@ -50,6 +52,33 @@ After you’re done, you can delete your downloaded binary and re-link the origi
 ```
 sudo rm /usr/local/bin/ddev
 brew link --force ddev
+```
+
+
+### Linux
+
+Start by removing your current binary, for Debian flavor systems use this:
+
+```
+sudo apt remove ddev
+```
+
+Next, unzip the binary you downloaded, make it executable, and move it to a folder in your path. Check with `echo $PATH`:
+
+```
+unzip ddev.zip
+chmod +x ddev && mv ddev ~/.local/bin/ddev
+```
+
+Verify the replacement worked by running `ddev version`. The output should be something like `DDEV version v1.22.3-39-gfbb878843`, instead of the regular `DDEV version v1.22.3`.
+
+You need to restart DDEV, to download the Docker images that it needs
+
+After you’re done, you can delete your downloaded binary and re-install the latest DDEV:
+
+```
+rm ~/.local/bin/ddev
+sudo apt install ddev
 ```
 
 ## Open in Gitpod
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -57,6 +57,7 @@ brew link --force ddev
 ### Installing a downloaded binary in the $PATH
 
 Normally, you can put any binary in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. In this example, `~/.local/bin` is used. Since not every distro has `~/.local/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
+
 ```
 mkdir -p ~/.local/bin
 export PATH="$HOME/.local/bin:$PATH"
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -54,7 +54,6 @@ sudo rm /usr/local/bin/ddev
 brew link --force ddev
 ```
 
-
 ### Linux
 
 Start by removing your current binary, for Debian flavor systems use this:
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -54,12 +54,12 @@ sudo rm /usr/local/bin/ddev
 brew link --force ddev
 ```
 
-### Linux
+### Installing a downloaded binary in the $PATH
 
-Start by removing your current binary, for Debian flavor systems use this:
-
+Normally, you can put any binary in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. In this example, `~/.local/bin` is used. Since not every distro has `~/.local/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
 ```
-sudo apt remove ddev
+mkdir -p ~/.local/bin
+export PATH="$HOME/.local/bin:$PATH"
 ```
 
 Next, unzip the binary you downloaded, make it executable, and move it to a folder in your path. Check with `echo $PATH`:
@@ -69,15 +69,14 @@ unzip ddev.zip
 chmod +x ddev && mv ddev ~/.local/bin/ddev
 ```
 
-Verify the replacement worked by running `ddev version`. The output should be something like `DDEV version v1.22.3-39-gfbb878843`, instead of the regular `DDEV version v1.22.3`.
+Now, close and reopen your terminal, and verify the replacement worked by running `ddev version`. The output should be something like `DDEV version v1.22.3-39-gfbb878843`, instead of the regular `DDEV version v1.22.3`.
 
-You need to restart DDEV, to download the Docker images that it needs
+You need to run `ddev poweroff` and `ddev start` to download the Docker images that it needs.
 
-After you’re done, you can delete your downloaded binary and re-install the latest DDEV:
+After you’re done testing, you can delete your downloaded binary, restart your terminal, and again use the standard DDEV:
 
 ```
 rm ~/.local/bin/ddev
-sudo apt install ddev
 ```
 
 ## Open in Gitpod
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -19,7 +19,7 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
 
 Download and unzip the appropriate binary and place it in your `$PATH`.
 
-### Homebrew with macOS
+### Homebrew with macOS or Linux
 
 If you’re using Homebrew, start by unlinking your current binary:
 
@@ -34,7 +34,7 @@ unzip ddev.zip
 chmod +x ddev && sudo mv ddev /usr/local/bin/ddev
 ```
 
-Verify the replacement worked by running `ddev -v`. The output should be something like `ddev version v1.19.1-42-g5334d3c1`, instead of the regular `ddev version v1.19.1`.
+Verify the replacement worked by running `ddev -v`. The output should be something like `ddev version v1.22.5-alpha1-70-g0852fc2df`, instead of the regular `ddev version v1.22.5`.
 
 !!!tip "macOS and Unsigned Binaries"
     macOS doesn’t like these downloaded binaries, so you’ll need to bypass the automatic quarantine to use them:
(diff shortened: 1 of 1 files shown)


docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
 brew link --force ddev
 ```
 
+### Linux
+
+Start by removing your current binary, for Debian flavor systems use this:
```
stasadev, line 55: I want to add that normally you don't need to remove ddev because when you put any binary in `~/.local/bin` it takes precedence.

docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
...
+
+```
+sudo apt remove ddev
+```
+
+Next, unzip the binary you downloaded, make it executable, and move it to a folder in your path. Check with `echo $PATH`:
```
stasadev, line 61: Not every distro has `~/.local/bin` in `$PATH`, so you can add an instruction how to do that (e.g. `export PATH="$HOME/.local/bin:$PATH"` to `~/.bashrc`).

docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
...
+
+```
+unzip ddev.zip
+chmod +x ddev && mv ddev ~/.local/bin/ddev
+```
+
+Verify the replacement worked by running `ddev version`. The output should be something like `DDEV version v1.22.3-39-gfbb878843`, instead of the regular `DDEV version v1.22.3`.
```
stasadev, line 68: Also we can check the correct binary with `which ddev` and usually you need `source ~/.bashrc` to see the changes, but I usually just restart the terminal.

docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
...
+
+You need to restart DDEV, to download the Docker images that it needs
```
stasadev, line 70: Dot at the end is missing.

stasadev (commented):
Thank you!

gitressa:
Thanks for the comments @stasadev. Before addressing them, I am having an odd situation, where my DDEV 1.22.3 _without_  a DDEV in `.local/bin` gives me the `ddev dbeaver` feature ... Do you have any idea how to clean Docker images, and start fresh? I ran `ddev poweroff && docker system prune -a --volumes` but maybe I need to clear some more Docker or DDEV caches as well?

stasadev:
>I am having an odd situation, where my DDEV 1.22.3 without a DDEV in `.local/bin` gives me the `ddev dbeaver` feature

It is because the file is here `~/.ddev/commands/host/dbeaver`.
You can simply remove it.

If you want to have a complete fresh DDEV, you can backup `~/.ddev/global_config.yaml` (you need only `project_info` key section from there), remove `~/.ddev` folder, and then run some ddev command (like `ddev version`), it will recreate `~/.ddev` folder, and you can add back `project_info` to `~/.ddev/global_config.yaml`.

>Do you have any idea how to clean Docker images.

I usually do `ddev poweroff && ddev delete images --all` when I want to clean DDEV related stuff.

and

```
docker rm -f $(docker ps -aq)
docker rmi -f $(docker images -q)
docker builder prune
```

when I want to clean everything else.

Or as you did: `docker system prune -a --volumes` - this command cleans everything. I usually run this if I've messed something up and want to start over.

docs/content/developers/building-contributing.md
rfay: My approach is to use a symlink in /usr/local/bin/ddev, which points to either the system-installed ddev or the locally built one.

rfay (changes requested):
This really isn't about linux at all. It's about understanding how $PATH works, and putting the ddev binary in your PATH ahead of where the system-installed binary is.

I'd love to have you rethink this as "installing a downloaded binary in the $PATH" instead of "Linux instructions".

docs/content/developers/building-contributing.md
gitressa: Great suggestion, I tried to update the text with that tip.

docs/content/developers/building-contributing.md
gitressa: Perfect, added.

docs/content/developers/building-contributing.md
gitressa: Done

docs/content/developers/building-contributing.md
gitressa: We could add the symlink option as well? The reason I prefer `~/.local/bin` is that symlinks can be harder to understand.
Which commands would you use for the symlink solution?

rfay:
Another minor nuance is that this isn't just about testing PRs, it's the same with testing HEAD version by downloading the artifacts.

rfay:
And... It is in fact hard for lots of people. So this is a very valuable thing to cover.

docs/content/developers/building-contributing.md
rfay: I use symlinking. I switch back and forth with a script I call relink_ddev.sh. Of course I have to do this all the time. My strategy is to have a symlink /usr/local/bin/ddev.link that points to the built version and then have /usr/local/bin/ddev be a symlink that points to it. But delete /usr/local/bin/ddev when using system-installed version. I use this script on Linux/WSL2 as well as macOS.

```
#!/bin/bash

set -x
set -o errexit

if command -v brew; then
  brew unlink ddev || true
fi

SUDO=""
if  [ ! -w /usr/local/bin ] || [ ! -f /usr/local/bin/ddev ] || [ ! -w /usr/local/bin/ddev ]; then
  SUDO="sudo "
fi
${SUDO} ln -sf /usr/local/bin/ddev.link /usr/local/bin/ddev
hash -r
```

tyler36:
Using `echo $PATH` and `which ddev` are valuable commands for debugging.

docs/content/developers/building-contributing.md
gitressa: Cool method @rfay, thanks for sharing. We could include it as an alternative method, if you suggest the changes (?), I can accept the suggestion and commit it to the patch.

gitressa:
Awesome explanation about Docker and the different DDEV parts @stasadev. I didn't get to comment about it before now, but thanks a lot for that, I really appreciate it.

Maybe the steps to test a PR (or the structure of DDEV itself and tools and add-ons?) needs to get a short description, since there are a few moving parts ...?

Maybe the help text could be expanded, to cover testing a PR for different situations and what to be aware of, such as testing a PR which changes code only inside DDEV (which I suppose would follow the latest binary) or testing a PR, which changes code in other places, such as the `~/.ddev` folder?

@tyler36: Great suggestion, I have updated the text.

docs/content/developers/building-contributing.md
gitressa: Yes, you're probably right. At least, it's now documented in this issue, if anyone is interested. Resolving this thread, since it seems to be so.

docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
 brew link --force ddev
 ```
 
+### Installing a downloaded binary in the $PATH
```
mattstein, line 53: We’ve almost got consistent title casing for headings on this page, so might as well continue here. And wrap `$PATH` in backticks since it’s a literal reference to a system variable. (Even though backtick styling in headings is kind of ugly.)

```suggestion
### Installing a Downloaded Binary in the `$PATH`
```

docs/content/developers/building-contributing.md
```diff
@@ -48,6 +50,32 @@ sudo rm /usr/local/bin/ddev
...
+
+Normally, you can put any binary in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. In this example, `~/.local/bin` is used. `echo $PATH` and `which ddev` are valuable commands for debugging. Since not every distro has `~/.local/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
```
mattstein, line 55: Being picky, but I’d try to swap “is used” for more active language. Also helps break up a long stretch of comma clauses:

```suggestion
Normally, you can put any binary in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. This example uses `~/.local/bin`. `echo $PATH` and `which ddev` are valuable commands for debugging. Since not every distro has `~/.local/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
```

mattstein (changes requested):
Left some nitpicky suggestions but otherwise looks good to me!

gitressa:
Perfect @mattstein, both were great suggestions, thanks!

gitressa:
In a follow up issue, perhaps that page can be re-visited about the use of the word "Binary"?

It doesn't mean anything to me.

So perhaps the page can start with stating something like this?

> After each PR, an individually-compiled DDEV binary ("DDEV binary") is created containing the new code.

And after that, use "DDEV binary" on the page or whatever we agree on is a fitting word for it, and not just "binary"?

gitressa:
In a follow up issue, perhaps the page can be re-visited about the use of the word "Binary"?

It doesn't mean anything to me.

So perhaps the page can start with stating something like this?

> After each PR, an individually-compiled DDEV binary ("DDEV binary") is created containing the new code.

And after that, use "DDEV binary" on the page or whatever we agree on is a fitting word for it, and not just "binary"?

Also, this could be tied in with @rfay's comment about DDEV’s latest-committed HEAD version:

> Another minor nuance is that this isn't just about testing PRs, it's the same with testing HEAD version by downloading the artifacts.

I am not certain how to phrase it though ...

rfay:
I'm interested in what wording to use, if we can do better. "Artifacts" is a pretty technical term which is not useful to people who aren't normally in a build environment. "Binary" is one of many words that imply a compiled artifact. In the PHP world there's aren't binaries (except `php` itself of course) because things are scripted. Looking forward to your thoughts about better words to use. Thoughts, @mattstein ?

mattstein:
Maybe [executable](https://go.dev/doc/tutorial/compile-install) instead?

rfay:
I could live with "executable". It's kind of a traditional Windows-ish word, but maybe people would understand it better.

mattstein:
If that’s the predominant term in the GoLang docs, I don’t think Clippy crossover warrants great concern. 🤷‍♂️

docs/content/developers/building-contributing.md
```diff
@@ -15,6 +15,8 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
 
 Download and unzip the appropriate binary and place it in your `$PATH`.
 
+### Homebrew with macOS
```
rfay, line 18:
```suggestion
### Homebrew with macOS or Linux
```

rfay (changes requested):
Thanks for this contribution and initiative.

Overall, this is adding too much training for too little value. I think we have to (mostly) assume that people understand what $PATH is. We can give them a little guidance about the place to put it. 

IMO ~/bin is a better place to put things than `.local/bin`, it's easier to work with and supported by most platforms' .bash_profile. 

Overall the problem though is how much handholding we can do here. 

Could we do a gist or something with the full handholding and keep this more concise? 

Surely there's a good tutorial somewhere we can point people to, or a gist or something?

Or maybe we need a PATH tutorial on ddev.com?

gitressa:
> Or maybe we need a PATH tutorial on ddev.com?

That's a great idea! We can move the text there and link to it.

Also, `~/bin` isn't in my path on Ubuntu ...

rfay:
On Ubuntu and most systems, this is in /etc/skel/.profile, which is used by default by sh and bash:

```
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/bin" ] ; then
    PATH="$HOME/bin:$PATH"
fi

# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

That's most likely how your .local/bin gets into the PATH, and if ~/bin exists, it will be added. 

You have to start a new bash session to see it.

gitressa:
I didn't know about `/etc/skel/.profile` @rfay, thanks for explaining. And you're right, it's there, and a copy of it in `~/.profile`, and if I create `~/bin` it does get picked up as part of the `$PATH`.

I prefer to keep my home folder as clean as possible of any program folders, which is why I use the hidden folder `~/.local/bin` for this, to house Drush, Composer, and so on.

gitressa:
Perfect, thanks @mattstein, @stasadev, @tyler36, and @rfay.

Actually, I have since come to the conclusion that using `~/bin` for DDEV-testing is the superior solution, since after you're done, you can just delete the folder, to clean up everything. And then ... might this be slightly better, since you then don't have to take care of the ZIP-file?

> ```
> cd ddev ~/bin
> ```
> Next, download the ZIP-file, unzip it and make it executable. Check with `echo $PATH`:
> ```
> unzip ddev.zip && chmod +x ddev
> ```

Or maybe just add the the `rm` command like so `unzip ddev.zip && rm ddev.zip` in the current solution?

Also, both v1.22.3 and v1.22.5 are used, but that's a minor detail.

rfay:
I always keep a ~/bin, and it's a fairly normal thing to do since the beginning of the Unix age. 

Let's let this be for now. Maybe you'll want to return to it in a few months :)

gitressa:
Hah, yes, I may realize that as well :)

gitressa:
I have warmed to this method @rfay :-)

First run:
`mkdir ~/bin`

... close the terminal, and then run:
`source ~/.profile
`

PR #5452, opened 2023-10-24
fix: add project config validation when using args, fixes #5445

### LINKED ISSUE #5445: No validation when running `ddev start project-name`

```
```
</details>         

### Expected Behavior

When you run `ddev start project-name`, and you have some invalid config, DDEV should be stopped early.

### Actual Behavior

When you have some invalid config, DDEV runs some checks on `ddev start`, but there are no checks when you run `ddev start project-name`, and project continues to start.

### Steps To Reproduce

1. `mkdir start-validation && cd start-validation && ddev config --auto`
2. open `.ddev/config.yaml` and make some invalid change, I did `php_version: "9.0"`
3. `ddev start`
4. see an error as it should be:
```
$ ddev start
Failed to get project(s): unsupported PHP version: 9.0, DDEV only supports the following versions: [5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2 8.3]
```
5. `ddev start start-validation`
6. see this:

```
$ ddev start start-validation
Starting start-validation... 
 Container ddev-ssh-agent  Created 
ssh-agent container is running: If you want to add authentication to the ssh-agent container, run 'ddev auth ssh' to enable your keys. 
Building project images... 
Project images built in 3s. 
 Network ddev-start-validation_default  Created 
 Container ddev-start-validation-web  Created 
 Container ddev-start-validation-db  Created 
 Container ddev-start-validation-db  Started 
 Container ddev-start-validation-web  Started 
Waiting for web/db containers to become ready: [web db] 
Failed waiting for web/db containers to become ready: web container failed: log=, err=container exited, please use 'ddev logs -s web' to find out why it failed 
Something is wrong with Docker or Colima and /mnt/ddev_config is not mounted from the project .ddev folder. This can cause all kinds of problems. 
Waiting for additional project containers to become ready... 
Failed to start start-validation: container(s) failed to become healthy before their configured timeout or in 120 seconds. This might be a problem with the healthcheck and not a functional problem. (container 'web' exited, please use 'ddev logs -s web' to find out why it failed)
```

### Anything else?

I believe this can be caused when you upgrade an old DDEV to a new one with old configs and start a project with `ddev start project-name` and see that it fails without any explanation

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/5445

Runs validation check for every projects that you pass to any DDEV command that uses args, so fixes more than just `ddev start`.

See the issue.

### DIFF
--- a/cmd/ddev/cmd/utils.go
+++ b/cmd/ddev/cmd/utils.go
@@ -46,7 +46,7 @@ func getRequestedProjects(names []string, all bool) ([]*ddevapp.DdevApp, error) 
 		if requestedProjectsMap[name], exists = allProjectMap[name]; !exists {
 			p := globalconfig.GetProject(name)
 			if p != nil && p.AppRoot != "" {
-				requestedProjectsMap[name] = &ddevapp.DdevApp{Name: name}
+				requestedProjectsMap[name] = &ddevapp.DdevApp{Name: name, AppRoot: p.AppRoot}
 			} else {
 				return nil, fmt.Errorf("could not find requested project '%s', you may need to use \"ddev start\" to add it to the project catalog", name)
 			}
--- a/cmd/ddev/cmd/utils.go
+++ b/cmd/ddev/cmd/utils.go
@@ -55,6 +55,11 @@ func getRequestedProjects(names []string, all bool) ([]*ddevapp.DdevApp, error) 
 
 	// Convert map back to slice
 	for _, project := range requestedProjectsMap {
+		err = project.ValidateConfig()
+		if err != nil {
+			return nil, err
+		}
+
 		requestedProjects = append(requestedProjects, project)
 	}
 
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -461,8 +461,11 @@ func (app *DdevApp) ValidateConfig() error {
 		return err
 	}
 
-	// Stop here if no config found
+	// Skip any validation below this check if there is nothing to validate
 	if err := CheckForMissingProjectFiles(app); err != nil {
+		// Do not return an error here because not all DDEV commands should be stopped by this check
+		// It matters when you start a project, but not when you stop or delete it
+		// This check is reused elsewhere where appropriate
 		return nil
 	}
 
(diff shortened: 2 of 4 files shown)


rfay (commented):
A little more work to do on tests, changed it back to draft.

rfay:
Rebased

pkg/ddevapp/config.go
```diff
@@ -462,6 +461,11 @@ func (app *DdevApp) ValidateConfig() error {
 		return err
 	}
 
+	// Stop here if no config found
+	if err := CheckForMissingProjectFiles(app); err != nil {
+		return nil
```
rfay, line 466: Shouldn't we be returning the error here?

rfay (changes requested):
I'm confused about returning nil on error there, but it does function correctly in manual testing.

The one important thing to do is to improve the message. Currently we see something like `Failed to get project(s): unsupported PHP version: xx.1, DDEV only supports the following versions: [5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2 8.3]` but we need to identify the project that failed, so `Failed to load project 'projectname': ...

pkg/ddevapp/config.go
stasadev: Added more comments on this check.

cmd/ddev/cmd/utils.go
```diff
@@ -46,7 +46,7 @@ func getRequestedProjects(names []string, all bool) ([]*ddevapp.DdevApp, error)
 		if requestedProjectsMap[name], exists = allProjectMap[name]; !exists {
 			p := globalconfig.GetProject(name)
 			if p != nil && p.AppRoot != "" {
-				requestedProjectsMap[name] = &ddevapp.DdevApp{Name: name}
+				requestedProjectsMap[name] = &ddevapp.DdevApp{Name: name, AppRoot: p.AppRoot}
```
stasadev, line 49: When you run `ddev start <project>` for a project with missing files, the error does not include the path:

```
Failed to start test-project: ddev can no longer find your project files at . If you would like...
```

And now it does:

```
Failed to start test-project: ddev can no longer find your project files at /home/stas/code/ddev/test-project. If you would like...
```

stasadev (commented):
I added the project name to the error message for each validation check.

rfay (approved):
This looks great to me. Lots of improvements to user communication. 💯 

One weird thing I encountered was on `ddev start -a`  with a broken config on one of the projects (ddev.com), `t:ype: xxx`.  It happily tried to start ddev.com (and didn't report errors) even though the config was invalid. I imagine that `ddev start -a` doesn't validate config. We don't have to fix that here.

PR #5453, opened 2023-10-24
refactor: update to Node.js v20 LTS as current default

peter-neumann-dev:
## The Issue

Today, Node.js v20 has its active LTS start. As explained in the documentation: The default for the config option `nodejs_version` is the current LTS version.

This PR raises the default used Node.js version from 18 to 20

### DIFF
--- a/.gitpod/images/Dockerfile
+++ b/.gitpod/images/Dockerfile
@@ -4,7 +4,7 @@ SHELL ["/bin/bash", "-c"]
 USER root
 
 RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
-RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main" > /etc/apt/sources.list.d/nodesource.list
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" > /etc/apt/sources.list.d/nodesource.list
 
 RUN install -m 0755 -d /etc/apt/keyrings
 RUN curl -fsSL https://pkg.ddev.com/apt/gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/ddev.gpg > /dev/null
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -54,10 +54,10 @@ ARG BUILDPLATFORM
 
 SHELL ["/bin/bash", "-c"]
 
-RUN wget -q -O /tmp/nginx_signing.key http://nginx.org/keys/nginx_signing.key && \
-    apt-key add /tmp/nginx_signing.key && \
-    rm /tmp/nginx_signing.key && \
-    echo "deb http://nginx.org/packages/debian/ $(lsb_release -sc) nginx" > /etc/apt/sources.list.d/nginx.list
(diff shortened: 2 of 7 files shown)


rfay:
I think 18 remains LTS for some time yet, doesn't it? https://endoflife.date/nodejs

rfay:
From https://nodejs.dev/en/about/releases/ :

image

rfay:
And I should have said at the first: Thanks for tracking this and contributing!

peter-neumann-dev:
Hi @rfay, 

thank you! I thought about Issue vs PR, but my mind was: Come on you can do this as well … At least the preparation 😄 

I checked the release schedule of Node.js here: https://github.com/nodejs/release#release-schedule
CleanShot 2023-10-24 at 16 36 52@2x

Also since today 20 is listed as LTS on https://nodejs.org/en
CleanShot 2023-10-24 at 16 36 10@2x

Maybe the other resources will get updated later today or the next days 🙂

peter-neumann-dev:
They are on it, it's called Iron this time:
CleanShot 2023-10-24 at 16 54 07@2x

rfay:
This won't be able to go in until DDEV v1.23.0, which is likely in the Spring of 2024. Do you want to pursue it now anyway? I do very much appreciate you keeping us up-to-date and paying attention!

peter-neumann-dev:
Sure, no problem, I was changing it in another project where I set up DDEV and saw it was 18 by default.

Should we add this to the documentation [here](https://ddev.readthedocs.io/en/stable/users/configuration/config/#nodejs_version), because 18 as the default is not the "current LTS version" or "active LTS version" anymore, instead "maintenance LTS version"? Maybe in a separate PR or something 🙂

The release table is now updated on GitHub: https://github.com/nodejs/Release#release-schedule
Here is the current LTS release: https://github.com/nodejs/node/releases/tag/v20.9.0

rfay:
The "latest" docs are at https://ddev.readthedocs.io/en/latest/users/configuration/config/#nodejs_version, so that's a *little* better. Not sure exactly how to do interim docs about this. It's not *that* important to most people. 

One of the biggest concerns I have is that we put a `nodejs_version` in the config.yaml by default, and I wish we didn't, because they get old so fast.

peter-neumann-dev:
That's an argument 👍🏻 Would the version get updated in exiting `config.yaml` files?

peter-neumann-dev:
Ok, I understand. I assumed the changes I did to raise the default version to 20 will only be noticeable when creating a new project without any existing `node_version`, why I thought it is not a problem, because for new projects that's what you want 🙂

rfay:
You might be right... We'll think about it. Maybe it's not even intrusive.

rfay (changes requested):
Since some of the code changed here is in the ddev-php-base Dockerfile, for example, this requires a push of that and of ddev-webserver. I'll change this to "draft" until we get that done.

rfay:
This is pushed with the image now. I'm not 100% convinced about it, but now it's testable. Moving to "ready to review"

stasadev (approved):
Looks good!

```
$ ddev exec node -v
v20.9.0
```

As for when to make the switch, I'm good with any decision.

By the way, a few days ago AWS Lambda released a managed runtime for Node.js 20 https://aws.amazon.com/blogs/compute/introducing-the-amazon-linux-2023-runtime-for-aws-lambda/

rfay:
Rebased and pushed updated image. For testing, update using `docker pull ddev/ddev-webserver:20231024_peter-neumann-dev_nodejs_v20_lts`

rfay:
This will need to wait for DDEV v1.23.

rfay:
Working on this, rebased, trying to run the php-base image build, getting an unexpected failure, easy to recreate locally.

PR #5454, opened 2023-10-24
fix: add PATH_INFO CGI parameter to all nginx configs, fixes #3854, fixes #4001

### LINKED ISSUE #3854: Add PATH_INFO CGI variable to nginx configuration

longwave:
## The Issue

When using SimpleSAMLphp inside a ddev project with nginx, it crashes because it cannot find the $_SERVER['PATH_INFO'] variable, which is expected to be set. This is documented at https://simplesamlphp.org/docs/stable/simplesamlphp-install#configuring-nginx

`PATH_INFO` is added to all nginx config files. This is only needed to handle special URLs that have a `.php` script file in the _middle_ of the URL, e.g. `/some/path/index.php/extra/arguments` where `/some/path/index.php` is executed and the `extra/arguments` are picked up by the script.

Start a new project with SimpleSAMLphp:
```
mkdir samltest
cd samltest
composer require simplesamlphp/simplesamlphp:^1

mkdir config
cp vendor/simplesamlphp/simplesamlphp/config-templates/config.php config/config.php
ln -s vendor/simplesamlphp/simplesamlphp/www simplesaml

ddev config --project-type drupal10 --web-environment SIMPLESAMLPHP_CONFIG_DIR=/var/www/html/config
ddev start
```

Then visit https://samltest.ddev.site/simplesaml/module.php/core/frontpage_welcome.php 

Without the changes in this PR, you get an error page because SimpleSAMLphp cannot parse the path.
With the changes in this PR, you get the SimpleSAMLphp welcome page.

## Related Issue Link(s)

* #3854
* #4001

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/nginx/sites-enabled/nginx-site-default.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/nginx/sites-enabled/nginx-site-default.conf
@@ -39,6 +39,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/containers/ddev-webserver/tests/ddev-webserver/testdata/nginx-site.conf
+++ b/containers/ddev-webserver/tests/ddev-webserver/testdata/nginx-site.conf
@@ -47,6 +47,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
@@ -48,6 +48,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-craftcms.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-craftcms.conf
@@ -53,6 +53,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal10.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal10.conf
@@ -47,6 +47,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
@@ -58,6 +58,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
         fastcgi_index index.php;
         include fastcgi_params;
         fastcgi_intercept_errors off;
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal7.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal7.conf
@@ -51,6 +51,7 @@ server {
         fastcgi_buffer_size 32k;
         fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
         fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        fastcgi_param PATH_INFO $fastcgi_path_info;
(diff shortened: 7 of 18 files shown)


rfay:
Thanks so much for the contribution!

stasadev:
Updated the webserver image and rebased.

But I don't know how test it.

https://www.php.net/manual/en/reserved.variables.server.php

> Contains any client-provided pathname information trailing the actual script filename but preceding the query string, if available. For instance, if the current script was accessed via the URI http://www.example.com/php/path_info.php/some/stuff?foo=bar, then $_SERVER['PATH_INFO'] would contain /some/stuff.

When I try to check the value it always returns an empty string.

rfay:
I confirm that the PATH_INFO got pushed successfully into the image. I had forgotten that in all normal DDEV contexts we mount or copy the per-CMS one from user space in there, so almost all of the things that could be done would have been correct already without the image push. 

@longwave please add manual testing instructions, and should we consider a test?

rfay:
I don't think PATH_INFO is supported for php_fpm at all, that's why we don't see it. PATH_INFO is supported for CGI, like obsolete apache CGI.

rfay:
The link provided above for this is 404, I think we want https://simplesamlphp.org/docs/stable/simplesamlphp-install.html

Current stable of simplesamlphp is 1.21.

rfay:
Key link: https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/

This says it's needed in /etc/nginx/fastcgi_params as well. 

I see lots of failures to get this going on the internet.

Could you please debug this using just .ddev/nginx_full/nginx-site.conf @longwave ? Please see if you can get it going properly; you can do that with your own project easily. 

I'm not sure the push to the default nginx config is even necessary here.

When you have it going by just adjusting your project nginx config, a PR update here is welcome. 

I note that you've been working on this for a long time, and have contributed a PR before... but this PR does not seem to solve the problem, and you should be able to do anything you need with .ddev/nginx_full/nginx-site.conf, so work that out and when it's working and we have a good test technique we'll go forward.

rfay:
@longwave thanks so much for your consistent work on this. But if you don't respond by Nov 17 I'll close this. We can certainly open it again, and we can continue the conversation. But to date, nobody knows how to get it actually working. Please make sure that you manually test, which you can do with this PR's artifacts and pushed image, but you could also do with a simple .ddev/web-build/Dockerfile that updates the nginx config. https://github.com/ddev/ddev/pull/5454#issuecomment-1777532059

rfay:
Sorry to do this, but closing due to no response. Happy to reopen. Sorry to have you keep returning to this. When you get back to it I'm happy to help you test it locally and figure out what is required. Happy to continue the conversation here or elsewhere.

longwave:
Steps to reproduce:

1. Start a new project with SimpleSAMLphp:
```
mkdir samltest
cd samltest
composer require simplesamlphp/simplesamlphp

mkdir config
cp vendor/simplesamlphp/simplesamlphp/config/config.php.dist config/config.php
ln -s vendor/simplesamlphp/simplesamlphp/public simplesaml

ddev config --project-type drupal10 --web-environment SIMPLESAMLPHP_CONFIG_DIR=/var/www/html/config
ddev start
```

2. Visit https://samltest.ddev.site/simplesaml/module.php/core/welcome and see a 404.

3. Edit .ddev/nginx_full/nginx-site.conf:
* Remove `#ddev-generated`
* Change the PHP `location` block regex to `location ~ '\.php(/|$)|^/update.php' {`
* Add `fastcgi_param PATH_INFO $fastcgi_path_info;` in the PHP location block.

4. https://samltest.ddev.site/simplesaml/module.php/core/welcome will now show th SimpleSAMLphp welcome page.

Both the location regex and `PATH_INFO` need to be changed for this to work - which I guess will need changing in all the places in this PR.

It's also further complicated because it doesn't seem to work on the `php` project type, I think because of `try_files`? But I am using this inside a Drupal install so the `drupal` project types work for me.

rfay:
Thanks, reopened. But your repro steps require editing nginx on local? 

How about the ways to test this PR?

Please edit the OP to show both repro steps and testing steps, thanks. And please try it out and see if this PR solves it.

https://github.com/ddev/ddev/pull/5454#issuecomment-1777532059

longwave:
Figured out how to reproduce with this PR alone with SimpleSAMLphp 1.x; 2.x has slightly different paths and might need more work. Will update the issue summary.

longwave:
The binary built earlier in this PR solves the issue, as described in steps to reproduce in the issue summary.

rfay:
I spent quite a bit of time with this this morning, 

* I didn't understand that PATH_INFO is about the path *after* the script. For example, in `https://samltest.ddev.site/pathinfotest.php/one/two`, the PATH_INFO is one/two, not `pathinfotest`. https://stackoverflow.com/questions/56640592/path-info-in-server-always-empty-nginx-fpm-7-3-ubuntu-18-04 got me out of that mode.
*  I can use your repro to make simplesaml work. However, PATH_INFO still isn't actually populated, it remains empty.


I am using this script in your setup:
```php
<?php
if (isset($_SERVER['PATH_INFO'])) {
    echo "PATH_INFO='" . $_SERVER['PATH_INFO'] . "'\n";
    echo "SCRIPT_NAME='" . $_SERVER['SCRIPT_NAME'] . "'\n";
    echo "SCRIPT_FILENAME='" . $_SERVER['SCRIPT_FILENAME'] . "'\n";
} else {
    echo "PATH_INFO is not set";
}
```

It does show PATH_INFO as empty, which is apparently what fixes simplesaml.

However, there is apparently no way with the current config to hit this:

`curl -I https://samltest.ddev.site/pathinfotest.php/one/two` always gets a 404, so you may have another problem ahead of you. I don't think our config actually supports that arrangement.

If you can get this to where we can demonstrate PATH_INFO actually *working* I'm all for it, of course if it doesn't break anything else.

Also, do you want it to work for apache as well?

rfay:
Looking forward to your responses @longwave - I think this should be fine, but want to make sure it's completely fine with you, maybe you can give a snippet showing the actual use of PATH_INFO.

rfay:
Rebased, hoping for response from @longwave

rfay:
I think the problem here is that PATH_INFO is actually for Apache + mod_php, and is not relevant for nginx + php-fpm. See https://stackoverflow.com/a/2261971/215713

DDEV has not supported Apache + mod_php (was `apache-cgi`) for 4-5 years, because it's so slow and nobody used it.

rfay:
Unless I hear from you with discussion about this, I think I'll close this about Feb 15, 2024. I'm pretty sure this is a misunderstanding of the use of PATH_INFO and a bug in SimpleSaml. And, of course, you have a workaround any time you need it.

longwave:
I think we should go ahead and close this for now. The SimpleSAMLphp fix did work for v1.x, but as of last week all sites I work on have upgraded to SimpleSAMLphp 2.x (it uses Symfony 6, and therefore only works with Drupal 10), and the same fix doesn't work there; I have to add an entire new stanza to my nginx.conf.

I am still convinced `PATH_INFO` is the right way to solve this as SimpleSAMLphp relies on it being set correctly more than most PHP applications, due to the way it has the PHP script in the middle of the URL and also requires the end of the URL to be passed in correctly, but if I manage to get a unified solution working with SimpleSAMLphp 2.x then I will post that as a new PR.

rfay:
Thanks so much for all your work on this. Are you sure that PATH_INFO is even a thing for Apache2 with fpm? I'm sorry for yet another discouraging effort on this front.

rfay:
If you're still interested in this after all these years,
* https://github.com/ddev/ddev/pull/8844

PR #5461, opened 2023-10-26
fix: remove unnecessary use of fmt.Sprintf

testwill:


rfay (approved):
Thanks for noticing and fixing!

PR #5467, opened 2023-10-27
docs: add info about disk space checking on wsl2

rfay:
## The Issue

Disk space problems need to be mentioned in Troubleshooting section.

### DIFF
--- a/docs/content/users/usage/troubleshooting.md
+++ b/docs/content/users/usage/troubleshooting.md
@@ -8,7 +8,7 @@ Things might go wrong! In addition to this page, consider checking [Stack Overfl
 * Temporarily disable firewalls, VPNs, tunnels, network proxies, and virus checkers while you’re troubleshooting.
 * Temporarily disable any proxies you’ve established in Docker’s settings.
 * Use [`ddev debug dockercheck`](commands.md#debug-dockercheck) and [`ddev debug test`](commands.md#debug-test) to help sort out Docker problems.
-* Make sure you do not have disk space problems on the host. This can be especially tricky on WSL2, where you need to check both the host disk space and WSL2 disk space as well.
+* Make sure you do not have disk space problems on your computer. This can be especially tricky on WSL2, where you need to check both the host disk space and WSL2 disk space as well.
 * On macOS, check to make sure Docker Desktop or Colima are not out of disk space. In *Settings* (or *Preferences*) → *Resources* → *Disk image size* there should be ample space left; try not to let usage exceed 80% because the reported number can be unreliable. If it says zero used, something is wrong.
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/troubleshooting.md
```diff
@@ -8,6 +8,7 @@ Things might go wrong! In addition to this page, consider checking [Stack Overfl
 * Temporarily disable firewalls, VPNs, tunnels, network proxies, and virus checkers while you’re troubleshooting.
 * Temporarily disable any proxies you’ve established in Docker’s settings.
 * Use [`ddev debug dockercheck`](commands.md#debug-dockercheck) and [`ddev debug test`](commands.md#debug-test) to help sort out Docker problems.
+* Make sure you do not have disk space problems on the host. This can be especially tricky on WSL2, where you need to check both the host disk space and WSL2 disk space as well.
```
mattstein, line 11: By “the host,” do we mean “your workstation?” I’m not sure if “your host workstation” is an accurate phrase here, but it might be good to reinforce that language and be more specific.

mattstein (approved):

docs/content/users/usage/troubleshooting.md
rfay: I'd like to improve this terminology in general. Probably "on your computer" is best.

rfay, line 11:
```suggestion
* Make sure you do not have disk space problems on your computer. This can be especially tricky on WSL2, where you need to check both the host disk space and WSL2 disk space as well.
```

mattstein: In that case maybe “on your computer” and then “both on your computer’s disk space and the space available to WSL2.”

rpkoller: if you change the term in the first sentence from the host to your computer then i would change the host disk space also to your computer's disk space in the second sentence. otherwise it might be even more confusing?

docs/content/users/usage/troubleshooting.md
```diff
@@ -8,6 +8,7 @@ Things might go wrong! In addition to this page, consider checking [Stack Overfl
 * Temporarily disable firewalls, VPNs, tunnels, network proxies, and virus checkers while you’re troubleshooting.
 * Temporarily disable any proxies you’ve established in Docker’s settings.
 * Use [`ddev debug dockercheck`](commands.md#debug-dockercheck) and [`ddev debug test`](commands.md#debug-test) to help sort out Docker problems.
+* Make sure you do not have disk space problems on your computer. This can be especially tricky on WSL2, where you need to check both the host disk space and WSL2 disk space as well.
```
rfay, line 11:
```suggestion
* Make sure you do not have disk space problems on your computer. This can be especially tricky on WSL2, where you need to check both the main Windows disk space and WSL2 disk space as well.
```

PR #5473, opened 2023-10-28
ci: gather debugging information about possible TestExtraPackages failure

rfay:
## The Issue

TestExtraPackages often fails due to 
* https://github.com/oerdnj/deb.sury.org/issues/2046

Experiment with gathering more information about the problem. This is explored in
* https://askubuntu.com/a/990019/18314
* https://askubuntu.com/a/841968/18314

The suggestions have to do with disk space (this happens at docker build time) and with removing /var/lib/apt/lists


rfay:
Bizarrely, I can't make it fail with these additions. This one will be a candidate for reversion if we can discover the real cause.

PR #5475, opened 2023-10-30
Update dbeaver command to allow usage with dbeaver flatpak installs o…

sschoeppe0:
…n user level

In most cases you don't install flatpaks on system level but on user level unfortually its not possible to use the "$HOME" variable inside "## HostBinaryExists:" there for IMHO it have to be removed from the command

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
@@ -5,7 +5,6 @@
 ## Usage: dbeaver
 ## Example: "ddev dbeaver [db] [user]"
 ## OSTypes: darwin,linux
-## HostBinaryExists: /Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
 
 database="${1:-db}"
 user="${2:-root}"
@@ -25,6 +24,7 @@ case $OSTYPE in
     BINARIES=(
       /usr/bin/dbeaver{,-ce,-le,-ue,-ee}
       /var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity
+      $HOME/.local/share/flatpak/app/io.dbeaver.DBeaverCommunity/current/active/export/bin/io.dbeaver.DBeaverCommunity
       /snap/bin/dbeaver-ce
     )
     for binary in "${BINARIES[@]}"; do


stasadev:
Hi, @sschoeppe0.

Unfortunately we can't remove `HostBinaryExists` because if we don't have that line, `ddev dbeaver` will be listed for all DDEV users even if they have not installed DBeaver.

For now, you can make your own changes to the script and remove the `#ddev-generated` line from the `~/.ddev/commands/host/dbeaver`. That way it will work the way you want it to.

However, I think it is possible add an environment variable expansion to DDEV using a regular expression here:

https://github.com/ddev/ddev/blob/34d7fd76bb9068ac1d8d304c1f928624ff0ded22/cmd/ddev/cmd/commands.go#L162-L174

A new feature request or PR is welcome.

sschoeppe0:
Hi @stasadev
thanks for your quick reply, i just have reviewed the code you mentioned and have one more question. Do you think such an regex would maybe usefull in the` fileutil.FileExists` function in general?

rfay (changes requested):
Thanks so much for your work on this, and we're looking forward to success. And good creative thinking also about using a regex.

sschoeppe0:
Hi @stasadev  @rfay  thanks for your feedback i have updated the fileExists function for Linux/Mac (sorry but i'm not familiar with windows so i skiped it) and readded the HostBinaryExists part

pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
```diff
@@ -5,7 +5,7 @@
 ## Usage: dbeaver
 ## Example: "ddev dbeaver [db] [user]"
 ## OSTypes: darwin,linux
-## HostBinaryExists: /Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
+## HostBinaryExists: /Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce,$HOME/.local/share/flatpak/app/io.dbeaver.DBeaverCommunity/current/active/export/bin/io.dbeaver.DBeaverCommunity
```
rfay, line 8: Thanks, were you able to test this manually? I wouldn't have necessarily expected that $HOME would be replaced in this context.

pkg/fileutil/file_exists.go
```diff
@@ -3,11 +3,23 @@
...
+	"regexp"
+)
+
+func replaceBashVariables(input string) string {
+	r := regexp.MustCompile(`\$(\w+)|\$\{(\w+)\}`)
+	return r.ReplaceAllStringFunc(input, func(s string) string {
+		varName := r.ReplaceAllString(s, "$1$2")
+		return os.Getenv(varName)
+	})
+}
```
rfay, line 17: These changes must be suggested in a separate PR dedicated to its purpose please.

rfay (changes requested):
PRs need to have a single purpose, so when DDEV golang code is going to be changed, let's suggest that and make it happen. But not in this PR.

pkg/fileutil/file_exists.go
sschoeppe0: ok i will close this PR and open 2 new PR's

rfay:
There was no need to close this one, it was on target. Just needed to revert the golang changes. But it's all good, thanks. If you want you can reopen this one after your other suggestion gets in.

PR #5476, opened 2023-10-30
feat: make it possible to use SHELL variables in FileExists function

sschoeppe0:

### DIFF
--- a/pkg/fileutil/file_exists.go
+++ b/pkg/fileutil/file_exists.go
@@ -3,11 +3,24 @@
 
 package fileutil
 
-import "os"
+import (
+	"os"
+	"regexp"
+)
+
+// Replace SHELL variables with real value
+func replaceShellVariables(input string) string {
+	r := regexp.MustCompile(`\$(\w+)|\$\{(\w+)\}`)
+	return r.ReplaceAllStringFunc(input, func(s string) string {
+		varName := r.ReplaceAllString(s, "$1$2")
+		return os.Getenv(varName)
+	})
+}
 
 // FileExists checks a file's existence
 func FileExists(name string) bool {
-	if _, err := os.Stat(name); err != nil {
+	expandedName := replaceShellVariables(name)
+	if _, err := os.Stat(expandedName); err != nil {
 		if os.IsNotExist(err) {
 			return false
 		}


pkg/fileutil/file_exists.go
```diff
@@ -3,11 +3,24 @@
...
+)
+
+// Replace SHELL variables with real value
+func replaceShellVariables(input string) string {
+	r := regexp.MustCompile(`\$(\w+)|\$\{(\w+)\}`)
+	return r.ReplaceAllStringFunc(input, func(s string) string {
+		varName := r.ReplaceAllString(s, "$1$2")
+		return os.Getenv(varName)
+	})
+}
```
rfay, line 18: This logic already exists in a more general form in https://github.com/ddev/ddev/blob/16870030d464e7bb64a569685b65871ce5669469/pkg/ddevapp/provider.go#L429-L441

but perhaps it needs to be generalized to be available elsewhere. I thought we had something that already did this, as we used to preprocess the nginx config file this way, but that isn't done the same any more.

Note that the Linux command `envsubst` exists for this purpose as well, but we can't count on that being available everywhere. 

One suggestion is to process the command file, rather than making a specific exception in the FileExists(). Adding this to FileExists() has vast consequences that are unpredictable.

Please consider processing the command file with variable replacements instead.

rfay (changes requested):
Thanks for this contribution and for the willingness to take on new territory!

rfay:
And, of course, when creating a PR please include all the details. This one will require a test as well.

pkg/fileutil/file_exists.go
rfay: I understand that it was suggested. Sorry, I think it's too intrusive, at least the way I look at it right now.

stasadev: @sschoeppe0 Sorry, I should point out that @rfay is the lead developer so the final decision is not up to me.

rfay:
Thanks for your contribution but I'm closing this for now. I don't think this approach is safe enough, and don't accept PRs that don't have adequate information in the headings requested for PRs. I do think we should do environment and go template replacement on files like this, and that will be a nice future improvement.

PR #5481, opened 2023-10-30
build: always use the latest go and golangci-lint for gitpod [skip ci]

stasadev:
## The Issue

It is easy to forget to update some packages in [Dockerfile](https://github.com/ddev/ddev/blob/34d7fd76bb9068ac1d8d304c1f928624ff0ded22/.gitpod/images/Dockerfile#L17-L19) for Gitpod after release.

Try to use the latest version all the time.

Found the solution for `go` [here](https://stackoverflow.com/a/77085404).
And it was super easy for `golangci-lint` - I just removed the hard-coded version.

```
$ docker buildx build -t ddev/ddev-gitpod-base --platform=linux/amd64 .gitpod/images

$ docker run --rm -it ddev/ddev-gitpod-base go version
go version go1.21.3 linux/amd64

$ docker run --rm -it ddev/ddev-gitpod-base golangci-lint version
golangci-lint has version 1.55.1 built with go1.21.3 from 9b20d49d on 2023-10-25T10:03:43Z
```

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: ddev/ddev-gitpod-base:20231026
+image: ddev/ddev-gitpod-base:20231030
 tasks:
   - name: build-run
     init: |
--- a/.gitpod/images/Dockerfile
+++ b/.gitpod/images/Dockerfile
@@ -14,9 +14,9 @@ RUN apt-get update >/dev/null && sudo apt-get install -y aspell autojump ddev fi
 
 RUN pip3 install mkdocs pyspelling pymdown-extensions
 RUN npm install -g markdownlint-cli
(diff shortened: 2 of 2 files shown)


rfay (approved):
Looks as good as it gets! Thanks!

rfay:
Followup will be to push a new version of the image, probably with a new tag, and update here and in ddev-gitpod-launcher. I think it would be OK to push now and add the new tag to this PR.

stasadev:
Pushed a new tag and updated it here and in ddev-gitpod-launcher.

rfay:
```
gitpod /workspace/ddev (master) $ go version && ddev --version
go version go1.21.3 linux/amd64
ddev version v1.22.4
```

Thanks so much!

PR #5484, opened 2023-10-30
docs: specify ddev tap in upgrading instructions for macOS with brew

bencroker:
## The Issue

The brew tap at https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/ is not specific enough and results in an error when attempting to upgrade.

Screenshot 2023-10-30 at 23 15 24

This PR specifies the tap `ddev/ddev/ddev`, bringing it in line with the installation instructions at https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/


rfay (approved):
Thanks, this is a good idea.

But you should also `brew untap drud/ddev` on your homebrew.

https://twitter.com/randyfay/status/1719118192994631840

PR #5488, opened 2023-11-01
docs: update custom commands example to show some of the possible annotations

rfay:
## The Issue

The custom commands example didn't include example annotations, which would be a nice feature for people to understand and use by default.


docs/content/users/extend/custom-commands.md
```diff
@@ -17,14 +17,16 @@ The [`ddev mysql`](../usage/commands.md#mysql) runs the `mysql` client inside th
 
 ## Host Commands
 
-To provide host commands, place a Bash script in `.ddev/commands/host`. For example, a PhpStorm launcher to make the `ddev phpstorm` command might go in `.ddev/commands/host/phpstorm` with these contents:
+To provide host commands, place a Bash script in `.ddev/commands/host`. For example, a PhpStorm launcher to make the `ddev phpstorm` command might go in `.ddev/commands/host/phpstorm` with these contents. The OSTypes and HostBinaryExists annotations are optional, but are useful to prevent the command from showing up if it's not useful to the user.
```
mattstein, line 20: Those annotation keys seem like literal references, so you might want to wrap them in backticks:

```suggestion
To provide host commands, place a Bash script in `.ddev/commands/host`. For example, a PhpStorm launcher to make the `ddev phpstorm` command might go in `.ddev/commands/host/phpstorm` with these contents. The `OSTypes` and `HostBinaryExists` annotations are optional, but are useful to prevent the command from showing up if it's not useful to the user.
```

mattstein (changes requested):

mattstein (approved):

Issue #4972, opened 2023-06-13
Drupal 10.1 nginx configuration change


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Per https://www.drupal.org/node/2888767#nginx-php-fpm there's a potential change required for Drupal 10.1 nginx config. 

### Describe your solution

Change the config.

It would be nice if we could make it completely compatible with 10.0 as well.

catch56:
The ddev config looks fine for Drupal 10.1

```
# js and css always loaded
    location ~* \.(js|css)$ {
        try_files $uri @rewrite;
        expires -1;
        log_not_found off;
    }
    ```
    
    This passes through all js and css - so it catches core aggregates, including with https://www.drupal.org/project/drupal/issues/3027639
    
 I also tested locally with ddev and 10.1.0, and aggregation is working, files created on disk etc.
 
 If there was a stricter directory pattern, it might have needed a change, but the blanket rule for me at least seems fine.

rfay:
Based on @catch56 report I'm going to close this for now, thanks!

PR #5494, opened 2023-11-02
fix: handle Drupal 10.1 asset aggregation URLs.

catch56:
## The Issue
Drupal 10.1 asset aggregation uses a similar mechanism to image derivatives/image styles. These directories don't fall through to PHP in the default ddev nginx config which leads to 404s

Allow more directories

Install Drupal 10.1 and enable js/css aggregation

## Automated Testing Overview

Just a default config change, would need to install Drupal to really test that it fixes what it says it does.

## Release/Deployment Notes

### DIFF
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal10.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal10.conf
@@ -34,8 +34,8 @@ server {
         rewrite ^ /index.php;
     }
 
-    # Handle image styles for Drupal 7+
-    location ~ ^/sites/.*/files/styles/ {
+    # Passes image style and asset generation to PHP.
+    location ~ ^/sites/.*/files/(css|js|styles)/ {
         try_files $uri @rewrite;
     }
 


rfay:
Thanks for the contribution!

rfay:
You say 
> Install Drupal 10.1 and enable js/css aggregation

But how do you tell whether it's working?

We do in fact install Drupal10 in tests, in TestDdevFullSiteSetup. I don't know if there's a way to validate this properly or not?

catch56:
I realised this already works... was a different local issue causing it not to. There is a wildcard rule further down this file that matches any .js or .css file anyway. However, it might be worth making the change regardless since it matches the Drupal.org ngnix recommendation and updates the docs a bit anyway?

Also realised we already discussed this a bit in https://github.com/ddev/ddev/issues/4972

rfay:
Since it already works, and has already been discussed and solved, I'm going to go ahead and close this for now, but happy to continue the conversation, and I so much appreciate you looking out for things like this.

Happy to reopen if you think that's best.

Issue #5171, opened 2023-07-20
Improve usability of of django4 and python projects


rfay:
### The problem

DDEV introduced Python/Django support in v1.22.0.

The good things:
* DDEV has direct support for [Django4](https://ddev.readthedocs.io/en/latest/users/quickstart/#django-4-experimental) (but not yet Django5) and also for more generic [Python/Flask](https://ddev.readthedocs.io/en/latest/users/quickstart/#pythonflask-experimental) projects. More discussion in [Experimental Configurations](https://ddev.readthedocs.io/en/latest/users/configuration/experimental/#django4-and-python-project-types)
* It has decent automated tests, including a Django4 configuration, a simple Flask configuration. Each of these passes our automated testing, and problems are immediately noticed.

The bad things:
* These features do not yet have a community. As far as I know, less than five people have experimented with these capabilities. DDEV is fundamentally dependent on its community, which is key to its success with PHP projects. Users notice what they need, and they let us know. DDEV would have long drifted away except for its community involvement, and we do not have anything like that for Python/Django.
* We chose Gunicorn for the webserver, and Gunicorn, unlike php-fpm, absolutely refuses to run if the supported app's settings are not correct, and setting those requires some expertise and the ability to read logs to sort it out. Every user that I'm familiar with has stumbled at this point. DDEV tries to prompt about the basic issues, and the docs mention this, but users just stumble here.

Platform.sh has stated its clear plans for DDEV to support Python projects, but we are definitely not there yet, and I'd love to have more feedback about how we could be more successful, especially getting people started at the beginning. I'm willing to use something other than gunicorn if that can solve the problems. But the community problem will require far more significant efforts.

### Possible solutions

* Write better tutorials with specific examples
* Do screencasts
* Improve visibility of what's going on with gunicorn, or use another webserver

chellman:
Hey @rfay sorry for the radio silence, was traveling.  I appreciate the changes that made it into 1.22, and I'll try to help however I can with Django improvements.

It seems like PyCharm is improving support for remote interpreters; the jankiness of working on a container-based project has made me slow to push on this before ddev support was available.  So if JetBrains can get support for everything working well, I'll be trying to switch to using DDEV on my main Django project at work.

My work project is the first one I've ever worked on for a meaningful length of time, which means I'm not overly familiar with gunicorn, but I am familiar with bashing on a project until I can get it running. 🤣 

[...]

PR #5499, opened 2023-11-03
feat: avoid having to remove files in `ddev composer create` #5493

### LINKED ISSUE #5493: Prevent execution of `ddev config` in the wrong directory

The other day I was creating a new DDEV instance using the `ddev composer create` command.

I wasn't concentrating (I know, bad move) and accidentally didn't enter the folder I had just created for this purpose. These should have been the steps:

```
cd ~/dev
mkdir drupal10
cd drupal10
ddev config --auto
```

Except, I got distracted and skipped the `cd drupal10` step ...

I quickly hit "Enter" through the prompts from the `ddev config` and `ddev composer create` commands, which looked all right. But since I wasn't in `~/dev/drupal10`, but instead in the `~/dev` folder, where all my projects are (around 15) they were almost all deleted, until the process got stopped by a write-only folder.

Yes I know, I should have double checked the prompts.

### Describe your solution

It would be awesome if it was possible to designate folders, where DDEV will gladly run project commands such as `ddev composer create` (i.e. those where you can get a `could not find a project` message) and reject them elsewhere. `ddev list` and similar commands which doesn't alter anything should still be allowed from any folder.

We only need to check for `ddev config`, since without a `.ddev` sub-folder present, no file altering commands will run, but simply return a "Failed to get project(s): could not find a project in /home/username.".

For example, I only ever run DDEV commands in project sub-folders within a development folder called `~/dev` (`/home/username/dev`), such as `/home/username/dev/drupal10`.

So a rule could be, to only allow DDEV to execute file altering commands in sub-folders of `~dev` with the pattern `/home/username/dev/*`.

Meaning: Only allow running `ddev config` from for example `/home/username/dev/drupal10`

Everywhere else, for example `/home/username/dev` or `/home/username` you get a message like this:

> You can only run ddev config to start new DDEV projects from `/home/username/dev/projectname`

Defining the rule in plain speech: "Only allow running `ddev config` from these folders". Since there can be more than one, it should probably be an array, like this:

`allow_ddev_config: ["/home/rfay/workspace/*", "/tmp/*"]`

How it could look in `.ddev/global_config.yaml`:

```
# Only allow running ddev config from these folders
# Separate multiple values with comma, quoted
# allow_ddev_config: ["/home/username/workspace/*", "/tmp/*"]
```

The `ddev config` command already checks if it is run from the home folder, which it doesn't allow and returns the message "could not create new config: ddev config is not useful in your home directory (/home/username)".

https://github.com/ddev/ddev/blob/master/pkg/ddevapp/config.go#L67-L70

Maybe the `allow_ddev_config` setting can be added similarly?

### Describe alternatives

_No response_

### Additional context

_No response_

hanoii:
Basically what I am doing instead of removing everything is:

- If composer root is different from the app root, it validates that it's empty
- If composer root is the same as the app root, it validates that:
  - there's nothing in the directory other than `.ddev`, `.git`, `.tarballs` and, if configured, the docroot.
  - It also adds a method so that each project type can whitelist extra paths (settings, upload dirs, etc..) that would be accepted as valid paths to be present for running composer create

If any of the validations above fails, the command does not run. This forces you to start on a clean directory without doing any removal, if you have to remove things, you have to do it yourself.

Helps with #5493


hanoii:
GitHub is all over the place - https://www.githubstatus.com/

hanoii:
rebased to restart tests

rfay:
I'd be a bit interested in you revisiting the very promising
* https://github.com/ddev/ddev/pull/5058

before going forward with this one. If you wouldn't mind taking a look at that and thinking about it, I'm sure we can figure out a way for you to either push into it or just start a new PR by forking it.

cmd/ddev/cmd/composer-create.go
```diff
@@ -62,54 +61,52 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
...
-		objs, err := fileutil.ListFilesInDir(composerRoot)
-		if err != nil {
-			util.Failed("Failed to create project: %v", err)
-		}
-
-		for _, o := range objs {
-			// Preserve .ddev, .git, .tarballs
-			if o == ".ddev" || o == ".git" || o == ".tarballs" {
-				continue
+		// If composer root is not the app root, make sure it's empty
```
rfay, line 72: Comment may not say what you want it to? You're saying "fail if it's not empty" right?

pkg/fileutil/files.go
```diff
@@ -135,6 +135,26 @@ func IsDirectory(path string) bool {
...
+func IsDirectoryEmpty(path string) bool {
+	if IsDirectory(path) {
+		f, err := os.Open(path)
+		if err != nil {
+			return false
+		}
+		defer f.Close()
+
+		// read in ONLY one file
+		_, err = f.Readdir(1)
```
rfay, line 148: Pretty sure `.` and `..` are always there. Does this work right?

rpkoller:
that might be also the functionality @rfay asked for in https://github.com/ddev/ddev/pull/5058#issuecomment-1685461783 for https://github.com/ddev/ddev/pull/5058 ?

cmd/ddev/cmd/composer-create.go
hanoii: isn't fail if it's not empty the same as make sure it's empty?

pkg/fileutil/files.go
hanoii: I took this code out of https://socketloop.com/tutorials/golang-determine-if-directory-is-empty-with-os-file-readdir-function - it did work on all my tests.

rfay: OK, as long as you've debugged through it. Oh, please add a test. Most of the functions in there have tests, maybe all of them.

rfay: Also, please include a link in the code comment when you've copied code from somewhere, thanks!

hanoii:
@rfay I looked at #5058 , from what I gathered, the other one is much bigger and not so related to this one. This one just changes the approach to not remove anything and performs useful validations to whether it should run of not. If the two should be combined into one I'd say to fork the otherone on top of this one, but maybe this one is quicker to get accepted? If you like the approach the changes should be pretty harmless.

pkg/fileutil/files.go
hanoii: Code added

rfay:
Thanks! This is going to need tests as well, because it's super important.

And another thing to think about (another PR) is controlling `ddev config` as discussed in #5493

hanoii:
@rfay 
> Thanks! This is going to need tests as well, because it's super important.

I would assume that all of the tests already there for this functionality should be as useful, if you can thing of anything else to test.

One test could be make sure it fails when it has to.

I will comment on #5493 as I don't see that much of a need there if this gets in.

hanoii:
@rfay can you let me know why on https://github.com/ddev/ddev/blob/64c9e469427785debfdda7db2e8caf1b8632433c/cmd/ddev/cmd/composer_test.go#L97-L114 you run two composer create on the same app. Up to know that would work because the next one would wipe everything out which obivously makes the test fail with this new approach. I am thinking how two modfiy, one option is to remove the second run, or if needed, run them from within new "ddev instances".. just trying to understand exactly what use case you are testing there.

rfay:
It's just testing different things, but using the same project to do it, which saves time. You could change it to deliberately remove (have the test remove the contents) or you could change it to create new projects.

hanoii:
@rfay this obviously adds a step to some quickstarts, see https://ddev--5499.org.readthedocs.build/en/5499/users/quickstart/#drupal for example. As not every project adds files to the webroot on start this should only be on those that do. I can over other composer create quickstart and check that out.

rfay:
> this obviously adds a step to some quickstarts

Please say why :) It's not so obvious.

rfay:
It's not possible to remove some things if they're bind-mounted, like sites/default/files with mutagen. Not sure what's up there.

hanoii:
> It's not possible to remove some things if they're bind-mounted, like sites/default/files with mutagen. Not sure where you're going with this.

I've thought about this, the way I undrestood your current code, you are removing them and then recreated on the first restart.

I figured this is not necessary, as another restart happens after `composer create` finishes. Correct me if I am wrong, and I'd love for you to test this (which is ready to test except from tests).

hanoii:
see if https://ddev--5499.org.readthedocs.build/en/5499/users/quickstart/#drupal is clear enough now.

hanoii:
Because you are removing things on the host, i don't think that affects the container, the bind mount will be there on restart.

rfay:
Not very happy about that :(

hanoii:
> Not very happy about that :(

About what? And can you elaborate what would you recommend doing instead?

rfay:
`rm -fr ../my-drupal10-site/web/*` is 
1. Not going to work in many cases
2. Open to failures by people not familiar with the command line
3. Using `*` with rm should never be done by people casually

Why is the code not doing this programatically?

hanoii:
> Why is the code not doing this programatically?

Well, the main goal of this PR was to offload arbitrary removal of ddev to the user, so that ddev doesn't have any chance of removing unnecessary things. As it is down, doing a composer create on a wrong folder could have a very bad data loss, that's why I consider this safer.

But safer doesn't mean user friendly and you know more about your users than me. I used `*` on purpose as that means anything inside that dir, which is less prone to issues.

But I thought of something else that was more work so I didn't go that route, which is removing ONLY what ddev start adds, which probably means adding another hook to your projectMatrix... I can explore that. I prefer not go the route of having ddev remove EVERYTHING inside the docroot, as again, there might be things there that ddev should not remove.

Again, let's say you accidentally init in a parent dir that happen to have a web docroot and you wiped that out, not great.

rfay:
I guess we're finding out why `composer create-project` refuses to operate on anything but a completely empty directory. 

Another strategy may be to gleefully remove composer_root/vendor, but balk at other subdirectories.

rfay:
For the quickstarts, since people are clearly in a new directory, we should be able to sort this out.

rfay:
Some other possible safety valves:
*  refuse to operate if .ddev directories were found in subdirectories; perhaps there are other red flags we could use.
* Don't try to remove dotfiles

hanoii:
I think this is mostly the reason: https://github.com/composer/composer/issues/7455#issuecomment-403965801, but it makes sense..

thinking of what to remove is not easy because it is also based on the composer project, drupal doesn't just use vendor as its output from create project, but also on web..

I have an idea to explore...

hanoii:
@rfay let's see if you are happy with this, if you can review properly the idea I can continue and tell me what you think.

Basically composer create will allow a certain amount of files to be present, some constant for every project (like current `.ddev`, `.git` and `.tarballs`, and some others that will be exposed by each project type. Anything other than that composer create will fail to run.

This should be easy to test as we just have to run composer create on every project type after a clean init.

There's no need for the user to remove anything from a quickstart.

rfay:
Thanks, I'll be back Tuesday.

hanoii:
Improved the logic a bit and the error message.

cmd/ddev/cmd/composer-create_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	assert := asrt.New(t)
+
+	origDir, err := os.Getwd()
+	assert.NoError(err)
+
+	types := ddevapp.GetValidAppTypes()
+	if os.Getenv("GOTEST_SHORT") != "" {
+		types = []string{nodeps.AppTypePHP, nodeps.AppTypeDrupal10}
+	}
```
hanoii, line 33: @rfay I continued to work on this and added some tests, while I still would appreciate you trying this out and vouching on the approach (🤞) I have an immediate question here..  I added this by looking and some other tests but the tests suite here on GH actions run it with GOTEST_SHORT set to 12 which not exactly sure what that mean. I think it's sensible for this test to run for every single project, is there a proper way of doing that.. shall I remove this if clause and always make it full, or maybe do a `!= 'false'` so it can be set locally... thanks!

hanoii:
Added tests for all util functions

pkg/fileutil/files.go
hanoii: Test added!

cmd/ddev/cmd/composer-create.go
hanoii: Changed it to what you said.

hanoii:
Skipping django4 on test because of #5171

cmd/ddev/cmd/composer-create_test.go
rfay: I guess you mean you want it to run for every project type. 

This test is extremely resource-intensive, so that's why it's only run with the one. 

GOTEST_SHORT=12 means that by default TestDdevFullSiteSetup and another test or two will only use stanza 12 in the test setup, 12 is drupal10, https://github.com/ddev/ddev/blob/64c9e469427785debfdda7db2e8caf1b8632433c/pkg/ddevapp/ddevapp_test.go#L225-L238

If GOTEST_SHORT is empty, it allows running all. In this case it's just doing php and drupal10 when GOTEST_SHORT is set to *anything*.

Does that answer your question?

cmd/ddev/cmd/composer_test.go
```diff
@@ -90,11 +85,66 @@ func TestComposerCmd(t *testing.T) {
 		assert.NoError(err)
 
 		assert.FileExists(filepath.Join(tmpDir, composerRoot, "Psr/Log/LogLevel.php"))
+	}
+}
+
+func TestComposerCmd2(t *testing.T) {
```
rfay, line 91: Try a descriptive name, it can be whatever you want.

pkg/nodeps/utils.go
```diff
@@ -135,3 +136,19 @@ func GrepStringInBuffer(buffer string, needle string) []string {
...
 	return matches
 }
+
+// PathExplode returns an array of all possible paths separated by slashes out
+// of a single one.
+// i.e. path/to/file will return {"path", "path/to", "path/to/file"}
+func PathExplode(path string) []string {
+	var paths []string
+	var partial string
+	for _, p := range strings.Split(path, "/") {
```
rfay, line 146: You can't be using strings.Split() on filepaths. Instead use the go `filepath` package to do what you need. Remember... Windows. And weird embedded thigns of other types.

And shouldn't this be in the fileutil package, not utils?

pkg/nodeps/utils_test.go
```diff
@@ -10,5 +10,27 @@ func TestRandomString(t *testing.T) {
...
+
+// TestPathExplode tests PathExplode
+func TestPathExplode(t *testing.T) {
+	assert := asrt.New(t)
+
+	testSources := []string{
+		"sites/default/files",
+		"/sites/default/files",
+		"./sites/default/files",
+	}
```
rfay, line 24: Ah, but where are the Windows examples :)

pkg/ddevapp/drupal.go
```diff
@@ -593,3 +593,31 @@ func drupalImportFilesAction(app *DdevApp, uploadDir, importPath, extPath string
 
 	return nil
 }
+
+// getDrupalComposerCreateAllowedPaths returns paths that are allowed to be present when running composer create
+func getDrupalComposerCreateAllowedPaths(app *DdevApp) ([]string, error) {
```
rfay, line 598: I have to confess that this looks pretty complex compared to what we expect it to need to be.

rfay (changes requested):
Thanks for all the dedicated work on this! 

The *ComposerCreateAllowed* makes me feel better, but I'm concerned about the things other than backdrop and drupal and WP. Maybe they all use .env files in the root, not sure.  I don't think that's true of Magento/Magento2 though.

What happens with people who are using the `php` project type to do a Drupal project, which is quite common... ?

hanoii:
> The _ComposerCreateAllowed_ makes me feel better, but I'm concerned about the things other than backdrop and drupal and WP. Maybe they all use .env files in the root, not sure. I don't think that's true of Magento/Magento2 though.

Each project type will whitelist whatever file they allow to be there for a composer create to work.

> What happens with people who are using the `php` project type to do a Drupal project, which is quite common... ?

If they are starting with an empty project, this would just work.

pkg/ddevapp/drupal.go
hanoii: is there something to do here?

pkg/nodeps/utils.go
hanoii: The package I haven't thought about, if you think it has to go there, I'll move it. About slashes, you seem to use slashes on other places and I added the documentation that this is expecting slashes. It's more a string function that a path function. I would like to understand this better before spending more time unless it's really required.

pkg/nodeps/utils_test.go
hanoii: Same as above, I expect this to work only on slashes, even in windows.. unless I understood things wrongly, but in the drupal.go you reference paths with slashes too.

cmd/ddev/cmd/composer_test.go
hanoii: Done

pkg/ddevapp/drupal.go
rfay: Take a look and see if it has to be this much code to accomplish a list of paths.

pkg/nodeps/utils_test.go
rfay: Every place we use file paths, we use the filepath stuff, it's important to be consistent about that. You can't just start randomly creating new ways to handle paths.  Did you want filepath.Split()? https://pkg.go.dev/path/filepath

hanoii: It's not doing the same as filepath.Split, it's a very specific thing I did (which might not be the bets) but it's connected to the way I am `filepath.Wallk`ing the dir on composer create command but then running all paths through a function that does the same  think as https://github.com/ddev/ddev/blob/0225d6f483adc606e9da1ffaf650c06d56f2882d/pkg/ddevapp/utils.go#L405. I made a function more general that does exactly the same thing.

I don't know how to change this withought being able to test it and it's not a simple change. If you have the time today, I'd like to walk you through what I implemented and why I think this will work the same on Windows as ToSlashes is being run and I only care about relative this. `PathRelativeSlashesExplode` could be a better name.

hanoii: This also confused me:

https://github.com/ddev/ddev/blob/0225d6f483adc606e9da1ffaf650c06d56f2882d/pkg/ddevapp/drupal.go#L501

it looks like that logics is supposed to work for windows and elsewhere, but it's only using slashes.

rfay: That snippet is known to always run inside the container I believe, where it's always linux. You can probably tell I've had some bad experiences with bizarre results of not using filepath where it's required.

cmd/ddev/cmd/composer-create_test.go
hanoii: Not really, but I understood of what GOTEST_SHORT does with an integer. My question is, can I allow my test of composer create to run for every project type?

pkg/ddevapp/drupal.go
hanoii: I simplified it slightly moving some of what it does upwards.

gitressa:
Thanks @hanoii, [I tested it yesterday](https://github.com/ddev/ddev/issues/5493#issuecomment-1800307952) and it worked well.

It looks like there is still development going on, but please let me know if you'd like me to try again at some point.

hanoii:
rebased

hanoii:
I think tests are all passing now, just restarted them as the failures were all random network issues

stasadev (approved):
Tested it on Linux with recipes for backdrop, drupal, and plain php.
The only thing that I noticed is a small copypaste mentioned above.

pkg/ddevapp/backdrop.go
```diff
@@ -224,3 +224,13 @@ func backdropPostStartAction(app *DdevApp) error {
 	}
 	return nil
 }
+
+// getDrupalComposerCreateAllowedPaths returns fullpaths that are allowed to be present when running composer create
+func getBackdropComposerCreateAllowedPaths(app *DdevApp) ([]string, error) {
+	var allowed []string
+
+	// drushrc.php path
+	allowed = append(allowed, filepath.Join(filepath.Dir(app.SiteSettingsPath), "drushrc.php"))
```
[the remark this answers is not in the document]
hanoii: No, backdrop does create the drushrc.php, whether that's necesasary or not, I am not sure as I have never used backdrop, but as it's a D7 fork, I'd assume it's ok. I'll fix the comment.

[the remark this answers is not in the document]
rfay: drushrc.php is still created in backdrop, https://github.com/ddev/ddev/blob/16870030d464e7bb64a569685b65871ce5669469/pkg/ddevapp/backdrop.go#L217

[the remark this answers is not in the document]
stasadev: Got it, thank you!

cmd/ddev/cmd/composer-create.go
```diff
@@ -64,52 +63,44 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
...
-
-		for _, o := range objs {
-			// Preserve .ddev, .git, .tarballs
-			if o == ".ddev" || o == ".git" || o == ".tarballs" {
-				continue
+		appRoot := app.GetAbsAppRoot(false)
+		if appRoot != composerRoot {
+			// If composer root is not the app root, fail if it's not fully empty.
+			if !fileutil.IsDirectoryEmpty(composerRoot) {
+				util.Failed("Failed to create project: '%v' has to be empty", composerRoot)
```
rfay, line 75: I don't understand this limitation completely. It's often the case that the composer_root is not the approot (Acquia always does it this way, composer_root: docroot). But that means that we would be creating docroot/sites/default/settings.php for example there. What is the limitation here? Fail if composer_root is not fully empty?

rfay (changes requested):

cmd/ddev/cmd/composer-create.go
hanoii: Good catch, I never consider the possibility of docroot and composer_root being the same. I'll remove that and do the validation on composer_root all the time!

hanoii: this simplifies it even more!

pkg/nodeps/utils_test.go
hanoii: Resolved as per our zoom chat

cmd/ddev/cmd/composer-create_test.go
hanoii: We also sorted this one out on our zoom call

cmd/ddev/cmd/composer-create.go
hanoii: Fixed and tests improved!

hanoii:
@rfay I fixed that and improved the tests to also account for failures and a special case (only for drupal10) where docroot and composer root are the same.

stasadev (approved):
Tested with a custom `--composer-root` and it is correctly whitelisted.

rfay:
Rebased

rfay:
Thanks for all the work on this. I don't think it will go into v1.22.5 this week just because we don't have enough time to review and test carefully again, and since I'll be in the southern hemisphere with you in December and January it will linger until next year I think. I apologize for that, but we'll keep it rebased.

hanoii:
@rfay that's fine, what you consider best

rfay (commented):
Could you take a look at the case of a Drupal10 site with composer_root=docroot and docroot=docroot and `mkdir -p docroot/sites/default/files/xxx` (that should be allowed, right?)

I get:
```
$ ddev composer create drupal/recommended-project -y
Failed to create project: '/home/rfay/workspace/broken/docroot/sites/default/files/xxx' is not allowed to be present. composer create needs to be run on a recently init project with only the following paths: [docroot docroot/sites docroot/sites/default docroot/sites/default/.gitignore docroot/sites/default/files docroot/sites/default/files/sync docroot/sites/default/settings.ddev.php docroot/sites/default/settings.php]
```

hanoii:
No, it expects it to be empty, only the path (both dirs and files) whitelisted are allowed, not recursively.

rfay:
Understood, thanks.

rfay (approved):
I got to spend some time with this and it's good to go, thanks! 

I added one commit, please take a look and see if you're OK with it @hanoii . 

It's great that this prevents accidental problems using this command, and I think even if there are any flaws people will understand and clean up their project successfully.

This can go into v1.22.5 this week.

hanoii:
@rfay all good with your last commit!

gitressa:
Thanks @hanoii and @rfay! Maybe add this in the issue summary, so this PR and its issue gets linked and automatically closed as well? Closes #5493.

A very impressive effort, and getting the feature available already in the next release is great!

hanoii:
@gitressa thanks! it's already merged so adding that wouldn't make a difference, and also, I think that @rfay steer that isssue into another direction which goes beyond what this PR fixes, which is running ddev config in wrong directories.

gitressa:
All right, well as long as a lot of folders are not deleted accidentally I am glad, so I closed the issue, and will update the title :)

PR #5502, opened 2023-11-04
docs: updates building-contributing.md pr title to comply with the linter

bmartinez287:
The linter complains unless the PR titles are all lowercase or at least start with a lowercase.

## The Issue

The linter complains unless the PR titles are all lowercase or at least start with a lowercase.

Provides an example that matches said expectation.

### DIFF
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -300,13 +300,13 @@ The pull request title must follow this convention which is based on the [Conven
 
 Must be one of the following:
 
-* **build**: Changes that affect the build or external dependencies
-* **ci**: Changes to our CI configuration files and scripts
-* **docs**: Documentation only changes
-* **feat**: A new feature
-* **fix**: A bugfix
-* **refactor**: A code change that neither fixes a bug nor adds a feature
-* **test**: Adding missing tests or correcting existing tests
+* **build**: changes that affect the build or external dependencies
+* **ci**: changes to our CI configuration files and scripts
+* **docs**: documentation only changes
+* **feat**: a new feature
+* **fix**: a bugfix
+* **refactor**: a code change that neither fixes a bug nor adds a feature
+* **test**: adding missing tests or correcting existing tests
 
 Regarding SemVer, all types above except `feat` increase the patch version, `feat` increases the minor version.
 


rfay:
How about if we just make the linter less oppressive?

bmartinez287:
That's a valid point. What did you have in mind?
Make it so it allows the first letter after the colon to be uppercase or allow any uppercases regardless of placement? the ladder is easier but could open a can of worms. That said, you may have something else in mind.

bmartinez287:
Checking commit messages against "^(build|ci|docs|feat|fix|refactor|test)(|!): [a-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)"
Looks like that regular expression is what checks on the title.

bmartinez287:
/(build|ci|docs|feat|fix|refactor|test)(|!): [A-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)/gm

Looks like switching [a-z] for [A-z] fixes this issue by making that check case insensitive.
thoughts? @rfay

bmartinez287:
The one above allows any character to be uppercase anywhere in the string. If we just want the first letter to be uppercase like in the docs we could do this.
^(build|ci|docs|feat|fix|refactor|test)(|!): [A-Z]\S+[a-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)

[A-Z]\S+
is the key.
That said, let's you set only the first letter after the colon and the space as an uppercase.

bmartinez287:
https://regex101.com/r/d9JiAk/1
^(build|ci|docs|feat|fix|refactor|test)(|!): ([A-Z]{1})([a-z]).+?((, fixes #\d+)*)($|\n|\r|\n\r)

The pattern above is interesting because it forces one to make the first letter after the colon space an upercase

this will pass
docs: Updates building-contributing.md PR title to comply with the linter.

this will fail.
docs: updates building-contributing.md DEMO title to comply with the linter.

rfay:
There's no reason to make it fail on lower-case. Just allow either. So `[A-Za-z]+` would work there. I think we can get more bang for the rules there in general, not just this one, hope we can start to use these features more like some other projects do.

rfay:
Closing this in favor of a follow-up PR that would make the linter more generous. Thanks!

PR #5503, opened 2023-11-04
docs: add feedback to the analytics section in mkdocs.yml

rpkoller:
## The Issue
@rfay is currently trying to add search priorities for the search field on ddev.readthedocs.com to prioritize search results (https://github.com/ddev/ddev/pull/5500). The problem is there is currently no direct way for a visitor to provide some sort of in context feedback about the search results and the docs page the person was pointed to in particular. Therefore there is no direct way to get an idea about potential problems and ambiguities users face in the docs experience. 
 
There is the option to add a `Was this page helpful?`widget to each docs page in mkdocs (see https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/#was-this-page-helpful). It is basically adding two icons (thumbs up and thumbs down) at the bottom of the page. on thumbs up everything was all right and it is forwarding the value 1 to the analytics software (per default google analytics - not sure how well it will work with plausible). on thumbs down the value 0 is forwarded to analytics and there is the option to leave a comment what actually went wrong. out of the box the mkdocs outline two options. squidfunk itself uses google forms. on the upside that doesnt require any account but on the downside it has gdpr implications therefore i would would vote against it. in addition there is also no way to get a reply to your feedback. so if you have a question you as the person providing a feedback depend on the maintainers to act on your feedback and solve your issue or provide some sort of reaction proactively. the other option is to instead of pointing a google form link to simply point to a github issue. the downside is a person reporting needs to have a github account but on the upside the maintainers have a direct way to provide some actual feedback on the feedback. i consider that a way better experience and a win win for both sides. 

i've created an initial PR #5503 where the github feedback link points to 

`https://github.com/ddev/ddev/issues/new/?title=[Feedback]+{title}+-+{url}&labels= documentation&template=bug_report.yml` 

creating a bug_report with a label of documentation. until a dedicated repo for mkdocs feedback is set up. otherwise the ddev queue might be "spammed". and i wanted to put the PR up to get some feedback on the microcopy for the ratings and the widget in general. as soon as the other repo is created the link could be changed to something like this: 

`https://github.com/ddev/mkdocs-feedback/issues/new/?title=[Feedback]+{title}+-+{url}&labels=feedback&template=mkdocs_feedback.yml`

any thoughts and or feedback?

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -89,7 +89,7 @@ extra:
           name: Search results and/or this page could be improved
           data: 0
           note: >-
-            Thanks for your feedback! Help us improve this page and the search experience in general by creating an issue <a href="https://github.com/ddev/ddev/issues/new/?title=[Feedback]+{title}+-+{url}&labels=feedback&template=mkdocs_feedback.yml" target="_blank" rel="noopener">on Github</a>.
+            Thanks for your feedback! Help us improve this page and the search experience in general by creating an issue <a href="https://github.com/ddev/ddev/issues/new/?title=[Feedback]+{title}+-+{url}&labels=feedback&template=bug_report.yml&labels=documenation" target="_blank" rel="noopener">on Github</a>.
 
 # Extensions
 markdown_extensions:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -77,6 +77,19 @@ extra:
   generator: false
   analytics:
     provider: plausible
+    feedback:
+      title: Did this page help you solve your problem?
+      ratings:
+        - icon: material/thumb-up-outline
+          name: This page was helpful
+          data: 1
+          note: >-
+            Thank you for your feedback! Happy
+        - icon: material/thumb-down-outline
+          name: Search results and/or this page could be improved
+          data: 0
+          note: >-
+            Thanks for your feedback! Help us improve this page and the search experience in general by creating an issue <a href="https://github.com/ddev/ddev/issues/new/?title=[Feedback]+{title}+-+{url}&labels=feedback&template=mkdocs_feedback.yml" target="_blank" rel="noopener">on Github</a>.
 
 # Extensions
 markdown_extensions:


rpkoller:
oh it looks like it might only work with google analytics according to the following comment on one issue: https://github.com/squidfunk/mkdocs-material/issues/4151#issuecomment-1312093549

cuz when i've taken a look at the docs preview the feedback widget isnt showing up :/

rfay:
How can I manually test this? I visited https://ddev--5503.org.readthedocs.build/en/5503/users/providers/ for example, and see nothing. 

Please fill in the manual testing section.

You reference https://github.com/ddev/ddev/pull/5499 but it doesn't seem to have anything to do with this.

rfay:
Rebased to current (where search priorities are in there)

rpkoller:
> How can I manually test this? I visited https://ddev--5503.org.readthedocs.build/en/5503/users/providers/ for example, and see nothing.

As mentioned on Discord it looks like that the feature requires Google Analytics. i've also added the link in the comment here for reference: https://github.com/ddev/ddev/pull/5503#issuecomment-1793462972

The docs were sort of misleading, creating the impression it would work with analytics in general in stead of requiring Google Analytics :/ not sure if there would be a way to use it with plausible as well.

> Please fill in the manual testing section.
> 
> You reference #5499 but it doesn't seem to have anything to do with this.

That was an error. No idea how i added that PR number. I've updated the issue summary. thanks for the note!

rfay:
Should we close this if it requires Google Analytics?

rpkoller:
Probably the best choice except adding support for another analytics software would be super easy and simple. basically it might be only one or two functions and it might be possible to use the GA code as a template for plausible.

rfay:
So the obvious path forward is just to close this because it can't work with Plausible?

rpkoller:
Then let's close the issue for now. And in case some new movement comes in (perhaps some way to get plausible integrated into mkdocs) then reopen this issue.

PR #5507, opened 2023-11-07
docs: mention normal upload_dirs for shopware 5

waldicom:
add the default directory for the shopware 6 projects

### DIFF
--- a/docs/content/users/usage/cli.md
+++ b/docs/content/users/usage/cli.md
@@ -62,7 +62,8 @@ To import static file assets for a project, such as uploaded images and document
 * For Drupal projects, these are the `sites/default/files` and `../private` directories.
 * For Magento 1 projects, this is the `media` directory.
 * For Magento 2 projects, this is the `pub/media` directory.
-* For Shopware projects, this is the `media` directory.
+* For Shopware 5 projects, this is the `media` directory.
+* For Shopware 6 projects, this is the `public/media` directory.
 * For TYPO3 projects, this is the `fileadmin` directory.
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/cli.md
```diff
@@ -62,7 +62,8 @@ To import static file assets for a project, such as uploaded images and document
 * For Drupal projects, these are the `sites/default/files` and `../private` directories.
 * For Magento 1 projects, this is the `media` directory.
 * For Magento 2 projects, this is the `pub/media` directory.
-* For Shopware projects, this is the `media` directory.
+* For Shopware 5 projects, this is the `media` directory.
```
rfay, line 65: I'm not sure exactly how this works since there isn't explicit support for Shopware 5. Maybe:

```suggestion
* For Shopware 5 projects (usually of type `php`), this will be the `media` directory.
```

rfay (changes requested):
I think this is OK, just suggest a little nuance in supporting something we don't have explicit support for.

docs/content/users/usage/cli.md
rfay: I guess the problem here is that unless you set `upload_dirs` then DDEV won't do anything special with Shopware 5.

rfay:
I don't think this is correct. Unless your Shopware 5 project has `upload_dirs: [media]` your suggestion won't work, and since DDEV has no explicit support for Shopware 5, it won't be set by DDEV, and people haven't been instructed.

Closing this for now, happy to continue the conversation here or elsewhere.

PR #5514, opened 2023-11-08
docs: allow uppercase letters on pr titles.

bmartinez287:
## The Issue

Allows the PR title to hold an uppercase.
relates #5502

It changes the regular expression to allow uppercase letters.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Check Type
         uses: gsactions/commit-message-checker@v2
         with:
-          pattern: '^(build|ci|docs|feat|fix|refactor|test)(|!): [a-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-check.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Check Type
         uses: gsactions/commit-message-checker@v2
         with:
-          pattern: '^(build|ci|docs|feat|fix|refactor|test)(|!): [a-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)'
+          pattern: '^(build|ci|docs|feat|fix|refactor|test)(|!): [A-Za-z].+?((, fixes #\d+)*)($|\n|\r|\n\r)'
```
rfay, line 22: Looks right to me. Can you test it on your fork?

rfay: Don't forget to change the docs with this.

bmartinez287: Well, the docs already match this pattern since they ask for the first letter to be an upperacase.
It would be tricky to test it maybe as a post-commit hook but that be a bit different.

rfay: Oh, didn't even realize that was the reason for your other PR.

PR #5515, opened 2023-11-08
fix: allow editing for platform.yaml

stasadev:
## The Issue

When testing Upsun, I copied the `platform.yaml` and checked its settings.
When I wanted to customize it, I removed the `#ddev-generated` line and ran `ddev restart`, but the file was reset.
That's because there is a second `#ddev-generated` in this file.

Removes `#` from the extra `#ddev-generated`.
Removes an extra line at the end of the file.
Adds a space between `#set -x` - to make it look like others `# set -x` in this file.

1. Use a regular DDEV binary (no need to generate a new binary for this PR)
2. Run `ddev start`
3. Edit `.ddev/providers/platform.yaml` - remove the `#ddev-generated` from the top.
4. Run `ddev stop` - see that file is reset.
5. Edit `.ddev/providers/platform.yaml` - remove the `#ddev-generated` from the top and edit the second entry of `#ddev-generated` on the third line to `ddev-generated`
6. Run `ddev start` - see that file is not reset.

### DIFF
--- a/pkg/ddevapp/dotddev_assets/providers/platform.yaml
+++ b/pkg/ddevapp/dotddev_assets/providers/platform.yaml
@@ -1,6 +1,6 @@
 #ddev-generated
 # Platform.sh provider configuration. This works out of the box, but can be edited to add
-# your own preferences. If you edit it, remove the `#ddev-generated` line from the top so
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks for the fix!

PR #5516, opened 2023-11-08
feat: add `TestComposerCreateCmd` to `push-pull-test-platforms`, tests #5499

hanoii:
Add TestComposerCreateCmd to push-pull-test-platforms so that we can fully test #5499

PR #5517, opened 2023-11-09
fix: allow testcmd to run as make tarket on `pull-push-test-platforms`, follow up to #5516

hanoii:
On #5516 we added the test, but it didn't run because testcmd was not part of the make target.


rfay (approved):
Yup, right thing to do, thanks.

PR #5521, opened 2023-11-09
feat: replace deb.nodesource.com with n, fixes #5509

### LINKED ISSUE #5509: Use n for managing node version
I like to pin the node version, it's a personal preference, but I normally do this on most of my projects:

```
RUN npm install -g n
RUN n install 20.9.0
```

This works very nicely and you can use almost [any node version.](https://www.npmjs.com/package/n#specifying-nodejs-versions).

I wonder if you'd consider this instead of what ddev is doing right now? I would go one step forward and add node_extra_packages and add that to the dockerfile automatically. 

Just an idea.

hanoii:
I figured this one was quick to do if you want to give it a try.

### DIFF
--- a/pkg/ddevapp/nodejs_test.go
+++ b/pkg/ddevapp/nodejs_test.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/exec"
-	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/testcommon"
 	"github.com/ddev/ddev/pkg/util"
 	asrt "github.com/stretchr/testify/assert"
@@ -41,7 +40,8 @@ func TestNodeJSVersions(t *testing.T) {
 	err = app.Start()
 	require.NoError(t, err)
 
-	for _, v := range nodeps.GetValidNodeVersions() {
+	// Testing some random versions
+	for _, v := range []string{"6", "10", "14", "16", "20"} {
 		app.NodeJSVersion = v
 		err = app.Restart()
 		assert.NoError(err)
--- a/pkg/ddevapp/nodejs_test.go
+++ b/pkg/ddevapp/nodejs_test.go
@@ -40,8 +40,8 @@ func TestNodeJSVersions(t *testing.T) {
 	err = app.Start()
 	require.NoError(t, err)
 
-	// Testing some random versions
-	for _, v := range []string{"6", "10", "14", "16", "20"} {
+	// Testing some random versions, both complete and incomplete
+	for _, v := range []string{"6", "10", "14.20", "16.0.0", "20"} {
 		app.NodeJSVersion = v
 		err = app.Restart()
 		assert.NoError(err)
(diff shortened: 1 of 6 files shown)


rfay (commented):
Requires installation in the Dockerfile right?

hanoii:
> Requires installation in the Dockerfile right?

what do you mean? It modifies the Dockerfile yes, but not the main ddev image, just the one for each project. It modifies as before when you specify a version.

rfay:
I think the initial nodejs installation should be done via n probably? 

https://github.com/ddev/ddev/blob/49323379ad5cea0217a22fc0ee1d0725061022a0/containers/ddev-php-base/Dockerfile#L64-L65

pkg/ddevapp/nodejs_test.go
```diff
@@ -41,7 +40,8 @@ func TestNodeJSVersions(t *testing.T) {
 	err = app.Start()
 	require.NoError(t, err)
 
-	for _, v := range nodeps.GetValidNodeVersions() {
+	// Testing some random versions
+	for _, v := range []string{"6", "10", "14", "16", "20"} {
```
rfay, line 44: Shouldn't this have some examples with minor versions?

rfay (changes requested):
This seems like it's close, but the docs have to be updated explaining the range of versions allowed, and the template.go comment instructions need statement that minor versions are supported right?

What about adding info about "Why do we need nvm at all then"? (Do we?)

pkg/ddevapp/nodejs_test.go
hanoii: Sure

hanoii:
> What about adding info about "Why do we need nvm at all then"? (Do we?)

This an interesting question. I don't think we do, but some people are used to it, so removing it altogether is a bit step, I guess that if you removed phpmyadmin you can do that with this and move it to an add on, which might not be a bad idea but I'd do that as a separate issue/PR.

I'll add some more info to template.go, I did change that to account for any "node version" but I can try thinking of something more specific.

hanoii:
Further more, using n once you start the project is not possible (this can be improved with permissions):

```
n install 18
  installing : node-v18.18.2
       mkdir : /usr/local/n/versions/node/18.18.2
mkdir: cannot create directory ‘/usr/local/n/versions/node/18.18.2’: Permission denied

  Error: sudo required (or change ownership, or define N_PREFIX)
```

1. Obviously N_PREFIX can point to something else, even the shared mnt, but again, if people are used to nvm I thought easier to just leave `n` as a system node installer and `nvm` as a user runtime `node` installer. I wonder how much nvm is used, maybe you have that information out of your analytics?

hanoii:
Updated docs and tests

stasadev (approved):
Impressive simplification!

It also "fixes" the problem described here:
- https://github.com/ddev/ddev/pull/5473

rfay:
Rebased.

rfay (approved):
Rebased and added minor docs changes. I think this is pretty fantastic. It will resolve the conflict people always feel between major version and nvm.

mandrasch:
Awesome change, thanks so much @hanoii! 👏 👏 👏

rfay:
I'm quite a fan of this as well. Simply fantastic.

tyler36:
Brilliant!

rfay:
A followup on this:
* `nodejs` inside the container is still the one in /usr/bin
* `node` is in /usr/local/bin/node
* Docs refer to `nodejs` a few places
* We could easily update to add a link /usr/local/bin/nodejs -> /usr/local/bin/node to solve this.

PR #5524, opened 2023-11-09
build: use latest version of mailpit at image build time

syssi:
## The Issue

Mailpit is outdated.

Bumps the mailpit version to the recent release.

`ddev launch -m` -> About -> v1.9.10

## Related Issue Link(s)

This project (https://github.com/pushpak1300/cypress-mailpit) cannot be used at the moment because it uses recent API features which aren't available in v1.8.1.

## Release/Deployment Notes

Bumps the mailpit version to v1.9.10.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -47,7 +47,7 @@ RUN chmod -f ugo+rwx /usr/local/bin /usr/local/bin/composer
 ### ---------------------------ddev-webserver-dev-base--------------------------------------
 ### Build ddev-webserver-dev-base from ddev-webserver-base
 FROM ddev-webserver-base as ddev-webserver-dev-base
-ENV MAILPIT_VERSION="v1.8.2"
+ENV MAILPIT_VERSION="v1.9.10"
 ENV CAROOT /mnt/ddev-global-cache/mkcert
 ENV PHP_DEFAULT_VERSION="8.1"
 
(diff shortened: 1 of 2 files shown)


rfay:
I think it would be better to get latest mailpit release like we do for so many other things, want to change it to do that?

syssi:
Could you point me to an example how to lookup the latest release?

rfay:
Thanks, all of these do it, https://github.com/ddev/ddev/blob/0225d6f483adc606e9da1ffaf650c06d56f2882d/containers/ddev-webserver/Dockerfile#L103-L109

Please manually test the process, test your curl. I'll have to push an image with your change before it actually gets used, but we'll let it run the container tests anyway.

Since we have an automated test for mailpit now I think it's OK to do this.

stasadev:
Built a new image for mailpit and ran tests ⬇️

stasadev (approved):
Thanks! Tested the binary from PR, Mailpit is v1.9.10.

rfay (approved):
Yay, thanks! And thanks for the maintenance push @stasadev !

mediaessenz:
What can I do to get the latest version of mailpit NOW?
I need it because of it's /view/latest endpoint

stasadev:
Hi @mediaessenz,

You can use the DDEV build from HEAD.

How to get it: https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-latest-commits-on-head

PR #5525, opened 2023-11-10
feat: launch port global custom command helper, fixes #5518

### LINKED ISSUE #5518: Add facility to make it easy to add `launch` type custom commands that open browser with correct URL

Recently, I created a custom command to launch the xhgui service in a browser.
This is currently quite complex;
- checks if HTTP / HTTPS
- checks for Gitpod or Codespaces
- checks if DEBUG is enabled
- case for linux
- case for darwin
- case for window

It would be convient to have a centralized helper function that custom command developers can call.
This would allow all open-in-browser commands to be updated if, say, Codespaces is updated.

### Describe your solution

I would like DDEV to provide a helper function to create a new command.
For example:

```
#!/bin/bash shell

## #ddev-generated
## Description: Launch a browser to XHGui
## Usage: xhgui
## Example: "xhgui"

HTTP_PORT = 8142
HTTPS_PORT = 8143

launch_service(HTTP_PORT, HTTPS_PORT )
```

### Describe alternatives

_No response_

### Additional context

Several modules, including the `launch` command would benefit from this:

Eg.
- ddev-xhgui
- ddev-adminer

tyler36:
## The Issue
When creating custom commands and addons, its common to supply a helper function to open the browser at a specific port.

These commands must account for different operating system and cloud environments leading to a lot of duplication.

This PR attempts to centralize these functions, so that a custom command may be created as such:

```
#!/bin/bash

## #ddev-generated
## Description: Launch xhgui
## Usage: xhgui
## Example: "xhgui"

DDEV_XHGUI_HTTP_PORT=8142
DDEV_XHGUI_HTTPS_PORT=8143

ddev port --http-port $DDEV_XHGUI_HTTP_PORT --https-port $DDEV_XHGUI_HTTPS_PORT
```

## Related Issue Link(s)
Fixes
- #5518


rfay (commented):
You may want to add this to a test.

But definitely one of the launchers needs to be changed by this PR to demonstrate the usage.

The way it is right now, `launch_port` will show up in the commands list, which we probably don't want.

tyler36:
> The way it is right now, launch_port will show up in the commands list, which we probably don't want.

Is there a reason why not?

> You may want to add this to a test.
If `launch` command is converted to use this, that should cover partiy testing.

rfay:
> Is there a reason why not show launch_port?

I just try to keep the total commands shown to a minimum, so people don't have to deal with things that don't apply to them. For example, Drupal users don't see `ddev typo3`.

rfay (commented):
This is pretty great! What do you think about making `launch_port` a bash function instead, and just including it? Then it wouldn't show up in the list of commands, and it also should be more performant, not having to go through the command building process again?

tyler36:
How do I do that? Just wrap it in `function launch_port(){}`?
Where would it live?

rfay:
I *think* it could live in ~/.ddev/commands/host, it's a predictable place, DDEV can install it, it's just not a command.

tyler36:
OK, time for review and feedback.

The latest update inverts the responsibility from the initial commit.

- `ddev launch` has returned to its primary function of "launch"-ing the browser as specified.
   - new `-s` & `-p`  set preferred ports to used depending on wether the site is HTTPS or HTTP
   - new `-u` can be used to specify a specific URL. Useful to quickly open _any_ URL such as offical docs or production version.

- `ddev port` is a helper function, that passes values to `ddev launch`. 
  - acts as an alias command and improves readability. 
  - great for testing because it shorter than typing `ddev launch -s 3000` vs. `ddev port 3000`

- `mailpit` command was updated (https://github.com/ddev/ddev/pull/5642)
   - "launch" command defers to the "mailpit" command which is responsible for setting the requests parameters using the "-s", "-p" or "-u" arguments as required. This 
Ideally, addons will follow whatever convention set here.  I'm not sure if using `ddev port` or `ddev launch` is better.

stasadev:
Rebased because the artifact bot comment links were invalid (see #5662).

rfay (changes requested):
Please take a look at TestLaunchCommand and get it fixed, thanks so much for all your contributions!

tyler36:
The last push fixed an issue and hopefully the tests. Colima failed yesterday and is currently running again ... 🤞

tyler36:
Updated PR summary.

tyler36:
Anything else for this?

rfay:
Sorry it's neglected, thanks for getting it here. Will get to it.

rfay:
Rebased, manually reviewing. @stasadev @GuySartorelli would appreciate if you could take a final look at this, thanks!

pkg/ddevapp/global_dotddev_assets/commands/host/port
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+## #ddev-generated
+## Description: Helper function to launch a browser to the project URL based on the port.
+## It has 3 main uses:
+##
+## - "ddev port 3000": Open the site on port 3000.
+## - "ddev port -s 3000": Open the site, preferring a secure port.
+## - "ddev port -p 3000": Open the site, preferring an insecure (-p) port.
```
stasadev, line 9: Insecure port doesn't work as expected:
```
ddev port -p 8025
# http://https//project.ddev.site:-p
```

rfay (changes requested):
It seems like everything works OK. I guess I would prefer if I used `-p` it would use the http port?

pkg/ddevapp/global_dotddev_assets/commands/host/port
GuySartorelli, line 1: This is really intended to be used by other commands, right? And not used in the CLI directly?
If so, it might be good to add an annotation that allows commands to be hidden, so they don't appear in the commands list for `ddev help`.

That can obviously be done separately, just thinking "out loud" here.

pkg/ddevapp/global_dotddev_assets/commands/host/port
GuySartorelli, line 9: What is a "secure" or "insecure" port? We're passing the port in, so does this actually change which port is used? Or does this just change whether `http` or `https` is used?
Either way, this needs to be clarified in the annotation.

GuySartorelli (changes requested):
Since this is intended to be used in custom commands, there should probably be some documentation about its usage on the custom commands documentation page.

pkg/ddevapp/global_dotddev_assets/commands/host/port
GuySartorelli, line 1: Seems like this command should fail if no port is passed in?

GuySartorelli:
Not sure if it's just a gitpod thing or not, but with gitpod all attempts to use `ddev launch` or `ddev port` just open the https URL for the project, ignoring all arguments and flags passed to those commands.

rfay:
On gitpod we don't use the ddev-router, so we have to do complicated things there - Anything that's going to work has to use their router (and their TLS endpoint).

tyler36:
> I guess I would prefer if I used -p it would use the http port?

You would prefer `-p` for HTTPS? What to use for HTTP?

pkg/ddevapp/global_dotddev_assets/commands/host/port
tyler36: The additional switches make it easier to add custom commands. 
When creating new add-ons, I often need to open the site to a specific port and this makes that easier too. 
Also helps with Discord support.

GuySartorelli: Right. But my point is it's not intended to run `ddev port` from the CLI directly, so ideally this command wouldn't display for `ddev help`.

And again this could be done in a separate PR since that capability isn't implemented yet for custom commands.

pkg/ddevapp/global_dotddev_assets/commands/host/port
GuySartorelli: It is a bit confusing to me that port is an argument, but you also pass it as flags...
What happens if I go `ddev 3000 -p 8080 -s 443`?

I'd recommend:
1. removing the argument (the idea is this is used by custom commands not directly in CLI, so clarity is preferred over convenience)
1. renaming the flags to `https` and `http` with no short syntax
1. clarifying in the usage (and in docs which need to call out this new feature) what it means to "prefer" one port over the other (i.e. even without the argument, what happens with `ddev port --http=8080 --https=443`?)

GuySartorelli: Might even make sense to only allow a single port - and then `--http` and `--https` could be booleans to indicate which protocol to use. Use https by default and if `--https` is passed, and `--http` only if that flag is passed. Fail if both are passed.

stasadev:
Rebased and tried to fix what was discussed here.

I used `-p` for HTTPS and `-P` for HTTP, renamed --https_port` to `--https-port` and `--http_port` to `--http-port`.

Changed things:
- the flag for HTTPS port has the highest priority (HTTP still will be used if HTTPS is not available)
- flags in `ddev port` have higher priority than the port provided as an argument (see `ddev port -h`)
- HTTP port now opens correctly
- added validation check for ports in the form of numbers, otherwise the default values will be used (443 for HTTPS and 80 for HTTP)

rfay (commented):
I am very sorry to say after all the effort put into this that I'm not feeling like it's going to be successful. We've ended up adding complexity (I think) in search of a simple approach. Unless I'm mistaken, the problem ended up being larger than expected. 

I guess I'd rather have duplicated code than obscure configuration directives.

Saying it out loud for discussion.

GuySartorelli:
The original issue description suggests a bash function would be created, and that bash function could be called from custom commands:

```sh
launch_service($HTTP_PORT, $HTTPS_PORT)
```

I'd change it to one of these:

```sh
# $PORT is the port to launch
# $IS_HTTPS is either `true` or falsy - is the port for HTTPS or not?
# $PATH is an optional path to append to the end of the URL
launch_service($PORT, $IS_HTTPS, $PATH)
```
or
```sh
# $HTTP_PORT is the the HTTP port if HTTPS is unavailable
# $HTTPS_PORT is the HTTPS port (if there is one)
# $PATH is an optional path to append to the end of the URL
launch_service($HTTP_PORT, $HTTPS_PORT, $PATH)
```

> [!IMPORTANT]
> The first above seems nicer but assumes that all projects have HTTPS, which some may not if they're not using the router. The second is more in line with how `ddev launch` currently works and allows for projects without HTTPS support, but requires commands to provide two ports.

I think if we took a step back and used that approach instead of adding a new `ddev port` command, this might all be a fair amount simpler.

Then each command just needs to know:
1. What port should I use?
1. Is that port for HTTP or HTTPS?
1. Is there any path to append to the URL (e.g. `ddev launch` accepts an argument for this)

and then call the function.

There's no complexity of having an argument but also two flags, and what to do when an argument **and** flags are passed.... instead, we effectively just pull the following existing logic from the `launch` command into that function:

```sh
# Before all of the below, we should do some quick validation e.g. check that a port has been passed and is a number - and exit 1 if not.

# Start project if not already running
  ddev start
fi

# Get URL and check if HTTPS or not
# This is directly copied from `ddev launch` - and it goes with my second suggested change above.
# IF using the first suggested change above (where the script calling the function declares whether it's HTTP or HTTPS),
# we could build the full URL using the first item in $DDEV_HOSTNAME and slapping the appropriate protocol in front.
FULLURL=${DDEV_PRIMARY_URL}
HTTPS=""
if [ ${DDEV_PRIMARY_URL%://*} = "https" ]; then HTTPS=true; fi

# Set correct port - this example assumes we're using the second option above.
if [ "${HTTPS}" = "" ]; then
    FULLURL="${FULLURL%:[0-9]*}:${1}"
else
    FULLURL="${FULLURL%:[0-9]*}:${2}"
fi

# Append path argument to URL, if there is one
if [ -n "${3:-}" ] ; then
  if [[ ${3::1} != "/" ]] ; then
    FULLURL="${FULLURL}/";
  fi

  FULLURL="${FULLURL}${3}";
fi

# Just output debug
if [ ! -z ${DDEV_DEBUG:-} ]; then
    printf "FULLURL $FULLURL\n" && exit 0
fi

# Do the actual launch
case $OSTYPE in
  linux-gnu)
    if [[ ! -z "${GITPOD_INSTANCE_ID}" ]]; then
        gp preview ${FULLURL}
    else
        xdg-open ${FULLURL}
    fi
    ;;
  "darwin"*)
    open ${FULLURL}
    ;;
  "win*"* | "msys"*)
    start ${FULLURL}
    ;;
esac
```

And `ddev launch` just needs its existing while loop (modified to set the relevant variables), followed by 
```sh
launch_service($HTTP_PORT, $HTTPS_PORT, $PATH)
```

Mailpit can similarly just call the function, passing in its ports and no path.

The only tricky bit then is figuring out a nice way to give custom bash commands access to the function - but I'm sure there's a clean solution to that if we look for it.

GuySartorelli:
> The only tricky bit then is figuring out a nice way to give custom bash commands access to the function - but I'm sure there's a clean solution to that if we look for it.

In fact we can probably just do something like `source <path-to-script-defining-that-function> &&` before running the host command.

tyler36:
I think DDEV need a command to `launch` to a specific port. I'm just not sure of the best way. 
I was tryng to keep all the "launch stuff" centralized.

I'd be OK only targeting HTTPS ports.

GuySartorelli:
> I think DDEV need a command to launch to a specific port.

I don't see the need for that - but even if that's a desire, that's not what the linked issue is about.

GuySartorelli:
Even if a `ddev port` command was added, it could use the function I described above.

stasadev:
Oh, I see, this complication is not good. I didn't expect to fail the tests.

I like the way `ddev launch` can take a full URL in this PR and I think we don't even need a new `--url` flag for this, it can be smart enough to take a URL as an argument.

In this case, adding ports to launch something will be much easier, and if we agree to use HTTPS all the time, the new `ddev port` command won't even be needed:

```
ddev launch "${DDEV_PRIMARY_URL}:${CUSTOM_PORT}"
```

rfay:
This has been a good learning opportunity, but I don't think this is going to solve the general problem in its current form, and it would be better with a restart. I do apologize because you've put so much effort into this @tyler36 but I hope that the things we've learned can inform a new approach to this. Closing for now.

stasadev:
Thanks for your input @tyler36, I've opened 

- #5986

Hope a small change can make things easier for you.

PR #5528, opened 2023-11-10
docs: use make test instead of make testpkg

hanoii:
When I first starting working with tests I got used to doing make testpkg TESTARGS="-run Test.." and I didn't know why this hand't worked with commands. Because we are providing the test as argument I gured we can include all tests in the make command so that this is less prone to happen.


rfay (approved):
Thanks, good addition.

PR #5531, opened 2023-11-12
docs: provide infromation about mapping error during CLI debugging

RuZniki:
## The Issue
On PHPStorm PHP server name setting is important.

Documentation is great, but I spend several hours before finding solution how to make CLI debugging. Add some information how it can be solved.

## Related Issue Link(s)
https://github.com/php-perfect/ddev-intellij-plugin/issues/227

### DIFF
--- a/docs/content/users/debugging-profiling/step-debugging.md
+++ b/docs/content/users/debugging-profiling/step-debugging.md
@@ -77,7 +77,6 @@ However, if you’ve not yet used PhpStorm with Xdebug for a regular web request
 
 If you have error: "Can't find a source position. Server with name 'SITE_NAME.ddev.site' doesn't exist.". Try to change the "PHP | Servers" **Name** to _SITE_NAME.ddev.site_. You can see on screenshot above that both Name, Host are "_d8composer.ddev.site_".
     
-
 ### Visual Studio Code (VS Code) Debugging Setup
 
 1. Install the [PHP Debug](https://marketplace.visualstudio.com/items?itemName=felixfbecker.php-debug) extension.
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for this contribution. I'd rather see it solved in 
* https://github.com/php-perfect/ddev-intellij-plugin/issues/259

We'll see what @nico-loeber says.

Mostly I try to avoid adding "every little thing" to the docs. Would Stack Overflow be a better place to put this? 

Regardless, please change the title of this issue to have the error encountered, so people will find it.

RuZniki:
Randy, thanks for feedback, I updated issue description and title.

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -75,6 +75,8 @@ If you need to debug command-line PHP processes, especially code that’s outsid
 However, if you’ve not yet used PhpStorm with Xdebug for a regular web request, do that to automatically create the PhpStorm “server” with the same name as your primary URL (*Languages and Frameworks* → *PHP* → *Servers*). The key job of the server is to map filesystem locations on the workstation (your computer) to filesystem locations on the remote server—in this case the `ddev-webserver` container. Often, PhpStorm automatically sets up a mapping that doesn’t include the entire project, meaning things like the `vendor` directory may not be mapped. You can fix this by mapping the top-level directory of your project to `/var/www/html` in the container:
 ![PhpStorm mapping](../../images/phpstorm-server-mapping.png)
 
+If you have error: "Can't find a source position. Server with name 'SITE_NAME.ddev.site' doesn't exist.". Try to change the "PHP | Servers" **Name** to _SITE_NAME.ddev.site_. You can see on screenshot above that both Name, Host are "_d8composer.ddev.site_".
```
rfay, line 78:
```suggestion
If you encounter the error: "Can't find a source position. Server with name 'SITE_NAME.ddev.site' doesn't exist", change the "PHP | Servers" **Name** to _SITE_NAME.ddev.site_. In the example screenshot above both Name and Host are "_d8composer.ddev.site_".
```

rfay (changes requested):
I think this may belong here OK, as long as you remember to remove it when the extension gets fixed :) Suggested minor verbiage change that you can just accept if you like it.

RuZniki:
Thanks for you suggestion. Accepted. I do my best to remove it as soon it would be fixed.

rfay (approved):
Thanks!

PR #5508, opened 2023-11-07
fix: ensure no duplicates for docker networks, fixes #5193

stasadev:
## The Issue

- https://github.com/docker/compose/issues/6532#issuecomment-769263484
- #5193

## Summary:

Sometimes when you start a project, multiple project networks are created (e.g. `ddev-your-project_default`) which prevents the project from starting.

```
Error response from daemon: network ddev-your-project_default is ambiguous (16 matches found on name)
```

The project network is *internal* (created from `docker-compose.yaml`) in v1.22.4, and we cannot manipulate its state.
But we can make it *external* (created programmatically from DDEV) in v1.22.5 and control it however we want.

This set of PRs are intended to fix the need for user to do manually `docker network rm`, because the project cannot be started when there are duplicate networks.

An attempt has been made to fix this on the Docker side (but it will only be available with Docker 25):

- https://github.com/moby/moby/pull/46251

To make things easier for DDEV users, we can use external networks for DDEV projects and automatically remove duplicates.

## Code changes for this issue are made in:
- #5305
- #5422
- #5436
- #5508

## How This PR Solves The Issue

DDEV should remove Docker networks by ID, not by name. This way users won't need to worry about duplicate networks, because they will be removed automatically on `ddev start` and `ddev poweroff` - regardless of what users run.

## Manual Testing Instructions

Download the binary below, this is a special broken binary that creates duplicate networks:

* [all-ddev-executables.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925836.zip)
* [ddev-linux-amd64.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925838.zip)
* [ddev-linux-arm64.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925840.zip)
* [ddev-macos-amd64.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925842.zip)
* [ddev-macos-arm64.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925844.zip)
* [ddev-windows-amd64-installer.zip](https://nightly.link/ddev/ddev/actions/artifacts/1034925846.zip)

```
# use the binary from above for ddev start, the more you do it, the more duplicates will be created
ddev start
ddev start
ddev start

# then run poweroff (networks will not be removed)
ddev poweroff

# see a lot of duplicates for ddev_default and ddev-your-project_default
docker network ls

# switch the binary to the one in the second comment (from the bot) and run

ddev start

# you will see a lot of these:

[...]

PR #5533, opened 2023-11-13
fix: internal project network with duplicate check, revert for #5305

stasadev:
## The Issue

- #5305
- #5508
- #5193

After switch to external project networks in #5305, there were issues with compatibility between different DDEV version, which made downgrade not smooth.

The last issue with downgrade from HEAD to v1.22.4:

```
$ ddev start
...
Building project images... 
Project images built in 3s. 
time="2023-11-07T08:29:14-07:00" level=warning msg="a network with name ddev-d10_default exists but was not created by compose.\nSet `external: true` to use an existing network" 
network ddev-d10_default was found but has incorrect label com.docker.compose.network set to "" 
Failed to start d10: composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-d10 docker-compose -f /Users/rfay/workspace/d10/.ddev/.ddev-docker-compose-full.yaml up -d', action='[]', err='exit status 1', stdout='', stderr='time="2023-11-07T08:29:14-07:00" level=warning msg="a network with name ddev-d10_default exists but was not created by compose.\nSet `external: true` to use an existing network"
network ddev-d10_default was found but has incorrect label com.docker.compose.network set to ""'
```

Move back to internal project networks, and create them before `docker-compose up` to ensure no duplicates.

I've refactored the code to make it clearer what's going on. In the previous PR #5305, it was unclear where and why the project network was created.

Use DDEV as usual.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1028,6 +1028,10 @@ func (app *DdevApp) Start() error {
 
 	app.DockerEnv()
 	dockerutil.EnsureDdevNetwork()
+	// "docker-compose up", which is called in this function later, may create
+	// duplicate project networks, we can create this network in advance
+	// see https://github.com/ddev/ddev/pull/5533
+	dockerutil.EnsureProjectNetwork()
 
 	if err = dockerutil.CheckDockerCompose(); err != nil {
 		util.Failed(`Your docker-compose version does not exist or is set to an invalid version.
@@ -2556,11 +2560,6 @@ func (app *DdevApp) Stop(removeData bool, createSnapshot bool) error {
 			return fmt.Errorf("failed to process post-stop hooks: %v", err)
 		}
 	}
-
-	// Remove current project network
-	// Working around duplicate network creation problem
-	// see https://github.com/ddev/ddev/pull/5508
-	dockerutil.RemoveNetworkWithWarningOnError(os.Getenv("COMPOSE_PROJECT_NAME") + "_default")
 
 	return nil
 }
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -41,7 +41,7 @@ type ComposeCmdOpts struct {
 }
 
 // EnsureNetwork will ensure the Docker network for DDEV is created.
-func EnsureNetwork(client *docker.Client, name string) error {
+func EnsureNetwork(client *docker.Client, name string, labels map[string]string) error {
 	// Pre-check for network duplicates
 	RemoveNetworkDuplicates(client, name)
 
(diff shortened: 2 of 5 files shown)


stasadev:
The only difference between a network created with `docker-compose up` and from DDEV is this set of labels:

```
$ docker inspect ddev-your-project_default
...
        "Labels": {
            "com.docker.compose.network": "default",
            "com.docker.compose.project": "ddev-your-project",
            "com.docker.compose.version": "2.23.0"
        }
...
```

And when you have a network with these labels created programmatically, `docker-compose up` runs as usual.

I see this as a win because with this approach we can have a project network at an early stage, I guess there might be some kind of race condition that creates duplicates for the networks in `docker-compose up`.

pkg/ddevapp/ddevapp.go
```diff
@@ -1028,6 +1028,10 @@ func (app *DdevApp) Start() error {
 
 	app.DockerEnv()
 	dockerutil.EnsureDdevNetwork()
+	// "docker-compose up", which is called in this function later, may create
+	// duplicate project networks, we can create this network in advance
+	// see https://github.com/ddev/ddev/pull/5533
+	dockerutil.EnsureProjectNetwork()
```
stasadev, line 1034: This PR is easy to revert, you only need to remove this function call and the function itself.

pkg/dockerutil/dockerutils.go
```diff
@@ -95,7 +107,8 @@ func NetworkExists(netName string) bool {
 func RemoveNetwork(netName string) error {
 	client := GetDockerClient()
 	networks, _ := client.ListNetworks()
-	var err error
+	// the loop below may not contain such a network
+	var err error = &docker.NoSuchNetwork{ID: netName}
```
stasadev, line 111: This is a fix for a bug that I introduced in #5508

When you run `ddev stop` or `ddev poweroff` several times in a row, `Network <name> removed` is shown each time, even if there is no such network.

pkg/testcommon/testcommon.go
```diff
@@ -262,12 +262,8 @@ func ClearDockerEnv() {
...
 // ContainerCheck determines if a given container name exists and matches a given state
 func ContainerCheck(checkName string, checkState string) (bool, error) {
-	// Ensure we have Docker network
-	client := dockerutil.GetDockerClient()
-	err := dockerutil.EnsureNetwork(client, dockerutil.NetName)
-	if err != nil {
-		log.Fatal(err)
-	}
+	// Ensure we have DDEV network
+	dockerutil.EnsureDdevNetwork()
```
stasadev, line 266: There is only refactoring.

stasadev:
There are some testing errors, I restarted them and they failed again.
Don't really know if these failures are related to this PR.

PR #5538, opened 2023-11-14
docs: add instructions how to add colima to the macos login items

rpkoller:
## The Issue
with the release of colima 0.6.0 it is possible to add colima to the macos login items by using `brew services`. just a small update to the docker-installation.md file. i am just not sure if it makes sense to add `brew services list` for educational services as well so people are able to validate what the command actually did, where things where copied, and verify everything was done as expected.

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -25,7 +25,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     After the initial run above, you can use `colima start` or use `colima start -e` to edit the configuration file. Run `colima status` at any time to check Colima’s status.
 
-    When your computer restarts, you’ll need to `colima start` again. Starting with Colima 0.6.0 you are able to automate the startup process. `brew services start colima` adds Colima to the list of login items for your MacOS user account. With `brew services list` you are able to verify if the service's plist file was correctly added to `~/Library/LaunchAgents`.
+    When your computer restarts, you’ll need to `colima start` again. Starting with Colima 0.6.0 you are able to automate the startup process. `brew services start colima` adds Colima to the list of login items for your MacOS user account. With `brew services list` you are able to verify if the service's property list file was correctly added to `~/Library/LaunchAgents`.
 
     !!!tip "Colima disk allocation"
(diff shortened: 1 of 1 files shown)


docs/content/users/install/docker-installation.md
```diff
@@ -22,10 +22,10 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     colima start --cpu 4 --memory 6 --disk 100 --vm-type=qemu --mount-type=sshfs --dns=1.1.1.1
     ```
     (On macOS versions before Ventura, omit the `--vm-type=qemu` flag as it doesn't work on older OS versions.)
-    
+
     After the initial run above, you can use `colima start` or use `colima start -e` to edit the configuration file. Run `colima status` at any time to check Colima’s status.
 
-    When your computer restarts, you’ll need to `colima start` again. This will eventually be automated in later versions of Colima.
+    When your computer restarts, you’ll need to `colima start` again. Starting with Colima 0.6.0 you are able to automate the startup process. `brew services start colima` adds Colima to the list of login items for your macOS user account. With `brew services list` you are able to verify if the service's property list file was correctly added to `~/Library/LaunchAgents`.
```
rfay, line 28:
```suggestion
    When your computer restarts, you’ll need to `colima start` again. If you prefer Colima to start automatically on reboot, in Colima `v0.6+` use `brew services start colima` to configure auto-start.
```

rfay (changes requested):
Minor suggestions.

docs/content/users/install/docker-installation.md
rpkoller: You would strike the part about `brew services list` again? But i am ok with removing it again. The only i am unsure about is using Colima twice. would something like that work? 

```
If you prefer Colima to start automatically on reboot, use `brew services start colima` in version 0.6+ to configure auto-start.
```

rfay: People don't really need to know anything except the `brew services start` part. It gets pretty wordy otherwise. Nobody knows what a plist is or cares, for example

rpkoller: i agree (i think the only person who was interested in that `brew services list` detail was me probably) . and would my rephrased suggestion work?

rfay: Yes, minor changes:

> If you prefer to start Colima automatically on reboot, use `brew services start colima` in Colima version 0.6+ to configure auto-start. 

should work.

rpkoller: I've pushed the changes with the version you've quoted. i'll set the conversation to resolved.

rfay (approved):
Thanks!

PR #5541, opened 2023-11-15
test: check for duplicate networks, for #5508

stasadev:
## The Issue

- #5508

Docker networks can be created programmatically with the same name (this is no longer possible with Docker >=25.x.x):

- https://github.com/moby/moby/pull/46251

This test creates two networks with the same name, removes duplicates, and checks that everything is OK.

## Automated Testing Overview

See `TestNetworkDuplicates`

### DIFF
--- a/cmd/ddev/cmd/networks_test.go
+++ b/cmd/ddev/cmd/networks_test.go
@@ -1,0 +1,62 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/ddev/ddev/pkg/dockerutil"
+	docker "github.com/fsouza/go-dockerclient"
+	asrt "github.com/stretchr/testify/assert"
+)
+
+// TestNetworkDuplicates makes sure that Docker network duplicates
+// are deleted successfully with DDEV
+// See https://github.com/ddev/ddev/pull/5508
+// Note: duplicate networks cannot be created with Docker >=25.x.x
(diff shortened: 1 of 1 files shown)


cmd/ddev/cmd/networks_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Create the first network
+	_, err := client.CreateNetwork(netOptions)
+	assert.NoError(err)
+
+	// Create a second network with the same name
+	_, err = client.CreateNetwork(netOptions)
+	if err != nil {
+		t.Logf("This test is run with Docker >=25.x.x, making it unrepresentative: %v", err)
+	}
```
stasadev, line 53: I'm not sure how to properly check for a duplicate here.

If I add `assert.NoError(err)`, it works for Docker 24, but not for Docker 25:
```
API error (409): network with name ddev-TestNetworkDuplicates_default already exists
```

So I can leave it as it is or add a strict `assert.NoError(err)` check and then the test will fail on every runner where Docker 25 is installed.

rfay: Would it be worth checking for docker server version? Shouldn't be hard to do that using the api.

stasadev: That's what I should be doing here. Thank you!

rfay (approved):
Thanks!

PR #5550, opened 2023-11-19
ci: don't run buildkite tests on docs

rfay:
## The Issue

Buildkite tests sometimes run when only docs are changed, no code. 

Try to use depends to prevent this instead of checking inside the test.sh


rfay:
This may not actually have value; I think it's just doing the same thing that was being done before in test.sh. I was hoping for something that happened earlier in the pipeline.

rfay:
Response from Buildkite support:

> The most Buildkite-y (or elegant) way to skip a step from running is using our [step conditionals](https://buildkite.com/docs/pipelines/conditionals), However, there is currently no straightforward way to do that if the step condition required is on identifying if code changes exists on a PR.

If the step condition cannot be evaluated with the supported if conditional variables, the suggested approach is to run a step to evaluate the condition, and upload another step as a result of the previous step.
```
 steps:
- label: "Check PR code change"
command: .buildkite/test_code_diff.sh
key: "gitdiff-check"
- label: "Run remaining tests if code changes exists"
depends_on: "gitdiff-check"
command: |
if [ $$(buildkite-agent step get "outcome" --step "gitdiff-check") == "passed" ]; then
cat <<- YAML | buildkite-agent pipeline upload
steps:
- command: ".buildkite/test.sh"
plugins:
- docker-login#v2.1.0:
username: druddockerpullaccount
password-env: DOCKERHUB_PULL_PASSWORD
agents:
- "os=macos"
- "dockertype=dockerformac"
- "architecture=arm64"
env:
BUILDKITE_CLEAN_CHECKOUT: true
BUILDKITE_BUILD_PATH: ~/tmp/buildkite_builds
BUILDKIT_PROGRESS: plain
DDEV_TEST_SHARE_CMD: "true"
DDEV_RUN_GET_TESTS: "true"
# M1 has some NFS failures (TestComposer, TestProcessHooks, TestNFSMount) and as of 2021-02-18; test without NFS by default
#DDEV_TEST_USE_NFSMOUNT: true
parallelism: 1
YAML
fi
```
Cheers,

PR #5555, opened 2023-11-21
fix: add config validation for --all flag, for #5452

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/5452#pullrequestreview-1742421480

> `ddev start -a` doesn't validate config

Adds validation when you pass `--all` flag for a DDEV command.

1. Break some config value in the project's `.ddev/config.yaml` (e.g. `php_version: "test"`)
2. Run `ddev start -a` and immediately see the error.


rfay:
Rebased, mostly just to check buildkite tests.

rfay (approved):
Code is good and works perfectly in manual testing, thanks!

PR #5561, opened 2023-11-22
docs: improve steps for macos buildkite agent

stasadev:
## The Issue

Setup buildkite agent for macOS remote machines.

Simplify or make certain steps clearer.


rfay (approved):
Thanks for all the work and for catching all the things!

PR #5563, opened 2023-11-23
build: bump docker-compose to v2.23.3

stasadev:
## The Issue

Trying out docker-compose v2.23.3.


rfay (approved):
Looks like it's going to be fine, thanks.

PR #5570, opened 2023-11-25
refactor: order host, port, driver the same both places in Drupal 10 settings file

gitressa:
## The Issue

It is not a big problem, but the order of host, port, driver is switched, causing a slight "hmm" when you scan the values.

```
$host = "db";
$port = 3306;
$driver = "mysql";

$databases['default']['default']['database'] = "db";
$databases['default']['default']['username'] = "db";
$databases['default']['default']['password'] = "db";
$databases['default']['default']['host'] = $host;
$databases['default']['default']['driver'] = $driver;
$databases['default']['default']['port'] = $port;
```

Makes the order identical to the one in the default settings.php file:

```
 * $databases['default']['default'] = [
 *   'database' => 'databasename',
 *   'username' => 'sqlusername',
 *   'password' => 'sqlpassword',
 *   'host' => 'localhost',
 *   'port' => '3306',
 *   'driver' => 'mysql',
[...]
```
https://github.com/drupal/drupal/blob/10.1.x/sites/default/default.settings.php#L80-L85

### DIFF
--- a/pkg/ddevapp/drupal/drupal10/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal10/settings.ddev.php
@@ -15,8 +15,8 @@ $databases['default']['default']['database'] = "{{ $config.DatabaseName }}";
 $databases['default']['default']['username'] = "{{ $config.DatabaseUsername }}";
 $databases['default']['default']['password'] = "{{ $config.DatabasePassword }}";
 $databases['default']['default']['host'] = $host;
-$databases['default']['default']['driver'] = $driver;
 $databases['default']['default']['port'] = $port;
+$databases['default']['default']['driver'] = $driver;
 
 $settings['hash_salt'] = '{{ $config.HashSalt }}';
 
--- a/pkg/ddevapp/drupal/drupal8/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal8/settings.ddev.php
@@ -22,8 +22,8 @@ $databases['default']['default']['database'] = "{{ $config.DatabaseName }}";
 $databases['default']['default']['username'] = "{{ $config.DatabaseUsername }}";
 $databases['default']['default']['password'] = "{{ $config.DatabasePassword }}";
 $databases['default']['default']['host'] = $host;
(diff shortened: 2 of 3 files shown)


rfay (approved):
Not sure why it makes a difference, but fine with me :)

I assume that the d8 and d9 have the same difference?

gitressa:
It's a tiny change, but now they are aligned in order :)

rfay:
Order is good, right?

gitressa:
Yes, indeed :)

The reason I noticed it, was because I wanted to run [MySQLTuner](https://github.com/major/MySQLTuner-perl), and needed the host/port info. It took me a while to finally just try replacing `127.0.0.1` with `db`, and then I got a reply:

```
$ ddev ssh
ddev:$ git clone --depth 1 -b master https://github.com/major/MySQLTuner-perl
ddev:$ cd MySQLTuner-perl/
ddev:/MySQLTuner-perl$ perl mysqltuner.pl --host db
 >>  MySQLTuner 2.3.2
   * Jean-Marie Renouard <jmrenouard@gmail.com>
   * Major Hayden <major@mhtx.net>
 >>  Bug reports, feature requests, and downloads at http://mysqltuner.pl/
 >>  Run with '--help' for additional options and output filtering

[--] Skipped version check for MySQLTuner script
[!!] The --forcemem option is required for remote connections
[!!] Assuming RAM memory is 1Gb for simplify remote connection usage
[!!] The --forceswap option is required for remote connections
[!!] Assuming Swap size is 1Gb for simplify remote connection usage
[--] Performing tests on db:3306
[--] Using mysql to check login
[...]
```

rfay:
`ddev describe` tells you all the things, 

and https://ddev.readthedocs.io/en/latest/users/usage/database-management/

gitressa:
Thanks for the tips, reading the docs closer, this would have gotten me on track:

> The web and db containers are each ready with MySQL/PostgreSQL clients, so you can [ddev ssh](https://ddev.readthedocs.io/en/latest/users/usage/commands/#ssh) or ddev ssh -s db and use mysql or psql.

Issue #962, opened 2018-07-02
ddev logs --json-output (ddev logs -j) no longer works (just emits normal logs)


rfay:
**Describe the bug**

`ddev logs -j` should present logs in json format, and it used to (and I think it's important to ddev-ui). It doesn't any more.

**To Reproduce**

`ddev logs -j`

**Expected behavior**

Logs encapsulated by json

**Version and configuration information (please complete the following information):**

ddev v0.20.0

bmartinez287:
https://github.com/ddev/ddev/blob/master/pkg/output/sirupsen_orig/json_formatter.go.sirupsen

This PR look like it might have been the last one where said feature worked.
https://github.com/ddev/ddev/commit/d7b5745c00d253cb003d39cb33fdb7afd81a4d40#diff-c0a58b146c62bd1f23410e7f387a07d9be492cd2c4e3a629fcb7167cd8c1bbaf

bmartinez287:
I wonder if the reason this does not work is the same as why.
`ddev debug --json-output` does not work.
It seems like both are missing github.com/ddev/ddev/pkg/ddevapp

bmartinez287:
Actually it might be the fact is missing this package "github.com/ddev/ddev/pkg/output"

rfay:
Want to fix it @bmartinez287 ? It would be welcome!

bmartinez287:
It might tricky to fix since it's the oldest issue but I will give it a go.

GuySartorelli:
I was going to start looking at this but then I realised - I don't really know what the output should look like.
Should it be an array with each line as its own entry?
And how should it behave when the `--tail` flag is used?

rfay:
> I'm starting to lean towards "ddev logs should not support json output"

I'm ok with that. Maybe just make it so if it's used with `-j` it gives an info statement saying not supported or something.

GuySartorelli:
Are there any other commands (other than composer) which currently don't respect that flag? And if so do they also output a similar message?

rfay:
I don't think so. And I think that 6-7 years ago this one may have worked with -j :)

PR #5572, opened 2023-11-26
fix: ensure help information respects `--json-output` flag, fixes #5567

### LINKED ISSUE #5567: `ddev help` doesn't respect `--json-output`

```
Running bash [-c /tmp/test_ddev.sh] 
Please try running this in an existing DDEV project directory, preferably the problem project.
It doesn't work in other directories.
Failed running test_ddev.sh: exit status 2
. You can run it manually with `curl -sL -O https://raw.githubusercontent.com/ddev/ddev/master/cmd/ddev/cmd/scripts/test_ddev.sh && bash test_ddev.sh`
```

Which makes sense - this bug report has nothing to do with projects. I don't have any current ddev projects to run the command in, and I don't think the output is relevant so I'm not gonna bother spinning one up unless I really have to.
</details>         

### Expected Behavior

The help information should respect the global `--json-output` or `-j` flag, and output its information in JSON format. This would make the output easy to interpret programmatically.

### Actual Behavior

The output is the same as though you hadn't used the `--json-output` or `-j` flag. This means it is not easy to interpret programmatically.

### Steps To Reproduce

1. Run `ddev help <command> --json-output`
2. Observe that the output is _not_ in JSON format.

### Anything else?

I'm writing a wrapper around DDEV that allows me to further customise the DDEV CLI. Part of this requires being able to programmatically interpret the output from the help command so that my wrapper can use that information to know what flags any given DDEV command will accept, among other things. I'm doing that with REGEX at the moment, but theoretically the way the output is displayed could change tomorrow and break it. If I could get the information as a JSON object, the normal text output could change without breaking my code (along with being easier to fetch the relevant information in the first place).

GuySartorelli:
## The Issue

- #5567

Replaces the default help function with one that outputs a JSON-formatted representation of a command when the `--json-output` flag is used.

Run any version of the help command or `-h` flag in conjunction with the `--json-output` (or `-j`) flag, e.g:
- `ddev help -j`
- `ddev auth -hj`
- `ddev auth ssh -hj`

## Automated Testing Overview

I haven't added tests yet, because I want to make sure the approach is correct and acceptable before I do so. I wouldn't want to write tests only for them to be entirely invalidated by sweeping changes required in the implementation.

Once a maintainer has told me that the approach for the implementation is good, I'll add some tests.

## Related Issue Link(s)

#5567 

## Release/Deployment Notes

N/A

## Other notes

This is my first time contributing here and my first time writing anything in go - so it's likely I've done things in a completely backwards way - I don't have a good background on go conventions, nor this codebase, nor the cobra codebase or conventions, so please make sure to explain things clearly and in full when requesting changes.

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -194,6 +194,8 @@ func init() {
 			util.Warning("Adding custom/shell commands failed: %v", err)
 		}
 	}
+
+	setHelpFunc(RootCmd)
 }
 
 // checkDdevVersionAndOptInInstrumentation() reads global config and checks to see if current version is different
--- a/cmd/ddev/cmd/root_help_json.go
+++ b/cmd/ddev/cmd/root_help_json.go
@@ -1,0 +1,153 @@
+package cmd
+
+import (
+	"github.com/ddev/ddev/pkg/output"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// Sets the help function of a command to one which respects the --json-output flag.
+func setHelpFunc(command *cobra.Command) {
+	originalHelpFunc := command.HelpFunc()
+	command.SetHelpFunc(func(command *cobra.Command, strings []string) {
+		if !output.JSONOutput {
+			// Use the default help function for plain text output
+			originalHelpFunc(command, strings)
+		} else {
+			// LogSetup already happened in init, but that's before flags were parsed.
+			// It seems like the --help flag bypasses the second call of LogSetUp after parsing
+			// flags, which means the output doesn't want to output JSON.
+			output.LogSetUp()
+
+			// JSON representation of a subcommand
+			type jsonSubCommand struct {
+				Name        string
+				Description string
+			}
+
+			// JSON representation of a group of commands
+			type jsonCommandGroup struct {
+				Name     string
+				Commands []*jsonSubCommand
+			}
+
+			// JSON representation of a CLI flag
+			type jsonFlag struct {
+				Name                string
+				Shorthand           string
+				Usage               string
+				Default             string
+				Deprecated          string
+				Hidden              bool
+				ShorthandDeprecated string
+			}
+
+			var jsonCommands []*jsonSubCommand
+			var jsonCommandGroups []*jsonCommandGroup
+			var jsonAdditionalCommands []*jsonSubCommand
+			var jsonAdditionalHelpCommands []*jsonSubCommand
+			var jsonFlags []*jsonFlag
+			var jsonGlobalFlags []*jsonFlag
+
+			// Build list of subcommands. Logic reflects the default "usage" template from cobra
+			if len(command.Groups()) == 0 {
+				// Direct subcommands, if there are no groups
+				for _, subCmd := range command.Commands() {
+					if subCmd.IsAvailableCommand() || subCmd.Name() == "help" {
+						jsonCommands = append(jsonCommands, &jsonSubCommand{
+							Name:        subCmd.Name(),
+							Description: subCmd.Short,
+						})
+					}
+				}
+			} else {
+				// Groups of subcommands
+				for _, group := range command.Groups() {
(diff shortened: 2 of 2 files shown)


rfay (changes requested):
Thanks so much for giving this a try! This is looking great. The one thing I'd say is probably structure the json the way all the other commands are structured; take a look at `ddev describe -j` for example, where the `raw` has the json, `msg` has the plain text. SO impressive! Thanks.

rfay:
(And since your head is in this, I'd love it if you'd take a look at 
* https://github.com/ddev/ddev/issues/962

which is our oldest open issue :) )

GuySartorelli:
I've updated this to use `output.UserOut.WithField()` to output the JSON, which seems to be what the describe command is doing... and that's fine for using the help command like so:
```shell
ddev help -j
ddev help list -j
```

But if I try to use the help _flag_, we get the "test" output which indicates the `--json-output` flag is captured correctly, but we _don't_ get any of the data I'm shoving into `raw`:
```shell
ddev -hj
ddev list -h -j
```

Do you know what might be causing the JSON to not be output correctly when `-h` is used as a flag?

GuySartorelli:
Nevermind, I think I found it. `output.LogSetUp()` needs to be called a second time. It's called once in root.init but that's _before_ the `--json-output` flag is parsed. See https://github.com/ddev/ddev/blob/1016bc7b623086cb200137e5d06b1e5665022b80/cmd/ddev/cmd/root.go#L40-L42

Let me know what further changes are required with the implementation as a whole - once you're happy with that I'll get some tests written.

cmd/ddev/cmd/root.go
```diff
@@ -194,6 +195,149 @@ func init() {
...
+			rawResult["Commands"] = jsonCommands
+			rawResult["CommandGroups"] = jsonCommandGroups
+			rawResult["AdditionalCommands"] = jsonAdditionalCommands
+			rawResult["AdditionalHelpCommands"] = jsonAdditionalHelpCommands
+			rawResult["Flags"] = jsonFlags
+			rawResult["GlobalFlags"] = jsonGlobalFlags
+			rawResult["Deprecated"] = command.Deprecated
+			rawResult["Hidden"] = command.Hidden
+
+			output.UserOut.WithField("raw", rawResult).Print()
```
GuySartorelli, line 338: Note that I can't pass the raw help output into the `Print()` call, since the original help function prints directly to the terminal rather than returning its result.

rfay:
I guess the one thing I don't like right now is that it's inline in the root.go init(). init() is evil anyway, but could you refactor it to put it in a function in root_help_json.go or something?

GuySartorelli:
> could you refactor it to put it in a function in root_help_json.go or something?

I can - but unless you say otherwise I'm going to call it literally `root_help_json.go` - "or something" is too ambiguous for me to do anything with :p

rfay:
I can live without a test on this one, but would love fix and test for the logs one :) (new PR :) )

GuySartorelli:
JSON format for logs is less important to me personally, so I won't promise anything on that front.

stasadev (approved):
It works good.

The only exceptions I found (for example, `ddev help composer`) do not support the `-j` flag, which is expected:
- https://github.com/ddev/ddev/pull/4283

GuySartorelli:
> I can live without a test on this one

Given this statement, is there anything left to do for this PR? Seems ready to merge, I think?

rfay:
I hope to give it a run.

GuySartorelli:
> I hope to give it a run.

Does that mean you want to try run it locally to test it before merging?
I know you said you're okay without automated tests for this, but if I provided some would that speed things along at all?

rfay:
Sorry I didn't get to test before leaving. Maybe @stasadev will get a chance to look. Thanks for the contribution.

stasadev (approved):
The code looks fine and doesn't seem to break anything, I'll pull it today.

GuySartorelli:
Thanks!

PR #5573, opened 2023-11-26
fix: php8.3 should allow environment variables to php-fpm

rfay:
## The Issue

@rpkoller and I were studying the failure of Drupal 10 to install and we discovered a regresssion - $IS_DDEV_PROJECT was not available to settings.php, meaning that it didn't behave correctly and load settings.ddev.php.

The problem was that the PHP8.3 fpm/pool.d/www.conf didn't have the correct setting for `clear_env`, which should be "no" for DDEV

* Set it correctly
* Also noted another change from php8.2 that was not intentional, daemonize = no and catch_workers_output=yes
* Added test to make sure we get environment variables

Do a Drupal10 installation with php8.3, it should work.

## Automated Testing Overview

Coverage of this regression was added in TestPHPConfig

### DIFF
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -968,14 +968,12 @@ func TestPHPConfig(t *testing.T) {
 
 	// Most of the time there's no reason to do all versions of PHP
 	phpKeys := []string{}
-	phpKeys = nodeps.GetValidPHPVersions()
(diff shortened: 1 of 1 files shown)


rpkoller:
I've tested with the same ansible setup i've previously noticed the issue with. with the downloaded artifact i was able to install a site and the settings.ddev.php was found and the set database credentials available for the drupal install on php8.3.

PR #5576, opened 2023-11-27
build: update .gitpod.yml for v1.22.5 release [skip ci]

stasadev:
## The Issue

DDEV v1.22.5 is out.

Push new gitpod image.


rfay (approved):
Thanks. I imagine we should improve this process so we don't have to do this.

PR #5577, opened 2023-11-27
ci: don't set default context to default - fix rancher desktop failures

rfay:
## The Issue

Multiple failures of rancher-desktop tests (see https://github.com/ddev/maintainer-info/issues/1#issuecomment-1828804959) have resulted in the docker context being left as "default" but the default context isn't wired to any provider. Orbstack and Docker Desktop both claim "default", but Rancher Desktop doesn't.

* Don't set context to default
* Leave the provider running that was running at the end of test

We'll see if this works.

**This temporarily allows rancher to run on this PR, but that should be removed before pulling**

### DIFF
--- a/.buildkite/macos-rancher-desktop.yml
+++ b/.buildkite/macos-rancher-desktop.yml
@@ -7,7 +7,7 @@
       - "os=macos"
       - "rancher-desktop=true"
       - "architecture=arm64"
-    branches: "master"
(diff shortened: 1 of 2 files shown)


rfay:
I guess we should leave rancher desktop running on PRs for now because otherwise it's too hard to study
* https://github.com/ddev/ddev/issues/5578

PR #5579, opened 2023-11-28
build: bump ddev-php-base and ddev-webserver to base on Debian 12 bookworm, replaces #5278

rfay:
## The Issue

This replaces #5278 - At the time that was done the build didn't succeed, but now it may. 

Resurrect #5278 

Try it out.


rfay:
The bookworm ddev-webserver image is about 40MB bigger than bullseye, https://hub.docker.com/r/ddev/ddev-webserver/tags

rfay:
Now we have to think about when it's appropriate to introduce this. Maybe it's no big deal. It's possible it could change some people's Dockerfiles or webimage_extra_packages.

rfay:
Yay, this works!

rfay:
Thanks for pointing out the sources.list change. Sounds *very* positive, what a great idea.

Any thoughts about when we should pull this? I'm thinking that since it doesn't involve docs it could be quite soon, and then we'd discover anything about Bookworm we need to know in the next month or two before release.

stasadev:
I think we can pull it very soon.

rfay:
I'll pull it. Any open PRs with image dependencies will need a new image rebuild.

PR #5586, opened 2023-11-29
fix: disable Xdebug for composer command, fixes #4219

### LINKED ISSUE #4219: Executing `ddev composer` should disable XDebug

stasadev:
## The Issue

- #4219

Adds `XDEBUG_MODE=off` for `ddev composer`.
`os.Setenv("XDEBUG_MODE", "off")` is not working here, because `ddev composer` runs inside `docker-compose exec`.

1. `ddev start`
2. `ddev xdebug`
3. `ddev composer -V` - should not should Xdebug output.


rfay:
Rebased

rfay (approved):
This seems like an elegant fix, and works correctly. I also manually tested with PhpStorm listening for xdebug. And tried with `nc -l 9003` on the host. Without this, xdebug connects to 9003 when you do a `ddev composer` command, and with this patch it does not.

rfay:
Related: https://github.com/composer/xdebug-handler

PR #5592, opened 2023-11-30
docs: GitHub Codespaces: Adjust examples to make sure docker is running 

eiriksm:
## The Issue
I was trying to follow the guide and something has probably changed in codespaces since then. I found postAttachCommand works though

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -403,7 +403,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     A lot more customization is possible via the [`devcontainer.json`-configuration](https://containers.dev/implementors/json_reference/). You can install Visual Studio Code extensions by default or run commands automatically. 
 
-    #### `postCreateCommand`
+    #### postCreateCommand
 
     The [`postCreateCommand`](https://containers.dev/implementors/json_reference/) lets you run commands automatically when a new codespace is launched. DDEV commands are available here.
 
@@ -411,7 +411,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     You usually want to use a separate bash script to do this, as docker [might not yet be available when the command starts to run](https://github.com/devcontainers/features/issues/780).
 
-    ```
+    ```json
     {
     "image": "mcr.microsoft.com/devcontainers/universal:2",
     "features": {
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -403,7 +403,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     A lot more customization is possible via the [`devcontainer.json`-configuration](https://containers.dev/implementors/json_reference/). You can install Visual Studio Code extensions by default or run commands automatically. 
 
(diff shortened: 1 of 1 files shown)


rfay:
@mandrasch could you review this please?

rfay (changes requested):
Failed spellcheck due to postAttachCommand

mandrasch:
Hi, thanks for contribution! 

As posted in Discord

- postCreateCommand is the correct event in my opionion
- the devcontainer / codespaces team fixed it previously - docker must be available there, otherwise it's a bug in docker-in-docker on their side (universal image)

So please don't remove this from the docs, it's the correct event to install things on first creation of a codespace (e.g. install scripts, etc.)

Saw this in other codespaces demos as well, they all used postCreate

(Scripts / routines would run twice if postAttach is used and the container is restarted as far as I understand - postCreate is really for the first initial steps on a fresh codespace)

Depends on the use case of course - we could add postAttach to the docs as second option?

Please change PR accordingly if relevant / useful in your opinion :)

Cheers

rfay:
I'm going to change this to draft status for now; @mandrasch if you can keep us up-to-date with the upstream problem it will be appreciated. 
* https://github.com/devcontainers/features/issues/671#issuecomment-1834370756

eiriksm:
Amazing, thanks! 😍

I'm just glad it's nothing wrong with the docs or ddev, so I just agree with everything said in this thread basically

mandrasch:
> Amazing, thanks! 😍
> 
> I'm just glad it's nothing wrong with the docs or ddev, so I just agree with everything said in this thread basically

👍 👍 

just fiy: I submitted a new issue here https://github.com/devcontainers/features/issues/780

mandrasch:
fyi: new comment https://github.com/devcontainers/features/issues/780#issuecomment-1851200127

rfay:
It does sound like perhaps a new approach is in order @mandrasch ? based on 
* https://github.com/devcontainers/features/issues/780#issuecomment-1851200127

mandrasch:
> It does sound like perhaps a new approach is in order @mandrasch ? based on
> 
> * [Docker not available in postCreateCommand (universal image) devcontainers/features#780 (comment)](https://github.com/devcontainers/features/issues/780#issuecomment-1851200127)

yeah... I'll see when I'll find time in the next weeks and I'll post here again. happy travelling in the meantime!

(if anyone else here reading this and wants to try it sooner - please feel free to test the new retry approach 😎)

eiriksm:
Tested the script and slightly altered it. Seems to work ✌️

eiriksm:
Updated with new command suggestion. Removed the inline bash example since that must be considered unstable at this point I guess 🤓✌️

mandrasch:
Awesome work @eiriksm, thanks very much for taking the time to solve this and editing the docs! 👏 🎉  🎉  🎉  

(I'll update https://github.com/mandrasch/ddev-craftcms-vite in the next days or weeks with your new retry logic)

mandrasch:
Works fine for me, good to go from my point of view! 👍 

Thanks very much again! 

One last thing: @eiriksm, could you please change the title of your PR?

eiriksm:
👍done

Thanks for your hard work @mandrasch ♥️

stasadev:
Thank you!

PR #5594, opened 2023-12-01
ci: make TestMailpit and TestProjectPortOverride more forgiving of Rancher Desktop, fixes #5578

### LINKED ISSUE #5578: Rancher tests are failing TestMailpit (intermittent?)
TestMailpit should succeed on all platforms. It's failing on Rancher Desktop, https://buildkite.com/ddev/macos-rancher-desktop/builds/95#_

Oddly, after several failures on master, it succeeded on PR -  https://buildkite.com/ddev/macos-rancher-desktop/builds/100#018c1357-d690-4514-aacc-0786506cdc2a

rfay:
## The Issue

* #5578

TestMailpit fails intermittently on Rancher Desktop

This may be due to the mailpit API coming up after our healthcheck says it should be ready, because we're not polling the API.

Wait in the test a few seconds if necessary

It's possible this could also be fixed by adding more extensive healthcheck.sh to check the mailpit API, but that seems like overkill for something that is not used a huge amount.

### DIFF
--- a/pkg/ddevapp/mailpit_test.go
+++ b/pkg/ddevapp/mailpit_test.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 )
 
 // TestMailpit does a basic test of mailpit.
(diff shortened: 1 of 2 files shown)


rfay:
TestProjectPortOverride is also flaky on Rancher at this line, failed here: https://buildkite.com/ddev/macos-rancher-desktop/builds/120#018c22cd-9094-41f8-9452-60e03ce15bcd

https://github.com/ddev/ddev/blob/7e872a7e1f6ba690ac2141d79ba8be6469b53a24/pkg/ddevapp/router_test.go#L116

rfay:
This seems to be working OK for TestMailpit at this point. 

However, TestProjectPortOverride often fails on rancher. It's possible there's something wrong with our test runner[s] and the port is still occupied somehow as well.

rfay:
It looks like the healthcheck for traefik may also be checking not-exactly the right thing; it waits for the web server, but that's not the same as all the work being done. We may be able to improve the traefik healthcheck at some point, but https://github.com/ddev/ddev/pull/5594/commits/4e58ac214a0878366c8fadf0ca2c5c52c2db9b2a just tries to give the test a little more time.

rfay:
I think it worked!

rfay (commented):
This is tests only, so I'm going to go ahead and pull it as it can stop the bleeding on some of the other PRs. We'll still have bleeding on Colima tests though.

rfay:
If you have things you'd rather do with this @stasadev they're welcome as followup.

stasadev:
I think it is good enough for tests.

PR #5600, opened 2023-12-01
ci: Skip TestGetLocalHTTPResponse on Colima

rfay:
I think Colima is probably failing everywhere, but this PR just skips TestGetLocalHTTPResponse for now. If it gets farther as a result we'll know something; I imagine it will just fail on another test.


rfay:
This is a step forward although it doesn't fix all the problems with Colima.

PR #5603, opened 2023-12-04
fix: dont auto start host custom commands, fixes #2688

### LINKED ISSUE #2688: Host shell commands (custom commands) should not auto-start a project
**Describe the bug**

Custom shell commands currently start the project they're in if it's not already started. This makes sense for web commands, but not so much for host commands.

**To Reproduce**

With a project stopped, issue a command that is a host command. It will start the project.

**Expected behavior**

**Screenshots**

- Host computer OS and Version: [e.g. Windows 10, macOS Catalina]
- Docker Desktop version if on macOS or Windows (from "About Docker Desktop")

**Additional context**

rfay:
## The Issue

* #2688 

Host shell commands don't always need to start project

* No auto-start for host commands
* Host commands that need project running can start it

Try out `ddev launch` and the various database launcher commands
Try out `ddev self-upgrade` and it should not auto start, or try another custom host command.

## Release/Deployment Notes

* We probably need to visit some add-ons that provide custom commands and update them if needed.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
@@ -7,6 +7,11 @@
 ## OSTypes: darwin,linux
 ## HostBinaryExists: /Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
 
+if [ ${DDEV_PROJECT_STATUS} != "running" ]; then
+  echo "Project ${DDEV_PROJECT} is not running, starting it"
+  ddev start
+fi
+
 database="${1:-db}"
 user="${2:-root}"
 type="$(echo $DDEV_DATABASE | sed 's/:.*//')"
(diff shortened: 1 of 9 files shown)


pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
```diff
@@ -7,6 +7,11 @@
 ## OSTypes: darwin,linux
 ## HostBinaryExists: /Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
 
+if [ ${DDEV_PROJECT_STATUS} != "running" ]; then
```
stasadev, line 10:
```suggestion
if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
```

pkg/ddevapp/global_dotddev_assets/commands/host/sequelpro
```diff
@@ -8,6 +8,10 @@
 ## HostBinaryExists: /Applications/Sequel Pro.app
 ## DBTypes: mysql,mariadb
 
+if [ ${DDEV_PROJECT_STATUS} != "running" ]; then
```
stasadev, line 11:
```suggestion
if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
```

pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
```diff
@@ -9,6 +9,11 @@
 ## OSTypes: darwin
 ## HostBinaryExists: /Applications/TablePlus.app
 
+if [ ${DDEV_PROJECT_STATUS} != "running" ]; then
```
stasadev, line 12:
```suggestion
if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
```

stasadev (changes requested):
It works great!

Small suggestion: if `${DDEV_PROJECT_STATUS}` is undefined: `[: !=: unary operator expected`, so it is better to wrap it.

> We probably need to visit some add-ons that provide custom commands and update them if needed.

This can help with addons:
```
#!/bin/bash

ddev get --list -j | jq -c '.raw[]' | while read -r item; do
    html_url=$(echo "$item" | jq -r '.html_url')
    default_branch=$(echo "$item" | jq -r '.default_branch')
    url="$html_url/tree/$default_branch/commands/host"

    if [ "$(curl -s -o /dev/null -w "%{http_code}" "$url")" -eq 200 ]; then
        echo "$url"
    fi
done
```

stasadev (approved):

Issue #5620, opened 2023-12-12
ddev-webserver container build fails because of unavailable packages.blackfire.io debian repo


syssi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
```
</details>         

Skipped.

### Expected Behavior

Successful container build. `ddev start` returns `0`.

### Actual Behavior

```
#13 [web 4/6] RUN apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests webp
#13 2.896 E: Failed to fetch http://packages.blackfire.io/debian/dists/any/InRelease  403  Forbidden [IP: 3.5.66.150 80]
#13 2.896 E: The repository 'http://packages.blackfire.io/debian any InRelease' is not signed.
#13 ERROR: process "/bin/bash -c apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::=\"--force-confold\" --no-install-recommends --no-install-suggests webp" did not complete successfully: exit code: 100
------
 > [web 4/6] RUN apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests webp:
2.896 E: Failed to fetch http://packages.blackfire.io/debian/dists/any/InRelease  403  Forbidden [IP: 3.5.66.150 80]
2.896 E: The repository 'http://packages.blackfire.io/debian any InRelease' is not signed.
------
```

### Steps To Reproduce

```
ddev config --docroot=. --project-type=php --webimage-extra-packages webp && ddev start
```

### Anything else?

Is there a workaround to get rid of blackfire.io temporary? We are dependent on the custom container build but we don't need blackfire.io usually.

mattvb91:
same issue here. Cant even downgrade either because other versions broken too with ` ComposeCmd failed to run` issues. Completely broken now across multiple versions.

nostadt:
Would love to be able to disable blackfire when not needed, or actually have it disabled by default and enable if needed. Given the price of blackfire, I assume many people don't use this.

pbowyer:
Seems to be resolved for me. Approx 1.25h after I first encountered it. Thanks Blackfire team for sorting the outage!

[...]

PR #5623, opened 2023-12-12
fix: do not fail on `apt-get update` for extra packages, for #5620

stasadev:
## The Issue

- #5620

We must not allow `apt-get update` to fail if some repository is down, because it can stop the webserver from starting.

Allows `apt-get install` to continue after a fail for `apt-get update`.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1085,7 +1085,6 @@ RUN (groupadd --gid $gid "$username" || groupadd "$username" || true) && (userad
 	if extraPackages != nil {
 		contents = contents + `
 ### DDEV-injected from webimage_extra_packages or dbimage_extra_packages
-RUN echo "Disk usage problems: " && df -h / && ls -l /var/lib/apt/lists
 RUN (apt-get -qq update || true) && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests ` + strings.Join(extraPackages, " ") + "\n"
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/config.go
```diff
@@ -1085,8 +1085,8 @@ RUN (groupadd --gid $gid "$username" || groupadd "$username" || true) && (userad
 	if extraPackages != nil {
 		contents = contents + `
 ### DDEV-injected from webimage_extra_packages or dbimage_extra_packages
-RUN echo "Disk usage problems: " && df -h /tmp && ls -l /var/lib/apt/lists
-RUN apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests ` + strings.Join(extraPackages, " ") + "\n"
+RUN echo "Disk usage problems: " && df -h / && ls -l /var/lib/apt/lists
```
stasadev, line 1088: @rfay I'm not sure if we still need this check, if we skip failures for `apt-get update`, but anyway:
I changed `df -h /tmp` to `df -h /` - we discussed this on Discord earlier and `df -h /tmp` produced less output.

rfay: Agreed we probably won't need it at all, can remove. Install of PHP related packages can still fail, but we have no control. And this check didn't actually do anything but somehow seemed to result in less failures

rfay:
Looks fine to me. I thought maybe there was a second place we do apt update.

stasadev:
There are several `apt-get update` added in Go files, all of which already have `|| true`.
This is the last one without `|| true`.

PR #5625, opened 2023-12-13
feat: add toggle arg to xdebug command, fixes #3593

### LINKED ISSUE #3593: Add ddev xdebug toggle flag
Currently ddev xdebug commands allows the following flags:

`ddev xdebug on|off|enable|disable|true|false|status`

It can either enable, disable or return the status of the xdebug.
In practice when we need to use xdebug we turn it on and then turn it off when we don't need it any more.
Those are two separate commands when we run them in our terminal window.

A good DX would be to have a 
`ddev xdebug toggle`
flag so it basically reads the current status of xdebug and inverts it.
We would then be able to use the same command like we are used with drush cr for example.

"ddev xdebug toggle" will also return the current new status after being executed.

GuySartorelli:
## The Issue

- #3593 

Adds new "toggle" argument to the `xdebug` command

1. Run `ddev xdebug toggle` - see it enable xdebug
1. Run `ddev xdebug toggle` - see it disable xdebug

Optionally run `ddev xdebug status` inbetween to see that the status has _actually_ changed.

## Automated Testing Overview

Added tests to validate that `ddev xdebug toggle` outputs the correct text to indicate the new status, and that running `ddev xdebug status` afterwards reports that the status was in fact toggled.

I'm not sure how to run tests locally so just wrote them blind and hoping they pass - but they're based on existing tests and it seems pretty straight forward so should be fine.

## Related Issue Link(s)

#3593 

## Release/Deployment Notes

I don't think this has any other ramifications.

## Other notes

I wanted to also change the default for `ddev xdebug` to be a shortcut for `ddev xdebug toggle` (it's currently a shortcut for `ddev xdebug on`) - but I figure that's probably a breaking change so if that does get done it should probably be done in a separate PR.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
@@ -15,12 +15,12 @@ fi
 xdebug_version=$(php --version | awk '/Xdebug v/ {print $3}')
 
 case $1 in
-    on|true|enable)
-	enable_xdebug
-	;;
-	off|false|disable)
-	disable_xdebug
-	;;
+  on|true|enable)
+    enable_xdebug
+    ;;
+  off|false|disable)
+    disable_xdebug
+    ;;
   toggle)
     case ${xdebug_version} in
     v3*)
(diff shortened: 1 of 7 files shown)


pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
GuySartorelli, line 1: This file already has a mix of tabs and spaces for indentation so I'm really not sure what the indentation should be - and there's no guidance on that (at least that I could find) in https://ddev.readthedocs.io/en/stable/developers/building-contributing/#coding-style

Let me know if you want something specific to be done with the indents in this file.

stasadev: Let's use spaces. You can add `[skip ci]` to the commit message to avoid running all the tests.

stasadev (approved):
@GuySartorelli, thank you!

It looks like you've made all the necessary changes in all the necessary places.

pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
stasadev: I'm going to make a commit here (will do a release soon, and want to merge it right now).

GuySartorelli:
Awesome, thank you for the quick turn around! 💙

PR #5632, opened 2023-12-14
fix: global project_tld should default to 'ddev.site', not empty, fixes #4290

### LINKED ISSUE #4290: Use the default project_tld `ddev.site` instead of "" in the `global_config.yaml`

When https://github.com/drud/ddev/issues/2746 has landed `project_tld: ""` was added to `~/.ddev/global_config.yaml` and project files got a commented out `# project_tld: ddev.site` line. As a user it isn't clear, if you only take a look at the `global_config.yaml`, if the tld entered needs to be wrapped in quotation marks and if it is necessary to prepend a period. 

### Describe your solution

Things could be more self explanatory by simply changing the entry in `~/.ddev/global_config.yaml` from `project_tld: ""` to `project_tld: ddev.site` utilizing the default tld domain. that way the user directly sees the syntax for the tld active and  globally applied per default. and that way the active tld would be visible at any given timen. 

the only caveat with my suggestion @rfay pointed out on discord is that you have to be careful in case you are changing the `global project_tld` . ddev sets the default (`project_tld: DdevDefaultTLD = "ddev.site"`) in 
https://github.com/drud/ddev/blob/af1918778c56f789568951437f0a7b6ed4b15752/pkg/nodeps/values.go#L109 
and then compares the constant `DdevDefaultTLD` with `ProjectTLD` in
https://github.com/drud/ddev/blob/5049bdfb0a2dadf1185c63972d2e238cd2120969/pkg/ddevapp/config.go#L103-L106

### Describe alternatives

_No response_

### Additional context

I've posted the idea in a comment in an already merged and closed pull request a few days ago (https://github.com/drud/ddev/pull/4218#issuecomment-1270770198). thought it would be easier to follow and track if a dedicated issue is created now instead appending the followup in an already closed pull request.

rfay:
## The Issue

* #4290

@rpkoller pointed out that the global_config.yaml would be more useful if it just *specified* ddev.site as the project_tld.

Make it explicit in the config.

- [ ] Test to make sure changing global project_tld still works fine
- [ ] Test to make sure default config with ddev.site works fine. Automated tests ought to do this just fine.

### DIFF
--- a/cmd/ddev/cmd/config-global_test.go
+++ b/cmd/ddev/cmd/config-global_test.go
@@ -65,7 +65,7 @@ func TestCmdGlobalConfig(t *testing.T) {
 	assert.Contains(out, "use-hardened-images=false\n")
 	assert.Contains(out, "fail-on-hook-fail=false")
 	assert.Contains(out, fmt.Sprintf("required-docker-compose-version=%s\nuse-docker-compose-from-path=false", globalconfig.DdevGlobalConfig.RequiredDockerComposeVersion))
-	assert.Contains(out, "project-tld=\nxdebug-ide-location=")
+	assert.Contains(out, "project-tld="+globalconfig.DdevGlobalConfig.ProjectTldGlobal+"\nxdebug-ide-location=")
 	assert.Contains(out, "router=traefik")
 	assert.Contains(out, "wsl2-no-windows-hosts-mgt=false")
 	assert.Contains(out, "router-http-port=80\nrouter-https-port=443")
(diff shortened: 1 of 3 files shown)


stasadev (changes requested):
Noticed a bug, it's not even related to the changes made in this PR (reproducible on the stable DDEV), but it probably fits here:

When you set the global TLD to `test` and the project TLD to `ddev.site`, the project config is not set, it seems that there is some hardcode in the code for `nodeps.DdevDefaultTLD`:

```
ddev config global --project-tld test
ddev config --project-tld ddev.site
```

On the other hand, this override works if I set it manually: `project_tld: ddev.site` in `.ddev/config.yaml`. But it is lost when I run `ddev config --auto` or `ddev config --project-tld ddev.site`.

rfay:
OK with me for you to fix that here.

stasadev:
I made a change in the `app.WriteConfig()`, will see how the tests go.

stasadev (approved):

stasadev:
It should be fine, I tested it again and didn't find any flaws.

PR #5635, opened 2023-12-15
fix: Solve the WSL connection failure issue, fixes #5634

### LINKED ISSUE #5634: WSL2 installation is failing to complete due to name resolution problems

<details><summary>Can't run  `ddev debug test` as this is prior to ddev being installed</summary></details>

### Expected Behavior

Follow the installation instructions and ddev should be ready to use.

### Actual Behavior

Follow the installation instructions and the installation doe not complete .

### Steps To Reproduce

On a Windows machine, follow the installation instructions at:
https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#windows

When running the command:
`iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ddev/ddev/master/scripts/install_ddev_wsl2_docker_inside.ps1'))`

The installation stops without ddev being available to use.

The error seems to be coming from one of the apt update, upgrade or install steps, e.g.: 
`wsl -u root apt-get update`
`wsl -u root -e bash -c "apt update && apt install -y ddev docker-ce docker-ce-cli containerd.io wslu"`
`wsl -u root -e bash -c "apt-get upgrade -y >/dev/null"`

It seems there is an issue with WSL resolving the server locations at the installation stage, not sure if it persists after the installation completes.

I have researched and found others having this problem which seems to be with the WSL installation rather than ddev, but the fix/workaround can be included to help here.
The issue is here: https://github.com/microsoft/WSL/issues/8365
</details>

### Anything else?

The general consensus from some of the duplicate tickets (e.g.: https://github.com/microsoft/WSL/issues/4285) seems to be to add the google nameservers to the resolve.conf file to allow WSL to find the required repositories and then complete the update steps.
```
bash -c 'echo "nameserver 8.8.8.8" >> /etc/resolv.conf';
chattr +i /etc/resolv.conf
bash -c 'echo "\n[network]\generateResolvConf=false" >> /etc/wsl.conf';
```
What I am not sure about is whether these changes are required to persist after the installation has been completed.
Hopefully, Windows will fix the issue with WSL and this will stop being an issue.

thegbomb:
## The Issue
[5634](https://github.com/ddev/ddev/issues/5634)

This adds the google nameservers to the `resolv.conf` file, which allows apt to find the repositories needed to run apt install.

Follow the instructions at as is: https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#windows to see the issue.

Remove the WSL and re run the instructions, but when running one of commands in the instructions to trigger the ddev installation script, use this command instead:
```
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/thegbomb/ddev/master/scripts/install_ddev_wsl2_docker_desktop.ps1'))
```
This command uses the forked version to install ddev, and the installation should complete as required.

## Related Issue Link(s)
Open issue for WSL: https://github.com/microsoft/WSL/issues/8365
Duplicated issue with the resolution: https://github.com/microsoft/WSL/issues/4285

### DIFF
--- a/scripts/install_ddev_wsl2_docker_desktop.ps1
+++ b/scripts/install_ddev_wsl2_docker_desktop.ps1
@@ -44,11 +44,11 @@ mkcert -install
 $env:CAROOT="$(mkcert -CAROOT)"
 setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV }
 
+wsl -u root -e bash -c "touch /etc/resolv.conf && printf 'nameserver 8.8.8.8\n' >>/etc/resolv.conf;"
 wsl -u root -e bash -c "rm -f /etc/apt/keyrings/ddev.gpg && curl -fsSL https://pkg.ddev.com/apt/gpg.key | gpg --dearmor | tee /etc/apt/keyrings/ddev.gpg > /dev/null"
 wsl -u root -e bash -c 'echo deb [signed-by=/etc/apt/keyrings/ddev.gpg] https://pkg.ddev.com/apt/ \* \* > /etc/apt/sources.list.d/ddev.list'
 wsl -u root -e bash -c "apt update && apt install -y ddev wslu"
 wsl -u root -e bash -c "apt-get upgrade -y >/dev/null"
-
 
 wsl bash -c 'echo $CAROOT'
 wsl -u root mkcert -install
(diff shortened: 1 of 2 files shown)


rfay (changes requested):
Unfortunately, this has side-effects that would be unacceptable. If a non-internet DNS server is supposed to be in play, this bypasses it and ignores it. So people in a corporate network could not access resources on their local network, or look up domain names, for example, that resolve to test machines, etc.

What we need to do is figure out what's wrong with your system, not change the instructions for everybody.

rfay:
Your contribution (and research) are very much appreciated, but this wouldn't be appropriate for the general public. We may want to add some documentation about this workaround, or perhaps we'll want to add into the script a basic test for name resolution (like `curl -f -I https://www.google.com`) but this exact approach isn't appropriate. Again, thanks for the research and let's continue the conversation in #5634

PR #5640, opened 2023-12-16
feat: first experiment with mysql8.2 and bitnami/mysql

rfay:
## The Issue

This is a very early and rough experiment with using bitnami/mysql as the upstream for mysql 8.0 and 8.2 (and 5.7)

I got an 8.0 to build. I'm not sure whether or not this will work out in the end or not. I'm only making a PR so I'll be able to find my way back to it in the future.


rfay:
Right now the problem I'm having is that they don't use /var/lib/mysql. Instead they use /bitnami/mysql. 

The environment variables specified in https://hub.docker.com/r/bitnami/mysql/ are not necessarily there at all.

Options include a hard link `ln /opt/bitnami/mysql /var/lib/mysql`

rfay:
This remains a good idea, but we won't be doing non-LTS MySQL at this point. 

It's also my understanding that VMWare (owner of Bitnami) is in quite some disruptive state right now, it's possible we should watch for a little while to see if Bitnami gets proper maintenance.

The one problem is that we have missed the window of capturing MySQL 8.0.35 arm64 from the Ubuntu build, as it's moved on to 8.0.36, but xtrabackup is still at 8.0.35. That means we're stuck with the same 8.0.34 we have in DDEV v1.22.7 for DDEV v1.23.0 unless Percona suddenly catches up with a new release of Xtrabackup.

PR #5647, opened 2023-12-18
fix: disable Content-Type auto-detection for Traefik 2, fixes #5346

### LINKED ISSUE #5346: Disable Traefik Content-Type auto-detection

I created a fresh Typo3 12 project and the install tool security check shows the following warnings:

Unexpected server response
https:// test.ddev.site/typo3temp/assets/29bd3687.tmp/c8059a1e.wrong: unexpected content-type text/html

Unexpected server response
https:// test.ddev.site/typo3temp/assets/29bd3687.tmp/c8059a1e.html.wrong: unexpected content-type text/html

As of the traefik documentation there is an auto-detection for the Content-Type which is enabled by default.
https://doc.traefik.io/traefik/middlewares/http/contenttype/

In this case traefik changes the Content-Type so the security checks fail as the response is not as expected. In my case this is not a problem for my local dev environment but could be a problem in other circumstances as you should not expect traefik to change the Content-Type. The traditional router before traefik did not change anything.

### Describe your solution

As mentioned in the traefik documentation this behavior can be disabled via middleware.

I suggest changing the autogenerated project traefik config file in the following way:

Adding these lines at the top under http:

    http:    
        middlewares:
            autodetect:  
                contentType:  
                    autoDetect: false

The middleware has than to be added to every router where this behavior should be disabled. I suggest disabling it in every router. This can be done this way:

    test-web-80-http:
      rule: HostRegexp(`test.ddev.site`)
      middlewares:
        - autodetect
      service: "test-web-80-http"
      tls: false
      entrypoints:
        - http-80

This fixed the problem and all the security checks passed.

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #5346

Disables autodetect for Content-Type in Traefik project yaml config.
Note that this fix needs to be reverted when Traefik 3 is out.

From https://github.com/ddev/ddev/issues/5346#issuecomment-1860409926
I could reproduce it when just created a fresh TYPO3 Install (Apache Webserver) and then in the Backend (Install Tool) under "Environment" and "Environment Status" I get the messages mentioned in my first post. After the changes I mentioned there there are no more unexpected server responses listed.

## Release/Deployment Notes

This fix is not needed in the upcoming Traefik 3:

- https://github.com/traefik/traefik/pull/9546

### DIFF
--- a/pkg/ddevapp/traefik_config_template.yaml
+++ b/pkg/ddevapp/traefik_config_template.yaml
@@ -10,6 +10,8 @@ http:
       {{else}}
       rule: Host({{ range $i, $h := $s.ExternalHostnames }}{{if $i}}, {{end}}`{{$h}}`{{end}})
       {{end}}
+      middlewares:
+        - autodetect
       service: "{{$appname}}-{{$s.InternalServiceName}}-{{$s.InternalServicePort}}-http"
       tls: false
       entrypoints:
@@ -22,6 +24,8 @@ http:
       {{else}}
       rule: Host({{ range $i, $h := $s.ExternalHostnames }}{{if $i}}, {{end}}`{{$h}}`{{end}})
       {{end}}
+      middlewares:
+        - autodetect
       service: "{{$appname}}-{{$s.InternalServiceName}}-{{$s.InternalServicePort}}-https"
       {{ if not $.UseLetsEncrypt }}
       tls: true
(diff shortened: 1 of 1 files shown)


rfay (commented):
This looks like a simple path, but I'm not going to be able to test it or study it. But should be fine. Does it impact docs at all?

stasadev:
Does not impact docs. This looks like a really rare case.

PR #5653, opened 2023-12-20
fix: update global schema

tyler36:
## The Issue
While editing the global config file, VScode reported 3 invalid keys.

Strangely, after a couple of minutes, the warning disappeared. However, the same 3 keys reported invalid again after a system restart.

The apostrophes also displayed a warning in VSCode:
"The character U+2019 "’" could be confused with the ASCII character U+0060 "`", which is more common in source code."

This PR 
- fixes a single typo.
- adds the missing keys to the the global schema file.
- replaces apostrophes with single quote

## Release/Deployment Notes

I understand apostrophes vs. single quotes is a personal preference. 
My intention is to remove warnings displayed when using the following setting: 
`"editor.unicodeHighlight.ambiguousCharacters": true,`.  We can revert that commit, as required.

### DIFF
--- a/pkg/globalconfig/schema.json
+++ b/pkg/globalconfig/schema.json
@@ -224,7 +224,7 @@
       "type": "boolean"
     },
     "use_letsencrypt": {
-      "description": "hether to enable Let’s Encrypt integration. (Works in conjunction with letsencrypt_email.) (Not currently compatible with Traefik router.)",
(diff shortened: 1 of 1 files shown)


jonaseberle (approved):

stasadev (approved):
Thank you!

PR #5657, opened 2023-12-22
ci: fix PostgreSQL client for testbot on WSL2

stasadev:
## The Issue

Noticed a wrong formula in tests:
```
Error: No available formula with the name "postgresql-client". Did you mean postgrest?
```
https://buildkite.com/ddev/wsl2-docker-inside/builds/3039#018c9045-77f3-48e4-879b-01dae0cfaf66/99-116

Uses the same formula from macOS setup.


rfay (approved):
Thanks for paying attention! If it works, it works.

Issue #3460, opened 2021-12-19
Replace fsouza/go-dockerclient with docker/docker/client


rfay:
[fsouza/go-dockerclient](https://github.com/fsouza/go-dockerclient) is barely maintained. We started using it (and it was created) before docker had a golang api.

Replace it with docker/client, see https://docs.docker.com/engine/api/sdk/

An immediate win for this would be the ability to support `ssh://` docker endpoints, which are not supported in go-dockerclient.

PR #5658, opened 2023-12-22
build: update all vendor module dependencies

stasadev:
## The Issue

I am working on 

- #3460

It would be nice to update all dependencies before that, so I can use all the latest libraries in combination with the new docker library.

It runs:
```
go get -u ./...
go mod vendor
go mod tidy
```


rfay (approved):
This looks fine. A lot of times I only try to update one or two things that dependabot is complaining about, but it makes sense to do them all from time to time. Good to go.

PR #5662, opened 2023-12-27
build: revert bump actions/upload-artifact from 3 to 4, for #5648

stasadev:
## The Issue

- #5648

Go to https://github.com/ddev/ddev/actions/runs/7340705862?pr=5661 and click on the artifact and see:

```
Failed to generate URL to download artifact.
```

There is no update guide at the moment, go back to `actions/upload-artifact@v3` for now.

See 

- https://github.com/actions/upload-artifact/issues/472

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -244,32 +244,32 @@ jobs:
         run: ./.ci-scripts/generate_artifacts.sh ${{ github.workspace }}/artifacts
 
       - name: Upload all artifacts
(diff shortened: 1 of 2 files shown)


stasadev:
I checked the build of the main branch, it works https://github.com/ddev/ddev/actions/runs/7340970014

Before merging #5648, I checked in a separate repository that the artifacts appeared on the Checks page and the bot comment looked legitimate, but I didn't click the link to check the artifact itself.

I will be more careful about such changes in the future.

PR #5663, opened 2023-12-28
docs: add instructions for Search API Pantheon module (for Drupal)

darrenoh:


stasadev (changes requested):
@darrenoh thank you for the contribution to the DDEV docs.

Please fill out the issue details when you have a chance.

File paths such as .crt and others should be included in the `.crt` backticks to make it look readable.

A link to the Pantheon documentation https://docs.pantheon.io/guides/solr-drupal is needed somewhere, and please add a note that this is for Drupal.

darrenoh:
All I am doing is summarizing what I shared on Slack. I have a lot of other things to do. Anyone who wants to take this further is welcome to.

rfay:
As far as I can tell this isn't something that belongs in the DDEV docs. I think it also is a very obscure case where you're using the upstream Pantheon service with local development, right?   It's great that you sorted it out, but it would be better in Stack Overflow I think. Could you post it there? Do it as a question-and-answer where you click the checkbox to answer your own question. Thanks!

PR #5668, opened 2024-01-01
fix: remove CRAFT_WEB_URL, fixes #5665

### LINKED ISSUE #5665: DDEV should not manage CRAFT_WEB_URL in .env (Craft CMS)

```
All project containers are now ready.
Successfully started tryddevproject-18527
Project can be reached at https://tryddevproject-18527.ddev.site https://127.0.0.1:32782
======== Curl of site from inside container:
HTTP/1.1 200 OK
Server: nginx
Date: Sat, 30 Dec 2023 19:21:59 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Vary: Accept-Encoding

======== curl -I of http://tryddevproject-18527.ddev.site from outside:
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
Date: Sat, 30 Dec 2023 19:21:59 GMT
Server: nginx
Vary: Accept-Encoding

======== full curl of http://tryddevproject-18527.ddev.site from outside:
Success accessing database... db via TCP/IP
ddev is working. You will want to delete this project with 'ddev delete -Oy tryddevproject-18527'
======== Project ownership on host:
drwxr-xr-x  4 timkelty  staff  128 Dec 30 14:21 ../tryddevproject-18527
======== Project ownership in container:
drwxr-xr-x 4 timkelty dialout 128 Dec 30 19:21 /var/www/html
======== In-container filesystem:
Filesystem     Type     1K-blocks      Used Available Use% Mounted on
mac            virtiofs 971350180 817160468 154189712  85% /var/www/html
======== curl again of tryddevproject-18527 from host:
Success accessing database... db via TCP/IP
ddev is working. You will want to delete this project with 'ddev delete -Oy tryddevproject-18527'
Thanks for running the diagnostic. It was successful.
Please provide the output of this script in a new gist at gist.github.com
Running ddev launch in 5 seconds
```
</details>         

### Expected Behavior

Setting `CRAFT_WEB_URL` to DDEV's primary URL is too prescriptive for Craft CMS and breaks multi-site functionality.

I should be able to `ddev start` a Craft CMS project, and not have the `@web` alias overwritten.

### Actual Behavior

On every start, `CRAFT_WEB_URL` is written to .env file, using the DDEV primary url, potentially breaking any Craft multisite installation.

### Steps To Reproduce

_No response_

### Anything else?

This was added as a result of https://github.com/ddev/ddev/issues/4650.

In that discussion, I was mistaken and thought these would only be added _initally_ (`ddev config`) as a starting point/suggestion, when in fact they are added every `start`.

For further clarification, `CRAFT_WEB_URL` is meant to be configured to the hostname from the current request. If you are using DDEV's `additional_hostnames`, this will not be accurate, so should not be set.

timkelty:
## The Issue
The `CRAFT_WEB_URL` env variable is not necessary for a Craft project, and prevents multi-site features from working.

Does not manage `CRAFT_WEB_URL`.

It also removes the comment about backward compatibility, as it isn't quite true. While `PRIMARY_SITE_URL` isn't an "official" var that Craft uses, it is a de-facto standard, as [Craft's installer script](https://github.com/craftcms/cms/blob/c905cc975a57ee3e41a317adf8cd7e303353a35b/src/console/controllers/InstallController.php#L189) will attempt to set it, and it is helpful to have the primary site set to reference in your config.

Start a craftcms ddev project:

```
ddev config --project-type=craftcms --docroot=web --create-docroot
touch .env
ddev start
```

Ensure `CRAFT_WEB_URL` was not written to `.env` (command should have no output):

```
cat .env | grep "CRAFT_WEB_URL="
```

## Related Issue Link(s)

Fixes https://github.com/ddev/ddev/issues/5665

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -134,11 +134,10 @@ func craftCmsPostStartAction(app *DdevApp) error {
 			"CRAFT_DB_DATABASE":     "db",
 			"CRAFT_DB_USER":         "db",
 			"CRAFT_DB_PASSWORD":     "db",
-			"CRAFT_WEB_URL":         app.GetPrimaryURL(),
 			"CRAFT_WEB_ROOT":        app.GetAbsDocroot(true),
 			"MAILPIT_SMTP_HOSTNAME": "127.0.0.1",
 			"MAILPIT_SMTP_PORT":     "1025",
-			"PRIMARY_SITE_URL":      app.GetPrimaryURL(), // for backward compatibility only
+			"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
 		}
 	}
 


rfay:
If CRAFT_WEB_URL isn't necessary for an ordinary project then I'm all in favor of not having it added to .env.

rfay:
Please add manual testing instructions (and do them) thanks!

timkelty:
@rfay added manual testing instructions

stasadev (approved):
This looks fine to me.

And I see that CraftCMS uses `PRIMARY_SITE_URL` since 2020 https://github.com/craftcms/cms/commit/7e18bc5f336181b40cfbdfea75ef60499e1c9a34

PR #5673, opened 2024-01-03
docs: add human readable headers tip in writing-style-guide.md

### LINKED ISSUE #5641: Add README guideline, for example to streamline add-on titles
[Add-on](https://ddev.readthedocs.io/en/latest/users/extend/additional-services/) headers in their individual repos are either missing, they can be all capitals, or "What is this?"

It would be great to streamline the style, by following some simple guidelines, like the [Drupal README template](https://www.drupal.org/docs/develop/managing-a-drupalorg-theme-module-or-distribution-project/documenting-your-project/readmemd-template).

As a start, for a title template, I'd like to nominate the format from "DDEV Adminer Service" https://github.com/ddev/ddev-adminer using this format:

> # DDEV Service Name

So for example for [ddev-drupal9-solr](https://github.com/ddev/ddev-drupal9-solr), the header should be:

> # DDEV Drupal9 Solr

## Proposed solution

A brief README guideline section could be added under https://ddev.readthedocs.io/en/latest/developers/writing-style-guide/

gitressa:
## The Issue

Fixes #5641.

Adds a guideline for README headers.

### DIFF
--- a/docs/content/developers/writing-style-guide.md
+++ b/docs/content/developers/writing-style-guide.md
@@ -242,6 +242,7 @@ One-off tips that don’t fit nicely into any of the sections above:
 <!-- textlint-disable -->
 
 - Pluralize “CMS” as “CMSes”, not “CMSs”.
+- Use human readable format in headers, for example in README files.
 - Use all-caps references for file *types* like JSON, YAML and CSS.
 - Wrap file *extensions* in backticks like `.json`, `.yaml`, and `.css`.
 - Wrap references to files, directories, images and commands in backticks.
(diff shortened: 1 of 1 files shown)


docs/content/developers/writing-style-guide.md
```diff
@@ -242,6 +242,7 @@ One-off tips that don’t fit nicely into any of the sections above:
 <!-- textlint-disable -->
 
 - Pluralize “CMS” as “CMSes”, not “CMSs”.
+- Use human readable format in headers, for example in README files.
```
stasadev, line 245: Isn't that the same as a few lines below?
```
- Use Title Case for headings wherever it makes sense.
```

stasadev:
Looking at it again, I think the "Writing Style Guide" only applies to documentation, and this change applies to add-ons. People who create add-ons are unlikely to look for guidance here.

I think this should be clarified in the "Getting Started" section on https://github.com/ddev/ddev-addon-template, in particular here:

https://github.com/ddev/ddev-addon-template/blob/e06657f57cddc35e02d0aa603a961c1f9365da28/README.md?plain=1#L31

Thank you for your attention to such details, I'm closing this, feel free to open a new PR there.

gitressa:
Yes, that's probably better. But don't you think the problem already happens in step 1?

> Choose a good descriptive name for your add-on. It should probably start with "ddev-" and include the basic service or functionality. If it's particular to a specific CMS, perhaps `ddev-<CMS>-servicename`.

I don't know enough about the add-on template system, but if you had time, it would be awesome if you could go through the process, and check that the outcome in the add-on README file will be `DDEV Project Name` and not `ddev-project-name`, to make sure this will be the README title format for future add-ons.

rfay:
The best place for add-on related things would be the template, I would think. https://github.com/ddev/ddev-addon-template

gitressa:
Thanks @rfay, I created a MR for that README: https://github.com/ddev/ddev-addon-template/pull/41.

PR #5674, opened 2024-01-03
docs: add WorkingDirectory= directive to systemd ddev.service example

vanWittlaer:
## The Issue

ddev.service fails at startup on Ubuntu 22.04 and > 1 ddev projects

Add the WorkingDirectory= directive to the ddev.service example


rfay:
This seems reasonable, but if you could please find out what directory it was using in the first place that would help. Use `pwd &&` to figure out what was going on. What directory was it launched in, and why does it work fine for me, see Discord thread, https://discordapp.com/channels/664580571770388500/1192036482068774962/1192183849338286100

rfay:
This workaround was only an accidental fix (it made the `chmod ugo+w .` work, but it shouldn't be allowed to work.

Actual fix is 
* #5676

Thanks so much for exploring this and exposing this bug!

PR #5676, opened 2024-01-04
fix: when settings files not set, don't chmod, fixes #5675

### LINKED ISSUE #5675: Project type `php` shouldn't try to  create settings files, should it?

```
```
</details>         

### Expected Behavior

PHP project type shouldn't try to create settings files or related activities

### Actual Behavior

https://github.com/ddev/ddev/blob/master/pkg/ddevapp/php.go#L7-L14

I don't know how or why this got in there but it doesn't make any sense. PHP project type doesn't have settings.

Discovered in Discord, https://discordapp.com/channels/664580571770388500/1192036482068774962

### Steps To Reproduce

See Discord thread.

### Anything else?

_No response_

rfay:
## The Issue

* #5675 

Projects with no settings may fail when starting outside homedir. They shouldn't even be chmodding in that case

* Remove the chmod action if no settings files are configured (as with php project type)
* Completely remove phpPostStartAction as it contained nothing useful. PHP projects never have settings.

Run `ddev start <somephpproject>` in the root directory of a linux machine. It should work fine.
Run `ddev start <somedrupalproject>` in the root directory on linux, it should work and unwriteable settings files should be  chmodded properly.

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -253,27 +253,29 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 	// Drupal and WordPress love to change settings files to be unwriteable.
 	// Chmod them to something we can work with in the event that they already
 	// exist.
-	chmodTargets := []string{filepath.Dir(app.SiteSettingsPath), app.SiteDdevSettingsFile}
-	for _, fp := range chmodTargets {
-		fileInfo, err := os.Stat(fp)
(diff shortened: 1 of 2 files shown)


stasadev (approved):
I tested both use cases and confirmed that they work.

The actual error message (if someone tries to search for it):

```
failed to write settings file : could not change permissions on file . to make it writeable: chmod .: operation not permitted
```

PR #5678, opened 2024-01-04
docs: fix composer command code block

mttsch:
## The Issue

Screenshot 2024-01-04 at 18 51 44

The `config` section of the commands page is not properly rendered.

Correctly close the previous code block.


stasadev (approved):
Nice catch 👍

rfay (approved):
Thanks!

PR #5679, opened 2024-01-04
docs: fix broken numeration in Pantheon

stasadev:
## The Issue

The list should not start again from 1:
image

Removes an extra line.


stasadev:
I didn't add `<!-- markdownlint-disable-next-line -->` to the new line, because that would break the ordered list.

rfay (approved):
Thanks!

PR #5686, opened 2024-01-09
fix: remove redundant chown in ddev-webserver start.sh, for #4942

rfay:
## The Issue

* #4942 

This at least removes the redundant chown in container start.sh


stasadev (approved):
I deleted the volume, started the project, and checked the directory owners. Everything is fine.

```
ddev poweroff
docker volume rm ddev-global-cache
ddev start
ddev exec "ls -l /mnt && ls -l /mnt/ddev-global-cache"
```

PR #5690, opened 2024-01-10
feat: create db query log command, fixes #5689

### LINKED ISSUE #5689: Add-on for easily turning on/off or tail-ing the db query log

When using ddev with a framework with a database abstraction layer, it can be finicky to set up being able to see the actual database queries generated by the application. 

### Describe your solution

A custom ddev command that allows a user to very very easily turn on or off query logging and optionally to tail that log. 

Due to the noisy nature of the db log output, it would be best to exclude it from the general `ddev logs` output.

- [X] I have contributed an add-on https://github.com/ChromaticHQ/ddev-dblog

### Describe alternatives

I considered adding configuration to the db container that would send all queries to the ddev logs. But I realized that would create a lot of noise, would slow down the application, and would make it difficult to toggle on and off. Looking at the db logs is generally something only necessary to do occasionally. 

### Additional context

I have written a command as described above and am using it successfully. I am posting this issue to see whether it would be considered generally worthwhile enough for it to be contributed to the ddev project as an included command.

apotek:
Create initial version of db query log command

This command should probably be expanded to be mysql or pgsql neutral (or a separate version written for pgsql parity).

Fixes: [#5869](https://github.com/ddev/ddev/issues/5689)

## The Issue
While using frameworks that use db abstraction layers, it can be difficult to see what queries are actually being executed on the database. While there are numerous ways to solve for this, this command contribution aims to make it dead simple for a developer to turn on/off query logging and tail it if needed.

We create a command that allows a user to (for example) tail the actual db queries with a simple `ddev dblog tail`

Run `ddev dblog on` and verify that queries are being logged inside the db container at /tmp/dblog <--May want to use standard /var/log/query.log or /home/mysql/query.log

Run `ddev dblog off` and verify that queries are no longer being logged inside the db container. 

Run `ddev dblog tail` and verify that you are now successfully tailing the queries in your terminal window. Hit CTRL+C to break out of it, and observe that the dblog state is reverted to its previous state, and a goodbye message is printed.

## Related Issue Link(s)
No related issues.

## Release/Deployment Notes
Do default commands get documented somewhere? If so, this merge request should probably also include documentation. 

This merge request will not change anything for users or for current installations unless a user chooses to run the contributed command.

This was inspired by the xdebug on|off command and a short [post by Dries Buytaert](https://dri.es/effortless-inspecting-of-drupal-database-queries).

### DIFF
--- a/pkg/ddevapp/dotddev_assets/commands/db/dblog
+++ b/pkg/ddevapp/dotddev_assets/commands/db/dblog
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+## #ddev-generated
+## Description: Turn on query logging for diagnostic purposes.
+## Usage: dblog on|off|tail
+## Example: "ddev dblog on" or "ddev dblog off" or "ddev dblog tail"
+## DBTypes: mysql,mariadb
+## ExecRaw: true
+
+case "$@" in
+  on)
+    state=1
+    ;;
+  off)
+    state=0
+    ;;
+  tail)
+    state=1
+    previous_state=$(mysql -u root -proot -e "SHOW VARIABLES LIKE 'general_log'\G;" | awk '/Value:/ { print $2 }')
+    trap ctrl_c INT
+    ;;
+  *)
+    echo "Tell us whether you want the query log on or off, or if you want to tail it."
+    echo "Usage: querylog on|off|tail"
+    exit 1
+    ;;
+esac
+
+function ctrl_c() {
+  if [ "$previous_state" = "OFF" ]; then
+    echo "Turning off general log..."
+    mysql -u root -proot -e "SET global general_log = 0"
+  fi
+  echo "Goodbye."
+  exit 0
+}
+
+mysql -u root -proot <<EOX
+SET global general_log = $state;
+SET global log_output = 'file';
+SET global general_log_file = '/tmp/dblog';
+EOX
+
+if [ "$@" = "tail" ]; then
+  tail -f /tmp/dblog
+fi


rfay:
Nice work. I do think this would probably be better as an add-on, as @tyler36 suggested. It's not often used, but very, very useful when it is. I've always rolled it when needed, but a maintained add-on would be better, and that makes it so you'd have control of the development cycle instead of being tied to DDEV's cycle.

apotek:
> Nice work. I do think this would probably be better as an add-on, as @tyler36 suggested. It's not often used, but very, very useful when it is. I've always rolled it when needed, but a maintained add-on would be better, and that makes it so you'd have control of the development cycle instead of being tied to DDEV's cycle.

Thank you for the response. I confess to some confusion about the add-on suggestion, as I always thought add-ons were additional services (such as an elastic search service, or a couchdb service) that you could plug into the app. Whereas here I am adding tooling for an _existing_ service, not adding a container.

Would you be able to point me to some clarifying docs on that? The docs I read on add-ons are oriented towards adding additional services rather than adding a command or feature to an existing service. Thank you.

apotek:
> > Nice work. I do think this would probably be better as an add-on, as @tyler36 suggested. It's not often used, but very, very useful when it is. I've always rolled it when needed, but a maintained add-on would be better, and that makes it so you'd have control of the development cycle instead of being tied to DDEV's cycle.
> 
> Thank you for the response. I confess to some confusion about the add-on suggestion, as I always thought add-ons were additional services (such as an elastic search service, or a couchdb service) that you could plug into the app. Whereas here I am adding tooling for an _existing_ service (db), not adding a container.
> 
> Would you be able to point me to some clarifying docs on that? The docs I read on add-ons are oriented towards adding additional services rather than adding a command or feature to an existing service. Thank you.

Never mind @rfay , I see @tyler36 pointed me to docs. I learned something new today! I'll see if I can re-roll this as an add-on.

tyler36:
I like this. I added it to my `global_commands/db` folder and it works well.

I'm curious, does this work for all MariaDB & mySQL versions?

The only other suggestion I have is Postgresql support.

apotek:
> I like this. I added it to my `global_commands/db` folder and it works well.
> 
> I'm curious, does this work for all MariaDB & mySQL versions?
> 
> The only other suggestion I have is Postgresql support.

It should work for all MariaDB and MySQL versions which use the following variables:

 - general_log
 - log_output
 - general_log_file

Which I don't have time to research but I think this goes way way back, maybe even to MySQL 3. 

@rfay I am closing this merge request in favor of the addon I just set up. Once it passes tests, I'll post a link to it in the original feature request issue. 

Thank you both so much for making it easy to get started. Hopefully I can add some value to ddev users soon.

tyler36:
It would be great if you can post the link to your addon here too. Makes it easier for search etc.

Drop into the Discord channel if you need help with the addon or want to say hi.  
Here's an invite if you are not there already: https://discord.gg/5wjP76mBJD

Thanks for contributing!

apotek:
> It would be great if you can post the link to your addon here too. Makes it easier for search etc.
> 
> Drop into the Discord channel if you need help with the addon or want to say hi. Here's an invite if you are not there already: https://discord.gg/5wjP76mBJD
> 
> Thanks for contributing!

@tyler36 : https://github.com/ChromaticHQ/ddev-dblog

PR #5695, opened 2024-01-15
fix: remove --json-output flag from `logs` command, fixes #962

### LINKED ISSUE #962: ddev logs --json-output (ddev logs -j) no longer works (just emits normal logs)

GuySartorelli:
## The Issue

`ddev logs` doesn't support JSON formatted output, and there's no clear way to add that functionality given the way the logs are currently fetched and outputted.
This is discussed in #962 

We don't display the flag when running `ddev help logs` or `ddev logs -h`.
This is the same way this is handled for the composer command (see [composer.go](https://github.com/ddev/ddev/blob/9b30cac3f0c7201da56611b5ed4cde6fead56de8/cmd/ddev/cmd/composer.go#L51-L56)) and custom commands (see [commands.go](https://github.com/ddev/ddev/blob/9b30cac3f0c7201da56611b5ed4cde6fead56de8/cmd/ddev/cmd/commands.go#L232-L238))

Run `ddev help logs` and `ddev logs -h` - the `--json-output` flag shouldn't be shown as an option.

## Automated Testing Overview

No automated tests added - I couldn't find associated tests for composer or for custom commands so it didn't seem sensible to add them for this.

## Related Issue Link(s)

https://github.com/ddev/ddev/issues/962

## Release/Deployment Notes

Doesn't affect deployments.

### DIFF
--- a/cmd/ddev/cmd/logs.go
+++ b/cmd/ddev/cmd/logs.go
@@ -43,6 +43,21 @@ func init() {
 	DdevLogsCmd.Flags().BoolVarP(&timestamp, "time", "t", false, "Add timestamps to logs")
 	DdevLogsCmd.Flags().StringVarP(&serviceType, "service", "s", "web", "Defines the service to retrieve logs from. [e.g. web, db]")
 	DdevLogsCmd.Flags().StringVarP(&tail, "tail", "", "", "How many lines to show")
-	RootCmd.AddCommand(DdevLogsCmd)
 
+	// JSON formatted output isn't supported, so we should hide the global --json-output flag
+	// We have to do that within the help function, since the global flag hasn't been added yet during init()
+	originalHelpFunc := DdevLogsCmd.HelpFunc()
+	DdevLogsCmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
+		_ = command.Flags().MarkHidden("json-output")
+		originalHelpFunc(command, strings)
+	})
+	// We need to do the same with the flag error function in case flags can't be parsed correctly,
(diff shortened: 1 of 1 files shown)


stasadev (commented):
> Run `ddev help logs` and `ddev logs -h` - the `--json-output` flag shouldn't be shown as an option.

Tried these, the flag is hidden as described.

And I found another edge case, that doesn't really matter: `--json-output` is still there when you pass a wrong flag to `ddev logs` like this `ddev logs --foobar`.

GuySartorelli:
> And I found another edge case, that doesn't really matter: `--json-output` is still there when you pass a wrong flag to `ddev logs` like this `ddev logs --foobar`.

Oh! That's unexpected... I wonder where that's getting its output from. I presume that same thing is happening for custom commands as well then (presumably not for composer - I bet that's handled differently).

I'll spend some time trying to figure out where that's coming from.

GuySartorelli:
Turns out that's an entirely separate function called `FlagErrorFunc()` which gives identical output to the help function, but generated separately. So we just have to treat that function the same way as the help function.

This doesn't need to be done for composer or for custom commands, because flag parsing isn't handled by cobra for those (everything is just passed straight through, it seems)

stasadev (approved):
Looks good to me, a great study of [Cobra](https://github.com/spf13/cobra)'s capabilities.

Issue #5705, opened 2024-01-17
‹ddev poweroff› on more than one project sometimes leads to "Failed to stop project: Could not kill running container <traefik-router ID>" 


jonaseberle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
======= Existing project config =========
These config files were loaded for project a: [/home/jonas/tmp/a/.ddev/config.yaml]
name: a
type: php
php_version: 8.1
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.22.6
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.4}
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 18
default_container_timeout: 120
======= Creating dummy project named  tryddevproject-1217 in ../tryddevproject-1217 =========
OS Information: Linux e480 6.5.0-14-generic #14-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 14 14:59:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
User information: uid=1000(jonas) gid=1000(jonas) groups=1000(jonas),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),122(lpadmin),132(lxd),133(sambashare),137(docker)
DDEV version:  ITEM             VALUE                                   
 DDEV version     v1.22.6-26-g94d77509a                   
 architecture     amd64                                   
 db               ddev/ddev-dbserver-mariadb-10.4:v1.22.6 
 ddev-ssh-agent   ddev/ddev-ssh-agent:v1.22.6             
 docker           24.0.5                                  
 docker-compose   v2.23.3                                 
 docker-platform  linux-docker                            
 mutagen          0.17.2                                  
 os               linux                                   
 router           ddev/ddev-traefik-router:v1.22.6        
 web              ddev/ddev-webserver:v1.22.6             
PROXY settings: HTTP_PROXY='' HTTPS_PROXY='' http_proxy='' NO_PROXY=''
======= DDEV global info =========
Global configuration:
instrumentation-opt-in=true
omit-containers=[]
performance-mode=none
router-bind-all-interfaces=false
internet-detection-timeout-ms=3000
disable-http2=false
use-letsencrypt=false
letsencrypt-email=
table-style=default
simple-formatting=false
use-hardened-images=false
[...]

PR #5707, opened 2024-01-17
fix: use init option to prevent zombie ddev-router, for #5705

stasadev:
## The Issue

- #5705

> API error (500): Could not kill running container d8079abded3b448b54fc3d22bc0abfe10ffbeb69cd7066027655bbf5598992d5, cannot remove - container d8079abded3b PID 1968501 is zombie and can not be killed. Use the --init option when creating containers to run an init inside the container that forwards signals and reaps processes

The error suggests to use `init`, it doesn't harm to try it for `ddev-router`:

> Use the --init option when creating containers to run an init inside the container that forwards signals and reaps processes

I cannot reproduce this error, at least it should not break the tests. See the issue.

## Related Issue Link(s)

- https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter

### DIFF
--- a/pkg/ddevapp/router_compose_template.yaml
+++ b/pkg/ddevapp/router_compose_template.yaml
@@ -1,10 +1,10 @@
 services:
   ddev-router:
     image: {{ .router_image }}
-    # Prevent zombie container
-    init: true
 
     {{ if eq .Router "traefik" }}
+    # Prevent zombie container
+    init: true
     command:
       - --configFile=/mnt/ddev-global-cache/traefik/static_config.yaml
     user: {{ .UID }}:{{ .GID }}
--- a/pkg/ddevapp/router_compose_template.yaml
+++ b/pkg/ddevapp/router_compose_template.yaml
@@ -1,6 +1,8 @@
 services:
   ddev-router:
     image: {{ .router_image }}
+    # Prevent zombie container
+    init: true
 
     {{ if eq .Router "traefik" }}
     command:


jonaseberle (approved):

jonaseberle:
That fixes it for me. Thank you Stanislav!

stasadev:
Great! This will make things a little easier.

rfay:
The SO article implies that this injects a `tini` process as parent of the traefik process. It seems like that could possibly have other implications. `init` is a strange name for them to give that behavior.

stasadev:
Yeah, I don't really know what this change can bring and how safe it is.

rfay (changes requested):
I think this is fine. However, since we have only had the problem with the traefik router, why don't we just add the init option if we're running traefik? (Just move the init down a line.) 

Apparently there is a path where the traefik binary doesn't handle signals correctly, and maybe that's what's happening. 

With this, we see /sbin/docker-init injected as pid 1 in the container:

```
rfay@rfay-surface-goo:~/workspace/d10$ docker exec -it ddev-router bash
f70c5494795b:/mnt/ddev-global-cache/traefik$ ps -ef
PID   USER     TIME  COMMAND
    1 1000      0:00 /sbin/docker-init -- /entrypoint.sh --configFile=/mnt/ddev-global-cache/traefik/static_config.yaml
    7 1000      0:00 traefik traefik --configFile=/mnt/ddev-
```

docker-init is just `tini1 apparently packaged for docker's use, not sure why they use a different name:

```
f70c5494795b:/mnt/ddev-global-cache/traefik$ /sbin/docker-init -h
docker-init (tini version 0.19.0 - git.de40ad0)
Usage: docker-init [OPTIONS] PROGRAM -- [ARGS] | --version

Execute a program under the supervision of a valid init process (docker-init)

Command line options:

  --version: Show version and exit.
  -h: Show this help message and exit.
  -s: Register as a process subreaper (requires Linux >= 3.4).
  -p SIGNAL: Trigger SIGNAL when parent dies, e.g. "-p SIGKILL".
  -v: Generate more verbose output. Repeat up to 3 times.
  -w: Print a warning when processes are getting reaped.
  -g: Send signals to the child's process group.
  -e EXIT_CODE: Remap EXIT_CODE (from 0 to 255) to 0.
  -l: Show license and exit.

Environment variables:

  TINI_SUBREAPER: Register as a process subreaper (requires Linux >= 3.4).
  TINI_VERBOSITY: Set the verbosity level (default: 1).
  TINI_KILL_PROCESS_GROUP: Send signals to the child's process group.
  ```

I'm surprised not to see issues in the traefik queue about zombies.

File: pkg/ddevapp/router_compose_template.yaml [hunk 1]
```diff
@@ -1,6 +1,8 @@
 services:
   ddev-router:
     image: {{ .router_image }}
+    # Prevent zombie container
+    init: true
 
     {{ if eq .Router "traefik" }}
```
rfay, line 7:
```suggestion

    {{ if eq .Router "traefik" }}
    # Prevent zombie container
    init: true
```

stasadev:
Great, thank you. I didn't know that this didn't happen with nginx-proxy, so it makes sense to only use it with Traefik.

I've restarted the tests to make sure everything is fine.

PR #5716, opened 2024-01-19
docs: temporarily revert docs on not needing --create-docroot

rfay:
## The Issue

People have been stumbling over the fact that `--create-docroot` is no longer in the quickstarts.. because it's removed in latest dev versions. But it's required in latest stable.

Temporarily revert the changes to the quickstarts and re-add `-- create-docroot`

This PR should be reverted before the next release.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -35,7 +35,7 @@ Environment variables will be automatically added to your `.env` file to simplif
     cd my-craft-project
 
     # Set up the DDEV environment:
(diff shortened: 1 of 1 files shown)


stasadev (changes requested):
One more for Ibexa DXP:

https://github.com/ddev/ddev/blob/94d77509a582d92b5d975be97e84c66e0ad0a6de/docs/content/users/quickstart.md?plain=1#L201-L201

rfay:
Thanks, fixed that one.

stasadev (approved):

PR #5732, opened 2024-01-24
feat: add autocomplete for project name arguments, fixes #4880

### LINKED ISSUE #4880: ADD autocomplete project names #completion

there are lots of projects in `ddev list` with long names and common prefixes on my pc

### Describe your solution

I type `ddev command` and type a letter and hit Tab
I get a list of project names that match

### Describe alternatives

using the mouse -- oh the humanity!

### Additional context

not to be confused with `ddev completion`, which works well

GuySartorelli:
## The Issue

Lots of the go commands take a project name as an argument, but don't provide autocompletion for those arguments.

Adds autocompletion for commands that take a project name as an argument.

- If only one argument is allowed, it doesn't give completion for more.
- If multiple arguments are allowed, it gives completion for all of them.
- For `start` and `stop` specifically, if only gives names of the relevant (inactive and active respectively) projects - for all other commands, all names are given.

**NOTE** This makes the `--select` flag for `ddev start` and `ddev stop` redundant. I've marked the flag as hidden and deprecated for both of those commands - it can be removed in some future release.

I've also removed the docs about that flag - but if you want me to add it back with a note that they're deprecated, I can happily do that.

Type `ddev start <tab>` and see the autocompletion results.
- Try this with any command that takes a project name as an argument
- Check that commands that only accept one argument only perform completion for one argument
- Check that commands that accept multiple arguments perform completion for as many arguments as you like
- Check that `ddev start` only gives completion for currently inactive projects
- Check that `ddev stop` only gives completion for currently active projects
- Check that a project name isn't repeated if it's already included as an argument

## Automated Testing Overview

I've added tests that work based on the `TestSites` array - though since these tests won't run with less than 2 test sites, the tests are skipped if there's not enough tests.

## Related Issue Link(s)

- https://github.com/ddev/ddev/issues/4880

## Release/Deployment Notes

This is fully backwards compatible and shouldn't affect anything else.

### DIFF
--- a/cmd/ddev/cmd/clean.go
+++ b/cmd/ddev/cmd/clean.go
@@ -14,8 +14,9 @@ import (
 )
 
 var CleanCmd = &cobra.Command{
-	Use:   "clean [projectname ...]",
-	Short: "Removes items DDEV has created",
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 0),
+	Use:               "clean [projectname ...]",
+	Short:             "Removes items DDEV has created",
 	Long: `Stops all running projects and then removes downloads and snapshots
 for the selected projects. Then clean will remove "ddev/ddev-*" images.
 
--- a/cmd/ddev/cmd/debug-compose-config.go
+++ b/cmd/ddev/cmd/debug-compose-config.go
@@ -1,8 +1,9 @@
 package cmd
 
 import (
-	"github.com/ddev/ddev/pkg/fileutil"
 	"strings"
+
+	"github.com/ddev/ddev/pkg/fileutil"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/output"
@@ -12,8 +13,9 @@ import (
 
 // DebugComposeConfigCmd implements the ddev debug compose-config command
 var DebugComposeConfigCmd = &cobra.Command{
-	Use:   "compose-config [project]",
-	Short: "Prints the docker-compose configuration of the current project",
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
+	Use:               "compose-config [project]",
+	Short:             "Prints the docker-compose configuration of the current project",
 	Run: func(cmd *cobra.Command, args []string) {
 		projectName := ""
 
--- a/cmd/ddev/cmd/logs.go
+++ b/cmd/ddev/cmd/logs.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
 )
@@ -13,9 +14,10 @@ var (
 
 // DdevLogsCmd contains the "ddev logs" command
 var DdevLogsCmd = &cobra.Command{
-	Use:   "logs [projectname]",
-	Short: "Get the logs from your running services.",
-	Long:  `Uses 'docker logs' to display stdout from the running services.`,
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("active", 1),
+	Use:               "logs [projectname]",
+	Short:             "Get the logs from your running services.",
+	Long:              `Uses 'docker logs' to display stdout from the running services.`,
 	Example: `ddev logs
 ddev logs -f
 ddev logs -s db
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -15,9 +15,10 @@ var sshDirArg string
 
 // DdevSSHCmd represents the SSH command.
 var DdevSSHCmd = &cobra.Command{
-	Use:   "ssh [projectname]",
-	Short: "Starts a shell session in the container for a service. Uses web service by default.",
-	Long:  `Starts a shell session in the container for a service. Uses web service by default. To start a shell session for another service, run "ddev ssh --service <service>`,
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("active", 1),
+	Use:               "ssh [projectname]",
+	Short:             "Starts a shell session in the container for a service. Uses web service by default.",
+	Long:              `Starts a shell session in the container for a service. Uses web service by default. To start a shell session for another service, run "ddev ssh --service <service>`,
 	Example: `ddev ssh
 ddev ssh -s db
 ddev ssh <projectname>
--- a/pkg/ddevapp/utils.go
+++ b/pkg/ddevapp/utils.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 	"text/template"
@@ -19,6 +20,7 @@ import (
 	"github.com/docker/docker/api/types/versions"
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/spf13/cobra"
 )
 
 // GetActiveProjects returns an array of DDEV projects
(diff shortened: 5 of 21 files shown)


GuySartorelli:
Whitespace changes and imports moving around were done automatically by my IDE

pkg/ddevapp/utils.go
```diff
@@ -400,6 +401,44 @@ func ExtractProjectNames(apps []*DdevApp) []string {
...
+			// This is an error state - but we just return nothing
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		// Return nothing if we have nothing, or return all of the project names.
+		// Note that if there's nothing to return, we don't let cobra pick completions
+		// from the files in the cwd.
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		return ExtractProjectNames(apps), cobra.ShellCompDirectiveNoFileComp
```
stasadev, line 438:
```suggestion
		// Do not show arguments that are already written on the command line
		var appNames []string
		appNamesFromCmd := make(map[string]struct{})
		for _, name := range cmd.Flags().Args() {
			appNamesFromCmd[name] = struct{}{}
		}
		for _, name := range ExtractProjectNames(apps) {
			if _, exists := appNamesFromCmd[name]; !exists {
				appNames = append(appNames, name)
			}
		}
		return appNames, cobra.ShellCompDirectiveNoFileComp
```

stasadev (changes requested):
Wow, this is HUGE improvement 🎉

> This makes the `--select` flag for `ddev start` and `ddev stop` redundant

We can mark that flag as deprecated and hidden, and remove it later.

> should I create a new `command_completion_test.go` and do a few tests in there?

I think a new file with tests is suitable for this.

Thank you, @GuySartorelli 👍

pkg/ddevapp/utils.go
stasadev: The suggestion is a fix for:

- `ddev start project-name <TAB>` and `project-name` is still in the list.
- one project is running and you run `ddev stop <TAB><TAB><TAB>`, the result is `ddev stop project-name project-name project-name`.
- and so on.

GuySartorelli: Ooh good catch.

GuySartorelli: I did something a little different than the suggestion, but the result is the same.
- Called the variable `projectNames` instead of `appNames` since that's the terminology used in the rest of the function
- Used the `args` argument that cobra is giving us instead of fetching the args manually.

stasadev: Thanks, that looks better 👍, because I just googled how to do `array_intersect()` in go and used the first option I found.

GuySartorelli:
I've marked the `--select` flag for `ddev start` and `ddev stop` as hidden and deprecated.

I've also made a start on tests (which will currently fail) - I need to see if there's a tidy way to either:
- a) only create a few specific sites so I can rely on specific site names
- b) write the tests in a way that generically uses names of whatever test sites happen to be spun up

I ran out of time to do that tonight. Any pointers would be welcome.
Marking as draft for now until I get them sorted.

stasadev:
> a) only create a few specific sites so I can rely on specific site names

My opinion: the only requirement is that the number of sites should be sufficient to test all possible cases.

GuySartorelli:
I've added tests now. They work using the existing `TestSites` array because I ran into weird problems when I tried to define my own. The only caveat there is that the tests will be skipped if there's less than 2 tests in the array (which if I'm not mistaken will only happen if someone is testing locally and uses the `GOTEST_SHORT` env variable).

I also removed the docs about the deprecated/hidden `--select` flag but if you want to keep that doc around I can bring it back with a note about it being deprecated. Makes more sense to me to just not mention it in the docs though.

I've also updated the PR description.

stasadev (approved):
Wow, thank you so much for this!

LGTM, tested and checked the list of changed commands.

rfay:
You are completely amazing @GuySartorelli . I won't have time to adequately grok this for at least a few days, but so appreciative! This will be a much-loved feature.

rfay:
Rebased, reviewing.

@GuySartorelli please see if you can convince your IDE to use standard gofmt formatting, which is what's generally required. As you point out, it makes a mess and is hard to review when things get changed that aren't relevant to the PR. Happy to chat with you about this, I know the timezones are hard. But it would be cool if you could experiment a bit to see if you can calm it down a bit. What IDE are you using? How do you have it set up? We can discuss in another issue if that's the easiest.

Thanks so much for your incredible contributions. You know we're looking forward to more. I'm going to learn a lot from you.

cmd/ddev/cmd/debug-compose-config.go
```diff
@@ -12,8 +13,9 @@ import (
 
 // DebugComposeConfigCmd implements the ddev debug compose-config command
 var DebugComposeConfigCmd = &cobra.Command{
-	Use:   "compose-config [project]",
-	Short: "Prints the docker-compose configuration of the current project",
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
```
rfay, line 16: I imagine you already explored to see if this could be handled somewhere other than each Cobra Cmd.

rfay (changes requested):
This is fantastic and creative work.

Right now, I can't get it to work at all. When I click `ddev start <tab>` I just get a list of filenames in current directory. This is with bash 3, macOS.  What do I need to do to manually test it?

cmd/ddev/cmd/logs.go
```diff
@@ -13,9 +14,10 @@ var (
 
 // DdevLogsCmd contains the "ddev logs" command
 var DdevLogsCmd = &cobra.Command{
-	Use:   "logs [projectname]",
-	Short: "Get the logs from your running services.",
-	Long:  `Uses 'docker logs' to display stdout from the running services.`,
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
```
rfay, line 17: Should only allow running projects I think.

rfay (changes requested):
This is looking great and ready to go. I did find a couple of things that should only show running projects, please take a casual look and see if there are others. thanks for the amazing work!

GuySartorelli:
> please see if you can convince your IDE to use standard gofmt formatting, which is what's generally required.

If you can point me to something that explains what that specifically means, I'll be happy to

cmd/ddev/cmd/clean.go
```diff
@@ -14,8 +14,9 @@ import (
 )
 
 var CleanCmd = &cobra.Command{
-	Use:   "clean [projectname ...]",
-	Short: "Removes items DDEV has created",
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 0),
```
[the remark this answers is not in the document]
GuySartorelli: Oooh I'll look further into that (separately), that seems useful.

rfay:
Github discussion is great. Some vscode plugins use rather exotic and opinionated golang setup, whereas we're trying mostly to just use what `gofmt` says to do, so `gofmt -w -s` is what's expected everywhere. Your IDE should be able to do that. That said, I didn't really see all that much to worry about in this PR.

cmd/ddev/cmd/ssh.go
```diff
@@ -15,9 +15,10 @@ var sshDirArg string
 
 // DdevSSHCmd represents the SSH command.
 var DdevSSHCmd = &cobra.Command{
-	Use:   "ssh [projectname]",
-	Short: "Starts a shell session in the container for a service. Uses web service by default.",
-	Long:  `Starts a shell session in the container for a service. Uses web service by default. To start a shell session for another service, run "ddev ssh --service <service>`,
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
```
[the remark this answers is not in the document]
GuySartorelli: Fixed

cmd/ddev/cmd/logs.go
GuySartorelli: Fixed

rfay:
This is working wonderfully for me in testing just now.

PR #5733, opened 2024-01-24
docs: Update Drupal 9 & 10 CMS Quickstarts

Nick-Hope:
## The Issue

`ddev drush uli` generates a one-time link to the page for editing the uid1 (admin) account details. If the site is opened with that, there is no need to also run `ddev launch`, so that is confusing.

Add instructions on how to open the link generated by `ddev drush uli`. Remove `ddev launch` from the initial instructions but refer to its use for further site visits.

An alternative would be to drop the `ddev drush uli` line, but then the user would not be logged in when `ddev launch` is run.

### DIFF
--- a/docs/content/users/debugging-profiling/xhprof-profiling.md
+++ b/docs/content/users/debugging-profiling/xhprof-profiling.md
@@ -25,7 +25,7 @@ Another example: you could exclude memory profiling so there are fewer columns t
 ## Information Links
 
 * [php.net xhprof](https://www.php.net/manual/en/book.xhprof.php)
-* [Old facebook xhprof docs](http://web.archive.org/web/20110514095512/http://mirror.facebook.net/facebook/xhprof/doc.html)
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Thank you, I think this is a reasonable improvement 👍

tyler36:
A side note:

All the other quickstart guides result in the launch of website.
This change would result in a link displayed in the terminal.

For consistency, I would prefer to see the quickstart guide end with `ddev launch`. Followed by a comment below that shares the `ddev drush uli` tip.

stasadev (changes requested):
@tyler36 I agree with standardization.

@Nick-Hope, maybe it can be written like this:

```
# use the one-time link (CTRL/CMD + Click) from the command below to edit your admin account details.
ddev drush uli
ddev launch
```

Nick-Hope:
A number of approaches here would be acceptable. Decided the @stasadev suggestion was as good as any.

stasadev (approved):

PR #5737, opened 2024-01-25
docs: added sql:sanitize to Drupal hook examples.

karenmurky:
## The Issue
It would be nice to encourage database sanitizing when importing a database to a local environment. 

Updated examples to include sql:sanitize to post-import-db hooks for Drupal 10 installations.
Changed example heading from Drupal 8 to Drupal 10 to demonstrate this is the current Drupal version.
Also updated commands to use their full name instead of the alias, for better visibility.

### DIFF
--- a/docs/content/users/configuration/hooks.md
+++ b/docs/content/users/configuration/hooks.md
@@ -43,13 +43,13 @@ DDEV currently supports these tasks:
 
 Value: string providing the command to run. Commands requiring user interaction are not supported. You can also add a “service” key to the command, specifying to run it on the `db` container or any other container you use.
 
-Example: _Use Drush to clear the Drupal cache and get a user login link after database import_.
+Example: _Use Drush to rebuild all caches and get a user login link after database import_.
 
 ```yaml
 hooks:
   post-import-db:
-    - exec: drush cr
-    - exec: drush uli
+    - exec: drush cache:rebuild
+    - exec: drush user:login
 ```
 
 Example: _Use wp-cli to replace the production URL with development URL in a WordPress project’s database_.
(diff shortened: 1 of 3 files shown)


karenmurky:
Sorry, last commit message should say `Use drupal10Hooks text for Drupal 9 install`

pkg/ddevapp/drupal.go
```diff
@@ -256,6 +256,13 @@ func getDrupalUploadDirs(_ *DdevApp) []string {
 	return uploadDirs
 }
 
+// Drupal10Hooks adds a d10-specific hooks example for post-import-db
+const Drupal10Hooks = `# post-import-db:
+#		- exec: drush sql:sanitize
+#		- exec: drush updatedb
+#		- exec: drush cache:rebuild
```
stasadev, line 263: Tabs to spaces:

```suggestion
#   - exec: drush sql:sanitize
#   - exec: drush updatedb
#   - exec: drush cache:rebuild
```

docs/content/users/configuration/hooks.md
```diff
@@ -143,24 +151,22 @@ hooks:
     - exec: "drush cc all"
 ```
 
-## Drupal 8 Example
+## Drupal 10 Example
```
stasadev, line 154: I'm not sure, but isn't "Drupal 8 Example" still relevant for Drupal 8, and "Drupal 10 Example" might have its own section?

stasadev (changes requested):
Thank you!

> Sorry, last commit message should say `Use drupal10Hooks text for Drupal 9 install`

Don't worry about it, we use GitHub squash for PRs, it will be one commit (PR title) in the end.

docs/content/users/configuration/hooks.md
karenmurky: There was command differences between Drupal 7 and Drupal 8 (eg: drush cc all to drush cr) but D8 onwards the commands are the same.

Drupal.org doesn't list Drupal 8 or 9 as current versions (D8 was end of life in 2021 and D9 was November 2023)
so it was just to keep with consistency with the Project.

stasadev: Okay, that sounds good.

stasadev (approved):

PR #2679, opened 2020-12-03
Create new HeidiSQL command for Windows and WSL2

andreashager:
## The Problem/Issue/Bug:

Create a command for Windows/WSL2 that connects with the current project database (like `ddev sequelpro`).
Otherwise you always have to hardcode the database port and connect with database tool manually.

The command checks if i am on Windows directly or using DDEV with WSL2. The program used for this is HeidiSQL (https://www.heidisql.com/).

## How this PR Solves The Problem:

This PR uses the Windows database tool HeidiSQL to connection to the database automatically from the command line.

PR #5738, opened 2024-01-25
feat: add WSL2 support to tableplus command

plebward:
## The Issue
Add WSL2 support to the ddev tableplus command
Looks like TablePlus have put out a release since the blocking issue was encountered here - https://github.com/ddev/ddev/pull/2679#issuecomment-740371838

Adds a case that launches the Windows version of TablePlus, taking advantage of the WSL2 OS identifier

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/tableplus
@@ -2,12 +2,12 @@
 
 #ddev-generated
 # Support for TablePlus, https://tableplus.com/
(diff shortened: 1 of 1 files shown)


stasadev (approved):
Thank you!

Tested TablePlus on WSL2, and it works.

Would you like to extend it for Windows as well (like in [HeidiSQL](https://github.com/ddev/ddev/blob/a066c4259b8bc2e0baace172bc00dfc8329c53a2/pkg/ddevapp/global_dotddev_assets/commands/host/heidisql#L17))?

PR #5745, opened 2024-01-26
build: bump docker-compose to v2.24.3

stasadev:
## The Issue

It's time to update `docker-compose`.

The most significant change in [v2.24.0](https://github.com/docker/compose/releases/tag/v2.24.0):

Breaking change ⚠️
service hash computation logic has been updated to fully ignore replicas/scale. Due to this change, after upgrade all services will be recreated.

### DIFF
--- a/pkg/versionconstants/versionconstants.go
+++ b/pkg/versionconstants/versionconstants.go
@@ -43,4 +43,4 @@ var MutagenVersion = ""
 
 const RequiredMutagenVersion = "0.17.2"
 
-const RequiredDockerComposeVersionDefault = "v2.23.3"
+const RequiredDockerComposeVersionDefault = "v2.24.3"


stasadev:
The new `docker-compose` has new aggressive checks, tests are failing:

```
can't set healthcheck.start_interval as feature require Docker Engine v25 or later
```

Related issue:

- https://github.com/docker-library/docker/issues/473

rfay:
Pretty weird! Nice job chasing it. But it needs to be fixed somewhere or we're stuck with old compose. Seems like compose bug.

stasadev:
Yes, I will see what can be changed.

pkg/ddevapp/config.go
```diff
@@ -747,6 +747,7 @@ type composeYAMLVars struct {
 	Hostnames                       []string
 	Timezone                        string
 	ComposerVersion                 string
+	DockerVersion                   string
```
rfay, line 750: Let's use DockerEngineVersion here

rfay (changes requested):
Thanks for working on this, I'm probably going to be OK with it, but it does look fragile for the future. I'd sure rather it be fixed in docker-compose. Interested in doing a PR there?

pkg/ddevapp/config.go
stasadev: I decided to use a different technique: introduce a new custom function `templateCanUse(feature string)` for use in templates, which can perform more complex checks on the go side.

stasadev:
> I'd sure rather it be fixed in docker-compose.

I think they made a reasonable change with error, because `healthcheck.start_interval` didn't work before Docker v25 and should be considered an invalid option.

> it does look fragile for the future

Agree, and for this reason I added a check specifically for the Docker API version https://docs.docker.com/engine/api/#api-version-matrix
This API version (e.g. 1.44) is much more stable than the Docker version (e.g. 25.0.1). The same comparison is used inside `docker-compose`.

stasadev:
Looking deeper into the code base, there is a check for the minimum version of Docker to use with DDEV:

https://github.com/ddev/ddev/blob/2defedb31dd103d945fac90f260909796a0a5dc0/pkg/dockerutil/dockerutils.go#L1733-L1739

And it does some exotic things in comparison, like removing `_ce` for OpenSUSE (they add this suffix to the version):

https://github.com/ddev/ddev/blob/2defedb31dd103d945fac90f260909796a0a5dc0/pkg/dockerutil/dockerutils.go#L772-L783

From what I can see, people who build Docker can override the Docker version as they wish, for example from the Arch package (people can set any `VERSION` env they want when building):

https://gitlab.archlinux.org/archlinux/packaging/packages/docker/-/blob/381210628f4f895b2cee908ffa1179f7ff1150d4/PKGBUILD#L87-94

I would rather set the minimum [API version](https://docs.docker.com/engine/api/#api-version-matrix) to use with DDEV - **this value is set and controlled only by Docker**.

@rfay, what do you think about this? (Of course, this is beyond the scope of this PR, and can be done in a new PR).

rfay:
API version sounds like a fine thing to me.

In general, we've been trying to support as much as we can all the time, and been pretty successful with it over the long term.

stasadev:
I've updated the PR description with more information about the changes made, it should be fine to merge.

rfay:
I'll try to manually test. Looks good to me also.

rfay:
I guess my major worry would be that since healthcheck.start_interval was previously ignored, it *not* being ignored when using Docker v25 might cause some new issues.

rfay (approved):
It looks great to me. I tested against Docker 25.0.2 (Docker Desktop) and against Docker  24.0.7 (colima) and the behavior was correct. 

The reason I was a little hesitant was that I have misunderstood these configuration values and misused them many times during DDEV's history :)

Issue #4416, opened 2022-11-30
Allow global shell commands/custom commands to be run from everywhere (not just inside a project)


OlivierSC:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

As it's possible to start a specific project from everywhere on host with :
From my home folder ~/
`ddev start projectname`
-> projectname starts

Currently, custom commands seems to need to already be in a project folder to run.
If run from outside existing ddev project folder, global commands (from ~/.ddev/commands/host ) are not found : 
> unknown command "phpstorm" for "ddev"

it could be very useful to run command (ex with "cd" on linux/mac to go to folder) on the specified host project folder from everywhere.
example : 
I'm in my home folder ~/
`ddev cd projectname`
-> open the projectname folder path.

It could be also usefull for existing command  :
I'm in my home folder ~/
`ddev phpstorme projectname`
-> open the projectname project in phpstorm.

This could be very useful when working with lots of projects rather than manually change folder / open project.

### Describe your solution

The first step would be to allow commands to be run from everywhere, even outside from project folders (ex my "home" folder).

The second step would be to find the best way to provide the "project name" parameter to the command, even for existing command like phpstorm.

A bonus step would be that ddev could try to find local commands ([project folder/.ddev/commands) for the specified "project name", that way not all commands need to be "globals".

chrisfromredfin:
There's also the idea that some host commands might not even need a project, for example the new self-upgrade. It would be nice if that (or ones like it) could run without a project context, and without needing to start one project first.

rfay:
I've been thinking about this a bit and it has potential.

1. Instead of copying global commands into project .global_commands in-container commands should be copied into /mnt/ddev-global-config
2. host global commands can be used in place, and should be available from anywhere.
3. Commands themselves will have to determine whether they require project context or not, and maybe not show. I think perhaps an annotation for that would be an elegant way to do it.

GuySartorelli:
I'm keen for this to be implemented. @rfay is https://github.com/ddev/ddev/issues/4416#issuecomment-1838665202 a full set of acceptance criteria for this feature?
If not, what other criteria would you like for this?

I'd be keen to help get this implemented.

rfay:
It's good enough for a first draft, we may encounter complexities.

GuySartorelli:
I'm going to tackle this in a series (probably just two) of PRs, since there are a nice set of distinct pieces of work this can be split into.

[...]

PR #5753, opened 2024-01-30
feat: copy commands into volume, not into project, for #4416

GuySartorelli:
## The Issue
As part of improving global commands, Randy suggested:
> Instead of copying global commands into project .global_commands in-container commands should be copied into /mnt/ddev-global-config

Copies global commands (except host commands) into `/mnt/ddev-global-cache/commands/<service>/` directory for each service.

Since these are global commands, we don't need to have a separate directory for each project. This keeps global commands in sync across projects, where previously those commands could be out of sync for a given project if you change global commands while a project is running and *don't* subsequently run `ddev start` or `ddev debug fix-commands` for that project.

Now when the global commands are copied into the volume (which still happens on both `ddev start` and `ddev debug fix-commands`), since the directory in the volume is shared for all projects, it updates them for *all* projects.

Run any custom service command in a project.

## Automated Testing Overview

- Updated existing tests for the new behaviour

## Related Issue Link(s)

- #4416

## Release/Deployment Notes

Custom containers no longer need to need to add the volume mentioned in https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/#container-commands (or, if it is needed, documentation needs to be updated to say why it's still needed)

Custom containers *do explicitly* need to add the `ddev-global-cache` volume (though that was probably already needed for other things)

Host commands are now run in place from `~/.ddev/commands/host`, meaning they can't get out of sync anymore (it used to be possible to update a host command, and then run the *old* command because you didn't run `ddev start` or `ddev debug fix-commands` first)

The `.ddev/.global_commands/` directory is no longer in your project directories.

`ddev debug fix-commands` now works from anywhere, since it isn't tied to a specific project.


rfay (changes requested):
Thanks so much for taking this up! Sorry I haven't been responsive, we just finished our last riding day and will fly home tomorrow and I'll start digging myself out of all the things I'll be buried in.

I don't think this is what I was suggesting, that doesn't mean it's wrong. I did respond in the other issue, see if that raises any thoughts.

GuySartorelli:
Yup no worries, now that I know what volume you had intended, I'll go that route.

rfay:
Rebased, reviewing.

GuySartorelli:
Probably a bit early to review this, which is why it's in draft. This pr still copies them into a fairly arbitrary directory in the containers rather than the volume (due to confusion that you've since cleared up)

rfay:
Thanks, I'll leave it alone for now. Got sidetracked anyway.

GuySartorelli:
Still a ways to go with this, but just checking in at this stage: @rfay is this more along the lines of what you were thinking?

rfay (changes requested):
I haven't really fully manually tested this yet, but did run across this SEGV:

```
$ ddev config --docroot=xxx
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104de1e64]

goroutine 1 [running]:
github.com/ddev/ddev/cmd/ddev/cmd.addCustomCommandsFromDir(0x14000490180?, 0x0, {0x14000490180, 0x1f}, {0x140004b6200, 0xd, 0x1400031fb78?}, 0x1, 0x1400031fb58?)
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/commands.go:106 +0xd4
github.com/ddev/ddev/cmd/ddev/cmd.addCustomCommands(0x0?)
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/commands.go:59 +0xf0
github.com/ddev/ddev/cmd/ddev/cmd.init.51()
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/root.go:192 +0x168
```

docs/content/users/extend/custom-commands.md
```diff
@@ -36,11 +36,11 @@ open -a PhpStorm.app ${DDEV_APPROOT}
 
 To provide a command which will execute in a container, add a Bash script to `.ddev/commands/<container_name>`, for example, `.ddev/commands/web/mycommand`. The Bash script will be executed inside the named container. For example, see the [several standard DDEV script-based web container commands](https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web).
 
-You can run commands in custom containers as well as standard DDEV `web` and `db` containers. Use the service name, like `.ddev/commands/solr/<command>`. The only catch with a custom container is that your service must mount `/mnt/ddev_config` like the `web` and `db` containers do; the `volumes` section of `docker-compose.<servicename>.yaml` needs:
+You can run commands in custom containers as well as standard DDEV `web` and `db` containers. Use the service name, like `.ddev/commands/solr/<command>`. The only catch with a custom container is that your service must mount `/mnt/ddev-global-cache` like the `web` and `db` containers do; the `volumes` section of `docker-compose.<servicename>.yaml` needs:
```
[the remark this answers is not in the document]
GuySartorelli: Bear in mind that this page isn't about how to make custom extensions or docker containers etc. See [docs/content/users/extend/custom-compose-files.md](https://github.com/ddev/ddev/pull/5753/files#diff-0f22c880ae6590eb0f249987709640c16ceb7cba5946361032b7328e4a3f665b) which does still mention both.

If you want to mention `/mnt/ddev_config` here the docs should probably say how that ties into custom commands, which after this change I don't think it will, directly? Maybe it's just better to not mention either mount directly and instead link to the page that does?

GuySartorelli:
@rfay I was really hoping for a "yes that's along the right lines" or "no, because X"... but I'm going to assume that it's along the right lines since you haven't stated otherwise?

rfay:
Sorry for the delay, I need to spend a little time with it. Just went ahead and submitted the review early because of the accidentally discovered SEGV when I was about to start testing.

rfay:
`ddev drush` gets
```
$ ddev drush
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x100d6de64]

goroutine 1 [running]:
github.com/ddev/ddev/cmd/ddev/cmd.addCustomCommandsFromDir(0x1400027d6a0?, 0x0, {0x1400027d6a0, 0x1f}, {0x14000386c00, 0xd, 0x1400029fb58?}, 0x1, 0x1400029fb58?)
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/commands.go:106 +0xd4
github.com/ddev/ddev/cmd/ddev/cmd.addCustomCommands(0x0?)
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/commands.go:59 +0xf0
github.com/ddev/ddev/cmd/ddev/cmd.init.51()
	/Users/rfay/workspace/ddev/cmd/ddev/cmd/root.go:192 +0x168
```

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,83 @@
...
+		return err
 	}
 
-	// Make sure it's empty
-	err = fileutil.PurgeDirectory(copiedGlobalCommandPath)
-	if err != nil {
-		util.Error("Unable to remove %s: %v", copiedGlobalCommandPath, err)
-		return nil
-	}
+	for _, serviceDirOnHost := range commandDirs {
```
rfay, line 43: This could also be all dirs, just ignore the irrelevant `host`. I agree that's not completely correct. It could also copy everything but `host`, not make the decisions about services. But this is fine.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,83 @@
...
+		if err != nil {
+			return err
+		}
+		// Any command we find will want to be executable on Linux
+		for _, commandName := range commandFiles {
+			onHostFullPath := filepath.Join(serviceDirOnHost, commandName)
+			_ = os.Chmod(onHostFullPath, 0755)
+		}
+		// Copy commands into container (this will create the directory if it's not there already)
+		err = dockerutil.CopyIntoContainer(serviceDirOnHost, containerName, commandDirInService, "")
```
rfay, line 76: I would rather see one call here rather than multiple calls. That's an argument in favor of just copying everything. We already do too many `CopyIntoContainer` calls. Maybe it doesn't matter so much.

rfay (changes requested):
This is what I was hoping for, and works pretty well except of course attention to some SEGVs. Thanks!

It's conceivable that this could be piggy-backed onto one of the CopyIntoVolume commands that is already used to populate the ddev-global-cache, but the CopyIntoContainer here is probably OK, although I'd rather it were just one call. You can do a little performance analysis of the impact if you want, maybe it doesn't matter.

This is what 
* https://github.com/ddev/ddev/issues/4942

is about kind of, and since I didn't actually solve all the issues there it worries me.

Thanks!

pkg/ddevapp/commands.go
GuySartorelli: Yeah I probably will just copy everything - I had it this way because originally it wasn't copying into a shared volume, so with that implementation it made sense for each container to only have it's own commands.

GuySartorelli:
SEGV should be resolved - but I wasn't getting those before so I can't check that locally. Can you please double check for me?

pkg/ddevapp/commands.go
GuySartorelli: Copying everything _except_ host now. Copying host could lead to confusion at best, as those will only ever be run in place in the `.global_commands` dir.
All service commands are copied regardless of whether there's a container for that service in the project (mostly 'cause that was easier)

GuySartorelli: Does still need a separate call to delete the old commands dir though, so we don't have commands in the volume that don't exist in `~/.ddev/commands` anymore.

GuySartorelli:
> It's conceivable that this could be piggy-backed onto one of the CopyIntoVolume commands that is already used to populate the ddev-global-cache

This needs to happen both on `ddev start` and on `ddev debug fix-commands` and I don't think any other `CopyIntoVolume` calls are (or should be) made in both of those places.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,83 @@
...
-	copy2 "github.com/otiai10/copy"
-	"os"
-	"path/filepath"
 )
 
 // PopulateCustomCommandFiles sets up the custom command files in the project
 // directories where they need to go.
 func PopulateCustomCommandFiles(app *DdevApp) error {
+	// Remove legacy directory
+	legacyCommandDir := app.GetConfigPath(".global_commands")
```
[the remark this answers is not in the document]
GuySartorelli: Done. Thanks for pointing that out, good to know there's a place to put that sorta thing.

docs/content/users/extend/custom-commands.md
```diff
@@ -36,11 +36,11 @@ open -a PhpStorm.app ${DDEV_APPROOT}
...
 
 ```
     volumes:
-    - ".:/mnt/ddev_config"
+      - ddev-global-cache:/mnt/ddev-global-cache
```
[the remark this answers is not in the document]
GuySartorelli: Discussion about this in https://github.com/ddev/ddev/pull/5753#discussion_r1481935167

GuySartorelli:
Phew! That one was trickier than I expected, but should be all good now.
Docs CI failure is unrelated and fixed in #5858
The `buildkite/wsl2-docker-inside` CI failure seems to be unrelated as well - that's failing in other PRs right now too.

rfay:
Rebased to get docs test fixed. Looks like tests failed in general this go-around though. Move to "ready for review" when you're totally happy with it. 

Thanks so much for all your reviews, contributions, careful thought. Your reviews make a huge difference. They make my own life better. I'm still buried in this PR backlog, but at least when I see the PRs being addressed and other trusted reviewers looking carefully at them it gives me more confidence and makes my review a little easier.

rfay:
This no longer gets a panic on `ddev config --docroot=xxx`

rfay:
Looks like [TestCleanAndDelete](https://github.com/ddev/ddev/actions/runs/7942743050/job/21686342728?pr=5753#step:17:12090) went bad somehow.

GuySartorelli:
> Looks like [TestCleanAndDelete](https://github.com/ddev/ddev/actions/runs/7942743050/job/21686342728?pr=5753#step:17:12090) went bad somehow.

That's good 'cause that's the test I'll be getting rid of when I simplify this to not have a directory for each project 😅

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,48 @@
...
+	"github.com/ddev/ddev/pkg/nodeps"
+	"github.com/ddev/ddev/pkg/util"
 )
 
-// PopulateCustomCommandFiles sets up the custom command files in the project
+// PopulateGlobalCustomCommandFiles sets up the custom command files in the project
 // directories where they need to go.
-func PopulateCustomCommandFiles(app *DdevApp) error {
-
+func PopulateGlobalCustomCommandFiles() error {
```
GuySartorelli, line 18: This rename isn't really necessary, but it just makes it more explicit that this is only touching **global** commands. I initially thought it did something with project-level commands as well before I read through the code.

cmd/ddev/cmd/debug-fixcommands.go
```diff
@@ -2,31 +2,19 @@ package cmd
...
 		if err != nil {
 			util.Warning("Failed to populate custom command files: %v", err)
 		}
-		// If no-bind-mounts we have to do a start to push the commands back in there again.
-		if globalconfig.DdevGlobalConfig.NoBindMounts {
-			err = app.Start()
-			if err != nil {
-				util.Failed("Failed to restart with NoBindMounts set: %v", err)
-			}
-		}
```
GuySartorelli, line 29: I **think** it's okay to remove this, but I'd like another set of eyes over it since I'm not quite sure.

rfay: I do not believe this should be removed, but haven't studied your situation carefully.

You can easily experiment with `--no-bind-mounts` and test it. It's very rarely used, is intended for remote docker server. It was actually created when Colima didn't have bind mounts. But just turn it on and see if commands work.

GuySartorelli: Just tried this, and it seems it's not needed? I ran `ddev config global --no-bind-mounts=true`, then `ddev start` to get my project up.
Then modified a global web command and ran `ddev debug fix-commands` and the change was reflected in `/mnt/ddev-global-cache/global-commands/web/`.

GuySartorelli:
Rebased on master

rfay:
Rebased

rfay:
Oh, then I noticed that you just rebased :(

rfay:
I'm getting there. Have to get through the Advisory Group meeting this week and prep for it, then can start trying to review PRs better. Thanks for your patience and all the great work!

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,48 @@
...
+	dockerImages "github.com/ddev/ddev/pkg/docker"
+	"github.com/ddev/ddev/pkg/dockerutil"
+	"github.com/ddev/ddev/pkg/globalconfig"
+	"github.com/ddev/ddev/pkg/nodeps"
+	"github.com/ddev/ddev/pkg/util"
+)
 
+// PopulateGlobalCustomCommandFiles sets up the global command files
+// ddev-global-cache where they belong
+func PopulateGlobalCustomCommandFiles() error {
```
rfay, line 18: In this function we do **three** different container runs, so the `ddev start` performance impact is nontrivial. 

* RunSimpleContainer()
* CopyIntoVolume()
* RunSimpleContainer() again

Some thoughts about mitigation would be to 
1. Do this work after the webserver is started, where we have ddev-global-cache already mounted? Then it's just one app.Exec() maybe?
2. Consolidate it into a script so we only have one container run?

I think the idea is all fine, I just don't want to add more container runs to get us through the `ddev start`

rfay (changes requested):

pkg/ddevapp/commands.go
GuySartorelli: > Do this work after the webserver is started, where we have ddev-global-cache already mounted? Then it's just one app.Exec() maybe?

Can you please explain how that would cut down the number of executions? We still need to do 3 things;
1. Remove the old contents. This is necessary to avoid keeping old commands that have since been removed from `~/.ddev/commands`
1. Copy commands into the volume.
1. `chmod` the files so commands and completion scripts can be executed.

I don't know how to copy the commands into the volume without using `dockerutil.CopyIntoVolume()`. If you can show me a way to do that which can be combined with the first and third steps above, I'd be happy to try that.

GuySartorelli: I guess an alternative approach would be to bind `~/.ddev/commands` directly as a volume for all containers - then we only need the `chmod` and we can do that on the host.
I don't know what flow-on effects that might have though (I don't know much about docker volumes), and it's definitely a larger departure from how these are currently handled.

GuySartorelli: Ahh I've been mulling this over - is the problem more about the fact that anonymous containers are being started for 1 and 3 above, which is slower than just a `docker exec` in an existing container?

If that's the problem, I can swap those out for running those commands in the web container if it's available (I say "if it's available" because remember this is for `ddev debug fix-commands` as well, which might run when the project is stopped)

rfay: > is the problem more about the fact that anonymous containers are being started for 1 and 3 above

Yes, exactly. The Exec() doesn't take much time, but running a new container does. You're on the right track. It's just that we already have too many of those CopyIntoVolume() calls in `ddev start`. 

Since `ddev debug fix-commands` is a one-off it can be done any way we want. It would be sad for it to have different code, perhaps an if-started/else would do the job?

GuySartorelli: > It's just that we already have too many of those CopyIntoVolume() calls in ddev start.

As I mentioned earlier I'm not sure I can do anything about the `CopyIntoVolume()` call, but I can add conditionals to use the web container for the other two if it's available, so I'll do that.

If you can think of a way to use the web container to avoid calling `CopyIntoVolume()`, please let me know.

GuySartorelli: There we go - if there's an active running app we use `app.Exec()` but otherwise (e.g. if running from outside a project dir) it uses the anonymous container.

GuySartorelli: Looks like it did actually run after starting the containers during `ddev start`.
I've changed the order now so that the containers are set up and ready first. See https://github.com/ddev/ddev/compare/6ac57710c75a3840425def8178aa333ab25941b8..fec7914a43a1b7c758d721efbf28a543357ec9ae

The post-start hook runs *after* this, so I think there is no effective change for users.

rfay:
Rebasing for review, and the test runners need some exercise any way :)

rfay:
Testing this I ran into https://github.com/ddev/ddev-redis/pull/22 right away :)

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 
-	if !fileutil.FileExists(projectCommandPath) || !fileutil.IsDirectory(projectCommandPath) {
-		return nil
-	}
 	return nil
 }
+
+// performTaskInContainer runs a command in the web container if it's available,
+// but uses an anonymous container otherwise.
+func performTaskInContainer(command []string) (string, string, error) {
```
rfay, line 50: Thanks for this. I think it could probably be a global utility function, but good as is.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
+	app, err := GetActiveApp("")
+	if err == nil {
+		status, _ := app.SiteStatus()
```
rfay, line 53: This could actually be more focused and just see if web container is running. OK like this.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
+			opts := &ExecOpts{
+				RawCmd:    command,
+				Tty:       false,
+				NoCapture: false,
+			}
+			return app.Exec(opts)
+		}
+	}
+
+	// If there is no running active site, use an anonymouse container instead.
```
rfay, line 65:
```suggestion
	// If there is no running active site, use an anonymous container instead.
```

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 	if err != nil {
 		return nil
 	}
 
-	projectCommandPath := app.GetConfigPath("commands")
-	// Make sure our target global command directory is empty
-	copiedGlobalCommandPath := app.GetConfigPath(".global_commands")
-	err = os.MkdirAll(copiedGlobalCommandPath, 0755)
+	// Remove contents of the directory, if the directory exists and has some contents
+	commandDirInVolume := "/mnt/ddev-global-cache/global-commands/"
```
rfay, line 26: This does seem like overkill, but same as what we're doing everywhere. It would be cool if we could do an rsync that would update+delete, but I don't see how to do that easily.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
-		util.Error("Unable to create directory %s: %v", copiedGlobalCommandPath, err)
-		return nil
+		return fmt.Errorf("unable to rm %s: %v (stderr=%s)", commandDirInVolume, err, stderr)
 	}
 
-	// Make sure it's empty
-	err = fileutil.PurgeDirectory(copiedGlobalCommandPath)
+	// Copy commands into container (this will create the directory if it's not there already)
+	uid, _, _ := util.GetContainerUIDGid()
+	err = dockerutil.CopyIntoVolume(sourceGlobalCommandPath, "ddev-global-cache", "global-commands", uid, "host", false)
```
rfay, line 34: This is still running an anon container. But we're down to one now :) We do have to figure out how to get things in there. We could do the equivalent of a `docker cp` here using the docker API, and use the existing running container. If you're interested in that, it could be a really useful thing and could improve things elsewhere. Otherwise, thanks for getting this down to one.

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 	if err != nil {
-		util.Error("Unable to remove %s: %v", copiedGlobalCommandPath, err)
-		return nil
+		return err
 	}
 
-	err = copy2.Copy(sourceGlobalCommandPath, copiedGlobalCommandPath)
+	// Make sure all commands can be executed
+	_, stderr, err = performTaskInContainer([]string{"sh", "-c", "find " + commandDirInVolume + " -type f -exec chmod 0755 {} \\;"})
```
rfay, line 40: I don't think we need a find here; in there it's not a problem that we might be changing modes on a few things that don't matter. This should be a little faster.

```suggestion
	_, stderr, err = performTaskInContainer([]string{"sh", "-c", "chmod -R u+rwx " + commandDirInVolume})
```

rfay (changes requested):
This is working great. I manually tested and stepped through both Start() and fixcommands() (outside the project directory). 

Some minor suggestions for improvement.

Looking forward to @stasadev review

stasadev (approved):
It looks good to me, I tested it and didn't find any flaws.

I agree with @rfay's suggestions.

GuySartorelli:
Haven't had much time to look at this - I'll try to get to this in the next week or so but feel free to make changes yourselves if you need it done quicker.

rfay:
Sure appreciate all your efforts!

rfay:
We're getting closer to the v1.23.0 release process, and this would be lovely to have in there. Let us know if there's anything you need. I know you don't have unlimited time.

GuySartorelli:
I'll try to make some time this weekend to work on this and #5918

pkg/ddevapp/commands.go
GuySartorelli: Given we're gearing up for a release and I've got a little less energy right now I'll opt to leave this as an enhancement someone can make in the future.

pkg/ddevapp/commands.go
GuySartorelli: Yeah I couldn't think of a way to do that - definitely worth investigating as a future enhancement though.

pkg/ddevapp/commands.go
GuySartorelli: As above - if you're getting close to a release I'd say lets leave this as a future enhancement.

GuySartorelli: I've turned the comment into [an issue](https://github.com/ddev/ddev/issues/6004)

pkg/ddevapp/commands.go
GuySartorelli: Done

pkg/ddevapp/commands.go
GuySartorelli: Oops! 😅 Fixed.

GuySartorelli:
Addressed feedback. Assuming it all goes green, it should be good now.

GuySartorelli:
All green! 🥳

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 		return nil
 	}
 
-	projectCommandPath := app.GetConfigPath("commands")
-	// Make sure our target global command directory is empty
-	copiedGlobalCommandPath := app.GetConfigPath(".global_commands")
-	err = os.MkdirAll(copiedGlobalCommandPath, 0755)
+	// Remove contents of the directory, if the directory exists and has some contents
+	commandDirInVolume := "/mnt/ddev-global-cache/global-commands/"
+	_, stderr, err := performTaskInContainer([]string{"sh", "-c", "rm -r " + path.Join(commandDirInVolume, "*") + " || true"})
```
rfay, line 27: I don't think the `sh -c` is required here, how about just this? There will never be any stderr though, still worth checking err

```suggestion
	_, _, err := performTaskInContainer([]string{"rm" "-rf", commandDirInVolume})
```

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 	if err != nil {
-		util.Error("Unable to create directory %s: %v", copiedGlobalCommandPath, err)
-		return nil
+		return fmt.Errorf("unable to rm %s: %v (stderr=%s)", commandDirInVolume, err, stderr)
```
rfay, line 29:
```suggestion
		return fmt.Errorf("unable to rm %s: %v ", commandDirInVolume, err)
```

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
+				RawCmd:    command,
+				Tty:       false,
+				NoCapture: false,
+			}
+			return app.Exec(opts)
+		}
+	}
+
+	// If there is no running active site, use an anonymous container instead.
+	containerName := "PopulateGlobalCustomCommandFiles" + nodeps.RandomString(12)
```
rfay, line 66: Refactored function (thanks!) didn't get updated anon name. Doesn't matter but might confuse some poor dev someday.
```suggestion
	containerName := "performTaskInContainer" + nodeps.RandomString(12)
```

pkg/ddevapp/commands.go
```diff
@@ -1,47 +1,69 @@
...
 
-	if !fileutil.FileExists(projectCommandPath) || !fileutil.IsDirectory(projectCommandPath) {
-		return nil
-	}
 	return nil
 }
+
+// performTaskInContainer runs a command in the web container if it's available,
+// but uses an anonymous container otherwise.
+func performTaskInContainer(command []string) (string, string, error) {
```
rfay, line 50: This can probably be public PerformTaskInContainer() here or in the future?

pkg/ddevapp/config.go
```diff
@@ -712,6 +711,21 @@ func (app *DdevApp) FixObsolete() {
 			}
 		}
 	}
+
+	// Remove old .global_commands directory
+	legacyCommandDir := app.GetConfigPath(".global_commands")
+	if fileutil.IsDirectory(legacyCommandDir) {
```
rfay, line 717: Why not just `os.RemoveAll(legacyCommandDir)` instead of this whole stanza?

rfay (changes requested):
This is pretty amazingly great and works fine in the manual testing I've done. It's so much more elegant than the previous approach, thanks!

I made a few smaller suggestions that I think can make the code simpler.

pkg/ddevapp/commands.go
GuySartorelli: It's only ever used from `PopulateGlobalCustomCommandFiles` which is why I kept the name how it was. But I'll make this requested change.

pkg/ddevapp/commands.go
GuySartorelli: I don't think this is the right place for it if it'll be public - and would warrant having its own tests and thinking more carefully about the implementation than I thought about it for its current purposes.
You mentioned earlier that you're preparing for a release and given my time constraints at the moment maybe this is better left as a future enhancement?

pkg/ddevapp/config.go
GuySartorelli: Because I didn't know that existed 😂

rfay:
Rebased, fixed conflict.

rfay:
I'm pretty sure this is very close. You can actually just accept the suggestions from the web UI if you like them.

GuySartorelli:
Feel free to update the PR - I haven't had time to look at this and don't want to commit changes I haven't tried myself.

pkg/ddevapp/commands.go
GuySartorelli: Done

pkg/ddevapp/commands.go
GuySartorelli: Done. I don't get why this works here but doesn't work for the chmod below (I did try it) - but it doesn't really matter why. It is how it is.

pkg/ddevapp/commands.go
GuySartorelli: Done

pkg/ddevapp/config.go
GuySartorelli: Done

GuySartorelli:
Rebased and requested changes made.

rfay:
All checks passed!

cmd/ddev/cmd/debug-fixcommands.go
```diff
@@ -2,31 +2,19 @@ package cmd
 
 import (
 	"github.com/ddev/ddev/pkg/ddevapp"
-	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
 )
 
-// DebugFixCommandsCmd fixes up custom/shell commands without having to do a ddev start
+// DebugFixCommandsCmd fixes up global container commands without having to do a ddev start
```
rfay, line 9: There isn't even a reason for this `ddev debug fixcommands` to exist any more is there, after this PR?

rfay (approved):
You are a brave one to take this on and stick to it all this time. Thanks! We can consider removing `ddev debug fixcommands` in a followup. Congrats!

cmd/ddev/cmd/debug-fixcommands.go
GuySartorelli: It certainly becomes a less important command, but I think it will be useful to have a command that syncs global commands into the volume (if someone adds or removes one for example) without having to run `ddev start` on a project.

PR #5754, opened 2024-01-30
fix: wording on install docs home page about Docker providers should be more generic

joachim-n:
## The Issue

The install docs home page at https://ddev.readthedocs.io/en/stable/users/install/ only mentions two Docker providers, but DDEV now recommends more than that.

Change the wording to be generic, so we don't have a second list to maintain here.

### DIFF
--- a/docs/content/users/install/index.md
+++ b/docs/content/users/install/index.md
@@ -1,5 +1,5 @@
 # Installation
 
-Whatever system you’re on, you’ll first need to [Install Docker or Colima](docker-installation.md), then [Install DDEV](ddev-installation.md).
+Whatever system you’re on, you’ll first need to [install a Docker provider](docker-installation.md), then [Install DDEV](ddev-installation.md).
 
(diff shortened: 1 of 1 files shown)


docs/content/users/install/index.md
```diff
@@ -1,5 +1,5 @@
 # Installation
 
-Whatever system you’re on, you’ll first need to [Install Docker or Colima](docker-installation.md), then [Install DDEV](ddev-installation.md).
+Whatever system you’re on, you’ll first need to [install a Docker provider](docker-installation.md), then [Install DDEV](ddev-installation.md).
```
stasadev, line 3:
```suggestion
Whatever system you’re on, you’ll first need to [install a Docker provider](docker-installation.md), then [install DDEV](ddev-installation.md).
```

stasadev (approved):
Thank you!

rfay (approved):
Thanks!

PR #5758, opened 2024-01-31
fix: update project schema nodejs_version, for #5400

stasadev:
## The Issue

- #5400

After changes in `nodejs_version`, we should accept any possible value.

Removes enum restrictions for `nodejs_version`.

Removes all unicode characters as in

- #5653

### DIFF
--- a/pkg/ddevapp/schema.json
+++ b/pkg/ddevapp/schema.json
@@ -218,13 +218,8 @@
       "type": "boolean"
     },
     "nodejs_version": {
-      "description": "Node.js version for the web container’s “system” version.",
-      "type": "string",
-      "enum": [
-        "16",
-        "18",
-        "20"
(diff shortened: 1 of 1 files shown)


rfay (approved):
Glad you got that! I wonder if there's a way to specify a 'numeric | decimal'

Regardless, it's great.

stasadev:
There is actually a way https://json-schema.org/understanding-json-schema/reference/type

```json
{ "type": ["number", "string"] }
```

But running `ddev config --auto` will convert a number to a string in `config.yaml` (since the yaml type is string), so I don't think it's necessary:

https://github.com/ddev/ddev/blob/2defedb31dd103d945fac90f260909796a0a5dc0/pkg/ddevapp/ddevapp.go#L127

Issue #4854, opened 2023-04-27
Create a quickstart for a simple Node.js app (without PHP)


MurzNN:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

I know that DDEV is mostly for PHP projects, but it is also can be used for Node.js apps pretty easily! And to simplify things will be good to add a quickstart script for a simple Node.js app (without PHP).

### Describe your solution

The quickstart script should just disable launching Nginx in the container, to allow mapping the Node.js app to the 80 port, or just route web traffic to another port. The most popular ports for Node.js apps are: 3000, 1337, 8080, so let's just use 3000 for it?

So, the task is to make https://node-js-project.ddev.site be mapped to localhost:3000 of the container, and that's all.

And if we make this, we can significantly extend the DDEV userbase with many JS-only developers!

rfay:
If you can suggest what the right thing is for a "standard" backend, this can go places @MurzNN.

MurzNN:
The "standard" is just a proxy to the 3000 port (or any other default port), yes - it can be achieved via Nginx config too, if this will be easier, so something like this:
https://www.daeyounglim.com/post/configuring-nginx-for-nodejs/
or this: https://tutorials.tinkink.net/en/nginx/nginx-reverse-proxy-nodejs.html#basic-configuration

```
server {
    listen 80;
    server_name xxx.ddev.site;

    error_log /var/log/nginx/error.log;
    access_log /var/log/nginx/accss.log;

    location / {
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Host $http_host;
        proxy_set_header X-Nginx-Proxy true;
        proxy_http_version 1.1;
        # Native address of Node.js, note the port
        proxy_pass http://localhost:3000;
    }
}
```
And will be good to ask the user about the port number to configure (3000 by default, 1337, 8080, 4000 as suggestions, and any custom value), if it's possible.

rfay:
Thanks!

andy-blum:
+1 for a node-proxy type. I did a custom nginx conf file for a project recently that ran multiple services at once and fed a domain to each. Any solution for a node-type quick start should have an easily configurable port option in the config.yml. While 3000 is probably the _most_ common, and all allow customizing the port in the run command, some have different defaults like `:1234` `:7777`, etc

rfay:
[...]

PR #5761, opened 2024-02-01
docs: adds second snippet to Multiple Docroots in nginx

bmartinez287:
Added notes from Andy Blum on issue #4854


docs/content/users/extend/customization-extendibility.md
```diff
@@ -214,6 +216,78 @@ For example, to make all HTTP URLs redirect to their HTTPS equivalents you might
...
+
+    error_log /dev/stdout info;
+    access_log /var/log/nginx/access.log;
+
+    include /etc/nginx/common.d/*.conf;
+    include /mnt/ddev_config/nginx/*.conf;
+}
+```
+
+.ddev/nginx_full/randomnamesecond.conf
```
bmartinez287, line 256:
```suggestion
.ddev/nginx_full/first-random-name.conf
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -214,6 +216,78 @@ For example, to make all HTTP URLs redirect to their HTTPS equivalents you might
 
 After adding a snippet, run `ddev restart` to make it take effect.
 
+#### Second snippet
+
+To route the *.ddev.site URLs to the right node apps, you'll probably want to make new Nginx config files. The default one (.ddev/nginx_full/nginx-site.conf) will route all incoming traffic to /var/www/html and assume a static HTML file or PHP app is there to accept the incoming request. Since node apps tend to run on other ports, we need to proxy those requests to the right ports
+
+.ddev/nginx_full/randomname.conf
```
bmartinez287, line 223:
```suggestion
.ddev/nginx_full/original-random-name.conf
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -202,6 +202,8 @@ The generated `.ddev/nginx_full/seconddocroot.conf.example` demonstrates how to
 
 ### nginx Snippets
 
+#### First snippet
```
rfay, line 205: Please use a descriptive title instead of "First" and "Second". 

```suggestion
#### Example Nginx Snippet: Redirecting http to https
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -214,6 +216,78 @@ For example, to make all HTTP URLs redirect to their HTTPS equivalents you might
 
 After adding a snippet, run `ddev restart` to make it take effect.
 
+#### Second snippet
```
rfay, line 219:
```suggestion
#### Example Nginx Snippet: Using Node.JS Server as Back-End
```

rfay (changes requested):
Thanks for capturing this. There are still some problems with the docs check, https://github.com/ddev/ddev/actions/runs/7735247282/job/21090580613?pr=5761 etc.

rfay:
Still some textlint trouble, see https://github.com/ddev/ddev/pull/5761/files/606796bfec681a469df552b9b83c7d65a352e1fd..d12d5c6b11ba7a722c6fb4119d736ba6cea9d3a6

rfay (changes requested):
Ooops, unfortunately your new node.js one is not a snippet, it's an example of using nginx_full. Could you please move it? Snippets are really quite different.

bmartinez287:
Sure, where do examples like this could go? Any preference?

rfay:
The underlying problem is that there isn't good distinction in that docs page between snippets (.ddev/nginx) and full config (.ddev/nginx_full). 

At one time I tried to remove the snippet feature and people screamed :)  Maybe the docs got confused.

If you wouldn't mind trying to add the difference into your PR it would be appreciated. 

* "Multiple Docroots" is an example of nginx_full (where the whole nginx config is replaced)
* The HTTP redirect is an example of a snippet
* Yours is an example of nginx_full, so the very least would be to turn multiple docroots and yours into nginx_full examples.
* the idea of "snippet" is never explained, needs to be explained (where the snippet is placed into the config)

Sorry to have you stumble on this weakness, would love it if you could sort it out.

rfay:
Sorry to have you discovering that problem. I'll be happy to take it up if you don't want to. Seems some docs must have been lost somewhere along the way.

bmartinez287:
I think I got most everything but explaining the snippets. That I'm not so sure.

rfay:
Snippets (that go in .ddev/nginx) are included in the last line of the nginx-site.conf:
```
    include /mnt/ddev_config/nginx/*.conf;
```

https://github.com/ddev/ddev/blob/c9d8f544333e7d8f4692280279169df5f228fdcc/pkg/ddevapp/webserver_config_assets/nginx-site-php.conf#L70

That means that they can do anything they want in the `server` stanza, but can't do anything outside the `server` context.

nginx_full, of course, is the complete contents of nginx-site.conf, which can actually contain more than `server`

bmartinez287:
I rephrased it a little, but I will between that and the full example right below it it provides a clear picture. 
I could have done the nginx custom config for ddev.com ddev config after looking and Andy example.
Without it there are too many unknowns.

rfay:
I just realized that this probably actually belongs in https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/#using-nodejs-with-ddev instead of in the nginx section (but the nginx section to refer to it)

rfay:
I'm spending a little time with this.

rfay:
I know tremendously little about the node.js ecosystem. But I assume there are websites entirely implemented in node.js, which could then be used with an nginx front-end and node.js back-end. That's essentially what you're doing here, but you're putting multiple back-ends in there and leaving the php-fpm back-end in use.

rfay:
I think this ends up being too complex for the regular docs. Would you consider doing this as an article in ddev.com, or Stack Overflow?

I do think we can improve what we have currently, and I also agree that we should capture @andy-blum 's work, but I think this PR is going to be too much.

bmartinez287:
Sounds good

andy-blum:
@bmartinez287 I started a blog post for my own site based on that same comment. Feel free to use it/modify it/etc

https://github.com/andy-blum/andy-blum/pull/8/files

PR #5775, opened 2024-02-03
test: experiment with rancher desktop start to find out problem

rfay:
## The Issue

Since upgrade Rancher Desktop startup has been failing

Experiment to find out what the problem is.


rfay:
The upgrade of the buildkite docker login plugin caused `docker context` to fail, seemed to be using a location other than the homedir for ~/.docker.

PR #5780, opened 2024-02-04
test: CircleCI upgrade to go v1.21

rfay:
## The Issue

* #5732 introduced the slices package which appeared in go v1.21, and CircleCI was set up for v1.20

Update CircleCI config

I'm not 100% sure that we need to keep testing Linux arm64. It's very lightly used, just by people with Raspberry Pi or MS Surface Pro. And the actual testing is probably adequate on macOS. We *could* make a mistake with images, etc.

### DIFF
--- a/.ci-scripts/linux_arm64_setup.sh
+++ b/.ci-scripts/linux_arm64_setup.sh
@@ -5,7 +5,7 @@ set -o errexit
 # Basic tools
 
 set -x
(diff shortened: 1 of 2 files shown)


rfay:
Moved to draft; this can't be pulled until https://github.com/ddev/ddev/pull/5780/commits/1872ab26f58950c05b2b8974364b2f603162215f is reverted.

PR #5785, opened 2024-02-05
feat: enable autocompletion for custom commands, fixes #5783

### LINKED ISSUE #5783: Enable custom commands to have custom autocompletion

Currently there's no autocompletion for custom command argumentss. This means users have to constantly check `help` to see what arguments are available to them.

### Describe your solution

There are two kinds of autocompletion at play:
1. static completion - i.e. specific values that can be used for arguments which are known ahead of time
2. dynamic completion - i.e. values that are evaluated at runtime, most likely through a bash function

The static completion would be really easy to add via a new annotation, e.g. the annotation for `ddev xdebug` might look like this:
```bash
## AutocompleteTerms: ["on","off","enable","disable","toggle","status"]
```

Dynamic completion would require a bash script that's separate to the main command script.
I recommend having a `autocompletion` subdirectory inside each service command dir. Any script inside the `autocompletion` directory which matches the name of a command will be used as the script to call for autocompletions.

e.g.
```text
~/.ddev/commands/
├─ host/
   ├─ autocompletion/
      └─ script2
   ├─ script1
   ├─ script2
   └─ ...
├─ web/
   ├─ autocompletion/
      ├─ nvm
      └─ ...
   ├─ script1
   ├─ nvm
   └─ ...
└─ ...
```
Above, when setting up the `nvm` command, the `~/.ddev/commands/web/autocompletion/nvm` script will be used for any autocompletions.

### Describe alternatives

_No response_

### Additional context

Static Flag autocompletion is already provided. We could look at doing dynamic flag autocompletion as well (e.g. completion for values that can be passed to flags).

Any discussion about that should be included in this issue, because it will likely affect the way this is implemented. Forking that off as a separate conversation would risk implementing autocompletion for args in a way that makes it difficult to introduce dynamic flag autocompletion without breaking changes.

GuySartorelli:
## The Issue

- #5783

For static completions, a new annotation is introduced which provides an array of valid argument values. They aren't validated against, they're only used for completions.

For dynamic completions, if there's a script in the `autocompletions` subdirectory with the same name as a command in that service command directory, it'll be called and the current cli args will be passed in. That script can then use those args to generate a string of line-break-separated values which are valid (see edit to docs)

### Autocompletion scripts

All CLI args (including the name of the command, since this seems to be standard with bash autocompletion) are passed to the script as arguments. This includes an empty argument if `tab` is pressed after a space.

The return value from scripts should be a space-separated string of value arguments. The script doesn't have to check that what it's suggesting actually starts with the letters in the last argument - cobra handles that for us.
Note that we may want to do something other than space separated values, in case spaces are valid inside arguments somehow.... but for now this is more of a POC than anything.

I originally wanted to just set the `COMP_WORDS` and `COMP_CWORD` environment variables, and have the scripts check those and then set `COMPREPLY` with the result (see [this how-to to see why, and how those are involved in bash completion](https://www.baeldung.com/linux/shell-auto-completion)), but scripts can't set the environment variable in the shell docker is executing, so there'd be no way to get the result that way.
So since we can't use the normal bash completion result, we may as well make the arguments easier to pass in and parse as well.

I've added the new annotation to all the commands that seem to have obvious pre-determined valid arguments.

I've added dynamic completion for `nvm` and `npm`. There are other commands like `drush` that could probably also benefit from this, and I'd welcome developers who use those commands to submit PRs that introduce completion for them. The main purpose of this PR is not to add completion to all the commands, but to add the *capability* of adding completion to commands.

Wherever possible, completion should ideally be forwarded to a completion script being maintained by whoever maintains the underlying tool (such as I've done with `nvm` and `npm`). Sometimes (like with `yarn`) that won't be an option, so for some commands there will have to either not be any completion support, or the completion will need to be manually kept in sync with any changes to the underlying command.

`ddev restart`
`ddev debug fix-commands`
Try `ddev nvm <tab>`, or `ddev xdebug <tab>`

## Automated Testing Overview

Added tests both for the new annotation and for dynamic autocomplete scripts

## Related Issue Link(s)

- #5783

## Release/Deployment Notes

Developers will be able to add the new annotation and dynamic autocompletion scripts to their projects, global commands, and extensions.
There are no BC concerns with this approach.

There is a change to the `.bashrc` for the web image, so that image will need to be rebuilt.

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
+++ b/containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
@@ -116,6 +116,9 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && source "$NVM_DIR/nvm.sh"
 [ -s "$NVM_DIR/bash_completion" ] && source "$NVM_DIR/bash_completion"
 
+# bash completion for npm
+[ -n "$(type -t npm)" ] && eval "$(npm completion)"
+
 for f in /etc/bashrc/*.bashrc; do
   source $f;
 done
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -65,6 +65,24 @@ Changes to the command files in the global `.ddev` directory need a `ddev start`
 
 There are many examples of [global](https://github.com/ddev/ddev/tree/master/pkg/ddevapp/global_dotddev_assets/commands) and [project-level](https://github.com/ddev/ddev/tree/master/pkg/ddevapp/dotddev_assets/commands) custom/shell commands that ship with DDEV you can adapt for your own use. They can be found in your `~/.ddev/commands/*` directories and in your project’s `.ddev/commands/*` directories. There you’ll see how to provide usage, examples, and how to use arguments provided to the commands. For example, the [`xdebug` command](https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/xdebug) shows simple argument processing and the [launch command](https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/host/launch) demonstrates flag processing.
 
+## Command Line Completion
+
+If your custom command has a set of pre-determined valid arguments it can accept, you can use the [`AutocompleteTerms`](#autocompleteterms-annotation).
+
+For dynamic completion, you can create a separate script with the same name in a directory named `autocomplete`.
+For example, if your command is in `~/.ddev/commands/web/my-command`, your autocompletion script will be in `~/.ddev/commands/web/autocomplete/my-command`.
+
+When you press tab on the command line after your command, the associated autocomplete script will be executed. The current command line (starting with the name of your command) will be passed into the completion script as arguments. If there is a space at the end of the command line, an empty argument will be included.
+
+For example:
+
+* `ddev my-command <tab>` will pass `my-command` and an empty argument into the autocomplete script.
+* `ddev my-command som<tab>` will pass `my-command`, and `som` into the autocomplete script.
+
+The autocomplete script should echo the valid arguments as a string separated by line breaks. You don't need to filter the arguments by the last argument string (e.g. if the last argument is `som`, you don't need to filter out any arguments that don't start with `som`). That will be handled for you before the result is given to your shell as completion suggestions.
+
+The web container's [`nvm` autocomplete script](https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/autocomplete/nvm) shows how this can be used to forward completion requests to a relevant script in the container.
+
 ## Environment Variables Provided
 
 A number of environment variables are provided to these command scripts. These are generally supported, but please avoid using undocumented environment variables. Useful variables for host scripts are:
@@ -185,6 +203,14 @@ The following fields can be used for a flag definition:
 * `DefValue`: default value for usage message
 * `NoOptDefVal`: default value, if the flag is on the command line without any options
 * `Annotations`: used by cobra.Command Bash autocomplete code (see <https://github.com/spf13/cobra/blob/main/site/content/completions/bash.md>)
+
+### `AutocompleteTerms` Annotation
+
+If your command accepts specific arguments, and you know ahead of time what those arguments are, you can use this annotation to provide those arguments for autocompletion.
+
+Usage: `## AutocompleteTerms: [<list-of-valid-arguments>]`
+
+Example: `## AutocompleteTerms: ["enable","disable","toggle","status"]`
 
 ### "CanRunGlobally" Annotation
 
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/autocomplete/npm
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/autocomplete/npm
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+#ddev-generated
+# set env variables required for nvm's bash completion script
+COMP_WORDS=("$@")
+COMP_CWORD=$(($# - 1))
+# run the actual script
+_npm_completion
+# output the result (which was stored in COMPREPLY) as a new-line delimited string
+printf "%s\n" "${COMPREPLY[@]}"
--- a/cmd/ddev/cmd/autocompletion_test.go
+++ b/cmd/ddev/cmd/autocompletion_test.go
@@ -1,10 +1,14 @@
 package cmd
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/fileutil"
 	asrt "github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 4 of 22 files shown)


rfay:
Yay!

containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
```diff
@@ -116,6 +116,9 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && source "$NVM_DIR/nvm.sh"
 [ -s "$NVM_DIR/bash_completion" ] && source "$NVM_DIR/bash_completion"
 
+# bash completion for npm
+[ -n "$(type -t npm)" ] && eval "$(npm completion)"
+
```
GuySartorelli, line 121: Is this the right file to put this in for it to apply to the webserver? I haven't touched anything to do with the containers before and the naming here is a little confusing.

pkg/ddevapp/global_dotddev_assets/commands/web/autocomplete/npm
GuySartorelli, line 1: This won't work without the change to bashrc

rfay:
> Docs CI failure is due to a file being linked to which is added in this PR 

That's fine. But it probably doesn't need a link in docs?... `https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/autocomplete/nvm`

GuySartorelli:
> But it probably doesn't need a link in docs?

No worries, I'll remove it then. I had it there to follow the style of other docs on that page, which link to files to provide examples rather than giving extensive code examples. I think it's important to show what the code for an autocomplete script might look like since it's not the way a normal normal bash completion script works.

GuySartorelli:
Added completion func for host commands and removed that link from the docs.

GuySartorelli:
And rebased

GuySartorelli:
The failing lint CI rather frustratingly doesn't seem to say _where_ the failing comes from... but the files tab in the PR points those failures at lines that aren't actually changes being made in this PR.

rfay:
I'm frustrated when linter changes as well. Spent an ugly few hours fixing all the complaints it had in https://github.com/ddev/ddev/pull/5812 - I should have pointed out to you that I was working on it. 

Rebasing this to fix that, everything else is coming through OK.

rfay:
BTW, `make staticrequired` locally runs those tests; you do have to have the linters installed.

rfay:
Pushed commit to pacify linter.

docs/content/users/extend/custom-commands.md
```diff
@@ -186,6 +204,14 @@ The following fields can be used for a flag definition:
 * `NoOptDefVal`: default value, if the flag is on the command line without any options
 * `Annotations`: used by cobra.Command Bash autocomplete code (see <https://github.com/spf13/cobra/blob/main/site/content/completions/bash.md>)
 
+### "AutocompleteTerms" Annotation
```
rfay, line 207:
```suggestion
### `AutocompleteTerms` Annotation
```

With this you won't need it in spellchecklist any more

.spellcheckwordlist.txt
```diff
@@ -7,6 +7,7 @@ Apache
 Acquia
 amazee.io
 Autocompletion
+AutocompleteTerms
```
rfay, line 10:
```suggestion
```

Shouldn't be necessary, was forced by simple error in markdown.

containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
rfay: This seems like it ought to be a fine place, of course only for bash, and only works if login shell.

docs/content/users/extend/custom-commands.md
```diff
@@ -65,6 +65,24 @@ Changes to the command files in the global `.ddev` directory need a `ddev start`
...
 
+## Command Line Completion
+
+If your custom command has a set of pre-determined valid arguments it can accept, you can use the [`AutocompleteTerms`](#autocompleteterms-annotation).
+
+For dynamic completion, you can create a separate script with the same name in a directory named `autocomplete`.
+For example, if your command is in `~/.ddev/commands/web/my-command`, your autocompletion script will be in `~/.ddev/commands/web/autocomplete/my-command`.
```
rfay, line 73: Probably needs link to real example. You probably removed the real example link because it was breaking link check :) You can exclude a link check by adding to the markdown-link-check.json.

rfay (approved):
This is pretty amazing, and works very nicely (tested with bash). Thanks so much. Takes another piece of friction out of the DDEV world.

Made some minor requests.

docs/content/users/extend/custom-commands.md
GuySartorelli: https://github.com/ddev/ddev/pull/5785#issuecomment-1934236761
> it probably doesn't need a link in docs

Just to clarify, since you're saying the opposite now... You *do* want the link now?

containers/ddev-webserver/ddev-webserver-etc-skel/etc/skel/.bashrc
GuySartorelli: It's fine that it's only for bash, because this is inside the web container and not on the host.

The auto completion script is explicitly executed in the container in bash, and the result gets passed back to cobra which then outputs it correctly for whichever she'll the host is using.

rfay: People *can* use zsh inside the container, but it's pretty exotic, and not sure it affects this case anyway. https://ddev.com/blog/oh-my-zsh-using-custom-commands-and-other-goodies-to-add-to-ddev/

docs/content/users/extend/custom-commands.md
rfay: Yes, I think it needs real, maintained example. Just exclude it from the check or whatever. I didn't realize before out of context how important a real maintained example was.

docs/content/users/extend/custom-commands.md
GuySartorelli: That would be inconsistent with all of the other annotation headings... I'd lean towards having them all either all using backticks or all using double quotes, rather than mixing and matching.

That said, happy to make whatever change you think is appropriate - I'll leave as is for now, please confirm if you want just this one in backticks, all in backticks, or leave in double quotes.

rfay: It's not that important, but `AutoCompleteTerms` is not a word, should not be spell-checked, and it's an "annotation". The others are probably wrong :) But perhaps they also happen to be words.

docs/content/users/extend/custom-commands.md
GuySartorelli: Sweet, added the link back.

.spellcheckwordlist.txt
GuySartorelli: Removed

docs/content/users/extend/custom-commands.md
GuySartorelli: I've swapped this one instance to backticks. I'll probably open a new issue in a few minutes to standardise it so they're all using backticks. Consistency will both make it easier for contributers in the future (since they'll be able to just copy the existing formatting without running into this issue) and remove any possible confusion from users (having one different from the rest might make it seem like there's something special or different about its usage or syntax or etc when there isn't)

GuySartorelli:
Added automated tests and addressed feedback. Should be good now, assuming all goes green.

docs/content/users/extend/custom-commands.md
GuySartorelli: > You can exclude a link check by adding to the markdown-link-check.json.

I don't think we should exclude this link there just to pass CI in the PR, because then as soon as the PR gets merged we'd need a separate PR just to remove it from the file, so that the link _can_ be checked from then onwards (in case the file gets deleted or renamed in the future, we still _do_ want to check the link!)

rfay:
Thanks for all the great work on this (and in this whole area!)

rpkoller:
@GuySartorelli i think i found a small bug in the npm autocomplete file. i was trying to understand the inner workings of that autocomplete to update the file for drush and just tested npm and nvm. nvm works but for npm i get ```ddev npm /mnt/ddev_config/.global_commands/web/autocomplete/npm:\ line\ 8:\ _npm_completion:\ command\ not\ found ```

comparing with nvm i suppose `_npm_completion`has to be changed to `_npm`?

rfay:
If you've found a bug, please open a new issue describing it @rpkoller , thanks. Or if you know how to solve it we'll welcome your PR!

GuySartorelli:
@rpkoller the nom completion relies on a change to the bashrc in the web container which was included in this PR - but which has not yet been included in a released build of that docker image.

To test the npm completion, add the snippet from the .bashrc change to your web container's .bashrc

rpkoller:
> If you've found a bug, please open a new issue describing it @rpkoller , thanks. Or if you know how to solve it we'll welcome your PR!

i wanted to ask on the PR that added the change first, since there is always the option that i have missed something. and as it turned out that was the case. i dont want to add any extra noise creating a new issue or trying to fix the issue with the PR.

rfay:
Actually, there's nothing to be done here, those changes did make it into current image. 
```
# bash completion for npm
[ -n "$(type -t npm)" ] && eval "$(npm completion)"
```

If you look at ddev/ddev-webserver:20240213_php_8.2_default (or run anything with a recent DDEV head, you'll see that those lines are in /etc/skel/.bashrc.

PR #5796, opened 2024-02-07
docs: add instructions for Grav CMS to quickstart.md

rpkoller:
## The Issue
There were no quickstart guide for Grav CMS

It is adding instructions for a composer based install and a git clone based on to the quickstart.md

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -260,7 +260,7 @@ ddev launch
      ddev exec bin/gpm update -f
      ```
 
-Visit the [Grav Documentation](https://learn.getgrav.org/17 for more information.
+Visit the [Grav Documentation](https://learn.getgrav.org/17 for more information about Grav in general and visit [Local Development with DDEV](https://learn.getgrav.org/17/webservers-hosting/local-development-with-ddev) for more details about the usage of Grav with DDEV.
 
 
 ## Ibexa DXP
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -247,18 +247,18 @@ ddev launch
     ddev exec bin/gpm install admin -y
     ```
 
- !!!tip "How to update?"
-     Upgrade Grave core:
+!!!tip "How to update?"
+    Upgrade Grave core:
 
-     ```bash
-     ddev exec bin/gpm selfupgrade -f
-     ```
+    ```bash
+    ddev exec bin/gpm selfupgrade -f
+    ```
 
-     Update plugins and themes:
+    Update plugins and themes:
 
-     ```bash
-     ddev exec bin/gpm update -f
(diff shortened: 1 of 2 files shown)


rfay:
Would be great to get grav community feedback here. Also, since Grav has long explicitly shown DDEV as their local dev option, link to https://learn.getgrav.org/16/webservers-hosting/local-development-with-ddev

rpkoller:
> Would be great to get grav community feedback here. Also, since Grav has long explicitly shown DDEV as their local dev option, link to https://learn.getgrav.org/16/webservers-hosting/local-development-with-ddev

good point. I wasn't aware of that particular page, but I've extended the last sentence pointing to the page you've raised. thanks!

rpkoller:
the tests are green again now and all styling glitches and typos are fixed. it is ready to be reviewed now.

docs/content/users/quickstart.md
```diff
@@ -194,6 +194,48 @@ ddev launch
...
+
+    ```bash
+    mkdir grav
+    cd grav
+    git clone -b master https://github.com/getgrav/grav.git .
+    ddev config --php-version=8.2 --omit-containers=db
+    ddev start
+    ddev composer install --no-dev -o
+    ddev exec bin/grav install
+    ddev exec bin/gpm install admin -y
```
stasadev, line 221: I think we need to install `require-dev` dependencies, and add `ddev launch`.

```suggestion
    ddev composer install
    ddev exec bin/grav install
    ddev exec bin/gpm install admin -y
    ddev launch
```

stasadev (changes requested):
I tried all the instructions provided, it looks good to me.
I do have a small suggestion regarding `git clone`.

docs/content/users/quickstart.md
rpkoller: good catch with the missing `ddev launch`.  about the composer install line. i've used the `--no-dev -0` because those are the options used in the official grav docs: https://learn.getgrav.org/17/basics/installation (see option 3). therefore i stuck to those in this case. i am fine changing it to the variant you've suggest (i would have used the same) but i thought the grav folks added those for a reason and therefore used that when i created the PR. that is the background. but just wanted to let you know that info before committing the suggestions. so just let me know if i should still go with the changes for the composer line and adding the launch line or if i should only add the launch line?

stasadev: Given that DDEV is a local development tool, the `require-dev` dependencies should always be installed.
I think there's a mistake in their docs (at least they *should* mention that these flags are usually used for production).

rpkoller: yep makes total sense. but just wanted to make sure. committed your proposed suggestion. thank you for taking a look and testing!

stasadev (approved):

PR #5804, opened 2024-02-08
fix: isMagento2App() should detect version 2.4.5+, update test URLs

Morgy93:
## The Issue

* If you run `ddev config --auto` on the latest Magento 2 version, the project type `magento2` is not detected and `php` is used.
* Magento2 test tarballs are unmaintainable and out-of-date

* The check is for a string `https://hackerone.com/magento` in the root SECURITY.md.
(This is odd because that file is often deleted.)
However, the string has been changed to `https://hackerone.com/adobe` since version 2.4.5 as you can review here: https://github.com/magento/magento2/commit/3cb720f9ed01f3d53ba123fba9c048eb3b73f2b7
* Use new ddev/test-magento2 repository for tests.

Install Magento 2 version 2.4.5 or following via CMS Quickstart.
Run `ddev config --auto`.
See that the project type is still `magento2`.

## Automated Testing Overview

The test for `isMagento2App` already exists.

## Release/Deployment Notes

I don't think so.

### DIFF
--- a/pkg/ddevapp/magento.go
+++ b/pkg/ddevapp/magento.go
@@ -23,7 +23,7 @@ func isMagentoApp(app *DdevApp) bool {
 
 // isMagento2App returns true if the app is of type magento2
 func isMagento2App(app *DdevApp) bool {
-	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/magento`)
+	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/adobe`)
 	if err == nil && ism2 {
 		return true
 	}
--- a/pkg/ddevapp/magento.go
+++ b/pkg/ddevapp/magento.go
@@ -23,7 +23,7 @@ func isMagentoApp(app *DdevApp) bool {
 
 // isMagento2App returns true if the app is of type magento2
 func isMagento2App(app *DdevApp) bool {
-	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/adobe`)
+	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/`)
 	if err == nil && ism2 {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


Morgy93:
This is in draft because the test tarball at https://github.com/ddev/ddev/blob/v1.22.7/pkg/ddevapp/ddevapp_test.go#L152 needs to be updated.

And I don't seem to be able to.

rfay:
We're trying to create separate repos for those test setups. Would you be willing to update that if I create a repo and give you privileges on it? The basic idea is to have a full repo which is maintained with code, and create the tarball and put it there.

Examples: 
* https://github.com/ddev/test-shopware6
* https://github.com/ddev/test-laravel
* https://github.com/ddev/test-silverstripe
* https://github.com/ddev/test-drupal
* https://github.com/ddev/test-flask-sayhello

(https://github.com/orgs/ddev/repositories?q=test-&type=all&language=&sort=)

I took the liberty of creating the repo, https://github.com/ddev/test-magento2 and giving you maintainer access to it.

I know this is outside the scope of what you signed up for but it would be wonderful if you were willing to take it on. 

Basically the old way of providing code as tarballs on a random unmaintained repo was hopeless. This way is at least a bit better.

rfay:
I also added you to the project so you can restart tests, and sent you an invite to buildkite so you can restart those tests.

Morgy93:
What's the philosophy for DDEV supported frameworks regarding versions?
Only support latest release or everything that is maintained?

Adobe supports Magento 2.4.4, 2.4.5 and 2.4.6 but the current check would be different for 2.4.4 and 2.4.5/.6.

rfay:
In general we want to support the current stable version. 

There are situations in TYPO3 where we check for more than one situation and handle it, but normally we want to just stay with the most likely thing people are using. isMagento2App() could certainly check more than one thing.

rfay:
And thanks for your very important work on this!

rfay:
@Morgy93 Please review, thanks!

stasadev (approved):
Tested, `ddev config --auto` detects `magento2`.

pkg/ddevapp/magento.go
```diff
@@ -23,7 +23,7 @@ func isMagentoApp(app *DdevApp) bool {
 
 // isMagento2App returns true if the app is of type magento2
 func isMagento2App(app *DdevApp) bool {
-	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/magento`)
+	ism2, err := fileutil.FgrepStringInFile(filepath.Join(app.AppRoot, app.Docroot, "..", "SECURITY.md"), `https://hackerone.com/`)
```
Morgy93, line 26: I am not a fan of looking for the string `https://hackerone.com/` in a `SECURITY.md` file. That is too generic.

If we want to stick with that file, we could search for the string `magento` which is present multiple times in that file: https://github.com/magento/magento2/blob/2.4.6-p4/SECURITY.md?plain=1

rfay: It's your PR! It can use anything reasonable to detect, but should work with older and newer magento2 and be expected to be relatively unique.

rfay:
Followup for a better detection is very welcome!

Issue #5767, opened 2024-02-02
Update default PHP version to 8.2


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

PHP 8.2 is the right default PHP version for DDEV v1.23.0

### Describe your solution

Update it.

PR #5805, opened 2024-02-08
fix: Remove Craft CMS PHP version override

AugustMiller:
## The Issue

We have just tagged Craft 5.0.0-beta.1, and with that, are requiring PHP 8.2. The `craftcms` project type currently uses PHP 8.1, which will not satisfy the requirements.

Our upgrade guide covers switching PHP versions using the `config` command (`ddev config --php-version=8.2`), but we would like for all projects moving forward to use the most recent/default DDEV PHP version (8.2)—with which Craft 4 is also fully compatible.

The current database version (MySQL 8.0) is still our recommended engine, so no changes have been made to that configuration.

The `craftCmsConfigOverrideAction()` function no longer overrides the PHP version used when creating new projects.

Create a new DDEV project:

```bash
mkdir craft-five
cd craft-five
ddev composer create -y --no-scripts craftcms/craft
ddev craft install
```

If Craft doesn't complain about the PHP version (or emit a lower-level language error), you’re all set!

## Automated Testing Overview

No automated tests have been added or updated.

## Release/Deployment Notes

- Craft CMS project types now use DDEV's default PHP version.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -126,7 +126,7 @@ Environment variables will be automatically added to your `.env` file to simplif
     ddev launch
     ```
 
-    Craft CMS projects use PHP 8.1 and MySQL 8.0, by default. You can override these settings during setup with [`config` command flags](./usage/commands.md#config) or after setup via the [configuration files](./configuration/config.md).
+    Craft CMS projects use MySQL 8.0, by default. You can override this setting (and the PHP version) during setup with [`config` command flags](./usage/commands.md#config) or after setup via the [configuration files](./configuration/config.md).
 
     !!!tip "Upgrading or using a generic project type?"
         If you previously set up DDEV in a Craft project using the generic `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
--- a/pkg/ddevapp/apptypes_test.go
+++ b/pkg/ddevapp/apptypes_test.go
@@ -62,7 +62,7 @@ func TestConfigOverrideAction(t *testing.T) {
 	appTypes := map[string]string{
 		nodeps.AppTypeBackdrop:     nodeps.PHPDefault,
 		nodeps.AppTypeCakePHP:      nodeps.PHP83,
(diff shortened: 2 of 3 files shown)


rfay:
Hmm, that line doesn't probably need to be there in the first place, since 8.1 is default.

Also, 8.2 will be the default in DDEV v1.23, coming as next release probably in a couple of months. 
* https://github.com/ddev/ddev/issues/5767

I'm still OK with this, just things to consider. It would also be OK to update the quickstart and add it there, which would make it effective immediately in the "latest" docs.

rfay:
And of course, thanks so much for watching these things and keeping up the maintenance!

AugustMiller:
Noted! I'll talk with the crew and revise or close. We are currently planning on a 6–8 week beta period, so the docs route may be preferable, in the event our timelines don't quite line up. ✌️

rfay:
Still, there's nothing wrong with your change here, does no harm, just requires future maintenance as time moves along and PHP versions move along.  And you do a great job keeping track, so should be fine anyway.

rfay:
PHP 8.2 default:
* https://github.com/ddev/ddev/pull/5835

But I note that Craft has a bunch of overrides that set it to 8.1, so we'll definitely want to update things here. The two ways of doing it would be to let it track the defaults (not do overrides) or just update the overrides.

Do we also want to force `composer_version: 2.7` ?

rfay:
@AugustMiller this needs a restart now, since PHP 8.2 is now default, etc. Do you want to continue with this one or start over again?

rfay:
Closing for now, happy to reopen, happy to continue the conversation here or elsewhere. Definitely want this to be up-to-date for Craft 5 and DDEV v1.23.0

rfay:
We should finish this! The new release is almost ready. Reopening, we'll have to get it done.

AugustMiller:
As we're out of beta now, I've removed the version override so that we can take advantage of DDEV's default, moving forward. This way, when DDEV is ready to move to 8.3 as the default, Craft projects will automatically get the latest version (rather than needlessly locking users to an older release).

Thanks for reviving, Randy! Holler if you need anything else from us. 💞

rfay (approved):
Fixed the broken test, after that it looks fine to me. Thanks!

PR #5812, opened 2024-02-09
test: golangci-lint got pickier, satisfy it

rfay:
## The Issue

golangci-lint just got way more particular about unused function parameters

Pacify the linter


rfay:
Seems to satisfy static validation now.

PR #5813, opened 2024-02-09
feat: generalize HOMEDIR management and respect XDG_CONFIG_HOME, fixes #5807

### LINKED ISSUE #5807: Support XDG base directory specification

I'm frustrated when software such as DDEV doesn't respect the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), leading to a cluttered home directory fully of one-off config directories like `~/.ddev`.

### Describe your solution

Follow the XDG base directory spec. For instance, in the absence of any XDG environment variables, move `~/.ddev` to `~/.config/ddev`. If XDG_CONFIG_HOME is set, move `~/.ddev` to `$XDG_CONFIG_HOME/ddev`.

As a transitional solution (so this doesn't need to be a breaking change), load and use the `~/.ddev` directory if it already exists instead of following XDG_CONFIG_HOME.

Note that there are other XDG directories besides the config home; I haven't investigated whether DDEV can or should make use of those.

### Describe alternatives

_No response_

### Additional context

_No response_

danepowell:
## The Issue

- #5807 

This is my first time touching Go so please "go" easy on me! 😄 Thanks for having great documentation, good code, and an easy build process that made it a pretty easy learning process.

1. Allows to move the global `~/.ddev` directory to `$XDG_CONFIG_HOME/ddev` or predefined OS specific location:

    * `~/.config/ddev` on Linux and WSL2 only

    **Important: if `$XDG_CONFIG_HOME` is set, DDEV will automatically create `$XDG_CONFIG_HOME/ddev` dir, if it doesn't exist.**

2. If there are multiple DDEV global configurations, they are checked with the following priority:

	1. `$XDG_CONFIG_HOME/ddev`
	2. `~/.config/ddev` on Linux and WSL2 only
	3. `~/.ddev`

	The current location of the configuration directory can be checked using `ddev version | grep global-ddev-dir`

3. Switches from `~/.ddev/mutagen_data_directory` to `~/.ddev/.mdd`

4. Adds check for Mutagen socket length. See:

	- https://github.com/mutagen-io/mutagen/issues/453
	- https://github.com/garden-io/garden/issues/4527#issuecomment-1584286590
	- https://unix.stackexchange.com/a/367012

1. Check moving ~/.ddev without using `$XDG_CONFIG_HOME` (on Linux and WSL2):
	1. Move `~/.ddev` to OS specific config folder:
	    * `~/.config/ddev` on Linux and WSL2
	2. Check that old Mutagen folder `~/.ddev_mutagen_data_directory` does exist.
	3. Run `ddev list`
	4. Run `ddev version | grep global-ddev-dir` to confirm that directory is different from `~/.ddev`
	5. Check that `~/.ddev` is not recreated
	6. Run `ddev start`
	7. Old `~/.ddev_mutagen_data_directory` should be deleted.
	8. Move `.ddev` back to `~/.ddev` and confirm that everything works as before

2. Repeat all steps from 1, but this time use `export $XDG_CONFIG_HOME=$HOME/.config && mkdir -p $XDG_CONFIG_HOME/ddev  && mv ~/.ddev $XDG_CONFIG_HOME/ddev`

3. Repeat all steps from 1, but this time use `export $XDG_CONFIG_HOME=$HOME/.config`, and run DDEV commands, the directory `$XDG_CONFIG_HOME/ddev` will be created automatically.

4. Check Mutagen error for really long path to its socket:
	1. `ddev config global --performance-mode=mutagen`
	2. check `ddev mutagen version` output
	3. Set `$XDG_CONFIG_HOME` to some really long path, in result the full path to `$XDG_CONFIG_HOME/ddev/.mdd/daemon/daemon.sock` should be at least 104 characters on macOS or 108 characters on Linux
	6. `mkdir -p $XDG_CONFIG_HOME/ddev`
	7. check `ddev mutagen version` output (the locations should be different).
	8. run `ddev start` and see a warning for too long path

## Automated Testing Overview

Tests that changed `$HOME` are rewritten to use `$XDG_CONFIG_HOME` (except for `TestCreateGlobalDdevDir`, that needs to create a directory).

## Related Issue Link(s)

Fixes #5807 

## Release/Deployment Notes

* This PR doesn't introduce any breaking changes, it adds more options for moving the DDEV global config directory to another location.
* However, `ddev poweroff` is required to get mutagen data into the new directory.

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -192,9 +192,11 @@ fi
 # Make sure we start with mutagen daemon off.
 unset MUTAGEN_DATA_DIRECTORY
 if [ -f ~/.ddev/bin/mutagen -o -f ~/.ddev/bin/mutagen.exe ]; then
+  # This line can be removed when ~/.ddev/.mdd is well established in master, probably July 2024
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.mutagen ~/.ddev/bin/mutagen daemon stop || true
+  # This line can be removed when ~/.ddev/.mdd is well established in master, probably July 2024
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd/ ~/.ddev/bin/mutagen daemon stop || true
 fi
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -192,9 +192,9 @@ fi
 # Make sure we start with mutagen daemon off.
 unset MUTAGEN_DATA_DIRECTORY
 if [ -f ~/.ddev/bin/mutagen -o -f ~/.ddev/bin/mutagen.exe ]; then
-  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
+  MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.mutagen ~/.ddev/bin/mutagen daemon stop || true
-  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
+  MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
 fi
 if command -v killall >/dev/null ; then
   killall mutagen || true
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -192,8 +192,10 @@ fi
 # Make sure we start with mutagen daemon off.
 unset MUTAGEN_DATA_DIRECTORY
 if [ -f ~/.ddev/bin/mutagen -o -f ~/.ddev/bin/mutagen.exe ]; then
+  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.mutagen ~/.ddev/bin/mutagen daemon stop || true
+  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
 fi
 if command -v killall >/dev/null ; then
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -193,10 +193,10 @@ fi
 unset MUTAGEN_DATA_DIRECTORY
 if [ -f ~/.ddev/bin/mutagen -o -f ~/.ddev/bin/mutagen.exe ]; then
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
-  MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
+  MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.mutagen ~/.ddev/bin/mutagen daemon stop || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
-  MUTAGEN_DATA_DIRECTORY=~/.ddev/mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
+  MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd/ ~/.ddev/bin/mutagen daemon stop || true
 fi
 if command -v killall >/dev/null ; then
   killall mutagen || true
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -35,8 +35,7 @@ func TestCustomCommands(t *testing.T) {
 	app, err := ddevapp.NewApp("", false)
 	assert.NoError(err)
 
-	tmpHome, origHome, err := makeTempHomeDir(app, t)
-	require.NoError(t, err)
+	tmpXdgConfigHomeDir, originalMutagenDataDir := testcommon.SetTmpXdgConfigHomeDir(t)
 
 	testdataCustomCommandsDir := filepath.Join(origDir, "testdata", t.Name())
 
@@ -46,24 +45,18 @@ func TestCustomCommands(t *testing.T) {
 		assert.NoError(err)
 		err = app.Stop(true, false)
 		assert.NoError(err)
-		// Stop the Mutagen daemon running in the bogus homedir
-		ddevapp.StopMutagenDaemon()
+		testcommon.CleanupTmpXdgConfigHomeDir(t, tmpXdgConfigHomeDir, originalMutagenDataDir)
 		runTime()
 		app.Type = origType
 		err = app.WriteConfig()
 		assert.NoError(err)
-		_ = os.RemoveAll(tmpHome)
 		_ = fileutil.PurgeDirectory(filepath.Join(site.Dir, ".ddev", "commands"))
 	})
 	// We must start the app before copying commands, so they don't get copied over
 	err = app.Start()
 	require.NoError(t, err)
 
-	// We can't use the standard getGlobalDDevDir here because *our* global hasn't changed.
-	// It's changed via $HOME for the DDEV subprocess
-	err = os.MkdirAll(filepath.Join(tmpHome, ".ddev"), 0755)
-	assert.NoError(err)
-	tmpHomeGlobalCommandsDir := filepath.Join(tmpHome, ".ddev", "commands")
+	tmpHomeGlobalCommandsDir := filepath.Join(globalconfig.GetGlobalDdevDir(), "commands")
 	err = os.RemoveAll(tmpHomeGlobalCommandsDir)
 	assert.NoError(err)
 
@@ -76,7 +69,7 @@ func TestCustomCommands(t *testing.T) {
 	assert.NoError(err)
 
 	// We need to run some assertions outside of the context of a project first
-	err = os.Chdir(tmpHome)
+	err = os.Chdir(tmpXdgConfigHomeDir)
 	require.NoError(t, err)
 
 	// Check that only global host commands with the XXX annotation display here
@@ -530,50 +523,4 @@ func TestNpmYarnCommands(t *testing.T) {
 		err = app.MutagenSyncFlush()
 		require.NoError(t, err)
 	}
-}
-
-// makeTempHomeDir makes a temporary home directory named for the test being executed.
-// Don't forget to run `ddevapp.StopMutagenDaemon()` and `os.RemoveAll(tmpHome)` in the
-// test's cleanup function.
-func makeTempHomeDir(app *ddevapp.DdevApp, t *testing.T) (string, string, error) {
-	// Before changing HOME, make sure that Mutagen is already running if we are using it,
-	// so we don't accidentally start it in the wrong directory
-	err := globalconfig.ReadGlobalConfig()
-	if err != nil {
-		return "", "", err
-	}
-	if globalconfig.DdevGlobalConfig.IsMutagenEnabled() {
-		out, err := exec.RunHostCommand(globalconfig.GetMutagenPath(), "daemon", "start")
-		if err != nil {
-			return "", "", fmt.Errorf("unable to run Mutagen daemon start, out='%s', err=%v", out, err)
-		}
-	}
-
-	// Must be stopped before changing homedir or Mutagen will lose track
-	// of sessions which are also tracked in the homedir.
-	err = app.Stop(true, false)
-	if err != nil {
-		return "", "", err
-	}
-
-	origHome, err := os.UserHomeDir()
-	require.NoError(t, err)
-	if runtime.GOOS == "windows" {
-		origHome = os.Getenv("USERPROFILE")
-	}
-
-	tmpHome := testcommon.CreateTmpDir(t.Name() + "-tempHome")
-
-	// Change the homedir temporarily
-	t.Setenv("HOME", tmpHome)
-	t.Setenv("USERPROFILE", tmpHome)
-	t.Setenv("DDEV_DEBUG", "")
-
-	// Make sure we have the .ddev/bin dir we need
-	err = fileutil.CopyDir(filepath.Join(origHome, ".ddev/bin"), filepath.Join(tmpHome, ".ddev/bin"))
-	if err != nil {
-		return "", "", err
-	}
-
-	return tmpHome, origHome, nil
 }
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -299,10 +299,6 @@ func TestCustomCommands(t *testing.T) {
 		assert.FileExists(filepath.Join(projectCommandsDir, f))
 	}
 
-	// Make sure we haven't accidentally created anything inappropriate in ~/.ddev
-	assert.False(fileutil.FileExists(filepath.Join(tmpHome, ".ddev", ".globalcommands")))
-	assert.False(fileutil.FileExists(filepath.Join(origHome, ".ddev", ".globalcommands")))
-
 	// Make sure that the old launch, mysql, and xdebug commands aren't in the project directory
 	for _, command := range []string{"db/mysql", "host/launch", "web/xdebug"} {
 		cmdPath := app.GetConfigPath(filepath.Join("commands", command))
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -35,7 +35,7 @@ func TestCustomCommands(t *testing.T) {
 	app, err := ddevapp.NewApp("", false)
 	assert.NoError(err)
 
-	tmpXdgConfigHomeDir, originalMutagenDataDir := testcommon.SetTmpXdgConfigHomeDir(t)
+	tmpHomeDir := testcommon.MoveGlobalDdevDir(t)
 
 	testdataCustomCommandsDir := filepath.Join(origDir, "testdata", t.Name())
 
@@ -45,7 +45,7 @@ func TestCustomCommands(t *testing.T) {
 		assert.NoError(err)
 		err = app.Stop(true, false)
 		assert.NoError(err)
-		testcommon.CleanupTmpXdgConfigHomeDir(t, tmpXdgConfigHomeDir, originalMutagenDataDir)
+		testcommon.ResetGlobalDdevDir(t, tmpHomeDir)
 		runTime()
 		app.Type = origType
 		err = app.WriteConfig()
@@ -69,7 +69,7 @@ func TestCustomCommands(t *testing.T) {
 	assert.NoError(err)
 
 	// We need to run some assertions outside of the context of a project first
-	err = os.Chdir(tmpXdgConfigHomeDir)
+	err = os.Chdir(tmpHomeDir)
 	require.NoError(t, err)
 
 	// Check that only global host commands with the XXX annotation display here
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -35,7 +35,7 @@ func TestCustomCommands(t *testing.T) {
 	app, err := ddevapp.NewApp("", false)
 	assert.NoError(err)
 
-	tmpXdgConfigHomeDir := testcommon.MoveGlobalDdevDir(t)
+	tmpXdgConfigHomeDir := testcommon.CopyGlobalDdevDir(t)
 
 	testdataCustomCommandsDir := filepath.Join(origDir, "testdata", t.Name())
 
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -35,7 +35,7 @@ func TestCustomCommands(t *testing.T) {
 	app, err := ddevapp.NewApp("", false)
 	assert.NoError(err)
 
-	tmpHomeDir := testcommon.MoveGlobalDdevDir(t)
+	tmpXdgConfigHomeDir := testcommon.MoveGlobalDdevDir(t)
 
 	testdataCustomCommandsDir := filepath.Join(origDir, "testdata", t.Name())
 
@@ -45,7 +45,7 @@ func TestCustomCommands(t *testing.T) {
 		assert.NoError(err)
 		err = app.Stop(true, false)
 		assert.NoError(err)
-		testcommon.ResetGlobalDdevDir(t, tmpHomeDir)
+		testcommon.ResetGlobalDdevDir(t, tmpXdgConfigHomeDir)
 		runTime()
 		app.Type = origType
 		err = app.WriteConfig()
@@ -69,7 +69,7 @@ func TestCustomCommands(t *testing.T) {
 	assert.NoError(err)
 
 	// We need to run some assertions outside of the context of a project first
-	err = os.Chdir(tmpHomeDir)
+	err = os.Chdir(tmpXdgConfigHomeDir)
 	require.NoError(t, err)
 
 	// Check that only global host commands with the XXX annotation display here
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -202,8 +202,8 @@ Whether [`ddev start`](../usage/commands.md#start) should be interrupted by a fa
 
 Absolute path to the `global_config.yaml` file. Can be changed by [moving](../usage/architecture.md#global-files) the `~/.ddev` global directory.
 
-| Type | Default                          | Usage
-| -- |----------------------------------| --
+| Type | Default | Usage
+| -- | -- | --
 | :octicons-globe-16: global | `$HOME/.ddev/global_config.yaml` | Controlled by `$XDG_CONFIG_HOME` env variable.
 
 ## `hooks`
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -198,6 +198,14 @@ Whether [`ddev start`](../usage/commands.md#start) should be interrupted by a fa
 | -- | -- | --
 | :octicons-file-directory-16: project<br>:octicons-globe-16: global | `false` | Can be `true` or `false`.
 
+## `global_config_path`
+
+Absolute path to the `global_config.yaml` file. Can be changed by [moving](../usage/architecture.md#global-files) the `~/.ddev` global directory.
+
+| Type | Default                          | Usage
+| -- |----------------------------------| --
+| :octicons-globe-16: global | `$HOME/.ddev/global_config.yaml` | Controlled by `$XDG_CONFIG_HOME` env variable.
+
 ## `hooks`
 
 DDEV-specific lifecycle [hooks](hooks.md) to be executed.
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -198,14 +198,6 @@ Whether [`ddev start`](../usage/commands.md#start) should be interrupted by a fa
 | -- | -- | --
 | :octicons-file-directory-16: project<br>:octicons-globe-16: global | `false` | Can be `true` or `false`.
 
-## `global_config_path`
-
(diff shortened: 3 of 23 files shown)


pkg/globalconfig/global_config.go
```diff
@@ -537,25 +537,33 @@ func WriteProjectList(projects map[string]*ProjectInfo) error {
...
 func GetGlobalDdevDir() string {
+	// Check for legacy directory.
 	userHome, err := os.UserHomeDir()
 	if err != nil {
 		logrus.Fatal("Could not get home directory for current user. Is it set?")
 	}
 	ddevDir := filepath.Join(userHome, ".ddev")
 
-	// Create the directory if it is not already present.
+	// If legacy directory doesn't exist, create new one.
```
rfay, line 549: This is creating the .ddev directory, nothing legacy about it. It may be trying to create the parent as a result also...

rfay (commented):
Looking forward to test results. Will require excellent manual testing.

rfay:
Static validation failures are due to upgrade of golanci-lint, will be fixed shortly in 
* https://github.com/ddev/ddev/pull/5812

That should go in at least by tomorrow.

danepowell:
Thanks. I'll circle back to this Monday to clean up tests and any other issues.

rfay:
@danepowell could you please PM me any way you want about some Acquia things? Discord probably best, but email and other contacts are on https://github.com/rfay

rfay:
Looks like TestCreateGlobalDdevDir will need some attention.

When you rebase this the static validation will be fixed.

pkg/globalconfig/global_config.go
danepowell: The language here is ambiguous, sorry. I'll clarify in the comment.

danepowell:
The tests now pass locally (woot! 😄 )

The problem is they inherit certain properties of the host environment, namely the OS config dir. So I have to hardcode `~/Library/Application Support` which will obviously break on a Linux runner. Not sure how to fix that yet.

danepowell:
This should be ready for review. The tests are still a little flaky on my local but I think that's due to some kind of interference between the test fixtures and my host environment. It also seems like some of the failing tests in CI are unrelated to this PR.

rfay:
One thing to know is that since it's your first PR, each push has to be approved for testing. So ping me if you push and need to see the new results. I see your pushes, so will try to make sure I start when it happens. Use `[skip ci]` in commit message if you don't want to run tests on a commit.

GuySartorelli (approved):
This works great! Uses the `~/.ddev/` dir if that exists (which it will for existing installations), but if that dir doesn't exist it'll create (or use if it exists) the `ddev/` dir in the configured config dir.

Can't test for mac or windows, but works great in gitpod and locally on ubuntu.

The failing colima test is probably a red herring, that test likes to fail (@rfay I thought you'd given me access to rerun these but it looks like I can't? I can only rerun buildkite CI it looks like)

GuySartorelli:
Just had a thought - there's a bunch of code comments (and maybe docs?) that refer to `~/.ddev` - these should be updated to avoid confusion.

rfay:
> but works great in gitpod and locally on ubuntu.

One of the problems with Gitpod is that the homedir is not persistent. I think we could actually use this to solve that problem on gitpod! /cc @shaal . /workspace is what persists there, so we could use actually /workspace itself as $XDG_CONFIG_HOME. Huge win!

rfay:
> there's a bunch of code comments (and maybe docs?) that refer to ~/.ddev - these should be updated to avoid confusion.

I think there's a lot of little things that do need to be taken care of, https://github.com/search?q=repo%3Addev%2Fddev%20~%2F.ddev&type=code - most of them are not important but there are a few that are. 

I'm OK with doing this as a followup, but some of those look really important.

The code situations (not comments) absolutely have to be taken care of.

danepowell:
I hadn't considered how many (156, at least 😄 ) hardcoded references to `~/.ddev` we'd need to deal with.

I don't think we actually _have_ to fix any of those immediately. In fact, we might be better off waiting until the next major release, when/if support for `~/.ddev` is actually removed, in order to fix most of those. Otherwise there's a chance we do all of that work and never end up actually deprecating `~/.ddev` for whatever reason.

I'd suggest that the one exception is public-facing docs. We should update those in order to avoid confusion where users go looking for `~/.ddev` where it doesn't exist.

Let me know what you think.

GuySartorelli:
> I don't think we actually have to fix any of those immediately.

I'd strongly disagree with that. Who knows what random bugs will suddenly appear because someone does a fresh install, their config is put in the new place, but the hardcoded references are looking in the wrong place...

rfay:
The ones in Dockerfiles aren't relevant.
list.go looks important
Location of router_compose.yaml is important
Location of mutagen_data_directory is critical
Location of ~/.ddev/bin is critical


I haven't looked through all of them, but it seems to me like we have more to do than originally intended.

rfay:
I think this is a great initiative. The ones that affect scripts are *not* all relevant, and the ones that take place inside containers are not relevant. But the golang ones that are for things other than ~/.ddev (like ~/.mutagen_data_directory) are obviously critical. Of course... moving .mutagen_data_directory into ~/.ddev would make a lot more sense.

Thanks so much for this work, and I hope it works out.

rfay:
Moved back to draft so it can't be accidentally pulled.

stasadev:
Rebased on the master.

This is an interesting PR and I'm going to continue working on it.

I think it's worth making `$XDG_CONFIG_HOME/ddev` optional, since DDEV has been around for a long time and there are many tutorials that depend on `~/.ddev`.

I like the way it is done for [git](https://git-scm.com/docs/git-config#Documentation/git-config.txt---global):

> write to global `~/.gitconfig` file rather than the repository `.git/config`, write to `$XDG_CONFIG_HOME/git/config` file if this file exists and the `~/.gitconfig` file doesn’t.

So, if the user decides to move `~/.ddev` to `~/.config/ddev` (or other equivalent for another OS), we'll let them do it, otherwise we'll leave it as it is. From this point of view, we will not need to change all the documentation.

danepowell:
Sounds good to me, thanks for picking this up!

rfay:
One of the phenomenally awesome things about this is it's going to simplify the tests where we have to do awful things like temporarily changing $HOME. Wow, will that be fantastic.

stasadev:
- Rebased
- Reverted `TestCreateGlobalDdevDir` and removed `TestLegacyGlobalDdevDir`
- Modified `GetGlobalDdevDir` to use config folder for DDEV only if it exists and `~/.ddev` doesn't exist
- Added `TestCreateGlobalDdevDirInUserConfigDir` with test cases for macOS, Windows, Linux
- Updated DDEV architecture page in the documentation

stasadev:
Rebased and moved `~/.ddev_mutagen_data_directory` to `~/.ddev/mutagen_data_directory`.

stasadev:
The tests look good, the only thing I'm not sure about is the previous location for `MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory` - should we do something with it?

Updated OP, ready for review.

stasadev:
Back to draft to make `$XDG_CONFIG_HOME` work on each OS

stasadev:
The problem with the timeout of the Mutagen daemon in the tests turned out to be a problem with `MUTAGEN_DATA_DIRECTORY` length:

- https://github.com/mutagen-io/mutagen/issues/453

```
Error: unable to connect to daemon: connection timed out (is the daemon running?)
```

https://buildkite.com/ddev/macos-orbstack/builds/1259#018f1bb0-7980-4bd0-abe2-45fc29bdfe3b/11685-11737

---

TODO:

- [ ] Find out the maximum length that can be set for `MUTAGEN_DATA_DIRECTORY` on different platforms and let the user know.

rfay:
Wow! Asking in Mutagen slack.

rfay:
Here's the code for MUTAGEN_DATA_DIRECTORY:
https://github.com/mutagen-io/mutagen/blob/84e22936fce1eb60e7b24b4ac18fc08dafc73821/pkg/filesystem/mutagen.go#L73-L129

rfay:
If it helps, I see no reason not to change to use $XDG_CONFIG_HOME/.ddev/.mdd

stasadev:
Okay, I'll see what I can do, I just need to find out why I broke the tests again 😅

stasadev (commented):
The tests are finally working and I have updated the OP with everything that has been done here.

I'm pretty sure there are no breaking changes, but it still needs some thorough manual testing.

I'm not moving this to ready for review as @rfay and I decided that these changes can be pulled later (after v1.23.1 bugfix release).

cmd/ddev/cmd/commands_test.go
```diff
@@ -297,10 +290,6 @@ func TestCustomCommands(t *testing.T) {
 		assert.FileExists(filepath.Join(projectCommandsDir, f))
 	}
 
-	// Make sure we haven't accidentally created anything inappropriate in ~/.ddev
-	assert.False(fileutil.FileExists(filepath.Join(tmpHome, ".ddev", ".globalcommands")))
-	assert.False(fileutil.FileExists(filepath.Join(origHome, ".ddev", ".globalcommands")))
```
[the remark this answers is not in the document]
rfay: The assertion about .globalcommands was to make sure that we hadn't left those laying around when the global commands were redesigned. It's probably OK to lose the assertion now.

rfay:
For the MUTAGEN_DATA_DIRECTORY too long problem, @xenoscopic mentions:
* https://github.com/garden-io/garden/pull/4867
* https://github.com/garden-io/garden/issues/4527

stasadev:
This is an interesting solution for Mutagen in Garden.io, but they have a different structure, they have a separate `MUTAGEN_DATA_DIRECTORY` for each project at `~/.garden/mutagen/[project-name]`. So they use `sha256` for `[project-name]` because it can be too long.

And for DDEV, the user will be able to move the entire global config directory, so I guess it's the user's responsibility to make sure that `$XDG_CONFIG_HOME` is not too deep. I've added a warning just for this case.

P.S.: one more thing - I added a length check for all platforms, including Windows, but I haven't tested it on Windows, so it might not be needed there.

.buildkite/test.sh
```diff
@@ -193,8 +193,10 @@ fi
 unset MUTAGEN_DATA_DIRECTORY
 if [ -f ~/.ddev/bin/mutagen -o -f ~/.ddev/bin/mutagen.exe ]; then
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
```
rfay, line 195:
```suggestion
  # This line can be removed when ~/.ddev/.mdd is well established in master, probably July 2024
  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen sync terminate -a || true
```

.buildkite/test.sh
```diff
@@ -193,8 +193,10 @@ fi
...
+  MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd/ ~/.ddev/bin/mutagen sync terminate -a || true
   MUTAGEN_DATA_DIRECTORY=~/.mutagen ~/.ddev/bin/mutagen daemon stop || true
   MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
```
rfay, line 198:
```suggestion
  # This line can be removed when ~/.ddev/.mdd is well established in master, probably July 2024
  MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory/ ~/.ddev/bin/mutagen daemon stop || true
```

docs/content/users/configuration/config.md
```diff
@@ -198,6 +198,14 @@ Whether [`ddev start`](../usage/commands.md#start) should be interrupted by a fa
 | -- | -- | --
 | :octicons-file-directory-16: project<br>:octicons-globe-16: global | `false` | Can be `true` or `false`.
 
+## `global_config_path`
```
rfay, line 201: Interested to know why this was needed, and why it isn't the *directory* instead of the *file*. project_list.yaml should still be a peer of this in the same directory, true?

If it's controlled by XDG_CONFIG_HOME, why does it have its own config element?

Should this be mostly hidden (maybe it is, haven't gotten that far...)

docs/content/users/install/performance.md
```diff
@@ -150,12 +150,14 @@ Mutagen is enabled by default on Mac and traditional Windows, and it can be disa
...
     * Avoid having Mutagen sync large binaries, which can cause `ddev start` to take a long time. The `.tarballs` directory is automatically excluded, so Mutagen will ignore anything you move there. To see what Mutagen is trying to sync, run `ddev mutagen status -l` in another window.
     * `export DDEV_DEBUG=true` will provide more information about what’s going on with Mutagen.
-    * As of DDEV v1.21.2, DDEV’s Mutagen daemon keeps its data in a DDEV-only `MUTAGEN_DATA_DIRECTORY`, `~/.ddev_mutagen_data_directory`.
-    * DDEV’s private Mutagen binary is installed in `~/.ddev/bin/mutagen`. You can use all the features of Mutagen with `export MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory` and running the Mutagen binary in `~/.ddev/bin/mutagen`, for example:and `~/.ddev/bin/mutagen daemon stop`.
+    * As of DDEV v1.21.2, DDEV’s Mutagen daemon keeps its data in a DDEV-only `MUTAGEN_DATA_DIRECTORY`:
+        * `.mdd` located in `~/.ddev/.mdd` in DDEV v1.23.2+
+        * `.ddev_mutagen_data_directory` located in `~/.ddev_mutagen_data_directory` in DDEV v1.21.2 to v1.23.1
```
rfay, line 155: IMO we confuse matters by talking about this too much. Mostly, in the years since mutagen has been introduced, nobody ever noticed or cared where this was stored.

```suggestion
    * * DDEV’s Mutagen daemon keeps its data in a DDEV-only `MUTAGEN_DATA_DIRECTORY`, by default in `~/.ddev/.mdd`.
```

docs/content/users/install/performance.md
```diff
@@ -150,12 +150,14 @@ Mutagen is enabled by default on Mac and traditional Windows, and it can be disa
...
+    * DDEV’s private Mutagen binary is installed in `~/.ddev/bin/mutagen`. You can use all the features of Mutagen with `export MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd` and running the Mutagen binary in `~/.ddev/bin/mutagen`, for example:
 
         ```bash
         export DDEV_DEBUG=true
-        export MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory
+        export MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd
```
rfay, line 160: Should this be

```suggestion
        export MUTAGEN_DATA_DIRECTORY=${XDG_CONFIG_HOME:-~/}/.ddev/.mdd
```

docs/content/users/usage/architecture.md
```diff
@@ -102,6 +102,27 @@ Files beginning with `.` are hidden because they shouldn’t be fiddled with; mo
 
 There’s only one global `.ddev` directory, which lives in your home directory: `~/.ddev` (`$HOME/.ddev`).
 
+!!!tip "Where is my global `.ddev` config?"
+    Use `ddev config global` (look at `global-config-path`) to check which location is currently used for the `.ddev` global directory.
```
rfay, line 106: Now I see why you have the global config element, but it seems it's not config but reporting?

docs/content/users/usage/architecture.md
```diff
@@ -102,6 +102,27 @@ Files beginning with `.` are hidden because they shouldn’t be fiddled with; mo
...
+
+!!!tip "What if I don't want to clutter up my `$HOME` with a `.ddev` directory?"
```
rfay, line 108: There's a risk here that people won't know that XDG_CONFIG_HOME is a system-wide setting that may affect other applications. Perhaps we should be saying things like "If XDG_CONFIG_HOME is set (with link) then DDEV's global directory will be stored there.

docs/content/users/usage/architecture.md
```diff
@@ -102,6 +102,27 @@ Files beginning with `.` are hidden because they shouldn’t be fiddled with; mo
...
+    `~/.ddev` can be moved to `$XDG_CONFIG_HOME/ddev` directory if `$XDG_CONFIG_HOME` is set:
+
+    ```bash
+    # permanently set environment variable depending on your OS (e.g. in ~/.bashrc):
+    export XDG_CONFIG_HOME="/my-desired-path"
+    # restart the terminal and run:
+    mv ~/.ddev /my-desired-path/ddev
```
rfay, line 115:
```suggestion
    mv ~/.ddev ${XDG_CONFIG_HOME}/ddev
```

docs/content/users/usage/architecture.md
```diff
@@ -102,6 +102,27 @@ Files beginning with `.` are hidden because they shouldn’t be fiddled with; mo
...
+    ```
+
+    Otherwise, if `$XDG_CONFIG_HOME` is not set, `~/.ddev` can be moved to:
```
rfay, line 118: I didn't know about this feature, and this worries me a bit. I was expecting only two places, ~/.ddev and $XDG_CONFIG_HOME/ddev

docs/content/users/usage/architecture.md
```diff
@@ -124,11 +145,15 @@ Again, these files are mostly regenerated on every `ddev start` so it’s best t
 `.gitignore`
 : Prevents files from getting checked in when they shouldn’t be.
 
+`.mdd` (`~/.ddev_mutagen_data_directory` previously)
```
rfay, line 148:
```suggestion
`.mdd` 
```

pkg/ddevapp/config.go
```diff
@@ -722,6 +723,18 @@ func (app *DdevApp) FixObsolete() {
...
 	}
 
+	// Remove old ~/.ddev_mutagen_data_directory directory
+	legacyMutagenDataDir := filepath.Join(homedir.Get(), ".ddev_mutagen_data_directory")
+	if fileutil.IsDirectory(legacyMutagenDataDir) {
+		originalMutagenDataDir := os.Getenv("MUTAGEN_DATA_DIRECTORY")
+		_ = os.Setenv("MUTAGEN_DATA_DIRECTORY", legacyMutagenDataDir)
+		StopMutagenDaemon()
+		_ = os.Setenv("MUTAGEN_DATA_DIRECTORY", originalMutagenDataDir)
+		err := os.RemoveAll(legacyMutagenDataDir)
```
rfay, line 733: This will fail certainly on Windows, maybe elsewhere if the daemon is still running. Have we killed off the daemon here? It's probably OK with the existing logic (tried and failed notification) but worth talking about.

pkg/ddevapp/mutagen.go
```diff
@@ -344,8 +344,8 @@ func mutagenSyncSessionExists(app *DdevApp) (bool, error) {
 func (app *DdevApp) MutagenStatus() (status string, shortResult string, mapResult map[string]interface{}, err error) {
 	syncName := MutagenSyncName(app.Name)
 
-	mutagenDataDirectory := os.Getenv("MUTAGEN_DATA_DIRECTORY")
 	fullJSONResult, err := exec.RunHostCommandSeparateStreams(globalconfig.GetMutagenPath(), "sync", "list", "--template", `{{ json (index . 0) }}`, syncName)
+	mutagenDataDirectory := os.Getenv("MUTAGEN_DATA_DIRECTORY")
```
rfay, line 348: Perhaps for these we should be using DDEV logic to find the directory instead of using the environment variable?

pkg/testcommon/testcommon.go
```diff
@@ -214,6 +215,86 @@ func CreateTmpDir(prefix string) string {
 	return fullPath
 }
 
+// MoveGlobalDdevDir creates a temporary global config directory for DDEV
```
rfay, line 218: Wouldn't this be better as CopyGlobalDdevDir()? We're doing a temp copy for tests, true?

scripts/diagnose_mutagen.sh
```diff
@@ -4,7 +4,7 @@
 # directory where you're having trouble and provide its
 # output in a gist at gist.github.com with any issue you open about mutagen.
 
-export MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen_data_directory
+export MUTAGEN_DATA_DIRECTORY=~/.ddev/.mdd
```
rfay, line 7: I forgot that this script even existed :)

rfay (changes requested):
This is great stuff and amazing work, thanks to both of you. 

* I would prefer that we use only two locations (~/.ddev and $XDG_CONFIG_HOME/ddev) for global config + mutagen
* I suspect that the new global item that tells people where the config is should be something other than a config item, since it's not config, right? Maybe it should be a `ddev debug` item? Not sure where it really should live. 
* Needs solid testing on both macOS and Windows (for the mutagen impact especially)

@danepowell we'd sure appreciate. your review and taking this for a spin.

docs/content/users/install/performance.md
stasadev: Not exactly, because in `XDG_CONFIG_HOME` there is a `ddev` directory, not `.ddev`.
But I understood what needs to be changed here, thanks!

docs/content/users/usage/architecture.md
stasadev: Yes, that's right, I thought it would be nice to have this when users send us `ddev debug test`.

docs/content/users/usage/architecture.md
stasadev: Okay, I will add the link to the XDG documentation.

docs/content/users/usage/architecture.md
stasadev: Okay, it makes sense, I will make it work with XDG_CONFIG_HOME only without and locations by default.

pkg/ddevapp/config.go
stasadev: I thought the daemon would be stopped here because DDEV asks for poweroff during the upgrade.

pkg/testcommon/testcommon.go
stasadev: It is always hard to come up with a good name, thanks!
I will rename the function.

pkg/ddevapp/mutagen.go
stasadev: You are right, I will refactor these.

stasadev (commented):
@rfay, thank you for the review!

> I would prefer that we use only two locations (~/.ddev and $XDG_CONFIG_HOME/ddev) for global config + mutagen

Sounds good with less things to maintain and think about.

> I suspect that the new global item that tells people where the config is should be something other than a config item, since it's not config, right? Maybe it should be a ddev debug item? Not sure where it really should live.

I agree, thanks for pointing me in the right direction, I'll think about `ddev debug something` (or maybe add it to some command).

docs/content/users/usage/architecture.md
danepowell: The XDG base directory spec states:

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

So I do think we should support `~/.config/ddev` by default if the variable is not set.

Things get murky for macOS and Windows, but the directories you've listed here seem to match convention.

stasadev: I previously added `TestGetXdgConfigHomeDir` specifically to test these different locations.

@danepowell, I think in case of Linux you are right.

Linux apps with support for XDG_CONFIG_HOME (see Notes column) to show the current situation with this topic:
https://wiki.archlinux.org/title/XDG_Base_Directory#Supported

Maybe I should just remove the default locations for macOS and Windows.

stasadev:
Changes:

- update the documentation
- remove `global_config_path` from `ddev config global`
- add `ddev mutagen version` command (to show the binary path and `MUTAGEN_DATA_DIRECTORY`)
- add `global-ddev-dir` to `ddev version` output
- use fallback value for empty `$XDG_CONFIG_HOME` for Linux only (`$HOME/.config`)
- rename `MoveGlobalDdevDir` to `CopyGlobalDdevDir`
- run mutagen length check only for Linux and macOS

rfay:
Moved to "Ready for Review" and rebased. We should get this in early in this cycle.

rfay:
Just keeping notes here.

* When I upgraded to this version my mutagen-enabled project was immediately broken due to the change to ~/.ddev/.mdd. I hadn't done `ddev start` or anything at all, and of course `ddev start` would have prompted me for `poweroff`.

rfay:
What am I doing wrong here @stasadev ? When I set XDG_CONFIG_HOME it doesn't seem to be using it at all...

```
rfay@ub-2204:~/workspace/d10$ export XDG_CONFIG_HOME=/tmp
rfay@ub-2204:~/workspace/d10$ ddev version
 ITEM             VALUE
 DDEV version     v1.23.1-47-gec81a1e06
 architecture     amd64
 cgo_enabled      0
 db               ddev/ddev-dbserver-mariadb-10.11:v1.23.1
 ddev-ssh-agent   ddev/ddev-ssh-agent:v1.23.1
 docker           26.1.3
 docker-api       1.45
 docker-compose   v2.27.0
 docker-platform  wsl2-docker-ce
 global-ddev-dir  /home/rfay/.ddev
 mutagen          0.17.2
 os               linux
 router           ddev/ddev-traefik-router:v1.23.1
 web              ddev/ddev-webserver:v1.23.1
```

stasadev:
This feature is completely optional: you need to create `${XDG_CONFIG_HOME}/ddev` directory.

rfay:
Thanks, I suspect DDEV should create the $XDG_CONFIG_HOME/ddev directory if it doesn't exist (but XDG_CONFIG_HOME is set). What do you think?

stasadev:
Yes, I can change the behavior.

I can see that this can get confusing, so it's better to have DDEV handle it.

rfay:
If you disagree, I'm certainly interested in discussion.

stasadev:
At first I thought that we should not create a directory (but only because I had different default folders for all OS's)

Given that `$XDG_CONFIG_HOME` is only used in the Linux world by default and is empty, I agree that we can create a `ddev` directory, but only if `$XDG_CONFIG_HOME` is not empty, so `~/ .config /ddev` will not be used by default.

rfay:
I agree. 

AFAICT, XDG_CONFIG_HOME is not set by default in Ubuntu or Debian.

rfay (changes requested):
I tested this on traditional Windows (git-bash), macOS, and WSL2/Ubuntu, and it worked everywhere. 

Minor changes as discussed include creating $XDG_CONFIG_HOME/ddev if XDG_CONFIG_HOME is set. 

Congrats to both of you on all this work!

docs/content/users/usage/architecture.md
```diff
@@ -102,6 +102,27 @@ Files beginning with `.` are hidden because they shouldn’t be fiddled with; mo
...
+    DDEV can use the `$XDG_CONFIG_HOME` environment variable from [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) to move `~/.ddev` to the `$XDG_CONFIG_HOME/ddev` directory if `$XDG_CONFIG_HOME` is [defined](https://superuser.com/questions/365847/where-should-the-xdg-config-home-variable-be-defined/):
+
+    ```bash
+    # permanently set environment variable depending on your OS:
+    export XDG_CONFIG_HOME="/my-desired-path"
+    # restart the terminal and run:
+    mv ~/.ddev ${XDG_CONFIG_HOME}/ddev
+    ```
+
+    Otherwise (Linux and WSL2 only):
```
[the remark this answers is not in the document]
stasadev: > Otherwise, on Linux/WSL2 only, the default `$HOME/.config/ddev` will be used if `$HOME/.ddev` does not exist:

Not exactly, `$HOME/.config/ddev` will be used only if this directory exists, it doesn't depend on `$HOME/.ddev` existence.

stasadev, line 118:
```suggestion
    Otherwise, on Linux/WSL2 only, the default `$HOME/.config/ddev` can be used when you move the config:
```

stasadev:
Linux tests fail, it seems that `$XDG_CONFIG_HOME` is set by default in GitHub runners and used there:

https://github.com/ddev/ddev/actions/runs/9164941883/job/25197327805?pr=5813#step:15:486

I didn't debug it, but if I'm right, macOS tests will succeed. Not sure if the last change I made is good, if it gives different results for different platforms.

stasadev:
> Linux tests fail, it seems that $XDG_CONFIG_HOME is set by default in GitHub runners and used there

That's true.

I thought about it a bit more and realized that `~/.ddev` check shouldn't be hard-coded, because I could run these tests locally with `$XDG_CONFIG_HOME`, but it should work regardless.

I changed the tests a little bit, this time I didn't check for `~/.ddev` directly, I used a function `GetGlobalDdevDirLocation()`, which has its own test.

rfay:
@danepowell could you please take a look and see if this is what you were hoping for? Thanks!

docs/content/users/usage/architecture.md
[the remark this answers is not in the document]
danepowell: I think Randy has it right. The logic I'd expect, and that's specified by the XDG standard, is that `$HOME/.ddev` is used if and only if it already exists or `XDG_CONFIG_HOME=$HOME`.

This PR looks like it's doing the opposite: using `$HOME/.ddev` by default unless XDG_CONFIG_HOME is set or `$HOME/.config/ddev` already exists.

danepowell:
I'm not sure it's handling the "legacy" directory the way I'd expect (see my previous comment), but otherwise this looks great and I appreciate how much work everyone's contributed. I'd never have gotten this all done on my own.

rfay (approved):
Congratulations on an amazing long haul. Hopefully we can all be using this for the next while so we can make sure there are no unintended side-effects. This is going to be great for the tests, and will make a significant improvement for Gitpod. And it will make @danepowell SO MUCH HAPPIER!

I think the day may come that we'll just move to ~/.config and XDG_CONFIG_HOME by default, now that I've seen how many things are actually storing config there, even on macOS.

stasadev:
Note that the directory order has changed from:

1. `$XDG_CONFIG_HOME/ddev`
2. `~/.config/ddev` on Linux and WSL2 only
3. `~/.ddev`

to

1. `$XDG_CONFIG_HOME/ddev`
2. `~/.ddev`
3. `~/.config/ddev` on Linux and WSL2 only if `~/.ddev` doesn't exist

due to some problems with the enabled Mutagen. It could create the `.ddev` folder in the old location, which was unexpected.

See

- #6239

PR #5814, opened 2024-02-10
fix: follow the zsh naming conventions for completion scripts

rpkoller:
## The Issue

* https://github.com/ddev/ddev/pull/4179 prepended an underscore to the zsh completion script following the zsh naming convention. The PR changed that only on the bump_homebrew.sh file. on .goreleaser.yaml which is used meanwhile that convention wasn't followed.


rfay:
The goreleaser example in https://carlosbecker.com/posts/golang-completions-cobra/ uses `_zsh`

rfay (approved):
This is correct, thanks!

rfay:
Try to train your IDE not to make changes unrelated to the PR, thanks. Something about it is being too aggressive.

Issue #5815, opened 2024-02-10
Support TablePlus application located in setapp directory


peterkraume:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

TablePlus is not only available via download from the vendor website but is  also part of the [SetApp subscription model](https://setapp.com/en/apps/tableplus). In that case, the application is installed in a different directory and thus  the `ddev tableplus` command doesn't work.

### Describe your solution

It's easily possible to check for the alternative directory and thus support both application paths:
Default path is `/Applications/TablePlus.app`
SetApp path is `/Applications/Setapp/TablePlus.app`

rfay:
This was fixed in https://github.com/ddev/ddev/pull/5816

PR #5816, opened 2024-02-10
feat: support TablePlus application located in setapp directory (#5815)

peterkraume:
## The Issue

TablePlus is not only available via download from the vendor website but is  also part of the [SetApp subscription model](https://setapp.com/en/apps/tableplus). In that case, the application is installed in a different directory and thus  the `ddev tableplus` command doesn't work.

This PR checks of tableplus is located in the setapp directory. If not, the default path is used.
Default path is `/Applications/TablePlus.app`
SetApp path is `/Applications/Setapp/TablePlus.app`

`ddev tableplus` works for both, normal users and SetApp users.


rfay (approved):
Rebased to fix the static validation problem. 

This seems good to me, please make sure you've manually tested in both install situations.

PR #5825, opened 2024-02-12
docs: update the section about zsh completions

rpkoller:
## The Issue
The doc for completions in Zsh are outdated.  

the draft is mainly aligned to the instructions on https://docs.brew.sh/Shell-Completion#configuring-completions-in-zsh and as already noted on discord. aside the changes for the zsh section i've made some changes how the page in general is styled. that is only a proposal and could be rolled back at any point. but it is easier to see it in a preview than simply talking about it.

Review these changes in https://ddev--5825.org.readthedocs.build/en/5825/users/install/shell-completion/

## Summary by CodeRabbit

- **Documentation**
	- Enhanced the guide for setting up shell completions for DDEV, now including detailed instructions for Zsh with Homebrew, addressing caching issues, and manual deployment on Linux.
	- Added guidance for Fish, Git Bash, and PowerShell completions, expanding support for various shells and operating systems.

### DIFF
--- a/docs/content/users/install/shell-completion.md
+++ b/docs/content/users/install/shell-completion.md
@@ -4,64 +4,87 @@ Most people like to have shell completion on the command line. In other words, w
 
 Shells like Bash and zsh need help to do this though, they have to know what the options are. DDEV provides the necessary hint scripts, and if you use Homebrew, they get installed automatically. But if you use oh-my-zsh, for example, you may have to manually install the hint script.
 
-=== "macOS Bash + Homebrew"
 
-    ## macOS Bash + Homebrew
+## macOS Bash + Homebrew
 
-    The easiest way to use Bash completion on macOS is install it with Homebrew. `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file:
+The easiest way to use Bash completion on macOS is install it with Homebrew. `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file:
 
+```
+[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+```
+
+!!!note "Bash profile"
+    You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
+
+Link completions by running `brew completions link`.
+
+When you install DDEV via Homebrew, each new release will automatically get a refreshed completions script.
+
+## MacOS Zsh + Homebrew
+
+To make Homebrew completions available in Zsh the Homebrew-managed path `zsh/site-functions` has to be added to the FPATH. There are two scenarios:
+
+=== "Plain macOS Zsh"
+
+    Add the following block to your `~/.zshrc` file:
+
+    ```bash
+    if type brew &>/dev/null
+    then
+      FPATH="$(brew --prefix)/share/zsh/site-functions:${FPATH}"
+      autoload -Uz compinit
+      compinit
+    fi
     ```
-    [[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+
+    Note that the sourcing of the FPATH has to happen before Zsh's completion facility is initialized with `compinit`.
+
+=== "macOS Zsh with Oh-My-Zsh"
+
+    Oh-My-Zsh is calling `compinit` for you when `oh-my-zsh.sh` is sourced. Instead of adding the block that was necessary for `Plain macOS Zsh` place the following line right before `oh-my-zsh.sh` is sourced:
+
+    ```bash
+    FPATH="$(brew --prefix)/share/zsh/site-functions:${FPATH}"
     ```
 
-    !!!note "Bash profile"
-        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
+To avoid any potential caching issue forcibly rebuild the `.zcompdump` file, which is the index for Zsh completions:
 
-    Link completions by running `brew completions link`.
+```bash
+rm -f ~/.zcompdump; compinit
+```
 
-    When you install DDEV via Homebrew, each new release will automatically get a refreshed completions script.
+In case you run into any `zsh compinit: insecure directories` warnings simply run:
 
-=== "Bash/Zsh/Fish on Linux"
+```bash
+chmod -R go-w "$(brew --prefix)/share"
+```
 
-    ## Bash/Zsh/Fish on Linux
+Otherwise, you can extract the zsh completions file (`ddev_zsh_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
 
-    On Debian and Yum based systems, if you installed DDEV using `apt install ddev`, the `bash`, `zsh`, and `fish` completions should be automatically installed at `/usr/share/bash-completion/completions/ddev`, `/usr/share/zsh/vendor-completions/_ddev` and `/usr/share/fish/completions/ddev.fish` respectively, and the `bash` completions should be automatically installed at `/usr/share/bash-completion/completions/bash`.
 
-    Otherwise, you can download the completion files for manual installation as described [below](#tar-archive-of-completion-scripts-for-manual-deployment). Every Linux distro requires a different manual installation technique. On Debian/Ubuntu, you could deploy the `ddev_bash_completion.sh` script where it needs to be by running `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`.
+## Bash/Zsh/Fish on Linux
 
-=== "Oh-My-Zsh"
+On Debian and Yum based systems, if you installed DDEV using `apt install ddev`, the `bash`, `zsh`, and `fish` completions should be automatically installed at `/usr/share/bash-completion/completions/ddev`, `/usr/share/zsh/vendor-completions/_ddev` and `/usr/share/fish/completions/ddev.fish` respectively, and the `bash` completions should be automatically installed at `/usr/share/bash-completion/completions/bash`.
 
-    ## Oh-My-Zsh
+Otherwise, you can download the completion files for manual installation as described [below](#tar-archive-of-completion-scripts-for-manual-deployment). Every Linux distro requires a different manual installation technique. On Debian/Ubuntu, you could deploy the `ddev_bash_completion.sh` script where it needs to be by running `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`.
 
-    If you installed zsh using Homebrew, DDEV’s completions should be automatically installed when you install DDEV using `brew install ddev/ddev/ddev`.
+## Fish
 
-    Otherwise, you can extract the zsh completions file (`ddev_zsh_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
-    
-    Oh-My-Zsh may be set up very differently in different places, so you’ll need to put `ddev_bash_completion.sh` where it belongs. `echo $fpath` will show you the places that it’s most likely to belong. To install in `~/.oh-my-zsh/completions` (an obvious choice), you can run `mkdir -p ~/.oh-my-zsh/completions && cp ddev_zsh_completion.sh ~/.oh-my-zsh/completions/_ddev`, then `autoload -Uz compinit && compinit`.
+`fish` shell completions are automatically installed at `/usr/local/share/fish/vendor_completions.d/ddev_fish_completion.sh` when you install DDEV via Homebrew.
 
-=== "Fish"
+If you installed `fish` without Homebrew, you can extract the fish completions (`ddev_fish_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
 
-    ## Fish
+## Git Bash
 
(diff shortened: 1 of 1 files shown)


rfay:
One thing we haven't figured out yet is why `brew info zsh-completions` says this:

```
  if type brew &>/dev/null; then
    FPATH=$(brew --prefix)/share/zsh-completions:$FPATH

    autoload -Uz compinit
    compinit
  fi
```

but instead of putting things there, homebrew installs ddev's completions in share/zsh/site-functions

rpkoller:
no need to put both paths in. if you do: 
```
$> echo $(brew --prefix)
/opt/homebrew
```
and when a formula gets installed symlinks are created. if the formula has completions symlinks are placed in `(brew --prefix)/share/zsh-completions`. that way an installed formula is contained within the cellar. everything also in other directories, the touch points with the outer world, are via symlinks. so i dont see a need to add both paths. in particular with the zsh completion behaving correctly again.

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-Shells like Bash and zsh need help to do this though, they have to know what the options are. DDEV provides the necessary hint scripts, and if you use Homebrew, they get installed automatically. But if you use oh-my-zsh, for example, you may have to manually install the hint script.
+Shells like Bash and Zsh need help to do this though, they have to know what the options are. DDEV provides the necessary hint scripts, and if you use Homebrew, they get installed automatically.
 
-=== "macOS Bash + Homebrew"
+## macOS Bash + Homebrew
 
-    ## macOS Bash + Homebrew
+The easiest way to use Bash completion on macOS is install it with Homebrew. `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file:
```
rfay, line 9:
```suggestion
The easiest way to use Bash completion on macOS is install it with Homebrew ([docs](https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash)). `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file (or if that doesn't exist, to your `~/.profile`:
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    The easiest way to use Bash completion on macOS is install it with Homebrew. `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file:
+```
+[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+```
 
-    ```
-    [[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+!!!note "Bash profile"
+    You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
```
rfay, line 16:
```suggestion
    You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately in your current terminal window.
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+Link completions by running `brew completions link`.
```
rfay, line 18: This seems to be a one-time configuration, and people would always want it. This also seems to affect all types of completions. I imagine we should be telling people at the beginning of each section to do this.

> Shell completions for external Homebrew commands are not automatically installed. To opt-in to using completions for external commands (if provided), they need to be linked to HOMEBREW_PREFIX by running brew completions link.

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+When you install DDEV via Homebrew, each new release will automatically get a refreshed completions script.
```
rfay, line 20: I don't think we need to say this?

```suggestion
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+## macOS Zsh + Homebrew
```
rfay, line 22:
```suggestion
## macOS Zsh with Homebrew
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+To make Homebrew completions available in Zsh the Homebrew-managed path `zsh/site-functions` has to be added to the FPATH. There are two scenarios:
```
rfay, line 24:
```suggestion
To make Homebrew completions available in Zsh the Homebrew-managed path `zsh/site-functions` has to be added to the FPATH. There are two scenarios, plain zsh and Oh-My-Zsh, see tabs below:
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+    then
+      FPATH="$(brew --prefix)/share/zsh/site-functions:${FPATH}"
+      autoload -Uz compinit
+      compinit
+    fi
     ```
 
-    !!!note "Bash profile"
-        You must add the include to your `.bash_profile` or `.profile` or nothing will work. Use `source ~/.bash_profile` or `source ~/.profile` to make it take effect immediately.
+    Note that the sourcing of the FPATH has to happen before the Zsh completion index is initialized with `compinit`.
```
rfay, line 39:
```suggestion
    Note that the updating of the FPATH variable has to happen before the Zsh completion index is initialized with `compinit`.
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    Link completions by running `brew completions link`.
+=== "macOS Zsh with Oh My Zsh"
 
-    When you install DDEV via Homebrew, each new release will automatically get a refreshed completions script.
+    Oh-My-Zsh is calling `compinit` for you when `oh-my-zsh.sh` is sourced. Instead of adding the block that was necessary for `Plain macOS Zsh` place the following line right before `oh-my-zsh.sh` is sourced:
```
rfay, line 43:
```suggestion
    Oh-My-Zsh is calling `compinit` for you when `oh-my-zsh.sh` is sourced. Instead of adding the block that was necessary for  simple `macOS Zsh` place the following line right before `oh-my-zsh.sh` is sourced:
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+    ```bash
+    FPATH="$(brew --prefix)/share/zsh/site-functions:${FPATH}"
+    ```
 
-=== "Bash/Zsh/Fish on Linux"
+To avoid any potential caching issue forcibly rebuild the `.zcompdump` file, which is the index for Zsh completions:
```
rfay, line 49:
```suggestion
To avoid any potential caching issue remove and rebuild the `.zcompdump` file, which is the index for Zsh completions:
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    ## Bash/Zsh/Fish on Linux
+```bash
+rm -f ~/.zcompdump; compinit
+```
 
-    On Debian and Yum based systems, if you installed DDEV using `apt install ddev`, the `bash`, `zsh`, and `fish` completions should be automatically installed at `/usr/share/bash-completion/completions/ddev`, `/usr/share/zsh/vendor-completions/_ddev` and `/usr/share/fish/completions/ddev.fish` respectively, and the `bash` completions should be automatically installed at `/usr/share/bash-completion/completions/bash`.
+In case you run into any `zsh compinit: insecure directories` warnings, simply run:
```
rfay, line 55:
```suggestion
In case you run into any `zsh compinit: insecure directories` warnings, run:
```

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    Otherwise, you can download the completion files for manual installation as described [below](#tar-archive-of-completion-scripts-for-manual-deployment). Every Linux distro requires a different manual installation technique. On Debian/Ubuntu, you could deploy the `ddev_bash_completion.sh` script where it needs to be by running `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`.
+```bash
+chmod -R go-w "$(brew --prefix)/share"
```
rfay, line 58: I actually think we should just tell them to do this in the first place.

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+```
 
-=== "Oh-My-Zsh"
+Otherwise, you can extract the Zsh completions file (`ddev_zsh_completion.sh`) from the tar archive of completion scripts included with each release. Read more in the [tar archive of completion scripts for manual deployment](#tar-archive-of-completion-scripts-for-manual-deployment) section.
```
rfay, line 61: I think this paragraph belongs elsewhere, wherever we describe all the many other possibililties, rather than here.

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    Otherwise, you can extract the zsh completions file (`ddev_zsh_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
-    
-    Oh-My-Zsh may be set up very differently in different places, so you’ll need to put `ddev_bash_completion.sh` where it belongs. `echo $fpath` will show you the places that it’s most likely to belong. To install in `~/.oh-my-zsh/completions` (an obvious choice), you can run `mkdir -p ~/.oh-my-zsh/completions && cp ddev_zsh_completion.sh ~/.oh-my-zsh/completions/_ddev`, then `autoload -Uz compinit && compinit`.
+Otherwise, you can download the completion files for manual installation as described [below](#tar-archive-of-completion-scripts-for-manual-deployment). Every Linux distro requires a different manual installation technique. On Debian/Ubuntu, you could deploy the `ddev_bash_completion.sh` script where it needs to be by running `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`.
 
-=== "Fish"
+## Fish
```
rfay, line 69: Isn't this redundant after we just had Bash/Zsh/Fish on Linux above?

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
 
-    ## Fish
+`fish` shell completions are automatically installed at `/usr/local/share/fish/vendor_completions.d/ddev_fish_completion.sh` when you install DDEV via Homebrew.
 
-    `fish` shell completions are automatically installed at `/usr/local/share/fish/vendor_completions.d/ddev_fish_completion.sh` when you install DDEV via Homebrew.
-    
-    If you installed `fish` without Homebrew, you can extract the fish completions (`ddev_fish_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
+If you installed `fish` without Homebrew, you can extract the fish completions (`ddev_fish_completion.sh`) from the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
```
rfay, line 73: If this section is about homebrew and fish, and if it's about macOS, let's label it as such.

rfay (changes requested):
Thanks so much for all this great work. Please note which things you actually tested and which you didn't. 

I'll try to test a few more permutations, including Linux. Not sure I'll get back to git-bash and PS

docs/content/users/install/shell-completion.md
```diff
@@ -2,66 +2,87 @@
...
+
+When you install DDEV via Homebrew, each new release will automatically get a refreshed completions script.
+
+## macOS Zsh + Homebrew
+
+To make Homebrew completions available in Zsh the Homebrew-managed path `zsh/site-functions` has to be added to the FPATH. There are two scenarios:
+
+=== "macOS Zsh"
+
+    Add the following block to your `~/.zshrc` file:
```
rfay, line 28:
```suggestion
    Add the following block to your `~/.zshrc` file (see [docs](https://docs.brew.sh/Shell-Completion#configuring-completions-in-zsh)):
```

rfay:
Another random worry that we don't actually have to deal with, or maybe we want to put one note in somewhere: If you use homebrew on Linux, it works like macOS with homebrew.

docs/content/users/install/shell-completion.md
rpkoller: I would vote for sticking with my version for two reasons. First i would avoid directional language like "see tabs below". That is something that should be avoided if possible. And second I tried to avoid "plain" for the title plus the naming for both is different. It should be Zsh and Oh My Zsh according to each of the upstream repos.

docs/content/users/install/shell-completion.md
rpkoller: I will apply that in a follow up that i draft locally and commit after committing your suggestions here

docs/content/users/install/shell-completion.md
rpkoller: technically it has nothing to do with macos zsh with homebrew. i've already striked a part but kept that part. but better do a clear separation of concern here.

docs/content/users/install/shell-completion.md
rpkoller: as mentioned on discord the initial pr was intended to only focus on the zsh part and do the other shells in a follow up. and i agree also about that. there has some more untwining happen.

docs/content/users/install/shell-completion.md
rpkoller: On second thought i would even strike `There are two scenarios:`

docs/content/users/install/shell-completion.md
rpkoller: at first i thought sourcing would be the more appropriate and clear term but i guess updating is fine as well and in particular easier to understand.

docs/content/users/install/shell-completion.md
rpkoller: hm on second thought. i think the current order is correct. it is a good thing to reset the index and in case the user runs in the permission issue that would be the next step. so probably the correct order.

rfay:
@coderabbit ignore

rfay:
@coderabbit pause

PR #5834, opened 2024-02-13
docs: find all DDEV binaries using `which -a ddev`

stasadev:
## The Issue

Multiple `ddev` can be in `PATH`.
And the older `ddev` may have a higher priority, so users wonder why their `ddev` is not "updating".

Adds the `which -a ddev` technique in several places in the documentation. My priority is that when you type `old ddev` in the search, it will be shown.
Uses `which -a ddev` in `ddev self-upgrade`.

Run `ddev debug fix-commands`.
Run `ddev self-upgrade` with one `ddev` in `PATH`, repeat with multiple `ddev` in `PATH`.

### DIFF
--- a/docs/content/users/usage/faq.md
+++ b/docs/content/users/usage/faq.md
@@ -186,7 +186,7 @@ $ which -a ddev
 /usr/bin/ddev
 ```
 
-You can check each binary version by its full path (`/usr/bin/ddev version`) to find old versions. Remove them preferably in the same way you installed them, i.e. `/home/linuxbrew/.linuxbrew/bin/ddev` should be removed with Homebrew.
+You can check each binary version by its full path (`/usr/bin/ddev version`) to find old versions. Remove them preferably in the same way you installed them, i.e. `/home/linuxbrew/.linuxbrew/bin/ddev` should be removed with Homebrew: `brew uninstall ddev`.
 
 ### How can I back up or restore all project databases?
 
--- a/docs/content/users/usage/faq.md
+++ b/docs/content/users/usage/faq.md
@@ -176,6 +176,18 @@ If you’re using Homebrew, first run `brew unlink ddev` to get rid of the versi
 3. On Debian/Ubuntu/WSL2 with DDEV installed via apt, you can run `sudo apt update && sudo apt install ddev=<version>`, for example `sudo apt install ddev=1.21.5`.
 4. If you want the very latest, unreleased version of DDEV, run `brew unlink ddev && brew install ddev/ddev/ddev --HEAD`.
 
+### Why do I have an old DDEV?
+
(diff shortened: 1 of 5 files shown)


docs/content/users/usage/faq.md
```diff
@@ -176,6 +176,18 @@ If you’re using Homebrew, first run `brew unlink ddev` to get rid of the versi
...
+
+You may have installed DDEV several times using different techniques. Use `which -a ddev` to find all installed binaries. For example, you could install a DDEV in WSL2 with Homebrew, forget about it for a while, and then install it again with `apt`:
+
+```bash
+$ which -a ddev
+/home/linuxbrew/.linuxbrew/bin/ddev
+/usr/bin/ddev
+```
+
+You can check each binary version by its full path (`/usr/bin/ddev version`) to find old versions. Remove them preferably in the same way you installed them, i.e. `/home/linuxbrew/.linuxbrew/bin/ddev` should be removed with Homebrew.
```
rfay, line 189:
```suggestion
You can check each binary version by its full path (`/usr/bin/ddev version`) to find old versions. Remove them preferably in the same way you installed them, i.e. `/home/linuxbrew/.linuxbrew/bin/ddev` should be removed with Homebrew: `brew uninstall ddev`.
```

docs/content/users/usage/troubleshooting.md
```diff
@@ -246,6 +246,10 @@ docker rmi -f $(docker images -q)
 
 You should then be able to start your DDEV machine.
 
+## `ddev version` Shows the old version
```
rfay, line 249:
```suggestion
## `ddev --version` shows an old version
```

`ddev --version` is slightly preferable because there's less noise going on.

docs/content/users/usage/troubleshooting.md
```diff
@@ -246,6 +246,10 @@ docker rmi -f $(docker images -q)
...
+
+If you have installed the latest version of DDEV, and when you check the actual version with `ddev version`, it shows an older version, please refer to [Why do I have an old DDEV?](./faq.md#why-do-i-have-an-old-ddev)
```
rfay, line 251:
```suggestion
If you have installed the latest version of DDEV, but when you check the actual version with `ddev --version`, it shows an older version, please refer to [Why do I have an old DDEV?](./faq.md#why-do-i-have-an-old-ddev)
```

rfay (changes requested):
Fantastic, thanks. Minor suggestions.

docs/content/users/usage/faq.md
stasadev, line 189:
```suggestion
You can check each binary version by its full path (`/usr/bin/ddev --version`) to find old versions. Remove them preferably in the same way you installed them, i.e. `/home/linuxbrew/.linuxbrew/bin/ddev` should be removed with Homebrew.
```

rfay: Thanks

tyler36:
Thats's great! I didn't know about `-a` .

PR #5837, opened 2024-02-14
docs: make the shell completion page more consistent

rpkoller:
## The Issue
Initial draft to make the shell completion sections more consistent among each other as well as more inline with https://docs.brew.sh/Shell-Completion . Not finished yet. Git Bash, PowerShell and the Tar archive section still need work. But feedback about bash and fish is welcome (as well as how consistent it is with the previous zsh changes)

### DIFF
--- a/docs/content/users/install/shell-completion.md
+++ b/docs/content/users/install/shell-completion.md
@@ -2,7 +2,7 @@
 
 Most people like to have shell completion on the command line. In other words, when you're typing a command, you can hit `<TAB>` and the shell will show you what the options are. For example, if you type `ddev <TAB>`, you'll see all the possible commands. `ddev debug <TAB>` will show you the options for the command. And `ddev list -<TAB>` will show you all the flags available for [`ddev list`](../usage/commands.md#list).
 
-Shells like Bash and Zsh need help to do this though, they have to know what the options are. DDEV provides the necessary hint scripts, and if you use Homebrew, they get installed automatically.
+Shells like Bash and Zsh need help to do this though, they have to know what the options are, and DDEV provides the necessary hint scripts.
 
 ## macOS Bash with Homebrew
 
--- a/docs/content/users/install/shell-completion.md
+++ b/docs/content/users/install/shell-completion.md
@@ -137,7 +137,7 @@ On Debian and Yum based systems, if you installed DDEV using `apt install ddev`,
 
 Otherwise, you can download the completion files for manual installation as described [below](#tar-archive-of-completion-scripts-for-manual-deployment). Every Linux distro requires a different manual installation technique. On Debian/Ubuntu, you could deploy the `ddev_bash_completion.sh` script where it needs to be by running `sudo mkdir -p /usr/share/bash-completion/completions && sudo cp ddev_bash_completion.sh /usr/share/bash-completion/completions/ddev`.
 
-## Git Bash
+## Git Bash on Microsoft Windows
 
 Git Bash completions (`ddev_bash_completion.sh`) are provided in the tar archive of completion scripts included with each release. See [below](#tar-archive-of-completion-scripts-for-manual-deployment).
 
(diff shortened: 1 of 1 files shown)


rfay:
Rebased

rfay:
Rebased

rfay:
Rebased

docs/content/users/install/shell-completion.md
```diff
@@ -2,22 +2,65 @@
...
+To make Homebrew completions available in Bash the Homebrew-managed path, either `etc/profile.d` or `etc/bash_completion.de` has to be sourced.
 
-```
-[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+=== "macOS Bash"
+
+     Add the following block to your `~/.bashprofile` or if it doesn't exist to your `~/.profile` file (see [docs](https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash)):
+
+    ```bash
+    if type brew &>/dev/null
```
rfay, line 16: Where did this come from and why is it required? Why the for loop? 

Why would /etc/profile.d not have the bash_completion.sh?

docs/content/users/install/shell-completion.md
```diff
@@ -2,22 +2,65 @@
...
+Shells like Bash and Zsh need help to do this though, they have to know what the options are, and DDEV provides the necessary hint scripts.
 
 ## macOS Bash with Homebrew
 
-The easiest way to use Bash completion on macOS is install it with Homebrew ([docs](https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash)). `brew install bash-completion`. When you install it though, it will warn you with something like this, which **may vary on your system**. Add the following line to your `~/.bash_profile` file (or if that doesn't exist, to your `~/.profile`:
+To make Homebrew completions available in Bash the Homebrew-managed path, either `etc/profile.d` or `etc/bash_completion.de` has to be sourced.
 
-```
-[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
+=== "macOS Bash"
```
rfay, line 11: Why do we have a "macOS Bash" and a "macOS Bash with Bash Completion"? 

image

rfay (changes requested):
I'm afraid this is not an improvement over what we currently have. Maybe it's trying to do too many things?

If this is just about consistency, a new PR could address that. I wasn't able to sort out what this was actually trying to do. 

What we want is for people to be able to quickly and easily get completion working. On macOS that means using homebrew for bash completion.

rfay:
Closing this for now. Thanks for all your work on all these things!

rpkoller:
> I'm afraid this is not an improvement over what we currently have. Maybe it's trying to do too many things?
> 
> If this is just about consistency, a new PR could address that. I wasn't able to sort out what this was actually trying to do.
> 
> What we want is for people to be able to quickly and easily get completion working. On macOS that means using homebrew for bash completion.

not about consistency at all. i was trying to improve the usability of that page. not requiring stuff like bash completion or oh my zsh. save the user the time to read through homebrew instructions and other sources. and that is what is accomplished with the changes. just havent had the time to return and finish the missing  manual step at the end.

docs/content/users/install/shell-completion.md
rpkoller: that is from the homebrew docs. that way you dont need the bash completion as stated in the next tab.

docs/content/users/install/shell-completion.md
rpkoller: for people that dont want to install bash completion for now (like me back then) you have the installation step taken from homebrew and for those who want more those can install bash completion. more or less the same case like for zsh and oh my zsh.

PR #5842, opened 2024-02-14
docs: use PHP 8.3 for the TYPO3 Quickstart

jonaseberle:
The current stable requires PHP 8.2+ (making the current snippet fail) and v11+ supports PHP 8.3 so it seems like a good default.

## The Issue
PHP 8.1 is not supported by the latest stable TYPO3 version (v13.0.1) which gets installed by default
Follow the steps from the paragraph.

## Automated Testing Overview
none

## Related Issue Link(s)
none
## Release/Deployment Notes
none


jonaseberle:
You mean a default PHP version for that project type? 
Hmm, not sure. I don't think it is needed.

PR #5787, opened 2024-02-05
feat: replace fsouza/go-dockerclient with docker/docker/client, fixes #3460

stasadev:
## The Issue

- #3460

## How This PR Solves The Issue

Uses `docker/docker/client` instead of `fsouza/go-dockerclient`.

TODO:

- [x] pass all tests.
- [ ] move all `github.com/docker/docker/...` imports to `pkg/dockerutil/dockerutils.go`.
- [ ] deal with Docker context workarounds (maybe not in this PR).

PR #5702, opened 2024-01-17
feat: allow global host commands to be run from anywhere, for #4416

GuySartorelli:
## The Issue

Custom global host commands currently only be run inside a project directory - even if they don't need a project in order to do their stuff.

A great example of this is `ddev self-upgrade` which should be able to run from anywhere - but currently it has to be run from inside a project directory.

## How This PR Solves The Issue

This PR adds a new `CanRunGlobally` annotation. If custom global host commands set this annotation to `true`, they will be added to the cobra command set, and can be run from anywhere, even outside project directories.

## Manual Testing Instructions

1. Run `ddev help` - you should see the `self-upgrade` command there.
1. Run `ddev self-upgrade` - it should work from anywhere, even outside a project directory.
1. Add `## CanRunGlobally: true` to a custom global host command, then perform the first two steps again, but with your custom command rather than `self-upgrade`.

## Automated Testing Overview

I've added checks for the following the following:
- Check a global host command with `CanRunGlobally: true` shows and can be run from arbitrary working directories
- Check a global host command with `CanRunGlobally: true` still shows and can be run from inside a project directory
- Check commands without `CanRunGlobally: true` do not show from arbitrary working directories

This is my first time really looking at how to add tests and I've gone in with a fairly light touch - if you want more test coverage I'll be happy to add it.

## Related Issue Link(s)

- https://github.com/ddev/ddev/issues/4416

**NOTE:** This PR will be one of several iterative features that together will fulfil the requirements set forth in that issue. The issue should not be closed until all of its requirements are fulfilled.

## Release/Deployment Notes

Anyone who maintains a DDEV extension or who has some custom commands in projects they manage should review their custom global host commands, and add the new `CanRunGlobally` annotation to any commands that should be allowed to run even outside of project directories.

PR #5848, opened 2024-02-14
fix: do not create project network programmatically on `ddev start`, fixes #5810

### LINKED ISSUE #5810: Let docker-compose manage the Docker project network

```
```
</details>         

### Expected Behavior

Custom `docker-compose.yaml` configurations should work for the Docker project network.

### Actual Behavior

From [Discord](https://discord.com/channels/664580571770388500/1205594721330659349/1205594721330659349):

> When setting the default.driver_opts.'com.docker.network.driver.mtu' ... it seems to not carry though

It doesn't work because of

- #5508

It affects DDEV v1.22.5, v1.22.6, v1.22.7

It's better to partially revert #5508 (only the part where the Docker project network is pre-created, duplicates will be checked anyway) to allow overrides for the project network configs.

Additional note: Docker v25+ doesn't allow to create duplicate networks at all.

### Steps To Reproduce

Create `docker-compose.network.yaml` with:
```yaml
networks:
  default:
    driver_opts:
      com.docker.network.driver.mtu: 1440
```
And inspect the Docker project network for this value:
```
docker inspect ddev-project-name_default
```

### Anything else?

_No response_

stasadev:
## The Issue

- #5810

This PR may return this error during `ddev start`, see #5508:

```
Error response from daemon: network ddev-your-project_default is ambiguous (16 matches found on name)
```

But I really doubt it, because we have a test for this behavior, and after the migration from `fsouza/go-dockerclient` to `docker/docker/client` in #5787, I had to rewrite the `networks_test.go`, because neither Docker 24.0.7 nor Docker 25.0.3 could create a duplicate network.

https://github.com/ddev/ddev/blob/cbb96e3a666d33ce7fb7053cfdc7f2e31c1ad0a0/cmd/ddev/cmd/networks_test.go#L44-L53

It only uses a check for network duplicates (which doesn't hurt at all), given that in Docker v25+ you can no longer have duplicates.

This allows users to create additional configurations for project networks in custom `docker-compose.yaml` files.

```
echo "networks:
  default:
    driver_opts:
      com.docker.network.driver.mtu: 1440" > .ddev/docker-compose.network.yaml

ddev restart
```

And inspect the Docker project network for this `mtu` value:
```
docker inspect ddev-$(ddev describe --json-output | jq -r .raw.name)_default | grep -C1 mtu
```
Should show this:
```json
"Options": {
    "com.docker.network.driver.mtu": "1440"
}
```

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -253,7 +253,6 @@ networks:
   ddev_default:
     name: ddev_default
     external: true
-  {{/* the default project network is created programmatically, see EnsureProjectNetwork() */}}
   default:
     name: ${COMPOSE_PROJECT_NAME}_default
     {{ if .IsGitpod }}{{/* see https://github.com/ddev/ddev/issues/3766 */}}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1030,10 +1030,12 @@ func (app *DdevApp) Start() error {
 
 	app.DockerEnv()
 	dockerutil.EnsureDdevNetwork()
-	// "docker-compose up", which is called in this function later, may create
-	// duplicate project networks, we can create this network in advance
-	// see https://github.com/ddev/ddev/pull/5533
-	dockerutil.EnsureProjectNetwork()
+	// The project network may have duplicates, we can remove them here.
+	// See https://github.com/ddev/ddev/pull/5508
+	if os.Getenv("COMPOSE_PROJECT_NAME") != "" {
+		ctx, client := dockerutil.GetDockerClient()
+		dockerutil.RemoveNetworkDuplicates(ctx, client, os.Getenv("COMPOSE_PROJECT_NAME")+"_default")
+	}
 
 	if err = dockerutil.CheckDockerCompose(); err != nil {
 		util.Failed(`Your docker-compose version does not exist or is set to an invalid version.
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -80,40 +80,6 @@ func EnsureDdevNetwork() {
 	}
 }
 
-// EnsureProjectNetwork creates or ensures the project network exists or
-// exits with fatal.
-func EnsureProjectNetwork() {
-	if os.Getenv("COMPOSE_PROJECT_NAME") == "" {
-		log.Fatalf("dockerutil.EnsureProjectNetwork() must be called after app.DockerEnv()")
-	}
-	networkName := os.Getenv("COMPOSE_PROJECT_NAME") + "_default"
-	ctx, client := GetDockerClient()
-	netOptions := dockerTypes.NetworkCreate{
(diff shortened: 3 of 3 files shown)


rfay:
The Colima build has failed 5 times with
`        	Error:      	"bash: line 1: /mnt/ddev_config/commands/web/autocomplete/project-web-cmd: Permission denied\n[Debug] error: bash: line 1: /mnt/ddev_config/commands/web/autocomplete/project-web-cmd: Permission denied,composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-testcmdwordpress docker-compose -f /Users/runner/tmp/ddevtest/TestCmdWordpress576168873/.ddev/.ddev-docker-compose-full.yaml exec -w /var/www/html/ -T web bash -c set -eu && ( /mnt/ddev_config/commands/web/autocomplete/project-web-cmd project-web-cmd anArg)', action='[]', err='exit status 126', stdout='', stderr='bash: line 1: /mnt/ddev_config/commands/web/autocomplete/project-web-cmd: Permission denied'\n:0\nCompletion ended with directive: ShellCompDirectiveDefault\n" does not contain "TestCmdWordpress-web"
`

I think I might have seen something like that on another OS as well. 

Normally we just ignore Colima and restart, but maybe this has happened too many times.

rfay:
Oh, same thing on Windows (master) in https://buildkite.com/ddev/ddev-windows-mutagen/builds/4432#018dae38-41ae-40f6-ab51-ad605ea8a2d0

rfay:
I think the problem we're seeing happens when you have a clean ~/.ddev (`commands` not yet created).  This happens on Colima and should be happening on the github tests. Test fails because of that. It's a result of 
* https://github.com/ddev/ddev/pull/5702

rfay:
I haven't been able to reproduce locally on macOS. I guess this is going to require tmate

rfay (approved):
This looks good to me, and can go in when you're ready. Anything that removes more code than it adds is a happy thing. I manually tested with the instructions and it seemed fine.

Please do update the testing instructions to add manual testing instructions that cover the history of this class of problem, not just this one. We want to make sure that we don't have to do any more of these, if possible. So think back through the problems that have been fixed over time and make sure we've tested them, thanks!

rfay:
@jonaseberle please review with your extensive knowledge of this class of issues, thanks!

stasadev:
Rebased to resolve conflicts.

jonaseberle:
I understand that we let docker-compose create the network. We gain less custom code to maintain ❤️  and configurability of the network 🚀  as @stasadev showed in the manual testing instructions.

I like that a lot. docker-compose has given us some trouble but in my experience it has become really stable "again". I like the approach to go back to it where we can.

I checked that all the labels (that we used to set manually) are now created by docker-dompose.
```
        "Labels": {
            "com.ddev.platform": "ddev",
            "com.docker.compose.network": "default",
            "com.docker.compose.project": "ddev-brillingerde",
            "com.docker.compose.version": "2.24.5"
        }
```

It looks good to me. Thank you!

jonaseberle (approved):

PR #5853, opened 2024-02-16
docs: add auto light / dark mode

stasadev:
## The Issue

I want to try auto light / dark mode for the docs.

Uses https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/?h=dark+mode#automatic-light-dark-mode

Try out https://ddev--5853.org.readthedocs.build/en/5853/

image

### DIFF
--- a/docs/content/assets/extra.css
+++ b/docs/content/assets/extra.css
@@ -2,6 +2,12 @@
 :root {
     /* #4051b5 (indigo primary) in decimal RGB */
     --custom-bg-color: rgba(64, 81, 181, .1);
+    --custom-header-color: var(--md-primary-fg-color);
+}
+
+/* dark color scheme overrides */
+[data-md-color-scheme="slate"] {
+    --custom-header-color: var(--md-primary-fg-color--light);
 }
 
 /* table does not need a scrollbar, see https://github.com/ddev/ddev/pull/3290#issuecomment-942888867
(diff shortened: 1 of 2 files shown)


rfay (approved):
Works for me. Seems like I have to hit it more than once to get it to switch. If you like it it can go in. Will request @bmartinez287 and @mattstein

rfay:
Please review @bmartinez287 thanks!

mattstein (approved):
Looks like the multiple clicks are because it toggles between the default system setting (A for automatic), then light or dark. First I’ve seen that pattern, but it makes sense. Ideally it’d behave in the same way as ddev.com I guess, but better to offer the option regardless I think. Good idea!

stasadev:
I have a global dark theme and it works automatically without any clicks.
And yes, I chose this intentionally (auto/light/dark) so as not to restrict anyone.

bmartinez287:
Pretty cool, I like it. At some point we should add that toggle option to ddev.com. The biggest challenge there is featured sponsors section. Multicolor SVGs have their challenges.

GuySartorelli:
Looks great!

There are a couple of contrast issues picked up by [WAVE](https://wave.webaim.org/extension/) - might pay to make that bluey/purple used in headings a little lighter when in dark mode.

rfay:
Rebased

bmartinez287:
https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/
If we switch from indigo to
Primarily blue and accent light blue on dark mode that might fix the a11y issue.

stasadev:
Thanks, I fixed the problem with the colors.

PR #5854, opened 2024-02-16
fix: make `ddev --version` work even if not in project and ~/.ddev/commands not yet created

rfay:
## The Issue

If no ~/.ddev exists, and `ddev --version` is executed in a directory that is not a project, the global commands directory doesn't get created. You end up getting an error like this:

```
$ mv ~/.ddev ~/.ddev.bak
rfay@ub-2204:~/workspace/ddev$ ddev --version
Adding custom/shell commands failed: open /home/rfay/.ddev/commands/host: no such file or directory
ddev version v1.22.7-23-gefa8cea57
```

Create the global directory structure before existing the test for "am I in a project"

```
mkdir -p ~/tmp/junk && cd ~/tmp/junk
mv ~/.ddev ~/.ddev.bak
ddev --version
```

You shouldn't get an error message. Don't forget to copy your ~/.ddev back into place if you care about it.

### DIFF
--- a/pkg/ddevapp/assets.go
+++ b/pkg/ddevapp/assets.go
@@ -2,10 +2,10 @@ package ddevapp
 
 import (
 	"embed"
+	"github.com/ddev/ddev/pkg/globalconfig"
 	"path/filepath"
 
 	"github.com/ddev/ddev/pkg/fileutil"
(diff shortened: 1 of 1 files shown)


rfay:
This problem existed in v1.22.7

PR #5856, opened 2024-02-16
fix: error message 'Error converting to int' on `ddev import-db`

rfay:
## The Issue

I've been noticing in tests and in manual testing of `ddev import-db` the message "Error converting '' to int":

```
% ddev import-db --file=.tarballs/db.sql.gz
13.9MiB 0:00:01 [7.00MiB/s] [================================>] 100%
Error converting '' to int
Successfully imported database 'db' for d10
```

Allow empty and don't try to convert it to a zero. I'm not sure this is a change in go's behavior.

I was unable to find a replacement technique for this stanza. The original idea was to watch the rows come in until they stopped coming in. I don't see how to do that now. But I did my traditional huge database import-db and it seemed to work fine.

- [x] `ddev import-db` shouldn't show this
- [x] Make sure `ddev import-db` is behaving OK on huge db import, which is what this code is for.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -709,36 +709,6 @@ func (app *DdevApp) ImportDB(dumpFile string, extractPath string, progress bool,
 		return fmt.Errorf("failed to import database: %v\nstdout: %s\nstderr: %s", err, stdout, stderr)
 	}
 
-	// Wait for import to really complete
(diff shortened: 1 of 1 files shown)


rfay:
The problem here is that Innodb_rows_inserted was removed in MariaDB 10.10, https://mariadb.com/kb/en/innodb-status-variables/#innodb_rows_inserted

PR #5857, opened 2024-02-17
docs: mostly warn people off of using xdebug_ide_location

rfay:
## The Issue

A user in Discord set xdebug_ide_location to `container` and didn't know why it didn't work

Try to get people not to use it except in the obscure situations it's needed for.

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -666,7 +666,7 @@ Whether Xdebug should be enabled for [step debugging](../debugging-profiling/ste
 
 ## `xdebug_ide_location`
 
-Adjust Xdebug listen location for WSL2 or in-container.
+Adjust Xdebug listen location for WSL2 or in-container. This is used very rarely, please don't change it unless you really know why.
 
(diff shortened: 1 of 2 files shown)


docs/content/users/configuration/config.md
```diff
@@ -666,7 +666,7 @@ Whether Xdebug should be enabled for [step debugging](../debugging-profiling/ste
 
 ## `xdebug_ide_location`
 
-Adjust Xdebug listen location for WSL2 or in-container.
+Adjust Xdebug listen location for WSL2 or in-container. This is used very rarely, please don't change it unless you really know why.
```
GuySartorelli, line 669:
```suggestion
Adjust Xdebug listen location for WSL2 or in-container. This is used very rarely, please don't change it unless you really know what you're doing.
```

GuySartorelli (commented):
Might make sense to add something like "seek advice in one of our support channels before making any of these changes unless you really know what you're doing" and link to https://ddev.com/support-ddev/#help-others or similar location

PR #5858, opened 2024-02-17
docs: stop checking all stack overflow links

rfay:
## The Issue

Stack Overflow is giving us 403 for all URL link checks, so stop trying.

### DIFF
--- a/markdown-link-check.json
+++ b/markdown-link-check.json
@@ -46,7 +46,7 @@
       "pattern": "^https://analytics.amplitude.com"
     },
     {
-      "pattern": "^https://(stackoverflow|superuser).com/"
+      "pattern": "^https://(meta.stackoverflow|stackoverflow|superuser).com/"
     },
     {
       "pattern": "^https://packages.debian.org/stable/"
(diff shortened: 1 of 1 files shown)


markdown-link-check.json
```diff
@@ -46,10 +46,10 @@
       "pattern": "^https://analytics.amplitude.com"
     },
     {
-      "pattern": "^https://stackoverflow.com/questions/tagged/ddev"
+      "pattern": "^https://(meta.stackoverflow|stackoverflow|superuser).com/"
```
GuySartorelli, line 49:
```suggestion
      "pattern": "^https://(meta.)?stackoverflow|superuser).com/"
```

GuySartorelli (approved):
Looks like this will cover all the failing links in https://github.com/ddev/ddev/actions/runs/7939391160/job/21679447737?pr=5753

I've proposed a minor change below but it doesn't affect the result so not a big deal if you don't make the change.

markdown-link-check.json
rfay: Thanks, more elegant and much appreciated, but I think the less elegant one is easier to understand and expand when the time comes.

PR #5855, opened 2024-02-16
docs: show a warning banner for the latest docs (readthedocs+mkdocs)

stasadev:
## The Issue

- #5832

## Alternative

- #5859

## How This PR Solves The Issue

Adds a warning banner.

Resources used in making this:
- https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS_VERSION_NAME
- https://docs.readthedocs.io/en/stable/versions.html#version-warning
- https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=outdated#version-warning
- https://github.com/readthedocs/readthedocs.org/issues/8529

Fixes mkdocs warnings on build.

## Manual Testing Instructions

See https://ddev--5855.org.readthedocs.build/en/5855/

image

## Alternative

As an alternative we could use this https://docs.readthedocs.io/en/stable/versions.html#version-warning

image

But it appears over and over again when you navigate the latest docs, so I decided not to use it (disabled it here https://beta.readthedocs.org/dashboard/ddev/addons/edit/).

PR #5859, opened 2024-02-17
docs: show a warning banner for the latest docs

bmartinez287:
This PR is an alternative to https://github.com/ddev/ddev/pull/5855
It accomplishes similar results using jinja and mkdocs native functionality.


bmartinez287:
@stasadev since both branches seem to have a slightly different mkdocs.yml I think this might work.

rfay:
I do like this approach, using mkdocs-material feature directly.

rfay:
Rebased to get rid of docs failure

GuySartorelli:
Implementation approach notwithstanding, I prefer the appearance of https://github.com/ddev/ddev/pull/5855 since it stands out a lot more.

I've also suggested a change of wording on that PR - if we go with this one, I recommend the changes I suggested there be made here.

bmartinez287:
Yeah, I am not biased towards the content in it, it can be anything. I'm good with whatever the maintainers lean towards.
The color I could see why yellow is helpful but that requires extra CSS changes and leaving closer to its default state might be MVP(minimum viable product) maybe. We could change that by overriding their base.html file (complex solution) or adding a CSS style that targets the announce html attribute (simpler but less flexible long term)

The biggest challenge so far is that the announce bar as it stands never disappears because of the core implementation of the material mkdocs block layout(see link below). In other words, if we make the bar yellow, it be yellow anywhere even if it has no content.

https://github.com/squidfunk/mkdocs-material/blob/master/src/templates/base.html#L229

GuySartorelli:
It honestly sounds like the built in announcement bar might not be fit for this purpose - something that only displays when there's content to display (i.e. when warning users about the content they're viewing) stands out more when it does display. 
Otherwise people may be used to seeing the bar there (even more so if it's a colour that stands out) and assume it's something they've already read and chosen to ignore.

bmartinez287:
Maybe. 
It is collapsed when unused even though it will be visible as in the second pic above. 
That said, it does avoid the layout shift issue JavaScript might have since it will load as the document does.
There is always a tradeoff. It could also be a pop-up if we wanted people to see it, but that tends to be more annoying and outdated.

Something is better than nothing that's for sure and either way it's an improvement.

docs/overrides/main.html
```diff
@@ -4,3 +4,13 @@
...
+
+{% block announce %}
+  {% if config.extra.version == "latest" %}
+        {{ super() }}
+        You’re viewing the latest unreleased version.
+        <a href="https://ddev.readthedocs.io/en/stable/">
+          <strong>Released docs</strong>
+        </a>
+  {% endif %}
+{% endblock %}
```
stasadev, line 16:
```suggestion
{% block announce %}
  {% if config.extra.version == "latest" %}
    {{ super() }}
    You're viewing documentation for an unreleased version of DDEV.
    <a id="stable-docs-link" href="https://ddev.readthedocs.io/en/stable/">
      <strong>Click here to see the stable documentation.</strong>
    </a>
    <script>
      document.getElementById('stable-docs-link').href = `https://ddev.readthedocs.io/${new URL(window.location.href).pathname.replace(/\/en\/[^/]+/, 'en/stable')}`;
    </script>
  {% else %}
    <style>
      div[data-md-component="announce"] {
        display: none;
      }
    </style>
  {% endif %}
{% endblock %}
```

stasadev (changes requested):
I like this, and I have a fix for empty announce bar, see my suggestion.

The main question for me is how we can change `version: latest` to something else in stable docs. I doubt that it will change by itself.

bmartinez287:
Great question, don't we keep separate mkdocs files for lastest and stable?
https://github.com/ddev/ddev/blob/dont-use-this/mkdocs.yml that one seems to have lates as well as a few other branches.
Is it done during the CI/CD? what changes the mkdocs site_url for them?

bmartinez287:
If anything maybe the variable could be overridden here. Once the docs are about to be published grep that variable and replace it's value.
https://github.com/ddev/ddev/blob/master/.github/workflows/docspublish.yml

stasadev:
I'm going to look into this after I'm done with my other priorities.

PR #5871, opened 2024-02-19
fix: ddev start should not continue if web/db containers aren't healthy

rfay:
## The Issue

On a number of [rancher desktop tests](https://buildkite.com/ddev/macos-rancher-desktop/builds/586#018dc0d5-43b7-4b9d-87ee-30c8d521908b) we're seeing docker having failed, and DDEV having detected startup failure, but continuing on in processing app.Start(), eventually seeing the misleading "Something is wrong with Docker or Colima and /mnt/ddev_config is not mounted from the project .ddev folder. This can cause all kinds of problems."
 
We also saw this misleading output on 
* https://github.com/ddev/ddev-sqlsrv/issues/20

If the web and db containers can't come up, stop the app.Start() and error out.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1428,7 +1428,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	}
 
 	if waitErr != nil {
-		util.Failed("Failed waiting for web/db containers to become ready: %v", err)
+		util.Failed("Failed waiting for web/db containers to become ready: %v", waitErr)
 	}
 
 	// WebExtraDaemons have to be started after Mutagen sync is done, because so often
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -1430,6 +1427,10 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		}
 	}
 
+	if waitErr != nil {
+		util.Failed("Failed waiting for web/db containers to become ready: %v", err)
```
stasadev, line 1431:
```suggestion
		util.Failed("Failed waiting for web/db containers to become ready: %v", waitErr)
```

stasadev (changes requested):

pkg/ddevapp/ddevapp.go
rfay: Great eye, thanks.

stasadev (approved):

GuySartorelli (approved):
Great idea!

PR #5874, opened 2024-02-20
test: Fix TestPantheonPush, broken due to php8.2 default

rfay:
## The Issue

TestPantheonPush has been failing.

Studying the problem showed that it was the move to PHP8.2 as default that broke it; drush couldn't be installed on Drupal9.

* Use Drupal10
* Use PHP8.2 

None needed.


rfay:
I think this should be good to go, test change only, related to PHP8.2 default.

PR #5878, opened 2024-02-20
fix: cache Docker client instead of creating a new one all the time, fixes #5861

### LINKED ISSUE #5861: Using Docker API broke Colima/Lima/Rancher usage
It seems that Colima doesn't work at all any more since 
* https://github.com/ddev/ddev/pull/5787

We have always had trouble with Colima, so we just kind of ignored the failures there I think. 

Now Colima never works at all

This is easy to recreate  running TestComposerVersion, and Colima isn't actually dead after it completes.

I checked out the commit before the Docker API change, and it passes this test just fine using Colima.

For a test environment, feel free to take one of the macOS Test Runners offline `brew services stop buildkite-agent` and use Goland or vscode there.

## Manual Testing

1. Install Colima (current release) using normal DDEV instructions, https://ddev.readthedocs.io/en/latest/users/install/docker-installation/#macos
  This can also be done as a separate profile if you want to, for example `colima start qemu --cpu 4 --memory 6 --disk 100 --vm-type=qemu --mount-type=sshfs --dns=1.1.1.1` to create a profile called `qemu` and not disturb anything you already have set up.

1. Stop all other docker providers so that they can't be part of the problem.
2. Verify that the context is set correctly and is pointing to your new colima instance, `docker context ls`
3. Verify that `docker ps` works OK.

To run the actual test, you can run it manually (below) or in your IDE:
```
GOTEST_SHORT=true make testpkg TESTARGS='-run TestComposerVersion --count=1'
```

It's instructive to `ddev list` or `docker ps` while docker is broken.

## Testing environment

I was unable to get this to fail on tb-macos-amd64-3, but it fails just fine on tb-macos-arm64-7. tb-macos-arm64-7 is all set up for testing, except you may want an ide there to step through things.

```
brew services stop buildkite-agent
orb stop
# Make sure rancher is stopped as well.
colima start
cd ~/workspace/ddev
GOTEST_SHORT=true make testpkg TESTARGS='-run TestComposerVersion --count=1'
```

stasadev:
## The Issue

- #5861

Do not create a lot of requests for new Docker client.
Close client connection.

I run `TestComposerVersion` several times on **tb-macos-arm64-7**, all green:

```
GOTEST_SHORT=true make testpkg TESTARGS='-run TestComposerVersion --count=1'
```

## Automated Testing Overview

Waiting for Colima tests.

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -139,6 +139,8 @@ func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, n
 
 var DockerHost string
 var DockerContext string
+var DockerCtx context.Context
+var DockerClient *dockerClient.Client
 
 // GetDockerClient returns a Docker client respecting the current Docker context
 // but DOCKER_HOST gets priority
@@ -159,14 +161,19 @@ func GetDockerClient() (context.Context, *dockerClient.Client) {
 		util.Debug("GetDockerClient: Setting DOCKER_HOST to '%s'", DockerHost)
 		_ = os.Setenv("DOCKER_HOST", DockerHost)
 	}
-	ctx := context.Background()
-	client, err := dockerClient.NewClientWithOpts(dockerClient.FromEnv, dockerClient.WithAPIVersionNegotiation())
-	if err != nil {
-		output.UserOut.Warnf("Could not get Docker client. Is Docker running? Error: %v", err)
-		// Use os.Exit instead of util.Failed() to avoid import cycle with util.
-		os.Exit(100)
+	if DockerClient == nil {
+		DockerCtx = context.Background()
+		DockerClient, err = dockerClient.NewClientWithOpts(dockerClient.FromEnv, dockerClient.WithAPIVersionNegotiation())
+		if err != nil {
+			output.UserOut.Warnf("Could not get Docker client. Is Docker running? Error: %v", err)
+			// Use os.Exit instead of util.Failed() to avoid import cycle with util.
+			os.Exit(100)
+		}
 	}
-	return ctx, client
+
+	defer DockerClient.Close()
+
+	return DockerCtx, DockerClient
 }
 
 // GetDockerContext returns the currently set Docker context, host, and error
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -169,10 +169,8 @@ func GetDockerClient() (context.Context, *dockerClient.Client) {
 			// Use os.Exit instead of util.Failed() to avoid import cycle with util.
 			os.Exit(100)
 		}
+		defer DockerClient.Close()
 	}
-
-	defer DockerClient.Close()
-
 	return DockerCtx, DockerClient
 }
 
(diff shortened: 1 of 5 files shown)


pkg/dockerutil/dockerutils.go
```diff
@@ -159,14 +161,19 @@ func GetDockerClient() (context.Context, *dockerClient.Client) {
...
+		DockerClient, err = dockerClient.NewClientWithOpts(dockerClient.FromEnv, dockerClient.WithAPIVersionNegotiation())
+		if err != nil {
+			output.UserOut.Warnf("Could not get Docker client. Is Docker running? Error: %v", err)
+			// Use os.Exit instead of util.Failed() to avoid import cycle with util.
+			os.Exit(100)
+		}
 	}
-	return ctx, client
+
+	defer DockerClient.Close()
```
stasadev, line 174: This should probably be inside `if DockerClient == nil`, but I'll wait for the tests to go through.

rfay:
Let's enable Rancher and macOS amd64 on this PR.

rfay:
Only TestConfigValidate failed, https://github.com/ddev/ddev/actions/runs/7979112678/job/21785753458?pr=5878

Restarting. This kind of failure is quite common.

rfay:
Failed again on TestConfigValidate, same thing. 

Retry

rfay:
We may have to use tmate to try TestConfigValidate, or see how it works on tb-macos-arm64-6

rfay:
Oh look at TestConfigValidate:
```
func TestConfigValidate(t *testing.T) {
	if nodeps.IsAppleSilicon() {
		t.Skip("Skipping on mac M1 to ignore problems with 'connection reset by peer'")
	}
```

If I remove the "skip on Apple Silicon" then I get the same failures on my machine using colima:

```
All project containers are now ready.
    config_test.go:628: 
        	Error Trace:	/Users/rfay/workspace/ddev/pkg/ddevapp/config_test.go:628
        	            	Get "http://127.0.0.1//readme.html": dial tcp 127.0.0.1:80: connect: connection refused
        	Test:       	TestConfigValidate
    config_test.go:630: 
        	Error Trace:	/Users/rfay/workspace/ddev/pkg/ddevapp/config_test.go:630
        	            	Get "http://127.0.0.1//readme.html": dial tcp 127.0.0.1:80: connect: connection refused
        	Test:       	TestConfigValidate
```

These are the lines that are failing:

```
		_, _, err = testcommon.GetLocalHTTPResponse(t, "http://x.ddev.site/"+staticURI)
		assert.NoError(err)
		_, _, err = testcommon.GetLocalHTTPResponse(t, "http://somethjingrandom.any.ddev.site/"+staticURI)
		assert.NoError(err)
```

I do not get those failures on Orbstack or Docker Desktop.

I'm pretty sure the exclusion for Apple Silicon was put in there in the early unstable days of Docker Desktop on arm64. Not that it's all that stable now. But it wasn't intended to exclude Colima on arm64. 

TestConfigValidate also passes on HEAD with Docker Desktop. 
TestConfigValidate FAILS on HEAD with Colima, but it's a bit different, same lines, but it's the EOF error
TestConfigValidate FAILS sometimes but not always on Colima at ca7e2ab3f28f7b094fb0b79480af690c05843400 (right before API change) in the same way as above (connection refused)

It's possible that this has been a flaky test for colima for a really long time and we might be able to ignore it. It does not appear to be new with this patch.

rfay:
I'm in favor of adding `if IsColima() t.Skip()` on TestConfigValidate. I think it's OK.

rfay:
Yay, it looks like we did finally get a clean colima on https://github.com/ddev/ddev/pull/5878/commits/538c5ca3b850108fb41a37c28ca7126c5179f352

Congratulations. I think TestConfigValidate is just too flaky on Colima so we'll be fine to be rid of it.

rfay (approved):
Thanks for the amazing work on this. I think we may have almost dug out of that hole :)

GuySartorelli:
Amazing work!

PR #5879, opened 2024-02-20
test: Minor improvement to TestComposerVersion

rfay:
## The Issue

This is a very minor improvement to TestComposerVersion that I needed when working on 
* https://github.com/ddev/ddev/issues/5861

so that I could restart the test even if I aborted in the middle of it.


GuySartorelli (approved):
Seems harmless at worst 👍

PR #5883, opened 2024-02-22
test: exclude TestPHPOverrides on Colima

rfay:
## The Issue

Colima is always difficult to get through a run. Lately we're seeing TestPHPOverrides fail a lot, and it seems no reason to work any harder on it.

Don't test it on Colima.

HOWEVER: Run local tests with colima on TestPHPOverrides and see if it can be reproduced.


rfay:
I manually tested TestPHPOverrides on macOS (with the t.Skip removed) and it tested fine (Colima)

PR #5885, opened 2024-02-22
docs: updating troubleshooting.md with instructions on how to solve ipv4 address pool issue

scotteuser:
## The Issue

At least for myself, I frequently hit this error, yet Troubleshooting section of docs does not cover it.

```
YYYY/MM/DD HH:II:SS Failed to ensure Docker network ddev-sitenamehere_default: API error (404): could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network
```
I decided to dig further to actually solve the error rather than just pruning the network. The error is now solved permanently for me and I would like to share the fix.

This provides instructions as to how to solve it.

Attempt to add new `default-address-pools` to your `daemon.json` as per the instructions and see the
address pool size increase.

## Automated Testing Overview

Documentation only, no tests.

## Related Issue Link(s)

Could not find any related issues

## Release/Deployment Notes

Documentation only change

### DIFF
--- a/docs/content/users/usage/troubleshooting.md
+++ b/docs/content/users/usage/troubleshooting.md
@@ -246,6 +246,41 @@ docker rmi -f $(docker images -q)
 
 You should then be able to start your DDEV machine.
 
+## `ddev start` Fails with "could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network"
+
+This typically happens on a Linux installation like Ubuntu with many ddev projects where you have installed Docker manually rather than using
+something like Docker Desktop. Pruning the network generally solves this problem, but it does mean already existing non-running ddev projects
+will need a `ddev restart`. To prune the network run:
+```
+docker network prune
+```
+
+If you encounter this problem often you can add to your `/etc/docker/daemon.json`:
+
+```
+{
+  ...
+  "default-address-pools": [
+    {"base": "172.17.0.0/16", "size": 24}
+  ]
+  ...
+}
+```
+If the file does not exist, it can be created and set to:
+```
+{
+  "default-address-pools": [
+    {"base": "172.17.0.0/16", "size": 24}
+  ]
+}
+```
+To have the changes take effect you need to reload the configuration and restart docker (which will stop any running ddev projects). You can do this
+as follows:
+```
+sudo systemctl daemon-reload
+sudo systemctl restart docker
+```
+
 ## `ddev --version` shows an old version
 
 If you have installed the latest version of DDEV, but when you check the actual version with `ddev --version`, it shows an older version, please refer to [Why do I have an old DDEV?](./faq.md#why-do-i-have-an-old-ddev)


rfay:
This really isn't a DDEV thing. Do you have a link to a more general article about the problem? Maybe that should go in there?

scotteuser:
I could only find random ad-filled aggregate sites that had pulled in blog posts on it. But fair enough, if you don't think it's helpful np.

scotteuser:
Basically had thought it was similar to things already in the docs like 'Web Server Ports Already Occupied' which are similarly perhaps not ddev problems but things that someone a ddev user may encounter...

scotteuser:
Actually here is one on stack overflow for future reference https://stackoverflow.com/a/62176334/814142

scotteuser:
[ddev-debug-test-pull-5885.txt](https://github.com/ddev/ddev/files/14382323/ddev-debug-test-pull-5885.txt)

Sure here it is; note that I renamed particular client site names to redacted-siteX where X increments per differnent size in case the quantity of each single line repeating itself is helpful. Thanks!

scotteuser:
It does not happen every time but it seems to be the sheer number of projects I have, and as that keeps increasing over time I have been hitting this issue nearly daily now which prompted me to investigate more permanent solutions. I have a ton of different client sites as I tend to support other developers with prototyping and the projects they are working up + booting up ddev sites for different module development/drupal core issues. It does not seem to be related to how many projects are actually running (ie, `ddev start`ed) at once.

Here is my docker setup, just a straightforward install with apt-get
```
scott@scott:~$ apt list --installed | grep "docker"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

docker-buildx-plugin/now 0.10.2-1~ubuntu.20.04~focal amd64 [installed,local]
docker-ce-cli/now 5:23.0.1-1~ubuntu.20.04~focal amd64 [installed,local]
docker-ce-rootless-extras/now 5:23.0.1-1~ubuntu.20.04~focal amd64 [installed,local]
docker-ce/now 5:23.0.1-1~ubuntu.20.04~focal amd64 [installed,local]
docker-compose-plugin/now 2.16.0-1~ubuntu.20.04~focal amd64 [installed,local]
docker-scan-plugin/now 0.23.0~ubuntu-focal amd64 [installed,local]

scott@scott:~$ which docker
/usr/bin/docker

scott@scott:~$ docker --version
Docker version 23.0.1, build a5ee5b1

scott@scott:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
```

I keep everything up to date in terms of all ubuntu software + ddev version and shut the laptop down nightly rather than just suspending, so I think good habits there :)

scotteuser:
Thanks for the tips! Will try and see if it helps

rfay:
I'm going to be interested to hear if this happens to you if you do a `ddev poweroff` from time to time. 

If it only happens when you have 100 active projects, I think this may be better as a Stack Overflow answer tagged with "ddev", but will wait for your results as you spend some time thinking about it.

rfay:
I'm going to close this for now, but happy to reopen based on your experience when you keep less than 100 projects active at a time :)  Happy to continue the conversation here or elsewhere, and this will document the right strategy regardless

PR #5886, opened 2024-02-22
test: Run colima tests native instead of github, fixes #5598

### LINKED ISSUE #5598: Move colima tests to buildkite
Running colima + ddev on github actions runner is just too terribly painfully slow.

Move it to our buildkite runners.

rfay:
## The Issue

* #5598 

We have never been able to get predictable results from colima tests on Github Actions. Maybe the VMs are too small
Also, we were never able to do normal https tests, so there are a bunch of things that require http, we might be able to remove those.

Run the builds on our larger buildkite test machines.

## Detail

Testing by running TestExtraPortExpose with:
`GOTEST_SHORT=true make testpkg TESTARGS="-run TestExtraPortExpose -count=1"`

It seems to work after a `colima start` and then a `colima restart`. And then it stays working. So weird.

### DIFF
--- a/.buildkite/macos-colima.yml
+++ b/.buildkite/macos-colima.yml
@@ -1,0 +1,17 @@
+  - command: ".buildkite/test.sh"
+    plugins:
+      - docker-login#v2.1.0:
+          password-env: DOCKERHUB_PULL_PASSWORD
+    agents:
+      - "os=macos"
+      - "colima=true"
+      - "architecture=arm64"
+    #branches: "master"
+    env:
+      BUILDKITE_CLEAN_CHECKOUT: true
+      BUILDKITE_BUILD_PATH: ~/tmp/buildkite_builds
+      BUILDKIT_PROGRESS: plain
+      DDEV_TEST_SHARE_CMD: "false"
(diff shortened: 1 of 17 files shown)


rfay:
Both colima tests (vz and qemu) ran in about an hour and a half (and succeeded). Yay!

stasadev (approved):
Looks great to me!

And I think we can add comments to all buildkite config yml files with clear info about where it can be run at the top of the file, without using the `branches:` stanza, to understand the expected behavior for each test:

For example:
add to `.buildkite/macos-orbstack.yml`
```
# See https://buildkite.com/ddev/macos-orbstack/settings/repository
# Runs on master, PRs, and 3rd party PRs
```

This is not ideal, of course, but it gives us a better understanding of the test, and when we return to it in six months or a year, we can already see how it should behave.

rfay:
Right now we're running 3 tests on every PR (colima, colima_vz, orbstack) that require tb-macos-arm64-(5-6-7), so they're fully occupied for one PR. 

On master, we're running 4 (colima, colima_vz, orbstack, rancher)

I think we'll probably want to dial back one of the colima ones soon.

PR #5890, opened 2024-02-23
feat: change upgrade pester interval to 4 hours, fixes #5889

### LINKED ISSUE #5889: Notify the user when a ddev update is available as the last item in output

I can spend hours trying to debug a ddev-related issue only to discover there's an update for ddev that solves the issue. 

### Describe your solution

Have DDEV notify users of DDEV updates on a more regular basis (daily, every use) and have the message be the _last_ thing output to the user. 

### Describe alternatives

_No response_

### Additional context

As someone who only occasionally works on projects where DDEV is a suitable local development tool, it can be months between when I use it. As I understand it, DDEV does not inform users of an update per-use but once a week (or month?). It also appears to be the very first message (or very early) once ddev begins its process (e.g. outputting after walking through the interactivity for `ddev config`). Given the large amount of DDEV sends to stdout during its rouines, it is *very* easy to miss the message before it scrolls out of view. Combine that with only notifying users once a week (month), it is very easy for users to miss the notification.

rfay:
## The Issue

* #5889 

People are running old versions of DDEV, and they say that they don't get properly notified about that.

Pester them on 4-hour cycle instead of one-week cycle.

Here's the updated view of it (with the correct URL):

image

Try using an out-of-date version. You should get notified every 4 hours.

You can change the time on ~/.ddev/.update to force this cycle. `touch -d "5 hours ago" ~/.ddev/.update` or `touch -m -t 202402230500 ~/.ddev/.update`


stasadev (approved):
I didn't even know about it because I install updates on my Arch-based system almost every day.

Tested, it works as expected.

PR #5892, opened 2024-02-23
feat: Add native lima Docker provider support

rfay:
## The Issue

Colima has gotten quite unstable for lots of people, but it's just a wrapper on Lima, and so we should be able to use Lima directly as another open-source alternative.

It's worth considering whether we should do this at all, as adding more options doesn't necessarily help people or our support environment.

## TODO

- [x] Docs
- [x] Turn on tests (preferably without putting red marks on other PRs)
- [ ] Make it pass tests

Use the docs to start a lima instance, then do a simple quickstart test.

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -10,6 +10,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     * [OrbStack](#orbstack)
     * [Colima](#colima)
+    * [Lima](#lima)
     * [Docker Desktop](#docker-desktop-for-mac)
     * [Rancher Desktop](#rancher-desktop)
 
@@ -35,8 +36,6 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     After the initial run above, you can use `colima start` or use `colima start -e` to edit the configuration file. Run `colima status` at any time to check Colima’s status.
 
-    When your computer restarts, you’ll need to `colima start` again. If you prefer to start Colima automatically on reboot, use `brew services start colima` in Colima version 0.6+ to configure auto-start.
-
     !!!tip "Colima disk allocation"
         In Colima versions starting with 0.5.4 you can increase—but not decrease—the disk allocation by editing `~/.colima/default/colima.yaml` to change the `disk` setting to a higher value. For example, `disk: 200` will increase allocation to 200 gigabytes. Then `colima restart` will result in the new disk allocation.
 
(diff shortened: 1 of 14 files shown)


rfay:
Fails TestDdevAllDatabases with non-writable filesystem:
https://buildkite.com/ddev/macos-lima/builds/12#018e1f6d-091a-4022-8753-f14e26bb1c74

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     !!!warning "Colima can only work in your home directory unless you do further configuration"
         By default, Colima only mounts your home directory, so it’s easiest to use it in a subdirectory there. See the `~/.colima/default/colima.yaml` for more information, or notes in [colima.yaml](https://github.com/abiosoft/colima/blob/fc948f8f055600986f87e29e3e632daf56ac8774/embedded/defaults/colima.yaml#L130-L143).
 
+    ### Lima
+
+    [Lima](https://github.com/lima-vm/lima) is a free and open-source project supported by the [Cloud Native Computing Foundation](https://cncf.io/).
+
+    1. Install Lima with `brew install lima`.
+    2. Install the `docker` client if you don't already have it with `brew install docker`.
+    3. Create a default Lima setup with 4 CPUs, 6GB memory, 100GB storage, and Cloudflare DNS, adjusting as needed:
```
rpkoller, line 54:
```suggestion
    3. Create a 100GB VM in Lima with 4 CPUs, 6GB memory, and Cloudflare DNS. Adjust to your own needs:
```

rpkoller: Also oriented to the proposed changes on the colima quickstart PR

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+    ```
+    limactl create --vm-type=vz --mount-type=virtiofs --mount-writable --memory=6 --cpus=4 --disk=100 template://docker
```
rpkoller, line 56: i am cautious and a bit afraid of mounting the entire home directory, most of the lima maintainers share the same worry. That is the reason why out of the box only the lima temp directory is added. but the rancher maintainer is the only lima maintainer who mounts his entire homedirectory for convenience reasons. just as a note. i am ok with it but personally i only use --mount="~/Sites:w" (with Sites being the directory i store all my projects)

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+    docker context create lima-default --docker "host=unix://$HOME/.lima/default/sock/docker.sock"
```
rpkoller, line 57: it probably should be noted that this command has only to be run once after the initial creation of the VM. and this line as well as the next (docker context use) are also output by default. they are part of the `message`that is defined in the instances yaml file.

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+    docker context use lima-default
+    ```
+    After the initial run above, you can use `limactl start`.  Run `limactl list` to see configured setup.
```
rpkoller, line 60:
```suggestion
    After the initial setup, you can start your instance with `limactl start`. Run `limactl --help` to see the list of available commands.
```
instead of adding the command to see the configured setup i think providing the help command to get the list of all available command might be more helpful (which also includes the list command)

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+
+    When your computer restarts, you’ll need to `limactl start` again.
+
+    !!!warning "Docker contexts let the Docker client point at the right Docker server"
+        The Docker provider you're using is selected with `docker context`. You can see the available contexts with `docker context ls` and the currently selected one with `docker context show`. With the setup above you'll want `docker context use lima-default`.
+
+    !!!warning "Lima only mounts filesystems in your home directory unless you do further configuration"
```
rpkoller, line 67: same as for the colima quickstart. i think it would make sense to clarify that those mounts are not necessary for lima to run but for ddev projects to run?

rpkoller (commented):
just left a few comments early on. just after leaving comments on the colima quickstart i noticed that there are certain overlaps for those two and it might be a good thing to keep things consistent across quickstarts. and would it make sense since colima and rancher are based on lima to place lima in front of the two?

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     !!!warning "Colima can only work in your home directory unless you do further configuration"
         By default, Colima only mounts your home directory, so it’s easiest to use it in a subdirectory there. See the `~/.colima/default/colima.yaml` for more information, or notes in [colima.yaml](https://github.com/abiosoft/colima/blob/fc948f8f055600986f87e29e3e632daf56ac8774/embedded/defaults/colima.yaml#L130-L143).
 
+    ### Lima
+
+    [Lima](https://github.com/lima-vm/lima) is a free and open-source project supported by the [Cloud Native Computing Foundation](https://cncf.io/).
```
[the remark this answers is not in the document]
rfay: Sorry, can't understand your question but feel free to propose something.

docs/content/users/install/docker-installation.md
```diff
@@ -46,6 +45,27 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+
+    1. Install Lima with `brew install lima`.
```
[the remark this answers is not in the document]
rfay: I like it better this way, didn't like the other one as much.

docs/content/users/install/docker-installation.md
rfay: I understand that your caution, but this is the default behavior and doing it otherwise creates enormous support problems.

docs/content/users/install/docker-installation.md
rfay: This is shown as part of the first-time setup.

docs/content/users/install/docker-installation.md
rfay: I like it as is. People have to branch out to learn more.

docs/content/users/install/docker-installation.md
rfay: Here we're only interested in people running DDEV projects. People can do other things with Lima if they want to. 

If people don't put their projects on a filesystem that Lima mounts then DDEV is nonfunctional. Most people put their projects somewhere in the homedir.

rfay (commented):
Thanks for taking such a good look @rpkoller and thanks for all your expertise in this area!

PR #5894, opened 2024-02-25
feat: add ddev ssh --cwd option, fixes #5839

### LINKED ISSUE #5839: Allow `ddev ssh` to start a shell in the equivalent directory inside the guest

When working on a project you often navigate your terminal to a directory inside of it, example `docroot/themes/custom/something`, but then realize that you need to do something _inside_ the guest _in_ this directory, but `ddev ssh` logs into `/var/www/html` or to a directory you pass as `--dir`.

### Describe your solution

I would like to be able to call `ddev ssh -c` (or similar) to be equivalent to `ddev ssh --dir /var/www/html/themes/custom/something` when the host terminal has `$CWD` as `approot/docroot/themes/custom/something`.

Since `ddev` knows all the above information, it should be possible to substract everything up to `docroot/` from the host's `$CWD`, and prefix the resulting string with the guest's absolute path to its docroot.

### Describe alternatives

Passing the `--dir` option requires an absolute path, from the perspective of the guest so you need to manually edit your `cwd` to use the `/var/www/html` prefix instead of whatever you have on your host `$CWD`.

### Additional context

`ddev` rocks ✌️

diegoe:
## The Issue
* https://github.com/ddev/ddev/issues/5839

Implements a `--cwd` option into `ddev ssh` so that you are logged into the equivalent directory _inside_ the guest.

1. Go to a `ddev` project
2. Browse into children directory under the `docroot`
3. `ddev ssh --cwd`
4. You should now be in the equivalent directory _inside_ the guest

## Automated Testing Overview

A _minimal_ test was hastily included. But I only tested against `GOTEST_SHORT=11`

## Release/Deployment Notes
* `ddev ssh` is now able to log you into the `web` container's equivalent directory to the host's current working directory. For example, if on the host your current working directory is `your-app/your-docroot/themes/custom/your-theme` and you call `ddev ssh --cwd` you will be logged into the guest under `/var/www/html/docroot/themes/custom/your-theme`. This option only works for `web` containers and is mutually exclusive with `--dir`.

### DIFF
--- a/cmd/ddev/cmd/ssh.go
+++ b/cmd/ddev/cmd/ssh.go
@@ -1,8 +1,10 @@
 package cmd
 
 import (
+	"errors"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/nodeps"
@@ -12,6 +14,10 @@ import (
 
 // sshDirArg allows a configurable container destination directory.
 var sshDirArg string
+
+// cwdArg: If true, --cwd, try to cd into the equivalent of $CWD, inside
+// the guest.
+var cwdArg bool
 
 // DdevSSHCmd represents the SSH command.
 var DdevSSHCmd = &cobra.Command{
@@ -40,11 +46,32 @@ ddev ssh -d /var/www/html`,
 		if !nodeps.ArrayContainsString([]string{"web", "db", "solr"}, serviceType) {
 			shell = "sh"
 		}
-		err = app.ExecWithTty(&ddevapp.ExecOpts{
-			Service: serviceType,
-			Cmd:     shell + " -l",
-			Dir:     sshDirArg,
-		})
+
+		if cwdArg && serviceType != "web" {
+			err = errors.New("The --cwd option can only be used with service type \"web\"")
+		}
+
+		if cwdArg && sshDirArg == "" {
+			guestWorkingDir := app.GetWorkingDir(serviceType, "")
+			hostWorkingDir, err := os.Getwd()
+
+			if err == nil {
+				relPath, err := filepath.Rel(app.GetAbsDocroot(false), hostWorkingDir)
+
+				if err == nil {
+					sshDirArg = filepath.Join(guestWorkingDir, app.GetDocroot(), relPath)
+				}
+			}
+		}
+
+		if err == nil {
+			err = app.ExecWithTty(&ddevapp.ExecOpts{
+				Service: serviceType,
+				Cmd:     shell + " -l",
+				Dir:     sshDirArg,
+			})
+		}
+
 		if err != nil {
 			if exiterr, ok := err.(*exec.ExitError); ok {
 				os.Exit(exiterr.ExitCode())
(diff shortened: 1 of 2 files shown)


diegoe:
I'm not super enthusiastic about the name of the option, but I guess that's the clearest way of saying "you'll go to whatever $CWD is equivalent of inside the guest"? Needless to say that I have no clue about Go, and this is largely "context clues inspired hacking" :)

rfay:
Rebased to get current testing setup

stasadev (changes requested):
> I'm not super enthusiastic about the name of the option, but I guess that's the clearest way of saying "you'll go to whatever $CWD is equivalent of inside the guest"?

@diegoe, you meant `$PWD`, not `$CWD`, right?

I found that `cwd` is used as a config option in VSCode (but it can be hard to guess what it is on the first try).

---

This works fine with the default working dir

```
ddev config --web-working-dir /var/www/html
```

But not with something else like

```
ddev config --web-working-dir /var/www/html/test
```

rfay:
I think you might be happier (and everybody else too) with just having a global custom web command `sshhere`:
```
#!/bin/bash

#ddev-generated
## Description: Equivalent of 'ssh' but execs into same directory as on host
## Usage: sshhere 
## Example: "ddev sshhere"
## ExecRaw: true
## HostWorkingDir: true

bash 
```

Put that in ~/.ddev/commands/web/sshhere and `ddev restart` and then use `ddev sshhere`, for example:

```
rfay@rfay-tag1-m1:~/workspace/d10/web/sites/default$ ddev sshhere
rfay@d10-web:/var/www/html/web/sites/default$
```

diegoe:
> > I'm not super enthusiastic about the name of the option, but I guess that's the clearest way of saying "you'll go to whatever $CWD is equivalent of inside the guest"?
> 
> @diegoe, you meant `$PWD`, not `$CWD`, right?
> 
> I found that `cwd` is used as a config option in VSCode (it can be hard to guess what it is on the first try) - but I'm okay with this naming.

For this specific context the two terms would be interchangeable. I'm personally used to `cwd`, whether in APIs or shell, because of PHP and Python, that use `getcwd()` and `os.getcwd()` in their standard libraries.

> This works fine with the default working dir
> 
> ```
> ddev config --web-working-dir /var/www/html
> ```
> 
> But not with something else like
> 
> ```
> ddev config --web-working-dir /var/www/html/test
> ```

Oh, I wasn't familiar with that option. Perhaps it should match explicitly to the `docroot`? In that case, the option could explain itself to be limited to equivalents _inside_ the `docroot` only (?), which I believe is the most useful aspect of this.

Perhaps the counter argument to making it more flexible, like allowing anything under `working_dir` (now that I know the option) or `$HOME` or all of the above, is that it would make it unpredictable (?). I have limited familiarity with more complex `ddev` use cases so I might be ignoring obvious conflicts, or opportunities, here.

Can you suggest what would be a preferred, or more predictable, criteria? My intuition is to choose the behavior that is the most predictable, less surprising, to the user (`docroot` only?). But again I'm limited to very simple use cases (`ddev start` of whatever was provided to me), so I might be overlooking something obvious.

Thanks!

rfay:
`ddev ssh` once used to land in the docroot, but then things like doing `composer install` don't typically work in the docroot. Etc. That's why it lands in the project root.

diegoe:
> Put that in ~/.ddev/commands/web/sshhere and `ddev restart` and then use `ddev sshhere`, for example:

This works perfectly for what I need. Thanks 👍 

Feel free to close this PR and issue if you don't see it fitting into the command itself.

rfay:
We have a number of commands like `ssh` and `exec` and `share` that predate shell-based commands... and would be better reimplemented the other way.

rfay:
Thanks for your work on this and for taking the time to create a contribution PR. I hope you'll do it again!  (Next time, please do an issue before you invest in a lot of work in the PR so you can make sure the conversation gets done before you do the work. Although most of the time it's not this much work.)

Since this adds complexity and there's already a `ddev ssh -d` option, and since you have a good workaround, I think I'm going to go ahead and close it. 

It still would have needed updates to docs, and the `cwd` vs `pwd` conversation would have been confusing in the long term. (Note that golang uses `os.Getwd()` to skip the whole conversation :) )

PR #5897, opened 2024-02-26
test: make sure buildkite leaves docker in sane setup

rfay:
## The Issue

We're having a little trouble with buildkite/macos cleanup, especially when it seems like Colima doesn't start right

Try to simplify the startup clean situation
* Make sure everything is stopped, then orbstack or Docker Desktop (or some docker provider) running
* Make sure we are on a context with docker running

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -52,7 +52,7 @@ if [ "${OSTYPE%%[0-9]*}" = "darwin" ]; then
       ;;
 
     "docker-desktop")
-      open -a Docker &
+      open -a /Applications/Docker.app &
       docker context use desktop-linux
       ;;
 
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -1,10 +1,12 @@
 #!/bin/bash
 # This script is used to build ddev/ddev using buildkite
 
+set -eu -o pipefail
+
 export PATH=$PATH:/home/linuxbrew/.linuxbrew/bin
 
-# GOTEST_SHORT=8 means drupal9
-export GOTEST_SHORT=8
+# GOTEST_SHORT=12 means drupal0
+export GOTEST_SHORT=12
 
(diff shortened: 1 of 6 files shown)


.buildkite/test.sh
```diff
@@ -1,80 +1,67 @@
 #!/bin/bash
 # This script is used to build ddev/ddev using buildkite
 
+set -eu -o pipefail
+
 export PATH=$PATH:/home/linuxbrew/.linuxbrew/bin
 
-# GOTEST_SHORT=8 means drupal9
-export GOTEST_SHORT=8
+# GOTEST_SHORT=12 means drupal0
```
stasadev, line 8:
```suggestion
# GOTEST_SHORT=12 means drupal10
```

.buildkite/test.sh
```diff
@@ -1,80 +1,67 @@
...
+    ~/.rd/bin/rdctl shutdown || true
+    docker context use default
+    # Leave orbstack running as the most likely to be reliable
+    orb start &
     true
   }
   trap cleanup EXIT
 
-  echo "original docker context situation:"
+    # Start with a predictable setup orbstack running
```
stasadev, line 30:
```suggestion
  # Start with a predictable setup orbstack running
```

.buildkite/test.sh
```diff
@@ -1,80 +1,67 @@
...
-#      docker context use default
+    orb stop &
+    killall com.docker.backend || true
+    colima stop || true
+    colima stop vz || true
+    limactl stop lima-vz || true
+    ~/.rd/bin/rdctl shutdown || true
+    docker context use default
+    # Leave orbstack running as the most likely to be reliable
+    orb start &
```
stasadev, line 25: `orb` is not installed on all runners:

```suggestion
    command -v orb &>/dev/null && orb start &
```

stasadev (changes requested):

Issue #5898, opened 2024-02-26
ddev start failed with php-gmp due to permissions on /run/php


stasadev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
```
</details>         

### Expected Behavior

DDEV should start normally.

### Actual Behavior

```
ddev start
Waiting for web/db containers to become ready: [web]
Failed waiting for web/db containers to become ready: web container failed: log=, err=container exited, please use 'ddev logs -s web' to find out why it failed

ddev logs -s web
ERROR: Unable to create the PID file (/run/php/php8.2-fpm.pid).: Permission denied (13)
ERROR: FPM initialization failed
```

### Steps To Reproduce

Used latest HEAD with Debian Bookworm, tested this on macOS arm64 and Linux amd64:

```
mkdir php-gmp && cd php-gmp
ddev config --auto --php-version 8.2 --webimage-extra-packages php8.2-gmp --omit-containers db
ddev start
```

### Anything else?

Related issue:

- https://github.com/ddev/ddev-sqlsrv/issues/20

stasadev:
Fixed in:

- #5900

PR #5900, opened 2024-02-26
fix: set permissions on /run/php so php-fpm can run

stasadev:
## The Issue

- #5898

Although this seems to be a problem with php packaging, we can solve it on our side to prevent this in the future with `chmod 777 /run/php`.

```
mkdir php-gmp && cd php-gmp
ddev config --auto --php-version 8.2 --webimage-extra-packages php8.2-gmp --omit-containers db
ddev start
```

The project should be started normally.


rfay (approved):
I'm not sure exactly what's going wrong here, but this is fine. 

I note that in 
* #5449 

we removed checking php packages in [TestExtraPackages](https://github.com/ddev/ddev/blob/d7163e7197bc25b4c9e1ff1be69af51823608da7/pkg/ddevapp/config_test.go#L950https://github.com/ddev/ddev/blob/d7163e7197bc25b4c9e1ff1be69af51823608da7/pkg/ddevapp/config_test.go#L950), probably because of upstream instability. Perhaps we should add them back.

PR #5905, opened 2024-02-27
refactor: point go code to the stable docs

stasadev:
## The Issue

We should point go code to the stable docs after

- #5855

Updates the links.


rfay (approved):
Good job, thanks.

PR #5423, opened 2023-10-13
fix: supervisord should not bring down the whole web container when non-essential service fails, fixes #5358

mootensai:
## The Issue
When a command that runs by the Supervisor gets an error or crash, it automatically turns off the web container. Making the supervisor auto restart feature useless.

## How This PR Solves The Issue
This PR check when there's an error happen in supervisord whether nginx or apache is working properly or not. if one of them is down, kill supervisord

## Manual Testing Instructions
I cannot test this on real use case because I need the file to run when supervisord start. please guide me how to test this in my env.

## Automated Testing Overview
I cannot test this because of the same reason above.

## Related Issue Link(s)
https://github.com/ddev/ddev/issues/5358

## Release/Deployment Notes
this shouldn't affect anything else.

PR #5906, opened 2024-02-27
fix: supervisord should not bring down the whole web container when non-essential service fails, fixes #5358

### LINKED ISSUE #5358: supervisord should not bring down the whole web container when non-essential service fails

```
Running bash [-c /var/folders/k2/l6n0dl6j4dg_8x02bwvjfwrr0000gn/T/test_ddev.sh] 
======= Existing project config =========
These config files were loaded for project my-project: [/Users/yohanescandrajaya/Documents/GitHub/ddev/my-project/.ddev/config.yaml /Users/yohanescandrajaya/Documents/GitHub/ddev/my-project/.ddev/config.development.yaml /Users/yohanescandrajaya/Documents/GitHub/ddev/my-project/.ddev/config.local.yaml /Users/yohanescandrajaya/Documents/GitHub/ddev/my-project/.ddev/config.selenium-standalone-chrome.yaml]
name: my-project
type: php
docroot: web
php_version: 8.2
webserver_type: apache-fpm
webimage: ddev/ddev-webserver:20230803_php_serialize_precision
xdebug_enabled: true
database: {mariadb 10.11}
performance_mode: none
hooks: map[post-start:[map[exec-host:composer config github-oauth.github.com 2>/dev/null | xargs -I {} ddev exec "composer config github-oauth.github.com {}"] map[exec:sh d/initLocal.sh] map[exec:sh d/migrate.sh] map[exec-host:composer config github-oauth.github.com 2>/dev/null | xargs -I {} ddev exec "composer config github-oauth.github.com {}"] map[exec:sh d/initLocal.sh] map[exec:sh d/migrate.sh]] pre-start:[map[exec-host:[ ! -f .ddev/docker-compose.selenium-chrome.yaml ] || rm .ddev/docker-compose.selenium-chrome.yaml] map[exec-host:[ ! -f .ddev/docker-compose.selenium-firefox.yaml ] || rm .ddev/docker-compose.selenium-firefox.yaml] map[exec-host:sh fetchDeploy.sh] map[exec-host:[ ! -f .ddev/docker-compose.selenium-chrome.yaml ] || rm .ddev/docker-compose.selenium-chrome.yaml] map[exec-host:[ ! -f .ddev/docker-compose.selenium-firefox.yaml ] || rm .ddev/docker-compose.selenium-firefox.yaml] map[exec-host:sh fetchDeploy.sh]]]
upload_dirs: [/var/www/html/storage]
host_db_port: 59099
mailhog_port: 8025
mailhog_https_port: 8026
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 18
default_container_timeout: 120
web_extra_exposed_ports: [{web 8088 8088 8089}]
======= Creating dummy project named  tryddevproject-25810 in ../tryddevproject-25810 =========
```
</details>         

### Expected Behavior

When a command that runs by the Supervisor gets an error or crashes, it automatically restarts the app with autorestart configuration on the supervisor.

### Actual Behavior

When a command that runs by the Supervisor gets an error or crash, it automatically turns off the web container. Making the supervisor auto restart feature useless.

### Steps To Reproduce

1. create a daemon that runs by supervisord
2. ddev start
3. make the daemon crash by making an error/throw an exception
4. see in the docker dashboard that the web container is shut down

### Anything else?

I think it's related to this issue here
https://github.com/ddev/ddev/issues/1137

and this commit
https://github.com/ddev/ddev/commit/58f80232c65a961c935f85417a92c2df310a5f34

I could make a turnaround by replacing the supervisor.conf without the `[eventlistener:child_exit_monitor]` section. but supervisor behavior should work as expected out of the box with ddev since I've spent some hours getting to this turnaround. This issue does not exist in the documentation as well.

tjarrett:
## The Issue

If a user loads an extra daemon (in my case a laravel worker queue) into supervisor and that worker dies for whatever reason (in my case, because it's PHP, I don't want it too live too long... also I'm developing so I want it to reload the code base often) it takes the whole container with it.

Also Mailpit failing will bring down the whole server.

I'd like to be able to restart Laravel queue workers while developing without needing to restart the whole container.

This adjusts the the kill_supervisor.py (aka child_exit_monitor supervisor event listener) so that it does not bring down the container unless one of the core services is down (nginx, apache2, php-fpm, gunicorn). It knows which ones to check for by peaking inside of the config.yaml for the project.

This needs to go into the container as I understand it (I am new to DDEV). However you can test with the default DDEV project by:

First, put the file from this pull request at `.dev/web-build/kill_supervisor.py`.

Then add these lines to `.ddev/web-build/Dockerfile`:
```
ADD kill_supervisor.py /usr/local/bin/kill_supervisor.py
RUN chmod 777 /usr/local/bin/kill_supervisor.py
```

Do a `ddev restart`.

Then do a `ddev ssh` and try something like:
```
killall mailpit
```

Note that the container does not come down (and mailpit restarts).

Then try something like:
```
killall php-fpm
```

Note that the container does indeed come down.

You can change the `webserver_type` in `.ddev/config.yaml` to any of the allowed options and repeat the tests with the essential services.

## Automated Testing Overview

I was unsure how to do an automated test for this.

## Related Issue Link(s)

#5358

Also previous pull request that was closed at #5423 

## Release/Deployment Notes

I do not think so. I don't think we really want the container coming down if a queue worker fails. But we do want it to come down if nginx fails. This accomplishes that.

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/kill_supervisor.py
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/kill_supervisor.py
@@ -2,27 +2,74 @@
 import sys
 import os
 import signal
+import subprocess
+
 
 # From https://blog.zhaw.ch/icclab/process-management-in-docker-containers/
+# http://supervisord.org/events.html#example-event-listener-implementation
 
 def write_stdout(s):
-   sys.stdout.write(s)
-   sys.stdout.flush()
+    sys.stdout.write(s)
+    sys.stdout.flush()
 def write_stderr(s):
-   sys.stderr.write(s)
-   sys.stderr.flush()
+    sys.stderr.write(s)
+    sys.stderr.flush()
+# Function to check if a process is running using ps, grep, and awk
+def is_process_running(process_identifier):
+    try:
+        # Use ps and grep to search for the process, excluding the grep command itself
+        output = subprocess.check_output(
+            f"ps aux | grep '[{process_identifier[0]}]{process_identifier[1:]}' | grep -v grep", shell=True)
+        # If the output is not empty, then the process is considered running
+        return bool(output.decode('utf-8').strip())
+    except subprocess.CalledProcessError:
+        # If the grep command does not find anything, it will raise a CalledProcessError
+        return False
+# Function to parse the config.yaml file for the webserver_type
+def read_webserver_type_from_yaml(file_path):
+    webserver_type = None
+    try:
+        with open(file_path, 'r') as file:
+            for line in file:
+                if line.strip().startswith('webserver_type:'):
+                    webserver_type = line.split(':', 1)[1].strip()
+                    break
+    except Exception as e:
+        write_stderr(f"Error reading webserver_type from YAML: {str(e)}\n")
+    return webserver_type
 def main():
-   while 1:
-       write_stdout('READY\n')
-       line = sys.stdin.readline()
-       write_stdout('This line kills supervisor: ' + line)
-       try:
-               pidfile = open('/var/run/supervisord.pid','r')
-               pid = int(pidfile.readline())
-               os.kill(pid, signal.SIGQUIT)
-       except Exception as e:
-               write_stdout('Could not kill supervisor: ' + e.strerror + '\n')
-       write_stdout('RESULT 2\nOK')
+    # Read the webserver_type from the config.yaml
+    webserver_type = read_webserver_type_from_yaml('/var/www/html/.ddev/config.yaml')
+    critical_processes = {
+        'nginx-fpm': ['nginx: ', 'php-fpm: '],
+        'apache-fpm': ['apache2 ', 'php-fpm: '],
+        'nginx-gunicorn': ['nginx: ', 'start-gunicorn.py'],
+    }.get(webserver_type, [])
+
+    while 1:
+        write_stdout('READY\n')
+        line = sys.stdin.readline()
+        try:
+            # Check if any critical process is not running
+            process_failure = False
+            for process in critical_processes:
+                if not is_process_running(process):
+                    process_failure = True
+                    write_stderr(f'Critical process {process} for {webserver_type} is not running. Action required.\n')
+                    break
+
+            if process_failure:
+                # Proceed with original behavior if a critical process is not running
+                write_stdout('This line kills supervisor: ' + line)
+                pidfile = open('/var/run/supervisord.pid', 'r')
+                pid = int(pidfile.readline())
+                os.kill(pid, signal.SIGQUIT)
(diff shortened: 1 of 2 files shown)


rfay (commented):
Currently this does nothing until you build an image with the related behavior. 

Have you tested this approach?

Is it possible that we should just remove this behavior (stop killing the container if one of the services fails?) We could rely on the healthcheck instead of this.

tjarrett:
Right. This is working for me by injecting my updatd `kill_supervisor.py` into my container by tweaking the `Dockerfile` in the `.ddev/web-build` folder in my project. I've added these lines to to that file:
```
ADD kill_supervisor.py /usr/local/bin/kill_supervisor.py
RUN chmod 777 /usr/local/bin/kill_supervisor.py
```

I then proceeded to change the `webserver_type` setting in `.ddev/config.yaml` to each of the three options and tested by killing mailpit and noting that the container stays up but killing php causes the container to go down (as expected).

I can send logs or a screen recording or something if you would like to show it working.

As to whether or not it should be removed; I would be ok with that. I'm new-ish to Docker and entirely new to DDEV so my opinion should not count for much besides noting that I did not expect the whole container to come down because one process being run by Supervisor decided to kill itself.

Maybe it could become a setting in the `.ddev/config.yaml`? That way folks that need it could keep it but folks who do not just would not have to deal with it.

Happy to help in any way I can to resolve this though.

stasadev (commented):
I have tested this PR, and I am in favor of adding it.
I'll rebase it to run the tests.

rfay:
@stasadev it needs a pushed image and reference to the image in versionconstants.go to be relevant or for the tests to do anything.

stasadev:
Oh, right, thank you, I'll do it now.

rfay:
I suspect we're not doing this right. I think that somehow we should be using `autorestart` to our advantage, rather than making the killer script super complex (it was already too complex).

stasadev:
Yesterday I removed `kill_supervisor.py `and tested healthcheck behavior for `ddev-webserver` with `killall php-fpm`.
The result was an `unhealthy` container, but not `exited`.

Found this SO answer https://stackoverflow.com/a/63508479/8097891, maybe we need to change our healthcheck.

rfay:
> The result was an unhealthy container, but not exited

I'm not sure that "unhealthy" isn't a better situation than "exited"

rfay:
@stasadev is going to do an experimental replacement PR that just removes the killer script entirely.

stasadev:
Experimental PR:

- #5936

rfay:
* https://github.com/ddev/ddev/pull/5936

is going to succeed, yay! Closing this. We'll all be a lot happier!

tjarrett:
Thanks guys!

PR #5907, opened 2024-02-28
feat: disable password policies for TYPO3 12

kitzberger:
Since TYPO3 12 there's a new feature to help hardening the passwords of TYPO3 backend and frontend users. This makes it unnecessarily complicated to developers to log into TYPO3.

## The Issue

TYPO3 12 forces users to follow a certain password policy. That's slowing developers down when working on their local machine.

The autogenerated config for TYPO3 now makes sure that this feature is disabled, see https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/PasswordPolicies/Index.html#disable-password-policies-globally

### DIFF
--- a/pkg/ddevapp/typo3/AdditionalConfiguration.php
+++ b/pkg/ddevapp/typo3/AdditionalConfiguration.php
@@ -10,6 +10,9 @@ if (getenv('IS_DDEV_PROJECT') == 'true') {
     $GLOBALS['TYPO3_CONF_VARS'] = array_replace_recursive(
         $GLOBALS['TYPO3_CONF_VARS'],
         [
+            'BE' => [
+                'passwordPolicy' => '',
+            ],
             'DB' => [
                 'Connections' => [
                     'Default' => [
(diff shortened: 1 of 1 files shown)


rfay (approved):
This looks fine to me, thanks! I'd like to have an approval from another TYPO3 community member too.

rfay:
@bmack @ochorocho could you review this please?

schliesser:
I don't like this to be added as a default by ddev. Rather add that Snippet to your site package as shown in the docs.

ochorocho:
I agree on @schliesser suggestion. I would prefer not to remove the password policy for local development.

rfay:
Interestingly, the TYPO3 password constraints are so weak it's pretty silly. When I install a T3 site I use `junker99` and it's totally happy. For other project types I use `admin`. But `junker99` is *not* a very good password :)

rfay:
Closing, seems like this isn't the community preference at this point, but happy to reopen if comments indicate otherwise. Happy to continue the conversation here or elsewhere.

PR #5916, opened 2024-03-01
feat: invoke dbeaver inside wsl2, fixes #5375

### LINKED ISSUE #5375: Add dbeaver host command support

* #4950 

The dbeaver client is not currently supported by a ddev command.

### Describe your solution

Support a ddev dbeaver command that auto connects, saves the connection and puts the connection in the DDEV folder for DBeaver for the current running DDEV project. Assumes db name/user/password are DDEV default.

This builds off of the effort started by @MakerTim. 

Add a `dbeaver` command:

```
#!/bin/bash

## Description: Open project database in DBeaver.
## Usage: dbeaver
## Example: "ddev dbeaver"

# Note that this example uses $DDEV_HOST_DB_PORT to get the port for the connection
# DBeaver command line options can be viewed here https://dbeaver.com/docs/dbeaver/Command-Line/
# DBeaver can be obtained here https://dbeaver.io

CONNECTION="driver=mysql|user=db|password=db|database=db|host=127.0.0.1|port=${DDEV_HOST_DB_PORT}|name=localhost|savePassword=true|folder=DDEV|openConsole=true"

case $OSTYPE in
  "win*"* | "msys"*)
    '/c/Program Files/DBeaver/dbeaver.exe' -con "$CONNECTION" &
    ;;
  "linux-gnu")
    if [ -f "/mnt/c/Program Files/DBeaver/dbeaver.exe" ]; then
      # wsl2 needs cmd to start the dbeaver
      /mnt/c/Windows/System32/cmd.exe /C 'C:"Program Files"\DBeaver\dbeaver.exe -con '\"$CONNECTION\"'' 2>nul &
    fi
    if [ -f "/snap/bin/dbeaver-ce" ]; then
      "/snap/bin/dbeaver-ce" -con "$CONNECTION" &
    fi
    if [ -f "/snap/bin/dbeaver-ee" ]; then
      "/snap/bin/dbeaver-ee" -con "$CONNECTION" &
    fi
    if [ -f "/snap/bin/dbeaver" ]; then
      "/snap/bin/dbeaver" -con "$CONNECTION" &
    fi
    ;;

  "darwin"*)
    open -a dbeaver.app --args -con $CONNECTION &
    echo "Attempted to launch DBeaver.app"
    ;;
esac
```

This will:

- Open DBeaver
- Create a connection with the local DDEV parameters
- Save that connection
- Create a DDEV folder to organize this connection

Running the command will open DBeaver with the SQL console open every time, even after a restart and the published port changes.

**Tested on**: MacOS Ventura

**Needs testing**:

- Verification on Windows 7/10/11 and/or WSL2
- Verification on Linux (Fedora, Ubuntu, Debian)

I can check Fedora when I am home in two weeks... I don't have a good way to test those OS above otherwise.

image

### Describe alternatives

_No response_

### Additional context

_No response_

Brupes:
DBeaver support inside wsl2

## The Issue
We are not able to call `ddev dbeaver` inside wsl2

Command dbeaver was modified to support dbeaver community installed to everyone

Install DBeaver Community edition in windows and call `ddev dbeaver` from wsl2.
It should open dbeaver and add the connection options to current project

## Related Issue Link(s)

* #5375


rfay:
Artifacts are ready for testing, https://github.com/ddev/ddev/pull/5916#issuecomment-1972365704

Brupes:
everything working on my side 😊
image

rfay:
Yay @Brupes - so you installed the built DDEV version and it worked OK? Your test isn't valid if you had an altered `dbeaver` in your ~/.ddev/commands/host, so please make sure you delete that when testing. It should be recreated on `ddev start` with the one from the artifacts here.

Brupes:
Just checked it now. everything looks ok.
image

rfay (changes requested):
I noticed that this requires an update in the docs as well, 
* https://github.com/ddev/ddev/blob/master/docs/content/users/usage/commands.md
* https://github.com/ddev/ddev/blob/master/docs/content/users/usage/database-management.md

Brupes:
For commands
image

For database management
image

does this look ok? i am not so sure about commands documentation :/

rfay:
It looks like you're on the right track to me, thanks. Push it up!

rfay:
Unfortunately this doesn't work if Dbeaver is installed with the defaults - it ends up in `C:\Users\randy\AppData\Local\DBeaver\dbeaver.exe` for me. 

It does work if I install with "install for all users". I'm not sure how we can explain this to people, or if we can add something that adds the new path. 

But it does work fine on WSL2 if installed for all users.

Of course, folks on WSL2 may also want to use the DBeaver linux version, but we'll leave that for another time.

rfay:
Still works fine on macOS

Brupes:
hmm... investigating the issue with user profile path which is a dynamic one, i discovered it is actually possible to get it from wsl2

```bash
cmdWSL2Path=$(wslpath -au 'C:\Windows\System32\cmd.exe')
userProfileWindowsPath=$("$cmdWSL2Path" /c 'echo %USERPROFILE%')
userProfileWSL2Path=$(wslpath -au "$userProfileWindowsPath")
```

the var `userProfileWSL2Path` will contain the path to current windows profile path from wsl2

the problem is to add it to `HostBinaryExists` :/

Brupes:
i got a better way of doing it now. i wanted to remove an output that would get us trouble later because of `UNC paths` and i ended with this code now

```bash
userProfileWindowsPath=$(pushd /mnt/c > /dev/null; cmd.exe /c 'echo %USERPROFILE%'; popd > /dev/null)
userProfileWSL2Path=$(wslpath -au "$userProfileWindowsPath")
```

Brupes:
This is working. The problem now is how i can add this path to `HostBinaryExists`...

```bash
case $OSTYPE in
  "linux-gnu")
    # Check for different binaries. Launch the first one found.
    BINARIES=(
      /usr/bin/dbeaver{,-ce,-le,-ue,-ee}
      /var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity
      /snap/bin/dbeaver-ce
      '/mnt/c/Program Files/dbeaver/dbeaver.exe'
    )
    for binary in "${BINARIES[@]}"; do
      if [ -x "$binary" ]; then
        echo "Launching $binary"
        "$binary" -con "$CONNECTION" &> /dev/null & disown
        exit 0
      fi
    done

    # Check for different binaries inside user profile wsl2. Launch the first one found.
    if [ -x "/mnt/c/Windows/System32/cmd.exe" ]; then
      BINARIES=(
        AppData/Local/dbeaver/dbeaver.exe
      )
      prefix="$(wslpath -au "$(pushd /mnt/c > /dev/null; cmd.exe /c 'echo %USERPROFILE%' | tr -d '\r\n'; popd > /dev/null)")"
      for binary in "${BINARIES[@]}"; do
        if [ -x "$prefix/$binary" ]; then
          echo "Launching $prefix/$binary"
          "$prefix/$binary" -con "$CONNECTION" &> /dev/null & disown
          exit 0
        fi
      done
    fi
    ;;
  "darwin"*)
    open -a dbeaver.app --args -con "$CONNECTION" &
    echo "Attempted to launch DBeaver.app"
    ;;
esac
```

image

rfay:
I think it's probably OK to go with the global installation.  Add something to https://ddev.readthedocs.io/en/latest/users/usage/commands/#dbeaver that explains that it has to be the global installation.

Brupes:
sorry the delay. i was working in other stuff 😅
here you have. just added "to all users". think it's good. the path is already there to reinforce that soo... 😊

Open [DBeaver](https://dbeaver.io/) with the current project’s database (global shell host container command). This command is only available if `DBeaver.app` is installed as `/Applications/DBeaver.app` for macOS, if `dbeaver.exe` is installed to all users as `C:/Program Files/dbeaver/dbeaver.exe` for WSL2, and if `dbeaver` (or another binary like `dbeaver-ce`) available inside `/usr/bin` for Linux (Flatpak and snap support included).

rfay (approved):
Thanks, looks good to me! It was testing fine before, I assume nothing has changed that makes me do manual test again.

Brupes:
> Thanks, looks good to me! It was testing fine before, I assume nothing has changed that makes me do manual test again.

Same. i just updated the documentation files. it should have the same behaviour as before

PR #5888, opened 2024-02-23
feat: add cakephp project type, fixes #5713

ajibarra:
## The Issue

CakePHP projects must use  generic `php` type which does not allow automatic configuration of settings, access to cake shell through command, etc

## How This PR Solves The Issue

This PR adds cakephp project type to DDEV

## Manual Testing Instructions

```bash
    mkdir my-cakephp-app
    cd my-cakephp-app
    ddev config --project-type=cakephp --docroot=webroot --create-docroot --php-version=8.3
    ddev composer create--prefer-dist cakephp/app:~5.0
    ddev composer install
    ddev cake
    ddev launch
```

## Automated Testing Overview

* Added `cakephp` app type to `pkg/ddevapp/ddevapp_test.go`
* Added `cake` command to `cmd/ddev/cmd/commands_test.go`

## Related Issue Link(s)

#5713 

## Release/Deployment Notes

* CakePHP test app: https://github.com/ddev/test-cakephp

PR #5923, opened 2024-03-01
fix: don't call app.WriteConfig twice

rfay:
## The Issue

@ajibarra points out in 
* https://github.com/ddev/ddev/pull/5888#issuecomment-1972943332 

that the PostConfigAction is being called twice. It's being called twice because when `ddev config` is called with args, the WriteConfig() happens both inside handleMainConfigArgs() and also in the parent handleConfigRun. It should only be called once. 

* When `ddev config` is called without args, we do the WriteConfig() in its context
* When called *with* args, let handleMainConfigArgs() do the write

I think it requires debugger or adding a print statement, but just add a print statement to a PostConfigAction or to WriteConfig.

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -171,6 +171,10 @@ func handleConfigRun(cmd *cobra.Command, args []string) {
 		if err != nil {
 			util.Failed("There was a problem configuring your project: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


rfay:
This could also be done with the opposite logic.

stasadev (approved):
Tested, this is fine.

PR #5926, opened 2024-03-04
refactor: Use project_list.yaml as source of truth, fixes #5918

### LINKED ISSUE #5918: Move to use project_list.yaml exclusively for project list
### Expected Behavior

We don't want duplicates of project list

In 
* https://github.com/ddev/ddev/pull/5651

the ~/.ddev/project_list.yaml approach was created, but it duplicates ~/.ddev/global_config.yaml

Now sometimes people try to delete from one and can't sort out problems they find. 

Another thing to take a look at is whether we want to move anything else out of ~/.ddev/global_config.yaml. 

Things that aren't really config items but actually managed by DDEV:

* mkcert_caroot
* last_started_version

GuySartorelli:
## The Issue
Users have been confused by having two copies of the project list, and haven't known which to update if they need to manually update the list.

1. If there's no `project_list.yaml` file and the `global_config.yaml` file contains a project list (i.e. if someone is upgrading from an older version of DDEV), copy the list into `project_list.yaml`
    - This is a departure from the previous behaviour which would copy the list from `global_config.yaml` if there was any difference between the two lists
1. Remove `ProjectList` from the `GlobalConfig` type
    - This means the next time `global_config.yaml` is written, the project list will be removed from that file.

Note that I have done this in two commits to show why some of the changes are necessary. I'll add PR comments to help clarify this.

1. Ensure you have at least one project in your project list prior to checking out this PR
1. Remove `project_list.yaml` if you have it
1. Run `ddev list` - it should list the project(s) correctly, regardless of what state they were in
1. Confirm that you now have a `project_list.yaml` file
1. Run any command that will result in global config being written (e.g. if your `last_started_version` has changed, run `ddev start` in a project)
1. Confirm that `global_config.yaml` no longer has a list pf projects

## Automated Testing Overview

No automated tests need to change - they were all already relying on `project_list.yaml`.

## Related Issue Link(s)
- https://github.com/ddev/ddev/issues/5918

## Release/Deployment Notes

This shouldn't affect anyone other than reducing confusion.

### DIFF
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -514,6 +514,8 @@ func TestConfigDatabaseVersion(t *testing.T) {
 	// add a database into the config and nothing else
 	for _, dbTypeVersion := range versionsToTest {
 		_ = app.Stop(true, false)
+		_, err := exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
+		assert.NoError(err)
 		parts := strings.Split(dbTypeVersion, ":")
 		err = os.RemoveAll(filepath.Join(testDir, ".ddev"))
 		assert.NoError(err)
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -514,8 +514,6 @@ func TestConfigDatabaseVersion(t *testing.T) {
 	// add a database into the config and nothing else
 	for _, dbTypeVersion := range versionsToTest {
 		_ = app.Stop(true, false)
-		_, err := exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
-		assert.NoError(err)
 		parts := strings.Split(dbTypeVersion, ":")
 		err = os.RemoveAll(filepath.Join(testDir, ".ddev"))
 		assert.NoError(err)
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
@@ -75,7 +74,8 @@ type GlobalConfig struct {
 	WSL2NoWindowsHostsMgt            bool                        `yaml:"wsl2_no_windows_hosts_mgt"`
 	WebEnvironment                   []string                    `yaml:"web_environment"`
 	XdebugIDELocation                string                      `yaml:"xdebug_ide_location"`
-	ProjectList                      map[string]*ProjectInfo     `yaml:"project_info"`
+	// ProjectList is retained to make upgrading easier and should be removed after a few releases
+	ProjectList map[string]*ProjectInfo `yaml:"project_info"`
 }
 
 // New() returns a default GlobalConfig
@@ -93,9 +93,10 @@ func New() GlobalConfig {
 		NoBindMounts:                 nodeps.NoBindMountsDefault,
 		Router:                       types.RouterTypeDefault,
 		MkcertCARoot:                 readCAROOT(),
-		ProjectList:                  make(map[string]*ProjectInfo),
-		TraefikMonitorPort:           nodeps.TraefikMonitorPortDefault,
-		ProjectTldGlobal:             nodeps.DdevDefaultTLD,
+		// ProjectList is retained to make upgrading easier and should be removed after a few releases
+		ProjectList:        nil,
+		TraefikMonitorPort: nodeps.TraefikMonitorPortDefault,
+		ProjectTldGlobal:   nodeps.DdevDefaultTLD,
 	}
 
 	return cfg
@@ -308,6 +309,11 @@ func WriteGlobalConfig(config GlobalConfig) error {
 		return err
 	}
 
+	// Remove empty project list from global config
+	// In a future release it will be removed from the struct but
+	// for now it's needed to make upgrading easier.
+	cfgbytes = []byte(strings.Replace(string(cfgbytes), "project_info: {}", "", 1))
+
 	// Append current image information
 	instructions := `
 # You can turn off usage of the
@@ -465,6 +471,7 @@ func WriteGlobalConfig(config GlobalConfig) error {
 // Or creates the file
 func ReadProjectList() error {
 	globalProjectsFile := GetProjectListPath()
+	fileAlreadyExists := true
 
 	// Can't use fileutil.FileExists() here because of import cycle.
 	if _, err := os.Stat(globalProjectsFile); err != nil {
@@ -475,6 +482,7 @@ func ReadProjectList() error {
 			return nil
 		}
 		if os.IsNotExist(err) {
+			fileAlreadyExists = false
 			// write an empty file
 			err := os.WriteFile(GetProjectListPath(), make([]byte, 0), 0644)
 			if err != nil {
@@ -496,13 +504,22 @@ func ReadProjectList() error {
 		return err
 	}
 
-	// For backwards compatability we're keeping the project_list in global config
-	// in sync with the list in project_list.yaml. If someone upgrades from an earlier
-	// version the global config will have correct content that isn't in projects.yaml.
-	// For now, treat global config as the source of truth when the two lists differ.
-	if !reflect.DeepEqual(DdevGlobalConfig.ProjectList, DdevProjectList) {
+	// Check legacy global config projects list for easier upgrade.
+	// If someone upgrades from an earlier version the global config will have
+	// correct content that isn't in projects.yaml.
+	// If the project list file didn't exist and the global config has a project list,
+	// assume global config is correct.
+	if !fileAlreadyExists && DdevGlobalConfig.ProjectList != nil && len(DdevGlobalConfig.ProjectList) > 0 {
 		DdevProjectList = DdevGlobalConfig.ProjectList
 		err := WriteProjectList(DdevProjectList)
+		if err != nil {
+			return err
+		}
+	}
+	if DdevGlobalConfig.ProjectList != nil && len(DdevGlobalConfig.ProjectList) > 0 {
+		// Clear global config project list - in a future release it won't be used at all.
+		DdevGlobalConfig.ProjectList = nil
+		err = WriteGlobalConfig(DdevGlobalConfig)
 		if err != nil {
 			return err
 		}
@@ -513,15 +530,6 @@ func ReadProjectList() error {
 
 // WriteProjectList writes the global projects list into ~/.ddev.
 func WriteProjectList(projects map[string]*ProjectInfo) error {
-	// Write to global config for backwards compatability.
-	// This allows devs to downgrade to an earlier version without
-	// worrying about copying project info into their global config file.
-	DdevGlobalConfig.ProjectList = projects
-	err := WriteGlobalConfig(DdevGlobalConfig)
-	if err != nil {
-		return err
-	}
-
 	// Prepare projects file content
 	projectsBytes, err := yaml.Marshal(projects)
 	if err != nil {
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -691,12 +691,9 @@ func RemoveProjectInfo(projectName string) error {
 	_, ok := DdevProjectList[projectName]
 	if ok {
 		delete(DdevProjectList, projectName)
-		err := WriteProjectList(DdevProjectList)
-		if err != nil {
-			return err
-		}
 	}
-	return nil
+	err := WriteProjectList(DdevProjectList)
+	return err
 }
 
 // GetGlobalProjectList returns the global project list map
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -500,12 +500,11 @@ func ReadProjectList() error {
 			}
 
 			// Write an empty file - we have no known projects
-			err := os.WriteFile(GetProjectListPath(), make([]byte, 0), 0644)
-			// Whether there's an error or nil here we want to return
-			return err
-		} else {
-			return err
+			err = os.WriteFile(GetProjectListPath(), make([]byte, 0), 0644)
 		}
+
+		// Whether there's an error or nil here we want to return
+		return err
 	}
 
 	source, err := os.ReadFile(globalProjectsFile)
--- a/pkg/globalconfig/global_config.go
+++ b/pkg/globalconfig/global_config.go
@@ -32,8 +32,6 @@ var (
 	DdevGlobalConfig GlobalConfig
 	// DdevProjectList is the list of all existing DDEV projects
 	DdevProjectList map[string]*ProjectInfo
-	// Old versions of DDEV stored the project list in the global config.
-	ddevLegacyGlobalConfig LegacyGlobalConfig
 )
 
 type ProjectInfo struct {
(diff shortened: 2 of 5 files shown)


pkg/globalconfig/global_config.go
```diff
@@ -308,6 +309,11 @@ func WriteGlobalConfig(config GlobalConfig) error {
 		return err
 	}
 
+	// Remove empty project list from global config
+	// In a future release it will be removed from the struct but
+	// for now it's needed to make upgrading easier.
+	cfgbytes = []byte(strings.Replace(string(cfgbytes), "project_info: {}", "", 1))
```
GuySartorelli: This is to ensure we don't have an **empty** project list in global config. It felt quite hacky, but was better than keeping an empty list in the file which would lead to further confusion.

pkg/globalconfig/global_config.go
```diff
@@ -75,7 +74,8 @@ type GlobalConfig struct {
 	RouterMailpitHTTPSPort           string                      `yaml:"mailpit_https_port,omitempty"`
 	Messages                         MessagesConfig              `yaml:"messages,omitempty"`
 	RemoteConfig                     RemoteConfig                `yaml:"remote_config,omitempty"`
-	ProjectList                      map[string]*ProjectInfo     `yaml:"project_info"`
+	// ProjectList is retained to make upgrading easier and should be removed after a few releases
+	ProjectList map[string]*ProjectInfo `yaml:"project_info"`
```
GuySartorelli: I didn't like keeping this in the type since it's no longer intended to be kept there, but this is how we'd do this if we don't have an intermediary legacy type.

pkg/globalconfig/global_config.go
```diff
@@ -74,7 +76,13 @@ type GlobalConfig struct {
...
-	// ProjectList is retained to make upgrading easier and should be removed after a few releases
+}
+
+// Old versions of DDEV stored the project list in the global config.
+// For ease of upgrade, we need to be able to extract that data and move it to
+// the new project_list.yaml
+// In a future release, remove this.
+type LegacyGlobalConfig struct {
 	ProjectList map[string]*ProjectInfo `yaml:"project_info"`
 }
```
GuySartorelli, line 87: This lets us remove `ProjectList` from the `GlobalConfig` type. It means any time the global config is saved, the project list (if there was one in the file) is removed without having to do something hackish like an above comment points out.

It also gives us a clean way to marshal the list from the YAML file - if we removed `ProjectList` from `GlobalConfig` but *didn't* add something like this, we'd have to parse the file manually which would be even worse.

pkg/globalconfig/global_config.go
```diff
@@ -239,6 +248,12 @@ func ReadGlobalConfig() error {
 		return err
 	}
 
+	// Get the legacy project list if there is one
+	err = yaml.Unmarshal(source, &ddevLegacyGlobalConfig)
+	if err != nil {
+		return err
+	}
```
GuySartorelli, line 255: We do this in `ReadGlobalConfig` because this is the best point to guarantee that:
1. We haven't overwritten the config file yet
2. We haven't run `ReadProjectList` yet

pkg/globalconfig/global_config.go
```diff
@@ -475,11 +491,18 @@ func ReadProjectList() error {
...
-			// write an empty file
-			err := os.WriteFile(GetProjectListPath(), make([]byte, 0), 0644)
-			if err != nil {
+			// If someone upgrades from an earlier version, the global config may hold the project list.
+			if ddevLegacyGlobalConfig.ProjectList != nil && len(ddevLegacyGlobalConfig.ProjectList) > 0 {
+				DdevProjectList = ddevLegacyGlobalConfig.ProjectList
+				err := WriteProjectList(DdevProjectList)
+				// Whether there's an error or nil here we want to return
 				return err
 			}
```
GuySartorelli, line 500: We only want to check legacy config if the file doesn't exist. If the file exists, it's the source of truth even if it's empty.

pkg/globalconfig/global_config.go
```diff
@@ -475,14 +491,20 @@ func ReadProjectList() error {
...
-		} else {
-			return err
+
+			// Write an empty file - we have no known projects
+			err = os.WriteFile(GetProjectListPath(), make([]byte, 0), 0644)
 		}
+
+		// Whether there's an error or nil here we want to return
+		return err
```
GuySartorelli, line 507: We either created a new file here, or we had an unexpected error. Either way there's no new projects to read so we can skip the "read the projects list from the file" logic below.

rfay:
Moved to draft until tests start working

rfay:
This is essential for DDEV v1.23.0 IMO, thanks for all your work on it, and we know you don't have unlimited time for it.

rfay:
Rebased

GuySartorelli:
Rebased again 'cause GitHub said it was still out of date with Master.

From what I can tell, the failing unit tests are failing because test logic isn't quite working - not because runtime logic isn't quite working. In fact it seems to be relying on a bug that this PR inadvertently fixes.

The error can be easily reproduced by running
```bash
GOTEST_SHORT=1 make test  TESTARGS="-run TestConfigDatabaseVersion"
```

```text
    config_test.go:521: 
        	Error Trace:	/home/gsartorelli/dump/temp/ddev/cmd/ddev/cmd/config_test.go:521
        	            	exit status 1
        	Test:       	TestConfigDatabaseVersion
        	Messages:   	Failed to run ddev config --database mariadb:10.11: this project root /home/gsartorelli/tmp/ddevtest/TestConfigDatabaseVersion3259693668 already contains a project named TestConfigDatabaseVersion. You may want to remove the existing project with "ddev stop --unlist TestConfigDatabaseVersion"
```

If we keep this part of the code from `ReadProjectList()` (along with keeping `ProjectList` in the `GlobalConfig` struct), the tests pass:
```go
if !reflect.DeepEqual(DdevGlobalConfig.ProjectList, DdevProjectList) {
	DdevProjectList = DdevGlobalConfig.ProjectList
	err := WriteProjectList(DdevProjectList)
	if err != nil {
		return err
	}
}
```

In other words, something in the tests is either writing directly to the project list in global config, or writing *the wrong thing* directly to the new project list. The two get out of sync, so it grabs everything from the global config and discards whatever's in the new list.

I **think** the problem might lie in `RemoveProjectInfo` somehow? Or maybe its in the logic of the `TestConfigDatabaseVersion` test logic? I'm not quite sure, and could use a hand tracing this.

GuySartorelli:
I tried adding
```go
err = globalconfig.RemoveProjectInfo(t.Name())
assert.NoError(err)
```
into the `for _, dbTypeVersion := range versionsToTest {` loop just after running `app.stop()` but that didn't work.

Appending `time.Now().String()` after `t.Name()` for the project name in the `ddev config` command in the test also doesn't work. It still complains about an existing project named "TestConfigDatabaseVersion". It seems that it's not the name of the project being created that matters.

GuySartorelli:
Changing the name of the **first** project that's created in the test changes the name of the project that is included in the error message. So it's clearly explicitly complaining that that first-created project exists and points to the same directory that the `config` call in the loop is trying to create a project in.

I just for the life of me don't seem to be able to find a way to remove that project correctly before creating the next one.

GuySartorelli:
Okay, I have something that passes that test locally now.
Fingers crossed.

GuySartorelli:
`auth-ssh_test.go` is failing, but only in one CI run... any guesses as to whether that's intermittent and would clear up if it was re-run?

rfay:
I re-ran the nginx-fpm test. It's always good to understand what's going on with a fail, but you're always welcome to restart.

GuySartorelli:
I only have the ability to restart the buildkite builds - I don't seem to have the right permissions for restart gha ones.

rfay:
Sent you invitation. Hopefully that will give you the power.

GuySartorelli:
Aha! Yes, thank you, I have the option now for next time 💙

GuySartorelli:
All green! 🥳

tyler36:
Great work @GuySartorelli!

cmd/ddev/cmd/config_test.go
```diff
@@ -514,6 +514,8 @@ func TestConfigDatabaseVersion(t *testing.T) {
 	// add a database into the config and nothing else
 	for _, dbTypeVersion := range versionsToTest {
 		_ = app.Stop(true, false)
+		_, err := exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
```
rfay, line 517: I'm not sure how this does any harm, but what did you hit here that made you do this? Is it related to this PR? Did some old project get left laying around?

cmd/ddev/cmd/config_test.go
```diff
@@ -514,6 +514,8 @@ func TestConfigDatabaseVersion(t *testing.T) {
...
+		assert.NoError(err)
```
rfay, line 518: I've been leaning more toward require() lately, which stops the test. Less time chasing irrelevant things. It's not necessary to accept this, I'm just commenting.
```suggestion
		require.NoError(t, err)
```

pkg/globalconfig/global_config.go
```diff
@@ -33,6 +32,8 @@ var (
 	DdevGlobalConfig GlobalConfig
 	// DdevProjectList is the list of all existing DDEV projects
 	DdevProjectList map[string]*ProjectInfo
+	// Old versions of DDEV stored the project list in the global config.
+	ddevLegacyGlobalConfig LegacyGlobalConfig
```
rfay, line 36: Interesting technique, but couldn't we just allow the old ProjectList in current config, but not write it out?

pkg/globalconfig/global_config.go
```diff
@@ -75,7 +76,14 @@ type GlobalConfig struct {
...
 	RemoteConfig                     RemoteConfig                `yaml:"remote_config,omitempty"`
-	ProjectList                      map[string]*ProjectInfo     `yaml:"project_info"`
+}
+
+// Old versions of DDEV stored the project list in the global config.
+// For ease of upgrade, we need to be able to extract that data and move it to
+// the new project_list.yaml
+// In a future release, remove this.
+type LegacyGlobalConfig struct {
+	ProjectList map[string]*ProjectInfo `yaml:"project_info"`
```
rfay, line 86: As above, tempted to allow the ProjectList in current global config, just not write it (just make sure it's empty and `omitempty`

rfay (changes requested):
I'm sure this is fine, and will run through a quick test, but would love to discuss/consider just doing

```yaml
	ProjectList                      map[string]*ProjectInfo     `yaml:"project_info,omitempty"`
```

and make sure that it gets emptied after processing.

cmd/ddev/cmd/config_test.go
GuySartorelli: See https://github.com/ddev/ddev/pull/5926#issuecomment-2013819550

pkg/globalconfig/global_config.go
GuySartorelli: I assume that's what the `omitempty` does in https://github.com/ddev/ddev/pull/5926#pullrequestreview-1956645524?

If so, I have no problem doing that. I just wasn't aware that was an option.
When I tried doing it without `omitempty` it ended up with an empty map in the yaml file which is obviously misleading - for anyone who looks at that file it'd look like DDEV has incorrectly said there's no projects!

rfay:
I did manually test and look through, and it works perfectly as is.

cmd/ddev/cmd/config_test.go
GuySartorelli: Might be worth making a push to change all of them at some stage and get CI to catch that. 
I'll change this if you want me to use the `omitempty` discussed in all the other comments - but if you're happy leaving that as is then I'll leave this as well.

GuySartorelli:
Let me know if you want the `omitempty` thing or not - it's unclear from your comments whether it's fine as is or whether that's blocking merging.

cmd/ddev/cmd/config_test.go
rfay: It's not important. You're just such an important contributor that I'm doing a bit of coaching. Nothing wrong with what's here.

cmd/ddev/cmd/config_test.go
rfay: Could you please revisit this without the delete here to see if you can find out what the real problem is? It appears that you're having to patch something that is caused by something entirely different.  This kind of adaptation of a test is really suspicious, as you know.

rfay:
> Let me know if you want the omitempty thing or not

I would really appreciate it if you could experiment with that approach. I think it might be cleaner.

cmd/ddev/cmd/config_test.go
GuySartorelli: It looks to me like this code is just trying to call `ddev config` inside a directory which already has a ddev project created and that this should have been failing previously

GuySartorelli:
> > Let me know if you want the omitempty thing or not
> 
> I would really appreciate it if you could experiment with that approach. I think it might be cleaner.

Sweet, I'll try make some time this week to get that done.

cmd/ddev/cmd/config_test.go
rfay: Except... it doesn't seem to have failed except in this PR. Thanks!

GuySartorelli: Yup. The bit of code I mentioned in https://github.com/ddev/ddev/pull/5926#issuecomment-2013819550 is being removed in this PR - if we keep that there it still works, because there's no projects in the global config list and there _are_ projects in the new projects list. That discrepancy combined with the mentioned code results in clearing the new project list, so as far as DDEV is concerned there's no project there.

This PR removes that code, which means DDEV is now fully aware that there *is* a project there, causing the error.

I'm not sure if this was something introduced when I added the new project list file, or since then, or maybe I'm just wrong about the cause here? But that's what it looks like to me.

rfay: Thanks for chasing it. I'm often unable to follow your references, too many context switches.  If you could re-mention things instead of linking to other comments maybe I'll do better. It sounds like you're saying that another test has left a project around instead of doing an `app.Stop(true, false)` or something.

GuySartorelli: Not another test - this same test. It's running `ddev config` in a loop in the same directory. It _is_ doing `app.Stop(true, false)` but that seems to be insufficient to remove the project from the project list file.

rfay: If app.Stop(true,false) isn't removing the project from the project list file, it's an introduced bug. Every single test would fail if that were common. Please step through in debugger and figure out what's going on. Thanks!

GuySartorelli: Good to know - I'll introduce a new test that explicitly tests `app.Stop(true,false)` and debug it.
I'll try to find time this week to do this.

rfay:
Rebased, fixed merge conflicts.

stasadev:
> I think the problem might lie in RemoveProjectInfo somehow?

@GuySartorelli, your idea was in the right direction, the project list was not updated, when it was needed.

That's because we had `globalconfig.WriteGlobalConfig()` inside `app.Stop()` - this statement is responsible for updating the project list.

I rewrote `RemoveProjectInfo` and removed `globalconfig.WriteGlobalConfig()` from `app.Stop()` - this is not needed anymore.

> I'm sure this is fine, and will run through a quick test, but would love to discuss/consider just doing `omitempty`

Done, running tests.

stasadev (approved):

rfay (approved):
I reviewed this again and manually tested, and it looks great to me! Thanks so much! 🙏🏼

PR #5939, opened 2024-03-06
build: bump docker-compose to v2.25.0

rfay:
## The Issue

docker-compose v2.25.0 was released. Run tests on it.

### DIFF
--- a/pkg/ddevapp/traefik.go
+++ b/pkg/ddevapp/traefik.go
@@ -33,51 +33,33 @@ func detectAppRouting(app *DdevApp) ([]TraefikRouting, error) {
 	if services, ok := app.ComposeYaml["services"]; ok {
 		for serviceName, s := range services.(map[string]interface{}) {
 			service := s.(map[string]interface{})
-			var envMap map[string]interface{}
-			if envInterface, ok := service["environment"]; ok {
-				switch env := envInterface.(type) {
-				// Backward compatibility for older docker-compose
-				case map[string]interface{}:
-					envMap = env
-				// The config structure changed in docker-compose v2.24.7
-				// See https://github.com/docker/compose/issues/11589
-				case []interface{}:
-					envMap = make(map[string]interface{})
-					for _, val := range env {
-						if str, ok := val.(string); ok {
-							parts := strings.SplitN(str, "=", 2)
-							if len(parts) == 2 {
-								envMap[parts[0]] = parts[1]
-							}
-						}
+			if env, ok := service["environment"].(map[string]interface{}); ok {
+				var virtualHost string
+				var ok bool
+				if virtualHost, ok = env["VIRTUAL_HOST"].(string); ok {
+					util.Debug("VIRTUAL_HOST=%v for %s", virtualHost, serviceName)
+				}
+				if virtualHost == "" {
+					continue
+				}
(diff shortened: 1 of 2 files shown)


rfay:
Wow, completely 100% failed!

rfay:
It seems that the router is broken and is sending the request elsewhere? Maybe some labels are messed up or something.

* Using the 127.0.0.1 URL provided by `ddev start` works fine
* `docker exec -it ddev-router bash` and `curl http://ddev-d10-web` works fine
* Watching logs from `docker logs -f ddev-router` when doing `curl https://d10.ddev.site` gets a 404 and
```
172.24.0.1 - - [07/Mar/2024:00:47:21 +0000] "GET / HTTP/2.0" 404 19 "-" "-" 139 "-" "-" 0ms
```

I think the router is probably trying to hit itself?

Oh, I see, the name resolution is wrong. It's trying to hit 172.24.0.1 for ddev-d10-web, and `ddev-d10-web` actually has the address 172.24.0.3

rfay:
It works fine with docker-compose v2.24.6. `ddev config global --required-docker-compose-version=v2.24.6`

But nothing at all works with 2.24.7. 

Unless it's the networks change in https://github.com/docker/compose/releases/tag/v2.24.7 I don't know what it would be. git bisect may be the path to study it (of compose). Or just wait for issues to show up in their queue?

stasadev:
I tested commits for v2.24.7 docker-compose release.

The bug is somewhere in:

- https://github.com/docker/compose/pull/11556/

rfay:
Thanks for the research!

I think it's actually breaking something about ddev-router/traefik usage.

stasadev:
> Or just wait for issues to show up in their queue?

Yes, found a bug report for this exact problem 😅:

- https://github.com/docker/compose/issues/11589

stasadev:
> I think it's actually breaking something about ddev-router/traefik usage.

Compared the contents of `~/.ddev/.router-compose-full.yaml`:

v2.24.5:

```yaml
networks:
  ddev_default:
    external: true
    name: ddev_default
```

v2.24.7

```yaml
networks:
  ddev_default:
    external: true
    name: ddev_default
  default:
    name: ddev-d10_default
```

and when you start another project:

```yaml
networks:
  ddev_default:
    external: true
    name: ddev_default
  default:
    name: ddev-project_default
```

rfay:
Thanks for the great chasing @stasadev 

Closing this one for now, we'll watch 
* https://github.com/docker/compose/issues/11589

stasadev:
Reopening with rebase and trying to fix our code for docker-compose v2.24.7

stasadev:
There is still a [bug](https://github.com/ddev/ddev/pull/5939#issuecomment-1984037083) with the `default` network, which is created when we are not asked to do so, but this should not prevent the code from running:

stasadev:
It still doesn't work because I made changes in only one place, the fix for the output should be made in:

https://github.com/ddev/ddev/blob/a4bf20b2246790a607b3a0b2fa12e657af5ab796/pkg/ddevapp/compose_yaml.go#L76

rfay:
I guess we should go ahead and go with this? Or should we try to adopt their new format at this time?

stasadev (approved):
I think we should update docker-compose to v2.25.0.

PR #5945, opened 2024-03-07
build: add review in gitpod comment since badge doesn't work any more [skip ci]

rfay:
## The Issue

We used to get a badge on every PR with the "review in gitpod" but they seem to have lost it.

Add the gitpod comment and see if it happens.


rfay:
Opened
* https://github.com/gitpod-io/gitpod/issues/19519

Closing this as it clearly demonstrates the problem

PR #5948, opened 2024-03-08
docs: change recommendation for colima to vz

rfay:
## The Issue

* VZ is the most performant and best behaved setup for Colima at this point
* We should only be testing VZ on PRs and then test both QEMU and VZ on master

* Change docs
* Manually changed config on colima in buildkite to run on master and not on PRs.

Run through the install. Have done it many times and currently using it.

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -32,7 +32,7 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     ```
     colima start --cpu 4 --memory 6 --disk 100 --vm-type=vz --mount-type=virtiofs --dns=1.1.1.1
     ```
-    (On macOS versions before Ventura, use `colima start --cpu 4 --memory 6 --disk 100 --mount-type=sshfs --dns=1.1.1.1` as the other options are not supported on older macOS versions.)
+    (On macOS versions before Ventura, use `colima start --cpu 4 --memory 6 --disk 100 --mount-type=sshfs --dns=1.1.1.1` as the `--vm-type` option are not supported for older macOS versions.)
 
     After the initial run above, you can use `colima start` or use `colima start -e` to edit the configuration file. Run `colima status` at any time to check Colima’s status.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/install/docker-installation.md
```diff
@@ -29,22 +29,22 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
     2. Install Colima with `brew install colima`.
     3. Start Colima with 4 CPUs, 6GB memory, 100GB storage, and Cloudflare DNS, adjusting as needed:
     ```
-    colima start --cpu 4 --memory 6 --disk 100 --vm-type=qemu --mount-type=sshfs --dns=1.1.1.1
+    colima start --cpu 4 --memory 6 --disk 100 --vm-type=vz --mount-type=virtiofs --dns=1.1.1.1
     ```
-    (On macOS versions before Ventura, omit the `--vm-type=qemu` flag as it doesn't work on older OS versions.)
+    (On macOS versions before Ventura, use `colima start --cpu 4 --memory 6 --disk 100 --mount-type=sshfs --dns=1.1.1.1` as the other options are not supported on older macOS versions.)
```
rpkoller, line 34:
```suggestion
    On versions before macOS Ventura, where Virtiofs was not available, use QEMU instead.
    ```
    colima start --cpu 4 --memory 6 --disk 100 --vm-type=qemu --mount-type=sshfs --dns=1.1.1.1
    ```
```

docs/content/users/install/docker-installation.md
```diff
@@ -29,22 +29,22 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
+        In Colima you can increase—but not decrease—the disk allocation by editing `~/.colima/default/colima.yaml` to change the `disk` setting to a higher value. For example, `disk: 200` will increase allocation to 200 gigabytes. Then `colima restart` will create the new disk allocation.
 
     !!!warning "Docker contexts let the Docker client point at the right Docker server"
         Colima activates its own Docker context to prevent conflicts with Docker Desktop. If you run `docker context ls`, you’ll see a list of available contexts where the currently-active one is indicated with a `*`—which will be `colima` after you’ve started it. You can change to the default (Docker Desktop) with `docker context use default` or change back with `docker context use colima`. This means you can run Docker Desktop and Colima at the same time, but be mindful of which context you’re pointing at!
 
     !!!warning "Colima can only work in your home directory unless you do further configuration"
```
rpkoller, line 46:
```suggestion
    !!!warning "Without additional configuration, Colima can only provide Docker to DDEV projects located in your home directory"
```
not the perfect headline yet. i just wanted to communicate the gist i try convey. but in the previous version it could be misread that colima only works in the homedirectory but technically they details the colima documentation is about deals with adding additional (writeable) mounts outside of the home directory which is the default

docs/content/users/install/docker-installation.md
```diff
@@ -29,22 +29,22 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
...
-        By default, Colima only mounts your home directory, so it’s easiest to use it in a subdirectory there. See the `~/.colima/default/colima.yaml` for more information, or notes in [colima.yaml](https://github.com/abiosoft/colima/blob/fc948f8f055600986f87e29e3e632daf56ac8774/embedded/defaults/colima.yaml#L130-L143).
+        By default, Colima only mounts your home directory, so it’s easiest to use it in a subdirectory there. See the `~/.colima/default/colima.yaml` for more information, or notes in [colima.yaml](https://github.com/abiosoft/colima/blob/main/embedded/defaults/colima.yaml#L160-L173).
```
rpkoller, line 47: that implies "so it’s easiest to use it in a subdirectory" that colima is used in a subdirectory. but instead it is about where the ddev based projects are located.

rpkoller (commented):
left a few comments. but one general thought. at the moment the order of available docker providers is not in a alphabetical order therefore me as a user would read some sort of ranking into. orbstack is the best and rancher desktop the "worst" ( i exaggerate ). and there are already four options. me as a user would be frozen uncertain which solution to pick. on option might be separate the commercial/paied solutions (orbstack and in part docker) from the open source ones (colima and rancher). those thoughts are definitely out of the scope of this issue but i just wanted to raise them. but i would definitely think it would be good to provide some more guidance or at least some sort of decision tree based help for someone uncertain.

docs/content/users/install/docker-installation.md
rfay: Improved this.

PR #5952, opened 2024-03-11
docs: Improve suggestions about WSL2 docker options

rfay:
## The Issue

The WSL2 install docs don't make clear that docker-ce is considered best.

Improve the wording.

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -141,7 +141,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     2. **Commercial, often problematic:** [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
     3. **Old school:** [Installing directly on traditional Windows](#traditional-windows) with an installer
 
-    **We strongly recommend using WSL2.** While its Linux experience may be new for some Windows users, it’s worth the performance benefit and common experience of working with Ubuntu and Bash.
+    **We strongly recommend using WSL2 for your Windows DDEV development environment.** While its Linux experience may be new for some Windows users, it’s worth the performance benefit and common experience of working with Ubuntu and Bash.
 
     ### Important Considerations for WSL2 and DDEV
 
@@ -204,7 +204,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ### WSL2 + Docker Desktop Install Script
 
-    WSL2 with Docker Desktop is a less-favored choice because Docker Desktop is often buggy and lightly supported, and has many features not required for use with DDEV. It is also not free software (although smaller organizations can use it free of charge) and it is not open-source.
+    WSL2 with Docker Desktop is a less-favored choice because Docker Desktop may be lightly supported, and has many features not required for use with DDEV that do not add particular value. It is also not free software (although smaller organizations can use it free of charge) and it is not open-source.
     
     The script here prepares your default WSL2 Ubuntu distro for use with Docker Desktop, and you can run the script multiple times without breaking anything.
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -137,9 +137,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     You can install DDEV on Windows three ways:
 
-    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
-    2. [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
-    3. [Installing directly on traditional Windows](#traditional-windows) with an installer
+    1. **Most reliable, free, open-source:** [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
+    2. **Commercial, often problematic:** [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
+    3. **Old school:** [Installing directly on traditional Windows](#traditional-windows) with an installer
 
     **We strongly recommend using WSL2.** While its Linux experience may be new for some Windows users, it’s worth the performance benefit and common experience of working with Ubuntu and Bash.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/install/ddev-installation.md
```diff
@@ -137,9 +137,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
...
     You can install DDEV on Windows three ways:
 
-    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
-    2. [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
-    3. [Installing directly on traditional Windows](#traditional-windows) with an installer
+    1. **Most reliable, free, open-source:** [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
+    2. **Commercial, often problematic:** [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
+    3. **Old school:** [Installing directly on traditional Windows](#traditional-windows) with an installer
 
     **We strongly recommend using WSL2.** While its Linux experience may be new for some Windows users, it’s worth the performance benefit and common experience of working with Ubuntu and Bash.
```
kiatng, line 144: **We strongly recommend using WSL2.**

I now get that it means as compared to WSL1. How about

```suggestion
    **We strongly recommend updating your WSL to WSL2 for the best performance.**

**Comment by kiatng** [comment]
I'm not sure if I represent a typical developer journey, so I read about WSL 2 and invested in a new machine and installed Windows 11. Because WSL2 is very strongly tied to Docker Desktop, that's what I get when I google "wsl docker". So, I proceeded to install both. Then days/months later, I learned about DDEV. In this scenario, what do I do? I would proceed with Docker Desktop.

File: docs/content/users/install/ddev-installation.md (hunk 1 above)
Comment (rfay) [reply docs/content/users/install/ddev-installation.md:144]: Thanks for reviewing! No, actually WSL1 is not usable with DDEV or Docker, and none of the instructions will work with WSL1, and no docker platform will work there. 

```suggestion
    **We strongly recommend using WSL2 for your Windows DDEV development environment.** While its Linux experience may be new for some Windows users, it’s worth the performance benefit and common experience of working with Ubuntu and Bash.
```

docs/content/users/install/ddev-installation.md
```diff
@@ -202,7 +204,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ### WSL2 + Docker Desktop Install Script
 
-    This prepares your default WSL2 Ubuntu distro for use with Docker Desktop, and you can run the script multiple times without breaking anything.
+    WSL2 with Docker Desktop is a less-favored choice because Docker Desktop is often buggy and lightly supported, and has many features not required for use with DDEV. It is also not free software (although smaller organizations can use it free of charge) and it is not open-source.
```
rfay, line 207:
```suggestion
    WSL2 with Docker Desktop is a less-favored choice because Docker Desktop may be lightly supported, and has many features not required for use with DDEV that do not add particular value. It is also not free software (although smaller organizations can use it free of charge) and it is not open-source.
```

rfay:
> Because WSL2 is very strongly tied to Docker Desktop, that's what I get when I google "wsl docker". So, I proceeded to install both. Then days/months later, I learned about DDEV. In this scenario, what do I do? I would proceed with Docker Desktop

Thanks for your feedback on this, and for reviewing. Glad you got things working. 

This discussion originally came from https://discord.com/channels/664580571770388500/1216277811585617940/1216582232857247744

docs/content/users/install/ddev-installation.md
```diff
@@ -137,11 +137,11 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     You can install DDEV on Windows three ways:
 
-    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
-    2. [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
-    3. [Installing directly on traditional Windows](#traditional-windows) with an installer
+    1. **Most reliable, free, open-source:** [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script)
+    2. **Commercial, often problematic:** [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script)
+    3. **Old school:** [Installing directly on traditional Windows](#traditional-windows) with an installer
```
kiatng, line 142:
```suggestion
    3. **Old school:** [Installing without WSL2 on traditional Windows](#traditional-windows) with an installer
 ```

Comment (tyler36) [inline docs/content/users/install/ddev-installation.md:142]:
```suggestion
    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script) **Recommended, best performance**
    2. [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script) **May require licence **
    3. [Installing directly on traditional Windows](#traditional-windows) with an installer **Legacy, slower performance**
```

kiatng:
https://github.com/ddev/ddev/blob/a4bf20b2246790a607b3a0b2fa12e657af5ab796/docs/content/users/install/ddev-installation.md?plain=1#L221-L232

Items 5 and 10 have the same instruction:

> In *Windows Update Settings* → *Advanced Options* enable *Receive updates for other Microsoft products*. You may want to occasionally run `wsl.exe --update` as well.

docs/content/users/install/ddev-installation.md
rfay: I wonder how to say this better.  This goes back to the same confusion about WSL2. You can't install Docker Desktop (required for old-school Windows setup) without WSL2, because that's what it uses under the hood. But with this solution, **DDEV** is used from git-bash or PowerShell, and it uses Docker Desktop.

rfay:
```suggestion
    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script) **Recommended, best performance, most reliable**
    2. [Using WSL2 with Docker Desktop](#wsl2-docker-desktop-install-script) **May require license, less reliable**
    3. [Installing directly on traditional Windows](#traditional-windows) with an installer **Legacy, slower performance**```

rfay:
Thanks so much @kiatng - removed step 10.

penyaskito:
Reviewed this after merge as a WSL2+Docker Desktop user, LGTM (still never found the time to migrate to docker-ce).

rfay:
Thanks @penyaskito !

BTW it doesn't take much effort to migrate:

1. `ddev snapshot -a`
2. Disable all Docker Desktop integrations with WSL2 (including the default)
3. Exit Docker Desktop
4. Install docker-ce or run the docker-ce install script, should work OK.
5. Restore snapshots as needed.

PR #5956, opened 2024-03-11
fix: ddev debug test should check to make sure project is in homedir, for #3679

rfay:
## The Issue

* #3679

* `ddev debug test` should say if project is not in home directory
* Gather version numbers of colima/lima/orbstack/rancher
* Remove some mutagen output


stasadev (approved):
Looks good to me :+1:

PR #5960, opened 2024-03-12
build: use traefik 3.0

rfay:
## The Issue

Traefik 3.0 is near release, currently in rc2.

See if we can use it.

## Release/Deployment Notes

It's possible that we need to check the docs for customization, and that something might be impacted in customization.


rfay:
It looks like this has trouble with our config when handling multiple hostnames.

For example, TestConfigValidate fails when trying to do extra hostnames, and TestDdevStartMultipleHostnames fails.

I don't think I have the patience to take a deep dive into Traefik 3.0 at this point. Closing for now.

PR #5893, opened 2024-02-23
fix: make a copy of the bytes array before passing to the channel, fixes #5470

deviantintegral:
## The Issue

* #5470

Docker Compose files can become corrupt. This is due to a race condition in the code that calls docker compose commands. I think this only comes up with ddev start because it's the only code that parses docker compose output - namely `docker compose config`.

## How This PR Solves The Issue

https://github.com/ddev/ddev/pull/5340 added support for showing output while the containers are built. https://github.com/hanoii/ddev/commit/35151743309dfadc29ff9e5add9141a2ac403cdb switched from calling `Text()` to `Bytes()` to assist supporting real time output in debug commands. However, [Bytes() has subtly different behaviour](https://pkg.go.dev/bufio#Scanner.Bytes):

> The underlying array may point to data that will be overwritten by a subsequent call to Scan. It does no allocation.

Unfortunately, this issue can take over 1000 instances to actually occur. That's why it hasn't come up yet on our Linux users locals, and also why retries in CI generally work. It's also why we originally thought it was a 1.22.4 issue, and it was! It was just hard to tell with such rare reoccurrence.

This PR fixes the issue by allocating and copying a new value to return through the channel. I did a search for any other calls to `Bytes()` and didn't find any.

## Manual Testing Instructions

~~So far, I'm unable to reproduce this on colima / arm64~~. One time it only took ~20 loops, the last time it took nearly 1200. We can see an error we've seen before, and that in the case the compose file is completely empty.

Edit: I just reproduced the below on my mac. It took 614 iterations:

```bash
#!/bin/bash

echo "Generating first cases..."
ddev debug compose-config > /tmp/1.ddev.compose.yaml
md5sum /tmp/1.ddev.compose.yaml
ddev debug compose-config > /tmp/2.ddev.compose.yaml
md5sum /tmp/2.ddev.compose.yaml

echo "Looping..."
I=2
while diff -up /tmp/1.ddev.compose.yaml /tmp/2.ddev.compose.yaml; do
  ddev debug compose-config > /tmp/2.ddev.compose.yaml
  echo Try $I....
  ((I++))
  md5sum /tmp/2.ddev.compose.yaml
done
```

image

On macOS:

```console
Try 614....
6ba83c77fda3524b8fdf2898c0c32cdd  /tmp/2.ddev.compose.yaml
--- /tmp/1.ddev.compose.yaml	2024-02-23 16:35:27
+++ /tmp/2.ddev.compose.yaml	2024-02-23 16:39:16
[...]

PR #5964, opened 2024-03-12
fix: Rewrite ComposeCmd() with just dots goroutine, fixes #5913

### LINKED ISSUE #5913: Followup TODO: Rework ComposeCmd() to remove goroutines (mostly)
Followup from 
* https://github.com/ddev/ddev/pull/5893

ComposeCmd uses goroutine when it really doesn't need to for the vast majority of situations.

Either make it simpler again or make it only use goroutines for the one case where they're needed

This PR could also add a race detection test, or that could be done in another PR.

rfay:
## The Issue

* #5913 
* #5893
* #5470 

* Only use goroutine when we need it, which is just for outputting dots during build.

- [x] Try out `ddev start` and `ddev debug refresh` and `DDEV_DEBUG=true ddev start`
- [x] Verify number of goroutines left over at exit

## Release notes

`ddev debug refresh` builds only the web image, making the output less confusing.

### DIFF
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -42,13 +42,6 @@ var DebugRefreshCmd = &cobra.Command{
 			util.Failed("Failed to get project: %v", err)
 		}
 
-		status, _ := app.SiteStatus()
-		if status != ddevapp.SiteRunning {
-			if err = app.Start(); err != nil {
-				util.Failed("Failed to start %s: %v", app.Name, err)
-			}
-		}
-
 		app.DockerEnv()
 		if err = app.WriteDockerComposeYAML(); err != nil {
 			util.Failed("Failed to get compose-config: %v", err)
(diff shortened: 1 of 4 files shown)


stasadev (approved):
This is great!

Tested and verified goroutines count (3 without amplitude and 12 with amplitude on `ddev debug refresh`, and 13 with amplitude on `ddev start`)

rfay:
> My only bit of architecture feedback; I think goroutines are a killer feature of the language. If anything, I'd suggest ddev lean more into goroutines as a way to improve performance. It's something I miss in PHP and nodejs (where with node, you only have async IO, but not real CPU threads). But in this specific case, I think this cleanup is good!

Thanks!

Where would you see performance being improved (or especially needing improvement?) Mostly the biggest work is done by `ddev start` and the pieces there are largely sequential.

deviantintegral:
I don't have any clear examples to point to - my comment was more from the perspective that it is more challenging to turn a big single-threaded codebase into a multithreaded one as compared to thinking about parallel processing from the start. And for ddev, ideally much of that work (say, starting containers) should already be run in parallel by docker compose.

PR #5965, opened 2024-03-13
docs: Use mysql:8.0 instead of mariadb for ExpressionEngine quickstart

ssssct:
## The Issue

Even if mariadb works, there are times that specific joins can be much slower than in mysql. 

ExpressionEngine documentation (and team) recommends using mysql instead of the default mariadb.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -178,7 +178,7 @@ ddev launch
     ```bash
     mkdir my-ee && cd my-ee
     unzip /path/to/ee-zipfile.zip
-    ddev config --auto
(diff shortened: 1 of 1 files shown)


rfay:
Yay, first PR!!! 💯 🥇

rfay (approved):
Congrats on first PR! When there are some docs on the "why" for mysql:8.0 (since nothing like that is currently in EE docs) please update the OP with a link, thanks.

rfay:
You can still update the PR here at your convenience, maybe when your PR upstream goes in. Thanks for updating!

ssssct:
Great! Will do! Thanks Randy!

Issue #3214, opened 2021-09-03
Use compose-go to build and parse docker-compose files, remove `version 3.6`


rfay:
the compose-spec has moved along significantly in recent years, and there's even a go library to build and parse compose files. https://github.com/compose-spec/compose-go

We can use this instead of running `docker-compose config` and can probably build the whole file using this technique instead of templating.

There may be issues with docker-compose v1 compatibility, so this might have to be deferred a bit. 

Related: the `version` key is no longer in the compose spec, so our `version: 3.6` is obsolete, but some versions of docker-compose that we support still need that. But the time will come that we can remove it.

rfay:
DDEV's processing of the `docker-compose.*.yaml` files can remove this from existing add-on recipes.

stasadev:
We already use `compose-go` with:

- #7422

`docker-compose config` is still used to gather all `docker-compose.*.yaml` files before `fixupComposeYaml` is called.

I think that's fine, closing this issue.

PR #5967, opened 2024-03-13
refactor: use github.com/compose-spec/compose-go/v2/cli, for #3214

stasadev:
## The Issue

- #3214

This PR doesn't change the existing logic for templating. It simply tries to replace the `docker-compose config` with a library call.

TODO:
- [ ] I don't know yet how to separate `stdout` and `stderr` in `MarshalYAML()`.
- [ ] `context.Background()` is not reused.


rfay:
Size change for ddev binary (arm64 macOS) is modest, from 20,505,554 to 22,048,082

rfay:
Next step (or maybe done here) is to use their structs to create the config in the first place.

stasadev:
I'm going to close this PR for now.

It's a good place to start, but rewriting all the `docker-compose config` logic is not easy and may take longer than expected.

PR #5975, opened 2024-03-15
test: Check goroutine count on several tests at completion, fixes #5920

### LINKED ISSUE #5920: Clean up goroutines, Add an automated test for leftover goroutines
### Expected Behavior

In 
* https://github.com/ddev/ddev/pull/5878

We discovered that Colima was broken due to many docker client connections being left open (via many running goroutines at shutdown)

Since we worked on that, we know that we can test for open goroutines at shutdown, so we should add an automated test that detects a situation like this. It can be simple, just detecting a number greater than we would expect.

Some experimentation with pprof on this line is in https://github.com/rfay/ddev/tree/20240221_pprof_experiments, but it's just experimentation. However, it would be really easy to move it forward.

rfay:
## The Issue

* #5920 

We want to know if goroutines get out of control as they did in https://github.com/ddev/ddev/pull/5878

Test on several of the TestCmd tests to make sure we have a reasonable number of goroutines at stop.

## Manual testing

- [ ] `DDEV_GOROUTINES=true ddev start`
- [ ] `DDEV_GOROUTINES=true DDEV_VERBOSE=true DDEV_DEBUG=true ddev start`

You can also use any other command, not just `ddev start`

### DIFF
--- a/cmd/ddev/cmd/logs_test.go
+++ b/cmd/ddev/cmd/logs_test.go
@@ -30,9 +30,9 @@ func TestLogsNoConfig(t *testing.T) {
 
 // TestCmdLogs tests that the ddev logs functionality is working.
 func TestCmdLogs(t *testing.T) {
-	//if nodeps.IsAppleSilicon() {
-	//	t.Skip("Skipping on mac M1 to ignore problems with 'connection reset by peer'")
-	//}
+	// Gather reporting about goroutines at exit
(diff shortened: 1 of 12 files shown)


stasadev (approved):
Looks good to me!

I added `DDEV_DEBUG=true` to the manual testing instructions, because without it there were no details.

PR #5976, opened 2024-03-15
fix: add restart for database commands after ddev start, fixes #5949

### LINKED ISSUE #5949: Global database commands cannot open a database when the project is stopped

```
```
</details>         

### Expected Behavior

It should open the database (or ask the user to start the project).

### Actual Behavior

When I tested `ddev dbeaver` with a stopped project, it started the project and tried to open the database with `3306` port (which is not the correct host port), and on the second run the db was opened with the actual working port.

### Steps To Reproduce

1. `ddev stop`
2. `ddev dbeaver` (or any other db command)
3. see the database connection with the wrong port

### Anything else?

I assume that this is because the DDEV env variables in the `dbeaver` script (and any other script for opening a database) are not updated after `ddev start`, so it only works when the project is started.

Is it possible to update env in Bash after `ddev start`? Or maybe we need to change the logic for db commands from:

```bash
  ddev start
fi
```
to
```bash
  echo >&2 "Project ${DDEV_PROJECT} is not running, please start it"
  exit 1
fi
```

stasadev:
## The Issue

- #5949

The simplest way I found, is to restart the database command again after the project is running and all nessesary env variables are available.

I also added the same thing to `ddev launch`.

TODO: add the same thing to official add-ons:

https://github.com/ddev/ddev-phpmyadmin/pull/10
https://github.com/ddev/ddev-adminer/pull/20

```
ddev stop

ddev debug fix-commands

# or any other db command
ddev dbeaver

# the database should open normally
```

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
@@ -10,6 +10,9 @@
 if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
   echo "Project ${DDEV_PROJECT} is not running, starting it"
   ddev start
+  # run this script again with the updated environment
+  DDEV_PROJECT_STATUS=running ddev "$(basename "$0")" "$@"
+  exit $?
 fi
 
 database="${1:-db}"
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
@@ -7,11 +7,15 @@
 ## OSTypes: darwin,linux,wsl2
 ## HostBinaryExists: /mnt/c/Program Files/dbeaver/dbeaver.exe,/Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
 
-if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
+if [ "${DDEV_PROJECT_STATUS}" != "running" ] && [ -z "$no_recursion" ]; then
   echo "Project ${DDEV_PROJECT} is not running, starting it"
   ddev start
-  # run this script again with the updated environment
-  DDEV_PROJECT_STATUS=running ddev "$(basename "$0")" "$@"
+  start_exit_code=$?
+  if [ $start_exit_code -ne 0 ]; then
+    exit $start_exit_code
+  fi
+  # run this script again, as the environment is updated after "ddev start"
+  no_recursion=true ddev "$(basename "$0")" "$@"
   exit $?
 fi
 
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -103,6 +103,7 @@ A number of environment variables are provided to these command scripts. These a
 * `DDEV_PHP_VERSION`: Current PHP version
 * `DDEV_PRIMARY_URL`: Primary project URL
 * `DDEV_PROJECT`: Project name, like `d8composer`
+* `DDEV_PROJECT_STATUS`: Project status determined from the `web` and `db` services health, like `starting`, `running`, `stopped`, `paused`, or another status returned from Docker, including `healthy`, `unhealthy`, `exited`, `restarting` 
 * `DDEV_PROJECT_TYPE`: `drupal8`, `typo3`, `backdrop`, `wordpress`, etc.
 * `DDEV_ROUTER_HTTP_PORT`: Router port for HTTP
 * `DDEV_ROUTER_HTTPS_PORT`: Router port for HTTPS
--- a/docs/content/users/extend/custom-commands.md
+++ b/docs/content/users/extend/custom-commands.md
@@ -103,7 +103,7 @@ A number of environment variables are provided to these command scripts. These a
 * `DDEV_PHP_VERSION`: Current PHP version
 * `DDEV_PRIMARY_URL`: Primary project URL
 * `DDEV_PROJECT`: Project name, like `d8composer`
-* `DDEV_PROJECT_STATUS`: Project status determined from the `web` and `db` services health, like `starting`, `running`, `stopped`, `paused`, or another status returned from Docker, including `healthy`, `unhealthy`, `exited`, `restarting` 
+* `DDEV_PROJECT_STATUS`: Project status determined from the `web` and `db` services health, like `starting`, `running`, `stopped`, `paused`, or another status returned from Docker, including `healthy`, `unhealthy`, `exited`, `restarting`
 * `DDEV_PROJECT_TYPE`: `drupal8`, `typo3`, `backdrop`, `wordpress`, etc.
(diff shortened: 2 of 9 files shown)


GuySartorelli:
If there is a problem with `ddev start` that results in containers not being started correctly, could this result in infinite recursion?

rfay:
> If there is a problem with ddev start that results in containers not being started correctly, could this result in infinite recursion?

A classic example would be something that happens to me all the time because I'm always changing Docker providers for testing. I end up with the wrong database type in the db volume, and on start it fails and complains. I think we could have restart after restart in that situation?

pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
```diff
@@ -10,6 +10,9 @@
 if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
   echo "Project ${DDEV_PROJECT} is not running, starting it"
   ddev start
+  # run this script again with the updated environment
+  DDEV_PROJECT_STATUS=running ddev "$(basename "$0")" "$@"
```
stasadev, line 14: > If there is a problem with ddev start that results in containers not being started correctly, could this result in infinite recursion?

`DDEV_PROJECT_STATUS=running` is here to prevent an infinite loop (that's why I added this override).

It's easy to test, the code below will cause an infinite run:

```
DDEV_PROJECT_STATUS=stopped ddev "$(basename "$0")" "$@"
```

GuySartorelli: Does DDEV respect environment variables that tell it the status of its projects? I'd expect it to set it each time from within ddev itself, based on the actual status of the project.
Is there a way to test that?

rfay (changes requested):
`DDEV_PROJECT_STATUS` isn't in https://ddev.readthedocs.io/en/latest/users/extend/custom-commands/#environment-variables-provided - we could add it here or elsewhere.

rfay:
There's an odd thing here: Custom/Shell commands already start DDEV if they're launched when it's stopped. So why do we need to do this anyway? If we do `ddev dbeaver` with the project stopped, we will always have a started project before we get the script being executed.

rfay (changes requested):
I think an easier fix here is to call (again) DockerEnv() before starting custom commands, and then the dbPortStr will be set correctly. See if that's an answer here.

stasadev:
> `DDEV_PROJECT_STATUS` isn't in ddev.readthedocs.io/en/latest/users/extend/custom-commands/#environment-variables-provided - we could add it here or elsewhere.

I agree, and will add this to the docs after we decide what to do with this PR.

> There's an odd thing here: Custom/Shell commands already start DDEV if they're launched when it's stopped. So why do we need to do this anyway?

Because the `ddev start` logic was moved to the shell script in:

- #5603

The project starts, but the shell script does not know about the new environment.

> I think an easier fix here is to call (again) DockerEnv() before starting custom commands, and then the dbPortStr will be set correctly. See if that's an answer here.

`ddev start` is called from a shell script, not from Go code, I don't think it's possible to update env in the current situation without calling a new DDEV instance (which is what I'm trying to do in this PR).

rfay:
> Because the ddev start logic was moved to the shell script

Thanks, I didn't remember that at all.

rfay (changes requested):
This is on the right track and works well most of the time. But the problem mentioned by @GuySartorelli is legit and is a problem. To demonstrate, `ddev stop` then  change the mariadb version, then `ddev dbeaver` (this output is with a `set -x` added)

```
rfay@rfay-tag1-m1:~/workspace/d10$ ddev dbeaver
+ '[' stopped '!=' running ']'
+ echo 'Project d10 is not running, starting it'
Project d10 is not running, starting it
+ ddev start
Starting d10...
Failed to start d10: unable to start project d10 because the configured database type does not match the current actual database. Please change your database type back to mariadb:10.11 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.11' and then you might want to try 'ddev debug migrate-database mariadb:10.6', see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/
++ basename /Users/rfay/workspace/d10/.ddev/.global_commands/host/dbeaver
+ DDEV_PROJECT_STATUS=running
+ ddev dbeaver
+ '[' stopped '!=' running ']'
+ echo 'Project d10 is not running, starting it'
Project d10 is not running, starting it
+ ddev start
Starting d10...
Failed to start d10: unable to start project d10 because the configured database type does not match the current actual database. Please change your database type back to mariadb:10.11 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.11' and then you might want to try 'ddev debug migrate-database mariadb:10.6', see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/
++ basename /Users/rfay/workspace/d10/.ddev/.global_commands/host/dbeaver
+ DDEV_PROJECT_STATUS=running
+ ddev dbeaver
+ '[' stopped '!=' running ']'
+ echo 'Project d10 is not running, starting it'
Project d10 is not running, starting it
+ ddev start
Starting d10...
Failed to start d10: unable to start project d10 because the configured database type does not match the current actual database. Please change your database type back to mariadb:10.11 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.11' and then you might want to try 'ddev debug migrate-database mariadb:10.6', see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/
++ basename /Users/rfay/workspace/d10/.ddev/.global_commands/host/dbeaver
```

rfay:
It's possible that it's not necessary to solve the recursion in this case, but worth thinking about.

stasadev:
Okay, thanks for the example, it's worth solving.

pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
```diff
@@ -7,9 +7,16 @@
 ## OSTypes: darwin,linux,wsl2
 ## HostBinaryExists: /mnt/c/Program Files/dbeaver/dbeaver.exe,/Applications/DBeaver.app,/usr/bin/dbeaver,/usr/bin/dbeaver-ce,/usr/bin/dbeaver-le,/usr/bin/dbeaver-ue,/usr/bin/dbeaver-ee,/var/lib/flatpak/exports/bin/io.dbeaver.DBeaverCommunity,/snap/bin/dbeaver-ce
 
-if [ "${DDEV_PROJECT_STATUS}" != "running" ]; then
+if [ "${DDEV_PROJECT_STATUS}" != "running" ] && [ -z "$no_recursion" ]; then
```
rfay, line 10: Simple and tricky technique!

rfay (approved):
Thanks for seeing this and chasing it!

pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver
GuySartorelli: Do we have a test that validates that env variables passed in like this to ddev are passed through to custom commands? If not, this seems like something that could regress easily if we change anything about how the DDEV env vars are passed in to custom commands.

rfay: They have to be, or all the environment variables we support would not work. 

For example, `DDEV_DEBUG=true ddev launch` will demonstrate this, and TestLaunchCommand depends on this.

https://github.com/ddev/ddev/blob/6945dd019d6f378721210dbc1b83f7c8a70c4cdd/cmd/ddev/cmd/commands_test.go#L301

PR #5978, opened 2024-03-15
refactor: standardize suggested commands for container log, fixes #5899

### LINKED ISSUE #5899: The error message for an unhealthy container does not contain information on how to view logs

```
```
</details>         

### Expected Behavior

`unhealthy` error should show how to run `ddev logs` in the same way as for `exited` error?

### Actual Behavior

Using Linux amd64, the container health is `unhealthy` and it becomes `exited` almost instantly, but because of early return I don't see `please use 'ddev logs -s web' to find out why it failed`.

It is not clear to the user what to do here:

```
ddev start
Waiting for web/db containers to become ready: [web] 
Failed waiting for web/db containers to become ready: web container failed: log=&{2024-02-26 10:14:17.395185894 +0200 EET 2024-02-26 10:14:17.511316278 +0200 EET 1 /var/www/html:OK mailpit:FAILED phpstatus:FAILED }, err=container /ddev-php-gmp-web unhealthy: &{2024-02-26 10:14:17.395185894 +0200 EET 2024-02-26 10:14:17.511316278 +0200 EET 1 /var/www/html:OK mailpit:FAILED phpstatus:FAILED }
```

---

At the same time using macOS arm64, the container health is `exited` immediately (it quickly passes the `unhealty` state).
I see `please use 'ddev logs -s web' to find out why it failed`:

```
ddev start
Waiting for web/db containers to become ready: [web]
Failed waiting for web/db containers to become ready: web container failed: log=, err=container exited, please use 'ddev logs -s web' to find out why it failed
```

---

This is a result of
- #5871 

Without #5871:
The last error message includes `please use 'ddev logs -s web' to find out why it failed` ( because `unhealthy` becomes `exited`):

```
ddev start
Waiting for web/db containers to become ready: [web] 
Failed waiting for web/db containers to become ready: web container failed: log=&{2024-02-26 10:21:44.883932923 +0200 EET 2024-02-26 10:21:44.990602754 +0200 EET 1 /var/www/html:OK mailpit:FAILED phpstatus:FAILED }, err=container /ddev-php-gmp-web unhealthy: &{2024-02-26 10:21:44.883932923 +0200 EET 2024-02-26 10:21:44.990602754 +0200 EET 1 /var/www/html:OK mailpit:FAILED phpstatus:FAILED } 
Something is wrong with Docker or Colima and /mnt/ddev_config is not mounted from the project .ddev folder. This can cause all kinds of problems. 
Waiting for additional project containers to become ready... 
Failed to start php-gmp: container(s) failed to become healthy before their configured timeout or in 120 seconds. This might be a problem with the healthcheck and not a functional problem. (container 'web' exited, please use 'ddev logs -s web' to find out why it failed)
```

### Steps To Reproduce

Test this on Linux:
- #5898
```
mkdir php-gmp && cd php-gmp
ddev config --auto --php-version 8.2 --webimage-extra-packages php8.2-gmp --omit-containers db
ddev start
```

### Anything else?

_No response_

stasadev:
## The Issue

- #5899

There are three functions in the code `ContainerWait`, `ContainersWait`, `ContainerWaitLog` that use different methods to provide the user with more detailed information about what to do with the failed containers.

This PR tries to provide the same output for all functions, giving the user as much information as possible about what to check.

Try breaking project containers on `ddev start` and see what happens.

I don't really know how to break them, so I've only really tested a timeout scenario using:

```
echo 'RUN chmod 000 /run/php' > .ddev/web-build/Dockerfile
```


rfay:
This is going to be great. I think that (either here or elsewhere) we should also limit the number of retries for apache, nginx, and php-fpm, as restarting 10 times is not useful. And I'm not sure we're limiting it to 10 tries. I think we should reconsider failing the container start if these three fail.

rfay (approved):
This looks great to me, thanks! I do think we need a bit of a followup to fine-tune the healthcheck and apache/nginx/php-fpm restart situation. A complete and immediate failure shouldn't require people to wait 2 minutes to figure it out.

stasadev:
> I think that (either here or elsewhere) we should also limit the number of retries for apache, nginx, and php-fpm, as restarting 10 times is not useful. And I'm not sure we're limiting it to 10 tries

I think it's better to do a new PR because it will require a new Docker image.
And this PR is a simple refactoring.

> I think we should reconsider failing the container start if these three fail.

Does this mean we need bring back `kill_supervisor.py`?

rfay:
> Does this mean we need to bring back kill_supervisor.py?

No, definitely not. I was thinking about a smarter healthcheck that would fail in those cases. Maybe we can brainstorm about that.

stasadev:
OK, I get it, maybe something like this inside the `healthcheck.sh`:

```
kill -SIGTERM $(ps aux | grep supervisord | grep -v grep | awk '{print $2}')
```

And we need to define the conditions for terminating the process (ideally, if the number of restarts exceeds a certain limit), but I don't know if supervisord provides such information.

rfay:
I was just thinking we should make the healthcheck show failed much earlier than the normal timeout if one of the key processes isn't working.

PR #5979, opened 2024-03-16
docs: Improve WSL2 instructions to explain how to open WSL2 location from vscode on Windows

kiatng:
## The Issue
I added an important step on how to launch VS Code in WSL2. I didn't know about this, I launched it in Windows and couldn't figure out why certain things weren't working correctly. I am using PHP, Intelephense was wreck and spent hours trying different things.

### DIFF
--- a/docs/content/users/debugging-profiling/step-debugging.md
+++ b/docs/content/users/debugging-profiling/step-debugging.md
@@ -88,7 +88,7 @@ If you encounter the error: "Can't find a source position. Server with name 'SIT
 6. In a browser, visit your project and confirm you hit the breakpoint.
 
 !!!tip "If you’re using VS Code on Windows with WSL2"
-    VS Code should suggest two extensions if you have WSL2 enabled along with a PHP project: “[PHP Debug](https://marketplace.visualstudio.com/items?itemName=xdebug.php-debug)” and “[WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl)”. You’ll need to enable both of these extensions in your distro (e.g. Ubuntu).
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
There may be a bit more finesse involved here, since these instructions are for all platforms.

In my opinion `code .` in WSL2 is the normal way to launch vscode. But you can also open \\wsl$\Ubuntu\home\... and it should work fine.

rfay:
I tested here and this is actually not correct. I opened a little-used vscode and updated it.  When I opened a project in \\WSL$\Ubuntu\home\rfay\workspace\d10 it prompted me with something like "Are you using WSL? You need to install the WSL support". I did that.  This is explained in https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl

I certainly agree that WSL2 adds complexity (and value) to everything, and also agree that using `code .` is the fastest way to get going because it makes these things happen right away. Maybe you can add a commit to tone this down and make it more accurate? Happy to continue the conversation here or in Discord.

image

rfay:
No response from OP, closing. I'm certainly open to adding more about this situation, but I don't think it belongs where this landed, and this one is not correct.

kiatng:
Sorry for not getting back. I revisited ddev again to give it another try. I believe this is a valid PR, I tested it again.

> There may be a bit more finesse involved here, since these instructions are for all platforms.

This PR is specifically for WSL2, the instruction is after line 90 `!!!tip "If you’re using VS Code on Windows with WSL2"`

> In my opinion code . in WSL2 is the normal way to launch vscode. But you can also open \wsl$\Ubuntu\home... and it should work fine.

The above should be documented somewhere but I couldn't find it. That is what this PR is about.

> I tested here and this is actually not correct.

To clarify, for novices not familiar with WSL2 like myself, I always launch VSCode from windows (the incorrect method ), it would launch OK, but Intelephense would be broken. 

image

This PR can potentially saves hours for novices trying to figure out why certain things don't work.

rfay:
I tested with the open-from-vscode method (with the wsl2 extension loaded) and it was fine. 

Please, when people take the time to look at your issue/PR, you can at a minimum respond in a reasonable timeframe.

You're welcome to take another look, and fine-tune your recommendation with a new PR.

PR #5981, opened 2024-03-16
docs: add tip about using 0.0.0.0 not localhost

rfay:
## The Issue

People trying to run node.js daemons inside web container keep stumbling on the need to expose to all interfaces (0.0.0.0)

Add tip and suggestion about this.


docs/content/users/extend/customization-extendibility.md
```diff
@@ -49,7 +49,10 @@ There are several ways to run processes inside the `web` container.
 2. Run them with a `post-start` [hook](../configuration/hooks.md).
 3. Run them automatically using `web_extra_daemons`.
 
-### Running Extra Daemons with `post-start` Hook
+!!!tip "Daemons requiring network access should bind to `0.0.0.0`, not to `localhost` or `127.0.0.1`"
+    Many examples on the internet show starting daemons starting up and binding to `127.0.0.1` or `localhost`. Those examples are assuming that network consumers are on the same network interface, but with a DDEV-based solution the network server is essentially on a different computer from the host computer (workstation). If the host computer needs to have connectivity, then bind to `0.0.0.0` (meaning "all network interfaces") rather than `127.0.0.1` or `localhost` (which means only allow access from the local network). A [ReactPHP example](https://github.com/orgs/ddev/discussions/5973) would be `$socket = new React\Socket\SocketServer('0.0.0.0:3000');` instead of `$socket = new React\Socket\SocketServer('127.0.0.1:8080');`. (To expose your daemon to the workstation and browser, see [Exposing Extra Ports via `ddev-router`](#exposing-extra-ports-via-ddev-router).)
```
GuySartorelli, line 53: Is the change from port `8080` to `3000` here relevant or necessary?
If not I'd recommend just keeping the ports the same to avoid confusion.
If it is necessary to also change the port, that probably warrants some additional explanation.

rfay: Just a big mistake. OH, it came from the issue, https://github.com/orgs/ddev/discussions/5973 - thanks for catching it.

```suggestion
    Many examples on the internet show starting daemons starting up and binding to `127.0.0.1` or `localhost`. Those examples are assuming that network consumers are on the same network interface, but with a DDEV-based solution the network server is essentially on a different computer from the host computer (workstation). If the host computer needs to have connectivity, then bind to `0.0.0.0` (meaning "all network interfaces") rather than `127.0.0.1` or `localhost` (which means only allow access from the local network). A [ReactPHP example](https://github.com/orgs/ddev/discussions/5973) would be `$socket = new React\Socket\SocketServer('0.0.0.0:3000');` instead of `$socket = new React\Socket\SocketServer('127.0.0.1:3030');`. (To expose your daemon to the workstation and browser, see [Exposing Extra Ports via `ddev-router`](#exposing-extra-ports-via-ddev-router).)
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -49,7 +49,10 @@ There are several ways to run processes inside the `web` container.
 2. Run them with a `post-start` [hook](../configuration/hooks.md).
 3. Run them automatically using `web_extra_daemons`.
 
-### Running Extra Daemons with `post-start` Hook
+!!!tip "Daemons requiring network access should bind to `0.0.0.0`, not to `localhost` or `127.0.0.1`"
+    Many examples on the internet show starting daemons starting up and binding to `127.0.0.1` or `localhost`. Those examples are assuming that network consumers are on the same network interface, but with a DDEV-based solution the network server is essentially on a different computer from the host computer (workstation). If the host computer needs to have connectivity, then bind to `0.0.0.0` (meaning "all network interfaces") rather than `127.0.0.1` or `localhost` (which means only allow access from the local network). A [ReactPHP example](https://github.com/orgs/ddev/discussions/5973) would be `$socket = new React\Socket\SocketServer('0.0.0.0:3000');` instead of `$socket = new React\Socket\SocketServer('127.0.0.1:3000');`. (To expose your daemon to the workstation and browser, see [Exposing Extra Ports via `ddev-router`](#exposing-extra-ports-via-ddev-router).)
```
stasadev, line 53: I would like to reformat the example code into separate lines for better readability:

````suggestion
    Many examples on the internet show starting daemons starting up and binding to `127.0.0.1` or `localhost`. Those examples are assuming that network consumers are on the same network interface, but with a DDEV-based solution the network server is essentially on a different computer from the host computer (workstation). If the host computer needs to have connectivity, then bind to `0.0.0.0` (meaning "all network interfaces") rather than `127.0.0.1` or `localhost` (which means only allow access from the local network). A [ReactPHP example](https://github.com/orgs/ddev/discussions/5973) would be:
    ```php
    $socket = new React\Socket\SocketServer('0.0.0.0:3000');
    ```
    instead of:
    ```php
    $socket = new React\Socket\SocketServer('127.0.0.1:3000');
    ```
    To expose your daemon to the workstation and browser, see [Exposing Extra Ports via `ddev-router`](#exposing-extra-ports-via-ddev-router).
````

stasadev (changes requested):

stasadev (approved):

rfay:
Thanks for fixing that!

PR #5986, opened 2024-03-18
feat: accept full URLs in `ddev launch`, fixes #5685

### LINKED ISSUE #5685: Feature Request: Expand `ddev launch` to do full URL's

When using ddev launch it's only capable to do paths.

Now when i do commands like `drush uli` you cannot let the browser open the URL.
But inside WSL it uses `https://wslutiliti.es/wslu/` and that opens/uses the url. Then my one-time login has been opened by WSLU.
Then when wslu redirects its request to the host-browser the url is no longer valid.
I can imagen many more usecases where you want to open the browser from out the wsl environment, and ddev could be the winner for implementing it.

### Describe your solution

If we look into expanding `ddev launch [path]` into `ddev launch [url]` then we can use ddev to open the browser from scripts without editing the url in scripts.

### Describe alternatives

It currently works by editing the uli so it no longer contains the url but only the path
```ddev launch `drush uli | sed 's/.*\.\w*//'````` 

But that feels dirty :/
And doesn't work when you need services like solr

### Additional context

I think ddev uses still WSL open-browser / linux side of the host.
because ```ddev launch `drush uli | sed 's/.*\.\w*//'` ``` has the same issue that the URL is already visited
image

### PR DESCRIPTION
**Pull request by stasadev**
## The Issue

- #5685

Opens user-defined full URLs, so users are not limited to using `$DDEV_PRIMARY_URL` all the time.

```
DDEV_DEBUG=true ddev launch https://www.google.com
DDEV_DEBUG=true ddev launch https://www.google.com:1234
DDEV_DEBUG=true ddev launch http://example.com
DDEV_DEBUG=true ddev launch $DDEV_PRIMARY_URL/test
```

### DIFF
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -330,8 +330,10 @@ func TestLaunchCommand(t *testing.T) {
 	desc, err := app.Describe(false)
 	require.NoError(t, err)
 	cases := map[string]string{
-		"":   app.GetPrimaryURL(),
-		"-m": desc["mailpit_https_url"].(string),
+		"":                       app.GetPrimaryURL(),
+		"-m":                     desc["mailpit_https_url"].(string),
+		"/test":                  app.GetPrimaryURL() + "/test",
+		"https://www.google.com": "https://www.google.com",
 	}
 	if globalconfig.DdevGlobalConfig.MkcertCARoot == "" {
 		cases["-m"] = desc["mailpit_url"].(string)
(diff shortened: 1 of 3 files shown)


stasadev (commented):
> It doesn't have the full URL option option in the help, but needs it.

Thanks, I updated the help.

> It doesn't work with http urls, right?

It also works with `http`, I added this check to the tests previously.

rfay:
> It also works with http, I added this check to the tests previously

I guess I must be being tricked by my browser. I can certainly make it show the correct behavior with `DDEV_DEBUG=true ddev launch http://secondaryhostname.ddev.site/one/two/three`. But only sometimes does the browser stay with http. I haven't been able to figure out what the case is.

stasadev:
> But only sometimes does the browser stay with http. I haven't been able to figure out what the case is.

I also sometimes see this behavior when a website opens via https and then refuses to open via http (redirects to https), the only thing that helps is to open the URL in incognito mode or clear the site's cookies and restart the browser.

rfay:
Solution that seems to have worked for me: https://stackoverflow.com/a/77351518/215713

Open chrome://settings/content/insecureContent and add to "Allowed to show insecure content" `[*.]ddev.site`

rfay (approved):
Thanks, this is perfect and much appreciated!

rfay:
We need a followup on traditional Windows, TestLaunchCommand fails there, https://buildkite.com/ddev/ddev-windows-mutagen/builds/4496#018e5c9a-9fb5-4f64-af5d-8c3d262c8696

PR #5988, opened 2024-03-19
feat: enable corepack for yarn/pnpm version detection, fixes #5987

### LINKED ISSUE #5987: Enable Corepack in the web container

Drupal 11 has switched to Yarn 4, which requires [Corepack](https://nodejs.org/api/corepack.html) to be enabled. This will then automatically download the correct version of `yarn` based on the `packageManager` key in package.json; the fallback is Yarn 1, the same as before, so there should be no surprises to anyone who isn't already using Corepack.

### Describe your solution

Enable Corepack by default in the web container by running `corepack enable` at an appropriate point after installing NodeJS.

### Describe alternatives

_No response_

### Additional context

_No response_

longwave:
## The Issue

DDEV ships with Yarn 1 installed globally, but Drupal 11 uses Yarn 4.

[Corepack](https://nodejs.org/api/corepack.html) is enabled instead, which adds Yarn 1 by default but will download specific versions of Yarn if specified in package.json.

Ensure Yarn 1 is still available by default:

```
$ yarn --version
1.22.22
```

Start a project on Drupal 11 and ensure Yarn 4 works:
```
$ cd core
$ yarn --version
4.1.1
```

## Related Issue Link(s)

#5987

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -3,6 +3,7 @@ package ddevapp
 import (
 	"bytes"
 	"fmt"
+	"go/version"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -942,6 +943,13 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	if app.NodeJSVersion != nodeps.NodeJSDefault {
 		extraWebContent = extraWebContent + "\nRUN npm install -g n"
 		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN n install %s && ln -sf /usr/local/bin/node /usr/local/bin/nodejs", app.NodeJSVersion)
+		comp := version.Compare(app.NodeJSVersion, "16.0-alpha1")
+		util.Success("comp=%v", comp)
+		c, _ := semver.NewConstraint(">=16")
+		v, _ := semver.NewVersion(app.NodeJSVersion)
+		if !c.Check(v) {
+			extraWebContent = extraWebContent + fmt.Sprintf("\nRUN hash -r && corepack disable && npm install yarn")
+		}
 	}
 
 	// Some installed packages can change the permissions of /run/php
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -3,7 +3,6 @@ package ddevapp
 import (
 	"bytes"
 	"fmt"
-	"go/version"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -943,13 +942,13 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	if app.NodeJSVersion != nodeps.NodeJSDefault {
 		extraWebContent = extraWebContent + "\nRUN npm install -g n"
 		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN n install %s && ln -sf /usr/local/bin/node /usr/local/bin/nodejs", app.NodeJSVersion)
-		comp := version.Compare(app.NodeJSVersion, "16.0-alpha1")
-		util.Success("comp=%v", comp)
-		c, _ := semver.NewConstraint(">=16")
-		v, _ := semver.NewVersion(app.NodeJSVersion)
-		if !c.Check(v) {
-			extraWebContent = extraWebContent + fmt.Sprintf("\nRUN hash -r && corepack disable && npm install yarn")
-		}
+	}
+	// If node version 18+ enable corepack for yarn to be determined by project
+	// See https://yarnpkg.com/blog/release/4.0
+	c, _ := semver.NewConstraint(">=18")
+	v, _ := semver.NewVersion(app.NodeJSVersion)
+	if c.Check(v) {
+		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN corepack enable")
 	}
 
 	// Some installed packages can change the permissions of /run/php
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -948,7 +948,7 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	c, _ := semver.NewConstraint(">=18")
 	v, _ := semver.NewVersion(app.NodeJSVersion)
 	if c.Check(v) {
-		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN corepack enable")
+		extraWebContent = extraWebContent + "\nRUN corepack enable"
 	}
 
 	// Some installed packages can change the permissions of /run/php
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -80,7 +80,7 @@ RUN apt-get -qq install --no-install-recommends --no-install-suggests -y \
     postgresql-client \
     sqlite3
 
-RUN corepack enable && npm install --unsafe-perm=true --global gulp-cli
+RUN npm install --unsafe-perm=true --global gulp-cli yarn
 # Normal user needs to be able to write to php sessions
 RUN set -eu -o pipefail && LATEST=$(curl -L --fail --silent "https://api.github.com/repos/nvm-sh/nvm/releases/latest" | jq -r .tag_name) && curl --fail -sL https://raw.githubusercontent.com/nvm-sh/nvm/${LATEST}/install.sh -o /usr/local/bin/install_nvm.sh && chmod +x /usr/local/bin/install_nvm.sh
 
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -80,7 +80,7 @@ RUN apt-get -qq install --no-install-recommends --no-install-suggests -y \
     postgresql-client \
     sqlite3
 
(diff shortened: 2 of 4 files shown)


rfay:
Thanks!

How does this affect people who use older node.js versions? For example, if they were using `nodejs_version: 12` ?

longwave:
So this is fine for build time because the Dockerfile here always runs with NodeJS 20.

`nodejs_version` appears to just use `n` to install the correct version, and unfortunately this breaks on NodeJS 14 and below:

```
$ sudo n install 20 && nodejs --version && yarn --version
     copying : node/20.11.1
   installed : v20.11.1 (with npm 10.2.4)
v20.11.1
1.22.22

$ sudo n install 18 && nodejs --version && yarn --version
     copying : node/18.19.1
   installed : v18.19.1 (with npm 10.2.4)
v18.19.1
1.22.22

$ sudo n install 16 && nodejs --version && yarn --version
     copying : node/16.20.2
   installed : v16.20.2 (with npm 8.19.4)
v16.20.2
1.22.22

$ sudo n install 14 && nodejs --version && yarn --version
     copying : node/14.21.3
   installed : v14.21.3 (with npm 6.14.18)
v14.21.3
/usr/lib/node_modules/corepack/dist/lib/corepack.cjs:43431
    process.exitCode ??= code;
                     ^^^

SyntaxError: Unexpected token '??='
    at wrapSafe (internal/modules/cjs/loader.js:1029:16)
    at Module._compile (internal/modules/cjs/loader.js:1078:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
    at Module.load (internal/modules/cjs/loader.js:979:32)
    at Function.Module._load (internal/modules/cjs/loader.js:819:12)
    at Module.require (internal/modules/cjs/loader.js:1003:19)
    at require (internal/modules/cjs/helpers.js:107:18)
    at Object.<anonymous> (/usr/lib/node_modules/corepack/dist/yarn.js:2:1)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
```

https://github.com/tj/n/issues/789 seems related.

Perhaps we should purge corepack before https://github.com/ddev/ddev/blob/6d2e5013cd6d6056edc84178948788ad6d6135fb/pkg/ddevapp/config.go#L944 and then run `corepack enable` after if the version is 16 or greater?

rfay:
Testable artifacts are now in https://github.com/ddev/ddev/pull/5988#issuecomment-2007391779 - please give them a lot of use...

rfay:
It looks to me like corepack has a fatal flaw here:
* https://github.com/nodejs/corepack/issues/112

There's no way to properly disable it as it has destroyed the original links.

rfay:
Reverted the image-build approach and just did corepack on nodejs_version 18+, please see what you think. Updated artifacts will soon be in https://github.com/ddev/ddev/pull/5988#issuecomment-2007391779

rfay:
I imagine this needs a test, maybe bundle into the Drupal 11 tests, but maybe it needs its own.

longwave:
Works great on Drupal 11 with nodejs_version 20. The default version of Yarn is 1.22.22 except if there is a valid package.json when it downloads the correct version automatically:
```
$ ddev --version
ddev version v1.22.7-108-g93e2acfe9
$ ddev ssh
dave@drupal-web:/var/www/html/drupal$ nodejs --version
v20.11.0
dave@drupal-web:/var/www/html/drupal$ yarn --version
1.22.22
dave@drupal-web:/var/www/html/drupal$ cd core
dave@drupal-web:/var/www/html/drupal/core$ yarn --version
4.1.1
```

Also works as expected if I drop to Node 16 - this uses `yarn` installed globally via npm in the Dockerfile:
```
$ ddev ssh
dave@drupal-web:/var/www/html/drupal$ nodejs --version
v16.20.2
dave@drupal-web:/var/www/html/drupal$ yarn --version
1.22.21
```

pkg/ddevapp/config.go
```diff
@@ -943,6 +943,13 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 		extraWebContent = extraWebContent + "\nRUN npm install -g n"
 		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN n install %s && ln -sf /usr/local/bin/node /usr/local/bin/nodejs", app.NodeJSVersion)
 	}
+	// If node version 18+ enable corepack for yarn to be determined by project
+	// See https://yarnpkg.com/blog/release/4.0
+	c, _ := semver.NewConstraint(">=18")
+	v, _ := semver.NewVersion(app.NodeJSVersion)
+	if c.Check(v) {
+		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN corepack enable")
```
longwave, line 951: No need for sprintf:
```suggestion
		extraWebContent = extraWebContent + "\nRUN corepack enable"
```

longwave:
Regarding testing you could add a basic package.json with `packageManager` key in nodejs_test.go and check that the correct version of Yarn is installed for Node 18/20 (and that Yarn 1 is available in Node <18 I guess), but my Go is not that good for me to attempt writing that at the moment.

rfay:
@longwave my yarn is not so good. If you can just provide a little demo github repo with simplest possible package.json for each case it would be awesome.

hanoii:
I am not sure on this.. nothing too strong, but I was just going around this on a project and thought I chip in.

IF we were to include this, I would do it only for drupal project types so maybe not on the Dockerfile, so maybe as part of the start.sh or something else.

Still, if node made this opt-in, I would expect ddev to do the same. Obviously you can add this as a post start hook, even setting up an add-on and what not.

There's a lot of discussion on node for corepack et all:

https://github.com/nodejs/node/issues/50963

In particular I found https://github.com/nodejs/node/issues/50963#issuecomment-1862957925 interesting, is more a stance of npm and corepack that anything else, but I relate to some of the issues presented there.

rfay:
Interesting about https://github.com/nodejs/node/issues/50963#issuecomment-1862957925 - but npm is not enabled with corepack currently, just yarn and pnpm... Right?

hanoii:
correct

stasadev (approved):
Even if the corepack is *experimental*, it doesn't seem to be doing any harm as it has a fallback for the yarn v1.

rfay:
@hanoii I'd sure like. you to expand (even more) on your reservations about using corepack. 

Since the project type will be `drupal` now (https://github.com/ddev/ddev/pull/5991) we can't go just on the project type.

But we probably could add a build layer if here only on drupal11+, updating it when we detect the change.

rpkoller:
i've manually tested `ddev version v1.22.7-108-g93e2acfe9`(downloaded the artifact via the actions bot link). after that i tried to setup a new instance adding the following addon:  https://github.com/justafish/ddev-drupal-core-dev/

the steps i've took: 

git clone https://git.drupalcode.org/project/drupal.git . 
ddev config --project-type=drupal10 --omit-containers=db
ddev start
ddev get justafish/ddev-drupal-core-dev

executing the post install i ran into the following error: 

```
Executing post-install actions: 
yarn install v1.22.22
[1/5] Validating package.json...
error Drupal@: The engine "yarn" is incompatible with this module. Expected version ">= 4.1.1". Got "1.22.22"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Failed to execute command yarn --cwd=core: exit status 1
 
👎  
Could not process post-install action (6) '' 
```

hanoii:
> @hanoii I'd sure like. you to expand (even more) on your reservations about using corepack.

As I said above, I don't really have any other than the fact that it's not enabled by default on node. I am curious about that and why that is the reason, I couldn't really gather a strong reason, just wonder if it's the right call to enable it on ddev by default. I guess It's more philosophical than anything else. What you deice is OK with me.

rfay:
* corepack is clearly [marked as experimental](https://nodejs.org/api/corepack.html) and of "stabilty:1"
* However, [Yarn is "all in" on corepack](https://yarnpkg.com/getting-started/install)

But corepack is a bit of a mess, even from the beginning, as we discovered from the first version of this PR. If you `corepack enable` it changes the symlink at /usr/bin/yarn to point to /usr/lib/node_modules/corepack/dist/yarn.js

However, when you `corepack disable` it leaves no link at all in /usr/bin/yarn, so yarn is dead. No yarn 1, nothing. That kind of poor behavior seems like "very experimental".

Could you folks in the know say why Drupal 11 has decided to focus in on yarn, even though it seems to be less popular than npm for example? Or has it? Could you please give links to the Drupal 11 discussion of choosing/using yarn 4?

rfay:
It looks like this PR should be setting COREPACK_HOME to point to /mnt/ddev-global-config/corepack or something so that projects can share corepack stuff.

rfay:
@rpkoller came up with this important summary of what's simmering in the various communities: [Node.js Community Debate Intensifies Over Enabling Corepack by Default and Potentially Unbundling npm ](https://socket.dev/blog/node-community-debates-enabling-corepack-unbundling-npm)

rfay:
Why isn't it adequate to just do this for D11 core devs:

```yaml
hooks:
  post-start:
  - exec: corepack enable
```

Why would we want more than that?

PR #5989, opened 2024-03-19
build: show gitpod testing link in PR comment

rfay:
## The Issue

Gitpod long ago stopped showing its badge in a comment on a PR, but we should be able to do it ourselves. 

Try to add the comment.

### DIFF
--- a/.github/workflows/pr-artifacts-comment.yml
+++ b/.github/workflows/pr-artifacts-comment.yml
@@ -49,7 +49,9 @@ jobs:
               return acc;
             }, 'Download the artifacts for this pull request:\n');
 
-            body += `\n\n See [Testing a PR](https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr)`;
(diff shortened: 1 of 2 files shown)


rfay:
I tested this on ddev/ddev-test ( https://github.com/ddev-test/ddev/pull/1 )

and after much work got this:

image

stasadev (approved):
This is great!

tyler36:
> I tested this on ddev-test/ddev ( https://github.com/ddev-test/ddev/pull/1 )

I assume that repo only produces mac artifacts. DDEV PRs usually show a lot more. (Just checking we didn't disable the others)

image

rfay:
Thanks for paying attention @tyler36 ! There's no change to the artifacts created, if you see one we'll fix right away. I was limiting the loops to make things go faster.

tyler36:
Thank you for confirming @rfay . PR didn't look like it borked anything but was worth the double check. :)

rfay:
Here it is now, https://github.com/ddev/ddev/pull/6002#issuecomment-2011125538

image

PR #5991, opened 2024-03-20
feat: Add a `drupal` project type that handles Drupal8-11, fixes #4957

### LINKED ISSUE #4957: Add new project type `drupal` for drupal8+ (and support drupal11)
## Problem

You cannot create a new Drupal 11 project:

```
$ ddev config --docroot=web --project-type=drupal11
apptype must be one of backdrop, craftcms, drupal10, drupal6, drupal7, drupal8, drupal9, laravel, magento, magento2, php, shopware6, typo3, wordpress
```

## Solution

- Unify Drupal 8+ under a `drupal` project type, following the recommended PHP version for the current release (Drupal 10 and PHP 8.1)  

OR

- Add Drupal 11 as a project type option, requiring a [minimum PHP version of 8.2](https://www.drupal.org/project/drupal/issues/3330874).

### Describe alternatives

Use Drupal 10 config.

### Additional context

_No response_

rfay:
## The Issue

* #4957 

* Add `drupal` project type
* drupal8-10 become aliases to `drupal`
* Add tests for drupal11

- [ ] Use `ddev config` and autodetect
- [ ] Use `ddev config --project-type=drupal10 --docroot=web`
- [ ] Use ddev with previously created projects that have type of `drupal8/9/10`


docs/content/users/quickstart.md
```diff
@@ -130,14 +130,16 @@ ddev launch
 
 ## Drupal
 
-=== "Drupal 10"
+=== "Drupal 8+"
```
rpkoller, line 133: I am uncertain if i would call the Tab Drupal 8+. If you take a look at the latest state on the docs you only have Drupal 10 and Drupal 6/7 https://ddev.readthedocs.io/en/latest/users/quickstart/#__tabbed_3_2 . Drupal 8 and 9 are EOL, the apptype got the version removed and is called "just" drupal. not sure if it is necessary to stick to a version numbering here. i also wonder if it is necessary to keep "drupal 6" named. it is also EOL, it just happens to have the same instructions as drupal 7. but not sure if it is necessary to have versions that are EOL named in a quickstart guide. people who need the functionality running drupal 6 already have an install of drupal 6 and they know what to do. so what not simply call the tabs Drupal and Drupal 7 (legacy) something like that?

docs/content/users/quickstart.md
```diff
@@ -130,14 +130,16 @@ ddev launch
...
-    mkdir my-drupal10-site
-    cd my-drupal10-site
-    ddev config --project-type=drupal10 --docroot=web
+    mkdir my-drupal-site
+    cd my-drupal-site
+    ddev config --project-type=drupal --docroot=web
     ddev start
-    ddev composer create drupal/recommended-project
+    ddev composer create drupal/recommended-project:^10
+    # For Drupal9: ddev composer create drupal/recommended-project:^9
```
rpkoller, line 141: In line with my previous comment i would simply drop the how to install drupal 9 and only cover versions that are officially supported here? in the latest docs there was only drupal 10 in this tab. which got already extended by the dev version for drupal 11.x-dev

docs/content/users/configuration/config.md
```diff
@@ -518,10 +518,12 @@ The DDEV-specific project type.
 
 | Type | Default | Usage
 | -- | -- | --
-| :octicons-file-directory-16: project | `php` | Can be `backdrop`, `craftcms`, `django4`, `drupal6`, `drupal7`, `drupal8`, `drupal9`, `drupal10`,  `laravel`, `magento`, `magento2`, `php`, `python`, `shopware6`, `silverstripe`, `typo3`, or `wordpress`.
+| :octicons-file-directory-16: project | `php` | Can be `backdrop`, `craftcms`, `django4`, `drupal6`, `drupal7`, `drupal`,  `laravel`, `magento`, `magento2`, `php`, `python`, `shopware6`, `silverstripe`, `typo3`, or `wordpress`.
 
 The `php` and `python` types don’t attempt [CMS configuration](../../users/quickstart.md) or settings file management and can work with any project.
 
+The legacy `drupal8`, `drupal9`, and `drupal10` project types can also be specified, but if you just use `drupal` DDEV will autodetect the type and use the recommended settings.
```
rpkoller, line 525:
```suggestion
The legacy `drupal8`, `drupal9`, and `drupal10` project types can also be specified, but by using `drupal` DDEV will autodetect the type and use the recommended settings.
```

rpkoller (commented):
again thanks so much for the change of plans and adding this project type now instead of in 1.23.1.   I went through and tried to install drupal 8,9,10 and 11 with the drupal project type - everything worked like a charm (each version had php 8.2 mariadb10.11 and teh transaction level read committed). the only exception was drupal 8 which had several version issue on composer create. so i failed to install that. but i havent investigated any further. it is EOL anyway. but over all this looks pretty solid and good. thank you!

docs/content/users/configuration/config.md
tyler36: > The legacy `drupal8`, `drupal9`, and `drupal10`

There has been talk about using the terms "modern drupal" for Drupal 8+, and "legacy Drupal" for Drupal =<7. 
With that in mind, I think we should avoid using "legacy" in this doc context.

rfay: Good point @tyler36 - They're "legacy" only in the sense that we're not going to use specific terms in DDEV any more.  Please suggest alternative wording, thanks.

rpkoller: How about the following?  
`The project types drupal8, drupal9, and drupal10 can still be directly specified,...`

tyler36: > The project types drupal8, drupal9, and drupal10 can still be directly specified,...

@rpkoller, I like that! It keeps with the DDEV terminology.

stasadev (approved):
I tested it with an existing Drupal 10 and also tried the Drupal 9,10,11 quickstart, everything looks good :+1:

docs/content/users/quickstart.md
rfay: Changed tab name to "drupal"

docs/content/users/quickstart.md
rfay: drupal9 is still widely talked about and used, no reason not to leave it in there for a while. Let's remove it in a few months or a year. It also gives a good demonstration of how to use explicit composer recipe pattern.

docs/content/users/configuration/config.md
rfay: Updated to 
> The former DDEV project types `drupal8`, `drupal9`, and `drupal10` can still be manually specified; by using `drupal` instead, DDEV will autodetect the correct type and its corresponding settings.

rfay:
Moved to draft waiting for 
* https://github.com/ddev/ddev/pull/6012

to go in, then we'll use corepack enable for Drupal 11

stasadev:
More places that may or may not require an update for the `drupal` type:

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/pkg/ddevapp/templates.go#L10-L10

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/containers/ddev-webserver/tests/ddev-webserver/test.sh#L93-L93


https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/docs/content/users/usage/commands.md#L168-L168

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/docs/content/users/usage/commands.md#L524-L524

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/docs/content/users/usage/commands.md#L788-L788

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/cmd/ddev/cmd/config.go#L140-L140

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/cmd/ddev/cmd/list.go#L19-L20

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/docs/content/users/extend/custom-commands.md#L107

https://github.com/ddev/ddev/blob/96fd091844b9a9cf4b09196fe0963231ed1170a4/docs/content/users/extend/custom-commands.md#L127

rfay:
Added corepack support and fixed up the many wonderful things @stasadev had noticed in https://github.com/ddev/ddev/pull/5991#issuecomment-2017879124

docs/content/users/quickstart.md
```diff
@@ -159,14 +159,17 @@ ddev launch
 
 ## Drupal
 
-=== "Drupal 10"
+=== "Drupal"
```
rpkoller, line 162: not listed, but mentioned in the first sentence, drupal 8 is already failing on the `composer require` step due to the fact that php 8.2 is too new. (php 8.2 and mariadb 10.11 are set for drupal 8 with the drupal project type as well) 

and based on the results from this comment and the two in regards of drupal 9 and 11 the first sentence creates false expectations. if i read the first line i expect i choose `drupal` as the project type and ddev adjusts everything according to the project at hand (which is impossible since composer create is run at a later step)

rpkoller (commented):
hmmm due to the fact that it is impossible that ddev recognizes the version of drupal before `composer create` is run i am unsure if it is possible at all to have a single project type for drupal 8, 9 10, and 11. basically drupal 10, the current latest stable major version, is the only one for which the quickstart guide works out of the box without any stumbling block. 
across such a broad number of versions of drupal it is basically impossible to have a single project type based on the fact that the default and recommended way to install drupal is via composer. and if me, as the user still have to adjust the project type with up to three flags to work in any other version than the current stable major (drupal 10) it is tricky and potentially confusing. 
not sure about the conclusion. keep the drupal project type and add aside drupal 8 and 9 also a drupal 11 to directly address it? or go with creating a dedicated project  type for every deprecated one as suggested here https://github.com/ddev/ddev/issues/4957#issuecomment-1853760286 and use the drupal project type only for the current (10 and 11) but for that there is then still the newly arisen problem with drupal 11 requiring php 8.3  and corepack. tricky. i dont know. but just from a users and in particular new users perspective the current quickstart is challenging.

docs/content/users/quickstart.md
```diff
@@ -159,14 +159,17 @@ ddev launch
...
-    cd my-drupal10-site
-    ddev config --project-type=drupal10 --docroot=web
+    mkdir my-drupal-site
+    cd my-drupal-site
+    ddev config --project-type=drupal --docroot=web
     ddev start
-    ddev composer create drupal/recommended-project
+    ddev composer create drupal/recommended-project:^10
+    # For Drupal9: ddev composer create drupal/recommended-project:^9
+    # For Drupal 11: ddev composer create drupal/recommended-project:^11.x-dev
```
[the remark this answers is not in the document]
rfay: The quickstart needs to change to `ddev config --project-type=drupal --docroot=web --php-version=8.3`

docs/content/users/quickstart.md
```diff
@@ -159,14 +159,17 @@ ddev launch
...
-    mkdir my-drupal10-site
-    cd my-drupal10-site
-    ddev config --project-type=drupal10 --docroot=web
+    mkdir my-drupal-site
+    cd my-drupal-site
+    ddev config --project-type=drupal --docroot=web
     ddev start
-    ddev composer create drupal/recommended-project
+    ddev composer create drupal/recommended-project:^10
+    # For Drupal9: ddev composer create drupal/recommended-project:^9
```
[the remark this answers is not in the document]
rfay: I think people have to figure out old php versions for themselves?

Or maybe we need a whole discussion? Or maybe we need to say what to do with the various Drupal versions?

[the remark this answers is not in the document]
rpkoller: no i would vote against let people figure the necessary settings out themselves. that shouldnt be the goal of a quickstart guide. from my perspective you should be able without any research and not even with much thinking to simply copy and paste the instructions and be done running through any of the quicktarts without any error or stumbling block. 

the easiest short term fix would be adding a dedicated tab for each of the composer create examples. one for 11 one for 10 and one for 9 (and maybe even one for 8 - would be necessary with a single project type drupal). 

but then you could question the purpose and the level of convenience of unifying everything into a single project type if you have to override everything to adjust the php version and the corepack setting. the basic benefit a project type should have that i am able to for example start up drupal 10 without any extra setting, same for the other supported versions. but with the project type drupal you only have that level of convenience for drupal 10. 

as i said not sure. now retesting through all the projects each of the projects had its stumbling blocks with the drupal project type. not sure. what do do you think and what do others think?

docs/content/users/quickstart.md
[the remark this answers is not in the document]
rpkoller: also --corepack-enable=true is needed as well at least for drupal 11

rfay:
Thanks for the careful review, @rpkoller - Please take a look at https://ddev--5991.org.readthedocs.build/en/5991/users/quickstart/#drupal and see if it has some potential.

In general, what we've been doing with other CMSs is to reflect the current stable, and it seems to work out OK. 

So I added a bit about D9 and D11 here with more context.

docs/content/users/quickstart.md
```diff
@@ -159,19 +159,30 @@ ddev launch
...
+=== "Drupal"
+    For all versions of Drupal 8+ the Composer techniques work. The settings configuration is done differently for each Drupal version, but the project type is "drupal".
 
     ```bash
-    mkdir my-drupal10-site
-    cd my-drupal10-site
-    ddev config --project-type=drupal10 --docroot=web
+    mkdir my-drupal-site
+    cd my-drupal-site
+    ddev config --project-type=drupal --php-version=8.3 --docroot=web
```
rpkoller, line 168: I wouldn't make a drupal 10 instance use php 8.3. the recommended version of php for 10.x is still only 8.1.6 according to https://www.drupal.org/docs/getting-started/system-requirements/php-requirements (8.3. for example is only supported by drupal 10.2+ and i think a version should be picked that supports drupal 10.0). so i think for drupal 10 the drupal app type default php 8.2 might be the right pick  so `--php-version=8.3` might be dropped here.

php 8.3 and --corepack-enable for drupal 11 looks good. 

and going with php 8.1 for drupal 9 as well. gotta test all quickstarts tomorrow with drush another time to make sure.

and in general packing everything into a single tab and codeblock for setting up drupal 9 , 10 and 11 i consider hard to process. i think the better choice would be to go with dedicated tabs for each of the versions.

rfay: Since Drupal 10.2 supports PHP8.3, and since that's what we're getting with composer create, I think it's OK, https://www.drupal.org/docs/getting-started/system-requirements/php-requirements

TYPO3 has been getting along with this kind of ambiguity successfully for years and many major releases, but of course they had @gilbertsoft watching out for them. 

I think your concerns are valid and we'll see what people's experience is. Separate tabs is definitely a possibility. Also, I was thinking about a `ddev config --detect` which would detect the current version and update config explicitly. That could be added to quickstarts after the initial setup.

rpkoller: i am not a fan of ambiguity and making the user think. therefore are situations like that thin ice if the person using the functionality doesnt know all the details about. 

separate tabs would be a short term fix. but as i said not that elegant and it would be sort of countering the elegance and the fact that there is a `drupal`app type in place. but your idea of having a functionality of `ddev config --detect`to trigger the detection of the current version and updating the config explicitly would solve all of that in a way more elegant manner and solve the tricky situation with drupal. that way also only a single drupal tab would be necessary. and one idea would it make sense to trigger `ddev config --detect`automatically when running  `ddev composer create`? just as an additional idea. but in general big plus one for the config detect idea!

PR #5997, opened 2024-03-20
feat: remove segment.io support, fixes #5365

### LINKED ISSUE #5365: Remove support for Segment.io instrumentation

rfay:
## The Issue

* #5365 

We stopped actually using Segment.io in v1.22.0 because it's too expensive and we switched to using Amplitude

Now for v1.23, we can just remove that.

Manual testing is super-critical here. We need to make absolutely sure that we haven't broken Amplitude

- [ ] With amplitude enabled, take actions and see results in Amplitude


stasadev (changes requested):
I tested this manually and I can see my user events in Amplitude with this PR.

To remove: `SegmentKey` in `Makefile` and in `.github/workflows/master-build.yml`.

I noticed comments for `InitAmplitude`, should it be done here as well?:

https://github.com/ddev/ddev/blob/55053f055955fe5e146e56954ed34ba22fc2c447/cmd/ddev/main.go#L14-L16

rfay:
Thanks for the good eye. turned out that the check on `ddev config global` wasn't valid any more, we do it only on start (and now on ddev version)

On 
> This should be cleaned up

That was Simon's hope to remove init() usage in the future, which is definitely a valid hope.

stasadev (changes requested):
Noticed one more leftover for `SegmentKey` in `.github/workflows/master-build.yml`

rfay:
> Noticed one more leftover for SegmentKey

I guess I got distracted fixing that and adding the warning to `ddev version`

Turned out that hadn't been working for ages; we changed a long time ago to only check on `ddev start` but now added the check to `ddev version`.

stasadev (approved):

rfay:
Rebased this and retested, and it seems to be right. I checked `go mod tidy` and `go mod vendor` and they seem to be unchanged.

PR #5999, opened 2024-03-20
fix: make `ddev launch` test work on Windows, for #5986

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/5986#issuecomment-2010306973

Fixes a test, changes the path in the test cases to make it clear which one failed.


rfay:
Thanks for the super-fast fix!

PR #6000, opened 2024-03-20
build: bump docker library to v26.0.0

stasadev:
## The Issue

https://github.com/moby/moby/releases/tag/v26.0.0 is out.

Checks if tests work with v26.0.0.
Replaces some deprecated imports.


cmd/ddev/cmd/delete-images.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/ddev/ddev/pkg/versionconstants"
-	dockerTypes "github.com/docker/docker/api/types"
+	dockerImage "github.com/docker/docker/api/types/image"
```
rfay, line 14: Interesting!

rfay (approved):
Looks good to me, passes tests. Should we pull it? 

It will likely conflict with a couple of other vendor dir changes.

The time will probably come that you'll talk me into not committing the vendor dir.

stasadev:
Yes, let's pull it.

> The time will probably come that you'll talk me into not committing the vendor dir.

I see the benefits of commiting the vendor directory, and if we don't update it too often, it's not too annoying. But yes, there is a room for discussion.

PR #6002, opened 2024-03-21
fix: improve `ddev debug test`, fixes #3679

### LINKED ISSUE #3679: Improve `ddev debug test`

rfay:
## The Issue

Chores fixing up 
* #3679


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -30,66 +34,119 @@ function docker_desktop_version {
...
+  else
+    echo "MODIFIED .ddev/mutagen/mutagen.yml found"
+  fi
+else
+  echo ".ddev/mutagen/mutagen.yml not found"
+fi
+
 PROJECT_DIR=../${PROJECT_NAME}
-echo "======= Creating dummy project named  ${PROJECT_NAME} in ${PROJECT_DIR} ========="
+header "Creating dummy project named  ${PROJECT_NAME} in ${PROJECT_DIR}"
```
stasadev, line 62:
```suggestion
header "Creating dummy project named ${PROJECT_NAME} in ${PROJECT_DIR}"
```

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -30,66 +34,119 @@ function docker_desktop_version {
...
-echo "DDEV version: $(ddev version)"
-echo "PROXY settings: HTTP_PROXY='${HTTP_PROXY:-}' HTTPS_PROXY='${HTTPS_PROXY:-}' http_proxy='${http_proxy:-}' NO_PROXY='${NO_PROXY:-}'"
+header "User information"
+id -a
+
+header "ddev version"
+ddev version
+docker_platform=$(ddev version -j | docker run -i --rm ddev/ddev-utilities jq -r  '.raw."docker-platform"' 2>/dev/null)
+router_http_port=$(ddev config global -j | docker run -i --rm ddev/ddev-utilities jq -r  '.raw."router_http_port"' 2>/dev/null)
+router_https_port=$(ddev config global -j | docker run -i --rm ddev/ddev-utilities jq -r  '.raw."router_https_port"' 2>/dev/null)
```
stasadev, line 81:
```suggestion
router_http_port=$(ddev config global -j | docker run -i --rm ddev/ddev-utilities jq -r  '.raw."router-http-port"' 2>/dev/null)
router_https_port=$(ddev config global -j | docker run -i --rm ddev/ddev-utilities jq -r  '.raw."router-https-port"' 2>/dev/null)
```


```
======== curl -I of http://tryddevproject-6699.ddev.site:null from outside ========
curl: (3) URL rejected: Port number was not a decimal number between 0 and 65535


======== Full curl of http://tryddevproject-6699.ddev.site:null from outside ========
curl: (3) URL rejected: Port number was not a decimal number between 0 and 65535


======== Full curl of https://tryddevproject-6699.ddev.site:null from outside ========
curl: (3) URL rejected: Port number was not a decimal number between 0 and 65535
```

stasadev (changes requested):
Nice!

This results in 404, we don't need this anymore?

```
echo "Running ddev launch in 5 seconds" && sleep 5
ddev launch
```

There is definitely a lot to see, and I will continue the review on Monday.

rfay:
`ddev launch` should not get a 404. Was that because of the `router-http-port` problem?

Are you saying we should remove the `ddev launch`?

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -105,37 +162,29 @@ ddev start -y || ( \
...
-  echo "Unable to curl the requested project Please provide this output in a new gist at gist.github.com."
-  exit 1
-fi
 
-echo "Thanks for running the diagnostic. It was successful."
-echo "Please provide the output of this script in a new gist at gist.github.com"
+header 'Thanks for running the diagnostic!'
 echo "Running ddev launch in 5 seconds" && sleep 5
 ddev launch
+ddev stop
```
stasadev, line 190: > ddev launch should not get a 404. Was that because of the router-http-port problem?

I get 404 because I use [braus](https://braus.properlypurple.com/) to open links.

This means, I have some delay after `ddev launch`, because I select the browser to open the website.

And right after `ddev launch` there is `ddev stop`.

Maybe add some `sleep` between the two commands? Because I may not be the only one who will be affected by this.

rfay: thanks, that makes sense.

rfay:
* Working on gitpod
* Working on codespaces
* Working on macOS

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -30,66 +30,123 @@ function docker_desktop_version {
...
 
-echo "======= DDEV global info ========="
+header "DDEV global info"
 ddev config global | (grep -v "^web-environment" || true)
-printf "\n======= DOCKER info =========\n"
-echo -n "docker location: " && ls -l "$(which docker)"
-if [ ${OSTYPE%-*} != "linux" ]; then
+
+header "DOCKER provider info"
+echo -n "docker client location: " && ls -l "$(which docker)" && echo
```
stasadev, line 93:
```suggestion
echo -n "docker client location: " && ls -l "$(which -a docker)" && echo
```

stasadev (changes requested):
This is great!

Let's add more info about  all `ddev` and `mkcert` locations:

```
which -a ddev
which -a mkcert
```

rfay:
Added `which -a ddev` and `which -a mkcert`, good idea.

stasadev (approved):
> Working on Ubuntu, but it may hang at end (ddev launch not returning?)

Works good on Arch-based Linux.

PR #6003, opened 2024-03-21
build: use the official script to install Upsun

stasadev:
## The Issue

We can use the recommended official way to install the `upsun` cli.

Changes the installation script

```
ddev exec upsun version
ddev exec sudo apt update | grep upsun
```

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -100,12 +100,12 @@ RUN phpdismod xhprof
 RUN set -x && set -o pipefail && tag=$(curl -L --fail --silent "https://api.github.com/repos/axllent/mailpit/releases/latest" | jq -r .tag_name) && curl --fail -sSL "https://github.com/axllent/mailpit/releases/download/${tag}/mailpit-linux-${TARGETPLATFORM##linux/}.tar.gz" -o /tmp/mailpit.tar.gz && tar -zx -C /usr/local/bin -f /tmp/mailpit.tar.gz mailpit && rm /tmp/mailpit.tar.gz
 
 RUN curl -sSL --fail --output /usr/local/bin/phive "https://phar.io/releases/phive.phar" && chmod 777 /usr/local/bin/phive
+# Install terminus cli
 RUN set -o pipefail && curl --fail -sSL https://github.com/pantheon-systems/terminus/releases/download/$(curl -L --fail --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/terminus.phar --output /usr/local/bin/terminus && chmod 777 /usr/local/bin/terminus
+# Install platform cli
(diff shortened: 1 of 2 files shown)


containers/ddev-webserver/Dockerfile
```diff
@@ -101,7 +101,7 @@ RUN set -x && set -o pipefail && tag=$(curl -L --fail --silent "https://api.gith
 
 RUN curl -sSL --fail --output /usr/local/bin/phive "https://phar.io/releases/phive.phar" && chmod 777 /usr/local/bin/phive
 RUN set -o pipefail && curl --fail -sSL https://github.com/pantheon-systems/terminus/releases/download/$(curl -L --fail --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m{"tag_name": "\K.*?(?=")}g')/terminus.phar --output /usr/local/bin/terminus && chmod 777 /usr/local/bin/terminus
-RUN set -o pipefail && curl -fsSL https://raw.githubusercontent.com/platformsh/cli/main/installer.sh | bash
+RUN set -o pipefail && tag=$(curl -L --fail --silent "https://api.github.com/repos/platformsh/cli/releases/latest" | jq -r .tag_name) && curl --fail -sSL "https://github.com/platformsh/cli/releases/download/${tag}/platform_${tag}_linux_${TARGETPLATFORM##linux/}.tar.gz" -o /tmp/platform.tar.gz && tar -zx -C /usr/local/bin -f /tmp/platform.tar.gz platform && rm /tmp/platform.tar.gz
```
rfay, line 104: Since this is a go app, it should be fine like this, but remind me why you prefer to do it this way.  Why not do the Debian installation? https://github.com/platformsh/cli#ubuntudebian

stasadev: I did this to prevent situations like this:

- #5620

We have a workaround with `sudo apt update || true`, but it still increases the wait time if something is wrong with some apt repos.

I checked the script from https://github.com/platformsh/cli#ubuntudebian and saw that it also installs an additional apt repo for platform.sh.

stasadev:
I'm fine with the recommended installation method from the vendor.

And we don't need to change the script for Platform.sh because the current script calls the Debian script under the hood:

https://github.com/platformsh/cli/blob/483dca9a7f9a79c53d81a2d6de6c6f2d57e180a9/installer.sh#L436

@rfay, do you want me to replace the Upsun script with the official one here in this PR (since this is a related topic)?

```bash
curl -fsSL https://raw.githubusercontent.com/platformsh/cli/main/installer.sh | VENDOR=upsun bash
```

rfay:
> do you want me to replace the Upsun script with the official one here in this PR

If you'd like to do that I'm fine with changing upsun. Might as well make use of the PR while we have it going!

rfay (approved):
thanks for doing this!

GuySartorelli (approved):

rfay (approved):
This seems right, although it hasn't run the upsun integration tests yet, but should be fine. I'll pull it.

PR #6005, opened 2024-03-22
docs: Fix "webimage_extra_packages" code example

tyler36:
## The Issue

When I copied the original code into my config, the project did not start and I recieved the following error:

```shell
 ddev restart 
Failed to get project(s): /home/user13/code/drupal/modules/iq_ldap/.ddev/config.yaml exists but cannot be read. It may be invalid due to a syntax error: unable to load config file /home/user13/code/drupal/modules/iq_ldap/.ddev/config.yaml: invalid configuration in /home/user13/code/drupal/modules/iq_ldap/.ddev/config.yaml: yaml: line 14: did not find expected ',' or ']'
```

This PR updates the doc to resolve this.

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -10,7 +10,7 @@ It’s common to have a requirement for the `web` or `db` images which isn’t b
 You can add extra Debian packages with lines like this in `.ddev/config.yaml`:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


tyler36:
Also tested with the following which installs a valid version of pcov and enables code coverage in phpunit:

```yaml
webimage_extra_packages: ["php${DDEV_PHP_VERSION}-pcov"]
```

rfay (approved):
Thanks for this!

PR #6007, opened 2024-03-22
test: minor updates to buildkite testbot_maintenance.sh [skip ci]

rfay:
## The Issue

I noticed in https://github.com/ddev/maintainer-info/issues/1#issuecomment-2015447823 that brew stuff wasn't getting generally updated

* do `brew upgrade` on macOS instead of just doing a few packages
* In WSL2 we don't need mkcert because it's installed as dependency by ddev

The possible downside is that when buildkite-agent gets upgraded it may have the popups that prevent execution asking permissions.

### DIFF
--- a/.buildkite/testbot_maintenance.sh
+++ b/.buildkite/testbot_maintenance.sh
@@ -27,19 +27,20 @@ fi
 # Upgrade various items on various operating systems
 case $os in
 darwin)
+    brew upgrade
     for item in ddev/ddev/ddev golang golangci-lint libpq mkcert mkdocs; do
-        brew upgrade $item || brew install $item || true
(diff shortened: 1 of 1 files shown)


stasadev:
> The possible downside is that when buildkite-agent gets upgraded it may have the popups that prevent execution asking permissions.

I see this as a persistent problem with idle tests when the `buildkite-agent` is updating in the background.

We could [pin](https://stackoverflow.com/questions/10093918/ignore-formula-on-brew-upgrade) it to upgrade manually:
```
brew pin buildkite-agent
brew unpin buildkite-agent
```

rfay:
I like it.

stasadev (approved):

PR #6009, opened 2024-03-22
fix: hide obsolete ddev config --http-port

rfay:
## The Issue

In testing 
* https://github.com/ddev/ddev/pull/6002

I happened to notice that `--http-port` and `--https-port` are still shown in `ddev config -h`. But we don't really want those to be used any more.

Hide them

* `ddev config -h` should not show those
* `ddev config --http-port=8080 --https-port=8443` should still work and add that config to .ddev/config.yaml

Don't forget to remove the config from your .ddev/config.yaml


cmd/ddev/cmd/config.go
```diff
@@ -215,8 +215,10 @@ func init() {
 	ConfigCommand.Flags().StringVar(&projectTypeArg, "project-type", "", projectTypeUsage)
 	ConfigCommand.Flags().StringVar(&phpVersionArg, "php-version", "", "The version of PHP that will be enabled in the web container")
 	ConfigCommand.Flags().StringVar(&httpPortArg, "http-port", "", "The router HTTP port for this project")
-	ConfigCommand.Flags().StringVar(&httpPortArg, "router-http-port", "", "The router HTTP port for this project")
-	ConfigCommand.Flags().StringVar(&httpsPortArg, "https-port", "", "The router HTTPS port for this project")
+	ConfigCommand.Flags().StringVar(&httpPortArg, "router-http-port", "", "The router HTTP port for this project; alias for `--router-http-port`")
```
GuySartorelli, line 218: Am I reading this wrong? It's an alias for itself?

cmd/ddev/cmd/config.go
```diff
@@ -215,8 +215,10 @@ func init() {
...
+	_ = ConfigCommand.Flags().MarkDeprecated("http-port", "--http-port is a deprecated alias for `--router-http-port`")
+	ConfigCommand.Flags().StringVar(&httpsPortArg, "https-port", "", "The router HTTPS port for this project; alias for `--router-https-port`")
+	_ = ConfigCommand.Flags().MarkDeprecated("https-port", "--https-port is a deprecated alias for `--router-https-port`")
```
GuySartorelli, line 221: Could be worth explicitly saying "use `--router-https-port` instead" as a call to action, but doesn't ultimately matter too much.

cmd/ddev/cmd/config.go
rfay: Thanks, I put the text on the wrong one. http-port is alias for router-http-port, https-port is alias for router-https-port. You're awesome.

cmd/ddev/cmd/config.go
rfay: Since I ended up using the MarkDeprecated() I don't think it's too important. And nobody really should discover it.

cmd/ddev/cmd/config.go
GuySartorelli: Ahh that makes more sense 😅

GuySartorelli (approved):

PR #6010, opened 2024-03-22
feat: add PR build link to test on Codespaces [skip ci]

rfay:
## The Issue

We now provide a link on PRs to test in gitpod.. Why not codespaces also?


rfay:
Testing on https://github.com/ddev-test/ddev/pull/1/:

image

PR #6012, opened 2024-03-24
feat: provide corepack_enable config, fixes #5987, replaces #5988

### LINKED ISSUE #5987: Enable Corepack in the web container

rfay:
## The Issue

* #5987
* Replaces #5988

* New `corepack_enable` config for .ddev/config.yaml
* New `ddev config --corepack-enable`
* docs
* TestCorepackEnable

## TODO

- [x] Set COREPACK_HOME to point to /mnt/ddev-global-cache/corepack

* Use `ddev config --corepack-enable=true` and false (restart after)
* Review docs

## Automated Testing Overview

* TestCorepackEnable

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -84,11 +84,21 @@ The relative path, from the project root, to the directory containing `composer.
 
 ## `composer_version`
 
-Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command.
+Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command._
 
 | Type | Default | Usage
 | -- | -- | --
 | :octicons-file-directory-16: project | `2` | Can be `2`, `1`, or empty (`""`) for latest major version at container build time.<br><br>Can also be a minor version like `2.2` for the latest release of that branch, an explicit version like `1.0.22`, or a keyword like `stable`, `preview` or `snapshot`. See Composer documentation.
+
+## `corepack_enable`
+
+Whether to `corepack enable` on Node.js configuration.
+
+| Type | Default | Usage
+| -- | -- | --
+| :octicons-file-directory-16: project | `false` | Can be `true` or `false`.
+
+When `true`, `corepack_enable` will be executed, making latest `yarn` and `pnpm` package managers available.
 
 ## `database`
 
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -84,7 +84,7 @@ The relative path, from the project root, to the directory containing `composer.
 
 ## `composer_version`
 
-Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command._
+Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -1447,7 +1447,10 @@ ddev xhprof off
 Run [`yarn` commands](https://yarnpkg.com/cli) inside the web container in the root of the project (global shell host container command).
 
 !!!tip
-    Use `--cwd` for another directory.
+    Use `--cwd` for another directory, or you can change directories to the desired directory and `ddev yarn` will act on the same relative directory inside the container.
+
+!!!tip
+    If you want latest yarn versions, set `corepack_enable: true` in `.ddev/config.yaml` or `ddev config --corepack-enable`
 
 Example:
 
@@ -1458,6 +1461,14 @@ ddev yarn install
 # Use Yarn to add the Lerna package
 ddev yarn add lerna
 
+# Use yarn in a relative directory
+cd web/core && ddev yarn add lerna
+
 # Use Yarn to add the Lerna package from the `web/core` directory
 ddev yarn --cwd web/core add lerna
+
+# Use latest yarn or specified yarn
+ddev config --corepack-enabled && ddev restart
+ddev yarn set version stable
+ddev -yarn --version
 ```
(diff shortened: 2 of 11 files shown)


rpkoller:
I've tested the artifact while testing https://github.com/justafish/ddev-drupal-core-dev/pull/16/files. i've executed `ddev config --project-type=drupal10 --omit-containers=db --docroot="" --disable-settings-management --nodejs-version=20 --corepack-enable=true`

after that `corepack_enable: true`got added to the config,yaml and when checking the version of a drupal 11.x codebase i got version 1.x in the root and 4.x within the core folder on `ddev yarn --version`

docs/content/users/usage/commands.md
```diff
@@ -1458,6 +1461,14 @@ ddev yarn install
...
 ddev yarn add lerna
 
+# Use yarn in a relative directory
+cd web/core && ddev yarn add lerna
+
 # Use Yarn to add the Lerna package from the `web/core` directory
 ddev yarn --cwd web/core add lerna
+
+# Use latest yarn or specified yarn
+ddev config --corepack-enabled && ddev restart
```
stasadev, line 1471:
```suggestion
ddev config --corepack-enable && ddev restart
```

docs/content/users/usage/commands.md
```diff
@@ -1458,6 +1461,14 @@ ddev yarn install
...
+ddev yarn set version stable
+ddev -yarn --version
```
stasadev, line 1473:
```suggestion
ddev yarn --version
```

pkg/ddevapp/ddevapp.go
```diff
@@ -127,6 +127,7 @@ type DdevApp struct {
 	DisableSettingsManagement bool                   `yaml:"disable_settings_management,omitempty"`
 	WebEnvironment            []string               `yaml:"web_environment"`
 	NodeJSVersion             string                 `yaml:"nodejs_version,omitempty"`
+	CorepackEnable            bool                   `yaml:"corepack_enable"`
```
stasadev, line 130:
```suggestion
	CorepackEnable            bool                   `yaml:"corepack_enable,omitempty"`
```

docs/content/users/usage/commands.md
```diff
@@ -1447,7 +1447,10 @@ ddev xhprof off
 Run [`yarn` commands](https://yarnpkg.com/cli) inside the web container in the root of the project (global shell host container command).
 
 !!!tip
-    Use `--cwd` for another directory.
+    Use `--cwd` for another directory, or you can change directories to the desired directory and `ddev yarn` will act on the same relative directory inside the container.
+
+!!!tip
+    If you want latest yarn versions, set `corepack_enable: true` in `.ddev/config.yaml` or `ddev config --corepack-enable`
```
stasadev, line 1453:
```suggestion
    If you want latest `yarn` versions, set `corepack_enable: true` in `.ddev/config.yaml` or `ddev config --corepack-enable`
```

stasadev (changes requested):
This is indeed a much better solution 👍

docs/content/users/usage/commands.md
rpkoller: Instead wrapping the word Yarn in single quotes and making it a command i would vote to make it upper case and follow the wordmark which is used https://yarnpkg.com so it becomes a name/label

```suggestion
    If you want latest Yarn versions, set `corepack_enable: true` in `.ddev/config.yaml` or `ddev config --corepack-enable`
```
and since it sounds slightly odd wanting several of the latest versions why not slightly adjust the first part of the sentence to the gist of the third paragraph over here: https://yarnpkg.com/corepack

```suggestion
    If you want to define your Yarn version on a per project basis, set `corepack_enable: true` in `.ddev/config.yaml` or `ddev config --corepack-enable`
```

rpkoller (commented):
and would it make sense to add the term `Yarn` to the `.spellcheckwordlist.txt` file?

docs/content/users/usage/commands.md
rfay: Thanks, good stuff. Since "yarn" is a regular word, I don't think it needs to be added to spellcheck list.

pkg/ddevapp/ddevapp.go
rfay: I prefer to have this in the config I think, rather than disappear. I usually prefer to `omitempty` things that almost nobody ever needs to know about, but this one seems like it will want to be visible.

stasadev: OK, thanks.

docs/content/users/configuration/config.md
```diff
@@ -84,12 +84,22 @@ The relative path, from the project root, to the directory containing `composer.
 
 ## `composer_version`
 
-Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command.
+Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command._
```
stasadev, line 87:
```suggestion
Composer version for the web container and the [`ddev composer`](../usage/commands.md#composer) command.
```

stasadev (changes requested):

pkg/ddevapp/config.go
```diff
@@ -939,6 +939,9 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 		extraWebContent = extraWebContent + "\nRUN npm install -g n"
 		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN n install %s && ln -sf /usr/local/bin/node /usr/local/bin/nodejs", app.NodeJSVersion)
 	}
+	if app.CorepackEnable && app.NodeJSVersion >= "18" {
```
stasadev, line 942:
```suggestion
	if app.CorepackEnable {
```

It's probably better to let the user decide what to do. This check would have unexpected results with `nodejs_version` `20.1.0`, `v20`, etc.

stasadev (changes requested):

stasadev (approved):

Issue #5469, opened 2023-10-28
Add details for "INFO" in ddev describe e.g. for addons


ochorocho:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

No

### Describe your solution

I wanted to have more details when running `ddev describe` to inform a user about addon details e.g. username and password or other important information for the user.

We need to find a place to configure this for add-ons and non-add-ons.

### Additional context

<img width="1033" alt="Bildschirmfoto 2023-10-28 um 14 08 18" src="https://github.com/ddev/ddev/assets/4623070/a2b4ac06-f389-44c9-9e20-91d58faa6cf8">

My idea was simply to add it to `install.yaml`
```
name: rabbitmq
description: "Username: rabbitmq, Password: rabbitmq"
```

Downside: it would be add-ons only!

rfay:
It's a great idea. I think people have also wanted to just add info for specific projects, perhaps not in the section of a specific service.

adrianrudnik:
Is there a way to influence `URL/PORT` and `INFO` in a non-addon? I've look at mailpit and that seems to come from a [hardcoded](https://github.com/ddev/ddev/blob/master/cmd/ddev/cmd/describe.go#L184) line.

Would it be possible to not to limit it to `install.md`?

In my current project I need to add additional containers and services through custom docker compose files and I would love to add additional details to both columns.

Maybe also provide it inside the `config.yaml`

```yaml
describe:
  [servicename]:
    info: >
       Text that can be
       multiline
    urls:
      - Documentation: https://mydocsforservice
```

rfay:
@adrianrudnik All add-ons or 3rd party services,  if they have a HTTP_EXPOSE or HTTPS_EXPOSE they show up in `ddev describe` with the port information.  The code that takes care of this searches through all services, and does not depend on add-ons at all.

barkbyte:
is it possible to have a .secret file (or some other name) in each ddev addon, that would contain this info and have ddev describe "read" it?

rfay:
It would probably need to be a special hint somehow in the `docker-compose.*.yaml`, which is what `ddev describe` studies, or perhaps an environment variable?

adam-vessey:
> It would probably need to be a special hint somehow in the `docker-compose.*.yaml`, which is what `ddev describe` studies, or perhaps an environment variable?

[...]

PR #6019, opened 2024-03-26
docs: clarify additional compose service configuration

manuelmeister:
## The Issue
We made the mistake, of naming the container differently from the service, and thus not being able to access the service. We somehow overlooked the part about the container name. Also we wanted to have a different hostname and did not know if we have to set `VIRTUAL_HOST=$DDEV_HOSTNAME` or if this has side effects. If everything is configured correctly we can expose a completely different service on an arbitrary hostname.

I've clarified the wording. I mention traefik in the hostname conventions to explain why this is needed for smooth operations.
I've added an example in the VIRTUAL_HOST part to show what would be possible there. (hostname taken from a [ddev test](https://github.com/ddev/ddev/blob/v1.22.7/pkg/ddevapp/testdata/TestTraefikVirtualHost/docker-compose.extra.yaml#L12))

Maybe it would be sensible to add a note to https://ddev.readthedocs.io/en/stable/users/extend/additional-hostnames/ and mention the ability to set arbitrary hostnames for custom services besides the `web` service.

### DIFF
--- a/docs/content/users/extend/custom-compose-files.md
+++ b/docs/content/users/extend/custom-compose-files.md
@@ -52,7 +52,7 @@ To better understand how DDEV parses your custom docker-compose files, run `ddev
 
 When defining additional services for your project, we recommended following these conventions to ensure DDEV handles your service the same way DDEV handles default services.
 
-* The container name should be `ddev-${DDEV_SITENAME}-<servicename>`.
+* The container name should be `ddev-${DDEV_SITENAME}-<servicename>`. This ensures the auto-generated [traefik routing configuration](./traefik-router.md#project-traefik-configuration) matches your custom service.
 * Provide containers with required labels:
 
     ```yaml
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks good to me, thanks!

manuelmeister:
The only thing I was considering was adding a link/note on [Additional Project Hostnames](https://ddev.readthedocs.io/en/latest/users/extend/additional-hostnames/) to [Defining Additional Services with Docker Compose](https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/#defining-additional-services-with-docker-compose). But I think this PR improves the situation already.

rfay:
> The only thing I was considering was adding a link/note on [Additional Project Hostnames](https://ddev.readthedocs.io/en/latest/users/extend/additional-hostnames/) to [Defining Additional Services with Docker Compose](https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/#defining-additional-services-with-docker-compose). But I think this PR improves the situation already.

Feel free to add another commit!

manuelmeister:
I just figured out, that it is not possible to customize the describe output of an additional compose service. Therefore I don''t think this is necessary. I'd rather have the ability to customize part of the describe and be able to write plaintext info for my custom service. 
I've never coded in Go and would need additional guidance, or should I open an issue for this?

rfay:
* https://github.com/ddev/ddev/issues/5469

It fits with your question. And adding VIRTUAL_HOST wouldn't be too hard. It would be in app.Describe.

A contribution would be welcome! Lots of support in https://ddev.readthedocs.io/en/stable/developers/

docs/content/users/extend/custom-compose-files.md
```diff
@@ -52,7 +52,7 @@ To better understand how DDEV parses your custom docker-compose files, run `ddev
 
 When defining additional services for your project, we recommended following these conventions to ensure DDEV handles your service the same way DDEV handles default services.
```
rfay, line 53:
```suggestion
When defining additional services for your project, we recommend following these conventions to ensure DDEV handles your service the same way DDEV handles default services.
```

docs/content/users/extend/custom-compose-files.md
```diff
@@ -52,7 +52,7 @@ To better understand how DDEV parses your custom docker-compose files, run `ddev
...
 
-* The container name should be `ddev-${DDEV_SITENAME}-<servicename>`.
+* The container name should be `ddev-${DDEV_SITENAME}-<servicename>`. This ensures the auto-generated [traefik routing configuration](./traefik-router.md#project-traefik-configuration) matches your custom service.
```
rfay, line 55:
```suggestion
* The container name should be `ddev-${DDEV_SITENAME}-<servicename>`. This ensures the auto-generated [Traefik routing configuration](./traefik-router.md#project-traefik-configuration) matches your custom service.
```

rfay (approved):
Thanks!

PR #6020, opened 2024-03-27
feat: add pint command to laravel projects

penyaskito:
## The Issue

Laravel includes its own linter, pint, which is installed by default with 11.x

Adds the pint shortcut command.

I've tested this on Gitpod:

* Clean the Drupal test data first:

```
ddev delete -Oy d10simple
rm d10simple/ -rf
```

* Create a laravel project:

```
mkdir laravel
cd laravel/
ddev config --project-type=laravel --docroot=public --php-version=8.2
ddev start
``` 

* As you didn't install laravel yet,  `pint` should fail with instructions to install:

```
ddev pint
```

* Install laravel:

```
ddev composer create --prefer-dist laravel/laravel:^11
```

* Now `ddev pint` should succeed:

```
ddev pint
```

* And you will see

```
PASS   26 files  
```

## Automated Testing Overview

Added a test verifying the command is added to Laravel projects.

Run the individual test with `go test -v -run TestCustomCommands ./cmd/ddev/cmd/`

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/pint
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/pint
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run Pint inside the web container
+## Usage: pint [flags] [args]
+## Example: "ddev pint --dirty" for fixing your current changes lint errors or "ddev pint --test app/Models" for only analyzing your models without autofix. 
+## ProjectTypes: laravel
+
+./vendor/bin/pint "$@"
+
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/pint
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/pint
@@ -6,5 +6,10 @@
 ## Example: "ddev pint --dirty" for fixing your current changes lint errors or "ddev pint --test app/Models" for only analyzing your models without autofix. 
 ## ProjectTypes: laravel
 
-./vendor/bin/pint "$@"
+if ! command -v pint >/dev/null; then
+  echo "pint is not available. You may need to 'ddev composer require laravel/pint'"
+  exit 1
+fi
+
+pint "$@"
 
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/global_dotddev_assets/commands/web/pint
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run Pint inside the web container
+## Usage: pint [flags] [args]
+## Example: "ddev pint --dirty" for fixing your current changes lint errors or "ddev pint --test app/Models" for only analyzing your models without autofix. 
+## ProjectTypes: laravel
+
+./vendor/bin/pint "$@"
```
rfay, line 9: vendor/bin is already in the $PATH, so this can just be
```suggestion
pint "$@"
```

However, `pint` is not always available/installed, right? So you might want to do what the drush command does and tell people what to do,

https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/drush#L10-L13

rfay (changes requested):
Biggest thing I see right now is ... does everybody on laravel need this? Or would it be better as an Add-on?

pkg/ddevapp/global_dotddev_assets/commands/web/pint
penyaskito: > vendor/bin is already in the $PATH, so this can just be

Is this new? Didn't work for me with v1.22.7 when I initially tried it that way.

> However, pint is not always available/installed, right?

Yeah, wanted to try this on gitpod, but that required I create the PR first (and that's why I used a draft)
I will check what the last releases of laravel have, but probably the sanity check makes sense anyway (even if it's a default, I could remove the package on my project)

rfay: > Is this new? Didn't work for me with v1.22.7 when I initially tried it that way.

vendor/bin has been in $PATH for years.

> Yeah, wanted to try this on gitpod, but that required I create the PR first (and that's why I used a draft)
I will check what the last releases of laravel have, but probably the sanity check makes sense anyway (even if it's a default, I could remove the package on my project)

You may like it better as an add-on, don't know.

rfay:
BTW, because you're special, you're a member of the DDEV org, and that allows you to push a branch, but please work on your fork rather than in ddev/ddev (the branch for this PR was created in ddev/ddev)

stasadev (approved):
Thanks!

Tested with this to see the code change:
```
ddev pint --preset psr12 -v
ddev pint -v
```

rfay (approved):
Thanks!

PR #6024, opened 2024-03-27
feat: accept port in `ddev launch`, for #5525

stasadev:
## The Issue

- #5525

I added an example in #5986 to run `ddev launch` with a specific port:

```
# Open your project’s base URL using a specific port
ddev launch $DDEV_PRIMARY_URL:3000
```

But this never actually worked:

```
DDEV_DEBUG=true ddev launch $DDEV_PRIMARY_URL:3000
FULLURL https://d10.ddev.site/:3900
```

`$DDEV_PRIMARY_URL` is not expanded because it is not available on the host side and `$DDEV_PRIMARY_URL:3000` and `:3000` are the same thing.

`ddev launch $DDEV_PRIMARY_URL:3000` only works if it is run in another DDEV custom script.

Adds actual support to launch ports.

```
ddev debug fix-commands
```

```
$ DDEV_DEBUG=true ddev launch :3000
FULLURL https://d10.ddev.site:3000
```

```
$ DDEV_DEBUG=true ddev launch :3000/test
FULLURL https://d10.ddev.site:3000/test
```

See a tip for HTTP to HTTPS redirect at https://ddev--6024.org.readthedocs.build/en/6024/users/usage/commands/#launch

```
$ DDEV_DEBUG=true ddev launch :8025
HTTP may redirect to HTTPS in your browser
See https://ddev.readthedocs.io/en/stable/users/usage/commands/#launch
FULLURL http://d10.ddev.site:8025
```

```
$ DDEV_DEBUG=true ddev launch :8026
FULLURL https://d10.ddev.site:8026
```

### DIFF
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -350,6 +350,8 @@ func TestLaunchCommand(t *testing.T) {
 		"nested/path":                      app.GetPrimaryURL() + "/nested/path",
 		"/path-with-slash":                 app.GetPrimaryURL() + "/path-with-slash",
 		app.GetPrimaryURL() + "/full-path": app.GetPrimaryURL() + "/full-path",
+		"http://example.com":               "http://example.com",
+		"https://example.com:443/test":     "https://example.com:443/test",
 		":8080":                            desc["httpurl"].(string),
 		":8080/http-port-path":             desc["httpurl"].(string) + "/http-port-path",
 		":8443":                            desc["httpsurl"].(string),
@@ -357,10 +359,8 @@ func TestLaunchCommand(t *testing.T) {
 		":18025":                           "http://" + app.GetHostname() + ":18025",
 		":18026":                           "https://" + app.GetHostname() + ":18026",
 		// if it is impossible to determine the http/https scheme, the default site protocol should be used
-		":3000":                        primaryURLWithoutPort + ":3000",
-		":3000/with-default-scheme":    "https://" + app.GetHostname() + ":3000/with-default-scheme",
-		"http://example.com":           "http://example.com",
-		"https://example.com:443/test": "https://example.com:443/test",
+		":3000":                     primaryURLWithoutPort + ":3000",
+		":3000/with-default-scheme": "https://" + app.GetHostname() + ":3000/with-default-scheme",
 	}
 	if runtime.GOOS == "windows" {
 		// Git-Bash converts single forward slashes to a Windows path
@@ -368,8 +368,9 @@ func TestLaunchCommand(t *testing.T) {
 		cases["//path-with-slash"] = app.GetPrimaryURL() + "/path-with-slash"
 		delete(cases, "/path-with-slash")
 	}
-	if globalconfig.DdevGlobalConfig.MkcertCARoot == "" {
+	if app.CanUseHTTPOnly() {
 		cases["-m"] = desc["mailpit_url"].(string)
+		cases[":3000/with-default-scheme"] = "http://" + app.GetHostname() + ":3000/with-default-scheme"
 	}
 	for partialCommand, expect := range cases {
 		// Try with the base URL, simplest case
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -331,6 +331,7 @@ func TestLaunchCommand(t *testing.T) {
 		_ = os.RemoveAll(testDir)
 	})
 
+	primaryURLWithoutPort := app.GetPrimaryURL()
 	// This only tests the https port changes, but that might be enough
 	app.RouterHTTPSPort = "8443"
 	err = app.WriteConfig()
@@ -347,6 +348,7 @@ func TestLaunchCommand(t *testing.T) {
 		"nested/path":                      app.GetPrimaryURL() + "/nested/path",
 		"/path-with-slash":                 app.GetPrimaryURL() + "/path-with-slash",
 		app.GetPrimaryURL() + "/full-path": app.GetPrimaryURL() + "/full-path",
+		":3000":                            primaryURLWithoutPort + ":3000",
 		"http://example.com":               "http://example.com",
 		"https://example.com:443/test":     "https://example.com:443/test",
 	}
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -332,8 +332,10 @@ func TestLaunchCommand(t *testing.T) {
 	})
 
 	primaryURLWithoutPort := app.GetPrimaryURL()
-	// This only tests the https port changes, but that might be enough
+	app.RouterHTTPPort = "8080"
 	app.RouterHTTPSPort = "8443"
+	app.MailpitHTTPPort = "18025"
+	app.MailpitHTTPSPort = "18026"
 	err = app.WriteConfig()
 	assert.NoError(err)
 	err = app.Start()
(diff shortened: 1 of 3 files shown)


cmd/ddev/cmd/commands_test.go
```diff
@@ -347,6 +348,7 @@ func TestLaunchCommand(t *testing.T) {
 		"nested/path":                      app.GetPrimaryURL() + "/nested/path",
 		"/path-with-slash":                 app.GetPrimaryURL() + "/path-with-slash",
 		app.GetPrimaryURL() + "/full-path": app.GetPrimaryURL() + "/full-path",
+		":3000":                            primaryURLWithoutPort + ":3000",
```
penyaskito, line 351: Should we add another test for ":3000/test"?

```suggestion
		":3000":                            primaryURLWithoutPort + ":3000",
		":3000/test":                     primaryURLWithoutPort + ":3000/test",
```

penyaskito (commented):
Tested this on gitpod and worked as expected.

pkg/ddevapp/global_dotddev_assets/commands/host/launch
```diff
@@ -52,15 +52,16 @@ while :; do
...
-    FULLURL="${FULLURL}/";
-  fi
-
+if [ -n "${1:-}" ]; then
   if [[ $1 =~ ^https?:// ]]; then
+    # full url
     FULLURL="${1}";
+  elif [[ $1 =~ ^: ]]; then
+    # specific port
+    FULLURL="${FULLURL%:[0-9]*}${1}";
```
GuySartorelli, line 61: What happens if the port is for `http` but the `FULLURL` is https? Or the other way around?

cmd/ddev/cmd/commands_test.go
rfay: I'm surprised it worked on gitpod, which is very picky about these things, and doesn't use `ddev-router`.

penyaskito: As gitpod is special in its port handling, I'll be more concrete:

- `ddev launch`: opens https://8080-stasadev-ddev-ajl5zssk1i2.ws-eu110.gitpod.io (made sense to me)
- `ddev launch :3000`: opens https://3000-stasadev-ddev-ajl5zssk1i2.ws-eu110.gitpod.io, not found (made sense to me)
- `ddev launch :3000/test`: opens https://3000-stasadev-ddev-ajl5zssk1i2.ws-eu110.gitpod.io/test, not found (made sense to me)

pkg/ddevapp/global_dotddev_assets/commands/host/launch
penyaskito: The protocol is not changed on ddev (might be redirected by e.g. nginx), so you can end on

```
400 Bad Request
The plain HTTP request was sent to HTTPS port
nginx
```

cmd/ddev/cmd/commands_test.go
penyaskito: Also:

```console
gitpod /workspace/d10simple (main) $ DDEV_DEBUG=true ddev launch :3000
FULLURL http://127.0.0.1:3000
gitpod /workspace/d10simple (main) $ DDEV_DEBUG=true ddev launch :3000/test
FULLURL http://127.0.0.1:3000/test
```

pkg/ddevapp/global_dotddev_assets/commands/host/launch
GuySartorelli: That's what I thought was probably the case.
It feels like this PR is missing a trick there. With the original PR (and the [alternative I suggested](https://github.com/ddev/ddev/pull/5525#issuecomment-1979759118)) checking HTTP vs HTTPS and setting the correct port is managed, so we don't get into this situation.

Right now `ddev launch` has logic to check whether it should be using HTTP or HTTPS - but that logic isn't going to be leveraged here.

GuySartorelli: Looking at [ddev adminer](https://github.com/ddev/ddev-adminer/blob/main/commands/host/adminer) for example, there's just as much code figuring out what the URL should be (http vs https etc) as there is launching it.

cmd/ddev/cmd/commands_test.go
stasadev: Thanks, I added more tests.

pkg/ddevapp/global_dotddev_assets/commands/host/launch
stasadev: > What happens if the port is for http but the FULLURL is https? Or the other way around?
> It feels like this PR is missing a trick there. 

Good point, thank you. I added detection for http/https.

stasadev: > Looking at [ddev adminer](https://github.com/ddev/ddev-adminer/blob/main/commands/host/adminer) for example, there's just as much code figuring out what the URL should be (http vs https etc) as there is launching it. Ideally that could all be abstracted away together.

The port for adminer can be shortened right now to (assuming someone is using DDEV v1.23.0-alpha1):

```bash
DDEV_ADMINER_PORT=9100
DDEV_ADMINER_HTTPS_PORT=9101

if [ ${DDEV_PRIMARY_URL%://*} = "https" ]; then
    ddev launch $DDEV_PRIMARY_URL:$DDEV_ADMINER_HTTPS_PORT
else
    ddev launch $DDEV_PRIMARY_URL::$DDEV_ADMINER_PORT
fi
```

or after this PR:

```bash
DDEV_ADMINER_PORT=9100
DDEV_ADMINER_HTTPS_PORT=9101

if [ ${DDEV_PRIMARY_URL%://*} = "https" ]; then
    ddev launch :$DDEV_ADMINER_HTTPS_PORT
else
    ddev launch :$DDEV_ADMINER_PORT
fi
```

But I'm not going to change it in the near future, because people can use the old DDEV with the new add-on version. And we have no restrictions on this. Ideally, add-ons should work like composer packages and pull only compatible add-on version, but at the moment that's probably too much work.

tyler36:
Maybe I'm missing something, but this works in `1.22.7`, no?

```shell
$ ddev -v          
ddev version v1.22.7

$ DDEV_DEBUG=true ddev launch :8026
FULLURL https://tempest-l11-demo.ddev.site:8026

$ DDEV_DEBUG=true ddev launch $DDEV_PRIMARY_URL:3000
FULLURL https://tempest-l11-demo.ddev.site:3000
```

tyler36:
Please ignore my previous comment. 

I updated to HEAD, played around and have now `brew link` back to a `1.22.7`. I'm see the original issue as described and *not* seeing what I wrote above. 

```shell
$ ddev -v
ddev version v1.22.7

$ DDEV_DEBUG=true ddev launch :8026
FULLURL https://tempest-l11-demo.ddev.site/:8026
```

Appologies for the false alarm.

stasadev:
> I'm see the original issue as described and *not* seeing what I wrote above. 

I had a similar situation, the port worked for me at some point, and I added it to the documentation. But when I tried it later, it turned out that it didn't work at all.

tyler36 (commented):
Tested on WSL2 (Ubuntu 22.04.4)

```shell
ddev version v1.23.0-alpha1-4-g71b07fb61
$ ddev -v

$ DDEV_DEBUG=true ddev launch :3000
FULLURL https://lara11-base-demo.ddev.site:3000
```

rfay (approved):
Thanks for the great work on this, and especially adding the extra cases to the test. It tests great manually and code is great. Thanks!

PR #6025, opened 2024-03-27
fix: handle wildcard hostnames when name resolution not available, fixes #4881

### LINKED ISSUE #4881: Using wildcard in `additional_hostnames` prompts for admin privileges

`ddev debug test` diagnostic information: https://gist.github.com/pyrello/81b43cba5124f073bd6bcc730c1d55a1  

Opening this bug per @rfay in conversation in Discord.

### Expected Behavior

Not requiring administrative privileges to run `ddev start`

### Actual Behavior

Running `ddev start` requires admin privileges with our setup.

There are two bugs here:

1. We shouldn't try to resolve an invalid hostname (`*.x.ddev.site`)
2. We shouldn't try to add a wildcard to /etc/hosts, as it won't work there anyway.

Probably we should just notify people that if their wildcard is not resolvable (`*.*.ddev.site` should normally be resolvable) then they need to make it resolvable one way or another.

### Steps To Reproduce

```
ddev config --additional-hostnames="*.junk"
ddev start
```

### Anything else?

In our project, we use DDEV for an application that runs more than 700 sites. We have it setup to provide all the sites locally at an address like `*.uiowa.ddev.site`. After upgrading to a new computer that does not allow me to have continuous access to administrative privileges, I am running into an issue where I see the following output when I run `ddev start`:
```
ddev start
Starting uiowa... 
The hostname *.uiowa.ddev.site is not currently resolvable, trying to add it to the hosts file 
ddev needs to run with administrative privileges.
You may be required to enter your password for sudo or allow escalation. ddev is about to issue the command:
  sudo --preserve-env=HOME /opt/homebrew/bin/ddev hostname *.uiowa.ddev.site 127.0.0.1
 
Password:
Not populating custom commands or hostadditions because running with root privileges
Failed to add hosts entry *.uiowa.ddev.site: hostname is not a valid dns name: *.uiowa.ddev.site
```

### Workaround

Adding a `config.local.yaml` file with the following settings allowed me to start ddev without it asking for my password:
```yml
override_config: true
additional_hostnames:
  - sppa.uiowa
```

This allows the site https://sppa.uiowa.ddev.site to be available locally. The `config.local.yaml` file does not get checked in.

rfay:
## The Issue

* #4881 

There are a few problems with creating hostnames in /etc/hosts when name lookup fails. 

Tell when we can't do it and don't try doing it.

```
ddev config --use-dns-when-possible=false
ddev config --additional-hostnames="*.junk"
ddev start
```

You should get a warning that `*.junk.ddev.site`  can't be added to hosts file... because hosts file doesn't do wildcards.

## Release/Deployment Notes

- [ ] Followup: A ddev.com article explaining all of the issues with name resolution and wildcards.

### DIFF
--- a/pkg/ddevapp/hostname_mgt.go
+++ b/pkg/ddevapp/hostname_mgt.go
@@ -112,7 +112,7 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 			continue
 		}
 		if !govalidator.IsDNSName(name) {
-			util.Warning("DDEV cannot add unresolvable hostnames like `%s` to your hosts file.\nSee docs for more info, https://ddev.readthedocs.io/en/stable/users/configuration/config/#additional_hostnames.", name)
+			util.Warning("DDEV cannot add unresolvable hostnames like `%s` to your hosts file.\nSee docs for more info, https://ddev.readthedocs.io/en/stable/users/configuration/config/#additional_hostnames", name)
 		} else {
 			util.Warning("The hostname %s is not currently resolvable, trying to add it to the hosts file", name)
 			out, err := escalateToAddHostEntry(name, dockerIP)
--- a/pkg/ddevapp/hostname_mgt.go
+++ b/pkg/ddevapp/hostname_mgt.go
@@ -87,7 +87,7 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 
 		// If we're able to resolve the hostname via DNS or otherwise we
 		// don't have to worry about this. This will allow resolution
-		// of *.ddev.site for example
+		// of <whatever>.ddev.site for example
 		if app.UseDNSWhenPossible && globalconfig.IsInternetActive() {
 			// If they have provided "*.<name>" then look up the suffix
 			checkName := strings.TrimPrefix(name, "*.")
@@ -110,13 +110,16 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 			util.Warning("Unable to open hosts file: %v", err)
 			continue
 		}
-		util.Warning("The hostname %s is not currently resolvable, trying to add it to the hosts file", name)
-
-		out, err := escalateToAddHostEntry(name, dockerIP)
-		if err != nil {
-			return err
+		if strings.Contains(name, `*`) {
+			util.Warning("DDEV cannot add unresolvable wildcard hostnames like `%s` to your hosts file", name)
+		} else {
+			util.Warning("The hostname %s is not currently resolvable, trying to add it to the hosts file", name)
+			out, err := escalateToAddHostEntry(name, dockerIP)
+			if err != nil {
+				return err
+			}
+			util.Success(out)
 		}
-		util.Success(out)
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


pkg/ddevapp/hostname_mgt.go
```diff
@@ -110,13 +110,16 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 			util.Warning("Unable to open hosts file: %v", err)
 			continue
 		}
-		util.Warning("The hostname %s is not currently resolvable, trying to add it to the hosts file", name)
-
-		out, err := escalateToAddHostEntry(name, dockerIP)
-		if err != nil {
-			return err
+		if strings.Contains(name, `*`) {
```
stasadev, line 113: We can use this check to cover all possible scenarios (and not run `sudo`):

https://github.com/ddev/ddev/blob/18e53dcc49440f621596387b40d1b942333e1587/vendor/github.com/goodhosts/hostsfile/hosts.go#L187-L189

This will require an update to `go.mod` (move the dependency from the indirect section).

stasadev (changes requested):

pkg/ddevapp/hostname_mgt.go
rfay: I like it.

pkg/ddevapp/hostname_mgt.go
```diff
@@ -110,13 +110,16 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
...
+			util.Warning("DDEV cannot add unresolvable wildcard hostnames like `%s` to your hosts file", name)
```
skeemer, line 114: Could this also point to docs or mention that there are probably ways to resolve this? The message is completely true, but still makes me think I misconfigured the project rather than having a general networking issue.

pkg/ddevapp/hostname_mgt.go
rfay: Currently we have used ValidateProjectName for this, https://github.com/ddev/ddev/blob/1daf7841659c3a020d044bb49d73ffa7fe9f6a7d/pkg/ddevapp/config.go#L433-L439

pkg/ddevapp/hostname_mgt.go
rfay: Added link to new doc.

pkg/ddevapp/hostname_mgt.go
```diff
@@ -110,13 +111,16 @@ func (app *DdevApp) AddHostsEntriesIfNeeded() error {
 			util.Warning("Unable to open hosts file: %v", err)
 			continue
 		}
-		util.Warning("The hostname %s is not currently resolvable, trying to add it to the hosts file", name)
-
-		out, err := escalateToAddHostEntry(name, dockerIP)
-		if err != nil {
-			return err
+		if !govalidator.IsDNSName(name) {
+			util.Warning("DDEV cannot add unresolvable hostnames like `%s` to your hosts file.\nSee docs for more info, https://ddev.readthedocs.io/en/stable/users/configuration/config/#additional_hostnames.", name)
```
stasadev, line 115: The anchor doesn't work work with dot.

```suggestion
			util.Warning("DDEV cannot add unresolvable hostnames like `%s` to your hosts file.\nSee docs for more info, https://ddev.readthedocs.io/en/stable/users/configuration/config/#additional_hostnames", name)
```

rfay: Thanks, wonder where that came from.

stasadev (approved):
Tested, works fine, no sudo for unresolvable hostnames.

A little confusion with:

```
ddev start
...
Successfully started d10 
Project can be reached at https://d10.ddev.site https://*.junk.ddev.site https://127.0.0.1:32774
```

`https://*.junk.ddev.site` is added to the list of hostnames anyway.

I thought about moving the warning to `app.GetHostnames()`, but that's probably not the best idea because it's used in so many places.

And hiding it in `pkg/ddevapp/hostname_mgt.go` doesn't hide it when running `ddev describe` or when checking the value of `$DDEV_HOSTNAME`.

Either way, I'm fine with this PR, the user is informed of the problem and has to solve it on their own.

Issue #5073, opened 2023-07-04
<none> image is being created when an add-on downloads assets to the web container


rpkoller:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
$> ddev debug test                                                  
Running bash [-c /var/folders/6h/vbkvz2wj38j0rlp7v_jcwj0m0000gn/T/test_ddev.sh] 
======= Existing project config =========
These config files were loaded for project try: [/Users/rkoller/Sites/try/.ddev/config.yaml /Users/rkoller/Sites/try/.ddev/config.ddev-spidergram.yaml /Users/rkoller/Sites/try/.ddev/config.mailpit.yaml]
name: try
type: drupal10
docroot: web
php_version: 8.1
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.22.0-beta1
database: {mariadb 10.4}
hooks: map[post-start:[map[exec:perl -pi -e "s/(url?:\ https?:\/\/).*\d{4}/\1$DDEV_PROJECT.$DDEV_TLD:8529/g" ./spidergram.config.yaml] map[exec:perl -pi -e "s/(url?:\ https?:\/\/).*\d{4}/\1$DDEV_PROJECT.$DDEV_TLD:8529/g" ./spidergram.config.yaml]]]
upload_dirs: []
mailhog_port: 9025
mailhog_https_port: 9026
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 18
default_container_timeout: 120
Instrumentation is opted in, but SegmentKey is not available. This usually means you have a locally-built ddev binary or one from a PR build. It's not an error. Please report it if you're using an official release build.
Instrumentation is opted in, but AmplitudeAPIKey is not available. This usually means you have a locally-built ddev binary or one from a PR build. It's not an error. Please report it if you're using an official release build.
======= Creating dummy project named  tryddevproject-8624 in ../tryddevproject-8624 =========
OS Information: Darwin elysium.local 21.6.0 Darwin Kernel Version 21.6.0: Thu Sep 29 20:13:56 PDT 2022; root:xnu-8020.240.7~1/RELEASE_ARM64_T6000 arm64
ProductName:	macOS
ProductVersion:	12.6.1
BuildVersion:	21G217
[...]

PR #6029, opened 2024-03-28
fix: clean up missing project if it's gone, fixes #5322

### LINKED ISSUE #5322: 'unable to create project at project root' error message should say how to fix the problem

```
```
</details>         

### Expected Behavior

I did `ddev list` and there is a warning:

> unable to create project at project root '/Users/joachim/Sites/pants': project root /Users/joachim/Sites/pants does not exist

I googled for how to delete a project and at https://ddev.readthedocs.io/en/latest/users/usage/uninstall/ it says to do `[ddev delete <project>](https://ddev.readthedocs.io/en/latest/users/usage/commands/#delete)`. But I assume that's a project name - I don't know what that is.

There's a mention of this case at #2314.

But the error message should give the user next steps to fix the problem.

### Actual Behavior

Error message that the user can't act on.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #5322

People periodically delete a project's filesystem without doing a `ddev delete` first

When the filesystem is gone, remove the project from the project list

```
mkdir junk && cd junk && ddev config --auto
cd .. && rm -r junk
ddev list
```

You should see the project mentioned while it's being removed from registry,  but it won't error and won't show up on the next `ddev list`.

## Automated Testing Overview

Added TestGetProjectsMissingApp

### DIFF
--- a/cmd/ddev/cmd/describe_test.go
+++ b/cmd/ddev/cmd/describe_test.go
@@ -4,15 +4,12 @@ import (
 	"fmt"
 	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/stretchr/testify/require"
-	"runtime"
 	"strings"
 	"testing"
 
 	"encoding/json"
 
 	"os"
-
-	"path/filepath"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/exec"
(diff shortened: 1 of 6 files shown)


rfay:
One thing we may still want to do here is to support `ddev delete <project>` when the project is not in the global list. That way people having trouble with an existing database could explicitly delete it in the easiest way.

stasadev (approved):
Looks good to me, I like the automatic removal of a project from the list.

> One thing we may still want to do here is to support `ddev delete <project>` when the project is not in the global list. That way people having trouble with an existing database could explicitly delete it in the easiest way.

This is a good idea, but probably beyond the scope of this PR.
I'll think about it when I come back to the task of cleaning up the docker's resources:

- #5073

rfay:
> This is a good idea, but probably beyond the scope of this PR.
> I'll think about it when I come back to the task of cleaning up the docker's resources: #5073 

That one is about docker images, here we're talking about the docker volume where the database is stored. I agree it belongs in a separate PR.

PR #6031, opened 2024-03-29
fix: revert removal of Colima host.docker.internal setting

rfay:
## The Issue

We're having a bit of trouble with NFS in colima/lima tb-macos-arm64-5/6/7, I'm not sure why. But we don't need to be testing NFS on macOS anyway, certainly not on orbstack/lima/colima.

* https://github.com/ddev/ddev/pull/5994

Lima provides 192.168.5.2 as host.docker.internal, but colima does not seem to have it any more. I assume colima lost it in a recent Lima upgrade.

This probably means that Xdebug is not working on colima since #5994 , but I don't know why it wouldn't have been caught by automated tests.

Revert #5994 to put explicit host.docker.internal back in there again

`ddev debug nfsmount` using colima

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -1253,6 +1253,11 @@ func GetNFSServerAddr() (string, error) {
 	nfsAddr := "host.docker.internal"
 
 	switch {
+	case IsColima():
+		// Lima specifies this as a named explicit IP address at this time
+		// see https://github.com/lima-vm/lima/blob/master/docs/network.md#host-ip-19216852
+		nfsAddr = "192.168.5.2"
+
 	// Gitpod has Docker 20.10+ so the docker-compose has already gotten the host-gateway
 	// However, NFS will never be used on Gitpod.
 	case nodeps.IsGitpod():
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -1587,6 +1587,20 @@ func IsLima() bool {
 	return false
 }
 
+// IsOrbstack detects if running on Orbstack
+func IsOrbstack() bool {
+	ctx, client := GetDockerClient()
+	info, err := client.Info(ctx)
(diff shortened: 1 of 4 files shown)


pkg/dockerutil/dockerutils.go
```diff
@@ -1587,6 +1587,20 @@ func IsLima() bool {
 	return false
 }
 
+// IsOrbstack detects if running on Orbstack
+func IsOrbstack() bool {
+	ctx, client := GetDockerClient()
+	info, err := client.Info(ctx)
+	if err != nil {
+		util.Warning("IsLima(): Unable to get Docker info, err=%v", err)
```
stasadev, line 1595:
```suggestion
		util.Warning("IsOrbstack(): Unable to get Docker info, err=%v", err)
```

rfay:
I'm probably chasing the wrong problem here, have to look closer.
> Details: error=failed to StartContainer: Error response from daemon: error while mounting volume '/var/lib/docker/volumes/testnfsmount/_data': error resolving passed in network volume address: lookup host.docker.internal on 127.0.0.53:53: no such host

https://buildkite.com/ddev/macos-colima-vz/builds/380#018e86b2-655e-4ed5-b24e-5bf8be98d32c

stasadev:
Can it be related to this change?

- #5994

stasadev:
> Yes, that's what it is. I tested and colima is not resolving host.docker.internal any more even though lima is.

Is this just a partial revert for NFS or is the Colima check needed again in `GetHostDockerInternalIP()` ?

stasadev (approved):

rfay:
Here's the reason we didn't catch this in automated tests
https://github.com/ddev/ddev/blob/9d73fe52bf21203e00cc97efc78484be0ce43564/pkg/ddevapp/ddevapp_test.go#L905-L911

rfay:
Oh I see, we're running the check using brew-installed DDEV v1.23.0-alpha1, which has this bug.

PR #6033, opened 2024-03-29
fix: shutdown supervisor in healthcheck.sh on fatal error, fixes #5993

### LINKED ISSUE #5993: ddev-webserver healthcheck should exit early on error with apache/nginx/php-fpm

```
```
</details>         

### Expected Behavior

The user should not wait two minutes for a timeout if something goes wrong with apache/nginx/php-fpm

### Actual Behavior

If the container fails to become healthy on `ddev start`, the supervisor tries to restart apache/nginx/php-fpm until the 2 minute timeout. Previously, this led to early failure, but now we have a timeout.

The logic changed after
- #5936

### Steps To Reproduce

See:

- https://github.com/ddev/ddev/issues/5899#issuecomment-2000105112

### Anything else?

The issue is partially discussed in:
- #5978

stasadev:
## The Issue

- #5993

I checked https://docs.docker.com/reference/dockerfile/#healthcheck before doing any change and when we use `start_period`, no matter what we do during this period, the container cannot become `unhealthy`.

> start period provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries.

I also searched for a fast fail when starting the container, but that's all I found:
https://forums.docker.com/t/healthcheck-fail-fast-feature-during-container-start-up/121201

So, in the end, I decided to do the same thing as suggested in:

- #5906

but in a simpler way.

Also I decreased `startretries=10` to `startretries=3` in supervisor, there is no reason to try more and wait longer, if the process cannot start several times already.

See error for php-fpm after 20-30 seconds (without waiting for 2 minutes default timeout):

```
echo 'RUN chmod 000 /run/php' > .ddev/web-build/Dockerfile
ddev restart
...
Failed waiting for web/db containers to become ready: web container failed: log=&{2024-03-29 21:42:41.347467771 +0200 EET 2024-03-29 21:42:42.086281673 +0200 EET 143 php-fpm:FATAL Shut down
}, err=ddev-d10-web container is unhealthy: &{2024-03-29 21:42:41.347467771 +0200 EET 2024-03-29 21:42:42.086281673 +0200 EET 143 php-fpm:FATAL Shut down
}, please use 'ddev logs -s web' and 'docker logs ddev-d10-web' and 'docker inspect --format "{{ json .State.Health }}" ddev-d10-web' to find out why it failed
```

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/apache.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/apache.conf
@@ -1,6 +1,5 @@
 [program:apache2]
 stopwaitsecs = 20
-startretries=10
 stopsignal = WINCH
 command=/usr/sbin/apache2ctl -D "FOREGROUND"
 # Great hints at https://advancedweb.hu/supervisor-with-docker-lessons-learned/
@@ -13,4 +12,4 @@ redirect_stderr=true
 exitcodes=0
 startsecs=1 # Must stay up 1 sec
 autorestart=true
-startretries=10
+startretries=3
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/gunicorn.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/gunicorn.conf
@@ -6,4 +6,4 @@ stdout_logfile=/var/tmp/logpipe
 stdout_logfile_maxbytes=0
 redirect_stderr=true
 autorestart=true
-startretries=10
+startretries=3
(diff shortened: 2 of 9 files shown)


rfay:
Just looking at the code says this is fine, will manually test, thanks!

stasadev:
Rebased, and pushed the image again.

rfay (approved):
This worked great on my first try. I set up https://github.com/ddev/test-django4-bakerydemo and didn't remember that [other things had to be done](https://github.com/ddev/ddev/blob/bdae26f0b85ef7545a5079f249d0b32ed27dcc9e/pkg/ddevapp/ddevapp_test.go#L299-L303). So it failed. 

But I was able to `ddev ssh` and study it, which is the good thing here. 

And after I did the things it was OK.

PR #6034, opened 2024-03-30
test: make circleci setup noninteractive so doesn't hang [skip buildkite]

rfay:
## The Issue

* Minor maintenance
* The CircleCI arm64 build has been failing waiting for interactive services restart:

image

Use DEBIAN_FRONTEND=noninteractive


rfay:
We should be considering stopping circleci usage. We get enough arm64 information via the macOS tests, and circleci costs at least $45/month the way we currently use it. And there are few linux arm64 users.

PR #6027, opened 2024-03-28
docs: add privacy plugins to not use external fonts, fixes #4982 

gitressa:
## The Issue

Fixes #4982.

We should keep an eye out for possibly slowed down automated builds, after this change.

## How This PR Solves The Issue

Bundles the fonts with the generated documents, as opposed to getting them from a third-party.

## Manual Testing Instructions

- **Now:**
    Open https://ddev--6027.org.readthedocs.build/en/6027/ in Firefox with uBlock Origin, and see that currently fonts are requested from fonts.googleapis.com
- **After adding the privacy plugin:**
    See that fonts are included and served from https://ddev--6027.org.readthedocs.build/en/6027/ 

Alternative method to verify that no third-parties are contacted:

1. Open developer tools in your browser with F12
1. Open "Network"
1. Reload the page, and see that no third-party resources are contacted, in this case Google and Plausible.

PR #6048, opened 2024-04-04
docs: add a sticky warning about the latest version

stasadev:
## The Issue

People do not know that they are using the documentation for an unreleased version of DDEV when they open the page and scroll down.

- I took the HTML code of the header from the upstream repo and added a warning banner to it.
- Add an anchor to the link leading to stable docs (`${window.location.hash}`)
- Add `/.cache` to `.gitignore` (follow up for #6027)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /authors.json
 /docs/cache
 /dist
+/.cache
 /.ddev
 /artifacts
 *.DS_Store
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -5,8 +5,50 @@
   <meta name="google-site-verification" content="uFpolYSfqt2_75ubmaMBdPrmc9LjnEkhZUhSF0ejXWg" />
 {% endblock %}
 
+<!-- This is a copy of https://github.com/squidfunk/mkdocs-material/blob/master/src/templates/partials/header.html with a warning banner added. -->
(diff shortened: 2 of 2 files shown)


docs/overrides/partials/header.html
```diff
@@ -0,0 +1,141 @@
+<!--
+  Copyright (c) 2016-2024 Martin Donath <martin.donath@squidfunk.com>
```
rfay, line 2: I assume this came from the mkdocs-material repo somewhere? Or is it a copy of something we already have included? Either way, please include a reference to the source. If it's a modified version of something from upstream, please indicate where the modification is.

rfay (changes requested):
It sure works great! I just want to know where the extra file comes from and how it can be maintained.

docs/overrides/partials/header.html
stasadev: Thanks, the file already contained a link to the upstream, but not at the top of the file.
It should be clearer now.

rfay:
> we force users to use stable docs anyway

What does that mean?

stasadev:
What I mean is that we recommend using the stable version of the documentation and giving stable links as references.

rfay:
Agreed, but it will be a very long time before Google starts suggesting only stable.

stasadev:
Please check https://ddev--6059.org.readthedocs.build/en/6059/users/quickstart/#drupal

- #6059

bmartinez287:
glad it worked.

PR #6050, opened 2024-04-04
fix: make sure CGO_ENABLED is normally off, fixes #6046

### LINKED ISSUE #6046: Unable to update / install ddev 1.23.0 alpha on Ubuntu 20.04

```
```
</details>         

### Expected Behavior

I should be able to install or update ddev. ( I had previously been using ddev, but needed to update it to support php 8.3

I tried manually installing the latest and using the .deb

### Actual Behavior

ddev --version
ddev: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ddev)
ddev: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ddev)

### Steps To Reproduce

I am on Ubuntu 20.04

### Anything else?

_No response_

rfay:
## The Issue

* #6046 

In DDEV v1.23.0-alpha1, the released binaries were released with CGO_ENABLED=1, which means they can't run everywhere they're intended to run. This includes Ubuntu 20.04

* Make sure we build it right
* Add cgo_enabled to `ddev version` so we can check it
* Update Makefile and github builds to test and make sure we have it right.

You can already test the artifacts built this way from the experimental release https://github.com/ddev-test/ddev/releases/tag/v1.22.10-aaa3 (Do not use this for anything but testing this issue). 

* `ddev version` should show cgo_enabled=0

## Automated Testing Overview

We now check in the build pipeline to make sure the CGO_ENABLED version is correct.

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -221,6 +221,9 @@ jobs:
         if: steps.notarizedmac.outputs.cache-hit != 'true'
         run: exit 1
 
+      - name: "Verify that cgo_enabled is 0 in Linux binary"
+        run: .gotmp/bin/linux_amd64/ddev version | grep "^cgo_enabled *0$"
+
       # Goreleaser does GitHub release artifacts, homebrew, AUR, deb/rpm
       - name: goreleaser
         uses: goreleaser/goreleaser-action@v5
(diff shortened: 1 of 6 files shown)


stasadev (approved):
I checked the artifacts from https://github.com/ddev-test/ddev/releases/tag/v1.22.10-aaa3 and from this PR, both show cgo_enabled 0.

Issue #6035, opened 2024-03-30
Improve support for the new `drupal`project type


rpkoller:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

While testing around with https://github.com/justafish/ddev-drupal-core-dev i've identified two scenarios where the newly added project type `drupal` might be improved further: 

1. If you take a look at the current recommended quickstart guides using the `drupal` project type: https://ddev.readthedocs.io/en/latest/users/quickstart/#drupal . you run `ddev config` on an empty directory there. the logics behind that command has no codebase at hand to determine the correct version of drupal that way. the codebase is pulled two steps after the first step on `ddev composer create`. for that reason ddev is falling back to the default values, php 8.2 and corepack enable false. but if you want to install a drupal 11 instance that requires PHP 8.3 and corepack enabled for example. 

2. The other case is if you have an existing project with the drupal10 project type and after upgrading from drupal 1.22.7 to 1.23.0 and you change the project type on `ddev config` from `drupal10` to `drupal` the settings remain the same. so in case you were on a drupal 11 instance before upgrading the newly added corepack_enable will stay set to false.

### Describe your solution

For point one. the config might be reevaluated and updated after `ddev composer create` is run and the codebase is in place. @rfay already ideated to introduce an `--update` flag to `ddev config` as well. i am not sure if i would make something that is altering the settings completely free accessibly (might entail unwanted sideeffects). but triggering such a functionality right after `ddev composer create` is run might be safe and fix the problem. but on the other hand a freely available command `ddev config --update` would also fix the second point. Not sure just wanted to open a discussion.

rfay:
I'd say:
- [ ] Add a `ddev config --update` or `ddev config --detect`
- [ ] Update quickstarts to add drupal9/drupal10/drupal11

rpkoller:
yeah probably the automatically run the config update in combination right after `ddev composer create` is run is probably overcomplicating things.  simply another manual step in the quickstart right after `ddev composer create` should be totally fine on second thought. and i like the idea with `ddev config --detect` even a bit better than `--update`. just one thought about that flag aside detecting the drupal version based on the codebase at hand and upgrading the settings would it make sense to output which settings are being altered and which are being added or removed?

rpkoller:
[...]

PR #6053, opened 2024-04-05
feat: POC for ddev config --update

penyaskito:

For supporting discussion on https://github.com/ddev/ddev/issues/6035.

If we end up going this path I'd create a cleaner PR.

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -128,7 +128,7 @@ var (
 	ddevVersionConstraint string
 
 	// update tries to update the configured php version with auto-detection.
-	update string
+	updateArg bool
 )
 
 var providerName = nodeps.ProviderDefault
@@ -312,7 +312,7 @@ func init() {
 	ConfigCommand.Flags().Bool("disable-upload-dirs-warning", true, `Disable warnings about upload-dirs not being set when using performance-mode=mutagen.`)
 	ConfigCommand.Flags().StringVar(&ddevVersionConstraint, "ddev-version-constraint", "", `Specify a ddev version constraint to validate ddev against.`)
 	ConfigCommand.Flags().Bool("corepack-enable", true, `Do 'corepack enable' to enable latest yarn/pnpm'`)
-	ConfigCommand.Flags().Bool("update", true, `Do 'update' to auto-detect the right settings`)
+	ConfigCommand.Flags().BoolVar(&updateArg, "update", false, `Do 'update' to auto-detect the right settings`)
 
 	RootCmd.AddCommand(ConfigCommand)
 
@@ -372,7 +372,7 @@ func handleMainConfigArgs(cmd *cobra.Command, _ []string, app *ddevapp.DdevApp) 
 		}
 	}
 
-	if cmd.Flag("update").Changed {
+	if updateArg {
 		friendlyMsg := fmt.Sprintf("Update flag detected")
 		output.UserOut.Print(friendlyMsg)
 
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/ddev/ddev/pkg/fileutil"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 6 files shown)


rfay:
WOW!

cmd/ddev/cmd/config.go
```diff
@@ -368,6 +373,63 @@ func handleMainConfigArgs(cmd *cobra.Command, _ []string, app *ddevapp.DdevApp)
...
+		output.UserOut.Print(detectedApptypeMsg)
+
+		// For the POC we need ConfigFileOverrideAction to omit app.ConfigExists() check
+		err = app.ConfigFileOverrideAction()
+
+		apptypeMsg := fmt.Sprintf("Apptype is %s", app.Type)
+		output.UserOut.Print(apptypeMsg)
+
+		if app.Type == nodeps.AppTypeDrupal {
+			// I can't call getDrupalVersion from drupal.go here?
```
penyaskito, line 391: TIL lowercase/uppercase defines visibility. So needs to be renamed to GetDrupalVersion if we want this function to be exported. https://go.dev/tour/basics/3

rfay: Your IDE can easily do this with a refactor action.

cmd/ddev/cmd/config.go
```diff
@@ -368,6 +373,63 @@ func handleMainConfigArgs(cmd *cobra.Command, _ []string, app *ddevapp.DdevApp)
...
+		detectedApptypeMsg := fmt.Sprintf("Detected %s", detectedApptype)
+		output.UserOut.Print(detectedApptypeMsg)
+
+		// For the POC we need ConfigFileOverrideAction to omit app.ConfigExists() check
+		err = app.ConfigFileOverrideAction()
+
+		apptypeMsg := fmt.Sprintf("Apptype is %s", app.Type)
+		output.UserOut.Print(apptypeMsg)
+
+		if app.Type == nodeps.AppTypeDrupal {
```
rfay, line 390: I don't think you should need to mention specific project types here.

rfay (changes requested):
The way I was thinking about this was that it would just be like any other config action (including --auto), except that it would not respect existing config. The key there is one line IIRC. Will take a look.

cmd/ddev/cmd/config.go
penyaskito: Yeah, it was only for debugging purposes when I did that. I just thought that we might want to take this chance to change apptype drupal8, drupal9, drupal10 to drupal if we find those.

penyaskito:
> Essentially we need to do what it was doing before that... but only on update?

Oh right I was accidentally hit by that one! Makes sense, it would be probably cleaner.

One question I have if we go either path is: should we really change db? That might be dangerous. Even if I don't think data loss could happen, people might not find their data if the db versions change. I suggest if a change in db is found, we leave it as is and provide instructions on updating/migrating it.

pkg/ddevapp/apptypes.go
```diff
@@ -388,9 +388,9 @@ func (app *DdevApp) PostImportDBAction() error {
 }
 
 // ConfigFileOverrideAction gives a chance for an apptype to override any element
-// of config.yaml that it needs to (on initial creation, but not after that)
-func (app *DdevApp) ConfigFileOverrideAction() error {
-	if appFuncs, ok := appTypeMatrix[app.Type]; ok && appFuncs.configOverrideAction != nil && !app.ConfigExists() {
+// of config.yaml that it needs to
+func (app *DdevApp) ConfigFileOverrideAction(overrideExistingConfig bool) error {
+	if appFuncs, ok := appTypeMatrix[app.Type]; ok && appFuncs.configOverrideAction != nil && (overrideExistingConfig || !app.ConfigExists()) {
```
penyaskito, line 393: @rfay Even if we go the other path, we probably still need this.

rfay:
> if we go either path is: should we really change db

We should not change database type or version. It would make people crazy. Thanks for paying attention to this. 

I see that this would affect CraftCMS, which really wants MySQL 8.0. But I don't see why we'd change existing.

I see drupal8/9 have override to mariadb:10.4, but I don't think those belong in there in the first place? Why would we not use defaults?

Silverstripe is probably a mistake. @GuySartorelli should we be overriding database type and version in https://github.com/ddev/ddev/blob/64c9e469427785debfdda7db2e8caf1b8632433c/pkg/ddevapp/silverstripe.go#L78-L79 ? For that matter, should we even have silverstripeConfigOverrideAction or should it be removed?

penyaskito:
@rfay I've created https://github.com/ddev/ddev/pull/6055. It's definitively cleaner that way. I'll leave this one open still in case we want to keep these conversations here, but feel free to close it if you feel so.

rfay:
Closing just to avoid confusion, can be reopened.

GuySartorelli:
I recognise this was closed but just answering what was asked:

> Silverstripe is probably a mistake. @GuySartorelli should we be overriding database type and version in? For that matter, should we even have silverstripeConfigOverrideAction or should it be removed?

I'm a bit removed from the workflow of someone who uses Silverstripe CMS to make websites - my workflow is very specific to the maintenance work I do so I'm not sure whether this would be a useful or harmful.

It is worth noting that there have been a few discussions about how DDEV overrides values in `.env` (e.g. database name, user, and password) which some people have found to disrupt their preferred workflow. I think that's a separate concern, but worth mentioning. I actually explicitly avoid using the "silverstripe" project type because of it.

PR #6058, opened 2024-04-05
fix: Update defaults for config overrides

rfay:
## The Issue

We're about to release a major version and we want to make sure that the config overrides match current requirements.

Links are in the code for the changes to Laravel, Silverstripe, Magento1/Openmage

Try out the quickstarts for the changed items.

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -369,7 +369,6 @@ func drupalConfigOverrideAction(app *DdevApp) error {
 		app.Database = DatabaseDesc{Type: nodeps.MariaDB, Version: nodeps.MariaDB104}
 	case "9":
 		app.PHPVersion = nodeps.PHP81
-		app.Database = DatabaseDesc{Type: nodeps.MariaDB, Version: nodeps.MariaDB104}
 	case "10":
 		app.PHPVersion = nodeps.PHP83
 	case "11":
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -151,9 +151,8 @@ func init() {
 		},
 
 		nodeps.AppTypeLaravel: {
-			appTypeDetect:        isLaravelApp,
-			postStartAction:      laravelPostStartAction,
-			configOverrideAction: laravelConfigOverrideAction,
+			appTypeDetect:   isLaravelApp,
+			postStartAction: laravelPostStartAction,
 		},
 
 		nodeps.AppTypeSilverstripe: {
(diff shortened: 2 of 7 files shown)


pkg/ddevapp/drupal.go
```diff
@@ -369,7 +369,6 @@ func drupalConfigOverrideAction(app *DdevApp) error {
 		app.Database = DatabaseDesc{Type: nodeps.MariaDB, Version: nodeps.MariaDB104}
 	case "9":
 		app.PHPVersion = nodeps.PHP81
-		app.Database = DatabaseDesc{Type: nodeps.MariaDB, Version: nodeps.MariaDB104}
```
rpkoller, line 372: good idea removing that one. with the current default 10.4 i ran into the following error testing with @penyaskitos artifact: 

```
$> ddev config --update
You are reconfiguring the project at /Users/rkoller/Sites/d9.
Configuring a 'drupal' project with docroot 'web' at /Users/rkoller/Sites/d9/web
unable to configure project d9 with database type mariadb:10.11 because that database type does not match the current actual database. Please change your database type back to mariadb:10.11 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.11', and you can try a migration with 'ddev debug migrate-database mariadb:10.4' see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/```

rfay: @penyaskito is going to make it not override existing database config for this exact reason.

stasadev (approved):
I tested
- Laravel
- Magento1/Openmage
- Silverstripe
- Drupal9

And checked the source code for all `nodeps.PHP*`, everything looks good.

pkg/ddevapp/drupal.go
rpkoller: yep. so removing 10.4 for d9 is a good call. just wanted to echo that.

rpkoller:
in regards of drupal, d8 requires php 7 which is covered with 7.4 and the problem for drupal 9 in regards of database is solved by removing mariadb 7.4. looks good

pkg/ddevapp/drupal.go
penyaskito: Is this because 10.4 is the global default?
If not, I'll keep this and just avoid overriding the DB in the other PR. But for new D9 sites might make sense.

rfay: No, mariadb:10.11 is the new default.  But D9 doesn't really care what version of mariadb it has.

PR #6059, opened 2024-04-05
docs: fix the scroll margin for links with anchors, for #6048

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/6048#issuecomment-2040194194

Links with anchors open in the wrong position:

Stable:
https://ddev.readthedocs.io/en/stable/users/quickstart/#drupal

Latest:
https://ddev.readthedocs.io/en/latest/users/quickstart/#drupal

The code that calculates this:
https://github.com/squidfunk/mkdocs-material/blob/68126abb567ed9f9c5940af5ba31c6005c499596/src/templates/assets/stylesheets/main/extensions/markdown/_toc.scss#L61-L71

Ideally we would need to add another `+ 48px` here for a warning banner:

```scss
--md-scroll-margin: #{px2rem(48px + 24px)};
```

Uses a fixed value of `6rem` for the margin. This is not ideal, as it may not work for all `font-size`, but at least it should satisfy most users.

Open on different devices https://ddev--6059.org.readthedocs.build/en/6059/users/quickstart/#drupal

## Related Issue Link(s)

- https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors

### DIFF
--- a/docs/overrides/partials/header.html
+++ b/docs/overrides/partials/header.html
@@ -34,10 +34,17 @@
 
   <!-- Warning banner -->
   {% if config.extra.version != "stable" %}
-    <!-- remove the shadow, as it looks strange with a warning banner next to it -->
     <style>
+      /* remove the shadow, as it looks strange with a warning banner next to it */
       .md-header--shadow {
         box-shadow: unset;
+      }
+      /* fix the scroll margin for links with anchors */
+      /* see https://github.com/squidfunk/mkdocs-material/blob/master/src/templates/assets/stylesheets/main/extensions/markdown/_toc.scss */
+      .md-typeset {
+        :target {
+          --md-scroll-margin: 6rem;
+        }
       }
     </style>
     <div data-md-color-scheme="default" data-md-component="outdated">


GuySartorelli:
Looks good on my tablet. 
I don't understand the way this is separated from the "stable" docs well enough - will this affect the stable docs at all or is it only ever going to affect latest?

rfay:
Good question @GuySartorelli - I think this should affect only latest, since it's all about getting people to move most of the time from "latest" to "stable"

GuySartorelli:
I agree that it *should* only affect latest from an acceptance criteria pov.

My question is more about whether it could accidentally cause a change to the "stable" docs. I'm guessing not but seemed worth asking just in case.

stasadev:
> I don't understand the way this is separated from the "stable" docs well enough - will this affect the stable docs at all or is it only ever going to affect latest?

Look for `{% if config.extra.version != "stable" %}` in the code.

`export READTHEDOCS_VERSION_NAME=latest` can be used locally to test the warning banner.

How it was added:

- #5855

PR #6060, opened 2024-04-07
docs: Add drupal9/11 tabs and symfony to quickstart

rfay:
## The Issue

* Add drupal9/11 tabs to quickstarts
* Add symfony

Rendered versions at:
* https://ddev--6060.org.readthedocs.build/en/6060/users/quickstart/#drupal
* https://ddev--6060.org.readthedocs.build/en/6060/users/quickstart/#symfony

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -627,6 +627,17 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like 
     ddev launch
     ```
 
+=== "Git Clone"
+
+    ```bash
+    git clone <your-symfony-repo>
+    cd <your-symfony-repo>
+    ddev config --docroot=public --php-version=8.3
+    ddev start
+    ddev composer install
+    ddev launch
+    ```
+
 ## TYPO3
 
 === "Composer"
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -618,6 +618,8 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like 
 
 There are many ways to install Symfony, here are a few of them based on the [Symfony docs](https://symfony.com/doc/current/setup.html).
 
+If your project uses a database you'll want to set the [DB connection string](https://symfony.com/doc/current/doctrine.html#configuring-the-database) in the `.env`. If using the default MariaDB configuration, you'll want `DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=10.11"`. If you're using a different database type or version, see `ddev describe` for the type and version.
+
 === "Composer"
 
     ```bash
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -635,18 +635,11 @@ If your project uses a database you'll want to set the [DB connection string](ht
 
     In a future release the Symfony CLI will be provided by default in `ddev-webserver`, but for now it needs to be configured.
 
-    * Create a `.ddev/web-build/Dockerfile.symfony-cli` with the contents:
-
-    ```bash
-    RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | sudo -E bash
-    RUN sudo apt install -y symfony-cli
-    ```
-
-    Then:
-
     ```bash
     mkdir my-symfony && cd my-symfony
     ddev config --docroot=public
+    echo "RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | sudo -E bash
+    RUN sudo apt install -y symfony-cli" >.ddev/web-build/Dockerfile.symfony-cli
     ddev restart
     ddev exec symfony check:requirements
     ddev exec symfony new temp --version="7.0.*" --webapp
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -162,6 +162,7 @@ ddev launch
 ## Drupal
 
 === "Drupal"
+
     For all versions of Drupal 8+ the Composer techniques work. The settings configuration is done differently for each Drupal version, but the project type is "drupal".
 
     ```bash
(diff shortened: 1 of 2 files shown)


docs/content/users/quickstart.md
```diff
@@ -162,6 +162,7 @@ ddev launch
 ## Drupal
 
 === "Drupal"
```
rpkoller, line 164: It reads odd to have at first a tab called drupal and then tabs with specific version numbering and technically the order is 10, 11, 9, 6/7... plus it feels odd to have the explanation inside the box on just one out of the three tabs this explanation applies to. I wonder if the following would make sense. 

move ```For all versions of Drupal 8+ the Composer techniques work. The settings configuration is done differently for each Drupal version, but the project type is “drupal”.``` outside the box, between the title Drupal and the box with the drupal tabs. and in regards of the tab naming and ordering, how about switching the position between drupal 10 and 11 and renaming `drupal` to `Drupal 10` and `Drupal 11` to `Drupal 11 (dev)` (or something similar)?

docs/content/users/quickstart.md
```diff
@@ -587,6 +614,30 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+=== "Git Clone"
+
+    ```bash
+    git clone <your-symfony-repo>
+    cd <your-symfony-repo>
+    ddev config --docroot=public --php-version=8.3
+    ddev start
+    ddev composer install
+    ddev launch
+    ```
```
penyaskito, line 639: I'd suggest a couple of links with more information until we can have a symfony project-type 

```suggestion
    ddev launch
    ```
As Symfony projects can be so flexible, you might need different customizations for your project.

You might want to start by editing your `.env` file for setting the DB connection string if you require one. See the [official docs](https://symfony.com/doc/current/doctrine.html#configuring-the-database) for more information.
If you require your current database engine version, you can use `ddev -s db exec --raw mysql -udb -pdb db -e 'SELECT VERSION();'`

Read more about customizing the environment with different [Config Options](https://ddev.readthedocs.io/en/latest/users/configuration/config/) and [Defining Additional Services with Docker Compose](https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/).    
    
```

rpkoller (commented):
and the quickstarts for drupal we should revisit in a follow up issue  when the `ddev config --update`PR is finalized imho

docs/content/users/quickstart.md
rfay: It's hard to follow your request. Maybe you can do a mockup, or just edit as you see fit and show what happens with `make mkdocs-serve`

docs/content/users/quickstart.md
rfay: I guess you don't know about `ddev mysql`. You'll like it! Please edit to use it. 

But... I don't understand what "if you require your current database engine version" means. Wouldn't they just `ddev describe` ?

The links should be relative links (in docs) instead of explicit links. Can you edit?

penyaskito: Updated those.

The version suggestion is because by default doctrine uses a DSN, and looks like it's a bit picky about those with mariadb/mysql according to the official Doctrine DBAL docs, so you might want to include the version in the DSN connection string.

https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/configuration.html#automatic-platform-version-detection

docs/content/users/quickstart.md
rpkoller: Screenshot 2024-04-07 at 03 01 57

I meant to move the description out of the drupal tab and perhaps slightly adjust it to something like: 

```
For all versions of Drupal 8+ the installation is done via Composer. The settings configuration is different for every version of Drupal, but the project type remains just "drupal".
```
and changing the tab labels  to something like:  

Drupal 11 (dev), Drupal 10, Drupal 9, Drupal 6/7,  Git Clone

docs/content/users/quickstart.md
rfay: Just `ddev mysql  -e 'SELECT VERSION();'` :)

But not sure how it's better than `ddev describe`

penyaskito: It gives the exact major.minor.patch version. Honestly, don't know how relevant that is, but most of examples in those docs specify at that level.

docs/content/users/quickstart.md
```diff
@@ -175,16 +176,36 @@ ddev launch
...
+    ddev start
+    ddev composer create drupal/recommended-project:^9
+    ddev composer require drush/drush
+    ddev drush site:install --account-name=admin --account-pass=admin -y
+    # use the one-time link (CTRL/CMD + Click) from the command below to edit your admin account details.
+    ddev drush uli
+    ddev launch
     ```
 
 === "Drupal 6/7"
```
rpkoller, line 211:
```suggestion
=== "Drupal 6 (EOL)/7"
```
A real good idea adding EOL to Drupal 9! would it make sense to clarify that for drupal 6 as well?

docs/content/users/quickstart.md
rfay: I added a commit with your suggestion and minor changes.

docs/content/users/quickstart.md
rfay: It's now like this: 
image

docs/content/users/quickstart.md
rfay: IMO they're so long gone that people should know this themselves, and will be using them deliberately if needed.

docs/content/users/quickstart.md
```diff
@@ -449,23 +470,27 @@ The Laravel project type can be used for [Lumen](https://lumen.laravel.com/) lik
...
-!!!tip
-    Moodle relies on a periodic cron job—don’t forget to set that up! See [ddev/ddev-cron](https://github.com/ddev/ddev-cron).
+    Visit the [Moodle Admin Quick Guide](https://docs.moodle.org/400/en/Admin_quick_guide) for more information.
+
+    !!!tip
+        Moodle relies on a periodic cron job—don’t forget to set that up! See [ddev/ddev-cron](https://github.com/ddev/ddev-cron).
 
 ## Pimcore
 
+=== "Composer"
```
stasadev, line 492: Formatting is broken, please add tabs for the text below.

docs/content/users/quickstart.md
```diff
@@ -587,6 +614,57 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+    ```bash
+    RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | sudo -E bash
+    RUN sudo apt install -y symfony-cli
+    ```
+
+    Then:
+
+    ```bash
+    mkdir my-symfony && cd my-symfony
+    ddev config --docroot=public
```
stasadev, line 649: We can use `echo` to add the file, as in other quickstarts:

````suggestion
    ```bash
    mkdir my-symfony && cd my-symfony
    ddev config --docroot=public
    echo "RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | sudo -E bash
    RUN sudo apt install -y symfony-cli" >.ddev/web-build/Dockerfile.symfony-cli
````

**Review by stasadev — CHANGES_REQUESTED** [review]
Tested Symfony quickstart. Works great 👍

I only have suggestions for formatting.

**Comment by sanjaybip** [comment]
Looks good to me.

File: docs/content/users/quickstart.md (hunk 5 above)
Comment (sanjaybip) [reply docs/content/users/quickstart.md:649]: Why this? Is  this better?

Comment (stasadev) [reply docs/content/users/quickstart.md:649]: Because the order of actions is wrong:

The example starts by creating `.ddev/web-build/Dockerfile.symfony-cli` and then you create a project folder
```
mkdir my-symfony && cd my-symfony
ddev config --docroot=public
```

sanjaybip: Make sense.

rfay:
I think I got everything @stasadev , thanks!

rfay:
Thanks all, I pulled so I could build on these docs in
* #6055

PR #6066, opened 2024-04-08
build: bump docker-compose to v2.26.1

stasadev:
## The Issue

docker-compose v2.26.1 contains a few bugfixes that may not be important, but are still worth a try.

https://github.com/docker/compose/releases/tag/v2.26.1


rfay:
Oh, I think
* https://github.com/docker/compose/pull/11555

might be a good thing!

PR #6074, opened 2024-04-09
test: Remove docker login from buildkite tests

rfay:
## The Issue

We have perpetual problems with the `docker login` on Docker Desktop for windows, 

`:docker: Logging into docker registry
Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded`

This can be demonstrated on the command line.

https://buildkite.com/ddev/wsl2-docker-desktop/builds/5372#018ec23f-52ae-4ba9-beb3-826af374d24c

The whole business of logging on buildkite tests was to prevent rate-limiting by Dockerhub. But in the years since it was introduced, they put us on the DSOS program, where rate-limiting isn't a problem any more.

Remove the docker login from buildkite tests.

### DIFF
--- a/.buildkite/macos-colima.yml
+++ b/.buildkite/macos-colima.yml
@@ -3,10 +3,6 @@
 # Runs on master and PRs, including forked PRs
 
   - command: ".buildkite/test.sh"
-    plugins:
-      - docker-login#v2.1.0:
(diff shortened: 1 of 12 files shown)


stasadev (approved):
Looks good 👍

PR #6080, opened 2024-04-11
feat: Add symfony cli

rfay:
## The Issue

In
* https://github.com/ddev/ddev/pull/6060

we added the symfony quickstart. But we had the awkward duty of adding it with a Dockerfile.

If we're going to push another ddev-webserver:v1.23.0-1... why not get this in?

### DIFF
--- a/.spellcheckwordlist.txt
+++ b/.spellcheckwordlist.txt
@@ -138,6 +138,7 @@ SQLite
 Starts
 Statamic
 Support
+Symfony
 TablePlus
 Traefik
 TravisCI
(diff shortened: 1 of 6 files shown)


GuySartorelli (approved):
Tested using the modified instructions in this PR - works great locally and in gitpod.

It would be nice to have a `ddev symfony` shortcut for `ddev exec symfony` but I appreciate that's probably out of scope for this PR.

rfay:
Yes, I thought about `ddev symfony` as well, but since we don't have a symfony project type to turn it on or off, it would be there for everybody. Future!

stasadev (approved):
I'm okay with adding Symfony CLI, tested and it works fine.

P.S. I want the same thing for Laravel in DDEV, but unfortunately their installer is based on composer only.

PR #6081, opened 2024-04-11
fix: Skip hidden directories when creating commands

GuySartorelli:
## The Issue

DDEV thinks hidden directories in `~/.ddev/commands/` hold commands for services with names that start with a `.`

This can result in commands like this showing up:

Screenshot from 2024-04-11 17-53-05

Note that the `composer.lock`, `composer.json`, and `config.yml` files all live in a hidden directory.

(For reference, `.php-utils/` here contains code and dependencies that my global host commands share - see https://github.com/GuySartorelli/my-ddev-commands if you're curious)

Skips hidden directories, so that files in those directories are not treated like commands.

Add a hidden directory in `~/.ddev/commands/` with some files in it and run `ddev` - before the PR they will show up in the commands list, but afterwards they shouldn't.

## Automated Testing Overview

Tests that a command in a hidden directory doesn't show in the commands list

## Related Issue Link(s)
N/A

## Release/Deployment Notes

In the very unlikely scenario where someone does actually have a service where the name of the services starts with a `.`, they will no longer be able to add global commands for it. That strikes me as exceedingly unlikely though.

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -82,6 +82,10 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 			if !fileutil.IsDirectory(serviceDirOnHost) {
 				continue
 			}
+			// Skip hidden directories as well.
+			if strings.HasPrefix(filepath.Base(serviceDirOnHost), ".") {
+				continue
+			}
 			commandFiles, err := fileutil.ListFilesInDir(serviceDirOnHost)
 			if err != nil {
 				return err
--- a/cmd/ddev/cmd/commands_test.go
+++ b/cmd/ddev/cmd/commands_test.go
@@ -89,6 +89,7 @@ func TestCustomCommands(t *testing.T) {
 	assert.NotContains(out, "testwebglobal global (global shell web container command)")
 	assert.NotContains(out, "testhostcmd global")
 	assert.NotContains(out, "testwebcmd global")
+	assert.NotContains(out, "not-a-command")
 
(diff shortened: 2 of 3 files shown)


GuySartorelli:
If this is at all controversial I can live with moving my directory `.phputils` directory into `commands/host/.phputils` instead since that will be ignored - but this seems like something someone else will eventually run into as well so I figured I'd offer up this solution first.

GuySartorelli:
> So you use directories named .something in your ~/.ddev/commands directory? What are they for?

It's right in the PR description. :p

> (For reference, .php-utils/ here contains code and dependencies that my global host commands share - see https://github.com/GuySartorelli/my-ddev-commands if you're curious)

GuySartorelli:
I don't think we did get stuck - I mentioned an approach in that PR which wasn't accepted. `xhprof` and `xdebug` for example have a function which is defined elsewhere. But the implementors decided to go with a different approach.
For any bash commands provided by DDEV we can do the same thing to provide any manner of utility scripts that can be shared across commands.

For custom commands outside the DDEV codebase that obviously won't be as helpful - maybe it's worth opening an issue about that.

stasadev (approved):
Tested, it looks good to me 👍

rfay (approved):
Thanks for this! There's only one problem that I see... changing the code for exactly one dev's workflow :) But that one person is a valued contributor, so thanks! And thanks for checking the test coverage while you were there.

GuySartorelli:
> changing the code for exactly one dev's workflow

One developer's workflow *today* ;p Fixing small issues like this can reduce friction for others to do similar things if they want to.

PR #6082, opened 2024-04-11
feat: Add toggle option for xhprof command, rework xdebug, fixes #5782

### LINKED ISSUE #5782: xhprof command should have a `toggle` argument

Not a problem per se - but it can be a bit annoying to get the right command when you're toggling something like this on and off multiple times.

### Describe your solution

xdebug was given a toggle argument in https://github.com/ddev/ddev/pull/5625
It should be trivial to add the same thing for xhprof as well.

### Describe alternatives

_No response_

### Additional context

_No response_

GuySartorelli:
## The Issue

The `xdebug` command has a toggle option, but `xhprof` doesn't. They're both really similar in their operation so it makes sense to add a `ddev xhprof toggle` option.

Adds the option.

Also some unrelated-but-adjacent refactoring of xdebug, which was pretty convoluted.

- Try `ddev xhprof toggle` and validate the status with `ddev xhprof status`
- Check that hitting `tab` after typing `ddev xhprof ` gives you `toggle` amongst the other options.
- Try `ddev xdebug toggle` and `ddev xdebug status` as a sanity check for the refactor

## Automated Testing Overview

Existing tests should cover the xdebug refactoring.

There's no tests at all for xhprof that I could find, so adding them seems out of scope. Happy to add if maintainers disagree with that assessment though.

## Related Issue Link(s)

- #5782 

## Release/Deployment Notes

N/A

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/xhprof
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/xhprof
@@ -2,11 +2,11 @@
 
 ## #ddev-generated
 ## Description: Enable or disable xhprof
-## Usage: xhprof on|off|enable|disable|true|false|status
-## Example: "ddev xhprof" (default is "on"), "ddev xhprof off", "ddev xhprof on", "ddev xhprof status"
+## Usage: xhprof on|off|enable|disable|true|false|toggle|status
+## Example: "ddev xhprof" (default is "on"), "ddev xhprof off", "ddev xhprof on", "ddev xhprof toggle", "ddev xhprof status"
 ## ExecRaw: false
 ## Flags: []
-## AutocompleteTerms: ["on","off","enable","disable","status"]
+## AutocompleteTerms: ["on","off","enable","disable","toggle","status"]
 
 if [ $# -eq 0 ]; then
   enable_xhprof
@@ -19,6 +19,14 @@ on | true | enable)
   ;;
 off | false | disable)
   disable_xhprof
+  ;;
+toggle)
+  status=$(php -m | grep 'xhprof')
+  if [ "${status}" = "xhprof" ]; then
+    disable_xhprof
+  else
+    enable_xhprof
+  fi
   ;;
 status)
   status=$(php -m | grep 'xhprof')
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/xdebug
@@ -15,6 +15,25 @@ fi
 
 xdebug_version=$(php --version | awk '/Xdebug v/ {print $3}')
 
+get_xdebug_status() {
+    case ${xdebug_version} in
+    v3*)
+      status=$(php -r 'echo ini_get("xdebug.mode");' 2>/dev/null)
+      if [[ "${status}" =~ .*"debug".* ]]; then
+        echo "1"
+      else
+        echo "0"
+      fi
+      ;;
+    v2*)
+      echo $(php -r 'echo ini_get("xdebug.remote_enable");')
+      ;;
+    *)
+      echo "0"
+      ;;
+    esac
+} 
+
 case $1 in
   on|true|enable)
     enable_xdebug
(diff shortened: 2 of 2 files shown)


rfay:
See https://github.com/ddev/ddev/blob/b0fb8cea16ee67564a0cf02991d61a59f3d0e4ef/pkg/ddevapp/ddevapp_test.go#L1097-L1209 for TestDdevXhprofEnabled.

Your IDE should have been able to find this! :)

rfay:
Does this mean that you often use xhprof?

GuySartorelli:
> Does this mean that you often use xhprof?

I almost never use it, but if there's anyone using DDEV who does a lot of profiling, I could see this being helpful - while at the same time it adds next to no additional maintenance.

GuySartorelli:
> Your IDE should have been able to find this! :)

If I had done a blanket search for `xhprof` it would have - I was just expecting a separate file like `xdebug` has since the commands are basically identical.

Looking at that test, it's testing the underlying `enable_xhprof` and `disable_xhprof` functions, not the `ddev xhprof` command itself. Unless I missed something.

rfay:
The xhprof command just runs enable_xhprof, etc.

I found the test by using symbol search and typing TestXhprof and it brought it right up (Goland)

GuySartorelli:
> The xhprof command just runs enable_xhprof, etc.

By that argument there's no need for me to update any tests, because that's still what it's doing. That doesn't test whether the existing `ddev xhprof status` command outputs correctly or whether this new `toggle` option works as expected - but as I say, that's a completely different test than the one that currently exists.

GuySartorelli:
I'm happy to add testing coverage for the actual `ddev xhprof` command itself if you want - just noting that that's not what the current test covers, which is why I didn't add tests for the toggle option there.

rfay:
It's all good, no need to go further IMO. Maybe when we integrate xhprof more thoroughly later this year with integrating the ddev-xhgui behavior.

stasadev (approved):
Tested, looks good to me 👍

P.S. Nice refactoring for the `xdebug` command, thanks!

rfay:
In testing this PR I did `ddev xdebug toggle` a number of times. When xdebug was not enabled, it did the right thing. But often when xdebug *was* enabled, it would hang. The hang is in `php -r 'echo ini_get("xdebug.mode");'`. 

I was testing on macOS, and where this happened I was on php 8.2

* It has nothing to do with this PR. I tested with HEAD and saw the same.
* It doesn't matter whether mutagen is enabled or not

GuySartorelli:
Ohhh, interesting. I guess that's 'cause in order to get the xdebug status, we're running php while xdebug is enabled - and xdebug is trying to connect to an IDE or anything that'll listen, and not finding anything to connect to?

Probably we should use `php -i`, or check the INI file(s) directly, or something else that doesn't require executing code via PHP to get the status.

GuySartorelli:
`php --ri xdebug` will exit with a 0 status if xdebug is enabled, and a non-0 status if it's disabled, so that's another option.

rfay:
You're right, congratulations! This doesn't happen if PhpStorm IDE isn't listening!

rfay:
I think that `php --version`, which is widely used in tests and such is probably an extremely robust technique. `php --version | grep xdebug` for example.

rfay:
Followups in 
* https://github.com/ddev/ddev/issues/6191

pkg/ddevapp/global_dotddev_assets/commands/web/xhprof
```diff
@@ -20,6 +20,14 @@ on | true | enable)
 off | false | disable)
   disable_xhprof
   ;;
+toggle)
+  status=$(php -m | grep 'xhprof')
+  if [ "${status}" = "xhprof" ]; then
```
rfay, line 25: This could be simpler with 
```
if php -m | grep xhprof >/dev/null; then
```
but it's fine the way it is.

rfay (approved):
This is fine, thanks so much, and sorry for the long wait. 

I've asked our local xhprof expert, @tyler36, to take a quick look, but this can be pulled in the morning if he either has no objections or can't get to it.

tyler36:
`XHprof` works as expected. No conflicts with `tyler36/ddev-xhgui`.
`xdebug` also continues to function after "rework".

Seems good to me.

## Test

1. Confirm DDEV version
```shell
$ ddev -v                                                          
ddev version v1.23.0-beta1-23-gbb51e6fc8
```

2. Start a new Drupal 10 project following quick start guide.
3. Toggle xhprof

```shell
$ ddev xhprof toggle
Enabled xhprof.
After each web request or CLI process you can see all runs, most recent first, at
https://xhprof-test.ddev.site/xhprof
```

4. Visit https://xhprof-test.ddev.site/xhprof to confirm working.
5. Refresh page to confirm XHprof is generating content


## Test with tyler36/ddev-xhgui

1. Install tyler36/ddev-xhgui addon

```shell
ddev get tyler36/ddev-xhgui
ddev restart
```

2. Following Drupal instructions, add `perftools/php-profiler`

```shell
ddev composer require perftools/php-profiler --dev
```

3. Following Drupal instructions, update `web/sites/default/settings.php`

```php
if (file_exists("/mnt/ddev_config/xhgui/collector/xhgui.collector.php")) {
 require_once "/mnt/ddev_config/xhgui/collector/xhgui.collector.php";
}
```

4. Toggle xhprof

```shell
ddev xhprof toggle
```

5. Launch XHGui page

```shell
ddev xhgui
```

6. Visit main Drupal page and confirm reports are generated in XHGui.

## Test Xdebug still functions

1. Add breakpoint to `web/index.php` on lin 16: `$kernel = new DrupalKernel('prod', $autoloader);`
2. Start xdebug listening in Vscode
3. Toggle xdebug on

```shell
ddev xdebug on
```

4. Attempt to homepage; page does NOT finish loading as breakpoint is hit (expected behavior).

PR #6089, opened 2024-04-12
fix: don't label transient containers and run without healthcheck, fixes #6085

### LINKED ISSUE #6085: RunSimpleContainer can show inappropriate healthcheck information
I just did a `ddev start` on a Drupal 9 test project while testing for release (WSL2/Ubuntu24.04/DDEV v1.23.0-rc1) and saw this:
```
Failed to start d9: container(s) failed to become healthy before their configured timeout or in 120 seconds. This might be a problem with the healthcheck and not a functional problem. (start-chown-ISqDmm container exited, more info with [
        docker logs start-chown-ISqDmm
        docker inspect --format "{{ json .State.Health }}" start-chown-ISqDmm | docker run -i --rm ddev/ddev-utilities jq -r
])
```

This is apparently from https://github.com/ddev/ddev/blob/bd73ada6677e358e6e4cc6bc76c681c11bb59b03/pkg/ddevapp/ddevapp.go#L1176

* It's not useful output
* RunSimpleContainer shouldn't be waiting for healthcheck in the first place should it?

Healthcheck output was:
```
{
  "Status": "unhealthy",
  "FailingStreak": 0,
  "Log": []
}
```

### Actual Behavior

x

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #6085 

* RunSimpleContainer() should use no healthcheck
* Don't attach full label to these containers, so they're not detected in waiting for container start

See #6085

### DIFF
--- a/cmd/ddev/cmd/debug-dockercheck.go
+++ b/cmd/ddev/cmd/debug-dockercheck.go
@@ -73,14 +73,14 @@ var DebugDockercheckCmd = &cobra.Command{
 		}
 
 		uid, _, _ := util.GetContainerUIDGid()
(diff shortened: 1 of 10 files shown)


stasadev (approved):
Thank you, I tested it and now everything works great, I really like that we don't need to introduce a new label for these temporary containers.

PR #6092, opened 2024-04-13
docs: Updated Laravel Quickstart to use PHP Version 8.3 by default

genericrohan:

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -388,7 +388,7 @@ The Laravel project type can be used for [Lumen](https://lumen.laravel.com/) lik
     ```bash
     mkdir my-laravel-app
     cd my-laravel-app
-    ddev config --project-type=laravel --docroot=public --php-version=8.2
+    ddev config --project-type=laravel --docroot=public --php-version=8.3
     ddev composer create --prefer-dist laravel/laravel:^11
     ddev launch
     ```
@@ -398,7 +398,7 @@ The Laravel project type can be used for [Lumen](https://lumen.laravel.com/) lik
     ```bash
     git clone <your-laravel-repo>
     cd <your-laravel-project>
-    ddev config --project-type=laravel --docroot=public --php-version=8.2
+    ddev config --project-type=laravel --docroot=public --php-version=8.3
     ddev start
     ddev composer install
     ddev php artisan key:generate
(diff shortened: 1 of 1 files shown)


tyler36:
Is there are particular reason for PHP `8.3`?

tyler36:
The [Laravel docs](https://laravel.com/docs/11.x/deployment#server-requirements) for version `11` (latest/current) lists server requirements as:

> - PHP >= 8.2

The "[latest](https://ddev.readthedocs.io/en/latest/users/quickstart/#laravel)" version of the Lavavel Quickstart guide has:

> ddev config --project-type=laravel --docroot=public --php-version=8.2

End-of-list list[ PHP 8.2](https://endoflife.date/php) end-of-life as:

> - Active support: 08 Dec 2024
> - Security support: 08 Dec 2025

I think while active support is still available, the guide is best keep to minimum requirement.

stasadev:
I agree with @tyler36.

I don't have anything against 8.3, but I also tend to have the minimum required version in DDEV.

rfay:
There are two issues here IMO:

1. What does the framework require?
2. How can we make the docs as robust as possible? (and last the longest)

We haven't really had "policies" about what's best for the quickstarts, probably we need to think them through.

In this case we're best here without specifying the php version at all IMO, so it gets the default, which is 8.2, but next fall 8.4 (or 9.0?) will be out and 8.3 will become the default, and we'll bump the default, and it will most likely match what's needed by Laravel.

So in general I'd prefer to omit the `--php-version` if the current DDEV default is appropriate.

I'll leave this open for discussion, but think we should be **removing** php-version rather than changing it.

stasadev:
> So in general I'd prefer to omit the `--php-version` if the current DDEV default is appropriate.
> I'll leave this open for discussion, but think we should be **removing** php-version rather than changing it.

I remembered why I didn't touch the `--php-version` when I updated the quickstart to Laravel 11 - I just thought that if it was added before, there was no reason to remove it. Now I see that it was only added as an override, and was only relevant at that time.

@knightofrohan, I appreciate your efforts in creating this PR, but we don't need to override the default PHP settings at this point. I think it's still worth to continue here, you could change the PR name and remove all the default `--php-version=8.2` from the quickstarts.

rfay:
This has been fixed in 
* https://github.com/ddev/ddev/pull/6095

Closing this one, happy to continue the conversation. It normally turns out better to use the DDEV default PHP version if that's a match for what the framework needs, and in the case of Laravel, that works.

PR #6094, opened 2024-04-14
fix: use ip route show default to get WSL2 addr of windows

rfay:
## The Issue

In testing the quickstarts, we found that Kirby didn't work. I was working on a WSL2 machine. I couldn't make Xdebug work. 

It does seem that the instructions in https://learn.microsoft.com/en-us/windows/wsl/networking#accessing-windows-networking-apps-from-linux-host-ip have changed from recommending looking at /etc/resolv.conf, which is what we've been doing for some time. Anyway, there was no way I could use that IP address.

Use `ip -4 route show default` and parse the result to get the Windows address

Try using xdebug on WSL2.


pkg/dockerutil/dockerutils.go
```diff
@@ -1305,31 +1306,59 @@ func GetNFSServerAddr() (string, error) {
...
+
+	if err != nil {
+		util.Warning("Unable to run 'ip -4 route show default' to get Windows IP address")
+		return ""
 	}
-	util.Warning("inappropriately using wsl2ResolvConfNameserver() but not on WSL2")
-	return ""
+	parts := strings.Split(out, " ")
+	if len(parts) < 3 {
+		util.Warning("Unable to parse output of ip -4 route show default, result was %v", parts)
```
stasadev, line 1350:
```suggestion
		util.Warning("Unable to parse output of 'ip -4 route show default', result was %v", parts)
```

pkg/dockerutil/dockerutils.go
```diff
@@ -1305,31 +1306,59 @@ func GetNFSServerAddr() (string, error) {
...
+		return ""
+	}
+
+	ip := parts[2]
+
+	if parsedIP := net.ParseIP(ip); parsedIP == nil {
+		util.Warning("Unable to validate IP address '%s' from ip -4 route show default", ip)
```
stasadev, line 1357:
```suggestion
		util.Warning("Unable to validate IP address '%s' from 'ip -4 route show default'", ip)
```

stasadev (approved):
It looks great to me, I suggest adding single quotes to the warning output for the command `ip -4 route show default`.

PR #6095, opened 2024-04-14
docs: Updates to the quickstarts

rpkoller:
## The Issue
a few changes based on the latest testing

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -275,7 +275,7 @@ For all versions of Drupal 8+ the Composer techniques work. The settings configu
     ddev config --php-version=8.2 --omit-containers=db
     ddev start
     ddev composer create getgrav/grav
-    ddev exec bin/gpm install admin -y
+    ddev exec gpm install admin -y
     ddev launch
     ```
 
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -306,7 +306,7 @@ Install [Ibexa DXP](https://www.ibexa.co) OSS Edition.
 
 ```bash
 mkdir my-ibexa-site && cd my-ibexa-site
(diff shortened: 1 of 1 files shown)


rpkoller:
but in follow ups pick up on the effort in trying to make the quickstarts more consistent in between CMSes. At the moment some are extremely verbose while others are quite brief, as well as @rfay suggested to standardized the used patterns like how directories are created, how login credentials are specified and so forth. that users are able to recognize the patterns for wayfinding if they go off trail to another CMS

rpkoller:
> * I would prefer `mkdir && cd xxx` for each of these. It just makes them simpler to test.

completely agree, also had that on my radar but thought it would be out of the scope for this issue and should go in a followup. but i can already do it now. the only question. should it be `my_xxxxxxx_project` or `my_xxxxxx_site`?

> * Magento2: Let's break up the setup:install line with backslashes and line breaks:

will do

rfay (changes requested):
This is looking great, and thanks for curating these.

I was hoping for all of the 
```
mkdir my-cakephp-site
cd my-cakephp-site
```

to be changed to `mkdir my-cakephp-site && cd my-cakephp-site`

rpkoller:
> This is looking great, and thanks for curating these.
> 
> I was hoping for all of the
> 
> ```
> mkdir my-cakephp-site
> cd my-cakephp-site
> ```
> 
> to be changed to `mkdir my-cakephp-site && cd my-cakephp-site`

gosh should have read things more thoroughly in the first place. i was fixated onto the naming pattern and totally glossed over the `mkdir && cd xxx` part ... updated. but i left the the quicktarts with git.

rfay (approved):
Looks great to me, should be ready to go, thanks!

rfay:
Thanks!

stasadev:
I noticed a few more cosmetic changes, created:

- #6107

PR #6101, opened 2024-04-15
fix: codespaces needs https port as base

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/6102

The URL provided for codespaces doesn't seem to work right, see https://discord.com/channels/664580571770388500/1229209241714163802/1229209241714163802

Use the https port for codespaces

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2684,7 +2684,7 @@ func (app *DdevApp) GetAllURLs() (httpURLs []string, httpsURLs []string, allURLs
 		codespaceName := os.Getenv("CODESPACE_NAME")
 		previewDomain := os.Getenv("GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN")
 		if codespaceName != "" && previewDomain != "" {
-			url := fmt.Sprintf("https://%s-%s.%s", codespaceName, app.HostWebserverPort, previewDomain)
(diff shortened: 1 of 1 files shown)


stasadev:
While this PR won't fix the issue, I should note:

If we decide to change the port here, we will also need to change `$DDEV_HOST_WEBSERVER_PORT` to `$DDEV_HOST_HTTPS_PORT` in `ddev launch`:

https://github.com/ddev/ddev/blob/0806ba7a385c88b4f7fe8b5c5439fc184f14f524/pkg/ddevapp/global_dotddev_assets/commands/host/launch#L29-L32

rfay:
AFAICT, this is the only way known to fix the problem with drupal10, and it does work.

stasadev:
If it works with drupal10, that's good. But it needs more changes, as I mentioned in my previous comment, because changing the port breaks `ddev launch -m`, `ddev mailpit`:

```
DDEV_DEBUG=true ddev mailpit
FULLURL https://solid-guide-jr4pvqvxxgqf56rj-8443.app.github.dev
```

The port in the URL is not replaces from `8443` to `8027`.

rfay:
I returned to this and spent a little time with it. It's really hard to implement https://github.com/ddev/ddev/pull/6101#issuecomment-2125684046 now because it's all been mashed into "cloud" :( 

I did find that the same problem still exists, and that this fixes it (but breaks `ddev launch -m`)

rfay:
This was an unusual approach, and apparently just took advantage of a Codespaces bug.

Closing in favor of
* #6398

PR #6107, opened 2024-04-16
docs: updates to the quickstarts, for #6095

stasadev:
## The Issue

- #6095

Follow up for quickstarts.


rpkoller:
uhhh good catches! in particular the remaining `your-...` instead of also sticking to the `my-...` pattern completely slipped through. thank you!

PR #6108, opened 2024-04-16
feat: default node version to .nvmrc when available, fixes #6071

### LINKED ISSUE #6071: Allow DDEV `nodejs_version` to use external file as reference for node version

Currently I have a `.nvmrc` file, other tools in the pipeline runs the following in it's set up

```shell
nvm install $(cat .nvmrc)
nvm alias default $(cat .nvmrc)
```

I came across this as I noticed that the `nodejs_version` default, did not match my package.json `engine` definition.
That is already managed for other people not on DDEV and CI tooling via .nvmrc

Ultimately I'd like to not have to manage multiple places...
 

### Describe your solution

~~It would be nice if I could remove the `nodejs_version` and DDEV would check for a `.nvmrc` file and take the number from there.~~

Allowing `nodejs_version: path/to/file OR version` 

It could to take the version number from a defined file
It can break if you pass something wrong, that's on you.

Arguably then this is agnostic.
eg
- `.n-node-version` n's file
- `.node-version` many many many more version managers
- `.nvmrc` nvm

### Describe alternatives

- (was initially main solution)
It would be nice if I could remove the `nodejs_version` and DDEV would check for a `.nvmrc` file and take the number from there.

- additional config for `nvmrc_path` or something perhaps to use instead or something...
Though usually that is only in root.

### Additional context

_No response_

GuySartorelli:
## The Issue

Projects with a `.nvmrc` need to duplicate the node version into DDEV config as well.
When collaborating with developers who aren't using DDEV, there is a risk of the `.nvmrc` value being updated without updating the reference in `.ddev/config.yaml` for that project.
There are other scenarios as well where this desync could be problematic.

If a `.nvmrc` exists in the project root and no `nodejs_version` is defined in the project's `.ddev/config.yaml` file, the `.nvmrc` file is used to define the default node js version.
If that file doesn't exist, the existing hardcoded DDEV default value is used.

Run `ddev status` in a project with/without a `.nvmrc` file and see the nodejs version update.
Validate that adding `nodejs_version` in config.yaml still overrides any default values.

## Automated Testing Overview

I'm not sure how to test this. Options include:
1. Make `getNodeJSDefault` a public function so it can be called directly by tests
2. Use reflection in tests and call the function using reflection (are there other cases where that's happening in existing tests?)
3. Change the way the default value is cached - it's currently cached after the first time it's grabbed primarily to avoid outputting warnings multiple times if something goes wrong, which makes it hard to test as part of an integration test.
4. ?? I'm open to suggestions about how to test this.

## Related Issue Link(s)

- #6071

## Release/Deployment Notes

Projects which have a `.nvmrc` file in the project root and which have not explicitly defined the nodejs version for that project with `nodejs_version` will swap to using the version defined in `.nvmrc` instead of the hardcoded DDEV default value.


GuySartorelli:
Marking as draft as there have been concerns raised around this approach in the linked issue.

rfay:
I'm open to general solutions to this issue, but unfortunately the problem with .nvmrc is that it pushes nvm onto people, and that certainly is not a majority workflow. We don't need to integrate one population's workflow onto another population.

Closing this for now, we can continue the conversation in the issue.

PR #6126, opened 2024-04-24
build: update hardened/prod healthcheck.sh, for #6033

stasadev:
## The Issue

- #6033

I noticed that I forgot to update `healthcheck.sh` for production.

1. Copies the supervisorctl changes from #6033
2. Small reformatting (remove extra spaces, use `printf` output for variables)


rfay (commented):
Thanks for this. The duplication in this area is certainly regrettable.

PR #6133, opened 2024-04-26
fix: getDatabaseBackups() should check error on .downloads deletion

rfay:
## The Issue

From https://discord.com/channels/664580571770388500/1232668554550186024/1232668554550186024

The .ddev/.downloads/files directory may exist somehow when trying to just get databases. 

Currently in HEAD `ddev pull` ignores failure to delete and recreate the .downloads directory.

Check the error.

### DIFF
--- a/pkg/ddevapp/provider.go
+++ b/pkg/ddevapp/provider.go
@@ -325,8 +325,14 @@ func (p *Provider) doFilesPullCommand() (filename []string, error error) {
 // getDatabaseBackups retrieves database using `generic backup database`, then
 // describe until it appears, then download it.
 func (p *Provider) getDatabaseBackups() (filename []string, error error) {
-	_ = os.RemoveAll(p.getDownloadDir())
-	_ = os.Mkdir(p.getDownloadDir(), 0755)
+	err := os.RemoveAll(p.getDownloadDir())
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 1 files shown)


stasadev (commented):
`GetBackup()` has a call to `prepDownloadDir()`:

https://github.com/ddev/ddev/blob/6cb25692aababe47c3bf8b1b02e5f474b3599de1/pkg/ddevapp/provider.go#L211-L218

And `prepDownloadDir()` recreates `files` directory:

https://github.com/ddev/ddev/blob/6cb25692aababe47c3bf8b1b02e5f474b3599de1/pkg/ddevapp/provider.go#L290-L297

> The .ddev/.downloads/files directory may exist somehow when trying to just get databases.

`prepDownloadDir()` does not check the type of backup (from `backupType`) and always creates `files`.

stasadev (approved):
Hmm, I looked at the code again, and the `.ddev/.downloads` (or `.ddev/.downloads/files`) directory is deleted and then created again later.

But in any case, if `files` dir is created all the time before call to `getDatabaseBackups()`, then if the next command fails, it will remain, the check added here is correct.

rfay:
> But in any case, if files dir is created all the time before call to getDatabaseBackups(), then if the next command fails, it will remain, the check added here is correct.

Agreed. 

The support issue in Discord appears to be something going wrong with mutagen or mutagen configuration. Even though mutagen sync is done, the files directory still exists.

PR #6136, opened 2024-04-27
feat: output the VIRTUAL_HOST if specified in environment variables of a service

kevinquillen:
## The Issue

Right now, `ddev describe` outputs most of the information one needs to access the various services in a DDEV project. There are cases however where it neglects to output custom `VIRTUAL_HOST` values from `environment` sections of a Docker Compose service. These values can allow for vanity domains of a service. This could/would be common with add-on services like Solr, NodeJS applications or similar.

This PR modifies describe.go to try and pull in the information of the `environment` section of a service.

- Build binary of this PR.
- Have a test DDEV project that has a service and set its VIRTUAL_HOST environment variable to "`VIRTUAL_HOST=site1.$DDEV_HOSTNAME`"
- Start the project with the test binary.
- Run the `describe` command.
- The output should contain your custom virtual host like the below example.

Before:

Screenshot 2024-04-27 at 4 27 01 PM

Note the NodeJS and Solr services.

After:

Screenshot 2024-04-27 at 4 26 42 PM

Note that the NodeJS and Solr services now show their correct `VIRTUAL_HOST` values.

These services are now available in the browser at those hostnames. Otherwise, it can appear to be unclear that this is possible to end users. There are cases where users may prefer to access services at custom hostnames instead of default + port and this will add the output to `ddev describe`.

I could not figure out however how to prepend http:// or https:// to these values.

## Automated Testing Overview

These changes do not modify the structure of the describe output, therefore the current `TestCmdDescribe` needs no change.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -350,7 +350,7 @@ func (app *DdevApp) Describe(short bool) (map[string]interface{}, error) {
 			}
 
 			for name, portMapping := range envMap {
-				if name == "VIRTUAL_HOST" {
+				if name != "HTTP_EXPOSE" && name != "HTTPS_EXPOSE" {
 					continue
 				}
 
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -349,35 +349,33 @@ func (app *DdevApp) Describe(short bool) (map[string]interface{}, error) {
 				appHostname = appDesc["hostname"].(string)
 			}
 
-			if httpExpose, ok := envMap["HTTP_EXPOSE"]; ok {
-				envValStr := fmt.Sprintf("%s", httpExpose)
-				portSpecs := strings.Split(envValStr, ",")
-				// There might be more than one exposed UI port, but this only handles the first listed,
-				// most often there's only one.
-				if len(portSpecs) > 0 {
-					// HTTP portSpecs typically look like <exposed>:<containerPort>, for example - HTTP_EXPOSE=1359:1358
-					ports := strings.Split(portSpecs[0], ":")
-
-					services[shortName]["http_url"] = "http://" + appHostname
-
-					if ports[0] != "80" {
-						services[shortName]["http_url"] = services[shortName]["http_url"] + ":" + ports[0]
-					}
+			for name, port := range envMap {
+				if name == "VIRTUAL_HOST" {
+					continue
 				}
-			}
 
-			if httpsExpose, ok := envMap["HTTPS_EXPOSE"]; ok {
-				envValStr := fmt.Sprintf("%s", httpsExpose)
+				portCheck := "80"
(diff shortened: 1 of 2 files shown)


kevinquillen:
I actually see a better place for this - will update the PR.

kevinquillen:
I've moved the logic to `ddevapp.go` instead, which is where it seems that most of this information is assembled to DDEV app-wide.

Now I correctly get http/https URL output if VIRTUAL_HOST is defined along with the mapped ports or their default (80/443).

image

This appears good to me, but I am new to Go so its possible it could be more compact than this.

kevinquillen:
Buildkite apparently failed for testnotddevapp - is there a way to quickly check that locally?

rfay:
> Buildkite apparently failed for testnotddevapp - is there a way to quickly check that locally

It's likely a transient failure. If you PM me your email I'll give you privileges on buildkite and you can restart tests.

kevinquillen:
Yes, appears that it was just a transient failure.

rfay:
Rebased, mostly just to run a new test runner.

pkg/ddevapp/ddevapp.go
```diff
@@ -312,34 +312,70 @@ func (app *DdevApp) Describe(short bool) (map[string]interface{}, error) {
...
+			if ok {
+				appHostname = hostname
+			} else {
+				appHostname = appDesc["hostname"].(string)
+			}
+
+			for name, portMapping := range envMap {
+				if name == "VIRTUAL_HOST" {
+					continue
+				}
```
stasadev, line 355: I think it's better to check which values can be explicitly included in this loop, otherwise the modification to `envMap` above may lead to unexpected results:

```suggestion
				if name != "HTTP_EXPOSE" && name != "HTTPS_EXPOSE" {
					continue
				}
```

stasadev (changes requested):
Tested, `ddev describe` looks good, I just have a small suggestion.

kevinquillen:
Yeah that makes sense, accepted the change.

stasadev (approved):
Thank you!

rfay:
I'm sorry we didn't get this into v1.23.1, we were trying to make it as minimally disruptive as possible. It should get into v1.23.2.

rfay:
Rebased.

rfay (approved):
Thanks, looks great to me. Manually tested in a number of situations.

Maybe a followup docs PR explaining how to use VIRTUAL_HOST as used here would be a good thing. 

I used a .ddev/docker-compose.solr_extra.yaml with
```
services:
  solr:
    environment:
      - VIRTUAL_HOST=mysolr.${DDEV_HOSTNAME}
```

And a config.solr.yaml with:
```
additional_hostnames: [mysolr.d10]
```

kevinquillen:
@rfay The docs sort of allude to it when I was looking for an applicable place. I made a small edit that shows a subdomain approach.

https://github.com/ddev/ddev/pull/6229

PR #6144, opened 2024-05-01
fix: Add host.docker.internal and customizations to ddev debug test

rfay:
## The Issue

Add a bit more detail to `ddev debug test`:
* host.docker.internal ping
* All customizations to current project
* add-ons installed in current project

Run `ddev debug test`

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -46,6 +46,13 @@ fi
 
 ddev debug configyaml | grep -v web_environment
 
+header "existing project customizations"
+grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build 2>/dev/null
+
+header "installed DDEV add-ons"
+
+ddev get --installed
+
 header "mutagen situation"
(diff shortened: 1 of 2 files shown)


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -46,6 +46,13 @@ fi
 
 ddev debug configyaml | grep -v web_environment
 
+header "existing project customizations"
+grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build 2>/dev/null
```
stasadev, line 50:
```suggestion
grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres 2>/dev/null
```

stasadev (approved):
Tested and it looks good to me 👍

The only suggestion is to show mysql and postgres files.

cmd/ddev/cmd/scripts/test_ddev.sh
rfay: Oh thanks! Spaced them out!

PR #6152, opened 2024-05-03
fix: hide custom ddev pull and ddev push commands from Amplitude, fixes #6128

### LINKED ISSUE #6128: DDEV sends a bit too much to Amplitude on commands (ddev pull arguments?)
We go to great lengths to make sure that no non-standard information leaks in the instrumentation/telemetry to Amplitude. 

However, I noticed the other day way down at the bottom of https://app.amplitude.com/analytics/ddev/chart/bzk6dz2u?linkingDashboardId=kd4mm9ft that there are some *arguments* to commands that are being reported. For example, `pantheon-news` is there, probably a `ddev pull` argument. And `dev00`, perhaps the same. 

Maybe they're all `ddev pull` arguments and we just need to massage that.

stasadev:
## The Issue

- #6128

Marks the command as custom, but only if the name is not in predefined provider list (like acquia, git, lagoon, etc.)

I haven't tested yet.

## Related Issue Link(s)

- #5110


rfay:
This looks fine to me, thanks. Does need manual testing. DDEV_VERBOSE=true might be good enough, but probably step-debugging it would be adequate as well. We can just rename one of the existing ones like lagoon or whatever and use it and see what happens.

stasadev:
I tested what would be sent to Amplitude after my change:

`ddev pull platform123`

```json
"event_properties": {
    "Arguments": null,
    "Called As": "platform123",
    "Command Name": "custom-command",
    "Command Path": "ddev pull custom-command"
}
```

`Command Name` and `Command Path` are hidden, but not `Called As`.

I think this is unintentional and should be fixed.

`cmd.CalledAs()` is a getter function that has access to a non-public attribute, and we can't override it here:

https://github.com/ddev/ddev/blob/195334dbf8ea974453183b535dc213c980edca63/cmd/ddev/cmd/root.go#L42-L48

I tried to add `IsUserDefinedCustomCommand()` here, where `cmd.CalledAs()` is added to Amplitude:

https://github.com/ddev/ddev/blob/195334dbf8ea974453183b535dc213c980edca63/pkg/amplitude/amplitude.go#L77-L81

Like this:
```go
import (
    cmd2 "github.com/ddev/ddev/cmd/ddev/cmd"
)

// Anonymize user defined custom commands.
calledAs := cmd.CalledAs()
if cmd2.IsUserDefinedCustomCommand(cmd) {
    calledAs = "custom-command"
}

builder := ampli.Command.Builder().
    Arguments(args).
    CalledAs(calledAs).
    CommandName(cmd.Name()).
    CommandPath(cmd.CommandPath())
```

But without success because of `import cycle not allowed`.

What is the best way to add a check for custom commands in `amplitude.go`?

stasadev:
I found the solution (assign a new command instead of reusing), now it sends it as:

```json
"event_properties": {
    "Arguments": null,
    "Called As": "",
    "Command Name": "custom-command",
    "Command Path": "ddev pull custom-command"
}
```

rfay (approved):
This is working great, thanks for the excellent work.

image

I did note with surprise that every time autocompletion happens, it comes through as a command. 

image

I'll open an issue to exclude __complete

PR #6153, opened 2024-05-06
fix: close global traefik file after creation

testwill:
## The Issue
Global traefik file not closed after creation
close global traefik file after creation


rfay (approved):
Thanks!

PR #6165, opened 2024-05-09
feat: set Drupal database to mariadb default on config update, fixes #6160

### LINKED ISSUE #6160: config --update do not update mariadb version for Drupal 11 and does not trigger any information

```
```
</details>

### Expected Behavior

When running `ddev config --update` in a project with latest Drupal 11 codebase, and the container was set to mariadb 10.4, the output should indicate that mariadb has to be updated manually to 10.6 after ensuring that volume data is saved. 

### Actual Behavior

The output shows that `ddev config --update` execution is successful.
```
$ ddev config --update                      
You are reconfiguring the project at /home/plopesc/projects/lullabot/core.
Auto-updating project configuration because update is requested.
Configuring a 'drupal' project with docroot '' at /home/plopesc/projects/lullabot/core

```

### Steps To Reproduce

_No response_

### Anything else?

_No response_

stasadev:
## The Issue

- #6160

Updates `ddev config --update`:
Sets the default database type if there is no database volume, otherwise displays a warning message to let users know that they may want to migrate their database.

```
mkdir my-drupal-site && cd my-drupal-site
ddev config --project-type=drupal --php-version=8.3 --docroot=web --database=mariadb:10.4
ddev start
ddev composer create drupal/recommended-project:^11.x-dev
ddev config --update
```

See the warning:

```
Default database type is mariadb:10.11, but the current actual database type is mariadb:10.4, you may want to migrate with 'ddev debug migrate-database mariadb:10.11'.
```

Then:

```
ddev delete -Oy
ddev config --update
```

See:

```
Configuring drupal project with database type 'mariadb:10.11'
```

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -407,12 +407,14 @@ func drupalConfigOverrideAction(app *DdevApp) error {
 		recommendedPostgresVersion = nodeps.Postgres16
 	}
 
-	if app.Database.Type == nodeps.MariaDB && slices.Contains(nonCompatibleMariaDBVersions, app.Database.Version) {
-		app.Database.Version = recommendedMariaDBVersion
-	} else if app.Database.Type == nodeps.MySQL && slices.Contains(nonCompatibleMySQLVersions, app.Database.Version) {
-		app.Database.Version = recommendedMySQLVersion
-	} else if app.Database.Type == nodeps.Postgres && slices.Contains(nonCompatiblePostgresVersions, app.Database.Version) {
-		app.Database.Version = recommendedPostgresVersion
+	if !nodeps.ArrayContainsString(app.GetOmittedContainers(), "db") {
+		if app.Database.Type == nodeps.MariaDB && slices.Contains(nonCompatibleMariaDBVersions, app.Database.Version) {
+			app.Database.Version = recommendedMariaDBVersion
+		} else if app.Database.Type == nodeps.MySQL && slices.Contains(nonCompatibleMySQLVersions, app.Database.Version) {
+			app.Database.Version = recommendedMySQLVersion
+		} else if app.Database.Type == nodeps.Postgres && slices.Contains(nonCompatiblePostgresVersions, app.Database.Version) {
+			app.Database.Version = recommendedPostgresVersion
+		}
 	}
 
 	return nil
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"slices"
 	"text/template"
 
 	"github.com/ddev/ddev/pkg/archive"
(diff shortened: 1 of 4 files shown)


rpkoller:
I've tested the artifact, ran `ddev config, set the docroot to `web` and the project to `drupal`, edited the config.yaml and changed mariadb to 10.4, saved and ran `ddev composer create drupal/recommended-project:11.x-dev@dev`, after that i ran `ddev config --update` and got: 

```
$> ddev config --update
You are reconfiguring the project at /Users/rkoller/Sites/oldstuff.
Auto-updating project configuration because update is requested.
Configuring a 'drupal' project with docroot 'web' at /Users/rkoller/Sites/oldstuff/web
drupal project already has database type set to non-recommended: mariadb:10.4, not changing it to recommended mariadb:10.11
```

So that works. I would have added comments on the file changes but somehow i am unable to find the string that is returned in case a not recommended database is in place so i add my comment in here. 

i think it might make sense in case an old not supported database is added to change not that single line but the last two lines and maybe even add direct instructions. cuz when i simply ran (even with no site installed yet) 

```
$> ddev config --database "mariadb:10.11"
You are reconfiguring the project at /Users/rkoller/Sites/oldstuff.
Configuring a 'drupal' project with docroot 'web' at /Users/rkoller/Sites/oldstuff/web
unable to configure project oldstuff with database type mariadb:10.4 because that database type does not match the current actual database. Please change your database type back to mariadb:10.4 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.4', and you can try a migration with 'ddev debug migrate-database mariadb:10.11' see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/
```

i get those bespoke instructions. so why not provide those in the first place and avoid the detour for someone not understanding the underlaying constraints? so instead of 

```
drupal project already has database type set to non-recommended: mariadb:10.4, not changing it to recommended mariadb:10.11
```
use something like:
```
This project is using [old database and version] old and unsupported database type. Export your database ddev export-db --file=db.sql.gz, delete your project ddev delete --omit-snapshot, change your database type ddev config --database "mariadb:10.11" (*note use the default version here for the project type), restart ddev restart and reimport ddev import-db --file=db.sql.gz
```
that is just a rough draft, but that way it would be directly actionable and no need to consult the docs

rfay (commented):
This seems to be more work than I'd expect. Can we just check to see if they have a database version already (like we do any time we config db) and if not, update to default mariadb:10.11? That seems like the default behavior I'd expect from `ddev config --update`

stasadev:
Thank you @rpkoller for your suggestions, we can make it better, but as @rfay said it adds more work.

I've updated the PR along with the manual instructions, it should be much easier now.

rfay (approved):
This looks good and tests fine. When no db exists, it updates the db config, when db does exist, it gives feedback about the situation.

I had to cherry-pick 
* https://github.com/ddev/ddev/pull/6175

to test it.

PR #6168, opened 2024-05-09
fix: check for valid db version in `ddev debug migrate-database`

stasadev:
## The Issue

```
$ ddev debug migrate-database mariadb:12345
Is it OK to attempt conversion from mariadb:10.11 to mariadb:12345?
This will export your database, create a snapshot,
then destroy your current database and import into the new database type.
It only migrates the 'db' database [Y/n] (yes):
```

Fixed the condition for valid db versions. I came across it and couldn't understand how it worked before. During debugging, I saw that instead of the version, for example, `8.0`, the whole string `mysql:8.0` was passed inside.

### DIFF
--- a/pkg/nodeps/values.go
+++ b/pkg/nodeps/values.go
@@ -1,8 +1,10 @@
 package nodeps
 
 import (
-	"github.com/maruel/natural"
 	"sort"
+	"strings"
+
+	"github.com/maruel/natural"
 
 	"github.com/ddev/ddev/pkg/config/types"
 )
@@ -175,8 +177,8 @@ func GetValidDatabaseVersions() []string {
 
 // IsValidMariaDBVersion is a helper function to determine if a MariaDB version is valid, returning
 // true if the supplied MariaDB version is valid and false otherwise.
-func IsValidMariaDBVersion(MariaDBVersion string) bool {
-	if _, ok := ValidMariaDBVersions[MariaDBVersion]; !ok {
+func IsValidMariaDBVersion(v string) bool {
+	if _, ok := ValidMariaDBVersions[strings.TrimPrefix(v, MariaDB+":")]; !ok {
 		return false
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/nodeps/values.go
```diff
@@ -207,7 +209,7 @@ func GetValidMariaDBVersions() []string {
 // IsValidPostgresVersion is a helper function to determine if a PostgreSQL version is valid, returning
 // true if the supplied version is valid and false otherwise.
 func IsValidPostgresVersion(v string) bool {
-	if _, ok := ValidPostgresVersions[v]; !ok {
+	if _, ok := ValidPostgresVersions[strings.TrimPrefix(v, Postgres+":")]; !ok {
```
rfay, line 212: Should this be `Postgres` or `postgres` ?

stasadev: `Postgres`, this is a constant:

https://github.com/ddev/ddev/blob/195334dbf8ea974453183b535dc213c980edca63/pkg/nodeps/values.go#L17-L22

pkg/nodeps/values.go
```diff
@@ -175,8 +177,8 @@ func GetValidDatabaseVersions() []string {
 
 // IsValidMariaDBVersion is a helper function to determine if a MariaDB version is valid, returning
 // true if the supplied MariaDB version is valid and false otherwise.
-func IsValidMariaDBVersion(MariaDBVersion string) bool {
-	if _, ok := ValidMariaDBVersions[MariaDBVersion]; !ok {
+func IsValidMariaDBVersion(v string) bool {
+	if _, ok := ValidMariaDBVersions[strings.TrimPrefix(v, MariaDB+":")]; !ok {
```
rfay, line 181: So these were only used in `ddev debug migrate-database` and were always wrong?

rfay (approved):
This works perfectly and looks right. I just didn't understand the validator change.. was it always wrong???

pkg/nodeps/values.go
stasadev: Yes, that's right.

I didn't even know the validator was broken, the condition in `ddev debug migrate-database` looked strange and I investigated it.

PR #6172, opened 2024-05-10
test: Use newer technique to cancel in-progress builds

rfay:
## The Issue

Someone pointed out that the use of https://github.com/styfle/cancel-workflow-action is no longer necessary. I think it was @tyler36, thanks!

There's a native technique, and the top of stylfle/cancel-workflow-action even calls it out. 

image

Use the newer technique.


stasadev (approved):
Looks good, thanks!

The idea comes from this PR:

- https://github.com/ddev/ddev-addon-template/pull/49

stasadev:
I don't know why, but after this change, when I create a new PR, I always get an email about this:

https://github.com/ddev/ddev/actions/runs/9078236290

```
PR Check
Canceling since a higher priority waiting request for 'PR Check-refs/heads/master' exists
```

Looking at the log of various actions, I see that some of them also have `Canceling since a higher priority waiting request`.

rfay:
I saw several of your PR runs get hung up yesterday, never getting assigned a runner. I restarted them, but didn't read them closely.

PR #6175, opened 2024-05-10
fix: retrieve Drupal 11 sqlite3 packages from snapshot.debian.org

rfay:
## The Issue

Suddenly we're not able to support sqlite/libsqlite 3.45.1 because it's gone from the repo we were pulling it from, so Drupal 11 startup fails.

Pull from snapshot.debian.org

I do wonder if we should only grab the 3.45 sqlite if they have omit_containers[db].

I don't understand why we didn't catch this in latest test on master (https://github.com/ddev/ddev/actions/runs/9035899711/job/24831603173) unless they just removed the package between the time that started and when I was testing. (Search for "DdevFullSiteSetup for TestPkgDrupal11").  Code is in https://github.com/ddev/ddev/blob/a96ab265933cc9c94c7533740fd40d78add0ccc5/pkg/ddevapp/ddevapp_test.go#L2221-L2232

I did manually test TestDdevFullSiteSetup with drupal11 on master and it failed there and succeeds here.

Start a d11 project

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -958,6 +958,8 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	}
 	if app.Type == nodeps.AppTypeDrupal {
 		// TODO: When ddev-webserver has required drupal 11+ sqlite version we can remove this.
+		// These packages must be retrieved from snapshot.debian.org. We hope they'll be there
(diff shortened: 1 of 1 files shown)


stasadev (approved):
Thank you, it's a good thing we didn't make a release with `ftp.debian.org`.

PR #6177, opened 2024-05-13
docs: fix broken anchor links

stasadev:
## The Issue

I found broken anchors in [`mkdocs build`](https://github.com/ddev/ddev/actions/runs/9035899706/job/24831601269#step:6:88) step.

Updates the links.


rfay (approved):
Awesome, thanks! Wish the link checker would check, but I think anchor links get a 200 even if the anchor is wrong.

PR #6182, opened 2024-05-14
fix: typo in auth ssh command description

stasadev:
## The Issue

I noticed the old image name in the `ddev auth -h`.

Renames it.


rfay (approved):
Thanks!

PR #6192, opened 2024-05-16
fix: Drupal - switch to sendmail transport always in settings.ddev.php

doradovictor:
## The Issue
When you have other default transport than sendmail, DDEV doesn't switch correctly Drupal config to use Mailpit.

We ran into this issue because we were using a custom transport in order to send our mails through an external service (Sendgrid). If "sendmail" is not the default transport, current DDEV tweaking technique for redirecting mail to Mailpit is not taking effect.

It also sets the default transport as sendmail, besides changing the sendmail transport configuration.

Configure another transport in Drupal Symfony Mailer and test that it is correctly overriden by DDEV to use Mailpit.


rfay:
Thanks @doradovictor - Could you please update the OP to show what your situation was and why you ran into this? Thanks!

doradovictor:
> Thanks @doradovictor - Could you please update the OP to show what your situation was and why you ran into this? Thanks

I tried to explain it better! 😊

rfay:
Sorry for the conflict here @doradovictor but there was a phpcs-based refactor that went in before yours, 
* https://github.com/ddev/ddev/pull/6217

Would you mind rebasing this (or just recreating it if that's easier)?

doradovictor:
@rfay I followed Github instructions and merged master with my branch, resolving conflicts. Is it ok now? :)

penyaskito (approved):
I didn't test this but makes sense to me.

pkg/ddevapp/drupal/drupal10/settings.ddev.php
```diff
@@ -38,7 +38,8 @@
   $settings['config_sync_directory'] = 'sites/default/files/sync';
 }
 
-// Override drupal/symfony_mailer default config to use Mailpit.
+// Override drupal/symfony_mailer default config to use Mailpit
```
penyaskito, line 41:
```suggestion
// Override drupal/symfony_mailer default config to use Mailpit.
```

Fixing phpcs that was fixed in the conflicting PR.

PR #6193, opened 2024-05-16
fix: do not send autocomplete commands to Amplitude, fixes #6189

### LINKED ISSUE #6189: Don't send amplitude events for autocompletion events
### Expected Behavior

DDEV doesn't need to treat autocompletion as events for amplitude.

### Actual Behavior

It reports __complete events - lots of them.

image

### Steps To Reproduce

Do something that does autocompletion and watch the output, or watch on Amplitude. Like `ddev debug instrumentation fl<tab>`

### Anything else?

_No response_

stasadev:
## The Issue

- #6189

Adds a check for the `__complete` command before sending it to Amplitude.


rfay (approved):
This seems to be working perfectly to me, thanks!

PR #6195, opened 2024-05-16
feat: add `--skip-hooks` global flag to avoid running hooks, fixes #2129, fixes #6185

### LINKED ISSUE #2129: Add `--skip-hooks` flag for ddev start, restart, etc.

My goal is to always have my commands executed under DDEV. So for example I don't want to run `composer install`, but instead `ddev composer install`. To do so I need my project to be started.

However, I often have my `config.local.yaml` defined with `post-start` hook that will `drush site-install`. So what happens is, on first install, it will start the container, try to install (and fail because composer didn't run yet), and then it will  `composer install`.

I was able to workaround it, with some [trickery](https://github.com/Gizra/multi-repo#installation). Specifically:

```
ddev composer install
cp .ddev/config.local.yaml.example .ddev/config.local.yaml
ddev restart
```

Instead of this late copy, or having to delete the `config.local.yaml`, it could be convenient to do something like:

```
ddev start  --skip-config-hooks
ddev composer install

# Restart, to run the post-start hooks.
ddev restart
```

### LINKED ISSUE #6185: flag to skip post-import-db-* hooks when using import-db
A colleague asked me about this and I was surprised to not find a native way of doing it.

I wonder if this is something logical to add?

hanoii:
## The Issue

* #2129 

Also 
* #6185

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -127,7 +127,7 @@ func Execute() {
 func init() {
 
 	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "json-output", "j", false, "If true, user-oriented output will be in JSON format.")
-	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "skip-hooks", "", false, "If true, any hook normally run by the command will be skipped.")
+	RootCmd.PersistentFlags().BoolVarP(&ddevapp.SkipHooks, "skip-hooks", "", false, "If true, any hook normally run by the command will be skipped.")
 
 	output.LogSetUp()
 
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -127,6 +127,7 @@ func Execute() {
 func init() {
 
 	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "json-output", "j", false, "If true, user-oriented output will be in JSON format.")
+	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "skip-hooks", "", false, "If true, any hook normally run by the command will be skipped.")
 
 	output.LogSetUp()
 
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -137,6 +137,10 @@ type DdevApp struct {
 	ComposeYaml               map[string]interface{} `yaml:"-"`
 }
 
+// Global variable that's set from --skip-hooks global flag.
+// If true, all hooks would be skiped.
+var SkipHooks = false
+
 // GetType returns the application type as a (lowercase) string
 func (app *DdevApp) GetType() string {
 	return strings.ToLower(app.Type)
@@ -1004,6 +1008,10 @@ func (app *DdevApp) ComposeFiles() ([]string, error) {
 
 // ProcessHooks executes Tasks defined in Hooks
 func (app *DdevApp) ProcessHooks(hookName string) error {
+	if SkipHooks {
+		output.UserOut.Debugf("Skipping the execution of %s hook...", hookName)
+		return nil
+	}
 	if cmds := app.Hooks[hookName]; len(cmds) > 0 {
 		output.UserOut.Debugf("Executing %s hook...", hookName)
 	}
--- a/pkg/ddevapp/hooks_test.go
+++ b/pkg/ddevapp/hooks_test.go
@@ -97,6 +97,18 @@ func TestProcessHooks(t *testing.T) {
 	assert.FileExists(filepath.Join(app.AppRoot, fmt.Sprintf("TestProcessHooks%s.txt", app.GetRouterHTTPSPort())))
 	assert.FileExists(filepath.Join(app.AppRoot, "touch_works_after_and.txt"))
 
+	// Make sure skip hooks work
+	ddevapp.SkipHooks = true
+	app.Hooks = map[string][]ddevapp.YAMLTask{
+		"hook-test-skip-hooks": {
+			{"exec": "\"echo TestProcessHooks > /var/www/html/TestProcessHooksSkipHooks${DDEV_ROUTER_HTTPS_PORT}.txt\""},
+		},
+	}
+	err = app.ProcessHooks("hook-test")
+	require.NoError(t, err)
+	assert.NoFileExists(filepath.Join(app.AppRoot, fmt.Sprintf("TestProcessHooksSkipHooks%s.txt", app.GetRouterHTTPSPort())))
+	ddevapp.SkipHooks = false
+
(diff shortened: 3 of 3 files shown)


cmd/ddev/cmd/root.go
```diff
@@ -125,6 +125,7 @@ func Execute() {
 func init() {
 
 	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "json-output", "j", false, "If true, user-oriented output will be in JSON format.")
+	RootCmd.PersistentFlags().BoolVarP(&output.JSONOutput, "skip-hooks", "", false, "If true, any hook normally run by the command will be skipped.")
```
hanoii, line 128: Hi, @rfay I timidly started, I have a question before moving forward, initially I added the flag right next to `--json-output`, that' sensible to me, let me know otherwise, but the main question here is where to store the flag content. I tried to look but it doesn't look like there's any other flag similar to this or so scoped as json output.

Should this be on ddevapp? Anyway, probably best to share your insight on this first.

jonesrussell: My _guess_ is to create a new global var at around line 19:
```
var skipHooksArg bool
```

rfay: I guess something like that would be the same as what we're doing of course with `--json-output`. I hate to do it with a global, even scoped, but don't have any other suggestions atm.

hanoii:
Current code is pretty simple and working, let's see how tests goes, and this will need its own test(s).

rfay (changes requested):
Seems like it will work fine to me, thanks!

Please add a cmd test that demonstrates the correct behavior.

rfay:
Rebased, looking forward to this finding its way in.

rfay:
Rebased.

hanoii:
I'll try to wrap this up this week

hanoii:
@rfay I added one test, on `TestProcessHooks` which is more a unit test, but maybe that's OK enough. I was thinking to add a cmd functional test passing on --skip-hooks to the cmd, would you want this?

I do have a question, `GOTEST_SHORT=true make test TESTARGS="-run TestProcessHooks"` fails on my local, I never what the output on the tasks, and I am not sure why. Any ideas?

I get things like
```
hooks_test.go:90: 
        	Error Trace:	/Users/ariel/Sites/work/ddev/pkg/ddevapp/hooks_test.go:90
        	Error:      	"Building project images...\n\x1b[32mProject images built in 1s.\x1b[0m\n Container ddev-TestPkgWordpress-db  Recreate\n Container ddev-TestPkgWordpress-web  Recreate\n Container ddev-TestPkgWordpress-db  Recreated\n Container ddev-TestPkgWordpress-web  Recreated\n Container ddev-TestPkgWordpress-web  Started\n Container ddev-TestPkgWordpress-db  Started\nStarting Mutagen sync process...\n\x1b[32mMutagen sync flush completed in 1s.\nFor details on sync status 'ddev mutagen st TestPkgWordpress -l'\x1b[0m\nWaiting for containers to become ready: [web db]\nStarting ddev-router if necessary...\n Container ddev-router  Running\nWaiting for additional project containers to become ready...\nAll project containers are now ready.\n" does not contain "Running task: Exec command 'ls /usr/local/bin'"
        	Test:       	TestProcessHooks
        	Messages:   	task: exec: ls /usr/local/bin
```

(does not contain)

rfay:
This one does require DDEV_DEBUG=true to work I'm pretty sure, so that's probably the problem. Task output doesn't happen these days without DDEV_DEBUG=true.

rfay:
I recently had to work around this same class of problem by explicitly setting DDEV_DEBUG in the test. Not sure what I did was the most elegant, but it got the job done. 

https://github.com/ddev/ddev/blob/213f46972bf95e1fafdb0259f55f6e601cb83662/cmd/ddev/cmd/mutagen-cmd_test.go#L22-L59

hanoii:
But the same is set in that test, only it doesn’t seem to work.  I haven’t
tried setting DDEV_DEBUG in the shell

Reply might shorter than usual / Respuesta más resumida qué de costumbre

rfay:
Run manually, see what the output of that task is. Debug it. In your debug setup, set DDEV_DEBUG=true. Step through it. See what the output is.

hanoii:
This is an existing test. The test works fine in CI. I only add added a
tiny bit. Will check later. I figured I was missing something but
apparently not

Reply might shorter than usual / Respuesta más resumida qué de costumbre

hanoii:
So if I export `DDEV_DEBUG` all works on my locak, but without out it doesn't with `GOTEST_SHORT=true make test TESTARGS="-run TestProcessHooks"`. Odd because you do have https://github.com/ddev/ddev/blob/45b6251381cfc7256de27ab302fb3e7da8def2f7/pkg/ddevapp/hooks_test.go#L27 on th test, but I guess this is a different issue. 

So my original question stands, is this an OK enough test for you, our you'd also like a cmd like test testing the actual passing of the flag. For now I just added a unit test of SkipHooks to `TestProcessHooks`

rfay:
Will study it, thanks! Maybe I don't understand t.SetEnv() properly, maybe we just need more study of how things are working.

rfay:
Rebased for review

rfay (approved):
The code looks good.

I manually tested with
```yaml
hooks:
  pre-start:
    - exec-host: echo "pre-start exec-host"
  post-start:
    - exec-host: echo "post-start exec-host"
    - exec: echo "post-start exec"
    - exec: echo "post-start exec in db"
      service: db
  post-import-db:
    - exec: echo "post-import-db exec in db"
      service: db
  pre-config:
    - exec-host: echo "pre-config exec-host"
  post-config:
    - exec-host: "echo post-config exec-host"
  pre-import-db:
    - exec: echo "pre-import-db in db"
      service: db
  pre-snapshot:
    - exec: echo "pre-snapshot exec"
  post-snapshot:
    - exec: echo "post-snapshot exec"
```

It worked fine with `ddev start --skip-hooks`, etc (and without).

The `ddev help start` (and other commands) have the related notation:

```
Global Flags:
      --skip-hooks    If true, any hook normally run by the command will be skipped.
```

rfay (commented):
I guess the one touch-up is that the docs should mention this in https://ddev.readthedocs.io/en/stable/users/configuration/hooks/

rfay:
wrt to the test failure TestProcessHooks: I get your same result. If DDEV_DEBUG=true is not set externally (in the goland run configuration, for example) then it doesn't seem to take effect. I have no idea why that would be. I even tried implementing what I suggested earlier (os.SetEnv() and later optionally remove) and it had no effect.

rfay:
Related to TestProcessHooks:
* https://github.com/ddev/ddev/issues/6313

stasadev (approved):
Tested, the change looks concise and simple, thank you!

PR #6196, opened 2024-05-16
build: fix Codespaces link

stasadev:
## The Issue

[Open in GitHub Codespaces](https://codespaces.new/ddev/ddev/pull/6196)

The link above ⬆️ opens the wrong repo, and the branch is not selected:

image

Fixes the link.
I also removes autostart from Gitpod link, because it should really have confirmation before creating a new workspace.

The link below works ⬇️

[Open in GitHub Codespaces](https://codespaces.new/?ref=20240516_stasadev_fix_codespaces_link&repo=669738338)

image

### DIFF
--- a/.github/workflows/pr-artifacts-comment.yml
+++ b/.github/workflows/pr-artifacts-comment.yml
@@ -24,11 +24,15 @@ jobs:
             // get a PR number
             const {owner, repo} = context.repo;
             const pullHeadSHA = '${{github.event.workflow_run.head_sha}}';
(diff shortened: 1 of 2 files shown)


rfay (changes requested):
Let's please make it open in new windows/tab while we're here, thanks!

stasadev:
I'd also like to open it in a new window, but that's not possible at the moment:

- https://github.com/mojombo/github-flavored-markdown/issues/28

PR #6197, opened 2024-05-16
docs: Fix a couple of broken link fragments

rfay:
## The Issue

Two Markdown link fragments were broken, and markdownlint was catching it, at least locally.

PR #6198, opened 2024-05-16
docs: Fix markdownlint problems by using different headers instead of different links

rfay:
This reverts commit 41f38c70098d6ba3613cc890e0334c88d2f4ad6a.

See discussion in #6197 

I had to do a --no-verify to push this because I have pre-push checks enabled


stasadev:
I don't see a way how to fix it quickly.

Perhaps we can simply not use such anchors, but change them:

```diff
-## Maintaining `ddev-dbserver` MySQL 5.7 & 8.0 ARM64 Images
+## Maintaining `ddev-dbserver` MySQL 5.7 and 8.0 ARM64 Images
```

```diff
-## Fork / Clone the DDEV Repository
+## Fork or Clone the DDEV Repository
```

rfay:
That's a good solution.

rfay:
OK, see what you think of this @stasadev

stasadev (approved):
It looks great, thank you!

I like that it is settled now.

PR #6208, opened 2024-05-17
fix: add more checks for existing approot

stasadev:
## The Issue

`app.CheckExistingAppInApproot()` does not work properly if you run `ddev config` several times and change the project name:

```
$ ddev config --project-name=d10
Creating a new DDEV project config in the current directory (/home/stas/code/ddev/d10) 
Once completed, your configuration will be written to /home/stas/code/ddev/d10/.ddev/config.yaml
 
Configuring a 'php' project with docroot '' at /home/stas/code/ddev/d10 
Configuration complete. You may now run 'ddev start'. 

$ ddev config --project-name=d10-rename
You are reconfiguring the project at /home/stas/code/ddev/d10.
The existing configuration will be updated and replaced.
Configuring a 'php' project with docroot '' at /home/stas/code/ddev/d10
Configuration complete. You may now run 'ddev start'.

$ ddev config --project-name=d10-rename-2
this project root /home/stas/code/ddev/d10 already contains a project named d10. You may want to remove the existing project with "ddev stop --unlist d10"
```

The `~/.ddev/project_list.yaml` file:

```yaml
d10:
    approot: /home/stas/code/ddev/d10
d10-rename:
    approot: /home/stas/code/ddev/d10
```

## How This PR Solves The Issue

Adds `app.CheckExistingAppInApproot()` to more places.

## Manual Testing Instructions

```
$ ddev config --project-name=d10

$ ddev config --project-name=d10-rename
You are reconfiguring the project at /home/stas/code/ddev/d10.
The existing configuration will be updated and replaced.
this project root /home/stas/code/ddev/d10 already contains a project named d10. You may want to remove the existing project with "ddev stop --unlist d10"

$ echo d10-rename | ddev config
You are reconfiguring the project at /home/stas/code/ddev/d10.
The existing configuration will be updated and replaced.
Project name (d10): this project root /home/stas/code/ddev/d10 already contains a project named d10. You may want to remove the existing project with "ddev stop --unlist d10"
```

PR #6209, opened 2024-05-17
fix: do not validate config in ddev clean, delete -O, stop -U

stasadev:
## The Issue

While testing #6208 I noticed a regression in:

- #5452
- #5555

```
ddev config --project-name=d10
ddev config --project-name=d10-rename
ddev config --project-name=d10-rename2
this project root /home/stas/code/ddev/d10 already contains a project named d10. You may want to remove the existing project with "ddev stop --unlist d10"

ddev stop --unlist d10
Project 'd10-rename' was found in configured directory /home/stas/code/ddev/d10 and it is already used by project 'd10'. If you have changed the name of the project, please "ddev stop --unlist d10" 
Failed to get project(s): the d10 project has an invalid hostname: 'd10.', see https://en.wikipedia.org/wiki/Hostname#Syntax for valid hostname requirements
```

And you are stuck with `ddev stop --unlist d10` until you manually clean `~/.ddev/project_list.yaml` (or run `ddev list`)

---

Also the same problem occurs when you are working on some new project type, then you change the DDEV binary and try to delete the test project with this new type, for example, you have `foobar` type:

```
ddev delete -Oy
Failed to get project(s): the test project has an invalid app type: foobar
```

And there is no way to `ddev delete -Oy` until you fix the config.

Do not run `ValidateConfig()` on a command with project arguments that don't actually need validation:
- `ddev clean`
- `ddev delete --omit-snapshot`
- `ddev stop --unlist`

The general idea is that we don't need to check the config if we delete something (I think it's annoying, just let me delete it, I don't care), and sometimes users can get stuck in a loop and we don't want that.

---

Allows to delete non-existent projects with `ddev delete -O`.

```
ddev config --project-name=d10
ddev config --project-name=d10-rename
ddev config --project-name=d10-rename2

ddev stop --unlist d10
Project d10 is already stopped. 
Project d10 has been stopped.

ddev stop --unlist d10-rename
Project d10-rename is already stopped. 
Project d10-rename has been stopped.

# manually edit `type` to something else in .ddev/config.yaml and run
ddev delete -Oy
```

And check Docker volumes and the contents of `ddev-global-cache` for:

```
ddev config --project-name=d10
ddev start
ddev stop --unlist
cd ..

docker volume ls | grep d10

docker run --rm -it -v ddev-global-cache:/mnt/ddev-global-cache ddev/ddev-webserver:v1.23.1 bash
ls /mnt/ddev-global-cache/*/d10-{web,db} /mnt/ddev-global-cache/traefik/*/d10.{yaml,crt,key}

ddev delete -O d10

docker volume ls | grep d10
# should show nothing

docker run --rm -it -v ddev-global-cache:/mnt/ddev-global-cache ddev/ddev-webserver:v1.23.1 bash
ls /mnt/ddev-global-cache/*/d10-{web,db} /mnt/ddev-global-cache/traefik/*/d10.{yaml,crt,key}
# should show nothing
```

### DIFF
--- a/cmd/ddev/cmd/clean.go
+++ b/cmd/ddev/cmd/clean.go
@@ -38,7 +38,12 @@ Additional commands that can help clean up resources:
 			util.Failed("No project provided. See ddev clean --help for usage")
 		}
 
+		// Skip project validation
+		originalRunValidateConfig := ddevapp.RunValidateConfig
+		ddevapp.RunValidateConfig = false
 		projects, err := getRequestedProjects(args, cleanAll)
+		ddevapp.RunValidateConfig = originalRunValidateConfig
+
 		if err != nil {
 			util.Failed("Failed to get project(s) '%v': %v", args, err)
 		}
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -453,7 +453,7 @@ func TestCmdDisasterConfig(t *testing.T) {
 		_, err = exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
 		assert.NoError(err)
 		_, err = exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name()+"_subdir")
-		assert.Error(err)
+		assert.NoError(err)
 		_ = os.RemoveAll(tmpDir)
 	})
 
--- a/cmd/ddev/cmd/delete.go
+++ b/cmd/ddev/cmd/delete.go
@@ -34,7 +34,7 @@ ddev delete --all`,
 		// Skip project validation if --omit-snapshot is provided
 		originalRunValidateConfig := ddevapp.RunValidateConfig
 		ddevapp.RunValidateConfig = !omitSnapshot
-		projects, err := getRequestedProjects(args, deleteAll)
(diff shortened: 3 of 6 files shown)


stasadev:
I'm not sure, but maybe we don't need the validation in `ddev delete` and `ddev stop` at all.
Let me know if I need to change anything.

rfay:
I would like `ddev delete` to delete related volumes even if the project can't be validated or doesn't exist any more. For example, `ddev delete -Oy tempproj` should kill off volumes that would have been for `tempproj` and probably things in ddev-global-cache even if the files had long since been removed. This doesn't necessarily belong in this PR, but you asked :)

stasadev:
> I would like `ddev delete` to delete related volumes even if the project can't be validated or doesn't exist any more.

Good idea and doesn't require much change, done.

rfay:
Rebased. I was testing and the problems mentioned in the OP are gone... because #6208 was fixed :)

stasadev:
Yes, that's right, now you have to manually break the config.

rfay (approved):
This seems like a really nice improvement, thanks! This has been a piece of friction for DDEV users for a long time. And allowing deletion of non-existing projects (getting rid of their databases and mutagen state) has been on my personal to-do list for some time.

PR #6210, opened 2024-05-17
feat: Replace underscores with hyphens in project-name, fixes #6206

### LINKED ISSUE #6206: Support project directories containing underscores

Drupal modules frequently contain underscores in the project names. It's annoying to have to specify a different project name, or manually convert underscores to hyphens, each time I create a DDEV project for one. It seems a bit arbitrary that underscores aren't allowed.

### Describe your solution

Allow underscores in project names.

### Describe alternatives

Automatically convert underscores to hyphens and use that as the default project name when running `ddev config`.

### Additional context

_No response_

mbomb007:
## The Issue

* #6206 

Drupal modules frequently contain underscores in the project names. It's annoying to have to specify a different project name, or manually convert underscores to hyphens, each time I create a DDEV project for one. It seems a bit arbitrary that underscores aren't allowed.

When getting the project name from the directory name, underscores are replaced with hyphens.

## Automated Testing Overview

This needs an additional test to ensure that a project directory containing underscores now has the default project-name working.

## Related Issue Link(s)

https://github.com/ddev/ddev/issues/6206

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -381,7 +381,7 @@ func handleMainConfigArgs(cmd *cobra.Command, _ []string, app *ddevapp.DdevApp) 
 	} else { // No siteNameArg passed, c.Name not set: use c.Name from the directory
 		pwd, err := os.Getwd()
 		util.CheckErr(err)
-		app.Name = filepath.Base(pwd)
+		app.Name = ddevapp.NormalizeProjectName(filepath.Base(pwd))
 	}
 
 	err = app.CheckExistingAppInApproot()
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -298,6 +298,47 @@ func TestConfigCommand(t *testing.T) {
 	}
 }
 
+// TestConfigCommandProjectNormalization tests behavior when normalizing project names.
+// For example, a directory like "some_dir" should result in project name "some-dir"
+func TestConfigCommandProjectNormalization(t *testing.T) {
+	assert := asrt.New(t)
+	origDir, _ := os.Getwd()
+
+	testMatrix := []struct {
+		name        string
+		expectation string
+	}{
+		{"normalname", "normalname"},
+		{"with-hyphen", "with-hyphen"},
+		{"with_underscore", "with-underscore"},
+		{"with.dot", "with.dot"},
+	}
+
+	for _, tc := range testMatrix {
+		t.Run(tc.name, func(t *testing.T) {
+
+			baseTestDir := t.TempDir()
+			testDir := filepath.Join(baseTestDir, tc.name)
+			err := os.Mkdir(testDir, 0755)
+			require.NoError(t, err)
+
+			// Create the app we'll use for testing.
+			app, err := ddevapp.NewApp(testDir, false)
+			require.NoError(t, err)
+
+			require.Equal(t, tc.expectation, app.Name)
+
+			t.Cleanup(func() {
+				_ = app.Stop(true, false)
+				_ = os.Chdir(origDir)
+			})
+
+			err = ddevapp.PrepDdevDirectory(app)
+			assert.NoError(err)
+		})
+	}
+}
+
 // TestConfigCommandDocrootDetection asserts the default docroot is detected.
 func TestConfigCommandDocrootDetection(t *testing.T) {
 	// Set up tests and give ourselves a working directory.
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -258,51 +258,23 @@ func TestConfigCommand(t *testing.T) {
 		name := strings.ToLower(util.RandString(16))
 		invalidAppType := strings.ToLower(util.RandString(8))
 
-		// Create an example input buffer that uses the default projectname, then a
-		// valid document root, then a valid app type.
-		input := fmt.Sprintf("\ndocroot\n%s", testValues[apptypePos])
+		// Create an example input buffer that writes the sitename, a valid document root,
+		// an invalid app type, and finally a valid app type (from test matrix)
+		input := fmt.Sprintf("%s\ndocroot\n%s\n%s", name, invalidAppType, testValues[apptypePos])
 		scanner := bufio.NewScanner(strings.NewReader(input))
 		util.SetInputScanner(scanner)
 
 		restoreOutput := util.CaptureUserOut()
 		err = app.PromptForConfig()
-		assert.NoError(err)
-		out := restoreOutput()
-
-		// Ensure we have expected vales in output.
-		projectRoot := filepath.Base(testDir)
-		assert.Contains(out, testDir)
-		assert.NotContains(out, fmt.Sprintf("%s is not a valid project name", projectRoot))
-
-		// Ensure values were properly set on the app struct.
-		// Project root contains an underscore. Make sure it was replaced with a
-		// hyphen for the project name.
-		// One of the test directories has a capital letter to ensure that the site
-		// name is lowercased.
-		assert.Equal(strings.ReplaceAll(projectRoot, "_", "-"), app.Name)
-		assert.Equal(testValues[apptypePos], app.Type)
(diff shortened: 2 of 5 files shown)


rfay (changes requested):
This is looking great, thanks! Do you want to try creating the relevant test? It will be fun!

rfay:
Here's a situation where your code couldn't compile locally, I don't think, so where manual testing is so critical. I'm pretty sure `make` didn't work here.

It's also a place where you want to use the rebase workflow instead of merging master in, which hides so many things. 

I'm quite sure that if you look at this you'll understand what happened to `input`, which is no longer defined.

rfay:
Congrats!

rfay:
Would you mind adding a commit to this making the project name lower-case as well? docker-compose needs that. 

We also remove dots/periods for docker-compose project name, but I don't think people would like removing those from the project name.

rfay:
Improving tests would be awesome.

cmd/ddev/cmd/config.go
```diff
@@ -381,7 +381,7 @@ func handleMainConfigArgs(cmd *cobra.Command, _ []string, app *ddevapp.DdevApp)
 	} else { // No siteNameArg passed, c.Name not set: use c.Name from the directory
 		pwd, err := os.Getwd()
 		util.CheckErr(err)
-		app.Name = filepath.Base(pwd)
+		app.Name = ddevapp.NormalizeSiteName(filepath.Base(pwd))
```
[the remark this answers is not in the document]
stasadev: Okay, that's a valid point.

I see we already have `ValidateProjectName` function, and the arg is `ddev config --project-name`, I think it's better to rename `NormalizeSiteName` to `NormalizeProjectName` for consistency.

Thank you for working on this!

rfay:
@mbomb007 somewhere you made a mess, I think by using the "merge" option on the github UI instead of rebase.

Anyway, there are an enormous number of conflicts here now. I started to rebase it, but I'm hoping you remember what you were trying to do and you can successfully rebase it.

I don't know if the default "update branch" (does a merge) will get you where you want to go now, but I'm certainly hesitant. Please do what you need to do, try "update branch" with merge commit if you want, but make sure you're getting the results you were aiming for.

rfay:
We can also fix this by simply rebasing it into a single commit and we'll get it fixed that way.

rfay:
I'll fix it.

stasadev:
I am wondering what is the best way to squash all the commits here into one. My strategy would be to do a soft reset to the point where this branch started, commit, rebase on master and resolve the conflicts.

BTW, the tests fail on `TestRouterPortsCheck` because of `strings.ToLower()` in `NormalizeProjectName()`.
https://buildkite.com/ddev/macos-lima/builds/713#018fdee8-586b-4ceb-bb52-19e219b3c5a2/4678-4739

rfay:
Yes, I think just squashing and then fixing. I think their merge did update it, but merge commits are pretty squirrely. But... Github offers them as the default!

rfay:
I squashed this into one commit, but then reverted the ToLower() as it had unknown consequences.

pkg/ddevapp/config_test.go
```diff
@@ -230,7 +230,7 @@ func TestConfigCommand(t *testing.T) {
 	testMatrix := map[string][]string{
 		"magentophpversion": {nodeps.AppTypeMagento, nodeps.PHPDefault},
 		"drupal7phpversion": {nodeps.AppTypeDrupal7, nodeps.PHP82},
-		"drupalphpversion":  {nodeps.AppTypeDrupal, nodeps.PHPDefault},
+		"Drupalphpversion":  {nodeps.AppTypeDrupal, nodeps.PHPDefault},
```
rfay, line 233: I don't understand why this change was made. Seems not to have anything to do with anything.

rfay (approved):
This tests out OK manually, and is relatively simple, and has test coverage. I think it's ready to go. Thanks!

rfay:
Rebased.

pkg/ddevapp/config_test.go
rfay: Ah, thank you. But I guess it doesn't do any harm anyway?

pkg/ddevapp/ddevapp.go
```diff
@@ -2838,6 +2838,7 @@ func GetActiveAppRoot(siteName string) (string, error) {
 // To use the current working directory, siteName should be ""
 func GetActiveApp(siteName string) (*DdevApp, error) {
 	app := &DdevApp{}
+	siteName = NormalizeProjectName(siteName)
```
stasadev, line 2841: If we do this only for the enclosing directory, this function should be called only if `siteName == ""`, right?

pkg/ddevapp/ddevapp.go
```diff
@@ -2863,6 +2864,11 @@ func GetActiveApp(siteName string) (*DdevApp, error) {
 	return app, nil
 }
 
+// Lowercase and replace underscores in the site name with hyphens.
```
stasadev, line 2867:
```suggestion
// NormalizeProjectName replaces underscores in the site name with hyphens.
```

pkg/ddevapp/config_test.go
```diff
@@ -258,23 +258,51 @@ func TestConfigCommand(t *testing.T) {
...
+		// Ensure we have expected vales in output.
+		projectRoot := filepath.Base(testDir)
+		assert.Contains(out, testDir)
+		assert.NotContains(out, fmt.Sprintf("%s is not a valid project name", projectRoot))
+
+		// Ensure values were properly set on the app struct.
+		// Project root contains an underscore. Make sure it was replaced with a
+		// hyphen for the project name.
+		// One of the test directories has a capital letter to ensure that the site
+		// name is lowercased.
```
stasadev, line 281: I think the test didn't actually check for lower case, because if it had, we would have seen the error here after removing `strings.ToLower()`.

stasadev (changes requested):

pkg/ddevapp/ddevapp.go
stasadev: `NormalizeProjectName` should probably be called somewhere here?:

https://github.com/ddev/ddev/blob/28a5afcdd56cc190aca33303fc1405c0c8ebabe2/pkg/ddevapp/ddevapp.go#L2791-L2804

> When getting the project name from the directory name, underscores are replaced with hyphens.

Actually replacing underscores for the directory name only makes it more complex, than we replace underscores in all cases.

pkg/ddevapp/config_test.go
```diff
@@ -258,23 +258,51 @@ func TestConfigCommand(t *testing.T) {
...
+		assert.Equal(ddevapp.NormalizeProjectName(projectRoot), app.Name)
```
stasadev, line 282: The test always passes here because `NormalizeProjectName()` is called directly.

This should be:

```suggestion
		assert.Equal(strings.ReplaceAll(projectRoot, "_", "-"), app.Name)
```

pkg/ddevapp/ddevapp.go
rfay: AFAICT GetActiveAppRoot() (if called with siteDir) is using the registration info from project_list.yaml, so doesn't need to do anything with NormalizeProjectName()

rfay:
I tried to simplify this by separating out the normalization test into TestConfigCommandProjectNormalization

pkg/ddevapp/ddevapp.go
stasadev: So we don't even need to call it here?

stasadev (approved):
Thanks, the separate test is much easier to understand.

Tested it for a prompt in `ddev config` and for `ddev config --auto`, everything works.

pkg/ddevapp/ddevapp.go
rfay: I guess that's probably true.

rfay:
We're awfully close @mbomb007 - sorry this was such a winding road, and sorry the lower-case normallization didn't work out after all the work you put into it.

rfay:
Yay!

PR #6211, opened 2024-05-18
fix: TYPO3 no longer has typo3cms command, update other out-of-date

rfay:
## The Issue

* The `typo3cms` command is no longer in [helhum/TYPO3-Console](https://github.com/TYPO3-Console/TYPO3-Console) so we shouldn't provide it either. (Thanks @rpkoller )
> Important information for longtime users
>The CLI binary typo3cms was removed with version 8.0.0 and is replaced by the TYPO3 CLI binary typo3. Commands that were called e.g. typo3cms database:updateschema before, must now be called with typo3 database:updateschema.

* A number of references to AdditionalConfiguration.php were updated to the current v12+ "additional.php" and also "settings.php"

* Use the TYPO3 quickstart to install
* Also use the [TYPO3 docs quickstart](https://docs.typo3.org/m/typo3/tutorial-getting-started/main/en-us/Installation/Install.html#install) - the GUI version is easier as it respects the additional.php


mandrasch:
Curious question: Should we test if install guides for older versions like v11.5 and v12 still work with this change? (backwards-compatibility) 

https://docs.typo3.org/m/typo3/tutorial-getting-started/11.5/en-us/Installation/TutorialDdev.html
https://docs.typo3.org/m/typo3/tutorial-getting-started/12.4/en-us/Installation/TutorialDdev.html

EOL roadmap: https://typo3.org/cms/roadmap

rfay:
> Should we test if install guides for older versions like v11.5 and v12 still work with this change?

Our general approach to TYPO3 (and now really Drupal as well) has mostly been just to *explicitly* support the current LTS. 

In general, people who understand what they're doing can make settings files or do GUI installs as need be without any trouble. The extra DDEV settings are really just to be helpful and get people started fast. But sometimes they get dependent and think it's magic :)

mandrasch:
Thanks very much for insights, didn't know about the LTS part! 💡

The link in manual testing instructions is wrong if I see it correctly - it refers to latest version (v13)
https://docs.typo3.org/m/typo3/tutorial-getting-started/main/en-us/Installation/Install.html#install

Should be this for v12 https://docs.typo3.org/m/typo3/tutorial-getting-started/12.4/en-us/Installation/Install.html 

On the "install via DDEV" page they also display a warning on branch main docs:



Just a personal note, guess there will be no breaking changes for "older" versions:

I experienced it in web agencies that projects were not updated immediately due to budget reasons / big changes needed. Starter templates also remained on "older" versions for  some time - so not breaking one or two versions before current LTS would be good for developer experience. EOL roadmap for supported versions could be a good indicator imho (if maintanable). Extended support is also often paid. Guess that's no real news, just wanted to let you know.

rfay:
DDEV provides so very much support for older things, with old database and PHP versions. We just don't have the bandwidth to try to keep things up for many, many versions. So we started the LTS strategy with TYPO3 and have mostly done that, although as you know DDEV has some code to detect older TYPO3, and now also the `drupal` type which can handle a variety. It's always a balancing act.

PR #6214, opened 2024-05-18
build: Remove concurrency/cancellation check on pr-build as it's useless [skip ci]

rfay:
## The Issue

Since we introduced the concurrency check, we've been getting useless emails about the pr-check concurrency.

Remove it.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -4,10 +4,6 @@ on:
   pull_request_target:
     types: [opened, edited, reopened, synchronize, ready_for_review, review_requested]
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
-
 permissions:
   contents: read
   issues: write


bmartinez287 (approved):

tyler36:
There is a discussion to allow disabling the notification but no resolution yet. https://github.com/orgs/community/discussions/13015

This MR removes the Github's concurrecy feature but did not restore the previous method right? 
We should re-instate the old method.

rfay:
Thanks for paying attention! 

The only workflow affected here is `pr-check` (I said in the title pr-build, but I had the wrong one). 

`pr-check`'s only annoying job is to see if the title matches the requirements and to apply labels. So it runs quickly and does no damage if it isn't cancelled. 

I could live with just removing it, personally. There is hope that we could use the PR title more successfully in the future, but we definitely don't need to worry about auto-cancellation on concurrency for this one IMO. 

None of the more important workflows were affected by this.

rfay:
This one is particularly annoying because it triggers when you actually change the PR title, for example. 

This is mentioned in https://github.com/orgs/community/discussions/13015#discussioncomment-9330564 (thanks for the pointer to the discussion). 

> I just realized that the emails even fire when you MANUALLY cancel the workflow, not just if it was from a cancel-in-progress in the YML.

> I do not need an email for an action I just manually undertook, it would be great if we could see some movement on this. It is really hard for me to notice any actual failures amongst all the cancelled ones.

PR #6232, opened 2024-05-22
test: fix TestConfigGitignore on Windows

stasadev:
## The Issue

Buildkite [error](https://buildkite.com/ddev/ddev-windows-mutagen/builds/4587#018f9da8-3c29-4a0e-a0d1-80631cf1d19e/12929-12950):

```
ls: cannot access '/C:\Users\testbot\tmp\ddevtest\Home_VRAjq683975830\ddev': No such file or directory
```

The error is a result of my refactoring in:

- #5813

I thought the path would look like `/c/Users/...`, but it is `C:\Users\...`

### DIFF
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -684,10 +684,10 @@ func TestConfigGitignore(t *testing.T) {
 	out = strings.ReplaceAll(out, "new file:   .ddev/config.yaml", "")
 	assert.NotContains(out, "new file:")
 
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks for watching and catching!

rfay:
I think we have trouble with Rancher and Colima as well. People wonder how we spend all our time as software developers :)

rfay:
Restarted a number of tests. Note that Colima has had a version update which was not supposed to affect anything, but it's updated via brew on all 3 arm64 machines.

rfay:
Thanks for being so smart to create this branch on ddev/ddev so we can properly solve this problems. If it's OK, let's fix all 3 things here in this PR.

PR #6233, opened 2024-05-22
feat: Provide TARGETARCH, TARGETOS, TARGETPLATFORM to docker builds

rfay:
## The Issue

I happened to notice that we kept having to provide `ARG TARGETPLATFORM` to some Dockerfile recipes

Automatically provide TARGETARCH, TARGETOS, TARGETPLATFORM so people don't need that. 

I used this Dockerfile:

```
RUN printf "TARGETPLATFORM=${TARGETPLATFORM}\nTARGETARCH=${TARGETARCH}\nTARGETOS=${TARGETOS}\n" >/var/lib/targetplatform.txt
```

And then ran `ddev exec at /var/lib/targetplatform.txt` after `ddev start`


stasadev (commented):
It looks great!

Please add these variables to https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/#build-time-environment-variables

And maybe we need an example how use them.

rfay:
Converted to draft. TODO includes:
* Docs
* test
* Example

stasadev (approved):
Looks good to me.

Issue #6235, opened 2024-05-23
Upgrade failing due to Blackfire "payment required" error


tyler36:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

@iainsaxon  reported that they are having troubles upgrading DDEV. #https://github.com/ddev/ddev/issues/5620

> This has become an issue again when I ddev start or ddev d refresh a project:
> 

Also seen in Discord: https://discord.com/channels/664580571770388500/1243100859756183612/1243100859756183612

Starting this issue to track it but have done no research yet (ticked boxes are meaningless).

### Expected Behavior

Upgrade possible

### Actual Behavior

```
1.674 Get:18 http://nginx.org/packages/debian bookworm/nginx arm64 Packages [7973 B]
1.811 Err:19 http://packages.blackfire.io/debian any InRelease
1.811   402  Payment Required [IP: 79.125.77.192 80]
2.089 Get:20 https://apt.postgresql.org/pub/repos/apt bookworm-pgdg InRelease [123 kB]
```

tyler36:
Could someone a complete copy of the debug log?
It will be helpful for debugging. 😄

andrewbelcher:
I believe the issue is only when you have custom packages to install - the `apt-get update` fails as the repo endpoint returns a 402. Here's an example from a project I've tried to start:
```
#16 [web 7/8] RUN apt-get update -y &&     apt-get install -y --no-install-recommends --no-install-suggests poppler-utils
#16 0.330 Get:1 http://deb.debian.org/debian bookworm InRelease [151 kB]
#16 0.345 Get:2 http://deb.debian.org/debian bookworm-updates InRelease [55.4 kB]
#16 0.345 Get:3 http://deb.debian.org/debian-security bookworm-security InRelease [48.0 kB]
#16 0.352 Get:4 http://nginx.org/packages/debian bookworm InRelease [2860 B]
#16 0.383 Get:5 https://packages.sury.org/php bookworm InRelease [7542 B]
[...]

Issue #6078, opened 2024-04-10
blackfire upgrade is failing, needs to be fixed


rfay:
### Expected Behavior

```
ddev ssh
sudo apt update && sudo apt upgrade -y
```

gets

```
Preparing to unpack .../blackfire_2%3a2.26.3_arm64.deb ...
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
Unpacking blackfire (2:2.26.3) over (2:2.26.1) ...
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
dpkg: warning: old blackfire package post-removal script subprocess returned error exit status 1
dpkg: trying script from the new package instead ...
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
dpkg: error processing archive /var/cache/apt/archives/blackfire_2%3a2.26.3_arm64.deb (--unpack):
 new blackfire package post-removal script subprocess returned error exit status 1
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
dpkg: error while cleaning up:
 new blackfire package post-removal script subprocess returned error exit status 1
```

`apt-cache policy blackfire`:

```
root@82d8e99a1c9d:/# apt list --upgradable
Listing... Done
blackfire/any 2:2.26.3 arm64 [upgradable from: 2:2.26.1]
nodejs/nodistro 20.12.2-1nodesource1 arm64 [upgradable from: 20.12.1-1nodesource1]
root@82d8e99a1c9d:/# apt-cache policy blackfire
blackfire:
  Installed: 2:2.26.1
  Candidate: 2:2.26.3
  Version table:
     2:2.26.3 500
        500 http://packages.blackfire.io/debian any/main arm64 Packages
     2:2.26.2 500
        500 http://packages.blackfire.io/debian any/main arm64 Packages
 *** 2:2.26.1 500
        500 http://packages.blackfire.io/debian any/main arm64 Packages
        100 /var/lib/dpkg/status
     2:2.26.0 500
        500 http://packages.blackfire.io/debian any/main arm64 Packages
     2:2.25.0 500
        500 http://packages.blackfire.io/debian any/main arm64 Packages
     2:2.24.4 500
```

rfay:
I am able to do ddev-webserver build successfully, so the upgrade may fail, but fresh install is OK. Short-term fix works to just rebuild ddev-webserver.

[...]

PR #6236, opened 2024-05-23
build: standardize apt-get usage, remove blackfire from apt sources, for #6235, for #6078

stasadev:
## The Issue

- #6235

`apt-get update` can fail in the user's custom Dockerfiles.

I thought we could search for `apt-get update` using a regex and automatically replace it in the resulting Dockerfile, but that seems like overkill because people can write it with different variations, like:

- `apt-get update`
- `apt-get update -y`
- `apt-get update -qq`
- `apt-get update -qq >/dev/null`

And the same with `apt`, so it's probably impossible to write a good regex to cover all cases and not break anything.

Replaces `apt-get update` with `(apt-get update || true)` in the docs.

Uses `apt-get` instead of `apt` everywhere.

Removes `/etc/apt/sources.list.d/blackfire.list` as we pin the blackfire package to a specific version and don't need it in `apt-get update`:

- #6078

1. `ddev start`
2. `ddev ssh`
3. `sudo apt-get update | grep blackfire` should give an empty result


rfay (approved):
Thanks, looks perfect.

PR #6239, opened 2024-05-23
fix: prevent various Mutagen conflicts between daemons when changing versions or global directory, fixes #6234

### LINKED ISSUE #6234: Critical global configuration misbehavior: Accidental switching between ~/.ddev and ~/.config/ddev
### Expected Behavior

The default directory should be ~/.ddev in the absence of XDG_CONFIG_HOME

On WSL2/Linux:

With mutagen enabled in the project, 

If I remove all config, `rm -r ~/.ddev ~/.config/ddev` and then `ddev start`

* Global config is created in ~/.ddev
* However, an empty ~/.config/ddev is also created

If I then `ddev stop`

* The various things in ~/.config/ddev are populated
* ~/.config/ddev becomes the global config dir

rfay:
## The Issue

* #6234 

With our new ability to change the global configuration directory from `~/.ddev` to other places come new risks for mutagen. 

(Removed): Mutagen now keeps its directory in `<globaldir>/.mdd` instead of `<home>/.ddev_mutagen_data_directory`. 

There's a risk in changing that mutagen could continue using an obsolete mutagen session or accidentally use a docker volume that did not match.

* When computing the mutagen hash for uniqueness, include not only the `mutagen.yml` contents but also the location of the `mutagen.yml` and the global configuration location.
* Revert to using ~/.ddev_mutagen_data_directory to resolve risks of upgrades and environment variable changes (XDG_CONFIG_HOME)

Still working on explicit instructions that can demonstrate the underlying problem.

## Automated Testing Overview

I think the existing tests probably already do enough.

### DIFF
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -97,7 +97,11 @@ func PauseMutagenSync(app *DdevApp) error {
 // PauseAllMutagenSync pauses all mutagen sync sessions found in
 // a provided MUTAGEN_DATA_DIRECTORY
 func PauseAllMutagenSync(mutagenDataDirectory string) error {
-	_, err := exec.RunHostCommandWithEnv(globalconfig.GetMutagenPath(), []string{"MUTAGEN_DATA_DIRECTORY=" + mutagenDataDirectory}, "sync", "pause", "-a")
+	env := []string{"MUTAGEN_DATA_DIRECTORY=" + mutagenDataDirectory}
+	out, err := exec.RunHostCommandWithEnv(globalconfig.GetMutagenPath(), env, "sync", "pause", "-a")
+	util.Verbose("ran mutagen sync pause -a(%s) output=%s err=%v", mutagenDataDirectory, out, err)
+	out, err = exec.RunHostCommandWithEnv(globalconfig.GetMutagenPath(), env, "daemon", "stop")
+	util.Verbose("ran mutagen daemon stop output=%s err=%v", mutagenDataDirectory, out, err)
 	return err
 }
 
@@ -105,6 +109,7 @@ func PauseAllMutagenSync(mutagenDataDirectory string) error {
 // that may belong to other configured DDEV setups
 // especially from previous versions
 func PauseOldMutagenSync() {
+	util.Debug("Attempting to terminate any mutagen sessions from other versions of DDEV")
 	userHome, _ := os.UserHomeDir()
 
 	allKnownMutagenDataDirectories := []string{
@@ -114,7 +119,8 @@ func PauseOldMutagenSync() {
 	}
 	ourMutagenDataDirectory := globalconfig.GetMutagenDataDirectory()
 	for _, d := range allKnownMutagenDataDirectories {
-		if d != ourMutagenDataDirectory {
+		if d != ourMutagenDataDirectory && fileutil.FileExists(filepath.Join(d, "sessions")) {
+			util.Debug("Pausing mutagen sessions for '%s'", d)
 			err := PauseAllMutagenSync(d)
 			if err != nil {
 				util.Debug("Error pausing mutagen sync for directory %s", d)
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -135,7 +135,9 @@ func GetMutagenConfigFilePath(app *DdevApp) string {
 // GetMutagenConfigFileHash returns the SHA1 hash of the mutagen.yml
 func GetMutagenConfigFileHash(app *DdevApp) (string, error) {
 	f := GetMutagenConfigFilePath(app)
-	hash, err := fileutil.FileHash(f)
+	// Create hash based on mutagen.yml file contents, location,
+	//and global config
+	hash, err := fileutil.FileHash(f, globalconfig.GetGlobalDdevDirLocation())
 	if err != nil {
 		return "", err
 	}
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -9,7 +9,6 @@ import (
 	osexec "os/exec"
 	"path"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"time"
@@ -646,22 +645,8 @@ func MutagenReset(app *DdevApp) error {
 }
 
 // GetMutagenVolumeName returns the name for the Mutagen Docker volume
-// It's based on the app.Name, Docker host ID, and MUTAGEN_DATA_DIRECTORY
-// This should generate a unique volume name and prevent deleting a docker volume
-// which belongs to another docker host or MUTAGEN_DATA_DIRECTORY sync
 func GetMutagenVolumeName(app *DdevApp) string {
-	name := app.Name + "_" + "project_mutagen" + "_" + globalconfig.GetMutagenDataDirectory() + "_" + dockerutil.GetDockerHostID()
-	name = strings.ToLower(name)
-	re := regexp.MustCompile(`[^a-z0-9-_]`)
-	name = re.ReplaceAllString(name, "-")
-	// Docker volume name is probably 255, but not clearly documented
-	// We'll limit to 127. We probably won't hit that but if we do the
-	// Docker hostID will be the first thing truncated, and it's not likely
-	// part of the problem anyway.
-	if len(name) > 127 {
-		name = name[:127]
-	}
-	return name
+	return app.Name + "_" + "project_mutagen"
 }
 
 // MutagenMonitor shows the output of `mutagen sync monitor <syncName>`
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -101,6 +101,28 @@ func PauseAllMutagenSync(mutagenDataDirectory string) error {
 	return err
 }
 
+// PauseOldMutagenSync attempts to pause any mutagen sessions
+// that may belong to other configured DDEV setups
+// especially from previous versions
+func PauseOldMutagenSync() {
+	userHome, _ := os.UserHomeDir()
+
+	allKnownMutagenDataDirectories := []string{
+		filepath.Join(userHome, ".ddev_mutagen_data_directory"), // through v1.23.1
+		filepath.Join(userHome, ".ddev", ".mdd"),                // default current
+		filepath.Join(userHome, ".config", "ddev", ".mdd"),
+	}
+	ourMutagenDataDirectory := globalconfig.GetMutagenDataDirectory()
+	for _, d := range allKnownMutagenDataDirectories {
+		if d != ourMutagenDataDirectory {
+			err := PauseAllMutagenSync(d)
+			if err != nil {
+				util.Debug("Error pausing mutagen sync for directory %s", d)
+			}
+		}
+	}
+}
+
 // SyncAndPauseMutagenSession syncs and pauses a Mutagen sync session
 func SyncAndPauseMutagenSession(app *DdevApp) error {
 	if !app.IsMutagenEnabled() {
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -94,32 +94,6 @@ func PauseMutagenSync(app *DdevApp) error {
 	return nil
 }
 
-// StopOldMutagenDaemons attempts to stop any mutagen daemons
-// that may belong to other configured DDEV setups
-// especially from previous versions or other values of XDG_CONFIG_HOME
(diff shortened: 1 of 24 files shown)


pkg/globalconfig/global_config.go
```diff
@@ -597,15 +597,23 @@ func GetGlobalDdevDir() string {
...
 	if xdgConfigHomeDir != "" {
 		return filepath.Join(xdgConfigHomeDir, "ddev")
 	}
-	// If Linux and ~/.config/ddev exists, use it,
+	// If Linux and ~/.ddev doesn't exist and
+	// ~/.config/ddev exists, use it,
 	// we don't create this directory.
-	if runtime.GOOS == "linux" {
+	_, userHomeDotDdevErr := os.Stat(userHomeDotDdev)
+	if runtime.GOOS == "linux" && userHomeDotDdevErr != nil {
```
stasadev, line 616:
```suggestion
	if runtime.GOOS == "linux" && os.IsNotExist(userHomeDotDdevErr) {
```

pkg/globalconfig/global_config.go
```diff
@@ -616,10 +624,7 @@ func GetGlobalDdevDirLocation() string {
 	}
 	// Otherwise, use ~/.ddev
 	// It will be created if it doesn't exist.
-	userHome, err := os.UserHomeDir()
-	if err != nil {
-		logrus.Fatal("Could not get home directory for current user. Is it set?")
-	}
+
 	return filepath.Join(userHome, ".ddev")
```
stasadev, line 628:
```suggestion
	return userHomeDotDdev
```

stasadev (changes requested):

stasadev (changes requested):
I can't make a suggestion for this line, but we can check for directory with:

```diff
-if _, err := os.Stat(linuxDir); err == nil {
+if stat, err := os.Stat(linuxDir); err == nil && stat.IsDir() {
```

rfay:
@stasadev please take a very skeptical look at this. It over-does a couple of things, and it probably actually changes your intent for default global config location on Linux. Now it does XDG_CONFIG_HOME if set, then ~/.ddev if exists, then ~/.config/ddev if exists, and that should be the same on all platforms. AFAICT the test failures are upstream ones at this point, I'm still working on those, but I think this can go in when you and I have looked at it again.

stasadev (approved):
I checked for active Mutagen processes and confirmed that they stop after running the new version of DDEV.

There is also an edge case where people decide to use `XDG_CONFIG_HOME`, run `ddev start`, and then set it to a different location or unset it, and leave the old Mutagen running.
But I don't think there is anything we can do about that (since we have no way of knowing the old value for `XDG_CONFIG_HOME`).

Thank you, I think this PR will smooth things over for the Mutagen.

rfay:
This one was intended to go in before the other ones and somehow I lost track of it while working on those :)

rfay:
I refactored this AGAIN to detect when the MUTAGEN_DATA_DIRECTORY has changed and to use `mutagen daemon stop` per @xenoscopic's suggestion in https://github.com/ddev/ddev/issues/6234#issuecomment-2135929437 - instead of pausing sessions when it has. Added global config for last_mutagen_data_directory.

pkg/ddevapp/mutagen.go
```diff
@@ -94,6 +94,27 @@ func PauseMutagenSync(app *DdevApp) error {
...
+func StopOldMutagenDaemons() {
+	ourMutagenDataDirectory := globalconfig.GetMutagenDataDirectory()
+	util.Debug("Attempting to terminate any mutagen daemons from other versions of DDEV. Current MUTAGEN_DATA_DIRECTORY=%s", ourMutagenDataDirectory)
+
+	userHome, _ := os.UserHomeDir()
+	allKnownMutagenDataDirectories := []string{
+		filepath.Join(userHome, ".ddev_mutagen_data_directory"), // through v1.23.1
+		filepath.Join(userHome, ".ddev", ".mdd"),                // default current
+		filepath.Join(userHome, ".config", "ddev", ".mdd"),
+	}
```
stasadev, line 109: If we allow `MUTAGEN_DATA_DIRECTORY` to be changed from the env variable, and that value is saved in the DDEV global configuration, perhaps we should add it to this list:

```go
if globalconfig.DdevGlobalConfig.LastMutagenDataDirectory != "" && !slices.Contains(allKnownMutagenDataDirectories, globalconfig.DdevGlobalConfig.LastMutagenDataDirectory) {
	allKnownMutagenDataDirectories = append(allKnownMutagenDataDirectories, globalconfig.DdevGlobalConfig.LastMutagenDataDirectory)
}
```

This will cover the case when someone adds a custom `MUTAGEN_DATA_DIRECTORY`, but then changes it, or unsets it:
```
export DDEV_DEBUG=true
export MUTAGEN_DATA_DIRECTORY=~/.ddev_mutagen2
ddev start
# if we don't check for LastMutagenDataDirectory, mutagen daemon in ~/.ddev_mutagen2 won't be stopped
unset MUTAGEN_DATA_DIRECTORY
ddev start
```

pkg/testcommon/testcommon.go
```diff
@@ -260,12 +260,14 @@ func CopyGlobalDdevDir(t *testing.T) string {
 		require.NoError(t, err)
 	}
 	// Reset $MUTAGEN_DATA_DIRECTORY
-	err = os.Unsetenv("MUTAGEN_DATA_DIRECTORY")
-	require.NoError(t, err)
+	_ = os.Unsetenv("MUTAGEN_DATA_DIRECTORY")
+	// globalconfig.GetMutagenDataDirectory gets new version of MUTAGEH_DATA_DIRECTORY
```
stasadev, line 264:
```suggestion
	// globalconfig.GetMutagenDataDirectory gets new version of MUTAGEN_DATA_DIRECTORY
```

stasadev (changes requested):

pkg/ddevapp/mutagen.go
rfay: I don't think we ever allow MUTAGEN_DATA_DIRECTORY to be set in the user's environment do we? We force it to the configuration determined by the code.

stasadev: I think this was an unintentional change from me in #5813, but `GetMutagenDataDirectory()` now reads `MUTAGEN_DATA_DIRECTORY` env before setting the directory.

rfay: But we *set* MUTAGEN_DATA_DIRECTORY in the init() in a.go.
https://github.com/ddev/ddev/blob/b3f7cf6df550f6b6d2cedaca9d64c98521ec2fcd/cmd/ddev/cmd/a.go#L17

rfay: I see. I'm going to have to take another look at that. Thanks for pointing that out.

stasadev: Oh, this was added from the start:

- #4109

https://github.com/ddev/ddev/blob/b3f7cf6df550f6b6d2cedaca9d64c98521ec2fcd/pkg/globalconfig/global_config.go#L143-L151

Some may rely on using a custom `MUTAGEN_DATA_DIRECTORY` even if this is not documented in our docs.

rfay: I don't think it's likely that people are setting this, but I sure appreciate you pointing this out. 

It's quite awkward for mutagen not to have a command-line arg instead of env var. I'll open an issue about that.

rfay: * https://github.com/mutagen-io/mutagen/issues/499

rfay:
I have tested this:
* macOS with mutagen
* Gitpod without mutagen
* Ubuntu without mutagen

I'll still try to test Ubuntu with mutagen.

@stasadev please make sure I didn't break the docs. At one point I did. It's complicated how it works slightly differently on Linux.

stasadev (commented):
I see that we cannot prevent all edge cases with Mutagen. No matter how hard we try, there is a way to break it.

Of course, there is a chance that you might break your project files while playing around with it. I don't mind if you decide to use `~/.ddev_mutagen_data_directory` again if this change might be too disruptive.

rfay:
The worst-case scenario is caused by the fact that we delete/empty the docker volume when we don't have a matching mutagen sync. It's always possible that that has been a bad idea.  It sure has worked well to this point.

stasadev:
> The worst-case scenario is caused by the fact that we delete/empty the docker volume when we don't have a matching mutagen sync.

What if we name the volume based on the `MUTAGEN_DATA_DIRECTORY`?

https://github.com/ddev/ddev/blob/e0b5fb9d3cbf69709d9494c395ad4c25e82c7bc8/pkg/ddevapp/mutagen.go#L614-L617

We can calculate `sha1` of the `MUTAGEN_DATA_DIRECTORY` path, get first 5 symbols, add it to the volume name. In this case we will have different volumes.

rfay:
I like the idea. It abandons an old docker volume, but that's OK.

rfay:
It was such a good idea! But my brain was off-kilter. The mutagen daemon sync session syncs between filesystem on the host and the *container*'s /var/www/html. So the more of these we have going, the more trouble we have. 

Here's the situation on my computer right now:

<details><summary>MUTAGEN_DATA_DIRECTORY=/Users/rfay/tmp/dd1/ddev/.mdd</summary>
```
$ export MUTAGEN_DATA_DIRECTORY=/Users/rfay/tmp/dd1/ddev/.mdd; ~/.ddev/bin/mutagen sync list
--------------------------------------------------------------------------------
Name: d10
Identifier: sync_X4NOdoqOMTlIgEMo6lpDn9TK06jtsRBMBPhgqxRCRtW
Alpha:
	URL: /Users/rfay/workspace/d10
	Connected: Yes
	Synchronizable contents:
		7040 directories
		27066 files (136 MB)
		4 symbolic links
Beta:
	URL: docker://ddev-d10-web/var/www/html
		DOCKER_HOST=unix:///Users/rfay/.orbstack/run/docker.sock
	Connected: Yes
	Synchronizable contents:
		7040 directories
		27066 files (136 MB)
		4 symbolic links
Status: Watching for changes
--------------------------------------------------------------------------------
```
</details>

<details><summary>export MUTAGEN_DATA_DIRECTORY=/Users/rfay/.ddev/.mdd; ~/.ddev/bin/mutagen sync list</summary>
```
$ export MUTAGEN_DATA_DIRECTORY=/Users/rfay/.ddev/.mdd; ~/.ddev/bin/mutagen sync list
--------------------------------------------------------------------------------
Name: d10
Identifier: sync_xT9dQybWgGvLks1FxjkrXtUhyQDyaK1AlgW6mJyQ4Bf
Alpha:
	URL: /Users/rfay/workspace/d10
	Connected: Yes
	Synchronizable contents:
		7040 directories
		27066 files (136 MB)
		4 symbolic links
Beta:
	URL: docker://ddev-d10-web/var/www/html
		DOCKER_HOST=unix:///Users/rfay/.orbstack/run/docker.sock
	Connected: Yes
	Synchronizable contents:
		7040 directories
		27066 files (136 MB)
		4 symbolic links
Status: Watching for changes
--------------------------------------------------------------------------------
```
</details>

Now both of these are synchonizing from the same alpha (host) to the same beta (here the docker container) at docker://ddev-d10-web/var/www/html - both fighting away!

With this situation I was able to get the host-side code deleted nearly every time :) I happened to have GoLand set up with a debug config that had XDG_CONFIG_HOME=~/tmp/dd1 and an iterm terminal that had XDG_CONFIG_HOME unset (so using ~/.ddev)

Every time (almost?) I do the start in the debugger it deletes everything. And if I kill the debug session... it leaves yet another mutagen session running, synchronizing to the same place. I had 8 of them :)

rfay:
Oh so sad. Latest commits 
* Revert (previously unpushed) GetMutagenVolumeName() "improvements", which were pretty fancy but offered no value in the end. And they caused a lot of calls to `docker context`
* Revert use of MUTAGEN_DATA_DIRECTORY being in the global ddev dir + .mdd. It seemed impossible to sort out the risks with this.

stasadev:
I prepared a revert patch for the docs:

```
curl https://github.com/ddev/ddev/commit/d3b8ee8ca59875ac13102337b261e1f4d4b13079.patch -o mutagen.patch
git apply mutagen.patch
```

rfay:
Thanks so much for that. I'll apply that after we see how the tests do.

(It would probably have been easier and easier to review to do a PR against this PR)

rfay:
I think this is ready to go. Or I'll die. I'm pretty sure that we've removed most of the risk by going back to the traditional MUTAGEN_DATA_DIRECTORY.

Anyway, it's important to get it in and get experience with it. I think it will do fine.

PR #6242, opened 2024-05-24
fix: create `working_dir` for `web` if it doesn't exist, fixes #6158

### LINKED ISSUE #6158: DDEV support in Drupal Starshot

This is just to make the DDEV developers aware of the issue https://github.com/phenaproxima/starshot-prototype/issues/49 in https://github.com/phenaproxima/starshot-prototype for https://www.drupal.org/starshot. (Maybe I could mention a DDEV developer, in that issue and the person would get pinged? I am not sure ....)

### Describe your solution

Perhaps the DDEV developers got an idea how to solve this best?

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- https://github.com/phenaproxima/starshot-prototype/issues/94
- #6158

If there is a config in `.ddev/config.yaml`:

```yaml
working_dir:
  web: /var/www/html/something
```

And the `something` directory doesn't exist on the host side, it will be created by Docker as root.

We can create the directory beforehand and it is only needed if it is in the approot.

1. Use Linux/WSL2.
2. `ddev config global --performance-mode=none`
3. `ddev config --web-working-dir=/var/www/html/something`
4. `ddev start`
5. check file permissions `ls -l something`

## Related Issue Link(s)

- https://github.com/moby/moby/issues/2259

### DIFF
--- a/pkg/ddevapp/utils.go
+++ b/pkg/ddevapp/utils.go
@@ -522,27 +522,6 @@ func (app *DdevApp) CanUseHTTPOnly() bool {
 	return false
 }
 
-// CreateWebWorkingDir creates a web working_dir if it does not exist on the host side.
-// If the app has a custom working_dir for the web service, like `/var/www/html/some/deep/dir`,
-// where `/var/www/html/` path corresponds to the project root,
-// and `some/deep/dir` does not exist on the host side, we should create it.
-// Otherwise, Docker will create it as root user (when Mutagen is disabled).
-// This problem (particularly for Docker volumes) is described in
-// https://github.com/moby/moby/issues/2259
-func (app *DdevApp) CreateWebWorkingDir() {
-	if app.WorkingDir != nil {
-		if workingDir := app.WorkingDir["web"]; workingDir != "" {
-			workingDirPrefix := app.DefaultWorkingDirMap()["web"]
-			if strings.HasPrefix(workingDir, workingDirPrefix) {
-				fullPath := filepath.Join(app.AppRoot, strings.TrimPrefix(workingDir, workingDirPrefix))
-				if _, err := os.Stat(fullPath); os.IsNotExist(err) {
-					_ = os.MkdirAll(fullPath, 0755)
-				}
-			}
-		}
-	}
-}
-
 // Turn a slice of *DdevApp into a map keyed by name
 func AppSliceToMap(appList []*DdevApp) map[string]*DdevApp {
 	nameMap := make(map[string]*DdevApp)
--- a/pkg/ddevapp/utils.go
+++ b/pkg/ddevapp/utils.go
@@ -522,6 +522,27 @@ func (app *DdevApp) CanUseHTTPOnly() bool {
 	return false
 }
 
+// CreateWebWorkingDir creates a web working_dir if it does not exist on the host side.
+// If the app has a custom working_dir for the web service, like `/var/www/html/some/deep/dir`,
+// where `/var/www/html/` path corresponds to the project root,
+// and `some/deep/dir` does not exist on the host side, we should create it.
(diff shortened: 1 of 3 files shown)


rfay:
I guess our quickstarts used to say `--create-docroot` and now they do it without even asking for that. So we always create the docroot in advance. 

Adding other mounts with a `docker-compose.mounts.yaml` would be an interesting test. 

Also, we normally mount the upload_dirs  when using mutagen. We should be having trouble with that in this situation. But of course there we already have created the host-side directory also (because it can't be mounted otherwise)

rfay:
Rebased.

rfay (changes requested):
I would prefer that we do this work in a separate function. Even though GetWorkingDir() isn't called a lot of places, the only place we need to do this work is in a function called right before RenderComposeYaml(). 

Thanks for your flexibility :)

We should probably have long ago put "/var/www/html" into a constant so it could be used more reliably. We don't have to do that here.

rfay:
There's another problem here. If you `ddev config --web-working-dir=/var/www/html/some/deep/dir` then `some` and `some/deep` and `some/deep/dir` all get root ownership on Linux. Maybe the existing approach here is fine and solves that though.

stasadev:
> Maybe the existing approach here is fine and solves that though.

Yes, this works for nested directories as well.

pkg/ddevapp/utils.go
```diff
@@ -522,6 +522,27 @@ func (app *DdevApp) CanUseHTTPOnly() bool {
...
 
+// CreateWebWorkingDir creates a web working_dir if it does not exist on the host side.
+// If the app has a custom working_dir for the web service, like `/var/www/html/some/deep/dir`,
+// where `/var/www/html/` path corresponds to the project root,
+// and `some/deep/dir` does not exist on the host side, we should create it.
+// Otherwise, Docker will create it as root user (when Mutagen is disabled).
+// This problem (particularly for Docker volumes) is described in
+// https://github.com/moby/moby/issues/2259
+func (app *DdevApp) CreateWebWorkingDir() {
+	if app.WorkingDir != nil {
```
rfay, line 533: Do we need to do any of this logic, or just use the working dir whatever it is, and os.MkdirAll() whatever it is? Since it mostly exists, it will do no harm right?

rfay: BTW, this organization is what I was hoping for, thanks.

stasadev: Yes, we need to do all this logic because it's done *outside* the container, before `docker-compose build`.

And yes, we need to check that `working_dir` is set to some path inside `/var/www/html/`, otherwise we won't have a corresponding directory on the host side.

rfay: I would be more comfortable with this if it were encapsulated a bit more. How about creating an app.GetHostWorkingDir(servicename) that does what needs to be done, removing the proper prefix. Then this code here would just be
```
d := GetHostWorkingDir("web")
os.MkDirAll(d, 0755)
```

I think it would be easier to follow.  

Another option is to return two values (in-container and on-host) from app.GetWorkingDir(). It's only used 5 places, so might not be too painful.

rfay (changes requested):

stasadev:
I refactored it to `app.GetHostWorkingDir`, and even found a bug in my previous code: some project types had a specific `docrootWorkingDir` that overrides the default, which leads to a different prefix, so I switched to using `app.GetAbsAppRoot`.

gitressa:
This is great, thanks! I look forward to trying it, as soon as the DDEV install for Drupal Starshot gets fixed.

- https://github.com/phenaproxima/starshot-prototype/issues/100

Issue #6249, opened 2024-05-26
`ddev import-db` fails with "Error: Unknown command '\-'" because of new directive in mariadb-dump output


ioanma:
## TL;DR:

If  `ddev import-db` or `ddev pull` is failing with `Error: Unknown command '\-'"`

* Make sure you're using DDEV v1.23.1+
* Update the appropriate ddev-dbserver image based on your version of Mariadb in the .ddev/config.yaml
```
docker pull ddev/ddev-dbserver-mariadb-10.11:v1.23.1
docker pull ddev/ddev-dbserver-mariadb-10.6:v1.23.1
docker pull ddev/ddev-dbserver-mariadb-10.5:v1.23.1
```

If you are not using one of those mariadb versions (10.11, 10.6, 10.5) you'll need to migrate to one as one to get the newer version of `mysql` in those versions. 

### Expected Behavior

Uninterrupted db file import

### Actual Behavior

```
Error:

ddev drush sql-cli < dump_file.sql
> ERROR at line 1: Unknown command '\-'.

In Process.php line 270:

  The command "mysql --defaults-file=/tmp/drush_Bh6glU --database=db --host=db --port=3306 -A" failed.

  Exit Code: 1(General error)

  Working directory: /var/www/html/web

  Output:
  ================


  Error Output:
  ================
  ERROR at line 1: Unknown command '\-'.
```

### Steps To Reproduce

Import db dump (from older mysql client) using:

`ddev import-db src={PATH_TO_DB_FILE}`

or

`ddev drush sql-cli {PATH_TO_DB_FILE}`

For older mysql clients, the following line is inserted in the db dump file, causing the above error to occur.

```
/*!999999\- enable the sandbox mode */ 
```

### Anything else?

The issue is described here: https://mariadb.org/mariadb-dump-file-compatibility-change/

rfay:
Hi @ioanma - MariaDB last week changed their dump format in a patch version of the product. 

This is discussed a number of places, especially
* https://github.com/ddev/ddev/issues/6083#issuecomment-2116597297
* https://mariadb.org/mariadb-dump-file-compatibility-change/
* https://stackoverflow.com/questions/78530098/ddev-import-db-error-unknown-command

The bottom line is that the problem is in your sql dump, you'll have to use a workaround for now, given in one of those places. 

It's quite surprising and concerning that a vendor would make a massive and incompatible change like this in a patch version of the product. 

I'll leave this issue open as it will have a better title the the other discussion.

[...]

PR #6253, opened 2024-05-28
fix: ImportDB() and ddev import-db should remove new mariadb-dump line, for #6249

rfay:
## The Issue

* #6249 

The new (formerly illegal) line at the top of some mariadb-dump files has
```
/*!999999\- enable the sandbox mode */
```

This breaks so many things all over the internet. It breaks `ddev import-db` and etc. And it breaks our TestDebugMigrateDatabase test in HEAD now that we are supporting the new versions of MariaDB, via the immediate fix in https://github.com/ddev/ddev/issues/6249

Remove the offending line from the import when importing via app.ImportDB. We already remove a couple of other lines, so might as well remove this one also.

* With a mariadb:10.11 project, `ddev debug migrate-database mysql:8.0`
* With a dump that contains the offending line, `ddev import-db`

## Automated Testing Overview

The automated tests caught this problem, and they should be good enough with it.

It's worth considering adding a dump that contains it instead of just the single test that pointed this out, TestDebugMigrateDatabase

## Release/Deployment Notes

In the next release, this will make things easier for lots of people, so they don't have to deal with this new complexity.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -691,7 +691,10 @@ func (app *DdevApp) ImportDB(dumpFile string, extractPath string, progress bool,
 		return err
 	}
 	// The Perl manipulation removes statements like CREATE DATABASE and USE, which
-	// throw off imports. This is a scary manipulation, as it must not match actual content
+	// throw off imports.
+	// It also removes the new `/*!999999\- enable the sandbox mode */` introduced in
+	// LTS versions of MariaDB 2024-05.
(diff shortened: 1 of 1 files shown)


rfay:
@stasadev I'm going to pull this to get HEAD tests one step closer to working (and of course it helps with our mariadb-backup troubles).  Please do review it after the fact though, thanks!

stasadev (commented):
Tested the change, `ddev debug migrate-database` works again. Thank you!
I also successfully imported the dump from `mariadb:10.11` to `mariadb:10.4`.

PR #6256, opened 2024-05-29
docs: Initial draft for adding a quickstart for Sulu, fixes #6230

### LINKED ISSUE #6230: Create Sulu CMS Quickstart

[Sulu](https://sulu.io/) is a Symfony-based CMS which could very well profit from using DDEV.

### Describe your solution

Let's start by creating a PR for a quickstart.

### Describe alternatives

_No response_

### Additional context

_No response_

rpkoller:
## The Issue

* #6230

Review the rendered content at https://ddev--6256.org.readthedocs.build/en/6256/users/quickstart/#sulu

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -626,9 +626,7 @@ ddev config --project-type=php --docroot=public --upload-dirs=uploads --database
 ddev start
 ddev composer create sulu/skeleton
 ```
-
-
-The content management part of Sulu is built upon webspaces. Each webspace configures a content tree. Each content tree may contain translations for different locales. Create your default webspace configuration `mv config/webspaces/example.xml config/webspaces/my-sulu-site.xml` and adjust the values for `<name>` and `<key>` so that they are matching your project:
+Create your default webspace configuration `mv config/webspaces/example.xml config/webspaces/my-sulu-site.xml` and adjust the values for `<name>` and `<key>` so that they are matching your project:
 
 ```bash
 <?xml version="1.0" encoding="utf-8"?>
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -622,7 +622,7 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like 
 
 ```bash
 mkdir my-sulu-site && cd my-sulu-site
-ddev config --project-type=php --docroot=public
+ddev config --project-type=php --docroot=public --upload-dirs=uploads
 ddev start
 ddev composer create sulu/skeleton
 ```
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -643,7 +643,7 @@ Create your default webspace configuration `mv config/webspaces/example.xml conf
 !!!warning "Caution"
     Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
 
-The information for the database connection is set in the environment variable `DATABASE_URL`. The installation should have already created a `.env.local` file which allows you overwrite the `DATABASE_URL` if no `.env.local` exists create it and define also the `APP_ENV` variable:
+The information for the database connection is set in the environment variable `DATABASE_URL`. The installation will have created a `.env.local` file.  Set `DATABASE_URL` in the `.env.local` file so it looks like this:
 
 ```bash
 APP_ENV=dev
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -661,8 +661,7 @@ The content management part of Sulu is built upon webspaces. Each webspace confi
 The information for the database connection is set in the environment variable `DATABASE:URL`. For development, clone the `.env` file by `cp .env .env.local` and change the environment variable to:
 
 ```bash
-DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0.32&charset=utf8mb4"
-```
+DATABASE_URL="mysql://db:db@db:3306/db?charset=utf8mb4"
 
 Now build the database. Building with the `dev` argument adds a user `admin`with the the password `admin` to your project.
 
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -618,6 +618,63 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like 
     ddev launch /cp
     ```
 
+## Sulu
+
+```bash
+mkdir my-sulu-site && cd my-sulu-site
+ddev config --project-type=php --docroot=public
+ddev start
+ddev composer create sulu/skeleton
+```
+
+!!!tip "Multilingual"
+    To set up the administration interface for example in French alongside the default English, add the following snippet on top of `config/packages/sulu_admin.yaml` ([all available localizations](https://sulu.crowdin.com/sulusulu)):
+    ```bash
+    sulu_core:
+    locales:
+        en: English
+        fr: Francais
+    translations:
+        - en
+        - fr
+    ```
+    `ddev exec bin/console sulu:admin:download-language` downloads any set language. The only exceptions are English and German which are shipping with Sulu.
+
+The content management part of Sulu is built upon webspaces. Each webspace configures a content tree. Each content tree may contain translations for different locales. Create your default webspace configuration `mv config/webspaces/example.xml config/webspaces/my-sulu-site.xml` and adjust the values for `<name>` and `<key>` so that they are matching your project:
+
+```bash
+<?xml version="1.0" encoding="utf-8"?>
+<webspace xmlns="http://schemas.sulu.io/webspace/webspace"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://schemas.sulu.io/webspace/webspace http://schemas.sulu.io/webspace/webspace-1.1.xsd">
+    <!-- See: http://docs.sulu.io/en/latest/book/webspaces.html how to configure your webspace-->
+
+    <name>My Sulu CMS</name>
+    <key>my-sulu-cms</key>
+```
+
+!!!warning "Caution"
+    Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
+
+The information for the database conntection is set in the environment variable `DATABASE:URL`. For development simply clone the `.env` file by `cp .env .env.local` and change the environement variable to:
+
+```bash
+DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0.32&charset=utf8mb4"
+```
+
+Now build the database. Building with the `dev` argument adds a user `admin`with the the password `admin` to your project.
+
+```bash
+ddev exec bin/adminconsole sulu:build dev
+ddev launch /admin
+```
+
+!!!tip
+    If you don't want to add an admin user use the `prod` argument instead
+    ```bash
+    ddev execute bin/adminconsole sulu:build prod
+    ```
+
 ## Symfony
 
 There are many ways to install Symfony, here are a few of them based on the [Symfony docs](https://symfony.com/doc/current/setup.html).
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -658,7 +658,7 @@ The content management part of Sulu is built upon webspaces. Each webspace confi
 !!!warning "Caution"
     Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
 
-The information for the database connection is set in the environment variable `DATABASE:URL`. For development, clone the `.env` file by `cp .env .env.local` and change the environment variable to:
+The information for the database connection is set in the environment variable `DATABASE_URL`. The installation should have already created a `.env.local` file which allows you overwrite the `DATABASE_URL` if no `.env.local` exists create it and define also the `APP_ENV` variable:
 
 ```bash
 DATABASE_URL="mysql://db:db@db:3306/db?charset=utf8mb4"
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -661,7 +661,8 @@ The content management part of Sulu is built upon webspaces. Each webspace confi
 The information for the database connection is set in the environment variable `DATABASE_URL`. The installation should have already created a `.env.local` file which allows you overwrite the `DATABASE_URL` if no `.env.local` exists create it and define also the `APP_ENV` variable:
 
 ```bash
-DATABASE_URL="mysql://db:db@db:3306/db?charset=utf8mb4"
+APP_ENV=dev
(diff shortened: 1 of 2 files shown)


docs/content/users/quickstart.md
```diff
@@ -607,6 +607,66 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+```
+
+!!!warning "Caution"
+    Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
+
+The information for the database connection is set in the environment variable `DATABASE:URL`. For development, clone the `.env` file by `cp .env .env.local` and change the environment variable to:
+
+```bash
+DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0.32&charset=utf8mb4"
+```
```
rfay, line 654: This section should go earlier I think. People need to know this at the beginning. Maybe we can give them a standard way to do this with an `echo` or something.

We don't want to put an explicit patch version of mysql in there, it will break often. If we used `--database=mysql:8.0` above that does the job.

```suggestion
```bash
DATABASE_URL="mysql://db:db@db:3306/db?charset=utf8mb4"
```
```

File: docs/content/users/quickstart.md [hunk 2]
```diff
@@ -607,6 +607,66 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
     ddev launch /cp
     ```
 
+## Sulu
+
+```bash
+mkdir my-sulu-site && cd my-sulu-site
+ddev config --project-type=php --docroot=public
```
rfay, line 614:
```suggestion
ddev config --project-type=php --docroot=public --database=mysql:8.0
```

rfay (changes requested):
Thanks for doing this. I took a first quick look through, and invited Sulu folks to do that also.

I think the creation of the .env should be much higher so it doesn't get lost in the detail, and preferably it can be done with a sed/awk/echo/cat or something.

docs/content/users/quickstart.md
rpkoller: in regards of the position. i've followed the order sulu was using which seemed clear and logical to me. first you download the skeleton then you get the optional instructions for multiilingual  and the command how to download those, then you set the name and key which is necessary same as multilingual before building your environment and in the last step you prepare the env file with the database credentials. then you build it. so moving the env file before the other step i am not sure about.

docs/content/users/quickstart.md
rfay: I assumed from your `DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0.32&charset=utf8mb4"
` that mysql 8.0 was somehow important. If it's not, then we can use the defaults.

rpkoller: i first had the same assumption but then took a second look at the docs and realized all the standard db ddev supports are supported by sulu as well

chirimoya:
@alexander-schranz could you please take a look? Thanks.

docs/content/users/quickstart.md
```diff
@@ -607,6 +607,65 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+    <!-- See: http://docs.sulu.io/en/latest/book/webspaces.html how to configure your webspace-->
+
+    <name>My Sulu CMS</name>
+    <key>my-sulu-cms</key>
+```
+
+!!!warning "Caution"
+    Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
+
+The information for the database connection is set in the environment variable `DATABASE:URL`. For development, clone the `.env` file by `cp .env .env.local` and change the environment variable to:
```
alexander-schranz, line 650: This line is for all Symfony based applications not correct the `.env`. You never do `cp .env .env.local` in Symfony based applicatoins as the `.env` is always loaded. You only write into the `.env.local` the things you want to change so in this case DATABASE_URL is only required to be added the `.env.local` normally already exists as the composer create project already set `APP_ENV=dev` for quick development start by Sulu.

```suggestion
The information for the database connection is set in the environment variable `DATABASE_URL`. The installation should have already created a `.env.local` file which allows you overwrite the `DATABASE_URL` if no `.env.local` exists create it and define also the `APP_ENV` variable:
```

docs/content/users/quickstart.md
```diff
@@ -607,6 +607,65 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+
+```bash
+DATABASE_URL="mysql://db:db@db:3306/db?charset=utf8mb4"
```
alexander-schranz, line 653: Missing "```"

alexander-schranz, line 653: All Symfony or even doctrine based applications should define the server_version else you run into unexpected errors:

```suggestion
```bash
APP_ENV=dev
DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0.32&charset=utf8mb4"
```

docs/content/users/quickstart.md
```diff
@@ -607,6 +607,65 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+
+Now build the database. Building with the `dev` argument adds a user `admin`with the the password `admin` to your project.
+
+```bash
+ddev exec bin/adminconsole sulu:build dev
+ddev launch /admin
```
alexander-schranz, line 659: what kind of Webserver is DDEV using do we need adjust something to support the media download and image generation. We have docs for [apache](https://docs.sulu.io/en/latest/cookbook/web-server/apache.html) and [nginx](https://docs.sulu.io/en/latest/cookbook/web-server/nginx.html).

alexander-schranz: Beside the `var` directory which need to writeable in any Symfony application in dev. The `public/uploads` also need to writeable and in our case special vhost routing.

The Symfony CLI Webserver does not require special VHost the PHP Internal webserver requires a special routing file `php -S 127.0.0.1:8000 -t public config/router.php`.

alexander-schranz (changes requested):

docs/content/users/quickstart.md
rpkoller: at first thank you for the review! and uh good point, cloning is not the right way to handle things here since `.env.local` is just overriding, that was my fault. 
but one question when i am applying the steps in the current draft `.env.local` is not created automatically by `ddev composer create sulu/skeleton` only `.env`, `.env.stage`, and `.env.test` is being created. To stay directly actionable and under the assumption that the `.env.local` file is not being created yet, how about using something like that instead of your second sentence? 

```
For development, override the `APP_ENV` and `DATABASE_URL` variables in the `.env.local` file with:

echo -n "APP_ENV=local\nDATABASE_URL=mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7" >> .env.local
```

docs/content/users/quickstart.md
rpkoller: argh good catch, thank you!

docs/content/users/quickstart.md
rpkoller: with ddev you are able to use apache as well as nginx, per default nginx is used. i've also pushed another commit to add `public/uploads` as the upload folder (are there more? - those are ignored by mutagen).

in regards of the other details you've raised i better hand over to @rfay and/or @stasadev, myself i am not a developer.

docs/content/users/quickstart.md
rpkoller: as mentioned in one other comment i am not a developer and was following the recommendation by @rfay who raised 

> We don't want to put an explicit patch version of mysql in there, it will break often. If we used --database=mysql:8.0 above that does the job.

in https://github.com/ddev/ddev/pull/6256#discussion_r1625078826 . with the serverversion in the user would have to keep the version version updated in case the database image is getting updated in a new version of ddev?

docs/content/users/quickstart.md
rpkoller: and one other question @alexander-schranz in regards of multilingual. in the example i went with english in french to include one language aside the two sulu is shipping with. but with the snippet used in the quickstart i still get: 

```

Language: de
------------

 ! [NOTE] Download skipped because Sulu includes a translation for the language "de" per default.


Language: en
------------

 ! [NOTE] Download skipped because Sulu includes a translation for the language "en" per default.
```
when running `ddev exec bin/console sulu:admin:download-language` , why `de` is skipped? in the `sulu_admin.yaml` only english and french is set?

docs/content/users/quickstart.md
alexander-schranz: @rpkoller does ddev skip the composer create project scripts? Its create by the basic composer root install part:

https://github.com/sulu/skeleton/blob/6f9ff17f8862c482a5a4155f916b22b7db216112/composer.json#L120

That part should not be skipped, I recommend that for any framework / system:


sulu-create-project

docs/content/users/quickstart.md
alexander-schranz: `de` and `en` is shipped via `vendor/sulu/sulu` so there is nothing to download as only none core languages need to be download from https://sulu.crowdin.com/

docs/content/users/quickstart.md
rpkoller: oh that is interesting. it looks like that some of the project scripts are skipped and not executed when running  `ddev composer create sulu/skeleton`. here is the output of the first two initial steps: https://gist.github.com/rpkoller/f4be5ff370afd60fe972b70275deddbd . those steps after the project is being create are missing. for example the app secret in the .env file is also not being replaced. i guess that bug in ddev first has to be fixed,  cuz it makes a few of the steps in the draft obsolete.

stasadev: There is some problem with Mutagen sync in DDEV v1.23.1, `.env.local` is not created on `ddev composer create sulu/skeleton`

But @rfay recently made some changes for Mutagen, I tried `ddev composer create sulu/skeleton` with DDEV HEAD, and this problem is fixed, `.env.local` will be there is DDEV v1.23.2.

docs/content/users/quickstart.md
alexander-schranz: yes but a project normally should not get automatically a new minor version, as doctrine did in past behave differently based on minor versions (e.g. JSON was stored as TEXT in older mariadb/mysql versions) and if it does not match your "production" version you even could accidently create false databasemigrations which do work locally but not on online server.

docs/content/users/quickstart.md
rpkoller: i am on the latest version of HEAD (ddev version v1.23.0-84-g213f46972 - just tested if there was another commit). so the output in https://gist.github.com/rpkoller/f4be5ff370afd60fe972b70275deddbd shows that those scripts are  still not executed unfortunately. and i am also unsure how mutagen could interfere with the scripts not being executed?

alexander-schranz: The problem is that somebody add `--no-scripts` see output line 24:

```
Executing Composer command: [composer create-project sulu/skeleton --no-plugins --no-scripts --no-install /tmp/UPPlPA]
```

rpkoller: one script is executed at a later point `composer run-script post-create-project-cmd` but not all scripts in general https://gist.github.com/rpkoller/f4be5ff370afd60fe972b70275deddbd#file-2-ddev-composer-create-txt-L543 aside line 536-540

alexander-schranz: The `post-root-package-install` should be executed before `Executing Composer command: [composer install]` (L54)

stasadev: We have `composer run-script post-root-package-install` script instruction (before `composer install`), but I tested this on Linux w/ and w/o Mutagen.

I'll definitely have to check it out again on macOS.

rpkoller: the setup the gist is created with is macos 14.5 running on lima with VZ

stasadev: I reproduced it with macOS, and also with Linux (I don't know why I couldn't the first time, maybe I mixed the binaries)

I opened this:

- #6281

rogoit:
Tested and looks good 4 me. Randy add my feedback. Thx for your passion.

docs/content/users/quickstart.md
rfay: Database version is updated with DDEV versions; only the major version (like 8.0, 8.4 for MySQL, and 10.11, 11.4 for MariaDB, or `14` or `15` for PostgreSQL) is available. Maintaining a patch version in settings code won't work out in the long run (or the short run :) )

rfay: Minor rewrite of @alexander-schranz suggestion, without MySQL patch version.

```suggestion
APP_ENV=dev
DATABASE_URL="mysql://db:db@db:3306/db?serverVersion=8.0&charset=utf8mb4"
```

docs/content/users/quickstart.md
rfay: It looks to me like the default nginx config is working fine for displaying images:

image

rfay (changes requested):
It's getting close. I do think it may be too verbose still, we can probably get rid of a little more of the detail.

docs/content/users/quickstart.md
```diff
@@ -607,6 +607,67 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+            en: English
+            fr: Francais
+        translations:
+            - en
+            - fr
+    ```
+
+    `ddev exec bin/console sulu:admin:download-language` downloads any set language. The only exceptions are English and German which are shipping with Sulu.
+
+The content management part of Sulu is built upon webspaces. Each webspace configures a content tree. Each content tree may contain translations for different locales. Create your default webspace configuration `mv config/webspaces/example.xml config/webspaces/my-sulu-site.xml` and adjust the values for `<name>` and `<key>` so that they are matching your project:
```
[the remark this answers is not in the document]
rpkoller: i've cut the first three sentences

docs/content/users/quickstart.md
```diff
@@ -607,6 +607,67 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
+    <!-- See: http://docs.sulu.io/en/latest/book/webspaces.html how to configure your webspace-->
+
+    <name>My Sulu CMS</name>
+    <key>my-sulu-cms</key>
+```
+
+!!!warning "Caution"
+    Changing the `<key>` for a webspace later on causes problems. It is recommended to decide on the value for the key before the database is build in the next step.
+
+The information for the database connection is set in the environment variable `DATABASE_URL`. The installation should have already created a `.env.local` file which allows you overwrite the `DATABASE_URL` if no `.env.local` exists create it and define also the `APP_ENV` variable:
```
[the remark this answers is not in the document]
rpkoller: how about? 
```suggestion
The information for the database connection is set in the environment variable `DATABASE_URL`. The installation will create a `.env.local` file. Add the `DATABASE_URL` that the `.env.local` file looks like this:
```

rpkoller:
the tests would have to be restarted. the linter failed on one "dead" link (which was a false report)

rfay (approved):
This looks great and tests great, thanks all!

PR #6257, opened 2024-05-29
fix: Make export-db also remove new mariadb directive in dumpfile , fixes #6249

### LINKED ISSUE #6249: `ddev import-db` fails with "Error: Unknown command '\-'" because of new directive in mariadb-dump output

rfay:
## The Issue

* #6249

We've already done a lot to mitigate this. 
* Pushed updated ddev-dbserver images for v1.23.1 that have the new client binaries so folks on v1.23.1 can at least get by
* https://github.com/ddev/ddev/pull/6253 fixed import-db so it wouldn't stumble on the new line

But we still had trouble of course *pushing* to remote. So TestPantheonPush failed for all the same reasons.

Rather than adding workarounds to our push tests, I figured I should just make ExportDB() and `ddev export-db` also remove the offending line, which should fix all of our Push  tests. 

ExportDB also removes the new mariadb line

Use `ddev export-db` with a new-style mariadb (10.11) and verify that the dump file does not have the offending directive.

## Automated Testing Overview

I think all the tests stand on their own. I had to run many of them to sort out all the issues.

## Release/Deployment Notes

The whole class of problem will be noted in release notes.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -778,22 +778,33 @@ func (app *DdevApp) ImportDB(dumpFile string, extractPath string, progress bool,
 // targetDB is the db name if not default "db"
 func (app *DdevApp) ExportDB(dumpFile string, compressionType string, targetDB string) error {
 	app.DockerEnv()
-	exportCmd := []string{"mysqldump"}
-	if app.Database.Type == "postgres" {
-		exportCmd = []string{"pg_dump", "-U", "db"}
-	}
 	if targetDB == "" {
 		targetDB = "db"
 	}
-	exportCmd = append(exportCmd, targetDB)
 
-	if compressionType != "" {
-		exportCmd = []string{"bash", "-c", fmt.Sprintf(`set -eu -o pipefail; %s | %s`, strings.Join(exportCmd, " "), compressionType)}
+	exportCmd := "mysqldump " + targetDB
+	if app.Database.Type == "postgres" {
+		exportCmd = "pg_dump -U db " + targetDB
 	}
+
+	if app.Database.Type == nodeps.MariaDB {
+		// The `tail +2` is a workaround that removes the new mariadb directive added
+		// 2024-05 in mariadb-dump. It removes the first line of the dump, which has
+		// the offending /*!999999\- enable the sandbox mode */. See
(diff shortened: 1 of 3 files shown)


rfay:
> As far as I understand, the issue that MariaDB developers were trying to fix was never an issue for DDEV because the database runs in an isolated Docker container and the system command doesn't have host-side access anyway, so it's okay to remove their fix completely.

I believe that's true, but OTOH now that we're doing dumps that don't have their magic lines, those dumps could be used to attack a machine where the dumps were imported, for example with a `ddev push platform`

rfay:
I'm going to pull this even though it's not perfect yet (TestDdevAllDatabases fails when the whole test gets run).  I want to make sure that none of the things I'm chasing in #6258 is related.

PR #6263, opened 2024-05-30
feat: don't override ddev-generated built-in project files created by add-on, fixes #6173, fixes #6190

stasadev:
## The Issue

Add-ons can't properly manage built-in DDEV resources because when you remove `#ddev-generated` the file is no longer "installed" with the add-on. When you remove the add-on the file without the `#ddev-generated` is not removed. But when you leave out `#ddev-generated`, the file is replaced with DDEV:

- #6173

Built-in DDEV assets are copied every time:
- #6190

## How This PR Solves The Issue

Do not override bundled assets in the project directory if they are listed in add-on manifest in `project_files` and have `#ddev-generated`.

This will allow to override files generated by DDEV in `.ddev` directory, by moving control over them to the add-on.

And do not copy the embedded assets if their content is the same.

## Manual Testing Instructions

```
mkdir test-pr && cd test-pr
ddev config --auto
mkdir -p test-addon/xhprof && printf "name: test-addon\nproject_files:\n- xhprof/xhprof_prepend.php\n" > test-addon/install.yaml && printf "<?php\n#ddev-generated\n" > test-addon/xhprof/xhprof_prepend.php
ddev get test-addon
ddev start
```

Check the contents of `.ddev/xhprof/xhprof_prepend.php`.

PR #6266, opened 2024-05-31
build: bump upload-artifact to v4 and fix mkcert.exe in tarball/zipball [skip ci]

rfay:
## The Issue

* upload-artifact is deprecated and needs to bump to v4
* mkcert.exe in tarball/zipball is wrong.

### DIFF
--- a/.ci-scripts/generate_artifacts.sh
+++ b/.ci-scripts/generate_artifacts.sh
@@ -73,7 +73,7 @@ popd >/dev/null
 
 # generate windows-amd64 tarball/zipball
 pushd $BASE_DIR/.gotmp/bin/windows_amd64 >/dev/null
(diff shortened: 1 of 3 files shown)


stasadev:
I will test it, the previous attempt to bump `upload-artifact@v4` was not successful:

- #5662

stasadev (approved):
It works, thank you!

rfay:
They're really putting the pressure on to upgrade now, so I guess we'll try it. We'll back it out if we fail again. I thought there had been a previous attempt.

stasadev:
Tested both changes with #6263

The artifacts are fine and the `mkcert.exe` binary inside `all-ddev-executables.zip` is correct.

rfay:
Thanks!

PR #6269, opened 2024-06-01
test: Don't use php package in TestExtraPackages

rfay:
## The Issue

Failures from packages.sury.org have broken our tests for a day or so, and this has happened before. 

TestExtraPackages doesn't really need to test a PHP package, so at least for now it would be nice to move forward, so using a non-deb.sury.org package there.

* https://github.com/oerdnj/deb.sury.org/issues/2046#issuecomment-2143131325

Use a non-deb.sury.org package for TestExtraPackages.

### DIFF
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -3,7 +3,6 @@ package ddevapp_test
 import (
 	"bufio"
 	"fmt"
-	copy2 "github.com/otiai10/copy"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -11,6 +10,8 @@ import (
 	"sort"
 	"strings"
 	"testing"
+
+	copy2 "github.com/otiai10/copy"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/ddev/ddev/pkg/ddevapp"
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/config_test.go
```diff
@@ -986,9 +987,7 @@ func TestExtraPackages(t *testing.T) {
 	require.NoError(t, err)
 
 	addedDBPackage := "sudo"
-	// php-gmp is flaky on Debian 12 Bookworm,
-	// we can test it https://github.com/ddev/ddev/issues/5898
-	addedWebPackage := "php" + app.PHPVersion + "-" + "gmp"
```
rfay, line 991: @stasadev I know you put this here for a specific reason when we were having trouble with php-gmp. If you think we should still be testing that situation we should figure out a safer way to test it, probably not in this test.

rfay:
I'm going to pull this in the interest of getting green tests that we can build from, but looking forward to any input you may have on Monday @stasadev .

pkg/ddevapp/config_test.go
stasadev: I don't think we will remove the file permissions workaround for the PHP pid file in `/run/php` because it has proven to work well over time.

And the addition of `php-gmp` was done solely to test this workaround. When a test fails due to an unstable connection to deb.sury.org, I don't mind changing the test package here.

PR #6270, opened 2024-06-03
refactor: make "waiting for" message more generic

tyler36:
## The Issue

On a test project, I disabled the database using:

```yml
omit_containers: 
- "db"
```

When I started the project, I saw:

```
$ ddev start
...
Waiting for web/db containers to become ready: [web] 
```

This (nitpic) PR makes the output more generic:

```diff
- Waiting for web/db containers to become ready: [web] 
+ Waiting for containers to become ready: [web] 
```

The line continues to show container names ('web') at the end.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1423,7 +1423,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	if !nodeps.ArrayContainsString(app.GetOmittedContainers(), "db") {
 		dependers = append(dependers, "db")
 	}
-	output.UserOut.Printf("Waiting for web/db containers to become ready: %v", dependers)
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -1423,7 +1423,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	if !nodeps.ArrayContainsString(app.GetOmittedContainers(), "db") {
 		dependers = append(dependers, "db")
 	}
-	output.UserOut.Printf("Waiting for web/db containers to become ready: %v", dependers)
+	output.UserOut.Printf("Waiting for containers to become ready: %v", dependers)
```
rfay, line 1426: I think this is fine. The reason this was stated the way it was is that we do two waits, the first for core containers, and the second for anything else.

rfay (approved):

PR #6272, opened 2024-06-04
build: Use github instead of dl.filippo.io to get mkcert binary

rfay:
## The Issue

Several add-on tests were failing on HEAD last night apparently because dl.filippo.io was broken.

Makefile should use the github repo as source for `mkcert`.

The reality is that the homebrew formula already installs mkcert as a dependency, so at least in the add-on tests, this is pretty much irrelevant.


stasadev (approved):
Looks good to me, thank you!

PR #6277, opened 2024-06-05
docs: fix mutagen example command

stasadev:
## The Issue

The example command is missing `ddev` at the start, `jq` may not be available.

Fixes it.


rfay (approved):
Thanks! I actually left the docker part out on purpose (or laziness) figuring that anybody who got in this deep would know how to use/get jq. But this is an improvement!

Issue #6276, opened 2024-06-05
Remove backticks from `ddev mailpit` in `ddev describe` info text


gitressa:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Problem

Under `ddev describe` the Mailpit help text is shown as `` `ddev mailpit` ``, using backticks, which don't really add anything.

```
$ ddev describe
┌─────────────────────────────────────────────────────────────────────────────────────────┐
│ Project: solr-test ~/dev/solr-test https://solr-test.ddev.site                          │
│ Docker platform: linux-docker                                                           │
│ Router: traefik                                                                         │
├──────────┬──────┬─────────────────────────────────────────────────┬─────────────────────┤
│ SERVICE  │ STAT │ URL/PORT                                        │ INFO                │
├──────────┼──────┼─────────────────────────────────────────────────┼─────────────────────┤
[... ]
├──────────┼──────┼─────────────────────────────────────────────────┼─────────────────────┤
│ Mailpit  │      │ Mailpit: https://solr-test.ddev.site:8026       │                     │
│          │      │ `ddev mailpit`                                  │                     │
├──────────┼──────┼─────────────────────────────────────────────────┼─────────────────────┤
```

### Solution

Update `` `ddev mailpit` `` and remove the backticks, to be just `ddev mailpit`.

stasadev:
I think we have these back ticks to make it clear that this is a command that you can run and not something else.

We actually use all of these in the DDEV output, depending on the developer's preference 🙂 :
- `
- "
- '

gitressa:
Thanks for a fast response @stasadev, it's great.

But in my opinion, the backticks (or any other indication) are not needed. From the context, the user will understand the meaning. If anything (if help is needed) it could be something like this?

> Launch: ddev mailpit

stasadev:
This is a good suggestion, I like it. Feel free to contribute it with PR.

gitressa:
Thanks! I created a PR. 

[...]

PR #6280, opened 2024-06-05
docs: update PR template

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/6276#issuecomment-2149679514

The template could be better.

Removes `## Related Issue Link(s)` and adds boilerplate for the issue number at the top.

I'm not using `## Related Issue Link(s)` section, it's too far away, so people reviewing the PR might not even notice it.
The best place for this is at the top.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,10 @@
 
 ## The Issue
 
+- #<issue number>
+
+<!-- Provide a brief description of the issue. -->
+
 ## How This PR Solves The Issue
 
 ## Manual Testing Instructions
@@ -13,9 +17,6 @@
 
 <!-- Please describe the tests introduced by this PR, or explain why no tests are needed. -->
 
-## Related Issue Link(s)
-
 ## Release/Deployment Notes
 
 <!-- Does this affect anything else or have ramifications for other code? Does anything have to be done on deployment? -->
-


rfay (approved):
Thanks!

gitressa:
This is a nice little improvement, thanks! Perhaps update it to this format?

`- Fixes #<issue number>`

Otherwise, the issue will not be automatically linked in the right hand side, here:

> **Development**
> Successfully merging this pull request may close these issues.

It could also be "Closes", or some other trigger word, see [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

See this issue and PR for an example:
- #6306
- #6307

stasadev:
We use `, fixes #<number>` in the title, according to https://ddev.readthedocs.io/en/stable/developers/building-contributing/#pull-request-title-guidelines

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,6 +5,10 @@
 
 ## The Issue
 
+- #<issue number>
```
gitressa, line 8: Maybe update to `- Fixes #<issue number>` here?

gitressa:
Sure, but I think adding it in the PR template does no harm, and makes sure that PR and issue are connected.

.github/PULL_REQUEST_TEMPLATE.md
stasadev: I don't know, it's not always the situation when the PR fixes something.
From what I see for the PRs I make, I usually add a link here to the issue or another PR or Discord.

But the title must have `fixes` in it if possible, which I find very useful because I use "blame" feature in GitLens with VSCode and in GitToolbox with JetBrains, and when I hover over any line, I see the commit author and the message, and I instantly know which issue/PR is associated with the change I'm looking at.

stasadev:
> Successfully merging this pull request may close these issues

I didn't know about it, maybe because I didn't look to the right (ads are usually on the sides of websites and I instinctively refused to read).

gitressa:
All right, so what do you think about my suggestion now?

stasadev:
If it can improve GitHub's PR interface, I'm fine with it.

gitressa:
Sounds good, thanks!

PR #6281, opened 2024-06-05
fix: Composer post-root-package-install didn't work with Mutagen enabled

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/6256#discussion_r1627427307

`.env.local` is not created in the Sulu project quickstart, but it should.
`ddev composer create` doesn't run `post-root-package-install` composer script because Mutagen doesn't perform a sync.

Adds additional `app.MutagenSyncFlush()` right before checking for `composer.json`

```
ddev config global --performance-mode=mutagen

mkdir my-sulu-site && cd my-sulu-site
ddev config --project-type=php --docroot=public --upload-dirs=uploads
ddev composer create sulu/skeleton

# see this in the output:
Executing composer command: [composer run-script post-root-package-install]

# and check if the file exists, it should:
ls .env.local
```

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -164,6 +164,12 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			util.Failed("failed to create project: %v", err)
 		}
 
+		// Make sure composer.json is here with Mutagen enabled
+		err = app.MutagenSyncFlush()
+		if err != nil {
+			util.Failed("Failed to flush Mutagen: %v", err)
+		}
+
 		composerManifest, _ := composer.NewManifest(path.Join(composerRoot, "composer.json"))
 
 		if !preserveFlags && composerManifest != nil && composerManifest.HasPostRootPackageInstallScript() {


rpkoller:
after extracting the skeleton archive and before stopping and restarting the containers to the install the dependencies 

```
Executing composer command: [composer run-script post-root-package-install]

> @php -r "file_put_contents('.env.local', 'APP_ENV=dev' . PHP_EOL);"
> @php -r "file_put_contents('.env', str_replace('APP_SECRET=\'\$ecretf0rt3st\'', 'APP_SECRET=' . bin2hex(random_bytes(16)), file_get_contents('.env')));"
 Container ddev-testinger-db  Stopped
```
is run. i've also checked, `.env.local` got created containing the app env variable with the `dev` value and an app secret is correctly set in the `.env` file. before it stuck to the default value. looks good, thanks for the fast fix!

rpkoller (approved):

stasadev:
I checked our quickstarts for `post-root-package-install`:

Found it only in `laravel/laravel` and `statamic/statamic`

`https://github.com/laravel/laravel/blob/3b3f9f13faab1753e7b7cad6a0e7098e54c8199f/composer.json#L40-L42

https://github.com/statamic/statamic/blob/c5a6394575223b1cea893f55b00dfde3bf660a6f/composer.json#L48-L50

All it does is copy `.env.example` to `.env`, which is also done in DDEV, so I think it won't break anything.

rpkoller:
yeah somehow statamic isnt affected that post root package install wasnt run. the .env file is created anyway including all the setting set in the following. a bit surprising. but i'Ve quickly tested statamic once with the patch applied once without. in both cases it worked. in the case of sulu it had more consequences.

stasadev:
> somehow statamic isnt affected

Yes, that's because statamic quickstart is based on the laravel recipe, and DDEV handles it.

rpkoller:
out of curiosity what happens with the binary where the patch is applied. does composer create the .env file and then ddev overwrites that .env file with the same or does ddev skip the creation of the .env file because it is already in place?

stasadev:
This one:

> ddev skip the creation of the .env file because it is already in place

`fileutil.CopyFile()` only if `os.IsNotExist(err)`

https://github.com/ddev/ddev/blob/ae65b58c2bc42bbe823a9bdda03675780392bd07/pkg/ddevapp/laravel.go#L30-L36

rpkoller:
oh that looks perfect! awesome, thanks

rfay (approved):
There's nothing dangerous about this, sounds good to me.

Normally the `ddev composer create` process does its sync with a restart, so not exactly sure how this comes out differently.

The one thing you might consider is adding a test to make sure the post-root-package-install works. Add to one of the composer tests or create a new one?

PR #6285, opened 2024-06-05
test: fix TestFileHash on Windows, fixes #6284

### LINKED ISSUE #6284: TestFileHash is broken on Traditional Windows
### Expected Behavior

TestFileHash should work everywhere. But it's not working. https://buildkite.com/ddev/ddev-windows-mutagen/builds/4611

If Mutagen is working fine on traditional Windows, we can just disable TestFileHash on Windows. Or we can fix it.

stasadev:
## The Issue

- #6284

`TestFileHash` is broken on Windows due to different slashes in the filepath.

Uses `filepath.ToSlash()` to normalize behavior for different OSs.

### DIFF
--- a/cmd/ddev/cmd/auth-ssh.go
+++ b/cmd/ddev/cmd/auth-ssh.go
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
-	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/exec"
 	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/nodeps"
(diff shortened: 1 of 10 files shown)


rfay (approved):
Wow thanks!

rfay:
Looks like this is not quite there yet, https://buildkite.com/ddev/ddev-windows-mutagen/builds/4614#018fea0a-1cfb-4964-8cf0-4a92009027fc

stasadev:
I debugged it on Windows and the actual error is:

```
sha1sum: can't open '/c/Users/testbot/tmp/ddevtest/TestFileHash_nUSXBdMSAm4228215173/TestFileHash_0
```

It doesn't want to mount the directory. I tried to mount a volume in git-bash and I had success with `MSYS_NO_PATHCONV=1`

```
mkdir -p /c/Users/testbot/tmp/ddevtest/TestFileHash && touch /c/Users/testbot/tmp/ddevtest/TestFileHash/test
MSYS_NO_PATHCONV=1 docker run -i --rm -v /c/Users/testbot/tmp/ddevtest/TestFileHash:/c/Users/testbot/tmp/ddevtest/TestFileHash busybox sha1sum /c/Users/testbot/tmp/ddevtest/TestFileHash/test
da39a3ee5e6b4b0d3255bfef95601890afd80709  /c/Users/testbot/tmp/ddevtest/TestFileHash/test
```

But it doesn't work inside the test, no matter how I try to set it:
```go
_ = os.Setenv("MSYS_NO_PATHCONV", "1")
t.Setenv("MSYS_NO_PATHCONV", "1")
```

`GOTEST_SHORT=true make test TESTARGS="-run TestFileHash"` always fails.

rfay:
I took the liberty of pursuing this. 

I think the bottom line is that both the hash function and the related test need to use a canonical filepath for their work, and it needs to be predictable.

This works for me. 

I imagine that mutagen wasn't working right on Windows either, because it would find a different hash every time and recreate the mutagen sync session. So yay for tests!

rfay:
Manual testing of mutagen on Windows says it's OK with this. I didn't see any mutagen sessions or volumes getting recreated.

rfay:
Yay, it passed the Windows test!

stasadev (commented):
Wow, that's great!

Issue #6083, opened 2024-04-11
Install matching psql/pg_dump client inside ddev-webserver based on postgres version


rfay:
In general, the `mysql` client shipped with the ddev-webserver works fine with all mariadb and mysql back-end databases. 

However, there are a few cases where their behavior is slightly different. 

And the fact that the `mysql` client reports a different version is often confusing to people. 

For example, with database mysql:8.0 we see these outputs:

```
rfay@rfay-mbp-2021:~/workspace/d10$ ddev exec mysql --version
mysql  Ver 15.1 Distrib 10.11.6-MariaDB, for debian-linux-gnu (aarch64) using  EditLine wrapper

rfay@rfay-mbp-2021:~/workspace/d10$ ddev exec -s db mysql --version
mysql  Ver 8.0.33-0ubuntu0.20.04.2 for Linux on aarch64 ((Ubuntu))
```

The web container has the mariadb version, and the db container has the mysql:8.0 version. 

On Craft CMS and perhaps some other places, people have the habit of using the built-in tools; Craft actually calls out to the mysql client (on ddev-webserver) rather than connecting directly to the database server, so this can cause confusion there. 

* https://discord.com/channels/664580571770388500/1227934688505630721/1227934688505630721
* https://discord.com/channels/664580571770388500/1189581731301429318
* https://discord.com/channels/664580571770388500/1160230833601392791/1160258585633247302

### Proposal

Add a layer to the ddev-webserver adding a matching `mysql` client to the ddev-webserver, especially where the database type is mysql:8.0, which is the most likely place for concern.

It also might be possible to implement this as an add-on for the few people it matters to. It's tricky getting the proper match between the OS-provided `mysql` client and the ddev-webserver, and it could introduce instabilities.

stasadev:
Answering to https://github.com/ddev/ddev/issues/6139#issuecomment-2086929352 in the relevant issue:

> Is there something like that where we can get the mysql version of the client also?

There is `mysql-apt-config` https://dev.mysql.com/doc/refman/8.4/en/linux-installation-apt-repo.html

The current latest version is `0.8.30` https://dev.mysql.com/get/mysql-apt-config_0.8.30-1_all.deb

How to install it in non-interactive way https://stackoverflow.com/a/37267411/8097891

rfay:
I'm not sure there's a client package, but if there is this can be great!

rfay:
[...]

PR #6290, opened 2024-06-07
fix: Use mariadb repositories for mariadb-client, for #6083

rfay:
## The Issue

- #6083

mysql and mysqldump in the `ddev-webserver` don't handle mysql's new sandbox directive. 

This installs the 10.11.8-MariaDB of the mariadb client (mysql) and the related mysqldump in the ddev-webserver

We already have matching versions in the ddev-dbserver, but some tools like drush and craft cms's backup tools use local tools in the ddev-webserver.

Try creating database dumps on ddev-webserver and importing them various places

You'll note `ddev exec 'mysqldump db >/tmp/db.sql'` will create a db.sql that has the dreaded `/*!999999\- enable the sandbox mode */`

`ddev import-db` and `ddev export-db` now strip this, but if processing is done by application-level things like drush, your mileage may vary. The sql dump that you create may not be import-able by a different version of mariadb or mysql.

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -82,6 +82,9 @@ RUN curl https://nginx.org/keys/nginx_signing.key | gpg --dearmor \
 RUN echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] \
 http://nginx.org/packages/debian `lsb_release -cs` nginx" > /etc/apt/sources.list.d/nginx.list
 
+# Install mariadb_repo_setup to get mariadb-client from them directly
+RUN curl -LsSf https://r.mariadb.com/downloads/mariadb_repo_setup | bash -s -- --mariadb-server-version="mariadb-10.11"
+
 RUN curl -sSLo /tmp/debsuryorg-archive-keyring.deb https://packages.sury.org/debsuryorg-archive-keyring.deb && \
     dpkg -i /tmp/debsuryorg-archive-keyring.deb && rm -f /tmp/debsuryorg-archive-keyring.deb && \
     echo "deb [signed-by=/usr/share/keyrings/deb.sury.org-php.gpg] https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/php.list && apt-get update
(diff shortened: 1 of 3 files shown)


stasadev (changes requested):
This is great, and even can be better 👍

I researched the error with MariaDB 10.6, it tried to pull from `bookworm` but there is only a repo for `bullseye`.

Let's save `curl -LsSf https://r.mariadb.com/downloads/mariadb_repo_setup` to `/usr/local/bin/mariadb_repo_setup`, add `chmod +x /usr/local/bin/mariadb_repo_setup`, and run `mariadb_repo_setup --mariadb-server-version="mariadb-10.11"` in the `ddev-php-base/Dockerfile`.

We can add this to `RenderComposeYAML()`:

```go
// Install the latest version of MariaDB for https://mariadb.org/mariadb-dump-file-compatibility-change/
if app.Database.Type == nodeps.MariaDB && (app.Database.Version == nodeps.MariaDB104 || app.Database.Version == nodeps.MariaDB105 || app.Database.Version == nodeps.MariaDB106) {
	osVersion := "bookworm"
	if app.Database.Version == nodeps.MariaDB104 || app.Database.Version == nodeps.MariaDB105 {
		osVersion = "buster"
	} else if app.Database.Version == nodeps.MariaDB106 {
		osVersion = "bullseye"
	}
	extraDBContent = extraDBContent + fmt.Sprintf(`RUN (mariadb_repo_setup --mariadb-server-version="mariadb-%s" --os-type=debian --os-version=%s && rm -f /etc/apt/sources.list.d/mariadb.list.old_*) || true`, app.Database.Version, osVersion)
}
```

stasadev (approved):
Oh, I didn't try to install the client.

```
sudo apt-get install mariadb-client

The following packages have unmet dependencies:
 mariadb-client-10.6 : Depends: mariadb-client-core-10.6 (>= 1:10.6.18+maria~deb11) but it is not going to be installed
                       Depends: libssl1.1 (>= 1.1.1) but it is not installable
```

It's a rabbit hole of dependencies, I think we are good now with MariaDB 10.11.

rfay:
https://github.com/ddev/ddev/issues/6083#issuecomment-2155559450 explains the situation here and links to the mariadb issue. I don't think we'll be able to use the package-install technique reliably with multiple mariadb-client versions.

I am in fact working on the same thing for mysql though, and am able to build clients for mysql 5.7, 8.0, 8.4. It seems to matter more with mysql in general. We *could* do the same thing with mariadb-client, but I suspect at this point we don't have to go there.

rfay:
Nothing about this prevents us from going back and adding per-version clients, this just means we start with the latest 10.11.

PR #6291, opened 2024-06-08
build: Bump mysql to 8.0.36, fixes #6250

### LINKED ISSUE #6250: Update mysql:8.0 to highest available patch version (8.0.36?)

xtrabackup is not at 8.0.35, see https://www.percona.com/downloads - so we can at least update mysql 8.0 to that.

rfay:
## The Issue

- #6250
- https://github.com/ddev/mysql-arm64-images/pull/8
- https://github.com/ddev/mysql-arm64-images/pull/9
- https://github.com/ddev/xtrabackup-build/releases/tag/8.0.35-30
- https://perconadev.atlassian.net/browse/PXB-3304

It looks like we can bump to mysql 8.0.36 now, with xtrabackup 8.0.35. 

Update

Note that this updates the mysql:8.0 image to Ubuntu 22.04

Use mysql:8.0 and do snapshots and snapshot restores. 

## Automated Testing Overview

Existing tests may be adequate.

### DIFF
--- a/containers/ddev-dbserver/Makefile
+++ b/containers/ddev-dbserver/Makefile
@@ -10,7 +10,7 @@ CURRENT_ARCH=$(shell ../get_arch.sh)
 # So has to explicitly declare anything it might need from there (like SHELL)
 SHELL = /bin/bash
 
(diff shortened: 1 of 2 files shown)


rfay:
The percona folks confirmed that we can use xtrabackup 8.0.35 with mysql 8.0.36. They also said that arm64 versions of their packages will be coming soon with 8.0.37.

Issue #6275, opened 2024-06-05
Custom Command not Recognized


introfini:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Please make sure that you have already looked at troubleshooting guide ========
Troubleshooting guide: https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/ 
Simple things to check:
* ddev poweroff
* Restart Docker Provider
* Reboot computer
* Temporarily disable VPN and firewall
Press any key to continue:


======== Output file will be in /var/folders/x0/l2nysd4577ngb586867ml4jr0000gn/T/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/local/bin/ddev

These config files were loaded for project bstorelondon: [/Users/josefernandes/Dev/bstorelondon/.ddev/config.yaml]
name: bstorelondon
type: drupal10
docroot: web
php_version: 8.1
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.23.1
router_http_port: 80
router_https_port: 443
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.4}
hooks: map[post-import-db:[map[exec:drush en -y stage_file_proxy devel twig_vardumper]]]
upload_dirs: [sites/default/files ../database]
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 16
default_container_timeout: 120


======== existing project customizations ========
.ddev/docker-compose.expose-mailpit.yaml
.ddev/php/php-example.ini.example


======== installed DDEV add-ons ========
No registered add-ons were found. Add-ons installed before DDEV v1.22.0 will not be listed.
Update them with `ddev get` so they'll be shown.


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /Users/josefernandes/Dev/bstorelondon
#ddev-generated
unmodified #ddev-generated found in .ddev/mutagen/mutagen.yml


[...]

Issue #5523, opened 2023-11-09
Drush Error: unknown command "drush" for "ddev" on *second* `ddev drush st`


stloc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
➜  bo_drupal git:(main) ✗ ddev debug test 
Running bash [-c /tmp/test_ddev.sh] 
======= Existing project config =========
These config files were loaded for project bmp-bo: [/home/l.guigo@telegramme.fr/dev/bretagne-marchespublics/bo_drupal/.ddev/config.yaml /home/l.guigo@telegramme.fr/dev/bretagne-marchespublics/bo_drupal/.ddev/config.platformsh.yaml]
name: bmp-bo
type: drupal10
docroot: web
php_version: 8.1
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.22.2
database: {mariadb 10.4}
hooks: map[post-start:[map[exec:platform self:update -qy --no-major || true] map[exec:mkdir -p ${PLATFORM_CACHE_DIR} || true] map[exec:[ ! -z "${PLATFORMSH_CLI_TOKEN:-}" ] && (platform ssh-cert:load  -y || true)] map[composer:install] map[exec:set -e
] map[exec:set -e
php ./drush/platformsh_generate_drush_yml.php
# if drupal is installed, will call the following drush commands:
#   - `cache-rebuild`
#   - `updatedb`
#   - and if config files are present, `config-import`
cd web
bash $PLATFORM_APP_DIR/drush/platformsh_deploy_drupal.sh
] map[exec:phpenmod blackfire] map[exec:platform self:update -qy --no-major || true] map[exec:mkdir -p ${PLATFORM_CACHE_DIR} || true] map[exec:[ ! -z "${PLATFORMSH_CLI_TOKEN:-}" ] && (platform ssh-cert:load  -y || true)] map[composer:install] map[exec:set -e
] map[exec:set -e
php ./drush/platformsh_generate_drush_yml.php
# if drupal is installed, will call the following drush commands:
#   - `cache-rebuild`
#   - `updatedb`
#   - and if config files are present, `config-import`
cd web
bash $PLATFORM_APP_DIR/drush/platformsh_deploy_drupal.sh
] map[exec:phpenmod blackfire]]]
[...]

PR #6293, opened 2024-06-10
feat: add warnings for global commands when they are "unknown", for #5523, for #6275

stasadev:
## The Issue

- #5523
- #6275

DDEV users may run some legitimate bash commands built into DDEV and see errors like this:

```
$ ddev drush
Error: unknown command "drush" for "ddev"

Did you mean this?
        push

Run 'ddev --help' for usage.
```

For our users, this is confusing, they don't know what to do and how to fix it. I myself have been hit by an "unknown command" several times.

We need to provide better output to indicate why the specified command does not work.

Adds a warning before "unknown command" error for custom bash commands, explaining why they don't work.

```
$ cd ~ && ddev mailpit
Command 'mailpit' cannot be used outside the project directory.

$ cd ~/workspace && mkdir command-warnings && cd command-warnings && ddev config --auto --database="postgres:16"

$ ddev drush
Command 'drush' is not available for the 'php' project type.
If you intend to use 'drush', change the project type to one of the supported types [
        ddev config --project-type=drupal
        ddev config --project-type=drupal10
        ddev config --project-type=drupal9
        ddev config --project-type=drupal8
        ddev config --project-type=drupal7
]

$ ddev heidisql # using macOS or Linux
Command 'heidisql' cannot be used with your OS.

$ ddev sequelpro # using macOS without Sequel Pro
Command 'sequelpro' cannot be used, because the binary is not found at [
        /Applications/Sequel Pro.app
]

$ ddev sequelpro # using macOS with Sequel Pro
Command 'sequelpro' is not available for the 'postgres' database type.
```

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -132,6 +132,9 @@ func addCustomCommandsFromDir(rootCmd *cobra.Command, app *ddevapp.DdevApp, serv
 		// Skip host commands that need a project if we aren't in a project directory.
 		if service == "host" && app == nil {
 			if val, ok := directives["CanRunGlobally"]; !ok || val != "true" {
+				if isCustomCommandInArgs(commandName) {
+					util.Warning("Command '%s' cannot be used outside the project directory.", commandName)
+				}
 				continue
 			}
 		}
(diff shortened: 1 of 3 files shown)


rfay:
Great initiative!

rfay (approved):
This is a great improvement, and great way to calm a number of support activities. It's not completely comprehensive yet, but I think that's fine. 

For example, `ddev drush` outside a project directory doesn't give the suggestion about doing it in a project directory.

stasadev:
> For example, `ddev drush` outside a project directory doesn't give the suggestion about doing it in a project directory.

Yes, that's one thing I tried figure out but gave up because if I add these project commands to the global scope:
```
$ ddev drush
Failed to RunSimpleContainer to inspect database version/type: failed to create/start Docker container GetExistingDBType--KnjrHa (&{   false true true map[] false false false [] [sh -c ( test -f /var/tmp/mysql/db_mariadb_version.txt && cat /var/tmp/mysql/db_mariadb_version.txt ) || ( test -f /var/tmp/postgres/PG_VERSION && cat /var/tmp/postgres/PG_VERSION) || true] <nil> false busybox:stable map[]  [] false  [] map[com.ddev.site-name:]  <nil> []}, &{[-mariadb:/var/tmp/mysql -postgres:/var/tmp/postgres]  { map[]}  map[] { 0} false  [] [0 0] map[] [] []  [] [] [] [host.docker.internal:host-gateway] []   [] 0  false false false [] map[] map[]   0 map[]   {0 0 0  0 [] [] [] [] [] 0 0 0 0   [] [] [] 0 0 0 0 <nil> <nil> <nil> [] 0 0 0 0} [] [] [] <nil>}): Error response from daemon: create -mariadb: "-mariadb" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path, output=
```
Basically it starts touching parts of DDEV not meant for it.

PR #6296, opened 2024-06-10
docs: update first step in project-types.md

kdambekalns:
This fixes the pointer to where a new project type should be added to.


rfay (approved):
Thanks!

PR #6297, opened 2024-06-11
fix: Add matching mysql clients to mysql:5.7 and mysql:8.0, for #6083

rfay:
## The Issue

- #6083

We're trying to match mysql client versions to server versions in at least some cases.

This does the job for mysql:8.0 and mysql:5.7. It prepares for mysql:8.4

It also adds TestWebserverDBClient to test all this and make sure the mysqldump etc work properly on the host side.

It depends on the new mysql client builder project, https://github.com/ddev/mysql-client-build, so review of that is requested as well.

* Start a mysql-based project. 
* `ddev exec mysql --version`
* `ddev exec mysqldump --version`

## Automated Testing Overview

* New TestWebserverDBClient exercises the clients on the webserver to make sure they work.

## Release/Deployment Notes

It's more important than ever that people match their DDEV database version with the upstream server version, so that the mysqldump format will match, etc.


stasadev (approved):
I tested this and it works great! I think that after this version is released, people will have a better experience with MySQL databases.

Not something super important, but shouldn't we also bump mysql 5.7 Docker image to 5.7.44 to be fully compliant?
https://github.com/ddev/mysql-arm64-images/blob/main/5.7/base_version.txt

```
$ ddev exec mysqldump --version
mysql  Ver 14.14 Distrib 5.7.44, for Linux (x86_64) using  EditLine wrapper

$ ddev exec -s db mysql --version
mysql  Ver 14.14 Distrib 5.7.42, for Linux (x86_64) using  EditLine wrapper
```

rfay:
> shouldn't we also bump mysql 5.7 Docker image to 5.7.44 to be fully compliant

1. I don't expect to have matching clients on everything at the patch level, it will be just too painful and difficult, although our ability to build them is now enhanced. (Note that the 5.7.42 client build failed due to something I didn't chase in the cmake configuration, but since it worked with 5.7.44 I went with that)
2. Because we get our 5.7 from Ubuntu 18.04 (our source for multiple architectures at this point) we get the latest in there, which is 5.7.42 (and it may never be updated?):

```
mysql-server/bionic-updates,bionic-security 5.7.42-0ubuntu0.18.04.1 all
  MySQL database server (metapackage depending on the latest version)

mysql-server-5.7/bionic-updates,bionic-security 5.7.42-0ubuntu0.18.04.1 arm64
  MySQL database server binaries and system database setup
  ```

stasadev:
Okay, thank you for the confirmation.

rfay:
I updated the .goreleaser.yml to provide the URLs for pushing the README to dockerhub, and updated https://github.com/ddev/mysql-client-build/blob/main/README.md to the required format.

rfay:
Rebased, pushed new ddev-webserver

stasadev (approved):
I think it's ready to be pulled.

stasadev:
I didn't notice, but `/tmp/mysql.tgz` is not deleted after installation. @rfay, could you add the cleanup in another PR with the webserver change?

https://github.com/ddev/ddev/blob/32bc2f065ee34fd488a822e91529b2b3a0f5bdd3/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mysql-client-install.sh#L21

rfay:
Oh, I see. You're saying that it gets left there during the build if the database type is mysql. It's only 4.8MB but I agree it should be removed. It could be removed in the build process as well.

rfay:
I snuck this into 
* https://github.com/ddev/ddev/pull/6304

PR #6299, opened 2024-06-11
build: bump yq to v4.44.1

juanjol:
## The Issue

yq yaml parser has a new version with multiple improvements 

bump version to the latests stable release (v4.44.1)

install ddev with this PR and launch yq --version inside web container. Version should be v4.44.1.

## Automated Testing Overview

None 

## Release/Deployment Notes

None

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -63,7 +63,7 @@ ARG PHP_DEFAULT_VERSION="8.2"
 ENV DDEV_PHP_VERSION=$PHP_DEFAULT_VERSION
 ENV PHP_VERSIONS="php5.6 php7.0 php7.1 php7.2 php7.3 php7.4 php8.0 php8.1 php8.2 php8.3"
 ENV PHP_INI=/etc/php/$PHP_DEFAULT_VERSION/fpm/php.ini
-ENV YQ_VERSION=v4.30.5
+ENV YQ_VERSION=v4.44.1
 ENV DRUSH_VERSION=8.4.12
 ENV NODE_VERSION=20
 # composer normally screams about running as root, we don't need that.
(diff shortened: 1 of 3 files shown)


rfay:
Just changing it here doesn't make it actually change. A new image has to be pushed and be referred to in the versionconstants.go.

Can you say why this is important to you?

I note that `yq` is available via apt, I don't think it used to be.

```
yq/stable 3.1.0-3 all
  Command-line YAML processor - jq wrapper for YAML documents
```

stasadev:
There are two `yq`:
https://github.com/kislyuk/yq
https://github.com/mikefarah/yq - we use this one, it's better, because it has more features. I know this because I also use it outside of DDEV and I used to use https://github.com/kislyuk/yq at first but then switched to this one.

I think we could even pull the latest version, similar to Mailpit:

- #5524

rfay:
Getting latest works for me.

stasadev:
@juanjol, please update this PR with my [suggestion](https://github.com/ddev/ddev/pull/6299#issuecomment-2161342500), I think you will be happy to have the latest `yq` with every new DDEV release.

After that, we will push the new Docker images and update `versionconstants.go`.

juanjol:
Hi @stasadev, @rfay, thank you for your quick feedback! The main reason to update yq is to have a version that is capable of output yaml contents in a shell compatible mode. I would like to add a custom yaml config inside .ddev folder to keep some configs relate to our custom environments but not to ddev. I like the idea of always have the latests versions, so I updated this PR. Is there something to do with versionconstants.go??

rfay:
One request for the future @juanjol  - Please don't use your fork's `master` branch for a PR. It means that you are stuck with it, and if you push anything else into it, it makes a mess of the PR.  Instead, please use a meaningful branch name that is devoted to the PR (something like `20240611_juanjol_short_description`, as is described in https://ddev.readthedocs.io/en/stable/developers/building-contributing/#pull-requests-and-pr-preparation

rfay:
I pushed the images and updated the versionconstants.go

juanjol:
Thank you very much, understood! Sorry for the inconvenience, the truth is that I did it directly editing in github without downloading the repo and it was not the best idea. 

Thanks again for the quick response and for the advice.

PR #6303, opened 2024-06-12
fix: pass all valid options to nested commands in `ddev composer create`, fixes #6300, fixes #6246, for #5058

### LINKED ISSUE #6246: Composer create does not pass "--no-interaction" flag for "post-create-project-cmd" (Craft CMS)
### Preliminary checklist

https://gist.github.com/mandrasch/9c5093ae8407a877f0b3859ab7b3842d
</details>

### Expected Behavior

Craft CMS has the following (https://github.com/craftcms/craft/blob/5.x/composer.json) `post-create-project-cmd`:

```
 "scripts": {
    "post-create-project-cmd": [
      "@php -r \"file_exists('.env') || copy('.env.example.dev', '.env');\"",
      "@php -r \"unlink('composer.json');\"",
      "@php -r \"rename('composer.json.default', 'composer.json');\"",
      "@composer dump-autoload -o",
      "@php craft setup/welcome"
    ]
  }
  ```
  
 I want to run an install without interaction, therefore I want to use the following
  
 `ddev composer create -y --no-scripts --no-interaction craftcms/craft`
 
 with DDEV v.1.21.3 (since post-create-cmd is executed now)

### Actual Behavior

Executing  `ddev composer create -y --no-scripts --no-interaction craftcms/craft` get's stuck at this screen with a prompt:

image

Console output: 

```
Executing composer command: [composer run-script post-create-project-cmd]
 
> @php -r "file_exists('.env') || copy('.env.example.dev', '.env');"
> @php -r "unlink('composer.json');"
> @php -r "rename('composer.json.default', 'composer.json');"
Generating optimized autoload files
Generated optimized autoload files containing 5444 classes
> @php craft setup/welcome

   ______ .______          ___       _______ .___________.
  /      ||   _  \        /   \     |   ____||           |
 |  ,----'|  |_)  |      /  ^  \    |  |__   `---|  |----`
 |  |     |      /      /  /_\  \   |   __|      |  |
 |  `----.|  |\  \----./  _____  \  |  |         |  |
  \______|| _| `._____/__/     \__\ |__|         |__|
 
     A       N   E   W       I   N   S   T   A   L   L
               ______ .___  ___.      _______.
              /      ||   \/   |     /       |
             |  ,----'|  \  /  |    |   (----`
             |  |     |  |\/|  |     \   \
             |  `----.|  |  |  | .----)   |
              \______||__|  |__| |_______/


Generating an application ID ... done (CraftCMS--e8279XXXX)
Generating a security key ... done (UgSm61wTXXXX)

Welcome to Craft CMS!

Are you ready to begin the setup? (yes|no) [no]:
```

### Steps To Reproduce

1. Execute:

```
mkdir craft5-ddev && \ 
  cd craft5-ddev && \
  ddev config \
    --project-type=craftcms \
    --docroot=web \
    --create-docroot \
    --php-version="8.2" \
    --database="mysql:8.0" \
    --nodejs-version="20" && \
  ddev start -y && \
  ddev composer create -y --no-scripts --no-interaction craftcms/craft && \
  ddev craft install/craft \
    --username=admin \
    --password=password123 \
    --email=admin@example.com \
    --site-name=Testsite \
    --language=en \
    --site-url='$DDEV_PRIMARY_URL' && \
  echo 'Nice, ready to launch!' && \
  ddev launch
  ```
  
2. Check the normal composer installation, works with `--no-interaction`:
  
```
ddev ssh
composer create-project craftcms/craft --no-interaction my-craft-project
```

### Anything else?

- CraftCMS related source code:  `@php craft setup/welcome` - ActionWelcome checks for if (!$this->interactive https://github.com/craftcms/cms/blob/5.x/src/console/controllers/SetupController.php#L168 --> and this is set here https://github.com/craftcms/cms/blob/c2f58dce722f79baa208b879f821f74960627a12/src/console/ControllerTrait.php#L171.

```
    protected function checkTty(): void
    {
        // Don't treat this as interactive if it doesn't appear to be a TTY shell
        if ($this->interactive && !Process::isTtySupported()) {
            $this->interactive = false;
        }
    }
 ```
 
So maybe this needs to be changed here in DDEV when `--no-interaction` is passed? 🤔 (just a guess)
https://github.com/ddev/ddev/blob/70fc4cd7b8b84600c18f0435f2ca82f1dba8b730/cmd/ddev/cmd/composer-create.go#L284C36-L284C41
 
 
 ```
 		stdout, stderr, _ = app.Exec(&ddevapp.ExecOpts{
			Service: "web",
			Dir:     app.GetComposerRoot(true, false),
			RawCmd:  composerCmd,
			Tty:     isatty.IsTerminal(os.Stdin.Fd()),
		})
```
 
 Or this https://github.com/ddev/ddev/blob/70fc4cd7b8b84600c18f0435f2ca82f1dba8b730/cmd/ddev/cmd/composer-create.go#L237C4-L244C1
 
 ```
 for _, osarg := range osargs {
				for _, supportedArg := range supportedArgs {
					if strings.HasPrefix(osarg, supportedArg) {
						composerCmd = append(composerCmd, osarg)
					}
				}
			}
```

needs to be appended to the second command to 
``` 
// Try to run post-create-project-cmd.
composerCmd = []string{
```

as well (second guess).

- DDEV improved ddev composer create in v1.23.1 - the post-create-project-cmd is now being executed properly  (https://github.com/ddev/ddev/pull/5058). Wasn't executed before as far as I understand. 
- My guide for previous versions of DDEV: https://dev.to/mandrasch/install-craft-cms-v5-alpha-with-one-command-via-ddev-4ne5)

### LINKED ISSUE #6300: Run CakePHP Quickstart with no interaction (skip setting directory permissions)
### Preliminary checklist

```
```
</details>

### Expected Behavior

Copy and pasting the CakePHP quickstart should:
- install website
- automatically launch the browser to the functionaling website

### Actual Behavior

Composer stops and asks to "Set Folder Permissions":

```shell
$ ddev composer create --prefer-dist cakephp/app:~5.0
...
PHP CodeSniffer Config installed_paths set to ../../cakephp/cakephp-codesniffer,../../slevomat/coding-standard
Executing composer command: [composer run-script post-create-project-cmd]
 
> App\Console\Installer::postInstall
Created `config/app_local.php` file
Created `/var/www/html/logs` directory
Created `/var/www/html/tmp/cache/views` directory
Set Folder Permissions ? (Default to Y) [Y,n]? 
```

### Steps To Reproduce

Follow quickstart guide for CakePHP (replicated below)

```shell
mkdir my-cakephp-site && cd my-cakephp-site
ddev config --project-type=cakephp --docroot=webroot
ddev composer create --prefer-dist cakephp/app:~5.0
ddev cake
ddev launch
```

### Anything else?

I tried use `--no-interaction` on the composer line but it still stopped.
Eg. `ddev composer create --prefer-dist cakephp/app:~5.0 --no-interaction`

I believe this is because of the `composer run-script post-create-project-cmd` line which doesn't inherit the flag.

Adding a `.ddev/.env` file with `COMPOSER_NO_INTERACTION=1` does NOT cause composer to stop. But we probably don't want this to always be active, only during the initial "ddev composer create" step.

stasadev:
## The Issue

- #6300
- #6246
- #5058

Composer options were not respected by `composer run-script`.

Makes `ddev composer create` fully compatible with all possible composer options. I removed all the hard-coded options.
Adds composer options to `composer run-script`.
Adds `--no-interaction` to CakePHP quickstart.

---

`--preserve-flags` (which doesn't add `--no-plugins` and `--no-scripts`) didn't work as expected, I decided to remove it.

This is not a breaking change, because `--preserve-flags` didn't work before (it was always `false`).

---

I moved `composer run-script post-create-project-cmd` outside the condition for `composer install`, to make it work the same way as in `composer create-project`.

---

If you passed `--no-scripts`, DDEV tried to run `composer run-script`, which was wrong behavior, fixed it.

---

If `--help` or `--version` flags were provided, composer showed it's help page or it's version, and DDEV tried to continue creating the project, which was wrong behavior, fixed it.

---

I refactored the test logic to make it more clear and added more tests. I removed the testing for the `psr/log` package and replaced it with custom-made fake packages.

1. The order of commands executed for `ddev composer create` must be the same as for `composer create-project`.

To check this behavior, you can use a test repository (it is used for `testdata` here) which contains all the necessary situations to check:

```
cd ~/workspace && mkdir composer-create && cd composer-create
ddev config --auto
mkdir -p .ddev/test-ddev-composer-create/{test-ddev-require,test-ddev-require-dev}

curl -o .ddev/test-ddev-composer-create/composer.json https://raw.githubusercontent.com/stasadev/ddev/20240612_stasadev_composer_create/cmd/ddev/cmd/testdata/TestComposerCreateCmd/.ddev/test-ddev-composer-create/composer.json
curl -o .ddev/test-ddev-composer-create/test-ddev-require/composer.json https://raw.githubusercontent.com/stasadev/ddev/20240612_stasadev_composer_create/cmd/ddev/cmd/testdata/TestComposerCreateCmd/.ddev/test-ddev-composer-create/test-ddev-require/composer.json
curl -o .ddev/test-ddev-composer-create/test-ddev-require-dev/composer.json https://raw.githubusercontent.com/stasadev/ddev/20240612_stasadev_composer_create/cmd/ddev/cmd/testdata/TestComposerCreateCmd/.ddev/test-ddev-composer-create/test-ddev-require-dev/composer.json

export repo='{"type": "path", "url": ".ddev/test-ddev-composer-create", "options": {"symlink": false}}'

# see the result for:
composer create-project --repository $repo test/ddev-composer-create # add different flags here

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf

# see the result for:
ddev composer create --repository $repo test/ddev-composer-create # add different flags here

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf
````

`test/ddev-composer-create` package (`.ddev/test-ddev-composer-create/composer.json`) contains this:
- it requires `test/ddev-require` from `require` section:
  - `vendor/test/ddev-require` will be here, but only if there is no `--no-install` flag
- it requires `test/ddev-require-dev` from `require-dev` section:
  - `vendor/test/ddev-require-dev` will be here, but only if there is no `--no-dev`/`--no-install` flag
- it runs `post-root-package-install` script:
  - this script should run before `composer install`
  - it creates `created-by-post-root-package-install` file, but only if there is no `--no-scripts` flag
- it runs `post-create-project-cmd` script:
  - this script should run after `composer install`
  - it creates `created-by-post-create-project-cmd` file, but only if there is no `--no-scripts` flag

And use different flag combinations to see how it works for `composer create-project` and `ddev composer create`. It is important to check if the commands run in the same order.

Available flags for `composer create-project` https://getcomposer.org/doc/03-cli.md#create-project

Ignore the `test-ddev-require` and `test-ddev-require-dev` directories created in the project root (they are expected to be created every time as they are part of the composer package).

Every `ddev composer create` will have this output:
```
Executing composer command: [composer ...]
```
You can see what flags are used for each of these composer commands, and flags like `--no-interaction` will be also used for `composer run-script` and `composer install`.

For example:

This should not create `vendor` folder, and there should be no `created-by-*` files
```
# see the result for:
composer create-project --repository $repo test/ddev-composer-create --no-install --no-scripts

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf

# see the result for:
ddev composer create --repository $repo test/ddev-composer-create --no-install --no-scripts

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf
```

This should create `vendor/test/ddev-require` folder, (but not `vendor/test/ddev-require-dev`) and there should be no `created-by-*` files
```
# see the result for:
composer create-project --repository $repo test/ddev-composer-create --no-scripts --no-dev

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf

# see the result for:
ddev composer create --repository $repo test/ddev-composer-create --no-scripts --no-dev

# clear the directory
ls | grep -xv ".ddev" | xargs rm -rf
```

And so on.

---

2. Try CakePHP quickstart (without interaction), it should not ask for input:

```
mkdir my-cakephp-site && cd my-cakephp-site
ddev config --project-type=cakephp --docroot=webroot
ddev composer create --prefer-dist --no-interaction cakephp/app:~5.0
```

---

3. Try `Steps To Reproduce` example (it should not ask for input) from:

- #6246

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -12,7 +12,6 @@ import (
 	"github.com/ddev/ddev/pkg/composer"
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/fileutil"
-	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/output"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/mattn/go-isatty"
@@ -33,25 +32,28 @@ ddev composer create "typo3/cms-base-distribution:^10"
 ddev composer create drupal/recommended-project --no-install
 ddev composer create --repository=https://repo.magento.com/ magento/project-community-edition
 ddev composer create --prefer-dist --no-interaction --no-dev psr/log
-ddev composer create --preserve-flags --no-interaction psr/log
 `,
 	ValidArgsFunction: getComposerCompletionFunc(true),
 	Run: func(cmd *cobra.Command, _ []string) {
-		preserveFlags, _ := cmd.Flags().GetBool("preserve-flags")
-
 		// We only want to pass all flags and args to Composer
 		// cobra does not seem to allow direct access to everything predictably
-		osargs := []string{}
+		var osargs []string
 		if len(os.Args) > 3 {
 			osargs = os.Args[3:]
-			// I don't know why, but cmd.Flags().GetBool("preserve-flags") is always false, so we can check it here
-			preserveFlags = slices.Contains(osargs, "--preserve-flags")
-			osargs = nodeps.RemoveItemFromSlice(osargs, "--preserve-flags")
-			expandedOsargs := []string{}
+			var expandedOsargs []string
 			for _, osarg := range osargs {
 				// If this is a combined short option like "-test", split it into "-t -e -s -t"
 				// This is necessary, because each of these options will be checked for validity
 				if strings.HasPrefix(osarg, "-") && !strings.HasPrefix(osarg, "--") && len(osarg) > 2 {
+					// -vvv and -vv are exceptions, that should not be split
+					if strings.HasPrefix(osarg, "-vvv") {
+						expandedOsargs = append(expandedOsargs, "-vvv")
+						continue
+					}
+					if strings.HasPrefix(osarg, "-vv") {
+						expandedOsargs = append(expandedOsargs, "-vv")
+						continue
+					}
 					for _, char := range osarg[1:] {
 						expandedOsargs = append(expandedOsargs, "-"+string(char))
 					}
@@ -60,6 +62,9 @@ ddev composer create --preserve-flags --no-interaction psr/log
 				}
 			}
 			osargs = expandedOsargs
+		} else {
+			_ = cmd.Help()
+			return
 		}
 
 		app, err := ddevapp.GetActiveApp("")
@@ -134,9 +139,10 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			return !strings.Contains(out, fmt.Sprintf(`"%s" option does not exist`, option))
 		}
 
-		// Add some args to avoid troubles while cloning as long as
-		// --preserve-flags is not set.
-		createArgs := []string{}
+		// Add some args to avoid troubles while cloning the project.
+		// We add the three options to "composer create-project": --no-plugins, --no-scripts, --no-install
+		// These options make the difference between "composer create-project" and "ddev composer create".
+		var createArgs []string
 
 		for _, osarg := range osargs {
 			if isValidComposerOption("create-project", osarg) {
@@ -144,13 +150,18 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			}
 		}
 
-		if !preserveFlags && !slices.Contains(createArgs, "--no-plugins") {
+		// this slice will be used for nested composer commands
+		validCreateArgs := createArgs
+
+		if !slices.Contains(createArgs, "--no-plugins") {
+			// Don't run plugin events for "composer create-project", but run them for "composer run-script" and "composer install"
 			createArgs = append(createArgs, "--no-plugins")
 		}
 
 		// Remember if --no-scripts was provided by the user
 		noScriptsPresent := slices.Contains(createArgs, "--no-scripts")
 		if !noScriptsPresent {
+			// Don't run scripts for "composer create-project", but run them for "composer install"
 			createArgs = append(createArgs, "--no-scripts")
 		}
 
@@ -218,6 +229,7 @@ ddev composer create --preserve-flags --no-interaction psr/log
 		}
 
 		composerManifest, _ := composer.NewManifest(path.Join(composerRoot, "composer.json"))
+		var validRunScriptArgs []string
 
 		if !noScriptsPresent && composerManifest != nil && composerManifest.HasPostRootPackageInstallScript() {
 			// Try to run post-root-package-install.
@@ -227,11 +239,13 @@ ddev composer create --preserve-flags --no-interaction psr/log
 				"post-root-package-install",
 			}
 
-			for _, osarg := range osargs {
-				if isValidComposerOption("run-script", osarg) {
-					composerCmd = append(composerCmd, osarg)
+			for _, validCreateArg := range validCreateArgs {
+				if isValidComposerOption("run-script", validCreateArg) {
+					validRunScriptArgs = append(validRunScriptArgs, validCreateArg)
 				}
 			}
+
+			composerCmd = append(composerCmd, validRunScriptArgs...)
 
 			output.UserOut.Printf("Executing Composer command: %v\n", composerCmd)
 
@@ -266,9 +280,9 @@ ddev composer create --preserve-flags --no-interaction psr/log
 				"install",
 			}
 
-			for _, osarg := range osargs {
-				if isValidComposerOption("install", osarg) {
-					composerCmd = append(composerCmd, osarg)
+			for _, validCreateArg := range validCreateArgs {
+				if isValidComposerOption("install", validCreateArg) {
+					composerCmd = append(composerCmd, validCreateArg)
 				}
 			}
 
@@ -306,11 +320,16 @@ ddev composer create --preserve-flags --no-interaction psr/log
 				"post-create-project-cmd",
 			}
 
-			for _, osarg := range osargs {
-				if isValidComposerOption("run-script", osarg) {
-					composerCmd = append(composerCmd, osarg)
+			// If the flags for "run-script" were already validated, don't validate them again.
+			if validRunScriptArgs == nil {
+				for _, validCreateArg := range validCreateArgs {
+					if isValidComposerOption("run-script", validCreateArg) {
+						validRunScriptArgs = append(validRunScriptArgs, validCreateArg)
+					}
 				}
 			}
+
+			composerCmd = append(composerCmd, validRunScriptArgs...)
 
 			output.UserOut.Printf("Executing Composer command: %v\n", composerCmd)
 
@@ -360,7 +379,6 @@ for basic project creation or 'ddev ssh' into the web container and execute
 }
 
 func init() {
-	ComposerCreateCmd.Flags().Bool("preserve-flags", false, "Do not append `--no-plugins` flag")
 	ComposerCmd.AddCommand(ComposerCreateProjectCmd)
 	ComposerCmd.AddCommand(ComposerCreateCmd)
 }
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/ddev/ddev/pkg/composer"
@@ -43,6 +44,8 @@ ddev composer create --preserve-flags --no-interaction psr/log
 		osargs := []string{}
 		if len(os.Args) > 3 {
 			osargs = os.Args[3:]
+			// I don't know why, but cmd.Flags().GetBool("preserve-flags") is always false, so we can check it here
+			preserveFlags = slices.Contains(osargs, "--preserve-flags")
 			osargs = nodeps.RemoveItemFromSlice(osargs, "--preserve-flags")
 			expandedOsargs := []string{}
 			for _, osarg := range osargs {
@@ -91,10 +94,10 @@ ddev composer create --preserve-flags --no-interaction psr/log
 
 				checkPath := app.GetRelativeDirectory(walkPath)
 
-				if walkInfo.IsDir() && nodeps.ArrayContainsString(skipDirs, checkPath) {
+				if walkInfo.IsDir() && slices.Contains(skipDirs, checkPath) {
 					return filepath.SkipDir
 				}
-				if !nodeps.ArrayContainsString(composerCreateAllowedPaths, checkPath) {
+				if !slices.Contains(composerCreateAllowedPaths, checkPath) {
 					return fmt.Errorf("'%s' is not allowed to be present. composer create needs to be run on a clean/empty project with only the following paths: %v - please clean up the project before using 'ddev composer create'", filepath.Join(appRoot, checkPath), composerCreateAllowedPaths)
 				}
 				if err != nil {
@@ -141,16 +144,16 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			}
 		}
 
-		if !preserveFlags && !nodeps.ArrayContainsString(createArgs, "--no-plugins") {
+		if !preserveFlags && !slices.Contains(createArgs, "--no-plugins") {
 			createArgs = append(createArgs, "--no-plugins")
 		}
 
-		if !preserveFlags && !nodeps.ArrayContainsString(createArgs, "--no-scripts") {
+		if !preserveFlags && !slices.Contains(createArgs, "--no-scripts") {
 			createArgs = append(createArgs, "--no-scripts")
 		}
 
 		// Remember if --no-install was provided by the user
-		noInstallPresent := nodeps.ArrayContainsString(createArgs, "--no-install")
+		noInstallPresent := slices.Contains(createArgs, "--no-install")
 		if !noInstallPresent {
 			// Add the --no-install option by default to avoid issues with
 			// rsyncing many files afterwards to the project root.
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -148,7 +148,9 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			createArgs = append(createArgs, "--no-plugins")
 		}
 
-		if !preserveFlags && !slices.Contains(createArgs, "--no-scripts") {
+		// Remember if --no-scripts was provided by the user
+		noScriptsPresent := slices.Contains(createArgs, "--no-scripts")
+		if !noScriptsPresent {
 			createArgs = append(createArgs, "--no-scripts")
 		}
 
@@ -188,6 +190,11 @@ ddev composer create --preserve-flags --no-interaction psr/log
 			output.UserErr.Println(stderr)
 		}
 
+		// If options contain help or version flags, do not continue
+		if slices.Contains(createArgs, "-h") || slices.Contains(createArgs, "--help") || slices.Contains(createArgs, "-V") || slices.Contains(createArgs, "--version") {
+			return
+		}
+
 		output.UserOut.Printf("Moving install to Composer root")
 
 		rsyncArgs := "-rltgopD" // Same as -a
@@ -212,7 +219,7 @@ ddev composer create --preserve-flags --no-interaction psr/log
 
 		composerManifest, _ := composer.NewManifest(path.Join(composerRoot, "composer.json"))
 
-		if !preserveFlags && composerManifest != nil && composerManifest.HasPostRootPackageInstallScript() {
+		if !noScriptsPresent && composerManifest != nil && composerManifest.HasPostRootPackageInstallScript() {
 			// Try to run post-root-package-install.
 			composerCmd = []string{
 				"composer",
@@ -226,7 +233,7 @@ ddev composer create --preserve-flags --no-interaction psr/log
 				}
 			}
 
-			output.UserOut.Printf("Executing composer command: %v\n", composerCmd)
+			output.UserOut.Printf("Executing Composer command: %v\n", composerCmd)
 
 			stdout, stderr, _ = app.Exec(&ddevapp.ExecOpts{
 				Service: "web",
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by rfay — CHANGES_REQUESTED** [review]
Could you please add test coverage for this? Thanks! We have been fiddling with `ddev composer create` for years and years now, and we keep incrementally improving it, but we never seem to completely get it finished.

**Comment by stasadev** [comment]
Yes, I will add some new test or modify an existing one.

File: cmd/ddev/cmd/composer-create.go [hunk 1]
```diff
@@ -43,9 +44,22 @@ ddev composer create --preserve-flags --no-interaction psr/log
 		osargs := []string{}
 		if len(os.Args) > 3 {
 			osargs = os.Args[3:]
-			osargs = nodeps.RemoveItemFromSlice(osargs, "--yes")
-			osargs = nodeps.RemoveItemFromSlice(osargs, "-y")
+			// I don't know why, but cmd.Flags().GetBool("preserve-flags") is always false, so we can check it here
+			preserveFlags = slices.Contains(osargs, "--preserve-flags")
```
stasadev, line 48: This baffles me, I don't understand why Cobra reads `--preserve-flags` as `false` no matter how I try to pass this flag.

Anyway, at least `--preserve-flags` now works as described (it didn't before).

stasadev:
I added more conditions and checks to the test, now it covers most possible situations.

tyler36:
Confirmed. This PR fixes #6300 .  Thank you for the quick turn around.

## Test

1. Confirm DDEV version

```shell
$ ddev -v
ddev version v1.23.1-62-g4f748c85a
```

1. Follow (updated) cakePHP quickstart. 

```shell
mkdir my-cakephp-site && cd my-cakephp-site
ddev config --project-type=cakephp --docroot=webroot
ddev composer create --prefer-dist --no-interaction cakephp/app:~5.0
ddev cake
ddev launch
```

> [!NOTE]  
This PR updates the quickstart to include `--no-interaction` . This makes no difference in DDEV `1.23.1` (see #6300 )

tyler36 (approved):

stasadev:
There is still some difference how composer scripts work in:

- `ddev composer create --no-install` runs only `post-root-package-install` and nothing else
- `composer create-project --no-install` runs `post-root-package-install` and `post-create-project-cmd`

I will investigate this further today and adjust the behavior.

stasadev:
`--preserve-flags` (which doesn't add `--no-plugins` and `--no-scripts`) didn't work as expected:

The order was:

- `composer create-project`
- `composer run-script post-root-package-install`
- `composer run-script post-create-project-cmd`
- `composer install`

The right order:

- `composer create-project`
- `composer run-script post-root-package-install`
- `composer install`
- `composer run-script post-create-project-cmd`

This is why I made `--preserve-flags` not add only `--no-plugins`

This is not a breaking change, because `--preserve-flags` didn't work before (it was always `false`).

---

I moved `composer run-script post-create-project-cmd` outside the condition for `composer install`, to make it work the same way as in `composer create-project`.

---

If you passed `--no-scripts`, DDEV tried to run `composer run-script`, which was wrong behavior, fixed it.

---

If `--help` or `--version` flags were provided, composer showed it's help page or it's version, and DDEV tried to continue creating the project, which was wrong behavior, fixed it.

---

I refactored the test logic to make it more clear and added more tests. I removed the testing for the `psr/log` package and replaced it with custom-made fake packages.

stasadev:
When I split combined short flags, it broke `-vvv` functionality.
It turned out that `-v -v -v` is not the same as `-vvv`, so I will see how to fix it.

stasadev:
Rebased.

I fixed `-vvv` and changed a bit the logic for flags.

I updated our documentation with a more detailed explanation of what steps we do with `ddev composer create`.

I researched the Composer documentation, learned more about what `--no-plugin`, `--no-scripts` and `--no-install` flags do. In result I decided to remove `--preserve-flags` completely, because if we do not add `--no-plugins`, `--no-scripts` and `--no-install` to `composer create-project`, it breaks the original logic, so I don't think we ever needed it.

`--preserve-flags` didn't work before this PR, so this removal is not a breaking change.

---

I'm pretty sure the most of the original logic here works right.

There are some conflicts with the `composer create-project` flags, such as `--ask` and `--working-dir`, using which results in error for `ddev composer create`, but I decided not to hard-code them, as the composer may add/remove some flags in the future, and I don't want to add unnecessary checks.

docs/content/users/usage/developer-tools.md
```diff
@@ -40,7 +40,14 @@ For example:
 
 When using `ddev composer create` your project should be essentially empty or the command will refuse to run, to avoid loss of your files.
 
-By default `--no-plugins` and `--no-scripts` options are used while cloning the project and `composer install` is executed afterwards as long as `--no-install` is not provided. When using `--preserve-flags` plugins and scripts are allowed during the cloning step.
+Standard order of operations for `ddev composer create`:
```
rfay, line 43: I sure like this section!

```suggestion
`ddev composer create` does these things in this order:
```

docs/content/users/usage/developer-tools.md
```diff
@@ -40,7 +40,14 @@ For example:
...
+
+1. `composer create-project --no-plugins --no-scripts --no-install` clones a bare Composer package without any additional actions.
+2. `composer run-script post-root-package-install` runs if `--no-scripts` is not provided manually.
```
rfay, line 46:
```suggestion
2. `composer run-script post-root-package-install` runs if `--no-scripts` is not given as a flag to `ddev composer create`.
```

docs/content/users/usage/developer-tools.md
```diff
@@ -40,7 +40,14 @@ For example:
...
+3. `composer install` runs if `--no-install` is not provided manually.
```
rfay, line 47:
```suggestion
3. `composer install` runs if `--no-install` is not given as a flag to `ddev composer create`.
```

docs/content/users/usage/developer-tools.md
```diff
@@ -40,7 +40,14 @@ For example:
...
+4. `composer run-script post-create-project-cmd` runs if `--no-scripts` is not provided manually.
```
rfay, line 48:
```suggestion
4. `composer run-script post-create-project-cmd` runs if `--no-scripts` is not given as a flag to `ddev composer create`.
```

rfay (commented):
I just looked through the docs, not code, had minor comments.

rfay:
@rpkoller would you be willing to validate the quickstarts with this PR? I know it's loads of work, but your attention is so valuable.

rpkoller:
> @rpkoller would you be willing to validate the quickstarts with this PR? I know it's loads of work, but your attention is so valuable.

argh that slipped through again. was on my to do list then i completely forgot. will try to run through either tomorrow or on thursday hopefully.

cmd/ddev/cmd/composer-create.go
```diff
@@ -27,25 +27,44 @@ var ComposerCreateCmd = &cobra.Command{
...
-		osargs := []string{}
+		var osargs []string
 		if len(os.Args) > 3 {
 			osargs = os.Args[3:]
-			osargs = nodeps.RemoveItemFromSlice(osargs, "--yes")
-			osargs = nodeps.RemoveItemFromSlice(osargs, "-y")
-			osargs = nodeps.RemoveItemFromSlice(osargs, "--preserve-flags")
+			var expandedOsargs []string
+			for _, osarg := range osargs {
+				// If this is a combined short option like "-test", split it into "-t -e -s -t"
```
rfay, line 45: Weird that we have to do this, when Cobra has these capabilities. I know there's *lots* of history in this code.

cmd/ddev/cmd/composer-create.go
```diff
@@ -213,38 +280,9 @@ ddev composer create --preserve-flags --no-interaction psr/log
...
-				"--verbose",
-			}
-
-			for _, osarg := range osargs {
-				for _, supportedArg := range supportedArgs {
-					if strings.HasPrefix(osarg, supportedArg) {
-						composerCmd = append(composerCmd, osarg)
-					}
+			for _, validCreateArg := range validCreateArgs {
+				if isValidComposerOption("install", validCreateArg) {
```
rfay, line 284: Thanks for splitting this out.

cmd/ddev/cmd/composer-create.go
```diff
@@ -329,8 +379,6 @@ for basic project creation or 'ddev ssh' into the web container and execute
 }
 
 func init() {
-	ComposerCreateCmd.Flags().BoolP("yes", "y", false, "Yes - skip confirmation prompt")
```
rfay, line 332: It's always great to get evil things out of the init() functions

cmd/ddev/cmd/composer-create.go
stasadev: I don't think we are using Cobra here, because `osargs = os.Args[3:]` at the top
In any case, there is always a way to improve it in the future.

stasadev:
> we're in hell here and we're going to have to figure out how to get out of it

It sounds like a challenge and I like it. I haven't touched any code related to the mutagen, a few restarts here seems redundant. But still, I think I'll come back to it some time.

> Note that a --no-interaction install of Craft CMS does not result in a working site, because it still needs to do a ddev craft setup to work. (#6246)

I don't think we can update this quickstart with @mandrasch's technique right now, but after some time when this code is mature, we can update it for `--no-interaction`.

mandrasch:
> The original idea was just to allow composer create-project without doing it in a subdirectory. I'm not sure why we didn't end up doing all the work in a temporary directory and then copy it into the project, and I'm not sure why we can't do that now.

Just as sidenote: I wondered about this when I first tried to understand why there is a special command for composer create in ddev. But I haven't had time to find out why (seems like there is git involved, so directory really must be empty?  https://github.com/composer/composer/issues/7455#issuecomment-536991910) 🤓 Thanks for all the effort which went into making this work! 🙏

mandrasch:
Regarding CraftCMS, it's a bit confusing - the official docs state this (https://craftcms.com/docs/5.x/install.html#quick-start):

```bash
mkdir my-craft-project
cd my-craft-project/
ddev config --project-type=craftcms --docroot=web --php-version=8.2
ddev composer create -y --no-scripts "craftcms/craft"
ddev craft install
```

Using `--no-scripts` means, the post-create hook with `craft setup` is never used anyway. And at least I always used `ddev craft install` afterwards as well. There is no need for `--no-interaction` so far when following the official docs approach. Hope this does not lead to more confusion!

docs/content/users/quickstart.md
```diff
@@ -54,7 +54,7 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
     ```bash
     mkdir my-cakephp-site && cd my-cakephp-site
     ddev config --project-type=cakephp --docroot=webroot
-    ddev composer create --prefer-dist cakephp/app:~5.0
+    ddev composer create --prefer-dist --no-interaction cakephp/app:~5.0
     ddev cake
```
rpkoller, line 58:
```suggestion
```
i think the `ddev cake` line can be dropped since it is missing an actual command doing something. at the moment it results in:  ```$> ddev cake
No command provided. Choose one of the available commands.```

the cake php docs also dont use the cake command setting up a new project: https://book.cakephp.org/5/en/installation.html#installation-using-ddev

stasadev: Yes, it makes sense to remove that line, thanks.

stasadev:
@rfay and I discussed our current implementation for `ddev composer create`, and decided to try using `composer create-project` as is.

I did that and it generally worked, but when I started testing it with our quickstarts, things went wrong.

Our current logic:
- run `composer create-project --no-plugins --no-scripts --no-install`, which creates the project skeleton including important CMS settings, like `.env.example`.
- run `composer run-script post-root-package-install`
- restart the project, which creates `.env` from `.env.example`, *a really important step* for some frameworks like Laravel and CraftCMS
- run `composer install`
- run `composer run-script post-create-project-cmd`, which *heavily depends on CMS settings* in `.env` for some frameworks like Laravel and CraftCMS

For example:

Laravel quickstart creates `.env` with settings for sqlite, and runs database migrations. Without manual `composer run-script post-create-project-cmd` we don't run it for our mariadb.

CraftCMS asks some interactive questions and without DDEV override it suggested me this:
```
Database server name or IP address: [127.0.0.1]
```
instead of
```
Database server name or IP address: [db]
```

So we are tied to the current implementation.

rfay:
Thanks for studying so carefully!

stasadev:
> DDEV does the magic for .env in ddev restart (I also wondered why ddev's composer create needs a ddev restart)

And it also restarts Mutagen for macOS users.

---

> faster installations without a ddev restart

We could probably review what we do inside `ddev restart` for the CMS settings and do it without restarting.
The problem is that when we start the project, it checks many different settings, so it's just easier to restart it.

---

> install cms with CMSes regular CLI command, evaluates `.env` and so prefill should be correct ddev value
> `ddev craft install/craft` or `typo3 install:setup`

For CraftCMS: it runs `php craft setup/welcome` inside `post-create-project-cmd` script (when you run in without `--no-scripts`)
For Laravel: it runs `php artisan migrate` inside `post-create-project-cmd` script
As so on.

If we move to the original behavior for `composer create-project`:

1. this will mean that some actions will have to be performed twice (as in `post-create-project-cmd`). Of course, we can use a flag for `--no-scripts`, but this will change the initial behavior for `composer create-project`.
2. we will need to add all these framework-specific commands such as `ddev craft install/craft`, `ddev artisan migrate` and so on to our quickstarts again, which will increase the maintenance burden on DDEV. Today the framework has this set of commands, tomorrow they will add something else, so I don't like this idea. Notice how we cleaned up our quickstarts so that they will work all the time, even if the framework changes something inside the `post-create-project-cmd`. And we achieve this only because we intercept the `composer create-project` halfway before it executes the `post-create-project-cmd`.

mandrasch:
Great additions, thanks for insights @stasadev 👍👍 

After reading your post, I noticed the (third) big question is: Does DDEV want devs to follow DDEV quickstart or the official docs?

Because https://laravel.com/docs/11.x/installation for instance also says "run migrate if you need db". 

My personal experience is that DDEV automagic also might hide some important commands which are needed for deployment later on (but this argument is only true for inexperienced beginners + I wasn't aware of all the post create hooks in Composer since there were not executed in the past in DDEV)

mandrasch:
PS: I also wonder what framework/CMS creators would prefer - no scripts or with scripts? 🤔 

PPS: When I think about a DDEV UI in a far future (maybe some day 😁), this would also be a question - guess no-interaction would be needed here. 

(And thanks for the exchange, super interesting to learn more about how this works👍)

stasadev:
> Does DDEV want devs to follow DDEV quickstart or the official docs?

That's a good question. Usually, the `composer create-project` configures everything for you, but sometimes it doesn't 🙂 (for example, Drupal requires more actions). I think we need to have a certain balance, if the framework provides everything in `composer create-project` then we rely on it completely, otherwise we add more steps to our quickstarts.

> Because https://laravel.com/docs/11.x/installation for instance also says "run migrate if you need db".

Yes, it's in the Laravel documentation and also in the `post-create-project-cmd` script. They added it to the script quite recently, after they switched to SQLite in Laravel 11.

rfay:
> Does DDEV want devs to follow DDEV quickstart or the official docs?

We want devs to be successful in their own world, and be comfortable there. The quickstarts are normally adaptations of the official docs. But where the official docs don't reflect a docker-based environment (or reflect an opinionated one of their own) we have to adapt (and usually simplify).

Sometimes we do too much for the user. Drupal users can get so complacent about what DDEV does with settings that they forget how the whole settings system works and then they don't know how to debug it, which a few years ago would have been natural and easy for every user.

PR #6304, opened 2024-06-12
test: Fix TestAcquiaPush and TestLagoonPush, add mariadb-old clients, fixes #6295

### LINKED ISSUE #6295: `ddev push acquia` fails with new mariadb client: `(ERROR at line 1: Unknown command '\-'.`
### Preliminary checklist

```
```
</details>

### Expected Behavior

`ddev push acquia` should work

### Actual Behavior

Now that we have the latest version of the mariadb client in the container, apparently used by `acli`, and which adds the break-everything directive:

* https://github.com/ddev/ddev/issues/6249

our [TestAcquiaPush](https://github.com/ddev/ddev/actions/runs/9448960180/job/26024113928) fails because acli is using mysql and then pushing database with that directive; Acquia seems to be running mysql:5.7:

```
$ mysql --version
mysql  Ver 14.14 Distrib 5.7.42-46, for debian-linux-gnu (x86_64) using  6.3
```

### Steps To Reproduce

`ddev push acquia`

### Anything else?

Ideas include:
* Solve the in-ddev-webserver client version problem, which is in progress (https://github.com/ddev/ddev/issues/6083#issuecomment-2116597297). That's good, but will still leave people confused, because they'll run into this problem and have to change their local version to mysql:5.7. It's what they should be doing (matching upstream with local) but we don't really want to have them have to learn this.
* Provide a wrapper script solution inside ddev-webserver, and have acquia.yaml use that

Note that acquia.yaml is due for maintenance anyway since this got fixed:
* https://github.com/acquia/cli/issues/1629

https://github.com/ddev/ddev/blob/0100013265cb48d21ff94b67ec7b770879daac8a/pkg/ddevapp/dotddev_assets/providers/acquia.yaml#L44-L50

rfay:
## The Issue

* #6295
* TestLagoonPush broke for the same reason
* We want to put a `neutral` mysql/mysqldump in there for people in this situation, so it's now in /usr/local/mariadb-old/bin

Acquia uses MySQL 5.7. We were using the default MariaDB:10.11 for the TestAcquiaPush test, and of course it has the mariadb-client 10.11.8, which has the broken directive in it.

This changes the tests to use MySQL 5.7, which now gets the MySQL 5.7 client in it, which should work OK.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -50,7 +50,7 @@ RUN chmod -f ugo+rwx /usr/local/bin /usr/local/bin/composer
 RUN cd /tmp && curl -L -s --fail -O https://dlm.mariadb.com/3607865/MariaDB/mariadb-10.11.6/repo/debian/pool/main/m/mariadb/mariadb-client_10.11.6+maria~deb12_$(dpkg --print-architecture).deb
 RUN cd /tmp && curl -L -s --fail -O https://dlm.mariadb.com/3607494/MariaDB/mariadb-10.11.6/repo/debian/pool/main/m/mariadb/mariadb-client-core_10.11.6+maria~deb12_$(dpkg --print-architecture).deb
 RUN set -x && cd /tmp && for item in *.deb; do dpkg-deb -x ${item} extracted; done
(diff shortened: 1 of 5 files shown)


stasadev (approved):
It looks good to me 👍

Do we need to add info about the old MariaDB somewhere in the docs, or will the release notes suffice?

stasadev:
I think a blog post will do the job.

stasadev (approved):
I checked it again, all good, thank you.

PR #6305, opened 2024-06-12
fix: Don't terminate mutagen sessions on upgrade/version change

rfay:
## The Issue

This has apparently been an issue for quite some time, but if you change DDEV versions (upgrade, test a new one, whatever - or just change the `last_started_version` in ~/.ddev/global_config.yaml, it prompts you to poweroff. I guess I always did this... until recently. I didn't want to wait for it.

But then I would see that all of my other projects (the ones I didn't want to impact with a poweroff) were broken because their mutagen session was gone.

image

If people do not agree to a poweroff, do not delete their mutagen sessions.

Change the last_started_version in global config and `ddev start` a project.  Do not accept poweroff.The other projects should not be broken.

### DIFF
--- a/cmd/ddev/cmd/root.go
+++ b/cmd/ddev/cmd/root.go
@@ -175,8 +175,6 @@ func checkDdevVersionAndOptInInstrumentation(skipConfirmation bool) error {
 			if okPoweroff {
 				ddevapp.PowerOff()
 			}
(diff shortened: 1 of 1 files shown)


stasadev (approved):
Some time ago I switched to Mutagen on Linux and saw this error several times. But I thought the problem was that I was doing something wrong. Thank you!

PR #6307, opened 2024-06-13
refactor: set default drupal config to php 8.3 for drupal 11 compatibility, fixes #6306

### LINKED ISSUE #6306: Set default PHP version to 8.3 for `ddev config --project-type=drupal` for Drupal 11 release late July 2024
### Preliminary checklist

### Problem

Drupal 11 will be released the week of July 29, 2024.

https://www.drupal.org/about/core/policies/core-release-cycles/schedule

It would be nice if this command gave you PHP 8.3 before the week of July 29, 2024:

`ddev config --project-type=drupal --docroot=web`

### Workaround

`ddev config --project-type=drupal --docroot=web --php-version=8.3`

### Solution

I found https://github.com/rfay/ddev/blob/master/pkg/nodeps/php_values.go where it looks like the default PHP version for all systems is PHP 8.2. Perhaps other systems are not ready for PHP 8.3 ... so should the default `drupal` config be changed from using `PHPDefault` to PHP 8.3? (I am not sure where)

gitressa:
## The Issue

- Fixes #6306

Make default `drupal` config Drupal 11 compatible.

### DIFF
--- a/pkg/ddevapp/apptypes_test.go
+++ b/pkg/ddevapp/apptypes_test.go
@@ -65,7 +65,7 @@ func TestConfigOverrideAction(t *testing.T) {
 		nodeps.AppTypeCraftCms:     nodeps.PHPDefault,
 		nodeps.AppTypeDrupal6:      nodeps.PHP56,
 		nodeps.AppTypeDrupal7:      nodeps.PHP82,
-		nodeps.AppTypeDrupal:       nodeps.PHPDefault,
+		nodeps.AppTypeDrupal:       nodeps.PHP83,
 		nodeps.AppTypeLaravel:      nodeps.PHP82,
 		nodeps.AppTypeMagento:      nodeps.PHPDefault,
 		nodeps.AppTypeMagento2:     nodeps.PHP82,
--- a/pkg/ddevapp/config_test.go
+++ b/pkg/ddevapp/config_test.go
@@ -230,7 +230,7 @@ func TestConfigCommand(t *testing.T) {
 	testMatrix := map[string][]string{
 		"magentophpversion": {nodeps.AppTypeMagento, nodeps.PHPDefault},
 		"drupal7phpversion": {nodeps.AppTypeDrupal7, nodeps.PHP82},
-		"drupalphpversion":  {nodeps.AppTypeDrupal, nodeps.PHPDefault},
+		"drupalphpversion":  {nodeps.AppTypeDrupal, nodeps.PHP83},
 	}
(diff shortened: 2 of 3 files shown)


rfay (changes requested):
Wouldn't it be better to do this for `ddev config --update` when Drupal 11 is detected, rather than doing it all for drupal8+ ?

As a matter of fact, that is already true: https://github.com/ddev/ddev/blob/a319644bf269cd49738dbb9fe230907769435cae/pkg/ddevapp/drupal.go#L383-L385

So I'm not understanding completely why this needs to be done.

stasadev:
I explained my thoughts on this in https://github.com/ddev/ddev/issues/6306#issuecomment-2165226994

I think the basic idea is to set the default PHP for Drupal to the same version that we explicitly set in our quickstarts (targeting non-EOL Drupal).

As far as I understand it, this will simplify things for people who *don't* use the Drupal quickstart.

Of course I could be wrong because I don't use it.

gitressa:
PHP 8.3 is recommended for Drupal 10.3 and Wordpress

- https://www.php.net/releases/8.3/en.php
- https://docs.pantheon.io/release-notes/2024/04/php-8-3-recommended
- https://www.drupal.org/project/drupal/issues/3446936

Very soon Drupal 11 will be the official Drupal release (July 2024) where PHP 8.3 is required.

https://www.drupal.org/docs/getting-started/system-requirements/php-requirements

Updating the default DDEV PHP constant `const PHPDefault = PHP82` to `PHP83` in DDEV is another option, making this PR redundant ...

https://github.com/ddev/ddev/blob/master/pkg/nodeps/php_values.go#L18

It would also make `--php-version=8.3` redundant for Drupal 10 and 11 Quickstarts:

https://ddev.readthedocs.io/en/stable/users/quickstart/#drupal

rfay:
Thanks for keeping us in touch with Drupal needs, and for doing this PR.

However, in part due to your concerns earlier this year, we already implemented the `drupal` type and the `ddev config --update` option which updates the default to PHP 8.3 when Drupal 11 is discovered. 

Please review how that works for you with Drupal 11 using the regular quickstart and see how it does. 

Closing this for now, happy to continue the conversation.

stasadev:
Thanks for chasing this, but it looks like the Drupal quickstart works fine with `--php-version`, and we actually need this flag because people don't use the same version of DDEV, and don't update it often. Adding the PHP flag make DDEV consistent for all DDEV users.

PR #6309, opened 2024-06-13
fix: Support MySQL 5.5 and 5.6 on ddev-webserver by using MySQL 5.7 clients

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/6083#issuecomment-2165714455

It seems that people still do use MySQL 5.6 and maybe somebody out there still uses 5.5

We can't build the clients for those long-obsolete versions for Debian 12 Bookworm, but they should work fine with the MySQL 5.7 client. 

If MySQL 5.5 or 5.6, use the mysql 5.7 client in the ddev-webserver.

Try using mysql 5.6. `ddev exec mysql --version` should show the 5.7 client.

I tested this on Gitpod with mysql 5.5 and 5.6 and it seems to be fine.


stasadev (approved):
Tested and it looks good to me, thanks for adding the timeout.

PR #6310, opened 2024-06-13
docs: update instructions for cleaning up old `ddev` binaries

stasadev:
## The Issue

From Discord:
ddev version different when trying to upgrade
https://discord.com/channels/664580571770388500/1250826678096236554/1250826678096236554

Updates instructions to include more use cases.
Adds a tip to the "Upgrading DDEV" page.

Rendered version is at https://ddev--6310.org.readthedocs.build/en/6310/users/install/ddev-upgrade/


rfay (approved):
Thanks!

PR #6314, opened 2024-06-13
test: Attempt to debug TestProcessHooks intermittent failures, fixes #6313

### LINKED ISSUE #6313: TestProcessHooks is failing intermittently
TestProcessHooks is failing intermittently in a number of environments. for example:

* race-detection, https://github.com/ddev/ddev/actions/runs/9505897980/job/26201865009
* mutagen, https://github.com/ddev/ddev/actions/runs/9503012441/job/26192443776
* WSL2-docker-inside (NOT mutagen): https://buildkite.com/ddev/wsl2-docker-inside/builds/4466#019023c3-0677-43a8-8c1e-674be0d25e46

It seems to me like it's always happening appStart() after
`failed to app.Start() for task '{exec: "echo TestProcessHooks > /var/www/html/TestProcessHooks${DDEV_ROUTER_HTTPS_PORT}.txt"  Running task: Exec command 'echo TestProcessHooks > /var/www/html/TestProcessHooks${DDEV_ROUTER_HTTPS_PORT}.txt'}`

rfay:
## The Issue

* #6313

This has just failed so commonly it's worth figuring out

Perhaps the app.Stop() at end of the for loop is relevant. app.Stop() really isn't designed for repetitive use like it would otherwise be used here.

### DIFF
--- a/pkg/ddevapp/hooks_test.go
+++ b/pkg/ddevapp/hooks_test.go
@@ -44,7 +44,7 @@ func TestProcessHooks(t *testing.T) {
 		_ = os.Setenv(`DDEV_DEBUG`, origDdevDebug)
 	})
 	err = app.Start()
-	assert.NoError(err)
+	require.NoError(t, err)
 
 	// Create a composer.json so we can do actions on it.
 	fName := filepath.Join(app.AppRoot, "composer.json")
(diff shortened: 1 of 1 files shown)


rfay:
With the extra `Stop()` here it doesn't seem to fail.

PR #6319, opened 2024-06-17
docs: suggestion ddev --version, move mkcert -install to later

rfay:
## The Issue

* A fairly common question is "I upgraded DDEV but `ddev --version` still shows the old version.
* `mkcert` is installed as a dependency most places, so a separate mention of the download etc is not needed.

Update docs

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -20,7 +20,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     ### Install Script
 
-    The [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) is another option. It downloads, verifies, and sets up the `ddev` binary:
+    The [install script](https://github.com/ddev/ddev/blob/master/scripts/install_ddev.sh) is another option. It downloads, verifies, and sets up the `ddev` executable:
 
     ```bash
     # Download and run the install script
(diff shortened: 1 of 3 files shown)


docs/content/users/install/ddev-installation.md
```diff
@@ -107,11 +105,19 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
...
+    ```
+
     ### Homebrew (AMD64 only)
 
     ```bash
     # Install DDEV using Homebrew
     brew install ddev/ddev/ddev
+    # One-time initialization of mkcert
+    mkcert -install
+
```
stasadev, line 120:
```suggestion

    # One-time initialization of mkcert
    mkcert -install
```

docs/content/users/install/ddev-installation.md
```diff
@@ -94,7 +90,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     # Install DDEV
     sudo sh -c 'echo ""'
     sudo dnf install --refresh ddev
-    ```
+
+    # One-time initialization of mkcert
+    mkcert -install
```
stasadev, line 95:
````suggestion
    mkcert -install
    ```
````

File: docs/content/users/install/ddev-installation.md [hunk 3]
```diff
@@ -107,11 +105,19 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     yay -S ddev-bin
     ```
 
+    ```bash
+    # One-time initialization of mkcert
+    mkcert -install
+    ```
```
stasadev, line 111: This can be in the same block as `yay` command. (I can't make a suggestion here.)

docs/content/users/install/ddev-installation.md
```diff
@@ -132,6 +138,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
         curl -fsSL https://ddev.com/install.sh | bash -s v1.21.4
         ```
 
+    ??? "Do you still have an old version after installing or upgrading?"
```
stasadev, line 141: I think it can also be copied after Debian/Ubuntu: because it's the most used Linux and people are unlikely to scroll down here.

stasadev (changes requested):

pkg/ddevapp/global_dotddev_assets/commands/host/self-upgrade
```diff
@@ -44,3 +44,6 @@ for ddev_path in $(which -a ddev | uniq); do
 
   esac
 done
+printf "\n'ddev --version' will show you the version you currently have installed.\n"
+printf "'which ddev' will show you where that executable is on your filesystem\n"
```
stasadev, line 48:
```suggestion
printf "'which ddev' will show you where that executable is on your filesystem.\n"
```

stasadev (approved):
Looks good to me.

PR #6323, opened 2024-06-17
docs: Improve ddev debug test by offering suggestions early [skip ci]

rfay:
## The Issue

`ddev debug test` can do a better job coaching people about debugging. This tries to focus them more on that at the beginning.

Eventually we should actually have a clearer diagnosis script maybe.

image


stasadev (approved):
Great, I like this simplification.

PR #6324, opened 2024-06-17
fix: generated gitignore is not working correctly

rfay:
## The Issue

In studying 
* #6317

I found that the project .ddev/.gitignore was not being created correctly at all. Mostly it was failing.

It's intended to look for named files, check them for #ddev-generated, and add them to the .gitignore if that's found *or* if there's an error in opening the file.

However, due to a bug in the logic, it wasn't successfully opening *any* file, and as a result was not adding files to the .gitignore that should be added (like if they had the #ddev-generated removed).

This PR:
* Removes a number of obsolete items from the CreateGitignore() call. I didn't recognize several of them, and didn't find them in a search.
* Fixes so we actually grep the right file before adding it to the .gitignore

Remove the "#ddev-generated" from an important file, like in providers dir, or traekfik/config. `ddev start` and it should show in `git status`

## Automated Testing Overview

I thought TestConfigGitignore would have caught this, but it seems to have been created for slightly different reason.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1405,7 +1405,9 @@ func PrepDdevDirectory(app *DdevApp) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Nice catch, I tested it and `.ddev/.gitignore` is updated correctly.

rfay:
This is embarrassing for me because I've maintained for years that DDEV did this right :(

PR #6325, opened 2024-06-18
docs: update the link to upgrade DDEV in faq.md

kiatng:
## The Issue
For someone who has very little knowledge on Linux, I end up trying to install DDEV by following the installation instructions (from docker desktop and so on) only to realize it's unnecessary.

It should help DDEV users like me.


rfay (approved):
Thanks!

rfay:
Thanks so much for pointing this out. I added another commit removing the duplicated instructions in the FAQ.

PR #6327, opened 2024-06-19
refactor: error handling for PR artifact comment [skip ci]

stasadev:
## The Issue

PR artifact comment is not created:

- https://github.com/ddev/ddev/actions/runs/9581560892

```
Unhandled error: TypeError: Cannot destructure property 'prNumber' of '(intermediate value)' as it is undefined.
```

I'm not sure what's wrong here.
I refactored the action to at least better understand the error.

### DIFF
--- a/.github/workflows/pr-artifacts-comment.yml
+++ b/.github/workflows/pr-artifacts-comment.yml
@@ -21,13 +21,21 @@ jobs:
           # https://github.com/oprypin/nightly.link/blob/master/.github/workflows/pr-comment.yml
           # https://github.com/AKSW/submission.d2r2.aksw.org/blob/main/.github/workflows/pr-comment.yml
           script: |
-            // get a PR number
-            const {owner, repo} = context.repo;
-            const pullHeadSHA = '${{github.event.workflow_run.head_sha}}';
-            const {prNumber, prRef, prRepoId} = await (async () => {
-              for await (const {data} of github.paginate.iterator(github.rest.pulls.list, {owner, repo})) {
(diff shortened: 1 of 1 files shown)


rfay:
You'll probably have to commit it in master in ddev-test/ddev to be able to see if it works of course.

stasadev:
> You'll probably have to commit it in master in ddev-test/ddev to be able to see if it works of course.

Done.

I tested it here:

- https://github.com/ddev-test/ddev/pull/1

rfay (approved):
Thanks so much for fixing this so fast!

stasadev:
I ran the PR build again (from the Actions tab) for:

- #6317

It has unrecognized SHA https://github.com/ddev/ddev/actions/runs/9584935394 which I can't find at https://api.github.com/repos/ddev/ddev/pulls

Another thing I noticed, there is not `PR Build` job in #6317 at the bottom.

I can't say for sure, but the last commit was `[skip ci]` and it was force-pushed, maybe it's an edge case that we can't handle.

Let's push #6317 again, but without `[skip ci]`.

rfay:
Sounds OK.

The sha is probably the built-in 
```
${{ secrets.GITHUB_TOKEN }}
```

stasadev:
No, it's this one:

```
${{ github.event.workflow_run.head_sha }}
```

It should match `github.rest.pulls.list[].head.sha` so that the corresponding PR number can be found.

My understanding is that it should just match the last SHA commit, but something went wrong here.

PR #6330, opened 2024-06-20
docs: revisit WSL Docker Desktop setup instructions

rfay:
## The Issue

Because of having to rebuild tb-wsldd-05 I wanted to capture some of the things that needed to be updated. 

This PR is created against upstream/ddev so perhaps we'll find out whether the setup was complete.


rfay:
Pulling. Not convinced tb-wsldd-05 is completely working, but too hard to chase. Will follow up if I see it failing.

PR #6243, opened 2024-05-24
feat: add support for MariaDB 11.4 LTS, fixes #6061

hussainweb:
## The Issue

MariaDB 11.4 is LTS. We can add support for that here.

## How This PR Solves The Issue

This PR adds MariaDB 11.4 to Makefile for the Docker image and to the allowed options list.

## Manual Testing Instructions

On a DDEV project, run `ddev config --database=mariadb:11.4` and make sure it works.

## Automated Testing Overview

I mainly tried to copy the PR #4902 and I couldn't find any tests to change here.

## Related Issue Link(s)

Fixes #6061.

PR #6334, opened 2024-06-20
build: use IPv4 inside a container, mariadb-client 11.4 for the webserver

stasadev:
## The Issue

From Discord https://discord.com/channels/664580571770388500/1252792407775514686

Something changed in Debian Bookworm, which results in ping being using IPv6 inside a container:

```
$ ddev exec ping -c 1 localhost
PING localhost(localhost (::1)) 56 data bytes
64 bytes from localhost (::1): icmp_seq=1 ttl=64 time=0.026 ms

--- localhost ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.026/0.026/0.026/0.000 ms
```

I don't have a pattern here, but I remember a time when it always used IPv4. We can make it more predictable to always prefer IPv4 over IPv6.

1. There is a commented section in `/etc/gai.conf`, that we can uncomment:

- https://wiki.archlinux.org/title/IPv6#Prefer_IPv4_over_IPv6

```
#
#    For sites which prefer IPv4 connections change the last line to
#
precedence ::ffff:0:0/96  100
```

2. After MariaDB 11.4 goes in, we can use `mariadb_repo_setup` to install newer `mariadb-client`, there is no need to pull this script every time (because the connection might fail and we don't want that):

- #6243

1. `ddev exec ping -c 1 localhost` should ping `127.0.0.1`
2. `ddev debug migrate-database mariadb:11.4 && ddev exec mysql -V` should show `11.4`

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/gai.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/gai.conf
@@ -1,0 +1,69 @@
+# This is the typical /etc/gai.conf from Debian Bookworm with this option enabled:
+# Prefer IPv4 over IPv6 for commands like "ping" inside a container.
+# https://wiki.archlinux.org/title/IPv6#Prefer_IPv4_over_IPv6
+
+# Configuration for getaddrinfo(3).
+#
+# So far only configuration for the destination address sorting is needed.
+# RFC 3484 governs the sorting.  But the RFC also says that system
+# administrators should be able to overwrite the defaults.  This can be
+# achieved here.
+#
+# All lines have an initial identifier specifying the option followed by
+# up to two values.  Information specified in this file replaces the
+# default information.  Complete absence of data of one kind causes the
+# appropriate default information to be used.  The supported commands include:
+#
+# reload  <yes|no>
+#    If set to yes, each getaddrinfo(3) call will check whether this file
+#    changed and if necessary reload.  This option should not really be
+#    used.  There are possible runtime problems.  The default is no.
+#
+# label   <mask>   <value>
+#    Add another rule to the RFC 3484 label table.  See section 2.1 in
+#    RFC 3484.  The default is:
+#
+#label ::1/128       0
+#label ::/0          1
+#label 2002::/16     2
+#label ::/96         3
+#label ::ffff:0:0/96 4
+#label fec0::/10     5
+#label fc00::/7      6
+#label 2001:0::/32   7
+#
+#    This default differs from the tables given in RFC 3484 by handling
+#    (now obsolete) site-local IPv6 addresses and Unique Local Addresses.
+#    The reason for this difference is that these addresses are never
+#    NATed while IPv4 site-local addresses most probably are.  Given
+#    the precedence of IPv6 over IPv4 (see below) on machines having only
+#    site-local IPv4 and IPv6 addresses a lookup for a global address would
+#    see the IPv6 be preferred.  The result is a long delay because the
+#    site-local IPv6 addresses cannot be used while the IPv4 address is
+#    (at least for the foreseeable future) NATed.  We also treat Teredo
+#    tunnels special.
+#
+# precedence  <mask>   <value>
+#    Add another rule to the RFC 3484 precedence table.  See section 2.1
+#    and 10.3 in RFC 3484.  The default is:
+#
+#precedence  ::1/128       50
+#precedence  ::/0          40
+#precedence  2002::/16     30
+#precedence ::/96          20
+#precedence ::ffff:0:0/96  10
+#
+#    For sites which prefer IPv4 connections change the last line to
+#
+precedence ::ffff:0:0/96  100
+
+#
+# scopev4  <mask>  <value>
+#    Add another rule to the RFC 6724 scope table for IPv4 addresses.
+#    By default the scope IDs described in section 3.2 in RFC 6724 are
+#    used.  Changing these defaults should hardly ever be necessary.
+#    The defaults are equivalent to:
+#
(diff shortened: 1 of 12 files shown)


rfay:
Rebased to pick up MariaDB 8.4 support

stasadev:
I added mariadb-client support for MariaDB 11.4 in the webserver.

I'll think about the test coverage on Monday.

I also changed `mysql-client-install.sh` a bit: use timeout only for `curl`, because we don't really need it for the whole script.

I repushed the image here:
```
docker pull ddev/ddev-webserver:20240620_stasadev_ipv4
```

rfay (changes requested):
I think you might like this better just putting /etc/gai.conf into ddev-webserver-base-files, https://github.com/ddev/ddev/tree/b38b631b5616f7c9190d8629128255e5abc8d3e1/containers/ddev-webserver/ddev-webserver-base-files/etc

containers/ddev-webserver/ddev-webserver-base-files/etc/gai.conf
```diff
@@ -0,0 +1,69 @@
...
+#
+#precedence  ::1/128       50
+#precedence  ::/0          40
+#precedence  2002::/16     30
+#precedence ::/96          20
+#precedence ::ffff:0:0/96  10
+#
+#    For sites which prefer IPv4 connections change the last line to
+#
+precedence ::ffff:0:0/96  100
```
rfay, line 58: For next time, a meaningful comment here would be good. It's hard to find the useful information in this file. Maybe we can slip it in the next time we work in ddev-webserver.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
```diff
@@ -0,0 +1,27 @@
...
+# This script is intended to be run with root privileges (normally in a docker build phase)
+
+set -eu -o pipefail
+
+DDEV_DATABASE_FAMILY=${DDEV_DATABASE%:*}
+if [ "${DDEV_DATABASE_FAMILY}" != "mariadb" ]; then
+  echo "This script is to be used only with a project using mariadb" && exit 1
+fi
+MARIADB_VERSION=${DDEV_DATABASE#*:}
+# Add all required MariaDB versions here
```
rfay, line 15: It's hard to understand this comment. Are we supposed to add more required MariaDB versions? I guess maybe it means "All mariaDB versions that require the compat packages?"

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ "${MARIADB_VERSION}" != "11.4" ]; then
+  echo "This script is not intended to run with mariadb:${MARIADB_VERSION}" && exit 0
```
rfay, line 17: Is this not an error? If it's an error, why `exit 0`? BTW, just `exit` should be same as `exit 0` but nothing wrong with exit 0.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mysql-client-install.sh
```diff
@@ -23,9 +23,9 @@ TARBALL_URL=https://github.com/ddev/mysql-client-build/releases/download/${TARBA
 
 # Install the related mysql client if available
 set -x
-cd /tmp && curl -L -o /tmp/mysql.tgz --fail -s ${TARBALL_URL}
+cd /tmp && timeout 30 curl -L -o /tmp/mysql.tgz --fail -s ${TARBALL_URL}
```
rfay, line 26: We may want to consider using curl's built-in timeout options for these cases in the future, https://unix.stackexchange.com/questions/94604/does-curl-have-a-timeout

rfay (approved):
This is working great. Future maintenance will be hard with some of the comments or lack of comments, so we'll want to return to those and try to sort them out. 

I note in passing that we will still be annoyed by the stupid mariadb whining:

```
$ ddev mysql --version
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
```

So I guess we'll have to add `ddev mariadb` and I guess people will have to learn how to use that tool.

containers/ddev-webserver/ddev-webserver-base-files/etc/gai.conf
stasadev: The comment is at the top of the file. I think it's better to have it in two places at once.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
stasadev: You're right, I thought about that, but then forgot. I will push a new image to replace the exit code.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mysql-client-install.sh
stasadev: Yes, we can use the curl timeout for cases like this, but I'm very confused by having multiple different timeouts in curl. So I'll leave it as it is.

rfay:
Sorry to delay this again. Thanks for your work on it.

stasadev:
I added missing `mariadb:11.4` to the README.md generation in GoReleaser.

And reworked the code a bit, now it should be clearer.

PR #6337, opened 2024-06-23
docs: Enable Rosetta on Apple Silicon machines with lima / colima

deviantintegral:

This encourages users to set up Rosetta for running Intel images on ARM, if needed.

On Apple Silicon, create the VM following these steps and run:

```
docker run --rm -it --platform=linux/amd64 busybox arch
```

You should get `x86_64`.

## Automated Testing Overview

Docs only!

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -27,9 +27,16 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     1. Install Lima with `brew install lima`.
     2. If you don't have the `docker` client (if `docker help` fails) then install it with `brew install docker`.
-    3. Create a 100GB VM in Lima with 4 CPUs, 6GB memory, and Cloudflare DNS. Adjust to your own needs:
+    3. Create a 100GB VM in Lima with 4 CPUs, 6GB memory, and Cloudflare DNS. Adjust to your own needs. For Intel systems run:
     ```
     limactl create --name=default --vm-type=vz --mount-type=virtiofs --mount-writable --memory=6 --cpus=4 --disk=100 template://docker
+    ```
+    On Apple Silicon systems, enable Rosetta in case you need to run Intel / x86_64 Docker images:
+    ```
+    limactl create --name=default --vm-type=vz --mount-type=virtiofs --mount-writable --memory=6 --cpus=4 --disk=100 --rosetta template://docker
+    ```
+    Then, set up the Docker context so it uses Lima to run containers:
+    ```
     docker context create lima-default --docker "host=unix://$HOME/.lima/default/sock/docker.sock"
     docker context use lima-default
     ```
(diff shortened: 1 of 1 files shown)


rfay (commented):
I'm not much in favor of people using Rosetta without knowing the full set of things they're dealing with, and they usually don't. Results can vary.

Would a link to https://ddev.com//blog/amd64-on-apple-silicon-ddev/ be more effective and cover several Docker providers for this use-case?

deviantintegral:
Using these flags doesn't change any defaults - it just makes Lima / colima match Orb and Docker Desktop in that they _can_ run x64 containers if requested. And, I've yet to find an x64 container that won't run well on Rosetta (I did have trouble with qemu back before Apple supported this).

Your blog post describes how to change the default platform, which seems to be contrary to your note that "results can vary".

rfay:
I apologize, but I feel that the very existence of Rosetta confuses people beyond usefulness. It's a brilliant technology, but doing support all the time I wish people would just disable it and leave it disabled.

However, I recognize the value, and obviously you're finding value with this approach. Would you be willing to do a guest blog on ddev.com explaining how you use this approach? I think that would be better than adding more options here where people are already confused...

Closing this for now because it adds too much complexity, hoping for another approach that will communicate the good stuff without complicating the basics.

deviantintegral:
So the thing is, nearly always I'm using Rosetta in a context outside of ddev. Most recently, it was to build x64 container images locally that I was pushing to a remote server for testing. I'd most recently reset my local docker environment, and copy-pasted the settings from the ddev docs, and was confused why my builds didn't work.

I have occasionally run x64 images for ddev. One time was an old project that needed an old Redis container, from before they had arm64 containers. And actually, the other was for Drupal work (but not ddev) running a Microsoft SQL container to analyze a database from DotNetNuke for a migration.

rfay:
I'm sure you know there's no need to use Rosetta to build amd64 images on arm64 (or arm64 on amd64). We build all the images typically on amd64.

> I have occasionally run x64 images for ddev. One time was an old project that needed an old Redis container, from before they had arm64 containers. And actually, the other was for Drupal work (but not ddev) running a Microsoft SQL container to analyze a database from DotNetNuke for a migration.

These are the really classic reasons, things like this!

deviantintegral:
Yeah, I was building images on my Macbook and pushing to my NAS, so I needed cross-compiling. It's also common for folks with Raspberry PIs, because they are so slow to build and many are on x64 workstations.

rfay:
The overall doc on multi-platform building is at https://docs.docker.com/build/building/multi-platform/ - it's amazing how well it works. 

The groundbreaking build scenario, which we started using before the mac M1 even came out, thanks to @dennisameling was explained in https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

A more recent update: https://www.docker.com/blog/how-to-rapidly-build-multi-architecture-images-with-buildx/

I'm not trying to talk you out of the excellent techniques you have, of course.

rfay:
Regardless, I certainly appreciate the contribution and hope that you'll be willing to explain your advanced usages in a blog post that people at your level can follow, https://github.com/ddev/ddev.com

PR #6346, opened 2024-06-24
fix: Drupal 10 and Drupal 11 settings have migrated, set state_cache

rfay:
## The Issue

Drupal 10 and Drupal 11 settings have migrated slightly.

The most important thing is $state_cache in Drupal 10.3

* Update Drupal 11 settings to current defaults
* Update Drupal 10 settings to current defaults and set `$settings['state_cache'] = TRUE;` in the settings.ddev.php
* Use Drupal 10 settings if version can't be detected
* Add comment to all Drupal and Backdrop settings.php/settings.ddev.php to show where they have come from.
* Removed the "drupal" settings.php/settings.ddev.php. Instead we'll use drupal10 by default if can't detect the actual version.

Spin up Drupal 11 and Drupal 10 sites. See that the warning about state_cache does appear.


rfay:
@rpkoller points out that the correct drupal10 settings.php is not being installed.

rfay:
@rpkoller I hope you'll be willing to take another look at this. Thanks for catching the earlier problem with wrong settings.php being installed.

pkg/ddevapp/backdrop.go
```diff
@@ -76,7 +76,7 @@ func createBackdropSettingsFile(app *DdevApp) (string, error) {
 	} else {
 		output.UserOut.Printf("Existing %s file does not include %s, modifying to include ddev settings", settings.SiteSettings, settings.SiteSettingsDdev)
 
-		if err = appendIncludeToDrupalSettingsFile(app.SiteSettingsPath, app.Type); err != nil {
+		if err = appendIncludeToDrupalSettingsFile(nil); err != nil {
```
stasadev, line 79: I didn't test it, but i think `nil` can cause panic here.

```suggestion
		if err = appendIncludeToDrupalSettingsFile(app); err != nil {
```

pkg/ddevapp/drupal.go
```diff
@@ -101,7 +105,7 @@ func manageDrupalSettingsFile(app *DdevApp, drupalConfig *DrupalSettings, appTyp
 	} else {
 		util.Debug("Existing %s file does not include %s, modifying to include DDEV settings", drupalConfig.SiteSettings, drupalConfig.SiteSettingsDdev)
 
-		if err := appendIncludeToDrupalSettingsFile(app.SiteSettingsPath, app.Type); err != nil {
+		if err := appendIncludeToDrupalSettingsFile(nil); err != nil {
```
stasadev, line 108:
```suggestion
		if err := appendIncludeToDrupalSettingsFile(app); err != nil {
```

stasadev (changes requested):

pkg/ddevapp/backdrop.go
rfay: Must be `app` anyway. thanks for the good eye.

rfay:
It's great when IDEs can refactor and change signatures for you... but wow! Thanks @stasadev

rpkoller (approved):
The artifacts were already ready so i was able to quickly test before dinner. tested on two drupal 10.3 instances that failed yesterday. removed the settings again so on startup the settings.php and settings.ddev.php got regenerated. and now all the settings default.settings.php, settings.ddev.php and settings.php have the state cache setting set to true properly added. thank you!

stasadev (approved):
Looks good to me.

PR #6349, opened 2024-06-25
docs: add link for config arguments as environment variables

stasadev:
## The Issue

From Slack https://drupal.slack.com/archives/C5TQRQZRR/p1719307140534229

It would be really useful if https://ddev.readthedocs.io/en/stable/users/configuration/config said when a config value gets exposed as an env var.

Adds a mention that some config arguments are exposed as env variables.
I haven't added the equivalent of an env variable for each config, as I find it hard to maintain and better to have it in one place.

See https://ddev--6349.org.readthedocs.build/en/6349/users/configuration/config/#setting-options

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -30,6 +30,13 @@ You can hand-edit the YAML files DDEV creates for you after running [`ddev confi
     ```
 
     Run `ddev help config` to see all the available config arguments.
+=== "Environment variable"
+
+    ```shell
+    ddev exec 'echo $DDEV_PHP_VERSION'
+    ```
+
+    Some config arguments are available as readable environment variables. See [environment variables provided to custom commands](../extend/custom-commands.md#environment-variables-provided).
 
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
I don't think this is accomplishing the goal. Environment variables cannot be used for configuration. This is pointing out that you can *read* some configuration values using environment variables, but the context here is configuration, which is about *writing* config values. I suspect that the original docs about environment variables are the ones we still need. Maybe we can just move that link out of the tab and send people to the available env vars.

rfay:
I don't think it's correct to think of config values and environment variables as equivalent or representative. I kind of wish they were, and maybe they could be some day. Be even if they were, the env vars would be just a mirror of the config, not a way to set it. So I recommend we keep thinking about this, and in the meantime, maybe we can figure out how to make this doc more prominent, https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/#environment-variables-provided

PR #6354, opened 2024-06-27
docs: Fix textlint usage, make textlint work on PRs

rfay:
## The Issue

A release in https://github.com/sapegin/textlint-rule-terminology broke all usages of "VS Code", which is common usage, and it now insists on "Visual Studio Code". However, that would make docs searches less likely to succeed, etc. 

See
* https://github.com/sapegin/textlint-rule-terminology/discussions/69#discussioncomment-9889155

Problems in 
* https://github.com/sapegin/textlint-rule-terminology/commit/787da86db165e0cf46d8e3680809a091238ce482#diff-a20da60dfec71fb7418916186c1249fe4746058a7f50348a43cd17f6157caca0R85
* https://github.com/sapegin/textlint-rule-terminology/releases/tag/v5.0.17
* https://github.com/sapegin/textlint-rule-terminology/releases/tag/v5.0.16

Also this (after being pulled) makes textlint/reviewdog work on PRs.


rfay:
Fixed this now, thanks to @sapegin help in https://github.com/sapegin/textlint-rule-terminology/discussions/69#discussioncomment-9889654 (thanks!)

The key is you have to find exactly how the rule is specified in https://github.com/sapegin/textlint-rule-terminology (in this case it's in https://github.com/sapegin/textlint-rule-terminology/blob/master/terms.jsonc#L83 - so you have to look at terms.json and use exactly the format shown. )

rfay:
@stasadev I'm still baffled why this seems to run or get reported only on master builds (https://github.com/ddev/ddev/runs/26731121109)

I did finally get my local textlint and related updated to latest so I could demonstrate the problem locally, https://github.com/ddev/ddev/blob/master/Makefile#L201-L210

rfay:
Ah, here's the reason it doesn't report:

image

rfay:
@stasadev textlint+reviewdog seems to work OK after landing on master. I tried pushing it into ddev-test/ddev and then in updated https://github.com/ddev-test/ddev/pull/1 it does do the check correctly on PRs, see https://github.com/ddev-test/ddev/actions/runs/9702012785/job/26776877801?pr=1

My only concern is that I'd rather not use `permissions: write`, which seems to be the thing here. I think it should be OK but would appreciate your concurrence.

stasadev (approved):
It looks good to me.

> I did finally get my local textlint and related updated to latest

I also updated my textlint to the latest version and confirmed its behavior.

PR #6359, opened 2024-06-28
fix: Check to see if IP address is local to determine whether to add to hosts

rfay:
## The Issue

In testing Let's Encrypt/Casual Hosting in
* https://github.com/ddev/ddev/pull/6317

I noted that even when the IP address of the machine was a local (non-localhost) address, it was still adding 127.0.0.1 to the /etc/hosts on that machine. For example, if the hostname DDEV wanted to use was `test.something.thefays.us` and `test.something.thefays.us` properly resolved to the current machine, DDEV would still add `test.something.thefays.us` to /etc/hosts as 127.0.0.1

This was a result of the logic in AddHostsEntriesIfNeeded() that said
```go
// If we had successful lookup and dockerIP matches
// with adding to hosts file.
if err == nil && len(hostIPs) > 0 && hostIPs[0] == dockerIP {
	continue
}
```

Essentially, it was only allowing the looked-up value of 127.0.0.1 (typical dockerIP) as "local", but it could be other things.

Instead of comparing with dockerIP, check to see if the IP address is assigned to this machine.

Note that the small overhead required by this is only incurred during `ddev start` and should be irrelevant.

* Create a DNS entry or wildcard that points to the local machine. For example, `*.something.ddev.com` might be an A record that contains `192.168.5.13`, the IP address of the local machine.
* `ddev config --project-name=xx`
* `ddev config --project-tld=something.ddev.com`
* `ddev start`

It should start up without trying to add `xx.something.ddev.com` to the /etc/hosts

## Automated Testing Overview

I wanted to add a test for IsLocalIP() but I couldn't figure out a useful test, since it would be dependent on the local configuration.

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -1,7 +1,6 @@
 package netutil
 
 import (
-	"fmt"
 	"net"
 	"os"
 	"slices"
@@ -47,16 +46,16 @@ func IsPortActive(port string) bool {
 // assigned to the local computer
 func IsLocalIP(ipAddress string) bool {
 	dockerIP, err := dockerutil.GetDockerIP()
-	
+
 	if err != nil {
 		util.Warning("Failed to get Docker IP address: %v", err)
 		return false
 	}
-	
+
 	if ipAddress == dockerIP {
 		return true
 	}
-	
+
 	localIPs, err := GetLocalIPs()
 
 	if err != nil {
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -46,9 +46,21 @@ func IsPortActive(port string) bool {
 // IsLocalIP returns true if the provided IP address is
 // assigned to the local computer
 func IsLocalIP(ipAddress string) bool {
-	localIPs, err := GetLocalIPs()
+	dockerIP, err := dockerutil.GetDockerIP()
+	
 	if err != nil {
-		fmt.Printf("Error fetching local IPs: %v\n", err)
+		util.Warning("Failed to get Docker IP address: %v", err)
+		return false
+	}
+	
+	if ipAddress == dockerIP {
+		return true
+	}
+	
+	localIPs, err := GetLocalIPs()
+
+	if err != nil {
+		util.Warning("Failed to get local IPs: %v", err)
 		return false
 	}
 
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -1,11 +1,14 @@
 package netutil
 
 import (
-	"github.com/ddev/ddev/pkg/dockerutil"
-	"github.com/ddev/ddev/pkg/util"
+	"fmt"
 	"net"
 	"os"
+	"slices"
 	"syscall"
+
+	"github.com/ddev/ddev/pkg/dockerutil"
+	"github.com/ddev/ddev/pkg/util"
 )
 
 // IsPortActive checks to see if the given port on Docker IP is answering.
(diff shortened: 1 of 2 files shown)


pkg/netutil/netutil.go
```diff
@@ -39,3 +42,42 @@ func IsPortActive(port string) bool {
...
 }
+
+// IsLocalIP returns true if the provided IP address is
+// assigned to the local computer
+func IsLocalIP(ipAddress string) bool {
+	localIPs, err := GetLocalIPs()
+	if err != nil {
+		fmt.Printf("Error fetching local IPs: %v\n", err)
+		return false
+	}
```
stasadev, line 53:
```suggestion
	dockerIP, err := dockerutil.GetDockerIP()
	
	if err != nil {
		util.Warning("Failed to get Docker IP address: %v", err)
		return false
	}
	
	if ipAddress == dockerIP {
		return true
	}
	
	localIPs, err := GetLocalIPs()

	if err != nil {
		util.Warning("Error fetching local IPs: %v", err)
		return false
	}
```

stasadev (changes requested):
I tested this on Linux.

`127.0.0.1` is a loopback address:

```
$ ddev start
Starting d10... 
The hostname d10.ddev.site is not currently resolvable, trying to add it to the hosts file 
DDEV needs to run with administrative privileges.
This is normally to add unresolvable hostnames to the hosts file.
You may be required to enter your password for sudo or allow escalation.
DDEV is about to issue the command:
  sudo --preserve-env=HOME /usr/bin/ddev hostname d10.ddev.site 127.0.0.1
 
[sudo] password for stas:
```

And I can see that any IP address that starts with `127` is a loopback:
```go
// IsLoopback reports whether ip is a loopback address.
func (ip IP) IsLoopback() bool {
	if ip4 := ip.To4(); ip4 != nil {
		return ip4[0] == 127
	}
	return ip.Equal(IPv6loopback)
}
```

I don't understand why the tests didn't fail here.

rfay:
Wow, thank you so much for your careful testing. Manual testing is the best! 

The reason it passed the tests is when `DDEV_NONINTERACTIVE` is true, we use the Host header and don't bother with name resolution.

You're the greatest!

pkg/netutil/netutil.go
rfay: This is also a tiny bit better performance-wise, thanks!

rfay (commented):
I tested this now with these project_tlds
* `local-host.thefays.us` resolved to 127.0.0.1 (worked fine)
* `traefik.thefays.us`  - on jumphost.thefays.us, where it resolving to actual host's IP address (jumphost.thefays.us) - worked, did not add to /etc/hosts
* `traefik.thefays.us` but on my local machine, where the IP address does not resolve to local, it added to /etc/hosts
*  `example.com`, which can't resolve anywhere, added to /etc/hosts

Thanks so much for catching the incorect behavior @stasadev

stasadev (approved):
Tested it again, it looks good to me.

I tried adding my local computer's IP to `/etc/hosts` with `192.168.1.100 test.example.com`.

And then:
```
ddev config --additional-fqdns=test.example.com && ddev restart
```

As a result, DDEV did not attempt to replace it with `127.0.0.1 test.example.com`.

But of course it only works with Let's Encrypt, so `test.example.com` didn't work in my browser.

I'm not sure, but maybe `GetLocalIPs()` only needs to be checked if Let's Encrypt is enabled?

rfay:
Thanks for the review, and for the careful check in the first place.

It works fine without Let's Encrypt, but you have to `ddev config global --router-bind-all-interfaces` for the router to listen to your other interface. 

```
rfay@rfay-tag1-m1:~/workspace/d10$ nslookup d10.artful.thefays.us
Server:		1.1.1.1
Address:	1.1.1.1#53

Non-authoritative answer:
Name:	d10.artful.thefays.us
Address: 192.168.2.249
```

image

stasadev:
Thanks for pointing this out, now I see it works.

rfay:
Thanks for making me test it to make sure! I was concerned at first. Then I tricked myself by walking to the coffee shop after setting it up. And it didn't work any more! Because my local IP had changed :)

PR #6361, opened 2024-06-30
feat: add `debug version-constraint` command

stasadev:
## The Issue

- https://github.com/ddev/ddev-xhgui/pull/31

The current approach when you need to add a version constraint to an add-on involves using `ddev debug capabilities`, which is not obvious.

This isn't good because capabilities aren't added for each DDEV release and you can't add a version constraint for some minor release.

Adds `debug version-constraint` command that allows you to check the semver constraint.

Try `ddev debug version-constraint ">=v1.23.0"`.

And a nice bonus is that the following bash command will also work with older versions of DDEV:

```bash
if [ "$(ddev debug version-constraint ">=v1.23.3" 2>/dev/null)" != "true" ]; then
    echo "This add-on requires DDEV v1.23.3 or higher, please upgrade." && exit 2
fi
```

## Release/Deployment Notes

Need to update https://github.com/ddev/ddev-addon-template/blob/8d7dac3113074fbf6e6b37696a24d837fa323056/install.yaml#L24-L26 after v1.23.3 release.

### DIFF
--- a/cmd/ddev/cmd/debug-capabilities_test.go
+++ b/cmd/ddev/cmd/debug-capabilities_test.go
@@ -2,12 +2,12 @@ package cmd
 
 import (
 	"encoding/json"
-	"github.com/ddev/ddev/pkg/nodeps"
-	asrt "github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"testing"
 
 	"github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/nodeps"
+	asrt "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // TestDebugCapabilitiesCmd tests that ddev debug capabilities works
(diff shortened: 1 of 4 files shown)


rfay (commented):
This seems useful, but what about just supporting the constraint directly in `ddev-get` so it can be specified in the install.yaml? Just add `ddev_version_constraint` to install.yaml?

stasadev:
I didn't think about adding `ddev_version_constraint` to `install.yaml`, because I was wondering how to make it work with any DDEV that supports `ddev-get` since v1.22.0.

Of course `ddev_version_constraint` seems much better in the long run.

> Just add `ddev_version_constraint` to `install.yaml`?

Okay, I'll use this instead, just like `ddev_version_constraint` in `config.yaml`.

rfay:
I don't think there's anything wrong with what you've already done here of course.

rfay:
I note that this basic problem is being handled by updating config.*.yaml in some add-ons like https://github.com/justafish/ddev-drupal-core-dev/blob/main/config.ddev-drupal-core-dev.yaml

It's a legitimate thing to do, but the constraint only shows up at `ddev restart`

stasadev:
> this basic problem is being handled by updating `config.*.yaml` in some add-ons

I wonder how this will work when you have installed multiple add-ons with `ddev_version_constraint` - it will probably pick the first or last constraint just by the name of the `config.*.yaml` file, which is not good.

rfay:
Yes, it would pick up the last one when merging, typically the last one alphabtically. But it can be forced with [override_config](https://ddev.readthedocs.io/en/stable/users/configuration/config/#override_config), I doubt that would be good.

stasadev:
Closing this in favor of:

- #6433

PR #6367, opened 2024-07-02
docs: remove bad link to deb.sury.org, confusing extra verbiage

rfay:
## The Issue

deb.sury.org removed support for buster packages today.

We had a reference to the buster packages, so docs link broke.

However, the reference was in a confusing repetitive section where better instructions had already been given above, so I ended up removing the paragraph where it was contained.


stasadev (approved):
I'm okay with this removal.

PR #6370, opened 2024-07-02
build: Use ENV = in Dockerfile

rfay:
## The Issue

We noticed warnings in the build about `ENV var <value>` being obsolete and Docker wanting `ENV var=<value>`

This fixes those.

I'm not even going to run tests on this one, we can absorb them later.


stasadev (approved):
Looks good to me.

I noticed one more place to update:
https://github.com/ddev/ddev/blob/eb86a27e76f6a56e92b8918e63016f53d33fcea3/pkg/ddevapp/config.go#L1063

rfay:
Brought this in, will be tested perhaps more successfully in
* https://github.com/ddev/ddev/pull/6374

PR #6371, opened 2024-07-03
build: Bump gitpod image for DDEV v1.23.3 [skip ci]

rfay:
## The Issue

After release we have to bump the gitpod image for use with v1.23.3. This does it.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: ddev/ddev-gitpod-base:20240614
+image: ddev/ddev-gitpod-base:20240702
 tasks:
   - name: build-run
     init: |


rfay:
Tested this after implementing in ddev-gitpod-launcher and it's working OK. I'm pretty sure we can just add and `apt update && apt install -y ddev` to the .gitpod.yml in both places and we won't have to update for every release any more.

stasadev:
> I'm pretty sure we can just add and `apt update && apt install -y ddev` to the .gitpod.yml in both places

I'm all for it.

rfay:
`apt-get install -y ddev` is added in
* https://github.com/ddev/ddev-gitpod-launcher/pull/44

We don't need to add it to the .gitpod.yml here because we build from source here. When it goes in there, we can update the developer docs here.

PR #6374, opened 2024-07-03
build: percona xtrabackup changed install technique

rfay:
## The Issue

Percona seems to have changed their method of installing xtrabackup.

https://docs.percona.com/percona-software-repositories/percona-release.html

Add their new technique to Dockerfile.


rfay:
It's possible that this change could affect people who added packages to the db container, but I'm not sure.

Issue #6357, opened 2024-06-28
Preserve the sanctity of #ddev-generated files via new customization options


nickchomey:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It was discovered during testing of the [Traefik v3 migration PR](https://github.com/ddev/ddev/pull/6317#issuecomment-2197231485) that an error related to `additional_hostnames` and `additional_fqdns` was due to my having overriden the `.ddev/traefik/static_config.yaml` file by removing the `#ddev-generated` line from the top. This prevented ddev from setting some new settings in static_config.yaml, notably 
```
core:
  defaultRuleSyntax: v2 
```

This `static_config.yaml` issue is just one more example of frustrations related to `#ddev-generated` files. 

For example, even without this migration to v3, in order to use the Traefik middleware plugins, I had to override both the global static_config.yaml and each project's traefik dynamic yaml. But, then if I wanted to change the project config.yaml - say to add `additional_hostnames` - those changes could only be applied if I added `#ddev-generated` to the yaml, which then wipes out my middleware config. So, you have to copy the middleware config, ddev restart to regenerate, re-add the middleware config, and ddev restart again. And then get confused when it inevitably gets out of sync again in the future. 

Likewise, when I excitedly shared DDEV with people in the Wordpress world recently, the immediate feedback by some people was "why does it keep breaking my `wp-config.php` file? This is ridiculous" I told them what was happening, but I suspect they just stopped using it altogether and will never return. Thats a great shame and shouldn't happen again.

---
### Root Problem Diagnosis

The problem, fundamentally, is that DDEV dynamically generates various config files at runtime, but doesn't expose sufficient ways for users to add customizations to them.

It does so with `config.*.yaml` and `docker-compose.*.yaml`, but doesn't do so with respect to the Traefik router, and some of the project types (at least Wordpress) generate config files that conflict with user workflows. 

This forces users to manually override (at DDEV's instruction!) files that really should be completely, 100%, off-limits. 

---
### Describe your solution

`#ddev-generated` files should be sacred and off-limits - all customizations should be done in other config files which then get dynamically merged at runtime.  

We should:

- [ ] remove any documentation or other mention that `#ddev-generated` files can be overridden
- [ ] add something that throws a fatal error (or at least a very prominent, stern and educational warning) if `#ddev-generated` is not present in files where it is expected
- [ ] ensure that all `#ddev-generated` files have extension mechanisms that are also well-documented

[...]

PR #6375, opened 2024-07-04
feat: Allow traefik merged dynamic configuration override with <project>.yaml

nickchomey:
## The Issue

Currently the only way to define any Traefik Middleware (or modify anything at all in the Traefik configs) is by removing `#ddev-generated` from the `/pkg/ddevapp/traefik_config_template.yaml` and `~/.ddev/traefik/static_config` which, as described in #6357, invites all sorts of other issues. The solution is to make all `#ddev-generated` files extensible. 

This was recently implemented for the static config via #6317. This PR is for the dynamic config

This PR adds a generalized merge mechanism to allow extra `/project/.ddev/traefik/dynamic_config.*.yaml` files to be merged into the `<project>.yaml` file, with the last alphanumeric file "winning". This can be used for adding middlewares as well as overriding any field in the dynamic config.

The `dynamic_config.*.yaml` files will also be able to contain {{go templates}} so that middlewares can dynamically use app variables to do things like rewrite app URLs in the response body (useful for Wordpress and Magento), or automatically "namespace" the middleware per active DDEV project.

Most users should never need to touch this - it would be best implemented via addons or other core code. 

* create a ddev project with `ddev config`
* copy/rename `/project/traefik/dynamic_config.middleware.yaml.example` to remove `.example` extension, which will use the fail2ban plugin
* Likewise for ~/.ddev/traefik/static_config.plugins.yaml` to add fail2ban globally
* `ddev start` 
* Check that routers and middleware are loading and working properly at http://localhost:10999/dashboard/#/
* Do the same for a 2nd project and check the traefik dashboard to see that separate middleware has been created per router
* Feel free to try defining other middlewares in both of those files, or changing the router assignments. Make sure to follow the same formatting, particularly with the `{{ .App.Name }}-` prefixes
* Other fields in the dynamic config can be changed as well - e.g. `tls:`

## Automated Testing Overview

A test has been added that injects a middleware definition via an extra config file. 

## Release/Deployment Notes

This should not affect anything else. It adds opt-in functionality, which would be best implemented via addons or other built-in mechanisms.


rfay:
Looks like the next hurdle is just TestCmdRestartJSON, https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8124#01907be5-07b3-4384-a0e9-aa1b1b01f567/15650-15654

> failed to unmarshal logLine='Failed to read Extension YAML file. Probably doesn't exist: open /Users/testbot/tmp/ddevtest/TestCmdWordpress192791668/.ddev/traefik/.traefik-dynamic-config-extensions-template.yaml: no such file or directory'

Looks like a secondary failure, first the .traefik-dynamic-config-extensions-template.yaml doesn't exist, second the output of that causes the TestCmdRestartJSON to fail.

You can run tests manually really easily in vscode, just click the little arrowhead. Since this is a TestCmd test, make sure you have the correct built version of ddev in path.

rfay:
I don't understand why you didn't fill in the PR template here. Please update it to explain. (But thanks for the manual testing instructions!)

Please edit the OP with the required information:

`## The Issue`

Explain there why this PR is necessary, 

`## How this PR solves the problem. 

`## Automated Testing Overview`

rfay (changes requested):
Just beginning to look at this, but your comment about the creation of traefik-dynamic-config-extensions-template.yaml.example says:

> I don't believe that existing projects will receive the /project/traefik/.traefik-dynamic-config-extensions-template.yaml.example file, as it is created with ddev config so that the file would be available prior to the first ddev start. Should that stay there, but also add some conditional logic to traefik.go to check for its existence, creating it if it doesn't exist?

This creation is being done in the wrong place. It should be processed configureTraefikForApp() in traefik.go, along with the other project-level traefik configuration. 

Also, you're giving it an inappropriate name. `.traefik_dynamic_config_extensions_template.yaml.example` is a hidden file. We don't want an example to be hidden.  I imagine it shouldn't have the name `template` in it either?

rfay:
Please improve the testing instructions above. I took a look at the example file and it has all kinds of go templating in it.

```
{{ $webEnv := .App.WebEnvironment }}
middlewareAssignments:
  {{ .App.Name }}-web-80-http:
      - {{ .App.Name }}-redirectHttps
      - {{ .App.Name }}-my-fail2ban
  {{ .App.Name }}-web-80-https:
      - {{ .App.Name }}-my-fail2ban
```

Do we expect users to be able to do that? Or was the example file supposed to have been processed?

rfay:
Your comments say
> Finally, you can debug it by going to http://localhost:10999/dashboard/ and seeing if there were any errors for Routers or Middleware. You can click on any that are red to see the error message

Shouldn't `ddev start` have failed long before you can see router or middleware failures here?

pkg/ddevapp/config.go
```diff
@@ -289,6 +283,24 @@ RUN echo "Built on $(date)" > /build-date.txt
...
 	}
 
+	//write example Traefik .traefik_dynamic_config_extensions_template.yaml.example. Contains some sample plugin configs
+	data, err := bundledAssets.ReadFile("traefik_dynamic_config_extensions_template.yaml")
+	if err != nil {
+		// Handle error (e.g., file not found in embedded FS)
+		log.Fatal(err)
+	}
+
+	err = WriteImageDockerfile(app.GetConfigPath("/traefik/.traefik-dynamic-config-extensions-template.yaml.example"), data)
```
rfay, line 293: Why would you add anything to the Dockerfile here? This looks all wrong. Project configuration has nothing to do with the Dockerfile

pkg/ddevapp/traefik_static_config_template.yaml
```diff
@@ -43,4 +43,4 @@ entryPoints:
   {{ range $port := .RouterPorts }}
   http-{{$port}}:
     address: ":{{ $port }}"
-  {{ end }}
+  {{ end }}
```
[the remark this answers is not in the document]
rfay: It seems to have no conflict, but you have removed the trailing newline apparently.

rfay (changes requested):
I'll take another look when I can understand the PR testing instructions and when Dockerfile stuff is removed and the example makes sense (I assume without go templating requirements?)

rfay:
You don't have to create a new PR. Just edit the OP. Pull down the little editor in the upper right of the OP :)

rfay:
Congrats on tests passing!

pkg/ddevapp/config.go
```diff
@@ -1425,7 +1425,7 @@ func PrepDdevDirectory(app *DdevApp) error {
...
+	err = CreateGitIgnore(dir, "**/*.example", ".dbimageBuild", ".ddev-docker-*.yaml", ".*downloads", ".homeadditions", ".importdb*", ".sshimageBuild", ".venv", ".webimageBuild", "apache/apache-site.conf", "commands/.gitattributes", "config.local.y*ml", "db_snapshots", "mutagen/mutagen.yml", "mutagen/.start-synced", "nginx_full/nginx-site.conf", "postgres/postgresql.conf", "providers/acquia.yaml", "providers/lagoon.yaml", "providers/pantheon.yaml", "providers/platform.yaml", "providers/upsun.yaml", "sequelpro.spf", "settings/settings.ddev.py", "traefik/", fmt.Sprintf("traefik/config/%s.yaml", app.Name), fmt.Sprintf("traefik/certs/%s.crt", app.Name), fmt.Sprintf("traefik/certs/%s.key", app.Name), "xhprof/xhprof_prepend.php", "**/README.*")
```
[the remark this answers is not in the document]
rfay: Yeah, has to be actually fixed. This ignores the entire traefik directory.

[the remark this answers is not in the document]
rfay: I will look at it today.

pkg/ddevapp/traefik_config_template.yaml
```diff
@@ -16,9 +16,8 @@ http:
       rule: Host({{ range $i, $h := $s.ExternalHostnames }}{{if $i}}, {{end}}`{{$h}}`{{end}})
       {{ end }}
       service: "{{$appname}}-{{$s.Service.InternalServiceName}}-{{$s.Service.InternalServicePort}}"
+      middlewares: []
```
rfay, line 19: I don't seem to be able to comment above, but the warnings on #ddev-generated at top of file should point people to the *correct* place to make changes. The comment remains correct, of course, but since there's a *good* way to do this now, point them to it.

pkg/ddevapp/dotddev_assets/traefik/dynamic_config.middleware.yaml.example
```diff
@@ -0,0 +1,37 @@
+#ddev-generated
+# In order to use Traefik Middleware plugins that are defined globally in ~/.ddev/traefik/static_config.*.yaml files you must insert here the dynamic YAML config provided by the plugin page->Install Plugin at https://plugins.traefik.io/plugins
```
rfay, line 2: Nit: Try to use readable line lengths here, not everybody has wrapping turned on.

rfay (changes requested):
This is going to be great, thanks!

-  [ ] As discussed, I think avoiding go templates in the dynamic config example. (preprocess the example). If you think it's super important, you can preprocess whtat the user puts in dynamic_config.*.yaml (even if the example doesn't show it) or you can add a section that demonstrates using the templating.
- [ ]  Add a README.md to the .ddev/traefik directory explaining how this is done
- [ ] Add the new technique to the docs
- [ ]  Needs test

pkg/ddevapp/traefik.go
```diff
@@ -409,6 +412,69 @@ func configureTraefikForApp(app *DdevApp) error {
...
+
+				// Unmarshal yaml plugin configs to intermediate map
+				extraConfigMap := make(map[string]interface{})
+				err = yaml.Unmarshal(extraConfigProcessedYAML.Bytes(), &extraConfigMap)
+				if err != nil {
+					return fmt.Errorf("error unmarshaling YAML to intermediate map: %s", err)
+				}
+
+				// merge processed file with baseConfig
+				err = mergo.Merge(&baseConfig, extraConfigMap, mergo.WithOverride)
```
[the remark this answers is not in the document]
rfay: I would prefer to use those, as otherwise we have lots of duplicate code. Maybe we can brainstorm how to do it.

pkg/ddevapp/traefik.go
```diff
@@ -370,6 +370,71 @@ func configureTraefikForApp(app *DdevApp) error {
 		UseLetsEncrypt:  globalconfig.DdevGlobalConfig.UseLetsEncrypt,
 	}
 
+	// The following block reads the project/.ddev/traefik/traefik_dynamic_config_extensions_template.yaml file, fills any template placeholders in it with templateData, and populates the Extension struct. Will later be added to templateData, to be used in generating the final .ddev/traefik/config/project.yaml file. This ultimately allows traefik extension plugins to be seamlessly used in the DDEV Traefik Router on a per-project basis. Add-ons can inject their extensions as-needed
+
+	extensionsYamlFilePath := projectTraefikDir + "/traefik_dynamic_config_extensions_template.yaml"
```
rfay, line 375: Why are we reading a file from the traefik directory instead of reading it from go:embed? There's no reason for this file to be written out is there?

rfay (changes requested):
Besides being all broken right now (we'll work on that) please use this logic:

* Process the config template from go:embed. You now have what would be the default config.
* For each dynamic_config.*.yaml, util.MergeYamlFiles() to that string. 
* The resultant string should be written to <projectname>.yaml, prepended with the ddev-generated comment and a reference to docs or dynamic_config.*.yaml readme

docs/content/users/extend/traefik-router.md
```diff
@@ -55,16 +55,8 @@ Project-specific configuration is automatically generated in the project’s `.d
 * The `certs` directory contains the `<projectname>.crt` and `<projectname>.key` certificate generated for the project.
 * The `config/<projectname>.yaml` file contains the configuration for the project, including information about routers, services, and certificates.
     * If you alter a `config/<projectname>.yaml` file, you need to remove the `#ddev-generated` from the top of the file, so it won't be overwritten on the next `ddev start`.
```
rfay, line 57: Let's just not suggest this. People may be familiar with the technique and they'll know what to do, and it will work.
```suggestion
```

docs/content/users/extend/traefik-router.md
```diff
@@ -55,16 +55,8 @@ Project-specific configuration is automatically generated in the project’s `.d
...
-
-      ```yaml
-      http:
-        middlewares:
-          example-redirectHttps:
-            redirectScheme:
-              scheme: https
-              permanent: true
-      ```
+    * However, as with extending the `static_config.yaml` file, the new way to modify a project's dynamic configuration is by adding `dynamic_config.*.yaml` files to `.ddev/traefik/`. These will be merged with the same alphanumeric sort **override** strategy.
```
rfay, line 58:
```suggestion
    * You can modify a project's dynamic configuration by adding `dynamic_config.*.yaml` files to `.ddev/traefik/`. These will be merged with an alphanumeric sort **override** strategy, so anything that is needed can be added or replaced.
```

docs/content/users/extend/traefik-router.md
```diff
@@ -55,16 +55,8 @@ Project-specific configuration is automatically generated in the project’s `.d
...
+    * A `dynamic_config.middlewares.yaml.example` file is now created in `.ddev/traefik/` which can be modified to define and assign different middlewares, as well as overriding any default parameters. Please refer to the `.ddev/traefik/README.md` file for more details.
```
rfay, line 59:
```suggestion
    * An example `.ddev/traefik/dynamic_config.middlewares.yaml.example` file is provided in `.ddev/traefik/. It can be modified to define and assign different middlewares, as well as overriding any default parameters. Please see `.ddev/traefik/README.md` file for more details.
```

pkg/ddevapp/dotddev_assets/traefik/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The `certs` directory contains default certs, and the `config` directory normally contains only
+default dynamic configuration in the `config/<project>.yaml` file.
+
+Additional dynamic configuration can be merged into the DDEV-generated `<project>.yaml` by
+adding `dynamic_config.*.yaml` files to the `traefik` directory. As with the global `static_config.*.yaml` files,
+merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a
+particular element of the YAML structure wins.
+
+There is a `dynamic_config.middleware.yaml.example` file that you can use as a starting point for adding
```
rfay, line 13:
```suggestion
The `dynamic_config.middleware.yaml.example` can be used as a starting point for adding
```

pkg/ddevapp/traefik.go
```diff
@@ -409,6 +409,78 @@ func configureTraefikForApp(app *DdevApp) error {
 		}
 	}
 
+	/* The following section reads the project/.ddev/traefik/dynamic_config.*.yaml files, fills any template placeholders in them with the
```
rfay, line 412: It templates these still? Or not?

pkg/ddevapp/traefik_config_middlewares_template.yaml
```diff
@@ -0,0 +1,29 @@
+#ddev-generated
```
rfay, line 1: Please rename this file to show its usage (It's a general *example* file and has nothing really to do with middlewares, right?)

Also, the example file does not seem to have been processed here, and still has go templates in it.

pkg/ddevapp/traefik_config_template.yaml
```diff
@@ -2,6 +2,11 @@
 # If you remove the ddev-generated line above you
 # are responsible for maintaining this file. DDEV will not then
 # update it, for example if you add `additional_hostnames`, etc.
+# However, it is now unnecessary and imprudent to take control of
+# this file, as there is now a dynamic merge mechanism. To use it
+# create <project>/.ddev/traefik/dynamic_config.*.yaml files which
+# will then be merged in alphanumeric order. See the readme in the
+# /traefik directory for more information
```
rfay, line 9: None of this is coming out in the resultant file, which just looks like this:
```
#ddev-generated
http:
    middlewares:
```

You'll need to prepend this string to the overall string before writing it out.

pkg/ddevapp/traefik.go
```diff
@@ -409,6 +409,78 @@ func configureTraefikForApp(app *DdevApp) error {
...
+	App's templateData (for	targeting the appropriate routers (e.g. {projectname}-web-80-http) or for rewriting the App's URL in the
+	response body), then merges their content into the base dynamic config YAML generated above, and is finally written to /project/.ddev/
+	traefik/config/<project>.yaml. Allows for adding middlewares, overriding settings, etc...
+	*/
+
+	extraDynamicConfigFiles, err := fileutil.GlobFilenames(projectTraefikDir, "dynamic_config.*.yaml")
```
rfay, line 418: All of this processing must happen *before* the traefik `<project>.yaml` is created, or at least before it's populated. Probably before it's created. 

* Template the base config into a string
* Find the dynamic config files if they exist
* util.MergeYamlFile() into the string
* Write it out

rfay (changes requested):
Unless I'm misunderstanding, the processing is not correct yet in the ConfigureTraefikForApp() function. 

Manual testing is working fine though, yay!

rfay:
I gave specific instructions at the places they belong. If you want to talk we can look at them together. It's always possible I don't understand something, but I explained that you need to create the config file *after* all the processing is done, not before.

pkg/ddevapp/traefik.go
rfay: it's OK with me to template-process those dynamic_config.*.yaml files, but please don't show it in the example.

rfay:
I hope to get a look today. Github's ideas of "Changes requested" is pretty mechanical and not very useful. It has no idea what's been addressed. The reality is that "a lot of water under the bridge and it has to be looked at again" :)

pkg/ddevapp/dotddev_assets/traefik/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The `certs` directory contains default certs, and the `config` directory normally contains only
+default dynamic configuration in the `config/<project>.yaml` file.
+
+Additional dynamic configuration can be merged into the DDEV-generated `<project>.yaml` by
+adding `dynamic_config.*.yaml` files to the `traefik` directory. As with the global `static_config.*.yaml` files,
+merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a
+particular element of the YAML structure wins.
+
+The `dynamic_config.middleware.yaml.example` can be used as a starting point for adding
```
stasadev, line 13:
```suggestion
The `dynamic_config.middlewares.yaml.example` can be used as a starting point for adding
```

pkg/ddevapp/traefik_config_template.yaml
```diff
@@ -2,6 +2,11 @@
 # If you remove the ddev-generated line above you
 # are responsible for maintaining this file. DDEV will not then
 # update it, for example if you add `additional_hostnames`, etc.
+# However, it is now unnecessary and imprudent to take control of
+# this file, as there is now a dynamic merge mechanism. To use it
```
stasadev, line 6:
```suggestion
# this file, as there is now a dynamic merge mechanism. To use it,
```

pkg/ddevapp/global_dotddev_assets/traefik/static_config.fail2ban.yaml.example
stasadev, line 1: Please update all the references in the project from `static_config.fail2ban.yaml` to `static_config.plugins.yaml`.
I see two `*.md` files that mention it.

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,117 @@ func configureTraefikForApp(app *DdevApp) error {
...
+				"# /traefik directory for more information\n\n" +
+				extraConfigProcessedYAML.String())
+
+		} else {
+
+			// if there aren't any dynamic_config.*.yaml files defined, then check if there's at least a dynamic_config.middlewares.yaml.example file.
+			// If not, create it from the go template, populating {{.App.Name}} where appropriate (e.g for router namespacing)
+			// This is needed because using dotddev_assets/traefik/ to inject the example file does not really allow for filling out the templates,
+			// which is desired so that users don't have to see or touch go templates
+			// Also, the example should only be created if there aren't any extra config files in use already, to avoid it being recreated on each start
```
stasadev, line 480: I think this `dynamic_config.middlewares.yaml.example` file needs to be constantly recreated (as it works for `static_config.plugins.yaml.example`), because if people copy it into some `dynamic_config.*.yaml` and change it to suit their needs, it will not appear again.

This means that if we decide to add some changes to this example file in the future, people with already changed configuration will not know about it.

pkg/ddevapp/traefik_dynamic_config_example_template.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - {{ .App.Name }}-my-fail2ban
+
+  middlewares:
+    {{ .App.Name }}-redirectHttps:
+      redirectScheme:
+        scheme: https
+        permanent: true
+    {{ .App.Name }}-my-fail2ban:
+      plugin:
+        fail2ban:
```
stasadev, line 19: I think this template should have a comment that you need to enable fail2ban with `static_config.plugins.yaml` or it won't work.

docs/content/users/extend/traefik-router.md
```diff
@@ -54,17 +54,9 @@ Project-specific configuration is automatically generated in the project’s `.d
...
-            redirectScheme:
-              scheme: https
-              permanent: true
-      ```
+    * You can modify a project's dynamic configuration by adding `dynamic_config.*.yaml` files to `.ddev/traefik/`. These will be merged with an alphanumeric sort **override** strategy, so anything that is needed can be added or replaced.
+    * An example `.ddev/traefik/dynamic_config.middlewares.yaml.example` file is provided in `.ddev/traefik/`. It can be modified to define and assign different middlewares, as well as overriding any default parameters. Please see `.ddev/traefik/README.md` file for more details.
+    * The extra config files may contain Go template placeholders, for injecting App-specific data into the config. But this is an advanced option that most can ignore. If you use them, note that template `{{ }}` placeholders cannot go at the start of a key (e.g. `{{.App.Name}}-web-80-http`). They must be preceded by any character, or used in YAML values. This means that if a middleware needs to be prefixed with the app's name, it will either need to be done manually or prefixed when an add-on creates its dynamic_config.*.yaml file from its own go template.
```
stasadev, line 59:
```suggestion
    * The extra config files may contain Go template placeholders, for injecting App-specific data into the config. But this is an advanced option that most can ignore. If you use them, note that template `{{ }}` placeholders cannot go at the start of a key (e.g. `{{.App.Name}}-web-80-http`). They must be preceded by any character, or used in YAML values. This means that if a middleware needs to be prefixed with the app's name, it will either need to be done manually or prefixed when an add-on creates its `dynamic_config.*.yaml` file from its own go template.
```

stasadev (changes requested):
Thank you @nickchomey for working hard on this!

I've tested this on multiple projects at once and it works.
I have some small suggestions.

pkg/ddevapp/traefik_dynamic_config_example_template.yaml
stasadev: Yes, that's correct, but people often don't read the docs or don't read them carefully enough.

It looks fine to me, I'm just wondering how it would look to others.

docs/content/users/extend/traefik-router.md
```diff
@@ -12,59 +12,52 @@ Before continuing, it's important to note that **very few users ever experiment
...
-* `config/default_config.yaml` contains global *dynamic* configuration, including pointers to the default certificates. It is possible to add other Traefik configuration in the `config` directory, which will apply to all projects. For example, a `config/router_middlewares.yaml` file might provide middleware implementations that would apply to all projects.
+- `.static_config.yaml` (a hidden file) is the configuration that gets used. It is not to be edited; it is generated from DDEV's base configuration while merging any files named `static_config.*.yaml`. It is read on router startup, and does not change until the router starts again (normally after `ddev poweroff`).
+- Additional static configuration may be added by adding `static_config.*.yaml` files, which will be merged into the generated `.static_config.yaml`. For example, a `static_config.plugin.yaml` might contain external Traefik plugins, or a `static_config.dnschallenge.yaml` might provide configuration for additional `certificatesResolvers`. Merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a particular element of the YAML structure wins.
```
stasadev, line 25: One more `plugin` > `plugins`:

```suggestion
- Additional static configuration may be added by adding `static_config.*.yaml` files, which will be merged into the generated `.static_config.yaml`. For example, a `static_config.plugins.yaml` might contain external Traefik plugins, or a `static_config.dnschallenge.yaml` might provide configuration for additional `certificatesResolvers`. Merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a particular element of the YAML structure wins.
```

stasadev (approved):
Looks good to me 👍

You can also rebase:

image

pkg/ddevapp/traefik_config_template.yaml
rfay: This comment content is lost in the later merge, so this heading is irrrelevant, you can remove it. To see this, just step through the code and look at the string that results from the MergeYamlFiles(), it's gone.

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,113 @@ func configureTraefikForApp(app *DdevApp) error {
...
+			return err
 		}
+
 		t, err := template.New("traefik_config_template.yaml").Funcs(getTemplateFuncMap()).ParseFS(bundledAssets, "traefik_config_template.yaml")
 		if err != nil {
 			return fmt.Errorf("could not create template from traefik_config_template.yaml: %v", err)
 		}
 
-		err = t.Execute(f, templateData)
+		err = t.Execute(dynamicConfigTemp, templateData)
```
rfay, line 417: We don't need to do this whole templating stage do we, since we template the results of the merge?

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,113 @@ func configureTraefikForApp(app *DdevApp) error {
...
+		*/
+
+		// find all dynamic_config.*.yaml files in the project's .ddev/traefik directory
+		extraDynamicConfigFiles, err := fileutil.GlobFilenames(projectTraefikDir, "dynamic_config.*.yaml")
+		if err != nil {
+			return err
+		}
+		var finalYaml []byte
+		// Check if extra config files were found. If not, no need to proceed.
+		if len(extraDynamicConfigFiles) > 0 {
```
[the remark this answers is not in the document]
rfay: To make this simpler, initialize finalYaml with its default before the MergeYamlFiles, and then the "if" would just do that merge. After the merge, add to or replace finalYaml with the new value. That would be the only content of the if. After that the templating would proceed against the result, whatever it is. (And of course, only needing to do one template operation, against the already-merged finalYaml)

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,113 @@ func configureTraefikForApp(app *DdevApp) error {
...
+				"# However, it is now unnecessary and imprudent to take control of\n" +
+				"# this file, as there is now a dynamic merge mechanism. To use it,\n" +
+				"# create <project>/.ddev/traefik/dynamic_config.*.yaml files which\n" +
+				"# will then be merged in alphanumeric order. See the readme in the\n" +
+				"# /traefik directory for more information\n\n" +
+				extraConfigProcessedYAML.String())
+
+		} else {
+
+			// Since there weren't any dynamic_config.*.yaml files, we can just write the contents of the base temp file to .ddev/traefik/<project.yaml>
```
rfay, line 476: This gets much simpler as suggested above, always process the finalYaml, no "else" required.

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,113 @@ func configureTraefikForApp(app *DdevApp) error {
...
+			return fmt.Errorf("could not write to traefikYamlFile: %v", err)
+		}
+
+		// Recreate the example config file on each ddev start, so as to both keep it up to date if we change things, as well as to always have
+		// an example available there to be modified.
+		// It is created from a template file, populating {{.App.Name}} where appropriate (e.g for router namespacing)
+		// This is needed because using dotddev_assets/traefik/ to inject the example file does not really allow for filling out the templates,
+		// which is desired so that users don't have to see or touch go templates
+
+		dynamicExampleFile := filepath.Join(projectTraefikDir, "dynamic_config.middlewares.yaml.example")
```
rfay, line 496: The example file (thanks for processing it) does need to be done, but it's a bit of work here in the middle of everything and distracts from the other important work. It could be in a separate function, etc. No need to change. But we just have such density here already that it's worthy thinking how to reduce it.

rfay (changes requested):
Lots of comments, most super-easy. The part that isn't quite as easy is simplifying the flow of ConfigureTraefikForApp(). It does what it needs to do now, but it works a little too hard at it, resulting in too many lines of code and a harder flow. If I'm not mistaken, we can do the templating just once, and we can remove a number of if/else cases. I'll be happy to help with this if it's not obvious enough from the comments. I know you've already worked hard on this, you're almost to the finish line!

pkg/ddevapp/traefik.go
```diff
@@ -394,19 +402,113 @@ func configureTraefikForApp(app *DdevApp) error {
...
+			// convert the output to a string and prepend "#ddev-generated" to the string, and then convert all of it to []byte to be written to file
+			finalYaml = []byte("#ddev-generated\n" +
+				"# If you remove the ddev-generated line above you\n" +
+				"# are responsible for maintaining this file. DDEV will not then\n" +
+				"# update it, for example if you add `additional_hostnames`, etc.\n" +
+				"# However, it is now unnecessary and imprudent to take control of\n" +
+				"# this file, as there is now a dynamic merge mechanism. To use it,\n" +
+				"# create <project>/.ddev/traefik/dynamic_config.*.yaml files which\n" +
+				"# will then be merged in alphanumeric order. See the readme in the\n" +
+				"# /traefik directory for more information\n\n" +
```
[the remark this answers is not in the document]
rfay: Single quotes mean something entirely different in go, sadly.

docs/content/users/extend/traefik-router.md
```diff
@@ -54,17 +54,9 @@ Project-specific configuration is automatically generated in the project’s `.d
...
-            redirectScheme:
-              scheme: https
-              permanent: true
-      ```
+    * You can modify a project's dynamic configuration by adding `dynamic_config.*.yaml` files to `.ddev/traefik/`. These will be merged with an alphanumeric sort **override** strategy, so anything that is needed can be added or replaced.
+    * An example `.ddev/traefik/dynamic_config.middlewares.yaml.example` file is provided in `.ddev/traefik/`. It can be modified to define and assign different middlewares, as well as overriding any default parameters. Please see `.ddev/traefik/README.md` file for more details.
+    * The extra config files may contain Go template placeholders, for injecting App-specific data into the config. But this is an advanced option that most can ignore. If you use them, note that template `{{ }}` placeholders cannot go at the start of a key (e.g. `{{.App.Name}}-web-80-http`). They must be preceded by any character, or used in YAML values. This means that if a middleware needs to be prefixed with the app's name, it will either need to be done manually or prefixed when an add-on creates its `dynamic_config.*.yaml` file from its own go template.
```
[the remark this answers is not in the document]
rfay: Maybe this can go in the README (more obscure) but be left out of docs (here?)

pkg/ddevapp/traefik.go
```diff
@@ -257,7 +257,15 @@ func PushGlobalTraefikConfig() error {
 	if err != nil {
 		return err
 	}
-	err = os.WriteFile(staticConfigFinalPath, []byte(resultYaml), 0755)
+
+	finalYaml := "#ddev-generated\n" +
+		"# Traefik \"static configuration\" file for DDEV\n" +
+		"# DO NOT EDIT this file, your edits will be replaced.\n" +
+		"# Although Traefik calls this \"static configuration\" it is changed\n" +
+		"# regularly to update `entryPoints`\n\n" +
```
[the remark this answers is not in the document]
rfay: The answer here is super easy. Just use single quotes around 'entrypoints'. It doesn't need the backticks anyway,

pkg/ddevapp/traefik.go
rfay: You're already creating a temp file with the contents if I'm not mistaken.

rfay: If it's not valid yaml with the templates in it, then you're right. But how is that going to work below, where you have the same problem? All of the files would have to be preprocessed in that case.

pkg/ddevapp/traefik.go
rfay: Thanks, it's easy and will just clean up a tiny bit.

pkg/ddevapp/traefik_config_template.yaml
rfay: You already have enough in the final yaml don't you?

rfay:
> The definitions will merge just fine, but the assignments from the 2nd one will override the 1st

Isn't that the intended behavior?

rfay:
I think what you're describing is the intended override behavior. In config.yaml we have two modes. The first (default) is merging, so `config.yaml` and `config.*.yaml` are all merged. Direct hits are overridden, but things like arrays (your case) are merged. For the explicit override case we have a special `override_config` directive, https://ddev.readthedocs.io/en/stable/users/configuration/config/#override_config, which makes it work as an explicit replacement instead of merge.

rfay:
I have decided not to accept this PR, mostly because of the explicit threat. I wish you well on your journey, and am sorry that this was such a difficult process.

PR #6377, opened 2024-07-04
fix: improve installation method for mariadb-client

stasadev:
## The Issue

TestWebserverDBClient fails fairly regularly on MariaDB 11.4
- https://github.com/ddev/ddev/actions/runs/9788192607/job/27025855851
- https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8122#01907bda-1be5-4dc2-a1cd-d1887c6e38d5

It seems that mariadb-client for 11.4 is not always installed, the offending line is:
```
timeout 30 mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}"
```

There are some problems with `mariadb_repo_setup`:
1. It adds three repos (server, tools, maxscale), but we only need the server:
	```
	$ ddev exec sudo apt update | grep mariadb
	Hit:4 https://downloads.mariadb.com/Tools/debian bookworm InRelease
	Hit:11 https://dlm.mariadb.com/repo/mariadb-server/10.11/repo/debian bookworm InRelease
	Hit:12 https://dlm.mariadb.com/repo/maxscale/latest/apt bookworm InRelease
	```
2. It runs `apt-get -qq update` (which can take a long time), but we really need to refresh only mariadb repo here.

This PR tries to make setup faster by skipping unnecessary options.

```
ddev config --database="mariadb:11.4"
ddev start
ddev exec mariadb -V
```

## Automated Testing Overview

The tests just have to work.

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -81,11 +81,6 @@ RUN curl https://nginx.org/keys/nginx_signing.key | gpg --dearmor \
 RUN echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] \
 http://nginx.org/packages/debian `lsb_release -cs` nginx" > /etc/apt/sources.list.d/nginx.list
 
-# Install mariadb_repo_setup to get mariadb-client from them directly
-RUN curl -LsSf https://r.mariadb.com/downloads/mariadb_repo_setup -o /usr/local/bin/mariadb_repo_setup && chmod +x /usr/local/bin/mariadb_repo_setup
-# Search for CHANGE_MARIADB_CLIENT to update related code.
-RUN /usr/local/bin/mariadb_repo_setup --mariadb-server-version="mariadb-10.11" --skip-maxscale --skip-tools
-
 RUN curl -sSLo /tmp/debsuryorg-archive-keyring.deb https://packages.sury.org/debsuryorg-archive-keyring.deb && \
     dpkg -i /tmp/debsuryorg-archive-keyring.deb && rm -f /tmp/debsuryorg-archive-keyring.deb && \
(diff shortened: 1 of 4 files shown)


containers/ddev-php-base/Dockerfile
```diff
@@ -84,7 +84,7 @@ http://nginx.org/packages/debian `lsb_release -cs` nginx" > /etc/apt/sources.lis
 # Install mariadb_repo_setup to get mariadb-client from them directly
 RUN curl -LsSf https://r.mariadb.com/downloads/mariadb_repo_setup -o /usr/local/bin/mariadb_repo_setup && chmod +x /usr/local/bin/mariadb_repo_setup
 # Search for CHANGE_MARIADB_CLIENT to update related code.
-RUN /usr/local/bin/mariadb_repo_setup --mariadb-server-version="mariadb-10.11"
+RUN /usr/local/bin/mariadb_repo_setup --mariadb-server-version="mariadb-10.11" --skip-maxscale --skip-tools
```
rfay, line 87: This is fine, but why would we install it in ddev-php-base anyway, I wonder.  Seems like it would make more sense just in ddev-webserver.

rfay (approved):
Looking great to me, thanks!

containers/ddev-php-base/Dockerfile
stasadev: I think it was only added to `ddev-php-base` because `apt-get install mariadb-client` was there in the first place.

I agree that it's better to have it in `ddev-webserver` because it's so tedious to wait for the Docker image to be ready and who knows how many times we'll need to change it in the future.

I will move it completely to `ddev-webserver`.

rfay (approved):
Looks great to me, thanks so much. Sorry to request moving from the base image, but it really is a pain building that one.

PR #6387, opened 2024-07-09
docs: stop checking aur.archlinux.org link [skip ci]

stasadev:
## The Issue

https://github.com/ddev/ddev/actions/runs/9847531874/job/27190957328#step:11:348

```
ERROR: 1 dead links found!
[✖] https://aur.archlinux.org/packages/ddev-bin/ → Status: 0
```

This is not the first time we see an error for this link, time to ignore it.

### DIFF
--- a/markdown-link-check.json
+++ b/markdown-link-check.json
@@ -56,6 +56,9 @@
     },
     {
       "pattern": "^https://community.chocolatey.org"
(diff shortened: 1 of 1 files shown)


rfay (approved):
I agree, thanks.

PR #6388, opened 2024-07-09
fix: update regex for traefik v3 usage

rfay:
## The Issue

* https://github.com/ddev/ddev/pull/6317 was supposed to convert to Traefik v3, but it wasn't quite successful. The rules were thinking they were using HostRegexp() but they weren't really successful.

* Use `ruleSyntax: v3` in our generated rules
* Use HostRegexp() in the v3 way - `HostRegexp() || HostRegexp()` etc

Try a wildcard like `ddev config --additional-hostnames="*.wild"`

## Automated Testing Overview

I think existing tests do it right.

### DIFF
--- a/pkg/ddevapp/traefik.go
+++ b/pkg/ddevapp/traefik.go
@@ -370,11 +370,11 @@ func configureTraefikForApp(app *DdevApp) error {
 		UseLetsEncrypt:  globalconfig.DdevGlobalConfig.UseLetsEncrypt,
 	}
 
-	// Convert externalHostnames wildcards like `*.<anything>` to `{subdomain:.+}.wild.ddev.site`
+	// Convert externalHostnames wildcards like `*.<anything>` to `[a-zA-Z0-9-]+.wild.ddev.site`
 	for i, v := range routingTable {
 		for j, h := range v.ExternalHostnames {
 			if strings.HasPrefix(h, `*.`) {
-				h = `{subdomain:.+}` + strings.TrimPrefix(h, `*`)
+				h = `[a-zA-Z0-9-]+` + strings.TrimPrefix(h, `*`)
 				routingTable[i].ExternalHostnames[j] = h
 			}
 		}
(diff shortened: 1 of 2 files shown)


rfay:
Would appreciate your review here @stasadev and @nickchomey

stasadev (approved):
Tested, it works fine.

h1nds1ght:
Hi @rfay ,

I am using ddev v1.23.4 with `additional_fqdns` also (nothing else), i.e.

```
name: myproject
additional_fqdns: [myproject.example.com]
```

Actually, if letsencrypt is enabled., this compiles to:

```
rule: Host(`^myproject\.ddev\.site$`,`^myproject\.example\.com$`)
```

Traefik logs then says:

```
Error while adding route for host error="error while adding rule HostSNI(`^myproject\\.ddev\\.site$`): invalid value for HostSNI matcher, \"^myproject\\\\.ddev\\\\.site$\" is not a valid hostname"
```

Could it be that this line in `traefik_config_template.yaml` inserts regexes instead of hostnames?

```
rule: Host({{ range $i, $h := $s.ExternalHostnames }}{{if $i}}, {{end}}`^{{$h | replace "." "\\."}}$`{{end}})
```

If I manually change `traefik/config/myproject.yaml` to the following, letsencrypt works as aspected:
```
rule: Host(`myproject.example.com`)
```

I am not that much into traefik, but my understanding is, that `HostRegexp` supports regexes, while `Host` only supports full hostnames, refering to [this](https://doc.traefik.io/traefik/routing/routers/#rule).

rfay:
Please open a new issue. I'm sure you're onto something. But iirc you can't use hostregex with let's encrypt. I'm on my phone on bike trip in Ireland. For now, use a single host name.

h1nds1ght:
Sorry, I will do so. Have a nice trip!

rfay:
I confirmed your result and reported in
* https://github.com/ddev/ddev/issues/6486

PR #6315, opened 2024-06-14
build: use $XDG_CONFIG_HOME in Gitpod for v1.23.2 release, for #6228 [skip ci]

stasadev:
## The Issue

- #6228

## How This PR Solves The Issue

Sets `XDG_CONFIG_HOME=/workspace/.config` inside the Gitpod container.
DDEV global configuration will be stored in a non-volatile location: `${XDG_CONFIG_HOME}/ddev`
I added a symlink from `~/.ddev` to `${XDG_CONFIG_HOME}/ddev`, so it won't break a custom user script (e.g. in hooks) that uses the `~/.ddev` location.

## Manual Testing Instructions

[Open in Gitpod](https://gitpod.io/#https://github.com/ddev/ddev/pull/6315)

Click Open in Gitpod, change some global config setting, e.g.:
```
ddev config global --internet-detection-timeout-ms 4000
```
Stop the workspace in https://gitpod.io/workspaces, open it again, and confirm that setting is the same:
```
ddev config global | grep internet-detection-timeout-ms
internet-detection-timeout-ms=4000
```
Check that `~/.ddev` points to `${XDG_CONFIG_HOME}/ddev`
```
ls -la ~ | grep .ddev
lrwxrwxrwx 1 gitpod gitpod   23 Jun 14 11:00 .ddev -> /workspace/.config/ddev
```

PR #6390, opened 2024-07-10
fix: omit ddev-router for Gitpod (for DDEV PRs) [skip ci]

stasadev:
## The Issue

`omit_containers: [ddev-router]` is not written to global DDEV config after:

- #6315

It led to problems with add-ons:

- https://github.com/ddev/ddev-phpmyadmin/issues/15

This is caused by the fact that `/workspace` appears to be a special folder that is overridden by Gitpod. It doesn't pick up any content from the pre-built Docker image.

Omits the `ddev-router` on the project init.

[Open in Gitpod](https://gitpod.io/#https://github.com/ddev/ddev/pull/6390)

Confirm that `ddev config global | grep omit` contains `ddev-router`.
Stop the Gitpod workspace, start it again and confirm the same.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -4,6 +4,7 @@ tasks:
     init: |
       # Compile ddev
       make
+      ddev config global --omit-containers=ddev-router
       ddev debug download-images
       mkcert -install
     command: |


stasadev (commented):
I tested and it worked and the `global_config.yaml` is saved on workspace stop/start.

rfay:
This only affects DDEV PRs, right, doesn't solve any of the general gitpod problems?

rfay (approved):
Looks right to me, thanks! This only affects "Open in Gitpod" on ddev/ddev, as far as I understand.

```
gitpod /workspace/d10simple (main) $ ddev config global |grep omit
omit-containers=[ddev-router]
```

stasadev:
Yes, this for `Open in Gitpod` only.

And I opened a similar PR for the ddev-gitpod-launcher:

- https://github.com/ddev/ddev-gitpod-launcher/pull/45

PR #6395, opened 2024-07-12
feat: add ddev/ddev-utilities to the pulled images

stasadev:
## The Issue

https://hub.docker.com/r/ddev/ddev-utilities is used in:

- `ddev debug test`
- `ddev launch`
- the docs

But we don't pull it.

Let's pull it to avoid a couple of seconds of waiting.

```
docker rmi ddev/ddev-utilities
DDEV_DEBUG=true ddev debug download-images
```


rfay:
Seems like a reasonable idea. I guess it escaped me that we were actually using this routinely. I sure wish `jq` was everywhere!

rfay (approved):
Thanks!

PR #6403, opened 2024-07-15
docs: correct config used to add extra-packages

matlads:
## The Issue

- wrong `ddev config` option for extra-packages

- documentation fix


rfay (approved):
Awesome, thanks for catching this, and thanks so much for experimenting with python use!

PR #6404, opened 2024-07-16
docs: Update 'ddev get' example

mfickers:
## The Issue

Searching the docs for "opensearch" will give no results. This might lead some to believe that it is not officially supported.

Update the `ddev get --list` example output in the docs to include a mention of all supported additional services.

-

## Automated Testing Overview

Changes are limited to docs, no testing needed.


tyler36:
Thank you for for also including the additional missing offical addons.

PR #6406, opened 2024-07-17
feat: expand `.ddev/.env.*` files on `ddev start`, add `ddev dotenv`, add variable expansion to `ddev add-on get`, fixes #4430

### LINKED ISSUE #4430: `ddev add-on get` should allow specifying optional settings

Some `ddev add-on get` add-ons can use minor configuration overrides. For example, the version of a redis image might want to be overridden. 

### Describe your solution

Use a flag, `--environment`? 

### Describe alternatives

People currently have to manually edit `docker-compose.*.yaml` and then remove the #ddev-generated from a file that is part of an add-on.

### Additional context

_No response_

stasadev:
## The Issue

- #4430

* Previously, variables from the `.env` file were expanded at the `docker-compose config` build stage for all `docker-compose.*.yaml` files, but were passed only to the `web` container. And the documentation says another thing:

	> An optional, project-level `.ddev/.env` file provides environment variables to all DDEV containers, including any additional services or add-ons.

	This PR changes the behavior to pass all variables from `.ddev/.env` to all containers.

* Adds support for custom long flags with `ddev dotenv set`, which allows to do things like:

	```
	ddev dotenv set .ddev/.env.redis --test-variable=foo --some-addon-value 123 --enable-debug --test-variable2 test
	```
	which adds `.ddev/.env.redis` with:
	```dotenv
	TEST_VARIABLE="foo"
	SOME_ADDON_VALUE="123"
	ENABLE_DEBUG=""
	TEST_VARIABLE2="test"
	```
	As you can see above, `--enable-debug` is converted to an empty value, which makes it easier for add-on maintainers to add simple flags without values, to check the existence of a variable instead of checking the value.

	Uppercase flags or short flags will not work, the below flags will be ignored:
	```
	ddev dotenv set .ddev/.env.redis --TEST=123 -a -B 123 -test=foo
	```

	These variables can be read with `ddev dotenv get .ddev/.env.redis --variable-name`.

	This can be useful in bash scripts:
	```bash
    TEST_VARIABLE=$(ddev dotenv get .ddev/.env.redis --test-variable 2>/dev/null)
    exit_code=$?
    if [[ $exit_code -eq 0 && "${TEST_VARIABLE}" == "foo" ]]; then
        # do something
    fi

* Uses custom `.ddev/.env.*` files to expand env variables in `docker-compose.*.yaml` services.
* The variables are also expanded to `pre_install_actions` and `post_install_actions` in `install.yaml` for add-ons.
* The variables are passed to the respective service containers, i.e. everything from `.env.redis` will go to the `redis` service.
* `.env.*` files can override values that are already defined in the service, for example, `.ddev/docker-compose.busybox.yaml`:

	```yaml
	services:
	  busybox:
	    environment:
	      - THIS_VARIABLE_CAN_BE_CHANGED_FROM_ENV=some-value
	```
	And the value from `.ddev/.env.busybox` will be available in the `busybox` container:
	```dotenv
    THIS_VARIABLE_CAN_BE_CHANGED_FROM_ENV=new-value
	```

1. Check updated `ddev/ddev-redis` add-on:

- https://github.com/ddev/ddev-redis/pull/24

```
ddev add-on get https://github.com/ddev/ddev-redis/tarball/20240717_stasadev_optional_version
ddev dotenv set .ddev/.env.redis --redis-tag=7 # the order of commands is not important in this case
cat .ddev/.env.redis # REDIS_TAG="7"
ddev dotenv get .ddev/.env.redis --redis-tag # REDIS_TAG="7"
ddev restart
ddev redis-cli INFO | grep redis_version # redis_version:7.4.0
# the command below will take into account the env from .ddev/.env.redis
ddev add-on get https://github.com/ddev/ddev-redis/tarball/20240717_stasadev_optional_version
```

2. Check updated `rfay/ddev-php-patch-build` add-on:

- https://github.com/rfay/ddev-php-patch-build/pull/2

```
ddev dotenv set .ddev/.env.php-patch-build --static-php-version=8.0.10 # the order of commands is important in this case, because env is used in post_install_actions
ddev add-on get https://github.com/stasadev/ddev-php-patch-build/tarball/20240717_stasadev_optional_version
cat .ddev/.env.php-patch-build # STATIC_PHP_VERSION="8.0.10"
ddev dotenv get .ddev/.env.php-patch-build --static-php-version # STATIC_PHP_VERSION="8.0.10"
ddev restart
# the command below will take into account the env from .ddev/.env.php-patch-build
ddev add-on get https://github.com/stasadev/ddev-php-patch-build/tarball/20240717_stasadev_optional_version
```


docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
 
 ## Adding Custom Configuration to an Add-on
 
-Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis-7`](https://github.com/ddev/ddev-redis-7) the `image` is set to `image: redis:7.2-alpine`. If you wanted to change this to use `7.0-alpine` instead, you would have two choices:
+Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis`](https://github.com/ddev/ddev-redis) the `image` is set to `image: redis:${REDIS_VERSION:-6-bullseye}`. If you wanted to change this to use `7.0-alpine` instead, you would have three choices:
```
rfay, line 108: This is a nit, but I would recommend not suggesting a completely different OS and build, as they won't necessarily be compatible.

```suggestion
Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis`](https://github.com/ddev/ddev-redis) the `image` is set to `image: redis:${REDIS_VERSION:-6-bullseye}`. If you wanted to change this to use `7.0-bullseye` instead, you would have three choices:
```

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
...
-  services:
-    redis:
-      image: redis:7.0-alpine
-```
+    ```bash
+    ddev get ddev/ddev-redis --environment="REDIS_VERSION=7.0-alpine"
+    ddev restart
+    ```
+
+2. Remove the `#ddev-generated` line in `.ddev/docker-compose.redis.yaml` and edit the relevant line.
```
rfay, line 117: Why don't we just stop recommending this, since we have better ways? The general docs explain this a few places, but it's not even needed here.

```suggestion
```

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
...
+3. Add a `.ddev/docker-compose.redis_extra.yaml` with the contents:
+
+    ```yaml
+    services:
+      redis:
+        image: redis:7.0-alpine
```
rfay, line 123: Recommend using similar example, 
```suggestion
        image: redis:7.0-bullseye
```

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
...
+    ```
 
-Using the second option (`docker-compose.redis-7_extra.yaml`) allows you to update to future versions of `ddev-redis-7` without losing your configuration, and without the upgrade being blocked because you removed the `#ddev-generated` line from the upstream `docker-compose.redis-7.yaml`.
+Using the first (`.env.redis`) and the third option (`docker-compose.redis_extra.yaml`) allows you to update to future versions of `ddev-redis` without losing your configuration, and without the upgrade being blocked because you removed the `#ddev-generated` line from the upstream `docker-compose.redis.yaml`.
```
rfay, line 126:
```suggestion
Using the options here allow you to update to future versions of `ddev-redis` without losing your configuration, and without the upgrade being blocked because you removed the `#ddev-generated` line from the upstream `docker-compose.redis.yaml`.
```

rfay (commented):
This looks like it has great promise, and it's a totally appropriate and general way to solve this problem. I like the `.env.service` technique because the configuration sticks around even if a new `ddev get` is done.

rfay:
Will return to this when it's ready for review.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
...
-    redis:
-      image: redis:7.0-alpine
-```
+    ```bash
+    ddev get ddev/ddev-redis --environment="REDIS_VERSION=7.0-alpine"
+    ddev restart
+    ```
+
+2. Remove the `#ddev-generated` line in `.ddev/docker-compose.redis.yaml` and edit the relevant line.
+3. Add a `.ddev/docker-compose.redis_extra.yaml` with the contents:
```
stasadev, line 118:
```suggestion
2. Add a `.ddev/docker-compose.redis_extra.yaml` with the contents:
```

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,25 @@ If you have add-ons that were installed before v1.22, update them with `ddev get
 
 ## Adding Custom Configuration to an Add-on
 
-Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis-7`](https://github.com/ddev/ddev-redis-7) the `image` is set to `image: redis:7.2-alpine`. If you wanted to change this to use `7.0-alpine` instead, you would have two choices:
+Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis`](https://github.com/ddev/ddev-redis) the `image` is set to `image: redis:${REDIS_VERSION:-6-bullseye}`. If you wanted to change this to use `7.0-alpine` instead, you would have three choices:
 
-1. Remove the `#ddev-generated` line in `docker-compose.redis-7.yaml` and edit the relevant line.
-2. Add a `.ddev/docker-compose.redis-7_extra.yaml` with the contents:
+1. With DDEV v1.23.4+, the override can be done using a special `.ddev/.env.*` file, where `*` is the name of the add-on, in this case `.ddev/.env.redis`:
```
stasadev, line 110:
```suggestion
1. With DDEV v1.23.5+, the override can be done using a special `.ddev/.env.*` file, where `*` is the name of the add-on, in this case `.ddev/.env.redis`:
```

rfay:
This has the same lima failure on port 33000, https://buildkite.com/ddev/macos-lima/builds/1370#01917ff1-750e-469e-b745-1b7955008fc0/5713-6094

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -48,7 +48,7 @@ echo
 ddev debug configyaml | grep -v web_environment
 
 header "existing project customizations"
-grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres 2>/dev/null
+grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres .ddev/.env .ddev/.env.!(example|*.example) 2>/dev/null
```
stasadev, line 51: It's good to show that users have `.env.*` in `ddev debug test`

cmd/ddev/cmd/testdata/TestCmdAddonGetEnvironmentFlag/busybox/docker-compose.busybox.yaml
```diff
@@ -0,0 +1,22 @@
...
+      com.ddev.site-name: ${DDEV_SITENAME}
+      com.ddev.approot: $DDEV_APPROOT
+    environment:
+      # This variable cannot be overridden with env_file, because it's already declared here
+      - THIS_VARIABLE_CANNOT_BE_CHANGED_FROM_ENV=true
+      # This variable can never be overridden from the .env.busybox.example file
+      - BUSYBOX_EXAMPLE_VARIABLE=${BUSYBOX_EXAMPLE_VARIABLE:-notset}
+      # DDEV_REDIS_VERSION from .env.redis can be passed here during the 'docker-compose config' step,
+      # when all .env.* files are passed to 'docker-compose config'
+      - CUSTOM_REDIS_VERSION=${DDEV_REDIS_VERSION:-notset}
```
stasadev, line 18: It's a chicken situation when I pass all the `.env.*` files to `docker-compose config` because I can't know what services I have until that config is complete, but at the time it is running, the variable `DDEV_REDIS_VERSION` is already expanded, and I can't do nothing about it.

This is the first thing I don't like about this PR.

pkg/ddevapp/compose_yaml.go
```diff
@@ -78,17 +87,54 @@ func (app *DdevApp) WriteDockerComposeYAML() error {
 		return err
 	}
 
+	// Run docker-compose config on the full rendered yaml file again after fixes.
+	// This action will override the existing file with the new one.
+	fullContentsFinal, _, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+		ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+		Action:       append(action, "config"),
+	})
```
stasadev, line 95: I have to run `docker-compose config` again because I add `env_file` to services and I need to convert it to an `environment` stanza.

This is the second thing I don't like about this PR.

pkg/ddevapp/ddevapp.go
```diff
@@ -1097,6 +1097,59 @@ func (app *DdevApp) ComposeFiles() ([]string, error) {
 	return orderedFiles, nil
 }
 
+// EnvFiles returns a list of env files for a project.
+// It has to put the .ddev/.env first
+// It has to put the .ddev/.env.* second
+// It has to put the .ddev/.env.*.override third
+// It has to put the .ddev/.env.override last
+// Env files ending with .example are ignored.
```
stasadev, line 1105: Maybe I didn't fully describe this behavior in the docs, probably it could be better.

pkg/ddevapp/envfile.go
```diff
@@ -35,7 +35,11 @@ func WriteProjectEnvFile(envFilePath string, envMap map[string]string, envText s
 			envText = exp.ReplaceAllString(envText, fmt.Sprintf(`$1$2="%s"`, v))
 		} else {
 			envText = strings.TrimSuffix(envText, "\n")
-			envText = fmt.Sprintf("%s\n%s=%s\n", envText, k, v)
+			if envText != "" {
+				envText = fmt.Sprintf("%s\n%s=\"%s\"\n", envText, k, v)
+			} else {
+				envText = fmt.Sprintf("%s=\"%s\"\n", k, v)
+			}
```
stasadev, line 42: This is required to remove empty lines from `.env` files.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,31 @@ If you have add-ons that were installed before v1.22, update them with `ddev add
...
+    ddev add-on get ddev/ddev-redis --environment="DDEV_REDIS_VERSION=7.0-alpine"
+    ddev restart
+    ```
+   
+    You can override `.ddev/.env.*` with `.ddev/.env.*.override` file in case you need more customization.
+
+    !!!warning "Environment variables cannot be overridden in some cases"
+        If your `docker-compose.*.yaml` file already has some variables in the `environment` or `env_file` stanza, you cannot override them with `.ddev/.env.*` or `.ddev/.env.*.override` files.
+        
+        If this is the case, move these variables to the appropriate `.ddev/.env.*` file to be able to override them with a `.ddev/.env.*.override` file.
```
stasadev, line 122: This is why I removed this section from the app template:

```yaml
    {{ if .EnvFile }}
    env_file: '{{ .EnvFile }}'
    {{ end }}
```

Otherwise, I couldn't override any existing variables from `.ddev/.env` with `.ddev/.env.override`.

stasadev (commented):
I left a few comments explaining what I did and what I didn't like.
I covered all the behavior in the test.

rfay:
What happens when you do `ddev add-on get https://github.com/ddev/ddev-redis/tarball/20240717_stasadev_optional_version --environment="DDEV_REDIS_VERSION=7"`

and then later do a `ddev add-on get https://github.com/ddev/ddev-redis/tarball/20240717_stasadev_optional_version` (without the option)? I guess it keeps the special configuration on into the future?

rfay (changes requested):
This is a creative and careful approach to the problem at hand. I'm not sure how much you like it, and I'm not sure how much I like it. Let's talk this through before committing to it. 

I should mention that the `ddev pull --environment` approach that was probably a model here isn't all that great either, even though it does solve a problem.

Anyway, let's do a voice call and talk about it.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,31 @@ If you have add-ons that were installed before v1.22, update them with `ddev add
 
 ## Adding Custom Configuration to an Add-on
 
-Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis-7`](https://github.com/ddev/ddev-redis-7) the `image` is set to `image: redis:7.2-alpine`. If you wanted to change this to use `7.0-alpine` instead, you would have two choices:
+Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis`](https://github.com/ddev/ddev-redis) the `image` is set to `image: redis:${DDEV_REDIS_VERSION:-6-bullseye}`. If you wanted to change this to use `7.0-bookworm` instead, you would have two choices:
 
-1. Remove the `#ddev-generated` line in `docker-compose.redis-7.yaml` and edit the relevant line.
-2. Add a `.ddev/docker-compose.redis-7_extra.yaml` with the contents:
+1. With DDEV v1.23.5+, the override can be done using a special `.ddev/.env.*` file, where `*` is the name of the add-on (and its service), in this case `.ddev/.env.redis`:
```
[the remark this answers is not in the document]
stasadev: `something` in `.env.something` is only relevant if env variables are to be passed inside a service.

Otherwise the filename is irrelevant, it will be expanded by `docker-compose config` anyway.
I use the add-on name to name this `.env.*` file somehow.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,31 @@ If you have add-ons that were installed before v1.22, update them with `ddev add
...
-      image: redis:7.0-alpine
-```
+    ```bash
+    ddev add-on get ddev/ddev-redis --environment="DDEV_REDIS_VERSION=7.0-alpine"
+    ddev restart
+    ```
+
+    You can override `.ddev/.env.*` with `.ddev/.env.*.override` file in case you need more customization.
+
+    !!!warning "Environment variables cannot be overridden in some cases"
```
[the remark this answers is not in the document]
stasadev: I should rethink how it works. Because I don't like it.

pkg/ddevapp/compose_yaml.go
stasadev: I have an idea how to get rid of this second `config` call.

I should manually adjust `environment` stanza, instead of using `env_file` stanza, and it will also fix the problem with:

```
!!!warning "Environment variables cannot be overridden in some cases"
```

pkg/ddevapp/compose_yaml.go
```diff
@@ -128,6 +179,25 @@ func fixupComposeYaml(yamlStr string, app *DdevApp) (map[string]interface{}, err
 			"ddev_default": nil,
 			"default":      nil,
 		}
+
+		for _, envFile := range envFiles {
+			filename := filepath.Base(envFile)
+			// Files like .env and .env.override should go in the web container
```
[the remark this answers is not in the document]
stasadev: Maybe I'm overthinking here and we don't need this "override" at all?

This is not supported by `docker-compose config` originally, it was just my idea that I was trying to implement.

pkg/ddevapp/ddevapp.go
```diff
@@ -1055,7 +1055,7 @@ func (app *DdevApp) ImportFiles(uploadDir, importPath, extractPath string) error
 
 // ComposeFiles returns a list of compose files for a project.
 // It has to put the .ddev/docker-compose.*.y*ml first
-// It has to put the docker-compose.override.y*l last
+// It has to put the .ddev/docker-compose.override.y*ml last
```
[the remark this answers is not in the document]
stasadev: Yes, I got the idea for `.env.*.override` from here, but it can be very confusing.

stasadev (commented):
@rfay, thanks for the feedback, it made me rethink some logic, I'll try to simplify it and request a second round of review.

stasadev:
I had a discussion about it with @rfay and I'm going to make another round:

- remove override technique for `.env.*` files, because it's too complex

- try to implement Cobra capabilities for unknows flags to change this:

    ```
    ddev add-on get ddev/ddev-redis --environment="DDEV_REDIS_VERSION=7"
    ```
    to
    ```
    ddev add-on get ddev/ddev-redis --ddev-redis-version=7
    ```

    `--ddev-redis-version=7` should be converted to `DDEV_REDIS_VERSION=7` automatically, and then stored in `.env.redis`

- remove the second `docker-compose config` by changing/adding env variables in yaml on the DDEV side

GuySartorelli:
Feel free to tag me again when you've finished the next round of changes.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,31 @@ If you have add-ons that were installed before v1.22, update them with `ddev add
...
+2. Add a `.ddev/docker-compose.redis_extra.yaml` with the contents:
+
+    ```yaml
+    services:
+      redis:
+        image: redis:7.0-bookworm
+    ```
 
-Using the second option (`docker-compose.redis-7_extra.yaml`) allows you to update to future versions of `ddev-redis-7` without losing your configuration, and without the upgrade being blocked because you removed the `#ddev-generated` line from the upstream `docker-compose.redis-7.yaml`.
+Using the options here allow you to update to future versions of `ddev-redis` without losing your configuration, and without the upgrade being blocked because you removed the `#ddev-generated` line from the upstream `docker-compose.redis.yaml`.
```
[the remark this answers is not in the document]
stasadev: I added a note below.

docs/content/users/extend/customization-extendibility.md
```diff
@@ -13,20 +13,27 @@ You can set custom environment variables in several places:
 
 1. An optional, project-level `.ddev/.env` file provides environment variables to all DDEV containers, including any additional services or add-ons. It can look something like this:
```
stasadev, line 14: This statement is not true for DDEV v1.23.4, because `.ddev/.env` adds variables inside containers only for the `web` container.

But it will become true after this PR goes in.

stasadev (commented):
@rfay and @GuySartorelli, I think I've achieved what I wanted to do here, it's ready for review.

Please see the updated explanations and instructions in the first post.

Makefile
```diff
@@ -9,7 +9,7 @@ BUILD_BASE_DIR ?= $(PWD)
 GOTMP=.gotmp
 SHELL = /bin/bash
 PWD = $(shell pwd)
-GOFILES = $(shell find $(SRC_DIRS) -type f)
+GOFILES = $(shell find $(SRC_DIRS) -type f ! -path "*/testdata/*")
```
stasadev, line 12: There was a problem on `make build`:

https://github.com/ddev/ddev/actions/runs/10881287727/job/30189913793?pr=6406#step:16:29
```
tangle cmd/ddev/cmd/testdata/TestCmdAddonGetWithEnvironmentVariablesFromUnknownFlags/env_files/.env.web
/bin/bash: line 1: tangle: command not found
make: *** [<builtin>: cmd/ddev/cmd/testdata/TestCmdAddonGetWithEnvironmentVariablesFromUnknownFlags/env_files/.env.p] Error 127
```

The reason for this behavior:

https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

> TeX and Web
> n.dvi is made from n.tex with the recipe ‘$(TEX)’. n.tex is made from n.web with ‘$(WEAVE)’, or from n.w (and from n.ch if it exists or can be made) with ‘$(CWEAVE)’. n.p is made from n.web with ‘$(TANGLE)’ and n.c is made from n.w (and from n.ch if it exists or can be made) with ‘$(CTANGLE)’.

Excluding all `/testdata/` files from the `Makefile` helped.

Thanks to @rfay for showing me the rules for `make`.

stasadev:
Rebased to resolve conflicts and squashed everything into one commit.

stasadev:
> What about when you want to change a version (or the entire image, eg. redis -> valkey) for testing Do you re-install the addon? That seems pointless at best, and more likely just dangerous since a re-install's various actions are not unlikely to wipe out any user customizations.
> a separate `ddev setdotenv {optional service} --var1 value1 --var2 command`

I like this idea, creating a new command will not break the existing flag validation, as it does in this PR for `ddev add-on get`.
I'm going to implement it.

> `ddev redis version {version number}`

Yes, this is easy to do with bash, and when we have `ddev setdotenv`, use it like this:

```bash
$ cat .ddev/commands/host/redis

if [[ "${1:-}" == "version" && "${2:-}" != "" ]]; then
    ddev setdotenv redis --redis-tag $2
    ddev restart
    exit 0
fi
```

But in the case of `ddev/ddev-redis-7` it will overlap with `.ddev/commands/web/redis`, so we should think about how to name it or rename the command for `redis-cli`.

> It would also be useful to expose yq and jq functionality outside of the containers

We use this mini-image in our scripts, and it can also be used in add-ons (example https://ddev.com/blog/anatomy-advanced-ddev-addon/):

```
docker run -i --rm ddev/ddev-utilities jq -V
```

There is no `yq` inside yet.

rfay (commented):
How can one discover the possible flags for an add-on? Just via the add-on's README?

rfay:
Working my way through. What great work on making .env.* work properly, fantastic. And the flag parsing seems to work perfectly, although I haven't studied the code carefully yet. All the manual testing worked out perfectly.

docs/content/users/extend/additional-services.md
```diff
@@ -105,18 +105,32 @@ If you have add-ons that were installed before v1.22, update them with `ddev add
 
 ## Adding Custom Configuration to an Add-on
 
-Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis-7`](https://github.com/ddev/ddev-redis-7) the `image` is set to `image: redis:7.2-alpine`. If you wanted to change this to use `7.0-alpine` instead, you would have two choices:
+Sometimes it's necessary to add custom configuration to an add-on. For example, in [`ddev-redis`](https://github.com/ddev/ddev-redis) the `image` is set to `image: redis:${REDIS_TAG:-6-bullseye}` (it is important to have the default value for `$REDIS_TAG`). If you wanted to change this to use `7-bookworm` instead, you would have two choices:
 
-1. Remove the `#ddev-generated` line in `docker-compose.redis-7.yaml` and edit the relevant line.
-2. Add a `.ddev/docker-compose.redis-7_extra.yaml` with the contents:
+1. With DDEV v1.23.5+, the override can be done using a special `.ddev/.env.*` file, where `*` is the name of the add-on, in this case `.ddev/.env.redis` (file is created automatically):
```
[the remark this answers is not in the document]
rfay: I think we have a little bit of trouble here. 

The name of an add-on *may* be the same as the name of the *service*. But it may not be. There's nothing that forces that. In fact, in ddev-ioncube, the name is nonstandard... and there is also not a docker-compose.*.yaml, so it's impossible to do a .env anyway. 

But if there *were* a `docker-compose.*.yaml`, it might contain more than one service. 

I think what we probably have here at this point is just a naming problem. I think you're completely handling all *services* by looking and the fully constructed compose yaml. But that may not be the exact name of the add-on. And if a `docker-compose.*.yaml` had more than one service in it (I think some do) then this would mostly work OK?

rfay (approved):
This seems amazingly elegant and solid to me, great work. I read through it and manually tested and also step-debugged through some various cases. 

I think there's probably is a little ambiguity left between an "add-on", a "docker-compose.*.yaml" file, and a "service". 

I think this is completely addressing the concept of "service" (the `service` tag in a docker-compose file), but it just happens to be done in the context of an `add-on`. ) Since a `docker-compose.*.yaml` (perhaps provided by an add-on) can have multiple services, it will probably work fine as is. But it's the service, in a `docker-compose.*.yaml`, that is being managed and helped with the `.env` file. And of course, an add-on may not have a `docker-compose.*.yaml` file.

Congrats! Perhaps just a little verbiage cleanup and thinking and pull away?

cmd/ddev/cmd/addon-get.go
```diff
@@ -343,6 +349,52 @@ func createManifestFile(app *ddevapp.DdevApp, addonName string, repository strin
...
+		// Skip short flags
+		if strings.HasPrefix(flag, "--") {
+			// Convert flags to env variables, e.g. "--redis-tag" -> "REDIS_TAG"
+			envName := strings.ToUpper(strings.ReplaceAll(strings.TrimLeft(flag, "-"), "-", "_"))
+			// Remove double quotes because we wrap the value in double quotes.
+			envMap[envName] = strings.ReplaceAll(value, `"`, "")
+		}
+
+	}
+	if len(envMap) > 0 {
```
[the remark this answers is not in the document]
stasadev: I think this is expected behavior, given that I don't use `#ddev-generated` here.

cmd/ddev/cmd/addon-get.go
```diff
@@ -343,6 +349,52 @@ func createManifestFile(app *ddevapp.DdevApp, addonName string, repository strin
...
+		err = ddevapp.WriteProjectEnvFile(envFile, envMap, envText)
+		if err != nil {
+			util.Failed("Error writing .env file: %v", err)
+		}
+	}
+	envMap, _, err = ddevapp.ReadProjectEnvFile(envFile)
```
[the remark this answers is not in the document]
stasadev: I changed this function several times, and probably used the second call for something else before, I should remove it.

cmd/ddev/cmd/addon-get.go
```diff
@@ -343,6 +349,52 @@ func createManifestFile(app *ddevapp.DdevApp, addonName string, repository strin
...
+	if err != nil && !os.IsNotExist(err) {
+		util.Failed("Unable to read %s file: %v", envFile, err)
+	}
+	if len(envMap) > 0 {
+		if verbose {
+			util.Warning("Using env file %s", envFile)
+		}
+		injectedEnv = "export"
```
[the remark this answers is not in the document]
stasadev: I have a test with two env variables, so it should be good.

docs/content/users/extend/additional-services.md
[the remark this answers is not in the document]
stasadev: The add-on name matters only in the `ddev add-on get --something=value` command, and as you mention, this is a problem.

And luckily we can make it better and obvious using a new command `ddev dotenv .env.ioncube --flags`, that I described in:
- https://github.com/ddev/ddev/pull/6406#issuecomment-2363869271

stasadev (commented):
> Congrats! Perhaps just a little verbiage cleanup and thinking and pull away?

I want to extract logic with flags to the new `ddev dotenv` command.

The reasons for this are simple:

1. When we parse unknown flags, we lose flag validation, and in case of `ddev add-on get` it can be confusing.
2. I see more potential in having an explicit command to manipulate/update `.env` files, which makes it easier to update them without using `sed`. It goes beyond add-ons functionality, which makes it more useful.

There's not much work to do, I'll just move the logic a little bit.

stasadev (commented):
> Congrats! Perhaps just a little verbiage cleanup and thinking and pull away?

I want to extract logic with flags to the new `ddev dotenv` command.

The reasons for this are simple:

1. When we parse unknown flags, we lose flag validation, and in case of `ddev add-on get` it can be confusing.
2. I see more potential in having an explicit command to manipulate/update `.env` files, which makes it easier to update them without using `sed`. It goes beyond add-ons functionality, which makes it more useful.

There's not much work to do, I'll just move the logic a little bit.

stasadev:
The `ddev add-on get addon --some=value` command was quite unclear about the location of the generated `.env` file.

Now that we are using the syntax with `ddev dotenv`, it should be much easier to understand the logic, as it becomes clear where the changes are made.

Ready for another look.

docs/content/users/extend/additional-services.md
```diff
@@ -134,7 +150,7 @@ Anyone can create an add-on for `ddev add-on`. See [this screencast](https://www
 The `install.yaml` is a simple YAML file with a few main sections:
 
 * `name`: The add-on name. This will be used in the various `ddev add-on` commands. If the repository name is `ddev-solr`, for example, a good `name` will be `solr`.
-* `pre_install_actions`: an array of Bash statements or scripts to be executed before `project_files` are installed. The actions are executed in the context of the target project’s root directory.
+* `pre_install_actions`: an array of Bash statements or scripts to be executed before `project_files` are installed. The actions are executed in the context of the target project’s root directory. With DDEV v1.23.5+, variables from the `.ddev/.env.name` file, where `name` is the name of the add-on, are available in this context.
```
rfay, line 153:
```suggestion
* `pre_install_actions`: an array of Bash statements or scripts to be executed before `project_files` are installed. The actions are executed in the context of the target project’s root directory. With DDEV v1.23.5+, variables from the `.ddev/.env.name` file, where `name` is the name of the service in the add-on's `docker-compose.*.yaml`, are available in this context.
```

docs/content/users/extend/additional-services.md
```diff
@@ -144,7 +160,7 @@ The `install.yaml` is a simple YAML file with a few main sections:
     ```
 
 * `dependencies`: an array of add-ons that this add-on depends on.
-* `post_install_actions`: an array of Bash statements or scripts to be executed after `project_files` and `global_files` are installed. The actions are executed in the context of the target project’s `.ddev` directory.
+* `post_install_actions`: an array of Bash statements or scripts to be executed after `project_files` and `global_files` are installed. The actions are executed in the context of the target project’s `.ddev` directory. With DDEV v1.23.5+, variables from the `.ddev/.env.name` file, where `name` is the name of the add-on, are available in this context.
```
rfay, line 163: So here, the .env.name is for the add-on name, true? But elsewhere, service name?

rfay (approved):
This seems to me to be quite a masterwork, and against all odds it may have gotten simpler with age. Thanks for all the work on this!

The only thing I seem to be confused about still is where the `.env.<add-on-name>` is used. I think it's there for the *service* that matches the name in a `docker-compose.*.yaml`. But you have used it in `ddev-php-patch-build` without needing a `docker-compose.*.yaml`. But I think that one is completely dependent on the behavior of the post-install action, so it's actually just reaching in and grabbing the variable, not relying on the behavior of DDEV.

docs/content/users/extend/additional-services.md
stasadev: Yes.

cmd/ddev/cmd/addon-get_test.go
```diff
@@ -85,6 +85,10 @@ func TestCmdAddonComplex(t *testing.T) {
 
 // TestCmdAddonDependencies tests the dependency behavior is correct
 func TestCmdAddonDependencies(t *testing.T) {
+	if os.Getenv("DDEV_RUN_GET_TESTS") != "true" {
```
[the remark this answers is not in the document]
stasadev: Okay, thank you, I'll review the tests.

stasadev (commented):
> The only thing I seem to be confused about still is where the `.env.<add-on-name>` is used. I think it's there for the *service* that matches the name in a `docker-compose.*.yaml`. But you have used it in `ddev-php-patch-build` without needing a `docker-compose.*.yaml`. But I think that one is completely dependent on the behavior of the post-install action, so it's actually just reaching in and grabbing the variable, not relying on the behavior of DDEV.

There are all the situations where `.ddev/.env.something` can be used:

1. If you name it `.ddev/.env.does-not-matter`, the variables from this file will always be expanded in `docker-compose.*.yaml` files. This works because I pass all `--env-file` to `docker-compose config`. And I can only control the alphabetical order that the files are passed here. If there is two files with the same variable, it will use the value of the last one.
2. If you name it `.ddev/.env.addon-name`, it will be available for `addon-name` add-on `pre_install_actions` and `post_install_actions`.
3. If you name it `.ddev/.env.service-name`, the variables from it will be available inside the `service-name` container.

If we look at the `ddev/ddev-redis` add-on, this is a combined case of all three situations.

If we look at the `rfay/ddev-php-build` add-on, I named the file `.ddev/.env.php-patch-build` just to make it available for `post_install_actions`, otherwise the filename there won't be important.

docs/content/users/extend/additional-services.md
stasadev: And here `.env.name` is always the name of the add-on. I can't use the service name here because add-ons may not have services, and I can't get the service name without running `docker-compose config`, which creates a catch-22.

rfay:
Congrats on a great effort!

cmd/ddev/cmd/addon-get_test.go
```diff
@@ -18,10 +18,6 @@ import (
 
 // TestCmdAddonComplex tests advanced usages
 func TestCmdAddonComplex(t *testing.T) {
-	if os.Getenv("DDEV_RUN_GET_TESTS") != "true" {
-		t.Skip("Skipping because DDEV_RUN_GET_TESTS is not set")
-	}
-
```
stasadev, line 24: This change broke Windows test, see:

- #6585

PR #6407, opened 2024-07-17
test: make sure that volumes in buildkite are deleted at test start

rfay:
## The Issue

There were [quite](https://buildkite.com/ddev/macos-colima/builds/396#0190acea-35ee-4f1b-817b-e36e04d286e8) [a](https://buildkite.com/ddev/macos-colima-vz/builds/1292#0190b7e9-91cd-4a55-a719-17b5731a413a) [lot](https://buildkite.com/ddev/macos-colima-vz/builds/1308#0190c165-cc86-4048-8713-31d281a48c3c) of suspicious TestSSHAuth failures on colima-vz on tb-macos-arm64-6, so I finally thought I'd take a look. 

I found a lot of volumes, especially the SSH volumes, stuck there. 

It looks to me like we haven't been successful deleting volumes at test startup.

* Make sure to delete volumes after all containers are stopped
* Double-delete
* Stop ignoring the output.

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -162,9 +162,6 @@ fi
 
 set -x
 
-# We don't want any docker volumes to be existing and changing behavior
-docker volume prune -a -f >/dev/null 2>&1 || true
-
 # Run any testbot maintenance that may need to be done
 echo "--- running testbot_maintenance.sh"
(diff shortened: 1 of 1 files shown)


stasadev (approved):
I think this is correct, I tested it too, all the volumes I had locally were deleted.
(I wanted to do fresh db imports anyway.)

PR #6414, opened 2024-07-21
feat: adds support to ephemeral ports for 80/443, fixes #5326

### LINKED ISSUE #5326: Feature: Use alternate ephemeral port for router if ports 80/443 are occupied
It seems Lando chooses an alternate port when a given port is not available. DDEV checks for port availability, but does not switch. It's a nice feature. It would make first-timers have a little better experience.

agviu:
## The Issue

- #5326 

Ports 80 and 443 and commonly occupied. DDEV should be able to successfully start by creating ephemeral ports as an alternative.

New functionality is added that checks as soon as possible if the ports HTTP and HTTPS, used by the router, are occupied.
If they are occupied, a ephemeral port is assigned and used for configuring the router.

The router registers if it's using ephemeral ports in an environment variable. This information is readed by other projects that start and is useful to detect if the router itself is occuping the http and https ports, or not. 

# TODO

- [x] `ddev describe` does not show to correct URLs in the bottom line "All URLs"
- [x] `router_http_port` and `router_https_port` are not updated in the project. Should they be? Or should this just be temporary? We do see the expected:
    > HTTP port 28080 is busy, using 50080 instead.
    > HTTPS port 28443 is busy, using 50443 instead.
- [ ] `app.Stop()` pushes [removes site config from router](https://github.com/ddev/ddev/blob/c91869a5a840f55eeb020d71d09e80158542dcfc/pkg/ddevapp/ddevapp.go#L2568-L2575). This is good, and may be enough. However, the router itself still is configured with an entrypoint (no longer used, perhaps) in ~/.ddev/traefik/.static_config.yaml. And in this case the router is still running bound to the ports that are perhaps no longer in use. To demonstrate: `docker exec -it ddev-router cat .static_config.yaml`. This is not a bug related to the work in this PR, may not be a bug at all but it may be related to some of the weirdness we see in other tests after the ephemeral tests run.
- [x] The logic has to handle *all* ports handled by the router, not just http/https. 
- [x] There's no real reason to distinguish between HTTP and HTTPS I don't think. We just need the ports.
- [x] Handle the mailpit ports at least
- [x] Handle HTTP_EXPOSE and HTTPS_EXPOSE related to `web_extra_exposed_ports`, a very common scenario
- [x] Consider if there's a way to handle conflicts on 3rd party services, which would require overriding their HTTP_EXPOSE and HTTPS_EXPOSE.
- [x] I removed the second site start from TestUseEphemeralPort, but it's a good thing to have that second site, as I've discovered while breaking things. It should go back in there.
- [x] Don't use "80", "443", "8025" and "8026" as literals. Make those global and configurable. We already have constants for these in values.go, DdevDefaultRouterHTTPPort and DdevDefaultRouterHTTPSPort etc. But we can make them variables, and override them with a DDEV_TEST environment variable, so that multiple non-conflicting tests might possibly be runnable at the same time. (followup in https://github.com/ddev/ddev/pull/6470)
- [x] Comment out or make DDEV_VERBOSE all the massive output showing HTTP_EXPOSE etc.

First, make sure any DDEV container is off by running:
`ddev poweroff`

Then, occupy the 80 and 443 ports. For example, by running:
`docker run -d -p 80:80 -p 443:443 nginx`

Then, go to a DDEV project that is configured to use the same ports. This is managed in the config variables:
router_http_port="80"
router_https_port="443"

Then run
`ddev start`
You will see a message, very early on, indicating that the ports are occupied and an ephemeral port wil be used instead.
The start process should finish normally, and DDEV already indicates where to find the project by displaying the URL.
You can also run
`ddev describe`
to see information about the URL.

Then, repeat the same operation with a different project that is also configured to use the same ports.
DDEV should detect the ephemeral ports and assign them to the new project.

## Automated Testing Overview

There are 2 tests created:
- One for checking that DDEV is able to find an available port in a given range.
- Another for checking that the app get's the ephemeral port when it has to, and that a second app will also get the same ports.

## Release/Deployment Notes

This feature should not have any negative effect in other parts of the code.


rfay:
Wow, so timely! Ping me if you push additional commits so I can start the github tests.

rfay:
The easy stuff is cleaning up the linting, `make staticrequired` should sort out the things in https://github.com/ddev/ddev/actions/runs/10030444886/job/27721100040?pr=6414

rfay:
It looks like maybe your test didn't clean up after itself after changing router ports, and tests after that fail? If you don't have access to the buildkite builds or can't make sense of the github builds, let me know.

rfay:
I went ahead and rebased this, so please make sure to update your local remote, there are many ways to do it, but `git pull --rebase` should do the job.

rfay:
A note: It will be important to validate the behavior of `ddev debug test` with this.

agviu:
Hi @rfay , 
I ran 'make staticrequired' and updated the code as requested.
I do not have access to the buildkite, so I'm not sure what's happening. The apps created during the test have also a Cleanup function attached, which among other things restore the original global HTTP/HTTPS ports.

About running 'ddev debug test', I also ran it and it went fine.

rfay:
Ran your tests. If you PM me your email I'll add you to buildkite. (Sometimes the buildkite tests can be flaky, especially Windows and colima)

rfay:
Sent you invitation to buildkite.

rfay:
It looks like TestUseEphemeralPort is failing oddly (linux) with a permissions error. 
https://github.com/ddev/ddev/actions/runs/10066810230/job/27829479804#step:15:11777

If Linux is not your home territory, you might want to try running that test manually on Gitpod.

rfay:
Oh, then I see 
https://buildkite.com/ddev/macos-colima-vz/builds/1345#0190e18c-b285-478e-9309-1c7e758ff37b

```
invalid containerPort: 72080
  | config_test.go:1258:
  | Error Trace:	/Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-6/ddev/macos-colima-vz/pkg/ddevapp/config_test.go:1258
  | Error:      	Received unexpected error:
  | composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-testpkgdrupal10 docker-compose -f /Users/testbot/.ddev/.router-compose.yaml config', action='[config]', err='exit status 15', stdout='', stderr='invalid containerPort: 72080'
  | Test:       	TestCustomBuildDockerfiles
 ```

It looks like something is allowing ports greater than 65535? Can't go higher than that. 

<br class="Apple-interchange-newline">

**Comment by rfay** [comment]
You can comment here anytime you want to talk, and if you want to do a zoom/screenshare make an appointment at https://cal.com/randyfay/30min - thanks for working on this!

**Comment by rfay** [comment]
Looks like great progress. I was able to reproduce the test failure locally with 
```
make testpkg TESTARGS="-run 'Test(.*Ephemeral|Traefik)'"
```

or with this Goland configuration

image

**Comment by rfay** [comment]
It looks to me like the problem is that traefik has been left configured with nonstandard ports (50080/50443). So there's a cleanup problem in the Test*Ephemeral tests.

```
entryPoints:
    http-8025:
        address: :8025
    http-8026:
        address: :8026
    http-50080:
        address: :50080
    http-50443:
        address: :50443
    traefik:
        address: :10999
```

**Comment by stasadev** [comment]
@agviu, one way to resolve these conflicts is to use a different global config directory for your tests:

https://github.com/ddev/ddev/blob/c2aca52a18687e678086dd232573cf51914dba56/pkg/dockerutil/docker_compose_version_test.go#L28-L32

It is already used in several DDEV tests. The main idea is to `CopyGlobalDdevDir` before testing something, and then `ResetGlobalDdevDir` during cleanup.

**Comment by rfay** [comment]
It looks like TestRouterPortsCheck is the next frontier :)

**Comment by rfay** [comment]
One step at a time!

I see this in https://github.com/ddev/ddev/actions/runs/10166920841/job/28118414199?pr=6414 - I expect it's caused by the fact that ports 1024 and below are not allowed bind permissions without root privileges, which is a key reason why people use things like 8080 and 8443. But it's not clear to me why this is happening to you here. But a fairer test will always involve ports 1024+


```
=== RUN   TestFindEphemeralPort
    router_test.go:284: 
        	Error Trace:	/home/runner/work/ddev/ddev/pkg/ddevapp/router_test.go:284
        	            	listen tcp 127.0.0.1:439: bind: permission denied
        	Test:       	TestFindEphemeralPort
```

**Comment by agviu** [comment]
I managed to reproduce the error locally by running
`go test -v -timeout 3600s -run "TestUseEphemeralPort|TestTraefikStaticConfig" ./pkg/ddevapp`
So that error was found and fixed.

**Comment by agviu** [comment]
> I see TestRouterPortsCheck failed in https://buildkite.com/ddev/macos-colima-vz/builds/1383#019104ab-2a51-4498-b099-f83ac863ab73 - It's suspicious of course because you're working in that area, but that is a bit of a flaky test and fails sometimes. I'm going to restart it.

yes, it suspicious. It would make sense that this PR affect that test. I did not manage to reproduce it locally, though...

**Comment by rfay** [comment]
Tests are WORKING!!!! Now to get to actual review :) I'm going to rebase this and move to to "ready to review". Don't forget to reset your branch with `git fetch <your_remote> && git reset --hard <your_remote>/5326...`

**Comment by agviu** [comment]
@rfay @stasadev I also updated the documentation with a mention of this feature. Please feel free to rewrite as you consider.

**Comment by tyler36** [comment]
@agviu
I'm looking forward to this. This looks really exciting.

**Comment by rfay** [comment]
Looks like a little remaining test trouble, https://buildkite.com/ddev/macos-lima/builds/1234#01910a23-7cea-4891-abc7-19061fde95fa, not sure why it doesn't affect every test.

```

ddevapp_test.go:4381:
--
  | Error Trace:	/Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-5/ddev/macos-lima/pkg/ddevapp/ddevapp_test.go:4381
  | Error:      	Not equal:
  | expected: "80"
  | actual  : "50080"
  |  
  | Diff:
  | --- Expected
  | +++ Actual
  | @@ -1 +1 @@
  | -80
  | +50080
  | Test:       	TestEnvironmentVariables
  | ddevapp_test.go:4381:
  | Error Trace:	/Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-5/ddev/macos-lima/pkg/ddevapp/ddevapp_test.go:4381
  | Error:      	Not equal:
  | expected: "443"
  | actual  : "50443"
  |  
  | Diff:
  | --- Expected
  | +++ Actual
  | @@ -1 +1 @@
  | -443
  | +50443
```

File: docs/content/users/install/ddev-upgrade.md [hunk 1]
```diff
@@ -13,6 +13,7 @@ Installing and upgrading DDEV are nearly the same thing, because you're upgradin
 
     ```bash
     # Upgrade DDEV to the latest version
+    brew tap ddev/ddev
```
rfay, line 16:
```suggestion
```

This is not necessary, since `brew install ddev/ddev/ddev` does the tap for you.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
...
+		DBTarURL:                      "https://github.com/ddev/ddev_test_tarballs/releases/download/v1.1/d8_umami.sql.tar.gz",
+		DBZipURL:                      "https://github.com/ddev/ddev_test_tarballs/releases/download/v1.1/d8_umami.sql.zip",
+		FullSiteTarballURL:            "",
+		Dir:                           testcommon.CreateTmpDir(t.Name() + "_2"),
+		Type:                          nodeps.AppTypeDrupal,
+		Docroot:                       "",
+		Safe200URIWithExpectation:     testcommon.URIWithExpect{URI: "/README.txt", Expect: "Drupal is an open source content management platform"},
+		DynamicURI:                    testcommon.URIWithExpect{URI: "/node/2", Expect: "Vegan chocolate and nut brownies"},
+		FilesImageURI:                 "/sites/default/files/vegan-chocolate-nut-brownies.jpg",
+	}
```
rfay, line 339: There isn't actually a need for all this. We don't need (and aren't using all this TestSite stuff. All we need is a trivial tmp directory with a `ddev config` in it in order to test this.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
...
+	app, err := ddevapp.NewApp(site1.Dir, false)
+	require.NoError(t, err)
+	app2, err := ddevapp.NewApp(site2.Dir, false)
+	require.NoError(t, err)
+
+	// Configure both apps to use the same target ports. Keep original configured ports for undoing the configuration later.
+	appHTTPPort, appHTTPSPort := app.RouterHTTPPort, app.RouterHTTPSPort
+	app.RouterHTTPPort, app.RouterHTTPSPort = targetHTTPPort, targetHTTPSPort
+	app2HTTPPort, app2HTTPSPort := app.RouterHTTPPort, app.RouterHTTPSPort
+	app2.RouterHTTPPort, app2.RouterHTTPSPort = targetHTTPPort, targetHTTPSPort
```
rfay, line 350: "short variable declaration", I had never seen this before, initializing 2 things at once. I'm not sure it's useful here though :)  I think an array or map assignment might be simpler. Not sure.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
...
+		// Undo the configuration of the ports.
+		app.RouterHTTPPort, app.RouterHTTPSPort = appHTTPPort, appHTTPSPort
+		app2.RouterHTTPPort, app2.RouterHTTPSPort = app2HTTPPort, app2HTTPSPort
+
+		err = app.WriteConfig()
+		assert.NoError(err)
+		err = app2.WriteConfig()
+		assert.NoError(err)
+		app.RemoveGlobalProjectInfo()
+		app2.RemoveGlobalProjectInfo()
```
rfay, line 367: We don't have to do any of these things if we're using a unique tmpdir. They will all be long gone.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
...
+	l0, err := net.Listen("tcp", "127.0.0.1:"+targetHTTPPort)
+	require.NoError(t, err)
+	defer l0.Close()
+	l1, err := net.Listen("tcp", "127.0.0.1:"+targetHTTPSPort)
+	require.NoError(t, err)
+	defer l1.Close()
+
+	// Find out which ephemeral ports the apps will use.
+	ephemeralHTTPPort, ok := ddevapp.FindAvailableRouterPort(ddevapp.MinEphemeralHTTPPort, ddevapp.MaxEphemeralHTTPPort)
+	assert.Exactly(true, ok)
```
rfay, line 390:
```suggestion
	require.True(t, ok)
```

Wouldn't that be easier?

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
...
+		if router != nil && err == nil && router.State == "running" {
+			err = dockerutil.RemoveContainer(nodeps.RouterContainer)
+			assert.NoError(err)
+		}
+	})
+
+	// Occupy target router ports:
+	l0, err := net.Listen("tcp", "127.0.0.1:"+targetHTTPPort)
+	require.NoError(t, err)
+	defer l0.Close()
```
rfay, line 383: I think these will be better in t.Cleanup(), as there we can check and assert on their success. Since we're having some odd results in various places, I guess we'd better know whether ports are properly being cleaned up.

rfay (changes requested):
Hopefully with your permission, I think it will be better for me to try to clean up this test than to feed you changes piecemeal. I hope I'm not overstepping, but am hoping you'll be willing to continue your ambitious contributions, and you'll be too frustrated to continue if we keep this going for months.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +269,140 @@ func TestDisableHTTP2(t *testing.T) {
 	assert.Equal("HTTP/1.1 200 OK\r\n", out)
 
 }
+
+// Test the function FindEphemeralPort
+func TestFindEphemeralPort(t *testing.T) {
```
rfay, line 274: I'm not sure, but I think there's a basic misunderstanding here. Both apps should be able to use the same target ports (normally 80 and 443) because ddev-router is routing them correctly. This is the way DDEV normally works.

rfay:
Rebased and simplified TestUseEphemeralPort

pkg/ddevapp/router.go
```diff
@@ -16,12 +17,23 @@ import (
...
 )
 
 // RouterProjectName is the "machine name" of the router docker-compose
-const RouterProjectName = "ddev-router"
+const (
+	RouterProjectName     = "ddev-router"
+	MinEphemeralHTTPPort  = 50080
+	MaxEphemeralHTTPPort  = 50442
+	MinEphemeralHTTPSPort = 50443
+	MaxEphemeralHTTPSPort = 50500
```
rfay, line 31: There's an unfortunate assumption here that the HTTP and HTTPS ports are the only thing we're dealing with, when in reality there may be many ports, supported by HTTP_EXPOSE or HTTPS_EXPOSE in extra services. (We don't have a problem with the database port, which is not handled by the router, but many other ports are)

rfay:
Overall, I'm worried about the lack of generality here (only handling router_http_port and router_https_port and nothing else). However, those are the most common conflicts (80 and 443) so there could be great advantage in just doing this. 

As noted, one of the areas lacking generality is the handling of available ports. We don't need a separate range for http and https, we just need something that gives us an available port when one is not available. Like `p := GimmeAnAvailablePortNow()`

rfay:
@agviu do you have time for a screenshare session, preferably today? I'd like to talk through options. https://cal.com/randyfay/30min

Where I'm at, I'm sure that a generalized solution at `ddev start` will be fine. But I can't figure out how to handle storing the data for `ddev describe`, `ddev list`, and the like. What you were doing was to recreate the port information, which I'm pretty sure isn't a sustainable or reliable technique.  We could create a new text file, which I hate to do. We could stash in the router (you were kind of doing this with an env var, can be done other ways). I don't really like anything I've thought of. And we probably can't update the .ddev/config.yaml, as it would be changing their configuration.

stasadev (changes requested):
I ran a random test with two projects and it generally worked.

Note: `$DDEV_PRIMARY_URL` is not updated on the host side, `ddev launch` doesn't have the port.

Which explains why `TestEnvironmentVariables` failed.

pkg/ddevapp/router.go
```diff
@@ -137,6 +149,15 @@ func generateRouterCompose() (string, error) {
...
 	uid, gid, username := util.GetContainerUIDGid()
 
+	var ephemeralPorts []string
+	if ephemeralRouterHTTPPort != "" && originalRouterHTTPPort != "" {
+		ephemeralPorts = append(ephemeralPorts, strings.Join([]string{originalRouterHTTPPort, ephemeralRouterHTTPPort}, ":"))
+	}
+	if ephemeralRouterHTTPSPort != "" && originalRouterHTTPSPort != "" {
+		ephemeralPorts = append(ephemeralPorts, strings.Join([]string{originalRouterHTTPSPort, ephemeralRouterHTTPSPort}, ":"))
+	}
+	ephemeralPortsStr := strings.Join(ephemeralPorts, ",")
```
rfay, line 159: This strategy can't work at all, because one project may have a different router_http*_port from another, and this technique overwrites the environment variable in the router, losing track of the first project's information I'm pretty sure this strategy can't work in that case at all.

rfay:
Thanks @agviu ! My current exploration is using the HTTP_EXPOSE and HTTPS_EXPOSE, which seem to be working right and are already set correctly.

> Each container has a writting available layer where we can store a file that would be ephemeral

I think you're talking about each *image*. Images have layers, containers don't. But the image would not work for this. If I'm wrong, explain more :)

> Docker provides a tmpfs option

That's an interesting option, maintaining a tmpfs mount could work. 

BTW, the env var on ddev-router definitely can't work, because one project would overwrite another's env var.

agviu:
Hi @rfay ,

> I think you're talking about each image. Images have layers, containers don't. But the image would not work for this. If I'm wrong, explain more :)
I am talking about the last layer of an image, that is writable but ephemeral. It's more explained here: https://docs.docker.com/storage/storagedriver/#storage-drivers-versus-docker-volumes

> That's an interesting option, maintaining a tmpfs mount could work.
I think so too. It could be like the ddev-global-cache volume which is attached to pretty much all the contaniers generated by DDEV (as far as I have found), but in memory and ephemeral.
However the implementation would be also more complex and I'm not sure if it is worth for what we need to store in there... maybe this solution would give you ideas for other DDEV temporal data to be stored there in the future.  In this case I think it would be worht the effort.

> the env var on ddev-router definitely can't work, because one project would overwrite another's env var.
I get your point. Would it work if we add a functionality that reads the value in that env variable before overwritting it, ensuring that both the existing value and the new value are mixed accordingly?

rfay:
I think this is probably ready for review again. Thanks to both of you for your patience.

The strategy here is that we use HTTP_EXPOSE and HTTPS_EXPOSE to get the port information. Those are already gathered into app.ComposeYAML, so they're already available. This technique is probably extensible if we ever decide to do it, but it currently supports the http, https, and mailpit http/https ports. 

The magic piece is that we empty app.ComposeYaml early in app.Start(), so we have to go look and find out what the ports are, and then the new ports (if needed) are gathered into the writing of the new compose yaml right before docker-compose up.

There is no extra storage anywhere for this technique, but OTOH it may be a bit too magical. It uses techniques that were already built into DDEV.

Please try it out and see what you think.

The one thing that it doesn't do any more (that it did do in @agviu 's version) is that if the regular configured ports are unavailable, each new project will use a new set of ports. In @agviu 's version, if 80/443 were already in use, a new set of ports was chosen and (mostly) new projects would get those same ports. However, since this is really for people who already don't have 80/443 available, or who don't have global router_https_port set to something else, it seems like it's not that big a problem, and if they're in that situation they may not even know the difference.

agviu:
Hi @rfay ,
wow you did a great job!
I have been testing this manually. Everything works as expected, with the only exception of the mailpit command, which does not generate the correct URL and uses always the default https mailtpit port.
I could not make this work, but I noticed that there're 2 env variables that seems to be used for the same: DDEV_MAILPIT_HTTP_PORT and DDEV_MAILPIT_PORT. According to documentation, only DDEV_MAILPIT_HTTP_PORT is provided...

stasadev (approved):
Wow, this works really nice.

I tested five projects at the same time with different configurations, some using explicit ports for the router/mailpit, some using busy ports and some using defaults. The `ddev launch` and `ddev mailpit` commands worked as expected for each of them.

I really like the way `ddev debug test` works now, no more errors for `curl` and the debug output is clear.

> Everything works as expected, with the only exception of the mailpit command, which does not generate the correct URL and uses always the default https mailtpit port when an ephemeral port is used for mailpit.

@agviu, could you please provide a way to reproduce it because it works for me no matter what configuration I try.

pkg/ddevapp/router_test.go
```diff
@@ -15,19 +18,27 @@ import (
...
 	"github.com/stretchr/testify/require"
 )
 
 // TestGlobalPortOverride tests global router_http_port and router_https_port
 func TestGlobalPortOverride(t *testing.T) {
+	if dockerutil.IsLima() || dockerutil.IsColima() {
+		// Intermittent failures in CI due apparently to https://github.com/lima-vm/lima/issues/2536
+		// Expected port is not available, so it allocates another one.
+		t.Skip("Lima and Colima allocate another port, so skip")
+	}
```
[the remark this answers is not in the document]
rfay: I don't think so, they claim that Rancher has a different technique for ports. But I guess we'll find out when it lands, will have to do a followup if that same malady happens to Rancher.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +281,109 @@ func TestDisableHTTP2(t *testing.T) {
...
+// Test the function FindEphemeralPort
+func TestFindEphemeralPort(t *testing.T) {
+	assert := asrt.New(t)
+
+	localIP, _ := dockerutil.GetDockerIP()
+
+	// Get a random port number in the dynamic port range
+	startPort := 49152 + rand.Intn(65535-49152+1)
+	goodEndPort := startPort + 3
+	badEndPort := startPort + 2
```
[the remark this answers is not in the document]
rfay: It probably doesn't need to be so fancy anyway.

pkg/ddevapp/router_test.go
```diff
@@ -266,3 +281,109 @@ func TestDisableHTTP2(t *testing.T) {
...
+	require.Equal(t, startPort+3, port)
+}
+
+// Test that the app assigns an ephemeral port if the default one is not available.
+func TestUseEphemeralPort(t *testing.T) {
+	if dockerutil.IsColima() || dockerutil.IsLima() {
+		// Intermittent failures in CI due apparently to https://github.com/lima-vm/lima/issues/2536
+		// Expected port is not available, so it allocates another one.
+		t.Skip("Skipping on Lima/Colima as ports don't seem to be released properly in a timely fashion")
+	}
```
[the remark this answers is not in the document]
rfay: We'll find out, don't think so at this point.

pkg/ddevapp/router.go
```diff
@@ -53,16 +67,62 @@ func StopRouterIfNoContainers() error {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			util.Debug("Timeout reached, stopping check.")
+			return
+		case <-ticker.C:
+			allReleased := true
+			for _, port := range ports {
+				if netutil.IsPortActive(port) {
```
agviu, line 111:
```suggestion
				if PortIsAvailable(port) {
```

agviu:
> > Everything works as expected, with the only exception of the mailpit command, which does not generate the correct URL and uses always the default https mailtpit port when an ephemeral port is used for mailpit.
> 
> @agviu, could you please provide a way to reproduce it because it works for me no matter what configuration I try.

I can give you a bit more of information here:
- Using 'ddev launch -m' correctly opens the mailpit's dashboard.
- Using 'ddev mailpit' uncorrectly opens the project's URL :8026, giving an error.

I tried this with 2 different projects I have and it happens in both.
Can you check ? @stasadev @rfay

agviu:
I updated the code with minor changes, mostly updating comments.

rfay:
Thanks @agviu . I simplified the logic of TestFindEphemeralPort just a bit, and used the constant there instead of a numeric literal. Also excluded Lima/Colima from another test that they don't do well with.

rpkoller:
i've manually tested with the instructions provided in the issue summary. used the nginx docker line and then added the following two lines to the config.yaml of two projects:

```
router_http_port: "80"
router_https_port: "443"
```

an excerpt from the startup process of the first project 
```
Network ddev_default removed 
Starting drupalcore... 
Port 80 is busy, using 33000 instead. 
Port 443 is busy, using 33001 instead. 
Network ddev_default created 
```
and the excerpt from the second
```
Starting yoyuu... 
Port 80 is busy, using 33002 instead. 
Port 443 is busy, using 33003 instead. 
Using custom PHP configuration: [
	/Users/rkoller/Sites/yoyuu/.ddev/php/something.ini
] 
Custom configuration is updated on restart.
If you don't see your custom configuration taking effect, run 'ddev restart'. 
Building project images... 
```

on the second project i was using a custom php ini. it only contains: 
```
[PHP]
apc.shm_size ="33M"
```
is that the expected behavior? not sure i saw that kind of output on regular starts with the regular version of HEAD i am using. aside that minor detail it looks like the functionality this issue is about is working as expected.

rfay:
I'm not sure I understand the question. If you have custom PHP, it always notifies you about it on `ddev start`, has nothing to do with this PR.

rfay:
I want to take another round on this, both looking at the code and manual testing. I'm pretty sure we could revert the key parts fairly easily if we end up not liking it, but don't really want that to have to happen.

rfay (approved):
I updated a number of comments, docs, and the name of one test and looked through it all. I think it's ok. 

- [x] Works with router_http_port/router_https_port set globally or on project
- [x] Works if nonstandard ports are blocked (router_http_port=8080 but owned by another process)
- [x] Works on gitpod
- [x] `ddev describe` shows correct values and URLs
- [x] `ddev list` shows correct default URL
- [ ] critical look at ddev debug test 

I would appreciate it if both of you @stasadev and @agviu  would give another look and manual test and make sure you think it's OK, and then let's get it in and get some experience with it.

agviu:
Hi @rfay , I have been testing it manually with 2 different projects. They both get different ports and these ports are correctly displayed when running list & describe commands. Also, the launch & mailpit commands work as expected.

stasadev (approved):
Looks good to me!
This feature will definitely become a favorite of many people, thank you @agviu and @rfay.

rfay:
I think it will make the first start for some people easier. But people with 80/443 blocked will be confused by the new port every time. Hopefully they'll then find their way to `router_http_port` and `router_https_port`

Probably we should say something about those settings when we're saying "port occupied, using ..." ?

stasadev:
> Probably we should say something about those settings

We could print a link to the docs https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#web-server-ports-already-occupied
If it's too long, you can shorten it with https://github.com/ddev/ddev.com/blob/main/public/_redirects

rfay:
I didn't even know we had that redirect feature! Awesome!

I do think the whole secton on ports is now too long and convoluted. It will need some work, I don't think we should mess with it right now. Love the redirects!

rfay:
Congrats @agviu ! thanks for taking this on!

rfay:
We probably have a problem with colima vz:
* https://buildkite.com/ddev/macos-colima-vz/builds/1538#01917fcd-2838-4324-840d-0263762f3b0f/5755-6041
> ddevapp_test.go:2131:
  | Error Trace:	/Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-6/ddev/macos-colima-vz/pkg/ddevapp/ddevapp_test.go:2131
  | Error:      	Received unexpected error:
  | unable to listen on required ports, port 33000 is already in use,
  | Troubleshooting suggestions at https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#unable-listen
  | Test:       	TestWebserverMariaMySQLDBClient
 

Same on lima:
* https://buildkite.com/ddev/macos-lima/builds/1368#01917fcd-1b6f-4167-bb3e-aa70214e2ac6/5820-6390

> 2024-08-23T09:40:25.682 Copied global Traefik config in /Users/testbot/.ddev/traefik/certs to ddev-global-cache/traefik
  | ddevapp_test.go:2131:
  | Error Trace:	/Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-7/ddev/macos-lima/pkg/ddevapp/ddevapp_test.go:2131
  | Error:      	Received unexpected error:
  | unable to listen on required ports, port 33002 is already in use,
  | Troubleshooting suggestions at https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#unable-listen
  | Test:       	TestWebserverMariaMySQLDBClient

tyler36:
> Probably we should say something about those settings when we're saying "port occupied, using ..." ?

Sounds like a good canidate for the message-of-the-day too: [ddev/remote-config](https://github.com/ddev/remote-config)

PR #6415, opened 2024-07-22
fix: default_container_timeout should work right, fixes #5133 again

### LINKED ISSUE #5133: default_container_timeout not working

rfay:
## The Issue

* #5133 - over and over again

`default_container_timeout` hasn't had the behavior we expect, even though we've changed that behavior lots of times.

## TODO

- [x] app.GetMaxContainerWaitTime() needs to estimate a start_period if none is specified
- [x] Make TestConfigDefaultContainerTimeout more comprehensive (include 3rd party services)

Resolve the differences between the docker-compose `timeout` and what we need, `start_period` or defaults for `start_period`

* Read [blog on this](https://ddev.com/blog/ddev-and-docker-healthchecks-technote)
* Try `ddev snapshot restore` with a larger snapshot
* Try it with different `default_container_timeout` - larger ones should result in more time for the snapshot restore
* Try `docker-compose.*.yaml` with a variety of healthcheck values, especially `start_period` There are a number of these for examples in the TestConfigDefaultContainerTimeout directory.

## Automated Testing Overview

New `TestConfigDefaultContainerTimeout`, maybe it can get me out of hell eventually. This has a ways to go yet.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1595,9 +1595,9 @@ func FindNotOmittedImages(app *DdevApp) []string {
 }
 
 // FindMaxTimeout looks through all services and returns the max timeout found
-// Defaults to 120
+// Defaults to DefaultContainerTimeout (usually 120 unless configured)
 func (app *DdevApp) FindMaxTimeout() int {
-	const defaultContainerTimeout = 120
+	defaultContainerTimeout, _ := strconv.Atoi(app.DefaultContainerTimeout)
 	maxTimeout := defaultContainerTimeout
 	if app.ComposeYaml == nil {
 		return defaultContainerTimeout
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1628,7 +1628,11 @@ func (app *DdevApp) GetMaxContainerWaitTime() int {
 					if retriesSpecified, ok := i.(map[string]interface{})["retries"]; ok {
 						retries = retriesSpecified.(int)
 					}
-					maxWaitTime = retries * interval
+					// If the guessed maxWaitTime is greater than what we've found before
+					// then use it. This will be unusual.
+					if retries*interval > maxWaitTime {
+						maxWaitTime = retries * interval
+					}
 				}
 			}
 		}
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1600,6 +1600,7 @@ func FindNotOmittedImages(app *DdevApp) []string {
 func (app *DdevApp) GetMaxContainerWaitTime() int {
 	defaultContainerTimeout, _ := strconv.Atoi(app.DefaultContainerTimeout)
 	maxWaitTime := defaultContainerTimeout
+
 	if app.ComposeYaml == nil {
 		return defaultContainerTimeout
 	}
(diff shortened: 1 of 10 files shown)


rfay:
Ready for review. Maybe I won't have to do this again for a few months???

pkg/ddevapp/ddevapp.go
```diff
@@ -1594,31 +1594,52 @@ func FindNotOmittedImages(app *DdevApp) []string {
...
 					t := int(duration.Seconds())
-					if t > maxTimeout {
-						maxTimeout = t
+					if t > maxWaitTime {
+						maxWaitTime = t
+					}
+				} else { /* In this case we didn't have a specified start_period, so guess at one */
+					interval := 1
+					retries := 3
+					maxWaitTime = 15 /* Use a generic default for start_period/maxWaitTime based on 3*5 */
```
stasadev, line 1622: The comment is confusing.
I don't understand where `15` comes from, `3*5`, if `3` is retries, than what is `5` ?
Does it come from `600/120` ? (600 is the value I see in `snapshot.go`.)
In this case the comment about `start_period/maxWaitTime` should be `maxWaitTime/start_period`.

stasadev (approved):
Look good to me.

I just have one question about the default value for the guessed `maxWaitTime`.

pkg/ddevapp/ddevapp.go
rfay: 5 is the default `start_period`. If the `test` were to fail 3 times, it would be retried in 5 seconds, meaning 15 seconds or more. I'll see if I can add a comment there.

pkg/ddevapp/ddevapp.go
```diff
@@ -1594,31 +1594,54 @@ func FindNotOmittedImages(app *DdevApp) []string {
...
-					if t > maxTimeout {
-						maxTimeout = t
+					if t > maxWaitTime {
+						maxWaitTime = t
+					}
+				} else { /* In this case we didn't have a specified start_period, so guess at one */
+					interval := 1
+					retries := 3
+					// Use a generic default for start_period/maxWaitTime based on
+					// `retries` (3) * default `interval` (5)
```
stasadev, line 1623: I believe it's `start_interval`, not `interval`, and let's add a reference to where it comes from.

```suggestion
					// https://docs.docker.com/reference/dockerfile/#healthcheck
					// `retries` (3) * default `start_interval` (5)
```

stasadev (approved):
It looks good 👍

PR #6420, opened 2024-07-23
feat: add support for `n install auto`, add `N_PREFIX` cache, replace node repo with `n`, fixes #6418

### LINKED ISSUE #6418: Allow DDEV nodejs_version to use external file as reference for node version

* https://github.com/ddev/ddev/issues/6071

Currently I have a file that holds my node version used by various other parts of the project

Ultimately I'd like to not have to manage DDEV's config independantly of the other details that read this file.

### Describe your solution

Allowing `nodejs_version:` to accept `path/to/file` OR `version`

It would just take the version number from a defined file.
It can break if you pass something wrong, that's on you.

The filename is irrelevant.

### Describe alternatives

N/A

### Additional context

https://github.com/ddev/ddev/issues/6071

stasadev:
## The Issue

- #6418

From https://www.npmjs.com/package/n#specifying-nodejs-versions:

There is an `auto` label to read the target version from a file in the current directory, or any parent directory. `n` looks for in order:

- `.n-node-version`: version on single line. Custom to `n`.
- `.node-version`: version on single line. Used by multiple tools: [node-version-usage](https://github.com/shadowspawn/node-version-usage)
- `.nvmrc`: version on single line. Used by `nvm`.
- if no version file found, look for `engine` as below.

The `engine` label looks for a `package.json` file and reads the `engines` field to determine compatible Node.js. Requires an installed version of `node`, and uses `npx semver` to resolve complex ranges.

---

This PR adds:
- cache for `n` just like it works for `nvm`.
- full compatibility for `n install`, as previously it was run from a `Dockerfile` with no mounted code, which limited its use.

Review the docs https://ddev--6420.org.readthedocs.build/en/6420/users/configuration/config/#nodejs_version

```
ddev config --nodejs-version=auto
echo "10" > .nvmrc
ddev restart
ddev logs -s web # to see how it was installed
ddev exec node --version
```

Repeat the same with different options from https://www.npmjs.com/package/n#specifying-nodejs-versions

```
ddev config --nodejs-version=engine
echo '{
    "engines": {
        "node": "14.20"
    }
}' > package.json
ddev restart
ddev logs -s web # to see how it was installed
ddev exec node --version
```


docs/content/users/configuration/config.md
```diff
@@ -393,6 +393,9 @@ Node.js version for the web container’s “system” version.
 
 There is no need to configure `nodejs_version` unless you want a version other than the default version.
 
+!!!tip "Select the Node.js version from a file such as `.nvmrc` or `package.json`"
+    `nodejs_version` can be set to `auto`, which will use `n install auto` under the hood.
```
GuySartorelli, line 397: Probably worth either linking to `n` docs or explaining what that means. People who want this functionality may not know what `n` is or what `n install auto` does.

pkg/ddevapp/templates.go
```diff
@@ -70,7 +70,7 @@ const ConfigInstructions = `
 # change from the default system Node.js version to any other version.
 # Numeric version numbers can be complete (i.e. 18.15.0) or
 # incomplete (18, 17.2, 16). 'lts' and 'latest' can be used as well along with
-# other named releases.
+# other named releases, 'auto' reads the version from a file such as '.nvmrc' or 'package.json'.
```
GuySartorelli, line 73: In the pr description you mentioned an `engine` setting for checking `package.json`, but here you say `auto` will do that?

GuySartorelli: May be worth mentioning the other files that it reads from (there's only 3 after all) to avoid people thinking it's only `.nvmrc`

GuySartorelli (commented):
Will test tomorrow, but glancing at it, this is a brilliant solution.

Couple things that stood out in the communication about the feature below.

containers/ddev-webserver/ddev-webserver-prod-scripts/start.sh
```diff
@@ -92,6 +92,10 @@ ls /var/www/html >/dev/null || (echo "/var/www/html does not seem to be healthy/
 
 mkdir -p /mnt/ddev-global-cache/{bashhistory/${HOSTNAME},mysqlhistory/${HOSTNAME},nvm_dir/${HOSTNAME},npm,yarn/classic,yarn/berry,corepack}
 
+if [ "${N_INSTALL_VERSION:-}" != "" ]; then
+  sudo n install "${N_INSTALL_VERSION}"
```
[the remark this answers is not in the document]
stasadev: My mistake with `sudo`, thanks, and yes, it should also be available for prod/hardened, just like it works for `nvm`.

docs/content/users/configuration/config.md
stasadev: Thanks, that's a good suggestion.

pkg/ddevapp/templates.go
stasadev: I didn't even notice that I copied the part with `engine` 🙂, it doesn't work now, but it will.
I'm going to remove my hardcoded condition for `auto` so that it accepts anything that works for `n`.

stasadev:
I have updated the code:

- Run `docker pull ddev/ddev-webserver:20240723_stasadev_n_install_auto` if you have already tested this PR.
- Fixed the issue where the node version was pulled every time by adding the `N_PREFIX` cache directory.
- Added `n install --offline`, so that it doesn't crash if you don't have internet.
- Updated the docs and added an example of what to do if you use `n install auto` and your file, for example `.nvmrc`, is not in the root of the project.

containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
```diff
@@ -1,4 +1,6 @@
-export PATH="~/bin:$PATH:/var/www/html/vendor/bin:/var/www/html/bin:$DDEV_COMPOSER_ROOT/vendor/bin"
+PATH="$PATH:/var/www/html/vendor/bin:/var/www/html/bin:$DDEV_COMPOSER_ROOT/vendor/bin"
```
rfay, line 1: Perhaps slightly better

```suggestion
PATH="${PATH}:/var/www/html/vendor/bin:/var/www/html/bin:${DDEV_COMPOSER_ROOT}/vendor/bin"
```

However, I don't understand the removal of ~/bin. I think people might add things there with homeadditions.

containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
```diff
@@ -93,8 +93,16 @@ ls /var/www/html >/dev/null || (echo "/var/www/html does not seem to be healthy/
...
 sudo mkdir -p ${TERMINUS_CACHE_DIR}
 
-sudo mkdir -p /mnt/ddev-global-cache/{bashhistory/${HOSTNAME},mysqlhistory/${HOSTNAME},nvm_dir/${HOSTNAME},npm,yarn/classic,yarn/berry,corepack}
+sudo mkdir -p /mnt/ddev-global-cache/{bashhistory/${HOSTNAME},mysqlhistory/${HOSTNAME},n_prefix/${HOSTNAME},nvm_dir/${HOSTNAME},npm,yarn/classic,yarn/berry,corepack}
 sudo chown -R "$(id -u):$(id -g)" /mnt/ddev-global-cache/ /var/lib/php
+
+if command -v n >/dev/null 2>&1 && [ "${N_PREFIX:-}" != "" ] && [ "${N_INSTALL_VERSION:-}" != "" ]; then
+  ln -sf /mnt/ddev-global-cache/n_prefix/${HOSTNAME} "${N_PREFIX}"
+  # try a normal install that also uses cache and try again offline if it fails
+  n install "${N_INSTALL_VERSION}" || n install "${N_INSTALL_VERSION}" --offline
```
rfay, line 102:
```suggestion
  n install "${N_INSTALL_VERSION}" || n install "${N_INSTALL_VERSION}" --offline || true
```

`n install --offline` fails if it doesn't find the version in the cache; I think we'd be better continuing.

containers/ddev-webserver/ddev-webserver-prod-scripts/start.sh
```diff
@@ -90,7 +90,14 @@ disable_xhprof
 
 ls /var/www/html >/dev/null || (echo "/var/www/html does not seem to be healthy/mounted; docker may not be mounting it., exiting" && exit 101)
 
-mkdir -p /mnt/ddev-global-cache/{bashhistory/${HOSTNAME},mysqlhistory/${HOSTNAME},nvm_dir/${HOSTNAME},npm,yarn/classic,yarn/berry,corepack}
+mkdir -p /mnt/ddev-global-cache/{bashhistory/${HOSTNAME},mysqlhistory/${HOSTNAME},n_prefix/${HOSTNAME},nvm_dir/${HOSTNAME},npm,yarn/classic,yarn/berry,corepack}
+
+if command -v n >/dev/null 2>&1 && [ "${N_PREFIX:-}" != "" ] && [ "${N_INSTALL_VERSION:-}" != "" ]; then
+  ln -sf /mnt/ddev-global-cache/n_prefix/${HOSTNAME} "${N_PREFIX}"
+  # try a normal install that also uses cache and try again offline if it fails
+  n install "${N_INSTALL_VERSION}" || n install "${N_INSTALL_VERSION}" --offline
```
rfay, line 98:
```suggestion
  n install "${N_INSTALL_VERSION}" || n install "${N_INSTALL_VERSION}" --offline || true
```

docs/content/users/configuration/config.md
```diff
@@ -385,13 +385,37 @@ Whether to skip mounting project into web container.
 
 ## `nodejs_version`
 
-Node.js version for the web container’s “system” version.
+Node.js version for the web container’s “system” version. [`n`](https://www.npmjs.com/package/n) tool is under the hood.
+
+There is no need to configure `nodejs_version` unless you want a version other than the default version.
```
rfay, line 390:
```suggestion
There is no need to reconfigure `nodejs_version` unless you want a version other than the version already specified, which will be the default version at the time the project was configured.
```

docs/content/users/configuration/config.md
```diff
@@ -385,13 +385,37 @@ Whether to skip mounting project into web container.
...
 
 | Type | Default | Usage
 | -- | -- | --
 | :octicons-file-directory-16: project | current LTS version | any [node version](https://www.npmjs.com/package/n#specifying-nodejs-versions), like `16`, `18.2`, `18.19.2`, etc.
 
-There is no need to configure `nodejs_version` unless you want a version other than the default version.
+!!!tip "How to install the Node.js version from a file"
+    There is an `auto` label (see [full documentation](https://www.npmjs.com/package/n#specifying-nodejs-versions)):
```
rfay, line 397:
```suggestion
Your project team may specify the Node.js version in a more general way than in the `.ddev/config.yaml`. For example, you may use a `.nvmrc` file, the `package.json`, or a similar technique. In that case, DDEV can use the external configuration provided by that file.

    There is an `auto` label (see [full documentation](https://www.npmjs.com/package/n#specifying-nodejs-versions)):
```

pkg/ddevapp/config.go
```diff
@@ -973,8 +973,11 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 	extraWebContent := "\nRUN mkdir -p /home/$username && chown $username /home/$username && chmod 600 /home/$username/.pgpass"
 	extraWebContent = extraWebContent + "\nENV NVM_DIR=/home/$username/.nvm"
 	if app.NodeJSVersion != nodeps.NodeJSDefault {
-		extraWebContent = extraWebContent + "\nRUN npm install -g n"
-		extraWebContent = extraWebContent + fmt.Sprintf("\nRUN n install %s && ln -sf /usr/local/bin/node /usr/local/bin/nodejs", app.NodeJSVersion)
+		extraWebContent = extraWebContent + fmt.Sprintf(`
+RUN npm install -g n
```
rfay, line 977: It's a little confusing that we try to do this at build time and then later during start.sh. But it seems reasonable.

```suggestion
RUN npm install -g n || true
```

pkg/ddevapp/templates.go
```diff
@@ -68,12 +68,24 @@ const ConfigInstructions = `
...
+# or any parent directory. 'n' looks for in order:
+#   - ".n-node-version": version on single line. Custom to 'n'.
+#   - ".node-version": version on single line. Used by multiple tools https://github.com/shadowspawn/node-version-usage
+#   - ".nvmrc": version on single line. Used by 'nvm'.
+#   - if no version file found, look for "engine" as below.
+#
+# The "engine" label looks for a 'package.json' file and reads the engines field to determine compatible Node.js.
+#
+# Note that using of 'ddev nvm' is discouraged because "nodejs_version" is much easier to use,
+# can specify any version, and is more robust than using 'nvm'.
```
rfay, line 88: I think this section would be better as a link to the docs. It's probably too verbose here. We should probably reconsider most of what's here in template.go. I wonder if people find it useful.

rfay (changes requested):
Well, this is pretty awesome! I tested it and it worked without any questions at all. I have a number of minor suggestions. One of the things I really appreciate about you is how you study the possibilities and really understand them, rather than coming up with the most obvious "quick think" solution. All that conversation in the issue(s) and then you latched on to the excellent `n` techniques and adopted them. It's amazing work.

containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
stasadev: > I don't understand the removal of `~/bin`

I didn't remove it, see the line below. I preserved the original order, so the binaries inside `N_PREFIX` will not override something in `~/bin`.

containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
stasadev: I thought about this, but didn't implement it.

I think it's for good, I wish we could somehow notify the user of the error here.

Maybe I could check for `node --version` somewhere later in the `ddev start`, compare it with config and show a Go warning...

pkg/ddevapp/config.go
stasadev: This is a different thing, we run only `n install` during `start.sh`.

Maybe we should install it on `ddev-webserver` so it's available right away?

pkg/ddevapp/templates.go
stasadev: Okay, a link to our docs sounds like a good idea to avoid this duplication.

pkg/ddevapp/config.go
rfay: I agree, we should have `n` already installed in ddev-webserver.

containers/ddev-webserver/Dockerfile
```diff
@@ -55,6 +55,9 @@ RUN curl --fail -sSL https://raw.githubusercontent.com/netz98/n98-magerun/develo
 RUN curl --fail -sSL https://files.magerun.net/n98-magerun2-latest.phar -o /usr/local/bin/magerun2 && chmod 777 /usr/local/bin/magerun2
 RUN curl --fail -sSL https://raw.githubusercontent.com/netz98/n98-magerun2/develop/res/autocompletion/bash/n98-magerun2.phar.bash -o /etc/bash_completion.d/n98-magerun2.phar && chmod +x /usr/local/bin/magerun
 
+# n for nodejs_version
+RUN npm install -g n
+
```
rfay, line 60: I think it would be better to install `n` directly and then install nodejs using it. That way we don't have the /usr/bin/node and /usr/bin/npm hanging around and causing possible confusion.

containers/ddev-webserver/ddev-webserver-base-scripts/start.sh
```diff
@@ -93,8 +93,28 @@ ls /var/www/html >/dev/null || (echo "/var/www/html does not seem to be healthy/
...
+  if [ "${n_install_result}" = "true" ]; then
+    ln -sf ${N_PREFIX}/bin/node ${N_PREFIX}/bin/nodejs
+    # we don't need this file if everything is fine
+    rm -f /tmp/n-install-stderr.txt
+  else
+    # remove the symlink on error so that the system Node.js can be used
+    rm -f ${N_PREFIX}
+  fi
+  hash -r
+fi
```
rfay, line 116: Could we try moving this into a script that can be maintained separately? Like /usr/local/bin/n_install.sh or something?

pkg/ddevapp/ddevapp.go
```diff
@@ -1473,6 +1473,16 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
...
 
+	if app.NodeJSVersion != nodeps.NodeJSDefault {
+		util.Debug(`checking nodejs_version: "%s" install for errors`, app.NodeJSVersion)
+		nInstallStderr, _, _ := app.Exec(&ExecOpts{
+			Cmd: "cat /tmp/n-install-stderr.txt || true",
+		})
+		if nInstallStderr != "" {
+			util.Warning("Unable to install nodejs_version: \"%s\", error output from `n install %s`:\n%s", app.NodeJSVersion, app.NodeJSVersion, nInstallStderr)
+		}
+	}
```
rfay, line 1484: Consider generalizing this approach for other things that we skip over when offline, etc. Doesn't have to be done in this PR, but definitely worth thinking about. (This works quite nicely btw)

rfay (changes requested):
I tested what I think are the key changes from last time, and made a few suggestions. It works great.

stasadev:
Implemented the requested changes.

Do not forget to run `docker pull ddev/ddev-webserver:20240723_stasadev_n_install_auto` before testing.

stasadev:
Resolved conflicts, pushed `ddev-php-base` and `ddev-webserver`.

stasadev:
I didn't check at first, but this change for `n` adds +46Mb to the container.

@rfay, should I still add the change from the node repository to `n`?

ddev-php-base
ddev-php-base

ddev-webserver
ddev-webserver

rfay:
Let's try to figure out where the size increase comes from. It doesn't make too much sense, since we're getting rid of node and npm, and moving them to a docker volume. There may be something that got lost. Thanks for paying attention. We do want to avoid increasing size this much on a PR like this.

stasadev:
Now everything is fine
image

containers/ddev-webserver/Dockerfile
```diff
@@ -165,7 +165,7 @@ RUN chmod -R 777 /var/log
 # we need to create the /var/cache/linux and /var/lib/nginx manually for the arm64 image and chmod them, please don't remove them!
 RUN mkdir -p /mnt/ddev-global-cache/mkcert /run/{php,blackfire} /var/cache/nginx /var/lib/nginx && chmod -R ugo+rw /mnt/ddev-global-cache/
 
-RUN chmod -fR ugo+w /usr/sbin /usr/bin /etc/nginx /var/cache/nginx /var/lib/nginx /run /var/www /etc/php/*/*/conf.d/ /var/lib/php/modules /etc/alternatives /usr/lib/node_modules /etc/php /etc/apache2 /var/log/apache2/ /var/run/apache2 /var/lib/apache2 /mnt/ddev-global-cache/*
+RUN chmod -fR ugo+w /usr/sbin /usr/bin /etc/nginx /var/cache/nginx /var/lib/nginx /run /var/www /etc/php/*/*/conf.d/ /var/lib/php/modules /etc/alternatives /usr/local/lib/node_modules /etc/php /etc/apache2 /var/log/apache2/ /var/run/apache2 /var/lib/apache2 /mnt/ddev-global-cache/*
```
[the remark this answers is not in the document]
stasadev: Yes, `n` uses this folder.

GuySartorelli:
Sorry it took me so long to test this, but just tested now and it works exactly as expected.

I did notice that `ddev describe` will show `NodeJS:auto` instead of the version being used. It would be _nice_ if it showed the actual version in use... but on the other hand maybe saying "auto" here is better in some ways because it's telling the dev "it's not hardcoded in ddev configuration".
Either way having it say "auto" is the easier solution so probably no reason to change it.

rfay (approved):
This is great, I'm sure you didn't expect to spend this long on this. Thank you.

Testing, I note that if `auto` is set and nothing done to configure it, we see

```
Unable to install nodejs_version: "auto".
Error output from `n install auto`:

  Error: no file found for auto version (.n-node-version, .node-version, .nvmrc, or package.json)

     offline : matching cached versions

  Error: no file found for auto version (.n-node-version, .node-version, .nvmrc, or package.json)
```

node v20 is installed though. In a perfect world it would probably say "installing default nodejs_version 20" instead.  This is minor though and you're sick of this, so OK to get it in!

containers/ddev-php-base/Dockerfile
```diff
@@ -98,9 +96,14 @@ RUN apt-get -qq install --no-install-recommends --no-install-suggests -y \
     jq \
     msmtp \
     nginx \
-    nodejs \
     sqlite3
 
+RUN curl -fsSL https://raw.githubusercontent.com/tj/n/master/bin/n | bash -s "${NODE_VERSION}"
```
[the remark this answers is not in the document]
shadowspawn: You could just save the curl download rather than piping it into bash, if that is a good enough install for your purposes. Basic idea:

```
curl -fsSL https://raw.githubusercontent.com/tj/n/master/bin/n -o /usr/local/bin/n
chmod a+x /usr/local/bin/n
n install "${NODE_VERSION}"
```

Longer explanation: https://github.com/tj/n/issues/809#issuecomment-2264129789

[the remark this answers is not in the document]
rfay: If @shadowspawn says it, it must be true! (thanks for being here!) @stasadev had gone the direction of the documented `n` install-with-bash technique, and it required `make` to be used.

rfay:
Congrats all!

containers/ddev-php-base/Dockerfile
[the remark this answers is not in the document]
stasadev: Thank you, @shadowspawn.

I will change this instruction to `n` when we prepare the next release of DDEV.

PR #6422, opened 2024-07-24
fix: show containers we're waiting for and better error message on timeout/unhealthy

rfay:
## The Issue

* In a complex project, there may be many containers we're waiting for. Show what we're waiting for when waiting for them
* Although we try hard with our error message, people seem not to read or understand it. Try to improve readability

* Show containers we're waiting for
* Improve error message

image

You can add a trivial failing extra service with `.ddev/docker-compose.longtimeout.yaml` with contents
```yaml
services:
  longtimeout:
    container_name: "ddev-${DDEV_SITENAME}-longtimeout"
    image: debian:12
    command: sleep infinity
    labels:
      com.ddev.approot: ${DDEV_APPROOT}
    healthcheck:
      start_period: "10s"
      interval: "11s"
      start_interval: "11s"
      retries: 2
      timeout: "15s"
      test: "sleep 9 && false"
```

Or adjust as you like.

## Automated Testing Overview

Nothing new.

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -417,7 +417,7 @@ func ContainerWait(waittime int, labels map[string]string) (string, error) {
 				return logOutput, nil
 			case "unhealthy":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
-				return logOutput, fmt.Errorf("%s container is unhealthy: %s, Troubleshot with: \n%s", name, logOutput, suggestedCommand)
+				return logOutput, fmt.Errorf("%s container is unhealthy, log=%s\nTroubleshoot this with these commands: \n%s", name, logOutput, suggestedCommand)
 			case "exited":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
 				return logOutput, fmt.Errorf("%s container exited,\nTroubleshoot this with these commands:\n%s", name, suggestedCommand)
@@ -471,7 +471,7 @@ func ContainersWait(waittime int, labels map[string]string) error {
 					continue
 				case "unhealthy":
 					name, suggestedCommand := getSuggestedCommandForContainerLog(&container)
-					return fmt.Errorf("%s container is unhealthy: %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
+					return fmt.Errorf("%s container is unhealthy, log=%s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
 				case "exited":
 					name, suggestedCommand := getSuggestedCommandForContainerLog(&container)
 					return fmt.Errorf("%s container exited.\nTroubleshoot this with these commands:\n%s", name, suggestedCommand)
@@ -529,7 +529,7 @@ func ContainerWaitLog(waittime int, labels map[string]string, expectedLog string
 				return logOutput, nil
 			case status == "unhealthy":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
-				return logOutput, fmt.Errorf("%s container is unhealthy: %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
+				return logOutput, fmt.Errorf("%s container is unhealthy, log=%s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
 			case status == "exited":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
 				return logOutput, fmt.Errorf("%s container exited\nTroubleshoot this with these commands:\n%s", name, suggestedCommand)
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -400,7 +400,7 @@ func ContainerWait(waittime int, labels map[string]string) (string, error) {
 				health, _ := GetContainerHealth(container)
 				if health != "healthy" {
 					name, suggestedCommand := getSuggestedCommandForContainerLog(container)
-					desc = desc + fmt.Sprintf(" %s:%s - more info with %s", name, health, suggestedCommand)
+					desc = desc + fmt.Sprintf(" %s:%s\nTroubleshoot this with these commands:\n%s", name, health, suggestedCommand)
 				}
 			}
 			return "", fmt.Errorf("health check timed out after %v: labels %v timed out without becoming healthy, status=%v, detail=%s ", durationWait, labels, status, desc)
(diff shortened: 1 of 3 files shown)


pkg/dockerutil/dockerutils.go
```diff
@@ -416,10 +416,10 @@ func ContainerWait(waittime int, labels map[string]string) (string, error) {
 				return logOutput, nil
 			case "unhealthy":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
-				return logOutput, fmt.Errorf("%s container is unhealthy: %s, more info with %s", name, logOutput, suggestedCommand)
+				return logOutput, fmt.Errorf("%s container is unhealthy: %s, Troubleshot with: \n%s", name, logOutput, suggestedCommand)
```
stasadev, line 419:
```suggestion
				return logOutput, fmt.Errorf("%s container is unhealthy: %s, Troubleshoot this with these commands: \n%s", name, logOutput, suggestedCommand)
```

pkg/dockerutil/dockerutils.go
```diff
@@ -470,10 +470,10 @@ func ContainersWait(waittime int, labels map[string]string) error {
 					continue
 				case "unhealthy":
 					name, suggestedCommand := getSuggestedCommandForContainerLog(&container)
-					return fmt.Errorf("%s container is unhealthy: %s, more info with %s", name, logOutput, suggestedCommand)
+					return fmt.Errorf("%s container is unhealthy: %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
```
stasadev, line 473:
```suggestion
					return fmt.Errorf("%s container is unhealthy, %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
```

pkg/dockerutil/dockerutils.go
```diff
@@ -528,10 +528,10 @@ func ContainerWaitLog(waittime int, labels map[string]string, expectedLog string
 				return logOutput, nil
 			case status == "unhealthy":
 				name, suggestedCommand := getSuggestedCommandForContainerLog(container)
-				return logOutput, fmt.Errorf("%s container is unhealthy: %s, more info with %s", name, logOutput, suggestedCommand)
+				return logOutput, fmt.Errorf("%s container is unhealthy: %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
```
stasadev, line 531:
```suggestion
				return logOutput, fmt.Errorf("%s container is unhealthy, %s\nTroubleshoot this with these commands:\n%s", name, logOutput, suggestedCommand)
```

stasadev (changes requested):
This looks better, thank you!

On Linux I see `unhealthy` error message instead, and it showed some strange output after `unhealthy: &{2024-07-24...`

```
The error was 'ddev-d10-longtimeout container is unhealthy: &{2024-07-24 19:25:47.021519304 +0300 EEST 2024-07-24 19:25:56.0959853 +0300 EEST 1 }
Troubleshoot this with these commands:
[
        ddev logs -s longtimeout
        docker logs ddev-d10-longtimeout
        docker inspect --format "{{ json .State.Health }}" ddev-d10-longtimeout | docker run -i --rm ddev/ddev-utilities jq -r
]'
```

It turned out, I forgot to update some syntax after migrating from `fsouza/go-dockerclient`, please change:

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/dockerutil/dockerutils.go#L600-L600

```diff
-logOutput = fmt.Sprintf("%v", inspect.State.Health.Log[numLogs-1])
+logOutput = fmt.Sprintf("%v", inspect.State.Health.Log[numLogs-1].Output)
```

And we need to trim spaces here (because our healthcheck for web container may return output with trailing spaces):

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/dockerutil/dockerutils.go#L613-L613

```diff
-return status, logOutput
+return status, strings.TrimSpace(logOutput)
```

And often this `logOutput` is empty, so I suggested to change `%s container is unhealthy: %s` to `%s container is unhealthy, log=%s` to make it look better.

In this case it will show the correct `logOutput` for `unhealthy`. Which makes me think about this:
```
docker inspect --format "{{ json .State.Health }}" ddev-d10-longtimeout | docker run -i --rm ddev/ddev-utilities jq -r
```
I think people already see the long output (a lot of text!), and when it opens in a small terminal window, it's not obvious that it's a command that can be run:

```
        docker inspect --format "{{ json .State.Health }}" ddev-d10-longtim
eout | docker run -i --rm ddev/ddev-utilities jq -r
```

Maybe we don't need to suggest it at all here in two places?:

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/dockerutil/dockerutils.go#L553-L553

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/dockerutil/dockerutils.go#L560-L560

I originally added it just because I saw it in the output for `ddev-router` and `ddev-ssh-agent`, but it may add value for other containers:

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/ddevapp/router.go#L115-L115

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/ddevapp/ssh_auth.go#L71-L71

rfay:
> BTW, ddev-ssh-agent could also show logOutput=%s as for ddev-router

We've never seen a problem with ddev-ssh-agent, so not worth a lot of effort IMO

rfay:
It may be OK to try to get this into v1.23.4 if we are confident it doesn't introduce bugs. We're going to need to do a little manual testing with all these last-minute PRs going in.

stasadev:
I meant this:

https://github.com/ddev/ddev/blob/2de19808b6b891433128d46775c64a417ad5b63a/pkg/dockerutil/dockerutils.go#L553-L553

```diff
-suggestedCommands = append(suggestedCommands, fmt.Sprintf("docker logs %s", name), fmt.Sprintf("docker inspect --format \"{{ json .State.Health }}\" %s | docker run -i --rm ddev/ddev-utilities jq -r", name))
+suggestedCommands = append(suggestedCommands, fmt.Sprintf("docker logs %s", name))
```

Do we really need to suggest this `docker inspect` ? Because I never saw useful info inside this command.

---

Tests seem to be failing because I suggested adding `TrimSpace()`.

rfay:
> Do we really need to suggest this docker inspect ? Because I never saw useful info inside this command.

There are a number of cases where the healthcheck gives clues when a timeout happens. That's what that is about. 

An example, if you change the `docker-compose.longtimeout.yaml` to have 
```yaml
    healthcheck:
      start_period: "10s"
      interval: "11s"
      start_interval: "11s"
      retries: 2
      timeout: "15s"
      test: "sleep 9 && echo 'oh dear this is taking a long time' && false"
```

You'll see this, which tells us why the timeout is failing
```
$      docker inspect --format "{{ json .State.Health }}" ddev-d10-longtimeout | docker run -i --rm ddev/ddev-utilities jq -r
{
  "Status": "unhealthy",
  "FailingStreak": 7,
  "Log": [
    {
      "Start": "2024-08-05T10:58:56.108347218-06:00",
      "End": "2024-08-05T10:59:05.140396323-06:00",
      "ExitCode": 1,
      "Output": "oh dear this is taking a long time\n"
    },
    {
      "Start": "2024-08-05T10:59:16.141631636-06:00",
      "End": "2024-08-05T10:59:25.17326936-06:00",
      "ExitCode": 1,
      "Output": "oh dear this is taking a long time\n"
    },
```

rfay:
I agree that maybe a `ddev healthcheck-diagnose` would be better...

stasadev:
I see, okay, let's leave it for now. The only complaint is that it is a very long command.

stasadev (approved):
Looks good to me.

Issue #6419, opened 2024-07-22
`ddev start` fails due to no internet connectivity at start time in Docker system


Adrianm6254:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
```

</details>

https://gist.github.com/Adrianm6254/74c5625e26bb4cb2ca390c1c3adebe06#file-gistfile1-txt

### Expected Behavior

Drupal 11 site to startup

### Actual Behavior

<details><summary>Drupal 11 ddev start</summary>

```
ddev start
Starting n2c-base11-d... 
Building project images... 
.failed to solve: process "/bin/bash -c mkdir -p /tmp/sqlite3 && wget -O /tmp/sqlite3/sqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && wget -O /tmp/sqlite3/libsqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/libsqlite3-0_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && apt-get install -y /tmp/sqlite3/*.deb && rm -rf /tmp/sqlite3" did not complete successfully: exit code: 4 
Failed to start n2c-base11-d: docker-compose build failed: composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-n2c-base11-d docker-compose -f /media/drive2/dev/project-builds/n2c-base11/.ddev/.ddev-docker-compose-full.yaml --progress=plain build', action='[--progress=plain build]', err='exit status 17', stdout='#0 building with "default" instance using docker driver

#1 [db internal] load build definition from Dockerfile
#1 transferring dockerfile: 656B done
[...]

PR #6426, opened 2024-07-25
fix: fail gently if sqlite3 can't be downloaded for drupal11, fixes #6425, for #6419

### LINKED ISSUE #6425: Drupal 11 sqlite3 installation should fail more gracefully

```
```
</details>

### Expected Behavior

See
* https://github.com/ddev/ddev/issues/6419

When there's no internet connectivity, a `ddev start` with a drupal11 site can fail

### Actual Behavior

It fails if it can't download sqlite3

We should fail gracefully, as we do with normal composer install or apt update. Try once or twice and then move on.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #6425
* #6419

If we can't download and install sqlite 3.45, just move on.

Most Drupal11 users don't need it anyway.

* Turn off internet
* Create new Drupal11 project
* `ddev start`
* ddev config --update
* `ddev restart`

`ddev exec dpkg -l sqlite3` should show older sqlite

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -988,11 +988,11 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 			extraWebContent = extraWebContent + "\n" + fmt.Sprintf(`
 ### Drupal 11+ requires a minimum sqlite3 version (3.45 currently)
 ARG SQLITE_VERSION=%s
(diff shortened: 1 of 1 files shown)


stasadev (approved):
Tried this PR by manually changing the install URL to the junk one and it doesn't break `ddev start`
Looks good to me.

PR #6427, opened 2024-07-26
fix: create mutagen-agents tarball if it doesn't exist, fixes #6421 

### LINKED ISSUE #6421: DownloadMutagenIfNeeded() should check for existence of agent tarball and download if it doesn't exist
### Preliminary checklist

```
```
</details>

### Expected Behavior

At `ddev start`, the correct `mutagen` and `mutagen-agents.tar.gz` should be available in ~/.ddev/bin

### Actual Behavior

Slack: https://drupal.slack.com/archives/C5TQRQZRR/p1721753115234839

Something deleted ~/.ddev/mutagen-agents.tar.gz, and as a result mutagen couldn't function, although `mutagen --version` works as a test.

### Steps To Reproduce

Delete ~/.ddev/bin/mutagen-agents.tar.gz and `ddev start`

### Anything else?

Simple workaround is `rm -rf ~/.ddev/bin` and `ddev restart`

We should be able to fix this by just testing in DownloadMutagenIfNeeded to download if the tarball is missing, and perhaps validate at the end of DownloadMutagen() to make sure the tarball arrived.

davearch:
## The Issue

- #6421 

The app is unable to start without a mutagen-agents.tar.gz file in the global bin directory.

This PR adds a check to make sure the tarbell exists in the DownloadMutagenIfNeeded function, and if it does not, downloads mutagen. It also adds a test to test this functionality works.

Remove the tarbell file in the global bin directory e.g. rm ~/.ddev/bin/mutagen-agents.tar.gz and restart the app e.g. ddev restart

## Automated Testing Overview
Adds the TestDownloadMutagen test which has the same setup as TestMutagenSimple but removes the tarbell if it exists, then starts the app. Then it verifies that the tarbell has been downloaded. Then it removes it again and restarts the app, and finally verifies that the tarbell has been downloaded again.

### DIFF
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -582,16 +582,9 @@ func DownloadMutagenIfNeeded() error {
 	if err != nil {
 		return err
 	}
-	curVersion, err := version.GetLiveMutagenVersion()
-	if err != nil || curVersion != versionconstants.RequiredMutagenVersion {
-		err = DownloadMutagen()
-		if err != nil {
-			return err
-		}
-	}
-	// Also make sure the mutagen-agents.tar.gz is in place
 	agentsFile := filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz")
-	if !fileutil.FileExists(agentsFile) {
+	curVersion, err := version.GetLiveMutagenVersion()
+	if err != nil || curVersion != versionconstants.RequiredMutagenVersion || !fileutil.FileExists(agentsFile) {
 		err = DownloadMutagen()
 		if err != nil {
 			return err
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -589,6 +589,14 @@ func DownloadMutagenIfNeeded() error {
 			return err
 		}
 	}
+	// Also make sure the mutagen-agents.tar.gz is in place
+	agentsFile := filepath.Join(globalconfig.GetGlobalDdevDirLocation(), "mutagen-agents.tar.gz")
+	if !fileutil.FileExists(agentsFile) {
+		err = DownloadMutagen()
+		if err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
--- a/pkg/ddevapp/mutagen.go
+++ b/pkg/ddevapp/mutagen.go
@@ -590,7 +590,7 @@ func DownloadMutagenIfNeeded() error {
 		}
 	}
 	// Also make sure the mutagen-agents.tar.gz is in place
-	agentsFile := filepath.Join(globalconfig.GetGlobalDdevDirLocation(), "mutagen-agents.tar.gz")
+	agentsFile := filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz")
 	if !fileutil.FileExists(agentsFile) {
 		err = DownloadMutagen()
 		if err != nil {
--- a/pkg/ddevapp/mutagen_test.go
+++ b/pkg/ddevapp/mutagen_test.go
@@ -250,3 +250,48 @@ func TestMutagenConfigChange(t *testing.T) {
 
 	util.Debug("origSyncID=%s afterRestartSyncID=%s afterChangeSyncID=%s", origSyncID, afterRestartSyncID, afterChangeSyncID)
 }
+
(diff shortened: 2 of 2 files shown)


rfay:
Thanks!

stasadev (approved):
Thank you!
Tested, works as it should.

pkg/ddevapp/mutagen.go
```diff
@@ -589,6 +589,14 @@ func DownloadMutagenIfNeeded() error {
...
 		}
 	}
+	// Also make sure the mutagen-agents.tar.gz is in place
+	agentsFile := filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz")
+	if !fileutil.FileExists(agentsFile) {
+		err = DownloadMutagen()
+		if err != nil {
+			return err
+		}
+	}
```
rfay, line 599: This is great! How about doing this with one if statement (incorporating it in the "or" above? You can calculate the agentsFile before the if, and add the !FileExists to the OR?

pkg/ddevapp/mutagen_test.go
```diff
@@ -250,3 +250,48 @@ func TestMutagenConfigChange(t *testing.T) {
 
 	util.Debug("origSyncID=%s afterRestartSyncID=%s afterChangeSyncID=%s", origSyncID, afterRestartSyncID, afterChangeSyncID)
 }
+
+// Test app starts despite a missing mutagen-agents.tar.gz file
```
rfay, line 254: Could this be done at the end of TestMutagenSimple and require less setup that way?

rfay (changes requested):
So impressive, and passed tests right away. I think it could be simplified a bit as mentioned, but thanks for taking this on and solving it, and it will be successful!

pkg/ddevapp/mutagen.go
davearch: Yes, makes sense!

pkg/ddevapp/mutagen_test.go
davearch: Yes, probably lol

pkg/ddevapp/mutagen.go
```diff
@@ -582,8 +582,9 @@ func DownloadMutagenIfNeeded() error {
 	if err != nil {
 		return err
 	}
+	agentsFile := filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz")
 	curVersion, err := version.GetLiveMutagenVersion()
-	if err != nil || curVersion != versionconstants.RequiredMutagenVersion {
+	if err != nil || curVersion != versionconstants.RequiredMutagenVersion || !fileutil.FileExists(agentsFile) {
```
rfay, line 587: Thanks, this is perfect.

pkg/ddevapp/mutagen_test.go
```diff
@@ -150,6 +150,26 @@ func TestMutagenSimple(t *testing.T) {
...
+	require.NoError(t, err)
+	// Verify that the mutagen-agents.tar.gz file was downloaded
+	assert.FileExists(filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz"))
+	// Remove the mutagen-agents.tar.gz file again and restart the app
+	err = os.Remove(filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz"))
+	require.NoError(t, err)
+	err = app.Restart()
+	require.NoError(t, err)
+	// Verify again that the mutagen-agents.tar.gz file was downloaded
+	assert.FileExists(filepath.Join(globalconfig.GetDDEVBinDir(), "mutagen-agents.tar.gz"))
```
rfay, line 171: Why do this a second time? I know you had a reason.

rfay (approved):
Looking great, thanks! Just had the one question.

pkg/ddevapp/mutagen_test.go
davearch: I wanted to specifically check that stopping the app and starting again (the restart) would not cause any issues. Mostly because in the stop method there are some mutagen calls (like SyncAndPauseMutagenSession and TerminateMutagenSync). This _may_ not be beneficial because I am not 100% on the details of mutagen but I wanted to sanity check it anyway. I can remove it if you think it is overkill. Probably that last FileExists check might be overkill

PR #6433, opened 2024-07-26
feat: add `ddev_version_constraint` for add-ons, remove `#ddev-nodisplay`, fixes #5969

### LINKED ISSUE #5969: docs: Remove docs and test for `#ddev-nodisplay`

_No response_

### Expected Behavior

I searched the code and found one test but I think all this stanza behavior was removed from ddev, not sure if by accident of on purpose. So if it was removed on purpose, probably worth removing it altogether and fixing https://github.com/ddev/ddev-addon-template/blob/main/install.yaml

project-wide search of #ddev-nodisplay

```
Searching 5610 files for "ddev-nodisplay"

~/Sites/work/ddev/cmd/ddev/cmd/testdata/TestCmdGetComplex/recipe/install.yaml:
    3  pre_install_actions:
    4  - |
    5:   #ddev-nodisplay
    6    echo "This action should not have any output"
    7  - |

~/Sites/work/ddev/docs/content/users/extend/additional-services.md:
  102  In any stanza of `pre_install_actions` and `post_install_actions` you can:
  103  
  104: * Use `#ddev-nodisplay` on a line to suppress any output.
  105  * Use `#ddev-description:<some description of what stanza is doing>` to instruct DDEV to output a description of the action it's taking.
  106  

2 matches across 2 files
```

### Actual Behavior

n/a

### Steps To Reproduce

_No response_

### Anything else?

_No response_

stasadev:
## The Issue

- https://github.com/ddev/ddev-xhgui/pull/31
- #5969

The current approach when you need to add a version constraint to an add-on involves using `ddev debug capabilities`, which is not obvious.

This isn't good because capabilities aren't added for each DDEV release and you can't add a version constraint for some minor release.

Adds `ddev_version_constraint` to `install.yaml` for add-ons that allows you to check the semver constraint.

TODO:
- [x] update https://github.com/ddev/ddev-addon-template/blob/8d7dac3113074fbf6e6b37696a24d837fa323056/install.yaml#L24-L26.
- https://github.com/ddev/ddev-addon-template/pull/55
- [x] update https://ddev.com/blog/advanced-add-on-contributor-training/
- https://github.com/ddev/ddev.com/pull/235

```
git clone https://github.com/ddev/ddev-phpmyadmin
echo "ddev_version_constraint: '>= v1.23.4'" >> ddev-phpmyadmin/install.yaml
ddev get ddev-phpmyadmin # should show error
sed -i 's/v1.23.4/v1.23.3/' ddev-phpmyadmin/install.yaml # lower the constraint
ddev get ddev-phpmyadmin # should be installed
```

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -127,7 +127,7 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 You can configure a [version constraint](https://github.com/Masterminds/semver#checking-version-constraints) for DDEV that will be validated against the running DDEV executable and prevent `ddev start` from running if it doesn't validate. For example:
 
 ```yaml
-ddev_version_constraint: '>=v1.23.0-alpha1'
+ddev_version_constraint: '>= v1.23.0-alpha1'
 ```
 
 This is only supported with DDEV versions above v1.22.4; older DDEV versions will ignore this setting.
--- a/docs/content/users/extend/additional-services.md
+++ b/docs/content/users/extend/additional-services.md
@@ -133,7 +133,7 @@ Anyone can create an add-on for `ddev get`. See [this screencast](https://www.yo
 
 The `install.yaml` is a simple YAML file with a few main sections:
 
-* `name`: The add-on name. This will be used in `ddev get --remove` and similar commands. If the repository name is `ddev-solr`, for example, a good `name` will be `solr`. 
+* `name`: The add-on name. This will be used in `ddev get --remove` and similar commands. If the repository name is `ddev-solr`, for example, a good `name` will be `solr`.
 * `pre_install_actions`: an array of Bash statements or scripts to be executed before `project_files` are installed. The actions are executed in the context of the target project’s root directory.
 * `project_files`: an array of files or directories to be copied from the add-on into the target project’s `.ddev` directory.
 * `global_files`: is an array of files or directories to be copied from the add-on into the target system’s global `.ddev` directory (`~/.ddev/`).
--- a/docs/content/users/extend/additional-services.md
+++ b/docs/content/users/extend/additional-services.md
@@ -138,9 +138,11 @@ The `install.yaml` is a simple YAML file with a few main sections:
 * `project_files`: an array of files or directories to be copied from the add-on into the target project’s `.ddev` directory.
 * `global_files`: is an array of files or directories to be copied from the add-on into the target system’s global `.ddev` directory (`~/.ddev/`).
 * `ddev_version_constraint` (available with DDEV v1.23.4+): a [version constraint](https://github.com/Masterminds/semver#checking-version-constraints) for DDEV that will be validated against the running DDEV executable and prevent add-on from being installed if it doesn't validate. For example:
+
     ```yaml
     ddev_version_constraint: '>= v1.23.4'
     ```
+
 * `dependencies`: an array of add-ons that this add-on depends on.
 * `post_install_actions`: an array of Bash statements or scripts to be executed after `project_files` and `global_files` are installed. The actions are executed in the context of the target project’s root directory.
 * `removal_actions`: an array of Bash statements or scripts to be executed when the add-on is being removed with `ddev get --remove`.
--- a/docs/content/users/extend/additional-services.md
+++ b/docs/content/users/extend/additional-services.md
@@ -133,7 +133,7 @@ Anyone can create an add-on for `ddev get`. See [this screencast](https://www.yo
 
 The `install.yaml` is a simple YAML file with a few main sections:
 
-* `name`: an add-on name.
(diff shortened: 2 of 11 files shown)


docs/content/users/configuration/config.md
```diff
@@ -127,7 +127,7 @@ Example: `dbimage_extra_packages: ["less"]` will add the `less` package when the
 You can configure a [version constraint](https://github.com/Masterminds/semver#checking-version-constraints) for DDEV that will be validated against the running DDEV executable and prevent `ddev start` from running if it doesn't validate. For example:
 
 ```yaml
-ddev_version_constraint: '>=v1.23.0-alpha1'
+ddev_version_constraint: '>= v1.23.0-alpha1'
```
rfay, line 130: Does this matter? Was the lack of space an error?

docs/content/users/extend/additional-services.md
```diff
@@ -133,9 +133,16 @@ Anyone can create an add-on for `ddev get`. See [this screencast](https://www.yo
 
 The `install.yaml` is a simple YAML file with a few main sections:
 
+* `name`: an add-on name.
```
rfay, line 136: While we're here, why don't we add a little detail. 
```suggestion
* `name`: The add-on name. This will be used in `ddev get --remove` and similar commands. If the repository name is `ddev-elasticsearch`, for example, a good `name` will be `solr`. 
```

pkg/ddevapp/config_test.go
```diff
@@ -534,14 +534,14 @@ func TestConfigValidate(t *testing.T) {
...
-	assert.Contains(err.Error(), "constraint that is not valid")
+	assert.Contains(err.Error(), "constraint is not valid")
 	app.DdevVersionConstraint = ""
 
 	versionconstants.DdevVersion = "v1.22.0"
 	app.DdevVersionConstraint = ">= 1.23"
 	err = app.ValidateConfig()
 	assert.Error(err)
-	assert.Contains(err.Error(), "project has a DDEV version constraint of '>= 1.23' and the version of DDEV you are using ('v1.22.0') does not meet the constraint")
+	assert.Contains(err.Error(), "your DDEV version 'v1.22.0' doesn't meet the constraint '>= 1.23'")
```
rfay, line 544: Interestingly, there has been long-running problems with version comparisons that don't have patch version. I guess gien this test maybe that's resolved. So I would have always used `v1.23.0`.

pkg/ddevapp/config_test.go
```diff
@@ -564,7 +564,7 @@ func TestConfigValidate(t *testing.T) {
 	app.DdevVersionConstraint = ">= 1.23"
 	err = app.ValidateConfig()
 	assert.Error(err)
-	assert.Contains(err.Error(), "project has a DDEV version constraint of '>= 1.23' and the version of DDEV you are using ('v1.22.3-11-g8baef014e') does not meet the constraint")
+	assert.Contains(err.Error(), "your DDEV version 'v1.22.3-11-g8baef014e' doesn't meet the constraint '>= 1.23'")
```
rfay, line 567: This set of comparisons would be easier to create and manage with a test array and t.Run(). I've recently started using that more for repetitive things in tests and like it quite a lot. For example, https://github.com/ddev/ddev/blob/c2aca52a18687e678086dd232573cf51914dba56/pkg/ddevapp/config_test.go#L1545-L1568

rfay (approved):
Minor suggestions and comments. Thanks! It's perfect. 

We should probably add a schema for install.yaml, so opened 
* https://github.com/ddev/ddev/issues/6435


I manually tested with a selection of versions and it worked flawlessly. 

A note: `make VERSION=v1.25.0` will create a DDEV version that thinks it's v1.25.0

docs/content/users/configuration/config.md
stasadev: It doesn't matter, I just noticed we have this space everywhere else, so for consistency and better readability I added it here.

pkg/ddevapp/config_test.go
stasadev: That's a good point.

rfay:
Can pull when you're ready. There are a couple of checkmarks still listed.

pkg/ddevapp/config_test.go
stasadev: Done.

pkg/ddevapp/config_test.go
```diff
@@ -531,79 +531,40 @@ func TestConfigValidate(t *testing.T) {
...
-	app.DdevVersionConstraint = ""
-	versionconstants.DdevVersion = ddevVersion
-
-	versionconstants.DdevVersion = "v1.22.3-beta1"
-	app.DdevVersionConstraint = ">= v1.22.3-alpha3"
-	err = app.ValidateConfig()
-	assert.NoError(err)
-	app.DdevVersionConstraint = ""
-	versionconstants.DdevVersion = ddevVersion
+	testCases := []struct {
```
rfay, line 534: Very nice, thanks!

PR #6439, opened 2024-07-29
docs: improve placement of .env for environment variables

rfay:
## The Issue

Probably the best way these days to provide environment variables to containers/services is to use the `.ddev/.env` file.

* Move the Environment variable section to the front page the page it's on
* Put the `.env` discussion first
* Push up the search preference value

I still don't get a good hit in search except with "variables", "environment variables" doesn't do much.

Rendered version is at https://ddev--6439.org.readthedocs.build/en/6439/users/extend/customization-extendibility/#environment-variables-for-containers-and-services

### DIFF
--- a/docs/content/users/extend/customization-extendibility.md
+++ b/docs/content/users/extend/customization-extendibility.md
@@ -1,11 +1,68 @@
 ---
 search:
-  boost: 2
+  boost: 3
 ---
 
 # Extending and Customizing Environments
 
 DDEV provides several ways to customize and extend project environments.
+
+## Environment Variables for Containers and Services
+
+You can set custom environment variables in several places:
+
+1. An optional, project-level `.ddev/.env` file provides environment variables to all DDEV containers, including any additional services or add-ons. It can look something like this:
+
+    ```
+    MY_ENV_VAR='someval'
+    MY_OTHER_ENV_VAR='someotherval'
+    ```
+
+2. The global `web_environment` setting in `.ddev/global_config.yaml`.
+
+    ```yaml
+    web_environment:
+    - MY_ENV_VAR=someval
+    - MY_OTHER_ENV_VAR=someotherval
+    ```
+
+3. The project’s [`web_environment`](../configuration/config.md#web_environment) setting in `.ddev/config.yaml` or `.ddev/config.*.yaml`:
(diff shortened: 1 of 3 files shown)


docs/content/users/extend/customization-extendibility.md
```diff
@@ -1,12 +1,69 @@
...
+    ```
+
+If you’d rather use the CLI to set the project or global `web_environment` value, you can use the [`ddev config`](../usage/commands.md#config) command:
+
+```sh
+# Set MY_ENV_VAR for the project
+ddev config --web-environment-add="MY_ENV_VAR=someval"
+
+# Set MY_ENV_VAR globally
+ddev config global --web-environment-add="MY_ENV_VAR=someval
```
stasadev, line 44:
```suggestion
ddev config global --web-environment-add="MY_ENV_VAR=someval"
```

docs/content/users/extend/customization-extendibility.md
```diff
@@ -1,12 +1,69 @@
...
+```
+
+You can use the `--web-environment` flag to overwrite existing values rather than adding them.
+
+!!!warning "Don’t check in sensitive values!"
+Sensitive variables like API keys should not be checked in with your project. Typically you might use an `.env` file and _not_ check that in, but offer `.env.example` with expected keys that don’t have values. Some use global configuration for sensitive values, as that’s not normally checked in either.
```
stasadev, line 50:
```suggestion
	Sensitive variables like API keys should not be checked in with your project. Typically you might use an `.env` file and _not_ check that in, but offer `.env.example` with expected keys that don’t have values. Some use global configuration for sensitive values, as that’s not normally checked in either.
```

stasadev: `.env.example` is ignored by `.ddev/.gitignore`.

We could mention how to add it to git here.

stasadev (approved):
Looks good to me, only small suggestions.

docs/content/users/extend/customization-extendibility.md
rfay: Although this is something that people stumble on regularly, I'm always torn about how much we should teach about how many things.

Issue #6380, opened 2024-07-06
Installation of Magento 2 on Linux failed due to wrong command when following docs


InteractiveWebdesign:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


WARNING: Project should usually be in a subdirectory of the user's home directory.
Instead it's in /var/www/magento2



======== Output file will be in /tmp/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/bin/ddev
/bin/ddev

These config files were loaded for project magento2: [/var/www/magento2/.ddev/config.yaml]
name: magento2
type: magento2
docroot: pub
php_version: 8.2
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.23.3
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.6}
upload_dirs: [media]
project_tld: loc
use_dns_when_possible: true
composer_version: 2
disable_settings_management: true
nodejs_version: 20
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
┌───────────────┬─────────┬─────────────────────────┬───────────────────────────┐
│ ADD-ON        │ VERSION │ REPOSITORY              │ DATE INSTALLED            │
├───────────────┼─────────┼─────────────────────────┼───────────────────────────┤
│ elasticsearch │ v0.3.2  │ ddev/ddev-elasticsearch │ 2024-07-06T15:38:28+02:00 │
└───────────────┴─────────┴─────────────────────────┴───────────────────────────┘



======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /var/www/magento2
.ddev/mutagen/mutagen.yml not found


======== Creating dummy project named tryddevproject-2839 in ../tryddevproject-2839 ========
Creating a new DDEV project config in the current directory (/var/www/tryddevproject-2839)
Once completed, your configuration will be written to /var/www/tryddevproject-2839/.ddev/config.yaml

[...]

PR #5932, opened 2024-03-04
docs: Improve magento2 quickstart and remember composer auth

rfay:
## The Issue

* The Magento2 install forgets the composer auth for the private repo
* People always forget what to do if the directory name is different
* Backdrop tabs are broken by messed-up indentation

## How This PR Solves The Issue

Make it a bit more resilient, and capture the auth.json where it will persist.

## Manual Testing Instructions

Review at https://ddev--5932.org.readthedocs.build/en/5932/users/quickstart/

Try it out

PR #6442, opened 2024-07-29
docs: store magento2 credentials in the global config, for #5932, for #6380

stasadev:
## The Issue

- #5932

The documentation says https://ddev.readthedocs.io/en/stable/users/quickstart/#magento

> Note that you can install the Adobe/Magento composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to find them again.

But the current quickstart doesn't check composer's global configuration for this auth, resulting in a situation where you have to add these credentials each time for each project.

Stores magento2 credentials in the global config `~/.ddev/homeadditions/.composer/auth.json` (and respects `$XDG_CONFIG_HOME`)

This is one-time action. If you run it again, it will simply check if your credentials are correct.

Review at https://ddev--6442.org.readthedocs.build/en/6442/users/quickstart/#magento

Try it out.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -418,13 +418,13 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     Normal details of a Composer build for Magento 2 are on the [Magento 2 site](https://experienceleague.adobe.com/docs/commerce-operations/installation-guide/composer.html). You must have a public and private key to install from Magento’s repository. When prompted for “username” and “password” in `composer create`, it’s asking for your public key as "username" and private key as "password".
 
     !!!tip "Store Adobe/Magento Composer credentials in the global DDEV config"
-        If you already have Composer installed on you system you can reuse the `auth.json` by making a symlink. See [In-Container Home Directory and Shell Configuration](extend/in-container-configuration.md):
+        If you have Composer installed on your workstation and have an `auth.json` you can reuse the `auth.json` by making a symlink. See [In-Container Home Directory and Shell Configuration](extend/in-container-configuration.md):
 
         ```
         mkdir -p ~/.ddev/homeadditions/.composer && ln -s ~/.composer/auth.json ~/.ddev/homeadditions/.composer/auth.json
         ```
 
-        Alternatively, you can install the Adobe/Magento Composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to find them again (see below):
+        Alternately, you can install the Adobe/Magento Composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to enter them again (see below):
     
         ??? "Script to store Adobe/Magento Composer credentials (click me)"
             ```bash
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -421,20 +421,14 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
 
     ```bash
     mkdir my-magento2-site && cd my-magento2-site
-    ddev config --project-type=magento2 --docroot=pub --disable-settings-management \
-    --upload-dirs=media
+    ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
+
+    # a one-time action to save your Magento credentials in the global DDEV config
+    # enter your username/password and agree to store your credentials
+    ddev_dir="$(ddev version -j | docker run -i --rm ddev/ddev-utilities jq -r ".raw.\"global-ddev-dir\" | select (.!=null) // \"$HOME/.ddev\"" 2>/dev/null)" && mkdir -p $ddev_dir/homeadditions/.composer && docker run -it --rm -v "$ddev_dir/homeadditions/.composer:/composer" --workdir=/composer -e COMPOSER_HOME=/composer --user $(id -u):$(id -g) $(ddev version -j | docker run -i --rm ddev/ddev-utilities jq -r ".raw.web | select (.!=null)" 2>/dev/null) bash -c "cd /tmp && composer create --repository https://repo.magento.com/ magento/project-community-edition --no-install"
 
     ddev get ddev/ddev-elasticsearch
     ddev start
-
-    # if the Magento auth is not found in the Composer global config, save it at the project level
-    if ! ddev exec "composer config --list --global 2>/dev/null | grep -q repo.magento.com" 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


rfay:
Just a note that... the config could also be in project .ddev/homeadditions/.composer. So many things.

stasadev:
> Just a note that... the config could also be in project .ddev/homeadditions/.composer. So many things.

Fixed by adding another check.

docs/content/users/quickstart.md
```diff
@@ -427,8 +427,8 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     ddev get ddev/ddev-elasticsearch
     ddev start
 
-    # if the Magento auth is not found in the Composer global config, save it at the project level
-    if ! ddev exec "composer config --list --global 2>/dev/null | grep -q repo.magento.com" 2>/dev/null; then
+    # if the Magento auth is not found in the Composer config, save it at the project level
+    if ! ddev exec "composer config --list 2>/dev/null | grep -q repo.magento.com || composer config --list --global 2>/dev/null | grep -q repo.magento.com" 2>/dev/null; then
```
rfay, line 431: I might be wrong, but I don't think that `composer config --list --global` is relevant here. The difference is between *DDEV*'s global and local homeadditions, both of which land in the *user* directory inside the project, `~/.composer/auth.json` - So maybe my comment was naive in the first place. Anyway, nothing we do makes it into composer's global config.

stasadev: > So maybe my comment was naive in the first place.

Your suggestion was correct because someone could override `COMPOSER_HOME` in the DDEV's global config.

> both of which land in the user directory inside the project, `~/.composer/auth.json`

Not both, when you have `COMPOSER_HOME` set, it will use a different path.

> I don't think that `composer config --list --global` is relevant here.

It's relevant because it runs after `ddev start`, so `~/.ddev/homeadditions/.composer/auth.json` is in `~/.composer/auth.json` at this point.

It just needs some manual testing for a better understanding.

stasadev: I reverted changes for `composer config --list` as it only shows the config for `composer.json` in the current directory.

On the other hand, `composer config --list --global` shows the config you have in `~/.composer/auth.json` or `COMPOSER_HOME` inside the container.

rfay: I was confused, of course. `global` config in composer is (normally) ~/.composer, and regular is in the project composer.json.

rfay (changes requested):
I don't think this is a very good idea in general.

*  I don't much like the added complexity, or the fact that we're doing this for them.  
* It still requires entering the credentials twice
* The auth.json gets committed via git (not sure why the gitignore was added here)
* Changing COMPOSER_HOME means that none of the cached packages are available, so a full download of everything is required.

Just a note, it's seems odd that we have that one 
> ddev php bin/magento info:adminuri

at the end. It should just be `ddev magento info:adminuri` right?

docs/content/users/quickstart.md
```diff
@@ -422,13 +422,21 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
     ddev start
-    ddev composer create --repository=https://repo.magento.com/ magento/project-community-edition
+
+    # if the Magento auth is not found in the Composer global config, save it at the project level
+    if ! ddev exec "composer config --list --global 2>/dev/null | grep -q repo.magento.com" 2>/dev/null; then
+        ddev config --web-environment-add=COMPOSER_HOME="/var/www/html/.ddev/homeadditions/.composer"
+        mkdir -p .ddev/homeadditions/.composer
+        echo "*\n\!.gitignore" >.ddev/homeadditions/.composer/.gitignore
+        ddev restart
+    fi
```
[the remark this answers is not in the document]
stasadev: I didn't test `echo "*\n\!.gitignore"` with bash 😅
It doesn't work as expected there, but it is fine with zsh, I will fix it.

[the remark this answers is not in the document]
stasadev: > was it supposed to gitignore the auth.json?

Yes, it was supposed to ignore everything in that folder except `.gitignore`, but the resulting file in bash looked like this:

```
*\n\!.gitignore
```

instead of

```
*
!.gitignore
```

stasadev:
For this I added a very long bash line 🙂

It stores magento2 credentials in the global config ~/.ddev/homeadditions/.composer/auth.json (and respects `$XDG_CONFIG_HOME`)

This is one-time action. If you run it again, it will simply check if your credentials are correct.

docs/content/users/quickstart.md
```diff
@@ -417,18 +417,40 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
 
     Normal details of a Composer build for Magento 2 are on the [Magento 2 site](https://experienceleague.adobe.com/docs/commerce-operations/installation-guide/composer.html). You must have a public and private key to install from Magento’s repository. When prompted for “username” and “password” in `composer create`, it’s asking for your public key as "username" and private key as "password".
 
-    Note that you can install the Adobe/Magento composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to find them again. See [In-Container Home Directory and Shell Configuration](extend/in-container-configuration.md).
+    !!!tip "Store Adobe/Magento Composer credentials in the global DDEV config"
+        If you already have Composer installed on you system you can reuse the `auth.json` by making a symlink. See [In-Container Home Directory and Shell Configuration](extend/in-container-configuration.md):
```
rfay, line 421:
```suggestion
        If you have Composer installed on your workstation and have an `auth.json` you can reuse the `auth.json` by making a symlink. See [In-Container Home Directory and Shell Configuration](extend/in-container-configuration.md):
```

docs/content/users/quickstart.md
```diff
@@ -417,18 +417,40 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+
+        ```
+        mkdir -p ~/.ddev/homeadditions/.composer && ln -s ~/.composer/auth.json ~/.ddev/homeadditions/.composer/auth.json
+        ```
+
+        Alternatively, you can install the Adobe/Magento Composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to find them again (see below):
```
rfay, line 427:
```suggestion
        Alternately, you can install the Adobe/Magento Composer credentials in your global `~/.ddev/homeadditions/.composer/auth.json` and never have to enter them again (see below):
```

rfay (changes requested):
Minor suggestions. That's some script! Thanks for the work on this. I wish Adobe weren't so weird.

rfay (approved):
The new approach to the password is really nice and works perfectly. Thanks!

PR #6447, opened 2024-07-30
docs: show actual running context in `install.yaml` for add-ons

stasadev:
## The Issue

While working on

- https://github.com/ddev/ddev-varnish/pull/28

I noticed that different add-on actions are performed in different contexts, which is confusing.

- `pre_install_actions` in `$DDEV_APPROOT`
- `post_install_actions` in `$DDEV_APPROOT/.ddev`
- `removal_actions` in `$DDEV_APPROOT/.ddev`

Related change:

- https://github.com/ddev/ddev-addon-template/pull/54

Fixes documentation.

```
mkdir miniaddon && echo "name: miniaddon\npre_install_actions:\n- pwd\npost_install_actions:\n- pwd\nremoval_actions:\n- pwd" > miniaddon/install.yaml
# the output should show project root for pre_install_actions and .ddev for post_install_actions
ddev get miniaddon
# the output should show  .ddev for removal_actions
ddev get --remove miniaddon
```

### DIFF
--- a/cmd/ddev/cmd/get.go
+++ b/cmd/ddev/cmd/get.go
@@ -267,6 +267,21 @@ ddev get --remove ddev-someaddonname
 				}
 			}
 		}
+
+		origDir, _ := os.Getwd()
+
+		defer func() {
+			err = os.Chdir(origDir)
+			if err != nil {
+				util.Failed("Unable to chdir to %v: %v", origDir, err)
+			}
+		}()
+
+		err = os.Chdir(app.GetConfigPath(""))
+		if err != nil {
+			util.Failed("Unable to chdir to %v: %v", app.GetConfigPath(""), err)
+		}
+
 		if len(s.PreInstallActions) > 0 {
 			util.Success("\nExecuting pre-install actions:")
 		}
@@ -328,19 +343,6 @@ ddev get --remove ddev-someaddonname
 			} else {
 				util.Warning("NOT overwriting %s. The #ddev-generated signature was not found in the file, so it will not be overwritten. You can remove the file and use DDEV get again if you want it to be replaced: %v", dest, err)
 			}
-		}
-		origDir, _ := os.Getwd()
-
-		defer func() {
-			err = os.Chdir(origDir)
-			if err != nil {
-				util.Failed("Unable to chdir to %v: %v", origDir, err)
-			}
-		}()
-
-		err = os.Chdir(app.GetConfigPath(""))
-		if err != nil {
-			util.Failed("Unable to chdir to %v: %v", app.GetConfigPath(""), err)
 		}
 
 		if len(s.PostInstallActions) > 0 {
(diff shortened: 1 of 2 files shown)


stasadev:
This PR normalizes the running context for everything except `yaml_read_files` which runs in `$DDEV_APPROOT`, I wonder if we should change it to `$DDEV_APPROOT/.ddev`

rfay:
> I wonder if we should change it to $DDEV_APPROOT/.ddev

Agreed that we should. It's not unlikely that ddev -platformsh is the only user. 

IMO it's quite naive that it doesn't already use $DDEV_APPROOT to explicitly name the directory.

https://github.com/ddev/ddev-platformsh/blob/bb7365e30ae68797602dd0f648bf16bb46cd62b3/install.yaml#L330-L333

rfay:
Looks like TestCmdGet has some problems. Not sure why I see those only on macOS.

rfay (changes requested):
This looks good to me. As you suggested, we should
* do yaml_read_files as well
* Update the ddev-addon-template (including explicitly using $DDEV_APPROOT in there)
* Fix ddev-platformsh proactively
* Carefully run the monitoring script after this goes in and fix what it breaks.

rfay:
* https://github.com/ddev/ddev-platformsh/pull/126

rfay:
`yaml_read_files` already puts together the full path, so adding ${DDEV_APPROOT} on the front of those files ends up with 
> /Users/rfay/tmp/test-platformsh/Users/rfay/tmp/test-platformsh/.platform.app.yaml: unable to read file /Users/rfay/tmp/test-platformsh/Users/rfay/tmp/test-platformsh/.platform.app.yaml (open /Users/rfay/tmp/test-platformsh/Users/rfay/tmp/test-platformsh/.platform.app.yaml: no such file or directory)

rfay:
I guess maybe `yaml_read_files` code should check to see if path is rooted?

stasadev:
Yes, it can be done in a similar way to:

https://github.com/ddev/ddev/blob/c2aca52a18687e678086dd232573cf51914dba56/cmd/ddev/cmd/auth-ssh.go#L40-L45

rfay:
This also makes `ddev get` on relative directories fail.

For example, 

`ddev get ../ddev-platformsh` fails
`ddev get ${PWD}/../ddev-platformsh` succeeds

stasadev:
It might be better to leave everything as is, but change the docs properly with `ddev-addon-template` as they are incorrect:

https://ddev.readthedocs.io/en/latest/users/extend/additional-services/#sections-and-features-of-ddev-get-add-on-installyaml

rfay:
Shucks. Well, we tried!

stasadev:
Related change:

- https://github.com/ddev/ddev-addon-template/pull/54

rfay (approved):
Thanks!

Issue #6451, opened 2024-08-01
Show users warnings from stderr when running updates for composer, sqlite, etc


stasadev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Sometimes `web` container fails to update some components, like:
- `composer self-update`
- sqlite for Drupal 11
- ...

We silently continue here with `|| true`, so the container still runs, but users don't know about the errors.

### Describe your solution

Ideally it should be a script, that can be used in `/start.sh` or Dockerfile.

```dockerfile
RUN log-stderr.sh "composer self-update"
```

It should store `stderr` in some file in `/tmp`.

Something similar is done for `n install` in:
- #6420

But we should generalize it as much as possible.

rfay:
Wow!

stasadev:
It is important to handle situation when a failure occurs when executing `RUN` instruction inside a `Dockerfile`, for example, `composer self-update`, and it is cached by Docker:

1. add a `dirty` label to the `ddev-webserver-built` image (if possible), and run rebuild on the next start/restart? And do not break the offline mode.
2. or ignore it and just show the additional warning that suggests `ddev debug refresh` if something failed in the `Dockerfile`.

rfay:
I think image *labels* might be the best technique for a "dirty" marker. I guess it can be tricky to add label after we're done, but maybe this approach will work for us, didn't study it carefully, https://stackoverflow.com/questions/44828219/how-to-add-docker-label-after-image-is-made

PR #6455, opened 2024-08-02
feat: add `log-stderr.sh` command, for #6451, for #6420

stasadev:
## The Issue

- #6451
- #6420

- Adds a new bash command `log-stderr.sh` that stores `stderr` output. (I've only used it for `n install` so far, I will add it to other commands later, after some testing and probably after the next DDEV release.)
- Adds `export DDEV_SKIP_NODEJS_TEST=true` for Buildkite tests.
- Changes `n` installation technique according to upstream recommendations.
- Optimizes error output for `n install` (removes `--offline` logging).

Look at the newer error output for:

```
$ ddev config --nodejs-version auto # when you don't have .nvmrc file or another file
...
Warning: Command 'n install auto' run as 'stas' failed with exit code 2:
  Error: no file found for auto version (.n-node-version, .node-version, .nvmrc, or package.json)

$ ddev config --nodejs-version foobar # invalid version
...
Warning: Command 'n install foobar' run as 'stas' failed with exit code 2:
  Error: invalid version 'foobar'

$ ddev config --nodejs-version 6 # without internet
...
Warning: Command 'n install 6' run as 'stas' failed with exit code 2:
curl: (6) Could not resolve host: nodejs.org

  Error: failed to download version index (https://nodejs.org/dist/index.tab)
```

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/log-stderr.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/log-stderr.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# This script is used to run a command with optional timeout
+# and save stderr to /tmp/ddev-log-stderr-${whoami}.txt
+set +eu
+
+# Function to display usage information
+usage() {
+  echo "Usage: $(basename "$0") [-t timeout] command"
+  echo "  -t, --timeout: Timeout in seconds (default: no timeout)"
+  exit 1
+}
+
+# Default timeout value
+timeout=0
+
+if whoami &>/dev/null; then
+  whoami=$(whoami)
+else
+  whoami=$(id -u)
+fi
+
+# Parse command line options
+while [[ "$#" -gt 0 ]]; do
+  case "$1" in
+    -t|--timeout)
+      timeout="$2"
+      shift 2
+      ;;
+    -*)
+      usage
+      ;;
+    *)
+      command=("$@")
+      break
+      ;;
+  esac
+done
+
+# If no command is provided, show usage
+if [ -z "${command[*]}" ]; then
+  usage
+fi
+
+tmp_error_file="/tmp/tmp-ddev-log-stderr-${whoami}-$(echo -n "${command[*]} $(date +%s)" | md5sum | awk '{print $1}').txt"
+
+# Run the command with timeout if specified
+if [ "${timeout}" -gt 0 ]; then
+  timeout "${timeout}" "${command[@]}" 2> >(tee -a "${tmp_error_file}" >&2)
+  exit_code=$?
+else
+  "${command[@]}" 2> >(tee -a "${tmp_error_file}" >&2)
+  exit_code=$?
+fi
+
+# Exit on success
+if [ "${exit_code}" -eq 0 ]; then
+  rm -f "${tmp_error_file}"
+  exit "${exit_code}"
+fi
+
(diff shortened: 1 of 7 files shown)


containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/log-stderr.sh
```diff
@@ -0,0 +1,116 @@
...
+[ "${debug}" = "true" ] && set -x
+
+# show stderr output
+if [ "${show}" = "true" ]; then
+  # find /tmp -maxdepth 1 -name 'ddev-log-stderr-*.txt': Searches for files matching the pattern.
+  # -printf "%C@ %p\n": Outputs the creation time (%C@) in seconds since the epoch followed by the file path (%p).
+  # sort -n: Sorts the output numerically (-n).
+  # awk '{print $2}': Extracts the file paths from the sorted output.
+  # xargs cat --squeeze-blank: Concatenates content by collapsing multiple blank lines into a single blank line.
+  find /tmp -maxdepth 1 -name 'ddev-log-stderr-*.txt' -printf "%C@ %p\n" | sort -n | awk '{print $2}' | xargs --no-run-if-empty cat --squeeze-blank
```
rfay, line 60: I would have expected a simpler `ls /tmp/ddev-log-stderr*` here, but there's nothing wrong with this.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/log-stderr.sh
```diff
@@ -0,0 +1,116 @@
...
+# If no command is provided, show usage
+if [ -z "${command[*]}" ]; then
+  usage
+fi
+
+if whoami &>/dev/null; then
+  whoami=$(whoami)
+else
+  whoami=$(id -u)
+fi
```
rfay, line 73: This is fine. Given that we have a predictable environment I would think just $(whoami) would be adequte, but it's still fine.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/n-install.sh
```diff
@@ -31,28 +31,29 @@ fi
 
 ln -sf "/mnt/ddev-global-cache/n_prefix/${HOSTNAME}" "${N_PREFIX}"
 
-# try normal install that also uses cache
+# try online install that also uses cache
 n_install_result=true
-timeout 30 n install "${N_INSTALL_VERSION}" 2> >(tee /tmp/n-install-stderr.txt >&2) || n_install_result=false
+log-stderr.sh --timeout 30 n install "${N_INSTALL_VERSION}" || n_install_result=false
```
rfay, line 36: I wonder if our scattered 30-second timeouts are a general mistake, or if we should figure out a way to parameterize them. No need to change anything right now, just food for thought.

rfay (approved):
Thanks for this!

I tried a bunch of different permutations, online and offline, and everything worked as expected.

I studied the new `log-stderr.sh` and like it. I'm glad we're going to be able to report these kind of things when we skip something due to internet trouble.

It's interesting that we switched from dynamic updates to ddev-router (traefik) into building our own just because of problems like this, but we're still doing it lots of place. composer was due to a request from a user. sqlite we do because of D11 and only there. I'm not sure there's anything we can do about node. 

However, with sqlite, we could stash the 3.45 version somewhere and only use it for D11. Or we could just use it everywhere and not worry, and it would get us around the problem of failure.

PR #6467, opened 2024-08-07
docs: add instruction how to brew upgrade ddev head

rpkoller:
## The Issue

so far i was using an alias for regularly updating my version of DDEV head `alias ddevup="ddev poweroff && brew unlink ddev && brew install --HEAD --fetch-head ddev"`. after a while it happened that the version of ddev was showing up on `brew outdated`. that was the symptom that within the next few upgrades running ddevup will fail with a permission issue where homebrew is unable to replace the old with the new ddev binary after the make step. i simply closely followed the instructions in the docs back then creating that alias. @rfay suggested dropping `--HEAD` and also using upgrade instead of brew install and going with this command `brew upgrade --fetch-head ddev` which worked fine so far. I've simply added one additional sentence to the section in the docs explaining the of someone waning to regularly test and use the latest version of DDEV HEAD.


docs/content/developers/building-contributing.md
```diff
@@ -9,7 +9,7 @@ search:
...
-* **Homebrew install HEAD**: On macOS and Linux, run `brew unlink ddev && brew install ddev/ddev/ddev --HEAD --fetch-HEAD` to get the latest DDEV commit, even if it’s unreleased. Since you’re building this on your own computer, it’s not signed or notarized, and you’ll get a notification that instrumentation doesn’t work, which is fine. If you’re using Linux/WSL2, you’ll likely need to install build-essential by running the following command: `sudo apt-get install -y build-essential`.
+* **Homebrew install HEAD**: On macOS and Linux, run `brew unlink ddev && brew install ddev/ddev/ddev --HEAD --fetch-HEAD` to get the latest DDEV commit, even if it’s unreleased. Since you’re building this on your own computer, it’s not signed or notarized, and you’ll get a notification that instrumentation doesn’t work, which is fine. If you’re using Linux/WSL2, you’ll likely need to install build-essential by running the following command: `sudo apt-get install -y build-essential`. In case you want to regularly test and use the latest version of DDEV HEAD run `brew unlink ddev && brew upgrade --fetch-head ddev` instead.
```
rfay, line 12:
```suggestion
* **Homebrew install HEAD**: On macOS and Linux, run `brew unlink ddev && brew install ddev/ddev/ddev --HEAD --fetch-HEAD` to get the latest DDEV commit, even if it’s unreleased. Since you’re building this on your own computer, it’s not signed or notarized, and you’ll get a notification that instrumentation doesn’t work, which is fine. If you’re using Linux/WSL2, you’ll likely need to install build-essential by running the following command: `sudo apt-get install -y build-essential`. In case you want to regularly test and use the latest version of DDEV HEAD run `brew upgrade --fetch-head ddev` instead.
```

The unlink isn't required is it? And won't this leave it unlinked?

rfay (changes requested):

docs/content/developers/building-contributing.md
rpkoller: i think i would have to test omitting the unlink when a new version of head is available. but i thought it is the saver approach and inline with the install variant. you unlink the old and then upgrade (while on upgrade ddev is linked again automatically)

rfay: There's a new version there now for you to test with :)

rpkoller: i've checked the version and the path of the ddev binary before and after and only ran the upgrade fetch head. it got properly upgraded. and on second thought with the regular upgrade of stable formulas you also dont have an unlink. and i also tested the start of a project after the upgrade. everything works. good call, i was simply too overcautious. committed your suggestion

rfay (approved):
If you like it, I like it, thanks!

PR #6469, opened 2024-08-07
Add back the command to renew LE certs.

jonpugh:
## The Issue

- #TBD

The documentation for casual hosting was missing instructions for renewing letsencrypt certificates. I googled it and turns out Google cached the old docs?

> You'll need to regularly renew the Let's Encrypt certificates. This is often done on a system reboot, but that may not be soon enough. A cron with the command docker exec ddev-router bash -c "certbot renew && nginx -s reload" will do the renewals.

Adds this paragraph to the letsencrypt notes.

### DIFF
--- a/docs/content/users/topics/hosting.md
+++ b/docs/content/users/topics/hosting.md
@@ -34,8 +34,6 @@ Here’s how to try it for yourself:
             caServer: "https://acme-staging-v02.api.letsencrypt.org/directory"
     ```
 
-You'll need to regularly renew the Let's Encrypt certificates. This is often done on a system reboot, but that may not be soon enough. A cron with the command docker exec ddev-router bash -c "certbot renew && nginx -s reload" will do the renewals.
-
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
This is no longer required, as traefik does the work.

jonpugh:
Ah that explains it. 

I can't remember why, but I needed to use `nginx-proxy` for my ddev site runner thing. 

I'm using it for production for www.thinkdrop.net!

rfay:
Have you set up the new traefik technique? I haven't yet done it on my little server.

rfay:
BTW, misspelling Persistant -> Persistent

Interested in hearing more about Operations Site Runner. Are you interested in doing a guest blog on ddev.com?

rfay:
The suggested commands won't work with traefik router and shouldn't be needed any more. We'll see!

PR #6471, opened 2024-08-08
docs: ddev debug test should suggest current stable [skip ci]

rfay:
## The Issue

After a support issue I was looking to make sure we bring up the "latest current version" thing.

This adds it to the front of `ddev debug test`, not sure that will help.

I'm going to leave this in draft for some time, if https://github.com/ddev/ddev/pull/6414 goes in there may be some adjusting to do here.


rfay:
I'm tired of waiting for this :) I guess we can bring it in soon.

PR #1368, opened 2019-01-07
Web container maintenance, fixes #1350, fixes #920, fixes #1337

rfay:
## The Problem/Issue/Bug:

We had some trivial web container maintenance in various issues.

* #1350 wants memcached added for php7.3, it's ready now
* #920 requests ghostscript as it's important for specific TYPO3 needs
* #1337 requests that the ~/.ssh directory have ownership that matches the running user.
* Add php-apcu package per [stack overflow answer](https://stackoverflow.com/a/54049172/215713), and add @mglaman's Drupal 8 settings.ddev.php improvement with `$settings['class_loader_auto_detect'] = FALSE;`

## Manual Testing Instructions:

* use the Circleci build or `webimage: drud/ddev-webserver:20190107_web_container_maint` in config.yaml and `ddev start`
* Verify that php7.3-memcached and apcu are in the container (php -i, admin/reports/status/php)
* Verify that ~/.ssh is owned by the container user.
* Run ghostscript (gs)

PR #6478, opened 2024-08-11
fix: drupal class_loader_auto_detect no longer needed

rfay:
## The Issue

Drupal #ddev slack https://drupal.slack.com/archives/C5TQRQZRR/p1723338555563719 points out that settings.ddev.php has

```
// Don't use Symfony's APCLoader. ddev includes APCu; Composer's APCu loader has
// better performance.
$settings['class_loader_auto_detect'] = FALSE;
```

but that comes from long ago. Apparently it's not needed in Drupal 9+, so can be removed. 

It was probably introduced in https://github.com/ddev/ddev/pull/1368 (https://github.com/ddev/ddev/pull/1368/commits/dd3b86188e13e14adce26c15ee35d82502fdcb81#diff-0196cec3f796f7d78f6e01b3d20405d825d0d874dd0aa9aff53930171d394217R161-R162) 

However, the need for it seems to have been removed in https://www.drupal.org/project/drupal/issues/3020296 (Drupal 9+)

* [Change record: Only Composer's classloader is used](https://www.drupal.org/node/3116384)
* 

Remove `$settings['class_loader_auto_detect'] = FALSE;` from settings.ddev.php in drupal9+

Try it out in Drupal project

### DIFF
--- a/pkg/ddevapp/drupal/drupal10/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal10/settings.ddev.php
@@ -32,10 +32,6 @@ $settings['skip_permissions_hardening'] = TRUE;
 // names. Additional host patterns can be added for custom configurations.
 $settings['trusted_host_patterns'] = ['.*'];
 
-// Don't use Symfony's APCLoader. ddev includes APCu; Composer's APCu loader has
-// better performance.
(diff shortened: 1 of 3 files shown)


weitzman:
LGTM.

Issue #6298, opened 2024-06-11
Deprecate and remove `ddev service` commands


rfay:
`ddev service` has been mostly replaced with add-on commands.

Deprecate and remove it.

stasadev:
For the context, how it's replaced by add-on commands:

Now you can do this:
```
ddev add-on remove redis
```

Instead of this:
```
ddev service disable redis
```

rfay:
Our approach for this will be:
* In v1.24: Change to a message saying "Use ddev add-on commands ..." and make a hidden command and remove from docs
* In v1.25: Remove it completely

PR #6482, opened 2024-08-14
refactor: Refactor `ddev get` to `ddev add-on` and into sub-commands, fixes #6146

### LINKED ISSUE #6146: Rename `ddev get` to `ddev addon` with sub-commands

`ddev get` isn't immediately obvious by the name what it does. What's more, `ddev get --list` and `ddev get --installed`, an especially `ddev get --remove` are all semantically a bit strange.

Having the command be `ddev addon` would be more immediately obvious what the command is for, and splitting the semantically different actions into sub commands means you don't have to worry about what happens if someone tries something weird like `ddev get --list --all --installed --remove my-addon` or just `ddev get --all`

"get" is also a very common word which makes it harder to search for, e.g. there's 19 results for "get" in the docs (only one is about this command) vs 3 for "addon", and searching "get" and "addon" in github issues similarly shows a lot more results for "get" which are just issues where that word happened to be used, not at all related to the concept of addons nor about the `get` command.

### Describe your solution

Add a new `ddev addon` command, with sub commands for each semantically different action (`get` to get an addon, `remove` to remove one, `list` to list them)
```bash
ddev addon get my-addon
ddev addon remove my-addon
ddev addon list --all
ddev addon list --installed
```

Then deprecate `ddev get`, and update it so all it does is call the appropriate `ddev addon` command. Eventually, remove `ddev get` in a future release.

I would suggest also hiding `ddev get` so it doesn't appear in documentation or in `ddev help`, `ddev --help`, etc. But that could be done later on closer to the time it gets removed, if there's any concern about people wondering where it went.
If it's not hidden, its description should be updated to indicate that it's deprecated and to use the new command instead.

#### Proposed structure of the new command

The `addon` command would have three sub-commands: `list`, `get`, and `remove`.

|0ld|new|
|---|---|
|`ddev get --list`|`ddev addon list`|
|`ddev get --list --all`|`ddev addon list --all`|
|`ddev get --installed`|`ddev addon list --installed`|
|`ddev get my-addon`|`ddev addon get my-addon`|
|`ddev get --remove my-addon`|`ddev addon remove my-addon`|

### Describe alternatives

The command could be called `add-on` instead of `addon` if that's preferable, since the documentation broadly refers to "add-ons". Personally I prefer `addon` because it's nicer to type - and I think it's obvious that the `addon` command is related to what the documentation broadly refers to as "add-ons". But either way is fine.

### Additional context

_No response_

GuySartorelli:
## The Issue

- #6146

`ddev get` isn't immediately obvious by the name what it does, and it seems to be doing three separate things (installing, removing, and listing add-ons).

Creates an umbrella command `ddev add-on`, which has a sub-command for each of the three responsibilities of the `ddev get` command:
1. `ddev add-on get`
2. `ddev add-on remove`
3. `ddev add-on list`

Appropriate aliases are added - e.g. `ddev add-on`, `ddev addon`, `ddev add-ons`, and `ddev addons` all work. This means people don't have to remember whether it's plural or not, nor whether it's hyphenated (I forgot both of those while I was testing this PR, so this makes it easy to use).

The existing `ddev get` command is turned into an alias for the various `ddev add-on` commands, so the code only exists in one place. It is also marked as deprecated, so that users get a message to use the new `ddev add-on` command instead.

In a future release, once people have gotten used to using `ddev add-on`, the `ddev get` command can be removed.

Rendered docs at https://ddev--6482.org.readthedocs.build/en/6482/users/extend/additional-services/ (and scattered around, look carefully)

Try `ddev add-on` on its own - it should just list usage info.

Try `ddev add-on` with flags and arguments. It should still just list usage info.

Try `ddev add-on get <add-on>` to install some add-ons.

Try `ddev add-on remove <add-on>` to remove some installed add-ons.

Try `ddev add-on list` with and without the various flags to list official, unnoficial, and installed add-ons.

Try the old `ddev get` command - that should still work exactly as expected, just with the additional output at the start saying to use the new command instead.

Check the usage info for each of the `ddev add-on` sub-commands is correct.

## Automated Testing Overview

I've duplicated the existing tests for `ddev get`. This means we are still testing `ddev get` to ensure that the alias works as expected, and we're also testing the new `ddev add-on` sub-commands.

This does mean the tests are in two places at once, so a change to the tests in one place needs to also be made in the other place. I couldn't find a clean way to alias the tests.

If you think it's better to just test `ddev add-on` with the existing test coverage, and instead just write some minimal tests for the `ddev get` alias, I'm happy to make that change.

## Release/Deployment Notes

Everything that used to work still works. Nothing will break as a result of this PR.

We should highlight that `ddev add-on` is the new best practice, and that `ddev get` will eventually be removed.

People should update any scripts they have that use `ddev get` to instead use the relevant `ddev add-on` sub-command.

People should update the instructions in their add-on readme files to use `ddev add-on get` instead of `ddev get`.


GuySartorelli:
Looks like there's a lot of lint failures that aren't related to the changes in this PR.

rpkoller:
this is fantastic! thank you so much! i've given it a spin. one thought i have. with `ddev addon`you get the help text shown. would it be reasonable to do the same for `ddev get`? at the moment you only get `You must specify an add-on to download` and you have to run `ddev addon get --help`. probably same for `ddev addon remove`. but there i get ```$> ddev addon remove
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x103158384]

goroutine 1 [running]:
github.com/ddev/ddev/cmd/ddev/cmd.init.func3(0x103950b00, {0x1039c9f80?, 0x0?, 0x0?})
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/addon-remove.go:41 +0x1c4
github.com/spf13/cobra.(*Command).execute(0x103950b00, {0x1039c9f80, 0x0, 0x0})
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:987 +0x828
github.com/spf13/cobra.(*Command).ExecuteC(0x103959ac0)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1039
github.com/ddev/ddev/cmd/ddev/cmd.Execute()
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/root.go:122 +0x24
main.main()
	/home/runner/work/ddev/ddev/cmd/ddev/main.go:27 +0xd0```

and i also get errors when i try to remove an previously installed addon with either ```$> ddev addon remove adminer
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x102f64384]

goroutine 1 [running]:
github.com/ddev/ddev/cmd/ddev/cmd.init.func3(0x10375cb00, {0x14000496970?, 0x1?, 0x1?})
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/addon-remove.go:41 +0x1c4
github.com/spf13/cobra.(*Command).execute(0x10375cb00, {0x14000496930, 0x1, 0x1})
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:987 +0x828
github.com/spf13/cobra.(*Command).ExecuteC(0x103765ac0)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1039
github.com/ddev/ddev/cmd/ddev/cmd.Execute()
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/root.go:122 +0x24
main.main()
	/home/runner/work/ddev/ddev/cmd/ddev/main.go:27 +0xd0``` 

or

```$> ddev addon remove ddev/ddev-adminer 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x1028c0384]

goroutine 1 [running]:
github.com/ddev/ddev/cmd/ddev/cmd.init.func3(0x1030b8b00, {0x140000a87d0?, 0x1?, 0x1?})
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/addon-remove.go:41 +0x1c4
github.com/spf13/cobra.(*Command).execute(0x1030b8b00, {0x140000a8790, 0x1, 0x1})
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:987 +0x828
github.com/spf13/cobra.(*Command).ExecuteC(0x1030c1ac0)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/work/ddev/ddev/vendor/github.com/spf13/cobra/command.go:1039
github.com/ddev/ddev/cmd/ddev/cmd.Execute()
	/home/runner/work/ddev/ddev/cmd/ddev/cmd/root.go:122 +0x24
main.main()
	/home/runner/work/ddev/ddev/cmd/ddev/main.go:27 +0xd0```

File: cmd/ddev/cmd/addon-list.go [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/google/go-github/v52/github"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/spf13/cobra"
+)
+
+// AddonListCmd is the "ddev add-on list" command
+var AddonListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List available or installed DDEV add-ons",
```
rpkoller, line 23: and one nitpick i havent looked at all the help texts yet. but list available addons is correct but i am unsure if "installed" addons are listed too. list just lists official addons and with the all flag also unofficial one but installed addons are not have any additional label in the outputted list?

stasadev:
Wow!

> Looks like there's a lot of lint failures that aren't related to the changes in this PR.

Thanks, I'll fix it today.

rfay:
Rebased

GuySartorelli:
> with `ddev addon` you get the help text shown. would it be reasonable to do the same for `ddev addon get`? at the moment you only get You must specify an add-on to download and you have to run `ddev addon get --help`. probably same for `ddev addon remove`

Good point - I think there's validation available via cobra to mark that arguments are required. I'll look into that tonight.

GuySartorelli:
@rpkoller I've added validation for the number of arguments passed to `ddev add-on get` and `ddev add-on remove` - which has the side effect of showing the help info if you don't pass any arguments.

I've also fixed the problem with `ddev add-on remove`. Can you please retest to confirm?

Fingers cross that fixes the broken CI tests too, but I'll give those a chance to rerun.

rpkoller:
yay looks good, I can confirm the following:  
- the runtime errors are fixed
- i am able to remove an addon based on the addon name as well as the repository name 
- just entering `addon remove` or `addon get` without any flag outputs the helptext now. but is it necessary to output also an error at all? the error could easily read over and be missed and by just entering the command without any flag the user notices anyway that the command had no effect except showing the help text which provides the necessary actual usage?  

```
$> ddev addon remove
Error: accepts 1 arg(s), received 0
Usage:
  ddev add-on remove [flags]

Examples:
ddev add-on remove someaddonname,
ddev add-on remove someowner/ddev-someaddonname,
ddev add-on remove ddev-someaddonname


Flags:
  -h, --help      help for remove
  -v, --verbose   Extended/verbose output

Global Flags:
      --skip-hooks    If true, any hook normally run by the command will be skipped.
```

and strictly speaking `ddev addon`also needs a command  and a flag?

GuySartorelli:
> is it necessary to output also an error at all? 

I suppose it's not strictly necessary, but I think it makes sense to use the validation that cobra provides our of the box instead of rolling our own way to display help info when no Argos are provided. 

> and strictly speaking ddev addonalso needs a command and a flag?

I'm not sure what you mean

GuySartorelli:
Will look into the remaining ci failures tomorrow when I get a chance.

rpkoller:
> > and strictly speaking ddev addonalso needs a command and a flag?
> 
> I'm not sure what you mean

I meant if you enter `ddev addon` it is also missing a command and flag. so shouldnt there be an error there as well then? with `ddev addon` no error while for example `ddev addon remove` returns an error. it is no way a problem. it just felt inconsistent and "sort of" unexpected. i am fine either way, i just noticed, and wanted to note it.

GuySartorelli:
> I meant if you enter ddev addon it is also missing a command and flag. so shouldnt there be an error there as well then?

There are no flags or arguments for that command, so all it can do it display usage info, which tells you about the sub commands. But it is a bit subjective, I can see that.

GuySartorelli:
Turns out I forgot to update the testdata dirnames. CI should pass now, I think.

stasadev:
Thanks for these great contributions @GuySartorelli!
Sorry for not reviewing this sooner, will do so next week.

rfay:
TestCmdGetInstalled seems to fail in https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8465#01915879-fc46-41e8-bc85-3abfc09b3ea1, looks probably related to this PR, (and with DDEV_DEBUG=true). Not sure why it wouldn't have shown up elsewhere, maybe because the `get` tests aren't done everywhere. 

We have a rest day tomorrow so I'll try to take a first review.

rfay:
It took me quite a bit of searching to find the `--installed` option, which is now `ddev add-on list --installed`. Almost everything else was immediately intuitive.

rfay:
While we're here, let's remove the "Add-ons installed before DDEV v1.22.0 will not be listed.", there's enough water under the bridge now that we don't have to say that any more.

> $ ddev add-on list --installed
No registered add-ons were found. Add-ons installed before DDEV v1.22.0 will not be listed.

cmd/ddev/cmd/service-enable.go
```diff
@@ -13,7 +14,7 @@ const disabledServicesDir = ".disabled-services"
 var ServiceEnable = &cobra.Command{
 	Use:     "enable service [project]",
 	Short:   "Enable a 3rd party service",
-	Long:    fmt.Sprintf(`Enable a 3rd party service. The service must exist as .ddev/%s/docker-compose.<service>.yaml. Note that you can use "ddev get" to obtain a service not already on your system.`, disabledServicesDir),
+	Long:    fmt.Sprintf(`Enable a 3rd party service. The service must exist as .ddev/%s/docker-compose.<service>.yaml. Note that you can use "ddev add-on get" to obtain a service not already on your system.`, disabledServicesDir),
```
rfay, line 17: I think we should probably go ahead and deprecate `ddev service` at this point. 
* https://github.com/ddev/ddev/issues/6298
Would it be reasonable to do that in this PR or not?

docs/content/users/extend/additional-services.md
```diff
@@ -79,29 +79,29 @@ For example,
 
 Officially-supported add-ons:
```
rfay, line 80: Overall, we want to do better with this, of course. It needs to be automatic, and it's a bit of a mess anyway. And we need a registry web page. Ideas welcome.

docs/content/users/extend/additional-services.md
```diff
@@ -79,29 +79,29 @@ For example,
...
-* [PDFreactor](https://github.com/ddev/ddev-pdfreactor): `ddev get ddev/ddev-pdfreactor`
-* [Platform.sh](https://github.com/ddev/ddev-platformsh): `ddev get ddev/ddev-platformsh`
-* [Proxy Support](https://github.com/ddev/ddev-proxy-support): `ddev get ddev/ddev-proxy-support`.
-* [Redis Commander](https://github.com/ddev/ddev-redis-commander): `ddev get ddev/ddev-redis-commander`.
-* [Redis](https://github.com/ddev/ddev-redis): `ddev get ddev/ddev-redis`.
-* [Selenium Standalone Chrome](https://github.com/ddev/ddev-selenium-standalone-chrome): `ddev get ddev/ddev-selenium-standalone-chrome`.
-* [Varnish](https://github.com/ddev/ddev-varnish): `ddev get ddev/ddev-varnish`.
+* [Adminer](https://github.com/ddev/ddev-adminer): `ddev add-on get ddev/ddev-adminer`.
+* [Apache Solr for Drupal](https://github.com/ddev/ddev-drupal-solr): `ddev add-on get ddev/ddev-drupal-solr`.
+* [Beanstalkd](https://github.com/ddev/ddev-beanstalkd): `ddev add-on get ddev/ddev-beanstalkd`.
```
rfay, line 84: Beanstalkd is gone
```suggestion
```

Please do a `ddev add-on list` to get the correct contents of this section.

docs/content/users/extend/additional-services.md
```diff
@@ -79,29 +79,29 @@ For example,
...
+* [Proxy Support](https://github.com/ddev/ddev-proxy-support): `ddev add-on get ddev/ddev-proxy-support`.
+* [Redis Commander](https://github.com/ddev/ddev-redis-commander): `ddev add-on get ddev/ddev-redis-commander`.
+* [Redis](https://github.com/ddev/ddev-redis): `ddev add-on get ddev/ddev-redis`.
+* [Selenium Standalone Chrome](https://github.com/ddev/ddev-selenium-standalone-chrome): `ddev add-on get ddev/ddev-selenium-standalone-chrome`.
+* [Varnish](https://github.com/ddev/ddev-varnish): `ddev add-on get ddev/ddev-varnish`.
 
 ## Managing Installed Add-Ons
 
-Add-ons installed in DDEV v1.22+ are versioned and can be viewed by running `ddev get --installed`.
+Add-ons installed in DDEV v1.22+ are versioned and can be viewed by running `ddev add-on list --installed`.
```
rfay, line 100:
```suggestion
Installed add-ons are versioned and can be viewed by running `ddev add-on list --installed`.
```

rfay (changes requested):
This is pretty great! Thanks for taking the time to explore and implement. It seems like it has little risk.

The one point of friction (happened to @rpkoller too I guess) was figuring out `ddev add-on list --installed`, so if we can figure out how to redirect people from that, we're already almost there.

I haven't taken a careful read of the rendered docs.

rfay:
And hi from Ireland where we have a rest day on our bike trip. Wish us well as the coming week looks really rainy.

cmd/ddev/cmd/service-enable.go
GuySartorelli: I'd rather keep this PR down to just what it needs to touch (especially since there's a separate issue for that work already). There's a lot of room for scope creep here but it'll be easiest to work on and to review if we just keep it narrow, and do separate work separately.

docs/content/users/extend/additional-services.md
GuySartorelli: I'd rather not touch things that aren't related to what this PR is for. If we remove that then we need to remove from the table as well, and maybe the spellcheck list.... I'd much rather that all be done in a separate piece of work, probably along side work to automate it.

docs/content/users/extend/additional-services.md
GuySartorelli: That's not really related to this PR but it's a tiny change so I'll make it.

GuySartorelli:
> It took me quite a bit of searching to find the --installed option, which is now ddev add-on list --installed. Almost everything else was immediately intuitive. I guess maybe additional (even deprecated?) ddev add-on --installed or ddev add-on installed ? Or they could be just errors pointing to the correct usage?

Moving from `ddev get --list --installed` to `ddev add-on list --installed` - they both have "list" and "installed", I'm not sure what else I can do to improve the overall intuitiveness of this.

I think `ddev add-on --installed` or `ddev add-on installed` that function fully would just encourage using the wrong command, but I could add one that just errors with a message telling people about the correct command, if you think people will have that much trouble with them. I haven't done that yet because to me it doesn't feel necessary to me but I'm happy to be overruled on that.

In the meantime I've added an example to the `ddev add-on` usage. We could also add a "message of the day" saying something like "Want to see which add-ons you have installed? Run 'ddev add-on list --installed'"

> While we're here, let's remove the "Add-ons installed before DDEV v1.22.0 will not be listed.", there's enough water under the bridge now that we don't have to say that any more.

Done

> TestCmdGetInstalled seems to fail in https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8465#01915879-fc46-41e8-bc85-3abfc09b3ea1, looks probably related to this PR, (and with DDEV_DEBUG=true). Not sure why it wouldn't have shown up elsewhere, maybe because the get tests aren't done everywhere.

I would expect the get tests to be done everywhere? I'd be very surprised if something so important was limited to only being tested on mac and only when mutagen is used...

The `TestCmdGetInstalled` and `TestCmdGetComplex` tests do run for me locally on linux. I can't test with a mac locally so I can't look into this failure any further than that.

docs/content/users/extend/additional-services.md
rfay: The fix is just to do `ddev add-on list` and inject it here, since the format changed. You don't have to care about beanstalkd.

rfay: It's probably just that copy-pasting this list is silly and we need to move to the add-on registry.

rfay:
I see we only run TestCmdGetInstalled in [one place](https://github.com/ddev/ddev/blob/52a71ca3cad6b723377b5628a86351050d532355/.buildkite/macos-docker-desktop-arm64.yml#L15), macOS-arm64, and nowhere else, so the fact that it's failing there is particularly important,

```
func TestCmdGetInstalled(t *testing.T) {
	if os.Getenv("DDEV_RUN_GET_TESTS") != "true" {
		t.Skip("Skipping because DDEV_RUN_GET_TESTS is not set")
	}
```

rfay:
I guess we should change the names of the tests that are affected also, from Get -> AddOn (hyphen won't work I don't think, OK if it does)

rfay:
> The TestCmdGetInstalled and TestCmdGetComplex tests do run for me locally on linux. I can't test with a mac locally so I can't look into this failure any further than that.

If TestCmdGetInstalled runs for you locally (as opposed to being skipped, it means you have DDEV_RUN_GET_TESTS=true set in your test environment, which is good. But it's not succeeding anywhere in automated testing

The reason these are largely skipped is because do what we might, Github would always hit a rate-limit and turn them down, resulting in intermittent failures.

docs/content/users/extend/additional-services.md
rfay: It's important to improve docs as we go on all these things. And of course since I gave the suggestion and you can just click a button to add it it's not too hard to fix.

image

GuySartorelli: > it's not too hard to fix.

That's why I did make this change :p

GuySartorelli:
> I guess we should change the names of the tests that are affected also, from Get -> AddOn (hyphen won't work I don't think, OK if it does)

I did do this - as noted in the PR description I kept the originals (to validate that the proxy is working as expected) as well as copying the tests to test the new command, and updated the names of the copies.

rfay:
I think you'll be able to demonstrate the error in https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8471#019162d5-0fb7-44d0-b1d4-f6ecda1c5029/16498-16500 if you set up your test to run with `export DDEV_DEBUG=true DDEV_RUN_GET_TESTS=true`

It's still using `ddev get` instead of using `ddev add-on`, so the test needs to be updated to use the new command.

cmd/ddev/cmd/get_test.go
```diff
@@ -18,7 +18,10 @@ import (
 	asrt "github.com/stretchr/testify/assert"
 )
 
-// TestCmdGet tests various `ddev get` commands .
+// The tests in this file are duplicates of the various addon*_test.go files
+// Remove these when the deprecated get command is removed.
+
+// TestCmdGet tests various `ddev get` commands.
 func TestCmdGet(t *testing.T) {
```
rfay, line 25: Change the name and the commend for new usage.

cmd/ddev/cmd/get_test.go
```diff
@@ -130,7 +133,7 @@ func TestCmdGetComplex(t *testing.T) {
 	_, err = os.Create(app.GetConfigPath("extra/no-ddev-generated.txt"))
 	require.NoError(t, err)
 
-	out, err := exec.RunHostCommand(DdevBin, "get", filepath.Join(origDir, "testdata", t.Name(), "recipe"))
+	out, err := exec.RunHostCommand(DdevBin, "get", filepath.Join(origDir, "testdata", "TestCmdAddonComplex", "recipe"))
```
rfay, line 136: Use the new add-on command throughout.

rfay (changes requested):
A few more things to fix in tests. TestCmdGet changes name, maybe there's other things in that file. Change `get` to `add-on` in that test.

cmd/ddev/cmd/get_test.go
GuySartorelli: I've mentioned this a few times now but I'll say it again.
I've intentionally kept these tests exactly as they are. They're intentionally testing `ddev get` so that we know that the proxy is working the way we expect it to. This will be especially important if functionality is added to or changed in the various `ddev add-on` sub-commands.

The tests in this file have all been copied into `addon_test.go` and `addon-get_test.go`, with appropriate names.

If you like, I can delete these old ones - but I think it makes sense to keep them until you eventually remove `ddev get`, or else you don't actually know that `ddev get` is still working as a proxy for `ddev add-on`.

GuySartorelli:
> I think you'll be able to demonstrate the error in https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8471#019162d5-0fb7-44d0-b1d4-f6ecda1c5029/16498-16500 if you set up your test to run with `export DDEV_DEBUG=true DDEV_RUN_GET_TESTS=true`

Ahh, yes, now that I run the following command I can see the failure locally:
`DDEV_DEBUG=true DDEV_RUN_GET_TESTS=true make test TESTARGS="-run TestCmdGetInstalled"`

I'll look into this failure.

> It's still using `ddev get` instead of using `ddev add-on`, so the test needs to be updated to use the new command.

I've already responded to this several times. We intentionally have tests both for `ddev get` and `ddev add-on`. If you want me to remove the tests for `ddev get` I can do so, but I think that would be a mistake.

GuySartorelli:
Looks like the CI failure was caused by the deprecation warning being emitted in a non-JSON format even when `--json-output` was used.

I've changed the way we're doing this so that the deprecation notice is now built as part of running the command, and simply isn't emitted for JSON output.
The bonus for this is that we have more control over the deprecation warning message, so I've made it dynamic based on the command being used.

For example, when running `ddev get --installed`, the deprecation warning will say to use `ddev add-on list --installed` instead. This should help to alleviate the confusion around that command that was mentioned earlier.

`DDEV_DEBUG=true DDEV_RUN_GET_TESTS=true make test TESTARGS="-run TestCmdGetInstalled"` now passes fully locally.

rfay:
Thanks for the good work on this.

If you want to keep tests for `get`, it's fine with me, but they should be in addition to the mainline tests for the mainline command. So do a loop, or a matrix, or however you prefer, and they should preferably be in the tests for `ddev add-on` Thanks!

GuySartorelli:
> If you want to keep tests for get, it's fine with me, but they should be in addition to the mainline tests for the mainline command

Please read my many comments about this where I've already mentioned several times that this PR *does include tests for `ddev add-on`*!

See the `addon_test.go` and `addon-get_test.go` files, which contain those tests. They're copies of the tests that run against `ddev get`, except these ones run against `ddev add-on`.

I don't know how many different ways I can tell you that... I'm starting to think I've misunderstood what you want, but every time I read what you said, it sounds like you're asking for what I've already done.

rfay:
You have a tendency to be a very verbose writer, people like me don't necessarily grok all the things you say on all the places :) I'm also on a bike trip in Ireland with incomplete attention span.

Here's what I want:

* The new command is `ddev add-on`
* The tests should be for `ddev add-on` and should be named and used appropriately
* Optional capabilities in the add-on tests to test `ddev get` would be welcome, with "TODO" on them saying "Can be removed after 2025-08-01" or some such.

This is all about long-term maintenance, helping make it clear what's important in the long term.

Thanks for the great work and all the initiative on this. It really is excellent.

GuySartorelli:
> The new command is `ddev add-on`

Done.

> The tests should be for `ddev add-on` and should be named and used appropriately

Done, see the files `addon_test.go` and `addon-get_test.go`

> Optional capabilities in the add-on tests to test ddev get would be welcome, with "TODO" on them saying "Can be removed after 2025-08-01" or some such.

Done, these are left as they used to be in the original file, so it's easy to find and remove when the `ddev get` command itself gets removed.

The commend there is as follows. I can change it to explicitly say "TODO" if you need that. 

> // The tests in this file are duplicates of the various addon*_test.go files
> // Remove these when the deprecated get command is removed.

rfay:
Actually, on second thought, let's just remove the `ddev get` tests if there's equivalent coverage for `ddev add-on`. Thanks.

GuySartorelli:
> Actually, on second thought, let's just remove the `ddev get` tests if there's equivalent coverage for `ddev add-on`. Thanks.

Done.

Bear in mind this means there's no longer any test coverage for `ddev get`. Any changes to `ddev add-on` going forward will need to have very good *manual* testing of `ddev get` to ensure there are no regressions or breaking changes to that command until it gets deleted..

rfay:
Thanks. IMO deprecated things don't need quite as much care. But I do think you're right to worry a bit, because old habits are hard to break.

rfay:
Oh, just realized that nearly every single add-on will need a PR to change their README. It's only just a bit over 100 :(

GuySartorelli:
> Oh, just realized that nearly every single add-on will need a PR to change their README. It's only just a bit over 100 :(

Thankfully there's no immediate rush on that - I imagine `ddev get` will still work for a fairly long while.
In the meantime if you can give me a list of all of them, I can write a script to automate the process of updating the readme and generating PRs.

rpkoller:
there are currently 112: 

```
2ndkauboy/ddev-elasticvue
a11ywatch/ddev-a11ywatch
b13/ddev-keycloak
b13/ddev-rabbitmq
b13/ddev-stirling-pdf
biati-digital/ddev-wp-rename-tables-prefix
brandonkramer/ddev-vite
bricebou/ddev-manticoresearch
bserem/ddev-selenium-standalone-firefox
bserem/ddev-typesense
carsten-walther/ddev-typo3-solr
catch56/ddev-gander
ChromaticHQ/ddev-dblog
claudiu-cristea/ddev-virtuoso
codemonauts/ddev-s3-http-proxy
davereid/ddev-drupal-contrib-dev
ddev/ddev-adminer
ddev/ddev-browsersync
ddev/ddev-cron
ddev/ddev-drupal-contrib
ddev/ddev-drupal-solr
ddev/ddev-elasticsearch
ddev/ddev-ioncube
ddev/ddev-memcached
ddev/ddev-minio
ddev/ddev-mongo
ddev/ddev-opensearch
ddev/ddev-pdfreactor
ddev/ddev-phpmyadmin
ddev/ddev-platformsh
ddev/ddev-proxy-support
ddev/ddev-redis
ddev/ddev-redis-7
ddev/ddev-redis-commander
ddev/ddev-selenium-standalone-chrome
ddev/ddev-solr
ddev/ddev-sqlsrv
ddev/ddev-typo3-solr
ddev/ddev-varnish
ddev/ddev-xhgui
dermatz/ddev-woodoo-buildtools-magento
DiffyWebsite/ddev-diffy
eakcize/ddev-timescaledb
eakcize/ddev-timescaledb-with-mongo-and-mysql-fdw
echavaillaz/ddev-gotenberg
envsa/ddev-pnpm
feuertelegraph/ddev-firebird
fwust/ddev-clamav
gebruederheitz/ddev-headless-chrome
gebruederheitz/ddev-taskfile
GetDKAN/ddev-dkan
gschrisstadler/ddev-snowflake-odbc
haase-fabian/ddev-neos
hanoii/ddev-pimp-my-shell
hanoii/ddev-readme
hanoii/ddev-sshd
iamntz/ddev-rename-project
iljapolanskis/ddev-buggregator
JanoPL/ddev-kibana
javivf/ddev-keycloak
julienloizelet/ddev-playwright
julienloizelet/ddev-portainer
justafish/ddev-drupal-core-dev
kevinquillen/ddev-meilisearch
kevinquillen/ddev-typesense
Lullabot/ddev-playwright
machine-rc/ddev-dotnet
machine-rc/ddev-frontend-nextjs
machine-rc/ddev-graphql
machine-rc/ddev-python
madebydaniz/ddev-orthanc
mandrasch/ddev-addon-statamic-please
mattocchi/ddev-db-mssql
MelaineGerard/ddev-directus
Metadrop/ddev-aljibe
mmunz/ddev-backstopjs
Morgy93/ddev-bun
Morgy93/ddev-gum
MurzNN/ddev-grafana
MurzNN/ddev-pgadmin
netz98/ddev-n8n
netz98/ddev-qdrant
nireneko/ddev-mkdocs
oblakstudio/ddev-xhgui-pro
penyaskito/ddev-hugo
reloxx13/ddev-swagger-ui
rfay/ddev-drushonhost
rfay/ddev-php-patch-build
rpkoller/ddev-spidergram
s2b/ddev-vite-sidecar
sebastian-ehrling/ddev-opensearch
skeemer/ddev-wkhtmltox
ssmarco/ddev-apache-tika
ssmarco/ddev-enterprise-search
Tapkaros-rs/ddev-stripe
thunder/ddev-selenium-chrome
thursdaybw/ddev-tightvnc
torenware/ddev-viteserve
torhoehn/ddev-opensearch
torhoehn/ddev-typesense
tyler36/ddev-cypress
tyler36/ddev-dbslow
tyler36/ddev-gitpod-setup
tyler36/ddev-laravel-queue
tyler36/ddev-locale
tyler36/ddev-php-dumper
tyler36/ddev-qr
tyler36/ddev-storybook
tyler36/ddev-tinker
tyler36/ddev-vite
tyler36/ddev-vscode-devcontainer
valicm/ddev-dynamodb
```

GuySartorelli:
@rpkoller If I stick `www.github.com/` in front of all those will I get the correct repo?

rpkoller:
yes... i just quickly sanitized you the list that gets output from `ddev get --list --all`... and prepending github.com will get you to the repo

GuySartorelli:
Sweet. I'll try to get to making those PRs sometime this week. It'll update to say something like

> For DDEV v1.24.0 or above run
> 
> ```bash
> ddev add-on get the/add-on
> ```
> 
> For earlier versions of DDEV run
> 
> ```bash
> ddev get the/add-on
> ```

@rfay If this is slated for an earlier release than 1.24 let me know so I can set the correct version in the PRs

rfay:
I think we can work on it after the next release, which will probably be v1.23.5. We'll do a major release (v1.24.0) changing PHP and Node.JS defaults later in the fall.

The related https://github.com/ddev/ddev-add-on-monitoring may also give us some tools to work with on this.

docs/content/users/extend/additional-services.md
```diff
@@ -7,14 +7,14 @@ search:
...
+Anyone can create their own services with a `.ddev/docker-compose.*.yaml` file, and a growing number of popular services are supported and tested, and can be installed using the [`ddev add-on get`](../usage/commands.md#add-on-get) command.
 
-Use `ddev get --list` to see available add-ons. To see all possible add-ons (not necessarily supported or tested), use `ddev get --list --all`.
+Use `ddev add-on list` to see available add-ons. To see all possible add-ons (not necessarily supported or tested), use `ddev add-on list --all`.
 
 For example,
 
 ```
-→  ddev get --list
+→  ddev  add-on list
```
stasadev, line 17:
```suggestion
→  ddev add-on list
```

stasadev (changes requested):
I noticed a panic here:

```
$ ddev get
Command "get" is deprecated, use 'ddev add-on get' instead
panic: runtime error: slice bounds out of range [1:0]
```

GuySartorelli:
> I noticed a panic here:

Oh, does using `output.stdErr` automatically create a panic? If so that kinda sucks but I guess I'll just have to use `stdOut` instead...

I'll try to find some time this weekend to sort that out.

cmd/ddev/cmd/get.go
```diff
@@ -45,454 +24,49 @@ ddev get --remove someowner/ddev-someaddonname,
...
-			util.Failed("Unable to create manifest file: %v", err)
-		}
-
-		util.Success("\nInstalled DDEV add-on %s, use `ddev restart` to enable.", sourceRepoArg)
-		if argType == "github" {
-			util.Success("Please read instructions for this add-on at the source repo at\nhttps://github.com/%v/%v\nPlease file issues and create pull requests there to improve it.", owner, repo)
-		}
-		output.UserOut.WithField("raw", manifest).Printf("Installed %s:%s from %s", manifest.Name, manifest.Version, manifest.Repository)
+		// Hand off execution for ddev get <add-on>
+		AddonGetCmd.Run(cmd, args)
```
stasadev, line 66: > Oh, does using output.stdErr automatically create a panic?

No, it's from here, we need to check the length:

```suggestion
		if len(args) > 1 {
			AddonGetCmd.Run(cmd, args)
		}
```

rfay:
Just want to say again how happy I am that you took on this important terminology change. Sometimes I can be so reluctant to change things for fear of breaking people's workflow, but this will be a great addition.

rfay:
This has some minor conflicts to be resolved.

docs/content/users/extend/additional-services.md
GuySartorelli: Updated

cmd/ddev/cmd/get.go
GuySartorelli: Ahh, interesting. I had expected the validation of the new command to kick in and say "requires at least 1 arg(s), only received 0" but it seems that validation is skipped when running the function directly like this.

Have added back the original messaging from the old `ddev get` when no args are passed.

GuySartorelli:
Rebased and requested changes made

GuySartorelli:
Nit sure what's causing that one build kits failure. It failed an arbitrary test after 4 hours running, but when I try rerunning it doesn't get even 5 minutes in before failing.

rfay:
The failure on https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/8491#019182be-ef02-4283-89eb-8733fa61a52e is a broken test runner, I'll reboot it and restart the test.

cmd/ddev/cmd/get.go
```diff
@@ -45,454 +25,52 @@ ddev get --remove someowner/ddev-someaddonname,
...
-		requestedVersion := ""
-
-		if cmd.Flags().Changed("version") {
-			requestedVersion = cmd.Flag("version").Value.String()
-		}
-
-		if cmd.Flags().Changed("verbose") {
-			verbose = true
+		// Create deprecation notice.
+		// Not we don't just set this as the "Deprecated" property of the cobra.Command
```
rfay, line 29: Oh-so-trivial, add with [skip ci]
```suggestion
		// Note we don't just set this as the "Deprecated" property of the cobra.Command
```

rfay (approved):
Looking great, thanks!

* There's a trivial reference to `ddev get` in https://github.com/ddev/ddev/blob/master/pkg/ddevapp/dotddev_assets/providers/platform.yaml#L6 that could be fixed with [skip ci]

Thanks so much for this!

GuySartorelli:
> There's a trivial reference to ddev get in https://github.com/ddev/ddev/blob/master/pkg/ddevapp/dotddev_assets/providers/platform.yaml#L6

Looks like I got that one: https://github.com/ddev/ddev/pull/6482/files#diff-a671f8ba15fea3be28d3307afa46681463f4029f1559ba45c91de0489eb84fd7R6

rfay:
Yes, you did in fact. There must have been a spare .ddev created in my local working tree, in ddevapp, by a test. That means there's a test somewhere that isn't properly changing directories :) Has nothing to do with this PR though.

stasadev (approved):
This works perfectly 🙌

I especially like that I have autocomplete with `ddev addon <tab>` because I was afraid it wouldn't work if we didn't have autocomplete for the alias itself.

PR #6484, opened 2024-08-14
fix: always show `pv` output on `ddev import-db`

stasadev:
## The Issue

- https://github.com/orgs/ddev/discussions/6483

Users cannot redirect `pv` output on `ddev import-db`

To reproduce:

```
ddev exec "pv /dev/zero | head -c 5M | while read -r line; do sleep 0.01; done" >stdout.txt 2>stderr.txt
```
And view inside `stderr.txt` - it is empty.

Uses `pv -f` instead of `pv`

```
$ pv -h
Usage: pv [OPTION] [FILE]...
Concatenate FILE(s), or standard input, to standard output, with monitoring.
...
  -f, --force                    output even if standard error is not a terminal
...
```

```
ddev exec "pv -f /dev/zero | head -c 5M | while read -r line; do sleep 0.01; done" >stdout.txt 2>stderr.txt
```
And view inside `stderr.txt` - it should contain the output from `pv`.

Or
```
ddev import-db --file=db.sql.gz >stdout.txt 2>stderr.txt
```
And view inside `stderr.txt` - it should contain the output from `pv`.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -717,7 +717,7 @@ func (app *DdevApp) ImportDB(dumpFile string, extractPath string, progress bool,
 		}
 
 		// Case for reading from file
-		inContainerCommand = []string{"bash", "-c", fmt.Sprintf(`set -eu -o pipefail && mysql -uroot -proot -e "%s" %s && pv %s/*.*sql |  perl -p -e 's/^(\/\*.*999999.*enable the sandbox mode *|CREATE DATABASE \/\*|USE %s)[^;]*(;|\*\/)//' | mysql %s %s`, preImportSQL, nodeps.MySQLRemoveDeprecatedMessage, insideContainerImportPath, "`", targetDB, nodeps.MySQLRemoveDeprecatedMessage)}
+		inContainerCommand = []string{"bash", "-c", fmt.Sprintf(`set -eu -o pipefail && mysql -uroot -proot -e "%s" %s && pv -f %s/*.*sql |  perl -p -e 's/^(\/\*.*999999.*enable the sandbox mode *|CREATE DATABASE \/\*|USE %s)[^;]*(;|\*\/)//' | mysql %s %s`, preImportSQL, nodeps.MySQLRemoveDeprecatedMessage, insideContainerImportPath, "`", targetDB, nodeps.MySQLRemoveDeprecatedMessage)}
 
 		// Alternate case where we are reading from stdin
 		if dumpFile == "" && extractPath == "" {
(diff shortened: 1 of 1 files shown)


Mathijs-Zuiderlicht:
I ran your manual testing instructions in Gitpod: ddev import-db --file=db.sql.gz >stdout.txt 2>stderr.txt
Here's the contents of stderr.txt

```
 896KiB 0:00:01 [ 818KiB/s] [<=>                                               ]
1.50MiB 0:00:02 [ 676KiB/s] [ <=>                                              ]
2.25MiB 0:00:03 [ 698KiB/s] [  <=>                                             ]
2.88MiB 0:00:04 [ 691KiB/s] [   <=>                                            ]
3.62MiB 0:00:05 [ 683KiB/s] [    <=>                                           ]
4.25MiB 0:00:06 [ 678KiB/s] [     <=>                                          ]
4.88MiB 0:00:07 [ 698KiB/s] [      <=>                                         ]
5.00MiB 0:00:07 [ 698KiB/s] [       <=>                                        ]
```

Looks alright to me. 
I tried your test command on my machine without your changes and got an empty stderr.txt.

rfay (commented):
I was afraid that this might break importing from stdin, but it seems to be fine.

stasadev:
@Mathijs-Zuiderlicht,

I just tried this in Gitpod:

```
ddev export-db --file=db.sql.gz
ddev import-db --file=db.sql.gz >stdout.txt 2>stderr.txt
```
stderr.txt:
```
9.03MiB 0:00:01 [8.98MiB/s] [========================>         ] 76% ETA 0:00:00
11.7MiB 0:00:01 [7.35MiB/s] [================================>] 100%            
```

---

```
ddev export-db --file=db.sql --gzip=false
ddev import-db --file=db.sql >stdout.txt 2>stderr.txt
```
stderr.txt:
```
9.03MiB 0:00:01 [8.84MiB/s] [========================>         ] 76% ETA 0:00:00
11.7MiB 0:00:01 [7.27MiB/s] [================================>] 100%        
```

rfay (changes requested):
What this PR does is to force output even if stderr is not a terminal. From `pv -h`:

>   -f, --force              output even if standard error is not a terminal

The purpose of (not) forcing is to not display things in a non-interactive environment that don't belong there. For example, the test instructions given here that redirect `2>/stderr.txt`. The reality is that we really wouldn't want those to happen most of the time.

The simple answer for the OP in https://github.com/orgs/ddev/discussions/6483 is to change your node process to pretend to be a terminal. I'm not sure that's possible, but that's the easy answer. The reason that pv isn't showing output is that the stderr isn't a terminal. 

I'm open to conversation, but changing working code to unexpected code for a very unusual situation seems like it may be the wrong thing, and the OP code could possibly adapt instead.

stasadev:
> Could this break any existing automated tests that use `ddev import-db` ?

I think this change to `stderr` may have an impact if someone's code checks the output in `stderr`.

And we can't redirect the output of `pv` to `stdout` because that would completely override it.

rfay:
I think it may be OK, just wonder if the experimental node code could just act like a terminal and that would solve the problem.

rfay:
There may be some answers in https://stackoverflow.com/questions/68344362/how-do-i-get-full-visual-control-of-a-terminal-with-node

In general, the answer would probably be "behave like a terminal in order to get code like pv to think you're a terminal"

Mathijs-Zuiderlicht:
I ended up doing something like this:
`const importCommand = 'wsl -d Ubuntu pv -f ' + dumpFilePath + ' | wsl -d Ubuntu --cd ' + appRoot + ' ddev mysql -uroot -proot ' + myDB`
instead of: 
`const importCommand = 'wsl -d Ubuntu --cd ' + appRoot + ' ddev import-db --file=${dumpFileName}';`

It works perfectly for my use case. This also doesn't require any changes in DDEV ;)

@rfay and @stasadev thank you for looking into it.

stasadev:
Closing this.

A workaround is to pipe the `ddev import-db` command to `pv` as shown above.

PR #6485, opened 2024-08-14
ci: add new check from golangci-lint

stasadev:
## The Issue

- https://github.com/golang/go/issues/60529

```
golangci-lint: 
pkg/exec/exec.go:114:25: printf: non-constant format string in call to (*github.com/sirupsen/logrus.Logger).Printf (govet)
                output.UserOut.Printf("RunHostCommand: " + command + " " + strings.Join(args, " "))
...
```

Pacifies golangci-lint.

### DIFF
--- a/cmd/ddev/cmd/debug-mutagen.go
+++ b/cmd/ddev/cmd/debug-mutagen.go
@@ -1,12 +1,13 @@
 package cmd
 
 import (
+	"os"
+
 	"github.com/ddev/ddev/pkg/exec"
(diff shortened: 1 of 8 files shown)


rfay (approved):
Thanks. IIRC a person went through and checked for these earlier, but there must have been some missed. Several of thse could also have been fixed by using the non -`f` version of the output command.

rfay:
Looks like we're having lots of trouble with TestWebserverMariaMySQLDBClient, will probably also have to sort that out. But this is fine to pull and will be a good thing.

PR #6488, opened 2024-08-15
docs: add instructions to access DDEV sites on iOS simulator

cenoura:
## The Issue

Self signed certificates from host machine are not available automatically on iOS simulator.
Given that, when trying to open a site served by DDEV, the user will see an error:

ssl-issue

This PR update docs and include a new item on FAQ with instructions to install the self signed certificate on the iOS simulator to be able to open the site properly.

https://ddev--6488.org.readthedocs.build/en/6488/users/usage/faq/#how-to-access-ddev-sites-https-urls-on-ios-simulator

## Automated Testing Overview

documentation updates only, no test updates needed

## Release/Deployment Notes

documentation updates only, no release notes needed

### DIFF
--- a/.spellcheckwordlist.txt
+++ b/.spellcheckwordlist.txt
@@ -167,6 +167,7 @@ WSL
 WSL2
 WantedBy
 Webhosting
+Xcode
 XDebug
 XDG
 Xhprof
--- a/docs/content/users/usage/faq.md
+++ b/docs/content/users/usage/faq.md
@@ -150,6 +150,21 @@ Probably! You’ll need to install an older, unsupported version of Docker Deskt
 
 Check out [this Stack Overflow answer](https://stackoverflow.com/a/69964995/897279) for a walk through the process.
 
+### How to access DDEV sites HTTPS URLs on iOS Simulator?
+
+When using the iOS Simulator (provided with Xcode), you may encounter issues accessing DDEV sites that use SSL.
+
+This is because the iOS Simulator does not trust the self-signed SSL certificates that DDEV uses by default.
+
+To resolve this issue, you need to install the DDEV root certificate on the iOS Simulator.
+
+Here’s how to do it:
+
+1. Open iOS Simulator and keep it on the home screen.
+2. On your Mac, open Finder and navigate to `~/Library/Application Support/mkcert/`.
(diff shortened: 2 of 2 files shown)


stasadev (commented):
Thank you @cenoura.

I think this iOS simulator is part of Xcode, right?
Maybe we also need to mention Xcode somewhere for better searchability?

stasadev:
Our spell checker doesn't like the word `Xcode`, please add it here https://github.com/ddev/ddev/blob/master/.spellcheckwordlist.txt#L169

Or I can add it later myself.

stasadev (approved):
Looks good to me, thanks.

rfay (changes requested):
I apologize, but this doesn't belong in the FAQ, but it's super valuable.

However, it really is wrong to focus on IOS simulator. It's the same problem if you want to use any device outside the current machine. 

The `mkcert` docs at https://github.com/FiloSottile/mkcert?tab=readme-ov-file#mobile-devices explain exactly what to do, which I imagine is where you got this, but it's also the solution if you want to use a browser on the local network, etc.

A much easier technique is to use `ddev share`, see https://ddev.readthedocs.io/en/stable/users/topics/sharing/

That's probably what most people do for this kind of situation.

I'd love to have you do a guest blog on ddev.com (see https://github.com/ddev/ddev.com) explaining all the ways you can get an external browser to trust `mkcert` certificates. 

But at the very least, please do a Stack Overflow question-and-answer-your-own-question, explaining exactly what you're doing and why, it will be very useful.

But this is definitely not an FAQ, and is not even mainstream enough for the docs, unless you rewrite it as a very general answer (which also includes your emulator) and put it on the Sharing page linked above.

stasadev:
> It's the same problem if you want to use any device outside the current machine.

It's true, I remember having to deal with certificates in Postman to get HTTPS working.

> this doesn't belong in the FAQ, but it's super valuable

Yeah, this will probably get lost in the big FAQ page, I didn't think of that at first.

---

@cenoura, sorry for the quick approval, but it needs to be done elsewhere.

cenoura:
@rfay thank you for the feedback and explanations.
TBH, I didn't know about `ddev share` - this need to access the site on other devices only happened to me two days ago.
I'm closing the PR and will try to get some time to do one of your suggestions.

PR #6496, opened 2024-08-19
fix: flush mutagen after deletion of .downloads directory

rfay:
## The Issue

Twice people have reported having trouble with `ddev pull pantheon --skip-files` where it tried to import files for some reason, or tried to import things that weren't there. 
* https://discord.com/channels/664580571770388500/1274137556858175642
* https://discord.com/channels/664580571770388500/1232668554550186024

We delete the .downloads directory before downloading, but we don't do a mutagen sync, so this all may happen too fast, and sync may be incomplete, leading to files being in the .downloads/files directory, messing up the import. 

This does a sync before recreating the .downloads directory

In addition, we should never be trying to import directories during a database import process, so it's an error to try to traverse directories as was shown in the Discord OP.

In https://discord.com/channels/664580571770388500/1274137556858175642/1274137556858175642, @travisbutterfield reported the below. Clearly we shouldn't have been trying to import the subdirectory `files` even though it existed.

```
ddev pull pantheon --skip-files -y
Authenticating...
 [notice] Logging in via machine token.
 [notice] Found a machine token for ************@***.***.
 [notice] Logged in via machine token.
Obtaining databases...
+ set -eu -o pipefail
+ ls /var/www/html/.ddev
+ pushd /var/www/html/.ddev/.downloads
+ terminus backup:get asu-isearch.live --element=db --to=db.sql.gz
 [notice] Downloading asu-isearch_live_2024-08-16T20-00-00_UTC_database.sql.gz to db.sql.gz
Importing databases [/Users/tbutterf/web/asu-isearch/.ddev/.downloads/db.sql.gz /Users/tbutterf/web/asu-isearch/.ddev/.downloads/files]

1.12GiB 0:01:11 [16.1MiB/s] [====================================================>] 100%
Pull failed: unable to validate import asset /Users/tbutterf/web/asu-isearch/.ddev/.downloads/files: invalid asset: provided path is not a .sql file or archive
```

We're hoping that @travisbutterfield will be able to recreate the problem and we can test against it.

Otherwise, put some files in .ddev/.downloads/files/... and see if it can make the problem happen (without this fix)

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -73,7 +73,7 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 	}
 
 	for _, commandSet := range []string{projectCommandPath, globalCommandPath} {
-		commandDirs, err := fileutil.ListFilesInDirFullPath(commandSet)
+		commandDirs, err := fileutil.ListFilesInDirFullPath(commandSet, false)
 		if err != nil {
 			return err
 		}
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -271,7 +271,7 @@ func GatherAllManifests(app *DdevApp) (map[string]AddonManifest, error) {
 		return nil, err
 	}
 
-	dirs, err := fileutil.ListFilesInDirFullPath(metadataDir)
+	dirs, err := fileutil.ListFilesInDirFullPath(metadataDir, false)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/ddevapp/envfile_test.go
+++ b/pkg/ddevapp/envfile_test.go
@@ -1,14 +1,15 @@
 package ddevapp_test
 
 import (
-	"github.com/ddev/ddev/pkg/ddevapp"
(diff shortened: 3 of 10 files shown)


rfay:
We need to do a bit more. 

We're trying to import *directories* in

```
	sqlTarballs, err := fileutil.ListFilesInDirFullPath(p.getDownloadDir())
	if err != nil || sqlTarballs == nil {
		return nil, fmt.Errorf("failed to find downloaded files in %s: %v", p.getDownloadDir(), err)
	}
```

because ListFilesInDirFullPath() will get subdirs.

rfay:
I'm not particularly proud of `ListFilesInDirNoSubdirsFullPath` as opposed to adding an arg to `ListFilesInDirFullPath` with an argument. But I did add a test. I just didn't want to add multiple changes to other things. Maybe I should rethink that.

rfay:
Refactored as suggested to just add a param to the `ListFilesInDirFullPath()` instead of adding a new function.

stasadev (commented):
Great, this is what I hoped for 👍

Issue #6499, opened 2024-08-22
When offline: prevent Docker pulls if a new Docker image is required


jenlampton:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

I'll revisit this issue once I have a stable internet connection and add the results of `ddev debug test`, but I suspect they may not be necessary since this is a known feature request? (I did check the box for this... promise not to do that regularly!)

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Please make sure that you have already looked at troubleshooting guide ========
Troubleshooting guide: https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/ 
Simple things to check:
* ddev poweroff
* Restart Docker Provider
* Reboot computer
* Temporarily disable VPN and firewall
Press any key to continue:


======== Output file will be in /var/folders/mw/_p1cl_rx69b3ll553w9cdvcc0000gn/T/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/local/bin/ddev
/usr/local/bin/ddev

These config files were loaded for project emerisa: [/Users/jlampton/Sites/emerisa/drupal/docroot/.ddev/config.yaml]
name: emerisa
type: drupal7
php_version: 7.4
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.23.2
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.11}
project_tld: ddev.site
composer_version: 2
nodejs_version: 20
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
No registered add-ons were found. Add-ons installed before DDEV v1.22.0 will not be listed.
Update them with `ddev get` so they'll be shown.


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /Users/jlampton/Sites/emerisa/drupal/docroot
#ddev-generated
unmodified #ddev-generated found in .ddev/mutagen/mutagen.yml


======== Creating dummy project named tryddevproject-14427 in ../tryddevproject-14427 ========
[...]

PR #6502, opened 2024-08-23
docs: clearer statement about `ddev start` before offline, for #6499

rfay:
## The Issue

* #6499 

The docs on offline usage need to be stronger about `ddev start` on each project before going offline.

Review new page at https://ddev--6502.org.readthedocs.build/en/6502/users/usage/offline/

### DIFF
--- a/docs/content/users/usage/offline.md
+++ b/docs/content/users/usage/offline.md
@@ -1,31 +1,24 @@
 # Using DDEV Offline
 
-DDEV attempts to work smoothly offline, and you shouldn’t have to do anything to make it work:
+!!!warning "Make sure to do a `ddev start` on every project you want to use before doing offline"
 
-* It doesn’t attempt instrumentation or update reporting if offline
-* It falls back to using `/etc/hosts` entries if DNS resolution fails
+    To work offline, you have to already have any needed Docker images pulled. That means you should do a `ddev start` on all of the projects before going offline. 
 
-However, it does not (yet) attempt to prevent Docker pulls if a new Docker image is required, so you’ll want to make sure that you try a [`ddev start`](../usage/commands.md#start) before going offline to make sure everything has been pulled.
+DDEV attempts to work smoothly offline, and you shouldn’t have to do anything to make it work.
 
-If you have a project running when you’re online (using DNS for name resolution) and you then go offline, you’ll want to do a [`ddev restart`](../usage/commands.md#restart) to get the hostname added into `/etc/hosts` for name resolution.
+However, it cannot pull needed Docker images when offline if a new Docker image is required, so you’ll want to make sure that you try a [`ddev start`](../usage/commands.md#start) before going offline to make sure everything has been pulled.
 
-You have general options as well:
+If you have a project running when you’re online (using DNS for name resolution) and you then go offline, do a [`ddev restart`](../usage/commands.md#restart) to get the hostname added into `/etc/hosts` for name resolution.
+
+You have some general options as well:
 
 In `.ddev/config.yaml`, [`use_dns_when_possible: false`](../configuration/config.md#use_dns_when_possible) will make DDEV never try to use DNS for resolution, instead adding hostnames to `/etc/hosts`. You can also use `ddev config --use-dns-when-possible=false` to set this configuration option.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/offline.md
```diff
@@ -1,31 +1,24 @@
...
+You have some general options as well:
 
-In `.ddev/config.yaml`, you can use `project_tld: example.com` to have DDEV use a project TLD that won’t be looked up via DNS. You can do the equivalent with `ddev config --project-tld=example.com`. This also works as a global option in `~/.ddev/global_config.yaml` or running `ddev config global --project-tld=example.com`.
+In `.ddev/config.yaml`, [`use_dns_when_possible: false`](../configuration/config.md#use_dns_when_possible) will make DDEV never try to use DNS for resolution, instead adding hostnames to `/etc/hosts`. You can also use `ddev config --use-dns-when-possible=false` to set this configuration option.
 
 You can also set up a local DNS server like [dnsmasq](https://dnsmasq.org) (Linux and macOS, `brew install dnsmasq`) or ([unbound](https://github.com/NLnetLabs/unbound) or many others on Windows) in your own host environment that serves the project_tld that you choose, and DNS resolution will work fine. You’ll likely want a wildcard A record pointing to 127.0.0.1 on most DDEV installations. If you use dnsmasq, you must configure it to allow DNS rebinding.
```
stasadev, line 17:
```suggestion
You can also set up a local DNS server like [dnsmasq](https://dnsmasq.org) (Linux and macOS, `brew install dnsmasq`) or ([unbound](https://github.com/NLnetLabs/unbound) or many others on Windows) in your own host environment that serves the `project_tld` that you choose, and DNS resolution will work fine. You’ll likely want a wildcard A record pointing to `127.0.0.1` on most DDEV installations. If you use dnsmasq, you must configure it to allow DNS rebinding.
```

docs/content/users/usage/offline.md
```diff
@@ -1,31 +1,24 @@
 # Using DDEV Offline
 
-DDEV attempts to work smoothly offline, and you shouldn’t have to do anything to make it work:
+!!!warning "Make sure to do a `ddev start` on every project you want to use before doing offline"
```
stasadev, line 3:
```suggestion
!!!warning "Make sure to do a `ddev start` on every project you want to use before going offline"
```

stasadev (changes requested):
While we're at it, I see no config example for `dnsmasq`, maybe we can add this somewhere on this page:

```conf
address=/*.ddev.site/127.0.0.1
```

rfay:
I don't think a dnsmasq setup is particular to this page, it's something that people might want to use in a number of situations. We should think about where we could put it. I think that docksal *requires* it and sets it up by default.

stasadev (approved):
That's interesting, I didn't know about Docksal using it under the hood.
I don't know where to put it, it will definitely get lost in the FAQ.

rfay:
I just hate for people to get lost in the weeds on things they don't normally have to fiddle with. If it weren't for that we also could have provided dnsmasq out of the box, then we could resolve everything ourselves, including additional_fqdns, etc. But then we'd have a whole other class of support requests. It's worth thinking about though, for people with complex needs.

jenlampton:
I think the warning is a useful improvement, but it also seems fairly obvious. 

I see the following has been removed:

> However, it does not (yet) attempt to prevent Docker pulls if a new Docker image is required,

Does this mean that ddev does now prevent Docker pulls while offline? How does it know if they are needed or not? How does it know that you are offline? 

I was thinking doing something like `ddev start --offline` might be a good way to tell it not to check for updated images.

All the failures I saw in https://github.com/ddev/ddev/issues/6499 appeared as though ddev was still trying to pull the images, but failing. (Now that we have established that I was running v1.23.2, it is entirely possible this was added in v1.23.3 or v1.23.4)

rfay:
DDEV doesn't pull images if it already has them. If it doesn't have them, it will try to pull them. As far as I can tell, you may have been in the process of upgrading DDEV, and got in a situation where some or none of your images were present in your current Docker environment. Or perhaps you tried changing Docker providers, which is basically the same thing.

PR #6506, opened 2024-08-25
test: Rancher Desktop and other problems when non-default ports used

rfay:
## The Issue

* There were some tests that were creating URLs without respecting the actual URLs in use.  TestPHPConfig, TestPHPOverride. They were doing things like this with app.GetHostname(), etc:
`	_, _ = testcommon.EnsureLocalHTTPContent(t, "http://"+app.GetHostname()+"/phpinfo.php", `max_input_time</td><td class="v">999`, 60)`
* Those tests sometimes may fail on Rancher Desktop as they do on Colima/Lima, so stop testing there.
* TestHTTPSRedirection couldn't work properly with non-default ports. Adjusted it.

### DIFF
--- a/cmd/ddev/cmd/ssh_test.go
+++ b/cmd/ddev/cmd/ssh_test.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/nodeps"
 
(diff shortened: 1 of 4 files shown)


rfay:
I got myself in trouble working on tests in 3 different PRs and finding the remaining errors in the other two while working, etc. So I went ahead and pulled and will have to deal with the results. The two that I pulled were clean and mostly clean.

PR #6509, opened 2024-08-26
test: fix  TestExtraPortExpose with correct URL

rfay:
## The Issue

In https://buildkite.com/ddev/macos-rancher-desktop/builds/957#01918e49-185d-4044-bcf0-8db01bb419df/10062-10136 I see that there is still some erroneous logic in tests when the primary URL may have a port.

> failed to get hit url https://testpkgdrupal10.ddev.site:33003:3000/testfile.html, out=, resp=<nil> err=Get "https://127.0.0.1:3000/testfile.html": dial tcp 127.0.0.1:3000: connect: connection refused

Note `https://testpkgdrupal10.ddev.site:33003:3000/testfile.html`

Add BaseURLFromFullURL() to get the base URL from any URL, without the port.

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -1,7 +1,9 @@
 package netutil
 
 import (
+	"fmt"
 	"net"
+	"net/url"
 	"os"
 	"slices"
 	"syscall"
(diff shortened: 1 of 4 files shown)


pkg/netutil/netutil.go
```diff
@@ -92,3 +94,13 @@ func GetLocalIPs() ([]string, error) {
 
 	return localIPs, nil
 }
+
+// BaseURLFromFullURL returns the base url (http://hostname.example.com) from a URL, without port
+func BaseURLFromFullURL(fullURL string) string {
```
stasadev, line 99: Side note: this will be very useful to use when we have `ddev launch` written in go.

stasadev (approved):
Looks good to me.

PR #6510, opened 2024-08-27
fix: don't touch `.ddev` dir in every `ddev` command, avoid overwriting `.ddev/.gitignore` if unchanged, fixes php-perfect/ddev-intellij-plugin#361

stasadev:
## The Issue

- https://github.com/php-perfect/ddev-intellij-plugin/issues/361#issuecomment-2310857181

1. Adds more commands/flags that you can run without Docker available:
    1. `-v` flag is the same as `--version`
    2. you can run `ddev` without args without Docker, and it's the same as running it with the help flags I've also added.
2. `.ddev/.gitignore` file is removed/created each time you run any command, it's fixed now, it will be repopulated only if something should be changed.

1. `.ddev/.gitignore` file is not updated when there is not changes:
	```
	ddev config --auto
	stat .ddev/.gitignore # remember stats
	ddev config --auto
	stat .ddev/.gitignore # stats should be the same
	```

2. `.ddev/.gitignore` file is not updated when there are some changes:
	```
	ddev config --auto
	echo '/test' >>.ddev/.gitignore
	stat .ddev/.gitignore # remember stats
	ddev config --auto
	stat .ddev/.gitignore # stats should be updated
	```

3. `ddev --version` doesn't touch `.ddev`:
	```
	rm -rf .ddev/providers
	ddev --version
	ls -l .ddev/providers # should be not found
	ddev start
    ls -l .ddev/providers # should appear
	```

### DIFF
--- a/pkg/ddevapp/utils.go
+++ b/pkg/ddevapp/utils.go
@@ -1,6 +1,7 @@
 package ddevapp
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path"
@@ -226,8 +227,9 @@ type ignoreTemplateContents struct {
 
 // CreateGitIgnore will create a .gitignore file in the target directory if one does not exist.
 // Each value in ignores will be added as a new line to the .gitignore.
-func CreateGitIgnore(targetDir string, ignores ...string) error {
+func CreateGitIgnore(targetDir string, showWarnings bool, ignores ...string) error {
 	gitIgnoreFilePath := filepath.Join(targetDir, ".gitignore")
+	existingContent := ""
 
 	if fileutil.FileExists(gitIgnoreFilePath) {
 		sigFound, err := fileutil.FgrepStringInFile(gitIgnoreFilePath, nodeps.DdevFileSignature)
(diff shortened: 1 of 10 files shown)


pkg/dockerutil/dockerutils.go
```diff
@@ -1904,3 +1904,19 @@ func GetContainerNames(containers []dockerTypes.Container) []string {
 func IsErrNotFound(err error) bool {
 	return dockerClient.IsErrNotFound(err)
 }
+
+// CanRunWithoutDocker returns true if the command or flag can run without Docker.
+func CanRunWithoutDocker() bool {
```
rfay, line 1909: This is a good thing.

rfay:
What `ddev` command is creating the .ddev directory to begin with? It seems to me like the fundamental problem is that ddev should only create that on `ddev config`, and update it on `ddev start` ...

stasadev:
> What `ddev` command is creating the `.ddev` directory to begin with?

`ddevapp.PopulateExamplesCommandsHomeadditions("")` in `init()`.

https://github.com/ddev/ddev/blob/fae659e4d94cb2e70568c420046360e184a50982/cmd/ddev/cmd/root.go#L127-L148

stasadev:
I thought it was important to be there in `root.go` so I didn't look into changing this behavior.

It seems easy to "fix" with additional check for not empty `appName = ""` in `ddevapp.PopulateExamplesCommandsHomeadditions("")`

I'll look into this.

rfay:
Thanks, yeah, it doesn't seem like commands like `ddev --version` have any right to create any files.

rfay:
The commands that have no right to create files may be the same ones that can be run without docker.

AkibaAT:
> The commands that have no right to create files may be the same ones that can be run without docker.

Probably a superset of that. I wouldn't expect something like `ddev describe` to create files either for example. Should it require files that get created during start to report on anything, it should rather abort with a message reflecting that.

stasadev:
> Probably a superset of that. I wouldn't expect something like `ddev describe` to create files either for example.

That's right, this PR does this not only for `ddev --version`, but for everything else as well.

rfay:
https://github.com/php-perfect/ddev-intellij-plugin/issues/361#issuecomment-2327452607 explains how to create the problem with current DDEV.

pkg/ddevapp/utils.go
```diff
@@ -240,50 +242,60 @@ func CreateGitIgnore(targetDir string, ignores ...string) error {
...
 			return nil
 		}
-		// Otherwise, remove the existing file to prevent surprising template results
-		err = os.Remove(gitIgnoreFilePath)
-		if err != nil {
-			return err
+		// Read the existing content for future comparison.
+		if gitIgnoreFileBytes, err := os.ReadFile(gitIgnoreFilePath); err == nil {
+			existingContent = string(gitIgnoreFileBytes)
+		} else {
```
rfay, line 248: I suspect this should be "If we can't read or if it's empty then delete it"

There have been a few odd instances of ~/.ddev/commands/ files going empty, and as a result we delete them if they're found empty.

pkg/dockerutil/dockerutils.go
```diff
@@ -1904,3 +1904,19 @@ func GetContainerNames(containers []dockerTypes.Container) []string {
...
+	}
+	commands := map[string]bool{
+		"-v":        true,
+		"--version": true,
+		"-h":        true,
+		"--help":    true,
+		"help":      true,
+		"hostname":  true,
+	}
+	return commands[os.Args[1]]
```
rfay, line 1921: This will work correctly (if map element not found, it returns the empty value, which is false here) but I think this might be clearer with 

```
val, ok := commands[os.Args[1]]
return ok && val 
```

Maybe not, your call. I had to go look at what happens on non-existent map entry.

pkg/dockerutil/dockerutils.go
```diff
@@ -1904,3 +1904,19 @@ func GetContainerNames(containers []dockerTypes.Container) []string {
 func IsErrNotFound(err error) bool {
 	return dockerClient.IsErrNotFound(err)
 }
+
+// CanRunWithoutDocker returns true if the command or flag can run without Docker.
+func CanRunWithoutDocker() bool {
+	if len(os.Args) < 2 {
+		return true
+	}
+	commands := map[string]bool{
```
rfay, line 1913: There are some interesting variants that I think we can ignore, like `ddev help -h` and `ddev -h help`

And what about all the things like `ddev config -h` and `ddev config --help` ?

pkg/ddevapp/utils.go
```diff
@@ -240,50 +242,60 @@ func CreateGitIgnore(targetDir string, ignores ...string) error {
...
 		Signature:    nodeps.DdevFileSignature,
 		IgnoredItems: generatedIgnores,
 	}
-
-	//nolint: revive
-	if err = tmpl.Execute(file, parms); err != nil {
+	if err = t.Execute(&buf, ignoredItems); err != nil {
 		return err
 	}
+	// Only write the file if the generated content differs from the existing content.
```
rfay, line 286: This is great stuff of course. The next time we need it, it will be nice to generalize it into a function that can do the work on any file instead of focusing on .ddev/.gitignore.

rfay (approved):
This is great, thanks for chasing it and understanding it! 

I made some minor comments, you can pursue them if you're interested and think they're relevant. 

I've always been a little embarrassed that we create all those files all the time, regardless of whether they're already there and are already valid. Here we make one small step forward on that, but we can probably do better in the future.

pkg/ddevapp/utils.go
stasadev: > this should be "If we can't read or if it's empty then delete it"

Okay, done.

pkg/dockerutil/dockerutils.go
stasadev: > what about all the things like `ddev config -h` and `ddev config --help` ?

I added support for Cobra help flag.

rfay:
Pull when ready, thanks!

stasadev:
I noticed another bug in the project directory, there is a warning when there is no `.ddev/commands` folder:

```
$ ls -l .ddev
total 4
-rw-r--r-- 1 stas stas 306 Sep  5 00:38 config.yaml

$ ddev --version
Adding custom/shell commands failed: open /home/stas/ddev/d11/.ddev/commands: no such file or directory
ddev version v1.23.4-32-ga98ade6bb
```

Running the tests again after fix.

PR #6516, opened 2024-08-28
test: apt key expiration management, fixes #5829 [skip buildkite]

### LINKED ISSUE #5829: Container test for expiring Debian package keys no longer works (apt-key), monitor debian repositories and keys
### Expected Behavior

* https://github.com/ddev/ddev/issues/5795 surprised us, and we should have been notified by our own tests
* We would be best to monitor upstream Debian repositories as well, so we don't get surprised by things like https://github.com/ddev/ddev/issues/5620

### Actual Behavior

The test is https://github.com/ddev/ddev/blob/6302f01c85c4e1c5bce9372353e4aa114eef81b8/containers/ddev-dbserver/test/image_general.bats#L15-L31 but it uses apt-key, which doesn't look at /usr/share/keyrings

### Steps To Reproduce

_No response_

### Anything else?

We should be able to rewrite that test.

rfay:
## The Issue

* #5829 

Add a script that uses `gpg` to check expiration dates of both .gpg and .asc files

* Review the test output for the tests (Test Container) to see the keys and their expiration and/or
* In a supported project like ddev-dbserver, ddev-ssh-agent, ddev-webserver

To show success, with evaluation of all the keys:
```
export DDEV_MAX_DAYS_BEFORE_CERT_EXPIRATION=200
make test
```

To show a failure
```
export DDEV_MAX_DAYS_BEFORE_CERT_EXPIRATION=1000
make test
```

I didn't do `ddev-traefik-router` because it's apk based instead of debian
I didn't do `ddev-php-base` because everything in it is in `ddev-webserver`

It's ironic that I'm doing this now, when we're more than 500 days from any expiration, when we had TWO accidental expirations in the last several months. 

Right now, it shows all the keys. I think before pulling I'll change `bats --show-output-of-passing-tests` to remove that, so it doesn't show all the keys and their expiration dates.

### DIFF
--- a/.ci-scripts/linux_arm64_setup.sh
+++ b/.ci-scripts/linux_arm64_setup.sh
@@ -26,7 +26,7 @@ echo "capath=/etc/ssl/certs/" >>~/.curlrc
 
 curl -sSL https://golang.org/dl/go${GO_VERSION}.linux-arm64.tar.gz -o /tmp/go.tgz && sudo rm -rf /usr/local/go && sudo tar -zxf /tmp/go.tgz -C /usr/local
 
-git clone --branch v1.2.1 https://github.com/bats-core/bats-core.git /tmp/bats-core && pushd /tmp/bats-core >/dev/null && sudo ./install.sh /usr/local
+git clone --branch v1.11.0 https://github.com/bats-core/bats-core.git /tmp/bats-core && pushd /tmp/bats-core >/dev/null && sudo ./install.sh /usr/local
 
 # Install mkcert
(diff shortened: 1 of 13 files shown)


rfay:
For review, probably just poke around, read the output in https://github.com/ddev/ddev/actions/runs/10610080993/job/29406883861, look for sanity. 

It's probably OK, and much better than what we (didn't have any more)

stasadev (approved):
Looks good to me.
I like the the output, and I didn't know that `awk` could look so similar to JS code, wow!

rfay:
Before I pull this I need to figure out if I want to show the more verbose expiration information normally, and probably use a better way to do it (instead of just showing output of successful tests).

Issue #6462, opened 2024-08-05
Review docker builds for warnings and errors


rfay:
We have some warnings in Docker image builds. It would be great to chase these and resolve them. Some are a result of `docker buildx build` tightening up. I think there may be one from `docker-compose build` at start time as well.

We should review the build logs and address these.

rfay:
I see this in the ddev-dbserver builds:
```
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "MYSQL_PASSWORD") (line 12)
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "MYSQL_ROOT_PASSWORD") (line 13)
  ```
But in this situation, those aren't actually secrets. I guess they could be in some other situations, but in DDEV they're widely known. The same is true in `ddev-ssh-agent`:

```
 3 warnings found (use docker --debug to expand):
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "SSH_KEY_DIR") (line 10)
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "SSH_AUTH_SOCK") (line 12)
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "SSH_AUTH_PROXY_SOCK") (line 13)
```

Those are *really* not secrets. 

The explanation is https://docs.docker.com/reference/build-checks/secrets-used-in-arg-or-env/ and the build secrets docs are at https://docs.docker.com/build/building/secrets/

So we'll have to ignore these or move to using build secrets I guess. I don't know how it infers what's a secret.

**Comment by stasadev** [comment]
I think we can get around this with:

- https://github.com/moby/buildkit/pull/5410

Where any variable containing `public` is ignored.

```Dockerfile
FROM scratch

ARG PUBLIC_MYSQL_ROOT_PASSWORD=root
ENV MYSQL_ROOT_PASSWORD=${PUBLIC_MYSQL_ROOT_PASSWORD}
```

But it doesn't work yet, as there's no release for `moby/buildkit` with this feature.

rfay:
I think this "works" now in the database builds, but unfortunately, it won't accept the MYSQL_ROOT_PASSWORD because it doesn't have PUBLIC on it. 

[...]

PR #6520, opened 2024-09-02
build: fix FromAsCasing warning from Dockerfile, for #6462 [skip ci]

stasadev:
## The Issue

- #6462
- https://docs.docker.com/reference/build-checks/from-as-casing/

```
#1 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 6)
```

* uses the same casting for `FROM ... AS ...`
* removes extra lines at the end of files
* uses an indent of 4 spaces


rfay (approved):
Excellent, thanks!

IMO we don't need to create new images at this time for changes like this.

stasadev:
Sure, I didn't start building anything here.

PR #6531, opened 2024-09-08
test: experimental use of 1password for test secrets

rfay:
## The Issue

Experiment with using 1Password for secrets required in tests

This first attempt does only TestPlatformPull


rfay:
I didn't mean to target this at ddev/ddev. Experiment ongoing on ddev-test/ddev

PR #6536, opened 2024-09-11
test: Use 1password for test secrets [skip buildkite]

rfay:
## The Issue

We'd like to use 1Password to manage secrets

This is a first step, using 1Password for the testing secrets, like TestAcquiaPull, etc.

Review the test logs. 
- [x] The push-pull tests should run successfully in that one
- [x] Tests like TestPlatformPull should *not* run in the other github tests

## Release/Deployment Notes

These secrets live in the `test-secrets` vault
The `test-secrets` vault allows access to the service account [`tests`](https://team-ddev.1password.com/developer-tools/infrastructure-secrets/serviceaccount/76URGQSBMVEUHPEWVHQRFJ4N3Q) whose auth token is in "1Password Service Account Auth Token: Tests"

The technique is documented in https://developer.1password.com/docs/ci-cd/github-actions/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,9 +67,7 @@ jobs:
 
     env:
       CGO_ENABLED: 0
-      DDEV_ACQUIA_SSH_KEY: ${{ secrets.DDEV_ACQUIA_SSH_KEY }}
       DDEV_NONINTERACTIVE: "true"
-      DDEV_PANTHEON_SSH_KEY: ${{ secrets.DDEV_PANTHEON_SSH_KEY }}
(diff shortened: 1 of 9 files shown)


rfay:
Before pulling this, I removed the related secrets from ddev/ddev Github Actions secrets.

stasadev:
And with a fork PR:

```
Run 1password/load-secrets-action@v2
Error: Authentication error with environment variables: you must set either 1) OP_SERVICE_ACCOUNT_TOKEN, or 2) both OP_CONNECT_HOST and OP_CONNECT_TOKEN.
```

https://github.com/ddev/ddev/actions/runs/10848814835/job/30106737701?pr=6501#step:11:32

PR #6501, opened 2024-08-23
fix: allow overrides for Docker networks, fixes #5706, fixes #6376

stasadev:
## The Issue

- #5706
- #6376

## How This PR Solves The Issue

Allows users to make modifications to the `ddev_default` and `default` networks, but with some limitations, that were allowed before, such as:
- it is now forbidden to change `ddev_default` and `default` networks name.
- it is now forbidden to change the `external` setting for `ddev_default` and `default` networks.

Allows users to add new networks to the Docker services.

Adds `com.ddev.platform` labels to all internal networks.

## Manual Testing Instructions

To test this, use the examples from the related issues description or create a new `.ddev/docker-compose.networks.yaml`:

```yaml
networks:
  dummy:
    name: dummy_name
  ddev_default:
    name: this_name_and_external_false_should_be_ignored_by_ddev
    external: false
  default:
    name: this_name_and_external_true_should_be_ignored_by_ddev
    external: true

services:
  web:
    networks:
      default:
        priority: 1
      dummy:
        priority: 2
```

Start the project and look into `.ddev/.ddev-docker-compose-full.yaml`:
1. It should add new `dummy` network to the `web` service, and change `priority` setting.
2. The `ddev_default` and `default` networks should have the original `name` and `external` settings.
3. The `dummy` and `default` networks must have the `com.ddev.platform` label set.

## Automated Testing Overview

I added tests to the existing `TestMultipleComposeFiles` test.

PR #6537, opened 2024-09-12
fix: Remove obsolete code that fixed up absolute mountpoints

rfay:
## The Issue

In reviewing 
* #6501

I noted some old code that was built around a docker-compose anomaly from long ago

Remove the old adaptive code

Figure out the original bug and make sure it's properly handled now.
* https://youtrack.jetbrains.com/issue/WI-61976
* https://github.com/docker/compose/issues/8503#issuecomment-930969241

## Release/Deployment Notes

This will cause a conflict with #6501 so can go in after that one with modest work.

### DIFF
--- a/pkg/ddevapp/compose_yaml.go
+++ b/pkg/ddevapp/compose_yaml.go
@@ -1,12 +1,10 @@
 package ddevapp
 
 import (
-	"os"
-	"strings"
-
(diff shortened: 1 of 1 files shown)


rfay:
@stasadev you can choose whether to get this one in first and then adjust the other one, or the opposite way. Thanks!

stasadev (approved):
Let's pull this PR.

PR #6540, opened 2024-09-13
test: Make forked PRs skip loading secrets

rfay:
## The Issue

* https://github.com/ddev/ddev/pull/6536#issuecomment-2348824760

With the addition of 1Password secrets to the tests workflow, forked PRs started failing

Don't load secrets on forked PRs


rfay:
I think this does the job @stasadev I'll pull it if you agree.

stasadev (approved):
Thank you, I see it works good.

Just in case, leaving an alternative approach here:

```yaml
if: ${{ matrix.pull-push-test-platforms && env.OP_SERVICE_ACCOUNT_TOKEN != '' }}
```

rfay:
Thanks.  That requires setting the environment to the secret in advance though right?

stasadev:
No additional variables, I reused this:

https://github.com/ddev/ddev/blob/dbc1b8104e7f57ca201b62704a1839288d5066de/.github/workflows/tests.yml#L133

PR #6546, opened 2024-09-16
feat: show warnings from stderr, auto rebuild for web image w/o cache on warnings, fixes #6451, for #5073

### LINKED ISSUE #6451: Show users warnings from stderr when running updates for composer, sqlite, etc

stasadev:
## The Issue

- #6451
- #5073

Adds warnings for:
- mariadb-client
- mysql-client
- postgresql-client
- sqlite3 (Drupal 11)
- composer

Automatically runs `docker-compose build web --no-cache` on warnings if there is Internet available.

1. Disable internet.
2. Run `ddev debug refresh`, it should show warnings.
3. Enable internet.
4. Run `ddev restart`, it should rebuild web container without cache.
    If you run `ddev start` instead, it will show a warning, which will disappear on one more start or restart:
    ```
	Unable to delete sha256:a9ee9b4b2665f78697259e5b6a703ac6d24afb1dd8452c6b8d02ae48167fc0c5: Error response from daemon: conflict: unable to delete a9ee9b4b2665 (cannot be forced) - image is being used by running container cda3b436ea7d
	```
5. Run `ddev restart` again, it should not trigger rebuild without warnings.
6. Repeat these steps for different warnings from this PR.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -990,11 +990,11 @@ ENV N_INSTALL_VERSION="%s"
 			extraWebContent = extraWebContent + "\n" + fmt.Sprintf(`
 ### Drupal 11+ requires a minimum sqlite3 version (3.45 currently)
 ARG SQLITE_VERSION=%s
-RUN ( mkdir -p /tmp/sqlite3 && \
+RUN log-stderr.sh bash -c "mkdir -p /tmp/sqlite3 && \
 wget -O /tmp/sqlite3/sqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && \
 wget -O /tmp/sqlite3/libsqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/libsqlite3-0_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && \
 apt-get install -y /tmp/sqlite3/*.deb && \
-rm -rf /tmp/sqlite3 ) || true
+rm -rf /tmp/sqlite3" || true
 			`, versionconstants.Drupal11RequiredSqlite3Version)
 		}
 	}
@@ -1196,7 +1196,7 @@ RUN (apt-get -qq update || true) && DEBIAN_FRONTEND=noninteractive apt-get -qq i
 		// selecting a branch instead of a major version only.
 		contents = contents + fmt.Sprintf(`
 ### DDEV-injected composer update
-RUN export XDEBUG_MODE=off; composer self-update --stable || composer self-update --stable || true; composer self-update %s || composer self-update %s || true
+RUN export XDEBUG_MODE=off; composer self-update --stable || log-stderr.sh composer self-update --stable || true; composer self-update %s || log-stderr.sh composer self-update %s || true
 `, composerSelfUpdateArg, composerSelfUpdateArg)
 
 		// For Postgres, install the relevant PostgreSQL clients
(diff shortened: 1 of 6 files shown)


stasadev:
This PR is still WIP because docker-compose rebuilds in this PR generate dangling Docker images, which I should cleanup here.

pkg/ddevapp/config.go
```diff
@@ -990,11 +990,11 @@ ENV N_INSTALL_VERSION="%s"
 			extraWebContent = extraWebContent + "\n" + fmt.Sprintf(`
 ### Drupal 11+ requires a minimum sqlite3 version (3.45 currently)
 ARG SQLITE_VERSION=%s
-RUN ( mkdir -p /tmp/sqlite3 && \
+RUN log-stderr.sh bash -c "mkdir -p /tmp/sqlite3 && \
```
rfay, line 993: For next time, this would be a great place to try out the docker HEREDOC technique. Wow, is this hard to read :)

rfay:
Additional testing technique: `docker buildx prune` to force it to build without cache. After that, `ddev start` shows the errors.

rfay (approved):
I tested this with looking for sqlite3 version on a d11 project and I like it quite a lot. I needed `docker buildx prune` to get full value. It confused me that it seemed to be working offline.

Shouldn't we do a wrap-up statement when there are errors, explaining to people that some parts of the project build failed, and especially if they're offline, explain what happened?

rfay (commented):
I didn't completely understand how "dirty" is calculated. If you do a `ddev start` with cache, it recreates the dirty build from docker cache and we see it and do-over?

stasadev (commented):
> Shouldn't we do a wrap-up statement when there are errors, explaining to people that some parts of the project build failed, and especially if they're offline, explain what happened?

Good idea, I'll add it.

> I didn't completely understand how "dirty" is calculated. If you do a ddev start with cache, it recreates the dirty build from docker cache and we see it and do-over?

An interesting situation, at first I added a label to make the build dirty, but then I found that we don't even need that, because we can check if it is dirty by looking if there are any warnings in the `/tmp/ddev-log-stderr-*.txt`.

This is done by running `dockerutil. RunSimpleContainer()`, which checks for output from `log-stderr.sh --show` immediately after `docker-compose build`.

rfay:
I guess I wonder why `/tmp/ddev-log-stderr-*.txt` is still there at startup. Build caching?  It's not part of a layer is it? Or maybe it did get added to a layer.

stasadev:
> why `/tmp/ddev-log-stderr-*.txt` is still there at startup. Build caching?

Because I only delete it if there are no errors. Otherwise, I leave it there to show on the `ddev start`. It is part of the Docker layer.

pkg/ddevapp/ddevapp.go
```diff
@@ -1650,6 +1677,14 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		return err
 	}
 
+	if logStderr != "" {
+		util.Warning(`Some components of the project %s were not installed properly.
+The project is running anyway, but see the warnings above for details.
+If offline, run 'ddev restart' once you are back online.
+If online, check your connection and run 'ddev restart' later.
+If this seems to be a config issue, update it accordingly.`, app.Name)
+	}
```
stasadev, line 1686: Added warning message at the end of `ddev start` and pushed new Docker image.

PR #6547, opened 2024-09-16
docs: fix link to ngrok agent config

rfay:
## The Issue

Ngrok moved their config docs, breaking link check


tyler36:
Confirmed new link is correct.

PR #6549, opened 2024-09-17
build: experimentally change default innodb_page_size to 32768 for Craft CMS monster rows with too many columns

rfay:
## The Issue

Strictly experimental:

* https://github.com/orgs/ddev/discussions/6543

On Craft CMS, it huge field/row/column sizes may require increasing the innodb_page_size

Experiment with this. Note that it will require some additional work as mysql:5.5 and 5.6 don't seems to support this.

Try using with mysql 8.0 on a new project. `ddev mysql` and `SHOW GLOBAL VARIABLES LIKE 'innodb_page_size';`

Try starting an existing project, should stay at default

## Automated Testing Overview

mysql 5.5/5.6 will fail

## Release/Deployment Notes

New projects created this way (or deleted and recreated) will have the new value. It shouldn't affect existing projects.

### DIFF
--- a/containers/ddev-dbserver/files/etc/my.cnf
+++ b/containers/ddev-dbserver/files/etc/my.cnf
@@ -80,6 +80,7 @@ innodb-buffer-pool-size        = 1024M
 ;innodb_file_format=barracuda
 innodb_file_per_table=true
 innodb_doublewrite=0
+innodb_page_size=32768
 
 # LOGGING #
 log-error                      = /var/log/mysqld.err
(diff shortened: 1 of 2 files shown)


brandonkelly:
I would vote against this, because most hosts aren’t going to have the same setting. It’s better to find out you’re running against a database limit while developing locally, then to think everything is fine until you try to deploy to production.

brandonkelly:
…or at least it should be something the developer has to explicitly opt into, knowing the risks.

rfay:
@brandonkelly 's argument is good, and this is probably not something that DDEV should be solving, as suggested also in https://github.com/orgs/ddev/discussions/6543

Closing this for now.

PR #6551, opened 2024-09-18
docs: add Magento tools to docs

cmuench:
Add n98-magerun and n98-magerun2 to list of pre-installed tools in web container.
Only a small doc addition.


rfay (approved):
Awesome, thanks!

PR #6552, opened 2024-09-19
fix: ddev debug test - show customizations in .ddev/nginx if they exist [skip ci]

rfay:
## The Issue

A user has customizations in .ddev/nginx/custom.conf that broke the site, but it didn't show up in `ddev debug test`

Make it show up

Do `ddev debug test` with and without a .ddev/nginx/custom.conf file

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -50,6 +50,11 @@ ddev debug configyaml | grep -v web_environment
 header "existing project customizations"
 grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres 2>/dev/null
 
+if ls .ddev/nginx >/dev/null 2>&1 ; then
+  echo "Customizations in .ddev/nginx:"
+  ls -l .ddev/nginx
+fi
+
 header "installed DDEV add-ons"
 
 ddev add-on list --installed


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -50,6 +50,11 @@ ddev debug configyaml | grep -v web_environment
 header "existing project customizations"
 grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres 2>/dev/null
```
stasadev, line 51: But there is already `.ddev/nginx*`.

Run:
```
mkdir -p .ddev/nginx && touch .ddev/nginx/custom.conf
grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres 2>/dev/null
```
Result:
```
.ddev/nginx/custom.conf
```

I think the problem may be that the user had `#ddev-generated` in their file.

rfay:
Thanks for the careful thinking. I tested and you're right, they must have had that in there.

That raises more questions, of course. 

It is often possible for people to copy the nginx_full/seconddocroot.conf.example to seconddocroot.conf, and not remove the #ddev-generated. Of course... ddev did not generate it, as is the case in all files in .ddev/nginx, because we don't populate that directory. 

This minor change is probably still correct in this exact case because DDEV never creates anything in .ddev/nginx.

In a perfect world, we'd list the files that DDEV creates and check those instead of looking at a wildcard. 

We probably should also look in .ddev/*-build for anything (not just non-ddev-generated) as that's another easy example to do what we're doing here.  And perhaps exclude "*.example" in all cases.  Any docker-compose.*.yaml is a customization, as we don't create those. But an add-on might :) So I think the existing logic is OK for that one.

rfay:
I don't think this does any harm. And I guess I won't add anything else to it at this point.

stasadev (approved):
It looks good.

PR #6555, opened 2024-09-20
refactor: Use __DIR__ instead of dirname(__FILE__)

mxr576:
## The Issue

The code generated by DDEV to include its own settings.php could be simplified like this.


rfay (approved):
I don't see any reason why not. `__DIR__` was introduced in PHP 5.3, so it's not as if we're going to not be able to support it with some PHP version.

stasadev (changes requested):
Let's update it everywhere, I see four hits total, three others:

https://github.com/ddev/ddev/blob/507cfca2508b97786b80e8b8c83ea17f5c0fea20/docs/content/users/quickstart.md?plain=1#L833-L833

https://github.com/ddev/ddev/blob/507cfca2508b97786b80e8b8c83ea17f5c0fea20/pkg/ddevapp/drupal/backdrop/settings.php#L526-L526

https://github.com/ddev/ddev/blob/507cfca2508b97786b80e8b8c83ea17f5c0fea20/pkg/ddevapp/wordpress.go#L99-L99

And two more in wp-config.php:

https://github.com/ddev/ddev/blob/507cfca2508b97786b80e8b8c83ea17f5c0fea20/pkg/ddevapp/wordpress/wp-config.php#L32-L36

rfay:
Thank you @mxr576 !

PR #6562, opened 2024-09-23
build: change to new gitpod image with current ddev, make codespaces work [skip ci]

rfay:
## The Issue

It turns out that we *do* need a new image when each new release is done. 

Gitpod just isn't well-enough behaved to do an apt upgrade when needed

* Bump to new image
* Add the tyler36/ddev-gitpod-setup add-on to the docs

While studying this I figured out how to make codespaces work, based on @mandrasch 's work. It turns out it ignores the package if there's not a postCreateCommand!

So added that to the docs here.


stasadev (approved):
Looks good to me.

This is a really interesting finding about `postCreateCommand` 👍

PR #6565, opened 2024-09-26
feat: Don't force Drush's Xdebug environment variable to 1

weitzman:
Drush launcher (deprecated a long time ago) used to care about this env variable. I dont know why we would have set this to 1 but we did. This env variable is now being repurposed by Drush itself at https://github.com/drush-ops/drush/pull/6119/files. So lets not manage the variable in DDEV anymore.

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -198,7 +198,6 @@ services:
     {{ if not .DisableSettingsManagement }}
     - DRUSH_OPTIONS_URI=$DDEV_PRIMARY_URL
     {{ end }}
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks like a no-brainer to me! Thanks!

hanoii:
Correct me if I am wrong, but wouldn't this affect people running drush before 13.x? Affect in the sense that something that was working for them before without doing anything would stop to work. It just happened to colleague of mine that asked me why his drush commands were not being xdebugg'ed. I searched the release docs for xdebug and that's how I landed on this issue. I am about to submit a PR to ddev docs, just thought of mentioning this.

rfay:
As far as we could tell, that hadn't been doing anything previously.

Issue #5984, opened 2024-03-17
Trim a little fat off of ddev-webserver (locale)


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

ddev-webserver is large.

We can sometimes get rid of some of it. 

The locale collection is particularly large when you use one of the docker image inspectors. We can probably use a lot less locales, or add needed locales as additional layers.

### Describe your solution

Inspect the image and see if we can limit locales or other fat

GuySartorelli:
I wonder if there's a clean way to declare required locales as (either global or project) config, and pull in those locales as layers based on that?

rfay:
We install the locales-all package, with results in 236MB uncompressed. I'm not sure why we install locales-all. 

Although there are a number of tools around for studying these things, the easiest technique for me was to use tar. 

```
docker save ddev/ddev-webserver:<tag> >~/tmp/ddev-webserver.tar
cd ~/tmp/ddev-webserver.tar
tar -xf ddev-webserver.tar
# Find the biggest blob
# Untar the biggest blob
# Examine the contents of the untarred blob for size using a filesystem explorer
```

rfay:
locales-all was added in 
* https://github.com/ddev/ddev/pull/1106
in response to 
* https://github.com/ddev/ddev/issues/822

I think it's clear we can't just remove it. And it would be a breaking change to change how locales are configured and used. But we could do it, would just have to provide completely new configuration, which might not be a bad thing. But it's not a quick fat-trimming exercise.

rfay:
It's been a few years since I tried docker-slim (now slimtoolkit/slim), but I just tried it against the ddev-webserver. It reduced the size by 5x. But, like the last time I tried it, the image was then unusable. For example, during the build process we use `groupadd` in one of the extra layers. But... it had been removed by slim.

rfay:
One thing I'm absolutely going to do: Write an FAQ about ddev-webserver size in the docs, explaining why it's big (because it's for developers, providing what they need). Probably add a tip-of-the-day about it. I definitely think the size needs to be viewed in terms of its value.

rfay:
I don't think we'll get to this in this release cycle, but it's a worthy piece of work.

PR #6570, opened 2024-09-27
refactor: install limited locales by default, provide configuration for additional locales, for #5984

deviantintegral:
## The Issue

Locales are big! Yet in practice, most users will only need a small number of locales. This adds a few hundred MB to images, slowing pulls and builds for everyone.

* https://github.com/ddev/ddev/issues/5984

We replace the locales-all package with locales, and then install just a few likely locales. This shouldn't have any effect on PHP apps because [they use ICU](https://www.php.net/manual/en/intl.requirements.php), not the glibc locale. It's the same with nodejs: https://nodejs.org/api/intl.html#build-with-a-pre-installed-icu-system-icu.

People who want all locales, just use `webimage_extra_packages: [locales-all]`

## ## Manual Testing Instructions

- [x] `ddev start` and `ddev exec locale -a` should show the limited locales
- [ ] Add `webimage_extra_packages: [locales-all]` and `ddev restart` and `ddev exec locale -a` will show all locales.

## Automated Testing Overview

I don't think we need to do anything.

## Release/Deployment Notes

* This change will need to be in release notes and preferably will land in a major release, v1.24.0
* In case this doesn't make progress, https://github.com/ddev/ddev/pull/6570/commits/bfdc677720e3cd458edcc28cfc0bf6ccd24e3460 should be implemented elsewhere. It makes sure that errors in the implementation of the Dockerfile are respected. (It really should be in a different PR anyway)


rfay:
Thanks for this! 

I wonder what impact this may have on the few users. I would love to hear from folks in non En-US environments.

Setting to draft as it needs to push an image and the image tag has to go into versionconstants.go

stasadev:
Thank you!

> I would love to hear from folks in non En-US environments

I used only `en_US` and `C` (for time, paper, measurement) on the host, so I don't have experience here.
But it looks easy enough to install all the locales if necessary, I agree that it's a waste to have them all.

---

We also need to add information about how to install locales to the documentation somewhere:

1. Either install them all:
```
ddev config --webimage-extra-packages locales-all
```
2. Or install some of them by selecting a locale with `ddev exec cat /etc/locale.gen` and creating a file, for example, for the French language:
```
echo "RUN echo 'fr_FR.UTF-8 UTF-8' >> /etc/locale.gen && locale-gen" > .ddev/web-build/Dockerfile.locale
```

---

With DDEV v1.23.4:

```
$ ddev exec locale -a
C
C.utf8
POSIX
aa_DJ
aa_DJ.utf8
aa_ER
aa_ER@saaho
aa_ET
af_ZA
af_ZA.utf8
agr_PE
ak_GH
am_ET
an_ES
an_ES.utf8
anp_IN
ar_AE
ar_AE.utf8
ar_BH
ar_BH.utf8
ar_DZ
ar_DZ.utf8
ar_EG
ar_EG.utf8
ar_IN
ar_IQ
ar_IQ.utf8
ar_JO
ar_JO.utf8
ar_KW
ar_KW.utf8
ar_LB
ar_LB.utf8
ar_LY
ar_LY.utf8
ar_MA
ar_MA.utf8
ar_OM
ar_OM.utf8
ar_QA
ar_QA.utf8
ar_SA
ar_SA.utf8
ar_SD
ar_SD.utf8
ar_SS
ar_SY
ar_SY.utf8
ar_TN
ar_TN.utf8
ar_YE
ar_YE.utf8
as_IN
ast_ES
ast_ES.utf8
ayc_PE
az_AZ
az_IR
be_BY
be_BY.utf8
be_BY@latin
bem_ZM
ber_DZ
ber_MA
bg_BG
bg_BG.utf8
bhb_IN.utf8
bho_IN
bho_NP
bi_VU
bn_BD
bn_IN
bo_CN
bo_IN
br_FR
br_FR.utf8
br_FR@euro
brx_IN
bs_BA
bs_BA.utf8
byn_ER
ca_AD
ca_AD.utf8
ca_ES
ca_ES.utf8
ca_ES@euro
ca_ES@valencia
ca_FR
ca_FR.utf8
ca_IT
ca_IT.utf8
ce_RU
chr_US
ckb_IQ
cmn_TW
crh_UA
cs_CZ
cs_CZ.utf8
csb_PL
cv_RU
cy_GB
cy_GB.utf8
da_DK
da_DK.utf8
de_AT
de_AT.utf8
de_AT@euro
de_BE
de_BE.utf8
de_BE@euro
de_CH
de_CH.utf8
de_DE
de_DE.utf8
de_DE@euro
de_IT
de_IT.utf8
de_LI.utf8
de_LU
de_LU.utf8
de_LU@euro
doi_IN
dsb_DE
dv_MV
dz_BT
el_CY
el_CY.utf8
el_GR
el_GR.utf8
el_GR@euro
en_AG
en_AU
en_AU.utf8
en_BW
en_BW.utf8
en_CA
en_CA.utf8
en_DK
en_DK.iso885915
en_DK.utf8
en_GB
en_GB.iso885915
en_GB.utf8
en_HK
en_HK.utf8
en_IE
en_IE.utf8
en_IE@euro
en_IL
en_IN
en_NG
en_NZ
en_NZ.utf8
en_PH
en_PH.utf8
en_SC.utf8
en_SG
en_SG.utf8
en_US
en_US.iso885915
en_US.utf8
en_ZA
en_ZA.utf8
en_ZM
en_ZW
en_ZW.utf8
eo
es_AR
es_AR.utf8
es_BO
es_BO.utf8
es_CL
es_CL.utf8
es_CO
es_CO.utf8
es_CR
es_CR.utf8
es_CU
es_DO
es_DO.utf8
es_EC
es_EC.utf8
es_ES
es_ES.utf8
es_ES@euro
es_GT
es_GT.utf8
es_HN
es_HN.utf8
es_MX
es_MX.utf8
es_NI
es_NI.utf8
es_PA
es_PA.utf8
es_PE
es_PE.utf8
es_PR
es_PR.utf8
es_PY
es_PY.utf8
es_SV
es_SV.utf8
es_US
es_US.utf8
es_UY
es_UY.utf8
es_VE
es_VE.utf8
et_EE
et_EE.iso885915
et_EE.utf8
eu_ES
eu_ES.utf8
eu_ES@euro
eu_FR
eu_FR.utf8
eu_FR@euro
fa_IR
ff_SN
fi_FI
fi_FI.utf8
fi_FI@euro
fil_PH
fo_FO
fo_FO.utf8
fr_BE
fr_BE.utf8
fr_BE@euro
fr_CA
fr_CA.utf8
fr_CH
fr_CH.utf8
fr_FR
fr_FR.utf8
fr_FR@euro
fr_LU
fr_LU.utf8
fr_LU@euro
fur_IT
fy_DE
fy_NL
ga_IE
ga_IE.utf8
ga_IE@euro
gd_GB
gd_GB.utf8
gez_ER
gez_ER@abegede
gez_ET
gez_ET@abegede
gl_ES
gl_ES.utf8
gl_ES@euro
gu_IN
gv_GB
gv_GB.utf8
ha_NG
hak_TW
he_IL
he_IL.utf8
hi_IN
hif_FJ
hne_IN
hr_HR
hr_HR.utf8
hsb_DE
hsb_DE.utf8
ht_HT
hu_HU
hu_HU.utf8
hy_AM
hy_AM.armscii8
ia_FR
id_ID
id_ID.utf8
ig_NG
ik_CA
is_IS
is_IS.utf8
it_CH
it_CH.utf8
it_IT
it_IT.utf8
it_IT@euro
iu_CA
ja_JP.eucjp
ja_JP.utf8
ka_GE
ka_GE.utf8
kab_DZ
kk_KZ
kk_KZ.rk1048
kk_KZ.utf8
kl_GL
kl_GL.utf8
km_KH
kn_IN
ko_KR.euckr
ko_KR.utf8
kok_IN
ks_IN
ks_IN@devanagari
ku_TR
ku_TR.utf8
kw_GB
kw_GB.utf8
ky_KG
lb_LU
lg_UG
lg_UG.utf8
li_BE
li_NL
lij_IT
ln_CD
lo_LA
lt_LT
lt_LT.utf8
lv_LV
lv_LV.utf8
lzh_TW
mag_IN
mai_IN
mai_NP
mfe_MU
mg_MG
mg_MG.utf8
mhr_RU
mi_NZ
mi_NZ.utf8
miq_NI
mjw_IN
mk_MK
mk_MK.utf8
ml_IN
mn_MN
mni_IN
mnw_MM
mr_IN
ms_MY
ms_MY.utf8
mt_MT
mt_MT.utf8
my_MM
nan_TW
nan_TW@latin
nb_NO
nb_NO.utf8
nds_DE
nds_NL
ne_NP
nhn_MX
niu_NU
niu_NZ
nl_AW
nl_BE
nl_BE.utf8
nl_BE@euro
nl_NL
nl_NL.utf8
nl_NL@euro
nn_NO
nn_NO.utf8
nr_ZA
nso_ZA
oc_FR
oc_FR.utf8
om_ET
om_KE
om_KE.utf8
or_IN
os_RU
pa_IN
pa_PK
pap_AW
pap_CW
pl_PL
pl_PL.utf8
ps_AF
pt_BR
pt_BR.utf8
pt_PT
pt_PT.utf8
pt_PT@euro
quz_PE
raj_IN
rif_MA
ro_RO
ro_RO.utf8
ru_RU
ru_RU.cp1251
ru_RU.koi8r
ru_RU.utf8
ru_UA
ru_UA.utf8
rw_RW
sa_IN
sah_RU
sat_IN
sc_IT
sd_IN
sd_IN@devanagari
se_NO
sgs_LT
shn_MM
shs_CA
si_LK
sid_ET
sk_SK
sk_SK.utf8
sl_SI
sl_SI.utf8
sm_WS
so_DJ
so_DJ.utf8
so_ET
so_KE
so_KE.utf8
so_SO
so_SO.utf8
sq_AL
sq_AL.utf8
sq_MK
sr_ME
sr_RS
sr_RS@latin
ss_ZA
st_ZA
st_ZA.utf8
sv_FI
sv_FI.utf8
sv_FI@euro
sv_SE
sv_SE.iso885915
sv_SE.utf8
sw_KE
sw_TZ
syr
szl_PL
ta_IN
ta_LK
tcy_IN.utf8
te_IN
tg_TJ
tg_TJ.utf8
th_TH
th_TH.utf8
the_NP
ti_ER
ti_ET
tig_ER
tk_TM
tl_PH
tl_PH.utf8
tn_ZA
to_TO
tpi_PG
tr_CY
tr_CY.utf8
tr_TR
tr_TR.utf8
ts_ZA
tt_RU
tt_RU@iqtelif
ug_CN
uk_UA
uk_UA.utf8
unm_US
ur_IN
ur_PK
uz_UZ
uz_UZ.utf8
uz_UZ@cyrillic
ve_ZA
vi_VN
wa_BE
wa_BE.utf8
wa_BE@euro
wae_CH
wal_ET
wo_SN
xh_ZA
xh_ZA.utf8
yi_US
yi_US.utf8
yo_NG
yue_HK
yuw_PG
zh_CN
zh_CN.gb18030
zh_CN.gbk
zh_CN.utf8
zh_HK
zh_HK.utf8
zh_SG
zh_SG.gbk
zh_SG.utf8
zh_TW
zh_TW.euctw
zh_TW.utf8
zu_ZA
zu_ZA.utf8
```

With this PR:

```
$ ddev exec locale -a
C
C.utf8
POSIX
en_US.utf8
```

rfay:
The key to this one is understanding how people have been using this. I know it was added for a reason, and many people outside en-us use DDEV. But many of them don't care or don't use it.

garvinhicking:
Thanks for pointing me/us to this. In fact, the development community of TYPO3 is large european, so we do have a lot of developers using: de(at/CH), fr, dk, no and some others. Again, for us it's really helpful to have all kind of contributors and test installations be able to utilize different locales easily, without any extra packages.

I haven't tested this since a longer time, but at least with PHP 7 I had the need to install system locales to be able to properly use TYPO3s multilanguage features, which depend on the system locale to be available. The language picker only shows the locales existing on the server.

Most webservers where TYPO3 is hosted also utilize the system locale languages, so people are used to this working in usual setups. It would feel a bit awkward to have it not work "out of the box" in a development environment, and would create friction. I think this one is another good example of where ddev boasts with convenience over explicit configuration.

($0.02)

rfay:
> It seems to me that it is unlikely that anyone who uses ddev doesn't have at least a working understanding of English - all the docs, commands etc are written in English. (whether it should all be internationalized is another topic).

The problem here is the behavior of the project (inside the container) rather than the user's preferred locale. I agree with you that users can cope with the default locale (and the `ddev` command itself on the host has no customization for this. But the behavior of the project when serving websites is what might be affected here.

I agree that something like the @stasadev suggestion might be completely adequate. I asked the TYPO3 folks to speak up here, as they may have some projects that depend on non-default locales.

sbuerk:
TYPO3 does not only use ICU. Even in the current main (upcoming v13 LTS) we still have real `setlocal()` codes and even reading available system locales, which are currently really essential. 

Most TYPO3 agencies may reside and provide projects in the DACH region countries or around EUROPE, but some has really multilanguage/multilocale projects.

Even in non-TYPO3 project this may be an issue. Simply downstripping that to a single local should thus be considered breaking as it may hurt a lot people silently simply by an ddev update.

I understand the general approach to minimize the images and still want to question what ddev should be for users. Until now it was a really good local kickstart project for all developers around the world and having basic stuff and beeing happy on board. Special needs needed special addressment, but in the majority of project it simply worked.

This change would now enforce agengies to update this througout all projects in a hurry if that is done silently.

Literally, downloading the prebuild image with all locals takes lesser time the when developers have to pull them on startup of projects and having long web image builds.

Combining this with the ghostscrip removal discussion this increases a startup for many project drastically.

Is it that worth ? Does `en_US`developer really only work wrok with that one locale ? 

I would love to hear that reconsidered, but at least ensure a grace period informing ddev users beforehand with the next version that it will dropped at some point .. next minor version and not patchlevel bugfix version or similar.

At least fire up the press/socialmedia information chain weeks beforehand so people can prepare their projects where needed instead of running into isssues to do so when no time is available and not so firm developers needs the help from other collegues to fullfill this correctly in projects.

Even if "PHP" itself does not use system locals directly, that does not mean that PHP application does not rely on that, specially for cli command operations.

I fear that many people will resort to pure docker system builds with pre-build images if project startups on lesser powerfull system takes a lot of time to install and update all these thins now getting dropped from ddev and has been a strength of ddev.

> Note not a daily ddev user myself, but looking at collegues and developer in my bubble I fear that this may become a real "ranting shitstorm".

In the end - it's up to you, just my five cents.

(or at least narrow it down to at least some more locales)

* en_US.utf-8
* en_GB.utf-8
* de_DE.utf-8
* de_AT.utf-8
* fr_FR.utf-8
* ru_RU.utf-8
* ....

In my last company we had a dozen of projects requiring 20 or 25 different locales in multiple projects ....

sbuerk:
Full for some verrsions until a specific image has been selected in the configuation and at some point switching to "lean/small" image (grace period).

I mean - all this is the dna of ddev and that one selling point that even started without and (web)server / linux administration experience could setup local project setups for a team. Without all that even simpler setups requires these experience - which would require experienced dev ops/admins to take care of that again (and ending up using docker-compose directly again with fully pre-build images specific to projects / company base images).

As already written to randy, in end it is not my decision and I only tried to emphasize whcih this move would mean to a lot of people.

I mean ... why not dropping nvm/nodejs also ? We usual use dedicated images anyway for that and others eventualy to - why should that be in the image ? 

At some point we will and on a pure pure base os image.

I guess, it should be really considered what ddev is and stands for. In my eye it is not about beeing a slim and narrowed down image. It's more about being a good, direct and easys starting poing for a lot of casual use-cases around the globe.

sbuerk:
> Whatever happens, the default should probably remain as stable/unchanged/full as possible - "power users" can more easily switch to a "lean" build (and a config flag could be provided to make it easy to do so).

Maybe I missed that - and literally, that was my point. The full fledged happy images should stay for easy onboarding. Having additional spezialized images may be good thing. Just not the way around, in my eyes.

I can live with all of that because I'm simply having all these things anway directy on my host systems. Just wanted to share the experiences not-so firm people had with ddev and what they would mean and what impact it may have on project and agencies (with getting the intention wrong as it seems).

Lean/smaller/fine splitted images should be opt-in then - do I get that right ?

deviantintegral:
> many people outside en-us use DDEV

Yep, I'm an `en-CA` and had to take a breath before committing code with `en-US`!

Going down to the top 5-10 locales instead of _everything_ could be a good way to improve this.

Something else to keep in mind is that Debian itself doesn't default to all locales. When you install, you get a dialog to select which ones you want.

> Which one should be the default, I'm not sure. Probably full

100% agreed on this! Let's only remove things from ddev's default images that are truly stale and have minimal impact on users.

> I don't think anyone wants to implement anything that would negatively affect anyone. Perhaps there's a language barrier?

I'm glad to add context to this. Our team has benefitted greatly from ddev's "batteries included" philosophy, which is why these PRs and issues very much request feedback! None of these are anywhere close to planned work - it's something we've talked about at Lullabot for a while, @rfay mentioned off-hand that he _thought_ we may be able to reduce locales, and I had a Friday afternoon where I needed a palate cleanser of writing code :D.

Even if we end up making no changes to locales in the image, I do think we should add some docs. Right now, ddev doesn't have the word "locale" at all in docs, and if including every locale is decided to be a worthwhile feature then we should promote it!

rfay:
I'm interested in specifics of how using different locales in the web container changes website behavior.

tyler36:
I have an addon that I use to Japanese-ify the DDEV web container. I created it to help co-workers, but do not use it.

rfay:
@tyler36 I'd appreciate if you'd ask them if it affects actual website behavior. I can see how it would be useful to them with `ddev ssh`.

rfay:
Asked about this for broader responses on Social, https://x.com/randyfay/status/1842937026884075843 etc

kaystrobach:
Go for it. Maybe add an option to switch it, but i doubt someone is using it.

koehnlein:
Will there be an easy to use configuration similar to `webimage_extra_packages` to install additional locales? Then I would be fine with it.

If this is not planned and users should install locales with Dockerfiles or similar, this would be to complicated for many users.

rfay:
Thanks so much @koehnlein - Could you say what locale you use in the Debian image and what impact it has for you? 

We don't want to do anything disruptive. If we removed locales-all, then an individual locale would be added with something like this, but would not be as simple as adding a package:
```
echo "RUN echo 'fr_FR.UTF-8 UTF-8' >> /etc/locale.gen && locale-gen" > .ddev/web-build/Dockerfile.locale
```

One option is to add a project-level configuration for locale, `locale: fr_FR.UTF-8 ` or something. But then there are people who have mentioned having needed multiple locales in one project in the past (still do-able with the Dockerfile approach above.)

kaystrobach:
Hoe sbout making the option an array. Once it’s there add the corresponding locale. If it’s configured globally, merge both.

and i need to correct me. We used the locales to output dates in php with the applied locale

rfay:
Agreed that it should be a list, thanks @kaystrobach - Does php actually use the system locales, or does it have its own techniques?

Could you provide a trivial PHP script that demonstrates the appropriate use of locale and its dependence on the system locale? We can incorporate that into an automated test.

tyler36:
> I'd appreciate if you'd ask them if it affects actual website behavior. I can see how it would be useful to them with ddev ssh.

They said they didn't know of any difference.

rfay:
IMO the consensus here is to 
* Include several key locales by default
* Make it easy to add locales in the DDEV config

I have added a set of base locales (open to discussion and addition) and pushed the new image and image tag.

I updated the OP to show the several tasks still to be accomplished.

Conversation is still very welcome.

This trims a nontrivial 55MB compressed off of the image, more than 10%:

image

schliesser:
The first time I had contact with Locales was a few years back, when the date formatting suddenly was in MM/DD/YYYY instead of DD.MM.YYYY for my german website on the webserver. Due to the fact that ddev delivered all locales it just worked fine locally. Therefore, I was able to trace it down to the locales - I've never heard about before - and was able to request the server admin to add the necessary locales to the webserver.

I think ddev should keep this user friendly approach and deliver all locales by default. The UX is worth way more than 50mb.
For local development I really don't care much about the image size. When it comes to CI, I see the point, but for CI there's almost every time some custom image needed.

If you decide to go with only en-US, then there needs to be a config array to add all the needed locales per project. I've got some projects running in de-DE only and some using 10-20 locales. I rarely use en-US only.

kaystrobach:
```php
<?php


$today = new \DateTime();

setlocale(LC_TIME, 'de_DE');

echo @strftime("%A %e %B %Y") . PHP_EOL; // deprecated
echo $today->format("F") . PHP_EOL;

setlocale(LC_TIME, 'en_US');

echo @strftime("%A %e %B %Y") . PHP_EOL; // deprecated
echo $today->format("F");


```

This behavior is noted in the php docs on [DateTime::format](http://www.php.net/manual/en/datetime.format.php): This method does not use locales. All output is in English.

* strftime is problematic, but deprecated ...
* https://www.php.net/manual/en/class.numberformatter.php needs locales as well

I would still opt for the reduction of the locales with a BC.

To make the transition smooth, I suggest the following way:

1. add 2 config options: 
  * install all locales
  * install specific locales
2. for existing projects / installs ask the users on the first run after the update, if the install all locales option should be set to true
  * first question for global setting (should not ask for local one again)
  * if not globally set, ask for the local project
3. if a new project is setup try to get the locale settings from the host
  * with a german host system ddev should add german to the list of locales to install, etc.

jonaseberle:
Thank you for finding a good solution here. There can always be programs on a webserver that require a specific locale. So far TYPO3 relies on it.
But reducing image size is also an important goal, so thank you all for working together here.

Regarding PHP, I had to play around with it. The OP was of course ;) right: `intl` functions don't need the system locale.

For sorting, date and number format that relies on `setlocale()` the system locales are required. 

I'll leave Kay's extended playground here if someone else wants needs a refresher like me...

```php
<?php

function sysLocaleTests($locale)
{
    echo PHP_EOL . "----- setlocale() " . $locale . " -----" . PHP_EOL;
    if (false === setlocale(LC_ALL, $locale)) {
        echo "[!!] Could not set locale $locale" . PHP_EOL;
        return;
    }
    $locale = localeconv();
    echo number_format(
            1234567.89,
            $locale['frac_digits'],
            $locale['decimal_point'],
            $locale['thousands_sep']
        ) . PHP_EOL;

    $today = new DateTime();
    echo @strftime("%A %e %B %Y") . PHP_EOL; // deprecated
    echo $today->format("F") . PHP_EOL;

    $sort = ["nap", "nop", "ñop", "á", "abc", "äbc", "b"];
    sort($sort, SORT_LOCALE_STRING);
    echo implode(', ', $sort) . PHP_EOL;
}

function intlTests($locale)
{
    echo PHP_EOL . "----- intl " . $locale . " -----" . PHP_EOL;

    echo (new NumberFormatter($locale, NumberFormatter::DECIMAL))->format(1234567.89123456789) . PHP_EOL;
    // that's fun ;)
    echo (new NumberFormatter($locale, NumberFormatter::SPELLOUT))->format(1234567.89123456789) . PHP_EOL;

    $today = new DateTime();
    $formatter = new IntlDateFormatter($locale, IntlDateFormatter::SHORT, IntlDateFormatter::NONE);
    echo $formatter->format($today) . PHP_EOL;
    $formatter = new IntlDateFormatter($locale, IntlDateFormatter::LONG, IntlDateFormatter::NONE);
    echo $formatter->format($today) . PHP_EOL;

    $sort = ["nap", "nop", "ñop", "á", "abc", "äbc", "b"];
    (new Collator($locale))->sort($sort);
    echo implode(', ', $sort) . PHP_EOL;
}

foreach (['en_US.utf8', 'es_ES.utf8', 'de_DE.utf8'] as $locale) {
    sysLocaleTests($locale);
    intlTests($locale);
}
```

jonaseberle:
I have a little doubt about `ddev config global --extra-locales` if that is meant to be merged with the local project configuration.

I prefer to have reproducible test/dev systems for all developers and testers as far as possible. Anything that is not in the local project configuration should not be taken into account in my opinion.

rfay:
> I have a little doubt about ddev config global --extra-locales if that is meant to be merged with the local project configuration.

Thanks @jonaseberle - The project config would override the global config (not merge). (Any time we add just project config somebody asks for it to be global as well, I'm sure that would be the case here as well)

josefglatz:
Would be cool to not remove languages. DDEV as dev environment should be running all over the world without installing locales. With that, also testing multilanguage apps is possible without any need to think about it.

jonaseberle:
> Would be cool to not remove languages. DDEV as dev environment should be running all over the world without installing locales. With that, also testing multilanguage apps is possible without any need to think about it.

In that case the admin knows which **system** locales are necessary and can add them to the DDEV project config.

bmack:
I recommend to not ship additional locales, but just introduce a new setting like this:
```
webimage_extra_packages: [screen, poppler-utils]
```

and we then have
```
webimage_locales: [de, de_AT, fr, fr_FR]
```

fe-hicking:
(If it really has to be removed, please also add a "*" setting then so we don't have to think about it when implementing new locales into an existing project)

rfay:
@fe-hicking the proposal in the updated OP is to have an "all" setting, which would be like current.

rfay:
Working approach here:
* Create a list of all locales with both their short name (locale -a) and their long name (that goes into /etc/local3.gen)
* Allow people to use the short name in extra_locales
* Look up the short name to get the long name (with the character code)
* Allow people to look up possible locales with an easy `ddev exec`

rfay:
I've studied this and now have:
* [gzipped file with all locales](https://github.com/ddev/ddev/blob/96d9e3ad5bf3d2e747342b87cc8b0a6524a6d411/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/all_locales.txt.gz). This has 3 fields per line. The first is the locale name as would be shown in `locale -a`, like `en_US.utf8`. The second is the base name of the locale, like `en_US`. The third field is the character set, like UTF-8. 
* [awk generator](https://github.com/ddev/ddev/blob/96d9e3ad5bf3d2e747342b87cc8b0a6524a6d411/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/locale.awk)

Generate with `locale -a -v | awk -f locale.awk`

All available locale bases are in /usr/share/i18n/locales

/etc/locale.gen takes like with
```
basename codeset
en_US UTF-8
```

Unfortunately, the normal way to specify a locale is with something like `en_US.utf8`, so that's why we use that as the first field.

rfay:
Please review @deviantintegral , thanks!  

image

containers/ddev-webserver/Dockerfile
```diff
@@ -16,6 +16,7 @@ ENV CAROOT=/mnt/ddev-global-cache/mkcert
 # BUILDPLATFORM is the platform of the build host (e.g. linux/amd64)
 ARG TARGETPLATFORM
 ARG BUILDPLATFORM
+ENV DEFAULT_LOCALES="en_CA.UTF-8 UTF-8\nen_US.UTF-8 UTF-8\nen_GB.UTF-8 UTF-8\nde_DE.UTF-8 UTF-8\nde_AT.UTF-8 UTF-8\nfr_CA.UTF-8 UTF-8\nfr_FR.UTF-8 UTF-8\nja_JP.UTF-8 UTF-8\nru_RU.UTF-8 UTF-8"
```
stasadev, line 19:
```suggestion
ARG DEFAULT_LOCALES="en_CA.UTF-8 UTF-8\nen_US.UTF-8 UTF-8\nen_GB.UTF-8 UTF-8\nde_DE.UTF-8 UTF-8\nde_AT.UTF-8 UTF-8\nfr_CA.UTF-8 UTF-8\nfr_FR.UTF-8 UTF-8\nja_JP.UTF-8 UTF-8\nru_RU.UTF-8 UTF-8\n"
```

stasadev, line 19: I don't see `$DEFAULT_LOCALES` in the resulting image (why? maybe `\n` is to blame), but let's use `ARG` to be sure. And I added a new line at the end, so `cat /etc/locale.gen` looks better.

```suggestion
ARG DEFAULT_LOCALES="en_CA.UTF-8 UTF-8\nen_US.UTF-8 UTF-8\nen_GB.UTF-8 UTF-8\nde_DE.UTF-8 UTF-8\nde_AT.UTF-8 UTF-8\nfr_CA.UTF-8 UTF-8\nfr_FR.UTF-8 UTF-8\nja_JP.UTF-8 UTF-8\nru_RU.UTF-8 UTF-8\n"
```

stasadev (approved):
Looks good to me.

containers/ddev-webserver/Dockerfile
rfay: I'm pretty sure the reason you don't see DEFAULT_LOCALES in the resulting image is that it's from a different stage. It's an ENV there in ddev-webserver-base, but the final build is ddev-webserver, 2 stages later. ENVs don't survive those stages. 

The `\n` is actually useful here because I did a printf instead of `echo`. The echo would have added its own.

rfay: I do sure like using ARG vs ENV, will eventually learn to do that :)

deviantintegral:
This is also working well for me. The uncompressed size of the image compared to 1.23.5 is nearly 200MB smaller. [I ran some jobs in CI](https://github.com/Lullabot/drainpipe/pull/730) to [compare performance](https://github.com/Lullabot/drainpipe/pull/729), and installing ddev is ~5% faster. Nice!

rfay:
Congratulations and thanks!

PR #6575, opened 2024-09-30
build: add bash-completion to standard packages

rfay:
## The Issue

Some people use git inside the container, and I note that git autocompletion doesn't work there.

Because bash completion is not installed, it's likely that plenty of other autocompletes don't work

Add bash-completion.

Check out something inside the container and use git completion, especially of branches. For example, clone https://github.com/ddev/ddev and `git checkout 202409` and hit tab to get suggested branches.

## Release/Deployment Notes

This seems to add less than 1 MB to the image size, see https://hub.docker.com/r/ddev/ddev-webserver/tags

image


ultimike:
Woot!

PR #6580, opened 2024-10-02
feat: rename refresh to rebuild in `ddev debug refresh`, allow to build different services, add cache flag

stasadev:
## The Issue

`ddev debug refresh` is not flexible enough. It runs rebuild without cache only for the web service.

- adds `--service=name` and `--all` flag to build specified service.
- adds `--cache` flag that will not add `--no-cache` to the underlying `docker-compose build` command.
- adds `--progress plain` to `docker-compose build`, because it was squashed by default.

```
ddev debug refresh # this command is the same as it was before
ddev debug refresh --cache # should show cached output
ddev debug refresh --all # should rebuild all services (if you have web, db, redis: all of them should be rebuilt)
ddev debug refresh --service db # the same as usual command, but for the db service
```

### DIFF
--- a/cmd/ddev/cmd/debug-refresh.go
+++ b/cmd/ddev/cmd/debug-refresh.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	exec2 "github.com/ddev/ddev/pkg/exec"
 	"github.com/ddev/ddev/pkg/globalconfig"
+	"strings"
 	"time"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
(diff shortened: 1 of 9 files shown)


rfay:
Wow, exciting!

cmd/ddev/cmd/debug-refresh.go
```diff
@@ -12,12 +13,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	buildAll bool
```
rfay, line 17: In general, I don't like these globals for this, and prefer checking the flags as we do in most places. Of course, this is just in `ddev debug` so not important at all, so no big deal.

cmd/ddev/cmd/debug-refresh.go
```diff
@@ -12,12 +13,17 @@ import (
...
+	service  string
+)
+
 // DebugRefreshCmd implements the ddev debug refresh command
 var DebugRefreshCmd = &cobra.Command{
 	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
```
rfay, line 23: `refresh` was a great name for this when it was created, but how about changing it to `rebuild` with `refresh` being an alias?

rfay (approved):
Super nice, can't wait to have this!

cmd/ddev/cmd/debug-refresh.go
stasadev: The variable here is only needed because the flag has a short and long form, I just copied the logic from some other command.

cmd/ddev/cmd/debug-refresh.go
stasadev: Okay, `rebuild` gives more context, I'll change it.

stasadev:
> `refresh` was a great name for this when it was created, but how about changing it to `rebuild` with `refresh` being an alias?

Renamed, added the alias, and updated the documentation.

PR #5266, opened 2023-08-16
fix: add dedicated ddev-traefik-router, fixes #5239, fixes #5209

rfay:
## The Issue

* #5239 
* #5209 

In DDEV v1.22.0 the default router was changed to traefik, and a build stage was added. There were a couple of side-effects in unusual cases in the above issues.

## How This PR Solves The Issue

Switch to a pre-built ddev-traefik-router image so we have control of the build process.

## Manual Testing Instructions

- [x] Just use it.
- [x] Make sure the ~/.ddev/router-build directory is empty after using it

## Automated Testing Overview

- [x] Needs container tests for ddev-traefik-router

<a href="https://gitpod.io/#https://github.com/ddev/ddev/pull/5266"></a>

PR #6581, opened 2024-10-02
fix: remove unused file for router, for #5266

stasadev:
## The Issue

Follow up for:

- #5266

Removes `router_Dockerfile_template` file, that is not used anywhere.


rfay:
This got lost when we moved to a completely pre-built traefik router, instead of having it built at start. That approach got too many failures.

PR #6582, opened 2024-10-02
build: fix Docker's InvalidDefaultArgInFrom warning

stasadev:
## The Issue

When running `ddev debug refresh`:

```
WARN: InvalidDefaultArgInFrom: Default value for ARG $BASE_IMAGE results in empty or invalid base image name (line 6)
```

https://docs.docker.com/reference/build-checks/invalid-default-arg-in-from/

Adds defaults, that will never be used, because we always pass the `ARG BASE_IMAGE`.

This PR goal is to silence this warning.

Run `ddev debug refresh`

### DIFF
--- a/containers/ddev-dbserver/Dockerfile
+++ b/containers/ddev-dbserver/Dockerfile
@@ -1,5 +1,5 @@
-ARG BASE_IMAGE
-ARG DB_PINNED_VERSION
+ARG BASE_IMAGE="scratch"
+ARG DB_PINNED_VERSION="latest"
 FROM ${BASE_IMAGE}:${DB_PINNED_VERSION}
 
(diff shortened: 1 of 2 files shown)


rfay (commented):
Thanks for doing this. To get tested of course, it needs to have db images pushed., right? Maybe we don't need to do that if you want to just see how the container tests work.

stasadev:
> To get tested of course, it needs to have db images pushed., right?

Right, and I built one image locally. No need to run it here.

rfay:
Looks happy in https://github.com/ddev/ddev/actions/runs/11147950246/job/30983464615?pr=6582, although I didn't read every line.

rfay (approved):

Issue #6583, opened 2024-10-02
`ddev export-db > file` contains DDEV startup output if Docker container is not currently running, making invalid dump


cooperka:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
I'm uncomfortable posting all these detailed diagnostics publicly online,
so I'll at least paste the relevant version numbers, and I can include more
info if you fail to reproduce it:

======== DDEV version ========
 ITEM             VALUE                                    
 DDEV version     v1.23.4                                  
 architecture     arm64                                    
 cgo_enabled      0                                        
 db               ddev/ddev-dbserver-mariadb-10.11:v1.23.4 
 ddev-ssh-agent   ddev/ddev-ssh-agent:v1.23.4              
 docker           27.2.0                                   
 docker-api       1.47                                     
 docker-compose   v2.29.1                                  
 docker-platform  docker-desktop                           
 global-ddev-dir  /Users/cooperka/.ddev
 mutagen          0.17.2                                   
 os               darwin                                   
 router           ddev/ddev-traefik-router:v1.23.4         
 web              ddev/ddev-webserver:v1.23.4              
```
</details>

### Expected Behavior

- DDEV should export properly using any variation of the [instructions](https://ddev.readthedocs.io/en/stable/users/usage/commands/#export-db)

### Actual Behavior

* Even though these files should all be identical, the first one `export-offline.sql.gz` will contain console logs that corrupt the file (example below)

### Steps To Reproduce

1. Set up a DDEV project, for example by following Drupal's [new project startup guide](https://www.drupal.org/docs/official_docs/local-development-guide)
1. Turn **off** the new Docker container that was created
1. Run `ddev export-db > export-offline.sql.gz` (which will start up the container while exporting)
1. Run `ddev export-db > export-online.sql.gz` (which will use the already-started container)
1. Run `ddev export-db --file=export-file.sql.gz` (which will work regardless)
1. Open the exported files in a text editor

### Anything else?

Example logging info at the top of the file:

```
[...]

PR #6584, opened 2024-10-04
docs: remove import/export 'redirect operator' examples

cooperka:
### Reason for change

Neither of these examples works reliably, and they should not be recommended to the average user.

### Example of export issue

See 
* #6583.

### Example of import issue

Using `ddev import-db < ../db.sql.gz` (note that the file is gzipped) drops the database and then fails with something like:

```
grep: (standard input): binary file matches
Error: failed to import database 'db' for project: failed to import database: composeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-project docker-compose -f /Users/cooperka/project/.ddev/.ddev-docker-compose-full.yaml exec -T db bash -c set -eu -o pipefail && mysql -uroot -proot -e "DROP DATABASE IF EXISTS db; CREATE DATABASE IF NOT EXISTS db; GRANT ALL ON db.* TO 'db'@'%';"  2> >(grep -v 'Deprecated program name.' >&2)  && perl -p -e 's/^(CREATE DATABASE \/\*|USE `)[^;]*;//' | mysql db  2> >(grep -v 'Deprecated program name.' >&2) ', action='[exec -T db bash -c set -eu -o pipefail && mysql -uroot -proot -e "DROP DATABASE IF EXISTS db; CREATE DATABASE IF NOT EXISTS db; GRANT ALL ON db.* TO 'db'@'%';"  2> >(grep -v 'Deprecated program name.' >&2)  && perl -p -e 's/^(CREATE DATABASE \/\*|USE `)[^;]*;//' | mysql db  2> >(grep -v 'Deprecated program name.' >&2) ]', err='exit status 1', stdout='', stderr='grep: (standard input): binary file matches'
stdout:
stderr: grep: (standard input): binary file matches
```

It's not clear in the docs that the redirect operator ONLY works for uncompressed files, unlike every other example which works for both compressed and uncompressed files.

### DIFF
--- a/docs/content/users/usage/commands.md
+++ b/docs/content/users/usage/commands.md
@@ -624,9 +624,6 @@ ddev export-db --gzip=false --file /tmp/db.sql
 # Dump and compress the current project’s `foo` database instead of `db`
 ddev export-db --database=foo --file=/tmp/db.sql.gz
 
-# Output the current project’s database and use `>` to write to `/tmp/db.sql.gz`
-ddev export-db > /tmp/db.sql.gz
-
 # Dump my-project’s database, without compressing it, to `/tmp/my-project.sql`
 ddev export-db my-project --gzip=false --file=/tmp/my-project.sql
 ```
@@ -827,12 +824,6 @@ ddev import-db --file=.tarballs/db.sql.gz
 
 # Import the compressed `.tarballs/db.sql.gz` dump to a `other_db` database
 ddev import-db --database=additional_db --file=.tarballs/db.sql.gz
-
-# Import the `db.sql` dump to the project database
-ddev import-db < db.sql
-
-# Import the `db.sql` dump to the `my-project` default database
-ddev import-db my-project < db.sql
 
 # Uncompress `db.sql.gz` and pipe the result to the `import-db` command
 gzip -dc db.sql.gz | ddev import-db


rfay:
The redirect operator works fine for compressed tings.

`ddev export-db >junk.sql.gz` works and creates a perfect file, if nothing else is going on (like starting DDEV)

cooperka:
I agree it works if nothing else is going on, but can we make the assumption that the average user knows that a prerequisite for this command to work reliably is that "nothing else is going on"? That seems risky, and in my case caused a full day of redoing work after my backups turned out to be unusable.

I did confirm by the way that additional stdout logs (including blank lines) corrupted the MIDDLE of the gzipped binary file, so it was very difficult to fix. In some cases you can simply remove the first few lines of stdout to fix the backup, but not always.

I don't think the docs should recommend a command that may silently fail to produce a valid backup. If you disagree, feel free to close.

stasadev:
Let's simply add `ddev start &&` to these commands with redirect:

```diff
-ddev export-db > /tmp/db.sql.gz
+ddev start && ddev export-db > /tmp/db.sql.gz
```

This works fine.

And we need update here:

https://github.com/ddev/ddev/blob/cccd95557898033d74c83bedea1b608ac6651345/cmd/ddev/cmd/export-db.go#L22-L23

And here:

https://github.com/ddev/ddev/blob/cccd95557898033d74c83bedea1b608ac6651345/cmd/ddev/cmd/import-db.go#L41-L43

Don't forget about pipe:

```diff
-gzip -dc db.sql.gz | ddev import-db
+ddev start && gzip -dc db.sql.gz | ddev import-db
```

Thank you.

cooperka:
Thanks for the suggestion @stasadev. I can make a change across those files too.

Just to clarify, ddev does not need to be running to import (it will start automatically). The issue with import is that it doesn't accept gzipped files with a redirect operator.

So, since the redirect operator is universally worse than just using the `--file=` argument, I'd still argue we should simply avoid recommending the redirect operator. It's less useful for both import and export. It's not an equivalent option and should not be documented as such. We could add all this context to explain how and when it works, but it's so much easier to just remove.

rfay:
You're right that `ddev import-db` expects plain-vanilla text as input, as most other utilities do. 

For example:
```
$ gzip -dc .tarballs/db.sql.gz | ddev import-db
mysql: [Warning] Using a password on the command line interface can be insecure.
Successfully imported database 'db' for d11
```

rfay:
I agree that it's odd that `ddev export-db` has always output gzipped content by default. It was one of those things that was like that very early and was hard to change after that.

cooperka:
I'm having a difficult time working with you @rfay, your comments are reading as defensive to me and I'm just trying to help. Could you please be explicit about what action item you're looking for? More feedback from others perhaps, or a specific change from me?

rfay:
I'm very much appreciating your proactive response to the friction you encountered! I'm trying to explain the history and the possible paths to improving the docs. That's why I suggested the `gzip -dc <db.sql.gz | ddev import-db` approach. Since you like pipes (and I like them very much) I thought you might want to include that.

rfay:
And a follow-up: Thanks for being open about feeling like it's difficult working with me. I think most people would walk away instead of saying that, and it's a marvelous thing that you know how to continue the conversation. If you'd like to see if we can take the edge off with a call about this (or anything else), let's talk at https://cal.com/randyfay/30min

rfay (changes requested):
The items that are removed in this PR all work correctly. I'm certainly open to more examples.

cooperka:
Like I said, I don't think the docs should recommend a command that may silently fail to produce a valid backup. If you disagree, feel free to close.

rfay:
Closing, as this removes valid examples. Improvement to the examples and addition of new ones will be welcome.

Issue #6344, opened 2024-06-24
Support Windows and WSL2 ARM64


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

There are many ARM64 Windows machines coming out this summer. People are going to be wanting them.

We have already had some ARM64 support for years, because WSL2 works there. @dennisameling did the original ARM64 work for his own ARM64 Surface Pro machine. 

However, we don't have
- [ ] Windows Installer (NSIS - doubt it will become available for arm64). 
- [ ] Required plugins for NSIS, see https://github.com/ddev/ddev/issues/6344#issuecomment-2192648799
- [ ] Distribution via Chocolatey (They don't have plans to support: https://github.com/chocolatey/choco/issues/1803#issuecomment-2187200057 ) (consider Winget?)
- [ ] Documentation
- [ ] Makefile assumes that Windows is amd64 and doesn't even build an arm64 binary (that part is easy, just add the type)
- [ ] Automated testing

### Describe your solution

Provide support

### Additional context

This will probably require getting an arm64 computer, which would be fun. 

Getting automated testing going will likely require yet another windows test machine when they start to become available.

dennisameling:
> This will probably require getting an arm64 computer, which would be fun.

You can also [run Azure VMs on Windows 11 ARM64](https://learn.microsoft.com/en-us/windows/arm/create-arm-vm) if that helps.

> Getting automated testing going will likely require yet another windows test machine when they start to become available.

GitHub hosted runners on arm64 Windows are available to GitHub Team and Enterprise users today, with support for open source projects coming later this year: https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/

rfay:
Yay @dennisameling ! Great to see you here! 

I heard they were coming later this year on Github. We haven't ever gotten Windows virtualized machines to work adequately, but maybe should give Github another chance. They always lack nested virtualization (or it's just too slow).

dennisameling:
> Yay @dennisameling ! Great to see you here!

It's been a while! Have been quite busy with other projects, and haven't been using DDEV lately as I've been working mostly on Kotlin-based backend and infra-related projects for work 😊 

Love to see that DDEV is still going strong! I can't commit much time now, but feel free to tag me if you need something to be tested on my Surface Pro X, and I'll test when I have time 👍🏼

rfay:
[...]

PR #6587, opened 2024-10-04
feat: Create Windows ARM64 installer, for #6344, fixes #5817

### LINKED ISSUE #5817: Windows 11 has introduced its own sudo.exe, make sure we're using gsudo

```
```
</details>         

### Expected Behavior

DDEV should behave predictably on Windows and WSL2, probably using its own gsudo.exe

However, [Windows has introduced its own sudo](https://devblogs.microsoft.com/commandline/introducing-sudo-for-windows/), so we'll likely have confusion.

### Actual Behavior

Make sure that where we invoke sudo.exe on the Windows side we do it explicitly as `gsudo.exe`

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

* #6344
* #5817 

This now adds a Windows ARM64 installer

- [x] Install and use on arm64 and amd64
- [x] Make sure installed binaries are signed (from release, https://github.com/ddev-test/ddev/releases/tag/v1.23.5-aaa-testwinarm64.2 )
- [x] Should not be able to run the arm64 installer on amd64, and vice versa.

## TODO:
- [ ] Follow up with a docs update explaining Windows ARM64 installation and use (traditional, docker-desktop, docker-ce)
- [ ] Update the WSL install scripts. Stop using chocolatey and instead get the installer directly from github release. This probably needs to be done after v1.23.5 comes out.


rfay:
@dennisameling thanks for all the work you've done over so many years with this. The groundwork you laid has paid off over and over. Would appreciate your review here and trying the installer out.  (There's a fully-built and signed installer at https://github.com/ddev-test/ddev/releases/tag/v1.23.5-aaa-testwinarm64.2 )

stasadev (approved):
Looks good to me, I didn't test the actual binary, but I looked for the `windows` keyword in the project to see if everything was updated in this PR.

rfay:
This isn't installing gsudo.exe successfully right now, at least on arm64.

rfay:
I had a release of this on ddev-test/ddev where the amd64 installer was not signed. I fiddled and tried to make it check and prevent that possibility. I hope I was successful. But it seems that the release after that went fine. We'll need to check carefully at actual install time.

Issue #6586, opened 2024-10-04
TestComposerCreateCmd doesn't work with Composer 2.8.1


rfay:
### Actual Behavior

Panic SEGV in pkg/composer/composer.go:65

I note this panic in TestComposerCreateCmd, https://buildkite.com/ddev/macos-colima-vz/builds/1710#01925778-6aa1-439e-a5a2-800234aee613/15326-15327

```


panic: runtime error: invalid memory address or nil pointer dereference [recovered]
  | panic: runtime error: invalid memory address or nil pointer dereference
  | [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x105300470]
  |  
  | goroutine 737 [running]:
  | testing.tRunner.func1.2({0x1055ce140, 0x105bc7190})
  | /opt/homebrew/Cellar/go/1.23.2/libexec/src/testing/testing.go:1632 +0x1bc
  | testing.tRunner.func1()
  | /opt/homebrew/Cellar/go/1.23.2/libexec/src/testing/testing.go:1635 +0x334
  | panic({0x1055ce140?, 0x105bc7190?})
  | /opt/homebrew/Cellar/go/1.23.2/libexec/src/runtime/panic.go:785 +0x124
  | github.com/ddev/ddev/pkg/composer.(*Manifest).keyExists(...)
  | /Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-6/ddev/macos-colima-vz/pkg/composer/composer.go:65
  | github.com/ddev/ddev/pkg/composer.(*Manifest).HasPostRootPackageInstallScript(...)
  | /Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-6/ddev/macos-colima-vz/pkg/composer/composer.go:104
  | github.com/ddev/ddev/cmd/ddev/cmd.TestComposerCreateCmd(0x14000334ea0)
  | /Users/testbot/tmp/buildkite-agent/builds/tb-macos-arm64-6/ddev/macos-colima-vz/cmd/ddev/cmd/composer-create_test.go:222 +0x2280
  | testing.tRunner(0x14000334ea0, 0x1056a0cb0)
  | /opt/homebrew/Cellar/go/1.23.2/libexec/src/testing/testing.go:1690 +0xe4
  | created by testing.(*T).Run in goroutine 1
  | /opt/homebrew/Cellar/go/1.23.2/libexec/src/testing/testing.go:1743 +0x314
 ```

[...]

PR #6588, opened 2024-10-05
test: Use composer 2.8.0 for TestComposerCreateCmd until composer bug is fixed, for #6586

rfay:
## The Issue

* #6586 

Composer 2.8.1 introduces new probably unwanted behavior that breaks TestComposerCreateCmd

Upstream issue: 
* https://github.com/composer/composer/issues/12150

* Temporarily use composer 2.8.0 for that test.
* Use `require` instead of `assert` so we don't run along and use invalid data and get a panic
* Minor fiddles with test flow

Read the test output.

## Misc

A comment about TestComposerCreateCmd: This is so impressive how much test coverage you built into this @stasadev, so great. In retrospect, I imagine all the if statements would be cleaner as a switch statement, or perhaps they could be a test matrix.

### DIFF
--- a/cmd/ddev/cmd/composer-create_test.go
+++ b/cmd/ddev/cmd/composer-create_test.go
@@ -29,7 +29,7 @@ func TestComposerCreateCmd(t *testing.T) {
 	assert := asrt.New(t)
 
 	origDir, err := os.Getwd()
-	assert.NoError(err)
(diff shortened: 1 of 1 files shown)


stasadev:
Thank you!

In the meantime, I'll be working on the packagist package.

PR #6589, opened 2024-10-06
fix: Use bitnami/mysql as base for mysql:8.x images, and support MySQL 8.4, fixes #6241, fixes #6511

### LINKED ISSUE #6241: Support for MySQL 8.4

MySQL 8.4 was released last month and it is an LTS version. It includes several changes and it would be a great addition to the project.

The release notes here: https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html

### Describe your solution

Add MySQL 8.4 support

### Describe alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #6511: mysql:8.0 - "mysql_native_password' is deprecated and will be removed in a future release"
`ddev start` with mysql:8.0 shows this output in `ddev logs -s db`:

`2024-08-27T17:20:08.116145Z 17 [Warning] [MY-013360] [Server] Plugin mysql_native_password reported: ''mysql_native_password' is deprecated and will be removed in a future release. Please use caching_sha2_password instead'`

It would be good to use the newer approach, if it doesn't break things.

This is in the initial setup, in create_db.sh:

https://github.com/ddev/ddev/blob/8900d5cdaa079244f962a00f990773e27fb878a5/containers/ddev-dbserver/files/create_base_db.sh#L76-L82

rfay:
## The Issue

Fixes
* #6241 
* https://github.com/ddev/ddev/issues/6511

Outstanding Percona Xtrabackup installation issue: https://perconadev.atlassian.net/browse/PXB-3361
Instructions for installation are at https://docs.percona.com/percona-xtrabackup/8.4/apt-repo.html#install-percona-xtrabackup-through-percona-release
The released URL for percona-release when things get deployed will be `https://repo.percona.com/apt/percona-release_latest.generic_all.deb`

This also tries to resolve longstanding configuration differences and annoyances and warnings, trying to fix warnings in the Dockerfile builds as well as in the first-time and later startup of mysql/mariadb databases. There were lots of them.

And it turns off log-bin, assuming we don't actually need logs in development, and assuming that they're costly in terms of performance and storage. So the configuration related to binlogs also gets removed.

There are a number of outstanding issues here, but I've made some progress:

- [x] **mysql:5.7 will have to continue to be built from ddev/mysql:5.7, which will still have to be maintained by us**: xtrabackup 2.4 will never be available on arm64 from Percona, so our mysql:5.7 support will apparently have to continue to be built on ubuntu:18.04, where both mysql-server (5.7) and percona-xtrabackup are available. There once was a percona-xtrabackup in [Debian snapshot](http://snapshot.debian.org/archive/debian/20141013T184415Z/pool/main/p/percona-xtrabackup/) but it was v2.2.3 and based on mysql 5.6. The one that we get with ddev/mysql:5.7 (from ubuntu:18.04) is "2.4.9 based on MySQL server 5.7.13".
- [x] **mysql:8.0 is based on bitnami**: I think that when Percona Xtrabackup rolls out their 8.0 arm64 support that we'll be able to build ddev-dbserver for mysql 8.0 based on bitnami/mysql:8.0, which will be a big win getting rid of our [xtrabackup-build](https://github.com/ddev/xtrabackup-build) and probably free us from having to pin matching xtrabackup/mysql-server versions.
- [x] I haven't tried building mariadb versions from bitnami/mariadb, and won't try it in this PR.
- [x] Update schema.json with mysql:8.4
- [ ] Update the percona-setup URL when it gets out of "testing"

## Outstanding significant problems (besides waiting for arm64 xtrabackup)

- [x] The bitnami config files in /opt/bitnami override what we put in /etc/my.cnf. AFAICT this is a result of how mysqld is built. This has to be investigated, and we either have to override what they provide or use their docker-entrypoint.sh or something.
- [x] We need separate my.cnf (or other config file) for different versions. We've used a common config for all maria/mysql for a long time, but now mysql:8.0 whines about deprecations and mysql:8.4 actually implements them.
- [x] The bitnami image has done something to make $HOME be set to `/` regardless of what it should be. As a hack, I set it in app_compose_template.yaml, but probably need to figure out why it doesn't get set. We could also set it in our code that adds the user layer to the image. Opened * https://github.com/bitnami/containers/issues/75578
- [x] Make sure we're not pinning mysql 8.0 version
- [x] Review TODOs here and make sure they're removed/accounted for
- [x] Our developer docs about the build process will need to be updated, removing the xtrabackup-build, but keeping the ddev/mysql:5.7, etc. xtrabackup-build can hopefully be archived.
- [x] Resolve configuration TODOs
- [x] mariabackup: io_uring_queue_init() failed with errno 0
- [x] [Warning] InnoDB: liburing disabled: falling back to innodb_use_native_aio=OFF
- [x] mariadb 10.11: 'innodb-large-prefix' was removed. It does nothing now and exists only for compatibility with old my.cnf files.
- [x] mariadb 10.11: 'innodb-file-format' was removed. It does nothing now and exists only for compatibility with old my.cnf files.
- [x] mariadb 10.11: 'innodb-log-files-in-group' was removed. It does nothing now and exists only for compatibility with old my.cnf files.
- [x] mariadb 10.11: You need to use --log-bin to make --expire-logs-days or --binlog-expire-logs-seconds work.
 - [x] mysql 8.0: [MY-010101] [Server] Insecure configuration for --secure-file-priv: Location is accessible to all OS users. Consider choosing a different directory.
 - [x] mysql 8.0: [Warning] [MY-013869] [InnoDB] Ignored deprecated configuration parameter innodb_log_file_size. Used innodb_redo_log_capacity instead
 - [x] mariadb 10.11: [Warning] 'innodb-large-prefix' was removed. It does nothing now and exists only for compatibility with old my.cnf files
 - [ ] 
- [x] I won't deal with the mariadb name changes in this PR
    - [ ] mariadb 11.4: `mysqld: Deprecated program name. It will be removed in a future release, use '/usr/sbin/mariadbd' instead`
    - [ ] mariadb 11.4: `/usr/bin/mysql_install_db: Deprecated program name. It will be removed in a future release, use 'mariadb-install-db' instead`
    - [ ] mariadb 11.4: `mysql_tzinfo_to_sql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb-tzinfo-to-sql' instead`
    - [ ] `mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead` (does this mean we haven't installed the compatibility package?)
    - [ ] mysqladmin: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb-admin' instead
    - [ ]  mariabackup: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb-backup

- [ ] Use mysql:8.4 and mysql:8.0 in both arm64 and amd64 environments.
- [ ] Examine whether we might have accidentally changed a mysql/mariadb config that people were counting on
- [ ] Try to compare mysql import speed per previous issues.
- [ ] Manually test on gitpod and codespaces
- [ ] Verify that configuration in HEAD is not changed where it's not wanted to be changed. Verify the new tests and compare. 

## Automated Testing Overview

- [x] mysql 8.4 tests added most places
- [x] Test coverage for correct config on (all?) (most?) database type/version


rfay:
At this point we can build the mysql:8.0 and 8.4 images, but everything else fails:

https://github.com/rfay/ddev/actions/runs/11203465444

containers/ddev-dbserver/Dockerfile
```diff
@@ -33,32 +40,60 @@ RUN set -x; if ( command -v mariadbd && ! command -v mysqld ); then \
...
 RUN mkdir -p /var/lib/mysql
 
 ADD files /
 
+# On bitnami-derived images, remove their default config so we get ours
+RUN <<BITNAMI_EOF
+if [ -d /opt/bitnami/mysql ]; then
+    rm -rf /opt/bitnami/mysql/conf
+fi
+BITNAMI_EOF
```
rfay, line 91:
```suggestion
RUN rm -rf /opt/bitnami/mysql/conf
```

containers/ddev-dbserver/Dockerfile
```diff
@@ -33,32 +40,60 @@ RUN set -x; if ( command -v mariadbd && ! command -v mysqld ); then \
...
+
+ENV MYSQL_UNIX_PORT=/var/tmp/mysql.sock
```
rfay, line 93: This seems to have been a thing through a long part of mysql's history, https://dev.mysql.com/doc/refman/5.7/en/environment-variables.html (shows in 8.0 and 8.4)

containers/ddev-dbserver/Dockerfile
```diff
@@ -33,32 +40,60 @@ RUN set -x; if ( command -v mariadbd && ! command -v mysqld ); then \
...
 RUN mkdir -p /var/log /var/tmp/mysqlbase /etc/mysql/conf.d && chmod -R ugo+wx /var/log /var/tmp/mysqlbase /etc/mysql/conf.d
-RUN ln -s -f /dev/stderr /var/log/mysqld.err
+RUN  ln -sf /tmp/mysqlx.sock ${MYSQL_UNIX_PORT}
+RUN if ! id mysql &>/dev/null ; then useradd -u 112 mysql; fi
```
rfay, line 96: 112 is the standard mysql user uid.

containers/ddev-dbserver/files/create_base_db.sh
```diff
@@ -23,23 +25,24 @@ mysqld_version=${mysqld_version%%-*}
...
     # mysql 5.5 requires running mysql_install_db in /usr/local/mysql
     if command -v mysqld | grep usr.local; then
         cd /usr/local/mysql
     fi
-    mysql_install_db --defaults-file=/var/tmp/my.cnf --force --datadir=/var/lib/mysql
+    mysql_install_db --defaults-file=/var/tmp/my.cnf --force --datadir=${DATADIR:-/var/lib/mysql}
 fi
-echo "Starting mysqld --skip-networking --socket=${SOCKET}"
-mysqld --defaults-file=/var/tmp/my.cnf --user=root --socket=$SOCKET --innodb_log_file_size=48M --skip-networking --datadir=/var/lib/mysql --server-id=0 --skip-log-bin &
+# TODO: Why was this changed to MYSQL_UNIX_PORT, review
```
rfay, line 37: Comment above shows it. See https://dev.mysql.com/doc/refman/8.4/en/environment-variables.html
```suggestion
```

containers/ddev-dbserver/test/image_general.bats
```diff
@@ -24,13 +24,3 @@ function setup {
   docker exec -u root -e "DDEV_MAX_DAYS_BEFORE_CERT_EXPIRATION=$DDEV_MAX_DAYS_BEFORE_CERT_EXPIRATION" ${CONTAINER_NAME} /tmp/check_key_expirations.sh >&3
 
 }
-
-@test "verify xtrabackup version equal to mysql-server version" {
```
rfay, line 28: This is now tested elsewhere in the config tests

containers/ddev-dbserver/test/testdata/database_config/mariadb_10.0.ini
```diff
@@ -4,7 +4,6 @@ datadir=/var/lib/mysql/
 secure_file_priv=/var/lib/mysql-files/
 innodb_use_native_aio=ON
 expire_logs_days=1
-log_bin_trust_function_creators=ON
```
rfay, line 7: I think we can safely remove this since we have binlog off.

pkg/ddevapp/config.go
```diff
@@ -1209,7 +1225,8 @@ ARG uid
 ARG gid
 ARG DDEV_PHP_VERSION
 ARG DDEV_DATABASE
-RUN (groupadd --gid $gid "$username" || groupadd "$username" || true) && (useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' --uid $uid "$username" || useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' "$username" || useradd  -l -m -s "/bin/bash" --gid "$gid" --comment '' "$username" || useradd -l -m -s "/bin/bash" --comment '' $username )
+RUN getent group tty || groupadd tty
```
rfay, line 1228: Debian images have the tty group, but ddev-router may not

rfay (commented):
It looks to me like this is likely to be OK, and it looks like the configuration expectations are correct.

rfay:
If you wouldn't mind taking a first look at this @stasadev I'd appreciate it. I'm hoping the main risks are mitigated, and I think we get rid of a lot of complex maintenance. But I haven't compared import performance, and turning off binlogs is I think a good thing but it is a bit of a change.

containers/ddev-dbserver/Dockerfile
```diff
@@ -33,32 +40,56 @@ RUN set -x; if ( command -v mariadbd && ! command -v mysqld ); then \
...
+                ;;
+
+            "8.4")
+                percona-release setup pxb84lts
+                apt-get update && apt-get -qq install -y percona-xtrabackup-84
+                ;;
+            # Older versions of mysql and a few mariadb use pxb24
+            *)
+                percona-release setup pxb24
+                apt-get update && apt install -qq -y percona-xtrabackup-24
```
stasadev, line 73:
```suggestion
                apt-get update && apt-get install -qq -y percona-xtrabackup-24
```

containers/ddev-dbserver/files/create_base_db.sh
```diff
@@ -71,24 +73,24 @@ mysql -uroot <<EOF
...
 
-mysqladmin -uroot password root
+mysqladmin -uroot --socket=${MYSQL_UNIX_PORT} password root
 
-if [  "${mysqld_version%%%.*}" = "8.0" ]; then
-    mysql -uroot -proot <<EOF
-        ALTER USER 'db'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_PASSWORD';
-        ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
-        ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
+if [[  "${mysqld_version%%%.*}" = ^8.[04]$ ]]; then
```
stasadev, line 78: Should be this?
```suggestion
if [[  "${mysqld_version%%%.*}" =~ ^8.[04]$ ]]; then
```

containers/ddev-dbserver/files/etc/my.cnf
```diff
@@ -6,44 +6,25 @@ default-character-set = utf8mb4
...
 secure-file-priv=/var/lib/mysql-files
+tmpdir=/tmp
 
 innodb_use_native_aio=ON
-expire-logs-days = 1
-
-# log_bin_trust_function_creators is required when log_bin is on for creating triggers
-log_bin_trust_function_creators=on
+# We don't need binlogs at all
+skip-log-bin
```
stasadev, line 19: Nice.

docs/content/developers/release-management.md
```diff
@@ -110,16 +110,12 @@ While it’s more error-prone, images can be pushed from the command line:
 5. `make push VERSION=<release_version> DOCKER_ARGS=--no-cache` for most of the images. For `ddev-dbserver` it’s `make PUSH=true VERSION=<release_version> DOCKER_ARGS=--no-cache`. There’s a [push-all.sh](https://github.com/ddev/ddev/blob/master/containers/push-all.sh) script to update all of them, but it takes forever.
 6. `ddev-dbserver` images can be pushed with `make PUSH=true VERSION=<release_version> DOCKER_ARGS=--no-cache` from the `containers/ddev-dbserver` directory.
 
-## Maintaining `ddev-dbserver` MySQL 5.7 and 8.0 ARM64 Images
+## Maintaining `ddev-dbserver` MySQL 5.7 ARM64 Images
 
-Sadly, there are no ARM64 Docker images for MySQL 5.7 and 8.0, so we have our own process to maintain [ddev/mysql-arm64-images](https://github.com/ddev/mysql-arm64-images) and [ddev/xtrabackup-build](https://github.com/ddev/xtrabackup-build) images for DDEV.
+There are no ARM64 Docker images for MySQL 5.7, so we have our own process to maintain [ddev/mysql-arm64-images](https://github.com/ddev/mysql-arm64-images).
```
stasadev, line 115: To be clear, ARM64 for MySQL 5.7 that I see here is not what we want?
https://hub.docker.com/r/bitnami/mysql/tags?name=5.7

stasadev (changes requested):
Big move!

`TestDBTypeVersionFromString` should also include MySQL 8.4.

containers/ddev-dbserver/files/create_base_db.sh
rfay: I wonder how we got by without this working.

docs/content/developers/release-management.md
rfay: I think it can be better stated. Seems the problem is we don't have a way to use xtrabackup with mysql 5.7 on arm64.

containers/ddev-dbserver/files/create_base_db.sh
```diff
@@ -71,24 +73,24 @@ mysql -uroot <<EOF
...
-if [  "${mysqld_version%%%.*}" = "8.0" ]; then
-    mysql -uroot -proot <<EOF
-        ALTER USER 'db'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_PASSWORD';
-        ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
-        ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
+if [[  "${mysqld_version%%%.*}" =~ ^8.[04]$ ]]; then
+    mysql -uroot -proot --socket=${MYSQL_UNIX_PORT} <<EOF
+    ALTER USER 'db'@'%' IDENTIFIED WITH caching_sha2_password BY '$MYSQL_PASSWORD';
+    ALTER USER 'root'@'%' IDENTIFIED WITH caching_sha2_password BY '$MYSQL_ROOT_PASSWORD';
+    ALTER USER 'root'@'localhost' IDENTIFIED WITH caching_sha2_password BY '$MYSQL_ROOT_PASSWORD';
```
stasadev, line 82: I think we should move this condition to other place. (it's not needed here because `caching_sha2_password` is the default.)

---

I have an old php 7.3 + mysql 8.0 project, `caching_sha2_password` is not supported there https://www.php.net/manual/en/mysqli.requirements.php

If I upgrade DDEV to this PR and run that project, the database works fine.

And I have this config set in `.ddev/mysql/my.cnf`:

```
[mysqld]
default_authentication_plugin = mysql_native_password
```

But when I try to import the database on that project using this PR, the connection doesn't work
```
Could not connect: SQLSTATE[HY000] [2054] The server requested authentication method unknown to the client
```
Because users have `caching_sha2_password`.

I know that I need to alter these users to have `mysql_native_password`, but others may not know that.

I think we should have `/start.sh` in dbserver, that looks at the current `default_authentication_plugin` and alters users using that plugin, so users don't need to figure it out themselves.

---

This is what we have in DDEV v1.24.1 for MySQL 8.0:

```
$ ddev exec "mysql -uroot -proot --execute='SELECT user, host, plugin FROM mysql.user;'"
mysql: [Warning] Using a password on the command line interface can be insecure.
+------------------+-----------+-----------------------+
| user             | host      | plugin                |
+------------------+-----------+-----------------------+
| db               | %         | mysql_native_password |
| root             | %         | mysql_native_password |
| db               | localhost | caching_sha2_password |
| mysql.infoschema | localhost | caching_sha2_password |
| mysql.session    | localhost | caching_sha2_password |
| mysql.sys        | localhost | caching_sha2_password |
| root             | localhost | mysql_native_password |
+------------------+-----------+-----------------------+
```

This is what we have in this PR for MySQL 8.0 (that's why nothing broke here before, it was already set to `caching_sha2_password`):

```
$ ddev exec "mysql -uroot -proot --execute='SELECT user, host, plugin FROM mysql.user;'"
mysql: [Warning] Using a password on the command line interface can be insecure.
+------------------+-----------+-----------------------+
| user             | host      | plugin                |
+------------------+-----------+-----------------------+
| db               | %         | caching_sha2_password |
| root             | %         | caching_sha2_password |
| db               | localhost | caching_sha2_password |
| mysql.infoschema | localhost | caching_sha2_password |
| mysql.session    | localhost | caching_sha2_password |
| mysql.sys        | localhost | caching_sha2_password |
| root             | localhost | caching_sha2_password |
+------------------+-----------+-----------------------+
```

stasadev (changes requested):

containers/ddev-dbserver/files/create_base_db.sh
stasadev: Something like this:

```bash
mysqld_version=$(mysqld --version | awk '{ print $3 }')

if [[ ("${mysqld_version%%%.*}" == "8.0" && "$(mysql -uroot -proot --socket=${MYSQL_UNIX_PORT} -e 'SELECT @@default_authentication_plugin;' -Ns 2>/dev/null)" == "mysql_native_password") ||
      ("${mysqld_version%%%.*}" == "8.4" && "$(mysql -uroot -proot --socket=${MYSQL_UNIX_PORT} -e 'SHOW PLUGINS;' -Ns 2>/dev/null | grep -q 'mysql_native_password'; echo $?)" == "0") ]]; then
    mysql -uroot -proot --socket=${MYSQL_UNIX_PORT} <<EOF
    ALTER USER 'db'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_PASSWORD';
    ALTER USER 'db'@'localhost' IDENTIFIED WITH mysql_native_password BY '$MYSQL_PASSWORD';
    ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
    ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '$MYSQL_ROOT_PASSWORD';
EOF
fi

```

To enable `mysql_native_password` in:

MySQL 8.0:

```
[mysqld]
default_authentication_plugin = mysql_native_password
```

MySQL 8.4:

```
[mysqld]
mysql_native_password = ON
```

stasadev: Or if there is no way to have `/start.sh` or somehow configure `/docker-entrypoint.sh` for this, then we should have the instruction for using `mysql_native_password` in our docs?

stasadev: Another option:

Add a script to `dbserver`, that will alter `db` and `root` users to the specific plugin:

```bash
ddev -s db exec alter_db_user_auth_plugin.sh mysql_native_password
```

And add this hook to the docs:

```yaml
hooks:
    post-start:
        - exec: alter_db_user_auth_plugin.sh mysql_native_password
          service: db
```

containers/ddev-dbserver/files/etc/mysql/version-conf.d/mysql_8.cnf.txt
```diff
@@ -0,0 +1,7 @@
+[mysqld]
+
+# caching_sha2_password is the default for mysql 8.0 and 8.4, no config required
+# default-authentication-plugin=caching_sha2_password
```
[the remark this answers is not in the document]
rfay: The good news is that because of our work with the client installs, we shouldn't have a *lot* of trouble with this, but thanks for pointing that out.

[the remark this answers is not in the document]
stasadev: Also, mysql-client in ddev-webserver has version 8.4.0, can be bumped to 8.4.3.

stasadev:
using dump from DDEV v1.24.1 and `mysql:8.0`:
1. ✅ `ddev delete -Oy && ddev config --database=mysql:8.0 && ddev import-db --file=/path/to/dump.sql.gz`
2. ✅ `ddev debug migrate-database mysql:8.4`
3. ✅ `ddev delete -Oy && ddev config --database=mysql:8.4 && ddev import-db --file=/path/to/dump.sql.gz`
4. ✅ `ddev snapshot` and `ddev snapshot restore` - tested both 8.0 and 8.4
5. ✅ large dump `ddev delete -Oy && ddev config --database=mysql:8.4 && ddev import-db --file=/path/to/dump.sql.gz`

containers/ddev-dbserver/Dockerfile
```diff
@@ -33,32 +40,56 @@ RUN set -x; if ( command -v mariadbd && ! command -v mysqld ); then \
 # look there when doing apt-get update. And we don't use new packages from there.
 # And we're going to install our own percona.list if needed, so get that if needed
 # and remove it here
+USER root
 RUN rm -f /etc/apt/sources.list.d/mariadb.list /etc/apt/sources.list.d/percona.list
+RUN mkdir -p /var/lib/apt/lists/partial
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" bzip2 curl gnupg2 less lsb-release pv tzdata vim-tiny wget
```
stasadev, line 46: We can add `vim` alternative, as in #6839

```suggestion
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" bzip2 curl gnupg2 less lsb-release pv tzdata vim-tiny wget
RUN update-alternatives --install /usr/bin/vim vim /usr/bin/vim.tiny 10
```

containers/ddev-dbserver/files/create_base_db.sh
stasadev: Or we can change the plugin to `mysql_native_password` for PHP 7.3 and below.

stasadev (commented):
Rebased.

stasadev (changes requested):
I tried `ddev dbeaver` with `mysql:8.0` and it couldn't open the database.

```
Public Key Retrieval is not allowed
```

The fix is to add `prop.allowPublicKeyRetrieval=true` and `prop.useSSL=false` to the connection string here:

https://github.com/ddev/ddev/blob/625ecc5ed7693c96d12e7e8ce81f973a715f0e61/pkg/ddevapp/global_dotddev_assets/commands/host/dbeaver#L32

See https://stackoverflow.com/a/74245704/8097891

containers/ddev-dbserver/files/etc/mysql/version-conf.d/mysql_8.cnf.txt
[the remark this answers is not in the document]
rfay: I'm so glad you mentioned that. I didn't even remember how that works!

stasadev:
I compared `ddev import-db` for `mysql:8.0` dump (22 GB) and tested it on `--database=mysql:8.0` in v1.24.1 and in this PR, and got the same import time ~75 minutes.

However, when I tried to do the same `ddev import-db` using `--database=mysql:8.4`, it was completed in ~95 minutes. Not sure if it's relevant (because the dump was done with `mysql:8.0` before).

containers/ddev-dbserver/files/etc/mysql/version-conf.d/mysql_8.cnf.txt
```diff
@@ -0,0 +1,11 @@
...
+
+# enable mysql-native-password only in case it needs to be used by php 7.3 and below
+# This is removed in MySQL 9
+mysql-native-password=1
```
stasadev, line 8: This option works only for mysql 8.4 (should be in the docs only, not here)

And correct option for mysql 8.0:

```
[mysqld]
default_authentication_plugin=mysql_native_password
```

pkg/ddevapp/ddevapp.go
```diff
@@ -1654,6 +1654,20 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
...
+		_, _, err = app.Exec(&ExecOpts{
+			Cmd:     fmt.Sprintf(`mysql -uroot -proot -e "%s"`, alterString),
+			Service: `db`,
+		})
+		if err != nil {
+			util.Warning("unable to set mysql_native_password db password: %v", err)
+		}
+		util.Debug(`mysql 8, php 5-7.3, set mysql_native_password`)
+	}
+
```
stasadev, line 1670: From https://github.com/ddev/ddev/pull/6589#discussion_r1886571895

(The code is not complete, because it's not easy to write here in suggestions.)

```suggestion
	if !slices.Contains(app.OmitContainers, "db") && app.Database.Type == nodeps.MySQL && (app.Database.Version == nodeps.MySQL80 || app.Database.Version == nodeps.MySQL84) && slices.Contains([]string{nodeps.PHP73, nodeps.PHP72, nodeps.PHP71, nodeps.PHP70, nodeps.PHP56}, app.PHPVersion) {
		// if mysql8.0, run app.Exec for:
		// `mysql -uroot -proot -e 'SELECT @@default_authentication_plugin;' -Ns 2>/dev/null` == "mysql_native_password"
		// else if mysql8.4, run app.Exec for:
		// `mysql -uroot -proot -e 'SELECT PLUGIN_STATUS FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME="mysql_native_password";' -Ns 2>/dev/null` == "ACTIVE"
		// if one of the condition above is true, run this alterString:
		alterString := `ALTER USER 'db'@'%' IDENTIFIED WITH mysql_native_password BY 'db';
			ALTER USER 'db'@'localhost' IDENTIFIED WITH mysql_native_password BY 'db';
			ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY 'root';
			ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';`
		_, _, err = app.Exec(&ExecOpts{
			Cmd:     fmt.Sprintf(`mysql -uroot -proot -e "%s"`, alterString),
			Service: `db`,
		})
		if err != nil {
			util.Warning("unable to set mysql_native_password db password: %v", err)
		}
		util.Debug(`mysql 8, php 5-7.3, set mysql_native_password`)
	}

```

stasadev (changes requested):

pkg/ddevapp/ddevapp.go
stasadev, line 1670: This is too obscure for the users, we can move these `ALTER USER` to the documentation and add a warning here with a link to the docs:

```suggestion
	if !slices.Contains(app.OmitContainers, "db") && app.Database.Type == nodeps.MySQL && (app.Database.Version == nodeps.MySQL80 || app.Database.Version == nodeps.MySQL84) && slices.Contains([]string{nodeps.PHP73, nodeps.PHP72, nodeps.PHP71, nodeps.PHP70, nodeps.PHP56}, app.PHPVersion) {
		util.Warning("mysql:%v and PHP %v are not compatible without extra steps, see the docs https://...", app.Database.Version, app.PHPVersion)
	}

```

stasadev (approved):
About `mysql_native_password`:

```
$ mysqld --verbose --help
...
  --mysql-native-password[=name] 
                      Enable or disable mysql_native_password plugin. Possible
                      values are ON, OFF, FORCE (don't start if the plugin
                      fails to load).
```

The `FORCE` option (`mysql-native-password=2`) activates the plugin, but doesn't make it the default.

I tested it with:
- php 7.3 + mysql 8.0
- php 7.3 + mysql 8.4
- php 7.4 + mysql 8.4 (it uses `mysql_native_password` if I switch from php7.3 with existing database, otherwise it uses `caching_sha2_password`)

Looks good to me!

rfay:
Wow, more than a year after starting on it. Followup will be using the production repo when they get it working, and there are perhaps docs, but we don't have to do the php7.3 docs any more I guess.

PR #6595, opened 2024-10-08
feat: Remove 127.0.0.1 from start and restart message, place in one function, fixes #6500

### LINKED ISSUE #6500: Don't output `127.0.0.1` URL on `ddev start` or `ddev restart` (if router is enabled)
On `ddev start` and `ddev restart` we show this at the end:

> Project can be reached at https://d11.ddev.site https://127.0.0.1:32810

If there are multiple hostnames or multiple fqdns we show all of them.

But the https://127.0.0.1 or http://127.0.0.1 sometimes confuses people who come from a world where all local dev is done on localhost. 

I think it would be better to just output this **when the router is enabled**

> Project can be reached at https://d11.ddev.site

(with any other other non-127.0.0.1 URLs)

If the router is disabled, the 127.0.0.1 URL is the only one available, so then it needs to be output.

davearch:
## The Issue

- #6500 

We don't want to output the 127.0.0.1 URL when the router is enabled.

Only shows the primary URL when the app is started/restarted

Run `ddev stop && ddev start` or `ddev restart`. Then you should see the updated output on the final line:
"Your project can be reached at mysite.ddev.site - see ddev describe for alternate URLs"

With the router omitted you should see the 127.0.0.1 URL.
Run: `ddev config global --omit-containers=ddev-router && ddev restart`

## Automated Testing Overview

Updated TestCmdRestart and TestCmdStart

### DIFF
--- a/cmd/ddev/cmd/restart.go
+++ b/cmd/ddev/cmd/restart.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"strings"
-
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/output"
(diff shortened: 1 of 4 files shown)


rfay (approved):
Looks great to me, congratulations. Manually tested with and without router.

davearch:
Done!

stasadev (approved):
Thank you!

rfay:
@davearch Your improvement made it onto ddev.com thanks to @stasadev !

image

davearch:
cool!

PR #6603, opened 2024-10-09
fix: improve timezone detection, for #6468

stasadev:
## The Issue

While attending today's contributor training on `ddev debug test`, I noticed a problem with timezone detection on Randy's MBP:

```
$ DDEV_DEBUG=true ddev start -y
...
2024-10-09T07:42:08.796 Unable to autodetect timezone: unable to read timezone from /usr/share/zoneinfo.default/America/Denver file
...
```

- #6468

After research, we found that `filepath.EvalSymlinks()` evaluates hidden symlinks on macOS:

```
$ ls -l /etc/localtime
lrwxr-xr-x  1 root  wheel  40 Oct  6 13:46 /etc/localtime -> /var/db/timezone/zoneinfo/America/Denver
```

```
$ ls -ltd /var/db/timezone/zoneinfo
lrwxr-xr-x  1 root  wheel  38 May 21 14:48 /var/db/timezone/zoneinfo -> /var/db/timezone/tz/2024a.1.0/zoneinfo
```

```
$ ls -ltd /var/db/timezone/zoneinfo
lrwxr-xr-x  1 root  wheel  27 Oct  6 13:45 /var/db/timezone/zoneinfo -> /usr/share/zoneinfo.default
```

---

And I noticed my own comment https://github.com/ddev/ddev/pull/6468#issuecomment-2273287479
> TODO remove optional right/ and posix/ from the timezone symlink https://stackoverflow.com/q/67885319/8097891

Which I forgot about, so I added it in this PR.

Run `DDEV_DEBUG=true ddev start` on macOS Sequoia.

The output should have this:
```
Using automatically detected timezone: TZ=America/Denver
```

And `ddev exec 'echo $TZ'` should show an appropriate timezone from the host.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


rfay (commented):
Thanks! Do we need another instance in the test coverage or this?

stasadev:
We have E2E testing here, I rely on what `/etc/localtime` returns.

I didn't find this path in the test runners, so it was unexpected for me to see `zoneinfo.default`

I could probably add unit testing here, but that would require moving some logic to a new function. Is it reasonable to do this?

PR #6609, opened 2024-10-10
feat: validation for database versions in `schema.json` [skip ci]

stasadev:
## The Issue

We don't have validation for database versions:

image

Adds validation:

image


rfay (approved):
So great! Let's not forget this when mysql:8.4

PR #6612, opened 2024-10-12
fix: don't try to release not-bound ports

rfay:
## The Issue

I've been using Colima as Docker Provider for the last few days, and I kept seeing these unexpected warnings:

```
Unable to properly check port status: dial tcp 127.0.0.1:0: connect: can't assign requested address
```

It turns out that we were trying to wait for release of ports that weren't bound, and the reason was that GetRouterBoundPorts() was walking through the router's ports, and some of those were not actually bound. Port 80 in this case was active inside but not bound.  So on Lima/Colima/Rancher, when it waits for ports to be unbound, this was shown.

If port PublicPort is 0 (not externally bound) then don't include it in GetRouterBoundPorts()

Using Colima or Lima:

Force it to use alternate ports (so 80 doesn't get bound):
```
echo 80 443 8025 8026 | xargs -n 1 -P 4 -I{} nc -k -l {}
```

Start and stop a project, using alternate ports
```
ddev start
DDEV_DEBUG=true ddev stop
```

This will show the actual port release sequence. You shouldn't see warnings.

### DIFF
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -283,7 +283,9 @@ func GetRouterBoundPorts() ([]uint16, error) {
 	}
 
 	for _, p := range r.Ports {
-		boundPorts = append(boundPorts, p.PublicPort)
+		if p.PublicPort != 0 {
+			boundPorts = append(boundPorts, p.PublicPort)
+		}
 	}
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Looks good to me.

PR #6613, opened 2024-10-13
build: push gitpod and codespaces images/packages during release

rfay:
## The Issue

We've had the gitpod image and codespaces package pushes as separate tasks for some time, and they weren't always done. 

* Push gitpod image after release build
* Push codespaces/vscode package after release build
* If building on ddev-test, push images to ddevhq Docker org so things are safer (the push happens to the variable DOCKER_ORG)
* Rename containers_shared.mak to containers_shared.mk so that it's recognized as a Makefile and tabs get used
* Adds a `latest` tag for Docker images for stable releases. We don't need to use these, and probably shouldn't, but this will remove maintenance of ddev-gitpod-launcher, and may help with how things work for drupalpod eventually. There is a place for `latest`

## Notes

I was overly ambitious here and tried to create a multi-architecture image of the upstream gitpod base image. That attempt was removed in https://github.com/ddev/ddev/pull/6613/commits/608a00d92476fa706d2ce5a9c38268c374f4aaf0 because it didn't seem to work quite right. I was trying to recreate the gitpod-sourced image with an alternate build technique, not using their `dazzle` technique. It didn't seem to me that it was working completely correctly, and I think my proposed use of the arm64 image was probably not as useful as expected.

* Build a release with normal release process on ddev-test. See https://github.com/ddev-test/ddev/releases/tag/v1.23.5-aaa-testgitpod.21 and https://github.com/ddev-test/ddev/actions/runs/11346993948
* Manual push of ddev-gitpod-base done in https://github.com/rfay/ddev/actions/runs/11346980563


rfay:
- [ ] TODO: Unfortunately the FROM we're using, [gitpod/workspace-base](https://hub.docker.com/r/gitpod/workspace-base/tags) is built only for amd64, so we're kind of fooling ourselves building a multi-arch image for it. I guess we have to build our own copy of the gitpod/workspace-base in multi-arch.

rfay:
OK, ready for your review. That turned out to be a lot more than a Sunday afternoon excursion :)

containers/ddev-dbserver/Makefile
```diff
@@ -1,12 +1,13 @@
...
+DOCKER_REPO ?= $(DOCKER_ORG)/ddev-dbserver
 
 VERSION := $(shell git describe --tags --always --dirty)
 
 .PHONY: build clean test
 
 CURRENT_ARCH=$(shell ../get_arch.sh)
 
-# This Makefile explicitly does not include containers/containers_shared.mak,
+# This Makefile explicitly does not include containers/containers_shared.mk,
```
stasadev, line 10: I see test build error for ddev-dbserver:
```
./build_image.sh: line 123: DOCkER_ORG: unbound variable
```

`build_image.sh` needs to pass `DOCKER_ORG` inside:

https://github.com/ddev/ddev/blob/aeb5ee06fd012de52d3332b23f85005439863172/containers/ddev-dbserver/Makefile#L63-L67

.github/workflows/master-build.yml
```diff
@@ -336,3 +339,82 @@ jobs:
...
+          GITHUB_TOKEN: "op://push-secrets/DDEV_GITHUB_TOKEN/credential"
+
+      - uses: actions/checkout@v4
+        with:
+          # We need to get all branches and tags for git describe to work properly
+          fetch-depth: 0
+
+      - name: Replace version in devcontainer-feature.json
+        run: |
+          sed -i 's/\${DDEV_VERSION}/'"${GITHUB_REF##*/}"'/g' ./containers/devcontainers/install-ddev/devcontainer-feature.json
```
stasadev, line 363: `${GITHUB_REF##*/}` looks the same as `${GITHUB_REF_NAME}` from https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

.github/workflows/master-build.yml
```diff
@@ -336,3 +339,82 @@ jobs:
...
+          username: ${{ vars.DOCKERHUB_USERNAME }}
+          password: ${{ env.DOCKERHUB_TOKEN }}
+
+      - name: Push ddev-gitpod-base image
+        if: startsWith(github.ref, 'refs/tags/v1')
+        run: |
+          pushd containers/ddev-gitpod-base >/dev/null
+          # For testing, push to ddevhq docker org
+          set -x
+          make push VERSION="${GITHUB_REF##*/}" DOCKER_ORG=${DOCKER_ORG}
```
stasadev, line 418: And here, maybe it's easier to use `${GITHUB_REF_NAME}`

.gitpod.yml
```diff
@@ -40,25 +40,6 @@ vscode:
...
-    pullRequestsFromForks: true
-    # add a check to pull requests (defaults to true)
-    addCheck: true
-    # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
-    addComment: false
-    # add a "Review in Gitpod" button to the pull request's description (defaults to false)
-    addBadge: true
-    # add a label once the prebuild is ready to pull requests (defaults to false)
-    addLabel: true
-
```
stasadev, line 61: Is this config not needed? I want to understand why we remove it.
https://www.gitpod.io/docs/integrations/github-gitpod-yaml

containers/ddev-gitpod-base/Dockerfile
```diff
@@ -36,6 +38,8 @@ RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
 # a gcc instance named gcc-5 is required for some vscode installations
 RUN sudo ln -sf $(which gcc) /usr/local/bin/gcc-5
 
+RUN mkdir -p ~/go/bin
+
 RUN for item in golang.org/x/tools/gopls@latest github.com/go-delve/delve/cmd/dlv@latest; do \
         go install $item; \
     done
```
stasadev, line 45: We should have `go clean -modcache` here in the same `RUN`, because `go install` downloads cache to `~/go/pkg`, and it's not removed.

stasadev (changes requested):

containers/ddev-dbserver/Makefile
rfay: Working on that, fails locally too.

containers/ddev-gitpod-base/Dockerfile
```diff
@@ -36,7 +38,9 @@ RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
 # a gcc instance named gcc-5 is required for some vscode installations
 RUN sudo ln -sf $(which gcc) /usr/local/bin/gcc-5
 
-RUN for item in golang.org/x/tools/gopls@latest github.com/go-delve/delve/cmd/dlv@latest; do \
+# Some ~/.cache and related have been created by root
+RUN sudo chown -R gitpod ~
+RUN go clean -modcache && for item in golang.org/x/tools/gopls@latest github.com/go-delve/delve/cmd/dlv@latest; do \
```
stasadev, line 43: Yes, but `go clean -modcache` should run after the loop.

containers/ddev-gitpod-base/Dockerfile
rfay: I wish we'd had this conversation earlier, I might have gotten out of the trouble I was in. I was really surprised about how this popped up when it didn't seem to have been there before. Perhaps upstream changed. Think it's sorted now.

rfay:
Hard to believe I could get so lost and destabilize so many things in an afternoon journey, but I think it's going now, https://github.com/ddev-test/ddev/releases/tag/v1.23.5-aaa-testgitpod.22 and https://github.com/ddev-test/ddev/actions/runs/11357139682

Please take another look and we'll see where we get.

stasadev (approved):
Looks good to me!

PR #6615, opened 2024-10-15
feat: add validation for `ddev dotenv` flags, for #6406

stasadev:
## The Issue

- #6406

Users should be aware that they have passed invalid flags to the `ddev dotenv`, as it may be unclear why some flags are simply ignored.

Adds flag validation.

Fixes a small fail in the test, that was caused by random behavior from `util.RandString(4)`:

https://github.com/ddev/ddev/actions/runs/11346006773/job/31554123302?pr=6615#step:16:83

```
=== RUN   TestValidateAsset
    appimport_test.go:38: 
        	Error Trace:	/home/runner/work/ddev/ddev/pkg/appimport/appimport_test.go:38
        	Error:      	Should be false
        	Test:       	TestValidateAsset
--- FAIL: TestValidateAsset (0.00s)
```

Before:

```
$ ddev dotenv set .env --test=123 -1asd=asd
Updated /home/stas/code/ddev/d11/.env with:

TEST="123"

$ ddev dotenv get .env --1asd
// shows help for the command without any error
```

After:

```
$ ddev dotenv set .env --test=123 -1asd=asd
Error reading command flags: the flag must be lowercase and start with a letter, but received -1asd=asd

$ ddev dotenv get .env --1asd
Error reading command flags: the flag must be lowercase and start with a letter, but received --1asd
```

### DIFF
--- a/cmd/ddev/cmd/dotenv-get.go
+++ b/cmd/ddev/cmd/dotenv-get.go
@@ -54,7 +54,10 @@ ddev dotenv get .ddev/.env.redis --redis-tag`,
 		}
 
 		// Get unknown flags and ensure only one flag is passed
-		envFlags := GetUnknownFlags(cmd)
+		envFlags, err := GetUnknownFlags(cmd)
+		if err != nil {
+			util.Failed("Error reading command flags: %v", err)
+		}
 		if len(envFlags) < 1 {
 			_ = cmd.Help()
 			return
@@ -69,7 +72,7 @@ ddev dotenv get .ddev/.env.redis --redis-tag`,
 		}
 
 		if !strings.HasPrefix(flag, "--") {
-			util.Failed("The flag must be in a long format.")
+			util.Failed("The flag must be in long format, but received %s", flag)
 		}
 
 		// Extract the environment variable name
--- a/cmd/ddev/cmd/dotenv-set.go
+++ b/cmd/ddev/cmd/dotenv-set.go
@@ -63,7 +63,10 @@ ddev dotenv set .ddev/.env.redis --redis-tag 7-bookworm`,
 		}
 
 		// Get unknown flags and convert them to env variables
-		envSlice := GetUnknownFlags(cmd)
+		envSlice, err := GetUnknownFlags(cmd)
+		if err != nil {
+			util.Failed("Error reading command flags: %v", err)
+		}
 		hasUnknownFlags := false
 		changedEnvMap := make(map[string]string)
 		for flag, value := range envSlice {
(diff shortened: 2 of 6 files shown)


cmd/ddev/cmd/dotenv-get.go
```diff
@@ -54,7 +54,10 @@ ddev dotenv get .ddev/.env.redis --redis-tag`,
 		}
 
 		// Get unknown flags and ensure only one flag is passed
-		envFlags := GetUnknownFlags(cmd)
+		envFlags, err := GetUnknownFlags(cmd)
+		if err != nil {
+			util.Failed("Error reading command flags: %v", err)
```
rfay, line 59: Should this be a fail or a warning?

cmd/ddev/cmd/dotenv-get.go
```diff
@@ -69,7 +72,7 @@ ddev dotenv get .ddev/.env.redis --redis-tag`,
 		}
 
 		if !strings.HasPrefix(flag, "--") {
-			util.Failed("The flag must be in a long format.")
+			util.Failed("The flag must be in long format, but received %s", flag)
```
rfay, line 75: Should this be a fail or a warning?

rfay (commented):
Looks pretty safe, and I commented about a couple of util.Failed(). However, since we're only failing the `add-on get` it's probably not that big a deal. I guess if something changed and a scripted flag were obsoleted it would make a difference.

cmd/ddev/cmd/dotenv-set.go
```diff
@@ -72,6 +75,8 @@ ddev dotenv set .ddev/.env.redis --redis-tag 7-bookworm`,
 				envMap[envName] = value
 				changedEnvMap[envName] = value
 				hasUnknownFlags = true
+			} else {
+				util.Failed("The flag must be in long format, but received %s", flag)
```
stasadev, line 79: And I fail too here, so users can see that they passed a wrong flag to the command.

I don't think changing the rules would break this behavior because I'm already using the strictest check I could think of.

cmd/ddev/cmd/dotenv-get.go
stasadev: I used fail because I wanted to stop the command execution.

In `ddev dotenv get` it doesn't really matter, because it accepts only one flag anyways.

cmd/ddev/cmd/dotenv-get.go
stasadev: `ddev dotenv get` accepts only one flag, should fail here.

rfay (approved):
Thanks!

PR #6620, opened 2024-10-16
fix: Some UPSUN environment variables are now PLATFORM_ environment variables

rfay:
## The Issue

I was using the `upsun` CLI and noted that it doesn't respect UPSUN_PROJECT and UPSUN_ENVIRONMENT

Move to the PLATFORM_ environment variables.

`ddev pull upsun` and `ddev push upsun`

## Automated Testing Overview

Everything should stay the same.


stasadev (approved):
I remember when I did this integration, some variables were called `PLATFORM_` and I named them to `UPSUN_` because I thought it was a beta version and would be renamed soon and it didn't matter for `ddev pull` and `ddev push`.

But if someone is using the `upsun` CLI directly, then I agree that it should be renamed, thanks.

rfay:
It didn't actually matter because of how we were using them explicitly. And it seems that they decided on this down the line.

Anyway, this PR supports both now. 

But probably should move to an add-on.

PR #6622, opened 2024-10-16
build: make latest tag same hash, push latest for most

rfay:
## The Issue

When we were reviewing the results of 
* https://github.com/ddev/ddev/pull/6613

today, Stas pointed out that

* `latest` didn't have the correct matching hash
* `ddev-dbserver` didn't get `latest` tag at all

I couldn't help but work on it a bit, so here it is.

This should not go into DDEV v1.23.5, and testing was done against ddevhq org

Use the web UI to push images.


stasadev (approved):
Looks good to me.

PR #6624, opened 2024-10-17
build: update Gitpod ports, update Gitpod README.md, for #6613 [skip ci]

stasadev:
## The Issue

- #6613

- Copy-paste Gitpod ports from https://github.com/ddev/ddev-gitpod-launcher/blob/main/.gitpod.yml
- Update links, that became outdated after #6613

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -41,19 +41,42 @@ vscode:
     - rogalmic.bash-debug
 
 ports:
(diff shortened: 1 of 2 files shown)


rfay (approved):
I tried this out at https://gitpod.io/?autostart=true#https://github.com/ddev/ddev/pull/6624 and everything seems fine. thanks!

PR #6626, opened 2024-10-17
docs: mention how to use xdebug with drush 13+

hanoii:
The change on #6565 made it less obvious how to run drush through xdebug, specially for those of us who were used to just working.

I don't quite like suggesting to do `ddev exec 'DRUSH_ALLOW_XDEBUG=1 drush st'`, instead one would want to do `DRUSH_ALLOW_XDEBUG=1 ddev drush st` but this will not work unless ddev takes care of adding the env var, which I think would be a nice addition if it's possible.


rfay:
@weitzman will want to comment here, but as far as we could tell that wasn't useful in the past or for the future.

hanoii:
Right now it won't run through xdebug unless you manually add the envar or use `--xdebug`. At least with drush 13, not sure on previous versions.

hanoii:
Ah. Probably the disabling of xdebug was not there on previous versions so it would still work for drush 12 and before. For drush 13 we have to do something for it to use xdebug

hanoii:
I confirm with drush 12 and before it works like before.

hanoii:
So adding a note for Drush 13+ to the docs then.

hanoii:
@rfay @weitzman In writing the docs I get more confused..  ..  If you have `DRUSH_ALLOW_XDEBUG=1` as we had before, even with D13, but xdebug is disabled on ddev (never enabled or disabled by `ddev xdebug off`, would this still have a performance hit? I think not.

By removing `DRUSH_ALLOW_XDEBUG` from being on ddev we are now forcing everyone to want to debug drush calls on ddev to have to do something, while before it was enough with `ddev xdebug on`.

hanoii:
Nevermind, the idea is that one should be able to debug the site without affecting drush commands, I get it.

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -16,6 +16,9 @@ It will remain enabled until you start or restart the project.
...
 * The IDE’s debug server port must be set to Xdebug’s default 9003, which is already the case in popular IDEs. If the unusual case that you have a port conflict, see [Using Xdebug on a Port Other than the Default 9003](#using-xdebug-on-a-port-other-than-the-default-9003) below.
 * In the case of using running your IDE inside WSL2 (using WSLg) or with a proxy setup like JetBrains Gateway, you can set that with `ddev config global --xdebug-ide-location=wsl2`. If you're running your IDE with a proxy inside the web container, you can set that with `ddev config global --xdebug-ide-location=container`.
+* [Drush 13+ disables Xdebug](https://www.drush.org/13.x/commands/#xdebug) even when Xdebug is enabled on DDEV. As per their docs you can:
+    * Enable it for a single Drush command by running: `ddev drush --xdebug`
+    * Set `DRUSH_ALLOW_XDEBUG=1` [environment variable](../../extend/customization-extendibility.md#environment-variables-for-containers-and-services), allowing every drush call to be run with Xdebug if enabled. Not recommended as one will normally want to debug the site without that having a hit on drush commands.
```
weitzman, line 21:
```suggestion
    * Set `DRUSH_ALLOW_XDEBUG=1` [environment variable](../../extend/customization-extendibility.md#environment-variables-for-containers-and-services), allowing every Drush call to be run with Xdebug when the PHP extension is enabled.
```

weitzman: IMO we don't want to make a recommendation. Just let user do the easiest. They have been payingn the perf penalty for years and are fine.

hanoii: fine!

hanoii:
I made a small edit in where I put the link to drush 13 xdebug docs

docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -16,6 +16,9 @@ It will remain enabled until you start or restart the project.
 * You may need to open port 9003 in your firewall to allow Xdebug access. (See [Troubleshooting Xdebug](#troubleshooting-xdebug) below.)
 * The IDE’s debug server port must be set to Xdebug’s default 9003, which is already the case in popular IDEs. If the unusual case that you have a port conflict, see [Using Xdebug on a Port Other than the Default 9003](#using-xdebug-on-a-port-other-than-the-default-9003) below.
 * In the case of using running your IDE inside WSL2 (using WSLg) or with a proxy setup like JetBrains Gateway, you can set that with `ddev config global --xdebug-ide-location=wsl2`. If you're running your IDE with a proxy inside the web container, you can set that with `ddev config global --xdebug-ide-location=container`.
+* Drush 13+ disables Xdebug even when Xdebug is enabled on DDEV. As per [their docs](https://www.drush.org/13.x/commands/#xdebug) you can:
```
rfay, line 19:
```suggestion
* Drush 13+ disables Xdebug even when Xdebug is enabled on DDEV. As per [drush docs](https://www.drush.org/13.x/commands/#xdebug) you can:
```

rfay (changes requested):
Thanks for sorting this out.

I'd rather have this somewhere other than the step-debugging page, as DDEV is not a Drupal-only tool, and a large number of people won't even know what the context is here, or what drush is. Could this go in the "Drupal specifics" section of cms-settings.md? Perhaps with a minor link here?

hanoii:
@rfay I obviously move it to where you think it goes best, but I honestly think this is where I would expect to find something like this. Maybe add a Drupal specific section here? Or add a Drupal note in front of, like: 

- For Drupal projects, [what is now on the PR]

rfay (changes requested):
Thanks for chasing this and figuring it all out. Please add the description of what's going on and how to deal with it in the drupal-specific section, then add a single sentence here saying something like "To debug Drupal `drush` commands see [instructions]()..."

rfay:
Minor complaint about list indentation. Thanks for your flexibility. https://github.com/ddev/ddev/actions/runs/11595260734/job/32283443215#step:10:12

hanoii:
@rfay fixed

rfay:
textlint absolutely insists that we use `Drush` instead of `drush`. I would be OK with removing that from `.texlintrc` in this PR rather than bowing down to it. What do you think? The complaint can also be fixed by saying `drush` with backticks.

hanoii:
But the checks all passed 🤔

rfay:
Sort of passed. Never have looked into why textlint is reported only inline. https://github.com/ddev/ddev/pull/6626/files

hanoii:
@rfay Drush is with a capital D on their docs so I am OK maintaining that.

docs/content/users/usage/cms-settings.md
```diff
@@ -53,6 +53,15 @@ With the default Drupal configuration, it’s very difficult to debug Twig templ
 $settings['container_yamls'][] = DRUPAL_ROOT . '/sites/development.services.yml';
 ```
 
+#### Drush and Xdebug
+
+Drush 13+ disables Xdebug even when [Xdebug is enabled on DDEV](../debugging-profiling/step-debugging.md).
+
+As per [drush docs](https://www.drush.org/13.x/commands/#xdebug) you can:
```
rfay, line 60:
```suggestion
As per [Drush docs](https://www.drush.org/13.x/commands/#xdebug) you can:  
```

rfay:
Thanks so much for working so hard on such a simple thing.

PR #6604, opened 2024-10-09
fix: use Cmd instead of RawCmd to execute composer and put vendor/bin first in PATH, fixes #6602

rfay:
## The Issue

- #6602

We should really prefer the COMPOSER_ROOT/vendor/bin in our $PATH instead of having it later in the PATH.

If composer/composer is in vendor/bin, it should be used from there when possible

## How This PR Solves The Issue

* Change the PATH setup that the image uses
* Change `ddev composer` to use `Cmd` instead of `RawCmd` so it gets interpreted by bash

## Manual Testing Instructions

What I did was `ddev composer require composer/composer:v2.8.0` and then used `ddev composer --version` and `ddev exec composer --version` for testing.

## Release/Deployment Notes

Right now, this is experimental, since this could possibly affect things we don't expect to affect: It can certainly change behavior when composer/composer is required, but it could also change something where something is in vendor/bin otherwise and thus masks something elsewhere in the container.

This also makes the vendor/bin/composer *override* anything that might be set for `composer_version`

PR #6627, opened 2024-10-18
fix: escape composer args, for #6604

stasadev:
## The Issue

From Discord https://discord.com/channels/664580571770388500/1296719909367054420

```
ddev composer config repositories.0 "{\"type\": \"composer\", \"url\":\"hello\",  \"exclude\": [\"magento/composer-dependency-version-audit-plugin\"]}"
```

(We can pass a json)

Since today (1.23.5), it fails with this message: 

```
In ConfigCommand.php line 766:You must pass the type and a url. Example: php composer.phar config reposit  ories.foo vcs https://bar.com
```
As I'm passing a type and url, I suspect there are lost somewhere

The culprit:
- #6604

This was discussed earlier in:
- #2547

Adds proper escaping.
Adds missing test coverage.

Test any composer command, that has `"`, `'`, `' '` (whitespace), or `{}`:

```
ddev composer config description "test description"
ddev composer config description "''''"
ddev composer config description "\""
ddev composer config repositories.local '{"type": "composer", "url": "https://example.com"}'
```

And check `composer.json` file for changes.


rfay:
Double-bashing is a never-ending pain.

rfay:
@stasadev thanks so much for explaining why this couldn't be solved with a more general fix (https://github.com/ddev/ddev/pull/6627#issuecomment-2422364014). I was wondering about that. 

More followups in the issue.

stasadev:
I updated this PR to pass `PATH` to the `docker-compose exec`.

This is better than escaping.

I explicitly used this condition:

```go
if err == nil && path != "" {
	env = append(env, "PATH="+path)
}
```

Because it would be strange to fail the whole `composer` command if `echo $PATH` somehow fails.

rfay (commented):
Does this maybe need a test where composer/composer is composer-installed?

I do think this is the best approach in general.

cmd/ddev/cmd/composer_test.go
```diff
@@ -162,6 +163,25 @@ func TestComposerCmdCreateRequireRemove(t *testing.T) {
...
+		out, err = exec.RunHostCommand(DdevBin, args...)
+		assert.NoError(err, "failed to run %v: err=%v, output=\n=====\n%s\n=====\n", args, err, out)
+		assert.NotContains(out, "2.8.0")
+		args = []string{"composer", "require", "composer/composer:v2.8.0"}
+		out, err = exec.RunHostCommand(DdevBin, args...)
+		assert.NoError(err, "failed to run %v: err=%v, output=\n=====\n%s\n=====\n", args, err, out)
+		args = []string{"composer", "--version"}
+		out, err = exec.RunHostCommand(DdevBin, args...)
+		assert.NoError(err, "failed to run %v: err=%v, output=\n=====\n%s\n=====\n", args, err, out)
+		assert.Contains(out, "2.8.0")
```
stasadev, line 184: > Does this maybe need a test where composer/composer is composer-installed?

I have already added such a test.

rfay: Thank you!

pkg/ddevapp/composer.go
```diff
@@ -18,14 +16,22 @@ func (app *DdevApp) Composer(args []string) (string, string, error) {
...
 	}
 
-	argString := strings.Join(args, " ")
+	// Prevent Composer from debugging when Xdebug is enabled
+	env := []string{"XDEBUG_MODE=off"}
+	// Let Composer know which binary to run from the PATH
+	path, _, err := app.Exec(&ExecOpts{
+		Cmd: "echo $PATH",
+	})
+	if err == nil && path != "" {
```
rfay, line 25:
```suggestion
	path = strings.Trim(path, "\n")
	if err == nil && path != "" {
```

rfay (approved):
This is a very elegant solution and gets us where we want to be. Thanks for exploring the fact that `docker-compose exec` doesn't add any PATH no matter what we do. I have the one suggestion of trimming the path after we get it. Stepping through I noticed the `\n` on the end.

rfay:
Sorry, the suggestion also requires `strings` to be imported.

PR #6631, opened 2024-10-19
docs: separate Debian and Homebrew install instructions for previous DDEV versions, link to it from Building, Testing, and Contributing page

gitressa:
## The Issue

The instructions under [How can I install a specific version of DDEV?](https://ddev.readthedocs.io/en/stable/users/usage/faq/#how-can-i-install-a-specific-version-of-ddev) contain a mix of Homebrew and Debian-flavor instructions.

The single step for Debian-flavored OS'es is very simple, so let's place it separately, at the top, followed closely by the steps for Homebrew.

Also, many users who want to switch to an older version of DDEV will find [Building, Testing, and Contributing](https://ddev.readthedocs.io/en/stable/developers/building-contributing/) so perhaps add a link to "How can I install a specific version of DDEV" there as well?

Updates the text under [How can I install a specific version of DDEV?](https://ddev.readthedocs.io/en/stable/users/usage/faq/#how-can-i-install-a-specific-version-of-ddev) and adds a link to it on [Building, Testing, and Contributing](https://ddev.readthedocs.io/en/stable/developers/building-contributing/).


docs/content/developers/building-contributing.md
```diff
@@ -19,6 +19,10 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
 
 Download and unzip the appropriate binary and place it in your `$PATH`.
 
+## Install a previous version
+
+You can also [install an old version of DDEV](https://ddev.readthedocs.io/en/stable/users/usage/faq/#how-can-i-install-a-specific-version-of-ddev).
```
stasadev, line 24: Thank you, I propose to add two more word for search: rollback and downgrade.

```suggestion
## Rollback to a previous version

You can also [downgrade to a previous version of DDEV](../users/usage/faq.md#how-can-i-install-a-specific-version-of-ddev).
```

stasadev (changes requested):

docs/content/developers/building-contributing.md
gitressa: Great suggestion, thanks @stasadev!

rfay (approved):
I think this is fine. I kind of wish people wouldn't use the scripted install_ddev.sh any more because they invariably end up confused about having multiple versions of DDEV installed, but it's still a useful technique, especially for this case.

rfay:
Thank you!

gitressa:
You're welcome! Thank you all for maintaining and expanding DDEV with new features -- it's a wonderful tool, which just keeps on getting better and better.

PR #6632, opened 2024-10-21
docs: add docker volumes to ddev uninstall docs

IanEdington:
I was trying to wipe the database and didn't know that volumes were being used. This is an attempt to surface that information for whoever comes across this next.


rfay:
You would use `ddev delete` to delete those volumes, which would have been part of your uninstallation right?

IanEdington:
yep. I'm not having a technical difficult. I just noticed part of the docs where it sounds like it's mentioning all the parts of docker it uses and it misses mentioning volumes.

rfay (approved):
This is fine, thanks. The thing I guess I didn't understand is that the docs do explain that you need to `ddev delete` to get rid of the volumes, so not sure what happened in your situation. This docs change wouldn't have helped you with that problem.

Issue #6633, opened 2024-10-21
Allow `ddev config` to specify a particular drupal sub-version and retain that configuration


hanoii:
In working on https://github.com/ddev/ddev-drupal-contrib/pull/87 I have an issue

Imagine the following scenario:

- I have a contrib module
- I download the addon
- I start the project with php 8.2 because the module was already setup for php 8.2
- I run `ddev poser` fails because it needs php 8.3

So the rest doesn't work, and `ddev config --update` doesn't work either. 

**I know php 8.3 is the recommended command, but this could fail on newer versions**.

So personally I would suggest one of two things:

- Be verbose on the addos on the settings that ddev needs, forcefully adding corepack and php 8.3 for D11 and php 8.2 for D10 if those are the minimum requirements.
- Better yet, as ddev already [handles that](https://github.com/ddev/ddev/blob/a6d7d237c0a403e5d878b8378bf24f3d7f5ab8eb/pkg/ddevapp/drupal.go#L390) provide a way to pass this drupal version through a command line override.

Would something like this be accepted?

Something like `ddev config --project-type=drupal --drupal-version=11`?

rfay:
I think what I'd like best is for `ddev describe` to say `drupal(11)` or something like that. Should be pretty easy. So make clear that it's project type `drupal` with the detected subtype `11`

rfay:
This is what DDEV does:
https://github.com/ddev/ddev/blob/fddf8c8df8e38ffd9214d6220140f9275640f636/pkg/ddevapp/drupal.go#L326-L342

hanoii:
Yeah I noted that on issue summary. I am trying to address the alternative to configure a project type with a Drupal version without anything to look the version for

rfay:
DDEV doesn't track that, so that would be going back to the older way. There was definitely a significant conversation before doing it the way we did.

hanoii:
I understand. I guess I was seeing if there was a way away to bring the old way back but only through an optional override or something. Like if want to force it for any reason

rfay:
So the title should be something like "Allow `ddev config` to specify a particular `drupal` subversion and retain that configuration" ?

hanoii:
Something like that, yes, or "use the default configuration for that drupal subversion"

rfay:
Related conversation in 
* https://github.com/ddev/ddev/issues/6306

rfay:
Let's solve this in 
* https://github.com/ddev/ddev/issues/6635

[...]

PR #6634, opened 2024-10-22
feat: ignore config.*.local.y*ml

hanoii:
This is really a small addition, my goal here is so that an addon can configure something up and make sure it's going to be ignored.

As I am working on ddev/ddev-drupal-contrib I was thinking in providing a command to switch core. The readme says changing and env var on config.local.yaml, which obviously is fine, but If I wanted to automate this somewhow I was hoping to do something like:

`ddev init-drupal --version 11` which could create a `config.contrib.local.yaml`

With the following:

```yml
php: 8.3
corepack_enable: true
web_environment:
  - DRUPAL_CORE=^11
```

And that file will automatically be ignored.

This is very tangentially related to #6633


hanoii:
I just thought that `config.contrib.local.yaml` will not be read, but it is!

rfay:
`config.contrib_local.yaml` is your fastest (and most recommended) way out of this :)

hanoii:
> `config.contrib_local.yaml` is your fastest (and most recommended) way out of this :)

but this is not current ignored, the PR is to be able to add something that I know will be ignored  by ddev, if it makes sense.

rfay:
Ah, 
> make sure it's going to be ignored.

That may be something that should/could be figured out using the content of the files, rather than their names? (I didn't understand the intent first time around.)

hanoii:
We use filenames a lot in general, I don't know, this looks easy enough?

rfay (changes requested):
OK, I can live with this, but it needs docs in https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#extending-configyaml-with-custom-configyaml-files

docs/content/users/extend/customization-extendibility.md
```diff
@@ -318,7 +318,7 @@ router_http_port: 8080
 
 Extra `config.*.yaml` files are loaded in lexicographic order, so `config.a.yaml` will be overridden by `config.b.yaml`.
 
-Team members may choose to use `config.local.yaml` for local non-committed config changes, for example. `config.local.yaml` is gitignored by default.
+Team members may choose to use `config.local.yaml` for local non-committed config changes, for example. `config.*.local.yaml` files can also be used to modify the order in which the local overrides are loaded. `config.local.yaml` and `config.*.local.yaml` are gitignored by default.
```
rfay, line 321: I guess I didn't understand this had anything to do with overrides? And how does it change order? Looks to me like it's just another config.*.yaml, which happens to be gitignored.

hanoii:
@rfay docs added. I realized that this have the added benefit of one can locally alter addons configs by using something the lexicographic order. Unfortunately, `config.redis.yaml` comes after `config.redis.local.yaml`, but one can always do `config.redisoverrides.local.yaml` or whatever filename you want to make sure that comes after the one you want to override. I think any local file should come after the non-local ones, but that's another discussion.

docs/content/users/extend/customization-extendibility.md
hanoii: @rfay see my comment bellow in the issue queue. I just realize that, that's why I mentioned it on the docs.

rfay:
The way we do things so they sort right is `config.redis_extra.yaml`, for example. 

The way to *override* is to use [override_config](https://ddev.readthedocs.io/en/stable/users/configuration/config/#override_config)

hanoii:
right, that's what I meant, or attempt to mean.  you can now do a `config.redis_extra.local.yaml` that will be gitignored by default. If you think what I wrote is confusing I can keep it simpler, will edit.

hanoii:
@rfay this was actually what I did first.

rfay (approved):
Thanks, makes sense to me. I wish we hadn't done `config.local.y*ml` as opposed to `config.local.yaml` in the distant past whenever it was, but I guess that's where we are.

PR #6644, opened 2024-10-24
feat: add `ddev cd` command, fixes #6639

### LINKED ISSUE #6639: Add a `ddev cd <projectname>` command

No.

### Describe your solution

[homeshick](https://github.com/andsens/homeshick), which I use to manage my dot files on my Mac, provides a really handy command to `cd` into a given "castle", as it calls them, without having to type its whole path--or even know it: `homeshick cd CASTLE`.

I think a comparable command for DDEV would be awesome: `ddev cd <projectname>`

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #6639

Adds the `ddev cd` command.

I tested it with bash, zsh, fish on Linux.

```
# do what it says, restart shell and use it (autocompletion should also work)
ddev cd my-project
```

### DIFF
--- a/cmd/ddev/cmd/cd.go
+++ b/cmd/ddev/cmd/cd.go
@@ -1,42 +1,47 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/fileutil"
+	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/output"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
+	"path/filepath"
 )
 
 // CdCmd is the top-level "ddev cd" command
 var CdCmd = &cobra.Command{
 	Use:   "cd [project-name]",
 	Short: "Uses shell built-in 'cd' to change to a project directory",
-	Long: `For bash/zsh add this function to your ~/.bashrc or ~/.zshrc,
-then restart your shell, and use 'ddev cd project-name':
+	Long: fmt.Sprintf(`To enable 'ddev cd' command, source the ddev.sh script from your rc-script.
 
-ddev() {
-  if [ "$1" = "cd" ] && [ -n "$2" ]; then
-    cd "$(DDEV_VERBOSE=false command ddev cd "$2")"
-  else
-    command ddev "$@"
-  fi
-}
+From bash:
 
-For fish add this function to your ~/.config/fish/config.fish,
-then restart your shell, and use 'ddev cd project-name':
+printf '\nsource "%s"' >> ~/.bashrc
 
-function ddev
-  if test (count $argv) -eq 2 -a "$argv[1]" = "cd"
-    cd "$(DDEV_VERBOSE=false command ddev cd $argv[2])"
-  else
-    command ddev $argv
-  end
-end`,
+From zsh:
+
+printf '\nsource "%s"' >> ~/.zshrc
+
+From fish:
+
+echo \n'source "%s"' >> ~/.config/fish/config.fish
+
+Restart your shell, and use 'ddev cd project-name'.
+`, filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.sh"), filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.sh"), filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.fish")),
 	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
 	Example: `ddev cd
 command ddev cd project-name
 ddev cd project-name`,
 	Run: func(cmd *cobra.Command, args []string) {
+		for _, file := range []string{"ddev.sh", "ddev.fish"} {
+			if !fileutil.FileExists(filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells", file)) {
+				util.Failed("Unable to find %s in %s", file, globalconfig.GetGlobalDdevDir())
+			}
+
+		}
 		if len(args) < 1 {
 			output.UserOut.Println(cmd.Long)
 			return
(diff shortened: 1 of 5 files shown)


cmd/ddev/cmd/cd.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var bashFile = filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.sh")
+var fishFile = filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.fish")
+
+// CdCmd is the top-level "ddev cd" command
+var CdCmd = &cobra.Command{
+	Use:   "cd [project-name]",
+	Short: "Uses shell built-in 'cd' to change to a project directory",
+	Long: fmt.Sprintf(`To enable 'ddev cd' command, source the ddev.sh script from your rc-script.
```
TravisCarden, line 21:
```suggestion
	Long: fmt.Sprintf(`To enable the 'ddev cd' command, source the ddev.sh script from your rc-script.
```

cmd/ddev/cmd/cd.go
```diff
@@ -0,0 +1,66 @@
...
+
+For fish:
+
+echo \n'[ -f "%s" ] && source "%s"' >> ~/.config/fish/config.fish
+
+Restart your shell, and use 'ddev cd project-name'.
+`, bashFile, bashFile, bashFile, bashFile, fishFile, fishFile),
+	ValidArgsFunction: ddevapp.GetProjectNamesFunc("all", 1),
+	Example: `ddev cd
+command ddev cd project-name
```
TravisCarden, line 39: Without any explanation, this looks like a typo--at least it did to me until I saw the comments in `docs/content/users/usage/commands.md`.

TravisCarden (commented):
It works great for me on zsh! I left a couple of thoughts on the code.

cmd/ddev/cmd/cd.go
stasadev: Thanks, I guess we don't need this example here (docs are enough).

cmd/ddev/cmd/cd.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+var bashFile = filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.sh")
+var fishFile = filepath.Join(globalconfig.GetGlobalDdevDir(), "commands/host/shells/ddev.fish")
+
+// CdCmd is the top-level "ddev cd" command
+var CdCmd = &cobra.Command{
+	Use:   "cd [project-name]",
+	Short: "Uses shell built-in 'cd' to change to a project directory",
+	Long: fmt.Sprintf(`To enable the 'ddev cd' command, source the ddev.sh script from your rc-script.
```
rfay, line 21: Is this needed? I didn't need it... I love this.

docs/content/users/usage/commands.md
```diff
@@ -227,6 +227,20 @@ Run the `cake` command; available only in projects of type `cakephp`, and only a
...
+
+Uses shell built-in `cd` to change to a project directory.
+
+```shell
+# Shows help to make `ddev cd` work in your shell
+ddev cd
+# Returns the path to the project-name
+command ddev cd project-name
+# Goes to the project-name directory
+# Works only if you applied the instructions from `ddev cd`
```
rfay, line 240: I didn't apply them... (bash)

rfay (changes requested):
This is going to make me very, very happy. However, it looks to me like you outdid yourself, and the extra instructions are no longer necessary?

cmd/ddev/cmd/cd.go
stasadev: It's needed, I just tested my PR on a test runner, and it prints the directory:

```
$ ddev cd TestPkgDrupal10
/Users/testbot/tmp/ddevtest/TestPkgDrupal10581214420
```

It doesn't change the directory.

Please check if you already have some customizations.

TravisCarden: Same for me, on both bash and zsh. Without the shell function, it just prints the directory.

rfay:
Oh, I see. It's cheating me by action just like `cd` but not doing anything.

rfay:
Probably it shouldn't look so much like the output of a `cd` command, or we'll have support issues. Is there a way to detect whether it's configured or not? If it isn't tell them what to do?

stasadev:
I understand that it can have support issues.

The only way I can think of is to add a new command in `ddev debug`, so

```
$ ddev debug get-project-path <project-name>
/path/to/the/project
```

And

```
$ ddev cd
# shows help
$ ddev cd <project-name>
# shows help
$ ddev cd <wrong-project-name>
# shows error
```

Right now only `ddev cd` shows help.

stasadev:
I've improved the command with a hidden flag.

Now, it should be easier to understand.

And I found a bug (`ddev cd -h` didn't work before, but now it does).

docs/content/users/usage/commands.md
```diff
@@ -227,6 +227,16 @@ Run the `cake` command; available only in projects of type `cakephp`, and only a
...
 
+## `cd`
+
+Uses shell built-in `cd` to change to a project directory.
+
+```shell
+# Goes to the project-name directory
+# Works only if you applied the instructions from this command
+ddev cd project-name
+```
```
rfay, line 238:
```suggestion
Uses shell built-in `cd` to change to a project directory. For example, `ddev cd some-project` will change directories to the project root of the project named `some-project`. 

Note that this command can't work until you make a small addition to your `.bashrc`, `.zshrc`, or `config.fish`. To see the explanation of what you need to do:

`ddev cd show-me-how`
```

rfay (approved):
Where have you been all my life! All those times I've done `ddev list` and copied the approot to work on a test problem???? This is amazing.

I made minor suggestions to doc and perhaps to the long help. 

Now I wonder if this could be integrated into autojump... probably could... But I don't see it in https://github.com/wting/autojump

docs/content/users/usage/commands.md
stasadev: I updated `ddev cd show-me-how` part - because it showed this error:

```
$ ddev cd show-me-how
Failed to find path for project: could not find a project named 'show-me-how'. Run 'ddev list' to see currently active projects
```

The real name of the project should be provided here to show the help.

PR #6645, opened 2024-10-24
feat: add `ddev debug match-constraint`

hanoii:
In playing with tests on an add-one I was going to query for the version of ddev with some bashism (like https://github.com/ddev/ddev/blob/fddf8c8df8e38ffd9214d6220140f9275640f636/.buildkite/sanetestbot.sh#L13) but then thought that ddev could easily do a much better job with the feature we already have, so I crafted this small PR.

To test this:

`ddev debug match-constraint ">= 1.23.5"` will work
`ddev debug match-constraint ">= 1.23.6"` will fail

### DIFF
--- a/cmd/ddev/cmd/debug-match-constraint.go
+++ b/cmd/ddev/cmd/debug-match-constraint.go
@@ -6,7 +6,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// DebugCheckDBMatch verified that the DB Type/Version in container matches configured
 var DebugMatchConstraint = &cobra.Command{
 	Use:     "match-constraint",
 	Short:   "Check if the currently installed ddev matches the specified version constraint.",
--- a/cmd/ddev/cmd/debug-match-constraint.go
+++ b/cmd/ddev/cmd/debug-match-constraint.go
@@ -6,6 +6,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// DebugMatchConstraint Compares a constraint against the installed DDEV version.
 var DebugMatchConstraint = &cobra.Command{
 	Use:     "match-constraint",
 	Short:   "Check if the currently installed ddev matches the specified version constraint.",
--- a/cmd/ddev/cmd/debug-match-constraint.go
+++ b/cmd/ddev/cmd/debug-match-constraint.go
@@ -23,5 +23,5 @@ var DebugMatchConstraintCmd = &cobra.Command{
 }
 
 func init() {
-	DebugCmd.AddCommand(DebugMatchConstraint)
+	DebugCmd.AddCommand(DebugMatchConstraintCmd)
 }
--- a/cmd/ddev/cmd/debug-match-constraint.go
+++ b/cmd/ddev/cmd/debug-match-constraint.go
@@ -1,0 +1,27 @@
+package cmd
+
+import (
+	"github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+// DebugCheckDBMatch verified that the DB Type/Version in container matches configured
+var DebugMatchConstraint = &cobra.Command{
+	Use:     "match-constraint",
+	Short:   "Check if the currently installed ddev matches the specified version constraint.",
+	Args:    cobra.ExactArgs(1),
+	Example: `ddev debug match-constraint ">= 1.24.0"`,
+	Run: func(_ *cobra.Command, args []string) {
+		versionConstraint := args[0]
+
+		err := ddevapp.CheckDdevVersionConstraint(versionConstraint, "", "")
+		if err != nil {
+			util.Failed(err.Error())
+		}
+	},
+}
+
+func init() {
+	DebugCmd.AddCommand(DebugMatchConstraint)
+}
--- a/cmd/ddev/cmd/debug-match-constraint.go
+++ b/cmd/ddev/cmd/debug-match-constraint.go
@@ -6,8 +6,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// DebugMatchConstraint Compares a constraint against the installed DDEV version.
-var DebugMatchConstraint = &cobra.Command{
+// DebugMatchConstraintCmd Compares a constraint against the installed DDEV version.
+var DebugMatchConstraintCmd = &cobra.Command{
 	Use:     "match-constraint",
 	Short:   "Check if the currently installed ddev matches the specified version constraint.",
 	Args:    cobra.ExactArgs(1),
--- a/cmd/ddev/cmd/debug-match-constraint_test.go
+++ b/cmd/ddev/cmd/debug-match-constraint_test.go
@@ -8,9 +8,11 @@ import (
 
 // TestDebugMatchConstraintCmd checks to see match-constraint behaves as expected
 func TestDebugMatchConstraintCmd(t *testing.T) {
+	version, err := exec.RunHostCommand(DdevBin, "--version")
+
 	out, err := exec.RunHostCommand(DdevBin, "debug", "match-constraint", "> 1.0")
-	require.NoError(t, err, "Match constraint should not have errored for > 1.0", out)
+	require.NoError(t, err, "Match constraint should not have errored for > 1.0, version='%s', out='%s'", version, out)
 
 	out, err = exec.RunHostCommand(DdevBin, "debug", "match-constraint", "< 1.0")
-	require.Error(t, err, "Match constraint should have errored for < 1.0", out)
+	require.Error(t, err, "Match constraint should have errored for < 1.0, version='%s', out='%s'", version, out)
 }
--- a/cmd/ddev/cmd/debug-match-constraint_test.go
+++ b/cmd/ddev/cmd/debug-match-constraint_test.go
@@ -2,9 +2,9 @@ package cmd
 
 import (
 	"github.com/ddev/ddev/pkg/exec"
-	"github.com/ddev/ddev/pkg/util"
 	"github.com/ddev/ddev/pkg/versionconstants"
 	"github.com/stretchr/testify/require"
+	"strings"
 	"testing"
 )
 
@@ -13,8 +13,9 @@ import (
 func TestDebugMatchConstraintCmd(t *testing.T) {
 
 	constraint := "= " + versionconstants.DdevVersion
-	out, _ := exec.RunHostCommand(DdevBin, "--version")
-	util.Debug("DdevBin=%s, version=%s", DdevBin, out)
+	v, _ := exec.RunHostCommand(DdevBin, "--version")
+	v = strings.Trim(v, "\r\n ")
+	t.Logf("DdevBin='%s', version from binary='%s', version from versionconstants='%s'", DdevBin, v, constraint)
 
 	out, err := exec.RunHostCommand(DdevBin, "debug", "match-constraint", constraint)
 	require.NoError(t, err, "Match constraint should not have errored for %s, out='%s'", constraint, out)
--- a/cmd/ddev/cmd/debug-match-constraint_test.go
+++ b/cmd/ddev/cmd/debug-match-constraint_test.go
@@ -1,0 +1,16 @@
+package cmd
+
+import (
+	"github.com/ddev/ddev/pkg/exec"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// TestDebugMatchConstraintCmd checks to see match-constraint behaves as expected
+func TestDebugMatchConstraintCmd(t *testing.T) {
+	out, err := exec.RunHostCommand(DdevBin, "debug", "match-constraint", "> 1.0")
+	require.NoError(t, err, "Match constraint should not have errored for > 1.0", out)
+
+	out, err = exec.RunHostCommand(DdevBin, "debug", "match-constraint", "< 1.0")
+	require.Error(t, err, "Match constraint should have errored for < 1.0", out)
+}
--- a/cmd/ddev/cmd/debug-match-constraint_test.go
+++ b/cmd/ddev/cmd/debug-match-constraint_test.go
@@ -8,7 +8,7 @@ import (
 
 // TestDebugMatchConstraintCmd checks to see match-constraint behaves as expected
 func TestDebugMatchConstraintCmd(t *testing.T) {
-	version, err := exec.RunHostCommand(DdevBin, "--version")
+	version, _ := exec.RunHostCommand(DdevBin, "--version")
 
 	out, err := exec.RunHostCommand(DdevBin, "debug", "match-constraint", "> 1.0")
 	require.NoError(t, err, "Match constraint should not have errored for > 1.0, version='%s', out='%s'", version, out)
--- a/cmd/ddev/cmd/debug-match-constraint_test.go
+++ b/cmd/ddev/cmd/debug-match-constraint_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/util"
 	"github.com/ddev/ddev/pkg/versionconstants"
 	"github.com/stretchr/testify/require"
 	"testing"
@@ -12,6 +13,9 @@ import (
 func TestDebugMatchConstraintCmd(t *testing.T) {
 
 	constraint := "= " + versionconstants.DdevVersion
+	out, _ := exec.RunHostCommand(DdevBin, "--version")
+	util.Debug("DdevBin=%s, version=%s", DdevBin, out)
+
 	out, err := exec.RunHostCommand(DdevBin, "debug", "match-constraint", constraint)
 	require.NoError(t, err, "Match constraint should not have errored for %s, out='%s'", constraint, out)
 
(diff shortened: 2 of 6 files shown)


rfay:
Your intent here is not to verify the constraint, but rather to see if the DDEV version is valid with the constraint you offer?

stasadev:
This looks like a simpler version of:

- #6361

hanoii:
> Your intent here is not to verify the constraint, but rather to see if the DDEV version is valid with the constraint you offer?

correct

hanoii:
Maybe `--match-constraint` is better?

rfay:
But for add-ons we have `ddev_version_constraint` for this purpose. What were you trying to do with your test?

(I certainly agree that checking in go is easier than bash)

hanoii:
i.e. in the context of a bats test:

```sh
if ddev version --verify-constraint ">= 1.23.5"; then
  ddev add-on get https://....
else
  ddev get https://....
fi
```

hanoii:
The above example is exactly what prompted be tu submit this, but I can think how this can be useful elsewhere, like in githooks and other potential team workflows.

stasadev (changes requested):
> Maybe `--match-constraint` is better?

I think this one is better.

---

I think it's a nice addition because even if we have a `ddev_version_constraint` in `.ddev/config.yaml` and in the `install.yaml` for add-ons, it's still not flexible enough.

As @hanoii pointed out, it can be used on a larger scale.

---

@hanoii, we also need to update the documentation, thanks.

https://github.com/ddev/ddev/blob/4de65075ee0a00fad6b0ddabae91969fed33ab57/docs/content/users/usage/commands.md#L1539-L1548

hanoii:
rebased

hanoii:
docs added and flag name changed

docs/content/users/usage/commands.md
```diff
@@ -1540,16 +1540,30 @@ ddev typo3 site:show
...
 ```shell
 # Print DDEV and platform version details
 ddev version
+
+# ddev add-on command was added in v1.23.5, conditionally run the new command
+if ddev version --match-constraint ">= 1.23.5" 2>&1 > /dev/null; then
+  ddev add-on get ....
+else
+  ddev get ....  
+fi
```
stasadev, line 1558: It looks good, but it won't work as expected with v1.23.5 because there's no `--match-constraint' flag.

I think we need to mention that this flag will run the actual check only from DDEV v1.24.0.

hanoii: @stasadev I improved this, added an extra check and mention about when it was added. I also added it to the flag.

I didn't see any other reference in the docs to when features were introduced, at least not on the commands page.

stasadev (approved):
Thank you @hanoii,

It looks perfect to me.

Side note: `2>&1 > /dev/null` is the same as `&>/dev/null` which is shorter (I don't ask to change it here).

hanoii:
> Side note: `2>&1 > /dev/null` is the same as `&>/dev/null` which is shorter (I don't ask to change it here).

Thanks, wasn't aware of that!

rfay:
Another side note, `&>/dev/null` works only in bash 4+, not bash 3 (macOS).

hanoii:
The example is a bit of a catch-22, as you are asking if ddev is over 1.1.9 and then if it is over 1.23.5, but it only works on 1.24+ :D - but it is a good example of what you could have done or can do in the future

rfay (changes requested):
Sorry to ask this, and I should have perked up earlier. But could you please just put this in `ddev debug match-constraint` instead? It's a little mysterious why it would go in `ddev version` (which is an output/reporting command). So throwing it into `ddev debug` is super easy and doesn't pollute the command-space.

hanoii:
@rfay excellent suggestion and done.

@stasadev if you want to review again

cmd/ddev/cmd/debug-match-constraint.go
```diff
@@ -0,0 +1,27 @@
+package cmd
+
+import (
+	"github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+// DebugMatchConstraint Compares a constraint against the installed DDEV version.
+var DebugMatchConstraint = &cobra.Command{
```
stasadev, line 10:
```suggestion
// DebugMatchConstraintCmd Compares a constraint against the installed DDEV version.
var DebugMatchConstraintCmd = &cobra.Command{
```

cmd/ddev/cmd/debug-match-constraint.go
```diff
@@ -0,0 +1,27 @@
...
+
+		err := ddevapp.CheckDdevVersionConstraint(versionConstraint, "", "")
+		if err != nil {
+			util.Failed(err.Error())
+		}
+	},
+}
+
+func init() {
+	DebugCmd.AddCommand(DebugMatchConstraint)
```
stasadev, line 26:
```suggestion
	DebugCmd.AddCommand(DebugMatchConstraintCmd)
```

stasadev (commented):
It looks good to me, but the command should be named with the `Cmd` suffix.

hanoii:
Committed

rfay (commented):
Most of these have tests...  Can we get by without one @stasadev ?

hanoii:
I don't think one is needed, as the constraint logic is tested elsewhere, but it's also a simple test to add at least to make sure the command is there and works. Added test

rfay:
You also have an error in there with "(EXTRA string=)", need to do the `%` right.

rfay:
Does the test run locally successfully?

rfay:
You have an ineffassign too, https://github.com/ddev/ddev/actions/runs/11615220037/job/32345253597?pr=6645

Make sure to do `make staticrequired` (have golangci-lint installed)

hanoii:
sorry, fixed that go error

rfay:
So much pain! Seems to be building without tag information for some reason. This isn't my experience I don't think, and we have other things that do this class of test. You can use tmate to go into the test and use `make testcmd TESTARGS="-run TestDebugMatchConstraintCmd"` to see what's actually happening.

stasadev:
@hanoii, see example from my closed PR https://github.com/ddev/ddev/pull/6361/files#diff-037e1285dac63dbc94244f8cddceca76b2bc71adc43b9fb90827196a451e69b6R16-R27

We can reliably check only the invalid constraint, otherwise we need to skip the test.

hanoii:
How would I use tmate :)?

> We can reliably check only the invalid constraint, otherwise we need to skip the test.

I see what you did, might give that a try if what I just pushed is not acceptable, but I think it might be for such a simple test.

stasadev:
> How would I use tmate :)?

From your own fork, but you need to enable Actions first in Settings, and then:

image

> I see what you did, might give that a try if what I just pushed is not acceptable, but I think it might be for such a simple test.

I like it, no need for more.

rfay:
> How would I use tmate :)?

tmate is a great tool for these confusing GitHub situations. It was [Last week's Contributor Training](https://ddev.com/blog/tmate-github-actions-contributor-training/).

rfay:
And... we have some more things to learn. The buildkite tests didn't run here, saying

> Skipping buildkite build since no code changes found
 
I was wondering why they didn't fail. 

https://github.com/ddev/ddev/blob/608267b6f858ea1ff670fbd48b1d945b41807e94/.buildkite/test.sh#L160-L164

I don't see the problem at first glance, assume it's related to only *new* code. 

I'll open an issue for it. 

Thanks so much for working so diligently on these things @hanoii , sorry the pain level has been higher than any of us would expect.

hanoii:
If I cannot test the command behaving as expected, I don't see much of the purpose of adding what @stasadev mentioned, which is basically testing the functionality of the version constraint logic that already have tests.

I vote to remove the test.

Also I wonder, what would be the effort to have test build ddev as local with the proper tags or version?

rfay:
> Also I wonder, what would be the effort to have test build ddev as local with the proper tags or version?

I assume that the problem here is 

https://github.com/ddev/ddev/blob/master/.github/workflows/pr-build.yml#L39-L44


But it absolutely says we're to be getting all branches and tags.

You could also do a `make VERSION=<whatever>` and the resultant ddevBin would have the named version. Maybe that would be fun!

hanoii:
But that's for building the artifacts. The tests seems to be using the hash

rfay:
I added a little debugging and made the Makefile get all tags. Can't harm to try! 
Rebased.

rfay:
> But that's for building the artifacts. The tests seems to be using the hash

The build is using the hash because it can't find any tags in `git describe --tags --always --dirty`. 

`make` overrides the version.

rfay:
Your fork has no recent tags in it, so I imagine that is related to what's going on.

Try this for fun:
```
git ls-remote --tags hanoii # assuming your fork's remote is named `hanoii`; you'll see no recent tags
git fetch upstream --tags #gets all the tags from the upstream repo, assume you've named it that
git push hanoii --tags --no-verify  # some may fail, but it's ok
git ls-remote --tags hanoii. # now all the tags will be there.
```

hanoii:
I just pushed all tags, so maybe you need to restart tests or push something else?

rfay:
All you would have to do to restart all tests is `git commit -m "run tests" --allow-empty" and `git push`

We'll see what we learn, or feel free to do that if you want to do it in advance.

hanoii:
@rfay had fun last night with this, I see it was a matter of having the tags, I think though that something like what @stasadev needs to be added, otherwise this will fail for every other fork that doesn't pull in the tags, which will be a nightmare for you.

hanoii:
I'll fix the lint, add that, and I also went back to a more natural constraint.

hanoii:
ok, let's see, I cleaned up, I think doing --version is unnecessary as that is exactly the same that comes out of `versionconstants.DdevVersion`, correct me if I am wrong, but that seems to be the case.

This tests should pass, I will try to remove the tags and see if I can test this without them and make sure it's properly skipped.

hanoii:
Locally:

```
❯ GOTEST_SHORT=true make VERSION=24bfb07 testcmd TESTARGS="-run TestDebugMatchConstraintCmd"
...
Copying complete
=== RUN   TestDebugMatchConstraintCmd
    debug-match-constraint_test.go:18: Skipping because ddev version doesn't start with any valid version tag, it's '24bfb07'
--- SKIP: TestDebugMatchConstraintCmd (0.00s)
PASS
...
```

seems ok, if this is good enough for you maybe removing the tags is not necessary.

cmd/ddev/cmd/debug-match-constraint_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// TestDebugMatchConstraintCmd checks to see match-constraint behaves as expected
+// @see https://github.com/Masterminds/semver#checking-version-constraints
+func TestDebugMatchConstraintCmd(t *testing.T) {
+
+	versionRegex := regexp.MustCompile(`^v[0-9]+\.`)
+	if !versionRegex.MatchString(versionconstants.DdevVersion) {
+		t.Skip(fmt.Sprintf("Skipping because ddev version doesn't start with any valid version tag, it's '%v'", versionconstants.DdevVersion))
+	}
```
stasadev, line 19: `t.Skip` won't run because `versionconstants.DdevVersion = v0.0.0-overridden-by-make` when there is no tags.

Let me add another commit here. I'll add a better condition.

hanoii:
That’s not the case. And that is a valid version anyway

Reply might be shorter than usual / Respuesta más resumida que de costumbre

cmd/ddev/cmd/debug-match-constraint_test.go
hanoii: that is a valid version, but it will fail the constraint I added, bug if you look at the previous failed ones, there version was the sha hash, not that one.

hanoii: See https://github.com/ddev/ddev/actions/runs/11617743434/job/32353619870#step:16:16831

```
=== RUN   TestDebugMatchConstraintCmd
    debug-match-constraint_test.go:16: 
        	Error Trace:	/home/runner/work/ddev/ddev/cmd/ddev/cmd/debug-match-constraint_test.go:16
        	            	exit status 1
        	Test:       	TestDebugMatchConstraintCmd
        	Messages:   	Match constraint should not have errored for = fc8df38, out='error: the '= fc8df38' constraint is not valid. See https://github.com/Masterminds/semver#checking-version-constraints for valid constraints format
        	            	'
--- FAIL: TestDebugMatchConstraintCmd (0.03s)
```

hanoii: `fc8df38` is the version

hanoii: So if with tags we get the proper version and without we get the hash, I think we are good.

stasadev: Your test failed here, but it shouldn't.

stasadev: > Your test failed here, but it shouldn't.

I mean, I ran your latest commit without tags, and it failed

```
=== RUN   TestDebugMatchConstraintCmd
    debug-match-constraint_test.go:27: 
        	Error Trace:	/home/stas/code/ddev/ddev-stasadev-build/cmd/ddev/cmd/debug-match-constraint_test.go:27
        	Error:      	An error is expected but got nil.
        	Test:       	TestDebugMatchConstraintCmd
        	Messages:   	Match constraint should have errored for < 1.0, out=''
--- FAIL: TestDebugMatchConstraintCmd (0.09s)
```

---

And now I replaced `versionconstants.DdevVersion` with `ddev --version`, and your regex works fine, and `t.Skip` works.

hanoii: there's a fix on another file unrelated to this PR, assume that's OK?

stasadev: Yes, I got it from my closed PR https://github.com/ddev/ddev/pull/6361
(I forgot about it before, I hope it's okay to add it here so I don't forget again 🙂)

hanoii: That's fine with me.. I think that with no tags, on the tests at least, you get the hash, I wonder why you got nil locally. If, however, I should not that if ever end up with `v0.0.0-overridden-by-make` the test will fail because of the test expectation that > 1.0 will always succeeded and also because of what you added before that:

Locally `GOTEST_SHORT=true make VERSION="v0.0.0-overridden-by-make" testcmd TESTARGS="-run TestDebugMatchConstraintCmd"`:

```
=== RUN   TestDebugMatchConstraintCmd
    debug-match-constraint_test.go:26: 
        	Error Trace:	/Users/ariel/Sites/work/ddev/cmd/ddev/cmd/debug-match-constraint_test.go:26
        	           	exit status 1
        	Test:       	TestDebugMatchConstraintCmd
        	Messages:   	Match constraint should not have errored for != v0.0.0-overridden-by-make, out='error: your DDEV version 'v0.0.0-overridden-by-make' doesn't meet the constraint '!= v0.0.0-overridden-by-make'. Please update to a DDEV version that meets this constraint 
```

Just saying

stasadev: > t.Skip won't run because versionconstants.DdevVersion = v0.0.0-overridden-by-make when there is no tags.

I said that, but I didn't check.

> I think that with no tags, on the tests at least, you get the hash

You are right. In reality `v0.0.0-overridden-by-make` is not set in any case.

We already have a check for `!= v0.0.0-overridden-by-make` here, in another test:

https://github.com/ddev/ddev/blob/00407fd09b9633a52001f94bc280c8d336bc3747/cmd/ddev/cmd/testdata/TestCmdAddonDdevVersionConstraint/valid_constraint_recipe/install.yaml#L3

And `TestCmdAddonDdevVersionConstraint` didn't fail on your PR with and without tags.

Actually, I think it's a good thing, because it ensures that this value is always changed to something else (except when you see it yourself with `make VERSION="v0.0.0-overridden-by-make"`).

https://github.com/ddev/ddev/blob/00407fd09b9633a52001f94bc280c8d336bc3747/pkg/versionconstants/versionconstants.go#L3-L4

So if this `TestDebugMatchConstraintCmd` or `TestCmdAddonDdevVersionConstraint` test fails, we'll immediately know that there's something wrong with make.

Thank you @hanoii for paying attention to this.

rfay:
> @rfay had fun last night with this, I see it was a matter of having the tags, I think though that something like what @stasadev suggested needs to be added, otherwise this will fail for every other fork that doesn't pull in the tags, which will be a nightmare for you.

Fixing this didn't have anything to do with you pushing those tags. It was that we (in only this case) weren't using the right actions/checkout option, so didn't fetch_depth. https://github.com/ddev/ddev/commit/00407fd09b9633a52001f94bc280c8d336bc3747

I think we all looked at that early, but I was looking at the wrong yaml file, so thought it was there, and it wasn't.

rfay:
Note that `v0.0.0-overridden-by-make` means that the ddev binary was not built via make, so the git tag didn't get put in there.

stasadev:
> It was that we (in only this case) weren't using the right actions/checkout option, so didn't fetch_depth.

Okay, I see.

---

And I understand now why I got nil here https://github.com/ddev/ddev/pull/6645#discussion_r1825775682

That's because I ran the test from PhpStorm, not from the terminal, so `versionconstants.DdevVersion` is good.

I'll revert my change for `ddev --version`, sorry for the fuss.

rfay (approved):
Thanks! Sometimes these little things can go right off into the weeds. Thanks for haning with it!

PR #6529, opened 2024-09-07
fix: help mutagen work with Drupal's read-only settings files, fixes #6491

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/6491#issuecomment-2336466328

## How This PR Solves The Issue

The overall problem is an obscure interaction of Drupal install and mutagen:

* Drupal install (especially before drupal 8) setting sites/default/settings.php to read-only
* Edit settings.php on the host (which hasn't been synced)
* Mutagen can't then sync settings.php, so nothing happens inside the container, which is where it matters

## Manual Testing Instructions

```
# Drupal 7 project
ddev start
ddev drush si -y
ddev exec ls -l sites/default # See settings.php is read-only now
# Edit file on host side
ddev mutagen st # will report the transitionProblem
ddev mutagen st -l # Will show failure as in issue
ddev restart
ddev exec ls -l sites/default # See settings.php is now writeable
ddev mutagen st -l. # will now be happy
```

## Automated Testing Overview

I don't think anything needs to be added.

PR #6648, opened 2024-10-24
build: try new mutagen 0.18.0

rfay:
## The Issue

Mutagen 0.18.0 has been released, and  @xenoscopic says it may benefit us and shouldn't hurt us. This should be a good point in release cycle to try it out.

Bump mutagen version.

Use it! Watch `ddev mutagen st -l`

### DIFF
--- a/pkg/testcommon/testcommon_test.go
+++ b/pkg/testcommon/testcommon_test.go
@@ -2,11 +2,6 @@ package testcommon
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


rfay:
I think this should be OK but we sure need to get the miles on it. Could possibly interact with 
* https://github.com/ddev/ddev/pull/6529

stasadev (approved):
Looks good to me!

PR #6649, opened 2024-10-24
refactor: add `drush` and `composer` command aliases, fixes #6642

### LINKED ISSUE #6642: Create command aliases for `ddev composer` and `ddev drush`

No.

### Describe your solution

I use Composer and Drush so much that I have convenience shell aliases: `co` for `composer` and `dr` for `drush`. I _think_ those are fairly common. (I didn't come up with them, anyway.) I wonder if we could get command aliases for `ddev composer` and `ddev drush`: `ddev co` and `ddev dr` respectively.

### Describe alternatives

_No response_

### Additional context

_No response_

TravisCarden:
## The Issue

- #6642

This adds command aliases for `ddev composer` and `ddev drush`: `ddev co` and `ddev dr` respectively.

I only had a few minutes to throw this together. Did I do it right?

## Automated Testing Overview

I assume no tests are needed for a simple, non-functional change like this.

### DIFF
--- a/cmd/ddev/cmd/composer.go
+++ b/cmd/ddev/cmd/composer.go
@@ -17,6 +17,7 @@ var ComposerCmd = &cobra.Command{
 	Use:                "composer [command]",
 	Short:              "Executes a Composer command within the web container",
 	Long: `Executes a Composer command at the Composer root in the web container. Generally,
+	Aliases:            []string{"co"},
 any Composer command can be forwarded to the container context by prepending
 the command with 'ddev'.`,
 	Example: `ddev composer install
--- a/cmd/ddev/cmd/composer.go
+++ b/cmd/ddev/cmd/composer.go
@@ -17,9 +17,9 @@ var ComposerCmd = &cobra.Command{
 	Use:                "composer [command]",
 	Short:              "Executes a Composer command within the web container",
 	Long: `Executes a Composer command at the Composer root in the web container. Generally,
-	Aliases:            []string{"co"},
 any Composer command can be forwarded to the container context by prepending
 the command with 'ddev'.`,
+	Aliases:            []string{"co"},
 	Example: `ddev composer install
 ddev composer require <package>
 ddev composer outdated --minor-only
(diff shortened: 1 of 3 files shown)


stasadev (commented):
> I only had a few minutes to throw this together. Did I do it right?

Yes, thank you, and we also need to update documentation like this:

https://github.com/ddev/ddev/blob/4de65075ee0a00fad6b0ddabae91969fed33ab57/docs/content/users/usage/commands.md#L170-L172

TravisCarden:
Cool. There you go.

cmd/ddev/cmd/composer.go
```diff
@@ -17,6 +17,7 @@ var ComposerCmd = &cobra.Command{
 	Use:                "composer [command]",
 	Short:              "Executes a Composer command within the web container",
 	Long: `Executes a Composer command at the Composer root in the web container. Generally,
+	Aliases:            []string{"co"},
```
stasadev, line 20: This line is in wrong place (it is in `Long` string)

stasadev (changes requested):

cmd/ddev/cmd/composer.go
TravisCarden: Oh! Good catch. Fixed.

cmd/ddev/cmd/composer.go
```diff
@@ -19,6 +19,7 @@ var ComposerCmd = &cobra.Command{
 	Long: `Executes a Composer command at the Composer root in the web container. Generally,
 any Composer command can be forwarded to the container context by prepending
 the command with 'ddev'.`,
+	Aliases:            []string{"co"},
```
stasadev, line 22:
```
golangci-lint: 
cmd/ddev/cmd/composer.go:22: File is not `gofmt`-ed with `-s` (gofmt)
        Aliases:            []string{"co"},
make: *** [Makefile:290: golangci-lint] Error 1
```

```suggestion
	Aliases: []string{"co"},
```

cmd/ddev/cmd/composer.go
stasadev: One more issue with linting (I made a suggestion), and we can run the tests.

rfay (approved):
Thanks, and welcome to this world! Looks good, works right.

PR #6672, opened 2024-11-01
build: Improve ddev-php-base and ddev-webserver image build

rfay:
## The Issue

Preparing for working on including less PHP versions by default in the ddev-php-base image. 

I found a few things to work on:

- [ ] PHP config files were partly in ddev-php-base and partly in ddev-webserver. They need to be consolidated in ddev-php-base and removed from ddev-webserver.
- [ ] PHP 8.3 and PHP 8.4 config files didn't get fully sorted out and updated.
- [ ]


containers/ddev-php-base/generic-files/etc/php-packages.yaml
```diff
@@ -0,0 +1,66 @@
...
+  arm64: ["apcu", "bcmath", "bz2", "curl", "cli", "common", "fpm", "gd", "imagick", "intl", "ldap", "mbstring", "memcached", "mysql", "opcache", "pgsql", "readline", "redis", "soap", "sqlite3", "uploadprogress", "xdebug", "xhprof", "xml", "xmlrpc", "zip"]
+
+php81:
+  # formerly apcu bcmath bz2 curl cli common fpm gd imagick intl ldap mbstring memcached mysql opcache pgsql readline redis soap sqlite3 uploadprogress xdebug xhprof xml xmlrpc zip
+  amd64: ["apcu", "bcmath", "bz2", "curl", "cli", "common", "fpm", "gd", "imagick", "intl", "ldap", "mbstring", "memcached", "mysql", "opcache", "pgsql", "readline", "redis", "soap", "sqlite3", "uploadprogress", "xdebug", "xhprof", "xml", "xmlrpc", "zip"]
+  # formerly apcu bcmath bz2 curl cli common fpm gd imagick intl ldap mbstring memcached mysql opcache pgsql readline redis soap sqlite3 uploadprogress xdebug xhprof xml xmlrpc zip
+  arm64: ["apcu", "bcmath", "bz2", "curl", "cli", "common", "fpm", "gd", "imagick", "intl", "ldap", "mbstring", "memcached", "mysql", "opcache", "pgsql", "readline", "redis", "soap", "sqlite3", "uploadprogress", "xdebug", "xhprof", "xml", "zip"]
```
stasadev, line 47: `xmlrpc` is missing?

rfay: Thank you. I guess this means that the test at https://github.com/rfay/ddev/blob/05adf8d72a456302ef35cfd84d7c0150ccb447b0/containers/ddev-webserver/tests/ddev-webserver/php_webserver.bats#L78-L97 must be invalid. I had already missed xmlrpc in several of these and fixed, but missed this one in 8151da7b8ac3c56ecd571470c0b14d8573d01610 - But now I'm wondering why the test caught it then and didn't now. The go tests also caught it.

rfay:
Thanks for catching that @stasadev . You may be able to get a better feel of this by looking just at the major changes to Dockerfile(s) and the script and the php-packages.yaml file.

stasadev (changes requested):
>  Update PHP default version to 8.3

This value needs an update (it will probably break some tests):

https://github.com/ddev/ddev/blob/92cd790994c8df04b383956b392bd482d8124b11/pkg/nodeps/php_values.go#L18-L19

---

In everything else, looks good. The only question I have is why there are only 4 INI files in PHP 5.6-8.2?

```
$ ls containers/ddev-php-base/ddev-php-files/etc/php/5.6/mods-available/
assert.ini  blackfire.ini  xdebug.ini  xhprof.ini
```

And much more in PHP 8.3, 8.4:

```
$ ls containers/ddev-php-base/ddev-php-files/etc/php/8.3/mods-available/
apcu.ini       fileinfo.ini   msgpack.ini     readline.ini   uploadprogress.ini
assert.ini     ftp.ini        mysqli.ini      redis.ini      xdebug.ini
bcmath.ini     gd.ini         mysqlnd.ini     shmop.ini      xhprof.ini
blackfire.ini  gettext.ini    opcache.ini     simplexml.ini  xml.ini
bz2.ini        iconv.ini      pdo.ini         soap.ini       xmlreader.ini
calendar.ini   igbinary.ini   pdo_mysql.ini   sockets.ini    xmlrpc.ini
ctype.ini      imagick.ini    pdo_pgsql.ini   sqlite3.ini    xmlwriter.ini
curl.ini       intl.ini       pdo_sqlite.ini  sysvmsg.ini    xsl.ini
dom.ini        ldap.ini       pgsql.ini       sysvsem.ini    zip.ini
exif.ini       mbstring.ini   phar.ini        sysvshm.ini
ffi.ini        memcached.ini  posix.ini       tokenizer.ini
```

Do we have to maintain all these files?

Actually there is a bug in https://github.com/ddev/ddev/blob/master/containers/ddev-php-base/ddev-php-files/etc/php/8.3/mods-available/redis.ini (and in 8.4)

It needs priority, but there isn't one, see https://github.com/oerdnj/deb.sury.org/issues/2201#issuecomment-2323376757

rfay:
> Update PHP default version to 8.3

Could have sworn I already had to do this :(  Was forced to do it by broken test. But I don't see it!

> The only question I have is why there are only 4 INI files in PHP 5.6-8.2? And much more in PHP 8.3, 8.4

I'm confused as well. I guess IMO we should only do the ones that we require specific changes in. So 5.6-8.2 would probably be the strategy we'd want. 

I guess when we upgraded to 8.3, we just took all the files. Then copied then when we did 8.4. 

I'd rather only do the ones that possibly have values we want to override.

rfay:
It was foolish of me to try to change the PHP and Node.js defaults in this issue, so I have reverted those changes. Not sure what I was thinking.

rfay:
Here's hoping that my removal of the version changes didn't lose anything I had encountered along the way. It's a lot easier to look at now though.

stasadev (approved):
> IMO we should only add files for the things we're changing.

Thank you, this totally answers my question.

---

Looks good to me!

PR #6684, opened 2024-11-04
docs: note that setting a nodejs version always downloads something

deviantintegral:
## The Issue

I noticed that even if a `.nvmrc` file specifies the version of nodejs that ddev already has, that it is redownloaded and installed. I think this makes sense (so all specified versions are installed consistently), but it surprised me to see on my local that this added ~5 seconds of startup time for new projects. Surprisingly, this even happened given the global cache, so perhaps there's a bug lurking as well.

On a basic php project with no code, default settings for me give a startup time of ~14.5 seconds:

```
$ hyperfine --prepare 'ddev delete -Oy || true' 'ddev start'
Benchmark 1: ddev start
  Time (mean ± σ):     14.525 s ±  0.122 s    [User: 0.561 s, System: 0.289 s]
  Range (min … max):   14.358 s … 14.730 s    10 runs
```

But once a nodejs_version is set, it goes up:

```
$ hyperfine --prepare 'ddev delete -Oy || true' 'ddev start'
Benchmark 1: ddev start
  Time (mean ± σ):     19.828 s ±  0.636 s    [User: 0.589 s, System: 0.278 s]
  Range (min … max):   19.371 s … 21.059 s    10 runs
```

Adds the note to the docs!

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -392,7 +392,7 @@ Node.js version for the web container’s “system” version. [`n`](https://ww
 
 There is no need to reconfigure `nodejs_version` unless you want a version other than the version already specified, which will be the default version at the time the project was configured.
 
-Note that specifying any Node.js version will cause ddev to download and install that version when running `ddev start`. If optimizing startup speed is your biggest concern (such as in CI), consider only using the version of Node that is included with DDEV by default.
+Note that specifying any non-default Node.js version will cause DDEV to download and install that version when running `ddev start` the first time on a project. If optimizing first-time startup speed (as in Continuous Integration) is your biggest concern, consider using the default version of Node.js.
 
 | Type | Default | Usage
 | -- | -- | --
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -392,6 +392,8 @@ Node.js version for the web container’s “system” version. [`n`](https://ww
 
 There is no need to reconfigure `nodejs_version` unless you want a version other than the version already specified, which will be the default version at the time the project was configured.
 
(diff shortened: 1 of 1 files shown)


docs/content/users/configuration/config.md
```diff
@@ -392,6 +392,8 @@ Node.js version for the web container’s “system” version. [`n`](https://ww
 
 There is no need to reconfigure `nodejs_version` unless you want a version other than the version already specified, which will be the default version at the time the project was configured.
 
+Note that specifying any Node.js version will cause ddev to download and install that version when running `ddev start`. If optimizing startup speed is your biggest concern (such as in CI), consider only using the version of Node that is included with DDEV by default.
```
rfay, line 395: Isn't it actually this?

```suggestion
Note that specifying any non-default Node.js version will cause DDEV to download and install that version when running `ddev start` the first time on a project. If optimizing first-time startup speed (as in Continuous Integration) is your biggest concern, consider using the default version of Node.js.
```

rfay (changes requested):

stasadev:
> it surprised me to see on my local that this added ~5 seconds of startup time for new projects. Surprisingly, this even happened given the global cache, so perhaps there's a bug lurking as well.

This is an intentional behavior, I did not configure a global `n` cache for all projects. There is a cache for each project.

I chose to do this because otherwise `ddev-global-cache` would grow very quickly without the ability to clean it up with `ddev delete` (since I don't know which version of `n` is in use and which can be removed).

docs/content/users/configuration/config.md
deviantintegral: Agreed! Committed.

rfay (approved):
This works for me. But the same thing is true of so many things. Changing composer version, webimage_extra_packages. This is just one example of DDEV doing the build for you the first time and caching it for you.

Issue #6679, opened 2024-11-04
composer versions from running self-update can be stuck in the build cache


deviantintegral:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [ ] I have run `ddev debug test` to include output below

### Expected Behavior

`ddev delete -Oy && ddev start` gets you the latest version of composer.

### Actual Behavior

To get docker to rerun `composer self-update` while building the project, you need to run `docker buildx prune` first.

### Anything else?

We ran into this because we hit a composer issue in 2.8.1 and require 2.8.2 for our project to work. Our instructions were to run `composer self-update` until you needed to do a full rebuild later (or a new version of ddev came out). However, one team member kept getting composer 2.8.1.

This is because https://github.com/ddev/ddev/blob/92cd790994c8df04b383956b392bd482d8124b11/pkg/ddevapp/config.go#L1193 runs self-update but has nothing that invalidates the build if the composer version has changed.

I'm not sure how to solve this other than somehow fetching the latest composer version and saving it to a file in the docker directory that would then invalidate the build cache.

rfay:
Wouldn't the proper way to do it be `ddev config --composer-version=2.8.2 && ddev restart` ?

deviantintegral:
Doing that has the disadvantage of pinning to 2.8.2 when 2.8.3 or 2.9 comes out. On the one hand, that seems like a benefit, on the other it's yet another difference from an out-of-the-box ddev project.

It would be nice if the composer version supported semver-like strings (like https://github.com/deviantintegral/composer-gavel), but then we'd be back at the issue with stale caches.

stasadev:
> To get docker to rerun `composer self-update` while building the project, you need to run `docker buildx prune` first.

Or use `ddev debug rebuild`

> I'm not sure how to solve this other than somehow fetching the latest composer version and saving it to a file in the docker directory that would then invalidate the build cache.

The only way to update it every time an update is available is to move the `composer self-update` to [start.sh](https://github.com/ddev/ddev/blob/master/containers/ddev-webserver/ddev-webserver-base-scripts/start.sh), but I'm not sure if it's worth it.

[...]

Issue #6602, opened 2024-10-09
Project vendor/bin should have priority over /usr/local/bin in the $PATH variable.


juanjol:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Output file will be in /tmp/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/bin/ddev

These config files were loaded for project aljibe-test-project: [/home/juanjo/projects/aljibe/aljibe-test-project/.ddev/config.yaml]
name: aljibe-test-project
type: drupal
docroot: web
php_version: 8.3
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.23.4
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.11}
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 20
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
┌──────────────────┬─────────┬────────────────────────────────┬───────────────────────────┐
│ ADD-ON           │ VERSION │ REPOSITORY                     │ DATE INSTALLED            │
├──────────────────┼─────────┼────────────────────────────────┼───────────────────────────┤
│ adminer          │ v0.3.4  │ ddev/ddev-adminer              │ 2024-09-23T13:39:29+02:00 │
├──────────────────┼─────────┼────────────────────────────────┼───────────────────────────┤
│ aljibe           │ v1.0.1  │ metadrop/ddev-aljibe           │ 2024-09-23T13:39:36+02:00 │
├──────────────────┼─────────┼────────────────────────────────┼───────────────────────────┤
│ aljibe-assistant │ 1.0.0   │ Metadrop/ddev-aljibe-assistant │ 2024-09-23T13:39:35+02:00 │
├──────────────────┼─────────┼────────────────────────────────┼───────────────────────────┤
[...]

PR #6688, opened 2024-11-05
docs: Show how to use a post-start hook to set the composer version

deviantintegral:
## The Issue

- #6679 

This includes @stasadev ’s example of using a post-start hook to make sure all team members have the same composer version. This is useful in the odd case where a team needs to work around a composer bug, or require a new release for a new feature.

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -94,6 +94,14 @@ Composer version for the web container and the [`ddev composer`](../usage/comman
 | -- | -- | --
 | :octicons-file-directory-16: project | `2` | Can be `2`, `1`, or empty (`""`) for latest major version at container build time.<br><br>Can also be a minor version like `2.2` for the latest release of that branch, an explicit version like `1.0.22`, or a keyword like `stable`, `preview` or `snapshot`. See Composer documentation.
 
+Composer is updated during the first `ddev start` of a project, or when other changes like a new DDEV version require a rebuild of the web container. If it is important that every team member be on the exact same version of Composer, consider adding a post-start hook to your project such as:
+
+```yaml
+hooks:
+    post-start:
+        - exec: composer self-update 2.8.2
+```
+
 !!!note "If your `composer.json` requires `composer/composer` that version will be used instead"
     If your project `composer.json` includes `composer/composer`, then the version specified there will normally be used instead of any version specified by `composer_version`, since `vendor/bin/composer` will come first in the in-container `$PATH`.
 


docs/content/users/configuration/config.md
```diff
@@ -94,6 +94,14 @@ Composer version for the web container and the [`ddev composer`](../usage/comman
 | -- | -- | --
 | :octicons-file-directory-16: project | `2` | Can be `2`, `1`, or empty (`""`) for latest major version at container build time.<br><br>Can also be a minor version like `2.2` for the latest release of that branch, an explicit version like `1.0.22`, or a keyword like `stable`, `preview` or `snapshot`. See Composer documentation.
 
+Composer is updated during the first `ddev start` of a project, or when other changes like a new DDEV version require a rebuild of the web container. If it is important that every team member be on the exact same version of Composer, consider adding a post-start hook to your project such as:
+
+```yaml
+hooks:
+    post-start:
+        - exec: composer self-update 2.8.2
```
stasadev, line 102: The only situation I can think of where this `post-start` hook might be useful is when you have:

```
ddev config --composer-version "2"
```

Then decide to pin a version to:
```
ddev config --composer-version "2.8.2"
```
(Which builds a new image.)

And suddenly rollback to
```
ddev config --composer-version "2"
```
(and here Docker will not build a new image, it will use the cache from the beginning.)

But with this hook, Docker cache won't matter.

---

But this `post-start` hook has disadvantage of needing internet connection and downloading Composer on each `ddev start` (if the version inside differs).

---

What about a temporary pin
```
ddev config --composer-version "2.8.2" && ddev restart
```

And when a new DDEV release comes out, and you'll want to use the latest stable Composer again:
```
ddev config --ddev-version-constraint ">= v1.24.0" --composer-version "2" && ddev restart
```
The constraint will ensure that no Docker cache will be used, because you will be forced to update.

stasadev:
> I wonder if we should remove or update that note? It seems like if ddev automatically prioritizes vendor/bin/composer that’s a problem for projects that use composer as an API.

And here people asked why it doesn't prioritize `vendor/bin/composer`, if inside the container (`ddev ssh`), the `composer`  was different:

- #6602

> It seems like if ddev automatically prioritizes vendor/bin/composer that’s a problem for projects that use composer as an API.

I think they can use a different `post-start` hook 🙂 (`~/bin` is the first item in the `PATH`):

```
hooks:
    post-start:
        - exec: mkdir -p ~/bin && ln -sf /usr/local/bin/composer ~/bin/composer
```

rfay (commented):
I suspect this might be better as a great question-and-answer in Stack Overflow. Here it's probably TMI. There are so many ways to solve this, and it's such an edge case, and it's difficult mostly because of your own requirements @deviantintegral .  And... it was one problematic release of composer out of so many.

deviantintegral:
That's fine. At least there's something now that can be found in search.

Future readers, if you come across this problem space in the future and this would have saved you time, feel free to leave a note!

Issue #6628, opened 2024-10-18
v1.23.5 regression: `ddev composer` command error when passing JSON or arg with spaces


stasadev:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
```
</details>

### Expected Behavior

`ddev composer` should work as `composer` inside the container.

### Actual Behavior

Since today (1.23.5), it fails with this message: 

```
In ConfigCommand.php line 766:You must pass the type and a url. Example: php composer.phar config reposit  ories.foo vcs https://bar.com
```
As I'm passing a type and url, I suspect there are lost somewhere.

### Steps To Reproduce

From Discord https://discord.com/channels/664580571770388500/1296719909367054420

```
ddev composer config repositories.0 "{\"type\": \"composer\", \"url\":\"hello\",  \"exclude\": [\"magento/composer-dependency-version-audit-plugin\"]}"
```

(We can pass a json)

### Anything else?

I've just found a workaround: 

```
ddev exec --raw composer  config repositories.0 "{\"type\": \"composer\", \"url\":\"hello\",  \"exclude\": [\"magento/composer-dependency-version-audit-plugin\"]}"
```

stasadev:
The fix is available in:

- https://github.com/ddev/ddev/pull/6627#issuecomment-2422346715

rfay:
Aw, we did look at that with a little bit of worry :(

rfay:
I don't understand:
* Why I took the risk of changing that
* Why I *needed* to take the risk of changing from raw (we should look together)
* How this class of problem can haunt us for years and years :(

stasadev:
> Why I took the risk of changing that
> Why I *needed* to take the risk of changing from raw (we should look together)

Because only `Cmd` uses `$PATH` from `bash`:

[...]

PR #6689, opened 2024-11-05
fix: respect `~/bin` for composer commands, for #6627

stasadev:
## The Issue

This comment:

- https://github.com/ddev/ddev/pull/6688#issuecomment-2457253832

For:

- #6627

It turned out that when you pass `PATH=~/bin:...` as env, it doesn't understand what `~` means.

```
ddev composer require composer/composer:v2.8.0 --dev
```

Add this to `.ddev/config.yaml`:

```
hooks:
    post-start:
        - exec: mkdir -p ~/bin && ln -sf /usr/local/bin/composer ~/bin/composer
```

```
ddev restart
ddev composer -V # should be the same as in /usr/local/bin/composer
```


rfay:
Why was `~/bin/composer` in use here anyway (in addition to vendor/bin/composer and /usr/local/bin/composer)?

stasadev:
> Why was `~/bin/composer` in use here anyway (in addition to vendor/bin/composer and /usr/local/bin/composer)? 

It's the first item in the `$PATH`, I guess we should respect it, and let people to do such overrides for `composer`:

https://github.com/ddev/ddev/blob/92cd790994c8df04b383956b392bd482d8124b11/containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc#L1

rfay:
Sorry, my question is not whether it should work, but why we would have used it as a composer instance.  As we've seen with the `ddev` binary, having it in multiple places can only lead to confusion. So I was asking why we would ever put composer in ~/bin to begin with.

stasadev:
> why we would ever put composer in ~/bin to begin with.

To use a different `composer`, as @deviantintegral said in https://github.com/ddev/ddev/pull/6688#issuecomment-2457114357 what if people have `composer/composer` in their `composer.json`, but they use it for a different purpose.

Even though I don't use it this way, I see a problem here as well, for example, after installing it, you can't uninstall it because it can't uninstall itself:

```
ddev composer require composer/composer:v2.8.0 --dev
ddev composer remove composer/composer --dev # throws error
```

docs/content/users/configuration/config.md
```diff
@@ -95,7 +95,13 @@ Composer version for the web container and the [`ddev composer`](../usage/comman
 | :octicons-file-directory-16: project | `2` | Can be `2`, `1`, or empty (`""`) for latest major version at container build time.<br><br>Can also be a minor version like `2.2` for the latest release of that branch, an explicit version like `1.0.22`, or a keyword like `stable`, `preview` or `snapshot`. See Composer documentation.
 
 !!!note "If your `composer.json` requires `composer/composer` that version will be used instead"
-    If your project `composer.json` includes `composer/composer`, then the version specified there will normally be used instead of any version specified by `composer_version`, since `vendor/bin/composer` will come first in the in-container `$PATH`.
+    If your project `composer.json` includes `composer/composer`, then the version specified there will normally be used instead of any version specified by `composer_version`, since `vendor/bin/composer` will come first in the in-container `$PATH`. However, if you prefer to use the version specified by `composer_version`, add the following hook to your `.ddev/config.yaml`:
```
rfay, line 98: This seems like a very unusual workaround. Why would people not want to use the version they specify in their composer.json?

I worry that adding too much detail in places like this just makes it so people can't make it through the docs, or find the right thing when they need it.

pkg/ddevapp/composer.go
```diff
@@ -20,8 +20,9 @@ func (app *DdevApp) Composer(args []string) (string, string, error) {
 	// Prevent Composer from debugging when Xdebug is enabled
 	env := []string{"XDEBUG_MODE=off"}
 	// Let Composer know which binary to run from the PATH
+	// And expand `~` in the PATH, because it is not expanded in this situation
 	path, _, err := app.Exec(&ExecOpts{
-		Cmd: "echo $PATH",
+		Cmd: `echo $PATH | sed "s|~|$HOME|g"`,
```
rfay, line 25: I wonder if homedir.Expand() could be used successfully here. Probably not, since this is in the context of the container. But transformations like this are notoriously fragile. 

(I don't understand what homedir.Expand() replaces the ~ with. It it's $HOME, it's good, I don't think it is.)

rfay (changes requested):
Minor suggestions. 

I generally don't think we should add every workaround for every obscure case we encounter to the docs, because they get cluttered. I usually point to Stack Overflow for cases like that.

docs/content/users/configuration/config.md
deviantintegral: > Why would people not want to use the version they specify in their composer.json?

It's less about matching the version, and more about issues upgrading composer itself since PHP can't reload classes it's already loaded: https://github.com/composer/composer/discussions/11672

I wonder if this should also be improved upstream. Perhaps composer should never install `vendor/bin/composer`, or if it needs to it should be something like `vendor/bin/composer.internal`.

pkg/ddevapp/composer.go
stasadev: This PR is not only about Composer binary override, but also if in the future we have some commands that will work just like `ddev composer` with `PATH`.

This entry in `PATH=~/bin` should really be `PATH=$HOME/bin`, then we don't need these ugly workarounds.

I'll also look into the `PATH` problem with subshells while I'm here:

- https://github.com/ddev/ddev/issues/6628#issuecomment-2428795844

docs/content/users/configuration/config.md
stasadev: > I worry that adding too much detail in places like this just makes it so people can't make it through the docs, or find the right thing when they need it.

I think if people use Composer in such an advanced way, they'll know how to change something in `$PATH`.
I'll remove this example from the docs because it's unlikely to be found.

containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
```diff
@@ -1,4 +1,14 @@
-export PATH="~/bin:$PATH:/var/www/html/bin"
+case ":$PATH:" in
+    *":$HOME/bin:"*) ;;
+    *) PATH="$HOME/bin:$PATH" ;;
```
stasadev, line 3: So the fix is to write `$HOME/bin` instead of `~/bin:$PATH` just like Debian does.

rfay:
Rebased for colima/lima problem

containers/ddev-webserver/ddev-webserver-base-files/etc/bash.nointeractive.bashrc
```diff
@@ -1,7 +1,13 @@
 # This file is loaded in non-interactive bash shells through $BASH_ENV
-export PATH="${DDEV_COMPOSER_ROOT:-/var/www/html}/vendor/bin:${PATH}"
+case ":$PATH:" in
+    *":${DDEV_COMPOSER_ROOT:-/var/www/html}/vendor/bin:"*) ;;
+    *) PATH="${DDEV_COMPOSER_ROOT:-/var/www/html}/vendor/bin:$PATH" ;;
+esac
```
rfay, line 5: I think this is a fine solution, but it's really, really cryptic. Please add a comment to both places explaining what it does.

containers/ddev-webserver/ddev-webserver-base-files/etc/bashrc/commandline-addons.bashrc
rfay: This also needs a comment explaining what it's doing.

But it's great, and an excellent solution.

rfay (changes requested):
Minor suggestions just to make sure future readers can grok what's going on.

stasadev:
Comments added (without a new image rebuild).

rfay (approved):
Thanks!

rfay:
Rebased, fixed versionconstants, needs image push. Could you push the ddev-webserver image or give me perms on your fork?

PR #6693, opened 2024-11-06
test: run full tests that are sometimes short-circuited

rfay:
## The Issue

In studying the test run and coverage for 
* https://github.com/ddev/ddev/pull/6691

I was a little suspicious that not all PHP versions had been tested. 

Sure enough, we have a number of costly tests that only run the full set of possibilities when GOTEST_SHORT is empty.

So the only test that would have tried all my PHP versions was TestPHPConfig, and it was short-circuited in all our tests.

We only run the most costly tests without short circuits in GitHub tests.yml `pull-push-test-platforms`, so the tests that are like this have to be explicitly mentioned there. 

Then I discovered that the logic in some tests (most importantly TestPHPConfig) was incorrect, so when GOTEST_SHORT="" it wasn't running *any* of the PHP versions.

* Look for tests of this behavior
* Add them to the list
* Reorganize the list so it's more readable.
* Fix tests that weren't running the right thing on GOTEST_SHORT=

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -124,8 +124,26 @@ jobs:
       - name: Override environment variables for push-pull-test-platforms
         run: |
           echo "MAKE_TARGET=test" >> $GITHUB_ENV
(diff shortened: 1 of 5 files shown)


stasadev (approved):
Looks good.

PR #6695, opened 2024-11-06
feat: add WordPress WP_ENVIRONMENT_TYPE support to wp-config-ddev.php

graham73may:
## The Issue

`wp_get_environment_type()` returns `production` (which is expected as nothing is telling WordPress otherwise). This PR sets DDEV projects to identify themselves as a `local` environment.

This PR defines your DDEV local site as a local environment using the `WP_ENVIRONMENT_TYPE` constant in the `wp-config-ddev.php` file. 

As outlined here: 
https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/

Execute `wp_get_environment_type()` anywhere and it should return `local`. 

## Automated Testing Overview

N/A

## Release/Deployment Notes

This would only affect WordPress projects using `wp_get_environment_type()` and expecting their local site to be one of the other values `development | staging | production`

### DIFF
--- a/pkg/ddevapp/wordpress/wp-config-ddev.php
+++ b/pkg/ddevapp/wordpress/wp-config-ddev.php
@@ -29,7 +29,7 @@ if ( getenv( 'IS_DDEV_PROJECT' ) == 'true' ) {
 	defined( 'WP_DEBUG' ) || define( 'WP_DEBUG', true );
 
 	/** WordPress environment type. */
-	define( 'WP_ENVIRONMENT_TYPE', 'local' );
+	defined( 'WP_ENVIRONMENT_TYPE' ) || define( 'WP_ENVIRONMENT_TYPE', 'local' );
 
 	/**
 	 * Set WordPress Database Table prefix if not already set.
--- a/pkg/ddevapp/wordpress/wp-config-ddev.php
+++ b/pkg/ddevapp/wordpress/wp-config-ddev.php
@@ -28,6 +28,9 @@ if ( getenv( 'IS_DDEV_PROJECT' ) == 'true' ) {
 	/** Enable debug */
 	defined( 'WP_DEBUG' ) || define( 'WP_DEBUG', true );
 
+	/** WordPress environment type. */
+	define( 'WP_ENVIRONMENT_TYPE', 'local' );
+
 	/**
 	 * Set WordPress Database Table prefix if not already set.
 	 *


rfay:
So the possible values are these?
> This would only affect WordPress projects using wp_get_environment_type() and expecting their local site to be one of the other values development | staging | production

Why "local" if the values might include "development" and not "local"?

graham73may:
Correct, possible values are: 
* local
* development
* staging
* production

I opted for `local` on this PR because DDEV is primarily a local development tool (RIP DDEV Live) and because some hosting providers have dev / staging / production environments baked in.
* [Pantheon](https://docs.pantheon.io/guides/environment-configuration) - dev / test / live
* [WP Engine](https://wpengine.com/support/environments/#About_Environments) - dev / staging / production

rfay:
Thanks!

stasadev (approved):
I think `local` is good because:

> Note: When `development` is returned by `wp_get_environment_type()`, `WP_DEBUG` will be set to `true` if it is not defined in the site’s `wp-config.php` file.

https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/

graham73may:
@nickchomey very interesting approach! Were relying on ddev core to do a few of these I think. 

For URL changing we handle this with a post import hook in dev’s config.yaml. We’ve also got some other hooks to set up node packages for build processes, composer packages, etc. 

With hooks and our ddev provider scripts we’ve got it to a sweet spot where we can run ddev pull and it will:
- Pull a database off of the environment 
- Import the database
- Find and replace URLs
- Install development plugins (stage file proxy) and configure it
- Disable SMTP plugins to stop emails leaking out
- Install composer packages
- Install node packages required for the build processes

Mostly everything we need to get running is automated ❤️

graham73may:
@nickchomey It's not really a "package" to make public, but essentially this is an example of what's in our `config.yaml`: 

```
hooks:
  post-start:
    - composer: install
    - exec-host: npm install --prefix wp-content/themes/example/assets/src
    - exec-host: cd wp-content/themes/example/assets/src && npx mix --production
  post-import-db:
    - exec: wp search-replace https://www.example.com https://example.com.ddev.site
    - exec: wp plugin deactivate wp-mail-smtp
    - exec: wp plugin install stage-file-proxy --activate
    - exec: wp eval "update_option('stage-file-proxy-settings', ['source_domain'=>'https://www.example.com','method'=>'redirect']);"
```

graham73may:
I recently got really into them. Would recommend taking a look at this page: https://ddev.readthedocs.io/en/stable/users/configuration/hooks/

There’s way more hooks and tasks than I thought! 👍🏻

rfay:
Since we're off in the weeds...

```
    - exec-host: npm install --prefix wp-content/themes/example/assets/src
    - exec-host: cd wp-content/themes/example/assets/src && npx mix --production
  post-import-db:
```

I don't recommend using npm on the host as this does, better to use `ddev npm` or put it in your exec hook. Otherwise you may be installing things using one npm and using them with another. Better to do everything inside the container.

```
  post-import-db:
    - exec: wp search-replace https://www.example.com https://example.com.ddev.site
    - exec: wp plugin deactivate wp-mail-smtp
    - exec: wp plugin install stage-file-proxy --activate
    - exec: wp eval "update_option('stage-file-proxy-settings', ['source_domain'=>'https://www.example.com','method'=>'redirect']);"
```
This would be simpler and perhaps more efficient if you rolled it into a script and executed the script. 

But wow, there's a stage-file-proxy for WP??? YAY!

graham73may:
@rfay I’m using npm on the host so I can execute my build tools using Phpstorm’s UI. I’ve tried really hard to get Phpstorm to use a remote node interpreter (is that the right name for it?) but last I checked I think I ran into a  YouTrack bug/feature request for it. 

It does work inside my container if I set it up that way and I can fire a watch task and leave it running, but I’ve found I like my build tool controls in my IDE.

rfay:
PhpStorm is bad with node, agreed. Do it as you will of course, just wanted to point out the team risks there.

graham73may:
Of course 🙂. We’re all comfortable with NVM so it’s not been too problematic I’m glad to say!

timnolte:
This should have never been merged in. This has messed up our workflow. There should be a way to prevent DDev from forcing this. There is no reason that people can't just set this themselves in the `.env`.

graham73may:
Sorry you're facing issues @timnolte.

As a quick fix if you remove `* #ddev-generated: Automatically generated WordPress settings file.` from your `wp-config-ddev.php` file it won't force it.

timnolte:
> Sorry you're facing issues @timnolte.
> 
> As a quick fix if you remove `* #ddev-generated: Automatically generated WordPress settings file.` from your `wp-config-ddev.php` file it won't force it and you can manage your own config.

@graham73may I'm aware of that however we have over 50 sites that will now all need manual fixes to account for this change.

PR #6702, opened 2024-11-08
test: Revert "Try to get lima and colima sane with LIMA_SSH_PORT_FORWARDER (#6697)"

rfay:
## The Issue

* #6697

Lima 1.0.0 port forwarder broke tests, it was fixed via #6697, but now in 1.0.1 they reverted that, so we can revert the workaround.


rfay:
This still seems to hang without the override.

PR #6704, opened 2024-11-08
fix: match snapshot using full filename, fixes #6694

### LINKED ISSUE #6694: Deleting a snapshot might delete a different snapshot with the same prefix
When one snapshot's name consists of another snapshot's name followed by a dash, it might be erroneously selected during cleanup. The following shell history shows how I create two snapshots, `foo` and `foo-bar`, and then try to delete `foo`, but `foo-bar` is deleted instead.

```bash
$ ddev snapshot -n foo
$ ddev snapshot -n foo-bar
$ ddev snapshot -l
┌──────────┬────────────┐
│ SNAPSHOT │ CREATED    │
├──────────┼────────────┤
│ foo-bar  │ 2024-11-06 │
├──────────┼────────────┤
│ foo      │ 2024-11-06 │
└──────────┴────────────┘
$ ddev snapshot -n foo -C
$ ddev snapshot -l
┌──────────┬────────────┐
│ SNAPSHOT │ CREATED    │
├──────────┼────────────┤
│ foo      │ 2024-11-06 │
└──────────┴────────────┘
$ ddev -v
ddev version v1.23.5
```

castanearie:
## The Issue

- #6694

Only matching a snapshot file by prefix can accidentally select a different snapshot that is lexicographically earlier.

Now a fully anchored regular expression is used to match the full filename. The regular expression is based on the one in `ListSnapshots`.

Create the snapshots `foo` and `foo-bar`. Attempt to delete the snapshot `foo`. List snapshots to see which actually got deleted.

## Automated Testing Overview

Such a small change didn't seem like it needed its own test. If a test is desired anyway, `TestDdevSnapshotCleanup` can probably be extended to cover this case as well.

## Release/Deployment Notes

Should not have any effects or require any action.

### DIFF
--- a/pkg/ddevapp/snapshot_test.go
+++ b/pkg/ddevapp/snapshot_test.go
@@ -44,7 +44,7 @@ func TestDdevSnapshotCleanup(t *testing.T) {
 	snapshotName, err := app.Snapshot(t.Name() + "_1")
 	assert.NoError(err)
 
-	// Provide a second snapshot that should not be deleted and whose filename will be lexicographically earlier (Issue #6694).
+	// Provide a second snapshot that should not be deleted and whose filename will be lexicographically earlier (See https://github.com/ddev/ddev/issues/6694).
 	_, err = app.Snapshot(t.Name() + "_1-a")
 	assert.NoError(err)
 
--- a/pkg/ddevapp/snapshot_test.go
+++ b/pkg/ddevapp/snapshot_test.go
@@ -44,6 +44,10 @@ func TestDdevSnapshotCleanup(t *testing.T) {
 	snapshotName, err := app.Snapshot(t.Name() + "_1")
 	assert.NoError(err)
 
+	// Provide a second snapshot that should not be deleted and whose filename will be lexicographically earlier (Issue #6694).
+	_, err = app.Snapshot(t.Name() + "_1-a")
+	assert.NoError(err)
+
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Works well for me too, thank you!

pkg/ddevapp/snapshot_test.go
```diff
@@ -44,6 +44,10 @@ func TestDdevSnapshotCleanup(t *testing.T) {
 	snapshotName, err := app.Snapshot(t.Name() + "_1")
 	assert.NoError(err)
 
+	// Provide a second snapshot that should not be deleted and whose filename will be lexicographically earlier (Issue #6694).
```
stasadev, line 47: Let's use the full URL here because IDE won't understand what `#6694` is.

```suggestion
	// Provide a second snapshot that should not be deleted and whose filename will be lexicographically earlier (See https://github.com/ddev/ddev/issues/6694).
```

castanearie:
Extended `TestDdevSnapshotCleanup` and confirmed that it fails without the fix.

pkg/ddevapp/snapshot_test.go
castanearie: should this be an amend or a new commit?

stasadev: Doesn't matter, we squash everything in one commit when it lands in master.

PR #6706, opened 2024-11-09
fix: use /usr/bin/env bash instead of direct /bin/bash in scripts

rfay:
## The Issue

After many years of knowing the option of using `#!/usr/bin/env bash` at the top of bash scripts, I finally hit a system where it mattered, 

See discord: https://discord.com/channels/664580571770388500/1303975088605102110

In nixOS, /bin/bash is not provided. This causes much weeping and gnashing of teeth. 

But there's no reason we can't use the standard technique.

Change `#!/bin/bash` and `#!/bin/sh`

## TODO

This still needs an image push for all the affected images.


cmd/ddev/cmd/testdata/TestCmdAuthSSH/ddevauthssh.expect
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env expect -f
+#!/usr/bin/env -S expect -f
```
stasadev, line 1: For reference:
```
$ env --help
  -S, --split-string=S  process and split S into separate arguments;
                        used to pass multiple arguments on shebang lines
```

stasadev (approved):
Looks good to me.

PR #6707, opened 2024-11-09
fix: not-included PHP version should work with webimage_extra_packages

rfay:
## The Issue

After 
* https://github.com/ddev/ddev/pull/6691

The ddev-platformsh add-on tests started failing on HEAD, https://github.com/ddev/ddev-platformsh/actions/runs/11758075483/job/32757492720

I had to use tmate to debug, but it turned out that ddev-platformsh installs several php modules for drupal9 and drupal10. And those use php8.0, which is not there by default. 

After they install modules, /var/lib/php/modules is not writable, and ddev fails to start.

* chmod /var/lib/php later
* Add some test coverage by adding redis.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1280,7 +1280,7 @@ fi`, app.Database.Version, psqlVersion) + "\n\n"
 		contents = contents + fmt.Sprintf(`
 ### DDEV-injected folders permission fix
 RUN chmod 777 /run/php /var/log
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Very nice, it's amazing that we can catch this kind of things with add-ons.

rfay:
The [ddev-platformsh builds](https://github.com/ddev/ddev-platformsh/actions/runs/11783195430) are green again, yay!

PR #6708, opened 2024-11-09
test: Fix TestGetAllURLs failure on tb-wsldd-16

rfay:
## The Issue

TestGetAllURLs has been failing consistently with 5 URLs instead of 10 on tb-wsldd-16.  https://buildkite.com/ddev/wsl2-docker-desktop/builds/5721#01930e11-788c-4527-ab15-2db459abd3ec

This adds a little output on the fail that might help us understand why.


rfay:
The global config `mkcert_caroot` was set to empty in ~/.ddev/global_config.yaml, so as a result it was only using http urls, not https. And the expectation was backward in the assert, making it look like we were expecting 5 and getting 10, but we were expecting 10 and getting 5.

I deleted that line, and it was properly recreated.

Improved the resiliency of the test as well.

stasadev (approved):
Looks good to me.

PR #6710, opened 2024-11-11
build: fix static errors from golangci-lint

stasadev:
## The Issue

golangci-lint 1.62.0 added more checks:

```
golangci-lint: 
pkg/globalconfig/global_config_test.go:40:2: redefines-builtin-id: redefinition of the built-in function max (revive)
        max := i + 100
        ^
pkg/ddevapp/provider.go:304:61: redefines-builtin-id: redefinition of the built-in type error (revive)
func (p *Provider) doFilesPullCommand() (filename []string, error error) {
                                                            ^
pkg/ddevapp/provider.go:327:61: redefines-builtin-id: redefinition of the built-in type error (revive)
func (p *Provider) getDatabaseBackups() (filename []string, error error) {
                                                            ^
make: *** [Makefile:290: golangci-lint] Error 1
```

Fixes it.

### DIFF
--- a/pkg/ddevapp/provider.go
+++ b/pkg/ddevapp/provider.go
@@ -301,7 +301,7 @@ func (p *Provider) getDownloadDir() string {
 	return destDir
 }
 
-func (p *Provider) doFilesPullCommand() (filename []string, error error) {
+func (p *Provider) doFilesPullCommand() ([]string, error) {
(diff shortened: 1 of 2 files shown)


rfay (approved):
Thanks, we should pull as soon as static test succeeds.

PR #6711, opened 2024-11-11
fix: use TARGETARCH instead of TARGETPLATFORM in user Dockerfiles

stasadev:
## The Issue

From Drupal Slack https://drupal.slack.com/archives/C5TQRQZRR/p1731341586120089

```
Warning: command 'bash -c mkdir -p /tmp/sqlite3 && wget -O /tmp/sqlite3/sqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_3.45.1-1_arm64/v8.deb && wget -O /tmp/sqlite3/libsqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/libsqlite3-0_3.45.1-1_arm64/v8.deb && apt-get install -y /tmp/sqlite3/*.deb && rm -rf /tmp/sqlite3' run as 'root' failed with exit code 8:
--2024-11-11 16:30:05--  https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_3.45.1-1_arm64/v8.deb
Resolving snapshot.debian.org (snapshot.debian.org)... 185.213.153.170
Connecting to snapshot.debian.org (snapshot.debian.org)|185.213.153.170|:443... connected.
HTTP request sent, awaiting response... 404 NOT FOUND
2024-11-11 16:30:05 ERROR 404: NOT FOUND. 
```

Looking at that URL, I suspect that the Docker build engine is using `arm64/v8` as the target platform, which makes sense, but that maybe makes the URL of the SQLite `.deb` wrong.
If I remove the `/v8` from that URL in my browser, it actually tries to download something.
Is there a way I can override the `TARGETPLATFORM` build argument in `.ddev/config.yaml` to force it to be `arm64` , as a workaround?

From https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/#adding-extra-dockerfiles-for-webimage-and-dbimage, it looks like maybe the workaround that installs SQLite should be using `TARGETARCH` instead of `TARGETPLATFORM`.

Uses `TARGETARCH` in the user Dockerfile, this change is also important for the newly added 'install_php_extensions.sh'

Should pass the tests.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -985,8 +985,8 @@ ENV N_INSTALL_VERSION="%s"
 ### Drupal 11+ requires a minimum sqlite3 version (3.45 currently)
 ARG SQLITE_VERSION=%s
 RUN log-stderr.sh bash -c "mkdir -p /tmp/sqlite3 && \
-wget -O /tmp/sqlite3/sqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && \
-wget -O /tmp/sqlite3/libsqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/libsqlite3-0_${SQLITE_VERSION}-1_${TARGETPLATFORM##linux/}.deb && \
+wget -O /tmp/sqlite3/sqlite3.deb https://snapshot.debian.org/archive/debian/20240203T152533Z/pool/main/s/sqlite3/sqlite3_${SQLITE_VERSION}-1_${TARGETARCH}.deb && \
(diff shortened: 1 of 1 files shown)


phenaproxima:
Can confirm in manual testing that this worked.

stasadev:
@rfay, I also want to point out that `snapshot.debian.org` can be fragile, I hit this problem with `ddev-python2` add-on (the link was not available for 2 days, until I "fixed" it by downloading python2 from `ftp.debian.org`:

- https://github.com/stasadev/ddev-python2/issues/2

Maybe we need to bundle the sqlite3.45.1 binaries to `ddev-webserver`, similar to this:

https://github.com/ddev/ddev/blob/924ad6f11ae26cdd6fe26c062cefabfb1a9928fc/containers/ddev-webserver/Dockerfile#L70-L75

Let me know what you think about it.

rfay:
Now that we sorted out what we were chasing, I think TARGETARCH might have been `arm64/v8`. But then when he tested the PR it would have failed the same way.

(He had an ancient Colima with an ancient docker server version in it, that apparently passed TARGETPLATFORM=linux/arm64/v8)

As far as I can tell, TARGETPLATFORM in this context should only have been linux/amd64 or linux/arm64.

rfay (approved):
I think we should go with this because it's simpler and more elegant.

PR #6712, opened 2024-11-11
refactor: use `n install --cleanup`

stasadev:
## The Issue

- https://github.com/tj/n/issues/809#issuecomment-2465891240

Uses new `n install --cleanup` flag instead of `n install && n rm`.
This is the same thing, but with only one command.

```
$ ddev exec 'ls -la /usr/local/n/versions/node' # should be no files here
total 8
drwxr-xr-x 2 root root 4096 Nov 12 20:48 .
drwxr-xr-x 3 root root 4096 Nov 12 20:48 ..

$ ddev exec 'sudo n install --cleanup 20'
installing : node-v20.18.0
       mkdir : /usr/local/n/versions/node/20.18.0
       fetch : https://nodejs.org/dist/v20.18.0/node-v20.18.0-linux-x64.tar.xz
     copying : node/20.18.0
   installed : v20.18.0 (with npm 10.8.2)
     cleanup : removing cached node/20.18.0

$ ddev exec 'ls -la /usr/local/n/versions/node' # should be no files here
drwxr-xr-x 1 root root 4096 Nov 12 22:37 .
drwxr-xr-x 1 root root 4096 Nov 12 20:48 ..
```

### DIFF
--- a/containers/ddev-php-base/Dockerfile
+++ b/containers/ddev-php-base/Dockerfile
@@ -100,8 +100,8 @@ RUN apt-get -qq install --no-install-recommends --no-install-suggests -y \
     sqlite3
 
 RUN curl -L --fail -o /usr/local/bin/n -sSL https://raw.githubusercontent.com/tj/n/master/bin/n && chmod ugo+wx /usr/local/bin/n
-# Install node, remove it from cache, make a symlink for nodejs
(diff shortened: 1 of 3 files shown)


rfay (commented):
This is great, but does need pushed ddev-php-base and ddev-webserver.

stasadev:
Instructions updated.

rfay (approved):
It looks good and all works for me in manual testing, thanks!

PR #6678, opened 2024-11-04
feat: allow files and symlinks in `ddev auth ssh`, fixes #5465, fixes #6677

stasadev:
## The Issue

- #5465

- #6677

## How This PR Solves The Issue

- Allows single file or symlink to a file or directory to be passed in `ddev auth ssh`.
- Fixes a warning from `grep -l` by mounting only files.
- Allows to mount several files and directories.

I didn't add a new flag for the file, I decided it would only complicate the code.

## Manual Testing Instructions

```
mkdir ~/temp_ssh
cp ~/.ssh/id_ed25519 ~/temp_ssh/id_ed25519
cp ~/.ssh/id_ed25519 ~/temp_ssh/id_ed25519_copy
ln -s ~/.ssh/id_ed25519 ~/temp_ssh/id_ed25519_symlink
ln -s ~/.ssh ~/temp_ssh/.ssh

ddev auth ssh -d ~/.ssh
ddev auth ssh -d ~/temp_ssh
ddev auth ssh -d ~/temp_ssh/.ssh
ddev auth ssh -f ~/temp_ssh/id_ed25519
ddev auth ssh -f ~/temp_ssh/id_ed25519_copy -f ~/temp_ssh/id_ed25519_symlink

# also works if there are files with the same name
ddev auth ssh -f ~/temp_ssh/id_ed25519 -f ~/.ssh/id_ed25519

rm -r ~/temp_ssh
```

PR #6720, opened 2024-11-13
refactor: remove `docker run` from `ddev auth ssh`

stasadev:
## The Issue

I made initial attempt to remove `docker run` in:

- #6678

But couldn't get it work, when `ddev auth ssh` required a passphrase for provided keys.

The code from `docker run` may help:

- https://github.com/docker/cli/blob/master/cli/command/container/run.go#L252-L298

It still doesn't work for keys with passphrase, but works for keys without it.

```
# create a key with passphrase
ssh-keygen -t rsa -b 4096 -C "test@example.com" -f $HOME/.ssh/test_key

# it fails, because cannot accept input.
ddev auth ssh ~/.ssh/test_key
Command 'bash -c "cp -r /tmp/sshtmp ~/.ssh && chmod -R go-rwx ~/.ssh && cd ~/.ssh && grep -l '^-----BEGIN .* PRIVATE KEY-----' * | xargs -d '\n' ssh-add"' failed: container run failed with exit code 123; output='Enter passphrase for id_rsa:'
```

`ddev auth ssh` should accept input for passphrases in all provided SSH keys.

### DIFF
--- a/cmd/ddev/cmd/auth-ssh.go
+++ b/cmd/ddev/cmd/auth-ssh.go
@@ -4,12 +4,15 @@ import (
 	"fmt"
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/docker"
-	"github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/fileutil"
 	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/heredoc"
 	"github.com/ddev/ddev/pkg/nodeps"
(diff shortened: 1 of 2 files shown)


stasadev:
@rfay, I still don't know how to make it work correctly.
You might be able to take a look into it. Thank you.

rfay:
I don't know that I'll do any better, but I sure appreciate you exploring the territory and trying to make this cleaner.

rfay:
I agree we should leave this alone for now. I also spent quite a bit of time on this. The answer seems to be to open in tty mode with raw tty, because of course ssh-add  doesn't use stdin to read the password, it reads directly from tty as it should. I imagine that we could study the docker/cli code more (I spent a little time with it). But it's hard to spend time on something like this that already works and has worked for ages. Thanks for the courageous attempt on this.

PR #6725, opened 2024-11-14
feat: add Symfony project type

IndraGunawan:
## The Issue

using default `php` project type works well, but several Symfony specific actions needs to be done manually

### Default web environments

On `ddev start` command, some environments will be automatically added to `.env.local` file.

`MAILER_*` to use [Mailpit](https://ddev.readthedocs.io/en/stable/users/usage/developer-tools/#email-capture-and-review-mailpit) as mail catcher.
`DATABASE_*` based on database selection unless [--omit-containers=db](https://ddev.readthedocs.io/en/stable/users/configuration/config/#omit_containers) is set.
`REDIS_*` if `ddev/ddev-redis` or `ddev/ddev-redis-7` add-on installed.

### Want to run Symfony Console (bin/console)?

```bash
ddev console list
# ddev console doctrine:schema:update --force
```

### Consuming Messages (Running the Worker)

Edit `.ddev/config.yaml` in your project directory and uncomment `post-start` hook
To see `messenger:consume` command logs

```bash
ddev exec symfony server:log
```

- detect environment variables automatically and create `.env.local` file on `ddev start`
- have the template how to run messenger consume worker on start
- allow to use `ddev console` to execute symfony console command

https://ddev--6725.org.readthedocs.build/en/6725/users/quickstart/#symfony

## Automated Testing Overview

tests added

## Release/Deployment Notes

-


rfay:
You know that the symfony CLI is already included in the web container, right? `ddev exec symfony version`

IndraGunawan:
> You know that the symfony CLI is already included in the web container, right? `ddev exec symfony version`

yes, i utilized that as well

rfay:
Is that the same as the symfony console (sorry for my ignorance)

stasadev:
> Is that the same as the symfony console (sorry for my ignorance)

It's not the same, but symfony CLI can run `bin/console`:

```
$ ddev exec symfony -h
...
Available wrappers:
Runs PHP (version depends on project's configuration).
Environment variables to use Platform.sh/Upsun relationships or Docker services are automatically defined.

  composer                                                       Runs Composer without memory limit
  console                                                        Runs the Symfony Console (bin/console) for current project
  php, pecl, pear, php-fpm, php-cgi, php-config, phpdbg, phpize  Runs the named binary using the configured PHP version
```

So running `ddev exec symfony console` will be the same as running `ddev exec bin/console` in the context of the project.

@IndraGunawan, please add `ddev symfony` command instead if `ddev console`, and use `ddev symfony console` in your examples.

IndraGunawan:
Symfony CLI is like a wrapper for a lot of things, some of them
- it can act as development web server (symfony server:start)
- run composer without memory limit (symfony composer)
- run a process (symfony run)
- run PHP using specific PHP version (symfony php)
- integration with platform.sh
- run symfony console (bin/console) with pre-populated environment variable if web server has start (symfony console)

while Symfony console (bin/console) is about command for the project like create db migration, create fixtures. and it will run with  default PHP version and no pre-populated environment variable

IndraGunawan:
@stasadev i don't see any benefits of adding `ddev symfony` if everything has been configured like PHP version and the webserver. the only missing is the console. if user really need symfony they can `ddev exec symfony ...`, is it?

stasadev:
So, the Symfony CLI is rarely used, and the Symfony Console is used every day, in the context of a project, right?

I'm just confused that they chose such a generic name for this tool.

Then, I'm fine with this (because typing `ddev console` is faster than `ddev symfony console`).

IndraGunawan:
> So, the Symfony CLI is rarely used, and the Symfony Console is used every day, in the context of a project, right?

Yes, creating controller, entity, run migration or reload fixtures could be more frequent and that is using `bin/console` or `symfony console` instead of creating project `symfony new` also ddev has already handle what `symfony serve` does on this case
Other command from symfony cli could be deployment to platform.sh which i never use it so have no comment on that

rfay:
Congrats on taking on this heavy lift!

Looks like you're making great progress; TestDdevFullSiteSetup not quite working, https://github.com/ddev/ddev/actions/runs/11840776970/job/32995964609

You can run the test manually locally, see https://ddev.readthedocs.io/en/stable/developers/building-contributing/#testing

There's also a number of contributor trainings that may help at https://ddev.com/blog/contributor-training/  (see https://ddev.com/blog/maintaining-ddev-tests-contributor-training/)


```
    ddevapp_test.go:2439: 
        	Error Trace:	/home/runner/work/ddev/ddev/pkg/ddevapp/ddevapp_test.go:2439
        	Error:      	"<br />\n<b>Warning</b>:  require_once(/var/www/html/vendor/autoload_runtime.php): Failed to open stream: No such file or directory in <b>/var/www/html/public/index.php</b> on line <b>5</b><br />\n<br />\n<b>Fatal error</b>:  Uncaught Error: Failed opening required '/var/www/html/vendor/autoload_runtime.php' (include_path='.:/usr/share/php') in /var/www/html/public/index.php:5\nStack trace:\n#0 {main}\n  thrown in <b>/var/www/html/public/index.php</b> on line <b>5</b><br />\n" does not contain "Welcome to the Symfony Demo application"
        	Test:       	TestDdevFullSiteSetup
        	Messages:   	expected Welcome to the Symfony Demo application on project TestPkgSymfony
```

IndraGunawan:
@rfay can't make test works on my local using WSL, but I've tested it using github actions

```
Volume TestPkgSymfony-mariadb for project TestPkgSymfony was deleted
Volume TestPkgSymfony-postgres for project TestPkgSymfony was deleted
Volume TestPkgSymfony_project_mutagen for project TestPkgSymfony was deleted
Project TestPkgSymfony was deleted. Your code and configuration are unchanged.
Copying directory /home/indra/.ddev/testcache/6b486da7_v2.6.0.tar.gz to /home/indra/tmp/ddevtest/TestPkgSymfony622454049

Copying complete
=== RUN   TestDdevFullSiteSetup
    ddevapp_test.go:2350: == BEGIN TestDdevFullSiteSetup for TestPkgSymfony (0)
Not trying to remove hostnames because DDEV_NONINTERACTIVE=true
Volume TestPkgSymfony-mariadb for project TestPkgSymfony was deleted
Volume TestPkgSymfony-postgres for project TestPkgSymfony was deleted
Volume TestPkgSymfony_project_mutagen for project TestPkgSymfony was deleted
Project TestPkgSymfony was deleted. Your code and configuration are unchanged.
Fix with 'ddev config global --required-docker-compose-version="" --use-docker-compose-from-path=false': Get "https://github.com/docker/compose/releases/download/v2.30.0/docker-compose-linux-x86_64": net/http: TLS handshake timeout
FAIL    github.com/ddev/ddev/pkg/ddevapp        22.436s
FAIL
make: *** [Makefile:127: testddevapp] Error 1
``` 

```
docker-compose --version
Docker Compose version v2.29.7-desktop.1
```

pkg/ddevapp/ddevapp_test.go
```diff
@@ -363,6 +363,21 @@ var (
...
+		{
+			Name:                          "TestPkgSymfony",
+			SourceURL:                     "https://github.com/symfony/demo/archive/refs/tags/v2.6.0.tar.gz",
+			ArchiveInternalExtractionPath: "demo-2.6.0/",
+			Type:                          nodeps.AppTypeSymfony,
+			Docroot:                       "public",
+			Safe200URIWithExpectation:     testcommon.URIWithExpect{URI: "/robots.txt", Expect: "User-agent"},
+			DynamicURI:                    testcommon.URIWithExpect{URI: "/", Expect: "Symfony Demo"},
+			FilesImageURI:                 "/apple-touch-icon.png",
+			PretestCmd:                    "ddev composer install --no-progress",
```
stasadev, line 376: We usually fork the upstream project, and prepare everything inside, including the commit for the Composer vendor dependencies (which makes testing faster).

Example with Laravel https://github.com/ddev/test-laravel

I suggest you to fork https://github.com/symfony/demo/tree/v2.6.0 and commit vendor dependencies, then you can transfer your fork to DDEV namespace later.

stasadev (commented):
> can't make test works on my local using WSL
> `Get "https://github.com/docker/compose/releases/download/v2.30.0/docker-compose-linux-x86_64": net/http: TLS handshake timeout`

Maybe `wsl --shutdown` or reboot can help https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#windows-wsl2-network-issues

pkg/ddevapp/ddevapp_test.go
stasadev: Or I can create a fork, if you like.

IndraGunawan: i have forked the demo repo here https://github.com/IndraGunawan/test-symfony and commit the vendor dir
I can't transfer to ddev org
`You don’t have the permission to create public repositories on ddev`

stasadev: Thanks, I invited you to the `ddevl` org, you can transfer your repo there, and then I will transfer it to the `ddev` org.

(This is how we usually do this.)

IndraGunawan: transfered

stasadev: Thanks, I did my part too. Let's update the URL in tests, and I'll restart them.

stasadev:
@IndraGunawan,

There were extra assets in `assets/vendor` and `var/sass` (that were gitignored and auto-created after `composer install`).

I committed them to https://github.com/ddev/test-symfony, updated `README.md` and added Symfony Demo db fixtures.

I hope you don't mind that I did it myself so as not to block this PR for too long.

stasadev:
`.env.local` is probably created too early, it should be there only on `ddev start`

```
        	Error:      	"\x1b[31mFailed to create project: '/home/runner/tmp/ddevtest/TestComposerCreateCmdsymfony953269503/.env.local' is not allowed to be present. composer create needs to be run on a clean/empty project with only the following paths: [] - please clean up the project before using 'ddev composer create'\x1b[0m\n" does not contain "touch1.txt"
```

IndraGunawan:
@stasadev since ddev is using nginx / apache. that `.env.local` will be read on every request hence I think it should be fine also it the file is not present on `ddev start` do you know another event that can fit for that?

docs/content/users/quickstart.md
```diff
@@ -737,27 +737,14 @@ ddev launch /admin
...
-=== "Composer"
-
-    ```bash
-    mkdir my-symfony-site && cd my-symfony-site
-    ddev config --docroot=public
-    ddev composer create symfony/skeleton
-    ddev composer require webapp
-    # When it asks if you want to include docker configuration, say "no" with "x"
-    ddev launch
-    ```
```
stasadev, line 751: Why is this removed?

docs/content/users/quickstart.md
```diff
@@ -767,12 +754,34 @@ If your project uses a database you'll want to set the [DB connection string](ht
...
     ddev composer install
     ddev launch
     ```
 
+!!!tip "Default web environments"
+    On `ddev start` command, some environments will be automatically added to `.env.local` file.
+
+    - `MAILER_*` to use [Mailpit](https://ddev.readthedocs.io/en/stable/users/usage/developer-tools/#email-capture-and-review-mailpit) as mail catcher.
+    - `DATABASE_*` based on database selection unless [--omit-containers=db](https://ddev.readthedocs.io/en/stable/users/configuration/config/#omit_containers) is set.
+    - `REDIS_*` if `ddev/ddev-redis` or `ddev/ddev-redis-7` add-on installed.
```
stasadev, line 768: I don't think we need this "Default web environments" tip.

The place for this kind of info is here (where the general logic is described)
https://ddev.readthedocs.io/en/stable/users/usage/cms-settings/

stasadev (changes requested):
> do you know another event that can fit for that?

Yes, I added it to `appTypeSettingsPaths`.

Looking forward to see if the tests are good now.

docs/content/users/quickstart.md
IndraGunawan: i thought that is just information those envs would be auto-populated and no user intervention was needed for that

docs/content/users/quickstart.md
IndraGunawan: even the exsiting guide using Symfony CLI also does a tricks by using a tmp dir  :stuck_out_tongue:

> I don't think it should be removed.
yeah, we can revert it

docs/content/users/quickstart.md
IndraGunawan: in the existing doc https://ddev.readthedocs.io/en/stable/users/quickstart/#symfony it state that the user need to modify the `.env` file with some string based on the configuration hence I do replace those statement with a tip that state user doesn't need to do anything

or we can copy from Laravel https://ddev.readthedocs.io/en/stable/users/quickstart/#laravel 
```
DDEV automatically updates or creates the .env file with the database information.
```

stasadev: I think we can have the same message as in Laravel quickstart (with `.env.local` instead of `.env`).

stasadev: Updated the docs with a short message instead.

rfay:
Yay, passing tests!

IndraGunawan:
yeayy, thank you @stasadev for your help

rfay:
> since ddev is using nginx / apache. that .env.local will be read on every request hence I think it should be fine also it the file is not present on ddev start do you know another event that can fit for that?

Just a note that nginx and apache are not involved in the application's `.env.local`, that's something that the PHP application can do (and I guess does with Symfony). The `.ddev/.env*` is read by `docker-compose` during `ddev start`, which is another different thing.

rfay:
@IndraGunawan I checked to make sure you have proper privs on https://github.com/ddev/test-symfony and I think you do. We'll be counting on you to pay attention to what it needs as time goes on (as well as this integration). Thanks so much for your work on this and your initiative with it!

pkg/ddevapp/symfony.go
```diff
@@ -0,0 +1,122 @@
...
+	if app.Database.Type == nodeps.Postgres {
+		dbPort = "5432"
+		dbDriver = "postgres"
+		dbVersion = app.Database.Version
+	} else if app.Database.Type == nodeps.MariaDB || app.Database.Type == nodeps.MySQL {
+		dbPort = "3306"
+		dbDriver = "mysql"
+		if app.Database.Type == nodeps.MariaDB {
+			// doctrine requires mariadb version until its patch version so add 0 as default patch version
+			dbVersion = fmt.Sprintf("%s.0-mariadb", app.Database.Version)
```
rfay, line 57: Can you add a link and a more detailed explanation for this please in the comment? This seems odd. I'm pretty sure Drupal uses symfony access to mariadb without this, but I could be wrong.  This seems very fragile and artificial.

pkg/ddevapp/global_dotddev_assets/commands/web/console
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+#ddev-generated
+## Description: Run symfony console inside the web container
+## Usage: console [flags] [args]
+## Example: "ddev console about" or "ddev console doctrine:schema:update --dump-sql"
+## ProjectTypes: symfony
+## ExecRaw: true
+
+if [ ! -f bin/console ]; then
```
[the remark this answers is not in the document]
IndraGunawan: > Is it always true that a symfony project installs into the project root, and bin/console will be right here?

i added it just to add a guard, if user initialize ddev with symfony project type but has yet initialize the symfony
I copied it from https://github.com/ddev/ddev/blob/master/pkg/ddevapp/global_dotddev_assets/commands/web/magento

> (And if it's somewhere else then symfony console would still work right?)

`symfony console` requires `bin/console` to be present

pkg/ddevapp/symfony.go
```diff
@@ -0,0 +1,122 @@
...
+	"path/filepath"
+	"slices"
+
+	"github.com/ddev/ddev/pkg/fileutil"
+	"github.com/ddev/ddev/pkg/nodeps"
+)
+
+// isSymfonyApp returns true if the app is of type symfony
+func isSymfonyApp(app *DdevApp) bool {
+	return fileutil.FileExists(filepath.Join(app.AppRoot, "bin", "console")) && fileutil.FileExists(filepath.Join(app.AppRoot, "src", "Kernel.php"))
```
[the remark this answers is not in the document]
IndraGunawan: if user install symfony using its [skeleton ](https://github.com/symfony/skeleton)(`composer create-project` or `symfony new`). the recipes will copy those those `bin/console` and 'src/Kernel.php` to the project

pkg/ddevapp/symfony.go
IndraGunawan: actually from Symfony docs should contain full database version https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration like from `mysql -V`

Note: I'm not using mariadb myself, if this is too fragile, I can remove it. i did that hacks because it was failing if I initialize symfony with default ddev config, so I debug the issue

only mariadb requires until its patch version https://github.com/doctrine/dbal/blob/4.2.x/src/Driver/AbstractMySQLDriver.php#L92-L108

while [postgres ](https://github.com/doctrine/dbal/blob/4.2.x/src/Driver/AbstractPostgreSQLDriver.php#L28-L33)only need the major version or no regex check on MySQL version https://github.com/doctrine/dbal/blob/4.2.x/src/Driver/AbstractMySQLDriver.php#L62-L76

pkg/ddevapp/symfony.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	return nil
+}
+
+// getSymfonyHooks for appending as byte array.
+func getSymfonyHooks() []byte {
+	symfonyHooks := `## Un-comment to consume async message.
+#    post-start:
+#    - exec: symfony run --daemon --watch=config,src,templates,vendor bin/console messenger:consume async -vv
```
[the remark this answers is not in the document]
IndraGunawan: i'm new on DDEV dev.

in here https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#running-extra-daemons-in-the-web-container I saw 
```
web_extra_daemons is a shortcut for adding a configuration to supervisord, which organizes daemons inside the web container. If the default settings are inadequate for your use, you can write a [complete config file for your daemon](https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#explicit-supervisord-configuration-for-additional-daemons).
```

if that's the case, `symfony run --daemon` runs the command as daemon and manage its lifecycle, I don't think we need to wrap it with supervisor, wdyt?

pkg/ddevapp/symfony.go
```diff
@@ -0,0 +1,122 @@
...
+				envMap["REDIS_USER"] = "redis"
+				envMap["REDIS_PASSWORD"] = "redis"
+				envMap["REDIS_URL"] = "redis://redis:redis@redis:6379"
+			} else {
+				envMap["REDIS_URL"] = "redis://redis:6379"
+			}
+		}
+	}
+
+	err := WriteProjectEnvFile(envFilePath, envMap, envText)
```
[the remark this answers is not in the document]
IndraGunawan: `WriteProjectEnvFile` is existing function and it is respect the existing key inside the env file

@stasadev can confirm this?

pkg/ddevapp/symfony.go
```diff
@@ -0,0 +1,122 @@
...
+func symfonyEnvMailer(app *DdevApp, envMap map[string]string) {
+	envMap["MAILER_AUTH_MODE"] = ""
+	envMap["MAILER_PASSWORD"] = ""
+	envMap["MAILER_USERNAME"] = ""
+	envMap["MAILER_CATCHER"] = "1"
+	envMap["MAILER_DRIVER"] = "smtp"
+	envMap["MAILER_DSN"] = "smtp://localhost:1025"
+	envMap["MAILER_HOST"] = "localhost"
+	envMap["MAILER_PORT"] = "1025"
+	envMap["MAILER_URL"] = "smtp://localhost:1025"
```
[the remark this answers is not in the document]
IndraGunawan: fixed

pkg/ddevapp/global_dotddev_assets/commands/web/console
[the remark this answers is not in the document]
rfay: > symfony console requires bin/console to be present

Thanks. I would have expected `symfony console` to find the console script wherever it was.

pkg/ddevapp/symfony.go
rfay: If you could take a look at it and verify that it's really required it would be great. I'm betting it's not required. It seems very awkward and error-prone to artificially add a ".0" there.

pkg/ddevapp/symfony.go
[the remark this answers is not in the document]
rfay: It looks to me like WriteProjectEnvFIle should respect existing values.

pkg/ddevapp/symfony.go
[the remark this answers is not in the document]
rfay: It's exactly the same. Provide it commented out. Try it out and see if you like it.

pkg/ddevapp/symfony.go
IndraGunawan: the ".0" is not required if we can get the full mariadb version string

or what if we exclude mariadb? if any mariadb user need this in the future, they can contribute on this

rfay: You can get the full mariadb version string but only after starting it up.

MariaDB is probably the most popular database for the majority of CMSs. This is certainly an interesting problem. 

One thing we *could* do that we've never had to deal with before is to detect the mariadb version during startup. It's easy in a post-start hook, 

`ddev exec 'mysql -e "SELECT VERSION();"'`

pkg/ddevapp/symfony.go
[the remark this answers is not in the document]
IndraGunawan: Could you share what func should i hook to?

pkg/ddevapp/symfony.go
IndraGunawan: Quick search on drupal repository and found this https://github.com/drupal/drupal/blob/11.x/core/modules/mysql/src/Driver/Database/mysql/Connection.php#L277
Drupal do check the db version

One thing i can do here is to exclude mariadb from auto-populated env var

rfay: So Drupal does it dynamically, but Symfony seems to rely on it being preconfigured. 

I don't think this is going to affect mariadb only. Doesn't it affect mysql? 

I just don't understand how Symfony wouldn't be able to handle a db server upgrade, for example. A new version of DDEV often has a new patch version of mariadb, mysql, and postgres.

pkg/ddevapp/symfony.go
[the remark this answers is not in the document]
rfay: Same as in docs. 

Please try it out and see how it works for you. Don't bother with trying it here, try it first locally.

https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#running-extra-daemons-in-the-web-container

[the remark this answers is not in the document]
IndraGunawan: I don’t think we are on the same page. I’m talking about how a project type modify the config.yaml and put the commented config

if i’m as user, it won’t be an issue to follow the docs

if you can show me examples how a project type automatically add web_extra_daemon would be good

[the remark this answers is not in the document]
IndraGunawan: Let it as is on post-start hook then

pkg/ddevapp/symfony.go
IndraGunawan: Commited to exclude mariadb

rfay (commented):
Minor comment here on the way, using PHPDefault.

I really don't want to exclude mariadb, it is more likely the default, so we need to figure out a way to handle that.

@fabpot has some comments, hoping he'll review here.

stasadev:
Rebased to resolve conflicts.

stasadev (approved):
I tested Composer and Symfony CLI quickstarts https://ddev--6725.org.readthedocs.build/en/6725/users/quickstart/#symfony
Everything is fine.

I wanted to test it with https://github.com/symfony/demo, but it failed for database migrations because that demo is not adapted for the recent change (a few days ago) in `doctrine/data-fixtures` https://stackoverflow.com/a/79234035/8097891

So I tested it with our fork instead to view blog posts generated from database fixtures.

Ignore this warning (it's because 8.2.0 is hardcoded in `composer.json`):

```
WARNING the current dir requires PHP 8.2.0 (composer.json from current dir: /var/www/html/composer.json), but this version is not available: fallback to 8.2
```

And ignore this warning https://stackoverflow.com/a/49817578:

```
[CAUTION] This operation should not be executed in a production environment!
```

For each database type:

```
# mariadb:10.11
git clone https://github.com/ddev/test-symfony && cd test-symfony
ddev config --auto
ddev start
ddev console doctrine:schema:create
ddev console doctrine:fixtures:load --no-interaction
# view loaded records from the database
ddev launch en/blog
# check table records in the database
ddev dbeaver

# mysql:8.0
ddev delete -Oy
ddev config --database=mysql:8.0
ddev restart
ddev console doctrine:schema:create
ddev console doctrine:fixtures:load --no-interaction
# view loaded records from the database
ddev launch en/blog
# check table records in the database
ddev dbeaver

# postgres:17
ddev delete -Oy
ddev config --database=postgres:17
ddev restart
ddev console doctrine:schema:create
ddev console doctrine:fixtures:load --no-interaction
# view loaded records from the database
ddev launch en/blog
# check table records in the database
ddev dbeaver
```

All database types worked for me.

docs/content/users/quickstart.md
```diff
@@ -679,13 +679,13 @@ ddev launch /admin
 
 There are many ways to install Symfony, here are a few of them based on the [Symfony docs](https://symfony.com/doc/current/setup.html).
```
rpkoller, line 680:
```suggestion
```
Is that sentence really necessary? It is the purpose of quickstarts to provide one or two ways how to best get a certain CMS spun up. there is not really a need to state that there are several ways to install symfony imho

rpkoller (commented):
tested the quickstarts for composer and symfony cli and in both cases i get to a functional landing page. (symfony doesnt have an "admin ui" the user could be point to?)

docs/content/users/quickstart.md
rfay: This is probably left over from previous quickstart. OTOH... there *are* many ways, and it's more of a framework. So this may well be valid.

IndraGunawan:
thank you @rfay , @stasadev for helping me to fix some issues here 😁

rfay:
Thanks for taking the initiative to do this!

PR #6727, opened 2024-11-15
docs: for phpstorm plugin, prerequisite is not necessary

ptmkenny:
Make it clear that the "prerequisite" is not necessary if using the PHPStorm plugin

## The Issue

- #<issue number>

Even though I was using the PHPStorm plugin, because of the structure of the docs page, I attempted to configure PHPStorm as described in the "Prerequisite" section even though that is not necessary.

Restructures the doc to make it clear that PHPStorm does not need to be configured if using the plugin.

### DIFF
--- a/docs/content/users/install/phpstorm.md
+++ b/docs/content/users/install/phpstorm.md
@@ -8,14 +8,6 @@ If you work with the [PhpStorm](https://www.jetbrains.com/phpstorm/) IDE, you ca
 - DDEV v1.21.1 or higher.
 - Make sure to get at least one project going with `ddev start` before trying to set up the plugin, because the plugin assumes it has a project to work with.
 
-## Prerequisite
-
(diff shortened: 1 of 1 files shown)


rfay:
TODO:
- [ ] Verify that the DDEV Integration Plugin does this for you, and that we can delay mentioning docker-compose location until after the discussion of the plugin.
- [ ] Use the word "absolute path" instead of "full path"

rfay:
I just tested and this is not correct. The plugin does not properly update `docker-compose` location in the PhpStorm config. Sorry I misled you.

rfay:
I tried to adapt your wording about docker-compose in https://github.com/ddev/ddev/pull/6740/commits/54b752ca6cc5b49e89046a6c7d7d9d4d44345a3e - Please see if you like it, thanks.

ptmkenny:
Thanks, that looks good!

PR #6728, opened 2024-11-15
build: Update ddev-ssh-agent build

rfay:
## The Issue

* Update ddev-ssh-agent to base on Debian 12 bookworm
* Put the complex add-keys stuff into a script and run the script
* Tiny updates to error handling and script env var usage

This is really prep for looking at 
* https://github.com/ddev/ddev/pull/6720

I found that the inline pipe there was too hard to understand or debug, and some of the other things needed some attention.

I experimented here with using PUBLIC_SSH_AUTH_SOCK for example, but didn't get it to work right. It appears that the ability to ignore things with that has made it into buildx, but 1) I didn't make it work at all and 2) the warnings didn't go away anyway. So I backed that out.

Normal usage of `ddev auth ssh`

## Automated Testing Overview

Hopefully everything is the same.


stasadev (approved):
Thank you for exploring this.

Tested, it works as expected.

PR #6731, opened 2024-11-17
fix: Improve interactive ddev config text, fixes #6542

### LINKED ISSUE #6542: `ddev config` should have better text about accepting default instead of "You may leave this value blank if your site files are in the project root"

>>You may leave this value blank if your site files are in the project root
Docroot Location (web):

having (web) in paranthesis suggests that it is the default value when no value is input

### Describe your solution

add web as default

### Describe alternatives

or change the label

### Additional context

_No response_

rfay:
## The Issue

- #6542 

Misleading text in interactive `ddev config` of new project

Improve the text.

* Use `ddev config` in an empty directory (uses default `php` type)
* Use `ddev config` on an already-configured project that has docroot = ""

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1412,8 +1412,8 @@ func DiscoverDefaultDocroot(app *DdevApp) string {
 func (app *DdevApp) docrootPrompt() error {
 
 	// Determine the document root.
-	util.Warning("\nThe docroot is the directory from which your site is served.\nThis is a relative path from your project root at %s", app.AppRoot)
-	output.UserOut.Println("You may leave this value blank if your site files are in the project root")
+	output.UserOut.Printf("\nThe docroot is the directory from which your site is served.\nThis is a relative path from your project root at %s\n", app.AppRoot)
+	output.UserOut.Printf("Leave docroot empty (hit <RETURN>) if your index.php is in the project root.\n")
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/config.go
```diff
@@ -1412,8 +1412,8 @@ func DiscoverDefaultDocroot(app *DdevApp) string {
 func (app *DdevApp) docrootPrompt() error {
 
 	// Determine the document root.
-	util.Warning("\nThe docroot is the directory from which your site is served.\nThis is a relative path from your project root at %s", app.AppRoot)
-	output.UserOut.Println("You may leave this value blank if your site files are in the project root")
+	output.UserOut.Printf("\nThe docroot is the directory from which your site is served.\nThis is a relative path from your project root at %s\n", app.AppRoot)
+	output.UserOut.Printf("Leave docroot empty (hit <RETURN>) if your index.php is in the project root.\n")
```
stasadev, line 1416: Maybe this is better?:

```suggestion
	output.UserOut.Printf("Leave docroot empty (hit <RETURN>) to use the location shown in parentheses.\nOr specify a custom path if your index.php is in a different directory.\n")
```

We need to emphasize that the default value is in parentheses.

stasadev (approved):
Looks good to me.

PR #6733, opened 2024-11-17
fix: remove python and django basic features and tests, fixes #6714

### LINKED ISSUE #6714: Remove python/django support in v1.24.0
Our python/django experiment has not been a success. I propose that we remove this feature in upcoming v1.24.0

## TODO

- [ ] Remove tests (python/django/etc)
- [ ] Remove extra packages added to ddev-webserver, especially gunicorn and related
- [ ] Remove related docs
- [ ] Remove python and django project types

rfay:
## The Issue

- #6714 

Python support hasn't been successful. Remove python/django.

Removal

ddev-webserver is about 25MB smaller, not a lot, but something. https://hub.docker.com/r/ddev/ddev-webserver/tags

Config of python/django should fail.
ddev-webserver should be smaller.

Here's what I looked for in the code:

* python
* django
* gunicorn


containers/ddev-webserver/Dockerfile
```diff
@@ -104,16 +104,11 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -o Dpkg::Options::="--for
     libldap-common \
     libpcre3 \
     libpq-dev \
-    libpython3-dev \
     nano \
     ncurses-bin \
     netcat-traditional \
     openssh-client \
     patch \
-    python-is-python3 \
```
rfay, line 113: I wonder if this is a mistake. Maybe this should stay in there.

stasadev: This package adds a symlink, that should be there:
```
$ python -V
bash: python: command not found
$ python3 -V
Python 3.11.2
```

stasadev (changes requested):
More places that should be reviewed for python/django usage:

https://github.com/ddev/ddev/blob/2d90a7e236536b6fa9ae4fe06d61d1ed321853e7/.circleci/README.arch-aur-builder.md#L49

https://github.com/ddev/ddev/tree/master/pkg/ddevapp/testdata/TestApptypeDetection/django4

stasadev (commented):
And one more leftover https://github.com/rfay/ddev/blob/20241117_remove_python_django/pkg/ddevapp/testdata/TestApptypeDetection/django4/manage.py

rfay:
Thanks, I removed those two AUR-related unused things with only slight reservations, because when I was working on the homebrew/process PR I did the upgrade manually using our README that is in https://ddev.readthedocs.io/en/stable/developers/release-management/#manually-updating-aur-repository and was successful. Had minor stumbles, but was able to do it.

PR #6738, opened 2024-11-18
fix: Move sqlite3 3.45 installation into ddev-webserver, fixes #6735

### LINKED ISSUE #6735: install sqlite3 3.45+ in upstream Dockerfile

We are doing special config for Drupal 11 with sqlite3.45 (at start time)

Do it instead in the base config upstream Dockerfile

rfay:
## The Issue

- #6735 

Sqlite 3.45 installation  per-project in Drupal 11 is fragile. We can control it better in the ddev-webserver Dockerfile, and it's unlikely to bother anybody else.


stasadev (approved):
It looks good to me, and I want to point out that 3.45.1 is the last version we can use in Debian Bookworm.
The other day I tried to install 3.46.1 and it didn't work because of some dependency.

PR #6739, opened 2024-11-18
fix: remove nginx proxy router, fixes #5940

### LINKED ISSUE #5940: Remove nginx proxy router

DDEV uses traefik by default, which works great for most if not all scenarios.

There is also a legacy `ddev config global --router=nginx-proxy` option. @rfay mentioned in the advisory group discussion that it would be good to remove it in the interest of reducing complexity.

### Describe your solution

Remove the nginx proxy router.

### Describe alternatives

_No response_

### Additional context

I have never used the nginx proxy router - I'm not sure if there are use cases for it that the traefic router doesn't handle, can't handle, or can handle but only with difficulty. It would be good to get some context around that before any work is done to remove it.

rfay:
## The Issue

- #5940

Remove the deprecated nginx proxy router

Try setting router to nginx-proxy router
Make sure ordinary things work


cmd/ddev/cmd/config-global.go
```diff
@@ -318,7 +298,6 @@ func init() {
 	configGlobalCommand.Flags().String("xdebug-ide-location", "", "For less usual IDE locations specify where the IDE is running for Xdebug to reach it")
 	configGlobalCommand.Flags().Bool("use-traefik", true, "If true, use Traefik for ddev-router")
 	_ = configGlobalCommand.Flags().MarkDeprecated("use-traefik", "please use --router instead")
-	configGlobalCommand.Flags().String("router", globalconfigTypes.RouterTypeTraefik, fmt.Sprintf("Valid router types are %s, default is %s", strings.Join(globalconfigTypes.GetValidRouterTypes(), ", "), globalconfigTypes.RouterTypeDefault))
```
stasadev, line 321: I think we'd better mark the `--router` flag as deprecated, just like we did for `--use-traefik` above.
Or make it accept `traefik` only.

docs/content/users/configuration/config.md
```diff
@@ -498,15 +488,7 @@ If set to `v2.8.0`, for example, it will download and use that version instead o
...
-## `router`
-
-Whether to enable the default [Traefik router](../extend/traefik-router.md) or the deprecated "nginx-proxy" router.
-
-| Type | Default | Usage
-| -- | -- | --
-| :octicons-globe-16: global | `traefik` | Can `traefik` or `nginx-proxy` (deprecated).
-
-May also be set via `ddev config global --router=traefik` or `ddev config global --router=nginx-proxy`.
+The only current value is `traefik`.
```
stasadev, line 491:
```suggestion
## `router`

The only current value is `traefik`.
```

stasadev (changes requested):
We shouldn't have `containers/ddev-nginx-proxy-router`, right?

rfay:
> We shouldn't have containers/ddev-nginx-proxy-router, right?

I was reluctant to just remove it right away. Do you disagree? It would be reasonable either way. It's obviously there in the source code.

stasadev:
> I was reluctant to just remove it right away. Do you disagree? It would be reasonable either way. It's obviously there in the source code.

I'm okay with it, just asked if this was intentional to leave it there.

stasadev (changes requested):
When I use `nginx-proxy` with v1.23.5, and then switch to this PR:

```
$ ddev version
FATA[0000] unable to read global config: invalid router: nginx-proxy, valid router types are [traefik]
```

I guess we should change `nginx-proxy` to `traefik` automatically?

pkg/globalconfig/global_config.go
```diff
@@ -209,7 +209,9 @@ func ValidateGlobalConfig() error {
 	}
 
 	if !types.IsValidRouterType(DdevGlobalConfig.Router) {
-		return fmt.Errorf("invalid router: %s, valid router types are %v", DdevGlobalConfig.Router, types.GetValidRouterTypes())
+		output.UserOut.Printf("\nThe only valid router type is %s, but you have router: %s in your global configuration, using %s instead.\n", types.RouterTypeTraefik, DdevGlobalConfig.Router, types.RouterTypeTraefik)
+		DdevGlobalConfig.Router = types.RouterTypeTraefik
+		return nil
```
stasadev, line 214: Return was added here to show a warning for the user each time, but when I tested it with:

```yaml
router: nginx-proxy
xdebug_ide_location: "blablabla"
```

DDEV didn't complain about wrong `xdebug_ide_location`.

I don't think it's worth skipping all the further validation checks here.

stasadev (changes requested):

pkg/globalconfig/global_config.go
rfay: Fixed, thanks for catching the erroneous return there.

stasadev (approved):
Thank you.

I tested upgrade/downgrade. DDEV did a great job in all cases.

Regarding the code in containers/ddev-nginx-proxy-router:
We need to keep it because we still use this image for versions 1.23.5 and earlier, and we are required to provide the source code (as requested by Docker-Sponsored Open Source Program). So, until the Docker image is available, we should keep it.

PR #6741, opened 2024-11-18
fix: get composer/git to stop complaining about safe dir

rfay:
## The Issue

Recent reports of problems like 

```
fatal: detected dubious ownership in repository at '/var/www/html'
To add an exception for this directory, call:

        git config --global --add safe.directory /var/www/html
```

For example,
* https://discord.com/channels/664580571770388500/1194832810129772554

Add safe directory config to the .gitconfig that we already provide.

In lima or Docker Desktop without mutagen on macOS, do a plain DDEV drupal 11 installation and ddev composer install.

You shouldn't see the warning.


rfay:
New image pushed, thanks!

stasadev (approved):
Thank you, tested, everything is fine.

PR #6745, opened 2024-11-19
build: add timeout to `docker-compose build`

stasadev:
## The Issue

https://github.com/ddev/ddev/actions/runs/11915882563/job/33207216994

```
=== RUN   TestPHPConfig
...
Building project images...
2024-11-19T15:17:48.612 Executing docker-compose -f /home/runner/tmp/ddevtest/TestPkgDrupal101551583127/.ddev/.ddev-docker-compose-full.yaml build --progress=plain
......................................panic: test timed out after 4h0m0s
	running tests:
		TestPHPConfig (3h55m4s)
...
```

We started to see tests randomly fail when running a `docker-compose build`, but we cannot debug it because we don't see where it is stuck.
Let's add a relatively large timeout of 1 hour to at least see where it doesn't work.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1496,6 +1496,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
 		Action:       []string{"--progress=" + progress, "build"},
 		Progress:     true,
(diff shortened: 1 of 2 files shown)


rfay:
This seems like a good idea to me. At least we can watch it to see what happens and if we don't like it we can back out. 

We can also dial back the test timeout from 4h, but here it's more important to find out what's happening, because if it happens to us in tests, it will eventually happen to users and they won't know how or what to report.

PR #6747, opened 2024-11-20
build: add more timeouts for `apt-get update`, remove custom `php8.0-xdebug`, for #6745

stasadev:
## The Issue

- #6745

https://github.com/ddev/ddev/actions/runs/11927136921/job/33242061289?pr=6739

Stuck here:
```
RUN (apt-get -qq update || true) && DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests dnsutils
```

I think the reason why it's stuck is that some repo, e.g. deb.sury downloads with a very low speed.

- Add more timeouts for `apt-get update` that are done on the user side:
	- I used 30 seconds (1/4 of the `--default-container-timeout`) for single repo `apt-get update`
	- And 120 seconds for full `apt-get update`.
- Removes `-qq` that hid the output
- Removes custom `php8.0-xdebug` build because multi-stage build can fail if you have slow connection.

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -87,7 +87,7 @@ function cleanup {
 
 ddev config --project-type=php --docroot=web --disable-upload-dirs-warning || (printf "\n\nPlease run 'ddev debug test' in the root of the existing project where you're having trouble.\n\n" && exit 4)
 
-printf "RUN apt-get update\nRUN curl -I https://www.google.com\n" > .ddev/web-build/Dockerfile.test
+printf "RUN timeout 120 apt-get update || true\nRUN curl -I https://www.google.com\n" > .ddev/web-build/Dockerfile.test
 
 set +eu
 
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -78,6 +78,7 @@ jobs:
         run: |
           source ~/.bashrc
           docker version
+          docker info
           mkcert --version
           set -eu -o pipefail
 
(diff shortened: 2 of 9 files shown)


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -87,7 +87,7 @@ function cleanup {
 
 ddev config --project-type=php --docroot=web --disable-upload-dirs-warning || (printf "\n\nPlease run 'ddev debug test' in the root of the existing project where you're having trouble.\n\n" && exit 4)
 
-printf "RUN apt-get update\nRUN curl -I https://www.google.com\n" > .ddev/web-build/Dockerfile.test
+printf "RUN timeout 120 apt-get update || true\nRUN curl -I https://www.google.com\n" > .ddev/web-build/Dockerfile.test
```
rfay, line 90: This does no harm, but wow, we shouldn't have to do this :)

rfay (approved):
Thank you for your work on this.

cmd/ddev/cmd/scripts/test_ddev.sh
stasadev: There is a possibility of getting stuck in `TestCmdDebugTest` because of this `apt-get update`.
So I added a timeout here.

https://github.com/ddev/ddev/actions/runs/11927128822/job/33242041249#step:16:18865

stasadev:
I tested this with:
```
ddev config --database=postgres:9 --webimage-extra-packages=neofetch --dbimage-extra-packages=postgresql-9.6-postgis-2.3,postgresql-9.6-postgis-scripts,postgis --default-container-timeout=150
```
And looked inside `.ddev/.dbimageBuild/Dockerfile` and `.ddev/.webimageBuild/Dockerfile` to see if timeout is fine.

While testing I had slow connection (temporary problem that was automatically resolved in 30 minutes), and php8.0-xdebug failed:

```
', stderr='failed to solve: failed to compute cache key: failed to calculate checksum of ref GG2L:B3IQ:5Y6L:42AW:GAF4:5CCH:E5P6:CWVU:XY6C:Q6HU:YOD3:4UBV::i8tt87k3z8p2o13xc2r1uhxo7: "/usr/lib/php/20200930/xdebug.so": not found'
```

It turned out that multi-stage builds run in parallel and it was not enough time, decided to remove it.

rfay:
When you decide to pull I'll update and re-test 
* https://github.com/ddev/ddev/pull/6747

PR #6750, opened 2024-11-21
fix: Move drupal11 to its own project type, make "drupal" an alias for latest drupal, fixes #6635

### LINKED ISSUE #6635: Change `drupal` to be alias for latest drupal version. Introduce `drupal11` type

In a recent DDEV release we introduced the `drupal` project type for Drupal 8+. 

It relies on inspection of the code laid out to determine the version, which means it's not easy to specify the Drupal version at `ddev config --project-type=drupal` time, except with `ddev config --project-type=drupal --php-version=8.3`. (And PHP version may not be the only problem in the end. For example, the inspection done in `ddev config --update` is used to set `corepack_enable` as well, for Drupal 11, and also determines the use of an appropriate version of Sqlite3.)

There are various complaints about this though, mostly because in the current setup, `ddev config --project-type=drupal` may result in a PHP 8.2 project (which should then be changed to 8.3 with `ddev config --update`). But people aren't doing the `ddev config --update` because it's not in their habits. (The next major release of DDEV will change most defaults to PHP 8.3, eliminating the biggest problems, but there are still complaints about how this is set or queried.)

Discussion is welcome about how we could:
* Specify the Drupal sub-version at `ddev config` time while maintaining the generic `drupal` project type (or should we abandon `drupal` and just go back to specific versions like `drupal10`, `drupal11`, `drupal12` ?
* Inspect a project to find out its sub-version (`ddev describe` improvement?)
* Other solutions?

We can probably do something about this for DDEV v1.24.0

Related:
* https://github.com/ddev/ddev/issues/6306#issuecomment-2429547889 and earlier
* https://github.com/ddev/ddev/issues/6633

rfay:
## The Issue

- #6635 

In progress.

## TODO:
- [x] Implement drupal type as alias for drupal11
- [x] Make sure `ddev config --project-type=drupal` works.
- [x] Update quickstarts to stop using `ddev config --update`
- [x] Change docs on `ddev config --update` - should it go away? Just be `ddev config --auto` ?
- [x] Improve docs about the `drupal` project type and what it means now (alias) and it's previous usage.
- [x] Several changes were made to use `drupal11` in tests, they may be able to use `drupal` and be more resilient for future versions.
- [x] `ddev list` and `ddev describe` should show the configured project type, not the detected project type
- [x] `ddev config --project-type=drupal` should create `drupal11` right now
- [x] Tests for complex "drupal" handling
- [x] With drupal11 codebase, restart ddev config manual. Gets php? then drupal10? may be detection is not working right. May be intermittent.

- [ ] Test with `ddev config --project-type=drupal --docroot=web` (should come out drupal, but `ddev list` and `ddev describe` will report it as drupal11)
- [ ] Test with interactive `ddev config`
- [ ] Test with various project types in config.yaml


pkg/ddevapp/providerAcquia_test.go
```diff
@@ -59,8 +59,8 @@ func TestAcquiaPull(t *testing.T) {
 	err := globalconfig.WriteGlobalConfig(globalconfig.DdevGlobalConfig)
 	assert.NoError(err)
 
-	// Use a Drupal 10 codebase (test CMS 12)
-	drupalCode := FullTestSites[12]
+	// Use a Drupal 11 codebase (test CMS 12)
```
stasadev, line 62:
```suggestion
	// Use a Drupal 11 codebase (test CMS 16)
```

pkg/ddevapp/templates.go
```diff
@@ -7,10 +7,9 @@ const ConfigInstructions = `
 # name: <projectname> # Name of the project, automatically provides
 #   http://projectname.ddev.site and https://projectname.ddev.site
 
-# type: <projecttype>  # backdrop, cakephp, craftcms, drupal, drupal6, drupal7, laravel, magento, magento2, php, shopware6, silverstripe, typo3, wordpress
+# type: <projecttype>  # backdrop, cakephp, craftcms, drupal6, drupal7, drupal8, drupal9, drupal10, drupal11, laravel, magento, magento2, php, shopware6, silverstripe, typo3, wordpress
```
stasadev, line 10: I added `drupal` back, because it was in the docs for the project type.
Perhaps a note about an alias (just like in the docs) is also needed here.

```suggestion
# type: <projecttype>  # backdrop, cakephp, craftcms, drupal, drupal6, drupal7, drupal8, drupal9, drupal10, drupal11, laravel, magento, magento2, php, shopware6, silverstripe, typo3, wordpress
```

stasadev (changes requested):
Please also update `schema.json` with `drupal11`.

cmd/ddev/cmd/composer-create_test.go
```diff
@@ -55,7 +55,7 @@ func TestComposerCreateCmd(t *testing.T) {
 				// Normally for Drupal the docroot would be web, and the composer root would be the
 				// project root (default). But here we're making sure we can use the docroot
 				// as the composer_root. Acquia sites often do this...
-				if projectType == nodeps.AppTypeDrupal {
+				if projectType == nodeps.AppTypeDrupal11 || projectType == nodeps.AppTypeDrupal10 || projectType == nodeps.AppTypeDrupal9 || projectType == nodeps.AppTypeDrupal8 {
```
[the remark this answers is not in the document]
rfay: Thanks, did that. It's a little bit of silliness in the test though.

pkg/ddevapp/templates.go
rfay: Thanks. I always wonder if this whole comment section is useful. Wonder if anybody reads it.

rpkoller: i am using and reading it every time i dont spin up a project with a copy and paste snippet or an ansible script. i prefer using `ddev config` then. 

@stasadev and in regard of re-adding `drupal`. that is one point i've asked randy to remove before on the PR. for me the main problem in that list is that it is long and drupal alone makes up  roughly one third of the entries with eight different project types. and drupal and drupal11 are redundant since drupal is an alias for drupal11 as the current major version. my thought was enable users that know and use the alias "drupal" simply use it but for the visible options go with the absolute explicit project types of drupal6-drupal11 only. that was my thought behind suggesting removing drupal in that list. but as i told already to randy probably the better choice for the flow used on `ddev config` instead of going with such a lengthy inline list might be for example adding a fourth step. where the user is able for CMSs with more than one project type to select the specific version there. ( but that is out of the scope for this PR)

stasadev (changes requested):
When I run `ddev config --project-type` with no `.ddev` folder:

```
ddev delete -Oy && rm -rf .ddev && ddev config --project-type=drupal11 && cat .ddev/config.yaml | grep '^corepack'
corepack_enable: true
```

But:
```
ddev delete -Oy && rm -rf .ddev && ddev config --project-type=drupal && cat .ddev/config.yaml | grep '^corepack'
corepack_enable: false
```

pkg/ddevapp/templates.go
stasadev: Yeah, I also don't like that we have too many drupal project types.

Probably a fourth step in the interactive `ddev config` would be better, for example:

```
Project Type [backdrop, cakephp, craftcms, drupal, laravel, magento, magento2, php, shopware6, silverstripe, typo3, wordpress] (php): drupal

Note: The project type 'drupal' is aliased 'drupal11' in the current DDEV.
Project Version [drupal, drupal6, drupal7, drupal8, drupal9, drupal10, drupal11] (drupal):
```

stasadev (changes requested):
One more thing, when I have `drupal`, and then run `ddev config --auto` or `ddev config --update`, it's replaced with `drupal11`.
DDEV does the right thing, but on the other hand, it should remain `drupal`.

rfay:
> One more thing, when I have drupal, and then run ddev config --auto or ddev config --update, it's replaced with drupal11.

This is intentional. 

* `ddev config --update` definitely wants to discover the "preferred" type, so it always alters project type when it discovers another project type
* `ddev config --auto` perhaps ought to respect the existing one, but it's a tossup.

stasadev (approved):
Okay, then it's fine, looks good to me.

Issue #6748, opened 2024-11-20
Why are `GOARCH` and `GOOS` set in the container?


hanoii:
I set myself to add go to my ddev projects, the reason behind this is multi-fold, but I wanted to

```
go install github.com/skx/sysbox@latest
```

I have other ways of installing it but I used this as a way to learn a bit on go on its own, and I noticed that go install github.com/skx/sysbox was building it as `darwin_arm64` and that led to to try to understand why and I learned that GOOS and GOARCH are exported from the host.

I have many ways around this, even unsetting this, but I wonder if there's an actual reason for this or if maybe these variables should not be set?

https://github.com/ddev/ddev/blob/dd09a1a3bc2aeb4b761e74926ea6937776bf631f/pkg/ddevapp/ddevapp.go#L2440-L2441

rfay:
They're the easiest and most consistent way to get the OS and architecture, that's all.

hanoii:
but that's the HOST OS and architecture... is that used elsewhere? You see how that would affect if anyone want's to develop with go inside the container, those crazy like me .. just wonder if these variables should have been something else, like DDEV_HOST_OS and DDEV_HOST_ARCH

rfay:
I just spent some time poking around.

This was added in
* https://github.com/ddev/ddev/pull/3898

I guess it's to provide that information to add-ons, which may not have that exact thing available. I see it being used (I think) in TestCmdAddonComplex but that's probably it. 

I would think it could be changed to something else indicative. Not sure how valuable it is.

PR #6752, opened 2024-11-21
refactor: replace GOOS and GOARCH for a DDEV_ alternative

hanoii:
## The Issue

- #6748 

By adding go to the web container in order to compile some go utilities I found that becase GOARCH and GOOS were present and they were set to the ddev host runtime, the go compiled binary **inside the web container** turned out to be of the host platform.

Although these variables seem to be hardly ever used, this PR keeps them but under the DDEV_ environment variable namespace.


hanoii:
Added NOTE to the issue summary.

rfay (approved):
Seems reasonable to me. It's not clear that this ever got used.

rfay:
I'd love to talk more about using go in add-ons. I find the shell + go templates to be a terrible environment for sophisticated add-ons. I would at least like to be able to parse, process, and write yaml and json like go can. I've looked at adding a go plugin to DDEV (it looks like it would either be fragile or require extra features like go compiler or required libraries) and have looked at shipping a go binary add-on in both architectures.

hanoii:
> I've looked at adding a go plugin to DDEV (it looks like it would either be fragile or require extra features like go compiler or required libraries)

isn't go a compiler on its own? I thought that was I was doing on my plugin, just installing go and doing go install would build stuff? 

> have looked at shipping a go binary add-on in both architectures.

Does this mean downloading a go binary like you do with mutagen and compose on the host? This might be a good alternative, but I am lacking go knowledge to know if there's anything else needed. This doesn't add go to the add on, only just making sure it can run go code I guess, but it's a good approach as any honestly? Like running node or python that are also on ddev, but at least this would be run from the host. You can also do `docker run` calls on the add-on with whatever.

And we touched this topic before, what I would love, and not sure it's possible, is a way to add "logic" to ddev being able to touch its internals, kind of how `.dll` works (or used to work) on windows - I haven't developed for windows in ages but I had early low experience with c/c++ on it.

rfay:
> And we touched this topic before, what I would love, and not sure it's possible, is a way to add "logic" to ddev being able to touch its internals, kind of how .dll works (or used to work) on windows - I haven't developed for windows in ages but I had early low experience with c/c++ on it.

Yes, it's possible to do this with one of the go plugin options, including Traefik's approach, but they seemed too complex or fragile to me.

hanoii:
> Yes, it's possible to do this with one of the go plugin options, including Traefik's approach, but they seemed too complex or fragile to me.

How, or anything you can point me to look at?

rfay:
* https://github.com/traefik/yaegi
* https://pkg.go.dev/plugin
* https://github.com/hashicorp/go-plugin

And then, of course, creating a built binary that could be executed in an add-on.

PR #6756, opened 2024-11-21
fix: use correct path for Windows in custom configs

stasadev:
## The Issue

We had `ddev-windows-mutagen` failures https://buildkite.com/ddev/ddev-windows-mutagen/builds/4923

It failed on installing `npm install --global forever`.

The contents of `.ddev/web-build/Dockerfile.example` was added on Windows by mistake:

```dockerfile
### From user Dockerfile C:\Users\testbot\tmp\ddevtest\TestPretestAndEnv1956093451\.ddev\web-build\Dockerfile.example:

#ddev-generated
# You can copy this Dockerfile.example to Dockerfile to add configuration
# or packages or anything else to your webimage
# These additions will be appended last to ddev's own Dockerfile
RUN npm install --global forever
RUN echo "Built on $(date)" > /build-date.txt
```

The contents of `Dockerfile.example` should never be added to the final `Dockerfile`: fixed using the correct file path in Windows.
And I added a check for `forever`, it should not be installed.
I think that `filepath.Glob` worked fine, but I added `filepath.Join` to make sure the path is right for Windows.
Change `~\.ddev` to `~/.ddev` in `ddev list` on Windows.
```
 Router │ healthy      │ ~\.ddev          │ http://127.0.0.1:10999 │ traefik
```

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -260,12 +260,12 @@ func (app *DdevApp) WriteConfig() error {
 
 	// Write example Dockerfiles into build directories
 	contents := []byte(`
-#ddev-generated
-# You can copy this Dockerfile.example to Dockerfile to add configuration
-# or packages or anything else to your webimage
-# These additions will be appended last to ddev's own Dockerfile
-RUN npm install --global forever
-RUN echo "Built on $(date)" > /build-date.txt
+## #ddev-generated
+## You can copy this Dockerfile.example to Dockerfile to add configuration
+## or packages or anything else to your webimage
+## These additions will be appended last to DDEV's own Dockerfile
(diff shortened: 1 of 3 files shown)


rfay (changes requested):
This is looking great, thanks.

Let's also

* Remove the forever example, it was a really bad idea. Just put something more relevant there, like something from the docs on Dockerfiles
* Comment out whatever we put in the example.

I think I must have introduced this, when did I do it, was that clear? I remember touching a piece of code using glob.

rfay (approved):
It's great. I do wonder why the `npm install` resulted in it apparently *running* `forever`.

Good to go, top priority. If it works manually on Windows and if one other set of automated tests has passed, we should get it pulled.

PR #6772, opened 2024-11-25
build: always use /usr/local/bin/composer, fixes #6771

### LINKED ISSUE #6771: vendor/bin/composer finally didn't have to take priority

Hi, we have to accept the mistakes we make and I think what I proposed in #6602 was a mistake.

Unfortunately, after proposing this solution, we have seen that by always running the composer from vendor/bin, there is another set of problems and failures when updating Drupal. This has led us to investigate the issue again and clarify whether this should really be the case or not, and we finally came across the following message: https://www.drupal.org/project/drupal/issues/3075785#comment-13227500

It seems that although composer is required as a development dependency in Drupal core, it is actually required as a library and the vendor/bin executable should not be used. This causes some problems if you want to run other executables such as drush from the same folder (making it impossible to use path prioritisation in the $PATH variable).

Given all this, we thought we might have to undo what was done in #6602, as it is not good practice. 

Sorry for the inconvenience.

### Describe your solution

Revert #6602.

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #6771

`vendor/bin/composer` doesn't seem to be what we want, it's confusing and people can't override it with `ddev config --composer-version=<version>`

Reverts behavior to use `/usr/local/bin/composer`.
Removes `vendor/bin/composer` from `$PATH` inside the container with `EXECIGNORE` https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

This way it will use `/usr/local/bin/composer` instead of `/var/www/html/vendor/bin/composer` with `ddev ssh`.

(Prior to this PR, `ddev ssh` && `composer -V` had always preferred `vendor/bin/composer` no matter what.)
`vendor/bin/composer` can still be called directly.

https://ddev--6772.org.readthedocs.build/en/6772/users/configuration/config/#composer_version

```
ddev composer require composer/composer:v2.8.0

ddev composer -V
Composer version 2.8.3 2024-11-17 13:13:04

ddev exec composer -V
Composer version 2.8.3 2024-11-17 13:13:04

ddev exec vendor/bin/composer -V       
Composer version 2.8.0 2024-10-02 16:40:29

ddev ssh

composer -V
Composer version 2.8.3 2024-11-17 13:13:04

vendor/bin/composer -V
Composer version 2.8.0 2024-10-02 16:40:29
```

The only downside is that `EXECIGNORE` has no effect on `which`, so it will still show `/var/www/html/vendor/bin/composer`, but will actually use `/usr/local/bin/composer`, so we will have consistency between inside and outside:

```
ddev ssh

which composer
/var/www/html/vendor/bin/composer

which -a composer
/var/www/html/vendor/bin/composer
/usr/local/bin/composer

composer -V
Composer version 2.8.3 2024-11-17 13:13:04

/var/www/html/vendor/bin/composer -V
Composer version 2.8.0 2024-10-02 16:40:29

/usr/local/bin/composer -V
Composer version 2.8.3 2024-11-17 13:13:04
```

### DIFF
--- a/pkg/ddevapp/composer.go
+++ b/pkg/ddevapp/composer.go
@@ -6,7 +6,6 @@ import (
 	"github.com/mattn/go-isatty"
 	"os"
 	"runtime"
-	"strings"
 )
 
 // Composer runs Composer commands in the web container, managing pre- and post- hooks
@@ -17,23 +16,13 @@ func (app *DdevApp) Composer(args []string) (string, string, error) {
 		return "", "", fmt.Errorf("failed to process pre-composer hooks: %v", err)
 	}
 
-	// Prevent Composer from debugging when Xdebug is enabled
-	env := []string{"XDEBUG_MODE=off"}
-	// Let Composer know which binary to run from the PATH
-	path, _, err := app.Exec(&ExecOpts{
-		Cmd: "echo $PATH",
-	})
-	path = strings.Trim(path, "\n")
-	if err == nil && path != "" {
-		env = append(env, "PATH="+path)
-	}
-
 	stdout, stderr, err := app.Exec(&ExecOpts{
 		Service: "web",
 		Dir:     app.GetComposerRoot(true, true),
 		RawCmd:  append([]string{"composer"}, args...),
 		Tty:     isatty.IsTerminal(os.Stdin.Fd()),
-		Env:     env,
+		// Prevent Composer from debugging when Xdebug is enabled
+		Env: []string{"XDEBUG_MODE=off"},
 	})
 	if err != nil {
 		return stdout, stderr, fmt.Errorf("composer command failed: %v", err)
--- a/cmd/ddev/cmd/composer-create_test.go
+++ b/cmd/ddev/cmd/composer-create_test.go
@@ -16,12 +16,6 @@ import (
 )
 
 func TestComposerCreateCmd(t *testing.T) {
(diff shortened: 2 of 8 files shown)


rfay:
Let's add to doc here what people need to do if they *do* want to use the composer-installed composer.

pkg/ddevapp/composer.go
```diff
@@ -17,23 +16,13 @@ func (app *DdevApp) Composer(args []string) (string, string, error) {
...
-	}
-
 	stdout, stderr, err := app.Exec(&ExecOpts{
 		Service: "web",
 		Dir:     app.GetComposerRoot(true, true),
 		RawCmd:  append([]string{"composer"}, args...),
 		Tty:     isatty.IsTerminal(os.Stdin.Fd()),
-		Env:     env,
+		// Prevent Composer from debugging when Xdebug is enabled
+		Env: []string{"XDEBUG_MODE=off"},
```
rfay, line 25: Is this related to the revert here? Or just a good idea? I thought it was in here before.

rfay (approved):
This seems correct. There's still going to be a little confusion, but I think it's ok.

The weird thing I encountered is this:
`ddev exec which composer` gets vendor/bin
`ddev exec composer --version` gets /usr/local/bin

Not sure how that can happen. I guess because `which` is a bash built-in.

`ddev composer` is always getting the right version.

Docs look good.

Issue #6722, opened 2024-11-14
The relative path ‘~/bin’ is included in the PATH environment variable, which is insecure in combination with the -execdir action of find


mkalkbrenner:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

ddev debug test is not relevant for this issue.

### Expected Behavior

find command with -execdir option should work within the web container if you enter it with `ddev ssh`.
This worked for years but it is broken in latest ddev versions.

### Actual Behavior

find exists with an error, outputting `The relative path ‘~/bin’ is included in the PATH environment variable, which is insecure in combination with the -execdir action of find`

This is caused by this section in `~/.profile`:
```sh
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/bin" ] ; then
    PATH="$HOME/bin:$PATH"
fi
```

Now some of our existing commands are broken.

### Steps To Reproduce

1. login into the web container using `ddev ssh`
2. run `find . -type d -execdir touch {}/foo \;`

stasadev:
Hi @mkalkbrenner,

It's already fixed in:

- #6689

The fix is available in the [DDEV HEAD](https://ddev.readthedocs.io/en/stable/developers/building-contributing/#testing-latest-commits-on-head).

PR #6777, opened 2024-11-26
docs: make installing a Downloaded Binary in the $PATH clearer

gitressa:
If you follow the current instructions, you lose the path, since it is only temporary. So we should clarify that you need to actually  add an extra line in `~.bashrc.`

This is easy for us (more or less) but for people just starting with Unix and the command line, it's not easy at all. I remember it well.

So let's make it more welcoming for those not yet CLI veterans :)

## The Issue

- #<issue number>

https://ddev--6777.org.readthedocs.build/en/6777/developers/building-contributing/#testing-a-pr


docs/content/developers/building-contributing.md
```diff
@@ -67,21 +67,28 @@ brew link --force ddev
 
 ### Installing a Downloaded Binary in the `$PATH`
 
-Normally, you can put any executable in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. This example uses `~/bin`. `echo $PATH` and `which ddev` are valuable commands for debugging. Since not every distro has `~/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
+Normally, you can put any executable in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. This example uses `~/bin`. `echo $PATH` and `which ddev` are valuable commands for debugging. Since not every distro has `~/bin` in `$PATH`, you can create the folder and add it to your path.
 
+Add `~/bin` to your path, by adding this in your `~/.bashrc` file:
 ```
-mkdir -p ~/bin
 export PATH="~/bin:$PATH"
```
stasadev, line 74: This should be `$HOME/bin`, not `~/bin`, because:

- #6722

Let me fix it.

rfay (commented):
I welcome modest improvement, but teaching people the basics of the command line leads to bloated and incomprehensible docs everywhere. And this is in the development section. That said, we do want people to try out PRs.

docs/content/developers/building-contributing.md
```diff
@@ -67,21 +67,28 @@ brew link --force ddev
...
 ```
+Create the `~/bin` folder:
+```
+mkdir ~/bin
+```
+Close and reopen your terminal, and verify that it worked with `echo $PATH`.
 
-Next, unzip the ZIP file you downloaded, make it executable, and move it to a folder in your path. Check with `echo $PATH`:
+Next, download the executable file inside the `~/bin` folder, unzip it, and make it executable:
```
stasadev, line 82: We download a ZIP archive, not a file.
It's better to add a real example with `wget` and `unzip`, because nobody probably uses that. (I will add it.)

gitressa: Sure, I am open for corrections and suggestions.

docs/content/developers/building-contributing.md
gitressa: Great, feel free to update it.

gitressa:
I agree that we should take care, that the docs don't get bloated, but the current instructions were not correct.

So it's a case of just stating "add the binary in your path, and you're good to go" for the CLI veterans, or something a little bit more elaborate ... it's all about striking the balance.

stasadev:
Please take a look
https://ddev--6777.org.readthedocs.build/en/6777/developers/building-contributing/#testing-a-pr

gitressa:
Thanks @stasadev, stellar work as always, it looks great and works well.

rfay (approved):
It's OK with me, but only because it's in a little-used section. The use of tabs makes it less onerous. This is a difficult problem to solve, a difficult balance to keep, but thanks for thinking of the people who need it.

rfay:
@rpkoller actually *does* this all the time, so let's wait for his review.

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
 
 Download and unzip the appropriate binary and place it in your `$PATH`.
 
-## Rollback to a previous version
+!!!tip "You can also [downgrade to an older version of DDEV](../users/usage/faq.md#how-can-i-install-a-specific-version-of-ddev) (perform a rollback)."
 
-You can also [downgrade to an older version of DDEV](../users/usage/faq.md#how-can-i-install-a-specific-version-of-ddev).
+=== "Homebrew with macOS or Linux"
```
rpkoller, line 31: At first i've thought this PR would have added the entire "homebrew with macos or linux" section but turns out after having the current and the new version side by side in preview that this section already existed. tbh  both sections ("homebrew  with macos or linux" and "installing a downloaded binary in the path" are sort of overlapping and doing basically the same. for me as a none technical user having more than one option where each is doing basically the same is rather stalling and freezing instead of providing clear and actionable instructions. i would rather suggest to drop the homebrew section entirely? plus the keyword homebrew is creating false expectations. my expectation based on the title would be the install of the PR binary would be based on homebrew but the sole reason why homebrew is named there is for unlinking ddev and linking it again after the testing is done.  plus by placing the binary in `/usr/local/bin/` sudo is needed. the other approach doesnt need it which would be preferable imho.

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
-unzip ddev.zip
-chmod +x ddev && mv ddev ~/bin
-```
+    Normally, you can put any executable in your path, and it takes precedence, so you don't need to remove or disable an already installed DDEV instance, which we will use here. This example uses `~/bin`. `echo $PATH` and `which ddev` are valuable commands for debugging. Since not every distro has `$HOME/bin` in `$PATH`, you can create the folder and add it to your path in `~/.bashrc` with these commands:
 
-Now, close and reopen your terminal, and verify the replacement worked by running `ddev version`. The output should be something like `DDEV version v1.22.3-39-gfbb878843`, instead of the regular `DDEV version v1.22.3`.
+    ```bash
+    mkdir ~/bin
+    echo 'export PATH="$HOME/bin:$PATH"' >>~/.bashrc
+    # Close and reopen your terminal, and verify that it worked with:
```
rpkoller, line 83:
```suggestion
    . ~/.zshrc
    # Verify that `~/bin` was added to your $PATH
```
instead of the instruction to close and reopen the terminal i would simply add a command to reload the .zshrc file inside the open terminal. and should a comment be added what `. ~/.zshrc` does as well?

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
+    echo $PATH
+    ```
 
-You need to run `ddev poweroff` and `ddev start` to download the Docker images that it needs.
+    Next, download the ZIP archive, unzip it, make it executable, and move it to the `~/bin` folder:
```
rpkoller, line 87: strictly speaking i would expect the screenshot after line 96 right here in this context. here is the call to action to download the zip archive. i would also consider to remove line 90. on macos wget isnt available by default. it might be confusing if you have clear instructions for something that is not available for everyone. why not keep it simple and dont rely on a dl command here. people using commands like wget or curl use those anyway, other less technical users might get confused. if you are already providing the screenshot with the download links then also expect that someone simply clicks one of the links and downloads the necessary artifact?

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
 
-After you’re done testing, you can delete your downloaded executable, restart your terminal, and again use the standard DDEV:
+    ```bash
+    wget -O ddev.zip [LINK TO ZIP ARCHIVE]
+    unzip ddev.zip
+    chmod +x ddev && mv ddev ~/bin/ddev
```
rpkoller, line 92: shouldnt the "macOS and Unsigned Binaries" box be added here as well (in particular in case the other sections would get dropped)?

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
+    rm ddev.zip
+    ```
 
-```
-rm ~/bin/ddev
-```
+    Replace `[LINK TO ZIP ARCHIVE]` with the corresponding link from the PR comment:
```
rpkoller, line 96: according to my second to last comment i would drop this line here.

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
+
+    ![Github Action PR Comment](../images/github-action-pr-comment.png)
+
+    Verify the replacement worked by running `ddev -v`. The output should be something like `ddev version v1.23.5-98-g3c93ae87e`, instead of the regular `ddev version v1.23.5`.
+
+    You need to run `ddev poweroff` and `ddev start` to download the Docker images that it needs.
+
+    After you’re done testing, you can delete your downloaded executable, restart your terminal, and again use the standard DDEV:
```
rpkoller, line 104: after you are done testing i would also recommend running ddev poweroff before deleting the artifact.

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,86 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
+    ```
 
-```
-rm ~/bin/ddev
-```
+    Replace `[LINK TO ZIP ARCHIVE]` with the corresponding link from the PR comment:
+
+    ![Github Action PR Comment](../images/github-action-pr-comment.png)
+
+    Verify the replacement worked by running `ddev -v`. The output should be something like `ddev version v1.23.5-98-g3c93ae87e`, instead of the regular `ddev version v1.23.5`.
```
rpkoller, line 100: the clearer way for me validating everything is set up correctly is instead of using `ddev -v` ( i am using the version of ddev head anyway and it is not easy to distinguish what is a version of head and what is the version of the PR) i would simply run `which ddev` and make sure the path to ddev is located in `~/bin/ddev` (will be the absolute path instead of the ~) instead of `/opt/homebrew/bin/ddev` (in case you are on mac and using homebrew for example like in my case). and i just noticed the introductory paragraph of this section also references `which ddev` so i would definitely adjust this validation step and replace `ddev -v` with `which ddev`

docs/content/developers/building-contributing.md
stasadev: > sort of overlapping and doing basically the same.

I agree.

> plus by placing the binary in `/usr/local/bin/` sudo is needed. the other approach doesnt need it which would be preferable imho.

I want to mention that if you are using PhpStorm, the binary is initially saved for each project.

So if you have the original install at `/usr/local/bin/ddev` (or have a symlink here), add override it with `~/bin/ddev`, and open the project with PhpStorm, it will still use `/usr/local/bin/ddev` for its tasks, which is frustrating (because it will re-download `~/.ddev/bin/docker-compose` all the time).

---

Ideally I prefer to replace the original DDEV binary, and make a backup of the old one:

```
# Make yourself the binary owner and do a backup
binary="$(readlink -f "$(which ddev)")"
sudo chown "${USER}:${USER}" "${binary}"
cp "${binary}" "${binary}.bak"

# Download ZIP file using wget, curl, dl, or using a browser and unzip it
unzip ddev.zip
chmod +x ddev && mv ddev "${binary}"

# Do the testing, and to go back, do the following:
binary="$(readlink -f "$(which ddev)")"
cp "${binary}.bak" "${binary}"
```

I think this will be better in terms of interaction with PhpStorm.

I am already using this technique myself, happy with it, and I can call the original DDEV with `ddev.bak` at the same time.

Let's replace the Homebrew section with this technique.

docs/content/developers/building-contributing.md
stasadev: Thank you, yes, sourcing the file should do the trick, in this case:

```
source ~/.bashrc
```

Edited your suggestion.

docs/content/developers/building-contributing.md
stasadev: Okay, we can replace wget line with a comment:

```
# Download ZIP file using wget, curl, dl, or using a browser and unzip it
```

rfay:
Let's simplify this by doing the same thing in each place, and we probably don't need two tabs.

* The only difference between the homebrew and the download-binary is that you need to `brew unlink ddev` first (perhaps) if you use homebrew, to make sure it doesn't show up in the path. So let's just add that as a step.
* Using ~/bin is probably better because it doesn't require sudo. It also doesn't require creating /usr/local/bin on macOS systems that don't have that (and might not have it in PATH)

The bottom line is, this isn't complex, so let's not make it complex.

rfay:
Probably we should make a script for people, the script could download the particular artifact, choose the right one, etc. Could validate that they have ~/bin in path, or ask them to add it. Could validate that they then have the right thing in their path.

stasadev:
> The only difference between the homebrew and the download-binary is that you need to `brew unlink ddev` first (perhaps) if you use homebrew, to make sure it doesn't show up in the path. So let's just add that as a step.

We add the binary to `~/bin/ddev`, and have `export PATH="$HOME/bin:$PATH"`, do we really need to unlink homebrew given that '$HOME/bin' will come first?

docs/content/developers/building-contributing.md
```diff
@@ -26,70 +26,87 @@ Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) cre
...
+
+    You need to run `ddev poweroff` and `ddev start` to download the Docker images that it needs.
```
tyler36, line 103:
```suggestion
    When DDEV detects a version change, it recommends powering down all running containters. Then, it will download the new images, if required.
```

DDEV does not need to "poweroff" to get new images:

```shell 
$ ddev start
You seem to have a new DDEV version.
During an upgrade it's important to `ddev poweroff`.
May I do `ddev poweroff` before continuing?
This does no harm and loses no data. [Y/n] (yes): n
Starting cypress...
20241122_stasadev_php_pecl: Pulling from ddev/ddev-webserver
132e290b53a3: Download complete
Digest: sha256:1111b1a5766693de8b432149941433a432c8d53f41c24ba1c1efd443a6e6f912
Status: Downloaded newer image for ddev/ddev-webserver:20241122_stasadev_php_pecl
```

rfay (approved):
This is OK and it sticks to the point. I hope it still works for you @gitressa

gitressa:
It works well, and is more straight forward than before, and will be a good guide for new users. It will also be helpful for me, next time I need to test a PR, and I forgot how to :)

The only thing that might give a false impression, is that the downloaded file will be called something like "2246879499.zip" and not nicely named "ddev-linux-amd64.zip" ... but that's a detail, and most will probably figure this out.

Perhaps it could be made clear where the artifacts are placed? Something like this? I have bolded and italicized my insert. It should just be plain text if it gets added:

> Each [PR build](https://github.com/ddev/ddev/actions/workflows/pr-build.yml) creates GitHub artifacts, you can use for testing, so you can download the one you need from the PR page **_under "Download the artifacts for this pull request"_**, install it locally, and test using that build.

Also, could this be added somewhere in case someone wants to try out the process, and just needs a quick zip.file? 

> You can use the nightly [build zip-files](https://nightly.link/ddev/ddev/workflows/master-build/master) if you need a zip-file to try out the process.

But these my comments are just details, and it will be a big improvement in the current state.

rpkoller:
> The only thing that might give a false impression, is that the downloaded file will be called something like "2246879499.zip" and not nicely named "ddev-linux-amd64.zip" ... but that's a detail, and most will probably figure this out.
if you hover over the different artifact links in the "download the artifacts for this pull request:" list then each file is numeric only that is true, but when downloaded it turns into the name that is listed on list created by github actions (and also seen on the screenshot). so at least i am ending up with the correctly named file name, ddev-macos-arm64.zip in my case. for you the file is not renamed automatically by the browser?

gitressa:
Thanks @rpkoller I did try that, and that's correct -- but I don't click -- I use `wget` :)

docs/content/developers/building-contributing.md
```diff
@@ -24,70 +24,45 @@ There are several ways to use DDEV’s latest-committed HEAD version:
...
+```bash
+mkdir ~/bin
+echo 'export PATH="$HOME/bin:$PATH"' >>~/.bashrc
+source ~/.bashrc
+# Verify that `$HOME/bin` is the first entry in your `$PATH`
+echo $PATH
 ```
 
-Next, unzip the binary you downloaded, make it executable, and move it to your bin folder:
+Choose a ZIP file for your OS and architecture, download it using `wget`, `curl`, `dl` or using a browser and unzip it, make it executable, and move it to the `~/bin` folder:
```
gitressa, line 39: Thinking more about this, the reason it irked me when I hovered over the links, was that I saw `[...]/2246879499.zip` which are "automagically" transformed.

I am not a fan of magic with computers, but in the end, clicking seems to be the easiest method.

It would work better for me, if there was a basic explanation of what it looks like, and what happens -- then I have no problem recommending the click-based solution. Something like this, recommending the click method, by moving it first and clarifying how it works:

```suggestion
Download a ZIP file for your OS and architecture by clicking the link (the result is something like `ddev-macos-arm64.zip`) or using `wget`, `curl`, `dl` and unzip it, make it executable, and move it to the `~/bin` folder:
```

docs/content/developers/building-contributing.md
```diff
@@ -24,70 +24,45 @@ There are several ways to use DDEV’s latest-committed HEAD version:
...
-chmod +x ddev && sudo mv ddev /usr/local/bin/ddev
+```bash
+# Example for macOS Apple Silicon:
+unzip ddev-macos-arm64.zip
+chmod +x ddev && mv ddev ~/bin/ddev
+rm ddev-macos-arm64.zip
 ```
 
-Verify the replacement worked by running `ddev -v`. The output should be something like `ddev version v1.22.5-alpha1-70-g0852fc2df`, instead of the regular `ddev version v1.22.5`.
+![Github Action PR Comment ZIP files](../images/github-action-pr-comment.png)
```
gitressa, line 48: Could a link to the nightly builds be added, in case someone wants to try out the "Testing a PR" process, and just needs a quick zip.file? I could easily see myself need this, at a later point in time, and not have to dig through PR's, to find a list of artifacts would be nice ...

```suggestion
![Github Action PR Comment ZIP files](../images/github-action-pr-comment.png)

Tip: If you need a zip-file to try out the "Testing a PR" process, see the [nightly builds](https://nightly.link/ddev/ddev/workflows/master-build/master).
```

gitressa:
After sleeping on it, I have realized that clicking is the way to go after all, and have added two comments.

gitressa:
Thanks to everyone who helped improve the docs, they look really good now.
https://ddev.readthedocs.io/en/latest/developers/building-contributing/#testing-a-pr

PR #6779, opened 2024-11-26
build: do not run `.buildkite/test.sh` if we want to skip

stasadev:
## The Issue

All builds are waiting for agents to be available, and we have this check:

https://github.com/ddev/ddev/blob/aa269d58ea9b13c4a1ccb51b249e168cde314314/.buildkite/test.sh#L154-L158

It only runs when the agent is available, but we can do it with Buildkite itself without waiting for the agent:

```yaml
if: |
  build.message !~ /\[(skip ci|skip buildkite)\]/
```

Adds conditional https://buildkite.com/docs/pipelines/configure/conditionals#conditionals-in-steps

Check if `[skip ci]` or `[skip buildkite]` will skip the test from running i.e. remove this endless waiting for the agent.

I mean these tests:

image

### DIFF
--- a/.buildkite/macos-colima.yml
+++ b/.buildkite/macos-colima.yml
@@ -1,8 +1,10 @@
 # Colima with qemu and sshfs
 # See https://buildkite.com/ddev/macos-colima/settings/repository
-# Runs on master and PRs, including forked PRs
+# Not currently used.
 
   - command: ".buildkite/test.sh"
+    if: |
+      build.message !~ /\[(skip ci|skip buildkite)\]/
     agents:
       - "os=macos"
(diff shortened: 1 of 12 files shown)


rfay (approved):
Works for me, thanks!

stasadev:
It works.

Look at the tests for each commit in this PR. (It doesn't wait for agents with `[skip ci]` or `[skip buildkite]`).

---

P.S. It would be nice to have this condition too, but I don't see a way how to do it:

https://github.com/ddev/ddev/blob/aa269d58ea9b13c4a1ccb51b249e168cde314314/.buildkite/test.sh#L160-L164

PR #6749, opened 2024-11-20
fix: Deprecate and hide ddev service commands, fixes #6298

rfay:
## The Issue

- #6298 

Deprecate and hide `ddev service enable` and `ddev service disable` as long-since replaced by `ddev add-on`

## Manual Testing Instructions

* Verify that `ddev service` doesn't show up in `ddev` or `ddev help`
* Verify that these commands can still be used
* Verify that they show a warning.

PR #6782, opened 2024-11-27
docs: Update version-history.md for v1.24 [skip buildkite]

rfay:
## The Issue

Update version-history.md for v1.24.0

### DIFF
--- a/version-history.md
+++ b/version-history.md
@@ -10,9 +10,9 @@ This version history has been driven by what we hear from our wonderful communit
 
 - [x] Added PHP 8.4. PHP 8.3 is default for new projects
 - [x] NodeJS 22 becomes default
-- [x] Extensive support for matching `mysql`, `mariadb`, and `pq` clients so the client matches the server.
+- [x] Extensive support for matching `mysql`, `mariadb`, and `psql` clients so the client matches the server.
 - [x] MariaDB 11.4 LTS support
-- [x] Traditional Windows ARM64 (Windows Copilot) support
+- [x] Traditional Windows ARM64 support
 - [x] `ddev_version_constraint` option added for both `config.yaml` and Add-on `install.yaml`
 - [x] `ddev get` changed to `ddev add-on get`
 - [x] Traefik router is default, and supports hosting mode, with Let's Encrypt.
--- a/version-history.md
+++ b/version-history.md
@@ -8,7 +8,7 @@ This version history has been driven by what we hear from our wonderful communit
 
 ## v1.24 (Released 2024-12)
 
-- [x] PHP 8.4, PHP 8.3 becomes default version on new projects
+- [x] Added PHP 8.4. PHP 8.3 is default for new projects
 - [x] NodeJS 22 becomes default
 - [x] Extensive support for matching `mysql`, `mariadb`, and `pq` clients so the client matches the server.
 - [x] MariaDB 11.4 LTS support
@@ -22,8 +22,8 @@ This version history has been driven by what we hear from our wonderful communit
 - [x] `ddev auth ssh` can be used with individual files and against symlinks.
 - [x] 25% smaller `ddev-webserver` Docker image
 - [x] `drupal` project type introduced as an alias for latest-stable Drupal version
-- [x] `ddev service enable` and `ddev service disable` commands removed in favor of `ddev add-on` commands.
-- [x] Only limited locales are now included by default in `ddev-webserver`, all locales can be enabled with `webimage_extra_packages: [locales-all]`
+- [x] `ddev service enable` and `ddev service disable` commands deprecated in favor of `ddev add-on` commands.
+- [x] Reduced number of locales included by default in `ddev-webserver`, all locales can be enabled with `webimage_extra_packages: [locales-all]`
 
 ## v1.23 (Released 2024-04)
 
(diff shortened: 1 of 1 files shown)


version-history.md
```diff
@@ -6,6 +6,25 @@ search:
 
 This version history has been driven by what we hear from our wonderful community of users. Please share your comments or ideas in the [issue queue](https://github.com/ddev/ddev/issues). We listen. Or talk to us in any of the [support locations](https://ddev.readthedocs.io/en/stable/#support).
 
+## v1.24 (Released 2024-12)
+
+- [x] PHP 8.4, PHP 8.3 becomes default version on new projects
```
tyler36, line 11:
```suggestion
- [x] Added PHP 8.4. PHP 8.3 is default for new projects
```

version-history.md
```diff
@@ -6,6 +6,25 @@ search:
...
+- [x] `ddev_version_constraint` option added for both `config.yaml` and Add-on `install.yaml`
+- [x] `ddev get` changed to `ddev add-on get`
+- [x] Traefik router is default, and supports hosting mode, with Let's Encrypt.
+- [x] `ddevcd` command allows easy switching between project directories
+- [x] `nginx-proxy` router type removed
+- [x] `python` and `django4` project types removed
+- [x] `ddev auth ssh` can be used with individual files and against symlinks.
+- [x] 25% smaller `ddev-webserver` Docker image
+- [x] `drupal` project type introduced as an alias for latest-stable Drupal version
+- [x] `ddev service enable` and `ddev service disable` commands removed in favor of `ddev add-on` commands.
```
tyler36, line 25:
```suggestion
- [x] `ddev service enable` and `ddev service disable` commands deprecated in favor of `ddev add-on` commands.
```

#6749 has it depreciated.

version-history.md
```diff
@@ -6,6 +6,25 @@ search:
...
+- [x] Only limited locales are now included by default in `ddev-webserver`, all locales can be enabled with `webimage_extra_packages: [locales-all]`
```
tyler36, line 26:
```suggestion
- [x] Reduced number of locales included by default in `ddev-webserver`, all locales can be enabled with `webimage_extra_packages: [locales-all]`
```

tyler36 (commented):
Some lines end in period and others do not. Not sure if you want consistency.

version-history.md
```diff
@@ -6,6 +6,25 @@ search:
...
+- [x] Traditional Windows ARM64 (Windows Copilot) support
+- [x] `ddev_version_constraint` option added for both `config.yaml` and Add-on `install.yaml`
+- [x] `ddev get` changed to `ddev add-on get`
+- [x] Traefik router is default, and supports hosting mode, with Let's Encrypt.
+- [x] `ddevcd` command allows easy switching between project directories
+- [x] `nginx-proxy` router type removed
+- [x] `python` and `django4` project types removed
+- [x] `ddev auth ssh` can be used with individual files and against symlinks.
+- [x] 25% smaller `ddev-webserver` Docker image
+- [x] `drupal` project type introduced as an alias for latest-stable Drupal version
```
rpkoller, line 24: strictly speaking, the project type `drupal` was already introduced in, was it 1.23.0(?), wouldn't it be more accurate to use something like "becomes"?

version-history.md
rpkoller: would it make sense to combine line 26 and 23... thematically those two belong together and are interconnected? and that part "...by default in `ddev-webserver`" feels incomplete and also somehow missing what `ddev-webserver` actually is (Docker image), something that was provided for lines 23?

rpkoller (commented):
left only two nitpicks, and another point to note that is applying to all the given points. some of the sentences end with a period some not, i think that should be consistent?

version-history.md
rfay: This is taking v1.24 as a whole, all the things from v1.23.1 up to v1.24.0, so the detail doesn't matter much.

version-history.md
rfay: From the user's perspective they are different things. User doesn't much care *why* it's smaller. User does care what to do if their favorite locale isn't there any more.

version-history.md
```diff
@@ -6,6 +6,25 @@ search:
 
 This version history has been driven by what we hear from our wonderful community of users. Please share your comments or ideas in the [issue queue](https://github.com/ddev/ddev/issues). We listen. Or talk to us in any of the [support locations](https://ddev.readthedocs.io/en/stable/#support).
 
+## v1.24 (Released 2024-12)
+
+- [x] Added PHP 8.4. PHP 8.3 is default for new projects
+- [x] NodeJS 22 becomes default
+- [x] Extensive support for matching `mysql`, `mariadb`, and `pq` clients so the client matches the server.
```
stasadev, line 13:
```suggestion
- [x] Extensive support for matching `mysql`, `mariadb`, and `psql` clients so the client matches the server.
```

version-history.md
```diff
@@ -6,6 +6,25 @@ search:
...
+- [x] MariaDB 11.4 LTS support
+- [x] Traditional Windows ARM64 (Windows Copilot) support
```
stasadev, line 15: I'm not sure what 'Windows Copilot' has to do with this?

stasadev (approved):

version-history.md
rfay: Most of those ARM64 computers of the current generation are labeled "Windows Copilot". However, since that will likely be obsolete before anybody reads this...

```suggestion
- [x] Traditional Windows ARM64 support
```

PR #6787, opened 2024-11-28
build: traefik: bump image, add timezone, show custom configs, update healthcheck, fixes #6463

### LINKED ISSUE #6463: Traefik router healthcheck.sh gives no useful output when everything fails
### Expected Behavior

When a healthcheck fails, it should give us some information about why it's failing. Right now, we don't have any.

See https://discord.com/channels/664580571770388500/1269927348447346689

### Actual Behavior

In the discord issue, we see that traefik is giving no output at all (https://gist.github.com/comanche76/d45b561a42f41e3684fe98e4bd9aff51#file-docker_inspect_ddev-router-txt): 
> {"Status":"unhealthy","FailingStreak":0,"Log":[]}

The healthcheck is https://github.com/ddev/ddev/blob/master/containers/ddev-traefik-router/traefik_healthcheck.sh

It does seem that if none of the happy cases work, it doesn't output any useful information. So in this case, it wasn't healthy, it wasn't initializing, and the curl on the router ping failed. 

However, we used curl -s on that, so we're not seeing that output. 

So we can improve the healthcheck.sh to give information on why it isn't healthy. Which case failed...

It may be that there are other healthchecks that aren't as good as they could be.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

stasadev:
## The Issue

- #6463

- Bumps Traefik to the latest v3, as of today it's v3.2.
- Adds an automatically detected timezone to `ddev-router`, so `docker logs ddev-router` shows the local time.
- Shows custom Traefik configs (including custom global configs) on `ddev start`.
- Uses standard `traefik healthcheck` script for healthcheck (the key point was to add `--configFile`).
- Removes noise from `docker-compose` v2.31.0.

### DIFF
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -202,6 +202,7 @@ func generateRouterCompose() (string, error) {
 	dockerIP, _ := dockerutil.GetDockerIP()
 
 	uid, gid, username := util.GetContainerUIDGid()
+	timezone, _ := util.GetLocalTimezone()
 
 	templateVars := map[string]interface{}{
 		"Username":                   username,
@@ -215,6 +216,7 @@ func generateRouterCompose() (string, error) {
 		"letsencrypt_email":          globalconfig.DdevGlobalConfig.LetsEncryptEmail,
 		"Router":                     globalconfig.DdevGlobalConfig.Router,
 		"TraefikMonitorPort":         globalconfig.DdevGlobalConfig.TraefikMonitorPort,
+		"Timezone":                   timezone,
 	}
 
 	t, err := template.New("router_compose_template.yaml").ParseFS(bundledAssets, "router_compose_template.yaml")
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -178,7 +178,7 @@ func StartDdevRouter() error {
 	util.Debug(`Waiting for ddev-router to become ready, timeout=%v`, routerWaitTimeout)
 	logOutput, err := dockerutil.ContainerWait(routerWaitTimeout, label)
 	if err != nil {
-		return fmt.Errorf("ddev-router failed to become ready; logOutput=%s, err=%v", logOutput, err)
+		return fmt.Errorf("ddev-router failed to become ready; log=%s, err=%v", logOutput, err)
 	}
 	util.Debug("ddev-router is ready")
 
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -175,10 +175,10 @@ func StartDdevRouter() error {
 	if globalconfig.DdevGlobalConfig.UseLetsEncrypt {
 		routerWaitTimeout = 180
 	}
-	util.Debug(`Waiting for ddev-router to become ready. docker inspect --format "{{json .State.Health }}" ddev-router`)
+	util.Debug(`Waiting for ddev-router to become ready, timeout=%v`, routerWaitTimeout)
 	logOutput, err := dockerutil.ContainerWait(routerWaitTimeout, label)
 	if err != nil {
-		return fmt.Errorf("ddev-router failed to become ready; debug with 'docker logs ddev-router' and 'docker inspect --format \"{{json .State.Health }}\" ddev-router'; logOutput=%s, err=%v", logOutput, err)
+		return fmt.Errorf("ddev-router failed to become ready; logOutput=%s, err=%v", logOutput, err)
 	}
 	util.Debug("ddev-router is ready")
 
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -727,7 +727,7 @@ func ComposeCmd(cmd *ComposeCmdOpts) (string, string, error) {
 	// Container (or Volume) ... Creating or Created or Stopping or Starting or Removing
 	// Container Stopped or Created
 	// No resource found to remove (when doing a stop and no project exists)
-	ignoreRegex := "(^ *(Network|Container|Volume) .* (Creat|Start|Stopp|Remov)ing$|^Container .*(Stopp|Creat)(ed|ing)$|Warning: No resource found to remove|Pulling fs layer|Waiting|Downloading|Extracting|Verifying Checksum|Download complete|Pull complete)"
+	ignoreRegex := "(^ *(Network|Container|Volume|Service) .* (Creat|Start|Stopp|Remov|Build|Buil)(ing|t)$|^Container .*(Stopp|Creat)(ed|ing)$|Warning: No resource found to remove|Pulling fs layer|Waiting|Downloading|Extracting|Verifying Checksum|Download complete|Pull complete)"
 	downRE, err := regexp.Compile(ignoreRegex)
 	if err != nil {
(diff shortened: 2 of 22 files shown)


pkg/ddevapp/router.go
```diff
@@ -175,10 +175,10 @@ func StartDdevRouter() error {
 	if globalconfig.DdevGlobalConfig.UseLetsEncrypt {
 		routerWaitTimeout = 180
 	}
-	util.Debug(`Waiting for ddev-router to become ready. docker inspect --format "{{json .State.Health }}" ddev-router`)
+	util.Debug(`Waiting for ddev-router to become ready, timeout=%v`, routerWaitTimeout)
 	logOutput, err := dockerutil.ContainerWait(routerWaitTimeout, label)
 	if err != nil {
-		return fmt.Errorf("ddev-router failed to become ready; debug with 'docker logs ddev-router' and 'docker inspect --format \"{{json .State.Health }}\" ddev-router'; logOutput=%s, err=%v", logOutput, err)
+		return fmt.Errorf("ddev-router failed to become ready; logOutput=%s, err=%v", logOutput, err)
```
stasadev, line 181: There was some duplication, `err` already included troubleshooting commands.

pkg/dockerutil/dockerutils.go
```diff
@@ -727,7 +727,7 @@ func ComposeCmd(cmd *ComposeCmdOpts) (string, string, error) {
 	// Container (or Volume) ... Creating or Created or Stopping or Starting or Removing
 	// Container Stopped or Created
 	// No resource found to remove (when doing a stop and no project exists)
-	ignoreRegex := "(^ *(Network|Container|Volume) .* (Creat|Start|Stopp|Remov)ing$|^Container .*(Stopp|Creat)(ed|ing)$|Warning: No resource found to remove|Pulling fs layer|Waiting|Downloading|Extracting|Verifying Checksum|Download complete|Pull complete)"
+	ignoreRegex := "(^ *(Network|Container|Volume|Service) .* (Creat|Start|Stopp|Remov|Build|Buil)(ing|t)$|^Container .*(Stopp|Creat)(ed|ing)$|Warning: No resource found to remove|Pulling fs layer|Waiting|Downloading|Extracting|Verifying Checksum|Download complete|Pull complete)"
```
stasadev, line 730: Remove noise from `docker-compose` v2.31.0

stasadev:
I tested the healthcheck, by modifying `/mnt/ddev-global-cache/traefik/.static_config.yaml` after starting the project.

Seems to work fine:

```
$ docker inspect --format "{{ json .State.Health }}" ddev-router | docker run -i --rm ddev/ddev-utilities jq -r
{
  "Status": "unhealthy",
  "FailingStreak": 473,
  "Log": [
    {
      "Start": "2024-11-29T17:22:15.691055385+02:00",
      "End": "2024-11-29T17:22:15.871498861+02:00",
      "ExitCode": 1,
      "Output": "Traefik healthcheck failed: {\"level\":\"error\",\"error\":\"command healthcheck error: field not found, node: address123\",\"time\":\"2024-11-29T15:22:15Z\",\"message\":\"Command error\"}"
    },
```

What interesting is that `http://127.0.0.1:10999/dashboard/#/` is still working, and `http://127.0.0.1:10999/ping` shows OK, but `traefik healthcheck --ping` shows the error correctly, and container status is `unhealthy` as it should be.

rfay:
I wonder if there's a way we can quiet this useless SNI message, `2024-11-29T20:07:12Z WRN No domain found in rule HostRegexp(`^d11\.ddev\.site$`), the TLS options applied for this router will depend on the SNI of each request entryPointName=http-443 routerName=d11-web-80-https@file`

rfay:
I have deployed this for my little prod server and sites.

rfay:
I had 38000 lines of traefik log in my little system after a few hours. I'm not sure where docker stores container logs or whether they're eating memory, but I think we might want to do LogLevel: ERROR, and then show people how to do a `static_config.loglevel.yaml` with

```
log:
  level: INFO
```

rfay (changes requested):
This is great, I just have one thing I'd like you to look at. If I break the configuration with an invalid rule in project, for example,
```
http:
  routers:
    d11-web-80-http:
      entrypoints:
        - http-80
        - nohting
      rule: HostRegexp(`^d11\.ddev\.site$`)
      rule: extrarule
 ```
 
 The router fails, kind of as described in #6463 (but the `docker logs ddev-router` does explain it)
 
 `2024-12-01T15:04:25Z ERR Error while building configuration (for the first time) error="/mnt/ddev-global-cache/traefik/config/d11.yaml: yaml: unmarshal errors:\n  line 12: mapping key \"rule\" already defined at line 11" providerName=file`

But it would be better if a complete failure like this (that breaks the output of the project's URL because it wasn't parsed) could show in the healthcheck. I think you already explored that and didn't find a way. 

Could we read the logs from ddev-router after start and look for ERR? Or look for "ERR Error while building configuration"?

**Comment by stasadev** [comment]
> This is great, I just have one thing I'd like you to look at. If I break the configuration with an invalid rule in project

> The router fails, kind of as described in #6463 (but the `docker logs ddev-router` does explain it)

>But it would be better if a complete failure like this (that breaks the output of the project's URL because it wasn't parsed) could show in the healthcheck. I think you already explored that and didn't find a way. 

It's disappointing that Traefik hasn't addressed requests for YAML validation, such as:

- https://github.com/traefik/traefik/issues/6694

> Could we read the logs from ddev-router after start and look for ERR? Or look for "ERR Error while building configuration"?  I think we could just read the logs after `ddev start` after it gets past the nginx section and then just warn if there were issues?

I tried capturing logs from `ddev-router` with `app.CaptureLogs()` right after `ddevapp.StartDdevRouter()` and filtering errors related to YAML with:

```go
re := regexp.MustCompile(fmt.Sprintf(`(?m)^.*ERR.*%s.yaml.*$`, app.Name))
matches := re.FindAllString(routerLogs, -1)
```

While this approach works, it add a significant delay to the `ddev start`. I decided not to do that.

---

I think it's better to display a warning to the user when they are using a custom Traefik configuration, so they know what to do if it doesn't work as expected:

```
Using custom Traefik configuration (use `docker logs ddev-router` for troubleshooting): [
    /path/to/traefik/file/here/.ddev/traefik/config/d11.yaml
]
```

---

Additionally, I passed `TZ` to the `ddev-router` container, so `docker logs ddev-router` shows the local time.

rfay:
> I think it's better to display a warning to the user when they are using a custom Traefik configuration, so they know what to do if it doesn't work as expected:

I can't believe we didn't have that in there before, it's super important. Good technique, thanks. Works fine. 

> Additionally, I passed TZ to the ddev-router container, so docker logs ddev-router shows the local time.

I'm so very happy about this. Studying the logs I very much missed this and got a bit confused.

pkg/dockerutil/dockerutils.go
rfay: I noticed the new chatter in the other traefik logs PR and was glad to see you fixed this here.

rfay (approved):
Thanks for this, I think it's a great improvement. 

I could swear that we have test coverage for warnings about config overrides, but can't find it. When we find it we should make sure it covers this. Doesn't have to be now. May not cover web-entrypoint.d either.

We can pull this when you're ready.

PR #6743, opened 2024-11-19
refactor: use `ddevcd` function with `ddev debug cd` instead of `ddev cd`, fixes #6737

stasadev:
## The Issue

- #6737

## How This PR Solves The Issue

Replaces global `ddev` function wrapper with `ddevcd` function while keeping autocompletion.

Hides top level `ddev cd` to `ddev debug cd`.

## Manual Testing Instructions

Follow instructions from `ddev debug cd`, restart your shell, and use `ddevcd <tab>`

PR #6789, opened 2024-11-28
fix: do not run project validation for `ddevcd`, for #6743

stasadev:
## The Issue

- #6743

I noticed that I couldn't `ddevcd` to a project with a broken configuration, but DDEV should allow it.

- Disables project validation for `ddevcd`.
- Hides `stderr` from completions for `ddevcd`, I noticed that when I press `<TAB>` when `ddevcd` is available and when I use v1.23.5, it shows an error, but I didn't even press enter.

Break project config and run `ddevcd`:

```
ddevcd d11

# break it, for example replace type with `type: blablabla`
vim .ddev/config.yaml

cd ~

# it should `cd` to `d11` project
ddevcd d11
```

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/shells/ddev.fish
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/shells/ddev.fish
@@ -8,11 +8,11 @@
 # "ddevcd project-name" to cd into the project directory.
 
 function ddevcd
-  cd (DDEV_VERBOSE=false ddev debug cd "$argv[1]" --get-approot)
+  cd (DDEV_VERBOSE=false DDEV_DEBUG=false ddev debug cd "$argv[1]" --get-approot)
 end
 
 function __ddevcd_autocomplete
-  DDEV_VERBOSE=false ddev debug cd --list 2>/dev/null
+  DDEV_VERBOSE=false DDEV_DEBUG=false ddev debug cd --list 2>/dev/null
 end
 
 complete -c ddevcd -f -a "(__ddevcd_autocomplete)"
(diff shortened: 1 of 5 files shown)


pkg/ddevapp/global_dotddev_assets/commands/host/shells/ddev.fish
```diff
@@ -12,7 +12,7 @@ function ddevcd
 end
 
 function __ddevcd_autocomplete
-  DDEV_VERBOSE=false ddev debug cd --list
+  DDEV_VERBOSE=false ddev debug cd --list 2>/dev/null
```
rfay, line 15: Should we have `DDEV_DEBUG=false` too, just for you?

rfay (approved):
Thanks, I certainly have no objections!

pkg/ddevapp/global_dotddev_assets/commands/host/shells/ddev.fish
stasadev: Although it is not needed now, it is still good to add it.

pkg/globalconfig/global_config.go
```diff
@@ -835,9 +835,9 @@ func IsInternetActive() bool {
 
 	// Internet is active (active == true) if both err and ctx.Err() were nil
 	active := err == nil && ctx.Err() == nil
-	if os.Getenv("DDEV_DEBUG") != "" {
+	if DdevDebug {
```
rfay, line 838: Better, thanks.

rfay (approved):
Thanks, looks like it's ready to pull.

PR #6794, opened 2024-12-01
fix: Change traefik log level to ERROR and limit accessLog

rfay:
## The Issue

In reviewing 
- #6787

I noted that we had loads of log info we don't need, and way too much (access logs) in general.

* Change the default log level to ERROR (affects traefik events like config fails, etc)
* Change access log to only "unusual" things like 300-510

Try `docker logs -f ddev-router` with a number of configurations.

Rendered Traefik configuration updates in https://ddev--6794.org.readthedocs.build/en/6794/users/extend/traefik-router/#traefik-static-configuration

### DIFF
--- a/docs/content/users/extend/traefik-router.md
+++ b/docs/content/users/extend/traefik-router.md
@@ -20,10 +20,10 @@ All Traefik configuration uses the *file* provider, not the *Docker* provider. E
 Static configuration is automatically generated in the `~/.ddev/traefik` directory. "Static" configuration means Traefik configuration which is only read when the router is started.
 
 * `.static_config.yaml` (a hidden file) is the configuration that gets used. It is not to be edited; it is generated from DDEV's base configuration while merging any files named `static_config.*.yaml`. It is read on router startup, and does not change until the router starts again (normally after `ddev poweroff`).
-* Additional static configuration may be added by adding `static_config.*.yaml` files, which will be merged into the generated `.static_config.yaml`. For example, a `static_config.logs.yaml` might override logging configuration, a `static_config.plugin.yaml` might contain external Traefik plugins, or a `static_config.dnschallenge.yaml` might provide configuration for additional `certificatesResolvers`. Merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a particular element of the YAML structure wins.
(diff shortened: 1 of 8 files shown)


docs/content/users/extend/traefik-router.md
```diff
@@ -20,18 +20,28 @@ All Traefik configuration uses the *file* provider, not the *Docker* provider. E
...
+* Additional static configuration may be added by adding `static_config.*.yaml` files, which will be merged into the generated `.static_config.yaml`. For example, a `static_config.logs.yaml` might override logging configuration, a `static_config.plugin.yaml` might contain external Traefik plugins, or a `static_config.dnschallenge.yaml` might provide configuration for additional `certificatesResolvers`. Merging is done with an **override** strategy, meaning that the final file in alphanumeric sort to touch a particular element of the YAML structure wins.
     Some examples of `static_config.*.yaml` files are:
+
+    * `static_config.logs.yaml`:
+
+        ```yaml
+        # Enable extensive error and access logging
+        log:
+          level: DEBUG
+        accessLog: {}
```
stasadev, line 32: Because `accessLog: {}` override doesn't work (it doesn't update `.static_config.yaml`):

```suggestion
        accessLog:
          filters:
            statusCodes: {}
```

stasadev (changes requested):
Thank you, we also need to add `.ddev/traefik/static_config.logs.yaml.example`.

rfay:
> Thank you, we also need to add .ddev/traefik/static_config.logs.yaml.example.

Added it, thanks.

pkg/ddevapp/global_dotddev_assets/traefik/static_config.loglevel.yaml.example
stasadev, line 1: The filename mentioned in elsewhere is `static_config.logs.yaml`, but here it is `static_config.loglevel.yaml`.

stasadev (changes requested):

pkg/ddevapp/global_dotddev_assets/traefik/static_config.loglevel.yaml.example
rfay: The example wasn't mentioned before I don't think, but I'll change the mention in the docs.

stasadev: Okay, I thought each `~/.ddev/traefik/static_config.*.yaml.example` had a corresponding section in the docs, now I can see the difference.

(I had some leftover files inside `~/.ddev/traefik` which made me think so.)

Thank you for the change.

stasadev (approved):
Looks good to me.

PR #6796, opened 2024-12-02
refactor: make `ddev auth ssh` flags mutually exclusive, fixes #6795

### LINKED ISSUE #6795: `ddev auth ssh -f` and `ddev auth ssh -d` should be mutually exclusive, simplify examples
### Preliminary checklist

```
```
</details>

### Expected Behavior

Easier usage of `ddev auth ssh`

### Actual Behavior

You can use the two together

### Steps To Reproduce

_No response_

### Anything else?

I think we'd be better off to make these mutually exclusive, and simplify the examples. 

I think `ddev auth ssh -f` will be very popular, perhaps used extensively

We should also note that you can run the command more than once, in addition to using multiple `-f` flags.

stasadev:
## The Issue

- #6795

- Makes `ddev auth ssh -f` and `ddev auth ssh -d` mutually exclusive/
- Clarifies that command can be run multiple times, and the flags can be passed multiple times.
- Explicitly says "private" for SSH keys (because they can be "public").

This should fail:

```
ddev auth ssh -d ~/.ssh -f ~/custom/path/to/ssh/id_ed25519
```

### DIFF
--- a/cmd/ddev/cmd/auth-ssh.go
+++ b/cmd/ddev/cmd/auth-ssh.go
@@ -23,7 +23,7 @@ var sshKeyFiles, sshKeyDirs []string
 var AuthSSHCommand = &cobra.Command{
 	Use:   "ssh",
 	Short: "Add SSH private key authentication to the ddev-ssh-agent container",
-	Long:  `Use this command to provide the password to your SSH private key to the ddev-ssh-agent container, where it can be used by other containers. The command can be executed multiple times to add more keys to the same ssh-agent.`,
+	Long:  `Use this command to provide the password to your SSH private key to the ddev-ssh-agent container, where it can be used by other containers. The command can be executed multiple times to add more keys.`,
 	Example: heredoc.DocI2S(`
 		ddev auth ssh
 		ddev auth ssh -d ~/custom/path/to/ssh
@@ -184,8 +184,8 @@ func echoDockerCmd(dockerCmd []string) string {
 }
 
 func init() {
-	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyFiles, "ssh-key-file", "f", nil, "full path to SSH private key file, use the flag multiple times to add more keys")
-	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyDirs, "ssh-key-path", "d", nil, "full path to directory with SSH private key(s), use the flag multiple times to add more directories")
+	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyFiles, "ssh-key-file", "f", nil, "path to SSH private key file, use the flag multiple times to add more keys")
+	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyDirs, "ssh-key-path", "d", nil, "path to directory with SSH private key(s), use the flag multiple times to add more directories")
 	// While both flags work well with each other, don't allow them to be passed at the same time to make it easier to use.
 	AuthSSHCommand.MarkFlagsMutuallyExclusive("ssh-key-file", "ssh-key-path")
 
--- a/cmd/ddev/cmd/auth-ssh.go
+++ b/cmd/ddev/cmd/auth-ssh.go
@@ -22,13 +22,12 @@ var sshKeyFiles, sshKeyDirs []string
 // AuthSSHCommand implements the "ddev auth ssh" command
 var AuthSSHCommand = &cobra.Command{
 	Use:   "ssh",
-	Short: "Add SSH key authentication to the ddev-ssh-agent container",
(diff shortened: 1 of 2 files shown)


stasadev:
> first the microcopy when trying to use both flags is sort of challenging to comprehend in particular for none developers like me

It comes from the Cobra package, and I cannot override the message from DDEV:

https://github.com/ddev/ddev/blob/5042e2732d2b01efb23acf0c85057867294c0273/vendor/github.com/spf13/cobra/flag_groups.go#L204

I don't understand the message itself either.

One thing I could do is override it right in the `verdor` folder, but that's not a good practice.

> i would rather be gravitating towards `private SSH key`? that would sound more seamless to me.

I don't know what is better:
- [GitHub](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) uses `SSH private key` terminology
- [GitLab](https://docs.gitlab.com/ee/user/ssh.html) uses `private SSH key` terminology

Let's wait for @rfay's opinion.

docs/content/users/usage/commands.md
```diff
@@ -186,23 +186,23 @@ Authentication commands.
 
 Add [SSH key authentication](../usage/cli.md#ssh-into-containers) to the `ddev-ssh-agent` container.
 
+The command can be executed multiple times to add more keys to the same `ssh-agent`.
```
rfay, line 189:
```suggestion
The command can be executed multiple times to add more keys, or you can use the `-f` flag for several individual keys for example `ddev ssh add -f /path/to/first_id_rsa -f /path/to/second_id_rsa`.
```

docs/content/users/usage/commands.md
```diff
@@ -186,23 +186,23 @@ Authentication commands.
...
 ddev auth ssh -f ~/.ssh/id_ed25519 -f ~/.ssh/id_rsa
-# Add your SSH keys from ~/.ssh and ~/custom/path/to/ssh/id_ed25519
-ddev auth ssh -d ~/.ssh -f ~/custom/path/to/ssh/id_ed25519
 ```
 
 Flags:
 
-* `--ssh-key-file`, `-f`: Full path to SSH key file.
-* `--ssh-key-path`, `-d`: Full path to SSH key directory.
+* `--ssh-key-file`, `-f`: Full path to SSH private key file, use the flag multiple times to add more keys.
```
rfay, line 204: Relative path works fine, right?

```suggestion
* `--ssh-key-file`, `-f`: Path to SSH private key file, use the flag multiple times to add more keys.
```

docs/content/users/usage/commands.md
```diff
@@ -186,23 +186,23 @@ Authentication commands.
...
+* `--ssh-key-path`, `-d`: Full path to directory with SSH private key(s), use the flag multiple times to add more directories.
```
rfay, line 205:
```suggestion
* `--ssh-key-path`, `-d`: Path to directory with SSH private key(s), use the flag multiple times to add more directories.
```

rfay (approved):
Looks great, minor comments about pathing.

docs/content/users/usage/commands.md
stasadev: Yes, the relative path works too.

cmd/ddev/cmd/auth-ssh.go
```diff
@@ -185,8 +184,10 @@ func echoDockerCmd(dockerCmd []string) string {
 }
 
 func init() {
-	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyFiles, "ssh-key-file", "f", nil, "full path to SSH key file")
-	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyDirs, "ssh-key-path", "d", nil, "full path to SSH key directory")
+	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyFiles, "ssh-key-file", "f", nil, "full path to SSH private key file, use the flag multiple times to add more keys")
+	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyDirs, "ssh-key-path", "d", nil, "full path to directory with SSH private key(s), use the flag multiple times to add more directories")
```
stasadev, line 188:
```suggestion
	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyFiles, "ssh-key-file", "f", nil, "path to SSH private key file, use the flag multiple times to add more keys")
	AuthSSHCommand.Flags().StringArrayVarP(&sshKeyDirs, "ssh-key-path", "d", nil, "path to directory with SSH private key(s), use the flag multiple times to add more directories")
```

rpkoller:
thank for the clarification @stasadev and i am glad its not only me who doesnt understand that error message :D and if the origin of the message is the cobra package then it is ok, better not to override it in the vendor folder if that is not a recommended practice. all good then. thank you!

PR #6799, opened 2024-12-03
docs: Remove twitter, add bluesky [skip ci]

rfay:
## The Issue

docs: remove twitter, add bluesky

Rendered at https://ddev--6799.org.readthedocs.build/en/6799/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -91,8 +91,7 @@ extra:
     link: https://discord.gg/5wjP76mBJD
   - icon: fontawesome/brands/stack-overflow
     link: https://stackoverflow.com/tags/ddev
-  - icon: fontawesome/brands/twitter
-    link: https://twitter.com/search?q=%23ddev&src=unknown&f=live
+
   generator: false
   analytics:
     provider: plausible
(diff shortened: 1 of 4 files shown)


mkdocs.yml
```diff
@@ -91,8 +91,7 @@ extra:
     link: https://discord.gg/5wjP76mBJD
   - icon: fontawesome/brands/stack-overflow
     link: https://stackoverflow.com/tags/ddev
-  - icon: fontawesome/brands/twitter
-    link: https://twitter.com/search?q=%23ddev&src=unknown&f=live
+
```
bmartinez287, line 94: This might leave a gap on the footer though. As the icon might be missing.

rfay: It's rendered at https://ddev--6799.org.readthedocs.build/en/6799/ and I don't see a gap, but I don't particularly like the result. Not sure if this is an artifact of the PR build.

bmartinez287: Screenshot 2024-12-02 at 8 13 40 PM
I guess that deletes it. I thought we would want Bluesky down there too but it might not be needed.

PR #6801, opened 2024-12-03
fix: improve `ddev debug test`

stasadev:
## The Issue

`RUN timeout 120 apt-get update || true` has too high timeout, people with network troubles may think that `ddev debug test` is stuck.

- Lowers the timeout.
- Adds info about `$JAVA_HOME` for mkcert (so we know if there is something wrong with it).
- Prints curl location (I remember there were some issues with curl installed with homebrew or other place).
- Shows customizations in `.ddev/traefik/config`.

Run

```
ddev debug test
```

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -48,7 +48,7 @@ echo
 ddev debug configyaml | grep -v web_environment
 
 header "existing project customizations"
-grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres .ddev/.env .ddev/.env.* 2>/dev/null | grep -v '\.example$' 2>/dev/null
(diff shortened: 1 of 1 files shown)


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -48,7 +48,7 @@ echo
 ddev debug configyaml | grep -v web_environment
 
 header "existing project customizations"
-grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres .ddev/.env .ddev/.env.* 2>/dev/null | grep -v '\.example$' 2>/dev/null
+grep -r -L "#ddev-generated" .ddev/docker-compose.*.yaml .ddev/php .ddev/mutagen .ddev/apache .ddev/nginx* .ddev/*-build .ddev/mysql .ddev/postgres .ddev/traefik/config .ddev/.env .ddev/.env.* 2>/dev/null | grep -v '\.example$' 2>/dev/null
```
rfay, line 51: Excellent

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -160,6 +160,7 @@ docker ps -a
 if command -v mkcert >/dev/null; then
   header "mkcert information"
   which -a mkcert
+  echo "JAVA_HOME=${JAVA_HOME:-}"
```
rfay, line 163: So glad you put that in there.

rfay (approved):
Looking great, thanks for getting the JAVA_HOME especially. 

If we don't have echo of $CAROOT and WSLENV in there those might be helpful.

Issue #6808, opened 2024-12-04
DDEV v1.24: Fix pre.Dockerfile handling for older PHP version installations


TerrorSquad:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

There is no summary. 

</details>

### Expected Behavior

Installing older PHP versions (7.2, 7.3, 7.4) should work, with the provided `pre.Dockerfile`.

`.ddev/web-build/pre.Dockerfile`

```Dockerfile
COPY certificates/* /usr/local/share/ca-certificates
RUN update-ca-certificates
```

### Actual Behavior

Installing older php versions is impossible because the container cannot fetch the latest package info from `https://packages.sury.org/php/dists/bookwork/InRelease` because of the missing SSL certificates.

The generated `.ddev/.webimageBuild/Dockerfile` should be changed so that the code injected from the `pre.Dockerfile` comes before installing PHP.

The older version 1.23.5 works fine. This issue is most likely the result of the new image trimming.

### Steps To Reproduce

1. Be in an environment that requires additional SSL certificates - e.g. Zscaler
2. Set up a project using older PHP version e.g. 7.2
3. Add a custom `pre.Dockerfile` that installs ZScaler certificate
4. Run `ddev start`

### Anything else?

I am posting this from my personal laptop, but I am experiencing this issue on my company's laptop. On the company's laptop, I can only use the Enterprise Github account (which doesn't allow posting here) and not my own, which is the reason for the missing debug info above.

rfay:
Thanks for studying this carefully. Please update the title to reflect your actual situation, which is pretty "special". 

A PR is welcome.

TerrorSquad:
You're welcome. I'll try to come up with a PR for this issue.

I've changed the title. Hopefully it's clearer now. If not, please suggest a new one.

rfay:
This was fixed by
* https://github.com/ddev/ddev/pull/6809

Thanks!

TerrorSquad:
Hi @rfay 

[...]

PR #6809, opened 2024-12-04
fix: move pre.Dockerfile injection before installing PHP

TerrorSquad:
## The Issue

- #6808 

This issue is only present when in an environment that requires custom SSL certificates.

`pre.Dockerfile` is injected too late, making it impossible to install older PHP versions, even when custom SSL certificates are added via the `pre.Dockerfile`

This PR changes the generated `.ddev/webimageBuild/Dockerfile` by moving the contents of the `pre.Dockerfile` before the commands that install PHP. If we add custom SSL certificates with the `pre.Dockerfile`, then the next command that installs the requested PHP version will work.

## Automated Testing Overview

I didn't add any tests, mainly because of a lack of time.

## Release/Deployment Notes

N/A

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1201,15 +1201,6 @@ ARG DDEV_PHP_VERSION
 ARG DDEV_DATABASE
 RUN (groupadd --gid $gid "$username" || groupadd "$username" || true) && (useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' --uid $uid "$username" || useradd  -l -m -s "/bin/bash" --gid "$username" --comment '' "$username" || useradd  -l -m -s "/bin/bash" --gid "$gid" --comment '' "$username" || useradd -l -m -s "/bin/bash" --comment '' $username )
 `
-	// If our PHP version is not already provided in the ddev-webserver, add it now
-	if strings.Contains(fullpath, "webimageBuild") {
-		if _, ok := nodeps.PreinstalledPHPVersions[app.PHPVersion]; !ok {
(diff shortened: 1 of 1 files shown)


stasadev (approved):
I thought that `pre.Dockerfile` might be dependent on the PHP installed, and I thought that running `install_php_extensions.sh` before `pre.Dockerfile` was justified.

But it turned out the way it is, thank you, @TerrorSquad for the quick fix!

TerrorSquad:
Well, `pre.Dockerfile` might depend on the installed PHP. It depends on what the user wants to do. They could add SSL certificates, but they could also try to run PHP.

Perhaps a more streamlined solution should be implemented.

We could add a new subdirectory `certs` into `web-build` and `db-build` directories. ddev could then internally read from that directory and add the certificates into the resulting docker image, instead of relying on the user to write a `pre.Dockerfile` that adds the SSL certs.

The order of execution would then be as follows:

1. Base dockerfile injection
2. Certificates Injection (if there are files in the {web-build,db-build}/certs directory)
3. Installation of PHP
4. Injecting `pre.Dockerfile`

The question is, should adding custom SSL certs be done by ddev with this `certs` dir, or by the user with the `pre.Dockerfile`.

stasadev:
I think we can introduce something like `top.Dockerfile` or `first.Dockerfile`, that will be added at the top, even earlier:

```dockerfile
ARG BASE_IMAGE="scratch"
FROM $BASE_IMAGE
SHELL ["/bin/bash", "-c"]
```

To give users complete freedom of what they can do, which will also allow multi-stage builds.

PR #6810, opened 2024-12-04
fix: don't copy Dockerfiles to `*imageBuild` folder, add examples to assets, fixes #6757

### LINKED ISSUE #6757: Managing files in `.ddev/web-build` and `.ddev/.webimageBuild`
### Preliminary checklist

```
```
</details>

### Expected Behavior

`.ddev/.webimageBuild` should contain only context files from `.ddev/web-build`, but not files like:
- `.ddev/web-build/*.example`
- `.ddev/web-build/README.txt`
- `.ddev/web-build/*Dockerfile*`

`.ddev/web-build/Dockerfile.example` and `.ddev/web-build/README.txt` should be created on `ddev start`.

### Actual Behavior

Everything from `.ddev/web-build` is copied to `.ddev/.webimageBuild`
We don't have `.ddev/web-build/README.txt`.
`.ddev/web-build/Dockerfile.example` is created only with `ddev config` but not with `ddev start` (need to add it to `dotddev_assets`)

### Steps To Reproduce

```
rm -rf .ddev/web-build && ddev start
ls -la .ddev/web-build # nothing here
```

### Anything else?

The same applies to `.ddev/db-build` and `.ddev/.dbimageBuild`.

stasadev:
## The Issue

- #6757

Creates example Dockerfiles in `.ddev/web-build` and `.ddev/db-build` on `ddev config` and `ddev start`. This lets people understand that they can customize Docker images when they get a project from someone else (i.e. they don't run `ddev config`).

Populates app assets on `ddev config` in addition to `ddev start`.

Adds README.txt to `.ddev/web-build`, `.ddev/db-build` and `.ddev/web-entrypoint.d`.

Skips all Dockerfiles that end on `*.example`, not only `Dockerfile.example`.

Updates tests.

```
$ rm -rf .ddev/web-build && ddev start

# assets should be here
$ ls .ddev/web-build 
Dockerfile.example  README.txt

# and they shouldn't be copied to the build directory
$ ls .ddev/.webimageBuild 
Dockerfile

# create context
$ touch .ddev/web-build/file && mkdir -p .ddev/web-build/folder && ddev start

# context should be here
$ ls .ddev/.webimageBuild 
Dockerfile  file  folder
```

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -101,14 +101,16 @@ These files’ content will be inserted into the constructed Dockerfile for each
 
 For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* everything else:
 
+* `.ddev/web-build/pre.Dockerfile`
 * `.ddev/web-build/pre.Dockerfile.*`
+* `.ddev/db-build/pre.Dockerfile`
 * `.ddev/db-build/pre.Dockerfile.*`
 
 Examine the resultant generated Dockerfile (which you will never edit directly), at `.ddev/.webimageBuild/Dockerfile`. You can force a rebuild with [`ddev debug rebuild`](../usage/commands.md#debug-rebuild).
 
-Examples of possible Dockerfiles are `.ddev/web-build/Dockerfile.example` and `.ddev/db-build/Dockerfile.example`, created in your project when you run [`ddev config`](../usage/commands.md#config).
+Examples of possible Dockerfiles are `.ddev/web-build/Dockerfile.example` and `.ddev/db-build/Dockerfile.example`, created in your project when you run [`ddev start`](../usage/commands.md#start).
 
-You can use the `.ddev/*-build` directory as the Docker “context” directory as well. So for example, if a file named `README.txt` exists in `.ddev/web-build`, you can use `ADD README.txt /` in the Dockerfile.
+You can use the `.ddev/*-build` directory as the Docker “context” directory as well. So for example, if a file named `file.txt` exists in `.ddev/web-build`, you can use `ADD file.txt /` in the Dockerfile.
 
 An example web image `.ddev/web-build/Dockerfile` might be:
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -101,10 +101,10 @@ These files’ content will be inserted into the constructed Dockerfile for each
 
 For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* everything else:
 
-* `.ddev/web-build/pre.Dockerfile`
 * `.ddev/web-build/pre.Dockerfile.*`
-* `.ddev/db-build/pre.Dockerfile`
+* `.ddev/web-build/pre.Dockerfile`
 * `.ddev/db-build/pre.Dockerfile.*`
+* `.ddev/db-build/pre.Dockerfile`
 
 Examine the resultant generated Dockerfile (which you will never edit directly), at `.ddev/.webimageBuild/Dockerfile`. You can force a rebuild with [`ddev debug rebuild`](../usage/commands.md#debug-rebuild).
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -108,7 +108,7 @@ For certain use cases, you might need to add directives very early on the Docker
 
 Examine the resultant generated Dockerfile (which you will never edit directly), at `.ddev/.webimageBuild/Dockerfile`. You can force a rebuild with [`ddev debug rebuild`](../usage/commands.md#debug-rebuild).
 
-Examples of possible Dockerfiles are `.ddev/web-build/Dockerfile.example` and `.ddev/db-build/Dockerfile.example`, created in your project when you run [`ddev start`](../usage/commands.md#start).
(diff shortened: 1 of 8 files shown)


pkg/ddevapp/config.go
```diff
@@ -1334,7 +1315,33 @@ fi`, app.Database.Version, app.GetStartScriptTimeout(), psqlVersion) + "\n\n"
 	// Assets in the web-build directory copied to .webimageBuild so .webimageBuild can be "context"
 	// This actually copies the Dockerfile, but it is then immediately overwritten by WriteImageDockerfile()
 	if userDockerfilePath != "" {
-		err = copy2.Copy(userDockerfilePath, filepath.Dir(fullpath))
+		err = copy2.Copy(userDockerfilePath, filepath.Dir(fullpath), copy2.Options{Skip: func(_ os.FileInfo, src, _ string) (bool, error) {
```
rfay, line 1318: Pretty fancy! Please add a comment above this line saying what it's doing and why, thanks.

Would be be better to turn this into an actual function instead of a closure?

rfay (approved):
Thanks for going back to this. It looks great to me. Thanks for the excellent test coverage.

rfay (changes requested):
Oops, manually testing, I see that no example files or README is being created in .ddev/web-build or db-build

stasadev:
> I see that no example files or README is being created in .ddev/web-build or db-build

Did you run `ddev start` ?
They were created with `ddev config` and now they are created with `ddev start` (like the rest of the assets in the `.ddev` directory).

rfay:
Yes, ran `ddev start` several times.

rfay:
Oh, I see. I had wrong link to ddev build, sorry

pkg/ddevapp/config.go
stasadev: I cannot get rid of closure completely, because I need `userDockerfilePath` inside that closure to detect top-level directory (so I can remove `*.example` files correctly, without touching subdirs).

Is it okay to use something like this?:

```go
err = copy2.Copy(userDockerfilePath, filepath.Dir(fullpath), copy2.Options{
	Skip: func(_ os.FileInfo, src, _ string) (bool, error) {
		return shouldSkipDockerfileOnCopy(src, userDockerfilePath)
	},
})
```

rfay: Yes, works for me. Just want future me to be able to easily understand it.  I would have accepted as it was, just looking for clearer presentation.

pkg/ddevapp/config.go
```diff
@@ -1524,18 +1515,6 @@ func PrepDdevDirectory(app *DdevApp) error {
...
-		"web-entrypoint.d",
-		"xhprof",
-	}
-	for _, subdir := range dirs {
-		err = os.MkdirAll(filepath.Join(dir, subdir), 0755)
-		if err != nil {
-			return err
-		}
-	}
-
```
stasadev, line 1538: I remember we had problems with this before, and tests failed here (again), when there was no `.ddev/web-build` after `ddev config`.

I decided to add `PopulateExamplesCommandsHomeadditions` on `ddev config` (see above), and added `web-entrypoint.d/README.txt`, so that this directory would be created as well.

pkg/ddevapp/config.go
```diff
@@ -257,36 +257,14 @@ func (app *DdevApp) WriteConfig() error {
...
-## You can copy this Dockerfile.example to Dockerfile to add configuration
-## or packages or anything else to your webimage
-## These additions will be appended last to DDEV's own Dockerfile
-## See examples here https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/#adding-extra-dockerfiles-for-webimage-and-dbimage
-# RUN echo "Built on $(date)" > /build-date.txt
-`)
-
-	err = WriteImageDockerfile(app.GetConfigPath("web-build")+"/Dockerfile.example", contents)
+	// The .ddev directory may still need to be populated, especially in tests
+	err = PopulateExamplesCommandsHomeadditions(appcopy.Name)
```
stasadev, line 261: I don't know why we didn't have this function here, I find it quite valuable to populate the app assets on `ddev config` so that people can change them before `ddev start`.

pkg/ddevapp/config.go
stasadev: Alternatively, I could add `web-build` and `db-build` here to the `dirs` slice - but why would I do that if we can add all the assets to the `ddev config`.

docs/content/users/extend/customizing-images.md
```diff
@@ -101,14 +101,16 @@ These files’ content will be inserted into the constructed Dockerfile for each
 
 For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* everything else:
 
+* `.ddev/web-build/pre.Dockerfile`
 * `.ddev/web-build/pre.Dockerfile.*`
+* `.ddev/db-build/pre.Dockerfile`
 * `.ddev/db-build/pre.Dockerfile.*`
```
rfay, line 107: nit: Let's put the `.*` first, best practice. People are better off to name these things. 

```suggestion
* `.ddev/web-build/pre.Dockerfile.*`
* `.ddev/web-build/pre.Dockerfile`
* `.ddev/db-build/pre.Dockerfile.*`
* `.ddev/db-build/pre.Dockerfile`
```

rfay (approved):
Great work, thanks. Minor nit in preference for `Dockerfile.*`

I guess we may want to consider waiting to pull until after v1.24.1, if we're doing that soon.

stasadev:
> I guess we may want to consider waiting to pull until after v1.24.1, if we're doing that soon.

Agreed.

rfay:
Converted to draft just because I don't want to pull by accident.

rfay:
Rebased for review reasons.

PR #6813, opened 2024-12-05
test: Stop testing for NFS capability

rfay:
## The Issue

* Our tests don't need to look for NFS being enabled any more, we don't test NFS. 
* Docs don't need to explain how to fiddle with Windows NFS. It's just a bad idea.

### DIFF
--- a/docs/content/users/install/performance.md
+++ b/docs/content/users/install/performance.md
@@ -236,7 +236,7 @@ Mutagen is enabled by default on Mac and traditional Windows, and it can be disa
     !!!warning "NFS is deprecated"
 
         NFS is deprecated and no longer recommended. It can be complex and unreliable. This feature will be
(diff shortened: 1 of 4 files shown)


docs/content/users/install/performance.md
```diff
@@ -235,8 +235,8 @@ Mutagen is enabled by default on Mac and traditional Windows, and it can be disa
 
     !!!warning "NFS is deprecated"
 
-        NFS is deprecated and no longer recommended. This feature may be
-        removed in a future release.
+        NFS is deprecated and no longer recommended. It can be complex and unreliable. This feature will be
+        removed DDEV v1.25.0.
```
stasadev, line 239:
```suggestion
        removed in DDEV v1.25.0.
```

stasadev (approved):

PR #6815, opened 2024-12-06
fix: remove python leftovers, for #6733

stasadev:
## The Issue

- #6733

Removes leftovers that were added here, but not removed:

- #4768

The `ddev-webserver` image size is 3.5 MB smaller.

### DIFF
--- a/.spellcheckwordlist.txt
+++ b/.spellcheckwordlist.txt
@@ -338,7 +338,6 @@ filesystem
 filesystems
 findstr
 fish
-flask
 for
 format
 fpath
@@ -537,7 +536,6 @@ provider
 ps
 psql
 public
-python
 Querious
 quickstart
 quickstarts
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -104,7 +104,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -o Dpkg::Options::="--for
     less \
     libldap-common \
     libpcre3 \
-    libpq-dev \
     nano \
     ncurses-bin \
     netcat-traditional \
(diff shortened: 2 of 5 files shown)


.spellcheckwordlist.txt
```diff
@@ -537,7 +536,6 @@ provider
 ps
 psql
 public
-python
```
rfay, line 540: I think people depend in this probably? Of course they could add it in webimage_extra_packages, but it seems like a normal thing to have.

containers/ddev-webserver/Dockerfile
```diff
@@ -104,7 +104,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -o Dpkg::Options::="--for
     less \
     libldap-common \
     libpcre3 \
-    libpq-dev \
```
rfay, line 107: This is for interacting with postgres. I think it's OK to remove, but not sure what it has to do with python leftovers.

pkg/ddevapp/config.go
```diff
@@ -1423,15 +1423,6 @@ func DiscoverDefaultDocroot(app *DdevApp) string {
 			}
 		}
 	}
-	dir, err := fileutil.FindFilenameInDirectory(app.AppRoot, []string{"manage.py"})
```
rfay, line 1426: Thanks for catching this!

rfay (approved):
Thank you! Minor suggestions/questions.

containers/ddev-webserver/Dockerfile
stasadev: `libpq-dev` was initially added in the Python PR https://github.com/ddev/ddev/pull/4768/files#diff-131a1297bdf1b5e28b280c2e8e473b47e9f959fe10efd73f421f2e0e058a9364R73

That's why I thought it was necessary for Python only?

rfay: Works for me.

PR #6819, opened 2024-12-06
docs: use shortlink for discord access

rfay:
## The Issue

* https://github.com/ddev/ddev.com/pull/288

We now have a shortlink for discord, so we don't have to use the odd one. Update that.

I also discovered a number of probably inappropriate references to ddev-contrib, so trimmed some of those.

I imagine this will run tests, but don't want to short-circuit that.


stasadev (approved):
Looks good to me.

One more place to update https://github.com/ddev/remote-config

PR #6821, opened 2024-12-09
fix: correct panic on ddev describe on gitpod

rfay:
## The Issue

In doing some debugging on gitpod for
* https://github.com/ddev/ddev/pull/6589

I noted a panic in `ddev describe`. It's a result of trying to convert nil to string. 

Check before converting.

`ddev describe` on gitpod


stasadev (approved):
Looks good to me.

PR #6822, opened 2024-12-09
docs: update magento2, pimcore, sulu quickstarts

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/6784#pullrequestreview-2486662032
- https://gist.github.com/rpkoller/d0fa272b81b7705dd3f56c35e01c8f8a

@rpkoller pointed out some problems in our quickstarts.

Magento 2:

I didn't notice any issue, the quickstart works for my in the current state.
But I decided to update the admin URL to the persistent value: `/admin_ddev`

Pimcore:

They require `amqp` extension:

- https://github.com/pimcore/skeleton/issues/186

Sulu:

They changed the filename to `website.xml`:

- https://github.com/sulu/skeleton/pull/248

Additionally, I automated setting `<name>` and `<key>` with `xmlstarlet`.

Test these quickstarts:

- https://ddev--6822.org.readthedocs.build/en/6822/users/quickstart/#magento
- https://ddev--6822.org.readthedocs.build/en/6822/users/quickstart/#pimcore
- https://ddev--6822.org.readthedocs.build/en/6822/users/quickstart/#sulu

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -454,15 +454,15 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
             ```
 
     ```bash
-    mkdir my-magento2-site && cd my-magento2-site
+    export MAGENTO_HOSTNAME=my-magento2-site
+    mkdir ${MAGENTO_HOSTNAME} && cd ${MAGENTO_HOSTNAME}
     ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
     ddev add-on get ddev/ddev-elasticsearch
     ddev start
     ddev composer create --repository https://repo.magento.com/ magento/project-community-edition
     rm -f app/etc/env.php
 
-    # Change the base-url below to your project's URL
-    ddev magento setup:install --base-url="https://my-magento2-site.ddev.site/" \
+    ddev magento setup:install --base-url="https://${MAGENTO_HOSTNAME}.ddev.site/" \
         --cleanup-database --db-host=db --db-name=db --db-user=db --db-password=db \
         --elasticsearch-host=elasticsearch --search-engine=elasticsearch7 --elasticsearch-port=9200 \
         --admin-firstname=Magento --admin-lastname=User --admin-email=user@example.com \
@@ -638,8 +638,8 @@ ddev config --project-type=php --docroot=public --upload-dirs=uploads --database
 ddev start
 ddev composer create sulu/skeleton
 # Create your default webspace configuration `config/webspaces/my-sulu-site.xml`
-# from `config/webspaces/website.xml`. The command below will adjust the values
-# for `<name>` and `<key>`, so that they are matching your project:
+# from `config/webspaces/website.xml`. Alternately, the command below will adjust the values
+# for `<name>` and `<key>` to match the project you have set up:
 # <name>My Sulu Site</name>
 # <key>my-sulu-site</key>
 ddev exec 'name="$(echo "${DDEV_PROJECT}" | sed "s/\b\(.\)/\U\1/g" | tr "-" " ")"; key="${DDEV_PROJECT}"; xmlstarlet ed -u "//_:webspace/_:name" -v "${name}" -u "//_:webspace/_:key" -v "${key}" -u "//_:portals/_:portal/_:name" -v "${name}" -u "//_:portals/_:portal/_:key" -v "${key}" config/webspaces/website.xml > config/webspaces/${DDEV_PROJECT}.xml && rm -f config/webspaces/website.xml'
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -473,7 +473,7 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     ddev config --disable-settings-management=false
     # Change the backend frontname URL to /admin_ddev
     ddev magento setup:config:set --backend-frontname="admin_ddev" --no-interaction
-    # Log into your account using `admin` and `Password123`
+    # Login using `admin` user and `Password123` password
     ddev launch /admin_ddev
     ```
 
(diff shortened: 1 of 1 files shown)


rpkoller (commented):
**magento2:** unfortunately i am still running into `404 page not found` at the end :( 
**pimcore:** works perfectly now, thank you! one nitpick would it be possible to remove the second `ddev start` in that quickstart in line 546. the project is already started so the start is unnecessary or is it required to make certain config changes active? if so i would at least rename it to `ddev restart`? 
**Sulu:** uh wow that one got even more convenient setting up. no need of manual altering the xml file, niiiice. and yep the quickstart works as expected as well.

rpkoller:
and one detail about the magento2 quickstart to note, somehow i have to enter the credentials twice during the setup process. the save location is in home and therefore it is not stored by ddev. out of the scope for this issue i suppose, and i also dont need it since i dont use magento, but i wonder if it would make sense for those regularly using it if it would be possible to store those credentials more persistent?

stasadev:
> one nitpick would it be possible to remove the second `ddev start` in that quickstart in line 546. the project is already started so the start is unnecessary or is it required to make certain config changes active? if so i would at least rename it to `ddev restart`? 

Thanks, it really should be `ddev restart`, needed for `web_extra_daemons`.

stasadev:
> one detail about the magento2 quickstart to note, somehow i have to enter the credentials twice during the setup process

I didn't enter anything because I used `Script to store Adobe/Magento Composer credentials (click me)`

image

That small script creates `~/.ddev/homeadditions/.composer/auth.json` file with:

```json
{
    "http-basic": {
        "repo.magento.com": {
            "username": "username_here",
            "password": "password_here"
        }
    }
}
```

rfay:
If you don't store the credentials it's really painful.

But... will we make the homedir nonvolatiile someday? Why don't we yet?

rpkoller:
argh i'Ve missed that script link on the docs. yep definitely more convenient that way. but on the downside, i've retested with the credentials saved and available but i still run into the `404 page not found` error in the end on ` ddev launch /admin_ddev`

rpkoller:
> If you don't store the credentials it's really painful.
> 
> But... will we make the homedir nonvolatiile someday? Why don't we yet?

making the homedir nonvolatile might also be beneficial in the context of working inside the container with ultimikes approach?

docs/content/users/quickstart.md
```diff
@@ -633,38 +634,27 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
-    <!-- See: http://docs.sulu.io/en/latest/book/webspaces.html how to configure your webspace-->
-
-    <name>My Sulu CMS</name>
-    <key>my-sulu-cms</key>
+# Create your default webspace configuration `config/webspaces/my-sulu-site.xml`
+# from `config/webspaces/website.xml`. The command below will adjust the values
+# for `<name>` and `<key>`, so that they are matching your project:
+# <name>My Sulu Site</name>
+# <key>my-sulu-site</key>
+ddev exec 'name="$(echo "${DDEV_PROJECT}" | sed "s/\b\(.\)/\U\1/g" | tr "-" " ")"; key="${DDEV_PROJECT}"; xmlstarlet ed -u "//_:webspace/_:name" -v "${name}" -u "//_:webspace/_:key" -v "${key}" -u "//_:portals/_:portal/_:name" -v "${name}" -u "//_:portals/_:portal/_:key" -v "${key}" config/webspaces/website.xml > config/webspaces/${DDEV_PROJECT}.xml && rm -f config/webspaces/website.xml'
```
rfay, line 645: (I think this is OK with the modest update above, saying that they can edit themselves or use this magic command. Still a bit uncomfortable with it.) 

I understand that you're trying to simplify the technique here, but I'm not actually convinced that people will understand or be able to recreate what we're doing here. I may prefer the old technique, which has clear docs. We want people to be able to clearly understand every step, and perhaps with this approach we're not doing that.

docs/content/users/quickstart.md
```diff
@@ -463,17 +463,18 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
 
     # Change the base-url below to your project's URL
```
rfay, line 464: Sorry to request other things here, but let's separate the domain name into an environment variable so they can mostly copy/paste. If we say `export MAGENTO_HOSTNAME= my-magento2-site` and then use `mkdir ${MAGENTO_HOSTNAME} && cd ${MAGENTO_HOSTNAME}` then `ddev magento setup:install with the ${MAGENTO_HOSTNAME}.ddev.site` will work without them having to stop and fiddle with it.

docs/content/users/quickstart.md
```diff
@@ -463,17 +463,18 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
     ddev magento setup:install --base-url="https://my-magento2-site.ddev.site/" \
```
rfay, line 465:
```suggestion
    export MAGENTO_HOSTNAME=my-magento2-site
    mkdir ${MAGENTO_HOSTNAME} && cd ${MAGENTO_HOSTNAME}
    ddev magento setup:install --base-url="https://${MAGENTO_HOSTNAME}.ddev.site/" \
```

docs/content/users/quickstart.md
```diff
@@ -633,38 +634,27 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
...
-<webspace xmlns="http://schemas.sulu.io/webspace/webspace"
-          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-          xsi:schemaLocation="http://schemas.sulu.io/webspace/webspace http://schemas.sulu.io/webspace/webspace-1.1.xsd">
-    <!-- See: http://docs.sulu.io/en/latest/book/webspaces.html how to configure your webspace-->
-
-    <name>My Sulu CMS</name>
-    <key>my-sulu-cms</key>
+# Create your default webspace configuration `config/webspaces/my-sulu-site.xml`
+# from `config/webspaces/website.xml`. The command below will adjust the values
+# for `<name>` and `<key>`, so that they are matching your project:
```
rfay, line 642:
```suggestion
# Create your default webspace configuration `config/webspaces/my-sulu-site.xml`
# from `config/webspaces/website.xml`. Alternately, the command below will adjust the values
# for `<name>` and `<key>` to match the project you have set up:
```

rfay (approved):
These all worked for me. I have reservations about the xmlstarlet usage, but suggested workaround. Sorry to suggest modest change to the magento2 one, but I often fail with changing the hostname there.

rpkoller:
> These all worked for me. I have reservations about the xmlstarlet usage, but suggested workaround. Sorry to suggest modest change to the magento2 one, but I often fail with changing the hostname there.

that was exactly my problem that lead into the 404s, i forgot to adjust `my-magento2-site` to my  bespoke folder name in: 

```
ddev magento setup:install --base-url="https://my-magento2-site.ddev.site/" \
    --cleanup-database --db-host=db --db-name=db --db-user=db --db-password=db \
    --elasticsearch-host=elasticsearch --search-engine=elasticsearch7 --elasticsearch-port=9200 \
    --admin-firstname=Magento --admin-lastname=User --admin-email=user@example.com \
    --admin-user=admin --admin-password=Password123 --language=en_US
```
after adjusting it during setup to the folder name everything works and i dont run into the 404 error anymore.

rfay:
One thing we could probably do better in the long term is to report to people where the 404 comes from. In @rpkoller 's case, I think it was coming from the router, because there was no back-end for it to hit. It's confusing what's going on when that happens.

docs/content/users/quickstart.md
```diff
@@ -633,38 +634,27 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like
 
 ```bash
 mkdir my-sulu-site && cd my-sulu-site
-ddev config --project-type=php --docroot=public --upload-dirs=uploads --database=mysql:8.0
+ddev config --project-type=php --docroot=public --upload-dirs=uploads --database=mysql:8.0 --webimage-extra-packages="xmlstarlet"
```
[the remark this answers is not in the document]
stasadev: Thanks, I didn't know it was abandoned in Debian. (I only knew that it was maintained in Arch.)

docs/content/users/quickstart.md
stasadev: I'll replace it with simpler `ddev exec sed ...` approach.

PR #6823, opened 2024-12-09
fix: unset DDEV_NONINTERACTIVE in Gitpod

stasadev:
## The Issue

In Gitpod:
```
$ ddev phpmyadmin
Nothing has been changed.
```

Because `DDEV_NONINTERACTIVE` is set.

Unsets it.


rfay (approved):
Thanks!

PR #6824, opened 2024-12-09
fix: show correct warning for Dockerfiles in `web-build` and `db-build`

stasadev:
## The Issue

It says `web-entrypoint.d` (but it should say `web-build`), and it doesn't show customizations for `db-build`:

```
$ ddev start
Starting d11... 
Using custom web-entrypoint.d configuration: [
        /home/stas/code/ddev/d11/.ddev/web-build/Dockerfile.test
        /home/stas/code/ddev/d11/.ddev/web-build/pre.Dockerfile
] 
Custom configuration is updated on restart.
If you don't see your custom configuration taking effect, run 'ddev restart'.
```

Fixes it:

```
$ ddev start
Starting d11... 
Using custom web-build configuration: [
        /home/stas/code/ddev/d11/.ddev/web-build/Dockerfile.test
        /home/stas/code/ddev/d11/.ddev/web-build/pre.Dockerfile
] 
Using custom db-build configuration: [
        /home/stas/code/ddev/d11/.ddev/db-build/Dockerfile.xample
        /home/stas/code/ddev/d11/.ddev/db-build/pre.Dockerfile.asd
] 
Custom configuration is updated on restart.
If you don't see your custom configuration taking effect, run 'ddev restart'.
```

Add custom Dockerfile to the project in `.ddev/web-build`, and `.ddev/db-build` and run `ddev start`.

## Release/Deployment Notes

This is part of the

- #6810

that we can safely pull in the next point release.

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -711,17 +711,22 @@ func (app *DdevApp) CheckCustomConfig() {
 		}
 	}
 
-	customDockerPath := filepath.Join(ddevDir, "web-build")
-	if _, err := os.Stat(customDockerPath); err == nil {
-		dockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "*Dockerfile*"))
-		util.CheckErr(err)
-		dockerFiles = slices.DeleteFunc(dockerFiles, func(s string) bool {
-			return strings.HasSuffix(s, ".example")
-		})
-		if len(dockerFiles) > 0 {
-			printableFiles, _ := util.ArrayToReadableOutput(dockerFiles)
-			util.Warning("Using custom web-entrypoint.d configuration: %v", printableFiles)
-			customConfig = true
+	for _, buildType := range []string{"web-build", "db-build"} {
+		customDockerPath := filepath.Join(ddevDir, buildType)
+		if _, err := os.Stat(customDockerPath); err == nil {
+			dockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "Dockerfile*"))
+			util.CheckErr(err)
+			preDockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "pre.Dockerfile*"))
+			util.CheckErr(err)
+			dockerFiles = append(dockerFiles, preDockerFiles...)
+			dockerFiles = slices.DeleteFunc(dockerFiles, func(s string) bool {
+				return strings.HasSuffix(s, ".example")
+			})
+			if len(dockerFiles) > 0 {
+				printableFiles, _ := util.ArrayToReadableOutput(dockerFiles)
+				util.Warning("Using custom %s configuration: %v", buildType, printableFiles)
+				customConfig = true
+			}
 		}
 	}
 


pkg/ddevapp/config.go
```diff
@@ -711,17 +711,22 @@ func (app *DdevApp) CheckCustomConfig() {
...
-		})
-		if len(dockerFiles) > 0 {
-			printableFiles, _ := util.ArrayToReadableOutput(dockerFiles)
-			util.Warning("Using custom web-entrypoint.d configuration: %v", printableFiles)
-			customConfig = true
+	for _, buildType := range []string{"web-build", "db-build"} {
+		customDockerPath := filepath.Join(ddevDir, buildType)
+		if _, err := os.Stat(customDockerPath); err == nil {
+			dockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "Dockerfile*"))
+			util.CheckErr(err)
```
rfay, line 718: I don't really like this usage any more, it seems to hide behavior. Even though the go usage `if err != nil` is ugly and verbose, it's very clear what it does. Would prefer to use that.

pkg/ddevapp/config.go
```diff
@@ -711,17 +711,22 @@ func (app *DdevApp) CheckCustomConfig() {
...
-			return strings.HasSuffix(s, ".example")
-		})
-		if len(dockerFiles) > 0 {
-			printableFiles, _ := util.ArrayToReadableOutput(dockerFiles)
-			util.Warning("Using custom web-entrypoint.d configuration: %v", printableFiles)
-			customConfig = true
+	for _, buildType := range []string{"web-build", "db-build"} {
+		customDockerPath := filepath.Join(ddevDir, buildType)
+		if _, err := os.Stat(customDockerPath); err == nil {
+			dockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "Dockerfile*"))
```
rfay, line 717: It wasn't clear to me why you couldn't glob the Dockerfile* and pre.Dockerfile* in one statement here.

pkg/ddevapp/config.go
```diff
@@ -711,17 +711,22 @@ func (app *DdevApp) CheckCustomConfig() {
...
+			util.CheckErr(err)
+			preDockerFiles, err := filepath.Glob(filepath.Join(customDockerPath, "pre.Dockerfile*"))
+			util.CheckErr(err)
+			dockerFiles = append(dockerFiles, preDockerFiles...)
+			dockerFiles = slices.DeleteFunc(dockerFiles, func(s string) bool {
```
rfay, line 722: Nice and elegant :)

rfay (approved):
Manual testing works fine, thanks. Minor nit about not liking the util.CheckErr().

pkg/ddevapp/config.go
stasadev: Because if you name some file `foo.Dockerfile`, it will not be added to `.ddev/.webimageBuild/Dockerfile`, but will be shown here, which is wrong.

pkg/ddevapp/config.go
stasadev: Yes, it's not obvious, but it's used everywhere in this function, so I didn't change it.

rfay: OK

PR #6828, opened 2024-12-10
docs: use `drupal` in Drupal 11 Quickstart

gitressa:
Since you get the latest Drupal with `drupal`, we could use that?

The same way for `ddev composer create drupal/recommended-project`.

Also, since `ddev launch $(ddev drush uli)` logs you in anyway, maybe we can remove the user and password setting bit from `site:install`?

## The Issue

- #<issue number>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -187,10 +187,10 @@ The legacy type `drupal` will be interpreted as the latest stable version of Dru
 
     ```bash
     mkdir my-drupal11-site && cd my-drupal11-site
-    ddev config --project-type=drupal11 --docroot=web
-    ddev composer create drupal/recommended-project:^11
+    ddev config --project-type=drupal --docroot=web
+    ddev composer create drupal/recommended-project
     ddev composer require drush/drush
-    ddev drush site:install --account-name=admin --account-pass=admin -y
+    ddev drush site:install -y
     ddev launch
     # or automatically log in with
     ddev launch $(ddev drush uli)
@@ -203,7 +203,7 @@ The legacy type `drupal` will be interpreted as the latest stable version of Dru
     ddev config --project-type=drupal10 --docroot=web
     ddev composer create drupal/recommended-project:^10
     ddev composer require drush/drush
-    ddev drush site:install --account-name=admin --account-pass=admin -y
+    ddev drush site:install -y
     ddev launch
     # or automatically log in with
     ddev launch $(ddev drush uli)


rpkoller:
i would strongly vote against reintroducing `drupal` as the project type in a quickstart. it would be ok and probably reasonable to add a commented out line before line 190 that educates the user that instead of `drupal11` one could also go with `drupal`as the alias for the latest major version. but i would keep things explicit. same for the composer create command.. this quickstart is about drupal 11 therefore i would also make the composer create step explicit. otherwise it might be more confusing than helpful. and the thought about using the `--account-name` and `--account-pass` flag was in one part educational and on the other hand it saves the user a copy and paste of the generated password if the person uses `ddev launch` instead of `ddev launch $(ddev drush uli)` .

rfay:
We want to use `drupal11`, thanks.  And the explicit use of admin/admin is deliberate.

You argued loudly against the `drupal` type, now you have to live with `drupal11` :)

gitressa:
Thanks for the feedback @rkoller and @rfay. In my opinion, the main focus of a Quickstart as a concept should be to get the user started as fast as possible, with the least cruft -- not be educational (though it can be).

But of course, we may all look at this differently.

About including a tip for Drupal 11, it already says this in the intro, which is great:

> The legacy type `drupal` will be interpreted as the latest stable version of Drupal, so in 2024, ddev config --project-type=drupal` will configure a Drupal 11 project.

rfay:
Sorry, `drupal11` will be the preferred approach.

gitressa:
All right, though I don't hear strong arguments against using just `drupal` to get the latest Drupal ... I think documentation should be As Lean As possible: [Make Composer create-project As Lean As Possible (ALAP)](https://www.drupal.org/project/drupal/issues/3104445).

Also, Drupal documentation does not specify version for the latest release:
- https://www.drupal.org/docs/develop/using-composer/manage-dependencies#s-create-a-project
- https://www.drupal.org/docs/develop/using-composer/starting-a-site-using-drupal-composer-project-templates#s-how-do-i-set-it-up

But I accept the decision -- it's not a case of black or white, and I do see upsides to explicitly including `11`, though more benefits by leaving it out :)

rfay:
It's always a question whether to put the ^11 in there.

https://ddev.readthedocs.io/en/stable/users/quickstart/#drupal-drupal-11

has it. But our docs get carefully checked by you every time there's a change, and I know perfectly well there will be a 12 when needed, and @rpkoller already did a drupal cms, please review it, 

* https://github.com/ddev/ddev/pull/6829

PR #6831, opened 2024-12-10
test: Add configuration verification tests for mariadb and mysql

rfay:
## The Issue

* https://github.com/ddev/ddev/pull/6589

has the potential to be disruptive and cause unintended configuration changes. We don't want that.

* Add tests beforehand that specify the configuration we expect so we're not caught by surprise


stasadev (approved):
Looks good to me.

PR #6834, opened 2024-12-11
fix: add CI to instrumentation tags

IndraGunawan:
## The Issue

I read this page https://ddev.com/blog/stats-on-ddev-usage-nov-2024/ and found

> ... `ddev drush` is excluded here because it was used 515,000 times in the last 7 days. I guess some people must be using it in an automated situation. If you’re using any of these in CI, please turn off reporting if it happens to be on, `ddev config global --instrumentation-opt-in=false` or `export DDEV_NO_INSTRUMENTATION=true`.

~instead of asking user to set `DDEV_NO_INSTRUMENTATION` env var, let's make DDEV automatically detect and set no instrumentation if it is run on a CI pipeline~

here are some popular CI tools that have `CI=true` env var 

- Github Actions https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
- Gitlab CI https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
- Buildkite https://buildkite.com/docs/pipelines/configure/environment-variables#CI
- Bitbucket https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/
- CircleCI https://circleci.com/docs/variables/#built-in-environment-variables

* Build using the development AmplitudeAPIKey
* Do some things and then `ddev debug instrumentation flush`
* Verify that `CI` comes through as false in https://app.amplitude.com/analytics/ddev/live-events using "DDEV - Development"
* export CI=true
* Do over, verify CI=true

## Release/Deployment Notes

- [ ] Followup in the future after this goes into release and has time for people to have it, we can update our reports to test whether CI is true.


stasadev (commented):
Thank you, @IndraGunawan, especially for the links to popular CI tools.

@rfay, I noticed that we don't mention `DDEV_NO_INSTRUMENTATION` / `CI` in our documentation. Should we?

rfay:
Thank you so much for the good proactive thinking! I guess now you've got me hooked, and I'd rather include CI in the Amplitude instrumentation, so we know to exclude from the regular reports, but we can actually see what's going on with CI!

IndraGunawan:
I pushed a new CI tag

Would excluding ci from analytics reduce amplitude cost? But if we get more benefits from them then better to include it.

but whatever the decision i will revert the unused commit later

tyler36:
What a great idea! 

I was just think about this the other day. 

Thanks for providing the links to various CI platforms.

rfay:
> Would excluding ci from analytics reduce amplitude cost? But if we get more benefits from them then better to include it.

Actually, we haven't ever ended up paying Amplitude, so hopefully it will stay that way. But even when charging, they charge by active users, not by amount of reporting.

rfay:
This is great and exciting, but adding to amplitude requires more, at least [I think it does](https://ddev.readthedocs.io/en/stable/developers/building-contributing/#instrumentation). Thanks for getting this started!

stasadev (commented):
I looked into the project files and I don't see any usage for `nodeps.InstrumentationTags`.

@rfay, is `pkg/ddevapp/instrumentation.go` a leftover file from a previous instrumentation logic, that has no use?

rfay:
I think you're right. `instrumentation.go` and `instrumentation_test.go` are orphaned code (from segment), and this effort needs to take place using the techniques in https://ddev.readthedocs.io/en/stable/developers/building-contributing/#instrumentation

rfay:
Updated using Amplitude techniques. Ready for verification.

stasadev (approved):
Manual testing worked fine, I see the `CI` variable in the Amplitude logs.

Looks good to me!

rfay:
I was studying this in the charts and I guess we should have put CI as a user property, not a project property :(

PR #6836, opened 2024-12-11
docs: how to add unsupported PHP versions without `ddev config`, fixes #6835

### LINKED ISSUE #6835: Unsupported PHP (5.6, 7.0, 7.1, 7.2, 7.3, 7.4, 8.0) is not available in the container (v1.24.0+) by default
The following PHP versions are not available by default in DDEV v1.24.0+

```bash
ddev exec php5.6 -v
ddev exec php7.0 -v
ddev exec php7.1 -v
ddev exec php7.2 -v
ddev exec php7.3 -v
ddev exec php7.4 -v
ddev exec php8.0 -v
```

To install one of these versions, you can switch to it in the configuration with:

```bash
ddev config --php-version=7.4
```

However, if you want to use multiple PHP versions simultaneously while keeping a newer version as the default, you can use the technique described below.

---

If your project requires multiple PHP versions (e.g., you're using PHP 8.3 but also need PHP 7.4 for certain scripts), and you don’t want to switch completely to PHP 7.4 using `ddev config --php-version=7.4`, you can install it using the `pre.Dockerfile.*` technique from the previous section.

Create a `.ddev/web-build/pre.Dockerfile.php7.4` file with the following content:

```dockerfile
RUN /usr/local/bin/install_php_extensions.sh "php7.4" "${TARGETARCH}"
```

After restarting the project, you can use PHP 7.4 with the command `ddev exec php7.4 -v`.

stasadev:
## The Issue

- #6835

Adds instruction to the docs.

https://ddev--6836.org.readthedocs.build/en/6836/users/extend/customizing-images/#adding-eol-versions-of-php

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -84,7 +84,7 @@ The web image ships by default with a small number of locales, which work for mo
 
 If you need other locales, you can install all of them by adding `locales-all` to your `webimage_extra_packages`. For example, in `.ddev/config.yaml`:
 
-```
+```yaml
 webimage_extra_packages: [locales-all]
 ```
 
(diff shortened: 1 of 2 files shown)


rfay (approved):
I'm not 100% sure this detail belongs in docs, but I guess I'm ok with it. It should be a pretty obscure workaround.

But since you made a great way to install things and it's so simple... yay!

docs/content/users/extend/customizing-images.md
```diff
@@ -178,6 +178,18 @@ An example of using `$TARGETARCH` would be:
 RUN curl --fail -JL -s -o /usr/local/bin/mkcert "https://dl.filippo.io/mkcert/latest?for=linux/${TARGETARCH}"
 ```
 
+## Adding unsupported PHP versions
```
tyler36, line 181:
```suggestion
## Adding older PHP versions
```

"Unsupported" describes it from a DDEV point of view. 
I feel developer using DDEV would naturally search for "older PHP versions". Thoughts?

stasadev: I added more keywords: EOL, older, unsupported, unmaintained.

PR #6842, opened 2024-12-13
feat: Add pager to database output

tyler36:
## The Issue

Database CLI commands (`ddev mysql`, `ddev psql`) are not configured to use a pager. This results is text wrapping on smaller terminal windows.

MySQL: 
image

Postgres:
image

This PR configures the pager option for databases, improving the readability of output.

- A PAGER environmental variable is defined
- It uses `less`  (see [man less](https://man7.org/linux/man-pages/man1/less.1.html) for `-SFX` switch descriptions.
- If output wraps, the pager is invoked. Press <kbd>right-arrow</kbd> to page right, <kbd>left-arrow</kbd> to page left.
- If output does not wrap, no pager is used (aka current behavior).

> [!IMPORTANT] 
> This PR changes the default behavior of the database CLI, and therefore may be "controversial"
> This PR using an environmental variable, developers can override the value through the various methods.

I manually testd on the following:

- MariaDB: `10.11`
- Postgres: `16 `
- MySQL: `8.0`

### MariaDB 10.11

image
Press <kbd>right-arrow</kbd> to page right

image

Example of wide terminal that does not invoke pager; 
image

### Postgres:16 

> If the environment variable PAGER is set, the output is piped to the specified program. Otherwise a platform-dependent default (such as more) is used.
https://www.postgresql.org/docs/7.4/app-psql.html

Postgres: paged
image

### MySQL 8.0

MySQL requires the `--pager` switch passed in the command. When no value is provided, it fallsback to the environmental variable defined. See https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-using-pager.html

MySQL: paged
image

Setup project. For existing project skip to "pager test".

1. Create a Laravel project following quickstart guide.
2. Migrate database.

```shell
ddev artisan migrate:fresh
```

3. Populate database with random data

```shell
$ ddev artisan tinker
User::factory(5)->create()
exit
```

### Test pager

5. Start database CLI via `ddev mysql` or `ddev psql`
6. Enter SQL command to view content. Eg. `select * from Users;`

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -212,6 +212,7 @@ services:
     - MYSQL_PWD=db
     - NODE_EXTRA_CA_CERTS=/mnt/ddev-global-cache/mkcert/rootCA.pem
     - npm_config_cache=/mnt/ddev-global-cache/npm
+    - PAGER=less -SFX
     - PGDATABASE=db
     - PGHOST=db
     - PGPASSWORD=db
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -71,6 +71,7 @@ services:
       - POSTGRES_PASSWORD=db
       - POSTGRES_USER=db
       - POSTGRES_DB=db
+      - PAGER=less -SFX
       - TZ={{ .Timezone }}
       - USER={{ .Username }}
     command: ${DDEV_DB_CONTAINER_COMMAND}
--- a/pkg/ddevapp/global_dotddev_assets/commands/db/mysql
+++ b/pkg/ddevapp/global_dotddev_assets/commands/db/mysql
@@ -8,4 +8,4 @@
 ## DBTypes: mysql,mariadb
 ## ExecRaw: true
 
-mysql -udb -pdb "$@"
+mysql -udb -pdb --pager "$@"


stasadev (commented):
It's a nice addition!

I'm thinking about the `PAGER` env, if it's appropriate to use it for the entire `db` container.

What if we add this pager to the config files (not saying if it should be added exactly there):
- mysql https://serverfault.com/a/955986
- psql https://stackoverflow.com/a/14474666/8097891

And if we talk about `mysql`, `mariadb`, `psql`, it would be good to have the same config in the `web` container.

tyler36:
This PR uses the environmental variable because it is the "lowest" level. Developers can easily add an environment to the config file to override. Or update the database config file, which overrides the ENV. Or update the command file which I believe is the hightest level.

So we prefer to go with the config file?

tyler36:
> And if we talk about mysql, mariadb, psql, it would be good to have the same config in the web container.

I'm not sure I understand this. Is the database configured in both the DB and WEB containers?

GuySartorelli:
Is `PAGER` too generic a name? Should it be `DDEV_PAGER` or similar so it's clearly related to DDEV?

Are there instructions to the user to press right/left arrows to scroll? If not, should that be added? I don't think I'd Intuit it personally

tyler36:
> Is PAGER too generic a name?

`PAGER` is the value used by the MySQL, MariaDB, and Postgres databases; its what they are hard-coded to look for. 

So I guess thats another vote towards setting config instead of ENV?

stasadev:
> I'm not sure I understand this. Is the database configured in both the DB and WEB containers?

@tyler36, I meant that we also have database client binaries in the web container. If we add it to `db`, should we also have it in `web`?

> Is `PAGER` too generic a name? Should it be `DDEV_PAGER` or similar so it's clearly related to DDEV?

`PAGER` is a common variable that is used by many, many Linux programs. I wonder if it's okay to have it set for the whole container. Could this break/change behavior for some other program?

rfay:
$PAGER is a really standard approach, and we do the same thing with $MYSQL_HISTFILE, for example. These are local variables, intended to change the behavior of the tools that use them. We don't need to use a DDEV_ variant. And there's nothing wrong with doing it as an environment variable.

stasadev:
I think it's okay to add `PAGER` globally.

What about adding `PAGER` to `ddev-webserver` (so database clients and other programs can also use it)?

Note: it would added to `git` commands inside the container. (There is also `GIT_PAGER` if someone wants to override it.)

rfay (approved):
This is great, thanks! Tests nicely. It can go in as it is or you can add for webserver. This could also have been done in the respective Dockerfiles I imagine.

tyler36:
`PAGER` was added to web container envionment. 
Ready to on approval.

rfay (approved):
This works and is fine. I do wish that `mysql` would just respect the PAGER without having to be run with `mysql --pager` but oh, well.

tyler36:
> I do wish that mysql would just respect the PAGER without having to be run with mysql --pager but oh, well.

Was disappointed too.

stasadev (approved):
Thank you!

rfay:
I'm finding this inconvenient. I prefer `SHOW TABLES;` to just show me all the tables. How are you two feeling about it? I'd be more comfortable without it, and of course it's trivial for any user to enable with a .env or otherwise.

stasadev:
I agree that it's annoying for big databases and `SHOW TABLES;`, but without pager you can't read any `SELECT * FROM ...`

I suggest several workarounds:

1. Use `ddev mysql -e "SHOW TABLES;"`
2. Long press `d` and it will scroll to the end and press `q`. 
3. Set `nopager` before running the command:
	```
	mysql> nopager
	PAGER set to stdout
	mysql> SHOW TABLES;
	```

rfay:
My previous strategy was always just to scroll up for `SELECT * FROM ...` :)

tyler36:
I mostly use `select *` statements which are difficult for me to read without a pager.

Feel free to revert this if you feel that its creates friction for your workflows.

stasadev:
> I'm finding this inconvenient. I prefer `SHOW TABLES;` to just show me all the tables. How are you two feeling about it? I'd be more comfortable without it, and of course it's trivial for any user to enable with a .env or otherwise.

Opened:

- #6900

PR #6846, opened 2024-12-19
docs: add CiviCRM Standalone to quickstart.md

gitressa:
## The Issue

Historically, [CiviCRM](https://civicrm.org/) required a CMS such as Drupal or Wordpress, but [CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running it without a CMS.

Adds CiviCRM Standalone to the [CMS Quickstarts](https://ddev.readthedocs.io/en/stable/users/quickstart) page.

https://ddev--6846.org.readthedocs.build/en/6846/users/quickstart/#civicrm-standalone

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -74,53 +74,27 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
 
 [CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
 
-=== "macOS and Linux"
-
-    ```bash
-    curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
-    tar -xzf civicrm-standalone.tar.gz
-    cd civicrm-standalone
-    ddev config --project-type=php
-    ddev start
-    curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
-    chmod +x bin/cv
-    # You can now install CiviCRM manually in your browser using `ddev launch`
-    # and selecting `db` for the server and `db` for database/username/password
-    # or do the same automatically using the command below:
-    # The parameter `-m loadGenerated=1` includes sample data
-    ddev exec cv core:install \
-        --cms-base-url='$DDEV_PRIMARY_URL' \
-        --db=mysql://db:db@db/db \
-        -m loadGenerated=1 \
-        -m extras.adminUser=admin \
-        -m extras.adminPass=admin \
-        -m extras.adminEmail=admin@example.com
-    ddev launch
-    ```
-
-=== "Windows"
-
-    ```bash
-    mkdir my-civicrm-site && cd my-civicrm-site
-    ddev config --project-type=php --upload-dirs=public/media
-    echo "RUN curl -L --fail -o /tmp/civicrm-standalone.tar.gz -sSL https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz" > .ddev/web-build/Dockerfile.civicrm-standalone
-    echo "RUN curl -L --fail -o /usr/local/bin/cv -sSL https://download.civicrm.org/cv/cv.phar && chmod ugo+wx /usr/local/bin/cv" > .ddev/web-build/Dockerfile.cv
-    ddev start
-    ddev exec tar --strip-components=1 -xzf /tmp/civicrm-standalone.tar.gz
-    # You can now install CiviCRM manually in your browser using `ddev launch`
-    # and selecting `db:3306` for the server and `db` for database/username/password
-    # or do the same automatically using the command below:
-    # The parameter `-m loadGenerated=1` includes sample data
-    ddev exec cv core:install \
-        --cms-base-url='$DDEV_PRIMARY_URL' \
-        --db=mysql://db:db@db/db \
-        -m loadGenerated=1 \
-        -m extras.adminUser=admin \
-        -m extras.adminPass=admin \
-        -m extras.adminEmail=admin@example.com
-    # Login using `admin` user and `admin` password
-    ddev launch
-    ```
+```bash
+mkdir my-civicrm-site && cd my-civicrm-site
+ddev config --project-type=php
+ddev start
+ddev exec curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
+ddev exec tar --strip-components=1 -xzf civicrm-standalone.tar.gz
+ddev exec curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
+ddev exec chmod +x bin/cv
+# You can now install CiviCRM manually in your browser using `ddev launch`
+# and selecting `db` for the server and `db` for database/username/password
+# or do the same automatically using the command below:
+# The parameter `-m loadGenerated=1` includes sample data
+ddev exec cv core:install \
+    --cms-base-url='$DDEV_PRIMARY_URL' \
+    --db=mysql://db:db@db/db \
+    -m loadGenerated=1 \
+    -m extras.adminUser=admin \
+    -m extras.adminPass=admin \
+    -m extras.adminEmail=admin@example.com
+ddev launch
+```
 
 Visit [Install CiviCRM (Standalone)](https://docs.civicrm.org/installation/en/latest/standalone) for more installation details.
 
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -76,7 +76,7 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
 
 ```bash
 mkdir my-civicrm-site && cd my-civicrm-site
-ddev config --project-type=php
+ddev config --project-type=php --upload-dirs=public/media
 echo "RUN curl -L --fail -o /tmp/civicrm-standalone.tar.gz -sSL https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz" > .ddev/web-build/Dockerfile.civicrm-standalone
 echo "RUN curl -L --fail -o /usr/local/bin/cv -sSL https://download.civicrm.org/cv/cv.phar && chmod ugo+wx /usr/local/bin/cv" > .ddev/web-build/Dockerfile.cv
 ddev start
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -79,7 +79,7 @@ mkdir my-civicrm-site && cd my-civicrm-site
 ddev config --project-type=php
 ddev start
 ddev exec curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
-ddev exec tar --strip-components=1 -xzf civicrm-standalone.tar.gz
+ddev exec tar --strip-components=1 -xzf civicrm-standalone.tar.gz && ddev exec rm civicrm-standalone.tar.gz
 ddev exec curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
 ddev exec chmod +x bin/cv
 # You can now install CiviCRM manually in your browser using `ddev launch`
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -70,6 +70,35 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
     ddev launch
     ```
 
+## CiviCRM (Standalone)
+
+[CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
+
+```bash
+curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
+tar -xzf civicrm-standalone.tar.gz
+cd civicrm-standalone
+ddev config --project-type=php
+ddev start
+curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
+chmod +x bin/cv
+# The parameter `-m loadGenerated=1` includes sample data
+ddev exec cv core:install \
+    --cms-base-url="https://civicrm-standalone.ddev.site" \
+    --db=mysql://db:db@db/db \
+    -m loadGenerated=1 \
+    -m extras.adminUser=admin \
+    -m extras.adminPass=admin \
+    -m extras.adminEmail=admin@example.com
+ddev launch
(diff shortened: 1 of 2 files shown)


tyler36:
Thank you.

Was able to install and setup a CiviCRM site.

A sugestion:

- Make a note that this is for MySQL install. Other DBs users should need to change the `--db=mysql://db:db@db/db` line

gitressa:
Thanks for a fast review @tyler36.

Adding a database tip could be considered, feel free to make a review, and add a suggested solution. Though, I don't see any other Quick Start doing that ...

stasadev (approved):
I refactored it a bit, because I'm not sure that `tar` is available everywhere.

> Make a note that this is for MySQL install. Other DBs users should need to change the `--db=mysql://db:db@db/db` line

I don't think we need that, because the docs say that it's MariaDB compatible.

gitressa:
Thanks @stasadev, but I think the Windows method looks a bit over engineered ... what if we separate the methods -- one for MacOS and Linux, and one for Windows?

stasadev:
@gitressa, we don't need new sections, because I *already* made it work for all platforms.

> I think the Windows method looks a bit over engineered

Using `.ddev/web-build` is pretty standard technique, especially for DDEV.

P.S. Note that you install the `cv` binary to a non-standard location at `/var/www/html/bin/cv`. CiviCRM suggests to use `/usr/local/bin/cv` https://docs.civicrm.org/installation/en/latest/general/cli-cv/#download-cv, as I did.

gitressa:
I don't doubt that it works on all platforms, I just prefer to keep Quick Starts as simple as possible, that's my only aim.

And about `PATH`, note their choice of word, "for example". So as long as it's in the DDEV `PATH`, I think that's fine.

gitressa:
Actually, how about we let DDEV untar it? Couldn't something like this work on all platforms??

```
mkdir my-civicrm-site && cd my-civicrm-site
ddev start
ddev exec curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
ddev exec tar --strip-components=1 -xzf civicrm-standalone.tar.gz
ddev exec curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
ddev exec chmod +x bin/cv
[...]
```

stasadev:
The install should be done to `my-civicrm-site` to align with all other quickstarts, your setup creates a subfolder - we don't want that, and `tar.gz` archive is not removed afterwards, also it should include upload-dirs (don't forget about macOS and Mutagen).

stasadev:
> Actually, how about we let DDEV untar it?

That's what I did, didn't I? 🙂

Yes, we can do the same in container without using `.ddev/web-build`.

gitressa:
Ok, I'll update the MR, and let's discuss :)

stasadev:
Okay, install it to `/tmp/civicrm-standalone.tar.gz`, so you don't have to clean it afterwards.

gitressa:
That doesn't work:

```
$ ddev exec tar --strip-components=1 -xzf /tmp/civicrm-standalone.tar.gz
tar (child): /tmp/civicrm-standalone.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
Failed to execute command tar --strip-components=1 -xzf /tmp/civicrm-standalone.tar.gz: exit status 2
```

docs/content/users/quickstart.md
```diff
@@ -70,6 +70,34 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
     ddev launch
     ```
 
+## CiviCRM (Standalone)
+
+[CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
+
+```bash
+mkdir my-civicrm-site && cd my-civicrm-site
+ddev config --project-type=php
```
stasadev, line 79: From https://docs.civicrm.org/installation/en/latest/standalone/

> `public/media/`	Public files uploaded to your site, like images that appear on a contribution page or in a mass email

```suggestion
ddev config --project-type=php --upload-dirs=public/media
```

docs/content/users/quickstart.md
```diff
@@ -70,6 +70,34 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
...
+ddev start
+ddev exec curl -L https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz > civicrm-standalone.tar.gz
+ddev exec tar --strip-components=1 -xzf civicrm-standalone.tar.gz && ddev exec rm civicrm-standalone.tar.gz
+ddev exec curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv
+ddev exec chmod +x bin/cv
```
stasadev, line 84:
```suggestion
ddev exec "curl -LsS https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz -o /tmp/civicrm-standalone.tar.gz"
ddev exec "tar --strip-components=1 -xzf /tmp/civicrm-standalone.tar.gz"
ddev exec "curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv && chmod +x bin/cv"
```

stasadev (changes requested):

docs/content/users/quickstart.md
gitressa: I never had any use for that, and for a Quick Start it's overkill IMHO.

docs/content/users/quickstart.md
gitressa: Thanks! These are great improvements, and work very well.

gitressa:
Thanks for the review, the `curl` and `tar` commands are much better now.

About your `--upload-dirs` comment:

>  also it should include upload-dirs (don't forget about macOS and Mutagen).

Is this a new thing? I found a single line, but it didn't explain why it's needed, or which OSes need it ...

> <code>--upload-dirs</code>: Sets the project’s upload directories, the destination directories of the import-files command, or <code>--upload-dirs=""</code> to remove previously configured values.

From: https://ddev.readthedocs.io/en/stable/users/usage/commands/#config

gitressa:
I see, thanks for clarifying. I found #5888 which had more info about this. But as I wrote, hardly anyone just starting out with CiviCRM will upload files.

So in this case, we would be adding that extra parameter `--upload-dirs=public/media` only to avoid a warning about something which will probably not be used anyway. Doesn't that seen backward?

Alternatively, the `php` recipe could get `public/media` added as a directory to exclude from Mutagen indexing? It's CivCRM's  "Public site files which should be accessible over the web", from the section "More detail: CiviCRM Standalone folder structure" on [Install CiviCRM (Standalone)](https://docs.civicrm.org/installation/en/latest/standalone/).

rfay:
Remember that quickstarts are not just to get people started, but to give instruction in best practices. That's why it makes sense to point out upload_dirs.

gitressa:
But why configure something, which is unlikely to be used? And if it will be used -- very, very few files will ever be added ... In other words, we are configuring for a scenario which will probably never cause any issues anyway.

A pragmatic approach would be to add the path to `php` recipe, like I suggested.

But of course, it's your project, and you should add it, if you feel strongly about it.

rfay:
So CiviCRM projects aren't ever likely to have significant collections of non-PHP assets like images? And nobody will ever use `ddev import-files` (which is also controlled by `upload_dirs`)

Note that if we just wanted to turn off the warning, we'd use `disable_upload_dirs_warning`

gitressa:
I can't say with certainty. But as I wrote, if you strongly feel that it should be added, you can do it.

How about simply adding the path to the `php` project, like I suggested a few times?

tyler36:
> Adding a database tip could be considered, feel free to make a review, and add a suggested solution. Though, I don't see any other Quick Start doing that ...

@gitressa 
It was a nitpik, for sure. DDEV does a great job managing the database settings for other frameworks which is why they don't include database settings.
If memory serves, this is one of the few guides that excplictly sets the database type.

rfay:
I'm hoping @vinugawade will create an add-on for CiviCRM, as discussed in 
* https://github.com/ddev/ddev/pull/6856#issuecomment-2561930976

That will be a great way to package up CiviCRM support, and might be added to these docs.

gitressa:
That's great news @rfay, thanks for sharing! In the mean time, until it's ready, this Quick Start could be added? I have added `--upload-dirs=public/media`, so there shouldn't be anything blocking it. When #6856 is ready, we can update the CiviCRM Quick Start.

docs/content/users/quickstart.md
gitressa: Let's add it, to be safe.

gitressa:
The test error looks unrelated ...

> (node:701) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.

rfay:
That's just a deprecation warning, the error is https://github.com/ddev/ddev/actions/runs/12505423285/job/34888766027?pr=6846

gitressa:
It looks like it's a dead link in another file:

> 
> FILE: docs/content/developers/maintainers.md
> [✖] https://readthedocs.org/dashboard/ddev/users/
> 
> 22 links checked.
> 
> ERROR: 1 dead links found!
> [✖] https://readthedocs.org/dashboard/ddev/users/ → Status: 0
> (node:151) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
> (Use `node --trace-deprecation ...` to show where the warning was created)

docs/content/users/quickstart.md
```diff
@@ -70,6 +70,33 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
...
+
+[CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
+
+```bash
+mkdir my-civicrm-site && cd my-civicrm-site
+ddev config --project-type=php --upload-dirs=public/media
+ddev start
+ddev exec "curl -LsS https://download.civicrm.org/latest/civicrm-STABLE-standalone.tar.gz -o /tmp/civicrm-standalone.tar.gz"
+ddev exec "tar --strip-components=1 -xzf /tmp/civicrm-standalone.tar.gz"
+ddev exec "curl -LsS https://download.civicrm.org/cv/cv.phar --create-dirs -o bin/cv && chmod +x bin/cv"
```
stasadev, line 83: Should we add more CLI tools from #6856? Ignore this change if you don't like it.
(I added `--no-scripts` because it didn't work otherwise.)

```suggestion
ddev composer require civicrm/cli-tools --no-scripts
```

docs/content/users/quickstart.md
```diff
@@ -70,6 +70,33 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
     ddev launch
     ```
 
+## CiviCRM (Standalone)
+
+[CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
+
+```bash
+mkdir my-civicrm-site && cd my-civicrm-site
+ddev config --project-type=php --upload-dirs=public/media
```
stasadev, line 79: For `ddev composer` to work correctly if users want to use Composer:

```suggestion
ddev config --project-type=php --composer-root=core --upload-dirs=public/media
```

stasadev (changes requested):

docs/content/users/quickstart.md
gitressa: Thanks @stasadev, that's a great suggestion, having Composer take care of the placement, permissions, etc. is a more elegant solution.

stasadev (approved):
Looks good to me!

stasadev:
Rebased to see how the new linkspector and mkdocs python dependency update (the last two commits on master) work here.

docs/content/users/quickstart.md
```diff
@@ -70,6 +70,33 @@ Please note that you will need to change the PHP version to 7.4 to be able to wo
     ddev launch
     ```
 
+## CiviCRM (Standalone)
+
+[CiviCRM Standalone](https://civicrm.org/blog/ufundo/next-steps-civicrm-standalone) allows running [CiviCRM](https://civicrm.org/) without a CMS.
```
rfay, line 75: I think this link should be to https://docs.civicrm.org/installation/en/latest/standalone/

Oddly, those docs don't seem to be ready at all, with no download link and no composer instructions. A composer install would be more in line with most of the other quickstarts.

```suggestion
[CiviCRM Standalone](https://docs.civicrm.org/installation/en/latest/standalone/) allows running [CiviCRM](https://civicrm.org/) without a CMS.
```

rfay (changes requested):
This is a bit of an unusual mix of tarball and composer.

I see this complaint during the installation of the cli-tools:
`Composer could not detect the root package (civicrm/civicrm-core) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version`

Is that supposed to happen? The tools are properly installed, and in the right path though, core/vendor/bin

I assume that it's normal for composer_root to be `core`, right?

This works for me, but I kind of wonder if it's premature.

docs/content/users/quickstart.md
gitressa: I linked to the blog post, to offer people some background. There's a link to the docs at the bottom.

gitressa:
@rfay: I also get the warning, but it works fine ... It was supposed to be fixed:

> It turns out that [Add a warning message when Composer is not able to guess the root package version #11858](https://github.com/composer/composer/pull/11858) was too broad, causing many false positives as we have seen. It will be fixed when [Only show warning about default version when not "project" type #11885](https://github.com/composer/composer/pull/11885) in the next release of Composer.

From https://www.drupal.org/forum/support/composer/2024-03-13/root-cause-of-could-not-detect-the-root-package#comment-15528494

The reason tar.gz is used, is because that's the only format available ... there is no Composer package yet for standalone CiviCRM.

docs/content/users/quickstart.md
rfay: I didn't find it in the blog and still don't find it after re-reviewing. Please update with links to docs and you can keep the blog link if you think it's useful. It's a pretty rambling piece and makes the project look very, very young :)

gitressa: It's right there at the bottom of the Quick Start :)

> Visit [Install CiviCRM (Standalone)](https://docs.civicrm.org/installation/en/latest/standalone) for more installation details.

- https://github.com/ddev/ddev/pull/6846/files#diff-cdacb5ab99042792ddb67ee7d62ac1d36c4699992949005880517b4955e10f7eR98

rfay: OK, I thought you were saying it was at the bottom of that blog. Please move the link to the top section, and reference the blog if you'd like. People don't read anything that far down on the page.

gitressa: All right, it's at the top now.

rfay (approved):
Thanks!

PR #6849, opened 2024-12-22
build: bump go-pretty and x/crypto, fixes #6844

rfay:
## The Issue

I wanted to bump go-pretty because I saw a new release that might help us with wrapping in cells. 
Then
* https://github.com/ddev/ddev/pull/6844

came along, so might as well do both.


stasadev (approved):
Thank you!

PR #6850, opened 2024-12-22
docs: add Laravel Installer quickstart

stasadev:
## The Issue

Laravel has its own installer, but it's distributed only with Composer.

This installer allows you to select a [starter kit](https://laravel.com/docs/11.x/starter-kits) with additional options when you create the app:

```
$ laravel new
   _                               _
  | |                             | |
  | |     __ _ _ __ __ ___   _____| |
  | |    / _` | '__/ _` \ \ / / _ \ |
  | |___| (_| | | | (_| |\ V /  __/ |
  |______\__,_|_|  \__,_| \_/ \___|_|


 ┌ Would you like to install a starter kit? ────────────────────┐
 │   ○ No starter kit                                           │
 │   ○ Laravel Breeze                                           │
 │ › ● Laravel Jetstream                                        │
 └──────────────────────────────────────────────────────────────┘

 ┌ Which Jetstream stack would you like to install? ────────────┐
 │   ○ Livewire                                                 │
 │ › ● Vue with Inertia                                         │
 └──────────────────────────────────────────────────────────────┘

┌ Would you like any optional features? ───────────────────────┐
 │   ◻ API support                                              │
 │   ◼ Dark mode                                                │
 │ › ◼ Email verification                                       │
 │   ◻ Team support                                             │
 │   ◼ Inertia SSR                                              │
 └──────────────────────────────────────────────────────────────┘

 ┌ Which testing framework do you prefer? ──────────────────────┐
 │ › ● Pest                                                     │
 │   ○ PHPUnit                                                  │
 └──────────────────────────────────────────────────────────────┘
```

Without the Laravel installer, all of this would need to be done manually, and I believe most people prefer having a starter kit from the beginning.

Adds a quickstart for this.

I chose to add the Laravel Installer temporarily via the Dockerfile, since it’s not needed once the app is created.

I don't believe we need to package this installer inside the `ddev-webserver` Docker image, as it’s more than just one file.

It also serves as a good example for users on how to install a global package with Composer.

https://ddev--6850.org.readthedocs.build/en/6850/users/quickstart/#laravel-laravel-installer

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -421,7 +421,7 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     ddev exec 'rsync -rltgopD temp/ ./ && rm -rf temp'
 
     # Remove the Laravel installer and the .env file
-    ddev exec 'rm -rf .ddev/web-build/Dockerfile.laravel .env'
+    rm -rf .ddev/web-build/Dockerfile.laravel .env
 
     # Restart the project and execute the post-install actions
     ddev restart
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -403,7 +403,7 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     mkdir my-laravel-site && cd my-laravel-site
     ddev config --project-type=laravel --docroot=public
 
-    # Temporarily add the Laravel installer
+    # Temporarily add the Laravel installer as /usr/local/bin/laravel in the web container
     echo 'ARG COMPOSER_HOME=/usr/local/composer
     RUN composer global require laravel/installer
     RUN ln -s $COMPOSER_HOME/vendor/bin/laravel /usr/local/bin/laravel
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -409,9 +409,11 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     RUN ln -s $COMPOSER_HOME/vendor/bin/laravel /usr/local/bin/laravel
     ' > .ddev/web-build/Dockerfile.laravel
 
-    # Start the project and select a starter kit of your choice
-    # The database is temporarily set to SQLite and will be switched to MariaDB
+    # Start the project
     ddev start
+
+    # Select a starter kit of your choice.
+    # The database is temporarily set to SQLite and will be switched to MariaDB
     ddev exec laravel new temp --database=sqlite
 
     # Move the installed files
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -397,6 +397,35 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     ddev launch
     ```
 
+=== "Laravel Installer"
+
+    ```bash
+    mkdir my-laravel-site && cd my-laravel-site
+    ddev config --project-type=laravel --docroot=public
+
+    # Temporarily add the Laravel installer
+    echo 'ARG COMPOSER_HOME=/usr/local/composer
+    RUN composer global require laravel/installer
+    RUN ln -s $COMPOSER_HOME/vendor/bin/laravel /usr/local/bin/laravel
+    ' > .ddev/web-build/Dockerfile.laravel
+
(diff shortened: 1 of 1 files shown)


docs/content/users/quickstart.md
```diff
@@ -397,6 +397,35 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    # Temporarily add the Laravel installer
+    echo 'ARG COMPOSER_HOME=/usr/local/composer
+    RUN composer global require laravel/installer
+    RUN ln -s $COMPOSER_HOME/vendor/bin/laravel /usr/local/bin/laravel
+    ' > .ddev/web-build/Dockerfile.laravel
+
+    # Start the project and select a starter kit of your choice
+    # The database is temporarily set to SQLite and will be switched to MariaDB
+    ddev start
+    ddev exec laravel new temp --database=sqlite
```
tyler36, line 415:
```suggestion
    # Start the project
    ddev start
    
    # Select a starter kit of your choice. 
    # The database is temporarily set to SQLite and will be switched to MariaDB
    ddev exec laravel new temp --database=sqlite
```

I feel like seperating these makes it clearer between what DDEV and Laravel setup. 
Strangely, it feels OK at the end though so feel free to ignore.

tyler36 (approved):
This is great.

Easy to follow instructions and a good example of using a Dockerfile.

tyler36:
I had been thinking about a PR to include `cpx` in the base image. It's basically like `npx` but for composer.
Then we could run `ddev cpx laravel/installer new` (assuming we had a helper command`)

https://laravel-news.com/cpx

tyler36:
> At this point, I don't see much value in adding cpx. Maybe we can reconsider it once it matures.

Yes, 100% agree. But I'll excited about the possibilities for "modern" PHP projects!

docs/content/users/quickstart.md
```diff
@@ -397,6 +397,37 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     ddev launch
     ```
 
+=== "Laravel Installer"
+
+    ```bash
+    mkdir my-laravel-site && cd my-laravel-site
+    ddev config --project-type=laravel --docroot=public
+
+    # Temporarily add the Laravel installer
```
rfay, line 406: Since this is instructional, this may be better as a HEREDOC or something. I was going line by line and stumbled here. 

```suggestion
    # Temporarily add the Laravel installer as /usr/local/bin/laravel in the web container
```

docs/content/users/quickstart.md
```diff
@@ -397,6 +397,37 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    ' > .ddev/web-build/Dockerfile.laravel
+
+    # Start the project
+    ddev start
+
+    # Select a starter kit of your choice.
+    # The database is temporarily set to SQLite and will be switched to MariaDB
+    ddev exec laravel new temp --database=sqlite
+
+    # Move the installed files
```
rfay, line 419: This needs a better, slightly longer explanation, since people may not understand either that the files are currently in a temp dir or what rsync is.

docs/content/users/quickstart.md
```diff
@@ -397,6 +397,37 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    ddev exec 'rsync -rltgopD temp/ ./ && rm -rf temp'
+
+    # Remove the Laravel installer and the .env file
+    ddev exec 'rm -rf .ddev/web-build/Dockerfile.laravel .env'
```
rfay, line 423: Why not do this on the host? (not that it matters) Expecting `rm` not to be available? `rm` even works in PowerShell, but not Cmd.

rfay (changes requested):
This is great, although it's as complex "quickstart". 

It can be improved by adding just a bit of text at the top to explain why one would do this, and then a few comments on the complex things that get done so that people can understand the "why".

docs/content/users/quickstart.md
stasadev: This is a copy-paste from the Symfony quickstart, I'll add the explanation.

docs/content/users/quickstart.md
stasadev: Okay, let's use the host.

rfay (approved):
Thanks, sorry for the nitpicks!

PR #6852, opened 2024-12-23
feat: disable `ddev config` in parent folders of a project, but allow it in subfolders, fixes #6783, fixes #5493

### LINKED ISSUE #5493: Prevent execution of `ddev config` in the wrong directory

### LINKED ISSUE #6783: Consider adding the ability to prevent the project creation in specific directories

I have my projects in a single directory `~/Sites` and i have aliases to change to different levels of parent directories by( .., ..., ....).so it could easily happen i dont end up in the root of my project but in the root of my Sites directory (by typing one period too much) and by accident i could then run `Ddev config` inside the Sites directory. Solvable with `ddev delete --omit-snapshot`  and then manually removing the `.ddev` directory, cuz it is easily noticeable since running the config step takes much longer than usual. but still it would be good if it could be prevented and avoided in the first place. 

### Describe your solution

Simply define an array in the global_config.yaml. @rfay suggested calling it `disallowed_project_directories` and maybe it should already contain a few common things like `$HOME` according to randy? 

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #6783
- #5493 

Automatically disable `ddev config` in:
- all project parent directories,
- home and its parent directories,
- global config directory `~/.ddev` and its nested directories,
- `ddev/ddev` repository root directory. (I myself accidentally ran `ddev config` here a few times.)

Allow to run `ddev config` in the project's subdirectory.

TODO:

- [x] Check if we can allow `ddev config` in the project subdirectory.
- [x] Write tests.

Check that you cannot run `ddev config` in any parent of any existing project:
```
mkdir -p ~/workspace/directory/projects/my-project
cd ~/workspace/directory/projects/my-project && ddev config --auto

cd ..
ddev config
Could not create a new config: 'ddev config' is not recommended in the /home/stas/workspace/directory/projects directory because a project exists in the subdirectory /home/stas/workspace/directory/projects/my-project
To use 'ddev config' here, run 'ddev stop --unlist' for all projects in subdirectories of this directory first

cd ..
ddev config
Could not create a new config: 'ddev config' is not recommended in the /home/stas/workspace/directory directory because a project exists in the subdirectory /home/stas/workspace/directory/projects/my-project
To use 'ddev config' here, run 'ddev stop --unlist' for all projects in subdirectories of this directory first

cd ..
ddev config
Could not create a new config: 'ddev config' is not recommended in the /home/stas/workspace directory because a project exists in the subdirectory /home/stas/workspace/directory/projects/my-project
To use 'ddev config' here, run 'ddev stop --unlist' for all projects in subdirectories of this directory first
```

You cannot run it in your home or parent directories of your home:
```
cd ~
ddev config
Could not create a new config: 'ddev config' is not recommended in your home directory (/home/stas)

cd /home
ddev config
Could not create a new config: 'ddev config' is not recommended in the parent directories of your home directory (/home)

cd /
ddev config
Could not create a new config: 'ddev config' is not recommended in the parent directories of your home directory (/)
```

You cannot run it in your global config directory and its nested directories:
```
cd ~/.ddev
ddev config
Could not create a new config: 'ddev config' is not recommended in your global config directory (/home/stas/.ddev)

cd ~/.ddev/commands
ddev config
Could not create a new config: 'ddev config' is not recommended in your global config directory (/home/stas/.ddev/commands)
```

You cannot run it in the directory where you're developing DDEV itself:
```
cd ~/workspace/ddev
ddev config
Could not create a new config: 'ddev config' is not allowed in the directory used for DDEV development (/home/stas/workspace/ddev)
```

Allow `ddev config` in the project subdirectory:

```
mkdir -p ~/workspace/directory/projects/my-project
cd ~/workspace/directory/projects/my-project && ddev config --auto

cd .ddev

# before
ddev config --php-version=8.2
it usually does not make sense to `ddev config` in a subdirectory of an existing project. Is it possible you wanted to `ddev config` in parent directory /home/stas/workspace/directory/projects/my-project?

# after
ddev config --php-version=8.2
You are reconfiguring the project at /home/stas/workspace/directory/projects/my-project.
```

## Release/Deployment Notes

This can break the user's workflow if they have a DDEV project inside another DDEV project, which I don't think is a good practice.

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -333,7 +333,7 @@ func getConfigApp(_ string) (*ddevapp.DdevApp, error) {
 	// Check for an existing config in a parent dir
 	otherRoot, _ := ddevapp.CheckForConf(appRoot)
 	if otherRoot != "" && otherRoot != appRoot {
-		return nil, fmt.Errorf("it usually does not make sense to `ddev config` in a subdirectory of an existing project. Is it possible you wanted to `ddev config` in parent directory %s?", otherRoot)
+		appRoot = otherRoot
 	}
 	app, err := ddevapp.NewApp(appRoot, false)
 	if err != nil {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -149,9 +150,9 @@ func handleConfigRun(cmd *cobra.Command, args []string) {
 		util.Failed(err.Error())
 	}
 
-	homeDir, _ := os.UserHomeDir()
-	if app.AppRoot == filepath.Dir(globalconfig.GetGlobalDdevDir()) || app.AppRoot == homeDir {
-		util.Failed("Please do not use `ddev config` in your home directory")
+	err = ddevapp.IsAllowedProjectLocation(app)
+	if err != nil {
+		util.Failed("Unable to run `ddev config`: %v", err)
 	}
 
 	err = app.CheckExistingAppInApproot()
@@ -196,6 +197,16 @@ func handleConfigRun(cmd *cobra.Command, args []string) {
 	err = app.ProcessHooks("post-config")
 	if err != nil {
 		util.Failed("Failed to process hook 'post-config'")
+	}
+
+	// Exclude project's parent directory from accidental 'ddev config'
+	projectParentDir := filepath.Dir(app.AppRoot)
+	if !slices.Contains(globalconfig.DdevGlobalConfig.NoConfigHere, projectParentDir) {
+		globalconfig.DdevGlobalConfig.NoConfigHere = append(globalconfig.DdevGlobalConfig.NoConfigHere, projectParentDir)
+		err = globalconfig.WriteGlobalConfig(globalconfig.DdevGlobalConfig)
+		if err != nil {
+			util.Failed("Failed to update global config: %v", err)
+		}
 	}
 
 	util.Success("Configuration complete. You may now run 'ddev start'.")
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -145,7 +145,7 @@ var ConfigCommand = &cobra.Command{
 func handleConfigRun(cmd *cobra.Command, args []string) {
 	app := getConfigApp(providerName)
 
-	err := ddevapp.HasRecommendedLocation(app)
+	err := ddevapp.HasAllowedLocation(app)
 	if err != nil {
 		util.Failed("Unable to run `ddev config`: %v", err)
 	}
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -334,6 +334,10 @@ func getConfigApp(_ string) (*ddevapp.DdevApp, error) {
 	otherRoot, _ := ddevapp.CheckForConf(appRoot)
 	if otherRoot != "" && otherRoot != appRoot {
 		appRoot = otherRoot
+		err = os.Chdir(appRoot)
+		if err != nil {
+			util.Failed("Unable to chdir to %v: %v", appRoot, err)
+		}
 	}
 	app, err := ddevapp.NewApp(appRoot, false)
 	if err != nil {
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -4,14 +4,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"slices"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/ddev/ddev/pkg/config/types"
 	"github.com/ddev/ddev/pkg/ddevapp"
-	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/output"
 	"github.com/ddev/ddev/pkg/util"
@@ -197,16 +195,6 @@ func handleConfigRun(cmd *cobra.Command, args []string) {
 	err = app.ProcessHooks("post-config")
 	if err != nil {
 		util.Failed("Failed to process hook 'post-config'")
-	}
-
-	// Exclude project's parent directory from accidental 'ddev config'
-	projectParentDir := filepath.Dir(app.AppRoot)
-	if !slices.Contains(globalconfig.DdevGlobalConfig.NoConfigHere, projectParentDir) {
-		globalconfig.DdevGlobalConfig.NoConfigHere = append(globalconfig.DdevGlobalConfig.NoConfigHere, projectParentDir)
-		err = globalconfig.WriteGlobalConfig(globalconfig.DdevGlobalConfig)
-		if err != nil {
-			util.Failed("Failed to update global config: %v", err)
-		}
 	}
 
 	util.Success("Configuration complete. You may now run 'ddev start'.")
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -143,12 +143,9 @@ var ConfigCommand = &cobra.Command{
 
 // handleConfigRun handles all the flag processing for any provider
 func handleConfigRun(cmd *cobra.Command, args []string) {
-	app, err := getConfigApp(providerName)
-	if err != nil {
-		util.Failed(err.Error())
-	}
+	app := getConfigApp(providerName)
 
-	err = ddevapp.IsAllowedProjectLocation(app)
+	err := ddevapp.HasRecommendedLocation(app)
 	if err != nil {
 		util.Failed("Unable to run `ddev config`: %v", err)
 	}
@@ -324,10 +321,10 @@ func init() {
 }
 
 // getConfigApp() does the basic setup of the app (with provider) and returns it.
-func getConfigApp(_ string) (*ddevapp.DdevApp, error) {
+func getConfigApp(_ string) *ddevapp.DdevApp {
 	appRoot, err := os.Getwd()
 	if err != nil {
-		return nil, fmt.Errorf("could not determine current working directory: %v", err)
+		util.Failed("Could not determine the current working directory: %v", err)
 	}
 
 	// Check for an existing config in a parent dir
@@ -341,9 +338,9 @@ func getConfigApp(_ string) (*ddevapp.DdevApp, error) {
 	}
 	app, err := ddevapp.NewApp(appRoot, false)
 	if err != nil {
-		return nil, fmt.Errorf("could not create new config: %v", err)
+		util.Failed("Could not create a new config: %v", err)
 	}
-	return app, nil
+	return app
 }
 
 // handleMainConfigArgs() validates and processes the main config args (docroot, etc.)
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -430,9 +430,9 @@ func TestConfigInvalidProjectname(t *testing.T) {
 
 }
 
-// TestCmdDisasterConfig tests to make sure we can't accidentally
-// config in homedir, and that config in a subdir is handled correctly
-func TestCmdDisasterConfig(t *testing.T) {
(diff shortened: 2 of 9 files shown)


rfay (commented):
Of course this doesn't prevent them from doing it wrong the first time. (`ddev config` in ~/workspace for example)

gitressa:
Thanks @stasadev, the PR will also solve this issue:
- #5493

rfay (commented):
I tried the simplest case of what I think this is supposed to prevent and it allowed me to do it. I keep my projects in ~/workspace, and I did a `ddev config` or `ddev config --auto` there and it allowed it.

stasadev:
`ddev config` is supposed to be run at the project level at least once to add the exclusion.

So, first run `ddev config --auto` in the project directory `~/workspace/project`, and then go to the parent directory.

stasadev:
I think I can make it better by reading `project_list.yaml` instead of relying on `ddev config`. Thank you for testing.

stasadev:
After reconsidering, I decided to disable `ddev config` in any parent directory of the project.

This is the simplest approach and should prevent confusion, except when `ddev config` is run in the wrong place initially, which we can't fully prevent.

At the same time, I allowed `ddev config` in the project subdirectories. I will see how the tests go.

rfay:
> could not create new config: 'ddev config' is not allowed in the /Users/rkoller/testung directory because it has a project in the subdirectory /Users/rkoller/testung/subdir/drupal
> To use 'ddev config' here, please 'ddev delete' the project in the subdirectory first

I haven't looked yet, but wouldn't one need to both `ddev delete` or `ddev stop --unlist` *and* remove the offending .ddev/config.yaml?

rpkoller:
> I haven't looked yet, but wouldn't one need to both `ddev delete` or `ddev stop --unlist` _and_ remove the offending .ddev/config.yaml?

delete does that, it stops the running project (if not running it starts it to be able to create a snapshot if you dont add --omit-snapshot,) and then removes it from the list of known project.

stasadev:
> Would it be possible to start the message upper case?

I'll take a look. It is written with a lowercase only because of `golangci-lint`.

> i wonder why to add a path to subdirectory? would it be enough to simply tell" there is a ddev project in a subdirectory" without providing an absolute path?

This subdirectory can be many levels deeper, so I think it's best to tell the user the exact location.

> and on second thought it is also sort of misleading, "To use 'ddev config' here, please 'ddev delete' the project in the subdirectory first" is referring to just a single project and saying deleting that project in the subdirectory will enable you to run ddev config in the given directory. but in my case that wouldnt be possible cuz after deleting the first project the next would be listed?

I wanted to give the user general instructions, but we can definitely improve it. Let's brainstorm a better message.

> and would it make sense to go with "recommended" instead of "useful"?

Yes, easy to change, thanks.

rpkoller:
out of curiosity one question first before going into brainstorming mode, is ddev aware how many projects are blockers and also knows all their paths?

stasadev:
> is ddev aware how many projects are blockers and also knows all their paths?

Yes, we have that information (we just need to count it), I used an early return to keep it simple without checking the number of projects.

cmd/ddev/cmd/config.go
```diff
@@ -149,9 +148,9 @@ func handleConfigRun(cmd *cobra.Command, args []string) {
 		util.Failed(err.Error())
 	}
 
-	homeDir, _ := os.UserHomeDir()
-	if app.AppRoot == filepath.Dir(globalconfig.GetGlobalDdevDir()) || app.AppRoot == homeDir {
-		util.Failed("Please do not use `ddev config` in your home directory")
+	err = ddevapp.IsAllowedProjectLocation(app)
```
rfay, line 151: Nice refactoring.

cmd/ddev/cmd/config_test.go
```diff
@@ -470,11 +470,20 @@ func TestCmdDisasterConfig(t *testing.T) {
 	err = os.Chdir(subdir)
 	assert.NoError(err)
 
-	// Make sure that ddev config in a subdir gives an error
+	// Make sure that 'ddev config' in a subdir doesn't give an error for existing project
```
rfay, line 473: TestCmdDisasterConfig needs a better name (and comment). OK to rename. It was built around the homedir failure of course, and now it's covering more ground.

pkg/ddevapp/config.go
```diff
@@ -1382,6 +1381,26 @@ func WriteImageDockerfile(fullpath string, contents []byte) error {
 	return nil
 }
 
+// IsAllowedProjectLocation returns an error if the project location is not allowed
+func IsAllowedProjectLocation(app *DdevApp) error {
+	homeDir, _ := os.UserHomeDir()
+	if app.AppRoot == filepath.Dir(globalconfig.GetGlobalDdevDir()) || app.AppRoot == homeDir {
```
rfay, line 1387: Can we expand this to parents of homedir? I haven't heard of it happening, but what if people accidentally did /home or /Users ?

pkg/ddevapp/config.go
```diff
@@ -1382,6 +1381,26 @@ func WriteImageDockerfile(fullpath string, contents []byte) error {
...
+		return fmt.Errorf("'ddev config' is not useful in your home directory (%v)", homeDir)
+	}
+	projectList := globalconfig.GetGlobalProjectList()
+	for _, project := range projectList {
```
rfay, line 1391: This is a good and adequate strategy, although I have accidentally created projects that have a .ddev in a subdirectory. It's typically when I have done something like a `ddev stop --unlist -a`, removing everything from global project list, or when my global config has been damaged somehow.

pkg/ddevapp/config.go
```diff
@@ -1382,6 +1381,26 @@ func WriteImageDockerfile(fullpath string, contents []byte) error {
...
+		if app.AppRoot == project.AppRoot {
+			return nil
+		}
+		// Do not allow 'ddev config' in any parent directory of any project
+		rel, err := filepath.Rel(app.AppRoot, project.AppRoot)
+		if err == nil && !strings.HasPrefix(rel, "..") {
+			return fmt.Errorf("'ddev config' is not allowed in the %s directory because it has a project in the subdirectory %s\nTo use 'ddev config' here, please 'ddev delete' the project in the subdirectory first", app.AppRoot, project.AppRoot)
```
rfay, line 1398: `ddev delete` is not really required, and does more than expected. `ddev stop --unlist` ? Or should we just always assume that delete is the best answer?

rfay (changes requested):
* `ddev config` in a subdirectory does in fact reconfig the parent directory project, but it inappropriately messes up the message about docroot creation,  (it actually doesn't create the docroot web/web or put that into the config.yaml):

```
$ ddev config --webserver-type=apache-fpm
You are reconfiguring the project at /home/rfay/workspace/d11.
Configuring a 'drupal11' project named 'd11' with docroot 'web' at '/home/rfay/workspace/d11/web/web'.
```
* One thing we may want to keep in mind for people is they don't actually need to use `ddev config` and can use manual creation of a .ddev/config.yaml to work around these restrictions if necessary.
* Is there anything we need to do in docs? I doubt it. But there may be some support events that get triggered by this and can probably be fixed with manual config.yaml creation.

pkg/ddevapp/config.go
stasadev: Yes, we can.

pkg/ddevapp/config.go
stasadev: I don't know yet, I will think about it.

stasadev:
> `ddev config` in a subdirectory does in fact reconfig the parent directory project, but it inappropriately messes up the message about docroot creation

Thanks, I'll take a look.

> One thing we may want to keep in mind for people is they don't actually need to use `ddev config` and can use manual creation of a .ddev/config.yaml to work around these restrictions if necessary.

I agree, if they want to do it, then it's their choice.

> Is there anything we need to do in docs? I doubt it. But there may be some support events that get triggered by this and can probably be fixed with manual config.yaml creation.

I don't know how people develop Drupal, but can they have Drupal modules that are DDEV projects inside other DDEV projects?

rfay:
> I don't know how people develop Drupal, but can they have Drupal modules that are DDEV projects inside other DDEV projects?

That's a good question. The best practice right now is https://github.com/ddev/ddev-drupal-contrib, which I think just does the project it the top directory. There is also https://github.com/joachim-n/drupal-project-contrib-development

But I don't think either one would explicitly use `ddev config`, which is the thing we're working with here.

rfay:
The ability to safely use `ddev config` in a subdirectory and have it config the right thing is a huge improvement to DDEV, thank you!

stasadev:
I implemented all the suggestions and slightly rephrased the error message.
Additionally, I added everything under `~/.ddev` to the exclusion list.

rfay (approved):
It's awesome, and such an excellent improvement, in both directions. `ddev config` in a subdirectory of valid project works *fine* instead of messing up. And `ddev config` in parent directory balks. And it was great that you disabled inside .ddev directory.

rpkoller (commented):
the changes are mostly about changing recommended to allowed after a brief discussion with @rfay and he recommended going with allowed which is the most precise in that context. aside that i did only two minor nitpicks . one change from plural to singular for directory and one rephrasing of a sentence open for discussion. aside that i hope i haven missed a recommend nor was too overeager. and functionwise the PR is still in excellent shape, thank you for all the excellent work!

stasadev:
Thanks @rpkoller, much better now!

I read the related issues and added one more check for `ddev/ddev` repository root. (I run `ddev config` here too 😅)

> * https://github.com/ddev/ddev/issues/5493#issuecomment-1791131045

rfay:
That is so great that you caught the "doing it in ddev repo" thing. Maybe it will only help you and me but it will sure help. Of course... I often do it in the pkg/ddevapp dir. But it's great.

rfay:
Conversation on this in https://drupal.slack.com/archives/C5TQRQZRR/p1737819544903659 - we may not be successful with this. Especially with ddev/ddev-drupal-contrib adding to top-level projects (drupal modules). Then you bring in one of those modules that has a checked-in .ddev directory, and you can no longer `ddev config` in the project that contains it.

PR #6862, opened 2024-12-26
docs: bump mkdocs/readthedocs dependencies for python 3.13

stasadev:
## The Issue

I got Python 3.13 locally, I use `venv` to run mkdocs, and I couldn't build `mkdocs` because of:

```
        File "/tmp/pip-install-bckwce7a/htmlmin_90b52ccd621645329c7baa6349f008e5/htmlmin/__init__.py", line 28, in <module>
          from .main import minify, Minifier
        File "/tmp/pip-install-bckwce7a/htmlmin_90b52ccd621645329c7baa6349f008e5/htmlmin/main.py", line 28, in <module>
          import cgi
      ModuleNotFoundError: No module named 'cgi'
      [end of output]
```

- https://github.com/mankyd/htmlmin/issues/66

Use `htmlmin2` instead of `htmlmin`, which doesn't have `cgi` dependency (which is deprecated from python 3.11)

Bump python mkdocs to 3.13

Fix small warnings, that were noticed by the textlint.

https://ddev--6862.org.readthedocs.build/en/6862/


docs/content/index.md
```diff
@@ -17,7 +17,7 @@ These environments can be extended, version controlled, and shared, so you can t
 
     Runs natively on ARM64 (Apple Silicon) and AMD64 machines.
 
-    * macOS Big Sur (11) or higher, [mostly](./users/usage/faq.md#can-i-run-ddev-on-an-older-mac)
+    * macOS Big Sur (11) or higher
```
rfay, line 20:
```suggestion
    * macOS Ventura (12) or higher. This is primarily driven by the available Docker providers.
```

Might as well update while we're here.  Homebrew is only the past two releases, so that limits colima and lima. Docker Desktop is only current release and previous two.

rfay (approved):
Thanks for keeping track of this!

rfay:
As you already know, `make mkdocs-serve` is easier and doesn't actually require installation of mkdocs (but is faster that way)

rfay:
Oh, I see that `make mkdocs-serve` is completely broken probably in every way, and the docker image we use is unmaintained.

rfay:
* https://github.com/ddev/ddev/pull/6864

PR #6863, opened 2024-12-27
docs: Add link to hostnames article on ddev.com

rfay:
## The Issue

The ddev.com article https://ddev.com/blog/ddev-name-resolution-wildcards/ gives lots of background that isn't really possible in the docs. Link to it.


stasadev (approved):
Looks good to me.

PR #6866, opened 2024-12-28
test: get more disk space on hosted runner, switch to ubuntu 24.04, fixes #6865

### LINKED ISSUE #6865: TestAcquiaPull is failing, probably due to disk space remaining on runner
TestAcquiaPull is failing on upstream/master, see https://github.com/ddev/ddev/actions/runs/12521760766/job/34941933752#step:17:13351

We've been seeing the disk space warning for some time, I guess it's time to fix it.

I can run TestAcquiaPull just fine locally. 

I see this recipe for reducing space: https://github.com/easimon/maximize-build-space

rfay:
## The Issue

- #6865 

We don't have enough room on disk. Try to resolve.

It may be that some test cleanup is in order as well.

* Clean up more things that aren't needed in test runner
* Switch to Ubuntu 24.04 runner
* "Just because" I switched pr-build and push-tagged-dbimage to 24.04 as well. docscheck remains at 22.04 because of misbehavior of one of the actions.
* Clean up after a couple of TestPull/Push tests that weren't being cleaned up
* Clean up after each test in TestDdevFullSiteSetup
* Disable drupal 6/8/9 in TestDdevFullSiteSetup

This gets us from 28GB available to 43GB available at the start. We'll see if it breaks anything.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -33,7 +33,7 @@ permissions:
 jobs:
   build:
     name: Build DDEV executables
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     env:
       AmplitudeAPIKey: ${{ secrets.AMPLITUDE_API_KEY_DEV }}
     steps:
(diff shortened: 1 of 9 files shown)


rfay:
I think the actual problem was a change in how the output of the database query was reported and a couple of fragile tests that insisted on it. But we'd have to deal with the space problem someday.

stasadev (approved):
Looks good to me.

PR #6867, opened 2024-12-28
build: dependabot complaint about net/html

rfay:
## The Issue

- https://github.com/ddev/ddev/security/dependabot/35

Dependabot is complaining about a security vulnerability in net/html

Upgrade it.


stasadev (approved):
Looks good to me.

Issue #6222, opened 2024-05-20
Implement automated tests for DDEV Quickstart docs


rfay:
It's really hard to keep all the [Quickstarts](https://ddev.readthedocs.io/en/stable/users/quickstart/) properly updated as the upstream projects change.

### Describe your solution

Implement either
* Automated tests for the quickstarts (include them into the docs as scripts?) or
* Create parallel automated tests that are intended to match each quickstart, and then carefully maintain them.

mandrasch:
👏👏 Will be awesome!

My two cents:

Personally I would argue against docs integration, because this can get messy fast with different versions + the quickstart docs currently are not focused on no-interaction / non-interactive install (which would be needed for automated testing). 

I'm not a technical expert on setting up such tests, but I could imagine that we could put 1 cms in 1 test file and delete/clean between cms versions - if this is meant to be executed with `ddev/github-action-setup-ddev@v1`? Or should it run as some kind of go test? 

Happy to help/contribute!

rfay:
I agree that docs integration will probably be too hard. I'm thinking we could have a single golang test (or bats, or ...) for each quickstart, and just maintain them in parallel with the quickstarts, so by process a change to the quickstart requires an update to the test.

rfay:
This is done, special thanks to @rpkoller !

PR #6868, opened 2024-12-29
test: Add bats test coverage for backdrop quickstart, for #6222

rfay:
## The Issue

- https://github.com/ddev/ddev/issues/6222

We need to know when the quickstarts stop working

* Add bats framework for testing
* Test the two variants of backdrop
* Add a [ddev/test-backdrop](https://github.com/ddev/test-backdrop) repository for test artifacts
* Update TestDdevFullSiteSetup config to use the new repo
* Minor updates to the backdrop quickstart, since it had a couple of problems (composer.json)

* Try the quickstart - Rendered at https://ddev--6868.org.readthedocs.build/en/6868/users/quickstart/#backdrop
* Review the test run

## Automated Testing Overview

Added `make quickstart-test` to run bats tests in docs/tests


rfay:
@jenlampton @quicksketch would love it if you'd take a quick look at the updated quickstart, https://ddev--6868.org.readthedocs.build/en/6868/users/quickstart/#backdrop

I was a little baffled by the composer install. Is there every a composer.json?

quicksketch:
New instructions for the "Existing projects" look great. Simplified down a bit from before. The "New projects" instructions are identical to before.

> I was a little baffled by the composer install. Is there every a composer.json?

Some site owners still use composer to install other dependencies, even if Backdrop itself does not use it. I think it's fairly safe to remove it to save most Backdrop users the confusion as composer is not a default tool out-of-the-box for Backdrop.

docs/tests/common-setup.bash
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+_common_setup() {
+    TEST_BREW_PREFIX="$(brew --prefix)"
+    load "${TEST_BREW_PREFIX}/lib/bats-support/load.bash"
+    load "${TEST_BREW_PREFIX}/lib/bats-assert/load.bash"
```
stasadev, line 6: I added `BATS_LIB_PATH` to make it work on Linux without `brew`, and people can set `BATS_LIB_PATH` if they have it installed somewhere else locally.

Found this technique here https://github.com/bats-core/homebrew-bats-core/blob/main/tests/tests_helper.bash#L1-L8

(Perhaps we can take advantage of this technique in add-ons, I will leave a note on this for myself.)

```suggestion
    TEST_BREW_PREFIX="$(brew --prefix 2>/dev/null || true)"
    export BATS_LIB_PATH="${BATS_LIB_PATH}:${TEST_BREW_PREFIX}/lib:/usr/lib/bats"
    bats_load_library bats-support
    bats_load_library bats-assert
```

docs/tests/common-setup.bash
```diff
@@ -0,0 +1,17 @@
...
+    mkdir -p ~/tmp
+    tmpdir=$(mktemp -d ~/tmp/${PROJNAME}.XXXXXX)
+    export DDEV_NO_INSTRUMENTATION=true
+    mkdir -p ${tmpdir} && cd ${tmpdir} || exit
```
stasadev, line 10: Should fail here?
```suggestion
    mkdir -p ${tmpdir} && cd ${tmpdir} || exit 1
```

docs/tests/common-setup.bash
```diff
@@ -0,0 +1,17 @@
...
+    ddev delete -Oy ${PROJNAME:-notset}
+}
+
+_common_teardown() {
+  ddev delete -Oy ${PROJNAME}
+  rm -rf ${tmpdir}/${PROJNAME}
```
stasadev, line 16: It removed only `~/tmp/my-backdrop-site.hXV1I6/my-backdrop-site`, but not `~/tmp/my-backdrop-site.hXV1I6`
```suggestion
  rm -rf ${tmpdir}
```

docs/content/users/quickstart.md
```diff
@@ -26,18 +26,25 @@ To get started with [Backdrop](https://backdropcms.org), clone the project repos
...
     ddev config --project-type=backdrop
 
-    # Boot the project and install Composer packages (if required):
+    # Boot the project and install Composer packages (if composer is in use):
     ddev start
-    ddev composer install
+    if [ -f composer.json ] ; then ddev composer install; fi
 
-    # Import a database backup and open the site in your browser:
+    # Import a f backup:
```
stasadev, line 41:
```suggestion
    # Import a database backup:
```

stasadev (changes requested):
I have some suggestions regarding the bats tests.

Everything else looks good to me.

docs/tests/common-setup.bash
rfay: Thanks for noting that, much better with fail return.

rfay:
Thanks @quicksketch - I removed the composer install.

PR #6869, opened 2024-12-29
test: Minor updates to linux-setup.sh

rfay:
## The Issue

There were warnings about mkcert etc in recent PRs, so trying to clean up a bit

We don't need the NFS setup any more

The ngrok setup was out of date.

### DIFF
--- a/.github/workflows/linux-setup.sh
+++ b/.github/workflows/linux-setup.sh
@@ -19,8 +19,8 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
   | sudo tee /etc/apt/trusted.gpg.d/ngrok.asc >/dev/null \
   && echo "deb https://ngrok-agent.s3.amazonaws.com buster main" \
   | sudo tee /etc/apt/sources.list.d/ngrok.list \
-  && sudo apt update \
-  && sudo apt install ngrok
+  && sudo apt-get update -qq >/dev/null \
+  && sudo apt-get install -y -qq ngrok
 
 # Without this .curlrc CircleCI linux image doesn't respect mkcert certs
 echo "capath=/etc/ssl/certs/" >>~/.curlrc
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-setup.sh
```diff
@@ -13,38 +13,26 @@ if [ ! -z "${DOCKERHUB_PULL_USERNAME:-}" ]; then
...
-sudo apt-get install -qq zip jq expect nfs-kernel-server build-essential curl git libnss3-tools libcurl4-gnutls-dev postgresql-client >/dev/null
+sudo apt-get install -y -qq build-essential expect libnss3-tools libcurl4-gnutls-dev postgresql-client >/dev/null
 
-curl -sSL --fail -o /tmp/ngrok.zip https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-amd64.zip && sudo unzip -o -d /usr/local/bin /tmp/ngrok.zip
-
-export PATH=/home/linuxbrew/.linuxbrew/bin:$PATH
-echo "export PATH=/home/linuxbrew/.linuxbrew/bin:$PATH" >>~/.bashrc
+curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
+  | sudo tee /etc/apt/trusted.gpg.d/ngrok.asc >/dev/null \
+  && echo "deb https://ngrok-agent.s3.amazonaws.com buster main" \
```
stasadev, line 20: It seems odd to see `buster` here, but if the official docs recommend it (https://download.ngrok.com/linux), I have no objections.

.github/workflows/linux-setup.sh
```diff
@@ -13,38 +13,26 @@ if [ ! -z "${DOCKERHUB_PULL_USERNAME:-}" ]; then
...
+  | sudo tee /etc/apt/sources.list.d/ngrok.list \
+  && sudo apt update \
+  && sudo apt install ngrok
```
stasadev, line 23:
```suggestion
  && sudo apt-get update -qq >/dev/null \
  && sudo apt-get install -y -qq ngrok
```

stasadev (approved):

.github/workflows/linux-setup.sh
rfay: That is very odd, agreed. But in general debian versions work the same.

PR #6870, opened 2024-12-30
test: Try to improve Docker Desktop reliability with docker desktop restart

rfay:
## The Issue

Reliability of Docker Desktop, especially on Windows WSL2 but also on macOS, has always been a problem.

Docker has recently added command-line control with `docker desktop start/stop/restart`, so we have a more reliable way of controlling it.

This PR attempts to do a stop where we were doing a kill before, and to do a restart at beginning of test.


rfay:
There's nothing wrong with this, but it doesn't stop the Docker Desktop/WSL2 problem, because what happens is WSL2 blows up and crashes, and there's no recovery possible, so the cleanup() doesn't get run, etc. I might take a look and see if there might be a way to restart docker desktop and wsl2 distro from windows side.

rfay:
I guess I'll return to this if the upstream problem gets solved. Closing for now.

PR #6872, opened 2024-12-30
feat: Node.js back-end to nginx, fixes #4854

### LINKED ISSUE #4854: Create a quickstart for a simple Node.js app (without PHP)

rfay:
## The Issue

- #4854 

## TODO

- [ ] Docs
  - [ ] General explanation (Call it "experimental"?)
  - [ ] Simple examples 
  - [ ] Explain how this differs from the earlier (and more common support) of general daemons and node daemons.
- [ ] Add `ddev npx` custom command
- [ ] Tests
- [x] Do we also need `apache-nodejs` webserver type?
- [ ] consider doing a `generic` project type (alias of `php`?) when working on node project type.
- [ ] Consider configuration of the node.js daemon port. Default to 3000 because they always do.
- [ ] Consider changing config.yaml setup so it only shows relevant items. For example, don't add `php_version` by default when it's not being used.
- [ ] Consider svelte or sveltekit project type
- [ ] Consider new configuration for node (the port would then have to be injected into the nginx configuration:
    ```yaml
    nodejs_start_command: "npm start"
    nodejs_port: 3000
    ```
- [ ] Or consider using `web_extra_daemons` and just use that, but the port would still need to be handled

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/nodejs.conf
+++ b/containers/ddev-webserver/ddev-webserver-base-files/etc/supervisor/nodejs.conf
@@ -1,5 +1,5 @@
 [program:nodejs]
-command = bash -c "cd /var/www/html/${DDEV_DOCROOT} && npm run dev"
+command = bash -c "cd /var/www/html && ./node_modules/http-server/bin/http-server -p 3000"
 priority=5
 stdout_logfile=/var/tmp/logpipe
 stdout_logfile_maxbytes=0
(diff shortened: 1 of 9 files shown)


rfay:
Closing in favor of 
* #6935

PR #6874, opened 2025-01-03
fix: In acquia.yaml, specify default site source for ddev pull acquia.

charginghawk:
If you have multiple databases, `acli pull:db acquia` doesn't automatically use the default (the one initially created for the site). This causes ddev pull to instead pull a non-default database.

## The Issue

- #<issue number>

It specifies the default database.

Have an Acquia site with two databases. Run the ddev pull command. Confirm that the default database is pulled.

### DIFF
--- a/pkg/ddevapp/dotddev_assets/providers/acquia.yaml
+++ b/pkg/ddevapp/dotddev_assets/providers/acquia.yaml
@@ -58,7 +58,7 @@ files_import_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     set -eu -o pipefail
-    acli -n pull:files ${ACQUIA_ENVIRONMENT_ID}
+    acli -n pull:files ${ACQUIA_ENVIRONMENT_ID} default
 
 # push is a dangerous command. If not absolutely needed it's better to delete these lines.
 db_push_command:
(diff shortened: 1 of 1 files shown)


rfay:
Hmm, I see this conversation in discord about the same exact thing, but about files.

https://discordapp.com/channels/664580571770388500/1323712310857957536

In that case could you do the same thing in both stanzas?

@danepowell do you think this is popping up related to a change in acli? I didn't see anything recent in a casual review of releases.

rfay:
Changed to draft. Automated tests won't really test this because a forked PR

charginghawk:
I created an MR in acli to try and fix it, worth a shot:

https://github.com/acquia/cli/pull/1839

It may be related to Acquia Cloud Next, we just switched over last month.

danepowell:
I'll take a look when I'm back in the office Monday

danepowell:
I can't think of anything that would have changed in ACLI to cause this; it's always sorted the sites alphabetically. But I agree, default should be first. That will be fixed in https://github.com/acquia/cli/pull/1839

Let me know if there's anything else I can do to help.

rfay:
This seems to test fine in 
* https://github.com/ddev/ddev/pull/6882

So it can go in.

Thanks!

PR #6881, opened 2025-01-07
docs: Stop checking stopwords 'just' and 'simply' [skip ci]

rfay:
## The Issue

The stopword setup for "just" and "simply" aren't helpful. Remove them.

Adjust .textlintrc


stasadev (approved):
Approved! *Simply* said, it's *just* right! 😄

PR #6883, opened 2025-01-07
fix: Fix Acquia push, Minor cleanup in provider integrations [skip buildkite]

rfay:
## The Issue

* Remove old lines from integrations that were there for bad NFS situations
* Make sure the TestAcquiaPush happens successfully

(I was trying to remove the necessity of `ddev auth ssh` for Acquia, but it's actually still required, even though `acli` is used for moving the files around)

I changed acquia.yaml to use user=root pass=root for the push test, because acli does `MYSQL_PWD=db mysqldump`, and that can't work because the ~/.my.cnf has password=root, which takes precedence.

### DIFF
--- a/docs/content/users/providers/acquia.md
+++ b/docs/content/users/providers/acquia.md
@@ -7,9 +7,7 @@ DDEV’s Acquia integration pulls database and files from an existing project in
 ## Acquia Quickstart
 
 1. Get your Acquia API token from *Account Settings* → *API Tokens*.
-2. Make sure you’ve added your SSH key to your Acquia account in *Account Settings* → *SSH Keys*.
-3. Run [`ddev auth ssh`](../usage/commands.md#auth-ssh). (Typically once per DDEV session, not every pull.)
-4. In `~/.ddev/global_config.yaml` add or update the [`web_environment`](../configuration/config.md#web_environment) section with the API keys:
+2. In `~/.ddev/global_config.yaml` add or update the [`web_environment`](../configuration/config.md#web_environment) section with the API keys:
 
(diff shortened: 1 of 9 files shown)


docs/content/users/providers/acquia.md
```diff
@@ -23,7 +21,7 @@ DDEV’s Acquia integration pulls database and files from an existing project in
     ddev config global --web-environment-add="ACQUIA_API_KEY=xxxxxxxx,ACQUIA_API_SECRET=xxxxx"
     ```
 
-5. In the project `.ddev/config.yaml` add the `ACQUIA_ENVIRONMENT_ID` environment variable:
+3. In the project `.ddev/config.yaml` add the `ACQUIA_ENVIRONMENT_ID` environment variable:
```
GuySartorelli, line 24: If you swap these all to `1.` you'll never have to update the numbering again. Markdown sees `1. 1. 1.` and renders out a correct ordered list.

rfay: Not true in our case as the markdown rules we have set up require the more readable 1, 2, 3. But agreed that it wouldn't be unreasonable to change the rules. Our goal has been to make all the markdown readable, and the 1, 1, 1 technique is definitely not as readable (but lots easier to maintain).

Thanks for reviewing!

rfay:
TestAcquiaPush is failing due to 
- https://github.com/ddev/ddev/issues/6886

but this PR disables it, which is probably fine for right now.

Goal for this PR then will be to verify that all the other provider integrations (especially platform and upsun) are working

PR #6895, opened 2025-01-14
test: Quickstart tests for TYPO3 quickstart 

rpkoller:
Initial draft for the composer based quickstart for setting up TYPO3. tested locally and it is working fine. the second available quickstart is missing so far for this PR since there is no github repo for a typo3  test install yet as far as i can see. 

## The Issue

- #<issue number>

### DIFF
--- a/docs/tests/typo3.bats
+++ b/docs/tests/typo3.bats
@@ -37,3 +37,36 @@ teardown() {
   assert_output --partial "location: /typo3/install.php"
   assert_output --partial "HTTP/2 302"
 }
+
+@test "TYPO3 git based quickstart with $(ddev --version)" {
+  skip
+  # PROJECT_GIT_URL=https://github.com/ddev/test-typo3.git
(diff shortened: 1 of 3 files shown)


docs/tests/typo3.bats
```diff
@@ -0,0 +1,71 @@
...
+  assert_success
+  # ddev launch
+  run bash -c "DDEV_DEBUG=true ddev launch"
+  assert_output --partial "FULLURL https://${PROJNAME}.ddev.site"
+  assert_success
+  # validate running project
+  run curl -sfI https://${PROJNAME}.ddev.site
+  assert_success
+  assert_output --partial "location: /typo3/install.php"
+  assert_output --partial "HTTP/2 302"
```
rfay, line 38: As we talked in Discord, it would be wonderful if we could actually validate that the install..php can be used, but that's for another time as we learn how to do this better.

rfay (approved):
It's great, thanks!

PR #6900, opened 2025-01-15
feat: override `PAGER` with `DDEV_PAGER` env variable, for #6842

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/6842#issuecomment-2590294868

We don't have a quick and easy way to override `PAGER` variable in the `web` and `db` containers.

Adds an optional environment variable `DDEV_PAGER`.

I'm unsure where to add information about DDEV_PAGER in the docs (should we?).
The section on [Environment Variables Provided](https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/#environment-variables-provided) is not the right place, as it covers variables inside the container, whereas `DDEV_PAGER` is used outside the container.

```
ddev start
ddev exec 'echo $PAGER'
less -SFX
```

```
export DDEV_PAGER=less
ddev start
ddev exec 'echo $PAGER'
less
```

```
export DDEV_PAGER=
ddev start
ddev exec 'echo $PAGER'
<empty>
```

```
unset DDEV_PAGER
ddev start
ddev exec 'echo $PAGER'
less -SFX
```

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -71,7 +71,7 @@ services:
       - POSTGRES_PASSWORD=db
       - POSTGRES_USER=db
       - POSTGRES_DB=db
-      - PAGER=less -SFX
+      - PAGER=${DDEV_PAGER-less -SFX}
       - TZ={{ .Timezone }}
       - USER={{ .Username }}
     command: ${DDEV_DB_CONTAINER_COMMAND}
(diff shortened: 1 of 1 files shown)


pkg/ddevapp/app_compose_template.yaml
```diff
@@ -71,7 +71,7 @@ services:
       - POSTGRES_PASSWORD=db
       - POSTGRES_USER=db
       - POSTGRES_DB=db
-      - PAGER=less -SFX
+      - PAGER=${DDEV_PAGER-less -SFX}
```
stasadev, line 74: This is intentional syntax; the below syntax will not work for an empty string: `export DDEV_PAGER=`, because it will use the default `less -SFX` instead of an empty string:

```suggestion
      - PAGER=${DDEV_PAGER:-less -SFX}
```

rfay: That fix is right.

rfay:
> I'm unsure where to add information about DDEV_PAGER in the docs (should we?).

It's a good question. We do have other things that can probably be overridden, but DDEV is usually overriding them.

rfay (approved):
This works for me, and will make me happy enough. Not sure my use-case is that important.

I guess the question would be, should the default be with-pager or without-pager? I guess we can find out whether we get pushback on the default pager.

stasadev:
> I guess the question would be, should the default be with-pager or without-pager?

Yes, let's use with-pager for now, otherwise we won't get any feedback.

rfay:
You'd get feedback from @tyler36 ! :)

PR #6901, opened 2025-01-15
docs: replace GOOS and GOARCH for a DDEV_ alternative, for #6752

stasadev:
## The Issue

- #6752

The documentation was not updated.

Updates it.


rfay (approved):
Thanks!

PR #6902, opened 2025-01-15
fix: provide a dummy ddev command inside container, allow web commands there, fixes #6574

### LINKED ISSUE #6574: Make some DDEV custom commands available inside the web container

Some people work inside the web container, for example with VS Code's "Remote Explorer" and Devcontainers setup, @ultimike teaches this approach in his classes. It's nice, because you're directly on the "server" instead of using `ddev` commands to filter in there. (It's almost the same as `ddev ssh` of course.)

However, he teaches people to use the undocumented `enable_xdebug` and `disable_xdebug` commands, for example.

### Describe your solution

Some `ddev` custom commands work great inside the container. xdebug is one of these:

`sudo ln -s /mnt/ddev-global-cache/global-commands/web/xdebug /usr/local/bin/xdebug` for example, will make an `xdebug` command work inside container just like it works outside (except without `ddev`)

This doesn't work with `xhprof` or `wp` because those namespaces are already claimed. 

It's not needed for `drush` or `php`, which are already in the path as is. 

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

- #6574 

* People who work inside the web container like in the @ultimike technique often type `ddev <whatever>` and waste their keystrokes.
* Some commands like `xdebug on` would be useful inside the web container, and they're already there.

Add a simple `ddev` command. Works when things are available and when not available:

```
rfay@d11-web:/var/www/html$ ddev
You executed 'ddev ' inside the web container
but many DDEV commands are not available.
```

```
rfay@d11-web:/var/www/html$ ddev drush cr
You executed 'ddev drush cr' inside the web container
but many DDEV commands are not available.

However, you may be able to use just the command 'drush cr'
```

Also, specially for @ultimike :

```
rfay@d11-web:/var/www/html$ xdebug on
Enabled xdebug
rfay@d11-web:/var/www/html$ xdebug off
Disabled xdebug
rfay@d11-web:/var/www/html$ xhprof on
Enabled xhprof.
After each web request or CLI process you can see all runs, most recent first, at
https://d11.ddev.site/xhprof
```

* Try using `ddev` inside `ddev ssh`
* try using some various things like `xdebug on`

## Automated Testing Overview

* Added check for xdebug and such in web container tests

## Release/Deployment Notes

It's possible this needs note in docs or release, but maybe better to just have people discover it?

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Imitation/replacement for `ddev` inside the web container
+# Partial functionality
+
+set -eu -o pipefail
+
+printf "You executed 'ddev $*' inside the web container\nbut many DDEV commands are not available.\n"
+
+if [ $# -ge 1 ] && command -v "$1" >/dev/null 2>&1; then
+  printf "\nHowever, you may be able to use just the command '$*'\n".
+fi
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
@@ -8,5 +8,5 @@ set -eu -o pipefail
 printf "You executed 'ddev $*' inside the web container\nbut many DDEV commands are not available.\n"
 
 if [ $# -ge 1 ] && command -v "$1" >/dev/null 2>&1; then
-  printf "\nHowever, you may be able to use just the command '$*'\n".
+  printf "\nHowever, you may be able to use a command like:\n\n$*\n".
 fi
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
@@ -8,5 +8,5 @@ set -eu -o pipefail
 printf "You executed 'ddev $*' inside the web container\nbut many DDEV commands are not available.\n"
 
 if [ $# -ge 1 ] && command -v "$1" >/dev/null 2>&1; then
-  printf "\nHowever, you may be able to use a command like:\n\n$*\n".
+  printf "\nHowever, you may be able to use a command like:\n\n$*\n"
 fi
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/wp
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/wp
@@ -6,4 +6,7 @@
 ## ProjectTypes: wordpress
 ## ExecRaw: true
 
+# Ignore anything we find in the mounted global commands
+PATH=${PATH//\/mnt\/ddev-global-cache\/global-commands\/web/}
+
 wp "$@"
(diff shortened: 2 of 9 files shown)


rfay:
@ultimike would love it if you could take the artifacts in https://github.com/ddev/ddev/pull/6902#issuecomment-2593583120 for a test ride.

stasadev (commented):
It's great, but I'm afraid it's not quite what we want.

In DDEV v1.24.1, when there is no `drush` command, then `ddev drush` shows me:
```
$ ddev drush
drush is not available. You may need to 'ddev composer require drush/drush'
```

But with this PR, `ddev drush` is stuck in a loop with no messages

```
$ ddev drush
<nothing-and-waiting>
```

when the only `drush` available is:

```
$ which -a drush
/mnt/ddev-global-cache/global-commands/web/drush
```

The same thing is inside the container.

rfay:
I guess that's because you have a project without drush in it? Otherwise it would be in vendor/bin and be in the path.

rfay:
When there is no drush inside the container, I get this result:

```
$ ddev drush
You executed 'ddev drush' inside the web container
but many DDEV commands are not available.

However, you may be able to use just the command 'drush'
```

I don't understand why the ddev script would hang, can you take a look in your env? If you haven't pulled that image, please check it. I imagine you have current. But the `ddev` script doesn't try to execute anything, so it shouldn't hang.

stasadev:
No, run `ddev drush` outside the container (or `drush` inside the container), and watch it hang.

I have a fix for this:

```diff
-if ! command -v drush >/dev/null; then
+if ! command -v drush >/dev/null || [ "$(which drush)" = "/mnt/ddev-global-cache/global-commands/web/drush" ]; then
```

The same check needs to be done for `pint`, `sake`, `typo3`, and `wp`.

rfay:
Oh, the problem is with `ddev drush` *outside* container, not *inside*, thanks.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/ddev
```diff
@@ -0,0 +1,12 @@
...
+
+# Imitation/replacement for `ddev` inside the web container
+# Partial functionality
+
+set -eu -o pipefail
+
+printf "You executed 'ddev $*' inside the web container\nbut many DDEV commands are not available.\n"
+
+if [ $# -ge 1 ] && command -v "$1" >/dev/null 2>&1; then
+  printf "\nHowever, you may be able to use a command like:\n\n$*\n".
```
stasadev, line 11: Removed the dot:

```suggestion
  printf "\nHowever, you may be able to use a command like:\n\n$*\n"
```

Because it was added here:

```
However, you may be able to use a command like:

drush
.stas@d11-web:/var/www/html$
```

rfay:
I'll try to add a generic fix for those commands.

pkg/ddevapp/global_dotddev_assets/commands/web/wp
```diff
@@ -6,4 +6,7 @@
 ## ProjectTypes: wordpress
 ## ExecRaw: true
 
+# Ignore anything we find in the mounted global commands
+PATH=${PATH//\/mnt\/ddev-global-cache\/global-commands\/web/}
+
```
stasadev, line 11: I didn't check at first, but there's `/usr/local/bin/wp` available all the time, so it's most likely not needed here.

rfay:
See what you think about https://github.com/ddev/ddev/pull/6902/commits/e2c6673e9a7247b126bb61a2c2702df590caf547

Thanks, glad you caught that!

stasadev (approved):
Amazing! Works as expected now.

pkg/ddevapp/global_dotddev_assets/commands/web/wp
rfay: Agreed. I did it anyway based on it being in your list. Does no harm.

PR #6904, opened 2025-01-15
docs: change the quickstart repo for typo3 to test-typo3

rpkoller:
@stasadev noticed that the quickstart for the git based approach to install typo3 is still using the phantasy url instead of the test-typo3 repo on the ddev org. this PR updates that. 

## The Issue

- #<issue number>


stasadev (approved):
Thanks!

rfay (commented):
Thanks!

Issue #6726, opened 2024-11-15
DDEV Ignoring Changes in settings.php on Drupal 11 Site


TorFJ:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

I am using ddev on a Drupal 11 site. The problem is when I edit the settings.php file changes are completly ignored. I have tried to clear all the cache and restarting the ddev container  but nothing works. 
Does anybody have any suggestion on how to fix this? 

### Expected Behavior

Changes in settings.php file should be applied.

### Actual Behavior

Changes in settings.php are not being applied.

TorFJ:
It seems like others are having the same problem, see this page: https://www.drupal.org/forum/support/post-installation/2024-06-13/settingslocalphp-seems-to-be-cached

rfay:
Hi @TorFJ - I believe you're using mutagen, you have `disable_settings_management: true` and you've done a Drupal install, and drupal has made the settings.php read-only (as it has a habit of doing), so mutagen can't edit it. 

Please do a `ddev mutagen reset` and `ddev start`. I think it's probably been warning you about this when you've done `ddev start`.  

If you had done the requested `ddev debug test` I could have said for sure. But I imagine if you do a `ddev mutagen reset` and a `ddev start` you'll be functioning again. For good measure, please make sure that you settings.php is writable.

TorFJ:
Thank you @rfay much appreciated! After making the sites/default writeable it works without problems. 
All the best and thanks for all the hard work you been doing on this project!

Issue #6491, opened 2024-08-16
Changes to settings.php do not affect the (Drupal) site


jenlampton:
### Preliminary checklist

- [X] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [X] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [X] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [X] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Please make sure that you have already looked at troubleshooting guide ========
Troubleshooting guide: https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/ 
Simple things to check:
* ddev poweroff
* Restart Docker Provider
* Reboot computer
* Temporarily disable VPN and firewall
Press any key to continue:


======== Output file will be in /var/folders/mw/_p1cl_rx69b3ll553w9cdvcc0000gn/T/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/local/bin/ddev
/usr/local/bin/ddev

These config files were loaded for project emerisa: [/Users/jlampton/Sites/emerisa/drupal/.ddev/config.yaml]
name: emerisa
type: drupal7
docroot: docroot
php_version: 7.4
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.23.2
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.11}
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 20
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
No registered add-ons were found. Add-ons installed before DDEV v1.22.0 will not be listed.
Update them with `ddev get` so they'll be shown.


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /Users/jlampton/Sites/emerisa/drupal
#ddev-generated
unmodified #ddev-generated found in .ddev/mutagen/mutagen.yml


======== Creating dummy project named tryddevproject-2594 in ../tryddevproject-2594 ========
Creating a new DDEV project config in the current directory (/Users/jlampton/Sites/emerisa/tryddevproject-2594)
[...]

PR #6913, opened 2025-01-17
fix: update Drupal 7 settings.ddev.php and settings.php to match Drupal 7.103

eporama:
- Adds new Trusted Host Patterns to eliminate the warning on status report. 
- Adds `skip_permissions_hardening` conf
- Replaces settings.php contents with entire `default.settings.php` file from Drupal 7.103

## The Issue
Mutagen also fails if attempting to modify the settings.ddev.php files after installation if `$conf['skip_permissions_hardening']` is not set to `TRUE`.

Drupal 7 has added many features to the `default.settings.php` including a version of Trusted Host Patterns similar to D10/11 which cause an error on status reports:

 ```
{
  "title": "Trusted Host Settings",
  "value": "Not enabled",
  "description": "The trusted_host_patterns setting is not configured in settings.php. This can  \nlead to security vulnerabilities. It is *highly recommended* that you  \nconfigure this. See Protecting against HTTP HOST Header attacks [1] for more  \ninformation.\n\n[1] https://www.drupal.org/node/1992030\n",
  "severity": "Error",
  "sid": 2
}
```

This pr introduces the correct settings.php with documentation on the new features and adds the following code to `settings.ddev.php`

```
// This will prevent Drupal from setting read-only permissions on sites/default.
$conf['skip_permissions_hardening'] = TRUE;

// This will ensure the site can only be accessed through the intended host
// names. Additional host patterns can be added for custom configurations.
$conf['trusted_host_patterns'] = ['.*'];
```

Installing Drupal 7 with this configuration should eliminate

### DIFF
--- a/pkg/ddevapp/drupal/drupal7/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal7/settings.ddev.php
@@ -29,6 +29,10 @@ $databases['default']['default']['port'] = $port;
 
 $drupal_hash_salt = '{{ $config.HashSalt }}';
 
+// This will ensure the site can only be accessed through the intended host
+// names. Additional host patterns can be added for custom configurations.
+$conf['trusted_host_patterns'] = ['.*'];
+
 // Enable verbose logging for errors.
 // https://www.drupal.org/docs/7/creating-custom-modules/show-all-errors-while-developing
 $conf['error_level'] = 2;
--- a/pkg/ddevapp/drupal/drupal7/settings.ddev.php
+++ b/pkg/ddevapp/drupal/drupal7/settings.ddev.php
@@ -29,6 +29,9 @@ $databases['default']['default']['port'] = $port;
 
 $drupal_hash_salt = '{{ $config.HashSalt }}';
 
+// This will prevent Drupal from setting read-only permissions on sites/default.
+$conf['skip_permissions_hardening'] = TRUE;
+
 // This will ensure the site can only be accessed through the intended host
 // names. Additional host patterns can be added for custom configurations.
 $conf['trusted_host_patterns'] = ['.*'];
--- a/pkg/ddevapp/drupal/drupal7/settings.php
+++ b/pkg/ddevapp/drupal/drupal7/settings.php
@@ -2,6 +2,7 @@
 
 // DDEV-created Drupal 7 settings.php from upstream default.settings.php
 
+
 /**
  * @file
  * Drupal site-specific configuration file.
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/ddevapp/drupal/drupal7/settings.php [hunk 1]
```diff
@@ -202,12 +203,17 @@
...
+ *
+ * NOTE: NO_AUTO_CREATE_USER was removed in MySQL 8.0.11.
+ * Some hosting providers/MySQL packages may report the wrong MySQL version.
+ * If this is the case, set 'sql_mode' manually:
  *
  * @code
  * $databases['default']['default'] = array(
  *   'init_commands' => array(
  *     'big_selects' => 'SET SQL_BIG_SELECTS=1',
+ *     'sql_mode' => "SET sql_mode = 'REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,STRICT_TRANS_TABLES,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO",
```
penyaskito, line 216: Should this also be a default for ddev connection?

eporama: That's in a section on more advanced changes that _might_ be necessary:

```
 * Advanced users can add or override initial commands to execute when
 * connecting to the database server, as well as PDO connection settings. For
 * example, to enable MySQL SELECT queries to exceed the max_join_size system
 * variable, and to reduce the database connection timeout to 5 seconds.
 *
 * NOTE: NO_AUTO_CREATE_USER was removed in MySQL 8.0.11.
 * Some hosting providers/MySQL packages may report the wrong MySQL version.
 * If this is the case, set 'sql_mode' manually:
```

I don't know that we need to set those modes by default, but it doesn't seem like it.

pkg/ddevapp/drupal/drupal7/settings.ddev.php
```diff
@@ -29,6 +29,13 @@
 
 $drupal_hash_salt = '{{ $config.HashSalt }}';
 
+// This will prevent Drupal from setting read-only permissions on sites/default.
+$conf['skip_permissions_hardening'] = TRUE;
```
rfay, line 33: Last I checked, Drupal 7 didn't have this feature. It would have saved lots of pain if it had! It's important! 

Drupal's ridiculous attempt at security by making sites/default and sites/default/settings.php non-writable during an install has annoyed me for a really, really long time :)

pkg/ddevapp/drupal/drupal7/settings.php
rfay: For settings.php, I prefer to use the exact upstream in-code settings.php. Any adjustments belong in settings.ddev.php

rfay (commented):
Thanks! I'm sure this is great. Please provide a link link to the upstream settings.php.

Recent go-arounds on this for Drupal 7 are:
* https://github.com/ddev/ddev/issues/6726
* https://github.com/ddev/ddev/issues/6491

I had no idea that the `skip_permissions_hardening` had made it into D7, makes me enormously happy. I was thinking on hike today that settings.ddev.php could make settings.php writable in *php* code.

rfay (approved):
I diffed against D7's default.settings.php and checked the settings.ddev.php, and D7 comes up perfectly. 

Thanks for this!

stasadev (approved):
Thank you!

PR #6911, opened 2025-01-16
fix: nginx-site-drupal11.conf wasn't being installed, fixes #6910

rfay:
## The Issue

- #6910

The correct nginx configuration (.ddev/nginx_full/nginx-site.conf) was not being installed for `drupal11` or `drupal` projects.

I didn't check, but I assume this was a regression in v1.24.0, when we introduced the `drupal11` project type.

## How This PR Solves The Issue

Provide the nginx config.

## Manual Testing Instructions

`ddev start` in a `drupal11` or `drupal` project and inspect .ddev/nginx_full/nginx-site.conf

## Automated Testing Overview

- [x] This still needs test coverage for this situation. Just looking at the nginx-site.conf after start may be adequate.
- [x] Test in TestPHPWebserverType can probably detect problems with this
- [x] Look at the docs for adding a project type and make sure the nginx config is there.
- [x] Make sure the `drupal` project type is working correctly

PR #6914, opened 2025-01-18
test: some tests should skip skippable testsites

rfay:
## The Issue

I noted in looking at the test in 
* https://github.com/ddev/ddev/pull/6911

That even though we've implemented skipping tests for some of the project types in ddevapp_test.go, not all tests respect the skip

Make them skip tests when Skip is set.


stasadev (commented):
I see 3 more `range TestSites` without `Disable` in autocompletion_test.go (TestAutocompletionForStopCmd, TestAutocompletionForStartCmd, TestAutocompletionForDescribeCmd)

stasadev (approved):
I didn't read it carefully enough, now I see that sites are being added to `addSites()`, where `Disable` is used.

rfay:
Thanks for studying it carefully! I'm so happy that you check these things.

PR #6924, opened 2025-01-21
build: rename mailhog.conf to mailpit.conf

rfay:
## The Issue

I happened to discover that in the ddev-webserver, the mailpit configuration was still called `mailhog.conf`

Rename it. That should be all required


stasadev (approved):
Looks good, `ddev mailpit` works for me as it should.

PR #6935, opened 2025-01-24
feat: Add "generic" webserver type for Node.js and other things like FrankenPHP, fixes #4854, fixes #5655, fixes #4312

### LINKED ISSUE #4312: Feature request: Caddy webserver support

In the PHP-world (and in general) there's currently two big players when it comes to webservers: apache and nginx, both supported by ddev. 
Since quite a while there's another one which is coming up more and more: Caddy.
It's being loved for its combination of performance and simplicity to configure.

### Describe your solution

Besides nginx and apache, it would be quite nice to also choose Caddy as webserver alternative in ddev.

### Additional context

Related:
* https://github.com/ddev/ddev/issues/5664

### LINKED ISSUE #4854: Create a quickstart for a simple Node.js app (without PHP)

### LINKED ISSUE #5655: FrankenPHP Support

FrankenPHP (frankenphp.dev) released its first version a few days ago. Do you have plans to integrate it with DDEV?

### Describe your solution

There is a docker image available for quick setup https://frankenphp.dev/docs/docker/

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

- #4854
- Replaces #6872 
- #5655 (This will be the most likely short-term support of FrankenPHP)
- #4312  (This will be the most likely short-term support of Caddy)

We have long wanted to be able to provide back-ends other than the nginx/apache/php ones, but in the Node.js world (and python/django) the webserving needs to be set up in the repository itself, not pre-set as it is in the php world.

* Provide a "generic" webserver_type that doesn't run php-fpm or apache or nginx
* Provide a "generic" project type
* Docs at https://ddev--6935.org.readthedocs.build/en/6935/users/quickstart/

## TODO

- [x] Test on codespaces
- [x] Test on gitpod
- [x] Verify with a generic setup and `ddev describe` and `ddev list`
- [x] Docs update for the new type
- [x] Docs example of how to do svelte
- [x] Docs example of a basic Node.js webserver
- [x] Consider doc/quickstart and test for frankenphp, or perhaps a blog post on next release explaining how it works as an experiment.
- [x] Quickstart and test ? for Sveltekit?
- [x] franken example shows incorrect project URL ` http://127.0.0.1:0`
- [x] Consider binding ports for web_extra_daemons items
- [x] Make normal project types work with `generic` webserver (drupal11, etc.)
- [ ] gitpod: Incorrect "Your project can be reached at" from `ddev start` with generic (it can't be reached)
- [x] Consider adding generic webserver_type and web_extra_exposed_ports and web_extra_daemons to Amplitude
- [x] If port 80 is exposed, it breaks the traefik config, because port 80 is duplicated there. 
- [x] localhost URLs not showing in "project URLs" section if the bound port is not 80/443. 
- [x] "Project URLs" is empty if there is nothing bound on port 80/443, meaning that $DDEV_PRIMARY_URL is empty, so `ddev launch` fails
- [ ] Consider a python/django example.

 - [ ] Use the quickstart to do Node.js, Sveltekit, FrankenPHP examples, or something else.
 - [ ] Read docs changes.  (https://ddev--6935.org.readthedocs.build/en/6935/users/quickstart/)

## Automated Testing Overview

- [x] Add test for generic type
- [x] Add test for svelte
- [x] Test for generic node.js
- [x] Consider a test for python/django
- [x] Consider a test for frankenphp

## Release/Deployment Notes

* GetExposedContainerPorts() (now GetBoundHostPorts()) was incorrect from its inception I believe. It was looking in the wrong place for bound ports. This may affect the behavior of the router, which queries this to see if it needs to bind more ports. So maybe it will be better? Or we'll have new troubles?


GuySartorelli:
This looks like a good overall approach 👍

rfay:
I can do frankenphp with this:

config.frankenphp.yaml:
```yaml
web_extra_daemons: 
  - name: "frankenphp"
    command: "frankenphp php-server --listen 0.0.0.0:80 --root /var/www/html/web -v -a"
    directory: /var/www/html/web
web_extra_exposed_ports:
  - name: "frankenphp"
    container_port: 80
    http_port: 80
    https_port: 443
```

.ddev/web-build/Dockerfile.frankenphp:
```Dockerfile
RUN curl https://frankenphp.dev/install.sh | sh
RUN mv frankenphp /usr/local/bin/
```

rfay:
Thanks for testing! If you didn't want `web` as the docroot, you could leave the `docroot` out of the config.yaml, or set it to ""

rfay:
> Also, as expected, xdebug doesn't work , even though it suggests that it does

I didn't originally think of this PR as even relating to anything PHP-ish (aimed at node and maybe python/django) but realized this morning that it could do frankenphp/caddy. For Xdebug to work, `host.docker.internal` would need to be provided. That might be possible to do. 

> Stepping back a bit, how would this work with the various DDEV project types?

My experiment was with drupal11, and I immediately missed the `drupal11` project type. But since all the PHP project types work fine with apache-fpm and nginx-fpm, I don't see any immediate reason why you couldn't use the `wordpress` project type with the `generic` webserver_type, so I imagine I'll try to make that work, in this PR or later. It seems like it should be fine. 

`db` already works out of the box, with no questions asked, same as always. (I did a drupal11 install with it.) Except when using the `generic` project type you have to do the db configuration yourself, as if with settings management disabled.

rfay:
I tested on gitpod and codespaces. It seems to do no harm to either one, but you can't use generic webservers started by `web_extra_daemons` because we don't use the router currently in those areas, and there's no binding of the web_extra_daemons ports.

rfay:
@nickchomey I took a quick look at xdebug with frankenphp and didn't find an easy way to enable it, and seemed to fail at configuring frankenphp at all. The docs say you can put a php.ini in the directory of frankenphp, which in this case is /usr/local/bin, but I wasn't successful in a quick work on that. Normally ddev would do a `phpenmod xdebug` in enable_xdebug, but frankenphp provides no similar command that I can sort out. But when you do a `ddev xdebug`, ddev just enables xdebug via enable_xdebug which does a `phpenmod`, which has no effect at all on frankenphp, because it has its own configuration system. `frankenphp php-server` does not provide an arg to give a root to config (if it did, we could point it to /etc/php/8.4/fpm)

stasadev:
> Here's an interesting error: when I change it to `docroot:""`, frankenphp fails with this error
> 
> `Error: loading new config: loading http app module: provision http: server php: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 3: loading handler modules: position 0: loading module 'php': provision http.handlers.php: unable to resolve root symlink: lstat /var/www/html/-v: no such file or directory`
> 
> For some reason it is now trying to find the `-v` directory...
> 
> I've reverted to /web for now

Fix for `docroot: ""` in `config.frankenphp.yaml`:

```diff
-command: "frankenphp php-server --listen 0.0.0.0:80 --root ${DDEV_DOCROOT} -v -a"
+command: "frankenphp php-server --listen=0.0.0.0:80 --root=${DDEV_DOCROOT} -v -a"
```

Debug enable:
```
ddev exec "echo environment=CADDY_GLOBAL_OPTIONS=debug | sudo tee -a /etc/supervisor/conf.d/frankenphp.conf"
ddev exec "supervisorctl reload && sleep 2 && (pkill frankenphp || true) && supervisorctl restart webextradaemons:frankenphp"
```

Debug disable:
```
ddev exec sudo sed -i '/environment=CADDY_GLOBAL_OPTIONS=debug/d' /etc/supervisor/conf.d/frankenphp.conf
ddev exec "supervisorctl reload && sleep 2 && (pkill frankenphp || true) && supervisorctl restart webextradaemons:frankenphp"
```

I have no idea how to test this debug, but `CADDY_GLOBAL_OPTIONS=debug` shows up in `phpinfo();`

P.S. I don't really understand why I needed `pkill frankenphp` ,but without it, the previous frankenphp processes still worked.

rfay:
One oddity is that it *feels* to me like the frankenphp implementation is sluggish compared to the traditional nginx-fpm or apache-fpm.

rfay:
> It would need to be added like this https://frankenphp.dev/docs/docker/#how-to-install-more-php-extensions

https://frankenphp.dev/ says "Popular PHP extensions, including OPcache and XDebug, are natively supported!". That would say to me that you don't have to install extra extensions for Xdebug.

rfay:
> Could be docker

Well, php-fpm is in exactly the same context, inside docker. IMO the most likely factor is that we have a finely tuned php.ini for php-fpm.

> tagging the frankenphp devs to chime in here

No thanks, the goal of this PR is to provide a first-class Node.js back-end, everything else is secondary. After this goes in we can do some figuring of how to properly get php.ini into frankenphp, etc.

rfay:
Thanks for taking a look at this! The problem with empty DDEV_DOCROOT (docroot in root of project) should be solved now for FrankenPHP. The issue was when it was empty there was no argument to `--root`, now there is. See quickstart at https://ddev--6935.org.readthedocs.build/en/6935/users/quickstart/#generic-frankenphp

rfay:
@andy-blum it would be lovely if you could take a look, thanks!

rpkoller:
i've tested both quickstarts and they are working as expected. aside that i am not really able to provide much more input nor nitpick due to my general lack of knowledge about the entire field of javascript. the only thing i know those quickstarts feel sort of complex to apply. with several steps. it is a good thing to have a generic project type but i wonder would it make sense to have a node.js project type alongside with two sub types svelt and node.js webserver. that way it would mirror the quickstarts and at the same time it might be possible to run some of setup up processes behind the scenes automatically? things like creating the yaml fie and running those curl and npx commands? so the quickstart guides could become even more condensed? or would that too opinionated and too much of a maintenance burden? ( i mean with quickstart tests in place it would noticeable easily if something breaks? )

rfay:
Thanks for reviewing @rpkoller - Agreed that these are complex. The problem is that as you look at just these two examples (Node.js and Svelte) they have nothing to do with each other, so it's impossible to standardize an approach. (That's why our Python/Django project types failed, they're the same level of every-project-is-completely-different.)

tyler36 (commented):
I gave this a try today, here are some random thoughts:

- SvelteKit Quickstart worked but I was prompted 2 times during the installation:

```shell
Need to install the following packages:
sv@0.6.16
_Ok to proceed? (y) y_

┌  Welcome to the Svelte CLI! (v0.6.16)
│
_◇  Directory not empty. Continue?_
│  Yes
```

- NodeJS quickstart result in a functioning website. Nice!

- I wonder, is it worth disabling the database for NodeJs projects by default? Maybe it's just me, but my javascript projects tend to be front-end API types.

- How do we manage all the potential new QuickStart guides for "latest-shiny-js-framework"?

tyler36:
Any ideas how to setup debugging with VSCode?
Usually, I start the "Javascript Debug Terminal" and run my code.
I tried setting up a DDEV version:
```js
      {
        "type": "node",
        "request": "launch",
        "name": "DDEV Javascript Debug Terminal",
        "port": 9229,
        "runtimeExecutable": "ddev",
        "args": [
          "ssh"
        ],
        "console": "integratedTerminal"
      },
 ```

... exposing the port ...

```
web_extra_exposed_ports:
  - name: node-debug
    container_port: 9229
    http_port: 9228
    https_port: 9229
```

and running 

```shell
$ vitest --inspect 0.0.0.0:9229 --no-file-parallelism
```

rpkoller:
> Thanks for reviewing @rpkoller - Agreed that these are complex. The problem is that as you look at just these two examples (Node.js and Svelte) they have nothing to do with each other, so it's impossible to standardize an approach. (That's why our Python/Django project types failed, they're the same level of every-project-is-completely-different.)

yes, but i not meant have one project type for every node.js but more like for example in drupal you have a unified based nodejs like drupal but then bespoke parts that cover or most of the steps outlined in the recipes. meaning one project type svelt, and node server. but i was one step ahead in my mind and envisioned one suggestion stas once made for ddev config with an additional step. where you select after the project type eg drupal then the version 9, 10, 11 .. and that is what i had in mind here, the project type nodejs and the available "versions" then svelt, nodejs server and more.?

rfay:
Thanks @tyler36 !

> Any ideas how to setup debugging with VSCode?

Since this is server-side stuff, I think you'd need to use the Remote Explorer/Dev Containers extensions to actually have VS Code working inside the container. This is the Ultimike technique, as in https://www.youtube.com/watch?v=Nn_L6N4ZQqo

> I wonder, is it worth disabling the database for NodeJs projects by default? Maybe it's just me, but my javascript projects tend to be front-end API types.

IMO people will need to sort this out for themselves. I'm baffled by all of it. The DB is there, it's easy to figure out how to disable it. I did not figure out a good example that used it, although I tried. 

> How do we manage all the potential new QuickStart guides for "latest-shiny-js-framework"?

We can put lots of quickstarts in there. But inviting people to ddev.com blogs is a great path too, so that we get people exposed to more than my very low level of knowledge.

rfay:
@rpkoller If I could figure out any of that technologically I would have done it for both python and this, but in both of these worlds, the amount of variation is so far off the chart it's a whole different thing from anything in the PHP world.

andy-blum:
I'll try to test this out on my own later, but wanted to add my two cents to some of the responses so far.

> those quickstarts feel sort of complex to apply ...  i wonder would it make sense to have a node.js project type alongside with two sub types svelt and node.js webserver

As @rfay mentioned, this would scale beyond maintainability _really_ quickly. I don't know that I'd even suggest adding quickstarts for svelte. It's a nice framework and _I_ like it, but it's not really all that common in the workplace. If we wanted a JS framework example, a react project is still probably the way to go, but I think a simple node server is enough documentation as a "here's how DDEV works" for incoming JS devs without prescribing an entire quickstart sequence for specific frameworks.

> That's why our Python/Django project types failed, they're the same level of every-project-is-completely-different.

Maybe then the right idea is to have a examples for node, django, and python for this project type. Instead of getting into the frameworks of those languages, what's the most _vanilla_ example possible?

> is it worth disabling the database for NodeJs projects by default? Maybe it's just me, but my javascript projects tend to be front-end API types.

I don't think so - especially since the DB container can be easily omitted in your config file.

Having just scanned the updated docs briefly, the only thing I think I would find "lacking" (and should be followup work, not shoehorned into this PR) is that there's still no way to mirror a backend/frontend pairing.

```yml
    web_extra_exposed_ports:
    - name: node-example
      container_port: 3000
      http_port: 80
      https_port: 443

    web_extra_daemons:
    - name: "node-example"
      command: "node server.js"
      directory: /var/www/html
 ```
 
 This allows us to run our node app on port 3000 and access it at `mysite.ddev.site`, but if that's just the frontend and I want to _also_ run a decoupled backend, it's got to be exposed through a separate non-default port like `mysite.ddev.site:3001`. That's probably fine for most of the time, but it would be really nice to be able do something like this:
 
 ```yml
     web_extra_exposed_ports:
    - name: node-backend
      container_port: 3001
      http_port: 80
      https_port: 443
      subproject: api
 ```
 
 the `subproject` key here could either create a new subdomain (`api.mysite.ddev.site`), or a new project-name-prefixed domain (`mysite-api.ddev.site`).

File: docs/content/users/quickstart.md [hunk 1]
```diff
@@ -618,6 +661,80 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+      http_port: 80
+      https_port: 443
+    web_extra_daemons:
+    - name: "sveltekit-demo"
+      command: "node build"
+      directory: /var/www/html
+    EOF
+
+    ddev exec "npx sv create --template=demo --types=ts --no-add-ons . --no-add-ons --no-install"
+    # When it prompts "Directory not empty. Continue?", choose Yes.
```
stasadev, line 691:
```suggestion
    ddev exec "yes | npx sv create --template=demo --types=ts --no-add-ons --no-install ."
```

docs/tests/frankenphp.bats
```diff
@@ -0,0 +1,64 @@
...
+
+  run ddev config --project-type=drupal11 --webserver-type=generic --docroot=web --php-version=8.4
+  assert_success
+  ddev start -y
+  assert_success
+
+  cat <<'EOF' > .ddev/config.frankenphp.yaml
+web_extra_daemons:
+    - name: "frankenphp"
+      command: "frankenphp php-server --listen 0.0.0.0:80 --root=\"/var/www/html/${DDEV_DOCROOT:-}\" -v -a"
```
stasadev, line 27:
```suggestion
      command: "frankenphp php-server --listen=0.0.0.0:80 --root=${DDEV_DOCROOT:-} -v -a"
```

docs/tests/sveltekit.bats
```diff
@@ -0,0 +1,63 @@
...
+  http_port: 80
+  https_port: 443
+web_extra_daemons:
+- name: "sveltekit-demo"
+  command: "node build"
+  directory: /var/www/html
+EOF
+  assert_success
+
+  run bash -c "yes | ddev exec 'npx sv create --template=demo --types=ts --no-add-ons . --no-add-ons --no-install'"
```
stasadev, line 37:
```suggestion
  run bash -c "ddev exec 'yes | npx sv create --template=demo --types=ts --no-add-ons --no-install .'"
```

.vscode/launch.json
```diff
@@ -30,9 +30,8 @@
             "request": "launch",
             "mode": "debug",
             "program": "${workspaceRoot}/cmd/ddev",
-            "cwd": "${workspaceRoot}/../d9",
             "env": {"DDEV_DEBUG": "true"},
-            "args": ["describe"],
+            "args": ["describe", "sv2"],
```
stasadev, line 34: I'm assuming `sv2` is SvetleKit, but where does it come from? I don't see any `sv2` here.

stasadev (changes requested):
It's amazing! I'll review this PR several times because there are too many changes 🙂

We also need to add a new `type` and `webserver_type` to the `schema.json`.

docs/content/users/configuration/config.md
```diff
@@ -451,12 +451,11 @@ Defines the performance optimization mode to be used. Currently [Mutagen asynchr
...
 | Type | Default | Usage
 | -- | -- | --
-| :octicons-file-directory-16: project<br>:octicons-globe-16: global | `` | Can be `global`, `none`, `mutagen`, or `nfs`.
+| :octicons-file-directory-16: project<br>:octicons-globe-16: global | `` | Can be `global`, `none`, `mutagen`, or (deprecated) `nfs`.
 
 This is typically a global setting. The project-specific value will override global config.
 
-!!!tip "Workstation configuration is required to use NFS!"
-    See the [NFS section](../install/performance.md#nfs) on the Performance page.
+The [NFS section in the "performance" doc](../install/performance.md#performance) explains the deprecated `nfs` option.
```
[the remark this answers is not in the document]
rfay: I had `#nfs` there before and the linkchecker failed on it :(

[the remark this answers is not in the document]
stasadev: There is another anchor available, we can try this:

```suggestion
The [NFS section in the "performance" doc](../install/performance.md#filesystem-performance-nfs) explains the deprecated `nfs` option.
```

docs/content/users/quickstart.md
rfay: This is a reasonable choice, and I once had it that way, but thought that `yes` was too obsure for the average user to be comfortable with. Still, it does prevent @tyler36 's initial experience with too many questions. (As I'm sure you noted, the automated test is set up that way.)

docs/tests/frankenphp.bats
rfay: I'm interested in trying this out, but I think I liked my version better. This particular instance is in the test, which is less important than the quickstart

docs/content/users/quickstart.md
stasadev: Okay, then the suggestion changes to this:

```suggestion
    ddev exec "npx sv create --template=demo --types=ts --no-add-ons --no-install ."
    # When it prompts "Directory not empty. Continue?", choose Yes.
```

.vscode/launch.json
rfay: I was using VS Code and experimenting with a project named sv2, it needs to go away :)

```suggestion
            "args": ["describe"],
```

docs/tests/frankenphp.bats
stasadev: If explicit `/var/www/html` is better then I propose this bash syntax for empty value:

```suggestion
      command: "frankenphp php-server --listen=0.0.0.0:80 --root=\"${DDEV_DOCROOT:-/var/www/html}\" -v -a"
```

docs/content/users/quickstart.md
```diff
@@ -328,6 +328,49 @@ Read more about customizing the environment and persisting configuration in [Pro
...
+```bash
+export FRANKENPHP_SITENAME=my-frankenphp-site
+mkdir ${FRANKENPHP_SITENAME} && cd ${FRANKENPHP_SITENAME}
+ddev config --project-type=drupal11 --webserver-type=generic --docroot=web --php-version=8.4
+ddev start
+
+cat <<'EOF' > .ddev/config.frankenphp.yaml
+web_extra_daemons:
+    - name: "frankenphp"
+      command: "frankenphp php-server --listen 0.0.0.0:80 --root=\"/var/www/html/${DDEV_DOCROOT:-}\" -v -a"
```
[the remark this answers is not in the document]
stasadev: Adding alternative:

```suggestion
      command: "frankenphp php-server --listen=0.0.0.0:80 --root=\"${DDEV_DOCROOT:-/var/www/html}\" -v -a"
```

rpkoller:
one general thought about semantics. after noticing that there is a menu point "generic (frankenphp)" i havent noticed in my first review it made me think.... the page is about cms quickstarts but neither frankenphp, nor nodejs with the option sveltkit or nodejs web server are cms'es. in addition to that frankenphp uses generic in the quickstart label while nodejs is not. i wonder if it would make sense to create a new page that contains all those different web server quickstarts. that way you could start the page by explaining the generic project type and then provide the different web server quickstarts. that way things would be chopped up more clearly with a separation of concern between web servers and cmses?  because in the current state webserver and cms and the concept of generic project types are intertwined somehow?

docs/tests/frankenphp.bats
stasadev: I just realized that all my suggestions here for `--root` will not use unified path, so it will be `/var/www/html` for empty value and `web` for `web` (not `/var/www/html/web`)

Let's just unify the syntax of long flags:

```suggestion
      command: "frankenphp php-server --listen=0.0.0.0:80 --root=\"/var/www/html/${DDEV_DOCROOT:-}\" -v -a"
```

cmd/ddev/cmd/describe.go
```diff
@@ -174,7 +175,12 @@ func renderAppDescribe(app *ddevapp.DdevApp, desc map[string]interface{}) (strin
...
-				extraInfo = append(extraInfo, fmt.Sprintf("%s PHP%s\n%s\ndocroot:'%s'", desc["type"], desc["php_version"], desc["webserver_type"], desc["docroot"]))
+				// For generic type, we don't show PHP version
+				if app.WebserverType != nodeps.WebserverGeneric {
+					extraInfo = append(extraInfo, fmt.Sprintf("%s \n%s\ndocroot:'%s'", desc["type"], desc["webserver_type"], desc["docroot"]))
+				} else {
+					extraInfo = append(extraInfo, fmt.Sprintf("%s", desc["type"]))
+				}
 				extraInfo = append(extraInfo, fmt.Sprintf("Perf mode: %s", desc["performance_mode"].(string)))
 				if v, ok := desc["nodejs_version"].(string); ok {
 					extraInfo = append(extraInfo, fmt.Sprintf("NodeJS:%s", v))
```
stasadev, line 186:
```suggestion
					extraInfo = append(extraInfo, fmt.Sprintf("Node.js: %s", v))
```

docs/tests/frankenphp.bats
```diff
@@ -0,0 +1,64 @@
...
+
+  run ddev config --project-type=drupal11 --webserver-type=generic --docroot=web --php-version=8.4
+  assert_success
+  ddev start -y
+  assert_success
+
+  cat <<'EOF' > .ddev/config.frankenphp.yaml
+web_extra_daemons:
+    - name: "frankenphp"
+      command: "frankenphp php-server --listen=0.0.0.0:80 --root=\"${DDEV_DOCROOT:-/var/www/html}\" -v -a"
```
stasadev, line 27:
```suggestion
      command: "frankenphp php-server --listen=0.0.0.0:80 --root=\"/var/www/html/${DDEV_DOCROOT:-}\" -v -a"
```

docs/tests/nodejs.bats
```diff
@@ -0,0 +1,56 @@
...
+  https_port: 443
+
+web_extra_daemons:
+- name: "node-example"
+  command: "node server.js"
+  directory: /var/www/html
+EOF
+  assert_success
+
+  run ddev exec curl -s -O https://raw.githubusercontent.com/ddev/test-nodejs/refs/heads/main/server.js
```
stasadev, line 41: Make the link a bit shorter:

```suggestion
  run ddev exec curl -s -O https://raw.githubusercontent.com/ddev/test-nodejs/main/server.js
```

docs/tests/sveltekit.bats
```diff
@@ -0,0 +1,63 @@
...
+- name: "sveltekit-demo"
+  command: "node build"
+  directory: /var/www/html
+EOF
+  assert_success
+
+  run ddev exec "yes | npx sv create --template=demo --types=ts --no-add-ons --no-install ."
+  assert_success
+
+  run ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/refs/heads/main/svelte.config.js
```
stasadev, line 40:
```suggestion
  run ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/main/svelte.config.js
```

docs/tests/sveltekit.bats
```diff
@@ -0,0 +1,63 @@
...
+  assert_success
+
+  run ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/refs/heads/main/vite.config.ts
```
stasadev, line 43:
```suggestion
  run ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/main/vite.config.ts
```

docs/content/users/quickstart.md
```diff
@@ -618,6 +661,80 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    - name: "sveltekit-demo"
+      command: "node build"
+      directory: /var/www/html
+    EOF
+
+    ddev exec "npx sv create --template=demo --types=ts --no-add-ons --no-install ."
+    # When it prompts "Directory not empty. Continue?", choose Yes.
+
+    # Install an example svelte.config.js that uses adapter-node
+    ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/refs/heads/main/svelte.config.js
```
stasadev, line 694:
```suggestion
    ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/main/svelte.config.js
```

docs/content/users/quickstart.md
```diff
@@ -618,6 +661,80 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    # Install an example vite.config.ts that sets the port and allows all hostnames
+    ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/refs/heads/main/vite.config.ts
```
stasadev, line 696:
```suggestion
    ddev exec curl -s -OL https://raw.githubusercontent.com/ddev/test-sveltekit/main/vite.config.ts
```

docs/content/users/quickstart.md
```diff
@@ -618,6 +661,80 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+      http_port: 80
+      https_port: 443
+    
+    web_extra_daemons:
+    - name: "node-example"
+      command: "node server.js"
+      directory: /var/www/html
+    EOF
+
+    ddev exec curl -s -O https://raw.githubusercontent.com/ddev/test-nodejs/refs/heads/main/server.js
```
stasadev, line 731:
```suggestion
    ddev exec curl -s -O https://raw.githubusercontent.com/ddev/test-nodejs/main/server.js
```

pkg/ddevapp/app_compose_template.yaml
```diff
@@ -162,11 +162,17 @@ services:
...
-      - "{{ .DockerIP }}:$DDEV_HOST_HTTPS_PORT:443"
-    {{ if .HostMailpitPort }}
-      - "{{ .DockerIP }}:{{ .HostMailpitPort }}:8025"
-    {{ end }}
+    {{- if ne .WebserverType "generic" }}
+      - "{{ $.DockerIP }}:$DDEV_HOST_HTTP_PORT:80"
+      - "{{ $.DockerIP }}:$DDEV_HOST_HTTPS_PORT:443"
+    {{- end }}
+    {{- range .WebExtraContainerPorts }}
+      - "{{ $.DockerIP }}::{{ . }}"
```
stasadev, line 170: Nice, I didn't know that such a syntax was allowed with double `::`, where you skip the host port.

pkg/ddevapp/ddevapp.go
```diff
@@ -563,33 +561,26 @@ func (app *DdevApp) GetWebserverType() string {
...
-	if httpExpose := app.GetWebEnvVar("HTTP_EXPOSE"); httpExpose != "" {
-		//util.Debug("GetRouterHTTPPort(): HTTP_EXPOSE=%s", httpExpose)
-		httpPort := app.PortFromExposeVariable(httpExpose, "80")
-		if httpPort != "" {
-			//util.Debug("GetRouterHTTPPort(): returning httpPort=%s found from HTTP_EXPOSE=%s", httpPort, httpExpose)
-			return httpPort
-		}
+// GetPrimaryRouterHTTPPort returns app's router primary http port
+// This can be port 80 (default) or the item in HTTP_EXPOSE
+// matching the default router port. If there is no port, returns ""
```
stasadev, line 566: If there is no port, then what can happen, how DDEV will act in this case, because it is not obvious to understand?

pkg/ddevapp/ddevapp.go
```diff
@@ -3032,23 +3059,33 @@ func (app *DdevApp) GetWebContainerDirectHTTPSURL() string {
...
+		return ""
+	}
 	return fmt.Sprintf("https://%s:%d", dockerIP, port)
 }
 
-// GetWebContainerPublicPort returns the direct-access public tcp port for http
-func (app *DdevApp) GetWebContainerPublicPort() (int, error) {
-
+// GetWebContainerPublicPort returns the first direct-access public tcp port for http
+// It excludes 8025, mailpit
```
stasadev, line 3070:
```suggestion
// It excludes 8025 (mailpit) and 443 (nginx and apache)
```

pkg/ddevapp/ddevapp_test.go
```diff
@@ -2313,7 +2361,7 @@ func readLastLine(fileName string) (string, error) {
 // TestDdevFullSiteSetup tests a full import-db and import-files and then looks to see if
 // we have a spot-test success hit on a URL
 func TestDdevFullSiteSetup(t *testing.T) {
-	if runtime.GOOS == "windows" || dockerutil.IsColima() || dockerutil.IsLima() || dockerutil.IsRancherDesktop() {
+	if os.Getenv("DDEV_RUN_TEST_ANYWAY") != "true" && (runtime.GOOS == "windows" || dockerutil.IsColima() || dockerutil.IsLima() || dockerutil.IsRancherDesktop()) {
```
stasadev, line 2364: Has `DDEV_RUN_TEST_ANYWAY` been introduced for any future use?
I don't see how it can be used, as I usually run one test locally. And I don't see any usage in GitHub Actions.

stasadev (changes requested):
Add `generic` here:

https://github.com/ddev/ddev/blob/17c5324949deabe2855bad835c3a860cf6ee9d58/cmd/ddev/cmd/config.go#L232-L232

https://github.com/ddev/ddev/blob/17c5324949deabe2855bad835c3a860cf6ee9d58/docs/content/users/usage/commands.md#L358

pkg/ddevapp/ddevapp.go
rfay: If no port, returns empty string, as you point out. Caller has to do what is required.  For example, in https://github.com/rfay/ddev/blob/d7edc40b90f251604edce0e0e6f93d9d9dacce10/pkg/ddevapp/ddevapp.go#L2998-L3011 the logic is adjusted if there is no value for the returned primary port.

pkg/ddevapp/ddevapp_test.go
rfay: This is the first time I used it. I was testing using Lima, and it was really annoying. So it's intended here only for local test running. In the past when I was testing something like this I would comment out this line or equivalent line, but then it's likely to get accidentally committed.

docs/tests/frankenphp.bats
```diff
@@ -0,0 +1,64 @@
...
+setup() {
+  PROJNAME=my-frankenphp-site
+  load 'common-setup'
+  _common_setup
+}
+
+# executed after each test
+#teardown() {
+#  _common_teardown
+#}
```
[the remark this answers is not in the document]
rfay: Thanks for catching that. It was a debugging thing.

docs/content/users/quickstart.md
```diff
@@ -328,6 +328,49 @@ Read more about customizing the environment and persisting configuration in [Pro
 
     When the installation wizard prompts for database settings, enter `db` for the _DB Server Address_, _DB Name_, _DB Username_, and _DB Password_.
 
+## Generic (FrankenPHP)
+
+This example of the `webserver_type: generic` puts [FrankenPHP](https://frankenphp.dev/) into DDEV as an experimental first step in using the innovative Golang-based PHP interpreter. It is in its infancy and may someday become a full-fledged `webserver_type`. Your feedback and improvements are welcome.
```
iluuu1994, line 333: Nit: FrankenPHP is a Golang-based PHP SAPI, not an alternative PHP interpreter. It integrates the same PHP interpreter you're already using into Caddy, similar to mod_php for Apache. That's great, because it means it has pretty much full feature parity with other way of executing PHP, and will not run into weird language incompatibilities.

rfay: Thanks @iluuu1994 - Why don't you do a docs PR to improve it, thanks! I'll be interested to see what you find with this. I see it as massively slower at webserving than our fpm-based configurations, and it's not clear to me why, since everybody is so excited about FrankenPHP performance.  It's just experimental here of course, but it would be nice to understand what's up.

iluuu1994: > since everybody is so excited about FrankenPHP performance

You'll get the biggest performance improvement out of FrankenPHP's worker mode. It sacrifices the "share nothing" methodology in favor of PHP workers that stay alive for multiple requests. This avoids bootstrapping the entire framework, instantiating services, establishing db connections, etc. for every single request.

This approach isn't new per-se, it was already implemented by https://roadrunner.dev/ years ago. However, FrankenPHP bundles the webserver, PHP interpreter, most PHP extensions and the worker manager into a single binary, which greatly simplifies the setup.

> Why don't you do a docs PR to improve it, thanks!

Sorry, I don't use ddev and know very little about it, I saw this by-chance because I was sent this link by a colleague.

rfay: I was just suggesting that you click the pencil in the upper right on https://ddev.readthedocs.io/en/latest/users/quickstart/#generic-frankenphp and improve the verbiage with a trivial docs PR, since you know more about this.

PR #6940, opened 2025-01-27
fix: unset JAVA_HOME to prevent mkcert hang

rfay:
## The Issue

In very occasional cases, JAVA_HOME may be set incorrectly, leading to failure or hang of the `mkcert` command, which is used to create trusted certs. Examples in https://stackoverflow.com/questions/78865612/ddev-mkcert-install-fails-or-hangs-when-java-home-misconfigured

Disables `mkcert` if `JAVA_HOME` is incorrect.
`mkcert` fails to run if `JAVA_HOME` is set, and `$JAVA_HOME/lib/security/cacerts` or `$JAVA_HOME/jre/lib/security/cacerts` is not found.

Set JAVA_HOME to something that causes failure. Try `ddev start`, it should work with this PR, and fail with v1.24.2:

```
export JAVA_HOME=/usr
ddev start
```

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2436,6 +2436,10 @@ func (app *DdevApp) DockerEnv() {
 		dbFamily = "postgres"
 	}
 
+	// JAVA_HOME is not useful to us and can make `mkcert` fail when set wrong
+	// see https://stackoverflow.com/questions/78865612/ddev-mkcert-install-fails-or-hangs-when-java-home-misconfigured
+	_ = os.Unsetenv("JAVA_HOME")
+
 	envVars := map[string]string{
 		// The compose project name can no longer contain dots; must be lower-case
 		"COMPOSE_PROJECT_NAME":           app.GetComposeProjectName(),


stasadev (approved):
I tested this PR with `export JAVA_HOME=/usr` and confirm that it works.

At first I wasn't able to reproduce the error using v1.24.2, until I renamed my `~/.ddev` directory.

```
$ mv ~/.ddev ~/.ddev.bak
$ ddev start
Starting d11... 
Building project images....
Project images built in 2s. 
 Network ddev-d11  Created 
 Container ddev-d11-db  Created 
 Container ddev-d11-web  Created 
 Container ddev-d11-web  Started 
 Container ddev-d11-db  Started 
Failed to create certificates for project, check mkcert operation: ERROR: failed to execute "keytool -list": exit status 1

keytool error: java.lang.Exception: Keystore file does not exist: 

; err=exit status 1
```

stasadev:
So the problem is that mkcert tries to install in the java cacerts file, and fails if there is an incorrect `JAVA_HOME`:

- https://github.com/FiloSottile/mkcert/pull/38

An ignored attempt to provide users with a clear error message was made in:

- https://github.com/FiloSottile/mkcert/pull/474

If we unset `JAVA_HOME` here,  I wonder if we should recommend this in our docs:

```diff
-mkcert -install
+JAVA_HOME= mkcert -install
```

stasadev:
Based on what I found, it doesn't seem right to completely unset `JAVA_HOME` because users might still want to use Java's cacerts for mkcert. I updated the PR to show a warning and disable mkcert if the cacerts file isn't found.

This way, we ensure DDEV continues to work without HTTPS, even if the user has a broken `JAVA_HOME`.

rfay:
Thank you, and congratulations on reproducing the problem too!

rfay:
The one thing about the approach here is that we could in fact have turned off JAVA_HOME for DDEV's use, even though people might use it for other things outside of DDEV, true?

stasadev:
Yes, I think there may be some edge cases, and turning `JAVA_HOME` off completely may not be what people want.

stasadev (approved):
After another round of testing and exploring `mkcert` source code, I can say, that unsetting `JAVA_HOME` is enough, it's simple, and doesn't require any maintenance.

```
export JAVA_HOME=$(readlink -f /usr/bin/javac | sed "s:/bin/javac::")
# JAVA_HOME=/usr/lib/jvm/java-23-openjdk for me on Arch

mkcert -install
Created a new local CA 💥
The local CA is now installed in the system trust store! ⚡
The local CA is now installed in the Firefox and/or Chrome/Chromium trust store (requires browser restart)! 🦊
The local CA is already installed in Java's trust store! 👍
```

`mkcert` uses all available stores, and we simply don't need Java store in DDEV, it'll fallback to a different available store.

PR #6941, opened 2025-01-27
fix: allow `ddev config` for existing parent projects, fixes #6937

### LINKED ISSUE #6937: v1.24.2: Preventing `ddev config` in parent directory has significant side-effects
### Expected Behavior

See Drupal Slack @kevinquillen : https://drupal.slack.com/archives/C5TQRQZRR/p1737819544903659

`ddev config` should work even if a module of a Drupal project has a .ddev in it (which often happens when using the ddev/ddev-drupal-contrib add-on)

### Actual Behavior

In v1.24.2, if you do a `ddev composer install` that happens to pull in a component (like a Drupal module) which had a .ddev directory, you can't `ddev config` the project. 

The example in the Slack thread was:
```
Failed to get project(s): 'ddev config' is not allowed in /home/kevinquillen/Projects/cci because a project exists in the subdirectory /home/kevinquillen/Projects/cci/docroot/modules/contrib/svg_image_field
Run 'ddev stop --unlist' for all projects in subdirectories of the current directory first to reenable 'ddev config'
```

This is an unexpected result of the wonderful
* https://github.com/ddev/ddev/pull/6852

### Steps To Reproduce

In a Drupal 11 project:
```
ddev composer require --dev drupal/svg_image_field:dev-2.3.x
```

Then `ddev config` the project.

### Anything else?

_No response_

stasadev:
## The Issue

- #6937

If users have existing projects in a forbidden place, let them run `ddev config` there.

```
$ mkdir -p ~/workspace/parent/subfolder/child/anotherdir

# create project in the child dir
$ cd ~/workspace/parent/subfolder/child && ddev config --auto
Creating a new DDEV project config in the current directory (/home/stas/workspace/parent/subfolder/child) 
Once completed, your configuration will be written to /home/stas/workspace/parent/subfolder/child/.ddev/config.yaml
 
Configuring a 'php' project named 'child' with docroot '' at '/home/stas/workspace/parent/subfolder/child'.

# create project in the parent dir doesn't work as expected
$ cd ~/workspace/parent && ddev config --auto
Could not create a new config: a project is not allowed in /home/stas/workspace/parent because another project exists in the subdirectory /home/stas/workspace/parent/subfolder/child
Unlist this project (if it exists) with 'cd "/home/stas/workspace/parent" && ddev stop --unlist'
Or run 'ddev stop --unlist' for all projects in the subdirectories of this project directory

# create dummy config file for parent
$ mkdir -p ~/workspace/parent/.ddev && touch ~/workspace/parent/.ddev/config.yaml

# create project in the parent dir works
$ cd ~/workspace/parent && ddev config --auto
You are reconfiguring the project at /home/stas/workspace/parent.
Configuring a 'php' project named 'parent' with docroot '' at '/home/stas/workspace/parent'.

# reconfiguring in the parent/subfolder belongs to the parent project
$ cd ~/workspace/parent/subfolder && ddev config --auto
You are reconfiguring the project at /home/stas/workspace/parent.
Configuring a 'php' project named 'parent' with docroot '' at '/home/stas/workspace/parent'.

# reconfiguring in the parent/subfolder/child/anotherdir belongs to the child project
$ cd ~/workspace/parent/subfolder/child/anotherdir && ddev config --auto
You are reconfiguring the project at /home/stas/workspace/parent/subfolder/child.
Configuring a 'php' project named 'child' with docroot '' at '/home/stas/workspace/parent/subfolder/child'.

# should show both projects, parent and child
$ ddev list
```

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1418,6 +1418,10 @@ func HasAllowedLocation(app *DdevApp) error {
 		if app.AppRoot == project.AppRoot {
 			return nil
 		}
+		// If this is an existing project, allow it
+		if fileutil.FileExists(app.GetConfigPath("config.yaml")) {
+			return nil
+		}
 		// Do not allow 'ddev config' in any parent directory of any project
 		rel, err = filepath.Rel(app.AppRoot, project.AppRoot)
 		if err == nil && !strings.HasPrefix(rel, "..") {
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -1402,6 +1402,11 @@ func HasAllowedLocation(app *DdevApp) error {
 	if fileutil.FileExists(filepath.Join(app.AppRoot, "cmd/ddev/main.go")) && fileutil.FileExists(filepath.Join(app.AppRoot, "cmd/ddev_gen_autocomplete/ddev_gen_autocomplete.go")) {
 		return fmt.Errorf("a project cannot be created in the DDEV source code (%v)", app.AppRoot)
 	}
+	// If this is an existing project, allow it.
+	if fileutil.FileExists(app.GetConfigPath("config.yaml")) {
+		return nil
+	}
+	// Check all projects if they are located in the subdirectories of the project we are in.
 	projectMap := globalconfig.GetGlobalProjectList()
 	projectList := make([]*globalconfig.ProjectInfo, 0, len(projectMap))
 	for _, project := range projectMap {
@@ -1416,10 +1421,6 @@ func HasAllowedLocation(app *DdevApp) error {
 		// Without sorting, a parent directory might be matched first,
 		// causing the function to return without checking the project in the subdirectory.
 		if app.AppRoot == project.AppRoot {
-			return nil
-		}
-		// If this is an existing project, allow it
-		if fileutil.FileExists(app.GetConfigPath("config.yaml")) {
 			return nil
 		}
 		// Do not allow 'ddev config' in any parent directory of any project
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -441,25 +441,25 @@ func TestCmdConfigHasAllowedLocation(t *testing.T) {
 	// Make sure we're not allowed to config in home directory.
 	home, _ := os.UserHomeDir()
 	err = os.Chdir(home)
-	assert.NoError(err)
+	require.NoError(t, err)
 	out, err := exec.RunHostCommand(DdevBin, "config", "--project-type=php")
-	assert.Error(err)
-	assert.Contains(out, "not allowed in")
+	require.Error(t, err)
+	require.Contains(t, out, "not allowed in")
 
 	// Create a temporary directory and switch to it.
 	tmpDir := testcommon.CreateTmpDir(t.Name())
 	err = os.Chdir(tmpDir)
-	assert.NoError(err)
+	require.NoError(t, err)
 	// Create a project
 	_, err = exec.RunHostCommand(DdevBin, "config", "--project-type=php", "--project-name="+t.Name())
-	assert.NoError(err)
+	require.NoError(t, err)
 
 	t.Cleanup(func() {
 		err = os.Chdir(origDir)
 		assert.NoError(err)
 		_, err = exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name())
 		assert.NoError(err)
-		_, err = exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name()+"_subdir")
+		_, err = exec.RunHostCommand(DdevBin, "delete", "-Oy", t.Name()+"-subdir")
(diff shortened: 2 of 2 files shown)


pkg/ddevapp/config.go
```diff
@@ -1402,6 +1402,11 @@ func HasAllowedLocation(app *DdevApp) error {
 	if fileutil.FileExists(filepath.Join(app.AppRoot, "cmd/ddev/main.go")) && fileutil.FileExists(filepath.Join(app.AppRoot, "cmd/ddev_gen_autocomplete/ddev_gen_autocomplete.go")) {
 		return fmt.Errorf("a project cannot be created in the DDEV source code (%v)", app.AppRoot)
 	}
+	// If this is an existing project, allow it.
+	if fileutil.FileExists(app.GetConfigPath("config.yaml")) {
```
rfay, line 1406: I'm not sure this solves the reported problem, which **I think** is that the subdirectory project was registered in project_list.yaml, and the main project had not yet had a .ddev created. I think it's described in https://github.com/ddev/ddev/issues/6937#issuecomment-2614141399

So just looking to see if the main project has a config.yaml wouldn't solve it. If you had a registered project from the module, then `ddev config` wouldn't be allowed in the project. 

Maybe I don't fully understand?

rfay (changes requested):
Thanks for taking care of this! My take is that the situation described in https://github.com/ddev/ddev/issues/6937#issuecomment-2614141399 isn't handled by this yet?

pkg/ddevapp/config.go
stasadev: The reported case mentions:

> It's just a contributed module required by composer

So, from what I understand, the DDEV project was initially created, then a Drupal module was pulled using Composer, and `ddev config` or `ddev start` was run in the module. Everything worked fine for a while, until the upgrade to v1.24.2, which broke the project. After removing the `.ddev` folder in the module and auto cleaning `project_list.yaml`, things worked again.

The purpose of this PR is to handle a situation where you already have a Drupal project running, then you add a module (you already have a working DDEV project here, since you need `ddev composer`). If you then navigate to the module directory, where a `.ddev` directory exists, and run `ddev start`, you end up with two projects running: a parent and a child. The goal is to ensure this setup continues to work, as it did before v1.24.2.

rfay: No, in my manual testing, that was not the case. The problem was a registered project with the module's registration.

As far as I could tell in manual testing, there was no problem at all with a module that happened to have a .ddev/config.yaml in it, the problem was if there was a project registered from that.

stasadev: > a registered project with the module's registration

I don't really know how Drupal is structured, but don't you have structure like this?:

```
├── .ddev
├── web
│    └── modules
│        └── some_module
│            └── .ddev
├── sites
...
```

rfay: > I see this as there is the existing project in /home/kevinquillen/Projects/cci, otherwise how could you ever get this module /home/kevinquillen/Projects/cci/docroot/modules/contrib/svg_image_field without having a project in the parent directory.

You could get that situation by having a checked-in project that was not yet configured for DDEV (but was built by composer). 

The OP did confirm that they had a registered project related to the module, and they said it was hard to get rid of, which likely means that there were leftover dead containers showing it (and they hadn't done `ddev poweroff`)

stasadev: I agree.

So, this PR just gives you the option to configure/start a project anywhere (except `$HOME`) if you check it in and you have a `.ddev/config.yaml` file, which should cover this specific case with Drupal.

The main problem with this, which I didn't understand at first, was the intermittent results, so sometimes it was possible to run a not allowed project, as I explained in the issue https://github.com/ddev/ddev/issues/6937#issuecomment-2624589419

rfay: `fileutil.FileExists(app.GetConfigPath("config.yaml"))`

How can that file (config.yaml) not exist? If we have an app, doesn't the config.yaml have to exist?

stasadev: It doesn't exist for new "projects", i.e. places where you run `ddev config`.

If it this file doesn't exist, then we check for the parent/child logic (below this condition).

But if it exists, then we simply accept this project (the parent project will simply work, even considering that fact that it's not allowed).

I want to add that we use this check not only for `ddev config`, but for`ddev start` and so on.

rfay: `fileutil.FileExists(app.GetConfigPath("config.yaml"))` works here because `HasAllowedLocation()` is called super early in NewApp(), before anything is done filling in the app, thanks.

rfay (approved):
I tried testing this any way I could and it works correctly.

The general check is only looking for *registered* projects or projects with containers existing even if the project didn't exist any more. It does not get disturbed by *potential* projects that might be described in .ddev/config.yaml in subdirectories, so the initial worry in #6937 wasn't really valid. The OP there did in fact have a project in a subdirectory, even though they didn't know how they got it.

Thanks for working this through!

Issue #6938, opened 2025-01-27
Drag and drop to windows terminal in WSL2 does not work for `ddev import-db` or `ddev import-files`


RoelsL:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
======== Output file will be in /tmp/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/bin/ddev
/bin/ddev

These config files were loaded for project projectname: [/home/username/projects/projectname-websites/.ddev/config.yaml]
name: projectname
type: drupal9
docroot: web
php_version: 8.1
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.24.2
router_http_port: 80
router_https_port: 443
additional_hostnames: [bonap buurtslagers]
database: {mysql 8.0}
hooks: map[post-start:[map[exec:mysql -uroot -proot -e "CREATE DATABASE IF NOT EXISTS buurtslagers; GRANT ALL ON buurtslagers.* to 'db'@'%';" service:db] map[exec:mysql -uroot -proot -e "CREATE DATABASE IF NOT EXISTS bonap; GRANT ALL ON bonap.* to 'db'@'%';" service:db] map[exec-host:(ddev drush @default.local status 2>&1 | grep -q Successful) || ddev @default.local install]] pre-start:[map[exec-host:cali check-version]]]
omit_containers: [dba]
host_db_port: 49033
webimage_extra_packages: [python3 python-is-python3 build-essential]
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
disable_settings_management: true
nodejs_version: 22
default_container_timeout: 120


======== existing project customizations ========
.ddev/docker-compose.redis.yaml
.ddev/php/xdebug_port.ini


======== installed DDEV add-ons ========
No registered add-ons were found.


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /home/username/projects/projectname-websites
.ddev/mutagen/mutagen.yml not found


[...]

PR #6942, opened 2025-01-28
docs: add more examples for testing a PR

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/6938#issuecomment-2616045373

@RoelsL was not sure which binary to use for testing a PR on WSL2.

Adds an example for Linux and WSL2.

https://ddev--6942.org.readthedocs.build/en/6942/developers/building-contributing/#testing-a-pr

### DIFF
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -42,14 +42,14 @@ Download a ZIP file for your OS and architecture by clicking the link (the resul
 # Example for macOS Apple Silicon:
 unzip ddev-macos-arm64.zip
 chmod +x ddev && mv ddev ~/bin/ddev
-rm ddev-macos-arm64.zip
+rm -f ddev-macos-arm64.zip
 ```
 
 ```bash
 # Example for Linux and WSL2:
 unzip ddev-linux-amd64.zip
 chmod +x ddev && mv ddev ~/bin/ddev
-rm ddev-linux-amd64.zip
+rm -f ddev-linux-amd64.zip
 ```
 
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
Shouldn't it have at least a comment for Intel mac users or ARM64 WSL2 users telling them what to use?

stasadev:
I updated these examples with tabs, and added more examples (except Windows, which is probably not needed).

rfay (changes requested):
Thanks so much for the mentoring here!

The `chmod` is not actually required, as the x bit is set inside the zipball.

The thing that worries me the most is that ~/bin may not exist and may not be in the PATH. In most bash environments, ~/bin is added to the PATH if it exists, but only when a new terminal is opened. 

Should we say `export PATH=~/bin:$PATH` just to change the environment in this one terminal session? and `mkdir -p ~/bin` ?

I do think this is great, because people do legitimately stumble on this, and since this is developer docs, it doesn't do any tremendous harm to be verbose.

RoelsL:
The explanation about ~/bin and it being in PATH exists above the examples. 
In a normal flow of following the docs, you should have created the folder and added it to your PATH.

See: https://github.com/ddev/ddev/blob/f99695177c23d7b8d4fc47d989e184bd2250393e/docs/content/developers/building-contributing.md?plain=1#L31

stasadev:
As @RoelsL says, we already handled `PATH` some time ago, it's only a few lines above the change made here.

rfay:
Oh thanks. Sometimes context in PRs is a problem!

rfay (approved):
Thanks!

PR #6946, opened 2025-01-29
refactor: improve docker version checks, set minimum supported docker API to 1.44, fixes #6916

### LINKED ISSUE #6916: Bump the minimum required version of Docker and Docker API

Sometimes we see user reports of non-working DDEV, where the problem is that they are using a fairly old Docker.

Our current constraint:

https://github.com/ddev/ddev/blob/7dea56f86cd0feecc13621baeddba483516f0ae6/pkg/dockerutil/dockerutils.go#L1836

### Describe your solution

Bump it to something newer.

Docker supports Ubuntu Focal 20.04 (LTS)
https://docs.docker.com/engine/install/ubuntu/#os-requirements

The previous Ubuntu Bionic 18.04 has support up to Docker 24.0.2
https://download.docker.com/linux/ubuntu/dists/bionic/pool/stable/amd64/

Debian Bullseye 11 is supported too
https://docs.docker.com/engine/install/debian/#os-requirements

We have this optional check for healthcheck.start_interval that is supported from Docker v25+ in our code https://github.com/ddev/ddev/blob/7dea56f86cd0feecc13621baeddba483516f0ae6/pkg/ddevapp/utils.go#L208

Docker API matrix
https://docs.docker.com/reference/api/engine/#api-version-matrix

### Additional context

I'm not sure which version we want to see minimal, I'm only sure that people had problems with early versions of 20.10+.

AndrewGearhart:
## The Issue

- #6916

The goal is to work toward reduce template branching by improving the version checks to be looking for the version of Docker and the Docker API.

This PR updates the `DockerVersionConstraint` in `ddev/pkg/dockerutil/dockerutils.go` from `">= 20.10.0-alpha1"` to a more recent version. This ensures that DDEV operates with supported and stable versions of Docker, reducing errors caused by outdated Docker versions. Additionally, the PR updates the Docker API version checks to align with the new minimum Docker version requirements.

1. **Verify Docker Version Constraint:**
   - Ensure you have a Docker version that meets the new minimum requirement.
   - Run `ddev start` on a project.
   - Confirm that DDEV starts without errors related to Docker version compatibility.

2. **Test with Unsupported Docker Version:**
   - Temporarily downgrade Docker to a version below the new minimum requirement.
   - Run `ddev start` on a project.
   - Confirm that DDEV displays an appropriate error message indicating the Docker version is unsupported.

3. **Check Docker API Version Handling:**
   - With a supported Docker version, run `ddev start`.
   - Monitor for any warnings or errors related to Docker API version compatibility.

## Automated Testing Overview

No new automated tests have been added in this PR. The version constraint changes rely on existing logic, and validation is expected to be handled through manual testing.

## Release/Deployment Notes

- **Impact on Users:** Users running Docker versions older than the new minimum requirement will need to upgrade Docker to continue using DDEV. This may have to wait for a major version upgrade, v1.25.0?
- **Documentation:** Update the documentation to reflect the new minimum Docker and Docker API version requirements.
- **Deployment:** No special deployment steps are required beyond standard release procedures.


pkg/dockerutil/dockerutils.go
```diff
@@ -44,6 +44,8 @@ import (
 
 // NetName provides the default network name for ddev.
 const NetName = "ddev_default"
+const requiredDockerVersion = ">= 24.0.0"
```
stasadev, line 47: We already have a variable for this in the same file, `DockerVersionConstraint`

pkg/dockerutil/dockerutils.go
```diff
@@ -44,6 +44,8 @@ import (
...
+const requiredDockerAPIVersion = ">= 1.43"
```
stasadev, line 48: This could be named `minimalDockerAPIVersion = "1.43"`, and built-in check from Docker library can be used for this (see `templateCanUse` func).

pkg/dockerutil/dockerutils.go
```diff
@@ -839,6 +841,35 @@ func CheckDockerVersion(versionConstraint string) error {
...
+			return apiErrs[0]
+		}
+
+		apiMsgs := "\n"
+		for _, err := range apiErrs {
+			apiMsgs = fmt.Sprint(apiMsgs, err, "\n")
+		}
+		return fmt.Errorf("%s", apiMsgs)
+	}
+
```
stasadev, line 872: This can be replaced by:

```go
import (
	...
	dockerVersions "github.com/docker/docker/api/types/versions"
	...
)

...

dockerAPIVersion, err := GetDockerAPIVersion()
if err != nil {
	return fmt.Errorf("unable to get Docker API version: %v", err)
}
if !dockerVersions.GreaterThanOrEqualTo(dockerAPIVersion, minimalDockerAPIVersion) {
	return fmt.Errorf("You %s doesn't match minimum Docker API version %s", dockerAPIVersion, minimalDockerAPIVersion)
}
return nil
```

rfay (changes requested):
Let's make sure to add a link in the docker error message that explains how to use the current version of Docker (and make sure the doc is adjusted to make that easy). This would make it a little less cumbersome if we decide to do this in a minor version upgrade.

AndrewGearhart:
@rfay @stasadev I've updated the docs to include a notation about how to install Docker on Linux distros, and a troubleshooting bullet at the end of that page. I included a link in the error message for when the user doesn't have a modern version installed. Let me know if this isn't what you had in mind!

pkg/dockerutil/dockerutils.go
```diff
@@ -45,6 +46,15 @@ import (
 // NetName provides the default network name for ddev.
 const NetName = "ddev_default"
 
+// DockerVersionConstraint is the current minimum version of Docker required for DDEV.
+// See https://godoc.org/github.com/Masterminds/semver#hdr-Checking_Version_Constraints
+// for examples defining version constraints.
+// REMEMBER TO CHANGE docs/ddev-installation.md if you touch this!
+// The constraint MUST HAVE a -pre of some kind on it for successful comparison.
+// See https://github.com/ddev/ddev/pull/738 and regression https://github.com/ddev/ddev/issues/1431
+const DockerVersionConstraint = ">= 24.0.0"
```
stasadev, line 55: As suggested in the comment above this line:

```suggestion
const DockerVersionConstraint = ">= 24.0.0-alpha1"
```

docs/content/users/install/docker-installation.md
```diff
@@ -159,3 +161,4 @@ If you get an error or don’t see the contents of your project directory, you
 * If you’re seeing “The path (...) is not shared and is not known to Docker”, find *File sharing* in your Docker settings make sure the appropriate path or drive is included.
 * “Error response from daemon: Get registry-1.docker.io/v2/” may mean Docker isn’t running or you don’t have internet access. Try starting or restarting Docker, and confirm you have a working internet connection.
 * If you’re seeing “403 authentication required” trying to [`ddev start`](../usage/commands.md#start), run `docker logout` and try again. Docker authentication is *not* required for any normal DDEV action.
+* If you see the error message "Your version of Docker API (A.B.C) doesn't meet the minimum (X.Y.Z)," you need to upgrade your Docker engine. The best way to do this is by updating Docker from the official Docker repository. If you're using a version from your distribution's default package repository, it's likely outdated. In that case, uninstall Docker Engine and follow the installation instructions above to install a supported version from the official Docker repository.
```
stasadev, line 164: Docker API always has two letters, so it should be `A.B` and `X.Y`, but I'm not sure if people are going to read/see this message. See my another comment.

pkg/dockerutil/dockerutils.go
```diff
@@ -839,6 +849,14 @@ func CheckDockerVersion(versionConstraint string) error {
 		}
 		return fmt.Errorf("%s", msgs)
 	}
+	
+	dockerAPIVersion, err := GetDockerAPIVersion()
+	if err != nil {
+		return fmt.Errorf("unable to determine Docker API version: %v", err)
+	}
+	if !dockerVersions.GreaterThanOrEqualTo(dockerAPIVersion, minimalDockerAPIVersion) {
+		return fmt.Errorf("your version of Docker API (%s) doesn't meet the minimum (%s) - see (%s) for more information", dockerAPIVersion, minimalDockerAPIVersion, "https://ddev.rtfd.io/page/users/install/docker-installation/#testing-and-troubleshooting-your-docker-installation")
```
stasadev, line 858: I think we can simply add a link to Docker API matrix table https://docs.docker.com/reference/api/engine/#api-version-matrix

Because people won't understand what version is required if they don't see that table.

stasadev (changes requested):
Thanks for the changes, I think we're almost done here, but I have some suggestions.

rfay:
Rebased. When you think everything has been addressed and this is ready for final review, set it to "Ready for review".

docs/content/users/install/docker-installation.md
AndrewGearhart: I updated to use two number semver example.

rfay:
I let the tests run, will try to pay attention to allow them again if I see more commits. (First-time PR submitters have to have the tests approved for security reasons. After the first time it's not an issue.)

rfay:
* You have a typo "respository"
* You have a misformatted file
> https://github.com/ddev/ddev/actions/runs/13253712403/job/36996937393?pr=6946

Fix with `gofmt -w -s pkg/dockerutil/dockerutils.go`

rfay:
Congrats on solving the static validation and docs problems!

pkg/dockerutil/dockerutils.go
```diff
@@ -839,6 +849,14 @@ func CheckDockerVersion(versionConstraint string) error {
 		}
 		return fmt.Errorf("%s", msgs)
 	}
+
+	dockerAPIVersion, err := GetDockerAPIVersion()
+	if err != nil {
+		return fmt.Errorf("unable to determine Docker API version: %v", err)
+	}
+	if !dockerVersions.GreaterThanOrEqualTo(dockerAPIVersion, minimalDockerAPIVersion) {
+		return fmt.Errorf("your Docker API version (%s) doesn't meet the minimum (%s) - see %s for version matrix (%s) for more help", dockerAPIVersion, minimalDockerAPIVersion, "https://docs.docker.com/reference/api/engine/#api-version-matrix", "https://ddev.rtfd.io/page/users/install/docker-installation/#testing-and-troubleshooting-your-docker-installation")
```
stasadev, line 858: Right now we have an optional check for `"1.44"` in our code, to modify template for the app:

https://github.com/ddev/ddev/blob/30fb117cb664273b085477edaacb76ed25166c5a/pkg/ddevapp/app_compose_template.yaml#L85-L87

https://github.com/ddev/ddev/blob/30fb117cb664273b085477edaacb76ed25166c5a/pkg/ddevapp/utils.go#L198-L211

If we change
```diff
-const minimalDockerAPIVersion = "1.43"
+const minimalDockerAPIVersion = "1.44"
```
and remove `templateCanUse`, making `start_interval: "1s"` available all the time, it will break `ddev start` for older Docker, because it'll simply show a warning, but it should fail completely.

---

So it should call `util.Failed` instead of returning an error.

stasadev (changes requested):
I ran `ddev start` with modified constraints (to actually test it):

```
const DockerVersionConstraint = ">= 24.0.0-alpha1"
const minimalDockerAPIVersion = "1.43"
```

I changed it to:
```diff
-const DockerVersionConstraint = ">= 24.0.0-alpha1"
+const DockerVersionConstraint = ">= 28.0.0-alpha1"
const minimalDockerAPIVersion = "1.43"
```

to see a warning:
```
$ ddev start
The Docker version currently installed does not seem to meet DDEV's requirements: 27.5.1 is less than 28.0.0-alpha1
...
```

And another check

```diff
const DockerVersionConstraint = ">= 24.0.0-alpha1"
-const minimalDockerAPIVersion = "1.43"
+const minimalDockerAPIVersion = "1.55"
```

```
$ ddev start
The Docker version currently installed does not seem to meet DDEV's requirements: your Docker API version (1.47) doesn't meet the minimum (1.55) - see https://docs.docker.com/reference/api/engine/#api-version-matrix for version matrix (https://ddev.rtfd.io/page/users/install/docker-installation/#testing-and-troubleshooting-your-docker-installation) for more help
...
```

I agree that Docker version check is a "recommendation" (because you can override it to whatever you want if you build `docker` yourself).

But Docker API version check should be a hard check, there is no point in showing a warning, if our code has something that requires newer API.

pkg/dockerutil/dockerutils.go
rfay: But didn't we decide that we can't go to 1.44 because we don't want to break people?

pkg/dockerutil/dockerutils.go
```diff
@@ -839,6 +849,14 @@ func CheckDockerVersion(versionConstraint string) error {
 		}
 		return fmt.Errorf("%s", msgs)
 	}
+
+	dockerAPIVersion, err := GetDockerAPIVersion()
+	if err != nil {
+		return fmt.Errorf("unable to determine Docker API version: %v", err)
```
stasadev, line 855: And if we can't determine the API version at all, I think it should fail too.

pkg/dockerutil/dockerutils.go
stasadev: Yes, but I'm thinking ahead 😅

When we decide to increase the `minimalDockerAPIVersion`, it'll simply show a warning for people with lower API version, it won't fail, which mean `ddev start` will continue to run, failing in another place.

Is it enough to show a warning in this case?

stasadev: You can change the `Version` shown in the `docker version` output to anything you want by specifying it as a build argument (which is why I believe we show it as a warning). Different Linux distributions can bundle Docker using various techniques, making this value unreliable.

However, the `API version` in `docker version` cannot be altered through build arguments, so it's something we can rely on. That's why I suggest we implement a hard failure if the API version doesn't meet the requirements.

stasadev:
@AndrewGearhart, do you mind if I experiment with this?

I think we should do a hard `API Version` check, but only on `ddev start`.

AndrewGearhart:
> @AndrewGearhart, do you mind if I experiment with this?
> 
> I think we should do a hard `API Version` check, but only on `ddev start`.

Sure, I'm okay with that. :) Thanks for checking.

stasadev:
I pushed a new commit:

- fail on `ddev start` if Docker API version is not compatible
- show Docker API version in `ddev debug dockercheck`
- small refactoring

stasadev (approved):
I tested it again with a fresh perspective, and I can confirm that it works.

As for failing on `ddev start` if the Docker API version doesn't match the required minimum, it's definitely a stricter approach, but I think it's for the best as it ensures everyone stays updated.

rfay:
What was your technique for using old docker? Or did you mock it?

stasadev:
I mocked it by changing the constraints.

rfay:
Manually tested on 
* Ubuntu 22.04 with docker.io package, worked fine. docker 26.1.3 with api 1.45
* Debian 11 with docker.io 20.10.5+dfsg1 I first get the warning `The Docker version currently installed does not seem to meet DDEV's requirements: 20.10.5+dfsg1 is less than 24.0.0-alpha1` and then I get `Unable to start project d11: your Docker API version (1.41) doesn't meet the required minimum (1.43)`

pkg/ddevapp/ddevapp.go
```diff
@@ -1243,6 +1243,9 @@ func (app *DdevApp) GetDBImage() string {
 // Start initiates docker-compose up
 func (app *DdevApp) Start() error {
 	var err error
+	if _, err = dockerutil.HasCompatibleDockerAPIVersion(dockerutil.DockerAPIVersionMinimum); err != nil {
+		util.Failed("Unable to start project %s: %s", app.GetName(), err)
```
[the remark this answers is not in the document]
rfay:
```suggestion
		util.Failed("Unable to start project %s: %v", app.GetName(), err)
```

Let's use `%v` for the replacement, since we never know what type `err` is.

pkg/dockerutil/dockerutils.go
```diff
@@ -45,6 +46,19 @@ import (
 // NetName provides the default network name for ddev.
 const NetName = "ddev_default"
 
+// DockerVersionConstraint is the current minimum version of Docker recommended for DDEV.
```
rfay, line 49:
```suggestion
// DockerVersionConstraint specifies the Docker server version required for DDEV.
```

pkg/dockerutil/dockerutils.go
```diff
@@ -45,6 +46,19 @@ import (
...
+// If this constraint is not met, we will warn the user, but not fail.
```
rfay, line 50: This isn't true now is it?

```suggestion
```

pkg/dockerutil/dockerutils.go
```diff
@@ -842,6 +856,26 @@ func CheckDockerVersion(versionConstraint string) error {
 	return nil
 }
 
+// HasCompatibleDockerAPIVersion checks if the Docker API version of the host system is compatible with the required minimum version.
+func HasCompatibleDockerAPIVersion(minimumAPIVersion string) (bool, error) {
+	const (
+		versionMatrixURL   = "https://docs.docker.com/reference/api/engine/#api-version-matrix"
+		troubleshootingURL = "https://ddev.readthedocs.io/en/stable/users/install/docker-installation/#testing-and-troubleshooting-your-docker-installation"
+		helpfulLinks       = "\nSee API version matrix at %s\nTroubleshoot at %s"
```
rfay, line 864: After checking several failures I'm not entirely sure that this link is helpful, but it's fine. This should be an unusual situation.

rfay (approved):
This seems fine to me. I offer a couple of nits.

Overall, it doesn't look to me like we're actually getting value in the long term by checking both Docker version and Docker API, it was a good idea, but in the practical world they seem to track exactly.

Thanks for all the hard work on this!

pkg/dockerutil/dockerutils.go
stasadev: Strictly speaking, we don't need a `DockerVersionConstraint`, `DockerAPIVersionMinimum` should be enough.

It's not required now, see my comment below.

pkg/dockerutil/dockerutils.go
stasadev: It's a warning, not an error:

https://github.com/ddev/ddev/blob/e6bf644d6c93216a4fa754b527b169fcd95920fa/cmd/ddev/cmd/root.go#L64-L73

We need to either remove this check from `root.go` or leave the comment as is.

pkg/dockerutil/dockerutils.go
rfay: It's true that we're probably doing too much in too many places.

pkg/dockerutil/dockerutils.go
stasadev: Do you mean `versionMatrixURL` or `troubleshootingURL` ?

pkg/dockerutil/dockerutils.go
stasadev: Let's remove `DockerVersionConstraint`, one check on `ddev start` should be enough.

pkg/dockerutil/dockerutils.go
rfay: The helpfulLinks with the API version matrix. Basically in our situation people don't need a matrix, they just need to update. And.. as discussed elsewhere, it's going to catch the Docker server version anyway.

stasadev:
I added a new commit: it will show a warning (not an error) in readable format for the user:

```
$ ddev start
Problem with your Docker provider: installed Docker version 24.0.2 is not supported, please update to version 25.0 or newer.
...
```

Under the hood, the check will compare the Docker API version instead of the Docker version.

What version can be considered outdated?

Here is Docker Engine EOL list https://endoflife.date/docker-engine If the version doesn't receive any updates, it's EOL. Using that list I see that 24.0 is not supported, and 23.0 will be discontinued in 3 months.

So I decided that we can recommend at least Docker 25.0 (API 1.44) and focus on that list in the future.

cmd/ddev/cmd/root.go
```diff
@@ -61,14 +61,12 @@ Support: https://ddev.readthedocs.io/en/stable/users/support/`,
 			return
 		}
 
-		err := dockerutil.CheckDockerVersion(dockerutil.DockerVersionConstraint)
+		err := dockerutil.CheckDockerVersion(dockerutil.DockerRequirements)
 		if err != nil {
 			if err.Error() == "no docker" {
-				if os.Args[1] != "version" {
```
stasadev, line 67: This check is not needed because of this condition right above:

```go
if command != "start" && command != "restart" {
	return
}
```

rfay:
> So I decided that we can recommend at least Docker 25.0 (API 1.44) and focus on that list in the future.

I can go with that, especially since we're going to warn instead of fail.

rfay:
```
rfay@ub-1804:~/workspace/d11$ ddev start
Problem with your Docker provider: installed Docker version 20.10.21 is not supported, please update to version 25.0 or newer.
```

pkg/dockerutil/dockerutils.go
```diff
@@ -45,6 +46,22 @@ import (
 // NetName provides the default network name for ddev.
 const NetName = "ddev_default"
 
+type DockerVersionMatrix struct {
+	APIVersion string
+	Version    string
+}
+
+// DockerRequirements defines the minimum Docker version required by DDEV.
```
rfay, line 54: This is a pretty nice approach in the end, thanks.

pkg/dockerutil/dockerutils.go
```diff
@@ -793,21 +810,18 @@ func GetContainerEnv(key string, container dockerTypes.Container) string {
...
 	if err != nil {
 		return fmt.Errorf("no docker")
 	}
-	// If Docker version has "_ce", remove it. This happens on OpenSUSE Tumbleweed at least
-	currentVersion = strings.TrimSuffix(currentVersion, "_ce")
-	dockerVersion, err := semver.NewVersion(currentVersion)
+	currentAPIVersion, err := GetDockerAPIVersion()
 	if err != nil {
-		return err
+		return fmt.Errorf("no docker")
```
rfay, line 824: This could certainly be improved with an actual error type, but no need to change. It's been this way a long time. But we should really be doing fancier.

pkg/dockerutil/dockerutils.go
```diff
@@ -822,22 +836,8 @@ func CheckDockerVersion(versionConstraint string) error {
...
-		if len(errs) <= 1 {
-			return errs[0]
-		}
-
-		msgs := "\n"
-		for _, err := range errs {
-			msgs = fmt.Sprint(msgs, err, "\n")
-		}
-		return fmt.Errorf("%s", msgs)
+	if !dockerVersions.GreaterThanOrEqualTo(currentAPIVersion, dockerVersionMatrix.APIVersion) {
```
rfay, line 839: This stanza is fine, but it's not obvious in the code what's going on. We're checking the API version, but then recommending the Docker version.

Suggestion below may not work right because of inability to insert tab here.

```suggestion
	// Check against recommended API version, if it fails, suggest the minimum Docker version that relates to supported API
	if !dockerVersions.GreaterThanOrEqualTo(currentAPIVersion, dockerVersionMatrix.APIVersion) {
```

rfay (approved):
I think it's great. Minor comments and maybe a suggestion. Works nicely in many environments, and doesn't break people, warns instead. Thanks for all the work on this, both of you. I'm sorry we started out with different requirements than we ended up with, but am pretty happy with the result. We aren't forcing people, but we're giving a clear indication of where they're at.  Also happy about the simplification of check-api-but-suggest-docker-version.

pkg/dockerutil/dockerutils.go
stasadev: Thanks, I do see a tab here (I'm using the [Refined GitHub](https://github.com/refined-github/refined-github) browser extension, which shows tabs and spaces in PRs and adds many other small, fancy features).

PR #6950, opened 2025-01-30
feat: bind IP to user-defined ports if not set, fixes #6943

### LINKED ISSUE #6943: Make `ports` in `docker-compose.*.yaml` respect `bind-all-interfaces=false` project option

Some add-ons like https://github.com/ddev/ddev-sqlsrv/blob/main/docker-compose.sqlsrv.yaml#L19 use `ports` directly.

But it's exposed on `0.0.0.0`, instead of `127.0.0.1`, which is default in DDEV.

To reproduce:

```
mkdir ports-test && cd ports-test
ddev config --bind-all-interfaces=false
printf "services:\n  web:\n    ports:\n      - 12345:12345\n" > .ddev/docker-compose.web.yaml
ddev start
```

```yaml
services:
  web:
    ports:
      - 12345:12345
```

Look for `ports` configuration in `.ddev/.ddev-docker-compose-full.yaml`:

```yaml
ports:
    - host_ip: 127.0.0.1
        mode: ingress
        protocol: tcp
        target: 80
    - host_ip: 127.0.0.1
        mode: ingress
        protocol: tcp
        target: 443
    - mode: ingress
        protocol: tcp
        published: "12345"
        target: 12345
```

And check the ports, I did this on Linux:

```
$ sudo netstat -tulpn | grep LISTEN | grep docker-proxy
tcp        0      0 127.0.0.1:10999         0.0.0.0:*               LISTEN      157680/docker-proxy 
tcp        0      0 127.0.0.1:443           0.0.0.0:*               LISTEN      157656/docker-proxy 
tcp        0      0 127.0.0.1:32798         0.0.0.0:*               LISTEN      154264/docker-proxy 
tcp        0      0 127.0.0.1:32799         0.0.0.0:*               LISTEN      154302/docker-proxy 
tcp        0      0 127.0.0.1:32797         0.0.0.0:*               LISTEN      154256/docker-proxy 
tcp        0      0 127.0.0.1:80            0.0.0.0:*               LISTEN      157649/docker-proxy 
tcp        0      0 0.0.0.0:12345           0.0.0.0:*               LISTEN      154229/docker-proxy 
tcp        0      0 127.0.0.1:8026          0.0.0.0:*               LISTEN      157671/docker-proxy 
tcp        0      0 127.0.0.1:8025          0.0.0.0:*               LISTEN      157665/docker-proxy 
tcp6       0      0 :::12345                :::*                    LISTEN      154245/docker-proxy
```

Specifically:
```
tcp        0      0 0.0.0.0:12345           0.0.0.0:*               LISTEN      154229/docker-proxy 
tcp6       0      0 :::12345                :::*                    LISTEN      154245/docker-proxy
```

### Describe your solution

Make it respect `ddev config --bind-all-interfaces=false`

The fix can be applied to `ddevapp.fixupComposeYaml()` for each `services[][ports]`.

### Describe alternatives

Manually add `127.0.0.1`:

```yaml
services:
  web:
    ports:
      - 127.0.0.1:12345:12345
```

### Additional context

_No response_

stasadev:
## The Issue

- #6943

Adds Docker IP to the port, but only if the IP is not already set, making it a non-breaking change.

- [x] Add a test

Default config should use `127.0.0.1`:

```
ddev config --bind-all-interfaces=false # default value
printf "services:\n  web:\n    ports:\n      - 12345:12345\n" > .ddev/docker-compose.web.yaml
ddev start

cat .ddev/.ddev-docker-compose-full.yaml | grep -B3 12345
            - host_ip: 127.0.0.1
              mode: ingress
              protocol: tcp
              published: "12345"
              target: 12345
```

`bind-all-interfaces` uses `0.0.0.0`:

```
ddev config --bind-all-interfaces=true
printf "services:\n  web:\n    ports:\n      - 12345:12345\n" > .ddev/docker-compose.web.yaml
ddev start

cat .ddev/.ddev-docker-compose-full.yaml | grep -B3 12345
            - host_ip: 0.0.0.0
              mode: ingress
              protocol: tcp
              published: "12345"
              target: 12345
```

Use default config, but set `0.0.0.0` for this port explicitly, DDEV should respect the user's choice:

```
ddev config --bind-all-interfaces=false
printf "services:\n  web:\n    ports:\n      - 0.0.0.0:12345:12345\n" > .ddev/docker-compose.web.yaml
ddev start

cat .ddev/.ddev-docker-compose-full.yaml | grep -B3 12345
            - host_ip: 0.0.0.0
              mode: ingress
              protocol: tcp
              published: "12345"
              target: 12345
```

### DIFF
--- a/pkg/ddevapp/compose_yaml.go
+++ b/pkg/ddevapp/compose_yaml.go
@@ -102,6 +102,11 @@ func fixupComposeYaml(yamlStr string, app *DdevApp) (map[string]interface{}, err
 		return nil, err
 	}
 
+	dockerIP, _ := dockerutil.GetDockerIP()
+	if app.BindAllInterfaces {
+		dockerIP = "0.0.0.0"
+	}
+
 	// Ensure that some important network properties are not overridden by users
 	for name, network := range tempMap["networks"].(map[string]interface{}) {
 		if network == nil {
@@ -171,6 +176,24 @@ func fixupComposeYaml(yamlStr string, app *DdevApp) (map[string]interface{}, err
 							// The same thing is done in `docker-compose config`
 							// See https://github.com/docker/compose/blob/361c0893a9e16d54f535cdb2e764362363d40702/cmd/compose/config.go#L405-L409
 							environmentMap[envKey] = strings.ReplaceAll(envValue, `$`, `$$`)
+						}
+					}
+				}
+			}
+		}
+		// Assign the host_ip for each port if it's not already set.
+		// This is needed for custom-defined user ports. For example:
+		// ports:
+		//   - 3000:3000
+		// Without this, Docker doesn't add a Docker IP, like this:
+		// ports:
+		//   - 127.0.0.1:3000:3000
+		if dockerIP != "" {
+			if ports, ok := serviceMap["ports"].([]interface{}); ok {
+				for _, port := range ports {
+					if portMap, ok := port.(map[string]interface{}); ok {
+						if _, exists := portMap["host_ip"]; !exists {
+							portMap["host_ip"] = dockerIP
 						}
 					}
 				}
--- a/pkg/ddevapp/compose_yaml.go
+++ b/pkg/ddevapp/compose_yaml.go
@@ -102,9 +102,9 @@ func fixupComposeYaml(yamlStr string, app *DdevApp) (map[string]interface{}, err
 		return nil, err
 	}
 
-	dockerIP, _ := dockerutil.GetDockerIP()
+	bindIP, _ := dockerutil.GetDockerIP()
 	if app.BindAllInterfaces {
-		dockerIP = "0.0.0.0"
+		bindIP = "0.0.0.0"
 	}
 
 	// Ensure that some important network properties are not overridden by users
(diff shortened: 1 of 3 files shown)


pkg/ddevapp/compose_yaml.go
```diff
@@ -102,6 +102,11 @@ func fixupComposeYaml(yamlStr string, app *DdevApp) (map[string]interface{}, err
 		return nil, err
 	}
 
+	dockerIP, _ := dockerutil.GetDockerIP()
```
rfay, line 105: nit here: IMO it would be better to use a variable name like `bindIP` here, to describe what it's actual purpose is.

pkg/ddevapp/ddevapp_test.go
```diff
@@ -3517,13 +3517,19 @@ func TestMultipleComposeFiles(t *testing.T) {
...
 	require.Equal(t, app.GetConfigPath("docker-compose.override.yaml"), files[len(files)-1])
 
 	require.NotEmpty(t, app.ComposeYaml)
 	require.True(t, len(app.ComposeYaml) > 0)
 
+	dockerIP, err := dockerutil.GetDockerIP()
+	require.NoError(t, err)
+	if app.BindAllInterfaces {
+		dockerIP = "0.0.0.0"
+	}
```
rfay, line 3531: Is the test covering the default situation, with regular 127.0.0.1?

rfay (approved):
Trivial comments. If you're fine with this after manual testing and consideration of the comments, it can go in.

pkg/ddevapp/ddevapp_test.go
```diff
@@ -3536,24 +3542,51 @@ func TestMultipleComposeFiles(t *testing.T) {
 				assert.Equal("2", env["DUMMY_COMPOSE_TWO"])
 				assert.Equal("override", env["DUMMY_COMPOSE_OVERRIDE"])
 			} else {
-				t.Error("Failed to parse environment")
+				t.Error("Failed to parse web environment")
```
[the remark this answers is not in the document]
stasadev: I didn't stumble on this, but it's a good idea, I will add it.

pkg/ddevapp/ddevapp_test.go
stasadev: Yes, you are right. I should check for `127.0.0.1` in the test, not use `dockerutil.GetDockerIP()` here.

rfay:
I doubt that there are automated tests in the add-ons that will warn us about problems with this, but we'll keep our eyes open.

PR #6860, opened 2024-12-26
test: Use linkspector instead of markdown-link-check, fixes #6859

rfay:
## The Issue

* https://github.com/ddev/ddev/issues/6859

Per https://github.com/gaurav-nelson/github-action-markdown-link-check new work is going on in https://github.com/UmbrellaDocs/linkspector so we should try using that instead.

PR #6951, opened 2025-01-31
fix: use linkspector for external links, use mkdocs for internal links, for #6860

stasadev:
## The Issue

We switched to `linkspector` in:

- #6860

And there is some strange behavior for `linkspector` in #6935:

`Running linkspector with reviewdog 🐶 ...`
`message:"Cannot reach ../install/performance.md#filesystem-performance-nfs. Status: 404 Cannot find section: #filesystem-performance-nfs in file: /home/runner/work/ddev/ddev/docs/content/users/install/performance.md." location:{path:"docs/content/users/configuration/config.md" range:{start:{line:458 column:5} end:{line:458 column:97}}} severity:ERROR source:{name:"linkspector" url:"https://github.com/UmbrellaDocs/linkspector"}`

The link exists and works, but `linkspector` doesn't understand that.

- Uses linkspector only for external links (use `filter_mode: added` for PRs and `filter_mode: nofilter` on main branch).
- Uses built-in mkdocs check for internal links

I'll add broken links here to see if all checks are working fine.

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -8,6 +8,7 @@ on:
       - "Makefile"
       - "vendor/**"
       - ".github/workflows/**"
+      - "!.github/workflows/docs*"
   push:
     branches: [ main, master ]
   workflow_dispatch:
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -8,7 +8,7 @@ on:
       - "Makefile"
       - "vendor/**"
       - ".github/workflows/**"
-      - "!.github/workflows/docs*"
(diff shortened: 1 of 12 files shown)


rfay:
Changes here are running all the tests, could we fix that in this PR so full tests don't run when just these files and docs are touched?

Thanks for chasing this down! Let's make sure we make sure that fragment link failures still fail.

stasadev:
> Changes here are running all the tests, could we fix that in this PR so full tests don't run when just these files and docs are touched?

I can't check if my exclusions are really working because I'm changing other workflow files here.

I will check if it works in `ddev-test`.

rfay (approved):
Thanks!

PR #6952, opened 2025-01-31
build: hide extra output from linkspector [skip ci]

stasadev:

Before:

```
$ make linkspector
linkspector: 
if command -v linkspector >/dev/null 2>&1; then \
        linkspector check; \
else \
        echo "Not running linkspector because it's not installed"; \
fi
⠹ Checking docs/content/index.md...
```

After:

```
$ make linkspector
linkspector: 
⠼ Checking docs/content/index.md...
```


rfay (approved):
Thanks!

PR #6953, opened 2025-01-31
build: run linkspector on the main branch, for #6860

stasadev:
## The Issue

- #6860

I noticed that linkspector doesn't run for main branch:
https://github.com/ddev/ddev/actions/runs/13079915327/job/36500915400
```
 Running linkspector with reviewdog :dog: ...
  reviewdog: this is not PullRequest build.
```

Changes the reporter to make it check for broken links on the main branch as well.

I tested it on `ddev-test`.

In addition, I confirmed that `filter_mode: ${{ github.event_name == 'pull_request' && 'added' || 'nofilter' }}` allows us to check only for changed links.

This means that if some link somewhere in the project is broken, and PR didn't touch it, then it won't be reported (the error will be shown only on the main branch build).

### DIFF
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -53,7 +53,7 @@ jobs:
         uses: umbrelladocs/action-linkspector@v1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks!

PR #6954, opened 2025-01-31
build: run golangci-lint only for changed code in PRs

stasadev:
## The Issue

Sometimes `golangci-lint` introduces new rules that break our tests. This affects the people who contribute to DDEV because these errors show up in their PRs. Usually, we deal with this by doing a hotfix PR, merging it, and then rebasing other people's PRs.

Runs `golangci-lint` in the PR only for the actually modified code, the full check will still be done on the main branch.

I got the idea from #6953, where I tested the same thing for `linkspector`.

I tested this PR on `ddev-test`. It works!

Log from this PR https://github.com/ddev/ddev/actions/runs/13081600537/job/36506164694?pr=6954

```
run golangci-lint
  only new issues on pull_request: /tmp/tmp-1924-DYZkqY0KDaC0/pull.patch
```

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -50,6 +50,8 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: latest
+          # Check code only for the changed lines for PRs
+          only-new-issues: ${{ github.event_name == 'pull_request' && true || false }}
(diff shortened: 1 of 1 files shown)


rfay (approved):
Golangci-lint doesn't really cost much time to run, but I'm fine with this.

PR #6962, opened 2025-02-05
feat: normalize docroot usage in `ddev config`

stasadev:
## The Issue

```
# set docroot to some value
ddev config --docroot=web

# try to unset it with this command, which won't work
ddev config --docroot=""

# works, but sets docroot to dot, not empty string
ddev config --docroot=.

# and try absolute path, which should not be possible (it fails, but the error should not come from os.Mkdir)
ddev config --docroot="/test"
Could not create docroot at /test: the provided docroot /test is outside the project root

# and this one works, but it shouldn't work
ddev config --docroot="/tmp"
```

- Normalizes docroot in `CreateDocroot` function.
- Accepts quoted input for docroot in `ddev config` (because we accept path here).
- Adds a test.

```
ddev config --docroot=web
ddev debug configyaml | grep docroot # docroot: web

ddev config --docroot=""
ddev debug configyaml | grep docroot # should return nothing

ddev config --docroot="."
ddev debug configyaml | grep docroot # should return nothing

ddev config --docroot="./"
ddev debug configyaml | grep docroot # should return nothing

ddev config --docroot="/test"
Could not create docroot at /test: the provided docroot /test is outside the project root

ddev config --docroot="/tmp"
Could not create docroot at /tmp: the provided docroot /tmp is outside the project root

# repeat the above with interactive input for docroot:
ddev config
```

### DIFF
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -386,7 +386,7 @@ func TestConfigSetValues(t *testing.T) {
 		{"empty docroot", "", "", ""},
 		{"dot docroot", ".", "", ""},
 		{"fail for outside approot", "..", "", "is outside the project root"},
-		{"fail for absolute path outside approot", "/test", "", "is outside the project root"},
+		{"fail for absolute path outside approot", "//test", "", "is outside the project root"},
 		{"ok for absolute path inside approot", tmpDir, "", ""},
 		{"dot with slash docroot", "./", "", ""},
 		{"dot with slash and dir docroot", "./test", "test", ""},
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -375,6 +375,51 @@ func TestConfigSetValues(t *testing.T) {
 	assert.Equal("FOO=bar", app.WebEnvironment[1])
 	assert.Equal("SPACES=with spaces", app.WebEnvironment[3])
 	assert.Equal(webEnv, app.WebEnvironment[2])
+
+	// test docroot locations
+	testMatrix := []struct {
+		description string
+		input       string
+		expected    string
+		error       string
+	}{
+		{"empty docroot", "", "", ""},
+		{"dot docroot", ".", "", ""},
+		{"fail for outside approot", "..", "", "is outside the project root"},
+		{"fail for absolute path outside approot", "/test", "", "is outside the project root"},
+		{"ok for absolute path inside approot", tmpDir, "", ""},
+		{"dot with slash docroot", "./", "", ""},
+		{"dot with slash and dir docroot", "./test", "test", ""},
+		{"subdir docroot", "test/dir", "test/dir", ""},
+		{"dir docroot", "test", "test", ""},
+		{"dot with slash and subdir docroot", "./test/dir", "test/dir", ""},
+	}
+
+	for _, tc := range testMatrix {
+		t.Run(tc.description, func(t *testing.T) {
+			args = []string{
+				"config",
+				"--docroot", tc.input,
+			}
+
+			out, err = exec.RunHostCommand(DdevBin, args...)
+			if tc.error != "" {
+				require.Error(t, err)
+				require.Contains(t, out, tc.error)
+				return
+			}
+			require.NoError(t, err)
+
+			configContents, err = os.ReadFile(configFile)
+			require.NoError(t, err, "Unable to read %s: %v", configFile, err)
+
+			app = &ddevapp.DdevApp{}
+			err = yaml.Unmarshal(configContents, app)
+			require.NoError(t, err, "Could not unmarshal %s: %v", configFile, err)
+
+			require.Equal(t, tc.expected, app.Docroot)
+		})
+	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/ddevapp/config.go
```diff
@@ -1494,31 +1524,22 @@ func (app *DdevApp) docrootPrompt() error {
...
 	// If there is a default docroot, display it in the prompt.
 	if defaultDocroot != "" {
 		docrootPrompt = fmt.Sprintf("%s (%s)", docrootPrompt, defaultDocroot)
-	} else if cd, _ := os.Getwd(); cd == filepath.Join(app.AppRoot, defaultDocroot) {
-		// Preserve the case where the docroot is the current directory
-		docrootPrompt = fmt.Sprintf("%s (current directory)", docrootPrompt)
 	} else {
-		// Explicitly state 'project root' when in a subdirectory
 		docrootPrompt = fmt.Sprintf("%s (project root)", docrootPrompt)
 	}
```
stasadev, line 1535: There is no point in `(current directory)` because it will always say `(current directory)`, after I added `os.Chdir(appRoot)` to `ddev config` in the nested directories here:

https://github.com/ddev/ddev/blob/f1ab6f553ed3c159e9489316db86512f2eba0939/cmd/ddev/cmd/config.go#L323-L344

It should show `(project root)` all the time.

rfay:
Are there other things that have this misbehavior? I guess it would be flags that take a string. The ones that take a bool probably all behave correctly.

stasadev:
There is a `--composer-root=""` that doesn't work (we have a `--composer-root-default` for this that should be deprecated/deleted when it is fixed).

At first I wanted to look at all the flags, but then I saw that there was no test coverage for `CreateDocroot`, so I decided to do this PR exclusively for `--docroot`, and look at any other flags in another PR.

cmd/ddev/cmd/config_test.go
```diff
@@ -375,6 +375,51 @@ func TestConfigSetValues(t *testing.T) {
...
 	assert.Equal("SPACES=with spaces", app.WebEnvironment[3])
 	assert.Equal(webEnv, app.WebEnvironment[2])
+
+	// test docroot locations
+	testMatrix := []struct {
+		description string
+		input       string
+		expected    string
+		error       string
+	}{
```
rfay, line 385: Thanks for this great set of tests. No need to change anything in this PR, but this test is already long and unreadable, it's OK to split into new and separate tests when working on something like this.

cmd/ddev/cmd/config_test.go
```diff
@@ -375,6 +375,51 @@ func TestConfigSetValues(t *testing.T) {
...
+		{"empty docroot", "", "", ""},
+		{"dot docroot", ".", "", ""},
+		{"fail for outside approot", "..", "", "is outside the project root"},
```
rfay, line 388: Maybe I don't understand, but it looks like this is looking for the literal "..", but we probably want to be testing for something that *begins with* ".."? No need to make it fancier, it's already a good test as is.

pkg/ddevapp/config.go
```diff
@@ -1481,6 +1481,36 @@ func AvailablePHPDocrootLocations() []string {
...
 }
 
+// CreateDocroot normalizes the docroot path and creates it for DDEV app if it doesn't exist
+func (app *DdevApp) CreateDocroot() error {
+	if app.Docroot == "" {
+		return nil
+	}
+	docrootAbsPath := ""
+	if filepath.IsAbs(app.Docroot) {
```
rfay, line 1490: This works correctly, but it's not really clear to the user that the absolute path is a host-side docroot.  No need to change, but I would have preferred simpler approach that did not allow absolute path here. I think it adds complexity and code, but doesn't offer much to the user.

pkg/ddevapp/config.go
```diff
@@ -1505,31 +1535,22 @@ func (app *DdevApp) docrootPrompt() error {
 
 	// Determine the document root.
 	output.UserOut.Printf("\nThe docroot is the directory from which your site is served.\nThis is a relative path from your project root at %s\n", app.AppRoot)
-	output.UserOut.Printf("Leave docroot empty (hit <RETURN>) to use the location shown in parentheses.\nOr specify a custom path if your index.php is in a different directory.\n")
+	output.UserOut.Printf("Leave docroot empty (hit <RETURN>) to use the location shown in parentheses.\nOr specify a custom path if your index.php is in a different directory.\nOr use '.' (a dot) to explicitly set it to the project root.\n")
```
rfay, line 1538: This is great and works great. Much clearer.

rfay (approved):
Your careful work on this, improving the whole thing, not just the non-interactive command, is fantastic. You pay such great attention to the user experience. I especially appreciate the improved experience with interactive `ddev config`, which I always forget exists, and I forget that anybody ever uses it.

rfay:
This is a great improvement, but it's still possible for the docroot to not be normalized properly in config.yaml, so docroot of `/web` gets this on startup, and there may be other problems:

```
Unable to set permissions inside container on settings files: ''
```

So we should normalize the docroot in general. See https://discord.com/channels/664580571770388500/1339793828173058128/1340710341704683571

I'll open an issue.

PR #6968, opened 2025-02-07
build: autoload Python and Node deps from Makefile with direnv [skip ci]

stasadev:
## The Issue

The [Makefile](https://github.com/ddev/ddev/blob/main/Makefile) suggests installing some tools globally, but it is better to install them in an isolated environment, accessible only within DDEV directory.

Adds a config file for `direnv` tool, that will autoinstall everything needed.

It is completely optional and not necessary, but I find it easy to use.

1. Install direnv: https://direnv.net/docs/installation.html
2. Hook direnv into your shell: https://direnv.net/docs/hook.html
3. Create the global config: https://github.com/direnv/direnv/blob/master/man/direnv.toml.1.md
    where `~/workspace/ddev/.envrc` is the path to this file.
    (Alternatively, `.envrc` file can be loaded with `direnv allow .`)
    ```  
    $ cat ~/.config/direnv/direnv.toml
    [global]
    hide_env_diff = true
    [whitelist]
    exact = ["~/workspace/ddev/.envrc"]
	```
4. After setting this up, when you `cd` into this directory, all required tools
    will automatically be installed in .gotmp/env and sourced.
5. To update tools, run `make clean`, then `cd` out of and back into this directory.

Example of use:
```
$ eval "$(direnv hook bash)" # sourced in ~/.bashrc
$ cd ~/workspace/ddev
direnv: loading ~/workspace/ddev/.envrc
$ mkdocs -V
mkdocs, version 1.6.1 from /home/stas/workspace/ddev/.gotmp/env/python/lib/python3.13/site-packages/mkdocs (Python 3.13)
$ cd ..
$ mkdocs -V
bash: mkdocs: command not found
$ cd ddev
direnv: loading ~/workspace/ddev/.envrc
$ make mkdocs
mkdocs: 
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /tmp/mkdocsbuild
INFO    -  Documentation built in 6.60 seconds
```

The following message can be disabled with `export DIRENV_LOG_FORMAT=`
```
direnv: loading ~/workspace/ddev/.envrc
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+
+# This file is used by `direnv` to install all Python and Node dependencies
+# specified in the Makefile into the virtual env located in the `.gotmp/env` directory.
+
+# Steps to set up:
+# 1. Install direnv: https://direnv.net/docs/installation.html
+# 2. Hook direnv into your shell: https://direnv.net/docs/hook.html
+# 3. Create the global config: https://github.com/direnv/direnv/blob/master/man/direnv.toml.1.md
+#    where `~/workspace/ddev/.envrc` is the path to this file.
+#    (Alternatively, `.envrc` file can be loaded with `direnv allow .`)
+#    $ cat ~/.config/direnv/direnv.toml
+#    [global]
+#    hide_env_diff = true
+#    [whitelist]
+#    exact = ["~/workspace/ddev/.envrc"]
+# 4. After setting this up, when you `cd` into this directory, all required tools
+#    will automatically be installed in .gotmp/env and sourced.
+# 5. To update tools, run `make clean`, then `cd` out of and back into this directory.
+
+PYTHON_ENV=.gotmp/env/python
+NODE_ENV=.gotmp/env/node
+
+if ! command -v python >/dev/null 2>&1 && ! command -v python3 >/dev/null 2>&1; then
+    echo "Python is not available."
+    exit 1
+fi
+
+if [[ ! -f "${PYTHON_ENV}/bin/activate" ]]; then
+    echo "Initializing python venv..."
+    python -m venv "${PYTHON_ENV}" || python3 -m venv "${PYTHON_ENV}"
+fi
+
+source "${PYTHON_ENV}/bin/activate"
+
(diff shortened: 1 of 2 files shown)


rfay:
Please say what motivates this, thanks! Why would we want this, and what's the Makefile from?

stasadev:
When I run make `mkdocs-serve` or `make staticrequired`, some tools need to be installed.

I never run `pip install --break-system-packages` on the host, so I install everything in a Python virtual environment, activate that environment, and then run `make mkdocs-serve`.

Until today, I had a custom script that handled this for me, it was essentially a wrapper around the `Makefile`, which activated the virtual environment with the required tools. It was easy to use, except that I couldn't run the installed tools directly in the directory.

The technique I want to introduce adds an optional `.envrc` file (which is loaded only if you have `direnv` installed and activated in `ddev/ddev` root folder). This file activates the virtual Python and Node environment inside `.gotmp/env` directory (which is easy to remove/update/reinstall), installs the necessary tools, and automatically sources the tools when you `cd` into the `ddev/ddev` repo or its subfolders. When I `cd` out of `ddev/ddev`, none of those tools are available (which is useful if I want to run a custom `mkdocs` build elsewhere, install a different version of `linkspector`, and so on).

So contributors will only need `direnv` and `python` locally, rather than installing all these Python and Node packages and their dependencies globally.
(This does not apply to golang tools, which should still be installed globally.)

And it's optional, if there is no `direnv` installed and activated, nothing will happen.

rfay (approved):
Thanks, didn't understand the context. 

I use `pipx install mkdocs` I guess, that doesn't seem to break things?

stasadev:
I tried `pipx` but it installs to `~/.local` which still loads globally.

Does `pipx install mkdocs` does the same thing as `pip install -r docs/mkdocs-pip-requirements`, which installs many more packages?

That's why I preferred python venv to be as close as possible to what is inside [docs/mkdocs-pip-requirements](https://github.com/ddev/ddev/blob/main/docs/mkdocs-pip-requirements).

rfay:
I guess you're right, and I don't know what I did most recently. I was just saying that I installed mkdocs (apparently) using pipx.

I don't do so much with python that it matters a lot...

Issue #6891, opened 2025-01-11
Remove gitpod and gitpod support from docs


rfay:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```
[COPY-PASTE HERE THE OUTPUT OF `ddev debug test`]
```
</details>

### Expected Behavior

Gitpod "Classic" is now deprecated and will be removed on April 1, 2025. 

People are already being confused by visiting gitpod and getting told about "gitpod flex", which isn't a related item.

I think we should probably remove 
- [ ] Docs about gitpod
- [ ] `ddev-gitpod-base` base image Dockerfile and build (although we probably need a replacement to use for the devcontainers extensions).
- [ ] ddev/ddev-gitpod-launcher is probably already obsolete and can be updated and archived.

### Actual Behavior

x

### Steps To Reproduce

x

rfay:
@dennisameling Would you consider taking on the challenge of building a multi-arch replacement for the standard devcontainers universal base image, which is only published in amd64 (https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/universal/tags)

It actually wouldn't need to be universal, but essentially we'd like a devcontainers image that could be used not only with codespaces but also with ARM64 machines and DDEV + VS Code + devcontainers.

stasadev:
EOL date for Gitpod Classic:

Gitpod Classic will stay accessible until October 15. After, logins will be disabled and any active subscriptions will be canceled. Here’s how to migrate → https://ona.com/stories/gitpod-classic-payg-sunset

rfay:
They changed the name quite thoroughly to "Ona", but the product offering seems same (and has a free tier) except that it has lots of AI gibberish.

stasadev:
Their new free tier: 40 OCUs (one time)

https://ona.com/pricing#pricing-comparison

I assume that this means that the OCU will not be renewed in the next month.

But I don't understand if the OCU is just used to run the app, or if it is only used for AI requests.

rfay:
[...]

PR #6974, opened 2025-02-11
docs: Remove gitpod/drupalpod from docs, for #6891 [skip buildkite]

rfay:
## The Issue

- #6891 

This is the docs section of the removal.

* Remove docs
* Remove the helper on PRs


rfay:
I think it's probably fine to go ahead and do this, doesn't look like they're going to back off or anything. There will be other work to do in code.

stasadev (approved):
Looks good to me.

This change also requires a follow up PR for https://ddev.com/get-started/

PR #6977, opened 2025-02-12
build: Provide new multi-arch ddev-devcontainers-base image to replace ddev-gitpod-base, for #6891 [skip ci]

rfay:
## The Issue

- #6891 

We need to remove Gitpod support and improve multi-arch codespaces/devcontainers support.

* Move ddev-gitpod-base to ddev-devcontainers-base
* Remove .gitpod.yml


rfay:
This definitely doesn't work with Codespaces yet; no docker-in-docker.

rfay:
My goal here was to get an image working which could work for Codespaces and also work for VS Code Devcontainers and also work for PhpStorm remote containers. The idea was to have a multi-arch image. (The only known one image that works with Codespaces is amd64 only). So far I haven't succeeded in getting any image to work with Codespaces.

rfay:
I think this is probably a misguided attempt right now, but will likely try again later and resurrect this.

PR #6976, opened 2025-02-11
build: remove Xdebug 3.2.2 builds for php8.1 and php8.2

stasadev:
## The Issue

From Discord https://discord.com/channels/664580571770388500/1338908217094180935

A user experienced an issue with PHP 8.2 and Xdebug 3.2.2 when debugging was enabled.

Some time ago, we reverted Xdebug from 3.3.x to 3.2.2 in:

- #6181

## How This PR Solves The Issue

Xdebug 3.4.1 is now available for PHP 8.1 and PHP 8.2.

Given that some time has passed, it's time to switch to the newer versions.

This PR removes the custom PECL Xdebug builds for the mentioned PHP versions.

## Manual Testing Instructions

It should show Xdebug 3.4.1:
```
$ ddev start
$ ddev xdebug

$ ddev exec php8.1 -v
PHP 8.1.31 (cli) (built: Nov 21 2024 10:26:13) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.31, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.31, Copyright (c), by Zend Technologies
    with Xdebug v3.4.1, Copyright (c) 2002-2025, by Derick Rethans

ddev exec php8.2 -v
PHP 8.2.27 (cli) (built: Dec 24 2024 06:13:38) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.27, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.27, Copyright (c), by Zend Technologies
    with Xdebug v3.4.1, Copyright (c) 2002-2025, by Derick Rethans
```

PR #6979, opened 2025-02-12
build: push tagged images with ubuntu 22.04 [skip ci]

rfay:
## The Issue

* https://github.com/docker/setup-qemu-action/issues/198
* https://github.com/tonistiigi/binfmt/issues/240
* https://github.com/atomvm/AtomVM/pull/1530

It seems that there's a problem with qemu on Ubuntu 24.04 currently, so I'm unable to push-tagged-image because of segfaults on apt install

Go to 22.04 for now.

### DIFF
--- a/.github/workflows/push-tagged-dbimage.yml
+++ b/.github/workflows/push-tagged-dbimage.yml
@@ -27,7 +27,7 @@ permissions:
 jobs:
   push-tagged-dbimage:
     name: "push tagged dbimage"
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Oh, this must be a similar problem to the one I had with the `RUN apt-get update` in https://github.com/ddev/ddev/pull/6976#discussion_r1951747395 where it broke with a strange error.

Thanks!

rfay:
Unfortunately, this simple change doesn't seem to do the job for me. 

https://github.com/rfay/ddev/actions/runs/13292729872/job/37117213574

I'm able to build and push just fine from arm64 machine, as one of the issues says.

rfay:
The failure was on the `apt-get install`, is that the same as what you saw?

stasadev:
It's not the same, but I also saw similar errors on `apt-get install` recently.

PR #6986, opened 2025-02-14
test: initial draft for civicrm bats test

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Seems like a good start to me! Thanks!

PR #6987, opened 2025-02-14
test: contao quickstart bats test

rpkoller:
## The Issue

Initial draft of bats tests for contao. the first test for composer fails on the following step `run ddev exec contao-console contao:migrate --no-interaction` when running the test locally (https://gist.github.com/rpkoller/4d4f09f6971db12d4f7cbd4d2ba4a7bf). I've created the PR already to test it with tmate in github action as well per the suggestion of @rfay (otherwise i wouldnt have considered pushing a PR at this point yet)

- #<issue number>

### DIFF
--- a/docs/tests/contao.bats
+++ b/docs/tests/contao.bats
@@ -1,0 +1,71 @@
+#!/usr/bin/env bats
+
+setup() {
+  PROJNAME=my-contao-site
+  load 'common-setup'
+  _common_setup
+}
+
+# executed after each test
+teardown() {
+  _common_teardown
+}
+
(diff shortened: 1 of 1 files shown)


rpkoller:
tests are passing now. tested with tmate and the problem in the migrate line is not reproducible on github actions (probably only a local thing for me), fixed another problem with an assertion and now tests are passing. so ready for review. i'Ve removed the draft label

rfay (approved):
Looks like a good start, thanks so much!

rfay:
Moved to draft, dependent on
* https://github.com/ddev/ddev/pull/6994

rfay (approved):
Thanks for this!

PR #6990, opened 2025-02-14
test: quickstart bats test for ibexa

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Looks like a great start, thanks!

PR #6991, opened 2025-02-14
test: quickstart bats test for moodle

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Great start, thanks! I wonder if moodle is normally installed via composer. I remember when this one went in some time ago.

PR #6993, opened 2025-02-14
test: shopware quickstart bats test

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Thank you! I do hope we can do something better with the actual quickstart in the future.

PR #6994, opened 2025-02-15
fix: add a mutagen sync at the end of the dotenv cmd

rpkoller:
## The Issue

the problem surfaced during the creation of the sulu bats test. even though the env.local file was correctly created the admin console command was unable to pickup the correct datanbase credentials leading to failing tests on macos in bash and zsh while the tests ran properly on github actions. adding  `ddev mutagen sync` per randys suggestion fixed the problem 
 
```
  run ddev dotenv set .env.local --app-env=dev --database-url="mysql://db:db@db:3306/db?serverVersion=8.0&charset=utf8mb4"
  run ddev mutagen sync
  assert_success
  run ddev exec bin/adminconsole sulu:build dev --no-interaction
```
but instead of adding the mutagen sync statement in the quickstart and the tests it is better to run the mutagen sync within the scope of the command. randy provided the code and i've just quickly done the grunt working setting up the PR. 

- #<issue number>

### DIFF
--- a/cmd/ddev/cmd/dotenv-set.go
+++ b/cmd/ddev/cmd/dotenv-set.go
@@ -2,12 +2,13 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/ddev/ddev/pkg/output"
 	"os"
 	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
+
+	"github.com/ddev/ddev/pkg/output"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/util"
@@ -19,7 +20,7 @@ var DotEnvSetCmd = &cobra.Command{
 	Use:   "set [file]",
 	Short: "Write values from the command line to a .env file",
 	Long: `Create or update a .env file with values specified via long flags from the command line.
-Flags in the format --env-key=value will be converted to environment variable names 
+Flags in the format --env-key=value will be converted to environment variable names
 like ENV_KEY="value". The .env file should be named .env or .env.<servicename> or .env.<something>
 All environment variables can be used and expanded in .ddev/docker-compose.*.yaml files.
 Provide the path relative to the project root when specifying the file.`,
@@ -102,6 +103,10 @@ ddev dotenv set .ddev/.env.redis --redis-tag 7-bookworm`,
 		var formattedVars []string
 		for _, k := range rawResultKeys {
 			formattedVars = append(formattedVars, fmt.Sprintf(`%s=%v`, k, ddevapp.EscapeEnvFileValue(changedEnvMap[k])))
+		}
+		err = app.MutagenSyncFlush()
+		if err != nil {
+			util.Failed("failed to app.MutagenSyncFlush: %v", err)
 		}
 		friendlyMsg := fmt.Sprintf("Updated %s with:\n\n%s", envFile, strings.Join(formattedVars, "\n"))
 		output.UserOut.WithField("raw", changedEnvMap).Print(friendlyMsg)
--- a/cmd/ddev/cmd/dotenv-set.go
+++ b/cmd/ddev/cmd/dotenv-set.go
@@ -91,6 +91,10 @@ ddev dotenv set .ddev/.env.redis --redis-tag 7-bookworm`,
 			if err := ddevapp.WriteProjectEnvFile(envFile, changedEnvMap, envText); err != nil {
 				util.Failed("Error writing .env file: %v", err)
 			}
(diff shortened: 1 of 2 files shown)


rfay:
TestCmdDotEnvGetAndSet() failed because it operates on a not-running project, and app.MutagenSyncFlush() failed on a stopped project. 

I debugged it and added https://github.com/ddev/ddev/pull/6994/commits/97eccd1a0e99d649632a44086d6d1a9fab2b1384

It should be viewed with a bit of suspicion but should be OK.

cmd/ddev/cmd/dotenv-set.go
```diff
@@ -103,6 +104,10 @@ ddev dotenv set .ddev/.env.redis --redis-tag 7-bookworm`,
 		for _, k := range rawResultKeys {
 			formattedVars = append(formattedVars, fmt.Sprintf(`%s=%v`, k, ddevapp.EscapeEnvFileValue(changedEnvMap[k])))
 		}
+		err = app.MutagenSyncFlush()
+		if err != nil {
+			util.Failed("failed to app.MutagenSyncFlush: %v", err)
+		}
```
stasadev, line 110: This can be added immediately after 'ddevapp.WriteProjectEnvFile` inside 'if !reflect.DeepEqual` condition, but it doesn't harm to have it here later.

stasadev (approved):
Thank you!

cmd/ddev/cmd/dotenv-set.go
rpkoller: you mean that way 
```
		// Write only if there are changes
		if !reflect.DeepEqual(originalEnvMap, envMap) {
			if err := ddevapp.WriteProjectEnvFile(envFile, changedEnvMap, envText); err != nil {
				util.Failed("Error writing .env file: %v", err)
			}
			err = app.MutagenSyncFlush()
			if err != nil {
				util.Failed("failed to app.MutagenSyncFlush: %v", err)
			}
		}
```

stasadev: Yes.

rpkoller: thanks for the clarification. i quickly commit the necessary changes

rpkoller:
something might be off. after committing the latest change, let the artifact build and download it, i've tested the sulu test (which uses the dotenv command) and the test took forever; i've did a hard quit. but i also ran an existing test for wordpress and the second bedrock based test failed (but it also failed after reverting back to the latest version of HEAD)

```
✗ WordPress Bedrock based quickstart with ddev version v1.24.2-47-ge6bf644d6
   (from function `assert_success' in file /opt/homebrew/lib/bats-assert/src/assert.bash, line 114,
    in test file docs/tests/wordpress.bats, line 64)
     `assert_success' failed
   Volume my-wp-site-mariadb for project my-wp-site was deleted
   Volume my-wp-site-postgres for project my-wp-site was deleted
   Volume my-wp-site_project_mutagen for project my-wp-site was deleted
   Project my-wp-site was deleted. Your code and configuration are unchanged.
   
   -- command failed --
   status : 1
   output : sed: 1: ".env": invalid command code .
   --
   
```
so not sure what the exact reason is but somehow something feels off. but not sure if in this PR or in HEAD or in both?

stasadev:
> not sure what the exact reason is but somehow something feels off

I'm not sure what could be wrong, I tested Sulu test PR with and without this PR with Mutagen on Linux and it worked as expected (failed without this PR and succeeded with it).

Maybe there were some leftovers, did you run `ddev poweroff` before running tests again?
Also check for test leftovers in `ddev list`, and check if there is something sitting inside `~/tmp` (a directory in which our tests are run).

rpkoller:
hm no, my list is clean (only projects that are supposed to be there), and the tmp folder is empty. not sure what is going on  but the problem persist. 

with the PR artifact: the second wordpress bats test failed same as shown above, and the sulu test takes forever and i have to forcequit. 

with the HEAD: same results. 

not sure what is causing that. but the wordpress errors might be unrelated but just went unnoticed so far? (but back then when i created those tests they worked all locally on my end - will investigate in the coming days whats up with the second wordpress test)

rfay:
I don't see any reason that the behavior of `ddev dotenv` here would be different from the previous version.

However, the assumption that it wasn't as good at the end of the code isn't correct I don't think.

* When there are no changes, a mutagen sync is of trivial impact
* When mutagen is not enabled, a mutagen sync is a no-op.

rfay:
I'm going to go ahead and pull this, we'll get a HEAD build, and we can retest the quickstart tests and see where we're at.

PR #6995, opened 2025-02-15
test: bats test for the sulu quickstart

rpkoller:
## The Issue

locally the tests are not working because they require a mutagen sync but the tests work on github actions already. the not working local tests are being fixed with https://github.com/ddev/ddev/pull/6994 (tested the artifact with the tests that failed before locally and work now with the artifact in PR6994)

- #<issue number>


rpkoller:
https://github.com/ddev/ddev/pull/6994 has to go in first

rfay:
Rebased to see what happens with buildkite tests.

rfay (approved):
Looking good to me, thanks! Hopefully we can do even better with actually exercising quickstarts like this in the future.

PR #6996, opened 2025-02-15
test: add yes flag to ddev start cmd

rpkoller:
## The Issue

ddev start and or ddev restart was missing the yes flag on several bats tests. without a test is stuck on the new version of ddev prompt and could never proceed

- #<issue number>


rfay (approved):
Thanks!

PR #7001, opened 2025-02-16
fix: validate docroot instead of using it blindly, fixes #6998

### LINKED ISSUE #6998: Normalize and/or validate docroot obtained from config.yaml `docroot`
### Preliminary checklist

```
```
</details>

### Expected Behavior

`docroot` in config.yaml should always be a relative path from the project root

* https://github.com/ddev/ddev/pull/6962 fixed loads of things here, but only when the docroot is done via `ddev config`

### Actual Behavior

One can put `docroot: /web` and cause a number of interesting problems, see https://discord.com/channels/664580571770388500/1339793828173058128/1340710341704683571

### Steps To Reproduce

Set `docroot: /web` and `ddev restart`.

You'll get 
```
Unable to set permissions inside container on settings files: ''
Unable to set permissions inside container on settings files: ''
```

### Anything else?

- [ ] We should normalize the docroot at config read time
- [ ] drupalEnsureWritePerms() should use the normalized docroot, probably prefixed by /var/www/html: https://github.com/ddev/ddev/blob/dbd83788a53c3211b22b205ec8d2d15a34678075/pkg/ddevapp/drupal.go#L536-L582

rfay:
## The Issue

* #6998

* Validate the docroot to make sure it isn't absolute or start with ..
* I didn't try to update tests
* This now does more validating and less normalizing (it doesn't change the offered docroot)

Try awkward and wrong docroot manually edited in config.yaml
* /web
* ../something
* other?

### DIFF
--- a/cmd/ddev/cmd/composer-create.go
+++ b/cmd/ddev/cmd/composer-create.go
@@ -54,7 +54,7 @@ ddev composer create --prefer-dist --no-interaction --no-dev psr/log
 		if status != ddevapp.SiteRunning {
 			err = app.CreateDocroot()
 			if err != nil {
-				util.Failed("Could not create docroot at %s: %v", app.Docroot, err)
+				util.Failed("Could not create docroot at %s: %v", app.GetAbsDocroot(false), err)
 			}
 			err = app.Start()
 			if err != nil {
@@ -475,7 +475,7 @@ func prepareAppForComposer(app *ddevapp.DdevApp) {
 	if !fileutil.IsDirectory(app.GetAbsDocroot(false)) {
 		err := app.CreateDocroot()
 		if err != nil {
-			util.Failed("Could not create docroot at %s: %v", app.Docroot, err)
+			util.Failed("Could not create docroot at %s: %v", app.GetAbsDocroot(false), err)
 		}
 		// Restart the project after creating docroot
 		if err := app.Restart(); err != nil {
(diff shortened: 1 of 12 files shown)


stasadev (changes requested):
I noticed these:
```yaml
docroot: "/web"
```
```
$ ddev exec 'echo $DDEV_DOCROOT'
/web
```
And:
```yaml
docroot: ".."
```
```
```
It created `sites` directory next to the project directory `d11` 🤯
It should not have access to anything outside the approot.
I think we should validate the docroot for absolute path and fail in `ddevapp.ValidateConfig()`

rfay:
TODO:
- [ ] Normalize most accesses to `docroot`
- [ ] Make sure DDEV_DOCROOT is normalized
- [ ] Add validation to make sure that `docroot` is reasonable

stasadev (commented):
It looks good, there is no normalization anymore (I'm okay with that, as far as we have validation).

But
```
ddev config --docroot=/test
failed to validate config: docroot ('/test') cannot be an absolute path, it must be a relative path from the project root
```
creates a `test` folder at the root of the project, even if it shows an error.

rfay:
Thanks for the good testing. This now does more validating and no normalizing.

stasadev (approved):
Looks good to me!

I tested it with `ddev config` and by manually editing the `.ddev/config.yaml` file.
I didn’t find any issues with the `docroot`. Thank you!

PR #7005, opened 2025-02-19
test: fix and improve the ability to set dotenv values on every OS

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Thanks! This isn't the only way to do this, but here we go!

PR #6930, opened 2025-01-22
fix: handle args with spaces in `ddev exec`, don't panic about invalid syntax in `exec_raw`, fixes #5865

stasadev:
## The Issue

- #5865
- #3603

## How This PR Solves The Issue

Uses the same technique I tried in:

- #6627

But this time I fixed the command with pipes `ddev exec "set | grep IS_DDEV_PROJECT"`

TODO:

- ~~I marked the `--raw` flag hidden. Mark it deprecated, remove it from the docs?~~
- ~~Remove `--raw` from tests~~
- ~~Check the hooks, maybe we can get rid of `exec_raw`~~

Edit: I decided not to remove the `--raw` flag, this PR simply runs `ddev exec` with *multiple* (2+) arguments, as it could be run using `ddev exec --raw`.

---

And it fixes panic for the wrong syntax on `ddev start` (this is indirectly related to `exec`, so I decided to fix it here):

```yaml
hooks:
    post-start:
        - exec:
          exec_raw: ls -la # the command should look like this [ls, la]
```

```
$ ddev start
...
panic: interface conversion: interface {} is string, not []interface {}
...
```

## Manual Testing Instructions

Using DDEV v1.24.2:
```bash
$ echo 'for i; do echo $i; done' > issue.sh

$ ddev exec bash issue.sh "string with spaces"
string
with
spaces

$ ddev exec bash issue.sh "string with spaces" --test="a b c"
string
with
spaces
--test=a
b
c

$ ddev exec 'bash issue.sh "string with spaces"'
string with spaces

$ ddev exec "set | grep IS_DDEV_PROJECT"
BASH_EXECUTION_STRING='set -eu && ( set | grep IS_DDEV_PROJECT)'
IS_DDEV_PROJECT=true

$ echo '<?php var_dump($_SERVER["argv"]);' > args.php

$ ddev exec php args.php --option="argument with string"
array(4) {
  [0]=>
  string(8) "args.php"
  [1]=>
  string(17) "--option=argument"
  [2]=>
  string(4) "with"
  [3]=>
  string(6) "string"
}
```

Using this PR:
```bash
$ echo 'for i; do echo $i; done' > issue.sh

$ ddev exec bash issue.sh "string with spaces"
string with spaces

$ ddev exec bash issue.sh "string with spaces" --test="a b c"
string with spaces
--test=a b c

$ ddev exec 'bash issue.sh "string with spaces"'
string with spaces

$ ddev exec "set | grep IS_DDEV_PROJECT"
[...]

PR #7006, opened 2025-02-19
fix: wrap args with spaces in `ddev exec` with double quotes instead of using raw, for #6930

stasadev:
## The Issue

- #6930

After merging the PR above, CiviCRM quickstart failed because the value of `$DDEV_PRIMARY_URL` was not expanded.

```
ddev exec cv core:install --cms-base-url='$DDEV_PRIMARY_URL' ...
...
#   ERROR: (system) cmsBaseUrl: The "cmsBaseUrl" ($DDEV_PRIMARY_URL) is unavailable or malformed. Consider setting it explicitly.
...
```

The use of `ExecRaw` broke variable expansion.

We can't use `ExecRaw` because variable expansion doesn't work.

I can add `os.ExpandEnv`, but it's not the same as bash expansion:

- https://github.com/golang/go/issues/47187

This means that any existing `ddev exec` code that depends on bash expansion will be broken.

So I reverted this change and added a double quote wrapping, which should be an appropriate solution that still fixes the arguments with spaces.

Using DDEV v1.24.2:
```bash
$ ddev exec echo '$IS_DDEV_PROJECT ${FOO_BAR:-foobar}'
true foobar

$ echo 'for i; do echo $i; done' > issue.sh

$ ddev exec bash issue.sh "string with spaces"
string
with
spaces

$ ddev exec bash issue.sh "string with spaces" --test="a b c"
string
with
spaces
--test=a
b
c

$ ddev exec 'bash issue.sh "string with spaces"'
string with spaces

$ ddev exec "set | grep IS_DDEV_PROJECT"
BASH_EXECUTION_STRING='set -eu && ( set | grep IS_DDEV_PROJECT)'
IS_DDEV_PROJECT=true

$ echo '<?php var_dump($_SERVER["argv"]);' > args.php

$ ddev exec php args.php --option="argument with string"
array(4) {
  [0]=>
  string(8) "args.php"
  [1]=>
  string(17) "--option=argument"
  [2]=>
  string(4) "with"
  [3]=>
  string(6) "string"
}
```

Using DDEV HEAD (which has broken variable expansion when using several args):
```bash
$ ddev exec echo '$IS_DDEV_PROJECT ${FOO_BAR:-foobar}'
$IS_DDEV_PROJECT ${FOO_BAR:-foobar}

$ echo 'for i; do echo $i; done' > issue.sh

$ ddev exec bash issue.sh "string with spaces"
string with spaces

$ ddev exec bash issue.sh "string with spaces" --test="a b c"
string with spaces
--test=a b c

$ ddev exec 'bash issue.sh "string with spaces"'
string with spaces

$ ddev exec "set | grep IS_DDEV_PROJECT"
BASH_EXECUTION_STRING='set -eu && ( set | grep IS_DDEV_PROJECT)'
IS_DDEV_PROJECT=true

$ echo '<?php var_dump($_SERVER["argv"]);' > args.php

$ ddev exec php args.php --option="argument with string"
array(2) {
  [0]=>
  string(8) "args.php"
  [1]=>
  string(29) "--option=argument with string"
}
```

Using this PR:
```bash
$ ddev exec echo '$IS_DDEV_PROJECT ${FOO_BAR:-foobar}'
true foobar

$ echo 'for i; do echo $i; done' > issue.sh

$ ddev exec bash issue.sh "string with spaces"
string with spaces

$ ddev exec bash issue.sh "string with spaces" --test="a b c"
string with spaces
--test=a b c

$ ddev exec 'bash issue.sh "string with spaces"'
string with spaces

$ ddev exec "set | grep IS_DDEV_PROJECT"
BASH_EXECUTION_STRING='set -eu && ( set | grep IS_DDEV_PROJECT)'
IS_DDEV_PROJECT=true

$ echo '<?php var_dump($_SERVER["argv"]);' > args.php

$ ddev exec php args.php --option="argument with string"
array(2) {
  [0]=>
  string(8) "args.php"
  [1]=>
  string(29) "--option=argument with string"
}
```

### DIFF
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -45,15 +45,13 @@ ddev exec --raw -- ls -lR`,
 		opts := &ddevapp.ExecOpts{
 			Service: serviceType,
 			Dir:     execDirArg,
-			Cmd:     strings.Join(args, " "),
+			Cmd:     quoteArgs(args),
 			Tty:     true,
 		}
 
 		// If they've chosen raw, use the actual passed values.
-		// If there are multiple arguments, treat them as a raw command.
-		// This avoids splitting quoted strings with spaces into separate arguments.
 		// Also, retrieve and preserve the current $PATH to ensure the environment is consistent.
-		if cmd.Flag("raw").Changed || len(args) > 1 {
+		if cmd.Flag("raw").Changed {
 			var env []string
 			path, _, err := app.Exec(&ddevapp.ExecOpts{
 				Service: serviceType,
@@ -73,6 +71,22 @@ ddev exec --raw -- ls -lR`,
 			util.Failed("Failed to execute command %s: %v", strings.Join(args, " "), err)
 		}
 	},
+}
+
+// quoteArgs quotes any arguments that contain spaces.
+// This avoids splitting quoted strings with spaces into separate arguments.
+func quoteArgs(args []string) string {
+	if len(args) > 1 {
+		for i, arg := range args {
+			// Check if the argument contains spaces
+			if strings.Contains(arg, " ") {
+				// Escape existing double quotes and wrap in double quotes
+				arg = `"` + strings.ReplaceAll(arg, `"`, `\"`) + `"`
+			}
+			args[i] = arg
+		}
+	}
+	return strings.Join(args, " ")
 }
 
 func init() {
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -61,6 +61,7 @@ ddev exec --raw -- ls -lR`,
 			if err == nil && path != "" {
 				env = append(env, "PATH="+path)
 			}
+			// opts.RawCmd is used instead of opts.Cmd
 			opts.RawCmd = args
 			opts.Env = env
 		}
@@ -68,25 +69,32 @@ ddev exec --raw -- ls -lR`,
 		_, _, err = app.Exec(opts)
 
 		if err != nil {
-			util.Failed("Failed to execute command %s: %v", strings.Join(args, " "), err)
+			util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
 		}
 	},
 }
 
 // quoteArgs quotes any arguments that contain spaces.
 // This avoids splitting quoted strings with spaces into separate arguments.
+// The function is adapted from the internal quoteArgs golang function.
 func quoteArgs(args []string) string {
-	if len(args) > 1 {
-		for i, arg := range args {
-			// Check if the argument contains spaces
-			if strings.Contains(arg, " ") {
-				// Escape existing double quotes and wrap in double quotes
-				arg = `"` + strings.ReplaceAll(arg, `"`, `\"`) + `"`
-			}
-			args[i] = arg
+	if len(args) < 2 {
+		return strings.Join(args, " ")
+	}
+	var b strings.Builder
+	for i, arg := range args {
+		if i > 0 {
+			b.WriteString(" ")
+		}
+		if strings.ContainsAny(arg, "\" \t\r\n#") {
+			b.WriteString(`"`)
+			b.WriteString(strings.ReplaceAll(arg, `"`, `\"`))
+			b.WriteString(`"`)
+		} else {
+			b.WriteString(arg)
 		}
 	}
-	return strings.Join(args, " ")
+	return b.String()
 }
 
 func init() {
--- a/cmd/ddev/cmd/exec_test.go
+++ b/cmd/ddev/cmd/exec_test.go
@@ -71,12 +71,13 @@ func TestCmdExec(t *testing.T) {
 	// Test with raw cmd
 	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "--", "ls", "/usr/local")
 	assert.NoError(err)
-	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), "expected '%v' to start with 'bin\\netc\\games\\include'", out)
+	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), `expected '%v' to start with '%s'`, out, `bin\netc\ngames\ninclude`)
 
 	// Test with raw cmd and bash
 	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "bash", "-c", "ls /usr/local")
 	assert.NoError(err)
-	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), "expected '%v' to start with 'bin\\netc\\games\\include'", out)
+	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), `expected '%v' to start with '%s'`, out, `bin\netc\games\include`)
+
 
 	// Test sudo
 	out, err = exec.RunHostCommand(DdevBin, "exec", "sudo", "whoami")
--- a/cmd/ddev/cmd/exec_test.go
+++ b/cmd/ddev/cmd/exec_test.go
@@ -109,10 +109,37 @@ func TestCmdExec(t *testing.T) {
 	assert.NoError(err)
 	assert.NotEqual("/var/www/html", strings.TrimSpace(out))
 
-	// Complex commands with spaces should also work, even without the --raw flag
-	out, err = exec.RunHostCommand(DdevBin, "exec", "bash", "-c", "for i; do echo $i; done", "_", "string with spaces")
+	// Create a bash script for testing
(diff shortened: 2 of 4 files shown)


stasadev:
Quickstart test worked fine, I'll add the actual test coverage for this in `TestCmdExec`.

deviantintegral:
Is this related to this failure on ddev HEAD? It looks like `~` is no longer being passed properly to the shell inside of ddev.

Called at: https://github.com/Lullabot/ddev-playwright/blob/main/tests/test.bats#L79-L81

The error: https://github.com/Lullabot/ddev-playwright/actions/runs/13431047279/job/37536159660#step:11:487

rfay:
Thanks so much for noting and reporting the problem @deviantintegral !

Notes from the peanut gallery about https://github.com/Lullabot/ddev-playwright/blob/main/tests/test.bats#L79-L81

* I would have done `ddev exec 'ls ~/.cache/ms-playwright'`, simple, obvious, no escaping needed. 
* I would have used a `run` and checked its success, but you seem to be repeating it in the next line.
* I would have done simpler `ddev exec 'ls ~/.cache/ms-playwright | wc -l | sed "s/ *//"'`, interested in why you used more complex. No escaping, nothing.

cmd/ddev/cmd/exec.go
```diff
@@ -75,6 +73,22 @@ ddev exec --raw -- ls -lR`,
 	},
 }
 
+// quoteArgs quotes any arguments that contain spaces.
```
rfay, line 76: It may be worth looking at the quoteArgs() used internally in go:
```
func quoteArgs(args []string) string {
	var b strings.Builder
	for i, arg := range args {
		if i > 0 {
			b.WriteString(" ")
		}
		if strings.ContainsAny(arg, "'"+argSepChars) {
			// Quote the argument to a form that would be parsed as a single argument.
			b.WriteString("'")
			b.WriteString(strings.ReplaceAll(arg, "'", "''"))
			b.WriteString("'")
		} else {
			b.WriteString(arg)
		}
	}
	return b.String()
}
```

cmd/ddev/cmd/exec.go
```diff
@@ -45,15 +45,13 @@ ddev exec --raw -- ls -lR`,
 		opts := &ddevapp.ExecOpts{
 			Service: serviceType,
 			Dir:     execDirArg,
-			Cmd:     strings.Join(args, " "),
+			Cmd:     quoteArgs(args),
```
rfay, line 48: I'm not convinced that it's appropriate to do this when using ExecRaw. Definitely want to talk about it.

rfay (changes requested):
This is so complicated and so fragile, you're a brave one to go there! 

Here's an example of something that worked fine in v1.24.2, but no longer works with this version:

```
rfay@MacBookAir:~/workspace/d11$ ddev exec --raw bash -c "ls -l"
bash: line 1: ls -l: command not found
Failed to execute command bash -c "ls -l": exit status 127
```

cmd/ddev/cmd/exec.go
stasadev: `Cmd` is not used when `RawCmd` is provided, I will set `opts.Cmd = ""` for `--raw` explicitly to make it clear.

cmd/ddev/cmd/exec.go
stasadev: Thank you!

I didn't know about this function, I'll copy it.

cmd/ddev/cmd/exec.go
stasadev: I got it, the problem was that I modified `args` inside `quoteArgs(args)`
I thought there was a copy inside, but it was a reference.

stasadev:
`TestCmdExec` is a fairly large test, but I decided not to rewrite it using `testCases` because the conditions are different, and unifying it might reduce readability and make it harder to understand.

cmd/ddev/cmd/exec_test.go
```diff
@@ -71,8 +71,12 @@ func TestCmdExec(t *testing.T) {
 	// Test with raw cmd
 	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "--", "ls", "/usr/local")
 	assert.NoError(err)
+	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), "expected '%v' to start with 'bin\\netc\\games\\include'", out)
```
rfay, line 74: nit: These two assertion messages don't show correctly, and the expectation and actual are switched. this can be written much more easily. (Just the failure message is bad.)

```suggestion
	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), `expected '%v' to start with '%s'`, out, `bin\netc\ngames\ninclude`)
```

cmd/ddev/cmd/exec_test.go
```diff
@@ -71,8 +71,12 @@ func TestCmdExec(t *testing.T) {
...
 
-	assert.True(strings.HasPrefix(out, "bin\netc\ngames"), "expected '%v' to start with 'bin\\netc\\games\\include'", out)
+	// Test with raw cmd and bash
+	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "bash", "-c", "ls /usr/local")
+	assert.NoError(err)
+	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), "expected '%v' to start with 'bin\\netc\\games\\include'", out)
```
rfay, line 79:
```suggestion
	assert.True(strings.HasPrefix(out, "bin\netc\ngames\ninclude"), `expected '%v' to start with '%s'`, out, `bin\netc\games\include`)

```

cmd/ddev/cmd/exec_test.go
```diff
@@ -109,10 +113,37 @@ func TestCmdExec(t *testing.T) {
...
+	// Bash expansion doesn't work with --raw, it's expected
+	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "echo", "$IS_DDEV_PROJECT")
+	assert.NoError(err)
+	assert.Equal("$IS_DDEV_PROJECT", strings.TrimSpace(out))
+
+	// Bash expansion doesn't work with --raw, it's expected (using arg with spaces)
+	out, err = exec.RunHostCommand(DdevBin, "exec", "--raw", "echo", "$IS_DDEV_PROJECT\n${DDEV_NON_EXISTING_TEST_VARIABLE:-foobar}")
 	assert.NoError(err)
-	assert.Equal("string with spaces", strings.TrimSpace(out))
+	assert.Equal("$IS_DDEV_PROJECT\n${DDEV_NON_EXISTING_TEST_VARIABLE:-foobar}", strings.TrimSpace(out))
```
rfay, line 146: This is a bit baffling, so a comment explaining why it's here would be an improvement.

cmd/ddev/cmd/exec_test.go
```diff
@@ -109,10 +113,37 @@ func TestCmdExec(t *testing.T) {
...
-	out, err = exec.RunHostCommand(DdevBin, "exec", "bash", "-c", "for i; do echo $i; done", "_", "string with spaces")
+	// Create a bash script for testing
+	_, err = exec.RunHostCommand(DdevBin, "exec", "echo 'for i; do echo $i; done' > echo_args.sh")
+	assert.NoError(err)
+	err = app.MutagenSyncFlush()
+	assert.NoError(err)
+	assert.FileExists(filepath.Join(site.Dir, "echo_args.sh"))
+
+	// Arguments with spaces should not be split, quotes should be preserved
+	out, err = exec.RunHostCommand(DdevBin, "exec", "bash", "echo_args.sh", "string with \"quotes\" and spaces", "another 'quoted' string")
```
rfay, line 124: Nice techniques here!

cmd/ddev/cmd/exec.go
```diff
@@ -63,18 +61,42 @@ ddev exec --raw -- ls -lR`,
...
-			util.Failed("Failed to execute command %s: %v", strings.Join(args, " "), err)
+			util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
 		}
 	},
 }
 
+// quoteArgs quotes any arguments that contain spaces.
+// This avoids splitting quoted strings with spaces into separate arguments.
+// The function is adapted from the internal quoteArgs golang function.
+func quoteArgs(args []string) string {
```
rfay, line 80: I can't claim to have studied this carefully :( It looks pretty fancy.

rfay (approved):
I think it's great, thanks! A couple of nits on error messages, and a suggestion on comment. This will help us get HEAD working again. The approach to testing for correct args and such is excellent.

cmd/ddev/cmd/exec_test.go
stasadev: Updated with comment.

Issue #6894, opened 2025-01-13
XHGui support integrated in DDEV core


rfay:
Related: 
- #3782 
- [TYPO3 Community Budget Process Announcement](https://typo3.org/article/four-ideas-to-be-funded-in-quarter-1-2025)

The TYPO3 Community Ideas Process has selected DDEV XHGui  integration as a project for Q1 2025. 

This issue will serve as the project plan for this.

Goals:
- [x] Improve the existing [ddev-xhgui](https://github.com/ddev/ddev-xhgui) add-on with a number of features, including removing the requirement for additional configuration.
- [x] Add xhgui command to DDEV. It should create required database if it doesn't exist, and enable xhprof. 
- [ ] Deprecate/hide the xhprof command
- [x] Avoid breaking the workflow of existing users that depend on xhprof in the traditional technique
- [x] Update docs for performance
- [ ] Deprecate the ddev-xhgui add-on
- [ ] Run xhgui container with the current user (instead of root)
- [x] Add bash to the xhgui built image
- [x] Consider pegging xhgui image instead of using `latest`.
- [x] Consider installing xhgui in the web container instead of using separate container.
- [x] Xhgui should not be in the prod/hardened image

Notes:
- [x] [xhgui/xhgui](https://hub.docker.com/r/xhgui/xhgui) is alpine-based and hasn't been pushed in a year, and has no information about its build. The [org](https://hub.docker.com/u/xhgui) also has no information. 
- [x] The [github repo](https://github.com/perftools/xhgui) appears to be maintained by the same user ([glensc](https://github.com/glensc)) who pushes the xhgui image.
- [x] His php-profiler shows recent but sporadic activity, https://github.com/perftools/php-profiler
- [x] Wow, xhgui has been around for a long time!

rfay:
> What do you mean with deprecating ddev xhprof

Just mean deprecate the command `ddev xhprof` in favor of `ddev xhgui`.

rfay:
Just a note here about the up-and-coming [php-spx](https://github.com/NoiseByNorthwest/php-spx). It's marked as experimental, does not run nightly tests, but some folks are thrilled with it.  There is an experimental DDEV add-on at https://github.com/fullfatthings/ddev-spx

sreichel:
I already was happy with xhgui, but now with spx on-top it becomes easy to find bottlenecks.

https://github.com/OpenMage/magento-lts/pull/4517#issuecomment-2628691004

❤

[...]

PR #7007, opened 2025-02-19
feat: Support docker compose optional profiles, allow `ddev start --profiles=list,of,profiles`, for #6894

rfay:
## The Issue

- #6894 

For xhgui, we don't want to start the container unless needed, so it can be limited to a specific profile, then started in `ddev xhgui`

[Docker Compose profiles](https://docs.docker.com/compose/how-tos/profiles/) can be used to label services for starting at an "optional" time. 

* Add handling of profiles in ComposeCmd()
* Support optional profiles
* Add `ddev start --profiles=list,of,profiles` as a way to do it manually

* Use the ddev-xhgui add-on, `ddev add-on get ddev/ddev-xhgui`

* Add this to the `docker-compose.xhgui.yaml`:
```yaml
profiles:
  - xhgui
```

* `ddev start`

* `ddev start --profiles=xhgui`

## TODO:

- [x] Don't just use hard-wired "optional", instead make "optional" the default, give it a constant, allow other profiles.
- [x] Tests for this feature
- [x] Add `ddev restart --profiles=<whatever>`
- [x] Consider whether this might make global add-ons possible, or at least global `docker-compose.*.yaml`

## Automated Testing Overview

* TestCmdStartOptionalProfiles
* TestDdevApp_StartOptionalProfiles


rfay:
I haven't figured out how to support more than one profile, mostly because the `docker-compose config` command only brings in all the profiles if you specify them. So to have a full .ddev/.ddev-docker-compose-full.yaml, we have to bring in all the profiles, or the start doesn't work.

pkg/ddevapp/compose_yaml.go
```diff
@@ -56,6 +58,7 @@ func (app *DdevApp) WriteDockerComposeYAML() error {
 	}
 	fullContents, _, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
 		ComposeFiles: files,
+		Profiles:     []string{nodeps.ComposeProfileOptional},
```
stasadev, line 61: > I haven't figured out how to support more than one profile, mostly because the `docker-compose config` command only brings in all the profiles if you specify them.

Does this solve the problem described?

```suggestion
		Profiles:     []string{"*"},
```

The docs say:

If you want to enable all profiles at the same time, you can run `docker compose --profile "*"`

https://docs.docker.com/compose/how-tos/profiles/#start-specific-profiles

rfay: However, is there a difference between that and "optional" in the end? What is clearer for our users?

I'll definitely consider "*".

rfay: Oh, this may solve things for me though! I cn use "*" in the `docker-compose config`, and then I'll be able to use arbitrary profiles in the start. Thanks!

stasadev: We can use `--optional` as quick and easy way for `optional` profile and add a new flag that gives people more flexibility, i.e. use the same `--profile` (array flag) as in `docker-compose` for a broader approach if it makes sense. Flags must be mutually exclusive.

pkg/dockerutil/dockerutils.go
```diff
@@ -712,6 +713,10 @@ func ComposeCmd(cmd *ComposeCmdOpts) (string, string, error) {
 		arg = append(arg, "-f", file)
 	}
 
+	if cmd.Profiles != nil {
+		arg = append(arg, "--profile", strings.Join(cmd.Profiles, ","))
+	}
```
stasadev, line 718: I think it won't work like this (although the code works, but only with one profile).

This should work:

```suggestion
	for _, profile := range cmd.Profiles {
		arg = append(arg, "--profile", profile)
	}
```

pkg/nodeps/values.go
```diff
@@ -43,6 +43,11 @@ const (
 	WebserverGeneric   = "generic"
 )
 
+// Docker Compose Profiles
+const (
+	ComposeProfileOptional = "optional"
+)
+
```
stasadev, line 50: Is it no longer needed?

pkg/ddevapp/ddevapp.go
```diff
@@ -1794,6 +1794,40 @@ If this seems to be a config issue, update it accordingly.`, app.Name)
...
+	if status, _ := app.SiteStatus(); status != SiteRunning {
+		err = app.Start()
+		if err != nil {
+			return err
+		}
+	}
+
+	_, stderr, err := dockerutil.ComposeCmd(&dockerutil.ComposeCmdOpts{
+		ComposeFiles: []string{app.DockerComposeFullRenderedYAMLPath()},
+		Profiles:     []string{profile},
```
stasadev, line 1809: It would be good to split profile by comma, and allow this:

```
ddev start --profile=xhgui,extra
```

stasadev (changes requested):

pkg/nodeps/values.go
rfay: Correct, no longer needed, will remove.

pkg/dockerutil/dockerutils.go
rfay: Interesting, I wasn't actually thinking of offering more than one profile. Probably shouldn't have been cmd.Profiles. Do you think there's a need for more than one, since a single profile can easily start multiple services?

stasadev: I think having more than one profile is fine, especially since we're not enforcing a specific profile name.

People are unlikely to name their profiles the same across different add-ons. So, if someone pulls multiple add-ons with profiles and starts them all, the profiles will likely differ. Instead of running `ddev start --profile` for each one, they could start them all with a single command.

stasadev: `ddev start --profile="*"` is useful, but having something more specific is even better. For example, starting only certain profiles, like this `ddev start --profile=xhgui,extra` allows you to start some profiles while excluding others.

docs/content/users/extend/custom-compose-files.md
```diff
@@ -136,3 +136,20 @@ RUN export TARGETPLATFORM=linux/$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm6
 RUN chmod +x /usr/local/bin/mkcert
 USER original_user
 ```
+
+## Optional Services
+
+Services in named Docker Compose profiles will not automatically be started on `ddev start`. This is useful when you want to define a service that is not always needed, but can be started by an additional command when it is time to use it. In this way, it doesn't use system resources unless needed. In this example, the `busybox` container will only be started if the `busybox` profile is requested, for example with `ddev start --profile=busybox`. More than one service can be labeled for a single Docker Compose profile.
```
stasadev, line 142:
```suggestion
Services in named Docker Compose profiles will not automatically be started on `ddev start`. This is useful when you want to define a service that is not always needed, but can be started by an additional command when it is time to use it. In this way, it doesn't use system resources unless needed. In this example, the `busybox` container will only be started if the `busybox` profile is requested, for example with `ddev start --profiles=busybox`. More than one service can be labeled for a single Docker Compose profile.

!!!tip "Run `ddev start --profiles='*'` to start all defined profiles."
```

stasadev (changes requested):
It works very well, I tried different combinations of `ddev start` with and without `--profiles`, and everything worked perfectly as expected.

---

I have a small suggestion regarding the docs.

---

> TODO: `Add ddev restart --profiles=<whatever>`

Is this feature planned for a future PR? The checkbox is marked as active, which caused some confusion, it looks like it's already implemented, but I don't see it included here.

rfay:
Hmm, I must have though I'd just "skip" the `ddev restart --profiles`, so checked it. 

The reality is that if this gets used much, people will immediately want `ddev restart --profiles` and `ddev stop --profiles`

But I guess I'd prefer to move on without those at this point, to get focused on the XHGui stuff.

stasadev:
Okay, thanks for the clarification, I'm okay with this.

stasadev (approved):

PR #7009, opened 2025-02-20
test: Makefile - try again if curl fails to download something it needs

rfay:
## The Issue

In tests, especially buildkite, we rather often have the mysterious test failure:

```
make: *** 
[.gotmp/bin/linux_amd64/mkcert] Error 92
```

This is just curl failing to download mkcert from the github release

* Try 5 times, it will likely work the second time
* Use `-S` to show the error if there is one

I did also fiddle with some other curls in the Makefile that have not been noticeably problemmatic.

`rm -r .gotmp/bin/*/mkcert && make`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ $(GOTMP)/bin/darwin_arm64/mkcert $(GOTMP)/bin/darwin_amd64/mkcert $(GOTMP)/bin/l
 	@export TARGET=$(word 3, $(subst /, ,$@)) && \
 	export GOOS="$${TARGET%_*}" GOARCH="$${TARGET#*_}" MKCERT_VERSION=v1.4.4 && \
 	mkdir -p $(GOTMP)/bin/$${GOOS}_$${GOARCH} && \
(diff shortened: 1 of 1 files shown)


stasadev (approved):
It looks good, but I noticed that `mkcert` is downloaded 4 times when I don't ask for it:

Maybe it can be changed to this?:
```diff
-mkcert: $(GOTMP)/bin/darwin_arm64/mkcert $(GOTMP)/bin/darwin_amd64/mkcert $(GOTMP)/bin/linux_arm64/mkcert $(GOTMP)/bin/linux_amd64/mkcert
+mkcert: $(GOTMP)/bin/$(BUILD_OS)_$(BUILD_ARCH)/mkcert
```

---

Edit:
My suggestion about `mkcert` doesn't work when I'm building for a different platform.
I don't know how to fix it, so it's fine with me to ignore it.

rfay:
Yeah, that's just a weakness of the Makefile, not related to this PR I don't think. And it only happens once on an existing system.

However, that probably means that on these test runners, we're downloading all each time we run, which increases the likelihood of failure. Thanks for noticing that!

rfay:
I can see how it ended up this way, because the target OS and ARCH aren't available for use in the target part of the make recipes. I don't see a way to improve it without mucking about quite a lot in the Makefile, and we never noticed this before (thanks for noticing!) 

IMO we should just go with it, and maybe look at it someday. It's been in there a long time, and didn't impact us locally because we always had all of them downloaded.

stasadev:
I agree.

PR #7011, opened 2025-02-22
docs: add quickstart tip on running Symfony Web Debug Toolbar with Apache

brezanac:
This PR adds a quickstart tip to the documentation on how to fix the issue of not being able to use the Symfony Web Debug Toolbar when the Apache web server is used (`webserver_type` in `.ddev/config.yaml` is set to `apache-fpm`).

symf-deb-tool

The Web Debug Toolbar needs support through `.htaccess` in order to find the appropriate Symfony profiler routes.

This support is added by including the official Symfony Apache pack through Composer: https://packagist.org/packages/symfony/apache-pack

```bash
ddev composer require symfony/apache-pack
```

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -1009,6 +1009,13 @@ DDEV automatically updates or creates the `.env.local` file with the database in
     ddev exec symfony server:log
     ```
 
+!!!tip "Web Debug Toolbar support for Apache (`apache-fpm`)"
+    
+    ```bash
+    ddev composer require symfony/apache-pack
+    # When asked to execute the recipe, choose "yes" with "y"
+    ```
+
 ## TYPO3
 
 === "Composer"


rfay:
Thanks for this contribution and for working through what it took you to get here! 

Since the quickstart uses nginx-fpm, and this does not seem to be the kind of thing that an initial Symfony user would need, I'm not convinced this is the right place for it. 

We'd love your note to get great search coverage. Some good ways to do that would be 

1. Put it over in https://ddev.readthedocs.io/en/latest/users/usage/cms-settings/#managing-cms-settings instead of here. (To do that you could just update this PR with a new commit moving the info)
1. If you use Stack Overflow, do a question-and-answer there. When creating the question, you use the checkbox at the bottom to "answer your own question")
2. I invite you to do a guest blog on ddev.com, super easy, explaining how you use symfony with ddev and giving this tip and any other. 
3. Consider creating a DDEV add-on that implements the way you like to use DDEV with Symfony.

rfay:
I'm going to close this as it's not appropriate as is, but your contribution is welcome in any of the above ways.

PR #7012, opened 2025-02-23
test: run ClearDockerEnv() before tests, fixes #472

### LINKED ISSUE #472: Clear docker environment variables after using docker-compose.
When using docker-compose we rely on environment variables for inserting things which are configurable via DDEV. We only set them within the context of the ddev binary, but rather than clearing them we rely on the process ending.

Unfortunately, when running tests, sometimes those environment variables stick around and cause unintended site effects. They could also cause problems if we tried to interact with multiple sites in the same `ddev command`.

Rather than relying on the ddev process finishing and these environment variables automatically going away, we should unset them after the docker-compose call ends. You can see the pain this causes in testing by looking at the testcommon.ClearDockerEnv() method. 

## How to reproduce this:
delete a few calls to testcommon.ClearDockerEnv() in local_test.go and see how things start failing and how confusing it is to figure out what is going on 😢

rfay:
## The Issue

* #472 

It's been a long time. This rarely causes trouble, but I spent a few hours with it today on #7007 so maybe it's time to fix it!

### DIFF
--- a/cmd/ddev/cmd/root_test.go
+++ b/cmd/ddev/cmd/root_test.go
@@ -2,6 +2,14 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	osexec "os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"testing"
+
 	"github.com/ddev/ddev/pkg/config/types"
 	"github.com/ddev/ddev/pkg/ddevapp"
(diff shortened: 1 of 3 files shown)


rfay:
The actual answer to the problem here is to stop using environment variables and move to consistent use of go templates, but the past and compatibility haunt us.  But it looks like this ancient thing pointed out by @beeradb may finally be solved.

stasadev (approved):
Looks good to me.

PR #6980, opened 2025-02-13
fix: do not show router URL in `ddev list` if it's stopped

stasadev:
## The Issue

```
❯ ddev list
┌──────────┬─────────┬──────────────────────┬────────────────────────┬─────────┐
│ NAME     │ STATUS  │ LOCATION             │ URL                    │ TYPE    │
├──────────┼─────────┼──────────────────────┼────────────────────────┼─────────┤
│ ddev.com │ stopped │ ~/code/ddev/ddev.com │                        │ php     │
├──────────┼─────────┼──────────────────────┼────────────────────────┼─────────┤
│ Router   │ stopped │ ~/.ddev              │ http://127.0.0.1:10999 │ traefik │
└──────────┴─────────┴──────────────────────┴────────────────────────┴─────────┘
```

We don't need the router URL if it's stopped.

## How This PR Solves The Issue

Don't show the URL for the `stopped` status.

```
❯ ddev list
┌──────────┬─────────┬──────────────────────┬─────┬─────────┐
│ NAME     │ STATUS  │ LOCATION             │ URL │ TYPE    │
├──────────┼─────────┼──────────────────────┼─────┼─────────┤
│ ddev.com │ stopped │ ~/code/ddev/ddev.com │     │ php     │
├──────────┼─────────┼──────────────────────┼─────┼─────────┤
│ Router   │ stopped │ ~/.ddev              │     │ traefik │
└──────────┴─────────┴──────────────────────┴─────┴─────────┘
```
```
❯ ddev start
❯ ddev list
┌──────────┬─────────┬──────────────────────┬────────────────────────────┬─────────┐
│ NAME     │ STATUS  │ LOCATION             │ URL                        │ TYPE    │
├──────────┼─────────┼──────────────────────┼────────────────────────────┼─────────┤
│ ddev.com │ OK      │ ~/code/ddev/ddev.com │ https://ddev.com.ddev.site │ php     │
├──────────┼─────────┼──────────────────────┼────────────────────────────┼─────────┤
│ Router   │ healthy │ ~/.ddev              │ http://127.0.0.1:10999     │ traefik │
└──────────┴─────────┴──────────────────────┴────────────────────────────┴─────────┘
```

## Manual Testing Instructions

```
ddev poweroff
ddev list # without URL
ddev start
ddev list # with URL
```

PR #7016, opened 2025-02-24
fix: show router errors on `ddev list` and `ddev describe`, for #6980

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/6980#issuecomment-2664268292

Run `ddev list --continuous`, open a new terminal and run `ddev start`

Look at `ddev list --continuous` (but this is not an error!):

image

In a couple of seconds:

image

Run `docker exec -it ddev-router bash -c "killall traefik"` in a new terminal.

Look at `ddev list --continuous` (it's ugly):

image

- Makes the `ddev list` output fancier
- Adds router error to `ddev describe` output because people don't always notice the output from `ddev start`, so give them a second chance to see if there is a problem with router on `ddev describe`.

Run `ddev list --continuous`, open a new terminal and run `ddev start`

Look at `ddev list --continuous`:

image

In a couple of seconds:

image

Run `docker exec -it ddev-router bash -c "killall traefik"` in a new terminal.

Look at `ddev list --continuous`:

image

Run `ddev describe` (new footer that will only be displayed in case of an error with the router):

image

### DIFF
--- a/cmd/ddev/cmd/describe.go
+++ b/cmd/ddev/cmd/describe.go
@@ -233,6 +233,15 @@ func renderAppDescribe(app *ddevapp.DdevApp, desc map[string]interface{}) (strin
 		if len(bindInfo) > 0 {
 			t.AppendRow(table.Row{"Network", "", strings.Join(bindInfo, "\n")})
 		}
+		if !ddevapp.IsRouterDisabled(app) {
+			// If there is a problem with the router, add it to the table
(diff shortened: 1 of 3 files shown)


rfay (approved):
This is fine and works great. Thanks for following it up!

rfay:
(No need to wait for tests to finish after the rebase)

PR #7017, opened 2025-02-24
test: Run quickstart tests with mutagen enabled [skip buildkite]

rfay:
## The Issue

We've been running quickstarts in a linux workflow, but we might get more information running them with mutagen on. 

We may want to think about running them on macOS, but for now we can enable mutagen here.

Turn on mutagen for these tests

### DIFF
--- a/.github/workflows/quickstart.yml
+++ b/.github/workflows/quickstart.yml
@@ -67,4 +67,6 @@ jobs:
 
       - name: Run quickstart test
         run: |
+          # Use mutagen by default as it may cause different behaviors
+          ddev config global --performance-mode=mutagen
           make quickstart-test
--- a/.github/workflows/quickstart.yml
+++ b/.github/workflows/quickstart.yml
@@ -69,4 +69,7 @@ jobs:
         run: |
           # Use mutagen by default as it may cause different behaviors
           ddev config global --performance-mode=mutagen
+          # bats refuses ever to exit until all processes started by it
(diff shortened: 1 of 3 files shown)


rfay:
The quickstart hangs after all the tests are done. Moving to draft until I can look at it.

rfay:
This was so very difficult to sort out. I thought perhaps one of the tests was the problem. I did lots of good binary searching on the tests, and also tested manually in tmate. But the tmate tests always seemed to succeed.

In the end, it turns out that `bats` refuses to ever exit until all background processes create by it have exited. This may be either the background child process itself, or perhaps it has to do with open file descriptors.

Anyway, if a bats test does a `ddev start` with mutagen enabled, in a github actions environment, that starts mutagen. But we don't typically stop mutagen in our environment, although a `ddev poweroff` would have stopped it I think. But with mutagen still running, bats never completes, even though all of its tests are successful.

The fix here is to start the mutagen daemon before the bats tests run. There are several other ways to fix it. `ddev debug mutagen daemon stop` in the cleanup, probably a `ddev poweroff` in the cleanup would fix it, etc. 

Perhaps related to
* https://github.com/bats-core/bats-core/issues/155

stasadev (approved):
> In the end, it turns out that `bats` refuses to ever exit until all background processes create by it have exited. This may be either the background child process itself, or perhaps it has to do with open file descriptors.

Wow, that sounds really challenging! Glad you managed to sort it out.

PR #7020, opened 2025-02-25
docs: add a note about rootless mode, show context in ddev debug test

stasadev:
## The Issue

From Discord https://discord.com/channels/664580571770388500/1343957710604992573

It's good to see all available contexts in `ddev debug test`.
Say explicitly that rootless mode is not supported.


rfay (approved):
Thank you! The context was there hidden in the `docker version` output, but this is better.

Issue #5277, opened 2023-08-18
Add-ons `#ddev-description:` on `install.yaml` can't have `:`


hanoii:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

I have a [`platform:tunnels` host command](https://github.com/hanoii/ddev-platformsh-lite/blob/main/commands/host/platform%3Atunnels) on my add-on and I wanted to [add a removal post_install_action](https://github.com/hanoii/ddev-platformsh-lite/blob/9304021d99c1bde136faa38428d671532f2321e2/install.yaml#L12) with a description mentioning the removal of such port.

If I add a colon (`:`) anywhere on the line after `#ddev-description:` the output of `ddev get` gets truncated at the colon.

### Actual Behavior

Either this working or way of escaping such characters?

rfay:
Thanks for the report!

PR #7022, opened 2025-02-26
fix: the `#ddev-description` stanza in add-on install actions not showing if it's the first line

hanoii:
I just noticed that my [add-on's post install actions](https://github.com/hanoii/ddev-pimp-my-shell/blob/637a873275cff867d82b48e707644df0f1854905/install.yaml#L21-L54) were not showing the `#ddev-description` as I was used to.

I looked at the code and found that the regex was not finding it if the stanza was right next to `- |`. One more newline and it was picked up.

I added an extra newline on the last commit of that file, but trying it with the latest release of the add-on show showcase the issue.

### DIFF
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -146,7 +146,7 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 
 // GetAddonDdevDescription returns what follows #ddev-description: in any line in action
 func GetAddonDdevDescription(action string) string {
-	descLines := nodeps.GrepStringInBuffer(action, `[\r\n]+#ddev-description:.*[\r\n]+`)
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks safe to me. Do you think this was a regression, or just an oversight? I do hope we can find more sophisticated ways to do this sort of thing in the future.

hanoii:
According to [blame](https://github.com/ddev/ddev/blame/244e0b46b9a5f7dc651a710a7ab2cd74935544e2/pkg/ddevapp/addons.go#L149) this goes back to 2 years so maybe it was an oversight.

hanoii:
Also in looking in the issue queue as I vaguely remember this being an issue before, I found https://github.com/ddev/ddev/issues/5277, which although not the same thing, it's worth linking the two.

stasadev (approved):
Looks good to me.

PR #7028, opened 2025-02-27
feat: Laravel 12 quickstart with tests

stasadev:
## The Issue

Laravel 12 was released https://laravel.com/docs/12.x/releases

- Updates quickstart to use Laravel 12
- Adds a dedicated SQLite quickstart for Laravel
- Updates Laravel Installer to make it clearer
- Adds Laravel quickstart tests

https://ddev--7028.org.readthedocs.build/en/7028/users/quickstart/#laravel

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -531,11 +531,12 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
 
     ```bash
     mkdir my-laravel-site && cd my-laravel-site
+
+    # To use MariaDB, apply the following command
     ddev config --project-type=laravel --docroot=public
 
-    # This config uses MariaDB, but if you want to use SQLite instead,
-    # uncomment this line:
-    #ddev config --omit-containers=db --disable-settings-management=true
+    # To use SQLite, uncomment and use the following command instead
+    #ddev config --project-type=laravel --docroot=public --omit-containers=db --disable-settings-management=true
 
     # Temporarily add the Laravel installer
     # as /usr/local/bin/laravel in the web container
@@ -550,6 +551,8 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
 
     # Follow the prompts, select a starter kit of your choice (or none),
     # and agree to run npm commands
+    # (SQLite is used here as other database types would fail due to
+    # the .env file not being ready, which DDEV will fix on 'ddev restart')
     ddev exec laravel new temp --database=sqlite
 
     # 'laravel new' can't install in the current directory right away,


rfay (approved):
Looks to me like it's just what we need! Thanks!

docs/content/users/quickstart.md
```diff
@@ -541,20 +542,21 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    ddev composer run-script post-root-package-install
+    ddev composer run-script post-create-project-cmd
     ddev launch
     ```
 
 !!!tip "Want to use a SQLite database for Laravel?"
     DDEV defaults to using a MariaDB database to better represent a production environment.
 
-    To select the [Laravel 11 defaults](https://laravel.com/docs/11.x/releases#application-defaults) for SQLite, use this command for `ddev config`:
+    To select the [Laravel defaults](https://laravel.com/docs/11.x/releases#application-defaults) for SQLite, use this command for `ddev config`:
```
tyler36, line 553:
```suggestion
    To use a SQLite database when creating a new project, replace the `ddev config` command with the following instead:
```

This removes the link, which is a moving target.

docs/content/users/quickstart.md
```diff
@@ -541,20 +542,21 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
     ```bash
     ddev config --project-type=laravel --docroot=public --omit-containers=db --disable-settings-management=true
     ```
```
tyler36, line 557: A problem with this section is that only applies when starting a new project.
For existing projects you need to:

- update `.env` to `DB_CONNECTION="sqlite"`

```bash
ddev config --project-type=laravel --docroot=public --omit-containers=db --disable-settings-management=true
ddev restart
ddev artisan migrate:fresh
 ```

If you don't update the `.env` you will get the following error (included for searchability): 

```
SQLSTATE[HY000] [2002] php_network_getaddresses: getaddrinfo for db failed: Name or service not known (Connection: mariadb, SQL: select * from `sessions` where `id` = jvwGaqOpdMyrsUZa412ljRqpuaNOIqPSw2NJE2XN limit 1)
```

**Review by tyler36 — CHANGES_REQUESTED** [review]
Generally looks good.

File: docs/tests/laravel.bats [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+  assert_success
+
+  # validate running project
+  run curl -sfI https://${PROJNAME}.ddev.site
+  assert_success
+  assert_output --partial "HTTP/2 200"
+
+  run curl -sf https://${PROJNAME}.ddev.site
+  assert_success
+  assert_output --partial "Laravel"
```
rpkoller, line 38: I wonder if it would make sense to slightly extend the assertion and use `<title>Laravel</title>`instead? would be still prone to potential updates but still more specific than just `Laravel`? (if so it would have to be applied to line 127 as well)

rpkoller (commented):
looks good (tested the composer and laravel isntaller steps), i left one comment on the tests and two observations about applying the laravel installer quickstart:

1) i am not sure if i have tested the laravel installer before, but somehow i only copy and pasted the echo line `echo 'ARG COMPOSER_HOME=/usr/local/composer` then thought i'Ve missed the single quote, after another try it dawned to me that it is a single command. for readability the current version is definitely better but it could be misleading like in my case. therefore i wonder if putting it in a single line would be the more clear choice? 

2) for `ddev exec laravel new temp --database=sqlite` would it make sense to add to the comment before the commant, that the the user should follow the defaults none and for picking no starterkit to install and yes for running the npm build?

docs/tests/laravel.bats
tyler36: > I wonder if it would make sense to slightly extend the assertion and use <title>Laravel</title>instead?

If the database setting are incorrectly set by DDEV, the resulting error page contains the word "Laravel" and the header "<title>Laravel</title>". 

image

If I recall, there were some shenanigans with collation settings and Laravel 11. Would the tests have caught that issue?

rpkoller: i just manually tested and changed the DB_HOST from db to dbs in the .env file. after that `curl -sf https://lari.ddev.site` returns nothing so i suppose the assertion would fail in that case

rpkoller: `curl -sf https://lari.ddev.site` returns nothing while `curl -sfI https://lari.ddev.site`returns: ```HTTP/2 500 
cache-control: no-cache, private
content-type: text/html; charset=UTF-8
date: Fri, 28 Feb 2025 04:20:43 GMT
server: nginx```

tyler36: Ah, so it would be caught by `assert_output --partial "HTTP/2 200"` then. 
Thank you for checking.

rpkoller: yep, even both assertions would fail. neither the 200 nor the Laravel one would pass.

docs/content/users/quickstart.md
stasadev: Nice catch!

I think we should put the SQLite config on a separate tab, where we mention how to migrate from one to the other.

If we add everything on one tab, it will complicate things for newcomers.

stasadev:
> therefore i wonder if putting it in a single line would be the more clear choice?

@rpkoller, I think it was the syntax highlighting that confused you.

I'll replace it with `cat <<'EOF'` like it's done for FrankenPHP.

> for `ddev exec laravel new temp --database=sqlite` would it make sense to add to the comment before the commant, that the the user should follow the defaults none and for picking no starterkit to install and yes for running the npm build?

If people use Laravel Installer, it usually meant that they *want* to use StarterKit. I will make the comment clearer about the choice, thank you!

stasadev:
I updated Laravel Quickstart with Composer (SQLite) and made some small changes to Laravel Installer.

docs/content/users/quickstart.md
```diff
@@ -494,10 +494,36 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
...
+    ```
+
+    To switch an existing Laravel project to SQLite:
+
+    ```bash
+    ddev config --project-type=laravel --docroot=public --omit-containers=db --disable-settings-management=true
+    ddev restart
+    ddev composer run-script post-root-package-install
+    ddev dotenv set .env --db-connection=sqlite
+    ddev composer run-script post-create-project-cmd
```
stasadev, line 526: `post-root-package-install` copies `.env.example` to `.env` if it doesn't exist, then we set the `sqlite` connection in `.env`, and finally run migrations in `post-create-project-cmd`.

tyler36 (approved):
Looks good.

I was able to create a project with the "default" mariadb database, then convert to Sqlite database.

Thank you!

docs/content/users/quickstart.md
```diff
@@ -507,28 +533,37 @@ The Laravel project type can be used for [StarterKits](https://laravel.com/docs/
     mkdir my-laravel-site && cd my-laravel-site
     ddev config --project-type=laravel --docroot=public
```
rpkoller, line 534: there is only one minor nitpick about a detail i've just noticed. line 536 refers to line 534 stating that it is using mariadb, while line 538 uses sqlite instead. the detail i dont understand, with the active default you set the project up for mariadb but in line 553 sqlite is used anyway? that is not entirely clear to me why you use mariadb and sqlite alongside? 

and in general the comment in 536 is hard to comprehend cuz it requires that you jump in-between lines referring to the lines before and after the comment. Plus if the quickstart states "This config uses.."  strictly speaking me as the recipient would expect that line 538 would set up mariadb. how about something like the following instead? 

```
# To use MariaDB apply the following command
ddev config --project-type=laravel --docroot=public

# To use SQLite, uncomment and use the following command instead
#ddev config --omit-containers=db --disable-settings-management=true
```

rpkoller (commented):
thanks for the changes. the syntax highlighting makes it way clearer now and the comment i've raised looks good now. and i'Ve ran through the two tests another time and all worked well. added only one more comment about one small stumbling block i havent noticed in the previous review.

docs/content/users/quickstart.md
stasadev: >  but in line 553 sqlite is used anyway?

It's because if you specify a different database (other than SQLite) at this point, the database migrations will fail, which would cause more confusion. You can try it with:

```
ddev exec laravel new temp
ddev exec laravel new temp --database=mariadb
```

I don’t have access to alter the `.env` content when running `laravel new`.

I tried to make the comment for `ddev exec laravel new temp --database=sqlite` a bit clearer.

> how about something like the following instead?
> `# To use SQLite, uncomment and use the following command instead`

Good idea! I think it's better to add `--project-type=laravel --docroot=public` in this case when we say "instead".

rpkoller: ah thanks for the explanation. i like the improvements you made to the comment! and yep i also agree adding `--project-type=laravel --docroot=public`. looks good, thank you!

PR #7014, opened 2025-02-23
docs: change code refs to include info about Backdrop config storage options, fixes #7013

laryn:
## The Issue

- #7013

DDEV provides helpful information regarding Backdrop's config after importing a database. As of Backdrop 1.28.0, this information is slightly outdated so I'd like to update it.

## How This PR Solves The Issue

Minor touch to adjust language slightly and give additional context about config storage options.

## Manual Testing Instructions

    1. Spin up a Backdrop site.
    2. Export the database.
    3. Import the database.
    4. Read the message that DDEV provides.

## Automated Testing Overview

No additional tests needed as it's just a minor text adjustment.

## Release/Deployment Notes

n/a

PR #7037, opened 2025-02-28
docs: add Backdrop-specific config considerations.

laryn:
## The Issue

Backdrop's config storage options can be confusing with local development, so it would be good to have a short description of the options as discussed in https://github.com/ddev/ddev/pull/7014#issuecomment-2677157148

As well, it is worth mentioning Backdrop's command line tool, `bee`. 

Review rendered at https://ddev--7037.org.readthedocs.build/en/7037/users/usage/cms-settings/#backdrop-specifics

### DIFF
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -34,6 +34,14 @@ This environment variable is set `true` by default in DDEV’s environment, and 
 
 ## CMS-Specific Help and Techniques
 
+### Backdrop Specifics
+
+#### Configuration considerations
+
+By default, Backdrop stores both active and staging configuration in the filesystem. In this case, if you are updating the database (either from a live site, or reloading while working locally) it can be important to also refresh the active configuration directory to make sure your database structure is in line with what your configuration is describing. 
+
+Backdrop also allows you to store your active configuration in the database ([as of Backdrop 1.28.0](https://docs.backdropcms.org/change-records/swappable-config-storage)), which can be simpler while developing locally, especially if you are making use of DDEV's snapshot feature. In this case, the database and the active configuration are effectively bundled and refreshing the database also refreshes the active configuration.
+
 ### Drupal Specifics
 
 #### Drupal Settings Files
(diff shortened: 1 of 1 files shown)


rfay (approved):
Thanks!

docs/content/users/usage/cms-settings.md
```diff
@@ -34,6 +34,21 @@ This environment variable is set `true` by default in DDEV’s environment, and
...
 
+### Backdrop Specifics
+
+#### Configuration considerations
+
+By default, Backdrop stores both active and staging configuration in the filesystem. In this case, if you are updating the database (either from a live site, or reloading while working locally) it can be important to also refresh the active configuration directory to make sure your database structure is in line with what your configuration is describing. 
+
+Backdrop also allows you to store your active configuration in the database ([as of Backdrop 1.28.0](https://docs.backdropcms.org/change-records/swappable-config-storage)), which can be simpler while developing locally, especially if you are making use of DDEV's snapshot feature. In this case, the database and the active configuration are effectively bundled and refreshing the database also refreshes the active configuration.
+
+#### Command line tool
```
rpkoller, line 45: I wonder would it make sense to move the command line part over to the backdrop quickstart? i mean for drupal drush is contained in every quickstart available (except for drupal cms but that is shipping with drush out of the box) so i suppose wouldnt `ddev add-on get backdrop-ops/ddev-backdrop-bee` be better suited in the two backdrop quickstarts? or is bee used less extensively with backdrop compared to drush with drupal?

rfay: It does seem like bee has become adequately standard now that it should just be installed from the add-on right away when setting up a project. That can be another PR of course.

rfay:
There are minor markdown complaints in https://github.com/ddev/ddev/actions/runs/13596373138/job/38014122763?pr=7037, easy to fix with another commit.

docs/content/users/usage/cms-settings.md
laryn: I've removed the CLI piece from this PR and will make another for the quickstarts.

Issue #7039, opened 2025-02-28
wp-cli doesnt work if you change the webroot


nickchomey:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

If you set something other than the project root for `webroot`,  `ddev wp` cli doesnt work. This is because it tries to use `/var/www/html` as the `cwd`. 

Yet, installing bedrock actually works. As it turns out, it is because bedrock adds a `wp-cli.yml` file in the project root

https://github.com/roots/bedrock/blob/master/wp-cli.yml

I propose that https://github.com/ddev/ddev/blob/0951e7848b699f83166029aed7f61e27fb462e3c/pkg/ddevapp/wordpress.go create such a file as well. Also, that there also be some mechanism when changing the `webroot` in a WordPress project-type that also changes the values in that file to match the webroot.

rfay:
Seems like a good idea to me! Thanks for sorting it out. Back when Drupal's drush needed a drushrc.php it used to be created by DDEV, as in https://github.com/ddev/ddev/blob/0951e7848b699f83166029aed7f61e27fb462e3c/pkg/ddevapp/drupal.go#L223-L271

I think just the wordpress.go should do the same basic idea, I'd think it would be done at Start time (as the settings are done, so can be part of the settings)

rfay:
As always, thanks for caring for the WordPress world, and for paying attention to all these things!

jakeparis:
Hi all, very grateful for this solution, as I ran into this issue today. Here in github it looks like the docs update has been merged, but it's not showing up on the stable documentation at https://ddev.readthedocs.io/en/stable/users/usage/cms-settings/#cms-specific-help-and-techniques

rfay:
Thanks @jakeparis - the "latest" docs move to "stable" when the new release happens, planned for Monday, March 31. So right now you'll find this at https://ddev.readthedocs.io/en/latest/users/usage/cms-settings/#wordpress-specifics

PR #7041, opened 2025-03-02
fix: allow usage of wp-cli when docroot has changed in a wordpress project

nickchomey:
## The Issue

- #7039 

Adds a wp-cli.yml file to the app root when the application starts up, setting the path to the app.docroot. Allows for overriding it by removing the `#ddev-generated` signature

I suppose you could create a new wordpress project, change the docroot, install wordpress then run something like `ddev wp plugin list` (which wont work without the yml file)

https://ddev.readthedocs.io/en/stable/users/quickstart/#wordpress

## Automated Testing Overview

I didnt add anything. If a test is required, please let me know what and where to add it.

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -244,6 +244,7 @@ func init() {
 			uploadDirs:           getWordpressUploadDirs,
 			hookDefaultComments:  getWordpressHooks,
 			appTypeSettingsPaths: setWordpressSiteSettingsPaths,
+			postStartAction:      wordpressPostStartAction,
 			appTypeDetect:        isWordpressApp,
 			importFilesAction:    wordpressImportFilesAction,
 		},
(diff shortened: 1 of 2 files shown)


rfay:
Looks like there's a little bit of work to be done for
* https://github.com/ddev/ddev/actions/runs/13617386329/job/38062205667?pr=7041

For  https://github.com/ddev/ddev/actions/runs/13617386417/job/38062205737?pr=7041 you'll need to manually test and sort out the change there.

rfay:
You'll need to solve it yourself IMO, because your change breaks both of these. Happy to help you understand the troubleshooting techniques. Some details about the quickstart tests are in https://ddev.readthedocs.io/en/stable/developers/quickstart-maintenance/

I'll restart both tests in case it was a one-off, but it looks like your changes broke both tests.

rpkoller:
hm played around with the artifact provided by this pr to understand the problem. in regard of the two wp cli based quickstarts, i agree it would be good if it would be possible to only create the yaml file if the default is overridden and the docroot is changed. sure an option might be to update the quickstart and the tests by adding the parameter but that feels suboptimal, the conditional creation would be preferable if possible? in regard of the composer based approach it wouldnt be entirely fixed cuz if the docroot is overridden you could still run into the error. would it make sense to add the yaml file on the list of allowed files that could exist when composer create is run?

rfay:
Solving it with docs is fine, and agreed it's a bit obscure. It would fit nicely into https://ddev.readthedocs.io/en/stable/users/usage/cms-settings/#cms-specific-help-and-techniques - and I imagine you will end up adding other things there?

PR #7046, opened 2025-03-04
build: fix new mkdocs failure on git

rfay:
## The Issue

mkdocs builds are failing, and docscheck builds are failing.

I'm not sure this has anything to do with us, but we might as well solve the complaint

> WARNING -  [git-revision-date-localized-plugin] Unable to find a git directory and/or git is not installed. To ignore this error, set option 'fallback_to_build_date: true'

In addition, we didn't have tmate set up for the docscheck workflow.


stasadev (commented):
Thank you!

PR #7036, opened 2025-02-28
fix: delete traefik images based on the pattern used for ddev-dbserver and use constants for targeting, fixes #6326

rpkoller:
## The Issue

- #6326

## How This PR Solves The Issue
The PR addresses two details. first it fixes #6326 by applying the same pattern used for db images. i've addressed the suggestion by @stasadev using constants instead of the image name. talking of constant i've also did a small cosmetic fix and use now also a constant for addressing the db images instead of the image name. now every if clause in the for loop uses constants instead of image names

## Manual Testing Instructions

```
docker pull ddev/ddev-ssh-agent:v1.24.2
docker pull ddev/ddev-traefik-router:v1.24.2

docker images | grep v1.24.2
ddev/ddev-traefik-router                  v1.24.2                        01f41d37afe2   6 weeks ago     229MB
ddev/ddev-ssh-agent                       v1.24.2                        bd674e56e923   6 weeks ago     136MB

ddev delete images

docker images | grep v1.24.2
(nothing)
```

PR #7047, opened 2025-03-04
fix: use the DBImg constant instead of the db image name

rpkoller:
## The Issue

- #<issue number>

delete-images.go was using `if strings.HasPrefix(tag, "ddev/ddev-dbserver") ` and as i've learned in https://github.com/ddev/ddev/pull/7036 there are constants to use instead of using the image names.


rpkoller:
i've cherrypicked the change and moved it over to https://github.com/ddev/ddev/pull/7036 per randys recommendation. closing this one again.

PR #7049, opened 2025-03-05
build: stop installing chocolatey, fixes #6636, fixes #6344 [skip buildkite]

### LINKED ISSUE #6344: Support Windows and WSL2 ARM64

### LINKED ISSUE #6636: Update WSL installation scripts to skip using Chocolatey

Since v1.23.5 I don't think we need to use Chocolatey in the WSL2 install scripts, and lots of people get hung up with that. We no longer install services, and we *do* ARM64 Windows, which Chocolatey doesn't.

I think we should update the scripts to just get the latest DDEV installer from Github and run it unattended.

### Describe your solution

x

### Describe alternatives

_No response_

### Additional context

_No response_

colans:
## The Issue

- #6636 
- #6344 (I think this is the final fix for Windows ARM64, stopping use of Choco)

We no longer need Chocolatey when installing inside WSL so let's remove it to prevent side effects that users are running into.

It removes the unnecessary dependency.

1. Run the script to install DDEV.
1. Ensure that there are no errors caused by Chocolatey not being installed.
1. Ensure that DDEV works.

## Automated Testing Overview

TBD

## Release/Deployment Notes

TBD

### DIFF
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -233,7 +233,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     In all cases:
 
-    4. Install WSL2 with an Ubuntu distro. On a system without WSL2, run:
+    1. Install WSL2 with an Ubuntu distro. On a system without WSL2, run:
         ```powershell
         wsl --install
         ```
@@ -245,19 +245,19 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
         If you prefer to use another Ubuntu distro, install it and set it as default. For example, `wsl --set-default Ubuntu-24.04`.
 
-    5. In *Windows Update Settings* → *Advanced Options* enable *Receive updates for other Microsoft products*. You may want to occasionally run `wsl.exe --update` as well.
+    2. In *Windows Update Settings* → *Advanced Options* enable *Receive updates for other Microsoft products*. You may want to occasionally run `wsl.exe --update` as well.
 
-    6. Install Docker Desktop. Download [Docker Desktop from Docker](https://www.docker.com/products/docker-desktop/).
-    7. Start Docker Desktop. You should now be able to run `docker ps` in PowerShell or Git Bash.
-    8. In *Docker Desktop* → *Settings* → *Resources* → *WSL2 Integration*, verify that Docker Desktop is integrated with your distro.
-    9. Open a non-administrative PowerShell terminal. However, if the commands below don't work (e.g. "the system cannot find all the information required"), you may need to reboot or open a new Powershell terminal before trying again.  Run [this PowerShell script](https://raw.githubusercontent.com/ddev/ddev/main/scripts/install_ddev_wsl2_docker_desktop.ps1) by executing:
+    3. Install Docker Desktop. Download [Docker Desktop from Docker](https://www.docker.com/products/docker-desktop/).
+    4. Start Docker Desktop. You should now be able to run `docker ps` in PowerShell or Git Bash.
+    5. In *Docker Desktop* → *Settings* → *Resources* → *WSL2 Integration*, verify that Docker Desktop is integrated with your distro.
+    6. Open a non-administrative PowerShell terminal. However, if the commands below don't work (e.g. "the system cannot find all the information required"), you may need to reboot or open a new Powershell terminal before trying again.  Run [this PowerShell script](https://raw.githubusercontent.com/ddev/ddev/main/scripts/install_ddev_wsl2_docker_desktop.ps1) by executing:
 
         ```powershell
         Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;
         iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ddev/ddev/main/scripts/install_ddev_wsl2_docker_desktop.ps1'))
         ```
 
-    10. For unusual browsers and situations that don't automatically support the `mkcert` certificate authority, [configure your browser](configuring-browsers.md).
+    7. For unusual browsers and situations that don't automatically support the `mkcert` certificate authority, [configure your browser](configuring-browsers.md).
 
     Now you can use the "Ubuntu" terminal app or Windows Terminal to access your Ubuntu distro, which has DDEV and Docker Desktop integrated with it.
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -265,16 +265,16 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     You can manually step through the process the install script attempts to automate:
     1. Download and run the [Windows-side installer](https://github.com/ddev/ddev/releases) (used for hosts-file management only).
-    3. In an administrative PowerShell, run `mkcert -install` and follow the prompt to install the Certificate Authority.
-    4. In an administrative PowerShell, run `$env:CAROOT="$(mkcert -CAROOT)"; setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV }`. This will set WSL2 to use the Certificate Authority installed on the Windows side. In some cases it takes a reboot to work correctly.
-    5. In administrative PowerShell, run `wsl --install`. This will install WSL2 and Ubuntu for you. Reboot when this is done.
-    6. **Docker Desktop for Windows:** If you already have the latest Docker Desktop, configure it in the General Settings to use the WSL2-based engine. Otherwise install the latest Docker Desktop for Windows and select the WSL2-based engine (not legacy Hyper-V) when installing. Download the installer from [docker.com](https://www.docker.com/products/docker-desktop/).  Start Docker. It may prompt you to log out and log in again, or reboot.
-    7. Go to Docker Desktop’s *Settings* → *Resources* → *WSL integration* → *enable integration for your distro*. Now `docker` commands will be available from within your WSL2 distro.
-    8. Double-check in PowerShell: `wsl -l -v` should show three distros, and your Ubuntu should be the default. All three should be WSL version 2.
-    9. Double-check in Ubuntu (or your distro): `echo $CAROOT` should show something like `/mnt/c/Users/<you>/AppData/Local/mkcert`
-    10. Check that Docker is working inside Ubuntu (or your distro) by running `docker ps`.
-    11. Open the WSL2 terminal, for example `Ubuntu` from the Windows start menu.
-    12. Install DDEV:
+    2. In an administrative PowerShell, run `mkcert -install` and follow the prompt to install the Certificate Authority.
+    3. In an administrative PowerShell, run `$env:CAROOT="$(mkcert -CAROOT)"; setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV }`. This will set WSL2 to use the Certificate Authority installed on the Windows side. In some cases it takes a reboot to work correctly.
+    4. In administrative PowerShell, run `wsl --install`. This will install WSL2 and Ubuntu for you. Reboot when this is done.
+    5. **Docker Desktop for Windows:** If you already have the latest Docker Desktop, configure it in the General Settings to use the WSL2-based engine. Otherwise install the latest Docker Desktop for Windows and select the WSL2-based engine (not legacy Hyper-V) when installing. Download the installer from [docker.com](https://www.docker.com/products/docker-desktop/).  Start Docker. It may prompt you to log out and log in again, or reboot.
+    6. Go to Docker Desktop’s *Settings* → *Resources* → *WSL integration* → *enable integration for your distro*. Now `docker` commands will be available from within your WSL2 distro.
+    7. Double-check in PowerShell: `wsl -l -v` should show three distros, and your Ubuntu should be the default. All three should be WSL version 2.
+    8. Double-check in Ubuntu (or your distro): `echo $CAROOT` should show something like `/mnt/c/Users/<you>/AppData/Local/mkcert`
+    9. Check that Docker is working inside Ubuntu (or your distro) by running `docker ps`.
+    10. Open the WSL2 terminal, for example `Ubuntu` from the Windows start menu.
+    11. Install DDEV:
 
         ```bash
         sudo apt-get update && sudo apt-get install -y curl
@@ -284,9 +284,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
         sudo apt-get update && sudo apt-get install -y ddev
         ```
 
-    13. In WSL2, run `mkcert -install`.
+    12. In WSL2, run `mkcert -install`.
 
-    14. For unusual browsers and situations that don't automatically support the `mkcert` certificate authority, [configure your browser](configuring-browsers.md).
+    13. For unusual browsers and situations that don't automatically support the `mkcert` certificate authority, [configure your browser](configuring-browsers.md).
 
     You have now installed DDEV on WSL2. If you’re using WSL2 for DDEV, remember to run all `ddev` commands inside the WSL2 distro.
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -216,7 +216,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
         * Verify that your Ubuntu default distro is WSL v2 using `wsl -l -v`.
 
-    2. In an administrative PowerShell run [this PowerShell script](https://raw.githubusercontent.com/ddev/ddev/main/scripts/install_ddev_wsl2_docker_inside.ps1) by executing:
+    2. Open a non-administrative PowerShell terminal. However, if it doesn't work (e.g. "the system cannot find all the information required"), try with an Administrator one (possibly after a reboot) as it may be necessary on some systems.  Run [this PowerShell script](https://raw.githubusercontent.com/ddev/ddev/main/scripts/install_ddev_wsl2_docker_inside.ps1) by executing:
 
         ```powershell
         Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;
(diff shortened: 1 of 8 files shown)


colans:
Yup, the first two of those are what I'm working on now (trying to figure out how to do this stuff in Powershell as I come from the Linux world).

Didn't know about that 3rd item so I'll look into that too!

colans:
@rfay : Where should I stick these EXEs after downloading?

colans:
> Where should I stick these EXEs after downloading?

I just created a new DDEV directory in Program Files.

What's left:
* [ ] Installing `mkcert`, and
* [ ] Doing the other script.

colans:
> I just created a new DDEV directory in Program Files.

In the end, I just ended up with a `mkcert` directory because the other apps use installers.  We're not actually placing the binaries anywhere; the installers take care of that.

Anyway, just took this out of draft; it's ready for review.

scripts/install_ddev_wsl2_docker_desktop.ps1
```diff
@@ -41,8 +34,41 @@ if (-not(wsl -e docker ps) ) {
     throw "Docker Desktop integration with the default distro does not seem to be enabled yet."
 }
 $ErrorActionPreference = "Stop"
-# Install needed choco items
-choco upgrade -y ddev gsudo mkcert
+
+# Install DDEV on Windows to manipulate the local hosts file.  (Also requires mkcert & sudo; see below.)
+$TempDir = $env:TEMP
+$DdevInstallerPath = Join-Path $TempDir "ddev-installer.exe"
```
rfay, line 40: This is just the right approach. I had been thinking of not using the installer, but using the installer is great. Then you get mkcert and gsudo for free, don't have to worry about those at all. Perfect.

scripts/install_ddev_wsl2_docker_desktop.ps1
```diff
@@ -41,8 +34,41 @@ if (-not(wsl -e docker ps) ) {
...
+# TODO: To always fetch the latest EXE (e.g. https://github.com/<OWNER>/<REPO>/releases/latest/download/myprogram.exe),
+# there can't be version numbers in the file name so we need to remove the version number from the release artefact.
+# Until then, we can simply fetch the installer from the latest release, which we'll hardcode.
+Invoke-WebRequest `
+    -Uri "https://github.com/ddev/ddev/releases/download/v1.24.3/ddev_windows_amd64_installer.v1.24.3.exe" `
```
rfay, line 45: You'll find a number of examples of getting latest version in our Dockerfile and in the [install_ddev.sh](https://github.com/ddev/ddev/blob/main/scripts/install_ddev.sh)

scripts/install_ddev_wsl2_docker_desktop.ps1
```diff
@@ -41,8 +34,41 @@ if (-not(wsl -e docker ps) ) {
...
+    -OutFile $DdevInstallerPath
+Start-Process $DdevInstallerPath -Wait
+Remove-Item $DdevInstallerPath
+
+# Install mkcert for Windows.
```
rfay, line 50: This isn't necessary because the installer does it, right?

rfay (changes requested):
I'm afraid I misled you and didn't think properly about mkcert and gsudo, and you're doing the right thing with the installer. So mkcert/gsudo get installed by the installer (I think).

You'll find lots of examples of how to download latest stable version of a package.

Please make sure this handles the architecture correctly. It should be able to handle windows amd64 and arm64.

rfay:
Please use `[skip buildkite]` in your commit messages (if you remember, no harm done if you don't) since this doesn't run any tests on buildkite. I'm not sure why they ran anyway, some rule must be wrong.

rfay:
Rebased this so you won't get messed up by changes to these scripts that went into upstream/main - make sure you get the latest version before continuing, thanks.

colans:
Next up is fetching the correct architecture...

scripts/install_ddev_wsl2_docker_desktop.ps1
colans: Done!

scripts/install_ddev_wsl2_docker_desktop.ps1
colans: Done!

scripts/install_ddev_wsl2_docker_desktop.ps1
```diff
@@ -41,8 +34,41 @@ if (-not(wsl -e docker ps) ) {
...
+    # Because this is an external dependency, pin the release so we're not implicitly trusting their branch forever.
+Invoke-WebRequest `
+    -Uri "https://github.com/FiloSottile/mkcert/releases/download/v1.4.4/mkcert-v1.4.4-windows-amd64.exe" `
+    -OutFile $MkcertBinaryPath
+
+# Install Sudo for Windows.
+Set-ExecutionPolicy RemoteSigned -scope Process
+[Net.ServicePointManager]::SecurityProtocol = 'Tls12'
+# Because this is an external dependency, pin the release so we're not implicitly trusting their branch forever.
+iwr -UseBasicParsing https://raw.githubusercontent.com/gerardog/gsudo/v2.6.0/installgsudo.ps1 | iex
```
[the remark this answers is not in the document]
colans: Done!

colans:
@rfay Ready for your review again:

* Apps that the installer pulls in are no longer installed explicitly.
* The release is no longer hard-coded.
* The architecture is no longer hard-coded.

rfay (commented):
Looking pretty great, looking forward to test drive.

rfay:
Sorry I haven't gotten to manually test yet, will get there. I just haven't spent a day with spare time on WSL2 in the last bit. As you know from your work, it's a lot of work to test. 

One thing I've been thinking about... it may not need to run in admin PS, true? Depends on what the behavior or the DDEV installer is in this situation.

rfay:
Rebased

rfay (changes requested):
I tested both of these scripts and they worked fine. 

There are a couple of things to be done:

- [ ] Currently, this runs the ddev installer and people have to step through it. I think that's OK, but if we're going to leave it like that we need to change the installer to not have that "open release notes" at the end by default, because that opens a browser that hides what happens from there.
- [ ] What do you think about running the installer in silent mode like chocolatey does? Run the installer with `/S /C`
- [ ] I'm pretty sure this no longer needs to be run in admin Powershell, do you agree? Perhaps the silent mode conflicts with that idea. But if it doesn't need to be run in silent mode then the doc has to be improved. 

Awesome work, just what I was hoping for!

colans:
Thanks for reviewing! I was away last week so will catch up on the above soon.

rfay:
Thank *you* for working on this, it's going to be really great. I apologize for requesting other things here, but we want to make this as easy to use as possible.

colans:
> Currently, this runs the ddev installer and people have to step through it. I think that's OK, but if we're going to leave it like that we need to change the installer to not have that "open release notes" at the end by default, because that opens a browser that hides what happens from there.

How do you think we should implement this?  If it's running from either of the Windows installers, skip that step?  Is this context available to the installer?

> What do you think about running the installer in silent mode like chocolatey does? Run the installer with /S /C

I don't like running things in silent mode because then you can't see what's happening, and if something goes wrong, then you don't know what went wrong.  Maybe Windows users are accustomed to this, but it drives me nuts.  (I run Ansible playbooks with ["-vv"](https://www.builddevops.com/post/verbosity-in-ansible) so that's my bias; I like to know what's going on.)

> I'm pretty sure this no longer needs to be run in admin Powershell, do you agree? Perhaps the silent mode conflicts with that idea. But if it doesn't need to be run in silent mode then the doc has to be improved.

When you tested it, did you try it without?  I haven't had a chance to get back into this yet.  Let's have at least one of us try it without, and see if it works.  That should be good enough of a test, right?  If so, we update the docs.  Otherwise, I think we can drop this one.

rfay:
> How do you think we should implement this? If it's running from either of the Windows installers, skip that step? Is this context available to the installer?

I think all you have to do is this, but I'm on mac right now need to check the build. `make windows` will build the windows installer:

```
diff --git a/winpkg/ddev.nsi b/winpkg/ddev.nsi
index 0103c38b3..bf540987d 100644
--- a/winpkg/ddev.nsi
+++ b/winpkg/ddev.nsi
@@ -226,6 +226,7 @@ Var ICONS_GROUP
 
 ; Finish page
 !define MUI_FINISHPAGE_SHOWREADME "${PRODUCT_RELEASE_NOTES_URL}"
+!define MUI_FINISHPAGE_SHOWREADME_NOTCHECKED
 !define MUI_FINISHPAGE_SHOWREADME_TEXT "Review the release notes"
 !define MUI_FINISHPAGE_LINK "${PRODUCT_PROJECT} (${PRODUCT_PROJECT_URL})"
 !define MUI_FINISHPAGE_LINK_LOCATION ${PRODUCT_PROJECT_URL}
```

> I don't like running things in silent mode because then you can't see what's happening, and if something goes wrong, then you don't know what went wrong. Maybe Windows users are accustomed to this, but it drives me nuts. (I run Ansible playbooks with ["-vv"](https://www.builddevops.com/post/verbosity-in-ansible) so that's my bias; I like to know what's going on.)

Agreed, although choco installs are classically that way. If we add a statement like "Press return to continue the DDEV WIndows-side installer, which is only used to update the hosts file, just accept the defaults" then they won't be surprised.

> When you tested it, did you try it without? I haven't had a chance to get back into this yet. Let's have at least one of us try it without, and see if it works. That should be good enough of a test, right? If so, we update the docs. Otherwise, I think we can drop this one.

I have not tested it that way yet, but have high hopes.

rfay:
I did the NSIS (ddev.nsi) fix in 
* https://github.com/ddev/ddev/pull/7147

So you don't have to think about that. Tested locally and it works.  Please consider whether silent-install is worth considering since we only are installing this in order to edit hostnames (when doing WSL2)

rfay:
Rebased; the fix to the Windows DDEV installer is now in HEAD and will be in release hopefully on Friday.

rfay:
Yes, that's what I was thinking. Does it work for you? Thanks!

colans:
I was having problems because running it multiple times (without silencing, in both admin and non-admin terminals) gives me:

```ps
PS C:\Users\username> iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/colans/ddev/3bf49b0b59bbf91e26ab0c0eed70888364a16f99/scripts/install_ddev_wsl2_docker_inside.ps1'))
Detected OS architecture: AMD64; using DDEV installer: amd64
The latest ddev/ddev version is v1.24.4.
Downloading from https://github.com/ddev/ddev/releases/download/v1.24.4/ddev_windows_amd64_installer.v1.24.4.exe...
Start-Process : This command cannot be run completely because the system cannot find all the information required.
At line:80 char:1
+ Start-Process $DdevInstallerPath -Wait
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Start-Process], InvalidOperationException
    + FullyQualifiedErrorId : InvalidOperationException,Microsoft.PowerShell.Commands.StartProcessCommand
```

I realized that you need to start with a fresh terminal.  And then it works (without admin).

However, if I try it with the latest commit (with silencing), I get:

```ps
PS C:\Users\username> iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/colans/ddev/9207f3c94278195af7e48f1659e1e0e270bb4eec/scripts/install_ddev_wsl2_docker_inside.ps1'))
Detected OS architecture: AMD64; using DDEV installer: amd64
The latest ddev/ddev version is v1.24.4.
Downloading from https://github.com/ddev/ddev/releases/download/v1.24.4/ddev_windows_amd64_installer.v1.24.4.exe...
Start-Process : This command cannot be run completely because the system cannot find all the information required.
At line:80 char:1
+ Start-Process $DdevInstallerPath -ArgumentList "/S", "/C" -Wait
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Start-Process], InvalidOperationException
    + FullyQualifiedErrorId : InvalidOperationException,Microsoft.PowerShell.Commands.StartProcessCommand
```

This happens in both admin and non-admin terminals.

However, I tried with the first commit again after a few tries, both admin and non-admin, just to confirm, and then I got the original error again so it stopped working entirely.  There's something non-deterministic going on with this client laptop that I don't understand.  (It's rather locked down, so maybe there's some AI guardian sometimes locking me out...)

Anyway, I got it working once!  Can you try over there so we can see if the problem is my device?  With both commits please.

rfay:
Rebased just on principle.

rfay:
I'm not sure how you were able to even try it as non-admin with the `Requires RunAsAdministrator` in there.

I removed that in https://github.com/ddev/ddev/pull/7049/commits/dfa4d21596d57344beb1fd153a0c6e9692daff52

Please give it another try. 

My strategy is to check out the project, run `Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;` so it will let me run a script, then run the script with (for example) `.\install_ddev_wsl2_docker_inside.ps1`

It seemed to work fine for me after the RunAsAdministrator was removed.

Don't forget that the docs will have to be updated.

rfay:
I just tried it with 
```
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ddev/ddev/dfa4d21596d57344beb1fd153a0c6e9692daff52/scripts/install_ddev_wsl2_docker_inside.ps1'))
```
and it seemed to do fine.

colans:
> I'm not sure how you were able to even try it as non-admin with the Requires RunAsAdministrator in there.

Thanks for removing that; I had it locally, and forgot to push it.

The problem was that I could only run it in a non-admin terminal *with* the admin requirement (because this laptop has software that lets you do admin stuff with a `gksudo`-type pop-up), and then it just didn't work at all in an admin terminal *with the admin requirement*.  After you take that out, it doesn't work in a regular terminal, but does in an admin one, even though sometimes a reboot is required first.  It's all backwards, but at least I understand the rules well enough now to document how to get it rolling on these machines.  Thanks for your patience!

Will push a commit for the docs now...

colans:
@rfay : As long as you're okay with the docs update, I think we're good?

rfay:
We're close if it actually works for you in non-admin PS.  (Note that the installer does the escalation.)

rfay:
Since commits of this to upstream/main immediately affect people's install, we have to be more careful than usual, it won't be after next release. OTOH if we get reports of problems, we can act.

colans:
> We're close if it actually works for you in non-admin PS. (Note that the installer does the escalation.)

Unlike on a Linux system (where I'd be able to comment), I'm going to have to defer to your judgment on this.  I really don't know enough about how Windows works to determine if it's this app, a Windows thing, or some wacky security software that my client uses.

What happens here is that I get the escalation pop-up, authenticate, and then it fails with that message.

If I start from an admin terminal, it works.  So 🤷‍♂️ .

I thought it be best handled by what I put in the docs, where it should work without admin, but may not in all cases.  I'm open to better ways to handle this though.

> Since commits of this to upstream/main immediately affect people's install, we have to be more careful than usual, it won't be after next release. OTOH if we get reports of problems, we can act.

I could also argue both sides of that one too.  No rush here as we can just run the script from the branch commit.

rfay:
I sure don't like that it doesn't work for you in your environment. 

Is your Administrative PS using the same (administrator) user, or a different user?

rfay:
Thanks! I ran the with-docker-desktop one last night when testing proxies etc, and it worked fine!

rfay:
Good job, thanks. The installer should be installing `mkcert.exe` on the Windows side

I think perhaps @colans and I didn't hit that because mkcert.exe was already in the path. Or it could be that a new windows (or reloaded PATH) has to be done after the DDEV installer runs.

rfay:
(However, note that the current scripts do not use choco at all, so if the script ran choco, it wasn't the right script)

rfay:
Here's the reason:
https://github.com/ddev/ddev/blob/4f0061093422a1fd566465558e42400cc87b6d5e/winpkg/ddev.nsi#L500-L505

colans:
Having unique installer names is a good move so deleting is no longer our problem.  I usually go with something like `filename-timestamp.ext`, but GUIDs work too. 😉  Thanks for figuring it out!

rfay:
I think it might work now, your review is appreciated. Apparently the mkcert wasn't installed immediately; it now waits for completion. No idea why.

It also now works with the current released installer; but the uninstall won't uninstall without the complaint until the next release (it tries to remove the directory when there's still a file in it). But the new installer does remove successfully.

docs/content/users/install/ddev-installation.md
```diff
@@ -265,17 +264,11 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
...
-
-    1. Install [Chocolatey](https://chocolatey.org/install):
-        ```powershell
-        Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;
-        iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))`
-        ```
-    2. In an administrative PowerShell, run `choco install -y ddev mkcert`.
+    1. Download and run the [Windows-side installer](https://github.com/ddev/ddev/releases) (used for hosts-file management only).
     3. In an administrative PowerShell, run `mkcert -install` and follow the prompt to install the Certificate Authority.
     4. In an administrative PowerShell, run `$env:CAROOT="$(mkcert -CAROOT)"; setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV }`. This will set WSL2 to use the Certificate Authority installed on the Windows side. In some cases it takes a reboot to work correctly.
```
stasadev, line 269: 2. is removed, the list numbers should be updated.

And `mkcert -install`, `mkcert -CAROOT` here in manual instructions, I think it should be `"C:\Program Files\DDEV\mkcert.exe" ...`

PR #7050, opened 2025-03-05
fix: nginx.conf should let index.php handle 404 errors for media files

barbieswimcrew:
## The Issue

Media stanza in `nginx.conf` doesn't work as expected in some cases:

| Framework | 404 `.png` handled by | 404 `.txt` handled by | Description |
| --------- | -------------- | ----------- |  ----------- |
| Shopware6 | nginx (this PR: app) | app | without PR `.png` is handled by nginx (this is wrong) |
| Magento1 | nginx (this PR: app) | app | has the same problem as Shopware6 |
| Magento2 | nginx (this PR: app) | app | has the same problem as Shopware6 |
| Typo3 | nginx (this PR: app) | app | has the same problem as Shopware6 |
| Wordpress | nginx (this PR: app) | app | has the same problem as Shopware6 |
| Laravel | app | app | has `error_page 404 /index.php;` |
| Symfony | app | app | because nginx doesn't have `Media` stanza |
| Drupal | app | app | Media stanza has `try_files $uri @rewrite;` |

- Adds links to upstream nginx files.
- Updates Shopware6, Magento1 and Wordpress nginx.conf with `try_files`.
- Removes Media stanza from Laravel, Magento2 and Typo3 (because upstream doesn't have it.)

Try quickstarts for Shopware6, Magento1, Typo3, Wordpress, Laravel

The below pages should show the same 404 from the app:

```
ddev launch /test/image.png
ddev launch /test/some.txt
```

Expected result (404 page from app):

image

Unexpected result (404 page from nginx):

image

### DIFF
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
@@ -1,4 +1,5 @@
 # ddev backdrop config
+# https://github.com/perusio/drupal-with-nginx
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
@@ -1,4 +1,6 @@
 # ddev backdrop config
+# https://www.drupal.org/project/drupal/issues/2937161
+# https://www.drupal.org/project/drupal/issues/3336659
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
@@ -1,5 +1,5 @@
 # ddev backdrop config
-# https://github.com/perusio/drupal-with-nginx
+# https://docs.backdropcms.org/documentation/system-requirements#web-server-details
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
@@ -1,6 +1,4 @@
 # ddev backdrop config
-# https://www.drupal.org/project/drupal/issues/2937161
-# https://www.drupal.org/project/drupal/issues/3336659
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal.conf
@@ -1,4 +1,6 @@
 # ddev drupal config (drupal8+)
+# https://www.drupal.org/project/drupal/issues/2937161
+# https://www.drupal.org/project/drupal/issues/3336659
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal.conf
@@ -1,4 +1,5 @@
-# ddev drupal config (drupal8+)
+# ddev drupal config (obsolete), should point to latest stable drupal version
+# This should no longer be in use.
 # https://www.drupal.org/project/drupal/issues/2937161
 # https://www.drupal.org/project/drupal/issues/3336659
 
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
@@ -1,6 +1,4 @@
 # ddev drupal6 config
-# https://www.drupal.org/project/drupal/issues/2937161
-# https://www.drupal.org/project/drupal/issues/3336659
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
@@ -1,4 +1,6 @@
 # ddev drupal6 config
+# https://www.drupal.org/project/drupal/issues/2937161
+# https://www.drupal.org/project/drupal/issues/3336659
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
@@ -1,4 +1,5 @@
 # ddev drupal6 config
+# https://github.com/perusio/drupal-with-nginx
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
@@ -1,4 +1,5 @@
 # ddev magento2 config
+# https://github.com/magento/magento2/blob/2.4-develop/nginx.conf.sample
 
 #ddev-generated
 # If you want to take over this file and customize it, remove the line above
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
@@ -58,15 +58,6 @@ server {
         fastcgi_pass_header "X-Accel-Redirect";
     }
 
-    # Expire rules for static content
-
-    # Media: images, icons, video, audio, HTC
-    location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|webp|htc)$ {
-        expires 1M;
-        access_log off;
-        add_header Cache-Control "public";
-    }
-
     # Prevent clients from accessing hidden files (starting with a dot)
     # This is particularly important if you store .htpasswd files in the site hierarchy
     # Access to `/.well-known/` is allowed.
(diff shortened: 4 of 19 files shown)


rfay:
Thanks for the contribution!

Since Shopware is based on Symfony, it raises questions about 
* Laravel
* Sulu
* Symfony itself
* Others? 

Could you take a look at those and see if they work (quickstarts at https://ddev.readthedocs.io/en/stable/users/quickstart/)

Also, would you mind pushing up a trivial Shopware6 project that demonstrates this problem, so we can see the before and after? Create a README.md on the repository explaining how to demonstrate the problem. Thanks! (Or consider forking and updating the repo that we use in our automated tests, https://github.com/ddev/test-shopware6 ) But it's really important to have a demonstration of the problem.

I imagine we have to change our testing on this when we understand it. Our automated tests (https://github.com/ddev/ddev/blob/6c62ea63106563fc201c458c849c2aa4e8cd06da/pkg/ddevapp/ddevapp_test.go#L203-L218) do in fact successfully get an image. So it will be interesting to know what the difference is in your situation.

rfay:
It would be unusual in any CMS for files to have to be downloaded using index.php; that has major performance implications. The only place I'm familiar with this model is when files are private (access is controlled via PHP). Drupal's "private files" mechanism uses this. But most files on most CMSs aren't delivered with PHP's involvement at all, they are delivered by the webserver (nginx or apache) and not sent to the php-fpm.

barbieswimcrew:
I agree with you @rfay that this brings performance issues with it. However, the situation is as follows:

1. Please see the official NGINX configuration of Shopware. There it is implemented as in my PR via try_files: https://developer.shopware.com/docs/resources/references/config-reference/server/nginx.html

2. In a pure Symfony application (i.e. without Shopware) it is also implemented in the same way. Please refer to the NGINX Config of Symfony, especially at line 8: https://symfony.com/doc/current/setup/web_server_configuration.html

3. Even in your nginx-site-symfony.conf it is like this, please check line 28:  https://github.com/ddev/ddev/blob/main/pkg/ddevapp/webserver_config_assets/nginx-site-symfony.conf

You could also remove the entire block

```nginx
    # Media: images, icons, video, audio, HTC
    location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|webp|htc)$ {
        expires 1M;
        access_log off;
        add_header Cache-Control "public";
    }
```

then it would also work. I just wanted to solve it in a minimally invasive way in my PR. I simply believe that your nginx-site-shopware6.conf is in an outdated state and should be adapted to the current recommendation of Shopware itself.

rfay:
I would prefer that we use the recommended nginx config (or close) for this (and similarly on related configs)

rfay:
Bottom line though... Please demonstrate the problem and help us to understand it before we make changes, as requested above.

tyler36:
> Since Shopware is based on Symfony, it raises questions about ... laravel

I didn't have a problem on Laravel 11 or 12.

## Test

1. Add a valid image file. Eg. `public/img/beach.jpg`
1. Add the following code to `resources/views/welcome.blade.php` (Laravel 11); replacing the filename as required

    ```html
        <img src="/img/beach.jpg" alt="" srcset="">
        <img src="{{ asset('/img/beach.jpg') }}" alt="" srcset="">
    ```

1.  View page and confirm the image is displayed twice.

image

stasadev (approved):
Looks good to me.

I think there's a bit of misunderstanding here. @barbieswimcrew is explaining that the app didn't behave as expected when there was a broken image.

Here's the demonstration:

- When trying to open a non-existent image with a `.jpg` extension, it's handled by nginx, not by the app.
- However, when trying to open a different extension, like `.jpg2`, it's handled by the app.

demo-shopware

The error from Symfony provides more details (and it looks different in production). The goal is to handle these 404s so they display what the app shows for 404 errors, not the default "404 Not Found" from nginx.

---

The situation is not the same with Laravel. Any routing seems to be handled by the app, no matter what I try to open, so I don't see raw 404 nginx:

laravel-404

Laravel has this nginx config:

```
error_page 404 /index.php;
```

barbieswimcrew:
Thank you @stasadev - i really had no idea how to describe this :-D 

I noticed the whole thing because I was playing around with the Symfony Asset Mapper component in a Shopware example project. In DEV mode, the asset mapper tricks the system into believing that the file you want to retrieve via {{ asset('...') }} is in the public directory, when in fact it is not. Only in PROD is the relevant file copied to the public directory using asset-map:compile. Please have a look at: https://symfony.com/doc/current/frontend/asset_mapper.html#serving-assets-in-dev-vs-prod

rfay:
I guess we need to update our test framework (hasn't been updated for 2 years) and maybe show how to demonstrate this.

So @stasadev I'm afraid I still don't really understand. You're saying that shopware wants to handle 404s on images itself. But does that mean that the normal case is that shopware handles the normal delivery of the images? Or that it only reacts on a 404?

- [ ] Update OP here to show the actual situation and describe it completely
- [ ] Compare our nginx config for shopware to the one suggested by shopware, make sure they're similar where they can be.
- [ ] Improve our test in ddevapp_test.go so that it uses this feature
- [ ] Update the ddev/test-shopware6 repository and make sure it can demonstrate this

stasadev:
We have the same problem in several frameworks, all because of Media nginx stanza without `try_files`:

| Framework | 404 `.png` handled by | 404 `.txt` handled by | Description |
| --------- | -------------- | ----------- |  ----------- |
| Laravel | app | app | has `error_page 404 /index.php;` |
| Symfony | app | app | because nginx doesn't have `Media` stanza |
| Drupal | app | app | Media stanza has `try_files $uri @rewrite;` |
| Shopware6 | app (PR) | app | without this PR `.png` is handled by nginx |
| Magento1 | nginx | app | has the same problem as Shopware6 |
| Typo3 | nginx | app | has the same problem as Shopware6 |
| Wordpress | nginx | app | has the same problem as Shopware6 |

rfay:
Fantastic analysis!

In general we should try to use the basic nginx suggested by the CMS; sometimes that's not really available. 

I'm surprised to see Drupal on the list here. Didn't think it processed image 404 in index.php.

This could be a good chance for an nginx config update; if you think that would be better done in another PR we could do that and pull this one where it's at or with minor additional changes.

rfay:
Thanks, that's fine. If you have upstream references for proper nginx format, please provide the links, both here and in the comments in the code.

rfay:
I moved to draft so we don't accidentally pull until you're ready.

rfay:
Drupal nginx suggestions:

* https://www.drupal.org/project/drupal/issues/2937161
* https://www.drupal.org/project/drupal/issues/3336659

stasadev:
- I added links to upstream nginx files.
- I updated Magento1, Typo3 and Wordpress conf with `try_files`.
- I removed Media stanza from Laravel (because upstream doesn't have it.)

More careful review and comparison of upstream vs DDEV nginx config should be done in a different PR.
(And it would be good to have all quickstart tests before doing this.)

pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
```diff
@@ -1,4 +1,6 @@
 # ddev backdrop config
+# https://www.drupal.org/project/drupal/issues/2937161
+# https://www.drupal.org/project/drupal/issues/3336659
```
rfay, line 3: Backdrop is probably not similar enough to Drupal8+ to deserve these links.  DDEV's drupal7 config is based on time-tested situation like backdrop.

pkg/ddevapp/webserver_config_assets/nginx-site-drupal.conf
```diff
@@ -1,4 +1,6 @@
 # ddev drupal config (drupal8+)
```
rfay, line 1:
```suggestion
# ddev drupal config (obsolete), should point to latest stable drupal version
# This should no longer be in use.
```

pkg/ddevapp/webserver_config_assets/nginx-site-drupal6.conf
```diff
@@ -1,4 +1,6 @@
 # ddev drupal6 config
+# https://www.drupal.org/project/drupal/issues/2937161
+# https://www.drupal.org/project/drupal/issues/3336659
```
rfay, line 3: Our configs for drupal6/7 are probably as good as anything is going to get. These links aren't valid.

rfay:
```suggestion
```

pkg/ddevapp/webserver_config_assets/nginx-site-backdrop.conf
rfay:
```suggestion
```

stasadev:
Updated the first post instructions.

pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
```diff
@@ -1,4 +1,5 @@
 # ddev magento2 config
+# https://github.com/magento/magento2/blob/2.4-develop/nginx.conf.sample
```
rfay, line 2: I'm amazed that we didn't have to change things here!

stasadev (approved):

rfay (changes requested):
Amazing research, so much appreciated! Please update title and OP so we can find this in the future when needed.

pkg/ddevapp/webserver_config_assets/nginx-site-magento2.conf
stasadev: It does look different, but I only deleted our media stanza, thanks for noticing this.

rfay:
DDEV's drupal6/7/backdrop nginx configs are probably descended from https://github.com/perusio/drupal-with-nginx, which was the approach I used to use for deploying sites back in those days.

rfay:
Backdrop's recommended nginx config is mighty minimal,

https://docs.backdropcms.org/documentation/system-requirements#web-server-details

rfay:
We'll try to get this in early in the upcoming release cycle.

rfay (approved):
I'm good with this, as @stasadev explained. For shopware6 I had to switch to `APP_ENV=prod` to get the described behavior instead of just a stack trace.

@barbieswimcrew we need you to review this please and make sure it still addresses your original issue. See https://github.com/ddev/ddev/pull/7050#issuecomment-2702440513

stasadev (approved):
Tested it again, and it behaves exactly as I expected.

barbieswimcrew:
> I'm good with this, as @stasadev explained. For shopware6 I had to switch to `APP_ENV=prod` to get the described behavior instead of just a stack trace.
> 
> @barbieswimcrew we need you to review this please and make sure it still addresses your original issue. See [#7050 (comment)](https://github.com/ddev/ddev/pull/7050#issuecomment-2702440513)

Looks good to me! Thanks for your work! ❤️

rfay:
Thanks for all your study on this. We'll get it in now, hopefully no side-effects!

Issue #5747, opened 2024-01-28
Backdrop CMS: Write ddev-specific settings in settings.local.php instead of directly to settings.php


klonos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

As it is currently, ddev writes this bit at the very end of the default `settings.php` file in Backdrop:
```php
// Automatically generated include for settings managed by ddev.
$ddev_settings = dirname(__FILE__) . '/settings.ddev.php';
if (getenv('IS_DDEV_PROJECT') == 'true' && is_readable($ddev_settings)) {
  require $ddev_settings;
}
```

Right before this, the default `settings.php` file in Backdrop contains the following:
```php
/**
 * Include a local settings file, if available.
 *
 * To make local development easier, you can add a settings.local.php file that
 * contains settings specific to your local installation, or to any secondary
 * environment (staging, development, etc).
 *
 * Typically used to specify a different database connection information, to
 * disable caching, JavaScript/CSS compression, re-routing of outgoing emails,
 * Google Analytics, and other things that should not happen on development and
 * testing sites.
 *
 * This local settings file can be ignored in your Git repository, so that any
 * updates to settings.php can be pulled in without overwriting your local
 * changes.
 *
 * Keep this code block at the end of this file to take full effect.
 */
if (file_exists(__DIR__ . '/settings.local.php')) {
  include __DIR__ . '/settings.local.php';
}
```

As the comments explain, keeping your `settings.php` file "clean" makes it easier/safer to pull and push changes without having to worry about overwriting things.

### Describe your solution

Writing the ddev-specific settings in `settings.local.php` will have no effect on current functionality (`settings.ddev.php` will still be loaded - only instead of directly from `settings.php` it will be loaded from `settings.local.php`), while at the same time:
- both `settings.ddev.php` as well as `settings.local.php` will be git-ignored (since Backdrop git-ignores `/settings*.php` by default - see https://github.com/backdrop/backdrop/blob/1.x/.gitignore#L5)
- `settings.php` will be left alone, and people can do things like `git add .` without having to worry that they might be including changes that should not be committed.

laryn:
Great idea.

rfay:
The reason we use `settings.ddev.php` is that DDEV owns it. DDEV tries to make as few changes as possible to anything about your system or project. And `settings.local.php` is intended for changes custom to the system it's on (and not to be in source control). 

[...]

PR #7052, opened 2025-03-06
feat: use settings.local.php by default with Backdrop

laryn:
## The Issue

- #5747

Changing `settings.php` is not ideal because folks may not want to commit changes to that file to their repo. Backdrop by default checks for the existence of `settings.local.php`, and `.gitignore`s it if it exists, but does not include one out of the box. It seems like it should be easy to simply stick the DDEV code into a new `settings.local.php` instead of the main file.

It's a bit of a guess as to what needs to happen, but simply changing `settings.php` to `settings.local.php` in the Backdrop script, and removing the old `settings.php` file (which has become outdated from the upstream copy, as an aside), replacing it with a slimmed down `settings.local.php`.


rfay:
Automated tests don't agree yet :(

laryn:
I think I just found out how to open those failing checks up and see the information inside. 🤦 

Looks like maybe the automated test `TestDrupalBackdropIncludeSettingsDdevInNewSettingsFile` needs some work... or to be split apart like the other functions I mentioned.

rfay:
Thanks so much for this PR and for taking care of the Backdrop community @laryn !

We're happy to help get this to the finish line, but if you're not back by April 30 I'll close the PR for now, but it can be reopened again later.

rfay:
We welcome you back to it, say the word, and we're happy to help. Great to have you here!

PR #7053, opened 2025-03-06
feat: backdrop add bee to quickstart

laryn:
## The Issue

In an earlier PR, I was going to add instructions on how to install the Bee CLI add-on for Backdrop. It was suggested that it might be more useful to include it by default in the Backdrop Quickstart.

I'm not 100% sure that what I've done will work or is the right place for it, but I found a spot where it was doing some Drush-related processing and added the `ddev addon get` line there.

Run instructions from https://ddev--7053.org.readthedocs.build/en/7053/users/quickstart/#backdrop

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -17,6 +17,8 @@ To get started with [Backdrop](https://backdropcms.org), clone the project repos
     mv backdrop/* my-backdrop-site/ && rm -rf backdrop
     cd my-backdrop-site
     ddev config --project-type=backdrop
+    # Add the Bee CLI add-on.
+    ddev add-on get backdrop-ops/ddev-backdrop-bee
     ddev start
     ddev launch
     ```
@@ -35,6 +37,9 @@ To get started with [Backdrop](https://backdropcms.org), clone the project repos
 
     # Set up the DDEV environment:
     ddev config --project-type=backdrop
+
+    # Add the Bee CLI add-on.
+    ddev add-on get backdrop-ops/ddev-backdrop-bee
 
     # Start the project
     ddev start
@@ -157,7 +162,7 @@ Further information on the DDEV procedure can also be found in the [Contao docum
 Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing one.
 
 !!!tip "Compatibility with Craft CMS 3"
-    The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
+The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
 Environment variables will be automatically added to your `.env` file to simplify the first boot of a project. For _new_ installations, this means the default URL and database connection settings displayed during installation can be used without modification. If _existing_ projects expect environment variables to be named in a particular way, you are welcome to rename them.
 
@@ -218,7 +223,7 @@ In order for `ddev craft` to work when Craft is installed in a subdirectory, you
 Read more about customizing the environment and persisting configuration in [Providing Custom Environment Variables to a Container](./extend/customization-extendibility.md#environment-variables-for-containers-and-services).
 
 !!!tip "Installing Craft"
-    Read more about installing Craft in the [official documentation](https://craftcms.com/docs).
+Read more about installing Craft in the [official documentation](https://craftcms.com/docs).
 
 ## Drupal
 
@@ -399,7 +404,7 @@ ddev launch $(ddev drush uli)
     ```
 
 !!!tip "How to update?"
-    Upgrade Grave core:
+Upgrade Grave core:
 
     ```bash
     ddev exec gpm selfupgrade -f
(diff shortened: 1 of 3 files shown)


rpkoller:
the quickstart refers to the quickstart markdown file which can be found at: `docs/content/users/quickstart.md`. the new project example would look like something like that with your proposed changes: 

=== "New projects"

    ```bash
    mkdir my-backdrop-site && cd my-backdrop-site
    curl -LJO https://github.com/backdrop/backdrop/releases/latest/download/backdrop.zip
    unzip ./backdrop.zip && rm -f backdrop.zip && mv -f ./backdrop/{.,}* . ; rm -rf backdrop
    # Add Bee CLI add-on.
    ddev add-on get backdrop-ops/ddev-backdrop-bee
    ddev start
    ddev launch
    ```

laryn:
Oops! Sorry. I'll revise the PR.

rfay:
Moved to draft awaiting update. Thanks for the work on this!

stasadev (changes requested):
Please update the quickstart tests too https://github.com/ddev/ddev/blob/main/docs/tests/backdrop.bats

rpkoller (commented):
one detail i've realized. now that we have bee available we should not only add it but actually use in the quickstart. first i thought it would make sense to simply add 

```
ddev bee si --username=admin --password=admin --db-name=db --db-user=db --db-pass=db --db-host=db --auto
```

after `ddev start`... but then it dawned on me it would be also possible to drop all the curl and unzip jumping through hoops and simply use `ddev bee download-core` so the backdrop quickstart would look like that: 

```
mkdir my-backdrop-site && cd my-backdrop-site
ddev add-on get backdrop-ops/ddev-backdrop-bee
ddev start
ddev bee download-core
ddev bee si --username=admin --password=admin --db-name=db --db-user=db --db-pass=db --db-host=db --auto
ddev launch
```

docs/content/users/quickstart.md
```diff
@@ -6,18 +6,23 @@ Before proceeding, make sure your installation of DDEV is up to date. In a new a
...
-    cd my-backdrop-site
+    mkdir my-backdrop-site && cd my-backdrop-site
     ddev config --project-type=backdrop
+    # Add the official Bee CLI add-on
+    ddev add-on get backdrop-ops/ddev-backdrop-bee
     ddev start
+    # Download Backdrop core
+    ddev bee download-core
+    # Sync files between host and container (only if Mutagen is enabled)
+    ddev mutagen sync
```
stasadev, line 22: @rpkoller,

I think we should have `ddev mutagen sync` in the quickstarts, not only during tests, because the Mutagen may or may not sync properly, but with this we are 100% sure that it will work.

stasadev (approved):
Rebased and refactored to use `ddev bee` per @rpkoller suggestion, thanks!

docs/content/users/quickstart.md
rpkoller: that makes sense. `ddev composer create` or `require` commands do a mutagen sync already arent they?

rpkoller (approved):
looks good to me. thank you!

docs/content/users/quickstart.md
rfay: > ddev composer create or require commands do a mutagen sync already arent they

Yes. In tests, the key issue is to make sure that things are synced before asserting them, that's why these tests have showed up a few things, they assert far faster than a human would do anything.

rfay: However... in a test, the sync never does any harm. In the actual quickstart it shouldn't be added unless it's fundamental, which is quite unusual.

stasadev: And if people have copied the whole quickstart as one command, it works very fast and we have files added to the project, just like here after `ddev start` we run `ddev bee download-core`, we should have `ddev mutagen sync`.

rfay: Additional idea, add `ddev mutagen sync` to the bee command in the add-on. But since bee, like wp-cli, does things inside the container that people expect to show up immediately on the host, it's a good thing.

stasadev: I opened:

- https://github.com/backdrop-ops/ddev-backdrop-bee/pull/8

I'll change tests here to use my fork, to see if it works well.

stasadev:
Coming back here after these two PRs are merged and a new `backdrop-ops/ddev-backdrop-bee` release is created:

- https://github.com/backdrop-ops/ddev-backdrop-bee/pull/8
- #7124

rpkoller:
this iss might need another rebase. the quickstart is failing with the openmage test that got fixed

rfay:
Rebased, but it's WIP anyway.

stasadev:
Rebased and it's good to merge after another round of manual testing.

stasadev (approved):
It works 🎉

rfay:
Does it still look ok to you @laryn ? Thanks!

Could you take another look @rpkoller ?

rpkoller:
everything still works. there havent been any changes since i last looked aside rebases? but also tested another time. the only nitpick i would have that the existing projects could strictly speaking also called Git clone plus for the git clone you dont end up with a working site at the end. cuz the site install step is missing due to the fact that there are none actionable examples for importing a database and importing files. at the moment ddev launch brings you into the backdrop installer.

stasadev:
> there havent been any changes since i last looked aside rebases?

No changes.

> Git clone plus for the git clone you dont end up with a working site at the end.

At least we say:
```
# Use your own repository URL, this is an example
```

> due to the fact that there are none actionable examples for importing a database and importing files

I don't think we should add real URLs from https://github.com/ddev/test-backdrop, because people can misuse it, and override their own database or files.

rpkoller:
i agree adding actual files referenced there would be a bad idea (and i have not implied that). but having instructions to set up a CMS that has that kind of none actionable steps always puzzled me personally and makes me freeze cuz i am uncertain how to proceed, afraid of doing anything wrong. that is in particular relevant for people unfamiliar with a CMS (how to get to such a database and those files, both sound like necessary/mandatory steps - that is me thinking out loud in the role of a naive none tech savvy new user) and/or even ddev (uncertain about the implications and exact functioning of import-db and and import-files). i always consider a quickstart as something i dont have to think about and strictly speaking i am able to copy and paste entirely and i am able to without previous knowledge to set up a running site within seconds - for the backdrop git clone approach that is not the case.

PR #3727, opened 2022-03-23
Styleguide

nico-loeber:
## The Problem/Issue/Bug:

When contributing to DDEV or developing DDEV related projects, developers might miss assets like logos.

## How this PR Solves The Problem:

This styleguide includes different variants of the DDEV logo and information on how to use them.

Also I have added some additional information on the the usage of colors and spelling of the DDEV word mark, as discussed with @rfay.

## Manual Testing Instructions:
N/A

## Automated Testing Overview:
N/A

## Related Issue Link(s):
N/A

## Release/Deployment notes:
N/A

PR #7055, opened 2025-03-06
docs: update mkdocs logo, update word/figurative mark svg for dark mode

stasadev:
## The Issue

Docs logo is not branded:

image

Can't see the brand logo in dark mode:

https://ddev.readthedocs.io/en/stable/developers/brand-guide/

image

- Updates the docs logo to use white logo without words
- Adds `@media(prefers-color-scheme:dark){.cls-2{fill:#e1ded8;}}` to `Logo_w_text.svg`
- `#e1ded8` (white) is an inverted color for previously used `#1e2127` (black)
- Uses SVG on the brand page
- Updates logos in https://github.com/stasadev/ddev/tree/20250306_stasadev_logo_word_dark_mode/docs/content/developers/logos

White on blue looks better (I tried blue on blue, it's not good):

image

https://ddev--7055.org.readthedocs.build/en/7055/developers/brand-guide/

result


docs/content/developers/brand-guide.md
```diff
@@ -17,13 +17,13 @@ If possible, use the [SVG version](https://github.com/ddev/ddev/tree/main/docs/c
 If the SVG format is not supported, you can use the exported PNG versions of the logo. Use [@2x](https://github.com/ddev/ddev/tree/main/docs/content/developers/logos/2x), [@3x](https://github.com/ddev/ddev/tree/main/docs/content/developers/logos/3x), and [@4x](https://github.com/ddev/ddev/tree/main/docs/content/developers/logos/4x) for high
 pixel density displays. Many applications support [@2x](https://github.com/ddev/ddev/tree/main/docs/content/developers/logos/2x) annotations in the image path and automatically choose the correct image for the display in use.
 
-Currently there is no prepared version for dark backgrounds of the word/figurative mark.
+Currently, there is no prepared version for dark backgrounds of the word/figurative mark.
```
stasadev, line 20: I have no idea how to generate matching 1x, 2x, 3x, 4x from SVG.

rfay: I imagine @mattstein would know if you DM him on Discord. May have to friend him first. He's also good with direct email IIRC.

mattstein: These should work, @stasadev, but let me know if I missed anything! This is a ZIP file of PNG images in 1–4x variants, both white and blue: [ddev-mark-square.zip](https://github.com/user-attachments/files/19116876/ddev-mark-square.zip)

stasadev: Thank you, @mattstein.

I checked the generated logos, and they all seem to have the same density.

I also tested what we already have, and the density doubles with each scaled logo. From what I understand, this is the intended behavior to maintain consistent quality.

I see these logos were introduced by @nico-loeber in:

- #3727

I’m not sure what was used to convert these logos to PNG, but we need a documented way to do that.

I did some research and created a script to generate the logos using `inkscape`, with additional optimization using lossless compression via `oxipng`.

Thanks for the effort! The original intention was to generate a logo with white text instead of black.
(I'm not sure if we need to provide the white logos without text, I did it only for the docs logo since the blue on blue looked out of place, but let me know if I should add it.)

stasadev:
Some of the images increased in size because Inkscape's default DPI for 1x is 96.
Previously, images were generated using 72 DPI for 1x, which is why the older images were smaller.
If we decrease the DPI, the images will also reduce in width and height.

tyler36 (approved):
Works well.
DDEV logo looks good on Chrome v134 in light and dark mode.

image

Brand guide: light mode
image
Brand guide: dark mode
image

PR #7057, opened 2025-03-07
build: don't use go 1.24 yet until docker issues resolved, fixes #7051

### LINKED ISSUE #7051: DDEV v1.24.4 probably needs to be built with go 1.23 (docker pull problems)

There are ongoing issues like
* https://github.com/moby/moby/issues/49513
* https://github.com/docker/cli/issues/5898
* https://bbs.archlinux.org/viewtopic.php?id=303943

Where people are having trouble using hub.docker.com (or perhaps all registries) with go-1.24-built applications

We should watch this and when doing that release, if necessary, back down to go 1.23 for the release. Or perhaps we should pin to 1.23 now.

### Describe your solution

x

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

- #7051 

Hopefully this is for a short time, but there have been multiple reports (see #7051) of problems pulling images from go code compiled with go 1.24. We're not having trouble locally, but there's no ready that go v1.23 is not fine.

Use go 1.23

Review the PR build output to verify that it uses go 1.23*

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -60,7 +60,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: '>=1.21'
+          go-version: '<1.24'
 
       - name: Build DDEV executables
         run: make linux_amd64 linux_arm64 darwin_amd64 darwin_arm64 completions mkcert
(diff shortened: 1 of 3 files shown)


stasadev (approved):
Looks good, thank you.

Issue #5763, opened 2024-02-01
Rework DDEV configuration model using Viper


rfay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

DDEV uses go-yaml to read config files, but Viper has much more capability, including overriding with environment variables, much more robust. We could solve lots of things and rework config if we switched, and probably avoid some of the classic go-yaml failures as well.

### Describe your solution

Use viper completely for config.

rfay:
@stasadev pointed out this article of frustration with some features of viper: https://www.linkedin.com/pulse/why-i-moving-away-from-spf13viper-koanf-joe-z-ifcqc/

stasadev:
Viper alternatives https://www.reddit.com/r/golang/comments/s59gws/im_looking_for_a_good_alternativ_to_viper/

rfay:
Related: 
* https://github.com/ddev/ddev/issues/4596

PR #7062, opened 2025-03-10
fix: make `ddev config` work with project name from `config.*.yaml`, fixes #7060

### LINKED ISSUE #7060: Cause ddev config commands to respect project name specified in config.local.yaml

For my projects, the config.yaml file is checked into the repo.  I often have multiple copies of the site and have a config.local.yaml where I override the name.  Some ddev commands get confused and operate on the name specified in the config.yaml rather than the name specified in the config.local.yaml.  I recently learned that one option is to remove the name line from the checked in config.yaml which causes ddev to default to using the directory name as the project name.  While this is an excellent option, perhaps it would be better to make ddev respect the new name.

### Describe your solution

Update ddev commands to use the name specified in the config.local.yaml file rather than to use the name specified in the config.yaml

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #7060

```
mkdir foo && cd foo
ddev config --omit-containers=db
ddev delete -Oy

# remove name
sed -i '1d' .ddev/config.yaml

# set a different name
echo "name: bar" > .ddev/config.local.yaml

ddev start
Starting bar...

ddev stop

ddev config --auto
this project root '/home/stas/code/ddev/foo' already contains a project named 'bar'. You may want to remove the existing project with "ddev stop --unlist bar"

ddev stop --unlist bar

# this adds "name: foo" to .ddev/config.yaml
ddev config --webserver-type apache-fpm

ddev start
Starting bar...
```

So you need to unlist the project before running `ddev config` to make it work.

But it must simply work with the `name` we already have from `.ddev/config.local.yaml`

- Sets app name to the appropriate value, so it won't throw an error.
- Doesn't write the `name` to `.ddev/config.yaml` if you already have it in `.ddev/config.*.yaml`.

```
mkdir foo && cd foo
ddev config --omit-containers=db
ddev delete -Oy

# remove name
sed -i '1d' .ddev/config.yaml

# set a different name
echo "name: bar" > .ddev/config.local.yaml

ddev start
Starting bar...

ddev stop

ddev config --auto
You are reconfiguring the project at /home/stas/code/ddev/foo.
Configuring a 'php' project named 'bar' with docroot '' at '/home/stas/code/ddev/foo'.
```

### DIFF
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -343,12 +343,9 @@ func getConfigApp(_ string) *ddevapp.DdevApp {
 	if err != nil {
 		util.Failed("Could not create a new config: %v", err)
 	}
-	// If the project name is changed to something else in the config overrides, update it here.
-	appWithOverrides, err := ddevapp.NewApp(appRoot, true)
-	if err != nil {
-		util.Failed("Could not create a new config: %v", err)
+	if err = app.RespectConfigNameOverride(); err != nil {
+		util.Failed("Could not check for name override: %v", err)
 	}
-	app.Name = appWithOverrides.Name
 	return app
 }
 
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -343,8 +343,8 @@ func getConfigApp(_ string) *ddevapp.DdevApp {
 	if err != nil {
 		util.Failed("Could not create a new config: %v", err)
 	}
-	if err = app.RespectConfigNameOverride(); err != nil {
-		util.Failed("Could not check for name override: %v", err)
+	if hasConfigNameOverride, newName := app.HasConfigNameOverride(); hasConfigNameOverride {
+		app.Name = newName
 	}
 	return app
 }
--- a/cmd/ddev/cmd/config.go
+++ b/cmd/ddev/cmd/config.go
@@ -343,6 +343,12 @@ func getConfigApp(_ string) *ddevapp.DdevApp {
 	if err != nil {
 		util.Failed("Could not create a new config: %v", err)
 	}
+	// If the project name is changed to something else in the config overrides, update it here.
+	appWithOverrides, err := ddevapp.NewApp(appRoot, true)
+	if err != nil {
+		util.Failed("Could not create a new config: %v", err)
+	}
+	app.Name = appWithOverrides.Name
 	return app
 }
 
--- a/cmd/ddev/cmd/config_test.go
+++ b/cmd/ddev/cmd/config_test.go
@@ -888,3 +888,53 @@ func TestDrupalAppTypeUsage(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, nodeps.AppTypeDrupalLatestStable, app.Type)
 }
+
+// TestConfigRespectNameOverride tests that `ddev config` respects name override from `.ddev/config.*.yaml` files
+func TestConfigRespectNameOverride(t *testing.T) {
+	assert := asrt.New(t)
+
+	projectName := strings.ToLower(t.Name())
+	origDir, _ := os.Getwd()
+	_, _ = exec.RunHostCommand(DdevBin, "stop", "--unlist", projectName)
+
+	// Create a temporary directory and switch to it.
+	tmpDir := testcommon.CreateTmpDir(t.Name())
+	_ = os.Chdir(tmpDir)
+
+	var err error
+
+	t.Cleanup(func() {
+		err = os.Chdir(origDir)
+		assert.NoError(err)
+		out, err := exec.RunHostCommand(DdevBin, "delete", "-Oy", projectName)
+		assert.NoError(err, "output=%s", out)
+		_ = os.RemoveAll(tmpDir)
+	})
+
+	_ = os.Chdir(tmpDir)
+
+	// Create a dummy config file to make it look like it's an existing project
+	// Otherwise DDEV will not read the override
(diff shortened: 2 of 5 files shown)


cmd/ddev/cmd/config.go
```diff
@@ -340,6 +340,12 @@ func getConfigApp(_ string) *ddevapp.DdevApp {
 	if err != nil {
 		util.Failed("Could not create a new config: %v", err)
 	}
+	// If the project name is changed to something else in the config overrides, update it here.
+	appWithOverrides, err := ddevapp.NewApp(appRoot, true)
+	if err != nil {
+		util.Failed("Could not create a new config: %v", err)
+	}
+	app.Name = appWithOverrides.Name
```
rfay, line 348: Why just do the name? Isn't this a general problem with anything that might be in the override file? Like PHP version, etc?

rfay (commented):
Thanks for taking this on!

cmd/ddev/cmd/config.go
stasadev: Because it will bring all overrides that you have in `.ddev/config.local.yaml` to `.ddev/config.yaml`.

If you have something like this in the override:

```yaml
additional_hostnames:
    - foo
    - bar
```

it will be copied to the main config file. Do we really want that?

stasadev: I mean, people use these overrides to keep the main config clean, and if you pull some add-on, run `ddev config --auto` and remove that add-on, the entire add-on config (defined in `config.*.yaml`) will still be in your main config.

rfay: Agreed. This is a difficult problem, and the long-term solution is probably in the other issue. Here are you rewriting the `name` in config.yaml or leaving it alone in the end?

Probably we should be only writing things that originally exist in the config.yaml and aren't overwritten by `config.*.yaml` That's probably difficult. OK with smaller scope, but hate to solve only one possible problem.

We may want to just say to people using `config.*.yaml` "do not use `ddev config`"

rfay: Or maybe `ddev config` should just detect the existence of `config.*.yaml` and refuse, saying "Please edit your files instead of using this"

stasadev: > Here are you rewriting the name in config.yaml or leaving it alone in the end?

`name` is written all the time when you run `ddev config`.

> Or maybe `ddev config` should just detect the existence of `config.*.yaml` and refuse, saying "Please edit your files instead of using this"

I think this will break a lot of automated workflows.

stasadev:
It should be good now.

`.ddev/config.yaml` doesn't add `name:` to `.ddev/config.yaml` if it's defined elsewhere.

BTW, we could omit the name by default, so it would always be set from the enclosing directory.
However, after thinking about it, I decided not to do that, as it could cause too much confusion for our users.

stasadev:
@selwynpolit, I’d like to know your opinion on this PR. Thanks!

rfay:
> BTW, we could omit the name by default, so it would always be set from the enclosing directory.
> However, after thinking about it, I decided not to do that, as it could cause too much confusion for our users.

I wish it were that way from the beginning, but don't think we can change courses at this point, so I agree with you.

stasadev:
We could replace the name with a commented line on `ddev config`:

Before:

```
$ cat .ddev/config.yaml
name: foobar
type: php
...
```

After:

```
# name: foobar
type: php
```

Or:

```
# Using name from enclosing folder
# name: foobar
type: php
```

So it can be clearer this way, but I'm not sure it's clear enough.

rfay:
We definitely don't want to make a change like this in haste, but we could consider longer term a global config that allowed changing the behavior, and then change the default for that even farther out.

rfay (changes requested):
This works fine, but I do wish it were more general. There are all those other things that might be overridden that should be respected just the same. 

I think most likely would be `php-version` and `web_environment` or `timezone`. Really anything that might exist in a `config.*.yaml` generallly or in an add-on. 

I'd also like it a bit better if the equivalent of OmitName were not in the actual app struct, but rather (probably?) a function?

If you think this is best, we can go with it, as it solves the specific request, but it seems to me that the need is more general.

stasadev:
> This works fine, but I do wish it were more general. There are all those other things that might be overridden that should be respected just the same. 

I think this should be considered in:

- #5763

---

The original intention of this PR was to make `ddev config` work (rather than show an error) when you have a `name:` in `.ddev/config.*.yaml`, only this and nothing else.

I think I have overdone it with `OmitName` here, so I'm going to undo my last commit.

rfay (changes requested):
This doesn't behave the way I'd expect. If `config.name.yaml` has a `name` in it, then `ddev config --whatever` updates the name in `config.yaml`. I would have expected it to *not* touch the `name` in the config.yaml.

stasadev:
> This doesn't behave the way I'd expect. If `config.name.yaml` has a `name` in it, then `ddev config --whatever` updates the name in `config.yaml`. I would have expected it to *not* touch the `name` in the config.yaml.

Okay. I refactored it again, this time using `app.ReadConfig` instead of calling `ddevapp.NewApp`, since the latter does a lot more than we need for a simple `app.Name` comparison.

rfay (approved):
I like the way this works now, thank you. Does it need any docs tweaks? I don't know how we can make them in a meaningful way, people will just stumble over the details. 

As always, I worry about adding complex code for specific use-cases, but maybe we can enlarge our overall config vision this year and make workarounds like this less necessary.

PR #7063, opened 2025-03-10
build: put both git and ssh in both webserver images, fixes #7054

### LINKED ISSUE #7054: Production / hardened webserver images no longer include git or ssh
### Preliminary checklist

Not `debug test` but I don't think that's useful in this case anyways!

```
❯ docker run --rm  ddev/ddev-webserver-prod:v1.24.3 which git
/usr/bin/git
❯ docker run --rm  ddev/ddev-webserver-prod:v1.24.3 which ssh
exit status 1
❯ docker run --rm  ddev/ddev-webserver-prod:v1.24.0 which ssh
/usr/bin/ssh
```

### Expected Behavior

`ssh` is available so `git` can work.

### Actual Behavior

```
Cloning into '/tmp/tmp.JkH5ATrbRF'...
error: cannot run ssh: No such file or directory
fatal: unable to fork
Cloning into '/tmp/tmp.JkH5ATrbRF'...
error: cannot run ssh: No such file or directory
fatal: unable to fork
```

### Steps To Reproduce

1. Turn on hardened images (in our case, we're using `ddev config global --use-hardened-images` in CI as it's a faster download).
2. Try to do git clone with an SSH remote. In our case, we need that to be able to pull a repository down from Pantheon and then later push back up to it.

### Anything else?

https://github.com/ddev/ddev/pull/6839#pullrequestreview-2509451718 suggested removing both git and ssh, but in practice git was left in. It seems like either:

1. git should be removed, because it is not that useful without ssh.
2. ssh should be added back in.
3. Even if nothing changes, can we update the release notes? It wasn't clear from me scanning them that the above PR touched the hardened images - I only figured it out by running `git log -S -p ssh` locally.

rfay:
## The Issue

- #7054

Make sure both git and ssh client are in both ddev-webserver and ddev-webserver-prod

```
docker run -it --rm ddev/ddev-webserver:20250307_rfay_add_hardened_ssh git --version
docker run -it --rm ddev/ddev-webserver-prod:20250307_rfay_add_hardened_ssh git --version
docker run -it --rm ddev/ddev-webserver:20250307_rfay_add_hardened_ssh ssh -V
docker run -it --rm ddev/ddev-webserver-prod:20250307_rfay_add_hardened_ssh ssh -V

## Automated Testing Overview

I did add a check for git and ssh in the image. It doesn't specifically check hardened though.

### DIFF
--- a/containers/ddev-webserver/Dockerfile
+++ b/containers/ddev-webserver/Dockerfile
@@ -42,6 +42,7 @@ RUN apt-get install -y postgresql-common && /usr/share/postgresql-common/pgdg/ap
 RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests -y \
     bash-completion \
     gettext \
+    git \
     graphviz \
     less \
     libcap2-bin \
@@ -49,6 +50,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq install -y -o Dpkg::Options::="--
     libmagickcore-6.q16-6-extra \
     locales \
     mariadb-client \
+    openssh-client \
     patch \
     postgresql-client \
     pv \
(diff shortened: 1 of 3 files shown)


stasadev (approved):
Looks good to me.

PR #7061, opened 2025-03-10
docs: Explain corp vpn and proxy, fixup FAQ, fixes #7048

rfay:
## The Issue

- #7048 

We occasionally have people with VPN or proxy situations that need help and reference about those

## How This PR Solves The Issue

* Add VPN and Proxy info
* Add more on DNS problems
* Add `name` info to FAQ

## TODO

- [x] Update [ddev/ddev-proxy-support](https://github.com/ddev/ddev-proxy-support) to show supported and explain the different techniques and point to doc.
- [x] Add info about `name` line to remote-config if it's not already there
- [x] Add a link to @darrenoh article, https://darren.oh.name/node/81
- [x] Verify that `ddev-proxy-support` can work when pulling new images (it probably can't)
- [x] Related discussion for [Global Protect](https://drupal.slack.com/archives/C5TQRQZRR/p1741964281950459?thread_ts=1741961709.100059&cid=C5TQRQZRR) with [fix mentioned](https://drupal.slack.com/archives/C5TQRQZRR/p1741965403742839?thread_ts=1741961709.100059&cid=C5TQRQZRR)
- [x] [Docker Desktop docs](https://docs.docker.com/desktop/troubleshoot-and-support/faqs/macfaqs/#add-custom-ca-certificates-server-side), additional [blog](https://blog.container-solutions.com/adding-self-signed-registry-certs-docker-mac)
- [x] [OrbStack issue](https://github.com/orbstack/orbstack/issues/1835)
- [x] Extensive slack [writeup](https://drupal.slack.com/archives/C5TQRQZRR/p1741975434827539?thread_ts=1741961709.100059&cid=C5TQRQZRR) by @shelane, who should review changes.

## Manual Testing Instructions

Review (mostly) at https://ddev--7061.org.readthedocs.build/en/7061/users/usage/networking/

Network test setup rendered at https://ddev--7061.org.readthedocs.build/en/7061/developers/network-test-environments/

FAQ is at https://ddev--7061.org.readthedocs.build/en/7061/users/usage/faq/#i-want-to-use-the-same-code-with-various-project-names

PR #7064, opened 2025-03-11
build: skip testing with buildkite if diff is not from that branch

stasadev:
## The Issue

I wondered why buildkite tests were run for:

- #7061

Buildkite's skip logic works well when you're always up to date, but not when you're pushing commits without rebasing.

Affected PRs are PRs for documentation.

This PR uses the base merge commit instead of the latest HEAD to compare changes, which helps reduce the carbon footprint.

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -166,9 +166,14 @@ fi
 
 # If this is a PR and the diff doesn't have code, skip it
 set -x
-if [ "${BUILDKITE_PULL_REQUEST:-false}" != "false" ] && ( ! git diff --name-only refs/remotes/origin/${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-} | egrep "^(\.buildkite|Makefile|pkg|cmd|vendor|go\.)" >/dev/null ); then
-  echo "Skipping buildkite build since no code changes found"
-  exit 0
+if [ "${BUILDKITE_PULL_REQUEST:-false}" != "false" ]; then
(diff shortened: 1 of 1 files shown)


rfay:
Pretty sure I'm in favor of this, but I think the reason it ran on that PR was the .spellcheckwordlist. Note that the buildkite tests run always (quickly on docs), they just short-circuit if they find there's no code to test.

stasadev:
I thought about `.spellcheckwordlist.txt` too, but when I checked it locally:

```
$ git diff --name-only refs/remotes/origin/main | egrep "^(\.buildkite|Makefile|pkg|cmd|vendor|go\.)"
cmd/ddev/cmd/debug-download-images.go
```

rfay:
I have absolutely seen a number of times that the buildkite tests did a full run (often just on one runner) for no reason, as in the PR under study. I blamed it on "who knows". I guess it can't be the fault of buildkite. (Last I checked, buildkite did not have a capability to prevent runs like github does; I corresponded with them about it. Time has passed though.)

So the problem here is when you push a second commit to the PR?

stasadev:
> So the problem here is when you push a second commit to the PR?

Yes, but only when the base branch contains new changes in the related locations.

rfay:
So the problem is when a second or nth commit is added, and the upstream has a new commit that changes the affected items? (assuming that the initial commit was against current)?

stasadev:
That's correct.

rfay (approved):
Let's go with it, thanks!

PR #7070, opened 2025-03-12
build: Use specific binfmt for qemu in push-tagged-image [skip ci]

rfay:
## The Issue

Regular failures of push-tagged-image due to "Cannot allocate memory" in arm64 apt-get install

Use specific binfmt workaround until this gets fixed upstream.

### DIFF
--- a/.github/workflows/push-tagged-image.yml
+++ b/.github/workflows/push-tagged-image.yml
@@ -39,7 +39,7 @@ permissions:
 jobs:
   push-tagged-image:
     name: "push tagged image"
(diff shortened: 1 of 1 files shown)


rfay:
Push attempt using this in https://github.com/rfay/ddev/actions/runs/13815961807/job/38649087271

stasadev (changes requested):
Let's do the same for db-image. I had failures there while building v1.24.3.

rfay:
Push worked

PR #7071, opened 2025-03-12
feat: support prepend.Dockerfile* files for multi-stage builds

hanoii:
## The issue

We can customize the default Dockerfile by adding appending stuff to the end (through `Dockerfile* variants`), and by injecting something before ddev own customizations, but still in between ddev's own Dockerfile (through `pre.Dockerfile*` variants). But we cannot add anything before everything else, limiting other customization like being able to define other build stages. 

It adds a new `prepend.Dockerfile*` variant that gets added on top of everything before anything added by ddev.

1. create a `.ddev/web-build/prepend.Dockerfile.pr-7071` with:

```dockerfile
# If we want to use any of the build time environment variables injected by ddev 
# on the prepend.Dockerfile* variants we need to manually declare them to make 
# them available using the ARG instruction.
ARG BASE_IMAGE
FROM $BASE_IMAGE AS build-stage-go

ARG uid
ARG gid

# install go
RUN set -eux; \
    GO_VERSION=$(curl -fsSL "https://go.dev/dl/?mode=json" | jq -r ".[0].version"); \
    AARCH=$(dpkg --print-architecture); \
    wget -q https://go.dev/dl/${GO_VERSION}.linux-${AARCH}.tar.gz -O go.tar.gz; \
    tar -C /usr/local -xzf go.tar.gz; \
    rm go.tar.gz;
```

2. Create a `.ddev/web-build/Dockerfile.pr-7071` with:

```dockerfile
# Copy entire go directory from the build stage defined above.
COPY --from=build-stage-go /usr/local/go /usr/local
```

3. Run `ddev start` and it should build without error.
4. Check that both files are reported as custom web-build configurations:

```
Using custom web-build configuration: [
	/tmp/gotest/.ddev/web-build/prepend.Dockerfile.pr-7071
	/tmp/gotest/.ddev/web-build/Dockerfile.pr-7071
] 
```

5. Run `ddev exec go version` and verify the command runs ok and you got the latest go version. 

## Automated Testing Overview

This verifies that that the all of the variants work by doing a simple build stage that creates a file and then it copies it over. It tests both `prepend.Dockerfile` and `prepend.Dockerfile.*`

## Release/Deployment Notes

It affects nothing.

## Additional context

Most important docs for review: https://ddev--7071.org.readthedocs.build/en/7071/users/extend/customizing-images/#adding-extra-dockerfiles-for-webimage-and-dbimage

This is a follow up for a discord thread: https://discord.com/channels/664580571770388500/1349036904015855636

The reasoning behind this is having a way to, if needed/wanted for any reason, be able to feed to ddev Dockefile build stages. This can help in splitting out big modifications of Dockerfiles into their own stages.

The change to ddev code is super simple actually, just adds any `stage.Dockerfile.*` that's get added before EVERYTHING in the current Dockerfile, allowing you to do whatever you want there.

While I am still working through a test project that uses this successfully, initial test works just fine and I think you can gather what we can do with this.

i.e.

```
#ddev-generated - Do not modify this file; your modifications will be overwritten.


### From user Dockerfile /Users/ariel/Sites/work/dockerfile-stage/.ddev/web-build/stage.Dockerfile.pimp-my-shell.ahoy:
#ddev-generated

ARG BASE_IMAGE="scratch"
FROM $BASE_IMAGE AS pimp-my-shell-ahoy

# ahoy
RUN set -eux; \
    AHOY_VERSION=2.2.0; \
    AARCH=$(dpkg --print-architecture); \
    wget -q https://github.com/ahoy-cli/ahoy/releases/download/v${AHOY_VERSION}/ahoy-bin-linux-${AARCH} -O - > /usr/local/bin/ahoy; \
    chmod +x /usr/local/bin/ahoy;


### From user Dockerfile /Users/ariel/Sites/work/dockerfile-stage/.ddev/web-build/stage.Dockerfile.pimpi-my-shell.kitty:
#ddev-generated

ARG BASE_IMAGE="scratch"
FROM $BASE_IMAGE AS pimp-my-shell-kitty

# kitty-terminfo from directoy from their GitHub repo @ master
RUN set -eux; \
    KITTY_VERSION=0.40.0; \
    url="https://github.com/kovidgoyal/kitty/raw/refs/tags/v${KITTY_VERSION}/terminfo/x/xterm-kitty"; \
    mkdir -p /usr/share/terminfo/x; \
    cd /usr/share/terminfo/x; \
    wget -q "${url}";
```

This is the first part of the `.webImageBuild`'s Dockerfile where it concatenates `stage.Dockerfile.pimpi-my-shell.ahoy` and `stage.Dockerfile.pimpi-my-shell.kitty`.

Further below I also have a `Dockerfile.pimp-my-shell` that does:

```
COPY --from=pimp-my-shell-ahoy /usr/local/bin/ahoy /usr/local/bin/ahoy
COPY --from=pimp-my-shell-kitty /usr/share/terminfo/x/xterm-kitty /usr/share/terminfo/x/xterm-kitty
```

Changing something in the ahoy stage, doesn't bust the kitty stage cache. The COPY entries gets busted in order (the first one that get busted busts the others, but the COPY performance is negible. 

> [!NOTE]
> At first I was trying to support more logic, like adding all of the ARGs or auto detect FROM statements to see if I should add the FROM base image but then I figured I was getting too complex into something that's inherently complex and if one would want to use this, he should cover anything what's required in this part of the dockerfile.

Docs: https://ddev--7071.org.readthedocs.build/en/7071/users/extend/customizing-images/#adding-extra-dockerfiles-for-webimage-and-dbimage

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -106,12 +106,12 @@ For certain use cases, you might need to add directives very early on the Docker
 * `.ddev/db-build/pre.Dockerfile.*`
 * `.ddev/db-build/pre.Dockerfile`
 
-Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), you can use `stage.` variants that are inserted *before* everything else, *on top* of the generated Dockerfile.
+Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/) and other more complex use cases, you can use `prepend.` variants that are inserted *before* everything else, *on top* of the generated Dockerfile.
 
-* `.ddev/web-build/stage.Dockerfile.*`
-* `.ddev/web-build/stage.Dockerfile`
-* `.ddev/db-build/stage.Dockerfile.*`
-* `.ddev/db-build/stage.Dockerfile`
+* `.ddev/web-build/prepend.Dockerfile.*`
+* `.ddev/web-build/prepend.Dockerfile`
+* `.ddev/db-build/prepend.Dockerfile.*`
+* `.ddev/db-build/prepend.Dockerfile`
 
 Multi-stage builds are useful to anyone who has struggled to optimize Dockerfiles while keeping them easy to read and maintain.
 
@@ -151,14 +151,14 @@ RUN chmod -R ugo+rw $COMPOSER_HOME
 ENV COMPOSER_HOME=""
 ```
 
-An example [Multi-stage](https://docs.docker.com/build/building/multi-stage/) web image could have a  `.ddev/web-build/stage.Dockerfile`:
+An example [Multi-stage](https://docs.docker.com/build/building/multi-stage/) web image could have a  `.ddev/web-build/prepend.Dockerfile`:
 
 ```dockerfile
-# We must declare the build enviroment variables we want to use on stage.Dockerfile
+# We must declare the build enviroment variables we want to use on prepend.Dockerfile
 ARG BASE_IMAGE="scratch"
 FROM $BASE_IMAGE AS build-stage-go
 
-# We must declare the build enviroment variables we want to use on stage.Dockerfile
+# We must declare the build enviroment variables we want to use on prepend.Dockerfile
 ARG uid
 ARG gid
 
@@ -193,8 +193,8 @@ The following environment variables are available for the web Dockerfile to use 
 * `$TARGETOS`: The build target operating system (always `linux`)
 * `$TARGETPLATFORM`: `linux/amd64` or `linux/arm64` depending on the machine it's been executed on
 
-!!!warning "These variables won't be automatically available on `stage.Dockerfile*` variants"
-    If you need to use any of these variables you will need to manually add them to your `stage.Dockerfile*` files using [ARG](https://docs.docker.com/reference/dockerfile/#arg) instructions.
+!!!warning "These variables won't be automatically available on `prepend.Dockerfile*` variants"
+    If you need to use any of these variables you will need to manually add them to your `prepend.Dockerfile*` files using [ARG](https://docs.docker.com/reference/dockerfile/#arg) instructions.
 
 For example, a Dockerfile might want to build an extension for the configured PHP version like this using `$DDEV_PHP_VERSION` to specify the proper version:
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -99,12 +99,19 @@ For more complex requirements, you can add:
 
 These files’ content will be inserted into the constructed Dockerfile for each image. They are inserted *after* most of the rest of the things that are done to build the image, and are done in alphabetical order, so `Dockerfile` is inserted first, followed by `Dockerfile.*` in alphabetical order.
 
-For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* everything else:
+For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* what DDEV adds to build the image:
 
 * `.ddev/web-build/pre.Dockerfile.*`
 * `.ddev/web-build/pre.Dockerfile`
 * `.ddev/db-build/pre.Dockerfile.*`
 * `.ddev/db-build/pre.Dockerfile`
+
+Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), you can use `stage.` variants for this that are inserted *before* everything else, *on top* of the generated Dockerfile.
+
+* `.ddev/web-build/stage.Dockerfile.*`
+* `.ddev/web-build/stage.Dockerfile`
+* `.ddev/db-build/stage.Dockerfile.*`
+* `.ddev/db-build/stage.Dockerfile`
 
 Examine the resultant generated Dockerfile (which you will never edit directly), at `.ddev/.webimageBuild/Dockerfile`. You can force a rebuild with [`ddev debug rebuild`](../usage/commands.md#debug-rebuild). `ddev debug rebuild` is also great because it shows you the entire process of the build for debugging.
 
@@ -156,6 +163,9 @@ The following environment variables are available for the web Dockerfile to use 
 * `$TARGETARCH`: The build target architecture, like `arm64` or `amd64`
 * `$TARGETOS`: The build target operating system (always `linux`)
 * `$TARGETPLATFORM`: `linux/amd64` or `linux/arm64` depending on the machine it's been executed on
+
+!!!warning "These variables won't be automatically available on `stage.Dockerfile*` variants"
+    If you need to use any of these variables you will need to manually add them to your `stage.Dockerfile*` files using [ARG](https://docs.docker.com/reference/dockerfile/#arg) instructions.
 
 For example, a Dockerfile might want to build an extension for the configured PHP version like this using `$DDEV_PHP_VERSION` to specify the proper version:
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -113,6 +113,8 @@ Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/
 * `.ddev/db-build/stage.Dockerfile.*`
 * `.ddev/db-build/stage.Dockerfile`
 
+Multi-stage builds are useful to anyone who has struggled to optimize Dockerfiles while keeping them easy to read and maintain.
+
 Examine the resultant generated Dockerfile (which you will never edit directly), at `.ddev/.webimageBuild/Dockerfile`. You can force a rebuild with [`ddev debug rebuild`](../usage/commands.md#debug-rebuild). `ddev debug rebuild` is also great because it shows you the entire process of the build for debugging.
 
 Examples of possible Dockerfiles are `.ddev/web-build/Dockerfile.example` and `.ddev/db-build/Dockerfile.example`, created in your project when you run [`ddev config`](../usage/commands.md#config).
@@ -148,6 +150,34 @@ RUN chmod -R ugo+rw $COMPOSER_HOME
 # Now turn it off, because ordinary users will want to be using the default.
 ENV COMPOSER_HOME=""
 ```
+
+An example [Multi-stage](https://docs.docker.com/build/building/multi-stage/) web image could have a  `.ddev/web-build/stage.Dockerfile`:
+
+```dockerfile
+# We must declare the build enviroment variables we want to use on stage.Dockerfile
+ARG BASE_IMAGE="scratch"
+FROM $BASE_IMAGE AS build-stage-go
+
+# We must declare the build enviroment variables we want to use on stage.Dockerfile
+ARG uid
+ARG gid
+
+# install go
+RUN set -eux; \
+    GO_VERSION=1.24.1; \
+    AARCH=$(dpkg --print-architecture); \
+    wget -q https://go.dev/dl/go${GO_VERSION}.linux-${AARCH}.tar.gz -O go.tar.gz; \
+    tar -C /usr/local -xzf go.tar.gz; \
+    rm go.tar.gz;
+```
+
+And then a `Dockerfile`:
+
+```dockerfile
+# Copy entire go directory from the build stage defined above.
+COPY --from=build-stage-go /usr/local/go /usr/local
+```
+
 
 **Remember that the Dockerfile is building a Docker image that will be used later with DDEV.** At the time the Dockerfile is executing, your code is not mounted and the container is not running, the image is being built. So for example, an `npm install` in `/var/www/html` will not do anything to your project because the code is not there at image building time.
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -106,7 +106,7 @@ For certain use cases, you might need to add directives very early on the Docker
 * `.ddev/db-build/pre.Dockerfile.*`
 * `.ddev/db-build/pre.Dockerfile`
 
-Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), you can use `stage.` variants for this that are inserted *before* everything else, *on top* of the generated Dockerfile.
+Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), you can use `stage.` variants that are inserted *before* everything else, *on top* of the generated Dockerfile.
 
 * `.ddev/web-build/stage.Dockerfile.*`
 * `.ddev/web-build/stage.Dockerfile`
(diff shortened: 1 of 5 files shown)


hanoii:
It's obvious that although I am using the base image as provided by ddev, I could have use any other image on each stage.

hanoii:
And I probably still need to add some notes to the README file, the docs and tests, but before continuing I just wanted your initial "approval" on adding this.

hanoii:
Added tests and README.txt tweaks

hanoii:
Added docs, this should be feature complete and ready to review.

pkg/ddevapp/config.go
```diff
@@ -1206,7 +1209,29 @@ func WriteBuildDockerfile(app *DdevApp, fullpath string, userDockerfilePath stri
 	// Normal starting content is the arg and base image
 	contents := `
 #ddev-generated - Do not modify this file; your modifications will be overwritten.
+`
+	// If there are user stage.Dockerfile.* files, insert their contents
+	if userDockerfilePath != "" {
+		files, err := filepath.Glob(filepath.Join(userDockerfilePath, "stage.Dockerfile.*"))
```
stasadev, line 1215: Docs say:

* `.ddev/web-build/stage.Dockerfile.*`
* `.ddev/web-build/stage.Dockerfile`
* `.ddev/db-build/stage.Dockerfile.*`
* `.ddev/db-build/stage.Dockerfile`

But this code doesn't support `stage.Dockerfile` variant without `.*`

stasadev (changes requested):

pkg/ddevapp/config.go
hanoii: nice catch! Fixing and will add test for this and for pre.Dockerfile which was not tested (the pre.Dockerfile - no suffix - variant)

docs/content/users/extend/customizing-images.md
```diff
@@ -99,13 +99,20 @@ For more complex requirements, you can add:
...
 
-For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* everything else:
+For certain use cases, you might need to add directives very early on the Dockerfile like proxy settings or SSL termination. You can use `pre.` variants for this that are inserted *before* what DDEV adds to build the image:
 
 * `.ddev/web-build/pre.Dockerfile.*`
 * `.ddev/web-build/pre.Dockerfile`
 * `.ddev/db-build/pre.Dockerfile.*`
 * `.ddev/db-build/pre.Dockerfile`
 
+Finally, to support [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), you can use `stage.` variants that are inserted *before* everything else, *on top* of the generated Dockerfile.
```
stasadev, line 109: Can we also give a small example of how to use it right here inside DDEV?

I think newbies won't understand what they can do, even after reading Docker docs, e.g add a simple thing inside the container and a note about the presence of a second `Dockerfile` with `COPY --FROM ,,,`.

Thanks.

pkg/ddevapp/dotddev_assets/web-build/README.txt
```diff
@@ -4,11 +4,17 @@ Files in this directory will be used to customize the webimage, you can add:
 * .ddev/web-build/Dockerfile
 * .ddev/web-build/Dockerfile.*
 
-Additionally, you can use `pre.` variants that are inserted before everything else:
+Additionally, you can use `pre.` variants that are inserted before what DDEV adds :
```
stasadev, line 7:
```suggestion
Additionally, you can use `pre.` variants that are inserted before what DDEV adds:
```

pkg/ddevapp/dotddev_assets/db-build/README.txt
```diff
@@ -4,11 +4,17 @@ Files in this directory will be used to customize the dbimage, you can add:
 * .ddev/db-build/Dockerfile
 * .ddev/db-build/Dockerfile.*
 
-Additionally, you can use `pre.` variants that are inserted before everything else:
+Additionally, you can use `pre.` variants that are inserted before what DDEV adds :
```
stasadev, line 7:
```suggestion
Additionally, you can use `pre.` variants that are inserted before what DDEV adds:
```

docs/content/users/extend/customizing-images.md
hanoii: @stasadev done!

docs/content/users/extend/customizing-images.md
```diff
@@ -142,6 +151,33 @@ RUN chmod -R ugo+rw $COMPOSER_HOME
...
+
+# We must declare the build enviroment variables we want to use on stage.Dockerfile
+ARG uid
+ARG gid
+
+# install go
+RUN set -eux; \
+    GO_VERSION=1.24.1; \
+    AARCH=$(dpkg --print-architecture); \
+    wget -q https://go.dev/dl/go${GO_VERSION}.linux-${AARCH}.tar.gz -O go.tar.gz; \
```
stasadev, line 169: Automate to install the latest `go`:

```suggestion
    GO_VERSION=$(curl -fsSL "https://go.dev/dl/?mode=json" | jq -r ".[0].version"); \
    AARCH=$(dpkg --print-architecture); \
    wget -q https://go.dev/dl/${GO_VERSION}.linux-${AARCH}.tar.gz -O go.tar.gz; \
```

stasadev (approved):
Works well, thank you!

I only have a suggestion for your `go` example.

docs/content/users/extend/customizing-images.md
hanoii: @stasadev sure, I copied this over from my dockerfile, I used hard-coded versions so I can update them and make sure all my local layers are gonna be cached and to make it specific, but for the example, that's better.

hanoii: @stasadev I am doing something similar to get the latest version here https://github.com/hanoii/ddev-pimp-my-shell/blob/feb0b4f693573ec1338f1be0063c0b51f6625fd9/update.sh#L59

hanoii:
shameless bump of this PR to see if it can be considered for next release :)

rfay:
I hope to get to it, and you have Stas's support, so it's likely. But from community point of view it's not as important as some other things.

rfay:
We'll try to get this in early in upcoming release cycle.

rfay (commented):
Your contributions are so very much appreciated in so many places.

But I'm sorry, I'm not very happy with this. I spent some time with it today and it seems to me like it adds complexity to a topic that is already too complex for most people, and primarily does it to support a single user's workflow. We'll leave it open and I'll return to it later.

rfay:
Could you please restate here exactly what your situation is? We were thinking we should try to solve it in a more generic way, instead of adding a specific feature.

hanoii:
I just wanted to be able to do a multi stage dockerfile build successfully,
being able to add stages before ddevs own image and without having to
maintain my own image

Reply might be shorter than usual / Respuesta más resumida que de costumbre

hanoii:
I think it's also in the discord thread linked in the PR description, but here it is:

- I have https://github.com/hanoii/ddev-pimp-my-shell/blob/main/web-build/Dockerfile.pimp-my-shell
- It's a longish Dockerfile, it has a lot of different things that gets installed
- I lock each one to a specific version (One could argue that it's bet to always install the the latest, I preferred this so I can bust cache easily on all projects and that I keep track of the new versions of tools). This gets auto-updated by a recent Github Action.
- These dependencies changes often and between that, ddev updates and me switching through a lot of different projects with different php versions I rebuild the images for for each project a lot of times. It's gotten a bit annoying for my taste so I went to look if I can improve the building time of my add-on.
- If the dependency that gets updated is very early of the dockerfile all of the other entries need to run, and even more with ddev updates.
- I figured my best choice was to be able to do it with build stages, so that the docker cache can be used more smartly. For a lot of this I can start with a simple images, download the dependency/install it/dowahtever and then simply copy over the file. If one gets updated, only that "build" stage needs to be rebuild, so now this becomes independent of other dependencies or even ddev images itself. I'd have a lot more options.

rfay:
Why is it important for it to change all the time?

hanoii:
> Why is it important for it to change all the time?

I guess it's not. I don't have an answer other than keeping it up to date

hanoii:
It only builds when something change. There are many tools. There are some
new release very so often.

Reply might be shorter than usual / Respuesta más resumida que de costumbre

hanoii:
Maybe `prepend.` is a better name than `stage.`, but you already have `pre.`

hanoii:
> Can you say what it is that gets out of date?

Looking through the CI commits, you can see what things changes:

https://github.com/hanoii/ddev-pimp-my-shell/commits/main/?author=github-actions%5Bbot%5D

rfay:
It does look like you're arbitrarily changing the install date, forcing change, https://github.com/hanoii/ddev-pimp-my-shell/commit/d1ffb5636e9e7fddab719245bbf3bfccd58b9e28#diff-93d7b8da8161a773eb584ec7e36de7ffbd894cdf860e3beb6649d8b7acf93020L4-L6

hanoii:
That date has nothing to do with the add-on on other projects, that's only because I test the add-on with ddev. So when an update happens, I do a `ddev add-on get .`  so that the internal ddev project of the add-on also has the update to the add-on itself. The update modifies three files:

- `web-build/Dockerfile.pimp-my-shell` (the add-on Dockerfile)
- `.ddev/web-build/Dockerfile.pimp-my-shell` and `.ddev/addon-metadata/pimp-my-shell/manifest.yaml` which are only the self-enclosed add-on on the .ddev project. These files have nothing to do with doing a `ddev add-on get https://github.com/hanoii/ddev-pimp-my-shell/tarball/main` on other projects.

What triggers a rebuild on the projects using the add-on is only the change on the Dockerfile which bust all Dockefile commands after that change. Even those Dockerfiles that comes after that are unrelated to this add-on in the usual lexicographic order.

hanoii:
It really wouldn't make any difference. I don't update the add-on on the projects that often, but often enough that it started to annoy me, and like I said, there are things that are unrelated to my add-on that also triggers a rebuild:

- ddev updates
- any setting to to `config.yaml` that changes the Dockerfile
- adding a new add-on that adds something before my add-on's Dockerfile.
- and yes, updating the add-on, which normally happens when I get to work on a project after some time or sometimes I happen to need something on a project which I feel is going to be nice to have them [available on all of my projects](https://github.com/hanoii/ddev-pimp-my-shell/commit/3891fdb897b254019c95f9c6fc06347f65b63d73).

I can avoid a lot of rebuilding of Docker layers by doing this multi-stage build, speeding up my builds a good amount.

I appreciate your willingness to help, on this and on everything else, but I just don't think there are ways around this. It's one of those things that either you choose to accept it or not. I will though attempt a final push into acceptance 😅 :

The goal of this really is just a way that allows me (or any other although I can see for now it's just me) to prepend something on top of the generated Dockerfile. For now I can see this being used for multi-stage builds, but maybe it can serve other purposes (I can't think of anything else right now). It's also gives you complete control on the Dockerfile, we have the normal additions to the end, the `pre.Dockerfile` that goes in between some of the ddev changes and now I am asking to have this `prepend.` addition that takes it one step further. I think `prepend.` is much better than `stage.` The complexity it adds is only because Dockerfiles are complex enough for most people, but it really would be something that most people won't even know it's there, so I see no major harm in accepting it. It changes nothing if not used.

hanoii:
Just also noting (and cross reference) that #3999  was added to support certain use cases and then I found https://github.com/ddev/ddev/pull/6809#issuecomment-2520102807 from @stasadev which is exactly this

hanoii:
> Now I remember that use-case. I'm inclined to add this, but I think we need more validation, i.e. FROM should be 100% here, otherwise it won't work, right?

You don't do any validation to any of the Dockerfiles so I really don't think you'd need validations here, I do think that's adding more complexity for a logic that depends on this. It you do something unsupported, the build will fail.

And not sure on that honestly, You might add an ARG or ENV or something. I wouldn't do that.

I let you guys pick the best naming if you decide to include this, I'll change it to whatever you decide:

I like (for now):

- `prepend.`
- `init.`

Prepend is very clear what it does.

stasadev:
> I really don't think you'd need validations here

I checked it, and I agree, having Docker validation is enough.

> Prepend is very clear what it does.

Let's use it then, but one more thought:

What if you pre-build your add-on `Dockerfile` to image on DockerHub or ghcr.io (all automated with GitHub Actions), and then use it like this with usual `.ddev/web-build/Dockerfile`:

```
COPY --from=ghcr.io/hanoii/pimp-my-shell /usr/local/bin/ahoy /usr/local/bin/ahoy
COPY --from=ghcr.io/hanoii/pimp-my-shell /usr/share/terminfo/x/xterm-kitty /usr/share/terminfo/x/xterm-kitty
```

hanoii:
Hmmm, I guess that could work, I hace never maintained an image so not fully grasp the implications. I think though that in this case, updates on the image wouldn't land on the project unless I do a docker pull, but correct me if I am wrong. I just wanted to avoid something that can be perfectly sorted out locally.

hanoii:
About to rename to prepend and rebase

stasadev:
> updates on the image wouldn't land on the project unless I do a `docker pull`.

It wouldn't land until you run `ddev debug rebuild`, or a new DDEV release comes out 😅

So it will work the same way it currently works with Composer: you start a project, it pulls out the latest Composer and that's it, Composer won't update until you rebuild the project or use a newer version of DDEV.

hanoii:
Rename done ✅

hanoii:
I also modified the order of the output dockerfiles when you list the custom configs

Now it is:

- Dockerfile*
- pre.Dockerfile*

With this PR:

- prepend.Docerfile*
- pre.Dockerfile*
- Dockerfile* 

Tiny change but much more logical (to me at least).

hanoii:
I am retrying the wsl2 buildkite, not sure if I should, it failed twice

rfay:
tb-wsl-14 is broken right now

rfay:
Rebased for convenience and just to run the wsl2/docker-ce tests on tb-wsl-14, which was failing before.

rfay:
OK, I'm willing to go with it. Please use the PR template to provide full information about this PR, including adding explicit (simple) manual testing instructions. Your narrative description is fine justifying the whole thing, but please put it as an addition to the template. After that gets done, Stas and I will do another round of review. It does look fine to me, although I don't know how people will really understand the difference between pre and prepend.

hanoii:
Done and pushed a minor docs adjustment to match the PR testing instructions ,which I found it clearer.

rfay:
You'll be happy to know that heredocs are fully supported these days (simpler examples than you have). I'm trying to move to them when I see opportunities. https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/

Just a comment, no need to change anything.

rfay (approved):
OK! I hope you like it and others like it!

It does add complexity, and will certainly mean some confusion between `pre` and `prepend`, but it also adds power, and hopefully power that helps your situation. Thanks for the contribution, and for scratching your own itch!

The good thing is the docs read well and prepend does come way down there after people have stopped reading.

stasadev (approved):
Looks good and works as expected. Thank you!

rfay:
Thanks for your contribution and your patience on this!

stasadev:
@hanoii, thank you very much for this contribution!

I reused this technique in:

- https://github.com/ddev/ddev-frankenphp/pull/39

Multi-stage build was required there, because variable expansion is not supported in `COPY --from=`, but the workaround is to use `FROM image:tag-${expanded_variable} AS build`, and then `COPY --from=build`.

PR #7077, opened 2025-03-13
fix: make sure mutagen agents file is fresh on upgrade, add shasum testing, test shasum for mutagen+docker-compose

rfay:
## The Issue

Discord: https://discord.com/channels/664580571770388500/1349663367178096650/1349755537377722520

A user upgraded to DDEV v1.24.3 and somehow got a bad version of the mutagen-agents.tar.gz which did not contain the linux agents.

```
rfay@rfay-mba-m4:~/Downloads/bin.bak$ tar -vtzf mutagen-agents.tar.gz
-rwxr-xr-x  0 0      0     9511958 Feb 24 15:34 aix_ppc64
-rwxr-xr-x  0 0      0     8176880 Feb 24 15:34 darwin_amd64
-rwxr-xr-x  0 0      0     7892416 Feb 24 15:34 darwin_arm64
-rwxr-xr-x  0 0      0     7848088 Feb 24 15:35 dragonfly_amd64
-rwxr-xr-x  0 0      0     7528610 Feb 24 15:35 freebsd_386
```

They did say they might have had internet troubles at the time, but I don't know how we could be untarring an agents file and getting different results.

## TODO

- [x] Check SHA256SUMS or checksums.txt on download
- [x] Check checksums.txt on docker-compose download as well
- [x] util.DownloadFile() now takes an optional arg for SHASUM file
- [x] If shasum validation fails, downloaded file is deleted.
- [x] Add TestRemoveFilesMatchingGlob
- [x] Add TestDownloadFile
- [x] Simplify docker-compose download logic (no longer needs to look for v1, etc.)

This probably doesn't actually solve the issue, but at the very least we should make sure that we can't have an old or bad mutagen-agents.tar.gz file (or docker-compose, etc.)

- [ ] With mutagen enabled, `ddev poweroff` and remove the ~/.ddev/bin/mutagen* , then `ddev start`.
- [ ] With mutagen enabled, `ddev poweroff` and untar a previous mutagen version into ~/.ddev/bin - you should get the new version successfully
- [ ] For extra credit, step through and verify the deletion.

## Automated Testing Overview

- [x] Added fileutil.TestRemoveFilesMatchingGlob()
- [x] TestDownloadFile

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -492,9 +492,6 @@ func DownloadAndExtractTarball(url string, removeTopLevel bool) (string, func(),
 		return "", nil, fmt.Errorf("unable to download %v: %v", url, err)
 	}
 	extractedDir, cleanup, err := ExtractTarballWithCleanup(tarball, removeTopLevel)
-	if err != nil {
-		_ = os.RemoveAll(extractedDir)
-	}
 	return extractedDir, cleanup, err
 }
 
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -487,7 +487,7 @@ func DownloadAndExtractTarball(url string, removeTopLevel bool) (string, func(),
 		_ = os.Remove(tarball)
 	}()
 
-	err = util.DownloadFile(tarball, url, true)
+	err = util.DownloadFile(tarball, url, true, "")
 	if err != nil {
 		return "", nil, fmt.Errorf("unable to download %v: %v", url, err)
 	}
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -502,12 +502,13 @@ func DownloadAndExtractTarball(url string, removeTopLevel bool) (string, func(),
 func ExtractTarballWithCleanup(tarball string, removeTopLevel bool) (string, func(), error) {
 	tmpDir, err := os.MkdirTemp("", fmt.Sprintf("ddev_%s_*", filepath.Base(tarball)))
 	if err != nil {
-		return "", nil, fmt.Errorf("unable to create temp dir: %v", err)
+		return "", func() {}, fmt.Errorf("unable to create temp dir: %v", err)
 	}
+	cleanupFunc := func() { _ = os.RemoveAll(tmpDir) }
 
 	err = Untar(tarball, tmpDir, "")
 	if err != nil {
-		return "", nil, fmt.Errorf("unable to untar %v: %v", tmpDir, err)
+		return "", cleanupFunc, fmt.Errorf("unable to untar %v: %v", tmpDir, err)
 	}
 
 	// If removeTopLevel then the guts of the tarball are the first level directory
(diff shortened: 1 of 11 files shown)


stasadev:
Copied my message from Discord:

I was able to reproduce it, but not within DDEV.

I went to `~/.ddev/bin` and removed the mutagen-related files. Then I ran `ddev start` and pressed `ctrl+c` halfway through, so I got a broken `mutagen.tgz` file.
```
$ ddev start 
Starting d11... 
mutagen.tgz 44.39 MiB / 97.44 MiB [================>-------------------]  45.55%
^C
```
When I tried to extract it using tar, I got two files and `mutagen-agents.tar.gz` was broken too.
```
$ tar -xvzf mutagen.tgz 
mutagen
mutagen-agents.tar.gz

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive

$ tar -xvzf mutagen-agents.tar.gz
aix_ppc64
darwin_amd64
darwin_arm64
dragonfly_amd64
freebsd_386
freebsd_amd64
freebsd_arm
freebsd_arm64
linux_386
linux_amd64
linux_arm
linux_arm64
linux_mips
linux_mipsle
linux_mips64
linux_mips64le
linux_ppc64
linux_ppc64le
linux_riscv64
linux_s390x

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
```
But when I extracted `mutagen.tgz` using Ark (https://apps.kde.org/ark/ which complained about corrupted file), I got a perfectly fine `mutagen-agents.tar.gz` file (with the same size as normal file), which I extracted later with `tar` just fine without any errors (13 agents out of 33 were missing.)

So, the corruption happened during the `mutagen.tgz` download step.

I tried to reproduce using this:
```diff
-reader = bar.NewProxyReader(resp.Body)
+reader = bar.NewProxyReader(io.LimitReader(resp.Body, 1024*1024*60))
```
But the `mutagen-agents.tar.gz` has become corrupted, so I don't think there is any error in the extract logic.

I believe that some parts of `mutagen.tgz` were lost during downloading (somewhere in between), resulting in a working archive.

Checking shasums is the right way.

pkg/archive/archive.go
```diff
@@ -487,11 +487,14 @@ func DownloadAndExtractTarball(url string, removeTopLevel bool) (string, func(),
...
 	}()
 
-	err = util.DownloadFile(tarball, url, true)
+	err = util.DownloadFile(tarball, url, true, "")
 	if err != nil {
 		return "", nil, fmt.Errorf("unable to download %v: %v", url, err)
 	}
 	extractedDir, cleanup, err := ExtractTarballWithCleanup(tarball, removeTopLevel)
+	if err != nil {
+		_ = os.RemoveAll(extractedDir)
```
rfay, line 496: I worry that this could be questionable if the extractedDir were assumed to already exist and have files in it.

stasadev (approved):
It works well and I was able to see the error while doing manual editing of the shasum code.

rfay:
Applied your changes, thanks!. You would have been welcome to push them in.

stasadev (approved):
Thanks, looks good to me.

PR #7078, opened 2025-03-14
build: Update go mod files except docker, replaces #7074

rfay:
## The Issue

- #7074 wanted us to update x/net, but it's probably time to update go packages

This updates everything except the docker packages, which have a number of deprecations and will need a PR of their own.

(docker/docker v27.3.1+incompatible does not pass golangci-lint due to deprecations when upgraded to 28.x)

```
go get -u ./...
go mod vendor
go mod tidy
# Roll back docker/docker back to v27.3.1
go mod vendor
go mod tidy
```

It's probably good if it passes tests


stasadev (approved):
Looks good to me.

PR #7080, opened 2025-03-14
docs: add Wordpress special handling info about wp-cli.yml

nickchomey:
## The Issue

- #7039 

Adds a doc for this, as discussed in #7041

Review rendered at https://ddev--7080.org.readthedocs.build/en/7080/users/usage/cms-settings/#cms-specific-help-and-techniques

### DIFF
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -146,7 +146,7 @@ See also [TYPO3 Documentation](https://docs.typo3.org/m/typo3/reference-coreapi/
 
 ### WordPress Specifics
 
-#### WP-CLI and changing webroot in config.yaml
+#### WP-CLI and changing the docroot in config.yaml
 If you use something other than the root directory (`''`) for your webroot, wp-cli will not work properly. To fix this, create a `wp-cli.yml` file in the root directory that contains `path: {webroot/path}`.
 
 For example, if your webroot is `public`, your `wp-cli.yml` file should contain:
--- a/docs/content/users/usage/cms-settings.md
+++ b/docs/content/users/usage/cms-settings.md
@@ -14,19 +14,19 @@ While this reduces setup time for new users, makes it easier to try out a CMS, a
 
 There are several ways to back off DDEV’s CMS settings management:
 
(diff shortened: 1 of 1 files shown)


docs/content/users/usage/cms-settings.md
```diff
@@ -143,3 +143,13 @@ baseVariants:
 ```
 
 See also [TYPO3 Documentation](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/SiteHandling/BaseVariants.html).
+
+### WordPress Specifics
+
+#### WP-CLI and changing webroot in config.yaml
```
rfay, line 149: DDEV uses `docroot` for this.

```suggestion
#### WP-CLI and changing the docroot in config.yaml
```

docs/content/users/usage/cms-settings.md
```diff
@@ -143,3 +143,13 @@ baseVariants:
...
+If you use something other than the root directory (`''`) for your webroot, wp-cli will not work properly. To fix this, create a `wp-cli.yml` file in the root directory that contains `path: {webroot/path}`.
```
rfay, line 150:
```suggestion
If you use something other than the root directory (`''`) for your docroot, wp-cli will not work properly. To fix this, create a `wp-cli.yml` file in the root directory that contains `path: {<docroot>/path}`.
```

docs/content/users/usage/cms-settings.md
```diff
@@ -143,3 +143,13 @@ baseVariants:
...
+
+For example, if your webroot is `public`, your `wp-cli.yml` file should contain:
```
rfay, line 152:
```suggestion
For example, if your docroot is `public`, your `wp-cli.yml` file should contain:
```

rfay (changes requested):
Thanks for the followup on this!

docs/content/users/usage/cms-settings.md
rfay: Don't understand why you marked it resolved without changing the usage from webroot to docroot.

rfay (approved):
Looks good to me, thanks.

PR #7081, opened 2025-03-14
build: upgrade docker/docker to v28 usages, followup to #7078, fixes #7079

### LINKED ISSUE #7079: build: Update github.com/docker/docker v27.3.1+incompatible to v28.x
### TODO

In #7078 we update go packages, but docker/docker needs work to fix deprecations.

It doesn't look like they should be hard. I started to do them in that PR, but figure it should be done more deliberately instead of mixed in with all those other changes.

rfay:
## The Issue

- #7079
- Related: #7078 

Update docker/docker to current v28 usage.

If it passes tests maybe we're ok.


stasadev:
I pushed a new commit, where I searched for `github.com/imdario/mergo` and replaced it with `dario.cat/mergo`

The last time I tried to do this in 2024, it failed because some other package required old named `mergo`, but now it worked fine.

Then I removed the whole `indirect` section from `go.mod` and ran:

```
go get -u ./...
go mod vendor
go mod tidy
```

Which updated the entire indirect section and removed some unused dependencies.

stasadev (approved):
Approved.

rfay (commented):
Seems fine to me when it passes tests. I do wish they wouldn't do weird vanity URLs for packages like this. Not sure how that helps anybody.

PR #7082, opened 2025-03-14
test: Add quickstart test for magento2 (by @rpkoller) [skip buildkite]

rfay:
## The Issue

This is a new Quickstart test for Magento2

@rpkoller created it; I tried to fix it so we have the needed composer keys in there

### DIFF
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -33,7 +33,7 @@ teardown() {
   assert_success
 
   # add the env variable credentials to auth.json
-  run cat <<EOF > .ddev/homeadditions/.composer/auth.json
+  cat <<EOF > .ddev/homeadditions/.composer/auth.json
 {
     "http-basic": {
         "repo.magento.com": {
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -12,6 +12,11 @@ teardown() {
 }
 
 @test "Magento 2 quickstart with $(ddev --version)" {
+
+  if [ "${MAGENTO2_PUBLIC_ACCESS_KEY}" = "" ]; then
+    skip "MAGENTO_PUBLIC_ACCESS_KEY not provided (forked PR)"
+  fi
+  
   # mkdir ${PROJNAME} && cd ${PROJNAME}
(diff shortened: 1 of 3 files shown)


docs/tests/magento2.bats
```diff
@@ -0,0 +1,107 @@
...
+  run touch ./.ddev/homeadditions/.composer/auth.json
+  assert_success
+
+  # add the env variable credentials to auth.json
+  run cat <<EOF > .ddev/homeadditions/.composer/auth.json
+{
+    "http-basic": {
+        "repo.magento.com": {
+            "username": "'"$MAGENTO2_PUBLIC_ACCESS_KEY"'",
+            "password": "'"$MAGENTO2_PRIVATE_ACCESS_KEY"'"
```
rpkoller, line 41:
```suggestion
            "username": "${MAGENTO2_PUBLIC_ACCESS_KEY}",
            "password": "${MAGENTO2_PRIVATE_ACCESS_KEY}"
```

that should fix the problem. working locally with that change

docs/tests/magento2.bats
```diff
@@ -0,0 +1,101 @@
...
+  # ddev add-on get ddev/ddev-elasticsearch
+  run ddev add-on get ddev/ddev-elasticsearch
+  assert_success
+
+  # ddev start -y
+  run ddev start -y
+  assert_success
+
+  # ddev composer create --repository https://repo.magento.com/ magento/project-community-edition
+  run ddev composer create --repository https://repo.magento.com/ magento/project-community-edition
```
rpkoller, line 51: just noticed the space after '.com/' in the test as well as the quickstart. it is correct but still feels so wrong that they are using a space in such a path?

rfay: These are two arguments, `--repository url-to-repository` and `project-name`

rpkoller: i was so focused on the space as soon as i noticed it so i overlooked the --repository flag

rpkoller (approved):
looks still good with your recent changes

rfay:
This is all you, thanks for the great work.

PR #7084, opened 2025-03-14
docs: add DevDb tip to database management documentation

damms005:
This PR adds DevDb tip to database management documentation

https://ddev--7084.org.readthedocs.build/en/7084/users/usage/database-management/#database-guis

### DIFF
--- a/docs/content/users/usage/database-management.md
+++ b/docs/content/users/usage/database-management.md
@@ -65,6 +65,8 @@ The PostgreSQL database container includes normal `pg` commands like `pg_dump`.
 
 If you’d like to use a GUI database client, you’ll need the right connection details and there may even be a command to launch it for you:
 
+!!!tip "If you’re using VS Code or any of its forks (Cursor, Windsurf, etc.)" You can use [DevDb]([url](https://marketplace.visualstudio.com/items?itemName=damms005.devdb)) extension to access your database right inside the IDE.
+
 * phpMyAdmin, formerly built into DDEV core, can be installed by running `ddev add-on get ddev/ddev-phpmyadmin`.
 * Adminer can be installed with `ddev add-on get ddev/ddev-adminer`
 * The [`ddev describe`](../usage/commands.md#describe) command displays the `Host:` details you’ll need to connect to the `db` container externally, for example if you're using an on-host database browser like SequelAce.
--- a/docs/content/users/usage/database-management.md
+++ b/docs/content/users/usage/database-management.md
@@ -74,7 +74,7 @@ If you’d like to use a GUI database client, you’ll need the right connection
     * Choose a static [`host_db_port`](../configuration/config.md#host_db_port) setting for your project. For example `host_db_port: 59002` (each project’s database port should be different if you’re running more than one project at a time). Use [`ddev start`](../usage/commands.md#start) for it to take effect.
     * Use the “database” tool to create a source from “localhost”, with the proper type `mysql` or `postgresql` and the port you chose, username `db` + password `db`.
     * Explore away!
-* If you're working with VS Code or any of its forks (Cursor, Windsurf, etc.), the [DevDb](https://marketplace.visualstudio.com/items?itemName=damms005.devdb) lets you access your database directly inside the editor.
+* VS Code or any of its forks (Cursor, Windsurf, etc.) has [DevDb](https://marketplace.visualstudio.com/items?itemName=damms005.devdb), an extension that lets you access your database directly inside the editor.
 * There’s a sample custom command that will run the free MySQL Workbench on macOS, Windows or Linux. To use it, run:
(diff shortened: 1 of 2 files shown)


docs/content/users/usage/database-management.md
```diff
@@ -65,6 +65,8 @@ The PostgreSQL database container includes normal `pg` commands like `pg_dump`.
 
 If you’d like to use a GUI database client, you’ll need the right connection details and there may even be a command to launch it for you:
 
+!!!tip "If you’re using VS Code or any of its forks (Cursor, Windsurf, etc.)" You can use [DevDb]([url](https://marketplace.visualstudio.com/items?itemName=damms005.devdb)) extension to access your database right inside the IDE.
```
stasadev, line 68: Fix the tip formatting with the extension link:

```suggestion
!!!tip "If you’re using VS Code or any of its forks (Cursor, Windsurf, etc.)"
    You can use the [DevDb](https://marketplace.visualstudio.com/items?itemName=damms005.devdb) extension to access your database right inside the IDE.
```

stasadev (changes requested):

stasadev (commented):
Thanks, I updated the first post with the link to the generated docs.

And now that I look at it, I see that it does stand out, which is good, but shouldn't it be listed right before or after the PhpStorm item in that list?

stasadev:
Also, please add `DevDb` here https://github.com/ddev/ddev/blob/main/.spellcheckwordlist.txt to pacify the linter.

stasadev:
And maybe a note that it's only for Laravel, right?

So, my suggestion is to turn it from a tip to an item list.

damms005:
Thanks for your time in checking this out @stasadev 

Was literally confused if it fits in the list (because it is an extension) or should be a tip. But yes makes more sense if it is after PhpStorm since it is also an IDE.

However, not sure about adding a note for Laravel. This is because I have a [features request page](https://github.com/damms005/devdb-vscode/discussions/new?category=ideas) where I track requested features. Whenever I implement DDEV request for a new environment, then this section of the docs becomes outdated.

I am thinking adding something like "If your DDEV environment is not yet supported, please consider requesting for it"

(I will then link the "please consider requesting for it" part to the features request page linked above)

Thoughts?

docs/content/users/usage/database-management.md
```diff
@@ -74,6 +77,7 @@ If you’d like to use a GUI database client, you’ll need the right connection
     * Choose a static [`host_db_port`](../configuration/config.md#host_db_port) setting for your project. For example `host_db_port: 59002` (each project’s database port should be different if you’re running more than one project at a time). Use [`ddev start`](../usage/commands.md#start) for it to take effect.
     * Use the “database” tool to create a source from “localhost”, with the proper type `mysql` or `postgresql` and the port you chose, username `db` + password `db`.
     * Explore away!
+* If you’re using VS Code or any of its forks (Cursor, Windsurf, etc.), you can use the [DevDb](https://marketplace.visualstudio.com/items?itemName=damms005.devdb) extension to access your database right inside the IDE.
```
stasadev, line 80:
```suggestion
* If you're working with VS Code or any of its forks (Cursor, Windsurf, etc.), the [DevDb](https://marketplace.visualstudio.com/items?itemName=damms005.devdb) lets you access your database directly inside the editor.
```

stasadev (changes requested):
> I am thinking adding something like "If your DDEV environment is not yet supported, please consider requesting for it"

I tested it just now with symfony and simple php project types, and it works too.

I think you can remove `Laravel` from the DDEV related configs, because this config already works with everything else.

(See suggestions: remove the tip, and rephrase the text a bit.)

damms005:
Thanks for confirming it works on those other PHP projects. I will also update the guide on my end to reflect that.

I have effected your suggestions, and also updated the grammar to match the style used in the preceeding PhpStorm part.

stasadev (approved):
Thank you, I like it.

tyler36 (approved):
Looks good to me.

rfay:
Unless I misunderstand @nickchomey your comment would be best in an issue in https://github.com/damms005/devdb-vscode ? Or is this something to do with DDEV?

damms005:
@nickchomey thanks for checking out DevDb! 

It may not be necessary to update the DDEV docs as you suggested because:

1. DevDb docs already covers how it loads databases, and it now includes a clarification about zero-config in multi-root environments 
2. It is not exactly accurate to say DevDb only works "when the workspace is at the project root", except if you are referring specifically to the [zero-config feature](https://github.com/damms005/devdb-vscode?tab=readme-ov-file#1-zero-config-automatic-database-loading), which is one of two ways databases are loaded in DevDb, the second being a [`.devdbrc` file](https://github.com/damms005/devdb-vscode?tab=readme-ov-file#2-config-based-database-loading) suitable for instances where zero-config is not available

image

PR #7087, opened 2025-03-14
[test] don't load 1password secrets if not available

rfay:
## The Issue

Dependabot PRs and some forked PRs fail in some places where they could continue without secrets.

Instead, this PR attempts to check if the base `secrets.TESTS_SERVICE_ACCOUNT_TOKEN` is available, and if it is not, then proceed without loading 1password secrets.


rfay:
Will reopen on upstream.

PR #7088, opened 2025-03-14
[test] don't load 1password secrets if not available [skip buildkite]

rfay:
## The Issue

Dependabot PRs and some forked PRs fail in some places where they could continue without secrets.

Instead, this PR attempts to check if the base `secrets.TESTS_SERVICE_ACCOUNT_TOKEN` is available, and if it is not, then proceed without loading 1password secrets.

I'm not sure this PR will actually use the changes added here, but recreated this on upstream to give it the best chance to see.


stasadev (approved):
Looks good to me.

PR #7090, opened 2025-03-14
test: Fix secret loading [skip buildkite]

rfay:
## The Issue

- #7088 did not work right and now secrets don't get loaded or run

Trying again here with @stasadev technique.


stasadev (approved):
I see `Authenticated with Service account.`

PR #7092, opened 2025-03-15
Update docker-installation.md to use 'Docker Engine' terminology

Nick-Hope:
## The Issue

- #<issue number>

### DIFF
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -109,13 +109,13 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     ## Windows
 
-    If you’re working inside WSL2, which we recommend, you can [install Docker Engine (docker-ce) inside of it](#docker-ce-inside-windows-wsl2). Otherwise, you can [install Docker Desktop](#docker-desktop-for-windows), which works with both traditional Windows and WSL2.
+    If you’re working inside WSL2, which we recommend, you can [install Docker Engine (docker-ce) inside of it](#docker-engine-inside-windows-wsl2). Otherwise, you can [install Docker Desktop](#docker-desktop-for-windows), which works with both traditional Windows and WSL2.
 
     ### Docker Engine inside Windows WSL2
 
     Many have moved away from using Docker Desktop in favor of the Docker-provided open-source Docker Engine (`docker-ce`) package inside WSL2.
 
-    The instructions for [DDEV Installation in WSL2](ddev-installation.md#wsl2-docker-ce-inside-install-script) include Docker Engine setup and a script that does almost all the work. Please use those.
+    The instructions for [DDEV Installation in WSL2](ddev-installation.md#wsl2-docker-engine-install-script) include Docker Engine setup and a script that does almost all the work. Please use those.
 
     ### Docker Desktop for Windows
 
--- a/docs/content/users/install/docker-installation.md
+++ b/docs/content/users/install/docker-installation.md
@@ -111,11 +111,11 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     If you’re working inside WSL2, which we recommend, you can [install Docker Engine (docker-ce) inside of it](#docker-ce-inside-windows-wsl2). Otherwise, you can [install Docker Desktop](#docker-desktop-for-windows), which works with both traditional Windows and WSL2.
 
-    ### Docker CE Inside Windows WSL2
+    ### Docker Engine inside Windows WSL2
 
-    Many have moved away from using Docker Desktop in favor of the Docker-provided open-source `docker-ce` package inside WSL2.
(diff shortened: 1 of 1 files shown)


rfay:
Your point is valid, but of course you'll note that the actual installation is `sudo apt-get install docker-ce`

One reason we use "Docker CE" is that they use "docker engine" for any Docker Desktop usage, including on macOS and Windows. Other Docker Providers don't use "engine" even though that's what they are too. 

Let's continue the conversation. We don't want to confuse people, but also Docker's terminology isn't perfect. 

What do you think about just adding a note about what "Docker CE" is (The open-source Community Edition of the Docker Engine) ?

docs/content/users/install/ddev-installation.md
```diff
@@ -149,7 +149,7 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     You can install DDEV on Windows three ways:
 
-    1. [Using WSL2 with Docker inside](#wsl2-docker-ce-inside-install-script) **Recommended, best performance, most reliable**
+    1. [Using WSL2 with Docker Engine](#wsl2-docker-engine-install-script) **Recommended, best performance, most reliable**
```
rfay, line 152: "Docker inside" is in fact very poor usage here, and we can do better. "Docker Engine" isn't great, because it allows confusion with Docker Desktop, which also has "Docker Engine" in it.

docs/content/users/install/ddev-installation.md
```diff
@@ -182,9 +182,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     * Installing DDEV inside your distro; this is normally done by running one of the two scripts below, but can be done manually step-by-step as well.
     * Verify that your distro uses WSL version 2 with `wsl.exe -l -v`.
 
-    ### WSL2 + Docker CE Inside Install Script
+    ### WSL2 + Docker Engine Install Script
```
rfay, line 185: I think "Docker CE" rather than "Docker CE Inside" might be better.

docs/content/users/install/docker-installation.md
```diff
@@ -109,13 +109,13 @@ You’ll need a Docker provider on your system before you can [install DDEV](dde
 
     ## Windows
 
-    If you’re working inside WSL2, which we recommend, you can [install Docker Engine (docker-ce) inside of it](#docker-ce-inside-windows-wsl2). Otherwise, you can [install Docker Desktop](#docker-desktop-for-windows), which works with both traditional Windows and WSL2.
+    If you’re working inside WSL2, which we recommend, you can [install Docker Engine (docker-ce) inside of it](#docker-engine-inside-windows-wsl2). Otherwise, you can [install Docker Desktop](#docker-desktop-for-windows), which works with both traditional Windows and WSL2.
```
rfay, line 112: This one is OK, although can use more standardization.

rfay (changes requested):
As in the previous note, this can definitely be improved. But shouldn't you be doing the Linux instructions at the same time, since Docker CE in WSL2 is almost completely the same thing as the Linux install.

Nick-Hope:
@rfay It is difficult because "Docker CE" doesn't really seem to have a home on the web under that name. A Google for "Docker CE" takes you to https://docs.docker.com/engine/install/, which does refer to it as "Docker Engine on Linux, also known as Docker CE". Anyway, I have reverted back to "Docker CE", added clarification about what it is in both documents, and cleaned up a few bits and pieces.

One subtle change is the lower case 'i' in the heading "WSL2 + Docker CE inside Install Script". I remember misguidedly searching for a product called 'Docker CE Inside' at one point.

_"But shouldn't you be doing the Linux instructions at the same time, since Docker CE in WSL2 is almost completely the same thing as the Linux install."_
I had the same thought but I could not find any references to "Docker CE" in those docs.

rfay:
Thanks for the good thinking on this! It is a problem, and glad to see it get some cleaning action.

docs/content/users/install/ddev-installation.md
```diff
@@ -182,9 +182,9 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
     * Installing DDEV inside your distro; this is normally done by running one of the two scripts below, but can be done manually step-by-step as well.
     * Verify that your distro uses WSL version 2 with `wsl.exe -l -v`.
 
-    ### WSL2 + Docker CE Inside Install Script
+    ### WSL2 + Docker CE Install Script
 
-    This technique is our favorite, as it uses the most reliable WSL2 Docker provider (`docker-ce`), which is also free and open-source.
+    This technique is our favorite, as it uses the most reliable WSL2 Docker provider, Docker CE - the free, open-source Community Edition of [Docker Engine](https://docs.docker.com/engine/).
```
rfay, line 187: I like this phrasing.

rfay (approved):
This is awesome, thanks!

PR #7096, opened 2025-03-16
fix: remove obsolete references to non-traefik router

rfay:
## The Issue

I happened to notice that we had some code checking for traefik/non-traefik routers, and since that's obsolete, it's better to remove.

Review carefully, but this should have no impact.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2846,11 +2846,13 @@ func (app *DdevApp) Stop(removeData bool, createSnapshot bool) error {
 		}
 	}
 
-	_, _, err = app.Exec(&ExecOpts{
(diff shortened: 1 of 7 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -2850,14 +2846,13 @@ func (app *DdevApp) Stop(removeData bool, createSnapshot bool) error {
 		}
 	}
 
-	if globalconfig.DdevGlobalConfig.IsTraefikRouter() && status == SiteRunning {
```
stasadev, line 2853: `status == SiteRunning` should still be here.

stasadev (changes requested):

pkg/ddevapp/ddevapp.go
rfay: Thanks, good eye, fixed.

stasadev (approved):
Looks good to me!

PR #7099, opened 2025-03-16
test: adding kirby quickstart bats test

rpkoller:
… to it

## The Issue

- #<issue number>

bats test for the new project quickstart... for existing quickstart there wont be a quickstart necessary imho. it is basically identically to the new project quickstart except the composer create step is missing and an existing codebase is expected


rfay (approved):
Thanks!

PR #7102, opened 2025-03-16
test: symfony bats tests

rpkoller:
## The Issue

- #<issue number>

bats tests for the symfony composer and the symfony cli quickstart. also made small adjustments to the quickstarts. 

only nagging point, at the moment it is only possible to check ddev launch. it is impossible to check the install with curl. the default page that is shown is not an actual page but rather a "virtual" on... the webprofiler in the footer bar shows a 404 and curl returns no output. so uncertain how to reliably and more thoroughly check for a successful install. am a bit out of ideas and dont know symfony well enough. only idea would be if it is possible to create a page with symfony cli if that is possible and then check that page and route as a proof that symfony was properly set up. but i only found the manual set up of a page and that entailed several steps that also required code snippets.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -1000,7 +1000,7 @@ DDEV automatically updates or creates the `.env.local` file with the database in
     ```bash
     mkdir my-symfony-site && cd my-symfony-site
     ddev config --project-type=symfony --docroot=public
-    ddev start -y
+    ddev start
     ddev composer create symfony/skeleton
     ddev composer require webapp
     # When it asks if you want to include docker configuration, say "no" with "x"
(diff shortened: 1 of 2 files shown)


rfay:
Does your question in the OP still stand? Do you want to work more on that? We can take a look together.

rpkoller:
> Does your question in the OP still stand? Do you want to work more on that? We can take a look together.

hm i would be curious how that works and if it would be testable in some way. but at the moment past my knowledge and skill level in the context of symfony. maybe there are regular symfony users on discord who could shine some light?

rpkoller:
ok added a few assertions and checks. i think if the tests pass this is good to go. and one tricky detail which was the last stumbling block. usually i have an assert_success when i test `curl -sfI` but in this case with an 404 the assert of access has to be omitted and only checked against the 404 and the server: 

```
  run curl -sfI https://${PROJNAME}.ddev.site
  assert_output --partial "server: nginx"
  assert_output --partial "HTTP/2 404"
``

rfay (approved):
Seems good to me, thanks!

PR #7103, opened 2025-03-17
test: Bats test for CakePHP composer quickstart

rpkoller:
…ment

## The Issue

- #<issue number>

bats test for the composer quickstart for cakephp. i'Ve omitted the git based for now since there is no actionable git repo to build a bats test around.


rfay (approved):
Thanks!

PR #7106, opened 2025-03-17
test: simplify the zip based test and quickstart for backdrop based on feedback from @stasadev

rpkoller:
## The Issue

- #<issue number>

and we should open an issue upstream asking that backdrop folks create zip files that expand flat instead of creating a subfolder


rfay (approved):
Seems fine to me, thanks.

stasadev (approved):
Looks good to me.

PR #7107, opened 2025-03-17
test: craftcms bats test

rpkoller:
## The Issue

- #<issue number>

adding a bats test for the composer based quickstart (only remaining question if the approach for setting up the site used on the test should be used on the quickstart as well - asked @mandrasch over in discord)

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -173,6 +173,13 @@ Environment variables will be automatically added to your `.env` file to simplif
     # Boot the project and install the starter project:
     ddev start
     ddev composer create craftcms/craft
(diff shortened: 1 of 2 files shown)


rfay (approved):
Looks like a good thing, thanks!

docs/content/users/quickstart.md
```diff
@@ -173,6 +173,13 @@ Environment variables will be automatically added to your `.env` file to simplif
     # Boot the project and install the starter project:
     ddev start
     ddev composer create craftcms/craft
```
stasadev, line 175: This line needs `--no-scripts`.

PR #7112, opened 2025-03-17
test: adding silverstripe quickstart bats test

rpkoller:
## The Issue

- #<issue number>

stuck to the pattern of adding the run command as a comment. best when all bats tests are in to unify everything at once.


rfay:
Rebased, (sorry), to get all those recent merges

rfay (approved):
Seems OK to me, thanks for this!

PR #7114, opened 2025-03-18
test: attempt to fix the returned 503 error on ee tests

rpkoller:
## The Issue

- #<issue number>


rfay (approved):
Thanks, looks like the right approach to me. I'll pull after HEAD tests are done.

PR #7116, opened 2025-03-18
test: bats test for Statamic Composer quickstart [skip buildkite]

rpkoller:
## The Issue

- #<issue number>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -922,7 +922,7 @@ The Laravel project type can be used for [Statamic](https://statamic.com/) like 
     mkdir my-statamic-site && cd my-statamic-site
     ddev config --project-type=laravel --docroot=public
     ddev composer create --prefer-dist statamic/statamic
-    ddev php please make:user
+    ddev php please make:user admin@example.com --password=admin1234 --super --no-interaction
     ddev launch /cp
     ```
 === "Git Clone"
--- a/docs/tests/statamic.bats
+++ b/docs/tests/statamic.bats
@@ -25,18 +25,10 @@ teardown() {
   assert_success
 
   # fill out the interactive form
-  run expect <<EOD
-spawn ddev php please make:user
-expect "Email"
-send "admin@mail.com\r"
-expect "Name"
-send "Admin\r"
-expect "Password"
-send "admin1234\r"
-expect "Super user?"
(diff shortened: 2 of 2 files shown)


rpkoller:
the only small detail i am not sure about is why selecting YES in the prompt is not working with expect/send .... when i manually tried, it was enough to press Y (that is what i send in line 37 on the bats test followed by a return key press)... but with automatic bats tests the selection remained on NO... no biggie, that way we dont get a super user which we dont need anyway but i would have been curious why it is not picking up on the entry of `y`

stasadev (commented):
> the only small detail i am not sure about is why selecting YES in the prompt is not working with expect/send

I don't think it's working properly, I don't see any user in the database, I noticed that it creates a file in the `users` folder, when I run it manually, so:

```
run ddev php please make:user admin@example.com --password=admin1234 --super --no-interaction
assert_file_exist admin@mail.com.yaml
```

rpkoller:
> > the only small detail i am not sure about is why selecting YES in the prompt is not working with expect/send
> 
> I don't think it's working properly, I don't see any user in the database, I noticed that it creates a file in the `users` folder, when I run it manually, so:
> 
> ```
> run ddev php please make:user admin@example.com --password=admin1234 --super --no-interaction
> assert_file_exist users/admin@example.com.yaml
> ```

i am either blind or not able to search properly the net. i searched everywhere if it would be possible not to use the prompt but a command line instead. what you've posted is exactly what i was searching for. that expect and send was buggy and also not really convenient. but there is one problem. when i test locally for assert file exist bats failes on macos with zsh with ```
  -- file does not exist --
   path : ./users/admin@example.com.yaml
   --
```
but the file actually exists... on my manual test as well as commenting out the teardown and then examine the automatic bats output. in both cases the file exists with the proper content within the users directory

**Comment by stasadev** [comment]
> `file does not exist`

Probably the same thing as here #7114:

```
ddev mutagen sync
```

rpkoller:
i've tried escaping the `@` but that had no effect. and the error output was the same no matter the `@` was escaped or not. so the character was already handled properly. but stas'es suspicion was the right one, a mutagen sync fixed it. i will never ever forget again to sync mutagen after any file was altered or created. thank you!

stasadev (approved):
Looks good to me.

Issue #7044, opened 2025-03-04
Versions 1.24.0 and above fail behind a corporate proxy. Earlier version work just fine


samgurung:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Output file will be in /tmp/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/bin/ddev
/bin/ddev

These config files were loaded for project testing: [/home/ad.dcs.smcs/skg/testing/.ddev/config.yaml]
name: testing
type: php
php_version: 8.3
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.24.3
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.11}
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 22
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
No registered add-ons were found.


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /home/ad.dcs.smcs/skg/testing
.ddev/mutagen/mutagen.yml not found


======== Creating dummy project named tryddevproject-6046 in ../tryddevproject-6046 ========
Creating a new DDEV project config in the current directory (/home/ad.dcs.smcs/skg/tryddevproject-6046)
Once completed, your configuration will be written to /home/ad.dcs.smcs/skg/tryddevproject-6046/.ddev/config.yaml

Configuring a 'php' project named 'tryddevproject-6046' with docroot 'web' at '/home/ad.dcs.smcs/skg/tryddevproject-6046/web'.
For full details use 'ddev describe'.
Configuration complete. You may now run 'ddev start'.


======== OS Information ========
Linux lin-smccomp401 6.11.0-17-generic #17~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 20 22:48:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux


[...]

PR #7118, opened 2025-03-18
fix: set `NO_PROXY=*` in ddev-router to allow internal connections, use default Traefik config file location

stasadev:
## The Issue

- From https://github.com/ddev/ddev/pull/7061#issuecomment-2726978833

	only IP access was working - I couldn't open it via domain name (myapp.ddev.site). [This comment](https://github.com/ddev/ddev/issues/7044#issuecomment-2697303134) was very helpful - after adding `ddev-myapp-web` (and `ddev-myapp-phpmyadmin`) to `noProxy` in `~/.docker/config.json`, I was finally able to run both my projects without problems (and without addon).

- From https://github.com/ddev/ddev/pull/7061#issuecomment-2727509330

	Adding container names in `noProxy` in `~/.docker/config.json` as described in my previous post. BTW I'm not sure why I can't use wildcards (`ddev-*-web,ddev-*-phpmyadmin`) - I need to setup names for every project, so my `~/.docker/config.json` looks like this:
	
	```json
	{
	  "proxies": {
	    "default": {
	       "httpProxy": "http://username:p%%40ssword@192.168.1.10:3128",
	       "httpsProxy": "http://username:p%%40ssword@192.168.1.10:3128",
	       "noProxy": "localhost,127.0.0.1/8,::1,*.ddev.site,ddev-myapp1-web,ddev-myapp1-phpmyadmin,ddev-myapp2-web,ddev-myapp2-phpmyadmin"
	    }
	  }
	}
	```

- From https://github.com/ddev/ddev/pull/7061#issuecomment-2727558052

	`"noProxy": "127.0.0.0/8,::1,localhost,*.ddev.site,*.ddev.local,.ddev.site,.ddev.local,ddev-,ddev-*,ddev-*-web,ddev-*-phpmyadmin,ddev-myapp-phpmyadmin"`
	I could only access phpmyadmin via domain - my app was still not accessible.
	
	Errors are coming from Squid (`(111) Connection refused`) - for me it looks like Traefik uses proxy data from `~/docker/config.json` and tries to access these containers via my proxy (when i use `docker logs ddev-router` I can see these container names there):
	
	```
	172.18.0.1 - - [16/Mar/2025:17:11:26 +0000] "GET / HTTP/2.0" 503 4348 "-" "-" 1 "myapp-web-80-https@file" "http://ddev-myapp-web:80" 92ms
	172.18.0.1 - - [16/Mar/2025:17:11:28 +0000] "GET / HTTP/2.0" 503 4373 "-" "-" 2 "myapp-phpmyadmin-80-https@file" "http://ddev-myapp-phpmyadmin:80" 2ms
	```

Related:

- https://github.com/docker/compose/issues/11433

Sets no_proxy variable:

```
$ docker exec -it ddev-router bash -c 'env | grep -i no_proxy'                
no_proxy=*
NO_PROXY=*
```

And in addition, this change with proxy env makes the PR below unneeded:

- #7045

And according to [Traefik docs](https://doc.traefik.io/traefik/getting-started/configuration-overview/#configuration-file), `/etc/traefik/traefik.yaml` is read by default, so I made it a symlink to `/mnt/ddev-global-cache/traefik/.static_config.yaml`. This change made it possible to use Traefik commands inside the container without remembering about `--configFile`, so easy:

```
$ docker exec -it ddev-router traefik healthcheck --ping   
OK: http://:10999/ping
````

---

I was able to reproduce it under proxy.

This is how it looks without the fix (with squid proxy):

image

The reason why I didn't see/check this before is that I didn't open the site in a browser through a proxy (I only checked the Traefik dashboard).

Configure a proxy in `~/.docker/config.json` according to #7061 (without adding stuff like `ddev-d11-web`)
Start a project and open it in the browser, it should be accessible.

### DIFF
--- a/containers/ddev-traefik-router/test/containertest.sh
+++ b/containers/ddev-traefik-router/test/containertest.sh
@@ -64,4 +64,4 @@ if ! containercheck; then
 fi
 
 # Make sure internal access to monitor port is working
-docker exec -t $CONTAINER_NAME bash -c 'traefik healthcheck --ping.entryPoint=ping --entryPoints.ping.address=127.0.0.1:${TRAEFIK_MONITOR_PORT} --ping 2>&1' || (echo "Failed to run http healthcheck inside container" && exit 104)
+docker exec -t $CONTAINER_NAME bash -c 'traefik healthcheck --ping 2>&1' || (echo "Failed to run http healthcheck inside container" && exit 104)
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -3546,7 +3546,6 @@ func TestMultipleComposeFiles(t *testing.T) {
 	assert.NoError(err)
 	_, err = app.ReadConfig(true)
 	require.NoError(t, err)
-	app.DockerEnv()
 	err = app.WriteDockerComposeYAML()
 	require.NoError(t, err)
 
@@ -3585,24 +3584,7 @@ func TestMultipleComposeFiles(t *testing.T) {
 			}
 			// Verify that users can add and override network properties
 			if networks, ok := w["networks"].(map[string]interface{}); ok {
-				if network, ok := networks["ddev_default"].(map[string]interface{}); ok {
-					if aliases, ok := network["aliases"].([]interface{}); ok {
-						aliasList := make([]string, len(aliases))
-						for i, alias := range aliases {
-							aliasList[i], ok = alias.(string)
-							if !ok {
-								t.Errorf("Failed to parse alias in web ddev_default at index %d: it is not a string: %v", i, alias)
-							}
-						}
-						// The aliases should contain the expected values
-						// From DDEV codebase (needed for Traefik load balancer to work under proxies)
-						assert.Contains(aliasList, ddevapp.GetContainerName(app, "web")+"."+app.ProjectTLD)
-						// From docker-compose.override.yaml
-						assert.Contains(aliasList, "web-extra-alias")
-					} else {
-						t.Errorf("Failed to parse 'aliases' in web ddev_default network: %v", network)
-					}
-				}
+				assert.Nil(networks["ddev_default"])
 				if network, ok := networks["default"].(map[string]interface{}); ok {
 					assert.Equal(1, network["priority"])
 				} else {
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -3546,6 +3546,7 @@ func TestMultipleComposeFiles(t *testing.T) {
 	assert.NoError(err)
 	_, err = app.ReadConfig(true)
 	require.NoError(t, err)
+	app.DockerEnv()
 	err = app.WriteDockerComposeYAML()
 	require.NoError(t, err)
 
@@ -3584,7 +3585,24 @@ func TestMultipleComposeFiles(t *testing.T) {
 			}
 			// Verify that users can add and override network properties
 			if networks, ok := w["networks"].(map[string]interface{}); ok {
-				assert.Nil(networks["ddev_default"])
+				if network, ok := networks["ddev_default"].(map[string]interface{}); ok {
+					if aliases, ok := network["aliases"].([]interface{}); ok {
+						aliasList := make([]string, len(aliases))
+						for i, alias := range aliases {
+							aliasList[i], ok = alias.(string)
+							if !ok {
+								t.Errorf("Failed to parse alias in web ddev_default at index %d: it is not a string: %v", i, alias)
+							}
+						}
+						// The aliases should contain the expected values
+						// From DDEV codebase (needed for Traefik load balancer to work under proxies)
+						assert.Contains(aliasList, ddevapp.GetContainerName(app, "web")+"."+app.ProjectTLD)
+						// From docker-compose.override.yaml
+						assert.Contains(aliasList, "web-extra-alias")
+					} else {
+						t.Errorf("Failed to parse 'aliases' in web ddev_default network: %v", network)
+					}
+				}
 				if network, ok := networks["default"].(map[string]interface{}); ok {
 					assert.Equal(1, network["priority"])
 				} else {
--- a/pkg/ddevapp/router_compose_template.yaml
+++ b/pkg/ddevapp/router_compose_template.yaml
@@ -35,8 +35,6 @@ services:
       - http_proxy=
       - HTTPS_PROXY=
       - https_proxy=
-      - NO_PROXY=
-      - no_proxy=
     restart: "no"
     healthcheck:
       test: "/healthcheck.sh"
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: pkg/ddevapp/ddevapp_test.go [hunk 1]
```diff
@@ -3546,6 +3546,7 @@ func TestMultipleComposeFiles(t *testing.T) {
 	assert.NoError(err)
 	_, err = app.ReadConfig(true)
 	require.NoError(t, err)
+	app.DockerEnv()
```
stasadev, line 3549: Added because the test didn't work if I only ran one test.

Caused by:

- #7012

rfay: Actually, caused by #472 true? I think #7012 was unsuccessful?

stasadev: Yes.

rfay:
Awesome and brilliant, although sad that we need to do this. 

I worry about side-effects.

Could you please add a bit more to the OP explaining what the practical effect of this is? Give a practical example of how the router now interacts with containers, etc. Thanks!

maciazek:
Tested today at my workplace with real app on fresh VM - it works without any problems. Thank you!

ddev version v1.24.3-42-g14a1a52ef
noProxy: `127.0.0.1,localhost,::1,*.ddev.site,*.ddev.local`
Tested the app itself, phpmyadmin, composer and npm.

Domain without asterisk work for me too.

rfay:
I'm unable to reproduce the perceived problem in my lab setup. I'm using an Ubuntu 22.04 system with no internet, which is using a proxy (tinyproxy running on macOS).  In Ubuntu have the docker daemon configured, the docker client configured, and apt configured for proxy, and the Ubuntu system itself has proxy settings, which show up in the terminal.

The relevant environment variables inside ddev-router are these:

```
no_proxy=127.0.0.1/8,.ddev.site
https_proxy=http://10.37.129.2:8888
NO_PROXY=127.0.0.1/8,.ddev.site
HTTPS_PROXY=http://10.37.129.2:8888
HTTP_PROXY=http://10.37.129.2:8888
http_proxy=http://10.37.129.2:8888
```

I have done nothing at all with special noproxy settings for containers. 

However, I *can* demonstrate the problem inside ddev-router:

```
rfay@ub-2204-use-proxy:~/workspace/d11$ docker exec -it ddev-router curl -I http://ddev-d11-web
HTTP/1.1 500 Unable to connect
Server: tinyproxy/1.11.2
Content-Type: text/html
Connection: close
```

My suspicion is that traefik itself isn't running in the context of the environment variables and is not using them.

rfay:
Updated comment above showing the error of my ways.

rfay:
> Errors are coming from Squid ((111) Connection refused) - for me it looks like Traefik uses proxy data from ~/docker/config.json and tries to access these containers via my proxy (when i use docker logs ddev-router I can see these container names there):

I think squid is reporting badly; but tinyproxy gives a 500 because it can't proxy. A 503 from squid would have been reasonable.

rfay:
Rebased just for convenience.

rfay (approved):
This all looks great and tests great, amazing thinking, excellent workaround.

My remaining questions:
* Wouldn't it have bene simpler to use [links](https://docs.docker.com/compose/how-tos/networking/#link-containers) instead instead of network alias?
* I looked for network alias in docker docs and didn't find it, if it's not already here, please add the link to the PR.
* I was surprised that the "network alias" was essentially the container FQDN (like a `link` would be, but I expected it to be just a TLD.

Congrats on an amazingly thoughtful approach.

rfay:
Oh, I had another thought on a completely different way to solve this: If we can prevent the HTTP*_PROXY environment variables from happening inside ddev-router, this problem would go away (I think) and ddev-router doesn't need to access the outside world.

stasadev:
> However, I *can* demonstrate the problem inside ddev-router:

```
rfay@ub-2204-use-proxy:~/workspace/d11$ docker exec -it ddev-router curl -I http://ddev-d11-web
HTTP/1.1 500 Unable to connect
Server: tinyproxy/1.11.2
Content-Type: text/html
Connection: close
```

I had the same result under proxy, and `curl -I http://ddev-d11-web.ddev.site` refused to connect too, so confusing.

---

> Oh, I had another thought on a completely different way to solve this: If we can prevent the HTTP*_PROXY environment variables from happening inside ddev-router, this problem would go away (I think) and ddev-router doesn't need to access the outside world.

Amazing idea, so simple and so powerful 🙌
Implemented it.

---

In addition, when we don't use proxy inside the container, it's not needed to check `127.0.0.1` explicitly, so I removed my `traefik healthcheck` workaround from:

- #7045

According to [Traefik docs](https://doc.traefik.io/traefik/getting-started/configuration-overview/#configuration-file), `/etc/traefik/traefik.yaml` is read by default, so I made it a symlink to `/mnt/ddev-global-cache/traefik/.static_config.yaml`. This change made it possible to use Traefik commands inside the container without remembering about `--configFile`, so easy:

```
$ docker exec -it ddev-router traefik healthcheck --ping   
OK: http://:10999/ping
````

File: containers/ddev-traefik-router/test/containertest.sh [hunk 2]
```diff
@@ -64,4 +64,4 @@ if ! containercheck; then
 fi
 
 # Make sure internal access to monitor port is working
-docker exec -t $CONTAINER_NAME bash -c 'traefik healthcheck --ping.entryPoint=ping --entryPoints.ping.address=127.0.0.1:${TRAEFIK_MONITOR_PORT} --ping 2>&1' || (echo "Failed to run http healthcheck inside container" && exit 104)
+docker exec -t $CONTAINER_NAME bash -c 'traefik healthcheck --ping 2>&1' || (echo "Failed to run http healthcheck inside container" && exit 104)
```
rfay, line 67: You made regular ping work again!

pkg/ddevapp/router_compose_template.yaml
```diff
@@ -31,6 +29,12 @@ services:
       - USE_LETSENCRYPT={{ .letsencrypt }}
         {{ end }}{{/* end if .letsencrypt */}}
       - TZ={{ .Timezone }}
+      # Unset proxy environment variables for Traefik to prevent blocking internal container connections
+      # Docker automatically sets both uppercase and lowercase variables
+      - HTTP_PROXY=
+      - http_proxy=
+      - HTTPS_PROXY=
+      - https_proxy=
```
rfay, line 37: Yay!

Now I wonder if something like `NO_PROXY=*` would solve it all.

rfay (approved):
Looks great, tests great, thanks!

pkg/ddevapp/router_compose_template.yaml
stasadev: Yes, `NO_PROXY=*` works too.

rfay:
Tested again, works great, thanks for the great work!

PR #7119, opened 2025-03-18
docs: Update quickstart.md to remove Drupal CMS zip file instructions

phenaproxima:
## The Issue
The `launch-drupal-cms.sh` script mentioned in the docs is removed from Drupal CMS 1.1.0, which releases tomorrow. We should remove all mention of it from DDEV's docs as well.

N/A

## Automated Testing Overview
N/A

## Release/Deployment Notes
N/A

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -241,15 +241,6 @@ Read more about customizing the environment and persisting configuration in [Pro
     ddev launch
     ```
 
-    or use the ZIP file download technique:
-
-    ```bash
-    curl -o my-drupal-site.zip -fL https://www.drupal.org/download-latest/cms
-    unzip my-drupal-site.zip && rm -f my-drupal-site.zip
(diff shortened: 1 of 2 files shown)


rfay (approved):
Looks great to me. @rpkoller I guess we'll need an update of the quickstart test to remove that section? Doesn't have to go here, but it can.

stasadev:
To remove

https://github.com/ddev/ddev/blob/72c67de02a1cd1d077bbcdd8474c0c96df87e22f/docs/tests/drupal.bats#L131

rpkoller:
> Looks great to me. @rpkoller I guess we'll need an update of the quickstart test to remove that section? Doesn't have to go here, but it can.

yep the test stas linked to should be removed i agree. but not pressing since it was skipped anyway. but would be good never the less to remove everything in one go.

rfay:
Removed quickstart for zipball

PR #7124, opened 2025-03-19
feat: add MutagenSync annotation for custom commands, for #7053

stasadev:
## The Issue

- #7053
- https://github.com/backdrop-ops/ddev-backdrop-bee/pull/8

We need to have a way to run `ddev mutagen sync` automatically.

Adds `## MutagenSync: true` annotation.

I added it to the most of our bundled `web` commands.
(I'm not 100% sure of the changes I've made, but I think we'll be able to figure out where we need `MutagenSync` or not in the review.)

```
ddev config global --performance-mode=mutagen
```

And:

```
mkdir my-backdrop-site && cd my-backdrop-site
ddev add-on get backdrop-ops/ddev-backdrop-bee
ddev start
ddev bee download-core
ddev bee si --username=admin --password=admin --db-name=db --db-user=db --db-pass=db --db-host=db --auto
ddev launch
```

Mutagen Sync should run before and after each of `ddev bee` commands:

```
$ DDEV_VERBOSE=true DDEV_DEBUG=true ddev bee status 2>&1 | grep 'Mutagen sync status'
2025-03-19T20:10:53.005 Mutagen sync status my-backdrop-site in MutagenSyncFlush(): status='ok', short='watching', err='<nil>'
2025-03-19T20:10:53.698 Mutagen sync status my-backdrop-site in MutagenSyncFlush(): status='ok', short='watching', err='<nil>'
```

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -203,7 +203,7 @@ func addCustomCommandsFromDir(rootCmd *cobra.Command, app *ddevapp.DdevApp, serv
 
 		// Run the command with mutagen sync or not
 		mutagenSync := false
-		if val, ok := directives["NutagenSync"]; ok {
+		if val, ok := directives["MutagenSync"]; ok {
 			if val == "true" {
 				mutagenSync = true
 			}
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -492,7 +492,9 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 		}
 		app.DockerEnv()
 
-		runMutagenSync(app, mutagenSync)
+		if service == "web" {
+			runMutagenSync(app, mutagenSync)
+		}
 
 		osArgs := []string{}
 		if len(os.Args) > 2 {
@@ -519,7 +521,9 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			util.Failed("Failed to run %s %v: %v", name, strings.Join(osArgs, " "), err)
 		}
 
-		runMutagenSync(app, mutagenSync)
+		if service == "web" {
+			runMutagenSync(app, mutagenSync)
+		}
 	}
 }
 
@@ -561,11 +565,11 @@ func findDirectivesInScriptCommand(script string) map[string]string {
 }
 
 func runMutagenSync(app *ddevapp.DdevApp, mutagenSync bool) {
-	if mutagenSync {
+	if mutagenSync && app.IsMutagenEnabled() {
 		if status, _ := app.SiteStatus(); status == ddevapp.SiteRunning {
 			err := app.MutagenSyncFlush()
 			if err != nil {
-				util.Warning("failed to app.MutagenSyncFlush: %v", err)
+				util.Warning("Could not flush Mutagen: %v", err)
 			}
 		}
 	}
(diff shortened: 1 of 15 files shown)


cmd/ddev/cmd/commands.go
```diff
@@ -479,6 +492,8 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 		}
 		app.DockerEnv()
 
+		runMutagenSync(app, mutagenSync)
```
rfay, line 495: These should only be run on the `web` container, true?

cmd/ddev/cmd/commands.go
```diff
@@ -542,3 +559,14 @@ func findDirectivesInScriptCommand(script string) map[string]string {
 
 	return directives
 }
+
+func runMutagenSync(app *ddevapp.DdevApp, mutagenSync bool) {
+	if mutagenSync {
+		if status, _ := app.SiteStatus(); status == ddevapp.SiteRunning {
+			err := app.MutagenSyncFlush()
+			if err != nil {
+				util.Failed("failed to app.MutagenSyncFlush: %v", err)
```
rfay, line 568: I suspect a `util.Warning()` would be gentler

```suggestion
				util.Warning("failed to app.MutagenSyncFlush: %v", err)
```

cmd/ddev/cmd/commands.go
```diff
@@ -542,3 +559,14 @@ func findDirectivesInScriptCommand(script string) map[string]string {
 
 	return directives
 }
+
+func runMutagenSync(app *ddevapp.DdevApp, mutagenSync bool) {
+	if mutagenSync {
```
rfay, line 564: Here or somewhere needs to check to see if mutagen is enabled and the current project is running.

docs/content/users/extend/custom-commands.md
```diff
@@ -288,10 +288,16 @@ Use `ExecRaw: true` to pass command arguments directly to the container as-is.
...
 For example, when `ExecRaw` is true, `ddev yarn --help` returns the help for `yarn`, not DDEV's help for the `ddev yarn` command.
 
-We recommend  using this annotation for all container commands. The default behavior is retained to avoid breaking existing commands.
+We recommend using this annotation for all container commands. The default behavior is retained to avoid breaking existing commands.
 
 Example: `## ExecRaw: true`
 
+### `MutagenSync` Annotation
+
+Use `MutagenSync: true` to ensure [Mutagen](../install/performance.md#mutagen) sync runs before and after the command.
```
rfay, line 297:
```suggestion
Use `MutagenSync: true` to ensure [Mutagen](../install/performance.md#mutagen) sync runs before and after the command (where mutagen is enabled and the project is running).
```

rfay (changes requested):
Nice work. I agree with the commands where you added this.

I haven't manually tested yet.

rfay:
Do you think this needs a test? Do the other annotations have tests?

cmd/ddev/cmd/commands.go
stasadev: `app.MutagenSyncFlush()` already checks for enabled/disabled Mutagen, but it doesn't harm to add it here as well.

stasadev:
> Do you think this needs a test? Do the other annotations have tests?

I didn't see any tests for annotations. They should be partially tested by checking the commands in `~/.ddev/commands/web`

rfay:
Rebased

rfay (approved):
Thanks for this. It seems great to me, the code is great and it tests OK. 

After this the backdrop quickstart change can go in if I remember right.

As we talked about in Discord, `nvm` probably doesn't need the annotation, since it only affects the in-container home directory, which isn't mutagen-synced.

PR #7126, opened 2025-03-19
test: adjust openmage bats test assertions to the now available demo content

rpkoller:
## The Issue

- #7091

according to randy tests were failing on main for the openmage bats test. turn out the assertion testing the frontend was failing the test. well turns out the fix by stas got already merged https://github.com/OpenMage/magento-lts/pull/4720 and due to the fact the test and quickstart is using `git clone https://github.com/OpenMage/magento-lts .`the fix was already available that way the demo site is completely different: 

Screenshot 2025-03-19 at 22 17 26

adjusted the assertion to the now working demo content. only question is is it a good idea to recommend a quickstart that is living on the edge aka the head for openmage?  ( a good thing in case of the bug but in the long run for less technical folks getting in touch with openmage for the first time i am unsure)

### DIFF
--- a/docs/tests/openmage.bats
+++ b/docs/tests/openmage.bats
@@ -45,7 +45,8 @@ teardown() {
   # Check if the frontend is working
   run curl -sf https://${PROJNAME}.ddev.site
   assert_success
-  assert_output --partial "2020 OpenMage Demo Store. All Rights Reserved."
+  assert_output --partial "<title>Madison Island</title>"
+  assert_output --partial "<meta name=\"keywords\" content=\"Magento, Varien, E-commerce\" />"
 
   # Check if the admin is working
   run curl -sf https://${PROJNAME}.ddev.site/index.php/admin/


stasadev (approved):
Thank you!

rpkoller:
@sreichel from your perspective what would be the best approach to stick to with the quickstart? i know we've talked about composer before and you wrote something like the composer approach isnt recommended at this point for openmage and the git based approach would be more robust but providing a version of openmage head (as outlined in the issue summary) seems still a bit challenging in particular you have less technical folks who want to just try openmage and are unaware of the implications.

sreichel:
I am busy atm, so i'll pass it to @kiatng @colinmollenhour @addison74

I'm not sure whats the best way to make custom DDEV-commands available for composer install. The live inside OM.  (https://github.com/OpenMage/magento-lts/pull/4709). 

For quickstart composer-install I thinks the easiest way is to use plain CLI install command, instead of `openmage-install` or `cp` commands to root DDEV. 

Please give me some days to think about it.

And, much thanks for all you help :)

PR #7128, opened 2025-03-20
test: improve `ddev debug test`

stasadev:
## The Issue

- `ddev debug test` creates test project, which is probably included to instrumentation
- `curl` on host runs without explicit timeout

- Adds `export DDEV_NO_INSTRUMENTATION=true` inside the script.
- Runs `DOCKER_HOST="" docker context ls` with empty DOCKER_HOST (because it always shown as a warning, but we set this variable ourselves in DDEV).
- Fixes `printf` warnings shown to me in the IDE.
- Adds `docker system df` output with tips how to clean it.
- Do not show warning about instrumentation in `ddev version` if `DDEV_NO_INSTRUMENTATION=true`.
- Adds `curl --connect-timeout 10 --max-time 20` on the host to prevent people from using `^C`.

Run `ddev debug test`

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -9,10 +9,13 @@
 # If you have NFS enabled globally, please temporarily disable it with
 # `ddev config global --performance-mode-reset`
 
+# Disable instrumentation inside `ddev debug test`
+export DDEV_NO_INSTRUMENTATION=true
+
 PROJECT_NAME=tryddevproject-${RANDOM}
 
 function header {
-  printf "\n\n======== $1 ========\n"
+  printf "\n\n======== %s ========\n" "$1"
 }
 
 function docker_desktop_version {
@@ -32,7 +35,7 @@ function docker_desktop_version {
 
 
 if [[ ${PWD} != ${HOME}* ]]; then
-  printf "\n\nWARNING: Project should usually be in a subdirectory of the user's home directory.\nInstead it's in ${PWD}\n\n"
+  printf "\n\nWARNING: Project should usually be in a subdirectory of the user's home directory.\nInstead it's in %s\n\n" "${PWD}"
 fi
 
 header "Output file will be in $1"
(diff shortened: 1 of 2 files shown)


cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -32,7 +35,7 @@ function docker_desktop_version {
 
 
 if [[ ${PWD} != ${HOME}* ]]; then
-  printf "\n\nWARNING: Project should usually be in a subdirectory of the user's home directory.\nInstead it's in ${PWD}\n\n"
+  printf "\n\nWARNING: Project should usually be in a subdirectory of the user's home directory.\nInstead it's in %s\n\n" "${PWD}"
```
rfay, line 38: Not sure why to make these printf changes, but they're fine.

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -62,8 +65,9 @@ ddev add-on list --installed
 header "mutagen situation"
 
 echo "looking for #ddev-generated in mutagen.yml in project ${PWD}"
+echo
 if [ -f .ddev/mutagen/mutagen.yml ]; then
-  if grep '#ddev-generated' .ddev/mutagen/mutagen.yml; then
+  if grep -q '#ddev-generated' .ddev/mutagen/mutagen.yml; then
```
rfay, line 70: Excellent. I always do `>/dev/null` and `-q` is better. And regardless, it's good to lose that output here.

cmd/ddev/cmd/scripts/test_ddev.sh
```diff
@@ -153,13 +160,30 @@ rancher-desktop)
...
+docker system df
+
+echo "
+  Tips:
+  1. Periodically check your Docker filesystem usage with 'docker system df'
+  2. Use 'docker builder prune' to remove unused Docker build cache (it doesn't remove your data)
+  3. To remove all containers and images (it doesn't remove your data):
+    \`\`\`
+    ddev poweroff
+    docker rm -f \$(docker ps -aq)
```
rfay, line 182:
```suggestion
    docker rm -f \$(docker ps -aq) || true
```

rfay (approved):
Thanks!

cmd/ddev/cmd/scripts/test_ddev.sh
stasadev: I followed this recommendation https://github.com/koalaman/shellcheck/wiki/SC2059

PR #7130, opened 2025-03-21
docs: fix example file name

stasadev:
## The Issue

Example has a wrong name.

Fixes it.

See https://ddev--7130.org.readthedocs.build/en/7130/users/extend/custom-compose-files/#third-party-services-may-need-to-trust-ddev-webserver


rfay (approved):
Wow, didn't remember at all that we had these docs :) 

Pretty sure it could be simplified, but probably works.

PR #7133, opened 2025-03-22
test: Add OpenMage composer quickstart and tests

sreichel:
Follow-up to https://github.com/ddev/ddev/pull/7091, adding composer quickstart.

docs:

https://ddev--7133.org.readthedocs.build/en/7133/users/quickstart/#openmage


docs/content/users/quickstart.md
```diff
@@ -767,6 +767,30 @@ ddev magento setup:upgrade
...
+    ddev composer config extra.enable-patching true
+    ddev composer config extra.magento-core-package-type "magento-source"
+    ddev composer config allow-plugins.cweagans/composer-patches true
+    ddev composer config allow-plugins.magento-hackathon/magento-composer-installer true
+    ddev composer config allow-plugins.aydin-hassan/magento-core-composer-installer true
+    ddev composer config allow-plugins.openmage/composer-plugin true
+    ddev composer require --no-update "aydin-hassan/magento-core-composer-installer":"^2.1.0" "openmage/magento-lts":"^20.13"
+    ddev . wget -O .ddev/commands/web/openmage-install https://raw.githubusercontent.com/OpenMage/magento-lts/refs/heads/main/.ddev/commands/web/openmage-install
+    ddev start
+    ddev composer install
```
rpkoller, line 786:
```suggestion
    ddev . wget -O .ddev/commands/web/openmage-install https://raw.githubusercontent.com/OpenMage/magento-lts/refs/heads/main/.ddev/commands/web/openmage-install
    ddev composer install
```

docs/content/users/quickstart.md
```diff
@@ -767,6 +767,30 @@ ddev magento setup:upgrade
     ddev launch /admin
     ```
 
+=== "Composer"
+
+    ```bash
+    mkdir my-openmage-site && cd my-openmage-site
+    ddev config --project-type=magento --docroot=public_test --php-version=8.1 --web-environment-add=MAGE_IS_DEVELOPER_MODE=1
+    ddev composer init --name "openmage/openmage-composer-test" --description "OpenMage starter project" --type "project" -l "OSL-3.0" -s "dev" -q
```
rpkoller, line 775:
```suggestion
    ddev config --project-type=magento --docroot=public_test --php-version=8.1 --web-environment-add=MAGE_IS_DEVELOPER_MODE=1
    ddev start
    ddev composer init --name "openmage/openmage-composer-test" --description "OpenMage starter project" --type "project" -l "OSL-3.0" -s "dev" -q
```
the `ddev start` has to go before the first `ddev composer`  command, since that is starting up the web container anyway, plus in the test you have the `ddev start -y` moved right before `ddev composer init`anyway

docs/tests/openmage.bats
```diff
@@ -53,3 +53,79 @@ teardown() {
...
+  # composer install
+  run ddev composer install
+  assert_success
+
+  # download OpenMage install command
+  ddev . wget -O .ddev/commands/web/openmage-install https://raw.githubusercontent.com/OpenMage/magento-lts/refs/heads/main/.ddev/commands/web/openmage-install
+  assert_success
+
+  # Silent OpenMage install with sample data
+  run ddev openmage-install -q
```
rpkoller, line 107:
```suggestion
  ddev . wget -O .ddev/commands/web/openmage-install https://raw.githubusercontent.com/OpenMage/magento-lts/refs/heads/main/.ddev/commands/web/openmage-install
  assert_success

  run ddev mutagen sync
  assert_success
  
  # Silent OpenMage install with sample data
  run ddev openmage-install -q
```

the test currently fails because the openmage-install command is not found. locally the test works just fine. so i wonder if it is another timing problem and files downloaded are not available yet for mutagen so lets see if ddev mutagen sync is fixing that failure

rpkoller (commented):
thanks for working on the composer based quickstart! found one small nitpick for the composer based quickstart in regard of the time when the ddev start is called. 

aside that i wonder if it  would make sense to switch the order of the two tabs from Git Clone Composer to Composer Git Clone. For other quickstarts the composer tab is usually first and the git clone comes after. and i wonder if it would make sense to also add the detail about git clone that it is mainly recommended for contributors on https://docs.openmage.org/users/install/use-composer/#create-new-project ? 

and one question that is out of the scope for this quickstart and more of a question for the openmage project. would it be possible to simplify the composer based install so you only have a single `ddev composer create`  command. that many composer config and init commands are sort of intimidating :D 

but everything, the quickstart same as the test, work out fine in manual tests locally.

docs/content/users/quickstart.md
sreichel: Thanks.

docs/content/users/quickstart.md
sreichel: >  would it be possible to simplify the composer based install so you only have a single `ddev composer create` command.

Already working on it - but i will take some days (internal reasons).

>  i wonder if it would make sense to also add the detail about git clone that it is mainly recommended for contributors on

Would it be enough to rename the tab to "Git clone (for contributors)" ?

docs/tests/openmage.bats
sreichel: Seems to work now. Thanks. 

Do you have an idea  for better access to `openmage-install` command, not using `wget`? Move that custum commands to ddev-addon? And add them autmatically when project-type is "openmage"?

rpkoller: hmmm that is probably a question better suited for randy or stas, since both know the whole development part and the ddev internals inside out, but i wonder if you already have a project type openmage why not provide the openmage-install command in the context of that project type? isnt that the purpose of a project type exactly for cases like that (in particular if there is a project type already in place)?  with a project type you are able to provide files (https://ddev.readthedocs.io/en/stable/developers/project-types/  "You’ll likely need templates for settings files, use the Drupal or TYPO3 templates as examples, for example pkg/ddevapp/drupal and pkg/ddevapp/typo3. Those templates have to be loaded at runtime as well")

docs/content/users/quickstart.md
rpkoller: > Already working on it - but i will take some days (internal reasons).

no rush 

> Would it be enough to rename the tab to "Git clone (for contributors)" ?

personally i am drawn. on one hand it might be enough renaming the tab, but if you take the perspective of a less technical person, and take into consideration that the concept of contribution isn't surfaced anywhere in the quickstarts guide, things might be potentially confusing for folks unfamiliar with the concept. so the other option might be instead of renaming the tab going with the Tip box inside the tab instead. that way it could be communicated that a git clone also contains all the branches of previous versions and so on. the consequences of that decision going with git clone might be unapparent other wise for some folks. and with a tip things could be clarified. what do others think about that detail?

docs/tests/openmage.bats
stasadev: I think the `openmage-install` command should be provided by DDEV if it's a DDEV-specific command.

However, before that, OpenMage should have a dedicated GitHub starter project that’s published to Packagist. This way, it won’t be limited to DDEV, and people will be able to use it with other tools as well.

It would be similar to projects like https://github.com/symfony/skeleton or https://github.com/laravel/laravel

(Initially, I thought https://github.com/OpenMage/magento-lts was that project, but I realize now that it’s not.)

docs/content/users/quickstart.md
stasadev: I don’t think we haveve had a quickstart setup similar to the current one for OpenMage source.

However, with @sreichel working on it, I don't mind at all that we're using the "Git clone (for contributors)". While the wording might be a bit long, it's quite descriptive to me.

docs/tests/openmage.bats
sreichel: > I don’t think we have had a quickstart setup similar to the current one for OpenMage source.

OpenMage is the inofficial successor to Magento 1, that hit EOL some years ago. There was no support for composer at all. We have added composer components, but never cared about a "starter project" like Drupal/Symfony/Laravel.

This had should be done long time ago, instead if relying on multiple composer commands.

Thanks for showing impotance of it. (I use git nearly 99%, i did not see it.).

>  think the `openmage-install` command should be provided by DDEV if it's a DDEV-specific command.
>
> However, before that, OpenMage should have a dedicated GitHub starter project that’s published to Packagist. 

That would be great. There will be a composer "starter project"" asap, while working on next [level](https://github.com/OpenMage/magento-lts/pull/4618).

I think that PR is ready to merge, but beyond that:

- [ ] create OM composer starter project
- [ ] complete https://github.com/ddev/ddev/pull/7135
- [ ] update DDEV tests and docs (syned to updated OM)

docs/content/users/quickstart.md
sreichel: > that way it could be communicated that a git clone also contains all the branches of previous versions and so on

Good point, but i think that details belongs to OM's documentation, not DDEV's. Imho, DDEV quickstart should contain the minimum instructions to start a new project,  When using git clone you normaly know about branches or it should be in OM docs.

stasadev (approved):
It looks good and the tests are good, thank you!

I pushed one more small commit [`b4cd3ff` (#7133)](https://github.com/ddev/ddev/pull/7133/commits/b4cd3ff565a2ddd713e1b64b68390d05e407c4c7)
- use `ddev exec` instead of `ddev .`, because dot is an alias, not the main command
- improve a note about `docroot`
- move info about OpenMage commands to "Git Clone (for contributors)" tab

docs/content/users/quickstart.md
rpkoller: > Good point, but i think that details belongs to OM's documentation, not DDEV's. Imho, DDEV quickstart should contain the minimum instructions to start a new project, When using git clone you normaly know about branches or it should be in OM docs.

problem is you cant expect that people necessarily go to the openmage website to research what the details and caveats of the given approaches are. it is usually preferable to enable the user to make an informed decision in the current context without the need to branch out and do their own research. and i think in this case it wouldnt do any harm to elaborate on that detail in quickstart of ddev as well as in the OM docs. by having that detail in the ddev docs as well it would make it easier for the user to decide which approach to take. since each approach is copy and paste based you dont necessarily have to know the underlaying details. so having a brief disclaimer pointing out that caveat might be helpful. 

from my point of view the idea of quickstarts is to be brief and actionable on one hand (so that you are able to install a new CMS you are not familiar within seconds) but at the same time to provide some important educational tidbits along the way necessary in the given context.

sreichel: There is no need to inform about other branches for git clone installation. `v19` branch is outdated at becomes EOL with April 4th. `v21`/`next` branch is far behind and not maintained. There is only the main branch, that should be used. 

Instead of adding more information here, https://docs.openmage.org/users/install/use-composer/#specify-php-engine-version has to be updated.

stasadev:
I want to merge this, so we can have a working Composer quickstart for OpenMage in the stable documentation (we have a release in the upcoming days.)

@rpkoller any comments?

sreichel:
@rpkoller anything i can do here? I'd leave as it is and update as soon there is a composer starter project (and updated docs).

rfay:
@sreichel one workflow comment - When you update the branch in a PR, could you use the "rebase" technique when possible? When you use the merge commit technique it prevents it from being rebased later. (I know GitHub offers the merge commit technique as default.) 
image

rfay:
Thanks so much for all the effort on behalf of OpenMage! We'll look forward to future maintenance as things progress!

sreichel:
@rfay thanks. 

I'll use rebase from now :)

PR #7134, opened 2025-03-22
docs: added version info for `mkdocs-material`

sreichel:
## The Issue

My mkdocs-material version was at 9.4.8, so `make mkdocs-serve` did not work.

docs:

https://ddev--7134.org.readthedocs.build/en/7134/developers/testing-docs/#preview-changes

### DIFF
--- a/docs/content/developers/testing-docs.md
+++ b/docs/content/developers/testing-docs.md
@@ -60,6 +60,13 @@ This will launch a web server on port 8000 and automatically refresh pages as th
     sudo --preserve-env pipx install mkdocs --pip-args "-r docs/mkdocs-pip-requirements"
     ```
 
+!!!warning "`mkdocs-material` v9.5.1 required!"
+    On lower versions of `mkdocs-material`, you may see an error like this:
+
+    ```bash
+    Config value 'plugins': The "privacy" plugin is not installed
+    ```
+
 ## Check Markdown for Errors
 
 Run `make markdownlint` before you publish changes to quickly check your files for errors or inconsistencies.


rfay (changes requested):
I think this means that you didn't do the required pip install -r , true?

sreichel:
True, b/c mkdocs was already installed, It only says "Installing mkdocs locally", not that is mandatory to use `-r` flag.

Question: pip-requirements has `mkdocs-material>=9` ... not `>=9.5.1`. Does it make a difference?

... maybe its still helpful for those that dont RTFM carefully. :)

rfay:
I recently did this without paying attention to the [docs](https://ddev.readthedocs.io/en/stable/developers/testing-docs/#preview-changes) and ended up using `--break-system-packages`, which was annoying. 

But please say if these docs (which use pipx and are pretty strong) would have been adequate:

image

stasadev:
I use `direnv` to install everything to the isolated env https://github.com/ddev/ddev/blob/main/.envrc

rfay:
Moved to draft while we are discussing this.

sreichel:
@rfay no need to keep it open. Its not that important to me.

I run into serveral errors when blinly starting `mkdos serve`.

1st, "go"-command was missing, but that was obious to fix. 2nd was missing mkdocs-plugin, that took me a few minutes to find 9.5.1 is required.

rfay:
I'll close, but happy to help. You have to install the extensions in mkdocs-pip-requirements as shown above. Let me know if you need any more help.

PR #7069, opened 2025-03-12
feat: Integrate XHGui into DDEV, fixes #6894

rfay:
## The Issue

- #6894 

Honors to @tyler36, who pioneered and then matured the xhgui add-on until it could be brought into DDEV here.

## How This PR Solves The Issue

Add integration for xhgui

## TODO

- [x] Handle case of existing ddev-xhgui add-on
- [x] Tests
- [x] Fix `ddev xhgui -j`, perhaps add some support
- [x] Make the xhgui port configurable
- [x] Make work without router like mailpit
- [x] FIgure out if `ddev xhgui` should launch browser. Adjust as needed
- [ ] Report to TYPO3 folks (for typo3.org) about the outcome at [link](https://typo3.org/community/teams/communication-coordination/content-request), [using](https://typo3.org/article/community-budget-idea-report-image-rendering-improvements) [these](https://typo3.org/article/community-budget-report-future-of-the-staticfilecache-extension) [examples](https://typo3.org/project/news/budget-idea).

## Manual Testing Instructions

Docs suggestions:
*  https://ddev--7069.org.readthedocs.build/en/7069/users/debugging-profiling/xhprof-profiling/#simplest-xhprof-usage-with-xhgui
*  https://ddev--7069.org.readthedocs.build/en/7069/users/usage/commands/#xhgui
* https://ddev--7069.org.readthedocs.build/en/7069/users/configuration/config/#xhprof_mode

- [ ] `ddev config global --xhprof-mode=xhgui`
- [ ] `ddev xhgui on`
- [ ] `ddev xhgui status`
- [ ] `ddev xhgui launch`
- [ ] `ddev xhprof on`
- [ ] Validate that xhprof still works with xhprof-mode=prepend
- [ ] Install ddev/ddev-xhgui and then `ddev start` to see removal
- [x] Test with a Postgres project

## Automated Testing Overview

- [x] TestDdevXhprofXhguiEnabled
- [x] TestCmdXHGui
- [x] TestXHGuiSetup

## Release/Deployment Notes

- [ ] Note that the ddev-xhgui add-on will not be compatible.
- [ ] Announce to supporting TYPO3 community!
- [ ] Blog and screencast

PR #7142, opened 2025-03-25
fix: remove XHGui volume for `/run/nginx`, for #7069

stasadev:
## The Issue

- #7069
- https://github.com/perftools/xhgui/issues/505

XHGui creates a volume in the [Dockerfile](https://github.com/perftools/xhgui/blob/cec362839f001a7f2f7f681abbcadbe0a26a400b/Dockerfile#L108):

```dockerfile
VOLUME "/run/nginx"
```

Related:

- https://github.com/moby/buildkit/issues/4802#issuecomment-2182544924
- https://github.com/moby/moby/issues/3465#issuecomment-2179958180

Overrides the parent Dockerfile using this technique:

- https://stackoverflow.com/questions/44020785/remove-a-volume-in-a-dockerfile/78645904

```
# check and remove all xhgui and anonymous volumes
docker volume ls

ddev start
ddev xhgui

# check it again, it should not create any new volumes
docker volume ls
```

### DIFF
--- a/containers/ddev-xhgui/Dockerfile
+++ b/containers/ddev-xhgui/Dockerfile
@@ -1,4 +1,20 @@
-FROM xhgui/xhgui:0.21 AS ddev-xhgui
+FROM xhgui/xhgui:0.21 AS xhgui-docker
+
+FROM alpine:3.15 AS ddev-xhgui
+
+### --------------------------------xhgui-docker----------------------------------
+### XHGui Docker image creates a volume for /run/nginx, which we want to remove.
+### We copy all files from xhgui/xhgui, and set ENV, WORKDIR, CMD, EXPOSE.
+### Source https://github.com/perftools/xhgui/blob/0.21.x/Dockerfile
+COPY --from=xhgui-docker / /
+ENV PHP_INI_DIR=/etc/php7
+ARG APPDIR=/var/www/xhgui
+ARG WEBROOT=$APPDIR/webroot
+WORKDIR $APPDIR
+RUN mkdir -p /run/nginx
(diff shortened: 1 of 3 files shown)


rfay (commented):
This is a great idea, congrats and thanks. 

I may not be able to look carefully, but when you like it it's good to go in.

containers/ddev-xhgui/Dockerfile
```diff
@@ -1,4 +1,20 @@
-FROM xhgui/xhgui:0.21 AS ddev-xhgui
+FROM xhgui/xhgui:0.21 AS xhgui-docker
+
+FROM alpine:3.15 AS ddev-xhgui
```
glensc, line 3: you should build `FROM scratch` then if you copy other image all files to avoid superfluous files or even cause conflicts because different os files don't merge well.

stasadev: Thank you.

containers/ddev-xhgui/Dockerfile
```diff
@@ -1,4 +1,20 @@
-FROM xhgui/xhgui:0.21 AS ddev-xhgui
+FROM xhgui/xhgui:0.21 AS xhgui-docker
+
+FROM scratch AS ddev-xhgui
```
rfay, line 3: I wasn't suggesting to re-do this :) The other way was fine too :)

stasadev: We don't need to edit this line in the future with `scratch` (when `xhgui/xhgui` changes the base image).

PR #7147, opened 2025-03-25
fix: don't auto show release notes in windows installer, for #7049

rfay:
## The Issue

- #7049 

We're trying a new install script in 7049, which will run the installer

However, it has always by default shown the release notes, which would leave people unable to figure out context when browser opens.

Clear the auto-set checkbox

Run the windows installer. On the last page, it should not automatically launch the browser.


rfay:
I don't think you'll object to this @stasadev , please pull when you've looked.

stasadev (approved):
I confirm, there is no checkbox by default:

image

PR #7151, opened 2025-03-26
feat: remove all images in `ddev delete images`

stasadev:
## The Issue

Not everything is removed.

Makes sure that we remove all the stuff in `ddev delete images`
TODO check `ddev clean`

- TODO


rpkoller:
ideally from my perspective `ddev delete images` should keep the ddev core images for the current version plus the addon images for current projects, but remove all the ddev core images of previous versions, and dangling addon images for projects that got already deleted

rfay:
I imagine you'll want to solve
* https://github.com/ddev/ddev/issues/7201

in this one.

stasadev:
Ready for review, the OP is updated.

I didn't add any new automated tests because these commands can't be tested with just one `ddev` binary.

stasadev:
> ideally from my perspective `ddev delete images` should keep the ddev core images for the current version plus the addon images for current projects, but remove all the ddev core images of previous versions, and dangling addon images for projects that got already deleted

@rpkoller, `ddev delete images` without any flags should now work exactly as you expect.

rfay:
I used `ddev delete images` the other day doing some cleanup on my server. It made me sad that it insisted on `ddev poweroff`, but that's a normal and reasonable requirement. But I did wish it didn't do that and (with permission) stop all my projects on there. It seems like in a perfect world it would be possible to clean up without doing that. IMO doesn't matter for local development, which is DDEV's target.

rpkoller (commented):
I've tested the artifact. kept all my backlog of images from plugins as well as ddev core for a while: https://gist.github.com/rpkoller/762009ebf6e004c15121f2525471fea5 

i love that none images get cleared now as well. i also had some confusion understanding the exact naming pattern. stas already cleared up that confusion and i can attest the cleanup worked very well in my test.  the only nitpick to note is that the output you have to confirm when running `ddev delete images` is sort of overwhelming : 

```
$> ddev delete images
Warning: the following 7 Docker image(s) will be deleted:
Image to be deleted from ddev-core: sha256:6ef8b6687bb46f4eb5961e73f5ae0dde3325368f0c431cad63a1c2e27c78494f
Image to be deleted from ddev-core: ddev/ddev-webserver:20250213_rfay_mariadb_11.8-core-built
Image to be deleted from ddev-lu: ddev-lu-xhgui:latest
Image to be deleted from ddev-uxmeeting: ddev/ddev-webserver:20250213_rfay_mariadb_11.8-uxmeeting-built
Image to be deleted from ddev-vali: ddev-vali-xhgui:latest
Image to be deleted: ddev/ddev-traefik-router:v1.24.6
Image to be deleted: ddev/ddev-webserver:20250213_rfay_mariadb_11.8
Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
Powering off DDEV to avoid conflicts
The ddev-ssh-agent container has been removed. When you start it again you will have to use 'ddev auth ssh' to provide key authentication again.
All DDEV images discovered were deleted.
```

i am not sure if i really need the list of images to be deleted. it is a non destructive operation that could cause a data loss. you are only deleting images that could be redownloaded anyway. i guess if something happens and i redownload an image then i might investigate. but that would be just inconvenient for the moment. therefore i would think that list is not really needed imho. and if folks think it would be nice to have then i would at least change the sorting. if you run `docker image ls` you have a sort order from newst to oldest while the sort order on the confirmation output is alphabetical ascending. i would keep the order consistent. or the longer i think about it, that list is just complicating things, i would simply drop it.

stasadev:
> But I did wish it didn't do that and (with permission) stop all my projects on there.

@rfay, I want to use `ddev delete images` myself in some daily `update` script I use (which runs many more things).
It would be nice to check if the images we want to remove are running or not. And don't do `ddev poweroff` when it's not needed.

> if you run docker image ls you have a sort order from newst to oldest while the sort order on the confirmation output is alphabetical ascending.

@rpkoller, it's grouped by project, then by name. I tried different sortings, and I'm not sure if newest to oldest will be readable at all.
Maybe it's better to use table here (like in `ddev list`) so it can be more readable.

> it is a non destructive operation that could cause a data loss.

We consider "data loss" only for volumes, not images.

rpkoller:
thought about it some more. in regard of readability probably the order might now be the most important aspect. it is rather the redundant block of text: 

```
$> ddev delete images
Warning: the following 7 Docker image(s) will be deleted:
Image to be deleted from ddev-core: sha256:6ef8b6687bb46f4eb5961e73f5ae0dde3325368f0c431cad63a1c2e27c78494f
Image to be deleted from ddev-core: ddev/ddev-webserver:20250213_rfay_mariadb_11.8-core-built
Image to be deleted from ddev-lu: ddev-lu-xhgui:latest
Image to be deleted from ddev-uxmeeting: ddev/ddev-webserver:20250213_rfay_mariadb_11.8-uxmeeting-built
Image to be deleted from ddev-vali: ddev-vali-xhgui:latest
Image to be deleted: ddev/ddev-traefik-router:v1.24.6
Image to be deleted: ddev/ddev-webserver:20250213_rfay_mariadb_11.8
Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
Powering off DDEV to avoid conflicts
The ddev-ssh-agent container has been removed. When you start it again you will have to use 'ddev auth ssh' to provide key authentication again.
All DDEV images discovered were deleted.
```

each line for an image is starting with `Image to be deleted from` that way you have a hard time processing, you have to skim past that redundant snippet in each line. in addition to that the context like ddev-core is hard to distinguish from the actual image name. how about something like? 

```
$> ddev delete images
Warning: the following 7 Docker image(s) will be deleted:
(ddev-core): sha256:6ef8b6687bb46f4eb5961e73f5ae0dde3325368f0c431cad63a1c2e27c78494f
(ddev-core): ddev/ddev-webserver:20250213_rfay_mariadb_11.8-core-built
(ddev-lu): ddev-lu-xhgui:latest
(ddev-uxmeeting): ddev/ddev-webserver:20250213_rfay_mariadb_11.8-uxmeeting-built
Iddev-vali): ddev-vali-xhgui:latest
ddev/ddev-traefik-router:v1.24.6
ddev/ddev-webserver:20250213_rfay_mariadb_11.8
Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
Powering off DDEV to avoid conflicts
The ddev-ssh-agent container has been removed. When you start it again you will have to use 'ddev auth ssh' to provide key authentication again.
All DDEV images discovered were deleted.
```

but looking at the inconsistent and rather confusing context with some images having ddev-core some a name related to project (ddev-uxmeeting), some at the end without any context with just the name. i wonder if the following would be the most straight forward ( i would also rename the sha256 to what it is called with `docker image ls` -> `<none>`


```
$> ddev delete images
Warning: the following 7 Docker image(s) will be deleted:

<none>
ddev/ddev-webserver:20250213_rfay_mariadb_11.8-core-built
ddev-lu-xhgui:latest
ddev/ddev-webserver:20250213_rfay_mariadb_11.8-uxmeeting-built
ddev-vali-xhgui:latest
ddev/ddev-traefik-router:v1.24.6
ddev/ddev-webserver:20250213_rfay_mariadb_11.8

Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
Powering off DDEV to avoid conflicts
The ddev-ssh-agent container has been removed. When you start it again you will have to use 'ddev auth ssh' to provide key authentication again.
All DDEV images discovered were deleted.
```
or as i said drop the listing of the image names entirely? the dont provide much of an additional benefit?

stasadev:
> But I did wish it didn't do that and (with permission) stop all my projects on there.

@rfay, done, `ddev poweroff` will only run if some container uses any of the images to delete.

> but looking at the inconsistent and rather confusing context with some images having ddev-core some a name related to project (ddev-uxmeeting), some at the end without any context with just the name. i wonder if the following would be the most straight forward ( i would also rename the sha256 to what it is called with `docker image ls` -> `<none>`

@rpkoller, sort by name only, without any projects, show short image IDs:

```
$ ddev delete images    
Warning: the following 5 Docker image(s) will be deleted:
IMAGE ID       REPOSITORY:TAG
64d94408a746   <none>:<none>
b5e47fc0c220   ddev/ddev-dbserver-mariadb-10.11:v1.24.6
a9574de44849   ddev/ddev-ssh-agent:v1.24.6
3e84b6b15ff1   ddev/ddev-traefik-router:v1.24.6
e887d55b5c2c   ddev/ddev-webserver:v1.24.6
Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
All DDEV images discovered were deleted.
Optionally, run `docker builder prune` to clean unused builder cache.
```

> or as i said drop the listing of the image names entirely? the dont provide much of an additional benefit?

I still think it's nice to see a list of what changes are being made here.

rpkoller:
interesting. when you've mentioned display the images in a table i've automatically thought it would mean the table would have borders (which is thought would look and feel too massive), but without any borders it improves the readability significantly. i like that. the sole two nitpicks.

would it make sense to add a blank line before and after? that way it makes easier to distinguish the textblocks:

```
$ ddev delete images    
Warning: the following 5 Docker image(s) will be deleted:

IMAGE ID              REPOSITORY:TAG
64d94408a746   <none>:<none>
b5e47fc0c220    ddev/ddev-dbserver-mariadb-10.11:v1.24.6
a9574de44849   ddev/ddev-ssh-agent:v1.24.6
3e84b6b15ff1     ddev/ddev-traefik-router:v1.24.6
e887d55b5c2c   ddev/ddev-webserver:v1.24.6

Deleting images is a non-destructive operation.
You may need to download images again when you need them.

OK to continue? [Y/n] (yes): 
All DDEV images discovered were deleted.
Optionally, run `docker builder prune` to clean unused builder cache.
```

and the other detail, i am not sure what benefit the image id has on that deletion step. the ID is only needed if you manually want to trigger the deletion for an image with `docker rmi`. but in that list it doesnt have any value? at least me the image id tells me nothing about the image. would it make sense to drop that column entirely or replace it with a category column. categories could be coarse grained, just something like: ddev core, ddev addon, docker ... that way one gets a rough context this line/image belongs to core that one to an addon and that one is just a docker image like for example none? meaning having a rough categorization to what an image belongs to i would consider helpful distinguishing the list.

stasadev:
> would it make sense to add a blank line before and after?

Okay.

> i am not sure what benefit the image id has on that deletion step

The `IMAGE ID` column is the primary identifier.

One use case: run `ddev delete images`, don't agree to remove anything, read the image list, notice a suspicious `<none>:<none>` or another image, copy the ID, and check its details with `docker inspect <image-id>` in another terminal window.

> would it make sense to drop that column entirely or replace it with a category column. categories could be coarse grained, just something like: ddev core, ddev addon, docker ...

This will only add more complexity to the code, it should already be clear what kind of image it is, as long as it's not `<none>:<none>`.

rfay:
I see
> Optionally, run `docker builder prune` to clean unused builder cache

Shouldn't it be `docker buildx prune` ?

rfay:
Trying to save away my 182 images before testing. It's probably not worth the effort, because the restore will be too costly.

`docker save $(docker images -q) | gzip > /tmp/all-images.tar.gz`

A better approach would probably be a copy of (Orbstack) in this case images directory. That's also probably too costly.

stasadev:
> Shouldn't it be `docker buildx prune` ?

No, because `buildx` is specific to Buildx, while `builder` runs any command for the builder *in use*.

rfay:
I'm sure you're right, but on my server I get this:

```
rfay@ddevprod:~$ docker builder prune
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

WARNING! This will remove all dangling build cache. Are you sure you want to continue? [y/N]
```

I assume it's complaining because I'm using the legacy builder, not because I'm using `docker builder`.  (I see the warning when I just do `docker builder` as well. 

I do not see the warning on macOS where I'm using buildx:

```
rfay@rfay-mba-m4:~/workspace/ddev$ docker builder prune
WARNING! This will remove all dangling build cache. Are you sure you want to continue? [y/N]
```

So I'm sure you're right and it's just poor messaging from the docker CLI.

stasadev:
I also have 'docker-buildx' installed and `docker builder prune` doesn't complain to me.

rfay:
This is not a request for changes or extra investigation, but I do see one or two suspicious images after `ddev delete images -a` (edited list)

```
REPOSITORY                                                                           TAG                      IMAGE ID       CREATED         SIZE
ddev/ddev-webserver                                                                  <none>                   be381812ac90   5 days ago      1.21GB
vsc-ddev-45e804f67f32a6616b4a214fafc6d4bc3e9be3135f5df8fbd6edd62e85c1eb25-features   latest                   8b21a7582369   3 weeks ago     3.3GB
opensearchproject/opensearch-dashboards                                              3.0.0                    a649ef48e6ff   6 weeks ago     1.64GB
redis                                                                                7                        11457eb59fe5   7 weeks ago     147MB
adminer                                                                              standalone               ed7c8a62c220   2 months ago    118MB
phpstorm_helpers                                                                     PS-243.25659.45          2dc29e561d7c   2 months ago    4.2MB
adminer                                                                              5.1.0-standalone         c313cf8b0ea9   2 months ago    118MB
opensearchproject/opensearch-dashboards                                              latest                   2b8f3a2f88a8   3 months ago    1.46GB
postgres                                                                             16                       ae29dcde45c8   3 months ago    457MB
bitnami/mariadb                                                                      10.6.18-debian-12-r0     962ca1adea62   13 months ago   413MB
busybox                                                                              stable                   0496d33b146b   2 years ago     4.04MB
phpmyadmin                                                                           5.2.0                    cdc96bbdba0b   2 years ago     479MB
traefik                                                                              2.2.0                    269e148e952b   5 years ago     72.3MB
```

rfay (changes requested):
Thanks for the great work on this. 

`ddev clean` in a project context seems to delete the ~/.ddev/bin contents, which breaks any other project that might be running.  And it even breaks the running project.

I think `ddev clean` should stop the running project, and shouldn't delete ~/.ddev/bin unless `ddev clean -a`

stasadev:
> Should `ddev clean -a` delete the project_list.yaml and/or global_config.yaml ? (I'd probably think it's better to leave them alone.)

I also believe that it should not touch these files.

> I think `ddev clean` should stop the running project, and shouldn't delete ~/.ddev/bin unless `ddev clean -a`

Agreed, will change the logic.

> (It also doesn't seem to warn about the deletion of ~/.ddev/bin in any case.)

Good point.

rfay:
It does look like we might be leaving busybox out there, because we use it without touching it? Here's after `ddev clean -a` on my Docker Desktop provider. I'm not convinced it's necessary that we delete busybox.
```
rfay@rfay-mba-m4:~/workspace/ddev$ docker images
REPOSITORY         TAG                IMAGE ID       CREATED        SIZE
redis              7                  7df1eeff67eb   7 weeks ago    195MB
phpstorm_helpers   PS-251.23774.436   df3c5c109475   2 months ago   6.3MB
<none>             <none>             01cf6854e23d   3 months ago   106MB
postgres           16                 cef2d22004db   3 months ago   636MB
alpine             latest             a8560b36e8b8   4 months ago   12.8MB
busybox            latest             37f7b378a29c   8 months ago   6.02MB
busybox            stable             7a5342b7662d   2 years ago    6.02MB
```

rfay (approved):
Overall it seems great to me. Minor suggestions in previous reviews and comments, that can be handled as you see fit. But I sure like it, and it doesn't seem to have any risk factors either.

rfay:
Tested this on all my other docker providers, colima, lima, rancher, Docker Desktop. In each case it did marvelous cleanup things, just as I'd hope.

rfay:
This fixes 
* https://github.com/ddev/ddev/issues/5073

also, true?

stasadev:
All requested changes have been implemented.

stasadev:
I tested it again, and `postgres:x-y-built` images are always removed.

I'll try to fix it.

Edit: fixed.

---

> It does look like we might be leaving busybox out there, because we use it without touching it? Here's after `ddev clean -a` on my Docker Desktop provider. I'm not convinced it's necessary that we delete busybox.

What I've noticed over time: we use `busybox:stable` and JetBrains uses `busybox:latest` 🙂

---

Another thing I noticed: if the image doesn't have a specific `ddev` version in its tag, it won't be rebuilt later.

So if users update `ddev` and use something like `ddev/ddev-solr`, the already built Solr image won't be rebuilt with the new `ddev`.

This could be solved by always setting the `image:` field, but that's outside the scope of this PR.

rpkoller:
> This fixes
> 
> * [`<none>` image is being created when a new and different image of same name is created #5073](https://github.com/ddev/ddev/issues/5073)
> 
> also, true?

not completely. the none images are still being created in the first place. the cleanup got now way more convenient with this PR.

rfay:
The `none` images are a feature of `docker buildx` and this is going to be as close as we get to fixing IMO.

PR #7152, opened 2025-03-27
fix: show correct ephemeral router ports in `ddev describe`, make generic webserver work with ephemeral ports

stasadev:
## The Issue

From https://discord.com/channels/664580571770388500/1352433228966789150

`ddev describe` shows incorrect project URL, without the correct ephemeral port.

Caused by changes to `ddevapp.go` in:

- #6935

To reproduce:

```
$ ddev poweroff

# leave it running
$ docker run -it --rm -p 80:80 -p 443:443 busybox sh

# in another terminal window
$ ddev start
Starting d11... 
Port 80 is busy, using 33000 instead, see https://ddev.com/s/port-conflict 
Port 443 is busy, using 33001 instead, see https://ddev.com/s/port-conflict
...
Your project can be reached at https://d11.ddev.site:33001

# describe doesn't show https://d11.ddev.site:33001 and http://d11.ddev.site:33000
$ ddev describe
...
├──────────────┼──────┼─────────────────────────────────────────────────┤
│ Project URLs │      │ https://127.0.0.1:33151, http://127.0.0.1:33150 │
└──────────────┴──────┴─────────────────────────────────────────────────┘

$ ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.httpURLs'
[
  "http://127.0.0.1:33150"
]

$ ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.httpsURLs'
[
  "https://127.0.0.1:33151"
]
```

Additionally:

I noticed that `generic` webserver tries to use ephemeral ports, but fails:

Using svetlekit quickstart https://ddev.readthedocs.io/en/stable/users/quickstart/#nodejs

```
$ ddev poweroff

# leave it running
$ docker run -it --rm -p 80:80 -p 443:443 busybox sh

$ ddev start
Starting my-sveltekit-site... 
Port 80 is busy, using 33000 instead, see https://ddev.com/s/port-conflict 
Port 443 is busy, using 33001 instead, see https://ddev.com/s/port-conflict 
Building project images....
Project images built in 2s. 
 Network ddev-my-sveltekit-site_default  Created 
 Container ddev-my-sveltekit-site-web  Created 
 Container ddev-my-sveltekit-site-db  Created 
 Container ddev-my-sveltekit-site-db  Started 
 Container ddev-my-sveltekit-site-web  Started 
Starting web_extra_daemons... 
Failed to start my-sveltekit-site: unable to listen on required ports, port 443 is already in use,
Troubleshooting suggestions at https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#unable-listen
```

When there is nothing to expose, and default ports are busy, DDEV says that it tries to use router ports (which is not true):

```
$ ddev poweroff

# leave it running
$ docker run -it --rm -p 80:80 -p 443:443 busybox sh

$ mkdir generic-no-ports && cd generic-no-ports
$ ddev config --webserver-type=generic
$ ddev start
Starting generic-no-ports...
Port 80 is busy, using 33000 instead, see https://ddev.com/s/port-conflict 
Port 443 is busy, using 33001 instead, see https://ddev.com/s/port-conflict
...

# why say "port is busy", if nothing is running there
```

- Refactors `GetPrimaryRouterHTTPPort` and `GetPrimaryRouterHTTPSPort` to use correct values.
- Makes generic webserver work when default ports (80, 443) are busy.
- Show correct values in `ddev describe` for usual webserver and `generic` one.

Repeat my three examples:

1. describe should show `https://d11.ddev.site:33001` and `http://d11.ddev.site:33000`
2. svetlekit quickstart show start on ephemeral ports, and `ddev describe` should show correct URLs
3. generic webserver without ports shouldn't say "Port X is busy, using Y"

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -558,9 +558,12 @@ func (app *DdevApp) GetWebserverType() string {
 	return v
 }
 
-// GetPrimaryRouterHTTPPort returns app's router primary http port
-// This can be port 80 (default) or the item in HTTP_EXPOSE
-// matching the default router port. If there is no port, returns ""
+// GetPrimaryRouterHTTPPort returns app's primary router http port
+// It has to choose from (highest to lowest priority):
+// 1. Empty string if webserver type is generic and no web_extra_exposed_ports are defined
+// 2. The actual port configured into running container via DDEV_ROUTER_HTTP_PORT
+// 3. The project router_http_port
+// 4. The global router_http_port
 func (app *DdevApp) GetPrimaryRouterHTTPPort() string {
 	proposedPrimaryRouterHTTPPort := "80"
 	if globalconfig.DdevGlobalConfig.RouterHTTPPort != "" {
@@ -569,18 +572,16 @@ func (app *DdevApp) GetPrimaryRouterHTTPPort() string {
 	if app.RouterHTTPPort != "" {
 		proposedPrimaryRouterHTTPPort = app.RouterHTTPPort
 	}
-	if httpExpose := app.GetWebEnvVar("HTTP_EXPOSE"); httpExpose != "" {
-		//util.Debug("GetPrimaryRouterHTTPPort(): HTTP_EXPOSE='%s'", httpExpose)
-		httpPort := app.RouterPortFromExposeVariable(httpExpose, proposedPrimaryRouterHTTPPort)
-		if httpPort == "" {
-			//util.Debug("GetPrimaryRouterHTTPPort(): no primary router port found in %s", httpExpose)
-			proposedPrimaryRouterHTTPPort = ""
-		}
+	if httpPort := app.GetWebEnvVar("DDEV_ROUTER_HTTP_PORT"); httpPort != "" {
+		proposedPrimaryRouterHTTPPort = httpPort
+	}
+	if app.WebserverType == nodeps.WebserverGeneric && len(app.WebExtraExposedPorts) == 0 {
+		proposedPrimaryRouterHTTPPort = ""
 	}
 	return proposedPrimaryRouterHTTPPort
 }
 
-// GetWebEnvVar() gets an environment variable from
+// GetWebEnvVar gets an environment variable from
 // app.ComposeYaml["services"]["web"]["environment"]
 // It returns empty string if there is no var or the ComposeYaml
 // is just not set.
@@ -595,8 +596,8 @@ func (app *DdevApp) GetWebEnvVar(name string) string {
 	return ""
 }
 
-// TargetPortFromExposeVariable() uses a string like HTTP_EXPOSE or HTTPS_EXPOSE, which is a
-// comma-delimted list of colon-delimited port-pairs
+// TargetPortFromExposeVariable uses a string like HTTP_EXPOSE or HTTPS_EXPOSE, which is a
+// comma-delimited list of colon-delimited port-pairs
 // Given a target port (often "80" or "8025") its job is to get from HTTPS_EXPOSE or HTTP_EXPOSE
 // the related port to be exposed on the router.
 // It returns an empty string if the HTTP_EXPOSE/HTTPS_EXPOSE is not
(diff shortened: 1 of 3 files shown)


deviantintegral (approved):
Manual testing and static review looks good to me. See my other comment for the exact steps I took for testing.

I also wanted to use this as an opportunity try out generating unit tests. I was able to generate this one for the router which looks OK (though I had to fix a mistake as it generated with `ddevapp.ExposedPort` instead of `ddevapp.WebExposedPort`. However, I was unable to try it for the app file as `ddevapp_test.go` is too big! I'm unlikely to have time to write the tests by hand today.

https://youtrack.jetbrains.com/issues?q=%22Request%20entity%20is%20larger%20than%20defined%20limits%22&preview=LLM-14961

Feel free to take this or not.

```go
// TestAssignRouterPortsToGenericWebserverPorts ensures that RouterHTTPPort and RouterHTTPSPort
// are assigned correctly based on WebExtraExposedPorts for Generic webservers.
func TestAssignRouterPortsToGenericWebserverPorts(t *testing.T) {
	assert := asrt.New(t)
	type testCase struct {
		name                 string
		webserverType        string
		webExtraExposedPorts []ddevapp.WebExposedPort
		expectedHTTPPort     string
		expectedHTTPSPort    string
	}

	tests := []testCase{
		{
			name:          "Generic webserver with valid ports",
			webserverType: nodeps.WebserverGeneric,
			webExtraExposedPorts: []ddevapp.WebExposedPort{
				{HTTPPort: 8080, HTTPSPort: 8443},
			},
			expectedHTTPPort:  "8080",
			expectedHTTPSPort: "8443",
		},
		{
			name:                 "Generic webserver with no extra ports",
			webserverType:        nodeps.WebserverGeneric,
			webExtraExposedPorts: []ddevapp.WebExposedPort{},
			expectedHTTPPort:     "",
			expectedHTTPSPort:    "",
		},
		{
			name:          "Non-Generic webserver should not assign ports",
			webserverType: "other-webserver",
			webExtraExposedPorts: []ddevapp.WebExposedPort{
				{HTTPPort: 8080, HTTPSPort: 8443},
			},
			expectedHTTPPort:  "",
			expectedHTTPSPort: "",
		},
		{
			name:          "Generic webserver with multiple ports uses first",
			webserverType: nodeps.WebserverGeneric,
			webExtraExposedPorts: []ddevapp.WebExposedPort{
				{HTTPPort: 8000, HTTPSPort: 8443},
				{HTTPPort: 8081, HTTPSPort: 8444},
			},
			expectedHTTPPort:  "8000",
			expectedHTTPSPort: "8443",
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			app := &ddevapp.DdevApp{
				WebserverType:        tc.webserverType,
				WebExtraExposedPorts: tc.webExtraExposedPorts,
			}

			ddevapp.AssignRouterPortsToGenericWebserverPorts(app)
			assert.Equal(tc.expectedHTTPPort, app.RouterHTTPPort)
			assert.Equal(tc.expectedHTTPSPort, app.RouterHTTPSPort)
		})
	}
}
```

stasadev:
> I also wanted to use this as an opportunity try out generating unit tests.

@deviantintegral, thanks, I added it with commit credit.

> The current TestUseEphemeralPorts just doesn't check this well enough (at all?), so it's fundamental that we get at least app.Describe() checked for this, and perhaps even a TestCmd test.

@rfay, I found out why it didn't catch anything: we need to have a fresh app after `ddev start`, otherwise it works with ports defined manually in the test.

> perhaps even a TestCmd test.

Not needed, because I added the necessary checks to the existing `TestUseEphemeralPorts` and confirmed that the test fails on the current DDEV HEAD.

pkg/ddevapp/ddevapp.go
```diff
@@ -569,18 +572,16 @@ func (app *DdevApp) GetPrimaryRouterHTTPPort() string {
...
 		proposedPrimaryRouterHTTPPort = app.RouterHTTPPort
 	}
-	if httpExpose := app.GetWebEnvVar("HTTP_EXPOSE"); httpExpose != "" {
-		//util.Debug("GetPrimaryRouterHTTPPort(): HTTP_EXPOSE='%s'", httpExpose)
-		httpPort := app.RouterPortFromExposeVariable(httpExpose, proposedPrimaryRouterHTTPPort)
-		if httpPort == "" {
-			//util.Debug("GetPrimaryRouterHTTPPort(): no primary router port found in %s", httpExpose)
-			proposedPrimaryRouterHTTPPort = ""
-		}
+	if httpPort := app.GetWebEnvVar("DDEV_ROUTER_HTTP_PORT"); httpPort != "" {
```
rfay, line 575: Why use this instead of app.GetPrimaryRouterHTTPPort() ? It seems like we're going in circles on that.

stasadev: I don't understand.
We are inside the `app.GetPrimaryRouterHTTPPort()`, if I call it here, it will be an endless loop.

stasadev: This variable `DDEV_ROUTER_HTTP_PORT` contains the used real port (either from config or ephemeral).

rfay: Right, but DDEV_ROUTER_HTTP_PORT is set to the value of app.GetPrimaryRouterHTTPPort(), so that what I mean by circular.

stasadev: Now I understand what you mean, `DDEV_ROUTER_HTTP_PORT` is set before `app.ComposeYaml` is populated (`GetWebEnvVar` reads `app.ComposeYaml`), so there is no issue.

This has no difference with what we use here before , `HTTP_EXPOSE` was handled in the same way. I just found that we don't need to read `HTTP_EXPOSE`, if it has it's own variable.

rfay (changes requested):
I'm sorry you've been waiting for feedback on this. I've spent some time on it but I get distracted and a bit confused by the two different problems in play. I know that you discovered the generic webserver problem while studying the `ddev describe` problem, but it doesn't seem related, and so it's hard to properly discuss this. I think I have some comments on the generic webserver problem, (which is probably very little used) but really want to get the describe problem sorted out soon.

Would you be willing to split this into two PRs if there's not a reason that it can't be split?

stasadev:
> Would you be willing to split this into two PRs if there's not a reason that it can't be split?

Done.

PR #7160, opened 2025-03-28
docs: update Craft CMS quickstart, for #7107

stasadev:
## The Issue

I didn't review this before it was merged:

- #7107 

Related:

- #6246
- https://dev.to/mandrasch/install-craft-cms-v5-alpha-with-one-command-via-ddev-4ne5

- Makes the quickstart for Craft CMS non-interactive.
- Removes not needed comments from Craft CMS quickstart.
- Removes not needed `-y` flag.

https://ddev--7160.org.readthedocs.build/en/7160/users/quickstart/#craft-cms

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -628,7 +628,7 @@ export MAGENTO_HOSTNAME=my-magento2-site
 mkdir ${MAGENTO_HOSTNAME} && cd ${MAGENTO_HOSTNAME}
 ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
 ddev add-on get ddev/ddev-elasticsearch
-ddev start -y
+ddev start
 ddev composer create --repository https://repo.magento.com/ magento/project-community-edition
 rm -f app/etc/env.php
 
@@ -866,7 +866,7 @@ If you have any questions there is lots of help in the [DDEV thread in the Proce
     ```bash
     mkdir my-shopware-site && cd my-shopware-site
     ddev config --project-type=shopware6 --docroot=public
-    ddev start -y
+    ddev start
     ddev composer create shopware/production:^v6.5
     # If it asks `Do you want to include Docker configuration from recipes?`
     # answer `x`, as we're using DDEV for this rather than its recipes.
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -172,14 +172,16 @@ Environment variables will be automatically added to your `.env` file to simplif
 
     # Boot the project and install the starter project:
     ddev start
-    ddev composer create craftcms/craft
+    ddev composer create --no-scripts craftcms/craft
     ddev craft install/craft \
         --username=admin \
-        --password=password123 \
-        --email=admin@mail.com \
+        --password=Password123 \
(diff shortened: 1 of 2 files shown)


docs/content/users/quickstart.md
```diff
@@ -172,14 +172,16 @@ Environment variables will be automatically added to your `.env` file to simplif
 
     # Boot the project and install the starter project:
     ddev start
-    ddev composer create craftcms/craft
+    ddev composer create --no-scripts craftcms/craft
```
rpkoller, line 175: i thought --no-scripts would be unnecessary to explicitly name since after your changes composer create is taking care of that. and i've manually rechecked: 

ddev composer create --no-scripts craftcms/craft
Executing Composer command: [composer create-project --no-scripts --no-plugins --no-install craftcms/craft /tmp/VBeFAq]

Executing Composer command: [composer create-project --no-plugins --no-scripts --no-install craftcms/craft /tmp/AjEQPy]

to the addition of that flag does no harm but is still unnecessary imho?

rpkoller (commented):
i dont think adding --no-script is necessary as demonstrated in my inline comment. 

aside the rest of the changes are fine. i would have revisited all the quickstarts at a later point after the release and consolidated the comments at a whole same as the ddev start removing the -y on the quickstart (there are still a few left i suppose (but all those do no harm and i would have moved that to the next version). but no object to make those changes for craft in this PR already. but i would still revisit all quickstarts for one of the next versions.

docs/content/users/quickstart.md
stasadev: No, there is a difference, you should look at the different part of the output:

```
$ ddev composer create craftcms/craft
...
Executing Composer command: [composer install]
...
Executing Composer command: [composer run-script post-create-project-cmd]
...
```

vs
```
$ ddev composer create --no-scripts craftcms/craft
...
Executing Composer command: [composer install --no-scripts]
...
```

rpkoller: oh huh how i could have missed that?! i am not really awake yet nor i have been awake when i created that previous PR. when posting my answer i havent waited until the composer create step went through, i indeed run into those prompts without the -no-script and i completely agree that that shouldnt happen. then i am fine with all the changes. 

but just for my own understanding cuz somehow that behavior of composer create is unexpected and sort of confusing. as stated in my initial comment the executing composer command for composer create project is identical with and without the --no-scripts... but why is the executing composer command for composer install then differing? on with and one without --no-scripts? i wouldnt have expected that adding the flag is still necessary in some context because of that composer install step?

rpkoller (approved):

docs/content/users/quickstart.md
stasadev: See https://ddev.readthedocs.io/en/stable/users/usage/developer-tools/#ddev-and-composer

Particularly,

> `ddev composer create` does these things in this order:

rpkoller: i wasnt aware of that fact nor really aware of that section in the docs. i was always under the impression composer create would handle all that automagically for a while now. but as i've learned now only for the create step no the install step (which would be impossible i suppose). interesting detail. thank you for the pointer!

rfay (approved):
Thanks!

PR #7161, opened 2025-03-28
build: bump XHGui image to 0.23, for #7142

stasadev:
## The Issue

XHGui doesn't have a Docker volume anymore:

- https://github.com/perftools/xhgui/pull/520

Removes workaround for xhgui volume.

This PR reverts:

- #7142

```
ddev xhgui
ddev logs -s xhgui -f
```

### DIFF
--- a/containers/ddev-xhgui/Dockerfile
+++ b/containers/ddev-xhgui/Dockerfile
@@ -1,20 +1,4 @@
-FROM xhgui/xhgui:0.21 AS xhgui-docker
-
-FROM scratch AS ddev-xhgui
-
-### --------------------------------xhgui-docker----------------------------------
-### XHGui Docker image creates a volume for /run/nginx, which we want to remove.
-### We copy all files from xhgui/xhgui, and set ENV, WORKDIR, CMD, EXPOSE.
-### Source https://github.com/perftools/xhgui/blob/0.21.x/Dockerfile
-COPY --from=xhgui-docker / /
-ENV PHP_INI_DIR=/etc/php7
-ARG APPDIR=/var/www/xhgui
-ARG WEBROOT=$APPDIR/webroot
-WORKDIR $APPDIR
-RUN mkdir -p /run/nginx
-CMD ["sh", "-c", "nginx && exec php-fpm"]
-EXPOSE 80
-### ------------------------------END xhgui-docker--------------------------------
+FROM xhgui/xhgui:0.23.x AS ddev-xhgui
 
 RUN apk add bash curl
 ADD /var /var
--- a/containers/ddev-xhgui/Dockerfile
+++ b/containers/ddev-xhgui/Dockerfile
@@ -1,4 +1,4 @@
-FROM xhgui/xhgui:0.23.x AS ddev-xhgui
+FROM xhgui/xhgui:0.23.0 AS ddev-xhgui
 
 RUN apk add bash curl
 ADD /var /var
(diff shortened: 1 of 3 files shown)


rfay (approved):
Agreed that we should be cautious about this, but if we are able to verify correct behavior I know @PierrePaul would be very happy to see his fixes show up in the release.  So there is value in bringing this in for the release if we're happy with it.

stasadev:
There is a new release for XHGui, testing it.

containers/ddev-xhgui/Dockerfile
```diff
@@ -1,20 +1,4 @@
...
-COPY --from=xhgui-docker / /
-ENV PHP_INI_DIR=/etc/php7
-ARG APPDIR=/var/www/xhgui
-ARG WEBROOT=$APPDIR/webroot
-WORKDIR $APPDIR
-RUN mkdir -p /run/nginx
-CMD ["sh", "-c", "nginx && exec php-fpm"]
-EXPOSE 80
-### ------------------------------END xhgui-docker--------------------------------
+FROM xhgui/xhgui:0.23 AS ddev-xhgui
```
glensc, line 1:
```suggestion
FROM xhgui/xhgui:0.23.0 AS ddev-xhgui
```

shouldn't you use actual tag? 0.23 is latest 0.23.x release, i.e it's ovewritten with next release

containers/ddev-xhgui/Dockerfile
```diff
@@ -1,20 +1,4 @@
...
 
 RUN apk add bash curl
```
glensc, line 3:
```suggestion
RUN apk add --no-cache bash curl
```

docker best practices not to store INDEX files to container.

containers/ddev-xhgui/Dockerfile
stasadev: I think it's fine to use 0.23 since 0.23.x will be minor releases that shouldn't break anything (plus, we have tests).

This way, we don't need to constantly remember to update it here. (And we handle Traefik in the same way.)

rfay: We are usually OK with following a branch (movable tag, 0.23), especially since we build and maintain a derivative image which is only updated at our release time.  And of course... when Stas did this there wasn't an 0.23.0 :)

glensc: xhgui doesn't guarantee any semver releases (no manpower). besides by semver, in 0.x release series any release is allowed to add a breaking change. 

and of course any change can break someone's workflow: 
- https://xkcd.com/1172/

containers/ddev-xhgui/var/www/xhgui/config/config.php
```diff
@@ -17,6 +17,7 @@
         'pass' => getenv('XHGUI_PDO_PASS') ?: 'db',
         'table' => getenv('XHGUI_PDO_TABLE') ?: 'results',
         'tableWatch' => getenv('XHGUI_PDO_TABLE_WATCHES') ?: 'watches',
+        'initSchema' => getenv('XHGUI_PDO_INIT_SCHEMA') ?: 'true',
```
glensc, line 20: This is out of sync with upstream:

```
src/config.default.php:        'initSchema' => getenv('XHGUI_PDO_INITSCHEMA') ?: 'true',
```

It may be confusing to use different env var names

stasadev: Oh thanks, I didn't actually type, it's all autocomplete.

containers/ddev-xhgui/Dockerfile
stasadev: We're okay with this, and it could actually be a win-win for both of us to catch any bugs as early as possible.

rfay:
We can do this however you prefer @stasadev , as long as we can manually test.

stasadev:
It's good that it will be in the next release.

PR #6593, opened 2024-10-07
fix: check only for IPv4 in DNS lookup, fixes #6592

stasadev:
## The Issue

- #6592

## How This PR Solves The Issue

Checks only for IPv4 in DNS lookup.

## Manual Testing Instructions

It should work when local DNS is configured to resolve IPv4 and IPv6 for the same host at the same time.

I tested it with custom TLD `*.test` (to be sure it doesn't collide with online `*.ddev.site`) and `dnsmasq`:

dnsmasq config:
for IPv4:
```
address=/*.test/127.0.0.1
```
for IPv6:
```
address=/*.test/::1
```

DDEV config:
```
ddev config --project-tld test
```

Cases:
1. Resolve IPv4 `*.test` added, resolve IPv6 `*.test` added - `ddev start` works
2. Resolve IPv4 `*.test` added, resolve IPv6 `*.test` not added - `ddev start` works
3. Resolve IPv4 `*.test` not added, resolve IPv6 `*.test` added - `ddev start` asks for sudo as expected
4. Resolve IPv4 `*.test` not added, resolve IPv6 `*.test` not added - `ddev start` asks for sudo as expected

PR #7166, opened 2025-03-28
fix: add timeout for netutil::IsPortActive check for WSL2 with "mirrored networking mode" as opposed to default "NAT mode", fixes #6245

### LINKED ISSUE #6245: Using WSL2 Mirrored mode networking delays ddev-router start up
### Preliminary checklist

_No response_

### Expected Behavior

Regardless of the networking mode configured in WSL2, ddev-router should start up fast.

### Actual Behavior

After enabling the [mirrored networking](https://learn.microsoft.com/en-us/windows/wsl/networking#mirrored-mode-networking) mode in WSL2, it takes a while (as in a few minutes) until ddev-router is started when running `ddev start` and ddev-router is not already running. This is the ddev output regarding ddev-router startup:
```
Unable to properly check port status: dial tcp 127.0.0.1:443: connect: connection timed out
Unable to properly check port status: dial tcp 127.0.0.1:80: connect: connection timed out
Unable to properly check port status: dial tcp 127.0.0.1:8025: connect: connection timed out
Unable to properly check port status: dial tcp 127.0.0.1:8026: connect: connection timed out
```

It looks like this is a "general" issue with how the networking mode works in detail, e.g when I run `telnet 127.0.0.1 443` in my WSL2 instance, it also hangs for ages. 

### Steps To Reproduce

_No response_

### Anything else?

_No response_

marvinhinz:
## The Issue

- #6245

WSL2 with networkingMode=mirrored seems to break port checks, as there is no response for the dial attempt and ddevs netutil IsPortActive waits around 5 minutes for each port to check when starting a project. This makes it quite unusable. The windows team seems not able to fix it in a reasonable time so im proposing this simple change as a workaround. 

Instead of waiting 5 minutes for each port check on WSL2 (mirrored networking mode), it waits for .2 second max per check.  The timeout approach is only used on WSL2, and it's possible that in NAT mode it doesn't hit it anyway, but if it did it should be no more than about 1s total added to `ddev start`.

- [ ] Use `export DDEV_DEBUG=true` to get more info about what it's doing
- [ ] Test `ddev poweroff && ddev start -y` on non-WSL2 to verify behavior
- [ ] Test `ddev poweroff && ddev start -y` on WSL2 in NAT mode
- [ ] Test `ddev poweroff && ddev start -y` on WSL2 in mirrored mode

Note that the delay in mirrored mode happens only when the port is *not* occupied, which is what's superbly annoying about this. It happens to new users or first start. 

To occupy ports manually, consider `docker run -it --rm -p 80 -p 443 busybox sh`, which occupies 80 and 443, a common situation. In that situation it should *not* time out on 80/443, but rather it finds them occupied.

## Automated Testing Overview

We may want to consider putting one of our WSL2 test runners into mirrored mode after this goes into HEAD.

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -2,6 +2,7 @@ package netutil
 
 import (
 	"fmt"
+	"github.com/ddev/ddev/pkg/nodeps"
 	"net"
 	"net/url"
 	"os"
@@ -16,19 +17,39 @@ import (
 
 // IsPortActive checks to see if the given port on Docker IP is answering.
 func IsPortActive(port string) bool {
+
+	dialTimeout := 0 * time.Millisecond
+	if nodeps.IsWSL2() {
+		dialTimeout = 200 * time.Millisecond
+	}
+
 	dockerIP, err := dockerutil.GetDockerIP()
 	if err != nil {
 		util.Warning("Failed to get Docker IP address: %v", err)
 		return false
 	}
 
-	conn, err := net.DialTimeout("tcp", dockerIP+":"+port, 1*time.Second)
+	util.Debug("Checking if port %s is active", port)
+	conn, err := net.DialTimeout("tcp", dockerIP+":"+port, dialTimeout)
 
 	// If we were able to connect, something is listening on the port.
 	if err == nil {
 		_ = conn.Close()
 		return true
 	}
+
+	// In WSL2 mirrored mode, when we test an unused port, we just get a timeout
+	// Assume that the port is available (not active) in that situation.
+	// This seems to be caused by https://github.com/microsoft/WSL/issues/10855
+	// We don't have a way to know whether WSL2 in mirrored mode, but
+	// we use the longer timeout in WSL2 and assume that timeout is unoccupied.
+	if nodeps.IsWSL2() {
+		if err, ok := err.(net.Error); ok && err.Timeout() {
+			util.Debug("In WSL2 and port %s is probably not active; timeout", port)
+			return false
+		}
+	}
+
 	// If we get ECONNREFUSED the port is not active.
 	oe, ok := err.(*net.OpError)
 	if ok {
@@ -38,6 +59,7 @@ func IsPortActive(port string) bool {
 		var WSAECONNREFUSED syscall.Errno = 10061
 
 		if ok && (syscallErr.Err == syscall.ECONNREFUSED || syscallErr.Err == WSAECONNREFUSED) {
+			util.Debug("port %s shows connection refused so not active", port)
 			return false
 		}
 	}
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/util"
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -21,7 +21,7 @@ func IsPortActive(port string) bool {
 		return false
 	}
 
-	conn, err := net.Dial("tcp", dockerIP+":"+port)
+	conn, err := net.DialTimeout("tcp", dockerIP+":"+port, 1*time.Second)
 
 	// If we were able to connect, something is listening on the port.
 	if err == nil {


rfay:
Thanks for the PR! 

Could you please open an issue with the complete description of your problem?  For example, please link to https://learn.microsoft.com/en-us/windows/wsl/networking and also explain why you want to use mirrored mode and what you're trying to accomplish.  I do see it's the "new normal", although WSL2 defaults to NAT mode. 

It seems like perhaps the port check needs to be more sophisticated, rather than just having a timeout. 

I wonder if [this discord thread](https://discord.com/channels/664580571770388500/1352833466076631050/1352833466076631050) relates to this. 

It also may be possible that the WSL2 project needs an issue.

marvinhinz:
Thanks for the feedback!

The issue linked is essentially pretty spot on. A bit on the background why i use the mirrored mode: For my projects i need to connect to some ipv6 only services, that are hard to mock (saltstack / prometheus & thanos). Other than that its easier to use host.docker.internal to access services running on my windows host that way. As ipv6 adoption grows faster and faster and v6-only-availability becomes a thing, the need to have proper ipv6 connectivity inside wsl seems well justified.

From the wsl docs:

`There are a few considerations to be aware of when working with networking apps and WSL. By default WSL uses a NAT based architecture, and we recommend trying the new Mirrored networking mode to get the latest features and improvements.` 

Currently its recommended to try the mirrored mode, who know if in a few months/years it will become the default setting. 
Then this issue might impact more ddev users.

I totally get that this ddev itself is not the culprit, but as some microsoft wsl bugfixes seem to take literal years to arrive, a workaround/improvement in ddev might be sensible. The wsl issue is open since November 2023 https://github.com/microsoft/WSL/issues/10855

The "bug" in ddev is that when the port check does not work because of obscure network problems, it will take minutes for each port check to fail, and that makes it completely unusable, as project startup takes nearly half an hour. 
I think regardless of the circumstances this should be prevented by ddev.

Do you have another idea to improve the IsPortActive check to make it more robust in such cases?

rfay:
So you're saying that the key concern is that you can't make ipv6 connections from inside a normal WSL2 distro when in NAT mode?

Docker itself has only recently started supporting IPV6 and I think it's spotty. Are you needing to access external (or internal) services from inside the web container, for example?

marvinhinz:
Yes WLS2 in NAT mode doesnt support ipv6. WLS1 did, but the overall performance was poor.

Im not talking about connections from inside docker containers to ipv6 services primarily - although ddev ipv6 support would be very nice too, but thats a pretty big task i guess. 

Our internal project infrastructure is mostly ipv6 only, and i cant interact with it from my ide or usual shell, i have to use cmd on the windows side for that. Also using the mirrored mode i have one hoop less to jump through when connecting to ipv6 only services from the containers by using socat inside wsl to map a service to host.docker.internal. With the nat mode i have to setup a windows portproxy and manage firewall settings first, and then use socat to proxy that service to the wsl host ip too. 

I get that this is rather an inconvenience than a show stopper, but i think the proposed change is minimal and might improve the overall ddev experience for some users.

rfay:
Is the message you get the one in the discord channel, `Unable to properly check port status for 127.0.0.1:80: err=dial tcp 127.0.0.1:80: connect: connection timed out` ? That discord thread never mentioned that the timeout took forever.

rfay:
Another question, the user in the Discord thread ended up using the Traditional Windows install, which requires Docker Desktop, which by default runs in WSL2. I wonder whether there are any problems with that approach in mirrored networking mode.

pkg/netutil/netutil.go
```diff
@@ -21,7 +22,7 @@ func IsPortActive(port string) bool {
 		return false
 	}
 
-	conn, err := net.Dial("tcp", dockerIP+":"+port)
+	conn, err := net.DialTimeout("tcp", dockerIP+":"+port, 1*time.Second)
```
stasadev, line 25: @marvinhinz, can you experiment with network here? Replace `tcp` with `tcp4`

```suggestion
	conn, err := net.Dial("tcp4", dockerIP+":"+port)
```

We already use IPv4 only in some other places, but not here:

- #6593

stasadev: P.S. From https://pkg.go.dev/net#Dial

It may fail because we don't pass IPv6 correctly (but we don't need it anyway?):

> If the host is a literal IPv6 address it must be enclosed in square brackets, as in "[2001:db8::1]:80" or "[fe80::1%zone]:80".

rfay:
Moved to draft while we study and learn about this.

rfay:
I experimented with a few things in mirrored mode.

1. This approach works, but it doesn't actually do the test we want to test, so we don't know if the port is occupied.
2. I tried doing a direct listen, but that has the disadvantage that ports below 1024 require extra privileges to listen; and of course the most important things we check are 80/443.
3. I tried running a docker container and having it listen (ddev-webserver is set up to allow that). But in mirrored mode it doesn't fail in that situation apparently.

rfay:
After looking at all those things and looking at your basic code, I just did some finesse. When in WSL2 we timeout, then the timeout is the same as "nothing there". The basic flow in other environments should remain the same.

pkg/netutil/netutil.go
rfay: I didn't end up doing this, but it can still be done. I experimented with it at one point but it wasn't really relevant it didn't seem.

rfay:
I manually tested on WSL2:
* Mirrored mode, testing not-listening port, it does the timeout and reports it (if Debug is on)
* NAT mode, not-listening port, it skips the timeout and behaves traditionally.

This may mean that we don't need the WSL2-specific code at all in the first place; the timeout doesn't do anything on other platforms because we immediately have success (connection, port is already listening) or failure (connection refused, nothing listening on port).

stasadev (approved):
I tested mirrored and NAT mode using Windows VM and using testbot machine.

Behaves as it should, thank you everyone!

etxaleku:
This fix is awesome, thanks!  Makes drupal development with ddev under mirrored networking work great!  And mirrored networking seems pretty essential to working with corporate VPNs remotely.

etxaleku:
Hmm, this is odd, the behavior seems to have regressed.

```
Unable to properly check port status for 127.0.0.1:80: err=dial tcp 127.0.0.1:80: connect: connection timed out
Unable to properly check port status for 127.0.0.1:443: err=dial tcp 127.0.0.1:443: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8025: err=dial tcp 127.0.0.1:8025: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8026: err=dial tcp 127.0.0.1:8026: connect: connection timed out
```

```
$ ddev --version
ddev version v1.24.4

$ wslinfo --networking-mode
mirrored
```

In an effort to try and "fix it" I just went through an "apt update/apt upgrade" cycle and a reboot to re-initialize everything but the timeouts remain.  It was working so smoothly 4 days ago through many "ddev stop/ddev start" iterations, including some wsl --shutdown's as well.

This new behavior is the same as before (few weeks ago) - those ports just sit in SYN_SENT state (serially) for 2 minutes each and then once all of them have timed out (twice) then everything runs smoothly after that.  Makes ddev restarts pretty painful unfortunately as every restart takes 20+ minutes.

```
$ ddev start
Starting pmr...
Unable to properly check port status for 127.0.0.1:80: err=dial tcp 127.0.0.1:80: connect: connection timed out
Unable to properly check port status for 127.0.0.1:443: err=dial tcp 127.0.0.1:443: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8025: err=dial tcp 127.0.0.1:8025: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8026: err=dial tcp 127.0.0.1:8026: connect: connection timed out
 Container ddev-ssh-agent  Created
Building project images....
Project images built in 2s.
 Network ddev-pmr_default  Created
 Container ddev-pmr-web  Created
 Container ddev-pmr-db  Created
 Container ddev-pmr-db  Started
 Container ddev-pmr-web  Started
Unable to properly check port status for 127.0.0.1:443: err=dial tcp 127.0.0.1:443: connect: connection timed out
Unable to properly check port status for 127.0.0.1:80: err=dial tcp 127.0.0.1:80: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8025: err=dial tcp 127.0.0.1:8025: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8026: err=dial tcp 127.0.0.1:8026: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8142: err=dial tcp 127.0.0.1:8142: connect: connection timed out
Unable to properly check port status for 127.0.0.1:8143: err=dial tcp 127.0.0.1:8143: connect: connection timed out
Successfully started pmr
Your project can be reached at https://pmr.ddev.site
See 'ddev describe' for alternate URLs.
$
```

Monitoring the timing on the SYN_SENT states it is consistently waiting 2 minutes for each port check, 4 of them the first time through and then 6 of them the second time through.  Here is an example of the first 4 port check -

```
11:37:15 tcp        0      1 127.0.0.1:47319         127.0.0.1:443           SYN_SENT
11:39:15 tcp        0      1 127.0.0.1:47319         127.0.0.1:443           SYN_SENT

11:39:25 tcp        0      1 127.0.0.1:48155         127.0.0.1:80            SYN_SENT
11:41:25 tcp        0      1 127.0.0.1:48155         127.0.0.1:80            SYN_SENT

11:41:35 tcp        0      1 127.0.0.1:47915         127.0.0.1:8025          SYN_SENT
11:43:36 tcp        0      1 127.0.0.1:47915         127.0.0.1:8025          SYN_SENT

11:43:46 tcp        0      1 127.0.0.1:46953         127.0.0.1:8026          SYN_SENT
11:45:46 tcp        0      1 127.0.0.1:46953         127.0.0.1:8026          SYN_SENT

11:45:56 tcp        0      1 127.0.0.1:44775         127.0.0.1:8142          SYN_SENT
11:47:57 tcp        0      1 127.0.0.1:44775         127.0.0.1:8142          SYN_SENT

11:48:07 tcp        0      1 127.0.0.1:45653         127.0.0.1:8143          SYN_SENT
11:50:17 ... back to normal operation
```

Is this expected and will correct itself at the next update or something?

rfay:
The fix here is in HEAD, not in v1.24.4.  https://ddev.readthedocs.io/en/stable/developers/building-contributing/#testing-latest-commits-on-head

It will be released in v1.24.5, but easy to use now.

etxaleku:
Ah!  Super thanks for the HEADs up!   :)

etxaleku:
Just a small FYI ... the original apt installed ddev as /usr/bin/ddev but this HEAD install placed it in /usr/local/bin/ddev.  I was able to straighten it out with a simple ```cp /usr/local/bin/ddev /usr/bin/ddev``` but thought I should mention it.  It's back to working smoothly now - thanks again!

rfay:
You could also handle that situation by putting /usr/local/bin first in your PATH, I usually have that. Then when you want to go back to apt-installed version, just remove the /usr/local/bin/ddev.

stasadev:
I used to override `ddev` in `/usr/local/bin/ddev` and `~/.local/bin/ddev`, but in the end, I just replace `/usr/bin/ddev` and keep a backup next to it.

This is all because the JetBrains DDEV Integration plugin has a hard-coded path to the binary, and it doesn't update for previously opened projects. That causes it to use different binary from what I have in the terminal, so I just replace the original to keep things consistent.

BTW, if you replace the original, you don't have to worry about it anymore, since the next DDEV release will simply overwrite it.

PR #7182, opened 2025-04-02
fix: XHGui launch command support custom ports, fixes #7181

### LINKED ISSUE #7181: XHGui launches to invalid URL when using non-standard ports
### Preliminary checklist

_No response_

### Expected Behavior

`ddev xhgui` will open the webinterface in the browser

### Actual Behavior

There is an error:
```
➜  acme-project git:(main) ddev xhgui
 Container ddev-acme-project-xhgui  Created 
 Container ddev-acme-project-xhgui  Started 
Started optional compose profiles 'xhgui' 
Enabled XHProf for XHGui. 
failed to launch xhgui: output='Unable to interpret 'https://acme-project.ddev.site:33001:8142' as a path or URL
Failed to run launch https://acme-project.ddev.site:33001:8142; error=exit status 1
', err=exit status 1 
```

### Steps To Reproduce

1. `ddev config global --router-http-port=8080 --router-https-port=8443` 
2. start a project
3. run `ddev xhgui`

The problem occurs in both cases:
1. when the router is configured for different ports than 80/443
2. when DDEV automatically detects a port conflict and assigns a random port number as in my example

### Anything else?

_No response_

PierrePaul:
## The Issue

- #7181 

This PR move the logic of building the xhgui url into the describe method, just like mailpit.

- ddev config global --router-http-port=8080 --router-https-port=8443
- ddev restart
- ddev xhgui launch

The launch should now work instead of throwing an error.

## Automated Testing Overview

Automated test has been updated to validate values are properly filled in by the describe method.

### DIFF
--- a/cmd/ddev/cmd/xhgui_test.go
+++ b/cmd/ddev/cmd/xhgui_test.go
@@ -64,7 +64,12 @@ func TestCmdXHGui(t *testing.T) {
 	time.Sleep(2 * time.Second)
 
 	// Now hit xhgui UI
-	xhguiURL := app.GetXHGuiURL()
+	desc, err := app.Describe(true)
+	require.NoError(t, err)
+	require.NotNil(t, desc["xhgui_url"])
+	require.NotNil(t, desc["xhgui_https_url"])
+	xhguiURL := desc["xhgui_https_url"].(string)
+
 	out, _, err = testcommon.GetLocalHTTPResponse(t, xhguiURL, 2)
 	require.NoError(t, err)
 	// Output should contain at least one run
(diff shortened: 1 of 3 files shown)


rfay:
A casual glance says this is a great improvment and will solve the problem (and is in keeping with what we've done many places)

Although we've been really sloppy in many places (like creating the problem here) with constructing URLs as strings on the fly, resulting in bugs like this, we could probably do better. 

Would you take a look at the actual go url creation functions, like used in https://github.com/ddev/ddev/blob/213e8cc607433e5c683b3ca0fc0df1148c069c2c/pkg/dockerutil/dockerutils.go#L926-L945 and see if using those might result in more robust overall code? Maybe I'm grasping, but it's obvious that I created this bug by doing silly things with strings :)

stasadev (approved):
Thank you!

Looks good and works well with different router ports and with ephemeral router ports.

> I did notice that when we do `ddev xhgui on` we see the router being recreated

@rfay, I don't see this behavior. My guess is that your router had some other state instead of `running` at that time.

When I run this several times, I see only this output, it's not recreating for me:

```
$ ~/.ddev/bin/docker-compose --progress=plain -f ~/.ddev/.router-compose-full.yaml -p ddev-router up --build -d
```

rfay:
Thanks @PierrePaul ! Congrats and looking forward to your amazing future contributions. Thanks for making XHGui so much better.

PR #7187, opened 2025-04-03
build: remove <1.24 constraint for go, for #7057

stasadev:
## The Issue

- #7057
- https://github.com/moby/moby/issues/49513

Upstream has a fix for go, which is already included in Arch Linux. (I'm already using it.)

The bug only affected Linux with kernel 6.11+

And I don't think DDEV has ever been affected by this bug seen in Docker.

---

I used `go-version: '>=1.23'` everywhere because this is what we have in `go.mod`:

https://github.com/ddev/ddev/blob/19a3e5fb27ca1793c4d2c4a809816b3245cd75dd/go.mod#L85

---

Adding `check-latest: true`, because it still used `go1.23.7` from cache.

Not needed.

### DIFF
--- a/.github/workflows/colima-tests.yml
+++ b/.github/workflows/colima-tests.yml
@@ -84,6 +84,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: '>=1.23'
+          check-latest: true
 
       - name: Set up Homebrew
         id: set-up-homebrew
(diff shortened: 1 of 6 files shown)


stasadev:
Adding `check-latest: true`, because it still used `go1.23.7` from cache:

```
Setup go version spec >=1.23
Found in cache @ /opt/hostedtoolcache/go/1.23.7/x64
Added go to the path
Successfully set up Go version >=1.23
```

And actually we should wait for go1.24.3 release. Moving this to draft.

rfay (approved):
Thanks!

PR #7188, opened 2025-04-04
docs: show commands to control xdebug inside the web container

MurzNN:
## The Issue

- No documentation about how to enable or disable xdebug directly from the container

Added info about available commands to control xdebug inside the container.

## Release/Deployment Notes

Document commands to control xdebug inside the container

### DIFF
--- a/docs/content/users/debugging-profiling/step-debugging.md
+++ b/docs/content/users/debugging-profiling/step-debugging.md
@@ -11,7 +11,7 @@ All IDEs basically work the same, listening on a port and reacting when they’r
 * Enable Xdebug by running [`ddev xdebug`](../usage/commands.md#xdebug) or `ddev xdebug on` from your project directory.
 It will remain enabled until you start or restart the project.
 * Disable Xdebug for better performance when not debugging with `ddev xdebug off`.
-* Toggle Xdebug on and off easily with `ddev xdebug toggle`.
+* Toggle Xdebug on and off easily with `ddev xdebug toggle` (or `disable_xdebug`/`enable_xdebug` inside the container).
 * `ddev xdebug status` will show Xdebug’s current status.
 * You may need to open port 9003 in your firewall to allow Xdebug access. (See [Troubleshooting Xdebug](#troubleshooting-xdebug) below.)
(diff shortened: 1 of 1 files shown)


docs/content/users/debugging-profiling/step-debugging.md
```diff
@@ -11,7 +11,7 @@ All IDEs basically work the same, listening on a port and reacting when they’r
 * Enable Xdebug by running [`ddev xdebug`](../usage/commands.md#xdebug) or `ddev xdebug on` from your project directory.
 It will remain enabled until you start or restart the project.
 * Disable Xdebug for better performance when not debugging with `ddev xdebug off`.
-* Toggle Xdebug on and off easily with `ddev xdebug toggle`.
+* Toggle Xdebug on and off easily with `ddev xdebug toggle` (or `disable_xdebug`/`enable_xdebug` inside the container).
```
rfay, line 14: Did you know that you can use many web commands directly? So instead of `ddev xdebug off` you can use `xdebug off` or `xdebug on`. 

If you accidentally use `ddev xdebug off` inside the web container it will tell you:

```
rfay@d11-web:/var/www/html$ ddev xdebug off
You executed 'ddev xdebug off' inside the web container
but many DDEV commands are not available.

However, you may be able to use a command like:

xdebug off
```

I imagine it's easier to show these commands than the `enable_xdebug` varieties?

rfay:
BTW, I imagine you know this but your LE cert for https://korepov.pro/ has expired.

docs/content/users/debugging-profiling/step-debugging.md
MurzNN: Wow, didn't know this, thanks! So, in that case this tip is not needed then ;)

MurzNN: But seems we need to extend another doc them - probably this one https://github.com/ddev/ddev/blob/main/docs/content/users/usage/commands.md by adding info which commands are available inside the container.

stasadev: All `web` commands are available inside:

- #6902

rfay:
Moved to draft since needs followup.

rfay:
Closing for now, happy to have a PR to add this info to docs.

Issue #2276, opened 2020-05-22
Support podman as a docker alternative


benhosmer:
**Is your feature request related to a problem? Please describe.**
For users in the RedHat family, ie. CentOS, Fedora, podman and cgroups v2 is the default and offers
other benefits that docker doesn't.

**Describe the solution you'd like**
I'd like to use ddev without needing to install docker.

**Describe alternatives you've considered**
None.

I realize docker-compose is the likely the trickiest part. One thing of note, podman doesn't run as root at all so you'd likely gain some more security.

rfay:
You'll still have to install docker-compose; ddev is a wrapper on docker-compose. And if you have to install docker-compose... Why not just install docker?

benhosmer:
> Why not just install docker?
Because Docker isn't the only container game in town any longer and podman will soon support compatibility with a docker-compose style syntax.

crydotsnake:
I dont think its possible to use ddev without Docker. Because DDEV is based on Docker. I think, you would need to recreate DDEV from the beginning, to make it running without Docker.

What are the reasons, why you didnt like docker?.

atomicptr:
> I dont think its possible to use ddev without Docker. Because DDEV is based on Docker. I think, you would need to recreate DDEV from the beginning, to make it running without Docker.
> 
> What are the reasons, why you didnt like docker?.
> 
> And here you have the install steps for installing docker-compose on linux:
> https://docs.docker.com/compose/install/
> 
> It should be work on all Linux Distributions.

There are a few advantages podman has over Docker like supporting cgroups v2 (which Fedora for instance has installed by default), not relying on a daemon and being able to run containers with your user permissions instead of as root.

You can generally go quite far with most Docker based scripts/tools by using:

```bash
$ alias docker=podman
# https://github.com/containers/podman-compose
$ alias docker-compose=podman-compose
```

Since this seemed kind of interesting, I played around a bit with the idea...

By enabling podmans API via:

```bash
$ podman system service --timeout 5000
```

and creating a link for the docker.sock

```bash
$ ln -s /$XDG_RUNTIME_DIR/podman/podman.sock /var/run/docker.sock
```

I got ddev to acknowledge it as a docker, sadly it will complain about the version by this point:

```bash
$ ddev start
The docker version currently installed does not meet ddev's requirements: 1.8.2 is less than 18.06.1-alpha1 
```

This might be worthwhile to look into further and possibly provide a "podman mode" :).

rfay:
[...]

PR #7189, opened 2025-04-05
refactor: replace docker command with docker/cli library, fixes #5862

### LINKED ISSUE #5862: GetDockerContext() should no longer be running `docker context`
In trying to debug recent perpetual Colima failures - 
* https://github.com/ddev/ddev/issues/5861

I stumbled on GetDockerContext() still calling out to `docker client` for context information. I'm sorry I missed this in the review of the Docker rework @stasadev , but GetDockerContext() should now be able to use native techniques rather than running `docker context inspect`

https://github.com/ddev/ddev/blob/49bf2a402f547776207896bf0b19d8ee580411aa/pkg/dockerutil/dockerutils.go#L207-L219

1. Nothing about docker contexts should require calling out to the `docker` client I don't think.
2. We shouldn't be setting or using DOCKER_HOST any more (it should be allowed as explicitly set as an override, but we shouldn't be using it internally):

https://github.com/ddev/ddev/blob/b3f7cf6df550f6b6d2cedaca9d64c98521ec2fcd/pkg/dockerutil/dockerutils.go#L192-L195

stasadev:
## The Issue

- #5862

- [x] Initial try to use `docker/cli` in `GetDockerContext()`
- [ ] Refactor `GetDockerClient()` to return both `client` and `cli`
- [ ] Replace all direct `docker` calls with `cli`
- [ ] Remove `DOCKER_HOST` env

I think all of this will be a step forward for using Podman (without installing the `docker` client).

## Release/Deployment Notes

This will increase the size of the `ddev` binary.


stasadev:
I investigated why the binary size is so big, this comes from the telemetry:

- https://github.com/docker/cli/pull/4889#pullrequestreview-1914189938

> I think integrating otel things into the cli like this could cause some issues down the road, since we'd be using otel structs and methods directly even in projects that import the cli to use this common code. Maybe it could be worth it to decouple our usage of telemetry things from our actual implementation using otel, also for keeping things more easily standardized across projects;

- https://github.com/docker/cli/pull/4889#issuecomment-1983877922

> My personal hope is that, with this in the CLI, it'll mostly standardize the version the various CLI programs use. I do think a follow up thing after that is maybe creating a separate repository for OTEL stuff in general so server and cli components are synced on the versions.

For now, telemetry comes with `github.com/docker/cli` 🙁

This can be "fixed" by copying all files from `vendor/github.com/docker/cli/cli/command` elsewhere and removing or modifying the telemetry parts. I did this, which helped reduce the size, but it requires constant maintenance on package update.

```
# DDEV HEAD ddev-linux-amd64.zip
$ ls -l /usr/bin/ddev 
-rwxr-xr-x 1 stas stas 19222712 Aug 11 14:18 /usr/bin/ddev

# This PR ddev-linux-amd64.zip
$ ls -l /usr/bin/ddev
-rwxr-xr-x 1 stas stas 25137336 Aug 11 17:33 /usr/bin/ddev

# manual removal of telemetry stuff
ls -l /usr/bin/ddev
-rwxr-xr-x 1 stas stas 19734712 Aug 11 17:33 /usr/bin/ddev
```

In result: before 19.2 MB, after 25.1 MB.

If we remove telemetry manually, then size is 19.7 MB - but it requires too much work for us.

rfay:
Unless I'm reading it wrong, we're talking about a few MB, which doesn't seem to be a big issue in the context of how DDEV is used.  So docker/cli is using Open Telemetry project, and that just adds lots of code?

stasadev:
Yes, that's correct.

The initial increase in size was from 17 MB to 33 MB, in the previous rebase, that's why I looked at this.

stasadev:
I removed `DOCKER_HOST` from here.

`DOCKER_CONTEXT` and `DOCKER_HOST` are fully respected without using them directly.

rfay:
With removal of DOCKER_HOST we have to make sure that remote docker environments still work, https://ddev.readthedocs.io/en/stable/users/topics/remote-docker/

We don't really have testing for that, and we have the testing for no_bind_mounts turned off, perhaps we should turn it on for the time being?  Or maybe that's unrelated, but normally remote docker and no-bind-mounts go together.

https://github.com/ddev/ddev/blob/2d50d821805e9be1462be3b05571ee5a2b38573e/.github/workflows/tests.yml#L56

stasadev:
> With removal of DOCKER_HOST we have to make sure that remote docker environments still work, [ddev.readthedocs.io/en/stable/users/topics/remote-docker](https://ddev.readthedocs.io/en/stable/users/topics/remote-docker/)

I [enabled](https://gist.github.com/styblope/dc55e0ad2a9848f2cc3307d4819d819f) TCP port 2375 in my Ubuntu 24.04 guest (VMWare Workstation), and then:

```
$ vmrun getGuestIPAddress "$(vmrun list | grep vmx)"
192.168.185.141

$ docker context create ubuntu --description "ubuntu" --docker "host=tcp://192.168.185.141:2375"

$ ddev debug dockercheck
Docker platform: linux-docker 
Using Docker context: default 
Using Docker host: unix:///var/run/docker.sock 
docker-compose: v2.38.2 
Docker version: 28.3.3 
Docker API version: 1.51 
Able to run simple container that mounts a volume. 
Able to use internet inside container.

$ DOCKER_CONTEXT=ubuntu ddev debug dockercheck      
Docker platform: linux-docker 
Using Docker context: ubuntu 
From DOCKER_CONTEXT=ubuntu 
Using Docker host: tcp://192.168.185.141:2375 
docker-compose: v2.38.2 
Docker version: 28.3.3 
Docker API version: 1.51 
Able to run simple container that mounts a volume. 
Able to use internet inside container.

$ DOCKER_HOST=192.168.185.141 ddev debug dockercheck
Docker platform: linux-docker 
Using Docker context: default 
Using Docker host: tcp://192.168.185.141:2375 
From DOCKER_HOST=192.168.185.141 
docker-compose: v2.38.2 
Docker version: 28.3.3 
Docker API version: 1.51 
Able to run simple container that mounts a volume. 
Able to use internet inside container.

$ docker context use ubuntu

$ ddev debug dockercheck
Docker platform: linux-docker 
Using Docker context: ubuntu 
Using Docker host: tcp://192.168.185.141:2375 
docker-compose: v2.38.2 
Docker version: 28.3.3 
Docker API version: 1.51 
Able to run simple container that mounts a volume. 
Able to use internet inside container.

$ ddev config global --no-bind-mounts && ddev start
...
Successfully started d11 
Your project can be reached at https://d11.ddev.site
See 'ddev describe' for alternate URLs.
```

stasadev:
```
# guest
$ sudo tcpdump -i any -A -s0 tcp port 2375

# host
$ ddev poweroff
```

Logs from DDEV v1.24.7:

```
GET /v1.48/containers/json?all=1&filters=%7B%22label%22%3A%7B%22com.ddev.site-name%22%3Atrue%7D%7D HTTP/1.1
Host: 192.168.185.141:2375
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip
```

Logs from this PR (I edited the `User-Agent` header, I can't edit it completely, but this is at least something):

```
GET /v1.51/containers/json?all=1&filters=%7B%22label%22%3A%7B%22com.ddev.site-name%22%3Atrue%7D%7D  HTTP/1.1
Host: 192.168.185.141:2375
User-Agent: Docker-Client/ddev-v1.24.7-36-g10e806f46-dirty (linux)
Accept-Encoding: gzip
```

By default it looked like:
```
User-Agent: Docker-Client/unknown-version (linux)
```

stasadev:
Podman testing:

- https://github.com/ddev/ddev/issues/2276#issuecomment-3180099753

(I'm not going to do anything for Podman in this PR, because it's out of scope.)

stasadev:
> Try Podman. There's no guarantee it will work out of the box, but it's worth a try.

This PR will make things easier for:
- Podman: can be used without `podman-docker` layer.
- Docker Desktop: has a setting to enable or disable the `docker` CLI (and now I can't find this setting).

> Do the same with rootless Docker on Linux. I don't know if it will work either, but it's worth trying.

This change is unrelated to rootless Docker. We can already run `ddev start` on v1.24.7, and removing `docker context` does not affect the `rootless` context.

pkg/dockerutil/dockerutils.go
```diff
@@ -1982,6 +1973,9 @@ func CanRunWithoutDocker() bool {
 	if output.ParseBoolFlag("version", "v") || output.ParseBoolFlag("help", "h") {
 		return true
 	}
+	if len(os.Args) == 2 && output.ParseBoolFlag("json-output", "j") {
+		return true
+	}
```
stasadev, line 1978: I noticed that `DOCKER_HOST=foo ddev -j` doesn't work without Docker, but it should.

pkg/dockerutil/dockerutils.go
```diff
@@ -171,75 +176,67 @@ func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, n
...
 
-	// This section is skipped if $DOCKER_HOST is set
-	if DockerHost == "" {
-		DockerContext, DockerHost, err = GetDockerContext()
-		// ddev --version may be called without Docker client or context available, ignore err
-		if err != nil && !CanRunWithoutDocker() {
-			util.Failed("Unable to get Docker context: %v", err)
+// GetDockerClient returns a Docker client respecting the current Docker context and host
+func GetDockerClient() (context.Context, dockerClient.APIClient) {
+	initOnce.Do(func() {
```
stasadev, line 193: `sync.Once` is used in the `docker/cli` too, prevents race conditions.

pkg/ddevapp/app_compose_template.yaml
```diff
@@ -306,8 +306,6 @@ services:
       - XHGUI_PDO_USER=db
       - XHGUI_PDO_PASS=db
       - TZ={{ .Timezone }}
-    links:
-      - db
```
stasadev, line 310: Deleted this for Podman, it didn't help, but still - `links:` are not required nowadays.

pkg/ddevapp/testdata/TestNetworkAmbiguity/docker-compose.test.yaml
```diff
@@ -6,6 +6,3 @@ services:
     labels:
       com.ddev.site-name: ${DDEV_SITENAME}
       com.ddev.approot: $DDEV_APPROOT
-  web:
-    links:
-      - "test:test"
```
stasadev, line 11: The same, removed `links:` for Podman.

cmd/ddev/cmd/debug-dockercheck.go
```diff
@@ -41,18 +41,21 @@ var DebugDockercheckCmd = &cobra.Command{
...
 		dockerContext := os.Getenv("DOCKER_CONTEXT")
 		if dockerContext != "" {
-			util.Success("Using DOCKER_CONTEXT=%s", dockerContext)
+			util.Success("From DOCKER_CONTEXT=%s", dockerContext)
+		}
+
+		util.Success("Using Docker host: %s", dockerutil.DockerHost)
+		dockerHost := os.Getenv("DOCKER_HOST")
+		if dockerHost != "" {
+			util.Success("From DOCKER_HOST=%s", dockerHost)
```
stasadev, line 54: I didn't remove `os.Getenv("DOCKER_HOST")` just because we need more debugging information in the `ddev debug test`.

docs/content/users/topics/remote-docker.md
```diff
@@ -12,7 +12,7 @@ You can use remote Docker instances, whether on the internet, inside your networ
 * Create a Docker context that points to the remote Docker instance. For example, if the remote hostname is `debian-11`, then `docker context create debian-11 --docker host=tcp://debian-11:2375 && docker use debian-11`. Alternately, you can use the `DOCKER_HOST` environment variable, e.g. `export DOCKER_HOST=tcp://debian-11:2375`.
```
stasadev:
```suggestion
* Create a Docker context that points to the remote Docker instance. For example, if the remote hostname is `debian-11`, then `docker context create debian-11 --docker host=tcp://debian-11:2375 && docker context use debian-11`. Alternately, you can use the `DOCKER_HOST` environment variable, e.g. `export DOCKER_HOST=debian-11` or `export DOCKER_HOST=tcp://debian-11:2375`.
```

pkg/dockerutil/dockerutils.go
```diff
@@ -153,7 +157,8 @@ func RemoveNetworkWithWarningOnError(netName string) {
 // RemoveNetworkDuplicates removes the duplicates for the named Docker network
 // This means that if there is only one network with this name - no action,
 // and if there are several such networks, then we leave the first one, and delete the others
-func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, netName string) {
+func RemoveNetworkDuplicates(netName string) {
```
rfay, line 160: Your old friend returns!

pkg/dockerutil/dockerutils.go
```diff
@@ -171,75 +176,67 @@ func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, n
...
-	dockerContext := ""
-	dockerHost := ""
+// GetDockerClientErr returns the error from GetDockerClient initialization
+func GetDockerClientErr() error {
+	return initErr
+}
 
-	// This is a cheap way of using Docker contexts by running `docker context inspect`
-	// I would wish for something far better, but trying to transplant the code from
-	// docker/cli did not succeed. rfay 2021-12-16
```
rfay, line 218: It's amazing that we've been finding our way through contexts all this time... and now you figured out how to do it right. Wow, do the years pass.

pkg/dockerutil/dockerutils.go
```diff
@@ -171,75 +176,67 @@ func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, n
...
-	contextInfo, err := ddevexec.RunHostCommand("docker", "context", "inspect", "-f", `{{ .Name }} {{ .Endpoints.docker.Host }}`)
+// initDockerCli sets up the Docker CLI client and initializes it
+// And sets DockerContext and DockerHost variables
+func initDockerCli() error {
+	var err error
+	// io.Discard is used to suppress stream output from DockerCli.DockerEndpoint().Host on error
+	DockerCli, err = dockerCliCommand.NewDockerCli(dockerCliCommand.WithErrorStream(io.Discard))
 	if err != nil {
-		return "", "", fmt.Errorf("unable to run 'docker context inspect' - please make sure Docker client is in path and up-to-date: %v", err)
+		return fmt.Errorf("unable to get Docker CLI client: %v", err)
```
rfay: Meaning of "docker CLI" has changed. It's pretty confusing anyway. Poor historical naming on their side I guess.

```suggestion
		return fmt.Errorf("newDockerCli() failed: %v", err)
```

pkg/dockerutil/dockerutils.go
```diff
@@ -171,75 +176,67 @@ func RemoveNetworkDuplicates(ctx context.Context, client *dockerClient.Client, n
...
 	}
-	contextInfo = strings.Trim(contextInfo, " \r\n")
-	util.Verbose("GetDockerContext: contextInfo='%s'", contextInfo)
-	parts := strings.SplitN(contextInfo, " ", 2)
-	if len(parts) != 2 {
-		return "", "", fmt.Errorf("unable to run split Docker context info %s: %v", contextInfo, err)
+	opts := dockerCliFlags.NewClientOptions()
+	if err := DockerCli.Initialize(opts); err != nil {
+		return fmt.Errorf("unable to initialize Docker CLI client: %v", err)
```
rfay:
```suggestion
		return fmt.Errorf("unable to DockerCli.Initialize(): %v", err)
```

pkg/dockerutil/dockerutils.go
```diff
@@ -620,6 +614,9 @@ func getSuggestedCommandForContainerLog(container *dockerContainer.Summary) (str
 
 // ContainerName returns the container's human-readable name.
 func ContainerName(container dockerContainer.Summary) string {
+	if len(container.Names) == 0 {
```
rfay, line 617: So you can have a container without a name? I wonder if I've seen some of those.

cmd/ddev/cmd/debug-dockercheck.go
```diff
@@ -41,18 +41,21 @@ var DebugDockercheckCmd = &cobra.Command{
...
 		}
-		util.Success("Using Docker context: %s (%s)", dockerutil.DockerContext, dockerutil.DockerHost)
-		util.Success("docker-compose: %s", versionInfo["docker-compose"])
 
-		dockerHost := os.Getenv("DOCKER_HOST")
-		if dockerHost != "" {
-			util.Success("Using DOCKER_HOST=%s", dockerHost)
-		}
+		util.Success("Using Docker context: %s", dockerutil.DockerContext)
 		dockerContext := os.Getenv("DOCKER_CONTEXT")
```
rfay, line 46: I didn't even remember that there was such a thing!

cmd/ddev/cmd/root.go
```diff
@@ -139,7 +139,7 @@ func init() {
 	_, err := dockerutil.GetDockerVersion()
 
 	if err != nil && !dockerutil.CanRunWithoutDocker() {
-		util.Failed("Could not connect to a Docker provider. Please start or install a Docker provider.\nFor install help go to: https://ddev.readthedocs.io/en/stable/users/install/docker-installation/")
+		util.Failed("Could not connect to a Docker provider. Please start or install a Docker provider.\nDetails: %v\nFor install help go to: https://ddev.readthedocs.io/en/stable/users/install/docker-installation/", err)
```
rfay: Let's add an FAQ and link to it here.  I can't imagine how many people get lost with contexts and can't find their way out. This error message assumes that the problem is not having a provider, but it can certainly be the context, so an FAQ could help a little with that.

rfay (approved):
I'm very glad that this didn't require as much of (my) brainpower as I feared. I know it required an enormous amount of yours, but luckily you have lots. It all seems to work fine for me and the code looks good and improved, and I think general approach is better, we don't have that ugly dependency on an unknown external resource (docker cli). 

I made small suggestions. Probably the most important is for the "can't connect to docker" error to link to a better troubleshooting resource, probably in FAQ, that explains contexts.

pkg/dockerutil/dockerutils.go
stasadev: This didn't happen to me, but I added it as a simple check to prevent a random panic.

cmd/ddev/cmd/root.go
stasadev: Updated docs: https://ddev--7189.org.readthedocs.build/en/7189/users/install/docker-installation/#troubleshooting-docker

rfay: Thx!

rfay:
Seems like it can be pulled when you think it's ready, thanks!

PR #7192, opened 2025-04-08
test: fix for magento2 quickstart and bats test, fixes #7191

### LINKED ISSUE #7191: Magento 2 Quickstart failing, "Search engine 'elasticsearch7' is not an available search engine"
The Magento 2 quickstart test is [failing](https://github.com/ddev/ddev/actions/runs/14334832011/job/40184703140#step:7:1857) with "Search engine 'elasticsearch7' is not an available search engine"

Could you take a look at this @rpkoller ?

rpkoller:
## The Issue

- #7191

See https://ddev--7192.org.readthedocs.build/en/7192/users/quickstart/#magento-2

### DIFF
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -20,8 +20,8 @@ teardown() {
   # mkdir ${PROJNAME} && cd ${PROJNAME}
   run mkdir ${PROJNAME} && cd ${PROJNAME}
   assert_success
-  # ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management --php-version=8.4
-  run ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management --php-version=8.4
+  # ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
+  run ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
   assert_success
 
   # mkdir -p .ddev/homeadditions/.composer
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -116,7 +116,7 @@ EOF
   run curl -sf https://${PROJNAME}.ddev.site/index.php/admin_ddev/
   assert_success
   assert_output --partial "Copyright &copy; 2025 Magento Commerce Inc. All rights reserved."
-  run curl -sf https://${PROJNAME}..ddev.site:5602/app/home#/
+  run curl -sf https://${PROJNAME}.ddev.site:5602/app/home#/
   assert_success
   assert_output --partial "<title>OpenSearch Dashboards</title>"
 }
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -39,6 +39,14 @@ teardown() {
 }
 EOF
 
+  # Copy the auth.json into var/composer_home for the deploying the sample data sep
+  run ddev exec "mkdir -p var/composer_home && cp ~/.composer/auth.json var/composer_home/auth.json"
+  assert_success
+
+  # run ddev mutagen sync
+  run ddev mutagen sync
+  assert_success
+
   # ddev add-on get ddev/ddev-opensearch
   run ddev add-on get ddev/ddev-opensearch
   assert_success
@@ -79,6 +87,14 @@ EOF
   run ddev magento setup:config:set --backend-frontname="admin_ddev" --no-interaction
   assert_success
 
+  # Add sample data for Magento
+  run ddev magento sampledata:deploy
+  assert_success
+
+  # Re-run the setup script to include the sample data
+  run ddev magento setup:upgrade
+  assert_success
+
   # validate ddev launch
   run bash -c "DDEV_DEBUG=true ddev launch /admin_ddev"
   assert_output "FULLURL https://${PROJNAME}.ddev.site/admin_ddev"
@@ -98,4 +114,7 @@ EOF
   run curl -sf https://${PROJNAME}.ddev.site/index.php/admin_ddev/
   assert_success
   assert_output --partial "Copyright &copy; 2025 Magento Commerce Inc. All rights reserved."
+  run curl -sf https://${PROJNAME}..ddev.site:5602/app/home#/
+  assert_success
+  assert_output --partial "<title>OpenSearch Dashboards</title>"
 }
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -107,6 +107,8 @@ EOF
   run curl -sf https://${PROJNAME}.ddev.site
   assert_success
   assert_output --partial "Copyright © 2013-present Magento, Inc. All rights reserved."
+  assert_output --partial "Here is what\`s trending on Luma right now"
+  assert_output --partial "title=\"Argus All-Weather Tank\""
   run curl -sfI https://${PROJNAME}.ddev.site/index.php/admin_ddev/
   assert_success
   assert_output --partial "server: nginx"
--- a/docs/tests/magento2.bats
+++ b/docs/tests/magento2.bats
@@ -20,8 +20,8 @@ teardown() {
   # mkdir ${PROJNAME} && cd ${PROJNAME}
   run mkdir ${PROJNAME} && cd ${PROJNAME}
   assert_success
-  # ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
-  run ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
+  # ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management --php-version=8.4
+  run ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management --php-version=8.4
   assert_success
 
   # mkdir -p .ddev/homeadditions/.composer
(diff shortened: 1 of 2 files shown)


JonasWijne:
the magento 2 docs state that elastic search support is dropped starting 2.4.8 in favor of opensearch,

example
```bash 
export MAGENTO_HOSTNAME=my-magento2-site
mkdir ${MAGENTO_HOSTNAME} && cd ${MAGENTO_HOSTNAME}
# Configure DDEV for Magento 2 with PHP 8.4
ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management --php-version=8.4
# Magento >=2.4.8 requires OpenSearch
ddev add-on get ddev/ddev-opensearch
ddev start
ddev composer create --repository https://repo.magento.com/ magento/project-community-edition
rm -f app/etc/env.php

ddev magento setup:install --base-url="https://${MAGENTO_HOSTNAME}.ddev.site/" \
    --cleanup-database --db-host=db --db-name=db --db-user=db --db-password=db \
    --search-engine=opensearch --opensearch-host=opensearch --opensearch-port=9200 \
    --admin-firstname=Magento --admin-lastname=User --admin-email=user@example.com \
    --admin-user=admin --admin-password=Password123 --language=en_US

ddev magento deploy:mode:set developer
ddev magento module:disable Magento_TwoFactorAuth Magento_AdminAdobeImsTwoFactorAuth
ddev config --disable-settings-management=false
# Change the backend frontname URL to /admin_ddev
ddev magento setup:config:set --backend-frontname="admin_ddev" --no-interaction
# Login using `admin` user and `Password123` password
ddev launch /admin_ddev
```

suggestion: https://github.com/jonaswijne/ddev/commit/085071a90bddcfcc3015ede69c1bdac7e8c82745

rpkoller:
@JonasWijne thanks for the pointer. was unaware about the recommended switch to opensearch. we only noticed the failing tests. i already applied your proposed changes locally. will test the new instructions manually first and then also extend the bats test to actually test the opensearch frontend in some  way as well. updated PR should be up soon

rpkoller:
and one question would it make sense instead of using the flag on ddev config to set the php version to 8.4 to update the php version in the magento2 project type instead?

docs/tests/magento2.bats
```diff
@@ -39,8 +39,16 @@ teardown() {
...
 
-  # ddev add-on get ddev/ddev-elasticsearch
-  run ddev add-on get ddev/ddev-elasticsearch
+  # Copy the auth.json into var/composer_home for the deploying the sample data sep
+  run ddev exec "mkdir -p var/composer_home && cp ~/.composer/auth.json var/composer_home/auth.json"
+  assert_success
+
+  # run ddev mutagen sync
+  run ddev mutagen sync
+  assert_success
```
rfay, line 48: We don't need this because we took action inside the container.

```suggestion
```

docs/tests/magento2.bats
```diff
@@ -91,11 +107,16 @@ EOF
...
+  assert_output --partial "Here is what\`s trending on Luma right now"
+  assert_output --partial "title=\"Argus All-Weather Tank\""
   run curl -sfI https://${PROJNAME}.ddev.site/index.php/admin_ddev/
   assert_success
   assert_output --partial "server: nginx"
   assert_output --partial "HTTP/2 200"
   run curl -sf https://${PROJNAME}.ddev.site/index.php/admin_ddev/
   assert_success
   assert_output --partial "Copyright &copy; 2025 Magento Commerce Inc. All rights reserved."
+  run curl -sf https://${PROJNAME}..ddev.site:5602/app/home#/
```
stasadev, line 119:
```suggestion
  run curl -sf https://${PROJNAME}.ddev.site:5602/app/home#/
```

docs/tests/magento2.bats
```diff
@@ -39,8 +39,16 @@ teardown() {
 }
 EOF
 
-  # ddev add-on get ddev/ddev-elasticsearch
-  run ddev add-on get ddev/ddev-elasticsearch
+  # Copy the auth.json into var/composer_home for the deploying the sample data sep
+  run ddev exec "mkdir -p var/composer_home && cp ~/.composer/auth.json var/composer_home/auth.json"
```
rfay, line 43: Has the project been started? Otherwise this couldn't work.

docs/tests/magento2.bats
```diff
@@ -39,8 +39,16 @@ teardown() {
...
+  assert_success
+
+  # run ddev mutagen sync
+  run ddev mutagen sync
+  assert_success
+
+  # ddev add-on get ddev/ddev-opensearch
+  run ddev add-on get ddev/ddev-opensearch
```
rfay, line 51: I think this should have been done much higher, before we even start the project. Same in the quickstart itself. Do this before ever starting project.

rfay (changes requested):
Looking great, thanks for this!

docs/tests/magento2.bats
rpkoller: i still mix up where it is necessary to run a sync. :/

docs/tests/magento2.bats
rpkoller: in the quickstart and in the bats test the addon is already installed before the project is started?

docs/tests/magento2.bats
rpkoller: good point. i'll move it after the ddev start

docs/tests/magento2.bats
rfay: It would make sense to install the add-on before the project is started, or else to do a `ddev restart` after installing it. If you do a `ddev restart` after installing, you may have to put that `ddev exec` after the restart.

rpkoller: i dont understand, it is already that way, or do i understand something wrong here?  the part from the quickstart 
```
ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
ddev add-on get ddev/ddev-opensearch
ddev start
```

the part from the bats test: 

```
  run ddev config --project-type=magento2 --docroot=pub --upload-dirs=media --disable-settings-management
  assert_success

  # mkdir -p .ddev/homeadditions/.composer
  mkdir -p ./.ddev/homeadditions/.composer

  # add the env variable credentials to auth.json
  cat <<EOF > .ddev/homeadditions/.composer/auth.json
{
    "http-basic": {
        "repo.magento.com": {
            "username": "${MAGENTO2_PUBLIC_ACCESS_KEY}",
            "password": "${MAGENTO2_PRIVATE_ACCESS_KEY}"
        }
    }
}
EOF

  # ddev add-on get ddev/ddev-opensearch
  run ddev add-on get ddev/ddev-opensearch
  assert_success

  # ddev start -y
  run ddev start -y
  assert_success
```

docs/tests/magento2.bats
rpkoller: and in this case it is created inside the container and also needed to be available only in the container providing the credentials for composer(which is also only available in the container)?

docs/tests/magento2.bats
rfay: I guess I was just confused.

docs/tests/magento2.bats
```diff
@@ -39,12 +39,16 @@ teardown() {
...
   assert_success
 
   # ddev start -y
   run ddev start -y
+  assert_success
+
+    # Copy the auth.json into var/composer_home for the deploying the sample data sep
+  run ddev exec "mkdir -p var/composer_home && cp ~/.composer/auth.json var/composer_home/auth.json"
```
stasadev, line 51:
```
   -- command failed --
   status : 1
   output : Failed to create project: '/home/stas/tmp/my-magento-site.9A6SFh/my-magento-site/var' is not allowed to be present. composer create-project needs to be run on a clean/empty project with only the following paths: [app app/etc app/etc/.gitignore app/etc/env.php pub] - please clean up the project before using 'ddev composer create-project'
   --
``

stasadev: I think this should be done after `ddev composer create`, and why do we only have this in the test?

rpkoller: in the quickstart the sample data is only optional at the end of the quickstart but idea was to ensure that this optional step is working by installing the sample data in the bats test.

but the detail i dont understand, i would have expected if the content would be copied into /var/www/html that composer would raise an error but in this case the auth.json is copied into  var/composer_home/. that way the error comes sort of unexpected?  ( am currently running the bats test locally to see if i run into that error as well.

stasadev: > in the quickstart the sample data is only optional

Now I see it, thanks.

> i would have expected if the content would be copied into /var/www/html that composer would raise an error but in this case the auth.json is copied into var/composer_home/

it's copied to `/var/www/html/var/composer_home`.

rpkoller: > it's copied to /var/www/html/var/composer_home, just move this cp after ddev composer create to see if it helps.

i misread and missed the missing `/` at the beginning of `var/composer_home/`, i read it as if the `/`was there and the `auth.json` would be copied to the system root (also somehow blended out the mkdir in the first step). but that way with the `composer_home` created inside `/var/www/html`the copy statement in the second step goes into the project root. nooooow i understand things again. yep that way i also understand the necessary moving of the copy statement after the composer create. my local test now also runs through. the previous test after placing the cp after the start and before the composer create took forever probably hanging somewhere in the process without a note or exit. will push an updated version in a minute.

stasadev (approved):
Tested it locally, works for me, thank you!

rfay (approved):
Thanks so much for the excellent work on this!

PR #7199, opened 2025-04-10
feat: add DDEV_APPROOT variable to web container and updates documentation, fixes #7198

### LINKED ISSUE #7198: Add DDEV_APPROOT web container variable equivalent of host

There is an environment variable for the host that will give the absolute path to the project root directory. It would be great to have the equivalent in web so that we can use the variable in custom commands in the web container.

### Describe your solution

Add `DDEV_APPROOT="/var/www/html"` to environment variables for the web service. (PR incoming)

### Describe alternatives

The alternative is for the user to hard code that value in custom command files or create their own variable.

### Additional context

_No response_

shelane:
## The Issue

- #7198 

Adds the variable to the web container.

```
ddev ssh
echo $DDEV_APPROOT
```

You should see "/var/www/html" as the output (without the quotes of course)

### DIFF
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -182,7 +182,7 @@ services:
     - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
     - COREPACK_HOME=/mnt/ddev-global-cache/corepack
     - DOCROOT=${DDEV_DOCROOT}
-    - DDEV_APPROOT="/var/www/html"
+    - DDEV_APPROOT=/var/www/html
     - DDEV_COMPOSER_ROOT
     - DDEV_DATABASE
     - DDEV_DOCROOT
--- a/pkg/ddevapp/app_compose_template.yaml
+++ b/pkg/ddevapp/app_compose_template.yaml
@@ -182,6 +182,7 @@ services:
     - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
     - COREPACK_HOME=/mnt/ddev-global-cache/corepack
     - DOCROOT=${DDEV_DOCROOT}
+    - DDEV_APPROOT="/var/www/html"
     - DDEV_COMPOSER_ROOT
     - DDEV_DATABASE
     - DDEV_DOCROOT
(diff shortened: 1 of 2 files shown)


tyler36 (approved):
Confirmed. Works as intended.

## Test

1. Get current DDEV
   
    ```shell
    $ ddev -v
    ddev version v1.24.4
    ```


1. Start existing project

    ```shell
    $ ddev start
    ...
    Your project can be reached at https://test-variable.ddev.site
    ```

1. Check variable is set in container

    ```shell
    $ ddev ssh
    $ echo $DDEV_APPROOT
    
    $ 
    ```

### PR DDEV

1. Check PR DDEV version

    ```shell
    $ ddev -v     
    ddev version v1.24.4-13-g9252322e1
    ```

1. Start existing project

    ```shell
    $ ddev start
    ...
    Your project can be reached at https://test-variable.ddev.site
    ```

1. Check variable is set in container

    ```shell
    $ ddev ssh
    $ echo $DDEV_APPROOT
    "/var/www/html"
    $
    ```

rfay (approved):
It seems like we've needed this for a long time. My only minor reservation is if there's a way to use `DDEV_APPROOT` wrong in the wrong context.  But I think it should be fine.

pkg/ddevapp/app_compose_template.yaml
```diff
@@ -182,6 +182,7 @@ services:
     - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
     - COREPACK_HOME=/mnt/ddev-global-cache/corepack
     - DOCROOT=${DDEV_DOCROOT}
+    - DDEV_APPROOT="/var/www/html"
```
stasadev, line 185:
```
ddev exec 'echo $DDEV_APPROOT/test'
"/var/www/html"/test
```

But I expected to see it without quotes.

```suggestion
    - DDEV_APPROOT=/var/www/html
```

stasadev, line 185:
```
$ ddev exec 'echo $DDEV_APPROOT/test'
"/var/www/html"/test
```

But I expected to see it without quotes.

```suggestion
    - DDEV_APPROOT=/var/www/html
```

stasadev (changes requested):

stasadev (approved):
I tested `DDEV_APPROOT` in host and web commands, and can confirm that it shows the correct value in both cases.

rfay:
Thanks!

PR #7200, opened 2025-04-11
docs: Improve buildkite windows setup script run [skip buildkite]

rfay:
## The Issue

I was rebuilding tb-wsl-14 and had just a little trouble, fixed here.


stasadev (approved):
I wonder why Windows uses such weird commands, they seem almost impossible to remember.

PR #7203, opened 2025-04-12
test: make the drupal cms bats test a bit more robust and trustworthy

rpkoller:
## The Issue

So far it was only tested if the installer page was loaded. after that much could break. the PR adds a drush site install and a drush recipe to apple the blog recipe. that way it is possible to test if the site is actually installed properly, is able to be launched and if the recipes was correctly applied. 

- #<issue number>


rfay (approved):
Thanks!

Issue #7204, opened 2025-04-12
Disable xhgui/xhprof via ENV variable


sreichel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Thanks for adding xhgui to core ❤ 

With the add-on from @tyler36 i had to set an ENV var to disable xhgui when running PHPUnit to prevent errors. 

- see https://github.com/OpenMage/magento-lts/pull/4325
- see https://github.com/OpenMage/magento-lts/pull/4325/files#diff-8b8fea56d71e37beefdeb85189a18b8d3d1567e913f89660aa2abc8d210d0f36

Now everything works out  of the box, but phpunit runs are also logged. Is it possible to add possibility to turn off profiling via ENV?

It could also help to not spam reports with cron tasks. :)

### Describe your solution

Detect `DISABLE_XHGUI` env, that defaults to `0`  (or something like that),

rfay:
Hi @sreichel - When you do a `ddev start`, XHGui is not enabled until you enable it.  The container is not running, and xhprof is not enable. 

Can you be more specific about the problem is and exactly and how to recreate it?

sreichel:
I'd like to disable profiling for PhpUnit tests and cron tasks, but have it enabled for all other requests.

I don't want to manually turn xhgui on/off everytime.

My idea was to add an env bar to xhgui/xhprof container that can be overwritten locally (e.g. from phpunit config)

rfay:
I'm not sure exactly how you'd do that. Essentially it's enabling or disabling the `xhprof` PHP extension. Can you say why you would have it enabled when you weren't profiling explicitly?

sreichel:
For the xhgui-addon it was necessary to manually include the xhgui collector.

See linked PR (etc/includes/ddev.xhgui.php) and check for 'NO_XHGUI'.

I don't know if it works to add ...

- env 'ENABLE_XHGUI=1'
- only run data collector if set to 1
- allow to override from project

... to xhgui.

To disable it for phpunit, I could set ENABLE_XHGUI=0 in its config.

rfay:
Please experiment. This new feature is mostly all adapted from the add-on.

sreichel:
```php
if (getenv('IS_DDEV_PROJECT') && getenv('NO_XHGUI') !== '1') {
    if (file_exists($xhgui = '/mnt/ddev_config/xhgui/collector/xhgui.collector.php')) {
        require_once $xhgui;
    }
[...]

PR #7206, opened 2025-04-12
feat: disable xhgui collector via env

sreichel:
## The Issue

- #7204

It should be possible to set XHGUI_ENABLED to 0 in phpunit config to exclude profiling.

Add to phpunit.dist.xml

```
    <php>
        <env name="XHGUI_ENABLED" value="0"/>
    </php>
```


containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/xhprof/xhprof_prepend.php
```diff
@@ -12,6 +12,10 @@
 }
 
 $collector_php = "/usr/local/xhgui.collector/xhgui.collector.php";
+if (getenv('XHGUI_ENABLED') === '0') {
+    # some output?
+    return;
+}
```
rfay, line 18: How does XHGUI_ENABLED get set to 0 (or to any other value)? I didn't find any reference to that variable anywhere on the internet except in laravel-xhgui - https://github.com/search?q=repo%3ARitvarsZ%2Flaravel-xhgui%20XHGUI_ENABLED&type=code

To take effect, this requires a build of the ddev-webserver, so won't work in casual testing here.

sreichel: > How does XHGUI_ENABLED get set to 0

From project. See phpunit.dist.xml example from above.

> I didn't find any reference to that variable anywhere on the internet

This var does not exist. It only acts as a flag that can be set.

rfay: The problem is not how to set an environment variable, it's what does the environment variable do. The link you gave doesn't say anything about XHGUI_ENABLED. 

The question is... what are valid variables of the environment variable? How does it get set? When does it get set? What does it do and why? And of course, why are we doing it. You want XHGui to stop collecting, but xhprof to continue being enabled?

rfay:
Moving to draft until this is actionable.

rfay:
Please edit the OP to explain exactly what you're trying to do and why; docs will need to be added as well.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/xhprof/xhprof_prepend.php
sreichel: > The problem is not how to set an environment variable, it's what does the environment variable do.

It does nothing. It can be named how you like. DDEV should only check if it exists and is set to 0. In this case do not load xhguis data collector.

So ...

#### what are valid variables of the environment variable?
only 0, to disable

#### How does it get set?
e.g from phpunit config (maybe from cli)

#### When does it get set?
when running phpunit

####What does it do and why?
When set to 0, xhgui does not collect data. You will see all your normal post/get request, but no profiling for phphunit tests

#### You want XHGui to stop collecting, but xhprof to continue being enabled?
Yes. It should be enabled, but i dont want to see logs from phpunit (nor cron).

rfay:
I think I'd prefer the opposite, something like `DDEV_XHGUI_DISABLED=true` for behavior like this. 

You can experiment with your idea by overwriting  `/usr/local/bin/xhprof/xhprof_prepend.php` with a .`ddev/web-build/Dockerfile.xhgui_disabler`

Something like
```
COPY xhprof_prepend.php /usr/local/bin/xhprof/xhprof_prepend.php
```
and provide the xhprof_prepend.php in .ddev/web-build

Try it out and tune it in and see what you get.

stasadev (commented):
I think the place where the env variable should be set is here:

https://github.com/ddev/ddev/blob/c382808a854acd41411ea7a65e291e9a3a01f5e5/containers/ddev-webserver/ddev-webserver-base-files/usr/local/xhgui.collector/xhgui.collector.config.php#L32-L35

See https://github.com/perftools/php-profiler#configure-profiling-rate

---

For example, we can use:

```php
  'profiler.enable' => function () {
    if (in_array(getenv('DDEV_XHGUI_PROFILER_DISABLED'), ['true', '1'])) {
      return false;
    }
    return true;
  },
```

---

Test:

```php
<?php

function isProfilerEnabled() {
  if (in_array(getenv('DDEV_XHGUI_PROFILER_DISABLED'), ['true', '1'])) {
    return false;
  }
  return true;
}

var_dump(isProfilerEnabled());

putenv('DDEV_XHGUI_PROFILER_DISABLED=1');

var_dump(isProfilerEnabled());

putenv('DDEV_XHGUI_PROFILER_DISABLED=0');

var_dump(isProfilerEnabled());

putenv('DDEV_XHGUI_PROFILER_DISABLED=true');

var_dump(isProfilerEnabled());
```

```
bool(true)
bool(false)
bool(true)
bool(false)
```

rfay:
Will need doc as well.

sreichel:
Looks great.

@stasadev want to create a new PR? Your idea. Or update this one.

Can't work on it atm.

stasadev:
It's okay to update this PR, I'll work on it tomorrow.

stasadev:
@sreichel, if I read this PR correctly:

- https://github.com/OpenMage/magento-lts/pull/4325

`perftools/php-profiler` is required at the app level, and loading for XHProf is done in the app itself.

(I'm still not sure how you managed to succeed when DDEV loads `xhprof_prepend.php` in addition to what you have.)

---

But DDEV loads `perftools/php-profiler` in `xhprof_prepend.php` *before* running the application when `xhprof` is enabled:

https://github.com/ddev/ddev/blob/ae78e8fabc92414fa674b6431c95dd41e7b2ccb4/containers/ddev-php-base/ddev-php-files/etc/php/8.3/mods-available/xhprof.ini#L1-L3

This means the environment variable is already read before it can be set in `phpunit.xml`:

```xml
<php>
    <env name="DDEV_XHGUI_PROFILER_DISABLED" value="true"/>
</php>
```

As a result, `ddev exec phpunit` still triggers a run.

The run is not triggered with this (but only for the `phpunit` command itself, not for HTTP requests inside the tests):
```
ddev exec 'DDEV_XHGUI_PROFILER_DISABLED=true phpunit'
```
But it makes no sense to add a variable for this if people can just disable it before running `phpunit`:
```
ddev xhgui off && ddev exec phpunit
```

stasadev:
I pushed new Docker images here (but I used `[skip ci]`, so the binaries are not generated in this PR).

To demonstrate, checkout this branch locally, and build the binary with `make`.

I used Laravel quickstart:

```
ddev config global --xhprof-mode=xhgui

mkdir my-laravel-site && cd my-laravel-site
ddev config --project-type=laravel --docroot=public
ddev start
ddev composer create-project "laravel/laravel:^12"

ddev xhgui on
ddev xhgui launch

# see +1 run in XHGui
ddev launch

# see +1 run in XHGui
ddev exec phpunit

# see no new runs in XHGui
ddev exec 'DDEV_XHGUI_PROFILER_DISABLED=true phpunit'
```

Add this env to `phpunit.xml`
```xml
<php>
  ...
  <env name="DDEV_XHGUI_PROFILER_DISABLED" value="true"/>
</php>
```

No effect, +1 run in XHGui, env is set after it's already read:
```
ddev exec phpunit
```

stasadev:
@sreichel, as stated in [the comment](https://github.com/ddev/ddev/pull/7206#issuecomment-2970114134), XHGui logic loads too early, so we cannot implement this environment variable. Closing.

Issue #7085, opened 2025-03-14
Make `ddev debug download-images` sensitive to project configuration / addons


jonhattan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

At present `ddev debug download-images` downloads a fixed list of images:

```
REPOSITORY                         TAG       IMAGE ID       CREATED         SIZE
ddev/ddev-dbserver-mariadb-10.11   v1.24.3   08f1db6a1f86   2 weeks ago     531MB
ddev/ddev-traefik-router           v1.24.3   a6501e545198   2 weeks ago     230MB
ddev/ddev-ssh-agent                v1.24.3   54e83ef8180b   2 weeks ago     136MB
ddev/ddev-webserver                v1.24.3   58b66a54716b   2 weeks ago     1.18GB
ddev/ddev-utilities                latest    09b7b0fddb95   5 months ago    68.3MB
busybox                            stable    83ef53509aa5   22 months ago   4.26MB
```

In projects based on mysql or another db backend, we want mysql-8 instead if mariadb-10.11.

### Describe your solution

Check `.ddev/config.yaml` (if present) and pick the correct db image.

### Describe alternatives

Alternative and more ambitious solution: pick all images used in a project.

Roughly I use this in shell to get the list of images and download it by other means: 

```
ddev debug compose-config |grep -i image: | grep -v built | awk '{print $2}' | sort | uniq -u > myproject.list
```

rfay:
It has potential! I think to make this happen you'll likely have to undertake it as a PR. Note that a dummy `ddev start` will get you what you want, I've had to do that in a couple of tests.

rfay:
This is done, right @stasadev ?

stasadev:
Yes, fixed in:

- #7483

PR #7217, opened 2025-04-18
feat: use docker-compose to parallel download images, fixes #7163

### LINKED ISSUE #7163: Enable parallel download via docker compose

On first boot (or when docker images are pruned) the docker images are downloaded in parallel.

docker-compose could be used to pull them in parallel:
1. create temporary docker compose file containing needed images
2. run docker-compose -f tmp.yml pull
3. continue with the rest of the flow

### Describe your solution

.

### Describe alternatives

_No response_

### Additional context

_No response_

AkibaAT:
## The Issue

- #7163

DDEV currently preloads all required images via individual docker pull commands. 

This PR replaces this with a docker compose call, enabling parallel download of multiple images.

Remove DDEV images from local Docker
Run ddev debug download-images

## Automated Testing Overview

-

## Release/Deployment Notes

-

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"fmt"
 	"os"
+	oexec "os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
@@ -1855,12 +1856,28 @@ func PullBaseContainerImages() error {
 	images := []string{ddevImages.GetWebImage(), versionconstants.BusyboxImage, versionconstants.UtilitiesImage}
 	images = append(images, FindNotOmittedImages(nil)...)
 
-	for _, i := range images {
-		err := dockerutil.Pull(i)
-		if err != nil {
-			return err
-		}
-		util.Debug("Pulled image for %s", i)
+	composeBinaryPath, err := globalconfig.GetDockerComposePath()
+	if err != nil {
+		return fmt.Errorf("failed to get docker-compose path: %v", err)
+	}
+
+	// Build compose YAML in-memory
+	var buf bytes.Buffer
+	buf.WriteString("services:\n")
+	r := strings.NewReplacer(":", "-", "/", "-")
+	for _, img := range images {
+		serviceName := r.Replace(img)
+		buf.WriteString(fmt.Sprintf("  %s:\n    image: %s\n", serviceName, img))
+		util.Debug("Adding service %s to pull image %s", serviceName, img)
+	}
+
+	cmd := oexec.Command(composeBinaryPath, "-f", "-", "pull")
+	cmd.Stdin = &buf
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err = cmd.Run(); err != nil {
+		return fmt.Errorf("failed to pull Docker images: %v", err)
 	}
 
 	return nil
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -5,7 +5,6 @@ import (
 	"embed"
 	"fmt"
 	"os"
-	oexec "os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
@@ -1269,11 +1268,6 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		return err
 	}
 
-	err = PullBaseContainerImages()
-	if err != nil {
-		util.Warning("Unable to pull Docker images: %v", err)
-	}
-
 	if !nodeps.ArrayContainsString(app.GetOmittedContainers(), "db") {
 		// OK to start if dbType is empty (nonexistent) or if it matches
 		if dbType, err := app.GetExistingDBType(); err != nil || (dbType != "" && dbType != app.Database.Type+":"+app.Database.Version) {
@@ -1453,7 +1447,7 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 	}
 
 	// This needs to be done after WriteDockerComposeYAML() to get the right images
-	err = app.PullContainerImages()
+	_, err = app.PullContainerImages()
 	if err != nil {
 		util.Warning("Unable to pull Docker images: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


rfay (changes requested):
I know you're not done with this yet, but I think it should be using ComposeCmd() to do its work.

AkibaAT:
I saw #7085 after this, so I'll likely rework this to cover that as well.

rfay:
I'm not entirely sure this idea doesn't already solve #7085 since you're using `.ddev/.ddev-docker-compose-full.yaml`

AkibaAT:
This draft doesn't yet, it's currently still the same image list as before. But yeah, using the docker compose file instead would be the goal.

stasadev:
Thank you, I didn't know you could use a buffer for this.

This is cool, but it's a little confusing to see errors when I already have the images and disconnect from the Internet:

```
$ ddev start
...
[+] Pulling 5/5
 ✘ ddev-ddev-ssh-agent-v1.24.4 Error      Get "https://registry-1.docker.io/v2/":...                                    0.1s 
 ✘ ddev-ddev-webserver-v1.24.4 Error      Get "https://registry-1.docker.io/v2/":...                                    0.1s 
 ✘ ddev-ddev-traefik-router-v1.24.4 Error Get "https://registry-1.docker.io/...                                         0.1s 
 ✘ busybox-stable Error                   Get "https://registry-1.docker.io/v2/": dial tcp: lo...                       0.1s 
 ✘ ddev-ddev-utilities Error              Get "https://registry-1.docker.io/v2/": dial tc...                            0.1s 
Error response from daemon: Get "https://registry-1.docker.io/v2/": dial tcp: lookup registry-1.docker.io: Temporary failure in name resolution
Unable to pull Docker images: failed to pull Docker images: exit status 1 
Building project images....
Project images built in 2s.
...
```

AkibaAT:
Hmmm... alright, I had not considered the behaviour when offline. Testing this, the current version handles this better, although the debug log claim isn't quite right either, claiming that the images got pulled.

```
(base) ➜  d11 DDEV_DEBUG=true ddev debug download-images
2025-04-21T22:34:54.541 Pulled image for ddev/ddev-webserver:v1.24.4
2025-04-21T22:34:54.568 Pulled image for busybox:stable
2025-04-21T22:34:54.583 Pulled image for ddev/ddev-utilities
2025-04-21T22:34:54.602 Pulled image for ddev/ddev-ssh-agent:v1.24.4
2025-04-21T22:34:54.625 Pulled image for ddev/ddev-traefik-router:v1.24.4
2025-04-21T22:34:54.672 Pulled ddev/ddev-dbserver-mariadb-10.11:v1.24.4
Successfully downloaded DDEV images
```

Properly handling this, we'd either need to detect what images are already present and not pull them at all (what about updates for moving targets like busybox:stable then?), or maybe detect if the user is currently offline, and provide a separate error message for that case. Or suppress all direct output and instead parse the results?

There is a `--ignore-pull-failures` flag according to https://docs.docker.com/reference/cli/docker/compose/pull/, but that would then ignore any sort of pull error, also for images that the user doesn't already have.

stasadev:
> what about updates for moving targets like busybox:stable then?

We don't have this behavior right now, because we check for the existing image, and pull nothing if it exists (but I would like it to do an actual pull, so this is good.)

> maybe detect if the user is currently offline, and provide a separate error message for that case.

Good idea, and we have a `globalconfig.IsInternetActive()` function for this check.

rfay:
I'm not sure how often busybox changes, but there's any reason to actually use `busybox:stable`, it's just a random utility image. We could use 1.37.0, for example, instead of `stable`

AkibaAT:
Changed the logic to pull in all project images if there's a valid project context, and otherwise falls back to the base list (minus MariaDB).
Internet connection check added, outputs a warning `Internet connection not detected, skipping Docker image pull. Some images may be missing.` if there's no connection, but doesn't interrupt the execution.

rfay (commented):
I don't understand why this doesn't just use existing ComposeCmd (with `pull` instead of `exec`, just like here: https://github.com/ddev/ddev/blob/9bb78d5ba819f3f9a1907995eb9e07de79d67417/pkg/ddevapp/ddevapp.go#L1088

That should be super trivial, with the code already written. That would pull all images in the "full" yaml

AkibaAT:
Try to run a pull on the full yaml in the shell.

Example: `docker compose -f .ddev-docker-compose-full.yaml pull`
Result:
```
 ✔ redis Pulled                                                                                                                                                                                                                                                                    1.4s
 ! web Warning  failed to resolve reference "docker.io/ddev/ddev-webserver:v1.24.4-fvn-li-built": docker.io/ddev/ddev-webserver:v1.24.4-fvn-li-built: not found                                                                                                                    1.3s
 ! db Warning   failed to resolve reference "docker.io/library/postgres:17-fvn-li-built": docker.io/library/postgres:17-fvn-li-built: not found                                                                                                                                    1.3s
WARNING: Some service image(s) must be built from source by running:
    docker compose build web db
2 errors occurred:
        * Error response from daemon: failed to resolve reference "docker.io/ddev/ddev-webserver:v1.24.4-fvn-li-built": docker.io/ddev/ddev-webserver:v1.24.4-fvn-li-built: not found
        * Error response from daemon: failed to resolve reference "docker.io/library/postgres:17-fvn-li-built": docker.io/library/postgres:17-fvn-li-built: not found
```

AkibaAT:
Additionally, even if we would take the full YAML and manipulate it the same way that FindAllImages does, we'd only be pulling 2 images by default, the one for web, and the one for the db, missing all the images that are required outside the project's full YAML:
- busybox
- ddev/ddev-ssh-agent
- ddev/ddev-traefik-router
- ddev/ddev-utilities

Which is why I'm building a YAML based on FindAllImages, to get all of them together.

rfay:
IMO it would be better to do the one-off images separately, super easy, already done that way. There's also the issue of things like codespaces/gitpod, where the router and probably ddev-ssh-agent don't need to be pulled. IMO ddev-utilities doesn't actually ever need to be pulled until needed. (Now that I think about it, we could use ddev-utilities everywhere we use busybox and drop busybox)

pkg/ddevapp/ddevapp.go
```diff
@@ -1826,44 +1821,72 @@ func (app *DdevApp) Restart() error {
...
+		return false, fmt.Errorf("failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
 
-// PullBaseContainerImages pulls only the fundamentally needed images so they can be available early.
-// We always need web image, busybox, and ddev-utilities for housekeeping.
-func PullBaseContainerImages() error {
-	images := []string{ddevImages.GetWebImage(), versionconstants.BusyboxImage, versionconstants.UtilitiesImage}
-	images = append(images, FindNotOmittedImages(nil)...)
+	// Create docker-compose.yaml file
```
rfay, line 1863: Sorry, I object to this, don't want another random file being created, and certainly not one with a generic name like this.

rfay (changes requested):
Please use the .ddev/.ddev-docker-compose-full.yaml using ComposeCmd, and then an extra pull for the one-off's. 

Thanks! Sorry I didn't give more guidance on this to start with.

pkg/ddevapp/ddevapp.go
AkibaAT: The pull on docker-compose-full doesn't work, as explained further up.

AkibaAT: The temporary YAML gets put into a temp directory: `tmpDir, err := os.MkdirTemp("", "ddev-pull-images-")`. I'd prefer to work completely in-memory, like an in earlier version of this PR, but dockerutil.ComposeCmd, which contains logic we do want to use here, requires a file.

rfay:
Just a note here on why we use `docker-compose` and why we use a private version of it. 

* Early on in compose v2 they broke lots of things with every release, to where we just had to "vendor" it to provide any reliability.
* If we were to use `docker compose` instead of our private `docker-compose` then we'd be using the plugin that the user provided, which we have no control of.
* Our overall strategy is not to make changes to the user's machine (like requiring specific versions of docker cli or compose plugin). In general, we try to have the user explicitly make any changes of this type, for example, with `mkcert -install`. The so-far unavoidable exception is editing the hosts file when people can't resolve `*.ddev.site` or use a non-ddev.site URL.

stasadev:
See:

- #7483

PR #7228, opened 2025-04-24
refactor: check for docroot/index.* on start, fixes #7157

### LINKED ISSUE #7157: Pre-check existence of index.php in docroot on `ddev start` and warn

People often get a 403 when they have the wrong docroot or no index.php in the docroot. We have this in our FAQ.

### Describe your solution

@dalin suggested that we do a check-and-warn during `ddev start` if there is no index.php in the specified docroot. We could save people a lot of trouble.

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

- #7157

On start, check to see if docroot and index.* exist - or warn. Don't do it on generic webserver_type.

- [ ] Rename or remove docroot, start
- [ ] misconfigure `docroot`, start
- [ ] Check with generic webserver_type, start

## Automated Testing Overview

- nothing added

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1268,6 +1268,8 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		return err
 	}
 
+	warnMissingDocroot(app)
+
 	err = PullBaseContainerImages()
 	if err != nil {
 		util.Warning("Unable to pull Docker images: %v", err)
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Awesome!

I wish it was more noticeable (but it probably doesn't work everywhere):

```
echo -e "\033[5mWARNING\033[0m"
```

warning

rfay:
image

stasadev (approved):
Thank you!

I'm used to seeing orange in DDEV, so it’s refreshing to have something different.

rfay:
We may find that this is confusing to people doing a `ddev composer create-project` or a `ddev start` before the create-project. If so maybe we can figure out what to do then... Perhaps the warning gives enough context for people to understand, but most people don't read the whole message. 

I think the [FAQ](https://ddev.readthedocs.io/en/stable/users/usage/faq/#why-do-i-get-a-403-or-404-on-my-project-after-ddev-launch) is probably still valid and useful, but perhaps less people will need it now.

PR #7229, opened 2025-04-24
build: Use some artifacts that we control building windows installer (nsis), fixes #7086

### LINKED ISSUE #7086: Move nsis-required artifacts to a repository that we control
We currently use some NSIS-related artifacts that we don't control that come from Sourceforge of all places.

We should at least have more control of them. There are also a couple that come from @gilbertsoft 's repositories, we should fork those and get them into ddev.

These are for the Windows installer.

As @stasadev says, we can at least move these to a Github repo we have control of, so if they suddenly go away we're not left with a rebuild.

https://github.com/ddev/ddev/blob/bc338507e529a82cb23d7605651a0ffc0484eab1/.ci-scripts/nsis_setup.sh#L21-L27

rfay:
## The Issue

- #7086 

We discovered that our NSIS Windows installer build was completely dependent on a couple of repos that probably aren't alive any more, and we have no way to recreate them.

* Fork the repos
* Unlink the forks
* Recreate the releases
* Change the link in our build code.

* Look at the resultant Windows installer
* Look at the build output


rfay:
Why didn't it build artifacts?

rfay:
- [ ] Test the installer on Windows machine and call it good.

rfay:
Installer worked fine.

PR #2519, opened 2020-09-22
Introduce help description for arguments in custom commands [skip ci][ci skip]

gilbertsoft:

PR #7232, opened 2025-04-25
feat: allow YAML for custom command annotations, fixes #2638

### LINKED ISSUE #2638: Idea: Support --- yaml frontmatter in custom commands for Annotations
Just an idea that I had while parsing the custom commands docs.

It would be nicer to write:

```bash
#!/usr/bin/env bash

## Frontmatter
: <<'
---'
description: 'Run MySQLWorkbench against current db'
usage: mysqlworkbench
example: 'ddev mysqlworkbench'
flags:
  flag:
    usage: 'sets the flag option'
---

echo "Hello"
```

The `: <<'---'` is a trick to redirect the output to an empty command.

The following is also possible:

```bash
#!/usr/bin/env bash

## Frontmatter
: <<---
description: 'Run MySQLWorkbench against current db'
usage: mysqlworkbench
example: 'ddev mysqlworkbench'
flags:
  flag:
    usage: 'sets the flag option'
---

echo "Hello"
```

GuySartorelli:
## The Issue

- #2638

Annotations for custom commands (especially for flags) can get pretty long since they have to be on a single line.
Allowing annotations to be defined in YAML would resolve this and give clear syntax rules that DDEV doesn't have to define or maintain.

Allows defining for custom commands in YAML

Create a custom command
Instead of using the `## Annotation: Value` syntax, use the following syntax:

```text
<<'###YAML_ANNOTATIONS'
annotation: value
###YAML_ANNOTATIONS
```

for example:

```text
<<'###YAML_ANNOTATIONS'
description: "A description for the command"
usage: "command-name [flags] [args]"
aliases:
  - "command-alias"
  - "my-command"
projectTypes:
  - "laravel"
  - "php"
execRaw: true
mutagenSync: true
###YAML_ANNOTATIONS
```

## Automated Testing Overview

Will be written after I have a good sense that the general approach is acceptable

## Release/Deployment Notes

The old syntax will still work, but should be considered deprecated.
In a future major release, the old syntax should be removed.

## Remaining tasks:

1. [ ] Confirm the general approach
2. [ ] Make sure flags work
3. [ ] Any other "todo" comments currently added in the PR
4. [ ] Update all commands included with DDEV
5. [ ] Update docs
6. [ ] Add tests for the new format

## Outstanding questions:

1. Is this a good way to delimit the start and end of the YAML command content?
1. Should the YAML format accept keys with both capital and lower case first characters (if `yaml.unmarshal` allows that)?
    - If `yaml.unmarshal` dictates it must be one or the other, should it accept upper-case or lower-case?
1. Should the documentation list both syntaxes or just the new YAML syntax?
    - If both, how should that look? e.g. show YAML as default but then "before x.y.z use `<old syntax>`"?
1. If a command uses the old syntax, should a warning or message of some sort be output to the terminal? (I would suggest no)

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -7,7 +7,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
@@ -18,7 +20,29 @@ import (
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
+
+type annotations = struct {
+	// @TODO maybe some should not be omit empty
+	// @TODO maybe allow or even enforce capital letter starts
+	// @TODO check for any missing annotations and make sure we parse and use them
+	Aliases           []string `yaml:"aliases,omitempty"`
+	AutocompleteTerms []string `yaml:"autocompleteTerms,omitempty"`
+	CanRunGlobally    bool     `yaml:"canRunGlobally,omitempty"`
+	DbTypes           []string `yaml:"dbTypes,omitempty"`
+	Description       string   `yaml:"description,omitempty"`
+	DisableFlags      bool     `yaml:"-"`
+	Example           string   `yaml:"example,omitempty"`
+	ExecRaw           bool     `yaml:"execRaw,omitempty"`
+	Flags             Flags    `yaml:"flags,omitempty"`
+	HostBinaryExists  string   `yaml:"hostBinaryExists,omitempty"`
+	HostWorkingDir    bool     `yaml:"hostWorkingDir,omitempty"`
+	MutagenSync       bool     `yaml:"mutagenSynx,omitempty"`
+	OsTypes           []string `yaml:"osTypes,omitempty"`
+	ProjectTypes      []string `yaml:"projectTypes,omitempty"`
+	Usage             string   `yaml:"usage,omitempty"`
+}
 
 const (
 	CustomCommand        = "customCommand"
@@ -107,7 +131,6 @@ func addCustomCommands(rootCmd *cobra.Command) error {
 // addCustomCommandsFromDir adds the custom commands from inside a given directory
 func addCustomCommandsFromDir(rootCmd *cobra.Command, app *ddevapp.DdevApp, serviceDirOnHost string, commandFiles []string, isGlobalSet bool, commandsAdded map[string]int) error {
 	service := filepath.Base(serviceDirOnHost)
-	var err error
 
 	for _, commandName := range commandFiles {
 		onHostFullPath := filepath.Join(serviceDirOnHost, commandName)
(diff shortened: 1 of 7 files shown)


GuySartorelli:
This is in a very early state - I'm mostly looking for answers to the "Outstanding questions" sections and to validate the overall approach.

Note I'm not sure how much time I have to dedicate to this - it is likely development of this will be quite drawn out unless someone else wants to pitch in and help flesh it out.

rfay:
Were you able to use or build on 
* https://github.com/ddev/ddev/pull/2519

at all?

(When I saw the title go by I thought it was about adding annotations for additional services, which has been an active topic recently.)

rfay (commented):
After a quick look, I don't see why this shouldn't have a future. It looks like a very straightforward and carefully implemented approach. I do recommend waiting for @stasadev 's response next week. 

The one thing I thought about a bit was our hope to move from using yamlv3 to viper or something.
* https://github.com/ddev/ddev/issues/5763

Using yaml directly in go has worked fine for us, but it's pretty low-level compared to some alternatives. 

I'm not suggesting a that the viper idea should be a blocker for this initiative.

GuySartorelli:
> Were you able to use or build on https://github.com/ddev/ddev/pull/2519 at all?

I did look at that - the flag refactoring _might_ be useful when I start looking at flags specifically, but I got the impression most of the PR was making the groundwork for automatically updating existing commands to use yaml, rather than actually parsing and using that yaml.

> I don't see why this shouldn't have a future. It looks like a very straightforward and carefully implemented approach.

Awesome, that's good to read. Did you have any thoughts about the outstanding questions?

> I do recommend waiting for @stasadev 's response next week.

Will do.

> The one thing I thought about a bit was our hope to move from using yamlv3 to viper or something.

I'll take a look at that - though I suspect my conclusion will be that that should be done separately, and that for now we should be internally consistent (i.e. this should probably use the same yaml unmarshalling that the rest of the codebase uses).

rfay:
> Is this a good way to delimit the start and end of the YAML command content?

I don't mind the `<<'###YAML_ANNOTATIONS'` but where does that come from? Would a HEREDOC format be more familiar?

> Should the YAML format accept keys with both capital and lower case first characters (if yaml.unmarshal allows that)?

I'm in favor of strict case, and normally lower-case

> If yaml.unmarshal dictates it must be one or the other, should it accept upper-case or lower-case?

Must be same.

> Should the documentation list both syntaxes or just the new YAML syntax?

Not IMO

> If both, how should that look? e.g. show YAML as default but then "before x.y.z use <old syntax>"?

Just use new

> If a command uses the old syntax, should a warning or message of some sort be output to the terminal? (I would suggest no)

Let it be; the message would go to the wrong person.

GuySartorelli:
> I don't mind the `<<'###YAML_ANNOTATIONS'` but where does that come from? Would a HEREDOC format be more familiar?

This is one of the suggested syntaxes from the linked issue. It is a HEREDOC I think.

cmd/ddev/cmd/commands.go
```diff
@@ -18,8 +20,30 @@ import (
...
+	CanRunGlobally    bool     `yaml:"canRunGlobally,omitempty"`
+	DbTypes           []string `yaml:"dbTypes,omitempty"`
+	Description       string   `yaml:"description,omitempty"`
+	DisableFlags      bool     `yaml:"-"`
+	Example           string   `yaml:"example,omitempty"`
+	ExecRaw           bool     `yaml:"execRaw,omitempty"`
+	Flags             Flags    `yaml:"flags,omitempty"`
+	HostBinaryExists  string   `yaml:"hostBinaryExists,omitempty"`
+	HostWorkingDir    bool     `yaml:"hostWorkingDir,omitempty"`
+	MutagenSync       bool     `yaml:"mutagenSynx,omitempty"`
```
stasadev, line 41:
```suggestion
	MutagenSync       bool     `yaml:"mutagenSync,omitempty"`
```

cmd/ddev/cmd/commands.go
```diff
@@ -18,8 +20,30 @@ import (
...
+	Aliases           []string `yaml:"aliases,omitempty"`
+	AutocompleteTerms []string `yaml:"autocompleteTerms,omitempty"`
+	CanRunGlobally    bool     `yaml:"canRunGlobally,omitempty"`
+	DbTypes           []string `yaml:"dbTypes,omitempty"`
+	Description       string   `yaml:"description,omitempty"`
+	DisableFlags      bool     `yaml:"-"`
+	Example           string   `yaml:"example,omitempty"`
+	ExecRaw           bool     `yaml:"execRaw,omitempty"`
+	Flags             Flags    `yaml:"flags,omitempty"`
+	HostBinaryExists  string   `yaml:"hostBinaryExists,omitempty"`
```
stasadev, line 39: I think it's better to make it an array for readability:

```suggestion
	HostBinaryExists  []string `yaml:"hostBinaryExists,omitempty"`
```

stasadev (commented):
I like this syntax.

I suggest using underscores to make it the same style we have in YAML files.

`execRaw` => `exec_raw`, and so on.

rfay:
Seems like it's ready to move forward when you have time for it. Needs nontrivial manual rebase in the meantime though.

rfay:
Let us know if you want to continue with this at this point or reopen it in the future, thanks!

ChloeSartorelli:
I do _want_ to continue, but I don't have time right now unfortunately. I'll close it for now and reopen when I have more time.

rfay:
Totally understand. Your amazing contributions are so much valued, and this can easily be reopened. Thanks!

PR #7234, opened 2025-04-25
refactor: Add retries to util.Download(), both the SHASUM and the target file

rfay:
## The Issue

A number of tests, especially on buildkite, have been interrupted by download failures, especially shasum download failures. We don't know what's going on with those machines, but it seems like a retry is a reasonable thing.

ChatGPT assisted.

* Try 4 times
* Wait 500ms
* Retry after 5s
* Complete timeout after 60s (per retry)

Hopefully this won't be very relevant for most real users, but it should improve our tests a bit.

```
rm -r ~/.ddev/bin/*
ddev restart
```

Also review logs to see what happens there. DDEV_DEBUG=true gives the actual retry info, which test logs have by default.

## Automated Testing Overview

Should pass existing tests.


stasadev (approved):
Looks good to me!

Issue #7235, opened 2025-04-25
Feature: `.ddev/.env.<anything>` or `.ddev/.env.global.<anything>`


rfay:
### Feature Request

We might want to have a way to use a unique `.ddev/.env.*` file for particular applications. 

This specifically relates to
* #7233

They would like a Craft-specific .env that is *not* in the application space. It's a good idea and would improve all the CMSs that use .env, and we wouldn't be writing their file or really needing the #ddev-generated as much, although we can still request it.

Currently .ddev/.env works great for all containers, and .ddev/.env.* refers to a particular container name like .ddev/.env.solr

stasadev:
Think about how we should gitignore these .env files in the `.ddev/` directory.

stasadev:
From Drupal Slack https://drupal.slack.com/archives/C5TQRQZRR/p1754664513749809

It's better not to deliver `.ddev/.env.web` in add-on, because other add-ons may deliver it as well.
I think we should extend the support on `.ddev/.env` for `.ddev/.env.<anything>.<service>`

PR #7236, opened 2025-04-25
refactor: Use `.ddev/.env.web` for Craft CMS DDEV settings, fixes #7233

### LINKED ISSUE #7233: Improve Craft CMS "settings management" (writing to project `.env`)

The Craft CMS project type writes a number of `CRAFT_*` vars to the project's `.env` file, which is not ideal for a few reasons:
- Craft itself doesn't natively enforce .env usage, so it relys on someone using `vlucas/phpdotenv`
- We've found it not uncommon for people to upload their local .env file to a server and use as a base for editing. Usually they are aware enough to change the `CRAFT_DB_` vars, but often times don't realize `CRAFT_WEB_ROOT` was added by DDEV, leave it in place, and break their remote app.

### Describe your solution

The way things are implemented now seems to be somewhat by chance and a fundamental misunderstanding of how the "settings management" feature is intended to be used.

Ideally, DDEV wouldn't touch the project's `.env` file, and instead inject the environment variables directly into the containers.

### Describe alternatives

 It appears that writing `.ddev/.env` is one option for that, but writing to YAML may also be an option.

### Additional context

- https://github.com/craftcms/cms/issues/17175
- https://github.com/craftcms/cms/pull/16980

rfay:
## The Issue

- #7233

Craft CMS settings for DDEV would be better in the .ddev namespace

Move DB settings and the like into .ddev/.env.web. Could be improved with better naming, see
* #7235 

## TODO

- [x] Changes to Craft CMS quickstart doc?
- [x] Figure out how to make this backward-compatible. Warn people about project-level .env if it contains things?
- [x] Add `#ddev-generated` to the generated file if we can. Previously we couldn't, but with this approach we mostly should be able to.

Use the [DDEV Craft CMS Quickstart](https://ddev.readthedocs.io/en/stable/users/quickstart/#craft-cms)

Note that the `composer create-project craftcms/craft` absolutely insists on creating a project-level .env with these values. It takes the correct values as default that have already been set by the .ddev/.env.web, but then they're place in the project .env, which is a pain.

## Automated Testing Overview

Existing tests should behave the same

### DIFF
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -2,6 +2,7 @@ package ddevapp
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 
@@ -100,7 +101,8 @@ func updateCraftCMSDotEnv(app *DdevApp) (string, error) {
 
 	// Read existing env file
 	_, existingEnvText, err := ReadProjectEnvFile(envFilePath)
-	if err != nil {
+	// If envFilePath doesn't exist, that's not really an error, continue
+	if !os.IsNotExist(err) {
 		return "", err
 	}
 
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -62,7 +62,7 @@ func craftCmsImportFilesAction(app *DdevApp, uploadDir, importPath, extPath stri
 }
 
 // Set up the .ddev/.env.web file for ddev
-func createCraftCMSDotEnv(app *DdevApp) (string, error) {
+func updateCraftCMSDotEnv(app *DdevApp) (string, error) {
 	// If settings management is disabled, do nothing
 	if app.DisableSettingsManagement {
 		return "", nil
@@ -85,7 +85,7 @@ func createCraftCMSDotEnv(app *DdevApp) (string, error) {
 		port = "5432"
 	}
 
-	envMap := map[string]string{
+	newCraftEnvMap := map[string]string{
 		"CRAFT_DB_DRIVER":       driver,
 		"CRAFT_DB_SERVER":       "db",
 		"CRAFT_DB_PORT":         port,
@@ -98,7 +98,13 @@ func createCraftCMSDotEnv(app *DdevApp) (string, error) {
 		"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
 	}
 
-	err := WriteProjectEnvFile(envFilePath, envMap, "")
+	// Read existing env file
+	_, existingEnvText, err := ReadProjectEnvFile(envFilePath)
+	if err != nil {
+		return "", err
+	}
+
+	err = WriteProjectEnvFile(envFilePath, newCraftEnvMap, existingEnvText)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -102,7 +102,7 @@ func updateCraftCMSDotEnv(app *DdevApp) (string, error) {
 	// Read existing env file
 	_, existingEnvText, err := ReadProjectEnvFile(envFilePath)
 	// If envFilePath doesn't exist, that's not really an error, continue
-	if !os.IsNotExist(err) {
+	if err != nil && !os.IsNotExist(err) {
 		return "", err
 	}
 
--- a/pkg/ddevapp/craftcms.go
+++ b/pkg/ddevapp/craftcms.go
@@ -2,7 +2,6 @@ package ddevapp
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"regexp"
 
@@ -62,7 +61,7 @@ func craftCmsImportFilesAction(app *DdevApp, uploadDir, importPath, extPath stri
 	return nil
 }
 
-// Set up the .env file for ddev
+// Set up the .ddev/.env.web file for ddev
 func createCraftCMSDotEnv(app *DdevApp) (string, error) {
 	// If settings management is disabled, do nothing
 	if app.DisableSettingsManagement {
@@ -77,32 +76,7 @@ func createCraftCMSDotEnv(app *DdevApp) (string, error) {
 		}
 	}
 
-	// If the .env file doesn't exist, try to create it by copying .env.example to .env
-	envFilePath := app.GetConfigPath(".env.web")
-	if !fileutil.FileExists(envFilePath) {
-		var exampleEnvFilePaths = []string{".env.example", ".env.example.dev"}
-		for _, envFileName := range exampleEnvFilePaths {
-			exampleEnvFilePath := filepath.Join(app.AppRoot, app.ComposerRoot, envFileName)
-			if fileutil.FileExists(exampleEnvFilePath) {
-				util.Success(fmt.Sprintf("Copied %s to .env", envFileName))
-				err := fileutil.CopyFile(exampleEnvFilePath, envFilePath)
-				if err != nil {
-					util.Error(fmt.Sprintf("Error copying %s to .env", exampleEnvFilePath))
-
-					return "", err
-				}
-			}
-		}
-	}
-	// If the .env file *still* doesn't exist, return early
-	if !fileutil.FileExists(envFilePath) {
-		return "", nil
-	}
-	// Read in the .env file
-	envMap, envText, err := ReadProjectEnvFile(envFilePath)
-	if err != nil && !os.IsNotExist(err) {
-		return "", fmt.Errorf("unable to read %s: %v", envFilePath, err)
-	}
+	envFilePath := app.SiteSettingsPath
 
 	port := "3306"
 	driver := "mysql"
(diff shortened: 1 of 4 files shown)


timkelty:
@rfay ran through this and have some notes:

1. I'm not clear why this PR is changing the file from `.env` to `.env.web`. Is that just for clarity until https://github.com/ddev/ddev/issues/7235 allows it to be `.env.craftcms`?
2. The env vars from the `.ddev/.env.web` need to be available in the container when running the `ddev craft install/craft`. Currently they are NOT. Ideally they is done programatically after the `.ddev/.env.web` is created. If that isn't feasible, the quickstart needs to be updated to have a `ddev restart` before the install command.
3. I don't think we should worry about copying the .env example template included by the `composer create` command. Ideally, the `.env.web` would be created on `ddev config` (or initial `start`), containing only the dynamic values DDEV can set. This would also make them available for the install command, nullifying **2**.

> Note that the composer create-project craftcms/craft absolutely insists on creating a project-level .env with these values. It takes the correct values as default that have already been set by the .ddev/.env.web, but then they're place in the project .env, which is a pain.

This is not my experience with this PR. I'm thinking maybe you didn't use the `--no-scripts` as shown in the quickstart guide? With that, there should be no project level .env created. The next step (`ddev craft install/craft`) will ask you to create one for values it doesn't yet have (and needs).

To clarifiy, after quickstart, what we want is something like this:

```
# <project>/.ddev/.env.web (dynamic DDEV values)
CRAFT_DB_DRIVER="mysql"
CRAFT_DB_SERVER="db"
CRAFT_DB_PORT="3306"
CRAFT_DB_DATABASE="db"
CRAFT_DB_USER="db"
CRAFT_DB_PASSWORD="db"
CRAFT_DB_SCHEMA=public
CRAFT_DB_TABLE_PREFIX=
CRAFT_WEB_ROOT="/var/www/html/web"
MAILPIT_SMTP_HOSTNAME="127.0.0.1"
MAILPIT_SMTP_PORT="1025"
PRIMARY_SITE_URL="https://my-craft-site.ddev.site"
```

```
# <project>/.env (values created by the craft installer)
CRAFT_APP_ID=CraftCMS--90c39817-d6e0-4632-89b3-dbae073004ad
CRAFT_ENVIRONMENT=dev
CRAFT_SECURITY_KEY=epH_oDAmHr1Wo4tMICZC5_7m9X9YG8ts
CRAFT_DEV_MODE=true
CRAFT_ALLOW_ADMIN_CHANGES=true
CRAFT_DISALLOW_ROBOTS=true
```

rfay:
> I'm not clear why this PR is changing the file from .env to .env.web. Is that just for clarity until https://github.com/ddev/ddev/issues/7235 allows it to be .env.craftcms?

Right now, the choices are `.ddev/.env` (which affects every container) and `.ddev/.env.web` (very lightly used, maybe not used anywhere). I used `.env.web` just because then it doesn't have to consider a conflict with an existing .env. It may not be a good enough choice.

More followup later.

timkelty:
> The .ddev/.env.web should be done in a function called from settingsCreator in apptypes.go

That makes sense, but in this case `postStart` should also work – but in this case it isn't.
If you follow the quickstart up until just before the `composer create`, you'll notice that the `.ddev/.env.web` is no present and the vars are not loaded in the container:

```
# Create a project directory and move into it:
mkdir my-craft-site && cd my-craft-site

# Set up the DDEV environment:

ddev start
```

timkelty:
> Right now, the choices are .ddev/.env (which affects every container) and .ddev/.env.web (very lightly used, maybe not used anywhere). I used .env.web just because then it doesn't have to consider a conflict with an existing .env. It may not be a good enough choice.

Gotcha. I'm thinking in terms of – in future, it will make sense for the Craft project type to include a queue/worker container, which would also need these vars, so `.ddev/.env` felt more appropriate. I take your meaning about conflicts with existing though – but for Craft CMS users I don't think this would be an issue.

rfay:
> Gotcha. I'm thinking in terms of – in future, it will make sense for the Craft project type to include a queue/worker container, which would also need these vars, so .ddev/.env felt more appropriate. I take your meaning about conflicts with existing though – but for Craft CMS users I don't think this would be an issue.

Most implementations (Like [Pimcore](https://ddev.readthedocs.io/en/stable/users/quickstart/#pimcore) for example) put a queue worker into the DDEV web container with `web_extra_daemons`, super easy.

timkelty:
> Most implementations (Like [Pimcore](https://ddev.readthedocs.io/en/stable/users/quickstart/#pimcore) for example) put a queue worker into the DDEV web container with `web_extra_daemons`, super easy.

I'll plan on that approach, then.

rfay:
Updated this and hope you can take it for another spin @timkelty , https://github.com/ddev/ddev/pull/7236#issuecomment-2831050404

Now it populates the .ddev/.env.web very early, so it's there on first `ddev start`

pkg/ddevapp/craftcms.go
```diff
@@ -111,42 +85,25 @@ func craftCmsPostStartAction(app *DdevApp) error {
...
+		"CRAFT_DB_USER":         "db",
+		"CRAFT_DB_PASSWORD":     "db",
+		"CRAFT_WEB_ROOT":        app.GetAbsDocroot(true),
+		"MAILPIT_SMTP_HOSTNAME": "127.0.0.1",
+		"MAILPIT_SMTP_PORT":     "1025",
+		"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
 	}
 
-	err = WriteProjectEnvFile(envFilePath, envMap, envText)
+	err := WriteProjectEnvFile(envFilePath, envMap, "")
```
stasadev, line 101: We need to get `ReadProjectEnvFile` back.

Now this code takes full control of the `.ddev/.env.web` file, and you can't set any more env variables there:
```bash
ddev dotenv set .ddev/.env.web --some-variable=foobar
ddev restart
cat .ddev/.env.web # no SOME_VARIABLE
```
But it should be possible.

stasadev (changes requested):

pkg/ddevapp/craftcms.go
rfay: > Now this code takes full control of the .ddev/.env.web file, and you can't set any more env variables there:

Yes, this is the difficulty with not currently having a way to control what's "owned by DDEV" in dotenv files. 

* https://github.com/ddev/ddev/issues/7235 would help with this (I'd use .ddev/.env.global.craftcms probably) but it still wouldn't be authoritative.

stasadev: But it's already working with `ddev dotenv set` (otherwise this command would override the whole file).

This PR should work the same way, it should change only env variables that are related to Craft CMS.

stasadev: Ideally, `WriteProjectEnvFile` should support append or create mode. It's annoying that we need extra code to append or replace env vars in an existing file.

rfay: Ah, so we could use the logic/function from `ddev dotenv set`??? Awesome!

timkelty:
@rfay it's lovely! PR'd the accompanying quickstart edits here: https://github.com/ddev/ddev/pull/7274

rfay:
Updated with the proper read of env file before writing it, tested with 
* https://github.com/ddev/ddev/pull/7274 ([rendered](https://ddev--7274.org.readthedocs.build/en/7274/users/quickstart/#craft-cms)

rfay:
@timkelty We have a release planned for this week, that this PR could possibly make it into. Would it be important to get it into this release, or would it be better to go into the next release cycle (early)?

rfay:
Related question @timkelty - Should we worry about migration of Craft CMS projects after this change? They'll have duplicate info in the project .env and the `.ddev/.env.web` I imagine. But it likely won't do any harm.

timkelty:
> @timkelty We have a release planned for this week, that this PR could possibly make it into. Would it be important to get it into this release, or would it be better to go into the next release cycle (early)?

Not particularly important, no. I can try to have the [quickstart PR](https://github.com/ddev/ddev/pull/7274) ready today, though.

pkg/ddevapp/craftcms.go
```diff
@@ -111,42 +86,32 @@ func craftCmsPostStartAction(app *DdevApp) error {
...
+		"CRAFT_WEB_ROOT":        app.GetAbsDocroot(true),
+		"MAILPIT_SMTP_HOSTNAME": "127.0.0.1",
+		"MAILPIT_SMTP_PORT":     "1025",
+		"PRIMARY_SITE_URL":      app.GetPrimaryURL(),
+	}
+
+	// Read existing env file
+	_, existingEnvText, err := ReadProjectEnvFile(envFilePath)
+	// If envFilePath doesn't exist, that's not really an error, continue
+	if !os.IsNotExist(err) {
```
stasadev, line 105: If I run:

```
ddev dotenv set .ddev/.env.web --craft-db-user=foobar && ddev config --auto
```

`CRAFT_DB_USER` is not updated anymore, that because of this missing condition:

```suggestion
	if err != nil && !os.IsNotExist(err) {
```

stasadev (changes requested):
In addition to my suggestion,

What if people do this:

```
$ ddev dotenv set .ddev/.env.web --craft-db-user=foobar
$ ddev restart
$ ddev exec 'echo $CRAFT_DB_USER'      
foobar
$ ddev dotenv get .ddev/.env.web --craft-db-user
db
$ ddev restart
$ ddev exec 'echo $CRAFT_DB_USER'      
db
$ ddev dotenv get .ddev/.env.web --craft-db-user
db
```

`app.CreateSettingsFile()` is called *after* the web container is started in `app.Start()`.

Not sure whether it is important, just mentioning this edge case.

rfay:
Your fix solves the one problem, thanks!

I'm not sure I completely understand the scenario in https://github.com/ddev/ddev/pull/7236#pullrequestreview-2834245608, but it *seems* like it would be mitigated by people understanding that a change of configuration requires a `ddev restart`.

stasadev (approved):
> it *seems* like it would be mitigated by people understanding that a change of configuration requires a `ddev restart`

Okay. I'm not saying anything needs to change here, it's just a bit awkward that the settings file is updated immediately on `ddev config`, while the update on `ddev start` only happens after the project is already running, not during the pre-start phase.

rfay:
> it's just a bit awkward that the settings file is updated immediately on ddev config, while the update on ddev start only happens after the project is already running, not during the pre-start phase

Drupal has worked that way over the years, but it's not using environment variables, but instead using php code, which is absolutely a fundamental difference. 

Even though the environment variables do get loaded before php-fpm starts... you're pointing out that the .env doesn't even get populated until start finishes.

stasadev:
Yes, modifying `.ddev/.env.*` files is not the same. And I don't know if we can move the logic to an early `ddev start` (didn't investigate it).

rfay:
I'll try to study it just a bit before pulling. I don't want to make any real structural changes to get this in, but OTOH the whole env var approach to settings is very different from what the original usage was (Drupal/WP/etc with php-based settings)

rfay:
The issue here is that CreateSettingsFile() needs to be called before we do the `docker-compose up`. There is no reason I can see not to do this, but I don't want to do it in this PR. It affects several project types, and should be a trivial change, but best not to bundle it here (and best to leave it for a release after this week's release). 

I'll do a PR after pulling this.

PR #7237, opened 2025-04-28
docs: wrap quotes around commands that use the caret symbol

nmangold:
## The Issue

When using Zsh, the `composer create ...` command will fail if using a caret symbol to specify the project version.

Drupal Slack conversation: https://drupal.slack.com/archives/C5TQRQZRR/p1745774586216839

Zsh issue:

- https://github.com/ohmyzsh/ohmyzsh/pull/3919

Wrapping the project and version number in quotes will prevent Zsh from globbing, and allow the command to execute as expected.

Review https://ddev--7237.org.readthedocs.build/en/7237/users/quickstart/

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -235,7 +235,7 @@ Read more about customizing the environment and persisting configuration in [Pro
     mkdir my-drupal-site && cd my-drupal-site
     ddev config --project-type=drupal11 --docroot=web
     ddev start
-    ddev composer create-project drupal/recommended-project:^11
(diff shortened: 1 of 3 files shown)


rfay:
Added quickstart tests and rebased

rfay (approved):
Looks good to me, passed quickstart tests.

rpkoller (approved):
adding the double quotes looks good to me. the fail save approach.

GuySartorelli:
Oh that's what causes it! Thank you for raising this. I occasionally stumble on this but never bothered to find out what part of the command causes it - I just quoted all the args and moved on.

PR #7238, opened 2025-04-28
fix: add warning about duplicate usage for global commands, for #6293

stasadev:
## The Issue

Sometimes I copy and paste global commands, edit them slightly, and then I don't understand why they don't work:

```
$ cp ~/.ddev/commands/host/self-upgrade ~/.ddev/commands/host/self-upgrade2
$ ddev self-upgrade2
Error: unknown command "self-upgrade2" for "ddev"

Did you mean this?
        self-upgrade
        self-upgrade

Run 'ddev --help' for usage.
```

For:
- #6293

- Adds a duplicate warning for `## Usage:`
- Adds a file path to all warnings for a better understanding of where they are coming from.

```
$ cp ~/.ddev/commands/host/self-upgrade ~/.ddev/commands/host/self-upgrade2
$ ddev self-upgrade2
Command 'self-upgrade2' cannot have usage 'self-upgrade' because it is already in use by command 'self-upgrade', skipping /home/stas/.ddev/commands/host/self-upgrade2
Error: unknown command "self-upgrade2" for "ddev"

Did you mean this?
        self-upgrade

Run 'ddev --help' for usage.
```

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -137,7 +137,7 @@ func addCustomCommandsFromDir(rootCmd *cobra.Command, app *ddevapp.DdevApp, serv
 		if service == "host" && app == nil {
 			if val, ok := directives["CanRunGlobally"]; !ok || val != "true" {
 				if isCustomCommandInArgs(commandName) {
-					util.Warning("Command '%s' cannot be used outside the project directory.", commandName)
+					util.Warning("Command '%s' cannot be used outside the project directory, skipping %s", commandName, onHostFullPath)
 				}
 				continue
 			}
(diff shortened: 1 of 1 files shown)


rfay (commented):
How does it know which one is the "real" command?

stasadev:
> How does it know which one is the "real" command?

It's based on the order of what is passed inside `addCustomCommandsFromDir()`.

```
$ cp ~/.ddev/commands/host/self-upgrade ~/.ddev/commands/host/a-self-upgrade

$ ddev a-self-upgrade
Command 'self-upgrade' cannot have usage 'self-upgrade' because it is already in use by command 'self-upgrade', skipping /home/stas/.ddev/commands/host/self-upgrade
```

So in this case `a-self-upgrade` was registered first.

I don't think it's too important to prefer bundled global commands because these changes are made by the user and the warning should give them enough clues that they have done something wrong.

PR #7247, opened 2025-05-01
fix: correct warning output formatting (Drupal) and don't try to do exec when not running

stasadev:
## The Issue

From Drupal Slack https://drupal.slack.com/archives/C5TQRQZRR/p1746102793754109

When I did `ddev config` it said: `Unable to set permissions inside container on settings files: ''` <- empty string in the quotes

- Adds `stderr` redirect to `stdout`.
- Removes `-f` silent flag.
- Doesn't run `app.Exec()` without a running container.

Now it outputs this (I reproduced it using a dummy go command):
```
Unable to set permissions inside container on settings files:
chmod: cannot access '/var/www/html/web/sites/default': No such file or directory
chmod: cannot access '/var/www/html/web/sites/default/settings.php': No such file or directory
```

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -560,17 +560,14 @@ func drupalEnsureWritePerms(app *DdevApp) error {
 			util.Warning("Unable to set permissions: %v", err)
 		}
 	}
+	// Don't run app.Exec() without a running container.
+	status, _ := app.SiteStatus()
 	// It is possible that a Drupal install has been done, setting sites/default and sites/default/settings.php
 	// to read-only, which breaks mutagen's access to them so mutagen then
 	// can't sync. See https://github.com/ddev/ddev/issues/6491
 	// So we chmod +w the two files that a Drupal install may set read-only
 	// *inside* the container, allowing mutagen access to it again
-	if app.IsMutagenEnabled() {
-		// Don't run app.Exec() without a running container.
-		status, _ := app.SiteStatus()
-		if status != SiteRunning {
-			return nil
-		}
+	if app.IsMutagenEnabled() && status == SiteRunning {
 		settingsFiles := []string{
 			path.Join(app.GetAbsDocroot(true), `sites/default`),
 			path.Join(app.GetAbsDocroot(true), `sites/default/settings.php`),
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -572,11 +572,12 @@ func drupalEnsureWritePerms(app *DdevApp) error {
 			path.Join(app.GetAbsDocroot(true), `sites/default`),
 			path.Join(app.GetAbsDocroot(true), `sites/default/settings.php`),
 		}
-		stdout, _, err := app.Exec(&ExecOpts{
+		stdout, stderr, err := app.Exec(&ExecOpts{
 			Cmd: fmt.Sprintf(`chmod u+w %s 2>&1`, strings.Join(settingsFiles, " ")),
 		})
 		if err != nil {
-			util.Warning("Unable to set permissions inside container on settings files:\n%s", stdout)
+			util.Warning("Unable to set permissions inside container on settings files; err='%v', stdout='%s', stderr='%s'", err, stdout, stderr)
+
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rfay:
I've been noticing that, thanks!

rfay:
Actually, I think this is in a particular situation (where settings file doesn't exist?). We should probably figure out exactly how to recreate and not show the message in that situation.

rfay:
Rebased for quickstart test.

I've seen this warning many, many times, but am unable to recreate today on WSL2 with mutagen enabled. Maybe it's only happening on macOS.

stasadev:
`chmod -f` means silent

```
$ chmod --help
  -f, --silent, --quiet  suppress most error messages
```

So we didn't get any error message at all.

I refactored it to redirect the output to stdout.

rfay:
This problem happens when doing `ddev config --project-type=drupal11 --docroot=web` when the project hasn't been started. The failure happens because the project isn't running, and we're trying to do an app.Exec(). IMO the logic should not go in there if the project is not running. But I think it's likely that app.IsMutagenEnabled() should only return true if the project is started; I'm reluctant to do that because it could have larger impacts.

https://github.com/ddev/ddev/blob/adb461684aec2e77011a345c6246c6393320bd8f/pkg/ddevapp/drupal.go#L573-L575

stasadev:
I added a condition for a running container.

pkg/ddevapp/drupal.go
```diff
@@ -566,15 +566,20 @@ func drupalEnsureWritePerms(app *DdevApp) error {
 	// So we chmod +w the two files that a Drupal install may set read-only
 	// *inside* the container, allowing mutagen access to it again
 	if app.IsMutagenEnabled() {
```
rfay, line 568: IMO instead of returning nil in the not-running condition, it would be better to change the condition to siterunning && IsMutagenEnabled, that way if something gets added in the future after the if stanza it won't cause unexpected behavior.

rfay (changes requested):

pkg/ddevapp/drupal.go
```diff
@@ -560,21 +560,23 @@ func drupalEnsureWritePerms(app *DdevApp) error {
...
 			path.Join(app.GetAbsDocroot(true), `sites/default/settings.php`),
 		}
-		_, stderr, err := app.Exec(&ExecOpts{
-			Cmd: fmt.Sprintf(`chmod -f u+w %s`, strings.Join(settingsFiles, " ")),
+		stdout, _, err := app.Exec(&ExecOpts{
+			Cmd: fmt.Sprintf(`chmod u+w %s 2>&1`, strings.Join(settingsFiles, " ")),
 		})
 		if err != nil {
-			util.Warning("Unable to set permissions inside container on settings files: '%s'", stderr)
+			util.Warning("Unable to set permissions inside container on settings files:\n%s", stdout)
```
rfay, line 579: I don't think this will even happen any more, but the error message can be better. Testing it, I see that stdout and stderr are useless here, but err has the full info. 
```suggestion
			util.Warning("Unable to set permissions inside container on settings files; err='%v', stdout='%s', stderr='%s'", err, stdout, stderr)

```

pkg/ddevapp/drupal.go
```diff
@@ -560,21 +560,23 @@ func drupalEnsureWritePerms(app *DdevApp) error {
...
 	// *inside* the container, allowing mutagen access to it again
-	if app.IsMutagenEnabled() {
+	if app.IsMutagenEnabled() && status == SiteRunning {
 		settingsFiles := []string{
 			path.Join(app.GetAbsDocroot(true), `sites/default`),
 			path.Join(app.GetAbsDocroot(true), `sites/default/settings.php`),
 		}
-		_, stderr, err := app.Exec(&ExecOpts{
-			Cmd: fmt.Sprintf(`chmod -f u+w %s`, strings.Join(settingsFiles, " ")),
+		stdout, _, err := app.Exec(&ExecOpts{
```
rfay, line 575: Might as well get it all, even though unlikely to happen.
```suggestion
		stdout, stderr, err := app.Exec(&ExecOpts{
```

rfay (approved):
Thanks, this works great. Minor suggestion to get more out of our error reporting.

PR #7252, opened 2025-05-01
refactor: Make DownloadFile() retry logic more robust, improve curling in Makefile

rfay:
## The Issue

- https://github.com/ddev/ddev/pull/7234

In CI on Buildkite, we're still seeing download failures, even though we implemented retries. 

I had a chat with ChatGPT about the setup, and it suggested implementing a CheckRetry function that would note the "context deadline exceeded" that we're often seeing, and retry in that case. 

I don't think this does any harm.

But the problems are probably CI network problems.

Also added improved curl syntax in the Makefile, where we regularly have similar failures downloading mkcert.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,26 +207,26 @@ textlint:
 darwin_amd64_signed: $(GOTMP)/bin/darwin_amd64/ddev
 	@if [ -z "$(DDEV_MACOS_SIGNING_PASSWORD)" ] ; then echo "Skipping signing ddev for macOS, no DDEV_MACOS_SIGNING_PASSWORD provided"; else echo "Signing $< ..."; \
 		set -o errexit -o pipefail; \
-		$(CURL) -s --retry 5 --retry-delay 5 --retry-connrefused https://raw.githubusercontent.com/ddev/signing_tools/master/macos_sign.sh | bash -s -  --signing-password="$(DDEV_MACOS_SIGNING_PASSWORD)" --cert-file=certfiles/ddev_developer_id_cert.p12 --cert-name="Developer ID Application: Localdev Foundation (9HQ298V2BW)" --target-binary="$<" ; \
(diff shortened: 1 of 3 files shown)


rfay (commented):
This immediately got the "Context deadline exceeded" error, so it doesn't seem to be making much progress.

rfay:
I wonder if all the disconnected/failed buildkites were related to this behavior. Restarting them. I don't think this PR does any harm, but don't think it solves the problem it was hoping to solve.

Makefile
```diff
@@ -87,12 +87,15 @@ $(GOTMP)/bin/completions.tar.gz: build
 
 mkcert: $(GOTMP)/bin/darwin_arm64/mkcert $(GOTMP)/bin/darwin_amd64/mkcert $(GOTMP)/bin/linux_arm64/mkcert $(GOTMP)/bin/linux_amd64/mkcert
 
+# Set CURL to the Homebrew-installed curl, fallback to default
+CURL := $(shell command -v /opt/homebrew/opt/curl/bin/curl || command -v /usr/local/opt/curl/bin/curl || echo curl)
```
stasadev, line 91:
```suggestion
CURL := $(shell command -v /opt/homebrew/opt/curl/bin/curl || command -v /usr/local/opt/curl/bin/curl || command -v curl)
```

stasadev (approved):
It doesn't seem change anything.

I replaced `echo curl` with a real check.

Makefile
rfay: This should be exactly the same, the `echo curl` will make CURL=curl, which will get curl from PATH. Your change is perfectly valid, will give the full path to the curl in the PATH instead of just `curl`

rfay:
Using the brew-installed curl here has some interesting problems. It mostly apparently doesn't impact us currently because we don't have a test on macOS using the homebrew curl against a DDEV project I guess. 

But the homebrew-installed curl does not respect the system CA, which means it doesn't respect our mkcert CA, unless you do some fiddling with the ~/.curlrc. And if you don't do that right, it doesn't work with anything else on the internet. 

Here we're using it apparently only in `make` and not adding it to the PATH, so other usages don't matter, and we don't use the homebrew curl against DDEV projects.

Still, it's pretty questionable. /cc @stasadev

PR #7258, opened 2025-05-03
fix: add BASE_IMAGE arg before everything else, for #7071

hanoii:
## The Issue

- Follow up to #7071 

I started playing with this with HEAD and found an issue which took me a bit to figure out. All of my examples and even tests that worked before was because `ARG BASE_IMAGE="scratch"` was added on the prepend file but I just didn't as I was using a smaller base image for a build stage and the Dockerfile build failed because of https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact.

That particular ARG has to be before that first FROM so it can be used in any FROM afterwards. While we can be documented and leave prepend as plain as possible, I think this might confuse those that are trying to use this feature (if any). I figured that as BASE_IMAGE is a very important ARG to have defined, so I think it's best this is added before everything else (including prepend.Dockerfile variants). 

How to reproduce this with the current HEAD:

- Add a simple `prepend.Dockerfile` with:

```Dockerfile
FROM busybox AS test1
RUN echo "aa"
```

- `ddev start` will fail with an error of $BASE_IMAGE not being present.

### DIFF
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -197,7 +197,7 @@ The following environment variables are available for the web Dockerfile to use 
 * `$TARGETOS`: The build target operating system (always `linux`)
 * `$TARGETPLATFORM`: `linux/amd64` or `linux/arm64` depending on the machine it's been executed on
 
-!!! warning "Only `$BASE_IMAGE` is automatically available in `prepend.Dockerfile*` variants"
+!!!warning "Only `$BASE_IMAGE` is automatically available in `prepend.Dockerfile*` variants"
     If you need to use any of the other variables you will need to manually add them to your `prepend.Dockerfile*` files using [ARG](https://docs.docker.com/reference/dockerfile/#arg) instructions.
 
 For example, a Dockerfile might want to build an extension for the configured PHP version like this using `$DDEV_PHP_VERSION` to specify the proper version:
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -157,7 +157,7 @@ An example [Multi-stage](https://docs.docker.com/build/building/multi-stage/) we
 # If we want to use any of the build time environment variables injected by ddev
 # on the prepend.Dockerfile* variants we need to manually declare them to make
 # them available using the ARG instruction.
-# Only $BASE_IMAGE is already added as it must be global be used on FROM 
+# Only $BASE_IMAGE is already added as it must be global to be used on FROM 
 # statements. 
 FROM $BASE_IMAGE AS build-stage-go
 
--- a/docs/content/users/extend/customizing-images.md
+++ b/docs/content/users/extend/customizing-images.md
@@ -161,6 +161,8 @@ An example [Multi-stage](https://docs.docker.com/build/building/multi-stage/) we
 # statements. 
 FROM $BASE_IMAGE AS build-stage-go
 
+# While we are not using $uid and $gid in the code below, this serves as an example
+# of how any of the other DDEV's build variables must be defined. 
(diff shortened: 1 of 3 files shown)


rfay:
So it's accepting the BASE_IMAGE from docker-compose (if it exists), otherwise using scratch, and has to be first for the reasons you said. 

Did this impact you only when using the prepend.Dockerfile.*, or were there other impacts?

Do the docs need another example where scratch is not used, but instead it's using the normal BASE_IMAGE from the docker-compose?

hanoii:
> Did this impact you only when using the prepend.Dockerfile.*, or were there other impacts?

Only with `prepend.Dockerfile` file, but the reason is that by adding anything before what is currently your ARG $BASE_IMAGE='scratch' it just stops to be in the [global scope](https://docs.docker.com/build/building/variables/#scoping) and it fails.

> Do the docs need another example where scratch is not used, but instead it's using the normal BASE_IMAGE from the docker-compose?

The example actually used $BASE_IMAGE, the reason why neither the examples or the test gave me issues is that because I was basing it on $BASE_IMAGE I added the ARG myself before, so I was inadvertently adding it to the global space, a concept I just learned and undesrtood.

Using scratch or not has nothing to do with this, this is only because your original code added. I could argue if `ARG $BASE_IMAGE=scratch` makes sense or one could simple left it without default,. but that's for another issue.

**I modified docs to reflect this better and tests be able to catch it**

hanoii:
Oh, and I have added an example `prepend.Dockerfile` to the PR description that allows you to reproduce the issue.

hanoii:
Not sure on buildkit but I assume one offs?

rfay:
Always buildkite these days.

docs/content/users/extend/customizing-images.md
```diff
@@ -194,8 +197,8 @@ The following environment variables are available for the web Dockerfile to use
 * `$TARGETOS`: The build target operating system (always `linux`)
 * `$TARGETPLATFORM`: `linux/amd64` or `linux/arm64` depending on the machine it's been executed on
 
-!!!warning "These variables won't be automatically available on `prepend.Dockerfile*` variants"
-    If you need to use any of these variables you will need to manually add them to your `prepend.Dockerfile*` files using [ARG](https://docs.docker.com/reference/dockerfile/#arg) instructions.
+!!! warning "Only `$BASE_IMAGE` is automatically available in `prepend.Dockerfile*` variants"
```
stasadev, line 200: Remove space (it still works this way, but we don't use space in other places)
```suggestion
!!!warning "Only `$BASE_IMAGE` is automatically available in `prepend.Dockerfile*` variants"
```

stasadev (approved):
Thank you!

I encountered a similar problem some time ago when I was updating add-ons and experimenting with [`dockerfile_inline`](https://docs.docker.com/reference/compose-file/build/#dockerfile_inline), the `ARG` used in `FROM` must be declared first.

rfay (approved):
Thanks!

hanoii:
In case you fancy trying this out, once this is in:

`ddev add-on get https://github.com/hanoii/ddev-pimp-my-shell/tarball/multi-stage`

It is too good for me. I have several different teams using this add-on and I can't just enforce the latest version of ddev, so for a while I'll keep both approaches :)

https://github.com/hanoii/ddev-pimp-my-shell/blob/f0a870f771a4803b46996c24114eb80dcea51f88/config.pimp-my-shell.yaml#L4-L12

Also, this test failed because of this issue:

https://github.com/hanoii/ddev-pimp-my-shell/actions/runs/14846871224/job/41682633387?pr=2 (`stderr='failed to solve: base name ($BASE_IMAGE) should not be blank')

PR #7260, opened 2025-05-05
fix: set `XDEBUG_MODE=off` for `ddev composer create-project`, fixes ddev/ddev-intellij-plugin#414

stasadev:
## The Issue

- ddev/ddev-intellij-plugin#414

We set `XDEBUG_MODE=off` for `ddev composer`, but not for `ddev composer create-project`.

Adds `XDEBUG_MODE=off` for `ddev composer create-project`.

See https://github.com/ddev/ddev-intellij-plugin/issues/414#issuecomment-2848884444


rfay (approved):
Looks completely correct to me, thanks!

PR #7263, opened 2025-05-05
refactor: improve `ddev add-on get` output, add warning exit code annotation

hanoii:
This is just an idea on improving slightly the `ddev add-on get` output.

I was working on hanoii/ddev-pimp-my-shell and wanted to support installing it in both pre-1.24.5 and later so that I can keep for a while maintaining the two approaches (single stage/multi-stage).

I noticed something that was bothering me before as well, it is unrelated to the above, but it was what triggered it.

First, when an action outputs something, it stays above the description:

Screenshot 2025-05-05 at 11 48 02

which is confusing, I rather have it below.

Also, I would have loved to have something other than 👍🏻  and 👎🏻  to make it visually clear that is neither success, nor error (call it a warning).

So I come up with this PR which translates to this:

Screenshot 2025-05-06 at 08 50 08

It doesn't report as error either.

It relies on a a new annotation like `#ddev-description`: `#ddev-warning-exit-code:`

i.e.

```yaml
post_install_actions:
  - |
    #ddev-description:Checking DDEV version
    #ddev-warning-exit-code: 64
    if ! ${DDEV_EXECUTABLE} debug match-constraint ">= 1.24.5" 2>/dev/null ; then
      echo "This add-on works best with DDEV v1.24.5+ or higher, please upgrade." && exit 64
    fi
```

# Manual test:

```bash
mkdir -p /tmp/ddev-add-on
wget https://raw.githubusercontent.com/hanoii/ddev/refs/heads/improve-add-on-action-output/cmd/ddev/cmd/testdata/TestCmdAddonActionsOutput/recipe/install.yaml -O /tmp/ddev-add-on/install.yaml

# success

mkdir -p /tmp/ddev-test-project1
cd /tmp/ddev-test-project1
ddev add-on get /tmp/ddev-add-on

# with error

mkdir -p /tmp/ddev-test-project2
cd /tmp/ddev-test-project2
touch .ddev/test_cmd_addon_actions_output_error.txt
ddev add-on get /tmp/ddev-add-on
```

### DIFF
--- a/docs/content/users/extend/additional-services.md
+++ b/docs/content/users/extend/additional-services.md
@@ -87,6 +87,7 @@ The `install.yaml` is a simple YAML file with a few main sections:
 In any stanza of `pre_install_actions` and `post_install_actions` you can:
 
 * Use `#ddev-description:<some description of what stanza is doing>` to instruct DDEV to output a description of the action it's taking.
+* Use `#ddev-warning-exit-code:<an integer between 1-255>` to instruct DDEV to consider a certain exit code of the stanza as a warning, not an error, continuing with other actions as normal.
 
 You can see a simple `install.yaml` in [`ddev-addon-template`’s `install.yaml`](https://github.com/ddev/ddev-addon-template/blob/main/install.yaml).
 
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -140,7 +140,7 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 			// Get the exit code
 			exitCode := exitErr.ExitCode()
 			if exitCode == 63 {
-				util.Warning("%c %s", '\U000026A1', desc)
+				util.Warning("%s %s", "\U000026A0\U0000FE0F", desc)
 				err = nil
 			} else {
 				util.Warning("%c %s", '\U0001F44E', desc)
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -3,8 +3,10 @@ package ddevapp
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
+	goexec "os/exec"
 	"path/filepath"
 	"strings"
 	"text/template"
@@ -131,17 +133,27 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 		action = "set -x; " + action
 	}
 	out, err := exec.RunHostCommand(bashPath, "-c", action)
+
+	if err != nil {
+		var exitErr *goexec.ExitError
+		if errors.As(err, &exitErr) {
+			// Get the exit code
+			exitCode := exitErr.ExitCode()
+			if exitCode == 63 {
+				util.Warning("%c %s", '\U000026A1', desc)
+				err = nil
+			} else {
+				util.Warning("%c %s", '\U0001F44E', desc)
+				err = fmt.Errorf("Unable to run action %v: %v, output=%s", action, err, out)
+			}
+		}
+	} else {
+		util.Success("%c %s", '\U0001F44D', desc)
+	}
 	if len(out) > 0 {
 		util.Warning(out)
 	}
-	if err != nil {
-		util.Warning("%c %s", '\U0001F44E', desc)
-		return fmt.Errorf("Unable to run action %v: %v, output=%s", action, err, out)
-	}
-	if desc != "" {
-		util.Success("%c %s", '\U0001F44D', desc)
-	}
-	return nil
+	return err
 }
 
 // GetAddonDdevDescription returns what follows #ddev-description: in any line in action
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -134,7 +134,6 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 		action = "set -x; " + action
 	}
 	out, err := exec.RunHostCommand(bashPath, "-c", action)
-
 	if err != nil {
 		warningCode := GetAddonDdevWarningExitCode(action)
 		if warningCode > 0 {
@@ -148,7 +147,6 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 				}
 			}
 		}
-
 		if err != nil {
 			util.Warning("%c %s", '\U0001F44E', desc)
 			err = fmt.Errorf("Unable to run action %v: %v, output=%s", action, err, out)
--- a/pkg/ddevapp/addons.go
+++ b/pkg/ddevapp/addons.go
@@ -142,17 +142,23 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
 				// Get the exit code
 				exitCode := exitErr.ExitCode()
 				if exitCode == warningCode {
-					util.Warning("%s %s", "\U000026A0\U0000FE0F", desc)
+					if desc != "" {
+						util.Warning("%s %s", "\U000026A0\U0000FE0F", desc)
+					}
 					err = nil
 				}
 			}
 		}
 		if err != nil {
-			util.Warning("%c %s", '\U0001F44E', desc)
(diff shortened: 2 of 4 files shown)


hanoii:
maybe ⚠️  is better but it's a two character unicode so I couldn't use '%c' I don't mind changing the code to be able to use that, but wanted to jsut give this a try quickly in case you'd be willing to accept something like this.

hanoii:
I went ahead and modified to use the warning emoji
Screenshot 2025-05-05 at 12 04 25

pkg/ddevapp/addons.go
```diff
@@ -131,17 +133,27 @@ func ProcessAddonAction(action string, dict map[string]interface{}, bashPath str
...
 	if err != nil {
-		util.Warning("%c %s", '\U0001F44E', desc)
-		return fmt.Errorf("Unable to run action %v: %v, output=%s", action, err, out)
-	}
-	if desc != "" {
+		var exitErr *goexec.ExitError
+		if errors.As(err, &exitErr) {
+			// Get the exit code
+			exitCode := exitErr.ExitCode()
+			if exitCode == 63 {
```
GuySartorelli, line 142: Is there a reason for `63` specifically?
Do we have somewhere sensible we can set this as a constant so it's not just a magic number floating around?

hanoii: This is what anthropic had to say about this:

I went for 63 as in the higher application-specific error code based on that but
1. it could be anything
2. I can't say if that is right

> ## Warning Handling Approaches
> 
> While there's no standard "warning" exit code, some applications follow these patterns:
> 
> 1. **Non-Zero but Documented**: Some tools use specific exit codes like **1** or **2** for warnings and document this behavior
>    
> 2. **LSB (Linux Standard Base)** suggests ranges:
>    - 1-63: Application-specific error codes
>    - 64-113: More specific errors (sometimes following `sysexits.h`)
> 
> 3. **sysexits.h** defines error codes (64-78) for specific situations but doesn't include warnings specifically

hanoii:
I went one step ahead and added support for an annotation like `#ddev-description:`

```yaml
  - |
    #ddev-description:Checking DDEV version
    #ddev-warning-exit-code: 64
    if ! ${DDEV_EXECUTABLE} debug match-constraint ">= 1.24.5" 2>/dev/null ; then
      echo "This add-on works best with DDEV v1.24.5+ or higher, please upgrade." && exit 64
    fi
```

So now, no hardcoding of magic numbers
Screenshot 2025-05-06 at 08 50 08

hanoii:
Edited PR description with this new approach

stasadev (changes requested):
This is definitely an improvement, thanks!

Having the warning output after the description makes it much easier to understand.

`#ddev-warning-exit-code` needs to be mentioned in the docs, in the same section where `#ddev-description` is explained. I didn't understand what it did at first. After testing, I see it changes the script behavior so it doesn't stop other scripts on `exit`.

hanoii:
Docs added, I was treating this more as an idea as I wasn't sure you might accept it, but I think with this pseudio-annoation is a nice addition and fairly harmless.

hanoii:
it's not the same thing, it's a separate outcome, a warning, not a success and not an error. It displays the emoji and continue with other actions, but it lest you "exit" your script at any point.

stasadev:
Oh, that's right. I didn't notice that the text after the emoji is not green.

rfay:
It's wonderful that we can learn from your experience with your many add-ons, and benefit from your experience improving this.

AFAICT there are two completely different things being addressed here, true? One is your new feature (which is pretty specific to your setup) of whitelisting an error code, and the other is improving the order of output.

Would it be possible to do them in separate PRs? It feels to me like the first is just fixing something for you and the second is a very general usability improvement for everybody. 

If @stasadev is OK with bundling these, it works for me, but I would rather have had them in separate PRs so the reviewing experience would be more likely to be successful.

hanoii:
I can do it, I actually have rebase ready to push, the only thing is that both things touches the same code so it would save me time and also the PR already had discussion for both things. I let @stasadev weight in. But I will definitely keep this in mind, I actually normally do, but this on this one I wasn't that careful.

rfay:
One thing I note, this doesn't solve the current problem with older versions of DDEV, only a future situation like you currently have now.

docs/content/users/extend/additional-services.md
```diff
@@ -87,6 +87,7 @@ The `install.yaml` is a simple YAML file with a few main sections:
 In any stanza of `pre_install_actions` and `post_install_actions` you can:
 
 * Use `#ddev-description:<some description of what stanza is doing>` to instruct DDEV to output a description of the action it's taking.
+* Use `#ddev-warning-exit-code:<an integer between 1-255>` to instruct DDEV to consider a certain exit code of the stanza as a warning, not an error, continuing with other actions as normal.
```
rfay, line 90: Is it possible to accept multiple exit codes?

rfay (changes requested):
I'm sure we can accept this. 

It's not clear to me how to test the failure/warning mode at this point.

Maybe you can put up a dummy add-on (without the `ddev-get` label) for demonstration.

docs/content/users/extend/additional-services.md
hanoii: I have actually thought about this but I figured it made no sense, if your use case requires different exit codes you can bundle them together into a single one in your action:

```bash
...
ret1=$?
...
ret2=$?
if [[ "$ret1" == "2" || "$ret2" == "3" ]]; then
  exit 127
fi
```

hanoii:
> It's not clear to me how to test the failure/warning mode at this point.
> Maybe you can put up a dummy add-on (without the ddev-get label) for demonstration.

I am working on a test, I don't hink a dummy add-on repo is necessary, just picking up on the other addons test I am currently playing with:

```yaml
name: sample_actions

post_install_actions:
- touch test_cmd_addon_actions_output.txt
- |
  echo "action with output and no #ddev-description"
- |
  #ddev-description:Action with #ddev-description and output
  touch test_cmd_addon_actions_output.txt
  echo "test_cmd_addon_actions_output.txt created"
- |
  #ddev-description:Action that errs if .ddev/test_cmd_addon_actions_output_error.txt is present
  if [ -f "test_cmd_addon_actions_output_error.txt" ]; then
    echo "test_cmd_addon_actions_output_error.txt found!"
    exit 1
  fi
  echo "test_cmd_addon_actions_output_error.txt not found!"
- |
  #ddev-warning-exit-code:128
  exit 128
- |
  #ddev-warning-exit-code:128
  echo "action with output, #ddev-warning-exit-code and no #ddev-description"
  exit 128
- |
  #ddev-description:Action with #ddev-description and no output
  touch ~/.ddev/commands/web/global-touched
- |
  #ddev-description:Action with #ddev-warning-exit-code and #ddev-description and no output
  #ddev-warning-exit-code:127
  exit 127
- |
  #ddev-description:Final action doing nothing
```

And will parse its output. Just sharing for now in case I am going in the wrong direction.

Will pick up on this tomorrow.

stasadev:
I'm fine with having both changes in one PR since there aren't many changed lines.

hanoii:
@stasadev @rfay tests added and rebased.

hanoii:
Fixing tests

hanoii:
Done, the oddest thing, goland for some reason didn't trim out whitespaces from the end of a line. I couldn't get it to do it later either. Will try to find out why, but maybe you know.

rfay:
Please provide an actionable manual test that demonstrates both of the features of this PR, thanks.

hanoii:
> Please provide an actionable manual test that demonstrates both of the features of this PR, thanks.

Added it to the PR description.

rfay:
A small request: It just makes things a little easier for me if you use the [suggested branch naming](https://ddev.readthedocs.io/en/stable/developers/building-contributing/#feature-branch-name) in the future if you remember, this one could have been been named `20250505_hanoii_improve-add-on-action-output` or something.

rfay (approved):
OK, I can go with it. Thanks for the great manual testing instructions.

PR #7268, opened 2025-05-06
fix: support '>,x,X,*' comparisons and extended conditions in version constraint, fixes #7266

### LINKED ISSUE #7266: `ddev_version_constraint` provides incorrect result: `> 1.24.4` is true for `1.24.4`
This relates to how and why we added pre-releases as part of the code:

https://github.com/ddev/ddev/blob/c5ad791badefaf9ecb6f05a3768cd2039fa022b7/pkg/ddevapp/utils.go#L551-L555

So `> 1.24.4` becomes `> 1.24.4-0` and a full release (1.24.4) is actually higher.

I played a bit with the code but I really couldn't come up with a solution I like. Furthermore, now that I looked into this, I don't like the normalization at all, specially as the constraint can really really complex, so just appending `-0` might not work for all cases. We did it because of dev versions (IIRC). I wonder if we should simply exclude this and let the user add pre-releases to the constraints when needed.

i.e.

This is a command I am using to work around this: `ddev debug match-constraint "> 1.24.4-0 <= 1.24.4-zzzzzz || > 1.24.4"` which is rather odd.

Ideally, I'd expect that > 1.24.4 would work for regular versions and also accept something like 1.24.5-44

The current version on dev using the git tags gives you 1.24.4-commits-hash-dirty which, evenmore for the semver library, is just wrong because it only uses the last part of the dash as pre-release.

I don't have a solution so far. I am inclined to remove any normalization and let the user define the constraints, if not we might have to look for another semver library.

stasadev:
## The Issue

- #7266

Using DDEV v1.24.4:

```
$ ddev debug match-constraint ">1.24.4"
(no error, but it should fail)
```

> Ideally, I'd expect that > 1.24.4 would work for regular versions and also accept something like 1.24.5-44

- Doesn't add the `-0` suffix if it's not a prerelease, which makes this comparison work `>1.24.4` as expected
- Adds `-0` suffix to all parts of expression for prereleases, which makes extended comparisons `>1.24.0 <1.24.5` work as expected
- Adds support for wildcards `x`, `X`, `*` for prereleases.

```
$ make

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.4"
2025-05-06T14:48:56.747 Comparing constraint '>1.24.4-0' against version 'v1.24.4-38-gaaf8b5b83'

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.0 <1.24.5" 
2025-05-06T14:49:16.744 Comparing constraint '>1.24.0-0 <1.24.5-0' against version 'v1.24.4-38-gaaf8b5b83'

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.4-38"
2025-05-06T14:51:20.587 Comparing constraint '>1.24.4-38-0' against version 'v1.24.4-38-gaaf8b5b83'

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.4-39"
2025-05-06T14:51:33.586 Comparing constraint '>1.24.4-39-0' against version 'v1.24.4-38-gaaf8b5b83' 
error: your DDEV version 'v1.24.4-38-gaaf8b5b83' doesn't meet the constraint '>1.24.4-39'. Please update to a DDEV version that meets this constraint

$ DDEV_DEBUG=true ddev debug match-constraint "1.24.x"             
2025-05-06T16:44:32.33 Comparing constraint '1.24.x-0' against version 'v1.24.4-38-gaaf8b5b83-dirty'
```

```
$ make VERSION=v1.24.4

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.4"
2025-05-06T14:49:44.872 Comparing constraint '>1.24.4' against version 'v1.24.4' 
error: your DDEV version 'v1.24.4' doesn't meet the constraint '>1.24.4'. Please update to a DDEV version that meets this constraint

$ DDEV_DEBUG=true ddev debug match-constraint ">1.24.0 <1.24.5" 
2025-05-06T14:49:52.95 Comparing constraint '>1.24.0 <1.24.5' against version 'v1.24.4'

$ DDEV_DEBUG=true ddev debug match-constraint "1.24.x"
2025-05-06T16:45:05.369 Comparing constraint '1.24.x' against version 'v1.24.4'
```


pkg/ddevapp/utils.go
```diff
@@ -563,10 +565,26 @@ func CheckDdevVersionConstraint(constraint string, errorPrefix string, errorSuff
...
+	// Split the constraint into tokens based on spaces
+	tokens := strings.Fields(constraintNoCommas)
+	for i, token := range tokens {
+		last := token[len(token)-1]
+		// If the token represents a version number (ends with a digit)
+		// and doesn't contain a suffix '-0', append '-0'
+		if !strings.HasSuffix(token, "-0") && last >= '0' && last <= '9' {
+			tokens[i] = token + "-0"
+		}
+	}
```
hanoii, line 588: Why hardly used, this will probably fail wild https://github.com/Masterminds/semver?tab=readme-ov-file#wildcards-in-comparisons, right?

hanoii (commented):
Other than this comment this looks really great!

I thought (theoretically) of doing something similar, glad you got to it.

pkg/ddevapp/utils.go
stasadev: I agree, I'll add support for wildcards, thank you.

rfay (approved):
Thank you, this seems to sort everything out!

I could wish that we didn't have to work so hard on this for ... years :)

PR #7269, opened 2025-05-06
fix: WSL2 install script should wait longer for DDEV Windows installer to complete [skip ci]

rfay:
## The Issue

Discord: https://discord.com/channels/664580571770388500/1367962405237166233

Using the new (without chocolatey) script for WSL2 install, the PS script calls the Windows installer, and then waited only 10s for the completion.

Apparently the NSIS installer hands off installation of files to a child process and can run to completion before the child process is done. In this situation the reporter said it was an older computer. There are ways to wait for the child process, but it seems simpler just to increase the wait time, so we increase it from 10s to 60s here.

If we could reproduce on older computer it would be great.

### DIFF
--- a/scripts/install_ddev_wsl2_docker_inside.ps1
+++ b/scripts/install_ddev_wsl2_docker_inside.ps1
@@ -92,7 +92,7 @@ Write-Host "DDEV installation complete."
 
 $mkcertPath = "C:\Program Files\DDEV\mkcert.exe"
 $maxWait = 60
-Write-Host "Waiting $maxWait seconds for $mkcertPath binary..."
+Write-Host "Waiting up to $maxWait seconds for $mkcertPath binary..."
 $waited = 0
 while (-not (Test-Path $mkcertPath) -and $waited -lt $maxWait) {
     Start-Sleep -Seconds 1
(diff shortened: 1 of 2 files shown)


gconry18 (approved):
I was the original inquirer on discord. Happy with this change.

scripts/install_ddev_wsl2_docker_inside.ps1
```diff
@@ -91,7 +91,7 @@ $env:PATH += ";C:\Program Files\DDEV"
 Write-Host "DDEV installation complete."
 
 $mkcertPath = "C:\Program Files\DDEV\mkcert.exe"
-$maxWait = 10
+$maxWait = 60
```
stasadev, line 94:
```suggestion
$maxWait = 60
Write-Host "Waiting $maxWait seconds for $mkcertPath binary..."
```

stasadev (changes requested):
I'm okay with this change, but it's a silent wait, right?
We should at least say that we're waiting for something.

stasadev (approved):
Looks good to me!

rfay:
Now it probably needs a manual run again. Or maybe I'll pull it and use the instructions.

stasadev:
Works well
```
Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/rfay/ddev/refs/heads/20250506_rfay_longer_timeout_wsl2_install/scripts/install_ddev_wsl2_docker_inside.ps1'))
```

```
...
DDEV installation complete.
Waiting up to 60 seconds for C:\Program Files\DDEV\mkcert.exe binary...
The local CA is already installed in the system trust store! 👍
...
```

PR #7270, opened 2025-05-06
fix: allow status code 429 for linkspector

stasadev:
## The Issue

We get 429 from GitHub links for linkspector https://github.com/ddev/ddev/actions/runs/14861937842/attempts/2

It seems that GitHub changed some limits.

- https://github.com/UmbrellaDocs/action-linkspector/issues/43

I noticed that we use `GITHUB_TOKEN` for actions (this is not the env itself, it's name for custom action variable) instead of `github_token` (IDE highlighted it for me)

Adds an exception for 429 status.

### DIFF
--- a/.github/workflows/docscheck.yml
+++ b/.github/workflows/docscheck.yml
@@ -64,13 +64,13 @@ jobs:
       - name: Run textlint
         uses: tsuyoshicho/action-textlint@v3
         with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


stasadev:
I got the same error in https://github.com/ddev-test/ddev/actions/runs/14863320774/job/41733676200

stasadev:
Success with 429 status allowed in https://github.com/ddev-test/ddev/actions/runs/14863503281/job/41734319606

rfay (approved):
I guess we should do this for now.

PR #7273, opened 2025-05-08
docs: add --disable-settings-management suggestion on hosting

gbere:
## The Issue

After configuring `MAILER_DSN` in `.env.local` and restarting the project, it keeps getting overridden by `MAILER_DSN="smtp://127.0.0.1:1025"`. To prevent this, I had to run `ddev config --disable-settings-management` within the project.

### DIFF
--- a/docs/content/users/topics/hosting.md
+++ b/docs/content/users/topics/hosting.md
@@ -19,10 +19,10 @@ Here’s how to try it for yourself:
 5. Configure your project with [`ddev config`](../usage/commands.md#config).
 6. Import your database and files using [`ddev import-db`](../usage/commands.md#import-db) and [`ddev import-files`](../usage/commands.md#import-files).
 7. Save all databases with `ddev snapshot --all` before beginning, so you can restore your database if you have to change the name of the project.
-8. Tell DDEV to listen on all network interfaces, omit the SSH agent, use hardened images, and enable Let’s Encrypt:
+8. Tell DDEV to listen on all network interfaces, omit the SSH agent, use hardened images, enable Let’s Encrypt and ensure that the auto-configuration does not modify your variables after every restart:
 
     ```
-    ddev config global --router-bind-all-interfaces --omit-containers=ddev-ssh-agent --use-hardened-images --performance-mode=none --use-letsencrypt --letsencrypt-email=you@example.com
+    ddev config global --router-bind-all-interfaces --omit-containers=ddev-ssh-agent --use-hardened-images --performance-mode=none --use-letsencrypt --letsencrypt-email=you@example.com --disable-settings-management
     ```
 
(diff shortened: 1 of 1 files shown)


rfay (changes requested):
Thanks for this contribution!

Please update the PR template above to explain the motivation for the PR.

This is also completely incorrect, because `--disable-settings-management` is not a global option (not an option for `ddev config global`)

But at the project level, where `ddev config --disable-settings-management` does work, I haven't needed it or used it in my many years of using DDEV for hosting. 

So you'll need to explain the "why" of your concern.

rfay:
> After configuring MAILER_DSN in .env.local and restarting the project, it keeps getting overridden by MAILER_DSN="smtp://127.0.0.1:1025". To prevent this, I had to run ddev config --disable-settings-management within the project.

I assume your project wants to send mail. If you say more about what CMS you're hosting it will help. 

Changing the docs for everybody is not going to be the right way to solve your problem (but disable-settings-management is probably the right way to solve your problem.) But you're solving a generic problem by changing the docs for everybody that uses hosting.

gbere:
> > After configuring MAILER_DSN in .env.local and restarting the project, it keeps getting overridden by MAILER_DSN="smtp://127.0.0.1:1025". To prevent this, I had to run ddev config --disable-settings-management within the project.
> 
> I assume your project wants to send mail. If you say more about what CMS you're hosting it will help.
> 
> Changing the docs for everybody is not going to be the right way to solve your problem (but disable-settings-management is probably the right way to solve your problem.) But you're solving a generic problem by changing the docs for everybody that uses hosting.

I am hosting a Symfony project.
I believed that, in general, when hosting any type of project, you wouldn’t want variables like those for Mailpit to auto-configure themselves.
However, perhaps the most appropriate behavior would be that, if you have `use_hardened_images` enabled which disables Mailpit then the related environment variables should not be auto-configured.

rfay:
Thanks for stepping up to provide your experience!

It's interesting, this feature is normally handled by the php.ini, which uses `sendmail_path=/usr/local/bin/mailpit sendmail -t --smtp-addr 127.0.0.1:1025`, so you might have to override that as well. But I guess Symfony is not using the sendmail approach, so you're not affected. 

If you'd like to give some docs guidance about various settings that might have to be overridden, or do a PR to change what [DDEV does for symfony projects when using hardened images](https://github.com/ddev/ddev/blob/main/pkg/ddevapp/symfony.go#L23-L38) it's welcome.  You also might want to open an issue explaining the challenge and your solution. And  a guest blog post on ddev.com would be welcome, super easy to do. https://github.com/ddev/ddev.com

But this PR as is won't be useful to most people besides you you, it requires much more finesse.  Closing for now, but your continued collaboration is very welcome. Happy to continue the conversation here or in [discord](https://ddev.com/s/discord).

PR #7274, opened 2025-05-08
docs: update Craft CMS quickstart for PR #7236

timkelty:
## The Issue

- #7236

See PR for details.

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -164,7 +164,7 @@ Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing on
 !!!tip "Compatibility with Craft CMS 3"
     The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
-Environment variables will be automatically added to your `.env` file to simplify the first boot of a project. For _new_ installations, this means the default URL and database connection settings displayed during installation can be used without modification. If _existing_ projects expect environment variables to be named in a particular way, you are welcome to rename them.
+Environment variables will be automatically added to the `.ddev/.env.web` file to simplify project configuration. This means that the primary site URL and the database connection settings can be used without any modification. To disable this behavior, see [`disable_settings_management`](configuration/config/#disable_settings_management).
 
 === "New projects"
 
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
     ddev config --project-type=craftcms --docroot=web
 
     # Boot the project and install the starter project:
-    ddev start
     ddev composer create-project --no-scripts craftcms/craft
-    ddev craft install/craft \
+
+    # Install Craft CMS:
+    ddev craft install \
         --username=admin \
         --password=Password123 \
         --email=admin@example.com \
-        --site-name='$DDEV_PROJECT' \
         --language=en \
-        --site-url='$DDEV_PRIMARY_URL'
+        --site-url='$PRIMARY_SITE_URL'
 
     # Login using `admin` user and `Password123` password
-    ddev launch
+    ddev launch admin
     ```
 
     Third-party starter projects can by used the same way—substitute the package name when running `ddev composer create-project`.
@@ -205,11 +205,10 @@ Environment variables will be automatically added to your `.env` file to simplif
     ddev config --project-type=craftcms
 
     # Boot the project and install Composer packages:
-    ddev start
     ddev composer install
 
     # Import a database backup and open the site in your browser:
-    ddev import-db --file=/path/to/db.sql.gz
+    ddev craft db/restore /path/to/db.sql.gz
     ddev launch
     ```
 


docs/content/users/quickstart.md
```diff
@@ -164,7 +164,7 @@ Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing on
 !!!tip "Compatibility with Craft CMS 3"
     The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
-Environment variables will be automatically added to your `.env` file to simplify the first boot of a project. For _new_ installations, this means the default URL and database connection settings displayed during installation can be used without modification. If _existing_ projects expect environment variables to be named in a particular way, you are welcome to rename them.
```
timkelty, line 167: Since the `.ddev/.env.web` is always created, the distinction between new/existing doesn't really matter. `CRAFT_DB_` variables will always override.

docs/content/users/quickstart.md
```diff
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
...
-    ddev start
     ddev composer create-project --no-scripts craftcms/craft
-    ddev craft install/craft \
+
+    # Install Craft CMS:
+    ddev craft install \
         --username=admin \
         --password=Password123 \
         --email=admin@example.com \
-        --site-name='$DDEV_PROJECT' \
```
timkelty, line 185: This is awkard:
- `--site-name` should be the "friendly name" of a project
- `$DDEV_PROJECT` (not the resolved value) will get persisted to `config/project.yaml`, which you don't want.

Best to just let the prompt ask them, or hardcode it

docs/content/users/quickstart.md
```diff
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
...
         --language=en \
-        --site-url='$DDEV_PRIMARY_URL'
```
timkelty, line 187: Since this gets persisted to yaml, we don't want the `DDEV_` version.

docs/content/users/quickstart.md
```diff
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
...
+        --site-url='$PRIMARY_SITE_URL'
 
     # Login using `admin` user and `Password123` password
-    ddev launch
```
timkelty, line 190: Bring them to the control panel login.

docs/content/users/quickstart.md
```diff
@@ -205,11 +205,10 @@ Environment variables will be automatically added to your `.env` file to simplif
     ddev config --project-type=craftcms
 
     # Boot the project and install Composer packages:
-    ddev start
     ddev composer install
 
     # Import a database backup and open the site in your browser:
-    ddev import-db --file=/path/to/db.sql.gz
```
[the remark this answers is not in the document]
rfay: `ddev import-db` does postgres importing just fine...

docs/content/users/quickstart.md
```diff
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
...
         --password=Password123 \
         --email=admin@example.com \
-        --site-name='$DDEV_PROJECT' \
         --language=en \
-        --site-url='$DDEV_PRIMARY_URL'
+        --site-url='$PRIMARY_SITE_URL'
 
     # Login using `admin` user and `Password123` password
-    ddev launch
+    ddev launch admin
```
timkelty, line 191: Bring them to the control panel login.

docs/content/users/quickstart.md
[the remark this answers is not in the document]
timkelty: Gotcha.
I'm fine with whatever you prefer here, then.

As-is, `craft db/backup` and `craft db/restore` will use `mysql`/`mysqldump`/`pg_dump`/`pg_restore` from the web container.

docs/content/users/quickstart.md
stasadev: We can't assume people using this quickstart have the latest DDEV, so it might not work for them.
(BTW, we use `$DDEV_PRIMARY_URL` for many quickstarts.)

docs/content/users/quickstart.md
```diff
@@ -164,7 +164,7 @@ Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing on
 !!!tip "Compatibility with Craft CMS 3"
     The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
-Environment variables will be automatically added to your `.env` file to simplify the first boot of a project. For _new_ installations, this means the default URL and database connection settings displayed during installation can be used without modification. If _existing_ projects expect environment variables to be named in a particular way, you are welcome to rename them.
+Environment variables will be automatically added to the `.ddev/.env.web` file to simplify project configuration. This means that the primary site URL and the database connection settings can be used without any modification. To disable this behavior, see [`disable_settings_management`](configuration/config/#disable_settings_management).
```
stasadev, line 167:
```suggestion
Environment variables will be automatically added to the `.ddev/.env.web` file to simplify project configuration. This means that the primary site URL and the database connection settings can be used without any modification. To disable this behavior, see [`disable_settings_management`](./configuration/config.md#disable_settings_management).
```

docs/content/users/quickstart.md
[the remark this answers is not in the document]
rfay: Fine to suggest the craft one.

docs/content/users/quickstart.md
rfay: I would normally put /admin, but it doesn't matter.

rfay (changes requested):
Changes to the quickstart require updates to the quickstart test, docs/tests/craftcms.bats,

docs/content/users/quickstart.md
timkelty: @stasadev not the end of the world if we don't change it – but in Craft's case, using it gets very awkward: The unparsed value `$DDEV_PRIMARY_URL` ends up getting built into a yaml config.

While it is normal to reference an env var in the yaml like this (especially for site url), you would never want the var to be called `DDEV_` as that would indicate that you would name it that way in another/remote environment.

timkelty:
> Changes to the quickstart require updates to the quickstart test, docs/tests/craftcms.bats,

Noted – I likely wont have time to get to that until next week sometime.

rfay:
Moved to draft, this has to wait for
* https://github.com/ddev/ddev/pull/7236

and has to get quickstart test update.

docs/content/users/quickstart.md
```diff
@@ -176,18 +176,18 @@ Environment variables will be automatically added to your `.env` file to simplif
     ddev config --project-type=craftcms --docroot=web
 
     # Boot the project and install the starter project:
-    ddev start
```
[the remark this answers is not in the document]
timkelty: Fair enough

stasadev (approved):
Thank you, it looks good to me.

Could you please update one more thing here https://ddev--7274.org.readthedocs.build/en/7274/users/quickstart/#craft-cms-existing-projects

image

I noticed we use "generic" for the `php` project type. That text was added a long time ago, but now we have a dedicated `generic` project type, so the tip might be confusing. I propose we remove or replace "generic" with "php":

```diff
-!!!tip "Upgrading or using a generic project type?"
+!!!tip "Upgrading or using a `php` project type?"
-    If you previously set up DDEV in a Craft project using the generic `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
+    If you previously set up DDEV in a Craft project using the `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
```

timkelty:
> Thank you, it looks good to me.
> 
> Could you please update one more thing here https://ddev--7274.org.readthedocs.build/en/7274/users/quickstart/#craft-cms-existing-projects
> 
> image
> 
> I noticed we use "generic" for the `php` project type. That text was added a long time ago, but now we have a dedicated `generic` project type, so the tip might be confusing. I propose we remove or replace "generic" with "php":
> 
> ```diff
> -!!!tip "Upgrading or using a generic project type?"
> +!!!tip "Upgrading or using a `php` project type?"
> -    If you previously set up DDEV in a Craft project using the generic `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
> +    If you previously set up DDEV in a Craft project using the `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
> ```

All set.

rfay:
Shouldn't we just consider removing that warning now? It's been a few years. Seems obsolete?

docs/content/users/quickstart.md
```diff
@@ -205,18 +209,17 @@ Environment variables will be automatically added to your `.env` file to simplif
...
+    ddev craft db/restore /path/to/db.sql.gz
     ddev launch
     ```
 
     Craft CMS projects use MySQL 8.0, by default. You can override this setting (and the PHP version) during setup with [`config` command flags](./usage/commands.md#config) or after setup via the [configuration files](./configuration/config.md).
 
-    !!!tip "Upgrading or using a generic project type?"
-        If you previously set up DDEV in a Craft project using the generic `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
+    !!!tip "Upgrading or using a `php` project type?"
+        If you previously set up DDEV in a Craft project using the `php` project type, update the `type:` setting in `.ddev/config.yaml` to `craftcms`, then run [`ddev restart`](../users/usage/commands.md#restart) apply the changes.
```
rfay, line 222:
```suggestion
```

rfay:
Rebased. Ready for you @rpkoller , sorry I jumped the gun before.

rfay (approved):
Thanks, seems good to me!

docs/content/users/quickstart.md
```diff
@@ -164,7 +164,7 @@ Start a new [Craft CMS](https://craftcms.com) project or retrofit an existing on
 !!!tip "Compatibility with Craft CMS 3"
     The `craftcms` project type is best with Craft CMS 4+, which is more opinionated about some settings. If you are using Craft CMS 3 or earlier, you may want to use the `php` project type and [manage settings yourself](https://github.com/ddev/ddev/issues/4650).
 
-Environment variables will be automatically added to your `.env` file to simplify the first boot of a project. For _new_ installations, this means the default URL and database connection settings displayed during installation can be used without modification. If _existing_ projects expect environment variables to be named in a particular way, you are welcome to rename them.
+Environment variables will be automatically added to the `.ddev/.env.web` file to simplify project configuration. This means that the primary site URL and the database connection settings can be used without any modification. To disable this behavior, see [`disable_settings_management`](./configuration/config.md#disable_settings_management).
```
rpkoller, line 167:
```suggestion
Environment variables will be automatically added to the `.ddev/.env` file to simplify project configuration. This means that the primary site URL and the database connection settings can be used without any modification. To disable this behavior, see [`disable_settings_management`](./configuration/config.md#disable_settings_management).
```


With https://github.com/ddev/ddev/pull/7236 in, i've retested with the instructions from this PR... `.ddev/.env.web` i can't find, it looks like the created env file is just  `.ddev/.env`

rpkoller (commented):
found one detail to note about the naming of the .env file

docs/content/users/quickstart.md
rpkoller: and i just compared .env with the other three example files created, and i wonder would it make sense to extend the content of the .env file with the settings from the other example files and then use the ddev database credentials there with db?

rfay: It's .ddev/.env.web, so if you're seeting anything else, something is wrong.  And there are no example files in play here, if there's a mention of any example files, that's also wrong, because we're not using any.

rpkoller: hm after upgrading to `/opt/homebrew/Cellar/ddev/HEAD-bf731bc` i followed the steps in the issue summary https://ddev--7274.org.readthedocs.build/en/7274/users/quickstart/#craft-cms 

```
ddev composer create-project --no-scripts craftcms/craft
ddev craft install \
    --username=admin \
    --password=Password123 \
    --email=admin@example.com \
    --site-name='My Craft Site' \
    --language=en \
    --site-url='$PRIMARY_SITE_URL'
```

and the first three lines of output of the craft install command were: 

```
Generating an application ID ... 
A .env file doesn't exist at /var/www/html/.env. Would you like to create one? (yes|no) [yes]:
/var/www/html/.env created. Note you still need to set up PHP dotenv for its values to take effect.
```
just retired and happend on the next try as well. but now i understood. 

in project root i have:

```
drwxr-xr-x  25 rkoller  staff   800B May 14 03:16 .ddev
-rw-------   1 rkoller  staff   113B May 14 03:17 .env
-rw-------   1 rkoller  staff   613B May 14 03:16 .env.example.dev
-rw-------   1 rkoller  staff   623B May 14 03:16 .env.example.production
-rw-------   1 rkoller  staff   619B May 14 03:16 .env.example.staging
``` (and more)

the .env contains the craft app id and craft security key while `.ddev/.env.web` is now there and contains: 

```
CRAFT_DB_SERVER="db"
CRAFT_DB_USER="db"
MAILPIT_SMTP_PORT="1025"
CRAFT_DB_PORT="3306"
CRAFT_DB_DATABASE="db"
CRAFT_DB_PASSWORD="db"
MAILPIT_SMTP_HOSTNAME="127.0.0.1"
CRAFT_DB_DRIVER="mysql"
CRAFT_WEB_ROOT="/var/www/html/web"
PRIMARY_SITE_URL="https://lo.ddev.site"
```
and i ran `ddev ssh`in the previous test and in that root i also got 

```
drwx------ 19 rkoller dialout 4.0K May 14 03:16 .ddev
-rw-r--r--  1 rkoller dialout  113 May 14 03:17 .env
-rw-r--r--  1 rkoller dialout  613 May  7 21:12 .env.example.dev
-rw-r--r--  1 rkoller dialout  623 May  7 21:12 .env.example.production
-rw-r--r--  1 rkoller dialout  619 May  7 21:12 .env.example.staging
```
and rechecking .ddev contains also .env.web

not sure what threw me off but i guess the following prompt distracted me and brought me on the wrong track `A .env file doesn't exist at /var/www/html/.env. Would you like to create one? (yes|no) [yes]:` so everything i behaving as expected it looks like. so you can forget my suggestion. false alarm

Comment (rfay) [reply docs/content/users/quickstart.md:167]: The craft .env now has very little in it. That one is in the project root and is no longer DDEV's responsibility

The key credentials (database credentials) go into the .ddev/.env.web, and they are DDEV's responsibility

The .ddev/.env.web is created on `ddev config` and you do not even need to start to look at it.

Comment (timkelty) [reply docs/content/users/quickstart.md:167]: ↑ what @rfay said. `<project>/.env` is fully the responsibility of the `craft install` command, and what ends up in there are the app's responsibility, not DDEV.

Comment (rfay) [reply docs/content/users/quickstart.md:167]: We'll go with this for now, and @timkelty if it seems like people are stumbling on that spot it can be improved.

Comment (timkelty) [reply docs/content/users/quickstart.md:167]: @rpkoller @rfay maybe it would help if DDEV _did_ write to the `<project>/.env` with something like:

```
# Additional environment variables loaded in `.ddev/.env.web`
```

rfay: That's an interesting suggestion, pretty easy, but OTOH I'm going to hate touching the project .env now, since it's truly owned by Craft CMS.

timkelty: > That's an interesting suggestion, pretty easy, but OTOH I'm going to hate touching the project .env now, since it's truly owned by Craft CMS.

Good point – we could just add that comment in our composer starter if we really wanted.

PR #7275, opened 2025-05-08
feat: add TYPO3 create-sitepackage global command

rfay:
## The Issue

TYPO3's complexity of creating a default site package; we'd like to have a simple way for n00bs to get started with a sitepackage, which is fundamentally required to do anything.

Try to provide a TYPO3-only global web command that creates a site package (theme)

This is taken as suggested by @linawolf from the files in https://github.com/TYPO3-Documentation/site-introduction/tree/main/Build/DownloadSitePackage

`ddev create-sitepackage` in a TYPO3 v13 composer-build

It should create the site package and composer-install it.

## Automated Testing Overview

- [x] If this is successful, it needs a test or at least a mention in a test.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/web/create-sitepackage
+++ b/pkg/ddevapp/global_dotddev_assets/commands/web/create-sitepackage
@@ -38,9 +38,6 @@ curl -X 'POST' \
   -H 'Content-Type: application/json' \
   -d @"/tmp/data.json" --output /tmp/my_site_package.zip
 
-composer require my-vendor/my-site-package:@dev
-
-rm -rf packages/my_site_package/*
-rm -rf packages/my_site_package/.*
 unzip /tmp/my_site_package.zip -d "packages/my_site_package/"
+composer require my-vendor/my-site-package:@dev
 rm -f /tmp/my_site_package.zip /tmp/data.json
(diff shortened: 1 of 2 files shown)


pkg/ddevapp/global_dotddev_assets/commands/web/create-sitepackage
```diff
@@ -0,0 +1,46 @@
...
+  'https://get.typo3.org/api/v1/sitepackage/' \
+  -s \
+  --fail \
+  -H 'accept: application/zip' \
+  -H 'Content-Type: application/json' \
+  -d @"/tmp/data.json" --output /tmp/my_site_package.zip
+
+composer require my-vendor/my-site-package:@dev
+
+rm -rf packages/my_site_package/*
```
linawolf, line 43: I think we should not remove the site package if one exist unless some kind of `--force` parameter was set

rfay: Your wish is my command. Updated.

rfay:
Overall, I think this is a pretty bad patch by DDEV on a pretty big shortcoming of TYPO3 (for beginners). But I don't see it doing any great harm.

Note that after running this (which composer-requires the new sample site package) you have to add the "site settings" in the "site configuration" to make it work.

linawolf:
i dont really like the current solution. I will do more thinking about this

rfay:
Moving to draft pending TYPO3 community review, especially @linawolf

rfay:
This certainly made things easier for me, but this doesn't have any traction from the TYPO3 community, and certainly seems like a problem that should be solved in TYPO3 rather than in DDEV. I'll keep the custom command for testing though!

Happy to reopen if there's interest from TYPO3.

PR #7277, opened 2025-05-08
fix: make xdebug work with WSL2 mirrored mode, fixes #7272

### LINKED ISSUE #7272: Xdebug doesn't work on WSL2 with networkingMode=mirrored (host.docker.internal incorrect/unreachable)

Edited:

I'm using networkingMode=mirrored in WSL2 with the docker-ce WSL2 approach ("docker inside"). 

Xdebug doesn't seem to work; inside the web container the hostname `host.docker.internal` resolves to an IP address that does not reach the listening IDE on port 9003. 

I'd like XDebug to work.

#### Original report is below

My ddev installation works fine under WSL except for this minor bug when I start containers:

```
Unable to validate IP address 'eth0' from 'ip -4 route show default'
Unable to validate IP address 'eth0' from 'ip -4 route show default'
```

The reason for this is because that command is only for WSL under NAT mode, per Microsoft sources:
https://learn.microsoft.com/en-us/windows/wsl/networking

Everything seems to work fine (including internet for inside containers, accessing them from the host, etc), so the message is probably just a code bug because I don't see to have any problems except with Xdebug which is the reason for this question. From the same page it says that using mirror mode you can connect directly to the ports using localhost or 127.0.0.1

I had no luck connecting to the PHPStorm Xdebug from DDEV, I know it's working fine from other systems, just ddev has a problem because the routing from docker container can't connect.

From inside WSL I can connect to the port, example:
`nc -zv 127.0.0.1 9003`

`Connection to 127.0.0.1 9003 port [tcp/*] succeeded!`~~~

When I SSH into the php web container, I can't because 127.0.0.1 inside the container is not resolving to the WSL host. ddev is using its own isolated network configuration. So xdebug fails.

Is there a setting in ddev to tell it to use mirror mode for my containers? I know mirror mode is supported in Docker Desktop but I am actually running the docker from inside WSL because this is what the official docs suggested as the preferred option over Docker Desktop on the window host.

So the entire issue is basically ddev not detecting here that WSL is under mirror networking mode.

### Describe your solution

ddev detect if WSL is under NAT or Mirror mode

### Describe alternatives

Set proper networking for docker as mirror for containers instead of NAT.

### Additional context

_No response_

rfay:
## The Issue

- #7272

WSL2 Mirrored Mode is yet another kink in the already complex WSL2 landscape. Containers can only connect to Windows-side IP addresses, not to, for example. `docker0` in WSL2.

* Derive a candidate Windows-side IP address to connect to and provide it as host.docker.internal
* Update `ddev debug test` to give this kind of information.
* See https://github.com/ddev/ddev/issues/7272#issuecomment-2862942816

## TODO

- [ ] I pretty much hate the (ChatGPT-assisted) PowerShell snippet being used to determine the appropriate IP address. Oh would I like something better. There might be a way to run a proxy in the WSL2 distro... or ???
- [x] Add docs for WSL2 configuration (~/.wslconfig, including `hostAddressLoopback=true`
- [x] It's conceivable that now that we have a way of detecting mirrored mode we could improve the timeout handling in https://github.com/ddev/ddev/blob/82c1988f0a11bb5d9474439c486f454b15b1334b/pkg/netutil/netutil.go#L41-L51

- [ ] This was tested with WSL pre-release 2.5.7.0. It probably works with current stable 2.4.x
- [ ] Use Xdebug with WSL2 in mirrored mode against an IDE on Windows. You must also have the experimental `hostAddressLoopback=true`. (This may require a WIndows reboot to be successful.) Example ~/.wslconfig (on Windows side):
```
[wsl2]
networkingMode=mirrored

[experimental]
hostAddressLoopback=true
```
- [ ] Verify that we can use Xdebug in WSL2 in mirrored mode with an IDE running in WSL2

## Automated Testing Overview

* No changed tests at this point, and we'll be forced to run a mirrored mode WSL2 test runner at some point.

This passes manual run of TestDdevXdebugEnabled and TestCmdXdebug on WSL2 as configured above

### DIFF
--- a/cmd/ddev/cmd/scripts/test_ddev.sh
+++ b/cmd/ddev/cmd/scripts/test_ddev.sh
@@ -62,6 +62,13 @@ header "installed DDEV add-ons"
 
 ddev add-on list --installed
 
+header "WSL2 information"
+
+if command -v wslinfo >/dev/null ; then
+  echo "WSL version=$(wsl.exe --version | tr -d '\0')";
+  echo "WSL2 networking mode=$(wslinfo --networking-mode)"
+fi
+
 header "mutagen situation"
 
 echo "looking for #ddev-generated in mutagen.yml in project ${PWD}"
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -36,16 +36,19 @@ These environments can be extended, version controlled, and shared, so you can t
 
     * RAM: 8GB
     * Storage: 256GB
-    * [Docker Desktop](https://www.docker.com/products/docker-desktop/) on the Windows side or [Docker CE](https://docs.docker.com/engine/install/ubuntu/) inside WSL2
+    * [Docker CE](https://docs.docker.com/engine/install/ubuntu/) inside WSL2 or [Docker Desktop](https://www.docker.com/products/docker-desktop/) on the Windows side.
     * Ubuntu or an Ubuntu-derived distro is recommended, though others may work fine
 
     **Next steps:**
 
     *You’ll need a Docker provider on your system before you can install DDEV.*
     
-    1. Install Docker with [recommended settings](users/install/docker-installation.md#windows).
-    2. Install [DDEV for Windows](users/install/ddev-installation.md#windows).
-    3. Launch your [first project](users/project.md) and start developing. 🚀
+    1. Install [DDEV for Windows](users/install/ddev-installation.md#windows).
+    2. Launch your [first project](users/project.md) and start developing. 🚀
+
+    **WSL2 in Mirrored Mode**
+
+    If you're using Windows WSL2 with the "Mirrored" Networking Mode, set the experimental `hostAddressLoopback=true` feature.
 
 === "Traditional Windows"
 
--- a/docs/content/users/install/ddev-installation.md
+++ b/docs/content/users/install/ddev-installation.md
@@ -259,7 +259,10 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
 
     7. For unusual browsers and situations that don't automatically support the `mkcert` certificate authority, [configure your browser](configuring-browsers.md).
 
-    Now you can use the "Ubuntu" terminal app or Windows Terminal to access your Ubuntu distro, which has DDEV and Docker Desktop integrated with it.
+    Now you can use the "Ubuntu" terminal app or Windows Terminal to access your Ubuntu distro.
+
+    !!!note "WSL2 with Networking Mode set to 'Mirrored'"
+        If you're using the ['Mirrored' networking mode](https://learn.microsoft.com/en-us/windows/wsl/networking#mirrored-mode-networking) with WSL2, you need to also set the experimental `hostAddressLoopback=true` in your `~/.wslconfig` file.
 
     ### WSL2/Docker Desktop Manual Installation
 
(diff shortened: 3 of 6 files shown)


rfay:
@marvinhinz @Yannici Could you please review and test this along with @SyntaxDreamer, thanks!

stasadev (commented):
I tested it using local virtual machine and https://xdebug.org/docs/dbgpClient (running from Windows binary, because I don't have any IDE inside this VM).

NAT (xdebug works):

no `~/.wslconfig`

```
ddev start
ddev xdebug
```

Then I open `dbgpClient.exe` in the new window, and `ddev ssh`, ping to `host.docker.internal` doesn't work (I don't know if it should work at all), and then I run `composer require psr/log`, which triggers Xdebug connection, which appears in `dbgpClient.exe` window.

nat

---

Mirrored (xdebug doesn't work):

with `~/.wslconfig`

```
[wsl2]
networkingMode=mirrored

[experimental]
hostAddressLoopback=true
```

```
ddev start
ddev xdebug
```

Then I open `dbgpClient.exe` in the new window, and `ddev ssh`, ping to `host.docker.internal` works, and `composer require psr/log` is not debugging anything.

mirrored

rfay:
Sounds like the IP address is not successfully being determined, which is definitely a risk.

In this situation I've been using ncat.exe to do the listening:

`ncat.exe -k -l 0.0.0.0 9003` - easier than dbgpClient

rfay:
If you can show `ipconfig.exe` it may give us a clue.

stasadev:
```
ipconfig.exe

Windows IP Configuration


Ethernet adapter Ethernet0:

   Connection-specific DNS Suffix  . : localdomain
   Link-local IPv6 Address . . . . . : fe80::cb6:1279:884a:2ab0%9
   IPv4 Address. . . . . . . . . . . : 192.168.185.143
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : 192.168.185.2
```

stasadev:
Sorry, I forgot to update WSL2

```
PS C:\Users\testbot> wsl --version
WSL version: 2.2.4.0
Kernel version: 5.15.153.1-2
WSLg version: 1.0.61
MSRDC version: 1.2.5326
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26091.1-240325-1447.ge-release
Windows version: 10.0.22631.3737

PS C:\Users\testbot> wsl --update
Checking for updates.
Updating Windows Subsystem for Linux to version: 2.4.13.

PS C:\Users\testbot> wsl --version
WSL version: 2.4.13.0
Kernel version: 5.15.167.4-1
WSLg version: 1.0.65
MSRDC version: 1.2.5716
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.22631.3737
```

Tested it again and now it's working:

image

stasadev (approved):
Looks good to me, and works well.

It's amazing to have this level of support for Xdebug.

rfay:
Thanks! Looking at it from the risk point of view, so you think it's safe for everybody else, whether or not it works for every single wsl2-with-mirrored-mode user ?

stasadev:
I think it's good enough and won't break the existing WSL2 NAT. I don't prefer using PowerShell here either, but I don't have a better solution.

PR #7283, opened 2025-05-13
feat: add shell completion for --service (-s) flag

stasadev:
## The Issue

We don't have autocompletion for services.

Adds it.

```
ddev start
ddev exec -s <TAB><TAB>
ddev logs -s <TAB><TAB>
ddev ssh -s <TAB><TAB>
ddev debug rebuild -s <TAB><TAB> # should contain all services

ddev logs d11 -s <TAB><TAB>
ddev ssh d11 -s <TAB><TAB>
ddev debug rebuild d11 -s <TAB><TAB> # should contain all services
```

```
ddev add-on get ddev/ddev-redis
ddev restart
ddev exec -s <TAB><TAB> # see redis
```

```
ddev stop
ddev exec -s <TAB><TAB> # should be empty
ddev logs -s <TAB><TAB> # should be empty
ddev ssh -s <TAB><TAB> # should be empty
ddev debug rebuild -s <TAB><TAB> # should contain all services

ddev logs d11 -s <TAB><TAB> # should be empty
ddev ssh d11 -s <TAB><TAB> # should be empty
ddev debug rebuild d11 -s <TAB><TAB> # should contain all services
```

### DIFF
--- a/pkg/ddevapp/utils.go
+++ b/pkg/ddevapp/utils.go
@@ -490,6 +490,33 @@ func GetProjectNamesFunc(status string, numArgs int) func(*cobra.Command, []stri
 	}
 }
 
+// GetServiceNamesFunc returns a function for autocompleting service names for service flag.
+// If existingOnly is true, only names of existing services will be returned.
+func GetServiceNamesFunc(existingOnly bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+	return func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+		app, err := GetActiveApp("")
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		rawServices, ok := app.ComposeYaml["services"].(map[string]interface{})
+		if !ok {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		var services []string
+		for service := range rawServices {
+			if existingOnly {
+				c, err := app.FindContainerByType(service)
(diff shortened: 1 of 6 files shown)


pkg/ddevapp/utils.go
```diff
@@ -490,6 +490,33 @@ func GetProjectNamesFunc(status string, numArgs int) func(*cobra.Command, []stri
 	}
 }
 
+// GetServiceNamesFunc returns a function for autocompleting service names for service flag.
+// If existingOnly is true, only names of existing services will be returned.
+func GetServiceNamesFunc(existingOnly bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+	return func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+		app, err := GetActiveApp("")
```
rfay, line 497: This makes it not work right everywhere because it's getting the current app, rather than a possibly named app. For example, `ddev ssh craftdemo -s <tab>` gets the services for the current app instead of "creaftdemo".

rfay (changes requested):
Overall, this works great and is a nice addition, thanks!

And thanks @GuySartorelli for all your work in this area that has made DDEV so much easier to work with! 

The only thing is the assumption that the target is in the current directory, which is not necessarily true.

pkg/ddevapp/utils.go
stasadev: Thanks, I rarely use this feature with project arguments, and I forgot about it.

rfay (approved):
Works great, thanks for the fix.

PR #7284, opened 2025-05-13
feat: restart only specified service in `ddev debug rebuild`

stasadev:
## The Issue

From StackOverflow https://stackoverflow.com/questions/79522674/how-to-restart-a-single-ddev-service-docker-container-without-restarting-the-f/79523060?noredirect=1#comment140420056_79523060

Improves `ddev debug rebuild`, so it doesn't restart the entire project when it's not needed.

```
# get additional containers
ddev add-on get ddev/ddev-opensearch
ddev add-on get ddev/ddev-phpmyadmin
ddev restart

# break them
ddev exec -s opensearch kill 1
ddev exec -s phpmyadmin kill 1
ddev describe

# rebuild and restart opensearch using cache
ddev debug rebuild -s opensearch --cache
ddev describe

# rebuild and restart phpmyadmin without cache
ddev debug rebuild -s phpmyadmin
ddev describe
```

### DIFF
--- a/cmd/ddev/cmd/debug-rebuild.go
+++ b/cmd/ddev/cmd/debug-rebuild.go
@@ -1,13 +1,13 @@
 package cmd
 
 import (
-	exec2 "github.com/ddev/ddev/pkg/exec"
-	"github.com/ddev/ddev/pkg/globalconfig"
-	"strings"
 	"time"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/dockerutil"
+	exec2 "github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/globalconfig"
+	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/output"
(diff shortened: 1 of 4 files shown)


stasadev:
@MurzNN, please review.

cmd/ddev/cmd/debug-rebuild.go
```diff
@@ -54,44 +54,87 @@ var DebugRebuildCmd = &cobra.Command{
...
+		}
+
+		labels := map[string]string{
+			"com.ddev.site-name":         app.GetName(),
+			"com.docker.compose.oneoff":  "False",
+			"com.docker.compose.service": service,
+		}
+
+		// Restart the specified service using docker-compose, if it is running
+		if container, err := dockerutil.FindContainerByLabels(labels); err == nil && container != nil {
```
stasadev, line 111: Note: I changed the logic here to restart the service only if it's already running, since the command is called "rebuild". There's no reason to start it if the project isn't running.

I kept the restart logic for the default case with no arguments, so it's backward compatible with the original `ddev debug rebuild` (`ddev debug refresh`).

PR #7286, opened 2025-05-13
fix: Make sure env-type settings are populated before start, don't override `.ddev/.gitignore` file, for #7236

rfay:
## The Issue

- #7236
- Discord (breakage of .ddev/.gitignore on craft in v1.24.5): https://discord.com/channels/664580571770388500/1372597483443589264

In https://github.com/ddev/ddev/pull/7236#issuecomment-2876493614 @stasadev pointed out that now that we're using .ddev/.env* for settings (instead of using php application-defined settings load techniques) the settings/env values may be loaded too late (after containers are up) in some situations.

Before #7236 I believe that all explicit use of settings, including project types that used project-level .env (loaded by PHP dotenv for example) were application-loaded, which meant that it didn't matter that we wrote the .env late (post-start). But now with #7236 we're using DDEV's .ddev/.env.* so those only get loaded during `docker-compose up`, and need to be written before `docker-compose up`. 

---

The CraftCMS config overrides `.ddev/.gitignore` completely (which is wrong), and our logic for settings files creates `.gitignore` file in the directory of the settings file. Since CraftCMS uses `.ddev/.env.web`, its parent is `.ddev`, so it used `.ddev/.gitignore`.

Move the call to app.CreateSettingsFile() in the start sequence to before the `docker-compose up`

Don't override `.ddev/.gitignore` for CraftCMS. 

Mostly automated tests should tell us about this; 

A good check of Craft CMS in light of https://github.com/ddev/ddev/pull/7236#issuecomment-2876493614 is a good idea

## Automated Testing Overview

No changes.

### DIFF
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -311,7 +311,9 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 
 		// Don't create gitignore if it would be in top-level directory, where
 		// there is almost certainly already a gitignore (like Backdrop)
-		if path.Dir(app.SiteSettingsPath) != app.AppRoot {
+		// Don't override the existing .ddev/.gitignore as well
+		// TODO: we may want to append to .ddev/.gitignore if needed.
+		if path.Dir(app.SiteSettingsPath) != app.AppRoot && path.Dir(app.SiteSettingsPath) != app.GetConfigPath("") && app.SiteDdevSettingsFile != "" {
 			if err = CreateGitIgnore(filepath.Dir(app.SiteSettingsPath), filepath.Base(app.SiteDdevSettingsFile), "drushrc.php"); err != nil {
 				util.Warning("Failed to write .gitignore in %s: %v", filepath.Dir(app.SiteDdevSettingsFile), err)
 			}
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -268,7 +268,7 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 	app.SetApptypeSettingsPaths()
 
 	if app.DisableSettingsManagement && app.Type != nodeps.AppTypePHP && app.Type != nodeps.AppTypeGeneric {
-		util.Warning("Not creating CMS settings files because disable_settings_management=true")
+		util.Warning("Not creating CMS settings files because disable_settings_management=true or project type doesn't require settings")
 		return "", nil
 	}
 
--- a/pkg/ddevapp/apptypes.go
+++ b/pkg/ddevapp/apptypes.go
@@ -309,9 +309,9 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 			util.Warning("Unable to create settings file '%s': %v", app.SiteSettingsPath, err)
 		}
 
-		// Don't create gitignore if it would be in top-level directory, where
+		// Don't create settings-directory .gitignore if it would be in top-level directory, where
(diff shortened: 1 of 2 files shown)


stasadev (approved):
Thank you, I tested the Craft CMS quickstart using this PR, and now it updates `.ddev/.env.web` as expected.

rfay:
Open to suggestions about whether we should get this into this release. I don't foresee problems with doing the settings early, but perhaps I'm being naive. I didn't see any of the settings creators that tried to use a running container either.

stasadev:
To be safe, it can be merged early in the next release.

The workaround is to leave .ddev/.env.web untouched and run ddev restart.

I don't see any issues with this fix, but we can test it for a longer time in DDEV HEAD.

rfay:
Rebased

stasadev:
I fixed the problem with `.ddev/.gitignore` here.

We may want to think about how to properly gitignore `.ddev/.env.*` files (but they were not intended to be gitignored, when I added them for add-ons).

This is probably a task for:

- #7235

pkg/ddevapp/ddevapp.go
```diff
@@ -1758,10 +1762,6 @@ Fix with 'ddev config global --required-docker-compose-version="" --use-docker-c
 		return err
 	}
 
-	if _, err = app.CreateSettingsFile(); err != nil {
-		return fmt.Errorf("failed to write settings file %s: %v", app.SiteDdevSettingsFile, err)
-	}
-
```
stasadev, line 1764: I think we should have `app.CreateSettingsFile()` in both places, on pre-start and post-start.

I suggest to bring back this removed condition.

This code runs only when site is running:

https://github.com/ddev/ddev/blob/7b0cb1482c2596be0cbf67f3310883e4973992d7/pkg/ddevapp/drupal.go#L565-L582

stasadev (changes requested):

pkg/ddevapp/ddevapp.go
rfay: IMO it's pretty safe to do that, but here it was inside mutagenEnabled right?

stasadev: Yes, inside mutagenEnabled.

rfay: I added it back where it was before (but that was late in Start, not in mutagen section)

pkg/ddevapp/apptypes.go
```diff
@@ -311,7 +311,9 @@ func (app *DdevApp) CreateSettingsFile() (string, error) {
 
 		// Don't create gitignore if it would be in top-level directory, where
 		// there is almost certainly already a gitignore (like Backdrop)
-		if path.Dir(app.SiteSettingsPath) != app.AppRoot {
+		// Don't override the existing .ddev/.gitignore as well
+		// TODO: we may want to append to .ddev/.gitignore if needed.
+		if path.Dir(app.SiteSettingsPath) != app.AppRoot && path.Dir(app.SiteSettingsPath) != app.GetConfigPath("") && app.SiteDdevSettingsFile != "" {
```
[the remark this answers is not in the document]
rfay: This is great; now I understand the whole history and the stumble here.

rfay:
I made minor changes and re-added the previous near-post-start settings update. I don't understand exactly why it would be needed, but don't see it doing any harm.

stasadev (approved):
Thank you!

Adding code without deleting it seems safer to me.

Tested it again, the `.ddev/.env.web` file is updated on `ddev start` (I edited it during the start, and it was updated again)

And `.ddev/.gitignore` is not broken.

PR #7288, opened 2025-05-13
build: add `make go-mod-update` 

stasadev:
## The Issue

I don't want to remember how to bump dependencies in `go.mod`.

Adds `make go-mod-update`.

```
make go-mod-update
make
```

I don't bump the dependencies here because I don't want to overload this PR with changed lines.

And I probably won't bump it for v1.24.5, I don't see any reason for that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,6 +287,19 @@ golangci-lint:
 		echo "Skipping golangci-lint as not installed"; \
 	fi
 
+# golang doesn't always clean indirect dependencies in go.mod
+# we can remove all '// indirect' lines, and they will be re-added
+bump:
+	@echo "bump dependencies: "
+	@if [ "$$(uname)" = "Darwin" ]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -287,6 +287,19 @@ golangci-lint:
 		echo "Skipping golangci-lint as not installed"; \
 	fi
 
+# golang doesn't always clean indirect dependencies in go.mod
+# we can remove all '// indirect' lines, and they will be re-added
+bump:
+	@echo "bump dependencies: "
+	@if [ "$$(uname)" = "Darwin" ]; then \
+		sed -i '' '/\/\/ indirect$$/d' go.mod; \
```
rfay, line 295: Probably best to simplify this using ddev/ddev-utilities so it's the same command?

Or maybe awk would be easier.

stasadev:
> How about make go-mod-update or something?

Renamed.

> use a tool that behaves the same on different platforms

I used `awk`.

PR #7290, opened 2025-05-13
fix: Improve devcontainer.json to use current Ubuntu and docker, fixes #7287 [skip ci]

### LINKED ISSUE #7287: build: `make` doesn't work on Codespaces (or other Ubuntu 20.04)
### Expected Behavior

I opened a PR with Codespaces by clicking on the link and expected to be able to use ddev there. 

### Actual Behavior

* `ddev` wasn't built or in the PATH
* No default project is set up.
* Building ddev didn't work: `make` gives: `curl: option --retry-all-errors: is unknown` (The immediate problem here is that Codespaces is giving us Ubuntu 20.04 by default. Maybe we can get past that.

### Steps To Reproduce

Try "open in github codespaces" on https://github.com/ddev/ddev/pull/7286#issuecomment-2877072100

### Anything else?

I think we have a fair bit of work to do with Codespaces.

* Base the devcontainer.json on Ubuntu 24.04
* Create a default project on startup (when doing a PR build)

rfay:
## The Issue

- #7287 

devcontainer.json has an ancient Ubuntu version (20.04)

Upgrade. Use 24.04

## TODO

- [ ] The golang feature is huge and takes a lot of time to install. We don't need all of that.
- [ ] (Probably separate PR) Update [ddev install docs](https://ddev.readthedocs.io/en/stable/users/install/ddev-installation/#github-codespaces) and the related devcontainer package
- [ ] Automatically push the new ddev-devcontainer image on build with correct tag (and configure tag)

Try out this PR using the Codespaces link: https://github.com/codespaces/new?ref=20250513_rfay_devcontainer_codespaces&repo=80927419

## Automated Testing Overview

We don't know how to test these things yet but maybe we'll learn

## Release/Deployment Notes

There are a number of other things wrong with codespaces still:
- [x] No port is mapped, so the d10simple we create isn't useful until manually mapped
- [x] d10simple is out of date. And why not d11? And why Drupal specifically?

However, this may be a first step toward and architecture-independent devcontainer that will work on arm64 mac (unrelated to Codespaces, but related to 
* https://github.com/ddev/ddev/issues/6891

### DIFF
--- a/containers/ddev-devcontainer/Dockerfile
+++ b/containers/ddev-devcontainer/Dockerfile
@@ -1,0 +1,106 @@
+# syntax=docker/dockerfile:1
+# check=skip=SecretsUsedInArgOrEnv
+
+# Adapted from MS Universal Devcontainer: https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/Dockerfile
+# Original Copyright:
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM ubuntu:24.04 AS ddev-devcontainer
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    # Restore man command
+    && yes | unminimize 2>&1
+
+ENV LANG="C.UTF-8"
+
+# Install basic build tools
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        make \
+        unzip \
+        # The tools in this package are used when installing packages for Python
+        build-essential \
+        swig3.0 \
+        # Required for Microsoft SQL Server
+        unixodbc-dev \
+        # Required for PostgreSQL
+        libpq-dev \
+        # Required for mysqlclient
+        default-libmysqlclient-dev \
+        # Required for ts
+        moreutils \
+        rsync \
+        zip \
+        libgdiplus \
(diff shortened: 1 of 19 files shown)


rfay:
I doubt I'll get this into any decent shape. It may be that this can't work without MS's Oryx. I am learning a little bit about where features come from and how they are configured. 

- [x] Created a new experimental ddev-devcontainer image based on the MS universal image, but not sure if it can be used successfully without Oryx, whatever that is.
- [x] Switched the demo environment to new ddev/d11simple (useful, should be retained even if this fails)
- [x] Made setup_test_project.sh use the proper architecture (useful, retain even if this fails)

rfay:
Oh my goodness, it might be close to working! It works in Codespaces, and it even works in VS Code with Devcontainers!

rfay:
I'll close this for now, but it's on our list of key initiatives for this year, so it will get reopened.

containers/ddev-devcontainer/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+# check=skip=SecretsUsedInArgOrEnv
+
+# Adapted from MS Universal Devcontainer: https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/Dockerfile
+# Original Copyright:
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM ubuntu:24.04 AS ddev-devcontainer
```
rfay, line 11: It looks like via 
* https://github.com/ddev/ddev/pull/7713

https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/base/tag/debian-12

could be a better base and lighter.

PR #7300, opened 2025-05-15
fix: change download timeout to 5 minutes instead of 20s, fixes #7298

### LINKED ISSUE #7298: Timeout upgrading docker-compose after upgrading ddev to 1.24.5 on slower networks (fixed in v1.24.6)
### Preliminary checklist

After upgrading DDEV to 1.24.5, each project I start stops with a timeout error:

<summary>The download of docker-compose is slow and fails in time-out after a while (less than 1 minute).</summary>

<details>
After many tries (ran `ddev debug test`, restarted WSL, restarted Windows, cleared docker images, etc.) I downgraded DDEV to 1.24.4 and now docker-compose is (slowly) downloading.

It took about 30 minutes. But now it's working. The slowness is caused by WSL networking settings. I use NAT because Mirroring causes issues starting projects.
</details>

It looks like 1.24.5 has a timeout in the docker-compose download function. Can this be overridden?

### Expected Behavior

ddev start to run normally

### Actual Behavior

```log
❯ ddev start
Network ddev_default created
Starting opigno-test...
docker-compose 1.14 MiB / 70.32 MiB [=>--------------------------------------------------------------------]   1.62% 18s
Fix with 'ddev config global --required-docker-compose-version="" --use-docker-compose-from-path=false': context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

- #7298 

On slower networks, the download of docker-compose times out in just 20s and gives an odd message

- [x] Increase the timeout
- [x] Improve the message given on failure

I used network conditioning in Parallels to slow the download speed to 20Mbps to reproduce this.

Running TestDockerComposeDownload with the network slowed down shows the failure at 20s, but it succeed with this fix.

### DIFF
--- a/pkg/util/network.go
+++ b/pkg/util/network.go
@@ -43,7 +43,7 @@ func DownloadFile(destPath string, fileURL string, progressBar bool, shaSumURL s
 	}
 	client.Backoff = retryablehttp.DefaultBackoff
 	client.Logger = nil
-	client.HTTPClient.Timeout = 20 * time.Second
+	client.HTTPClient.Timeout = 5 * time.Minute
 	client.RequestLogHook = func(_ retryablehttp.Logger, req *http.Request, attempt int) {
 		if attempt > 0 {
 			// attempt==1 is the first retry, 2 the second, etc
--- a/cmd/ddev/cmd/debug-download-images.go
+++ b/cmd/ddev/cmd/debug-download-images.go
@@ -21,7 +21,7 @@ var DebugDownloadImagesCmd = &cobra.Command{
 		}
 		_, err := dockerutil.DownloadDockerComposeIfNeeded()
 		if err != nil {
-			util.Warning("Unable to download docker-compose: %v", err)
+			util.Failed("Unable to download docker-compose: %v", err)
 		}
 		if globalconfig.DdevGlobalConfig.IsMutagenEnabled() {
 			err = ddevapp.DownloadMutagenIfNeeded()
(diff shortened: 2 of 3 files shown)


rfay:
New message example:

```
Network ddev_default created 
docker-compose 38.58 MiB / 68.63 MiB [================>------------]  56.22% 19s
Failed to download updated docker-compose binary.
This might be due to network issues or a slow response. 
Please ensure your network is stable and try again:
context deadline exceeded (Client.Timeout or context cancellation while reading body) 
```

stasadev (approved):
Thank you, looks good to me!

pkg/util/network.go
```diff
@@ -43,7 +43,7 @@ func DownloadFile(destPath string, fileURL string, progressBar bool, shaSumURL s
 	}
 	client.Backoff = retryablehttp.DefaultBackoff
 	client.Logger = nil
-	client.HTTPClient.Timeout = 20 * time.Second
+	client.HTTPClient.Timeout = 5 * time.Minute
```
GuySartorelli, line 46: Is this change only for downloading images on upgrade? If not, what other scenarios use this timeout?

rfay: Has nothing to do with images, only things that use DownloadFile(), which means docker-compose, mutagen, and `ddev add-on get` downloads at this point.

rfay:
OK, going with it. Having this in HEAD may make it easier for some people to test/use it.

PR #7230, opened 2025-04-24
fix: support TYPO3 v14 nginx configuration, fixes #6944

rfay:
## The Issue

- #6944

TYPO3 docroot setup is changing.

TYPO3 v14 LTS is still a year away (April 2026), but v14 will appear in a few months.

## How This PR Solves The Issue

Use nginx config that will support older and newer TYPO3.

## Manual Testing Instructions

- [x] Test TYPO3 quickstart with v13
- [x] Test TYPO3 quickstart with TYPO3 v14 (instructions below)
- [x] New quickstart (ddev typo3 setup) at https://ddev--7230.org.readthedocs.build/en/7230/users/quickstart/#typo3
- [x] Take a look at the new quickstart tests

PR #7301, opened 2025-05-15
fix: make TYPO3 nginx config work with TYPO3 v12, fixes #7299, for #7230

### LINKED ISSUE #7299: v1.24.5: TYPO3 v12 (and previous) backend login no longer works (fallout from #7230)
### Preliminary checklist

No ddev debug test needed, see explanation 

### Expected Behavior

TYPO3 backend login page should be shown

### Actual Behavior

You see an error page, that the page is unknown.

* #7230 updated the TYPO3 nginx configuration for TYPO3 v13/v14

### Steps To Reproduce

1) Go to https://your.site/typo3
2) See Error page instead of TYPO3 login

### Anything else?

The new inserted "map" to check the index.php location always uses the root index.php (and never the one in /typo3/ directory) so this why it works with TYPO3 v13 and v14 but not with v12 (and older) anymore.

I also do not understand to use the map ... the try_files syntax would have been

> ~~try_files $uri /typo3/index.php$is_args$args /index.php$is_args$args;~~ <= Wrong, try_files do not work as that

Yes, this won't work in v13 with the configurating a different entry point like

> $GLOBALS['TYPO3_CONF_VARS']['BE']['entryPoint'] = '/admin';

but this also didn't worked in ddev out of the box before.

stasadev:
## The Issue

- #7299

Culprit:

- #7230

Uses a different nginx technique. Spent long hours with ChatGPT.

Try TYPO3 quickstart for v11, v12, v13, v14 (full install with login)

https://ddev.readthedocs.io/en/stable/users/quickstart/#typo3

v14:
```
git clone git@github.com:TYPO3/TYPO3.CMS.BaseDistribution.git my-typo3-site14
cd my-typo3-site14
ddev config --project-type=typo3 --docroot=public --php-version=8.3
ddev composer install
ddev config --auto # create TYPO3 settings after TYPO3 classes are installed in vendor
ddev exec touch public/FIRST_INSTALL
ddev launch /typo3/install.php
```

v13:
```
PROJECT_NAME=my-typo3-site13
mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
ddev config --project-type=typo3 --docroot=public --php-version=8.3
ddev start
ddev composer create-project "typo3/cms-base-distribution:^13"
ddev exec touch public/FIRST_INSTALL
ddev launch /typo3/install.php
```

v12:
```
PROJECT_NAME=my-typo3-site12
mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
ddev config --project-type=typo3 --docroot=public --php-version=8.3
ddev start
ddev composer create-project "typo3/cms-base-distribution:^12"
ddev exec touch public/FIRST_INSTALL
ddev launch /typo3/install.php
```

v11:
```
PROJECT_NAME=my-typo3-site11
mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
ddev config --project-type=typo3 --docroot=public --php-version=8.1
ddev start
ddev composer create-project "typo3/cms-base-distribution:^11"
ddev exec touch public/FIRST_INSTALL
ddev launch /typo3/install.php
```

### DIFF
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
@@ -12,6 +12,13 @@ map $http_accept $webp_suffix {
     "~*webp"  ".webp";
 }
 
+# /index.php is used for TYPO3 v14+
+# /typo3/index.php is used for TYPO3 pre-v14
+map $typo3_index_exists $typo3_index {
+    default   "/index.php";
+    1         "/typo3/index.php";
+}
+
 server {
     listen 80 default_server;
     listen 443 ssl default_server;
@@ -63,14 +70,15 @@ server {
         rewrite ^ /typo3/;
     }
 
+    # check if /typo3/index.php exists
+    set $typo3_index_exists 0;
+    if (-f $document_root/typo3/index.php) {
+        set $typo3_index_exists 1;
+    }
+
     location /typo3/ {
         absolute_redirect off;
-        # TYPO3 pre-v14 rewrite
-        if (-f $document_root/typo3/index.php) {
-            rewrite ^/typo3(/.*)?$ /typo3/index.php$is_args$args last;
-        }
-        # TYPO3 v14 rewrite
-        rewrite ^/typo3(/.*)?$ /index.php$is_args$args last;
+        try_files $uri $typo3_index$is_args$args;
     }
 
     # pass the PHP scripts to FastCGI server listening on socket
--- a/pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
+++ b/pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
@@ -12,12 +12,6 @@ map $http_accept $webp_suffix {
     "~*webp"  ".webp";
 }
 
-# Try both pre-v14 in /typo3 and also v14+ location (/index.php)
-map $document_root/typo3/index.php $typo3_fallback_target {
-    default /typo3/index.php;
-    "~*" /index.php;
-}
-
 server {
     listen 80 default_server;
     listen 443 ssl default_server;
@@ -71,7 +65,12 @@ server {
 
     location /typo3/ {
         absolute_redirect off;
-        try_files $uri $typo3_fallback_target$is_args$args;
+        # TYPO3 pre-v14 rewrite
+        if (-f $document_root/typo3/index.php) {
+            rewrite ^/typo3(/.*)?$ /typo3/index.php$is_args$args last;
+        }
+        # TYPO3 v14 rewrite
+        rewrite ^/typo3(/.*)?$ /index.php$is_args$args last;
     }
 
     # pass the PHP scripts to FastCGI server listening on socket
--- a/docs/tests/typo3.bats
+++ b/docs/tests/typo3.bats
@@ -106,46 +106,6 @@ teardown() {
   assert_success
 }
 
-# bats test_tags=typo3-setup,t3v12
-@test "TYPO3 v12 'ddev typo3 setup' composer test with $(ddev --version)" {
-  PROJNAME=my-typo3-site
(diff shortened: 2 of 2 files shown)


rfay:
@mabolek would appreciate your review here as well, thanks!

pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
```diff
@@ -71,7 +65,12 @@ server {
 
     location /typo3/ {
         absolute_redirect off;
-        try_files $uri $typo3_fallback_target$is_args$args;
+        # TYPO3 pre-v14 rewrite
+        if (-f $document_root/typo3/index.php) {
```
rfay, line 69: This looks like a reasonable approach to me.

pkg/ddevapp/webserver_config_assets/nginx-site-typo3.conf
```diff
@@ -71,7 +65,12 @@ server {
...
+            rewrite ^/typo3(/.*)?$ /typo3/index.php$is_args$args last;
+        }
+        # TYPO3 v14 rewrite
+        rewrite ^/typo3(/.*)?$ /index.php$is_args$args last;
```
rfay, line 73: Does this happen in addition to the previous rewrite?

rpkoller:
i havent seen your PR, randy only asked it would be good to have a quickstart for v12. i'Ve created one https://github.com/ddev/ddev/pull/7302  adding tabs for v13 and v12 plus bats test for each. i then ran into a failing test for the login on v12 ^^, a good proof that it works. i then downloaded the artifact from this PR and with that the tests passed. i just realized you are adding tests for v12 in this PR as well. i wonder would it make sense to drop the test from this one and use the test from the other. for one the creation for the first install file is only necessary for installs in the web ui for install using typo3 setup it is not necessary plus i did a one or two more asserts. you may take a look. or we might update the test in here? either way would be ok for me.

rpkoller:
i've also tested the artifact with v11 there i run into a 500 error on `ddev launch /typo3/install.php` (i was using php 8.1 for that older version)

stasadev:
@rpkoller, let's keep your PR open.

I thought I was logged in and probably tested it on a version where it "worked".

But testing again today on a fresh head, I get a 500 error after login on TYPO3 v13.

stasadev:
@rpkoller, I think it's better to pull my PR first and then rebase yours.

(I looked at your PR, and I think we still need a quickstart without `ddev typo3 setup`, since that's the only way to manually check if TYPO3 automatic version detection that creates `settings.php` in DDEV works properly.)

I changed the nginx conditions, now it should work better.

rfay:
I think the problem was that the login worked but nothing from there on. We can always hope to make more sophisticated quickstart tests...

rfay:
I re-ran the quickstart tests as they failed on shopware6 of all things!

stasadev:
> i wonder why it then worked for me when i tested locally?

> I think the problem was that the login worked but nothing from there on.

Correct.

> and i am not sure why a quickstart for older versions would be necessary?

@rpkoller, the reason why the previous change in #7230 worked for the current TYPO3 is because TYPO3 v13 already deprecated `/typo/index.php` and used a wrapper:

```php
// Legacy wrapper for typo3/index.php
// @deprecated will be removed in TYPO3 v14, /index.php entrypoint should be used directly
```

I think it would not make any difference if I just replaced `/typo3/index.php` with `/index.php` at that time.

In case of TYPO3 v12, it uses a different approach in `/typo3/index.php`, so it doesn't work with DDEV v1.24.5.

And that's why we need to test TYPO3 v12.

stasadev:
I'm removing the TYPO3 v12 test from this PR. It'll be changed anyway in #7302, and I want to avoid PR conflicts.

rfay (approved):
Yay!

I tested in v11, v12, v13, v14, and also tested with alternate BE path /admin for v14. All worked great.

PR #7302, opened 2025-05-16
test: adding quickstarts for typo3 v13 and v12 plus bats tests

rpkoller:
## The Issue
the test for v12 will fail in current HEAD, it requires https://github.com/ddev/ddev/pull/7301

- #<issue number>

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -1088,16 +1088,56 @@ DDEV automatically updates or creates the `.env.local` file with the database in
 
 ## TYPO3
 
-=== "Composer"
+=== "TYPO3 v13"
 
     ```bash
     PROJECT_NAME=my-typo3-site
     mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
-    ddev config --project-type=typo3 --docroot=public --php-version=8.3
+    ddev config --project-type=typo3 --docroot=public --php-version 8.3
     ddev start
-    ddev composer create-project "typo3/cms-base-distribution"
-    ddev exec touch public/FIRST_INSTALL
-    ddev launch /typo3/install.php
+    ddev composer create-project "typo3/cms-base-distribution:^13"
+    ddev typo3 setup \
+        --admin-user-password="Demo123*" \
+        --driver=mysqli \
+        --create-site=https://${PROJECT_NAME}.ddev.site \
+        --server-type=other \
+        --dbname=db \
+        --username=db \
+        --password=db \
+        --port=3306 \
+        --host=db \
+        --admin-username=admin \
+        --admin-email=admin@example.com \
+        --project-name="My TYPO3 site" \
+        --force
+
+    ddev launch /typo3/
+    ```
+
+=== "TYPO3 v12"
+
+    ```bash
+    PROJECT_NAME=my-typo3-site
+    mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
+    ddev config --project-type=typo3 --docroot=public --php-version 8.1
+    ddev start
+    ddev composer create-project "typo3/cms-base-distribution:^12"
+    ddev typo3 setup \
+        --admin-user-password="Demo123*" \
+        --driver=mysqli \
+        --create-site=https://${PROJECT_NAME}.ddev.site \
+        --server-type=other \
+        --dbname=db \
+        --username=db \
+        --password=db \
+        --port=3306 \
+        --host=db \
+        --admin-username=admin \
+        --admin-email=admin@example.com \
+        --project-name="My TYPO3 site" \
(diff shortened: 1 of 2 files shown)


docs/content/users/quickstart.md
```diff
@@ -1088,16 +1088,56 @@ DDEV automatically updates or creates the `.env.local` file with the database in
...
+        --host=db \
+        --admin-username=admin \
+        --admin-email=admin@example.com \
+        --project-name="My TYPO3 site" \
+        --force
+
+    ddev launch /typo3/
+    ```
+
+=== "TYPO3 v12"
```
rfay, line 1117: I guess I don't think we need to show quickstart for this, although we absolutely need the test.

rfay (changes requested):
Thanks for the great work on this!

I don't think we need the quickstart for TYPO3 v12, but we do need the test.

Would you mind adding test for v11 as well? (but not quickstart)

docs/content/users/quickstart.md
rpkoller: agreed, the difference between the quickstart for 12 and 13 is marginal (more or less one number) ;) 

i'll update the PR later tonight. have to finish some drupal stuff first. and i also agree that having a test for v11 would be good. in particular cuz `typo3 setup`got in with v12, so with 11 you have the older setups.

rpkoller: i guess when i work on the update i will use the artifact from the PR stas is working on. would be a good way to test that everything is working as expected either way.

rpkoller:
the installer in v11 is odd. that is the output in the browser
Screenshot 2025-05-17 at 02 26 07

I tried to assert for example for `First the Install Wizard needs to inspect your web server and identify any issues that may prevent TYPO3 from running correctly.` but no match, but in the devtools it is there. but with curl it is not :D

```
$> curl -sfL https://v11.ddev.site/typo3/install.php
<!DOCTYPE html>
<html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers">
<head>
    <title>Installing TYPO3 CMS</title>
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
    <link rel="stylesheet" href="/typo3/sysext/backend/Resources/Public/Css/backend.css?88f713284cb7c03aa64597970226257242e44839" />
    <link rel="icon" type="image/vnd.microsoft.icon" href="/typo3/sysext/install/Resources/Public/Icons/favicon.ico?88f713284cb7c03aa64597970226257242e44839" />
</head>
<body class="backend t3js-body install-tool-installer typo3-install" data-package-resource-paths="{&quot;backend&quot;:&quot;\/typo3\/sysext\/backend\/Resources\/Public\/&quot;,&quot;core&quot;:&quot;\/typo3\/sysext\/core\/Resources\/Public\/&quot;,&quot;install&quot;:&quot;\/typo3\/sysext\/install\/Resources\/Public\/&quot;}"></body>
<script src="/typo3/sysext/install/Resources/Public/JavaScript/RequireJSConfig.js?88f713284cb7c03aa64597970226257242e44839" data-bust="88f713284cb7c03aa64597970226257242e44839" data-init="TYPO3/CMS/Install/Installer"></script>
<script src="/typo3/sysext/core/Resources/Public/JavaScript/Contrib/require.js?88f713284cb7c03aa64597970226257242e44839"></script>
</html>
```

odd i simply asserted for something that was put out with curl. now all tests are passing using the artifact from the PR from stas. i'll push the commit containing the updates for the tests in a second

rpkoller:
i've rebased the PR after https://github.com/ddev/ddev/issues/7299 being merged, the quickstart test should hopefully pass now.

docs/content/users/quickstart.md
```diff
@@ -1093,11 +1093,25 @@ DDEV automatically updates or creates the `.env.local` file with the database in
     ```bash
     PROJECT_NAME=my-typo3-site
     mkdir ${PROJECT_NAME} && cd ${PROJECT_NAME}
-    ddev config --project-type=typo3 --docroot=public --php-version=8.3
+    ddev config --project-type=typo3 --docroot=public --php-version 8.3
```
stasadev, line 1096: Not sure why this was changed, but we use equals everywhere:
```suggestion
    ddev config --project-type=typo3 --docroot=public --php-version=8.3
```

rpkoller: argh an oversight by me, i was blind, and i know the reason why that slipped in, i simply copied over the version from the quickstart on typo3.org. ( i tried to align the minimum setup for each version).

stasadev (approved):
Looks good to me!

PR #7311, opened 2025-05-19
build: bump docker-compose to v2.36.1

stasadev:
## The Issue

`docker-compose` [v2.36.1](https://github.com/docker/compose/releases/tag/v2.36.1) is out.

Bumps it.


rpkoller:
but wouldnt be another new one also a validation that things are back to normal download wise with the new release, when they see that everything is working out now again?

stasadev:
I think adding this PR is good. Visible fix right away.

PR #7326, opened 2025-05-22
feat: prune orphaned Node.js versions after install, fixes #7325

### LINKED ISSUE #7325: Prune orphaned node versions when node is updated

E.g. with `node_version: "18"`, each time a new version is released it’s added to:

`ddev-global-cache/n_prefix/sitename/n/versions/node/`

However, old versions aren’t removed, so the volume ends up increasing in size over time.

Image

### Describe your solution

I have no experience of DDEV internals, but a quick code search suggests it uses https://github.com/tj/n which does have a `prune` command that could be run in `n-install.sh`.

### Describe alternatives

_No response_

### Additional context

_No response_

lozcalver:
## The Issue

- #7325

Runs the `n prune` command

Open the `ddev-global-cache` volume and look at a long-standing project’s directory: `n_prefix/<project-name>/n/versions/node/`. It’ll likely already contain multiple Node.js versions.

After this change, the next time you start DDEV only the most recent Node.js version should be present

## Automated Testing Overview

I’m not really sure how you’d test this, but given that this is a non-essential enhancement (everything’s running fine without this change, it just saves some disk space) I don’t think additional testing is necessary

## Release/Deployment Notes

N/A


containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/n-install.sh
```diff
@@ -47,6 +47,9 @@ if [ "${n_install_result}" = "false" ]; then
   rm -f "${N_PREFIX}" && exit 6
 fi
 
+# prune orphaned Node.js versions
+log-stderr.sh n prune || true
```
lozcalver, line 51: I wasn’t sure if including `log-stderr.sh` here was worth it given that this command isn’t really essential. Happy to remove it.

stasadev: It's fine to use `log-stderr.sh` here.

stasadev (commented):
Thank you!

I'm pushing a new Docker image now.

Important: test starting/stopping the project, then disable Internet and start the project again. It should work as before.

stasadev:
A new Docker image is ready, build artifacts will appear shortly below.

lozcalver:
Hmm, it’s not working properly - it removes _all_ the versions. Node still “works” but when restarting without an internet connection it fails over to Node 22 (presumably whatever is installed with the image by default).

I think this may be because the prune command needs to be moved after the symlinks are added here:

https://github.com/ddev/ddev/blob/2842925c37731d0b623a8148fccb8952c44c6909/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/n-install.sh#L50-L57

`n`’s prune command will detect the current “active” node version before it runs, and it does so like this:

https://github.com/tj/n/blob/f52d2172f12cd76f0efe9524690723f52ab74f40/bin/n#L477-L494

I’ve pushed a commit that moves the command. If that works I can squash the commits + force push if you’d like, not sure if you’d prefer that done upfront or if you’ll squash + merge

stasadev:
> I’ve pushed a commit that moves the command. If that works I can squash the commits + force push if you’d like, not sure if you’d prefer that done upfront or if you’ll squash + merge

We always squash on merge, no need to do it here.

Pushing a new image.

stasadev:
The new Docker image is [ready](https://github.com/stasadev/ddev/actions/workflows/push-tagged-image.yml).

Before testing run:
```
ddev poweroff
docker pull ddev/ddev-webserver:20250522_lozcalver_n_prune
```

lozcalver:
Great, it’s working now. With a Node 18 project, only `18.20.8` remains after starting DDEV. Tested disabling my internet connection before starting, that works too:

Screenshot 2025-05-22 at 11 40 01

stasadev:
Great!

I'll test it later today as well.

(Ignore the "Container tests" failure: it's unrelated to this PR.)

stasadev (approved):
Thank you!

There might be cases where someone wants to install multiple Node versions, but they can work around it by using a different `N_PREFIX`.

The logic behaves as expected:

```
$ ddev config --nodejs-version="14.1.0"
$ ddev start

$ ddev exec node -v
v14.1.0

$ ddev exec "n install 14 && n install 16 && n install 18"

$ ddev exec n ls
node/14.1.0
node/14.21.3
node/16.20.2
node/18.20.8

$ ddev config --nodejs-version="14"
$ ddev restart

$ ddev exec node -v
v14.21.3

# only one version as expected
$ ddev exec n ls
node/14.21.3

$ ddev exec "n install 14.21.2 && n install 14.21.1 && n install 14.21.0"

$ ddev exec n ls
node/14.21.0
node/14.21.1
node/14.21.2
node/14.21.3

$ ddev poweroff

# turn off the Internet

$ ddev start

$ ddev exec node -v
v14.21.3

# only one version as expected
$ ddev exec n ls
node/14.21.3

$ ddev restart

$ ddev exec node -v
v14.21.3

# only one version as expected
$ ddev exec n ls
node/14.21.3
```

PR #7336, opened 2025-05-27
fix: make `ddev composer create-project` work with stdin

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/7323#pullrequestreview-2861494051

stdin doesn't seem to work properly with `ddev composer create-project`.

I.e. `echo x | ddev composer create-project shopware/production` doesn't pass `x`, Composer thinks there is no TTY, and doesn't ask any question.

And there is no error on failed `composer run-script`.

We want to have better test coverage by using stdin in our quickstart tests.

I investaged the problem and found out that:
1. `os.Stdin` is empty after it's passed to the first `app.Exec()` call, nothing there afterwards.
2. Composer doesn't recognize TTY at all when using pipes.

So I used `os.Pipe` to pass `os.Stdin` for each `app.Exec()` call and wrapped Composer commands in preudo-TTY with `script`:
- `composer run-script post-root-package-install` can have interaction
- `composer install` can have interaction in `post-install-cmd` (can be tested with https://ddev.readthedocs.io/en/stable/users/quickstart/#cakephp without `--no-interaction` flag)
- `composer run-script post-create-project-cmd` can have interaction

> [!NOTE]  
> Stdin is passed to each of the commands above. The *entire* input is forwarded, so if multiple steps include interactive questions, it may not behave as expected: there's no way to tell which input was consumed and which wasn't.
>
> Example: `composer run-script post-root-package-install` asks "What is your first name?" and `composer run-script post-create-project-cmd` asks "What is your last name?", If you run `printf "Stas\nZhuk\n" | ddev composer creat-project ...`, then `"Stas\nZhuk\n"` is passed to both scripts, which means `Stas` will be consumed in both cases.
>
> Considering that usually you only need to pass `Y`/`yes` to agree to something, or that most interaction happens only in `composer run-script post-create-project-cmd`, I think it an acceptable compromise, especially since it didn't work at all before.

And I added an error check for `composer run-script`.

Craft CMS quickstart using pipe should work now:

```
printf "admin\nadmin@example.com\nnewpassword\nnewpassword\ntestsite\n\n\n" | ddev composer create-project craftcms/craft
ddev launch # login with `admin@example.com` and `newpassword`
```

In addition, check Shopware quickstart https://ddev.readthedocs.io/en/stable/users/quickstart/#shopware with:
```
echo x | ddev composer create-project shopware/production
```
And Cake PHP quickstart https://ddev.readthedocs.io/en/stable/users/quickstart/#cakephp with:
```
echo n | ddev composer create-project --prefer-dist cakephp/app:~5.0
```

## Automated Testing Overview

Updates Shopware quickstart test by adding a real check to confirm the selected option was executed correctly.
This test fails in v1.24.6, but not in this PR.

## Release/Deployment Notes

It doesn't touch the existing interactive `ddev composer create-project` logic, so I consider this PR safe to pull.

### DIFF
--- a/cmd/ddev/cmd/composer-create-project.go
+++ b/cmd/ddev/cmd/composer-create-project.go
@@ -547,6 +547,10 @@ func wrapTTYCommandWithStdin(data []byte, cmd []string) []string {
 		_, _ = stdinWriter.Write(data)
 	}()
 	os.Stdin = stdinReader
+	// 'script' forces execution in a pseudo-terminal (PTY)
+	// '-q' suppresses script's start and end messages (quiet mode)
+	// '-c' specifies the command to run in the PTY
+	// '/dev/null' discards the session log
 	return []string{"script", "-q", "-c", strings.Join(cmd, " "), "/dev/null"}
 }
 
--- a/cmd/ddev/cmd/composer-create-project.go
+++ b/cmd/ddev/cmd/composer-create-project.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"path/filepath"
@@ -38,6 +39,8 @@ ddev composer create-project --prefer-dist --no-interaction --no-dev psr/log .
 `,
 	ValidArgsFunction: getComposerCompletionFunc(true),
 	Run: func(cmd *cobra.Command, args []string) {
+		inputData := readPipedInput()
+
 		app, err := ddevapp.GetActiveApp("")
 		if err != nil {
 			util.Failed(err.Error())
@@ -186,16 +189,21 @@ ddev composer create-project --prefer-dist --no-interaction --no-dev psr/log .
 			}
 
 			composerCmd = append(composerCmd, validRunScriptArgs...)
+			composerCmd = wrapTTYCommandWithStdin(inputData, composerCmd)
 
 			output.UserOut.Printf("Executing Composer command: %v\n", composerCmd)
 
-			stdout, stderr, _ = app.Exec(&ddevapp.ExecOpts{
+			stdout, stderr, err = app.Exec(&ddevapp.ExecOpts{
 				Service: "web",
 				Dir:     getComposerRootInContainer(app),
 				RawCmd:  composerCmd,
 				Tty:     isatty.IsTerminal(os.Stdin.Fd()),
 				Env:     []string{"XDEBUG_MODE=off"},
 			})
+
+			if err != nil {
+				util.Failed("Failed to run post-root-package-install: %v\nstderr=%v", err, stderr)
+			}
 
 			if len(stdout) > 0 {
 				output.UserOut.Println(stdout)
@@ -227,6 +235,7 @@ ddev composer create-project --prefer-dist --no-interaction --no-dev psr/log .
 				}
 			}
 
+			composerCmd = wrapTTYCommandWithStdin(inputData, composerCmd)
 			// Run install command.
 			output.UserOut.Printf("Executing Composer command: %v\n", composerCmd)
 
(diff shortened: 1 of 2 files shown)


rpkoller:
hmmmmm as i'Ve commented on the other issue the approach for filling out the prompts in craft cms works like a charm. but that example uses printf, for the other two with shopware and cakephp i am uncertain. both install properly, buuut i am not sure if the prompt was "pressed"/submitted correctly. cuz if i you take a look at the output. as you can see the echo'ed "n" is preinted way before the prompt:  
Screenshot 2025-05-28 at 23 08 43

for shopware i first didnt know what was entered at the prompt (As i said the install ran through)

Screenshot 2025-05-28 at 23 06 51
but scrolling up way up showed me when the "x" was entered. 
Screenshot 2025-05-28 at 23 09 18
on macos with zsh

stasadev:
> i am not sure if the prompt was "pressed"/submitted correctly.

You can compare the final result:

```
# `compose.yaml` and `compose.override.yaml` files created in the project root
$ echo y | ddev composer create-project shopware/production
```

```
# there are no `compose.yaml` and `compose.override.yaml` files in the project root
$ echo x | ddev composer create-project shopware/production
```

```
$ echo Y | ddev composer create-project --prefer-dist cakephp/app:~5.0

# check the output for:
Permissions set on /var/www/html/tmp/cache
Permissions set on /var/www/html/tmp/cache/models
Permissions set on /var/www/html/tmp/cache/persistent
Permissions set on /var/www/html/tmp/cache/views
Permissions set on /var/www/html/tmp/sessions
Permissions set on /var/www/html/tmp/tests
Permissions set on /var/www/html/tmp
Permissions set on /var/www/html/logs
```

```
# no "Permissions set on" messages
$ echo n | ddev composer create-project --prefer-dist cakephp/app:~5.0
```

---

About the positioning of `n`/`x` and any other inputs:

```
$ echo n | ddev composer create-project --prefer-dist cakephp/app:~5.0
...
n
> App\Console\Installer::postInstall
Created `config/app_local.php` file
Created `/var/www/html/logs` directory
Created `/var/www/html/tmp/cache/views` directory
Set Folder Permissions ? (Default to Y) [Y,n]? Updated Security.salt value in config/app_local.php
```

Of course it would be more readable like this:

```
$ echo n | ddev composer create-project --prefer-dist cakephp/app:~5.0
...
> App\Console\Installer::postInstall
Created `config/app_local.php` file
Created `/var/www/html/logs` directory
Created `/var/www/html/tmp/cache/views` directory
Set Folder Permissions ? (Default to Y) [Y,n]? n
Updated Security.salt value in config/app_local.php
```

The Composer commands run inside a pseudo-terminal using the `script` command (PTY). It's a wrapper. That's why all input appears before the prompt and the output appears after, no matter what shell is used.

You can check it yourself by creating a CakePHP project, go inside the container with `ddev ssh`, and:

1. Without using the `script` wrapper, input `n`, it doesn't ask the question `Set Folder Permissions ?`, and sets file permissions, but said no:
	```
    $ rm -rf tmp logs
	$ echo n | composer run-script post-create-project-cmd> App\Console\Installer::postInstall
	Created `/var/www/html/logs` directory
	Created `/var/www/html/tmp` directory
	Created `/var/www/html/tmp/cache` directory
	Created `/var/www/html/tmp/cache/models` directory
	Created `/var/www/html/tmp/cache/persistent` directory
	Created `/var/www/html/tmp/cache/views` directory
	Created `/var/www/html/tmp/sessions` directory
	Created `/var/www/html/tmp/tests` directory
	Permissions set on /var/www/html/tmp/cache
	Permissions set on /var/www/html/tmp/cache/models
	Permissions set on /var/www/html/tmp/cache/persistent
	Permissions set on /var/www/html/tmp/cache/views
	Permissions set on /var/www/html/tmp/sessions
	Permissions set on /var/www/html/tmp/tests
	Permissions set on /var/www/html/tmp
	Permissions set on /var/www/html/logs
	No Security.salt placeholder to replace.
	```

2. And check out the same thing inside a wrapper (No permissions were set as expected):
	```
    $ rm -rf tmp logs
	$ echo n | script -q -c "composer run-script post-create-project-cmd" /dev/null
	n
	> App\Console\Installer::postInstall
	Created `/var/www/html/logs` directory
	Created `/var/www/html/tmp` directory
	Created `/var/www/html/tmp/cache` directory
	Created `/var/www/html/tmp/cache/models` directory
	Created `/var/www/html/tmp/cache/persistent` directory
	Created `/var/www/html/tmp/cache/views` directory
	Created `/var/www/html/tmp/sessions` directory
	Created `/var/www/html/tmp/tests` directory
	Set Folder Permissions ? (Default to Y) [Y,n]? No Security.salt placeholder to replace.
	```

This isn't a Docker or stdin issue. The problem is with Composer itself: it doesn't handle interactive input correctly when passed with `echo Y |`, which is why wrapping it inside `script` is a necessary workaround.

PR #7349, opened 2025-06-03
refactor: suppress `docroot` warning when `.htaccess` file is present, fixes #7345

### LINKED ISSUE #7345: Ability to disable/ modify `ddev start` root index.* check

Changes present in https://github.com/ddev/ddev/pull/7228 now on every DDEV start output the following warning for our projects.

```
The index.php or index.html does not yet exist in ...
```

If it matters (project type is `Silverstripe`)

This is an unneeded warning for us though, as we make use of `.htaccess` and rewrite requests, this aligns with how the Silverstripe Cloud platform operates as well.

It would be great to not have this warning.

Specific code:
https://github.com/ddev/ddev/pull/7228/files#diff-de9be5de8b21f6465a7e0df5c9c2f44638c0a67659ccf812cd043bfee94aeeb1R1787

### Describe your solution

Ability via the `.ddev/config.yaml` to bypass this check.
`root_index_check: false` or something

### Describe alternatives

Perhaps an "extension" point to allow project types to override this before the final check.
I'd assume various types of things would possibly have some subtle differences, and this would allow some flexibility there.

### Additional context

_No response_

stasadev:
## The Issue

- #7345

Doesn't show warning for apache2 if there is `.htaccess` file.

Example file is https://github.com/silverstripeltd/demo-project/blob/master/.htaccess

I decided not to check the contents of the `.htaccess` file because the main warning goal is already met.
People who don't understand `docroot` might set it to a random folder that doesn't have the right files.

```
$ mkdir docroot-htaccess && cd docroot-htaccess
$ ddev config --webserver-type=nginx-fpm
$ touch .htaccess
$ ddev start

Starting docroot-htaccess... 
The index.php or index.html does not yet exist at this path:
/home/stas/tmp/docroot-htaccess/index.*
You may get 403 errors 'permission denied' from the browser until it does.
If the webserver rewrites to a different docroot, this warning may not apply.
Ignore if a later action (like 'ddev composer create-project') will create it. 

$ ddev config --webserver-type=apache-fpm
$ ddev start

Starting docroot-htaccess...
(no warning)
...
```

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ddev/ddev/pkg/exec"
 	"github.com/ddev/ddev/pkg/fileutil"
 	"github.com/ddev/ddev/pkg/globalconfig"
+	"github.com/ddev/ddev/pkg/heredoc"
 	"github.com/ddev/ddev/pkg/netutil"
 	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/output"
@@ -1802,7 +1803,10 @@ func warnMissingDocroot(app *DdevApp) {
 	}
 	docroot := app.GetAbsDocroot(false)
 	if !fileutil.FileExists(docroot) {
-		util.WarningWithColor("magenta", "The project docroot does not yet exist or is misconfigured at this path:\n%s\nYou may get 403 errors 'permission denied' from the browser until it does.\n", docroot)
+		util.WarningWithColor("magenta", heredoc.Doc(fmt.Sprintf(
+			`The project docroot does not yet exist or is misconfigured at this path:
+			%s
+			You may get 403 errors 'permission denied' from the browser until it does.`, docroot)))
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


n8-dev:
I don't have time to check this right now, but this looks pretty good, thank you for trying a solution 😄

tyler36 (approved):
Works as described; warning does _not_ appear when `.htaccess` file is present

## Test DDEV `v1.24.6`

```shell
$ ddev -v
ddev version v1.24.6
$ ddev config --webserver-type=apache-fpm
$ touch .htaccess
$ ddev start
...
The index.php or index.html does not yet exist in
/home/user47/tmp/docroot-htaccess/index.*
so you may get 403 errors 'permission denied' from the browser until it does.
Don't worry about this if a later action 
(like `ddev composer install`) will create the file.
...
Successfully started docroot-htaccess 
```

## Test PR

```shell
$ ddev -v
ddev version v1.24.6-12-gcffa66a35
$ ddev start
...
(no warning)
Successfully started docroot-htaccess 
```

rfay:
Could you please add to the OP to explain why we should skip the warning if a .htaccess is in the docroot?

n8-dev:
Sorry @rfay do you mean on this PR? Or this issue? https://github.com/ddev/ddev/issues/7345

rfay:
@n8-dev I meant on this PR.

rfay (changes requested):
While we absolutely respect the Silverstripe community's request for this general feature, I'm not convinced this is the right solution. What's going on with silverstripe seems to be a very unusual server-side configuration, and I don't think it makes sense to have DDEV jump through hoops to solve a problem like that. 

I also recognize that this is a pretty simple PR, but it's still likely to have unexpected consequences.

I guess as suggested in #7345 the way to solve it would be adding configuration to say "don't warn". But even that seems like jumping through modest hoops to solve a complexity introduced by one CMS and not even in the context of DDEV.

stasadev:
Closing to move the conversation back to the issue.

PR #7363, opened 2025-06-09
fix: Allow Docker Desktop for Linux (DDFL) but warn about it, fixes #7307

### LINKED ISSUE #7307: Consider calming Docker Desktop for Linux warning instead of saying "not supported"
### Preliminary checklist

- [] I have run `ddev debug test` to include output below

### Expected Behavior

DDEV should allow running projects with Docker Desktop on Linux if the environment is otherwise functional. Earlier versions like v1.23.0 worked without issue. Docker Desktop should not be blocked unless there is a specific technical limitation.

### Actual Behavior

After upgrading to DDEV v1.24.5, DDEV refuses to start and shows the following error:

```
Problem with your Docker provider: docker desktop on Linux is not yet compatible with DDEV.

Fix with 'ddev config global --required-docker-compose-version="" --use-docker-compose-from-path=false':
context deadline exceeded (Client.Timeout or context cancellation while reading body)

```

Even after running the recommended fix (ddev config global), the error persists. Projects that previously worked now fail to start.

### Steps To Reproduce

Use Ubuntu 22.04 with Docker Desktop for Linux installed.

Install or upgrade to DDEV v1.24.5.

Verify Docker Desktop is functional:

`docker info | grep 'Docker Root Dir'
`
Output:
`Docker Root Dir: /var/lib/docker-desktop`

Run:
`ddev start`

DDEV fails with Docker Desktop compatibility error and Docker Compose version timeout.

### Anything else?

Docker info:

```
Server Version: 27.5.1
Docker Root Dir: /var/lib/docker
```

Docker Compose version:
`Docker Compose version v2.27.0`

 I resolved this issue downgrading DDEV to v1.23.0 via:

`sudo apt-get install ddev=1.23.0`

This appears to be a regression in v1.24.x — please consider either relaxing the block or allowing override in cases where Docker Desktop is functional. Let me know if logs or further environment info are needed.

rfay:
## The Issue

- #7307 

We hadn't been supporting Docker Desktop for Linux due to mounting problems, but those seem to be resolved.

Calm the language in the docs and in the warning that DDEV gives.

With DDFL, do `ddev start`
Read docs.

## Automated Testing Overview

I had hopes of being able to add automated testing with Github actions, but you can't currently accept the DDFL license agreement from command line, only from GUI.

### DIFF
--- a/pkg/ddevapp/hooks_test.go
+++ b/pkg/ddevapp/hooks_test.go
@@ -82,7 +82,7 @@ func TestProcessHooks(t *testing.T) {
 		require.NoError(t, err)
 
 		captureOutputFunc, err := util.CaptureOutputToFile()
-		require.NoError(t, err, `failed to capture output to file for taxk='%v' err=%v`, task, err)
+		require.NoError(t, err, `failed to capture output to file for task='%v' err=%v`, task, err)
 		userOutFunc := util.CaptureUserOut()
 
 		err = app.Start()


rfay:
Manual run test failures:
* TestTimezoneConfig
* ~~TestConfigFunctionality~~ (due to missing mysql client on host)
* ~~TestProcessHooks~~ works with DDEV_DEBUG=true
* ~~TestGitPull~~

[make.out.txt](https://github.com/user-attachments/files/20663906/make.out.txt)

stasadev (approved):
I couldn't test it properly on Arch Linux due to port conflicts, I don't even want to investigate it and will just rollback my system:
```
...
Error response from daemon: ports are not available: exposing port TCP 127.0.0.1:80 -> 127.0.0.1:0: listen tcp4 127.0.0.1:80: bind: permission denied'
...
Port 8025 is busy, using 33000 instead, see https://ddev.com/s/port-conflict
...
unable to listen on required ports, port 8142 is already in use,
Troubleshooting suggestions at https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/#unable-listen
...
```

What I noticed is that I see this warning three times during `ddev start`:
```
Docker Desktop for Linux is not explicitly supported and may cause problems.
```

Not a big deal, but I wonder if it might make sense to add some kind of global array to the cache messages and have something like `util.WarningOnce()`, which also can be used for other repeating warnings (i.e. we can see several same disable-settings-management warnings).

rfay:
Hmm, I was worried about that kind of thing, but didn't notice it. It's because I used `util.Warning` whereas previously it had returned an error.

rfay:
It wasn't right to have the DDFL check in CheckDockerVersion() in the first place. Created CheckDockerProvider() for this purpose, and call it at startup.

cmd/ddev/cmd/debug-dockercheck.go
```diff
@@ -66,7 +66,10 @@ var DebugDockercheckCmd = &cobra.Command{
 				util.Warning("Problem with your Docker provider: %v.", err)
 			}
 		}
-
+		err = dockerutil.CheckDockerProvider()
+		if err != nil {
+			util.Warning("Problem with your Docker provider: %v.", err)
```
stasadev: To match the message in the `root.go`:

```suggestion
			util.Warning("Possible problem with your Docker provider: %v.", err)
```

stasadev (approved):
Excellent, tested it, the warning is displayed only once.

rfay:
I had odd results trying to manually test TestTimezoneConfig with DDFL, and wasn't able to sort them out. 

Ubuntu 24.04, DDFL, TestTimezoneConfig fails because with https://github.com/ddev/ddev/blob/ff5cffdb474723256eab50e711987d7621b8b0a5/pkg/ddevapp/config_test.go#L1190 it gets a HostAbbrev of UTC. 

Succeeds if I remove that line.

But...
* Test doesn't  fail on same system using docker-ce.
* Running a trivial go file that looks up host abbrev works (seems to only fail with `t.SetEnv("TZ", "")`

It's not that important to solve, but it sure is confusing. 

```
    config_test.go:1208: 
            Error Trace:    /home/rfay/workspace/ddev/pkg/ddevapp/config_test.go:1208
            Error:          Not equal: 
                            expected: "timezone=UTC\nphptz=America/Denver"
                            actual  : "timezone=MDT\nphptz=America/Denver"
                            
                            Diff:
                            --- Expected
                            +++ Actual
                            @@ -1,2 +1,2 @@
                            -timezone=UTC
                            +timezone=MDT
                             phptz=America/Denver
            Test:           TestTimezoneConfig
```

I have absolutely no idea why DDFL would be related to this, as it's a host-side lookup using go.

So somehow hostTimezoneAbbrev is UTC, just `hostTimezoneAbbr, _ = time.Now().In(time.Local).Zone()`

rfay:
Bizarre. It works fine with DDEV_DEBUG=true.

PR #7365, opened 2025-06-10
feat: add `DDEV_USERNAME` variable which is used for the web container user

stasadev:
## The Issue

With DDEV add-ons, people may want to have the same features as we have in the `web` or `db` containers, such as having the same username inside the container.

For example, I hardcoded `ddev` as the username here:

https://github.com/stasadev/ddev-frankenphp/blob/2ae854f67847afb642037efccbceb3c77fd96e7d/docker-compose.frankenphp.yaml#L11-L17

```dockerfile
ARG FRANKENPHP_USER=ddev
ARG DDEV_UID
ARG DDEV_GID
RUN <<EOF
  set -eu
  groupadd -g $${DDEV_GID} $${FRANKENPHP_USER}
  useradd -u $${DDEV_UID} -g $${DDEV_GID} $${FRANKENPHP_USER}
  ...
EOF
```

But I would like to have the same username as on the host.

We already have `DDEV_UID`, `DDEV_GID` available, why not have the same thing for the username.

Adds `DDEV_USERNAME` variable.

```
$ cat <<'EOF' > .ddev/docker-compose.username.yaml
services:
  web:
    environment:
      - DDEV_USERNAME=$DDEV_USERNAME
EOF

$ ddev start

$ ddev exec 'echo $DDEV_USERNAME'
stas
```


stasadev:
This works just like `DDEV_UID` and `DDEV_GID`: it's not needed inside the containers, only for build purposes.

What I should do is replace `{{ .Username }}` with `$DDEV_USER` in `app_compose_template.yaml`.

rfay:
> What I should do is replace {{ .Username }} with $DDEV_USER in app_compose_template.yaml.

I probably prefer go templating over environment variables in there. 

I guess since you're only aiming this environment variable at host-side usage it's fine. But I don't really know why it shouldn't be other places (like $USER is)

stasadev:
About the `USER` variable - I think it's special in Linux world, that's why it's passed inside.

rfay:
`$USER` is special, but we do in fact override it due to a variety of problems over the years with weird usernames. DDEV_USER could have those problems as well, but I think it's set correctly using all the complexities that have been discovered.

rfay:
Rebased to get rid of the quickstart fail.

stasadev:
Updated OP and test which was not in sync with our docs.

docs/content/users/extend/custom-commands.md
```diff
@@ -115,18 +116,22 @@ A number of environment variables are provided to these command scripts. These a
 * `DDEV_PROJECT_TYPE`: `backdrop`, `drupal`, `typo3`,`wordpress`, etc.
 * `DDEV_ROUTER_HTTP_PORT`: Router port for HTTP
 * `DDEV_ROUTER_HTTPS_PORT`: Router port for HTTPS
-* `DDEV_SCHEME`: Schema of primary project URL
+* `DDEV_SCHEME`: Scheme of primary project URL
```
rfay, line 119: Nice work on looking carefully at these things.

PR #7368, opened 2025-06-10
test: update check for starter page in Silverstripe CMS

stasadev:
## The Issue

The quickstart test for Silverstripe CMS is failing because of the new template in today's 6.0.0 release.
https://www.silverstripe.org/blog/silverstripe-cms-6-0/

Fixes it.


rfay:
I was wondering why that failed!

PR #7370, opened 2025-06-10
build: Update google/go-github to v72, remove obsolete dependencies, replaces #7369

rfay:
## The Issue

dependabot opened 
* #7369 to update cloudflare/circl

But I couldn't figure out why we needed it. It was an indirect dependency of go-github. But `go mod why` didn't show a reason. 

Update to v72 of go-github, which does not use cloudflare/circl at all

Test things that rely on go-github, mostly `ddev add-on get` and `ddev add-on list`

## Automated Testing Overview

No changes


stasadev (approved):
Looks good to me, `ddev add-on list` and `ddev add-on get` work well.

rfay:
Don't understand why this would break TestCmdDescribe only on one platform. https://buildkite.com/ddev/ddev-macos-arm64-mutagen/builds/10708#01975ee2-db7c-4e23-bca0-2f41bf879202/20639

stasadev:
Maybe it's something with Docker Desktop 4.42.0 on macOS arm64.

`ddev describe` output should contain all ports from `expose` and `ports`, but apparently it only contains ports from `ports`:

https://github.com/ddev/ddev/blob/2d90a7e236536b6fa9ae4fe06d61d1ed321853e7/cmd/ddev/cmd/testdata/TestCmdDescribe/docker-compose.override.yaml#L1-L10

rfay:
I'll try to take a look locally after sorting out some other things and switching back to macOS.

rfay:
This does seem to be a Docker Desktop problem. (Probably amd64 and arm64, we have only tested on arm64 I think)

* TestCmdDescribe succeeds with this code and also with upstream/master on OrbStack
* TestCmdDescribe fails on Docker Desktop with either code.

rfay:
TestCmdDescribe succeeds on macstadium-m1, which is DD 4.41.2

rfay:
I guess this may be 
* https://github.com/docker/for-mac/issues/7693

rfay:
Pulling this even though it still had trouble with arm64-mutagen-docker-desktop

PR #7374, opened 2025-06-13
build: Remove old WindowsPathToCygwinPath which supported docker toolbox

rfay:
## The Issue

Studying why DDEV doesn't work with Rancher Desktop on Windows I discovered old code that claimed it was only for long obsolete and not supported Docker Toolbox. 

Docker Toolbox support was finally removed in 2000.

Remove that code.


rfay:
Moved commits from this one to 
* #7375

PR #7375, opened 2025-06-13
fix: Remove COMPOSE_CONVERT_WINDOWS_PATHS to make Rancher Desktop (Windows) work

rfay:
## The Issue

We've had COMPOSE_CONVERT_WINDOWS_PATHS set for docker-compose for ages.  It converts something like C:\x\y to /c/x/y, which doesn't really work in most contexts other than CYGWIN/git-bash, but it does somehow work in Docker Desktop mounts.

This caused Rancher Desktop on Windows to fail to mount:
* https://github.com/rancher-sandbox/rancher-desktop/issues/8775

Remove the environment variable and do a test run.

## TODO

- [x] Soften docs about Rancher Desktop for Windows
- [x] Don't hardcode "Docker Desktop For Windows" in `ddev version`
- [x] Consider automated testing support

- [x] Basic testing on Docker Desktop for Windows (Review the .ddev/.ddev-docker-compose-full.yaml mounts)
- [x] Basic testing on Rancher Desktop for Windows

## Automated Testing Overview

- [x] Consider adding automated testing for RD


stasadev (approved):
I used VMware Workstation on Linux to test it with Windows 11 guest.

- ✅ Docker Desktop For Windows + v1.24.6
- ❌ Rancher Desktop For Windows + v1.24.6
- ✅ Docker Desktop For Windows + this PR
- ✅ Rancher Desktop For Windows + this PR

Here is what I see with Rancher Desktop and DDEV v1.24.6:

```
$ ddev start
...
ls: cannot access '/mnt/ddev_config/nginx_full/nginx-site.conf': No such file or directory

Something is wrong with your Docker provider and /mnt/ddev_config is not mounted from the project .ddev folder. Your project cannot normally function successfully with this situation. Is your project in your home directory?
...
```

Using this PR, `ddev start` works fine.

---

The difference in `.ddev/.ddev-docker-compose-full.yaml`:

```diff
services:
    db:
        volumes:
            - bind:
                create_host_path: true
-             source: /c/Users/testbot/workspace/l12/.ddev
+             source: C:\Users\testbot\workspace\l12\.ddev
              target: /mnt/ddev_config
              type: bind
            - bind:
                create_host_path: true
-             source: /c/Users/testbot/workspace/l12/.ddev/db_snapshots
+             source: C:\Users\testbot\workspace\l12\.ddev\db_snapshots
              target: /mnt/snapshots
              type: bind
    web:
        volumes:
            - bind:
                create_host_path: true
              read_only: true
-             source: /c/Users/testbot/workspace/l12/.ddev
+             source: C:\Users\testbot\workspace\l12\.ddev
              target: /mnt/ddev_config
              type: bind
            - bind:
                create_host_path: true
-             source: /c/Users/testbot/workspace/l12/.ddev/xhprof
+             source: C:\Users\testbot\workspace\l12\.ddev\xhprof
              target: /usr/local/bin/xhprof
              type: bind
```

---

There is a hardcoded string in `ddev version` that says "Docker Desktop For Windows" when using Rancher Desktop:

https://github.com/ddev/ddev/blob/c382808a854acd41411ea7a65e291e9a3a01f5e5/pkg/version/version.go#L53-L55

rfay:
> There is a hardcoded string in ddev version that says "Docker Desktop For Windows" when using Rancher Desktop:

Thanks for reminding me about this. I noticed it but could have forgotten.

rfay:
I guess the testing is just not going to work, I'll need to back off the docs and put it in the "should work but no automated testing" category. And turn off testing for it. Testing is so inconsistent!

rfay:
Reverted the strong support and "tested" statements in docs, going back to "Rancher seems to work"

stasadev (approved):
Looks good to me!

PR #7385, opened 2025-06-17
feat: return real exit code from `ddev exec` and add quiet flag to it, fixes #3518

### LINKED ISSUE #3518: Add option to not display result of `ddev exec`

andreashager:
## Issue

- #3518

DDEV adds by default an command line text `Failed to execute command  ...` to stderr.

- Returns the exit code from the subcommand in `ddev exec`
- Adds a flag to suppress  `Failed to execute command  ...` with a command line flag for command `ddev exec`.

Usual output:
```
$ ddev exec "echo stdout; echo >&2 stderr; exit 42"
stdout
stderr

Failed to execute command `echo stdout; echo >&2 stderr; exit 42`: exit status 42

$ echo $?
42
```

Using quiet:

```
$ ddev exec -q "echo stdout; echo >&2 stderr; exit 42"
stdout
stderr

$ echo $?
42
```

---

Previously, this logic could be used to suppress the error message:

```
$ ddev exec "echo stdout; echo >&2 stderr; exit 42" 2>/dev/null
stdout
stderr
```

### DIFF
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"errors"
-	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -80,7 +79,7 @@ ddev exec --raw -- ls -lR`,
 				exitCode = exiterr.ExitCode()
 			}
 			if !quiet {
-				output.UserErr.Errorf(util.ColorizeText(fmt.Sprintf("Failed to execute command `%s`: %v", opts.Cmd, err), "red"))
+				util.Error("Failed to execute command `%s`: %v", opts.Cmd, err)
 			}
 			output.UserErr.Exit(exitCode)
 		}
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -67,9 +67,9 @@ ddev exec --raw -- ls -lR`,
 		}
 
 		_, _, err = app.Exec(opts)
-		verbose, _ := cmd.Flags().GetBool("verbose")
+		quiet, _ := cmd.Flags().GetBool("quiet")
 
-		if err != nil && verbose {
+		if err != nil && !quiet {
 			util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
 		}
 	},
@@ -103,7 +103,7 @@ func init() {
 	_ = DdevExecCmd.RegisterFlagCompletionFunc("service", ddevapp.GetServiceNamesFunc(true))
 	DdevExecCmd.Flags().StringVarP(&execDirArg, "dir", "d", "", "Defines the execution directory within the container")
 	DdevExecCmd.Flags().Bool("raw", true, "Use raw exec (do not interpret with Bash inside container)")
-	DdevExecCmd.Flags().BoolP("verbose", "v", false, "Add DDEV default error output")
+	DdevExecCmd.Flags().BoolP("quiet", "q", false, "Disables DDEV default error output")
 	// This requires flags for exec to be specified prior to any arguments, allowing for
 	// flags to be ignored by cobra for commands that are to be executed in a container.
 	DdevExecCmd.Flags().SetInterspersed(false)
--- a/cmd/ddev/cmd/exec.go
+++ b/cmd/ddev/cmd/exec.go
@@ -1,10 +1,14 @@
 package cmd
 
 import (
+	"errors"
+	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/ddev/ddev/pkg/ddevapp"
+	"github.com/ddev/ddev/pkg/output"
 	"github.com/ddev/ddev/pkg/util"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


rfay:
I don't think it's a great idea to disable `ddev exec` status by default. Before taking the time to make a PR to change DDEV behavior, talking about it in an issue would be better.

If you don't want to show status, you can use

```
hooks:
  post-start:
    - exec: do-something || true
```

rfay:
Oh, I see #3518 - but that was about a different topic than disabling exec status by default. It was about an option.

andreashager:
First I was thinking to add a quiet-flag to DDEV exec but for me quiet would mean that there is no output 😅 Then I thougt its more natural, if you want more information on a command that you add the verbose flag to it. Thats why I made it this way around 😊

But you are right, changing the default behavior of DDEV exec is also not a good idea - I will adapt my PR to add a flag to disable this output. What name would you suggest for this flag? 😊

rfay:
There are two things, right? Behavior when a hook fails, and showing the output of an exec hook?

andreashager:
What do you mean? 🤔 
It has nothing todo with hooks - the command `ddev exec` itself should get the flag I think 😅

Or maybe I misunderstood something completely 😓

rfay:
I guess my head was in the wrong place. It was early. Sorry.

rfay:
I'm sure you saw the examples I just put in the related issue:

https://github.com/ddev/ddev/issues/3518#issuecomment-2980958973

When you fiddle with the PR, please update the title and add more detail to the OP, thanks. The examples given here might be examples you can use in the "Manual testing" section.

cmd/ddev/cmd/exec.go
```diff
@@ -67,9 +67,14 @@ ddev exec --raw -- ls -lR`,
...
 
 		_, _, err = app.Exec(opts)
+		quiet, _ := cmd.Flags().GetBool("quiet")
 
 		if err != nil {
-			util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
+			if !quiet {
+				util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
+			} else {
+				os.Exit(1)
```
rfay: Don't we know what error it exited with, and can use that here? The reality is we should probably be exiting with a useful error code above instead of util.Failed(). We do in fact report the exit code in the err of the util.Failed(), but it's not easily accessible by the caller.

rfay: And do we want to exit with an error, or ignore the error completely (os.Exit(0))?

rfay (approved):
This works for me, but I'm not 100% sure it is accomplishing everything hoped for. DDEV no longer complains if you use `-q`, but whatever might have happened inside the container still complains. Is that OK?
```
rfay@ubuntu-len:~/workspace/d11$ ddev exec ls kkk
ls: cannot access 'kkk': No such file or directory

Failed to execute command `ls kkk`: exit status 2


rfay@ubuntu-len:~/workspace/d11$ ddev exec -q ls kkk
ls: cannot access 'kkk': No such file or directory
```

andreashager:
> This works for me, but I'm not 100% sure it is accomplishing everything hoped for. DDEV no longer complains if you use `-q`, but whatever might have happened inside the container still complains. Is that OK?

I think yes! What I wanted to achive is that `ddev exec` behaves the same as I would execute the command native (ddev ssh + execute command). So if the directory kkk is not found the native error message of the ls command is fine I think 😊

cmd/ddev/cmd/exec.go
andreashager: I would go with an error code so that the caller could still check with `$?` that there was an error (even when the DDEV error message is supressed). 😊

If it is easily possible, the error code of the `ls` command would be a nice-to-have feature I think so that you can respond to the ls (or any other command) the right way? 🤔

stasadev: > Don't we know what error it exited with, and can use that here?

Good idea, I'll add that.

> And do we want to exit with an error, or ignore the error completely (os.Exit(0))?

Yes, we want to have the exit code, this is what I expect.

For example, both of the following fails, and I expect an error with a quiet flag:
```
echo test | grep foobar
echo test | grep -q foobar
```

cmd/ddev/cmd/exec.go
```diff
@@ -67,9 +71,18 @@ ddev exec --raw -- ls -lR`,
...
 
 		if err != nil {
-			util.Failed("Failed to execute command `%s`: %v", opts.Cmd, err)
+			exitCode := 1
+			var exiterr *exec.ExitError
+			if errors.As(err, &exiterr) {
+				exitCode = exiterr.ExitCode()
+			}
+			if !quiet {
+				output.UserErr.Errorf(util.ColorizeText(fmt.Sprintf("Failed to execute command `%s`: %v", opts.Cmd, err), "red"))
```
stasadev: It's not as polished as I wanted. Ideally, this should be a single call with automatic colored output. We have similar cases elsewhere. Since it's out of scope for this PR, I'll leave it as is.

stasadev (approved):
`ddev exec` now returns the exit code from the subcommand.

cmd/ddev/cmd/exec.go
stasadev: Found `util.Error`, will use that.

rfay:
Weird, in two different tests on different platforms (OrbStack and apache-fpm), TestUseEphemeralPort was broken. I didn't study carefully. Restarted Orbstack, but then went to look at the other one and my alarm bells went off. Haven't looked at it carefully.

Issue #4443, opened 2022-12-10
ddev adds a domain to /etc/hosts although it is resolvable on OSX (should be fixed with golang v1.20)


helhum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Run a Diagnostic and Paste Link Here

OS Information: Darwin 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:23 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T6000 arm64
ProductName:	macOS
ProductVersion:	12.5.1
BuildVersion:	21G83
DDEV version:  ITEM             VALUE                                   
 DDEV version     v1.21.3                                 
 architecture     arm64                                   
 db               drud/ddev-dbserver-mariadb-10.4:v1.21.3 
 dba              phpmyadmin:5                            
 ddev-ssh-agent   drud/ddev-ssh-agent:v1.21.3             
 docker           20.10.20                                
 docker-compose   v2.10.2                                 
 docker-platform  docker-desktop                          
 mutagen          0.16.0                                  
 os               darwin                                  
 router           drud/ddev-router:v1.21.3                
 web              drud/ddev-webserver:v1.21.3             

### Expected Behavior

I have set `project_tld: helhum.test` and have a dnsmasq server running on localhost with `/etc/resolver/test` instructed to use dnsmasq and also resolve to 127.0.0.1

This works, which I can verify with `ping foo-whatever.helhum.test` correctly resolving to localhost.

When I now do `ddev start`, I would expect, that ddev does not add the project tlds, that can perfectly be resolved by Mac OS to be added to the `/etc/hosts` file.

### Actual Behavior

ddev show the warning: `The hostname <project>.helhum.test is not currently resolvable, trying to add it to the hosts file ` and adds the entry to `/etc/host`

### Steps To Reproduce

1. Setup local domain resolving by configuring a [etc/resolver](https://medium.com/@phaneendrakrishnareddy/using-dnsmasq-for-local-development-on-os-x-7f17f78c7198)
2. Add a `project_tld: ` according to above configuration
3. `ddev start`

### Anything else?

Seems like [Go 1.20 changed DNS resolving](https://danp.net/posts/macos-dns-change-in-go-1-20/) on OSX when using go net lib, which would seemingly address and fix my request.

[...]

PR #7392, opened 2025-06-20
feat: Improve elevation/escalation for hosts editing, including WSL2, fixes #6440, fixes #5324

### LINKED ISSUE #5324: Improve DDEV's packaging of gsudo
Discussion with @gerardog regarding this in:
* https://github.com/gerardog/gsudo/pull/299
* https://github.com/gerardog/gsudo/issues/296

_Originally posted by @rfay in https://github.com/ddev/ddev/issues/5314#issuecomment-1708813273_
            
---
Hi @rfay, I wanted to understand better how ddev uses gsudo, so I cloned every ddev repo:

```
$user = "ddev"
$apiUrl = "https://api.github.com/users/$user/repos?per_page=100"
$outputDir = "C:\git\ddev"

$response = Invoke-RestMethod -Uri $apiUrl
foreach ($repo in $response) {
    $repoName = $repo.name
    $repoUrl = $repo.clone_url
    Write-Host "Cloning $repoName from $repoUrl"
    git clone $repoUrl "$outputDir\$repoName"
}
```
and search for the gsudo literal... I mostly found it's used in:

- docs (*.md)
- install scripts
- choco package as a dependency
- nullsoft installer

So, I didn't find a place where gsudo is used as part of the actual application... Am I right??
If that is the case, is it that you just want the end user to have gsudo available for convenience?

I guess I should first understand if gsudo is a hard dependency, a must-have, or a desired convenience tool. The difference is the first one requires gsudo, the later can prompt the user if wants to install it.

The friendly way of installing gsudo is thru the `.msi` file... It has the convenience that the user gets the complete experience, 
 plus a clean uninstall, can upgrade anytime later using winget upgrade. (Yeah, winget upgrades even if the msi is installed manually), etc...

Currently gsudo is portable but may misbehave if two different versions are running at the same time. That is because as one protection against hackers is that an elevated gsudo will only accept calls from the same `gsudo.exe` file. Running multiple versions simultaneously may fail, hence is prefered to have only one gsudo install. This problem may be mitigated with a code change, but currently is not.

We can dig into the alternatives which I think are:
- Download the `.msi` and install gsudo from ddev installer...
- Make ddev launch gsudo installer... (either .msi or the .ps1 [install script](github.com/gerardog/gsudo/tree/master/installgsudo.ps1) which has an possible inconvenient 'press a key' wait on the end that we can remove.
- Opting for the portable way, by a) mitigating the problem mentioned above and b) embedding a non-virus flagged version inside the installer.

### LINKED ISSUE #6440: WSL2: Use better escalation technique for editing /etc/hosts

In Windows or WSL2, we use Windows escalation to add non-resolvable hostnames to `C:\Windows\system32\drivers\etc\hosts`. Most folks on WSL2 use their browser on the Windows side, so this is important to most people.

However, the technique we use is really quite clunky. We use `ddev.exe` installed on the Windows side to run `ddev.exe hostname` , which invokes gsudo, etc. 

We don't actually need to have all of ddev.exe installed on the Windows side, which only confuses people, we could have a small exe which did only the hostname work.

### Describe your solution

Options include:
* A tiny exe which only had the hostname addition and removal job, super easy to write
* Bundle an exe like that with the WSL2 installation?
* Do escalation using go code from the WSL2 side?

### Describe alternatives

Other alternatives are always available, including having people run a local DNS service on the Windows side for their preferred project_tld. This is too complicated for the average user. (I note that the lando docs actually lay out how to do it.)

### Additional context

I think we might want to stop using Chocolatey for Windows-side installation by default, as we don't need it for a number of the things we used to use it for, like `nssm`. We still currently (might) need it for `gsudo` but I think we bundle gsudo anyway, so why bother?

rfay:
## The Issue

- #6440
- #5324 

Our escalation technique for editing hosts has been quite awkward, requiring a Windows-side ddev.exe on WSL2, for example, and requiring gsudo.exe to be installed.

- [x] Separate out the hosts-editing into its own binary (and `ddev-hostname.exe` on Windows)
- [x] Update build and packaging for `ddev-hostname` and `ddev-hostname.exe`

This will still require effort, but I'm adding the PR now so that I can add notes or TODOs that may be discovered along the way.

Basic functionality works already on macOS and Linux.

## TODO

- [x] deleting inactive hostnames is currently disabled.
- [x] Retain the `ddev hostname` command for backward compatibility (and make it call `ddev-hostname`)
- [x] Package `ddev-hostname.exe` with WSL2/Linux variants, must be correct architecture.
- [x] Windows installer needs to install `ddev-hostname.exe`
- [x] Consider running `ddev-hostname` from PATH instead of assuming exact location from ddev binary
- [x] `ddev-hostname` (macOS) and `ddev-hostname.exe` (windows and WSL2) must be signed/notarized; signed version must be included with linux/wsl2 package
- [ ] Remove the installation of ddev on windows side from the WSL2 install scripts. (It's possible that they won't need to be PowerShell scripts at all any more?) (Defer this to another PR)
- [ ] However, mkcert.exe is still required on Windows side, so has to be installed somehow (and it has to be in the PATH) (Defer this to another PR)
- [x] Add `ddev-hostname` to install script (install_ddev.sh) (must be deferred to another PR, preferably starting to use versioned script)
- [x] Consider adding `ddev-hostname.exe` to homebrew recipe for when used on wsl2
- [ ] `ddev-hostname` seems to be larger than it should be. Take a look to figure out why.
- [x] Consider adding `ddev-hostname.exe` installation to the Arch AUR repository. (There isn't apparently an easy way to delete it if not on WSL2, but maybe that doesn't matter)
- [x] Sort out the "This file is in a location outside your local network" problem on WSL2, either by copying file to Windows or something.
- [x] Add `--version` support to `ddev-hostname`
- [x] Homebrew (macOS) is not installing `ddev-hostname`
- [x] Compare Windows elevation code to gsudo code
- [x] Ask gsudo maintainer @gerardog for review

* Test releases are in https://github.com/ddev-test/ddev/releases

* To use homebrew from test releases: `brew install ddev-test/ddev/ddev`
* To use the install script, `export DDEV_OWNER=ddev-test`

* To use apt installs from test releases, use this for /etc/apt/sources.list.d/ddev.list:
```
deb [trusted=yes] https://apt.fury.io/rfay/ * *
```
* To use the ddev-test yum repo, /etc/yum.repos.d/ddev.repo:
```
[ddev]
name=ddev
baseurl=https://yum.fury.io/rfay
gpgcheck=0
enabled=1
```

### macOS Homebrew

- [x] Remove/uninstall any ddev installation or other binaries
- [ ] `brew install ddev-test/ddev/ddev`
- [x] `ddev config --project-tld=example.com` and `ddev start`, make sure it escalates properly without any other warning and actually does the deed so you can use the example.com url (check hosts file)
- [ ] Verify that the ddev binary is properly signed and notarized
```
$ codesign -dvvv ddev 2>&1 |grep Authority
Authority=Developer ID Application: Localdev Foundation (9HQ298V2BW)
Authority=Developer ID Certification Authority
```
- [ ] Verify ddev-hostname is signed and notarized

### macOS Install Script

- [x] Remove/uninstall any ddev installation or other binaries
- [ ] export GITHUB_OWNER=ddev-test
- [ ] `curl -fsSL https://raw.githubusercontent.com/ddev-test/ddev/refs/heads/20250619_rfay_wsl2_escalation/scripts/install_ddev.sh | bash`
- [ ] `ddev config --project-tld=example.com` and `ddev start`, make sure it escalates properly without any other warning and actually does the deed so you can use the example.com url (check hosts file)
- [ ] Verify that the ddev and ddev-hostname binaries are properly signed and notarized

### Traditional Windows
- [x] Remove/uninstall any ddev installation or other binaries
- [x] Install DDEV using the test installer, make sure there are no warnings about source of any item
- [x] Check "digital signatures" for the installer, ddev.exe, and ddev-hostname.exe to see that they got signed. 
- [x] Use it
- [x] `ddev config --project-tld=example.com` and `ddev start`, make sure it escalates properly without any other warning and actually does the deed so you can use the example.com url (check hosts file)

### WSL2/Ubuntu

- [x] Install ddev on windows using the Windows installer from ddev-test/ddev
- [x] Remove/uninstall any existing ddev
- [x] Install DDEV in WSL2 using the downloaded deb or rpm packages
- [x] `ddev config --project-tld=example.com` and `ddev start`, make sure it escalates properly using windows-side escalation.
- [x] Verify that the normal installation script still works. This requires a little bit of workaround to anticipate the new release. Probably the best way to do it is to use [ddev-test/ddev](https://github.com/ddev-test/ddev) script (on the correct branch). That should get latest ddev installer etc.
- [x] `ddev poweroff` and `ddev hostname --remove-inactive`, then `ddev config global --wsl2-no-windows-hosts-mgt` and `ddev start`. Hosts should be set in wsl2 /etc/hosts

### Ubuntu Linux

- [x] Use rfay repo for installation
- [x] Make sure `ddev-hostname` and `ddev hostname` work properly with escalation
- [x] Use install script (install_ddev.sh), requires minor adjustment to test. Verify behavior
- [x] Use homebrew (brew install ddev-test/ddev/ddev) and verify behavior

### RPM-based distro (Fedora)

- [x] Use rfay repo for installation
- [x] Make sure `ddev-hostname` and `ddev hostname` work properly

## Arch-based repo

- [x] Install via AUR https://aur.archlinux.org/pkgbase/ddev-test-bin/ (`yay -S ddev-test-bin`)
- [x] Make sure `ddev-hostname` and `ddev hostname` work properly

## Automated Testing Overview

In almost all cases escalation is not possible in our test environments (sudo requires password, for example, or windows requires escalation. So mostly this is all on us for manual testing.


rfay:
This is working at a basic level now in all the places.

Next steps:
- [x] provide the ddev_hostname.exe to WSL2 installation or build
- [x] make sure we have a signed ddev_hostname.exe

rfay:
WSL2 running ddev_hostname.exe from /usr/bin: "This file is in a location outside your local network"

wsl2_location_outside

This can be avoided by copying ddev_hostname.exe to the windows side and executing it from there.

Workarounds:
* Copy the file to Windows filesystem and execute it from there
* Control Panel → Internet Options → Security → Local intranet → Sites → Advanced, add `\\wsl.localhost`

image

A number of ideas for copying even during the apt install via [chatgpt suggestions](https://chatgpt.com/share/68589174-0890-8003-89ba-db99976505c3)

pkg/ddevapp/hostname_mgt.go
```diff
@@ -3,76 +3,18 @@ package ddevapp
...
-	}
-	if err != nil {
-		return false, fmt.Errorf("unable to open hosts file: %v", err)
-	}
-	return hosts.Has(dockerIP, hostname), nil
-}
-
 // AddHostsEntriesIfNeeded will run sudo ddev hostname to the site URL to the host's /etc/hosts.
 // This should be run without admin privs; the DDEV hostname command will handle escalation.
 func (app *DdevApp) AddHostsEntriesIfNeeded() error {
```
stasadev, line 19: - Ref: https://github.com/ddev/ddev/issues/4443#issuecomment-2996185968

I think we should make a replacement in `AddHostsEntriesIfNeeded()`:

```diff
-if app.UseDNSWhenPossible && globalconfig.IsInternetActive() {
+if app.UseDNSWhenPossible {
```

If we look up each host individually, we end up checking `test.ddev.site` for every hostname, even if the hostname isn't under `*.ddev.site`, which duplicates effort.

stasadev: Edit: `globalconfig.IsInternetActive()` checks `test.ddev.site` only once, not multiple times.
The point is that DNS lookup for each hostname may not depend on whether `test.ddev.site` is reachable, as long as it does not end with `.ddev.site`.

rfay: I guess you're saying that we might have a situation where DNS works but no internet access is available. The assumption in this code is that DNS would not be usable if we don't have internet access, which is usually not a bad assumption. I probably currently like this assumption, but I guess you're right that `IsInternetActive()` isn't fundamentally needed. We'll do some lookups in this case when offline, but they shouldn't take much time.
 https://github.com/ddev/ddev/blob/8be2eb764185f95b741ce2c09cfc5ed6c28e2567/pkg/ddevapp/hostname_mgt.go#L36-L46

rfay: Made that change in https://github.com/ddev/ddev/pull/7392/commits/a3fcbee5b606bc0f9ef87b48cedd962525ca9320

rfay:
It's possible that ddev_hostname.exe should be split out and deployed with a Windows installer, perhaps along with mkcert.exe.

rfay:
Experimenting with making the binaries smaller (and finding out why ddev_hostname is larger than expected) I didn't solve the ddev_hostname question, but did make the binaries significantly smaller. Here is the difference in size between before and after symbol stripping change to Makefile in https://github.com/ddev/ddev/pull/7392/commits/80334270b35845933c25d7845a510e018ef2e2dc 

```
rfay@ubuntu-len:~/workspace/ddev$ make
building .gotmp/bin/linux_amd64/ddev from ./cmd/... ./pkg/... GORACE=halt_on_error=1 CGO_ENABLED=0 BUILDARGS=
building .gotmp/bin/linux_amd64/ddev_hostname from ./cmd/... ./pkg/... GORACE=halt_on_error=1 CGO_ENABLED=0 BUILDARGS=
building .gotmp/bin/windows_amd64/ddev_hostname.exe from ./cmd/... ./pkg/... GORACE= CGO_ENABLED=0 BUILDARGS=
rfay@ubuntu-len:~/workspace/ddev$ du -sh .gotmp/bin/linux_amd64/ddev*
24M     .gotmp/bin/linux_amd64/ddev
24M     .gotmp/bin/linux_amd64/ddev_gen_autocomplete
11M     .gotmp/bin/linux_amd64/ddev_hostname
12M     .gotmp/bin/linux_amd64/ddev_hostname.exe
rfay@ubuntu-len:~/workspace/ddev$ git checkout 20250619_rfay_
20250619_rfay_remove_go_homedir   20250619_rfay
```_wsl2_escalation
rfay@ubuntu-len:~/workspace/ddev$ git checkout 20250619_rfay_wsl2_escalation
Previous HEAD position was e3445df1c try to build ddev_hostname.exe without -race
Switched to branch '20250619_rfay_wsl2_escalation'
Your branch is up to date with 'upstream/20250619_rfay_wsl2_escalation'.
rfay@ubuntu-len:~/workspace/ddev$ make
building .gotmp/bin/linux_amd64/ddev from ./cmd/... ./pkg/... GORACE=halt_on_error=1 CGO_ENABLED=0 BUILDARGS=
building .gotmp/bin/linux_amd64/ddev_hostname from ./cmd/... ./pkg/... GORACE=halt_on_error=1 CGO_ENABLED=0 BUILDARGS=
building .gotmp/bin/windows_amd64/ddev_hostname.exe from ./cmd/... ./pkg/... GORACE= CGO_ENABLED=0 BUILDARGS=
rfay@ubuntu-len:~/workspace/ddev$ du -sh .gotmp/bin/linux_amd64/ddev*
17M     .gotmp/bin/linux_amd64/ddev
17M     .gotmp/bin/linux_amd64/ddev_gen_autocomplete
7.4M    .gotmp/bin/linux_amd64/ddev_hostname
7.7M    .gotmp/bin/linux_amd64/ddev_hostname.exe
```

rfay:
@gerardog would you mind reviewing (mostly elevate_windows.go ?) This does remove the need for #5324 , but DDEV's use case is so narrow (just editing hosts file) that for security and installation purposes it's far better not to require a general-purpose tool like gsudo.  I know your experience with this is deep and would very much appreciate your comments.

cmd/ddev/cmd/hostname.go
```diff
@@ -54,14 +49,14 @@ to allow DDEV to modify your hosts file. If you are connected to the internet an
 
 		// If requested, remove the provided host name and exit
 		if removeHostnameFlag {
-			err = ddevapp.RemoveHostEntry(name, dockerIP)
+			out, err := exec.RunHostCommand(ddevHostnameBinary, "--remove", name, dockerIP)
 			if err != nil {
-				util.Warning("Failed to remove host entry %s: %v", name, err)
+				util.Warning("Failed to remove hosts entry %s:%s: %v (output='%v'", name, dockerIP, err, out)
```
stasadev:
```suggestion
				util.Warning("Failed to remove hosts entry %s:%s: %v (output='%v')", name, dockerIP, err, out)
```

cmd/ddev/cmd/hostname.go
```diff
@@ -71,10 +66,9 @@ to allow DDEV to modify your hosts file. If you are connected to the internet an
 			os.Exit(1)
 		}
 		// By default, add a host name
-		err = ddevapp.AddHostEntry(name, dockerIP)
-
+		out, err := exec.RunHostCommand(ddevHostnameBinary, "--add", name, dockerIP)
 		if err != nil {
-			util.Warning("Failed to add hosts entry %s: %v", name, err)
+			util.Warning("Failed to add hosts entry %s:%s: %v (output='%v'", name, dockerIP, err, out)
```
stasadev:
```suggestion
			util.Warning("Failed to add hosts entry %s:%s: %v (output='%v')", name, dockerIP, err, out)
```

cmd/ddev_hostname/ddev_hostname.go
```diff
@@ -0,0 +1,101 @@
...
+		// If operating on one host name, two arguments are required
+		if len(args) != 2 {
+			util.Failed("Invalid arguments supplied. Please use 'ddev_hostname [hostname] [ip]'")
+		}
+
+		name, dockerIP := args[0], args[1]
+		var err error
+
+		util.Debug("Elevating privileges to add host entry %s -> %s", name, dockerIP)
+		elevateIfNeeded()
```
stasadev: `hostname.IsHostnameInHostsFile(name)` should be run first, and depending on what result is and what flag is provided, run `elevateIfNeeded()` only when it's really needed.

If `AddHostEntry`, check if it's already there, and if it's not, only then call `sudo`.
The same with `RemoveHostEntry`.
And `checkHostnameFlag` doesn't need elevation.

The same thing applies to `cmd/ddev/cmd/hostname.go`.

cmd/ddev_hostname/elevate.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+func elevateIfNeeded() {
+	// If we’re not root (UID 0), re‐exec via sudo
+	if syscall.Geteuid() != 0 {
```
stasadev: It can be written like this by removing one nested `if`:

```go
if syscall.Geteuid() == 0 {
	// we’re already root, continue
	return
}
// If we’re not root (UID 0), re‐exec via sudo
...
```

winpkg/ddev.nsi
```diff
@@ -894,9 +679,6 @@ Section Uninstall
   Delete "$INSTDIR\mkcert_license.txt"
   Delete "$INSTDIR\mkcert.exe"
 
-  Delete "$INSTDIR\gsudo_license.txt"
-  Delete "$INSTDIR\${GSUDO_SETUP}"
-
   Delete "$INSTDIR\license.txt"
   Delete "$INSTDIR\ddev.exe"
```
stasadev, line 683:
```suggestion
  Delete "$INSTDIR\ddev.exe"
  Delete "$INSTDIR\ddev_hostname.exe"
```

pkg/hostname/hostname.go
```diff
@@ -0,0 +1,153 @@
...
+		util.Warning("DDEV_NONINTERACTIVE is set. You must manually run '%s'", strings.Join(args, " "))
+		return "", nil
+	}
+	_, err = os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("could not get home directory for current user. Is it set?")
+	}
+
+	if !IsDdevHostnameAvailable() {
+		return "", fmt.Errorf("%s is not installed, please install it", ddevhostnameBinary)
```
stasadev: I'm not sure if this is a useful error message, people on WSL2 after running `sudo apt-get upgrade` will only get a new `ddev` on the Linux side, it needs some kind of documentation reference so they can rerun the installation script.

rfay: I added a little bit of doc, and will link to it in this error, but also opened
* https://github.com/ddev/ddev/issues/7407

stasadev: Thank you, that's what I asked for.

cmd/ddev_hostname/elevate.go
rfay: Reworked this in both elevate.go and elevate_windows.go

cmd/ddev/cmd/hostname.go
```diff
@@ -52,29 +47,31 @@ to allow DDEV to modify your hosts file. If you are connected to the internet an
...
 		var err error
 
+		hostnameInHostsFile, err := hostname.IsHostnameInHostsFile(name)
+		if err != nil {
+			util.Warning("Could not check existence of %s in hosts file: %v", name, err)
+		}
 		// If requested, remove the provided host name and exit
-		if removeHostnameFlag {
-			err = ddevapp.RemoveHostEntry(name, dockerIP)
+		if removeHostnameFlag && hostnameInHostsFile {
```
stasadev: Fix condition check:

```suggestion
		if removeHostnameFlag {
			if !hostnameInHostsFile {
				return
			}
```

rfay: Oops. I had noticed that the combined condition wasn't going to work so was reverting but didn't do it quite right.

rfay:
Now for another whole round of manual testing (for me, not you)

rfay:
Partial retesting:
- [x] macOS/Homebrew
- [x] macOS/install_ddev.sh
- [x] Traditional Windows
- [x] WSL2/apt
- [x] Linux/apt

rfay:
```
rfay@rfay-mba-m4:/opt/homebrew/Cellar/ddev/1.23.73/bin$ codesign -dvvv ddev 2>&1 |grep Authority
Authority=Developer ID Application: Localdev Foundation (9HQ298V2BW)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
rfay@rfay-mba-m4:/opt/homebrew/Cellar/ddev/1.23.73/bin$ codesign -dvvv ddev-hostname 2>&1 |grep Authority
Authority=Developer ID Application: Localdev Foundation (9HQ298V2BW)
Authority=Developer ID Certification Authority
Authority=Apple Root CA

```

Issue #7366, opened 2025-06-10
Drop dependency on github.com/mitchellh/go-homedir


dolmen:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Remove dependency on `github.com/mitchellh/go-homedir` and use instead [`os.UserHomeDir()`](https://pkg.go.dev/os#UserHomeDir).

### Describe your solution

Use instead [`os.UserHomeDir()`](https://pkg.go.dev/os#UserHomeDir).

### Additional context

1. [`github.com/mitchellh/go-homedir`](https://github.com/mitchellh/go-homedir) is outdated. The reason why that module existed has disappeared. See https://github.com/mitchellh/go-homedir/issues/34#issuecomment-2011915588
2. `github.com/mitchellh/go-homedir` has been unmaintained for years and now has been archived.

rfay:
PR would be appreciated, thanks for exploring this!

rfay:
The reality is that we use os.UserHomeDir(). 

We use mitchellh/go-homedir only for homedir.Expand() in 

* https://github.com/ddev/ddev/blob/b6c3c7ba4694a3a5334ae75394bf414fd0325200/pkg/appimport/appimport.go#L22

* One place in providerGit_test.go

I don't believe there's an equivalent to homedir.Expand() in go library, so we'd have to copy that out I guess.

dolmen:
See https://github.com/rclone/rclone/pull/8598 as an example where I also had to copy the `Expand` func from `go-homedir`.

dolmen:
But maybe this is just enough:

```go
func expandTilde(p string) string {
    if strings.HasPrefix(p, "~/") {
        p = filepath.Join(os.UserHomeDir(), p[2:])
    }
    return p
}
```

rfay:
I'll have a PR for you to remove in a few minutes.

rfay:
Thanks for your review of
* https://github.com/ddev/ddev/pull/7390

PR #7396, opened 2025-06-23
build: upgrade mapstructure to v2

dolmen:
## The Issue

Go package [`github.com/mitchellh/mapstructure`](https://github.com/mitchellh/mapstructure) is archived, so no updates will be published. Similar issue to #7366.

`github.com/go-viper/mapstructure/v2` is the blessed fork. See https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc

Remove github.com/mitchellh/mapstructure (archived repository) and replace with its blessed fork at github.com/go-viper/mapstructure/v2


rfay:
I think we could use this very effectively for our processing of composeyaml, where we read in .ddev/.ddev-docker-compose-full.yaml to see all the things and adjust some. What we do there is quite complex, and this would be a lot easier.

PR #7399, opened 2025-06-25
feat: add Linux support for heidisql command

punkrock34:
- Adds support for Linux native and Wine-based installations.
- Preserves Windows and WSL2 compatibility.

## The Issue

No issue number; enhancement based on missing Linux support.

- Adds support for launching HeidiSQL on native Linux and Wine-based environments.
- Maintains existing compatibility with Windows and WSL2.

1. Run `ddev heidisql` on:
   - WSL2
   - Linux (native or Wine via Snap, installed through the Software app)
2. Confirm HeidiSQL launches and connects to the DDEV database.

Tested on WSL2 and Linux (native + Snap Wine via Software app)

## Automated Testing Overview

No automated tests required; this is a host-side integration script.

## Release/Deployment Notes

No special deployment steps. Improves compatibility for Linux users using HeidiSQL via Snap or Wine.

### DIFF
--- a/pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
+++ b/pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
@@ -2,31 +2,44 @@
 
 ## #ddev-generated: If you want to edit and own this file, remove this line.
 ## Description: Run HeidiSQL against current db
-## Usage: heidisql
-## Example: "ddev heidisql"
-## OSTypes: windows,wsl2
-## HostBinaryExists: /mnt/c/Program Files/HeidiSQL/heidisql.exe,C:\Program Files\HeidiSQL\Heidisql.exe
-
-arguments="--host=\"127.0.0.1\" --port=${DDEV_HOST_DB_PORT} --user=root --password=root --description=${DDEV_SITENAME}"
+## Usage: heidisql [database]
+## Example: "ddev heidisql" or "ddev heidisql my_other_db"
+## OSTypes: windows,wsl2,linux
+## HostBinaryExists: /mnt/c/Program Files/HeidiSQL/heidisql.exe,C:\Program Files\HeidiSQL\Heidisql.exe,/usr/bin/heidisql,/usr/local/bin/heidisql,/snap/bin/heidisql-wine
 
 if [ "${DDEV_PROJECT_STATUS}" != "running" ] && [ -z "$no_recursion" ]; then
   echo "Project ${DDEV_PROJECT} is not running, starting it"
-  ddev start
-  start_exit_code=$?
-  if [ $start_exit_code -ne 0 ]; then
-    exit $start_exit_code
-  fi
-  # run this script again, as the environment is updated after "ddev start"
+  ddev start || exit $?
   no_recursion=true ddev "$(basename "$0")" "$@"
   exit $?
 fi
+
+arguments=(
+  "--host=127.0.0.1"
+  "--port=${DDEV_HOST_DB_PORT}"
+  "--user=root"
+  "--password=root"
+  "--description=${DDEV_SITENAME}"
+)
+
 case $OSTYPE in
-  "win*"* | "msys"*)
-    '/c/Program Files/HeidiSQL/heidisql.exe' $arguments &
+  "win"* | "msys"*)
+    start "" "C:\Program Files\HeidiSQL\heidisql.exe" "${arguments[@]}" &
     ;;
-  # linux-gnu in this case is only WSL2 as selected in OSTypes above
   "linux-gnu")
-    # HeidiSQL is Microsoft only, but we want to start it from WSL2
-    "/mnt/c/Program Files/HeidiSQL/heidisql.exe" $arguments &
+    BINARIES=(
+      /usr/local/bin/heidisql
+      /usr/bin/heidisql
+      /snap/bin/heidisql-wine
+      "/mnt/c/Program Files/HeidiSQL/heidisql.exe"
+    )
+
+    for binary in "${BINARIES[@]}"; do
+      if [ -x "$binary" ]; then
+        echo "Launching $binary"
+        "$binary" "${arguments[@]}" &>/dev/null & disown
+        exit 0
+      fi
+    done
     ;;
 esac


pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
```diff
@@ -2,31 +2,134 @@
...
 
 if [ "${DDEV_PROJECT_STATUS}" != "running" ] && [ -z "$no_recursion" ]; then
   echo "Project ${DDEV_PROJECT} is not running, starting it"
-  ddev start
-  start_exit_code=$?
-  if [ $start_exit_code -ne 0 ]; then
-    exit $start_exit_code
-  fi
-  # run this script again, as the environment is updated after "ddev start"
+  ddev start || exit $?
```
rfay, line 13: Nice.

rfay (commented):
Are you using Wine somehow with DDEV? Can you say a bit more???

pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
punkrock34: Hello @rfay yes I did install heidisql via wine and also today noticed they just officially released for linux a .deb on their site

punkrock34:
Hi @stasadev 
Just wanted to let you know I finished adding PostgreSQL support.
Please let me know if there is anything else I should modify or anything that you believe is not correct.

pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
punkrock34: Just to clarify: DDEV is installed natively on Linux only HeidiSQL was installed via Wine.
I tested both the Wine version and the new native .deb and wanted to make sure the script works well for both.

rfay: I guess I would prefer not to support the wine option since it seems there is native linux support now (including a tarball). I do see that there's a long history of usage with wine. People must really love HeidiSQL if they're going that far to use it when there are so many great [database browser GUIS](https://ddev.readthedocs.io/en/stable/users/usage/database-management/#database-guis) that don't require that level of effort.

punkrock34: well then I can remove the wine version and keep only the native one working, is that okay?

rfay: Native is great, thanks. Wine just seems like such a stretch. But like I say, it must have been really important to people.

punkrock34: Removed wine in the latest commit, let me know if you need me to modify anything else

stasadev (commented):
@punkrock34, amazing, the code looks good at first glance, tomorrow I will test it in action.

What we need is to polish our documentation in two places, where it says about Windows binaries / Windows only:

https://github.com/ddev/ddev/blob/7ca1a2432ab2fefdc547c2debce3985b9477673e/docs/content/users/usage/commands.md#L807-L816

https://github.com/ddev/ddev/blob/7ca1a2432ab2fefdc547c2debce3985b9477673e/docs/content/users/usage/cli.md#L21-L21

stasadev (changes requested):
I tested it on Arch Linux:

- ❌ `ddev config --database=mariadb:10.11` with provided `--library` (not expected)
- ✅ `ddev config --database=mariadb:10.11` without `--library`
- ❌ `ddev config --database=mysql:8.0` with provided `--library` (not expected)
- ✅ `ddev config --database=mysql:8.0` without `--library`
- ✅ `ddev config --database=postgres:16` with provided `--library` 
- ❌ `ddev config --database=postgres:16` without `--library` (checked just to make sure)

We didn't provide `--library` by default for MariaDB/MySQL. Is it required for WSL2?

punkrock34:
@stasadev Good question!

The `--library` flag is something I added to let users explicitly choose the exact client library that HeidiSQL should use to connect. In HeidiSQL, when creating a connection, you have to select a library file (like `libmariadb.dll` or `libpq.dll`) to handle the connection.

On Windows, HeidiSQL comes bundled with most libraries in its installation folder, so users often don’t need to specify `--library`. But on Linux, those libraries aren’t included, and users might prefer to install and point to a specific `.so` library instead of relying on defaults. That’s why the script supports passing `--library`.

For WSL2, it’s usually not required to specify `--library` because the Windows binary of HeidiSQL already has the libraries it needs. But for native Linux binaries of HeidiSQL (e.g. `/usr/bin/heidisql`), it may be necessary to point to the correct `.so` library to avoid errors.

**Note:** Even though the [HeidiSQL help](https://www.heidisql.com/help.php#commandline) says the library “depends on the network protocol,” in practice HeidiSQL doesn’t always automatically pick the right library just because you set the nettype. For example, if you run a PostgreSQL connection with nettype `8` but don’t pass a library, HeidiSQL can still default to MariaDB’s library (like `libmariadb.dll`) instead of `libpq.dll`. So specifying the library explicitly avoids these mismatches.

More details on HeidiSQL’s command-line parameters are here: 
https://www.heidisql.com/help.php#commandline

pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
```diff
@@ -2,31 +2,133 @@
...
+  case "$db_type" in
+  postgres)
+    arguments+=("--library=${db_library:-libpq-10.dll}")
     ;;
-  # linux-gnu in this case is only WSL2 as selected in OSTypes above
-  "linux-gnu")
-    # HeidiSQL is Microsoft only, but we want to start it from WSL2
-    "/mnt/c/Program Files/HeidiSQL/heidisql.exe" $arguments &
+  mariadb | mysql)
+    arguments+=("--library=${db_library:-libmariadb.dll}")
```
stasadev: Do we actually need the default `--library` flag for MariaDB/MySQL here? Would it work for everyone if we removed it? Users could still set `--library` manually if needed.

On my system:

```
$ ldconfig -p | grep -o '/.*libmariadb\.so\.3' | head -n1
/usr/lib/libmariadb.so.3
```

But it doesn't match the database version used by DDEV (`mariadb:10.11`), which is probably why it fails for me when the `--library` flag is set:

```
$ pacman -Qo /usr/lib/libmariadb.so.3
/usr/lib/libmariadb.so.3 is owned by mariadb-libs 11.8.2-1
```

punkrock34: Makes sense now. (I misunderstood your original comment about the usage of `--library`.)

I’ll try dropping the default `--library` for MariaDB/MySQL and leave it only for Postgres. I’ll tweak the script so it uses `--library` for MariaDB/MySQL only if the user explicitly sets it.

Planning to test on Windows, WSL2, Ubuntu, and Arch to see if things work fine without it. I’ll let you know how it goes.

pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
```diff
@@ -2,31 +2,133 @@
...
     ;;
+  esac
+  start "" "C:\Program Files\HeidiSQL\heidisql.exe" "${arguments[@]}" &
```
rfay, line 29: Of course we should probably figure out some day how to use PROGRAMFILES64 or equivalent.

rfay (changes requested):
I wasn't able to get this going at first with either windows or WSL2.

The binary for Windows is in C:\Program Files\HeidiSQL, but when you run the installer it doesn't add that to the PATH. After I added to the system path, it worked fine in wsl2 using the windows binary.

The *.deb package didn't work for me at all in WSL2. The .tar.xz version gets 
`./heidisql: error while loading shared libraries: libgdk-x11-2.0.so.0: cannot open shared object file: No such file or directory`, probably there's lots of trouble with dependencies. I'll try it on native Ubuntu as well.

Update: It worked fine on native Ubuntu using the downloaded deb package. I tested on a plain mariadb database and it worked. However, specifying the database didn't really do anything. `ddev heidisql x db db` opened it; I expected it to default to the "x" database but it didn't (it landed on `db`)

I'm super happy to have Linux support here @punkrock34  and appreciate this contribution! Mostly I'm fine with it when you're good with it, it's a step forward, and it's great to have you maintaining it. It may need a tiny bit of installation help docs or something.

pkg/ddevapp/global_dotddev_assets/commands/host/heidisql
punkrock34: Hi @rfay, thanks so much for the feedback, really appreciate you testing this out!

Yeah, I agree it definitely needs some documentation. I’ve been a bit tied up the past few days, but I’ll try to push some updates over the weekend.

rfay:
Changed to draft for now. This is a great initiative, and it can go in when you think it's ready @punkrock34

rfay:
Rebased. If you can just take a look and see if there's anything else you'd like to do @punkrock34 we'll get it in.

punkrock34:
@rfay 
No worries will let you know sadly I didn't have any time to look again at this feature, I will try to keep it up to date with the main branch and come with an implementation this month most likely

rfay:
Rebased. I added a slight bit to the commands.md docs. All it needs now is a test on Linux.

rfay:
Could you push that change and I'll test later today?

punkrock34:
Thanks @rfay, sorry for not being able to support Postgres properly and thanks @stasadev for the changes

PR #7404, opened 2025-06-26
feat: Windows Native and WSL2 GUI installer

rfay:
## The Issue

Related to 
* #6440 and #7392

We've long had two pretty awkward PowerShell script that people needed to run to do basic WSL2 setup for DDEV.  One was for docker-ce and the other for Docker Desktop. People struggled a bit with the instructions. With an installer we can give more guidance 

This won't be ready for some time, and depends on other things that are in play. However, it currently builds an amd64 installer that does docker-ce, Docker Desktop, and works with Rancher Desktop

It could actually do the install for the traditional Windows as well.

## TODO

- [x] Consider consolidating all Windows installs into one installer. 
- [x] Add Rancher Desktop support to prompts in installer
- [x] Make sure that mkcert integration is working right for all. Can it be skipped if mkcert -install already run?
- [x] docs: Add Rancher Desktop limited support for WSL2
- [x] docs: Upgrade instructions (just run the installer)
- [x] docs: Add to manual install docs with instructions/suggestions for other distros. Improve docs in general.
- [x] installer: Add suggestion link if distro is not Ubuntu. 
- [x] installer: Consider allowing them to choose which distro to install in, not requiring them to use the default distro.
- [x] Makefile needs to support amd64/arm64 installers (and sign both)
- [ ] Do automated testing of the install (all 3 paths)
- [ ] Make sure the `recommends` status of packages is as we want it. (and followup possibly create a wsl2 package)
- [x] Review the nsi file carefully again
- [x] Program Files directory should be properly specified, not defaulted to C drive
- [x] amd64 and arm64 architectures
- [x] Improve Makefile logic so it only builds the exe files it needs
- [x] How will Chocolatey work? Assume that it's doing regular Windows install...
- [x] Consider a better icon and art for the initial screen
- [x] WSL2 paths should install current WSL2 binary (and ddev-hostname) if possible
- [x] Verify that apt-get upgrade after the "hold" actually updates to a new version (we replaced the package-installed binary with a built one, but we want apt-get to update it)
- [x] Add check for git-bash on traditional windows
- [x] Installer should have a page that invites people to support DDEV

* [Docker Installation docs](https://ddev--7404.org.readthedocs.build/en/7404/users/install/docker-installation/#docker-installation-windows)
* [DDEV Installation docs](https://ddev--7404.org.readthedocs.build/en/7404/users/install/ddev-installation/#ddev-installation-windows)
* [Upgrade docs](https://ddev--7404.org.readthedocs.build/en/7404/users/install/ddev-upgrade/)

- [ ] Install fresh Ubuntu, make it default, test with docker-ce, test ddev there
- [ ] Re-run installer on already installed Ubuntu, test ddev there. Verify that /etc/apt/sources.list.d entries are correct
- [ ] Install fresh Ubuntu, make it default, Enable Docker Desktop integration on the distro, test with Docker Desktop install
- [ ] Re-run installer with Docker Desktop setup, test ddev.
- [ ]


rfay:
I know this has a ways to go @stasadev but I'm almost to writing tests for it, and would appreciate if you could take the installer for a bit of a run. There are many permutations... I have been often recreating distros like Ubuntu-20.04, Ubuntu-22.04, Ubuntu-24.04. My normal one is just "Ubuntu".

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1302 @@
...
+    DetailPrint "Traditional Windows installation completed."
+    ${IfNot} ${Silent}
+        MessageBox MB_ICONINFORMATION|MB_OK "DDEV Traditional Windows installation completed successfully."
+    ${EndIf}
+FunctionEnd
+
+Function RunMkcertInstall
+    DetailPrint "Setting up mkcert.exe (Windows) for trusted HTTPS certificates..."
+    ${IfNot} ${Silent}
+        MessageBox MB_ICONINFORMATION|MB_OK "Now setting up mkcer.exet to enable trusted https. Please accept the mkcert dialog box that may follow."
```
stasadev:
```suggestion
        MessageBox MB_ICONINFORMATION|MB_OK "Now setting up mkcert.exe to enable trusted https. Please accept the mkcert dialog box that may follow."
```

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1302 @@
...
+    Pop $1
+    Pop $0
+
+    ; Install linux packages
+    DetailPrint "WSL($SELECTED_DISTRO): Installing required linux packages..."
+    nsExec::ExecToStack 'wsl -d $SELECTED_DISTRO -u root apt-get install -y ca-certificates curl gnupg gnupg2 libsecret-1-0 lsb-release pass'
+    Pop $1
+    Pop $0
+    ${If} $1 != 0
+        MessageBox MB_ICONSTOP|MB_OK "Failed to install dependencies. Please check the logs."
```
stasadev: It's unclear what logs I need to check:


I think it's better to show error details like it's done here:

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1302 @@
...
+    DetailPrint "WSL($SELECTED_DISTRO): Removing old Docker packages if present..."
+    nsExec::ExecToStack 'wsl -d $SELECTED_DISTRO -u root bash -c "apt-get remove -y -qq docker docker-engine docker.io containerd runc >/dev/null 2>&1"'
+    Pop $1
+    Pop $0
+
+    ; apt-get upgrade
+    DetailPrint "WSL($SELECTED_DISTRO): Doing apt-get upgrade..."
+    nsExec::ExecToStack 'wsl -d $SELECTED_DISTRO -u root bash -c "apt-get update && apt-get upgrade -y >/dev/null 2>&1"'
+    Pop $1
+    Pop $0
```
stasadev: I think all of these `apt-get` calls need `MessageBox` with error details.

mkdocs.yml
```diff
@@ -156,6 +156,7 @@ nav:
       - users/install/index.md
       - users/install/docker-installation.md
       - users/install/ddev-installation.md
+      - users/install/ddev-upgrade.md
       - 'Configuring Browsers': users/install/configuring-browsers.md
       - 'Upgrading': users/install/ddev-upgrade.md
```
stasadev: If this section is moved up, should be removed here
```suggestion
```

stasadev (changes requested):
WSL2 with Docker CE

No WSL2 found error is not fatal enough.
The installer gets stuck, but installs mkcert and files to `C:\Program Files\DDEV`
I expect that clicking Ok on error will either go back to install type selection or to close the installer.

no-wsl-found-1

WSL2 with Docker Desktop or Rancher Desktop

No WSL2 found error is not fatal enough.
The installer gets stuck, but installs mkcert and files to `C:\Program Files\DDEV`
I expect that clicking Ok on error will either go back to install type selection or to close the installer.

no-wsl-found-2

Traditional Windows

Docker provider check is too late and not fatal enough.
The installer gets stuck, but installs mkcert and files to `C:\Program Files\DDEV`, and only then shows error for Docker provider.
I expect that this check is done before `mkcert` and copying files.
I expect that clicking Ok on error will either go back to install type selection or to close the installer.

no-docker-provider-found

mkdocs.yml
rfay: Thanks. I couldn't actually see that it was there.  Now I can see that it is. Will just revert the change.

winpkg/ddev_windows_installer.nsi
rfay: Agreed.

rfay:
> No WSL2 found error is not fatal enough.

I tried this again with current code (`wsl --uninstall` and then try WSL2 installer). It was close to adequate.  (Fix system with prompted `wsl --update` and I didn't lose any distros)

> Fatal missing docker provider

On Docker Desktop/WSL2:

image


On Traditional Windows:

image

rfay:
Added check for git-bash/git
Improved experience when bash or docker desktop not available.
Has useful automated tests for all three approaches. The problem is they have to be escalated to run, so that's another hurdle.

rfay:
Running as "local service account", fully privileged:
```
gsudo nssm set buildkite-agent AppEnvironmentExtra "HOME=C:\buildkite-agent"
```

Restart service. 

Now we have to figure out how to get this agent to pick up this type of test. 

I set up tb-win11-10 to run as "local service account".  And https://buildkite.com/ddev/ddev-windows-mutagen/builds/5311 is running

rfay:
@tyler36 if you could try out as much of the installer as you're willing it would be very welcome. You can use additional WSL2 distros, it doesn't have  to touch your normal distro that you use.

colans:
@rfay: Is there another place to get the installer?  https://nightly.link/ seems blocked over here.

rfay:
HeidiSQL has a decent installer finish page:

image

rfay:
Happy to get it for you any way you like @colans , and thanks! What works for you? Dropbox? Google?

colans:
Maybe just upload it here somewhere as GitHub works (just in a comment is fine if there's no better place).

rfay:
[ddev-windows-amd64-installer.zip](https://github.com/user-attachments/files/21129805/ddev-windows-amd64-installer.zip)

rfay:
Thanks @tyler36 - The "Unknown publisher" is because you're using the unsigned PR build. I can push the same thing to ddev-test and get it signed, but this is a fine thing.

"Ubuntu-only": We don't know how to select distros except by name. Other suggestions are welcome. I did look into explicitly supporting Debian as well, but it's already too complicated (on the code side)

> Text says to click "Show details" but only see "OK". Clicked "OK"
> No "Show details" here either

In that screenshot you're showing details, so that's why the "Show details" button doesn't show. 

"Root user": It checks for that and complains early, so you shouldn't have gotten to that point. Could you do that again and see if you can give an exact repro path?

tyler36:
6.  "Programs and Features" / Uninstalling 

Control Panel\All Control Panel Items\Programs and Features

- Icon appears fuzzy/low quality
- Size is not reported
- Uninstalling appears to remove mkcert but not DDEV from WSL.
image

rfay:
Agreed, uninstalling would never touch the WSL2 distro. That seems to me to be a violation. It doesn't remove mkcert from WSL2, but it does remove mkcert.exe from the Windows side. 

I will take a look at the logo in the uninstall. Claude code was kind enough to set that up :)

colans:
Finally got a chance to try it.  I thought I'd start without uninstalling the old installation, using the recommended path.  I get a message saying I need to install a Docker provider, but Docker is already installed in WSL; that's what I've been using.

Is it possible that it's checking the wrong WSL?  There are a couple on here, the default Ubuntu one, and a new one that the org pushed out (that I don't use, and hasn't been set up yet).

rfay:
Thanks @colans - If you're using Docker Desktop and WSL2 and doing that installation path, you need to have `docker ps` working in the distro you're trying to install to. You have to enable integration with that distro in Docker Desktop settings. Am I answering the right question?

If you can say where and when you're getting the message it will help. From the installer? When trying to run ddev?

colans:
From the installer.  Nope, not Docker Desktop, Docker inside WSL:
image
image
image

rfay:
I'm sure `docker ps` works in that distro, right? And with the default user? Does `wsl -d <distro> docker ps` work from PowerShell? 

Would you mind trying it with a new fresh distro, `wsl --install Ubuntu-24.04` for example?

rfay:
It would be lovely to have you come over to discord for easier turnaround. https://ddev.com/s/discord

rfay:
I got your same experience. Must have been a recent change :( Thanks for testing, sorry it was broken.

colans:
Thanks for confirming that I'm not crazy!  Discord & Slack are blocked on these laptops, but I'm on [Element](https://element.io/)/[Matrix](https://matrix.org/) as `@colan:matrix.org`, if that helps.

stasadev:
> I know `apt-get` can take a while sometimes, but it looks busy in my terminal. When using the DDEV installing, there's no "activity indicator" so it "feels" like it's frozen. Perhaps a message on the previous screen, "Installing WSL components may take a couple of minutes. "

Not sure if `apt-get upgrade` is necessary. It upgrades everything, feels forced, and can be slow. It was probably added to ensure all packages are up-to-date, but that's the user's responsibility, not ours.

rfay:
Fixed and signed windows installer is at https://github.com/ddev-test/ddev/releases/tag/v1.99.4 - thanks for your effort on this! It turns out I was working on automated tests for this... and they have to skip checks like this.

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1657 @@
...
+        DetailPrint "ERROR: WSL user check failed - exit code: $1, output: $0"
+        Push "Could not check WSL user. Please ensure WSL is working."
+        Call ShowErrorAndAbort
+    ${EndIf}
+    ${If} $0 == "root"
+        DetailPrint "ERROR: Default WSL user is root - this is not supported"
+        Push "The default user in your WSL2 distro is root. Please configure an ordinary default user."
+        Call ShowErrorAndAbort
+    ${EndIf}
+    DetailPrint "Non-root user detected successfully."
```
stasadev: I think there is a problem with trailing lines in the output, the check for `root` doesn't catch it:

image

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1657 @@
...
+
+        ; Check if it starts with "Ubuntu"
+        StrCpy $R4 $R3 6
+        ${If} $R4 == "Ubuntu"
+            DetailPrint "Found Ubuntu distribution: $R3"
+            ${If} $R0 != ""
+                StrCpy $R0 "$R0|"
+            ${EndIf}
+            StrCpy $R0 "$R0$R3"
+        ${EndIf}
```
stasadev: Let's check for `Flavor` instead of `DistributionName`:

image

winpkg/ddev_windows_installer.nsi
stasadev: To reproduce, create a new WSL2, but do not boot it, i.e. do not create a new user.

winpkg/ddev_windows_installer.nsi
rfay: I like it! Requires exploring registry instead of wsl -l, but excellent

stasadev: Looks like we're already accessing the registry, aren't we? 🙂

rfay:
Review TODOS:
- [x] See if we can look at "flavor" for WSL type (in registry)
- [x] Fix problem with allowing root user
- [x] Remove apt-get upgrade
- [x] Consider seeing if size can be reported in apps/uninstall dialog
- [ ] Consider showing progress during apt setup

rfay:
Thanks for all your reviews. I think I've addressed the feedback, thanks for your patience!

There's a signed version of the installer at https://github.com/ddev-test/ddev/releases

And the unsigned one is updated in the PR here.

rfay:
I note that there's something wrong with the CTA on the final page

image

stasadev:
> I note that there's something wrong with the CTA on the final page

I see this:

image

But if it is broken in some cases, then it is better to remove this Unicode character.

rfay:
Thanks. I remember it looking OK for me before. I wonder how it can sometimes be bad and other times good.

stasadev:
Not really needed, but it would be nice if the installer remembered what was installed earlier in the Registry (I think people may forget what they chose the first time), so on the second install/upgrade the selection will be the predefined:

image

rfay:
> it would be nice if the installer remembered what was installed earlier in the Registry

Nice idea!

docs/content/users/install/ddev-installation.md
```diff
@@ -180,115 +174,62 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
...
-    * WSL2 is not the same as Docker Desktop’s WSL2 engine.  
-      Using WSL2 to install and run DDEV is not the same as using Docker Desktop’s WSL2 engine, which itself runs in WSL2, but can serve applications running in both traditional Windows and inside WSL2.
+    * Custom hostnames (non `ddev.site` hostnames) are managed via the Windows hosts file, not within WSL2.  
+      DDEV attempts to manage custom hostnames via the Windows-side hosts file—usually at `C:\Windows\system32\drivers\etc\hosts`.
```
stasadev:
```suggestion
      DDEV attempts to manage custom hostnames via the Windows-side hosts file—usually at `C:\Windows\System32\drivers\etc\hosts`.
```

docs/content/users/install/ddev-installation.md
```diff
@@ -180,115 +174,62 @@ Once you’ve [installed a Docker provider](docker-installation.md), you’re re
...
+    1. **Download the Windows installer** from the [DDEV releases page](https://github.com/ddev/ddev/releases).
+    2. **Run the installer** and choose your installation type:
 
-    1. Install WSL2 with an Ubuntu distro. On a system without WSL2, run:
-        ```powershell
-        wsl --install
-        ```
+           - **Docker CE inside WSL2** (Recommended): The installer will automatically install Docker CE in your WSL2 environment. This is the fastest and most reliable option.
+           - **Docker Desktop/Rancher Desktop**: Choose this if you already have Docker Desktop or Rancher Desktop installed or prefer to use them.
+           - **Traditional Windows**: Choose this for non-WSL2 installations (requires Docker Desktop or Rancher Desktop). Note that both Docker Desktop and Rancher Desktop themselves use WSL2, but you will interact with DDEV in Git Bash, PowerShell, or Cmd on the Windows side.
```
stasadev, line 195: > Note that both Docker Desktop and Rancher Desktop themselves use WSL2

I installed Docker Desktop in a virtual machine using WSL2 and Hyper-V methods, and DDEV worked in both cases. Do we recommend using WSL2 backend all the time?

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1967 @@
...
+            /docker-desktop /distro=<name> - WSL2 with Docker Desktop$\n\
+            /rancher-desktop /distro=<name> - WSL2 with Rancher Desktop$\n\
+            /traditional                  - Traditional Windows install$\n\
+            /S                           - Silent install$\n\
+            /help or /?                  - Show this help message$\n$\n\
+            Examples:$\n\
+            installer.exe /docker-ce /distro=Ubuntu-22.04$\n\
+            installer.exe /docker-desktop /distro=Ubuntu-20.04$\n\
+            installer.exe /traditional$\n\
+            installer.exe /traditional /S"
```
stasadev, line 1785: This looks a bit scattered, but I'm not sure how to make it look better:

image

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1967 @@
...
+!define MUI_HEADERIMAGE_BITMAP "graphics\ddev-header.bmp"
+!define MUI_WELCOMEFINISHPAGE_BITMAP "graphics\ddev-wizard.bmp"
+
+!define MUI_ABORTWARNING
+
+; Function declarations - must be before page definitions
+
+; InitializeDebugLog - Open debug log file for writing
+Function InitializeDebugLog
+    StrCpy $DEBUG_LOG_PATH "$TEMP\ddev_installer_debug.log"
```
stasadev, line 70: I can't find this file `ddev_installer_debug.log` inside `C:\Windows\Temp`, should it be created all the time?

What is `$TEMP` ?

stasadev (approved):
I tested all three setups and everything worked as expected.

Then upgraded from v1.24.6 to the PR: everything still worked.

It’s probably best to recommend that users uninstall DDEV via "Add or Remove Programs" before installing the new version, since `ddev.exe` on the Windows side is not removed. This could be mentioned in the release notes.

I have a few minor nitpicks. Thank you for the great work, especially removing the need for password entry during WSL2 configuration, adding the new installer graphics, and consolidating all setups in one place.

rfay:
I got an ARM64 Windows machine to test this with.  Starting from first landing in Windows (all the time/date/etc setup) it took:
* 10 minutes to get WSL2 + DDEV (docker-ce) with the ARM64 installer set up
  * A fresh ubuntu did in fact take a longer time to install than I was used to, it would probably be nice if we could show more action during that apt-get install.
  * There was a problem: I had to `sudo systemctl start docker`, it hadn't been done. I don't know if that's something different about docker-ce arm64 setup, but I'll test it again with a clean new ubuntu.
* It took 5 more minutes (total of 15 from start) to get a Drupal 11 instance running in DDEV, fully installed with drush, respecting https.

Yay!

The installer was signed properly, and installed ARM64 executables properly.

docs/content/users/install/ddev-installation.md
rfay: WSL2 is the recommendation of Docker, and Hyper-V is the old way. I'm fine with people using both, thanks for testing both. But "choice is a sickness of the mind"

winpkg/ddev_windows_installer.nsi
rfay: When a failure happens, it reports it. But 
```
PS C:\Windows> $env:TEMP
C:\Users\randy\AppData\Local\Temp
```

winpkg/ddev_windows_installer.nsi
rfay: Hopefully people won't ever need this. But most installers don't have this information anyway.

winpkg/ddev_windows_installer.nsi
stasadev: Thank you, I see it now.

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,1967 @@
...
+        Push "Failed to create Docker check dialog"
+        Call LogPrint
+        Abort
+    ${EndIf}
+
+    ; Create different messages based on installation type
+    ${If} $INSTALL_OPTION == "traditional"
+        ${NSD_CreateLabel} 0 0 100% 60u "Docker provider is required for traditional Windows installation but was not found or is not running.$\r$\n$\r$\nPlease install and start Docker Desktop (https://www.docker.com/products/docker-desktop/) or Rancher Desktop (https://rancherdesktop.io/) before installing DDEV.$\r$\n$\r$\nYou can exit now to install Docker, or cancel this installation."
+    ${ElseIf} $INSTALL_OPTION == "wsl2-docker-desktop"
+        ${NSD_CreateLabel} 0 0 100% 60u "Docker Desktop or Rancher Desktop is required for this installation but is not accessible in WSL2.$\r$\n$\r$\nPlease install Docker Desktop or Rancher Desktop and enable WSL2 integration for the '$SELECTED_DISTRO' distro.$\r$\n$\r$\nYou can exit now to configure Docker, or cancel this installation."
```
stasadev: The installer doesn't say that Docker Desktop or Rancher Desktop should be running:

```suggestion
        ${NSD_CreateLabel} 0 0 100% 60u "Docker Desktop or Rancher Desktop is required for this installation but is not accessible in WSL2.$\r$\n$\r$\nEnsure that Docker Desktop or Rancher Desktop is installed, running, and has WSL2 integration enabled for the '$SELECTED_DISTRO' distro.$\r$\n$\r$\nThe 'docker ps' command must succeed inside WSL2 before launching this installer.$\r$\n$\r$\nYou can exit now to configure Docker, or cancel this installation."
```

rfay:
> It’s probably best to recommend that users uninstall DDEV via "Add or Remove Programs" before installing the new version, since ddev.exe on the Windows side is not removed. This could be mentioned in the release notes.

I'm not convinced that it's important to remove the ddev.exe, and it adds complexity to bother people with this. It could (maybe?) also mess with their mkcert config to uninstall.

rfay:
The installer now remembers the install type and distro that were previously chosen and uses those as default. Removed the checkmark from the final page. Reworked the distro radio boxes as dynamic instead of stupid.  Probably needs a little more casual manual testing.

rfay:
Broke the apt-get install into 4 parts so people get feedback that something is happening. Updated signed release in https://github.com/ddev-test/ddev/releases/tag/v1.99.8 (or just see latest release there)

winpkg/ddev_windows_installer.nsi
```diff
@@ -0,0 +1,2077 @@
...
+        Push "ERROR: apt-get update failed - exit code: $1, output: $0"
+        Call LogPrint
+        Push "Failed to apt-get update. Error: $0"
+        Call ShowErrorAndAbort
+    ${EndIf}
+
+    ; Install linux packages
+    Push "WSL($SELECTED_DISTRO): Installing required linux packages..."
+    Call LogPrint
+    Push "Please be patient - installing required packages..."
```
stasadev: To match the previous line in the output:

```suggestion
    Push "Please be patient - installing required linux packages..."
```

PR #7405, opened 2025-06-26
test: use bats-core/bats-core homebrew tap instead of kaos

rfay:
## The Issue

Use the official bats-core/bats-core brew tap instead of the kaos one.


rfay:
Added link in docs.

PR #7415, opened 2025-06-30
test: Ignore existing golangci-lint/revive package complaints

rfay:
## The Issue

A golangci-lint/revive upgrade brings new warnings on something it's not worth changing (too disruptive). 

https://github.com/ddev/ddev/actions/runs/15973505093/job/45049930978

> Error: pkg/config/remoteconfig/types/remote_config_storage.go:1:9: var-naming: avoid meaningless package names (revive)
  package types

This new check was added in 
* https://github.com/mgechev/revive/pull/1312

Coaching on how to ignore it is in
* https://github.com/mgechev/revive/pull/1312#discussion_r2044165709

Ignore this on existing "violations", but the next time we create something like this we'll get warned.

### DIFF
--- a/pkg/config/remoteconfig/state.go
+++ b/pkg/config/remoteconfig/state.go
@@ -1,4 +1,4 @@
-package remoteconfig
+package remoteconfig //nolint:revive
 
 import (
 	"time"
--- a/pkg/config/remoteconfig/types/remote_config_storage.go
+++ b/pkg/config/remoteconfig/types/remote_config_storage.go
@@ -1,4 +1,4 @@
-package types
+package types //nolint:revive
 
 import (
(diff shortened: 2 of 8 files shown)


rfay:
Passes static validation now, https://github.com/ddev/ddev/actions/runs/15974362646/job/45052834420

stasadev:
Another way to do this in `.golangci.yml` (from https://golangci-lint.run/usage/linters/#revive):

```yaml
linters:
  settings:
    revive:
      rules:
        - name: var-naming
          arguments:
            - []
            - []
            - - skip-package-name-checks: true
```

But not sure which method is better.

rfay:
I figured that explicitly ignoring the current complaints was better than ignoring new complaints in the future.

PR #7416, opened 2025-06-30
ci: use `skip-package-name-checks`, pin `golangci-lint` to v2.2.1, remove `only-new-issues`, for #7415

stasadev:
## The Issue

There are some more warnings related to:

- #7415

And @rfay pointed out that behavior added in the PR below is confusing:

- #6954

- Uses `skip-package-name-checks` to disable this check. It often reports inconsistent results, likely due to caching issues (`golangci-lint cache clean && make golangci-lint`)
- Reverts changes made in #7415 as not needed.
- Removes `only-new-issues`, so full lint checks run on PRs.
- Pins for `golangci-lint` to v2.2.1, which avoids future breakage from upstream changes. This adds some maintenance since we need to bump it manually, but it prevents unexpected linter failures.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -50,9 +50,7 @@ jobs:
       - name: Lint Go Files
         uses: golangci/golangci-lint-action@v8
         with:
-          version: latest
-          # Check code only for the changed lines for PRs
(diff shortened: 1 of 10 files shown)


stasadev:
Static Validation doesn't complain anymore https://github.com/ddev/ddev/actions/runs/15978385040/job/45066665574?pr=7416#step:5:32

PR #3620, opened 2022-02-22
Change docker-compose yaml to insert networks stanza, fixes #3587

rfay:
## The Problem/Issue/Bug:

* #3587

Because the requirement of the `networks` stanza in ddev v1.19 is pretty intrusive, requiring existing `docker-compose.*.yaml` all to be edited, this PR unmarshalls in the fully normalized yaml text into a golang map, replaces the `networks` array, and then marshalls the yaml back out to .ddev/.ddev-docker-compose-full.yaml.

## Manual Testing Instructions:

- [x] Try it out with recipes from ddev-contrib that haven't had the `networks` stanza added
- [x] Try it out with recipes that *do* have the `networks` stanza.

## TODO
- [x] Remove the docs that ask for the `networks` stanza

## Automated Testing Overview:

I'm just hoping that all existing pass. 
- [x] Add a test with a 3rd-party service that does not have `networks`

## Release/Deployment notes:

- [ ] Release notes can remove the worry about the networks stanza.

<a href="https://gitpod.io/#https://github.com/drud/ddev/pull/3620"></a>

PR #7422, opened 2025-07-02
refactor: use `compose-spec/compose-go/v2` for `fixupComposeYaml`

stasadev:
## The Issue

It's hard to edit `.ddev/.ddev-docker-compose-full.yaml` by doing manual YAML parsing.

Uses `github.com/compose-spec/compose-go/v2` (adds ~2 MB to the `ddev` binary size).

Changes the `app.ComposeYaml`:
```diff
-ComposeYaml map[string]interface{} `yaml:"-"`
+ComposeYaml *composeTypes.Project  `yaml:"-"`
```

Bumps `DockerComposeVersionConstraint` from `>= 2.5.1` to `>= 2.24.3` (which is used in Docker Desktop with Docker Engine 25.0, see https://docs.docker.com/desktop/release-notes/) (and rename to `ComposeVersionConstraint`)
Docker 25.0 is our current recommended minimum.

## Release/Deployment Notes

It's important to keep `github.com/compose-spec/compose-go/v2` library in sync with what is used in `docker-compose`.


stasadev:
@rfay, from the git history I see you left `compose-spec/compose-go` comments in the imports here:

- #3620

Looks like I'm making a second attempt to implement it.

rfay:
Ambitious!

rfay:
Pretty amazing that you seem to have gotten that all the way to clean with a very modest set of line changes.

stasadev:
And it will be much better after the next refactor:

```diff
-ComposeYaml map[string]interface{} `yaml:"-"`
+ComposeYaml *composeTypes.Project  `yaml:"-"`
```

We parse `app.ComposeYaml` manually in many places.

stasadev:
I see potential for further refactoring, but it's better to leave that for other PRs. Ready for review.

rfay:
## Claude Code Summary of Changes

This PR modernizes how DDEV handles Docker Compose YAML files by switching from manual parsing to using the official Docker Compose library.

### Key Changes:

**1. New Library Integration**
- Added `github.com/compose-spec/compose-go/v2` - the official Docker Compose Go library
- This is the same library Docker Compose itself uses internally

**2. Type System Upgrade**
- **Before**: `ComposeYaml` was a generic `map[string]interface{}` - basically unstructured data
- **After**: `ComposeYaml` is now `*composeTypes.Project` - a properly structured type with fields for services, networks, volumes, etc.

**3. Better YAML Processing** (`pkg/ddevapp/compose_yaml.go:97-137`)
- New `createComposeProject()` function uses the official library to parse YAML
- Automatically handles all the complex Docker Compose specification details
- Ensures all required fields (Networks, Services, Volumes) are properly initialized

**4. Updated Docker Compose Version Requirement**
- **Before**: Required Docker Compose >= 2.5.1
- **After**: Now requires >= 2.24.3 (aligns with current Docker Desktop)

### Why This Matters:

**For Users**: More reliable handling of complex Docker Compose configurations, especially when editing `.ddev/.ddev-docker-compose-full.yaml`

**For Developers**: 
- Type safety - the compiler catches errors that would previously only show up at runtime
- Better IntelliSense/autocomplete when working with compose data
- Less manual YAML parsing code to maintain

**Trade-off**: The binary size increases by ~2MB due to the additional library, but this provides much more robust YAML handling.

This is essentially DDEV saying "instead of reinventing the wheel for Docker Compose parsing, let's use the official tools that Docker itself uses."

pkg/ddevapp/compose_yaml.go
```diff
@@ -92,72 +93,99 @@ func (app *DdevApp) WriteDockerComposeYAML() error {
 	return nil
 }
 
+// createComposeProject creates a compose project from a string
+func createComposeProject(yamlStr string) (*composeTypes.Project, error) {
+	project, err := composeLoader.LoadWithContext(
```
rfay, line 98: So great!

pkg/ddevapp/compose_yaml.go
```diff
@@ -92,72 +93,99 @@ func (app *DdevApp) WriteDockerComposeYAML() error {
...
+	// Initialize Networks, Services, and Volumes to empty maps if nil
+	if project.Networks == nil {
+		project.Networks = composeTypes.Networks{}
+	}
+	if project.Services == nil {
+		project.Services = composeTypes.Services{}
+	}
+	if project.Volumes == nil {
+		project.Volumes = composeTypes.Volumes{}
+	}
```
rfay, line 119: It would be weird if any of these things happened, true?

pkg/ddevapp/ddevapp.go
```diff
@@ -81,70 +81,70 @@ type WebExtraDaemon struct {
...
+	WebExtraExposedPorts      []WebExposedPort      `yaml:"web_extra_exposed_ports,omitempty"`
+	WebExtraDaemons           []WebExtraDaemon      `yaml:"web_extra_daemons,omitempty"`
+	OverrideConfig            bool                  `yaml:"override_config,omitempty"`
+	DisableUploadDirsWarning  bool                  `yaml:"disable_upload_dirs_warning,omitempty"`
+	DdevVersionConstraint     string                `yaml:"ddev_version_constraint,omitempty"`
+	XHGuiHTTPSPort            string                `yaml:"xhgui_https_port,omitempty"`
+	XHGuiHTTPPort             string                `yaml:"xhgui_http_port,omitempty"`
+	HostXHGuiPort             string                `yaml:"host_xhgui_port,omitempty"`
+	XHProfMode                types.XHProfMode      `yaml:"xhprof_mode,omitempty"`
+	ComposeYaml               *composeTypes.Project `yaml:"-"`
```
rfay, line 144: Yay!

pkg/ddevapp/ddevapp.go
```diff
@@ -3432,34 +3425,10 @@ func (app *DdevApp) StartAppIfNotRunning() error {
 	return err
 }
 
-// CheckAddonIncompatibilities looks for problems with docker-compose.*.yaml 3rd-party services
```
rfay, line 3435: I guess this has all been taken care of because we used compose to load?

pkg/ddevapp/traefik.go
```diff
@@ -31,39 +31,38 @@ type TraefikRouting struct {
 // VIRTUAL_HOST and HTTP(S)_EXPOSE environment variables to set up routing
 // for the project
 func detectAppRouting(app *DdevApp) ([]TraefikRouting, error) {
-	// app.ComposeYaml["services"];
 	var table []TraefikRouting
-	if services, ok := app.ComposeYaml["services"]; ok {
-		for serviceName, s := range services.(map[string]interface{}) {
-			service := s.(map[string]interface{})
-			if env, ok := service["environment"].(map[string]interface{}); ok {
-				var virtualHost string
```
rfay, line 40: So great to lose all this!

rfay (approved):
Well, it's beautiful, what can I say! Congratulations! Thank you!

I opened https://github.com/ddev/ddev/issues/7472, which is certainly not the fault of this PR. Stepping through your lovely compose PR I note that the project is first created in cmd/pull.go's init(). Yuck to inits! Then it happens again in push.go's init(). Then it finally happens in root.go's init(), snapshot_restore.go, start.go, ddevapp.go's Start() (probably the first time it was actually needed).

rfay:
Rebased to get Windows test fixup.

pkg/ddevapp/compose_yaml.go
stasadev: Yes, but to avoid some random panic, we need to have these.

pkg/ddevapp/ddevapp.go
stasadev: Yes, and it could be deleted even earlier, the `networks` were added automatically anyway.

PR #7434, opened 2025-07-07
refactor: centralize table style configuration

stasadev:
## The Issue

We have this code block used in several places:

```go
style = GetTableStyle("default")
style.Options.SeparateRows = false
style.Options.SeparateFooter = false
style.Options.SeparateColumns = false
style.Options.SeparateHeader = false
style.Options.DrawBorder = false
```

But this configuration is already available in `styles.SetGlobalTableStyle()`. We don't need to copy-paste it everywhere.

Adds `useSimpleFormatting` flag to `styles.SetGlobalTableStyle()`.

Should have the same simple formatting *without* borders:
```
ddev version
ddev mutagen version
```
Should have the same formatting *with* borders:
```
ddev list
```

### DIFF
--- a/cmd/ddev/cmd/addon-list.go
+++ b/cmd/ddev/cmd/addon-list.go
@@ -59,14 +59,14 @@ ddev add-on list --installed --project my-project
 	},
 }
 
-// ListInstalledAddons() show the add-ons that have a manifest file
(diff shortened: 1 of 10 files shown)


rfay (approved):
Thanks!

PR #7436, opened 2025-07-08
docs: move composer usage tip to composer_root section

schliesser:
## The Issue

None

The tip is more related to the composer_root setting than to the composer_version. Therefore I moved the tip to the other section.

https://ddev--7436.org.readthedocs.build/en/7436/users/configuration/config/#composer_root

### DIFF
--- a/docs/content/users/configuration/config.md
+++ b/docs/content/users/configuration/config.md
@@ -86,6 +86,9 @@ The relative path, from the project root, to the directory containing `composer.
 | -- | -- | --
 | :octicons-file-directory-16: project | &zwnj; | &zwnj;
 
+!!!tip "Using `vendor/bin/composer`"
+    See the example in [Composer from `vendor/bin/composer`](../usage/developer-tools.md#composer-from-vendorbincomposer).
+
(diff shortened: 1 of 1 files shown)


rfay:
I guess because the tip mentions COMPOSER_ROOT in  the linked-to doc, https://ddev--7436.org.readthedocs.build/en/7436/users/usage/developer-tools/#composer-from-vendorbincomposer

I agree that maybe composer version is the place to put the link? It's definitely an odd one.

schliesser:
Okay, I see, the new version has different text than the current released one.

As the comment in the tip refers to "composer_root", I thought it was misplaced.
image

The current preview doesn't mention it at all.
image

stasadev:
Yes, it's changed in the latest version of the docs https://ddev.readthedocs.io/en/latest/users/configuration/config/#composer_version

And there is a link back to `composer_root` here https://ddev.readthedocs.io/en/latest/users/usage/developer-tools/#composer-from-vendorbincomposer

So, this PR can be closed?

schliesser:
I'm closing it.

PR #7448, opened 2025-07-11
fix: avoid mutating input slice in `cmd.prettyCmd()`

stasadev:
## The Issue

I noticed a flaw in `cmd.prettyCmd()`:

https://go.dev/play/p/EVaaARj1uYn

```
cmd := []string{"bash", "-c", `echo "test"`}
fmt.Printf("original cmd: %v\n", cmd)            # original cmd: [bash -c echo "test"]
fmt.Printf("  pretty cmd: %v\n", prettyCmd(cmd)) #   pretty cmd: bash -c "echo "test""
fmt.Printf("original cmd: %v\n", cmd)            # original cmd: [bash -c "echo "test""] - bad
```

Makes a copy, escapes double quotes.

https://go.dev/play/p/ZDcMH1TgqE1

```
cmd := []string{"bash", "-c", `echo "test"`}
fmt.Printf("original cmd: %v\n", cmd)            # original cmd: [bash -c echo "test"]
fmt.Printf("  pretty cmd: %v\n", prettyCmd(cmd)) #   pretty cmd: bash -c "echo \"test\""
fmt.Printf("original cmd: %v\n", cmd)            # original cmd: [bash -c echo "test"] - good
```

### DIFF
--- a/cmd/ddev/cmd/utils.go
+++ b/cmd/ddev/cmd/utils.go
@@ -148,10 +148,13 @@ func GetUnknownFlags(cmd *cobra.Command) (map[string]string, error) {
 
 // prettyCmd formats the Docker command to be more readable.
 func prettyCmd(cmd []string) string {
+	out := make([]string, len(cmd))
 	for i, arg := range cmd {
 		if strings.Contains(arg, " ") {
-			cmd[i] = `"` + arg + `"`
+			out[i] = `"` + strings.ReplaceAll(arg, `"`, `\"`) + `"`
+		} else {
+			out[i] = arg
 		}
 	}
-	return strings.Join(cmd, " ")
+	return strings.Join(out, " ")
 }


rfay (commented):
Do I remember that this class of problem was fixed up in latest golang?

stasadev:
I'm testing it using the latest go1.24.5.

It was a rather random discovery, I was trying to debug `ddev auth ssh` and used `cmd.prettyCmd()` in several places and then noticed this mutation.

It doesn't affect our current code, since it's only used in the final error output in `ddev auth ssh` (the original input isn't used afterward) and in `ddev debug rebuild` (where the command args don't have spaces, so they are not modified), but it should be fixed.

rfay:
Thanks. I guess I was thinking of this change in go 1.22, https://go.dev/blog/loopvar-preview

rfay (approved):
Thanks! The go playground approach to demo was really nice too.

PR #7458, opened 2025-07-16
fix: temporarily allow write to `/etc/mysql/conf.d/*` for `db` container restart, fixes #7457

### LINKED ISSUE #7457: Providing extra config via .ddev/mysql/*.cnf breaks db server restart
### Preliminary checklist

_No response_

### Expected Behavior

I can use `docker restart ddev-project-db` at all times without having to rebuild (force-recreate) the whole project.
And this regardless of providing extra db config via `.ddev/mysql/*.cnf`.

### Actual Behavior

When providing extra db config via `.ddev/mysql/*.cnf` and I restart the db container with e.g.  `docker restart ddev-project-db` I get the error:
```
+ '[' /mnt/ddev_config/mysql/cando_cx.cnf '!=' '/mnt/ddev_config/mysql/*.cnf' ']'
2025-07-16T18:41:18.994036380Z + cp /mnt/ddev_config/mysql/my_custom_config.cnf /etc/mysql/conf.d
2025-07-16T18:41:18.996947404Z cp: cannot create regular file '/etc/mysql/conf.d/my_custom_config.cnf': Permission denied
```

Reason for this is that the current `containers/ddev-dbserver/files/docker-entrypoint.sh` modifies the permission of the initially copied custom config files to make them readonly.
On a restart these permissions still apply and hence the copy fails.

### Steps To Reproduce

1. Add a file `.ddev/mysql/my_custom_config.cnf`
2. Start project as usual (`ddev start`)
3. Restart ddev db container using docker `docker restart ddev-{PROJECT}-db`

### Anything else?

_No response_

das-peter:
## The Issue

- #7457

Makes sure that the custom config files are writeable before attempting to write them on startup.

Before:

```
$ ddev stop
$ touch .ddev/mysql/test.cnf
$ ddev start

$ ddev exec -s db ls -la /etc/mysql/conf.d 
total 12
drwxrwxrwx 1 root root  4096 Jul 17 18:59 .
drwxr-xr-x 1 root root  4096 Jul 16 19:58 ..
-rwxrwxrwx 1 root root     0 Jul 16 19:58 .gitmanaged
-r--r--r-- 1 stas mysql    0 Jul 17 18:59 test.cnf

$ ddev debug rebuild -s db --cache
...
Troubleshoot this with these commands:

  - ddev logs -s db
...
$ ddev logs -s db
+ cp /mnt/ddev_config/mysql/test.cnf /etc/mysql/conf.d
cp: cannot create regular file '/etc/mysql/conf.d/test.cnf': Permission denied
```

After:

```
$ ddev stop
$ touch .ddev/mysql/test.cnf

$ cat <<'EOF' > .ddev/docker-compose.db-entrypoint.yaml
services:
  db:
    entrypoint:
      - sh
      - -c
      - |
        cp /docker-entrypoint.sh ~/docker-entrypoint.sh
        sed -i '81i chmod -f -R u+w /etc/mysql/conf.d/* || true' ~/docker-entrypoint.sh
        exec ~/docker-entrypoint.sh
EOF

$ ddev start

$ ddev exec -s db ls -la /etc/mysql/conf.d
total 12
drwxrwxrwx 1 root root  4096 Jul 17 19:01 .
drwxr-xr-x 1 root root  4096 Jul 16 19:58 ..
-rwxrwxrwx 1 root root     0 Jul 16 19:58 .gitmanaged
-r--r--r-- 1 stas mysql    0 Jul 17 19:01 test.cnf

$ ddev debug rebuild -s db --cache

$ ddev exec -s db ls -la /etc/mysql/conf.d 
total 12
drwxrwxrwx 1 root root  4096 Jul 17 19:01 .
drwxr-xr-x 1 root root  4096 Jul 16 19:58 ..
-rwxrwxrwx 1 root root     0 Jul 16 19:58 .gitmanaged
-r--r--r-- 1 stas mysql    0 Jul 17 19:02 test.cnf
```

### DIFF
--- a/containers/ddev-dbserver/files/docker-entrypoint.sh
+++ b/containers/ddev-dbserver/files/docker-entrypoint.sh
@@ -78,6 +78,9 @@ rm -f /tmp/raw_mysql_version.txt
 
 # If we have extra cnf files from user, copy them to where they go.
 if [ -d /mnt/ddev_config/mysql ] && [ "$(echo /mnt/ddev_config/mysql/*.cnf)" != "/mnt/ddev_config/mysql/*.cnf" ] ; then
+  # Ensure re-copying of custom files works on restart. 
+  # Has to ignore errors that arise if no custom files have been copied yet.
+  chmod -f -R u+w /etc/mysql/conf.d/* || true
   cp /mnt/ddev_config/mysql/*.cnf /etc/mysql/conf.d
   # Ignore errors on files such as .gitmanaged
   chmod -f -R ugo-w /etc/mysql/conf.d/*
--- a/pkg/versionconstants/versionconstants.go
+++ b/pkg/versionconstants/versionconstants.go
@@ -17,7 +17,7 @@ var WebTag = "20250612_stasadev_rebuild_images" // Note that this can be overrid
 var DBImg = "ddev/ddev-dbserver"
 
 // BaseDBTag is the main tag, DBTag is constructed from it
-var BaseDBTag = "20250716_das-peter_fix-extra-db-config-breaks-container-restart"
+var BaseDBTag = "20250716_das-peter_cnf"
 
 // TraefikRouterImage is image for router
 var TraefikRouterImage = "ddev/ddev-traefik-router"
--- a/pkg/versionconstants/versionconstants.go
+++ b/pkg/versionconstants/versionconstants.go
@@ -17,7 +17,7 @@ var WebTag = "20250612_stasadev_rebuild_images" // Note that this can be overrid
 var DBImg = "ddev/ddev-dbserver"
 
 // BaseDBTag is the main tag, DBTag is constructed from it
-var BaseDBTag = "20250213_rfay_mariadb_11.8"
+var BaseDBTag = "20250716_das-peter_fix-extra-db-config-breaks-container-restart"
 
 // TraefikRouterImage is image for router
 var TraefikRouterImage = "ddev/ddev-traefik-router"


rfay (changes requested):
Thanks for the contribution!

Did you try this out using a `.ddev/db-build/Dockerfile.chmod` that adds the docker-entrypoint.sh? If not, please do.

It would surprise me if it works to solve your problem.

This PR is not a workable PR at this point because it requires an image push. Try it as a Dockerfile.chmod first.

das-peter:
Hi @rfay,

thanks for getting back.

> Did you try this out using a .ddev/db-build/Dockerfile.chmod that adds the docker-entrypoint.sh? If not, please do.

Just did and found an issue on initial start if there are no custom config files yet.
And an embarrassing copy paste artifact in the initial chmod command. 😞 

> It would surprise me if it works to solve your problem.

Be surprised 😄 Container restart works now like a charm in the same setup.

However, since I've realize that the db image is always built I'll most likely handle that with a custom Dockerfile indeed.
Then whole effort goes into providing a password reset feature for the db-server.
Reason for it is that we use already mariabackup to fetch db data from development servers - but these use higher security passwords. So when restoring these "snapshots" we need to reset the passwords to what ddev uses.
Doing so requires either db-restarts or an extra execution of  "mysqld" with the `--init-file=` parameter  just after the restore.
I think I inject the necessary behavior via custom Dokerfile into `docker-entrypoint.sh` - unless that's something interesting for ddev itself. If so I'd try to work that into an PR.

stasadev (approved):
I updated manual testing instructions.

I confirm that this fix works and can be merged.

containers/ddev-dbserver/files/docker-entrypoint.sh
```diff
@@ -78,6 +78,9 @@ rm -f /tmp/raw_mysql_version.txt
 
 # If we have extra cnf files from user, copy them to where they go.
 if [ -d /mnt/ddev_config/mysql ] && [ "$(echo /mnt/ddev_config/mysql/*.cnf)" != "/mnt/ddev_config/mysql/*.cnf" ] ; then
+  # Ensure re-copying of custom files works on restart. 
+  # Has to ignore errors that arise if no custom files have been copied yet.
+  chmod -f -R ugo+w /etc/mysql/conf.d/* || true
```
stasadev: @rfay pointed out that `u+w` will be enough:

```suggestion
  chmod -f -R u+w /etc/mysql/conf.d/* || true
```

das-peter: Thank you @rfay & @stasadev  for the feedback.
Good suggestion indeed - keep handling at a minimum.
I've updated the forked branch accordingly.
I couldn't find anything in the contrib doc defining whether multiple commits in a PR branch are encouraged or not. I guess it might depend if the PR merges are squashed.
Anyhow - original commit updated with the suggestion.

stasadev:
Okay, I'll push new images for this.

das-peter:
Absolutely agree regarding the urgency of this case - no rush at all!
We are indeed doing some things that are somewhat outside of the normal ddev tooling.

containers/ddev-dbserver/files/docker-entrypoint.sh
stasadev: Either way is fine - we squash everything.

PR #7461, opened 2025-07-18
fix: Windows installer shouldn't be so aggressive in removing CAROOT

rfay:
## The Issue

We have some installer test failures that are a result of being aggressive about removing CAROOT and when the CAROOT gets populated in a freshly started distro

It's my belief that this would not affect real users, or at least not affect real users unless they uninstalled DDEV and then reinstalled, etc, with some combination of not restarting the distro, etc.

- Don't be so aggressive in removing CAROOT during uninstall from the Windows side. I doubt this even matters to a normal user.
- Replace a couple of DetailPrint with LogPrint. 

- Run through install
- Review test output

It won't surprise me if at least the TestDesktop distro has to be restarted

### DIFF
--- a/winpkg/ddev_windows_installer.nsi
+++ b/winpkg/ddev_windows_installer.nsi
@@ -1586,11 +1586,11 @@ Function InstallWSL2Common
     ; Mark installation as complete for external monitoring
     Push "Marking installation as complete..."
     Call LogPrint
-    DetailPrint "Marking installation as complete..."
     nsExec::ExecToStack 'wsl -d $SELECTED_DISTRO bash -c "echo \"COMPLETED: Installation completed successfully\" >> /tmp/ddev_installation_status.txt"'
     Pop $1
     Pop $2
-    DetailPrint "DDEV installation completed successfully"
+    Push "DDEV installation completed successfully"
+    Call LogPrint
 
     ; Clean up temp directory
     Push "Cleaning up temporary files..."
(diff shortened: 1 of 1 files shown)


rfay:
The installer tests ran fine on tb-win11-10, but skipped Docker Desktop. Docker Desktop has a tendency to lose its integration with a distro, and then suddenly discover that, then you have to click a button. I don't think that's really our problem to solve, and the current skip behavior is OK in general, as long as we get it to run once in a while. Mostly DockerCE and Docker Desktop are the same in the installer and share most code.

Edit: Retrying successfully a couple of times on tb-win11-10. After that I'll disable 10 and try it on tb-win11-06

Edit: From time to time we do see `Installer failed: exit status 0xc0000005` and that's a complete NSIS crash. I haven't seen it in manual work, but have seen it a couple of times in tests.

rfay:
This has now run a few times successfully on both runners.

stasadev (approved):
Looks good to me!

rfay:
I did this:
* Completely destroyed WSL2 and removed all distros, rebooted
* `wsl --install` to get it installed again
* Created DDEV distro from Ubuntu and installed using installer, all worked including launching with TLS
* Created a Docker Desktop-using distro and installed using installer, all worked with trivial site
* Created traditional WIndows install. All worked
* Did an uninstall
* Deleted DockerCE distro and recreated
* Did DockerCE installation using installer, all worked including trivial project with TLS

PR #7463, opened 2025-07-18
docs: add `.wslconfig` example for mirrored networking mode in WSL2

stasadev:
## The Issue

It may be unclear what needs to be done for mirrored networking mode in WSL2.

Adds an example.

https://ddev--7463.org.readthedocs.build/en/7463/#system-requirements-windows-wsl2


rfay (approved):
Thanks!

PR #7467, opened 2025-07-19
docs: Add CLAUDE.md

rfay:
## The Issue

You can initialize Claude Code with a CLAUDE.md that give basic instructions about how to build and run, do static checking, etc. 

An initial set of prompts.


rfay:
We can improve this as time goes on if we end up sticking with Claude.

PR #7469, opened 2025-07-21
fix: Allow .DS_Store when doing ddev composer create-project [skip ci]

rfay:
## The Issue

Drupal Slack: https://drupal.slack.com/archives/C5TQRQZRR/p1753093589691309

Somebody had a .DS_Store (probably network drive?) on macOS

We don't need to insist on its removal

Add .DS_Store to allowed files


stasadev (approved):
The related Drupal report says:

```
'/Users/Workspace/drupalcms/web/.DS_Store' is not allowed to be present
```

So the open directory was `web`, and this PR only adds an exception to `/Users/Workspace/drupalcms/.DS_Store`

In any case, this is an edge case, so I think it's good.

rfay:
Oh, interesting, thanks for noticing that.

PR #7474, opened 2025-07-22
fix: Update obsolete WSL2 install scripts to reflect new ddev-wsl2 and not needing ddev.exe, fixes #7464 [skip ci]

### LINKED ISSUE #7464: Script cleanup: WSL2 manual scripts should be updated
### Expected Behavior

(old) PS scripts should work

### Actual Behavior

In DDEV v1.24.7, ddev.exe does not need to be installed on the Windows side (and shouldn't be)

Instead, the scripts should just set CAROOT and WSLENV and that's about it besides the main install

### Steps To Reproduce

_No response_

### Anything else?

_No response_

rfay:
## The Issue

- #7464

The Windows installer in v1.24.7+ obsoletes the installer scripts, but they remain simple examples to follow for manual installation of DDEV.

* Add warning at top (use installer)
* Remove installation of windows-side ddev
* Rework mkcert install to use wsl2-side mkcert.exe

Try running using the instructions at the top (instructions have been removed from docs)

## Automated Testing Overview

No testing

### DIFF
--- a/scripts/install_ddev_wsl2_docker_inside.ps1
+++ b/scripts/install_ddev_wsl2_docker_inside.ps1
@@ -30,6 +30,12 @@ if (wsl bash -c "test -d /mnt/wsl/docker-desktop >/dev/null 2>&1" ) {
 }
 $ErrorActionPreference = "Stop"
 
+# Remove old Windows ddev.exe if it exists using uninstaller
+if (Test-Path "$env:PROGRAMFILES\DDEV\ddev_uninstall.exe") {
+    Write-Host "Removing old Windows ddev.exe installation"
+    Start-Process "$env:PROGRAMFILES\DDEV\ddev_uninstall.exe" -ArgumentList "/SILENT" -Wait
+}
+
 wsl -u root bash -c "apt-get remove -y -qq docker docker-engine docker.io containerd runc >/dev/null 2>&1"
 wsl -u root apt-get update
 wsl -u root apt-get install -y ca-certificates curl gnupg lsb-release
--- a/scripts/install_ddev_wsl2_docker_inside.ps1
+++ b/scripts/install_ddev_wsl2_docker_inside.ps1
@@ -51,7 +51,7 @@ wsl bash -c 'sudo usermod -aG docker $USER'
 
 wsl mkcert.exe -install
 $env:CAROOT = & wsl mkcert.exe -CAROOT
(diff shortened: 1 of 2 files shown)


rfay:
I think this is probably "good enough" for an obsolete set of scripts. 

These are now run as ordinary user.

I found the Docker Desktop integration flaky, and it would sometimes exit and fail, but I don't know anything to do about that. We recommend docker-ce.

scripts/install_ddev_wsl2_docker_inside.ps1
```diff
@@ -117,18 +46,22 @@ wsl -u root -e bash -c 'echo deb [arch=$(dpkg --print-architecture) signed-by=/e
...
 wsl -u root -e bash -c 'echo deb [signed-by=/etc/apt/keyrings/ddev.gpg] https://pkg.ddev.com/apt/ \* \* > /etc/apt/sources.list.d/ddev.list'
 wsl -u root -e bash -c "apt-get update && apt-get install -y docker-ce docker-ce-cli containerd.io wslu"
-wsl -u root -e bash -c "apt-get install -y --no-install-recommends ddev"
-wsl -u root -e bash -c "apt-get upgrade -y >/dev/null"
+wsl -u root -e bash -c "apt-get install -y --no-install-recommends ddev ddev-wsl2"
 wsl bash -c 'sudo usermod -aG docker $USER'
 
+wsl mkcert.exe -install
+$env:CAROOT = & wsl mkcert.exe -CAROOT
+setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV
```
stasadev:
```suggestion
setx CAROOT $env:CAROOT; If ($Env:WSLENV -notlike "*CAROOT/up:*") { $env:WSLENV="CAROOT/up:$env:WSLENV"; setx WSLENV $Env:WSLENV }
```

rfay: I don't understand how this could have worked at all...

stasadev (approved):
I confirm that Docker Desktop is flaky, it crashed wsl for me on the last `docker ps` script step, but then worked fine.

I tested both scripts, mkcert works fine.

Looks good to me.

rfay:
Thanks for the review and for catching that curly brace! I'll pull and we'll see what the future holds. Hopefully this is irrelevant.

PR #7479, opened 2025-07-23
test: add a no-interaction flag to the install command in ibexa bats file

rpkoller:
## The Issue

the ibexa tests were failing according to @rfay . turned out that the `ddev exec console ibexa:install` now requires the manual entry of an admin password. it was probably added recently but according to https://doc.ibexa.co/en/latest/getting_started/install_ibexa_dxp/#create-a-database it is possible to override that with a no-interaction flag in the context of for example CI process like in this case. 

- #<issue number>

### DIFF
--- a/docs/tests/ibexa.bats
+++ b/docs/tests/ibexa.bats
@@ -24,8 +24,8 @@ teardown() {
   # ddev composer create-project ibexa/oss-skeleton
   run ddev composer create-project ibexa/oss-skeleton
   assert_success
-  # ddev exec console ibexa:install
-  run ddev exec console ibexa:install
+  # ddev exec console ibexa:install --no-interaction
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for chasing this down!

rfay (approved):
Thanks! If it passes I'm happy.

Should this change also be made to the quickstart?

rpkoller:
i probably wouldn't change it in the quickstart. the idea of the quickstart is that people walk manually through the process of installing a cms  and learn then there that the install step for ibexa requires the change of the admin password. and https://doc.ibexa.co/en/latest/getting_started/install_ibexa_dxp/#add-entropy-to-improve-cryptographic-randomness only recommends it in the context of automatic deployment and ci pipelines. so i would vote against adding it to the quickstart as well.

Issue #5215, opened 2023-07-31
Utilize terminus rsync plugin for pantheon.


TonyTheFerg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Pantheon provides a [plugin](https://github.com/pantheon-systems/terminus-rsync-plugin) for rsync integration. Maybe DDEV could use it. 

I too was having issues with pushing files to pantheon as seen in [this issue](https://github.com/ddev/ddev/issues/3905), and @rfay suggested opening an issue about using the plugin as a step forward.

### Describe your solution

Maybe add the plugin to DDEV, and use the commands available when the plugin exists.

DigitalFrontiersMedia:
I started looking at this while completing first experiments for #4760.  I see 
```
### From user Dockerfile /Users/stephen/Projects/dfm-rok2/.ddev/web-build/Dockerfile.experiment:
## Install terminus rsync plugin
RUN terminus self:plugin:install pantheon-systems/terminus-rsync-plugin
```
added to .webimageBuild/Dockerfile.  And I see it being installed (takes a little while) and I've seen a message about "already installed" on later attempts.  But when my WIP Pantheon provider runs the file push, I get "Command "rsync" is not defined."  When I go into the web container, indeed, terminus rsync is not listed.  So I'm open to any ideas on where I might be going wrong.  I also experimented a bit with the guidance at https://ddev.com/blog/ddev-docker-image-maintenance/ but hit a block at pushing branch image to hub.docker.com with equivalent changes to ddev-webserver/Dockerfile.

rfay:
Thanks for exploring this!

Your strategy of using a Dockerfile.* to start with is a good one, rather than building the image first.

I would imagine the problem is that when you run `RUN terminus self:plugin:install pantheon-systems/terminus-rsync-plugin` there, it's installing it for the root user and I'll bet it wants to install it per user. 

You can sort that out by manually running `terminus self:plugin:install pantheon-systems/terminus-rsync-plugin` and seeing if it works, and figuring out where it gets installed, and if there are options on where it can be installed (can it be global?)

DigitalFrontiersMedia:
It is, indeed, added for the root user.  It's added as a composer file at ~/.terminus/plugins-3.x/composer.json and then installed, I guess.  There aren't any options to install it anywhere other than in the user's home directory that I could see (https://docs.pantheon.io/terminus/commands/self-plugin-install).

[...]

PR #7486, opened 2025-07-26
refactor: Update Pantheon database pull to get fresh DB and file push to be CMS-agnostic

danny2p:
## The Issue

1. When executing `ddev pull pantheon`, ddev pulls a backup from Pantheon.  If this is a new site or if no backups exist for the environment, this will fail.  If there are no new backups, this may restore a very old snapshot. 
2. When executing `ddev push pantheon`, ddev uses drush rsync which is Drupal specific and will fail for WordPress sites.

1. Taking a page out of [Lando's book](https://github.com/lando/pantheon/blob/58cf95d9849e5b7aae01e5830edbaeadc2d19c0a/scripts/pull.sh#L129), we can grab a fresh mysqldump to ensure `ddev pull pantheon` gets us a current DB, whether or not there are backups.  This could also be remedied by firing off a `terminus backup:create`, but this would add significant time to the overall db pull operation.  A possible improvement here would be to allow a flag to specify whether the pull should get a realtime dump or an existing backup.
2. Instead of drush rsync, install the [terminus rsync plugin](https://github.com/pantheon-systems/terminus-rsync-plugin) and leverage terminus rsync for CMS-agnostic file push.  Since this function is not commonly used, I am OK with installing the plugin at time of execution but it may be desirable to install the terminus rsync plugin globally for Pantheon use, I don't feel strongly either way.

1. Fire up your ddev instance for a Pantheon Drupal or WordPress site
2. Using this updated yaml, execute `ddev pull pantheon --skip-files` to test the DB pull
3. To test the rsync, add some new files to your site's files directory (sites/default/files or wp-content/uploads), and execute `ddev push pantheon --skip-db`

## Release/Deployment Notes

Should have no functional impacts to existing instances, though db_pull may take longer now due to fresh mysqldump vs using existing archive

### DIFF
--- a/pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
+++ b/pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
@@ -67,13 +67,16 @@ auth_command:
     terminus auth:login --machine-token="${TERMINUS_MACHINE_TOKEN}" || ( echo "terminus auth login failed, check your TERMINUS_MACHINE_TOKEN" && exit 1 )
     terminus aliases 2>/dev/null
 
+#make sure env is awake, and grab a fresh DB export
 db_pull_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     set -eu -o pipefail
     echo "Using PANTHEON_SITE=${PANTHEON_SITE} PANTHEON_ENVIRONMENT=${PANTHEON_ENVIRONMENT}"
     pushd /var/www/html/.ddev/.downloads >/dev/null
-    terminus backup:get ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT} --element=db --to=db.sql.gz
+    terminus env:wake -- ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}
+    FRESH_DB_DUMP_STRING=$(terminus connection:info "$project" --field=mysql_command | sed 's,^mysql,mysqldump --no-autocommit --single-transaction --opt -Q,')
+    eval "$FRESH_DB_DUMP_STRING | gzip > db.sql.gz" > /dev/null 2>&1
 
 files_pull_command:
   command: |
@@ -94,10 +97,13 @@ db_push_command:
     gzip -dc db.sql.gz | terminus remote:drush ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT} -- sql-cli
 
 # push is a dangerous command. If not absolutely needed it's better to delete these lines.
+# terminus rsync plugin for CMS agnostic file push
+# this will push the files directory to Pantheon
 files_push_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     ssh-add -l >/dev/null || ( echo "Please 'ddev auth ssh' before running this command." && exit 1 )
     set -eu -o pipefail
     ls ${DDEV_FILES_DIR} >/dev/null # This just refreshes stale NFS if possible
-    drush rsync -y @self:%files @${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}:%files
+    terminus self:plugin:install terminus-rsync-plugin
+    terminus rsync -y ${DDEV_FILES_DIR}/ ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}:files


pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
```diff
@@ -67,13 +67,16 @@ auth_command:
...
+#make sure env is awake, and grab a fresh DB export
 db_pull_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     set -eu -o pipefail
     echo "Using PANTHEON_SITE=${PANTHEON_SITE} PANTHEON_ENVIRONMENT=${PANTHEON_ENVIRONMENT}"
     pushd /var/www/html/.ddev/.downloads >/dev/null
-    terminus backup:get ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT} --element=db --to=db.sql.gz
+    terminus env:wake -- ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}
+    FRESH_DB_DUMP_STRING=$(terminus connection:info "$project" --field=mysql_command | sed 's,^mysql,mysqldump --no-autocommit --single-transaction --opt -Q,')
```
rfay:
```suggestion
    FRESH_DB_DUMP_STRING=$(terminus connection:info "${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}" --field=mysql_command | sed 's,^mysql,mysqldump --no-autocommit --single-transaction --opt -Q,')
```

danny2p:
@rfay -- this is a quick and dirty take at what we talked about at Drupalcamp yesterday.  

Thinking about this further, it might be better to default to pulling existing db backup, and only grab the fresh mysqldump if a new backup is absent, but on Pantheon it's common that sandox sites and dev environments don't have fresh nightly backups, since only paid sites have automatic nightly backups, and dev / multidev environments may not have scheduled backups enabled.  I'm not sure if speed of execution, or freshness of DB is more important here.

rfay (changes requested):
WOW, awesome, thank you so much!

It was missing one key thing, so don't forget to manually test. 

But with the change suggested I was able to run TestPantheonPull and TestPantheonPush successfully.

pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
danny2p: oh crap, remnant of the fact I started on this before i pulled YOUR latest refactor from a few days ago, apologies!

rfay (changes requested):
What a great step forward, thanks! It passes TestPantheonPull and TestPantheonPush now. 

I see that these two docs and pseudo-docs need to be adjusted and it should be mighty close.
- [x] Number 3 here in the pantheon.yaml
- [x] Number 3 in the docs/content/users/providers/pantheon.md

If you have opinions about those two in other ways adjustment is welcome.

Thanks for the great work!

pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
```diff
@@ -67,13 +67,16 @@ auth_command:
...
+#make sure env is awake, and grab a fresh DB export
 db_pull_command:
   command: |
     # set -x   # You can enable bash debugging output by uncommenting
     set -eu -o pipefail
     echo "Using PANTHEON_SITE=${PANTHEON_SITE} PANTHEON_ENVIRONMENT=${PANTHEON_ENVIRONMENT}"
     pushd /var/www/html/.ddev/.downloads >/dev/null
-    terminus backup:get ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT} --element=db --to=db.sql.gz
+    terminus env:wake -- ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}
+    FRESH_DB_DUMP_STRING=$(terminus connection:info "${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}" --field=mysql_command | sed 's,^mysql,mysqldump --no-autocommit --single-transaction --opt -Q,')
```
rfay, line 73: Why is the sed in there?

rfay:
I did the suggested items; @danny2p if you could review and make sure you like the results it would be appreciated. And that sed in there... I'd like to know if it's important.

pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
```diff
@@ -67,13 +63,16 @@ auth_command:
...
+    eval "$FRESH_DB_DUMP_STRING | gzip > db.sql.gz" > /dev/null 2>&1
```
stasadev: I'm not sure about this change, it's very slow even for a small database and you don't see any useful progress.

I would prefer:

```
echo "Waking up the project environment..."
terminus env:wake -- ${PANTHEON_SITE}.${PANTHEON_ENVIRONMENT}
...
echo "Running mysqldump on the live database..."
...
```

And maybe it should be optional by default. Some new `env` can be introduced:

```
ddev pull pantheon --environment=PANTHEON_PULL_NO_BACKUP=true
```

rfay: There may be a faster way to do accomplish it. But it's a huge benefit because the confusion of having to create and wait for a backup upstream has always been quite awkward.

pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
stasadev: It comes from https://github.com/lando/pantheon/blob/58cf95d9849e5b7aae01e5830edbaeadc2d19c0a/scripts/pull.sh#L129

`terminus connection:info --field=mysql_command` receives a string like:

```
mysql -u ... -p ... -h ...
```

And `sed` replaces it, so the commands looks like this:

```
mysqldump --no-autocommit --single-transaction --opt -Q -u ... -p ... -h ...
```

pkg/ddevapp/dotddev_assets/providers/pantheon.yaml
stasadev: I see.
Then at least an `echo` can help with understanding what commands are executed while waiting.

rfay:
Please take another look @stasadev @danny2p 
- More feedback during `ddev pull pantheon`
- Compressed data stream during pull
- `ddev push pantheon` now uses a generic technique instead of using drush

This is a forked PR so TestPantheonPull and TestPantheonPush won't run. I ran them manually and they were successful

DigitalFrontiersMedia:
This is cool.  I was about to go back to #5215 in the next few weeks.  Sorry I didn't get to it in a timely manner.  I still think it would be better if the rsync plugin could be installed as a part of the image (as terminus itself is) instead of running the install command on every push.  But I was hitting permissions issues trying to do it the way the tools expect.  So long as there's not a performance hit by doing it this way, an advantage here is that the terminus rsync plugin will always be updated if a new version comes out--without the need of a new ddev image release.

PR #7489, opened 2025-07-28
fix: Disable corepack on config update for all Drupal versions except 11, fixes #7488

### LINKED ISSUE #7488: Switching back to `drupal10` as type does not disable corepack
### Preliminary checklist

N/A 

### Expected Behavior

When a project type changes from `drupal11` to `drupal10` then corepack should be disabled.

### Actual Behavior

Corepack remains enabled.

### Steps To Reproduce

1. Setup a new project by using `drupal11` as type.
2. Run `ddev yarn --version` and confirm Yarn 4.x.
3. Change type to `drupal10` and run `ddev restart`.
2. Run `ddev yarn --version` and you should see Yarn 1 but you will get 4.x

### Anything else?

_No response_

mxr576:
## The Issue

- #7488

1. Setup a new project by using `drupal11` as type.
1. Run `ddev yarn --version` and confirm Yarn 4.x.
1. Change type to `drupal10` and run `ddev restart`.
1. Run `ddev yarn --version` and you should see Yarn 1 but you will get 4.x without this change.

### DIFF
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -427,6 +427,7 @@ func drupalConfigOverrideAction(app *DdevApp) error {
 		app.PHPVersion = nodeps.PHP81
 		drupalVersion = 9
 	case nodeps.AppTypeDrupal10:
+		app.CorepackEnable = false
 		drupalVersion = 10
 	case nodeps.AppTypeDrupal:
 		fallthrough
--- a/pkg/ddevapp/drupal.go
+++ b/pkg/ddevapp/drupal.go
@@ -412,6 +412,7 @@ func drupal7ConfigOverrideAction(app *DdevApp) error {
 // drupalConfigOverrideAction selects proper versions for
 func drupalConfigOverrideAction(app *DdevApp) error {
 	var drupalVersion int
+	app.CorepackEnable = false
 	switch app.Type {
 	case nodeps.AppTypeDrupal6:
 		app.PHPVersion = nodeps.PHP56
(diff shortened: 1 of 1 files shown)


mxr576:
Tentative fix, needs double checking if my idea works or not.

pkg/ddevapp/drupal.go
```diff
@@ -427,6 +427,7 @@ func drupalConfigOverrideAction(app *DdevApp) error {
 		app.PHPVersion = nodeps.PHP81
 		drupalVersion = 9
 	case nodeps.AppTypeDrupal10:
+		app.CorepackEnable = false
```
stasadev: I think it should be disabled for all Drupal types below 11, so it would be easier to add this line before `switch`.

Thank you for your contribution.

mxr576: Makes sense, moved the line

stasadev (commented):
Testing with Drupal 10 project:
```
mkdir my-drupal-site && cd my-drupal-site
ddev config --project-type=drupal10 --docroot=web
ddev start
ddev composer create-project "drupal/recommended-project:^10"
```

Using v1.24.7:
```
$ ddev config --corepack-enable=true
$ ddev config --update
$ ddev debug configyaml | grep corepack
corepack_enable: true
```

Using this PR:
```
$ ddev config --corepack-enable=true
$ ddev config --update
$ ddev debug configyaml | grep corepack
(nothing, i.e. corepack_enable: false)
```

rfay:
Thank you so much for your contribution and for stepping up to this perceived problem. 

I have decided that this isn't something we want to do, and it would have immense and intrusive consequences if we did it for all project types. In general, people should turn off corepack if they don't want it, no matter what project type they are using.

mxr576:
No problem, thanks for the insightful conversation on this PR and the related issue.

Issue #7484, opened 2025-07-24
`ddev ssh` exits the shell early


GuySartorelli:
### Preliminary checklist

- [x] I am using the latest stable version of DDEV (see [upgrade guide](https://ddev.readthedocs.io/en/stable/users/install/ddev-upgrade/))
- [x] I have searched [existing issues](https://github.com/ddev/ddev/issues)
- [x] I have checked the [troubleshooting guide](https://ddev.readthedocs.io/en/stable/users/usage/troubleshooting/)
- [x] I have run `ddev debug test` to include output below

### Output of `ddev debug test`

<details><summary>Expand `ddev debug test` diagnostic information</summary>

```


======== Output file will be in /tmp/ddev-debug-test.txt ========


======== Existing project config ========
ddev installation alternate locations:
/usr/bin/ddev
/bin/ddev

These config files were loaded for project subfolder-installation: [/home/gsartorelli/projects/subfolder-installation/.ddev/config.yaml]
name: subfolder-installation
type: php
php_version: 8.3
webserver_type: nginx-fpm
webimage: ddev/ddev-webserver:v1.24.7
additional_hostnames: []
additional_fqdns: []
database: {mariadb 10.11}
project_tld: ddev.site
use_dns_when_possible: true
composer_version: 2
nodejs_version: 22
default_container_timeout: 120


======== existing project customizations ========


======== installed DDEV add-ons ========
No registered add-ons were found.


======== WSL2 information ========


======== mutagen situation ========
looking for #ddev-generated in mutagen.yml in project /home/gsartorelli/projects/subfolder-installation

.ddev/mutagen/mutagen.yml not found


======== Creating dummy project named tryddevproject-30948 in ../tryddevproject-30948 ========
Creating a new DDEV project config in the current directory (/home/gsartorelli/projects/tryddevproject-30948)
Once completed, your configuration will be written to /home/gsartorelli/projects/tryddevproject-30948/.ddev/config.yaml

Configuring a 'php' project named 'tryddevproject-30948' with docroot 'web' at '/home/gsartorelli/projects/tryddevproject-30948/web'.
For full details use 'ddev describe'.
Configuration complete. You may now run 'ddev start'.


======== OS Information ========
Linux silverstripe-gsartorelli 6.12.10-76061203-generic #202412060638~1751390742~22.04~79b9668 SMP PREEMPT_DYNAMIC Tue J x86_64 x86_64 x86_64 GNU/Linux


======== User information ========
[...]

Issue #7108, opened 2025-03-17
v1.25.0 deprecations and defaults changes


rfay:
We have a number of deprecations and other changes for DDEV v1.25.0

- [x] Change xhprof_mode default to `xhgui` https://github.com/ddev/ddev/issues/7808
- [x] Remove nfs support https://github.com/ddev/ddev/issues/7810
- [x] Actually remove `ddev service *` https://github.com/ddev/ddev/issues/7812
- [x] Remove `ddev debug capabilities`, https://github.com/ddev/ddev/issues/7174
- [x] Remove `defaultRuleSyntax: v2` from traefik global config; then we also won't need `ruleSyntax: v3` in the generated project traefik config.  https://github.com/ddev/ddev/pull/7823
- [x] Update node default version to v24, current LTS #7821
- [x] If v1.25.0 comes in Fall 2025, it can be used to update to newest PHP 8.5, possibly changing default MariaDB version.
- [x] Hope for ddev-webserver update to Debian Trixie (https://github.com/rfay/ddev/tree/20250728_rfay_debian_trixie)
- [x] Remove Drupal 11 hack for sqlite (I started work on this in https://github.com/rfay/ddev/tree/20250728_rfay_debian_trixie )
- [x] Remove mariadb_version and mysql_version silent compatibility and remove the code that handles them. Use `database` only.
- [x] Change default mariadb version for new projects to 11.8, see https://endoflife.date/mariadb
- [x] Default PHP version for new projects becomes 8.4 https://github.com/ddev/ddev/issues/7874
- [x] Remove nvm and replace it with instructions about how to use nvm if people want to.
- [x] Remove `--disable-settings-management=true` from the Laravel SQLite quickstart, see #7584
~~- [ ] Consider changing the default of --omit-project-name-by-default to true~~
- [x] Remove obsolete flags from global and local config (like `mutagen-enabled`)
- [x] Remove the code for the old nginx-proxy router https://github.com/ddev/ddev/pull/7818
- [x] Update all go modules/dependencies
- [x] Update docker-compose

Release notes:
- [x] Some projects with complex Dockerfiles or obsolete `webimage_extra_packages` may have to be updated

fallinboul:
I think I might a solution to your problem. Check it out here [click](https://javierlopez.eu/captha/)

rfay:
Added to list:
> Remove `defaultRuleSyntax: v2` from traefik global config; then we also won't need `ruleSyntax: v3` in the generated project traefik config.

[...]

PR #7492, opened 2025-07-29
build: Move ddev-webserver base image to debian trixie

rfay:
## The Issue

- #7108

In DDEV v1.25.0 we expect to migrate the base image for ddev-webserver to debian trixie v13.

I know this is jumping the gun. I was inspired by #7484 from @GuySartorelli but they reported later it wasn't a fix. I did it anyway. There were a few workarounds required to get the various components going.

## TODO

- [ ] Convince blackfire to go to a newer key type (or maybe they have and we're using an old recipe (revert workaround ) I requested this change in [their discord](https://discord.com/channels/1121795479785721957/1399735861465186354/1399735861465186354)
- [ ] Get mariadb to support Debian Trixie with their client installer package (revert workaround https://github.com/ddev/ddev/commit/d25f4c014f30422f3ef230e58d08bc8566ec4af7 ) https://jira.mariadb.org/browse/MDEV-37241
- [ ] Get full support for needed PHP extensions from deb.sury.org and (revert workaround https://github.com/ddev/ddev/commit/35ecc77b4c99a3f07d382f5e9c3c1011273a8ba9 ) - see https://github.com/oerdnj/deb.sury.org/issues/2330
- [ ] Re-enable memcached/redis tests after deb.sury.org updated, revert https://github.com/ddev/ddev/commit/35ecc77b4c99a3f07d382f5e9c3c1011273a8ba9
- [ ] The base install of mariadb-client is explicitly using debian 12 and probably needs to be updated, https://github.com/ddev/ddev/blob/87d4aa526fc22cfb161d840b8201b42ebc458ce7/containers/ddev-webserver/Dockerfile#L101-L102

* Use Trixie as base
* Remove terrible sqlite 3.45.1 hacks for Drupal 11
* Experimental: Use Debian trixie nginx packages instead of nginx.org
* Upgrade libmagiccore

Try it out

### DIFF
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
@@ -15,7 +15,7 @@ MARIADB_VERSION=${DDEV_DATABASE#*:}
 
 # Configure the correct repository for mariadb
 set -x
-log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --os-type=debian --os-version=bookworm || exit $?
+log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --os-type=debian --os-version=bookworm --skip-key-import || exit $?
 rm -f /etc/apt/sources.list.d/mariadb.list.old_*
 # --skip-key-import flag doesn't download the existing key again and omits "apt-get update",
 # so we can run "apt-get update" manually only for mariadb and debian repos to make it faster
--- a/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
+++ b/containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
@@ -15,7 +15,7 @@ MARIADB_VERSION=${DDEV_DATABASE#*:}
 
 # Configure the correct repository for mariadb
 set -x
-log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --skip-key-import || exit $?
+log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --os-type=debian --os-version=bookworm || exit $?
 rm -f /etc/apt/sources.list.d/mariadb.list.old_*
 # --skip-key-import flag doesn't download the existing key again and omits "apt-get update",
 # so we can run "apt-get update" manually only for mariadb and debian repos to make it faster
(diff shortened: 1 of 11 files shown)


containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
```diff
@@ -15,7 +15,7 @@ MARIADB_VERSION=${DDEV_DATABASE#*:}
 
 # Configure the correct repository for mariadb
 set -x
-log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --skip-key-import || exit $?
+log-stderr.sh --timeout "${START_SCRIPT_TIMEOUT:-30}" mariadb_repo_setup --mariadb-server-version="mariadb-${MARIADB_VERSION}" --skip-maxscale --skip-tools --os-type=debian --os-version=bookworm || exit $?
```
stasadev: `--skip-key-import` should be here because we already have it in the base image, with skip it's faster.

rfay: Interestingly, it failed with no key I had it that way (at DDEV build time). I'm not sure why.

stasadev: > (at DDEV build time)

On `ddev start` ?

Or while building `ddev-webserver` ?

---

Because `--skip-key-import` should not be there in `Dockerfile` for `ddev-webserver`.

rfay: Yes, it worked during Dockerfile build, but when I tested it manually inside ddev-webserver it failed due to lack of key. I didn't study it very carefully, but the key seemed not to be installed. My test technique was to run `sudo /usr/local/bin/mariadb_repo_setup --mariadb-server-version="mariadb-11.8" --skip-maxscale --skip-tools --os-type=debian --os-version=bookworm  ` inside `ddev ssh` - it succeeded but then the `sudo apt install mariadb-client` failed with no key.

stasadev: Okay, maybe this combination of flags requires a key.
I didn't carefully study the `mariadb_repo_setup` script.

rfay: Will test again, maybe my approach was wrong.

rfay:
It passes tests! I'll leave this as draft and try to poke at the action items. I don't expect that we would do this until v1.25.0, which should also include PHP 8.5.

containers/ddev-webserver/ddev-webserver-base-files/usr/local/bin/mariadb-client-install.sh
rfay: You were right, did this.

rfay:
@stasadev I don't remember a reason that we used the upstream nginx.org packages, but it probably dates back a couple of Debian versions.

https://github.com/ddev/ddev/pull/7492/files#diff-ddc28b17c63eedfc281d63de8ae258a7b18a40b47b68b275df4c460a630cdcf2R36-R40

stasadev:
> a reason that we used the upstream nginx.org packages

I traced it back to:

- https://github.com/ddev/ddev/pull/993/files#diff-131a1297bdf1b5e28b280c2e8e473b47e9f959fe10efd73f421f2e0e058a9364R41

But I didn't find a reason why nginx repo was added.

rfay:
Jessie in 2018! I think we should probably go with the Debian stuff.

rfay:
I'm closing this for now, will reopen when we get closer. All of the external requests have been made, and all can be worked around if we need to... except the problems with deb.sury.org.

rfay:
> Adjust MariaDB client setup expectations (and ours). See https://jira.mariadb.org/browse/MDEV-37241 "Because MariaDB downgrades, particularly with packages, aren't supported or tested the Trixie package will begin at 11.8 and there will be no 11.4 packages."

PR #7500, opened 2025-08-01
build: use ddev/ddev-utilities instead of busybox, fixes #7499

### LINKED ISSUE #7499: Use ddev/ddev-utilities instead of busybox

We have a little trouble managing the `busybox` image in the places we use it.

And now that we provide a small image for "utility" things, why don't we use that instead of `busybox` ?

### Describe your solution

Replace the usages of `busybox` with `ddev/ddev-utilities`

### Describe alternatives

_No response_

### Additional context

_No response_

rfay:
## The Issue

- #7499

We already require ddev/ddev-utilities and have it downloaded. Using it will help us to better manage images that we control.

* Use ddev/ddev-utilities in user-facing activities
* I avoided changes to tests, as they would be intrusive and not useful
* busybox is still used widely in docs and examples, and it's fine there.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -1873,11 +1873,11 @@ func (app *DdevApp) PullContainerImages(pullAlways bool) error {
 }
 
 // PullBaseContainerImages pulls only the fundamentally needed images so they can be available early.
-// We always need web image, busybox, and ddev-utilities for housekeeping.
+// We always need web image, and ddev-utilities for housekeeping.
 func PullBaseContainerImages(additionalImages []string, pullAlways bool) error {
 	base := []string{
 		ddevImages.GetWebImage(),
-		versionconstants.BusyboxImage,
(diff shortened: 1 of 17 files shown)


pkg/ddevapp/ddevapp.go
```diff
@@ -1873,11 +1873,11 @@ func (app *DdevApp) PullContainerImages(pullAlways bool) error {
 }
 
 // PullBaseContainerImages pulls only the fundamentally needed images so they can be available early.
-// We always need web image, busybox, and ddev-utilities for housekeeping.
+// We always need web image, and ddev-utilities for housekeeping.
 func PullBaseContainerImages(additionalImages []string, pullAlways bool) error {
 	base := []string{
 		ddevImages.GetWebImage(),
-		versionconstants.BusyboxImage,
+		versionconstants.UtilitiesImage,
```
stasadev, line 1880:
```suggestion
```

stasadev (approved):
Looks good to me.

PR #7504, opened 2025-08-03
docs: Extend the Drupal installation instructions

BaluErtl:
## The Issue

- [x] 1. macOS' default file manager app _Finder_ automatically creates a hidden `.DS_Store` file when opening an empty directory. The Composer script invoked by the `create-project` command checks whether the directory at the given path is entirely empty and exits with an error message if something is found within. The easiest way to ensure Composer's project creation script can make it through is to NOT visit this folder in Finder after getting manually created in the CLI, only after the script has finished running.<br><br>Some users with less experience/confidence regarding how the macOS file system works might stumble upon this issue and eventually give up the installation process. – Being resolved in MR #7469.
- [ ] 2. The `$ drush uli` command can take an extra argument for a valid URL path existing in Drupal, where the user gets redirected upon successful authentication. With no such argument provided, the default page of the website, the frontpage, is being presented, showing a friendly welcome message useful for newcomer evaluators.<br><br>Suggesting the invocation of a pure `$ ddev launch` command (without piping in the output of `$ drush uli`) is not useful at all, because the user might feel "closed out" themselves from their freshly installed local website instance.

1. Add a friendly reminder before the first step of the process
2. Delete the (probably) less-useful not-authenticating command
3. Replace it with both auto-authenticating commands with a choice offered

Go through the steps of the installation process.

## Automated Testing Overview

This is documentation update only.

## Release/Deployment Notes

After merging, one day it should appear on https://ddev.readthedocs.io/en/stable/users/quickstart/#drupal

### DIFF
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -228,9 +228,11 @@ Read more about customizing the environment and persisting configuration in [Pro
     ddev composer create-project "drupal/recommended-project:^11"
     ddev composer require drush/drush
     ddev drush site:install --account-name=admin --account-pass=admin -y
-    ddev launch
-    # or automatically log in with
+    # Visit the website by opening its frontpage with a welcome message:
+    ddev launch $(ddev drush uli)
+    # Visit the website by opening its more technical status report page:
     ddev launch $(ddev drush uli admin/reports/status)
+    # (Both commands automatically signs you in to the admin interface.)
     ```
 
     Read more about: [Drupal Core](https://new.drupal.org/about/overview/technical) & [Documentation](https://www.drupal.org/docs)
@@ -257,9 +259,11 @@ Read more about customizing the environment and persisting configuration in [Pro
     ddev composer create-project "drupal/recommended-project:^10"
     ddev composer require drush/drush
     ddev drush site:install --account-name=admin --account-pass=admin -y
-    ddev launch
-    # or automatically log in with
+    # Visit the website by opening its frontpage with a welcome message:
+    ddev launch $(ddev drush uli)
+    # Visit the website by opening its more technical status report page:
     ddev launch $(ddev drush uli admin/reports/status)
+    # (Both commands automatically signs you in to the admin interface.)
     ```
 
     Read more about: [Drupal Core](https://new.drupal.org/about/overview/technical) & [Documentation](https://www.drupal.org/docs)
--- a/docs/content/users/quickstart.md
+++ b/docs/content/users/quickstart.md
@@ -229,10 +229,8 @@ Read more about customizing the environment and persisting configuration in [Pro
     ddev composer require drush/drush
     ddev drush site:install --account-name=admin --account-pass=admin -y
     # Visit the website by opening its frontpage with a welcome message:
-    ddev launch $(ddev drush uli)
-    # Visit the website by opening its more technical status report page:
-    ddev launch $(ddev drush uli admin/reports/status)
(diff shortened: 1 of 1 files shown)


docs/content/users/quickstart.md
```diff
@@ -221,15 +221,18 @@ Read more about customizing the environment and persisting configuration in [Pro
 === "Drupal 11"
 
     ```bash
+    # Don't open this folder in Finder yet – otherwise the `.DS_Store` file blocks the `composer create-project` script.
```
rfay: The .DS_Store problem was already solved in
* https://github.com/ddev/ddev/pull/7469
```suggestion
```

docs/content/users/quickstart.md
```diff
@@ -249,15 +252,18 @@ Read more about customizing the environment and persisting configuration in [Pro
...
     ddev composer create-project "drupal/recommended-project:^10"
     ddev composer require drush/drush
     ddev drush site:install --account-name=admin --account-pass=admin -y
-    ddev launch
-    # or automatically log in with
+    # Visit the website by opening its frontpage with a welcome message:
     ddev launch $(ddev drush uli)
+    # Visit the website by opening its more technical status report page:
+    ddev launch $(ddev drush uli admin/reports/status)
+    # (Both commands automatically signs you in to the admin interface.)
```
rfay: Nice, I sure didn't know about that! 

Why do you think it's important for people to land on admin/reports/status?

BaluErtl: Meanwhile, I just realized I was incorrect in the issue description: leaving off the URL path argument from the `$ drush uli` command leads you not to the frontpage (where Drupal's welcome message awaits) but to the user profile (where nothing interesting is displayed).

To answer your question, the reason why I thought it could be helpful for a first-time evaluator to offer two options to choose from because that is the case in the current docs too.

However, I tend to follow the simpler-the-docs-easier-to-follow rule, so I revoke my suggestion of the status page.

docs/content/users/quickstart.md
BaluErtl: Thanks for the heads-up, now I deleted this advice as it will be (hopefully) not needed to be alarmed for.

docs/content/users/quickstart.md
```diff
@@ -227,9 +227,9 @@ Read more about customizing the environment and persisting configuration in [Pro
     ddev composer create-project "drupal/recommended-project:^11"
     ddev composer require drush/drush
     ddev drush site:install --account-name=admin --account-pass=admin -y
-    ddev launch
-    # or automatically log in with
-    ddev launch $(ddev drush uli)
+    # Visit the website by opening its frontpage with a welcome message:
+    ddev launch $(ddev drush uli "/")
```
rfay, line 231: I don't mind doing `ddev launch $(ddev drush uli)` as the default instruction, but I've never seen anybody use `ddev drush uli "/"` before, and it's exactly the same.

The purpose of these quickstarts is to just get somebody going with ddev quickly, not necessarily to give instruction about tools like drush. 

But I didn't know `drush uli` could take an argument after 20 years :) Am glad to know it!

rfay:
Thank you so much for taking the time to create this PR suggestion, and I sure learned something about drush!

Since the Quickstart is really about teaching DDEV usage, not drush usage, though, I think I prefer to pass on this one.  I hope you'll do the next one you discover, and I apologize for not accepting this one after the time you took on it.

PR #7507, opened 2025-08-04
fix(traefik): improve router port discovery and optimize YAML writes, fixes ddev/ddev-mongo#24

stasadev:
## The Issue

- https://github.com/ddev/ddev-mongo/issues/24

Router doesn't become healthy because of missing Traefik entrypoints for optional services.

- Adds host port detection for Traefik entrypoint by reading `.ddev/.ddev-docker-compose-full.yaml`
- Removes hardcoded XHGui ports, they are now read from YAML.
- Optimizes file writes for `.ddev/.ddev-docker-compose-base.yaml` and `.ddev/.ddev-docker-compose-full.yaml` if the content is not changed.

Start a project with optional profiles:

```
ddev add-on get ddev/ddev-mongo
ddev start
```

Check `~/.ddev/traefik/.static_config.yaml`:

```yaml
entryPoints:
    http-80:
        address: :80
    http-443:
        address: :443
    http-8025:
        address: :8025
    http-8026:
        address: :8026
    http-8142:                 <= comes from xhgui
        address: :8142
    http-8143::                <= comes from xhgui
        address: :8143
    http-9091:                 <= comes from mongo-express
        address: :9091
    http-9092:                 <= comes from mongo-express
        address: :9092
    traefik:
        address: :10999
```

### DIFF
--- a/pkg/ddevapp/router.go
+++ b/pkg/ddevapp/router.go
@@ -338,8 +338,9 @@ func GetRouterStatus() (string, string) {
 	return status, logOutput
 }
 
-// determineRouterPorts returns a list of port mappings retrieved from running site
-// containers defining HTTP_EXPOSE and HTTPS_EXPOSE env var
+// determineRouterPorts returns a list of port mappings retrieved from ports from
+// configuration files of all active projects, plus running site
+// containers defining HTTP_EXPOSE and HTTPS_EXPOSE env var.
 // It is only useful to call this when containers are actually running, before
 // starting ddev-router (so that we can bind the port mappings needed
 func determineRouterPorts() []string {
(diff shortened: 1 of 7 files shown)


rfay:
Oh, I expected the fix to be in ddev-mongo! So wonderful that we have tests using this from so many angles.

pkg/ddevapp/router.go
```diff
@@ -349,10 +350,41 @@ func determineRouterPorts() []string {
 		util.Failed("Failed to retrieve containers for determining port mappings: %v", err)
 	}
 
-	// Add potential ports that might not be in use by current containers
-	// 8142/8143 are for xhgui, which might not be enabled, but we don't want to
-	// have to rebuild traefik when it does get enabled.
-	routerPorts = []string{nodeps.DdevDefaultXHGuiHTTPSPort, nodeps.DdevDefaultXHGuiHTTPPort}
```
stasadev, line 355: This PR makes port detection dynamic, so we no longer need to hardcode XHGui.

Though if the service isn't started, it's unclear why the ports should be occupied.

But this comment gives some justification:

> we don't want to have to rebuild traefik when it does get enabled.

What I see is that if we don't add xhgui all the time, then the project configuration in `.ddev/traefik/config/<project>.yaml` won't work as expected for people removing `#ddev-generated` from that file.

I didn't look deep into it.

rfay (commented):
I haven't studied this carefully, but this seems like an important improvement that needs a test?

stasadev:
> Consider splitting `determineRouterPorts()` for better maintainability
> Add unit tests for the new `processExposePorts()` function

Done with Claude Code.

> Integration test verifying port discovery from stopped services

I modified `TestCmdStartOptionalProfiles()` for this.
It now "successfully" fails for both:
- DDEV v1.24.7 (while running Traefik API validation)
- DDEV HEAD (while running `ddev start`).

rfay:
It's picking up the `goroutines=3` stuff in [tests](https://github.com/ddev/ddev/actions/runs/16812848257/job/47622236017?pr=7507#step:17:21561):
```

Messages:   	failed to parse JSON response for http router: {"entryPoints":["http-18125"],"service":"TestCmdWordpress-busybox1-80","rule":"HostRegexp(`^testcmdwordpress\\.ddev\\.site$`)","ruleSyntax":"v3","priority":44,"observability":{"accessLogs":true,"metrics":true,"tracing":true,"traceVerbosity":"minimal"},"status":"enabled","using":["http-18125"],"name":"TestCmdWordpress-busybox1-80-http@file","provider":"file"}
--
  | goroutines=3 at exit of main()
```

rfay (approved):
It's still great, thanks, and thanks for understanding, studying, and solving this, and caring for ddev-mongo. 

I wonder if this means that we could now start/restart traefik before everything else is up and ready, which might slightly improve user experience.  (Related idea: Most people in most situations don't really need to wait for everything to be fully ready. We might want to give them an out like "press return to get to work")

stasadev:
> I wonder if this means that we could now start/restart traefik before everything else is up and ready, which might slightly improve user experience.

This is possible, most likely requires moving all dynamic Traefik configuration to the start of `ddev start`.

Issue #3878, opened 2022-06-01
Use 1Password as the ssh-agent (macOS) instead of using `ddev-sshagent`


plexmaniax:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

We started using 1pass SSH Agent https://blog.1password.com/1password-ssh-agent/.
In our usecase we don't need to save the ssh key pair local anymore. This is why ddev auth ssh does not find any ssh keys on the root system. 

Could you find a way to implement this feature in ddev?

### Describe the solution you'd like

It would be great if ddev would support 1pass ssh agent, so we could work without any ssh key pairs on the local machines.

rfay:
I don't know how to accomplish this. If we could do this, we would be able to support keys with a regular ssh-agent (as on mac or Linux, for example). But the containers that need this don't have access to that agent. That's why we run a separate ddev-ssh-agent that serves the containers in the docker network.

nico-loeber:
Hey @plexmaniax,
This is a very specific use case I think. Anyway,  Anyway, I may have an idea how to solve this.

@rfay is it possible to add a single file mount to the container by configuration?
In case it is, mount the socket of your local 1pass agent `~/.1password/agent.sock` to `/mnt/.1password/agent.sock` in the container.

Then add the following config to your `~/.ddev/homeadditions/.ssh/config` configuration.

```
# Share known_hosts across all users of the ssh-agent
UserKnownHostsFile=/home/.ssh-agent/known_hosts
# In the container host key checking can be a problem with non-interactive composer, etc.
# Turn it off. When we upgrade to openssh 7.6 we can use the "accept-new" option instead of "no"
StrictHostKeyChecking=no

Host *
    IdentityAgent "/mnt/.1password/agent.sock"
```

I assume this will override the default one.
I'm also not sure how this behaves in combination with the SSH container, it's just an idea....

rfay:
Actually, that's a good point. The whole reason ddev-ssh-agent had to be created back in the day was the lack of support for mounting a socket into the container. That now works on mac (where it didn't for years) and it's always worked on Linux. I imagine it doesn't work on Windows though. 

And yes, a `docker-compose.*.yaml` could add a mount of a socket.

scrummer:
Hi all,

Interesting approach overall and wanted to try it (tried several ways) but does not seem to work.
[...]

PR #7511, opened 2025-08-05
feat: replace `docker run` in `ddev auth ssh` with Docker API and accept stdin

stasadev:
## The Issue

- #6720

I couldn't reopen the closed PR, so this is a new attempt.

Replace the `docker run` approach with direct Docker client API calls for better control over container I/O and error handling. This enables proper support for both interactive terminal usage and piped input scenarios.

Key improvements:
- Use Docker client API instead of `docker run` for container management
- Add proper terminal raw mode handling to hide passphrase input
- Preserve `ssh-add` error codes and provide more detailed error messages
- Remove `xargs` because it doesn't work with stdin
- Add non-interactive stdin (I didn't add this feature to the docs, because I don't think we should encourage people to use it.)

Prepare keys:

- `key 0` has "key 0" passphrase
- `key_1` has "key_1" passphrase
- `key_2` has no passphrase
- `key_3` has "key_3" passphrase

```
mkdir ~/test && cd ~/test
ssh-keygen -t rsa -b 4096 -C "key_0@example.com" -f "key 0" -N "key 0"
ssh-keygen -t rsa -b 4096 -C "key_1@example.com" -f "key_1" -N "key_1"
ssh-keygen -t rsa -b 4096 -C "key_2@example.com" -f "key_2" -N ""
ssh-keygen -t rsa -b 4096 -C "key_3@example.com" -f "key_3" -N "key_3"
```

- With v1.24.7:

	Press `<Enter>`,  `<Enter>`,  `<Enter>` (it fails, but it should fail early after the first key):
	```
	$ ddev auth ssh -d ~/test
	Enter passphrase for key 0: 
	Enter passphrase for key_1: 
	Identity added: key_2 (key_2@example.com)
	Enter passphrase for key_3: 
	Docker command 'docker run -it --rm --volumes-from=ddev-ssh-agent --user=1000 --entrypoint= "--mount=type=bind,src=/home/stas/test/key 0,dst=/tmp/sshtmp/key 0" --mount=type=bind,src=/home/stas/test/key_1,dst=/tmp/sshtmp/key_1 --mount=type=bind,src=/home/stas/test/key_2,dst=/tmp/sshtmp/key_2 --mount=type=bind,src=/home/stas/test/key_3,dst=/tmp/sshtmp/key_3 ddev/ddev-ssh-agent:v1.24.7-built bash -c "cp -r /tmp/sshtmp ~/.ssh && chmod -R go-rwx ~/.ssh && cd ~/.ssh && grep -l '^-----BEGIN .* PRIVATE KEY-----' * | xargs -d '\n' ssh-add"' failed: exit status 123
	```
	
	Write `key 0`, `<Enter>`, write `key_1`, `<Enter>`, write `key_3`, `<Enter>`, works as expected:
	```
	$ ddev auth ssh -d ~/test
	Enter passphrase for key 0: 
	Identity added: key 0 (key_0@example.com)
	Enter passphrase for key_1: 
	Identity added: key_1 (key_1@example.com)
	Identity added: key_2 (key_2@example.com)
	Enter passphrase for key_3: 
	Identity added: key_3 (key_3@example.com)
	```
	
	Stdin doesn't work:
	```
	$ printf "key 0\nkey_1\nkey_3\n" | ddev auth ssh -d ~/test
	the input device is not a TTY
	Docker command 'docker run -it --rm --volumes-from=ddev-ssh-agent --user=1000 --entrypoint= "--mount=type=bind,src=/home/stas/test/key 0,dst=/tmp/sshtmp/key 0" --mount=type=bind,src=/home/stas/test/key_1,dst=/tmp/sshtmp/key_1 --mount=type=bind,src=/home/stas/test/key_2,dst=/tmp/sshtmp/key_2 --mount=type=bind,src=/home/stas/test/key_3,dst=/tmp/sshtmp/key_3 ddev/ddev-ssh-agent:v1.24.7-built bash -c "cp -r /tmp/sshtmp ~/.ssh && chmod -R go-rwx ~/.ssh && cd ~/.ssh && grep -l '^-----BEGIN .* PRIVATE KEY-----' * | xargs -d '\n' ssh-add"' failed: exit status 1
	```

- With this PR:

	Press `<Enter>`, fails immediately as expected:
	```
	$ ddev auth ssh -d ~/test
	Enter passphrase for key 0: 
	Failed to execute command `ddev auth ssh`: exit status 1
	```
	
	Write `key 0`, `<Enter>`, write `key_1`, `<Enter>`, write `key_3`, `<Enter>`, works as expected:
	```
	$ ddev auth ssh -d ~/test
	Enter passphrase for key 0: 
	Identity added: key 0 (key_0@example.com)
	Enter passphrase for key_1: 
	Identity added: key_1 (key_1@example.com)
	Identity added: key_2 (key_2@example.com)
	Enter passphrase for key_3: 
	Identity added: key_3 (key_3@example.com
	```
	
	Stdin works as expected:
	```
	$ printf "key 0\nkey_1\nkey_3\n" | ddev auth ssh -d ~/test
	Enter passphrase for key 0: *Identity added: key 0 (key_0@example.com)
	Enter passphrase for key_1: *Identity added: key_1 (key_1@example.com)
	*Identity added: key_2 (key_2@example.com)
	Enter passphrase for key_3: *Identity added: key_3 (key_3@example.com)
	```

### DIFF
--- a/cmd/ddev/cmd/auth-ssh.go
+++ b/cmd/ddev/cmd/auth-ssh.go
@@ -1,20 +1,27 @@
 package cmd
 
 import (
+	"bufio"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
 	"github.com/ddev/ddev/pkg/ddevapp"
 	"github.com/ddev/ddev/pkg/docker"
-	"github.com/ddev/ddev/pkg/exec"
+	"github.com/ddev/ddev/pkg/dockerutil"
 	"github.com/ddev/ddev/pkg/fileutil"
 	"github.com/ddev/ddev/pkg/globalconfig"
 	"github.com/ddev/ddev/pkg/heredoc"
 	"github.com/ddev/ddev/pkg/nodeps"
 	"github.com/ddev/ddev/pkg/util"
+	dockerContainer "github.com/docker/docker/api/types/container"
+	dockerMount "github.com/docker/docker/api/types/mount"
+	dockerStrslice "github.com/docker/docker/api/types/strslice"
 	"github.com/spf13/cobra"
-	"os"
-	"path/filepath"
-	"slices"
-	"strings"
+	"golang.org/x/term"
 )
 
 var sshKeyFiles, sshKeyDirs []string
@@ -80,7 +87,8 @@ var AuthSSHCommand = &cobra.Command{
 			util.Failed("Failed to start ddev-ssh-agent container: %v", err)
 		}
 
-		var mounts []string
+		// Prepare mounts for Docker API
+		var mounts []dockerMount.Mount
 		// Map to track already added keys
 		addedKeys := make(map[string]struct{})
 		for i, keyPath := range keys {
@@ -91,18 +99,24 @@ var AuthSSHCommand = &cobra.Command{
 				filename = fmt.Sprintf("%s_%d", filename, i)
 			}
 			addedKeys[filename] = struct{}{}
-			mounts = append(mounts, "--mount=type=bind,src="+keyPath+",dst=/tmp/sshtmp/"+filename)
+			mounts = append(mounts, dockerMount.Mount{
+				Type:     dockerMount.TypeBind,
+				Source:   keyPath,
+				Target:   "/tmp/sshtmp/" + filename,
+				ReadOnly: true,
+			})
 			// Mount optional OpenSSH certificate
 			if certPath, certName := getCertificateForPrivateKey(keyPath, filename); certPath != "" && certName != "" {
-				mounts = append(mounts, "--mount=type=bind,src="+certPath+",dst=/tmp/sshtmp/"+certName)
+				mounts = append(mounts, dockerMount.Mount{
+					Type:     dockerMount.TypeBind,
+					Source:   certPath,
+					Target:   "/tmp/sshtmp/" + certName,
+					ReadOnly: true,
+				})
 			}
 		}
 
-		dockerCmd := []string{"run", "-it", "--rm", "--volumes-from=" + ddevapp.SSHAuthName, "--user=" + uidStr, "--entrypoint="}
-		dockerCmd = append(dockerCmd, mounts...)
-		dockerCmd = append(dockerCmd, docker.GetSSHAuthImage()+"-built", "bash", "-c", `cp -r /tmp/sshtmp ~/.ssh && chmod -R go-rwx ~/.ssh && cd ~/.ssh && grep -l '^-----BEGIN .* PRIVATE KEY-----' * | xargs -d '\n' ssh-add`)
-
-		err = exec.RunInteractiveCommand("docker", dockerCmd)
+		err = runSSHAuthContainer(uidStr, mounts)
 
 		if err != nil {
 			helpMessage := ""
@@ -111,7 +125,7 @@ var AuthSSHCommand = &cobra.Command{
 			if strings.Contains(err.Error(), "bind source path does not exist") {
 				helpMessage = "\n\nThe specified SSH private key path is not shared with your Docker provider."
 			}
-			util.Failed("Docker command 'docker %v' failed: %v %v", prettyCmd(dockerCmd), err, helpMessage)
+			util.Failed("Failed to run SSH auth container: %v %v", err, helpMessage)
 		}
 	},
 }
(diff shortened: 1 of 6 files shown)


stasadev:
`runSSHAuthContainer()` is quite unique, so I decided not to extract the logic from there to `dockerutils` package.

Today I simplified the logic for `stdin`, I spent several hours on this with AI (it gave me such bizarre and complex ideas), but finally I refactored it myself in [`51c5bc2`](https://github.com/ddev/ddev/pull/7511/commits/51c5bc20c4e5ac2ec650fb66d2534124777ecbbe).

Ready for review.

pkg/exec/exec.go
```diff
@@ -122,22 +145,26 @@ func RunHostCommand(command string, args ...string) (string, error) {
...
 
-// RunHostCommandWithEnv executes a command on the host with optional
-// environment variables and returns the
-// combined stdout/stderr results and error
-// If all of the existing environment is required, it must be
-// passed in `env`, as it is not set by default
-func RunHostCommandWithEnv(command string, env []string, args ...string) (string, error) {
+// RunHostCommandWithOptions executes a command on the host with configurable options
+// and returns the combined stdout/stderr results and error
+func RunHostCommandWithOptions(command string, options []CmdOption, args ...string) (string, error) {
```
rfay, line 150: I think this can probably use a test.

cmd/ddev/cmd/auth-ssh.go
```diff
@@ -197,6 +217,108 @@ func getCertificateForPrivateKey(path string, name string) (string, string) {
 	return certPath, certName
 }
 
+// runSSHAuthContainer runs the SSH auth container using Docker client API
+func runSSHAuthContainer(uidStr string, mounts []dockerMount.Mount) error {
```
rfay: This looks a lot like RunSimpleContainer(). Would it have been possible to use/improve that?

I assume the difference is the fancy terminal handling/streaming?

pkg/ddevapp/mutagen.go
```diff
@@ -566,7 +566,7 @@ func StopMutagenDaemon(mutagenDataDirectory string) {
 	}
 	if fileutil.FileExists(globalconfig.GetMutagenPath()) {
 		env := []string{"MUTAGEN_DATA_DIRECTORY=" + mutagenDataDirectory, "HOME=" + os.Getenv(`HOME`), "PWD=" + os.Getenv(`PWD`)}
-		out, err := exec.RunHostCommandWithEnv(globalconfig.GetMutagenPath(), env, "daemon", "stop")
+		out, err := exec.RunHostCommandWithOptions(globalconfig.GetMutagenPath(), []exec.CmdOption{exec.WithEnv(env)}, "daemon", "stop")
```
rfay, line 569: I'm surprised to see something in mutagen.go change; is this changing just because it seemed like a better way to accomplish the same goal?

rfay (changes requested):
This looks good to me, a couple of questions and suggested test.

I read through and stepped through, and things seem like they'll be fine.

* One of the key stumbling blocks for people has always been when they had unexpected files in their ~/.ssh directory. Those have often turned into support requests. Does this PR help with that situation?
* A key problem that we haven't really solved (and isn't in scope here but is worth thinking about) is the use/proxying of additional ssh agents. Currently we're just running our own inside the ddev-ssh-agent, but in a perfect world, we'd proxy existing agents, whether the Windows agent or the macOS agent or the 1Password agent. https://github.com/anotherjames/ddev-1password does this in at least some limited situations. IMO this is all out of scope for this PR, but it's such a [long-term discussion](https://github.com/ddev/ddev/issues/3878) that it's worth mentioning here. 

One of the excellent and interesting things about this implementation is that now I can step through the authentication in goland, whereas it was impossible before because terminal not detected. However, that also means that when I type my password in the debugger in goland, it's shown in plain text instead of hidden.

rfay:
Rebased.

pkg/ddevapp/mutagen.go
stasadev: I needed to pass stdin to the host command. A quick search showed there was already a custom function `RunHostCommandWithEnv`.

I added `RunHostCommandWithOptions`, which is highly customizable and can support any needed options. As a result, I removed `RunHostCommandWithEnv` since we now have a better version.

cmd/ddev/cmd/auth-ssh.go
stasadev: I'll try moving the logic to `RunSimpleContainer()`, but it may not be a simple review since it will require full refactoring in both places.

rfay: I don't think you should break your brain or mine or your fingers trying to force RunSimpleContainer to do the work if you're happy with where it's at right now. Your call on that.

cmd/ddev/cmd/auth-ssh.go
```diff
@@ -197,6 +184,86 @@ func getCertificateForPrivateKey(path string, name string) (string, string) {
...
+	exitCode := 0
+	if err != nil {
+		exitCode = 1
+		var code int
+		n, scanErr := fmt.Sscanf(err.Error(), "container run failed with exit code %d", &code)
+		if n == 1 && scanErr == nil {
+			exitCode = code
+		}
+	}
+	return exitCode, err
```
stasadev, line 264: I did this because docker doesn't return `exec.ExitError`, so I can't do:

```go
exitCode := 1
var exiterr *exec.ExitError
if errors.As(err, &exiterr) {
	exitCode = exiterr.ExitCode()
}
```

pkg/exec/exec.go
stasadev: The test is added.

pkg/dockerutil/dockerutils.go
```diff
@@ -1053,37 +1083,93 @@ func RunSimpleContainer(image string, name string, cmd []string, entrypoint []st
...
+
+		// For interactive containers, wait for I/O forwarding to complete
+		if config.AttachStdin {
+			// Close hijacked connection to signal EOF to goroutines
+			hijackedResp.Close()
+
+			// Wait for output forwarding to complete (input may still be running)
+			if outputDone != nil {
+				<-outputDone
+			}
```
stasadev, line 1137: I noticed that sometimes `ddev auth ssh` is stuck, i.e. it doesn't ask me to `Enter passphrase`.

I don't fully understand where the problem is, but Claude Code suggested to add `outputDone` for this.

stasadev:
If the tests are good, then it's ready for review.

rfay (approved):
Thanks for the test(s) and all the good work. All working, all passing.

PR #7516, opened 2025-08-07
chore(localdev): add path management to .envrc [skip ci]

rfay:
## The Issue

- #<issue number>

Add PATH management to direnv so it can automatically pick up the `ddev` built in the current directory

`direnv allow`

`which ddev` should point to the ddev built in the current environment

## Note

`direnv` support with .envrc was added in
* https://github.com/ddev/ddev/pull/6968


.envrc
```diff
@@ -71,15 +76,6 @@ if [[ ! -f "${NODE_ENV}/bin/textlint" ]]; then
...
-OS=$(uname -s | tr '[:upper:]' '[:lower:]')
-ARCH=$(uname -m)
-case "$ARCH" in
-  x86_64) ARCH=amd64 ;;
-  aarch64 | arm64) ARCH=arm64 ;;
-esac
-
-BIN_DIR="$PWD/.gotmp/bin/${OS}_${ARCH}"
-
+BIN_DIR="$PWD/.gotmp/bin/$(go env GOHOSTOS)_$(go env GOHOSTARCH)"
```
rfay, line 79: This is better, thanks!

rfay:
Thanks, we'll see how this works out for us. I won't be surprised if we have to adjust it. 

BTW there is a plugin for jetbrains tools to make it do envrc, you probably already have it. and VS Code has the feature by default I think.

PR #7517, opened 2025-08-07
chore: revert 3 github copilot commits that it (I) shouldn't have done [skip ci]

rfay:
## The Issue

I allowed Github Copilot to post a review.

It didn't post a review, but instead it did a commit to upstream/master. 

It did THREE commits:

db468268150949f8b413ef517860dee09539f870
7a2264532605e11a79d5e8941a422adb2da246b4
29a33931962982674fff206182d46c40c47b3a38

It didn't ask full permission for what it was doing. 

My trust level has gone to zero.

Do a revert of those. I could have rebased them away, but it's a protected branch, so we can live with them.


stasadev:
I think we need to review/update our branch protection rules.

rfay:
* https://github.com/ddev/ddev/issues/7518

PR #7520, opened 2025-08-10
build(direnv): Make direnv .envrc idempotent so it doesn't do all that work all the time

rfay:
## The Issue

Every time I switch into a ddev directory (a lot) I get several seconds of useless work being done:

```
direnv: loading ~/workspace/ddev/.envrc
Initializing python nodeenv...
Requirement already satisfied: nodeenv in ./.gotmp/env/python/lib/python3.13/site-packages (1.9.1)

[notice] A new release of pip is available: 25.1.1 -> 25.2
[notice] To update, run: pip install --upgrade pip
 * Environment already exists: .gotmp/env/node
./.envrc:47: .gotmp/env/node/bin/activate: No such file or directory
Installing markdownlint...

changed 104 packages in 2s

54 packages are looking for funding
  run `npm fund` for details
Installing linkspector...
⠼direnv: ([/opt/homebrew/bin/direnv export bash]) is taking a while to execute. Use CTRL-C to give up.

changed 231 packages in 5s

93 packages are looking for funding
  run `npm fund` for details
Installing textlint...

changed 298 packages in 6s

92 packages are looking for funding
  run `npm fund` for details
direnv: export +VIRTUAL_ENV +VIRTUAL_ENV_PROMPT ~PATH
```

Make it check to see if things are already properly installed and not do them in that case. 

Rendered at https://ddev--7520.org.readthedocs.build/en/7520/developers/building-contributing/#helpers-for-mkdocs-linkspector-markdownlint

Use it in .envrc
- [ ] cd into ddev dir see easy fast response
- [ ] rm -r .gotmp/env
- [ ] Switch out
- [ ] Switch back in, see installs happen
- [ ] Switch out
- [ ] Switch back in, see fast result

## Questions

Why are we even installing these things in the project instead of an external directory like ~/.cache? It's conceivably possible that some dependency of mkdocs could change, but it doesn't seem like it ever happens. I think that's the key question. And in that case... does it need to be done in .envrc at all?

## Comments

Assisted by ChatGPT

### DIFF
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -96,6 +96,23 @@ After you’re done testing, you can delete your downloaded executable, restart 
 rm ~/bin/ddev
 ```
 
+## Helpers for `mkdocs`, `linkspector`, `markdownlint`
+
+When testing the documentation for various problems, tools like `mkdocs` and `markdownlint` are required, and these may be installed differently on different systems. (This helps with `make staticrequired`, `make mkdocs-serve`, `make markdownlint`.)
+
+DDEV provides a project-level `.envrc` which with `direnv` can install these for the project.
+
+1. Install `direnv` with `brew install direnv` or `sudo apt-get update && sudo apt-get install -y direnv` or whatever technique is appropriate for your system.
+2. Hook `direnv` into your shell, see [docs](https://direnv.net/docs/hook.html).
+3. Create global configuration for `direnv` in `~/.config/direnv/direnv.toml` allowing it to be loaded without the `direnv allow` command, see  [docs](https://github.com/direnv/direnv/blob/master/man/direnv.toml.1.md), adjusting for your project path:
+
+```
+[global]
+strict_env = true
+[whitelist]
+exact = ["~/workspace/ddev/.envrc"]
+```
+
 ## Making Changes to DDEV Images
 
 If you need to make a change to one of the DDEV images, it will need to be built with a specific tag that’s updated in `pkg/versionconstants/versionconstants.go`.
--- a/docs/content/developers/building-contributing.md
+++ b/docs/content/developers/building-contributing.md
@@ -96,23 +96,6 @@ After you’re done testing, you can delete your downloaded executable, restart 
 rm ~/bin/ddev
 ```
 
-## Helpers for `mkdocs`, `linkspector`, `markdownlint`
-
-When testing the documentation for various problems, tools like `mkdocs` and `markdownlint` are required, and these may be installed differently on different systems. (This helps with `make staticrequired`, `make mkdocs-serve`, `make markdownlint`.)
-
-DDEV provides a project-level `.envrc` which with `direnv` can install these for the project.
-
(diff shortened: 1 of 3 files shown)


stasadev (changes requested):
> direnv: loading ~/workspace/ddev/.envrc
> Initializing python nodeenv...
> Requirement already satisfied: nodeenv in ./.gotmp/env/python/lib/python3.13/site-packages (1.9.1)

I don't see any of this, my `~/.bashrc` contains:

```
export DIRENV_LOG_FORMAT=
```

But without it, I see only this line:

```
direnv: loading ~/code/ddev/ddev/.envrc
```

---

If this change can improve that behavior, I'm fine with it.

P.S. The current PR adds some logic instead of replacing it.

rfay:
Does your system take several seconds in the switch?

What do you think about a global location for these tools, instead of project location?

stasadev:
> Does your system take several seconds in the switch?

No.

And I didn't try this PR yet because of added changes.

> What do you think about a global location for these tools, instead of project location?

If it can be done in a different way, I would like to see it.

rfay:
Actually, I would probably just make a script that does the install work into a global location in the PATH, and not update it very often, instead of tieing it to a specific repository/branch/directory.

rfay:
I cleaned up the duplication, maybe you can experiment with it now. 

However, I'd still rather this be a script in the scripts directory that sets up the venv on the system instead of per project. I don't remember seeing a project where that would be a problem. 

If you're OK with this one, let's get it in so I can continue to use direnv without pain, and then I'll pursue a system-venv setup as an alternate technique in an alternate PR.

stasadev (approved):
Works for me.

There is no `direnv` in our docs, I think a mention would be nice, somewhere in https://ddev.readthedocs.io/en/stable/developers/building-contributing/

docs/content/developers/building-contributing.md
```diff
@@ -96,6 +96,25 @@ After you’re done testing, you can delete your downloaded executable, restart
...
+When testing the documentation for various problems, tools like `mkdocs` and `markdownlint` are required, and these may be installed differently on different systems. (This helps with `make staticrequired`, `make mkdocs-serve`, `make markdownlint`.)
+
+DDEV provides a project-level `.envrc` which with `direnv` can install these for the project.
+
+1. Install `direnv` with `brew install direnv` or `sudo apt-get update && sudo apt-get install -y direnv` or whatever technique is appropriate for your system.
+2. Hook `direnv` into your shell, see [docs](https://direnv.net/docs/hook.html).
+3. Create global configuration for `direnv` in `~/.config/direnv/direnv.toml` allowing it to be loaded without the `direnv allow` command, see  [docs](https://github.com/direnv/direnv/blob/master/man/direnv.toml.1.md), for example:
+
+    ```toml
+   [global]
```
stasadev:
```suggestion
    [global]
```

rfay:
Added docs.

stasadev (commented):
This looks good, but I found out that we already have docs here:

- https://ddev--7520.org.readthedocs.build/en/7520/developers/testing-docs/#preview-changes
- https://ddev--7520.org.readthedocs.build/en/7520/developers/testing-docs/#check-markdown-for-errors

I would either move it there, or add crosslinks.

PR #7523, opened 2025-08-13
feat: support add-ons written primarily in PHP, fixes #7316

rfay:
## The Issue

- #7316

Support add-ons written in PHP.

PR #7528, opened 2025-08-14
fix(testddevexportdb): postgres:14 output dump statement was suddenly more than 80 characters from end

rfay:
## The Issue

TestDdevExportDB started failing today on HEAD.

It did an export of the database, and then checked to see that "ump complete" was in the last "line", which it turned out was assumed to be 80 characters. I don't know what the \unrestrict might mean.

But as of today, that dump seemed to have exceeded 80 characters:
```
 ALTER TABLE ONLY public.users
      ADD CONSTRAINT users__user_field__uuid__value__key UNIQUE (uuid);


  --
  -- PostgreSQL database dump complete
  --

  \unrestrict 9xW0ONzgZT8GfP5CfMcEmH3BHvMExf5nWzLJEY2XQaFT2mTpzYbYbJ665LBa4aF

```

* the `readLastLine()` was silly in its wrong naming (didn't get last line) and in the fact that it didn't take an arg for how much to look at. (This function is only used here)
* The TestDdevExportDB now looks at 120 characters using the updated function.
* This is tests only, and should be harmless.

I think this affects only tests, and when it passes we should be fine. 

The one (big?) question is... why would this change today?

### DIFF
--- a/pkg/ddevapp/ddevapp_test.go
+++ b/pkg/ddevapp/ddevapp_test.go
@@ -1866,7 +1866,7 @@ func TestDdevExportDB(t *testing.T) {
 		f, err := os.Stat("tmp/users1.sql")
 		assert.NoError(err)
 		assert.Greater(f.Size(), int64(2000))
-		l, err := readLastLine("tmp/users1.sql")
+		l, err := readFileTail("tmp/users1.sql", 120)
 		assert.NoError(err)
 		assert.Contains(l, "ump complete")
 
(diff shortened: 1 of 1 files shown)


rfay:
I cherry-picked this onto
* https://github.com/ddev/ddev/pull/7523

To make the tests on that one work

rfay:
Thanks, I imagine Trixie is the change. Do you have any idea what that trailing stuff in the dump means?  It may have been there in the past but just shorter or something. I didn't look.
```
  \unrestrict 9xW0ONzgZT8GfP5CfMcEmH3BHvMExf5nWzLJEY2XQaFT2mTpzYbYbJ665LBa4aF
```

stasadev:
> Do you have any idea what that trailing stuff in the dump means?

No, I don't understand what this is either.

PR #7531, opened 2025-08-15
fix: make ddev describe work correctly with new Docker Desktop, for docker/for-mac#7742

rfay:
## The Issue

- docker/for-mac#7742

Docker Desktop changed their internal inspect format, breaking `ddev describe` and TestCmdDescribe

Use the (appropriate) ExposedPorts to get the full set of exposed ports, instead of inferring it from Ports.

This affects only `ddev describe` and any other usages of app.Describe(), probably unlikely. It works on all the other Docker providers without problems.

We probably can't update our Docker Desktop runners (macOS and Windows) until this goes in, so manual testing is required with DD.

With Docker Desktop 4.44.0:

1. Use `docker-compose.extra.yaml` like this and `ddev describe` should show the 11111 ports etc.
```
services:
  web:
    expose:
      - 11111/tcp
      - 11112/tcp
      - 11113
```
2. run `TestCmdDescribe` and it should succeed 

## Automated Testing Overview

No changes required as the original tests caught this problem in the first place.

### DIFF
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -329,12 +329,19 @@ func (app *DdevApp) Describe(short bool) (map[string]interface{}, error) {
 
 		var exposedPrivatePorts []int
 		exposedPublicPorts := make(map[int]int)
+
+		// Get all exposed ports from container config (works regardless of Docker Desktop changes)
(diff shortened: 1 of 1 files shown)


rfay:
Now that this is pulled we can upgrade Docker Desktop on buildkite runners. I'll try to do this when not much testing is going on.

PR #7535, opened 2025-08-18
build(webserver): add `*_PROXY` and `DDEV_*` env for sudo; testing docs, fixes #7413

### LINKED ISSUE #7413: Make some of the `web` container env variables available to the `root` user

Compare the output for:
- `ddev exec env`
- `ddev exec sudo env`

Environment variables from here are not available to the `root` user:

https://github.com/ddev/ddev/blob/4a01fc8e0c5064f0abddd291ff83f2746f37329f/pkg/ddevapp/app_compose_template.yaml#L187-L258

I always wondered why.

### Describe your solution

I found out that this behavior is controlled by `/etc/sudoers`.

For example, proxy variables can be forwarded with:

```bash
$ cat /etc/sudoers.d/05_proxy
Defaults env_keep += "*_proxy *_PROXY"
```

Ref: https://wiki.archlinux.org/title/Proxy_server#Keep_proxy_through_sudo

I think we should forward all the `DDEV_*`, `IS_DDEV_PROJECT`, `TZ`, `START_SCRIPT_TIMEOUT` in addition to the proxy, and possibly more.

I see at least two use cases:

- Running `ddev exec sudo apt-get update` under proxy will work.
- Testing some internal DDEV scripts not inside the `Dockerfile` will be easier:
  ```
  $ ddev exec sudo mariadb-client-install.sh
  /usr/local/bin/mariadb-client-install.sh: line 10: DDEV_DATABASE: unbound variable
  
  Failed to execute command `sudo mariadb-client-install.sh`: exit status 1
  ```

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- #7413

Makes env available for `root` in ddev-webserver:

- `*_proxy`
- `*_PROXY`
- `DDEV_*`
- `IS_DDEV_PROJECT`
- `PHP_DEFAULT_VERSION`
- `START_SCRIPT_TIMEOUT`

Adds docs for testing HTTP proxy.

See https://ddev--7535.org.readthedocs.build/en/7535/developers/http-proxy-test-environments/

Compare the output for:

```
ddev exec sudo env
```

### DIFF
--- a/docs/content/developers/http-proxy-test-environments.md
+++ b/docs/content/developers/http-proxy-test-environments.md
@@ -5,6 +5,9 @@ search:
 
 # HTTP Proxy Test Environments
 
+!!!tip "Advanced: SSL Interception Testing"
+    For testing SSL interception scenarios (like corporate VPN/firewall environments), see [Network Test Environments: Packet-Inspection VPN Simulation](network-test-environments.md). This complements the basic HTTP proxy testing covered here.
+
 This guide demonstrates how to set up HTTP proxy environments for testing DDEV behavior with proxy servers. This is useful for testing corporate environments that route traffic through HTTP proxies, simulating real-world network configurations where all traffic must pass through proxy servers.
 
 ## Setup Overview
--- a/docs/content/developers/network-test-environments.md
+++ b/docs/content/developers/network-test-environments.md
@@ -4,6 +4,9 @@ search:
 ---
 # Network Test Environments: Packet-Inspection VPN Simulation
 
+!!!tip "Basic HTTP Proxy Testing"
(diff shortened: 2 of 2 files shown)


stasadev:
> Didn't I already have a writeup about this somewhere?

There is https://ddev.readthedocs.io/en/stable/developers/network-test-environments/

But my docs are different, I don't set up any certificates, I just check if the traffic is going through the proxy.

rfay:
Ah, I couldn't find that :) 

I guess they need links to each other? Or consider other consolidation.

stasadev:
> I guess they need links to each other? Or consider other consolidation.

Good idea, added.

PR #7536, opened 2025-08-18
refactor(docker): suppress any output (stdout, stderr) from docker/cli, for #7189

stasadev:
## The Issue

- #7189

Ignores `stdout` from `docker/cli` in addition to ignoring `stderr`.

`docker/cli` does not produce `stdout` for the functionality we use, but ignoring it prevents breakage from random output in the future.

Custom writers instead of `io.Discard` can be used, but we don't need them, because our main usage is Docker API, not Docker CLI.

### DIFF
--- a/pkg/dockerutil/dockerutils.go
+++ b/pkg/dockerutil/dockerutils.go
@@ -201,6 +201,8 @@ func GetDockerClient() (context.Context, dockerClient.APIClient) {
 		DockerCtx = context.Background()
 		// Set the Docker CLI version for User-Agent header
 		dockerCliVersion.Version = "ddev-" + versionconstants.DdevVersion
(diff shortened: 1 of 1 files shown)


stasadev:
I tested it by modifying `vendor`:

```go
// DockerEndpoint returns the current docker endpoint
func (cli *DockerCli) DockerEndpoint() docker.Endpoint {
	...
	_, _ = fmt.Fprintln(cli.Out(), "testing")
	...
}
```

`WithErrorStream` didn't supress this output, but `WithCombinedStreams` did.

The only place where I see `fmt.Fprintln(cli.Out()` is used: https://github.com/ddev/ddev/blob/main/vendor/github.com/docker/cli/cli/command/registry.go

PR #7540, opened 2025-08-20
docs: fiddle with sponsorship title [skip ci]

rfay:
## The Issue

I didn't like the 1-space difference between the two lines of the sponsorship update.

image

Remove a space.

This could have been done by changing the source data for the second line, but we might as well do this now.

There are probably other ways we could improve this, probably the two hearts aren't necessary. But we can change the bulk of that without changing DDEV.


rfay:
Smallest PR in history?

PR #7546, opened 2025-08-21
feat: add DDEV_ env vars to commands

hanoii:
I normally work inside the shell in the web container, but a lot of what I ship within the projects I work with other team members run them from the host. I have previously relied on other techniques to "detect" wether a particular script was run from the shell inside the container or from the host, but I figured having a way to detect that univocally coming from ddev might be a good addition.

The PR is simple and looks safe.

Wanted to get your opinion on this.

To test you can run:

- `ddev exec set | grep -i DDEV_EXEC`
- (create a custom host and web command with just `set`)  and run `ddev custom-web-command | | grep -iE "DDEV_COMMAND|DDEV_EXEC"`

### DIFF
--- a/cmd/ddev/cmd/commands.go
+++ b/cmd/ddev/cmd/commands.go
@@ -460,6 +460,8 @@ func makeHostCmd(app *ddevapp.DdevApp, fullPath, name string, mutagenSync bool) 
 			_ = os.Setenv("DDEV_PROJECT_STATUS", "")
 		}
 
+		os.Setenv("DDEV_COMMAND", name)
+
 		osArgs := []string{}
 		if len(os.Args) > 2 {
 			osArgs = os.Args[2:]
@@ -519,6 +521,7 @@ func makeContainerCmd(app *ddevapp.DdevApp, fullPath, name, service string, exec
 			Dir:       app.GetWorkingDir(s, ""),
 			Tty:       isatty.IsTerminal(os.Stdin.Fd()),
 			NoCapture: true,
+			Env:       []string{"DDEV_COMMAND=" + name},
 		}
 		if relative {
 			opts.Dir = path.Join(opts.Dir, app.GetRelativeWorkingDirectory())
--- a/pkg/ddevapp/ddevapp.go
+++ b/pkg/ddevapp/ddevapp.go
@@ -2177,6 +2177,7 @@ func (app *DdevApp) Exec(opts *ExecOpts) (string, string, error) {
 		}
 	}
 
+	baseComposeExecCmd = append(baseComposeExecCmd, "-e", "DDEV_EXEC=true")
 	baseComposeExecCmd = append(baseComposeExecCmd, opts.Service)
 
 	// Cases to handle


rfay:
$IS_DDEV_PROJECT is set inside the container always.

hanoii:
but `IS_DDEV_PROJECT` is set regardless of if a command is run within the container or from the host with `ddev command`.

rfay:
I guess you should give examples of where you mean.

hanoii:
Not sure what to add to the description.

Imagine a `script.sh` that I need it to to de something different if run from within the container, by doing 
- `ddev ssh`
- `./script.sh`

And something else if run from within the host:
- `ddev exec script.sh`

Or if script was run from a custom container web command:
- `ddev script`

rfay:
If you're running `./script.sh` on the host, it will not have IS_DDEV_PROJECT.
If running it inside the container, it will have IS_DDEV_PROJECT=true

hanoii:
I am not running ./script.sh from the host, I am running ddev from the host that runs  the script

- `ddev exec ./script.sh` which runs it in the container. 


```
ddev exec set | grep IS_DDEV_PROJECT
IS_DDEV_PROJECT=true
```

This PR adds `DDEV_EXEC=true` for that case

And 

if "script" happens to also be a custom web command

a `DDEV_COMMAND=script`

hanoii:
As a concrete example I use [ahoy](https://github.com/ahoy-cli/ahoy)  a lot.

So if I am in the container (after doing a `ddev ssh`) and run `ahoy switch` it does something.

Others do:

`ddev ahoy switch` and I want to be able to do something slightly different in that case.

Edit: `ddev ahoy` is a container command (NOT a host command)

rfay:
Do you want `ddev exec script.sh` to behave differently than `ddev ssh` followed by `script.sh` ?

hanoii:
yes!

rfay:
My friend ChatGPT suggests several ways this could be done without changing DDEV, would any of these work for you?

--- 

# Alternatives to Modifying DDEV Core

## 1. Add Environment Vars in Custom Web Commands
Set `DDEV_COMMAND` and `DDEV_EXEC` inside each custom command script:

```bash
# .ddev/commands/web/cache-rebuild
#!/bin/bash
set -euo pipefail
export DDEV_COMMAND="$(basename "$0")"
export DDEV_EXEC=0   # custom web command, not direct exec
exec /usr/local/bin/your-script "$@"
```

---

## 2. Host-Side Wrapper for `ddev exec`
Create a shell function or alias to inject env vars when running `ddev exec`:

```bash
# ~/.bashrc or ~/.zshrc
dexec() {
  local name="$1"; shift
  ddev exec bash -lc "DDEV_EXEC=1 DDEV_COMMAND='${name}' \"${name}\" \"$@\""
}
```

Usage:
```bash
dexec env | grep -E 'DDEV_EXEC|DDEV_COMMAND'
dexec artisan migrate
```

---

## 3. Makefile Targets (Team-Friendly)
Centralize wrapper logic in a `Makefile`:

```makefile
exec = ddev exec bash -lc

.PHONY: cr
cr:
	$(exec) 'DDEV_EXEC=1 DDEV_COMMAND=cache-rebuild drush cr'

.PHONY: tests
tests:
	$(exec) 'DDEV_EXEC=1 DDEV_COMMAND=phpunit ./vendor/bin/phpunit -v'
```

---

## 4. Heuristic Detection Inside the Container
Infer whether the command came from `ddev exec` or `ddev ssh` by inspecting the parent process:

```bash
detect_ddev_invocation() {
  if [ -z "${DDEV_EXEC:-}" ]; then
    local ppid
    ppid="$(awk '/^PPid:/ {print $2}' /proc/$$/status 2>/dev/null || true)"
    local pcmd
    pcmd="$(tr '\0' ' ' </proc/${ppid}/cmdline 2>/dev/null || true)"

    case "${pcmd}" in
      *"bash -lc "*|*"sh -lc "*) export DDEV_EXEC=1 ;; # likely ddev exec
      *)                           export DDEV_EXEC=0 ;; # likely ddev ssh
    esac
  fi
}
```

---

## 5. Breadcrumb File Approach
Use a temporary marker file to signal invocation type:

```bash
# host
ddev exec bash -lc 'echo "myscript" >/run/ddev_command; /usr/local/bin/myscript; rm -f /run/ddev_command'
```

Inside the container:

```bash
if [ -f /run/ddev_command ]; then
  export DDEV_EXEC=1
  export DDEV_COMMAND="$(cat /run/ddev_command)"
fi
```

---

## 6. Use `$0` in Custom Web Commands
Derive the command name directly:

```bash
export DDEV_COMMAND="${DDEV_COMMAND:-$(basename "$0")}"
```

---

## 7. Project-Wide Static Env Injection
Use `web-environment:` in `.ddev/config.yaml` for static vars:

```yaml
web-environment:
  - IS_DDEV_PROJECT=true
  - PROJECT_NAME=${DDEV_PROJECT}
```

(Dynamic vars still need a wrapper.)

---

## Recommended Combo
- For **custom web commands** → set `DDEV_COMMAND` + `DDEV_EXEC=0`.  
- For **host-initiated runs** → use a `dexec` wrapper or Makefile targets.  
- For **fallbacks** → use heuristic detection.  

This gives you nearly the same functionality as the PR without touching DDEV itself.

hanoii:
I guess. 

1. Ok, works only for commands
2. No, not portable to other team members
3. Not sure I understood, do a `make switch`, don't really like it.
4. Sure, I am already doing my own heuristics.
5. Not portable, odd workaround
6. Like 1, works only for subcommands, if it works, I have my doubts.
7. As it said, no dynamic vars

I have ways of doing this, but like with most of the things I've contributed I think what would have made the more sense to me so that I don't have to, for instance, repeat my heuristics on every project. 

I remember another command that logs something different if run from the container or outside with ddev COMMAND.

stasadev:
`DDEV_EXECUTABLE` is set only on the host side, if that helps.

PR #7550, opened 2025-08-22
refractor: updated mkcert and certutil installation

michaelhiiva:
## The Issue

mkcert and certutil needs to be installed before using mkcert in debain and ubuntu linux

This adds installation instructions for how to install mkcert and its dependancies.  This makes it so that the requirments of mkcert are there before the next init step.

## Automated Testing Overview

This is only adding an installation step. No tests should be required for this.

## Release/Deployment Notes

No


rfay:
Thanks for taking the time to make this suggestion @michaelhiiva - If you can follow up saying why you thought it was necessary, we'll be happy to help. Maybe there's something weird about your Debian-based environment, or maybe you have an old DDEV version or something?

Closing for now, happy to continue the conversation here or elsewhere.

PR #7552, opened 2025-08-22
docs: Use docs.ddev.com instead of ddev.readthedocs.io

rfay:
## The Issue

I have long wanted our docs to appear on our own domain.

It turns out readthedocs has this feature with no effort at all: https://app.readthedocs.org/dashboard/ddev/domains/

It's turned on. It's set to "canonical" although we can turn that off any time.

Both https://ddev.readthedocs.io and https://docs.ddev.com work!

- [ ] Visit and navigate https://docs.ddev.com
- [ ] Visit and navigate https://ddev.readthedocs.io


mattstein:
I like this! My only concern was with redirects, but it looks like any URL segment at https://ddev.readthedocs.io/ is maintained just fine redirecting to https://docs.ddev.com/. 👍

I’m seeing the `cf-cache-status: HIT` header in responses and fast global pings, and I forget what analytics are available but they might point to any weirdness if it exists.

As long as you can push a documentation update and see it reflected on the site without requiring a cache clear, you should be all good! Otherwise you’ll either want to disable Cloudflare proxying or add a manual step to invalidate its cache for each documentation deployment.

rfay:
> Otherwise you’ll either want to disable Cloudflare proxying or add a manual step to invalidate its cache for each documentation deployment

Right now, Cloudflare is just DNS, and I haven't experimented with turning on proxying.

mattstein:
I guess they’re already using it then, so it doesn’t really matter!

tyler36 (approved):
This seems fine to me.

- I bounced around the "docs.ddev" site without issue.
- I tested some simple searches which also worked as expected

>  Visit and navigate https://ddev.readthedocs.io/
I wasn't able to test this as it redirects to https://docs.ddev.com/en/stable/ (which is the goal)

Issue #7539, opened 2025-08-20
Run docker buildx runners on native architecture for faster builds


stasadev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

1.  ARM64 build speed degraded (10 minutes on bookworm, 30 minutes on trixie) for `ddev-webserver` in:

    - #7492

2. Our QEMU workarounds:

    - #7070
    - #7073

### Describe your solution

See:
- [PSA: Don’t forget to run your buildx runners on native architecture for faster builds](https://www.reddit.com/r/docker/comments/1mtw0bh/psa_dont_forget_to_run_your_buildx_runners_on/)
- [docker.yml example](https://github.com/11notes/docker-distroless/blob/master/.github/workflows/docker.yml)

### Additional context

Related:

- #2621

rfay:
We can actually do this since GitHub has arm64 linux runners. I've always thought it might be complicated though. But what we have is complicated.

AkibaAT:
Super easy to use them, can choose ubuntu-24.04-arm or ubuntu-22.04-arm.
Just need to add an extra manifest merge step to bring the different outputs back together.

rfay:
It's easy to use the runners; the complexity of the push goes up because you're pushing two different builds to the same tag.

AkibaAT:
That's what the manifest merge step brings together again, so you end up with a multi-arch image like now.
Basically the build process needs to be changed to push tag-arch variants, then the manifest merge step takes those and builds the final tag multi-arch image again.

PR #7553, opened 2025-08-24
build: use native arm builder for building docker images

AkibaAT:
## The Issue

- #7539

### DIFF
--- a/.github/workflows/push-tagged-dbimage.yml
+++ b/.github/workflows/push-tagged-dbimage.yml
@@ -25,48 +25,94 @@ permissions:
   contents: read
 
 jobs:
-  push-tagged-dbimage:
-    name: "push tagged dbimage"
-    runs-on: ubuntu-24.04
+  build-db-arch:
+    name: build ${{ matrix.arch }} ${{ matrix.dbtype }}
     strategy:
-      matrix:
-        dbtype: [mariadb_5.5, mariadb_10.0, mariadb_10.1, mariadb_10.2, mariadb_10.3, mariadb_10.4, mariadb_10.5, mariadb_10.6, mariadb_10.7, mariadb_10.8, mariadb_10.11, mariadb_11.4, mariadb_11.8, mysql_5.5, mysql_5.6, mysql_5.7, mysql_8.0, mysql_8.4]
       fail-fast: false
-
+      matrix:
+        arch: [amd64, arm64]
+        dbtype: [mariadb_5.5, mariadb_10.0, mariadb_10.1, mariadb_10.2, mariadb_10.3, mariadb_10.4, mariadb_10.5, mariadb_10.6, mariadb_10.7, mariadb_10.8, mariadb_10.11, mariadb_11.4, mariadb_11.8, mysql_5.5, mysql_5.6, mysql_5.7, mysql_8.0, mysql_8.4]
+        exclude:
+          - arch: arm64
+            dbtype: mariadb_5.5
+          - arch: arm64
+            dbtype: mariadb_10.0
+          - arch: arm64
+            dbtype: mysql_5.5
+          - arch: arm64
+            dbtype: mysql_5.6
+    runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
     steps:
-    - name: Load 1password secret(s)
-      uses: 1password/load-secrets-action@v2
-      if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.owner.login == github.repository_owner }}
-      with:
-        export-env: true
-      env:
-        OP_SERVICE_ACCOUNT_TOKEN: "${{ secrets.PUSH_SERVICE_ACCOUNT_TOKEN }}"
-        DOCKERHUB_TOKEN: "op://push-secrets/DOCKERHUB_TOKEN/credential"
-
-    - uses: actions/checkout@v5
-
-    - name: Set up QEMU
-      uses: docker/setup-qemu-action@v3
-      with:
-        # special qemu version used due to many
-        # failures with "Cannot allocate memory" in
-        # apt-get install of arm64
-        image: tonistiigi/binfmt:qemu-v7.0.0-28
-
-    - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v3
-    - name: Login to DockerHub
-      uses: docker/login-action@v3
-      with:
-        username: ${{ vars.DOCKERHUB_USERNAME }}
-        password: ${{ env.DOCKERHUB_TOKEN }}
-    - name: Setup tmate session
-      uses: mxschmitt/action-tmate@v3
-      with:
-        limit-access-to-actor: true
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-      if: ${{ github.event_name == 'workflow_dispatch' && inputs.debug_enabled }}
-    - name: Push ${{  matrix.dbtype }}:${{ github.event.inputs.tag }}
-      run: |
-        cd "containers/ddev-dbserver"
-        make ${{ matrix.dbtype }} PUSH=true VERSION="${TAG}" DOCKER_ORG=${DOCKER_ORG}
+      - name: Load 1password secret(s)
+        uses: 1password/load-secrets-action@v2
+        if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.owner.login == github.repository_owner }}
+        with:
+          export-env: true
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: "${{ secrets.PUSH_SERVICE_ACCOUNT_TOKEN }}"
+          DOCKERHUB_TOKEN: "op://push-secrets/DOCKERHUB_TOKEN/credential"
+      - uses: actions/checkout@v5
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ vars.DOCKERHUB_USERNAME }}
+          password: ${{ env.DOCKERHUB_TOKEN }}
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        with:
+          limit-access-to-actor: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.debug_enabled }}
+      - name: Build and push single-arch db image
+        run: |
+          set -eu
+          cd containers/ddev-dbserver
+          base=${{ matrix.dbtype }}
+          target="${base}_${{ matrix.arch }}"
+          echo "Building target $target for arch ${{ matrix.arch }}"
+          make $target PUSH=true VERSION="${TAG}-${{ matrix.arch }}"
+      - name: Record digest
+        id: digest
+        run: |
+          # Attempt to record one digest for confirmation
+          img_name=${{ matrix.dbtype }}
+          dbtype=$(echo $img_name | cut -d_ -f1)
+          major=$(echo $img_name | cut -d_ -f2)
+          IMAGE=${DOCKER_ORG}/ddev-dbserver-${dbtype}-${major}:${TAG}-${{ matrix.arch }}
+          if docker buildx imagetools inspect $IMAGE >/dev/null 2>&1; then
+            dg=$(docker buildx imagetools inspect $IMAGE | awk -F: '/Digest/ {print $2; exit}')
+            echo "digest=$dg" >> $GITHUB_OUTPUT
+          fi
+  create-manifests:
+    name: create multi-arch db manifests
+    needs: build-db-arch
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Load 1password secret(s)
+        uses: 1password/load-secrets-action@v2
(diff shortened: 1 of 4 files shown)


rfay:
Wow, thank you!

I guess `make push` in the various Makefiles would still have to use QEMU since they're assuming single-builder?

AkibaAT:
> Wow, thank you!
> 
> I guess `make push` in the various Makefiles would still have to use QEMU since they're assuming single-builder?

None of the stages should require QEMU anymore. The images are built via the matrix runner, and then there's a final merge manifest step that runs on the regular AMD Ubuntu runner to merge the ARM & AMD build results together.

rfay:
You should be able to test this in your own repo and pushing to your own org on hub.docker.com. Yippee! 

DOCKER_ORG and DOCKERHUB_USERNAME ought to get it to point correctly. 

I guess the DOCKERHUB_TOKEN is harder, since we get that from 1password, but you can temporarily hack it in as a github secret instead of as 1Password secret.

rfay:
> None of the stages should require QEMU anymore. The images are built via the matrix runner, and then there's a final merge manifest step that runs on the regular AMD Ubuntu runner to merge the ARM & AMD build results together.

In testing we will occasionally do a `make push` (which uses the current architecture) or the various varieties. I would think we'd always want to be able to push an image without github, but I have to admit we almost always do it with github.

AkibaAT:
> > None of the stages should require QEMU anymore. The images are built via the matrix runner, and then there's a final merge manifest step that runs on the regular AMD Ubuntu runner to merge the ARM & AMD build results together.
> 
> In testing we will occasionally do a `make push` (which uses the current architecture) or the various varieties. I would think we'd always want to be able to push an image without github, but I have to admit we almost always do it with github.

The makefile still takes the _both variants to build both archs, in that case QEMU would need to have been set up beforehand. If multi-archs is not required for the specific usecase, then the Makefile now also accept _amd64 and _arm64 variants instead.

stasadev:
@AkibaAT,

You can test it by duplicating this branch at https://github.com/ddev-test/ddev, it's configured to push everything to https://hub.docker.com/u/ddevhq

rfay:
When you commit/push on this one, please use `[skip ci]` in the title for now, since it doesn't need to run any tests, thanks!

rfay (changes requested):
I updated the manual testing instructions
* Confused by 3 separate images (arm64, amd64, consolidated)
* There are 4 separate digests on the consolidated image
* The arm64 and amd64 images have both digests in them

I think all images have to be test-pushed for this, and they need to come out with the right names and digests.

Thanks for the great work on this!

rfay:
Please use the branch name for the image tag, so `20250824_akibaat_buildx_arm_builder`

AkibaAT:
> There are 4 separate digests on the consolidated image
> The arm64 and amd64 images have both digests in them

Fixed the bug for the non-DB images, those still built both platforms together, and then those results got merged, which is why the digests doubled up.

> Confused by 3 separate images (arm64, amd64, consolidated)

The -amd64 and -arm64 images are the build outputs from the intermediary and required for the consolidation step.
Cleanup could potentially be added, but as far as I could see, that's only possible via direct curl calls to the Docker Hub API. (i.e. curl -u $USERNAME:$PASSWORD -X "DELETE" https://cloud.docker.com/v2/repositories/$ORGANIZATION/$REPOSITORY/tags/$TAG/). Or maybe via https://github.com/docker/hub-tool, although that seems to just do the same internally as well.

rfay:
Awesome work. 

Now oddly the latest images are arm64 and amd64, https://hub.docker.com/r/ddevhq/ddev-dbserver-mariadb-10.4/tags - does that mean the consolidated image got pushed first and then got the early date and the others got added to it?

But on traefik, the consolidated one is newer, https://hub.docker.com/r/ddevhq/ddev-traefik-router/tags

Please do push all, thanks!

rfay:
I have seen other org's images where there were separate and combined images; it's not familiar with DDEV and not very useful, but I'm not sure how bad it is. Just 3 images when only one is needed.

AkibaAT:
> Please do push all, thanks!

There was a workflow error reading the resulting digest, I assume due to rate limiting for the user against Docker Hub. Now the workflow finished.

AkibaAT:
> I have seen other org's images where there were separate and combined images; it's not familiar with DDEV and not very useful, but I'm not sure how bad it is. Just 3 images when only one is needed.

Added a cleanup via the API calls mentioned earlier.

.github/workflows/push-tagged-dbimage.yml
```diff
@@ -25,48 +25,109 @@ permissions:
...
+          for v in $variants; do
+            img=${DOCKER_ORG}/ddev-dbserver-$v
+            if docker buildx imagetools inspect ${img}:${TAG}-amd64 >/dev/null 2>&1 && docker buildx imagetools inspect ${img}:${TAG}-arm64 >/dev/null 2>&1; then
+              docker buildx imagetools create -t ${img}:${TAG} ${img}:${TAG}-amd64 ${img}:${TAG}-arm64
+              if [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                docker buildx imagetools create -t ${img}:latest ${img}:${TAG}-amd64 ${img}:${TAG}-arm64
+              fi
+              # Clean up intermediary single-arch tags from remote registry
+              curl -s -X DELETE -H "Authorization: JWT $TOKEN" "https://hub.docker.com/v2/repositories/${DOCKER_ORG}/ddev-dbserver-$v/tags/${TAG}-amd64/" >/dev/null || true
+              curl -s -X DELETE -H "Authorization: JWT $TOKEN" "https://hub.docker.com/v2/repositories/${DOCKER_ORG}/ddev-dbserver-$v/tags/${TAG}-arm64/" >/dev/null || true
```
stasadev: I wonder if it's possible to use `docker buildx imagetools create` without pushing single arch images to the registry.

Edit: no, it's not possible.

stasadev (commented):
In addition, we need to test if the `latest` tag works fine for release images (when you push tags like `v1.x.x`)

.github/workflows/push-tagged-dbimage.yml
stasadev: Related:

- https://github.com/docker/cli/issues/3350

AkibaAT: Yeah, the build stage, being singe-arch, needs to push the image *somewhere*. It could be a different registry, an artifact, doesn't really matter, but it has to first be stored somewhere in between to then be merged together. And in the end, the layers have to be pushed to Docker Hub anyway.

For creating manifests of local images, they'd have to be local first, which they aren't. Unless the build stages themselves exported the images to tars, that the merge step then downloaded, imported, to then build a combined image before pushing. Which, while possible, is quite the overhead, also runtime wise.

One option, if you really wanted to remove the need for the cleanup step while keeping the process optimal, would be to push the intermediary tags to its own building namespace, which could even be set to private.

.github/workflows/push-tagged-image.yml
```diff
@@ -38,56 +38,93 @@ permissions:
...
+          password: ${{ env.DOCKERHUB_TOKEN }}
+      - name: Create and push manifest (non-db)
+        if: github.event.inputs.image != 'ddev-dbserver'
+        run: |
+          set -eu
+          IMAGE=${DOCKER_ORG}/${{ github.event.inputs.image }}
+          TAG=${{ github.event.inputs.tag }}
+          docker buildx imagetools create -t ${IMAGE}:${TAG} ${IMAGE}:${TAG}-amd64 ${IMAGE}:${TAG}-arm64
+          if [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            docker buildx imagetools create -t ${IMAGE}:latest ${IMAGE}:${TAG}-amd64 ${IMAGE}:${TAG}-arm64
```
stasadev: The `latest` should look like this to have the same hash (not tested):

```suggestion
            docker buildx imagetools create -t ${IMAGE}:latest ${IMAGE}:${TAG}
```

.github/workflows/push-tagged-dbimage.yml
```diff
@@ -25,48 +25,109 @@ permissions:
...
+          TAG=${{ github.event.inputs.tag }}
+          # Get Docker Hub token for cleanup
+          TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username":"${{ vars.DOCKERHUB_USERNAME }}","password":"${{ env.DOCKERHUB_TOKEN }}"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
+          variants="mariadb-10.3 mariadb-10.4 mariadb-10.5 mariadb-10.6 mariadb-10.7 mariadb-10.8 mariadb-10.11 mariadb-11.4 mariadb-11.8 mysql-5.7 mysql-8.0 mysql-8.4"
+          for v in $variants; do
+            img=${DOCKER_ORG}/ddev-dbserver-$v
+            if docker buildx imagetools inspect ${img}:${TAG}-amd64 >/dev/null 2>&1 && docker buildx imagetools inspect ${img}:${TAG}-arm64 >/dev/null 2>&1; then
+              docker buildx imagetools create -t ${img}:${TAG} ${img}:${TAG}-amd64 ${img}:${TAG}-arm64
+              if [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                docker buildx imagetools create -t ${img}:latest ${img}:${TAG}-amd64 ${img}:${TAG}-arm64
```
stasadev: The `latest` should look like this to have the same hash (not tested):

```suggestion
                docker buildx imagetools create -t ${img}:latest ${img}:${TAG}
```

.github/workflows/push-tagged-dbimage.yml
AkibaAT: That would also keep the single arch tags from appearing temporarily while building, and it would prevent tags from showing up before all the builds succeeded.

stasadev: > One option, if you really wanted to remove the need for the cleanup step while keeping the process optimal, would be to push the intermediary tags to its own building namespace, which could even be set to private.

This sounds like a good idea.

Do you mean adding a new image to the registry, like "ddev-temp-builder", and pushing all single-arch images there, for example:

- `ddev/ddev-temp-builder:ddev-dbserver-mariadb-10.7-v1.24.7-arm64`
- `ddev/ddev-temp-builder:ddev-dbserver-mariadb-10.7-v1.24.7-amd64`
- `ddev/ddev-temp-builder:ddev-traefik-router-v1.24.7-arm64`
- `ddev/ddev-temp-builder:ddev-traefik-router-v1.24.7-arm64`
- ...

AkibaAT: Yes, exactly. This would keep the build from even temporarily filling up the other namespaces with tags that aren't required there.

rfay:
The "latest" tag works: https://hub.docker.com/repository/docker/ddevhq/ddev-webserver/tags

The ddev-php-base image takes only like 4 minutes to build!!!

rfay:
`ddev-gitpod-base` doesn't work (because it was never built with arm64 before).  It's (almost) obsolete, but still the time might come to do an amd64-only build. 

Related:
* https://github.com/ddev/ddev/pull/6977

rfay:
I see that for db images we have `Build and push non-db image (single-arch)`, but I guess we need that for regular images too. Or a separate workflow, which would be fine.

rfay:
@AkibaAT want to take a look at goreleaser in this context? https://carlosbecker.com/posts/goreleaser-v2.12 ? They're always getting ahead of us!

rfay:
Rebased

rfay (changes requested):
I appreciate you exploring the Chocolatey implementation but it's hard to test and I don't think you figured out a way to test it. I would love to use their implementation. But could you put that in a different PR please?

AkibaAT:
Rebased and fixed conditions up for required single arch images.

rfay:
When you do take on the chocolatey stuff we can figure out how to add a new project for testing there like we do for AUR.

rfay:
Rebased. Seems like there's still a comment from @stasadev that is not addressed. This is a good time to try to get this in.

stasadev:
This week I'm going to do a review and experiment with this PR.

rfay:
Such great exploration! I never heard of local registry before.

stasadev:
I looked at the result of multiarch here https://github.com/11notes/docker-distroless/blob/master/.github/workflows/docker.yml

https://hub.docker.com/r/11notes/personal-medical-records-keeper/tags
Nothing is delete here, but what's interesting: Type Artifact.

I can't even pull that. This is probably what we need to have.

```
$ docker pull 11notes/personal-medical-records-keeper:buildcache-amd64
buildcache-amd64: Pulling from 11notes/personal-medical-records-keeper
unsupported media type application/vnd.buildkit.cacheconfig.v0
```

image

---

Edit:

Hmm, no, that's something else.
Because I can still pull single-arch tags from there.

stasadev (approved):
Incredible work @AkibaAT !

---

I'll do one more rebase (with push for db images) after #7608 comes in (to avoild conflicts).

rfay:
Rebased, needs image pushes and then test run.

stasadev:
I pushed new db images.

rfay:
It doesn't need all the others? They haven't changed?

stasadev:
There are no changes elsewhere.

Issue #7491, opened 2025-07-28
Add a way to search for add-ons by keyword


GuySartorelli:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Often if I'm looking for an add-on for a specific service, I have to grep to find them (e.g. `ddev add-on list --all | grep memcached`) because most of the add-ons aren't related to what I'm looking for.

### Describe your solution

Allow an argument (or multiple arguments) to the `ddev add-on list` command for search terms.
If any of the search terms is found in the name or description of an add-on, display it. Don't display unrelated add-ons.
If no add-ons have the search term(s), display a helpful message like "no repositories found with those search terms"

For example I could use `ddev add-on list --all memcached` to find any add-ons related to memcached. Or `ddev add-on list --installed memcached` to list any _installed_ add-ons which are related to memcached.

### Describe alternatives

A new `ddev add-on search` command could be added - though the only reason I can think for doing it this way is if you want to reserve arguments to the list command for some future functionality.

There is also the registry online - but I don't wanna go fish out the URL for that and use a browser-based solution when I'm already in my terminal working away.

rfay:
Does https://addons.ddev.com help you adequately?

ChloeSartorelli:
As mentioned in the issue, if I'm already working in my terminal, a terminal-based solution is preferred. I use `grep` instead of going to the addons registry online.

rfay:
I wonder if the `ddev add-on search` should actually use the addons json as it's source.

ChloeSartorelli:
I thought the plan was for the whole `ddev add-on` system to eventually use the `addons.ddev.com` registry to get its data.
I would personally recommend against having one system for getting results for `ddev add-on search` and a separate system for getting results for `ddev add-on list`.

Though I may have misunderstood what you meant.

rfay:
> Though I may have misunderstood what you meant.

Probably I misunderstood :)

PR #7554, opened 2025-08-24
feat: add add-on search subcommand

AkibaAT:
## The Issue

- #7491

Implemented a new search subcommand

`ddev add-on search "solr engine"`

## Automated Testing Overview

Added new TestCmdAddonSearch


nickchomey:
Something like this was suggested and implemented in this PR but ultimately abandonned. It's still not evident to me why. It also briefly used fuzzy search, but a commit removed that. Again, it's not evident to me why fuzzy wouldn't be desirable.

https://github.com/ddev/ddev/pull/6353

Some fuzzy search libraries. 

https://github.com/sahilm/fuzzy
https://github.com/lithammer/fuzzysearch
https://github.com/raphamorim/fuzzy

The latter is newer and fancier and compares itself to the other two. Not sure which would be appropriate here.

rfay:
Rebased and fixed gofmt problem

tyler36 (changes requested):
This works well.

However, I think it should default to `--all`, otherwise its not obvious if there are other addons, or that we are only searching a subset of addon. 

## Example workflow

I want a "browsersync" addon

```shell
$ ddev add-on search browsersync
┌───────────────────────┬────────────────────────────────────────────────────┐
│ ADD-ON                │ DESCRIPTION                                        │
├───────────────────────┼────────────────────────────────────────────────────┤
│ ddev/ddev-browsersync │ Live-reload and HTTPS auto-refresh on file changes │
│                       │ for DDEV*                                          │
└───────────────────────┴────────────────────────────────────────────────────┘
1 repositories found matching 'browsersync'. Add-ons marked with '*' are officially maintained DDEV add-ons.
```

Good, there's 1 addon. And its marked with "*" so its official. (wonder if there are other unofficial ones and how do I find them?)

I want a "vscode" addon:

```
$ ddev add-on search vscode 
No add-ons found matching 'vscode'
```

Ok, none listed, official or otherwise. But I know I have an addon ....

```shell
$ ddev add-on search vscode --all
┌──────────────────────────────────┬─────────────────────────────────┐
│ ADD-ON                           │ DESCRIPTION                     │
├──────────────────────────────────┼─────────────────────────────────┤
│ tyler36/ddev-vscode-devcontainer │ Open a DDEV container in VSCode │
└──────────────────────────────────┴─────────────────────────────────┘
1 repositories found matching 'vscode'. Add-ons marked with '*' are officially maintained DDEV add-ons.
```

## Suggestion

- Default to `--all` search.  (Personally prefer this option)
- Alternatively, update the non/found message to include something like to:

    ```
    No add-ons found matching 'vscode'. Use "--all" to search official and unofficial add-ons.
    ```

## Test

- ✅  Find official addon: `ddev addon search browsersync`
- ✅  Find unofficial addon: `ddev addon search qr --all`
- ✅  Find unofficial addon via description: `ddev addon search queries --all`
- ✅  Displays message when no addons found: `ddev addon search meatball --all`
- ✅  Displays relevant help message: `ddev addon search --all`

nickchomey:
I'd like to reiterate that this feature seems half-baked, at best, without fuzzy search. As it is, people have to already know exactly what they're looking for, and then spell it precisely correctly (or at least a subset of it). 

Fuzzy search would account for typos, slight differences (vs code vs vscode), and overall help with discoverability. 

It seems like a simple enough thing to add, no?

AkibaAT:
> ## Suggestion
> * Default to `--all` search.  (Personally prefer this option)
> * Alternatively, update the non/found message to include something like to:
>   ```
>   No add-ons found matching 'vscode'. Use "--all" to search official and unofficial add-ons.
>   ```

The search command is using the same default as the list command because I'd expect to find the same behaviour in both. It also has the same issue of not giving any hint about other addons, and the --all flag.

@rfay What do you think? Default to --all true (best replacing it with an --official flag to limit to those), or behaviour like now, but just with an added comment about the flag? For both the list, as well as the search commands.

rfay:
Rebased

stasadev (commented):
Previously, we had one `ddev get` command, and there was no choice between using different flags for different actions.

But now, with subcommands (thanks to @GuySartorelli), we can have different behavior.

I agree with @tyler36, the `search` subcommand is expected to search everywhere, so the `--all` flag is not needed here.

> The search command is using the same default as the list command because I'd expect to find the same behaviour in both. It also has the same issue of not giving any hint about other addons, and the `--all` flag.

@AkibaAT, what about using this in `addon-list.go`?

```diff
-func renderRepositoryList(repos []*github.Repository) string {
+func renderRepositoryList(repos []*github.Repository, officialOnly bool) string {
```

```diff
-return out.String() + fmt.Sprintf("%d repositories found. Add-ons marked with '*' are officially maintained DDEV add-ons.", len(repos))
+message := out.String() + fmt.Sprintf("%d repositories found. Add-ons marked with '*' are officially maintained DDEV add-ons.", len(repos))
+if officialOnly {
+	message = message + "\nTo view all add-ons, use `ddev add-on list --all`"
+}
+return message
```

rfay:
I agree as well that people would expect to search all add-ons when using `ddev add-on search`. 

One thing that has changed since the beginning of add-ons is that these days there are a number of popular add-ons that don't really need to move to the "official" ddev namespace, but are still valuable to particular groups of users.

AkibaAT:
Rebased and dropped --all from list & search

ChloeSartorelli:
Why are you dropping `--all` from the `list` command? From what I can see in the discussion above, the suggestion was to not require `--all` for the new `search` command with no mention of removing it from the existing `list` command? It's possible I misunderstood that though.

docs/content/users/extend/using-add-ons.md
```diff
@@ -30,18 +30,12 @@ Use [DDEV Add-on Registry](https://addons.ddev.com/) to discover, explore, and l
 
 ### Command Line Discovery
```
ChloeSartorelli, line 31: Given this is about discovery of add-ons, it probably makes sense to add an example using the new `ddev add-on search` command here.

cmd/ddev/cmd/addon-search.go
```diff
@@ -0,0 +1,106 @@
...
+
+// AddonSearchCmd is the "ddev add-on search" command
+var AddonSearchCmd = &cobra.Command{
+	Use:   "search <search-term>",
+	Args:  cobra.ExactArgs(1),
+	Short: "Search available DDEV add-ons",
+	Long:  `Search available DDEV add-ons by name or description.`,
+	Example: `ddev add-on search redis
+ddev add-on search database
+ddev add-on search "solr search"`,
```
ChloeSartorelli: Can we allow sending multiple args instead? e.g. `ddev add-on search solr search` instead of having to put quotes around it?

This would also allow for potential future enhancements where each argument is a separate search term, e.g. `ddev add-on search solr search` could find any instance of "solr" and any instance of "search", while `ddev add-on search "solr search"` might explicitly only find instances of the full string "solr search" and not "solr" or "search" individually.

AkibaAT:
> Why are you dropping `--all` from the `list` command? From what I can see in the discussion above, the suggestion was to not require `--all` for the new `search` command with no mention of removing it from the existing `list` command? It's possible I misunderstood that though.

I understood the last input as it no longer making sense to filter by "official" addons, and instead to always show all.
If I misunderstood, it is in its own commit.

docs/content/users/extend/using-add-ons.md
AkibaAT: I added it to the general commands documentation, but did forget about adding it to the specific add-ons one.

AkibaAT: Added

cmd/ddev/cmd/addon-search.go
AkibaAT: Added, they currently behave the same.

cmd/ddev/cmd/addon-list.go
```diff
@@ -133,12 +132,9 @@ func renderRepositoryList(repos []*github.Repository) string {
 }
 
 func init() {
-	AddonListCmd.Flags().Bool("all", false, `List unofficial DDEV add-ons for in addition to the official ones`)
```
stasadev, line 136: Let's make it hidden instead of removing it.

We can list it for removal in future releases.

BTW, we still have this:
```
ddev get --list --all
```

docs/content/users/usage/commands.md
```diff
@@ -173,16 +172,33 @@ Example:
 # List official add-ons
```
stasadev:
```suggestion
# List all available add-ons
```

cmd/ddev/cmd/addon-list.go
ChloeSartorelli: If making it hidden, will the default behaviour be to show all? If so, is hiding the `--all` flag just to not make an error happen when it's used (even though it now will do nothing)?

Should a `--official` flag be added to list only official add-ons?

IMO removing or hiding this flag opens a bit of a can of worms - might be worth just keeping it for now and opening a new issue for discussion instead so this PR can be merged without affecting the list functionality.

stasadev: There are a few hidden flags in our codebase that do nothing (for compatibility reasons).

Actually we can mark it as deprecated instead of making it hidden.

stasadev: > Should a `--official` flag be added to list only official add-ons?

I'm not sure.

---

Side note: It actually feels slow when DDEV makes all these requests to GitHub (and you hit the limit pretty quickly without `DDEV_GITHUB_TOKEN`.)

It would make sense to use `https://addons.ddev.com/addons.json` (with local cache) for add-on operations (using GitHub as fallback.)

But this is beyond the scope of this PR.

docs/content/users/usage/commands.md
AkibaAT: Done

cmd/ddev/cmd/addon-list.go
AkibaAT: Marked as deprecated like other existing flags

cmd/ddev/cmd/addon-search.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	t.Render()
+
+	return out.String() + fmt.Sprintf("%d repositories found matching '%s'. Add-ons marked with '*' are officially maintained DDEV add-ons.", len(repos), searchTerm)
+}
+
+func init() {
+	AddonSearchCmd.Flags().Bool("all", false, `Search unofficial DDEV add-ons in addition to the official ones`)
+	_ = AddonSearchCmd.Flags().MarkDeprecated("all", "this flag no longer has any effect. All add-ons are searched by default.")
```
ChloeSartorelli:
```suggestion
```
There's no need to add this flag here.
The reason it's being kept and marked as deprecated in `add-on list` is because _it already existed there_ so people will be used to using it.

Since this is a brand new command, nobody has ever used it before so they won't be used to using the `--all` flag. There's no reason to add it unless it has some functionality (which it doesn't).

cmd/ddev/cmd/get.go
```diff
@@ -89,6 +88,7 @@ ddev get --remove ddev-someaddonname
 func init() {
 	Get.Flags().Bool("list", true, `List available add-ons for 'ddev get'`)
 	Get.Flags().Bool("all", true, `List unofficial add-ons for 'ddev get' in addition to the official ones`)
+	_ = Get.Flags().MarkDeprecated("all", "this flag no longer has any effect. All add-ons are shown by default.")
```
ChloeSartorelli: I recommend either _not_ deprecating here (this whole command is deprecated anyway), or else removing lines 38-40 (i.e. don't pass `--all` through since it doesn't do anything anymore).

cmd/ddev/cmd/addon-search.go
AkibaAT: Changed

cmd/ddev/cmd/get.go
AkibaAT: Removing that deprecation notice again makes more sense to me than the alternative, for the reason given that the whole command is outdated anyway.

PR #7563, opened 2025-08-26
docs: emphasize the Orbstack free version is personal use only

mxr576:


rfay:
Thanks again, I don't think it's our job to figure this out for people.

mxr576:
I opened this PR because DDEV promotes OrbStack as the recommended solution for running DDEV on macOS, but I think the warnings about it being "not free for commercial use" are not prominent enough. It is easy for users to miss this detail, and one of my colleagues nearly overlooked it as well.

In my view, if DDEV promotes OrbStack as the recommended tool, it should also put stronger emphasis on the commercial limitations so that users are clearly aware of them.

image

mxr576:
> Users need to commit to the license of all of these (including Docker Desktop, which you don't seem to be editing here).

That could be also sorted out in this PR

rfay:
Please tell us about your experience here. I assume you're using macOS. Did you feel tricked or deceived or something? Help us to understand the context.

mxr576:
No, I am a happy Linux user :) so I was already more than satisfied with the Linux documentation.

The first thing that caught my attention was:

> It is easy for users to miss this detail, and one of my colleagues nearly overlooked it as well.

Later, when I started working on a DDEV migration guide for my colleagues, I realized that some of them use macOS for daily work. That is when I noticed the need to highlight that they should not just follow the macOS installation steps blindly, but instead read carefully and choose a company-approved alternative to OrbStack.

If we did not have developers on macOS, our install guide would simply be:

```
## Install

Follow the installation steps on https://ddev.com/get-started specific
to your operating system.
```

rfay:
Here's what ddev.com/get-started says:

image

mxr576:
Yes, that is the page which I also provided a screenshot above and tried to highlight and emphasize that the warning about the "not free for professional use" part of the warning is not prominent enough and easy to miss.

Is there a way to style that text as a warning message with a different background? 

PS.: I am not a UX expert either

mxr576:
@rfay What is the next step here?
PR reopen? Moving the conversation to an issue?

rfay:
It's just not our problem to solve. It seems like it's your problem to solve in your organization. Each of the places we mention OrbStack has the situation clearly labeled. 

Thanks so much for stepping up with this suggestion, and we look forward to many more contributions!

mxr576:
Fair enough! Thanks 🍻

PR #7564, opened 2025-08-26
test: jq is not available on Windows, use docker run -i ddev/ddev-utilities

rfay:
## The Issue

A new test in the PHP add-on stuff used `jq` on the host, but it's not available on the Windows test runners, so TestCmdAddonPHP was failing in t.Cleanup()

Use `docker run -i ddev/ddev-utilities jq` instead

If tests pass it's good enough

### DIFF
--- a/cmd/ddev/cmd/addon_test.go
+++ b/cmd/ddev/cmd/addon_test.go
@@ -234,7 +234,7 @@ func TestCmdAddonPHP(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Cleanup(func() {
-		addonList, err := exec.RunHostCommand("bash", "-c", fmt.Sprintf("%s add-on list --installed -j | docker run -i ddev/ddev-utilities jq -r .raw.[].Name", DdevBin))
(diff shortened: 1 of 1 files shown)


cmd/ddev/cmd/addon_test.go
```diff
@@ -234,7 +234,7 @@ func TestCmdAddonPHP(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Cleanup(func() {
-		addonList, err := exec.RunHostCommand("bash", "-c", fmt.Sprintf("%s add-on list --installed -j | jq -r .raw.[].Name", DdevBin))
+		addonList, err := exec.RunHostCommand("bash", "-c", fmt.Sprintf("%s add-on list --installed -j | docker run -i ddev/ddev-utilities jq -r .raw.[].Name", DdevBin))
```
stasadev: `--rm` flag:

```suggestion
		addonList, err := exec.RunHostCommand("bash", "-c", fmt.Sprintf("%s add-on list --installed -j | docker run -i --rm ddev/ddev-utilities jq -r .raw.[].Name", DdevBin))
```

PR #7567, opened 2025-08-26
fix(magerun): use stable branch for autocompletion

stasadev:
## The Issue

- https://github.com/netz98/n98-magerun2/issues/1773

> The Debian bash autocomplete `n98-magerun2.phar` uses `local` where it can't be used, so everybody doing a `ddev ssh` in DDEV gets this since latest release:
> 
> ```
> rfay@ubuntu-len:~/workspace/d11$ ddev ssh
> bash: local: can only be used in a function
> bash: local: can only be used in a function
> ```

Uses the stable `master` branch as recommended in https://github.com/netz98/n98-magerun2/issues/1773#issuecomment-3224697637:

> > Is that what you'd recommend? It's been a lot of years. Is that still the recommended install technique?
> 
> The phar file is from stable release. The completion script is from unstable release. It whould be better to load the completion from stable "master" branch to have the completion compatible.

```
ddev ssh
```

Shouldn't have this output `bash: local: can only be used in a function`


rfay:
Thanks @cmuench !

PR #7568, opened 2025-08-26
test(buildkite): do better cleaning up volumes before running test

rfay:
## The Issue

We had some weird problems with ddev-router not being able to come ready due to leftover traefik config.

Apparently the traefik directory of the ddev-global-cache wasn't cleaned up by normal shutdown/poweroff

And apparently the attempt by buildkite to delete the ddev-global-cache volume failed

* Don't do `|| true` on key things like poweroff or volume removal

If it doesn't do any harm in the tests it's probably OK.

### DIFF
--- a/.buildkite/test.sh
+++ b/.buildkite/test.sh
@@ -193,7 +193,7 @@ ${TIMEOUT} 60s bash "$(dirname "$0")/testbot_maintenance.sh"
 
 # Our testbot should be sane, run the testbot checker to make sure.
 echo "--- running sanetestbot.sh"
-${TIMEOUT} 60s ./.buildkite/sanetestbot.sh
+${TIMEOUT} 60s bash "$(dirname "$0")/sanetestbot.sh"
 
 # Make sure we start with mutagen daemon off.
 unset MUTAGEN_DATA_DIRECTORY


stasadev (approved):
Looks good to me.

stasadev (commented):
Let's remove `|| true` from here:

https://github.com/ddev/ddev/blob/42229020e1ec82b4a88e4677ac57e2c8939abf8c/.buildkite/test.sh#L187-L191

This is the main place where the errors are skipped.

rfay:
Good idea. Removed that and cleaned up the timeout logic.

.buildkite/test.sh
```diff
@@ -184,15 +189,11 @@ fi
...
-# Prefer gtimeout (macOS). Do 60s timeout.
-# Continue anyway if trouble.
-"$(command -v gtimeout || command -v timeout)" 60s \
-  bash "$(dirname "$0")/testbot_maintenance.sh" || true
+${TIMEOUT} 60s bash "$(dirname "$0")/testbot_maintenance.sh"
 
 # Our testbot should be sane, run the testbot checker to make sure.
 echo "--- running sanetestbot.sh"
-"$(command -v gtimeout || command -v timeout)" 60s ./.buildkite/sanetestbot.sh
+${TIMEOUT} 60s ./.buildkite/sanetestbot.sh
```
stasadev: For consistency with `testbot_maintenance.sh` a few lines above:

```suggestion
${TIMEOUT} 60s bash "$(dirname "$0")/sanetestbot.sh"
```

Or change everything the other way around.

PR #7569, opened 2025-08-27
chore(mkdocs): disable privacy plugin for local builds, for #6027 [skip ci]

stasadev:
## The Issue

- #6027

The `privacy` plugin generates `.cache` directory locally, but:

I don't like it because if it's cached, we can't see if there are any issues with external sources while working locally (and I don't like having one more `.cache` here because if I checkout some old git commit, this directory appears in unstaged files.)

- Used variable: https://docs.readthedocs.com/platform/stable/reference/environment-variables.html#envvar-READTHEDOCS
- Material for MkDocs config https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.enabled
  > If you want to disable the plugin, e.g., for local builds, you can use an environment variable in mkdocs.yml
- I didn't remove `.cache` from `.gitignore`

Remotely, nothing changes for loaded sources:

image

---

Run this locally:

```
# in ddev directory
# rm -r .cache

# no `.cache` directory created
make mkdocs

# `.cache` directory created
READTHEDOCS=True make mkdocs
```

---

```
make mkdocs-serve
```

image

---

```
READTHEDOCS=True make mkdocs-serve
```

image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -69,7 +69,8 @@ theme:
 # Plugins
 plugins:
 - search
-- privacy
+- privacy:
+    enabled: !ENV [READTHEDOCS, false]
 - redirects:
     redirect_maps:
 - minify:


rfay (approved):
Sure seems like it can't do any harm!

PR #7572, opened 2025-08-28
feat: provide `host.docker.internal` for all services, fixes #7537

### LINKED ISSUE #7537: Xdebug: figure out a better way for `host.docker.internal` inside all containers

DDEV has `host.docker.internal` discovery logic for the `web` container.

[FrankenPHP add-on](https://github.com/stasadev/ddev-frankenphp) uses the `frankenphp` container, and we need a simple way to connect to Xdebug. 

Ideally, `host.docker.internal` should be available in all containers by default, whether it’s needed or not.

The same can be said for using some custom PHP https://ddev.com/blog/legacy-projects-with-unsupported-php-and-mysql-using-ddev/

### Describe your solution

We already have a condition in the compose YAML template:

https://github.com/ddev/ddev/blob/d329e8d48e6c8636a4b107bca1bbf870d1e9df84/pkg/ddevapp/app_compose_template.yaml#L265-L270

This should be moved to `ddevapp.fixupComposeYaml` so it applies to all containers. The original template logic for this can be removed.

### Describe alternatives

_No response_

### Additional context

_No response_

stasadev:
## The Issue

- Fixes #7537

Consolidates `host.docker.internal` configuration by moving it from template-time to compose post-processing.

- Remove `.HostDockerInternalIP` and `.UseHostDockerInternalExtraHosts` from the template 
- Add `host.docker.internal` handling in `fixupComposeYaml()` for all services
- Add `HOST_DOCKER_INTERNAL_IP` env variable for all services

```
ddev config --webserver-type=generic
ddev add-on get stasadev/ddev-frankenphp
ddev restart

for s in $(ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.services | keys[]'); do
  echo "Service: $s"
  ddev exec -s "$s" getent hosts host.docker.internal
  ddev exec -s "$s" 'echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP'
  echo
done
```

Testing on Linux:

v1.24.7:
```
Service: db
Failed to execute command `getent hosts host.docker.internal`: exit status 2
HOST_DOCKER_INTERNAL_IP=

Service: frankenphp
Failed to execute command `getent hosts host.docker.internal`: exit status 2
sh: 1: HOST_DOCKER_INTERNAL_IP: parameter not set
Failed to execute command `echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP`: exit status 2

Service: web
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=
```

This PR:
```
Service: db
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1

Service: frankenphp
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1

Service: web
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1
```

---

```
ddev config global --xdebug-ide-location=192.168.1.1

ddev config --webserver-type=generic
ddev add-on get stasadev/ddev-frankenphp
ddev restart

for s in $(ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.services | keys[]'); do
  echo "Service: $s"
  ddev exec -s "$s" getent hosts host.docker.internal
  ddev exec -s "$s" 'echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP'
  echo
done

ddev config global --xdebug-ide-location=
```

v1.24.7:
```
Service: db
Failed to execute command `getent hosts host.docker.internal`: exit status 2
HOST_DOCKER_INTERNAL_IP=192.168.1.1

Service: frankenphp
Failed to execute command `getent hosts host.docker.internal`: exit status 2
sh: 1: HOST_DOCKER_INTERNAL_IP: parameter not set
Failed to execute command `echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP`: exit status 2

Service: web
192.168.1.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=192.168.1.1
```

This PR:
```
Service: db
192.168.1.1     host.docker.internal
HOST_DOCKER_INTERNAL_IP=192.168.1.1

Service: frankenphp
192.168.1.1     host.docker.internal
HOST_DOCKER_INTERNAL_IP=192.168.1.1

Service: web
192.168.1.1     host.docker.internal
HOST_DOCKER_INTERNAL_IP=192.168.1.1
```

### DIFF
--- a/pkg/ddevapp/config.go
+++ b/pkg/ddevapp/config.go
@@ -896,74 +896,72 @@ func (app *DdevApp) FixObsolete() {
 }
 
 type composeYAMLVars struct {
-	Name                            string
-	Plugin                          string
-	AppType                         string
-	WebserverType                   string
-	MailpitPort                     string
-	HostMailpitPort                 string
-	DBType                          string
-	DBVersion                       string
-	DBMountDir                      string
-	DBAPort                         string
-	DBPort                          string
-	DdevGenerated                   string
-	HostDockerInternalIP            string
-	NFSServerAddr                   string
-	DisableSettingsManagement       bool
-	MountType                       string
-	WebMount                        string
-	WebBuildContext                 string
-	DBBuildContext                  string
-	WebBuildDockerfile              string
-	DBBuildDockerfile               string
-	SSHAgentBuildContext            string
-	OmitDB                          bool
-	OmitDBA                         bool
-	OmitRouter                      bool
-	OmitSSHAgent                    bool
-	BindAllInterfaces               bool
-	MariaDBVolumeName               string
-	PostgresVolumeName              string
-	MutagenEnabled                  bool
-	MutagenVolumeName               string
-	NFSMountEnabled                 bool
-	NFSSource                       string
-	NFSMountVolumeName              string
-	DockerIP                        string
-	IsWindowsFS                     bool
-	NoProjectMount                  bool
-	Hostnames                       []string
-	Timezone                        string
-	ComposerVersion                 string
-	Username                        string
-	UID                             string
-	GID                             string
-	FailOnHookFail                  bool
-	WebWorkingDir                   string
-	DBWorkingDir                    string
-	DBAWorkingDir                   string
-	WebEnvironment                  []string
-	NoBindMounts                    bool
-	Docroot                         string
-	UploadDirsMap                   []string
-	GitDirMount                     bool
-	IsGitpod                        bool
-	IsCodespaces                    bool
-	DefaultContainerTimeout         string
-	UseHostDockerInternalExtraHosts bool
-	WebExtraContainerPorts          []int
-	WebExtraHTTPPorts               string
-	WebExtraHTTPSPorts              string
-	WebExtraExposedPorts            string
-	BitnamiVolumeDir                string
-	UseHardenedImages               bool
-	XHGuiHTTPPort                   string
-	XHGuiHTTPSPort                  string
-	XHGuiPort                       string
-	HostXHGuiPort                   string
-	XhguiImage                      string
-	XHProfMode                      types.XHProfMode
+	Name                      string
+	Plugin                    string
+	AppType                   string
+	WebserverType             string
+	MailpitPort               string
+	HostMailpitPort           string
+	DBType                    string
+	DBVersion                 string
+	DBMountDir                string
+	DBAPort                   string
+	DBPort                    string
+	DdevGenerated             string
+	NFSServerAddr             string
+	DisableSettingsManagement bool
+	MountType                 string
+	WebMount                  string
+	WebBuildContext           string
+	DBBuildContext            string
+	WebBuildDockerfile        string
+	DBBuildDockerfile         string
+	SSHAgentBuildContext      string
+	OmitDB                    bool
+	OmitDBA                   bool
+	OmitRouter                bool
+	OmitSSHAgent              bool
+	BindAllInterfaces         bool
+	MariaDBVolumeName         string
+	PostgresVolumeName        string
+	MutagenEnabled            bool
+	MutagenVolumeName         string
+	NFSMountEnabled           bool
+	NFSSource                 string
+	NFSMountVolumeName        string
+	DockerIP                  string
+	IsWindowsFS               bool
+	NoProjectMount            bool
+	Hostnames                 []string
+	Timezone                  string
+	ComposerVersion           string
+	Username                  string
+	UID                       string
+	GID                       string
+	FailOnHookFail            bool
+	WebWorkingDir             string
+	DBWorkingDir              string
+	DBAWorkingDir             string
+	WebEnvironment            []string
+	NoBindMounts              bool
+	Docroot                   string
+	UploadDirsMap             []string
+	GitDirMount               bool
+	IsGitpod                  bool
+	IsCodespaces              bool
+	DefaultContainerTimeout   string
+	WebExtraContainerPorts    []int
+	WebExtraHTTPPorts         string
+	WebExtraHTTPSPorts        string
+	WebExtraExposedPorts      string
+	BitnamiVolumeDir          string
+	UseHardenedImages         bool
+	XHGuiHTTPPort             string
+	XHGuiHTTPSPort            string
+	XHGuiPort                 string
+	HostXHGuiPort             string
+	XhguiImage                string
+	XHProfMode                types.XHProfMode
 }
 
 // RenderComposeYAML renders the contents of .ddev/.ddev-docker-compose*.
@@ -1019,7 +1017,6 @@ func (app *DdevApp) RenderComposeYAML() (string, error) {
 		DBMountDir:                "/var/lib/mysql",
 		DBPort:                    GetInternalPort(app, "db"),
 		DdevGenerated:             nodeps.DdevFileSignature,
-		HostDockerInternalIP:      hostDockerInternal.IPAddress,
 		NFSServerAddr:             nfsServerAddr,
 		DisableSettingsManagement: app.DisableSettingsManagement,
 		OmitDB:                    nodeps.ArrayContainsString(app.GetOmittedContainers(), nodeps.DBContainer),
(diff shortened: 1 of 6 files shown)


stasadev:
This PR is in draft, waiting for:

- #7574

pkg/ddevapp/config.go
```diff
@@ -896,85 +896,79 @@ func (app *DdevApp) FixObsolete() {
...
+	WebExtraHTTPSPorts        string
+	WebExtraExposedPorts      string
+	BitnamiVolumeDir          string
+	UseHardenedImages         bool
+	XHGuiHTTPPort             string
+	XHGuiHTTPSPort            string
+	XHGuiPort                 string
+	HostXHGuiPort             string
+	XhguiImage                string
+	XHProfMode                types.XHProfMode
```
stasadev, line 964: I removed:

- `HostDockerInternalIP`
- `UseHostDockerInternalExtraHosts`

stasadev:
The next PR for Docker labels will depend on this, because I have added `TestFixupComposeYaml` here, which will be used later.

pkg/dockerutil/dockerutils.go
```diff
@@ -1418,7 +1408,10 @@ func CreateVolume(volumeName string, driver string, driverOpts map[string]string
 	return vol, err
 }
 
-// GetHostDockerInternalIP returns either "" (will use the hostname as is)
+var HostDockerInternalIP = "empty"
```
rfay: I'm not much of a fan of using a magic string like "empty".  I'd be a little happier with "notset", but not that much. You didn't like the empty string as a marker? It seemed clearer.

pkg/dockerutil/dockerutils.go
```diff
@@ -1418,7 +1408,10 @@ func CreateVolume(volumeName string, driver string, driverOpts map[string]string
...
+var HostDockerInternalValue = "empty"
```
rfay: Please add a comment indicating what this is (as opposed to the IP)

pkg/dockerutil/dockerutils.go
```diff
@@ -1431,72 +1424,86 @@ func CreateVolume(volumeName string, driver string, driverOpts map[string]string
 // Inside WSL2, the way to access an app like PhpStorm running on the Windows side is described
 // in https://learn.microsoft.com/en-us/windows/wsl/networking#accessing-windows-networking-apps-from-linux-host-ip
 // and it involves parsing /etc/resolv.conf.
-func GetHostDockerInternalIP() (string, error) {
-	hostDockerInternal := ""
+func GetHostDockerInternal() (string, string) {
```
rfay: I don't think the new signature (string, string) is documented properly here in the header. Please explain what these two values are.

pkg/dockerutil/dockerutils.go
stasadev: I used "empty" word that is not an empty string because otherwise the function logic will be executed again and again on Linux, where it's empty by default:

```go
	if HostDockerInternalIP != "empty" && HostDockerInternalValue != "empty" {
		return HostDockerInternalIP, HostDockerInternalValue
	}
```

We either need to compare it against something that is not an empty string or cache it differently.

pkg/dockerutil/dockerutils.go
stasadev: OK, It's some kind of cache to not perform checks for IP and for Linux again and again to get a "host-gateway" value.

pkg/dockerutil/dockerutils.go
stasadev: Okay.

stasadev:
Rebased on top of:

- #7587

rfay:
Ready for rebase.

rpkoller:
the steps i did on DDEV HEAD, the latest Lima stable on macos. the outputs i've received::

```
ddev config --webserver-type=generic
ddev add-on get stasadev/ddev-frankenphp
ddev start
```
entering 

```
for s in $(ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.services | keys[]'); do
  echo "Service: $s"
  ddev exec -s "$s" getent hosts host.docker.internal
  ddev exec -s "$s" 'echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP'
  echo
done
```

i've received

```
Service: db
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=

Service: frankenphp
192.168.5.2     host.lima.internal host.docker.internal
sh: 1: HOST_DOCKER_INTERNAL_IP: parameter not set
Failed to execute command `echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP`: exit status 2

Service: web
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=
```

stopped swtiched to the PR artifact and ran the same command again and got: 

```
Service: db
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=

Service: frankenphp
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=

Service: web
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=
```

now switched back to HEAD and ran  `ddev config global --xdebug-ide-location=172.17.0.1`and ran 

```
for s in $(ddev describe -j | docker run -i --rm ddev/ddev-utilities jq -r '.raw.services | keys[]'); do
  echo "Service: $s"
  ddev exec -s "$s" getent hosts host.docker.internal
  ddev exec -s "$s" 'echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP'
  echo
done
```

now received: 

```
Service: db
192.168.5.2     host.lima.internal host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1

Service: frankenphp
192.168.5.2     host.lima.internal host.docker.internal
sh: 1: HOST_DOCKER_INTERNAL_IP: parameter not set
Failed to execute command `echo HOST_DOCKER_INTERNAL_IP=$HOST_DOCKER_INTERNAL_IP`: exit status 2

Service: web
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1
```

switched then another time over to the artifact and ran the command: 

```
Service: db
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1

Service: frankenphp
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1

Service: web
172.17.0.1      host.docker.internal
HOST_DOCKER_INTERNAL_IP=172.17.0.1
```

so the results look identical

rpkoller:
and is it intentional to have an additional binary `ddev-hostname`? cuz so far everything was included in a single `ddev` binary?

stasadev:
> and is it intentional to have an additional binary `ddev-hostname`? cuz so far everything was included in a single `ddev` binary? 

Yes, see https://ddev.com/blog/ddev-hostname-security-improvements/

rpkoller:
> > and is it intentional to have an additional binary `ddev-hostname`? cuz so far everything was included in a single `ddev` binary?
> 
> Yes, see https://ddev.com/blog/ddev-hostname-security-improvements/

uhhh i must have either forgot about it or must have lived under a rock. i really havent noticed that there is a second binary for head now - completely missed that. thanks for the link!

rfay:
This seems to be working fine, but what's the environment variable HOST_DOCKER_INTERNAL_IP for if it's always left empty (even on Linux) (as it is in your results in the OP)

But then there's the question of what HOST_DOCKER_INTERNAL_IP is for. I don't see it documented anywhere, and don't remember any use, and don't see references in the code. Maybe it's a special-usage thing for one environment?

stasadev:
I don't know why we need HOST_DOCKER_INTERNAL_IP, let's remove it.

rfay:
OK. I sure can't find anything in usage or even in history that justifies it.

rfay:
Another option is that we could keep it, but populate it.

stasadev:
> Another option is that we could keep it, but populate it.

Good idea, implemented it.

stasadev:
But I think the `HOST_DOCKER_INTERNAL_IP` will still be empty in case of using Docker Desktop.

rfay:
I think what you did is OK, but I'm not sure my thinking is all that mature. It would be better probably if the env var were always populated with the IP address (including non-linux). The lookup is not costly, and it only happens during `ddev start`, but it also seems kind of silly since we don't know of a use for it.

rfay:
I see two exact same failures in https://buildkite.com/ddev/macos-colima-vz/builds/4429#01992538-08ab-4d87-abc9-e59a717c0b18 which I'm sure have nothing to do with this PR. And on tb-macos-arm64-7 (both) there was a popup about using Apple's password manager instead of keychain-access. 

I rebooted; but since this happened exactly the same way twice, I thought I should mention it. We may have something asking for password or some kind of access on tb-macos-arm64-7 or other places.

rfay (approved):
It looks good to me, thanks. 

If the tb-macos-arm64-7 fails again I'll try to reproduce locally with colima

